Issue #20, opened 2019-12-03
tracee in container


itaysk:
tracee should be able to run in a container. this issue is for tracking container related requests and tasks.

PR #22, opened 2019-12-03
add dockerfile

itaysk:
assuming linux headers are present on the host and mounted in as well as debugfs
related: #20


itaysk:
pushed new change that mounts debugfs in the container, it may not be present on the host

PR #24, opened 2019-12-17
performance: improve performance and reduce lost events

yanivagman:


tracee/container_event_monitor_ebpf.c
```diff
@@ -15,7 +15,7 @@
 #include <linux/version.h>
 
 #define MAX_STRING_SIZE 4096                                // Choosing this value to be the same as PATH_MAX
-#define SUBMIT_BUFSIZE  524288                              // Need to be power of 2
+#define SUBMIT_BUFSIZE  (2 << BUF_SZ_EXPO)                  // Need to be power of 2
```
simar7: Where is `BUF_SZ_EXPO` defined?

simar7:
I must be doing something wrong here...

Using the defaults:
```
vagrant@vagrant:~/repos/contracer$ sudo ./start.py -j
Possibly lost event data! Consider using bigger submission buffer size
Possibly lost event data! Consider using bigger submission buffer size
Possibly lost event data! Consider using bigger submission buffer size
Possibly lost event data! Consider using bigger submission buffer size
<snip>
```

Trying with a bigger buffer size:

```
vagrant@vagrant:~/repos/contracer$ sudo ./start.py -b 20
TIME(s)        UTS_NAME         MNT_NS       PID_NS       UID    EVENT            COMM             PID    TID    PPID   RET          ARGS
Possibly lost 195 samples
Possibly lost 95 samples
Possibly lost 7 samples
Possibly lost 326 samples
Possibly lost 4 samples
<snip>
```

I also noticed that Tracee is really slow to startup up before the first events are logged. This definitely wasn't the case before.

tracee/container_tracer.py
```diff
@@ -1129,17 +1125,43 @@ def print_event(self, cpu, data, size):
         # log.info(json.dumps(events, indent=4))
         # exit()
 
+    # process event
+    def handle_event(self, cpu, data, size):
```
simar7: `size` is not used, remove.

simar7:
OK some profiling results:

```
Timer unit: 1e-06 s

Total time: 3.09307 s
File: tracee/container_tracer.py
Function: handle_event at line 1130

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
  1130                                               @profile
  1131                                               def handle_event(self, cpu, data, size):
  1132      1561       4060.0      2.6      0.1          self.prev_time = time.time()
  1133      1561      27873.0     17.9      0.9          event = self.bpf["events"].event(data)
  1134      1561    1448857.0    928.2     46.8          events_buf = self.bpf["submission_buf"][cpu].buf
  1135      1561    1573674.0   1008.1     50.9          events_buf_head = self.bpf["submission_buf"][cpu].head
  1136      1561      19027.0     12.2      0.6          if (events_buf_head - event.start_off) > (2**(self.buf_size-1)):
  1137                                                       print("Possibly lost event data! Consider using bigger submission buffer size")
  1138      1561       1337.0      0.9      0.0          if (event.end_off > event.start_off):
  1139      1561       1110.0      0.7      0.0              event_size = event.end_off - event.start_off
  1140      1561      15084.0      9.7      0.5              event_buf = (ctypes.c_char * event_size).from_buffer_copy(events_buf, event.start_off)
  1141                                                   else:
  1142                                                       event_size = event.max_off - event.start_off + event.end_off
  1143                                                       event_buf = (ctypes.c_char * event_size)()
  1144                                                       first_chunk_sz = event.max_off - event.start_off
  1145                                                       ctypes.memmove(event_buf, ctypes.byref(events_buf, event.start_off), first_chunk_sz)
  1146                                                       ctypes.memmove(ctypes.byref(event_buf, first_chunk_sz), ctypes.byref(events_buf, 0), event.end_off)
  1147      1561       2044.0      1.3      0.1          self.event_bufs.append(event_buf)
```

Indexing the `self.bpf["submission_buf"][cpu]` array is expensive as it can be seen here.

One improvement we can make is the following:
```diff --git a/tracee/container_tracer.py b/tracee/container_tracer.py
index 119391c..a9d0b52 100755
--- a/tracee/container_tracer.py
+++ b/tracee/container_tracer.py
@@ -835,6 +835,7 @@ class EventMonitor:
     # process event
     def handle_event(self, cpu, data, size):
         self.prev_time = time.time()
+        subbuf = self.bpf["submission_buf"][cpu]
         event = self.bpf["events"].event(data)
-        events_buf = self.bpf["submission_buf"][cpu].buf
-        events_buf_head = self.bpf["submission_buf"][cpu].head
+        events_buf = subbuf.buf
+        events_buf_head = subbuf.head
 ```

New profiling results:
```
Total time: 1.44939 s
tracee/container_tracer.py
Function: handle_event at line 1130

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
  1130                                               @profile
  1131                                               def handle_event(self, cpu, data, size):
  1132      2698       8204.0      3.0      0.6          self.prev_time = time.time()
  1133      2698    1345337.0    498.6     92.8          subbuf = self.bpf["submission_buf"][cpu]
  1134      2698      42021.0     15.6      2.9          event = self.bpf["events"].event(data)
  1135      2698       8070.0      3.0      0.6          events_buf = subbuf.buf
  1136      2698       1439.0      0.5      0.1          events_buf_head = subbuf.head
```

This reduces the runtime by nearly 50%.

tracee/container_tracer.py
yanivagman: This is a callback function of open_perf_buffer - signature can not be changed

yanivagman:
> I must be doing something wrong here...
> 
> Using the defaults:
> 
> ```
> vagrant@vagrant:~/repos/contracer$ sudo ./start.py -j
> Possibly lost event data! Consider using bigger submission buffer size
> Possibly lost event data! Consider using bigger submission buffer size
> Possibly lost event data! Consider using bigger submission buffer size
> Possibly lost event data! Consider using bigger submission buffer size
> <snip>
> ```
> 
> Trying with a bigger buffer size:
> 
> ```
> vagrant@vagrant:~/repos/contracer$ sudo ./start.py -b 20
> TIME(s)        UTS_NAME         MNT_NS       PID_NS       UID    EVENT            COMM             PID    TID    PPID   RET          ARGS
> Possibly lost 195 samples
> Possibly lost 95 samples
> Possibly lost 7 samples
> Possibly lost 326 samples
> Possibly lost 4 samples
> <snip>
> ```
> 
> I also noticed that Tracee is really slow to startup up before the first events are logged. This definitely wasn't the case before.

The default buffer size is actually the biggest (2^20). When I tried to change it to bigger values - I got not enough memory messages, so actually, you can keep using the defaults. Also note that you used -j, which gives json output. I'm not sure why you don't get the actuall output (both in json and regular output) - have you removed the logs from code?

Regarding lost samples/"Possibly lost event data..." messages - this can happen when event rate is high - it is a known issue of Tracee, which I try to reduce all the time.

Using python cProfile I see that the startup time is due to attaching kprobes/kretprobes. The more events we will have, the stratup time will take longer - there is nothing to do about that. As I didn't add any new event in this patch - I don't see a reason why startup time will be longer - are you sure you compared it to latest master?

tracee/container_tracer.py
yanivagman: This is a callback function - can't change arguments here

yanivagman:
> This reduces the runtime by nearly 50%.

Good catch - I'll update the commit with that change

yanivagman:
> Indexing the `self.bpf["submission_buf"][cpu]` array is expensive as it can be seen here.

I discovered an interesting performance penalty here, after investigating this line.
In this commit, I also doubled the default buffer size (2^19 to 26^20) and I can see that this change more than doubles the time spent in this line.

What I conclude is that retreiving the map with this line doesn't give a reference address, but fully copies the events buffer in each access!

I initialy used a submission buffer as a map because it can be much bigger than the perf event buffer (which is limited to 2^17 in it's size, thus can't hold a lot of events). But this is a tradeoff we will have to take into account, or find a more proper solution

yanivagman:
Closed in favor of https://github.com/aquasecurity/tracee/pull/27

PR #37, opened 2019-12-26
avoid fork handler code duplication

yanivagman:


tracee/event_monitor_ebpf.c
```diff
@@ -911,7 +911,7 @@ static __always_inline int trace_ret_generic(struct pt_regs *ctx, u32 id, u64 ty
     context_t context = {};
 
     if (init_context(&context) || init_submit_buf())
-        return 0;
+        return -1;
```
itaysk: comment without action items: this function is called by kprobe callback and the return value is returned as-is to the caller, which is not in tracee. couldn't find documentation on the meaning of the return value of a kprobe callback, but all the examples in BCC returns 0 (on all cases). it seems that returning -1 is harmless, but if we ever find out it's not we could translate this internal return value to an external return value in the wrapping function (created by TRACE_RET_FUNC macro)

PR #42, opened 2019-12-30
Add optional exec-env flag to show env in execve

yanivagman:


tracee/event_monitor_ebpf.c
```diff
@@ -758,6 +764,26 @@ static __always_inline int save_argv(submit_buf_t *submit_p, struct pt_regs *ctx
     return 0;
 }
 
+static __always_inline int save_str_arr(submit_buf_t *submit_p, const char __user *const __user *ptr)
```
itaysk: can you suffix the name with `_to_buf` so it's consistent with other similar functions?

tracee/event_monitor_ebpf.c
```diff
@@ -1114,26 +1140,15 @@ int syscall__execve(struct pt_regs *ctx,
         return 0;
 
     context.eventid = SYS_EXECVE;
-    context.argnum = 2;
+    context.argnum = 3;
```
itaysk: what if `show_evex_args` is not set and there are just 2 args saved?

tracee/event_monitor_ebpf.c
yanivagman: Will do

tracee/event_monitor_ebpf.c
yanivagman: You are right. I'll fix it

PR #44, opened 2020-01-02
Add sockaddr struct fields for unix, inet, inet6 sockets

yanivagman:


tracee/tracer.py
```diff
@@ -950,10 +1036,9 @@ def parse_event(self, event_buf):
                 elif argtype == ArgType.STR_T:
                     args.append(self.get_string_from_buf(event_buf))
                 elif argtype == ArgType.STR_ARR_T:
-                    self.get_str_arr_from_buf(event_buf, args)
+                    args.append(self.get_str_arr_from_buf(event_buf))
```
itaysk: Just to be clear, is this change related to the topic of the PR (adding support for sockets?)

yanivagman: No, not directly. I changed it to be in the same convention as other types

PR #46, opened 2020-02-16
Add lsm bprm_check hook to get exec absolute path

yanivagman:


tracee/event_monitor_ebpf.c
```diff
@@ -1304,6 +1387,28 @@ int trace_do_exit(struct pt_regs *ctx, long code)
...
+    submit_buf_t *submit_p = get_submit_buf();
+    if (submit_p == NULL)
+        return 0;
+
+    context.eventid = SECURITY_BPRM_CHECK;
+    context.argnum = 1;
+    context.retval = 0;
+
+    save_to_submit_buf(submit_p, (void*)&context, sizeof(context_t), NONE_T);
+    save_path_to_buf(submit_p, &bprm->file->f_path);
```
itaysk: what about the  return value here?

yanivagman: I neither check for return values in save_to_submit_buf and save_str_to buf - same here.
You are right it is a better practice to check for return values, and we should probably do a separate commit for this in the future.

PR #47, opened 2020-02-16
Rewrite Python code in Go

itaysk:
This is the initial rewrite of the userspace code in Go. there's still some work to do, but I'd like to merge it now as it's stable, and continue incrementally improving with additional PRs. This PR should have feature parity with commit 431eaae7d4107cb6179ee9486f70d9435e44ebe3.

Notes:

- There are some minor UX changes. For example: timestamp formatting, args are printed inside brackets, new output flag, events flag accepts one flag at a time but can be defined multiple times.

- @simar7  correctly noted that consts names for the argtypes doesn't match Go conventions (which say consts are no different than vars). However, after reading about it a bit, and considering that those consts are borrowed from the C and Python code, I'd like to keep it as is (CAPS_SNAKE). This is also demonstrated in the Go codebase where they borrow const names from C.

- The `context` struct field names are based on the `context` struct from the C code. Unlike the discussion about consts names, here this is not borrowed but redefined, hence I think it should be following the Go naming contentions. It was a bit challenging following go's "initialism" convention: https://github.com/golang/go/wiki/CodeReviewComments#initialisms since the names of the context fields are all initials, and exported, so technically should be all upper cased, but that didn't feel right to me so I ignored this convention in this case and treated those fields as regular names. Hence some linter errors which I'm fine with that in this particular case.

- I'm still deliberating the way Syscalls/Sysevents are defined as map (it was done to make lookup simple)

- @yanivagman pointed to 2  gobpf related issues: how to set buffer size, and how to handle lost events. I'll address those in a future PR.

- Tests are missing, as they are in the python version. We'll continue with what @simar7 has started in another PR.

I'll open a GH issue for each item but adding as well.


Makefile.old
```diff
@@ -0,0 +1,2 @@
+test:
```
simar7: Could we this to the same Makefile but under a different target? Something like `python-test` for instance.

tracee/printer.go
```diff
@@ -0,0 +1,62 @@
...
+
+func (p jsonEventPrinter) Preamble() {}
+
+func (p jsonEventPrinter) Print(ctx context, args []interface{}) {
+	e := printableEvent{context: ctx, EventName: getEventName(ctx.Eventid), Args: args}
+	eBytes, err := json.Marshal(e)
+	if err != nil {
+		fmt.Printf("error printing event: %v\n", err)
+	}
+	fmt.Printf("%s", string(eBytes))
```
yanivagman: In the python code we support a function named "get_events" which retreives the json output when needed.
We need to think of a way to mimic that here, as it is needed by ultrabox - maybe output to a file?

tracee/argprinters.go
```diff
@@ -0,0 +1,340 @@
+package tracee
+
+import (
+	"strconv"
+	"strings"
+)
+
+// PrintMknodMode prints the `mode` bitmask argument of the `mknod` syscall
+// http://man7.org/linux/man-pages/man7/inode.7.html
+func PrintMknodMode(mode uint32) string {
```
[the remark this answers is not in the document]
itaysk: Thanks. I still want to "qualify" the otherwise generic argument name "mode". How about `PrintInodeMode`?

tracee/argprinters.go
```diff
@@ -0,0 +1,340 @@
...
+		}
+	}
+
+	return strings.Join(f, "|")
+}
+
+// PrintMmapProt prints the `prot` bitmask argument of the `mmap` syscall
+// http://man7.org/linux/man-pages/man2/mmap.2.html
+// https://elixir.bootlin.com/linux/v5.5.3/source/include/uapi/asm-generic/mman-common.h#L10
+func PrintMmapProt(prot uint32) string {
```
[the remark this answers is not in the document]
itaysk: How about `PrintMemoryProt` (memprot?)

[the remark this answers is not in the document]
yanivagman: This can work

tracee/argprinters.go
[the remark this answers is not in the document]
yanivagman: Sounds good to me. Another option - PrintFileMode

tracee/printer.go
itaysk: as discussed, ultrabox cannot import this module anymore as it's now golang; ultrabox will have to exec the process. so either read stdout or redirect output to a file and read it.

tracee/tracee.go
```diff
@@ -0,0 +1,487 @@
...
+}
+
+// contex struct contains common metadata that is collected for all types of events
+type context struct {
+	Ts      uint64   `json:"ts"`
+	Pid     uint32   `json:"pid"`
+	Tid     uint32   `json:"tid"`
+	Ppid    uint32   `json:"ppid"`
+	Uid     uint32   `json:"uid"`
+	MntId   uint32   `json:"mnt_id"`
```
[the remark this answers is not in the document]
itaysk: I used the same names as in the `context_t` struct. will use the names from the python `print_event`

tracee/tracee.go
```diff
@@ -0,0 +1,487 @@
...
+	eventsChannel  chan []byte
+	printer        eventPrinter
+}
+
+// New creates a new Tracee instance based on a given valid TraceeConfig
+func New(cfg TraceeConfig) (*Tracee, error) {
+	var err error
+
+	// validation
+	err = cfg.Validate()
```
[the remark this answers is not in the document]
itaysk: The state of mind here is that tracee is a library. The part of creating the configuration is a distinct operation from running tracee, and at this point we can already fail early if the input is user provided invalid. The `New` function is the entry point for the library, and it should be able to care for itself and validate input regardless of if a user validated before. They might create the config manually instead of using the `NewConfig` helper, or change the config before passing it to `New`.

tracee/tracee.go
```diff
@@ -0,0 +1,487 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("validation error: %v", err)
+	}
+	bpfFile := "./tracee/event_monitor_ebpf.c"
```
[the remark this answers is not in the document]
itaysk: I don't think this should be a package level constant. In some point we might want to make it configurable or load an elf instead or some other thing. by making this data a function level local instead of a package level global I "hide" it from the library users, and reserve the right to change it's behavior in the future.

tracee/tracee.go
```diff
@@ -0,0 +1,487 @@
...
+
+	bpfConfig := bpf.NewTable(t.bpfModule.TableId("config_map"), t.bpfModule)
+	key := make([]byte, 4)
+	leaf := make([]byte, 4)
+	binary.LittleEndian.PutUint32(key, uint32(CONFIG_CONT_MODE))
+	binary.LittleEndian.PutUint32(leaf, boolToUInt32(t.config.ContainerMode))
+	bpfConfig.Set(key, leaf)
+	binary.LittleEndian.PutUint32(key, uint32(CONFIG_DETECT_ORIG_SYSCALL))
+	binary.LittleEndian.PutUint32(leaf, boolToUInt32(t.config.DetectOriginalSyscall))
+	bpfConfig.Set(key, leaf)
```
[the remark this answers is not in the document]
itaysk: the table is initialized in the bpf code, no? I just checked `-c` and it works

tracee/tracee.go
```diff
@@ -0,0 +1,487 @@
...
+
+	eventsBPFTable := bpf.NewTable(t.bpfModule.TableId("events"), t.bpfModule)
+	t.eventsChannel = make(chan []byte, 1000)
```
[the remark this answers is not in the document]
itaysk: no good reason. will do in the future

tracee/printer.go
```diff
@@ -0,0 +1,62 @@
...
+	fmt.Println()
+}
+
+func (p tableEventPrinter) Print(ctx context, args []interface{}) {
+	fmt.Printf("%-14d %-16s %-12d %-12d %-6d %-16s %-16s %-6d %-6d %-6d %-12d", ctx.Ts/1000000, ctx.UtsName, ctx.MntId, ctx.PidId, ctx.Uid, getEventName(ctx.Eventid), ctx.Comm, ctx.Pid, ctx.Tid, ctx.Ppid, ctx.Retval)
+	fmt.Printf("%v", args)
+	fmt.Println()
+}
+
+func (p tableEventPrinter) Epilogue() {}
```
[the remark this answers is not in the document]
itaysk: noted (out of scope)

tracee/tracee.go
```diff
@@ -0,0 +1,487 @@
...
+	t.bpfPerfMap, err = bpf.InitPerfMap(eventsBPFTable, t.eventsChannel)
```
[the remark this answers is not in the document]
itaysk: will do in the future

tracee/tracee.go
```diff
@@ -0,0 +1,487 @@
...
+	return uint32(0)
+}
+
+func (t Tracee) processEvents() {
+	for {
+		dataRaw := <-t.eventsChannel
+		dataBuff := bytes.NewBuffer(dataRaw)
+		ctx, err := readContextFromBuff(dataBuff)
+		if err != nil {
+			continue
```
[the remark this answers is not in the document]
itaysk: noted (out of scope)

itaysk:
Thanks @yanivagman ! You really caught some important issues there. 
I've addressed most of your comments:
- Those that I just "resolved" I fixed. 
- Some I didn't fix and answered why. 
- The one about not printing unwanted events (left opened) need some more planning, will to that tomorrow. 
- Some I wrote down to add in another PR:
  - show events and lost/err event count https://github.com/aquasecurity/tracee/pull/47/files#r381269827
https://github.com/aquasecurity/tracee/pull/47/files#r381326662
  - consider events channel size
https://github.com/aquasecurity/tracee/pull/47/files#r381313817
  - configurable perf map size 
https://github.com/aquasecurity/tracee/pull/47/files#r381322134

tracee/tracee.go
```diff
@@ -0,0 +1,489 @@
...
+	}
+	return string(tc[:len])
+}
+
+// MarshalText implements the TextMarshaler interface, which is used by JSONMarshaler and others
+func (tc taskComm) MarshalText() ([]byte, error) {
+	return []byte(tc.String()), nil
+}
+
+// contex struct contains common metadata that is collected for all types of events
```
yanivagman: contex->context

tracee/tracee.go
```diff
@@ -0,0 +1,487 @@
...
+			continue
+		}
+		args := make([]interface{}, ctx.Argnum)
+		for i := 0; i < int(ctx.Argnum); i++ {
+			args[i], err = readArgFromBuff(dataBuff)
+			if err != nil {
+				continue
+			}
+		}
+		t.printer.Print(ctx, args)
```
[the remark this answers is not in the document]
yanivagman: What about this one?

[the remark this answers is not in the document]
itaysk: As I wrote:
> The one about not printing unwanted events (left opened) need some more planning, will to that tomorrow.

I'll let you know once ready to re-review

itaysk:
@yanivagman in order to fix https://github.com/aquasecurity/tracee/pull/47#discussion_r381311011 with minimum performance penalty, I did some refactoring to how events and their ids are managed (I also took the opportunity to make some changes I wanted to).

I'm still unsure about the last commit https://github.com/aquasecurity/tracee/pull/47/commits/abf4e6e56eab40a80a6367677016c557b7222d80 which changes tracee to use event ids instead of strings internally - Interested in your opinion. If we don't like it we can revert it and fall back to https://github.com/aquasecurity/tracee/pull/47/commits/8383876cb9d3478d2c5be85a68a9a3997c89a87f which include all the other required changes.

main.go
```diff
@@ -15,7 +15,7 @@ func main() {
 		Usage: "Trace OS events and syscalls using eBPF",
 		Action: func(c *cli.Context) error {
 			if c.Bool("list") {
-				printList()
+				fmt.Println(tracee.EventsIDToName)
```
yanivagman: I think we should keep the separation between syscalls and other events when printing

tracee/tracee.go
```diff
@@ -254,13 +213,60 @@ func (t *Tracee) initBPF() error {
...
+	for _, essentialEvent := range essentialEvents {
+		if eName == essentialEvent {
+			isEssential = true
+			break
+		}
+	}
+	if !isEssential {
+		return true
+	}
+
```
yanivagman: If we reached this point, it means the event is an essential event.
As the number of the essential events is only 3, it is better to save a boolean for each of these ones once during init.
This will save the need to iterate through EventsToTrace which can be a big list

tracee/tracee.go
```diff
@@ -487,3 +495,16 @@ func readArgFromBuff(dataBuff io.Reader) (interface{}, error) {
 	}
 	return res, nil
 }
+
+// DedupStringSlice removes any duplicate items in the incoming slice array as a new slice array
+func DedupStringSlice(in []string) []string {
+	tmpset := make(map[string]bool, len(in))
+	for _, e := range in {
+		tmpset[e] = true
```
yanivagman: This logic can be done in one iteration instead of 2:
if _, ok := tmpset[e]; !ok {
    out = append(out, e)
}

tracee/consts.go
```diff
@@ -393,10 +393,10 @@ func init() {
 }
 
 // essentialEvents is a list of event ids (in EventsIDToName map) that are essential to the operation of tracee and therefore must be traced
-var essentialEvents = []string{
-	"do_exit",
-	"execve",
-	"execveat",
+var essentialEvents = []int32{
```
yanivagman: I don't really understand the reason for moving from string to int32 - can you please explain?

itaysk: We just need to pick one type as the native type and I wasn't sure which one so I tried both. the fact that I tried first string and then int was just a coincidence. using string will not save us conversions, they will just move to other places.

tracee/tracee.go
itaysk: 👍

tracee/tracee.go
itaysk: I have changed how shouldPrintEvent works altogether in https://github.com/aquasecurity/tracee/pull/47/commits/95dbdbe8586281a6ac4aaea089b4a15303b4f8b5. What do you think?

tracee/tracee.go
```diff
@@ -164,11 +164,24 @@ func (t *Tracee) initBPF() error {
...
+	}
+	for _, e := range t.config.EventsToTrace {
+		_, essential := tmpset[e]
+		if essential {
+			essentialEvents[e] = true // mark that this essential event was requested by the user
+		} else {
+			eventsToTraceFinal = append(eventsToTraceFinal, e)
+		}
+		tmpset[e] = true
```
yanivagman: In the previous loop, tmpset contained only essential events.
This is being used in "if essential" above.
Here you add non essential events to tmpset.
So the assumption that tmpset contains only essential events is invalidated after one iteration.

tracee/tracee.go
yanivagman: Looks good, but I added a comment there

tracee/tracee.go
```diff
@@ -0,0 +1,500 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("error reading capability arg: %v", err)
+		}
+		res = PrintCapability(cap)
+	case SYSCALL_T:
+		sc, err := readInt32FromBuff(dataBuff)
+		if err != nil {
+			return nil, fmt.Errorf("error reading syscall arg: %v", err)
+		}
+		if e, ok := EventsIDToName[sc]; ok {
```
yanivagman: This will not work with your recent changes!
We need to have a full map of syscall ids to syscall names

itaysk: Why do you think this will not work?

tracee/tracee.go
itaysk: there is not assumption that it contains only essential events, but I agree some of the var names like `essential` is misleading. There is another bug here where in the event that `eventsToTrace` contains duplicates, `essentialEvents` map will contain non-essential events. I was working under the assumption that the input is deduped, but I've addressed this concern in: 
https://github.com/aquasecurity/tracee/pull/47/commits/37b85b67c9d97608e2acfc1356767b10453011b8

main.go
itaysk: fixed in 7b9a03445e99c03999c3fa4bc922063eeba87111

tracee/tracee.go
itaysk: got it, fixed in 33e946e72f0ba58500b440622c1a7653afe6ade4

PR #58, opened 2020-03-19
tests for readArgFromBuff function

itaysk:
First step in starting to test Tracee. this adds tests for the various argtypes parsers (based on what Simar started). the tests are very simple, and doesn't cover all cases (actually just one case for each argtype), but it's a good start and it already uncovered some bugs, which I fixed as part of this PR as well.


tracee/argprinters.go
```diff
@@ -219,7 +220,7 @@ func PrintSocketType(st uint32) string {
 		10: "SOCK_PACKET",
 	}
 	var f []string
-	if stName, ok := socketTypes[st]; ok {
+	if stName, ok := socketTypes[st&0xf]; ok {
```
simar7: Why is this needed?

tracee/tracee_test.go
```diff
@@ -0,0 +1,201 @@
...
+				0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, //18446744073709551615
+			},
+			expectedArg: uint64(18446744073709551615),
+		},
+		{
+			name: "OFF_T_T",
+			input: []byte{5, //OFF_T_T
+				0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, //18446744073709551615
+			},
+			expectedArg: uint64(18446744073709551615),
```
simar7: The first byte determines the type, correct? So if that's the case why is there only a single case for them all here: https://github.com/aquasecurity/tracee/blob/fc957f7d2b03157ce57ee473be2ef4122723eef4/tracee/tracee.go#L543

tracee/tracee_test.go
```diff
@@ -0,0 +1,201 @@
...
+		{
+			name: "DEV_T_T",
+			input: []byte{7, //DEV_T_T
+				0xFF, 0xFF, 0xFF, 0xFF, //4294967295
+			},
+			expectedArg: uint32(4294967295),
+		},
+		{
+			name: "OFF_T_T",
+			input: []byte{8, //SIZE_T_T
```
simar7: Is this supposed to be `SIZE_T_T`?

simar7:
Could you help me run these tests locally? Currently I get this for the version that's included in `go.mod`

```
vagrant@vagrant:~/repos/tracee$ make test
make: Circular tracee/tracee_test.go <- tracee/tracee_test.go dependency dropped.
go test -v ./...
go: downloading github.com/stretchr/testify v1.5.1
go: extracting github.com/stretchr/testify v1.5.1
go: downloading github.com/iovisor/gobpf v0.0.0-20191219090757-e72091e3c5e6
go: extracting github.com/iovisor/gobpf v0.0.0-20191219090757-e72091e3c5e6
go: finding github.com/iovisor/gobpf v0.0.0-20191219090757-e72091e3c5e6
go: finding github.com/stretchr/testify v1.5.1
# github.com/iovisor/gobpf/bcc
../../go/pkg/mod/github.com/iovisor/gobpf@v0.0.0-20191219090757-e72091e3c5e6/bcc/module.go:98:40: too many arguments in call to _Cfunc_bpf_module_create_c_from_string
        have (*_Ctype_char, number, **_Ctype_char, _Ctype_int, _Ctype__Bool, nil)
        want (*_Ctype_char, _Ctype_uint, **_Ctype_char, _Ctype_int, _Ctype__Bool)
FAIL    github.com/aquasecurity/tracee/tracee [build failed]
Makefile:12: recipe for target 'test' failed
make: *** [test] Error 2
```

tracee/tracee_test.go
```diff
@@ -0,0 +1,201 @@
...
+			},
+			expectedArg: "PR_GET_TSC",
+		},
+		{
+			name: "R_PATH_T",
+			input: []byte{23, //R_PATH_T
+				16, 0, 0, 0, //len=16
+				47, 117, 115, 114, 47, 98, 105, 110, 47, 100, 111, 99, 107, 101, 114, 0, // /usr/bin/docker
+			},
+			expectedArg: "docker/bin/usr/",
```
simar7: For my own understanding: why is this backwards?

tracee/argprinters.go
itaysk: seems that only the last byte is used to encode the type: https://elixir.bootlin.com/linux/v5.5.3/source/arch/mips/include/asm/socket.h#L42
(also, look at tracee python)

tracee/tracee_test.go
itaysk: indeed. fixed

tracee/tracee_test.go
itaysk: with https://github.com/aquasecurity/tracee/pull/46/ @yanivagman added support to get the real executable path by walking up the directory hierarchy in the BPF code. the nature of the algorithm means the path os constructed in reverse. In order to avoid complex code in BPF, we send the reversed path as-is to the frontend and reverse it in python/go.

itaysk:
> Could you help me run these tests locally? Currently I get this for the version that's included in go.mod

@simar7 it looks like you have an older version of bcc (not gobpf). 
can you try to use the latest one? this might not be an easy task given the poor state of bcc delivery pipeline - I ended up building from source.

simar7:
> > Could you help me run these tests locally? Currently I get this for the version that's included in go.mod
> 
> @simar7 it looks like you have an older version of bcc (not gobpf).
> can you try to use the latest one? this might not be an easy task given the poor state of bcc delivery pipeline - I ended up building from source.

Hmm.. I tried but this is what I get:
```
vagrant@vagrant:~/repos/tracee$ make test
make: Circular tracee/tracee_test.go <- tracee/tracee_test.go dependency dropped.
go test -v ./...
# github.com/iovisor/gobpf/bcc
../../go/pkg/mod/github.com/iovisor/gobpf@v0.0.0-20191219090757-e72091e3c5e6/bcc/module.go:32:10: fatal error: bcc/bcc_common.h: No such file or directory
 #include <bcc/bcc_common.h>
          ^~~~~~~~~~~~~~~~~~
compilation terminated.
FAIL    github.com/aquasecurity/tracee/tracee [build failed]
Makefile:12: recipe for target 'test' failed
make: *** [test] Error 2
```

simar7:
Could the environment to run these tests be replicated in a CI? That way we don't have to run them locally on our laptops to see the output. I basically wanted to see if this test case that you added *fails* when you run https://github.com/aquasecurity/tracee/pull/58/files#diff-c53c88ae223915b8a9a7b2f946cbf3abR74-R79

If it's expected to fail, maybe we re-write it in another way where we expect failure and not fail the overall test results. Maybe the comment is about something completely different, I'm not sure since I can't run the tests to see what happens :[

itaysk:
@simar7 I think your bcc is not installed correctly. can you run other bcc programs? 
yeah there's no CI for Tracee at the moment, I think this should be another task though.
can you point to the specific test you refer to? the link just shows me the PR diff.

PR #61, opened 2020-03-29
print lost events at the end

itaysk:
printing lost event stats within the stream of events has a couple of drawback:
1. it's hard to discover how many lost event were there. the printing is unpredictable
2. consumers will have to ignore the lost events messages (currently this is not bothering ultrabox which is not using the stdout anyway)

### DIFF
--- a/tracee/tracer.py
+++ b/tracee/tracer.py
@@ -1194,7 +1194,6 @@ class EventMonitor:
 
     def lost_event(self, lost):
         self.total_lost += lost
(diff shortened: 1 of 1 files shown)


yanivagman:
The thing is that I would like to keep the lost events message within the context of the events, so we can know which part is missing (by knowing the events that happened before and after the lost events). Printing to stderr is a possibility. We can also add a timestamp to the lost events message so we can know the context

PR #62, opened 2020-04-06
fix missing threads in system mode

yanivagman:


itaysk:
What is system mode?

yanivagman:
Running Tracee without "-c" (container mode)

PR #73, opened 2020-04-30
rename events-to-trace flag to event

itaysk:
and touch up the relevant readme section


itaysk:
I thought about it but didn't want to break the python version. updating it now

Issue #69, opened 2020-04-21
release tracee


itaysk:
build and release artifacts for tracee. probably with goreleaser. will need to version before.
@lizrice what do you think about the version:
1. officially "start counting" - `0.0.1`
2. don't start counting `0.0.0`
3. 1 with exnension: `0.0.1-rc1`
4. 2 with extension: `0.0.0-rc1`
Or if you have another suggestion. Note that this will be the git tag so applied to python as well

lizrice:
Yes, let's "start counting". So long as the major version is 0, I don't think we need rc extensions. We could do a 0.0.1 to check that everything is going smoothly with the versioning, and for testing on different distributions, shortly followed by 0.1.0 - first minor release. Wdyt?

itaysk:
sounds good to me. I'll release 0.0.1 once https://github.com/aquasecurity/tracee/pull/71 is merged

PR #75, opened 2020-05-03
add release procedure using goreleaser

itaysk:
This PR is introducing a GoReleaser workflow that build tracee binary and Docker image, and can push them to GitHub/DockerHub. 
related to: https://github.com/aquasecurity/tracee/issues/69

GoReleaser is a popular tool for Go projects, and we (aqua) use it in our other OSS Go projects, so we wanted to see if we can use it here as well. It works, but there are a couple of points to consider:
1. The binary build needs the `EBPFPROGRAM_BASE64` environment variable. I didn't find a way to define this in the goreleaser yaml, so I create it in the makefile, meaning GoReleaser must be called through the makefile which is awkward solution but was the compromise I found. 
2. In a previous PR I introduced a Dockerfile that builds binary and then created a final image using it (multistage dockerfile). But if we're using goreleaser, then goreleaser should build the binary (not the docker builder). In order to accommodate both approaches, I created a new dockerfile for goreleaser which is not building the binary.

Essentially this is about supporting 3 partially overlapping tools for build automation: Makefile/Docker/GoReleaser. I tried to find a middleground between them, let me know if you think it's reasonable.


Makefile
```diff
@@ -14,5 +14,14 @@ test:
 clean:
 	rm -rf dist || true
 
+.PHONY: release
+# release by default will not publish. run with `publish=1` to publish
+goreleaserFlags = --skip-publish --snapshot
+ifdef publish
+	goreleaserFlags =
+endif
```
simar7, line 22: for my own understanding: how does this work?

itaysk: I wanted to be extra careful with actually pushing things, so I created the `goreleaserFlags` make variable that defaults to `--skip-publish --snapshot` (which assures nothing scary is done), and the command that is run is appending this variable to the `goreleaser` command (line 24). but if the use defines the `publish` make variable (it can have any value, 1 is just an example) then the variable `goreleaserFlags` variable is emptied so nothing is appended to the `goreleaser` command, and the defaults are to publish.

PR #81, opened 2020-05-14
Cleanup stderr logs

simar7:
It seems that the stderr logs are currently flooded with this message when running inside a container
```
open(/sys/kernel/debug/tracing/kprobe_events): No such file or directory
```
It seems to be fixed if you mount /sys/kernel/debug within the container. 

Reference: https://github.com/iovisor/bcc/issues/1878

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ We will use the Tracee Docker image, which includes glibc and BCC. The host that
 To run Tracee using docker:
 
 ```bash
-docker run --name tracee --rm --privileged -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro aquasec/tracee:latest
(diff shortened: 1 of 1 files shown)


itaysk:
Did you override the entypoint? because debugfs is being mounted there: https://github.com/aquasecurity/tracee/blob/master/entrypoint.sh#L3

It's mounted in the entrypoint because:
1. to simplify the docker run command that users need to compose
2. maybe debugfs was not mounted on the host

simar7:
> Did you override the entypoint? because debugfs is being mounted there: https://github.com/aquasecurity/tracee/blob/master/entrypoint.sh#L3
> 
> It's mounted in the entrypoint because:
> 
> 1. to simplify the docker run command that users need to compose
> 2. maybe debugfs was not mounted on the host

Ah yes, my bad. You are right, I had an entry point setup. This is not needed. Thanks!

PR #88, opened 2020-06-16
refactor events map

itaysk:
the events map used to contain a sequential list that was split into "syscalls" and "events" by convention of the event id used. The reason for the split is that there are different implementation mechanisms. This pr makes this destination more explicit by separating into 2 maps which simplifies the usage of these maps.


tracee/event_monitor_ebpf.c
```diff
@@ -394,6 +394,22 @@ enum event_id {
...
+    RESERVED341,
+    RESERVED342,
+    RESERVED343,
+    RESERVED344,
+    RESERVED345,
+    RESERVED346,
+    RESERVED347,
+    RESERVED348,
+    RESERVED349,
+    RAW_SYSCALLS,
```
yanivagman, line 412: Maybe you want to add RAW_SYSCALLS to the next commit?

itaysk: Yes it’s slipped here. Will update

PR #91, opened 2020-06-21
Use full buffer size

yanivagman:


tracee/event_monitor_ebpf.c
```diff
@@ -742,22 +741,29 @@ static __always_inline int save_context_to_buf(submit_buf_t *submit_p, void *ptr
 
 static __always_inline int save_to_submit_buf(submit_buf_t *submit_p, void *ptr, int size, u8 type)
 {
-    if (submit_p->off > SUBMIT_BUFSIZE_HALF)
-        // not enough space - return
-        return 0;
+// The biggest element that can be saved with this function should be defined here
+#define MAX_ELEMENT_SIZE sizeof(struct sockaddr_un)
```
itaysk, line 745: why do we use this and not the `size` arg?

yanivagman: Because the BPF validator needs to know the maximal size in COMPILE time and not in run time, otherwise, it will refuse to run the program

itaysk:
very small suggestion: maybe move the max element declaration next to the max string declaration, or alternatively move the max string declaration to it's save function. but we can merge regardless

yanivagman:
I don't want to move max element constant from the function, as I want that if someone uses this function in the future with a bigger element, he will see this constant and update it accordingly (it is only used by this function).
On the other hand, the max string constant is used by at least two functions, so I can't move it to its save string functions. Also, max_string should never be changed - so it is not the same case

PR #95, opened 2020-06-23
Add vfs write event

yanivagman:


itaysk:
Let's merge this now and I'm also recording here some of the future improvements we discussed:

in bpf profram:
- extract "should trace" logic to a function
- file extraction filter check - use a loop
- rename config helper functions or remove them

in userspace:
- read from buffer into struct
- refactor processFileWrites. separate event parsing from file writing
- maybe reconsider the event table design, with the new kretprobe map

user experience:
- rename variables: `--capture-files /path/prefix --capture-files /another`
- output-path - make it clear this is not related to captured files

PR #105, opened 2020-06-24
Fix go timestamp resolution

### LINKED ISSUE #92: fix timestamp resolution in go
should be the same as python

yanivagman:
fixes #92


itaysk:
@yanivagman I think this is a cleaner approach: https://github.com/aquasecurity/tracee/pull/108

yanivagman:
Ok. Let's go with #108 approach

PR #107, opened 2020-06-25
remove args brackets

yanivagman:


itaysk:
Why is it desired to remove bracket? Hey make the output more readable and help frame inline arrays

yanivagman:
In the python version, no brackets exists.
I find it more readable without the brackets - the args starts in a known position in the output, and ends by the end of the line - why do we need those brackets (I find them annoying)

itaysk:
I didn't add them on purpose it was a side affect of the switch to go, but I this it's better this way. since the args can contain an array, the brackets help you distinguish between an element of the args or an element of an array arg. for example: `[foo [bar, baz]]` would be `foo bar baz`

yanivagman:
This is not true.
It will show as
foo [bar, baz]
Which is clear by all means.
Just checked.

itaysk:
ok so this just removes the outer brackets?

PR #108, opened 2020-06-25
change printed time resolution from seconds to microseconds

### LINKED ISSUE #92: fix timestamp resolution in go

itaysk:
fix #92


yanivagman:
But here you have 2 timestamps: one in the printableEvent struct, and the other in context.
They both have `json:"time" - isn't it a problem?

yanivagman:
OK, I think I understand that - the top struct tag gets precedence - cool.
Let's go with that approach

PR #112, opened 2020-06-28
cleanup file event handling code

itaysk:


tracee/tracee.go
```diff
@@ -390,35 +383,43 @@ func (t *Tracee) processEvents() {
...
-			if dataBuff.Len() != int(meta.Count) {
+			if dataBuff.Len() <= int(meta.Size) {
 				t.errorCounter = t.errorCounter + 1
 				continue
 			}
 
-			pathname := t.config.OutputPath + "/" + strconv.Itoa(int(meta.MntId))
-			if err := os.MkdirAll(pathname, 0755); err != nil {
+			pathname := path.Join(t.config.OutputPath, strconv.Itoa(int(meta.MntID)))
+			if err := os.MkdirAll(pathname, 0750); err != nil {
```
yanivagman, line 415: Why changing the permissions for "other"? Remeber that tracee runs as root, and you do want to give r+x permissions to "other" as well

tracee/tracee.go
```diff
@@ -390,35 +383,43 @@ func (t *Tracee) processEvents() {
...
 				t.errorCounter = t.errorCounter + 1
 				continue
 			}
+			filename := fmt.Sprintf("write.dev-%d.inode-%d", meta.DevID, meta.Inode)
+			fullname := path.Join(pathname, filename)
 
-			filename := pathname + "/write.dev-" + strconv.Itoa(int(meta.DevId)) + ".inode-" + strconv.Itoa(int(meta.Inode))
-
-			f, err := os.OpenFile(filename, os.O_CREATE|os.O_WRONLY, 0644)
+			f, err := os.OpenFile(fullname, os.O_CREATE|os.O_WRONLY, 0640)
```
yanivagman, line 422: Same

tracee/tracee.go
itaysk: tracee runs as whatever the user runs it with, doesn't have to be root right? I'm afraid of this becoming a vulnerability, I didn't think out the entire attack vector but I thought it's better to restrict this directory. WDYT?

itaysk: perhaps something inspired by this one: https://labs.redyops.com/index.php/2020/04/27/symantec-endpoint-protection-sep-14-2-eop-via-arbitrary-write/?fbclid=IwAR1-9Pby5y639KSPdS6kZTQI8E6matgVslf56lVpq3Skh9v9-V81JmbXulc

itaysk: again, I'm not trying to fix a specific vuln, I just tought it would be a better practice to use minimal permissions. If you think otherwise I can change it back

yanivagman: Well, I see your concern, but:
1. Tracee already runs as privilged user - and whoever runs it has the capabilities to do many more malicious things if he'd like that (SYS_ADMIN is probably the most capable capability, and equivalent root)
2. We don't give execute files to files we write to the output path. So, event if we write some malicious file to the output, an unprivilged user isn't capable of executing it.
3. Giving execute permission to a directory is equivalent to be able to iterate over it - I don't see a reason to remove this permission.
4. I think that users should be able to read the output, event without having root permissions - for usability reasons

itaysk: what about data exfiltration concerns? any user in the system will be able to read these files and their contents.

itaysk: btw, we're having this discussion on the dir change, but it applies to the files permissions as well

yanivagman: Yes, data exfiltration is indeed an issue. Dir permissions in that case don't have to be the same like file permissions. We can let users list the files, but not read them

itaysk: ok I've changed just the files permissions now

PR #119, opened 2020-07-02
fix map update issue with old kernels

yanivagman:


itaysk:
can you explain what this does please?

yanivagman:
We found that with recent changes we added today to the send_bin function, Tracee doesn't run on kernel 4.15 (On my 4.19 there is no problem). This fixes the problem for 4.15

tracee/event_monitor_ebpf.c
```diff
@@ -1724,8 +1724,13 @@ int send_bin(struct pt_regs *ctx)
     chunk_size = bin_args->full_size - i*F_CHUNK_SIZE;
 
     if (chunk_size > F_CHUNK_SIZE) {
-        bin_args->full_size = chunk_size;
-        bin_args_map.update(&id, bin_args);
+        bin_args_t bin_args_new = {};
```
itaysk, line 1727:
```suggestion
        // create a copy of the map before submitting - workaround for an issue with old kernels
        bin_args_t bin_args_new = {};
```

yanivagman: It actually the same as it was before a previous commit - nothing odd here

PR #122, opened 2020-07-05
process event hook, raw_syscall show name, capture executed files

itaysk:
This PR has 3 commits that build on one another:
1. add a hook in the processevents loop that allows to further transform and enrich events. this is an initial implementation, it will be improved in a future PR. This also implements a minor change that shows the syscall names for raw_syscalls instead of the ID.
2. put vfs captured files in a sub directory, as preparation for capture executed files
3. use the hook in (1) to capture executed files (execve).

### DIFF
--- a/tracee/tracee.go
+++ b/tracee/tracee.go
@@ -341,6 +341,51 @@ func (t *Tracee) processEvent(ctx *context, args []interface{}) error {
 		}
 	}
 
+	//capture executed files
+	if t.config.CaptureFiles && (eventName == "execve") {
+		var err error
+
+		destinationDirPath := filepath.Join(t.config.OutputPath, strconv.Itoa(int(ctx.Mnt_id)), "executed-files")
+		err = ensureOutputDir(destinationDirPath)
+		if err != nil {
+			return err
+		}
+		sourceFilePath, ok := args[0].(string)
+		if !ok {
+			return fmt.Errorf("error parsing execve args")
+		}
+		destinationFilePath := filepath.Join(destinationDirPath, fmt.Sprintf("%d-%s", ctx.Ts, filepath.Base(sourceFilePath)))
+		err = copyFileByPath(sourceFilePath, destinationFilePath)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func copyFileByPath(src, dst string) error {
+	sourceFileStat, err := os.Stat(src)
+	if err != nil {
+		return err
+	}
+	if !sourceFileStat.Mode().IsRegular() {
+		return fmt.Errorf("%s is not a regular file", src)
+	}
+	source, err := os.Open(src)
+	if err != nil {
+		return err
+	}
+	defer source.Close()
+	destination, err := os.Create(dst)
+	if err != nil {
+		return err
+	}
+	defer destination.Close()
+	_, err = io.Copy(destination, source)
+	if err != nil {
+		return err
+	}
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


tracee/tracee.go
```diff
@@ -407,8 +441,9 @@ func (t *Tracee) processFileWrites() {
 				continue
 			}
 
-			pathname := path.Join(t.config.OutputPath, strconv.Itoa(int(meta.MntID)))
-			if err := os.MkdirAll(pathname, 0755); err != nil {
+			pathname := path.Join(t.config.OutputPath, strconv.Itoa(int(meta.MntID)), "written-files")
+			ensureOutputDir(pathname)
```
yanivagman, line 490: Why create a special (long) function for what os.MkdirAll does for "free"?

tracee/tracee.go
```diff
@@ -407,8 +441,9 @@ func (t *Tracee) processFileWrites() {
 				continue
 			}
 
-			pathname := path.Join(t.config.OutputPath, strconv.Itoa(int(meta.MntID)))
-			if err := os.MkdirAll(pathname, 0755); err != nil {
+			pathname := path.Join(t.config.OutputPath, strconv.Itoa(int(meta.MntID)), "written-files")
```
yanivagman, line 489: I'm don't think we need this extra hierarchy now ("written-files").
File names are different for vfs_writes (starts with "write.") and unpacker (not yet in PR, starts with "bin."), and executed files can start with "exe." prefix.
Do you see a resaon for the extra path here? we can always add it in the future

tracee/tracee.go
```diff
@@ -341,6 +341,51 @@ func (t *Tracee) processEvent(ctx *context, args []interface{}) error {
 		}
 	}
 
+	//capture executed files
+	if t.config.CaptureFiles && (eventName == "execve") {
```
yanivagman, line 345: You can use security_bprm_check instead and catch execveat as well

tracee/tracee.go
```diff
@@ -341,6 +341,51 @@ func (t *Tracee) processEvent(ctx *context, args []interface{}) error {
...
+		var err error
+
+		destinationDirPath := filepath.Join(t.config.OutputPath, strconv.Itoa(int(ctx.Mnt_id)), "executed-files")
```
yanivagman, line 348: See my other comment about the path hierarchy

tracee/tracee.go
```diff
@@ -331,6 +331,89 @@ func (t Tracee) shouldPrintEvent(e int32) bool {
...
+	}
+	defer destination.Close()
+	_, err = io.Copy(destination, source)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func ensureOutputDir(path string) error {
```
yanivagman, line 392: Don't see the point in this function

tracee/tracee.go
itaysk: my original thought was to be able to reuse directories while considering their permissions. mkdirall doesn't care about the permissions. but I'm not sure this is important anymore.

tracee/tracee.go
itaysk: I don't see a reason for hierarchy but I also don't see the problem. I don't know if adding the dir name to the file name is better.. but I don't have strong feelings about this so you prefer it this way I can change

tracee/tracee.go
itaysk: will do

itaysk:
I'm closing this and opening new smaller PRs

PR #124, opened 2020-07-06
capture executed files

itaysk:


tracee/tracee.go
```diff
@@ -341,6 +341,50 @@ func (t *Tracee) processEvent(ctx *context, args []interface{}) error {
 		}
 	}
 
+	//capture executed files
+	if t.config.CaptureFiles && (eventName == "security_bprm_check") {
```
yanivagman, line 345: Do you think its a good idea to use the same capture files flag for both vfs_write and executables?

tracee/tracee.go
```diff
@@ -341,6 +341,50 @@ func (t *Tracee) processEvent(ctx *context, args []interface{}) error {
...
+
+		destinationDirPath := filepath.Join(t.config.OutputPath, strconv.Itoa(int(ctx.Mnt_id)))
+		if err := os.MkdirAll(destinationDirPath, 0755); err != nil {
+			return err
+		}
+		sourceFilePath, ok := args[0].(string)
+		if !ok {
+			return fmt.Errorf("error parsing security_bprm_check args")
+		}
+		destinationFilePath := filepath.Join(destinationDirPath, fmt.Sprintf("exec.%d.%s", ctx.Ts, filepath.Base(sourceFilePath)))
```
yanivagman, line 356: I'm not sure we need the file path in the new filename. It is possible to extract the filename by looking at the event output. What do you think?

itaysk: it's not the file path, it's just the file name

tracee/tracee.go
itaysk: I don't think it will be common to want one and not the other, and even if so, it's easy to just ignore the other files (maybe in the future we will "capture" more files). If you think otherwise let me know. I generally would prefer to simplify the flags rather then have every new feature under a new flag.

tracee/tracee.go
itaysk: to explain why I chose this name: I wanted to include the file name, for example `ls`. but that is not unique for example if someone execed ls and then renamed another file to be called "ls", so I added the timestamp. so the true identifier is the timestamp, the filename is added for convenience

PR #127, opened 2020-07-06
Add event configuration map

yanivagman:


tracee/consts.go
```diff
@@ -416,24 +422,12 @@ var EventsIDToName map[int32]string
...
-		EventsNameToID[name] = id
-	}
-	for id, name := range EventsKretprobes {
-		EventsIDToName[id] = name
-		EventsNameToID[name] = id
+	for id, event := range EventsIDToEvent {
+		EventsIDToName[id] = event.Name
+		EventsNameToID[event.Name] = event.ID
 	}
 }
```
itaysk, line 433: we can remove essentialEvents now, of if we want the index, just build it in init

tracee/consts.go
```diff
@@ -31,382 +31,388 @@ const (
...
+	AttachMechanism  ProbeType
+	EnabledByDefault bool
+	EssentialEvent   bool
+	ProbeName        string
 }
 
-// EventsKretprobes is list of supported kretprobes events, indexed by their ID
-var EventsKretprobes = map[int32]string{
-	355: "vfs_write",
+// EventsSyscalls is list of supported syscall events, indexed by their ID
```
itaysk, line 58: wrong comment

tracee/tracee.go
```diff
@@ -181,50 +181,56 @@ func (t *Tracee) initBPF(ebpfProgram string) error {
 
 	sysPrefix := bpf.GetSyscallPrefix()
 	for _, e := range eventsToTraceFinal {
-		if eName, isSyscall := EventsSyscalls[e]; isSyscall {
-			kp, err := t.bpfModule.LoadKprobe(fmt.Sprintf("syscall__%s", eName))
+		event, ok := EventsIDToEvent[e]
```
itaysk, line 184: instead of lookup each event by id, we can change eventsToTraceFinal from []int32 to []EventConfig

tracee/tracee.go
```diff
@@ -181,50 +181,56 @@ func (t *Tracee) initBPF(ebpfProgram string) error {
...
+		if !ok {
+			continue
+		}
+		if event.AttachMechanism == SYSCALL {
```
itaysk, line 188: not a big deal but better use a switch statement here

tracee/consts.go
```diff
@@ -31,382 +31,388 @@ const (
...
-	//	349: reserved
-}
+// ProbeType is an enum that describes the mechanism used to attach the event
+type ProbeType uint8
 
-// EventsTracepoints is list of supported tracepoints events, indexed by their ID
-// Currently the item name is the tracepoint category and the actuall tracepoint is always sys_enter
+// Syscall tracepoints are explained here: https://github.com/iovisor/bcc/blob/master/docs/reference_guide.md#8-system-call-tracepoints
+// Kprobes are explained here: https://github.com/iovisor/bcc/blob/master/docs/reference_guide.md#1-kprobes
 // Tracepoints are explained here: https://github.com/iovisor/bcc/blob/master/docs/reference_guide.md#3-tracepoints
```
itaysk, line 39: would be better if each comment is adjacent the const it relates to

tracee/consts.go
yanivagman: See my next PR

tracee/tracee.go
yanivagman: This is not a switch case. If AttachMechanism is KPROBE_KRETPROBE i want to of these cases to happen

tracee/tracee.go
yanivagman: Let's do in another commit

tracee/consts.go
yanivagman: Will fix

tracee/consts.go
yanivagman: All the events are now grouped togeher, this is why I grouped the comments here

itaysk: yes but you grouped the comments above all the ProbeTypes, instead of every comment above it's related ProbeType

itaysk:
I'm approving this so that we can move forward and rebase your other PR on top of this one

PR #136, opened 2020-07-08
Add security alerts

yanivagman:


main.go
```diff
@@ -31,8 +31,11 @@ func main() {
 				ShowExecEnv:           c.Bool("show-exec-env"),
 				OutputFormat:          c.String("output"),
 				PerfBufferSize:        c.Int("perf-buffer-size"),
+				BlobPerfBufferSize:    c.Int("blob-perf-buffer-size"),
 				OutputPath:            c.String("output-path"),
 				FilterFileWrite:       c.StringSlice("filter-file-write"),
+				ExtractDynamicCode:    c.Bool("extract-dynamic-code"),
```
itaysk, line 37: I thought this should be an option for `capture-files`, no? today we have `write/exec` we an add `mem` capture type.

main.go
```diff
@@ -52,6 +55,10 @@ func main() {
 			if c.Bool("show-all-syscalls") {
 				cfg.EventsToTrace = append(cfg.EventsToTrace, tracee.EventsNameToID["raw_syscalls"])
 			}
+			if c.Bool("security-alerts") {
+				cfg.EventsToTrace = append(cfg.EventsToTrace, tracee.EventsNameToID["mmap_alert"])
+				cfg.EventsToTrace = append(cfg.EventsToTrace, tracee.EventsNameToID["mprotect_alert"])
+			}
```
itaysk, line 61: I think these should be a single alert type, about memory protection. the fact that we implement it using mmap as a starting point is not important for the user.

tracee/consts.go
```diff
@@ -28,6 +28,7 @@ const (
 	ACCESS_MODE_T ArgType = 20
 	PTRACE_REQ_T  ArgType = 21
 	PRCTL_OPT_T   ArgType = 22
+	ALERT_T       ArgType = 23
```
itaysk, line 31: I don't think we should use one arg type for all security alerts. it depends on my other comment about joining both memory alerts into one, if we do that then we can call this MPROT_ALERT_T or something similar

tracee/event_monitor_ebpf.c
```diff
@@ -1145,6 +1166,12 @@ static __always_inline int trace_ret_generic(struct pt_regs *ctx, u32 id, u64 ty
     if (!should_trace())
         return -1;
 
+    // Workaround to a bug in gobpf, where a kprobe event (e.g. mmap syscall) can't be attached to two different functions.
+    // As we need to use save_args in kprobe entry, but one of the events (functions) might be chosen and the other not,
+    // We need a mechanism that can tell if an event was chosen by the user, so we can save the args without enabling both events.
+    if (!event_chosen(id))
+        return 0;
+
```
itaysk, line 1174: can you please also open an issue in tracee to explain this workaround as a reminder for us to remove this in the future? I think you will explain it better than me

tracee/event_monitor_ebpf.c
```diff
@@ -1894,3 +1921,112 @@ int do_trace_ret_vfs_write(struct pt_regs *ctx)
     return 0;
 }
 
+int trace_mmap_alert(struct pt_regs *ctx)
+{
+    context_t context = {};
+    args_t args = {};
+
+    // Don't delete args if mprotect event is selected, as it will be used in mmap kretprobe
```
itaysk, line 1929: did you mean mmap?

tracee/tracee.go
```diff
@@ -141,7 +147,14 @@ func New(cfg TraceeConfig) (*Tracee, error) {
 	if cfg.CaptureFilesWrite {
 		essentialEvents[EventsNameToID["vfs_write"]] = false
 	}
-
+	if cfg.SecurityAlerts || cfg.ExtractDynamicCode {
+		essentialEvents[EventsNameToID["mmap"]] = false
+		essentialEvents[EventsNameToID["mprotect"]] = false
+	}
+	if cfg.ExtractDynamicCode {
```
itaysk, line 154: don't we need the *_alert events if SecurityAlerts is chosen?

main.go
yanivagman: Yes, I wanted to do this, but then I thought that:
1. This is not a file, so capture-files not describes it well
2. The description of capture-files help got really messy when combining both:
"capture files that were written or that were executed. captured files will apear in the 'output-path' directory. possible values: 'write'/'exec'/'all'. use this flag multiple times to choose multiple file types"
+
"attempt to extract dynamically loaded code (e.g. using packers)"

main.go
yanivagman: The user chooses "security-alerts", and doesn't have to know about mmap-alert and mprotect-alert. So, it's more flxible, and the user can ignore these event names

tracee/consts.go
yanivagman: Why adding too many argument types? The alert_t only encodes some message in most of the cases, and I don't see a reason now to create new special alerts

tracee/event_monitor_ebpf.c
yanivagman: Sure. Will do.

tracee/event_monitor_ebpf.c
yanivagman: Right

tracee/tracee.go
yanivagman: This is done in the main.go already

main.go
yanivagman: Maybe we can call it just "capture"? WDYT?
How would you describe the comment then?

tracee/event_monitor_ebpf.c
yanivagman: Actually when thinking about it - this workaround is actually not a bad thing if it will stay.
It is a good idea to have a mechanism that sends the event only if the user requested for it.
Maybe we should just remove the "workaround" comment? WDYT?

main.go
itaysk: let's change to capture then

itaysk: > capture artifacts that were written, executed or found to be suspicious. captured files will appear in the 'output-path' directory. possible values: 'write'/'exec'/'mem'/'all'. use this flag multiple times to choose multiple file types"

PR #137, opened 2020-07-08
Fix "memory leaks" in bin_args_map and args_map

yanivagman:
Not in all cases elements were removed from these maps when no longer
necessary.
This patch fixes this issue

### DIFF
--- a/tracee/event_monitor_ebpf.c
+++ b/tracee/event_monitor_ebpf.c
@@ -1042,13 +1042,12 @@ static __always_inline int get_encoded_arg_num(u64 types)
     return argnum;
 }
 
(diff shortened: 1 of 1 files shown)


tracee/event_monitor_ebpf.c
```diff
@@ -1042,13 +1042,12 @@ static __always_inline int get_encoded_arg_num(u64 types)
     return argnum;
 }
 
-static __always_inline int save_args_to_submit_buf(u64 types, bool delete_args, u32 event_id)
+static __always_inline int save_args_to_submit_buf(u64 types, bool delete_args, args_t *args)
```
itaysk, line 1045: delete_args is unnecessary now no?

yanivagman: Correct. Will do

PR #140, opened 2020-07-08
Don't print syscall twice

### LINKED ISSUE #102: Don't print raw_syscalls for syscalls with existing events
Currently we have duplication of syscall events where raw_syscalls prints an event in addition to an existing syscall event.
Remove the raw_syscalls print in that case

yanivagman:
If "show-all-syscalls" is chosen, a syscall that tracee already traces
will be printed twice to the output.
Avoid this case by not printing raw_syscalls of known syscalls

resolves #102

### DIFF
--- a/tracee/tracee.go
+++ b/tracee/tracee.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/base64"
 	"encoding/binary"
+	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


itaysk:
if you don't mind, put this on hold until I have a better idea of how the even processing logic will be like, then we will know how best to implement this feature

Issue #141, opened 2020-07-09
Don't copy executable multiple times


yanivagman:
When capture executables is enabled, every file is copied on each exec.
This wastes memory, as the same file can be copied multiple times.
If the binary was already copied (check using access + stat ctime and cache results), don't copy it

PR #142, opened 2020-07-09
don't capture same exec twice

itaysk:
fix: https://github.com/aquasecurity/tracee/issues/141

### DIFF
--- a/tracee/tracee.go
+++ b/tracee/tracee.go
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"sync/atomic"
+	"syscall"
 
 	bpf "github.com/iovisor/gobpf/bcc"
 )
@@ -104,6 +105,7 @@ type Tracee struct {
 	lostWrChannel chan uint64
 	printer       eventPrinter
 	stats         statsStore
+	capturedFiles map[string]int64
 }
 
 type counter int32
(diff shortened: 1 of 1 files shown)


tracee/tracee.go
```diff
@@ -367,15 +370,26 @@ func (t *Tracee) processEvent(ctx *context, args []interface{}) error {
...
 		sourceFilePath, ok := args[0].(string)
 		if !ok {
 			return fmt.Errorf("error parsing security_bprm_check args")
 		}
+		sourceFile, err := os.Stat(sourceFilePath)
+		if err != nil {
+			return err
+		}
+		lastMtime, ok := t.capturedFiles[sourceFilePath]
+		if ok && lastMtime == sourceFile.ModTime() {
```
yanivagman, line 382: Don't use modification time, as this can be altered by the user.
Use ctime (change time), which is altered by the kernel, also on metadata change

tracee/tracee.go
```diff
@@ -367,15 +370,26 @@ func (t *Tracee) processEvent(ctx *context, args []interface{}) error {
...
 
-		destinationDirPath := filepath.Join(t.config.OutputPath, strconv.Itoa(int(ctx.Mnt_id)))
-		if err := os.MkdirAll(destinationDirPath, 0755); err != nil {
-			return err
-		}
 		sourceFilePath, ok := args[0].(string)
 		if !ok {
 			return fmt.Errorf("error parsing security_bprm_check args")
 		}
+		sourceFile, err := os.Stat(sourceFilePath)
```
yanivagman, line 377: I would call sourceFile -> sourceFileStat to be more clear

PR #146, opened 2020-07-12
Change consts to use go naming conventions

yanivagman:


yanivagman:
In this commit, I just changed the consts to be in go style. Let's consider other things (e.g. context structs) when we'll do a more thorugh change to go conventions in all tracee go code.

PR #149, opened 2020-07-19
Fix unix socket name when there are leading zeros

yanivagman:


tracee/tracee.go
```diff
@@ -652,8 +652,13 @@ func readStringVarFromBuff(buff io.Reader, max int) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("error reading null terminated string: %v", err)
 	}
+	count := 1
 	for char != 0 {
 		res = append(res, byte(char))
+		if count == max {
+			break
+		}
+		count ++
```
itaysk, line 661: perhaps it would be clearer to use the for statement for this:
```
for count :=1; char != 0 && count < max; count++ {
```

PR #151, opened 2020-07-20
Change socket addr format

yanivagman:


tracee/tracee.go
```diff
@@ -922,7 +923,13 @@ func readArgFromBuff(dataBuff io.Reader) (interface{}, error) {
 		if err != nil {
 			return nil, err
 		}
-		res = fmt.Sprintf("%v", sockaddr)[3:] // remove the leading "map" string
+		var s string
+		for key, val := range sockaddr {
+			s += fmt.Sprintf("'%s': '%s',", key, val)
```
itaysk, line 928: i think it's better to use `"` instead of `'` for strings (as it's json compliant and more common in different parsers)

itaysk:
woudlnt it be easier/better to just json marshal the map?

PR #153, opened 2020-07-22
append capture write if char, socket or fifo file is given

yanivagman:


tracee/tracee.go
```diff
@@ -509,17 +509,31 @@ func (t *Tracee) processFileWrites() {
...
+
+		S_IFSOCK uint16 = 0140000 // socket
+		S_IFLNK  uint16 = 0120000 // symbolic link
+		S_IFREG  uint16 = 0100000 // regular file
+		S_IFBLK  uint16 = 0060000 // block device
+		S_IFDIR  uint16 = 0040000 // directory
+		S_IFCHR  uint16 = 0020000 // character device
+		S_IFIFO  uint16 = 0010000 // FIFO
+	)
+
```
itaysk, line 530: we already have this code in `PrintInodeMode` (in argprinters.go). can you reuse it?
you can extract the file type section from `PrintInodeMode` into another function: `ParseModeFileType` (in the same file argprinters.go) and call it from both `PrintInodeMode` and from here. WDYT?

yanivagman: This not code, it's constants declaration. In PrintInodeMode we don't declare the constants, just use the numbers themselves. We can have these constants shared btween them both

itaysk: if not sharing code (meaning line 566), then better to just leave it here.. I approved

PR #156, opened 2020-07-22
Fix chmod invalid file

### LINKED ISSUE #155: chmod trace bug
command:
```
chmod +x test
```

output:
```
17431.801997   ubuntu           4026531840   4026531836   1000   fchmodat         chmod            19716  19716  10285  0           -100 test invalid file type|S_IRWXU|S_IRWXG|S_IROTH|S_IXOTH -1 
```

notice the "invalid file type" in mode

yanivagman:
Resolves #155

### DIFF
--- a/tracee/argprinters.go
+++ b/tracee/argprinters.go
@@ -28,8 +28,6 @@ func PrintInodeMode(mode uint32) string {
 		f = append(f, "S_IFCHR")
 	case mode&010000 == 010000:
 		f = append(f, "S_IFIFO")
-	default:
-		f = append(f, "invalid file type")
 	}
 
 	// File Mode


itaysk:
How is this fixing the problem? it just hides it, but the file type is still not being reported correctly

yanivagman:
There is no "file type" here, these are just flags that need to be reported correctly, as they were inserted by the user who called the system call

itaysk:
what do you mean "there is no file type here"? there is a file type - S_IFREG and it's not being printed.
here's the output from strace:
```
stat("dist/tmp", {st_mode=S_IFREG|0775, st_size=0, ...}) = 0
```

PR #158, opened 2020-07-23
Add exclude event flag

yanivagman:


main.go
```diff
@@ -80,6 +83,11 @@ func main() {
 				Value:   nil,
 				Usage:   "trace only the specified event or syscall. use this flag multiple times to choose multiple events",
 			},
+			&cli.StringSliceFlag{
+				Name:    "exclude",
```
itaysk, line 87: I think it should be called exclude-event rather than exclude

main.go
```diff
@@ -20,7 +20,10 @@ func main() {
 				printList()
 				return nil
 			}
-			events, err := prepareEventsToTrace(c.StringSlice("event"))
+			if len(c.StringSlice("event")) > 0 && len(c.StringSlice("exclude")) > 0 {
```
itaysk, line 23: probably clearer to replace `len(c.StringSlice("flag")) > 0` with `c.IsSet("flag")`

PR #163, opened 2020-07-26
Beautify table print

yanivagman:


tracee/printer.go
```diff
@@ -80,20 +87,29 @@ func newEvent(ctx context, args []interface{}) (Event, error) {
...
+	}
 	fmt.Fprintln(p.out)
 }
 
 func (p tableEventPrinter) Print(event Event) {
-	fmt.Fprintf(p.out, "%-14f %-16s %-12d %-12d %-6d %-16s %-16s %-6d %-6d %-6d %-12d", event.Timestamp, event.HostName, event.MountNS, event.PIDNS, event.UserID, event.EventName, event.ProcessName, event.ProcessID, event.ThreadID, event.ParentProcessID, event.ReturnValue)
+	if p.verbose {
+		fmt.Fprintf(p.out, "%-14f %-16s %-12d %-12d %-6d %-20s %-16s %-6d %-6d %-6d %-16d", event.Timestamp, event.HostName, event.MountNS, event.PIDNS, event.UserID, event.EventName, event.ProcessName, event.ProcessID, event.ThreadID, event.ParentProcessID, event.ReturnValue)
+	} else {
+		fmt.Fprintf(p.out, "%-14f %-16s %-6d %-20s %-16s %-6d %-6d %-6d %-16d", event.Timestamp, event.HostName, event.UserID, event.EventName, event.ProcessName, event.ProcessID, event.ThreadID, event.ParentProcessID, event.ReturnValue)
```
itaysk, line 111: what do you think of removing tid,ppid,user from the non-verbose output?

yanivagman: I think that these are more informative than the namespaces numbers.
Maybe in the future we should consider adding the ability to format the message, and let the user decide, but this is for another commit

PR #165, opened 2020-07-27
Add argument names

yanivagman:


tracee/printer.go
```diff
@@ -63,10 +63,11 @@ type Event struct {
 	EventName       string        `json:"eventName"`
 	ArgsNum         int           `json:"argsNum"`
 	ReturnValue     int           `json:"returnValue"`
+	ArgsNames       []string      `json:"argsNames"`
```
itaysk, line 66: I think we should not just add the argnames array to the Event type here, but instead combine it with the args array, probably as a `map[string]interface{}`. The reason we created this Event type is that we could decouple the internal data structures (context, []args) that are required to process events from the external data type (Event) that is used to communicate events externally (gob) and for printing (printer). This is a good opportunity to take use the Event struct to adapt the internal representation (2 arrays) which were convenient for processing but not for consuming, and adjust it to be convenient for consuming (one map).

tracee/printer.go
```diff
@@ -110,8 +112,8 @@ func (p tableEventPrinter) Print(event Event) {
 	} else {
 		fmt.Fprintf(p.out, "%-14f %-16s %-6d %-20s %-16s %-6d %-6d %-6d %-16d", event.Timestamp, event.HostName, event.UserID, event.EventName, event.ProcessName, event.ProcessID, event.ThreadID, event.ParentProcessID, event.ReturnValue)
 	}
-	for _, value := range event.Args {
-		fmt.Fprintf(p.out, "%v ", value)
+	for i, value := range event.Args {
+		fmt.Fprintf(p.out, "%s: %v ", event.ArgsNames[i], value)
```
itaysk, line 116: you have only handled the table printer and not the others, but I think the other will be naturally fixed if address my other comment fro above

tracee/tracee.go
```diff
@@ -473,22 +473,33 @@ func (t *Tracee) processEvents() {
...
 					continue
 				}
+				argsTags[i] = argTag
+				args[i] = argVal
+				argName, ok := argNames[argTag]
+				if ok {
+				       argsNames[i] = argName
+				} else {
+				       t.handleError(err)
+				       continue
```
itaysk, line 492: do we want to fail here or to just skip the tag parsing? (I'm not sure myself)

tracee/tracee.go
```diff
@@ -473,22 +473,33 @@ func (t *Tracee) processEvents() {
...
+				argTag, argVal, err := readArgFromBuff(dataBuff)
 				if err != nil {
 					t.handleError(err)
 					continue
 				}
+				argsTags[i] = argTag
+				args[i] = argVal
+				argName, ok := argNames[argTag]
+				if ok {
+				       argsNames[i] = argName
```
itaysk, line 489: The event processing code doesn't care about the arg names, it just needs their tags. the names are for printing only, so I think this code doesn't belongs here rather in the `newEvent` code

tracee/printer.go
yanivagman: Nothing to do in the others - the json printer will just marshal it as well, same for the gob which will encode as well

tracee/printer.go
yanivagman: I'm handling this in some different way in my next PR which is based on this one. We can change this internal representations after these will be merged

tracee/tracee.go
yanivagman: You are right. I'm handling exactly this in my next PR

tracee/tracee.go
yanivagman: This is definetely an error. If an argtag doesn't have an argName, this means something went wrong in the event decoding, as this shouldn't happen

PR #168, opened 2020-07-29
Decouple event extraction from parsing

yanivagman:


tracee/consts.go
```diff
@@ -214,365 +204,726 @@ type EventConfig struct {
 	EssentialEvent   bool
 }
 
+// events should match defined values in ebpf code
+const (
```
itaysk, line 208: This is a conceptual duplication of `EventsNameToID`. we can eliminate one of them

yanivagman: Right. Will do in a future commit

PR #170, opened 2020-07-30
Support tracing by pid

yanivagman:


itaysk:
using this flag produces a lot more events than without. is that intentional?

PR #176, opened 2020-08-02
Handle events in a pipeline

yanivagman:
In order to achieve better performance, and as a preparation for the
addition of plugins, handle events in a pipeline

### DIFF
--- a/tracee/pipeline.go
+++ b/tracee/pipeline.go
@@ -1,0 +1,198 @@
+package tracee
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"sync"
+)
+
+func (t *Tracee) runEventPipeline(done <-chan struct{}) error {
+	var errcList []<-chan error
+
+	// Source pipeline stage.
+	rawEventChan, errc, err := t.decodeRawEvent(done)
+	if err != nil {
+		return err
+	}
+	errcList = append(errcList, errc)
+
+	processedEventChan, errc, err := t.processRawEvent(done, rawEventChan)
+	if err != nil {
+		return err
+	}
+	errcList = append(errcList, errc)
+
+	printEventChan, errc, err := t.prepareEventForPrint(done, processedEventChan)
+	if err != nil {
+		return err
+	}
+	errcList = append(errcList, errc)
+
+	errc, err = t.printEvent(done, printEventChan)
+	if err != nil {
+		return err
+	}
+	errcList = append(errcList, errc)
+
+	// Pipeline started. Waiting for pipeline to complete
+	return t.WaitForPipeline(errcList...)
+}
+
+type RawEvent struct {
+	Ctx      context
+	RawArgs  map[argTag]interface{}
+	ArgsTags []argTag
+}
+
+func (t *Tracee) decodeRawEvent(done <-chan struct{}) (<-chan RawEvent, <-chan error, error) {
+	out := make(chan RawEvent)
+	errc := make(chan error, 1)
+	go func() {
+		defer close(out)
+		defer close(errc)
+		for dataRaw := range t.eventsChannel {
+			dataBuff := bytes.NewBuffer(dataRaw)
+			var ctx context
+			err := binary.Read(dataBuff, binary.LittleEndian, &ctx)
+			if err != nil {
+				errc <- err
+				continue
+			}
+
+			rawArgs := make(map[argTag]interface{})
+			argsTags := make([]argTag, ctx.Argnum)
+			for i := 0; i < int(ctx.Argnum); i++ {
+				tag, val, err := readArgFromBuff(dataBuff)
+				if err != nil {
+					errc <- err
+					continue
+				}
+				argsTags[i] = tag
+				rawArgs[tag] = val
+			}
+			select {
+			case out <- RawEvent{ctx, rawArgs, argsTags}:
+			case <-done:
+				return
+			}
+		}
+	}()
+	return out, errc, nil
+}
+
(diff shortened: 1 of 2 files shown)


tracee/tracee.go
```diff
@@ -354,17 +355,24 @@ func (t *Tracee) initBPF(ebpfProgram string) error {
...
 	return nil
 }
 
 // Run starts the trace. it will run until interrupted
 func (t *Tracee) Run() error {
 	sig := make(chan os.Signal, 1)
 	signal.Notify(sig, os.Interrupt)
+	done := make(chan struct{})
+	// Signal pipeline that Tracee exits by closing the done channel
+	defer close(done)
```
itaysk, line 370: the rest of the "closing code" is at the end of this loop (right after `<-sig`, better to keep all of this code together

tracee/tracee.go
```diff
@@ -354,17 +355,24 @@ func (t *Tracee) initBPF(ebpfProgram string) error {
...
 	t.printer.Preamble()
 	t.eventsPerfMap.Start()
 	t.fileWrPerfMap.Start()
-	go t.processEvents()
+	go t.processErrors()
```
itaysk, line 374: why isn't `done` passed to all the gourutines?

tracee/tracee.go
```diff
@@ -423,11 +431,6 @@ func copyFileByPath(src, dst string) error {
 	return nil
 }
 
-func (t *Tracee) handleError(err error) {
```
itaysk, line 426: error handling doesn't feel like the kind of work that will benefit from parallelism. it might be better to process errors in the same execution flow as the code that caused it. wdyt?

tracee/tracee.go
yanivagman: Done is used by all the go routines in the pipeline. It is used to avoid one stage in the pipeline to block the stages which come after it. Which go routines do you refer to?

itaysk: the one I commented about, for example `processErrors` (it's outside the pipeline)

tracee/tracee.go
yanivagman: Well, not all errors have to go to the errors channel. We can handle errors locally if required.
In the case of error handling that use handleError - we just update an error counter in Tracee "class", and print the error - I thought it would be better to just let the sending go routing keep doing it's job, while the "processErrors" go routine will take care of the print and error counting (which may be time consuming).
I don't mind leaving it the way it was if you are not convinced

tracee/tracee.go
yanivagman: Will do

tracee/tracee.go
yanivagman: Well, we can add it there too, but it's not a must (it will not block anything like in the pipeline)

tracee/pipeline.go
```diff
@@ -0,0 +1,250 @@
...
+	"fmt"
+	"io"
+	"sync"
+)
+
+func (t *Tracee) runEventPipeline(done <-chan struct{}) error {
+	var errcList []<-chan error
+
+	// Source pipeline stage.
+	dataBuffChan, errc, err := t.readEventFromChannel(done)
```
itaysk, line 15: I think this step is redundant. you already have the events to be consumed in the `eventsChannel` what does `readEventFromChannel` add on top of that?

yanivagman:
Regarding out of order events - the answer is no. The channels themselves are fifo - so the stages in the pipeline still work in a serial way

tracee/tracee.go
```diff
@@ -570,55 +581,9 @@ type context struct {
 	Retval  int64
 }
 
-func (t *Tracee) processEvents() {
+func (t *Tracee) processLostEvents() {
 	for {
 		select {
```
itaysk, line 586: what is the purpose of the select here? previously it selected between processing an event an a lostevent, now I think it's redundant.

tracee/pipeline.go
```diff
@@ -0,0 +1,231 @@
...
+				args[i] = rawEvent.RawArgs[tag]
+				argName, ok := argNames[tag]
+				if ok {
+					argsNames[i] = argName
+				} else {
+					errc <- fmt.Errorf("Invalid arg tag")
+					continue
+				}
+			}
+			t.stats.eventCounter.Increment()
```
itaysk, line 133: better to move that to printEvent (after the printing), to be sure that the counter is correct

tracee/pipeline.go
```diff
@@ -0,0 +1,231 @@
...
+			select {
+			case out2 <- n:
+			case <-done:
+				return
+			}
+		}
+	}()
+	return out1, out2, errc, nil
+}
+*/
```
itaysk, line 231: I don't think this belongs here. maybe add as a comment to the PR if you want to save it for yourself in the context of the code

yanivagman:
In the future, we may want to split some of the pipeline stages.
The below code is an example of how we can do that:

// An example for splitting the pipeline
// Based on https://medium.com/statuscode/pipeline-patterns-in-go-a37bb3a7e61d
func splitter(ctx context.Context, in <-chan int64) (
	<-chan int64, <-chan int64, <-chan error, error) {
	out1 := make(chan int64)
	out2 := make(chan int64)
	errc := make(chan error, 1)
	go func() {
		defer close(out1)
		defer close(out2)
		defer close(errc)
		for n := range in {
			// Send the data to the output channel 1 but return early
			// if the context has been cancelled.
			select {
			case out1 <- n:
			case <-done:
				return
			}
			// Send the data to the output channel 2 but return early
			// if the context has been cancelled.
			select {
			case out2 <- n:
			case <-done:
				return
			}
		}
	}()
	return out1, out2, errc, nil
}

PR #188, opened 2020-08-05
improve table output

itaysk:
add host pids to container mode, remove unneeded columns from non-verbose, improve width of pid columns

### DIFF
--- a/tracee/printer.go
+++ b/tracee/printer.go
@@ -19,20 +19,22 @@ type eventPrinter interface {
 	Error(err error)
 }
 
-func newEventPrinter(kind string, out io.Writer, err io.Writer) eventPrinter {
+func newEventPrinter(kind string, containerMode bool, out io.Writer, err io.Writer) eventPrinter {
 	var res eventPrinter
 	switch kind {
 	case "table":
 		res = &tableEventPrinter{
-			out:     out,
-			err:     err,
-			verbose: false,
+			out:           out,
+			err:           err,
+			verbose:       false,
+			containerMode: containerMode,
 		}
 	case "table-verbose":
 		res = &tableEventPrinter{
-			out:     out,
-			err:     err,
-			verbose: true,
+			out:           out,
+			err:           err,
+			verbose:       true,
+			containerMode: containerMode,
 		}
 	case "json":
 		res = &jsonEventPrinter{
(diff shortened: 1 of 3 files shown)


tracee/printer.go
```diff
@@ -105,28 +107,45 @@ func newEvent(ctx context, argsNames []string, args []interface{}) (Event, error
...
 
 func (p tableEventPrinter) Init() {}
 
 func (p tableEventPrinter) Preamble() {
 	if p.verbose {
-		fmt.Fprintf(p.out, "%-14s %-16s %-12s %-12s %-6s %-16s %-6s %-6s %-6s %-16s %-20s %s", "TIME(s)", "UTS_NAME", "MNT_NS", "PID_NS", "UID", "COMM", "PID", "TID", "PPID", "RET", "EVENT", "ARGS")
+		if p.containerMode {
+			fmt.Fprintf(p.out, "%-14s %-16s %-12s %-12s %-6s %-16s %-11s %-11s %-11s %-16s %-20s %s", "TIME(s)", "UTS_NAME", "MNT_NS", "PID_NS", "UID", "COMM", "PID/host", "TID/host", "PPID/host", "RET", "EVENT", "ARGS")
+		} else {
+			fmt.Fprintf(p.out, "%-14s %-16s %-12s %-12s %-6s %-16s %-5s %-5s %-5s %-16s %-20s %s", "TIME(s)", "UTS_NAME", "MNT_NS", "PID_NS", "UID", "COMM", "PID", "TID", "PPID", "RET", "EVENT", "ARGS")
```
yanivagman, line 124: Using 5 chars for pid is not enough. Just ran Tracee and got pid: 875107 (just an example)

tracee/printer.go
```diff
@@ -105,28 +107,45 @@ func newEvent(ctx context, argsNames []string, args []interface{}) (Event, error
...
+		}
 	} else {
-		fmt.Fprintf(p.out, "%-14s %-16s %-6s %-16s %-6s %-6s %-6s %-16s %-20s %s", "TIME(s)", "UTS_NAME", "UID", "COMM", "PID", "TID", "PPID", "RET", "EVENT", "ARGS")
+		if p.containerMode {
+			fmt.Fprintf(p.out, "%-14s %-6s %-16s %-11s %-16s %-20s %s", "TIME(s)", "UID", "COMM", "PID/host", "RET", "EVENT", "ARGS")
+		} else {
+			fmt.Fprintf(p.out, "%-14s %-6s %-16s %-5s %-16s %-20s %s", "TIME(s)", "UID", "COMM", "PID", "RET", "EVENT", "ARGS")
```
yanivagman, line 130: If removing pids, I suggest keeping tid and not pid, as it's more informative and the process name (comm) already implies that different tids belong to the same process.

tracee/printer.go
itaysk: did you customize your max-pid config? as far as I understand the common max is 32768

tracee/printer.go
itaysk: Don't want to remove pid as it's useful for further analysis. I will keep both tid and pid then

tracee/printer.go
yanivagman: no

itaysk: expanded to 7

itaysk:
@yanivagman wdyt about adding separation between args in table?
```
pathname: /etc/ld.so.cache flags: O_RDONLY|O_LARGEFILE dev: 8388610 inode: 533737 
```
we can't use | or : as it's common in our args maybe:
```
<pathname: /etc/ld.so.cache> <flags: O_RDONLY|O_LARGEFILE> <dev: 8388610> <inode: 533737> 
```

yanivagman:
Adding a comma separator is possible,
Maybe with a space before and after the comma so it will be easier to parse?

itaysk:
I think that if someone wan't to parse tracee they should use json output. not sure we should tailor the table out for parsing.

itaysk:
also "comma separated" which is common also for parsing doesn't have a space before the comma

yanivagman:
So let's add a comma then

PR #191, opened 2020-08-05
fix table example

itaysk:


itaysk:
that's weird, it's shown as 2 commits. I'll squash before merge (this time I'll remember!)

PR #198, opened 2020-08-09
Add event tags

yanivagman:


main.go
```diff
@@ -169,26 +175,32 @@ func main() {
...
-func prepareEventsToTrace(eventsToTrace []string, excludeEvents []string) ([]int32, error) {
+func prepareEventsToTrace(eventsToTrace []string, tagsToTrace []string, excludeEvents []string) ([]int32, error) {
 	var res []int32
 	eventsNameToID := make(map[string]int32, len(tracee.EventsIDToEvent))
 	for _, event := range tracee.EventsIDToEvent {
 		eventsNameToID[event.Name] = event.ID
 	}
-	if eventsToTrace == nil {
+	tagsToEvents := make(map[string][]int32)
+	for id, event := range tracee.EventsIDToEvent {
```
itaysk, line 185: you can combine this loop with the one that appears 5 lines before

main.go
```diff
@@ -169,26 +175,32 @@ func main() {
...
+			isExcluded[id] = true
 		}
 		res = make([]int32, 0, len(tracee.EventsIDToEvent))
-		for _, event := range tracee.EventsIDToEvent {
-			if event.EnabledByDefault {
-				res = append(res, event.ID)
+		defaultEvents, _ := tagsToEvents["default"]
+		for _, id := range defaultEvents {
+			if !isExcluded[id] {
+				res = append(res, id)
```
itaysk, line 203: I think that instead of special casing "default", you can just set `tagsToTrace = "default` if `eventsToTrace == nil` and `tagsToTrace == nil`, and the process the tags as as usual

main.go
```diff
@@ -169,26 +175,32 @@ func main() {
...
-			tracee.EventsIDToEvent[id] = event
+			isExcluded[id] = true
 		}
 		res = make([]int32, 0, len(tracee.EventsIDToEvent))
-		for _, event := range tracee.EventsIDToEvent {
-			if event.EnabledByDefault {
-				res = append(res, event.ID)
+		defaultEvents, _ := tagsToEvents["default"]
+		for _, id := range defaultEvents {
+			if !isExcluded[id] {
```
itaysk, line 202: event exclusion is skipped if specifying a tag. need to include this in the lines 215-223 block

main.go
```diff
@@ -87,6 +87,12 @@ func main() {
 				Value:   nil,
 				Usage:   "trace only the specified event or syscall. use this flag multiple times to choose multiple events",
 			},
+			&cli.StringSliceFlag{
+				Name:    "tag",
```
itaysk, line 91: I think tag is not the best name for this feature. maybe "collection" or "events-set" or something like that?

main.go
yanivagman: Right, thanks

main.go
yanivagman: I thought about this too, the problem his is that you loose the mutual exclusion with "excluded-events", which is indeed a special case of the default behavior

main.go
yanivagman: This is on purpose. It's like "--events" and "--exclude-events" can be used on the same time. If the user chose a tag, it's like choosing events

main.go
yanivagman: OK. I'll go with events-set

main.go
yanivagman: The problem at first place of having both included and excluded events is that we will need to handle the cases where the user may give both an event to include (directly or by tag), and give the same event to exclude. This will require to resolve the conflict, or alert the user. We can do such additional checks and allow this behavior in a different commit

main.go
itaysk: why is this a special case of the default event-set? for example `--events-set network --exclude-event socket seems` like a legit use case no?

main.go
itaysk: I don't think we need conflict resolution or anything similar, exclusion should always win, so if you just [apply the event exclusion here as well](https://github.com/aquasecurity/tracee/pull/198#discussion_r467547780) then it will just filter out unwanted eventswork as expected right?

yanivagman: This is one kind of conflict resolution - that exclude always wins. But why is it reasonable that the user will both say "I want to include open event, but also exclude open event"? This is definetly a problem.
I think we should tell the user that there is a conflict and exit - but as I said, we can do it in a different commit

main.go
yanivagman: Yes, see my other comments below, as these two are related

main.go
itaysk: I disagree but we can still merge :)

yanivagman: I already changed it per your request (before your last comment)

PR #208, opened 2020-08-20
fix capture exec when sharing pidns

itaysk:
improve capture executed files and fix a bug if the container is sharing pid namespace. instead of caching the pid 1 of every mntns (sensitive to pid namespace and lost events), we will cache the first 5 (by default) pids of every mount namespace. like before, it will always include pid1 in the container.

note that I didn't implement the pid cache cleanup since this turned out to be tricky. this is not a "leak" problem because the pid cache is limited in size. In a future PR we can implement it in order to be able to add new pids to the cache.

### DIFF
--- a/tracee/tracee.go
+++ b/tracee/tracee.go
@@ -37,6 +37,7 @@ type TraceeConfig struct {
 	SecurityAlerts        bool
 	EventsFile            *os.File
 	ErrorsFile            *os.File
+	maxPidsCache          int // maximum number of pids to cache per mnt ns (in Tracee.pidsInMntns)
 }
 
 // Validate does static validation of the configuration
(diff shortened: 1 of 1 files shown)


yanivagman:
The problem with this approach is that mnt_ns id (inode number) may get recycled if, for example, a container starts then exits, and a new container then starts. As mount namespace ids for different containers may be the same, and the fact that we don't delete entries from the cache, neither override when the cache is full - this approach (as it is now) will not work correctly.

itaysk:
the problem you mention exists in master as well, right?

yanivagman:
No. In master we overwrite pid 1 of a given mount namespace when we encounter it

itaysk:
right, ok i'll improve it

itaysk:
I don't know why this got closed, and I can't reopen it.. opening a new pr

itaysk:
ok I think I got it :)

itaysk:
there's still an edge case where this won't work: if a container with shared pidns exists, and another is started that uses the same mntns, then the pids table won't get updated

itaysk:
to fix that we need a way to know when a mnt ns is created (or destroyed), and clear the mntns cache record

yanivagman (approved):

PR #210, opened 2020-08-24
Add 32bit syscalls support

### LINKED ISSUE #139: Support 32bit executables
Currently, tracee doesn't trace system calls of 32bit binaries (that run on x64).
This may be fixed by changing syscalls attach mechanism to tracepoints instead of kprobes

yanivagman:
resolves #139


yanivagman:
closes #139

itaysk (approved):

PR #218, opened 2020-09-03
signal readiness using a file in output dir

itaysk:
another process can lookup this file in the output directory to know when tracee is ready, meaning all the initialization was complete. especially useful for tests

### DIFF
--- a/tracee/tracee.go
+++ b/tracee/tracee.go
@@ -205,6 +205,14 @@ func New(cfg TraceeConfig) (*Tracee, error) {
 			t.mntNsFirstPid[uint32(mnt_ns)] = 1
 		}
 	}
+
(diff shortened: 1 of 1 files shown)


tracee/tracee.go
```diff
@@ -205,6 +205,14 @@ func New(cfg TraceeConfig) (*Tracee, error) {
 			t.mntNsFirstPid[uint32(mnt_ns)] = 1
 		}
 	}
+
+	if err := os.MkdirAll(t.config.OutputPath, 0755); err != nil {
```
yanivagman, line 209: Can we now remove mkdir from capture exec and capture write?

itaysk: no, because it's creating the mount subdirectory there

yanivagman (approved):

PR #222, opened 2020-09-06
clear output dir

itaysk:
add flag to clear the output dir before running
(I think output dir is better name than output path)

### DIFF
--- a/main.go
+++ b/main.go
@@ -39,7 +39,7 @@ func main() {
 				PerfBufferSize:        c.Int("perf-buffer-size"),
 				PidsToTrace:           c.IntSlice("pid"),
 				BlobPerfBufferSize:    c.Int("blob-perf-buffer-size"),
-				OutputPath:            c.String("output-path"),
+				OutputDir:             c.String("output-dir"),
 				FilterFileWrite:       c.StringSlice("filter-file-write"),
 				SecurityAlerts:        c.Bool("security-alerts"),
 				EventsFile:            os.Stdout,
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -39,7 +39,7 @@ func main() {
 				PerfBufferSize:        c.Int("perf-buffer-size"),
 				PidsToTrace:           c.IntSlice("pid"),
 				BlobPerfBufferSize:    c.Int("blob-perf-buffer-size"),
-				OutputPath:            c.String("output-path"),
+				OutputDir:             c.String("output-dir"),
```
yanivagman, line 42: Why changing path to dir?
Today we save to a dir, maybe tomorrow to tar.gz, who knows?

main.go
```diff
@@ -135,14 +138,20 @@ func main() {
...
 			&cli.StringFlag{
-				Name:  "output-path",
+				Name:  "output-dir",
 				Value: "/tmp/tracee",
-				Usage: "set output path",
+				Usage: "set output dir",
+			},
+			&cli.BoolFlag{
+				Name:    "clear-output-dir",
+				Aliases: []string{"rm"},
```
yanivagman, line 147: Using rm as a shortcut may be a bit confusing, don't you think?

main.go
itaysk: ok

main.go
itaysk: I changed it to clear, is it better?

yanivagman (approved):

PR #228, opened 2020-09-13
Update minimal kernel version requirement to 4.18

yanivagman:
As we have verifier issues with kernel 4.14,
Update minimal kernel version requirement to 4.18.
This will also allow us to support BPF CO-RE, and use BPF_FUNC_get_current_cgroup_id()


yanivagman:
Dropping this for now,
will reopen as necessary in the future

PR #233, opened 2020-09-24
Don't monitor events generated by tracee

yanivagman:


itaysk:
shouldn't we filter that in ebpf?

yanivagman:
good idea

itaysk (approved):

PR #240, opened 2020-10-06
added contribution guidelines

xmonish:


itaysk:
@xmonish thanks for noticing that we were lacking contribution guidelines, but we do not accept community contributions in this area. also, if this is hacktoberfest related, then please familiarize with the rules first. you are welcome to make other contributions as long as they are not administrative (like changing license/guidelines)

PR #247, opened 2020-10-10
Added automated code coverage

### LINKED ISSUE #152: Add automated code coverage
We should use codecov on Tracee like most of our other projects

sudiptog81:
Added CircleCI configuration in `.circleci/config.yml` and updated `Makefile` to include code coverage arguments.

Closes #152.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,85 @@
+version: 2.1
+orbs:
+  codecov: codecov/codecov@1.1.0
+
+defaults: &defaults
+  docker:
+    - image: cimg/go:1.15
+      auth:
+        username: $DOCKER_USER
+        password: $DOCKER_PASS
+  environment:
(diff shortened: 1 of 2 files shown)


itaysk:
Hi @sudiptog81 , thanks for the effort, I wish you had discussed it with us in the associated issue before, since this PR touches on a few related but separate issues:
1. Adding CI, which Tracee currently lacks. We are not using circle CI and not planning to. If we had a CI it would be Github Actions. I will add new issue for adding CI (without code coverage), and you can make your contribution there.
2. Adding code coverage, which we have the associated issue for, but since Tracee doesn't have any tests yet, it would be meaningless so I prefer to delay that until we added tests.
3. Adding tests, which is significant task for Tracee because of the complexities of testing OS level interactions. I will open a new issue to track it, but I do not recommend starting to work on it now since the codebase is still unstable and rapidly changing.

PR #249, opened 2020-10-11
Add support for tracing specific containers

### LINKED ISSUE #221: Additional tracing modes
We currently support three modes of operation. It may be useful to add three more.
As we don't want to have a flag for each of these modes, we better change the UX to use "--trace-target" as suggested below by @itaysk 

trace target | trace mode | status (issue)
--- | --- | ---
process | all | TODO
process | new | implemented, migrate UX
process | specific | implemented, migrate UX
container | all | TODO
container | new | implemented, migrate UX
container | specific | TODO https://github.com/aquasecurity/tracee/issues/255

siddharths2710:
With the intent to resolve #221, this PR provides support for tracing a collection of specific Container IDs by 
- Fetching the Container PID mapped to the host
-  Tracing the container in PID mode using the PID obtained from above

Signed-off-by: Siddharth Srinivasan <siddharths2710@yahoo.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/aquasecurity/tracee
 go 1.13
 
 require (
+	docker.io/go-docker v1.0.0
 	github.com/iovisor/gobpf v0.0.0-20200529092446-49b58e11a4b5
 	github.com/stretchr/testify v1.5.1
 	github.com/syndtr/gocapability v0.0.0-20180916011248-d98352740cb2
--- a/go.sum
+++ b/go.sum
@@ -1,3 +1,5 @@
+docker.io/go-docker v1.0.0 h1:VdXS/aNYQxyA9wdLD5z8Q8Ro688/hG8HzKxYVEVbE6s=
+docker.io/go-docker v1.0.0/go.mod h1:7tiAn5a0LFmjbPDbyTPOaTTOuG1ZRNXdPA6RvKY+fpY=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
 github.com/cpuguy83/go-md2man/v2 v2.0.0-20190314233015-f79a8a8ca69d h1:U+s90UTSYgptZMwQh2aRr3LuazLJIa+Pg3Kc1ylSYVY=
 github.com/cpuguy83/go-md2man/v2 v2.0.0-20190314233015-f79a8a8ca69d/go.mod h1:maD7wRr/U5Z6m/iR4s+kqSMx2CaBsrgA7czyZG/E6dU=
--- a/main.go
+++ b/main.go
@@ -38,6 +38,7 @@ func main() {
 				OutputFormat:          c.String("output"),
 				PerfBufferSize:        c.Int("perf-buffer-size"),
 				PidsToTrace:           c.IntSlice("pid"),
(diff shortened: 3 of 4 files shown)


yanivagman:
Hi @siddharths2710 ,

Thanks for this PR, it's nice to also see PRs which are not typo related ;-)

While the idea of tracing containers using docker API is nice,
we need an approach which is container runtime agnostic.
As you probably know, docker today is not the only container engine, and others exist (e.g. cri-o, containerd).
This can be tackled by, for example, using CRI instead of docker API.
I believe that it should also be possible to get the relevant pids info from the file system (using procfs and cgroupfs) -
but this approach may be a bit harder to implement.

Another problem exist when only adding the pid of the container entrypoint (which is what you get when inspecting  a container).
If the container is already running, it may have child processes which should also be added.
This can be done, again, by one of two ways:
1. recursively adding all the childrens pids found in procfs
2. instead of adding pids - by adding the mount namespace inode nr (given by /proc/'pid'/ns/mnt link), thus tracing all it's pids at once

Yaniv

yanivagman (changes requested):
See my comments above

siddharths2710:
Hi @yanivagman,

Sorry for not considering other container engines as well. My bad :(
If it were possible to be able to retrieve the pids/mount ns of the containers through the procfs/cgroupfs route, 
then wouldn't this be the better approach since it'd handle any container runtime in a generic manner?
Although I believe there is also a specific case if cgroup driver is systemd as well.

In any case, I can try out with the other option which you've provided. I've seen the go clients for cri-o and containerd to be straightforward, so maybe depending on
which runtime is being used, I can invoke the routines specific to a given runtime.

yanivagman:
Hi @siddharths2710,

No worries! This is not a simple issue, and just by exploring the options we can gain new insights.
Although cgroup/procfs may be more generic, I'm afraid it is also less stable, and may change between container runtimes and versions. Moreover, there's some differences between cgroups v1 and v2, which may cause changes in the future. If you want to explore this path, here is a good reference that handles the case of extracting a container id of a pod from cgroupfs: https://github.com/kinvolk/traceloop/blob/e527bffa5c1d370d40a6284bbfc72fc94e557f62/pkg/podcgroup/podcgroup.go

I think that in this case, the easier approach will be of using CRI, which is implemented by all container runtimes working with kubernetes (including docker, containerd and cri-o). A tool that uses this interface is crictl: https://github.com/kubernetes-sigs/cri-tools/blob/master/docs/crictl.md

Anyways, I think we should start with a much simpler case before diving into any of these approaches. Assuming the user can provide the pid of the container entrypoint (by inspecting the container to get this pid) - Tracee can go to procfs to get its pid ns inode number, and use it to trace all of the container pids (by adding the pid ns id to pids_map). This should be a good starting point, from where we will be able to automate more things in the future, like getting the container entrypoint pid from a container id for a given container runtime.

What do you think?

siddharths2710:
Yeah @yanivagman thank you for guiding me in the approach! To be frank I have zero technical background on cgroups, so like you suggested I guess I'd first start off then with the simpler case of taking the pid as input, and once I have a working check-in I can switch over to work on with using CRI afterwards :D

itaysk:
@siddharths2710 thanks again for the effort! I'm going to close this PR for now. When you are ready with the new one please refer it to this issue: https://github.com/aquasecurity/tracee/issues/255

siddharths2710:
@itaysk sure, no problem!

Issue #216, opened 2020-09-01
CWD is sometimes changed to kernel headers directory


itaysk:
(reported by Idan)
In a go program that loads tracee as a library and writes to a file simultaneously, meaning there are 2 goroutins, one that runs tracee, another that writes files to a relative path. Some files are being written to the kernel headers directory, instead of the directory where this program runs. The hunch is it's related to ebpf compilation, probably something there is `cd`ing around, and the process is affected.

yanivagman:
Which files exactly are written to the kernel headers directory? And where do we use relative paths?

itaysk:
another program that uses tracee as a library, not tracee itself. but the fact that tracee was loaded seems to have affected the host program.

yanivagman:
Ok, now it's clearer.
So why "some files"? we're not changing dir while running tracee, and compilation only happens on start - so it should have been permanent (always writing to kernel headers dir) according to the ebpf compilation theory.

If indeed BCC does a directory change on compilation, then the solution is to chdir back after bpf initialization is completed.

itaysk:
Maybe they cd into headers and then back, and code that happens to run
while ebpf compiles is written to the wrong place. In that case the
workaround is to just wait until initbpf is finished.

yanivagman:
Why workaround? As I said, the solution is to chdir back after init bpf is finished -
We can add one line to save current working dir in init_bpf function start and add one other line to change dir back in init_bpf function end

itaysk:
Cd *after* initbpf doesn’t fix because while initbpf is running the problem
remains. If you wait till after initbpf to cd then The cd is not needed.
 Of course we first need to repro and validate the hypothesis for the root
cause

yanivagman:
If the problem *only* exists during the bpf init phase, this is not an issue, and can be marked as a limitation, that the calling process should not perform any file writes during bpf init.

I thought that the problem persists after bpf init is finished (and the current dir is kept set to the kernel headers dir) - and in that case, the above solution works

yanivagman:
This should not happen now, as BPF code compilation is now done by tracee or in the makefile, and not by BCC.

Issue #209, opened 2020-08-24
Convert BCC to libbpf+CO-RE


yanivagman:
All the reasons why and a practical guide for how to do that: https://facebookmicrosites.github.io/bpf/blog/2020/02/20/bcc-to-libbpf-howto-guide.html

Work in progress in the branch `libbpf` https://github.com/aquasecurity/tracee/tree/libbpf

pathtofile:
Donig this would make tracee much more portable, very much in favour of this

yanivagman:
First part of this task is now done.
We opened a branch named libbpf, where we moved from bcc to libbpf.
You are more than encouraged to use this branch and report any problems you encounter.

pathtofile:
Hey @yanivagman, super keen to check this out, I'm having dependency issues when attempting to run `make build-docker`, as the builder docker container is missing clang and a few other deps.

But the regular `make build` works on Ubuntu 20.04, kernel 5.4 and clang 11 👍

pathtofile:
How long are you thinking of keeping this branch separate before merging it into main? Should we be targeting new PRs to this branch or main?

yanivagman:
Hi @pathtofile , there are few small todos left in this branch, which you can see in tracee.go file, initBPF function.
One of these things is adding support to build docker image.
After we will complete these todos - we can merge this branch into master

pathtofile:
awesome, thanks!

yanivagman:
Libbpf branch is now merged to master.
Our next step will be adding support for CO-RE

yanivagman:
initial CO-RE support was added in #759 
Other CO-RE related tasks should be handled in a new issue

PR #250, opened 2020-10-11
Move from gobpf to bpfwrap (libbpf)

yanivagman:
This PR replaces gobpf as the bpf library used by Tracee with bpfwrap.
bpfwrap is a wrapper to libbpf, which is written in C.
The advantages of using bpfwrap over gobpf are the following:
    - no bcc dependancy - only clang is required to compile the code
    - compile once - compilation can take place during installation
      on a given node. Then, the compiled object is used in future runs.
      This have the advantage that Tracee will not have unnecessary delays
      on start.
    - libbpf is being developed as part of the linux kernel. This means it
      will always get the latest features, and that it is well maintained.
    - the move to libbpf will allow us to add support for CO-RE in the
      future. With CO-RE, bpf object is Compiled Once, and can then
      Run Everywhere without the need to compile the code again for different
      kernels

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .vscode
 .idea
+.output
 /dist
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "3rdparty/libbpf"]
+	path = 3rdparty/libbpf
+	url = https://github.com/libbpf/libbpf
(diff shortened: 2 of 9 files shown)


itaysk:
will presumably fix https://github.com/aquasecurity/tracee/issues/216

itaysk:
fix: https://github.com/aquasecurity/tracee/issues/209

PR #256, opened 2020-10-13
Output formatting via gotemplate

### LINKED ISSUE #214: Custom output format using gotemplate
Allow consumers of tracee to customize the output format by providing a gotemplate file via the `output` flag:
`tracee -o gotemplate=/path/to/template/file`

siddharths2710:
To resolve #214, provided the `-o go-template=<path>` option.

### DIFF
--- a/tracee/printer.go
+++ b/tracee/printer.go
@@ -4,11 +4,16 @@ import (
 	"bytes"
 	"encoding/gob"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
+	"strings"
+	"text/template"
 )
 
 type eventPrinter interface {
+	// Init serves as the initializer method for every event Printer type
+	Init() error
 	// Preamble prints something before event printing begins (one time)
 	Preamble()
 	// Epilogue prints something after event printing ends (one time)
@@ -19,35 +24,47 @@ type eventPrinter interface {
 	Error(err error)
 }
 
-func newEventPrinter(kind string, containerMode bool, out io.Writer, err io.Writer) eventPrinter {
+func newEventPrinter(kind string, containerMode bool, out io.Writer, err io.Writer) (eventPrinter, error) {
 	var res eventPrinter
-	switch kind {
-	case "table":
+	var initError error
+	switch {
+	case kind == "table":
 		res = &tableEventPrinter{
 			out:           out,
 			err:           err,
 			verbose:       false,
 			containerMode: containerMode,
 		}
-	case "table-verbose":
+	case kind == "table-verbose":
 		res = &tableEventPrinter{
 			out:           out,
 			err:           err,
 			verbose:       true,
 			containerMode: containerMode,
 		}
-	case "json":
+	case kind == "json":
 		res = &jsonEventPrinter{
 			out: out,
 			err: err,
 		}
-	case "gob":
+	case kind == "gob":
 		res = &gobEventPrinter{
 			out: gob.NewEncoder(out),
 			err: gob.NewEncoder(err),
 		}
+	case strings.HasPrefix(kind, "go-template="):
+		res = &templateEventPrinter{
+			out:           out,
+			err:           err,
+			containerMode: containerMode,
+			templatePath: strings.Split(kind, "=")[1],
+		}
 	}
-	return res
+	initError = res.Init()
+	if initError != nil {
+		return nil, initError
+	}
+	return res, nil
 }
 
 // Event is a user facing data structure representing a single event
@@ -114,7 +131,8 @@ type tableEventPrinter struct {
 	containerMode bool
 }
 
-func (p tableEventPrinter) Init() {}
+
+func (p tableEventPrinter) Init() (error) { return nil }
 
 func (p tableEventPrinter) Preamble() {
 	if p.verbose {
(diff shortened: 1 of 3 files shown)


itaysk (changes requested):
Thanks for the PR, can you please align it with the UX guidelines in the associated issue? we don't want new flags. Also, there's no need for preamble and epilogue yet. the other output formats doesn't support them either. We might add preamble and epilogue support later in a consistent manner.

siddharths2710:
@itaysk I've made the change. Please review the same.

tracee/printer.go
```diff
@@ -47,6 +50,15 @@ func newEventPrinter(kind string, containerMode bool, out io.Writer, err io.Writ
 			err: gob.NewEncoder(err),
 		}
 	}
+
+	if strings.HasPrefix(kind, "go-template=") {
```
itaysk, line 54: is there a particular reason this is outside the `switch` block? if not, please make it a switch case

tracee/printer.go
```diff
@@ -167,6 +180,42 @@ func (p tableEventPrinter) Epilogue(stats statsStore) {
...
+
+func (p templateEventPrinter) Error(err error) {
+	fmt.Fprintf(p.err, "%v", err)
+}
+
+
+func (p templateEventPrinter) Print(event Event) {
+	tmplPath := p.templatePath
+	if tmplPath != "" {
+		tmpl, err := template.ParseFiles(tmplPath)
```
itaysk, line 204: It would be better to parse the template in the `Init` rather then on every event

tracee/tracee.go
```diff
@@ -30,6 +30,7 @@ type TraceeConfig struct {
 	PerfBufferSize        int
 	BlobPerfBufferSize    int
 	OutputPath            string
+	GoTemplatePath        map[string]string
```
itaysk, line 33: where is this used?

tracee/printer.go
```diff
@@ -167,6 +180,42 @@ func (p tableEventPrinter) Epilogue(stats statsStore) {
...
+		if err != nil {
+			p.Error(err)
+		}
+		err = tmpl.Execute(p.out, event)
+		if err != nil {
+			p.Error(err)
+		}
+	} else {
+		p.Error(errors.New("Please specify a go-template for event-based output"))
```
itaysk, line 213: We'll need some sort of basic validation to occur earlier. I think that a good place would be in the Printer's `Init` phase. For that, you can modify the signature of `Init` to return an error.

itaysk (changes requested):

tracee/printer.go
siddharths2710: @itaysk yeah because I'm don't think the switch statement supports pattern matching but instead only a full equality check. 
There's also the boolean version of switch for regex matching but all the other cases will have to be rewritten as a boolean condition for equality instead :(

tracee/printer.go
siddharths2710: Okay sure. I've made the change accordingly.

tracee/tracee.go
siddharths2710: This came from the previous check-in. I've removed it now.

tracee/printer.go
siddharths2710: Yeah the Init has been modified now, could you please have a look?

tracee/printer.go
itaysk: doesn't there's a problem with including the prefix check in the switch: https://play.golang.org/p/MWWHC5oWV4A

tracee/printer.go
itaysk: What I meant here is that Init should validate the input, and fail if it's invalid. In your implementation it doesn't fail it just prints that there was an error for every event, but the error is known in init time so there no reason to continue trying.
To be more specific, you need to change the Init function (and interface definition) to return an error, and bubble that error up until it fails the program (which is a good thing if the input is invalid)

tracee/printer.go
siddharths2710: Okay I wasn't sure if you'd be okay with a boolean switch with `case kind == value` .... I'll make the change then.

tracee/printer.go
siddharths2710: Ohh okk got it...I'll make the change then.....do I have to modify the Init signatures of other templates as well?

itaysk: You would need to change them as well just for the program to compile but you don't need to implement any logic there.

tracee/printer.go
itaysk: oh I understand what you meant now by boolean switch. Yes I think it's better this way because we are expecting to add more output formats which have arguments in the future

tracee/printer.go
siddharths2710: @itaysk I've made the change.....could you please have a look?

siddharths2710:
@itaysk I've made the changes....could you please have a look?

itaysk:
@siddharths2710 can you confirm that you've tested it and it works?

siddharths2710:
@itaysk yes, I've tested all the changes using sample templates referencing the attributes of the Event struct.

tracee/printer.go
itaysk: You have changed the signature of Init, but it is never called. We discussed you will add it to the interface of `eventPrinter` so that the calling code can initialize the eventPrinter.

itaysk:
After further reviewing the code, I'm doubtful that it works on your side because the code in Init is never called (see https://github.com/aquasecurity/tracee/pull/256#discussion_r507677007).
Please - make sure you test a working and failing case (manually) before asking for review.

siddharths2710:
Hi @itaysk, sorry for the late check-in. I've invoked the `Init` as well this time, and below is are the cases I've tested:

Sample template file ( Invoked as `./tracee -o go-template=/root/events.tmpl` )
```
===========================================
TIMESTAMP: {{.Timestamp}}
PID: {{.ProcessID}}
PName: {{.ProcessName}}
Event: {{.EventName}}
Args: {{.Args}}
===========================================
```

Output:

```
Event: security_file_open
Args: [{pathname /proc/28159/uid_map} {flags O_RDONLY|O_LARGEFILE} {dev 4} {inode 5284393}]
===========================================
TIMESTAMP: 1.417543947185e+06
PID: 28159
PName: runc
Event: openat
Args: [{dirfd -100} {pathname /proc/self/uid_map} {flags O_RDONLY|O_CLOEXEC} {mode 0}]
===========================================
TIMESTAMP: 1.417543947227e+06
PID: 28159
PName: runc
Event: close
Args: [{fd 5}]
===========================================
TIMESTAMP: 1.417543948179e+06
PID: 28159
PName: runc
Event: security_file_open
Args: [{pathname /run/docker/runtime-runc/moby/935f212285047d77262b493241879bb1a14a1785222b2b1e1e477f4ef0a81fa7/state.json} {flags O_RDONLY|O_LARGEFILE} {dev 23} {inode 694}]
===========================================
TIMESTAMP: 1.417543948199e+06
PID: 28159
PName: runc
Event: openat
Args: [{dirfd -100} {pathname /var/run/docker/runtime-runc/moby/935f212285047d77262b493241879bb1a14a1785222b2b1e1e477f4ef0a81fa7/state.json} {flags O_RDONLY|O_CLOEXEC} {mode 0}]
===========================================
TIMESTAMP: 1.417543954966e+06
PID: 28159
PName: runc
Event: security_file_open
Args: [{pathname /sys/fs/cgroup/freezer/docker/935f212285047d77262b493241879bb1a14a1785222b2b1e1e477f4ef0a81fa7/freezer.state} {flags O_RDONLY|O_LARGEFILE} {dev 36} {inode 853}]
===========================================
TIMESTAMP: 1.417543955019e+06
PID: 28159
PName: runc
Event: openat
Args: [{dirfd -100} {pathname /sys/fs/cgroup/freezer/docker/935f212285047d77262b493241879bb1a14a1785222b2b1e1e477f4ef0a81fa7/freezer.state} {flags O_RDONLY|O_CLOEXEC} {mode 0}]
===========================================
TIMESTAMP: 1.417543955044e+06
PID: 28159
PName: runc
Event: fstat
Args: [{fd 5} {statbuf 0xC00016F7C8}]
===========================================

```

Failure case: `./tracee -o go-template=/root/failure.tmpl`
Output:-

`2020/10/28 18:17:03 error creating Tracee: open /root/failure.tmpl: no such file or directory`

Please have a look.

itaysk:
LGTM! thank you @siddharths2710

itaysk (approved):

Issue #252, opened 2020-10-13
Tests are failing on master


itaysk:
`make test` fails on master

neha-viswanathan:
@itaysk It looks like a known issue: https://github.com/iovisor/gobpf/issues/214
```
# github.com/iovisor/gobpf/bcc
../go/pkg/mod/github.com/iovisor/gobpf@v0.0.0-20200529092446-49b58e11a4b5/bcc/module.go:32:10: fatal error: bcc/bcc_common.h: No such file or directory
   32 | #include <bcc/bcc_common.h>
      |          ^~~~~~~~~~~~~~~~~~
compilation terminated.
github.com/cpuguy83/go-md2man/v2/md2man
github.com/urfave/cli/v2
make: *** [Makefile:15: dist/tracee] Error 2
```
One of their suggestions is to build bcc from the source: https://github.com/iovisor/bcc/blob/master/INSTALL.md#ubuntu---source

sudiptog81:
@neha-viswanathan upon building `bcc` from source, I struck this error on my [`test`](https://github.com/sudiptog81/tracee/runs/1251932887) workflow: `tracee/tracee_test.go:189:15: assignment mismatch: 2 variables but readArgFromBuff returns 3 values`.

itaysk:
@neha-viswanathan you seem to have an issue with BCC on you machine. This is not what this issue discusses: after you solve the issue you have with bcc, you can run `make test` and see some tests failing. 

Regarding your error: There's a known issue if you use ubuntu and obtain bcc from official repo you will get an older version of bcc which is not compatible with Tracee. One solution is to build from source as you suggest.

itaysk:
@sudiptog81 yes this sounds more like what I was referring to

ronaudinho:
it looks like this change in function signature in https://github.com/aquasecurity/tracee/commit/50a7e1734b2856ebd53c330d53d8fcaaf4d6af73 is the cause, now
https://github.com/aquasecurity/tracee/blob/d0fe845c21b7d6613216ae1f4ea37d88b54bb155/tracee/tracee.go#L1148
but previously it only returns 2 values
https://github.com/aquasecurity/tracee/blob/378263e8ac10fc643dd1c2d2e28c237c17b2dfcb/tracee/tracee.go#L871

[...]

PR #259, opened 2020-10-14
feat: Add CI/CD Workflow

### LINKED ISSUE #254: CI/CD
It would be nice to add some Automated build, tests, and release. Currently most of the build/test/release logic is in the Makefile, which should be used as the basis for any work that is done on this issue.

sudiptog81:
Added CI/CD Workflow with GitHub Actions. There are certain libraries and dependencies that need to be built/installed in order to build Tracee from source easily, I have included those steps as well. I have used the existing routines in the `Makefile` for the various jobs.

* Workflows `build` and `test` shall be triggered on any push to the repository on any branch. As currently there is a single branch, this is fine as per my conscience.
* Workflow `release` shall be triggered only when a new tag is pushed to the repository. 

The [`build`](https://github.com/sudiptog81/tracee/runs/1251932885) workflow ran fine on my branch, however, the [`test`](https://github.com/sudiptog81/tracee/runs/1251932887) workflow failed, owing to #252. Please advise on that.

The workflows may be rudimentary and may need changes but can be easily extended in the future. Fixes #254.


.github/workflows/test.yaml
```diff
@@ -0,0 +1,45 @@
+name: test
```
itaysk, line 1: I think it would be better to merge this WF with the "build" WF, the build step takes alot of time and rebuilding here again for the test seems redundant to me. what I'm suggesting is to just take the `make test` step from here and move it into the `build` workflow (which would be renamed to build-test)

.github/workflows/release.yaml
```diff
@@ -0,0 +1,39 @@
+name: release
```
itaysk, line 1: This will not work right now since goreleaser will miss the credentials to push to dockerhub. If you want to merge this PR quickly, we can remove this publish WF from this PR and we merge it without it. We can add publish in a new PR

.github/workflows/build.yaml
```diff
@@ -0,0 +1,45 @@
...
+          sudo apt install -y bison build-essential cmake flex git libedit-dev \
+              libllvm8 llvm-8-dev libclang-8-dev libllvm9 llvm-9-dev libclang-9-dev \
+              python zlib1g-dev libelf-dev libtinfo5 libfl-dev
+          git clone https://github.com/iovisor/bcc.git
+          mkdir bcc/build; cd bcc/build
+          cmake ..
+          make
+          sudo make install
+
+      - name: Lint
```
itaysk, line 34: Lint should come before the Install Libraries step, no reason to wait for the prerequisites to install before linting

.github/workflows/build.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Setup Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.15
+        id: go
+
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      - name: Install Libraries
```
itaysk, line 22: We can simplify this job and eliminate this Install Libraries step if we just build using the docker containers. so, instead of using `make build` we would use `make build-docker`

itaysk (changes requested):
@sudiptog81 thanks for working on this, and sorry I was delayed with the review. Added a few suggestions for improvements

.github/workflows/build.yaml
sudiptog81: Running tests requires BCC (has to be built from source), so I believe, we have to install the dependencies.

.github/workflows/build.yaml
sudiptog81: Sure! Addressing this in the next commit.

.github/workflows/release.yaml
sudiptog81: I will remove the release workflow from this PR for now.

.github/workflows/test.yaml
sudiptog81: Sure! Addressing this in the next commit.

sudiptog81:
@itaysk I addressed most of the comments, however, I found running the tests requires BCC which has to be built from source, and therefore, I believe we should keep that step. Let me know your opinion.

.github/workflows/build.yaml
itaysk: My reservation with this approach is 
1. building bcc from source every time is inconsistent, when we release we use a specific commit of bcc that we know works for us.
2. It takes a while to install all these packages and if we use a docker image its all contained in the image
3. Potential code duplication between this script and the Dockerfile.

This is why I made the suggestion. You are correct that to test we will need bcc, but we can also test in a container. For now the tests are broken anyway, and also there's WIP to remove bcc from Tracee, so we can defer the decision for later. 
For now I think it would be best to do a build-docker in this WF and later we can improve it if needed (I also want to merge this by the end of the month to make you eligible for hacktoberfest)

sudiptog81: Got your point! Will make the required changes.

itaysk (approved):
Thank you @sudiptog81 ,
sorry we had to scope down your original PR. we can continue with the tests and with the release in separate PRs

PR #262, opened 2020-10-15
Typo Corrected in README.md to sound more meaningfull

echobash:


itaysk:
Thank you for spotting this typo. I think it's better to change it to `If you use the Tracee binary`, to work with the rest of the text. Please make the change and I'll accept it

itaysk (changes requested):
Thank you for spotting this typo. I think it's better to change it to If you use the Tracee binary, to work with the rest of the text. Please make the change and I'll accept it

echobash:
Yeah That sounds better. Thank you.

itaysk (approved):

PR #263, opened 2020-10-16
Update README.md

Abhi947223:


itaysk:
Thank you for your attempt to contribute, but we do not think this is a meaningful addition to the project and will not accept it. Also please note our requirement for all accepted PRs to relate to an associated Issue.

PR #268, opened 2020-10-21
Additional tracing modes

pathtofile:
Replaces PR https://github.com/aquasecurity/tracee/pull/265 as  this was quicker than undoing that bad rebase.

Discussion: https://github.com/aquasecurity/tracee/issues/221

### DIFF
--- a/tracee/consts.go
+++ b/tracee/consts.go
@@ -26,9 +26,11 @@ const (
 type binType uint8
 
 const (
-	modeSystem uint32 = iota
-	modePid
-	modeContainer
+	ModeProcessAll uint32 = iota + 1
+	ModeProcessNew
+	ModeProcessList
+	ModeContainerAll
+	ModeContainerNew
 )
 
 const (
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ You can get Tracee in any of the following ways:
 
 ### Permissions
 
-If run Tracee binary, you'll need to run it with root permissions in order to load the eBPF code. 
+If you use the Tracee binary, you'll need to run it with root permissions in order to load the eBPF code. 
 If you use the Docker container, you should run it with the `--privileged` flag.
 
 ### Quickstart with Docker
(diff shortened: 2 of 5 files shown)


pathtofile:
@yanivagman - Apologies, this was quicker for me than fixing the bad rebase on the other branch.

All changes to https://github.com/aquasecurity/tracee/pull/265 have been done, and it's now actually squashed into a single commit

tracee/consts.go
```diff
@@ -26,9 +26,11 @@ const (
 type binType uint8
 
 const (
-	modeSystem uint32 = iota
-	modePid
-	modeContainer
+	ModeProcessAll uint32 = iota
```
yanivagman, line 29: One thing that I noticed is that when I'm now running tracee in container mode (using your PR code, "-t c"), I'm getting events which are not related to containers.
Took me some time to figure this out, but the reason for this is that ModeProcessAll value is 0.
For this reason, when loading the bpf programs, and while we still didn't set the maps (specifically, the config map with  the mode value), the default mode chosen is ModeProcessAll, and possibly non relevant processes can be traced as well...

Lucky for us, the solution is simple - just change the enum to start from 1 (iota + 1), and respectively shift the defines in the C code by one as well

yanivagman (changes requested):
One last fix required :-)

pathtofile:
Done last fix, and still a single commit :-)

yanivagman:
Great!
Thanks @pathtofile !

yanivagman (approved):

yanivagman:
@pathtofile I just noticed that the commit you pushed has the wrong author details. This probably happened when you performed the merge of the commits, and took one commit more than needed.

As we want the credit for this commit will belong to the right owner, what we can do is that you will revert this commit, and re-commit it (do a PR with two commits, one for revert, and another one with the same code and correct author details) and I will approve and merge the changes.

pathtofile:
@yanivagman oof, really sorry, 2nd git mixup for this feature. I've pushed up a commit with an amended authorship: https://github.com/aquasecurity/tracee/pull/271

PR #269, opened 2020-10-21
Fix broken kernel 4.14 support

yanivagman:
Verifier in 4.14 has some issues which were fixed in 4.15.
Satisfy 4.14 verifier to be happy...


itaysk (approved):
should we add a comment to remove this once moving to 4.18 ?

yanivagman:
I considered adding comments, but then we'll have to add a comment in every location that does such trick to pass the verifier (and I think there are even more in the code)
As It's not a big deal having these changes (mostly substracting and adding '1') - I'll keep it as is, and if we will remember, we can remove it in the future

PR #271, opened 2020-10-24
Fix authorship for 'Additional Tracing Modes'

pathtofile:
Fixes authorship of commit https://github.com/aquasecurity/tracee/pull/268 for Additional Tracing Modes

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ You can get Tracee in any of the following ways:
 
 ### Permissions
 
-If run Tracee binary, you'll need to run it with root permissions in order to load the eBPF code. 
+If you use the Tracee binary, you'll need to run it with root permissions in order to load the eBPF code. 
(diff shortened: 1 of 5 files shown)


pathtofile:
doing ` commit --amend --no-edit` always makes the GitHub diff a bit odd, but this accomplishes the goal of showing who last altered the relevant parts of the files.

@yanivagman Is that OK, or do you want to do a full overwrite of the badly-authored commit to remove all traces of it?

yanivagman:
@pathtofile I don't think merging this change will have any impact, as our master branch uses fast-forward policy, meaning you can't change things that were already merged.
I think that in this case, what I proposed (revert+re-commit)) is the only way to accomplish this

pathtofile:
Hm, ok, Will break out into a new branch, new PR: https://github.com/aquasecurity/tracee/pull/272

PR #280, opened 2020-10-28
improve and organize build

itaysk:
- reorganize the makefile to use make best practices and conventions (I have incorporated some of your feedback from the previous PR)
- fix targets prerequisites, every artifact directly depends on the dependencies it requires. use actual file or dir names instead of phony targets when possible
- use make script instead of shell script inside rules when possible
- build all the artifact in the dist directory
- including libbpf is also now built in dist so the source stays clean
- renamed event_monitor_ebpf to tracee_ebpf. it's a better name and especially in the future when we "install" this file in some other directory it need to be meaningful
- discover the binary path at runtime and load the ebpf file from "next" to the binary (instead from a fixed path)
- if we discover the arch for building the ebpf, support cross arch build for the binary as well
- don't hardcode linking paths in cgo, inject them from the makefile


Makefile
```diff
@@ -1,96 +1,79 @@
...
-	go test -v ./...
-
-.PHONY: clean
-clean:
-	rm -rf dist || true
-	cd $(LIBBPF_SRC) && $(MAKE) clean;
-	rm -f $(BPF_OBJ)
-	rm -f *.ll
-	rm -rf .output
+$(OUT_BIN): $(BPF_OBJ) $(LIBBPF_OBJ) $(GO_SRC) | $(OUT_DIR)
```
yanivagman, line 31: Tracee binary (OUT_BIN) doesn't depend on BPF_OBJ - they can be built independently

Makefile
```diff
@@ -1,96 +1,79 @@
...
+		CGO_LDFLAGS="$(abspath $(LIBBPF_OBJ))" \
+		go build -v -o $(OUT_BIN) \
+		-ldflags "-X github.com/aquasecurity/tracee/tracee.ebpfProgramB64Injected=$$(base64 -w 0 $(BPF_SRC))"
 
-imageName ?= tracee
-.PHONY: docker
-docker:
-	docker build -t $(imageName) .
+$(OUT_BIN)_slim: $(BPF_OBJ) $(LIBBPF_OBJ) $(GO_SRC) | $(OUT_DIR)
+	GOOS=linux GOARCH=$(ARCH:x86_64=amd64) go build -v -o $(OUT_BIN)
```
yanivagman, line 39: Regarding ARCH substitution for x86 - Maybe do the same for ARM as well?
uname -m for arm64 gives aarch64, while go uses arm64 to cross compile

Makefile
```diff
@@ -1,96 +1,79 @@
...
+		CGO_CFLAGS="-I $(abspath $(LIBBPF_HEADERS))/bpf" \
+		CGO_LDFLAGS="$(abspath $(LIBBPF_OBJ))" \
+		go build -v -o $(OUT_BIN) \
+		-ldflags "-X github.com/aquasecurity/tracee/tracee.ebpfProgramB64Injected=$$(base64 -w 0 $(BPF_SRC))"
 
-imageName ?= tracee
-.PHONY: docker
-docker:
-	docker build -t $(imageName) .
+$(OUT_BIN)_slim: $(BPF_OBJ) $(LIBBPF_OBJ) $(GO_SRC) | $(OUT_DIR)
```
yanivagman, line 38: Here as well - BPF_OBJ is not a dependency

Makefile
```diff
@@ -1,96 +1,79 @@
...
+	cd $(LIBBPF_SRC) && $(MAKE) OBJDIR=$(abspath $(OUT_DIR))/libbpf BUILD_STATIC_ONLY=1 
 
-$(BPF_OBJ): ${BPF_C} $(LIBBPF_OBJ) | ${OUTPUT}
+linux_arch := $(ARCH:x86_64=x86)
+bpf_extra_headers := 3rdparty/include #copy to out?
+$(BPF_OBJ): $(BPF_SRC) $(LIBBPF_HEADERS) | $(OUT_DIR) $(bpf_compile_tools)
 	$(CLANG) -S \
-		-D __BPF_TRACING__ -D __KERNEL__ -D__TARGET_ARCH_$(ARCH) \
-		$(CFLAGS) \
+		-D __BPF_TRACING__ -D __KERNEL__ -D__TARGET_ARCH_$(linux_arch) \
```
yanivagman, line 65: Let's use this opportunity to fix this for arm as well.
Here linux_arch should be equal to aarch64 (as is the output of uname -m), opposed to the headers in the next lines which should be arm64 (see my comment below)

Makefile
```diff
@@ -1,96 +1,79 @@
...
+		-I $(LIBBPF_HEADERS) \
 		-include $(KERN_SRC)/include/linux/kconfig.h \
-		-I $(KERN_SRC)/arch/$(ARCH)/include \
-		-I $(KERN_SRC)/arch/$(ARCH)/include/uapi \
-		-I $(KERN_SRC)/arch/$(ARCH)/include/generated \
-		-I $(KERN_SRC)/arch/$(ARCH)/include/generated/uapi \
+		-I $(KERN_SRC)/arch/$(linux_arch)/include \
```
yanivagman, line 68: Here, linux_arch should be arm64 (and not aarch64 given by uname -m) for arm64 architecture

yanivagman (changes requested):

Makefile
```diff
@@ -1,96 +1,79 @@
-OUTPUT := .output
-TARGET := event_monitor_ebpf
+.PHONY: all
+all: build
+
+# environment:
+ARCH := $(shell uname -m)
```
yanivagman, line 5: Maybe we better use conditional assignment here:
ARCH ?= 
WDYT?
Same for tools like LLC, CLANG and LLVM-STRIP
This way, the user will be able to change the behavior from cmdline without editing the makefile

Makefile
itaysk: yes, it was depended before when I embedded it in. will fix

Makefile
itaysk: do we support ARM? :)

Makefile
itaysk: I'm removing this slim version for now

Makefile
itaysk: the user can change override vars regardless of how they're assigned. the ?= is just if we want to leave it undefined (which we aren't in this case)

yanivagman:
Regarding renaming to trace_ebpf.c
I agree we should rename.
The convention in libbpf-tools is to use naming like: "tracee.bpf.c"
Actually, today ebpf=bpf in the kerenl, as there is no more cbpf subsytem in the kernel.
It's not a must, just a name, so I don't really mind - what do you think?

Makefile
```diff
@@ -28,7 +28,7 @@ $(OUT_DIR):
 .PHONY: build
 build: $(OUT_BIN)
 
-$(OUT_BIN): $(BPF_OBJ) $(LIBBPF_OBJ) $(GO_SRC) | $(OUT_DIR)
+$(OUT_BIN): $(LIBBPF_OBJ) $(GO_SRC) | $(OUT_DIR)
```
yanivagman, line 31: Note that default build is OUT_BIN - meaning that if you remove this dependency, you should add the BPF_OBJ target to PHONY build (or build all that you removed)

yanivagman (changes requested):

itaysk:
> Regarding renaming to trace_ebpf.c

sure

Makefile
yanivagman: Are you sure this will not override what the user chosen?
From what I understand, ?= is conditionaly setting the variable if wasn't set by the user, but I may be wrong here:
https://stackoverflow.com/questions/53369903/use-environment-variable-if-set-otherwise-use-default-value-in-makefile

Makefile
yanivagman: Like I said in the other comment:
"Let's use this opportunity to fix this for arm as well."
One day we will support it :-)

itaysk:
@yanivagman  Regarding renaming to tracee_ebpf.c - do you want to change the obj as well to `tracee.ebpf.o`?

yanivagman:
@itaysk maybe to tracee.bpf.o? (without the 'e')

Makefile
itaysk: I just tested and it works.. also written here:
> When you override a variable with a command line argument, you can define either a recursively-expanded variable or a simply-expanded variable
https://www.gnu.org/software/make/manual/html_node/Overriding.html#Overriding

itaysk: having said that, there might be a different if people want to use environment variable to override cars.. I'll check

Makefile
itaysk: it feels weird having special build mode for arm when we don't intend to support it soon.. people might get the wrong impression. maybe it's better to add this to the makefile in the same pr that adds arm support no?

Makefile
itaysk: `make build LLC=myllc` works with `:=` but `LLC=myllc make build` doesn't. I'll change

Makefile
yanivagman: Sure, no problem

yanivagman:
@itaysk maybe can you squash to one or two commits before we merge this?

itaysk:
@yanivagman  I will squash it here in github to one commit after you approve

yanivagman (approved):

PR #285, opened 2020-11-02
bpf object build and discovery

itaysk:
search for bpf obj file in predefined locations, also customizable
built the obj file if it's not found, and install it in a permanent location in the system

This is still draft because I still need to embed the libbpf and extra header files instead of expecting to find them locally

### DIFF
--- a/main.go
+++ b/main.go
@@ -190,8 +190,8 @@ func main() {
 			},
 			&cli.StringFlag{
 				Name:        "install-path",
-				Value:       "/opt/tracee",
-				Usage:       "path where tracee will install permanent resources",
+				Value:       "/tmp/tracee",
+				Usage:       "path where tracee will install or lookup it's resources",
 				Destination: &traceeInstallPath,
 			},
 		},
@@ -403,15 +403,15 @@ func getBPFObject() (string, error) {
 		if debug {
 			fmt.Printf("could not find tracee bpf object file in any of %v. attempting to build it\n", searchPaths)
 		}
-		madeBPFObjFilePath := filepath.Join(traceeInstallPath, bpfObjFileName)
-		err = makeBPFObject(madeBPFObjFilePath)
+		bpfObjInstallPath := filepath.Join(traceeInstallPath, bpfObjFileName)
+		err = makeBPFObject(bpfObjInstallPath)
 		if err != nil {
 			return "", err
 		}
 		if debug {
-			fmt.Printf("successfully built ebpf obj file into: %s\n", madeBPFObjFilePath)
+			fmt.Printf("successfully built ebpf obj file into: %s\n", bpfObjInstallPath)
 		}
-		bpfObjFilePath = madeBPFObjFilePath
+		bpfObjFilePath = bpfObjInstallPath
 	}
 
 	return bpfObjFilePath, nil
--- a/main.go
+++ b/main.go
@@ -22,6 +22,7 @@ import (
 
 var debug bool
 var traceeInstallPath string
+var buildPolicy string
 
 // This var is supposed to be injected *at build time* with the contents of the ebpf c program
 var bpfBundleInjected string
@@ -170,7 +171,7 @@ func main() {
 			&cli.StringSliceFlag{
 				Name:  "capture",
 				Value: nil,
-				Usage: "capture artifacts that were written, executed or found to be suspicious. captured artifacts will appear in the 'output-path' directory. possible values: 'write'/'exec'/'mem'/'all'. use this flag multiple times to choose multiple capture options",
+				Usage: "capture artifacts that were written, executed or found to be suspicious. captured artifacts will appear in the 'output-path' directory. possible options: 'write'/'exec'/'mem'/'all'. use this flag multiple times to choose multiple capture options",
 			},
 			&cli.StringSliceFlag{
 				Name:  "filter-file-write",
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -324,3 +353,238 @@ func printList() {
...
+	}
+	bpfObjFilePath := locateFile(bpfObjFileName, searchPaths)
+	if bpfObjFilePath != "" && debug {
+		fmt.Printf("found bpf object file at: %s\n", bpfObjFilePath)
+	}
+	if bpfObjFilePath == "" {
+		if debug {
+			fmt.Printf("could not find tracee bpf object file in any of %v. attempting to build it\n", searchPaths)
+		}
+		madeBPFObjFilePath := filepath.Join(traceeInstallPath, bpfObjFileName)
```
yanivagman, line 399: made=make?

main.go
```diff
@@ -324,3 +353,238 @@ func printList() {
...
+		if err != nil {
+			return "", err
+		}
+		if debug {
+			fmt.Printf("succesfully built ebpf obj file into: %s\n", madeBPFObjFilePath)
+		}
+		bpfObjFilePath = madeBPFObjFilePath
+	}
+
+	return bpfObjFilePath, nil
```
yanivagman, line 410: What if we want to force re-comilation of the bpf code?
For example, when new kernel is installed?
Maybe we shuold fallback to compilation if bpf code loading failed?

main.go
```diff
@@ -324,3 +353,238 @@ func printList() {
...
+		fmt.Printf("building bpf object in: %s\n", dir)
+	} else {
+		defer os.RemoveAll(dir)
+	}
+	objFile := filepath.Join(dir, "tracee.bpf.o")
+	bpfSrc := getBPFSourceFile(dir)
+	if bpfSrc == "" {
+		return fmt.Errorf("could not get bpf source file")
+	}
+	clang := locateFile("clang", []string{os.Getenv("CLANG")})
```
yanivagman, line 443: Not sure this is a good idea (security wise).
Tracee is running with root privileges, we have to be sure we are not executing a malicious executables that was found during search.

yanivagman: Maybe we can drop privileges before execution of these files

tracee/tracee.go
```diff
@@ -388,7 +372,7 @@ func (t *Tracee) initBPF(ebpfObjectPath string) error {
 	//       5. Populate maps with values,
 	//       6. Attach probes,
 	//       7. Initialize perf buffers
-	t.bpfModule, err = bpf.NewModuleFromFile(ebpfObjectPath)
+	t.bpfModule, err = bpf.NewModuleFromFile(bpfObjectPath)
```
yanivagman, line 375: If bpf t.bpfModule.BpfLoadObject() failed (line 406) - we can give the user a hint that he may need to recompile the code (this may happen if kernel version changed, and we use old bpf code binary)

main.go
itaysk: I plan to add "install" and "uninstall" commands to tracee. for now the user can just remove /opt/tracee and it will recompile

main.go
itaysk: this is a good feedback, but I won't have time to implement it soon, maybe end of week or next week. Do you want to wait for it, or merge it and fix in a followup PR?

itaysk: we need to "install" as root anyway in order to copy the files into a permanent location

yanivagman: Doing make doesn't require root privileges.
Tracee will not start without root privileges.
We can install not as root, and save to some temp dir.
I think this is better to put the object file in a dir that is deleted after reboot anyways. This way, you can be sure that the bpf code was compiled with the correct kernel version. If the user will want to install it to a different location that requires root privileges, he will need to do this explicitly

tracee/tracee.go
yanivagman: If we install to a tmp dir that is deleted after reboot - this is not an issue

main.go
```diff
@@ -324,3 +367,278 @@ func printList() {
...
+		return fmt.Errorf("missing compilation dependency: kernelSource")
+	}
+	linuxArch := os.Getenv("ARCH")
+	if linuxArch == "" {
+		linuxArch = strings.Replace(runtime.GOARCH, "amd64", "x86", 1)
+	}
+
+	// from the Makefile:
+	// $(CLANG) -S \
+	// 	-D__BPF_TRACING__ \
```
yanivagman, line 523: I think this comment is redundant as it is clear from the code

tracee/tracee.bpf.c
```diff
@@ -26,8 +26,8 @@
 
 #undef container_of
 //#include "bpf_core_read.h"
-#include <bpf/bpf_helpers.h>
-#include <bpf/bpf_tracing.h>
+#include <bpf_helpers.h>
```
yanivagman, line 29: Does compilation with the regular make works after this change?

itaysk: yes

main.go
itaysk: Since this code section is basically a port of 3 commands from the makefile, I noted the original commands in context with the code that ports them. I want to make sure that the build from make and the build from go is *exactly* the same, and that if someone touches this code it's clear that they either need to change the makefile as well or they should reconsider their change because this is how it compiles from the makefile.  I also think we should update this comment if we change the makefile in the future.

yanivagman (approved):

PR #289, opened 2020-11-08
Fix lint errors

yanivagman:


itaysk:
is this based on `go fmt -l` ? what was missing before that we didn't see these recommendations?

yanivagman:
use gofmt without space (go fmt)

yanivagman:
I didn't see all errors though (different gofmt version?)

itaysk:
since lint commits are going to pollute the history, it's better to make sure once commit fixes all the issues once and for all.  
are you on go 1.15?

yanivagman:
I'm on go 1.15.
Lint test passed on my repo:
https://github.com/yanivagman/tracee/actions

itaysk (approved):

PR #290, opened 2020-11-08
Fix tests

### LINKED ISSUE #252: Tests are failing on master

yanivagman:
Fixes #252


itaysk (commented):
this "fixes" the tests by forcing them to pass but it doesn't actually tests the correctness of the code. This test was written to test how `readArgFromBuff` parses and translates binary data to user readable strings. Since than, the `readArgFromBuff` was changed to simply return the binary data. So the in order to actually fix these tests we should invoke the new piece of code that now does the translation, otherwise it is not worthwhile to keep this test

yanivagman:
I think you are wrong.
This test still validates the correctness of the parsing.
If you'll have a look at the expected values, you will see all the values from before still exists, just not as strings.
This test verifies that the parsing is correct, and the values (in types) are what we expect.
The printing stage (stringify these values), is another step, which we can write a different test for.
Anyways, I think we should merge this change now, so the tests will pass and we can add it to github actions

itaysk (approved):

sudiptog81:
As the tests now pass, should I go ahead with creating the CI job?

itaysk:
@sudiptog81 that would be great, thanks!

PR #292, opened 2020-11-08
fix docker builds for libbpf

itaysk:
I also took the opportunity to upgrade the base image to ubuntu focal.
note that the builder image version will need to change to go 1.15 when libbpf is merged with master (currently libbpf is still on 1.13)


itaysk:
one thing I'm not 100% sure about is libelf-dev (which includes zlib as well) vs libelf1. the fact that zlib is not needed at runtime means we can remove it from the cgo headers as well?

yanivagman (approved):

PR #293, opened 2020-11-10
Rename bpfwrap to libbpfgo

yanivagman:


itaysk:
maybe add the go.mod file as well?

itaysk (approved):

yanivagman:
Let's add the go module after merging with master.
Pulling the module fails as url doesn't exist (in master).

PR #294, opened 2020-11-10
feat: Add tests to CI/CD workflow

sudiptog81:
## Summary

Added tests as a job in the GitHub Actions workflow. Related to #254.

## Possible Alternative

We can also alternatively add a new section in the `Makefile` and run tests in a container. However, could not detect failing tests with that approach.

```makefile
.PHONY: test-docker
test-docker: clean
  img=$$(docker build --target builder -q .) && \
  cnt=$$(docker create $$img); \
  docker run -a stdout $$cnt make test; \
  docker rm $$cnt ; docker rmi $$img
```

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -3,7 +3,7 @@ name: build-test
 on: pull_request
 
 jobs:
-  build-test:
+  build:
     runs-on: ubuntu-20.04
 
     env:
@@ -31,3 +31,14 @@ jobs:
 
       - name: Build Tracee
         run: make build-docker
+  
+  test:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      - name: Run Tests
+        run: make test-docker
+      


yanivagman:
Hi @sudiptog81 ,
Thanks for your contributions so far!

I think we better wait with this one until we merge the other PR I opened (move to libbpf) which makes some major changes.
BCC is not required anymore (which I see in the "Installed libraries" you added here),
and also, it fixes the Lint problems introduced yesterday (which still fails on this branch)

.github/workflows/build-test.yaml
```diff
@@ -31,3 +31,45 @@ jobs:
...
+            echo "The following lines should be fixed."
+            gofmt -s -d .
+            exit 1
+          fi
+
+      - name: Install Libraries
+        run: |
+          sudo apt-get update
+          sudo apt install -y bison build-essential cmake flex git libedit-dev \
+              libllvm8 llvm-8-dev libclang-8-dev libllvm9 llvm-9-dev libclang-9-dev \
```
yanivagman, line 65: is there a reason to have both llvm8 and llvm9 (same for clang)?

.github/workflows/build-test.yaml
```diff
@@ -31,3 +31,45 @@ jobs:
...
+              python zlib1g-dev libelf-dev libtinfo5 libfl-dev
+          git clone https://github.com/iovisor/bcc.git
```
yanivagman, line 67: please rebase on latest changes.
bcc is not a dependency anymore (I guess these 5 lines can now be removed)

yanivagman (changes requested):

.github/workflows/build-test.yaml
sudiptog81: Was having some issues with the `ubuntu-20.04` image and building `bcc` when that was required. Removing these now as `bcc` is no longer a dependency.

yanivagman:
If you open "Run Tests" in checks, it seems like it didn't run:
"make: Nothing to be done for 'test'."

I opened a PR to fix this in: #307 
We should rebase after merge and verify tests run correctly

sudiptog81:
> If you open "Run Tests" in checks, it seems like it didn't run:
> "make: Nothing to be done for 'test'."

The relevant section in `Makefile` was commented out on the upstream. I think we can go ahead after #307 gets merged as this PR does not change that file.

yanivagman:
#307 is now merged.
Can you please rebase so we can be sure everything works as expected?

sudiptog81:
The `LIBBPF_HEADERS` artifacts have to be taken care of.

.github/workflows/build-test.yaml
```diff
@@ -31,3 +31,40 @@ jobs:
...
+      - name: Setup Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.15
+        id: go
+
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      - name: Lint
```
yanivagman, line 51: Do we need the Lint here?
This is already done in "build" job

sudiptog81: Removed the Lint from "test" as suggested.

itaysk:
@sudiptog81 I'm working to fix the `make test` target now. it should fix this issue

itaysk:
https://github.com/aquasecurity/tracee/pull/313 fixes this issue but also adds a `make test-docker` target that should simplify this for CI. I propose to use this instead of installing all the dependencies every time

sudiptog81:
I rebased the PR and used the `test-docker` target.

.github/workflows/build-test.yaml
```diff
@@ -31,3 +31,23 @@ jobs:
...
+
+    env:
+      CGO_ENABLED: "1"
+
+    steps:
+      - name: Setup Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.15
+        id: go
```
itaysk, line 46: do we need this now that we're using docker?

sudiptog81: Not needed now as the tests run in container itself, removed this step in the next commit.

itaysk: I see that you've removed the "Setup Go" step but left the "env" section. is it needed?

sudiptog81: The `env` section is not needed now as we are not using the variables in the test runner.

itaysk (approved):

yanivagman (approved):

PR #299, opened 2020-11-11
Don't make llvm-strip a dependency

yanivagman:


main.go
```diff
@@ -615,20 +613,22 @@ func makeBPFObject(outFile string) error {
...
-		cmd3.Stdout = os.Stdout
-		cmd3.Stderr = os.Stderr
-	}
-	err = cmd3.Run()
-	if err != nil {
-		return err
+	if llvmstrip != "" {
+		// from Makefile:
+		// $(LLVM_STRIP) -g $@
+		cmd3 := exec.Command(llvmstrip,
```
itaysk, line 619: really small nit:  1) copy the updated make command (with the `-`). 2) move the `if` to after the comment, since it's part of the implementation of the make command in go (`-`->`if`). I know it's tiny but since it's same BL it's better to keep both in sync

itaysk (changes requested):

itaysk (approved):

PR #301, opened 2020-11-11
Update readme with build comments

yanivagman:


README.md
```diff
@@ -20,6 +20,8 @@ Other than tracing, Tracee is also capable of capturing files written to disk or
 To run, Tracee requires the following:
 - Linux kernel version > 4.14
 - Kernel headers
+
+In addition, if running the binary directly (not via docker):
```
itaysk, line 24: I don't think this is needed.. libc and clang are always required. in docker we baked them in the image, but they are still required by tracee.

README.md
```diff
@@ -20,6 +20,8 @@ Other than tracing, Tracee is also capable of capturing files written to disk or
...
 - C standard library (currently tested with glibc)
 - clang (currently tested with version > 10)
```
itaysk, line 26: we need to test which minimum version is required. if you `apt-get install clang` the stable release in debian gets you clang7. and I think it works

README.md
```diff
@@ -28,7 +30,8 @@ To run, Tracee requires the following:
 You can get Tracee in any of the following ways:
 1. Download the binary from the GitHub Releases tab (`tracee.tar.gz`).
 2. Use the docker image from Docker Hub: `aquasec/tracee`. The image already includes the required dependencies but you will need to mount the kernel headers in (see below for example).
-3. Build from source, using `make build` (or via Docker using `make build-docker`).
+3. Build from source via docker, using `make build-docker`.
+4. Build from source, using `make build` (requires libelf-dev in addition to the above prerequisites).
```
itaysk, line 34: I think this will be a transitive dependency of clang anyway. I explicitly installed it in the dockerfile because I'm using 1) a minimal container 2) --no-install-recommends. this is also the reason why I install llvm there which we should not mention here.

itaysk (changes requested):

README.md
yanivagman: ok

README.md
yanivagman: Version 6 didn't work for me - So 7 it is

README.md
yanivagman: I got some gelf.h missing problems with clang until I installed libelf-dev - I don't think this is a transitive dependency like you say

itaysk (approved):

PR #302, opened 2020-11-12
rewrite release process

itaysk:
we've had doubts about the value of goreleaser. we wanted to support both makefile and docker builds, but still be able to release using goreleaser, but goreleaser wants to build it's own binary. This led to some workarounds like the `release.Dockerfile` and code duplication between the makefile and goreleaser. Now that we've moved to libbpf, the build toolchain has gotten even more complex and we have greater motivation to concentrate our efforts in the makefile. What's more, building the current project in goreleaser is not so straightforward anymore.
I have proposed a couple of suggestions in goreleaser that if accepted could improve the situation for us, but it doesn't seem like there is interest in them so we have chosen to replace goreleaser with our own release script.


yanivagman:
Seems like CI build-test fails...

yanivagman (approved):

PR #305, opened 2020-11-12
Fix asm_inline for kernel > 5.4

yanivagman:


yanivagman:
Tested with ubuntu 20.04

itaysk (commented):
is this part of the source code? I think this should be in a separate header file (in bpf "extra" headers). wdyt?

yanivagman:
What do you mean part of the source code?
This is a workaround
Open a header file just for this? I don't know...

itaysk (approved):

PR #306, opened 2020-11-12
Fix compilation on ubuntu 16

yanivagman:


yanivagman:
Tested on ubuntu 16.04

itaysk (commented):
I think it would be better to add any libbpf related headers to the `$(LIBBPF_HEADERS)` target, rather then importing directly from 3rdparty source. Also, there's a `make install_uapi_headers` target in libbpf that's may be what you were looking for? if should be called from our `$(LIBBPF_HEADERS)` target

yanivagman:
The problem with their `make install_uapi_headers` is that it install it to the same directory as `make install headers`, which causes the files bpf.h and btf.h to get overriden (and they are not the same files).

From libbpf makefile:
```
PREFIX ?= /usr
LIBDIR ?= $(PREFIX)/$(LIBSUBDIR)
INCLUDEDIR ?= $(PREFIX)/include
UAPIDIR ?= $(PREFIX)/include
```

Any suggestions?

yanivagman:
Nevermind, I think I got it

itaysk (approved):
approved, is the bpf subdir needed?

yanivagman:
Yes, the bpf subdir is where the libbpf headers are copied to, while the linux subdir is where the linux uapi headers are copied to.
Both are in the LIBBPF_HEADERS dir (dist/libbpf/usr/include)

PR #310, opened 2020-11-16
check minimum clang version

itaysk:


main.go
```diff
@@ -501,6 +501,18 @@ func makeBPFObject(outFile string) error {
...
+	cmdVer := exec.Command(clang, "--version")
+	verOut, err := cmdVer.CombinedOutput()
+	if err != nil {
+		return err
+	}
+	// version output looks like: clang version 9.0.0-2 (tags/RELEASE_900/final)\n more lines...
+	eol := strings.Index(string(verOut), "\n")
+	verStr := string(verOut[14:eol])
+	verMajor, err := strconv.Atoi(strings.SplitN(verStr, ".", 2)[0])
+	if err != nil || verMajor < 7 {
```
yanivagman, line 513: I would separate this into error, and less than 7.
If and error detected - warn, but continue
Exit if major<7

main.go
```diff
@@ -501,6 +501,18 @@ func makeBPFObject(outFile string) error {
...
+		return fmt.Errorf("required minimum clang version: 7")
```
yanivagman, line 514: Can you add the verStr as part of the output (something like: "detected version: xxx")

main.go
itaysk: if `clang --version` returned an unrecognized format, you want to ignore?

yanivagman: The problem is that "clang --version" is not guranteed to keep its format between versions...
Do we want to fail if this is changed in new versions?

itaysk: I'm ok with both options, just wanted to clarify your request. you prefer to ignore?

yanivagman: I think so...

itaysk:
@yanivagman I found a clang version that's not compatible, will send an update soon

yanivagman (approved):

Issue #312, opened 2020-11-16
Filter by UID


itaysk:
extend the current `--trace` flag to support in-kernel filtering by user id, like so: `--trace uid:1001`

grantseltzer:
What are your thoughts on having a flag for `--uid` instead of lumping it in on `--trace`?

The way I understand it, there are two 'modes'. Container mode, and process mode (or non-container mode). UID filtering is applicable for both modes so wouldn't affect that. There are also other filters that could be added that would be applicable in either mode. Some examples would be GID, PPID, particular namespaces, or cgroups (not sure how tracee handles k8s pods at the moment).

Here's my idea for a more intuitive design and am happy to iterate based on your thoughts:

- Remove the `--trace` flag.
- For each value a user wants to filter by, they specify it with a corresponding flag.
  - `--uid`, `--gid` (both default to all)
  - `--pid`, `--ppid` (both default to all, can also specify new, list of pids, or specify multiple times)
  - `--container` (default to all, specific container ID, multiple IDs, 'new' or 'all')
  - `--ns`, `--cgroup` 
  - ...
- All of the specified filters would be logical "AND"'ed together. For example `tracee --pid=new --uid=1000` would only produce new processes with uid 1000.
- The 'mode' (container/process) can be implied by the use of flags. If `--container='___'` is specified then the mode will be container and the `pid`/`ppid` flags (if specified) will be ignored. Applicable filter flags would be honored. 

Alternatively I can see:

- Adding a `--filter` flag
  - Can be specified any number of times.
  - Arguments would be of the form `<filter>:<value>` (example: `pid:56132`, `pid:new`, `uid:0`, `gid:10`)

yanivagman:
Hi @grantseltzer,

We had a long discussion in #221 about some of the things you wrote here.
In the end, there are different ways to tackle this, and we need to think about a clean and intuitive UX,
but it's probably also a matter of taste...

Like you said, we currently have "modes" of operation.
We can consider these "modes" as filters, but I think that what makes them different from other filters, is that they represent the traced entity(ies) or workload(s), which dynamically change over time, and as such also have the "new" flag.
Opposed to other filters which can be logical "AND"ed together, these modes are also usually mutual exclusive - there is no sense in telling tracee to filter both new processes and all processes, or all processes and newly created (or specific) containers. 

Now comes the question that we still need to discuss - do we think that it's a good idea to integrate the "--trace" flag in our new filters design? I still don't have a clear opinion about this.
[...]

PR #315, opened 2020-11-19
Add a `--filter` flag and support for filtering of events by uid

grantseltzer:
_(I marked this as a draft because of a verifier issue, see context below)_

This adds a `--filter` flag to tracee. Currently the only supported value to filter by is `uid`. In this proposed solution users can specify filters of the form `<field>=<value>`, or multiple by using a comma separated list or specifying the flag multiple times:      `--filter <field>=<value1>,<value2>` or `--filter <field>=<value1> --filter <field>=<value2>`

In the case of uid filtering, the way I've implemented this is by parsing passed values, and sharing two values between user and kernel space: an array of the uids in their own BPF_ARRAY and a field in the BPF_HASH `config_map` for the length of this array. 

#### Verifier issue:
The issue is that the verifier complains about the loop in bpf which reads from the array of uids. Having the length of array (and therefore # of loop iterations) pulled from a variable is enough to reject the program.

I have a few thoughts on how we can get around this and their trade-offs:

- Is there a way of working around the verifier in bpf code that i'm not familiar with? I thought maybe declaring the length value as a const would help but it didn't.

- If we set a hard limit on the number of UIDs we can satisfy the verifier by bounding the loop at that limit. 
  - Can we picture someone specifying more than a handful of uids? Why would they not just create a group and pass a gid (once we add that as a supported filter option).
  - If so, what would we set this limit to? 15? 50? Regardless it seems like it would be arbitrary.

- We can move the filtering into Go code.
  - Likely takes a performance hit, but not sure how much, we need a bench marking suite in CI/CD for tracee. 
  - This would reduce the complexity of the bpf code. Is having filtering in bpf worth the number of instructions it takes up?
  - There may be fields we want to support filtering by that aren't surfaced to Go code and need to be done on the bpf side. There should be parity between all filter fields for the sake of code complexity. 

- We can just not have this feature at all and let users rely on unix tools like `grep` or `jq`
  - If someone wanted to consume tracee as a package in their own code this would limit the usefulness of the API.

Relevant issue: #312

### DIFF
--- a/main.go
+++ b/main.go
@@ -47,10 +47,14 @@ func main() {
 			if err != nil {
 				return err
 			}
-
+			filter, err := prepareFilter(c.StringSlice("filter"))
+			if err != nil {
+				return err
+			}
 			cfg := tracee.TraceeConfig{
 				EventsToTrace:         events,
 				Mode:                  mode,
+				Filter:                filter,
 				DetectOriginalSyscall: c.Bool("detect-original-syscall"),
 				ShowExecEnv:           c.Bool("show-exec-env"),
 				OutputFormat:          c.String("output"),
@@ -134,7 +138,13 @@ func main() {
 				Name:    "trace",
 				Aliases: []string{"t"},
 				Value:   "process:new",
-				Usage:   "Set trace mode, whether to trace processes or containers, and if to trace new, all, or specific processes/container. run '--trace help' for more info",
+				Usage:   "set trace mode, whether to trace processes or containers, and if to trace new, all, or specific processes/container. run '--trace help' for more info",
+			},
+			&cli.StringSliceFlag{
+				Name:    "filter",
+				Aliases: []string{"f"},
+				Value:   nil,
+				Usage:   "set tracing filters for specific fields (such as UID or GID). run '--filter help' for more info.",
 			},
 			&cli.BoolFlag{
 				Name:  "detect-original-syscall",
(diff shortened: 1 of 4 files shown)


yanivagman:
Passing the verifier can be challenging sometimes, and require us to think differently about how to solve a given problem.
Have you tried using BPF_HASH instead of BPF_ARRAY for storing the uids and save the loop?

grantseltzer:
I thought of having a uid_filter struct which contains a `u32` array and placing that in the main configuration BPF_HASH, but how do you see that being able to avoid the loop?

yanivagman:
What if you store the uids given by the user in a hash map.
The value is not important, but if the key exists in the map - it means that the user requested to filter by it.
Then the kernel can check if the current uid exists as a key in the hash map

grantseltzer:
Ah @yanivagman, that's far more clever. I will do exactly that.

grantseltzer:
Alright I went ahead with your suggestion and it works great. I was thinking about the bpf maps too abstractly, after all it's just a map :)

Removing the 'draft' status of the PR since it's working to my liking for now.

main.go
```diff
@@ -210,11 +220,58 @@ func main() {
...
 }
 
+func prepareFilter(filters []string) (tracee.Filter, error) {
+
+	uids := []uint32{}
+
+	filterHelp := "\n--filter allows you to specify values to match on for fields of traced events:\n"
+	filterHelp += "The following options are currently supported:\n"
+	filterHelp += "uid: only trace processes or containers with specified uid(s).\n"
+	filterHelp += "\t--filter uid=0,1000                                           | only trace events from uid 0\n"
```
yanivagman, line 230: The text and comments of these two lines switched?

grantseltzer:
I can also add filtering by gid to this PR, it'd be a really minimal addition.

tracee/tracee.bpf.c
```diff
@@ -227,22 +228,23 @@ struct mount {
...
-BPF_HASH(params_types_map, u32, u64);               // Encoded parameters types for event
-BPF_HASH(params_names_map, u32, u64);               // Encoded parameters names for event
-BPF_ARRAY(file_filter, path_filter_t, 3);           // Used to filter vfs_write events
-BPF_ARRAY(string_store, path_filter_t, 1);          // Store strings from userspace
-BPF_PERCPU_ARRAY(bufs, buf_t, MAX_BUFFERS);         // Percpu global buffer variables
-BPF_PERCPU_ARRAY(bufs_off, u32, MAX_BUFFERS);       // Holds offsets to bufs respectively
-BPF_PROG_ARRAY(prog_array, MAX_TAIL_CALL);          // Used to store programs for tail calls
-BPF_PROG_ARRAY(sys_enter_tails, MAX_EVENT_ID);      // Used to store programs for tail calls
-BPF_PROG_ARRAY(sys_exit_tails, MAX_EVENT_ID);       // Used to store programs for tail calls
+BPF_HASH(config_map, u32, u32);                         // Various configurations
```
yanivagman, line 231: Please keep the original indentation of the maps if there is no reason to change it

tracee/tracee.bpf.c
```diff
@@ -495,15 +497,34 @@ static __always_inline int get_config(u32 key)
     return *config;
 }
 
+static __always_inline int uid_filter_matches()
+{
+	uid_t uid = (u32)bpf_get_current_uid_gid();
```
yanivagman, line 502: Please use spaces instead of tabs (this what happens when we have both go and c in the same repo)

tracee/tracee.go
```diff
@@ -423,6 +429,11 @@ func (t *Tracee) populateBPFMaps() error {
 		fileFilterMap.Update(uint32(i), []byte(t.config.FilterFileWrite[i]))
 	}
 
+	uidHash, _ := t.bpfModule.GetMap("uid_filter")
+	for i := 0; i < len(t.config.Filter.UIDs); i++ {
+		uidHash.Update(uint32(t.config.Filter.UIDs[i]), uint32(1))
```
yanivagman, line 434: Just an idea, tell me what you think
The value of the uid_map can take one of, say, "0", "1", "2", which will represent - "<", "=", ">"
Then the user will be able to specify uid filters as:
"--filter uid>0", for example

Anyways, we don't have to implement this now

yanivagman (commented):
Great!
Please see my comments

tracee/tracee.go
grantseltzer: I like it! Can also have one that represents "not" (like uid != 1000).

I can add that as part of this PR or in a follow up.

main.go
```diff
@@ -210,11 +220,58 @@ func main() {
...
 
+func prepareFilter(filters []string) (tracee.Filter, error) {
+
+	uids := []uint32{}
+
+	filterHelp := "\n--filter allows you to specify values to match on for fields of traced events.\n"
+	filterHelp += "The following options are currently supported:\n"
+	filterHelp += "uid: only trace processes or containers with specified uid(s).\n"
+	filterHelp += "\t--filter uid=0                                                | only trace events from uid 0 or uid 1000\n"
+	filterHelp += "\t--filter uid=0,1000                                           | only trace events from uid 0\n"
```
yanivagman, line 231: Comments of the above two lines mixed?

tracee/tracee.go
yanivagman: If it's not too many additions, you can add it here,
otherwise, we can do a follow up PR

main.go
grantseltzer: Ah thought I fixed this, Idk how I lost that change. Pushing now.

tracee/tracee.go
grantseltzer: I'll give it a shot and you can make the call on if you want me to cherry-pick the changes onto a new PR instead. I typically like small PRs.

tracee/tracee.bpf.c
```diff
@@ -495,15 +497,34 @@ static __always_inline int get_config(u32 key)
     return *config;
 }
 
+static __always_inline int uid_filter_matches()
+{
+    uid_t uid = (u32)bpf_get_current_uid_gid();
+    u32* isMatch = bpf_map_lookup_elem(&uid_filter, &uid);
+    return (isMatch == NULL) {
```
yanivagman, line 504: Seems like this won't compile
return -> if

yanivagman:
LGTM

Can you please squash into one commit?

grantseltzer:
You got it!

yanivagman (approved):

PR #319, opened 2020-11-19
improve building in docker

itaysk:
This PR is re-organizing the build in docker code, improving the build targets user experience, and adding visibility into the build progress. it also bumps the required llvm version from 7 to 9, as we found issues with 7.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@ WORKDIR /tracee
 
 FROM tracee-builder as build
 COPY . /tracee
-RUN git submodule update --init && make build
+RUN make build
 
 # base image for tracee which includes all tools to build the bpf object at runtime
 FROM ubuntu:focal as fat
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,11 @@ WORKDIR /tracee
 
 COPY . /tracee
 
+
+FROM builder as build
+
 RUN git submodule update --init && make build
+
 
 # must run privileged and with linux headers mounted
 # docker run --name tracee --rm --privileged --pid=host -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro aquasec/tracee
@@ -18,6 +22,6 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update -y && apt-get install -y --no-
 
 WORKDIR /tracee
 
-COPY --from=builder /tracee/dist/tracee /tracee/entrypoint.sh ./
+COPY --from=build /tracee/dist/tracee /tracee/entrypoint.sh ./
 
 ENTRYPOINT ["./entrypoint.sh", "./tracee"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,8 @@
 FROM golang:1.15-buster as builder
 
-RUN DEBIAN_FRONTEND=noninteractive apt-get update -y && apt-get install -y --no-install-recommends libelf-dev llvm clang
+RUN echo "deb http://apt.llvm.org/buster/ llvm-toolchain-buster-9 main" >> /etc/apt/sources.list && apt-key adv --keyserver hkps://keyserver.ubuntu.com --recv-keys 15CF4D18AF4F7421 && \
+    DEBIAN_FRONTEND=noninteractive apt-get update -y && apt-get install -y --no-install-recommends libelf-dev llvm-9-dev clang-9 && \ 
+    (for tool in "clang" "llc" "llvm-strip"; do path=$(which $tool-9) && ln -s $path ${path%-*}; done)
 
 WORKDIR /tracee
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,27 +1,29 @@
-FROM golang:1.15-buster as builder
+ARG BASE=fat
 
+FROM golang:1.15-buster as builder
 RUN echo "deb http://apt.llvm.org/buster/ llvm-toolchain-buster-9 main" >> /etc/apt/sources.list && apt-key adv --keyserver hkps://keyserver.ubuntu.com --recv-keys 15CF4D18AF4F7421 && \
     DEBIAN_FRONTEND=noninteractive apt-get update -y && apt-get install -y --no-install-recommends libelf-dev llvm-9-dev clang-9 && \ 
     (for tool in "clang" "llc" "llvm-strip"; do path=$(which $tool-9) && ln -s $path ${path%-*}; done)
-
 WORKDIR /tracee
 
+FROM tracee-builder as build
 COPY . /tracee
-
-
-FROM builder as build
-
 RUN git submodule update --init && make build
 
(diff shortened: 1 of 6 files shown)


itaysk:
I see the build failed, looking into it

README.md
```diff
@@ -21,15 +21,18 @@ To run, Tracee requires the following:
...
+
 You can get Tracee in any of the following ways:
-1. Download the binary from the GitHub Releases tab (`tracee.tar.gz`).
-2. Use the docker image from Docker Hub: `aquasec/tracee`. The image already includes the required dependencies but you will need to mount the kernel headers in (see below for example).
-3. Build from source via docker, using `make build-docker`.
-4. Build from source, using `make build` (requires libelf-dev in addition to the above prerequisites).
+1. Download the executable from the [GitHub Releases](https://github.com/aquasecurity/tracee/releases) (`tracee.tar.gz`).
+2. Use the docker image from Docker Hub: `aquasec/tracee`.
+3. Build the executable from source using `make build`. For that you will need additional developement prerequisites such as golang and libelf.
+4. Build from executable from source in a Docker container which includes all the prerequisites, using `make build DOCKER=1`.
```
yanivagman, line 34: Build the...

Dockerfile
```diff
@@ -1,13 +1,19 @@
...
+RUN echo "deb http://apt.llvm.org/buster/ llvm-toolchain-buster-9 main" >> /etc/apt/sources.list && apt-key adv --keyserver hkps://keyserver.ubuntu.com --recv-keys 15CF4D18AF4F7421 && \
+    DEBIAN_FRONTEND=noninteractive apt-get update -y && apt-get install -y --no-install-recommends libelf-dev llvm-9-dev clang-9 && \ 
+    (for tool in "clang" "llc" "llvm-strip"; do path=$(which $tool-9) && ln -s $path ${path%-*}; done)
 
 WORKDIR /tracee
 
 COPY . /tracee
 
+
+FROM builder as build
```
yanivagman, line 12: Why not changing line 1 instead (builder->build)?

itaysk: it's not that builder changed to build. builder remains. build is a new stage. the is to speed up the make DOCKER=1 targets, so they don't need to build the binary just tp prep the builder image (the make will rebuild anyway).
BTW I have a commit that changes the Dockerfile (currently the prod docker image is broken), I just want to fix the build in github first

README.md
```diff
@@ -21,15 +21,18 @@ To run, Tracee requires the following:
 - Linux kernel version > 4.14
 - Kernel headers
 - C standard library (currently tested with glibc)
-- clang > 7
+- clang > 9
 
 ### Getting Tracee
 
+Tracee is made of an executable that drives the eBPF code (`tracee`) and the eBPF code itself (`tracee.ebpf.o`). When the `tracee` executable is started, is will look for the eBPF code next to the executable, or in `/tmp/tracee`. If the eBPF is not found, the executable will attempt to build it (you can control this using the `--build-policy` flag).
```
yanivagman, line 28: tracee.bpf.o

README.md
```diff
@@ -13,44 +13,52 @@ Other than tracing, Tracee is also capable of capturing files written to disk or
...
-## Getting started
+## Getting Started
 
 ### Prerequisites
 
-To run, Tracee requires the following:
-- Linux kernel version > 4.14
-- Kernel headers
+To run, Tracee requires Linux kernel version > 4.14, and the relevant kernel headers available under `/usr/src`.
+The rest of dependencies are already built into the Docker image:
```
yanivagman, line 21: The rest of THE deoencdencies

README.md
```diff
@@ -13,44 +13,52 @@ Other than tracing, Tracee is also capable of capturing files written to disk or
...
 - C standard library (currently tested with glibc)
-- clang > 9
+- libelf, zlib libraries
```
yanivagman, line 23: This is not needed to run tracee.
Only to compile it

README.md
```diff
@@ -13,44 +13,52 @@ Other than tracing, Tracee is also capable of capturing files written to disk or
...
+3. Build the executable from source using `make build`. For that you will need additional development tooling.
+4. Build the executable from source in a Docker container which includes all development tooling, using `make build DOCKER=1`.
+
+Alternatively, you can pre-compile the eBPF program, and provide it to the `tracee` executable. There are some benefits to this approach since you will not need clang at runtime anymore, as well as reduced risk of invoking an external program at runtime. You can build the eBPF program in the following ways:
+1. `make bpf`
+2. `make bpf DOCKER=1` to build in a Docker container which includes all development tooling.
+3. There is also a handy `make all` (and the `make all DOCKER=1` variant) which builds both the executable and the eBPF program.
+
+Once you have the eBPF program artifact, you can provide it to Tracee in any of the locations mentioned above.  
+The `aquasec/tracee:slim` image is a lighter-weight image that cannot build the eBPF program at run time, and is meant to be used when you already compiled the eBPF program beforehand.
```
yanivagman, line 50: It is worth telling how to invoke it, especially how do you tell it where to find the bpf code

README.md
```diff
@@ -13,44 +13,52 @@ Other than tracing, Tracee is also capable of capturing files written to disk or
...
-
-```bash
-docker run --name tracee --rm --privileged --pid=host -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro aquasec/tracee:latest
-```
-
-This will run Tracee with no arguments which will collect all events from all newly created processes and print them as a table to the standard output.
+### Understanding the output
 
-Here is how the output looks:
+Here's an sample output of running Tracee with no additional arguments (defaults to tracing all events):
```
yanivagman, line 61: a sample

README.md
itaysk: are you sure? (not suggesting otherwise, just asking)

README.md
itaysk: you don't think "in any of the locations mentioned above" is clear?

README.md
yanivagman: libbpf.a is statically compiled (.a), and it uses libelf during it's compilation, so I'm quite sure, but worth testing this

itaysk: I'm not sure but isn't it needed to read the obj file?

README.md
yanivagman: In the case of the slim docker image - the user can't put it in the directory of tracee executable neither under /tmp/tracee.
He should mount it, then use the TRACEE_BPF_FILE environment variable. I think it is worth explaining this

itaysk: good point

README.md
yanivagman: I think that when we linked tracee with libbpf.a, the needed code which reads the obj file is already part of the code

itaysk:
```
error while loading shared libraries: libelf.so.1: cannot open shared object file: No such file or directory
```

yanivagman: Ok, good to know. Indeed was worth the test :-)

yanivagman (approved):

PR #337, opened 2020-11-29
fix a couple of issues with the makefile

itaysk:
1. make clean tried to read from a file after deleting it
2. libbpf headers target was `libbpf/usr/include`, but some targets required `libbpf/usr/include/bpf`, and there's no recipe for this target. This happened to work so far because usually some other rule tried to build `libbpf/usr/include` before the rule that needed `/bpf` run. but if the `/bpf` is called directly before the parent rule is called, which happens with `make release`, then it will try to build `libbpf/usr/include/bpf` which has no recipe. the solution is to change the libbpf headers target point to the specific desired path, (`/bpf`). another approach could be to specify exactly which headers files are needed whic I could do if you think it's better. another alternative is to define the `/bpf` as a target but since the parent dir is not used, I thought it's not better

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,11 +6,10 @@ RUN echo "deb http://apt.llvm.org/buster/ llvm-toolchain-buster-9 main" >> /etc/
     (for tool in "clang" "llc" "llvm-strip"; do path=$(which $tool-9) && ln -s $path ${path%-*}; done)
 WORKDIR /tracee
 
+FROM builder as build
 ARG VERSION
-FROM tracee-builder as build
-ENV VERSION=$VERSION
 COPY . /tracee
-RUN make build
+RUN make build VERSION=$VERSION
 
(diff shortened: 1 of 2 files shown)


yanivagman:
This change needs to be checked first on ubuntu 16 environment.
I added this changes in https://github.com/aquasecurity/tracee/commit/f35a8f393ea132322cb7077322e1060695f08d4b#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52
To fix a compilation issue in ubuntu16 (where the headers/uapi where needed as well as headers/bpf files)

itaysk:
since I don't want to dive into this now I changed the second commit to a quick fix instead. WDYT?

yanivagman:
LGTM,
please squash before merge

yanivagman (approved):

Issue #328, opened 2020-11-25
Add filter comparison operation options (>, <, !=) 


grantseltzer:
Now that the filter flag was merged in #315 it would be great to have more options besides '=' for matching on filters. For example:

--filter uid > 0

--filter uid != 1000

--filter return < 0

grantseltzer:
The way i'm thinking of going about this is as follows:

- Use the BPF_HASH `uid_filter` to store both the 'operator' and the values to compare the process uid to.
- `uid_filter[0]` contains the operator in first 3 bits (1 is '=', 2 is '!=', 3 is '>', 4 is '<') 
- If the operator is `=` or `!=` the `uid_filter` hash works the way it does now
  - To avoid collisions (i.e. `uid=0` is specified):
    - i'll `>> 3` the value in the hash before checking if the uid matches.
    - When checking operator i'll `&&` with FFFFFFF8 
- If the operator is `>` or `<` then `uid_filter[1]` contains the value to compare to.

I can also tune this to instead have operator values for multiple operators specified (i.e. uid<1000 and uid >5000) where uid_filter[2] contains the second value. I don't see too much usecase for supporting this but it couldn't hurt. 

Let me know if you have any thoughts.

yanivagman:
I don't know about using index 0 and 1 to store this metadata, as it requires special handling for these uids.
I understand that the previous apprach I suggested is problematic? (using the value of the uid to store ">", "<").
I guess this is because it may be hard to know where to find the ">" and "<" uids in the map, right? (and we want to avoid loops where possible).

Let me suggest a different approach.
We can use the uid_filter as we have today, and have 2 values (the key is the uid):
"1" will represent "="
"2" will represent "!="

Then we can have a different map for "<" and ">" , which will have a limited size (say 3).
Then, we can just "loop" over these 3 elements, and check if our given uid is bigger or smaller than, according to:
"1" that will represent "<"
"2" that will represent ">"

WDYT?

grantseltzer:
Sure, that is simpler. I can set the bpfConfigMap's 'configFilterByUid' field to the corresponding operator. 

Have to make sure all the combination of filters are supported:

`--filter uid=1000 --filter uid>5000` 
`--filter uid>0 --filter uid<1000`
`--filter uid<1000 --filter uid!=0` 
`--filter uid>0` --filter uid>1000` (>0 overwrites >1000)
`--filter uid<1000 --filter uid<500` (<1000 overwrites <500)
[...]

PR #339, opened 2020-12-01
Add the ability to specify filters (such as by UID) using comparison operators (=, !=, >, <).

grantseltzer:
- Add filter comparison operators (still just UIDs supported)
- Add tests in main_test.go

Relevant issue: #328

### DIFF
--- a/main.go
+++ b/main.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"io/ioutil"
 	"log"
+	"math"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -49,14 +50,10 @@ func main() {
 			if err != nil {
 				return err
 			}
-			filter, err := prepareFilter(c.StringSlice("filter"))
-			if err != nil {
-				return err
-			}
+
 			cfg := tracee.TraceeConfig{
 				EventsToTrace:         events,
 				Mode:                  mode,
-				Filter:                filter,
 				DetectOriginalSyscall: c.Bool("detect-original-syscall"),
 				ShowExecEnv:           c.Bool("show-exec-env"),
 				OutputFormat:          c.String("output"),
@@ -85,6 +82,12 @@ func main() {
 					return fmt.Errorf("invalid capture option: %s", cap)
 				}
 			}
+			filter, err := prepareFilter(c.StringSlice("filter"))
+			if err != nil {
+				return err
+			}
+			cfg.Filter = &filter
+
 			if c.Bool("security-alerts") {
 				cfg.EventsToTrace = append(cfg.EventsToTrace, tracee.MemProtAlertEventID)
 			}
@@ -224,49 +227,136 @@ func main() {
 
 func prepareFilter(filters []string) (tracee.Filter, error) {
 
-	uids := []uint32{}
-
 	filterHelp := "\n--filter allows you to specify values to match on for fields of traced events.\n"
 	filterHelp += "The following options are currently supported:\n"
 	filterHelp += "uid: only trace processes or containers with specified uid(s).\n"
 	filterHelp += "\t--filter uid=0                                                | only trace events from uid 0\n"
 	filterHelp += "\t--filter uid=0,1000                                           | only trace events from uid 0 or uid 1000\n"
 	filterHelp += "\t--filter uid=0 --filter uid=1000                              | only trace events from uid 0 or uid 1000 (same as above)\n"
+	filterHelp += "\t--filter 'uid>0'                                              | only trace events from uids greater than 0"
+	filterHelp += "\t--filter 'uid>0' --filter 'uid<1000'                          | only trace events from uids between 0 and 1000"
+	filterHelp += "\t--filter 'uid>0' --filter uid!=1000                           | only trace events from uids greater than 0 but not 1000"
 
 	if len(filters) == 1 && filters[0] == "help" {
 		return tracee.Filter{}, fmt.Errorf(filterHelp)
 	}
 
+	validFilterOptions := []string{"uid"}
+
+	filter := tracee.Filter{
+		UIDFilter: baseUIDFilter(),
+	}
+
 	for _, f := range filters {
-		s := strings.Split(f, "=")
-		if len(s) != 2 {
-			return tracee.Filter{}, fmt.Errorf(filterHelp)
+
+		// Parse the filter option (i.e. uid, ...)
+		isValid := false
+		filterSlice := []string{}
+		for i := range validFilterOptions {
+			if strings.HasPrefix(f, validFilterOptions[i]) {
+				filterSlice = strings.SplitAfter(f, validFilterOptions[i])
+				isValid = true
+				break
+			}
 		}
-		if !validFilterOption(s[0]) {
-			return tracee.Filter{}, fmt.Errorf("invalid filter: %s\n%s", s[0], filterHelp)
+		if !isValid {
+			return tracee.Filter{}, fmt.Errorf("invalid filter option specified, use '--filter help' for more info")
 		}
 
-		if s[0] == "uid" {
-			values := strings.Split(s[1], ",")
-			for _, v := range values {
-				uid, err := strconv.ParseUint(v, 10, 32)
-				if err != nil {
-					return tracee.Filter{}, fmt.Errorf("specified invalid uid: %s", v)
-				}
-				uids = append(uids, uint32(uid))
+		if filterSlice[0] == "uid" {
+			err := parseUIDFilter(filterSlice[1], filter.UIDFilter)
+			if err != nil {
+				return tracee.Filter{}, err
 			}
 		}
 	}
-	return tracee.Filter{
-		UIDs: uids,
-	}, nil
+
+	if len(filter.UIDFilter.Equal) > 0 &&
+		len(filter.UIDFilter.NotEqual) == 0 &&
+		filter.UIDFilter.Greater == -1 &&
+		filter.UIDFilter.Less == math.MaxUint32+1 {
+		filter.UIDFilter.Less = -1
+		filter.UIDFilter.Greater = math.MaxUint32 + 1
+	}
+
+	return filter, nil
 }
 
-func validFilterOption(s string) bool {
-	validOptions := map[string]bool{
-		"uid": true,
+// Operator represents a comparison operator such as '=', '!=', '>', or '<'
+type Operator uint8
+
+const (
+	opInvalid Operator = iota
+	opEqual
+	opNotEqual
+	opGreater
+	opLess
+)
+
+func parseUIDFilter(operatorAndValues string, uidFilter *tracee.UIDFilter) error {
+
+	valuesString := string(operatorAndValues[1:])
+	operatorString := string(operatorAndValues[0])
+
+	if operatorString == "!" {
+		operatorString = operatorAndValues[0:2]
+		valuesString = operatorAndValues[2:]
 	}
-	return validOptions[s]
+
+	operator := parseOperator(operatorString)
+	if operator == opInvalid {
+		return fmt.Errorf("invalid filter operator: %s", operatorString)
+	}
+
+	values := strings.Split(valuesString, ",")
+
+	for i := range values {
+		v, err := strconv.ParseUint(values[i], 10, 32)
+		if err != nil {
+			return fmt.Errorf("invalid UID value: %s", values[i])
+		}
+		if operator == opEqual {
+			uid := uint32(v)
+			uidFilter.Equal = append(uidFilter.Equal, uid)
+		}
+		if operator == opNotEqual {
+			uid := uint32(v)
+			uidFilter.NotEqual = append(uidFilter.NotEqual, uid)
+		}
+		uid := int64(v)
+		if operator == opGreater && uid > uidFilter.Greater {
+			uidFilter.Greater = uid
+		}
+		if operator == opLess && uid < uidFilter.Less {
+			uidFilter.Less = uid
+		}
+	}
+
+	return nil
+}
+
+func baseUIDFilter() *tracee.UIDFilter {
+	return &tracee.UIDFilter{
+		Equal:    []uint32{},
+		NotEqual: []uint32{},
+		Greater:  -1,
+		Less:     math.MaxUint32 + 1,
+	}
+}
+
+func parseOperator(operationString string) Operator {
+	switch operationString {
+	case "=":
+		return opEqual
+	case "!=":
+		return opNotEqual
+	case ">":
+		return opGreater
+	case "<":
+		return opLess
+	default:
+		return opInvalid
+	}
 }
 
 func prepareTraceMode(traceString string) (uint32, []int, error) {
--- a/main_test.go
+++ b/main_test.go
@@ -1,0 +1,182 @@
+package main
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/aquasecurity/tracee/tracee"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPrepareFilter(t *testing.T) {
+
+	testCases := []struct {
+		testName       string
+		filters        []string
+		expectedFilter tracee.Filter
+		expectedError  error
+	}{
+
+		{
+			testName:       "invalid operator",
+			filters:        []string{"uid==0"},
+			expectedFilter: tracee.Filter{},
+			expectedError:  errors.New("invalid UID value: =0"),
+		},
+		{
+			testName:       "invalid operator",
+			filters:        []string{"uid<=0"},
+			expectedFilter: tracee.Filter{},
+			expectedError:  errors.New("invalid UID value: =0"),
+		},
+		{
+			testName:       "invalid operator",
+			filters:        []string{"uid\t0"},
+			expectedFilter: tracee.Filter{},
+			expectedError:  errors.New("invalid filter operator: \t"),
+		},
+		{
+			testName:       "invalid filter type",
+			filters:        []string{"UID>0"},
+			expectedFilter: tracee.Filter{},
+			expectedError:  errors.New("invalid filter option specified, use '--filter help' for more info"),
+		},
+		{
+			testName:       "invalid filter type",
+			filters:        []string{"test=0"},
+			expectedFilter: tracee.Filter{},
+			expectedError:  errors.New("invalid filter option specified, use '--filter help' for more info"),
+		},
+		{
+			testName:       "invalid filter type",
+			filters:        []string{"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=0"},
+			expectedFilter: tracee.Filter{},
+			expectedError:  errors.New("invalid filter option specified, use '--filter help' for more info"),
+		},
+		{
+			testName:       "invalid uid",
+			filters:        []string{"uid>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>0"},
+			expectedFilter: tracee.Filter{},
+			expectedError:  errors.New("invalid UID value: >>>>>>>>>>>>>>>>>>>>>>>>>>>>>0"),
+		},
+		{
+			testName:       "invalid uid",
+			filters:        []string{"uid=a"},
+			expectedFilter: tracee.Filter{},
+			expectedError:  errors.New("invalid UID value: a"),
+		},
+		{
+			testName:       "invalid uid",
+			filters:        []string{"uid=4294967296"},
+			expectedFilter: tracee.Filter{},
+			expectedError:  errors.New("invalid UID value: 4294967296"),
+		},
+		{
+			testName:       "invalid uid",
+			filters:        []string{"uid=-1"},
+			expectedFilter: tracee.Filter{},
+			expectedError:  errors.New("invalid UID value: -1"),
+		},
+		{
+			testName:       "invalid uid",
+			filters:        []string{"uid=1 1"},
+			expectedFilter: tracee.Filter{},
+			expectedError:  errors.New("invalid UID value: 1 1"),
+		},
+		{
+			testName:       "invalid uid",
+			filters:        []string{"uid="},
+			expectedFilter: tracee.Filter{},
+			expectedError:  errors.New("invalid UID value: "),
+		},
+		{
+			testName: "invalid uid",
+			filters: []string{"uid=	"},
+			expectedFilter: tracee.Filter{},
+			expectedError: errors.New("invalid UID value: 	"),
+		},
+		{
+			testName:       "invalid uid",
+			filters:        []string{"uid=AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"},
+			expectedFilter: tracee.Filter{},
+			expectedError:  errors.New("invalid UID value: AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"),
+		},
+		{
+			testName:       "invalid uid",
+			filters:        []string{"uid=%s"},
+			expectedFilter: tracee.Filter{},
+			expectedError:  errors.New("invalid UID value: %s"),
+		},
+		{
+			testName:       "invalid uid",
+			filters:        []string{"uid=\t\t\t"},
+			expectedFilter: tracee.Filter{},
+			expectedError:  errors.New("invalid UID value: \t\t\t"),
+		},
+		{
+			testName:       "invalid uid",
+			filters:        []string{"uid=0\t"},
+			expectedFilter: tracee.Filter{},
+			expectedError:  errors.New("invalid UID value: 0\t"),
+		},
+		{
+			testName: "uid=0",
+			filters:  []string{"uid=0"},
+			expectedFilter: tracee.Filter{
+				UIDFilter: &tracee.UIDFilter{
+					Equal:    []uint32{0},
+					NotEqual: []uint32{},
+					Greater:  4294967296,
+					Less:     -1,
+				},
+			},
+			expectedError: nil,
+		},
+		{
+			testName: "uid!=0",
+			filters:  []string{"uid!=0"},
+			expectedFilter: tracee.Filter{
+				UIDFilter: &tracee.UIDFilter{
+					Equal:    []uint32{},
+					NotEqual: []uint32{0},
+					Greater:  -1,
+					Less:     4294967296,
+				},
+			},
+			expectedError: nil,
+		},
+		{
+			testName: "uid>0",
+			filters:  []string{"uid>0"},
+			expectedFilter: tracee.Filter{
+				UIDFilter: &tracee.UIDFilter{
+					Equal:    []uint32{},
+					NotEqual: []uint32{},
+					Greater:  0,
+					Less:     4294967296,
+				},
+			},
+			expectedError: nil,
+		},
+		{
+			testName: "uid<0",
+			filters:  []string{"uid<0"},
+			expectedFilter: tracee.Filter{
+				UIDFilter: &tracee.UIDFilter{
+					Equal:    []uint32{},
+					NotEqual: []uint32{},
+					Greater:  -1,
+					Less:     0,
+				},
+			},
+			expectedError: nil,
+		},
+	}
+	for _, testcase := range testCases {
+		t.Run(testcase.testName, func(t *testing.T) {
+			filter, err := prepareFilter(testcase.filters)
+			assert.Equal(t, testcase.expectedFilter.UIDFilter, filter.UIDFilter)
+			assert.Equal(t, testcase.expectedError, err)
+		})
+	}
+}
--- a/tracee/consts.go
+++ b/tracee/consts.go
@@ -11,7 +11,11 @@ const (
 	configCaptureFiles
 	configExtractDynCode
 	configTraceePid
-	configFilterByUid
+)
+
+const (
+	filterNotEqual int64 = iota
+	filterEqual
 )
 
 // an enum that specifies the index of a function to be used in a bpf tail call
--- a/tracee/tracee.bpf.c
+++ b/tracee/tracee.bpf.c
@@ -113,7 +113,6 @@
 #define CONFIG_CAPTURE_FILES    3
 #define CONFIG_EXTRACT_DYN_CODE 4
 #define CONFIG_TRACEE_PID       5
-#define CONFIG_FILTER_UID       6
 
 #define MODE_PROCESS_ALL        1
 #define MODE_PROCESS_NEW        2
@@ -122,6 +121,9 @@
 #define MODE_CONTAINER_NEW      5
 #define MODE_HOST_ALL           6
 #define MODE_HOST_NEW           7
+
+#define GREATER 0
+#define LESS    1
 
 #define DEV_NULL_STR    0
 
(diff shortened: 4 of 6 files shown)


tracee/tracee.go
```diff
@@ -429,10 +435,18 @@ func (t *Tracee) populateBPFMaps() error {
 		fileFilterMap.Update(uint32(i), []byte(t.config.FilterFileWrite[i]))
 	}
 
-	uidHash, _ := t.bpfModule.GetMap("uid_filter")
-	for i := 0; i < len(t.config.Filter.UIDs); i++ {
-		uidHash.Update(uint32(t.config.Filter.UIDs[i]), uint32(1))
+	// Set UID filter options
+	uidEqualityFilter, _ := t.bpfModule.GetMap("uid_equal_filter")
```
grantseltzer, line 439: I don't have the error checked here because it's not checked elsewhere in the file, but i'm curious why that is?

yanivagman: TL;DR - we should add those checks

When I wrote the libbpfgo, I wanted to keep the interface as close as possible to what we had with gobpf,
so it will be easy to understand the commit.
As gobpf doesn't return an error when using it's NewTable method (not sure about the name), I didn't handle it here. However, libbpfgo does return an error (for example, if the map doesn't exist).
This error should never happen with tracee, as we know these maps do exist in our bpf code, but it is always a good practice to check for errors

main_test.go
```diff
@@ -0,0 +1,182 @@
...
+		},
+		{
+			testName: "uid=0",
+			filters:  []string{"uid=0"},
+			expectedFilter: tracee.Filter{
+				UIDFilter: tracee.UIDFilter{
+					Equal:    []uint32{0},
+					NotEqual: []uint32{},
+					Greater:  4294967295,
+					Less:     0,
```
yanivagman, line 130: Shouldn't greater be 0 and less 4294967295?
I would expect that if no ">" or "<" were chosen, that these will have the default value (so they will always return true).
See my comment above about default values

main_test.go
```diff
@@ -0,0 +1,182 @@
...
+		},
+		{
+			testName: "uid>0",
+			filters:  []string{"uid>0"},
+			expectedFilter: tracee.Filter{
+				UIDFilter: tracee.UIDFilter{
+					Equal:    []uint32{},
+					NotEqual: []uint32{},
+					Greater:  0,
+					Less:     0,
```
yanivagman, line 156: shouldn't 'Less' have the 4294967295 default value as it is not set by this filter?

main_test.go
```diff
@@ -0,0 +1,182 @@
...
+			expectedError: nil,
+		},
+		{
+			testName: "uid<0",
+			filters:  []string{"uid<0"},
+			expectedFilter: tracee.Filter{
+				UIDFilter: tracee.UIDFilter{
+					Equal:    []uint32{},
+					NotEqual: []uint32{},
+					Greater:  4294967295,
```
yanivagman, line 168: Shouldn't greater have the '0' default value as it is not set?

main.go
```diff
@@ -224,49 +227,149 @@ func main() {
...
-		if len(s) != 2 {
-			return tracee.Filter{}, fmt.Errorf(filterHelp)
+
+		// Parse the filter option (i.e. uid, ...)
+		isValid := false
+		filterSlice := []string{}
+		for i := range validFilterOptions {
+			if strings.HasPrefix(f, validFilterOptions[i]) {
+				filterSlice = strings.SplitAfter(f, validFilterOptions[i])
+				isValid = true
```
yanivagman, line 258: You can break after this statement

main.go
```diff
@@ -224,49 +227,149 @@ func main() {
...
+		}
+		if operator == opLess && uid > uidFilter.Less {
+			uidFilter.Less = uid
+		}
+	}
+
+	return nil
+}
+
+func appendDefaultUIDs(uidFilter *tracee.UIDFilter) {
```
yanivagman, line 334: This function can be removed if we set default values that are always true.
This can be done if we choose our map value to be "int64" instead of "uint32".
Then, we can set UIDFilter.Greater to "-1" - which will always evaluate to true,
and UIDFilter.Less to math.MaxUint32 - which will also always be true.

main.go
```diff
@@ -224,49 +227,149 @@ func main() {
...
+	opNotEqual
+	opGreater
+	opLess
+)
+
+func parseUIDFilter(operatorAndValues string, uidFilter *tracee.UIDFilter) error {
+	validFilterOperators := []string{"=", "!=", ">", "<"}
+	isValid := false
+	valueSlice := []string{}
+	for i := range validFilterOperators {
```
yanivagman, line 293: You can remove this loop and above variables as you already check for this with parseOperator() below.
You can give it the first character of operatorAndValues as an argument  instead of slicing here

main.go
```diff
@@ -224,49 +227,149 @@ func main() {
...
+		uidFilter.Less = math.MaxUint32
+	}
+}
+
+func baseUIDFilter() tracee.UIDFilter {
+	return tracee.UIDFilter{
+		Equal:    []uint32{},
+		NotEqual: []uint32{},
+		Greater:  math.MaxUint32,
+		Less:     0,
```
yanivagman, line 356: These default values can be changes to be always true as suggested above

tracee/tracee.bpf.c
```diff
@@ -233,7 +233,8 @@ BPF_HASH(chosen_events_map, u32, u32);              // Events chosen by the user
 BPF_HASH(pids_map, u32, u32);                       // Save container pid namespaces
 BPF_HASH(args_map, u64, args_t);                    // Persist args info between function entry and return
 BPF_HASH(ret_map, u64, u64);                        // Persist return value to be used in tail calls
-BPF_HASH(uid_filter, u32, u32);                     // Used to filter events by UID
+BPF_HASH(uid_equal_filter, u32, u32);               // Used to filter events by UID, for specific UIDs either by == or !=
+BPF_HASH(uid_compare_filter, u32, u32);             // Used to filter events by UID, for ranges of UIDs by < or >
```
yanivagman, line 237: we can use value of s64 as suggested above

yanivagman (changes requested):

tracee/tracee.bpf.c
```diff
@@ -497,15 +498,26 @@ static __always_inline int get_config(u32 key)
     return *config;
 }
 
+// returns 1 if you should trace based on uid
```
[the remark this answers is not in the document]
grantseltzer: This made me realize a flaw in this filter code. The filter was always loaded into the bpf map regardless of if the user specified one, so the -1 shouldn't ever be reached. The commit I'm about to push includes your feedback from my initial commit but also redesigns some things to, I think, be more straightforward.

tracee/tracee.bpf.c
```diff
@@ -497,15 +498,26 @@ static __always_inline int get_config(u32 key)
...
 static __always_inline int uid_filter_matches()
 {
     uid_t uid = (u32)bpf_get_current_uid_gid();
-    u32* isMatch = bpf_map_lookup_elem(&uid_filter, &uid);
-    if (isMatch == NULL) {
-        return 0;
+    u32* equality = bpf_map_lookup_elem(&uid_equal_filter, &uid);
+    if (equality != NULL) {
+        return *equality;
```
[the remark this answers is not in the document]
grantseltzer: Why do you want to have that not trace anything? Would you expect `--filter uid=997,1000` to not produce any events? Would filters that just include '==' be OR'ed and otherwise AND'ed? 

That said I see the flaw in my current solution. Right now if someone were to specify `--filter 'uid>1' --filter 'uid<1000'` it wouldn't filter anything out. There are edge cases I appear to have overlooked.

For example, if a `!=` filter is specified every other UID should be included.

main.go
grantseltzer: Ah this is a good idea

tracee/tracee.bpf.c
```diff
@@ -497,15 +498,26 @@ static __always_inline int get_config(u32 key)
...
+    u32 zero = 0;
+    u32 one = 1;
+
+    u32* greaterThan = bpf_map_lookup_elem(&uid_compare_filter, &zero);
+    u32* lessThan = bpf_map_lookup_elem(&uid_compare_filter, &one);
+    if (greaterThan == NULL || lessThan == NULL) {
+        return -1;
+    }
+
+    return (uid > *greaterThan) || (uid < *lessThan);
```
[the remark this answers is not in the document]
grantseltzer: See my response to your comment on `tracee.bpf.c`:`507`

tracee/tracee.bpf.c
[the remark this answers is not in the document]
yanivagman: > I think a user would expect that tracee --filter 'uid>5' --filter 'uid=1' would be logical ORed. I think a user would also expect tracee --filter uid>1 --filter uid<1000 to be logical ANDed. If you disagree, do we even want to trouble ourselves with supporting multiple filter specifications?

I agree. This is more intuitive. It just makes it a bit harder to implement - mixing AND and OR together.
Do you want to give it another try or maybe we should decide if we want to AND everything or OR everything? I think it will make the implementation easier

grantseltzer:
@yanivagman Alright I think I've got the right idea now. 

Thinking about the UIDs specified with each operator as sets, the boolean equation (lives in bpf) for deciding if an event should be traced is as follows:

`(=) || (> && < && !(!=))`

Here's some examples:

User specified: `--filter uid=5,6 uid>500 uid!=1000`
User expects: events with uid 5,6 and over 500 are allowed, uids 0-5, 6-500, 1000 are not allowed

So if the UID of the event in bpf is represented by x and the sets are as follows:

`=`: [5, 6]
`>`: [500]
`<`: [u32_max + 1]
`!=`: [1000]

All the following would evaluate false (don't trace) x = 10, x = 3, x = 1000, x = 499, x = 500
All the following would evaluate true (trace) x = 5, x = 6, x = 999, x = 999999, x = 501

The couple important design elements are:

- We use int64 for the UIDs so we can specify the less than value as one over uint32 max and the greater than value as -1.
- `!=` evaluates true for what elements are in the set, so that's why it's negated in the above boolean logic.

I wrote out a ton of notes of examples that I believe this works how users would expect it to. Alternatively if you want to keep it as simple as possible i'd just recommend having support for one filter operation at a time. If it sounds good i'm just going to start from a fresh patch to clean up the code. Let me know what you think!

yanivagman:
@grantseltzer that makes sense.
I think that most of your current code is good.
After setting the defaults in main.go according to the comments, you can make the changes in the bpf code as you suggested.

grantseltzer:
Just pushed a commit with the update I described above, I still need to update the tests, should get to that in a few hours.

tracee/consts.go
```diff
@@ -1565,3 +1571,8 @@ var EventsIDToParams = map[int32][]param{
 	MemProtAlertEventID:        {{pType: "alert_t", pName: "alert"}},
 	SchedProcessExitEventID:    {},
 }
+
+const (
+	FilterUIDCeiling = int64(math.MaxUint32 + 1)
+	FilterUIDFloor   = int64(-1)
```
grantseltzer, line 1577: If this isn't API you want to export I can maneuver checking of the values into the BPF code.

tracee/tracee.bpf.c
```diff
@@ -497,14 +500,39 @@ static __always_inline int get_config(u32 key)
...
+        equal = 0;
+        not_equal = 0;
+    } else {
+        not_equal = *equality == 0;
+        equal = *equality == 1;
     }
-    return 1;
+
+    u32 greater = ZERO;
+    u32 less = ONE;
```
grantseltzer, line 521: Was this what you meant by using `#define` constants?

tracee/consts.go
yanivagman: I don't think we need these consts.
But what do you mean maneuver checking the values in the BPF code?

tracee/tracee.bpf.c
```diff
@@ -233,8 +235,8 @@ BPF_HASH(chosen_events_map, u32, u32);              // Events chosen by the user
 BPF_HASH(pids_map, u32, u32);                       // Save container pid namespaces
 BPF_HASH(args_map, u64, args_t);                    // Persist args info between function entry and return
 BPF_HASH(ret_map, u64, u64);                        // Persist return value to be used in tail calls
-BPF_HASH(uid_equal_filter, u32, u32);               // Used to filter events by UID, for specific UIDs either by == or !=
-BPF_HASH(uid_compare_filter, u32, u32);             // Used to filter events by UID, for ranges of UIDs by < or >
+BPF_HASH(uid_equal_filter, u32, int64_t);             // Used to filter events by UID, for specific UIDs either by == or !=
```
yanivagman, line 238: Do we need this to be 64 bits long or can we keep it 32?

tracee/tracee.bpf.c
```diff
@@ -233,8 +235,8 @@ BPF_HASH(chosen_events_map, u32, u32);              // Events chosen by the user
...
+BPF_HASH(uid_compare_filter, u32, int64_t);           // Used to filter events by UID, for ranges of UIDs by < or >
```
yanivagman, line 239: I think you can use s64 to keep types definitions in the same convention

tracee/tracee.bpf.c
yanivagman: I would give it a more meaningful name, such as GREATER and LESS.
The point is to use the same names (but in camel case) in the go code, as the maps are shared

tracee/tracee.bpf.c
```diff
@@ -498,26 +500,40 @@ static __always_inline int get_config(u32 key)
     return *config;
 }
 
-// returns 1 if you should trace based on uid
+// returns 1 if you should trace based on uid, 0 if not
 static __always_inline int uid_filter_matches()
 {
+
+    int equal;
+    int not_equal;
```
yanivagman, line 508: Do we really need to have both of these vars?
Seems redundant to me

tracee/consts.go
grantseltzer: I meant that we could check for default values in BPF instead of NULL. In that case the map would be updated regardless of if they were set in Go code. I think this is fine as is, got rid of the consts though.

tracee/tracee.bpf.c
grantseltzer: Ah, ok. To be honest i'm not that familiar with C type differences like this example, something to learn more about.

tracee/tracee.bpf.c
grantseltzer: They need to be signed 64 bits because they can be -1 or u32_max +1

grantseltzer: Hm, actually I put validation in the CLI portion of the code to make the user specify valid UIDs. Think it's worth it to support -1 and u32_max+1? Otherwise there _shouldn't_ ever be none u32 values here.

tracee/tracee.bpf.c
grantseltzer: Thinking about how I can make it more elegant, but the reason is because `NotEqual` does not mean `Equal`  was not specified

main.go
yanivagman: +1

main.go
yanivagman: +1

main.go
```diff
@@ -224,49 +228,132 @@ func main() {
...
+		if operator == opEqual {
+			uid := uint32(v)
+			uidFilter.Equal = append(uidFilter.Equal, uid)
+		}
+		if operator == opNotEqual {
+			uid := uint32(v)
+			uidFilter.NotEqual = append(uidFilter.NotEqual, uid)
+		}
+		uid := int64(v)
+		if operator == opGreater && uid < uidFilter.Greater {
```
yanivagman, line 324: Shouldn't this be the opposite? "uid > uidFilter.Greater"?

If we have "uid>5" + "uid>10" Don't we want to choose "uid>10"?

main.go
```diff
@@ -224,49 +228,132 @@ func main() {
...
+	}
+
+	return nil
+}
+
+func baseUIDFilter() *tracee.UIDFilter {
+	return &tracee.UIDFilter{
+		Equal:    []uint32{},
+		NotEqual: []uint32{},
+		Greater:  int64(math.MaxUint32 + 1),
```
yanivagman, line 339: Don't we need Greater to be equal to "-1-" and Less be equal to (math.MaxUint32 + 1) so defaults will always evaluate to true?

tracee/tracee.bpf.c
yanivagman: Right, but in this case, you have a given uid, which only has one value in uid_equal_filter map.
If equal=1, then not_equal=0 and the other way around, right?
If so - there is no new information in the second variable.
Actually, I think non of these variables are needed, and let me explain.
Thinking about the formula you came up with:  (=) || (> && < && !(!=))
If uid value in uid_equal_filter is not NULL, we can return right away with the correct result.

tracee/tracee.bpf.c
```diff
@@ -497,14 +500,39 @@ static __always_inline int get_config(u32 key)
...
     }
-    return 1;
+
+    u32 greater = GREATER;
+    u32 less = LESS;
+
+    s64* greaterThan = bpf_map_lookup_elem(&uid_compare_filter, &greater);
+    s64* lessThan = bpf_map_lookup_elem(&uid_compare_filter, &less);
+
+    int accumulator = !not_equal;
```
yanivagman, line 526: I don't think this variable is needed as well, according to my previous and next comments

tracee/tracee.bpf.c
```diff
@@ -497,14 +500,39 @@ static __always_inline int get_config(u32 key)
...
+
+    if (lessThan != NULL) {  
+        accumulator = accumulator && (uid < *lessThan);
```
yanivagman, line 529: At this point, we already tested for equality, and returned if necessary.
We can change this statement to:
if ((uid >= *lessThan)) return 0;

tracee/tracee.bpf.c
```diff
@@ -497,14 +500,39 @@ static __always_inline int get_config(u32 key)
...
+    }
+    if (greaterThan != NULL ) { 
+        accumulator = accumulator && (uid > *greaterThan);
```
yanivagman, line 532: same:
if ((uid <= *greaterThan)) return 0;

tracee/tracee.bpf.c
```diff
@@ -497,14 +500,39 @@ static __always_inline int get_config(u32 key)
...
+    }
+
+    return (equal || accumulator);
```
yanivagman, line 535: And as we already tested for all cases and everything is ok:
return 1;

main.go
grantseltzer: Hm, I was thinking we would want to choose "uid>5", I figured if there's conflicting filters go for the more inclusive one.

tracee/tracee.bpf.c
grantseltzer: What a ">" and a ""<" filter are specified? (i.e. `uid>0` `uid<1000`)

main.go
grantseltzer: This relates to my comment above. I have it set to to closed by default so when reading in user-supplied filter options I can compare based on it being lower. When loading into BPF it compares to this default value and won't load into the BPF map at all if it hasn't changed. I can swap it around if you'd like.

main.go
yanivagman: I think it is more intuitive to take the more strict filter, but if you think otherwise, that's ok, we can leave it like this.

main.go
yanivagman: If that works, and you think it's more intuitive, I don't mind keeping it this way

tracee/tracee.bpf.c
grantseltzer: So I find it confusing, but the issue with this is that if a uid doesn't have any matching filters, the event will be allowed. But in the case of lets say:

User specifies: `--filter uid=0` (user only wants events with uid=0)

If lets say an event has a uid 1000 it would pass ('equality', 'lessThan', 'greaterThan' all NULL). If we return `0` at the end it breaks the comparison filters.

Let me see how I can clean it up...

yanivagman: Good point, but I think it has an easy fix.
To avoid missing cases, I tried to rewrite this function:
static __always_inline int uid_filter_matches()
{
    uid_t uid = (u32)bpf_get_current_uid_gid();
    s64* equality = bpf_map_lookup_elem(&uid_equal_filter, &uid);
    if (equality != NULL) {
        return *equality;
    }

    u32 greater = GREATER;
    u32 less = LESS;

    s64* greaterThan = bpf_map_lookup_elem(&uid_compare_filter, &greater);
    s64* lessThan = bpf_map_lookup_elem(&uid_compare_filter, &less);

    if (lessThan != NULL) {
        if ((uid >= *lessThan)) 
            return 0;
    }

    if (greaterThan != NULL ) { 
        if ((uid <= *greaterThan)) 
            return 0;
    }

    if (greaterThan == NULL && lessThan == NULL)
        // If we reached this point, there is no filter match
        return 0;

    return 1;
}

If may have missed something, so please tell me what you think

grantseltzer: It looks good to me, but I'm being thrown for a serious loop by the verifier. When running your code I get:

```
257: (15) if r3 == 0x0 goto pc+3
 R0=map_value(id=0,off=0,ks=4,vs=8,imm=0) R1_w=inv(id=0) R2_w=inv(id=0,umax_value=4294967295,var_off=(0x0; 0xffffffff)) R3_w=map_value(id=0,off=0,ks=4,vs=8,imm=0) R6=invP0 R7=inv(id=0,umax_value=4294967295,var_off=(0x0; 0xffffffff)) R8=inv0 R9=inv(id=0,umax_value=4294967295,var_off=(0x0; 0xffffffff)) R10=fp0 fp-80=????mmmm fp-88=????mmmm fp-104=mmmmmmmm fp-112=mmmmmmmm fp-120=mmmmmmmm fp-128=mmmmmmmm fp-136=mmmmmmmm fp-144=mmmmmmmm fp-152=mmmmmmmm fp-160=mmmmmmmm fp-168=mmmmmmmm fp-176=mmmmmmmm fp-184=mmmmmmmm fp-192=mmmmmmmm fp-200=mmmmmmmm fp-208=mmmmmmmm fp-216=mmmmmmmm fp-224=mmmmmmmm fp-232=mmmmmmmm fp-240=mmmmmmmm fp-248=mmmmmmmm fp-256=mmmmmmmm fp-264=mmmmmmmm fp-272=mmmm???? fp-368=????mmmm fp-376=ctx fp-384=mmmmmmmm fp-392=map_value
; if ((greaterThan != NULL) && (uid <= *greaterThan)) { 
258: (79) r1 = *(u64 *)(r3 +0)
 R0=map_value(id=0,off=0,ks=4,vs=8,imm=0) R1_w=inv(id=0) R2_w=inv(id=0,umax_value=4294967295,var_off=(0x0; 0xffffffff)) R3_w=map_value(id=0,off=0,ks=4,vs=8,imm=0) R6=invP0 R7=inv(id=0,umax_value=4294967295,var_off=(0x0; 0xffffffff)) R8=inv0 R9=inv(id=0,umax_value=4294967295,var_off=(0x0; 0xffffffff)) R10=fp0 fp-80=????mmmm fp-88=????mmmm fp-104=mmmmmmmm fp-112=mmmmmmmm fp-120=mmmmmmmm fp-128=mmmmmmmm fp-136=mmmmmmmm fp-144=mmmmmmmm fp-152=mmmmmmmm fp-160=mmmmmmmm fp-168=mmmmmmmm fp-176=mmmmmmmm fp-184=mmmmmmmm fp-192=mmmmmmmm fp-200=mmmmmmmm fp-208=mmmmmmmm fp-216=mmmmmmmm fp-224=mmmmmmmm fp-232=mmmmmmmm fp-240=mmmmmmmm fp-248=mmmmmmmm fp-256=mmmmmmmm fp-264=mmmmmmmm fp-272=mmmm???? fp-368=????mmmm fp-376=ctx fp-384=mmmmmmmm fp-392=map_value
; if ((greaterThan != NULL) && (uid <= *greaterThan)) { 
259: (61) r2 = *(u32 *)(r10 -368)
; if ((greaterThan != NULL) && (uid <= *greaterThan)) { 
260: (7d) if r1 s>= r2 goto pc-64
 R0=map_value(id=0,off=0,ks=4,vs=8,imm=0) R1=inv(id=0) R2=inv(id=0,umax_value=4294967295,var_off=(0x0; 0xffffffff)) R3=map_value(id=0,off=0,ks=4,vs=8,imm=0) R6=invP0 R7=inv(id=0,umax_value=4294967295,var_off=(0x0; 0xffffffff)) R8=inv0 R9=inv(id=0,umax_value=4294967295,var_off=(0x0; 0xffffffff)) R10=fp0 fp-80=????mmmm fp-88=????mmmm fp-104=mmmmmmmm fp-112=mmmmmmmm fp-120=mmmmmmmm fp-128=mmmmmmmm fp-136=mmmmmmmm fp-144=mmmmmmmm fp-152=mmmmmmmm fp-160=mmmmmmmm fp-168=mmmmmmmm fp-176=mmmmmmmm fp-184=mmmmmmmm fp-192=mmmmmmmm fp-200=mmmmmmmm fp-208=mmmmmmmm fp-216=mmmmmmmm fp-224=mmmmmmmm fp-232=mmmmmmmm fp-240=mmmmmmmm fp-248=mmmmmmmm fp-256=mmmmmmmm fp-264=mmmmmmmm fp-272=mmmm???? fp-368=????mmmm fp-376=ctx fp-384=mmmmmmmm fp-392=map_value
; if ((greaterThan != NULL) && (uid <= *greaterThan)) { 
261: (4f) r3 |= r0
R3 pointer |= pointer prohibited
processed 966 insns (limit 1000000) max_states_per_insn 2 total_states 67 peak_states 67 mark_read 26
```

I suspected this has to do with checking `greaterThan` and `lessThan` after dereferencing the pointers, but I think I disproved that. My other theory was the third check (`(greaterThan == NULL && lessThan == NULL)`) quick fails if `greaterThan` is NULL so technically lessThan isn't checked but they're checked above regardless.

Is this happening when you run on your machine? I'm on mainline `5.9.11-200` kernel.

yanivagman: Just checked on my machine, and I don't get verifier problems.
Can you please try to break the if:

```
    if ((lessThan != NULL) {
        if (uid >= *lessThan))
            return 0;
    }

    if ((greaterThan != NULL) {
        if (uid <= *greaterThan))
            return 0;
    }
```

grantseltzer: First thing I tried, same issue...

grantseltzer: What's strange is that if I remove the third if statement I don't get the verifier issue, but if I add it back the error message has to do with the first if statement.

yanivagman: Which clang version do you use?
It seems to be related to the registers assignment (I think)

grantseltzer: I have 11.0.0 installed

yanivagman: I now also see a logical problem here.
If no uid filter was chosen, greaterThan and lessThan will be null, and will get no output event

grantseltzer: Yea I noticed that too. Locally I changed the Go code to have the base uid filter start open and to always Update the map in tracee.go.

yanivagman: Can you please push your latest changes so I can test them locally?

grantseltzer: Done

tracee/tracee.go
```diff
@@ -429,9 +435,39 @@ func (t *Tracee) populateBPFMaps() error {
 		fileFilterMap.Update(uint32(i), []byte(t.config.FilterFileWrite[i]))
 	}
 
-	uidHash, _ := t.bpfModule.GetMap("uid_filter")
-	for i := 0; i < len(t.config.Filter.UIDs); i++ {
-		uidHash.Update(uint32(t.config.Filter.UIDs[i]), uint32(1))
+	// Set filter options
+	if t.config.Filter != nil {
```
yanivagman, line 439: If no filter chosen, no events will show as uidCompareFilter defaults are not set

grantseltzer: Right - didn't get that far to test because I'm in verifier hell, are you have the same issue if you've updated your kernel?

tracee/tracee.bpf.c
```diff
@@ -497,13 +500,35 @@ static __always_inline int get_config(u32 key)
...
+        if (uid > *lessThan)
+            return 0;
+    }
+
+    if (greaterThan != NULL) {
+        if (uid < *greaterThan)
+            return 0;
+    }
+
+    if (greaterThan == NULL && lessThan == NULL) {
```
yanivagman, line 528: Ok, so I reproduced the verifier issue with kernel 5.4.
For some reason, it doesn't like this line.
Then I thought - maybe we can remove this last if.

But then we come back to the problem you said:
if uid=5 was given as a filter, and now we have an event with uid=7, then it should not pass the filter.
The problem is, that in this case, we want the defaults to be "close" for inequality.

So what I suggest is the following:
1. Remove this if and pass the verifier
2. If ONLY equality filter was given (no greater than, no less than, no !=) - default to close values
3. Otherwise, default to open values.

This is logic, I think, because it should satisfy the equation you wrote:
 (=) || ((<) && (>) && !(!=))

WDYT?

tracee/tracee.go
yanivagman: Verifier hell! I know the feeling....

See my suggestion in the other comment

grantseltzer: I'm really looking forward to one day understanding the nuances of the verifier, just wish I can figure out this particular issue so I can at least learn from it :upside_down_face:

tracee/tracee.bpf.c
grantseltzer: Glad you were able to reproduce. Yes that logic makes sense to me, working on implementing that and taking into account all your previous comments. I'm surprised this problem become so complex to be honest.

yanivagman: I'm also surprised about that. But I think two things made it complex:
1. A comment I wrote few days ago:
"This is more intuitive. It just makes it a bit harder to implement - mixing AND and OR together..."
2. When writing bpf code, you can never know where the verifier will block you (unless you are one of the developers of the verifier in the kernel)

grantseltzer: Codifying user intent always can be tough. There's a lot of work to get done for CI/CD, testing against various kernel/verifier versions. That should simplify catching future issues.

grantseltzer:
Alright I think I've got it, working for all manual tests on my end. Let me know what ya think!

tracee/tracee.bpf.c
```diff
@@ -497,13 +500,31 @@ static __always_inline int get_config(u32 key)
...
+        return *equality;
+    }
+
+    u32 greater = GREATER;
+    u32 less = LESS;
+
+    s64* greaterThan = bpf_map_lookup_elem(&uid_compare_filter, &greater);
+    s64* lessThan = bpf_map_lookup_elem(&uid_compare_filter, &less);
+
+    if (lessThan != NULL) {
```
yanivagman, line 518: Just for aesthetic, I think we can have one "if" here (as there is no verifier issue with this):
if ((lessThan != NULL) && (uid >= *lessThan))

tracee/tracee.bpf.c
```diff
@@ -497,13 +500,31 @@ static __always_inline int get_config(u32 key)
...
+        if (uid >= *lessThan)
+            return 0;
     }
+
+    if (greaterThan != NULL) {
```
yanivagman, line 523: same

main.go
yanivagman: What about this one?

main.go
yanivagman: And this one?

yanivagman (commented):
@grantseltzer that's great!

There are two comments of mine in main.go which I'm not sure you've seen

yanivagman:
Thanks @grantseltzer !
LGTM

Can you please squash the changes to one commit?

grantseltzer:
Thanks for all the feedback and patience working this out, squashed and ready to go.

Good luck at Black Hat!

yanivagman (approved):

PR #341, opened 2020-12-01
Use alpine image instead of ubuntu

### LINKED ISSUE #329: reduce the size of docker images
currently, the latest image weights 654MB we need to to reduce it. This is is mostly due to the `apt-get install` layer, which is 574.5MB alone, and confirmed by looking at the slim image which is 107MB.

### LINKED ISSUE #335: Use alpine docker image instead of ubuntu
Moving to alpine will reduce ~50MB (non compressed) by installing the following packages for "fat" image:

FROM golang:alpine
RUN apk update
RUN apk add clang llvm make gcc libc6-compat coreutils linux-headers musl-dev elfutils-dev libelf-static zlib-static
...

And for "slim" image, will reduce the image size to less than 10MB (non compressed):
FROM alpine
RUN apk update
RUN apk add libc6-compat elfutils-dev
...

Other advantages of moving to alpine:
Faster CI times (as apk update and apk add are much faster than apt parallels),
Smaller attack surface

If we choose to go in this path, we should also statically link tracee as alpine uses musl (standalone tracee binary compiled using alpine will have musl dependency).
Statically linking the code has its own advantages, as it will remove the libc, libelf and zlib dependencies.
Another security advantage if statically linking is that the binary then can't be preloaded

yanivagman:
Use docker alpine image instead of ubuntu.
This change reduces the attack surface and shrinks the image sizes.
Image sizes (before compression in dockerhub):
- Fat: 327MB
- Slim: 15MB

Close #335 
Close #329

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,7 @@
 ARG BASE=fat
 
-FROM golang:1.15-buster as builder
-RUN echo "deb http://apt.llvm.org/buster/ llvm-toolchain-buster-9 main" >> /etc/apt/sources.list && apt-key adv --keyserver hkps://keyserver.ubuntu.com --recv-keys 15CF4D18AF4F7421 && \
-    DEBIAN_FRONTEND=noninteractive apt-get update -y && apt-get install -y --no-install-recommends libelf-dev llvm-9-dev clang-9 && \ 
-    (for tool in "clang" "llc" "llvm-strip"; do path=$(which $tool-9) && ln -s $path ${path%-*}; done)
+FROM golang:alpine as builder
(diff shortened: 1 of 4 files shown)


yanivagman:
I'm putting this on draft while considering if static linking is the right thing to do.
While it removes the library dependencies, cgo is not meant to be statically linked

itaysk (commented):
you can further optimize the image size by adding `--no-cache` to apk

itaysk (commented):
now the binary in the container is built with a different version of LLVM than the binary for release. the easy fix for you will be to use llvm10 in the builder.

yanivagman:
I don't think it's a problem that the clang versions are different.
The compiler is just a tool used to compile

itaysk (approved):

PR #345, opened 2020-12-03
Split kernel headers to source and build

### LINKED ISSUE #342: Failed to run tracee on Kali linux
Kali linux (and other ditros?) has it's headers files seperated between /lib/modules/KERN_RELEASE/build for headers generated during build and /lib/modules/KERN_RELEASE/source for all other includes.
As a result, we fail to compile the bpf code as the headers are not where we expect them to be

yanivagman:
Close #342

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,8 @@ CMD_GITHUB ?= gh
 # environment:
 ARCH ?= $(shell uname -m)
 KERN_RELEASE ?= $(shell uname -r)
-KERN_SRC ?= /lib/modules/$(KERN_RELEASE)/build
+KERN_BLD_PATH ?= $(if $(KERN_HEADERS),$(KERN_HEADERS),/lib/modules/$(KERN_RELEASE)/build)
+KERN_SRC_PATH ?= $(if $(KERN_HEADERS),$(KERN_HEADERS),$(if $(wildcard /lib/modules/$(KERN_RELEASE)/source),/lib/modules/$(KERN_RELEASE)/source,$(KERN_BLD_PATH)))
 VERSION ?= $(if $(RELEASE_TAG),$(RELEASE_TAG),$(shell $(CMD_GIT) describe --tags))
 # inputs and outputs:
 OUT_DIR ?= dist
@@ -27,8 +28,9 @@ LIBBPF_HEADERS := $(OUT_DIR)/libbpf/usr/include
 LIBBPF_OBJ := $(OUT_DIR)/libbpf/libbpf.a
 OUT_DOCKER ?= tracee
 DOCKER_BUILDER ?= tracee-builder
-# DOCKER_BUILDER_KERN_SRC is where the docker builder looks for kernel headers
-DOCKER_BUILDER_KERN_SRC ?= $(if $(shell readlink $(KERN_SRC)),$(shell readlink $(KERN_SRC)),$(KERN_SRC))
+# DOCKER_BUILDER_KERN_SRC(/BLD) is where the docker builder looks for kernel headers
+DOCKER_BUILDER_KERN_BLD ?= $(if $(shell readlink $(KERN_BLD_PATH)),$(shell readlink $(KERN_BLD_PATH)),$(KERN_BLD_PATH))
+DOCKER_BUILDER_KERN_SRC ?= $(if $(shell readlink $(KERN_SRC_PATH)),$(shell readlink $(KERN_SRC_PATH)),$(KERN_SRC_PATH))
 # DOCKER_BUILDER_KERN_SRC_MNT is the kernel headers directory to mount into the docker builder container. DOCKER_BUILDER_KERN_SRC should usually be a decendent of this path.
 DOCKER_BUILDER_KERN_SRC_MNT ?= $(dir $(DOCKER_BUILDER_KERN_SRC))
 RELEASE_ARCHIVE := $(OUT_DIR)/tracee.tar.gz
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -12,7 +12,11 @@ CMD_GITHUB ?= gh
 # environment:
 ARCH ?= $(shell uname -m)
 KERN_RELEASE ?= $(shell uname -r)
-KERN_SRC ?= /lib/modules/$(KERN_RELEASE)/build
+KERN_BLD_PATH ?= $(if $(KERN_SRC),$(KERN_SRC),/lib/modules/$(KERN_RELEASE)/build)
+KERN_SRC_PATH ?= $(if $(KERN_SRC),$(KERN_SRC),/lib/modules/$(KERN_RELEASE)/source)
```
itaysk, line 16: why did we keep both KERN_SRC and KERN_SRC_PATH? don't they mean the same thing?

Makefile
```diff
@@ -12,7 +12,11 @@ CMD_GITHUB ?= gh
 # environment:
 ARCH ?= $(shell uname -m)
 KERN_RELEASE ?= $(shell uname -r)
-KERN_SRC ?= /lib/modules/$(KERN_RELEASE)/build
+KERN_BLD_PATH ?= $(if $(KERN_SRC),$(KERN_SRC),/lib/modules/$(KERN_RELEASE)/build)
```
itaysk, line 15: KERN_SRC is used to provide the value for KERN_BLD_PATH? what's the reasoning behind this?

Makefile
```diff
@@ -12,7 +12,11 @@ CMD_GITHUB ?= gh
...
+KERN_SRC_PATH ?= $(if $(KERN_SRC),$(KERN_SRC),/lib/modules/$(KERN_RELEASE)/source)
+ifeq ($(wildcard $(KERN_SRC_PATH)/.),)
+  KERN_SRC_PATH := $(KERN_BLD_PATH)
```
itaysk, line 18: There's a potential bug here, if the user sets KERN_SRC_PATH and it's empty, this assignment won't override it.

Makefile
```diff
@@ -12,7 +12,11 @@ CMD_GITHUB ?= gh
 # environment:
 ARCH ?= $(shell uname -m)
 KERN_RELEASE ?= $(shell uname -r)
-KERN_SRC ?= /lib/modules/$(KERN_RELEASE)/build
+KERN_BLD_PATH ?= $(if $(KERN_SRC),$(KERN_SRC),/lib/modules/$(KERN_RELEASE)/build)
+KERN_SRC_PATH ?= $(if $(KERN_SRC),$(KERN_SRC),/lib/modules/$(KERN_RELEASE)/source)
+ifeq ($(wildcard $(KERN_SRC_PATH)/.),)
```
itaysk, line 17: maybe instead of enumerating all the files just `test -d`?

Makefile
```diff
@@ -138,7 +144,10 @@ $(OUT_DIR)/$(DOCKER_BUILDER): $(GO_SRC) $(BPF_SRC) $(MAKEFILE_LIST) Dockerfile |
 
 # docker_builder_make runs a make command in the tracee-builder container
 define docker_builder_make
-	$(CMD_DOCKER) run --rm -v $(abspath $(DOCKER_BUILDER_KERN_SRC_MNT)):$(DOCKER_BUILDER_KERN_SRC_MNT) -v $(abspath .):/tracee --entrypoint make $(DOCKER_BUILDER) KERN_SRC=$(DOCKER_BUILDER_KERN_SRC) $(1)
+	$(CMD_DOCKER) run --rm \
+	-v $(abspath $(DOCKER_BUILDER_KERN_SRC_MNT)):$(DOCKER_BUILDER_KERN_SRC_MNT) \
+	-v $(abspath .):/tracee \
+	--entrypoint make $(DOCKER_BUILDER) KERN_BLD_PATH=$(DOCKER_BUILDER_KERN_BLD) KERN_SRC_PATH=$(DOCKER_BUILDER_KERN_SRC) $(1)
```
itaysk, line 150: when adding line breaks, there are cases where this might not work properly, like when this function is called form a substitute. we shouldn't have any of those so it's not a big deal

Makefile
yanivagman: There are two "flavors" of kernel headers, according to the distro.
One flavor is using one directory for all kernel headers, including generated. In this case KERN_SRC should be chosen.
The other flavor is when the headers are split into two directories - source and build.
If the user explicitly chose KERN_SRC, it should be used both for KERN_BLD_PATH and KERN_SRC_PATH, and override the automatic search for build and source paths.

Makefile
yanivagman: They don't. See my above comment for the meaning of KERN_SRC.
KERN_SRC_PATH is the path where the includes that are not generated from the build are located.
KERN_BUILD_PATH is the path where the includes that are generated from the build are located.

On some distros, these two will be on the same directory, while on others, they are splitted

Makefile
yanivagman: So what do you suggest? If it's not a big deal should we keep it with line breaks? (otherwise line is really long)

Makefile
yanivagman: I understand that this line checks if the directory exists, isn't it?
I'll try with "test -d" if this is better

Makefile
yanivagman: Not sure I understand the bug.
KERN_SRC_PATH was already set in line 16.
We override it here only if KERN_SRC_PATH doesn't exist

Makefile
itaysk: My misunderstanding was due to the names - KERN_SRC is not really the SOURCE rather the BUILD dir. This "fault" was also before but now it's standing out because we added separate vars for SOURCE and BUILD. 
Maybe instead of asking the user for the location of /build and /source we ask them for the location of /lib/modules/$version and we find /build or /source there?
From `linux/Makefile`:

```
MODLIB = $(INSTALL_MOD_PATH)/lib/modules/$(KERNELRELEASE)
...
ln -s $(abspath $(srctree)) $(MODLIB)/source
...
ln -s $(CURDIR) $(MODLIB)/build
```

Makefile
itaysk: I was worried that this override won't work:

```
$ cat Makefile
FOO ?= bar
FOO := baz

yo:
	@echo $(FOO)

$ make yo FOO=hi
hi
```

Makefile
yanivagman: This doesn't take into consideration the case that the user doesn't have "source" and "build" sub directories.
For example, if the user downloaded the sources, and gives it as a parameter to KERN_SRC.

I think the name KERN_SRC is ok, but we can change it to KERN_HDRS to be more clear.
Then we will have:
KERN_HDRS - that points to a directory with all headers included.
KERN_SRC_PATH, KERN_BLD_PATH - that points to the source and build directories repsectively

Makefile
itaysk: I have has this issue with the `release_docker_image` function, where adding line-breaks broke the caller, because of how it was being called. I guess either way is fine.

Makefile
itaysk: ok I understand. so isn't the KERN_SRC redundant? why don't we ask the user for the /source and the /build paths, and if the user doesn't give them we default, and if the user give just one we default the other one.

yanivagman: Changed to KERN_HEADERS as discussed

itaysk (approved):

PR #346, opened 2020-12-06
Print bpf cmd argument and make a default event

yanivagman:


tracee/argprinters.go
```diff
@@ -557,6 +557,57 @@ func PrintPtraceRequest(req int32) string {
 	return res
 }
 
+// PrintBpfCmd prints the `cmd` argument of the `bpf` syscall
+// https://man7.org/linux/man-pages/man2/bpf.2.html
+// https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/bpf.h
+func PrintBpfCmd(cmd int32) string {
```
itaysk, line 563: Bpf -> BPF
(didn't go vet complained about this?)

yanivagman: go vet didn't complained, and why should it?
What's the problem with using camel case Bpf?

itaysk: This is an infamous golang convention: https://github.com/golang/go/wiki/CodeReviewComments#initialisms
The rest of our codebase follows this convention

itaysk (approved):

PR #347, opened 2020-12-06
Add host only mode

### LINKED ISSUE #344: Add host tracing only mode
It might be useful to only trace the host, without the containers running on it. 
This can help reduce some noise when we only want to trace the environment running the containers, and not the containers themselves.

yanivagman:
Close #344


itaysk (approved):

itaysk (commented):
update the readme?

itaysk (approved):

PR #348, opened 2020-12-07
Add security_inode_unlink event

yanivagman:


itaysk (approved):
just out of curiosity, what are the get_d_..._from_dentry functions? I can't find them in linux?

yanivagman:
@itaysk these functions are part of tracee.bpf.c that I'm using to read kernel specific data structures

PR #351, opened 2020-12-08
Fix bpf compilation on redhat and centos with kernel 4.18

### LINKED ISSUE #349: Tracee doesn't work on CentOS 8
Tracee doesn't seem to be compatible with the stock kernel on CentOS 8.

Steps to reproduce:
1. Get a machine with CentOS 8.2 or 8.3, install latest updates and use the stock kernel.
2. Install prerequisites for Tracee.
3. Download a binary release of Tracee or build it from source.
4. Run it with debugging enabled.
5. Observe a failure to build the BPF program that Tracee uses.

Here is [the output](https://pastebin.com/raw/z8uYKBP3) of an attempt to run Tracee 0.3.0 on CentOS 8.3.2011 (Linux 4.18.0-240.1.1.el8_3.x86_64, clang 10.0.1). I've tried building Tracee from source (master) but it doesn't make any difference.

At the same time, BCC and bpftrace (latest versions built from source) work perfectly and do not produce any errors on the same machine.

yanivagman:
Fix #349

### DIFF
--- a/tracee/tracee.bpf.c
+++ b/tracee/tracee.bpf.c
@@ -156,6 +156,12 @@ BPF_MAP(_name, BPF_MAP_TYPE_PROG_ARRAY, u32, u32, _max_entries);
 #define BPF_PERF_OUTPUT(_name) \
 BPF_MAP(_name, BPF_MAP_TYPE_PERF_EVENT_ARRAY, int, __u32, 1024);
 
+#ifdef RHEL_RELEASE_CODE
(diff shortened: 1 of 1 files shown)


yanivagman:
Marked as draft as I still didn't check this fix on a relevant redhat/centos system

yanivagman:
Tested by @NeonSludge in #349

itaysk (approved):

PR #355, opened 2020-12-14
Add mnt ns and pid ns filters

yanivagman:


itaysk:
Nice!! will have to review it tomorrow cause I'm leaving early today. can't wait to try it

tracee/tracee.bpf.c
```diff
@@ -535,6 +542,40 @@ static __always_inline int uid_filter_matches()
...
+    struct task_struct *task;
+    task = (struct task_struct *)bpf_get_current_task();
+
+    u64 mnt_id = get_task_mnt_ns_id(task);
+    u8* equality = bpf_map_lookup_elem(&mnt_ns_filter, &mnt_id);
+    if (equality != NULL) {
+        return *equality;
+    }
+
+    if (get_config(CONFIG_FILTER_MNT_NS) == FILTER_IN)
```
itaysk, line 556: not sure why this is needed? (as well as related decelerations)

itaysk (approved):

tracee/tracee.bpf.c
yanivagman: We want to support both "=" and "!=" filters.
In case "!=" filters were chosen - you want to "filter out" all of the selected namespace ids. In this case, you want to trace (return 1) all other namespaces that were not selected.

In case "=" filters were chosen - you want to "filter in" all of the selected namespace ids. In this case, you don't want to trace (return 0) all other namespaces that were not selected

itaysk:
@yanivagman should have also updated `TestPrepareFilter` with this PR

PR #370, opened 2020-12-21
Set TRACEE_BPF_FILE to point to a file instead of a dir

### LINKED ISSUE #356: TRACEE_BPF_FILE should point to a file, not directory
currently TRACEE_BPF_FILE's value is expected to be a directory, which is not intuitive

yanivagman:
Close #356

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ This will run Tracee with no arguments, which defaults to collecting all events 
 
 ### Setup options
 
-Tracee is made of an executable that drives the eBPF program (`tracee`), and the eBPF program itself (`tracee.bpf.$kernelversion.$traceeversion.o`). When the `tracee` executable is started, it will look for the eBPF program next to the executable, or in `/tmp/tracee`, or in a directory specified in `TRACEE_BPF_FILE` environment variable. If the eBPF program is not found, the executable will attempt to build it automatically before it starts (you can control this using the `--build-policy` flag).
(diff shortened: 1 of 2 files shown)


itaysk (commented):
this has a change in behavior where the version check is skipped if the env var is specified (I don't know if you intended this). I'm ok with it, actually Liz asked if we can support this, but it should be documented.
If we want to keep the version check for the env var, instead of returning the path from the env var directly, we can add its basedir to the "searchPaths" and let the rest take case of itself.

yanivagman:
This behavior is intended.
It can help with the case of cross compilation, where you compile on a different machine, thus having the wrong kernel version in the bpf file name when you run it on the other machine.
I don't see in the readme that wer'e saying that TRACEE_BPF_FILE should point to a filename with the correct version, just to the bpf file. Do you want to point me to where you think we should update the readme file and I'll add it here?

itaysk:
ok then

itaysk (approved):

itaysk:
wait, on a second read of you comment

>It can help with the case of cross compilation, where you compile on a different machine, thus having the wrong kernel version in the bpf file name when you run it on the other machine.

in this case we don't check that the versions match

yanivagman:
When providing TRACEE_BPF_FILE environment variable, I think it is ok to assume that it is the user's responsibility to validate that the versions match

PR #371, opened 2020-12-21
Support ARM64 architecture

### LINKED ISSUE #234: ARM support
Tracee currently doesn't work on ARM architecture. If you are interested in this, please add the 👍 reaction to this issue, and optionally add a comment to describe your use case.

From: https://github.com/aquasecurity/tracee/discussions/231#discussioncomment-89249
> The reason why Tracee will not work on arm devices is that the event ids numbers match the x86_64 syscalls numbers.
I recently added a table that maps x86_64 syscalls to x86_32 syscalls to support 32bit applications running in compat mode.
We can do something similar for arm, but some other changes are also required, like reading the arguments to different registers in pt_regs struct in the bpf code.

yanivagman:
Add support for ARM64 and other architectures in the future by
seperating system calls constants to different files and by
conditionally compiling bpf code according given architecture.

Close #234


itaysk:
this is awesome! which arm versions do we support?

yanivagman:
Arm64 architecture (arm v8)

itaysk (commented):
doesn't `save_args_from_regs` also need to change?

yanivagman:
`save_args_from_regs` is guarded with ifdef. No more updates are required

itaysk (approved):

PR #375, opened 2020-12-21
Add ability to capture stack traces

pathtofile:
eBPF Has the ability to capture stack traces at the location the program was called. This PR adds the ability to capture the stack traces into the outputted event JSON.

As this can have a performance impact, this feature is behind a `--capture-stack-traces` cli flag.

The PR has had to add or alter a number of different parts of tracee:
- Adding a new eBPF map of type `BPF_MAP_TYPE_STACK_TRACE`, to keep track of the stack traces
- Adding a new config option `CONFIG_CAPTURE_STACK_TRACES` to track if we wish to capture stacks or not
- A new function in `libbpfgo` to lookup a value from an eBPF Map
- Altering the `Event` struct to have a `StackTrace` param

Where necessary I've tried to make sure to be architecture-agnostic, e..g when calculating pointer sizes, etc.

By eBPF's design, the Stack Traces Map has a max number of different stack traces to keep track of, as well as a max stack trace depth. Currently, these are just hard-coded for simplicity.

I felt adding the stack traces to the `table` format wasn't necessary, as I feel stack traces fall into the 'very verbose' bucket. But they are available in the JSON output.

Currently, the stack trace output is just a string of comma-sepearted hex values. This could be changed to be an array or strings, I'm happy either way.

### DIFF
--- a/main.go
+++ b/main.go
@@ -56,6 +56,7 @@ func main() {
 				SecurityAlerts:        c.Bool("security-alerts"),
 				EventsFile:            os.Stdout,
 				ErrorsFile:            os.Stderr,
+				StackAddresses:        c.Bool("stack-addresses"),
 			}
 			capture := c.StringSlice("capture")
 			for _, cap := range capture {
@@ -189,6 +190,11 @@ func main() {
 				Value:       "if-needed",
 				Usage:       "when to build the bpf program. possible options: 'never'/'always'/'if-needed'",
 				Destination: &buildPolicy,
+			},
+			&cli.BoolFlag{
+				Name:  "stack-addresses",
+				Value: false,
+				Usage: "Include stack memory addresses for each event",
 			},
 		},
 	}
--- a/tracee/pipeline.go
+++ b/tracee/pipeline.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"fmt"
+	"strconv"
 	"sync"
 
 	"github.com/aquasecurity/tracee/tracee/external"
@@ -108,6 +109,36 @@ func (t *Tracee) processRawEvent(done <-chan struct{}, in <-chan RawEvent) (<-ch
 	return out, errc, nil
 }
 
+func (t *Tracee) getStackAddresses(StackID uint32) ([]uint64, error) {
+	StackAddresses := make([]uint64, maxStackDepth)
+	stackFrameSize := (strconv.IntSize / 8)
+
+	// Lookup the StackID in the map
+	// The ID could have aged out of the Map, as it only holds a finite number of
+	// Stack IDs in it's Map
+	stackBytes, err := t.StackAddressesMap.GetValue(StackID, stackFrameSize*maxStackDepth)
+	if err != nil {
+		return StackAddresses[0:0], nil
+	}
+
+	stackCounter := 0
+	for i := 0; i < len(stackBytes); i += stackFrameSize {
+		StackAddresses[stackCounter] = 0
+		stackAddr := binary.LittleEndian.Uint64(stackBytes[i : i+stackFrameSize])
+		if stackAddr == 0 {
+			break
+		}
+		StackAddresses[stackCounter] = stackAddr
+		stackCounter++
+	}
+
+	// Attempt to remove the ID from the map so we don't fill it up
+	// But if this fails continue on
+	_ = t.StackAddressesMap.DeleteKey(StackID)
+
+	return StackAddresses[0:stackCounter], nil
+}
+
 func (t *Tracee) prepareEventForPrint(done <-chan struct{}, in <-chan RawEvent) (<-chan external.Event, <-chan error, error) {
 	out := make(chan external.Event, 1000)
 	errc := make(chan error, 1)
@@ -135,7 +166,14 @@ func (t *Tracee) prepareEventForPrint(done <-chan struct{}, in <-chan RawEvent) 
 					continue
 				}
 			}
-			evt, err := newEvent(rawEvent.Ctx, argMetas, args)
+
+			// Add stack trace if needed
+			var StackAddresses []uint64
+			if t.config.StackAddresses {
+				StackAddresses, _ = t.getStackAddresses(rawEvent.Ctx.StackID)
+			}
+
+			evt, err := newEvent(rawEvent.Ctx, argMetas, args, StackAddresses)
 			if err != nil {
 				errc <- err
 				continue
--- a/tracee/printer.go
+++ b/tracee/printer.go
@@ -69,7 +69,7 @@ func newEventPrinter(kind string, containerMode bool, out io.Writer, err io.Writ
 	return res, nil
 }
 
-func newEvent(ctx context, argMetas []external.ArgMeta, args []interface{}) (external.Event, error) {
+func newEvent(ctx context, argMetas []external.ArgMeta, args []interface{}, StackAddresses []uint64) (external.Event, error) {
 	e := external.Event{
 		Timestamp:           float64(ctx.Ts) / 1000000.0,
 		ProcessID:           int(ctx.Pid),
@@ -88,6 +88,7 @@ func newEvent(ctx context, argMetas []external.ArgMeta, args []interface{}) (ext
 		ArgsNum:             int(ctx.Argnum),
 		ReturnValue:         int(ctx.Retval),
 		Args:                make([]external.Argument, 0, len(args)),
+		StackAddresses:      StackAddresses,
 	}
 	for i, arg := range args {
 		e.Args = append(e.Args, external.Argument{
--- a/tracee/tracee.bpf.c
+++ b/tracee/tracee.bpf.c
@@ -48,6 +48,8 @@
 #define MAX_STRING_SIZE     4096          // Choosing this value to be the same as PATH_MAX
 #define MAX_STR_ARR_ELEM    40            // String array elements number should be bounded due to instructions limit
 #define MAX_PATH_PREF_SIZE  64            // Max path prefix should be bounded due to instructions limit
+#define MAX_STACK_ADDRESSES 1024          // Max amount of different stack trace addresses to buffer in the Map
+#define MAX_STACK_DEPTH     20            // Max depth of each stack trace to track
 #define MAX_STR_FILTER_SIZE 16            // Max string filter size should be bounded to the size of the compared values (comm, uts)
 
 #define SUBMIT_BUF_IDX      0
@@ -128,21 +130,22 @@
 #define SCHED_PROCESS_EXIT    1010
 #define MAX_EVENT_ID          1011
 
-#define CONFIG_SHOW_SYSCALL     1
-#define CONFIG_EXEC_ENV         2
-#define CONFIG_CAPTURE_FILES    3
-#define CONFIG_EXTRACT_DYN_CODE 4
-#define CONFIG_TRACEE_PID       5
-#define CONFIG_UID_FILTER       6
-#define CONFIG_MNT_NS_FILTER    7
-#define CONFIG_PID_NS_FILTER    8
-#define CONFIG_UTS_NS_FILTER    9
-#define CONFIG_COMM_FILTER      10
-#define CONFIG_PID_FILTER       11
-#define CONFIG_CONT_FILTER      12
-#define CONFIG_FOLLOW_FILTER    13
-#define CONFIG_NEW_PID_FILTER   14
-#define CONFIG_NEW_PIDNS_FILTER 15
+#define CONFIG_SHOW_SYSCALL         1
+#define CONFIG_EXEC_ENV             2
+#define CONFIG_CAPTURE_FILES        3
+#define CONFIG_EXTRACT_DYN_CODE     4
+#define CONFIG_TRACEE_PID           5
+#define CONFIG_CAPTURE_STACK_TRACES 6
+#define CONFIG_UID_FILTER           7
+#define CONFIG_MNT_NS_FILTER        8
+#define CONFIG_PID_NS_FILTER        9
+#define CONFIG_UTS_NS_FILTER        10
+#define CONFIG_COMM_FILTER          11
+#define CONFIG_PID_FILTER           12
+#define CONFIG_CONT_FILTER          13
+#define CONFIG_FOLLOW_FILTER        14
+#define CONFIG_NEW_PID_FILTER       15
+#define CONFIG_NEW_PIDNS_FILTER     16
 
 // get_config(CONFIG_XXX_FILTER) returns 0 if not enabled
 #define FILTER_IN  1
@@ -191,6 +194,17 @@ BPF_MAP(_name, BPF_MAP_TYPE_PROG_ARRAY, u32, u32, _max_entries);
 #define BPF_PERF_OUTPUT(_name) \
 BPF_MAP(_name, BPF_MAP_TYPE_PERF_EVENT_ARRAY, int, __u32, 1024);
 
+// Stack Traces are slightly different
+// in that the value is 1 big byte array
+// of the stack addresses
+#define BPF_STACK_TRACE(_name, _max_entries) \
+struct bpf_map_def SEC("maps") _name = { \
+  .type = BPF_MAP_TYPE_STACK_TRACE, \
+  .key_size = sizeof(u32), \
+  .value_size = sizeof(size_t) * MAX_STACK_DEPTH, \
+  .max_entries = _max_entries, \
+};
+
 #ifdef RHEL_RELEASE_CODE
 #if (RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(8, 0))
 #define RHEL_RELEASE_GT_8_0
@@ -218,6 +232,7 @@ typedef struct context {
     char uts_name[TASK_COMM_LEN];
     u32 eventid;
     s64 retval;
+    u32 stack_id;
     u8 argnum;
 } context_t;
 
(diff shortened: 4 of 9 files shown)


tracee/tracee.go
```diff
@@ -997,7 +999,8 @@ type context struct {
 	EventID  int32
 	Retval   int64
 	Argnum   uint8
-	_        [7]byte //padding
+	StackID  [4]byte
```
yanivagman, line 1002: We can save stackid as uint32, which is more natural

main.go
```diff
@@ -65,6 +65,7 @@ func main() {
 				SecurityAlerts:        c.Bool("security-alerts"),
 				EventsFile:            os.Stdout,
 				ErrorsFile:            os.Stderr,
+				CaptureStackTraces:    c.Bool("capture-stack-traces"),
```
yanivagman, line 68: We alreay have a "capture" flag (see 2 lines below).
We can add this as a new capture option - "stack-traces"

tracee/pipeline.go
```diff
@@ -106,6 +108,36 @@ func (t *Tracee) processRawEvent(done <-chan struct{}, in <-chan RawEvent) (<-ch
 	return out, errc, nil
 }
 
+func (t *Tracee) getStackTrace(StackID [4]byte) (string, error) {
+	stackFrameSize := (strconv.IntSize / 8)
```
yanivagman, line 112: Maybe I'm missing something, but shouldn't address size be 64bit on a 64bit system (and strconv.IntSize is 32bit, right?)

tracee/printer.go
```diff
@@ -85,6 +85,7 @@ type Event struct {
 	EventName           string     `json:"eventName"`
 	ArgsNum             int        `json:"argsNum"`
 	ReturnValue         int        `json:"returnValue"`
+	StackTrace          string     `json:"stackTrace"`
```
yanivagman, line 88: I think that a slice of int will fit better than string

tracee/tracee.bpf.c
```diff
@@ -176,6 +180,17 @@ BPF_MAP(_name, BPF_MAP_TYPE_PERF_EVENT_ARRAY, int, __u32, 1024);
 #error Minimal required kernel version is 4.14
 #endif
 
+// Stack Traces are slightly different
+// in that the value is 1 big byte array
+// of the stack addresses
+#define BPF_STACK_TRACE(_name, _max_entries) \
```
yanivagman, line 186: Can you please move this defintion 10 lines before this point (after `#define BPF_PERF_OUTPUT(_name)`...)

tracee/tracee.bpf.c
```diff
@@ -627,15 +647,23 @@ static __always_inline int save_context_to_buf(buf_t *submit_p, void *ptr)
...
     init_context(&context);
     context.eventid = id;
     context.argnum = argnum;
     context.retval = ret;
-    save_context_to_buf(submit_p, (void*)&context);
 
+    // Get Stack trace
+    if (get_config(CONFIG_CAPTURE_STACK_TRACES)) {
+        u32 stack_id = bpf_get_stackid(ctx, &stack_traces, BPF_F_REUSE_STACKID | BPF_F_USER_STACK);
+        if (stack_id >= 0) {
```
yanivagman, line 661: Out of curiousity, when does stack_id<0?

libbpfgo/libbpfgo.go
```diff
@@ -302,6 +302,40 @@ func (m *Module) GetMap(mapName string) (*BPFMap, error) {
...
+		keyPtr = unsafe.Pointer(&k[0])
+	} else {
+		return nil, fmt.Errorf("failed to update map %s: unknown key type %T", b.name, key)
+	}
+
+	// Create pointer to value byte array
+	if valueSize == 0 {
+		return nil, fmt.Errorf("valueSize must be > 0")
+	}
+	value := make([]byte, valueSize)
```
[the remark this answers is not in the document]
pathtofile: So firstly, again very happy to follow to and the core maintainer's lead in what you want your project to look like. But to me, a wrapper/healper library should stop you from making _invalid_ decisions, but not _bad_ ones.

If someone wanted to use libbpfgo to create a map with an unreasonable key or value size, that's on them - In the kernel source, [ArrayMap](https://github.com/torvalds/linux/blob/d635a69dd4981cc51f90293f5f64268620ed1565/kernel/bpf/arraymap.c#L72) values for example just have to be less than `KMALLOC_MAX_SIZE`, which is pretty big.

It would make sense for us to restrict the size to `KMALLOC_MAX_SIZE`, but I feel that that setting the limit to anything smaller (which might be more reasonable to you and me), is being more opinionated than a library should. 

Looking at the [libbpf source](https://github.com/libbpf/libbpf/blob/42baefba714b2cce03fbfd68ecfff507a80adc1d/src/bpf.c#L80), they impose no restriction on value or key size, only passing the values on to the kernel to deal with

tracee/pipeline.go
pathtofile: So I think it might be poorly named but on 64bit, `strconv.IntSize` returns `64`: https://play.golang.org/p/BNHTlmWD5Dg

pathtofile: I'm no go expert, I took the advice from stack overflow 😅  https://stackoverflow.com/questions/25741841/how-can-i-check-if-my-program-is-compiled-for-32-or-64-bit-processor

tracee/tracee.bpf.c
pathtofile: Sorry, this should be a signed int. From [doco](https://www.man7.org/linux/man-pages/man7/bpf-helpers.7.html):
```
The stack id retrieved is a 32 bit long integer handle
...
Return: The positive or null stack id on success, or a negative error in case of failure.
```

pathtofile: I made the decision not to care about errors on the eBPF side, to ensure we don't impact performance as much

yanivagman:
One thing I noticed, is that after leaving tracee to work with stack traces for some time, there is some point were no more stack traces are received, and I only see empty string from that point. Have you noticed this as well?

pathtofile:
> 
> 
> One thing I noticed, is that after leaving tracee to work with stack traces for some time, there is some point were no more stack traces are received, and I only see empty string from that point. Have you noticed this as well?

Investigated this one, I had a misunderstanding of how the stack traces map worked. It is I now believe a regular hash map, and so when `max_entries` is reached it is full and stops storing stack traces. To fix this, in `getStackTrace()` once I get the stack trace from the map, I delete the key from the map, 'freeing up' a slot for a new stack trace ID.

In theory, we could still hit issues if there are more than `MAX_STACK_TRACES` events in the pipeline at once, but considering I have this as 1024, I think that is probably acceptable, thoughts?

pathtofile:
@yanivagman Made all the above changes, should be good for another look :-)

main.go
```diff
@@ -216,6 +217,11 @@ func main() {
 				Usage:       "when to build the bpf program. possible options: 'never'/'always'/'if-needed'",
 				Destination: &buildPolicy,
 			},
+			&cli.BoolFlag{
```
yanivagman, line 220: No need for this new flag - we already have "capture" flag, which can capture "exec/mem/write/all" where we can also add "stack-traces"

main.go
```diff
@@ -65,6 +65,7 @@ func main() {
 				SecurityAlerts:        c.Bool("security-alerts"),
 				EventsFile:            os.Stdout,
 				ErrorsFile:            os.Stderr,
+				StackTraces:           c.Bool("stack-traces"),
```
yanivagman, line 68: Please see my other comment.
What I actually meant is adding this option to the capture flag

tracee/tracee.bpf.c
```diff
@@ -194,6 +209,7 @@ typedef struct context {
     u32 eventid;
     s64 retval;
     u8 argnum;
+    char stack_id[STACK_ID_LEN];
```
yanivagman, line 212: This can be u32 instead (the return value of bpf_get_stackid is int, and you ignore negative values)

tracee/tracee.bpf.c
```diff
@@ -464,6 +481,9 @@ static __always_inline int init_context(context_t *context)
     // Save timestamp in microsecond resolution
     context->ts = bpf_ktime_get_ns()/1000;
 
+    // Clean Stack Trace ID
+    __builtin_memset(&context->stack_id, 0, STACK_ID_LEN);
```
yanivagman, line 485: After changing to u32:
context->stack_id = 0;

yanivagman (commented):
I like the changes you made to libbpf - expanding it's capabilities will allow other projects to use it as a go wrapper to libbpf.
Please see my other comments

tracee/tracee.go
```diff
@@ -997,7 +999,8 @@ type context struct {
 	EventID  int32
 	Retval   int64
 	Argnum   uint8
-	_        [7]byte //padding
+	StackID  uint32
+	_        [3]byte //padding
```
yanivagman, line 1003: Not sure this padding is now correct.
It was 7 bytes when we had one byte at the end,
and the struct needs to be 8 bytes aligned.
Now I think the padding should be 4.
We had some issues with this in the passed which made @itaysk set this padding

main.go
itaysk: the capture flag captures artifacts as files into the output directory, wheres this just tells trace to print more info to stdout, not unlike --show-exec-env for example. I don't think it belongs in the capture flag

yanivagman: I don't think it's the same. show-exec-env is printing an argument (env) which is part of execve event, that we otherwise hide. Here we want to add stack traces, which will be captured for every event, and that require the use of a special mechanism to capture this data.
I get your point though that this data is part of the printed output and not collected as a file to the tracee output directory, so I don't mind leaving it here

yanivagman (changes requested):
Hi @pathtofile 

Are you still working on this?
I think there is not much left to change, and I want to take this change for the next release

main.go
pathtofile: Sorry, what's the conclusion here, would you prefer this to be part of the 'capture' flag, or kept as a seperate flag?

tracee/pipeline.go
```diff
@@ -106,6 +107,44 @@ func (t *Tracee) processRawEvent(done <-chan struct{}, in <-chan RawEvent) (<-ch
...
+			break
+		}
+		StackTrace[stackCounter] = stackAddr
+		stackCounter++
+	}
+
+	// Remove the ID from the map to we don't fill it up
+	err = stackTracesMap.DeleteKey(StackID)
+	if err != nil {
+		fmt.Printf("[**] Failed to delte key!\n")
```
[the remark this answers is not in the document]
pathtofile: Debug print left in, will remove

tracee/pipeline.go
```diff
@@ -106,6 +107,44 @@ func (t *Tracee) processRawEvent(done <-chan struct{}, in <-chan RawEvent) (<-ch
...
+
+	// Get Map that holds the stack traces
+	stackTracesMap, err := t.bpfModule.GetMap("stack_traces")
+	if err != nil {
+		return StackTrace[0:0], nil
+	}
+
+	// Lookup the StackTraceID in the map
+	// The Id could be missing for various reasons, including
+	// the id has aged out, or we are not collecting stack traces
```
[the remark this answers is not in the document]
pathtofile: No, it is not possible, so will remove that part of the comment.

main.go
```diff
@@ -216,6 +217,11 @@ func main() {
 				Usage:       "when to build the bpf program. possible options: 'never'/'always'/'if-needed'",
 				Destination: &buildPolicy,
 			},
+			&cli.BoolFlag{
+				Name:  "stack-traces",
+				Value: false,
+				Usage: "Include stack traces for each event",
```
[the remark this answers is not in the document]
pathtofile: That's very true, what if I call this feature "stack addresses" then, as well as add that to the description? That way it makes it obvious that say `--stack-addresses` will be returning addresses, not function pointers

tracee/pipeline.go
```diff
@@ -106,6 +107,44 @@ func (t *Tracee) processRawEvent(done <-chan struct{}, in <-chan RawEvent) (<-ch
...
 
+func (t *Tracee) getStackTrace(StackID uint32) ([]uint64, error) {
+	StackTrace := make([]uint64, maxStackDepth)
+	stackFrameSize := (strconv.IntSize / 8)
+
+	// Get Map that holds the stack traces
+	stackTracesMap, err := t.bpfModule.GetMap("stack_traces")
+	if err != nil {
+		return StackTrace[0:0], nil
+	}
```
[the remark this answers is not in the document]
pathtofile: Agreed, will do this once and save the map reference in the Tracee Struct

tracee/tracee.go
pathtofile: Is that correct? So on the C side, `context_t` with the new `stack_id` looks like this:
```
typedef struct context {
    u64 ts;                     // Timestamp
    u32 pid;                    // PID as in the userspace term
    u32 tid;                    // TID as in the userspace term
    u32 ppid;                   // Parent PID as in the userspace term
    u32 host_pid;               // PID in host pid namespace
    u32 host_tid;               // TID in host pid namespace
    u32 host_ppid;              // Parent PID in host pid namespace
    u32 uid;
    u32 mnt_id;
    u32 pid_id;
    char comm[16];
    char uts_name[16];
    u32 eventid;
    s64 retval;
    u8 argnum;
    u32 stack_id;
} context_t;
```
So thats `(8 + 4 + 4 + 4 + 4 + 4 + 4 + 4 + 4 + 4 + 16 + 16 + 4 + 8 + 1 + 4) == 93 bytes`. And I checked with a debug string that `sizeof(context_t) == 96`, i.e. has a 3 Byte padding

tracee/tracee.bpf.c
pathtofile: Sorry, this one is currently holding this review up.

I 100% agree that this should be a `uint32` and not a char array, but *something* is truncating the uint32 when unmarshalling it, and converting only the first 'byte' of the int. Looking into this atm, I'm sure it's an issue with my code, somewhere

pathtofile: Ok, worked it out - `uint32`s also want to be aligned on a boundary (makes sense), which the 1-byte `argnum` throws this off, resulting in the 'padding' being in between `argum` and `stackid`.

I think the correct solution here is to keep the 1-Byte `argnum` parameter always at the end of the struct, and just put `stackID` before it (and make sure the struct is the same on the Golang and C sides). I've also added a note a bout this in a comment, in case the next person to alter the context struct has the same issue.

pathtofile:
Hey @yanivagman and @itaysk ,  apologies for the delay, I think I've made the various changes requests, except for this discussion about ensuring people understand this is addresses and not expecting function names: https://github.com/aquasecurity/tracee/pull/375#discussion_r554509199

I've also pulled in the latest main and merged the conflicts

main.go
itaysk: we've recently agreed to make the capture flag even more about artifact configuration, so I think definitely not under capture. @yanivagman I propose we merge it in with its own flag, and perhaps reconsider later as part of the output flag refactor we plan

main.go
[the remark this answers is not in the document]
itaysk: sounds good to me

tracee/tracee.go
itaysk: you're logic makes sense. when working on this I remember learning that the napkin calculation sometimes failed me, so I eventually added a debug statement, saw the memory layout and reverse-explained how the padding worked. I'm sorry I don't remember what was the particular caveat that made my original calc fail, but If you confirmed it in practice, it should be good. my method was to add a print in the go code (today that would be here https://github.com/aquasecurity/tracee/blob/main/tracee/pipeline.go#L50), and print a hex-dump of the event, yours sounds good as well.

itaysk: @yanivagman do we need any special casing for ARM? I don't know how it works there just asking

main.go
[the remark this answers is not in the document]
pathtofile: Done, renamed in comments and code on both C and GoLang sides

tracee/tracee.go
yanivagman: From what I remember when I wrote the above comment, StackId (uint32) was the last struct field before the padding, and not the Argnum(u8).
Padding, as far as I remember, work as follows:
1. Each struct should be aligned to its biggest field size (in our case, u64 which is 8 bytes)
2. Each field of the struct should be aligned according to it's size (thus, u32 is 4 bytes aligned)

So, when StackId was the last field before the padding, it was 4 bytes aligned, thus needed 4 bytes padding to align the struct (to 8 bytes). After Argnum is moved to be before the padding, there is a needed to add 3 more bytes to align to 8 bytes, as required by the struct - so it should also be fine

yanivagman (approved):

itaysk (approved):

itaysk:
@pathtofile can you please rebase and resolve conflict with https://github.com/aquasecurity/tracee/pull/402, and then squash?

pathtofile:
Hey @itaysk , code's been rebased and squashing into a single commit, however the build is failing, as this change introduces the `GetValue` and `DeleteKey` functions to `golibbpf`, which in the go modules is pointing to the current 'main' of the library which doesn't have the new functions.

What's the correct solution here? I'm able to confirm things work by running `go mod edit -replace github.com/aquasecurity/tracee/libbpfgo=./libbpfgo` and building, but I'm not a golang dev, is that a 'hacky' thing to do in a real codebase?

yanivagman:
Hi @pathtofile 

I'm also not sure what's the best way to resolve this in one commit (maybe @itaysk will have a better idea) but here is a possible workaround: you can open a different PR only for the changes in libbpfgo, we will merge those, and then there should be no problem to merge the rest of this PR

pathtofile:
Yeah that's true, ok, put up PR: https://github.com/aquasecurity/tracee/pull/422

pathtofile:
Hey @yanivagman , pulled in and rebased everything, looking better now!

yanivagman:
Thanks @pathtofile this will definitely be a nice addition to Tracee!

PR #376, opened 2020-12-22
Add indexing of written files by path

### LINKED ISSUE #213: index capture written files
add a file to the output directory that indexes the written files by their paths. for example:
`/path/to/file -> write.dev-1.inode-33`

jan0ski:
* Monitors `vfs_write` and `vfs_writev` events
* Records file path, writes a file in the form `write.dev-#.inode-#` to the output directory
* Output directory is indexed by the source file's path

Resolves #213

### DIFF
--- a/tracee/tracee.go
+++ b/tracee/tracee.go
@@ -112,6 +112,7 @@ type Tracee struct {
 	printer       eventPrinter
 	stats         statsStore
 	capturedFiles map[string]int64
+	writtenFiles  map[string]string
 	mntNsFirstPid map[uint32]uint32
 	DecParamName  [2]map[argTag]string
 	EncParamName  [2]map[string]argTag
@@ -200,6 +201,7 @@ func New(cfg TraceeConfig) (*Tracee, error) {
 		return nil, err
 	}
 
+	t.writtenFiles = make(map[string]string)
 	t.capturedFiles = make(map[string]int64)
 	//set a default value for config.maxPidsCache
 	if t.config.maxPidsCache == 0 {
@@ -723,6 +725,22 @@ func (t *Tracee) Run() error {
 	t.eventsPerfMap.Stop()
 	t.fileWrPerfMap.Stop()
 	t.printer.Epilogue(t.stats)
+
+	// record index of written files
+	if t.config.CaptureWrite {
+		destinationFilePath := filepath.Join(t.config.OutputPath, "written_files")
+		f, err := os.OpenFile(destinationFilePath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+		if err != nil {
+			return fmt.Errorf("error logging written files")
+		}
+		defer f.Close()
+		for fileName, filePath := range t.writtenFiles {
+			if _, err := f.WriteString(fmt.Sprintf("%s %s\n", fileName, filePath)); err != nil {
+				return fmt.Errorf("error logging written files")
+			}
+		}
+	}
+
 	// Signal pipeline that Tracee exits by closing the done channel
 	close(done)
 	t.Close()
@@ -781,6 +799,38 @@ func (t *Tracee) shouldProcessEvent(e RawEvent) bool {
 
 func (t *Tracee) processEvent(ctx *context, args map[argTag]interface{}) error {
 	switch ctx.EventID {
+
+	//capture written files
+	case VfsWriteEventID, VfsWritevEventID:
+		if t.config.CaptureWrite {
+			filePath, ok := args[t.EncParamName[ctx.EventID%2]["pathname"]].(string)
+			if !ok {
+				return fmt.Errorf("error parsing vfs_write args")
+			}
+			// path should be absolute, except for e.g memfd_create files
+			if filePath == "" || filePath[0] != '/' {
+				return nil
+			}
+			dev, ok := args[t.EncParamName[ctx.EventID%2]["dev"]].(uint32)
+			if !ok {
+				return fmt.Errorf("error parsing vfs_write args")
+			}
+			inode, ok := args[t.EncParamName[ctx.EventID%2]["inode"]].(uint64)
+			if !ok {
+				return fmt.Errorf("error parsing vfs_write args")
+			}
+
+			// stop processing if write was already indexed
+			fileName := fmt.Sprintf("write.dev-%d.inode-%d", dev, inode)
+			indexName, ok := t.writtenFiles[fileName]
+			if ok && indexName == filePath {
+				return nil
+			}
+
+			// index written file by original filepath
+			t.writtenFiles[fileName] = filePath
+		}
+
 	case SecurityBprmCheckEventID:
 
 		//cache this pid by it's mnt ns


tracee/tracee.go
```diff
@@ -812,6 +812,41 @@ func (t *Tracee) shouldProcessEvent(e RawEvent) bool {
 
 func (t *Tracee) processEvent(ctx *context, args map[argTag]interface{}) error {
 	switch ctx.EventID {
+
+	//capture written files
+	case VfsWriteEventID, VfsWritevEventID:
+		if t.config.CaptureWrite {
+			filePath, ok := args[t.EncParamName[ctx.EventID%2]["pathname"]].(string)
+			if !ok {
+				return fmt.Errorf("error parsing security_bprm_check args")
```
yanivagman, line 821: security_bprm_check -> vfs_write

tracee/tracee.go
```diff
@@ -812,6 +812,41 @@ func (t *Tracee) shouldProcessEvent(e RawEvent) bool {
...
+			}
+			// path should be absolute, except for e.g memfd_create files
+			if filePath == "" || filePath[0] != '/' {
+				return nil
+			}
+
+			destinationDirPath := filepath.Join(t.config.OutputPath, strconv.Itoa(int(ctx.MntID)))
+			if err := os.MkdirAll(destinationDirPath, 0755); err != nil {
```
yanivagman, line 829: We only want to index written file paths, no need to create a directory here

tracee/tracee.go
```diff
@@ -812,6 +812,41 @@ func (t *Tracee) shouldProcessEvent(e RawEvent) bool {
...
+				return err
+			}
+			sourceFileStat, err := os.Stat(filePath)
```
yanivagman, line 832: filePath cannot be Stat-ed as it can be in a different mount namespace (always true for containers)

tracee/tracee.go
```diff
@@ -812,6 +812,41 @@ func (t *Tracee) shouldProcessEvent(e RawEvent) bool {
...
+			if err != nil {
+				return err
+			}
+
+			sourceFileInode := sourceFileStat.Sys().(*syscall.Stat_t).Ino
+			sourceFileDev := sourceFileStat.Sys().(*syscall.Stat_t).Dev
```
yanivagman, line 838: The file's inode number ('inode' argument name) and device id ('dev' argument name) are given as part of the vfs_write(v) event itself, no need to stat for that.

tracee/tracee.go
```diff
@@ -812,6 +812,41 @@ func (t *Tracee) shouldProcessEvent(e RawEvent) bool {
...
+			sourceFileCtime := sourceFileStat.Sys().(*syscall.Stat_t).Ctim.Nano()
+			capturedFileID := fmt.Sprintf("%d:%s", ctx.MntID, filePath)
+			destinationFilePath := filepath.Join(destinationDirPath, fmt.Sprintf("write.dev-%d.inode-%d", sourceFileDev, sourceFileInode))
+			err = CopyFileByPath(filePath, destinationFilePath)
+			if err != nil {
+				return err
+			}
+			//mark this file as captured
+			t.capturedFiles[capturedFileID] = sourceFileCtime
```
yanivagman, line 847: This map is meant to be used by 'capture executed files' feature.
We shouldn't use it here as well

tracee/tracee.go
```diff
@@ -812,6 +812,41 @@ func (t *Tracee) shouldProcessEvent(e RawEvent) bool {
...
+			if err != nil {
+				return err
+			}
+
+			sourceFileInode := sourceFileStat.Sys().(*syscall.Stat_t).Ino
+			sourceFileDev := sourceFileStat.Sys().(*syscall.Stat_t).Dev
+			sourceFileCtime := sourceFileStat.Sys().(*syscall.Stat_t).Ctim.Nano()
+			capturedFileID := fmt.Sprintf("%d:%s", ctx.MntID, filePath)
+			destinationFilePath := filepath.Join(destinationDirPath, fmt.Sprintf("write.dev-%d.inode-%d", sourceFileDev, sourceFileInode))
+			err = CopyFileByPath(filePath, destinationFilePath)
```
yanivagman, line 842: Why should we copy files?
We only want to update an indexing file as described in #213

yanivagman (changes requested):
@jan0ski welcome to the Tracee community!
#213 only asks to create a mapping between the filenames of the written files (which have a name of the format "write.dev-%d.inode-%d") and the original filenames.
It seems that this PR tries to do something else...
If you want, we can disscus this in #213 and I can try to give further clarifications

itaysk:
I don't think a review is necessary at this point since the instructions in the issue were wrong. we can review again once fixed

tracee/tracee.go
jan0ski: Yes!  This is what I was struggling with, thank you for the clarification.  I saw with capturing executed files, the `security_bprm_check` events had a `pathname` argument name.  Where would I find the fields for each event type?

e.g. `filePath, ok := args[t.EncParamName[ctx.EventID%2]["pathname"]].(string)`

yanivagman: Have a look in consts.go.
The table named EventIdsToParams will tell you exactly that

tracee/tracee.go
```diff
@@ -812,6 +814,43 @@ func (t *Tracee) shouldProcessEvent(e RawEvent) bool {
...
+			if err != nil {
+				return fmt.Errorf("error logging written files")
+			}
+			defer f.Close()
+			if _, err := f.WriteString(fmt.Sprintf("%s %s\n", filePath, fileName)); err != nil {
+				return fmt.Errorf("error logging written files")
+			}
+
+			// index written file by original filepath
+			t.writtenFiles[filePath] = fileName
```
yanivagman, line 851: This indexing is not used anywhere.
We should avoid writing the same file several times to the index.
I suggest mapping from dev+inode (as the key) to filePath (as the value). 
dev+inode uniquely identify a file in the system.
If a file is deleted, then recreated with the same name, a new inode will be created.
We will write two different files (with different inode numbers), but the path will be the same.

tracee/tracee.go
```diff
@@ -812,6 +814,43 @@ func (t *Tracee) shouldProcessEvent(e RawEvent) bool {
...
+			}
+			dev, ok := args[t.EncParamName[ctx.EventID%2]["dev"]].(int32)
+			if !ok {
+				return fmt.Errorf("error parsing vfs_write args")
+			}
+			inode, ok := args[t.EncParamName[ctx.EventID%2]["inode"]].(int32)
+			if !ok {
+				return fmt.Errorf("error parsing vfs_write args")
+			}
+
```
yanivagman, line 837: We can stop processing the event here if we already added it to the index

yanivagman (commented):
Great!
This looks much more like what we wanted to achieve in this task.
Please see my comments about the indexing

yanivagman (commented):
Thanks @jan0ski!
Just one typo and please squash into one commit

tracee/tracee.go
```diff
@@ -812,6 +814,49 @@ func (t *Tracee) shouldProcessEvent(e RawEvent) bool {
...
+				return nil
+			}
+
+			// write original filepath mapping to file
+			destinationFilePath := filepath.Join(t.config.OutputPath, "written_files")
+			f, err := os.OpenFile(destinationFilePath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+			if err != nil {
+				return fmt.Errorf("error logging written files")
+			}
+			defer f.Close()
```
itaysk, line 851: open/close for every processed event can introduce unnecessary overhead. I think it would be better to just collect the index here, and actually print it at then end of the run. wdyt?

jan0ski:
Typos fixed, all commits squashed.  Pending any requested changes from this [comment](https://github.com/aquasecurity/tracee/pull/376#discussion_r549242006).

tracee/tracee.go
yanivagman: Agree, that will be more efficient

jan0ski: Assuming this should likely be implemented in `eventPrinter.Epilogue()` in `printer.go` with the stats of the run?

itaysk: actually I wouldn't put it in a `eventPrinter` because it is not output format dependent (it's the same code regardless of the chosen output format, just like the other artifacts). You can call it from `Run` https://github.com/aquasecurity/tracee/blob/main/tracee/tracee.go#L725

tracee/tracee.go
```diff
@@ -723,6 +725,9 @@ func (t *Tracee) Run() error {
 	t.eventsPerfMap.Stop()
 	t.fileWrPerfMap.Stop()
 	t.printer.Epilogue(t.stats)
+
+	fmt.Printf("Written Files: %+v\n", t.writtenFiles)
```
yanivagman, line 729: This is the correct location to save the written files,
but instead of printing them to stdout, can you please save them to a file in tracee's output dir (I think you did that in the previous commit, but here you only open and write to the file once)

tracee/tracee.go
```diff
@@ -723,6 +725,15 @@ func (t *Tracee) Run() error {
...
 	t.printer.Epilogue(t.stats)
+
+	// record index of written files
+	destinationFilePath := filepath.Join(t.config.OutputPath, "written_files")
+	f, err := os.OpenFile(destinationFilePath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+	if err != nil {
+		return fmt.Errorf("error logging written files")
+	}
+	defer f.Close()
+
```
yanivagman, line 736: Missing write of t.writtenFiles to f?

yanivagman (changes requested):
I tested locally and got errors when I run with the following line:
`sudo ./dist/tracee --capture write -f e=vfs_write,vfs_writev`

The problem is that the types of "dev" and "inode" are incorrect

tracee/tracee.go
```diff
@@ -781,6 +799,38 @@ func (t *Tracee) shouldProcessEvent(e RawEvent) bool {
...
+			}
+			// path should be absolute, except for e.g memfd_create files
+			if filePath == "" || filePath[0] != '/' {
+				return nil
+			}
+			dev, ok := args[t.EncParamName[ctx.EventID%2]["dev"]].(uint32)
+			if !ok {
+				return fmt.Errorf("error parsing vfs_write args")
+			}
+			inode, ok := args[t.EncParamName[ctx.EventID%2]["inode"]].(uint32)
```
yanivagman, line 818: uint32 -> uint64

jan0ski: Apologies for the lack of testing here... only system available is WSL and was having difficulting getting kernel headers.

tracee/tracee.go
```diff
@@ -781,6 +799,38 @@ func (t *Tracee) shouldProcessEvent(e RawEvent) bool {
...
+		if t.config.CaptureWrite {
+			filePath, ok := args[t.EncParamName[ctx.EventID%2]["pathname"]].(string)
+			if !ok {
+				return fmt.Errorf("error parsing vfs_write args")
+			}
+			// path should be absolute, except for e.g memfd_create files
+			if filePath == "" || filePath[0] != '/' {
+				return nil
+			}
+			dev, ok := args[t.EncParamName[ctx.EventID%2]["dev"]].(uint64)
```
yanivagman, line 814: Please have a look at https://github.com/aquasecurity/tracee/blob/main/tracee/consts.go#L834
and https://github.com/aquasecurity/tracee/blob/main/tracee/tracee.go#L1326
This one is of type dev_t, which is uint32

tracee/tracee.go
```diff
@@ -781,6 +799,38 @@ func (t *Tracee) shouldProcessEvent(e RawEvent) bool {
...
+			if !ok {
+				return fmt.Errorf("error parsing vfs_write args")
+			}
+			inode, ok := args[t.EncParamName[ctx.EventID%2]["inode"]].(uint64)
```
yanivagman, line 818: This one is indeed uint64 (unsigned long)

tracee/tracee.go
jan0ski: Ah, that clears things up much.  Thank you for the direction and bearing with me.

yanivagman (approved):
Thanks @jan0ski!

PR #379, opened 2020-12-24
signal end of transmission for gob output

itaysk:
when tracee writes into a file, and something else read from this file, we can't rely on EOF to know when the producer has finished sending events. For example if the producer doesn't yet have events to report, the consumer will hit EOF immediately, or if the consumer is consuming events faster than the producer.
The solution is to explicitly signal end of transmission, using a special even that seals the events stream. For now I'm adding it by default for gob output, but in the future we should 1) make it optional and 2) add it to every output that can be consumed in "streaming" fashion

### DIFF
--- a/test/gob/gob.go
+++ b/test/gob/gob.go
@@ -7,9 +7,35 @@ import (
 	"log"
 	"os"
 	"os/signal"
-
-	"github.com/aquasecurity/tracee/tracee"
+	"time"
 )
+
+// Event is a user facing data structure representing a single event
+type Event struct {
+	Timestamp           float64    `json:"timestamp"`
+	ProcessID           int        `json:"processId"`
+	ThreadID            int        `json:"threadId"`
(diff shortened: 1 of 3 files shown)


test/gob/gob.go
```diff
@@ -7,10 +7,36 @@ import (
 	"log"
 	"os"
 	"os/signal"
-
-	"github.com/aquasecurity/tracee/tracee"
+	"time"
 )
 
+// Event is a user facing data structure representing a single event
+type Event struct {
```
yanivagman, line 14: Why having a copy of the struct here?

itaysk: Same reason we have a copy in tracee-rule. Importing entire tracee just for the struct is too much. Also tracee requires linux to compile which is not a requirement for this program. In the future I plan to move these types outside the tracee package and then we could import them

yanivagman (approved):

PR #383, opened 2020-12-28
Fix event listing

### LINKED ISSUE #382: Listing events is broken
When using `tracee -l` to list the events, the output includes many blank lines and "other events" is empty.
This was probably caused as a result of adding support for arm64 architecture

yanivagman:
Fix #382

### DIFF
--- a/main.go
+++ b/main.go
@@ -575,8 +575,8 @@ func printList() {
 	b.WriteString("System Calls:              Sets:\n")
 	b.WriteString("____________               ____\n\n")
 	for i := 0; i < int(tracee.SysEnterEventID); i++ {
-		event := tracee.EventsIDToEvent[int32(i)]
-		if event.Name == "reserved" {
+		event, ok := tracee.EventsIDToEvent[int32(i)]
+		if !ok {
 			continue
 		}
 		if event.Sets != nil {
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -588,7 +588,7 @@ func printList() {
 	}
 	b.WriteString("\n\nOther Events:              Sets:\n")
 	b.WriteString("____________               ____\n\n")
-	for i := int(tracee.SysEnterEventID); i < len(tracee.EventsIDToEvent); i++ {
+	for i := int(tracee.SysEnterEventID); i < int(tracee.MaxEventID); i++ {
```
itaysk, line 591: is this change needed for the fix?

yanivagman: The length of the map is now smaller than the maximum event id.
There is a gap between the last syscall number (arch dependent) and common events (starting from 1000)

main.go
```diff
@@ -576,7 +576,7 @@ func printList() {
 	b.WriteString("____________               ____\n\n")
 	for i := 0; i < int(tracee.SysEnterEventID); i++ {
 		event := tracee.EventsIDToEvent[int32(i)]
-		if event.Name == "reserved" {
+		if event.Name == "" {
```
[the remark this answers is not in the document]
yanivagman: We can't assume syscalls are sequential - there are unimplemented syscalls in arm, for example.
continue is good enough

itaysk (approved):

PR #388, opened 2020-12-30
refactor Event and params

itaysk:
this PR does a couple of refactors:
1. extract the `Event` type (and related baggage) from package `tracee` to another package that can be easily imported by consumers of Tracee (e.g tracee-rules, or any consumer of `-o gob`).
2. still on the topic of external consumers, when using `-o json` (like tracee-rules' `stdin-as`) all event arguments of numerical types are unmarshaled as float64 regardless of their original type (this is how json works). To fix this, I have included the original type information in the json payload, and added a custom unmarshaler to use this information and unmarshal to the original type


itaysk:
@roikol - this should solve the float64 issue in a better way

itaysk:
@yanivagman sorry it's actually not ready for review

itaysk:
I recommend starting the review with external_test to understand the end goal

tracee/external/external.go
```diff
@@ -0,0 +1,95 @@
...
+// Argument holds the information for one argument
+type Argument struct {
+	ArgMeta
+	Value interface{} `json:"value"`
+}
+
+// ArgMeta describes an argument
+type ArgMeta struct {
+	Name         string `json:"name"`
+	OriginalType string `json:"originalType"`
```
yanivagman, line 40: Why OriginalType and not just Type?

tracee/printer.go
```diff
@@ -112,11 +86,18 @@ func newEvent(ctx context, argsNames []string, args []interface{}) (Event, error
...
+	params, ok := EventsIDToParams[ctx.EventID]
+	if !ok {
+		return e, fmt.Errorf("event %s(%d) is missing from params registry", e.EventName, e.EventID)
 	}
 	for i, arg := range args {
-		e.Args = append(e.Args, Argument{
-			Name:  argsNames[i],
+		e.Args = append(e.Args, external.Argument{
+			ArgMeta: external.ArgMeta{
+				Name:         params[i].Name,
```
yanivagman, line 98: You can't assume the arg name to be found according to its index.
There are cases where the argument number changes for the same event.
for example: envp in execve event is only shown if the user chooses to do that.
Another example: bpf code failed to read some argument (e.g. string), thus this argument will not be sent at first place

tracee/external/external.go
itaysk: I originally called it just Type and then changed to OriginalType because when you look at the Arg struct, or the json of it, you see "name", "type" and "value" which implies "type" is the type of the value. but it's the type of the syscall arg. If you thing Type is better I will change it back, I also wasn't sure about this.

yanivagman: Not sure I get your point here. Which Arg struct? You mean Argument struct?
This is ArgMeta struct, I don't think it will be confusing to call it Type here

itaysk: > You mean Argument struct?

yes

> This is ArgMeta struct

but ArgMeta is embedded within Argument. The idea is that we can use ArgMeta as is (for param tables) but also in Argument (for serialization). So if someone looks at the Argument struct, or the serialized Argument, they see "name" "type" "value". which implies the type describes the value in this struct, which is not true, it describes the original syscall value

yanivagman: I don't see a problem with that. I think Type is fine

tracee/printer.go
```diff
@@ -112,11 +86,18 @@ func newEvent(ctx context, argsNames []string, args []interface{}) (Event, error
...
+	params, ok := EventsIDToParams[ctx.EventID]
+	if !ok {
+		return e, fmt.Errorf("event %s(%d) is missing from params registry", e.EventName, e.EventID)
 	}
 	for i, arg := range args {
-		e.Args = append(e.Args, Argument{
-			Name:  argsNames[i],
+		e.Args = append(e.Args, external.Argument{
+			ArgMeta: external.ArgMeta{
+				Name: params[i].Name,
```
yanivagman, line 98: You can't assume the arg name to be found according to its index.
There are cases where the argument number changes for the same event.
for example: envp in execve event is only shown if the user chooses to do that.
Another example: bpf code failed to read some argument (e.g. string), thus this argument will not be sent at first place

yanivagman (approved):

PR #390, opened 2020-12-31
add code injection signature

itaysk:
cc: @roikol

### DIFF
--- a/tracee-rules/signatures/golang/code_injection.go
+++ b/tracee-rules/signatures/golang/code_injection.go
@@ -100,7 +100,7 @@ func (sig *codeInjection) OnEvent(e types.Event) error {
 	case "execve":
 		envs, err := GetTraceeArgumentByName(ee, "envp")
 		if err != nil {
-			return err
+			break
 		}
 		envsSlice := envs.Value.([]string)
 		for _, env := range envsSlice {
--- a/tracee-rules/signatures/golang/code_injection.go
+++ b/tracee-rules/signatures/golang/code_injection.go
@@ -56,12 +56,12 @@ func (sig *codeInjection) OnEvent(e types.Event) error {
 	}
 	switch ee.EventName {
 	case "open", "openat":
-		flags, err := GetArgumentByName(ee, "flags")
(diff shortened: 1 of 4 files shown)


tracee-rules/signatures/golang/code_injection_test.go
```diff
@@ -0,0 +1,75 @@
...
+						{
+							Name:  "argv",
+							Value: []string{"ls"},
+						},
+					},
+				},
+			},
+			expect: true,
+		},
+	}
```
yanivagman, line 61: Not a must, but maybe you will also want to add a  test with expect:false

tracee-rules/signatures/golang/helpers.go
```diff
@@ -0,0 +1,25 @@
+package main
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/aquasecurity/tracee/tracee-rules/types"
+)
+
+func GetArgumentByName(event types.TraceeEvent, argName string) (types.TraceeEventArgument, error) {
```
yanivagman, line 10: As we expect more input sources to be added in the future, I think that the function name should include Tracee in it

tracee-rules/signatures/golang/code_injection.go
```diff
@@ -0,0 +1,131 @@
...
+		},
+	}, nil
+}
+
+func (sig *codeInjection) GetSelectedEvents() ([]types.SignatureEventSelector, error) {
+	return []types.SignatureEventSelector{
+		{Source: "tracee", Name: "ptrace"},
+		{Source: "tracee", Name: "open"},
+		{Source: "tracee", Name: "openat"},
+		{Source: "tracee", Name: "execve"},
```
yanivagman, line 44: execveat?

itaysk: I said the same thing to Roi when we wrote the code. I don't know why it was missing from the python signature, but we decided to just port the python sig as is, and @roikol should improve it if he think so

yanivagman (approved):

yanivagman:
Don't forget to squash!

PR #394, opened 2021-01-03
fix compat detection for older kernels

itaysk:


yanivagman:
Not sure we need this one.
We will soon move to kernel 4.18 as a minimal requirement (needed for co-re, and bpf cgroupid helper function)

itaysk:
this was a real world issue, I didn't look for it. because we advertise 4.14, someone with older kernel went and upgraded to 4.14 in order to use tracee. and tracee didn't work. 
I've confirmed that this works on 4.14-4.14.17 now so it's ready for review, if you don't want to merge it now, we need to bump the requirements asap because they're misleading

yanivagman (approved):

Issue #377, opened 2020-12-23
Consolidate filtering under filter flag


yanivagman:
To filter events, Tracee uses several flags: `--filter`, `--event`, `--exclude-event`, `--events-set`  and one more `--trace` flag which chooses the "mode".
As disscused in #366 we can consolidate all those flags into one filter flag.

To do that, following filters will remain unchanged (already used today with --filter flag):
```
--filter uid=uid1,uid2,...
--filter pidns=pidns1,pidns2...
--filter mntns=mntns1,mntns2...
--filter uts=name1,name2,....
--filter comm=comm1,comm2,...
```
It is also possible to choose != with the above filters.

- [x] The `--event` and `--exclude-event` flags will be merged into one `event` filter:

```
--filter event=event1,event2,...
--filter event!=event1,event2...
```

- [x] The `--events-set` flag will be moved to `--filter` flag:

`--filter set=set1,set2,...`
We may consider adding set!=set1,set2 in the future, as this is not supported today

The `--trace` flag, which included the following tracing modes:
```
process:new
process:pid1,pid2,...
process:follow
process:all
container:new
container:all
host:all
host:new
```
- [x] will be changed to include the following modes:
```
new
follow
all
```
, and it's filtering features will be moved to `--filter` flag as follows:

- [x] `--trace process:pid1,pid2,...` will become `--filter pid=pid1,pid2...` (we should also add support for != pid)

- [x] `container` filter will be added, which will have a boolean value:

`--filter container` or `--filter !container`
If no container filter was given - processes from the host and from containers will be traced (equals in some way to "process:xxx")
If container filter was given, only processes which are part of a container will be traced (equals in a sense to "container:xxx")
If !container filter was given, only processes which don't belong to any container will be traced (like in "host=xxx").

To conclude all the filter options after the change:
```
--filter uid=uid1,uid2,...
--filter uid!=uid1,uid2,...
--filter pidns=pidns1,pidns2...
--filter pidns!=pidns1,pidns2...
--filter mntns=mntns1,mntns2...
--filter mntns!=mntns1,mntns2...
--filter uts=name1,name2,....
--filter uts!=name1,name2,....
--filter comm=comm1,comm2,...
--filter comm!=comm1,comm2,...
--filter event=event1,event2,...
--filter event!=event1,event2...
--filter set=set1,set2,...
--filter pid=pid1,pid2...
--filter pid!=pid1,pid2...
--filter container
```

Edit: 
[...]

PR #395, opened 2021-01-04
Change trace modes and filter UX

yanivagman:
This PR introduces the changes as described in #377:

1. Tracing modes are simplified to: new/all/follow/pidns
2. pidns mode keeps the "container:new" semantics from before, meaning, tracing when "new container" is started. This name better describes how we traced new containers in the past - by looking for the creation of new pid namespaces (execution of pid 1).
3. Boolean container filter is added. Opposed to pidns (which is equivalent to container:new), filtering by container (with default 'new' trace mode) will trace all new processes created in any container, including existing containers. Using container filter with 'all' trace mode is equivalent to the old "container:all" mode.
4. move "--events" "--exclude-events" and "--set" flags into "--filter" flag

### DIFF
--- a/main.go
+++ b/main.go
@@ -141,14 +141,14 @@ func main() {
 			&cli.StringFlag{
 				Name:    "trace",
 				Aliases: []string{"t"},
-				Value:   "process:new",
-				Usage:   "set trace mode, whether to trace processes or containers, and if to trace new, all, or specific processes/container. run '--trace help' for more info",
+				Value:   "new",
+				Usage:   "trace mode: new/pidns/follow/all. run '--trace help' for more info",
 			},
 			&cli.StringSliceFlag{
 				Name:    "filter",
 				Aliases: []string{"f"},
 				Value:   nil,
-				Usage:   "set tracing filters for specific fields (such as UID or GID). run '--filter help' for more info.",
+				Usage:   "set tracing filters for specific fields (such as UID or PID). run '--filter help' for more info.",
 			},
 			&cli.BoolFlag{
 				Name:  "detect-original-syscall",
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -225,34 +200,40 @@ func main() {
...
-	filterHelp += "\t--filter pidns=12345678                                       | only trace events from pidns 12345678\n"
-	filterHelp += "\t--filter pidns!=12345678                                      | don't trace events from pidns 12345678\n"
-	filterHelp += "uts: only trace processes or containers with specified uts namespace(s) name.\n"
-	filterHelp += "\t--filter uts=8215606f23f4                                     | only trace events from uts 8215606f23f4\n"
-	filterHelp += "\t--filter uts!=ab356bc4dd554                                   | don't trace events from uts ab356bc4dd554\n"
-	filterHelp += "comm: only trace processes with specified command name.\n"
-	filterHelp += "\t--filter comm=ls                                              | only trace events from ls command\n"
-	filterHelp += "\t--filter comm!=ls                                             | don't trace events from ls command\n"
+	filterHelp += "Note: some of the above operators have special meanings in different shells\n"
+	filterHelp += "      to 'escape' those operators, please use single quotes, e.g.: 'uid>0'\n"
```
itaysk, line 236: can you please replace these string constructions with a raw string? (between `)

main.go
```diff
@@ -225,34 +200,40 @@ func main() {
 }
 
 func prepareFilter(filters []string) (tracee.Filter, error) {
-
 	filterHelp := "\n--filter allows you to specify values to match on for fields of traced events.\n"
-	filterHelp += "The following options are currently supported:\n"
-	filterHelp += "uid: only trace processes or containers with specified uid(s).\n"
+	filterHelp += "\n"
+	filterHelp += "uint filters use the following operators: '=', '!=', '<', '>'\n"
```
itaysk, line 205: I think it's better to rename "uint filter" to "numerical filter" it doesn't matter the actual type, just that it's a number and can be treated as a number in the filter predicate

main.go
```diff
@@ -225,34 +200,40 @@ func main() {
...
+	filterHelp += "available uint filters: uid, pid, mntns, pidns\n"
+	filterHelp += "Examples:\n"
```
itaysk, line 207: I think it's better to put all the examples at the end. if someone it reading this from start to end they see a block of examples and don't keep reading to see there a more filter types

main.go
```diff
@@ -301,51 +282,94 @@ func prepareFilter(filters []string) (tracee.Filter, error) {
...
+		// To avoid collisions between filters that share the same prefix, put the filters which should have an exact match first!
+		if filterName == "comm" {
+			err := parseStringFilter(operatorAndValues, filter.CommFilter)
+			if err != nil {
+				return tracee.Filter{}, err
+			}
+			continue
+		}
+
+		if strings.HasPrefix("container", f) || (strings.HasPrefix("!container", f) && len(f) > 1) {
```
itaysk, line 315: if `strings.HasPrefix("!container", f)` then for sure `len(f) > 1`, isn't it?

tracee/tracee.go
```diff
@@ -64,15 +65,20 @@ type StringFilter struct {
 	Enabled  bool
 }
 
+type BoolFilter struct {
+	Set     bool
```
itaysk, line 69: I think "set" is confusing in this context. in other place (in flag parsing) we use it to mean "enabled". can we change it to "value" or something?

itaysk: we also might want to consider renaming enabled to "isSet" like in flag parsing but not a must

itaysk (commented):
needs tests as well for the new filters

itaysk:
I'm not sure pidns fits under --trace along with new, all, follow. It's not the same category, pidns is a filter, not a mode, so I think it should be under --filter, like `-f pidns=new`

main.go
yanivagman: You mean string builder, right?

main.go
yanivagman: No. "!" filter

tracee/tracee.go
```diff
@@ -169,14 +175,14 @@ func New(cfg TraceeConfig) (*Tracee, error) {
 	t := &Tracee{
 		config: cfg,
 	}
-	ContainerMode := (t.config.Mode == ModeContainerAll || t.config.Mode == ModeContainerNew)
+	ContainerMode := t.config.Filter.ContFilter.Enabled && t.config.Filter.ContFilter.Set
```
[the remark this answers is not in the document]
yanivagman: Pidns is not related to container (although containers create a new pid ns, and that what we had until now)

yanivagman:
About pidns. It's complicated.
Our previous "container:new" mode was based on it, and on the fact that new containers mean new pidns.
I kept the semantics of pidns (as was for container:new) because it's mutually exclusive with new/all/follow, and can't work with them. I also had my thought about if I should keep it or not, so I decided to keep it for now, and postpone the decision to a future commit

main.go
itaysk: no. theres not string to build here at runtime, it's just one big string in the source code.

```go
s:= `this is 
a raw string`
```

main.go
itaysk: I don't understand... `f` is a string that starts with `!container`, how can it not be longer than 1?

tracee/tracee.go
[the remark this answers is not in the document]
itaysk: pidns means that I want to capture events only from new pidns's right?

main.go
yanivagman: It's the other way around. It checks if "f" is a prefix of "container" or "!container", so we can filter with prefix: "--filter c" or "--filter !c"

tracee/tracee.go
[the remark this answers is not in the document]
yanivagman: Yes. It's different from "container" flag, as "container" flag is true for all conatiner processes, including those that existed in pid namespaces that are not new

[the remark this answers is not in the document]
itaysk: ok. so we do need to add pidns here. why do you think we don't?

main.go
itaysk: ahh right

tracee/tracee.go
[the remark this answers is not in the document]
yanivagman: As I said, I kept pidns only to be able to trace NEW containers, which is not possible with the container flag, otherwise, I would remove it from other places too.
Also, a process unsharing it's pid namespace is not always a container. (actually, a mount namespace fits much better as a description for a container than pid namespace). We can change the way "container" filter is working in the future, but pidns stands for itself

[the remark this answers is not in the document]
itaysk: Your answer is missing the context of this discussion. we are talking about this particular line right? all I'm saying is that when printing the table, we should show host ns information if tracing pidns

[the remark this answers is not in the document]
yanivagman: The context of this particular line is **ContainerMode**, how am I missing the context?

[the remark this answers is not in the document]
itaysk: it's just the argument to `newEventPrinter`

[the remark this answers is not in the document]
yanivagman: Well, this argument also controls the printing of UTS_NAME and MNT_NS, which doesn't exactly fit to PIDNS. I don't mind doing this change though

itaysk (approved):
squash reminder

PR #396, opened 2021-01-05
simplify finding data

itaysk:
cc: @roikol


itaysk:
if you know what the sig returns, you look it up in the map. if you don't know what the sig returns, knowing the "type" won't help you

yanivagman (approved):

Issue #29, opened 2019-12-19
Additional filtering


lizrice:
I'd like to be able to filter by UTS name field. 

Option 1: add `--u <uts_name>[,<uts_name2>...]` / `--uts-name <uts_name1>[,<uts_name2>...]` option which would only output a line if it matches one of the specified UTS names

Option 2: add `-f <field name>=<value1>[,<value2>...]` / `--filter <field name>=<value>[,<value2>...]` which would only output a line if the field matches the specified values. If multiple filters are specified, they are ANDed together (so the line only gets matched if all filters have a match). 

Option 1 is easier to both implement and use, and we already have `--event` for filtering on particular events, so I'm leaning towards that. It wouldn't rule out option 2 for the future. Wdyt?

icarus-sparry:
I would use `awk` (for table output) or `jq` (for json output) to post-process the output. Not quite as efficient but then you have a full programming language available so you can say "if it is for this utsname or this utsname then do this, else if it is for this utsname and it is an openat system call do this else ...". Whilst `find` has proved you can write arbitrary complex expressions on the command line, no one actually wants to do it.

If this enhancement is accepted then I suggest using `find` as a model and allowing both `and` and `or` operators together with grouping. Having -f name=value1,value2 as an abbreviation for `\( -f name=value1 --or -f name=value2 \)` would be nice but then the rules for escaping commas might get non obvious.

itaysk:
I think tracee should just collect/generate events, and not process them which better be done with specialized tools like you mentioned @icarus-sparry. But I do think that @lizrice has a point specifically about hostname filtering, or if can generalize - container filtering, which I think is n important use case (I want to trace only container X) and can also be filtered in kernel. Container filtering is also discussed in https://github.com/aquasecurity/tracee/issues/21.

PR #403, opened 2021-01-07
Add support for filtering an event by its argument

### LINKED ISSUE #380: Add event argument filters
We already have the "filter" flag which is able to filter events based on some given values.
This filtering is done for ALL events which need to pass all the filters.
We should add support to filter by the argument value of ONE specific event, which will have no effect on the other events.
This can be done by using the following format:
`--filter eventName:argName=xxx`
For example, say we want to filter all execve events which have "pathname: /usr/bin/ls", we can do that as follows:
`--filter execve:pathname=/usr/bin/ls`
All non-execve events will not be affected, while execve events will be filtered according this pathname.

This feature shouldn't be implemented as part of the bpf code as the bpf code is not aware of the argument names of the events, and it might also complicate the logic in the bpf side without any real performance gains.
Instead, we can implement this in the shouldProcessEvent() function, which happens before any processing of the event takes place.

yanivagman:
close #380

### DIFF
--- a/main.go
+++ b/main.go
@@ -212,6 +212,13 @@ available string filters: event, set, uts, comm
 Boolean filters use the following operator: '!'
 available boolean filters: container
 
+Event argument filters are used to filter a specific event by its arguments.
+The following format should be used: "event_name:event_arg=val" or "event_name.event_arg!=val"
+
+Non-boolean filters can get multiple values using "," as a separator.
+When used with the equality operator ("="), the values given to the filter are logical ORed.
+When given with any other operator, the values are logical ANDed.
+
 Note: some of the above operators have special meanings in different shells.
 To 'escape' those operators, please use single quotes, e.g.: 'uid>0'
 
@@ -233,6 +240,8 @@ Examples:
 	--filter c                                                    | only trace events from containers (same as above)
 	--filter !container                                           | only trace events from the host
 	--filter '!c'                                                 | only trace events from the host (same as above)
+	--filter close.fd=5                                           | only trace 'close' events that have 'fd' equals 5
+	--filter openat.pathname!=/tmp/1,/bin/ls                      | don't trace 'openat' events that have 'pathname' equals /tmp/1 or /bin/ls
 
 `
 
@@ -247,7 +256,6 @@ Examples:
 			Less:     tracee.LessNotSet,
 			Greater:  tracee.GreaterNotSet,
 			Is32Bit:  true,
-			Enabled:  false,
 		},
 		PIDFilter: &tracee.UintFilter{
 			Equal:    []uint64{},
@@ -255,43 +263,41 @@ Examples:
 			Less:     tracee.LessNotSet,
 			Greater:  tracee.GreaterNotSet,
 			Is32Bit:  true,
-			Enabled:  false,
 		},
 		MntNSFilter: &tracee.UintFilter{
 			Equal:    []uint64{},
 			NotEqual: []uint64{},
 			Less:     tracee.LessNotSet,
 			Greater:  tracee.GreaterNotSet,
-			Is32Bit:  false,
-			Enabled:  false,
 		},
 		PidNSFilter: &tracee.UintFilter{
 			Equal:    []uint64{},
 			NotEqual: []uint64{},
 			Less:     tracee.LessNotSet,
 			Greater:  tracee.GreaterNotSet,
-			Is32Bit:  false,
-			Enabled:  false,
 		},
 		UTSFilter: &tracee.StringFilter{
 			Equal:    []string{},
 			NotEqual: []string{},
-			Enabled:  false,
 		},
 		CommFilter: &tracee.StringFilter{
 			Equal:    []string{},
 			NotEqual: []string{},
-			Enabled:  false,
 		},
-		ContFilter: &tracee.BoolFilter{
-			Value:   false,
-			Enabled: false,
+		ContFilter: &tracee.BoolFilter{},
+		ArgFilter: &tracee.ArgFilter{
+			Filters: make(map[int32]map[string]tracee.ArgFilterVal),
 		},
 		EventsToTrace: []int32{},
 	}
 
-	eventFilter := &tracee.StringFilter{Equal: []string{}, NotEqual: []string{}, Enabled: false}
-	setFilter := &tracee.StringFilter{Equal: []string{}, NotEqual: []string{}, Enabled: false}
+	eventFilter := &tracee.StringFilter{Equal: []string{}, NotEqual: []string{}}
+	setFilter := &tracee.StringFilter{Equal: []string{}, NotEqual: []string{}}
+
+	eventsNameToID := make(map[string]int32, len(tracee.EventsIDToEvent))
+	for _, event := range tracee.EventsIDToEvent {
+		eventsNameToID[event.Name] = event.ID
+	}
 
 	for _, f := range filters {
 		filterName := f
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -233,6 +236,8 @@ Examples:
 	--filter c                                                    | only trace events from containers (same as above)
 	--filter !container                                           | only trace events from the host
 	--filter '!c'                                                 | only trace events from the host (same as above)
+	--filter close:fd=5                                           | only trace 'close' events that have 'fd' equals 5
+	--filter openat:pathname!=/tmp/1,/bin/ls                      | don't trace 'openat' events that have 'pathname' equals /tmp/1 or /bin/ls
```
itaysk, line 240: I see that you allow multiple values in the same filter.
1. this is not mentioned in the help (outside the example)
2. it should be clear what `,` means - AND/OR. perhaps replace it with a symbol that better reflects
3. I think it would be better to not support multiple values, just like we don't allow for other filters. if someone wants to filter multiple values, they can specify the filter multiple times (like with other filters)

main.go
```diff
@@ -212,6 +212,9 @@ available string filters: event, set, uts, comm
 Boolean filters use the following operator: '!'
 available boolean filters: container
 
+Event argument filters are used to filter a specific event by its arguments.
+The following format should be used: "event_name:event_arg=val" or "event_name:event_arg!=val"
```
itaysk, line 216: don't you think `.` is more intuitive than `:` when specifying lookup?

main.go
```diff
@@ -481,6 +491,73 @@ func parseBoolFilter(value string, boolFilter *tracee.BoolFilter) error {
...
+	if !ok {
+		return fmt.Errorf("invalid argument filter event name: %s", eventName)
+	}
+
+	eventParams, ok := tracee.EventsIDToParams[id]
+	if !ok {
+		return fmt.Errorf("invalid argument filter event name: %s", eventName)
+	}
+
+	// check if argument name exists for this event
```
itaysk, line 515: this should probably be validated in package tracee as well (instead or in addition to this)

tracee/tracee.go
```diff
@@ -71,6 +72,21 @@ type BoolFilter struct {
 	Enabled bool
 }
 
+type ArgFilter struct {
+	Filters map[ArgFilterKey]ArgFilterVal
```
itaysk, line 76: looks like you've chosen a design that's similar to the ebpf but since we are evaluating this in go it could be simper, just compare strings. This will allow us to implement filters like startswith which is not really important for filtering on context fields, but for filtering on args it's very practical (e.g pathname starts with /etc)

main.go
yanivagman: Going backward:
3. All of our filters support multiple values: uid, pid, comm, uts... Why do you say they don't?
2. That depends. It leads me back to the discussion with @grantseltzer about the expected filter behavior. If '=' is chosen, it means logical OR, if '!=' is chosen it means logical AND. This is also why ',' suits here best
1. The example is the help. I can add a comment if you think that is necessary

main.go
yanivagman: Good idea. I'll do that.
This also made me find a small bug.
I checked if we don't have an event or argument with a dot in it, and I found the following:
{Type: "struct sched_param*", Name: "external.ArgMeta"}},
For 3 different system calls: SchedSetparamEventID, SchedGetparamEventID, SchedSetschedulerEventID
Probably happened during renaming of param to external.ArgMeta

tracee/tracee.go
yanivagman: I did want to check by prefix by decided to postpone it for a later commit. We also need it for vfs_write events filtering to get rid of the --filter-file-write - but that's another story.
I chose this implementation method because it is the most efficient one. Naive string comparison will require to loop over the available argfilters for every event received, which is not efficient. Let's keep it efficient for now and think about how to tackle the prefix issue in a different commit

main.go
grantseltzer: I do agree that ',' is cleaner looking but I can't think of other tools that have a similar interface where they wouldn't have to read the docs anyway.

grantseltzer:
Do you guys think this is starting to scope creep into other unix tools that are traditional used for filtering like this? If so, would it make more sense to expose these filtering options via the Go package but not via CLI?

itaysk:
@grantseltzer I tend to agree, this was also my original approach when filtering was initially discussed: https://github.com/aquasecurity/tracee/issues/29#issuecomment-654634416
it generally depends on the use case - if we plan people to use Tracee in a certain way and do to certain things that will really benefit from implementing a feature that would otherwise be provided by another tool, then we can add it to Tracee. In this case we never had this discussion but I guess we can excuse it by being a common task, e.g filter by pathname

tracee/tracee.go
itaysk: actually, since we operate on the `RawEvent`, you have the `RawArgs` map so don't need to iterate, right?

main.go
itaysk: if you're keeping `,`. can you clarify the behavior in the help?

itaysk: > The example is the help. I can add a comment if you think that is necessary

The example should demonstrate what's defined in the help. I do think it should be clearly defined when discussing the filter

yanivagman:
I'll add more points to what Itay said:
1. Performance. Tracee can output thousands of events in one second. Performing the filtering in an early stage (in this case at the start of the pipeline) is important for performance reasons.
2. This filter is not just a simple grep. All events which are not THE event defined by the filter will act as usual, while only the event defined by the filter will be filtered according to one or more of its arguments. While it is probably possible to do this with other unix tools, I don't think the common user of Tracee can easily grep it.
3. For writing rules which are based on events from tracee, it can be helpful to configure tracee for certain events, without the use of external tools

tracee/tracee.go
yanivagman: Not really. Say that you gave a filter to filter all the pathname given to execve. Then you will have to compare "pathname" (deoded from the actual argument) against all the argument filters given. Here a map is used to avoid this loop

itaysk: I'm not looking at the code, but iirc you can encode the value into argtag, then look it up in the RawArgs map, no iteration involved, right?

yanivagman: So you suggest iterating over all given arg filters instead of iterating the event's arguments?

itaysk: yes. it will probably be less iterations on average, but that's not my point. I suggest this only if you think it will make the code simpler considering this in go and considering your future plans for arg filtering

yanivagman: But doesn't this loose the purpose you started with? To compare by string or prefix of strings?
If we need to encode the value into argtag, a prefix of the string will not work

yanivagman: Not sure I understand what do you get by changing what you loop on.
I DO compare strings! Have a look at lines 845 and 855

itaysk: yes you are right it's actually just the order or iterations (still makes more sense to iterate on filters first, but not a big deal)

yanivagman: Yes, it will also simplify things - I'm on it

tracee/tracee.go
```diff
@@ -812,6 +828,38 @@ func (t *Tracee) handleError(err error) {
...
+			argMeta, ok := t.DecParamName[e.Ctx.EventID%2][tag]
+			if !ok {
+				continue
+			}
+			key := ArgFilterKey{
+				EventId: e.Ctx.EventID,
+				ArgName: argMeta.Name,
+			}
+			if argFilter, ok := t.config.Filter.ArgFilter.Filters[key]; ok {
+				argVal := fmt.Sprint(e.RawArgs[tag])
```
itaysk, line 843: not a clocker, but I have a feeling that Sprint will be much less efficient than type assertion. I normally wouldn't care but since this is happening for every filter check, maybe better to replace.

yanivagman: I know, but assertion will require to handle all argument types differently, and I didn't want to get into it now

itaysk (approved):

Issue #398, opened 2021-01-05
Print event arguments when listing available events


yanivagman:
Listing available events (using --list flag) currently show the event name and the sets it belongs to.
As we plan to add argument filters (#380), it can be helpful to also list the event arguments types and names.
This can be done using the EventsIDToParams table found in consts.go, which already contains all the necessary data

PR #407, opened 2021-01-07
Issue-398 add arguments to events

mtcherni95:
This commit provides the ability to display the system calls arguments when tracee runs with --list flag. #398

### DIFF
--- a/main.go
+++ b/main.go
@@ -579,28 +579,58 @@ func getSelfCapabilities() (capability.Capabilities, error) {
 	return cap, nil
 }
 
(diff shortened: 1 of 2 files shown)


yanivagman (changes requested):
Thanks @mtcherni95 !

Please rebase your changes on main.
EventsIDToParams table has been changed and pType and pName are now exported as Type and Name. Probably after rebasing, most of your changes won't be needed, only those that change the printList() function.

yanivagman (approved):
Thanks @mtcherni95

PR #409, opened 2021-01-10
fixed bug in filtering by comm or uts_name

pathtofile:
Recent changes broke the ability to filter by uts_name or comm.

You can confirm this by running `trace --filter comm=whoami` then running a `whoami`.

This bug was created when the shift to using `comm` as the key into the `comm_filter` map - the call to lookup the key is essentially a mem compare, but the filter string 32-Bytes (`MAX_STR_FILTER_SIZE`), but the com name in the context struct was only 16, making the compare/lookup alway fail.

Currently only `comm` and `uts_name` use the string filter option, both of which are only 16-Bytes long, so it's OK to keep the size of `string_filter` to this. But if something else comes along later that wants to filter on something more than that, then the code will have to change, probably to copy the string-to-compare to the larger array, prior to doing the key lookup.

### DIFF
--- a/main.go
+++ b/main.go
@@ -455,8 +455,8 @@ func parseStringFilter(operatorAndValues string, stringFilter *tracee.StringFilt
 	values := strings.Split(valuesString, ",")
 
 	for i := range values {
-		if len(values[i]) > 32 {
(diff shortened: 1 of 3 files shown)


itaysk:
thanks @pathtofile ! this is a dup of #402

pathtofile:
ah oops, sorry!

PR #411, opened 2021-01-11
Move follow trace mode to filter flag

yanivagman:


main.go
```diff
@@ -234,6 +234,10 @@ Examples:
 	--filter !container                                           | only trace events from the host
 	--filter '!c'                                                 | only trace events from the host (same as above)
 
+
+Processes for which all of the above filters evaluate to true can be followed so that all of their children will be traced as well:
```
itaysk, line 238: 1. better to move this before the examples
2. can we rephrase this?
3. I think the first line of the help is not great (not related to your pr)

so I suggest to rewrite the help intro as:

```
--filter lets you preclude events that match a criteria from being traced. The following types of filter expressions are supported:

// description of filter types, each followed by filters

It is possible to include events that descended from filtered processes using the special 'follow' filter. Specifying this filter will trace events from any filtered process, or a child process of a filtered process.

Examples:
// list of examples
--filter comm=bash --filter follow | tracee all events that originated from bash or from one of the processes spawned by bash
```

wdyt? (specifically about the structure)

itaysk (approved):

PR #414, opened 2021-01-12
add support for rego signatures

itaysk:


itaysk:
@yanivagman note that I've added another commit just now, it's a small one you can review it after the original bulk

yanivagman (approved):

PR #416, opened 2021-01-13
improve readme

itaysk:


README.md
```diff
@@ -17,16 +17,15 @@ Other than tracing, Tracee is also capable of capturing files written to disk or
 
 ### Prerequisites
 
-- To run, Tracee requires Linux kernel version >= 4.14
-
-Not required if using the Docker image:
-- C standard library (tested with glibc)
+- Linux kernel version >= 4.14
+- Relevant kernel headers available under `/usr/src`
```
yanivagman, line 21: This is not true for arch (/manjaro), where the headers (not links) are located at /lib/modules/$(KERN_RELEASE)/build

README.md
```diff
@@ -17,16 +17,15 @@ Other than tracing, Tracee is also capable of capturing files written to disk or
...
+- C standard library (tested with `glibc` and `musl`)
 - `libelf` and `zlib` libraries
 - clang >= 9
 
-Not required if pre-compiling the eBPF code (see [Installation options](#installation-options)):
-- clang >= 9
-- Kernel headers available under `/usr/src`, must be provided by user and match the running kernel version, not needed if building the eBPF program in advance
+Exceptions:
+- Tracee supports loading a pre-compiled eBPF file, in which case the kernel headers are required only for the one-time compilation, and not at runtime. See [Setup Options](#setup-options) for more info.
```
yanivagman, line 27: clang is not needed in that case as well

README.md
yanivagman: Actually, the location is not a requirement at all as it can be set by environment variables

itaysk: maybe I change this to :
Relevant kernel headers available under a conventional location?
?

lizrice: Let's not lose info - it was useful to me to be able to check that there was nothing in `/usr/src` and that was likely my problem. How about saying something like "Relevant kernel headers available under `/usr/src` for most distributions (`/lib/modules/$(KERN_RELEASE)/build` for arch / manjaro)" 

We also need to tell users how to fulfill this requirement. Even if we just say "For example, on Ubuntu, run `sudo apt-get install linux-headers-$(uname -r)`" that gives a hint for what you'd have to do on other distro's

README.md
lizrice: This now implies everyone needs to install C standard library, libel, zlib, clang if they are using the docker image?

README.md
```diff
@@ -17,16 +17,15 @@ Other than tracing, Tracee is also capable of capturing files written to disk or
...
+- When using Tracee's Docker image, all of the aforementioned requirements are built into the image. The only requirement left is the kernel headers or the pre-built eBPF. See [Setup Options](#setup-options) for more info.
```
lizrice, line 28: Ah, OK I see this is mentioned here... I wonder if it would make sense to structure something like this: 

```
=== Prerequisites

* Tracee requires Linux kernel version >= 4.14
* Kernel headers matching the running kernel available under `/usr/src`...(as above)

There are additional dependencies if you want to build Tracee and its eBPF code yourself - see <some link> below

=== Quickstart with Docker

docker run --name tracee --rm --privileged --pid=host -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro -v /tmp/tracee:/tmp/tracee aquasec/tracee
Note: You may need to change the volume mounts for the kernel headers based on your setup.

This will run Tracee with no arguments, which defaults to collecting all events from all newly created processes and printing them in a table to standard output. See below for additional configuration options

=== Building and running Tracee

Tracee is made of an executable....(text as now)
```
and then insert the additional dependencies on libs and clang where they are needed (I think clang is only needed when building the eBPF code; I'm not sure about libelf and zlib) - that is, include info on the fact they are needed just above the instructions for building that need them

README.md
yanivagman: Nope. These are not needed as well

README.md
yanivagman: > This will run Tracee with no arguments, which defaults to collecting all events from all...
Tracee doesn't collect all events by default (only the default set, as defined when listing the events with --list)

> I think clang is only needed when building the eBPF code; I'm not sure about libelf and zlib
clang (or gcc?) is also needed for tracee itself as it is using cgo. In additoin, libbpf also requires a C compiler.
libelf and zlib are required for the tracee binary (for libbpfgo)

itaysk: @lizrice sgtm, but the kernel headers requirement is part of the "additional dependencies if you want to build Tracee and its eBPF code yourself" group

itaysk: @lizrice actually.. this is not correct:

> There are additional dependencies if you want to build Tracee and its eBPF code yourself - see <some link> below

The extra requirements are also if you want tracee to build the ebpf code for you.

itaysk:
closing this as we have an upcoming PR that rewrites the entire readme

PR #421, opened 2021-01-14
Consolidate "trace" and "filter" flags

yanivagman:
This PR consolidates "trace" and "filter" flags and removes the tracing modes.
1. All previous mode functionality is now available as part of the trace flag.
2. All processes in the system are now traced by default (instead of the previous default which was tracing new processes).
3. To trace new processes, the trace expression "pid=new" should be used
4. To trace new containers, the trace expression "container=new" should be used
5. The events "vfs_write(v)" and "ioctl" are removed from the default set of events because we now trace all processes by default, and these events add a lot of noise to the output
6. To reduce noise when container trace expression is chosen, pid=new expression is also set implicitly. This is done due to the fact that we use pid namespace to decide if a processes belongs to a container, however, there are many processes in the system (e.g. systemd) which are in a different pid namespace than the host, which in combination with the new default of tracing all the processes in the system will be mistakenly traced as a container.


itaysk:
we've discussed the ambiguity of the word "filter" and now that I'm reading the help it's really standing out. We initially agreed that it's most natural for filter to specify exclusion. this is demonstrated by the opening, which is now becoming incorrect:

> filter lets you preclude events that match a criteria from being traced

Recently we discussed the UX refactor and you suggested to filter will specify which events to include, which this PR implements. Now not only we didn't change the opening, but the fact that the same flag can describe both contradictory settings, and we have to rely on the help to explain what it does means that its probably not the best word.

Now we have statements like this in the help:

> special value called 'new' which filters new containers

which has no meaning (what does it mean filters new containers, does it exclude or include them).

I know that we have agreed to keep it filter, but I really think we should reconsider this choice. To be clear I'm just talking about the name of the flag, not it's behavior.

itaysk:
wdyt of the folowing suggestion:

```
--trace lets you select which events to trace by defining trace expressions which operates on events or process metadata. Only events that match all trace expressions will be traced (trace flags are ANDed). The following types of expressions are supported:

Numerical expressions which compare numbers and allow the following operators: '=', '!=', '<', '>'.
Available numerical expressions: uid, pid, mntns, pidns.

String expressions which compares text and allow the following operators: '=', '!='.
Available string expressions: event, set, uts, comm.

Boolean expressions that check if a boolean is true and allow the following operator: '!'.
Available boolean expressions: container.

Event arguments can be accessed using 'event_name.event_arg'. Event arguments allow only the following operators: '=', '!='.

Non-boolean expressions can compare a field to multiple values separated by ','. Multiple values are ORed if used with equals operator '=', but are ANDed if used with any other operator.

The field 'container' and 'pid' also support the special value 'new' which selects new containers or pids, respectively.

The special 'follow' expression declares that not only processes that match the criteria will be traced, but also their descendants.
```

yanivagman:
I still think that the word "filter" best describes what this flag does, and that any other word (including "trace") will not fit. I also think that having one flag (instead of two) is the most intuitive and easy thing to do.
Let's agree (as we already did) that other than events/set, all the filters indeed reduce the amount of events - right?
Now let's talk about 'event' and 'set'. 'event' is always relative to set (which defaults to 'default'). We could have chosen to put it outside the filter flags , but in my opinion, it's more convenient and intuitive for it to be in the filter flag.
About the first sentence in the filter help - let me remind you that you asked me to change to this sentence in the previous PR.
If there is something that is not clear in the help description, we can always change this to better describe how the flag should be used.

yanivagman:
In the end it's just the name of the flag, I see that we do agree about all the rest...
Let's not waste the time discussing this - I'll change to trace (even though I think that filter is a better fit)

main.go
```diff
@@ -339,19 +334,32 @@ To 'escape' those operators, please use single quotes, e.g.: 'uid>0'
...
 				return tracee.Filter{}, err
 			}
 			continue
 		}
 
-		if filterName == "newpidns" || filterName == "!newpidns" {
-			err := parseBoolFilter(f, filter.NewPidNsFilter)
-			if err != nil {
-				return tracee.Filter{}, err
+		if strings.HasPrefix("container", filterName) {
```
itaysk, line 346: correct me if I'm wrong but I think that if `--trace container=new`, it will never reach this `if` because it will be caught in the previous `if strings.HasPrefix("container",f)`?

yanivagman: Nope. the previous if means that "f" should be a prefix of "container".
If "f" equals "container=new", it is not a prefix of "container"

itaysk (approved):

PR #423, opened 2021-01-17
Add prefix operator to argument filters

yanivagman:


itaysk (commented):
awesome!
sorry to be pedantic but perhaps a couple of tests?

main.go
```diff
@@ -210,7 +210,8 @@ Boolean expressions that check if a boolean is true and allow the following oper
 Available boolean expressions: container.
 
 Event arguments can be accessed using 'event_name.event_arg' and provide a way to filter an event by its arguments.
-Event arguments allow the following operators: '=', '!='.
+Event arguments allow the following operators: '=', '!=', '*'.
+The operator '*' defines a string prefix, and should be added to the end of the string to match.
```
itaysk, line 214: are you sure you want to refer to this as an "operator" in the help? even tough it's technically an operator behind the scenes, perhaps we should just say that strings can be compared as a prefix if ending with *?

yanivagman:
Tests for which function? This cannot be part of main_test.go as it is not found in main.go

itaysk (approved):

Issue #397, opened 2021-01-05
Expand output flag


yanivagman:
Today, we have the following output flags:
--output value, -o value            output format: table/table-verbose/json/gob/go-template=<path> (default: "table")
--output-path value                 path where tracee will save produced artifacts (default: "/tmp/tracee/out")

Non of these flags control where the output of tracee will be directed to, which defaults to stdin/stderr.
Although the user can redirect the output using shell redirection, ('>', '2>'), this approach is not applicable to programs which call Tracee programatically, or when Tracee is running in a detached docker container.
A better approach will be to let the user specify an output file path, where Tracee will save the output to.

Currently, "--output" flag actually controls output formatting and "output-path" controls the path to save the produced artifacts of tracee. We can expand the output flag to look as follows:
```
--output format:table/table-verbose/json/gob/go-template=<path>
--output artifacts:<path>
--output trace:<path>
```
The above names are just a suggestion, and are open to discussion

itaysk:
I generally agree.

1. what is `trace:path`? isn't it the same as `format:$format=path`?
2. we also need a way to signify output options, like EOT https://github.com/aquasecurity/tracee/pull/379, or stack trace https://github.com/aquasecurity/tracee/pull/375
3. I think we should also use this for output customization. for example, we have an issue for selecting table columns https://github.com/aquasecurity/tracee/issues/207, instead of more flags we can make it a customization of the table format. 
4. we now split the artifacts configuration between the capture and output flag.

suggestion:
- `-o [format:]table=comma,separated,format,options` (default: table)
  - `-o table=pid,cmd`
  - `-o table=all` (same as verbose)
  - `-o gotemplate=/path/to/template`
  - `-o json=ident`
- `-o file:/path/to/file`. if unset write to stdout. (default: unset)
- `-o option:comma,separated,options`
  - `-o option:eot`. send an EOT event before exiting.
  - `-o option:stacktrace`. enable stack trace

- `--capture [mode:]exec/mem/write` (default:none)
- `--capture dir:/path/to/dir` (default:/tmp/tracee)
- `--capture clear-dir` (default: false)

WDYT?

yanivagman:
1. The <path> in the format field is for g`o-template=<path>`. This field only contains the format. `trace:path` is for the output of tracee ("traces"). I wanted to differ between this kind of output and the artifacts output, but I think your suggestion is better. 
2. Agree
3. Agree
4. I take your suggestion and I raise :-)

[...]

PR #425, opened 2021-01-18
Change capture ux

### LINKED ISSUE #419: Decouple and remove filter-file-write flag
Currently, filter-file-write flag controls the filtering of two different things:
1. vfs_write(v) events which are output of tracee
2. capture files when "--capture write" is given

These two filters should not necessarily be coupled, and a user should be able to filter specific vfs_write events, while capturing different file writes.

As we now have argument filters, it is possible to decouple these filters as follows:
1. `--filter vfs_write.pathname=/path/prefix1`
2. `--capture write:/path/prefix2`

To do that, it is needed to add prefix support for argument filters (which can be done using '\*', for example, /path/prefix/*)
It is also needed to add filtering support to `--capture write`
Then, we can also remove the filter-file-write flag

yanivagman:
close #419

### DIFF
--- a/main.go
+++ b/main.go
@@ -46,7 +46,6 @@ func main() {
 				OutputFormat:          c.String("output"),
 				PerfBufferSize:        c.Int("perf-buffer-size"),
 				BlobPerfBufferSize:    c.Int("blob-perf-buffer-size"),
-				OutputPath:            c.String("output-path"),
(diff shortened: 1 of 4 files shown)


itaysk (approved):
minor comments

yanivagman:
This PR also added the capture features as described in #397

PR #427, opened 2021-01-19
lint: Address a few idiomatic Go improvements

simar7:

### DIFF
--- a/tracee-ebpf/tracee/tracee.go
+++ b/tracee-ebpf/tracee/tracee.go
@@ -19,8 +19,8 @@ import (
 	"github.com/aquasecurity/tracee/tracee/external"
 )
 
-// TraceeConfig is a struct containing user defined configuration of tracee
-type TraceeConfig struct {
+// Config is a struct containing user defined configuration of tracee
+type Config struct {
 	Filter             *Filter
 	Capture            *CaptureConfig
 	Output             *OutputConfig
@@ -119,7 +119,7 @@ func (cfg OutputConfig) Validate() error {
 }
 
 // Validate does static validation of the configuration
-func (tc TraceeConfig) Validate() error {
+func (tc Config) Validate() error {
 	if tc.Filter.EventsToTrace == nil {
 		return fmt.Errorf("eventsToTrace is nil")
 	}
@@ -176,7 +176,7 @@ func (tc TraceeConfig) Validate() error {
 
 // Tracee traces system calls and system events using eBPF
 type Tracee struct {
-	config            TraceeConfig
+	config            Config
 	eventsToTrace     map[int32]bool
 	bpfModule         *bpf.Module
 	eventsPerfMap     *bpf.PerfBuffer
@@ -216,8 +216,8 @@ type statsStore struct {
 	lostWrCounter counter
 }
 
-// New creates a new Tracee instance based on a given valid TraceeConfig
-func New(cfg TraceeConfig) (*Tracee, error) {
+// New creates a new Tracee instance based on a given valid Config
+func New(cfg Config) (*Tracee, error) {
 	var err error
 
 	err = cfg.Validate()
@@ -410,17 +410,17 @@ func supportRawTP() (bool, error) {
 	if ver == "" {
 		return false, fmt.Errorf("could not determine current release")
 	}
-	ver_split := strings.Split(ver, ".")
-	if len(ver_split) < 2 {
+	verSplit := strings.Split(ver, ".")
+	if len(verSplit) < 2 {
 		return false, fmt.Errorf("invalid version returned by uname")
 	}
-	major, err := strconv.Atoi(ver_split[0])
+	major, err := strconv.Atoi(verSplit[0])
 	if err != nil {
-		return false, fmt.Errorf("invalid major number: %s", ver_split[0])
+		return false, fmt.Errorf("invalid major number: %s", verSplit[0])
 	}
-	minor, err := strconv.Atoi(ver_split[1])
+	minor, err := strconv.Atoi(verSplit[1])
 	if err != nil {
-		return false, fmt.Errorf("invalid minor number: %s", ver_split[1])
+		return false, fmt.Errorf("invalid minor number: %s", verSplit[1])
 	}
 	if ((major == 4) && (minor >= 17)) || (major > 4) {
 		return true, nil
@@ -638,19 +638,19 @@ func (t *Tracee) populateBPFMaps() error {
 	if err != nil {
 		return fmt.Errorf("error getting BPF program trace_ret_vfs_write_tail: %v", err)
 	}
-	bpfProgArrayMap.Update(uint32(tailVfsWrite), uint32(prog.GetFd()))
+	bpfProgArrayMap.Update(tailVfsWrite, uint32(prog.GetFd()))
 
 	prog, err = t.bpfModule.GetProgram("trace_ret_vfs_writev_tail")
 	if err != nil {
 		return fmt.Errorf("error getting BPF program trace_ret_vfs_writev_tail: %v", err)
 	}
-	bpfProgArrayMap.Update(uint32(tailVfsWritev), uint32(prog.GetFd()))
+	bpfProgArrayMap.Update(tailVfsWritev, uint32(prog.GetFd()))
 
 	prog, err = t.bpfModule.GetProgram("send_bin")
 	if err != nil {
 		return fmt.Errorf("error getting BPF program send_bin: %v", err)
 	}
-	bpfProgArrayMap.Update(uint32(tailSendBin), uint32(prog.GetFd()))
+	bpfProgArrayMap.Update(tailSendBin, uint32(prog.GetFd()))
 
 	// Set filters given by the user to filter file write events
 	fileFilterMap, _ := t.bpfModule.GetMap("file_filter")
@@ -765,7 +765,7 @@ func (t *Tracee) initBPF(bpfObjectPath string) error {
 
 	supportRawTracepoints, err := supportRawTP()
 	if err != nil {
-		return fmt.Errorf("Failed to find kernel version: %v", err)
+		return fmt.Errorf("failed to find kernel version: %v", err)
 	}
 
 	// BPFLoadObject() automatically loads ALL BPF programs according to their section type, unless set otherwise
@@ -809,7 +809,7 @@ func (t *Tracee) initBPF(bpfObjectPath string) error {
 		return err
 	}
 
-	for e, _ := range t.eventsToTrace {
+	for e := range t.eventsToTrace {
 		event, ok := EventsIDToEvent[e]
 		if !ok {
 			continue
@@ -1137,8 +1137,8 @@ func (t *Tracee) prepareArgsForPrint(ctx *context, args map[argTag]interface{}) 
 			}
 		}
 		if ctx.EventID == CapCapableEventID {
-			if cap, isInt32 := args[t.EncParamName[ctx.EventID%2]["cap"]].(int32); isInt32 {
-				args[t.EncParamName[ctx.EventID%2]["cap"]] = PrintCapability(cap)
+			if capability, isInt32 := args[t.EncParamName[ctx.EventID%2]["cap"]].(int32); isInt32 {
+				args[t.EncParamName[ctx.EventID%2]["cap"]] = PrintCapability(capability)
 			}
 		}
 	case MmapEventID, MprotectEventID, PkeyMprotectEventID:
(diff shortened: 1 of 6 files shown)


tracee/tracee.go
```diff
@@ -765,7 +765,7 @@ func (t *Tracee) initBPF(bpfObjectPath string) error {
 
 	supportRawTracepoints, err := supportRawTP()
 	if err != nil {
-		return fmt.Errorf("Failed to find kernel version: %v", err)
```
itaysk, line 768: The errors that I write are usually lowercased, as mentioned here: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

we do have places where errors are mistakenly styles like a sentence, if you want you could change those instead

tracee/tracee_test.go
```diff
@@ -120,16 +120,40 @@ func TestReadArgFromBuff(t *testing.T) {
...
-		_, actual, err := readArgFromBuff(bytes.NewReader(tc.input))
+		b := bytes.NewReader(tc.input)
+		_, actual, err := readArgFromBuff(b)
 		assert.Equal(t, tc.expectedError, err, tc.name)
 		assert.Equal(t, tc.expectedArg, actual, tc.name)
+
+		// TODO: the following fails for:
+		// strArrT
+		// unknown
+		// sockAddrT - AF_UNIX
```
itaysk, line 156: can you please open an issue for this instead of code comment?

tracee/tracee.go
```diff
@@ -638,19 +638,19 @@ func (t *Tracee) populateBPFMaps() error {
 	if err != nil {
 		return fmt.Errorf("error getting BPF program trace_ret_vfs_write_tail: %v", err)
 	}
-	bpfProgArrayMap.Update(uint32(tailVfsWrite), uint32(prog.GetFd()))
```
itaysk, line 641: can you please explain this suggestion? `tailVfsWrite` is a uint32 anyway

itaysk (changes requested):

tracee/tracee.go
simar7: Hmm I didn't understand your comment, are you saying this is more of a log statement than an error itself? So in terms of code is your suggestion:

```
log.Printf("Failed to find kernel version: %v", err)
return err
```

tracee/tracee.go
simar7: Yes exactly - it's an uint32 already so this cast is unnecessary.

tracee/tracee_test.go
simar7: Yes https://github.com/aquasecurity/tracee/issues/451

tracee/tracee.go
itaysk: oh right, I read it the other way

itaysk: read this the wrong way :) thanks

tracee/tracee.go
itaysk: again, read this the wrong way :)

tracee/tracee_test.go
itaysk: thanks, so just remove the comment?

itaysk (dismissed):

yanivagman:
What about the failed test?

yanivagman (changes requested):
We shouldn't merge a commit which cause the tests to fail

tracee/tracee_test.go
```diff
@@ -120,16 +120,40 @@ func TestReadArgFromBuff(t *testing.T) {
 			},
 			expectedArg: map[string]string(map[string]string{"sa_family": "AF_INET", "sin_addr": "255.255.255.255", "sin_port": "65535"}),
 		},
+		{
+			name: "sockAddrT - AF_UNIX",
+			input: []byte{12, //sockAddrT
+				0,    // Dummy tag
+				1, 0, //sa_family=AF_UNIX
+				47, 116, 109, 112, 47, 115, 111, 99, 107, 101, 116, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 101, 110, 0, 0, 0, // sun_path=/tmp/socket
```
itaysk, line 128: the max length of `sun_path` is 108 bytes. here you give 114 bytes, which explains the extra 6 bytes that fails the test.

tracee/tracee_test.go
```diff
@@ -120,16 +120,40 @@ func TestReadArgFromBuff(t *testing.T) {
...
+		b := bytes.NewReader(tc.input)
+		_, actual, err := readArgFromBuff(b)
 		assert.Equal(t, tc.expectedError, err, tc.name)
 		assert.Equal(t, tc.expectedArg, actual, tc.name)
+
+		// TODO: the following fails for:
+		// strArrT
+		// unknown
+		// sockAddrT - AF_UNIX
+		assert.Empty(t, b.Len(), tc.name) // passed in buffer should be emptied out
```
itaysk, line 157: In the case of unknown, we can't consume the event from the buffer since we don't know the event's size. (we should stop tracee in this case). so I don't think we should assert for emptiness in this case

tracee/tracee_test.go
simar7: Ah nice catch - you're right.

tracee/tracee_test.go
simar7: Does that happen today? (The stoppage of Tracee that is) I just see all errors are collected and printed via a handler.

simar7: Anyways - that's tangential to this PR, we can take it offline.

yanivagman:
I think it is better to split this into 2 PRs - one for lint and go improvements, and the other for adding/changing tests

grantseltzer:
@simar7 It seems like this just needs a rebase. 

@yanivagman In general I agree with you but for this case I think it's small enough to let it slide.

tracee/tracee_test.go
simar7: Right - https://github.com/aquasecurity/tracee/pull/427/commits/3e460e1240b2cb6daae2a6779c4375f16c352874

simar7:
> I think it is better to split this into 2 PRs - one for lint and go improvements, and the other for adding/changing tests

I agree @yanivagman - I shouldn't have coupled those two. Will keep that in mind for the next one!

simar7:
> @simar7 It seems like this just needs a rebase.
> 
> @yanivagman In general I agree with you but for this case I think it's small enough to let it slide.

Just updated and rebased - PTAL

itaysk:
Thanks @simar7 , LGTM, but if we are fixing error capitalization, can you please fix the rest as well? a quick scan revealed more:
```
❯ grep -rn -e 'fmt.Errorf("[A-Z]' .
./libbpfgo/libbpfgo.go:312:             return fmt.Errorf("Failed to pin map %s to path %s", b.name, pinPath)
./libbpfgo/libbpfgo.go:324:             return fmt.Errorf("Failed to unpin map %s from path %s", b.name, pinPath)
./libbpfgo/libbpfgo.go:336:             return fmt.Errorf("Failed to set pin for map %s to path %s", b.name, pinPath)
./libbpfgo/libbpfgo.go:358:             return nil, fmt.Errorf("Unknown data type %T", data)
./libbpfgo/libbpfgo.go:648:                             return fmt.Errorf("Error polling perf buffer: %d", err)
./tracee-rules/signatures/rego/regosig/traceerego.go:150:       return fmt.Errorf("OnSignal is not implemented")
./tracee-ebpf/main.go:73:                               return fmt.Errorf("Insufficient privileges to run")
./tracee-ebpf/main.go:1178:             return fmt.Errorf("Failed to make BPF object (clang): %v. Try using --debug for more info", err)
./tracee-ebpf/main.go:1197:             return fmt.Errorf("Failed to make BPF object (llc): %v. Try using --debug for more info", err)
./tracee-ebpf/main.go:1214:                     return fmt.Errorf("Failed to make BPF object (llvm-strip): %v. Try using --debug for more info", err)
./tracee-ebpf/tracee/pipeline.go:165:                                   errc <- fmt.Errorf("Invalid arg tag for event %d", rawEvent.Ctx.EventID)
./tracee-ebpf/tracee/tracee.go:162:                     return fmt.Errorf("The length of a path filter is limited to 50 characters: %s", filter)
./tracee-ebpf/tracee/tracee.go:768:             return fmt.Errorf("Failed to find kernel version: %v", err)
```

(up to you if fix the ones in other projects in same commit or not)

itaysk:
I also looked for errors ending with period and found just one in libbpfgo:

```
❯ grep -rn -e 'fmt.Errorf(".*\."' .  
./libbpfgo/libbpfgo.go:253:             return nil, fmt.Errorf("failed to open BPF object %s: %v...", bpfObjName, bpfObjBuff[:20])
```

simar7:
> Thanks @simar7 , LGTM, but if we are fixing error capitalization can you please fix the rest as well? a quick scan revealed more:
> 
> ```
> ❯ grep -rn -e 'fmt.Errorf("[A-Z]' .
> ./libbpfgo/libbpfgo.go:312:             return fmt.Errorf("Failed to pin map %s to path %s", b.name, pinPath)
> ./libbpfgo/libbpfgo.go:324:             return fmt.Errorf("Failed to unpin map %s from path %s", b.name, pinPath)
> ./libbpfgo/libbpfgo.go:336:             return fmt.Errorf("Failed to set pin for map %s to path %s", b.name, pinPath)
> ./libbpfgo/libbpfgo.go:358:             return nil, fmt.Errorf("Unknown data type %T", data)
> ./libbpfgo/libbpfgo.go:648:                             return fmt.Errorf("Error polling perf buffer: %d", err)
> ./tracee-rules/signatures/rego/regosig/traceerego.go:150:       return fmt.Errorf("OnSignal is not implemented")
> ./tracee-ebpf/main.go:73:                               return fmt.Errorf("Insufficient privileges to run")
> ./tracee-ebpf/main.go:1178:             return fmt.Errorf("Failed to make BPF object (clang): %v. Try using --debug for more info", err)
> ./tracee-ebpf/main.go:1197:             return fmt.Errorf("Failed to make BPF object (llc): %v. Try using --debug for more info", err)
> ./tracee-ebpf/main.go:1214:                     return fmt.Errorf("Failed to make BPF object (llvm-strip): %v. Try using --debug for more info", err)
> ./tracee-ebpf/tracee/pipeline.go:165:                                   errc <- fmt.Errorf("Invalid arg tag for event %d", rawEvent.Ctx.EventID)
> ./tracee-ebpf/tracee/tracee.go:162:                     return fmt.Errorf("The length of a path filter is limited to 50 characters: %s", filter)
> ./tracee-ebpf/tracee/tracee.go:768:             return fmt.Errorf("Failed to find kernel version: %v", err)
> ```
> 
> (up to you if fix the ones in other projects in same commit or not)

OK - I have addressed them here https://github.com/aquasecurity/tracee/pull/427/commits/07bc04fc168561fa6ef3510205c3f497ad1d55d1

tracee-ebpf/tracee/tracee.go
```diff
@@ -1345,13 +1345,13 @@ func (t *Tracee) processFileWrites() {
...
 			if appendFile {
-				if _, err := f.Seek(0, os.SEEK_END); err != nil {
+				if _, err := f.Seek(0, io.SeekEnd); err != nil {
 					f.Close()
 					t.handleError(err)
 					continue
 				}
 			} else {
-				if _, err := f.Seek(int64(meta.Off), os.SEEK_SET); err != nil {
+				if _, err := f.Seek(int64(meta.Off), io.SeekStart); err != nil {
```
yanivagman, line 1354: This needs to be seek_set so tracee can rebuild the file according to the correct offsets. If you set it to seekstart instead, it will overwrite other data written, and the written file will be corrupted

yanivagman (changes requested):
Please see my comment about seek_set

tracee-ebpf/tracee/tracee.go
itaysk: @yanivagman `os.SEEK_SET` is equivalent to `io.SeekStart`. This change is merely using the same value from a different package (which is good)

itaysk (approved):

tracee-ebpf/tracee/tracee.go
yanivagman: Correct. My mistake.

yanivagman (approved):

PR #428, opened 2021-01-19
Add return value filter

### LINKED ISSUE #408: Filter by return value
It can be helpful to filter by the return value.
For example, we might be interested only in succssesful system calls or failed ones.

yanivagman:
close #408

### DIFF
--- a/main.go
+++ b/main.go
@@ -255,6 +255,9 @@ Event arguments can be accessed using 'event_name.event_arg' and provide a way t
 Event arguments allow the following operators: '=', '!='.
 Strings can be compared as a prefix if ending with '*'.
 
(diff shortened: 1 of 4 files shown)


itaysk (commented):
why is this an `event.` expression and not a regular numerical filter? If you think about it, every thing that we can filter is related to the event but some are expressed as event and some are not.

yanivagman:
It's an event expression because I think it's more useful than a numerical expression that will catch for all the events.
If you only want to filter failed execve's or failed open, for example, but keep all the other events as usual - having this option is much better. If you think about it, the context of the return value depends on the event itself. For some syscalls, '0' return value might signify success while for other, failure

itaysk (approved):

PR #429, opened 2021-01-20
Make '--capture clear-dir' safer

### LINKED ISSUE #426: Make clear-dir safer
When `--capture clear-dir` is given, we delete the path given by `--capture dir:/path`.
This is dangerous in the case that the user choose to use both options, and gives, for example, "/" or "~" as the capture dir.
In that case, Tracee will DELETE all of these folders content!

To avoid this situation, one solution is to create a subdirectory under the chosen capture directory (for example, create "out" directory)

yanivagman:
close #426

### DIFF
--- a/main.go
+++ b/main.go
@@ -174,7 +174,7 @@ Possible options:
 [artifact:]mem                     capture memory regions that had write+execute (w+x) protection, and then changed to execute (x) only.
 [artifact:]all                     capture all of the above artifacts.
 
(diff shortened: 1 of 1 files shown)


itaysk (approved):

itaysk:
what about tracee.pid file?

yanivagman:
I didn't change tracee.pid location - it will be in the output directory as before.
But like I said, this should be changed.
We will need to handle it in a different commit.

PR #431, opened 2021-01-20
refactor output flag

itaysk:
related: https://github.com/aquasecurity/tracee/issues/397

this also fixes a small issue with --capture documentation 

There are a few practices here that if approved I think we should implement in the other preparers in the future.

### DIFF
--- a/main.go
+++ b/main.go
@@ -41,16 +41,15 @@ func main() {
 			}
 
 			cfg := tracee.TraceeConfig{
-				DetectOriginalSyscall: c.Bool("detect-original-syscall"),
-				ShowExecEnv:           c.Bool("show-exec-env"),
-				OutputFormat:          c.String("output"),
-				PerfBufferSize:        c.Int("perf-buffer-size"),
-				BlobPerfBufferSize:    c.Int("blob-perf-buffer-size"),
-				SecurityAlerts:        c.Bool("security-alerts"),
-				EventsFile:            os.Stdout,
-				ErrorsFile:            os.Stderr,
-				StackAddresses:        c.Bool("stack-addresses"),
+				PerfBufferSize:     c.Int("perf-buffer-size"),
+				BlobPerfBufferSize: c.Int("blob-perf-buffer-size"),
+				SecurityAlerts:     c.Bool("security-alerts"),
 			}
+			output, err := prepareOutput(c.StringSlice("output"))
+			if err != nil {
+				return err
+			}
+			cfg.Output = &output
 			capture, err := prepareCapture(c.StringSlice("capture"))
 			if err != nil {
 				return err
@@ -99,15 +98,16 @@ func main() {
 				Value:   nil,
 				Usage:   "capture artifacts that were written, executed or found to be suspicious. run '--capture help' for more info.",
 			},
-			&cli.BoolFlag{
-				Name:  "detect-original-syscall",
-				Value: false,
-				Usage: "when tracing kernel functions which are not syscalls (such as cap_capable), detect and show the original syscall that called that function",
+			&cli.StringSliceFlag{
+				Name:    "output",
+				Aliases: []string{"o"},
+				Value:   cli.NewStringSlice("format:table"),
+				Usage:   "Control how and where output is printed. run '--output help' for more info.",
 			},
 			&cli.BoolFlag{
-				Name:  "show-exec-env",
+				Name:  "security-alerts",
 				Value: false,
-				Usage: "when tracing execve/execveat, show environment variables",
+				Usage: "alert on security related events",
 			},
 			&cli.IntFlag{
 				Name:    "perf-buffer-size",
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -163,20 +147,82 @@ func main() {
...
+
+	res := tracee.OutputConfig{}
+	if len(outputSlice) == 1 && outputSlice[0] == "help" {
+		return res, fmt.Errorf(outputHelp)
+	}
+
+	for _, o := range outputSlice {
+		outputParts := strings.SplitN(o, ":", 2)
+		if outputParts[0] == "out-file" {
+			res.OutPath = outputParts[1]
```
yanivagman, line 179: What if the input "out-file" was given? In that case, there is no index 1 to outputParts

main.go
```diff
@@ -163,20 +147,82 @@ func main() {
...
+		} else if outputParts[0] == "err-file" {
+			res.ErrPath = outputParts[1]
```
yanivagman, line 181: same

main.go
```diff
@@ -163,20 +147,82 @@ func main() {
...
+		} else if outputParts[0] == "option" {
+			switch outputParts[1] {
```
yanivagman, line 183: same

main.go
```diff
@@ -163,20 +147,82 @@ func main() {
...
+			default:
+				return res, fmt.Errorf("invalid output option: %s, use '--option help' for more info", outputParts[1])
+			}
+		} else {
+			numParts := len(outputParts)
+			if numParts == 1 {
+				res.Format = strings.TrimPrefix(outputParts[0], "format:")
+			} else if numParts == 2 {
+				res.Format = outputParts[1]
+			}
```
yanivagman, line 201: Is it ok to have an input of more than 2 splits? e.g.: "option:eot:3:2"

main.go
```diff
@@ -163,20 +147,82 @@ func main() {
...
+				res.DetectSyscall = true
+			case "exec-env":
+				res.ExecEnv = true
+			default:
+				return res, fmt.Errorf("invalid output option: %s, use '--option help' for more info", outputParts[1])
+			}
+		} else {
+			numParts := len(outputParts)
+			if numParts == 1 {
+				res.Format = strings.TrimPrefix(outputParts[0], "format:")
```
yanivagman, line 198: if numPart is 1, the string can't start with "format:" (as it will have a size of 2 or more) - am I wrong?

main.go
```diff
@@ -163,20 +147,82 @@ func main() {
...
 [artifact:]all                     capture all of the above artifacts.
 
-dir:/path/to/dir        path where tracee will save produced artifacts (default: /tmp/tracee).
+dir:/path/to/dir        path where tracee will save produced artifacts. the artifact will be saved into an 'out' subdirectory. (default: /tmp/tracee).
 clear-dir               clear the captured artifacts output dir before starting (default: false).
 
+Examples:
+	--capture exec                                           | capture executed files into the default output directory
+	--capture all --capture dir:/my/dir --capture clear-dir  | delete /my/dir/out and then capture all supported artifacts into it
+	--capture write=/usr/bin --capture write=/etc/*          | capture files that were written into /usr/bin directly, or anywhere below /etc
```
yanivagman, line 224: write=/usr/bin without "*" will not work as we don't support this syntax

main.go
itaysk: I'm doing `SplitN` with 2 splits, so we can't get more then 2 items.
the example you gave will be split into: `[option eot:3:2]`

main.go
```diff
@@ -163,20 +147,82 @@ func main() {
...
+func prepareOutput(outputSlice []string) (tracee.OutputConfig, error) {
+	outputHelp := `
+Control how and where output is printed.
+Possilbe options:
+
+[format:]{table,table-verbose,json,gob,gotemplate=/path/to/template}   output events in the specified format. for gotemplate, specify the mandatody template file
+out-file:/path/to/file                                                 write the output to a specified file. the path to the file will be created if not existing and the file will be deleted if existing (deafult: stdout)
+err-file:/path/to/file                                                 write the errors to a specified file. the path to the file will be created if not existing and the file will be deleted if existing (deafult: stderr)
+option:{eot,stacktrace,detect-syscall,exec-env}                        augment output according to given options (default: none)
+	eot                                                                add a final event that signals the end of event stream. this is an empty event with "EventName" set to the ASCII code 4, which is "End Of Transmission"
```
yanivagman, line 159: Alignment?

main.go
```diff
@@ -163,20 +147,82 @@ func main() {
...
+	stack-addresses                                                    include stack memory addresses for each event
+	detect-syscall                                                     when tracing kernel functions which are not syscalls, detect and show the original syscall that called that function
+	exec-env                                                           when tracing execve/execveat, show the environment variables that were used for execution
+Examples:
+	--output json --output option:eot                        | output as json and add an EOT event
+	--output gotemplate=/path/to/my.tmpl                     | output as the provided go template
+	--output out-file:/my/out err-file:/my/err                 | output to /my/out and errors to /my/err
```
yanivagman, line 166: Alignment?

yanivagman (approved):
Only a small comment about alignment

PR #432, opened 2021-01-21
Publish the image

krol3:
Publish the image tracee


itaysk:
Sorry but we are publishing images via our [current release script](https://github.com/aquasecurity/tracee/blob/7d2ce345dfca4f1f24f2aa1a0a99a97e5ad0952f/Makefile#L181).
There's a PR to run this script from GH Actions: #367

PR #433, opened 2021-01-21
update readme

itaysk:
note that I chose not to detail ALL of the available options as it will be a copy of the help. I highlighted the general patterns and important flags. In the future we should find a way to automatically share the help between the code and github

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,7 @@
 
 # Tracee - Container and system tracing using eBPF
 
-Tracee is a lightweight and easy to use container and system tracing tool. It allows you to observe system calls and other system events in real-time. A unique feature of Tracee is that it will only trace newly created processes and containers (that were started after Tracee has started), in order to help the user focus on relevant events instead of every single thing that happens on the system (which can be overwhelming). Adding new events to Tracee (especially system calls) is straightforward, and will usually require no more than adding few lines of code.
-
-Other than tracing, Tracee is also capable of capturing files written to disk or memory ("fileless"), and extracting binaries that are dynamically loaded to an application's memory (e.g. when an application uses a packer). With these features, it is possible to quickly gain insights about the running processes that previously required the use of dynamic analysis tools and special knowledge.
+Tracee is a lightweight and easy to use tracing tool for Linux, which is focused on security and forensics. It allows you to observe system calls and other system events in real-time, with comprehensive filtering mechanism so you can focus on the events that are relevant to you. Unlike other tracing tools, Tracee is a security tool, which is demonstrated by features like capturing forensic artifacts from running applications, tracing non-syscall security events, and producing security insights.
 
 [Check out this quick demo of tracee](https://youtu.be/WTqE2ae257o)
 
@@ -17,38 +15,47 @@ Other than tracing, Tracee is also capable of capturing files written to disk or
 
 ### Prerequisites
 
-- To run, Tracee requires Linux kernel version >= 4.14
-
-Not required if using the Docker image:
-- C standard library (tested with glibc)
-- `libelf` and `zlib` libraries
+- Linux kernel version >= 4.14
+- Relevant kernel headers available under conventional location (see [Linux Headers](#Linux-Headers) section for info)
+- libc, and the libraries: libelf and zlib
 - clang >= 9
 
-Not required if pre-compiling the eBPF code (see [Installation options](#installation-options)):
-- clang >= 9
-- Kernel headers available under `/usr/src`, must be provided by user and match the running kernel version, not needed if building the eBPF program in advance
+Exceptions:
+
+- Tracee supports loading a pre-compiled eBPF file, in which case the kernel headers are required only for the one-time compilation, and not at runtime. See Setup Options for more info.
+- When using Tracee's Docker image, all of the aforementioned requirements are built into the image. The only requirement left is the kernel headers or the pre-built eBPF. See Setup Options for more info.
 
 ### Quickstart with Docker
 
 ```bash
-docker run --name tracee --rm --privileged --pid=host -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro -v /tmp/tracee:/tmp/tracee aquasec/tracee
+docker run --name tracee --rm --privileged --pid=host -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro -v /tmp/tracee:/tmp/tracee aquasec/tracee:latest
 ```
 
-> Note: You may need to change the volume mounts for the kernel headers based on your setup.
+> Note: You may need to change the volume mounts for the kernel headers based on your setup. See [Linux Headers](#Linux-Headers) section for info.
 
-This will run Tracee with no arguments, which defaults to collecting all events from all newly created processes and printing them in a table to standard output.
+This will run Tracee with no arguments, which defaults to collecting a useful default set of events from all processes and print them in a table to standard output.
 
-### Setup options
+### Getting Tracee
 
-Tracee is made of an executable that drives the eBPF program (`tracee`), and the eBPF program itself (`tracee.bpf.$kernelversion.$traceeversion.o`). When the `tracee` executable is started, it will look for the eBPF program next to the executable, or in `/tmp/tracee`, or in a path specified in `TRACEE_BPF_FILE` environment variable. If the eBPF program is not found, the executable will attempt to build it automatically before it starts (you can control this using the `--build-policy` flag).
-
-The easiest way to get started is to let the `tracee` executable build the eBPF program for you automatically. You can obtain the executable in any of the following ways:
+You can obtain Tracee in any of the following ways:
 1. Download from the [GitHub Releases](https://github.com/aquasecurity/tracee/releases) (`tracee.tar.gz`).
 2. Use the docker image from Docker Hub: `aquasec/tracee` (includes all the required dependencies).
 3. Build the executable from source using `make build`. For that you will need additional development tooling.
 4. Build the executable from source in a Docker container which includes all development tooling, using `make build DOCKER=1`.
 
-Alternatively, you can pre-compile the eBPF program, and provide it to the `tracee` executable. There are some benefits to this approach since you will not need clang and kernel headers at runtime anymore, as well as reduced risk of invoking an external program at runtime. You can build the eBPF program in the following ways:
+### Setup options
+
+Tracee is made of a userspace executable (`tracee`) that drives the eBPF program, and the eBPF program itself (`tracee.bpf.$kernelversion.$traceeversion.o`). When the `tracee` is started, it will look for the eBPF program in specific places and if not found, it will attempt to build the eBPF program automatically before it starts (you can control this using the `--build-policy` flag).
+
+The eBPF program is searched in the following places (in order):
+1. Path specified in `TRACEE_BPF_FILE` environment variable
+2. Next to the executable (same directory)
+3. `/tmp/tracee`
+
+The easiest way to get started is to just let tracee build the eBPF program for you automatically, as demonstrated in the previous section [Quickstart with Docker](#quickstart-with-docker).  
+Alternatively, you can pre-compile the eBPF program, and provide it to the `tracee` executable. There are some benefits to this approach as you will not need clang and kernel headers at runtime anymore, as well as reduced risk of invoking an external program at runtime.
+
+You can build the eBPF program in the following ways:
 1. `make bpf`
 2. `make bpf DOCKER=1` to build in a Docker container which includes all development tooling.
 3. There is also a handy `make all` (and the `make all DOCKER=1` variant) which builds both the executable and the eBPF program.
@@ -63,7 +70,7 @@ If running in a container, regardless if it's the full or slim image, it's advis
 
 When using the `--capture exec` option, Tracee needs access to the host PID namespace. For Docker, add `--pid=host` to the run command.
 
-If you are building the eBPF program in a container, you'll need to make the kernel headers available in the container. The quickstart example has wide mounts that works in a variety of cases, for demonstration purposes. If you want, you can narrow those mounts down to a directory that contains the headers on your setup, for example: `-v /path/to/headers:/myheaders -e KERN_HEADERS=/myheaders`. As mentioned before, a better practice for production is to pre-compile the eBPF program, in which case the kernel headers are not needed at runtime.
+If you are building the eBPF program in a container, you'll need to make the kernel headers available in the container. The quickstart example has broader mounts that works in a variety of cases, for demonstration purposes. If you want, you can narrow those mounts down to the specific directory that contains the headers on your setup, for example: `-v /path/to/headers:/myheaders -e KERN_HEADERS=/myheaders`. As mentioned before, a better practice for production is to pre-compile the eBPF program, in which case the kernel headers are not needed at runtime.
 
 #### Permissions
 
@@ -74,11 +81,27 @@ For actually tracing, Tracee needs to run with sufficient capabilities:
 
 Alternatively, running as `root` or with the `--privileged` flag of Docker, is an easy way to start.
 
+#### Linux Headers
+
+In order to compile the eBPF program, Tracee needs some of the Linux kernel headers. Depending on your Linux distribution, there may be different ways to obtain them.  
+
+- On Ubuntu/Debian/Arch/Manjaro install the `linux-headers` package.
+- On CentOS/Fedora install the `kernel-headers` and `kernel-devel` packages.
+
+Normally the files will be installed in `/lib/modules/${kernel_version}/build` which is where Tracee expects them. If you have the headers elsewhere, you can set the `KERN_HEADERS` environment variable with the correct location.
+
+> Note that it's important that the kernel headers match the exact version of kernel you are running. To check the current kernel version, run the command `uname -r`. To install a specific kernel headers version append the version to the package name: `linux-headers-$(uname -r)`.
+
+> Note that more often than not the kernel headers files contains filesystem links to other files in other directories. Therefore, when passing the kernel headers to Tracee docker container, make sure all the necessary directories are mounted. This is why the quickstart example mounts `/usr/src` in addition to `/lib/modules`.
+
 ## Using Tracee
+
+Use `tracee --help` to see a full description of available options. Some flags has specific help sections that can be accessed by passing `help` to the flag, for example `--output help`.
+This section covers some of the more common options.
 
 ### Understanding the output
 
-Here's a sample output of running Tracee with no additional arguments (which defaults to tracing all events):
+Here's a sample output of running Tracee with no additional arguments:
 
 ```
 TIME(s)        UID    COMM             PID     TID     RET             EVENT                ARGS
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -74,11 +82,30 @@ For actually tracing, Tracee needs to run with sufficient capabilities:
 
 Alternatively, running as `root` or with the `--privileged` flag of Docker, is an easy way to start.
 
+#### Linux Headers
+
+In order to compile the eBPF program, Tracee needs some of the Linux kernel headers. Depending on your Linux distribution, there are different ways to obtain it, and different places they end up at.  
+
+> Note that it's important that the kernel headers match the exact version of kernel you are running. To check the current kernel version, run the command `uname -r`
+
+Here are a few examples of how to obtain Linux kernel headers in different Linux distributions:
```
yanivagman, line 91: You write a few, but give only one example

README.md
```diff
@@ -74,11 +82,30 @@ For actually tracing, Tracee needs to run with sufficient capabilities:
...
+
+Distribution | Command | Installation location
+--- | --- | ---
+Ubuntu | apt-get install linux-headers | /usr/bin
+
+When building the eBPF object file using Tracee's `make bpf`, it will assume the headers to be found in common location: `/lib/modules/${kernel_version}/build`. If you have the headers elsewhere, you need to provide them to `make` using `KERN_BLD_PATH` and `KERN_SRC_PATH` for the build and sources appropriately.
```
yanivagman, line 97: This can be a bit confusing. Not all users are aware of what is build and sources directories (even for us it took some time to figure this out). Also, the more common thing is to have all the headers in one directory, in which case only 'KERN_HEADERS" can be used

README.md
```diff
@@ -100,41 +127,47 @@ Each line is a single event collected by Tracee, with the following information:
...
+### Selecting what to trace
+
+Trace output can easily become unwieldily when tracing all of the events from all of the system. Luckily, Tracee has a powerful mechanism to accurately trace just the information that is relevant to you, using the `--trace` flag.
+Using the `--trace` you define expressions that tells Tracee what you are interested in by means of event metadata, and process metadata. Only events that match this criteria will be traced.
+
+You can filter by most of the visible fields on a Tracee event. For example to trace only events related to user ID 1000, use `--trace uid=1000`.  
+You can combine trace expressions into more complex criteria. For example, to trace only events related to user ID 1000, which come from process ID 1234, use `--trace uid=1000 --trace pid=1234`.  
+
+A special `pid` value is `new` which let's you trace all newly created processes (that were created after Tracee started tracing).  
+Tracee let's use easily trace events that originate in containers using `--trace container` or only new containers (that was created after Tracee started) using `--trace container=new`.
```
yanivagman, line 149: Tracee let's use? Not sure I understand this sentence

README.md
```diff
@@ -100,41 +127,47 @@ Each line is a single event collected by Tracee, with the following information:
...
+
+Event metadata can be used in trace expression as well. For example, to trace only `operat` syscalls, use `--trace event:openat`. But you can also filter on a specific argument of the event, e.g `--trace openat.pathname=/bin/ls` which will show only `openat` syscalls that operate on the file `/bin/ls`.
```
yanivagman, line 151: operat -> openat

README.md
```diff
@@ -100,41 +127,47 @@ Each line is a single event collected by Tracee, with the following information:
...
 
-1. UTS_NAME - uts namespace name. As there is no container id object in the kernel, and docker/k8s will usually set this to the container id, we use this field to distinguish between containers.
-2. MNT_NS - mount namespace inode number.
-3. PID_NS - pid namespace inode number. In order to know if there are different containers in the same pid namespace (e.g. in a k8s pod), it is possible to check this value
-4. PPID - parent pid of the calling process
+For a complete list of trace options, run `--trace help`.
```
yanivagman, line 153: I think that the "follow" flag is very useful and should be part of the help, don't you think?

README.md
```diff
@@ -100,41 +127,47 @@ Each line is a single event collected by Tracee, with the following information:
...
-Trace new processes (default) | no `--trace` flag, `--trace p`, `--trace process` or `--trace process:new`
-Trace existing and new processes | `--trace process:all`
-Trace specific PIDs | `--trace process:<pid>,<pid2>,...` or `--trace p:<pid>,<pid2>,...`
-Trace filtered process and all of its children | `--trace process:follow`
-Trace new containers | `--trace c`, `--trace container` or `--trace container:new`
-Trace existing and new containers | `--trace container:all`
-Trace new processes not in a container | `--trace h`, `--trace host` or `--trace host:new`
-Trace all processes not in a container | `--trace host:all`
+To use, `--capture exec`, `--capture mem`, and `--capture write` capture executed, memory, and written files respectively. 
+Written files can be filtered using an optional path prefix, e.g `--capture write=/etc/*` will capture anything written blow `/etc/`.
```
yanivagman, line 168: If you are writing about the write filter here, I don't think it is necessary in point 1 above

README.md
```diff
@@ -7,48 +7,56 @@
...
-
-Other than tracing, Tracee is also capable of capturing files written to disk or memory ("fileless"), and extracting binaries that are dynamically loaded to an application's memory (e.g. when an application uses a packer). With these features, it is possible to quickly gain insights about the running processes that previously required the use of dynamic analysis tools and special knowledge.
+Tracee is a lightweight and easy to use container and system tracing tool. It allows you to observe system calls and other system events in real-time. A unique feature of Tracee is it's comprehensive filtering abilities which allows you to focus on relevant events instead of every single thing that happens on the system. Unlike other tracing tools, Tracee is a security tool, which is demonstrated by features like capturing IoC artifacts from running applications, tracing non-syscall security events, and producing security insights.
```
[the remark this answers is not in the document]
itaysk: I wanted to remove this whole uniqueness part, but I remember you wanted to add it in the first place. are you ok with removing it? so it will be just what you wrote, and the rest they can read about later

README.md
itaysk: yes, I meant to ask you about manjaro and simar about centos

README.md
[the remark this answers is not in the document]
yanivagman: The situation is now different, because we no longer filter new processes by default, so modifications are needed, and I understand that.
If you are writing about the unique filters as I wrote - that should be enough

README.md
```diff
@@ -7,48 +7,55 @@
 
 # Tracee - Container and system tracing using eBPF
 
-Tracee is a lightweight and easy to use container and system tracing tool. It allows you to observe system calls and other system events in real-time. A unique feature of Tracee is that it will only trace newly created processes and containers (that were started after Tracee has started), in order to help the user focus on relevant events instead of every single thing that happens on the system (which can be overwhelming). Adding new events to Tracee (especially system calls) is straightforward, and will usually require no more than adding few lines of code.
-
-Other than tracing, Tracee is also capable of capturing files written to disk or memory ("fileless"), and extracting binaries that are dynamically loaded to an application's memory (e.g. when an application uses a packer). With these features, it is possible to quickly gain insights about the running processes that previously required the use of dynamic analysis tools and special knowledge.
+Tracee is a lightweight and easy to use tracing tool for Linux, which is focused on security and forensics.
```
yanivagman, line 10: I think we need to keep some of the intro, don't you think?

itaysk: I think we either say something general about unique features or don't say anything and let the demo speak for itself. what do you prefer?

yanivagman: I think we need here at least one paragraph which describes tracee

README.md
```diff
@@ -100,41 +126,50 @@ Each line is a single event collected by Tracee, with the following information:
...
+You can combine trace expressions into more complex criteria. For example, to trace only events related to user ID 1000, which come from process ID 1234, use `--trace uid=1000 --trace pid=1234`.  
+
+A special `pid` value is `new` which let's you trace all newly created processes (that were created after Tracee started tracing).  
+Tracee lets you easily trace events that originate in containers using `--trace container` or only new containers (that was created after Tracee started) using `--trace container=new`.
+
+Event metadata can be used in trace expression as well. For example, to trace only `openat` syscalls, use `--trace event:openat`. But you can also filter on a specific argument of the event, e.g `--trace openat.pathname=/bin/ls` which will show only `openat` syscalls that operate on the file `/bin/ls`.
+
+A useful trace mode is the `--trace follow` which, if specified, will trace not only processes that match the given trace expressions, but also their child processes.
+For example, the following will trace all the events that originate from zsh shell, including all of the processes that will spawns: `--trace command=zsh --follow`.
```
yanivagman, line 153: that will spawns -> that it will spawn, or, that is spawns

README.md
```diff
@@ -74,11 +81,30 @@ For actually tracing, Tracee needs to run with sufficient capabilities:
...
+
+> Note that it's important that the kernel headers match the exact version of kernel you are running. To check the current kernel version, run the command `uname -r`
+
+Here are a few examples of how to obtain Linux kernel headers in different Linux distributions:
+
+Distribution | Command | Installation location
+--- | --- | ---
+Ubuntu | apt-get install linux-headers | /usr/bin
+
+When building the eBPF object file using Tracee's `make bpf`, it will assume the headers to be found in a common location: `/lib/modules/${kernel_version}/build`. If you have the headers elsewhere, you need to provide them to `make` using `KERN_HEADERS` make variable.
```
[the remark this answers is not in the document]
itaysk: it's actually a make variable, not an environment variable

README.md
```diff
@@ -7,48 +7,55 @@
...
 1. Download from the [GitHub Releases](https://github.com/aquasecurity/tracee/releases) (`tracee.tar.gz`).
 2. Use the docker image from Docker Hub: `aquasec/tracee` (includes all the required dependencies).
 3. Build the executable from source using `make build`. For that you will need additional development tooling.
 4. Build the executable from source in a Docker container which includes all development tooling, using `make build DOCKER=1`.
 
-Alternatively, you can pre-compile the eBPF program, and provide it to the `tracee` executable. There are some benefits to this approach since you will not need clang and kernel headers at runtime anymore, as well as reduced risk of invoking an external program at runtime. You can build the eBPF program in the following ways:
+### Setup options
+
+Tracee is made of an executable (`tracee`) that drives the eBPF program, and the eBPF program itself (`tracee.bpf.$kernelversion.$traceeversion.o`). When the `tracee` is started, it will look for the eBPF program in specific places and if not found, it will attempt to build the eBPF program automatically before it starts (you can control this using the `--build-policy` flag).
```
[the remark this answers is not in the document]
itaysk: thanks @grantseltzer I've applied some of you're various suggestions into the current commit, instead of accepting the suggestion, because we generally like to keep the commit history simple.

README.md
```diff
@@ -7,48 +7,55 @@
...
-Tracee is a lightweight and easy to use container and system tracing tool. It allows you to observe system calls and other system events in real-time. A unique feature of Tracee is that it will only trace newly created processes and containers (that were started after Tracee has started), in order to help the user focus on relevant events instead of every single thing that happens on the system (which can be overwhelming). Adding new events to Tracee (especially system calls) is straightforward, and will usually require no more than adding few lines of code.
-
-Other than tracing, Tracee is also capable of capturing files written to disk or memory ("fileless"), and extracting binaries that are dynamically loaded to an application's memory (e.g. when an application uses a packer). With these features, it is possible to quickly gain insights about the running processes that previously required the use of dynamic analysis tools and special knowledge.
+Tracee is a lightweight and easy to use tracing tool for Linux, which is focused on security and forensics.
 
 [Check out this quick demo of tracee](https://youtu.be/WTqE2ae257o)
 
 ## Getting started
 
 ### Prerequisites
```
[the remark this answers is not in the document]
itaysk: the thing is that building the ebpf probe IS part of running tracee. So I think it would be misleading if we rename it to "building", as people would think it just prereqs to build the executable, but these are actually required on their target environments

README.md
```diff
@@ -74,11 +81,30 @@ For actually tracing, Tracee needs to run with sufficient capabilities:
...
+In order to compile the eBPF program, Tracee needs some of the Linux kernel headers. Depending on your Linux distribution, there are different ways to obtain it, and different places they end up at.  
+
+> Note that it's important that the kernel headers match the exact version of kernel you are running. To check the current kernel version, run the command `uname -r`
+
+Here are a few examples of how to obtain Linux kernel headers in different Linux distributions:
+
+Distribution | Command | Installation location
+--- | --- | ---
+Ubuntu | apt-get install linux-headers | /usr/bin
+
```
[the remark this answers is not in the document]
itaysk: thanks @grantseltzer is the kernel-devel actually required? if you have a box to verify this (as well as the install location), that would be great since we want to be sure before adding to the readme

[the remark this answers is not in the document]
itaysk: pinging @simar7 if you have anything to add here

README.md
itaysk: so how do you do it on manjaro? @yanivagman

README.md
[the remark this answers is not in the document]
yanivagman: For manjaro it's just /lib/modules/$(uname -r)/build/include/linux without any links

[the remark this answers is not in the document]
itaysk: how do you install headers on manjaro?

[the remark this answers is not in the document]
yanivagman: sudo pacman -S linux-headers

yanivagman (approved):

README.md
```diff
@@ -100,41 +121,50 @@ Each line is a single event collected by Tracee, with the following information:
...
+Event metadata can be used in trace expression as well. For example, to trace only `openat` syscalls, use `--trace event:openat`. But you can also filter on a specific argument of the event, e.g `--trace openat.pathname=/bin/ls` which will show only `openat` syscalls that operate on the file `/bin/ls`.
+
+A useful trace mode is the `--trace follow` which, if specified, will trace not only processes that match the given trace expressions, but also their child processes.
+For example, the following will trace all the events that originate from zsh shell, including all of the processes that it will spawn: `--trace command=zsh --follow`.
 
-1. UTS_NAME - uts namespace name. As there is no container id object in the kernel, and docker/k8s will usually set this to the container id, we use this field to distinguish between containers.
-2. MNT_NS - mount namespace inode number.
-3. PID_NS - pid namespace inode number. In order to know if there are different containers in the same pid namespace (e.g. in a k8s pod), it is possible to check this value
-4. PPID - parent pid of the calling process
+For a complete list of trace options, run `--trace help`.
```
simar7, line 150: When I tried to see the help today, I instinctively passed it as the following:
```
./tracee --trace --help
2021/01/22 11:34:22 invalid filter option specified, use '--filter help' for more info
```

I realize we currently only look at the arg to have "help" and not "--help": https://github.com/aquasecurity/tracee/blob/main/main.go#L345-L347

Should we add "--help" as well?

simar7: Should we return an error (invalid filter) in this case: `./tracee --trace 'uid<0'` same for `./tracee --trace 'pid<0'`? currently we start tracing

simar7: What is expected to happen with `--trace '!c'`? I currently don't see we handle it here: https://github.com/aquasecurity/tracee/blob/main/main.go#L440 but if `--trace '!container'` and `--trace c` are a thing, intuitively I think `--trace '!c'` would be too.

yanivagman: Doing --trace '!c' works exactly as you expect. Give it a try (and then you can also see why the hasprefix in line 440 works - it's 'f' being the prefix of "!container")

simar7: Ah yes - I missed that detail. It does get caught by '!c'. I think we should add it to the readme as an example.

README.md
```diff
@@ -100,41 +121,50 @@ Each line is a single event collected by Tracee, with the following information:
...
+
+To tell tracee to write events to a file instead of stdout, use `--output out-file:/path/to/file`.
+
+There are different ways you can augment the output to add useful information. For example: `--output eot` will add a terminating event to the stream which is useful if feeding Tracee's output to another program.
+
+For a full list of output options, run `--output help`.
+
+### Selecting what to trace
+
+Trace output can easily become unwieldily when tracing all of the events from all of the system. Luckily, Tracee has a powerful mechanism to accurately trace just the information that is relevant to you, using the `--trace` flag.
```
simar7, line 136:
```suggestion
Trace output can easily become unwieldy when tracing all of the events from all of the system. Luckily, Tracee has a powerful mechanism to accurately trace just the information that is relevant to you, using the `--trace` flag.
```

README.md
```diff
@@ -100,41 +121,50 @@ Each line is a single event collected by Tracee, with the following information:
...
+Using the `--trace` you define expressions that tells Tracee what you are interested in by means of event metadata, and process metadata. Only events that match this criteria will be traced.
+
+You can filter by most of the visible fields on a Tracee event. For example to trace only events related to user ID 1000, use `--trace uid=1000`.  
+You can combine trace expressions into more complex criteria. For example, to trace only events related to user ID 1000, which come from process ID 1234, use `--trace uid=1000 --trace pid=1234`.  
+
+A special `pid` value is `new` which let's you trace all newly created processes (that were created after Tracee started tracing).  
+Tracee lets you easily trace events that originate in containers using `--trace container` or only new containers (that was created after Tracee started) using `--trace container=new`.
```
simar7, line 143:
```suggestion
Tracee lets you easily trace events that originate in containers using `--trace container` or only new containers (that were created after Tracee started) using `--trace container=new`.
```

README.md
```diff
@@ -100,41 +121,50 @@ Each line is a single event collected by Tracee, with the following information:
...
-Here are a few commonly useful flags:
-- `--trace` Sets the trace mode. For more information see [Trace Mode Configuration](#Trace-Mode-Configuration) below
-- `--event` allows you to specify a specific event to trace. You can use this flag multiple times, for example `--event execve --event openat`.
-- `--list` lists the events available for tracing, which you can provide to the `--event` flag.
-- `--output` lets you control the output format, for example `--output json` will output as JSON lines instead of table.
-- `--capture` capture artifacts that were written, executed or found suspicious, and save them to the output directory. Possible values are: 'write'/'exec'/'mem'/'all'
+All captured artifacts are saved in Tracee's "output directory" which can be configured using `--capture dir:/path/to/dir`.
 
-### Trace Mode Configuration
+Tracee can capture the following types of IoCs:
```
simar7, line 158: What is IoC?

README.md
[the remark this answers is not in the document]
simar7: > pinging @simar7 if you have anything to add here

Should we add the section on selinux somewhere? Doesn't have to be here but I feel lots of people will trip considering how it's setup (enforcing) out of the box on certain distros (centos)

README.md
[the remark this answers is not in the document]
grantseltzer: Ah I see, if the object doesn't exist at $TRACEE_BPF_FILE. Yea this is a little tricky to document and implement. We should explore embedding the bpf object file. I think this all gets easier once we have CO:RE support.

This is fine for now 👍

README.md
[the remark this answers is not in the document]
grantseltzer: Ah ok, got it confused with `KERN_SRC`

README.md
grantseltzer: Indicator of compromise.

simar7: Roger - we should put that here as it's the first usage within this document.
```suggestion
Tracee can capture the following types of Indicator of Compromise (IoC):
```

README.md
yanivagman: Regarding pid/uid<0 - I agree. We should produce an error when such values are given.
Also agree about the --help thing (and -h as well).
These two, of course, are for different commits.

Regarding the '!c' in the readme - it previously was an example in the --trace help, but I removed it as I already put there "container", "c" and "!container", and I thought it was only logical that people will understand what "!c" does.

simar7: Roger - I created an issue to track it: https://github.com/aquasecurity/tracee/issues/450

PR #437, opened 2021-01-22
Add pin, unpin and setpinpath for maps in libbpfgo

ogozman:
Addid pinning for maps in libbpfgo

### DIFF
--- a/libbpfgo/libbpfgo.go
+++ b/libbpfgo/libbpfgo.go
@@ -321,7 +321,7 @@ func (b *BPFMap) Unpin(pinPath string) error {
 	errC := C.bpf_map__unpin(bpfMap, path)
 	C.free(unsafe.Pointer(cs))
 	if errC != 0 {
-		return fmt.Errorf("Failed unpin map %s from path %s", b.name, pinPath)
+		return fmt.Errorf("Failed to unpin map %s from path %s", b.name, pinPath)
 	}
 	return nil
 }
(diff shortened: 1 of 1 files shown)


itaysk (changes requested):
please free the allocated strings. you can take example from other functions that use CString

libbpfgo/libbpfgo.go
```diff
@@ -302,6 +302,39 @@ func (m *Module) GetMap(mapName string) (*BPFMap, error) {
...
+	path := C.CString(pinning)
+	bpfMap := C.bpf_object__find_map_by_name(b.module.obj, cs)
+	errC := C.bpf_map__unpin(bpfMap, path)
+	if errC != 0 {
+		return fmt.Errorf("Failed unpin path for map %s", b.name)
+	}
+	return nil
+}
+
+func (b *BPFMap) SetPinPath(pinning string) error {
```
itaysk, line 327: looks like this does the same as unpin? what did I miss here?

itaysk (commented):
can you please rename "pinning" to meaningful name, like "

libbpfgo/libbpfgo.go
ogozman: You did not miss anything, it was my fault, i did not push the change after testing

ogozman: Would "pinPath" be a suitable name?

yanivagman (approved):

grantseltzer:
This is really good to have, thanks @ogozman!

libbpfgo/libbpfgo.go
ogozman: I have freed pointer and changed the "pinning" to "pinPath"

itaysk (approved):

Issue #323, opened 2020-11-24
improve CI times


itaysk:
currently, on every pull request we build a new docker container with all the dependencies from scratch, and rebuild all the code.
If we could cache the go dependencies, and the docker layers, we would save most of the time and resources spent in the CI, the code build itself is not that intensive.
need to look into how to do that with Github Actions

krol3:
I´d would like to help

itaysk:
please do! take a look at the https://github.com/aquasecurity/tracee/blob/main/.github/workflows/build-test.yaml file and let me know if you have any questions

krol3:
@itaysk, to publish the image will be the 'make docker' that use this command `docker build --build-arg VERSION=v0.3.1-42-g3aa104c -t tracee:latest .`

itaysk:
@krol3 not sure I understand the question..
to clarify, this issue discusses the build/test phase, not the release (publish) phase

krol3:
Yes clarify the issue, maybe for the release this [PR](https://github.com/aquasecurity/tracee/pull/432) could help

itaysk:
ok, that's interesting. I can't accept that PR because publishing images is handled in our [current release script](https://github.com/aquasecurity/tracee/blob/7d2ce345dfca4f1f24f2aa1a0a99a97e5ad0952f/Makefile#L181), BUT:
I think your PR addresses also the caching issue, by using the `Cache Docker layers` action. If you can take just the caching parts, and integrate to the existing workflow, that would be helpful and resolve this issue.

rafaeldtinoco:
With the new building schema I believe this can be closed.

PR #439, opened 2021-01-24
Github action using cache

krol3:
Using cache with buildx, related the issue [323](https://github.com/aquasecurity/tracee/issues/323)


.github/workflows/build-test.yaml
```diff
@@ -15,12 +16,29 @@ jobs:
...
+
+      - name: Cache Docker layers
+        uses: actions/cache@v2
+        with:
+          path: /tmp/.buildx-cache
+          key: ${{ runner.os }}-buildx-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-buildx-
+
+      - name: Build tracee-builder
```
itaysk, line 35: this is unnecessary. we push the images from the release script

.github/workflows/build-test.yaml
```diff
@@ -15,12 +16,29 @@ jobs:
...
+
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Cache Docker layers
+        uses: actions/cache@v2
+        with:
+          path: /tmp/.buildx-cache
+          key: ${{ runner.os }}-buildx-${{ github.sha }}
```
itaysk, line 31: correct me if I'm wrong but won't this cache only between builds on the same commit?
we want to be able to cache across commits

itaysk (changes requested):

.github/workflows/build-test.yaml
```diff
@@ -29,16 +30,28 @@ jobs:
             exit 1
           fi
 
-      - name: Build Tracee
-        run: make all DOCKER=1
-  
-  test:
-    runs-on: ubuntu-20.04
+      - name: yaml-lint
```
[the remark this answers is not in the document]
krol3: It's only to validate .yaml format (empty spaces, etc), also if you want to check common misspellings in text files: codespell-project/actions-codespell.
I will remove this yaml-lint.

.github/workflows/build-test.yaml
krol3: Using this settings, we are doing cache across commits,
```
          cache-from: type=local,src=/tmp/.buildx-cache
          cache-to: type=local,dest=/tmp/.buildx-cache,mode=max
```

.github/workflows/build-test.yaml
```diff
@@ -1,9 +1,10 @@
+---
 name: build-test
 
 on: pull_request
 
 jobs:
-  build:
```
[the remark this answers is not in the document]
krol3: I go back to the origin/main branch

grantseltzer:
@itaysk Did you have further comment on this, and if not I believe this will be good to merge after @krol3 resolves conflict?

krol3:
I need to test it with the new updates. Please don't merge yet.

itaysk:
@grantseltzer issues weren't addressed. @krol3 no rush, thanks for working on this

.github/workflows/pr-ebpf.yaml
```diff
@@ -23,7 +23,29 @@ jobs:
             gofmt -s -d .
             exit 1
           fi
-      - name: Build
-        run: make all DOCKER=1
```
itaysk, line 27: the build step is now missing in the workflow isn't it?

krol3: Now, It's working
- First build time (2m 53s): https://github.com/krol3/tracee/actions/runs/541588005
- Second time using cache(1m 51s): https://github.com/krol3/tracee/actions/runs/541596189
The actually time in the job pr-ebpf (1m 45s)

itaysk: is it actually building the executable thought? i don't think so since you removed this command, but maybe I'm missing something

krol3:
@itaysk  using cache for the golang artifacts we gain some time in the pr-rules workflow.

image

krol3:
@itaysk I rerun the tracee-ebpf workflow. I have this result using the flag DOCKER_BUILDKIT=1

image

itaysk:
@krol3 isn't this missing the cache action? now the only change is to opt in to buildkit but how is the cache persisted between runs?

rafaeldtinoco:
@krol3 I believe this is outdated based on our recent efforts to have a new builder, new PR workflows, etc. I'm marking this as a draft so you can work this through, or close it, if you would like (just so we can better filter the on going PRs we have).

danielpacak:
I think it would be much easier to open a new PR once we are done with the repo restructure. We no longer split PR workflow to tracee-ebpf and tracee-rules.

PR #440, opened 2021-01-24
fix release workflow for github actions

itaysk:
by default GH actions does a shallow clone of just the latest commit. we need all history to build the changelog.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -9,6 +9,8 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
       - name: authenticate
         run: |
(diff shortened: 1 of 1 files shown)


knqyf263 (approved):
LGTM. Just to confirm, you need all history for all branches and tags, right?

itaysk:
no I need just "main", but in my understanding this is what it fetches, no?

> When checking out the repository that triggered a workflow, this defaults to the reference or SHA for that event

knqyf263:
The document says as follows.
>0 indicates all history for all branches and tags.

https://github.com/actions/checkout#Fetch-all-history-for-all-tags-and-branches

The explanation you shared looks like which branch to check out, not fetch. I didn't find a way to fetch all history of only the main branch, so your way might be the best at the moment.

PR #449, opened 2021-01-25
feat: add TracePrint to libbpfgo

lizrice:
Add function for printing out debug info written by the BPF helper bpf_trace_printk()

### DIFF
--- a/libbpfgo/libbpfgo.go
+++ b/libbpfgo/libbpfgo.go
@@ -164,7 +164,9 @@ err_out:
 import "C"
 
 import (
+	"bufio"
 	"fmt"
+	"os"
 	"strings"
 	"syscall"
 	"unsafe"
@@ -650,4 +652,29 @@ func (pb *PerfBuffer) poll() error {
 		}
 	}
 	return nil
+}
+
+// TracePrint reads data from the trace pipe that bpf_trace_printk() writes to,
+// and writes it to stdout. The pipe is global, so this function is not
+// associated with any BPF program. It is recommended to use bpf_trace_printk()
+// and this function for debug purposes only.
+// This is a blocking function intended to be called from a goroutine, for example:
+//		go libbpfgo.TracePrint()
+func TracePrint() {
+	f, err := os.Open("/sys/kernel/debug/tracing/trace_pipe")
+	if err != nil {
+		fmt.Println("TracePrint failed to open trace pipe: %v", err)
+		return
+	}
+	r := bufio.NewReader(f)
+	b := make([]byte, 1000)
+	for {
+		len, err := r.Read(b)
+		if err != nil {
+			fmt.Println("TracePrint failed to read from trace pipe: %v", err)
+			return
+		}
+		s := string(b[:len])
+		fmt.Println(s)
+	}
 }
--- a/libbpfgo/libbpfgo.go
+++ b/libbpfgo/libbpfgo.go
@@ -663,17 +663,20 @@ func (pb *PerfBuffer) poll() error {
 func TracePrint() {
 	f, err := os.Open("/sys/kernel/debug/tracing/trace_pipe")
 	if err != nil {
-		fmt.Println("TracePrint failed to open trace pipe: %v", err)
+		fmt.Fprintln(os.Stderr, "TracePrint failed to open trace pipe: %v", err)
 		return
 	}
+	defer f.Close()
+
 	r := bufio.NewReader(f)
-	b := make([]byte, 1000)
+	b := make([]byte, 1024)
 	for {
 		len, err := r.Read(b)
 		if err != nil {
-			fmt.Println("TracePrint failed to read from trace pipe: %v", err)
+			fmt.Fprintln(os.Stderr, "TracePrint failed to read from trace pipe: %v", err)
 			return
 		}
+
 		s := string(b[:len])
 		fmt.Println(s)
 	}


itaysk (commented):
added a couple of comments, but also: do we want this helpful utility in libbpf or in another package? wdyt?

itaysk (commented):
IIUC this prints each chunk of 1000 bytes in a line. wouldn't it be better to read until a newline, like the original trace_print function does?

libbpfgo/libbpfgo.go
```diff
@@ -651,3 +653,28 @@ func (pb *PerfBuffer) poll() error {
...
+// TracePrint reads data from the trace pipe that bpf_trace_printk() writes to,
+// and writes it to stdout. The pipe is global, so this function is not
+// associated with any BPF program. It is recommended to use bpf_trace_printk()
+// and this function for debug purposes only.
+// This is a blocking function intended to be called from a goroutine, for example:
+//		go libbpfgo.TracePrint()
+func TracePrint() {
+	f, err := os.Open("/sys/kernel/debug/tracing/trace_pipe")
+	if err != nil {
+		fmt.Println("TracePrint failed to open trace pipe: %v", err)
```
grantseltzer, line 666: Imo this should be printed to stderr instead

libbpfgo/libbpfgo.go
```diff
@@ -651,3 +653,28 @@ func (pb *PerfBuffer) poll() error {
...
+		return
+	}
+	r := bufio.NewReader(f)
+	b := make([]byte, 1000)
+	for {
+		len, err := r.Read(b)
+		if err != nil {
+			fmt.Println("TracePrint failed to read from trace pipe: %v", err)
```
grantseltzer, line 674: same here

libbpfgo/libbpfgo.go
itaysk: or return an error instead of printing it

libbpfgo/libbpfgo.go
```diff
@@ -651,3 +653,28 @@ func (pb *PerfBuffer) poll() error {
...
 }
+
+// TracePrint reads data from the trace pipe that bpf_trace_printk() writes to,
+// and writes it to stdout. The pipe is global, so this function is not
+// associated with any BPF program. It is recommended to use bpf_trace_printk()
+// and this function for debug purposes only.
+// This is a blocking function intended to be called from a goroutine, for example:
+//		go libbpfgo.TracePrint()
+func TracePrint() {
+	f, err := os.Open("/sys/kernel/debug/tracing/trace_pipe")
```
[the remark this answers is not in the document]
lizrice: Good point, I'll add a defer

libbpfgo/libbpfgo.go
lizrice: Can't use the return value if called from a goroutine. Could send it back in a channel, but that seems like overkill given this is really only suitable for very basic debug purposes (where you're probably logging to screen anyway). Will send it to stderr though, that's a good point.

itaysk: this one (unlike L674) can be returned, as it's determined before entering the loop

lizrice:
> IIUC this prints each chunk of 1000 bytes in a line. wouldn't it be better to read until a newline, like the original trace_print function does?

That was my first instinct too, to do it with a default Scanner that reads line-by-line, but it turns out that newlines don't get written to the pipe. 1000 is totally arbitrary, but Read will return whatever is available, won't exceed it so it's safe, and there will be no further allocations. 

Sometimes you get more than one "line" in the buffer at once like this: 

```
          <...>-106196  [001] .... 12309.540706: 0: I'm alive!           <...>-106196  [001] .... 12309.540709: 0: I'm alive!           <...>-106196  [001] .... 12309.540712: 0: I'm alive!  
```
We could parse the line to make it output more prettily, but I'm not sure it's worth the trouble for now.

libbpfgo/libbpfgo.go
lizrice: What I mean is the caller can't use the return value if they call it from a goroutine, which (per the doc comment) they would almost certainly want to do because it's blocking

itaysk: oh, I assumed you will take my suggestion to own the goroutine, in that case it would attempt to open the file before starting the goroutine

libbpfgo/libbpfgo.go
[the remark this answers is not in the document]
itaysk: I don't think this defer will ever be invoked, this function never gracefully terminates, does it?
if the entire app is killed then I guess it doesn't matter, but if it isn't, shouldn't there be a way to stop? and thereby release the file?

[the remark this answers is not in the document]
itaysk: to be clear, what I was suggesting in https://github.com/aquasecurity/tracee/pull/449#discussion_r564466691 is something like:
```go
func StartTracePrint(done chan struct{}) err {
  f,err := os.Open(...)
  if err!=nil { return err }
  defer f.Close()
  go func() {
    for {
      // if done is closed, break
      // print event
    }
  }
}
```

itaysk:
@lizrice I think we can merge this as is, despite my other comments which are mere suggestions.
But I do think it's better to put this in a new package, as this is not a libbpf wrapping, but an extra utility, wdyt?

> do we want this helpful utility under package libbpfgo or under another package in the libbpfgo module? wdyt? (I expect we will have more non-libbpf utils in the future)

lizrice:
I'm not so sure about the benefits of separate packages. As a consumer of this package, do I care whether it's tightly coupled to libbpf, or do I just want to be able to write eBPF code in Go? (For similar reasons I think there is code currently in Tracee that is more generally useful for eBPF apps, but my thoughts on that are not fully formed yet)

I do think it would be good if libbpfgo were split into multiple files, it's enormous at the moment, but that could be another PR.

itaysk:
ok.. if it's debatable then we can merge like this and maybe change in the future if needed

grantseltzer:
Can we merge this? @itaysk @lizrice

lizrice:
I believe @itaysk and @yanivagman are OK with it, but someone needs to add the review approval

grantseltzer (approved):
LGTM!

PR #454, opened 2021-01-26
Map pinning (potentially also other objects)

ogozman:
This is not really a complete pull request, more of a draft. Currently in this state i can pin various maps. However, a proper filter activation is also needed if, for example, we want to filter by entries in the map (in my use case inspector gadget creates map with mnt namespaces ids and we then trace only events in this namespaces, so, this is what currenty works). So, now i activate specific filter only if the map we are pinning is mnt_ns_filter. It would be great to receive some feedback on how i can improve general implementaiton so that we can merge additonal feature. Thank you

@kinvolk @alban @mauriciovasquezbernal


tracee/tracee.go
```diff
@@ -831,9 +831,15 @@ func (t *Tracee) initBPF(bpfObjectPath string) error {
 			case kprobe:
 				// todo: after updating minimal kernel version to 4.18, use without legacy
 				_, err = prog.AttachKprobeLegacy(probe.event)
+				if err != nil {
+					_, err = prog.AttachKprobe(probe.event)
```
alban, line 855: I think if it's better to do the reverse: try the new `AttachKprobe` and if it fails, fallback to `AttachKprobeLegacy`. In this way, the program would correctly be detached when the process terminates and it will help compatibility with other programs using the debugfs based kprobe, such as gobpf-based programs. See also: https://github.com/iovisor/gobpf/issues/223

And it would be more aligned on [the way bcc is doing it](https://github.com/iovisor/bcc/blob/6e9b4509fc7a063302b574520bac6d49b01ca97e/src/cc/libbpf.c#L1021-L1027).

You could also add a comment to explain why `AttachKprobeLegacy` might fail, such as
```
// Legacy kprobes have a system-wide name (chosen by libbpfgo.go poke_kprobe_events())
// that might conflict with other BPF programs unrelated to tracee. New perf based kprobes
// don't have this problem but require Linux 4.17.
```

ogozman: We have discussed this with @yanivagman in the previous PR and per his recommendation I have done it this way. But I can do it in reverse of course.

tracee/tracee.go
```diff
@@ -29,6 +29,9 @@ type TraceeConfig struct {
 	SecurityAlerts     bool
 	maxPidsCache       int // maximum number of pids to cache per mnt ns (in Tracee.pidsInMntns)
 	BPFObjPath         string
+	PinObjectType      string
```
yanivagman, line 32: Please use a "config" struct for all of these three (same as done for Capture and Output above).
I also think that this should be a slice of pin configs so we can support more pinning options in the future:

```
type TraceeConfig struct {
    ...
    BpfPinnings    []BpfPinConfig
}

type BpfPinConfig struct {
    ObjectType    string
    ObjectName    string
    Path          string
}
```

tracee/tracee.go
```diff
@@ -799,6 +807,18 @@ func (t *Tracee) initBPF(bpfObjectPath string) error {
 		}
 	}
 
+	if t.config.PinObjectName != "" && t.config.PinObjectType != "" && t.config.PinPath != "" {
+		if t.config.PinObjectType == "map" {
+			mapToPin, err := t.bpfModule.GetMap(t.config.PinObjectName)
+			if err != nil {
+				return err
+			}
+			mapToPin.SetPinPath(t.config.PinPath)
```
yanivagman, line 816: SetPinPath only work for maps that are already pinned (with some path), isn't it?
I think we should use the Pin() function (which should be called after BPFLoadObject())

tracee/tracee.go
yanivagman: @ogozman that's correct. In the original PR, you replaced AttachKprobeLegacy with AttachKprobe, so I mentioned that we need to support the legacy kprobe attach mechanism as we still support kernels 4.14-4.18.
I know that I wrote that we should default to the legacy attach mechanism, but it actually makes more sense to do the reverse, like @alban suggested.
Anyways, we are planning to bump the minimal kernel version to 4.18 soon, so that's only temporary.

tracee/tracee.go
ogozman: Yes, but then it does not work for our usecase. So, either we need two different usecases and the logic for each one accordingly

yanivagman: In that case, we don't want to pin a map (as it was already pinned by inspektor gadget), but to use an existing pinned map, right? If so,  the name "pin" that we gave to the flag is misleading, as a user may think that he can pin objects created by tracee (like maps, or programs in the future), but that's not the case.

And now I also wonder how is this even working? What you want to do here is to take a map created by tracee (by using the GetMap() function) and pin it to the same path as of the map created by inspektor gadget - how is that even possible? I think that I'm missing something

ogozman: I think @alban can explain it better, but as far as I understand bpf maps are used this way to communicate between various bpf programs. Inspector gadget uses this way of giving information to various tracers that are already in the library. I am not sure, but I suppose that bcc might not actually make any distinction between two cases (pinning map or linking to trhe map that already exists), but, apparently libbpf does. In our use case two maps point to the same file and are shared by two programs. This way gadget can dynamically change which mnt ns ids to trace

yanivagman: I don't think it's two maps, but one. Otherwise - how can you set the pin path of a map that still doesn't exist? libbpf creates the maps only when loading the object (https://github.com/libbpf/libbpf/blob/master/src/libbpf.c#L7489) which happens after this line

ogozman: Yes, which is why i call it before the loading of the bpf object. In this case map is "loaded" from the pinning path

yanivagman: What is the map name (PinObjectName) that you give to tracee? is it equal to the map name given by inspektor-gadget or to the mntns filter map defined by tracee?

ogozman: mnt_ns_filter defined by the tracee

yanivagman: So what you are saying is that you are using a map with a name defined by tracee, pinning it to the same path of another map which was already pinned by inspektor-gadget (with another name?), and updating the inspektor-gadget map successfuly filters by mntns in Tracee? I really don't know how to explain this, and I think we should fully understand what is going on here before we merge this change

mauriciovasquezbernal: > So what you are saying is that you are using a map with a name defined by tracee, pinning it to the same path of another map which was already pinned by inspektor-gadget (with another name?), and updating the inspektor-gadget map successfuly filters by mntns in Tracee? 

Yes, that's how it works. I was also very confused trying to understand this logic at the beginning. The key point here is that libbpf tries to reuse a map if it's already pinned. 

In order to do so, the pin path of the map has to be set **before** loading the bpf object, then libbpf will reuse that map. It performs some checks to be sure the definition of both maps is the same (type, key size, value size, ...). This "magic" happens here https://github.com/libbpf/libbpf/blob/767d82caab7e54238f2fc6f40ab1e4af285f2abe/src/libbpf.c#L4308 :)

yanivagman: I see, indeed a "magic".
I wonder if we can assume that in the future this won't change (https://github.com/libbpf/libbpf/blob/767d82caab7e54238f2fc6f40ab1e4af285f2abe/src/libbpf.c#L4100) and libbpf will not also check for map_name equality (which is constant for tracee, and variable for inspektor-gadget).
I can also see that max_entries field is checked as well. What is the value for this in your map? We need to have the same value here for this to work.

mauriciovasquezbernal: > libbpf will not also check for map_name equality 

I don't think having the same map_name is important. map_name is just the name of a variable in the BPF program, it's replaced at load time by the file descriptor of the map. 

> I can also see that max_entries field is checked as well. What is the value for this in inspektor-gadget map? We need to have the same value here for this to work.

It's already been updated to match : https://github.com/kinvolk/inspektor-gadget/pull/159.

yanivagman: Thanks @mauriciovasquezbernal 
I'm ok proceeding with this method, but we need to take into account that if either tracee or inspektor-gadget change one of these parameters (map size and type, key and value sizes) this feature will stop working.

yanivagman: Also, correct me if I'm wrong but the path given to Tracee "pin" flag have to be of a valid pinned map, right?
If so, I think we should rename this flag to be "bpf-pin reuse:map:_map_name_:_path_" (or maybe you have a better suggestion), because the functionality of it is not "pinning" a map of Tracee, but reusing an existing map

yanivagman: BTW, I see that there is also a libbpf API to directly reuse a bpf map:
https://github.com/libbpf/libbpf/blob/767d82caab7e54238f2fc6f40ab1e4af285f2abe/src/libbpf.c#L3640
Maybe this can help avoiding the above checks

mauriciovasquezbernal: > I'm ok proceeding with this method, but we need to take into account that if either tracee or inspektor-gadget change one of these parameters (map size and type, key and value sizes) this feature will stop working.

Yes, this is something we have to be aware of. 

> Also, correct me if I'm wrong but the path given to Tracee "pin" flag have to be of a valid pinned map, right?

Yes, in this case (when used with Inspektor Gadget) Tracee will reuse the map. I wondering if the use case where tracee creates and pins the maps and another external tool opens and populates it is important. If we only want Tracee to reuse the map then I think we should rename the flag, otherwise if we want to allow it to create it then the current name seems good to me. 

I'm a bit concerned because if the user intention is to reuse the map and passes a wrong path, Tracee will create a new one without giving any error and it's going to be fun to debug. I'm not sure how we could improve it, is Tracee printing the debug of libbpf with some debug flag is passed?

itaysk: @mauriciovasquezbernal I posed my feedback about reuse vs create before reading your comment, which explains it. Interesting use case we need to think about if it's important but I suggest to start without and maybe add it in the future.

we do have --debug flag which can be used for this purpose.

itaysk: @ogozman can you please add a comment in the code warning from making changes to the map spec and linking to the compatibility checks in libbpf please?

yanivagman: > Yes, in this case (when used with Inspektor Gadget) Tracee will reuse the map. I wondering if the use case where tracee creates and pins the maps and another external tool opens and populates it is important. If we only want Tracee to reuse the map then I think we should rename the flag, otherwise if we want to allow it to create it then the current name seems good to me.
> 
> I'm a bit concerned because if the user intention is to reuse the map and passes a wrong path, Tracee will create a new one without giving any error and it's going to be fun to debug. I'm not sure how we could improve it, is Tracee printing the debug of libbpf with some debug flag is passed?

@mauriciovasquezbernal Are you sure that it creates a map if the path is an invalid path? I see in libbpf that it fails if the fd is negative (path doesn't exist), isn't it? https://github.com/libbpf/libbpf/blob/767d82caab7e54238f2fc6f40ab1e4af285f2abe/src/libbpf.c#L4114

> @yanivagman Don't you think it belongs under the --trace flag? it does configure the mnt ns tracee filter for tracee (IIUC). We already have --tarce mntns=1234 we can add a new operator that reads from a pinned map like --trace mntns=map:pathtomap or similar.

@itaysk I don't know about this. We defined the trace flag to select which events to trace, and here we want to do another thing. Unlike mntns=1234, here we define a pinning (or reuse) functionality, that doesn't have to affect the traced events.
bpf-reuse is a good name IMO.

itaysk: > here we define a pinning (or reuse) functionality, that doesn't have to affect the traced events.

isn't the entire purpose of the pin is to define a which mnt ns to trace?

yanivagman: > isn't the entire purpose of the pin is to allow IG to select which mnt ns to trace?

Yes, you are right. Maybe I'm over-thinking this as an option to reuse other maps as well in the future, but in this case, that can work too.

itaysk: ok.. I see. I think that even if we will reuse other maps in the future, we will need to find how to incorporate their new functionality in the existing UX. this PR is about filtering functionality to I think that the map pin should be under the relevant flag, if we want to next reuse a pin to control something in the output, it should be under the output flag. If it will be something entirely new, the name of the flag should describe the new functionality, and not how it is implemented

ogozman: > @ogozman can you please add a comment in the code warning about making changes to the map spec and linking to the compatibility checks in libbpf please?

@itaysk I am not surre what you mean, I did not change any defintions of the maps in tracee. We changed them in inspector-gadget to match the number of entries

mauriciovasquezbernal: > @mauriciovasquezbernal Are you sure that it creates a map if the path is an invalid path? I see in libbpf that it fails if the fd is negative (path doesn't exist), isn't it? https://github.com/libbpf/libbpf/blob/767d82caab7e54238f2fc6f40ab1e4af285f2abe/src/libbpf.c#L4114

There's an exception for `-ENOENT` two lines below that, in that case it returns 0 and the map is created.

yanivagman: @mauriciovasquezbernal I understand your concern about tracee creating the map even if not reused.
We currently don't print debug messages of libbpf, only WARN level (https://github.com/aquasecurity/tracee/blob/main/libbpfgo/libbpfgo.go#L26). We can change this behavior using the --debug flag. If debug is set, we can set the print function level to LIBBPF_DEBUG (this can be done by giving the set_print_fn an extra "level" argument, which will choose another print function that prints all libbpf messages). It's a good idea to have this functionality anyways, so I'll add a task for doing this.

Another option to avoid this situation is by checking if the path indeed exists by calling `bpf_get_obj` directly (https://github.com/libbpf/libbpf/blob/42baefba714b2cce03fbfd68ecfff507a80adc1d/src/bpf.c#L563) and checking if `pin_fd<0` just like libbpf does (then using `close(pin_fd)` after the check), then return an error if it doesn't exist

itaysk: > Another option to avoid this situation is by checking if the path indeed exists by calling

I agree this is preferred. also noted it earlier:

> we do have --debug flag which can be used for this purpose but this can also be validated before using and error if the path doesn't exist (if we choose to not allow pinning map from tracee)

itaysk:
Hi @ogozman , saw your latest commit (`updated readme`) which I didn't quite understand. are you still planning to complete this PR eventually? (no pressure, just asking)

ogozman:
> Hi @ogozman , saw your latest commit (`updated readme`) which I didn't quite understand. are you still planning to complete this PR eventually? (no pressure, just asking)

Since I am no longer working on the project and this request clearly needs a lot of further work, I will close it.

PR #457, opened 2021-01-26
restructure repo

itaysk:
This PR implements the restructure of the repo as agreed.
A followup to the restructure is to refactor the release process which I'm leaving to a future PR to make this one reviewable.
TL;DR of the change:
- moved all the tracee-ebpf stuff into it's dir
- edited the readme appropriately
- edited the makefile to generate tracee-ebpf
- edited the GH workflows to work with the change
- update git submodule with the new location


itaysk:
After this PR is merged we should merge this branch to main

itaysk:
@yanivagman @simar7 @grantseltzer I think we should also freeze all work on tracee-ebpf until this is merged to avoid unnecessary merge conflicts. we can keep working on libbpfgo because it's unchanged in the restruct

itaysk:
I see that there's a problem with libbpf submode. I'm checking that now

yanivagman:
I think you should rebase on main before doing this change to avoid merge conflicts when we will merge to main

itaysk:
@yanivagman this is already rebased on main.

itaysk:
Whatever you prefer

yanivagman:
I think that it's better to do this after the merge. No code freeze will be needed.
Is there a reason not to merge now the tracee-rules branch?

tracee-ebpf/README.md
```diff
@@ -28,32 +28,32 @@ Exceptions:
 ### Quickstart with Docker
 
 ```bash
-docker run --name tracee --rm --privileged --pid=host -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro -v /tmp/tracee:/tmp/tracee aquasec/tracee:latest
+docker run --name tracee --rm --privileged --pid=host -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro -v /tmp/tracee:/tmp/tracee aquasec/tracee:latest ebpf
```
simar7, line 31: is ebpf an arg now?

simar7:
> I think that it's better to do this after the merge. No code freeze will be needed.
> Is there a reason not to merge now the tracee-rules branch?

Merging tracee-rules should make things easier as well. I'm currently working on an improvement PR for tracee-rules but I probably won't be able to land it in until tracee-rules branch is merged on to main. 

It would also require a rebase each time (of my feature branch) to ensure it can land since parent branch (tracee-rules) isn't merged just yet.

tracee-ebpf/README.md
itaysk: not yet, but we should have as part of the restructure a subcommand (command name TBD)

PR #459, opened 2021-01-27
Restructure repo as monorepo

itaysk:
The git repo is now home to: tracee-ebpf (what tracee was until now),
tracee-rules, and libbpfgo.
In the root we will have general resources like docs/tests/packaging.
There is still work to do with regactoring the release process.


Makefile
```diff
@@ -1,193 +0,0 @@
-.PHONY: all
```
yanivagman, line 1: Why does it show that his Makefile was deleted and recreated in the tracee-ebpf dir?
Can you use git mv instead?

tracee-ebpf/README.md
```diff
@@ -28,32 +28,32 @@ Exceptions:
 ### Quickstart with Docker
 
 ```bash
-docker run --name tracee --rm --privileged --pid=host -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro -v /tmp/tracee:/tmp/tracee aquasec/tracee:latest
+docker run --name tracee --rm --privileged --pid=host -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro -v /tmp/tracee:/tmp/tracee aquasec/tracee:latest ebpf
```
yanivagman, line 31: Do you want to update this readme before we actually add the ebpf subcommand?

tracee-ebpf/README.md
```diff
@@ -5,11 +5,11 @@
 [![License](https://img.shields.io/github/license/aquasecurity/tracee)](https://github.com/aquasecurity/tracee/blob/main/LICENSE)
 [![docker](https://badgen.net/docker/pulls/aquasec/tracee)](https://hub.docker.com/r/aquasec/tracee)
 
-# Tracee - Container and system tracing using eBPF
+# Tracee-eBPF - Container and system tracing using eBPF
```
yanivagman, line 8: Moving this readmen without creating a new one in the root directory will leave the repo without a readme

itaysk: I don't want to start debating the readme in this pr. Will add in future pr

tracee-ebpf/README.md
itaysk: You are right i will remove for now

tracee-ebpf/README.md
yanivagman: It's not about debating the readme - it's about having one for whoever enters the repo for the first time while there is no readme in the root dir

yanivagman: I think that if we don't have another readme yet, we should keep this one until we will have.
Another option - open a readme which links to tracee-ebpf and tracee-rules readmes

itaysk: I will add something minimal then that points to tracee ebf and rules

itaysk:
I have updated the pr to be less disruptive, just focus on the directory move to make it easier to review. the rest we will do in a future pr

Readme.md
```diff
@@ -0,0 +1,12 @@
+![Tracee Logo](images/tracee.png)
+
+# Tracee: Runtime security and forensics using eBPF
+
+Tracee is a Runtime Security and forensics tool for Linux. It is using Linux eBPF technology to trace your system and applications at runtime, and analyze collected events to detect suspicious behavioral patterns.
+
+This repo contains the following projects:
+- [Tracee-eBPF](tracee-ebpf) - Container and system tracing using eBPF
```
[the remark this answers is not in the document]
yanivagman: No change here - do you prefer to keep it as is?

Readme.md
```diff
@@ -0,0 +1,12 @@
+![Tracee Logo](images/tracee.png)
+
+# Tracee: Linux Runtime security and forensics using eBPF
```
yanivagman, line 3: Runtime -> runtime (or capitalize all words in the headline)

yanivagman (commented):
Why did you move 3rdparty/libbpf → tracee-ebpf/3rdparty/libbpf?
libbpf is used by libbpfgo, which is under root, isn't it?

itaysk:
> Why did you move 3rdparty/libbpf → tracee-ebpf/3rdparty/libbpf?
libbpf is used by libbpfgo, which is under root, isn't it?

not exactly: libbpfgo is is just go bindings to libbpf C library the user need to obtain somehow. we are not (can not) shipping libbpf as part of libbpfgo, we tell the user to either install it (apt-get install libbpf-dev) or build it. Tracee, as a user of libbpfgo, chooses to vendor libbpf and build it as part of it's build script, that's why it's under Tracee-eBPF. the example the Liz created, and which we would like to add to this repo, is actually installing libbbpf and not building it.

Readme.md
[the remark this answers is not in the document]
itaysk: my bad. fixed

yanivagman:
These readme lines should move to the readme in the root dir:
[GitHub release (latest by date)](https://github.com/aquasecurity/tracee/releases)
[Go Report Card](https://goreportcard.com/report/github.com/aquasecurity/tracee)
[License](https://github.com/aquasecurity/tracee/blob/main/LICENSE)
[docker](https://hub.docker.com/r/aquasec/tracee)

yanivagman (approved):

itaysk:
are you sure? for now the "tracee" artifacts (bins and images) relate only to tracee-ebpf. so I thought it's better to leave it there until we refactor the release as well.

yanivagman:
All of these lines give the address of the tracee repo, and not the eBPF submodule - so yes, I think it should move there

PR #477, opened 2021-01-28
feat: Add tests for core engine functionality

simar7:
This PR adds tests for some of the core engine functionality. 


simar7:
Looks like the CI runner only runs the tests in `tracee-ebpf` dir. Any ideas how to run tests in all packages (tracee-rules, etc.)?

itaysk:
https://github.com/aquasecurity/tracee/pull/479 please review

go.mod
```diff
@@ -0,0 +1,14 @@
...
+replace github.com/aquasecurity/tracee/libbpfgo => ./libbpfgo
+
+require (
+	github.com/aquasecurity/tracee/libbpfgo v0.0.0-00010101000000-000000000000
+	github.com/aquasecurity/tracee/tracee-rules v0.0.0-20210127115517-c43b1c3394ec // indirect
+	github.com/iovisor/gobpf v0.0.0-20200529092446-49b58e11a4b5
+	github.com/stretchr/testify v1.5.1
+	github.com/syndtr/gocapability v0.0.0-20180916011248-d98352740cb2
+	github.com/urfave/cli/v2 v2.3.0
+)
```
simar7, line 14: I see top level tracee dir doesn't have a go.mod. Should we add go.mod and go.sum on the top level .gitignore file? https://github.com/aquasecurity/tracee/blob/main/.gitignore

Regardless, I'll remove this as it doesn't belong here.

tracee-rules/signature.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/aquasecurity/tracee/tracee-rules/types"
 )
 
-func getSignatures(rulesDir string, rules []string) ([]types.Signature, error) {
+func GetSignatures(rulesDir string, rules []string) ([]types.Signature, error) {
```
simar7, line 15: I can swing either way - for with or without this change to export this function. I just thought it'd be easier in the future to use as an abstraction.

go.mod
itaysk: no.. root dir doesn't need a go.mod since it's not a module

tracee-rules/signature.go
itaysk: fine with me, but how can someone practically import this?

itaysk: I mean import main, from another package in this module?

simar7: Actually your second comment made me think again - yes, e.g. an integration test. But I think it's better to do that when the need arises. I'll remove this for now.

go.mod
simar7: hmph - I thought my IDE had added it in but it doesn't seem to be the case. Oh well cruft anyways - I've removed it. https://github.com/aquasecurity/tracee/pull/477/commits/91e527ce78084fb47118c00061276e70d5814fa2

tracee-rules/engine/engine.go
```diff
@@ -68,8 +68,8 @@ func NewEngine(sigs []types.Signature, sources EventSources, output chan types.F
 func (engine *Engine) Start(done chan bool) {
 	go engine.consumeSources(done)
 	for s, c := range engine.signatures {
-		defer close(c)
 		go func(s types.Signature, c chan types.Event) {
+			defer close(c)
```
itaysk, line 72: what's the motivation for this change? the current state feels more reasonable to me

tracee-rules/engine/engine.go
```diff
@@ -107,7 +107,6 @@ func (engine *Engine) consumeSources(done <-chan bool) {
 						}
 					}
 				}
-				engine.inputs.Tracee = nil // TODO: Setting the channel to nil would block any future writes https://play.golang.org/p/jWyYwjNK_NT
```
itaysk: what's the motivation for this change? this line is preventing the containing loop from continuing infinitely and wasting cycles. I'm open to another solution but we need to be mindful of these changes

tracee-rules/engine/engine.go
```diff
@@ -102,12 +102,11 @@ func (engine *Engine) consumeSources(done <-chan bool) {
 					}
 					for _, sel := range se {
 						if sel.Source == "tracee" {
-							sig.OnSignal(types.SignalSourceComplete("tracee"))
+							sig.OnSignal(types.SignalSourceComplete("tracee")) // TODO: Handle error?
```
itaysk, line 105: this is intentional because if the sig errors we just ignore it. if you want to log the error that's fine as well

tracee-rules/engine/engine_test.go
```diff
@@ -0,0 +1,250 @@
...
+		expectedNumEvents int
+		expectedError     string
+	}{
+		{
+			name: "happy path - with one matching selector",
+			inputEvent: types.TraceeEvent{
+				EventName: "test_event",
+				ProcessID: 2, ParentProcessID: 1,
+				Args: []types.TraceeEventArgument{{
+					Name: "test_foo",
```
itaysk, line 74: it isn't realistic for an arg to have no value. even though it's not enforced better to put something here or remove the entire arg (it's not used is it). same in all other place

tracee-rules/engine/engine_test.go
```diff
@@ -0,0 +1,250 @@
...
+
+			if tc.expectedNumEvents <= 0 {
+				assert.Nil(t, gotEvent, tc.name)
+				assert.Zero(t, gotNumEvents, tc.name)
+			} else {
+				assert.Equal(t, tc.expectedNumEvents, gotNumEvents, tc.name)
+			}
+
+			if tc.expectedError != "" {
+				assert.Equal(t, tc.expectedError, logger.String(), tc.name)
```
itaysk, line 246: I'm not sure we should test the log output. it means that anytime anyone wants to log something, they need to update all the assertions that were made against the log. e.g: let's say that start will notify when it's starting "engine has started" to the log, this innocent change will affect tests to other functions (like this one). I also don't consider the log as a way to return the error. AFAIC there's no error returned.

tracee-rules/signature.go
```diff
@@ -44,6 +44,7 @@ func getSignatures(rulesDir string, rules []string) ([]types.Signature, error) {
 	return res, nil
 }
 
+// TODO: Use filepath.Walk() to recursively scan
```
itaysk, line 47: please create an issue for this feature request and remove this from the code.. (same for the other similar comment)

itaysk (commented):
thanks @simar7 ! I've added a few comments but also, I think the last commit (https://github.com/aquasecurity/tracee/pull/477/commits/dc437b3aa5a35e243033b889a9282e7e46c9d92d) doesn't belong in this PR as it's a changing to the business logic, not just the tests. I also think that commit is unnecessary, but we can discuss this separately from the tests PR.

tracee-rules/engine/engine.go
```diff
@@ -102,12 +102,11 @@ func (engine *Engine) consumeSources(done <-chan bool) {
...
 							break
 						}
 					}
 				}
-				engine.inputs.Tracee = nil
```
itaysk, line 110: what's the motivation for this change? this line is preventing the containing loop from continuing infinitely and wasting cycles. I'm open to another solution but we need to be mindful of these changes

itaysk (commented):
idk this was omitted from the original review but it's important comment

go.mod
itaysk: also remove top level go.sum :)

tracee-rules/engine/engine.go
simar7: The current implementation will panic if an incorrectly written signature erroneously sends an event on the channel after signalling `done`. Here's a test to show this behaviour: https://gist.github.com/simar7/d9ec8ef5df938b1b8a4dc86d692f5076

In addition, `c` is a channel that's read by the engine (Start func acts as a consumer) and created by someone else https://github.com/aquasecurity/tracee/blob/efdc613024f04f45b4be877372bd474f31da890d/tracee-rules/main.go#L34-L39

So one can also debate that we shouldn't close at all, which is something that crossed my mind as I write this. 

What do you think?

tracee-rules/engine/engine.go
simar7: consumeSources is just a consumer of the channel. In this case if the following happens:

1. Engine is running and processing events
2. Producer closes channel but does *not* send `done`
3. Engine will assign `nil` to the input channel
4. Producer will open the channel (recreate) and try to send another event
5. Engine will continue to wait and do nothing

In this case if we don't set `engine.inputs.Tracee = nil` when the producer does step 4. the engine will automatically pick it up and continue on with the newly created channel.

Here's a test if you want to see it in action: https://gist.github.com/simar7/0576c9d5d61c3f6785883b1cf936769f

tracee-rules/engine/engine_test.go
simar7: IMHO logging should be tested as it's part of the code and in certain cases (not here) include undesirable items (secrets being logged, etc.) which should be tested and guarded against. Another point would for when we use external libraries to log, they could panic in runtime. Chance of that are low happening reduce, if logging was tested.

I agree my approach isn't correct here, we should decouple casual logging from actual errors being logged or returned. I'll take it off from this PR and when we get to the point of having structured and levelled logging, re-introduce it appropriately.

tracee-rules/engine/engine.go
itaysk: > incorrectly written signature sends an event

signatures don't send events, they receive them. only input sources send events (in package main). so for your test to repro IRL main needs to have a bug which it closed the engine and then sent an event to the closed engine.

> panic if ... sends an event on the channel after signalling done
this behavior sounds good to me. done means the engine is done and should not process any further events. perhaps an error would be better than panic, but given both main and engine are under our control, and this scenario is so unlikely to repro, I don't see sufficient motivation to improve it.

Also - moving the defer to the goroutine is equivalent to removing it altogether. the defer will never be called where you put it, which is why it doesn't panic in your test.

> c is a channel that's read by the engine (Start func acts as a consumer) and created by someone else

actually this is not the case.. I think you are confusing `c`, which is the signature's dedicated channel and is [created by engine](https://github.com/aquasecurity/tracee/blob/4fdcba8ad7ad51f7bff77faed1add657ecbbf2fa/tracee-rules/engine/engine.go#L34) and owned by it, with `engine.inputs.xxx` channel, which indeed is created by main and isn't owned by engine.

tracee-rules/engine/engine.go
itaysk: > 4. Producer will open the channel (recreate) and try to send another event

This is impossible. Producer, which is outside `package engine` (in our case `package main`), cannot manipulate `engine.inputs`.

simar7: I see your point about abstraction. You're right, I'll revert this. I also don't know how to prevent a busy loop within a select waiting on channels but it's something I'm interested in learning more about.

tracee-rules/engine/engine.go
simar7: > > incorrectly written signature sends an event
> 
> signatures don't send events, they receive them. only input sources send events (in package main). so for your test to repro IRL main needs to have a bug which it closed the engine and then sent an event to the closed engine.
> 

Correct. I do agree the chances of that happening are low as we should limit the amount of logic we write in main, as it's hard to test for.

> > panic if ... sends an event on the channel after signalling done
> > this behavior sounds good to me. done means the engine is done and should not process any further events. perhaps an error would be better than panic, but given both main and engine are under our control, and this scenario is so unlikely to repro, I don't see sufficient motivation to improve it.
> 
> Also - moving the defer to the goroutine is equivalent to removing it altogether. the defer will never be called where you put it, which is why it doesn't panic in your test.
> 

You're right - I didn't realize this.

> > c is a channel that's read by the engine (Start func acts as a consumer) and created by someone else
> 
> actually this is not the case.. I think you are confusing `c`, which is the signature's dedicated channel and is [created by engine](https://github.com/aquasecurity/tracee/blob/4fdcba8ad7ad51f7bff77faed1add657ecbbf2fa/tracee-rules/engine/engine.go#L34) and owned by it, with `engine.inputs.xxx` channel, which indeed is created by main and isn't owned by engine.

Gotcha. I understand it now. I'll revert this as well.

tracee-rules/engine/engine_test.go
simar7: Yep - this is cruft and should be removed.

tracee-rules/engine/engine_test.go
simar7: > I'm not sure we should test the log output. it means that anytime anyone wants to log something, they need to update all the assertions that were made against the log. e.g: let's say that start will notify when it's starting "engine has started" to the log, this innocent change will affect tests to other functions (like this one). I also don't consider the log as a way to return the error. AFAIC there's no error returned.

Thinking it with a fresh mind - I think there are two different cases at play here:

1. Places that just purely log and drop the error: https://github.com/aquasecurity/tracee/blob/a328618462e2b0697d291ea5c94489d0b9d7908b/tracee-rules/engine/engine.go#L76
2. Places that don't drop the error and take a conditional branch: https://github.com/aquasecurity/tracee/blob/a328618462e2b0697d291ea5c94489d0b9d7908b/tracee-rules/engine/engine.go#L98

For 1. I agree we can get away with  not testing at this time but for 2. I feel it's important to have a test for as its conditionally executed. Not testing it would mean any future logic added in the conditional branch will be untested. Thoughts?

As far as assertions go, it makes sense to loosen it a little by doing a `assert.Contains` rather than a `assert.Equal` as I have here. That way we won't be breaking tests on each new log line added that might be unrelated to the test itself as you mentioned.

tracee-rules/engine/engine.go
simar7: Dropping it is fine by me as it's not logic that belongs to the engine anyway (it's a signature error). I'll make that change.

tracee-rules/signature.go
simar7: Roger https://github.com/aquasecurity/tracee/issues/501

tracee-rules/engine/engine.go
simar7: Replied & fixed in another comment.

go.mod
simar7: Are you talking about this? https://github.com/aquasecurity/tracee/pull/477/files#diff-bddf0b78668bac130d388e24bf8daf6fa0db604d2dc07a78aca6806985376f0c Or something else?

tracee-rules/engine/engine_test.go
itaysk: personally I don't consider this kind of log a spec'ed functionality, so I wouldn't test it because no one says what's good or bad. It's just going to be like a snapshot test which tells us if the result has changed, not if it broke. If you strongly feel it should be tested, it's fine with me, as long as it doesn't hinder development as with the scenario we described earlier

tracee-rules/engine/engine.go
itaysk: thanks. if you want to investigate it further, put a print in the `consumeSources` where it handles the closed channel, then you can "see" the busy loop

go.mod
itaysk: yeah my bad

tracee-rules/engine/engine_test.go
```diff
@@ -0,0 +1,238 @@
...
+		inputEvent        types.TraceeEvent
+		inputSignature    fakeSignature
+		expectedNumEvents int
+		expectedError     string
+	}{
+		{
+			name: "happy path - with one matching selector",
+			inputEvent: types.TraceeEvent{
+				EventName: "test_event",
+				ProcessID: 2, ParentProcessID: 1,
```
yanivagman, line 72: newline? (for consistency with below tests)

tracee-rules/engine/engine_test.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+func (fs fakeSignature) OnSignal(signal types.Signal) error {
+	if fs.onSignal != nil {
+		return fs.onSignal(signal)
+	}
+	return nil
+}
+
+func TestEngine_consumeSource(t *testing.T) {
```
yanivagman, line 60: Shouldn't it be called TestEngineConsumeSource?

yanivagman (approved):
Nice!
GJ @simar7

tracee-rules/engine/engine_test.go
itaysk: actually it should be `TestConsumeSources`. engine is the package and file being tested

tracee-rules/engine/engine_test.go
simar7: Good observation - Actually these fields populated in a test or not makes no difference functionally. I'll keep them in the first happy case but remove from the rest to keep test cases simple.

tracee-rules/engine/engine_test.go
simar7: Roger that - I handled it here: https://github.com/aquasecurity/tracee/pull/477/commits/2dfb7cacb6c9bb50e1505f5bd4dd1a6580d3ac0e

tracee-rules/engine/engine_test.go
simar7: Updated https://github.com/aquasecurity/tracee/pull/477/commits/2cd3c45e11a781604fb0d5d261cc31e7d420777a

simar7:
I had to rebase on top of latest main so could you just double check if all is well @itaysk before I press the big green button to merge?

itaysk (approved):
thanks @simar7 LGTM
please squash before merging

Issue #450, opened 2021-01-25
Throw an error if pid/uid<0 are provided as input filters


simar7:
**What did I see:**
`tracee --trace 'uid<0'` started tracing normally
`tracee --trace 'pid<0'` started tracing normally

**What did I expect to see:**
`tracee --trace 'uid<0'` should throw an error as it's an invalid filter.
`tracee --trace 'pid<0'` should throw an error as it's an invalid filter.

**Acceptance criteria:**
`tracee --trace 'uid<0'` throws an invalid filter error.
`tracee --trace 'pid<0'` throws an invalid filter error.

itaysk:
This is requires a little bit further discussion about whether the "preparer" functions should do input validation (I think no), or just parse the input and let the config validate the parsed struct. 
A good example is from `prepareOutput` which I explicitly mention here:  https://github.com/aquasecurity/tracee/blob/main/main_test.go#L1023
`prepareOutput` will blindly parse `--output foo` into an `OutputConfig.Format: foo`, and then rely on `OutputConfig.Validate()` to do the validation for better code reuse. I think this approach is better but open for other opinions.

PR #478, opened 2021-01-28
Update contributing guidelines (aka team agreements)

itaysk:


CONTRIBUTING.md
```diff
@@ -1,22 +1,56 @@
...
+
+1. We have the following discussion topics: 
+  1. [Announcements](https://github.com/aquasecurity/tracee/discussions/categories/announcements): One way communication from the team to the community. Consider this like our mini blog
+  2. [Questions and Help](https://github.com/aquasecurity/tracee/discussions/categories/questions-and-help): For help and support. Consider this similar to StackOverflow.
+  3. [Development](https://github.com/aquasecurity/tracee/discussions/categories/development): For discussing potential features, and collaborating on their design.
+
+### Issues:
+
+1. Every issue needs to be actionable and assignable. Consider the scope of the issue if assigned to one person, and break down if necessary.
+1. Be clear and definitive when composing issues. For bug reports, include detailed error messages and environment description. For features, include a clear scope and acceptance criteria.
```
simar7, line 24: Should we add an example? People might not be familiar with the terms scope or acceptance criteria. This seems simple enough to be one https://github.com/aquasecurity/tracee/issues/450

grantseltzer (approved):
I like Simar's idea but otherwise this looks good to me

CONTRIBUTING.md
```diff
@@ -1,22 +1,56 @@
...
+  1. Try to mention the area that this commit affects (see area labels above).
+1. Body style: 
+  1. Wrap at 72 chars.
+1. Code related information should be in commit message, review related information should be in PR description.
+
+## Code
+
+1. Follow Golang's code review standards: https://github.com/golang/go/wiki/CodeReviewComments.
+1. Follow `gofmt` + `govet` + `goimports` formatting.
+1. Tests should be considered an integral part of the PR. For parts that are harder to test, and are not currently tested, it's not required.
```
simar7, line 56: Should we say all new code should have tests to go along with it unless it's impossible to test?

simar7, line 56:
```suggestion
1. Tests should be considered an integral part of the PR. Tests should be added, wherever applicable, when adding new functionality. For bug fixes, if the corresponding code block has a test in place, consider adding a test case to guard against regression for the bug being fixed. For parts that are harder to test, and are not currently tested, it's not required.
```

CONTRIBUTING.md
```diff
@@ -1,22 +1,56 @@
...
+  1. Written in imperative mood, try to begin the sentence with either "Fix" or "Add".
+  1. Try to mention the area that this commit affects (see area labels above).
+1. Body style: 
+  1. Wrap at 72 chars.
+1. Code related information should be in commit message, review related information should be in PR description.
+
+## Code
+
+1. Follow Golang's code review standards: https://github.com/golang/go/wiki/CodeReviewComments.
+1. Follow `gofmt` + `govet` + `goimports` formatting.
```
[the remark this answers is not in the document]
yanivagman: As you probably know, there are no automatic formatting tools or absolute conventions in C.
I wrote the bpf code according to some conventions that I'm used to.
Maybe we should define some conventions to use, but there aren't now, so the best practice for now is to just follow the existing conventions when adding new code.

yanivagman:
Numbering is all "1"

CONTRIBUTING.md
grantseltzer: I would change 'consider adding' to 'you should add'. I don't see any reason not to.

CONTRIBUTING.md
```diff
@@ -1,22 +1,56 @@
...
+1. Be clear and definitive when composing issues. For bug reports, include detailed error messages and environment description. For features, include a clear scope and acceptance criteria.
+1. Since we have different projects under the same monorepo, use labels to denote areas that the issue relates to:
+  1. `tracee-ebpf`
+  2. `tracee-rules`
+  3. `libbpfgo`
+  4. `signatures`
+  5.  If non of the labels is relevant don't add any (usually for top-level issues)
+1. We use the following labels to describe the type of issue:
+  1. `bug`
+  2. `good-first-issue`
```
yanivagman, line 33: Don't we want a label to denote a feature as well?

yanivagman (commented):
I would also add a comment about not mixing commits from different areas.
For example, a PR which adds some new functionality to libbpfgo shouldn't use the added functionality in tracee-ebpf

CONTRIBUTING.md
itaysk: if we had this label it would be added to almost every new issue (except bugs) which makes it redundant. that's why we removed the "enhancement" label

itaysk:
> I would also add a comment about not mixing commits from different areas.
For example, a PR which adds some new functionality to libbpfgo shouldn't use the added functionality in tracee-ebpf

You started by saying commits and ended by saying PRs. I agree commits can be separated, but PRs, which are composed of multiple commits, can address an issue across areas.

CONTRIBUTING.md
yanivagman: Yet, I think it is not redundant.
Not all enhancements are new features, which make it less common.
I also find it useful to be able to filter by new feature requirements.

yanivagman:
`You started by saying commits and ended by saying PRs. I agree commits can be separated, but PRs, which are composed of multiple commits, can address an issue across areas`
This is on purpose. I don't think that the same PR should mix commits from different areas. There might be some exception though (like the case you noted of addressing a bug across different areas)

CONTRIBUTING.md
itaysk: ok.. we can try but I think this will create a discussion, so can you please bring it up in a future team meeting?

itaysk:
ok, I've written it in a less authoritative manner, that cross area changes should be independent (because I think in many cases they should be in same PR but different commits)

PR #483, opened 2021-01-28
feat: Add ringbuffer support to libbpfgo

grantseltzer:
This adds support for ringbuffers to libbpfgo. A good reference for ringbuffers can be found [here](https://nakryiko.com/posts/bpf-ringbuf/). There is not much difference in userspace between ringbuffers and perfbuffers besides the fact that ringbuffers API does not support a callback for lost events. That will have to be handled in kernel space for consumers of libbpfgo, like Tracee. 

I have a separate PR awaiting this which I can open as a draft seeing how it would be consumed in Tracee.

### DIFF
--- a/libbpfgo/libbpf_cb.go
+++ b/libbpfgo/libbpf_cb.go
@@ -20,3 +20,9 @@ func perfLostCallback(ctx unsafe.Pointer, cpu C.int, cnt C.ulonglong) {
 		lostChan <- uint64(cnt)
 	}
 }
+
+//export ringbufferCallback
+func ringbufferCallback(ctx unsafe.Pointer, data unsafe.Pointer, size C.int) C.int {
+	eventChannels[uintptr(ctx)] <- C.GoBytes(data, size)
+	return C.int(0)
+}
--- a/libbpfgo/libbpfgo.go
+++ b/libbpfgo/libbpfgo.go
@@ -20,16 +20,29 @@ package libbpfgo
 extern void perfCallback(void *ctx, int cpu, void *data, __u32 size);
 extern void perfLostCallback(void *ctx, int cpu, __u64 cnt);
 
+extern int ringbufferCallback(void *ctx, void *data, size_t size);
+
 int libbpf_print_fn(enum libbpf_print_level level,
-			   const char *format, va_list args)
+               const char *format, va_list args)
 {
-	if (level != LIBBPF_WARN)
-		return 0;
-	return vfprintf(stderr, format, args);
+    if (level != LIBBPF_WARN)
+        return 0;
+    return vfprintf(stderr, format, args);
 }
 
 void set_print_fn() {
-	libbpf_set_print(libbpf_print_fn);
+    libbpf_set_print(libbpf_print_fn);
+}
+
+struct ring_buffer * init_ring_buf(int map_fd) {
+    struct ring_buffer *rb = NULL;
+    __u64 ctx = map_fd;
+    rb = ring_buffer__new(map_fd, ringbufferCallback, (void*)ctx, NULL);
+    if (!rb) {
+        fprintf(stderr, "Failed to initialize ring buffer\n");
+        return NULL;
+    }
+    return rb;
 }
 
 struct perf_buffer * init_perf_buf(int map_fd, int page_cnt) {
@@ -105,7 +118,7 @@ struct bpf_link* attach_kprobe_legacy(
         return NULL;
     }
 
-	pr = is_kretprobe ? 'r' : 'p';
+    pr = is_kretprobe ? 'r' : 'p';
 
     snprintf(
         fname,
@@ -176,6 +189,7 @@ type Module struct {
 	obj      *C.struct_bpf_object
 	links    []*BPFLink
 	perfBufs []*PerfBuffer
+	ringBufs []*RingBuffer
 }
 
 type BPFMap struct {
@@ -212,6 +226,14 @@ type PerfBuffer struct {
 	pb     *C.struct_perf_buffer
 	bpfMap *BPFMap
 	stop   chan bool
+}
+
+type RingBuffer struct {
+	rb      *C.struct_ring_buffer
+	bpfMap  *BPFMap
+	stop    chan bool
+	stopped bool
+	closed bool 
 }
 
 // BPF is using locked memory for BPF maps and various other things.
@@ -262,10 +284,13 @@ func NewModuleFromBuffer(bpfObjBuff []byte, bpfObjName string) (*Module, error) 
 
 func (m *Module) Close() {
 	for _, pb := range m.perfBufs {
-		C.perf_buffer__free(pb.pb)
+		pb.Close()
+	}
+	for _, rb := range m.ringBufs {
+		rb.Close()
 	}
 	for _, link := range m.links {
-		C.bpf_link__destroy(link.link)
+		C.bpf_link__destroy(link.link) // this call will remove non-legacy kprobes
 		if link.linkType == KprobeLegacy {
 			cs := C.CString(link.eventName)
 			C.remove_kprobe_event(cs, false)
@@ -598,6 +623,80 @@ func doAttachKprobeLegacy(prog *BPFProg, kp string, isKretprobe bool) (*BPFLink,
 var eventChannels = make(map[uintptr]chan []byte)
 var lostChannels = make(map[uintptr]chan uint64)
 
+func (m *Module) InitRingBuf(mapName string, eventsChan chan []byte) (*RingBuffer, error) {
+	bpfMap, err := m.GetMap(mapName)
+	if err != nil {
+		return nil, err
+	}
+
+	ctx := uintptr(bpfMap.fd)
+	if eventsChan == nil {
+		return nil, fmt.Errorf("events channel can not be nil")
+	}
+	eventChannels[ctx] = eventsChan
+
+	rb := C.init_ring_buf(bpfMap.fd)
+	if rb == nil {
+		return nil, fmt.Errorf("")
+	}
+
+	ringBuf := &RingBuffer{
+		rb:     rb,
+		bpfMap: bpfMap,
+		stop:   make(chan bool),
+	}
+	m.ringBufs = append(m.ringBufs, ringBuf)
+	return ringBuf, nil
+}
+
+func (rb *RingBuffer) Start() {
+	go rb.poll()
+}
+
+func (rb *RingBuffer) Stop() {
+	if rb.stopped {
+		return
+	}
+	select {
+	case rb.stop <- true:
+	default:
+	}
+}
+
+func (rb *RingBuffer) Close() {
+	if rb.closed {
+		return
+	}
+	if !rb.stopped {
+		rb.Stop()
+	}
+	C.ring_buffer__free(rb.rb)
+	rb.closed = true
+}
+
+func (rb *RingBuffer) poll() error {
+
+	go func() {
+		<-rb.stop
+		rb.stopped = true
+	}()
+
+	for {
+		err := C.ring_buffer__poll(rb.rb, 0)
+		if err < 0 {
+			if syscall.Errno(-err) == syscall.EINTR {
+				continue
+			}
+			return fmt.Errorf("error polling ring buffer: %d", err)
+		}
+		
+		if rb.stopped {
+			break
+		}
+	}
+	return nil
+}
+
 func (m *Module) InitPerfBuf(mapName string, eventsChan chan []byte, lostChan chan uint64, pageCnt int) (*PerfBuffer, error) {
 	bpfMap, err := m.GetMap(mapName)
 	if err != nil {
@@ -605,7 +704,7 @@ func (m *Module) InitPerfBuf(mapName string, eventsChan chan []byte, lostChan ch
 	}
 	ctx := uintptr(bpfMap.fd)
 	if eventsChan == nil {
-		return nil, fmt.Errorf("failed to init perf buffer: events channel can not be nil!")
+		return nil, fmt.Errorf("failed to init perf buffer: events channel can not be nil")
 	}
 	eventChannels[ctx] = eventsChan
 	lostChannels[ctx] = lostChan
--- a/libbpfgo/selftest/ringbuffers/.gitignore
+++ b/libbpfgo/selftest/ringbuffers/.gitignore
@@ -1,0 +1,3 @@
+vmlinux.h
+self
+self.bpf.o
--- a/libbpfgo/selftest/ringbuffers/Makefile
+++ b/libbpfgo/selftest/ringbuffers/Makefile
@@ -1,0 +1,29 @@
+TARGET := self
+TARGET_BPF := $(TARGET).bpf.o
+
+GO_SRC := $(shell find . -type f -name '*.go')
+LIBBPFGO_SRC := $(shell find ../.. -type f -name '*.go')
+BPF_SRC := $(shell find . -type f -name '*.bpf.c')
+PWD := $(shell pwd)
+
+LIBBPF_HEADERS := /usr/include/bpf
+LIBBPF_OBJ := /usr/lib64/libbpf.a
+
+.PHONY: all
+all: vmlinux.h $(TARGET) $(TARGET_BPF)
+
+vmlinux.h:
+	bpftool btf dump file /sys/kernel/btf/vmlinux format c > vmlinux.h
+
+go_env := CC=gcc CGO_CFLAGS="-I $(LIBBPF_HEADERS)" CGO_LDFLAGS="$(LIBBPF_OBJ)"
+$(TARGET): $(GO_SRC) $(LIBBPFGO_SRC)
+	$(go_env) go build -ldflags '-extldflags "-static"' -o $(TARGET) 
+
+$(TARGET_BPF): $(BPF_SRC)
+	clang \
+		-g -O2 -c -target bpf \
+		-o $@ $<
+
+.PHONY: clean
+clean: 
+	rm $(TARGET) $(TARGET_BPF) vmlinux.h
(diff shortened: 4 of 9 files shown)


libbpfgo/libbpfgo.go
```diff
@@ -32,6 +34,15 @@ void set_print_fn() {
 	libbpf_set_print(libbpf_print_fn);
 }
 
+struct ring_buffer * init_ring_buf(int map_fd, int page_cnt) {
+	struct ring_buffer *rb = NULL;
+	rb = ring_buffer__new(map_fd, ringbufferCallback, NULL, NULL); //
+	if (!rb) {
+		fprintf(stderr, "Failed to initialize ring buffer\n");
+        return NULL;
```
grantseltzer, line 42: fixing whitespace right now

simar7: This brings up a good point though - how should we format the C code? We talked about Go code formatting but I'm curious to hear what you all are using.

libbpfgo/libbpfgo.go
```diff
@@ -261,11 +280,15 @@ func NewModuleFromBuffer(bpfObjBuff []byte, bpfObjName string) (*Module, error)
 }
 
 func (m *Module) Close() {
+
```
yanivagman, line 283: extra newline

libbpfgo/libbpfgo.go
```diff
@@ -20,16 +20,28 @@ package libbpfgo
...
 }
 
 void set_print_fn() {
-	libbpf_set_print(libbpf_print_fn);
+    libbpf_set_print(libbpf_print_fn);
+}
+
+struct ring_buffer * init_ring_buf(int map_fd) {
+    struct ring_buffer *rb = NULL;
+    rb = ring_buffer__new(map_fd, ringbufferCallback, NULL, NULL);
```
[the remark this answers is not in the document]
grantseltzer: Just to be consistent with the perf buffer init function, and to avoid passing `C.ringBufferCallback`. I can get rid of it if you feel strongly about it.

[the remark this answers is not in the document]
yanivagman: I think it is better to avoid unnecessary wrappers and use go where possible to make code more readable.
Not a must though

libbpfgo/libbpf_cb.go
```diff
@@ -20,3 +20,9 @@ func perfLostCallback(ctx unsafe.Pointer, cpu C.int, cnt C.ulonglong) {
 		lostChan <- uint64(cnt)
 	}
 }
+
+//export ringbufferCallback
+func ringbufferCallback(ctx unsafe.Pointer, data unsafe.Pointer, size C.ulong) C.int {
```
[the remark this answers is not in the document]
grantseltzer: Fixed back to int

grantseltzer:
Marking this as ready for review, following up with another PR for selftests that include a runner script for it that I would subsequently add to github actions workflow.

grantseltzer:
I actually just found a deadlock in this and the perfbuffer code when `Stop()` is called and the thread is blocked waiting for the event polling. Working on fix.

grantseltzer:
Ok fixed

yanivagman:
About you perfbuffer deadlock comment -
How can the perf poll loop be blocked if we have a timeout there?

libbpfgo/libbpfgo.go
```diff
@@ -598,14 +621,79 @@ func doAttachKprobeLegacy(prog *BPFProg, kp string, isKretprobe bool) (*BPFLink,
...
+			err := C.ring_buffer__poll(rb.rb, -1)
+			pollChannel  <- err
+		}
+	}()
+
+	for {
+		select {
+		case <-rb.stop:
+			return nil
+		case err := <-pollChannel:
```
yanivagman, line 677: Why not using timeout instead? (like we do for the perf buffer)
`err := C.ring_buffer__poll(rb.rb, 300)`

grantseltzer: I'm trying to figure out what is the exact unexpected behavior that's causing `ring_buffer__poll` to block regardless of the specified timeout. The test program I wrote exposed a block where the write to `rb.stop` in ringbuffer's `Stop` method to block indefinitely. Perhaps a channel was closed somewhere that the poll function was caught trying to write to. Perhaps a reused map file descriptor that's not being properly cleaned up. 

Pushing up that test program now so you can take a look.

grantseltzer:
Alright so I just pushed a commit that I'd like you to look at @yanivagman 

The issue with having `C.ring_buffer__poll(rb.rb, 300)` (with the 300ms timeout) is that the ringbuffer can be free'd while waiting for an event to come in which causes a panic. The issue with my first approach of having the poll run in its own go routine is the same one, poll after free. 

I've removed the timeout from `ring_buffer__poll` (passing 0 for timeout) which seems to effectively solve the issue of getting caught with a free while polling.

yanivagman:
Ok. I think I understand what you say.
So the perfbuffer can be freed by the Close() function while it is still polling - because the 300ms timeout didn't expire.
The problem here is that the pb.Stop() function didn't have the chance to gracefully stop the polling, right?
Changing the timeout to 0 will cause a busy loop, which will consume CPU resources, so it is not the best way to resolve that.
Instead, what we can do, is to introduce a new boolean variable to the PerfBuffer struct called "stopped".
We can then set this variable to true in the` case <-pb.stop:`. Then, we can wait with the Close() function (which frees the perf buffer) until we see that the stopped variable was set to true.
WDYT?

grantseltzer:
@yanivagman I thought of this, but the problem there is there's still a chance for poll after free when the free happens after the check but before the poll call. Only way to safely do it that way is to use a lock which we want to avoid.

yanivagman:
@grantseltzer I don't understand why there is still a chance for a poll after free if we implement this suggestion.
If you set the `stopped` variable in the` case <-pb.stop:`, no poll can happen after this point as you return from the poll loop right after setting the `stopped` variable

grantseltzer:
Ah that's clever, I misunderstood your previous comment. I'll implement exactly this.

grantseltzer:
@yanivagman found a bit of an issue with this approach as well. The reason `perf_buffer__poll` can hang is because channel sends block if there aren't any receivers. In the perf/ring buffer callbacks the the send will sit forever if the consumer of the events channel has stopped receiving on it. 

For example:

In the selftest:
```
recvLoop:
	for {
		 _ = <-eventsChannel     // events sent to eventsChannel in the callback
        	numberOfEventsReceived++ 
	        if numberOfEventsReceived > 3 {
	 	    break recvLoop	
	        }
	}
```

In this case when events are received after the loop is broken the callback will hang indefinitely (the timeout in `perf_buffer__poll` will not interrupt it). Therefore the stop signal is never received in your select statement. 

The fix here is to use a select statement in the callback which will make the write non-blocking. The other thing i've added is a check to see if `Close()` has been called multiple times to avoid a segfault.

libbpfgo/selftest/ringbuffers/.gitignore
```diff
@@ -0,0 +1,2 @@
+self
+self.bpf.o
```
itaysk, line 2: depending on the response to my other comments about libbpf.a and vmlinux.h , maybe need to add them here as well

libbpfgo/libbpfgo.go
```diff
@@ -598,14 +623,80 @@ func doAttachKprobeLegacy(prog *BPFProg, kp string, isKretprobe bool) (*BPFLink,
...
+
+func (rb *RingBuffer) Start() {
+	go rb.poll()
+}
+
+func (rb *RingBuffer) Stop() {
+	rb.stop <- true
+}
+
+func (rb *RingBuffer) Close() {
```
itaysk, line 660: good catch, thinking about that what happens if Close is called before Stop. In this case you chose to wait until the buffer is closed, but I think an easier and reasonable approach, is to just call Stop? either way I don't think close should block the caller

itaysk: also, whatever we decide here probably applies to perf buffer (but can be implemented in a separate PR) cc: @yanivagman

libbpfgo/libbpfgo.go
itaysk: didn't really investigate this but just noting that it's impossible to write to a closed channel, more likely it's a double write to an unbuffered channel

libbpfgo/libbpfgo.go
itaysk: ok after posting the review I see that this was already discussed so I'll take a moment to re-read the history first

libbpfgo/selftest/ringbuffers/go.mod
```diff
@@ -0,0 +1,7 @@
+module github.com/aquasecurity/tracee/libbpfgo/selftests/5.8.15
+
+go 1.15
+
+replace github.com/aquasecurity/tracee/libbpfgo => /home/vagrant/go/src/github.com/aquasecurity/tracee/libbpfgo
```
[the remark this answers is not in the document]
grantseltzer: Yes :-)

libbpfgo/selftest/ringbuffers/Makefile
```diff
@@ -0,0 +1,31 @@
...
+BPF_SRC := $(shell find . -type f -name '*.bpf.c')
+PWD := $(shell pwd)
+
+LIBBPF_HEADERS := /usr/include/bpf
+LIBBPF_OBJ := $(PWD)/libbpf.a
+
+.PHONY: all
+all: $(TARGET) $(TARGET_BPF)
+
+.PHONY: vmlinux.h
```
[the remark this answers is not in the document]
grantseltzer: TARGET/TARGET_BPF should be rebuilt for each selftest invocation.

grantseltzer:
@itaysk You're right that the select statement causes dropped events. Back to the drawing board.

libbpfgo/selftest/ringbuffers/run.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
grantseltzer: Yes, I need to learn about bats, will this week

libbpfgo/selftest/ringbuffers/Makefile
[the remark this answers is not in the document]
itaysk: why is that? is there a reason to rebuild something if it's source hasn't changed?

libbpfgo/selftest/ringbuffers/run.sh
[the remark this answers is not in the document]
itaysk: it can be in a future PR

grantseltzer:
Alright! I am cautiously optimistic that i've gotten it right this time. If we're calling Stop/Close on the ringbuffer there's no reason to block waiting for a consumer of events. I've replaced the stop signal send in `poll` with a  goroutine as to avoid blocking. This should avoid dropped events (at least in userspace) and blocking channel read/writes.

Take a look and tell me what you think. I'm going to be really happy once I put together benchmarks so we can put some metrics behind my thinking that this avoids blocks.

tracee-ebpf/tracee/tracee.go
```diff
@@ -845,6 +845,8 @@ func (t *Tracee) initBPF(bpfObjectPath string) error {
 		}
 	}
 
+	//TODO: if kernel version supports ringbuffers, use them instead of perf
```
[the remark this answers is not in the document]
grantseltzer: Done :)

libbpfgo/selftest/ringbuffers/Makefile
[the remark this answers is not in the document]
grantseltzer: I was mistaken, I thought it wasn't picking up changes in the bpf code, but that's what the TARGET_BPF is depending on. Removing the PHONY tags.

libbpfgo/selftest/ringbuffers/self.bpf.c
```diff
@@ -0,0 +1,42 @@
...
+	struct process_info *process;
+	
+    // Reserve space on the ringbuffer for the sample
+	process = bpf_ringbuf_reserve(&events, sizeof(struct process_info), ringbuffer_flags);
+	if (!process) {
+		return 0;
+    }
+
+	process->pid = tgid;
+    bpf_get_current_comm(&process->comm, 100);
```
yanivagman, line 38: Indentation of this function (better use spaces as a convention like we do in tracee.bpf.c file)

yanivagman (approved):
LGTM!

libbpfgo/libbpfgo.go
```diff
@@ -598,14 +623,84 @@ func doAttachKprobeLegacy(prog *BPFProg, kp string, isKretprobe bool) (*BPFLink,
...
+	for {
+		err := C.ring_buffer__poll(rb.rb, 300)
+		if err < 0 {
+			if syscall.Errno(-err) == syscall.EINTR {
+				continue
+			}
+			return fmt.Errorf("error polling ring buffer: %d", err)
+		}
+		
+		if rb.stopped {
```
itaysk, line 689: just a readability suggestion (totally optional) to move this stop condition to the for line
```
for !rb.stopped {
```

libbpfgo/libbpfgo.go
```diff
@@ -598,14 +623,84 @@ func doAttachKprobeLegacy(prog *BPFProg, kp string, isKretprobe bool) (*BPFLink,
...
+	C.ring_buffer__free(rb.rb)
+	rb.closed = true
+}
+
+func (rb *RingBuffer) poll() error {
+
+	go func() {
+		<-rb.stop
+		rb.stopped = true
+	}()
```
itaysk, line 678: just a readability suggestion (totally optional) to move this out of poll (into Start)

libbpfgo/libbpfgo.go
```diff
@@ -598,14 +623,84 @@ func doAttachKprobeLegacy(prog *BPFProg, kp string, isKretprobe bool) (*BPFLink,
...
+
+func (rb *RingBuffer) Close() {
+	if rb.closed {
+		return
+	}
+	for {
+		if rb.stopped {
+			break
+		}
+	}
```
itaysk, line 668: not sure I understand: if Close was called before Stop, you wait for Stop to have happened? isn't it better to call Stop instead?

libbpfgo/libbpfgo.go
itaysk: ☝️

libbpfgo/libbpfgo.go
grantseltzer: Yes good point, need to add a check in Stop() to see if it's already stopped because of the nature of channels but this makes more sense

grantseltzer:
@itaysk I realized why I wanted the make targets to be marked `.PHONY`, if there's changes to libbpfgo the makefile wasn't picking that up and therefore not recompiling.

libbpfgo/libbpfgo.go
grantseltzer: This throws off the fact that there's only one consumer of `rb.stop` so can cause a block. Now that I think of it, do we even need a `Stop` and a `Close`? Who is going to stop a ringbuffer and then restart it later? I'd be fine with just removing `Stop` and having the write to `rb.stop` in `Close()`. 

cc @yanivagman

itaysk: I'll let you and Yaniv decide if you want to support start after stop (make sure to document it) but certainly we don't want to support start after close, so I think either way close should just call stop, no?

itaysk:
> @itaysk I realized why I wanted the make targets to be marked .PHONY, if there's changes to libbpfgo the makefile wasn't picking that up and therefore not recompiling.

You already have `$(TARGET): $(GO_SRC)` which should take care of this dependency, I don't see a reason to why it needs to be `.PHONY`, but I won't let it block the PR (we can fix the makefile later). If you want we could look into the makefile together lmk

grantseltzer:
@itaysk maybe I misunderstand how .PHONY works, if I make a change in libbpfgo (two directories back), how `$(TARGET): $(GO_SRC)` affect rebuilding the Go binary?

itaysk:
right (I thought GO_SRC included that), so I think the correct solution is to make TARGET depend on libbpfgo. I mean, regardless of the rebuild issue, it is correct to say that this test needs libbpfgo right? if we express the relationship correctly, make should be able to just do the right thing without workarounds like .PHONY. 
PHONY tells make that this target is not a real file, which is not a case here IMU

grantseltzer:
@itaysk Ah, I understand. I've added a LIBBPFGO_SRC dependency to the BUILD target (not sure if my terminology is correct). I think this is what you were getting at.

libbpfgo/libbpfgo.go
itaysk: I still see whitespace issues in the commented C

libbpfgo/libbpfgo.go
```diff
@@ -598,14 +623,88 @@ func doAttachKprobeLegacy(prog *BPFProg, kp string, isKretprobe bool) (*BPFLink,
...
+
+func (rb *RingBuffer) Start() {
+	go rb.poll()
+}
+
+func (rb *RingBuffer) Stop() {
+	if rb.stopped {
+		return
+	}
+	select {
```
itaysk, line 660: looks like the select is unnecessary?

libbpfgo/selftest/ringbuffers/Makefile
```diff
@@ -0,0 +1,29 @@
...
+	$(go_env) go build -ldflags '-extldflags "-static"' -o $(TARGET) 
+
+$(TARGET_BPF): $(BPF_SRC)
+	clang \
+		-g -O2 -c -target bpf \
+		-o $@ $<
+
+.PHONY: clean
+clean:
+	rm self self.bpf.o vmlinux.h
```
itaysk, line 29: nit: probably should use TARGET and TRAGET_BPF here

itaysk (approved):
LGTM :)
there are some small comments here and in the previous review if you want to address/ignore before merging.
(and please squash before merging)

libbpfgo/libbpfgo.go
grantseltzer: This is what makes the channel write non-blocking.

libbpfgo/libbpfgo.go
grantseltzer: So strange, not sure why that kept happening...

Issue #486, opened 2021-02-01
Testing Rego signatures


itaysk:
We have an elaborate testing suit for Go signatures. Need to have an equivalent for Rego signatures, probably using Rego as well.

PR #489, opened 2021-02-01
feat: add rego tests

itaysk:
close: #486


itaysk:
cc: @roikol

grantseltzer:
So is the idea here to have parity between the go signatures and rego ones? If so how is this automatically run?

itaysk:
currently it's not automatically run, one has to run the test command. there's a different issue for integrating rego tests with the makefile: https://github.com/aquasecurity/tracee/issues/491

grantseltzer (approved):

simar7 (approved):

Issue #465, opened 2021-01-27
fix release process


itaysk:
now that we have the monorepo structure in place, we need to rethink the release process. currently we have 1 artifact, `tracee-tar.gz` which is essentially tracee-ebpf. we can still use one artifact archive, but that will have to include tracee-rules as well.
as part of this pr we will need to extract the release script from the tracee-rules makefile into a makefile at the root.
this makefile will invoke the tracee-ebpf and tracee-ruels builds and bundle them into a release

PR #492, opened 2021-02-01
update release process after restructure

itaysk:
close: #465 
this doesn't address repackaging the container which will be done in a followup PR. this just "fixes" the release process to reflect the repo restructure

- rename tracee to tracee-ebpf
- adds a root level makefile 
- extracts release from the original tracee makefile (now tracee-ebpf) into root level makefile
- fix a few small issues with the makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,80 @@
+.PHONY: all
+all:
+	@echo "This make file is intended for releasing only. In order to build any of Tracee's components, use their respective make files."
+
+# RELEASE_TAG must be set to specify a name for the release
+OUT_DIR ?= dist
+OUT_DOCKER ?= tracee
+OUT_ARCHIVE := $(OUT_DIR)/tracee.tar.gz
+OUT_CHECKSUMS := $(OUT_DIR)/checksums.txt
+PUSH_DOCKER_REPO ?= aquasec/tracee
+PUSH_DOCKER_TAG ?= $(RELEASE_TAG:v%=%)
+
+CMD_DOCKER ?= docker
+CMD_GIT ?= git
+CMD_CHECKSUM ?= sha256sum
+CMD_GITHUB ?= gh
+CMD_TAR ?= tar
+CMD_CP ?= cp
+
+release_tools := $(CMD_DOCKER) $(CMD_GIT) $(CMD_CHECKSUM) $(CMD_GITHUB) $(CMD_TAR) $(CMP_CP)
+check_release_tools := $(shell for tool in $(release_tools); do command -v $$tool >/dev/null || (echo "missing required tool $$tool" ; false); done)
+
+$(OUT_DIR):
+	mkdir -p $@
+
+# make_artifact invokes another make from $1 to make the artifact $2 using additional flags $3
+define make_artifact
+	$(MAKE) -C $1 dist/$2 $3 OUT_DIR=dist && $(CMD_CP) -r $1/dist/$2 $(OUT_DIR)/$2
+endef
+
+$(OUT_DIR)/tracee: tracee-ebpf | $(OUT_DIR)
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -0,0 +1,80 @@
+.PHONY: all
+all:
+	@echo 'This make file is intended for releasing only. In order to build any of Tracee's components, use their respective make files.'
```
grantseltzer, line 3: You have a `'` in this line which breaks the makefile. Should change the outer quotes to double quotes

tracee-ebpf/Makefile
```diff
@@ -147,11 +139,15 @@ define docker_builder_make
 	--entrypoint make $(DOCKER_BUILDER) KERN_BLD_PATH=$(DOCKER_BUILDER_KERN_BLD) KERN_SRC_PATH=$(DOCKER_BUILDER_KERN_SRC) $(1)
 endef
 
+.PHONY: mostlyclean
+mostlyclean:
+	-rm -rf $(OUT_BIN) $(bpf_bundle_dir) $(OUT_BPF) $(BPF_BUNDLE)
```
grantseltzer, line 144: Why have two cleans?

grantseltzer (changes requested):

tracee-ebpf/Makefile
itaysk: clean should delete all resources that were created by the makefile but some resources are 1) not changed between builds and 2) take long time to rebuild so it's better to keep them around. For example: libbpf, or the lower docker layers are useful to keep around. I followed the convention described here https://www.gnu.org/software/make/manual/html_node/Goals.html

grantseltzer (approved):

tracee-ebpf/Readme.md
```diff
@@ -34,21 +34,21 @@ This will run Tracee-eBPF with no arguments, which defaults to collecting a usef
...
+2. Use the docker image from Docker Hub: `aquasec/tracee`.
 3. Build the executable from source using `make build`. For that you will need additional development tooling.
 4. Build the executable from source in a Docker container which includes all development tooling, using `make build DOCKER=1`.
 
 ### Setup options
 
-Tracee-eBPF is made of a userspace executable (`tracee`) that drives the eBPF program, and the eBPF program itself (`tracee.bpf.$kernelversion.$traceeversion.o`). When the `tracee` is started, it will look for the eBPF program in specific places and if not found, it will attempt to build the eBPF program automatically before it starts (you can control this using the `--build-policy` flag).
+Tracee-eBPF is made of a userspace executable (`tracee-ebpf`) that drives the eBPF program, and the eBPF program itself (`tracee.bpf.$kernelversion.$traceeversion.o`). When Tracee is started, it will look for the eBPF program in specific places and if not found, it will attempt to build the eBPF program automatically before it starts (you can control this using the `--build-policy` flag).
```
simar7, line 43: Open ended thought: I wonder if we should call Tracee (the executable) as Tracee-eBPF everywhere (except when its the rule engine) to avoid confusion.

simar7 (approved):
LGTM

tracee-ebpf/Readme.md
itaysk: Good question that I also pondered about it while writing this PR. Initially I changed every single instance of tracee to tracee-ebpf, but it started to read heavily. so I changed my approach and used tracee-ebpf only when necessary and kept it tracee when possible else. My logic was that if the word relates specifically to the executable, or a property of it, then I'll change it to tracee-ebpf, but if the sentence was relevant to Tracee in general (i.e runtime security), then I keep it Tracee.
The sentence in context here is a good example: I changed the first occurrence to tracee-ebpf because it explicitly refers to the binary, but kept the second occurrence since it applies to tracee in general (regardless of how you run it).

Issue #496, opened 2021-02-02
tracee rules should use TraceeEvent struct from Tracee-eBPF


itaysk:
currently tracee-rules redefines Tracee's Event type. This is because at the time that code was written it was complicated to import package tracee. but since then #388 was merged so now it should be possible.

Issue #494, opened 2021-02-01
Example in tracee-rules/README.md is invalid


grantseltzer:
The quickstart command in the tracee-rules readme fails: `sudo tracee-ebpf -o format:gob -o option:eot | tracee-rules --tracee-file /dev/stdin`

 In particular the following error is emitted `error handling event by signature Code injection: argument flags not found`

I believe this is because the default rules for tracee-ebpf do not emit events with the argument 'flags' which is looked for in the 'Code injection' rule which is used by default in tracee-rules.

I believe to resolve this issue a working quickstart example should be added to the README and a test should be added to verify that it's always working.

itaysk:
the example is working for me (I checked before I write it). which commit are you on?

itaysk:
thanks @grantseltzer . I did check that it was working before I added this to the readme. I'll investigate this today

itaysk:
I accidentally opened #496 today and it seems to be related to this issue. I'll push a fix today

grantseltzer:
Closed by #498

Issue #497, opened 2021-02-02
rename package import paths before releasing


itaysk:
#457 changed the go import path for tracee. currently importers are unaffected because go usually takes the latest release which is prior to this change. before releasing the next version we should make sure known consumers (tracee-rules, DTA) are being updated as well.

simar7:
Rename go modules for tracee-ebpf and elsewhere wherever applicable before we finish this issue so they can be imported.

PR #498, opened 2021-02-02
refactor: tracee-rules to use Event type from tracee-ebpf

itaysk:
close: #496, #494

### DIFF
--- a/tracee-rules/engine/engine.go
+++ b/tracee-rules/engine/engine.go
@@ -5,6 +5,7 @@ import (
 	"log"
 
 	"github.com/aquasecurity/tracee/tracee-rules/types"
+	tracee "github.com/aquasecurity/tracee/tracee/external"
 )
 
 // Engine is a rule-engine that can process events coming from a set of input sources against a set of loaded signatures, and report the signatures' findings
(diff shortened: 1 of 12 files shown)


tracee-rules/engine/engine.go
```diff
@@ -5,6 +5,7 @@ import (
 	"log"
 
 	"github.com/aquasecurity/tracee/tracee-rules/types"
+	tracee "github.com/aquasecurity/tracee/tracee/external"
```
grantseltzer, line 8: I'm confused by this import path. To me it looks like the `tracee/external` package is in `github.com/aquasecurity/tracee/tracee-ebpf/tracee/external`, how does this work when the `tracee-ebpf` part is dropped?

grantseltzer:
Looks like you have failed tests, on my end #494 is resolved by this though. Otherwise code looks good to me.

tracee-rules/engine/engine.go
itaysk: you are looking at HEAD but go import gets the packages from latest tag. in tandem with the release we would need to updated these paths, and I've opened #497 as a reminder

itaysk:
alright I've rebased on main and tests are now passing. we still have the caveat I mentioned in https://github.com/aquasecurity/tracee/pull/498#discussion_r569363976. if you have another idea to approach this, I'm open to suggestions

grantseltzer:
Ah I follow, this LGTM then!

grantseltzer (approved):

Issue #499, opened 2021-02-02
Update tracee-rules UX to align with tracee-ebpf


grantseltzer:
I think we should align with tracee-ebpf's UX. 
perhaps something like:

```
--input-tracee file:/path/to/file
--input-tracee option:{eot}
--input-tracee format:gob/json 
```

stdin can be given as a file (like in the readme) or we can create an alias for it

_Originally posted by @itaysk in https://github.com/aquasecurity/tracee/discussions/495#discussioncomment-330123_

grantseltzer:
@itaysk What did you have in mind for when a user specifies `--input-tracee option:eot`?

itaysk:
take a loot at the same option in tracee-ebpf, if it's still not clear let's have a chat and I'll explain

yanivagman:
I wrote this comment in the PR, but maybe its better that we also discuss this here:
To be future proof, I think we should change this flag name to be simply `input`.
Then, we can have something like:
`--input source:key=value`
WDYT?

itaysk:
closed by #500

PR #500, opened 2021-02-02
Refactor flags in tracee-rules

grantseltzer:
This addresses some of the points in #499. In particular:

1) Removes the `--trace-file`, and `--stdin-as` flags.
1) Adds a new flag called `--input-trace` that allows the user to specify the input file and input format for events to tracee-rules.

Still need to address:
1) ~Need to add unit tests for the above parsing logic~
1) What should the default options be?
1) ~https://github.com/aquasecurity/tracee/issues/499#issuecomment-771927000~

### DIFF
--- a/tracee-rules/input.go
+++ b/tracee-rules/input.go
@@ -170,7 +170,7 @@ Examples:
 
 'tracee-rules --input-tracee file:./events.json --input-tracee format:json'
 'tracee-rules --input-tracee file:./events.gob --input-tracee format:gob'
-'sudo tracee-ebpf -o format:gob -o option:eot | tracee-rules --input-tracee file:stdin --input-tracee format:gob'
+'sudo tracee-ebpf -o format:gob | tracee-rules --input-tracee file:stdin --input-tracee format:gob'
 `
 
 	fmt.Println(traceeInputHelp)
--- a/tracee-rules/input.go
+++ b/tracee-rules/input.go
@@ -16,7 +16,7 @@ import (
 	tracee "github.com/aquasecurity/tracee/tracee/external"
 )
 
-var helpErr = errors.New("help me if you can im feeling down")
+var helpErr = errors.New("user has requested help text")
 
 type inputFormat uint8
 
@@ -29,7 +29,6 @@ const (
 type traceeInputOptions struct {
 	inputFile   *os.File
 	inputFormat inputFormat
-	closeOnEOF  bool
 }
 
 func setupInputSource(opts *traceeInputOptions) (chan types.Event, error) {
@@ -121,11 +120,6 @@ func parseTraceeInputOptions(inputOptions []string) (*traceeInputOptions, error)
 			if err != nil {
 				return nil, err
 			}
-		} else if kv[0] == "option" {
-			err = parseTraceeInputOption(&inputSourceOptions, kv[1])
-			if err != nil {
-				return nil, err
-			}
 		} else if kv[0] == "format" {
 			err = parseTraceeInputFormat(&inputSourceOptions, kv[1])
 			if err != nil {
@@ -153,15 +147,6 @@ func parseTraceeInputFile(option *traceeInputOptions, fileOpt string) error {
 		return fmt.Errorf("invalid file: %s", fileOpt)
 	}
 	option.inputFile = f
-	return nil
-}
-
-func parseTraceeInputOption(option *traceeInputOptions, optionString string) error {
-	if strings.ToUpper(optionString) == "EOF" {
-		option.closeOnEOF = true
-	} else {
-		return fmt.Errorf("unkown tracee input option: %s", optionString)
-	}
 	return nil
 }
 
--- a/tracee-rules/input.go
+++ b/tracee-rules/input.go
@@ -4,47 +4,47 @@ import (
 	"bufio"
 	"encoding/gob"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"log"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/aquasecurity/tracee/tracee-rules/types"
 	tracee "github.com/aquasecurity/tracee/tracee/external"
 )
 
-func setupStdinSource(inputSource string) (chan types.Event, error) {
-	res := make(chan types.Event)
-	scanner := bufio.NewScanner(os.Stdin)
-	go func() {
-		for scanner.Scan() {
-			event := scanner.Bytes()
-			switch inputSource {
-			case "tracee":
-				var e tracee.Event
-				err := json.Unmarshal(event, &e)
-				if err != nil {
-					log.Printf("invalid json in %s: %v", string(event), err)
-				}
-				res <- types.Event(e)
-			}
-		}
-		close(res)
-	}()
-	return res, nil
+type inputFormat uint8
+
+const (
+	invalidInputFormat inputFormat = iota + 1
+	jsonInputFormat
+	gobInputFormat
+)
+
+type traceeInputOptions struct {
+	inputFile   *os.File
+	inputFormat inputFormat
+	closeOnEOF  bool
 }
 
-func setupTraceeSource(traceeFilePath string) (chan types.Event, error) {
-	_, err := os.Stat(traceeFilePath)
-	if err != nil {
-		return nil, fmt.Errorf("invalid Tracee input file: %s", traceeFilePath)
+func setupInputSource(opts *traceeInputOptions) (chan types.Event, error) {
+
+	if opts.inputFormat == jsonInputFormat {
+		return setupJSONInputSource(opts)
 	}
-	f, err := os.Open(traceeFilePath)
-	if err != nil {
-		return nil, fmt.Errorf("invalid file: %s", traceeFilePath)
+
+	if opts.inputFormat == gobInputFormat {
+		return setupGobInputSource(opts)
 	}
-	dec := gob.NewDecoder(f)
+
+	return nil, errors.New("could not set up input source")
+}
+
+func setupGobInputSource(opts *traceeInputOptions) (chan types.Event, error) {
+	dec := gob.NewDecoder(opts.inputFile)
 	res := make(chan types.Event)
 	go func() {
 		for {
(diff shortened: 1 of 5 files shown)


grantseltzer:
Also need to address: https://github.com/aquasecurity/tracee/issues/499#issuecomment-771927000

grantseltzer:
Thoughts on default options being reading `gob` from `stdin`? 

Also for the `eof` option, should this mean cut the stream/execution once an EOF is reached in the reading file?

tracee-rules/main.go
```diff
@@ -16,6 +17,9 @@ func main() {
 		Name:  "tracee-rules",
 		Usage: "A rule engine for Runtime Security",
 		Action: func(c *cli.Context) error {
+			if c.NArg() == 0 && c.NumFlags() == 0 {
+				return errors.New("tracee-rules requires arguments, please use `tracee-rules --help`")
```
itaysk, line 21: isn't it better to just print the help? I think you or Simar suggested that

tracee-rules/main.go
```diff
@@ -16,6 +17,9 @@ func main() {
 		Name:  "tracee-rules",
 		Usage: "A rule engine for Runtime Security",
 		Action: func(c *cli.Context) error {
+			if c.NArg() == 0 && c.NumFlags() == 0 {
```
itaysk, line 20: for my understanding, what are both needed?

tracee-rules/output.go
```diff
@@ -16,6 +16,7 @@ func setupOuput(webhook string) (chan types.Finding, error) {
 	go func() {
 		for res := range out {
 			fmt.Printf("%+v\n", res)
+			//FIXME: Don't do this check everytime there's a finding
```
itaysk, line 19: ok but can you please record this in an issue instead of a comment?

tracee-rules/input.go
```diff
@@ -4,46 +4,46 @@ import (
...
-				res <- types.Event(e)
-			}
-		}
-		close(res)
-	}()
-	return res, nil
+type inputFormat uint8
+
+const (
+	invalidInputFormat inputFormat = iota + 1
```
itaysk, line 21: curios why `+ 1`?

itaysk (commented):
added some small comments, but the big one is to properly handle EOT flag. I've commented on the issue, let's first make sure you understand the requirement first (this is crucial to how tracee-ebpf->rules work)

tracee-rules/main.go
grantseltzer: They're actually not, this stemmed from me confusing this CLI library with the other popular Go one (Cobra) which does positional argument validation. I'm removing the `NArg()` check. Tracee-rules doesn't take any positional args anyway.

tracee-rules/input.go
grantseltzer: Snuck in from a change I made. Removing.

grantseltzer:
@itaysk I addressed your feedback, can you also verify my usage of `tracee.Event` vs. `types.Event`. I had to rebase and solve the conflict so just want to make sure I got the idea right of where to use each.

tracee-rules/input.go
```diff
@@ -4,47 +4,49 @@ import (
...
-				if err != nil {
-					log.Printf("invalid json in %s: %v", string(event), err)
-				}
-				res <- types.Event(e)
-			}
-		}
-		close(res)
-	}()
-	return res, nil
+var helpErr = errors.New("help me if you can im feeling down")
```
yanivagman, line 19: :-)

tracee-rules/input.go
```diff
@@ -4,47 +4,49 @@ import (
...
+	inputFile   *os.File
+	inputFormat inputFormat
+	closeOnEOF  bool
 }
 
-func setupTraceeSource(traceeFilePath string) (chan types.Event, error) {
-	_, err := os.Stat(traceeFilePath)
-	if err != nil {
-		return nil, fmt.Errorf("invalid Tracee input file: %s", traceeFilePath)
+func setupInputSource(opts *traceeInputOptions) (chan types.Event, error) {
```
yanivagman, line 35: Is there something specific to tracee in the opts struct?
If not, better changing the name to `inputOptions`, especially if you change `setupTraceeSource`->`setupInputSource`

tracee-rules/input.go
```diff
@@ -66,8 +68,132 @@ func setupTraceeSource(traceeFilePath string) (chan types.Event, error) {
...
+			}
+			res <- tracee.Event(e)
+		}
+		opts.inputFile.Close()
 		close(res)
 	}()
 	return res, nil
 }
+
+func parseTraceeInputOptions(inputOptions []string) (*traceeInputOptions, error) {
```
yanivagman, line 96: Same question as above - is there something specific to tracee here?
If yes, is it possible to make this function generic?

tracee-rules/input.go
```diff
@@ -66,8 +68,132 @@ func setupTraceeSource(traceeFilePath string) (chan types.Event, error) {
...
+	} else {
+		option.inputFormat = invalidInputFormat
+		return fmt.Errorf("invalid tracee input format specified: %s", formatString)
+	}
+	return nil
+}
+
+func printHelp() {
+	traceeInputHelp := `
+tracee-rules --input-tracee <key:value>,<key:value> --input-tracee <key:value>
```
yanivagman, line 184: To be future proof, I think we should change this flag name to be simply `input`.
Then, we can have something like:
`--input source:key=value`
WDYT?

itaysk: It's unlikely that future input sources will share config options. consider kubernetes audit - it needs a URL, credentials, maybe a kubeconfig file. none of this is relevant to tracee, so I think it's better to keep separate config structs per input source and separate flags

tracee-rules/input.go
itaysk: a better way to implement this would be to return a typed error, and check for the error type, instead of it's string representation.

grantseltzer: I did this based on how the Go standard library does it (i.e. io.EOF in https://golang.org/pkg/io/#pkg-variables). 

Do you think it's necessary to create a new type? We'd then have to have a type switch on return and it'd be effectively doing the same thing.

tracee-rules/input.go
grantseltzer: This is a fair point, there's of course value in keeping code more 'general', but my philosophy is that we shouldn't complicate code for the sake of flexibility that we don't need yet. This isn't such a large chunk of code, when we support more input sources we can easily go back and refactor a bit.

tracee-rules/input.go
grantseltzer: _(see comment above)_

tracee-rules/input.go
grantseltzer: _(see my comment above)_

tracee-rules/input.go
itaysk: I don't think that's a comparable example, but it's not that important, this is fine as well

tracee-rules/input.go
yanivagman: Ok, got it.
In that case, I think the name should be changed to `setupTraceeInputSource` as I mentioned above

itaysk: if you do that then same for setup{GOB,JSON}InputSource (add tracee since they accept traceeOptions)

grantseltzer:
I believe that based on the discussion in #508 this is ready for further review and the final question I have is what the default behavior should be for tracee-rules. I think it should be reading gob format from standard input.

itaysk:
@grantseltzer the code in this PR doesn't implement the revised logic suggested in #508 , did you mean to say that it is?

regarding the defaults, if you want to include #508 in this PR, then:

```
--input-tracee file:{/path/to/file,stdin} (default: stdin)
--input-tracee format:{gob/json}
```

grantseltzer:
@itaysk I figured I would make that a separate PR but certainly can into this one.

grantseltzer:
I'm actually going to open it as a second PR, the two things are related but it's a different part of the repo.

itaysk:
> I figured I would make that a separate PR but certainly can into this one.

no that's fine, I just understood it this way because you mentioned the discussion.

so now, for gob you read until EOT but for JSON you don't. I'm not sure this is what we wanted, but we can let it slide because that's changing soon anyway

tracee-rules/input.go
itaysk: @grantseltzer this (and the others) still need to be changed

tracee-rules/input.go
```diff
@@ -66,8 +67,118 @@ func setupTraceeSource(traceeFilePath string) (chan types.Event, error) {
 				res <- event
 			}
 		}
-		f.Close()
+		opts.inputFile.Close()
```
itaysk, line 63: since you have unified stdin and file, you will now close stdin. I'm not sure whether this is good or not

tracee-rules/input.go
```diff
@@ -66,8 +67,118 @@ func setupTraceeSource(traceeFilePath string) (chan types.Event, error) {
...
+Specify various key value pairs for input options tracee-ebpf. The following key options are available:
+
+'file'   - Input file source. You can specify a relative or absolute path. You may also specify 'stdin' for standard input.
+'format' - Input format. Options currently include 'JSON' and 'GOB'. Both can be specified as output formats from tracee-ebpf.
+
+Examples:
+
+'tracee-rules --input-tracee file:./events.json --input-tracee format:json'
+'tracee-rules --input-tracee file:./events.gob --input-tracee format:gob'
+'sudo tracee-ebpf -o format:gob -o option:eot | tracee-rules --input-tracee file:stdin --input-tracee format:gob'
```
itaysk, line 173: remove eot

tracee-rules/main.go
```diff
@@ -51,28 +63,24 @@ func main() {
...
 				Name:  "list",
 				Usage: "print all available rules",
 			},
+			&cli.StringFlag{
+				Name:  "webhook",
+				Usage: "HTTP endpoint to call for every match",
+			},
+			&cli.StringSliceFlag{
+				Name:  "input-tracee",
+				Usage: "specify various key:value pairs for input options from tracee-ebpf, use '--input-tracee help' for more info",
```
itaysk, line 82: may I suggest an alternate text:
`configure tracee-ebpf as input source. see '--input-tracee help' for more info`

tracee-rules/input.go
```diff
@@ -66,8 +67,118 @@ func setupTraceeSource(traceeFilePath string) (chan types.Event, error) {
...
 		close(res)
 	}()
 	return res, nil
 }
+
+func setupJSONInputSource(opts *traceeInputOptions) (chan types.Event, error) {
+	res := make(chan types.Event)
+	scanner := bufio.NewScanner(opts.inputFile)
+	go func() {
+		for scanner.Scan() {
```
itaysk, line 73: restating what I wrote in a general comment. if we decided to always read until EOT, then this should be the behavior here as well. but since it's about to be removed, we can forgive this one this time

tracee-rules/input.go
grantseltzer: I think that's fine. At this point in the code we're going to be closing down so we won't need to read from this stdin fd anyway. Does feel a little weird though.

tracee-rules/input.go
grantseltzer: Got it. Agreed they need to be changed but this will require a bit more work because  when `parseTraceeInputOptions` is called, the channel of type `chan tracee.Events` can't be assigned to `chan types.Events` (channels of types are themselves their own types). I'm not sure fixing this should be in this PR, I can fix that in a follow up if you'd like.

tracee-rules/input.go
yanivagman: What about the case when there are multiple input sources? Not sure we can assume we're closing down
Also see my answer in #508

tracee-rules/input.go
itaysk: I don't understand what you wrote :) we are just suggesting the rename the function name, shouldn't have any implications. but if you want to leave it, that's fine

tracee-rules/input.go
itaysk: yes we can't assume we're closing the program but I guess closing the file won't do any harm

tracee-rules/input.go
grantseltzer: Lol I thought you were responding to https://github.com/aquasecurity/tracee/pull/500#issuecomment-772813353

tracee-rules/input.go
```diff
@@ -4,70 +4,174 @@ import (
...
+			event := scanner.Bytes()
+			var e tracee.Event
+			err := json.Unmarshal(event, &e)
+			if err != nil {
+				log.Printf("invalid json in %s: %v", string(event), err)
+			}
+			res <- tracee.Event(e)
```
simar7, line 80:
```suggestion
			res <- e
```

tracee-rules/input.go
```diff
@@ -4,70 +4,174 @@ import (
...
+
+	if opts.inputFormat == jsonInputFormat {
+		return setupTraceeJSONInputSource(opts)
 	}
-	f, err := os.Open(traceeFilePath)
-	if err != nil {
-		return nil, fmt.Errorf("invalid file: %s", traceeFilePath)
+
+	if opts.inputFormat == gobInputFormat {
+		return setupTraceeGobInputSource(opts)
```
simar7, line 40: nit: Replace with a switch statement.

tracee-rules/input.go
```diff
@@ -4,70 +4,174 @@ import (
...
 			if err != nil {
 				if err == io.EOF {
-					// ignore EOF because we assume events can keep streaming into the file
-					// this might create a backlog of events and depending on the implementation of the input source might lead to lost events
-					// TODO: investigate impact of this and research alternatives
-					time.Sleep(time.Millisecond * 150)
-					continue
+					break
 				} else {
 					log.Printf("Error while decoding event: %v", err)
```
simar7, line 57: Is it possible for the input file to be closed when we get here? If that happens we will be stuck in a loop forever.

simar7: (I realize it's not a change of logic here but it's just me looking at this with a fresh set of eyes)

tracee-rules/input.go
```diff
@@ -4,70 +4,174 @@ import (
...
+
+	if opts.inputFormat == gobInputFormat {
+		return setupTraceeGobInputSource(opts)
 	}
-	dec := gob.NewDecoder(f)
+
+	return nil, errors.New("could not set up input source")
+}
+
+func setupTraceeGobInputSource(opts *traceeInputOptions) (chan types.Event, error) {
```
simar7, line 46: This is mostly a question for myself but why does this simple test program not work?

```
func Test_setupTraceeGobInputSource(t *testing.T) {
	f, err := ioutil.TempFile("", "Test_setupTraceeGobInputSource-*")
	require.NoError(t, err)
	defer func() {
		_ = f.Close()
		_ = os.RemoveAll(f.Name())
	}()

	// write to file
	e := gob.NewEncoder(f)
	event := tracee.Event{
		EventName: "foo",
	}
	require.NoError(t, e.Encode(event))
	
	opts := traceeInputOptions{
		inputFile:   f,
		inputFormat: gobInputFormat,
	}

	res, err := setupTraceeGobInputSource(&opts)
	require.NoError(t, err)
	assert.Equal(t, "foo", (<-res).(tracee.Event).EventName)
}
```

simar7, line 46: Is this function tested somewhere else? Maybe we could write a simple test like:
```
func Test_setupTraceeGobInputSource(t *testing.T) {
	f, err := ioutil.TempFile("", "Test_setupTraceeGobInputSource-*")
	require.NoError(t, err)
	defer func() {
		_ = f.Close()
		_ = os.RemoveAll(f.Name())
	}()

	// write to file
	e := gob.NewEncoder(f)
	event := tracee.Event{
		EventName: "foo",
	}
	require.NoError(t, e.Encode(event))
	f.Seek(0, io.SeekStart)

	opts := traceeInputOptions{
		inputFile:   f,
		inputFormat: gobInputFormat,
	}

	res, err := setupTraceeGobInputSource(&opts)

	require.NoError(t, err)
	assert.Equal(t, "foo", (<-res).(tracee.Event).EventName)
}
```

itaysk (approved):
please also update the readme accordingly and squash before you merge

tracee-rules/input.go
grantseltzer: If it's all the same to you, I think this looks cleaner than this:

```

tracee-rules/input.go
grantseltzer: Hm, that is a good point but i'm not sure how it would?

tracee-rules/input.go
grantseltzer: In the interest of unblocking Itay I'm going to include this in a follow up PR, but yes you do make a good point, testing these like how you suggested would be a big win.

grantseltzer: Issue created: #517

Issue #501, opened 2021-02-02
Consider using filepath.Walk() to recursively scan for signatures


simar7:
Currently we do a custom logic to go through signatures in rules-dir: https://github.com/aquasecurity/tracee/blob/efdc613024f04f45b4be877372bd474f31da890d/tracee-rules/signature.go#L48-L98

We can probably improve this (also cut down on logic) if we use filepath.Walk() from stdlib.

please create an issue for this feature request and remove this from the code.. (same for the other similar comment)

_Originally posted by @itaysk in https://github.com/aquasecurity/tracee/pull/477#discussion_r566815240_

simar7:
Since the creation of this issue, Go 1.16 has released filepath.WalkDir which guarantees better perf than filepath.Walk. https://golang.org/pkg/path/filepath/#WalkDir

danielpacak:
Resolved by #901

PR #503, opened 2021-02-03
Updates scan for Signatures to use filepath.Walk()

NandoTheessen:
Hi 👋🏻 

I'm fairly new to go and wanted to make an attempt to contribute. 
I've started with the findGoSigs function, if the approach is correct I'll port the RegoSig function as well. 

Looking for any feedback and or pointers 👍🏻 

Addresses #501


tracee-rules/signature.go
```diff
@@ -45,27 +45,28 @@ func getSignatures(rulesDir string, rules []string) ([]types.Signature, error) {
...
-	if err != nil {
-		return nil, fmt.Errorf("error reading plugins directory %s: %v", dir, err)
-	}
 	var res []types.Signature
-	for _, file := range files {
-		if filepath.Ext(file.Name()) != ".so" {
-			continue
+	err := filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return fmt.Errorf("error opening plugin %s: %v", info.Name(), err)
```
itaysk, line 51: to my understanding this will cause the walker to abort. isn't it better to return SkipDir (and log it) so that other files might be still processed? same goes to all other errors returned from this func

tracee-rules/signature.go
```diff
@@ -45,27 +45,28 @@ func getSignatures(rulesDir string, rules []string) ([]types.Signature, error) {
...
+		}
+		if filepath.Ext(info.Name()) != ".so" {
```
itaysk, line 53: please add a check that this is a "regular" file. so that directories ending with ".so" won't be processed.

itaysk (commented):
thanks @NandoTheessen !
other than the comments inline, please also update the help to of the --rules-dir to indicate that this is a recursive search

itaysk:
oh and also, feel free to include the rego rules implementation in the same PR, it wouldn't make much sense to change just one and not the other

itaysk:
I'm also curious, what does `*upd:` means in the commit message?

NandoTheessen:
Thanks for the pointers itaysk 👋🏻
I've implemented the changes you requested, but wasn't sure on the wording for the readme file.
Please lmk if you'd like it differently.

I usually contain *upd (update) *fix or *merge in my commit messages to provide a context what kind of change I'm committing, I've changed that for the following commit messages, since I don't know how you handle it for this project.

tracee-rules/signature.go
itaysk: didn't you mean to SkipDir here too?

itaysk: also, log the error like in other cases

tracee-rules/signature.go
```diff
@@ -45,52 +45,55 @@ func getSignatures(rulesDir string, rules []string) ([]types.Signature, error) {
...
 		export, err := p.Lookup("ExportedSignatures")
 		if err != nil {
-			log.Printf("missing Export symbol in plugin %s", file.Name())
-			continue
+			log.Printf("missing Export symbol in plugin %s", info.Name())
+			return filepath.SkipDir
 		}
 		sigs := *export.(*[]types.Signature)
 		res = append(res, sigs...)
+		return filepath.SkipDir
```
itaysk, line 68: is this necessary?

tracee-rules/Readme.md
```diff
@@ -34,7 +34,7 @@ This will:
 When a detection is made by any of the signatures, it will be printed to stdout. Using the `--webhook` flag you can post detections into an HTTP endpoint that can further relay the detection.
 
 # Rules
-Rules are discovered from the local `rules` directory (unless changed by the `--rules-dir` flag). By default, all discovered rules will be loaded unless specific rules are selected using the `--rules` flag.
+Rules are discovered from the local `rules` directory and all it's subfolders (unless changed by the `--rules-dir` flag). By default, all discovered rules will be loaded unless specific rules are selected using the `--rules` flag.
```
itaysk, line 37: nit: change subfolders to sub-directories so that it's consistent with the beginning of the sentence

tracee-rules/signature.go
NandoTheessen: No, it's not. I've changed it to return `nil` instead.

simar7:
Thanks @NandoTheessen this is looking good! Could you also add a few basic tests to test this new functionality? I read that you are new to Go so here's a few tips to write your first test:

1. A simple test in Go would look like this:
```
func TestFoo(t *testing.T){
    want := "foo"
    got := fooFunc(input)
    assert.Equal(t, got, want)
}
```
to make a simple assert that `want == got` is true. 

2. In this PR you can call the `findGoSigs` func and the `findRegoSigs` func with `dir` as the input argument. In Go testing you can create a temporary directory just for a test and populate it with your test artifacts. See https://golang.org/pkg/io/ioutil/#TempDir and https://golang.org/pkg/io/ioutil/#TempFile for details.
3. When you have several input combinations, you can convert them into a table to form a table driven test. Here's an example of it in action https://github.com/aquasecurity/tracee/blob/main/tracee-rules/input_test.go

Let us know if you need help writing tests!

tracee-rules/signature.go
```diff
@@ -45,63 +44,62 @@ func getSignatures(rulesDir string, rules []string) ([]types.Signature, error) {
...
 	}
 
-	for _, file := range files {
-		if filepath.Ext(file.Name()) != ".rego" {
-			continue
-		}
-		if file.Name() == "helpers.rego" {
-			continue
+	var res []types.Signature
+	err = filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
```
simar7, line 84: The `err` passed into `filepath.Walk` must be checked before the actual function logic is executed. If a non nil error is passed in, the resulting info will be `nil` and this function will panic.

See another example here: https://github.com/aquasecurity/trivy/pull/208


Details: https://golang.org/pkg/path/filepath/#WalkFunc

tracee-rules/signature.go
```diff
@@ -45,63 +44,62 @@ func getSignatures(rulesDir string, rules []string) ([]types.Signature, error) {
...
+		if filepath.Ext(info.Name()) != ".rego" || info.IsDir() == true || info.Name() == "helpers.rego" {
+			return filepath.SkipDir
 		}
```
simar7, line 87: This isn't an error but rather a `return nil` instead. Returning an error here will completely skip this directory. I found this out by writing a simple test:

```
func TestFindRegoSigs(t *testing.T) {
	actual, err := findRegoSigs("signatures/rego")
	require.NoError(t, err)
	assert.NotEmpty(t, actual)
}
```

simar7 (changes requested):
Left some comments.

simar7:
Are you still interested in making those changes @NandoTheessen? Please let us know.

NandoTheessen:
Yes, been a little busy but would like to finish this up on the weekend :)

yanivagman:
As per @simar7 comment in #501 

> Since the creation of this issue, Go 1.16 has released filepath.WalkDir which guarantees better perf than filepath.Walk. https://golang.org/pkg/path/filepath/#WalkDir

Should we keep this PR open? Also, there were no changes made here for the last 5 months

itaysk:
looking at the feedback on this PR:
1. no tests - it would have been nice if the contributor added tests, but since the function was originally untested I wouldn't fight on this.
2. newer API slightly more efficient - remember this is small piece of code that is running once at init and is doing very simple work. I don't think it matters which API we use.
3. some code review comment - this IS a blocker. so let give it a couple of more days and if no response let's close.

Issue #435, opened 2021-01-22
create a dedicate documentation site


itaysk:
we should follow the same as demonstrated in https://github.com/aquasecurity/starboard , and use mkdocs, with similar design

ronaudinho:
I'll take a look and see if I can incorporate #436 as well.

Issue #255, opened 2020-10-13
Trace specific container


itaysk:
Add option to trace a specific container. This should be agnostic to container runtime, and probably use Container Runtime Interface. User experience need to be defined as per https://github.com/aquasecurity/tracee/issues/221

From https://github.com/aquasecurity/tracee/pull/249: 

> CRI, which is implemented by all container runtimes working with kubernetes (including docker, containerd and cri-o). A tool that uses this interface is crictl: https://github.com/kubernetes-sigs/cri-tools/blob/master/docs/crictl.md

> start with a much simpler case before diving into any of these approaches. Assuming the user can provide the pid of the container entrypoint (by inspecting the container to get this pid) - Tracee can go to procfs to get its pid ns inode number, and use it to trace all of the container pids (by adding the pid ns id to pids_map). This should be a good starting point, from where we will be able to automate more things in the future, like getting the container entrypoint pid from a container id for a given container runtime.

yanivagman:
Now that we added filters for UTS name, mount namespace id, and pid namespace id, it is possible to use these filters to trace a specific container(s).
We would like to have the ability to trace using the container id given by the container runtime, which is unknown to Tracee. The uts name is usually the same as the container id, and can be used for that purpose, but that cannot be trusted to always be the case.
Using cgroups is another filtering option, which we should add in the future to support container tracing.
Yet, we will probably have to implement some kind of interaction with the container runtime if we want to filter by container id correctly.

simar7:
> The uts name is usually the same as the container id, and can be used for that purpose, but that cannot be trusted to always be the case.

Just for my understanding, why it cannot be trusted to always be the case? What would be such a situation?

simar7:
I think I have found one case at least for that to happen. Running pods in k8s seem to have a different UTS name than what the CRI sees.

yanivagman:
UTS name is namespaced. As such, the hostname (uts name) can be changed from within the container

yanivagman:
For example, try to run a privileged container while tracee runs, then execute (in the container):
`hostname foobar`

yanivagman:
Now that we have container id as part of the context (#749) we can implement a filter by container id to complete this task.
Interaction with the container runtime through CRI should be handled in #430

PR #513, opened 2021-02-07
add general container and entrypoint

### LINKED ISSUE #463: repackage tracee container image
currently aquasec/tracee image is just tracee-ebpf. the next release will need to release tracee-ebpf as well as tracee-rules under the aquasec/tracee container. we should repackage the container (dockerfile) with these artifacts. the dockerfile should move out of tracee-ebpf to the root

### LINKED ISSUE #464: tracee entrypoint
We need a new entrypoint for the tracee container that orchestrates tracee the "runtime security" use case.
Also, this entrypoint should allow invoking tracee-ebpf directly when a subcommand is used (I suppose `trace`)

if running the container with no arguments it should:
1. start tracee-ebpf 
  1. write to stdout
  2. use gob format
  3. use EOT output option
2. pipe the output to tracee-rules
3. start tracee-rules
  1. use tracee input source from stdin
  2. load all build-in rules
  3. configure webhook to falco-sidekick
4. start falco-sidekick
  1. configured to read a config the will be mounted in

if running the container with `trace` subcommand
1. start tracee-ebpf
  1. pass all the additional args to the tracee-ebpf command

### LINKED ISSUE #466: create a readme for tracee
we now have individual readmes for tracee-rules and tracee-ebpf. we need the readme at the homepage to describe the e2e runtime security use case

itaysk:
close #463 
close #464
close #466

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,28 @@
+ARG BASE=fat
+
+FROM golang:alpine as builder
+RUN apk --no-cache update && apk --no-cache add git clang llvm make gcc libc6-compat coreutils linux-headers musl-dev elfutils-dev libelf-static zlib-static
+WORKDIR /tracee
+
+FROM builder as build
+ARG VERSION
+COPY . /tracee
+RUN make
+
+# base image for tracee which includes all tools to build the bpf object at runtime
+FROM alpine as fat
+RUN apk --no-cache update && apk --no-cache add clang llvm make gcc libc6-compat coreutils linux-headers musl-dev elfutils-dev libelf-static zlib-static
+
+# base image for tracee which includes minimal dependencies and expects the bpf object to be provided at runtime
+FROM alpine as slim
+RUN apk --no-cache update && apk --no-cache add libc6-compat elfutils-dev
+
+# final image
+FROM $BASE
+WORKDIR /tracee
+
+COPY --from=falcosecurity/falcosidekick@sha256:a32d8850d51e9b096a09f4ae73ba6cde038c3fe1fd9c58baf76333dfda7e7bbd /app/falcosidekick ./
+COPY --from=build /tracee/dist/tracee-ebpf /tracee/dist/tracee-rules /tracee/entrypoint.sh ./
+COPY --from=build /tracee/dist/rules ./rules
+
+ENTRYPOINT ["./entrypoint.sh"]
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ $(OUT_DIR):
 
 # make_artifact invokes another make from $1 to make the artifact $2 using additional flags $3
 define make_artifact
-	$(MAKE) -C $1 dist/$2 $3 OUT_DIR=dist && $(CMD_CP) -r $1/dist/$2 $(OUT_DIR)/$2
+	$(MAKE) -C $1 mostlyclean dist/$2 $3 OUT_DIR=dist && $(CMD_CP) -r $1/dist/$2 $(OUT_DIR)
 endef
 
 $(OUT_DIR)/tracee-ebpf: tracee-ebpf | $(OUT_DIR)
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,12 @@
-.PHONY: all
-all:
-	@echo "This make file is intended for releasing only. In order to build any of Tracee's components, use their respective make files."
-
-# RELEASE_TAG must be set to specify a name for the release
 OUT_DIR ?= dist
 OUT_DOCKER ?= tracee
 OUT_ARCHIVE := $(OUT_DIR)/tracee.tar.gz
 OUT_CHECKSUMS := $(OUT_DIR)/checksums.txt
 PUSH_DOCKER_REPO ?= aquasec/tracee
 PUSH_DOCKER_TAG ?= $(RELEASE_TAG:v%=%)
+RELEASE_FILES := LICENSE $(OUT_DIR)/tracee-ebpf $(OUT_DIR)/tracee-rules $(OUT_DIR)/rules
+# RELEASE_TAG must be set for the release target
+VERSION ?= $(if $(RELEASE_TAG),$(RELEASE_TAG),$(shell $(CMD_GIT) describe --tags 2>/dev/null || echo '0'))
 
 CMD_DOCKER ?= docker
 CMD_GIT ?= git
@@ -20,6 +18,12 @@ CMD_CP ?= cp
 release_tools := $(CMD_DOCKER) $(CMD_GIT) $(CMD_CHECKSUM) $(CMD_GITHUB) $(CMD_TAR) $(CMP_CP)
 check_release_tools := $(shell for tool in $(release_tools); do command -v $$tool >/dev/null || (echo "missing required tool $$tool" ; false); done)
 
+# DOCKER variable is used by sub-makefiles to build their artifact in a container
+export DOCKER
+
+.PHONY: all
+all: $(RELEASE_FILES)
+
 $(OUT_DIR):
 	mkdir -p $@
 
@@ -29,7 +33,7 @@ define make_artifact
 endef
 
 $(OUT_DIR)/tracee-ebpf: tracee-ebpf | $(OUT_DIR)
-	$(call make_artifact,$<,$(notdir $@),DOCKER=1)
+	$(call make_artifact,$<,$(notdir $@))
 
 $(OUT_DIR)/tracee-rules $(OUT_DIR)/rules: tracee-rules | $(OUT_DIR)
 	$(call make_artifact,$<,$(notdir $@))
@@ -40,14 +44,17 @@ $(OUT_DIR)/tracee.bpf.%.o: bpf ;
 bpf: tracee-ebpf | $(OUT_DIR)
 		$(MAKE) -C $< mostlyclean $@ OUT_DIR=dist && $(CMD_CP) $</dist/tracee.bpf.*.o $(OUT_DIR)
 
-release_archive_files := LICENSE $(OUT_DIR)/tracee $(OUT_DIR)/tracee-rules $(OUT_DIR)/rules
-$(OUT_ARCHIVE) $(OUT_CHECKSUMS) &: $(release_archive_files) | $(OUT_DIR)
-	$(CMD_TAR) -czf $(OUT_ARCHIVE) $(release_archive_files)
+$(OUT_ARCHIVE) $(OUT_CHECKSUMS) &: $(RELEASE_FILES) | $(OUT_DIR)
+	$(CMD_TAR) -czf $(OUT_ARCHIVE) $(RELEASE_FILES)
 	$(CMD_CHECKSUM) $(OUT_ARCHIVE) > $(OUT_CHECKSUMS)
 
 .PHONY: docker
-docker: tracee-ebpf
-	$(MAKE) -C tracee-ebpf docker docker-slim OUT_DOCKER=$(OUT_DOCKER)
+docker:
+	$(CMD_DOCKER) build --build-arg VERSION=$(VERSION) -t $(OUT_DOCKER):latest .
+
+.PHONY: docker-slim
+docker-slim:
+	$(CMD_DOCKER) build --build-arg VERSION=$(VERSION) -t $(OUT_DOCKER):slim --build-arg BASE=slim .
 
 # release_docker_image accepts a docker image $1, pushes it as $2 to remote repository, and records it in the release notes
 define release_docker_image
--- a/Readme.md
+++ b/Readme.md
@@ -9,9 +9,97 @@
 
 Tracee is a Runtime Security and forensics tool for Linux. It is using Linux eBPF technology to trace your system and applications at runtime, and analyze collected events to detect suspicious behavioral patterns.
 
-This repo contains the following projects:
+Tracee is delivered as a Docker image that once run, will start to monitor the OS and detect suspicious behavior based on a pre-defined set of behavioral patterns.
+
+Tracee is composed of the following sub-projects:
 - [Tracee-eBPF](tracee-ebpf) - Linux Tracing and Forensics using eBPF
 - [Tracee-Rules](tracee-rules) - Runtime Security Detection Engine
 - [libbpgo](libbpfgo) - Go library for eBPF programming using Linux's [libbpf](https://github.com/libbpf/libbpf)
 
-The repo is currently in a state of transition into a monorepo containing these multiple projects. If you are looking for the previous "Tracee" tool, look in the "tracee-ebpf" directory.
+## Getting started
+
+### Prerequisites
+
+- Linux kernel version >= 4.14
+- Relevant kernel headers available under conventional location (see [Linux Headers](#Linux-Headers) section for info)
+- libc, and the libraries: libelf and zlib
+- clang >= 9
+
+Exceptions:
+
+- Tracee supports loading a pre-compiled eBPF file, in which case the kernel headers are required only for the one-time compilation, and not at runtime. See Setup Options for more info.
+- When using Tracee's Docker image, all of the aforementioned requirements are built into the image. The only requirement left is the kernel headers or the pre-built eBPF. See Setup Options for more info.
+
+### Quickstart with Docker
+
+```bash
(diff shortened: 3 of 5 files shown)


entrypoint.sh
```diff
@@ -0,0 +1,32 @@
...
+  $TRACEE_MOUNT_EXE -t debugfs debugfs /sys/kernel/debug/
+fi
+
+if [ "$1" = "trace" ]; then
+	shift
+	$TRACEE_EBPF_EXE $@
+	exit
+fi
+
+if [ "$1" = "--rules" ]; then
```
yanivagman, line 20: Why we use "--" here, unlike "trace"?
Also, shouldn't we accept any other flags than "--rules"?

entrypoint.sh
```diff
@@ -0,0 +1,32 @@
...
+
+if [ -f $TRACEE_WEBHOOK_CONFIG ]; then
+	flag_webhook_config="--config-file=$TRACEE_WEBHOOK_CONFIG"
+fi
+
+if $TRACEE_WEBHOOK_EXE $flag_webhook_config& then
+	flag_webhook_url="--webhook=http://localhost:2801"
+fi
+
+$TRACEE_EBPF_EXE --output=format:gob --output=option:eot | $TRACEE_RULES_EXE --tracee-file=/dev/stdin $flag_webhook_url $flag_rules
```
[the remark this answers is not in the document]
itaysk: did you read the PR message? I'm waiting for it to be merged then I'll make this ready.

Is there a problem with stdin? why would it not scale?

entrypoint.sh
itaysk: > Why we use "--" here, unlike "trace"?

because trace is a subcommand, and this is a flag to the root command

> shouldn't we accept any other flags than "--rules"

yes eventually, this PR just satisfies the MVP scope so we can make a first appearance

entrypoint.sh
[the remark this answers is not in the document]
yanivagman: `why would it not scale?`
As we already discussed, adding other sources will require the usage of pipes, and can't use stdin in parallel

`Is it something that would stop the MVP? Or can be addressed later?`
Doesn't have to stop MVP, but it feels like a regression to me.
We had file input (not stdin), which could have supported streaming input source before the removal of eot.
Now we took step back - and only stdin supports streaming

entrypoint.sh
yanivagman: `because trace is a subcommand, and this is a flag to the root command`
Didn't we say that `rules` will be a subcommand just like `trace` is?
Also, if we choose to have a tracee-ebpf image like I suggested in #463 maybe we can just remove the `trace` subcommand.

entrypoint.sh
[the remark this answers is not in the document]
itaysk: > adding other sources will require the usage of pipes, and can't use stdin in parallel

oh ok.. but we have just one input source now, and file is not a viable option (see below)

> We had file input (not stdin), which could have supported streaming input source before the removal of eot.
Now we took step back - and only stdin supports streaming

I don't think it's a step back because the file based solution was practically unusable (file would indefinitely grow with no way to trim or partition it).

Completely agree that the best way forward is something like named pipes, but we want to release soon and for the sake of the MVP, I think this can be good enough.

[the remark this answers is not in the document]
yanivagman: Agree. For MVP that's enough.

> I don't think it's a step back because the file based solution was practically unusable (file would indefinitely grow with no way to trim or partition it).

File doesn't have to grow indefinetly. it's maximum size can be specified and drop events when full (like a circular buffer). But I don't think that going forward this is the correct solution anyways, I just wanted to make the point that using stdin would not scale

entrypoint.sh
itaysk: > Didn't we say that rules will be a subcommand just like trace is?

I don't think so, maybe I missed something. but I don't see the point in adding a `rules` subcommand, and even if we had it, we would probably have made it the default

> Also, if we choose to have a tracee-ebpf

seems like we have swapped opinions :) initially I thought like you're suggesting (separate images), and then you suggested to use the same image with subcommand, and I think it a great idea, because we already have an image with all the tools inside, why not allow people to use those tools directly. 
We can still add another container dedicated for tracee-ebpf if you want to (I don't see why, it's just a few MB difference), but that doesn't have to mean we can't expose tracee-ebpf also from this one

yanivagman: Yes, we can keep the `trace` subcommand.
I still think we need to have another image for tracee-ebpf as it is self-contained, and users who only need tracing (and forensics) capabilities, can just use it instead of the "tracee" image and have a simpler UX for their needs.
Moreover, the fact that today the difference is only just a few MB doesn't say that this will be the situation in the future - it will probably keep growing as we will add signatures and other functionality

itaysk: ok.. I didn't say we can't have another image. this is a discussion on the entrypoint of `tracee`

entrypoint.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh
```
grantseltzer, line 1: We should set pipefail in our bash scripts:`set -euf -o pipefail`

([ref](https://sipb.mit.edu/doc/safe-shell/))

itaysk: as a general rule of thimb I guess this is true, but in this case:
- `pipefail` is not portable
- `-u` will break checks for optional arguments (which may be legitimately unset)
- `-e` will break checking for non-zero exit codes (like grep didn't find or test eval to false)

maybe there's a way to rewrite the entire script to work in strict mode but I'm not sure it's worth the effort

simar7:
Just poking stuff but how do I run this? And how do I exit? A simple SIGINT doesn't seem to have an effect.

```
vagrant@ubuntu-groovy:/usr/vagrant/repos/tracee$ docker run itaysk:entry
mount: permission denied (are you root?)
2021/02/08 19:26:14 error opening plugin builtin.so: plugin.Open("/tracee/rules/builtin.so"): Error loading shared library /tracee/rules/builtin.so: Exec format error
2021/02/08 19:26:14 [INFO]  : Enabled Outputs : 
2021/02/08 19:26:14 [INFO]  : Falco Sidekick is up and listening on port 2801
2021/02/08 19:26:14 missing kernel source code compilation dependency
^C^C^C
```

itaysk:
@simar7 check out the tracee-ebpf readme, try something like:
```
docker run  --rm --privileged --pid=host -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro itaysk:entry 
```

I'm looking into the SIGINT issue

itaysk:
@simar7 I need more info about the SIGINT issue you're having not sure I can repro

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+FROM alpine as slim
+RUN apk --no-cache update && apk --no-cache add libc6-compat elfutils-dev
+
+# final image
+FROM $BASE
+WORKDIR /tracee
+
+COPY --from=falcosecurity/falcosidekick@sha256:a32d8850d51e9b096a09f4ae73ba6cde038c3fe1fd9c58baf76333dfda7e7bbd /app/falcosidekick ./
+COPY --from=build /tracee/tracee-ebpf/dist/tracee-ebpf /tracee/tracee-rules/dist/tracee-rules /tracee/entrypoint.sh ./
+COPY --from=build /tracee/tracee-rules/dist/rules ./rules
```
simar7, line 26: Should we rebuild these artifacts before we copy?

itaysk:
> @simar7 I need more info about the SIGINT issue you're having not sure I can repro

@simar7 I think what's happening is that you didn't attach TTY, you are just looking at it's stdout. try adding `-it` to docker run

simar7:
> > @simar7 I need more info about the SIGINT issue you're having not sure I can repro
> 
> @simar7 I think what's happening is that you are running in foreground but didn't attach TTY, you are just looking at it's stdout. try adding `-it` to docker run

Ah that would explain. But I'm running into another issue:

```
vagrant@ubuntu-groovy:/usr/vagrant/repos/tracee$ docker run -it  --rm --privileged --pid=host -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro itaysk:entry
Error relocating /tracee/tracee-rules: __vfprintf_chk: symbol not found
Error relocating /tracee/tracee-rules: __fprintf_chk: symbol not found
2021/02/08 19:43:40 [INFO]  : Enabled Outputs : 
2021/02/08 19:43:40 [INFO]  : Falco Sidekick is up and listening on port 2801
failed to add kprobe 'p:kprobes/psecurity_inode_unlink security_inode_unlink': -17
failed to create kprobe event: -17
```

itaysk:
looks like your container is packing execs that were build on your host using glibc. how did you build the image?

Dockerfile
itaysk: good catch. I'll fix it

itaysk:
just to be sure, delete all `dist` directories from the root and from `tracee-*` and `docker build -t something .` in the root

simar7:
> looks like your container is packing execs that were build on your host using glibc. how did you build the image?

To build this image:
```
docker build . -t itaysk:entry 
```



> looks like your container is packing execs that were build on your host using glibc. how did you build the image?


I removed all `dist` dirs and rebuilt both tracee-ebpf and tracee-rules prior to building this docker image. For building the first two I just ran `make all` in their respective directories and to build the image I ran `docker build --no-cache -t itaysk:entry .`

Here's the most recent output:
```
docker run -it  --rm --privileged --pid=host -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro itaysk:entry
Error relocating /tracee/tracee-rules: __vfprintf_chk: symbol not found
Error relocating /tracee/tracee-rules: __fprintf_chk: symbol not found
Error relocating /tracee/tracee-ebpf: __snprintf_chk: symbol not found
Error relocating /tracee/tracee-ebpf: __vfprintf_chk: symbol not found
Error relocating /tracee/tracee-ebpf: __memcpy_chk: symbol not found
Error relocating /tracee/tracee-ebpf: __strcpy_chk: symbol not found
Error relocating /tracee/tracee-ebpf: __sprintf_chk: symbol not found
```

I'm building on a Ubuntu 20.10 FWIW, I never ran into issues running tracee-ebpf or tracee-rules on their own.

yanivagman:
@simar7 note that we have two dockerfiles - Dockerfile and Dockerfile,builder.
The first Dockerfile is using alpine (musl libc), and is what we use while doing a release.
The second Dockerfile.builder is using debian base image (glibc) and is what used for local builds.
It may be that you mixed the two in some way - built locally the executables (using glibc), then build the images using the first Dockerfile

itaysk:
> rebuilt both tracee-ebpf and tracee-rules prior to building this docker image

don't do this. just clean all dists and run `docker build -t .` that's it

simar7:
> > rebuilt both tracee-ebpf and tracee-rules prior to building this docker image
> 
> don't do this. just clean all dists and run `docker build -t .` that's it

Thanks that did the trick. I think we should document how to use the entrypoint/image with this PR. Thoughts?

What if we clean the respective `dist/` directories when we run this entry point? Or is there any other way to ignore the current artifacts present within them to build the entry point image?

itaysk:
> I think we should document how to use the entrypoint/image

yes absolutely

> What if we clean the respective dist/ directories when we run this entry point? Or is there any other way to ignore the current artifacts present within them to build the entry point image?

I'll try to think of something

simar7:
> > I think we should document how to use the entrypoint/image
> 
> yes absolutely

It could be just as simple as the following (I didn't know where this should live to I'm writing it down here):

### Tracee Runner Image
In order to facilitate ease of use, we've created a general Tracee container that contains ready to use release binaries for tracee-ebpf and tracee-rules. It also contains an instance of Falco sidekick to facilitate webhook events.

To run this container, simply: 
`make build-tracee-runner` which runs the following docker command
`docker build -t aquasec/tracee-runner:latest` to build a docker image.

This will do the following:
1. Build tracee-ebpf and tracee-rules binaries.
2. Copy the above built as artifacts into the new runner container image.
3. Copy Falco sidekick (https://github.com/falcosecurity/falcosidekick) for Webhook delivery.

The runner image needs to run in docker privileged mode. A simple invocation is like the following:
```docker run -it  --rm --privileged --pid=host -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro aquasec/tracee-runner```

Or by invoking the Makefile target: `make run-tracee-runner`

Note here that I made the following assumptions:
1. We call this orchestration/demo container as "tracee-runner". I'm happy with another name. Also feel free to change the make target names.
2. Added a bunch of Makefile targets to run the docker commands we need to build.

File: entrypoint.sh [hunk 5]
```diff
@@ -0,0 +1,32 @@
...
+
+TRACEE_EBPF_EXE=${TRACEE_EBPF_EXE:="/tracee/tracee-ebpf"}
+TRACEE_MOUNT_EXE=${TRACEE_MOUNT_EXE:="mount"}
+TRACEE_WEBHOOK_CONFIG=${TRACEE_WEBHOOK_CONFIG:="/tmp/tracee/integrations-config.yaml"}
```
simar7, line 5: For my understanding: How is this used? I see Falco sidekick has a WebUI that I was expecting to see if I didn't pass in any arguments https://github.com/falcosecurity/falcosidekick-ui and https://github.com/falcosecurity/falcosidekick/blob/master/config_example.yaml#L164-L165

I'm just trying to configure Falco sidekick to be able to look at the events that Tracee will generate.

itaysk: IDK what's webUI, seems like a new feature added to master 2 days ago and yet unreleased

itaysk: I just configured it with MS Teams output to see that it works

Readme.md
```diff
@@ -9,9 +9,95 @@
...
+### Setup options
+
+Tracee is leveraging Linux's eBPF technology, which is kernel and version sensitive, and therefore needs to be specifically compiled for your hosts.
+
+The easiest way to get started is to just let Tracee build the eBPF program for you automatically when it starts, as demonstrated by the [Quickstart](#quickstart).  
+Alternatively, you can pre-compile the eBPF program, and provide it to Tracee. There are some benefits to this approach as you will not need clang and kernel headers at runtime anymore, as well as reduced risk of invoking an external program at runtime.
+
+You can build the eBPF program in the following ways:
+1. Clone the repo and `make bpf`.
+2. `make bpf DOCKER=1` to build in a Docker container which includes all development tooling.
```
itaysk, line 66: I'll add this in a future PR. don't want to make this PR larger then it already is

itaysk:
added readme to the root level. it should include all of the operational instructions that applies to tracee-ebpf so I copied most of it in. it's Once we complete #435 (post-MVP) we will refactor the docs anyway.
also added `--list` flag to the container as a way for users to see what rules are provided, until we have AVD integration

Readme.md
itaysk: #519

grantseltzer (approved):

simar7 (approved):

Makefile
```diff
@@ -1,14 +1,12 @@
...
-# RELEASE_TAG must be set to specify a name for the release
 OUT_DIR ?= dist
 OUT_DOCKER ?= tracee
 OUT_ARCHIVE := $(OUT_DIR)/tracee.tar.gz
 OUT_CHECKSUMS := $(OUT_DIR)/checksums.txt
 PUSH_DOCKER_REPO ?= aquasec/tracee
 PUSH_DOCKER_TAG ?= $(RELEASE_TAG:v%=%)
+RELEASE_FILES := LICENSE $(OUT_DIR)/tracee-ebpf $(OUT_DIR)/tracee-rules $(OUT_DIR)/rules
+# RELEASE_TAG must be set for the release target
+VERSION ?= $(if $(RELEASE_TAG),$(RELEASE_TAG),$(shell $(CMD_GIT) describe --tags 2>/dev/null || echo '0'))
```
yanivagman, line 9: Can we now remove these same lines from tracee-ebpf/Makefile and other similar stuff as well?

itaysk: no because it's used to determine the name of the bpf obj (tracee.bpf.linuxversion.traceeversion.o)

itaysk: but we could clean up the tracee-ebpf makefile depending on what we decide to do with #255 (I'm leaving it to that PR)

PR #523, opened 2021-02-09
fix: Move example sigs into own dir and exclude from build.

### LINKED ISSUE #514: exclude example signatures
We have example signatures in tracee-rules, which live next to the regular signatures, and are included in the final build. We need to make sure that the examples are excluded from final product (or at least not being activated). I think the best approach would be to move them to a different directory will not get picked up by the build. Maybe a "test" directory because at some point we should use those example sigs for testing

### LINKED ISSUE #524: Link to the examples are broken
Trying to follow the examples in `tracee-rules` https://github.com/aquasecurity/tracee/blob/fc534300281f1ca60498cf49b196825432054e07/tracee-rules/Readme.md 

The example.go and rego are pointing to broken links.

simar7:
I initially tried with go build flags `// +build ignore` but the Go compiler still picks it up. I wonder if it has to do anything with the `-buildmode=` flag we use. Nevertheless to get traction I did it in the Makefile instead.

**Before**
```
/dist/tracee-rules  --list
Code injection: Possible process injection detected during runtime
detect self debugging using PTRACE_ME: Process uses anti-debugging technique to block debugger
example1: 
example2: 
```

**After**
```
./dist/tracee-rules --list
Code injection: Possible process injection detected during runtime
detect self debugging using PTRACE_ME: Process uses anti-debugging technique to block debugger
```

Fixes #514 
Fixes #524 


tracee-rules/signatures/golang/examples/example.go
```diff
@@ -1,4 +1,4 @@
-package main
+package examples
```
itaysk, line 1: this is invalid in go plugins

tracee-rules/Makefile
```diff
@@ -6,10 +6,10 @@ OUT_BIN := $(OUT_DIR)/tracee-rules
 OUT_RULES := $(OUT_DIR)/rules
 GO_SRC := $(shell find . -type f -name '*.go')
 GOSIGNATURES_DIR ?= signatures/golang
-GOSIGNATURES_SRC := $(shell find $(GOSIGNATURES_DIR) -type f -name '*.go' ! -name '*_test.go')
+GOSIGNATURES_SRC := $(shell find $(GOSIGNATURES_DIR) -type f -name '*.go' ! -name '*_test.go' ! -path '*examples*')
```
itaysk, line 9: small one, but perhaps it'll be safer to change this to `$(GOSIGNATURES)/examples/*` same for the rego one. not a stopper though, whatever you think

tracee-rules/signatures/golang/examples/example.go
simar7: Thanks - I blame IDE auto refactoring 😄

tracee-rules/Makefile
simar7: Yeah why not - I added your suggestion.

itaysk (approved):

Issue #515, opened 2021-02-08
ID for rules


itaysk:
Currently rules are identified by their name. If a user wants to select which rules to load, they need to give their full name on the command like, which is not ideal. We should add an ID field to signatures' metadata and use that in the --rules option.

yanivagman:
Signatures will need to add their own signature id?
This may cause conflicts (in the future) when using signatures from two different sources, and require signature writers to check for an available id.

itaysk:
good point! on the other hand, auto generating IDs will be hard to make deterministic. another consideration is that we need to be able to support 3rd party sigs, that were not bundled with tracee

itaysk:
Here's a suggestion: ID takes the following form: `XXX-12345`.
`XXX` is a prefix that indicates the source of the sig, for example for builtin signatures we can have `TRACEE` prefix and the number is a running number that uniquely identifies signatures.
For buildin signatures, because it's controlled by us and every new sig has to go through a PR, we can control how numerical ids are issued (just look at the latest). and for others - just use your own prefix and then you're in charge it's numerical IDs.
examples:
`TRACEE-123`
`ITAYSK-321`

yanivagman:
I'm not sure I understand the benefit of using such IDs. What's the difference between the user entering an ID compared to signature name? Is it the size of the name of the signature? If so, there are other possible solution that don't use ID for that, like using a unique prefix of a signature name

itaysk:
1. format of the field: name can be any size and include any character. so entering a name on the command line, into a flag that takes csv, is not ideal experience
2. uniqueness: id is unique and name doesn't have to be

yanivagman:
> format: name can be any size and include any character. so entering a name on the command line, into a flag that takes csv, is not ideal experience

Is remembering the signature ID considered an ideal experience? I don't think so

> uniqueness: id is unique and name doesn't have to be

We can require that names will be unique, I don't see a reason that two different rules will have the same name.
If you still think that signatures IDs are required (for the MVP? why?), I think they should be auto generated.
An example of a deterministic ID generation can be taking the hash of the signature name or filename (in the case that we decide that signature names are not unique)

itaysk:
> Is remembering the signature ID considered an ideal experience? I don't think so

not sure what you mean by remember. you do `tracee-rules --list` and see which rules are available, and if you want to pick just some of them you see their ID right there

> If you still think that signatures IDs are required (for the MVP? why?)

[...]

PR #525, opened 2021-02-09
Rename signatures and add spacing to printing of them with --list flag

grantseltzer:
Small change but this will make output look like this:

Screen Shot 2021-02-09 at 2 55 39 PM

### DIFF
--- a/tracee-rules/main.go
+++ b/tracee-rules/main.go
@@ -28,12 +28,13 @@ func main() {
 				return err
 			}
 			if c.Bool("list") {
+				fmt.Printf("%-30s %s\n", "RULE NAME", "DESCRIPTION")
 				for _, sig := range sigs {
 					meta, err := sig.GetMetadata()
 					if err != nil {
 						continue
 					}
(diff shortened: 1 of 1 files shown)


tracee-rules/signatures/rego/anti_debugging_ptraceme.rego
```diff
@@ -1,8 +1,8 @@
 package main
 
 __rego_metadoc__ := {
-    "name": "detect self debugging using PTRACE_ME",
-    "description": "Process uses anti-debugging technique to block debugger",
+    "name": "PTRACE_ME",
```
itaysk, line 4: I'm not sure about the name change.. this is supposed to be eventually displayed in user facing UIs (DTA/AVD/etc). If the intent is to make it accessible for selecting on the command like, #515 should fix that

grantseltzer: Ah alright, will revert

itaysk: Other sig as well?

itaysk (approved):

Issue #491, opened 2021-02-01
integrate rego tests in the make test target


itaysk:
need to either wrap the tests in go code, or manually invoke the opa test command. consider the benefits of each.

PR #535, opened 2021-02-10
feat: Add OPA tests to Github Actions

simar7:
There aren't docs around to use this action but looking at the yaml, I was able to understand it. https://github.com/petroprotsakh/opa-test-action/blob/master/action.yml

Saves us the hassle from building our own setup each time (clone OPA binary, etc.)

In theory it should run this (from the tracee-rules dir):
```
$ opa test . --verbose --ignore="examples"
data.main.test_match_1: PASS (3.399054ms)
data.main.test_match_wrong_request: PASS (114.405µs)
--------------------------------------------------------------------------------
PASS: 2/2
```

Fixes https://github.com/aquasecurity/tracee/issues/491

Requires: https://github.com/aquasecurity/tracee/pull/567

### DIFF
--- a/.github/workflows/pr-rules.yaml
+++ b/.github/workflows/pr-rules.yaml
@@ -30,4 +30,4 @@ jobs:
       - name: Build signatures
         run: make rules
       - name: Run Tests
-        run: make test
+        run: make test DOCKER=1
--- a/.github/workflows/pr-rules.yaml
+++ b/.github/workflows/pr-rules.yaml
@@ -16,6 +16,8 @@ jobs:
         uses: actions/setup-go@v1
         with:
           go-version: 1.15
+      - name: Setup OPA
+        run: make setup-opa
       - name: Lint
         run: |
           if test -z "$(gofmt -l .)"; then
@@ -29,10 +31,5 @@ jobs:
         run: make build
       - name: Build signatures
         run: make rules
-      - name: Run Unit Tests
+      - name: Run Tests
         run: make test
-      - name: Run OPA Tests
-        uses: petroprotsakh/opa-test-action@v2.1
-        with:
-          tests: "tracee-rules/"
-          options: "--verbose --ignore='examples'"
--- a/.github/workflows/pr-rules.yaml
+++ b/.github/workflows/pr-rules.yaml
@@ -16,8 +16,6 @@ jobs:
         uses: actions/setup-go@v1
         with:
           go-version: 1.15
-      - name: Setup OPA
-        run: make setup-opa
       - name: Lint
         run: |
           if test -z "$(gofmt -l .)"; then
--- a/.github/workflows/pr-rules.yaml
+++ b/.github/workflows/pr-rules.yaml
@@ -29,5 +29,10 @@ jobs:
         run: make build
       - name: Build signatures
         run: make rules
-      - name: Test
+      - name: Run Unit Tests
         run: make test
+      - name: Run OPA Tests
+        uses: petroprotsakh/opa-test-action@v2.1
+        with:
+          tests: "tracee-rules/"
+          options: "--verbose --ignore='examples'"
--- a/tracee-ebpf/Dockerfile.builder
+++ b/tracee-ebpf/Dockerfile.builder
@@ -1,5 +1,5 @@
 FROM golang:1.15-buster as builder
 RUN echo "deb http://apt.llvm.org/buster/ llvm-toolchain-buster-9 main" >> /etc/apt/sources.list && apt-key adv --keyserver hkps://keyserver.ubuntu.com --recv-keys 15CF4D18AF4F7421 && \
-    DEBIAN_FRONTEND=noninteractive apt-get update -y && apt-get install -y --no-install-recommends libelf-dev llvm-9-dev clang-9 curl && \
+    DEBIAN_FRONTEND=noninteractive apt-get update -y && apt-get install -y --no-install-recommends libelf-dev llvm-9-dev clang-9 && \
     (for tool in "clang" "llc" "llvm-strip"; do path=$(which $tool-9) && ln -s $path ${path%-*}; done)
 WORKDIR /tracee
--- a/tracee-ebpf/Dockerfile.builder
+++ b/tracee-ebpf/Dockerfile.builder
@@ -1,6 +1,5 @@
 FROM golang:1.15-buster as builder
 RUN echo "deb http://apt.llvm.org/buster/ llvm-toolchain-buster-9 main" >> /etc/apt/sources.list && apt-key adv --keyserver hkps://keyserver.ubuntu.com --recv-keys 15CF4D18AF4F7421 && \
     DEBIAN_FRONTEND=noninteractive apt-get update -y && apt-get install -y --no-install-recommends libelf-dev llvm-9-dev clang-9 curl && \
-    (for tool in "clang" "llc" "llvm-strip"; do path=$(which $tool-9) && ln -s $path ${path%-*}; done) && \
-    curl -L -o /usr/bin/opa https://openpolicyagent.org/downloads/latest/opa_linux_amd64 && chmod 755 /usr/bin/opa
+    (for tool in "clang" "llc" "llvm-strip"; do path=$(which $tool-9) && ln -s $path ${path%-*}; done)
 WORKDIR /tracee
--- a/tracee-ebpf/Dockerfile.builder
+++ b/tracee-ebpf/Dockerfile.builder
@@ -1,5 +1,6 @@
 FROM golang:1.15-buster as builder
 RUN echo "deb http://apt.llvm.org/buster/ llvm-toolchain-buster-9 main" >> /etc/apt/sources.list && apt-key adv --keyserver hkps://keyserver.ubuntu.com --recv-keys 15CF4D18AF4F7421 && \
-    DEBIAN_FRONTEND=noninteractive apt-get update -y && apt-get install -y --no-install-recommends libelf-dev llvm-9-dev clang-9 && \
-    (for tool in "clang" "llc" "llvm-strip"; do path=$(which $tool-9) && ln -s $path ${path%-*}; done)
+    DEBIAN_FRONTEND=noninteractive apt-get update -y && apt-get install -y --no-install-recommends libelf-dev llvm-9-dev clang-9 curl && \
+    (for tool in "clang" "llc" "llvm-strip"; do path=$(which $tool-9) && ln -s $path ${path%-*}; done) && \
+    curl -L -o /usr/bin/opa https://openpolicyagent.org/downloads/latest/opa_linux_amd64 && chmod 755 /usr/bin/opa
 WORKDIR /tracee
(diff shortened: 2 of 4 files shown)


simar7:
I just realized that there's a limitation with this approach as it stands. We only have one rego rule at the moment. If we had multiple, there would be redeclaration issues that'll we will run into when running the `opa test` command. See an example below.

```
$ opa test -v .                
1 error occurred: example1.rego:3: rego_type_error: rule named __rego_metadoc__ redeclared at example2.rego:3
```

Another idea I thought of was keeping each signature in its own package rather than main. But I'm not sure if there's any inherent assumptions we make when we keep them under main. One place I see it being used is here https://github.com/aquasecurity/tracee/blob/da310b07bbc71705b1c12c2e6ccb6cb19a5cbf33/tracee-rules/signatures/rego/regosig/traceerego.go#L29-L31 

If we can build these query strings on the fly (generating them dynamically based on package name) we could name packages uniquely. 

Regardless, I think we should open a separate issue to keep track of the above as no new rules are being added in the MVP. Thoughts?

itaysk:
it's better to add this to the Makefile test target rather then just in GitHub actions. It's also easier. Since you opted for using the `opa` binary, it's just a one line change to the tracee-rules Makefile.

itaysk:
regarding the package names for rego rules - Yes as we discussed, that's expected. it is quite common in OPA land (to namespace policies), and if we have #536 we should use the id as the package name. and yes that won't be a problem to incorporate into the query string

simar7:
> it's better to add this to the Makefile test target rather then just in GitHub actions. It's also easier. Since you opted for using the `opa` binary, it's just a one line change to the tracee-rules Makefile.

We'll need to make sure OPA client is installed first so I updated the PR accordingly https://github.com/aquasecurity/tracee/pull/535/commits/d2c5c5e5c342855a64d9368a71f00389693e1c4c

itaysk:
the makefile has many other binary dependencies, which we don't install for the user but rather check for existence, I think this should be similar. and btw you have just introduced another binary dependency - curl :)
I think it would be best to add a `check_opa` requirement as well as a user overridable variable. see how we handle clang or git for example

itaysk:
Almost forgot, add to the docker builder as well

simar7:
> Almost forgot, add to the docker builder as well

What's a docker builder?

simar7:
> the makefile has many other binary dependencies, which we don't install for the user but rather check for existence, I think this should be similar. and btw you have just introduced another binary dependency - curl :)
> I think it would be best to add a `check_opa` requirement as well as a user overridable variable. see how we handle clang or git for example

But don't we need to have `opa` within Github Actions? I agree with your comment about checking first than installing for the user but in this case that won't work as Github Actions runner doesn't have it.

itaysk:
GitHub Actions runs the tests with DOCKER=1 which runs in a docker container. This is the "docker builder" container. So I suggest to just add opa to the dockerfile.builder, and in the makefile just ensure that it's there.

simar7:
> GitHub Actions runs the tests with DOCKER=1 which runs in a docker container. This is the "docker builder" container. So I suggest to just add opa to the dockerfile.builder, and in the makefile just ensure that it's there.

But the pr-rules.yaml isn't setting DOCKER=1 for running tests https://github.com/aquasecurity/tracee/blob/main/.github/workflows/pr-rules.yaml#L32-L33

Are you suggesting to run them inside of a docker container as well?

simar7:
I built a new docker builder to run tests within a docker container within GitHub Actions. 

> regarding the package names for rego rules - Yes as we discussed, that's expected. it is quite common in OPA land (to namespace policies), and if we have #536 we should use the id as the package name. and yes that won't be a problem to incorporate into the query string

Since #536 isn't merged just yet, running tests isn't possible since they all live under `package main`. This is a chicken and egg problem again, as the tests cannot be run until they live under proper namespace.

We can go two ways:
1. Add proper namespacing as part of this PR starting with Rego signatures first. Each rego signature will have its own namespace. Which would enable us to run tests.
2. Merge this PR as is - tests won't run until a new PR is created to fix signature namespacing.

I'm fine with either approach - thoughts @itaysk?

tracee-ebpf/Dockerfile.builder
```diff
@@ -1,5 +1,6 @@
 FROM golang:1.15-buster as builder
 RUN echo "deb http://apt.llvm.org/buster/ llvm-toolchain-buster-9 main" >> /etc/apt/sources.list && apt-key adv --keyserver hkps://keyserver.ubuntu.com --recv-keys 15CF4D18AF4F7421 && \
-    DEBIAN_FRONTEND=noninteractive apt-get update -y && apt-get install -y --no-install-recommends libelf-dev llvm-9-dev clang-9 && \
-    (for tool in "clang" "llc" "llvm-strip"; do path=$(which $tool-9) && ln -s $path ${path%-*}; done)
+    DEBIAN_FRONTEND=noninteractive apt-get update -y && apt-get install -y --no-install-recommends libelf-dev llvm-9-dev clang-9 curl && \
+    (for tool in "clang" "llc" "llvm-strip"; do path=$(which $tool-9) && ln -s $path ${path%-*}; done) && \
+    curl -L -o /usr/bin/opa https://openpolicyagent.org/downloads/latest/opa_linux_amd64 && chmod 755 /usr/bin/opa
```
itaysk, line 5: why did you add opa to tracee-ebpf's container?

tracee-rules/Dockerfile.builder
```diff
@@ -0,0 +1,5 @@
+FROM golang:1.15-buster as builder
+RUN echo "deb http://apt.llvm.org/buster/ llvm-toolchain-buster-9 main" >> /etc/apt/sources.list && apt-key adv --keyserver hkps://keyserver.ubuntu.com --recv-keys 15CF4D18AF4F7421 && \
```
itaysk, line 2: why do we need llvm in tracee-rules?

tracee-rules/Makefile
```diff
@@ -27,11 +31,38 @@ $(OUT_RULES): $(GOSIGNATURES_DIR) $(REGO_SIGNATURES_SRC) | $(OUT_DIR)
...
+
+# docker_builder_make runs a make command in the tracee-builder container
+define docker_builder_make
+	$(CMD_DOCKER) run --rm \
+	-v $(abspath .):/tracee/tracee-rules \
+	-w /tracee/tracee-rules \
+	--entrypoint make $(DOCKER_BUILDER) $(1)
+endef
+
+# record built image id to prevent unnecessary building and for cleanup
```
itaysk, line 64: did you mean to also clean it up?

tracee-rules/Makefile
```diff
@@ -1,6 +1,9 @@
 .PHONY: all
 all: build rules
 
+CMD_DOCKER ?= docker
+CMD_GO ?= go
+CMD_OPA ?= opa
```
itaysk, line 6: where are these used?

tracee-rules/Makefile
```diff
@@ -10,6 +13,7 @@ GOSIGNATURES_SRC := $(shell find $(GOSIGNATURES_DIR) -type f -name '*.go' ! -nam
 OUT_GOSIGNATURES := $(OUT_RULES)/builtin.so
 REGO_SIGNATURES_DIR ?= signatures/rego
 REGO_SIGNATURES_SRC := $(shell find $(REGO_SIGNATURES_DIR) -type f -name '*.rego' ! -name '*_test.rego' ! -path '$(REGO_SIGNATURES_DIR)/examples/*')
+DOCKER_BUILDER ?= tracee-rules
```
itaysk, line 16: I think it's misleading to call the builder container tracee-rules (since it doesn't contains tracee-rules)

.github/workflows/pr-rules.yaml
```diff
@@ -29,5 +29,5 @@ jobs:
         run: make build
       - name: Build signatures
         run: make rules
-      - name: Test
-        run: make test
+      - name: Run Tests
+        run: make test DOCKER=1
```
itaysk, line 33: where is this used?

itaysk (changes requested):

itaysk:
> running tests isn't possible since they all live under package main

It could be possible if you run each test individually. This is how it's still working for Engine.
but I agree it's better to implement the desired solution and not an intermediary

tracee-rules/Makefile
simar7: what should we call it then?

itaysk: tracee-rules-builder?

tracee-ebpf/Dockerfile.builder
```diff
@@ -1,5 +1,5 @@
 FROM golang:1.15-buster as builder
 RUN echo "deb http://apt.llvm.org/buster/ llvm-toolchain-buster-9 main" >> /etc/apt/sources.list && apt-key adv --keyserver hkps://keyserver.ubuntu.com --recv-keys 15CF4D18AF4F7421 && \
-    DEBIAN_FRONTEND=noninteractive apt-get update -y && apt-get install -y --no-install-recommends libelf-dev llvm-9-dev clang-9 && \
+    DEBIAN_FRONTEND=noninteractive apt-get update -y && apt-get install -y --no-install-recommends libelf-dev llvm-9-dev clang-9 curl && \
```
itaysk, line 3: no reason for curl here right?

tracee-rules/Dockerfile.builder
```diff
@@ -0,0 +1,5 @@
+FROM golang:1.15-buster as builder
+RUN apt-key adv --keyserver hkps://keyserver.ubuntu.com --recv-keys 15CF4D18AF4F7421 && \
```
itaysk, line 2: this is the key for llvm, it's no longer needed

tracee-rules/Dockerfile.builder
```diff
@@ -0,0 +1,5 @@
...
+    DEBIAN_FRONTEND=noninteractive apt-get update -y && apt-get install -y --no-install-recommends curl && \
+    curl -L -o /usr/bin/opa https://openpolicyagent.org/downloads/latest/opa_linux_amd64 && chmod 755 /usr/bin/opa
```
itaysk, line 4: would be better to fetch a specific version instead of latest

tracee-rules/Makefile
itaysk: not sure this was actually addressed

simar7: As discussed: Need to add these in to the clean target in this makefile https://github.com/aquasecurity/tracee/blob/main/tracee-ebpf/Makefile#L149-L150

simar7:
I rebased on latest main after merging the other PRs, the tests now work as expected:
```
opa test . --verbose --ignore="examples" --ignore="dist"
data.tracee.TRC_2.test_match_1: PASS (1.4527ms)
data.tracee.TRC_2.test_match_wrong_request: PASS (269.7µs)
data.tracee.TRC_3.test_match_1: PASS (477.6µs)
data.tracee.TRC_3.test_match_2: PASS (603.7µs)
data.tracee.TRC_3.test_match_3: PASS (2.3133ms)
data.tracee.TRC_3.test_match_wrong_request: PASS (658.1µs)
data.tracee.TRC_3.test_match_wrong_pathname: PASS (352µs)
data.tracee.TRC_3.test_match_pid: PASS (207µs)
data.tracee.TRC_4.test_match_1: PASS (211.8µs)
data.tracee.TRC_4.test_match_wrong_message: PASS (259µs)
data.tracee.TRC_4.test_match_wrong_event: PASS (163.1µs)
data.tracee.TRC_5.test_match_1: PASS (279µs)
data.tracee.TRC_5.test_match_2: PASS (290.9µs)
data.tracee.TRC_5.test_match_wrong_pathname: PASS (221.6µs)
data.tracee.TRC_6.test_match_1: PASS (164.5µs)
data.tracee.TRC_6.test_match_2: PASS (185.9µs)
data.tracee.TRC_6.test_match_wrong_event: PASS (211.8µs)
data.tracee.TRC_7.test_match_1: PASS (290.9µs)
data.tracee.TRC_7.test_match_no_ld_preload: PASS (286.1µs)
--------------------------------------------------------------------------------
PASS: 19/19
```

itaysk (approved):

PR #536, opened 2021-02-11
Add IDs to Signature Metadata

simar7:
This PR introduces Signature IDs as part of their Metadata as discussed in https://github.com/aquasecurity/tracee/issues/515

### DIFF
--- a/tracee-rules/signatures/golang/code_injection_test.go
+++ b/tracee-rules/signatures/golang/code_injection_test.go
@@ -3,12 +3,15 @@ package main
 import (
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
 	"github.com/aquasecurity/tracee/tracee-rules/signatures/signaturestest"
 	"github.com/aquasecurity/tracee/tracee-rules/types"
 	tracee "github.com/aquasecurity/tracee/tracee/external"
 )
 
-func TestCodeInjection(t *testing.T) {
+func TestCodeInjection_OnEvent(t *testing.T) {
 	SigTests := []signaturestest.SigTest{
 		{
 			Events: []types.Event{
@@ -182,4 +185,20 @@ func TestCodeInjection(t *testing.T) {
 			t.Error("unexpected result", st)
 		}
 	}
+}
+
+func TestCodeInjection_GetMetadata(t *testing.T) {
+	sig := codeInjection{}
+	got, err := sig.GetMetadata()
+	require.NoError(t, err)
+	assert.Equal(t, types.SignatureMetadata{
+		ID:          "TRC-1",
+		Name:        "Code injection",
+		Description: "Possible process injection detected during runtime",
+		Tags:        []string{"linux", "container"},
+		Properties: map[string]interface{}{
+			"Severity":     3,
+			"MITRE ATT&CK": "Defense Evasion: Process Injection",
+		},
+	}, got)
 }
--- a/tracee-rules/signatures/golang/examples/example.go
+++ b/tracee-rules/signatures/golang/examples/example.go
@@ -26,6 +26,7 @@ func (sig *counter) Init(cb types.SignatureHandler) error {
 // GetMetadata implements the Signature interface by declaring information about the signature
 func (sig *counter) GetMetadata() (types.SignatureMetadata, error) {
 	return types.SignatureMetadata{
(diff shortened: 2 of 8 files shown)


tracee-rules/signatures/golang/code_injection_test.go
```diff
@@ -183,3 +186,19 @@ func TestCodeInjection(t *testing.T) {
 		}
 	}
 }
+
+func TestCodeInjection_GetMetadata(t *testing.T) {
```
itaysk, line 190: are you sure we want to test this function? it doesn't contain any code, just a declaration. if we wanted to removed the error returned (discussed in meeting), then it's an indication that there's nothing here to test.

tracee-rules/signatures/golang/examples/example.go
```diff
@@ -26,6 +26,7 @@ func (sig *counter) Init(cb types.SignatureHandler) error {
 // GetMetadata implements the Signature interface by declaring information about the signature
 func (sig *counter) GetMetadata() (types.SignatureMetadata, error) {
 	return types.SignatureMetadata{
+		ID:   "TRC-12345",
```
itaysk, line 29: let's give the examples an invalid ID, like TRC-0 or XMP-1

itaysk (changes requested):
thanks @simar7 for taking this

as part of this PR we need to also add this to the `--list` output and the rules list in the readme (I would much rather remove that from the readme but that's another discussion)

if you want in a different PR, we should update the `--rules` flag to use this and update the readme accordingly

If you want in a followup PR we should update the

itaysk:
probably need to watch out for conflicts with https://github.com/aquasecurity/tracee/pull/528

tracee-rules/signatures/rego/anti_debugging_ptraceme.rego
```diff
@@ -1,6 +1,7 @@
 package main
 
 __rego_metadoc__ := {
+    "id": "TRC-1",
```
yanivagman, line 4: TRC-1 is used for both of this signature and code_injection.go signature - isn't it a problem?

simar7: Correct - I'll wait to update this PR until #528 gets merged.

Actually you bring up another good point – Sometimes (not in this case) we might have the same signature implemented in both Rego and Go. Do we give the same ID in that case?

tracee-rules/signatures/golang/code_injection_test.go
simar7: You're right, it doesn't do anything today. But having said that it may change tomorrow in which case a test would ensure behaviour correctness. Speaking of which I'm now leaning more towards keeping the error in the signature.

tracee-rules/signatures/rego/anti_debugging_ptraceme.rego
yanivagman: We will need to find a proper way to track the next available TRC number.
As I mentioned previously, looking forward this will be a point of collisions in the future - think about two open PRs that want to add a new signature - will they both use the same next available number? how will they find it?

itaysk: > Sometimes (not in this case) we might have the same signature implemented in both Rego and Go

I don't think this is a desired state to be in. this will also mean the sig will trigger twice. we shouldn't implement sigs in both

> Do we give the same ID in that case?

I don't think so, it's not the same sig, even if the BL is similar. but again we should never have this case

itaysk: > think about two open PRs that want to add a new signature - will they both use the same next available number? how will they find it?

I honestly don't think this is such a big issue (yet). It's the PR reviewer's responsibility to make sure the number is valid (not just if it's available). and if it's not, to correct the mistake. in rare cases, maybe the reviewer will manually coordinate between 2 open PRs and tell each one which number to use.

Another solution might be to keep a file in the repo that contains the last committed version, and each new sig PR will update it. kind of like a mutex. this can help both easily identify what's the next number and also avoid conflicts.
(it won't prevent 2 people TRYING to use the same number but it will prevent them from succeeding in using it)

simar7:
Closing in interest of https://github.com/aquasecurity/tracee/pull/567

PR #542, opened 2021-02-14
tracee-ebpf: Add commit_creds event

yanivagman:


tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1810,6 +1812,76 @@ int BPF_KPROBE(trace_security_inode_unlink)
...
+    kgid_t old_egid = READ_KERN(old->egid);
+    kgid_t new_egid = READ_KERN(new->egid);
+    kuid_t old_fsuid = READ_KERN(old->fsuid);
+    kuid_t new_fsuid = READ_KERN(new->fsuid);
+    kernel_cap_t old_cap_eff = READ_KERN(old->cap_effective);
+    kernel_cap_t new_cap_eff = READ_KERN(new->cap_effective);
+
+    // Currently (2021), there are ~40 capabilities in the Linux kernel which are stored in a 2 dimentional u32 array
+    // This might change in the (not so near) future as more capabilities will be added.
+    // For now, we use u64 to store this array in one piece
```
itaysk, line 1845: if you want, you could check if `_KERNEL_CAPABILITY_U32S == 2` to verify

itaysk (approved):

tracee-ebpf/tracee/tracee.bpf.c
yanivagman: It will take a few more years until this will be changed to length of 3 (more than 20 capabilities will have to be added first), and we do not support kernel version with cap length of 1 - so I think it is ok  to skip such test for now

PR #543, opened 2021-02-15
update import paths after restructure

itaysk:
closes: https://github.com/aquasecurity/tracee/issues/497
The restucture changed the paths in main but didn't
effect the Go import paths which looks for a given tag. This commit
updated the import paths in preparation for the release (which will
create the tag)


itaysk:
it's ready for review but I'm keeping it draft because it needs to be the last commit before the release.
Also not that the commit "DO NOT COMMIT - temp test for CI" should be reverted before merging, it is here only to allow the build to pass.

yanivagman (approved):

itaysk:
we can't change the go module path in the same PR as the reference to it, so I added a replace directive to workaround it, and the plan is to remove it in the next PR

PR #547, opened 2021-02-15
fix match for non af_inet sockets

roikol:


roikol:
hey @itaysk,
just a quick fix for a bug added in one of my last commits. can you review?

itaysk:
@roikol can you please share what was the problem with the current code? it's hard to tell the difference

roikol:
@itaysk 
in the signature code, for each PID i maintain a map of socket FDs to the IP this socket is connected to (FD -> IP). this means we are only interested in sockets of type AF_INET/AF_INET6. 
the bug was that i updated the map (FD -> IP) for every type of socket, with a default IP value of an empty string "".
this PR aims to fix that - only AF_INET/AF_INET6 sockets go into the map, with real IP values.

itaysk (approved):

tracee-rules/signatures/golang/stdio_over_socket.go
```diff
@@ -73,20 +73,24 @@ func (sig *stdioOverSocket) OnEvent(e types.Event) error {
...
-		}
 
-		_, pidExists := sig.processSocketIp[pid]
-		if !pidExists {
-			sig.processSocketIp[pid] = make(map[int]string)
-		}
+			_, pidExists := sig.processSocketIp[pid]
+			if !pidExists {
+				sig.processSocketIp[pid] = make(map[int]string)
+			}
```
yanivagman, line 90: These lines are the same as lines 78-81 and can be added once before this if statement

yanivagman (approved):
One comment about code duplication, but should not block merge

tracee-rules/signatures/golang/stdio_over_socket.go
itaysk: but then it will create a PID entry in the map for other socket types. I think this is what Roi wanted to fix here, no?

yanivagman: No. Previously, non inet sockfds where saved as well, and this is not the case if you create a new map for the pid (which is empty). To avoid creating an empty map, it is possible to put the dup code once in an if (inet or inet6)...

PR #550, opened 2021-02-16
add tests for entrypoint

itaysk:
the approach taken here is to mock the commands used (e.g system commands like "mount" or our commands like "tracee-ebpf") with mocks that satisfies the basic functionality (e.g we don't want mount to actually mount something, but it should say that it did, so we can know that it did it's job). Also the log of operations performed is kept and we can assert which commands were "executed" (their mocks did). 
note that all mocks are just symlinks to the same file.


simar7:
Let's document somewhere that this is using `bats-core` and not the original `bats`. Would be helpful to someone who isn't aware of the distinction.

simar7 (approved):
lgtm, we can add more bats based tests in the future 👍🏼

simar7:
Can we merge this? Or is it waiting on us wiring these test targets into a Github Action?

PR #552, opened 2021-02-16
Fix stdio over socket

roikol:
export the signature so it will be loaded by tracee-rules.

fix FD assignment to int32 rather than int, because it causes panic.

### DIFF
--- a/tracee-rules/signatures/golang/stdio_over_socket.go
+++ b/tracee-rules/signatures/golang/stdio_over_socket.go
@@ -61,7 +61,7 @@ func (sig *stdioOverSocket) OnEvent(e types.Event) error {
 			return err
 		}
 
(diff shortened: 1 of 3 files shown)


tracee-rules/signatures/golang/stdio_over_socket.go
```diff
@@ -61,7 +61,7 @@ func (sig *stdioOverSocket) OnEvent(e types.Event) error {
 			return err
 		}
 
-		sockfd := sockfdArg.Value.(int)
+		sockfd := int(sockfdArg.Value.(int32))
```
itaysk, line 64: can't you change the map type to be int32, instead of double casting?

itaysk (approved):
approving it because it's needed for the release today

PR #555, opened 2021-02-17
Revert "fix release as monorepo"

itaysk:
we decided to take another approach which is to use specific commit to identify a dependency instead of a tag


itaysk:
included in #543

PR #567, opened 2021-02-19
feat: Add IDs to Signature Metadata.

simar7:
This PR intends to introduces the following:

- [x] Signature IDs as part of their Metadata as discussed in #515
- [x] Namespaces the rego signatures appropriately.
- [x] Logic to handle new namespaced rego signatures

### DIFF
--- a/tracee-rules/signatures/rego/anti_debugging_ptraceme.rego
+++ b/tracee-rules/signatures/rego/anti_debugging_ptraceme.rego
@@ -1,6 +1,7 @@
 package main
 
 __rego_metadoc__ := {
+    "id": "TRC-2",
     "name": "Anti-Debugging",
     "description": "Process uses anti-debugging technique to block debugger",
     "tags": ["linux", "container"],
--- a/tracee-rules/signatures/rego/anti_debugging_ptraceme.rego
+++ b/tracee-rules/signatures/rego/anti_debugging_ptraceme.rego
@@ -1,4 +1,4 @@
-package main
+package anti_debugging_ptraceme
 
 __rego_metadoc__ := {
     "id": "TRC-2",
--- a/tracee-rules/signatures/rego/anti_debugging_ptraceme.rego
+++ b/tracee-rules/signatures/rego/anti_debugging_ptraceme.rego
@@ -1,4 +1,4 @@
-package anti_debugging_ptraceme
+package tracee.TRC_2
 
 __rego_metadoc__ := {
     "id": "TRC-2",
--- a/tracee-rules/signatures/rego/anti_debugging_ptraceme_test.rego
+++ b/tracee-rules/signatures/rego/anti_debugging_ptraceme_test.rego
@@ -1,4 +1,4 @@
-package main
+package anti_debugging_ptraceme
 
 test_match_1 {
     tracee_match with input as {
--- a/tracee-rules/signatures/rego/anti_debugging_ptraceme_test.rego
+++ b/tracee-rules/signatures/rego/anti_debugging_ptraceme_test.rego
@@ -1,4 +1,4 @@
-package tracee.TRC_3
+package tracee.TRC_2
 
 test_match_1 {
     tracee_match with input as {
--- a/tracee-rules/signatures/rego/anti_debugging_ptraceme_test.rego
+++ b/tracee-rules/signatures/rego/anti_debugging_ptraceme_test.rego
@@ -1,4 +1,4 @@
-package anti_debugging_ptraceme
+package tracee.TRC_3
 
 test_match_1 {
     tracee_match with input as {
--- a/tracee-rules/signatures/rego/regosig/traceerego.go
+++ b/tracee-rules/signatures/rego/regosig/traceerego.go
@@ -5,6 +5,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"regexp"
+	"strings"
 
 	tracee "github.com/aquasecurity/tracee/tracee-ebpf/tracee/external"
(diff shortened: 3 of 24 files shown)


tracee-rules/signatures/rego/anti_debugging_ptraceme.rego
```diff
@@ -1,6 +1,7 @@
-package main
+package anti_debugging_ptraceme
```
simar7, line 1: Note: I have called it `anti_debugging_ptraceme` rather than `tracee.TRC-2` this is because dashes don't seem to be allowed characters in rego package names https://www.openpolicyagent.org/docs/latest/policy-language/#packages

I thought it's just easier to rely on the filename here since filenames will always have to be unique.

simar7: I _could_ also call it `tracee.TRC2`, fine either way I believe.

itaysk: thanks for spotting this. Not that I'm suggesting this, just an allegory: If you want to use filename here, why not use filename for ID as well?
I tend to think we should stick with TRC2/TRC_2, and for consistency, rename the files as well. wdyt?

tracee-rules/signatures/rego/regosig/traceerego.go
```diff
@@ -26,18 +28,25 @@ type RegoSignature struct {
...
 func NewRegoSignature(regoCodes ...string) (types.Signature, error) {
 	var err error
 	res := RegoSignature{}
 	regoMap := make(map[string]string)
 
-	for index, regoCode := range regoCodes {
-		regoModuleName := fmt.Sprintf("rego_%d", index)
+	re := regexp.MustCompile(packageNameRegex)
+
+	var pkgName string
```
itaysk, line 44: not a big deal but I think this could be a struct field

itaysk (commented):
this to complete the associated issue, please also update the --rules flag accordingly

tracee-rules/signatures/rego/anti_debugging_ptraceme.rego
simar7: updated

simar7:
> this to complete the associated issue, please also update the --rules flag accordingly

I didn't quite get that, did you mean update the `--list` flag output? If so I have pushed that change now.

itaysk:
> didn't quite get that, did you mean update the --list flag output? 

no, I meant the `--rules` flag. today it expects name, it should expect id. this was the motivation for adding IDs to rules

simar7:
Update to include `--list` with the following fields:
```
ID         NAME                                DESCRIPTION
TRC-1      Standard Input/Output Over Socket   Redirection of process's standard input/output to socket
TRC-2      Anti-Debugging                      Process uses anti-debugging technique to block debugger
TRC-3      Code injection                      Possible code injection into another process
TRC-4      Dynamic Code Loading                Writing to executable allocated memory region
TRC-5      Fileless Execution                  Executing a precess from memory, without a file in the disk
TRC-6      kernel module loading               Attempt to load a kernel module detection
TRC-7      LD_PRELOAD                          Usage of LD_PRELOAD to allow hooks on process
```

tracee-rules/signatures/rego/anti_debugging_ptraceme_test.rego
```diff
@@ -1,4 +1,4 @@
-package main
+package tracee.TRC_3
```
itaysk, line 1: conflicts with code_injection

itaysk (changes requested):

tracee-rules/signatures/rego/anti_debugging_ptraceme_test.rego
simar7: Thanks for the catch, I updated it.

itaysk (approved):

tracee-rules/signatures/rego/regosig/traceerego.go
```diff
@@ -26,18 +28,25 @@ type RegoSignature struct {
...
+	for _, regoCode := range regoCodes {
+		regoModuleName := strings.Split(re.FindString(regoCode), " ")[1]
+		if !strings.Contains(regoCode, "package tracee.helpers") {
+			pkgName = regoModuleName
+		}
 		regoMap[regoModuleName] = regoCode
 	}
```
yanivagman, line 52: Shouldn't we check that pkgName is not an empty string (for the case of "package tracee.helpers")?

tracee-rules/signatures/rego/regosig/traceerego.go
```diff
@@ -26,18 +28,25 @@ type RegoSignature struct {
...
 
-	for index, regoCode := range regoCodes {
-		regoModuleName := fmt.Sprintf("rego_%d", index)
+	re := regexp.MustCompile(packageNameRegex)
+
+	var pkgName string
+	for _, regoCode := range regoCodes {
+		regoModuleName := strings.Split(re.FindString(regoCode), " ")[1]
+		if !strings.Contains(regoCode, "package tracee.helpers") {
+			pkgName = regoModuleName
```
yanivagman, line 48: I see that you assign here pkgName, which is used later, out of this for loop.
This will make only the last regoCode package name to be saved - is this the intended behavior?

itaysk: I think it is, we need the package name from just one of the (non-helpers) files since all of them are assumed to be in the same package.

itaysk: we could `break` once we found one though it's not a big deal

Issue #569, opened 2021-02-21
improve detection output


itaysk:
This is what we currently print to stdout on a rule match:
```
{Data:map[] Context:{Timestamp:3340.723838 ProcessID:9806 ThreadID:9806 ParentProcessID:9775 HostProcessID:9806 HostThreadID:9806 HostParentProcessID:9775 UserID:1000 MountNS:4026531840 PIDNS:4026531836 ProcessName:gdb HostName:yaniv-pc EventID:101 EventName:ptrace ArgsNum:4 ReturnValue:0 StackAddresses:[] Args:[{ArgMeta:{Name:request Type:long} Value:PTRACE_TRACEME} {ArgMeta:{Name:pid Type:pid_t} Value:0} {ArgMeta:{Name:addr Type:void*} Value:0x0} {ArgMeta:{Name:data Type:void*} Value:0x7FBC52AF553E}]} Signature:0xc0000e4500}
```

we should make this more useful, for example:
```
*** Detection ***
Time: {{time.Now().String()}}
Signature ID: {{.Signature.SignatureID}}
Signature: {{.Signature.Name}}
Data: {{.Data}}
Command: {{.Context.ProcessName}}
Hostname (container): {{.Context. HostName}}
```

simar7:
An example to reproduce this issue is the following:

```
dist/tracee-rules --input-tracee file:stdin --input-tracee format:json
{​​​​ "eventName": "ptrace", "args": [ {​​​​"name": "request", "value": "PTRACE_TRACEME"}​​​​ ] }​​​​
```

Which will give us the desired output:
```
{​​​​Data:map[] Context:{​​​​Timestamp:0 ProcessID:0 ThreadID:0 ParentProcessID:0 HostProcessID:0 HostThreadID:0 HostParentProcessID:0 UserID:0 MountNS:0 PIDNS:0 ProcessName: HostName: EventID:0 EventName:ptrace ArgsNum:0 ReturnValue:0 StackAddresses:[] Args:[{​​​​ArgMeta:{​​​​Name:request Type:}​​​​ Value:PTRACE_TRACEME}​​​​]}​​​​ Signature:0xc0000b6600}​​​​
```

simar7:
Closed via https://github.com/aquasecurity/tracee/pull/573

PR #573, opened 2021-02-23
feat: Better formatted output for detected events.

simar7:
Addresses: #569 

### DIFF
--- a/tracee-rules/output.go
+++ b/tracee-rules/output.go
@@ -16,13 +16,11 @@ import (
 const DetectionOutput string = `
 *** Detection ***
 Time: %s
-SignatureID: %s
+Signature ID: %s
 Signature: %s
-ProcessName: %s
-ProcessID: %d
-ParentProcessID: %d
+Data: %s
+Command: %s
 Hostname: %s
-EventName: %s
 `
 
 func setupOutput(resultWriter io.Writer, clock Clock, webhook string) (chan types.Finding, error) {
@@ -37,12 +35,9 @@ func setupOutput(resultWriter io.Writer, clock Clock, webhook string) (chan type
 
 			switch res.Context.(type) {
 			case tracee.Event:
-				processName := res.Context.(tracee.Event).ProcessName
-				processID := res.Context.(tracee.Event).ProcessID
-				parentProcessID := res.Context.(tracee.Event).ParentProcessID
+				command := res.Context.(tracee.Event).ProcessName
 				hostName := res.Context.(tracee.Event).HostName
-				eventName := res.Context.(tracee.Event).EventName
-				fmt.Fprintf(resultWriter, DetectionOutput, clock.Now().UTC().Format(time.RFC3339), sigMetadata.ID, sigMetadata.Name, processName, processID, parentProcessID, hostName, eventName)
+				fmt.Fprintf(resultWriter, DetectionOutput, clock.Now().UTC().Format(time.RFC3339), sigMetadata.ID, sigMetadata.Name, res.Data, command, hostName)
 			default:
 				log.Printf("unsupported event detected: %T\n", res.Context)
 				continue
--- a/tracee-rules/output.go
+++ b/tracee-rules/output.go
@@ -44,6 +44,7 @@ func setupOutput(resultWriter io.Writer, clock Clock, webhook string) (chan type
 				fmt.Fprintf(resultWriter, DetectionOutput, clock.Now().UTC().Format(time.RFC3339), sigMetadata.Name, processName, processID, parentProcessID, hostName, eventName)
 			default:
 				log.Printf("unsupported event detected: %T\n", res.Context)
+				continue
 			}
 
 			if webhook != "" {
--- a/tracee-rules/output.go
+++ b/tracee-rules/output.go
@@ -3,6 +3,7 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"log"
 	"net/http"
 	"strings"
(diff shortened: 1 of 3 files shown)


simar7:
Opening this up for discussion.

tracee-rules/output.go
```diff
@@ -12,11 +13,36 @@ import (
...
+const DetectionOutput string = `
+*** Detection ***
+Time: %d
+Signature: %s
+ProcessName: %s
+ProcessID: %d
+ParentProcessID: %d
+Hostname: %s
+EventName: %s
+`
```
simar7, line 25: What else can we add here? Can we always assume that we will get a `tracee.Event` type as input to process? Not having a real type but rather an interface makes this output little tricky to be deterministic (see my code comment).

itaysk: the context depends on the input source. currently we only have tracee input sources so context is always tracee.Event, but when we add more input sources, we'll have more types. we should assert the type of context and print the output accordingly.

itaysk: in a go template that will be more complicated though. I've separated the gotemplate out of this issue so that you can merge this PR first

tracee-rules/output.go
```diff
@@ -12,11 +13,36 @@ import (
 	"github.com/aquasecurity/tracee/tracee-rules/types"
 )
 
-func setupOutput(webhook string) (chan types.Finding, error) {
+const DetectionOutput string = `
+*** Detection ***
+Time: %d
```
itaysk, line 18: can you please format as time?

simar7: I'm printing Unix time which is an int64, do you want a different format instead?

itaysk: Anything human readable, don't you think?

simar7: OK I can change it to a more conventional format.

simar7: I changed to RFC3339 https://github.com/aquasecurity/tracee/pull/573/commits/29873e935ae6b87a46f2281863d0546e51699737

tracee-rules/output.go
simar7: Did you create a separate issue to track the templating stuff? Sorry if I missed it.

itaysk: was going to but then realized it has some complexity which needs to be specified which I didn't get around to. also I thought that this is less important now since stdout is just there for debugging/experimentation users are encouraged to use webhook anyway. I'll create a placeholder now, which I'll leave out of the next milestone: https://github.com/aquasecurity/tracee/issues/580

itaysk (approved):

tracee-rules/output.go
yanivagman: I think that each signature should have its own output context, which should be submitted by the signature itself.
The context can describe more complicated things like: "process x connected to ip y which is blacklisted".
This can be implemented by the signature adding an extra "output" field to the Finding.Data map, which we can then print here in case it was specified by the signature.
In case that the output field was specified by the signature, we can also remove the processName, pid, and ppid which may not be of interest to the context of the signature match.

One more field that should be added here after the PR will be merged is the signature ID

itaysk: regarding per sig output. this is mentioned in https://github.com/aquasecurity/tracee/issues/580, I've added your comment for extra context (that issue has different motivation but same implementation)

itaysk: regarding ID - yes, and more importantly AVD link

itaysk: @yanivagman do you have a blocking concern or can I merge this?

yanivagman: I don't think it's the same thing.
In #580 you want to change a user facing feature by using a gotemplate that can be customized by the user.
What I mean here is letting the signature specify it's output message, which can be different between different signatures, according to the context - this is not a thing that can be addressed by a gotemplate as it is different for each and every signature.
I think that we should add a new string field to the signature interface which is called "output", and which is controlled by the signature itself. If we choose to do so, this should be done as part of this PR that should take care of the signature output formatting

itaysk: yes I understand your suggestion, it's a good one. #580 was discussing a different use case but overlapping implementation. So I have converted that into discussion so we can continue and discuss how to implement it (more along your suggestion).
So I agree with your suggestion but I don't think it's blocking this one? if at some point we add a user friendly message to the sig we just add another field here, just like when we add ID we add another field here. does that make sense?

yanivagman: Not blocking, but I think that some of these fields should not be printed on every signature match.
These are: process name, process ID, PPID, hostname and event name - which are almost all of the added fields.
This relates to my previous comment that every signature has its own context.
Here are a few examples:
1. A signature that is triggered by multiple events (connect, accept, dup, etc...) - the current event name is not relevant
2. A signature that counts the number of events of a certain type - doesn't require the ppid, process name, etc.
3. An "inter-process" signature that correlates events of different processes - doesn't need any process context.

And there are more examples.

yanivagman: In a second thoght - it is blocking.
All of the fields added here are not a MUST for every signature.
This should be the signature's choice

itaysk: ok I agree.. but we have to choose how to proceed, either we abandon this PR and incorporate your suggestions to it (which requires further discussion), or merge this one (it's still a step forward from the current state), and improve it in a future PR (after some discussion)

yanivagman: How is this a step forward if we will remove all of these fields soon?
I think we should discuss this before merging this one

itaysk: This Isssue/PR was triggered by your comment, where you said regarding the output format that: "It's really hard to understand out of it what really happened. We should probably remove some of the event metadata, and give a more descriptive output so the users can understand what happened" which is good feedback that was echoed by other users. This PR was a quick fix to print a friendly message instead of dumping the raw object. This is why I think it's a step forward.
You now suggest not only to print a friendly message but a personalized message per signature, great suggestion but  I thought it belonged in a future PR, mostly because I know we will need to discuss it's design.

itaysk: If you feel that this PR shouldn't be merge as is, I respect that, so let's decide how the per sig usecase looks like. shall we continue here? https://github.com/aquasecurity/tracee/discussions/581

yanivagman: If we don't want to implement the message per signature for the "post-MVP" release - we can merge this.
Otherwise, let's discuss this now and decide on our next steps

simar7: As discussed offline, we have decided to merge this PR with a few changes as I have made but the rest of the work will be carried into a new PR.

PR #584, opened 2021-03-01
tracee-ebpf: remove validator workarounds

yanivagman:
As we moved to minimal kernel 4.18, these WA are not required anymore


grantseltzer:
Seems there's something wrong with the github actions environment, I'm going to investigate that. Otherwise this looks good.

Not that there's typically much sense to it, but what is the verifier issue that this prevented in 4.17?

yanivagman:
> Not that there's typically much sense to it, but what is the verifier issue that this prevented in 4.17?

I don't remember exactly the validator error message, but probably something like "R2 min value is negative..."
It was relevant only for 4.14 (4.15 fixed it), and it was related to the fact that the verifier couldn't handle a size of 0 (with this workaround, the minimal value is 1)

grantseltzer (approved):
LGTM

Issue #253, opened 2020-10-13
e2e sanity test


itaysk:
This issue is for discussing and tracking Tracee's tests posture.
We need to start testing the core functionality of Tracee (at least). This is not an easy task may involve refactoring of the codebase to improve testability (i.e break into modules, DI, etc), think of a way to simulate system events.
Part of this effort should include basic sanity tests across different runtime configurations.
Currently Tracee has minimal tests just for the functions that prints encoded structs in a human readable format.

yanivagman:
As part of the tests, I think we should have a "simple" unit-test program that works as follows:
1. Initialize itself (whatever that will be)
2. Start tracee
3. Generate one event (e.g. call one syscall)
4. Stop tracee
5. Check output matches expectation
6. Check no errors occured.

As we have ~350 events - we need the test to be able to generate each of these events.
After we finish writing this test, we can build on it for generating more complex scenarios (e.g. generating multiple events with different arguments), and test it with different arguments given to tracee CLI (e.g. in container mode, specific process, etc...)

yanivagman:
Another kind of test will be checking for different environments to run tracee.
As small changes to the bpf code of Tracee can make it fail passing the verifier of a given kernel, we need to run a test every time that the bpf code is changed, that will try to run tracee on VMs with different kernel versions installed.
We can then test that no verifier errors occured, meaning tracee successfully initialized itself.

itaysk:
> As we have ~350 events - we need the test to be able to generate each of these events. 

This is the challenge here. A while back I searched for something like this, hoping someone had already done this. Didn't find anything

yanivagman:
For many syscalls, this should be trivial.
There are some syscalls that it will be more challenging to have a VALID call,
but even then, we can supply dummy arguments that the syscall will fail, and yet tracee will trace it and we will be able to validate basic functionality.
This doesn't sound too challenging to me, but a lot of work, as the number of events is big.
We can start from a small set of events, and expand as we go

itaysk:
Do you think we can use this tool? https://github.com/google/syzkaller
I have discovered it a while back but didn't have time to explore yet

yanivagman:
I think that using this tool is like taking a machine gun for a task which requires a small knife.
We need something much simpler.
For each syscall, we need to write one line that calls that specific syscall.
That should be it for a basic test.

itaysk:
[...]

PR #586, opened 2021-03-01
tracee-ebpf: add magic_write event

yanivagman:
Adding a new event named magic_write. This event is triggered when a file header is written (currently set to the first 16 bytes).
This event can then be used to detect when a file of some given type is written (e.g. an ELF file).

To support the addition of this event, a new argument type is introduced: bytes.
When submitting an argument of this type, it is possible to transfer an arbitrary number of bytes as an argument

### DIFF
--- a/tracee-ebpf/tracee/consts.go
+++ b/tracee-ebpf/tracee/consts.go
@@ -152,6 +152,7 @@ const (
 	SchedProcessExitEventID
 	CommitCredsEventID
 	SwitchTaskNSEventID
+	MagicWriteEventID
 	MaxEventID
 )
 
@@ -523,6 +524,7 @@ var EventsIDToEvent = map[int32]EventConfig{
 	SchedProcessExitEventID:    {ID: SchedProcessExitEventID, ID32Bit: sys32undefined, Name: "sched_process_exit", Probes: []probe{{event: "sched:sched_process_exit", attach: rawTracepoint, fn: "tracepoint__sched__sched_process_exit"}}, EssentialEvent: true, Sets: []string{"default", "proc", "proc_life"}},
 	CommitCredsEventID:         {ID: CommitCredsEventID, ID32Bit: sys32undefined, Name: "commit_creds", Probes: []probe{{event: "commit_creds", attach: kprobe, fn: "trace_commit_creds"}}, Sets: []string{}},
 	SwitchTaskNSEventID:        {ID: SwitchTaskNSEventID, ID32Bit: sys32undefined, Name: "switch_task_ns", Probes: []probe{{event: "switch_task_namespaces", attach: kprobe, fn: "trace_switch_task_namespaces"}}, Sets: []string{}},
+	MagicWriteEventID:          {ID: MagicWriteEventID, ID32Bit: sys32undefined, Name: "magic_write", Probes: []probe{}, Sets: []string{}},
 }
 
 // EventsIDToParams is list of the parameters (name and type) used by the events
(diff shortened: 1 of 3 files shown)


grantseltzer:
If I understand correctly, in the below output the bytes `127 69 76 70 2 1 1 0 0 0 0 0 0 0 0 0` correspond to the magic number of an ELF (I cp'd the tracee binary). Can we add an arg-printer function that will translate these bytes to just say "we know it was an ELF" ? Is there any other information we get besides the file type from these bytes?  This possibly would make more sense to be in tracee-rules.

```
905.001378     0      NetworkManager   480     480     109              magic_write          pathname: /var/lib/NetworkManager/timestamps.LUZ6Y0, bytes: [91 116 105 109 101 115 116 97 109 112 115 93 10 97 98 98]
911.313399     1000   cp               1948    1948    131072           magic_write          pathname: /home/vagrant/go/src/github.com/aquasecurity/tracee/tracee-ebpf/dist/tracee-ebpf-magic1, bytes: [127 69 76 70 2 1 1 0 0 0 0 0 0 0 0 0]
```

yanivagman:
Yes, these bytes represnt an ELF file. I thought that a byte array in go would print as hex with %v, but that wasn't the case :-(
Anyways, like you wrote, it is tracee-rules' responsibility to decode the bytes and understand the file type out of it. This connects to the discussion we had today - tracee-ebpf should generate raw arguments for tracee-rules, which can then interpret them as it wants. We can add an arg printer for that in a future PR, which can then be used by tracee-rules or tracee-ebpf (for users who invoke it directly)

grantseltzer:
If you want the bytes to just be in hex you can use %x or %X, if the bytes combine to form a 4 or 8 byte datatype you can use the encoding package like this for example:

```
value := binary.LittleEndian.Uint64(data) 
```

grantseltzer (approved):
I assume you want to change the byte printing verb (%v -> %x or something like that) but otherwise this is great!

yanivagman:
I know that %x prints in hex, but I can't just change it as it is being used for all argument types:
https://github.com/aquasecurity/tracee/blob/main/tracee-ebpf/tracee/printer.go#L149

simar7:
As earlier discussed, I think there's some value in adding a selftest like the following:

```
func Test_MagicWrite(t *testing.T) {
	var out bytes.Buffer
	cmd := exec.Command("dist/tracee-ebpf", "--trace", "event=magic_write")
	cmd.Stdout = &out

	done := make(chan bool, 1)

	go func(cmd *exec.Cmd) {
		fmt.Println("running: ", cmd.String())
		go assert.NoError(t, cmd.Run())
		<-done
	}(cmd)
	time.Sleep(time.Second * 1) // wait for tracee init

	// create a temp dir for testing
	d, err := ioutil.TempDir("", "Test_MagicWrite-dir-*")
	require.NoError(t, err)

	// cp a file to trigger
	f, err := os.CreateTemp(d, "Test_MagicWrite-file-*")
	require.NoError(t, err)
	defer func() {
		os.Remove(d)
	}()

	f.WriteString(`foo.bar.baz`)
	f.Close()

	cpCmd := exec.Command("cp", f.Name(), filepath.Join(d+filepath.Base(f.Name())+"-new"))
	fmt.Println("executing: ", cpCmd.String())
	cpCmd.Stdout = os.Stdout
	assert.NoError(t, cpCmd.Run())

	done <- true
	assert.Contains(t, out.String(), "bytes: [102 111 111 46 98 97 114 46 98 97 122]")
}
```

Doesn't have to be part of this PR, just something to think of for the future.

yanivagman:
@simar7 I agree we should add tests to Tracee's events, and this event is no exception.
Other than this new event, we should also add simple selftests to all system calls, vfs_write and all other events.
We should probably add some test framework code first, which will be common to all of the events like starting tracee, triggering the event, reading the output of tracee and verifying that we got the expected results. Also see my first comment in #253, which subject was changed to e2e tests (but we should probably also open a new issue for event self tests)

PR #587, opened 2021-03-01
tracee-ebpf: fix docker builder

yanivagman:


simar7 (approved):
thanks for figuring this out

Issue #434, opened 2021-01-22
use golang's file embedding (1.16)


itaysk:
in order to embed the bpf code in tracee binary, we use go linker variable injection: https://github.com/aquasecurity/tracee/blob/main/Makefile#L52
This was always a temporary hack, and now we have a better solution:
go 1.16 bring native support for embedding static files, and we should switch to that. 1.16 should be out early Feb so we can start preparing the PR and commit it when 1.16 is stable.

This PR will need to also move tracee to 1.16 of course.

itaysk:
1.16 is not released

simar7:
I briefly had a look into this, just thinking out loud:

1. Currently we support both: 
  a) using an existing BPF object as-is if found
  b) building BPF object on-the-fly if not.
2. `go embed` assumes that a) is the case for the `embed` directive to work

Our b) approach doesn't fit in so nicely here. `embed` assumes that the file being embedded will be present upon compile time. If not compilation fails. 

While this may seem as a step back in terms of being able to support both a) and b), it actually would simplify a lot of the existing code, if we were to remove b) altogether and only support a). I'm not aware of the historical context of why b) is supported today but if it's not something that yields us too much value, I can see deprecating in favour of simplicity.

Thoughts?

yanivagman:
> Our b) approach doesn't fit in so nicely here. embed assumes that the file being embedded will be present upon compile time. If not compilation fails.

I'm not sure I understant this sentence. Why can't we embed the bpf.c file and compile on the fly? Is there a problem with that approach?

> While this may seem as a step back in terms of being able to support both a) and b), it actually would simplify a lot of the existing code, if we were to remove b) altogether and only support a). I'm not aware of the historical context of why b) is supported today but if it's not something that yields us too much value, I can see deprecating in favour of simplicity.

This is indeed a step back from tracee-ebpf perspective. To support a standalone binary of tracee-ebpf, it is required to support on the fly compilation. I think it is usefull to have a standalone binary like we do have today, and unless we will embed all bpf objects to the binary, compilation on the fly is required

itaysk:
@simar7 am I correct to understand that you interpret this issue as to embed the ebpf object file? if so, it's not the case, this is about embedding the source code so that the file can be compiled on the target machine.

itaysk:
also embed helpers.rego 
and perhaps some gotemplates that we'd want to ship with tracee

simar7:
[...]

PR #589, opened 2021-03-03
feat: Bump up to go1.16

simar7:
This is in preparation for https://github.com/aquasecurity/tracee/issues/434

### DIFF
--- a/.github/workflows/pr-rules.yaml
+++ b/.github/workflows/pr-rules.yaml
@@ -15,7 +15,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v1
         with:
-          go-version: 1.15
+          go-version: 1.16
       - name: Lint
         run: |
           if test -z "$(gofmt -l .)"; then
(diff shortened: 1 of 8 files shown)


.github/workflows/pr-rules.yaml
```diff
@@ -15,7 +15,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v1
         with:
-          go-version: 1.15
+          go-version: 1.16
```
simar7, line 18: I wonder if we should just eliminate this setup since Ubuntu 20.04 Github Actions runner includes this by default anyway: https://github.com/actions/virtual-environments/blob/ubuntu20/20210219.1/images/linux/Ubuntu2004-README.md#go

We actually current don't have it for tracee-ebpf action: https://github.com/aquasecurity/tracee/blob/main/.github/workflows/pr-ebpf.yaml#L17-L25

itaysk: we don't have it in tracee-ebpf because it runs in a container.
personally I think it's better to be specific which version we are using to avoid future bugs

itaysk (approved):

PR #591, opened 2021-03-03
Cleanup ringbuffer stop logic

grantseltzer:
fix: rb.stopped should be set in the Stop method

This cleans up the design of a solution to an issue where Stop()
 could be called after the sole consumer of the rb.stop channel
 has closed but rb.stopped has not yet been set. This removes
the need to have a non-blocking channel write with `select` as
it turns out we really don't need this channel at all.


yanivagman (approved):
Much cleaner!

PR #596, opened 2021-03-04
feat: Use //go:embed to bundle bpf code

simar7:
Fixes: https://github.com/aquasecurity/tracee/issues/434

Requires go 1.16 or latest

### DIFF
--- a/tracee-ebpf/main.go
+++ b/tracee-ebpf/main.go
@@ -1,9 +1,10 @@
 package main
 
 import (
+	_ "embed"
+
 	"archive/tar"
 	"compress/gzip"
-	"encoding/base64"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -26,6 +27,7 @@ var traceeInstallPath string
 var buildPolicy string
 
 // These vars are supposed to be injected at build time
+//go:embed dist/tracee.bpf.tar.gz
 var bpfBundleInjected string
 var version string
 
@@ -978,13 +980,12 @@ func getBPFObject() (string, error) {
 	return bpfObjFilePath, nil
 }
 
-// unpackBPFBundle unpacks the bundle (tar(gzip(b64))) into the provided directory
+// unpackBPFBundle unpacks the bundle (tar(gzip)) into the provided directory
 func unpackBPFBundle(dir string) error {
 	if bpfBundleInjected == "" {
 		return fmt.Errorf("missing embedded data")
 	}
-	b64Reader := base64.NewDecoder(base64.RawStdEncoding, strings.NewReader(bpfBundleInjected))
-	gzReader, err := gzip.NewReader(b64Reader)
+	gzReader, err := gzip.NewReader(strings.NewReader(bpfBundleInjected))
 	if err != nil {
 		return err
 	}
--- a/tracee-ebpf/main.go
+++ b/tracee-ebpf/main.go
@@ -1,3 +1,5 @@
+// +build go1.16
+
 package main
 
 import (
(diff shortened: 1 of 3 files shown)


simar7:
Note: I have left templates out of this PR. I feel it's better for them to exist on their own, as they're nice-to-haves but not essentials. Thereby keeping tracee-rules compact.

tracee-ebpf/main.go
```diff
@@ -1,9 +1,12 @@
+// +build go1.16
```
grantseltzer, line 1: I suppose it's not a huge deal but are we ok with requiring go 1.16? There's probably a way to use embed only if go1.16 but not otherwise, like having the embed directive in a separate file

tracee-ebpf/main.go
```diff
@@ -26,6 +29,7 @@ var traceeInstallPath string
 var buildPolicy string
 
 // These vars are supposed to be injected at build time
+//go:embed dist/tracee.bpf.tar.gz
 var bpfBundleInjected string
```
itaysk, line 33: I don't think string is the best type here. If we want to keep the legacy workflow, we should change to `[]byte`, but I perhaps it'll be better to embed as `FS` then we can eliminate that entire workflow from the makefile and also the unpackBPFBundle code. WDYT?

tracee-ebpf/main.go
simar7: Yeah I guess it's safe to leave it out. This is what happens without it (on go 1.15 or earlier):

```
GOOS=linux GOARCH=amd64 CC=clang CGO_CFLAGS="-I /tracee/tracee-ebpf/dist/libbpf/usr/include" CGO_LDFLAGS="/tracee/tracee-ebpf/dist/libbpf/libbpf.a" go build -v -o dist/tracee-ebpf \
-ldflags "-X main.version=v0.5.0-20-gd58ae46"
go: downloading github.com/syndtr/gocapability v0.0.0-20180916011248-d98352740cb2
go: downloading github.com/urfave/cli/v2 v2.1.1
go: downloading github.com/aquasecurity/tracee/libbpfgo v0.0.0-20210115081842-487d1e44fcda
go: downloading github.com/cpuguy83/go-md2man/v2 v2.0.0-20190314233015-f79a8a8ca69d
go: downloading github.com/russross/blackfriday/v2 v2.0.1
go: downloading github.com/shurcooL/sanitized_anchor_name v1.0.0
main.go:4:2: package embed is not in GOROOT (/usr/local/go/src/embed)
```

Which is fine, it's still a compile time error at the end of the day.

tracee-ebpf/main.go
simar7: It's an interesting point, I need to investigate it further.

simar7: Thanks for the idea – I implemented it here: https://github.com/aquasecurity/tracee/pull/596/commits/abb0987bd91d35eb73a09f26c3ecf3718e332318 this trimmed some fat.

itaysk (approved):
LGTM! thanks
Just a question, I see that the source now has references to `dist/tracee.bpf/` which isn't ideal. This presumably is configurable at build time by the user in which case embedding would be broken. also it creates a dependency between the source and the makefile (until now it was only in the other direction). It's not a big deal, but it would be nice if we would find an elegant solution for that in a future PR

simar7:
> LGTM! thanks
> Just a question, I see that the source now has references to `dist/tracee.bpf/` which isn't ideal. This presumably is configurable at build time by the user in which case embedding would be broken. also it creates a dependency between the source and the makefile (until now it was only in the other direction). It's not a big deal, but it would be nice if we would find an elegant solution for that in a future PR

Yeah I agree it's a bit cumbersome. In fact after trying out embed, I'm not fully onboard it is the right approach from a language semantic point of view (similar concerns for go generate). Such behaviours are harder to test and may still fail in ways we don't expect. But it's a step in the right direction which hopefully gets better as the language evolves.

Issue #599, opened 2021-03-06
move go helpers to tracee-rules


itaysk:
currently the go signature helpers are contained in the signature's package, which means they are inaccessible to other signature authors. we need to move them into tracee-rules and make them available to all

Issue #538, opened 2021-02-11
document signature helpers


itaysk:
document the fact that we have helpers, why/when/how to use them. both rego and go

simar7:
Could you add what's the acceptance criteria here? Is this to document `helpers.rego`?

itaysk:
document the fact that we have helpers, why/when/how to use them. both rego and go

itaysk:
related: https://github.com/aquasecurity/tracee/issues/599

PR #601, opened 2021-03-08
fix: move and document the signature helpers 

grantseltzer:
This moves both the rego and go helper functions into their own package, updates usage and Makefile accordingly. (#599) 

It also adds notes in the README about them (#538).

### DIFF
--- a/tracee-rules/Readme.md
+++ b/tracee-rules/Readme.md
@@ -52,7 +52,7 @@ Create a `.rego` file in the rules directory that has the following Rego Rules (
 2. `tracee_selected_events`: A *set* rule that defines the event selectors.
 3. `tracee_match`: A *boolean* or a *document* rule that defines the logic of the signature. If bool is "returned", a true evaluation will generate a Finding with no data if document is "returned", any non-empty evaluation will generate a Finding with the returned document as the Finding's "Data".
 
-See [example2.rego](/tracee-rules/signatures/rego/examples/example2.rego) and [example1.rego](/tracee-rules/signatures/rego/examples/example1.rego) for example Rego signatures.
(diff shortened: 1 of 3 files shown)


grantseltzer:
Couple of issues I need to fix here:

- `findRegoSigs(dir string)` breaks because it's looking for `helpers.rego` at an essentially hardcoded path. I feel this should not be the case, or at the very least should be an optional thing to look for (helpers don't have to be used).
- A linter error

tracee-rules/signature_test.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 func Test_getSignatures(t *testing.T) {
-	sigs, err := getSignatures("signatures/rego", []string{"TRC-2"})
+	sigs, err := getSignatures("./dist/rules", []string{"TRC-2"})
```
simar7, line 13: Passing `signatures/rego` makes the use of existing rego signatures as inputs, effectively using them as golden files for this test.

Was there another motivation to change it to `dist/rules`?

itaysk: I think it's ok to leave the helpers under the signatures directory, after all they are signature helpers. that will also address this concern I believe

itaysk (commented):
I do think the rego helpers should move to a dedicated package (just the name, not the directory). in rego conventions that would be like `package tracee.helpers`

itaysk:
also, probably worth adding code comments to document the helper functions themselves. but this can be a separate PR if you'd like

grantseltzer:
> I do think the rego helpers should move to a dedicated package (like you already did before) (just the name, not the directory). in rego conventions that would be like `package tracee.helpers`

How is this different from how it is now?

> also, probably worth adding code comments to document the helper functions themselves. but this can be a separate PR if you'd like

On it

itaysk:
> How is this different from how it is now?

my bad :) got confused for a moment

itaysk (approved):

PR #603, opened 2021-03-08
feat: add libbpfgo selftests to github actions

grantseltzer:


grantseltzer:
Perhaps premature to be opening this but it's failing because the test is running on an outdated kernel. Since we need to have selftests run on various kernels/distros we should have a test in the selftest run script for this, will add that to this PR.

grantseltzer:
Alright I changed it so that insufficient kernel version doesn't cause a failure.

grantseltzer:
There's a couple of changes in this that are just gofmt changes from a previous PR

grantseltzer:
@itaysk ptal

itaysk (approved):

itaysk:
LGTM

I'd prefer if I was not always the gatekeeper for merging things. I mean it's always OK to ping a specific maintainer for review, but in most cases I hope maintainers can unblock each other

Issue #594, opened 2021-03-04
Add tests common framework


yanivagman:
Issue #253 subject was changed to e2e tests as "adding tests" was too large scope.
I think we should add some common framework code that will be used by unit tests, and will enable us to add selftests to new and existing events on the go. This common framework code should probably do the following:

Initialize itself (whatever that will be)
For some_event := range Tracee's events:
a. Start tracee with -e "some_event" -p "self_pid"
b. Generate some_event (e.g. call the syscall that generates it)
c. Stop tracee
d. Check output matches expectation
e. Check no errors occured.

After having this common code, we can write simple selftests to many of Tracee's events. To trigger an event, it should be enough to give dummy arguments to the event and check that we traced the event correctly (the dummy arguments should be traced even if the event returned an error - for example in a failed syscall)

simar7:
After discussing this on the meeting today a few more interesting points were raised:

1. Using Tracee as a library to instantiate eBPF code once and use many times across various test cases
2. Parallelize test inputs
3. First cover those test inputs that don't cause a side effect on the system/setup (e.g. mounting fs calls)
4. Potentially take a look into https://github.com/testcontainers/testcontainers-go if it can help with isolation

PR #606, opened 2021-03-12
Add basic integration test framework

simar7:
This PR introduces a basic framework that loads Tracee as a binary akin to how an external user would use in real life scenario.

It covers a set of basic cases. Writing a test for each input is better suited to a different [kind](https://go.googlesource.com/proposal/+/master/design/draft-fuzzing.md) of testing or the options discussed [here](https://github.com/aquasecurity/tracee/issues/253).

For the time-being until we figure out the CI situation, these tests aren't being wired up to run on GitHub Actions as they involve loading BPF code into the kernel. 

I'm open to adding more cases here. 

Fixes https://github.com/aquasecurity/tracee/issues/594


integration/integration_test.go
```diff
@@ -0,0 +1,230 @@
...
+
+	for _, tc := range testCases {
+		tc := tc
+		t.Run(tc.name, func(st *testing.T) {
+			st.Parallel()
+
+			var gotOutput bytes.Buffer
+			done := make(chan bool, 1)
+			go loadTracee(st, &gotOutput, done, tc.args...)
+			time.Sleep(time.Second) // wait for tracee init
```
itaysk, line 224: the more accurate way is to look for the file `$outputdir/tracee.pid`

integration/pid.tmpl
```diff
@@ -0,0 +1 @@
+{{ .ProcessID }}
```
itaysk, line 1: just an idea, if you don't like it it's fine: instead of creating files for each template, which is not going to be manageable if we have lots of tests, we could just create a tmp file with the test-specific template within the test the needs it.

integration/Makefile
```diff
@@ -0,0 +1,2 @@
+test:
```
itaysk, line 1: I would suggest using the same pattern that we already use for tests, where GH actions calls the Tracee makefile, which calls into tracee-rules makefile, which knows how to run the tests. some thoughts:
1. this tests tracee-ebpf specifically, it doesn't test "Tracee" so it should be under the tracee-ebpf dir.
2. tracee-ebpf already has a tests directory. I would put `integration` there. (nit is this integration test or e2e test, given the mention scoping of tracee-ebpf)
3. if it's under tracee-ebpf, is there a reason why it needs it's own makefile, or could it be added to the tracee-rule's makefile.

integration/integration_test.go
```diff
@@ -0,0 +1,230 @@
...
+	"time"
+
+	ps "github.com/mitchellh/go-ps"
+	"github.com/onsi/gomega/gexec"
+	"github.com/stretchr/testify/require"
+
+	"github.com/stretchr/testify/assert"
+)
+
+const TraceeBinaryPath = "../tracee-ebpf/dist/tracee-ebpf"
```
itaysk, line 23: another case of go code creating a dependency on the makefile (I called out a similar one in a previous PR), could we manage this better? perhaps an env var that the makefile sets?

integration/integration_test.go
```diff
@@ -0,0 +1,230 @@
...
+	cpCmd.Stdout = os.Stdout
+	assert.NoError(t, cpCmd.Run())
+
+	// check tracee output
+	assert.Contains(t, gotOutput.String(), `bytes: [102 111 111 46 98 97 114 46 98 97 122]`)
+}
+
+// execute a ls command
+func checkExeccommand(t *testing.T, gotOutput *bytes.Buffer) {
+	_, _ = exec.Command("ls").CombinedOutput()
```
itaysk, line 82: do we need a sleep between the action and observing it's side affect? (same for other tests)

itaysk: currently the flow is 2 part: 1 - start tracee 2- exercise command an check output. 
a suggestion, to make it 3 phase:
1 - start tracee 2- exercise command and stop tracee 3 - check output.
this will address my previous comment and also possible ease on the test overhead. It might be a premature optimization though

itaysk (commented):
this is great, thank you

itaysk:
I think I'm going to remove the bats test that I added before. if the team prefers go tests also for these kind of cli tests, there's no reason to maintain 2 platforms

grantseltzer:
I do agree with Itay on this handling some things that should be handled in Makefiles, but regardless I think this is a great step in the right direction for the project.

There's some fragility with the way you're checking output, this test fails on my system:

```
CONT  Test_Events/trace_filesystem_events_from_comm_ls
running:  ../tracee-ebpf/dist/tracee-ebpf --trace s=fs --trace comm=ls --output gotemplate=eventName.tmpl
    integration_test.go:162:
        	Error Trace:	integration_test.go:162
        	            				integration_test.go:226
        	Error:      	[]string{"access", "openat", "fstat", "close", "read", "pread64", "statfs", "ioctl", "statx", "getdents64", "write"} does not contain "lgetxattr"
        	Test:       	Test_Events/trace_filesystem_events_from_comm_ls
```

I suppose somewhere on my system (shell, kernel, or lib version) uses a different system call or something of that nature. This again relates to our need for infrastructure where we can run github actions against lots of different kernel/distro versions. Just something to note for now.

integration/integration_test.go
simar7: That's interesting. I had a look at it but found an improvement we can make before we can use it. Today the code looks like was written to only run a single instance of Tracee at a time. https://github.com/aquasecurity/tracee/blob/228c6d329e86dd8f62bb5fe48761e78b67dafb39/tracee-ebpf/tracee/tracee.go#L328-L331

But this has changed now so we will need to support creating a unique pid file for each run. `WriteFile` truncates the file if it exists and rewrites it. This prevents us from relying on it as a readiness check across multiple goroutines. But I do agree with you this is the better approach.

simar7: Filed an issue here to track it https://github.com/aquasecurity/tracee/issues/607

integration/pid.tmpl
simar7: Good point, this was on my mind as well. I updated here https://github.com/aquasecurity/tracee/pull/606/commits/b98c571f2f2af083cec43dc7a301ac556d40077d

integration/integration_test.go
simar7: I haven't observed any noticeable delay/lag in executing a command and checking its output. The output buffer is shared between tracee and the check function. 

Having said that maybe it will become a problem if we have more test cases but I would defer solving that until it becomes a problem.

simar7:
> I do agree with Itay on this handling some things that should be handled in Makefiles, but regardless I think this is a great step in the right direction for the project.
> 
> There's some fragility with the way you're checking output, this test fails on my system:
> 
> ```
> CONT  Test_Events/trace_filesystem_events_from_comm_ls
> running:  ../tracee-ebpf/dist/tracee-ebpf --trace s=fs --trace comm=ls --output gotemplate=eventName.tmpl
>     integration_test.go:162:
>         	Error Trace:	integration_test.go:162
>         	            				integration_test.go:226
>         	Error:      	[]string{"access", "openat", "fstat", "close", "read", "pread64", "statfs", "ioctl", "statx", "getdents64", "write"} does not contain "lgetxattr"
>         	Test:       	Test_Events/trace_filesystem_events_from_comm_ls
> ```
> 
> I suppose somewhere on my system (shell, kernel, or lib version) uses a different system call or something of that nature. This again relates to our need for infrastructure where we can run github actions against lots of different kernel/distro versions. Just something to note for now.
> 

Thanks this is a good point. Yeah actually this assertion is incomplete. There are many more sys calls that belong to the `fs` set. https://github.com/aquasecurity/tracee/blob/228c6d329e86dd8f62bb5fe48761e78b67dafb39/tracee-ebpf/tracee/consts.go#L160-L528 that we will need to cover. I will find a better way to handle this.

> At the end of the day I think integration tests matter most for tracee-rules. It doesn't really matter if the output events contain `execve` or `execveat`, as long as detections are working properly.

I agree but as we discussed on a call in the past, I think it ultimately depends on what our user interface is going to be for people to interface with tracee. Some might want just use it as a detection tool like you mentioned in which case adding integration tests for tracee-rules is the right place to start. Until we can finalize, better to have some coverage in the meantime.

integration/integration_test.go
simar7: Updated https://github.com/aquasecurity/tracee/pull/606/commits/3cd02b3b20fc88d6a921bb54ee47b25b60f0d508

integration/Makefile
simar7: Updated https://github.com/aquasecurity/tracee/pull/606/commits/a630a561913db092b0a10602969048469576431b

itaysk: you moved it into tracee-ebpf rather than into tracee-ebpf/test as suggested. don't you think it belongs in the test dir?

tracee-ebpf/Makefile
```diff
@@ -124,6 +124,10 @@ test: $(DOCKER_BUILDER)
 	$(call docker_builder_make,$@)
 endif
 
+.PHONY: test-integration
+test-integration:
+	$(go_env) TRC_BIN=../$(OUT_BIN) go test -exec 'sudo -E' -v integration/*_test.go
```
itaysk, line 129: why is the `../` prefix required? I think it should be `TRC_BIN=$(OUT_BIN)` no?

itaysk: I suspect this has to do with running from the root makefile, in this case probably that makefile should set TRB_BIN to the correct location

tracee-ebpf/Makefile
```diff
@@ -124,6 +124,10 @@ test: $(DOCKER_BUILDER)
 	$(call docker_builder_make,$@)
 endif
 
+.PHONY: test-integration
+test-integration:
```
itaysk, line 128: missing prerequisite (the binary)

tracee-ebpf/Makefile
simar7: Actually I believe this path is relative to the integration module not to the Makefile. That's why going up a directory is needed. Go tests run relative to the directory they are in. 

> in this case probably that makefile should set TRB_BIN to the correct location

Isn't that what the Makefile is doing now? Or did you have something else in mind.

tracee-ebpf/Makefile
simar7: thanks updated https://github.com/aquasecurity/tracee/pull/606/commits/4ed2477e3cb3b7ecb1aa9f7f8ccb8532b73f28cc

integration/Makefile
simar7: I'm fine either way. I just took the conventional approach of placing it under `integration/`. One example is Trivy https://github.com/aquasecurity/trivy/tree/main/integration I reworked  as you asked: https://github.com/aquasecurity/tracee/pull/606/commits/efd2dcff53a5da95b8eedca40cd1cf65215170bd

itaysk (approved):

tracee-ebpf/test/integration/integration_test.go
```diff
@@ -0,0 +1,256 @@
...
+	traceePid := getPidByName(t, "tracee")
+
+	// run a command
+	_, _ = exec.Command("ls").CombinedOutput()
+
+	// output should only have events with pids greater (newer) than tracee
+	pids := strings.Split(strings.TrimSpace(gotOutput.String()), "\n")
+	for _, p := range pids {
+		pid, _ := strconv.Atoi(p)
+		assert.Greater(t, pid, traceePid)
```
yanivagman, line 110: Although rare, but this may fail if the pids in the system are wrapped.

tracee-ebpf/test/integration/integration_test.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+// check for filesystem set when ls is invoked
+func checkSetFs(t *testing.T, gotOutput *bytes.Buffer) {
+	_, _ = exec.Command("ls").CombinedOutput()
+
+	// check output length
+	require.NotEmpty(t, gotOutput.String())
+
+	// output should only have events with event name of execve
```
yanivagman, line 167: Comment should probably be related to set=fs

yanivagman (commented):
Thanks @simar7 this is great!

tracee-ebpf/test/integration/integration_test.go
simar7: Oh I'm not familiar with the concept of pids being wrapped, could you explain a little more?

yanivagman: Depends on how the system is configured, there is a max pid number, after which wrapping occur.
You can see the value by:
`cat /proc/sys/kernel/pid_max`
On 64 bit systems, this number will usually be 4194304

tracee-ebpf/test/integration/integration_test.go
simar7: Thanks fixed.

tracee-ebpf/test/integration/integration_test.go
simar7: > > I do agree with Itay on this handling some things that should be handled in Makefiles, but regardless I think this is a great step in the right direction for the project.
> > There's some fragility with the way you're checking output, this test fails on my system:
> > ```
> > CONT  Test_Events/trace_filesystem_events_from_comm_ls
> > running:  ../tracee-ebpf/dist/tracee-ebpf --trace s=fs --trace comm=ls --output gotemplate=eventName.tmpl
> >     integration_test.go:162:
> >         	Error Trace:	integration_test.go:162
> >         	            				integration_test.go:226
> >         	Error:      	[]string{"access", "openat", "fstat", "close", "read", "pread64", "statfs", "ioctl", "statx", "getdents64", "write"} does not contain "lgetxattr"
> >         	Test:       	Test_Events/trace_filesystem_events_from_comm_ls
> > ```
> > 
> > 
> > I suppose somewhere on my system (shell, kernel, or lib version) uses a different system call or something of that nature. This again relates to our need for infrastructure where we can run github actions against lots of different kernel/distro versions. Just something to note for now.
> 
> Thanks this is a good point. Yeah actually this assertion is incomplete. There are many more sys calls that belong to the `fs` set.
> 
> https://github.com/aquasecurity/tracee/blob/228c6d329e86dd8f62bb5fe48761e78b67dafb39/tracee-ebpf/tracee/consts.go#L160-L528
> 
> that we will need to cover. I will find a better way to handle this.
> > At the end of the day I think integration tests matter most for tracee-rules. It doesn't really matter if the output events contain `execve` or `execveat`, as long as detections are working properly.
> 
> I agree but as we discussed on a call in the past, I think it ultimately depends on what our user interface is going to be for people to interface with tracee. Some might want just use it as a detection tool like you mentioned in which case adding integration tests for tracee-rules is the right place to start. Until we can finalize, better to have some coverage in the meantime.

I added a helper func here: https://github.com/aquasecurity/tracee/pull/606/commits/2dabbfa227c78f23d5224f2a100c269144617748#diff-02603f54659cbb5a36b20ea4235fc2b31a45e179c0e3ebdbfe84822b2403d81eR205-R214

Issue #592, opened 2021-03-03
remove falcosidekick from container


itaysk:
the tracee container includes falcosidekick, it should be taken out of tracee but documented how to use tracee with falcosidekick as a second container

itaysk:
as suggested here https://github.com/aquasecurity/tracee/pull/582#discussion_r586771393 also remove the falcosidekick struct from the webhook code, and replace it with a gotemplate

PR #613, opened 2021-03-13
remove falcosidekick from container

itaysk:
fix: #592 

Also removed entrypoint tests given https://github.com/aquasecurity/tracee/pull/606#issuecomment-797523281 and since the #605 and now this PR dramatically simplified the entrypoint logic


Readme.md
```diff
@@ -59,11 +59,15 @@ LD_PRELOAD | Usage of LD_PRELOAD to allow hooks on process | "linux", "container
...
+--- | --- | ---
+`--webhook-url` | The webhook URL | `--webhook-url http://my.webhook/endpoint`
+`--webhook-template` | Path to Go-template that formats the payload to send. Tracee's [Finding](https://github.com/aquasecurity/tracee/blob/28fbc66be8c9f3efa53f617a654cafe7421e8c70/tracee-rules/types/types.go#L46-L50) type is available to use within the template | `--webhook-template /path/to/my.tmpl` <br> See template examples [here](tracee-rules/templates/).
+`--webhook-content-type` | If present, will set the Content-Type HTTP header to match the provided template | `--webhook-content-type application/json`
 
-A complete reference of falco-sidekick's configuration format is available [here](https://github.com/falcosecurity/falcosidekick/blob/master/config_example.yaml).
+A popular webhook server that can be used with Tracee is [falcosidekick](https://github.com/falcosecurity/falcosidekick), which can send detection events into other systems of your choosing (for example Slack, Teams, Datadog, Prometheus, Email, Elasticsearch, PagerDuty, and many more). For more information on falcosidekick see [here](https://github.com/falcosecurity/falcosidekick/blob/master/config_example.yaml).
```
yanivagman, line 70: Add postee as well?

entrypoint.sh
```diff
@@ -1,32 +1,58 @@
...
+				test $# -lt 2 && die "missing value for argument '$1'." 1
+				arg_webhook_url="$2"
+				shift
+				;;
+			--webhook-content-type)
+				test $# -lt 2 && die "missing value for argument '$1'." 1
+				arg_webhook_content_type="$2"
+				shift
+				;;
+			*)
```
yanivagman, line 44: What if both --rules and --webhook-url are given? using **-lt 2** will not work in that case.
As all of these arguments are handled by tracee-rules (only tracee-ebpf is not) - maybe we should remove this entrypoint and just call tracee-rules, and if **trace** is given - call tracee-ebpf

itaysk: > What if both --rules and --webhook-url are given? using -lt 2 will not work in that case.

why would it not work? 2<4. not sure what's the issue you are seeing?

itaysk: > remove this entrypoint and just call tracee-rules

who will orchestrate passing events from tracee-ebpf to tracee-rules (the pipe)? the entrypoint is needed for that

> and if trace is given - call tracee-ebpf

where is this `if` going to be if not in entrypoint?

yanivagman: > why would it not work? 2<4. not sure what's the issue you are seeing?

It will work, but miss the point of argument verification, for example, in the follwing case:
`tracee --rules --webhook-url http://bla`
There is a missing value for **rules** in that case.

> who will orchestrate passing events from tracee-ebpf to tracee-rules (the pipe)? the entrypoint is needed for that

tracee-rules can - by creating a pipe, and calling fork-execve. It can also create a pipe which is not stdin based, allowing us to add more sources in the future.

> where is this if going to be if not in entrypoint?

tracee-rules can have a new flag called "trace". If given - execve tracee-ebpf (without forking first)

itaysk: ok so you're suggesting to make tracee-rules the entrypoint.. this will turn tracee-rules to be not just a rule engine, but the frontend for tracee. we discussed it before and IIRC you didn't like that idea. anyway, I think it's beyond the scope of this PR

itaysk: > but miss the point of argument verification

I'm wasn't trying to do argument verification, just defensive programming. This also wasn't handled before this PR I think. But I'll change it to verify the arg

yanivagman: I think that argument verification should be left to tracee-rules - no need to do this here as well.

> ok so you're suggesting to make tracee-rules the entrypoint.. this will turn tracee-rules to be not just a rule engine, but the frontend for tracee. we discussed it before and IIRC you didn't like that idea. anyway, I think it's beyond the scope of this PR

When we discussed it in the past, I suggested to have 'trace' flag as well as 'rules' flag. In the end, only 'trace' flag was implemented, and tracee-rules became the de-facto frontend for tracee.  I think that we should first answer the question - what is the role of the entrypoint? (especially now, after we removed mounting debugfs). 
I don't think that the role of the entrypoint should be argument verification - this should be done either by tracee-rules or tracee-ebpf.
This leaves us with the role of creating a pipe between tracee-rules and whatever source we will give it in the future. But this role can be implemented by tracee-rules, isn't it?

yanivagman: For this PR, I suggest only removing the falcosidekick part from the entrypoint

itaysk: > for this PR, I suggest only removing the falcosidekick part from the entrypoint

part of the motivation to remove it from the container is that we're now able to call it or others using the new webhook flags. If we remove falcosidekick from the container we should also add support in those flags, which is what I did

itaysk: > what is the role of the entrypoint?

since we chose to build tracee-ebpf and tracee-rules as independent projects, the entrypoint is what ties everything together. the fact that it's now simpler is good, we should be happy about simplifying it.
if we move that functionality into tracee-rules, it is not longer an independent rules engine, we might as well rename it to "tracee" like originally suggested. 

as a suggestion - since currently the entrypoint's ux is almost the same as tracee-rules, I could just pass all flags to tracee-rules. but I'm not sure that going forward that'll be the best solution

itaysk: I pushed a new versions that you might like better, where tracee-rules handles most of the args directly

Readme.md
itaysk: unfortunately postee is not as simple as falcosidekick. someone needs to try it with Tracee and document it appropriately.

yanivagman (approved):

PR #615, opened 2021-03-14
Fix some of the toplevel "make clean" errors.

icarus-sparry:
Fix "make clean".
 - Use '-C' rather than '-c' to enter subdirectory
 - Check for existence of files before reading from them, fixes empty
 args to docker rmi.

Document Gnu Make 4.3 or better is required, Makefiles uses grouped targets.

Signed-off-by: Icarus Sparry <Icarus Sparry tracee@icarus.freeuk.com>

### DIFF
--- a/tracee-ebpf/Makefile
+++ b/tracee-ebpf/Makefile
@@ -145,10 +145,13 @@ mostlyclean:
 
 .PHONY: clean
 clean:
-	-$(CMD_DOCKER) rmi $(file < $(docker_builder_file))
+	-FILE="$(docker_builder_file)" ; \
+	if [ -r "$$FILE" ] ; then \
+		$(CMD_DOCKER) rmi "$$(< $$FILE)" ; \
+	fi
 	-rm -rf dist $(OUT_DIR)
 	$(MAKE) -C $(LIBBPF_SRC) clean
-	
+
 check_%:
 	@command -v $* >/dev/null || (echo "missing required tool $*" ; false)
 
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,8 @@ $(OUT_DIR)/tracee.bpf.%.o: bpf ;
 bpf: tracee-ebpf | $(OUT_DIR)
 		$(MAKE) -C $< mostlyclean $@ OUT_DIR=dist && $(CMD_CP) $</dist/tracee.bpf.*.o $(OUT_DIR)
 
+# Use Gnu Make 4.3 "grouped targets" to make both $(OUT_ARCHIVE)
+# and $(OUT_CHECKSUMS) with one command.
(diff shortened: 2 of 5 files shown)


tracee-ebpf/Makefile
```diff
@@ -145,10 +147,13 @@ mostlyclean:
 
 .PHONY: clean
 clean:
-	-$(CMD_DOCKER) rmi $(file < $(docker_builder_file))
+	-FILE="$(docker_builder_file)" ; \
+	if [ -n "$$FILE" ] && [ -r "$$FILE" ] ; then \
```
itaysk, line 151: I believe -n is redundant if -r is checked no?
(applies to other occurrence as well)

tracee-ebpf/Makefile
```diff
@@ -127,6 +127,8 @@ endif
 # record built image id to prevent unnecessary building and for cleanup
 docker_builder_file := $(OUT_DIR)/$(DOCKER_BUILDER)
 .PHONY: $(DOCKER_BUILDER)
+# Use Gnu Make 4.3 "grouped targets" to make both $(DOCKER_BUILDER)
+# and $(docker_builder_file) with one command.
```
itaysk, line 131: I think it's sufficient to mention this in the readme prereqs as we wouldn't want to document every time we use this feature of other "advanced" features. perhaps mention as an FAQ in github discussions. Do you think that would suffice?
(applies to other occurrence as well)

icarus-sparry: I am happy to remove this comment, or even to rewrite the Makefile to not require the feature. I haven't even decided in my own mind if it makes sense to have a .PHONY and a real target as "grouped targets".

tracee-ebpf/Makefile
icarus-sparry: It is defensive programming from a more elegant era that no longer exists. I will remove it.

tracee-ebpf/Makefile
itaysk: maybe just put a comment at the top of the file that this makefile required GNU Make 4.3, I think that's will help.

> rewrite the Makefile to not require the feature

happy to hear about ideas for improvements! If you have a concrete idea, please open an issue do discuss it

icarus-sparry:
@itaysk I believe I have addressed the review comments.

itaysk (approved):
LGTM, thanks!
I see that you've added go 1.16 requirement to the readme (good catch), could you please add that also to the tracee-ebpf readme before we merge?

itaysk:
closes: https://github.com/aquasecurity/tracee/pull/615

icarus-sparry:
> I see that you've added go 1.16 requirement to the readme (good catch), could you please add that also to the tracee-ebpf readme before we merge?

Done

PR #620, opened 2021-03-15
Fix RingBuffer shutdown

### LINKED ISSUE #619: RingBuffer.Stop is racy
There are currently a few issues with RingBuffer.Stop:
1. `rb.stopped` flag is read and updated without the use of `atomic`, leading to a race detected by the race detector.
2. `Stop` does not wait for the poll goroutine to exit. The followup call to `Close` will destroy the ring buffer and cause the poll goroutine to access bad memory.
3. If there are kernel events happening during the shutdown, a deadlock is possible. If the consumer first stops reading from the channel and then calls `Stop`, pending events can deadlock the `ring_buffer` callback (in the write to the channel).

eyakubovich:
Fixes RingBuffer's Stop to do a clean exit.
Adds a polling 300ms timeout to check for stop condition.
The Stop() method waits for the polling goroutine to exit
before returning to the caller.

Fixes #619

### DIFF
--- a/libbpfgo/libbpfgo.go
+++ b/libbpfgo/libbpfgo.go
@@ -181,6 +181,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"sync"
 	"syscall"
 	"unsafe"
 )
@@ -231,8 +232,9 @@ type PerfBuffer struct {
 type RingBuffer struct {
 	rb      *C.struct_ring_buffer
 	bpfMap  *BPFMap
-	stopped bool
-	closed bool 
+	stop    chan struct{}
+	closed  bool
+	wg      sync.WaitGroup
 }
 
 // BPF is using locked memory for BPF maps and various other things.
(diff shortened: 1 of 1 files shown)


libbpfgo/libbpfgo.go
```diff
@@ -648,39 +650,71 @@ func (m *Module) InitRingBuf(mapName string, eventsChan chan []byte) (*RingBuffe
...
-	rb.stopped = true
+	if rb.stop != nil {
+		// Tell the poll goroutine that it's time to exit
+		close(rb.stop)
+
+		// The event channel should be drained here since the consumer
+		// may have stopped at this point. Failure to drain it will
+		// result in a deadlock: the channel will fill up and the poll
+		// goroutine will block in the callback.
+		eventChan := eventChannels[uintptr(rb.bpfMap.fd)]
```
grantseltzer, line 667: Ah this was the pattern that I failed to recognize the need for in previous iterations of this code. I couldn't get past the fact that the callback was blocking (and the timeout passed to ring_buffer__poll would interrupt it).

grantseltzer (approved):
Thanks so much for this @eyakubovich! Figuring out how to avoid blocking while completely avoiding a race condition at shutdown was throwing me for quite a loop. It's a great feeling not only to get a code contribution but to also learn from it.

Can you share the command(s) you used for running the race detector against this? I'd love to add it to our github actions CI/CD.

eyakubovich:
The race detector ran as part of the unit tests (`go test -race`) but it's in another project that uses libbpfgo.

There's actually another race: `eventChannels` map is not protected by a mutex. However it's probably best to try to remove the map itself (at least for the ring buffer case) and just store the channel inside the `RingBuffer` object. However that will be a separate issue/PR.

grantseltzer:
Got it! If you'd like to work on that it's more than welcome, looking into it right now just to understand it. Is that project open source? If so could you share it?

eyakubovich:
Yes, it's open source and called Teleport (https://github.com/gravitational/teleport).
The patch that moves it from BCC to just `libbpf` is not there yet as I'm working out the kinks.
However you can look at the Makefile to see how the tests are ran.

grantseltzer:
@eyakubovich Thanks for the info. Please feel free to file issues as you work on moving it over. I'm actively adding functionality to libbpf, among other things I'm hoping to have the full libbpf API implemented in the next couple months.

PR #629, opened 2021-03-17
feat: Add support for wildcard event suffixes

### LINKED ISSUE #94: group similar syscalls in the ux
some sycalls have different variants: `open`/`openat`/`open_by_handle_at`,  `dup`/`dup2`/`dup3`.
perhaps we can accept a flag `-e open* -e dup*` which will catch all related syscalls. we can do a literal search by prefix which will work for most examples.

jan0ski:
Allows for grouping of similar syscalls in UX via wildcard suffixes, e.g. `event=open*` or `event!=dup*`

Resolves  #94


simar7:
hi @jan0ski - I noticed that the file perms have changed 
image

Is there a reason why that was required?

jan0ski:
> 
> hi @jan0ski - I noticed that the file perms have changed
> image
> 
> Is there a reason why that was required?

Good catch, completely unintentional.  I'll fix those permissions.

jan0ski:
Example output:
```
$ sudo ./dist/tracee-ebpf -t 'event=dup*'
TIME(s)        UID    COMM             PID     TID     RET              EVENT                ARGS
728598.819035  1000   sh               11908   11908   1                dup2                 oldfd: 4, newfd: 1
728598.820178  1000   sh               11909   11909   1                dup2                 oldfd: 4, newfd: 1
728598.831100  1000   bash             11910   11910   255              dup2                 oldfd: 3, newfd: 255
728598.831656  1000   tmux: client     11908   11908   6                dup                  oldfd: 0
728598.832653  1000   tmux: client     11909   11909   6                dup                  oldfd: 0
728598.832700  1000   bash             11911   11911   1                dup2                 oldfd: 4, newfd: 1
728598.834704  1000   tmux: client     11908   11908   6                dup                  oldfd: 1
728598.835877  1000   tmux: client     11909   11909   6                dup                  oldfd: 1
...
^C
End of events stream
Stats: {eventCounter:46 errorCounter:0 lostEvCounter:0 lostWrCounter:0}
```

grantseltzer:
Hi @jan0ski! Thank you for tackling this!  I need to take a closer look at this tomorrow, but could you also update the help text here: https://github.com/aquasecurity/tracee/blob/main/tracee-ebpf/main.go#L285

tracee-ebpf/main.go
```diff
@@ -798,23 +798,55 @@ func prepareEventsToTrace(eventFilter *tracee.StringFilter, setFilter *tracee.St
...
 	res = make([]int32, 0, len(tracee.EventsIDToEvent))
 	for _, name := range eventsToTrace {
-		id, ok := eventsNameToID[name]
-		if !ok {
-			return nil, fmt.Errorf("invalid event to trace: %s", name)
+		// Handle event prefixes with wildcards
+		if strings.HasSuffix(name, "*") {
+			var ids []int32
+			found := false
+			prefix := name[:strings.Index(name, "*")]
```
grantseltzer, line 832: Since `*` is the suffix, you can replace `strings.Index(name, "*")` with `len(name)-1`

tracee-ebpf/main_test.go
```diff
@@ -191,6 +191,18 @@ func TestPrepareFilter(t *testing.T) {
...
+			filters:        []string{"event=blah*"},
+			expectedFilter: tracee.Filter{},
+			expectedError:  errors.New("invalid event to trace: blah*"),
+		},
+		{
+			testName:       "invalid not wildcard",
+			filters:        []string{"event!=blah*"},
+			expectedFilter: tracee.Filter{},
+			expectedError:  errors.New("invalid event to exclude: blah*"),
+		},
```
grantseltzer, line 205: First off, thanks for adding tests! Could you add a few test cases for the following cases?

-  having the `*` elsewhere in the filters (i.e. 'bl*ah')
- combinations of filters, in some cases where both valid, or one invalid, or both invalid (i.e. 'event=open*,exec*' or 'event!=open* event=open*'

grantseltzer (changes requested):

tracee-ebpf/main_test.go
jan0ski: For testing the `event!=<prefix>*` test cases, I couldn't find a succinct way to set `EventsToTrace` to be "every event ID but x".  Since there are over 1000 event IDs didn't think we'd want hard code a long slice with nearly every event ID in it.  Am I thinking about this wrong?

grantseltzer: Ah, yea that'd be obnoxious. So I think the way this _should_ be done is having a separate test function where instead of having a comparison of each valid event it'd just check the length or that certain ones are contained in the list. 

However, that's way too much to ask of an open source contribution. If you feel particularly motivated to do that (no pressure) you can open up a second PR for it, but otherwise just changing the `strings.Index` -> `len` that I suggested above should be plenty for closing up this PR!

tracee-ebpf/main.go
jan0ski: - [ ] replace strings.Index(name, "*") with len(name)-1

tracee-ebpf/main_test.go
jan0ski: Sounds good.  I'll look into it and potentially open a new issue for this.

grantseltzer (approved):
LGTM! Thanks again @jan0ski

Issue #612, opened 2021-03-13
followup on gotemplate support


itaysk:
https://github.com/aquasecurity/tracee/pull/582 added support for gotemplates there is some  missing some key points. I've added the bug label just because these were part of the issue requirements:

- [ ] can't print current time in template
- [ ] template examples use `.Context.Timestamp` (see [here](https://github.com/aquasecurity/tracee/pull/582#discussion_r584263000) why it's wrong)
- [ ] can't use any of the SignatureMetadata in a template (e.g print the signature name)

Generally I think it would be better if we used templates also in to print stdout, this would have served as a real world test that would reveal the issues described here. I've opened #611 for this

Issue #614, opened 2021-03-13
add sprig to gotemplates


itaysk:
[sprig](http://masterminds.github.io/sprig/) is a popular and useful library of functions that helps authoring go templates. kind of like a stdlib for gotemplates. I think it would be a good addition to whomever writes templates in tracee-ebpf/tracee-rules. 

It might also help address some outstanding issues like https://github.com/aquasecurity/tracee/pull/582#discussion_r584263304

simar7:
Added in https://github.com/aquasecurity/tracee/pull/630

Issue #611, opened 2021-03-13
use gotemplates in stdout


itaysk:
We already use gotemplates for compiling wehbook payload. the same could be used to compile the stdout message. this has the following benefits:

1. generally, same features should be supported in stdout and webhook printing, so no reason of having two implementations (like now)
2. easier to detect bugs/issues
3. allows the user to customize the stdout template like in tracee-ebpf `--output gotemplate=/path/to/template`

simar7:
For my understanding, are the following assumptions correct to qualify as an _acceptance criteria_?

1. `tracee-rules` will support an `--output` flag.
2. The above mentioned `--output` flag will only support one argument as an option, i.e. a gotemplate param: `--output gotemplate=path/to/my.tmpl`.
3. If no `--output` flag is specified the output will be the human readable output we support today. Which for posterity is:
```
*** Detection ***
Time: 2021-03-15T23:45:41Z
Signature ID: TRC-2
Signature: Anti-Debugging
Data: map[]
Command: 
Hostname: 
```
4. Sample usage of the new feature will look as follows:
```
./tracee-rules --rules TRC-1  --output gotemplate=path/to/csv.tmpl
TRC-2,Anti-Debugging

```

Where the `csv.tmpl` is as follows:
```
{{ .ID }},{{ .Name }}

```

itaysk:
yes, with the addition that

>If no --output flag is specified the output will be the human readable output we support today.

will be using a gotemplate as well

simar7:
Closed via https://github.com/aquasecurity/tracee/pull/630

PR #630, opened 2021-03-18
Use Go templates for stdout

simar7:
This PR adds the following:

- `--output` flag to tracee-rules, which enables to BYOT (bring your own template) for stdout.
- Use wall time as part of event detection capture.
- Ability to use Signature Metadata fields inside of custom templates.
- If no custom template is specified for `--output`, tracee-rules uses a default predefined template.

Fixes: #612, #611 and https://github.com/aquasecurity/tracee/issues/614


tracee-rules/main.go
```diff
@@ -111,6 +112,10 @@ func main() {
 				Name:  "input-tracee",
 				Usage: "configure tracee-ebpf as input source. see '--input-tracee help' for more info",
 			},
+			&cli.StringFlag{
+				Name:  "output",
+				Usage: "configure output format via templates. Usage: --output=path/to/my.tmpl",
```
itaysk, line 117: can we make it output-template?

tracee-rules/types/types.go
```diff
@@ -49,3 +49,8 @@ type Finding struct {
 	Context   Event
 	Signature Signature
 }
+
+type FindingWithMetadata struct {
+	Finding
+	SignatureMetadata
+}
```
itaysk, line 56: @simar7 what do you think on changing Finding type to include the SignatureMetadata instead of the Signature interface? @yanivagman also curios what you think

yanivagman: Signature interface has GetMetadata() method - why not using it when needed?

itaysk: Yaniv and I discussed this offline, we should replace signature field with signaturemetadata in the finding. This will simplify many cases without sacrificing much.

simar7: Should we do that as part of this PR? I'm happy to do it just feel it's unrelated to this PR as modifying an existing interface requires all users of it (example signatures in this case) to be changed as well.

tracee-rules/output.go
```diff
@@ -37,6 +38,7 @@ func setupTemplate(inputTemplateFile string, clock Clock) (*template.Template, e
 	case inputTemplateFile != "":
 		return template.New(filepath.Base(inputTemplateFile)).
 			Funcs(funcMap).
```
simar7, line 40: NB: I'm still keeping this in because to my knowledge there's no way to inject a custom time interface into sprig which makes it untestable for our most common case of template (simple and verbose) where we need the current time.

itaysk: why do we need to test it specifically? if we want to test that the templating engine works, can we  craft a simpler template for the test?

tracee-rules/types/types.go
itaysk: whatever you prefer

itaysk (approved):
LGTM, although I do prefer to remove the timenow function before merging if you agree with my comment

tracee-rules/output.go
simar7: Yes fair point. I wanted to test the default output (without any templates, which includes a timestamp). I found a another way to do it https://github.com/aquasecurity/tracee/pull/630/commits/bf6a78f87e6175c2c1669448ccb383032c1b38de and cleaned this up.

tracee-rules/types/types.go
simar7: I'll draft a new PR

simar7: https://github.com/aquasecurity/tracee/pull/646

Issue #485, opened 2021-01-29
Move argprinters from tracee-ebpf to libbpfgo


itaysk:
These are static functions that are useful for anyone working with Linux internals in Go. It can be a nice companion to writing bpf programs with libbpfgo. Move those into a package in libbpfgo (not `package libbfgo`).

yanivagman:
Note that PrintAlert is unique to tracee, so it should probably be left in tracee-ebpf

PR #631, opened 2021-03-18
Use libbpfgo/helpers package in tracee-ebpf

grantseltzer:
This is a follow up to #493, which should fully close out #485 

Question I have is whether or not we want to copy over the `PrintUint32IP` and `Print16BytesSliceIP` to this helpers package. Not sure if it's relevant enough?


yanivagman:
> Question I have is whether or not we want to copy over the PrintUint32IP and Print16BytesSliceIP to this helpers package. Not sure if it's relevant enough?

I don't think we should copy these helpers. The string representation of an address can be considered as "raw" by itself

yanivagman (approved):
Overall LGTM, but maybe we should remove the argprinters.go file and move the functions left (PrintUint32IP, Print16BytesSliceIP, PrintAlert) to tracee.go?

grantseltzer:
@yanivagman I think 1600 lines is enough for tracee.go, to be honest I wouldn't mind breaking it up further to make it easier to read/navigate. 

Lets handle that in a future discussion/PR

Issue #560, opened 2021-02-17
tracee-ebpf panics on invalid flag usage


simar7:
**What I wanted to do**
Print events as JSON output to a file.

**What I observed**
```
sudo ./dist/tracee-ebpf -o json -o out.json
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x5a17a4]

goroutine 1 [running]:
github.com/aquasecurity/tracee/tracee-ebpf/tracee.newEventPrinter(0xc0000acc50, 0x8, 0xc000111e00, 0x73c8e0, 0xc0000ae008, 0x73c8e0, 0xc0000ae010, 0xc00012ca00, 0x3c, 0x0, ...)
        /usr/vagrant/repos/tracee/tracee-ebpf/tracee/printer.go:67 +0xc4
github.com/aquasecurity/tracee/tracee-ebpf/tracee.New(0xc000119d00, 0xc000111f00, 0xc000111ec0, 0x400, 0x400, 0x0, 0x0, 0xc00012ca00, 0x3c, 0xc000080460, ...)
        /usr/vagrant/repos/tracee/tracee-ebpf/tracee/tracee.go:272 +0x3a5
main.main.func1(0xc000111e80, 0xc00013e400, 0x14)
        /usr/vagrant/repos/tracee/tracee-ebpf/main.go:75 +0x5ad
github.com/urfave/cli/v2.(*App).RunContext(0xc000086480, 0x73e9a0, 0xc0000ac010, 0xc000080050, 0x5, 0x5, 0x0, 0x0)
        /home/vagrant/go/pkg/mod/github.com/urfave/cli/v2@v2.1.1/app.go:311 +0x70d
github.com/urfave/cli/v2.(*App).Run(...)
        /home/vagrant/go/pkg/mod/github.com/urfave/cli/v2@v2.1.1/app.go:211
main.main()
        /usr/vagrant/repos/tracee/tracee-ebpf/main.go:144 +0x832
```

**What I expected to see**
1. Tracee not panicking.
2. JSON event output in a file. It works when I specify the correct flags: `sudo ./dist/tracee-ebpf -o json -o out-file:out.json
`

**Acceptance Criteria**
Tracee should not panic but rather gracefully exit with an meaningful error message.

simar7:
[...]

PR #632, opened 2021-03-18
fix: trace-ebpf flag output

krol3:
Related issue https://github.com/aquasecurity/tracee/issues/560
- Using switch and validateOutputFormatValue method to validate the default output format values("table", "table-verbose", "json", "gob","gotemplate=").

### DIFF
--- a/tracee-ebpf/main.go
+++ b/tracee-ebpf/main.go
@@ -177,7 +177,7 @@ Use this flag multiple times to choose multiple capture options
 			res.Format = outputParts[0]
 			err := res.Validate()
 			if err != nil {
-				return res, fmt.Errorf("%s. Valid format values: 'table', 'table-verbose', 'json', 'gob' or 'gotemplate='. Use '--output help' for more info.", err)
+				return res, err
 			}
 			outputParts = append(outputParts, outputParts[0])
 			outputParts[0] = "format"
@@ -188,7 +188,7 @@ Use this flag multiple times to choose multiple capture options
 			res.Format = outputParts[1]
 			err := res.Validate()
 			if err != nil {
-				return res, fmt.Errorf("%s. Valid format values: 'table', 'table-verbose', 'json', 'gob' or 'gotemplate='. Use '--output help' for more info.", err)
+				return res, err
 			}
 		case "out-file":
 			res.OutPath = outputParts[1]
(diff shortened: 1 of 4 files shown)


tracee-ebpf/main.go
```diff
@@ -175,16 +175,24 @@ Use this flag multiple times to choose multiple capture options
 		outputParts := strings.SplitN(o, ":", 2)
 		numParts := len(outputParts)
 		if numParts == 1 {
+			if !validateOutputFormatValue(outputParts[0]) {
```
simar7, line 178: Instead of validating when we get inside `prepareOutput` it will be easier to validate outside, prior to reaching here. So I suggest that before we call `prepareOutput` we can pass `c.StringSlice(output)` to another function that can validate all types of possible inputs that can be passed to `--output` flag.

This will keep the abstraction of `prepareOutput` to assume the inputs it receives are valid and also make it easier to read and test as it will have less branches. Furthermore testing will be simplified as well as we can just test both functions individually.

itaysk: regarding validation, I don't think that preparer functions should have validation logic. the XXXConfig struct already has the validation logic https://github.com/aquasecurity/tracee/blob/f8df7da6a27f729610992b6bd52e89d510fcf384/tracee-ebpf/tracee/tracee.go#L114 so we should just use it whenever is needed. 

also I think other configs should follow this pattern. wdyt @simar7 ?

krol3: Thanks @simar7 . I used the Validate() method and removed the other method that I created.

tracee-ebpf/main.go
```diff
@@ -175,16 +175,27 @@ Use this flag multiple times to choose multiple capture options
...
 		}
-		if outputParts[0] == "format" {
+
+		switch outputParts[0] {
+		case "format":
 			res.Format = outputParts[1]
-		} else if outputParts[0] == "out-file" {
+			err := res.Validate()
+			if err != nil {
+				return res, fmt.Errorf("%s. Valid format values: 'table', 'table-verbose', 'json', 'gob' or 'gotemplate='. Use '--output help' for more info.", err)
```
grantseltzer, line 192: I think this error string should be returned in `Validate()` itself.

grantseltzer (approved):
One nit from me, otherwise this looks good! Will merge after that small change.

Thanks @krol3!

grantseltzer (approved):

grantseltzer:
@krol3 once https://github.com/aquasecurity/tracee/pull/688 merges you'll need to rebase in order to fix the build, then i'll merge this

simar7:
@krol3 looks like there's a few compile time issues. If you fix those and tests pass, we can merge this in as the rest of the code is good.

PR #633, opened 2021-03-18
add mkdocs documentation

### LINKED ISSUE #435: create a dedicate documentation site

ronaudinho:
resolves #435

the docs are taken from existing READMEs with some modifications, as far as I remember, tracee has different project structure a few months back, so there are a number of overlaps especially between the root README and tracee-ebpf/README, which is reflected in the current mkdocs structure as well.


.github/workflows/mkdocs-dev.yaml
```diff
@@ -0,0 +1,34 @@
...
+        run: |
+          pip install git+https://${GH_TOKEN}@github.com/squidfunk/mkdocs-material-insiders.git
+          pip install mike
+          pip install mkdocs-macros-plugin
+        env:
+          GH_TOKEN: ${{ secrets.ORG_GITHUB_TOKEN }}
+      - name: Configure the git user
+        run: |
+          git config user.name "ronaudinho"
+          git config user.email "doelaudi@gmail.com"
```
ronaudinho, line 32: might want to adjust this @itaysk

.github/workflows/mkdocs-latest.yaml
```diff
@@ -0,0 +1,33 @@
...
+        run: |
+          pip install git+https://${GH_TOKEN}@github.com/squidfunk/mkdocs-material-insiders.git
+          pip install mike
+          pip install mkdocs-macros-plugin
+        env:
+          GH_TOKEN: ${{ secrets.ORG_GITHUB_TOKEN }}
+      - name: Configure the git user
+        run: |
+          git config user.name "ronaudinho"
+          git config user.email "doelaudi@gmail.com"
```
ronaudinho, line 29: ditto for this

docs/rules/rego.md
```diff
@@ -0,0 +1,7 @@
+Create a `.rego` file in the rules directory that has the following Rego Rules (in this context rules are Rego's language constructs):
+
+1. `__rego_metadoc__`: A *document* rule that defines the rule's metadata (based on [WIP Rego convention](https://hackmd.io/@ZtQnh19kS26YiNlJLqKJnw/H1gAv5nB) for describing policy metadata).
```
ronaudinho, line 3: the link returns not found

itaysk: you can remove whatever's in parenthesis

.github/workflows/mkdocs-dev.yaml
itaysk: can we just remove these lines? in other workflows where we git push without overriding the username, it shows as the "Github Actions" user, which is better I think

.github/workflows/mkdocs-latest.yaml
itaysk: see previous comment

mkdocs.yml
```diff
@@ -0,0 +1,64 @@
+site_name: Tracee
+site_url: https://aquasecurity.github.io/tracee/
+site_description: Tracee is a runtime security and forensics tool for Linux, using Linux eBPF technology to trace your system and applications at runtime, and analyze collected events to detect suspicious behavioral patterns.
```
itaysk, line 3:
```suggestion
site_description: Runtime security and forensics using eBPF.
```

itaysk (commented):
good job! I think I'll follow up with another PR that re-arranges the docs, but that's great as a starting point.
let's address the minor comments and then we can merge

grantseltzer:
Would this then allow us to turn on github pages and have a webpage with the documentation?

itaysk:
> Would this then allow us to turn on github pages and have a webpage with the documentation?

yes that's the goal

itaysk (approved):

PR #635, opened 2021-03-19
fix relative link to quickstart-with-docker

dmitris:
Currently the "quickstart" relative link doesn't do anything since the target section has changed


simar7 (approved):
thank you!

PR #638, opened 2021-03-23
Fix PerfBuffer shutdown

### LINKED ISSUE #640: PerfBuffer.Stop is racy
Akin to the issue #619, PerfBuffer' Stop method just sets a non-atomic boolean and exits. It should use a synchronization primitive to signal stop and wait for the polling goroutine t exit.

eyakubovich:
This is anlogous to 09b2b47c (which fixed RingBuffer).
It does a clean PerfBuffer stop by waiting for the
polling goroutine to finish its work.

Fixes #640

### DIFF
--- a/libbpfgo/libbpfgo.go
+++ b/libbpfgo/libbpfgo.go
@@ -224,7 +224,9 @@ type BPFLink struct {
 type PerfBuffer struct {
 	pb     *C.struct_perf_buffer
 	bpfMap *BPFMap
-	stop   chan bool
+	stop   chan struct{}
+	closed bool
+	wg     sync.WaitGroup
 }
 
 type RingBuffer struct {
@@ -664,7 +666,7 @@ func (rb *RingBuffer) Stop() {
 		// goroutine will block in the callback.
 		eventChan := eventChannels[uintptr(rb.bpfMap.fd)]
 		go func() {
-			for _ = range eventChan {
+			for range eventChan {
 			}
 		}()
 
@@ -736,26 +738,65 @@ func (m *Module) InitPerfBuf(mapName string, eventsChan chan []byte, lostChan ch
 	perfBuf := &PerfBuffer{
 		pb:     pb,
 		bpfMap: bpfMap,
-		stop:   make(chan bool),
 	}
 	m.perfBufs = append(m.perfBufs, perfBuf)
 	return perfBuf, nil
 }
 
 func (pb *PerfBuffer) Start() {
+	pb.stop = make(chan struct{})
+	pb.wg.Add(1)
 	go pb.poll()
 }
 
 func (pb *PerfBuffer) Stop() {
-	pb.stop <- true
+	if pb.stop != nil {
+		// Tell the poll goroutine that it's time to exit
+		close(pb.stop)
+
+		eventsChan := eventChannels[uintptr(pb.bpfMap.fd)]
+		lostChan := lostChannels[uintptr(pb.bpfMap.fd)]
+
+		// The event and lost channels should be drained here since the consumer
+		// may have stopped at this point. Failure to drain it will
+		// result in a deadlock: the channel will fill up and the poll
+		// goroutine will block in the callback.
+		go func() {
+			for range eventsChan {
+			}
+
+			if lostChan != nil {
+				for range lostChan {
+				}
+			}
+		}()
+
+		// Wait for the poll goroutine to exit
+		pb.wg.Wait()
+
+		// Close the channel -- this is useful for the consumer but
+		// also to terminate the drain goroutine above.
+		close(eventsChan)
+		close(lostChan)
+
+		// This allows Stop() to be called multiple times safely
+		pb.stop = nil
+	}
 }
 
 func (pb *PerfBuffer) Close() {
+	if pb.closed {
+		return
+	}
+	pb.Stop()
 	C.perf_buffer__free(pb.pb)
+	pb.closed = true
 }
 
 // todo: consider writing the perf polling in go as c to go calls (callback) are expensive
 func (pb *PerfBuffer) poll() error {
+	defer pb.wg.Done()
+
 	for {
 		select {
 		case <-pb.stop:
--- a/libbpfgo/selftest/perfbuffers/run.sh
+++ b/libbpfgo/selftest/perfbuffers/run.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+GREEN='    \033[0;32m'
+RED='    \033[0;31m'
+NC='\033[0m' 
+
+VERSION_LIMIT=4.3
+CURRENT_VERSION=$(uname -r | cut -d '.' -f1-2)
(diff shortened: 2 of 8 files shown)


libbpfgo/selftest/perfbuffers/run.sh
```diff
@@ -0,0 +1,49 @@
...
+if (( $(echo "$MAJOR_CURRENT_VERSION < $MAJOR_VERSION_LIMIT") |bc -l)); then
+    echo -e "${RED}[*] OUTDATED MAJOR KERNEL VERSION${NC}"
+    exit 1
+fi
+
+# Check minor version
+MINOR_VERSION_LIMIT=$(echo $VERSION_LIMIT | cut -d '.' -f2)
+MINOR_CURRENT_VERSION=$(echo $CURRENT_VERSION | cut -d '.' -f2)
+if (( $(echo "$MINOR_CURRENT_VERSION < $MINOR_VERSION_LIMIT") |bc -l)); then
+    echo -e "${RED}[*] OUTDATED MINOR KERNEL VERSION${NC}"
```
yanivagman, line 22: if 
$MAJOR_CURRENT_VERSION > $MAJOR_VERSION_LIMIT - 
there is no reason to check for minor, otherwise, version like 6.0 will not pass this test, right?

libbpfgo/selftest/perfbuffers/run.sh
```diff
@@ -0,0 +1,49 @@
...
+    exit 0
```
yanivagman, line 23: don't we want to exit with a value different from 0? If we do want to use exit 0 here, why not for the above major test as well?

yanivagman (approved):
Thanks @eyakubovich!
LGTM
only two small comments

grantseltzer:
@yanivagman the minor/major checks look like they're just copied from the script I have for running the ringbuffer selftest so I can answer for them. 

For your first comment on checking minor version after major, you're absolutely right. I'm going to have to fix this.

The reason we exit 0 is for the sake of the github action not failing. I picture having multiple environments where the github actions run, including older kernels. We should just not run the test if we don't expect the kernel to support it. I suppose we can return a particular exit code and have the test runner make exceptions for this scenario but we're not there yet.

libbpfgo/libbpfgo.go
```diff
@@ -736,26 +738,65 @@ func (m *Module) InitPerfBuf(mapName string, eventsChan chan []byte, lostChan ch
...
+		// The event and lost channels should be drained here since the consumer
+		// may have stopped at this point. Failure to drain it will
+		// result in a deadlock: the channel will fill up and the poll
+		// goroutine will block in the callback.
+		go func() {
+			for _ = range eventsChan {
+			}
+
+			if lostChan != nil {
+				for _ = range lostChan {
```
grantseltzer, line 769:
```suggestion
				for range lostChan {
```

grantseltzer (changes requested):

libbpfgo/selftest/perfbuffers/run.sh
```diff
@@ -0,0 +1,49 @@
...
+else
+    echo -e "${GREEN}[*] SUFFICIENT KERNEL VERSION${NC}"
+fi
```
grantseltzer, line 26:
```suggestion
# Check minor version, only if MAJOR_CURRENT_VERSION is same as MAJOR_VERSION_LIMIT
MINOR_VERSION_LIMIT=$(echo $VERSION_LIMIT | cut -d '.' -f2)
MINOR_CURRENT_VERSION=$(echo $CURRENT_VERSION | cut -d '.' -f2)
if (( $(echo "$MAJOR_CURRENT_VERSION == $MAJOR_VERSION_LIMIT") |bc -l)); then
    if (( $(echo "$MINOR_CURRENT_VERSION < $MINOR_VERSION_LIMIT") |bc -l)); then
        echo -e "${RED}[*] OUTDATED MINOR KERNEL VERSION${NC}"
        exit 1
    else
        echo -e "${GREEN}[*] SUFFICIENT KERNEL VERSION${NC}"
    fi
fi
```

grantseltzer: @yanivagman @eyakubovich 

How does this look? We only need to check minor version if the major version is equal to the major version limit.

libbpfgo/libbpfgo.go
```diff
@@ -736,26 +738,65 @@ func (m *Module) InitPerfBuf(mapName string, eventsChan chan []byte, lostChan ch
...
+
+		eventsChan := eventChannels[uintptr(pb.bpfMap.fd)]
+		lostChan := lostChannels[uintptr(pb.bpfMap.fd)]
+
+		// The event and lost channels should be drained here since the consumer
+		// may have stopped at this point. Failure to drain it will
+		// result in a deadlock: the channel will fill up and the poll
+		// goroutine will block in the callback.
+		go func() {
+			for _ = range eventsChan {
```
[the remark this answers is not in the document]
eyakubovich: very nice! I didn't know about this variant of `for`.

libbpfgo/selftest/perfbuffers/run.sh
eyakubovich: Yes, I copied the whole file (as well as most of the files in the directory) as is from the ringbuffer test and didn't look too much into it. Since we're relying on `bc` anyway, you can just compare both major and minor in one shot:
```
if [ $(echo "$CURRENT_VERSION < $VERSION_LIMIT" | bc) -eq "1" ]; then
    echo -e "${RED}[*] OUTDATED KERNEL VERSION${NC}"
    exit 1
fi
```
Also perf buffer was added in 4.3 so I need to adjust the `VERSION_LIMIT` value.

itaysk: I'm not sure bc is good for version comparison:
```
❯ echo "1.30 < 1.3" | bc
0
```
I'd recommend `sort -v` which is version aware comparisons

grantseltzer: Got it, that works too, once you push that change i'll go ahead and approve!

eyakubovich: @itaysk you're right, of course, can't compare them as decimals. How do I do it with sort? All I can come up with is:
```
min=$(sort -V <<EOF | head -n 1
$CURRENT_VERSION
$VERSION_LIMIT
EOF
)
if [ "$min" -ne "$VERSION_LIMIT" ]; then
  echo "too low"
fi
```
Is there anything cleaner?

itaysk: maybe something like 
```bash
test $(echo "$CURRENT_VERSION\n$VERSION_LIMIT" | sort -V | head -n1) = "$VERSION_LIMIT"
```

grantseltzer (approved):
LGTM! Thanks so much @eyakubovich, much appreciated!

PR #645, opened 2021-03-24
feat: add functions to helper package for checking kernel config options

grantseltzer:
In development of CO:RE work for tracee-ebpf I needed functionality to check kernel config options (for example, `CONFIG_DEBUG_INFO_BTF`). Other projects like cilium appear to shell out to `bpftool feature` but then we rely on bpftool tracking the [needed config options](https://git.kernel.org/pub/scm/linux/kernel/git/bpf/bpf-next.git/commit/?id=b9fc8b4a591811546fec2dbef7e9f809362100c9). We also shouldn't exec. 

This PR just adds the Go side functionality for checking kernel config. Relevant config options will be read into a BPF map and checked bpf side.


libbpfgo/helpers/kernel_features.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+type KernelConfig map[string]string
+
+// NewKernelConfig simply creates the underlying map used by
+// the KernelConfig, you must use InitKernelConfig to read
+// in the values of the config
+func NewKernelConfig() KernelConfig {
+	kconfig := make(map[string]string)
+	return kconfig
```
simar7, line 22: could we inline this?

```suggestion
	return make(map[string]string)
```

libbpfgo/helpers/kernel_features.go
```diff
@@ -0,0 +1,111 @@
...
+// InitKernelConfig attempts to read the kernel config
+//  into the KernelConfig from either
+// - /proc/config-$(uname -r)
+// - /boot/config.gz
+func (k KernelConfig) InitKernelConfig() error {
+
+	err := k.getBootConfig()
+	if err == nil {
+		return nil
+	}
```
simar7, line 34: I assume you wanted to check `if err != nil { return err }`?

libbpfgo/helpers/kernel_features.go
grantseltzer: Yup, good call. Interestingly the compiler inlines it anyway!

libbpfgo/helpers/kernel_features.go
grantseltzer: Nope, some kernels are compiled with `CONFIG_IKCONFIG=y` which puts the config file at /proc/config.gz instead of `/boot/config-$(uname -r)`, so the err==nil is there to fall through

itaysk (commented):
this is great functionality to have in libbpfgo. May I suggest some UX improvements:
Instead of:
```go
import github.com/.../libbpfgo/helpers
...
helpers.NewKernelConfig()
helpers.InitKernelConfig()
conf := helpers.KernelConfig["something"]
``` 

we can have:
```go
import github.com/.../libbpfgo/kconfig
...
conf, err := kconfig.Get("something)
```


I don't see why a user would care to instantiate the map, then populate it manually every time they want to get a config. Instead we can just do all of that in the package's `init()`. WDYT?

grantseltzer:
> this is great functionality to have in libbpfgo. May I suggest some UX improvements:
> Instead of:
> 
> ```go
> import github.com/.../libbpfgo/helpers
> ...
> helpers.NewKernelConfig()
> helpers.InitKernelConfig()
> conf := helpers.KernelConfig["something"]
> ```
> 
> we can have:
> 
> ```go
> import github.com/.../libbpfgo/kconfig
> ...
> conf, err := kconfig.Get("something)
> ```
> 
> I don't see why a user would care to instantiate the map, then populate it manually every time they want to get a config. Instead we can just do all of that in the package's `init()`. WDYT?

That's how I originally designed it (with an init) but decided against it. I'm not sure _why_ this would happen, but I was worried about a situation where the config file is created after the init function is called. I suppose this could happen if the kernel is compiled with `CONFIG_IKCONFIG` but not `CONFIG_IKCONFIG_PROC` and the file is extracted to /proc after the fact (again, not sure why that'd happen). I'd say in general I found the design to be more dangerous for suppressing errors that can't be repaired. 

I will say, there's no reason to have both a `NewKernelConfig` and an `InitKernelConfig`, I am going to put the logic for Init into New.

libbpfgo/helpers/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/aquasecurity/tracee/libbpfgo/helpers
```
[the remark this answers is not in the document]
grantseltzer: I think I lack proper understanding of go modules to be honest. What is the difference here? (I will do reading on modules today)

libbpfgo/helpers/example_tracelisten_test.go
```diff
@@ -1,15 +1,13 @@
...
+package helpers
 
 import (
 	"fmt"
 	"os"
-
-	"github.com/aquasecurity/tracee/libbpfgo/helpers"
 )
 
 func ExampleTracePipeListen_usage() {
```
yanivagman, line 8: Now that you changed this file package to "helpers" do we want to expose this function as part of the helpers package?

libbpfgo/helpers/kernel_features.go
```diff
@@ -0,0 +1,110 @@
...
+
+	"golang.org/x/sys/unix"
+)
+
+type KernelConfig map[string]string
+
+// NewKernelConfig simply creates the underlying map used by
+// the KernelConfig, you must use InitKernelConfig to read
+// in the values of the config
+func NewKernelConfig() KernelConfig {
```
yanivagman, line 20: Why not merging this function with InitKernelConfig? Is there a meaning to call this function without calling InitKernelConfig after?

libbpfgo/helpers/kernel_features.go
```diff
@@ -0,0 +1,110 @@
...
+	err := k.getBootConfig()
+	if err == nil {
+		return nil
+	}
+
+	err2 := k.getProcGZConfig()
+	if err != nil {
+		return fmt.Errorf("%v %v", err, err2)
+	}
+
```
yanivagman, line 39: Another option can be when CONFIG_IKHEADERS is defined. In that case, the headers can be found in sysfs

libbpfgo/helpers/kernel_features.go
```diff
@@ -0,0 +1,110 @@
...
+	return nil
+}
+
+// GetKernelConfigValue retrieves a value from the kernel config
+// If the config value does not exist an error will be returned
+func (k KernelConfig) GetKernelConfigValue(key string) (string, error) {
+	v, exists := k[key]
+	if !exists {
+		return "", errors.New("kernel config value does not exist, it is either not known by your kernel version or you have not called InitKernelConfig")
```
yanivagman, line 48: I think that instead of forcing the user to call InitKernelConfig, we should use "call once". By using it, InitKernelConfig (which can then be not exported) will be called only once at the first time this function is called.
(I remember that go supports that in some way, I can find you a reference if you want).

libbpfgo/helpers/kernel_features.go
```diff
@@ -0,0 +1,110 @@
...
+		return fmt.Errorf("could not open %s: %v", procConfigPath, err)
+	}
+
+	zreader, err := gzip.NewReader(configFile)
+	if err != nil {
+		return err
+	}
+
+	scanner := bufio.NewScanner(zreader)
+	for scanner.Scan() {
```
yanivagman, line 98: This for loop is dup of getBootConfig and may be called from a common function

libbpfgo/helpers/kernel_features.go
grantseltzer: I agree, meant to push this when I made the comment above.

libbpfgo/helpers/example_tracelisten_test.go
grantseltzer: This was a mistake, reverting to `helpers_test` package.

libbpfgo/helpers/kernel_features.go
grantseltzer: This is a good idea, handling this.

libbpfgo/helpers/kernel_features.go
grantseltzer: See my response to Itay above,  combining New/Init I think works best.

yanivagman: Golang once (from the sync package) is not the same as using init. It is only called the first time you call this getter, and not when the package is initialized

libbpfgo/helpers/kernel_features.go
yanivagman: Actually, I think that I confused you here. When CONFIG_IKHEADERS is set, the headers themselves are available in sysfs, not the config options set for the kernel

libbpfgo/helpers/kernel_features.go
grantseltzer: Oh interesting! Checking it out

simar7: > > this is great functionality to have in libbpfgo. May I suggest some UX improvements:
> > Instead of:
> > ```go
> > import github.com/.../libbpfgo/helpers
> > ...
> > helpers.NewKernelConfig()
> > helpers.InitKernelConfig()
> > conf := helpers.KernelConfig["something"]
> > ```
> > 
> > 
> > we can have:
> > ```go
> > import github.com/.../libbpfgo/kconfig
> > ...
> > conf, err := kconfig.Get("something)
> > ```
> > 
> > 
> > I don't see why a user would care to instantiate the map, then populate it manually every time they want to get a config. Instead we can just do all of that in the package's `init()`. WDYT?
> 
> That's how I originally designed it (with an init) but decided against it. I'm not sure _why_ this would happen, but I was worried about a situation where the config file is created after the init function is called. I suppose this could happen if the kernel is compiled with `CONFIG_IKCONFIG` but not `CONFIG_IKCONFIG_PROC` and the file is extracted to /proc after the fact (again, not sure why that'd happen). I'd say in general I found the design to be more dangerous for suppressing errors that can't be repaired.
> 
> I will say, there's no reason to have both a `NewKernelConfig` and an `InitKernelConfig`, I am going to put the logic for Init into New.

I would personally vote for a no init solution. While on the surface it seems appealing by the user having to do less setup work, it is a little [_magical_](https://peter.bourgon.org/blog/2017/06/09/theory-of-modern-go.html). 

How about the following:
```
var kc KernelConfig
conf, err := helpers.NewKernelConfig(kc)
...
val, err := conf. GetKernelConfigValue("some key")
```

This enables the user to inject their own KernelConfig type which gets modified by the constructor.

simar7:
> > this is great functionality to have in libbpfgo. May I suggest some UX improvements:
> > Instead of:
> > ```go
> > import github.com/.../libbpfgo/helpers
> > ...
> > helpers.NewKernelConfig()
> > helpers.InitKernelConfig()
> > conf := helpers.KernelConfig["something"]
> > ```
> > 
> > 
> > we can have:
> > ```go
> > import github.com/.../libbpfgo/kconfig
> > ...
> > conf, err := kconfig.Get("something)
> > ```
> > 
> > 
> > I don't see why a user would care to instantiate the map, then populate it manually every time they want to get a config. Instead we can just do all of that in the package's `init()`. WDYT?
> 
> That's how I originally designed it (with an init) but decided against it. I'm not sure _why_ this would happen, but I was worried about a situation where the config file is created after the init function is called. I suppose this could happen if the kernel is compiled with `CONFIG_IKCONFIG` but not `CONFIG_IKCONFIG_PROC` and the file is extracted to /proc after the fact (again, not sure why that'd happen). I'd say in general I found the design to be more dangerous for suppressing errors that can't be repaired.
> 
> I will say, there's no reason to have both a `NewKernelConfig` and an `InitKernelConfig`, I am going to put the logic for Init into New.

I would personally vote for a no init solution. While on the surface it seems appealing by the user having to do less setup work, it is a little [_magical_](https://peter.bourgon.org/blog/2017/06/09/theory-of-modern-go.html). 

How about the following:
```
var kc KernelConfig
conf, err := helpers.NewKernelConfig(kc)
...
val, err := conf. GetKernelConfigValue("some key")
```

This enables the user to inject their own KernelConfig type which gets modified by the constructor.

libbpfgo/helpers/kernel_features.go
grantseltzer: Ah, I thought you were talking about `CONFIG_IKCONFIG` and therefore suggesting I add functionality to do this:

https://stackoverflow.com/a/45654468

I think I need to explore that more and do it in a follow up PR if needed. It's not done in `bpftool feature`, let me think about why.

grantseltzer:
@simar7 I really like this suggestion and implemented it accordingly!

libbpfgo/helpers/kernel_features.go
grantseltzer: I feel this is clever but wouldn't be the expected behavior of someone using this API so it's best avoided. Having them just make one call to `InitKernelConfig` is acceptable now that I merged New and Init.

grantseltzer:
@yanivagman @simar7 If either of you have a few minutes today to look at this, I think it's good to go!

simar7:
> @yanivagman @simar7 If either of you have a few minutes today to look at this, I think it's good to go!

Will do later today!

libbpfgo/helpers/kernel_features.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("%v %v", err, err2)
+	}
+
+	return k, nil
+}
+
+// GetKernelConfigValue retrieves a value from the kernel config
+// If the config value does not exist an error will be returned
+func (k KernelConfig) GetKernelConfigValue(key string) (string, error) {
```
simar7, line 88: Since we can inject the dependency easily here, could you add a test for this function?

libbpfgo/helpers/kernel_features.go
```diff
@@ -0,0 +1,138 @@
...
+	v, exists := k[key]
+	if !exists {
+		return "", errors.New("kernel config value does not exist, it is could not be known by your kernel version")
```
simar7, line 91:
```suggestion
		return "", errors.New("kernel config value does not exist, it could not be determined by your kernel version")
```

libbpfgo/helpers/kernel_features.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		return fmt.Errorf("could not open %s: %v", bootConfigPath, err)
+	}
+
+	scanner := bufio.NewScanner(configFile)
+	k.readConfigFromScanner(scanner)
+	configFile.Close()
+
+	return nil
+}
```
simar7, line 108: ditto here for testing. Props for neat refactoring here, helps us test.

simar7 (approved):
:shipit: 🚀

PR #649, opened 2021-03-25
Fix eventsChannels race

### LINKED ISSUE #648: libbpfgo access global maps without synchronization
In order to access Go objects from a C callback, libbpfgo utilizes global maps that are keyed by a numeric id (bpf map fd).
These maps lack a mutex which is problematic because multiple polling goroutines access them in addition to client code. Potentially also the client code could be creating perf and ring buffers from multiple goroutines.

eyakubovich:
Fixes eventsChannels and lostChannels accesses
- Both were accessed without a mutex
- The acceeses in polling goroutine are frequent,
  locks are undesired in this path
- Use a sparse array (implemented as a two level tree)
  to provide lock-free read-only access.

Fixes #648

### DIFF
--- a/libbpfgo/helpers/rwArray.go
+++ b/libbpfgo/helpers/rwArray.go
@@ -1,0 +1,74 @@
+package helpers
+
+import (
+	"sync"
+)
+
+type slot struct {
+	value interface{}
+	used  bool
+}
+
+// RWArray allows for multiple concurrent readers but
+// only a single writer. The writers lock a mutex while the readers
+// are lock free.
+// It is implemented as an array of slots where each slot holds a
+// value (of type interface{}) and a boolean marker to indicate if it's
+// in use or not. The insertion (Put) performs a linear probe
+// looking for an available slot as indicated by the in-use marker.
+// While probing, it is not touching the value itself, as it's
+// being read without a lock by the readers.
+type RWArray struct {
+	slots []slot
+	mux   sync.Mutex
+}
+
+func NewRWArray(capacity uint) RWArray {
+	return RWArray{
+		slots: make([]slot, capacity),
+	}
+}
+
+func (a *RWArray) Put(v interface{}) int {
+	a.mux.Lock()
+	defer a.mux.Unlock()
+
+	limit := len(a.slots)
+
+	for i := 0; i < limit; i++ {
+		if !a.slots[i].used {
+			a.slots[i].value = v
+			a.slots[i].used = true
+			return i
+		}
+	}
+
+	return -1
+}
+
+func (a *RWArray) Remove(index uint) {
+	a.mux.Lock()
+	defer a.mux.Unlock()
+
+	if int(index) >= len(a.slots) {
+		return
+	}
+
+	a.slots[index].value = nil
+	a.slots[index].used = false
+}
+
+func (a *RWArray) Get(index uint) interface{} {
+	if int(index) >= len(a.slots) {
+		return nil
+	}
+
+	// N.B. If slot[index].used == false, this is technically
+	// a race since Put() might be putting the value in there
+	// at the same time.
+	return a.slots[index].value
+}
+
+func (a *RWArray) Capacity() uint {
+	return uint(len(a.slots))
+}
--- a/libbpfgo/libbpf_cb.go
+++ b/libbpfgo/libbpf_cb.go
@@ -10,19 +10,21 @@ import (
 
 //export perfCallback
 func perfCallback(ctx unsafe.Pointer, cpu C.int, data unsafe.Pointer, size C.int) {
-	eventChannels[uintptr(ctx)] <- C.GoBytes(data, size)
+	pb := eventChannels.Get(uint(uintptr(ctx))).(*PerfBuffer)
+	pb.eventsChan <- C.GoBytes(data, size)
 }
 
 //export perfLostCallback
 func perfLostCallback(ctx unsafe.Pointer, cpu C.int, cnt C.ulonglong) {
-	lostChan := lostChannels[uintptr(ctx)]
-	if lostChan != nil {
-		lostChan <- uint64(cnt)
+	pb := eventChannels.Get(uint(uintptr(ctx))).(*PerfBuffer)
+	if pb.lostChan != nil {
+		pb.lostChan <- uint64(cnt)
 	}
 }
 
 //export ringbufferCallback
 func ringbufferCallback(ctx unsafe.Pointer, data unsafe.Pointer, size C.int) C.int {
-	eventChannels[uintptr(ctx)] <- C.GoBytes(data, size)
+	ch := eventChannels.Get(uint(uintptr(ctx))).(chan []byte)
+	ch <- C.GoBytes(data, size)
 	return C.int(0)
 }
--- a/libbpfgo/libbpfgo.go
+++ b/libbpfgo/libbpfgo.go
@@ -5,6 +5,7 @@ package libbpfgo
 
 #include <bpf/bpf.h>
 #include <bpf/libbpf.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <sys/resource.h>
@@ -34,9 +35,8 @@ void set_print_fn() {
     libbpf_set_print(libbpf_print_fn);
 }
 
-struct ring_buffer * init_ring_buf(int map_fd) {
+struct ring_buffer * init_ring_buf(int map_fd, uintptr_t ctx) {
     struct ring_buffer *rb = NULL;
-    __u64 ctx = map_fd;
     rb = ring_buffer__new(map_fd, ringbufferCallback, (void*)ctx, NULL);
     if (!rb) {
         fprintf(stderr, "Failed to initialize ring buffer\n");
@@ -45,12 +45,11 @@ struct ring_buffer * init_ring_buf(int map_fd) {
     return rb;
 }
 
-struct perf_buffer * init_perf_buf(int map_fd, int page_cnt) {
+struct perf_buffer * init_perf_buf(int map_fd, int page_cnt, uintptr_t ctx) {
     struct perf_buffer_opts pb_opts = {};
     struct perf_buffer *pb = NULL;
     pb_opts.sample_cb = perfCallback;
     pb_opts.lost_cb = perfLostCallback;
-    __u64 ctx = map_fd;
     pb_opts.ctx = (void*)ctx;
     pb = perf_buffer__new(map_fd, page_cnt, &pb_opts);
     if (pb < 0) {
@@ -182,6 +181,13 @@ import (
 	"sync"
 	"syscall"
 	"unsafe"
+
+	"github.com/aquasecurity/tracee/libbpfgo/helpers"
+)
+
+const (
+	// Maximum number of channels (RingBuffers + PerfBuffers) supported
+	maxEventChannels = 512
 )
 
 type Module struct {
@@ -222,16 +228,20 @@ type BPFLink struct {
 }
 
 type PerfBuffer struct {
-	pb     *C.struct_perf_buffer
-	bpfMap *BPFMap
-	stop   chan struct{}
-	closed bool
-	wg     sync.WaitGroup
+	pb         *C.struct_perf_buffer
+	bpfMap     *BPFMap
+	slot       uint
+	eventsChan chan []byte
+	lostChan   chan uint64
+	stop       chan struct{}
+	closed     bool
+	wg         sync.WaitGroup
 }
 
 type RingBuffer struct {
 	rb     *C.struct_ring_buffer
 	bpfMap *BPFMap
+	slot   uint
 	stop   chan struct{}
 	closed bool
 	wg     sync.WaitGroup
@@ -621,8 +631,7 @@ func doAttachKprobeLegacy(prog *BPFProg, kp string, isKretprobe bool) (*BPFLink,
 	return bpfLink, nil
 }
 
-var eventChannels = make(map[uintptr]chan []byte)
-var lostChannels = make(map[uintptr]chan uint64)
+var eventChannels = helpers.NewRWArray(maxEventChannels)
 
 func (m *Module) InitRingBuf(mapName string, eventsChan chan []byte) (*RingBuffer, error) {
 	bpfMap, err := m.GetMap(mapName)
@@ -630,20 +639,24 @@ func (m *Module) InitRingBuf(mapName string, eventsChan chan []byte) (*RingBuffe
 		return nil, err
 	}
 
-	ctx := uintptr(bpfMap.fd)
 	if eventsChan == nil {
 		return nil, fmt.Errorf("events channel can not be nil")
 	}
-	eventChannels[ctx] = eventsChan
 
-	rb := C.init_ring_buf(bpfMap.fd)
+	slot := eventChannels.Put(eventsChan)
+	if slot == -1 {
+		return nil, fmt.Errorf("max ring buffers reached")
+	}
+
+	rb := C.init_ring_buf(bpfMap.fd, C.uintptr_t(slot))
 	if rb == nil {
-		return nil, fmt.Errorf("")
+		return nil, fmt.Errorf("failed to initialize ring buffer")
 	}
 
 	ringBuf := &RingBuffer{
 		rb:     rb,
 		bpfMap: bpfMap,
+		slot:   uint(slot),
 	}
 	m.ringBufs = append(m.ringBufs, ringBuf)
 	return ringBuf, nil
@@ -664,7 +677,7 @@ func (rb *RingBuffer) Stop() {
 		// may have stopped at this point. Failure to drain it will
 		// result in a deadlock: the channel will fill up and the poll
 		// goroutine will block in the callback.
-		eventChan := eventChannels[uintptr(rb.bpfMap.fd)]
+		eventChan := eventChannels.Get(rb.slot).(chan []byte)
 		go func() {
 			for range eventChan {
 			}
@@ -688,6 +701,7 @@ func (rb *RingBuffer) Close() {
 	}
 	rb.Stop()
 	C.ring_buffer__free(rb.rb)
+	eventChannels.Remove(rb.slot)
 	rb.closed = true
 }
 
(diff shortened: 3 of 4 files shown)


grantseltzer:
Thanks for keeping the contributions coming @eyakubovich! At first glance this looks great, just want to throw out there that i'll give this an actual review on Monday :-)

grantseltzer:
I'm trying to understand the motivation for having the RWArray data structure, what does having a two level tree to store references to the channels accomplish? `tracee-ebpf`, which granted is developed independently from `libbpfgo`, only uses two channels, would we ever expect a project to have 1024^2 ringbuffers? I don't think linux even supports that many open file descriptors. 

I need to better understand the code for this PR itself, but for now could you tell me what this achieves that putting a mutex [here](https://github.com/aquasecurity/tracee/blob/main/libbpfgo/libbpfgo.go#L637) wouldn't do?

eyakubovich:
This can certainly be done with just a mutex. However it won't be enough to put it just there, it would also need to be put in the callback, e.g. https://github.com/aquasecurity/tracee/blob/main/libbpfgo/libbpf_cb.go#L26 . Unless I'm misunderstanding something. The callback can be in the hot path so that's why I was trying to avoid a mutex. If there's more than one ring/perf buffer, it could create a contention point. But the hot path is also all read-only so it was a good candidate to use an array.

As for the array capacity (1M) and the use of two-levels, it was probably an overkill. I was paranoid about putting too low of a limit but also didn't want to allocate a ton of memory if most of the time it'll be just a handful. For our use case, it's also just a few ring buffers so can certainly limit it to 1024 and use a single level. Will simplify the code.

eyakubovich:
Also, the kernel can support quite a bit of open handles. The system wide limit defaults to 1M: https://elixir.bootlin.com/linux/v5.11.10/source/fs/file.c#L27

yanivagman (changes requested):
Thanks @eyakubovich,

I wrote some comments about the rwArray data structure.

Other than that, I'm not sure that I fully understand the problem that you are trying to solve. Let's take the conventional use case where the user of libbpfgo initializes several perf/ring buffers on his program start, and poll them from different go routines during the program's runtime. In that case, we only write to the global maps during initialization, read-only during runtime, and delete on exit. If any buffer has its own key in the global hash map (the fd of the perf/ring map), no writing occurs to these maps, and these maps are only internal to libbpfgo - where is the contention?

From my understanding, there are three cases which might be problematic, but I don't think that any of them is a real problem or requires the use of the rwArray data structure:
1. Initializing/removing buffers during runtime so the global map is written while other buffers may still be polled. In this case, the key used to add/remove the buffer to the hash map is different, which shouldn't affect any of the existing buffers in the map
2. Several go routines poll from the same buffer. I'm not sure why would anyone do that, and even if there is a reason to do that - we can enforce just one reader from a given buffer.

eyakubovich:
@yanivagman Go's `map` is not thread safe. Therefore if one goroutine reads from it and another writes, it needs a mutex around it. Take a typical startup sequence that uses two ring buffers:

```
ringA := module.InitRingBuffer("ring_a", chanA)
ringA.Start()

ringB := module.InitRingBuffer("ring_b", chanB)
ringB.Start()
```

By the time we get to initialization of ring B, the polling goroutine for ring A is already running. It is therefore reading the `eventChannels` map. Since `InitRingBuffer` inserts into the same global map, there's a race.

This can be mitigated by first initializing the rings and then starting them:
```
ringA := module.InitRingBuffer("ring_a", chanA)
ringB := module.InitRingBuffer("ring_b", chanB)

ringA.Start()
ringB.Start()
```

However this can be problematic in practice where InitRingBuffer/Start are packaged together in some other unit. Of course we can use this mitigation, if need be.

I think in general when there are global variables involved in a package, it's best to make access to them thread-safe.

libbpfgo/helpers/rwArray.go
```diff
@@ -0,0 +1,111 @@
...
+		chunks: make([]chunk, numChunks),
+	}
+}
+
+func (a *RWArray) Put(v interface{}) int {
+	a.mux.Lock()
+	defer a.mux.Unlock()
+
+	for i := 0; i < numChunks; i++ {
+		if j := a.chunks[i].put(v); j != -1 {
```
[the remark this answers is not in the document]
eyakubovich: chunk's `put` has
```
	if c.used == slotsPerChunk {
		return -1
	}
```
to keep it O(numChunks).

libbpfgo/helpers/rwArray.go
```diff
@@ -0,0 +1,111 @@
...
+
+type chunk struct {
+	slots []slot
+	used uint
+}
+
+func (c *chunk) put(v interface{}) int {
+	if c.used == slotsPerChunk {
+		return -1
+	} else if c.used == 0 && c.slots == nil {
```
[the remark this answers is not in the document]
eyakubovich: I need to comment more as this is subtle. Since the `Get` path accesses `c.slots` without holding a lock, I need to make sure that `c.used == 0` before touching `c.slots`. I'm relying on Go's short-circuiting behavior here. `c.used == 0` guarantees that no readers exist.

libbpfgo/helpers/rwArray.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	return -1
+}
+
+func (a *RWArray) Remove(index uint) {
+	a.mux.Lock()
+	defer a.mux.Unlock()
+
+	a.chunks[index / slotsPerChunk].free(index % slotsPerChunk)
```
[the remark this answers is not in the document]
eyakubovich: Will add but should it be a nop, panic or `error` if it's out-of-bounds or unused?

libbpfgo/helpers/rwArray.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func (a *RWArray) Get(index uint) interface{} {
+	chunkIdx := index / slotsPerChunk
+	slotIdx := index % slotsPerChunk
+
+	return a.chunks[chunkIdx].slots[slotIdx].value
```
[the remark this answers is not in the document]
eyakubovich: Will add a check

libbpfgo/helpers/rwArray.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+	}
+
+	panic("used tracking is wrong")
+}
+
+func (c *chunk) free(i uint) {
+	c.slots[i].value = nil
+	c.slots[i].used = false
+	c.used--
```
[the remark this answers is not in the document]
eyakubovich: `chunk` is internal -- I tend to avoid error checking on non-public interfaces. However I can add these sanity checks.

libbpfgo/helpers/rwArray.go
```diff
@@ -0,0 +1,111 @@
...
+}
```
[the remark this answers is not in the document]
eyakubovich: Agreed.

libbpfgo/libbpfgo.go
```diff
@@ -34,9 +34,8 @@ void set_print_fn() {
     libbpf_set_print(libbpf_print_fn);
 }
 
-struct ring_buffer * init_ring_buf(int map_fd) {
+struct ring_buffer * init_ring_buf(int map_fd, uintptr_t ctx) {
```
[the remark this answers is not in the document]
eyakubovich: Good catch. It's defined in `stdint.h` but that file is not directly included. I'm sure it's included numerous times from the other includes. I'll add the include.

libbpfgo/helpers/rwArray.go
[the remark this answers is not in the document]
yanivagman: I think that nop should be ok

libbpfgo/helpers/rwArray.go
```diff
@@ -0,0 +1,136 @@
...
+	if chunkIdx >= numChunks {
+		return nil
+	}
+
+	c := &a.chunks[chunkIdx]
+	if c.slots == nil {
+		return nil
+	}
+
+	return c.slots[slotIdx].value
```
yanivagman, line 86: slotIdx might be out of bounds.
also, should check that not unused

eyakubovich:
Our use case also creates all the ring buffers at startup and destroys on exit. We use a couple of buffers in separate places so the two phase Init/Start is possible but a bit awkward. If we think more abstractly, an app might not have the luxury of creating everything on startup. Some admin command (e.g. SIGHUP to re-read a config) could force it to create additional channels (or destroy existing) while others are running.

I've changed the code to be a single level. You're right that it makes it simpler, less bug prone and faster. It's probably now faster than using a map -- it just reads a slice (a single memory load) instead of going through a hash table. The `RWArray` can now be sized on construction. `libbpfgo.go` initializes it to 512 elements (the size of the slot should be 16 bytes, 4096 bytes total -- that's one page on x86).

libbpfgo/helpers/rwArray.go
```diff
@@ -0,0 +1,76 @@
...
+
+	for i := 0; i < limit; i++ {
+		if !a.slots[i].used {
+			a.slots[i].value = v
+			a.slots[i].used = true
+			return i
+		}
+	}
+
+	return -1
```
yanivagman, line 43: Now that you use slice - why not letting go handle the capacity of the slice?
We can have a default capacity (say, 512), and if no space left - just append one more slot to the array, which will double its capacity. WDYT?

libbpfgo/helpers/rwArray.go
```diff
@@ -0,0 +1,76 @@
...
+	// a race since Put() might be putting the value in there
+	// at the same time.
+	return a.slots[index].value
+}
+
+func (a *RWArray) Capacity() uint {
+	return uint(len(a.slots))
+}
+
+type slot struct {
```
yanivagman, line 73: Just for readability, now that this struct doesn't have any associated functions, it can be placed right before the RWArray struct

libbpfgo/libbpfgo.go
```diff
@@ -688,6 +703,7 @@ func (rb *RingBuffer) Close() {
 	}
 	rb.Stop()
 	C.ring_buffer__free(rb.rb)
+	eventChannels.Remove(rb.slot)
```
yanivagman, line 706: Didn't `rb.Stop()` (two lines above) take care of this already?

libbpfgo/libbpfgo.go
```diff
@@ -790,6 +814,7 @@ func (pb *PerfBuffer) Close() {
 	}
 	pb.Stop()
 	C.perf_buffer__free(pb.pb)
+	eventChannels.Remove(pb.slot)
```
yanivagman, line 817: Didn't `rb.Stop()` take care of this already?

yanivagman (commented):
Thanks @eyakubovich,
It looks much simpler now

libbpfgo/libbpfgo.go
eyakubovich: Great catch! It does but it needs to be done here since it was inserted in Init. Otherwise someone can call Stop() and then Start() and it'll break.

libbpfgo/helpers/rwArray.go
eyakubovich: Unfortunately we can't reallocate since `Get` does not hold a lock.

libbpfgo/libbpfgo.go
eyakubovich: I've removed the one in Stop

libbpfgo/libbpfgo.go
eyakubovich: I've removed the one in Stop

yanivagman (approved):
LGTM!

libbpfgo/helpers/rwArray.go
```diff
@@ -0,0 +1,76 @@
+package helpers
+
+import (
+	"sync"
+)
+
+// Like RWMutex, RWArray allows for multiple concurrent readers but
```
grantseltzer, line 7: Can you format this comment to be in GoDoc format (i.e. directly above the RWArray definition, and starting with "RWArray ..."

libbpfgo/libbpf_cb.go
```diff
@@ -10,19 +10,21 @@ import (
...
+	if pb.lostChan != nil {
+		pb.lostChan <- uint64(cnt)
 	}
 }
 
 //export ringbufferCallback
 func ringbufferCallback(ctx unsafe.Pointer, data unsafe.Pointer, size C.int) C.int {
-	eventChannels[uintptr(ctx)] <- C.GoBytes(data, size)
+	ch := eventChannels.Get(uint(uintptr(ctx))).(chan []byte)
+	ch <- C.GoBytes(data, size)
```
grantseltzer, line 28: Apologies for the confusion, I know i'm likely repeating a question that Yaniv asked but I want to really make sure I understand this. 

Is the whole need for this patch because accessing of the `eventsChannels` map could happen while another ring/perfbuffer is added? If so, wouldn't the newly created ring/perfbuffer not be written as the value of an existing key since the fd will always be unique? Thus avoiding the possibility of this race?

My main point of concern is the above lines in the callback functions. Casting the data as a `*PerfBuffer`/`chan []byte` seems like an expensive operation to be doing every time an event is received. 

We unfortunately don't yet have a robust benchmarking pipeline set up, though we should somewhat soon. I'd feel a lot more comfortable with this change if we could at least use a [go test benchmark](https://golang.org/pkg/testing/#hdr-Benchmarks) to prove that. Does this sound reasonable?

eyakubovich: Yes, it's because accessing `eventsChannels` map could happen while an insert or delete are happening (when a perf/ring buffer is created or destroyed). The fd's are going to be distinct, of course, but the map (hash table) is the same. It could get re-hashed as a result of an insert. The language specifies it as not thread-safe in this case.

The cast is more of a compile time operation. The type assertion compiles to a compare, a conditional jump and a move. It's just an `if` statement and it'll get very well predicted (since it's a type check). I'm sure the hash table this is replacing has more than a single branch. Not sure if you like reading assembly but I set up a mock to demonstrate: https://godbolt.org/z/rrjxaadv3

I can try to do a go test benchmark if you still have doubts.

grantseltzer: Ah I believe I follow, thanks for that clarification! I did look at the disassembly before my last comment but didn't find it to be particularly conclusive.

I just read [this article](https://hackernoon.com/some-insights-on-maps-in-golang-rm5v3ywh) which clearly things up greatly for me.

Would this design accomplish the same thing as using a [sync.Map](https://golang.org/pkg/sync/#Map)?

Also I think going back and creating test benchmarks for a ton of libbpfgo functionality would be great for us to do sometime soon but it's fine for now.

eyakubovich: I forgot about `sync.Map` (I remember seeing it many years ago). Yes, we could just use that and not worry about the capacity limits. I looked at the implementation and since it's more general, it will be a bit slower than what I have but I'm not sure it'll be measurable. Compared to reading data off of perf/ring buffers and doing a channel send, this overhead is probably less than the noise.

grantseltzer: So do you think we should use the sync.Map instead? I'll rely on your expertise here.

libbpfgo/helpers/rwArray.go
```diff
@@ -0,0 +1,74 @@
...
+// are lock free.
+// It is implemented as an array of slots where each slot holds a
+// value (of type interface{}) and a boolean marker to indicate if it's
+// in use or not. The insertion (Put) performs a linear probe
+// looking for an available slot as indicated by the in-use marker.
+// While probing, it is not touching the value itself, as it's
+// being read without a lock by the readers.
+type RWArray struct {
+	slots []slot
+	mux   sync.Mutex
```
simar7, line 23: nit: Conventionally mutexes are named as `mu`. Routers are `mux`.
```suggestion
	mu   sync.Mutex
```

libbpfgo/libbpf_cb.go
simar7: re: `sync.Map`: Stability should be the deciding factor when it comes to using it. Can we ensure this map won't grow overtime? IOW, would it be _stable_? There are other factors besides performance like losing compile time type checking by using sync.Map vs a regular mutex+map. I'm not sure if the tradeoff is worth it.

I recommend reading this prior to going down that path: https://github.com/gophercon/2017-talks/blob/master/lightningtalks/BryanCMills-AnOverviewOfSyncMap/An%20Overview%20of%20sync.Map.pdf

grantseltzer: Thanks for the link to that Simar. My instincts tell me i'd rather tell consumers of libbpfgo to create all perf/ringbuffers up from before accessing (which tracee-ebpf does) than take a performance hit here. I am going to put together some benchmarks to confirm whether or not i'm being paranoid.

libbpfgo/helpers/rwArray.go
eyakubovich: Good point. C++ habit 😀

eyakubovich:
I’ll think a bit more about the options. I’m also fine with refactoring our code to do two phase initialization

eyakubovich:
There's another option to consider (similar to `sync.Map`) -- every time you make a change to the map, you make a copy and then atomically publish it. This is basically RCU but relies on GC for cleanup. Here are some benchmarks I ran on the options:

This is lookup only. The first test uses just a regular Go `map` -- this is what's currently used.
```
BenchmarkMap-6       	14687455	        87.42 ns/op
BenchmarkRWArray-6   	73920867	        15.53 ns/op
BenchmarkSyncMap-6   	 4364822	       252.5 ns/op
BenchmarkRCUMap-6    	12471579	        95.72 ns/op
```

As you can see, RCUMap incurs a bit of overhead due to two extra pointer dereferences. I used `sync.Value` out of laziness, this can be cut down to one extra dereference with `unsafe.Pointer`.

In this test, I added a channel send and receive to better simulate the whole callback. As you can see, the lookup cost gets amortized:

```
BenchmarkMap-6       	 2183679	       510.7 ns/op
BenchmarkRWArray-6   	 2351884	       479.8 ns/op
BenchmarkSyncMap-6   	 1800895	       633.0 ns/op
BenchmarkRCUMap-6    	 2198708	       533.8 ns/op
```

grantseltzer:
Thanks for your patience and contribution @eyakubovich!

Issue #637, opened 2021-03-23
Create architecture diagram of the interaction between tracee-ebpf and tracee-rules


grantseltzer:
We need more documentation explaining how the different pieces interact with one another. It'd be good to have a reference-able blog post about it as well.

simar7:
Recommend: https://excalidraw.com

simar7:
Are you working on this @grantseltzer? - if not I can tackle it today.

grantseltzer:
@simar7 I was going to but wouldn't be getting to it until early next week, go for it!

simar7:
@grantseltzer I added a high level overview. Usually for inter software communication I would've used a ladder diagram but as I was drafting it I realized there isn't a whole lot of back-and-forth between tracee-ebpf and tracee-rules (in fact the latter is just a pure consumer of the former).

PR #647, opened 2021-03-25
revamp documentation

itaysk:
rearranged all the documentation into the /docs folder and simplified the docs site structure.

the idea is that the readme will contain just a simple intro to the project, and the full docs will be separate (like in Trivy/Starboard).

PR #650, opened 2021-03-26
feat: Add high level overview to Readme

simar7:
Fixes: https://github.com/aquasecurity/tracee/issues/637


Readme.md
```diff
@@ -17,6 +17,7 @@ Tracee is composed of the following sub-projects:
 - [libbpgo](libbpfgo) - Go library for eBPF programming using Linux's [libbpf](https://github.com/libbpf/libbpf)
 
 ## Getting started
+![High Level Overview](images/highleveloverview.png)
```
itaysk, line 20: I wouldn't put this under getting started. The end user shouldn't even be aware of the intricacies of the internal components. In the new readme (#647 ) this would fit under the "components" section

grantseltzer:
I need to try excalidraw, diagram looks great! My only comment is the "event detected" by (3). If I understand correctly all of the events are fed to tracee-rules in which case I feel 'event detected' makes more sense to be inside of tracee-rules's box.

Perhaps a better thing to put there is "flow of events created by tracee-ebpf"? And the 'events sent' from (2) should be "events created"?

Readme.md
itaysk: Actually, perhaps it would be beneficial to have an architecture.md file, for the sake of introducing contributors to tracee (as opposed to users which have the docs site). at first this doc can just contain this diagram, perhaps in the future we will add content

itaysk: https://matklad.github.io//2021/02/06/ARCHITECTURE.md.html

simar7: Yeah sure I'm OK with that. I can change it live in architecture.md instead.

simar7:
> I need to try excalidraw, diagram looks great! My only comment is the "event detected" by (3). If I understand correctly all of the events are fed to tracee-rules in which case I feel 'event detected' makes more sense to be inside of tracee-rules's box.
> 
> Perhaps a better thing to put there is "flow of events created by tracee-ebpf"? And the 'events sent' from (2) should be "events created"?

Right, how about I change it to: 
(3) events collected
(4) finding detected

itaysk (approved):
The title over the tracee-rules block should say detection, imo. Otherwise lgtm

Issue #627, opened 2021-03-17
remove the falcosidekick struct from the webhook code


itaysk:
remove the falcosidekick struct from the webhook code, and replace it with a gotemplate
depands on #612

PR #653, opened 2021-03-26
refactor: Remove falcosidekick specific code and reuse templating

simar7:
This PR removes Falcosidekick specific code and reuses the templating logic to build the same.

Fixes https://github.com/aquasecurity/tracee/issues/627

### DIFF
--- a/tracee-rules/output.go
+++ b/tracee-rules/output.go
@@ -2,7 +2,7 @@ package main
 
 import (
 	"bytes"
-	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"log"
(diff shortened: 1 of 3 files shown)


itaysk (approved):

itaysk:
LGTM but I'm trying to understand why tests failed.
In the previous PR to this area the same tests failed but GH didn't show the workflow as failed https://github.com/aquasecurity/tracee/pull/646/checks#step:7:148

simar7:
> LGTM but I'm trying to understand why tests failed.
> In the previous PR to this area the same tests failed but GH didn't show the workflow as failed https://github.com/aquasecurity/tracee/pull/646/checks#step:7:148
> I think it's ok to merge this PR (just because the previous one failed as well), and immediately investigate the failure

Actually tests only failed in this PR (which is an oversight on my part to not run all of them prior to pushing). The run you have linked has all tests passing. The red warnings you see in that run are coming from rego based tests and are not actual errors. GitHub UI just shows them like that. We could probably do something to clear that noise.

Issue #652, opened 2021-03-26
Fix documentation content issues


itaysk:
- [x] make sure links are not broken, both in docs and readmes, change to relative links
- [x] Headers should match navigation
- [x] tables are not formatted right (probably different markdown parser)
- [x] small typos
- [ ] beef up content is possible (remember those docs are user facing, not contributor facing)

grantseltzer:
Using this to take notes on observations:

- Add libbpf to prerequisites
- Logo in top left is messed up
-

grantseltzer:
per @yanivagman:

- Links are broken
- No one-liner of tracee-ebpf
- Even if choosing to only keep the main readme tracee one-liner, the is no "-it" for interactive mode, which enables the use of ctrl+c
- No links from tracee-ebpf readme to the docs (capture, secure tracing, and other stuff)

itaysk:
fixed in: #673

PR #657, opened 2021-03-30
Small fixes for documentation

grantseltzer:
This handles #652


docs/config.md
```diff
@@ -1,3 +1,3 @@
 # Configuration Options
 
-Rules are discovered from the local `rules` directory (unless changed by the `--rules-dir` flag). By default, all discovered rules will be loaded unless specific rules are selected using the `--rules` flag.
\ No newline at end of file
+Rules are, by default, discovered from the `rules` directory next to the `tracee-rules` executable binary (you can specify a different location with the `--rules-dir` flag). By default, all discovered rules will be loaded unless specific rules are selected using the `--rules` flag.
```
simar7, line 3: avoid comma splice 
```suggestion
By default, rules are discovered from the `rules` directory next to the `tracee-rules` executable binary (you can specify a different location with the `--rules-dir` flag). By default, all discovered rules will be loaded unless specific rules are selected using the `--rules` flag.
```

mkdocs.yml
```diff
@@ -14,19 +14,20 @@ nav:
...
     - Output Options: tracee-ebpf/output.md
-    - Trace Options: tracee-ebpf/trace.md
+    - Trace Options: tracee-ebpf/trace-options.md
     - Capturing Artifacts: tracee-ebpf/capture.md
 
 theme:
     name: material
     language: 'en'
-    logo: images/tracee.png
+    logo: images/tracee_logo_only.png
```
simar7, line 30: Seems like we have two asset sources now, one under `docs/images/` and one under `images/` could we reconcile them somehow?

grantseltzer: This PR removes the `images/` directory for the base of the repo completely. This particular line is referencing the one in `docs/images/`

simar7 (approved):

docs/install/prerequisites.md
```diff
@@ -1,19 +1,19 @@
 # Prerequisites
 
 - Linux kernel version >= 4.18
-- Linux kernel headers available under conventional location (see [Linux Headers](#TODO) section for more info)
-- libc, and the libraries: libelf, zlib
```
itaysk, line 5: libbpf is not a prerequisite

grantseltzer: @itaysk are you saying that because we package libbpf source code with this repo?

Would all these library prerequisites be eliminated if we statically compile our releases? And headers once I finish with CO:RE?

itaysk: if someone wants to run tracee, they can download the binary, and run it, they don't need to have libbpf in order for tracee to run, right? the same can't be said for libelf, zlib which are required at runtime.

Yes I think if we statically compile tracee those reqs would be eliminated

Issue #661, opened 2021-03-31
Load kernel configuration options into a shared BPF_HASH map


grantseltzer:
In order to feature flag functionality in BPF code we can use the new feature check functions in `helpers` to populate a map with all kernel configuration. Then we can put checks in tracee.bpf.c for different kernel features.

PR #670, opened 2021-04-02
Load kernel config into bpf hashmap

### LINKED ISSUE #654: Add checks for kernel configuration options
Once #645 merges we will be able to check kernel configuration options. tracee-ebpf should fail if configuration options like `CONFIG_BPF` aren't set, but i'm not sure what the behavior is like.

We should check the configuration options and exit if the following ones aren't set on (or to `y`):

- CONFIG_BPF
- CONFIG_BPF_SYSCALL
- CONFIG_KPROBE_EVENTS
- CONFIG_BPF_EVENTS
- CONFIG_TRACING
- CONFIG_FTRACE_SYSCALLS

Not sure if we need: 
- CONFIG_HAVE_EBPF_JIT 
- CONFIG_BPF_JIT 
- CONFIG_BPF_JIT_ALWAYS_ON

Need for specific features:
- CONFIG_UPROBE_EVENTS
- CONFIG_DEBUG_INFO_BTF

grantseltzer:
closes #654 and #661 

Also adds unit tests to github actions

### DIFF
--- a/tracee-ebpf/main.go
+++ b/tracee-ebpf/main.go
@@ -14,6 +14,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/aquasecurity/tracee/libbpfgo/helpers"
 	"github.com/aquasecurity/tracee/tracee-ebpf/tracee"
 	"github.com/syndtr/gocapability/capability"
 	cli "github.com/urfave/cli/v2"
@@ -70,6 +71,10 @@ func main() {
 			cfg.BPFObjPath = bpfFile
 			if !checkRequiredCapabilities() {
 				return fmt.Errorf("insufficient privileges to run")
+			}
+			missingKernelOptions := missingKernelConfigOptions()
+			if len(missingKernelOptions) != 0 {
+				return fmt.Errorf("kernel is not properly configured, missing: %v", missingKernelOptions)
 			}
 			t, err := tracee.New(cfg)
 			if err != nil {
@@ -883,6 +888,33 @@ func checkRequiredCapabilities() bool {
 		return false
 	}
 	return caps.Get(capability.EFFECTIVE, capability.CAP_SYS_ADMIN)
+}
+
+func missingKernelConfigOptions() []string {
+
+	requiredConfigOptions := map[uint32]string{
+		helpers.CONFIG_BPF:           "y",
+		helpers.CONFIG_BPF_SYSCALL:   "y",
+		helpers.CONFIG_KPROBE_EVENTS: "y",
+		helpers.CONFIG_BPF_EVENTS:    "y",
+	}
+
+	kConfig := helpers.KernelConfig{}
+	kConfig.InitKernelConfig()
+
+	missing := []string{}
+
+	for requiredOption, setting := range requiredConfigOptions {
+		opt, err := kConfig.GetKernelConfigValue(requiredOption)
+		if err != nil {
+			missing = append(missing, helpers.KernelConfigKeyIDToString[requiredOption])
+			continue
+		}
+		if opt != setting {
+			missing = append(missing, helpers.KernelConfigKeyIDToString[requiredOption])
+		}
+	}
+	return missing
 }
 
 func getSelfCapabilities() (capability.Capabilities, error) {
--- a/tracee-ebpf/tracee/tracee.bpf.c
+++ b/tracee-ebpf/tracee/tracee.bpf.c
@@ -177,6 +177,45 @@
 
 #define DEV_NULL_STR    0
 
+#define KERNEL_CONFIG_BPF                       1
+#define KERNEL_CONFIG_BPF_SYSCALL               2
+#define KERNEL_CONFIG_HAVE_EBPF_JIT             3
+#define KERNEL_CONFIG_BPF_JIT                   4
+#define KERNEL_CONFIG_BPF_JIT_ALWAYS_ON         5
+#define KERNEL_CONFIG_CGROUPS                   6
+#define KERNEL_CONFIG_CGROUP_BPF                7
+#define KERNEL_CONFIG_CGROUP_NET_CLASSID        8
+#define KERNEL_CONFIG_SOCK_CGROUP_DATA          9
+#define KERNEL_CONFIG_BPF_EVENTS                10
+#define KERNEL_CONFIG_KPROBE_EVENTS             11
+#define KERNEL_CONFIG_UPROBE_EVENTS             12
+#define KERNEL_CONFIG_TRACING                   13
+#define KERNEL_CONFIG_FTRACE_SYSCALLS           14
+#define KERNEL_CONFIG_FUNCTION_ERROR_INJECTION  15
+#define KERNEL_CONFIG_BPF_KPROBE_OVERRIDE       16
+#define KERNEL_CONFIG_NET                       17
+#define KERNEL_CONFIG_XDP_SOCKETS               18
+#define KERNEL_CONFIG_LWTUNNEL_BPF              19
+#define KERNEL_CONFIG_NET_ACT_BPF               20
+#define KERNEL_CONFIG_NET_CLS_BPF               21
+#define KERNEL_CONFIG_NET_CLS_ACT               22
+#define KERNEL_CONFIG_NET_SCH_INGRESS           23
+#define KERNEL_CONFIG_XFRM                      24
+#define KERNEL_CONFIG_IP_ROUTE_CLASSID          25
+#define KERNEL_CONFIG_IPV6_SEG6_BPF             26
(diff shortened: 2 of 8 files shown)


grantseltzer:
Build breaks because I have to update the libbpfgo module version. Maybe we should use this as an opportunity to separate out into its own repo.

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -49,16 +49,16 @@
...
+#define MAX_PERCPU_BUFSIZE      (1 << 15)     // This value is actually set by the kernel as an upper bound
+#define MAX_STRING_SIZE         4096          // Choosing this value to be the same as PATH_MAX
+#define MAX_BYTES_ARR_SIZE      4096          // Max size of bytes array, arbitrarily chosen
+#define MAX_STR_ARR_ELEM        40            // String array elements number should be bounded due to instructions limit
+#define MAX_PATH_PREF_SIZE      64            // Max path prefix should be bounded due to instructions limit
+#define MAX_STACK_ADDRESSES     1024          // Max amount of different stack trace addresses to buffer in the Map
+#define MAX_STACK_DEPTH         20            // Max depth of each stack trace to track
+#define MAX_STR_FILTER_SIZE     16            // Max string filter size should be bounded to the size of the compared values (comm, uts)
+#define FILE_MAGIC_HDR_SIZE     16            // Number of bytes to save from a file's header (for magic_write event)
+#define FILE_MAGIC_MASK         15            // Mask used to pass verifier when submitting magic_write event bytes
```
yanivagman, line 61: Any reason to change this identation?

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -177,6 +177,45 @@
...
+#define KERNEL_CONFIG_NETFILTER_XT_MATCH_BPF    29
+#define KERNEL_CONFIG_BPFILTER                  30
+#define KERNEL_CONFIG_BPFILTER_UMH              31
+#define KERNEL_CONFIG_TEST_BPF                  32
+#define KERNEL_CONFIG_HZ                        33
+#define KERNEL_CONFIG_DEBUG_INFO_BTF            34
+#define KERNEL_CONFIG_DEBUG_INFO_BTF_MODULES    35
+#define KERNEL_CONFIG_BPF_LSM                   36
+#define KERNEL_CONFIG_BPF_PRELOAD               37
+#define KERNEL_CONFIG_BPF_PRELOAD_UMD           38
```
yanivagman, line 217: I think we can leave this out for now, and add new defines when needed

tracee-ebpf/tracee/tracee.go
```diff
@@ -713,6 +713,18 @@ func (t *Tracee) populateBPFMaps() error {
 	stringStoreMap, _ := t.bpfModule.GetMap("string_store")
 	stringStoreMap.Update(uint32(0), []byte("/dev/null"))
 
+	kernelConfig := helpers.KernelConfig{}
+	kconf, err := helpers.InitKernelConfig(kernelConfig)
+	if err != nil {
+		return fmt.Errorf("could not read in kernel configuration %v", err)
```
yanivagman, line 719: Do we really want to fail tracee in this case? Maybe just report to the user and continue - things may still work even if we fail to read the config

tracee-ebpf/main.go
```diff
@@ -872,6 +877,38 @@ func checkRequiredCapabilities() bool {
...
+		helpers.CONFIG_BPF_EVENTS:        "y",
+		helpers.CONFIG_TRACING:           "y",
+		helpers.CONFIG_FTRACE_SYSCALLS:   "y",
+		helpers.CONFIG_HAVE_EBPF_JIT:     "y",
+		helpers.CONFIG_BPF_JIT:           "y",
+		helpers.CONFIG_BPF_JIT_ALWAYS_ON: "y",
+	}
+
+	kConfig := helpers.KernelConfig{}
+	helpers.InitKernelConfig(kConfig)
```
yanivagman, line 895: Shouldn't you use the returned kernelconfig value and check for error like you do in populateBPFMaps()?

tracee-ebpf/tracee/tracee.go
```diff
@@ -713,6 +713,18 @@ func (t *Tracee) populateBPFMaps() error {
...
+	}
+
+	kernelConfigMap, _ := t.bpfModule.GetMap("kernel_config_map")
+	v, err := kconf.GetKernelConfigValue(helpers.CONFIG_BPF_SYSCALL)
+	if err == nil && v == "y" {
+		kernelConfigMap.Update(helpers.CONFIG_BPF_SYSCALL, uint32(1))
```
yanivagman, line 725: Do we have a reason to populate CONFIG_BPF_SYSCALL? It should always be set in our case

yanivagman (commented):
Nice!

tracee-ebpf/main.go
```diff
@@ -872,6 +877,38 @@ func checkRequiredCapabilities() bool {
...
+	requiredConfigOptions := map[uint32]string{
+		helpers.CONFIG_BPF:               "y",
+		helpers.CONFIG_BPF_SYSCALL:       "y",
+		helpers.CONFIG_KPROBE_EVENTS:     "y",
+		helpers.CONFIG_BPF_EVENTS:        "y",
+		helpers.CONFIG_TRACING:           "y",
+		helpers.CONFIG_FTRACE_SYSCALLS:   "y",
+		helpers.CONFIG_HAVE_EBPF_JIT:     "y",
+		helpers.CONFIG_BPF_JIT:           "y",
+		helpers.CONFIG_BPF_JIT_ALWAYS_ON: "y",
```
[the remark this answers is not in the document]
grantseltzer: I suppose we don't need any of the JIT configurations? Do I understand correctly that the JIT compiler is used as a fallback if the normal clang compilation isn't available?

tracee-ebpf/tracee/tracee.bpf.c
grantseltzer: No, i'm not actually sure how this happened? I'm reverting

tracee-ebpf/tracee/tracee.bpf.c
grantseltzer: I think it's warranted to have them in libbpfgo, and since the enum values are already set there, they should exist here.

tracee-ebpf/tracee/tracee.go
grantseltzer: Good point

tracee-ebpf/tracee/tracee.go
grantseltzer: I had this somewhat for testing purposes, i'm going to change this to `KERNEL_CONFIG_DEBUG_INFO_BTF` as that's the most relevant feature (will be using this in follow up PR).

grantseltzer:
Need to merge #678 before this in order to fix build.

tracee-ebpf/main.go
grantseltzer: Sloppy by me not catching this when I refactored. InitKernelConfig doesn't need to return the KernelConfig, no reason for this not to be a method.

grantseltzer:
@yanivagman This includes another commit that upgrades the libbpfgo module. This was my mistake to make changes to libbpfgo and tracee-ebpf in the same PR. Going forward we should never do this, and we should force tracee-ebpf to only import set releases of libbpfgo.

grantseltzer:
alright build keeps failing for a reason unrelated to this PR on the self hosted runner... debugging right now. 

Something to do with the location of the kernel headers on centos vs ubuntu/debian.

tracee-ebpf/tracee/tracee.bpf.c
yanivagman: Why should they exist here if they are not being used?

grantseltzer: To reserve the numbers since they must correspond to the ones in libbpfgo anyway.

yanivagman: But why should we reserve them? They will never be used in the BPF context. It is just a "define" directive which can be added when needed

grantseltzer: My thinking is that we want to add any possible constants to libbpfgo because people might want to use them even if tracee-ebpf doesn't. Since the constants are already defined in libbpfgo I felt that it would be a good idea to just have constants defined for simplifying future development. 

However upon thinking about it I realize that this isn't even the proper way of handling feature flagging in BPF code. In the case of CO:RE stuff, I'm using a conditional compile flag. I'm removing this from the bpf code altogether.

yanivagman (approved):

Issue #468, opened 2021-01-27
tracee kubernetes deployment yamls


itaysk:
create yaml deployment files for deploying tracee into a kubernetes cluster.
should create a daemonset for running tracee on each node.
should  be configurable via a configmap.

krol3:
I would like to help

itaysk:
let's take it step by step, I think that we can start without a configurable deployment, just run the tracee container and use the default options. another question is how the results are accessible to the user. the deployment should give an example of running falcosidekick (in the future maybe postee), which is independently configured by the user.

So to summarize
1. Deployment + Service that runs falcosidekick in the cluster. configurable by the user to send the event wherever they want. falcosidekick has a helm chart that we could use, so we can just add an instruction to helm install it (options that will install a minimal setup of just deployment+service)
2. Daemonset for Tracee. runs the aquasec/tracee container. the only configuration we need now is to point to the falcosidekick service endpoint.

potential improvements **for the future**:
1. control tracee-ebpf configs
2. use the slim image
3. persist /tmp/tracee as volume
4. select which rules to load

PR #680, opened 2021-04-06
tracee kubernetes deployment yamls 

krol3:
Related the issue https://github.com/aquasecurity/tracee/issues/468
- DaemonSet for tracee
- Deployment and service for the webhook, in this sample falcosidekick
- Configmap for the trace-rules templates
- Configmap for the falcosidekick config file.


deploy/kubernetes/tracee.yaml
```diff
@@ -0,0 +1,60 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  labels:
+    app: tracee
```
itaysk, line 5: can you please rename the labels according to https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels?

deploy/kubernetes/tracee.yaml
```diff
@@ -0,0 +1,60 @@
...
+        imagePullPolicy: Always
+        env:
+        - name: WEBHOOK_URL
+          value: http://webhook:2801
+        args:
+          - ./tracee-ebpf --output=format:gob | ./tracee-rules --input-tracee=file:stdin
+            --input-tracee=format:gob --webhook $WEBHOOK_URL --webhook-template /templates/falcosidekick.tmpl --webhook-content-type application/json
+        command:
+          - /bin/sh
+          - -c
```
itaysk, line 28: Is there a reason we override the container entrypoint? I think it's better to use it and pass webhook config as additional args

deploy/kubernetes/tracee.yaml
```diff
@@ -0,0 +1,60 @@
...
+        ports:
+        - containerPort: 8096
+          protocol: TCP
```
itaysk, line 31: what is this port?

deploy/kubernetes/tracee.yaml
```diff
@@ -0,0 +1,60 @@
...
+        - name: tmp-tracee
+          mountPath: /tmp/tracee
+        - name: usr-src
+          mountPath: /usr/src
+          readOnly: true
+        - name: lib-modules
+          mountPath: /lib/modules/
+          readOnly: true
+        - name: tracee-templates
+          mountPath: /templates
```
itaysk, line 45: we will start adding the template to the container since #683 so will not be necessary

deploy/kubernetes/webhook.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: webhook
```
itaysk, line 5: it's my understanding that app should be tracee in this case, as this is another component in the same app, wdyt?

deploy/kubernetes/webhook.yaml
```diff
@@ -0,0 +1,42 @@
...
+        imagePullPolicy: Always
+        ports:
+        - containerPort: 2801
+        args: ["-c","/app/config/cfg.yaml"]
+        volumeMounts:
+          - name: webhook-volume
+            mountPath: /app/config
+      volumes:
+        - configMap:
+            name: webhook-config
```
itaysk, line 29: could you rename this to `tracee-webhook-config`

deploy/kubernetes/cfg.yaml
```diff
@@ -0,0 +1,3 @@
+debug: true #
```
itaysk, line 1: can we make this a proper kubernetes configmap resource? (`kind: ConfigMap`)

deploy/kubernetes/Readme.md
```diff
@@ -0,0 +1,35 @@
+# Deploy Tracee on Kubernetes
+
+> NOTE `Minimal Requirements to run tracee in the node`
+````
+https://aquasecurity.github.io/tracee/install/prerequisites/
+````
+
+1. Create a ConfigMap that will hold the templates files for tracee-rules
```
itaysk, line 8: ideally, after my suggestions below, this entire example could be boiled down to `kubectl apply -f ./deploy`

deploy/kubernetes/Readme.md
```diff
@@ -0,0 +1,35 @@
+# Deploy Tracee on Kubernetes
+
+> NOTE `Minimal Requirements to run tracee in the node`
+````
+https://aquasecurity.github.io/tracee/install/prerequisites/
+````
```
itaysk, line 6: Can you add a note that 

> This approach assumes that kernel headers are available on the Kubernetes nodes under conventional location, e.g `/usr/src` and `/lib/modules`

itaysk: after merging this we should spend some time improving this experience: instructions to build the eBPF probe once using the tracee image, and then use the slim image in the daemonset. for a first step this is fine

itaysk (changes requested):

deploy/kubernetes/Readme.md
krol3: in the future we could have deploy/helm, similar folder starboard struct. Do you want to remove the kubernetes folder ?

deploy/kubernetes/webhook-cm.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: v1
+metadata:
+  name: tracee-webhook-config
+data:
+  cfg.yaml: |-
+    debug: true
+    slack:
+      webhookurl: "https://hooks.slack.com/xxxxxxxx" # Slack WebhookURL
+    teams:
+      webhookurl: "https://xxxxxxx"
```
itaysk, line 11: let's comment these out so it won't be shipped in a broken state

deploy/kubernetes/tracee.yaml
```diff
@@ -0,0 +1,49 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  labels:
+    app.kubernetes.io/name: tracee
+    app.kubernetes.io/instance: tracee-v1
+    app.kubernetes.io/version: "v0.5.0"
+    app.kubernetes.io/component: tracee
+    app.kubernetes.io/part-of: tracee
+    app.kubernetes.io/managed-by: aquasecurity
```
itaysk, line 10: I don't think manged-by means that, it means which tool manages this. we can remove it

deploy/kubernetes/tracee.yaml
```diff
@@ -0,0 +1,49 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  labels:
+    app.kubernetes.io/name: tracee
+    app.kubernetes.io/instance: tracee-v1
+    app.kubernetes.io/version: "v0.5.0"
```
itaysk, line 7: I'm not sure how to manage the version with static yamls. also the image ref is latest, so let's remove it

deploy/kubernetes/tracee.yaml
```diff
@@ -0,0 +1,49 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  labels:
+    app.kubernetes.io/name: tracee
+    app.kubernetes.io/instance: tracee-v1
```
itaysk, line 6: this instance label is meaningless if we put the same value in all instances, so let's remove it

deploy/kubernetes/Readme.md
```diff
@@ -0,0 +1,23 @@
+# Deploy Tracee on Kubernetes
```
itaysk, line 1: can you please move this to the docs under "installation"?

deploy/kubernetes/Readme.md
```diff
@@ -0,0 +1,23 @@
...
+ ```
+
+This sample deploy use falcosidekick.tmpl and a config map for the falcosidekick settings
+> NOTE `See the complete config file in` [falcosidekick](https://github.com/falcosecurity/falcosidekick)
+
+The service to integrate with tracee.
+ 
+``` bash
+https://tracee-webhook.default.svc.cluster.local:2801
+````
```
itaysk, line 23: I think this is all unnecessary after the latest changes right?
The only thing I would keep is the mention of falcosidekick, and a call out for the user to change the configmap before deploying

deploy/kubernetes/webhook-cm.yaml
```diff
@@ -0,0 +1,11 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: tracee-webhook-config
+data:
+  cfg.yaml: |-
```
itaysk, line 6: would be nice to add a comment here pointing to the reference: https://github.com/falcosecurity/falcosidekick/blob/master/config_example.yaml

itaysk (approved):
LGTM! (assuming you tested it on an actual k8s?)
I just added one small suggestion on the webhook configmap

krol3:
It was testing in the KUBE_VERSION=1.20.5

simar7:
Just revisiting this, how do I use the provided deployment yaml files? This is what I did:

```
vagrant@ubuntu-groovy:~/tracee$ k apply -f deploy/kubernetes/tracee.yaml 
daemonset.apps/tracee created

vagrant@ubuntu-groovy:~/tracee$ k get pods
NAME                        READY   STATUS              RESTARTS   AGE
tracee-qsgt2                0/1     Running   0          2s

vagrant@ubuntu-groovy:~/tracee$ k logs tracee-qsgt2
2021/06/11 02:20:45 kernel is not properly configured, missing: [CONFIG_BPF CONFIG_BPF_SYSCALL CONFIG_KPROBE_EVENTS CONFIG_BPF_EVENTS]
```


I understand the requirements and do have them on my host machine: https://github.com/aquasecurity/tracee/blob/main/docs/install/kubernetes.md#deploy-on-kubernetes – I'm curious how do they translate into the k8s environment? I assume if a pod is running in privileged mode, it should have access to the appropriate headers, etc. correct?

Issue #681, opened 2021-04-06
Build the tracee image after compile locally


krol3:
Steps to reproduce:
tracee> make all
tracee> make docker

Running the local build image . 

Also, I see that this output was reported in https://github.com/aquasecurity/tracee/pull/513

```
 docker run --name tracee --rm --privileged -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro -v /tmp/tracee:/tmp/tracee -it tracee:latest
Error relocating /tracee/tracee-rules: __vfprintf_chk: symbol not found
Error relocating /tracee/tracee-rules: __fprintf_chk: symbol not found
Error relocating /tracee/tracee-ebpf: __snprintf_chk: symbol not found
Error relocating /tracee/tracee-ebpf: __vfprintf_chk: symbol not found
Error relocating /tracee/tracee-ebpf: __memcpy_chk: symbol not found
Error relocating /tracee/tracee-ebpf: __strcpy_chk: symbol not found
Error relocating /tracee/tracee-ebpf: __sprintf_chk: symbol not found
```

itaysk:
@krol3 seems like incompatible libc (also mentioned in the discussion you linked). you don't need to `make all` before `make docker`. if you want to be sure, run 

```
sudo rm -rf dist tracee-ebpf/dist tracee-rules/dist
make docker
```

PR #682, opened 2021-04-06
Adding a dockerignore

krol3:
Related the issue https://github.com/aquasecurity/tracee/issues/681

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+./dist/*
+./tracee-ebpf/dist/*
+./tracee-rules/dist/*


itaysk (approved):
that's sounds like a good solution to me! let's get another reviewer though, just to make sure I'm not missing anything (haven't run it)

krol3:
Of course! I tested it using the same steps about the open issue. But maybe another collateral effect in the other dockerfiles could happen.

grantseltzer:
So this would mean when we run `COPY . /tracee` [here](https://github.com/aquasecurity/tracee/blob/main/tracee-ebpf/Dockerfile#L9) it wouldn't copy in the `dist` directories, right?

PR #683, opened 2021-04-06
integrate and document gotemplate

itaysk:
1. add important gotemplates to the container
2. better documentation of integrations page, including reference to 1


simar7 (approved):

simar7:
thanks for the docs!

Issue #686, opened 2021-04-07
tracee-ebpf doesn't build


itaysk:
```
github.com/aquasecurity/tracee/libbpfgo
net
# github.com/aquasecurity/tracee/libbpfgo
/usr/bin/ld: cannot find -lbpf
clang-10: error: linker command failed with exit code 1 (use -v to see invocation)
github.com/aquasecurity/tracee/libbpfgo/helpers
make[1]: *** [Makefile:45: dist/tracee-ebpf] Error 2
make[1]: Leaving directory '/tracee/tracee-ebpf'
make: *** [Makefile:40: dist/tracee-ebpf] Error 2
The command '/bin/sh -c make' returned a non-zero code: 2
make[1]: *** [Makefile:59: docker] Error 2
make[1]: Leaving directory '/home/runner/work/tracee/tracee'
make: *** [Makefile:78: release] Error 2
Error: Process completed with exit code 2.
```

itaysk:
removing the `-lbpf` flag from libbpf.go fixes it.

yanivagman:
seems to be related to https://github.com/aquasecurity/tracee/pull/643/files where "-lbpf" was added to libbpfgo.go file.
@grantseltzer any reason for this addition?

yanivagman:
As discussed, CGO_FLAGS and CGO_LDFLAGS env vars should be set when running this command

PR #687, opened 2021-04-07
Fix build with libbpfgo

yanivagman:
fix: https://github.com/aquasecurity/tracee/issues/686

### DIFF
--- a/libbpfgo/libbpfgo.go
+++ b/libbpfgo/libbpfgo.go
@@ -1,7 +1,7 @@
 package libbpfgo
 
 /*
-#cgo LDFLAGS: -lelf -lz -lbpf
+#cgo LDFLAGS: -lelf -lz
 
 #include <bpf/bpf.h>
 #include <bpf/libbpf.h>


itaysk (approved):
fine with me, but let's wait for @grantseltzer's approval before merging

grantseltzer:
@yanivagman @itaysk 

How were you able to verify that this fixes the issue? When I do the same thing, push the change to a local branch, then update the libbpfgo go module locally I get a build failure on the `go get` command.

And assuming this does fix the issue should we include upgrading the libbpfgo module in tracee-ebpf in this PR?

yanivagman:
@grantseltzer I changed tracee-ebpf/go.mod locally:
`replace github.com/aquasecurity/tracee/libbpfgo => ../libbpfgo`
and it now compiles correctly.

I think that upgrading libbpfgo module in tracee-ebpf should be done in a different commit.
In general, I prefer to split changes in tracee-ebpf and libbpfgo to different PRs

PR #693, opened 2021-04-09
Run libbpfgo self tests on self hosted github action runner

grantseltzer:

### DIFF
--- a/.github/workflows/pr-ebpf-selfhosted.yaml
+++ b/.github/workflows/pr-ebpf-selfhosted.yaml
@@ -1,14 +1,12 @@
 name: pr-ebpf
 on: 
-  pull_request:
-    paths: 
-      - "tracee-ebpf/**"
+  workflow_dispatch:
 jobs:
   build-test:
     runs-on: ${{ matrix.os }}
     strategy:
         matrix:
-            os: [ubuntu-20.04]
+            os: [centos-8.2-linux-4.18]
     defaults:
       run:
         working-directory: tracee-ebpf
--- a/.github/workflows/ebpf-selfhosted.yaml
+++ b/.github/workflows/ebpf-selfhosted.yaml
@@ -1,4 +1,4 @@
-name: pr-ebpf
+name: ebpf-selfhosted
 on: 
   workflow_dispatch:
 jobs:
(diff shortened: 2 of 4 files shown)


.github/workflows/pr-libbpfgo.yaml
```diff
@@ -5,15 +5,18 @@ on:
       - "libbpfgo/**"
 jobs:
   build-test:
-    runs-on: ubuntu-20.04
+    runs-on: ${{ matrix.os }}
```
simar7, line 8: Is this a secret stored somewhere? I'm just curious where the value comes from.

simar7: Ah nvm https://github.com/aquasecurity/tracee/pull/693/files#diff-b5f15b3b6c9c91e9bede2818583e85ac47b2f7efc6331846a73daf3b566c3578R11 must be this right? 

For my understanding, how does it know where the self hosted machine is located?

grantseltzer: Yes that is it. That label (centos-8.2-linux-4.18) corresponds to the machine which was setup in the settings of the repo. ([docs here](https://docs.github.com/en/actions/hosting-your-own-runners/adding-self-hosted-runners) if you're curious)

grantseltzer:
I've updated this so that the self hosted actions runners are run manually to avoid security concerns. This also makes sense to have in their own files because I have tests in the pipeline that make sense to only run on self hosted runners.

.github/workflows/pr-ebpf-selfhosted.yaml
```diff
@@ -0,0 +1,30 @@
+name: pr-ebpf
+on: 
+  workflow_dispatch:
```
simar7, line 3: I assume this is what makes it trigger manually right? I was reading the docs but didn't see the prompt like shown https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

Will I see it once we merge it?

simar7: Also who can trigger this? The maintainers of this repo or anyone in Aqua Github org?

grantseltzer: I merged this in my personal fork to confirm, yes it will show once merged:

Screen Shot 2021-04-21 at 4 06 42 PM

And I believe maintainers can trigger it.

grantseltzer: Also I renamed the self hosted workflow files.

simar7 (approved):
Litmus!

Issue #691, opened 2021-04-08
Tracee panic when cannot allocate memory


simar7:
Reported by @eidans 

```
michael@MichaelsLenovo:~/workarea/tracee$ sudo ./dist/tracee-ebpf --perf-buffer-size=16384 --security-alerts -t event=execve,security_bprm_check,security_file_open,fork,vfork,clone,sched_process_exit,sec
urity_inode_unlink,ptrace,bpf,process_vm_writev,connect,accept,bind,mem_prot_alert,init_module,finit_module,dup,dup2,dup3,close --output=format:gob | ./dist/tracee-rules --input-tracee=file:stdin --input
-tracee=format:gob
OOOOOOO
Loaded signature(s): [TRC-1 TRC-2 TRC-BYPASS TRC-3 TRC-RUNTIME-CONTAINER-START TRC-RUNTIME-CONTAINER-STOP TRC-4 TRC-5 TRC-6 TRC-7]
libbpf: failed to mmap perf buffer on cpu #7: Cannot allocate memory
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x1 addr=0x34 pc=0x642b87]

runtime stack:
runtime.throw(0x7b29b4, 0x2a)
/usr/local/go/src/runtime/panic.go:1117 +0x72
runtime.sigpanic()
/usr/local/go/src/runtime/signal_unix.go:718 +0x2e5

goroutine 23 [syscall]:
runtime.cgocall(0x631890, 0xc00003a738, 0x0)
/usr/local/go/src/runtime/cgocall.go:154 +0x5b fp=0xc00003a708 sp=0xc00003a6d0 pc=0x40581b
github.com/aquasecurity/tracee/libbpfgo._Cfunc_perf_buffer__poll(0xfffffffffffffff4, 0xc00000012c, 0x0)
_cgo_gotypes.go:467 +0x48 fp=0xc00003a738 sp=0xc00003a708 pc=0x573388
github.com/aquasecurity/tracee/libbpfgo.(*PerfBuffer).poll.func1(0xc0000b4b70, 0xc000096180)
/root/go/pkg/mod/github.com/aquasecurity/tracee/libbpfgo@v0.0.0-20210318031738-f66f7bedda26/libbpfgo.go:764 +0x65 fp=0xc00003a770 sp=0xc00003a738 pc=0x5783a5
[...]

PR #694, opened 2021-04-09
fix: Use a proper macro to handle libbpf error returns

simar7:
Fixes: https://github.com/aquasecurity/tracee/issues/691

### DIFF
--- a/libbpfgo/libbpfgo.go
+++ b/libbpfgo/libbpfgo.go
@@ -12,6 +12,7 @@ package libbpfgo
 
 #include <asm-generic/unistd.h>
 #include <errno.h>
+#include <err.h>
 #include <fcntl.h>
 #include <linux/perf_event.h>
 #include <linux/unistd.h>
@@ -38,7 +39,7 @@ void set_print_fn() {
 struct ring_buffer * init_ring_buf(int map_fd, uintptr_t ctx) {
     struct ring_buffer *rb = NULL;
     rb = ring_buffer__new(map_fd, ringbufferCallback, (void*)ctx, NULL);
-    if (!rb) {
+    if IS_ERR(rb) {
         fprintf(stderr, "Failed to initialize ring buffer\n");
         return NULL;
     }
@@ -52,7 +53,7 @@ struct perf_buffer * init_perf_buf(int map_fd, int page_cnt, uintptr_t ctx) {
     pb_opts.lost_cb = perfLostCallback;
     pb_opts.ctx = (void*)ctx;
     pb = perf_buffer__new(map_fd, page_cnt, &pb_opts);
-    if (pb < 0) {
+    if IS_ERR(pb) {
         fprintf(stderr, "Failed to initialize perf buffer!\n");
         return NULL;
     }
--- a/libbpfgo/libbpfgo.go
+++ b/libbpfgo/libbpfgo.go
@@ -1,5 +1,16 @@
 package libbpfgo
 
+import "C"
+import (
+	"fmt"
+	"strings"
+	"sync"
+	"syscall"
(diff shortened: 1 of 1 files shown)


simar7:
@yanivagman Any ideas how we can test this?

libbpfgo/libbpfgo.go
```diff
@@ -52,7 +52,7 @@ struct perf_buffer * init_perf_buf(int map_fd, int page_cnt, uintptr_t ctx) {
     pb_opts.lost_cb = perfLostCallback;
     pb_opts.ctx = (void*)ctx;
     pb = perf_buffer__new(map_fd, page_cnt, &pb_opts);
-    if (pb < 0) {
+    if IS_ERR_VALUE(pb) {
```
yanivagman, line 55: We should probably do the same for init_ring_buf

yanivagman (commented):
Shouldn't we include the err.h header explicitly? I'm not sure how this even works

yanivagman:
About how to test - I believe that we can just give some invalid argument.
For example, invalid map_fd should trigger an error

simar7:
> About how to test - I believe that we can just give some invalid argument.
> For example, invalid map_fd should trigger an error

I tried this but wasn't successful. I mentioned earlier in this comment https://github.com/aquasecurity/tracee/issues/691#issuecomment-816253557 

Any ideas what am I missing?

simar7:
> About how to test - I believe that we can just give some invalid argument.
> For example, invalid map_fd should trigger an error

Maybe I'm doing something wrong here but I wasn't able to reproduce this on my machine. https://github.com/aquasecurity/tracee/issues/691#issuecomment-816253557

Any idea what am I doing differently?

yanivagman:
You are probably not using your local copy of libbpfgo, but from the repo.
Did you edit go.mod of tracee-ebpf to use it?
`replace github.com/aquasecurity/tracee/libbpfgo => ../libbpfgo`

I tested with a negative map_fd given to init_perf_buf and it worked (meaning failed with an error :-) )

simar7:
> You are probably not using your local copy of libbpfgo, but from the repo.
> Did you edit go.mod of tracee-ebpf to use it?
> `replace github.com/aquasecurity/tracee/libbpfgo => ../libbpfgo`
> 
> I tested with a negative map_fd given to init_perf_buf and it worked (meaning failed with an error :-) )

Thanks that was it. I tested with a negative map_fd and it did work.

libbpfgo/libbpfgo.go
```diff
@@ -23,6 +23,14 @@ extern void perfLostCallback(void *ctx, int cpu, __u64 cnt);
 
 extern int ringbufferCallback(void *ctx, void *data, size_t size);
 
+#define MAX_ERRNO       4095
```
yanivagman, line 26: Can you please add a comment here that this was taken from err.h of libbpf?

simar7: Yes https://github.com/aquasecurity/tracee/pull/694/commits/1f4717ba14be975d4efe2aee3ee9f84b25aade57#diff-4108726d54769f59926ec61abdeaab04a651cdf8766a24a74a859b4d98befbc8R37

libbpfgo/libbpfgo.go
```diff
@@ -1,5 +1,16 @@
 package libbpfgo
 
+import "C"
```
yanivagman, line 3: Any reason to move all the imports here?
Import "c" should be after the c code (and from all the cgo examples I saw, all the other imports come after it)

simar7:
> @simar7 can we close this PR in favor of #709?

Yeah good idea.

Issue #690, opened 2021-04-07
Tracee-ebpf prints multiple EOF after shutdown


yanivagman:
When running tracee-ebpf and pressing `ctrl+c` to terminate the process, I get a lot of `EOF` printed to stdout.

I found that this behavior started after merging https://github.com/aquasecurity/tracee/pull/638.
When I comment out `close(eventsChan)` from `func (pb *PerfBuffer) Stop()`, the issue disappears.
Not sure yet why this happens

itaysk:
I get it too now

grantseltzer:
(just taking notes)

I traced the error back to here:

https://github.com/aquasecurity/tracee/blob/main/tracee-ebpf/tracee/tracee.go#L1282-L1290

itaysk:
fixed by #696

PR #696, opened 2021-04-12
Add check for empty bytes being written by file write channel 

grantseltzer:
This is to handle #690 

The "EOF" error is printed as a result of the File Write Channel sending empty raw data. I believe the reason for this has to do with allocating memory for []byte{} that'd be send over the file_write channel, but nothing ever being written, leading it to being drained in this way. 

This will skip over empty raw data that's been sent. Alternatively we can add a check for `err == io.EOF` and ignore it specifically.

### DIFF
--- a/tracee-ebpf/tracee/tracee.go
+++ b/tracee-ebpf/tracee/tracee.go
@@ -1280,6 +1280,9 @@ func (t *Tracee) processFileWrites() {
 	for {
 		select {
 		case dataRaw := <-t.fileWrChannel:
+			if len(dataRaw) == 0 {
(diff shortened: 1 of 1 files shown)


itaysk (approved):
do we need the same in `decodeRawEvent` ?

grantseltzer:
No I don't believe so, this is specifically related to the file write channel.

yanivagman:
Note that this doesn't fix the root cause of the bug.
Why can there be events of 0 bytes in the file write channel?
The callback shouldn't have been called in that case...

PR #697, opened 2021-04-13
Network lsm hooks

roikol:
added events for network lsm hooks:
security_socket_create
security_socket_bind
security_socket_listen
security_socket_accept
security_socket_connect


tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1422,6 +1448,85 @@ int trace_ret_##name(void *ctx)                                         \
...
+
+    u32 addr = 0;
+    bpf_probe_read(&addr, sizeof(addr), &inet->inet_rcv_saddr);
+    if ( !addr ) {
+        bpf_probe_read(&addr, sizeof(addr), &inet->inet_saddr);
+    }
+
+    if ( peer ) {
+
+        bpf_probe_read(&net_details->local_address, sizeof(net_details->local_address), &inet->inet_daddr);
```
yanivagman, line 1471: kernel structs (inet_sock in this case) may change between different kernel versions.
For this reason, we dereference structs using helper functions which are located in the "KERNEL VERSION DEPENDANT HELPER FUNCTIONS" section of the bpf source code.
Any field we read from a kernel struct (e.g. inet->inet_daddr) can have its own helper.

Other than that, you can use the `READ_KERN` helper instead of `bpf_probe_read`

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2064,6 +2169,309 @@ int BPF_KPROBE(trace_cap_capable)
...
+        return 0;
+    }
+
+    // getting destination details
+
+    struct sockaddr *address = (struct sockaddr *)PT_REGS_PARM2(ctx);
+    sa_family_t sa_fam;
+
+    // read family from struct sockaddr*
+    bpf_probe_read(&sa_fam, sizeof(sa_fam), &address->sa_family);
```
yanivagman, line 2326: You read family here and from sk->sk_family in 2315.
Any reason for the value to be different between these two?

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2064,6 +2169,309 @@ int BPF_KPROBE(trace_cap_capable)
...
+        // saving to submit buffer
+        save_to_submit_buf(submit_p, (void *)address, sizeof(struct sockaddr_in), SOCKADDR_T, DEC_ARG(0, *tags));
+
+    }
+    else if (sa_fam == AF_INET6) {
+        // saving to submit buffer
+        save_to_submit_buf(submit_p, (void *)address, sizeof(struct sockaddr_in6), SOCKADDR_T, DEC_ARG(0, *tags));
+    }
+    else {
+        // not AF_INET or AF_INET6
```
yanivagman, line 2337: If the family value is the same for my previous question, this else can be removed as you check for this in L2316

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2064,6 +2169,309 @@ int BPF_KPROBE(trace_cap_capable)
...
+        net_conn_v4_t net_details = {};
+
+        get_network_details_from_sock_v4(sk, &net_details, 0);
+
+        struct sockaddr_in local;
+        local.sin_family = family;
+        local.sin_port = net_details.local_port;
+        struct in_addr local_addr;
+        local_addr.s_addr = net_details.local_address;
+        local.sin_addr = local_addr;
```
yanivagman, line 2255: Why do we need local_addr var?
We can use instead:
local.sin_addr = net_details.local_address;

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2064,6 +2169,309 @@ int BPF_KPROBE(trace_cap_capable)
...
```
yanivagman, line 2391: same comment about local_addr

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2064,6 +2169,309 @@ int BPF_KPROBE(trace_cap_capable)
...
+    if (submit_p == NULL)
+        return 0;
+    set_buf_off(SUBMIT_BUF_IDX, sizeof(context_t));
+
+    // getting source details
+
+    u16 family;
+    struct sock *sk;
+
+    struct socket *sock = (struct socket *)PT_REGS_PARM1(ctx);
```
yanivagman, line 2361: Don't we want to use PARAM2 where we get `struct socket *newsock` as well?

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2064,6 +2169,309 @@ int BPF_KPROBE(trace_cap_capable)
...
+
+    // getting event tags
+    u64 *tags = bpf_map_lookup_elem(&params_names_map, &context.eventid);
+    if (!tags) {
+        return -1;
+    }
+
+    // getting destination details
+
+    struct sockaddr *address = (struct sockaddr *)PT_REGS_PARM2(ctx);
```
yanivagman, line 2456: If we only use PARAM2 to extract data about the socket and send to user space, why do we also need to read PARAM1 and check for the family above?

tracee-ebpf/tracee/tracee.go
```diff
@@ -905,7 +905,7 @@ func (t *Tracee) Close() {
 	if t.bpfModule != nil {
 		t.bpfModule.Close()
 	}
-	t.printer.Close()
+	//t.printer.Close()
```
yanivagman, line 908: Any reason for that?

yanivagman (changes requested):
Thanks @roikol - this is a very good and important addition.

tracee-ebpf/tracee/consts.go
```diff
@@ -153,735 +154,750 @@ const (
...
+	SecuritySocketListenEventID:	{ID: SecuritySocketListenEventID, ID32Bit: sys32undefined, Name: "security_socket_listen", Probes: []probe{{event: "security_socket_listen", attach: kprobe, fn: "trace_security_socket_listen"}}, Sets: []string{"default"}},
+	SecuritySocketConnectEventID:	{ID: SecuritySocketConnectEventID, ID32Bit: sys32undefined, Name: "security_socket_connect", Probes: []probe{{event: "security_socket_connect", attach: kprobe, fn: "trace_security_socket_connect"}}, Sets: []string{"default"}},
+	SecuritySocketAcceptEventID:	{ID: SecuritySocketAcceptEventID, ID32Bit: sys32undefined, Name: "security_socket_accept", Probes: []probe{{event: "security_socket_accept", attach: kprobe, fn: "trace_security_socket_accept"}}, Sets: []string{"default"}},
+	SecuritySocketBindEventID: 		{ID: SecuritySocketBindEventID, ID32Bit: sys32undefined, Name: "security_socket_bind", Probes: []probe{{event: "security_socket_bind", attach: kprobe, fn: "trace_security_socket_bind"}}, Sets: []string{"default"}},
```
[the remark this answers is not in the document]
roikol: i'm not sure that if we are adding these events to the default event set we should remove their corresponding syscalls. for example, 'open' syscall and 'security_file_open' lsm are both in the default set.

in any case, what do you suggest:
-leaving the Sets property empty
-add then to an already existing set
-add them to a new set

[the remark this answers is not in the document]
yanivagman: I think that now we have enough lsm events, so we can add a new "lsm_hooks" set, which will include all of the events you added here, plus "security_file_open", "security_bprm_check" and "security_inode_unlink". This can be done as part of this PR or in a different PR if you prefer

tracee-ebpf/tracee/consts.go
```diff
@@ -153,735 +154,750 @@ const (
...
+	CommitCredsEventID:         	{{Type: "int", Name: "old_euid"}, {Type: "int", Name: "new_euid"}, {Type: "int", Name: "old_egid"}, {Type: "int", Name: "new_egid"}, {Type: "int", Name: "old_fsuid"}, {Type: "int", Name: "new_fsuid"}, {Type: "u64", Name: "old_cap_eff"}, {Type: "u64", Name: "new_cap_eff"}},
+	SwitchTaskNSEventID:        	{{Type: "pid_t", Name: "pid"}, {Type: "u32", Name: "new_mnt"}, {Type: "u32", Name: "new_pid"}, {Type: "u32", Name: "new_uts"}, {Type: "u32", Name: "new_ipc"}, {Type: "u32", Name: "new_net"}, {Type: "u32", Name: "new_cgroup"}},
+	MagicWriteEventID:          	{{Type: "const char*", Name: "pathname"}, {Type: "bytes", Name: "bytes"}},
+	SecuritySocketCreateEventID:	{{Type: "int", Name: "family"}, {Type: "int", Name: "type"}, {Type: "int", Name: "protocol"}, {Type: "int", Name: "kern"}},
+	SecuritySocketListenEventID:	{{Type: "struct sockaddr*", Name: "local_addr"}, {Type: "int", Name: "backlog"}},
+	SecuritySocketConnectEventID:	{{Type: "struct sockaddr*", Name: "remote_addr"}},
+	SecuritySocketAcceptEventID:	{{Type: "struct sockaddr*", Name: "local_addr"}},
+	SecuritySocketBindEventID:		{{Type: "struct sockaddr*", Name: "local_addr"}},
```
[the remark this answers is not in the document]
roikol: `struct socket *` is the input for these LSM hooks, but the output I chose to reflect is the addresses (not the socket structs) which are of type `struct sockaddr*`.
so for each of these LSM hooks, the output consists of different parameters compared to the input.
but the Type will correspond to the Value

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -281,6 +293,20 @@ typedef struct alert {
     u8 payload; // Non zero if payload is sent to userspace
 } alert_t;
 
+typedef struct network_connection_v4 {
+    u32 local_address;
+    u16 local_port;
+    u32 remote_address;
+    u16 remote_port;
```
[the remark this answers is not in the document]
roikol: maybe I understood you wrongly, but this struct is just to hold addresses and ports, it is not supposed to represent a network flow. 
the network flow struct will be added in a different PR and will consist of local ip and port, and also protocol as per this comment.

tracee-ebpf/tracee/tracee.bpf.c
roikol: these values come from the user - the user creates the socket with a specified family, and the user decides which address to connect to as well as the family of the address. so these values can be different.
the connect syscall checks if the two families are aligned, if not it returns an error EAFNOSUPPORT. but this happens after the LSM hook is called.

tracee-ebpf/tracee/tracee.bpf.c
roikol: we want the output to be of type `struct sockaddr_in`, and `sockaddr_in.sin_addr` is of type `struct in_addr`.
when we want to get the actual address we use: `sockaddr_in.sin_addr.s_addr`.

tracee-ebpf/tracee/tracee.bpf.c
roikol: `newsock` holds no relevant information at the point of this LSM hook. the addresses are copied into it only after the LSM hook.

tracee-ebpf/tracee/tracee.bpf.c
roikol: as commented before - these values come from the user and can be different.

tracee-ebpf/tracee/tracee.go
roikol: accidentally added this here. it's not related to this PR

tracee-ebpf/tracee/tracee.bpf.c
yanivagman: In that case, I think it is enough to use the family given by the user's address.
We are more interested in the intentions of the user here, and if there is no match, we can see it in the return value from the syscall. This will allow us to remove some lines here and save some cycles

tracee-ebpf/tracee/tracee.bpf.c
yanivagman: See my other comment

tracee-ebpf/tracee/tracee.bpf.c
yanivagman: So can't we just use `local.sin_addr.s_addr = net_details.local_address;` ?

roikol: you're right

tracee-ebpf/tracee/tracee.go
```diff
@@ -1148,6 +1148,13 @@ func (t *Tracee) prepareArgsForPrint(ctx *context, args map[argTag]interface{})
 		if typ, isInt32 := args[t.EncParamName[ctx.EventID%2]["type"]].(int32); isInt32 {
 			args[t.EncParamName[ctx.EventID%2]["type"]] = PrintSocketType(uint32(typ))
 		}
+	case SecuritySocketCreateEventID:
+		if dom, isInt32 := args[t.EncParamName[ctx.EventID%2]["family"]].(int32); isInt32 {
+			args[t.EncParamName[ctx.EventID%2]["family"]] = PrintSocketDomain(uint32(dom))
```
yanivagman, line 1153: Please rebase on main. We now use helpers.ParseSocketDomain

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1422,6 +1503,91 @@ int trace_ret_##name(void *ctx)                                         \
...
+    }
+
+    return 0;
+}
+
+static __always_inline struct ipv6_pinfo *inet6_sk_own_impl(const struct sock *__sk, struct inet_sock *inet)
+{
+    volatile unsigned char sk_state_own_impl;
+    bpf_probe_read((void *)&sk_state_own_impl, sizeof(sk_state_own_impl), (const void *)&__sk->sk_state);
+//    sk_state_own_impl = (void *)get_sock_state(__sk);
```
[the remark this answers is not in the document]
roikol: this specific helper i implemented using bpf_probe_read instead of READ_KERN because of compilation warnings.

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -553,6 +579,75 @@ static inline struct mount *real_mount(struct vfsmount *mnt)
...
+}
+
+static __always_inline struct in6_addr get_sock_v6_daddr(struct sock *sock)
+{
+    return READ_KERN(sock->sk_v6_daddr);
+}
+
+static __always_inline volatile unsigned char get_sock_state(struct sock *sock)
+{
+    // return READ_KERN(sock->sk_state);
```
simar7, line 639: cruft?

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2064,6 +2238,254 @@ int BPF_KPROBE(trace_cap_capable)
...
+    events_perf_submit(ctx);
+    return 0;
+};
+
+SEC("kprobe/security_socket_connect")
+int BPF_KPROBE(trace_security_socket_connect)
+{
+
+    // trace the event security_socket_connect
+
```
yanivagman, line 2346: These lines can be removed

yanivagman (approved):
LGTM

PR #709, opened 2021-04-23
libbpfgo: Check for ERR_PTR return values

### LINKED ISSUE #691: Tracee panic when cannot allocate memory

### LINKED ISSUE #708: libbpfgo crashes when some of libbpf functions fail
When `libbpf` functions returning pointers fail, they sometimes return a `NULL` and sometimes encode `errno` inside the pointer (by using kernel's `ERR_PTR` mechanism). As such it is not sufficient to just check for NULL as it leads the error to not be detected and causes a subsequent crash.

eyakubovich:
Some libbpf functions fail and return an -errno
encoded inside the pointer. Therefore checking
for NULL is not enough.

Checks the returned pointers for NULL or ERR_PTR
encoding and converts the errno (if available)
to an error message.

Fixes #708 
Fixes #691

### DIFF
--- a/libbpfgo/libbpfgo.go
+++ b/libbpfgo/libbpfgo.go
@@ -18,6 +18,24 @@ package libbpfgo
 #include <string.h>
 #include <unistd.h>
 
+#ifndef MAX_ERRNO
+#define MAX_ERRNO       4095
+
+#define IS_ERR_VALUE(x) ((x) >= (unsigned long)-MAX_ERRNO)
+
+static inline bool IS_ERR(const void *ptr) {
+	return IS_ERR_VALUE((unsigned long)ptr);
+}
+
+static inline bool IS_ERR_OR_NULL(const void *ptr) {
+	return !ptr || IS_ERR_VALUE((unsigned long)ptr);
+}
+
+static inline long PTR_ERR(const void *ptr) {
+	return (long) ptr;
+}
+#endif
+
 extern void perfCallback(void *ctx, int cpu, void *data, __u32 size);
 extern void perfLostCallback(void *ctx, int cpu, __u64 cnt);
 
@@ -261,14 +279,25 @@ func bumpMemlockRlimit() error {
 	return nil
 }
 
+func errptrError(ptr unsafe.Pointer, format string, args... interface{}) error {
+	negErrno := C.PTR_ERR(ptr)
+	errno := syscall.Errno(-int64(negErrno))
+	if errno == 0 {
+		return fmt.Errorf(format, args...)
+	}
+
+	args = append(args, errno.Error())
+	return fmt.Errorf(format + ": %v", args...)
+}
+
 func NewModuleFromFile(bpfObjFile string) (*Module, error) {
 	C.set_print_fn()
 	bumpMemlockRlimit()
 	cs := C.CString(bpfObjFile)
 	obj := C.bpf_object__open(cs)
 	C.free(unsafe.Pointer(cs))
-	if obj == nil {
-		return nil, fmt.Errorf("failed to open BPF object %s", bpfObjFile)
+	if C.IS_ERR_OR_NULL(unsafe.Pointer(obj)) {
+		return nil, errptrError(unsafe.Pointer(obj), "failed to open BPF object %s", bpfObjFile)
 	}
 
 	return &Module{
@@ -285,8 +314,8 @@ func NewModuleFromBuffer(bpfObjBuff []byte, bpfObjName string) (*Module, error) 
 	obj := C.bpf_object__open_buffer(buffPtr, buffSize, name)
 	C.free(unsafe.Pointer(name))
 	C.free(unsafe.Pointer(buffPtr))
-	if obj == nil {
-		return nil, fmt.Errorf("failed to open BPF object %s: %v", bpfObjName, bpfObjBuff[:20])
+	if C.IS_ERR_OR_NULL(unsafe.Pointer(obj)) {
+		return nil, errptrError(unsafe.Pointer(obj), "failed to open BPF object %s: %v", bpfObjName, bpfObjBuff[:20])
 	}
 
 	return &Module{
@@ -556,13 +585,16 @@ func (p *BPFProg) SetTracepoint() error {
 
 func (p *BPFProg) AttachTracepoint(tp string) (*BPFLink, error) {
 	tpEvent := strings.Split(tp, ":")
+	if len(tpEvent) != 2 {
+		return nil, fmt.Errorf("tracepoint must be in 'category:name' format")
+	}
 	tpCategory := C.CString(tpEvent[0])
 	tpName := C.CString(tpEvent[1])
 	link := C.bpf_program__attach_tracepoint(p.prog, tpCategory, tpName)
 	C.free(unsafe.Pointer(tpCategory))
 	C.free(unsafe.Pointer(tpName))
-	if link == nil {
-		return nil, fmt.Errorf("failed to attach tracepoint %s to program %s", tp, p.name)
+	if C.IS_ERR_OR_NULL(unsafe.Pointer(link)) {
+		return nil, errptrError(unsafe.Pointer(link), "failed to attach tracepoint %s to program %s", tp, p.name)
 	}
 
 	bpfLink := &BPFLink{
@@ -579,8 +611,8 @@ func (p *BPFProg) AttachRawTracepoint(tpEvent string) (*BPFLink, error) {
 	cs := C.CString(tpEvent)
 	link := C.bpf_program__attach_raw_tracepoint(p.prog, cs)
 	C.free(unsafe.Pointer(cs))
-	if link == nil {
-		return nil, fmt.Errorf("failed to attach raw tracepoint %s to program %s", tpEvent, p.name)
+	if C.IS_ERR_OR_NULL(unsafe.Pointer(link)) {
+		return nil, errptrError(unsafe.Pointer(link), "failed to attach raw tracepoint %s to program %s", tpEvent, p.name)
 	}
 
 	bpfLink := &BPFLink{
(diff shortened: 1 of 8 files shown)


yanivagman:
Thanks @eyakubovich , we also noticed this bug.
Please see #694 which also attempts to fix it.
Your PR seems to handle cases which are not fixed by #694 but I don't see a reason to open a new PR for this.

eyakubovich:
Should've looked at existing PRs before starting.
- This adds the check to more functions that are affected (compared to #694)
- It's best to check for both the NULL and `PTR_ERR` because some functions are not consistent: some code paths return NULL and some use the `ERR_PTR`. I think it's bad form but that's the way it is.

yanivagman:
@simar7

libbpfgo/libbpfgo.go
```diff
@@ -260,14 +278,25 @@ func bumpMemlockRlimit() error {
 	return nil
 }
 
+func errnoError(ptr unsafe.Pointer, errFmt string, args... interface{}) error {
+	negErrno := C.PTR_ERR(ptr)
+	errno := syscall.Errno(-int64(negErrno))
+	if errno == 0 {
+		return fmt.Errorf(errFmt, args...)
+	} else {
```
simar7, line 286: nit: you can drop this else as there's already a `return` in the if.

libbpfgo/libbpfgo.go
```diff
@@ -260,14 +278,25 @@ func bumpMemlockRlimit() error {
...
 func NewModuleFromFile(bpfObjFile string) (*Module, error) {
 	C.set_print_fn()
 	bumpMemlockRlimit()
 	cs := C.CString(bpfObjFile)
 	obj := C.bpf_object__open(cs)
 	C.free(unsafe.Pointer(cs))
-	if obj == nil {
-		return nil, fmt.Errorf("failed to open BPF object %s", bpfObjFile)
+	if C.IS_ERR_OR_NULL(unsafe.Pointer(obj)) {
+		return nil, errnoError(unsafe.Pointer(obj), "failed to open BPF object %s", bpfObjFile)
```
simar7, line 299: Maybe this is a good time to add a unit test for this function in order to avoid such scenarios in the future.

simar7: ditto elsewhere where the check is done

eyakubovich: I think it would be beneficial to use a unit testing framework (builtin or `check`) instead of standalone selftests.
Should that be a separate issue/PR though?

simar7: A simple unit test would suffice IMO. I usually like to write the test for the change of code along with it.

eyakubovich: I agree, I'm saying that libbpfgo as a whole is missing unit tests. It would be beneficial for write the tests that cover the whole lib.

But I'll kick things off by adding at least a test for a one or two of these functions.

eyakubovich: I added the tests but nothing is running them yet. I used `//go:generate` instead of a `Makefile` (can change it) so the CI would need to run `go generate` prior to running `go test`. PTAL and let me know how to best invoke them in CI.

The tests also require 5.8+ kernel because of `AttachLSM` although I can put it behind a flag or something.

eyakubovich: /cc @simar7

simar7: Thanks I will take a look at it shortly and get back to you.

simar7: > I added the tests but nothing is running them yet. I used `//go:generate` instead of a `Makefile` (can change it) so the CI would need to run `go generate` prior to running `go test`. PTAL and let me know how to best invoke them in CI.
> 
> The tests also require 5.8+ kernel because of `AttachLSM` although I can put it behind a flag or something.

Thanks this looks wonderful @eyakubovich! Looks like it just needs to be told where to find the right headers:

```
vagrant@ubuntu-groovy:/usr/vagrant/repos/tracee/libbpfgo$ go generate
test/test.bpf.c:3:10: fatal error: 'bpf/bpf_helpers.h' file not found
#include <bpf/bpf_helpers.h>
         ^~~~~~~~~~~~~~~~~~~
1 error generated.
libbpfgo_test.go:2: running "clang": exit status 1
```

As you mentioned it would be better to add them to run via the Makefile as some of the pre-requisite headers will be set in it. We currently do this for running perfbuffer and ringbuffer tests and you can find an example here: https://github.com/aquasecurity/tracee/blob/77cf435059442695b6b604630406ddc8614820e8/libbpfgo/selftest/perfbuffers/Makefile#L4-L7

eyakubovich: @simar7 PTAL. The tests fail on CI (as expected, kernel version) so I didn't add them to the GH actions. Can you check if they run for you? I'm a bit surprised that `#include <bpf/bpf_helpers.h>` failed -- if libbpf is installed, it should be in `/usr/include` and should work. If it's installed elsewhere, I guess the user will need to specify.

simar7: Once we sort the running of the tests locally, we are good to merge this in. Appreciate your patience.

yanivagman:
@simar7 any reason why not to merge this PR now?
This is an important bug fix, and we can take care of any other required test additions after we merge this

simar7:
> @simar7 any reason why not to merge this PR now?
> This is an important bug fix, and we can take care of any other required test additions after we merge this

I'm still unable to run the added tests. Merging an important bug fix without adding a test that exercises it seems like an omission to me. If you think that's fine, we can merge it.

simar7 (approved):

Issue #365, opened 2020-12-19
Support LSM BPF programs


yanivagman:
Linux kernel 5.7 added support for LSM BPF programs, allowing to enforce policies using LSM hooks attached to BPF programs.
Tracee already traces some LSM functions using kprobes (e.g. security_file_open).
Supporting LSM BPF programs will allow us to audit security events as a first step, and enforce policies in the future

yanivagman:
Closing in favor of #1686

PR #710, opened 2021-04-23
libbpfgo: Add AttachLSM() method

eyakubovich:
Although LSM is attached via BPF_RAW_TRACEPOINT_OPEN, using
AttachRawTracepoint is not possible since the tp_name argument
to bpf_program__attach_raw_tracepoint must be NULL.

Partially addresses #365


yanivagman (approved):
Thanks @eyakubovich

PR #715, opened 2021-05-03
Add security_sb_mount lsm hook

AsafEitani:
Add security_sb_mount as tracee event

### DIFF
--- a/tracee-ebpf/tracee/tracee.bpf.c
+++ b/tracee-ebpf/tracee/tracee.bpf.c
@@ -152,7 +152,8 @@
 #define SECURITY_SOCKET_CONNECT 1016
 #define SECURITY_SOCKET_ACCEPT  1017
 #define SECURITY_SOCKET_BIND    1018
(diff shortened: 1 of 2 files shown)


tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1204,6 +1206,12 @@ static __always_inline int save_file_path_to_str_buf(buf_t *string_p, struct fil
     return buf_off;
 }
 
+static __always_inline int save_file_path_to_str_buf(buf_t *string_p, struct file* file)
+{
+    struct path f_path = get_path_from_file(file);
+    return save_path_to_str_buf(string_p, &f_path);
```
yanivagman, line 1212: We read struct path both here and in line 1138.
Maybe it is possible to do just one read by using &file->f_path and removing the previous line
(It is possible that the verifier won't like it)

yanivagman (approved):

PR #718, opened 2021-05-05
Sockfd to network lsm

roikol:
added sockfd arg to:
- security_socket_bind
- security_socket_listen
- security_socket_connect
- security_socket_connect


tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1388,6 +1399,46 @@ static __always_inline int del_retval(u32 event_id)
     return 0;
 }
 
+static __always_inline int save_sockfd(u32 sockfd)
+{
+    u64 pid_tgid = bpf_get_current_pid_tgid();
+    u32 tgid;
+    tgid = (u32)((pid_tgid & 0xFFFFFFFF00000000LL) >> 32);
```
yanivagman, line 1406:
```suggestion
    u32 tgid = pid_tgid >> 32;
```

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1664,6 +1715,10 @@ int tracepoint__raw_syscalls__sys_enter(struct bpf_raw_tracepoint_args *ctx)
             bpf_map_update_elem(&new_pids_map, &pid, &pid, BPF_ANY);
         }
     }
+    else if (id == SYSCALL_CONNECT || id == SYSCALL_ACCEPT || id == SYSCALL_ACCEPT4 || id == SYSCALL_BIND || id == SYSCALL_LISTEN) {
+        u32 sockfd = args_tmp.args[0];
+        save_sockfd(sockfd);
+    }
```
yanivagman, line 1721: Why not putting this after `should_trace()`?

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1725,6 +1780,10 @@ int tracepoint__raw_syscalls__sys_exit(struct bpf_raw_tracepoint_args *ctx)
     if (load_args(&saved_args, delete_args, id) != 0)
         return 0;
 
+    if (id == SYSCALL_CONNECT || id == SYSCALL_ACCEPT || id == SYSCALL_ACCEPT4 || id == SYSCALL_BIND || id == SYSCALL_LISTEN) {
+        del_sockfd();
+    }
+
```
yanivagman, line 1786: same

itaysk:
I had one concern that I shared with Roi about the thread safety of the sockfd map, if it was possible that a thread created multiple sockets which due to os scheduling would overwrite each other in the map. Roi explained that it's not possible due to the nature of how LSM hooks are executed. 

My other concern was that we add information to the LSM event that not part of its signature, which we need to consider or document. But we already have other cases like this that we need to address systematically so we can address it later

yanivagman:
Actually, there might be a race condition here as tgid is used instead of pid (which is actually tid in user space). If several threads of the same process create a socket at the same time, they may override each other value in the map. @roikol can you please remind me why you chose to use tgid as a key for this map and not pid? Even though file descriptors are shared between the threads of a process, I think it shouldn't be a problem to use (kernel) pid here as a key, since it is only used as a temp variable between the time the thread entered the syscall until when the lsm hook is being called, isn't it?

roikol:
you're right, it makes more sense to use pid instead of tgid to avoid a race condition.

tracee-ebpf/tracee/tracee.bpf.c
roikol: because if someone wants to only trace for example 'security_socket_connect' then the 'connect' event wouldn't be traced. in this scenario, if i'll put this if statement after the sould_trace(), the 'sockfd' argument of 'security_socket_connect' will be missing.

yanivagman: When two events depend on each other, we can set an "essential event" which will be traced, but not printed as part of the output. This happen, for example, when you choose magic_write event but not vfs_write, which is needed for magic_write to work. In this case, you can set the connect event if security_socket_connect is chosen. To do that, have a look at how we do this for magic_write in tracee.go, in the function New()

tracee-ebpf/tracee/tracee.bpf.c
roikol: we can assign directly to u32 to get the pid.

"By directly setting this to a u32, we discard the upper 32 bits."
(from here: https://github.com/iovisor/bcc/blob/master/docs/reference_guide.md)

tracee-ebpf/tracee/tracee.go
```diff
@@ -286,6 +286,23 @@ func New(cfg Config) (*Tracee, error) {
...
+
+	if t.eventsToTrace[SecuritySocketAcceptEventID] {
+		setEssential(AcceptEventID)
+		setEssential(Accept4EventID)
+	}
+
+	if t.eventsToTrace[SecuritySocketBindEventID] {
+		setEssential(BindEventID)
+	}
+
```
yanivagman, line 305: Looking at your changes at tracepoint__raw_syscalls__sys_enter() again, I don't think that this is even necessary.
sockfd should be saved on sys_entry because shuold_trace() doesn't check if the event was chosen.

Can you please try to remove the above lines and check if sockfd is saved in the lsm hooks events when accept/connect/listen/bind is not chosen?

roikol: you're right. it works without setting the syscalls as essential to the lsm hooks.

yanivagman (approved):

Issue #721, opened 2021-05-11
libbpfgo: Support first class bpf_program__attach_perf_event


brancz:
It appears that libbpfgo does not have direct support for `bpf_program__attach_perf_event` yet. I assume that's just because tracee doesn't need this functionality. Would the maintainers be open to adding this? I could give it a shot, but have pretty much never worked with these things, so if it's a 10-minute thing for someone else and you can spare the time then by all means please go for it :)

//edit I came from gobpf when I opened this, and I now realized that `perf_event_open` can just be called as a syscall (gobpf wrapped the functionality)

brancz:
I'm not sure at all if I got this right, but I opened #722 in an attempt to add this.

brancz:
Closed by #722

PR #722, opened 2021-05-11
libbpfgo: Add support for AttachPerfEvent

brancz:
An attempt to implement #721 

@yanivagman @grantseltzer


yanivagman:
Ok, Thanks @brancz 
I'll approve this one, as it is similar to the other wrappers we have, and doesn't change any of the existing flows, but rather only adds a new API. 
#723 can be handled by a different PR, and verify that there is no problem with this one.

yanivagman (approved):

brancz:
Sounds good! Will try to put something together for #723!

PR #725, opened 2021-05-12
feat: Tracee Profiler Mode

simar7:
```
$ docker run --name tracee --rm --privileged -v /proc:/proc -v /boot:/boot -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro -v /tmp/tracee:/tmp/tracee -it tracee:latest trace --capture exec --capture profile

<tracee event output as usual>


$ cat /tmp/tracee/out/tracee.profile
{
  "4026531840:/proc/1/root/usr/bin/ls:1610624031864906594": {
    "times": 1,
    "file_hash": "c1d23fb9fd854168e9c1da311eea9e33c7651bf6a858daa57da31d23d370a3dc"
  },
  "4026531840:/proc/1/root/usr/bin/rm:1610624031868906572": {
    "times": 1,
    "file_hash": "24c520ea8b7eda26c3bdbc8824656bd3f69a4260fa88dfea92f5de60f3d59acf"
  },
  "4026531840:/proc/1/root/usr/bin/sudo:1611764875237571229": {
    "times": 1,
    "file_hash": "883be9c50f3e263e7ca4dee9185782e7145777ce9201de7f5cd55adafac172ed"
  }
}
```


entrypoint.sh
```diff
@@ -6,7 +6,7 @@ TRACEE_RULES_EXE=${TRACEE_RULES_EXE:="/tracee/tracee-rules"}
 # handle sub commands
 if [ "$1" = "trace" ]; then
 	shift
-	$TRACEE_EBPF_EXE $@
+	exec $TRACEE_EBPF_EXE $@
```
simar7, line 9: Reason for adding: If Tracee is running inside a container and I try to do the following:
```
docker kill --signal=SIGINT tracee
```
from outside of the container, the SIGINT is not delivered to the Tracee. 

Let me know if there's a better approach than this.

tracee-ebpf/main.go
```diff
@@ -233,12 +233,14 @@ Possible options:
 [artifact:]all                     capture all of the above artifacts.
 
 dir:/path/to/dir        path where tracee will save produced artifacts. the artifact will be saved into an 'out' subdirectory. (default: /tmp/tracee).
+profile                 creates a runtime profile of captured artifacts and their metadata for forensics use. the profile will be saved into the 'out' directory. (default: /tmp/tracee/out/tracee.profile).
```
itaysk, line 236: needless to say where it's captured to since it's documented for all the capture flags. also none of the other ones say it

tracee-ebpf/tracee/tracee.go
```diff
@@ -1108,6 +1151,35 @@ func (t *Tracee) processEvent(ctx *context, args map[argTag]interface{}) error {
...
+		defer f.Close()
+		h := sha256.New()
+		_, _ = io.Copy(h, f)
+		return hex.EncodeToString(h.Sum(nil))
+	}
+	return ""
+}
+
+func (t *Tracee) updateProfile(ctx *context, sourceFilePath string, sourceFileCtime int64) {
+	sourceFileSHA := getFileHash(sourceFilePath)
```
itaysk, line 1166: can we move the hash calculation to the end of the trace please? to minimize overhead in the critical path

tracee-ebpf/main.go
```diff
@@ -285,6 +287,8 @@ Use this flag multiple times to choose multiple capture options
 			if len(outDir) == 0 {
 				return tracee.CaptureConfig{}, fmt.Errorf("capture output dir cannot be empty")
 			}
+		} else if cap == "profile" {
+			capture.Profile = true
```
itaysk, line 291: If I understand correctly, exec is required for profile, and profile without exec is non-functional (correct me if I'm wrong). In this case we should automatically set `capture.Exec=true` here as well.

tracee-ebpf/tracee/tracee.go
```diff
@@ -1108,6 +1151,35 @@ func (t *Tracee) processEvent(ctx *context, args map[argTag]interface{}) error {
...
+	if pf, ok := t.profiledFiles[sourceFilePath]; !ok {
```
itaysk, line 1167: sourceFilePath is not a good key for this map, as files may be overwritten. I think a decent key can be capturedFileID + sourceFileCtime

tracee-ebpf/tracee/tracee.go
```diff
@@ -1108,6 +1151,35 @@ func (t *Tracee) processEvent(ctx *context, args map[argTag]interface{}) error {
...
+		t.profiledFiles[sourceFilePath] = profilerInfo{
+			MountNS:   ctx.MntID,
+			TimeStamp: sourceFileCtime,
+			Times:     1,
+			FileHash:  sourceFileSHA,
```
itaysk, line 1172: If you agree and apply the rest of my comments, then the profilerInfo struct is unneeded. The map value can be just "times" (int)

simar7: What about the file hash?

tracee-ebpf/main.go
simar7: OK

tracee-ebpf/main.go
simar7: OK

tracee-ebpf/tracee/tracee.go
simar7: Do you mean call `getFileHash` after the pipeline ends but before we save the stats to disk? If so, that would leave a window of time open when the event is captured till the profile is written on disk. During this time the file can be modified and we wouldn't know about the hash prior to modification.

Capturing hash at event capture time mitigates the above.

itaysk: we already capture the binaries at the time of execution (-c exec). they could be used for calculating the hash right?

tracee-ebpf/tracee/tracee.go
itaysk: see https://github.com/aquasecurity/tracee/pull/725/files#r636341189

tracee-ebpf/tracee/tracee.go
simar7: I see yeah that would work as well. Taking the hash of captured artifacts. Although there is a possibility that something malicious removes those captured artifacts as soon as they are created (as the location is known) but it's a sidetrack. 

I'll implement taking the hash of captured objects for now as you said.

tracee-ebpf/tracee/tracee.go
simar7: I'm not sure if I follow. Are you suggesting we remove the SHA altogether from the profile thats saved to disk?

entrypoint.sh
simar7: Created https://github.com/aquasecurity/tracee/issues/736 to track it

tracee-ebpf/tracee/tracee.go
simar7: De-coupled from main loop https://github.com/aquasecurity/tracee/pull/725/commits/93f7ce2aae884a80a52aad0590387e9042770676

tracee-ebpf/main.go
simar7: https://github.com/aquasecurity/tracee/pull/725/commits/718b24868f23b657de987413eebf28c956137c67

tracee-ebpf/main.go
simar7: https://github.com/aquasecurity/tracee/pull/725/commits/718b24868f23b657de987413eebf28c956137c67

tracee-ebpf/tracee/tracee.go
simar7: As discussed offline https://github.com/aquasecurity/tracee/pull/725/commits/93f7ce2aae884a80a52aad0590387e9042770676

tracee-ebpf/tracee/tracee.go
```diff
@@ -173,6 +177,11 @@ func (tc Config) Validate() error {
 	return nil
 }
 
+type profilerInfo struct {
+	Times    int64  `json:"times,omitempty"`
```
simar7, line 181: I'm considering to remove `Times`. Sometimes, Tracee startup is slower than expected and as a result we miss on events which cause noise.
image


I think it's not very useful to know how many times a program was executed. Let me know if you think otherwise.

simar7: New output:
```
 cat /tmp/tracee/out/tracee.profile
{
  "4026531840:/proc/1/root/snap/microk8s/2213/sbin/xtables-multi:1620854501000000000": {
    "file_hash": "474a0724ecba3b22a267810f2b0c10c792d3f2090464e2d614b59959491e59b9"
  },
  "4026531840:/proc/1/root/usr/sbin/sshd:1615393641978322322": {
    "file_hash": "9169f72f64ff6e0cc7597449b90312889275c613b4fcec390a852683cdd085d0"
  }
}
```

tracee-ebpf/tracee/tracee.go
```diff
@@ -1085,9 +1122,16 @@ func (t *Tracee) processEvent(ctx *context, args map[argTag]interface{}) error {
 					//TODO: remove dead pid from cache
 					continue
 				}
-				//don't capture same file twice unless it was modified
+
 				sourceFileCtime := sourceFileStat.Sys().(*syscall.Stat_t).Ctim.Nano()
 				capturedFileID := fmt.Sprintf("%d:%s", ctx.MntID, sourceFilePath)
+
+				// create an in-memory profile
+				if t.config.Capture.Profile {
```
yanivagman, line 1130: Why not to put this after the check of already captured file below?
This will prevent unnecessary processing for duplications

tracee-ebpf/tracee/tracee.go
```diff
@@ -1108,6 +1152,29 @@ func (t *Tracee) processEvent(ctx *context, args map[argTag]interface{}) error {
...
+		defer f.Close()
+		h := sha256.New()
+		_, _ = io.Copy(h, f)
+		return hex.EncodeToString(h.Sum(nil))
+	}
+	return ""
+}
+
+func (t *Tracee) updateProfile(sourceFilePath string) {
+	t.profiledFiles[sourceFilePath] = profilerInfo{}
```
yanivagman, line 1167: What if the same file was written twice, but with different content? We will only capture the last instance of written file, won't we? By adding the ctime to the key of the map, we can save all of the relevant instances

tracee-ebpf/tracee/tracee.go
```diff
@@ -1108,6 +1152,29 @@ func (t *Tracee) processEvent(ctx *context, args map[argTag]interface{}) error {
 	return nil
 }
 
+func getFileHash(fileName string) string {
```
yanivagman, line 1155: UpdateFileSHA, and thus this function as well, is called on terminating tracee.
At this point in time, sourceFilePath, which is given as an argument to updateProfile and is composed from a path in "/proc/pid/root/..." might not be available.
Instead, as we already capture the executed files to destinationFilePath), we can perform the hash calculations on these captured files

yanivagman (changes requested):

tracee-ebpf/main.go
```diff
@@ -233,12 +233,14 @@ Possible options:
 [artifact:]all                     capture all of the above artifacts.
 
 dir:/path/to/dir        path where tracee will save produced artifacts. the artifact will be saved into an 'out' subdirectory. (default: /tmp/tracee).
+profile                 creates a runtime profile of captured artifacts and their metadata for forensics use. the profile will be saved into the 'out' directory.
```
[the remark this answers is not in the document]
itaysk: I agree, "profile of captured artifacts" is not the best description. maybe "profile of program executions" or something similar

tracee-ebpf/tracee/tracee.go
itaysk: I believe we said we will wait for tracee to start (tracee.pid created) before moving on

tracee-ebpf/main.go
itaysk: not just the default also the mention of the output directory. like the other descriptions

tracee-ebpf/tracee/tracee.go
itaysk: yes this was the intention with https://github.com/aquasecurity/tracee/pull/725#discussion_r637391823

tracee-ebpf/tracee/tracee.go
itaysk: see https://github.com/aquasecurity/tracee/pull/725/files/4cb2e9a8f852d01a77429611a5dd0e525e28d87f#r642904379

tracee-ebpf/tracee/tracee.go
itaysk: `sourceFilePath` does include the ctime after https://github.com/aquasecurity/tracee/pull/725#discussion_r636348145 
(see L1131). it's just a bad var name (let's rename it)

tracee-ebpf/tracee/tracee.go
yanivagman: Are you sure?
`updateProfile()` is being called as follows:
`t.updateProfile(fmt.Sprintf("%s:%d", capturedFileID, sourceFileCtime))`
were capturedFileID equals:
`fmt.Sprintf("%d:%s", ctx.MntID, sourceFilePath)`
so in total we get:
`t.updateProfile(fmt.Sprintf(""%d:%s:%d", ctx.MntID, sourceFilePath, sourceFileCtime))`
Which means that sourceFilePath is saved with a value of "/proc/pid/root/...", isn't it?

tracee-ebpf/tracee/tracee.go
simar7: OK I tested this approach and it seems to work. I will add the times back as discussed offline.

tracee-ebpf/main.go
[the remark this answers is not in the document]
simar7: OK updated

tracee-ebpf/tracee/tracee.go
simar7: You're right. I've updated this.

tracee-ebpf/tracee/tracee.go
itaysk: yes, I wrote this was the _intention_ of the comment I made. I didn't see it implemented yet.

tracee-ebpf/main.go
```diff
@@ -233,12 +233,14 @@ Possible options:
...
 
 dir:/path/to/dir        path where tracee will save produced artifacts. the artifact will be saved into an 'out' subdirectory. (default: /tmp/tracee).
+profile                 creates a runtime profile of program executions and their metadata for forensics use.
 clear-dir               clear the captured artifacts output dir before starting (default: false).
 
 Examples:
   --capture exec                                           | capture executed files into the default output directory
   --capture all --capture dir:/my/dir --capture clear-dir  | delete /my/dir/out and then capture all supported artifacts into it
   --capture write=/usr/bin/* --capture write=/etc/*        | capture files that were written into anywhere under /usr/bin/ or /etc/
+  --capture exec --capture profile                         | captures files executed files and create a runtime profile in the output directory
```
yanivagman, line 243: extra "files" in description "captures files executed files..."

tracee-ebpf/tracee/tracee.go
simar7: Actually we need this for a different reason. If we were only profiling the captured files once your suggestion will be valid. But in our case we need to also include the number of times a file was executed for which we cannot skip over duplicates.

yanivagman: Well, when I wrote this comment, "Times" was not part of the profiling...

tracee-ebpf/tracee/tracee.go
simar7: I added some logic that uses captured file info instead of using the "/proc/pid/root" files. https://github.com/aquasecurity/tracee/pull/725/commits/25c83dfb295b1346c7cb2c040196fe4dcd7a18b0

tracee-ebpf/main.go
simar7: fixed

tracee-ebpf/tracee/tracee.go
```diff
@@ -1108,6 +1155,40 @@ func (t *Tracee) processEvent(ctx *context, args map[argTag]interface{}) error {
...
+		return hex.EncodeToString(h.Sum(nil))
+	}
+	return ""
+}
+
+func (t *Tracee) updateProfile(sourceFilePath string, executionTs uint64) {
+	if pf, ok := t.profiledFiles[sourceFilePath]; !ok {
+		t.profiledFiles[sourceFilePath] = profilerInfo{
+			Times:            1,
+			FirstExecutionTs: executionTs,
```
itaysk, line 1173: won't this break the profile? meaning we it won't be comparable to previous run

tracee-ebpf/tracee/tracee.go
```diff
@@ -1085,13 +1124,21 @@ func (t *Tracee) processEvent(ctx *context, args map[argTag]interface{}) error {
...
 					continue
 				}
-				//don't capture same file twice unless it was modified
+
 				sourceFileCtime := sourceFileStat.Sys().(*syscall.Stat_t).Ctim.Nano()
 				capturedFileID := fmt.Sprintf("%d:%s", ctx.MntID, sourceFilePath)
+
+				// create an in-memory profile
+				if t.config.Capture.Profile {
+					t.updateProfile(fmt.Sprintf("%s:%d", filepath.Join(destinationDirPath, fmt.Sprintf("exec.%s", filepath.Base(filePath))), sourceFileCtime), ctx.Ts)
```
itaysk, line 1133: can we use the already calculated destinationFilePath instead of this?

tracee-ebpf/tracee/tracee.go
simar7: Good point but no, it's explicitly filtered out in the JSON that's saved https://github.com/aquasecurity/tracee/pull/725/files#diff-710a36e0ca7a07ea49cf4970d916ed0b74b1c3619920145d9db50477c4f96967R183

tracee-ebpf/tracee/tracee.go
simar7: Not quite.
```
destinationFilePath := filepath.Join(destinationDirPath, fmt.Sprintf("exec.%d.%s", ctx.Ts, filepath.Base(filePath)))
```
destinationFilePath includes ctx.Ts. If we use it as is, each invocation (of the same binary) will create a new entry in the map.

simar7:
@itaysk and @yanivagman any other feedback before we can merge this in?

tracee-ebpf/tracee/tracee.go
```diff
@@ -1108,6 +1155,40 @@ func (t *Tracee) processEvent(ctx *context, args map[argTag]interface{}) error {
...
+		h := sha256.New()
+		_, _ = io.Copy(h, f)
+		return hex.EncodeToString(h.Sum(nil))
+	}
+	return ""
+}
+
+func (t *Tracee) updateProfile(sourceFilePath string, executionTs uint64) {
+	if pf, ok := t.profiledFiles[sourceFilePath]; !ok {
+		t.profiledFiles[sourceFilePath] = profilerInfo{
```
yanivagman, line 1171: Possible race condition when updating this map

tracee-ebpf/tracee/tracee.go
itaysk: A simplification suggestion (not for this PR):
The ctx.Ts is included in the captured file name in order to distinguish between different captured files with same path name.
IIUC you track the FirstExecutionTs field in order to reconstruct the same file name when calculating the hash.
Suggestion is to replace the ctx.Ts in captured file names with the file's ctime (it makes more sense regardless of this PR even), and then the unique id of a captured/profiled file includes just the mntid, path, and ctime. the profile map from destinationFilePath to executions counter and when you calculate the hash you can just iterate over the map/files without reconstructing file names which might be error prone (and also simplify the code flow as no need to track ctx.Ts in the map and pass it around).
Let merge this PR and maybe discuss this in the future

itaysk:
A simplification suggestion (not for this PR).
The ctx.Ts is included in the captured file name in order to distinguish between different captured files with same path name.
IIUC you track the FirstExecutionTs field in order to reconstruct the same file name when calculating the hash.

Suggestion is to replace the ctx.Ts in captured file names with the file's ctime (it makes more sense regardless of this PR even), and then the unique id of a captured/profiled file includes just the mntid, path, and ctime and is consistent between the file names (which you use for hash calculation), and the map keys (which you use for the profile).
The profile maps from destinationFilePath to executions counter (and hash if needed). This eliminates reconstructing the file names which might be error prone, and simplify the code flow as no need to track ctx.Ts in the map and pass it around.

itaysk (approved):

tracee-ebpf/tracee/tracee.go
simar7: I tried to think about the various ways today and there are some that I can think of (for same key):

1. Write after Read hazard: This map could be written before reading. I believe this is acceptable.
2. Read after Write hazard. This map is read prior to being written. In this case we would under count the items. This is an issue.
3. Write after Write hazard: This is fine since the counter is only incrementing up.

But regardless, your suggestion made me use the race detector and I found two unrelated races elsewhere in the code which I filed an issue for here https://github.com/aquasecurity/tracee/issues/756

simar7:
> A simplification suggestion (not for this PR).
> The ctx.Ts is included in the captured file name in order to distinguish between different captured files with same path name.
> IIUC you track the FirstExecutionTs field in order to reconstruct the same file name when calculating the hash.
> 
> Suggestion is to replace the ctx.Ts in captured file names with the file's ctime (it makes more sense regardless of this PR even), and then the unique id of a captured/profiled file includes just the mntid, path, and ctime and is consistent between the file names (which you use for hash calculation), and the map keys (which you use for the profile).
> The profile maps from destinationFilePath to executions counter (and hash if needed). This eliminates reconstructing the file names which might be error prone, and simplify the code flow as no need to track ctx.Ts in the map and pass it around.

Right this is a good improvement. I'll create an issue to track it: https://github.com/aquasecurity/tracee/issues/757

tracee-ebpf/tracee/tracee.go
yanivagman: Yes, I was referring to case number 2.
I think that a simple mutex for writing will solve this problem, so it is only needed in this function.

simar7: I thought about this a bit more. Since the read is done once per invocation of this function I don't think there's a race condition here as we only increment the Times counter. 

Here's a simple race test I wrote to demonstrate it:
```
// +build race !unit

package tracee

import (
	"fmt"
	"io/ioutil"
	"os"
	"path/filepath"
	"sync"
	"testing"

	"github.com/stretchr/testify/require"
)

func Test_updateProfile_race(t *testing.T) {
	trc := Tracee{
		profiledFiles: make(map[string]profilerInfo),
	}

	d, err := ioutil.TempDir("", "Test_updateProfile_dir-*")
	require.NoError(t, err)

	f, err := ioutil.TempFile(d, "Test_updateProfile-*")
	require.NoError(t, err)
	defer os.RemoveAll(f.Name())

	captureFileID := fmt.Sprintf("%s.%s", d, filepath.Base(f.Name()))

	//run many concurrent writes
	var wgW sync.WaitGroup
	wgW.Add(10)
	go func() {
		for i := 0; i < 10; i++ {
			trc.updateProfile(captureFileID, 123)
			wgW.Done()
		}
	}()
	wgW.Wait()

	require.Equal(t, profilerInfo{
		Times:            10,
		FirstExecutionTs: 123,
	}, trc.profiledFiles[captureFileID])
}
```

And the output:
```
GOOS=linux GOARCH=amd64 CC=clang CGO_CFLAGS="-I /usr/vagrant/repos/tracee/tracee-ebpf/dist/libbpf/usr/include" CGO_LDFLAGS="/usr/vagrant/repos/tracee/tracee-ebpf/dist/libbpf/libbpf.a" go test -tags=race -race -v ./tracee
=== RUN   Test_updateProfile_race
--- PASS: Test_updateProfile_race (0.00s)
PASS
ok      github.com/aquasecurity/tracee/tracee-ebpf/tracee       0.021s

```

yanivagman: I think that 10 counts is not enough to see the race condition.
Here is the case I was thinking about:

Thread A:
1. calls updateProfile
2. pf, ok := t.profiledFiles[sourceFilePath]

Thread B:
1. calls updateProfile
2. pf, ok := t.profiledFiles[sourceFilePath]
3. pf.Times = pf.Times + 1 
4. t.profiledFiles[sourceFilePath] = pf

Thread A:
1. pf.Times = pf.Times + 1 
2. t.profiledFiles[sourceFilePath] = pf

After this sequence, pf.Times will be under count by 1.

I don't think that this issue is that important, as there is nothing functional here (only profiling), but if it is something that can be easilly fixed - why not?

itaysk: @yanivagman what are thread A and thread B? updateProfile is called from ProcessEvent which is a single thread. Can you please elaborate?

yanivagman: You are right. My mistake.

yanivagman (approved):

Issue #723, opened 2021-05-11
Support map iterator


brancz:
It appears that currently only simple get/update/delete functions are available on maps. Other libraries such as libbpf-rs appear to [implement iterators](https://github.com/libbpf/libbpf-rs/blob/55b681dfd5b723005da8e79bd3472070f9260f5a/libbpf-rs/src/map.rs#L374-L406).

Would maintainers be open to having a map iterator implementation?

PR #728, opened 2021-05-12
libbpfgo: Add map iterator support

brancz:
Closes https://github.com/aquasecurity/tracee/issues/723

This appears to work for my use case, I'll contribute fixes should I notice any bugs with this as is. It's mostly modeled after what libbpf-rs does.

@yanivagman @grantseltzer

### DIFF
--- a/libbpfgo/libbpfgo.go
+++ b/libbpfgo/libbpfgo.go
@@ -478,6 +478,62 @@ func (b *BPFMap) Update(key, value interface{}) error {
 	return nil
 }
 
+type BPFMapIterator struct {
+	b       *BPFMap
+	err     error
+	keySize int
+	prev    []byte
+	next    []byte
+}
+
+func (b *BPFMap) Iterator(keySize int) *BPFMapIterator {
+	return &BPFMapIterator{
+		b:       b,
+		keySize: keySize,
+		prev:    nil,
+		next:    nil,
+	}
+}
+
+func (it *BPFMapIterator) Next() bool {
+	if it.err != nil {
+		return false
+	}
+
+	prevPtr := unsafe.Pointer(nil)
+	if it.next != nil {
+		prevPtr = unsafe.Pointer(&it.next[0])
+	}
+
+	next := make([]byte, it.keySize)
+	nextPtr := unsafe.Pointer(&next[0])
+
+	errC, err := C.bpf_map_get_next_key(it.b.fd, prevPtr, nextPtr)
+	if errno, ok := err.(syscall.Errno); errC == -1 && ok && errno == C.ENOENT {
+		return false
+	}
+	if err != nil {
+		it.err = err
+		return false
+	}
+
+	it.prev = it.next
+	it.next = next
+
+	return true
+}
+
+// Key returns the current key value of the iterator, if the most recent call to Next returned true.
+// The slice is valid only until the next call to Next.
+func (it *BPFMapIterator) Key() []byte {
+	return it.next
+}
+
+// Err returns the last error that ocurred while table.Iter or iter.Next
+func (it *BPFMapIterator) Err() error {
+	return it.err
+}
+
 func (m *Module) GetProgram(progName string) (*BPFProg, error) {
 	cs := C.CString(progName)
 	prog := C.bpf_object__find_program_by_name(m.obj, cs)


libbpfgo/libbpfgo.go
```diff
@@ -478,6 +478,67 @@ func (b *BPFMap) Update(key, value interface{}) error {
...
+		prevPtr = unsafe.Pointer(&it.next[0])
+	}
+
+	next := make([]byte, it.keySize)
+	nextPtr := unsafe.Pointer(&next[0])
+
+	errC := C.bpf_map_get_next_key(it.b.fd, prevPtr, nextPtr)
+	if errC == -1 {
+		it.prev = it.next
+		it.next = next
```
yanivagman, line 517: Quoting from the man page (https://man7.org/linux/man-pages/man2/bpf.2.html):

> If key is the last element, -1 is returned and errno is set to ENOENT. 
> Other possible errno values are ENOMEM, EFAULT, EPERM, and EINVAL.

So, is it correct to do the above lines in case errno != ENOENT?

libbpfgo/libbpfgo.go
```diff
@@ -478,6 +478,67 @@ func (b *BPFMap) Update(key, value interface{}) error {
...
+		return false
+	}
+	if errC != 0 {
```
yanivagman, line 520: According to the man page, there are only two possible return values: 0 or -1.
If so, how can this line ever evaluate to true (since we already checked for `errC == -1` above)?

yanivagman (commented):
Thanks @brancz,
overall looks good, but I'm not sure that returned errors are handled correctly

libbpfgo/libbpfgo.go
brancz: I think you're right, we're missing the last iteration as is, let me fix that!

brancz:
I think this should be correct now. I tested with this code:

```go
    numbers, err := m.GetMap("numbers")                                                                                                                                                                                                          
    if err != nil {                                                                                                                                                                                                                           
        return err                                                                                                                                                                                                                            
    }                                                                                                                                                                                                                                         
    one := 1                                                                                                                                                                                                                                  
    err = numbers.Update(uint32(1), uint32(one))                                                                                                                                                                                              
    if err != nil {                                                                                                                                                                                                                           
        return err                                                                                                                                                                                                                            
    }                                                                                                                                                                                                                                         
    err = numbers.Update(uint32(2), uint32(one))                                                                                                                                                                                              
    if err != nil {                                                                                                                                                                                                                           
        return err                                                                                                                                                                                                                            
    }                                                                                                                                                                                                                                         
    err = numbers.Update(uint32(3), uint32(one))                                                                                                                                                                                              
    if err != nil {                                                                                                                                                                                                                           
        return err                                                                                                                                                                                                                            
    }                                                                                                                                                                                                                                         
    it := numbers.Iter(4)                                                                                                                                                                                                                     
    byteOrder := GetHostByteOrder()                                                                                                                                                                                                           
    for it.Next() {                                                                                                                                                                                                                           
        keyBytes := it.Key()                                                                                                                                                                                                                  
        log.Println(byteOrder.Uint32(keyBytes))                                                                                                                                                                                               
    }                                                                                                                                                                                                                                         
    if it.Err() != nil {                                                                                                                                                                                                                      
        return it.Err()                                                                                                                                                                                                                       
    }
```

Which printed:

```
2021/05/14 09:08:07 1                                                                                                  
2021/05/14 09:08:07 3                                                                                                  
2021/05/14 09:08:07 2
```

libbpfgo/libbpfgo.go
```diff
@@ -478,6 +478,65 @@ func (b *BPFMap) Update(key, value interface{}) error {
...
+
+	prevPtr := unsafe.Pointer(nil)
+	if it.next != nil {
+		prevPtr = unsafe.Pointer(&it.next[0])
+	}
+
+	next := make([]byte, it.keySize)
+	nextPtr := unsafe.Pointer(&next[0])
+
+	errC, err := C.bpf_map_get_next_key(it.b.fd, prevPtr, nextPtr)
```
grantseltzer, line 514: How does have two return values work with CGO calls work?

libbpfgo/libbpfgo.go
```diff
@@ -478,6 +478,65 @@ func (b *BPFMap) Update(key, value interface{}) error {
...
+}
+
+// Key returns the current key value of the iterator, if the most recent call to Next returned true.
+// The slice is valid only until the next call to Next.
+func (it *BPFMapIterator) Key() []byte {
+	return it.next
+}
+
+// Err returns the last error that ocurred while table.Iter or iter.Next
+func (it *BPFMapIterator) Err() error {
```
grantseltzer, line 536: Why have this instead of returning it from `Next`?

libbpfgo/libbpfgo.go
```diff
@@ -478,6 +478,65 @@ func (b *BPFMap) Update(key, value interface{}) error {
...
+func (b *BPFMap) Iter(keySize int) *BPFMapIterator {
+	return &BPFMapIterator{
+		b:       b,
+		keySize: keySize,
+		prev:    nil,
+		next:    nil,
+	}
+}
+
+func (it *BPFMapIterator) Next() bool {
```
grantseltzer, line 501: What is the `bool` return meant to represent?

libbpfgo/libbpfgo.go
```diff
@@ -478,6 +478,65 @@ func (b *BPFMap) Update(key, value interface{}) error {
...
+
+	err error
+
+	keySize int
+
+	prev []byte
+	next []byte
+}
+
+func (b *BPFMap) Iter(keySize int) *BPFMapIterator {
```
grantseltzer, line 492: Iter -> Iterator

grantseltzer (changes requested):
Having this would be great and partly enables the global data feature request you created, so thank you very much!

A few comments about API design, lmkwyt

grantseltzer:
I'm also curious if you can share what your use case is, entirely out of curiosity.

libbpfgo/libbpfgo.go
brancz: Excerpt from the cgo docs:

> Any C function (even void functions) may be called in a multiple assignment context to retrieve both the return value (if any) and the C errno variable as an error (use _ to skip the result value if the function returns void). For example:
>
> n, err = C.sqrt(-1)
> _, err := C.voidFunc()
> var n, err = C.sqrt(1)

libbpfgo/libbpfgo.go
brancz: This is the typical iterator pattern in go, this way the usage works like this:

```go
for iterator.Next() {
  // do something with the current value
}
if iterator.Err() != nil {
  // handle error
}
```

libbpfgo/libbpfgo.go
brancz: Same as the above statement :)

brancz:
I'm trying to gather statistics about observed kernel stack traces and I need to iterate over all observed stack traces for that.

libbpfgo/libbpfgo.go
grantseltzer: That makes sense but i'm not convinced that this pattern should be used in an API. It makes it a bit opinionated on how it should be used. I'd rather opt to have as much conformity to libbpf itself as possible.

i.e. I think this would be fine:

```
for err := iterator.Next(); if err == nil { 

} else {
   // handle error
}
```

eyakubovich: FWIW, I don't think what @brancz proposes is very opinionated. Then the only difference is having to check the boolean (return value of Iterator.Next) before checking the error. However with the idiomatic usage of checking the error at the end, it's nicer to see `for iterator.Next() {`

grantseltzer: Perhaps it's me that is opinionated then :-) 

To be frank I've always felt that Go shouldn't use any design patterns. Otherwise the whole package has to use the same pattern where applicable and projects that consume it will feel the need to conform to the same pattern. If anything, I think the most idiomatic way/pattern of doing iteration in Go is using callbacks. 

What if we just had `iterator.Next()` return both the key and the error, and got rid of those two methods. Is there any reason why someone would want to call `Next()` and not immediately check both the error and value?

I would also be happy to have a callback parameter to `Next`, in this way we could eliminate having the specify a KeySize when initializing the iterator altogether which further simplifies the API.

brancz: I just applied what I know and expect from go packages (which could be opinionated due to my experience, but is widely used like this in prometheus tsdb/cockroachdb pebble/various networking libraries eg. [netlink](https://github.com/mdlayher/netlink)).

grantseltzer: Yea I can't find any example of usage in the standard library so i'd prefer just the simple approach of having `Next()` return the key and the error. If you can make that change I think this is good to go. I'll open up an issue for writing a selftest for this.

I opened up the new libbpfgo repository. I should be able to apply your commit on top of that. If there's any problem with that I'll let you know so that you get the proper credit for it. After that i'll cut the first release.

grantseltzer: Also thank you very much for this contribution and your patience 😀

yanivagman: @grantseltzer I agree that staying as close as possible to libbpf and making thin wrappers is usually preferrable.
In that case, however, I don't think that we loose anything, and the code is still generic (isn't it?).
I also like that `for iterator.Next() {` can be used as @eyakubovich pointed out.

About having a callback to eliminate the need for KeySize, maybe there is another option - using libbpf's `bpf_map__def()` API, we can get the map def:
```
struct bpf_map_def {	
  unsigned int type;	
  unsigned int key_size;	
  unsigned int value_size;	
  unsigned int max_entries;	
  unsigned int map_flags;
};
```

We can even add this to the BPFMap struct, and use it when needed for other places (e.g. for map GetValue() where we also provide the value size)

grantseltzer: @yanivagman Fair enough. I concede about changing the iterator pattern. @brancz can you please just add a godoc comment above each of the API functions to clarify how it's expected to be used (like what the bool return represents)

Using [`bpf_map_def`](https://github.com/libbpf/libbpf/blob/master/src/libbpf.h#L439) would be fantastic, maybe that should that be part of this PR or have a subsequent change change this, and `GetValue()` as well.

yanivagman (approved):

libbpfgo/libbpfgo.go
brancz: Happy to do a follow up with `bpf_map_def`, I'll likely do more for batch operations as well.

grantseltzer (approved):
LGTM! Thank you very much @brancz!

Please hold off on opening more PRs for the next couple hours until I have the new libbpfgo repo in place. I'm going to redo what I previously opened so I can preserve all history.

brancz:
My work day is pretty much over here anyways, so any further contributions are easily 14h away :)

PR #738, opened 2021-05-30
switch to libbpf v0.4

yanivagman:


grantseltzer:
Works for me! 

Related is this issue https://github.com/aquasecurity/libbpfgo/issues/17. I'm going to open a PR for libbpfgo so that it links against local libbpf modules for the selftests, but I tested locally with 0.4 and there's no issues as far as I can tell.

grantseltzer (approved):

PR #740, opened 2021-06-01
tracee-ebpf: commit_creds: submit more credentials

yanivagman:


grantseltzer (approved):
This looks good to me, we should have a built-in signature in tracee-rules that consumes commit_creds so users can make sense of these new values. Would make for a great demo.

yanivagman:
> This looks good to me, we should have a built-in signature in tracee-rules that consumes commit_creds so users can make sense of these new values. Would make for a great demo.

This is the plan :-)

Issue #430, opened 2021-01-20
Enrich container events with image name and container id


yanivagman:
Container events metadata can be enriched by interfacing with the container runtime.
The two most useful fields are container id and image name, which can then be used, for example, by tracee-rules, and also help with the implementation of #255.
To get this metadata, Tracee can interface with the container runtime using CRI.
The implementation should be able to get the metadata before the container starts its execution.
For cri-o, this can be done by using a prestart hook, which is a binary that runs before the container starts. 
As containerd doesn't support the configuration of prestart hooks, we can use its container events mechanism, that produces an event when a new container is created.
We can then map the mount namespace that we get from the kernel with the collected metadata we have from the container runtime, and add the extra metadata to the event as required.

simar7:
> To get this metadata, Tracee can interface with the container runtime using CRI.

Are you sure CRI can support this? My understanding after asking a few people about it that containers created via higher level tools like docker or ctr aren't visible via CRI. I opened an issue to confirm this: https://github.com/kubernetes-sigs/cri-tools/issues/758

One idea that was suggested on slack was to use something like runc to get hold of all the containers that were created using it. Upon trying it there seems to be some success in that:
```
sudo runc --root=/run/docker/runtime-runc/moby list
ID                                                                 PID         STATUS      BUNDLE                                                                                                                CREATED                          OWNER
bd538a379da8fb4553272722d2c510092d5b5a104a1faa8c4494d3feb21c0039   117889      running     /run/containerd/io.containerd.runtime.v2.task/moby/bd538a379da8fb4553272722d2c510092d5b5a104a1faa8c4494d3feb21c0039   2021-04-23T19:26:33.232357786Z   root
```

Thoughts?

yanivagman:
CRI is an interface used by k8s to interact with container runtimes (docker, containerd, cri-o all implement this interface).
If a container is created by k8s (configured to use containerd by default if I'm not wrong), the container state metadata will be saved by this containerd instance. In case that the containers you tried to list were started with docker, it may be that a different containerd instance was used to create these containers, thus are not presented by crictl. (crictl can be given a unix socket of the container runtime to interact with).

If the same container runtime is used for both docker and k8s, it may be that the problem you encountered is containerd namespaces as described here:
https://github.com/containerd/containerd/issues/1815

The problem with using runc to solve our problem is that it does not have any information about the container image, because it is not its responsibility. runc is only responsible for the runtime related stuff, and not for images.

[...]

PR #749, opened 2021-06-07
tracee-ebpf: add container id to context

yanivagman:
Partially covers #430

### DIFF
--- a/tracee-ebpf/tracee/printer.go
+++ b/tracee-ebpf/tracee/printer.go
@@ -85,6 +85,7 @@ func newEvent(ctx context, argMetas []external.ArgMeta, args []interface{}, Stac
 		PIDNS:               int(ctx.PidID),
 		ProcessName:         string(bytes.TrimRight(ctx.Comm[:], "\x00")),
 		HostName:            string(bytes.TrimRight(ctx.UtsName[:], "\x00")),
+		ContainerID:         string(bytes.TrimRight(ctx.ContID[:], "\x00")),
 		EventID:             int(ctx.EventID),
 		EventName:           EventsIDToEvent[int32(ctx.EventID)].Name,
 		ArgsNum:             int(ctx.Argnum),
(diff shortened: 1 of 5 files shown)


tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -203,6 +205,8 @@
 
 #define DEV_NULL_STR    0
 
+#define CONT_ID_LEN 12
```
simar7, line 208: Curious why is this 12?

yanivagman: I could have chosen a bigger value, but I don't think that this is necessary to identify a container.
Docker also presents the first 12 chars in its output so I used the same convention here.

simar7 (approved):

tracee-ebpf/tracee/printer.go
```diff
@@ -114,13 +115,13 @@ func (p tableEventPrinter) Init() error { return nil }
 func (p tableEventPrinter) Preamble() {
 	if p.verbose {
 		if p.containerMode {
-			fmt.Fprintf(p.out, "%-14s %-16s %-12s %-12s %-6s %-16s %-15s %-15s %-15s %-16s %-20s %s", "TIME(s)", "UTS_NAME", "MNT_NS", "PID_NS", "UID", "COMM", "PID/host", "TID/host", "PPID/host", "RET", "EVENT", "ARGS")
```
itaysk, line 117: @yanivagman don't you think hostname is redundant in container mode now that we have container id?

itaysk:
@yanivagman this will not work with pre-existing container right?

itaysk:
just sharing a thought publicly (not asking for any change), there are some similarities between this code and the "capture exec" code, in the future we should consider if we once could benefit form the other

tracee-ebpf/tracee/printer.go
yanivagman: Yes, we can leave it to verbose only

yanivagman:
`@yanivagman this will not work with pre-existing container right?`
Correct. We should handle this case differently.
A simple approach to handle this will be to update the new "pid_to_cont_id_map" map on Tracee init.
This will have a small race condition though (if a new container starts/exits while we init tracee)

yanivagman:
> just sharing a thought publicly (not asking for any change), there are some similarities between this code and the "capture exec" code, in the future we should consider if we once could benefit form the other

Which similarities?

PR #753, opened 2021-06-08
fix gob type declaration

### LINKED ISSUE #748: Commit_creds event resulting and unexpected output while redirecting it to Tracee-rules
Hey,
I tried writing a simple signature based on "commit_creds" event using rego.
While executing tracce-ebpf with the flags " -t event=commit_creds" everything works fine.
But, redirecting the output into Tracee-rules results in some kind of an error + the signatures does not math on the event.
MicrosoftTeams-image (2)

Steps to reproduce:
clone tracee from main (v5.2)
*compile using* : "make DOCKER=1"
*run* : "./tracee-ebpf/dist/tracee-ebpf -o format:gob -t event=commit_creds | ./tracee-rules/dist/tracee-rules --input-tracee file:stdin --input-tracee format:gob --output-template=../templates/rawjson.tmpl"
*open new terminal and run*: "sudo ps"

itaysk:
fix #748

### DIFF
--- a/tracee-ebpf/tracee/printer.go
+++ b/tracee-ebpf/tracee/printer.go
@@ -250,6 +250,8 @@ type gobEventPrinter struct {
 
 func (p *gobEventPrinter) Init() error {
 	p.outEnc = gob.NewEncoder(p.out)
+	gob.Register(external.Event{})
+	gob.Register(external.SlimCred{})
 	p.errEnc = gob.NewEncoder(p.err)
 	return nil
 }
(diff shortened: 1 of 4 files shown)


yanivagman:
Shouldn't we update the go.mod of tracee-rules to point to the new tracee-ebpf so that external.SlimCred will be known by it?

itaysk:
Yeah I forgot to commit that change. Will fix

yanivagman (approved):

PR #754, opened 2021-06-08
Helper functions

michaelkatch:


yanivagman:
Some of these functions (e.g. drop_root_permissions, process_capabilities_changed...) seem to be specific to a given signature, and I'm not sure that they fit the definition of a "helper". Do you see them being used by more than one signature?

itaysk:
@michaelkatch are you still on this?

PR #758, opened 2021-06-09
Optimize save_path_to_str_buf in tracee.bpf.c

grantseltzer:
Since the large loop gets unrolled, the various variable
allocations caused programs to exceed the stack limit.

### DIFF
--- a/tracee-ebpf/tracee/tracee.bpf.c
+++ b/tracee-ebpf/tracee/tracee.bpf.c
@@ -1172,38 +1172,47 @@ static __always_inline int save_path_to_str_buf(buf_t *string_p, const struct pa
     int zero = 0;
     struct dentry *dentry = f_path.dentry;
     struct vfsmount *vfsmnt = f_path.mnt;
+    struct mount *mnt_parent_p;
+
     struct mount *mnt_p = real_mount(vfsmnt);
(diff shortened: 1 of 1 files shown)


tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1172,38 +1172,50 @@ static __always_inline int save_path_to_str_buf(buf_t *string_p, const struct pa
...
     struct dentry *dentry = f_path.dentry;
     struct vfsmount *vfsmnt = f_path.mnt;
+
+    struct dentry *mnt_mountpoint_p;
+    struct mount *mnt_parent_p;
+
     struct mount *mnt_p = real_mount(vfsmnt);
-    struct mount mnt;
-    bpf_probe_read(&mnt, sizeof(struct mount), mnt_p);
+    bpf_probe_read(&mnt_mountpoint_p, sizeof(struct dentry*), &mnt_p->mnt_mountpoint);
```
yanivagman, line 1180: We read into this variable but it never gets used. Do we really need it?

grantseltzer: You're right :-)

yanivagman (approved):
LGTM

PR #759, opened 2021-06-09
Add initial CO:RE support 

grantseltzer:
This changes the way that tracee reads kernel data structures
so that:

1) If a compile time 'CORE' flag is specified, the bpf_core_read
helper is used which takes into account btf information extracted
from the running kernel.

2) KERN_READ calls are not embedded in one another. This was an
issue because the attribute that the BTF helper requires cannot
be embedded in one another.

This also adds supporting artifacts in the vmlinux.h file which
contains baseline definitions of all kernel structures, and a
header file which contains all macros tracee uses that aren't
defined in the vmlinux.h file.

Note: This PR does not address how we handle packaging of a CORE enabled tracee binary, nor does it handle testing infra. 


tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -204,11 +217,20 @@
...
 #define DEV_NULL_STR    0
 
+#ifndef CORE
 #define READ_KERN(ptr) ({ typeof(ptr) _val;                             \
                           __builtin_memset(&_val, 0, sizeof(_val));     \
                           bpf_probe_read(&_val, sizeof(_val), &ptr);    \
                           _val;                                         \
                         })
+#else
+// Try using READ_KERN here, just don't embed them in each other
```
yanivagman, line 227: As we previously discussed, maybe we can change this macro to get the arguments in a comma separated list (which is also supported by the BPF_CORE_READ and BPF_PROBE_READ macros given by libbpf)

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -256,7 +278,7 @@ struct bpf_map_def SEC("maps") _name = { \
 
 /*=============================== INTERNAL STRUCTS ===========================*/
 
-typedef struct context {
+typedef struct internal_proc_context {
```
yanivagman, line 281: Any reason for changing this name?
It is not only the proc context that is in this struct

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1172,38 +1217,50 @@ static __always_inline int save_path_to_str_buf(buf_t *string_p, const struct pa
...
-            if (mnt_p != mnt.mnt_parent) {
+            if (mnt_p != mnt_parent_p) {
                 // We reached root, but not global root - continue with mount point path
-                dentry = mnt.mnt_mountpoint;
-                bpf_probe_read(&mnt, sizeof(struct mount), mnt.mnt_parent);
-                vfsmnt = &mnt.mnt;
+                bpf_probe_read(&dentry, sizeof(struct dentry*), &mnt_p->mnt_mountpoint);
+                bpf_probe_read(&mnt_p, sizeof(struct mount*), &mnt_p->mnt_parent);
+                bpf_probe_read(&mnt_parent_p, sizeof(struct mount*), &mnt_p->mnt_parent);
+                bpf_probe_read(&vfsmnt, sizeof(struct vfsmount*), (const void*)&mnt_p->mnt);
```
yanivagman, line 1251: Weren't these changes already part of your previously commit? rebase?

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1635,14 +1692,19 @@ static __always_inline struct ipv6_pinfo *inet6_sk_own_impl(struct sock *__sk, s
...
 static __always_inline int get_network_details_from_sock_v6(struct sock *sk, net_conn_v6_t *net_details, int peer)
 {
-    struct inet_sock *inet = inet_sk(sk);
+    struct inet_sock *inet = (struct inet_sock *)(sk);
     struct ipv6_pinfo *np = inet6_sk_own_impl(sk, inet);
 
     struct in6_addr addr = {};
     addr = get_sock_v6_rcv_saddr(sk);
-    if (ipv6_addr_any(&addr)){
+    if (ipv6_any_addr(&addr)){
```
yanivagman, line 1707: This function, and inet_sk as well were part of the headers which are now removed for CO-RE, right?
I think that it would be better to put them in the "missing_macros" file (rename?), which should compensate for all the missing things that we had in the headers

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1675,7 +1737,7 @@ int tracepoint__raw_syscalls__sys_enter(struct bpf_raw_tracepoint_args *ctx)
     int id = ctx->args[1];
     struct task_struct *task = (struct task_struct *)bpf_get_current_task();
 
-#if defined(CONFIG_ARCH_HAS_SYSCALL_WRAPPER)
+if (CONFIG_ARCH_HAS_SYSCALL_WRAPPER) {
```
yanivagman, line 1740: What about the case that CORE is not defined? Does it work?
For x86 arch with kernels >= 4.18 this might not be a problem as this macro is always defined, but for ARM64 this is not case (added in a later kernel version)

yanivagman (commented):
Great!
This is something we are very much looking forward to.

Regarding the vmlinux.h file - this file is huge! Do we have to include it as part of the repo? If yes, will we need to ever change it in the future as newer kernels will be introduced?

Also, have you verified that you code works for non-core compilation in kernel 4.18?

tracee-ebpf/Makefile
```diff
@@ -19,7 +19,7 @@ OUT_DIR ?= dist
 GO_SRC := $(shell find . -type f -name '*.go')
 OUT_BIN := $(OUT_DIR)/tracee-ebpf
 BPF_SRC := tracee/tracee.bpf.c 
-OUT_BPF := $(OUT_DIR)/tracee.bpf.$(subst .,_,$(KERN_RELEASE)).$(subst .,_,$(VERSION)).o
+OUT_BPF := $(OUT_DIR)/tracee.bpf.o
```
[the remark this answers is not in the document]
grantseltzer: Sounds good. I will pass a compile time variable into the Go code so that it knows what name to look for.

tracee-ebpf/Makefile
```diff
@@ -111,6 +114,22 @@ $(OUT_DIR)/tracee.bpf.%.o: $(BPF_SRC) $(LIBBPF_HEADERS) | $(OUT_DIR) $(bpf_compi
...
+		-D__TARGET_ARCH_$(linux_arch) \
+		-D__BPF_TRACING__ \
+		-D__VM_LINUX_H__ \
+		-D__KERNEL__ \
+		-DLINUX_VERSION_CODE=$(LINUX_VERSION_CODE) \
+		-target bpf \
+		-O2 -c -g -o $@ $<
+endif
+else
+#docker yes
```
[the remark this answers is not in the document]
grantseltzer: Good point. Going to pass the CORE variable in docker_build_make

tracee-ebpf/tracee/missing_macros.h
```diff
@@ -0,0 +1,134 @@
+/*
+ * The purpose of this file is to define macros
+ * that tracee.bpf.c relies on which are defined
+ * in linux kernel headers but not in vmlinux.h
```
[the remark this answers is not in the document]
grantseltzer: Yes, and good point!

tracee-ebpf/tracee/missing_macros.h
```diff
@@ -0,0 +1,134 @@
...
+#define AF_VSOCK	  40	/* vSockets	*/
+#define AF_KCM		  41	/* Kernel Connection Multiplexor*/
+#define AF_QIPCRTR	  42	/* Qualcomm IPC Router */
+#define AF_SMC		  43	/* smc sockets: reserve number for PF_SMC protocol family that reuses AF_INET address family */
+
+#define VM_NONE		0x00000000
+
+#define VM_READ		0x00000001	/* currently active flags */
+#define VM_WRITE	0x00000002
+#define VM_EXEC		0x00000004
```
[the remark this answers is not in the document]
grantseltzer: Tabs are a scourge

tracee-ebpf/tracee/tracee.bpf.c
grantseltzer: There's a name collision with the `enum context` in vmlinux.h. I'll change it to 'internal_context'.

yanivagman: So maybe "event_context"?

tracee-ebpf/tracee/tracee.bpf.c
grantseltzer: As far as I can tell this causes no issue, but I don't have somewhere to test on ARM64?

tracee-ebpf/tracee/tracee.bpf.c
grantseltzer: Agreed, this requires a little rewiring but I can do it.

grantseltzer:
I'd like to have the vmlinux.h file as a backup in case the build machine doesn't have the vmlinux file. It shouldn't make a difference in terms of functionality. For example, I don't think we have access to it in our builder container.

tracee-ebpf/tracee/tracee.bpf.c
yanivagman: This was added in kernel 4.17 for x86 and 4.20 for ARM64.
As tracee only supports kernels > 4.18, this should be tested on an ARM64 platform with a 4.18/4.19 kernel.
As we don't have any ARM64 platform with these kernels installed to test on, we can keep your changes and address any potential bug that might appear in the future.

tracee-ebpf/tracee/tracee.bpf.c
grantseltzer: I'll take a crack at it for this PR

yanivagman:
LGTM!

One of my last concerns is the missing macros file. What if some macros are changed between kernel versions? Also, what if these macros are not the same on different systems (for example PAGE_SIZE)?

I'm not sure if there are things that we can do to mitigate these problems, but we should be aware of them

grantseltzer:
@yanivagman That's a good point, certainly worth the concern. Another way we can achieve CO:RE is to apply the 'preserve_access_index' attribute to all the headers we import in non-co:re variant of tracee-ebpf.  Just like how it works in vmlinux.h:

`#pragma clang attribute push (__attribute__((preserve_access_index)), apply_to = record)`

That'd let us keep the macros from the headers but has its own sets of challenges.

grantseltzer:
I got a bpf object that was built on a 5.12 kernel running on 4.18 with no issue.

I realized that we definitely should be documenting this as part of this PR. Additional things I want to make sure we make follow up issues for/document:

- Tracee bpf code uses some data structure values that aren't present/renamed in various different kernel versions. This means that the binary likely runs perfectly fine on anything but doesn't build on older kernels. We should make sure we have compile flags in tracee.bpf.c for everything. Requires more investigation but is fine for now.
- The vmlinux file, generating a fresh one for builds, etc...
- Updating macros (see comment above)
- Usage examples

yanivagman:
I think that we should also try to tackle #776 errors in this PR, but not by generating vmlinux.h. Instead, we should probably wrap the existing kernel version ifdefs with an outer "ifndef CORE". WDYT?

yanivagman (approved):

docs/install/ebpf-compilation.md
```diff
@@ -23,3 +23,9 @@ If using Docker, the following `docker run` options demonstrate mounting a pre-c
 ```bash
 docker run ... -v /path/in/host/tracee.bpf.123.o:/path/in/container/tracee.bpf.o -e TRACEE_BPF_FILE=/path/in/container/tracee.bpf.o aquasec/tracee
 ```
+
+# eBPF CO:RE Compilation
+
+Tracee also can utilize CO:RE (Compile once, run everywhere) technology enabled by libbpf. With this enabled, you can compile the tracee bpf object file on one system, and run it on any kernel with BTF (BPF type format) enabled. To check if your kernel has BTF enabled, check for the `CONFIG_DEBUG_INFO_BTF` in your kernel config.
```
itaysk, line 29: > check for the `CONFIG_DEBUG_INFO_BTF` in your kernel config.

can we write the command to do that?
can we make sure Tracee-eBPF's main does that and prints a nice error?

PR #763, opened 2021-06-10
add sched_process_exec event

AsafEitani:
Add sched_process_exec to catch user mode process creations from the kernel. This event will monitor any sched_process_exec - including such that were invoked from the user space.
In order to check if the function was invoked from the kernel - the parent name should start with "kworker/".
This is also backup up with tracee-rules rego helper `is_process_kernel`.

### DIFF
--- a/tracee-ebpf/tracee/consts.go
+++ b/tracee-ebpf/tracee/consts.go
@@ -144,6 +144,7 @@ const (
 	SysEnterEventID int32 = iota + 1000
 	SysExitEventID
 	SchedProcessForkEventID
+	SchedProcessExecEventID
 	SchedProcessExitEventID
 	DoExitEventID
 	CapCapableEventID
@@ -526,6 +527,7 @@ var EventsIDToEvent = map[int32]EventConfig{
 	SysEnterEventID:              {ID: SysEnterEventID, ID32Bit: sys32undefined, Name: "sys_enter", Probes: []probe{{event: "raw_syscalls:sys_enter", attach: rawTracepoint, fn: "tracepoint__raw_syscalls__sys_enter"}}, EssentialEvent: true, Sets: []string{}},
 	SysExitEventID:               {ID: SysExitEventID, ID32Bit: sys32undefined, Name: "sys_exit", Probes: []probe{{event: "raw_syscalls:sys_exit", attach: rawTracepoint, fn: "tracepoint__raw_syscalls__sys_exit"}}, EssentialEvent: true, Sets: []string{}},
 	SchedProcessForkEventID:      {ID: SchedProcessForkEventID, ID32Bit: sys32undefined, Name: "sched_process_fork", Probes: []probe{{event: "sched:sched_process_fork", attach: rawTracepoint, fn: "tracepoint__sched__sched_process_fork"}}, EssentialEvent: true, Sets: []string{}},
+	SchedProcessExecEventID:      {ID: SchedProcessExecEventID, ID32Bit: sys32undefined, Name: "sched_process_exec", Probes: []probe{{event: "sched:sched_process_exec", attach: rawTracepoint, fn: "tracepoint__sched__sched_process_exec"}}, Sets: []string{"proc"}},
 	SchedProcessExitEventID:      {ID: SchedProcessExitEventID, ID32Bit: sys32undefined, Name: "sched_process_exit", Probes: []probe{{event: "sched:sched_process_exit", attach: rawTracepoint, fn: "tracepoint__sched__sched_process_exit"}}, EssentialEvent: true, Sets: []string{"default", "proc", "proc_life"}},
 	DoExitEventID:                {ID: DoExitEventID, ID32Bit: sys32undefined, Name: "do_exit", Probes: []probe{{event: "do_exit", attach: kprobe, fn: "trace_do_exit"}}, Sets: []string{"proc", "proc_life"}},
 	CapCapableEventID:            {ID: CapCapableEventID, ID32Bit: sys32undefined, Name: "cap_capable", Probes: []probe{{event: "cap_capable", attach: kprobe, fn: "trace_cap_capable"}}, Sets: []string{"default"}},
@@ -895,6 +897,7 @@ var EventsIDToParams = map[int32][]external.ArgMeta{
 	SysEnterEventID:              {{Type: "int", Name: "syscall"}},
 	SysExitEventID:               {{Type: "int", Name: "syscall"}},
 	SchedProcessForkEventID:      {{Type: "int", Name: "parent_pid"}, {Type: "int", Name: "parent_ns_pid"}, {Type: "int", Name: "child_pid"}, {Type: "int", Name: "child_ns_pid"}},
+	SchedProcessExecEventID:      {{Type: "const char *", Name: "filename"}, {Type: "int", Name: "invoked_from_kernel"}},
 	SchedProcessExitEventID:      {},
 	DoExitEventID:                {},
 	CapCapableEventID:            {{Type: "int", Name: "cap"}, {Type: "int", Name: "syscall"}},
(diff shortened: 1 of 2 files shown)


tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1999,6 +2000,66 @@ struct bpf_raw_tracepoint_args *ctx
     return 0;
 }
 
+SEC("raw_tracepoint/sched_process_exec")
+int tracepoint__sched__sched_process_exec(
+#if LINUX_VERSION_CODE < KERNEL_VERSION(4, 17, 0)
+void *ctx
+#else
```
yanivagman, line 2007: Please remove these lines as they are not required anymore

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1999,6 +2000,66 @@ struct bpf_raw_tracepoint_args *ctx
...
+        u32 pid_ns = get_task_pid_ns_id(task);
+        if (get_task_ns_pid(task) == 1) {
+            // If pid equals 1 - stop tracing this pid namespace
+            bpf_map_delete_elem(&new_pidns_map, &pid_ns);
+        }
+    }
+    if (get_config(CONFIG_NEW_PID_FILTER)) {
+        // Remove pid from new_pids_map
+        bpf_map_delete_elem(&new_pids_map, &pid);
+    }
```
yanivagman, line 2032: These lines don't belong here - please remove

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1999,6 +2000,66 @@ struct bpf_raw_tracepoint_args *ctx
...
+    struct linux_binprm *bprm = (struct linux_binprm *)ctx->args[2];
+
+    if (bprm == NULL) {
+        return -1;
+    }
+    struct task_struct *parent = (struct task_struct *)READ_KERN(p->parent);
+    if (parent == NULL) {
+            return -1;
+    }
+    const char *filename = (const char *)READ_KERN(bprm->filename);
```
yanivagman, line 2050: Reading values of internal kernel structures may change between different kernel versions.
You should:
1. Move these to the helper functions section
2. Check that the values you read are located and named the same in different kernel versions (starting from 4.18 to the latest version). If not, add a compilation flag according to their location for a specific kernel version

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1999,6 +2000,66 @@ struct bpf_raw_tracepoint_args *ctx
...
+
+    u64 *tags = bpf_map_lookup_elem(&params_names_map, &context.eventid);
+    if (!tags) {
+        return -1;
+    }
+    save_str_to_buf(submit_p, (void *)parent->comm, DEC_ARG(0, *tags));
```
yanivagman, line 2056: This is a generic exec event, which should match (more or less) what we already have in the execve(at) event.
If the only reason for having the parent->comm field is to know if this is a kernel thread, why not checking for this here? For example, you can check if parent_comm has a prefix of "kworker/" (we already have a `has_prefix()` function defined) and only send this information to userspace (instead of sending the whole string)

yanivagman (changes requested):

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2024,6 +2030,46 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
...
 }
 
+// include/trace/events/sched.h:
+//TP_PROTO(struct task_struct *p, pid_t old_pid, struct linux_binprm *bprm)
+SEC("raw_tracepoint/sched_process_exec")
+int tracepoint__sched__sched_process_exec(struct bpf_raw_tracepoint_args *ctx)
+{
+    // Note: removing container id from pid_to_cont_id_map shoule be performed before should_trace()
+    u32 pid = bpf_get_current_pid_tgid();
+    bpf_map_delete_elem(&pid_to_cont_id_map, &pid);
```
yanivagman, line 2040: Why are these here?

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2024,6 +2030,46 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
...
+
+    context_t context = init_and_save_context(ctx, submit_p, SCHED_PROCESS_EXEC, 2, 0);
+    struct task_struct *task = (struct task_struct *)ctx->args[0];
+    struct linux_binprm *bprm = (struct linux_binprm *)ctx->args[2];
+
+    if (bprm == NULL) {
+        return -1;
+    }
+
+    int invoked_from_kernel = has_prefix("kworker/", get_task_parent_comm(task), 9);
```
yanivagman, line 2058: Are you sure that the prefix count here should be 9?
The prefix length is 8. If you take 9, you also consider a NULL byte as part of the prefix, which will probably won't match a thing.

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2024,6 +2030,46 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
...
+    const char *filename = (const char *)READ_KERN(bprm->filename);
```
yanivagman, line 2059: Reading internal kernel structures may change between kernel versions.
Please add a helper function for this as well.

tracee-ebpf/tracee/consts.go
```diff
@@ -895,6 +897,7 @@ var EventsIDToParams = map[int32][]external.ArgMeta{
 	SysEnterEventID:              {{Type: "int", Name: "syscall"}},
 	SysExitEventID:               {{Type: "int", Name: "syscall"}},
 	SchedProcessForkEventID:      {{Type: "int", Name: "parent_pid"}, {Type: "int", Name: "parent_ns_pid"}, {Type: "int", Name: "child_pid"}, {Type: "int", Name: "child_ns_pid"}},
+	SchedProcessExecEventID:      {{Type: "int", Name: "invoked_from_kernel"}, {Type: "const char *", Name: "filename"}},
```
yanivagman, line 900: I would put the filename first, then the invoked_from_kernel parameter

tracee-ebpf/tracee/consts.go
```diff
@@ -526,6 +527,7 @@ var EventsIDToEvent = map[int32]EventConfig{
 	SysEnterEventID:              {ID: SysEnterEventID, ID32Bit: sys32undefined, Name: "sys_enter", Probes: []probe{{event: "raw_syscalls:sys_enter", attach: rawTracepoint, fn: "tracepoint__raw_syscalls__sys_enter"}}, EssentialEvent: true, Sets: []string{}},
 	SysExitEventID:               {ID: SysExitEventID, ID32Bit: sys32undefined, Name: "sys_exit", Probes: []probe{{event: "raw_syscalls:sys_exit", attach: rawTracepoint, fn: "tracepoint__raw_syscalls__sys_exit"}}, EssentialEvent: true, Sets: []string{}},
 	SchedProcessForkEventID:      {ID: SchedProcessForkEventID, ID32Bit: sys32undefined, Name: "sched_process_fork", Probes: []probe{{event: "sched:sched_process_fork", attach: rawTracepoint, fn: "tracepoint__sched__sched_process_fork"}}, EssentialEvent: true, Sets: []string{}},
+	SchedProcessExecEventID:      {ID: SchedProcessExecEventID, ID32Bit: sys32undefined, Name: "sched_process_exec", Probes: []probe{{event: "sched:sched_process_exec", attach: rawTracepoint, fn: "tracepoint__sched__sched_process_exec"}}, Sets: []string{"default", "proc"}},
```
yanivagman, line 530: For now, please remove this event from the default set

yanivagman (changes requested):

tracee-ebpf/tracee/tracee.bpf.c
AsafEitani: I think that the null byte is needed... I followed the flow of the has_prefix code and the only way it will return 1 is if it reaches a null byte on the prefix. Also it works for me.

yanivagman (approved):

PR #775, opened 2021-06-17
Category output

AsafEitani:
add category for each event - allowing for better understanding tracee-ebpf output and generalizing tracee-rules to use the category instead of multiple system calls (eg. using category `execution` instead of events `execve`, `security_bprm_check`, `execveat`, etc.)


yanivagman (changes requested):
Other than the inline comments, I want to say that I am still not fully convinced that this feature is necessary.
Can you please explain here the motivation behind it, and why do you think that it is necessary? Maybe then others will also share their opinion about it.

AsafEitani:
@yanivagman basically it provides some information regarding the event for users that are not familiar with linux internals and provides the option to stream the tracee-ebpf data directly to splunk or elasticsearch (and make is searchable), thus making tracee a very appealing option as a monitoring technology, like Sysmon is for Windows.

tracee-ebpf/tracee/printer.go
```diff
@@ -132,9 +133,9 @@ func (p tableEventPrinter) Preamble() {
 func (p tableEventPrinter) Print(event external.Event) {
 	if p.verbose {
 		if p.containerMode {
-			fmt.Fprintf(p.out, "%-14f %-16s %-16s %-12d %-12d %-6d %-16s %-7d/%-7d %-7d/%-7d %-7d/%-7d %-16d %-20s ", event.Timestamp, event.HostName, event.ContainerID, event.MountNS, event.PIDNS, event.UserID, event.ProcessName, event.ProcessID, event.HostProcessID, event.ThreadID, event.HostThreadID, event.ParentProcessID, event.ParentProcessID, event.ReturnValue, event.EventName)
+			fmt.Fprintf(p.out, "%-14f %-16s %-16s %-12d %-12d %-6d %-16s %-7d/%-7d %-7d/%-7d %-7d/%-7d %-16d %-12s %-20s", event.Timestamp, event.HostName, event.ContainerID, event.MountNS, event.PIDNS, event.UserID, event.ProcessName, event.ProcessID, event.HostProcessID, event.ThreadID, event.HostThreadID, event.ParentProcessID, event.ParentProcessID, event.ReturnValue, event.Category, event.EventName)
```
yanivagman, line 136: missing trailing space after %20s

grantseltzer:
@yanivagman Do you still have concerns about this feature? Lets avoid having stale PRs open if so

yanivagman:
> @yanivagman Do you still have concerns about this feature? Lets avoid having stale PRs open if so

Yes. I'm not sure that this is the design we want for adding a "category", and this is why I wanted 

> others will also share their opinion about it.

My concerns are:
1. There is no new data added here to the event as this new field only maps an event to a category. I think that tracee-ebpf should only generate events, while any enrichment that doesn't add any new data should be done after
2. What if we want to have other partition into (different) categories? This patch sets "network access"/"file access"/etc category names, but maybe there are other useful partitions (for example, more fine grained categories, like we have for the event sets)

So, I think that maybe we should wait with this one, and add it as a "plugin" in the future (when we will have support for such plugins). WDYT?

grantseltzer:
I'm going to go ahead and close this. @AsafEitani please feel free to add your thoughts and describe the problem you're trying to solve in a new issue.

Issue #777, opened 2021-06-18
tracee-rules panic observed


simar7:
One fine day the sun shine was shining bright and there were no clouds in the sky. But alas a panic was observed in tracee-rules...

```
sudo ./dist/tracee-ebpf -o format:gob | ./dist/tracee-rules --rules-dir=<path/to/rules-dir> --input-tracee file:stdin --input-tracee format:gob

<snip>

^Cpanic: send on closed channel

goroutine 137 [running]:
github.com/aquasecurity/tracee/tracee-rules/engine.(*Engine).consumeSources(0xc0008b3030, 0xc0001323f0)
	/usr/vagrant/repos/tracee/tracee-rules/engine/engine.go:114 +0x2b9
created by github.com/aquasecurity/tracee/tracee-rules/engine.(*Engine).Start
	/usr/vagrant/repos/tracee/tracee-rules/engine/engine.go:70 +0x65
```

This was just a normal invocation of tracee-ebpf and tracee-rules from the master branch. Nothing out of the ordinary to set it up.

simar7:
Another instance without `rules-dir`

```
sudo ./dist/tracee-ebpf -o format:gob | ./dist/tracee-rules  --input-tracee file:stdin --input-tracee format:gob
Loaded signature(s):  [TRC-1 TRC-2 TRC-3 TRC-4 TRC-5 TRC-6 TRC-7]
^Cpanic: send on closed channel

goroutine 44 [running]:
github.com/aquasecurity/tracee/tracee-rules/engine.(*Engine).consumeSources(0xc000174b60, 0xc00006a000)
	/usr/vagrant/repos/tracee/tracee-rules/engine/engine.go:114 +0x2b9
created by github.com/aquasecurity/tracee/tracee-rules/engine.(*Engine).Start
	/usr/vagrant/repos/tracee/tracee-rules/engine/engine.go:70 +0x65
```

mtcherni95:
@simar7 I am having some troubles in reproducing it. Any suggestion or specific action you did in order to get the error?

simar7:
> @simar7 I am having some troubles in reproducing it. Any suggestion or specific action you did in order to get the error?

There's a reproducible test case and discussion here https://github.com/aquasecurity/tracee/pull/785#discussion_r657541451

PR #785, opened 2021-06-23
test: Benchmark tracee-rules

danielpacak:

### DIFF
--- a/tracee-rules/benchmark/benchmark_test.go
+++ b/tracee-rules/benchmark/benchmark_test.go
@@ -1,0 +1,143 @@
+package benchmark
+
+import (
+	"os"
+	"testing"
+	"time"
+
+	"github.com/aquasecurity/tracee/tracee-rules/benchmark/signature/golang"
+	"github.com/aquasecurity/tracee/tracee-rules/benchmark/signature/rego"
+	"github.com/aquasecurity/tracee/tracee-rules/engine"
+	"github.com/aquasecurity/tracee/tracee-rules/types"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	inputEventsCount = 1000
+)
+
+func BenchmarkOnEventCodeInjectionRegoRule(b *testing.B) {
+	codeInjectSig, err := rego.NewCodeInjectionSignature()
+	require.NoError(b, err)
+
+	err = codeInjectSig.Init(ignoreFinding)
+	require.NoError(b, err)
+
+	for i := 0; i < b.N; i++ {
+		err = codeInjectSig.OnEvent(triggerCodeInjectorPtraceEvent)
+		require.NoError(b, err)
+	}
+}
+
+func BenchmarkOnEventCodeInjectionGoRule(b *testing.B) {
+	codeInjectSig := golang.NewCodeInjectionSignature()
+	err := codeInjectSig.Init(ignoreFinding)
+	require.NoError(b, err)
+
+	for i := 0; i < b.N; i++ {
+		err := codeInjectSig.OnEvent(triggerCodeInjectorPtraceEvent)
+		require.NoError(b, err)
+	}
+}
+
+func BenchmarkEngineWithCodeInjectionRegoRule(b *testing.B) {
+	// Prepare signatures
+	codeInjectSig, err := rego.NewCodeInjectionSignature()
+	require.NoError(b, err)
+
+	sigs := []types.Signature{
+		codeInjectSig,
+	}
+
+	for i := 0; i < b.N; i++ {
+		// Produce events
+		eventsCh := make(chan types.Event, inputEventsCount)
+		ProduceEventsInMemory(eventsCh, inputEventsCount)
+
+		inputs := engine.EventSources{
+			Tracee: eventsCh,
+		}
+		output := make(chan types.Finding, inputEventsCount)
+
+		e := engine.NewEngine(sigs, inputs, output, os.Stderr)
+		// Start rules engine and wait until all events are processed
+		e.Start(waitForEventsProcessed(eventsCh))
+
+		b.Logf("Test is done with %d findings", len(output))
+	}
+}
+
+func BenchmarkEngineWithCodeInjectionGoRule(b *testing.B) {
+	// Prepare signatures
+	codeInjectSig := golang.NewCodeInjectionSignature()
+
+	sigs := []types.Signature{
+		codeInjectSig,
+	}
+
+	for i := 0; i < b.N; i++ {
+		// Produce events
+		eventsCh := make(chan types.Event, inputEventsCount)
+		ProduceEventsInMemory(eventsCh, inputEventsCount)
+
+		inputs := engine.EventSources{
+			Tracee: eventsCh,
+		}
+		output := make(chan types.Finding, inputEventsCount)
+
+		e := engine.NewEngine(sigs, inputs, output, os.Stderr)
+		// Start rules engine and wait until all events are processed
+		e.Start(waitForEventsProcessed(eventsCh))
+
+		b.Logf("Test is done with %d findings", len(output))
+	}
+}
+
+func BenchmarkEngineWithMultipleRegoAndGoRules(b *testing.B) {
+	// Prepare signatures
+	codeInjectionRegoSig, _ := rego.NewCodeInjectionSignature()
+	antiDebuggingRegoSig, _ := rego.NewAntiDebuggingSignature()
+
+	sigs := []types.Signature{
+		codeInjectionRegoSig,
+		antiDebuggingRegoSig,
+
+		golang.NewCodeInjectionSignature(),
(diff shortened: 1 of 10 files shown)


simar7:
I think we need to add the new benchmark dir as part of the ignored dirs for opa test: https://github.com/aquasecurity/tracee/blob/main/tracee-rules/Makefile#L56

That would make the currently failing check pass.

tracee-rules/benchmark/benchmark_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+func waitForEventsProcessed(eventsCh chan types.Event) chan bool {
+	done := make(chan bool, 1)
+	go func() {
+		for {
+			if len(eventsCh) == 0 {
+				done <- true
+				return
+			}
+			time.Sleep(100 * time.Millisecond)
```
simar7, line 135: Reducing this wait down exposes a bug in the code: https://github.com/aquasecurity/tracee/issues/777

```suggestion
			time.Sleep(1 * time.Millisecond)
```

If patched the numbers look a lot different:

```

benchmark                                     iter           time/iter     bytes alloc             allocs
---------                                     ----           ---------     -----------             ------
BenchmarkOnEventCodeInjectionRegoRule-8        100      56327.00 ns/op      23318 B/op      476 allocs/op
BenchmarkOnEventCodeInjectionGoRule-8          100        813.80 ns/op        608 B/op        4 allocs/op
BenchmarkEngineWithCodeInjectionRegoRule-8     100   12355021.00 ns/op    5762957 B/op   112738 allocs/op
BenchmarkEngineWithCodeInjectionGoRule-8       100    1449184.00 ns/op     550480 B/op     2550 allocs/op
BenchmarkEngineWithMultipleRegoAndGoRules-8    100   15819138.00 ns/op   11116686 B/op   214733 allocs/op
ok      github.com/aquasecurity/tracee/tracee-rules/benchmark   3.179s
```

simar7: IOW waiting here makes the numbers artificially skew larger than they should be. 

Somewhat related to this:
The way to get around the bug is to see the above numbers is to not close the channel here https://github.com/aquasecurity/tracee/blob/ef7e96ace8592fe6eb2333008391f0bfb9b6ce8e/tracee-rules/engine/engine.go#L72 – Ideally this channel close should happen from the creator (test program) and not within the engine itself.

tracee-rules/benchmark/benchmark_test.go
```diff
@@ -0,0 +1,143 @@
...
+	inputEventsCount = 1000
+)
+
+func BenchmarkOnEventCodeInjectionRegoRule(b *testing.B) {
+	codeInjectSig, err := rego.NewCodeInjectionSignature()
+	require.NoError(b, err)
+
+	err = codeInjectSig.Init(ignoreFinding)
+	require.NoError(b, err)
+
```
simar7, line 25: We need a `b.ResetTimer()` call here. 


image


Same for all other benchmarks.

tracee-rules/benchmark/benchmark_test.go
simar7: Actually I take this back. The numbers were different for another reason. If you add `b.ResetTimer()` to all the appropriate places after the setup is done the numbers are much better:

```
PASS
benchmark                                     iter            time/iter     bytes alloc             allocs
---------                                     ----            ---------     -----------             ------
BenchmarkOnEventCodeInjectionRegoRule-8        100       45904.00 ns/op      17971 B/op      353 allocs/op
BenchmarkOnEventCodeInjectionRegoRule-8        100       52121.00 ns/op      17703 B/op      352 allocs/op
BenchmarkOnEventCodeInjectionRegoRule-8        100       40360.00 ns/op      17645 B/op      352 allocs/op
BenchmarkOnEventCodeInjectionRegoRule-8        100       41343.00 ns/op      17628 B/op      352 allocs/op
BenchmarkOnEventCodeInjectionRegoRule-8        100       33396.00 ns/op      17665 B/op      352 allocs/op
BenchmarkOnEventCodeInjectionGoRule-8          100         554.00 ns/op        560 B/op        4 allocs/op
BenchmarkOnEventCodeInjectionGoRule-8          100         567.60 ns/op        560 B/op        4 allocs/op
BenchmarkOnEventCodeInjectionGoRule-8          100         564.50 ns/op        560 B/op        4 allocs/op
BenchmarkOnEventCodeInjectionGoRule-8          100         568.20 ns/op        560 B/op        4 allocs/op
BenchmarkOnEventCodeInjectionGoRule-8          100         560.30 ns/op        560 B/op        4 allocs/op
BenchmarkOnEventCodeInjectionWASMRule-8        100       28182.00 ns/op       1751 B/op       26 allocs/op
BenchmarkOnEventCodeInjectionWASMRule-8        100       21746.00 ns/op       1751 B/op       26 allocs/op
BenchmarkOnEventCodeInjectionWASMRule-8        100       21070.00 ns/op       1751 B/op       26 allocs/op
BenchmarkOnEventCodeInjectionWASMRule-8        100       20489.00 ns/op       1751 B/op       26 allocs/op
BenchmarkOnEventCodeInjectionWASMRule-8        100       20153.00 ns/op       1751 B/op       26 allocs/op
```

WASM is nearly half! 🚀

simar7 (changes requested):
Just one comment to be addressed for adding the reset of timer after initial setup is done.

tracee-rules/benchmark/benchmark_test.go
```diff
@@ -0,0 +1,143 @@
...
+	sigs := []types.Signature{
+		codeInjectSig,
+	}
+
+	for i := 0; i < b.N; i++ {
+		// Produce events
+		eventsCh := make(chan types.Event, inputEventsCount)
+		ProduceEventsInMemory(eventsCh, inputEventsCount)
```
itaysk, line 55: maybe move this out of the bench loop? IMO we can produce the events once

tracee-rules/benchmark/benchmark_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+		inputs := engine.EventSources{
+			Tracee: eventsCh,
+		}
+		output := make(chan types.Finding, inputEventsCount)
+
+		e := engine.NewEngine(sigs, inputs, output, os.Stderr)
```
itaysk, line 62: I think we shouldn't include the engine initialization in the benchmark

tracee-rules/benchmark/helpers_test.go
```diff
@@ -0,0 +1,113 @@
...
+	"math/rand"
+	"os"
+	"testing"
+
+	tracee "github.com/aquasecurity/tracee/tracee-ebpf/tracee/external"
+	"github.com/aquasecurity/tracee/tracee-rules/types"
+)
+
+var (
+	innocentEvent = tracee.Event{
```
itaysk, line 17: would be better to use a full event, with data in the fields, so that the load is more realistic (same goes for other events)

tracee-rules/benchmark/benchmark_test.go
itaysk: Great find!
just to correct a small comment you made

> Ideally this channel close should happen from the creator (test program) and not within the engine itself

this close is in the right place, engine IS the creator:
https://github.com/aquasecurity/tracee/blob/ef7e96ace8592fe6eb2333008391f0bfb9b6ce8e/tracee-rules/engine/engine.go#L35

tracee-rules/benchmark/benchmark_test.go
danielpacak: How would you Stop then engine (and benchmark iteration) in this approach? We generate 1000 events in each benchmark loop and close done channel when the input events channel is empty.

tracee-rules/benchmark/benchmark_test.go
danielpacak: One way to do that, which I tried out in this PR is to use b.StopTimer and b.StartTime to skip timing events producer:

```go
func BenchmarkEngineWithCodeInjectionRegoRule(b *testing.B) {
	// Prepare signatures and what not

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		b.StopTimer()
		eventsCh := make(chan types.Event, inputEventsCount)
		// Produce events without timing it
		b.StartTimer()

		// Start rules engine and wait until all events are processed
		e.Start(waitForEventsProcessed(eventsCh))

		b.Logf("Test is done with %d findings", len(output))
	}
}
```

tracee-rules/benchmark/helpers_test.go
danielpacak: Added values for fields based on real events captured by tracee-ebpf

simar7 (approved):

simar7:
lgtm, merging to rebase on top with my work

PR #787, opened 2021-06-24
tracee-ebpf: fix compilation on ubuntu

### LINKED ISSUE #786: Cannot compile Tracee-eBPF on Ubuntu 20
Here a snapshot of compilation error:
image

yanivagman:
This PR fixes a regression caused by the recent merge of the initial CO-RE
support to match the fix in 85cf5b

Fixes #786

### DIFF
--- a/tracee-ebpf/tracee/tracee.bpf.c
+++ b/tracee-ebpf/tracee/tracee.bpf.c
@@ -8,18 +8,17 @@ SPDX-License-Identifier: GPL-2.0
 Copyright (C) Aqua Security inc.
 */
 
+#ifndef CORE
+/* In Linux 5.4 asm_inline was introduced, but it's not supported by clang.
+ * Redefine it to just asm to enable successful compilation.
+ * see https://github.com/iovisor/bcc/commit/2d1497cde1cc9835f759a707b42dea83bee378b8 for more details
+ * Note: types.h should be included before defining asm_inline or compilation might break
+ */
+#include <linux/types.h>
 #ifdef asm_inline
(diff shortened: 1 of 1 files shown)


tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -8,18 +8,20 @@ SPDX-License-Identifier: GPL-2.0
...
  * Redefine it to just asm to enable successful compilation.
  * see https://github.com/iovisor/bcc/commit/2d1497cde1cc9835f759a707b42dea83bee378b8 for more details
+ * Note: types.h should be included before defining asm_inline or compilation might break
  */
 #include <linux/types.h>
+#endif
+#ifdef asm_inline
+#undef asm_inline
+#define asm_inline asm
+#endif
```
grantseltzer, line 22: Do we need these lines if compiling for CO:RE? I removed them and compiled with CO:RE and it doesn't seem to cause any issues.

grantseltzer (approved):

tracee-ebpf/tracee/tracee.bpf.c
yanivagman: Not sure that we do.
Better to check that on an environment where this problem occurs (such as @mtcherni95 's) we can compile CORE succesfully when removing them.

yanivagman: Ok, we tested on @mtcherni95 environment, and this is not required.
I updated the PR accordingly

Issue #473, opened 2021-01-27
Use cgroup id for containers


yanivagman:
Starting from kernel 4.18 we can check for a container by cgroup id (using bpf_get_current_cgroup_id() helper).
This is a better way than using pidns as we do today

PR #788, opened 2021-06-24
tracee-ebpf: fix container id issues

yanivagman:
1. Fix container id when cgroup starts with "docker-"
2. Fix container id not shown in sched_process_exit event


grantseltzer (approved):
Is there an issue to link this to? #473?

yanivagman:
No, I didn't open an issue for this

PR #799, opened 2021-06-30
fix: Unsynchronized send and close operations on signature channels

danielpacak:
Resolve: #777

### DIFF
--- a/tracee-rules/engine/engine.go
+++ b/tracee-rules/engine/engine.go
@@ -67,7 +67,6 @@ func NewEngine(sigs []types.Signature, sources EventSources, output chan types.F
 // once done, it cleans all internal resources, which means the engine is not reusable
 // note that the input and output channels are created by the consumer and therefore are not closed
 func (engine *Engine) Start(done chan bool) {
-	go engine.consumeSources(done)
 	for s, c := range engine.signatures {
 		defer close(c)
 		go func(s types.Signature, c chan types.Event) {
@@ -80,7 +79,7 @@ func (engine *Engine) Start(done chan bool) {
 			}
 		}(s, c)
 	}
-	<-done
+	engine.consumeSources(done)
 }
 
 // matchHandler is a function that runs when a signature is matched


tracee-rules/engine/engine.go
```diff
@@ -69,7 +69,6 @@ func NewEngine(sigs []types.Signature, sources EventSources, output chan types.F
 func (engine *Engine) Start(done chan bool) {
 	go engine.consumeSources(done)
 	for s, c := range engine.signatures {
-		defer close(c)
```
danielpacak, line 72: Here, defer close(c) inside the loop don’t get executed until the Start func ends - not when each step of for loop ends.

itaysk: what you say is correct, but it's not a bug though, is it?

tracee-rules/engine/engine.go
```diff
@@ -107,6 +106,7 @@ func (engine *Engine) consumeSources(done <-chan bool) {
 							break
 						}
 					}
+					close(sigCh)
```
itaysk, line 109: what if there are multiple input channels?

itaysk (commented):
Quickly reviewing this and I'm not sure this does the right thing, but I might be wrong, could you please elaborate

itaysk:
to summarize my two comments and my understanding of this: currently the sig's channel gets closed at the end of the engine.Start(), which makes sense to me. The PR propose to close it when the Tracee input channel is gets closed

tracee-rules/engine/engine.go
simar7: Is that possible today? 
https://github.com/aquasecurity/tracee/blob/5cc1e8cc27d03b60c7178995a3e448337815630a/tracee-rules/input.go#L33-L44 

Returns at most one input channel back that gets assigned to be used.

In the case of having the ability to support multiple input channels, I think we need to rethink the design of consumeSources as a whole. Maybe have a goroutine for each input source processing events. But until then I believe this should suffice.

simar7 (approved):

simar7:
> to summarize my two comments and my understanding of this: currently the sig's channel gets closed at the end of the engine.Start() (i.e when the engine stops), which makes sense to me. The PR propose to close it when the Tracee input channel is gets closed, but engine's lifecycle shouldn't be tied to the input channel's lifecycle

Once the input is closed, it is never expected to be opened again right? I think this is just a common signalling problem. Someone has to signal when the end happens. In this PR I believe that just ties it with the producer. No more stuff on input channel, no need to keep engine running. That's my understanding of it.

tracee-rules/engine/engine.go
yanivagman: If I understand correctly, the panic is caused due to a race condition when a SIGINT or SIGTERM are recieved.
In #777, ^C was hit, which sends SIGINT.
When this happens, the done channel receives 'true' (see sigHandler() in main.go), and at this point two things happen in parallel:
1. The `Start()` function ends, which triggers all the defer close(c)
2. `consumeSources() `also receives 'done'.
The problem is that `consumeSources()` might be in the middle of handling other event, and try to send the event  through one of the sigs channels.

So, my suggestion is to close the signature channels in the `case <-done:` below.
This way, there is no problem if there will be multiple input channels.

yanivagman (changes requested):

tracee-rules/engine/engine.go
danielpacak: Hey @yanivagman I think your comment is spot on. I actually bumped into similar error while running benchmarks, not SIGINT or SIGTERM interruption.

However, I do agree that the problem is not with deferred syntax to close the channels created for each signature, but with two gorutines competing over these channels:
1. One that invoked the Start method and deferred close methods. (I assume Start and deferred close methods run in the same thread.)
2. consumeSources goroutine that may send event to channel already closed by 1 in parallel.

tracee-rules/engine/engine.go
danielpacak: Well, yes and no 🤓 Start is run in a different thread than consumeSources. I temporarily printed goroutine id and reproduced this error with by intercepting it with ctrl+C

```console
$ cat tracee_output.gob| go run . --rules-dir dist/rules --input-tracee=file:stdin --input-tracee=format:gob
consumeSources goroutine 82
^CStart() is about to return
closing sig channel 0xc00010e540 in goroutine 1
closing sig channel 0xc00010e540 in goroutine 1
closing sig channel 0xc00010e540 in goroutine 1
closing sig channel 0xc00010e540 in goroutine 1
closing sig channel 0xc00010e540 in goroutine 1
closing sig channel 0xc00010e540 in goroutine 1
closing sig channel 0xc00010e540 in goroutine 1
panic: send on closed channel

goroutine 82 [running]:
github.com/aquasecurity/tracee/tracee-rules/engine.(*Engine).consumeSources(0xc0004ae620, 0xc0001a6000)
	/Users/dpacak/go/src/github.com/aquasecurity/tracee/tracee-rules/engine/engine.go:138 +0x33c
created by github.com/aquasecurity/tracee/tracee-rules/engine.(*Engine).Start
	/Users/dpacak/go/src/github.com/aquasecurity/tracee/tracee-rules/engine/engine.go:89 +0x65
exit status 2
```

From this listing you can see that we're closing channels in goroutine `1` but goroutine `82` is still trying to send an event on the closed channel.

danielpacak: It's even more interesting to run this program with the -race flag and reproduce the panic

```console
$ cat tracee_output.gob| go run -race . --rules-dir dist/rules --input-tracee=file:stdin --input-tracee=format:gob
2021/07/02 15:17:54 error opening plugin builtin.so: plugin.Open("dist/rules/builtin"): plugin was built with a different version of package runtime
Loaded signature(s):  [TRC-2 TRC-3 TRC-4 TRC-5 TRC-6 TRC-7]
consumeSources goroutine 53

*** Detection ***
Time: 2021-07-02T13:18:03Z
Signature ID: TRC-2
Signature: Anti-Debugging
Data: map[]
Command: strace
Hostname: tracee
^CStart() is about to return
closing sig channel 0xc000110660 in goroutine 1
==================
WARNING: DATA RACE
Write at 0x00c000110670 by main goroutine:
  runtime.closechan()
      /usr/local/Cellar/go/1.16.5/libexec/src/runtime/chan.go:355 +0x0
  github.com/aquasecurity/tracee/tracee-rules/engine.(*Engine).Start.func1()
      /Users/dpacak/go/src/github.com/aquasecurity/tracee/tracee-rules/engine/engine.go:93 +0xe4
  github.com/aquasecurity/tracee/tracee-rules/engine.(*Engine).Start()
      /Users/dpacak/go/src/github.com/aquasecurity/tracee/tracee-rules/engine/engine.go:107 +0x23f
  main.main.func1()
      /Users/dpacak/go/src/github.com/aquasecurity/tracee/tracee-rules/main.go:87 +0x964
  github.com/urfave/cli/v2.(*App).RunContext()
      /Users/dpacak/go/pkg/mod/github.com/urfave/cli/v2@v2.3.0/app.go:322 +0xabe
  github.com/urfave/cli/v2.(*App).Run()
      /Users/dpacak/go/pkg/mod/github.com/urfave/cli/v2@v2.3.0/app.go:224 +0xb1c
  main.main()
      /Users/dpacak/go/src/github.com/aquasecurity/tracee/tracee-rules/main.go:125 +0xa95

Previous read at 0x00c000110670 by goroutine 45:
  runtime.chansend()
      /usr/local/Cellar/go/1.16.5/libexec/src/runtime/chan.go:158 +0x0
  github.com/aquasecurity/tracee/tracee-rules/engine.(*Engine).consumeSources()
      /Users/dpacak/go/src/github.com/aquasecurity/tracee/tracee-rules/engine/engine.go:138 +0x45a

Goroutine 45 (running) created at:
  github.com/aquasecurity/tracee/tracee-rules/engine.(*Engine).Start()
      /Users/dpacak/go/src/github.com/aquasecurity/tracee/tracee-rules/engine/engine.go:89 +0x76
  main.main.func1()
      /Users/dpacak/go/src/github.com/aquasecurity/tracee/tracee-rules/main.go:87 +0x964
  github.com/urfave/cli/v2.(*App).RunContext()
      /Users/dpacak/go/pkg/mod/github.com/urfave/cli/v2@v2.3.0/app.go:322 +0xabe
  github.com/urfave/cli/v2.(*App).Run()
      /Users/dpacak/go/pkg/mod/github.com/urfave/cli/v2@v2.3.0/app.go:224 +0xb1c
  main.main()
      /Users/dpacak/go/src/github.com/aquasecurity/tracee/tracee-rules/main.go:125 +0xa95
==================
closing sig channel 0xc000110600 in goroutine 1
closing sig channel 0xc0001105a0 in goroutine 1
closing sig channel 0xc000110540 in goroutine 1
closing sig channel 0xc0001104e0 in goroutine 1
closing sig channel 0xc000110480 in goroutine 1
Found 1 data race(s)
exit status 66
```

tracee-rules/engine/engine.go
```diff
@@ -80,7 +83,7 @@ func (engine *Engine) Start(done chan bool) {
 			}
 		}(s, c)
 	}
-	<-done
+	engine.consumeSources(done)
```
danielpacak, line 86: If we want to use deferred close(c) the consumeSources must run in the same thread as Start()

If for some reason consumeSources must run in its own goroutine we should not use deferred close(c) but consumeSources must do that in the loop when it receives data from the done channel as suggested by @yanivagman 

That's at least my understanding of what is going on here. I tested and we no longer observer panic on ctrl + C and when running benchmarks. I cannot see DATA RACES anymore

tracee-rules/engine/engine.go
danielpacak: https://golang.org/doc/articles/race_detector#Unsynchronized_send_and_close_operations

yanivagman (approved):
LGTM!

itaysk (approved):
thanks for the analysis, this solution sounds good to me, but PR https://github.com/aquasecurity/tracee/pull/801 might change this altogether so let's hold on merging this for now.

Issue #797, opened 2021-06-30
Tracee-Rules dynamic signature loading


itaysk:
we want to support loading/unloading signatures at runtime (today sigs are provided one time at engine-initialization time):

1. `engine.Load(types.Signature) error`
2. `engine.Unload(types.Signature) error`
3. on signature interface, add `signature.Close()` that should clean after `signature.Init()`

Now that we allow loading/unloading signatures at runtime, the internal maps need to be thread safe.

mtcherni95:
We should consider to change as well the following logics by waiting to run as until there are loaded signatures, and if there are not wait. We could do by simply using a condition variable. 
![image](https://user-images.githubusercontent.com/39140425/124021009-b0b71180-d9f3-11eb-87f0-86364b476bf4.png)

itaysk:
I don't see why the engine should care about which signatures are loaded, since signatures can come and go. What if all the signatures were unloaded should the engine stop? I don't think so. I think it would just keep consuming the input sources and dispatch to no signatures. WDYT?

itaysk:
BTW there is some logic from start/stop the will need to be in load/unload of course. Since it's not just adding and removing from the maps

mtcherni95:
exactly, that's what I am saying. We don't want to stop if there are no signatures. In the current code, if there are no signatures, then the loop at line 71 will exit and we will push 'done' to channel. We don't want such a scenario.

mtcherni95:
In other words, we need to add a synchronisation mechanism (condition variable suits best in this case), where the predicat of the condition variable is that there is at least one loaded signatures.

itaysk:
> We don't want to stop if there are no signatures.

the current code won't stop if there are not signatures, it will block on <-done, which is good IMO see next

> In the current code, if there are no signatures, then the loop at line 71 will exit and we will read from 'done' channel. We don't want such a scenario.

Isn't this what we want? at this point the engine is running and consuming sources, and you can still load signatures to it

itaysk:
reopening this issue since the closing PR didn't cover Go signatures

PR #801, opened 2021-07-01
dynamic load/unload signature capability (#797)

mtcherni95:
#797

### DIFF
--- a/tracee-rules/engine/engine.go
+++ b/tracee-rules/engine/engine.go
@@ -2,8 +2,10 @@ package engine
 
 import (
 	"errors"
+	"fmt"
 	"io"
 	"log"
+	"sync"
 
 	tracee "github.com/aquasecurity/tracee/tracee-ebpf/tracee/external"
 	"github.com/aquasecurity/tracee/tracee-rules/types"
@@ -14,6 +16,7 @@ type Engine struct {
 	logger          log.Logger
 	signatures      map[types.Signature]chan types.Event
 	signaturesIndex map[types.SignatureEventSelector][]types.Signature
+	signaturesMutex sync.RWMutex
 	inputs          EventSources
 	output          chan types.Finding
 }
@@ -26,14 +29,21 @@ type EventSources struct {
 // NewEngine creates a new rules-engine with the given arguments
 // inputs and outputs are given as channels created by the consumer
 func NewEngine(sigs []types.Signature, sources EventSources, output chan types.Finding, logWriter io.Writer) (*Engine, error) {
+	if sources.Tracee == nil || output == nil || logWriter == nil {
+		return nil, fmt.Errorf("nil input received")
+	}
 	engine := Engine{}
 	engine.logger = *log.New(logWriter, "", 0)
 	engine.inputs = sources
 	engine.output = output
+	engine.signaturesMutex.Lock()
 	engine.signatures = make(map[types.Signature]chan types.Event)
 	engine.signaturesIndex = make(map[types.SignatureEventSelector][]types.Signature)
+	engine.signaturesMutex.Unlock()
 	for _, sig := range sigs {
+		engine.signaturesMutex.Lock()
 		engine.signatures[sig] = make(chan types.Event)
+		engine.signaturesMutex.Unlock()
 		meta, err := sig.GetMetadata()
 		if err != nil {
 			engine.logger.Printf("error getting metadata: %v", err)
@@ -51,7 +61,9 @@ func NewEngine(sigs []types.Signature, sources EventSources, output chan types.F
 			if es.Source == "" {
 				engine.logger.Printf("signature %s doesn't declare an input source", meta.Name)
 			} else {
+				engine.signaturesMutex.Lock()
 				engine.signaturesIndex[es] = append(engine.signaturesIndex[es], sig)
+				engine.signaturesMutex.Unlock()
 			}
 		}
 		err = sig.Init(engine.matchHandler)
@@ -60,10 +72,23 @@ func NewEngine(sigs []types.Signature, sources EventSources, output chan types.F
 			continue
 		}
 	}
-	if len(sigs) != len(engine.signatures) {
+	engine.signaturesMutex.RLock()
+	lenSigs := len(engine.signatures)
+	engine.signaturesMutex.RUnlock()
+	if len(sigs) != lenSigs {
 		return nil, errors.New("one or more signatures are not uniquely identifiable")
 	}
 	return &engine, nil
+}
+
+// signatureStart is the signature handling business logics.
+func signatureStart(signature types.Signature, c chan types.Event) {
+	for e := range c {
+		if err := signature.OnEvent(e); err != nil {
+			meta, _ := signature.GetMetadata()
+			log.Printf("error handling event by signature %s: %v", meta.Name, err)
+		}
+	}
 }
 
 // Start starts processing events and detecting signatures
@@ -71,19 +96,24 @@ func NewEngine(sigs []types.Signature, sources EventSources, output chan types.F
 // once done, it cleans all internal resources, which means the engine is not reusable
 // note that the input and output channels are created by the consumer and therefore are not closed
 func (engine *Engine) Start(done chan bool) {
+	defer engine.unloadAllSignatures()
+	engine.signaturesMutex.RLock()
 	for s, c := range engine.signatures {
-		defer close(c)
-		go func(s types.Signature, c chan types.Event) {
-			for e := range c {
-				err := s.OnEvent(e)
-				if err != nil {
-					meta, _ := s.GetMetadata()
-					log.Printf("error handling event by signature %s: %v", meta.Name, err)
-				}
-			}
-		}(s, c)
+		go signatureStart(s, c)
 	}
+	engine.signaturesMutex.RUnlock()
 	engine.consumeSources(done)
+}
+
+func (engine *Engine) unloadAllSignatures() {
+	engine.signaturesMutex.Lock()
+	defer engine.signaturesMutex.Unlock()
+	for sig, c := range engine.signatures {
+		sig.Close()
+		close(c)
+		delete(engine.signatures, sig)
+	}
+	engine.signaturesIndex = make(map[types.SignatureEventSelector][]types.Signature)
 }
 
 // matchHandler is a function that runs when a signature is matched
@@ -98,6 +128,7 @@ func (engine *Engine) consumeSources(done <-chan bool) {
 		select {
 		case event, ok := <-engine.inputs.Tracee:
 			if !ok {
+				engine.signaturesMutex.RLock()
 				for sig := range engine.signatures {
 					se, err := sig.GetSelectedEvents()
 					if err != nil {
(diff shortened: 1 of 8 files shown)


tracee-rules/engine/engine.go
```diff
@@ -62,24 +67,34 @@ func NewEngine(sigs []types.Signature, sources EventSources, output chan types.F
...
 
+// signatureStartFunc is the signature handling business logics.
+func (engine *Engine) signatureStartFunc(signature types.Signature, c chan types.Event) {
+	for {
+		select {
+		case e, ok := <-c:
+			if !ok {
+				// channel closed
+				return
+			}
```
itaysk, line 78: why did you prefer this over the current code for `e:= range c` which automatically handles channel close?

tracee-rules/engine/engine.go
```diff
@@ -62,24 +67,34 @@ func NewEngine(sigs []types.Signature, sources EventSources, output chan types.F
 	return &engine
 }
 
+// signatureStartFunc is the signature handling business logics.
+func (engine *Engine) signatureStartFunc(signature types.Signature, c chan types.Event) {
```
itaysk, line 71: does this func need to be attached to engine?

tracee-rules/engine/engine.go
```diff
@@ -62,24 +67,34 @@ func NewEngine(sigs []types.Signature, sources EventSources, output chan types.F
...
+
 // Start starts processing events and detecting signatures
 // it runs continuously until stopped by the done channel
 // once done, it cleans all internal resources, which means the engine is not reusable
 // note that the input and output channels are created by the consumer and therefore are not closed
 func (engine *Engine) Start(done chan bool) {
 	go engine.consumeSources(done)
+	engine.signaturesMutex.RLock()
 	for s, c := range engine.signatures {
-		defer close(c)
```
itaysk, line 72: you removed the sig-channel close, and only added it in the signature unload. doesn't it mean we will leak channels if the engine stops? (unless someone unloads all the signatures manually which is unlikely)

tracee-rules/engine/engine.go
```diff
@@ -108,17 +124,81 @@ func (engine *Engine) consumeSources(done <-chan bool) {
...
 			}
 		case <-done:
 			return
 		}
 	}
 }
+
+//LoadSignature will store in Engine data structures the given signature and activate its handling business logics.
+// It will return the signature ID as well as error.
+func (engine *Engine) LoadSignature(signature types.Signature) (string, error) {
```
itaysk, line 147: I see that you went with just load/unload API which also does the startsig/stopsig we discussed. That's fine with me I think just wanted to clarify you made this is intentional

tracee-rules/engine/engine.go
```diff
@@ -13,6 +15,7 @@ type Engine struct {
 	logger          log.Logger
 	signatures      map[types.Signature]chan types.Event
 	signaturesIndex map[types.SignatureEventSelector][]types.Signature
+	signaturesMutex sync.RWMutex
```
itaysk, line 18: can you explain we we need a RWMutex here instead of the simpler Mutex?

tracee-rules/engine/engine.go
itaysk: another option is to defer-unload all signatures in start

itaysk: also, see https://github.com/aquasecurity/tracee/pull/799

tracee-rules/engine/engine.go
mtcherni95: We will have multiple readers on the map, for example ListSignatures().

tracee-rules/engine/engine.go
```diff
@@ -108,17 +124,81 @@ func (engine *Engine) consumeSources(done <-chan bool) {
...
+	if engine.signatures[signature] != nil {
+		// signature already exists
+		return metadata.ID, nil
+	}
+	c := make(chan types.Event)
+	engine.signatures[signature] = c
+
+	// insert in engine.signaturesIndex map
+	for _, selectedEvent := range selectedEvents {
+		engine.signaturesIndex[selectedEvent] = append(engine.signaturesIndex[selectedEvent], signature)
```
yanivagman, line 166: Seems to be missing some logic as in NewEngine():
```
			if es.Name == "" {
				es.Name = "*"
			}
			if es.Source == "" {
				engine.logger.Printf("signature %s doesn't declare an input source", meta.Name)
			} else {
				engine.signaturesIndex[es] = append(engine.signaturesIndex[es], sig)
			}
```

tracee-rules/engine/engine.go
```diff
@@ -108,17 +124,81 @@ func (engine *Engine) consumeSources(done <-chan bool) {
...
+	}
+	go engine.signatureStartFunc(signature, c)
```
yanivagman, line 168: You start receiving events before signature init:
```
		err = sig.Init(engine.matchHandler)
		if err != nil {
			...
		}
```

tracee-rules/engine/engine.go
```diff
@@ -108,17 +124,81 @@ func (engine *Engine) consumeSources(done <-chan bool) {
...
+			if metadata.ID == sigId {
+				// signature found, remove it
+				signatures = append(signatures[:i], signatures[i+1:]...)
+				engine.signaturesIndex[selectedEvent] = signatures
+				break
+			}
+		}
+	}
+	close(c)
+	return nil
```
yanivagman, line 203: You defined a new signature.Close(), which is good, but I don't see it is being called here

tracee-rules/engine/engine.go
```diff
@@ -108,17 +124,81 @@ func (engine *Engine) consumeSources(done <-chan bool) {
...
+	c := engine.signatures[signature]
+
+	// remove from engine.signaturesIndex map
+	for _, selectedEvent := range selectedEvents {
+		signatures := engine.signaturesIndex[selectedEvent]
+		for i, sig := range signatures {
+			metadata, _ := sig.GetMetadata()
+			if metadata.ID == sigId {
+				// signature found, remove it
+				signatures = append(signatures[:i], signatures[i+1:]...)
```
yanivagman, line 196: This will panic if the signature is the last element in the slice (index i+1 is out of bounds).
As we don't care about the order of the signatures, you can do something like:
```
s[i] = s[len(s)-1]
s = s[:len(s)-1]
```

yanivagman (changes requested):

tracee-rules/engine/engine.go
mtcherni95: I don't think it will panic. Note I am taking 'i+1'.
https://play.golang.org/p/5zyN-OUAe0K

yanivagman: You are right.
Anyways, I think that the method I suggested will be more efficient, but not a must

tracee-rules/engine/engine.go
mtcherni95: nope, forgot to remove it.

tracee-rules/engine/engine.go
mtcherni95: yes, it is.

tracee-rules/engine/engine.go
```diff
@@ -108,17 +124,81 @@ func (engine *Engine) consumeSources(done <-chan bool) {
...
+	selectedEvents, err := signature.GetSelectedEvents()
+	if err != nil {
+		return "", fmt.Errorf("failed to store signature: %w", err)
+	}
+
+	// insert in engine.signatures map
+	metadata, _ := signature.GetMetadata()
+	engine.signaturesMutex.Lock()
+	defer engine.signaturesMutex.Unlock()
+	if engine.signatures[signature] != nil {
```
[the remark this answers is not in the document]
mtcherni95: why? isn't the signature instance enough?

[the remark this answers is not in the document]
yanivagman: If someone loads the same signature twice, I think that you will have two different instances of the same signature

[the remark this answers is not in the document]
mtcherni95: are you sure about that? if so then we should change all the indexing of the map according to the signature id, and to get the signature instance we should add a mapping between sig ID to sig instance.

I guess two instances of the same struct (signature) will be mapped to the same value in the map.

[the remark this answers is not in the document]
itaysk: @mtcherni95  is right https://play.golang.org/p/4nEF_Xrs-me

tracee-rules/engine/engine.go
mtcherni95: @yanivagman do we still want to add to the engine.signatures map if es.Source == "" ?

tracee-rules/engine/engine.go
```diff
@@ -62,24 +67,27 @@ func NewEngine(sigs []types.Signature, sources EventSources, output chan types.F
...
-				if err != nil {
-					meta, _ := s.GetMetadata()
-					log.Printf("error handling event by signature %s: %v", meta.Name, err)
-				}
-			}
-		}(s, c)
+		go signatureStartFunc(s, c)
 	}
+	engine.signaturesMutex.RUnlock()
 	<-done
```
yanivagman, line 91: Please rebase on main.
This code has changed

tracee-rules/engine/engine.go
mtcherni95: ConsumeSources is a blocking call. 
I think UnloadSignature is should be the one to close the signature channel. deferring unload to all signatures can be an idea.

tracee-rules/engine/engine.go
mtcherni95: done.

yanivagman (approved):
LGTM

tracee-rules/engine/engine.go
itaysk: I don't see a reason to do this. also today we reject it

tracee-rules/engine/engine.go
```diff
@@ -50,7 +56,9 @@ func NewEngine(sigs []types.Signature, sources EventSources, output chan types.F
 			if es.Source == "" {
 				engine.logger.Printf("signature %s doesn't declare an input source", meta.Name)
 			} else {
+				engine.signaturesMutex.Lock()
```
itaysk, line 59: why is `signatures` map not protected by the mutex?

tracee-rules/engine/engine.go
```diff
@@ -59,29 +67,46 @@ func NewEngine(sigs []types.Signature, sources EventSources, output chan types.F
 			continue
 		}
 	}
-	return &engine
+	return &engine, nil
+}
+
+// signatureStartFunc is the signature handling business logics.
```
itaysk, line 73: why does this function name ends with `Func`? all functions are "Funcs" aren't they :) Can we change to `startSignature`?

tracee-rules/engine/engine.go
```diff
@@ -59,29 +67,46 @@ func NewEngine(sigs []types.Signature, sources EventSources, output chan types.F
...
-		}(s, c)
+		go signatureStartFunc(s, c)
 	}
+	engine.signaturesMutex.RUnlock()
 	engine.consumeSources(done)
 }
 
+func (engine *Engine) unloadAllSignatures() {
+	engine.signaturesMutex.Lock()
+	defer engine.signaturesMutex.Unlock()
```
itaysk, line 99: generally I don't like deferring mutex unlocks, even though it's the same effect, it's prone to bugs in the future. Not big deal, it's a style thing, not a bug. if you like it this way it's fine but I think mutexs should be tightly controlled and released as soon as possible

tracee-rules/engine/engine.go
```diff
@@ -59,29 +67,46 @@ func NewEngine(sigs []types.Signature, sources EventSources, output chan types.F
...
+	for sig, c := range engine.signatures {
+		sig.Close()
+		close(c)
+		delete(engine.signatures, sig)
+	}
+	for selectedEvents, _ := range engine.signaturesIndex {
+		engine.signaturesIndex[selectedEvents] = nil
+	}
```
itaysk, line 107: this doesn't really clean the map as the keys will remain. you should copy L39 here

tracee-rules/engine/engine.go
```diff
@@ -107,17 +133,102 @@ func (engine *Engine) consumeSources(done <-chan bool) {
 						}
 					}
 				}
+				engine.signaturesMutex.RUnlock()
 				engine.inputs.Tracee = nil
 			} else if event != nil {
-				for _, s := range engine.signaturesIndex[types.SignatureEventSelector{Source: "tracee", Name: event.(tracee.Event).EventName}] {
+				engine.signaturesMutex.RLock()
+				assertedEvt, ok := event.(tracee.Event)
```
itaysk, line 140: for clarity, could rename `assertedEvt` to `traceeEvt` (or `traceeEvent`), we will have multiple input sources at some point

tracee-rules/engine/engine.go
mtcherni95: I guess I previously used it as a callback and forgot to rename :)

tracee-rules/engine/engine.go
mtcherni95: signatures and signaturesIndex are protected by the same lock as they refer to same objects and you want to be safe you don't have discrepancies between the data of the two data structures.

itaysk: I see that you've added the mutex around the L44. maybe just move it here under the same lock? not a big deal though

itaysk (approved):

itaysk:
@mtcherni95 just rebase and we can merge

itaysk (approved):

PR #802, opened 2021-07-01
Capture network

yanivagman:
This PR adds the ability to capture network traffic using Tracee.
The captured artifact is a pcap file which can be used as an input to programs such as wireshark for further analysis.

What makes this different than other network capture solutions, is that it is possible to use the regular filters of Tracee to only capture network traffic of the monitored processes. For example, one can use:
`--capture net=eth0 --trace comm=wget` to only capture network traffic of the wget process(es).

In addition, it is possible to get network debug events, which include the process context of the packets sent/received by the monitored processes using the `--debug` flag.

Note: currently, syn and syn-ack packets are not captured


tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2874,20 +3024,262 @@ int BPF_KPROBE(trace_security_socket_bind)
...
+        // However, in the case of a local server program and a local client program, which communicate via the 'lo' interface,
+        // the change to the 'TCP_ESTABLISHED' state of the server's socket doesn't happen in the (process) context of the server.
+        // To update the network_map correctly in that case, we use the 'TCP_LISTEN' state.
+        if (connect_id.port) {
+            if (!sock_ctx_p) {
+                net_ctx_ext.host_tid = bpf_get_current_pid_tgid();
+                bpf_get_current_comm(&net_ctx_ext.comm, sizeof(net_ctx_ext.comm));
+                net_ctx_ext.local_port = connect_id.port;
+                bpf_map_update_elem(&sock_ctx_map, &sk, &net_ctx_ext, BPF_ANY);
+                bpf_map_update_elem(&network_map, &connect_id, &net_ctx_ext, BPF_ANY);
```
roikol, line 3244: shouldn't the value be of type net_ctx_t?

yanivagman: No. This is by design to save some cycles. Notice that net_ctx is a "sub-struct" of net_ctx_ext. By updating the map with the net_ctx_ext, only the relevant fields will be saved

itaysk:
should we add some validation on the user input?

itaysk:
why do we need a separate debug flag? I think there should be a single debug mode which increases the verbosity of the error logs, and depending on enabled features it would print information relevant to the current run.

yanivagman:
> should we add some validation on the user input?

I don't think so. `attachTcProg()` will fail if the interface given by `--capture net=iface` is invalid

yanivagman:
`why do we need a separate debug flag? I think there should be a single debug mode which increases the verbosity of the error logs, and depending on enabled features it would print information relevant to the current run.`

For now, I can make this part of the present 'debug' flag.
The thing is that this flag provides valuable data, that can be used on its own to gain more insights about the network traffic used by a specific process, which is not visible in the pcap file.
When we capture the network traffic according to the filters given to tracee, we produce a pcap file without any process context - it represents the network traffic used by any of the processes which are traced by tracee.
When providing the 'debug-net' flag, one can know for each packet which process sent/received it.
In the future, we may add new events that will output this data, but for this PR we chose to keep it as "debug" info, because it is already quite big.

itaysk:
> I don't think so. attachTcProg() will fail if the interface given by --capture net=iface is invalid

Is this the desired behavior? don't we want to tell the user that input what's wrong instead of giving an internal error? it's the same as with kprobe events, we validate that the event is known and don't let the attachment fail

grantseltzer:
Can you please update README/Docs as part of this PR?

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2874,20 +2997,200 @@ int BPF_KPROBE(trace_security_socket_bind)
...
+    // In these cases, we won't pass the should_trace() check.
+    // To overcome this problem, we save the socket pointer in sock_ctx_map in states that we observed to have the correct context.
+    // We can then check for the existence of a socket in the map, and continue if it was traced before.
+    net_ctx_ext_t *sock_ctx_p = bpf_map_lookup_elem(&sock_ctx_map, &sk);
+    if (!sock_ctx_p) {
+        if (!should_trace()) {
+            return 0;
+        }
+    }
+
```
rafaeldtinoco, line 3193: This logic seems VERY good for me (saving socket pointer, updating the net bpf map on ESTABLISHED, etc). I'm only concerned on socket being shared among different tasks and logic synchronization. ​

One corner case I can think of would be: tracee is starting and will trace only new pids(). Or specific pids will be traced, but not others. It gets to trace PID 1 of 3 pids. All sharing the same socket (due to fd passing). During its start it starts tracing PID 3 but not PIDs 1 and 2.

PID 1 would change socket state, trigger the tracepoint, socket would exist in sock_ctx_map because of PID 2 (but now we are coming to this trace point from PID 1). 

> This would succeed in setting the new port (from 0 to something):
> 
> bpf_map_update_elem(&sock_ctx_map, &sk, &net_ctx_ext, BPF_ANY);
>
> This would fail:
> 
> bpf_map_update_elem(&network_map, &connect_id, &net_ctx_ext, BPF_ANY);
>
> because of net_ctx_ext.host_tid being different, wouldn't it ?
>
> Would this to make this logic fail ?

Of course this is a corner case, I'm not even sure it is possible... but that raises the concern of atomicity of having 2 maps and different handles/keys to them with parallel executions of different entry points.

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -3447,5 +3750,156 @@ int BPF_KPROBE(trace_security_bpf_map)
...
+        if (*head + offset > *tail) {
+            return false;
+        }
+    }
+
+    return true;
+}
+
+static __always_inline int tc_probe(struct __sk_buff *skb, bool ingress) {
```
rafaeldtinoco, line 3859: I'm here on my review. Have almost finished the golang part, finishing this and I'm ready to finish it.

tracee-ebpf/tracee/tracee.bpf.c
yanivagman: We must have two different maps here. The explanation for this is simple:
1. The network map is used by the tc bpf program, where there is no process context, so the map key should only contain network related fields, and not process or socket related stuff
2. On the other hand, we want to save the context between two different tcp states, where either process context is "incorrect" or network context is incomplete (no local port).

So if we are using this tracepoint, I don't see any other choice.

Anyways, I don't see a problem with the use case you presented. The pid is not part of the key in any of the maps. So if a socket was inserted to the map, it will stay there until it is closed. Also, the network map don't care about the pid (which is part of the map value) as long as the key exists. One more thing, as we use BPF_ANY to update the map, it will be updated even if the key exist or not

tracee-ebpf/tracee/tracee.go
```diff
@@ -1493,6 +1673,46 @@ func (t *Tracee) processFileWrites() {
...
+			// Sanity check - timestamp, host tid, comm and packet length must exist in all net events
+			if len(in) < 32 {
+				continue
+			}
+
+			timeStamp := binary.LittleEndian.Uint64(in[0:8])
+			pktLen := binary.LittleEndian.Uint32(in[8:12])
+
+			info := gopacket.CaptureInfo{
+				Timestamp:      time.Unix(0, int64(timeStamp)),
```
rafaeldtinoco: I'm getting:

```
$ tcpdump -n -r ./capture.pcap
reading from file ./capture.pcap, link-type EN10MB (Ethernet)
-2:-41:-32.988136 IP 127.0.0.1.54207 > 127.0.0.1.8080: UDP, length 5
-2:-41:-32.988136 IP 127.0.0.1.54207 > 127.0.0.1.8080: UDP, length 5
-2:-41:-32.988137 IP 127.0.0.1.8080 > 127.0.0.1.54207: UDP, length 5
-2:-41:-32.988137 IP 127.0.0.1.8080 > 127.0.0.1.54207: UDP, length 5
-2:-41:-31.055386 IP 127.0.0.1.33117 > 127.0.0.1.8080: UDP, length 5
-2:-41:-31.055386 IP 127.0.0.1.33117 > 127.0.0.1.8080: UDP, length 5
-2:-41:-31.056352 IP 127.0.0.1.8080 > 127.0.0.1.33117: UDP, length 5
-2:-41:-31.056352 IP 127.0.0.1.8080 > 127.0.0.1.33117: UDP, length 5
```

this timeStamp seems broken. Trying to check what could be wrong I saw:

pkt.ts = bpf_ktime_get_ns()/1000;

You are transforming the nanosec into microsec in tracee.bpf.c -> tc_probe. 

I think It should be kept as nanosec as:

```
timeStamp := time.Unix(0, timeStamp))
```

will have ns as the second argument as nanosec:

```
func Unix(sec int64, nsec int64) Time {
```

I have believe it is right to consider it is little endian as this is not coming from the network stack but from task_struct.

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -3028,16 +3055,30 @@ int BPF_KPROBE(trace_security_socket_bind)
...
+        __builtin_memcpy(net_ctx.comm, context.comm, TASK_COMM_LEN);
         bpf_map_update_elem(&network_map, &connect_id, &net_ctx, BPF_ANY);
     }
 
     events_perf_submit(ctx);
 
+    // netDebug event
+    if (get_config(CONFIG_DEBUG_NET)) {
+        net_debug_t debug_event = {0};
+        debug_event.ts = bpf_ktime_get_ns()/1000;
```
rafaeldtinoco, line 3067: 1199583.960716  nc                2618160  udp_sendmsg          LocalIP: 127.0.0.1, LocalPort: 36895, Protocol: 17
1199583.960755  nc                2618160  packet               Len: 47, SrcIP: 127.0.0.1, SrcPort: 36895, DestIP: 127.0.0.1, DestPort: 14219, Protocol: 17
1199583.960771  nc                2618160  packet               Len: 47, SrcIP: 127.0.0.1, SrcPort: 36895, DestIP: 127.0.0.1, DestPort: 14219, Protocol: 17

Timestamp from debug-net looks good. I think only the pcap timestamp generation was broken (like said in the other commit's comment).

rafaeldtinoco: BTW as you can see I went from commit to commit (That is why debug port is still broken in this output). I hadn't applied the little->big endian fix yet.

rafaeldtinoco:
All the rest looks pretty good to me. I left the 2 previous comments just so you can re-verify the timestamp issue with pcap.. that was the only issue I could find when digging into code and trying to exercise the execution paths you added within this patchset.

rafaeldtinoco:
> `why do we need a separate debug flag? I think there should be a single debug mode which increases the verbosity of the error logs, and depending on enabled features it would print information relevant to the current run.`
> 
> For now, I can make this part of the present 'debug' flag.
> The thing is that this flag provides valuable data, that can be used on its own to gain more insights about the network traffic used by a specific process, which is not visible in the pcap file.
> When we capture the network traffic according to the filters given to tracee, we produce a pcap file without any process context - it represents the network traffic used by any of the processes which are traced by tracee.
> When providing the 'debug-net' flag, one can know for each packet which process sent/received it.
> In the future, we may add new events that will output this data, but for this PR we chose to keep it as "debug" info, because it is already quite big.

I imagine to get debug-net messages only we would have to grep for:

packet | security_socket_bind | udp_sendmsg | __udp_disconnect | udp_destroy_sock | udpv6_destroy_sock |  inet_sock_set_state


```
|SOCK_CLOEXEC, protocol: 0
1199583.468936 0      ip               189500  189500  0                bind                 sockfd: 3, addr: {'sa_family': 'AF_NETLINK'}, addrlen: 12
1199583.468953 0      ip               189500  189500  0                getsockname          sockfd: 3, addr: {'sa_family': 'AF_NETLINK'}, addrlen: 0x7FFDEFFC809C
1199583.469384 0      ip               189500  189500  0                sched_process_exit
1199583.960716  nc                2618160  udp_sendmsg          LocalIP: 127.0.0.1, LocalPort: 36895, Protocol: 17
1199583.960755  nc                2618160  packet               Len: 47, SrcIP: 127.0.0.1, SrcPort: 36895, DestIP: 127.0.0.1, DestPort: 14219, Protocol: 17
1199583.960771  nc                2618160  packet               Len: 47, SrcIP: 127.0.0.1, SrcPort: 36895, DestIP: 127.0.0.1, DestPort: 14219, Protocol: 17
1199584.472502 0      nordvpnd         189501  189501  0                execve               pathname: /usr/sbin/ip, argv: [ip -4 route list default]
1199584.472732 0      nordvpnd         189501  189501  0                security_bprm_check  pathname: /usr/bin/ip, dev: 142, inode: 2569
```

should we prefix debug-net messages so we can grep them easily (when developing net features, comparing debug msgs to what was writen to the pcap files, and so on) ?

tracee-ebpf/tracee/tracee.go
rafaeldtinoco: Working fine now:

```
14:14:34.409831 IP 127.0.0.1.53740 > 127.0.0.1.12345: Flags [.], ack 1741000412, win 512, options [nop,nop,TS val 1494266592 ecr 1494266592], length 0
14:14:34.409841 IP 127.0.0.1.53740 > 127.0.0.1.12345: Flags [.], ack 1, win 512, options [nop,nop,TS val 1494266592 ecr 1494266592], length 0
14:14:34.880155 IP 127.0.0.1.53740 > 127.0.0.1.12345: Flags [P.], seq 0:5, ack 1, win 512, options [nop,nop,TS val 1494267063 ecr 1494266592], length 5
```

rafaeldtinoco:
Was able to filter debug_net messages pretty well:

$ sudo ./dist/tracee-ebpf --debug --trace pid=new --capture net=lo --capture dir:/tmp/capture --capture clear-dir 2>&1 | grep debug_net

```
2021-07-20 14:17:49.232029067 -0300 -03  nc                532423   debug_net/packet               Len: 71, SrcIP: 127.0.0.1, SrcPort: 53748, DestIP: 127.0.0.1, DestPort: 12345, Protocol: 6
2021-07-20 14:17:49.232055455 -0300 -03  nc                532423   debug_net/packet               Len: 71, SrcIP: 127.0.0.1, SrcPort: 53748, DestIP: 127.0.0.1, DestPort: 12345, Protocol: 6
2021-07-20 14:17:49.232088472 -0300 -03  nc                532423   debug_net/packet               Len: 66, SrcIP: 127.0.0.1, SrcPort: 12345, DestIP: 127.0.0.1, DestPort: 53748, Protocol: 6
2021-07-20 14:17:49.232095969 -0300 -03  nc                532423   debug_net/packet               Len: 66, SrcIP: 127.0.0.1, SrcPort: 12345, DestIP: 127.0.0.1, DestPort: 53748, Protocol: 6
2021-07-20 14:17:50.301660096 -0300 -03  nc                532423   debug_net/packet               Len: 71, SrcIP: 127.0.0.1, SrcPort: 12345, DestIP: 127.0.0.1, DestPort: 53748, Protocol: 6
2021-07-20 14:17:50.301705547 -0300 -03  nc                532423   debug_net/packet               Len: 71, SrcIP: 127.0.0.1, SrcPort: 12345, DestIP: 127.0.0.1, DestPort: 53748, Protocol: 6
```

And they all seem fine. Timestamps from debug are also good in comparison to pcap file:

```
14:17:48.553420 IP 127.0.0.1.53748 > 127.0.0.1.12345: Flags [S], seq 3169578644, win 65495, options [mss 65495,sackOK,TS val 1494460864 ecr 0,nop,wscale 7], length 0
14:17:48.553436 IP 127.0.0.1.53748 > 127.0.0.1.12345: Flags [S], seq 3169578644, win 65495, options [mss 65495,sackOK,TS val 1494460864 ecr 0,nop,wscale 7], length 0
14:17:48.553470 IP 127.0.0.1.12345 > 127.0.0.1.53748: Flags [S.], seq 2668228879, ack 3169578645, win 65483, options [mss 65495,sackOK,TS val 1494460865 ecr 1494460864,nop,wscale 7], length 0
```

+1 from me.

rafaeldtinoco (approved):
+1 from me after testing the new changes. Loved the network capture feature!

PR #805, opened 2021-07-03
fix(benchmark): Use uniquely identifiable sigs in BenchmarkEngineWithNSignatures

danielpacak:

### DIFF
--- a/tracee-rules/benchmark/benchmark_test.go
+++ b/tracee-rules/benchmark/benchmark_test.go
@@ -83,7 +83,8 @@ func BenchmarkEngineWithCodeInjection(b *testing.B) {
 				s, err := bc.sigFunc()
 				require.NoError(b, err, bc.name)
 
-				e := engine.NewEngine([]types.Signature{s}, inputs, output, os.Stderr)
+				e, err := engine.NewEngine([]types.Signature{s}, inputs, output, os.Stderr)
+				require.NoError(b, err, "constructing engine")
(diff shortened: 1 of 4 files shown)


tracee-rules/benchmark/benchmark_test.go
```diff
@@ -175,9 +177,10 @@ func BenchmarkEngineWithNSignatures(b *testing.B) {
 	for _, bc := range benches {
 		for _, tc := range bc.sigCount {
 			b.Run(fmt.Sprintf("%s/%dSignatures", bc.name, tc), func(b *testing.B) {
-				sig, _ := bc.sigFunc()
```
danielpacak, line 178: We always ended up with 1 channel for N signatures due to this bug. It was pretty hard to catch so I also added an assert to the `engine.NewEngine` constructor.

itaysk (commented):
where is the code change that is implementing "uniquely identifiable sigs"? does simply moving the call to `sigFunc` do that?

danielpacak:
> where is the code change that is implementing "uniquely identifiable sigs"? does simply moving the call to `sigFunc` do that?



> where is the code change that is implementing "uniquely identifiable sigs"? does simply moving the call to `sigFunc` do that?

The fix is to invoke `sigFunc` inside the `for i := range sigs`. Apparently Go compares pointer values here.

itaysk (approved):

Issue #806, opened 2021-07-05
recording context timestamp at sys_enter


SericaLaw:
There are scenarios that events are taking place in a nested way, e.g. when typing `ls -a` in a terminal.

For references, LTTng's output likes:

```
[19:37:12.099134427] (+4.226055818) ubuntu syscall_entry_clone: { cpu_id = 2 }, { tid = 26435, vtid = 26435, ppid = 26434, vppid = 26434, procname = "bash", pid_ns = 4026531836, net_ns = 4026532000 }, { clone_flags = 0x1200011, newsp = 0x0, parent_tid = 0x0, child_tid = 0x7F22F2D99A10 }
[19:37:12.099283972] (+0.000149545) ubuntu sched_process_fork: { cpu_id = 2 }, { tid = 26435, vtid = 26435, ppid = 26434, vppid = 26434, procname = "bash", pid_ns = 4026531836, net_ns = 4026532000 }, { parent_comm = "bash", parent_tid = 26435, parent_pid = 26435, parent_ns_inum = 4026531836, child_comm = "bash", child_tid = 48877, _vtids_length = 1, vtids = [ [0] = 48877 ], child_pid = 48877, child_ns_inum = 4026531836 }
[19:37:12.099360860] (+0.000039624) ubuntu syscall_exit_clone: { cpu_id = 2 }, { tid = 26435, vtid = 26435, ppid = 26434, vppid = 26434, procname = "bash", pid_ns = 4026531836, net_ns = 4026532000 }, { ret = 48877 }
[19:37:12.099416895] (+0.000056035) ubuntu syscall_exit_clone: { cpu_id = 3 }, { tid = 48877, vtid = 48877, ppid = 26435, vppid = 26435, procname = "bash", pid_ns = 4026531836, net_ns = 4026532000 }, { ret = 0 }

[19:37:12.099713157] (+0.000296262) ubuntu syscall_entry_execve: { cpu_id = 3 }, { tid = 48877, vtid = 48877, ppid = 26435, vppid = 26435, procname = "bash", pid_ns = 4026531836, net_ns = 4026532000 }, { filename = "/usr/bin/ls", argv = 0x55579F56E6F0, envp = 0x55579F55FB90 }
[...]

PR #807, opened 2021-07-05
tracee-ebpf: record context timestamp at sys_enter

SericaLaw:
Hi @yanivagman thanks for your reply to my issue ( #806 ), I have created a PR to record timestamp at syscall entry. My approach is to add another bpf map to keep the records in order to affect least range of bpf code.

### DIFF
--- a/tracee-ebpf/tracee/tracee.bpf.c
+++ b/tracee-ebpf/tracee/tracee.bpf.c
@@ -327,7 +327,7 @@ typedef struct event_context {
 } context_t;
 
 typedef struct args {
-    unsigned long args[6];
+    unsigned long args[7]; // the last element of this array is used to save the function entry timestamp
 } args_t;
 
 typedef struct bin_args {
@@ -1477,6 +1477,7 @@ static __always_inline int load_args(args_t *args, bool delete, u32 event_id)
     args->args[3] = saved_args->args[3];
     args->args[4] = saved_args->args[4];
     args->args[5] = saved_args->args[5];
+    args->args[6] = saved_args->args[6];
 
     if (delete)
         bpf_map_delete_elem(&args_map, &id);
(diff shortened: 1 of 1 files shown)


yanivagman:
Thanks @SericaLaw.
Note that the function init_and_save_context is being called from almost any place that we send events to userspace, not only from sys_enter/sys_exit.
Also, now the code initialize context.ts twice - once in init_context, and then again in init_and_save_context.

I think that a better approach will be to use the already existing args_map by expanding the save_args and load_args functions to save the timestamp. WDYT?

SericaLaw:
Hi @yanivagman I thought about the expanding `save_args` and `load_args` idea, but I think it will change the semantic of these functions, so I wrote timestamp-specific functions.

I agree that the current commit will "init" `context.ts` twice, and if the map entry missing in other cases will cause performance loss. Since `sys_exit` will use `init_and_save_context` to directly submit context to the buffer, so we may need to let it know the ts when `init_context`, which leads to adding params to these functions or write new functions, or we may need to re-submit the context. What's the best practice you think?

yanivagman:
I'll try to explain the usage of the `save_args()` and `load_args()` functions and how I think that we can do this change without changing their semantics. These functions can be renamed (although not a must) to reflect the fact that they also save/load the ts (for example, to something like `save(load)_fentry_data()`)

save_args() is being called from the following locations:
1. sys_enter tracepoint - this is what the issue is about
2. save_args_from_regs() - which is only called from the TRACE_ENT_FUNC macro (used by vfs_write_writev functions)
3. sys_exit tracepoint - which is used by the execve tailcalls

load_args() is being called from the following locations:
1. sys_exit tracepoint - this is what this issue is about
2. TRACE_RET_FUNC macro - which is currently not in use
3. the execve tailcalls
4. vfs_write_writev functions
5. trace_mmap_alert and trace_mprotect_alert

For all of the above, I don't see a problem with saving/loading an extra timestamp, and using it to update the context instead of the existing timestamp which was initialized by init_context(). When we don't care about saving/updating the ts, we can just pass 0 for the ts argument.

Here is a possible implementation (which I didn't test) that saves the extra timestamp as an "argument" to avoid changing the map value type
```
typedef struct args {
    unsigned long args[7]; // the last element of this array is used to save the function entry timestamp
} args_t;

static __always_inline int save_fentry_data(u32 event_id, args_t *args, u64 ts)
{
    u64 id = event_id;
    u32 tid = bpf_get_current_pid_tgid();
    id = id << 32 | tid;
    args[7] = ts;
    bpf_map_update_elem(&fentry_data_map, &id, args, BPF_ANY);

    return 0;
}
```

SericaLaw:
Hi @yanivagman I also tested my motivating example ( `ls -a` ) with my updated commit and it worked correctly. I agree that resaving the context is concise and readable enough since it also appears in other places like updating argnum. I have updated the code.

yanivagman (approved):
LGTM

PR #808, opened 2021-07-05
fix panic due to slice outbound

mtcherni95:


tracee-rules/signatures/rego/regosig/traceerego.go
```diff
@@ -43,7 +43,13 @@ func NewRegoSignature(regoCodes ...string) (types.Signature, error) {
 
 	var pkgName string
 	for _, regoCode := range regoCodes {
-		regoModuleName := strings.Split(re.FindString(regoCode), " ")[1]
+		var regoModuleName string
+		splittedName := strings.Split(re.FindString(regoCode), " ")
+		if len(splittedName) > 1 {
+			regoModuleName = splittedName[1]
+		} else {
```
mtcherni95, line 50: not sure if we should return here...
anyways we could here panic if there is no entry at index 1.

itaysk:
Can you explain what you did to cause this panic? did you pass an invalid rego file? tracee-rules assumes it gets valid rego files, it doesn't try to validate them.

mtcherni95:
> Can you explain what you did to cause this panic? did you pass an invalid rego file? tracee-rules assumes it gets valid rego files, it doesn't try to validate them.
A scenario for the panic could be the following: 
1. read from disk a rego file.
2. for some reason, the read operation failed and the user of tracee-rules didn't check the error code from the read operation and pass the buffer to NewRegoSiganture API (buffer nil).

Tracee-Rules is exporting the API of NewRegoSignature so I wouldn't be comfortable with the assumption that it gets valid files. The library should validate the external input or at least not panic.

mtcherni95:
> Can you explain what you did to cause this panic? did you pass an invalid rego file? tracee-rules assumes it gets valid rego files, it doesn't try to validate them.
A scenario for the panic could be the following: 
1. read from disk a rego file.
2. for some reason, the read operation failed and the user of tracee-rules didn't check the error code from the read operation and pass the buffer to NewRegoSiganture API (buffer nil).

Tracee-Rules is exporting the API of NewRegoSignature so I wouldn't be comfortable with the assumption that it gets valid files. The library should validate the external input or at least not panic.

tracee-rules/signatures/rego/regosig/traceerego.go
itaysk: Continue?

mtcherni95: well, not exactly: if the signatures imports a helper.rego then it will output weird logs or have an incorrect behaviour. That's why I would prefer to just return error.

itaysk: failing the entire process just because one rego policy was invalid doesn't feel right to me (think also about the security implication of that). Of course today this is what happens because we don't handle the error at all, but if we take the opportunity to fix it, let's choose the correct outcome.

I think continue is better:
- signature invalid: only the sig will be skipped
- tracee helper is invalid: not possible, we wrote the helpers so we can assume it's valid
- user-defined helper is invalid: only the user-provided policies will fail

^^ that sounds like a more reasonable fail model than always quit the process.

WDYT?

mtcherni95: what do you mean by security implication? the user of the library tries to load an incomplete signature, you shouldn't allow that. but sure, if you want we can have a continue statement.

itaysk: it's not just my decision, I genuinely ask what you think. by security implications I meant that someone can drop a malformed rego file in the rules dir and break tracee (if you will return an error here)

grantseltzer: I think continuing is fine as long as long as it's properly logged (regardless of if debug is set) or otherwise communicated to the user. If we do return, the error message should be expanded so the user knows which rego file was invalid.

mtcherni95: I believe returning is safer as a user could load a signatures that uses a helper file (for rego) and succeeding in loading (the function will report any error, but once the signature is loaded then strange error logs will be output. by returning we will be on the safe side. that's my opinion. 
Using a continue statement + log will not prevent from printing errors once the signature will not find its helper file.

itaysk: after a closer look (thanks @yanivagman) I realized that returning an error does exactly what I wanted to happen, since the caller (findRegoSigs) will do that continue.

itaysk (approved):

PR #810, opened 2021-07-08
fix help message

mtcherni95:


yanivagman (approved):

yanivagman (commented):
Please also update test_main tests, as they fail after this change

yanivagman (changes requested):
tests fail

mtcherni95:
> Please also update test_main tests, as they fail after this change

done, tests pass now.

yanivagman (approved):

PR #812, opened 2021-07-10
tracee-ebpf: fix incomplete path

### LINKED ISSUE #811: saved path might be incomplete
When running tracee-ebpf with vfs_write event, path might be incomplete.
For example:
`sudo ./dist/tracee-ebpf --trace e=vfs_write --trace pid=new`
```
TIME(s)        UID    COMM             PID     TID     RET              EVENT                ARGS
50626.809466   1000   ls               96263   96263   173              vfs_write            pathname: /pts/1, dev: 27, inode: 4, count: 173, pos: 0
```

While the output should have been:
```
TIME(s)        UID    COMM             PID     TID     RET              EVENT                ARGS
50626.809466   1000   ls               96263   96263   173              vfs_write            pathname: /dev/pts/1, dev: 27, inode: 4, count: 173, pos: 0
```

yanivagman:
fix #811 

Bug introduced in #758

### DIFF
--- a/tracee-ebpf/tracee/tracee.bpf.c
+++ b/tracee-ebpf/tracee/tracee.bpf.c
@@ -1307,7 +1307,7 @@ static __always_inline int save_path_to_str_buf(buf_t *string_p, const struct pa
                 bpf_probe_read(&dentry, sizeof(struct dentry*), &mnt_p->mnt_mountpoint);
                 bpf_probe_read(&mnt_p, sizeof(struct mount*), &mnt_p->mnt_parent);
                 bpf_probe_read(&mnt_parent_p, sizeof(struct mount*), &mnt_p->mnt_parent);
-                bpf_probe_read(&vfsmnt, sizeof(struct vfsmount*), (const void*)&mnt_p->mnt);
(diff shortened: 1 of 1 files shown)


tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1307,7 +1307,7 @@ static __always_inline int save_path_to_str_buf(buf_t *string_p, const struct pa
                 bpf_probe_read(&dentry, sizeof(struct dentry*), &mnt_p->mnt_mountpoint);
                 bpf_probe_read(&mnt_p, sizeof(struct mount*), &mnt_p->mnt_parent);
                 bpf_probe_read(&mnt_parent_p, sizeof(struct mount*), &mnt_p->mnt_parent);
-                bpf_probe_read(&vfsmnt, sizeof(struct vfsmount*), (const void*)&mnt_p->mnt);
```
yanivagman, line 1310: The problem here is that mnt_p->mnt is of type `struct vfsmount` and not `struct vfsmount*`

grantseltzer (approved):
Good catch :+1:

Issue #783, opened 2021-06-22
Embed CO:RE bpf object file in Go binary at compile time


grantseltzer:
When CORE=y is specified with `make` the resulting object file should be embedded into the go binary. This avoids any issue with the go binary trying to find the correct object. 

Similarly we should consider statically compiling the go binary when CORE is enabled

yanivagman:
Done in #818

itaysk:
also need to update the release workflow @grantseltzer

PR #818, opened 2021-07-13
Add embed directive to embed the compiled CORE bpf object into go binary

grantseltzer:
Handles https://github.com/aquasecurity/tracee/issues/783


tracee-ebpf/Makefile
```diff
@@ -23,7 +23,7 @@ BPF_SRC := tracee/tracee.bpf.c
 ifndef CORE
 OUT_BPF := $(OUT_DIR)/tracee.bpf.$(subst .,_,$(KERN_RELEASE)).$(subst .,_,$(VERSION)).o
 else
-OUT_BPF := $(OUT_DIR)/tracee.bpf.core.$(subst .,_,$(VERSION)).o
```
yanivagman, line 26: Why did you remove tracee version from the file?
Tracee binary version should match the bpf code version

itaysk (commented):
also missing updates to the main Makefile

tracee-ebpf/Makefile
grantseltzer: For the embed directive I can't pass the build time variable into the actual directive comment, so it'd be impossible to find it. Plus if it's being embedded anyway, why does it need the version?

tracee-ebpf/Makefile
```diff
@@ -47,9 +47,9 @@ build: $(OUT_BIN)
 
 go_env := GOOS=linux GOARCH=$(ARCH:x86_64=amd64) CC=$(CMD_CLANG) CGO_CFLAGS="-I $(abspath $(LIBBPF_HEADERS))" CGO_LDFLAGS="$(abspath $(LIBBPF_OBJ))"
 ifndef DOCKER
-$(OUT_BIN): $(LIBBPF_HEADERS) $(LIBBPF_OBJ) $(filter-out *_test.go,$(GO_SRC)) $(BPF_BUNDLE) | $(OUT_DIR)
+$(OUT_BIN): $(LIBBPF_HEADERS) $(LIBBPF_OBJ) $(OUT_DIR)/tracee.bpf.core.o $(filter-out *_test.go,$(GO_SRC)) $(BPF_BUNDLE) | $(OUT_DIR)
 	$(go_env) go build -v -o $(OUT_BIN) \
-	-ldflags "-X main.version=$(VERSION) -X main.bpf_core=$(CORE)"
+	-ldflags "-extldflags '-static' -X main.version=$(VERSION) -X main.bpf_core=$(CORE)"
```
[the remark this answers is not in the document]
grantseltzer: This is a good call out, I suspect this actually breaks something when building on 4.18. Doing more investigation.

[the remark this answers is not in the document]
grantseltzer: Ok it turns out that this adds the dependencies of static libraries (i.e. `zlib-static`, `libelf-static`, etc...). Red Hat doesn't package these for RHEL (but does for fedora/centos), so perhaps it'd be better to leave out static, or make it an additional make flag option.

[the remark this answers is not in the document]
itaysk: If that's the only issue than it's not a big deal IMO as long as we document it. Not sure how a flag will help this is for us who are building tracee we will need to decide how to build it anyway

itaysk:
how is this fixing #783?

tracee-ebpf/Makefile
```diff
@@ -47,9 +47,9 @@ build: $(OUT_BIN)
 
 go_env := GOOS=linux GOARCH=$(ARCH:x86_64=amd64) CC=$(CMD_CLANG) CGO_CFLAGS="-I $(abspath $(LIBBPF_HEADERS))" CGO_LDFLAGS="$(abspath $(LIBBPF_OBJ))"
 ifndef DOCKER
-$(OUT_BIN): $(LIBBPF_HEADERS) $(LIBBPF_OBJ) $(filter-out *_test.go,$(GO_SRC)) $(BPF_BUNDLE) | $(OUT_DIR)
+$(OUT_BIN): $(LIBBPF_HEADERS) $(LIBBPF_OBJ) $(OUT_DIR)/tracee.bpf.core.o $(filter-out *_test.go,$(GO_SRC)) $(BPF_BUNDLE) | $(OUT_DIR)
```
itaysk, line 50: here you require the CORE binary explicitly (not respecting OUT_BPF) and most importantly for other regular (non-CORE) builds

tracee-ebpf/Makefile
```diff
@@ -119,7 +119,7 @@ $(OUT_DIR)/tracee.bpf.%.o: $(BPF_SRC) $(LIBBPF_HEADERS) | $(OUT_DIR) $(bpf_compi
 	rm $(@:.o=.ll)
 else
 #docker no, CO:RE yes
-$(OUT_DIR)/tracee.bpf.core.%.o: $(BPF_SRC) $(LIBBPF_HEADERS) $(CMD_CLANG)
+$(OUT_DIR)/tracee.bpf.core.o: $(BPF_SRC) $(LIBBPF_HEADERS) $(CMD_CLANG)
```
itaysk, line 122: doesn't respect OUT_BPF variable.

itaysk: I didn't get to review the original CORE PR in time so making a comment about it here: I think the user-defined CORE variable wasn't the best choice. core object is a different artifact than the regular object file, so it should be a simple make target. unlike the DOCKER variable which produces the same artifacts in a different way, the CORE variable augments the build process to produce an entirely new artifact. I think this might be unnecessarily complicating the makefile. don't want it to block anything now, just sharing my thoughts

tracee-ebpf/main.go
```diff
@@ -26,6 +26,7 @@ var buildPolicy string
 
 // These vars are supposed to be injected at build time
 //go:embed "dist/tracee.bpf/*"
+//go:embed "dist/tracee.bpf.core.o"
```
itaysk, line 29: the makefile target is customizable (user can choose the name of the artifact) but the go code hardcodes the default name. It's really not important for anyone so I don't think we should do anything different now (I have some ideas for the future) just making the note here in context for future reference.

tracee-ebpf/Makefile
itaysk: SGTM

tracee-ebpf/Makefile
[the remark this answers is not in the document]
grantseltzer: Leaving it out for now.

tracee-ebpf/main.go
```diff
@@ -1077,6 +1095,23 @@ func getBPFObject() (string, error) {
 	return bpfObjFilePath, nil
 }
 
+func unpackCOREBinary() (string, error) {
+	b, err := bpfBundleInjected.ReadFile("dist/tracee.bpf.core.o")
+	if err != nil {
+		return "", err
+	}
+	f, err := ioutil.TempFile("/tmp/tracee", "tracee.bpf.core.o")
```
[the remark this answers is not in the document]
grantseltzer: I'm not sure I understand your feedback. /tmp/tracee is just a place to intermediately place the core object when unloading it from the embedded bytes. One alternative is to rewire tracee to load the object from memory and use libbpfgo's `NewModuleFromBuffer` instead.

grantseltzer:
I've pushed a commit that seems to be the only way to conditionally use the embed directive. I use a "core" build flag with two different files that build depending on if the tag is present or not.

tracee-ebpf/main.go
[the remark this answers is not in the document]
itaysk: I'll try to elaborate. Let me know if I misunderstood something:
There are 2 special directories for Tracee: "installPath" and "outputPath". both are user-configurable, and both default to `/tmp/tracee`. Nothing in Tracee writes directly to `/tmp/tracee`, we always write either to "installPath" or "outputPath". In this case, I think you meant to put the CORE object file under "installPath".
Now what I was saying is that this "installPath" is a permanent directory (the fact that it defaults to /tmp is confusing.. it's not very temporary in its nature). So my feedback was that for that reason I don't think it's a good "place to intermediately place the core object when unloading it from the embedded bytes". 
But even if you did want to place the intermediary CORE file in the "installPath", the code is missing a few things:
1. use the variable name (installPath), instead of hardcoding the default path
2. ensure that the path exists (like we do in other cases where we write there), otherwise it will fail
3. clean after it (if it's really an intermediary file it should be gone as soon as possible).
I also think that if it's really an intermediary file, it doesn't matter where it is and it's better to let `TempFile` generate a path under `/tmp`

In the same comment I was also wondering if we really need to "unpack" the embedded file to a temp file at all, or we can just `Open` it directly from `embed.FS`. I think libbpfgo should accept an `io.reader` instead (or in addition) to `[]bytes` as this is better Go code, but if it's expanding the scope of the PR let's leave it for some other time.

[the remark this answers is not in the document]
grantseltzer: Ah ok this all makes sense. I'm going to change this to just use the file system's default tempfs. 

Regarding not writing to file at all - you're correct that there's a way to not write it to a file at all. The best way to do this would be to read it all into memory and load it directly with libbpf. When originally writing this I didn't think it was important to rewire tracee to support it but i'm open to it if you feel it's be a significant improvement.

tracee-ebpf/main.go
```diff
@@ -1077,6 +1090,28 @@ func getBPFObject() (string, error) {
 	return bpfObjFilePath, nil
 }
 
+func unpackCOREBinary() (string, error) {
+	b, err := coreObject.ReadFile("dist/tracee.bpf.core.o")
+	if err != nil {
+		return "", err
+	}
+	f, err := ioutil.TempFile("", "tracee.bpf.core.o")
```
itaysk, line 1098: we'll need to clean up after this file creation on the user machine

itaysk (commented):
just one last todo

tracee-ebpf/main.go
```diff
@@ -1051,6 +1053,17 @@ func getBPFObject() (string, error) {
...
 		traceeInstallPath,
 	}
+
+	bpfObjFileName := fmt.Sprintf("tracee.bpf.%s.%s.o", strings.ReplaceAll(tracee.UnameRelease(), ".", "_"), strings.ReplaceAll(version, ".", "_"))
+	if bpf_core {
+		s, err := unpackCOREBinary()
+		if err != nil {
+			return "", err
+		}
+		bpfObjFileName = s
```
yanivagman, line 1063: At this point, can't we already return as we have the bpf object at hand?

grantseltzer:
I've pushed a change which reads the CO:RE binary straight from the embedded go binary into a slice of bytes, and loads it using `libbpfgo.NewModuleFromBuffer`. I also have changed the regular bpf object reading logic to do the same thing. That is to keep the code clean, we're loading it into memory from file before regardless.

grantseltzer:
After discussing with @yanivagman we came to the agreement that for the sake of optimal portability, the CO:RE should always be built (since BTF kernel is not required, just the vmlinux.h file) and embedded. 

At runtime tracee-ebpf should check if it's running on a BTF enabled kernel (see if /sys/kernel/btf exists). If there is BTF, use that embedded CORE binary. If not, look for the compiled bpf object, and if it doesn't exist attempt to build it from the embedded source. 

I will push a commit with this soon. @itaysk @rafaeldtinoco any thoughts?

rafaeldtinoco:
> After discussing with @yanivagman we came to the agreement that for the sake of optimal portability, the CO:RE should always be built (since BTF kernel is not required, just the vmlinux.h file) and embedded.

+1. Will start giving us the flexibility we want/need.

> At runtime tracee-ebpf should check if it's running on a BTF enabled kernel (see if /sys/kernel/btf exists). If there is BTF, use that embedded CORE binary. If not, look for the compiled bpf object, and if it doesn't exist attempt to build it from the embedded source.

So, wrapping up.. with your change we will have:

1. check for /sys/kernel/btf
2. if yes, use CO-RE binary
3. if no, look for BPF.o for running kernel
4. if no, compile BPF.o for running kernel

And after BTFHUB we will have (@yanivagman pls confirm if this is right):

- CO-RE phase:

1. check for /sys/kernel/btf.
2. if yes, use CO-RE binary
3. if no, look for BPF support in running kernel.
4. if yes, get BTF for running kernel from HUB
5. if BTF exists use CO-RE binary with downloaded/cached BTF file

- Legacy phase:

6. if BTF does not exist look for BPF.o for running kernel
4. if no, compile BPF.o for running kernel

> I will push a commit with this soon. @itaysk @rafaeldtinoco any thoughts?

I like it =)

grantseltzer:
@rafaeldtinoco Just what I had in mind :-) 

I will turn your comment into an issue as part of merging this PR

yanivagman:
@rafaeldtinoco LGTM, just one thing I'm not sure about:

> 6. if BTFHUB file does not exist look for BPF.o for running kernel

Do we really want to look for bpf.o only then?
One case I can think about is that for some reason, core bpf will fail at runtime (which can happen if, for example, a field of a struct was moved to some other internal struct) - then the user (or us, for debugging purposes) might want to use a precompiled bpf object (with the specific running kernel headers). This can be done, for example, by providing an environment variable (which we already support) TRACEE_BPF_FILE.
WDYT?

tracee-ebpf/Makefile
```diff
@@ -19,14 +19,9 @@ OUT_DIR ?= dist
 GO_SRC := $(shell find . -type f -name '*.go')
 OUT_BIN := $(OUT_DIR)/tracee-ebpf
 BPF_SRC := tracee/tracee.bpf.c
-
-ifndef CORE
 OUT_BPF := $(OUT_DIR)/tracee.bpf.$(subst .,_,$(KERN_RELEASE)).$(subst .,_,$(VERSION)).o
 GO_BUILD_TAGS := -tags
```
yanivagman, line 23: Do we still need this line?

tracee-ebpf/Makefile
```diff
@@ -166,12 +158,13 @@ define docker_builder_make
...
 	-w /tracee/tracee-ebpf \
-	--entrypoint make $(DOCKER_BUILDER) KERN_BLD_PATH=$(DOCKER_BUILDER_KERN_BLD) CORE="$(CORE)" KERN_SRC_PATH=$(DOCKER_BUILDER_KERN_SRC) $(1)
+	--entrypoint make $(DOCKER_BUILDER) KERN_BLD_PATH=$(DOCKER_BUILDER_KERN_BLD)
+	KERN_SRC_PATH=$(DOCKER_BUILDER_KERN_SRC) $(1)
 endef
 
 .PHONY: mostlyclean
 mostlyclean:
-	-rm -rf $(OUT_BIN) $(bpf_bundle_dir) $(OUT_BPF) $(BPF_BUNDLE)
+	-rm -rf $(OUT_BIN) $(bpf_bundle_dir) $(OUT_BPF_CORE) $(BPF_BUNDLE)
```
yanivagman, line 167: Don't we now need both OUT_BPF AND OUT_BPF_CORE here?

itaysk:
>  If there is BTF, use that embedded CORE binary. If not, look for the compiled bpf object, and if it doesn't exist attempt to build it from the embedded source. ... any thoughts?

Yes, exactly. Thanks for clarifying it and implementing!

itaysk:
@rafaeldtinoco 

> if no sysfs btf, look for BPF support in running kernel.
Isn't it late to look for BPF support in running kernel?

Also, I agree with @yanivagman that specific object file, if present, should override the embedded one

tracee-ebpf/Makefile
```diff
@@ -132,7 +127,6 @@ $(OUT_DIR)/tracee.bpf.core.%.o: $(BPF_SRC) $(LIBBPF_HEADERS) $(CMD_CLANG)
 		-DLINUX_VERSION_CODE=$(LINUX_VERSION_CODE) \
 		-target bpf \
 		-O2 -c -g -o $@ $<
-endif
 else
```
itaysk, line 130: DOCKER mode seem to be inconsistently supported for the obj files now. We need to be able to build both obj targets with docker and without. currently I think it's not possible? (haven't tried, just by reading the code)
If I'm right, would be easier to implement, and also more readable to just copy the same stanza from the other targets (e.g the OUT_BIN target) where the DOKCER variance is inside the target, not outside.

itaysk (commented):
LVGTM :)

rafaeldtinoco:
> @rafaeldtinoco
> 
> > if no sysfs btf, look for BPF support in running kernel.
> 
> Isn't it late to look for BPF support in running kernel, at this stage?

CO-RE phase:

> - check for /sys/kernel/btf.
> - if yes, use CO-RE binary
> - if no sysfs btf, look for BPF support in running kernel.

If /sys/kernel/btf exists then BPF support is already there (so we are checking it in the first item). This item is to check for old kernels that have BPF support but no BTF support, they would need the BTFHUB (bellow).

> - if BPF support, get BTF for running kernel from BTFHUB
> - if BTFHUB file exists use CO-RE binary with downloaded/cached BTF file

> Also, I agree with @yanivagman that specific object file, if present, should override the embedded one

Yes, if environment variable provided, we should use that object no matter what, I'm +1 on that as well.

grantseltzer:
Opened https://github.com/aquasecurity/tracee/issues/833

tracee-ebpf/Makefile
```diff
@@ -47,9 +44,9 @@ build: $(OUT_BIN)
 
 go_env := GOOS=linux GOARCH=$(ARCH:x86_64=amd64) CC=$(CMD_CLANG) CGO_CFLAGS="-I $(abspath $(LIBBPF_HEADERS))" CGO_LDFLAGS="$(abspath $(LIBBPF_OBJ))"
 ifndef DOCKER
-$(OUT_BIN): $(LIBBPF_HEADERS) $(LIBBPF_OBJ) $(filter-out *_test.go,$(GO_SRC)) $(BPF_BUNDLE) | $(OUT_DIR)
-	$(go_env) go build -v -o $(OUT_BIN) \
-	-ldflags "-X main.version=$(VERSION) -X main.bpf_core=$(CORE)"
+$(OUT_BIN): $(LIBBPF_HEADERS) $(LIBBPF_OBJ) $(OUT_BPF_CORE) $(OUT_BPF) $(filter-out *_test.go,$(GO_SRC)) $(BPF_BUNDLE) | $(OUT_DIR)
```
yanivagman, line 47: I think that we should only add here OUT_BPF_CORE, shouldn't we?

tracee-ebpf/main.go
```diff
@@ -1037,8 +1069,8 @@ func locateFile(file string, dirs []string) string {
 	return ""
 }
 
-// getBPFObject finds or builds ebpf object file and returns it's path
-func getBPFObject() (string, error) {
+// getBPFObjectPath finds or builds ebpf object file and returns it's path
+func getBPFObjectPath() (string, error) {
```
yanivagman, line 1073: As discussed, we want to have to TRACEE_BPF_FILE to be looked also in core mode. I think that we can call this function an exit before the binary build (or just split it into two functions)

tracee-ebpf/Makefile
itaysk: is it resolved? I don't see any change in the code

grantseltzer: Ugh, I forgot to git push...

tracee-ebpf/Makefile
grantseltzer: We can build the CO:RE object regardless of if BTF is enabled on the kernel, but that doesn't mean it can run it. So building the kernel specific one as a backup I think makes sense. WDYT?

tracee-ebpf/main.go
```diff
@@ -67,11 +67,48 @@ func main() {
...
+				bpfFile, err := getBPFObjectPath()
+				if err != nil {
+					return err
+				}
+				cfg.BPFObjPath = bpfFile
+				bpfBytes, err := ioutil.ReadFile(bpfFile)
+				if err != nil {
+					return err
+				}
+				cfg.BPFObjBytes = bpfBytes
```
yanivagman, line 109: These four lines are the same for all cases - you can take it out of the if statement

tracee-ebpf/Makefile
yanivagman: The all target (see the header of this makefile) already builds the bpf target, which compiles both of these.
So actually, you can remove them both from here

yanivagman: Or not. Keep the core one, as it is necessary by the OUT_BIN.
Just remove the OUT_BPF as it is already built by the all and bpf targets

yanivagman (approved):
LGTM!

itaysk (approved):
approving but I'm still not sure why the docker vs non-docker bpf targets are asymmetrical? in docker it's $(OUT_BPF) and in non-docker it's $(OUT_DIR)/tracee.bpf.%.o. Not the end of the world but possibly future bugs potential

grantseltzer:
@itaysk I don't fully understand it either, I don't have the strongest make knowledge. Doesn't having OUT_BPF allow for calling the ` $(OUT_DIR)/tracee.bpf.%.o` target so arguments can be passed in via $@?

PR #820, opened 2021-07-13
example: Use tracee as a library

simar7:
This PR includes  an example program that shows how to use Tracee (eBPF and Rules) as a library.

Some of the logic for the tracee library is abstracted out into: https://github.com/simar7/libtracee I'm open to bringing this into the Tracee project. Just wanted to keep it out until it was OK to do so.

Furthermore, it includes a benchmark that acts as a end to end setup+benchmark for Tracee over all the 3 different signature varieties: Golang, Rego and WASM.


yanivagman:
I like this direction.

Eventually, I think that tracee should have one main which will combine tracee-ebpf and tracee-rules.
Anyways today the most common use case is to run both of these binaries on the same machine, and it might take some time before tracee-rules will have other input sources and run independently without tracee-ebpf running as well.
Combining the two into one main has several advantages:
1. No need for serialization/deserialization between tracee-ebpf and tracee-rules
2. The user will be able to capture foresic artifacts when running tracee (which is only possible today with tracee-ebpf)
3. User friendly and more coherent

We might combine the different options with make flavors to build one main, or two different mains for tracee-ebpf and tracee-rules.

benchmark/README.md
```diff
@@ -0,0 +1,8 @@
+How to run these benchmarks:
+```
+GOOS=linux GOARCH=amd64 CC=clang CGO_CFLAGS="-I /usr/vagrant/repos/tracee/tracee-ebpf/dist/libbpf/usr/include" CGO_LDFLAGS="/usr/vagrant/repos/tracee/tracee-ebpf/dist/libbpf/libbpf.a" go test -exec 'sudo -E' -v -bench=. -benchtime=1x -tags=opa_wasm
```
grantseltzer, line 3: Perhaps we should point this to a libbpf submodule in the repo (we have one in `tracee-ebpf/3rdparty/libbpf`) where we can find the headers and build libbpf.a

grantseltzer (commented):
Adding a README for the example would also be great!

itaysk:
I'm not sure we want to promote users using tracee as a library yet, the use-case is not feature complete, and the API will likely change. We are mostly interested in the one binary for benchmarking purposes.

Issue #819, opened 2021-07-13
tracee-ebpf printer should not work if tracee-ebpf is running as a lib (aka exporting events channel)


mtcherni95:
![image](https://user-images.githubusercontent.com/39140425/125477190-c61c7f01-1ca2-4470-be94-ce29499e04b9.png)

PR #821, opened 2021-07-14
skip printing out if library mode

itaysk:
fix: #819

### DIFF
--- a/tracee-ebpf/tracee/tracee.go
+++ b/tracee-ebpf/tracee/tracee.go
@@ -895,7 +895,9 @@ func (t *Tracee) Run() error {
 	sig := make(chan os.Signal, 1)
 	signal.Notify(sig, os.Interrupt)
 	done := make(chan struct{})
-	t.printer.Preamble()
+	if t.config.ChanEvents != nil {
+		t.printer.Preamble()
+	}
 	t.eventsPerfMap.Start()
 	t.fileWrPerfMap.Start()
 	go t.processLostEvents()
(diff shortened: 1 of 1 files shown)


tracee-ebpf/tracee/tracee.go
```diff
@@ -904,8 +906,9 @@ func (t *Tracee) Run() error {
 	<-sig
 	t.eventsPerfMap.Stop()
 	t.fileWrPerfMap.Stop()
-	t.printer.Epilogue(t.stats)
-
+	if t.config.ChanEvents != nil {
+		t.printer.Epilogue(t.stats)
```
yanivagman, line 910: I think there is some important information in the epilog that a consumer of tracee as a library may want to know. This information includes, for example, the number of lost events and the event count.

grantseltzer:
What do you think about instead exposing an API where the library consumer can pass a logger interface?

tracee-ebpf/tracee/tracee.go
itaysk: Sure but this shouldn't be "printed" for the consumer right? 
We should add another API to retrieve that other information. And also I think we should take all the printing out of Tracee library. 
Anyway, this is just a bug fix which should have been included with the original commit

itaysk:
IMO the library should never print stuff, it should only return data and "main" should print it. We can pass a logger for errors though I don't think that for outputs

yanivagman (approved):

PR #825, opened 2021-07-14
perf: Improve Rego sig performance with Engine Unification

simar7:
**Scenario**
This PR includes a new benchmark that tries to demonstrate the scenario of comparing the following:

1. Combining multiple Rego signatures into one policy and one engine (new)
vs
2. One rego policy per rego engine (existing)

**Motivation:**
By combining various Rego signatures into one, we eliminate the need to launch multiple engines per policy. If we're able to apply this approach in practice the numbers show that the results are faster for one engine than launching multiples engines.

**Results:**
```
benchmark                                   iter     time/iter     bytes alloc             allocs
---------                                   ----     ---------     -----------             ------
BenchmarkEngine***/2_separate_rego_sigs-8   1000   16.59 ms/op   10498544 B/op   226044 allocs/op
BenchmarkEngine***/1_combined_rego_sig-8    1000    1.13 ms/op        241 B/op        3 allocs/op
```

**Real World CPU Performance**
Left: AIO (1 combined rego file of TRC-2, TRC-3) 
Right: 2 separate signatures (TRC-2, TRC-3) loaded separately 

image

_In this graph: Less is better_

### DIFF
--- a/tracee-rules/benchmark/signature/rego/aio.rego
+++ b/tracee-rules/benchmark/signature/rego/aio.rego
@@ -1,0 +1,79 @@
+package tracee.aio
+
+import data.tracee.helpers
+
+__rego_metadoc__ := {
+    "id": "TRC-AIO",
+    "version": "0.1.0",
+    "name": "All in one Tracee signature",
+    "description": "Combines Anti-Debugging and Code-injection",
+    "tags": ["linux", "container"],
+    "properties": {
+        "Severity": 3,
+        "MITRE ATT&CK": "Defense Evasion: Execution Guardrails",
+    }
+}
+
+# anti_debugging
+tracee_anti_debugging_tracee_selected_events[eventSelector] {
+	eventSelector := {
+		"source": "tracee",
+		"name": "ptrace"
+	}
+}
+
+tracee_anti_debugging_tracee_match {
+    input.eventName == "ptrace"
+    arg := input.args[_]
+    arg.name == "request"
+    arg.value == "PTRACE_TRACEME"
+}
+
+# code_injection
+tracee_code_injection_eventSelectors := [
+    {
+        "source": "tracee",
+        "name": "ptrace"
+    },
+    {
+        "source": "tracee",
+        "name": "security_file_open"
+    },
+    {
+        "source": "tracee",
+        "name": "process_vm_writev"
+    }
+]
+
+tracee_code_injection_tracee_selected_events[eventSelector] {
+	eventSelector := tracee_code_injection_eventSelectors[_]
+}
+
+
+tracee_code_injection_tracee_match {
+    input.eventName == "ptrace"
+    arg_value = helpers.get_tracee_argument("request")
+    arg_value == "PTRACE_POKETEXT"
+}
+
+tracee_code_injection_tracee_match = res {
+    input.eventName == "security_file_open"
+    flags = helpers.get_tracee_argument("flags")
+
+    helpers.is_file_write(flags)
+
+    pathname := helpers.get_tracee_argument("pathname")
+
+    regex.match(`/proc/(?:\d.+|self)/mem`, pathname)
+
+    res := {
+        "file flags": flags,
+        "file path": pathname,
+    }
+}
+
+tracee_code_injection_tracee_match {
+    input.eventName == "process_vm_writev"
+    dst_pid = helpers.get_tracee_argument("pid")
+    dst_pid != input.processId
+}
--- a/tracee-rules/benchmark/signature/rego/aio.rego
+++ b/tracee-rules/benchmark/signature/rego/aio.rego
@@ -1,12 +1,12 @@
 package tracee.aio
+ import data.tracee.helpers
 
-import data.tracee.helpers
 
-__rego_metadoc__ := {
-    "id": "TRC-AIO",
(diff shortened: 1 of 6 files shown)


tracee-rules/benchmark/signature/rego/aio.rego
```diff
@@ -0,0 +1,79 @@
...
+    }
+}
+
+# anti_debugging
+tracee_anti_debugging_tracee_selected_events[eventSelector] {
+	eventSelector := {
+		"source": "tracee",
+		"name": "ptrace"
+	}
+}
```
simar7, line 23: In this PR this file is hand generated to include prefixes like `tracee_anti_debugging_` prior to each rego stanza to keep them separate. Now if this was an approach we ultimately end up taking we could do something of the following (thanks to @danielpacak for brainstorming earlier):

1. Keep the same contract we have today to accept Rego signatures from users.
2. During the first run, parse all rego files (signatures) we have and combine them into one file. We could potentially save this in the /tmp/tracee directory for re-use
3. Each signatures rego policy is kept unique within the combined rego file by prefixing it with the prefixes as seen here so that we can determine which rule was triggered within this single AIO policy.
4. We would probably need some rework to figure out which rule within the combined Rego file was triggered during evaluation in the tracee-rules processing side.

simar7: What do you think about this @itaysk – to me steps 1, 2 and 3 seem doable but step 4 is a little unclear as I'm not an expert on Rego policies. Is it something we can do with Rego?

itaysk: Thanks for sharing this is very interesting. @danielpacak discussed this with me earlier this week and I told him that I don't think we need to merge the rules, we can just load them into the same rego instance which will presumably give the same effect. no rewrite is needed since the rules are namespaced already.

itaysk: I also told there are possible downsides to unifying rego instances:
1. reduced parallelism, the entire rego engine becomes a big bottleneck.
2. the entire "dispatching" mechanism which is supposed to deliver the right events to the relevant signatures becomes useless if there's one big signature that is receiving all the events. (I believe this benchmark doesn't account for that). 

An improvement suggestion: don't unify all signatures into one rego instance, but unify all signatures with the same event selectors (one per signaturesIndexMap item).

simar7:
I did some more testing and added some logic to make this transition to engine unification seamless (all existing signatures should work plug-and-play this way).

The following test runs all open source Rego signatures + 1 open source Go signature. 
image


As you can see, engine unification yields a **_10% CPU_** performance improvement.

tracee-rules/benchmark/signature/rego/aio.rego
simar7: > I also told there are possible downsides to unifying rego instances:
> 
> 1. reduced parallelism, the entire rego engine becomes a big bottleneck.
> 2. the entire "dispatching" mechanism which is supposed to deliver the right events to the relevant signatures becomes useless if there's one big signature that is receiving all the events. (I believe this benchmark doesn't account for that).
> 

It does as it can be seen here: https://github.com/aquasecurity/tracee/pull/825/#issue-690256492 - the graph is an overall CPU utilization. 

> An improvement suggestion: don't unify all signatures into one rego instance, but unify all signatures with the same event selectors (one per signaturesIndexMap item).

As discussed earlier offline this real world test effectively shows a performance improvement https://github.com/aquasecurity/tracee/pull/825/#issuecomment-881782164

tracee-rules/benchmark/signature/rego/aio.rego
```diff
@@ -0,0 +1,260 @@
+package tracee.aio
```
danielpacak, line 1: Alternative solution without changing the interface between engine and signature definition might be to create aggregating signature and leverage incremental rules and then compile it with existing signatures provided by users:

```rego
# Compile all.rego with existing Rego signatures
package tracee.all

tracee_match = res {
    # If res allowed returning multiple detections, we could
    # reuse existing rego.NewSignature constructor and pass all
    # Rego source including this one
    res := {
        "detections": detect
    }
}

# detect is an incremental rule so we can evaluate one to many
# existing signatures
detect[resp] {
	resp := {
		"name": data.tracee.TRC_2.__rego_metadoc__.name,
		"matches": data.tracee.TRC_2.tracee_match
	}
}

detect[resp] {
	resp := {
		"name": data.tracee.TRC_3.__rego_metadoc__.name,
		"matches": data.tracee.TRC_3.tracee_match
	}
}

# Define detect[res] for all signatures that you want to evaluate
```

You could instantiate it as any existing Rego signature without commenting out code in other places:

```go
regosig.NewRegoSignature(helpersRego, codeInjectionRego, antiDebuggingPtracemeRego, allRego)
```

However, this would require logic to parse result set with zero to many detections in the `OnEvent` callback.

grantseltzer:
@simar7 Is there anything blocking further review of this PR?

simar7:
> @simar7 Is there anything blocking further review of this PR?

Nothing blocking but we've taken a different direction with this. One drawback of this approach is that we cannot determine which rule was triggered as all rules are combined into one. Besides that the new AIO rule indexer approach has superseded this. I'll close this one.

PR #827, opened 2021-07-19
Add args to sched process exec

yanivagman:


grantseltzer:
I'm getting a build error locally with your branch

```
tracee/tracee.bpf.c:689:42: error: static declaration of 'get_task_mm' follows non-static declaration
static __always_inline struct mm_struct* get_task_mm(struct task_struct *task)
                                         ^
/lib/modules/5.12.15-300.fc34.x86_64/source/include/linux/sched/mm.h:88:26: note: previous declaration is here
extern struct mm_struct *get_task_mm(struct task_struct *task);
                         ^
```

yanivagman:
This means that your kernel already has this function defined (mine don't), and it's good that you catched this!
Can you please rename this function and check again?

itaysk:
@yanivagman what do you mean rename this function? is this something some users will be required to do?

yanivagman:
@itaysk I just asked @grantseltzer to check if this works for him. I'll update the pr.

yanivagman:
Just noticed that in ubuntu there are verifier issues.
Converted to draft

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2128,14 +2245,54 @@ int tracepoint__sched__sched_process_exec(struct bpf_raw_tracepoint_args *ctx)
...
 
+    // Note: Starting from kernel 5.9, there are two new interesting fields in bprm that we should consider adding:
+    // 1. struct file *executable - which can be used to get the executable name passed to an interpreter
+    // 2. fdpath - generated filename for execveat (after resolving dirfd)
+
     save_str_to_buf(submit_p, (void *)filename, DEC_ARG(0, *tags));
-    save_to_submit_buf(submit_p, &invoked_from_kernel, sizeof(int), INT_T, DEC_ARG(1, *tags));
+    save_str_to_buf(submit_p, (void *)&string_p->buf[*off], DEC_ARG(1, *tags));
+    save_args_str_arr_to_buf(submit_p, (void *)arg_start, (void *)arg_end, argc, DEC_ARG(2, *tags));
+    //save_args_str_arr_to_buf(submit_p, (void *)env_start, (void *)env_end, envc, DEC_ARG(3, *tags));
```
grantseltzer, line 2292: why is this commented out?

grantseltzer:
No issue building after you renamed it. Also can you tell me a little more about what the difference is between sched_process_exec and regular execve?

tracee-ebpf/tracee/tracee.bpf.c
yanivagman: See the comment above for "too many instructions" on kernels before 5.2. 
I would like to add it in the future (already tested and working) for kernels > 5.2

yanivagman:
> No issue building after you renamed it. Also can you tell me a little more about what the difference is between sched_process_exec and regular execve?

This event is now a combination of execve and security_bprm_check and enjoys the advantages of both:
1. Having user given binary + real path (not available in execve)
2. Avoid TOCTTOU attacks by not using user pointers, but kernel data (what will be really used by the kernel) - both for binary name and arguments
3. Having device id and inode number as well
4. It is a tracepoint (unlike security_bprm_check)

tracee-ebpf/tracee/tracee.bpf.c
grantseltzer: Can you just add a compile time macro check for kernel version?

yanivagman: Yes, we can add such a check, but co-re...
Let's do it in a future PR

grantseltzer (approved):
**lgtm**

PR #829, opened 2021-07-20
tracee-rules: evaluate parsed input with OPA

danielpacak:
With OPA Go SDK we can significantly improve performance by passing
rego.EvalParsedInput instead of rego.EvalInput option to rego.Eval()
method.

Co-authored-by: Simar <simar@linux.com>

### DIFF
--- a/tracee-rules/engine/engine.go
+++ b/tracee-rules/engine/engine.go
@@ -5,10 +5,13 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"reflect"
+	"strings"
 	"sync"
 
 	tracee "github.com/aquasecurity/tracee/tracee-ebpf/external"
 	"github.com/aquasecurity/tracee/tracee-rules/types"
+	"github.com/open-policy-agent/opa/ast"
 )
 
 // Engine is a rule-engine that can process events coming from a set of input sources against a set of loaded signatures, and report the signatures' findings
@@ -152,17 +155,34 @@ func (engine *Engine) consumeSources(done <-chan bool) {
 					engine.signaturesMutex.RUnlock()
 					continue
 				}
+
+				pe, err := ToParsedEvent(traceeEvt)
+				if err != nil {
+					engine.logger.Printf("error converting tracee event to OPA ast.Value: %v", err)
+					engine.signaturesMutex.RUnlock()
+					continue
+				}
+
 				for _, s := range engine.signaturesIndex[types.SignatureEventSelector{Source: "tracee", Name: traceeEvt.EventName}] {
-					engine.signatures[s] <- event
+					engine.dispatchEvent(s, pe)
 				}
 				for _, s := range engine.signaturesIndex[types.SignatureEventSelector{Source: "tracee", Name: "*"}] {
-					engine.signatures[s] <- event
+					engine.dispatchEvent(s, pe)
 				}
 				engine.signaturesMutex.RUnlock()
 			}
 		case <-done:
 			return
 		}
+	}
+}
+
+func (engine *Engine) dispatchEvent(s types.Signature, pe ParsedEvent) {
+	switch {
+	case strings.Contains(reflect.TypeOf(s).String(), "rego"):
+		engine.signatures[s] <- pe
+	default:
+		engine.signatures[s] <- pe.Event
 	}
 }
 
@@ -245,4 +265,28 @@ func (engine *Engine) UnloadSignature(signatureId string) error {
 		}
 	}
 	return nil
+}
+
+// ParsedEvent holds the original tracee.Event and its OPA ast.Value representation.
+type ParsedEvent struct {
+	Event tracee.Event
+	Value ast.Value
+}
+
+// ToParsedEvent enhances tracee.Event with OPA ast.Value. This is mainly used
+// for performance optimization to avoid parsing tracee.Event multiple times.
+func ToParsedEvent(e tracee.Event) (ParsedEvent, error) {
+	u, err := e.ToUnstructured()
+	if err != nil {
+		return ParsedEvent{}, fmt.Errorf("unstructuring event: %w", err)
+	}
+	// TODO In OPA >= v0.30.0 we can try passing tracee.Event directly to get rid of ToUnstructured call.
+	v, err := ast.InterfaceToValue(u)
+	if err != nil {
+		return ParsedEvent{}, fmt.Errorf("converting unstructured event to OPA ast.Value: %w", err)
+	}
+	return ParsedEvent{
+		Event: e,
+		Value: v,
+	}, nil
 }
--- a/tracee-rules/engine/engine_test.go
+++ b/tracee-rules/engine/engine_test.go
@@ -6,13 +6,14 @@ import (
 	"testing"
 	"time"
 
+	"github.com/aquasecurity/tracee/tracee-ebpf/external"
 	tracee "github.com/aquasecurity/tracee/tracee-ebpf/external"
 	"github.com/aquasecurity/tracee/tracee-rules/types"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
-type fakeSignature struct {
+type regoFakeSignature struct {
 	getMetadata       func() (types.SignatureMetadata, error)
 	getSelectedEvents func() ([]types.SignatureEventSelector, error)
 	init              func(types.SignatureHandler) error
@@ -20,7 +21,7 @@ type fakeSignature struct {
 	onSignal          func(signal types.Signal) error
 }
 
-func (fs fakeSignature) GetMetadata() (types.SignatureMetadata, error) {
+func (fs regoFakeSignature) GetMetadata() (types.SignatureMetadata, error) {
 	if fs.getMetadata != nil {
 		return fs.getMetadata()
 	}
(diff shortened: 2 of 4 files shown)


danielpacak:
tracee_opa_eval_parsed_input

itaysk:
@danielpacak you mention that this PR is just partially implementing the optimization, do you want to merge it, or it's draft until you add the part in consumeSources? If you want to add one by one, then I think the `toParsedEvent` and related functions should be taken out of test and made available to engine? (Also it has some overlap with #830 which I wasn't clear how you plan to reconcile). I'm approving anyway, just please share your plan, and possible open an issue to track it if you think it's necessary

itaysk (approved):

danielpacak:
> @danielpacak you mention that this PR is just partially implementing the optimization, do you want to merge it, or it's draft until you add the part in consumeSources? If you want to add one by one, then I think the `toParsedEvent` and related functions should be taken out of test and made available to engine? (Also it has some overlap with #830 which I wasn't clear how you plan to reconcile). I'm approving anyway, just please share your plan, and possible open an issue to track it if you think it's necessary

Hey @itaysk Sorry for the confusion. I believe we should merge #830 first because we do need ToUnstructured method to do any optimisation. Then I'll rebase this one and add missing code in consumeSources. Does it make sense?

tracee-rules/engine/engine.go
```diff
@@ -152,11 +154,31 @@ func (engine *Engine) consumeSources(done <-chan bool) {
 					engine.signaturesMutex.RUnlock()
 					continue
 				}
+
+				u, err := traceeEvt.ToUnstructured()
+				if err != nil {
+					engine.logger.Printf("unstructuring tracee event: %v", err)
```
simar7, line 160: While I'm not sure of the original intentions of the existing code that has a RWLock for each signature, I believe we will need to release the read lock in the failure branches in this code.


```suggestion
					engine.logger.Printf("unstructuring tracee event: %v", err)
					engine.signaturesMutex.RUnlock()
```

tracee-rules/engine/engine.go
```diff
@@ -152,11 +154,31 @@ func (engine *Engine) consumeSources(done <-chan bool) {
...
+					continue
+				}
+
+				// TODO In OPA v0.30.0 we can pass tracee.Event and get rid of ToUnstructured call.
+				value, err := ast.InterfaceToValue(u)
+				if err != nil {
+					engine.logger.Printf("converting unstructured tracee event to OPA value: %v", err)
```
simar7, line 167: ditto here for RUnlock()

tracee-rules/engine/engine.go
```diff
@@ -152,11 +154,31 @@ func (engine *Engine) consumeSources(done <-chan bool) {
 					engine.signaturesMutex.RUnlock()
 					continue
 				}
+
+				u, err := traceeEvt.ToUnstructured()
+				if err != nil {
+					engine.logger.Printf("unstructuring tracee event: %v", err)
+					continue
+				}
```
simar7, line 162: Since this will be throwaway code post the update to v0.30.0, I think we can forgo writing a test here.

tracee-rules/engine/engine.go
simar7: Also a good read since we're talking about performance: https://github.com/golang/go/issues/17973

tracee-rules/engine/engine.go
```diff
@@ -166,6 +189,15 @@ func (engine *Engine) consumeSources(done <-chan bool) {
 	}
 }
 
+func (engine *Engine) dispatchEvent(s types.Signature, pe regosig.ParsedEvent, event types.Event) {
+	switch {
+	case strings.Contains(reflect.TypeOf(s).String(), "rego"):
```
simar7, line 194: See here for a quick benchmark: https://gist.github.com/simar7/7f671bf4ce4bd685d4c9346f44c1e9c1

simar7:
Observation:

While this is a step in the right direction, we still have more work to do. This is how the flamegraph looks today:
image

And this is how it looks after implementing this change:
image

As you can see the JSON round tripper was replaced with our InterfaceToValue call. While quantitatively this is a minor improvement, three additional tracks to work on can be identified as follows:

1. 20% of total CPU:  engine.signatureStart, primarily a track to optimize rego prepared queries and their eval.
2. 25% of total CPU: runtime.gcBgMarkWorker, the goroutine over head track. 
2. 20% of total CPU: engine.consumeSources, event handling track.

tracee-rules/engine/engine.go
danielpacak: Done

tracee-rules/engine/engine.go
danielpacak: 💯

tracee-rules/engine/engine.go
danielpacak: We can also look into contention / block / mutex profiles

tracee-rules/main.go
```diff
@@ -27,6 +28,9 @@ func (realClock) Now() time.Time {
 }
 
 func main() {
+	// TODO Introduce --pprof flag?
+	defer profile.Start(profile.CPUProfile, profile.ProfilePath("."), profile.NoShutdownHook).Stop()
```
danielpacak, line 32: I think we need a flag --pprof in tracee-rules to conditionally enable profiling. Exposing /debug/pprof is also very useful to look into other things such as memory, coroutines or mutex

simar7: I'd say lets do it in a different PR

danielpacak: Agree. Reverted that

simar7 (approved):

tracee-rules/engine/engine.go
itaysk: @danielpacak @simar7 why are we not using the type assert/switch here? I remember Daniel mentioned something the doesn't work but isn't this what we want? https://play.golang.org/p/en8jJCHWiqw

itaysk:
thanks for working on merging this. do we have a sense on how much this improves by? I believe @simar7 told my ~10%, is that true?

danielpacak:
> thanks for working on merging this. do we have a sense on how much this improves by? I believe @simar7 told my ~10%, is that true? @danielpacak

@itaysk Here are sample [results](https://github.com/danielpacak/opa-herculean/tree/6397f3aa6f41aaee9e49456916e7ae89e42781dc#3-evaluate-raw-input-versus-evaluate-parsed-input) from the benchmark of rego.Eval input versus rego.EvalParsedInput for 6 (open source) and  71 (closed source) signatures with 2 input events that trigger findings. Evaluating a prepared query seems to be 3-4 times faster.  However, overall performance of the app depends on the workload, i.e. the stream of input events and how often an event is selected for evaluation (signatureIndex impl.).

tracee-rules/engine/engine.go
danielpacak: I think our case is more like https://play.golang.org/p/wETpR604pdj

@itaysk I tried one more time using switch type and the following snippet seems to be working:

```go
func (engine *Engine) dispatchEvent(s types.Signature, pe ParsedEvent) {
	switch s.(type) {
	case *regosig.RegoSignature: // Make sure to use pointer because we use pointer receiver for the RegoSignature implementation
		engine.signatures[s] <- pe
	default:
		engine.signatures[s] <- pe.Event
	}
}
```

However, it will introduce the cycle because of the engine.ParsedEvent type. We could fix that by moving the ParsedEvent to a shared package, maybe types?

@simar7 Could you explain what was the advantage of using reflect.TypeOf(s) here?

simar7: There is no advantage in this case. The only problem there is that we lose testability if we hardcode the type to `*regosig.RegoSignature`. 

In the test for `consumeSources` we use a fake type called `regoFakeSignature` https://github.com/aquasecurity/tracee/blob/6df0969653c4567f7e6dabca5a77b639c042fb83/tracee-rules/engine/engine_test.go#L16 - When evaluating the following code, this will result in taking the `default` path instead which is not the desired behaviour. 
```
func (engine *Engine) dispatchEvent(s types.Signature, pe ParsedEvent) {
	switch s.(type) {
	case *regosig.RegoSignature: // Note this isn't the same as regoFakeSignature
		engine.signatures[s] <- pe
	default:
		engine.signatures[s] <- pe.Event
	}
}
```

The other approach if we really wanted to use switch type assert would be to manually list each gosig as a case (since they are all defined types) and default for each regosig. So something like the following:

```func (engine *Engine) dispatchEvent(s types.Signature, pe ParsedEvent) {
	switch s.(type) {
	case *stdioOverSocket, *antiDebugging: 
		engine.signatures[s] <- pe.Event
	default:
		engine.signatures[s] <- pe
	}
}
```
This doesn't scale well as it would require a change to this function every time a new Go signature is added/used.

For the above reasons I went with a `reflect.TypeOf(s)` instead. If I've missed something or if there is a better way to do it, happy to understand and discuss.

itaysk: if the only holdback from doing it using type assertion (assuming we agree it's better), is just the test, I'd sacrifice it and not he production code. But I think this is a symptom of a bigger problem, the fact that engine is knowing which concrete signature types it's invoking is breaking the entire premise of the Signature interface. I don't have a better suggestion yet, just wanted to bring this point to attention. @simar7  @danielpacak

PR #840, opened 2021-07-26
tracee-ebpf: update external module

yanivagman:
PR #837 changed the external module files.
Update go.mod to point to the updated module.

### DIFF
--- a/tracee-ebpf/external/external_test.go
+++ b/tracee-ebpf/external/external_test.go
@@ -21,12 +21,12 @@ func TestEventUnmarshalJSON(t *testing.T) {
 
 	testCases := []testCase{
 		{
-			json: `{"timestamp":260182.49532,"processId":12434,"threadId":12434,"parentprocessid":23921,
+			json: `{"timestamp":26018249532,"processId":12434,"threadId":12434,"parentprocessid":23921,
 			"hostprocessid":12434,"hostthreadid":12434,"hostparentprocessid":23921,"userid":1000,"mountnamespace":4026531840,
 			"pidnamespace":4026531836,"processname":"strace","hostname":"ubuntu","eventid":"101","eventname":"ptrace",
 			"argsnum":4,"returnvalue":0,"args":[{"name":"request","type":"long","value":"ptrace_seize"},
 			{"name":"pid","type":"pid_t","value":12435},{"name":"addr","type":"void*","value":"0x0"},{"name":"data","type":"void*","value":"0x7f6f1eb44b83"}]}`,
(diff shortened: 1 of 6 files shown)


itaysk (commented):
also need to update tracee-rules

itaysk:
@danielpacak I think we can also remove this? https://github.com/aquasecurity/tracee/blob/443955e3c087f8e7652420c9d947cf5fe3c51756/tracee-ebpf/external/external.go#L64:L67

danielpacak:
> @danielpacak I think we can also remove this? https://github.com/aquasecurity/tracee/blob/443955e3c087f8e7652420c9d947cf5fe3c51756/tracee-ebpf/external/external.go#L64:L67

You are right Itay. Now we can just

`"timestamp": json.Number(strconv.Itoa(e.Timestamp))`

instead of

`"timestamp": json.Marshal(e.Timestamp)`

yanivagman:
@danielpacak I can do this update here, but then again we will need to update the external module in a different PR.
So maybe do this change now and I'll rebase?

danielpacak:
> @danielpacak I can do this update here, but then again we will need to update the external module in a different PR (go.mod will have to be changed in tracee-ebpf and tracee-rules as we change the external.go of the external module).
> So maybe do this change now and I'll rebase?

Yeah, let's do it in scope of this PR. Someone will have to that anyway

yanivagman:
Ok, I think I'll open a different PR for this change (It will be funny to update external module twice)

yanivagman:
Opened #842.
Tests will not pass there, but let's merge and I'll rebase this PR on it, which will also fix the build

yanivagman:
Ok, so surprisingly after rebasing and doing "go get" nothing changed in go.mod (not sure why), so we can merge this one now

danielpacak:
> Ok, so surprisingly after rebasing and doing "go get" nothing changed in go.mod (not sure why), so we can merge this one now

Shouldn't we point to the latest commit in bot tracee-ebf/go.mod and tracee-rules/go.mod ?

```
-       github.com/aquasecurity/tracee/tracee-ebpf/external v0.0.0-20210727054623-f17c1d161b64
+       github.com/aquasecurity/tracee/tracee-ebpf/external v0.0.0-20210727091827-bbe411a2a167
```

I also realised that external_test.go if failing due to change in Timestamp type from float64 to int.

yanivagman:
> Shouldn't we point to the latest commit in both tracee-ebf/go.mod and tracee-rules/go.mod ?

What do you mean? In this PR they both point to f17c1d161b64

yanivagman:
> I also realised that external_test.go if failing due to change in Timestamp type from float64 to int.

We should add this test to the pipeline

danielpacak:
> > Shouldn't we point to the latest commit in both tracee-ebf/go.mod and tracee-rules/go.mod ?
> 
> What do you mean? In this PR they both point to [f17c1d1](https://github.com/aquasecurity/tracee/commit/f17c1d161b640966f77e7d9ee164ee535e141987)

Yes, but the latest commit on the main branch is bbe411a not f17c1d1. Unless I took it all wrong 🤯

yanivagman:
> Yes, but the latest commit on the main branch is bbe411a not f17c1d1. Unless I took it all wrong exploding_head

You are right. For some reason, it didn't pull the latest changes for me when I did `go get`. Now everything seems to be aligned.

danielpacak (approved):
LGMT!

Issue #815, opened 2021-07-13
Can't run tracee - getting error about long argument list


michaelkatch:
Hello,

I'm trying to run tracee-ebpf with no specific arguments on the following kernel :
*5.3.0-1023-aws #25~18.04.1-Ubuntu SMP Fri Jun 5 15:18:30 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux*

This is the error I get 
![MicrosoftTeams-image (3)](https://user-images.githubusercontent.com/70207874/125458506-c57884a1-1cde-4b72-8ff6-e9a66ecd650f.png)

Thank you

michaelkatch:
Hey Grant,

I'm compiling using Docker with the command "make DOCKER=1" so I don't have bpftool binary on the machine.
I ran the commands you wrote and compiled again via Docke and still the same error.

grantseltzer:
Could you paste the output of the first two commands in my last comment? Essentially what i'm looking for is to make sure the kernel you're running supports tracee/bpf.

grantseltzer:
```
...
last_idx 2807 first_idx 2799
regs=4 stack=0 before 2806: (b7) r2 = 16
2808: (61) r2 = *(u32 *)(r10 -236)
; len = (d_name.len+1) & (MAX_STRING_SIZE-1);
2809: (07) r2 += 1
; len = (d_name.len+1) & (MAX_STRING_SIZE-1);
2810: (57) r2 &= 4095
; if (off <= buf_off) { // verify no wrap occured
2811: (bf) r1 = r9
2812: (67) r1 <<= 32
2813: (77) r1 >>= 32
; if (off <= buf_off) { // verify no wrap occured
2814: (2d) if r2 > r1 goto pc+8132
[...]

PR #841, opened 2021-07-27
tracee-ebpf: Adjust MAX_PATH_COMPONENTS limit for kernels >= 5.2

danielpacak:
Resolves: #815


yanivagman (approved):
Thanks for finding the cause for this issue and fixing it @danielpacak @grantseltzer

PR #843, opened 2021-07-27
tracee-ebpf: add inode and dev to magic_write event

yanivagman:


rafaeldtinoco (approved):
Looks good to me. dev_t and unsigned long being added as extra 2 args for event MAGIC_WRITE.

PR #848, opened 2021-07-28
Feature/signature filtering infrastructure

AlonZivony:
Made an infrastructure to filtering the signatures that each event will be passed to, in order of reducing real-time resources use of tracee-rules.
The purpose is to make it easy to create new filters in the future, and by that upgrading tracee-rules efficiency.

### DIFF
--- a/tracee-rules/engine/engine.go
+++ b/tracee-rules/engine/engine.go
@@ -16,11 +16,12 @@ import (
 type Engine struct {
 	logger          log.Logger
 	signatures      map[types.Signature]chan types.Event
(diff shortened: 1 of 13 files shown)


grantseltzer:
@AlonZivony This was never marked as not a draft, is there anything blocking this? Is it still something you'd like to push through?

AlonZivony:
So this architecture is still option that could happen in the future.
After testing the performance of it it was decided that the SOLIDity of it does not worth its cost in this time being.
In the future if more filters will be needed maybe we will return to it, because the current architecture does not support it.

PR #850, opened 2021-07-28
Fix 4th syscall param

### LINKED ISSUE #849: mount syscall flags arg field data error
Hi, I used tracee to collect mount system calls, and found that the `mountflags` field is wrong.

For example:

```
TIME(s)        UTS_NAME         MNT_NS       PID_NS       UID    COMM             PID     TID     PPID    RET              EVENT                ARGS
33898.711492   ubuntu           4026531840   4026531836   0      dockerd          1086    1263    1       0                mount                source: overlay, target: /var/lib/docker/overlay2/3c2e1e237a9b4811ad810070500dfa367ef6bc1921511a542f1ff0057846274f-init/merged, filesystemtype: overlay, mountflags: 94364462432266, data: 0xC0008121A0
```

The flag is `94364462432266`.

However, I can confirm with LTTng that the real flag is `0`.

Other fields like source, target, filesystemtype and data are correct.

yanivagman:
fix #849

### DIFF
--- a/tracee-ebpf/tracee/tracee.bpf.c
+++ b/tracee-ebpf/tracee/tracee.bpf.c
@@ -2090,7 +2090,13 @@ if (CONFIG_ARCH_HAS_SYSCALL_WRAPPER) {
         args_tmp.args[0] = PT_REGS_PARM1(&regs);
         args_tmp.args[1] = PT_REGS_PARM2(&regs);
         args_tmp.args[2] = PT_REGS_PARM3(&regs);
+#if defined(bpf_target_x86)
+        // In x86-64, r10 is used instead of rcx to pass the fourth parameter of a syscall
+        // see also: https://stackoverflow.com/questions/21322100/linux-x64-why-does-r10-come-before-r8-and-r9-in-syscalls
+        args_tmp.args[3] = regs.r10;
+#else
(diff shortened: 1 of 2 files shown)


tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2090,7 +2090,13 @@ if (CONFIG_ARCH_HAS_SYSCALL_WRAPPER) {
         args_tmp.args[0] = PT_REGS_PARM1(&regs);
         args_tmp.args[1] = PT_REGS_PARM2(&regs);
         args_tmp.args[2] = PT_REGS_PARM3(&regs);
+#if defined(bpf_target_x86)
+        // In x86-64, r10 is used instead of rcx to pass the fourth parameter of a syscall
+        // see also: https://stackoverflow.com/questions/21322100/linux-x64-why-does-r10-come-before-r8-and-r9-in-syscalls
```
rafaeldtinoco, line 2095: Had to check in amd64 ABI before opening the url... and yep, convention for static chains is this indeed:

image

rafaeldtinoco (approved):
I'm approving because change is correct... unsure if you want reporter to make sure it fixes the issue before merging, feel free.

yanivagman:
@rafaeldtinoco I validated this locally and it fixes the issue, thanks for double checking

PR #853, opened 2021-07-29
tracee-ebpf: extend magic_write bytes

yanivagman:
Extending the number of bytes sent by the magic_write event. This is required for the analysis of ELF headers.


rafaeldtinoco (approved):
LGTM.

PR #858, opened 2021-07-29
tracee-ebpf: add SIGTERM support

josedonizetti:
Handle SIGTERM  to display the epilogue message, instead of abruptly stop. 

eg:
```
End of events stream
Stats: {eventCounter:7531 errorCounter:0 lostEvCounter:0 lostWrCounter:0 lostNtCounter:0}
```


rafaeldtinoco (approved):
Simple change, tested (for the sake of clean conscience), and accepting.

PR #859, opened 2021-07-29
tracee-ebpf: refactor to avoid two strings.Split

josedonizetti:
Refactor to avoid doing `strings.Split(k, ".")` twice. Let me know if you okay with simple refactorings like this.

### DIFF
--- a/tracee-ebpf/tracee/tracee.go
+++ b/tracee-ebpf/tracee/tracee.go
@@ -1399,9 +1399,9 @@ func (t *Tracee) updateProfile(sourceFilePath string, executionTs uint64) {
 
 func (t *Tracee) updateFileSHA() {
 	for k, v := range t.profiledFiles {
-		pathPrefix := strings.Split(k, ".")[0]
-		exeName := strings.Split(strings.Split(k, ".")[1], ":")[0]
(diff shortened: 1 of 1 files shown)


tracee-ebpf/tracee/tracee.go
```diff
@@ -1390,8 +1390,9 @@ func (t *Tracee) updateProfile(sourceFilePath string, executionTs uint64) {
 
 func (t *Tracee) updateFileSHA() {
 	for k, v := range t.profiledFiles {
-		pathPrefix := strings.Split(k, ".")[0]
-		exeName := strings.Split(strings.Split(k, ".")[1], ":")[0]
+		s := strings.Split(k, ".")
+		pathPrefix := s[0]
```
grantseltzer, line 1394: It may get optimized out anyway, but we don't need to even create 'pathPrefix', we can just use 's[0]'.

grantseltzer (commented):
Hi @josedonizetti, thanks for catching this! Just the one comment on it. 

And we're certainly okay with simple refactors, it's much appreciated!

grantseltzer (approved):
LGTM!

grantseltzer:
Thank you @josedonizetti!

PR #861, opened 2021-08-02
use tcp_connect kprobe to get tcp handshake packets

roikol:
using kprobe tcp_connect, in order to capture tcp SYN and SYN-ACK packets. 

this internal kernel function is a good place to identify outgoing tcp connections:
- this function is called both for ipv4 and ipv6. 
- this function is called after the kernel assigns a local port to the socket.
- this function is called before the first SYN packet is sent.

we no longer have use of the tcp states TCP_SYN_SENT and TCP_ESTABLISHED in the inet_sock_set_state tracepoint - this is because these states were only used to capture outgoing tcp connections.
TCP_LISTEN is still needed because it is used to capture incoming tcp connections.

TCP_CLOSE is now used for both outgoing and incoming tcp connections, but we have to update sock_ctx_map in tcp_connect, so we can remove this network flow from network_map on TCP_CLOSE state.

### DIFF
--- a/tracee-ebpf/tracee/tracee.bpf.c
+++ b/tracee-ebpf/tracee/tracee.bpf.c
@@ -198,6 +198,7 @@ extern bool CONFIG_ARCH_HAS_SYSCALL_WRAPPER __kconfig;
 #define DEBUG_NET_UDP_DESTROY_SOCK      4
 #define DEBUG_NET_UDPV6_DESTROY_SOCK    5
 #define DEBUG_NET_INET_SOCK_SET_STATE   6
+#define DEBUG_NET_TCP_CONNECT           7
 
 #define CONFIG_SHOW_SYSCALL         1
 #define CONFIG_EXEC_ENV             2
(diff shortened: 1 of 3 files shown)


tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -3391,19 +3392,7 @@ int tracepoint__inet_sock_set_state(struct bpf_raw_tracepoint_args *ctx)
...
-        net_ctx_ext.host_tid = bpf_get_current_pid_tgid();
-        bpf_get_current_comm(&net_ctx_ext.comm, sizeof(net_ctx_ext.comm));
-        net_ctx_ext.local_port = connect_id.port;
-        bpf_map_update_elem(&sock_ctx_map, &sk, &net_ctx_ext, BPF_ANY);
-        break;
     case TCP_LISTEN:
-    case TCP_ESTABLISHED:
         // Ideally, we would update the network map only in 'TCP_ESTABLISHED' state.
         // However, in the case of a local server program and a local client program, which communicate via the 'lo' interface,
         // the change to the 'TCP_ESTABLISHED' state of the server's socket doesn't happen in the (process) context of the server.
```
yanivagman, line 3398: This comment can also be removed

tracee-ebpf/tracee/tracee.go
```diff
@@ -1821,6 +1830,19 @@ func (t *Tracee) processNetEvents() {
...
+						comm,
+						hostTid,
+						netaddr.IPFrom16(pkt.LocalIP),
+						pkt.LocalPort,
+						netaddr.IPFrom16(pkt.RemoteIP),
+						pkt.RemotePort,
+						pkt.Protocol,
+						pkt.TcpOldState,
+						pkt.TcpNewState,
+						pkt.SockPtr)
```
yanivagman, line 1845: These lines are not relevant for this debug message.
You can use the same arguemtns as in DebugNetUdpDestroySock instead

roikol: you're right, fixed it.

yanivagman (approved):

Issue #862, opened 2021-08-02
containerid doesn't work for existing containers


itaysk:
![image](https://user-images.githubusercontent.com/7395852/127900726-2e2b5d61-2968-498f-94c8-8cc3002f8d1a.png)

Initial traces were from a previously started container, showing no containerid.

rafaeldtinoco:
We currently have a "pid_to_cont_id_map" with container_id <-> pid mappings. Whenever a pid is traced, its container_id comes from reading the BPF_HASH `key=pid(u32) -> value=container_id`.

This map only gets populated in the following tracepoints:

1. raw_tracepoint/sched_process_fork (when process is created)
2. raw_tracepoint/cgroup_attach_task (when task is attached to cgroup)

And gets deleted in:

1. raw_tracepoint/sched_process_exit (when process exits)

In order to have existing container_ids in the BPF MAP we can populate that map from tracee.go, during initialization, with the existing containers by the time tracee-ebpf is initialized (from populateBPFMaps() or similar).

Working on it...

itaysk:
> we can populate that map from tracee.go, during initialization

👍

yanivagman:
This is not a bug, but a missing feature as I noted here:
https://github.com/aquasecurity/tracee/pull/749#issuecomment-861373065

rafaeldtinoco:
@yanivagman I'll wrap lsns (from util-linux) output and populate the BPF_HASH pid_to_cont_id_map during startup if its okay for you. Writing our own tool to traverse procfs with all peculiarities dealt by lsns.c (from util-linux/sys-utils/lsns.c) seems unnecessary for me and the tool already provides JSON output with columns chooser and so on (to list all existing ns).

yanivagman:
@rafaeldtinoco I'm actually unfamiliar with this tool. If it only iterates procfs, how do you plan to get the container id with it?

rafaeldtinoco:
![image](https://user-images.githubusercontent.com/7395852/127916435-f5e9f825-65a3-4641-b406-731d87a848f1.png)

yanivagman:
But the map is pid to container id. To populate it, you will need to know the container id of the existing pids.

yanivagman:
The method I used to get the container ids in runtime is to use the cgroup (dentry) name, which contains the container id in it (initialized as such by the container runtimes).
So you can take a similar approach at init time - traverse cgroupfs, and for each non-host cgroup, extract the pids and container ids out of it.

[...]

Issue #814, opened 2021-07-12
Change libbpfgo map methods to use unsafe.Pointer


grantseltzer:
In https://github.com/aquasecurity/libbpfgo/pull/32 the `Update`, `DeleteKey`, and `GetValue` methods of BPFMap were changed to use unsafe.Pointer's instead of `interface{}`s.

grantseltzer:
@itaysk I want to make sure this is in the next release of tracee. This is a blocker from upgrading tracee to new versions of libbpfgo

rafaeldtinoco:
![image](https://user-images.githubusercontent.com/7395852/127369714-fb11909f-17fd-4185-9c0b-da405f988545.png)

I had to revert those changes locally to keep local sources of tracee and libbpfgo (so I can develop both together). I started making the changes but saw they would require some changes in setUintFilter (for UID, PID, ..., and anything using consts passed as arguments to setXXXFilter) then dropped it, and then Grant told me about this issue.

rafaeldtinoco:
Plan here is:

- pull request #864 finished (Fixes: #862), add more commits:
- tracee needs to use latest libbpfgo (Fixes: #894)
- change tracee to use new libbpfgo map methods (Fixes: #814)

I'm working on it and will pull #864 altogether soonish.

rafaeldtinoco:
For this change we will have to do some modifications in the way we rely in 'costs' for configuration sets into maps. 

Example:

```go
const (
	filterIn  uint8 = 1
	filterOut uint8 = 2
)
```

```go
	configFilteru32 := uint32(configFilter)
	configFilterIn := uint32(filterIn)
	configFilterOut := uint32(filterOut)

	bpfConfigMap, err := t.bpfModule.GetMap("config_map") // u32, u32
	if err != nil {
		return err
	}
	if len(filter.Equal) > 0 && len(filter.NotEqual) == 0 {
		bpfConfigMap.Update(unsafe.Pointer(&configFilteru32), unsafe.Pointer(&configFilterIn))
	} else {
		bpfConfigMap.Update(unsafe.Pointer(&configFilteru32), unsafe.Pointer(&configFilterOut))
	}
```

Instead of creating local variables every time we need to call map.Update() function.. we can, instead of relying in consts, rely in initialized object variables with the same value. This way they will always have stack addresses to unsafe.Pointer() point at.

[...]

PR #864, opened 2021-08-03
tracee.go: initialize pid_to_cont_id_map during startup

rafaeldtinoco:
Fixes: #862

Initialize pid_to_cont_id_map during tracee startup so that already
running containers can have their container_id resolved when being
traced.

Instead of traversing cgroups filesystem, like tracee eBPF code does, I
have preferred to rely in wrapped 'docker cli' commands to discover
container ids and tasks running in each active container. This made the
code cleaner and easier to be maintained.

Note: There is a small window of opportunity for a container to be
started while this logic is running. I could have made 'Containers'
logic to run in a go coroutine but that would also create another race
window: userland adding a pid to the map that has already been removed
by the - now active - trace event handler.

### DIFF
--- a/tracee-ebpf/tracee/containers.go
+++ b/tracee-ebpf/tracee/containers.go
@@ -1,0 +1,180 @@
+package tracee
+
+import (
+	"bufio"
+	"fmt"
+	"io/fs"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+)
+
+// Containers contain information about host running containers in the host.
+type Containers struct {
+	cgroupv1   bool
+	cgroupv1mp string
+	cgroupv2   bool
+	cgroupv2mp string
+	mapContIds map[string][]string
+}
+
+// init will (re)initialize the containers object.
+func (c *Containers) init() {
+
+	c.mapContIds = make(map[string][]string, 0)
+	c.cgroupv1 = false
+	c.cgroupv2 = false
+	c.cgroupv1mp = ""
+	c.cgroupv2mp = ""
+}
+
+// addContainer adds a container given its uuid.
+func (c *Containers) addContainer(contId string) {
+	_, ok := c.mapContIds[contId]
+	if !ok {
+		c.mapContIds[contId] = make([]string, 0)
+	}
+}
+
+// GetContainers provides a list of all added containers by their uuid.
+func (c *Containers) GetContainers() []string {
+	var conts []string
+	for k := range c.mapContIds {
+		conts = append(conts, k)
+	}
+	return conts
+}
+
+// addPid will add a given pid string to an existing container by given uuid.
+// It will also create the container if given container uuid does not exist.
+func (c *Containers) addPid(contId string, pid string) {
+	c.addContainer(contId)
+	c.mapContIds[contId] = append(c.mapContIds[contId], pid)
+}
+
+func (c *Containers) GetPids(contId string) []string {
+	return c.mapContIds[contId]
+}
+
+// Populate will populate all Containers information by reading mounted proc
+// and cgroups filesystems.
+func (c *Containers) Populate() error {
+	// do all the hard work
+
+	c.init()
+
+	err := c.procMountsCgroups()
+	if err != nil {
+		return err
+	}
+
+	return c.populate()
+}
+
+// procMountsCgroups finds cgroups v1 and v2 mountpoints for the procfs walks.
+func (c *Containers) procMountsCgroups() error {
+	// find cgroups v1 and v2 mountpoints for procfs walks
+
+	mountsFile := "/proc/mounts"
+	file, err := os.Open(mountsFile)
+	if err != nil {
+		return err
+	}
+	scanner := bufio.NewScanner(file)
+	for i := 1; scanner.Scan(); i++ {
+		sline := strings.Split(scanner.Text(), " ")
+		mountpoint := sline[1]
+		fstype := sline[2]
+		if fstype == "cgroup" {
+			if strings.Contains(mountpoint, "cgroup/pids") {
+				c.cgroupv1 = true
+				c.cgroupv1mp = mountpoint
+			}
+		} else if fstype == "cgroup2" {
+			c.cgroupv2 = true
+			c.cgroupv2mp = mountpoint
+		}
+	}
+	_ = file.Close()
+
+	return nil
+}
+
+// populate prepares the regex(es) to be used for the population to be done in
+// proc walk.
+func (c *Containers) populate() error {
+
+	if c.cgroupv1 {
+		var r []string
+		// docker:  <cgroupv1mp>/<random>/docker/<id>/tasks
+		r = append(r, ".*docker.*(.{64})/tasks$")
+		// podman:  <cgroupv1mp>/<random>/libpod-<id>.scope/tasks
+		r = append(r, ".*libpod.*(.{64})\\.scope/tasks$")
+		// generic: <cgroupv1mp>/<random>/<id>.scope/tasks
+		r = append(r, ".*(.{64})\\.scope/tasks$")
+
+		err := c.populateProcWalk(c.cgroupv1mp, r)
+		if err != nil {
+			return err
+		}
+	}
+	if c.cgroupv2 {
+		var r []string
+		// docker:  <cgroupv2mp>/<random>/docker-<id>.scope/cgroup.procs
+		r = append(r, ".*docker.*(.{64})\\.scope/cgroup.procs$")
+		// podman:  <cgroupv2mp>/<random>/libpod-<id>.scope/cgroup.procs
+		r = append(r, ".*libpod.*(.{64})\\.scope/cgroup.procs$")
+		// generic: <cgroupv2mp>/<random>/<id>.scope/cgroup.procs
+		r = append(r, ".*(.{64})\\.scope/cgroup.procs$")
+
+		err := c.populateProcWalk(c.cgroupv2mp, r)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// populateProcWalk walks through procfs for cgroups (v1 & v2) filesystems and
+// find files based on given regex(es). It also extracts containers
+// information from found files and creates containers by their uuid AND adds
+// running pids to those containers.
+func (c *Containers) populateProcWalk(basedir string, allstr []string) error {
+
+	var allres []*regexp.Regexp
+	allres = make([]*regexp.Regexp, len(allstr))
+	for j := 0; j < len(allstr); j++ {
+		allres[j] = regexp.MustCompile(basedir + allstr[j])
+	}
+	err := filepath.WalkDir(basedir,
+		func(fn string, fi fs.DirEntry, err error) error {
+			if err != nil {
+				_, _ = fmt.Fprintln(os.Stderr, err)
+				return nil
+			}
+			for l := 0; l < len(allres); l++ {
+				res := allres[l].FindStringSubmatch(fn)
+				if len(res) == 0 {
+					continue
+				}
+				contId := res[1] // 1st regexp match (Id)
+				c.addContainer(contId)
+				buf, e := ioutil.ReadFile(fn) // res[0] or fn == .tasks or .procs file
+				if e != nil {
+					return e // error if can't read pids file
+				}
+				for _, pid := range strings.Split(string(buf), "\n") {
+					if len(pid) > 0 {
+						c.addPid(contId, pid)
+					}
+				}
+				break // a single match is enough
+			}
+			return nil
+		})
+
+	return err
+}
--- a/tracee-ebpf/tracee/containers.go
+++ b/tracee-ebpf/tracee/containers.go
@@ -17,24 +17,26 @@ type Containers struct {
 	cgroupv1mp string
 	cgroupv2   bool
 	cgroupv2mp string
-	mapContIds map[string][]string
+	mapContIds map[string][]int32
 }
 
-// init will (re)initialize the containers object.
-func (c *Containers) init() {
-
-	c.mapContIds = make(map[string][]string, 0)
-	c.cgroupv1 = false
-	c.cgroupv2 = false
-	c.cgroupv1mp = ""
-	c.cgroupv2mp = ""
+// InitContainers initializes a Containers object and returns a pointer to it.
+// User should further call "Populate" and iterate with Containers data.
+func InitContainers() *Containers {
+	return &Containers{
+		mapContIds: make(map[string][]int32, 0),
+		cgroupv1:   false,
+		cgroupv2:   false,
+		cgroupv1mp: "",
+		cgroupv2mp: "",
+	}
 }
 
 // addContainer adds a container given its uuid.
 func (c *Containers) addContainer(contId string) {
 	_, ok := c.mapContIds[contId]
 	if !ok {
-		c.mapContIds[contId] = make([]string, 0)
+		c.mapContIds[contId] = make([]int32, 0)
 	}
 }
 
@@ -49,12 +51,12 @@ func (c *Containers) GetContainers() []string {
 
 // addPid will add a given pid string to an existing container by given uuid.
 // It will also create the container if given container uuid does not exist.
-func (c *Containers) addPid(contId string, pid string) {
+func (c *Containers) addPid(contId string, pid int32) {
 	c.addContainer(contId)
 	c.mapContIds[contId] = append(c.mapContIds[contId], pid)
 }
 
-func (c *Containers) GetPids(contId string) []string {
+func (c *Containers) GetPids(contId string) []int32 {
 	return c.mapContIds[contId]
 }
 
@@ -62,8 +64,6 @@ func (c *Containers) GetPids(contId string) []string {
 // and cgroups filesystems.
 func (c *Containers) Populate() error {
 	// do all the hard work
-
-	c.init()
 
 	err := c.procMountsCgroups()
 	if err != nil {
@@ -168,7 +168,12 @@ func (c *Containers) populateProcWalk(basedir string, allstr []string) error {
 				}
 				for _, pid := range strings.Split(string(buf), "\n") {
 					if len(pid) > 0 {
-						c.addPid(contId, pid)
+						var pidInt int32
+						_, err = fmt.Sscanf(pid, "%d", &pidInt)
+						if err != nil {
+							continue // ignore sscanf errors
+						}
+						c.addPid(contId, pidInt)
 					}
 				}
 				break // a single match is enough
--- a/tracee-ebpf/tracee/tracee.go
+++ b/tracee-ebpf/tracee/tracee.go
@@ -836,6 +836,31 @@ func (t *Tracee) populateBPFMaps() error {
 		}
 	}
 
+	// Initialize pid_to_cont_id_map if tracing containers
+	c := Containers{}
+	err = c.Populate()
+	if err != nil {
+		return err
+	}
+	bpfPidToContIdMap, _ := t.bpfModule.GetMap("pid_to_cont_id_map")
+	for _, contId := range c.GetContainers() {
+		for _, pidstr := range c.GetPids(contId) {
+			if t.config.Debug {
+				fmt.Println("Running container =", contId, "pid =", pidstr)
+			}
+			var pid uint32
+			_, err = fmt.Sscanf(pidstr, "%d", &pid)
+			if err != nil {
+				return err
+			}
+			contIdBytes := []byte(contId)
+			err = bpfPidToContIdMap.Update(unsafe.Pointer(&pid), unsafe.Pointer(&contIdBytes[0]))
+			if err != nil {
+				return err
+			}
+		}
+	}
+
 	// Initialize tail calls program array
 	errs = make([]error, 0)
 	errs = append(errs, t.initTailCall(tailVfsWrite, "prog_array", "trace_ret_vfs_write_tail"))
--- a/tracee-ebpf/tracee/tracee.go
+++ b/tracee-ebpf/tracee/tracee.go
@@ -837,21 +837,15 @@ func (t *Tracee) populateBPFMaps() error {
 	}
 
 	// Initialize pid_to_cont_id_map if tracing containers
-	c := Containers{}
-	err = c.Populate()
-	if err != nil {
+	c := InitContainers()
+	if err := c.Populate(); err != nil {
 		return err
 	}
 	bpfPidToContIdMap, _ := t.bpfModule.GetMap("pid_to_cont_id_map")
 	for _, contId := range c.GetContainers() {
-		for _, pidstr := range c.GetPids(contId) {
+		for _, pid := range c.GetPids(contId) {
 			if t.config.Debug {
-				fmt.Println("Running container =", contId, "pid =", pidstr)
-			}
-			var pid uint32
-			_, err = fmt.Sscanf(pidstr, "%d", &pid)
-			if err != nil {
-				return err
+				fmt.Println("Running container =", contId, "pid =", pid)
 			}
 			contIdBytes := []byte(contId)
 			err = bpfPidToContIdMap.Update(unsafe.Pointer(&pid), unsafe.Pointer(&contIdBytes[0]))


yanivagman:
@rafaeldtinoco the problem with that approach is that it will only work with the docker runtime. Other container runtimes, such as containerd and cri-o will not work using this method, which also means that it won't work with recent kubernetes versions which don't rely on docker anymore

rafaeldtinoco:
Alright, I'll traverse the filesystem them, will change it tomorrow!

itaysk:
> the problem with that approach is that it will only work with the docker runtime

not only that, it also relies on docker being installed and available (and what about tracee in container), which is a dependency we can't take IMO. Today Tracee doesn't have runtime dependencies other than libbpf's dependencies

rafaeldtinoco:
> > the problem with that approach is that it will only work with the docker runtime
> 
> not only that, it also relies on docker being installed and docker client being accessible (and what about tracee in container), which is a dependency we can't take IMO. Today Tracee doesn't have runtime dependencies other than libbpf's dependencies so I'll possible we don't want to regress

Yes, it was a bad idea. Let me explain why I had it... I saw that the logic for the container_id coming from the cgroups uuid, but I also saw that was dependent on the way cgroups was being configured to each container instance. 

If we take other forms of containers, like LXC/LXD, they are not configured like that:

image

so the container_id for those containers does not seem to be working. This made me thought that we were only supporting docker for this particular feature.

Changing this code to read cgroups fs is easy. 

Question: Should we care about LXC/LXD containers ? 

image

If not because of k8s, then because of stand-alone tracee-bpf feature ?

rafaeldtinoco:
This is good to be reviewed (cgroups filesystem method now).

yanivagman:
> Question: Should we care about LXC/LXD containers ?
> If not because of k8s, then because of stand-alone tracee-bpf feature ?

Actually, I'm not that familiar with LXC/LXD, is there such a thing 'container ID' for them? If so, were is it saved?

rafaeldtinoco:
There is:

```
$ lxc list -c n,volatile.uuid
+----------------+--------------------------------------+
|      NAME      |            VOLATILE UUID             |
+----------------+--------------------------------------+
| bionic         | 6ae430f1-0cd2-4835-baa3-5b29fa5733bc |
+----------------+--------------------------------------+
| bionic-vm      | 336c2f1c-1d55-441b-ae59-397f7882d7ee |
+----------------+--------------------------------------+
| centos7        | 2b1a572e-af0d-44b4-a409-ef3f6c06ebc9 |
+----------------+--------------------------------------+
| centos7-tracee | 2e048bbb-b765-4490-a1a3-27cdc51463cd |
+----------------+--------------------------------------+
| centos7-vm     | 7082023c-50f0-4687-af2c-8c7490b4ec62 |
+----------------+--------------------------------------+
| centos8        | 87974d78-c274-4ff6-8eb5-5b485f2a2a34 |
+----------------+--------------------------------------+
```

I believe this would be the same as container_id BUT LXD uses cgroups filesystem with other hierarchy:

```
# ls -1 -d lxc.monitor.*
lxc.monitor.bionic
lxc.monitor.hirsute
lxc.monitor.zcopel
lxc.monitor.zrouter
lxc.monitor.zunifi
lxc.monitor.zvirtua
```

each of those dirs have a `tasks` file with their pids.

Maybe, for now, we can ignore lxc.* from inside cgroups filesystem in tracee.bpf.c ? So we don't get issues with bad container_id. In the future, if needed, we can extend the Containers object from:

https://github.com/aquasecurity/tracee/commit/7c1045669fc17d22fda02e5fe13bef0a370b6fdd

and have other functions such as PopulateLXDContainers() if needed.

WDYT ?

yanivagman (commented):
See my comments inline. We will need to find a more generic way to implement this feature, which will work on all environments, and not only for docker.
Here is a possible approach for such an implementation:
0. Locate cgroup**2** fs path (which might be located at /sys/fs/cgroup/unified, and not /sys/fs/cgroup that might be cgroup_v1)
1. Recursively walk through the cgroup2 fs, which is heirarchycal.
2. For every directory:
2.1 Get the pids attached to it by reading cgroup.procs. If there are no pids -> continue to the next dir.
2.2 Save the directory name.

After having this list of direcorty_name->pids, we still need to extract the container ids from the directory name. However, not every directory represents a container cgroup. So, we can do the following:
1. Trim prefix if exists (I'm only aware of a possible "docker-" prefix), and extract the first next 12 characters. If the directory name is smaller than 12 - ignore it.
2. Check if the directory name substring contains a valid container id name (we will need to think of a regex that will fit here). If the name is not valid - ignore it. Otherwise, extract the container id name and save it to the pid_to_contid_map for the relevant pids.

Note: this approach will not work on environments where only cgroup v1 is available, but I think we can assume that tracee will run on environments with cgroupv2 enabled. Otherwise, we will need to think of a similar approach for cgroupv1 environments.

tracee-ebpf/tracee/tracee.go
```diff
@@ -2086,3 +2104,57 @@ func readArgFromBuff(dataBuff io.Reader) (argTag, interface{}, error) {
...
+
+func (c *Containers) Get() map[string][]string {
+	// easify range for map[string][]string
+	return c.mapContIds
+}
+
+func (c *Containers) PopulateDockerContainers() error {
+
+	// cgroupfs pids is enough to get all container tasks
+	baseDir := "/sys/fs/cgroup/pids/docker/"
```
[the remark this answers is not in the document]
rafaeldtinoco: Hum, that is actually true. I mean, I assumed that by now all distros would still have cgroups 1 mounted and compatible but that might not be the case. This should check for both versions, and perhaps find out about the mount points from /proc/mounts. Will fix it.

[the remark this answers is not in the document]
rafaeldtinoco: I see you said in your comment:
```
 (which might be located at /sys/fs/cgroup/unified, and not /sys/fs/cgroup that might be cgroup_v1)
```
for this, will consider.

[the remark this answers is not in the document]
rafaeldtinoco:
```
Note: this approach will not work on environments where only cgroup v1 is available, but I think we can assume that tracee will run on environments with cgroupv2 enabled. Otherwise, we will need to think of a similar approach for cgroupv1 environments.
```

I'll get the cgroups mounts from /proc/mounts, check which version OS is using and where it is mounted. Then I'll traverse v1 or v2 in a way that it finds all the pids for all existing containers.

tracee-ebpf/tracee/tracee.go
```diff
@@ -736,6 +736,24 @@ func (t *Tracee) populateBPFMaps() error {
...
+	if t.config.Filter.ContFilter.Value {
+		bpfPidToContIdMap, _ := t.bpfModule.GetMap("pid_to_cont_id_map")
+		c := Containers{}
+		err := c.Populate()
+		if err != nil {
+			return fmt.Errorf("failed to populate pid_to_cont_id_map")
+		}
+		for fullcontid, pidarray := range c.Get() {
+			for _, pidstr := range pidarray {
+				container_id := fullcontid[:12]
```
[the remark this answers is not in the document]
rafaeldtinoco: I thought all dirs from inside a 'docker' directory within a cgroup fs would have obligatory the same amount of chars for the container_id (and was substring-ing-it to 12 because of the eBPF map). I do see your point in checking if there are 'other dirs' inside cgroups docker directory.

rafaeldtinoco:
Wrapping up some investigation:

```
 archlinux - cgroups v2:
   docker /sys/fs/cgroup/system.slice/docker-xxx.scope/cgroup.procs
   podman /sys/fs/cgroup/machine.slice/libpod-xxx.scope/{cgroup.procs, container/cgroup.procs}

 ubuntu - cgroups v1:
   docker /sys/fs/cgroup/pids/docker/xxx/tasks
   podman /sys/fs/cgroup/pids/machine.slice/libpod-xxx.scope/tasks

 ubuntu - cgroups v2:
   docker <ubuntu pkg: no cgroups v2 usage, pkg from docker: no cgroups v2 usage>
   podman /sys/fs/cgroup/machine.slice/libpod-xxx.scope/cgroups.procs
```

I think it is clear that I can discover cgroups v1 mountpoint, traverse it looking for docker and podman containers, add them, and do the same with cgroups v2. If one linux has both (like Ubuntu podman), then duplicates won't get added to the map (and wouldn't be by default anyway).

Good to go I believe.

rafaeldtinoco:
I'm pushing a new change and I would like you to test if possible. I have tested with podman and docker in Ubuntu with cgroups v1 and v2. But I know you're on arch (I based my changes in arch and others as well).

```
TIME             UTS_NAME         CONTAINER_ID     UID    COMM             PID/host        TID/host        RET              EVENT                ARGS
04:08:39:374553  d29c3cd3988f     d29c3cd3988f     0      bash             7      /1623500 7      /1623500 0                execve               pathname: /usr/bin/ls, argv: [ls --color=auto]
04:08:42:494053  313fb32a5b59     313fb32a5b59     0      bash             12     /1623583 12     /1623583 0                execve               pathname: /usr/bin/ls, argv: [ls --color=auto]
04:08:45:208951  83c710f6dbbc     83c710f6dbbc     0      bash             17     /1623719 17     /1623719 0                execve               pathname: /usr/bin/ls, argv: [ls --color=auto]
04:08:47:819753  7befaa8e13b6     7befaa8e13b6     0      sh               9      /1623720 9      /1623720 0                execve               pathname: /usr/bin/ls, argv: [ls]
```

Change looks good and supporting a bunch of stuff. We still have to fix tracee.btf.c because new containers get wrong container_id for podman:

```
04:08:06:753771  d29c3cd3988f     libpod-conmo     0      3                1      /1616957 1      /1616957 0                execve               pathname: /bin/bash, argv: [/bin/bash]
04:08:06:759622  d29c3cd3988f     libpod-conmo     0      bash             2      /1616961 2      /1616961 0                execve               pathname: /usr/bin/groups, argv: [groups]
04:08:06:763677  d29c3cd3988f     libpod-conmo     0      bash             3      /1616962 3      /1616962 0                execve               pathname: /usr/bin/dircolors, argv: [dircolors -b]
04:08:07:364265  d29c3cd3988f     libpod-conmo     0      bash             4      /1616963 4      /1616963 0                execve               pathname: /usr/bin/ls, argv: [ls --color=auto]
```

I can do it tomorrow, no biggie.

tracee-ebpf/tracee/tracee.go
```diff
@@ -2086,3 +2104,228 @@ func readArgFromBuff(dataBuff io.Reader) (argTag, interface{}, error) {
...
+		re := regexp.MustCompile(c.cgroupv2mp + `.*docker.*(?P<id>.{64})\.scope/cgroup.procs$`)
+		err := c._PopulateProcWalk(c.cgroupv2mp, re)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (c *Containers) _PopulatePodman() error {
```
yanivagman, line 2287: This function and _PopulateDocker() are the same but with a different regex.
We can have one function that gets the regex string as an argument and avoid code duplication for different container runtimes.

tracee-ebpf/tracee/tracee.go
```diff
@@ -2086,3 +2104,228 @@ func readArgFromBuff(dataBuff io.Reader) (argTag, interface{}, error) {
...
+	}
+	err = c._PopulateDocker()
+	if err != nil {
+		return err
+	}
+	err = c._PopulatePodman()
+	if err != nil {
+		return err
+	}
+
```
yanivagman, line 2218: We should add a third populate that doesn't contain any prefix (e.g. docker/libpod) in its regex (I think this should work for cri-o/containerd)

yanivagman (commented):
@rafaeldtinoco works on my env, and indeed looks good!
I think we can move all the Containers struct methods to a separate file, WDYT?

rafaeldtinoco:
All good for a new review. It took some extra time due to a local bug I faced (host kernel) but it is all good now. I wrapped all regex strings into a slice of strings and compiled them before walking procfs so it has better performance. I have also added a 'generic' regex to cgroupv1 and cgroupv2 cases.

tracee-ebpf/tracee/containers.go
```diff
@@ -0,0 +1,240 @@
...
+func (c *Containers) _DelContainer(contid string) {
+	_, ok := c.mapContIds[contid]
+	if ok {
+		delete(c.mapContIds, contid)
+	}
+}
+
+func (c *Containers) _CheckContainer(contid string) bool {
+	_, ok := c.mapContIds[contid]
+	if !ok {
```
josedonizetti, line 70:
```
return ok
```

tracee-ebpf/tracee/containers.go
```diff
@@ -0,0 +1,240 @@
...
+	return conts
+}
+
+func (c *Containers) _AddPid(contid string, pid string) {
+	c._AddContainer(contid)
+	c.mapContIds[contid] = InsertSortUniq(c.mapContIds[contid], pid)
+}
+
+func (c *Containers) _DelPid(contid string, pid string) {
+	if c._CheckContainer(contid) == true {
```
josedonizetti, line 95:
```suggestion
	if c._CheckContainer(contid) {
```

tracee-ebpf/tracee/containers.go
```diff
@@ -0,0 +1,240 @@
...
+		c.mapContIds[contid] = DeleteSortUniq(c.mapContIds[contid], pid)
+	}
+}
+
+func (c *Containers) GetPids(contid string) []string {
+	if c._CheckContainer(contid) == false {
```
josedonizetti, line 101:
```suggestion
	if !c._CheckContainer(contid)  {
```

tracee-ebpf/tracee/containers.go
```diff
@@ -0,0 +1,240 @@
...
+
+func (c *Containers) Populate() error {
+	var err error
+
+	err = c._ProcMountsCgroups()
+	if err != nil {
+		return err
+	}
+	err = c._Populate()
+	if err != nil {
```
josedonizetti, line 117: No need to check nil, if it is returning err anyway below. 
it can be `return err`

tracee-ebpf/tracee/containers.go
```diff
@@ -0,0 +1,240 @@
...
+			if strings.Contains(mountpoint, "cgroup/pids") {
+				c.cgroupv1 = true
+				c.cgroupv1mp = mountpoint
+			}
+		} else if fstype == "cgroup2" {
+			c.cgroupv2 = true
+			c.cgroupv2mp = mountpoint
+		}
+	}
+	_ = file.Close()
```
josedonizetti, line 147:
```suggestion
	return file.Close()
```

or if you want to ignore the err, no need to assign it to `_`

tracee-ebpf/tracee/containers.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+func (c *Containers) _PopulateProcWalk(v2 bool, allstr []string) error {
+	// Walk through procfs for cgroups v1 or v2 filesystem and find files based on regex
+	// Extract container information from found files
+
+	var basedir string
+	var allres []*regexp.Regexp
+
+	if v2 == true {
```
josedonizetti, line 188:
```suggestion
	if v2 {
```

tracee-ebpf/tracee/containers.go
```diff
@@ -0,0 +1,240 @@
...
+
+	err := filepath.Walk(basedir,
+		func(fn string, fi os.FileInfo, err error) error {
+			if err != nil {
+				return nil // no error in procfs reads
+			}
+			var matched bool
+			var re *regexp.Regexp
+			for _, re = range allres {
+				if re.MatchString(fn) == true {
```
josedonizetti, line 206:
```suggestion
				if re.MatchString(fn)  {
```

tracee-ebpf/tracee/containers.go
```diff
@@ -0,0 +1,240 @@
...
+					matched = true // matched any of compiled regexes
+					break
+				}
+			}
+			if matched == false {
```
josedonizetti, line 211:
```suggestion
			if !matched  {
```

tracee-ebpf/tracee/tracee.go
```diff
@@ -736,6 +736,24 @@ func (t *Tracee) populateBPFMaps() error {
...
 	bpfConfigMap.Update(uint32(configDebugNet), boolToUInt32(t.config.Debug))
 
+	// Initialize pid_to_cont_id_map if tracing containers
+	c := Containers{}
+	err := c.Populate()
+	if err != nil {
+		return fmt.Errorf("failed to populate pid_to_cont_id_map")
+	}
+	bpfPidToContIdMap, _ := t.bpfModule.GetMap("pid_to_cont_id_map")
+	for _, cont_id := range c.GetContainers() {
```
josedonizetti, line 746:
```suggestion
	for _, contd := range c.GetContainers() {
```
naming convention: https://golang.org/doc/effective_go#mixed-caps

josedonizetti (commented):
@rafaeldtinoco I hope you don't mind, but I did a quick review more related to code style. Also, unless it is a convention on the tracee code base, I think it would be best to avoid starting method names with `_`. If the methods are not to be used on the public API, they should be `low case` to avoid exporting outside the package.

tracee-ebpf/tracee/containers.go
rafaeldtinoco: loved that one =)

tracee-ebpf/tracee/containers.go
rafaeldtinoco: Some of those are just so lint stop highlighting things :\

tracee-ebpf/tracee/tracee.go
rafaeldtinoco: Yep!!! thx

tracee-ebpf/tracee/containers.go
rafaeldtinoco: > @rafaeldtinoco I hope you don't mind, but I did a quick review more related to code style. Also, unless it is a convention on the tracee code base, I think it would be best to avoid starting method names with `_`. If the methods are not to be used on the public API, they should be `low case` to avoid exporting outside the package.

I don't mind at all and it is very good for me to improve my golang skills as I'm recent to the language. I have changed most of the suggestions, kept some so my local lint-ing does not complain too much.

About the convention, yep! We should stick to that for sure.

rafaeldtinoco:
Okay, the issue: #894 was driving me crazy and that is why this had so many force pushes and back and forth. I'm pushing my latest version for review BUT this can't be merged until issue #894 is solved.

tracee-ebpf/tracee/containers.go
```diff
@@ -0,0 +1,205 @@
...
+
+func (c *Containers) populate() error {
+	// Prepare regex to be used for population done in proc walk
+
+	if c.cgroupv1 {
+		var r []string
+		// docker:  <cgroupv1mp>/<random>/docker/<id>/tasks
+		r = append(r, ".*docker.*(.{64})/tasks$")
+		// podman:  <cgroupv1mp>/<random>/libpod-<id>.scope/tasks
+		r = append(r, ".*libpod.*(.{64})\\.scope/tasks$")
```
yanivagman, line 135: Are these needed now that you have "generic" regex?

tracee-ebpf/tracee/containers.go
```diff
@@ -0,0 +1,205 @@
...
+func (c *Containers) populateProcWalk(v2 bool, allstr []string) error {
+	// Walk through procfs for cgroups v1 or v2 filesystem and find files based on regex
+	// Extract container information from found files
+
+	var basedir string
+	if v2 {
+		basedir = c.cgroupv2mp
+	} else {
+		basedir = c.cgroupv1mp
+	}
```
yanivagman, line 171: You can give basedir as an argument instead of v2 (I don't see any other usage for v2 other than setting basedir)

tracee-ebpf/tracee/containers.go
```diff
@@ -0,0 +1,205 @@
...
+
+func (c *Containers) addPid(contId string, pid string) {
+	c.addContainer(contId)
+	c.mapContIds[contId] = append(c.mapContIds[contId], pid)
+}
+
+func (c *Containers) GetPids(contId string) []string {
+	if !c.checkContainer(contId) {
+		return nil
+	}
```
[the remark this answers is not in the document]
rafaeldtinoco: I wasn't sure about that as well. Afraid of generating a null de-reference.

tracee-ebpf/tracee/containers.go
rafaeldtinoco: I have not tested the generic regexes while I have tested the other 4. I want to make sure the 4 are good and the 3rd generic regex is a 'safe measure'. We can change later if redundant.

tracee-ebpf/tracee/tracee.go
```diff
@@ -741,6 +741,27 @@ func (t *Tracee) populateBPFMaps() error {
...
+		return err
+	}
+	bpfPidToContIdMap, _ := t.bpfModule.GetMap("pid_to_cont_id_map")
+	for _, contId := range c.GetContainers() {
+		for _, pidstr := range c.GetPids(contId) {
+			if t.config.Debug {
+				fmt.Println("Running container =", contId, "pid =", pidstr)
+			}
+			var pid uint32
+			_, err = fmt.Sscanf(pidstr, "%d", &pid)
```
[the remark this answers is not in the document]
rafaeldtinoco: leftover: I was checking and removed thinking there was no way it couldn't be a %d in that file (because its a procfs file).. but we never know, I'm putting it back.

tracee-ebpf/tracee/containers.go
```diff
@@ -0,0 +1,180 @@
...
+
+	var allres []*regexp.Regexp
+	allres = make([]*regexp.Regexp, len(allstr))
+	for j := 0; j < len(allstr); j++ {
+		allres[j] = regexp.MustCompile(basedir + allstr[j])
+	}
+	err := filepath.WalkDir(basedir,
+		func(fn string, fi fs.DirEntry, err error) error {
+			if err != nil {
+				_, _ = fmt.Fprintln(os.Stderr, err)
```
yanivagman, line 155: Nit: if you don't care about the return value, you can just `fmt.Fprintln(os.Stderr, err)`

tracee-ebpf/tracee/containers.go
```diff
@@ -0,0 +1,180 @@
...
+				}
+				contId := res[1] // 1st regexp match (Id)
+				c.addContainer(contId)
+				buf, e := ioutil.ReadFile(fn) // res[0] or fn == .tasks or .procs file
+				if e != nil {
+					return e // error if can't read pids file
+				}
+				for _, pid := range strings.Split(string(buf), "\n") {
+					if len(pid) > 0 {
+						c.addPid(contId, pid)
```
yanivagman, line 171: Not a must, but it is more intuitive to save the pids as ints (and change the map type accordingly).
This will also save the later sscanf in tracee.go.

yanivagman (dismissed):

tracee-ebpf/tracee/containers.go
rafaeldtinoco: Yes, but then goland default linter keeps complaining. I should seek to change that.

tracee-ebpf/tracee/containers.go
rafaeldtinoco: Yep, it crossed my mind also. I'll change it.

rafaeldtinoco:
Plan here is:

- pull request #864 finished (Fixes: #862), add more commits:
- tracee needs to use latest libbpfgo (Fixes: #894)
- change tracee to use new libbpfgo map methods (Fixes: #814)

I'm working on it and will pull #864 altogether soonish.

rafaeldtinoco:
We need the following 4 commits together in order to address the container_id issue, the race during bpf maps initialization AND the new libbpfgo version (which contains the fix for the bpf maps initialization issue).

```
commit 899848e (HEAD -> refs/heads/issue862, refs/remotes/rafaeldtinoco/issue862)
Date:   Sun Aug 15 16:40:52 2021

    tracee-ebpf: Change libbpfgo map methods to use unsafe.Pointer
    
    Closes: #814

commit 2add622
Date:   Sun Aug 15 13:17:25 2021

    tracee-ebpf: update libbpf to sync with libbpfgo

commit 6bc9418
Date:   Sun Aug 15 01:49:36 2021

    tracee-ebpf: update to libbpfgo v0.2.0-libbpf_0.4.0
    
    Fixes: #894

commit b786e37
Date:   Thu Aug 5 04:12:40 2021

    tracee.go: initialize pid_to_cont_id_map during startup
    
    Fixes: #862
```

I want to change the container_id fix to use pids as int() instead of string().. I'll add a commit to this patchset soonish. This can already be reviewed. I have tested and it seems to be working fine.

rafaeldtinoco:
Added 

```
commit 2b1ce1f (HEAD -> refs/heads/issue862, refs/remotes/rafaeldtinoco/issue862)
Date:   Mon Aug 16 00:29:14 2021

    tracee-ebpf: improve Containers object
    
    - start logic: InitContainers() function
    - mapContIds as a map[uuid][]int32: pids as int32 makes more sense

commit 0b5adbe
Date:   Sun Aug 15 23:39:21 2021

    tracee-ebpf: improve error handling and add missing ones
```

And this is good for a final review now (and further fixes if needed).

yanivagman:
@rafaeldtinoco can we split this into two?
I would first update tracee to use the new libbpfgo version and apply the required changes to compile tracee with it (e.g. Update() maps).
In a different PR (this PR) - I would take all the PRs that handle the pid_to_cont_id_map initialization.

Are you ok with doing this split?

rafaeldtinoco:
The split is at: https://github.com/aquasecurity/tracee/pull/926

Once that PR is merged, I'll rebase this one (branch issue862) and we can work it through.

Thank you!

yanivagman:
@rafaeldtinoco can you please rebase this one so it will be easier to review?

rafaeldtinoco:
@yanivagman rebased and working:

```
$ sudo ./dist/tracee-ebpf --debug --trace container --trace event=execve
BTF enabled, attempting to unpack CORE bpf object
unpacked CO:RE bpf object file into memory
Running container = 70c8252a6255ed105754cd040c545bfeba65f52c4feba417b035ef851c4364a8 pid = 1206770
Running container = 70c8252a6255ed105754cd040c545bfeba65f52c4feba417b035ef851c4364a8 pid = 1206767
Running container = 70c8252a6255ed105754cd040c545bfeba65f52c4feba417b035ef851c4364a8 pid = 1206769
Running container = 70c8252a6255ed105754cd040c545bfeba65f52c4feba417b035ef851c4364a8 pid = 1206770
Running container = 70c8252a6255ed105754cd040c545bfeba65f52c4feba417b035ef851c4364a8 pid = 1206767
Running container = a99b06d705af8bff86e787dccf665ce44183da525389ad1056e94856f57f7e35 pid = 1209971
Running container = a99b06d705af8bff86e787dccf665ce44183da525389ad1056e94856f57f7e35 pid = 1209968
Running container = a99b06d705af8bff86e787dccf665ce44183da525389ad1056e94856f57f7e35 pid = 1209970
Running container = a99b06d705af8bff86e787dccf665ce44183da525389ad1056e94856f57f7e35 pid = 1209971
Running container = a99b06d705af8bff86e787dccf665ce44183da525389ad1056e94856f57f7e35 pid = 1209968
Running container = 38301d81987a06c64b7dedf5c194593a5a85b43bfa50ce948590731a5362e340 pid = 1205289
Running container = 38301d81987a06c64b7dedf5c194593a5a85b43bfa50ce948590731a5362e340 pid = 3817400
Running container = b3ab2b73e261011a18860e1dccaef19f5ae8de5f5872d9a8c60334d26d46c0f7 pid = 1205908
Running container = b3ab2b73e261011a18860e1dccaef19f5ae8de5f5872d9a8c60334d26d46c0f7 pid = 3814886
TIME             UTS_NAME         CONTAINER_ID     UID    COMM             PID/host        TID/host        RET              EVENT                ARGS
14:09:29:639530  a99b06d705af     a99b06d705af     0      bash             46     /1644937 46     /1644937 0                execve               pathname: /usr/bin/ls, argv: [ls --color=auto]
14:09:34:025937  b3ab2b73e261     b3ab2b73e261     0      bash             62     /1645020 62     /1645020 0                execve               pathname: /usr/bin/ls, argv: [ls --color=auto]
14:09:42:492754  a0ee63dd079a     a0ee63dd079a     0      runc:[2:INIT]    1      /1648086 1      /1648086 0                execve               pathname: /bin/sh, argv: [/bin/sh]
14:09:52:680638  a0ee63dd079a     a0ee63dd079a     0      sh               7      /1648262 7      /1648262 0                execve               pathname: /usr/bin/bash, argv: [bash]
14:09:52:687362  a0ee63dd079a     a0ee63dd079a     0      bash             8      /1648263 8      /1648263 0                execve               pathname: /usr/bin/groups, argv: [groups]
14:09:52:691560  a0ee63dd079a     a0ee63dd079a     0      bash             9      /1648264 9      /1648264 0                execve               pathname: /usr/bin/dircolors, argv: [dircolors -b]

End of events stream
Stats: {eventCounter:6 errorCounter:0 lostEvCounter:0 lostWrCounter:0 lostNtCounter:0}
```

yanivagman (approved):

PR #867, opened 2021-08-03
fix rule name partially cropped in error message

AsafEitani:
When an error occurs in rego rule - the error message used to print the first 20 bytes - effectively cropping the full name of the rules for rule names longer than 20 bytes.
This PR should fix it.

### DIFF
--- a/tracee-rules/signature.go
+++ b/tracee-rules/signature.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	_ "embed"
 
 	"fmt"
@@ -97,7 +98,11 @@ func findRegoSigs(dir string) ([]types.Signature, error) {
 		}
 		sig, err := regosig.NewRegoSignature(string(regoCode), regoHelpersCode)
 		if err != nil {
-			log.Printf("error creating rego signature with: %s: %v ", regoCode[0:20], err)
+			newlineOffset := bytes.Index(regoCode, []byte("\n"))
+			if newlineOffset == -1 {
+				newlineOffset = len(regoCode)
+			}
+			log.Printf("error creating rego signature with: %s: %v ", regoCode[0:newlineOffset], err)
 			continue
 		}
 		res = append(res, sig)
--- a/tracee-rules/signature.go
+++ b/tracee-rules/signature.go
@@ -100,7 +100,12 @@ func findRegoSigs(dir string) ([]types.Signature, error) {
 		if err != nil {
 			newlineOffset := bytes.Index(regoCode, []byte("\n"))
 			if newlineOffset == -1 {
-				newlineOffset = len(regoCode)
+				codeLength := len(regoCode)
+				if codeLength < 22 {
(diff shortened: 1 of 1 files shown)


tracee-rules/signature.go
```diff
@@ -97,7 +98,7 @@ func findRegoSigs(dir string) ([]types.Signature, error) {
 		}
 		sig, err := regosig.NewRegoSignature(string(regoCode), regoHelpersCode)
 		if err != nil {
-			log.Printf("error creating rego signature with: %s: %v ", regoCode[0:20], err)
+			log.Printf("error creating rego signature with: %s: %v ", regoCode[0:bytes.Index(regoCode, []byte("\n"))], err)
```
grantseltzer, line 101: From the GoDOC for bytes.Index:

`Index returns the index of the first instance of sep in s, or -1 if sep is not present in s.`

So this can potentially cause a panic. Can we not do `regoCode[0:len(regoCode)]` or just `regoCode` ?

AsafEitani: That will print the entire rego code, which is a bit much in my opinion.
I'll add a check to see if the returned value is -1.
Good catch - thanks!

tracee-rules/signature.go
```diff
@@ -97,7 +98,11 @@ func findRegoSigs(dir string) ([]types.Signature, error) {
 		}
 		sig, err := regosig.NewRegoSignature(string(regoCode), regoHelpersCode)
 		if err != nil {
-			log.Printf("error creating rego signature with: %s: %v ", regoCode[0:20], err)
+			newlineOffset := bytes.Index(regoCode, []byte("\n"))
+			if newlineOffset == -1 {
+				newlineOffset = len(regoCode)
```
grantseltzer, line 103: So I think it doesn't really matter, but in this case it'll print the whole rego code. Maybe we limit it at 20 like above? Perhaps just change this to `len(regoCode)%20`? or some other arbitrary (but reasonable) limit number. WDYT?

tracee-rules/signature.go
```diff
@@ -97,7 +98,16 @@ func findRegoSigs(dir string) ([]types.Signature, error) {
...
 		if err != nil {
-			log.Printf("error creating rego signature with: %s: %v ", regoCode[0:20], err)
+			newlineOffset := bytes.Index(regoCode, []byte("\n"))
+			if newlineOffset == -1 {
+				codeLength := len(regoCode)
+				if codeLength < 22 {
+					newlineOffset = codeLength
+				} else {
+					newlineOffset = 22
+				}
```
grantseltzer, line 108: newlineOffset = 
```suggestion
                              newlineOffset = len(regoCode) % 22 
```

AsafEitani: problem is if the length of the code is exactly 22 characters - the message won't print anything...

grantseltzer: Ah makes sense, good point

grantseltzer (approved):

PR #868, opened 2021-08-03
Update docs to take into account CO:RE default

grantseltzer:

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -22,10 +22,18 @@ The full documentation of Tracee is available at [https://aquasecurity.github.io
 
 Before you proceed, make sure you follow the [minimum requirements for running Tracee](https://aquasecurity.github.io/tracee/dev/install/prerequisites/).
 
+If running on BTF enabled kernel:
+
 ```bash
-docker run --name tracee --rm --privileged -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro -v /tmp/tracee:/tmp/tracee -it aquasec/tracee:latest
+docker run --name tracee --rm --privileged -it aquasec/tracee:latest
 ```
 
+If running on BTF disabled kernel:
+
+```bash
+docker run --name tracee --rm --privileged -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro -v /tmp/tracee:/tmp/tracee -it aquasec/tracee:latest
+
+```
 This will run Tracee with default settings and start reporting detections to standard output.  
 In order to simulate a suspicious behavior, you can run `strace ls` in another terminal, which will trigger the "Anti-Debugging" signature, which is loaded by default.
 
--- a/docs/install/ebpf-compilation.md
+++ b/docs/install/ebpf-compilation.md
@@ -24,5 +24,5 @@ In this case, the full Docker image can be replaced by the lighter-weight `aquas
 If using Docker, the following `docker run` options demonstrate mounting a pre-compiled eBPF artifact into the container, and pointing Tracee to use it:
 
 ```bash
-docker run ... -v /path/in/host/tracee.bpf.123.o:/path/in/container/tracee.bpf.o aquasec/tracee
+docker run ... -v /path/in/host/tracee.bpf.123.o:/path/in/container/tracee.bpf.o -e TRACEE_BPF_FILE=/path/in/container/tracee.bpf.o aquasec/tracee
 ```
--- a/docs/install/ebpf-compilation.md
+++ b/docs/install/ebpf-compilation.md
@@ -1,9 +1,12 @@
 # eBPF Compilation
 
-Tracee is leveraging Linux's eBPF technology, which is kernel and version sensitive. Therefore, Tracee's eBPF component needs to be specifically compiled for your hosts.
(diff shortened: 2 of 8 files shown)


docs/install/ebpf-compilation.md
```diff
@@ -21,11 +24,5 @@ In this case, the full Docker image can be replaced by the lighter-weight `aquas
 If using Docker, the following `docker run` options demonstrate mounting a pre-compiled eBPF artifact into the container, and pointing Tracee to use it:
 
 ```bash
-docker run ... -v /path/in/host/tracee.bpf.123.o:/path/in/container/tracee.bpf.o -e TRACEE_BPF_FILE=/path/in/container/tracee.bpf.o aquasec/tracee
+docker run ... -v /path/in/host/tracee.bpf.123.o:/path/in/container/tracee.bpf.o aquasec/tracee
```
yanivagman, line 27: Why did you remove the environment variable?

yanivagman (commented):
Don't we also want to update the one liner (both in the readme and docs) to start tracee? If a kernel has BTF enabled, it is possible to remove the linux headers mounts

docs/install/ebpf-compilation.md
grantseltzer: The CO:RE object is embedded, so it's not needed to run

grantseltzer: On second thought, since this section is for mounting the precompiled binary, it should stay. Good call out.

Readme.md
```diff
@@ -22,10 +22,18 @@ The full documentation of Tracee is available at [https://aquasecurity.github.io
...
+If running on BTF enabled kernel:
+
 ```bash
-docker run --name tracee --rm --privileged -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro -v /tmp/tracee:/tmp/tracee -it aquasec/tracee:latest
+docker run --name tracee --rm --privileged -it aquasec/tracee:latest
 ```
 
+If running on BTF disabled kernel:
+
+```
```
yanivagman, line 33: Missing 'bash'?

yanivagman (approved):
LGTM!

docs/install/prerequisites.md
```diff
@@ -1,11 +1,18 @@
 # Prerequisites
 
+Portable (CO:RE) option:
 - Linux kernel version >= 4.18
-- Linux kernel headers available under conventional location (see [Linux Headers](../headers) section for more info)
+- `CONFIG_DEBUG_INFO_BTF` Linux compile option enabled (only needed for running, not compilation)
```
itaysk, line 5: IMO the kernel config is not the requisite, BTF is. I suggest something similar to my other comment about this

docs/install/prerequisites.md
```diff
@@ -1,11 +1,18 @@
...
 - libc, and the libraries: libelf, zlib
 - GNU Make >= 4.3
-- clang >= 9
+- clang >= 11
```
itaysk, line 8: I think with CO:RE we can start to remove the compile-time requirements from here, since this is the first thngs the user sees about tracee, and it's quite intimidating. wdyt about moving this to the ebpf-compilation doc?

Readme.md
```diff
@@ -22,10 +22,18 @@ The full documentation of Tracee is available at [https://aquasecurity.github.io
...
-docker run --name tracee --rm --privileged -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro -v /tmp/tracee:/tmp/tracee -it aquasec/tracee:latest
+docker run --name tracee --rm --privileged -it aquasec/tracee:latest
 ```
 
+If running on BTF disabled kernel:
+
+```bash
+docker run --name tracee --rm --privileged -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro -v /tmp/tracee:/tmp/tracee -it aquasec/tracee:latest
+
+```
```
itaysk, line 36: IMO remove from the quickstart, and link to the doc If non-btf

grantseltzer:
Noting here that it appears libbpf needs access to the Kconfig (`/boot/config-$(uname -r)`) for CO:RE to work.

itaysk:
as a general note, all of the comments are just because I'm considering these docs user facing and not developer (contributor) facing.

docs/tracee-ebpf/index.md
```diff
@@ -9,7 +9,7 @@ In some cases, you might want to leverage Tracee's eBPF event collection capabil
 Before you proceed, make sure you follow the [minimum requirements for running Tracee](install/prerequisites.md).
 
 ```bash
-docker run --name tracee --rm --privileged -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro -v /tmp/tracee:/tmp/tracee -it aquasec/tracee:latest trace
+docker run --name tracee --rm --privileged -it aquasec/tracee:latest trace
```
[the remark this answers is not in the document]
grantseltzer: It seems that sysfs is automatically read only mounted.

Readme.md
grantseltzer: Which doc do you think should be linked to?

grantseltzer: Perhaps a new document that talks about the differences between CO:RE and kernel-header-dependent versions of tracee

PR #869, opened 2021-08-03
add security_kernel_read_file lsm hook

roikol:


grantseltzer:
@roikol Code looks good, is there a way I can manually test this? How can I trigger the `security_kernel_read_file` hook?

PR #873, opened 2021-08-04
tracee-ebpf: make tracee-ebpf static for portability

rafaeldtinoco:
With CO-RE support, it makes sense that tracee-ebpf binary is also
portable. This way, one may compile tracee-ebpf in one distribution
version and execute in another one, avoiding problems such as:

./dist/tracee-ebpf: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by ./dist/tracee-ebpf)

### DIFF
--- a/tracee-ebpf/Makefile
+++ b/tracee-ebpf/Makefile
@@ -44,8 +44,8 @@ build: $(OUT_BIN)
 go_env := GOOS=linux GOARCH=$(ARCH:x86_64=amd64) CC=$(CMD_CLANG) CGO_CFLAGS="-I $(abspath $(LIBBPF_HEADERS))" CGO_LDFLAGS="$(abspath $(LIBBPF_OBJ))"
 ifndef DOCKER
 $(OUT_BIN): $(LIBBPF_HEADERS) $(LIBBPF_OBJ) $(OUT_BPF_CORE) $(filter-out *_test.go,$(GO_SRC)) $(BPF_BUNDLE) | $(OUT_DIR)
-	$(go_env) go build -v -o $(OUT_BIN) \
-	-ldflags "-X main.version=$(VERSION)"
+	$(go_env) go build -tags netgo -v -o $(OUT_BIN) \
+	-ldflags "-w -extldflags \"-static\" -X main.version=$(VERSION)"
 else 
 $(OUT_BIN): $(DOCKER_BUILDER) | $(OUT_DIR)
 	$(call docker_builder_make,$@ VERSION=$(VERSION))


rafaeldtinoco:
```
$ sudo ./dist/tracee-ebpf --debug --trace pid=new
BTF is not enabled
attempting to build the bpf object file
building bpf object in: /tmp/tracee-make736244672
/usr/bin/clang -S -D__BPF_TRACING__ -D__KERNEL__ -D__TARGET_ARCH_x86 -I/tmp/tracee-make736244672 -include/lib/modules/5.4.0-80-generic/build/include/linux/kconfig.h -I/lib/modules/5.4.0-80-generic/build/arch/x86/include -I/lib/modules/5.4.0-80-generic/build/arch/x86/include/uapi -I/lib/modules/5.4.0-80-generic/build/arch/x86/include/generated -I/lib/modules/5.4.0-80-generic/build/arch/x86/include/generated/uapi -I/lib/modules/5.4.0-80-generic/build/include -I/lib/modules/5.4.0-80-generic/build/include -I/lib/modules/5.4.0-80-generic/build/include/uapi -I/lib/modules/5.4.0-80-generic/build/include/generated -I/lib/modules/5.4.0-80-generic/build/include/generated/uapi -Wno-address-of-packed-member -Wno-compare-distinct-pointer-types -Wno-deprecated-declarations -Wno-gnu-variable-sized-type-not-at-end -Wno-pointer-sign -Wno-pragma-once-outside-heade -Wno-unknown-warning-option -Wno-unused-value -Wunused -Wall -fno-stack-protector -fno-jump-tables -fno-unwind-tables -fno-asynchronous-unwind-tables -xc -nostdinc -O2 -emit-llvm -c -g /tmp/tracee-make736244672/tracee.bpf.c -o/tmp/tracee-make736244672/tracee.bpf.ll
/usr/bin/llc -march=bpf -filetype=obj -o /tmp/tracee-make736244672/tracee.bpf.o /tmp/tracee-make736244672/tracee.bpf.ll
/usr/bin/llc: /tmp/tracee-make736244672/tracee.bpf.ll:940:92: error: expected ')' at end of argument list
define dso_local i32 @tracepoint__raw_syscalls__sys_enter(%struct.bpf_raw_tracepoint_args* %0) #0 section "raw_tracepoint/sys_enter" !dbg !12131 {
                                                                                           ^
2021/08/04 04:51:20 failed to make BPF object (llc): exit status 1. Try using --debug for more info
```

When trying to use tracee-ebpf without BTF.

rafaeldtinoco:
This is the fence in between preparing tracee-ebpf as a OS tool and package it to each distribution version, or, for now, keep it as is and guarantee that, at least, we can copy the binary to different flavors for development purposes.

yanivagman:
@rafaeldtinoco have a look at #565 .
I'm also in favor of statically linking, we just need to make sure we know all the implications.
Maybe we should add another build target instead of replacing the existing one?

itaysk:
+1 for static. But I don't see the value in another target. if we will use the static target in the release the other one will be redundant, and if we don't use it in the release, why is it there

grantseltzer:
An issue with static builds that i've run into is that fedora doesn't provide a libelf-static package. I had to build libelf.a from source in order to get it to work.

rafaeldtinoco:
This was closed by accident because of a local branch deletion when I was configuring auto-merge deletion. Will open another PR for the static building.

PR #874, opened 2021-08-04
add close on fileread finish to tracee-rules

AsafEitani:


AsafEitani:
Fixed issue of tracee-rules exiting before signatures finished printing.

tracee-rules/engine/engine.go
```diff
@@ -219,7 +234,7 @@ func (engine *Engine) LoadSignature(signature types.Signature) (string, error) {
 		engine.logger.Printf("error initializing signature %s: %v", metadata.Name, err)
 
 	}
-	go signatureStart(signature, c)
+	go signatureStart(signature, c, &engine.waitGroup)
```
yanivagman, line 237: Shouldn't you add `engine.waitGroup.Add(1)` before this line?

itaysk: preferably inside signature start

AsafEitani: I am :)
Thanks

yanivagman (approved):

PR #879, opened 2021-08-04
tracee-ebpf: add flags support for make test

josedonizetti:
Allowing external flags to `go test`. This simplify getting coverage reports or running single tests. Eg:
```
make test GOTEST_FLAGS="-run Test_updateFileSHA"
make test GOTEST_FLAGS="-cover"
...
```


itaysk (commented):
good idea. while we're doing that, do you mind adding the same to the other makefiles?

josedonizetti:
@itaysk  I added it to tracee-rules/Makefile, but opted to not change the tracee-rules/benchmark/Makefile which is more specific and already declared flags of its own.

yanivagman (approved):
LGTM

PR #881, opened 2021-08-05
tracee-ebpf: add output validate test

josedonizetti:
Add tests for `OutputConfig.Validate()`:

```
=== RUN   TestOutputConfigValidate
=== RUN   TestOutputConfigValidate/format:table
=== RUN   TestOutputConfigValidate/format:table-verbose
=== RUN   TestOutputConfigValidate/format:json
=== RUN   TestOutputConfigValidate/format:gob
=== RUN   TestOutputConfigValidate/format:gotemplate=tmpl
=== RUN   TestOutputConfigValidate/invalid_format
```

I also refactored the code in a way I consider more readable than the huge `if`, but I can revert it if you prefer the previous version.

### DIFF
--- a/tracee-ebpf/tracee/tracee.go
+++ b/tracee-ebpf/tracee/tracee.go
@@ -134,9 +134,14 @@ type netProbe struct {
 
 // Validate does static validation of the configuration
 func (cfg OutputConfig) Validate() error {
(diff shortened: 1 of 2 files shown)


grantseltzer:
Thanks so much for this @josedonizetti! Test looks great. I think i'd prefer the (admittedly ugly) single if statement rather than an if and a switch. Perhaps we can format like this if you'd like: https://play.golang.org/p/NOvSKBnhCjI

josedonizetti:
@grantseltzer thanks for the feedback. Reverted it and formatted.

grantseltzer (approved):
LGTM! Thanks!

Issue #876, opened 2021-08-04
tracee-ebpf --output ignore flag


josedonizetti:
When using the capture command we still see the stream of events continuously printed to stdout (or another output), but for capturing one might not care about the events, only the artifacts captured. To ignore the events, we may redirect the output to `/dev/null` but it still generates IO.

I wonder if supporting an extra flag for a [printer](https://github.com/aquasecurity/tracee/blob/main/tracee-ebpf/tracee/printer.go#L17) that ignores the output, avoiding any extra work on it, would make sense. eg:

`tracee-ebpf --capture write=/home/user/test.txt* --output ignore`

(I'm willing to contribute a PR if you agree the change makes sense.)

grantseltzer:
I think this is a great idea @josedonizetti! Would be happy to accept a PR with this addition.

Issue #752, opened 2021-06-08
separate error/output printers


itaysk:
currently the output flag controls the format of both the events output and the error output. it doesn't make much sense, in example that -o gob will write gob formatted events to the output file, and also gob formatted errors into the standard error. for now I think we should always print errors to standard error in text format

PR #882, opened 2021-08-05
tracee-ebpf: add --capture ignore

josedonizetti:
Fixes https://github.com/aquasecurity/tracee/issues/876

### DIFF
--- a/tracee-ebpf/tracee/tracee.go
+++ b/tracee-ebpf/tracee/tracee.go
@@ -309,7 +309,8 @@ func New(cfg Config) (*Tracee, error) {
 		bootTime:  uint64(bootTime),
 	}
 	outf := os.Stdout
-	if t.config.Output.OutPath != "" {
+	if t.config.Output.OutPath != "" &&
+		t.config.Output.OutPath != os.DevNull {
 		dir := filepath.Dir(t.config.Output.OutPath)
 		os.MkdirAll(dir, 0755)
 		os.Remove(t.config.Output.OutPath)
@@ -330,7 +331,13 @@ func New(cfg Config) (*Tracee, error) {
 	}
 	ContainerMode := (t.config.Filter.ContFilter.Enabled && t.config.Filter.ContFilter.Value) ||
 		(t.config.Filter.NewContFilter.Enabled && t.config.Filter.NewContFilter.Value)
-	printObj, err := newEventPrinter(t.config.Output.Format, ContainerMode, t.config.Output.RelativeTime, outf, errf)
+
+	printerKind := t.config.Output.Format
+	if t.config.Output.OutPath == os.DevNull {
+		printerKind = "ignore"
+	}
+
+	printObj, err := newEventPrinter(printerKind, ContainerMode, t.config.Output.RelativeTime, outf, errf)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


tracee-ebpf/tracee/tracee.go
```diff
@@ -134,8 +134,13 @@ type netProbe struct {
 
 // Validate does static validation of the configuration
 func (cfg OutputConfig) Validate() error {
-	if cfg.Format != "table" && cfg.Format != "table-verbose" && cfg.Format != "json" && cfg.Format != "gob" && !strings.HasPrefix(cfg.Format, "gotemplate=") {
-		return fmt.Errorf("unrecognized output format: %s. Valid format values: 'table', 'table-verbose', 'json', 'gob' or 'gotemplate='. Use '--output help' for more info.", cfg.Format)
+	if cfg.Format != "table" &&
+		cfg.Format != "table-verbose" &&
+		cfg.Format != "json" &&
+		cfg.Format != "gob" &&
+		cfg.Format != "ignore" && // TODO: add to test after merging https://github.com/aquasecurity/tracee/pull/881
```
grantseltzer, line 141: We try to avoid TODO's in comments, better to just make a github issue

josedonizetti: @grantseltzer okay, but the PR was just merged, I’ll rebase and resolve the TODO. Thanks

itaysk (commented):
Thanks for this PR and your other recent ones as well!

We need to consider a few things here:
1. do we want to turn off output events and errors? I think just events
2. regardless of (1), this is not an output format, so the config is in the wrong place.

If we do want to toggle just the output events, I think it belongs under the "out-file" setting. If we want to toggle everything, then it belongs at the top level of "output".

josedonizetti:
@itaysk Flag moved to `out-file`, but it was looking a bit weird to pass `out-file:ignore`, so instead I changed it to be `out-file:/dev/null` it reads better considering the option has `file` on the name. Also, ignorePrinter is only ignoring events, not errors, but if in the future it becomes a need to also ignore errors, we could use the same syntax `err-file:/dev/null`.

itaysk:
I'm not sure about that.. using a real path name as a special option that changes the behavior does not sound best to me (others please chime in). I'd prefer `-o out-file:none` if we are using the out-file argument.

But after thinking about this abit, I think that the `--output` flag should just control the main output - stdoud, and not the errors. If I'm consuming events over a pipe with gob, it doesn't mean I want the errors in gob as well. Errors are usually consumed by a different person, tool, etc and should also be printed to stderr as "regular" text messages. This could be done in a separate PR, but it means that in this PR we can choose to do `-o none` which will read better for you

josedonizetti:
@itaysk about printing to stdout/stderr, there is a separate issue to fix it https://github.com/aquasecurity/tracee/issues/752, what I didn’t understand is if you prefer ‘-o none’ or ‘-o out-file:none’

itaysk:
>I didn’t understand is if you prefer ‘-o none’ or ‘-o out-file:none’

`-o none`

grantseltzer:
I'm going to have to agree that  `-o out-file:/dev/null` and `-o none` should be two separate things. What if someone is using tracee and wants to see the writes without saving to disk?

yanivagman (commented):
LGTM
Can you please also add this option to the docs (in tracee-ebpf/output.md)?

yanivagman (approved):

Issue #736, opened 2021-05-25
Sending a SIGINT to Tracee container does not finish it


simar7:
### What I observed
```
first shell:
docker run --name tracee --rm --privileged -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro -v /tmp/tracee:/tmp/tracee aquasec/tracee:latest

second shell:
docker kill --signal=SIGINT tracee
```
Tracee container does not stop and keeps tracing

### What I expected
Tracee container stops, prints stats and gracefully exits upon receiving SIGINT

### My thoughts
Bash does not deliver the signal to Tracee running [0] within the container [1] Instead we could launch it with an exec prefix call and avoid bash altogether [2] 

### Open questions
What happens when we pipe tracee-ebpf output to tracee-rules, does the above approach still hold?

### References:
[0] https://github.com/aquasecurity/tracee/blob/main/entrypoint.sh#L9
[1] https://unix.stackexchange.com/a/196053
[2] https://hynek.me/articles/docker-signals/

PR #883, opened 2021-08-05
tracee: add tini tracee docker image

josedonizetti:
Fixes https://github.com/aquasecurity/tracee/issues/736. If adding a dependency is not a problem, [tini](https://github.com/krallin/tini) is a very common solution to handle signals properly on containers.

The image size doesn't  change:
```
tracee:with_tini         446MB
tracee:without_tini      446MB
```

Testing:
```
# first shell
# build trace image
docker build  -t tracee:with_tini .
# run tracee
docker run -v /boot/config-$(uname -r):/boot/config-$(uname -r) --name tracee --rm --privileged -it tracee:with_ini

# second shell
docker kill --signal=SIGINT tracee
# or
docker kill --signal=SIGTERM tracee
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,7 @@ RUN make
 
 # base image for tracee which includes all tools to build the bpf object at runtime
 FROM alpine as fat
-RUN apk --no-cache update && apk --no-cache add clang llvm make gcc libc6-compat coreutils linux-headers musl-dev elfutils-dev libelf-static zlib-static
+RUN apk --no-cache update && apk --no-cache add clang llvm make gcc libc6-compat coreutils linux-headers musl-dev elfutils-dev libelf-static zlib-static tini
 
 # base image for tracee which includes minimal dependencies and expects the bpf object to be provided at runtime
 FROM alpine as slim
@@ -39,4 +39,4 @@ LABEL org.label-schema.build-date=$BUILD_DATE \
     org.label-schema.vendor="Aqua Security" \
     org.label-schema.version=$VERSION
 
-ENTRYPOINT ["./entrypoint.sh"]
+ENTRYPOINT ["/sbin/tini", "-g", "--", "./entrypoint.sh"]


josedonizetti:
@simar7 Let me know what you think please.

simar7:
@josedonizetti we've already added an `exec` directive to launch so I don't think this is necessary as signals are handled appropriately. https://github.com/aquasecurity/tracee/blob/main/entrypoint.sh#L9

Did I miss something?

josedonizetti:
@simar7 well, maybe I am missing something, but testing here in two different computers (ubuntu 18.04, and 20.04) with `aquasec/tracee:latest` ( i have delete the image to have docker pull latest) and `SIGINT` is only handled if I hit `CTRL+C`, if I send a `SIGINT` or `SIGTERM` via `docker kill` or `kill` the container doesn't gracefully exit.

itaysk:
Only the "trace" subcommand is using `exec`, and therefore handled appropriately. the "regular" tracee cannot leverage `exec` since it's not a single command (<- currently. we do have plans to make it a single process).

Here's my data for the "regular" tracee

method | explanation | did it quit
--- | --- | ---
CTRL+C (in the terminal when running with `-it`) | send INT to foreground group, i.e to bash, tracee-ebpf, tracee-rules individually | yes 
kill -INT (to the container's root PID) | send SIGINT to bash | no
kill -TERM (to the container's root PID) |  send SIGTERM to bash) | yes (actually I don't understand why)
docker stop | send SIGTERM to the container's root PID | yes

itaysk:
anyway, I think it's fine to add tini to the container to get more predictable experience. @simar7 do you have another idea to fix the issue?
When we merge the processes we can reconsider if we need it.

josedonizetti:
@itaysk if you prefer I can investigate a little further if we can accomplish the same without `tini`. About your tests results, at least for me `docker stop` is also not working.

Docker will first attempt a SIGTERM, but if the container doesn't respect it, it forces a SIGKILL. 

Eg:
```
$ docker run --name tracee --rm --privileged -it tracee:with_tini
$ echo $?
143 # using tini, SIGTERM was respected

$ docker run --name tracee --rm --privileged -it aquasec/tracee:latest
$ echo $?
137 # without tini, a SIGKILL was forced here
```

simar7:
> anyway, I think it's fine to add tini to the container to get more predictable experience. @simar7 do you have another idea to fix the issue?
> When we merge the processes we can reconsider if we need it.

Docker 1.13 and above should have the signal handling functionally if we pass in `--init` flag. See below for example:
```
vagrant@ubuntu-groovy:~/linux$ docker run -it --name  tracee --rm --privileged -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro -v /tmp/tracee:/tmp/tracee aquasec/tracee:0.5.2

vagrant@ubuntu-groovy:~/linux$ echo $?
137

```

Now with `--init` flag:
```
vagrant@ubuntu-groovy:~/linux$ docker run --init -it --name  tracee --rm --privileged -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro -v /tmp/tracee:/tmp/tracee aquasec/tracee:0.5.2

vagrant@ubuntu-groovy:~/linux$ echo $?
143
```

Ref: https://docs.docker.com/engine/reference/commandline/run/

We could include `tini` instead of adding this to our docs. I just err on the side of including less dependencies if we can. 

Thoughts?

josedonizetti:
@simar7 Yes, `--init` is a good option, specially because it is  `tini` that docker uses to support it. But I wonder if this isn't too specific for docker only, it seems podman supports `--init` too, but I don't think cri-o and containerd do. Another consideration, for anyone running tracee as a k8s  dameonset, will also not handle SIGTERM, and k8s forcing a SIGKILL during a rolling update.

simar7 (approved):
Fair enough - sounds like there are enough reasons to have it then to not. I'll let @itaysk give it another read but lgtm

itaysk (approved):
Not sure if we should removed the exec from trace sub command

Issue #851, opened 2021-07-29
failed to attach to raw tracepoint 'sys_enter': invalid argument in kernel 5.4


rafaeldtinoco:
When developing external BTF feature I faced:

![image](https://user-images.githubusercontent.com/7395852/127429772-29bd73ff-8acb-40bf-a0cd-8301cf753dcc.png)

when trying to run tracee in a 5.4.0-80-generic kernel and the BTF file provided by btfhub. This issue is for me to investigate this further in a near future time. The same does not happen in a 5.8 kernel running on the host of the same docker container:

![image](https://user-images.githubusercontent.com/7395852/127430081-b2e92aa9-c826-4786-819c-71f0a82c5845.png)

rafaeldtinoco:
BTW I don't think external BTF has anything to do with this but haven't checked it yet (as I'm finishing the BTF from btfhub feature).

rafaeldtinoco:
```
$ sudo ./dist/tracee-ebpf --debug --trace pid=new
BTF is not enabled
attempting to build the bpf object file
building bpf object in: /tmp/tracee-make736244672
[...]

PR #890, opened 2021-08-09
tracee-ebpf: sys_enter: use temp args_t struct pointer for no args

rafaeldtinoco:
Fixes: #851

Somehow eBPF verifier did not like CO-RE object txt using the stack
address pointer from args_tmp as pointer for save_args() inline
function:

    libbpf: prog 'tracepoint__raw_syscalls__sys_enter': failed to
    attach to raw tracepoint 'sys_enter': Invalid argument

This commit creates a temporary args_t structure in the stack just
before calling save_args() and passing the structure pointer as an
argument. That is enough to allow the eBPF program to be loaded.

### DIFF
--- a/tracee-ebpf/tracee/tracee.bpf.c
+++ b/tracee-ebpf/tracee/tracee.bpf.c
@@ -2169,8 +2169,9 @@ if (CONFIG_ARCH_HAS_SYSCALL_WRAPPER) {
     // exit, exit_group and rt_sigreturn syscalls don't return - don't save args for them
     if (id != SYS_EXIT && id != SYS_EXIT_GROUP && id != SYS_RT_SIGRETURN) {
         // save the timestamp at function entry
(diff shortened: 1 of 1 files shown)


yanivagman (changes requested):
Doing this change results in zero arguments when running tracee (all syscalls arguments equal zero)

rafaeldtinoco:
> Doing this change results in zero arguments when running tracee (all syscalls arguments equal zero)

Stupid me.. I was reading == instead of != in:


```
    if (id != SYS_EXIT && id != SYS_EXIT_GROUP && id != SYS_RT_SIGRETURN) {
```

that is the reason of confusion =). Anyway, let me continue digging into it then.

rafaeldtinoco:
No no no, still pursuing the issue, sorry for the noise in this one.

Issue #888, opened 2021-08-08
Fix minimum required capabilities


josedonizetti:
Tracee should verify the proper required capabilities to allow a user to run without privilege. Currently tracee checks only for [`CAP_SYS_ADMIN`](https://github.com/aquasecurity/tracee/blob/main/tracee-ebpf/main.go#L970), but if we try to run either the direct process, or a docker container we get:

```
docker run --na tracee --rm --cap-add CAP_SYS_ADMIN -it aquasec/tracee:latest
libbpf: failed to mmap perf buffer on cpu #7: Operation not permitted
libbpf: failed to mmap perf buffer on cpu #7: Operation not permitted
fatal error: unexpected signal during runtime execution
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x1 addr=0x3f pc=0x7133d3]

runtime stack:
runtime.throw(0x7e9f70, 0x2a)
	/usr/local/go/src/runtime/panic.go:1117 +0x72
runtime.sigpanic()
	/usr/local/go/src/runtime/signal_unix.go:718 +0x2e5
```

Libbpf doesn't have the proper permissions. This also triggers a panic on golang, because libbpfgo doesn't handle the error properbly, but I have fixed this issue -> https://github.com/aquasecurity/libbpfgo/pull/49

To allow either a docker container or a process on the host to run, tracee also needs `CAP_IPC_LOCK`. 

eg:
```
docker run  --name tracee --rm --cap-add CAP_SYS_ADMIN --cap-add CAP_IPC_LOCK -it aquasec/tracee:latest
Loaded signature(s):  [TRC-1 TRC-2 TRC-3 TRC-4 TRC-5 TRC-6 TRC-7]
```

If you agree, I'm willing to make a PR to fix it.

josedonizetti:
@yanivagman Let me know what you think, pls.

yanivagman:
@josedonizetti agreed - many thanks for all your recent contributions!

Issue #747, opened 2021-06-05
Update checkRequiredCapabilities to support newer kernels


yanivagman:
We currently only check for CAP_SYS_ADMIN when running Tracee (https://github.com/aquasecurity/tracee/blob/main/tracee-ebpf/main.go#L885), which was correct for old kernels.
For newer kernels, CAP_SYS_ADMIN was split into combination of 3 other capabilities:
CAP_BPF, CAP_PERFMON and CAP_NET_ADMIN as described in https://lwn.net/Articles/820560/, and https://lwn.net/Articles/822362/
Update checkRequiredCapabilities() to support these newer kernels.

Also related: #286

rajibmitra:
would it be ok if I take this one @itaysk  @yanivagman

yanivagman:
@rajibmitra of course, please do

cdelzotti:
Hello ! It's me again :)

I see that this issue has been opened for a while, can I be assigned to it ?

yanivagman:
This is related to #1202 . I know that @AlonZivony tried to tackle this there, but we found out that it is more tricky than it may seem - @AlonZivony can you please relate to this?

AlonZivony:
So this one is a duplicate of #1815.
The issue here was that some distros didn't work without `CAP_SYS_ADMIN` event though it should work with their kernel version.
To do this bugfix, the reason for the error in #1803 should be resolved.
I don't have the time for it right now, so if @cdelzotti want to solve it, it would be great :)

cdelzotti:
Of course, thanks. I'll be away for a week but I'll check it out when I come back.

yanivagman:
I'm closing this one in favor of #1815

PR #891, opened 2021-08-09
tracee-ebpf: fix capabilities minimum requirements

josedonizetti:
Fixes https://github.com/aquasecurity/tracee/issues/888

Testing:
```
# on a binary
sudo setcap cap_ipc_lock,cap_sys_admin=pe tracee-ebpf
./tracee-ebpf # without sudo

# docker
# build tracee-ebpf docker images
docker run --cap-add CAP_SYS_ADMIN --cap-add CAP_IPC_LOCK  --rm -it tracee-ebpf:latest
```

I have not changed any documentation, I would prefer to show examples using capabilities than `--privileged`, but perhaps before fixing the documentation I could before fix issue https://github.com/aquasecurity/tracee/issues/747

### DIFF
--- a/tracee-ebpf/main.go
+++ b/tracee-ebpf/main.go
@@ -124,8 +124,9 @@ func main() {
 			cfg.BPFObjPath = bpfFilePath
 			cfg.BPFObjBytes = bpfBytes
 
-			if !checkRequiredCapabilities() {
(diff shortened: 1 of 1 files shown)


yanivagman (approved):
LGTM!

Issue #780, opened 2021-06-20
helpers.rego and helpers.go should be loaded dynamically 


idanr1986:

eidans:
@itaysk , do we have an ETA for this?

itaysk:
It's not in the backlog yet. needs to be discussed further as I don't think the use-case is obvious from the title, and then prioritized. can you elaborate on the use-case?

eidans:
We want to load go signatures that are not compiled with tracee-rules
binary. As I understand, it is not possible today as the helpers file is
statically compiles with tracee-rules.

itaysk:
> We want to load go signatures that are not compiled with tracee-rules
binary. As I understand, it is not possible today as the helpers file is
statically compiles with tracee-rules.

It's true that helpers.go is compiled with tracee-rules. that's good because it makes it available to everyone to use. But that doesn't have anything to do with your ability to dynamically load go signatures - that's possible today, just build the policies and put the file in the rules directory and tracee-rules will discover and load it.
Let me know if I missed something

itaysk:
To address the original request, I think we modify the helpers loading logic. Today we specifically look for helpers.rego and load it to all the rego instances. We can instead look for `*_helpers.rego` which will allow anyone to write their own helpers file which will be available to all the signatures.

PR #893, opened 2021-08-09
tracee-ebpf: add support to custom rego helpers

josedonizetti:
Fixes https://github.com/aquasecurity/tracee/issues/780

### DIFF
--- a/tracee-rules/signature.go
+++ b/tracee-rules/signature.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"plugin"
+	"strings"
 
 	"github.com/aquasecurity/tracee/tracee-rules/signatures/rego/regosig"
 	"github.com/aquasecurity/tracee/tracee-rules/types"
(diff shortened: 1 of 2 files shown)


tracee-rules/signature.go
```diff
@@ -83,19 +84,35 @@ func findRegoSigs(dir string) ([]types.Signature, error) {
 
 	var res []types.Signature
 
+	regoHelpers := []string{regoHelpersCode}
 	for _, file := range files {
-		if filepath.Ext(file.Name()) != ".rego" {
+		if file.Name() == "helpers.rego" {
 			continue
 		}
```
itaysk, line 91: seems like this test is included in L93, no?

josedonizetti: L93 checks if a file is **not** a helper, we don’t process it. But ‘helpers.go’ is a valid helper, but we don’t want to process it, because it is already embedded.

itaysk: right, sorry. 
I think we should not skip helpers.rego actually because it's embedded (that was the motivation to embed it). but I'll live that for a future PR maybe

itaysk (approved):

josedonizetti:
@itaysk Later when you have some free time, do you mind extending on your comment?
```
I think we should not skip helpers.rego actually because it's embedded (that was the motivation to embed it). but I'll live that for a future PR maybe
```

Not sure I understood, because the code for helpers.go already exists embedded, so we only add it to the [helpers slice](https://github.com/aquasecurity/tracee/blob/main/tracee-rules/signature.go#L87), if we re-read the `helpers.go` it will be overridden, is that the expected behavior?

itaysk:
You're right that was too vague :)
I wrote the answer in another issue, so we can track it as well https://github.com/aquasecurity/tracee/issues/902

Issue #835, opened 2021-07-23
Export system/kernel exploring functions


mtcherni95:
As part of integrating Tracee as a library we would like to have the ability to use functions like 'missingKernelConfigOptions', 'checkRequiredCapabilities' etc.

grantseltzer:
We can potentially put these in the helpers sub-package of libbpfgo

mtcherni95:
In particular: we would like to have the following APIs:
- GetKernelAndDistroVersion
- GetTraceeVersion
- LoadObjectFromPath

This will allow us to:
- Skip steps that are part of the init of Tracee that we don’t do in our app such as compiling eBPF locally with kernel headers.

itaysk:
Tracee version is not so easy to reason about, we could print the version but in your case it might be incorrect. Unlike the other runtime properties, this is a compile-time property. Since in your case the compilation is under your control (you're using Tracee as a library and compiling a different artifact then tracee-ebpf), then it's up to you to embed the correct version. Another alternative is that you inspect the module version at runtime (IDK if it's possible)

itaysk:
Thinking of this again.. Correct me if I'm wrong but I think the reason why you need the Tracee version is only so you can identify a compatible object file. This is a use case that we can and should support, even without exporting the Tracee version. I propose we start versioning the ABI between the eBPF probe and the userspace program. It doesn't have to align with Tracee versions. Actually, in most cases it wouldn't, which means we can reuse the probe in more versions.
This is important not only for this use case, but also for potentially BTF Hub. I'll start a new issue to track it.

yanivagman:
> In particular, we would like to have the following APIs:
> 
> GetKernelAndDistroVersion
> GetTraceeVersion
> LoadObjectFromPath
> This will allow us to:
> 
> Skip steps that are part of the init of Tracee that we don’t do in our app such as compiling eBPF locally with kernel headers.

**GetKernelAndDistroVersion** - Nothing special here - just a simple uname. I don't think that this is something we should export.
**LoadObjectFromPath** - You can already provide an environment variable to set a path for an object, so consider this already "exported".
**GetTraceeVersion** - See my comment in #839

itaysk:
closed by: https://github.com/aquasecurity/libbpfgo/pull/53

Issue #833, opened 2021-07-21
Implement BTF support, BTF & BPF decision path for CO-RE and non CO-RE BPF objects execution


grantseltzer:
Currently (or once #818 is merged) the following logic is followed for deciding  which bpf object to use:

1. check for /sys/kernel/btf
2. if yes, use CO-RE binary
3. if no, look for BPF.o for running kernel
4. if no BPF.o, compile BPF.o for running kernel

Once we have BTF Hub the following logic should be implemented:

- CO-RE phase:

1. check for /sys/kernel/btf.
2. if yes, use CO-RE binary
3. if no sysfs btf, look for BPF support in running kernel.
4. if BPF support, get BTF for running kernel from BTFHUB
5. if BTFHUB file exists use CO-RE binary with downloaded/cached BTF file

- Legacy phase:

6. if BTFHUB file does not exist look for BPF.o for running kernel
4. if no, compile BPF.o for running kernel

_See comment posted by @rafaeldtinoco in https://github.com/aquasecurity/tracee/issues/818#issuecomment-883759923_

rafaeldtinoco:
I'm currently working on this. Scratch for [btfhub retrieval and distro discovery](https://gist.github.com/rafaeldtinoco/3e7cdff16cccbcffc45d0cacaff61f01) to be included in tracee.

itaysk:
Should add that if user explicitly provides BPF file (TRACEE_BPF_FILE) it overrides all defaults.

itaysk:
this might be changed based on discussion in #839 
The possible change is to remove step legacy-6.
Also, I presume that we no longer for for object file that is next to the binary (same dir)

rafaeldtinoco:
I haven't read #839 discussion yet (will do it tomorrow my morning) but wanted to put WIP here based on previous talk with @itaysk about decisions based on existing tracee environment:

I'll try to read #839 tomorrow morning, for now WIP on decision making on startup:

https://github.com/rafaeldtinoco/tracee/commit/d23c48e6a7678f0b79a4465e58230bb7a0f3e400

rafaeldtinoco:
This is being implemented at:

rafaeldtinoco:
BTW, this issue is to track the full BTFHUB API usage as, from recent merges, you will see parts of BTF support are already being merged here and there (because of CNDR needs). With that said, I'll re-work 'external-btf' so it uses a BTF API that we will agree on.

PR #895, opened 2021-08-10
tracee: add btfinfo initial support

rafaeldtinoco:
Initial BTFInfo object for subsequent commits adding BTF support.

### DIFF
--- a/tracee-ebpf/btfinfo.go
+++ b/tracee-ebpf/btfinfo.go
@@ -1,0 +1,110 @@
+package main
+
+import (
+	"bufio"
+	"errors"
+	"github.com/aquasecurity/tracee/tracee-ebpf/tracee"
+	"os"
+	"strings"
+)
+
+func fileExists(path string) bool {
+	if _, err := os.Stat(path); err == nil {
+		return true
+	}
+	return false
+}
+
+func fileGrepUnique(path string, needle string) string {
+	file, _ := os.Open(path)
+	defer func(file *os.File) {
+		_ = file.Close()
+	}(file)
+	scanner := bufio.NewScanner(file)
+	for i := 1; scanner.Scan(); i++ {
+		if strings.Contains(scanner.Text(), needle) {
+			return scanner.Text()
+		}
+	}
+	return ""
+}
+
+// NewBTFInfo creates a BTFInfo object and runs DiscoverDistro()
+// on its creation. It returns nil if no discovery data was
+// avalable. To get specific error from DiscoverDistro method,
+// user should create a BTFInfo struct and call DiscoverDistro
+// by itself.
+func NewBTFInfo() *BTFInfo {
+	btfi := new(BTFInfo)
+	if err := btfi.DiscoverDistro(); err != nil {
+		return nil
+	}
+	return btfi
+}
+
+// BTFInfo object contains all BTF relevant information so user
+// is able to not only discover underlaying Linux distribution
+// but also, in future, manage acquisition & cache of external
+// BTF files to be used by the BPF CO-RE objects.
+type BTFInfo struct {
+	distroId     string
+	distroVer    string
+	distroKernel string
+}
+
+func (btfi *BTFInfo) GetDistroId() string {
+	return btfi.distroId
+}
+
+func (btfi *BTFInfo) GetDistroVer() string {
+	return btfi.distroVer
+}
+
+func (btfi *BTFInfo) GetDistroKernel() string {
+	return btfi.distroKernel
+}
+
+// DiscoverDistro tries to discover running Linux distribution either by reading
+// /etc/os-releases (https://man7.org/linux/man-pages/man5/os-release.5.html) or
+// by interpreting the version out of uname(2) system call.
+func (btfi *BTFInfo) DiscoverDistro() error {
+
+	ker := tracee.UnameRelease()
+	if ker == "" {
+		return errors.New("could not get uname")
+	}
+
+	osrelease := "/etc/os-release"
+	if fileExists(osrelease) {
+		id := fileGrepUnique(osrelease, "ID=")
+		id = id[strings.Index(id, "=")+1:]
+		id = strings.ReplaceAll(id, "\"", "")
+		ver := fileGrepUnique(osrelease, "VERSION_ID=")
+		ver = ver[strings.Index(ver, "=")+1:]
+		ver = strings.ReplaceAll(ver, "\"", "")
+		btfi.distroId = id
+		btfi.distroVer = ver
+		btfi.distroKernel = ker
+		return nil
+	}
+
+	// real examples:
+	// fedora:      4.18.16-300.fc29
+	// centos:      3.10.0-1160.31.1.el7.centos.x86_64
+	if strings.Contains(ker, "fc") {
+		btfi.distroId = "fedora"
+		btfi.distroVer = ker[strings.Index(ker, "fc")+2:]
+		btfi.distroKernel = ker[:strings.Index(ker, ".fc")]
+		return nil
+	} else if strings.Contains(ker, "centos") {
+		btfi.distroId = "centos"
+		t := ker
+		t = t[strings.Index(t, "el")+2:]
+		t = t[:strings.Index(t, ".")]
+		btfi.distroVer = t
+		btfi.distroKernel = ker[:strings.Index(ker, ".el")]
+		return nil
+	} // else if may continue...
+
+	return errors.New("distribution could not be discovered")
+}


rafaeldtinoco:
@mtcherni95 as we've spoke, this is the initial BTFinfo support you've asked. You may use parts of the following repository:

https://github.com/rafaeldtinoco/tracee/tree/external-btf

if you're willing to have external BTF files feature, just like we've spoken. Let me know if there is anything you need from the upstream repo in order to have your requirements satisfied.

@itaysk FYIO ^

tracee-ebpf/btfinfo.go
```diff
@@ -0,0 +1,126 @@
...
+	for i, b := range all {
+		temp[i] = byte(b)
+	}
+	conv := string(temp[:])
+	conv = conv[:strings.Index(conv, "\x00")]
+
+	return conv
+}
+
+func KernelVersion() string {
```
mtcherni95, line 42: shouldn't we return error as well?

tracee-ebpf/btfinfo.go
```diff
@@ -0,0 +1,126 @@
...
+	if btfi.distrokernel == "" {
+		return "", errors.New("should call discover first")
+	}
+	return btfi.distrokernel, nil
+}
+
+func (btfi *BTFinfo) DiscoverDistroByKernel() error {
+	// tries to discover running distribution by kernel version/name
+
+	kernel := KernelVersion()
```
mtcherni95, line 84: might be useful to inform user if we didn't manage to get the kernel version, WDYT?

mtcherni95:
> @mtcherni95 as we've spoke, this is the initial BTFinfo support you've asked. You may use parts of the following repository:
> 
> https://github.com/rafaeldtinoco/tracee/tree/external-btf
> 
> if you're willing to have external BTF files feature, just like we've spoken. Let me know if there is anything you need from the upstream repo in order to have your requirements satisfied.
> 
> @itaysk FYIO ^

Thanks a lot @rafaeldtinoco!

tracee-ebpf/btfinfo.go
rafaeldtinoco: Hum, only error here would come from:

	err := syscall.Uname(&utsname)

and I had panic() before. Only error for uname(2) is EFAULT if buffer is not valid.. but we're for sure allocating it from the function stack.. so I think we are good.. I'll let you resolve this if you agree, if not I can change.

tracee-ebpf/btfinfo.go
rafaeldtinoco: It goes with the first answer. All ears! Thank for reviewing this.

tracee-ebpf/btfinfo.go
```diff
@@ -0,0 +1,126 @@
...
+	err := syscall.Uname(&utsname)
+	if err != nil {
+		return ""
+	}
+
+	return ByteToString(utsname.Release)
+}
+
+// BTFinfo
+
```
grantseltzer, line 53:
```suggestion
```

tracee-ebpf/btfinfo.go
grantseltzer, line 42: I don't think this needs to be exported/capitalized

tracee-ebpf/btfinfo.go
```diff
@@ -0,0 +1,126 @@
...
+	for i := 1; scanner.Scan(); i++ {
+		if strings.Contains(scanner.Text(), needle) {
+			return scanner.Text()
+		}
+	}
+
+	return ""
+}
+
+func ByteToString(all [65]int8) string {
```
grantseltzer, line 31: I don't think this needs to be exported/capitalized

tracee-ebpf/btfinfo.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+// Helpers
+
+func FileExists(path string) bool {
+	_, err := os.Stat(path)
+	return err == nil
+}
+
+func FileGrepUnique(path string, needle string) string {
```
grantseltzer, line 18: I don't think this needs to be exported/capitalized

tracee-ebpf/btfinfo.go
```diff
@@ -0,0 +1,126 @@
...
+	"bufio"
+	"errors"
+	"os"
+	"strings"
+	"syscall"
+)
+
+// Helpers
+
+func FileExists(path string) bool {
```
grantseltzer, line 13: I don't think this needs to be exported/capitalized

tracee-ebpf/btfinfo.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (btfi *BTFinfo) GetDistroKernel() (string, error) {
+	if btfi.distrokernel == "" {
+		return "", errors.New("should call discover first")
+	}
+	return btfi.distrokernel, nil
+}
+
+func (btfi *BTFinfo) DiscoverDistroByKernel() error {
+	// tries to discover running distribution by kernel version/name
```
grantseltzer, line 82: You can put this in a GoDoc comment

tracee-ebpf/btfinfo.go
```diff
@@ -0,0 +1,126 @@
...
+
+	kernel := KernelVersion()
+	// fedora:      4.18.16-300.fc29
+	// centos:      3.10.0-1160.31.1.el7.centos.x86_64
```
grantseltzer, line 86: errant comments?

tracee-ebpf/btfinfo.go
```diff
@@ -0,0 +1,126 @@
...
+
+		return nil
+	} // else if ...
+
+	return errors.New("can't discover distribution by kernel version")
+}
+
+func (btfi *BTFinfo) DiscoverDistroByRelease() error {
+	// tries to discover distribution based on the os-release file:
+	// https://man7.org/linux/man-pages/man5/os-release.5.html
```
grantseltzer, line 108: This would also make a good godoc comment

grantseltzer (changes requested):
What do you think about, instead of needing to call a `Discover` method, have a single function for creating the BTFinfo struct which attempts to do one, and does the other if the first one fails?

tracee-ebpf/btfinfo.go
```diff
@@ -0,0 +1,126 @@
...
+import (
+	"bufio"
+	"errors"
+	"os"
+	"strings"
+	"syscall"
+)
+
+// Helpers
+
```
grantseltzer, line 12:
```suggestion
```

tracee-ebpf/btfinfo.go
```diff
@@ -0,0 +1,126 @@
...
+		btfi.distrover = t
+		btfi.distrokernel = kernel[:strings.Index(kernel, ".el")]
+
+		return nil
+	} // else if ...
+
+	return errors.New("can't discover distribution by kernel version")
+}
+
+func (btfi *BTFinfo) DiscoverDistroByRelease() error {
```
yanivagman, line 106: Instead of creating different discover methods, I would create just one Discover method, and try different methods until a match is found.
This function can have the following signature:
`func Discover () (BTFInfo, error)`
Then, if err=nil, the user can just extract the data from (the now exported) BTFInfo fields:
btfi.DistroVer, btfi.DistroID, btfi.DistroKernel.
WDYT?

tracee-ebpf/btfinfo.go
rafaeldtinoco: oh, ok, let me read about it and will do.

tracee-ebpf/btfinfo.go
rafaeldtinoco: Thought about having real examples.

rafaeldtinoco:
> What do you think about, instead of needing to call a `Discover` method, have a single function for creating the BTFinfo struct which attempts to do one, and does the other if the first one fails?

Ok, Yaniv also suggested that i glued the those discovery methods. Will do that.

tracee-ebpf/btfinfo.go
rafaeldtinoco: I'll mark this as resolved because Yaniv mentioned existing function: func UnameRelease() which I'll be using.

tracee-ebpf/btfinfo.go
rafaeldtinoco: Same here.

tracee-ebpf/btfinfo.go
rafaeldtinoco: Yes, I think it's ok doing that now. This was a subset of a major change from: 

https://github.com/rafaeldtinoco/tracee/tree/external-btf 

which contained an Object and more methods (to get and extract BTF files from the HUB). Intention was to isolate the BTFinfo logic, and then in near future start isolating things in a more objectification approach for existing code. I'm ok in having this as a single function now and changing later if needed (or keeping as is).

rafaeldtinoco: Actually, if you don't mind, I think I'll keep this as is (as an object). Just because my parallel logic (from the external-btf branch) will use it as an object for now and I dont want to re-write it. Let me know if you have hard preference on this and I can change it.

rafaeldtinoco: > Instead of creating different discover methods, I would create just one Discover method, and try different methods until a match is found.

I have done that though. Just 1 discovery method.. I believe this is better also.

rafaeldtinoco:
```go
// NewBTFInfo creates a BTFInfo object and runs DiscoverDistro()
// on its creation. It returns nil if no discovery data was
// avalable. To get specific error from DiscoverDistro method,
// user should create a BTFInfo struct and call DiscoverDistro
// by itself.
func NewBTFInfo() *BTFInfo {
	btfi := new(BTFInfo)
	if err := btfi.DiscoverDistro(); err != nil {
		return nil
	}
	return btfi
}
```

So using it might be as simple as:

```go
	btfi := NewBTFInfo()
	if btfi == nil {
		return errors.New("BTFInfo failed")
	}
	fmt.Println(btfi.GetDistroId()
```

tracee-ebpf/btfinfo.go
```diff
@@ -0,0 +1,108 @@
...
+// /etc/os-releases (https://man7.org/linux/man-pages/man5/os-release.5.html) or
+// by interpreting the version out of uname(2) system call.
+func (btfi *BTFInfo) DiscoverDistro() error {
+
+	ker := tracee.UnameRelease()
+	if ker == "" {
+		return errors.New("could not get uname")
+	}
+
+	osrelease := "/etc/os-release"
```
grantseltzer, line 75: Do all distros with `/etc/os-release` use this same format? I can confirm for Fedora :-)

tracee-ebpf/btfinfo.go
```diff
@@ -0,0 +1,108 @@
...
+		t := ker
+		t = t[strings.Index(t, "el")+2:]
+		t = t[:strings.Index(t, ".")]
+		btfi.distroVer = t
+		btfi.distroKernel = ker[:strings.Index(ker, ".el")]
+		return nil
+	} // else if may continue...
+
+	return errors.New("distribution could not be discovered")
+}
```
grantseltzer, line 108:
```suggestion
}

```

tracee-ebpf/btfinfo.go
```diff
@@ -0,0 +1,108 @@
...
+	btfi := new(BTFInfo)
+	if err := btfi.DiscoverDistro(); err != nil {
+		return nil
+	}
+	return btfi
+}
+
+// BTFInfo
+
+type BTFInfo struct {
```
grantseltzer, line 47: One line away from being a GoDoc comment :D

tracee-ebpf/btfinfo.go
rafaeldtinoco: http://0pointer.de/blog/projects/os-release

This is the announce that says systemd stopped supporting distros that does not implement it AND it is a standard by freedesktop iirc. 

Nevertheless, idea is to either expand discovery function or to have different discovery mechanisms ready if needed. For initial BTF support I think this is enough as we can always expand this later.

tracee-ebpf/btfinfo.go
rafaeldtinoco: Marking as resolved with my last push, please let me know if you prefer it changed. I'll expand the object with logic from external-btf branch in my repository. Thanks!

itaysk:
This seems to be more relevant to #835 than #833. Having a utility function to discover OS-level information is something we wanted to export a while ago, also from a refactoring perspective. Do you think it makes sense to reframe this PR as such? possibly also move this code to libbpfgo.

rafaeldtinoco:
> This seems to be more relevant to #835 than #833. Having a utility function to discover OS-level information is something we wanted to export a while ago, also from a refactoring perspective. Do you think it makes sense to reframe this PR as such? possibly also move this code to libbpfgo.

I think so. It may be more relevant now to #835 indeed as it became just a subset of what it was in my initial patch (with the entire btf download logic). It may sound weird, but.. shouldn't libbpfgo be a 1:1 to libbpf upstream (but the go version) ? With that said, what I mean is.. what about having a 'glue' library for both responsible for everything that isn't libbpf or tracee related in their nature ?

grantseltzer:
> > This seems to be more relevant to #835 than #833. Having a utility function to discover OS-level information is something we wanted to export a while ago, also from a refactoring perspective. Do you think it makes sense to reframe this PR as such? possibly also move this code to libbpfgo.
> 
> I think so. It may be more relevant now to #835 indeed as it became just a subset of what it was in my initial patch (with the entire btf download logic). It may sound weird, but.. shouldn't libbpfgo be a 1:1 to libbpf upstream (but the go version) ? With that said, what I mean is.. what about having a 'glue' library for both responsible for everything that isn't libbpf or tracee related in their nature ?

That'd be the helpers package in libbpfgo https://github.com/aquasecurity/libbpfgo/tree/main/helpers

rafaeldtinoco:
Okay, I'm closing this PR then and will open a new one for libbpfgo. This new PR, together with https://github.com/aquasecurity/libbpfgo/pull/50, will be what CNDR currently needs to be in upstream repos (for now). @grantseltzer if you could review both I would appreciate. Thank you.

itaysk:
> what about having a 'glue' library for both responsible for everything that isn't libbpf or tracee related in their nature ?

This came up in the past a few times. we are incrementally accumulating general linux helpers in tracee and libbpfgo. If I had to choose between those two I'd choose libbfgo, but a 3rd library (linux-helpers) is also fine with me.

rafaeldtinoco:
> > what about having a 'glue' library for both responsible for everything that isn't libbpf or tracee related in their nature ?
> 
> This came up in the past a few times. we are incrementally accumulating general linux helpers in tracee and libbpfgo. If I had to choose between those two I'd choose libbfgo, but a 3rd library (linux-helpers) is also fine with me.

Alright, like we spoke today.. and for @grantseltzer awareness... idea we had (based on this discussion also) was to merge this one into libbpfgo helpers now and open an issue to spin off (at least analyze) libbpfgo helpers into a possible 3rd lib for helpers only (that can be shared among the two).

rafaeldtinoco:
Alright, this pull request moved to: https://github.com/aquasecurity/libbpfgo/pull/53 as a libbpfgo helper.

PR #896, opened 2021-08-10
Fix CO:RE support for RHEL 8 and derivatives

NeonSludge:
This PR fixes CO:RE support in `tracee-ebpf` on RHEL 8 and its derivatives.
RedHat kernel support was added some time ago by #351 but additional changes are now needed to make that fix work for CO:RE since this mode doesn't use actual RedHat Linux kernel header files and some definitions are missing.

### DIFF
--- a/tracee-ebpf/Makefile
+++ b/tracee-ebpf/Makefile
@@ -35,6 +35,14 @@ DOCKER_BUILDER_KERN_SRC ?= $(if $(shell readlink $(KERN_SRC_PATH)),$(shell readl
 DOCKER_BUILDER_KERN_SRC_MNT ?= $(dir $(DOCKER_BUILDER_KERN_SRC))
 LINUX_VERSION_CODE := $(shell uname -r | awk '{split($$0,a,"."); split(a[3], b, "-"); print lshift(a[1], 16) + lshift(a[2],8) + b[1];}')
 
+ifneq (,$(wildcard /etc/redhat-release))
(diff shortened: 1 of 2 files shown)


rafaeldtinoco (approved):
LGTM. 

```
$ RHEL_MAJOR=$(grep -Eoh -m 1 '[0-9]+\.[0-9]+' /etc/redhat-release | cut -d '.' -f 1)
$ RHEL_MINOR=$(grep -Eoh -m 1 '[0-9]+\.[0-9]+' /etc/redhat-release | cut -d '.' -f 2)
$ echo $(($RHEL_MAJOR << 8 | $RHEL_MINOR))
2052
```

yanivagman:
LGTM, thanks @NeonSludge!

itaysk:
Shouldn't we've updated automatic compilation in main.go as well?

yanivagman:
Not sure I understand correctly what you mean. but this change is only relevant to the CO-RE binary, so I don't think there should be any impact on the compilation steps in main.go

itaysk:
Thanks for the clarification, I didn't understand that before. I still think we should make the compilation methods identical (they were before CORE) to reduce bug potential.

yanivagman:
main.go shouldn't build co-re object, as it shouldn't be built on the target (this is the whole point of CO-RE).
I don't see any problem here, and no reason to put this logic in main.go

PR #903, opened 2021-08-12
tracee-ebpf: fix get syscall id from regs

### LINKED ISSUE #900: get_syscall_ev_id_from_regs doesn't work when CO-RE is enabled
The get_syscall_ev_id_from_regs function is used to extract the syscall which called some internal kernel function.
This is used either when `--output option:detect-syscall` is chosen, or when `security_file_open` event is chosen (to check if it was originated from open/openat syscalls (which we shuold reconsider if still necessary, but that is for another issue).

When used with a BTF enabled kernel (Manjaro 5.10 in my case) - this functino returns 0, and not the syscall number as expected.
When I compile the code without BTF enabled (by changing the code directly in main.go), the function works as expected.

yanivagman:
Fix #900 

Specifically, the problem was that THREAD_SIZE was defined as follows:
`#define THREAD_SIZE (2*PAGE_SIZE)`
While it should have been:
`#define THREAD_SIZE (PAGE_SIZE << 2)`

I also took the opportunity to highlight the fact that some of these constants are architecture specific, and remove some x32 specific definitions

### DIFF
--- a/tracee-ebpf/tracee/co_re_missing_definitions.h
+++ b/tracee-ebpf/tracee/co_re_missing_definitions.h
@@ -51,31 +51,38 @@
 #define sk_flags          __sk_common.skc_flags
 #define sk_rxhash         __sk_common.skc_rxhash
 
-#define TS_COMPAT 0x0002    /* 32bit syscall active (64BIT)*/
-
 #define PF_KTHREAD  0x00200000	/* I am a kernel thread */
 
 #define TASK_COMM_LEN 16
 
-#define THREAD_SIZE_ORDER 1
+// include/uapi/linux/const.h
+#define __AC(X,Y)   (X##Y)
+#define _AC(X,Y)    __AC(X,Y)
 
-#define THREAD_SIZE (2*PAGE_SIZE)
+/*=============================== ARCH SPECIFIC ===========================*/
+#if defined(__TARGET_ARCH_x86)
+
+#define TS_COMPAT 0x0002    /* 32bit syscall active (64BIT)*/
+
+// arch/x86/include/asm/page_64_types.h
+#define KASAN_STACK_ORDER   0  /* We implicitly assume here that KASAN (memory debugger) is disabled */
+#define THREAD_SIZE_ORDER   (2 + KASAN_STACK_ORDER)
+#define THREAD_SIZE         (PAGE_SIZE << THREAD_SIZE_ORDER)
 
 #define PAGE_SHIFT 12
 #define PAGE_SIZE  (_AC(1,UL) << PAGE_SHIFT)
 #define PAGE_MASK  (~(PAGE_SIZE-1))
 
-#define _AC(X,Y)    X
+#define TOP_OF_KERNEL_STACK_PADDING 0
 
-#ifdef CONFIG_X86_32
-# ifdef CONFIG_VM86
-#  define TOP_OF_KERNEL_STACK_PADDING 16
-# else
-#  define TOP_OF_KERNEL_STACK_PADDING 8
-# endif
-#else
-# define TOP_OF_KERNEL_STACK_PADDING 0
+
+#elif defined(__TARGET_ARCH_arm64)
+extern bool CONFIG_ARM64_PAGE_SHIFT __kconfig;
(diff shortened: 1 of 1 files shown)


tracee-ebpf/tracee/co_re_missing_definitions.h
```diff
@@ -51,31 +51,38 @@
...
+#define KASAN_STACK_ORDER 0  /* We implicitly assume here that KASAN (memory debugger) is disabled */
+#define THREAD_SIZE_ORDER   (2 + KASAN_STACK_ORDER)
+#define THREAD_SIZE  (PAGE_SIZE << THREAD_SIZE_ORDER)
 
 #define PAGE_SHIFT 12
 #define PAGE_SIZE  (_AC(1,UL) << PAGE_SHIFT)
 #define PAGE_MASK  (~(PAGE_SIZE-1))
 
-#define _AC(X,Y)    X
+#define TOP_OF_KERNEL_STACK_PADDING 0
```
rafaeldtinoco, line 76: Alright from what I read from thread_info.h

tracee-ebpf/tracee/co_re_missing_definitions.h
```diff
@@ -51,31 +51,38 @@
...
-#ifdef CONFIG_X86_32
-# ifdef CONFIG_VM86
-#  define TOP_OF_KERNEL_STACK_PADDING 16
-# else
-#  define TOP_OF_KERNEL_STACK_PADDING 8
-# endif
-#else
-# define TOP_OF_KERNEL_STACK_PADDING 0
+#elif defined(__TARGET_ARCH_arm64)
+extern bool CONFIG_ARM64_PAGE_SHIFT __kconfig;
```
rafaeldtinoco, line 80: A comment saying arm64 8.2 started support 52-bit addressing (thus the need for external page shift) would be good here (for future readers). I'm not sure what is our arm64 use case nor if 52-bit addressing is widely adopted (I did not see many bugs from my time at Linaro).

rafaeldtinoco: As a side note: https://github.com/aquasecurity/tracee/issues/851#issuecomment-897498803 applies here for older kernels using 52-bit addressing and CO-RE.

rafaeldtinoco (approved):
Left some comments for clarification only.

tracee-ebpf/tracee/co_re_missing_definitions.h
```diff
@@ -51,31 +51,38 @@
...
+/*=============================== ARCH SPECIFIC ===========================*/
+#if defined(__TARGET_ARCH_x86)
 
-#define THREAD_SIZE (2*PAGE_SIZE)
+#define TS_COMPAT 0x0002    /* 32bit syscall active (64BIT)*/
+
+// arch/x86/include/asm/page_64_types.h
+#define KASAN_STACK_ORDER 0  /* We implicitly assume here that KASAN (memory debugger) is disabled */
+#define THREAD_SIZE_ORDER   (2 + KASAN_STACK_ORDER)
+#define THREAD_SIZE  (PAGE_SIZE << THREAD_SIZE_ORDER)
```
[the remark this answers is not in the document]
yanivagman: Actually, the thread size is 4 pages (this exactly was the problem that is fixed here)

tracee-ebpf/tracee/co_re_missing_definitions.h
yanivagman: The use case is android (for now): https://arxiv.org/pdf/2105.14344.pdf

yanivagman: I just used the PAGE_SHIFT as defined in arch/arm64/include/asm/page-def.h.
This one always uses CONFIG_ARM64_PAGE_SHIFT without mentioning 52-bit addressing:
`#define PAGE_SHIFT      CONFIG_ARM64_PAGE_SHIFT`
From what I understand, the config itself should take care of all the rest:
https://elixir.bootlin.com/linux/v5.13.11/source/arch/arm64/Kconfig#L232

PR #905, opened 2021-08-12
add category to external.go

AsafEitani:
Needed by #775.


yanivagman:
@AsafEitani note that we merged #949, so now there is no need to take this out of #775

Issue #904, opened 2021-08-12
tracee-rules doesn't load rules


itaysk:
regression from #893 , the `findRegoSigs` function used to filter discovered files to only `*.rego` files and now it doesn't, which causes it to try to load a Go Signature from that same rules directory (`*.so`). We simple need to reinstate the rego filter there.

cc: @josedonizetti  @jan0ski since you are both working on this section so possibly you'd want to fix this as well

josedonizetti:
@itaysk  Yup, I have the PR ready already, found this early today while testing. Sorry for the regression.

PR #906, opened 2021-08-12
tracee-rules: fix rego signature loading

josedonizetti:
Fixes https://github.com/aquasecurity/tracee/issues/904


itaysk (approved):
thanks! (I think those tests are excessive)

josedonizetti:
@itaysk should I remove the test? Or some of the options it is testing?

PR #908, opened 2021-08-12
Feature/fetch system info

AlonZivony:
Fetching system information at the start of tracee-ebpf run, and save it to file.
Added support in tracee-rules for the file.
The purpose is to give more sophisticated signatures access to interesting information from the machine tracee-ebpf runs on at the start of its runtime.

### DIFF
--- a/tracee-ebpf/tracee/system_info.go
+++ b/tracee-ebpf/tracee/system_info.go
@@ -1,0 +1,29 @@
+package tracee
+
+import (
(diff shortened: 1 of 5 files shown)


AlonZivony:
Will change to another solution, so will open again in the future

AlonZivony:
Reopen after changing of architecture - now create an event in the start of tracee run with the fetched info.

PR #918, opened 2021-08-15
tracee-ebpf: fix regression - program too large in kernel 4.19

yanivagman:
PR #897 caused a regression in kernel 4.19 - program instruction count was too big.
Seems like using `__builtin_memcpy` uses more instructions that bpf_probe_read().
Change back to use bpf_probe_read() to fix this regression


rafaeldtinoco (approved):
LGTM.

PR #920, opened 2021-08-15
tracee-ebpf: fix capabilities documentation

josedonizetti:
Fixes capabilities documentation to `tracee-ebpf` current state. I know we have https://github.com/aquasecurity/tracee/issues/747 to add support for the newer capabilities, but while it is not done let's have the documentation represent the current requirements. 

Also, although implementing https://github.com/aquasecurity/tracee/issues/747 is an easy change, currently, it doesn't make much sense considering the latest versions of `docker` and `podman` still doesn't support it, for `docker` it was recently re-added (it was disabled after it was first added) but it didn't make its way to a release yet. https://github.com/moby/moby/pull/42011

### DIFF
--- a/docs/install/prerequisites.md
+++ b/docs/install/prerequisites.md
@@ -19,6 +19,6 @@ Exceptions:
 
 For using the eBPF Linux subsystem, Tracee needs to run with sufficient capabilities: 
 - `CAP_SYS_RESOURCE` (to manage eBPF maps limits)
-- `CAP_BPF`+`CAP_TRACING` which are available on recent kernels (>=5.8), or `SYS_ADMIN` on older kernels (to load and attach the eBPF programs).
+- `CAP_SYS_ADMIN` and `CAP_IPC_LOCK` (to load and attach the eBPF programs).
 
 Alternatively, run as `root` or with the `--privileged` flag of Docker.


docs/install/prerequisites.md
```diff
@@ -19,6 +19,6 @@ Exceptions:
 
 For using the eBPF Linux subsystem, Tracee needs to run with sufficient capabilities: 
 - `CAP_SYS_RESOURCE` (to manage eBPF maps limits)
-- `CAP_BPF`+`CAP_TRACING` which are available on recent kernels (>=5.8), or `SYS_ADMIN` on older kernels (to load and attach the eBPF programs).
+- `CAP_SYS_ADMIN` and `CAP_IPC_LOCK` (to load and attach the eBPF programs).
```
yanivagman, line 22: cap_sys_admin and cap_ipc_lock are not enough for kernels >= 5.8, and cap_bpf+cap_tracing are required (instead of cap_sys_admin).
I did the following check to validate:
```
[yaniv@yaniv-aqua tracee-ebpf]$ sudo setcap 'cap_ipc_lock,cap_sys_admin+ep' ./dist/tracee-ebpf 
[yaniv@yaniv-aqua tracee-ebpf]$ getcap ./dist/tracee-ebpf 
./dist/tracee-ebpf cap_ipc_lock,cap_sys_admin=ep 
[yaniv@yaniv-aqua tracee-ebpf]$ ./dist/tracee-ebpf 
libbpf: map 'args_map': failed to create: Operation not permitted(-1)
2021/08/16 13:43:45 error creating Tracee: failed to load BPF object
```

so if docker doesn't support directly specifying these capabilities, then using '--priviliged' is required instead.

josedonizetti: @yanivagman Do you mind sharing which environment did you test it? I mean, if you think this change is unnecessary I can definitely close the PR. But  before changing this I did the same test you did with `setcap` on:

ubuntu 18.04 - kernel 5.8.0-generic
ubuntu 21.04 - kernel 5.11.0-25
ubuntu 21.04 - kernerl 5.12.5 (custom compiled)

And `CAP_SYS_ADMIN` and `CAP_IPC_LOCK` all worked.

yanivagman: Did you then run tracee as root as a non privileged user?
After using setcap, you should run tracee-ebpf as a non-privileged user.
This should be reproduced (I think) on any kernel >= 5.8

josedonizetti: @yanivagman  Non privileged user for sure! Eg:

```
$ rm -rf dist
$ make 
...

# first attempt
$ ./dist/tracee-ebpf 
2021/08/16 10:13:11 insufficient privileges to run: missing CAP_SYS_ADMIN

# setcap
$ sudo setcap cap_ipc_lock,cap_sys_admin=pe ./dist/tracee-ebpf

# second attempt
$ ./dist/tracee-ebpf 
10:13:52:689670  1000   ThreadPoolForeg  4854    4859    7149             clone                flags: CLONE_VM|CLONE_FS|CLONE_FILES|CLONE_SIGHAND|CLONE_THREAD|CLONE_SYSVSEM|CLONE_SETTLS|CLONE_PARENT_SETTID|CLONE_CHILD_CLEARTID, stack: 0x7F711C05FB70, parent_tid: 0x7F711C0609D0, child_tid: 0x7F711C0609D0, tls: 140123778189056
......
```

yanivagman: Ok. So I made some tests (I'm using Manjaro with kernel 5.10), and I found that either of the following two works (I removed the checks in main.go):
1. `sudo setcap 'cap_bpf,cap_perfmon,cap_sys_resource+ep' ./dist/tracee-ebpf` 
2. `sudo setcap 'cap_sys_admin,cap_sys_resource+ep' ./dist/tracee-ebpf`

Interesting to note that:
1. cap_ipc_lock wasn't required in my setup for tracee-ebpf to work, but cap_sys_resource was needed. This actually makes sense as cap_sys_resource is required to setrlimit (https://man7.org/linux/man-pages/man7/capabilities.7.html), and indeed we do that in https://github.com/aquasecurity/libbpfgo/blob/main/libbpfgo.go#L281. So I'm not sure why cap_ipc_lock works in your case, as you shouldn't be able to do setrlimit with it (and now I'm also not sure I understand what memory needs to be locked for libbpf to work).
2. The current comment in the code should be fixed anyways as there is no such cap_tracing, but cap_perfmon should be used instead

josedonizetti: @yanivagman awesome, I also did a few tests here:

- We need to check for `cap_sys_resource` on https://github.com/aquasecurity/tracee/blob/main/tracee-ebpf/main.go#L1032, on ubuntu for some reason the default for `RLIMIT_MEMLOCK` is huge, the reason why it is not complaining about the capability
- For ubuntu `cap_ipc_lock` is required, it fails without it to use the syscall `mmap` (even though the mmap call is not locking memory or using huge pages).
- For kernel >= 5.8, the proper caps are sudo setcap 'cap_bpf,cap_perfmon,cap_sys_resource', this must be documented and later checked on the code (there is an issue for this one already)

I'll change this PR to reflect the proper capabilities, and I'll also change the code to check for `cap_sys_resource`, but I wonder what to do about `IPC_LOCK` do we keep checking it even though it seems only ubuntu needs it?

yanivagman: I would keep the IPC_LOCK as it may be required on some environments, as you noted.
I also think we should add CAP_NET_ADMIN as we recently added tc programs to capture network traffic. (https://lwn.net/Articles/820560/)

yanivagman: BTW, it is possible that starting from kernel 5.11 both IPC_LOCK and SYS_RESOURCE caps won't be needed anymore as per https://lore.kernel.org/bpf/20201201215900.3569844-1-guro@fb.com/t/#u

rafaeldtinoco: > * We need to check for `cap_sys_resource` on https://github.com/aquasecurity/tracee/blob/main/tracee-ebpf/main.go#L1032, on ubuntu for some reason the default for `RLIMIT_MEMLOCK` is huge, the reason why it is not complaining about the capability

This is set by systemd in /etc/systemd/system.conf if you want to override the defaults. Different default values could be because of different systemd versions (not necessarily because of different kernel versions).

---- 

On another topic...

> * For ubuntu `cap_ipc_lock` is required, it fails without it to use the syscall `mmap` (even though the mmap call is not locking memory or using huge pages).

I think we all should **base ourselves in comparisons like this by using initially the upstream supported versions**:  5.14-rc7 , 5.13.12 , 5.10.60 , 5.4.142 , 4.19.204 , 4.14.244 , 4.9.280 , 4.4.281 , next-20210824 and, after and when this is not enough, mention distro specific kernels. This way everybody speaks the same language.

**Example 1) Ubuntus**

Ubuntu kernel is actually dividide into tons of different kernels per Ubuntu version:

- Ubuntu Bionic (older and initial with the release) == v4.15.
- Ubuntu Bionic HWE (newer kernel up to the next LTS, Focal, base kernel) == v5.4

- Ubuntu Focal (older and initial with the release) == v5.4.
- Ubuntu Focal HWE (newer kernel up to the next LTS base kernel) == currently v5.11 (will stop in the next LTS)

and then we have the kconfig dependant changes (-generic, -lowlatency, -aws, -gce, -virtual) which have different set of configurations.

**NOTE** that Ubuntu Bionic HWE v5.4 kernel is NOT the same build as Ubuntu focal base v5.4 kernel. They have different config options, debug symbols, etc.
**NOTE** All Ubuntu kernels follow the upstream stable-strees. This way we can base ourselves in the stable trees only, like I mentioned in the beginning.

**Example 3)** Debian/Archlinux

Those are usually 1:1 with upstream and, in case of Debian Sid and Archlinux, very close to latest upstream. Debian stable tries to mantain a supported/patched kernel in the same version but, from my experience, it is always outdated and not fully fixed, etc, so people tend to use kernels from sid.

**Example 3) CentOS / Alma / Oracle Linux / etc**

Kernels based in RedHat's work are even more complicated because they don't have a clear base kernel and, differently than Ubuntu, they have not only bug fixes but lots of backported features. For THOSE kernels we should always have their version mentioned as they don't follow upstream kernels.

yanivagman:
@josedonizetti I included some of the required changes discussed here in #996

josedonizetti:
@yanivagman sorry for the delay updating the PR. I’ll close it in favor of yours.

PR #922, opened 2021-08-16
tracee-ebpf: support network capture from multiple interfaces

### LINKED ISSUE #914: Add network capture limitations to the docs
As described in #910, we should give a better description of the network capture feature with its current limitations in the docs. This include:
1. Supporting one network interface
2. Supporting tcp+udp only

yanivagman:
Close #914


tracee-ebpf/tracee/tracee.go
```diff
@@ -114,7 +115,7 @@ type CaptureConfig struct {
 	Exec            bool
 	Mem             bool
 	Profile         bool
-	NetIfaces       []string
```
itaysk, line 117: are you sure about this? this is user facing configuration, in similar cases (for example, output path) we always ask for a simple type (string) and do the necessary work internally

docs/tracee-ebpf/capture.md
```diff
@@ -18,7 +18,7 @@ CLI Option | Description
 `[artifact:]write[=/path/prefix*]` | capture written files. A filter can be given to only capture file writes whose path starts with some prefix (up to 50 characters). Up to 3 filters can be given.
 `[artifact:]exec` | capture executed files.
 `[artifact:]mem` | capture memory regions that had write+execute (w+x) protection, and then changed to execute (x) only.
-`[artifact:]net=<network_interface>` | capture network packets generated by traced processes, that goes through <network_interface>.
```
itaysk, line 21: do we need to update the help in main.go as well

rafaeldtinoco (approved):
It all looks very sane to me. I checked var types and padding for the debug net pkt struct, the index map[int][int] logic, etc. I think adding multiple pcap files would be easy from this change. 

> Also checked upstream discussion and couldn't find any update on the qdisk check logic.

tracee-ebpf/tracee/tracee.go
```diff
@@ -423,6 +430,21 @@ func New(cfg Config) (*Tracee, error) {
...
+		for _, iface := range t.config.Capture.NetIfaces[1:] {
+			ngIface = pcapgo.NgInterface{
+				Name:       iface.Name,
+				Comment:    "tracee tc capture",
+				Filter:     "",
+				LinkType:   layers.LinkTypeEthernet,
+				SnapLength: uint32(math.MaxUint16),
+			}
+
+			_, err := pcapWriter.AddInterface(ngIface)
```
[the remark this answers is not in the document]
yanivagman: There are pros and cons to save it as a single file vs multiple files.
For example, with one file, it might be easier to correlate between events happening on different interfaces, while still being able to differentiate between the interfaces.
Let's consider this in a future PR, where we might also give an option to the user to choose to split the pcaps or not.

tracee-ebpf/tracee/tracee.go
yanivagman: I changed it back to slice of string. Thanks

PR #926, opened 2021-08-16
Update tracee-ebpf to libbpfgo v0.2.0-libbpf_0.4.0

rafaeldtinoco:
bdcf9b9 tracee-ebpf: improve error handling and add missing ones
e48e515 tracee-ebpf: Change libbpfgo map methods to use unsafe.Pointer
9e867b4 tracee-ebpf: update libbpf to sync with libbpfgo
c8d2f99 tracee-ebpf: update to libbpfgo v0.2.0-libbpf_0.4.0

### DIFF
--- a/tracee-ebpf/go.sum
+++ b/tracee-ebpf/go.sum
@@ -1,6 +1,6 @@
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
-github.com/aquasecurity/libbpfgo v0.1.1 h1:j/ipxI/l20lu5Tosn0o9JYj7bD3Cwzi18jIEEu3rMvQ=
-github.com/aquasecurity/libbpfgo v0.1.1/go.mod h1:/+clceXE103FaXvVTIY2HAkQjxNtkra4DRWvZYr2SKw=
+github.com/aquasecurity/libbpfgo v0.1.2-0.20210803032413-c500b6207bf6 h1:5DmiBAntdMSoMC4Be4D8h4cswNnkpkDREiWeluK0BC0=
+github.com/aquasecurity/libbpfgo v0.1.2-0.20210803032413-c500b6207bf6/go.mod h1:/+clceXE103FaXvVTIY2HAkQjxNtkra4DRWvZYr2SKw=
 github.com/aquasecurity/tracee/tracee-ebpf/external v0.0.0-20210727091827-bbe411a2a167 h1:QpuLpMxwd9fFH8p2aZ4umKqfpI9/6vf37D/fvLUyHMk=
 github.com/aquasecurity/tracee/tracee-ebpf/external v0.0.0-20210727091827-bbe411a2a167/go.mod h1:C6aED/3HEi4aKSyjtCBWC//7EDXVlV/cQAdDgSbF1eM=
 github.com/cpuguy83/go-md2man/v2 v2.0.0-20190314233015-f79a8a8ca69d h1:U+s90UTSYgptZMwQh2aRr3LuazLJIa+Pg3Kc1ylSYVY=
--- a/tracee-ebpf/tracee/pipeline.go
+++ b/tracee-ebpf/tracee/pipeline.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"strconv"
 	"sync"
+	"unsafe"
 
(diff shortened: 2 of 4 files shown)


tracee-ebpf/go.sum
```diff
@@ -1,6 +1,6 @@
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
-github.com/aquasecurity/libbpfgo v0.1.1 h1:j/ipxI/l20lu5Tosn0o9JYj7bD3Cwzi18jIEEu3rMvQ=
-github.com/aquasecurity/libbpfgo v0.1.1/go.mod h1:/+clceXE103FaXvVTIY2HAkQjxNtkra4DRWvZYr2SKw=
+github.com/aquasecurity/libbpfgo v0.1.2-0.20210803032413-c500b6207bf6 h1:5DmiBAntdMSoMC4Be4D8h4cswNnkpkDREiWeluK0BC0=
```
yanivagman, line 2: Shouldn't this be v0.2.0-libbpf_0.4.0?

tracee-ebpf/tracee/pipeline.go
```diff
@@ -110,13 +112,14 @@ func (t *Tracee) processRawEvent(done <-chan struct{}, in <-chan RawEvent) (<-ch
...
 func (t *Tracee) getStackAddresses(StackID uint32) ([]uint64, error) {
+	StackIDC := C.uint(StackID)
 	StackAddresses := make([]uint64, maxStackDepth)
 	stackFrameSize := (strconv.IntSize / 8)
 
 	// Lookup the StackID in the map
 	// The ID could have aged out of the Map, as it only holds a finite number of
 	// Stack IDs in it's Map
-	stackBytes, err := t.StackAddressesMap.GetValue(StackID)
+	stackBytes, err := t.StackAddressesMap.GetValue(unsafe.Pointer(&StackIDC))
```
yanivagman, line 122: I think it should be ok if you just give `unsafe.Pointer(&StackID)` and remove the "import C" part.

tracee-ebpf/tracee/tracee.go
```diff
@@ -616,212 +619,291 @@ func (t *Tracee) setUintFilter(filter *UintFilter, filterMapName string, configF
...
+	// 4. pid_ns_filter     u64, u32
 	equalityFilter, err := t.bpfModule.GetMap(filterMapName)
 	if err != nil {
 		return err
 	}
 	for i := 0; i < len(filter.Equal); i++ {
 		if filter.Is32Bit {
-			err = equalityFilter.Update(uint32(filter.Equal[i]), filterEqual)
+			EqualC := C.uint(filter.Equal[i])
+			err = equalityFilter.Update(unsafe.Pointer(&EqualC), unsafe.Pointer(&filterEqualC))
```
yanivagman, line 637: I think it should be ok if you just give unsafe.Pointer(&StackID) and remove the "import C" part, same for all other places below.
This is exactly how the old libbpfgo code worked, and it was fine.

yanivagman (commented):
I see that in many places here you first convert to C type, and then using the unsafe.Pointer.
Actually, this is not needed, and the way we did it before was by just giving the original go type as an unsafe.Pointer: https://github.com/aquasecurity/libbpfgo/pull/32/files#diff-c9fafdb5381ed3544be49e84b94f3e3342e61382a849cb5b4bf674765ad27a20L509

rafaeldtinoco:
> I see that in many places here you first convert to C type, and then using the unsafe.Pointer.
> Actually, this is not needed, and the way we did it before was by just giving the original go type as an unsafe.Pointer: https://github.com/aquasecurity/libbpfgo/pull/32/files#diff-c9fafdb5381ed3544be49e84b94f3e3342e61382a849cb5b4bf674765ad27a20L509

Yes. I thought they were mandatory because I was getting some errors from Interfaces{} - I had types being changed depending on conditionals) - and did not understand why. All set now.

tracee-ebpf/go.sum
rafaeldtinoco: go mod tidy changes it automagically. I set it as v0.2.0-libbpf_0.4.0 and it changes to this new way of setting versioning.

rafaeldtinoco:
Good for another review. Thanks!

yanivagman (approved):
LGTM!

Issue #836, opened 2021-07-23
Capability to filter out events by process ancestry


mtcherni95:

yanivagman:
Please explain. What is missing? You can already use pid!=123 or comm!=xyz

mtcherni95:
the use case is the following:
process A forks B, C, and D. I would like to not get any event from A, B, C and D knowing only that A exists.

yanivagman:
If the names of the processes is known in advance, you can filter out by it.
Also, are these processes part of the same container? If so, you can use, for example, mnt_ns!=12345 to not trace that comntainer.

mtcherni95:
> If the names of the processes is known in advance, you can filter out by it.
> Also, are these processes part of the same container? If so, you can use, for example, mnt_ns!=12345 to not trace that comntainer.

that's the thing, the name of the children processes are not known in advance.
using mnt_ns might help however the processes:
1. are running on host namespace
2. go in/out their ns

yanivagman:
Ok. I understand the use case.
There are several questions that need to be answered before implementing such a feature:
1. Do we want to implement this in the bpf code or userspace? Implementing another filter in the kernel means more instructions for every event, and if we do so, we need to make sure we don't pass the instructions limit in old kernel (4.18-5.2). On the other hand, implementing this in userspace is less efficient as we have to send the event to the userspace first, which requires more processing.
2. If we choose to implement this in the kernel, this can be done by adding the forked pids to the pid_filter. If we do that, should we also do that to other filters as well? The logic says yes, but then again, this means even more code added to the filters logic in the kernel.
3. How will this be exposed to the user from the command line? We currently have the "follow" filter to follow forks of processes that passed all filters. The required feature is somehow the same, but with the opposite functionality - If something was filtered out explicitly, then don't follow its descendants. Maybe we can make this flag more generic to reflect that?

grantseltzer:
> the use case is the following:
> process A forks B, C, and D. I would like to not get any event from A, B, C and D knowing only that A exists.

In this situation, what do you know about A? PID? Command name? or any existing filter?

Would having a filter by PPID satisfy this? Or any process in the process tree under a particular PID?

itaysk:
@mtcherni95 @yanivagman how do you suggest to address this request? from an end-user perspective I think it would make most sense if we could make follow an operator-level modifier, but I'm not sure if feasible.

mtcherni95:
[...]

PR #927, opened 2021-08-16
Implement process tree filter

grantseltzer:
Handles #836

Couple of design decisions I made worth discussing

- Having support for multiple process trees would have too much of an impact on # of instructions, validator implications, and performance. I instead opted to only support one process tree filter. Hopefully this meets the use-case of CNDR.
- I build the process tree after the bpf programs are attached. This removes the issue where procs start/end in between the process tree being built and when sched_proc_fork/sched_proc_exit take over maintenance 
- The flag is called `tree` and can be passed like `--trace tree!=3` or `--trace tree=1001`

We must verify that this works on older kernels than my development one however. I'm worried about the possibility that this blows the instruction limit.


yanivagman:
Doesn't work on 4.19:
```
[yaniv@yaniv-aqua tracee-ebpf]$ sudo ./dist/tracee-ebpf 
[sudo] password for yaniv: 
libbpf: load bpf program failed: Argument list too long
libbpf: Program too large (4795 insns), at most 4096 insns
libbpf: failed to load program 'tracepoint__raw_syscalls__sys_exit'
libbpf: failed to load object '/home/yaniv/src/tracee/tracee-ebpf/dist/tracee.bpf.4_19_201-1-MANJARO.v0_6_0-13-g7bb9807.o'
2021/08/17 10:29:37 error creating Tracee: failed to load BPF object
```

yanivagman (commented):
Keeping track of parent and childs relations complicates the implementation.
It requires to loop until parent is found, which wastes cpu cycles, and defines a max depth of the tree (currently set to 100, but probably we will need to lower that so it can work in older kernels <5.2).

Instead, you can use the implementation I described in #836:
1. Define a new map in the kernel - process_tree_map with a key that will represent a pid, and a value that will reresent if the process (and its descendants) should be traced or not.
2. Populate this map during the init phase of tracee, as described above
3. Update sched_process_fork event as follows: If the parent pid is in process_tree_map, add the child pid as well with the same value as of its parent.
4. Update should_trace to consult this filter (probably we can use the equality filter semantics here).
5. Update sched_process_exit to delete the pid entry from the map.

tracee-ebpf/tracee/proc.go
```diff
@@ -0,0 +1,50 @@
...
+
+	filterPidAndChildPids := map[uint32]uint32{
+		filteredPID: 1,
+	}
+
+	processMap := map[uint32][]uint32{}
+	for i := range procs {
+		stat, err := procs[i].Stat()
+		if err != nil {
+			return nil, err
```
yanivagman, line 28: will this retrun an error if a process exited while iterating procs?

tracee-ebpf/tracee/proc.go
```diff
@@ -0,0 +1,50 @@
...
+		return nil, err
+	}
+
+	procs, err := fs.AllProcs()
+	if err != nil {
+		return nil, err
+	}
+
+	filterPidAndChildPids := map[uint32]uint32{
+		filteredPID: 1,
```
yanivagman, line 21: Is the value always set to 1?
You can use the value to know if we shuold trace the process or not (0 value to indicate tree!=x)

tracee-ebpf/tracee/proc.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	processMap := map[uint32][]uint32{}
+	for i := range procs {
+		stat, err := procs[i].Stat()
+		if err != nil {
+			return nil, err
+		}
+		processMap[uint32(stat.PPID)] = append(processMap[uint32(stat.PPID)], uint32(stat.PID))
```
yanivagman, line 30: I think that using this method you will only have the pids, and not the tids.
This means that the threads of a process tree won't be included, which is not what we want.

tracee-ebpf/tracee/tracee.go
```diff
@@ -902,6 +940,33 @@ func (t *Tracee) populateBPFMaps() error {
 		}
 	}
 
+	err = t.populateProcessTreeMap(t.config.Filter.ProcessTreeFilter.PID)
```
yanivagman, line 943: This map is populated before the bpf programs are attached.
If processes are added/remved after this point, they will not be added/removed from the map.
We should probably do another iteration after the bpf programs are attached, and check if no changes occured to avoid race conditions

yanivagman (changes requested):
Comments inline.
Also, is there any reason to update libbpf files here?

tracee-ebpf/tracee/tracee.go
```diff
@@ -816,6 +844,16 @@ func (t *Tracee) populateBPFMaps() error {
...
+
+	err = t.setProcessTreeFilter(t.config.Filter.ProcessTreeFilter)
+	if err != nil {
+		return fmt.Errorf("error setting process tree filter: %v", err)
+	}
+
+	err = t.setStringFilter(t.config.Filter.UTSFilter, "uts_ns_filter", configUTSNsFilter)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
grantseltzer: Must have been an accident while rebasing

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2408,8 +2444,10 @@ int tracepoint__sched__sched_process_exit(struct bpf_raw_tracepoint_args *ctx)
     if (!should_trace()) {
         // Note: we need to remove the container id here as we always add it to the map in cgroup_attach_task event.
         bpf_map_delete_elem(&pid_to_cont_id_map, &pid);
+        bpf_map_delete_elem(&process_tree_map, &pid);
```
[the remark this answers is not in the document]
grantseltzer: But if a process tree filter was specified, this affects the logic of `should_trace`.

tracee-ebpf/tracee/proc.go
```diff
@@ -0,0 +1,91 @@
+package tracee
+
+import (
+	"github.com/prometheus/procfs"
+)
+
+//TODO:
+// != doesn't work
+//
```
yanivagman, line 9: Is this still a todo?

tracee-ebpf/tracee/proc.go
```diff
@@ -0,0 +1,87 @@
...
+		return nil, err
+	}
+
+	processMap := map[uint32][]uint32{}
+	for i := range procs {
+		stat, err := procs[i].Stat()
+		if err != nil {
+			return nil, err
+		}
+		processMap[uint32(stat.PID)] = processMap[uint32(stat.PID)]
```
yanivagman, line 24: What does this line do?

tracee-ebpf/tracee/proc.go
yanivagman: This is still a problem...

tracee-ebpf/tracee/proc.go
```diff
@@ -0,0 +1,87 @@
...
+		processMap[uint32(stat.PPID)] = append(processMap[uint32(stat.PPID)], uint32(stat.PID))
+	}
+
+	return processMap, nil
+}
+
+// populateProcessTreeFilterMap takes a map of the process tree (k=ppid, v=[]childpid)
+// and a filter specification map (k=pid,v=trace/not) and returns a fully populated map
+// where k=pid, v=trace/not for all pids
+func populateProcessTreeFilterMap(processTree map[uint32][]uint32,
```
yanivagman, line 34: I don't understand why we shuold first gather the process tree representation, and only then initialize the map.
There is a big window for race conditions using this method, and it wastes cycles for processes we don't really need (were not requested by the user tree filter)

tracee-ebpf/tracee/proc.go
```diff
@@ -0,0 +1,87 @@
...
+	var defaultFilter = true
+	for _, v := range filterSpecification {
+		defaultFilter = defaultFilter && v
+	}
+	defaultFilter = !defaultFilter
+
+	// Populate inital filter map  (keys representing all pids)
+	for pid, _ := range processTree {
+		filterMap[pid] = defaultFilter
+	}
```
yanivagman, line 52: I don't think we need to save this value for every pid in the system.
Instead, we can pass the default filter value in the config map, and use it when we don't find a given pid in the map

tracee-ebpf/tracee/proc.go
```diff
@@ -0,0 +1,87 @@
...
+			if shouldBeTraced {
+				for j := range descendentPIDs {
+					filterMap[descendentPIDs[j]] = true
+				}
+			} else if !shouldBeTraced {
+				for j := range descendentPIDs {
+					filterMap[descendentPIDs[j]] = false
+				}
+			}
+		}
```
yanivagman, line 70: I would loop over filterSpecification instead (related to my previous comment that we don't need to iterate over all the process tree)

tracee-ebpf/tracee/proc.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	return filterMap
+}
+
+// gatherAllDescedentPIDs takes a specific pid, and a map 'pids' which represents
+// a snapshot of the process tree where k = ppid and v = slice of child pids
+// and returns a slice of all the descedent pids
+func gatherAllDescedentPIDs(pid uint32, pids map[uint32][]uint32) []uint32 {
```
yanivagman, line 79: This is where I think you should implement the procfs walk - only for the pids selected in the filter

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -482,36 +483,37 @@ struct mount {
...
+BPF_LRU_HASH(sock_ctx_map, u64, net_ctx_ext_t);         // Socket address to process context
+BPF_LRU_HASH(network_map, local_net_id_t, net_ctx_t);   // Network identifier to process context
+BPF_ARRAY(file_filter, path_filter_t, 3);               // Used to filter vfs_write events
+BPF_ARRAY(string_store, path_filter_t, 1);              // Store strings from userspace
+BPF_PERCPU_ARRAY(bufs, buf_t, MAX_BUFFERS);             // Percpu global buffer variables
+BPF_PERCPU_ARRAY(bufs_off, u32, MAX_BUFFERS);           // Holds offsets to bufs respectively
+BPF_PROG_ARRAY(prog_array, MAX_TAIL_CALL);              // Used to store programs for tail calls
+BPF_PROG_ARRAY(sys_enter_tails, MAX_EVENT_ID);          // Used to store programs for tail calls
+BPF_PROG_ARRAY(sys_exit_tails, MAX_EVENT_ID);           // Used to store programs for tail calls
+BPF_STACK_TRACE(stack_addresses, MAX_STACK_ADDRESSES);  // Used to store stack traces
```
yanivagman, line 516: Please rebase

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1023,6 +1030,7 @@ static __always_inline int uint_filter_matches(int filter_config, void *filter_m
     return 1;
 }
 
+
```
yanivagman, line 1033: nit: extra line

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2398,13 +2421,17 @@ int tracepoint__sched__sched_process_exec(struct bpf_raw_tracepoint_args *ctx)
...
+    u32 pid = id;
+    u32 tgid = id >> 32;
 
     if (!should_trace()) {
         // Note: we need to remove the container id here as we always add it to the map in cgroup_attach_task event.
         bpf_map_delete_elem(&pid_to_cont_id_map, &pid);
+        bpf_map_delete_elem(&process_tree_map, &tgid);
         return 0;
     }
+    bpf_map_delete_elem(&process_tree_map, &pid);
```
yanivagman, line 2434: shouldn't this be tgid as well?

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1114,6 +1122,12 @@ static __always_inline int should_trace()
     {
         return 0;
     }
+
+    if (!equality_filter_matches(CONFIG_PROC_TREE_FILTER, &process_tree_map, &context.pid))
```
yanivagman, line 1126: If you are saving tgids in the map - why do you campre to pid?

tracee-ebpf/tracee/proc.go
grantseltzer: This is so that all PIDs are recorded in the map, even if they don't have any child processes.

tracee-ebpf/tracee/proc.go
grantseltzer: Sure this takes up some extra space, but this adds a few more instructions in each  `should_trace()` call to check the default value if the PID isn't in the map. I don't feel that strongly about this, I can if you'd like.

tracee-ebpf/tracee/proc.go
grantseltzer: The reason we gather the process tree first is so we can determine all of the descendant processes. Is there a way to query proc on a per pid basis to traverse through all its child pids that i'm not aware of?

grantseltzer: As in for each PID we can gather its PPID, but I don't see anything like a list of child pids

tracee-ebpf/tracee/proc.go
grantseltzer: The process tree map in bpf code is using tgid and in userspace we use pid, aren't these the same?

tracee-ebpf/tracee/tracee.bpf.c
grantseltzer: Ah yes good catch.

tracee-ebpf/tracee/tracee.bpf.c
grantseltzer: In the `struct event_context` definition you have the 'pid' field commented with `PID as in the userspace term` which appears to be accurate int `init_context`

tracee-ebpf/tracee/proc.go
yanivagman: But this is a self-assignment of processMap[uint32(stat.PID)] to itself, how does it achieves this?

tracee-ebpf/tracee/proc.go
yanivagman: Yes, I see that the bpf code is now using tgid. Should be ok now, however I wonder if it wouldn't have been better to set tids both in userspace and and bpf code for better granularity of the filter. Probably not that important, just a thought.

tracee-ebpf/tracee/proc.go
yanivagman: Yes, there is actually a way to do that. To every process, you can list all the threads and their children using:
/proc/pid/task/tid/children

tracee-ebpf/tracee/proc.go
yanivagman: equality_filter_matches (which you use in the bpf code) already takes care of that. You should just set CONFIG_PROC_TREE_FILTER to either FILTER_IN or FILTER_OUT in the config map.

tracee-ebpf/tracee/tracee.bpf.c
yanivagman: Yes, you are right

tracee-ebpf/tracee/proc.go
grantseltzer: Lol that's actually interesting, I never stopped to think that it shouldn't be valid code. It does work though.

Played with it here:
 https://play.golang.org/p/2uHNoyTcnfh

tracee-ebpf/tracee/proc.go
grantseltzer: Ahhh, thanks for teaching me that. Will have iteration #5 soon :-)

grantseltzer: @yanivagman What are your thoughts on the proc man page section on this file:

```
 /proc/[pid]/task/[tid]/children (since Linux 3.5)
        A  space-separated list of child tasks of this task.  Each child task is rep‐
        resented by its TID.

        This option is intended for use by the checkpoint-restore (CRIU) system,  and
        reliably  provides  a list of children only if all of the child processes are
        stopped or frozen.  It does not work properly if children of the target  task
        exit  while  the  file is being read!  Exiting children may cause non-exiting
        children to be omitted from the list.  This makes this  interface  even  more
        unreliable  than  classic  PID-based approaches if the inspected task and its
        children aren't frozen, and most code should probably not use this interface.
```

Also we'll still have to convert the TID to TGID, regardless of the stability concerns listed above.

grantseltzer: The current implementation (map entire proc tree to determine parent/child relationships) can have a data race where we record procs that exit after being set up (false positive). However since the map is built after events are installed the update in sched_process_exit should catch this (must confirm if this is reflected in the map).

If we use `/proc/[pid]/task/[tid]/children` we have the opposite problem (false negative) where by nature of procfs where we may not know about procs that do in fact exist. The map updates in sched_process_(fork|exit) won't handle this.

tracee-ebpf/tracee/proc.go
yanivagman: It seems that it doesn't work:
./prog.go:30:2: self-assignment of x[foo] to x[foo]

tracee-ebpf/tracee/proc.go
yanivagman: > The current implementation (map entire proc tree to determine parent/child relationships) can have a data race where we record procs that exit after being set up (false positive). However since the map is built after events are installed the update in sched_process_exit should catch this (must confirm if this is reflected in the map).

The race is between the time you read procfs data, to the time you update the map. If you update the map and after some process already exited, you will never receieve sched_process_exit again for this process, which means that the pid will stay in the map. This is why I suggested to traverse procfs twice

yanivagman: > @yanivagman What are your thoughts on the proc man page section on this file:

I see. Let's not use this interface then

yanivagman: Ok, so here is another suggestion:
For every process 'p' in procfs:
1. Get its ppid
2. If ppid = one of the given tree filters pids -> add 'p' to the map, and continue to the next process
3. If ppid = 1 -> continue to the next process (as we reached the root of the process tree)
4. Go to step 1 with ppid 

This method will find the relevant set of pids which are part of the tree of the pid given by the tree filter, and update the map in (almost) the same time, reducing the window of a possible race condition

yanivagman: @rafaeldtinoco to get the changes of the process tree in runtime is not the problem here, as we have sched_process_fork and sched_process_exit.
The problem here is - how can we initialize a map that contains a "snapshot" of the pids in the process tree whose root is some given pid. The challenge here is to initialize this map and avoid race conditions of exited/new processes in this process tree

rafaeldtinoco: > @rafaeldtinoco to get the changes of the process tree in runtime is not the problem here, as we have sched_process_fork and sched_process_exit.

I had that in mind when I wrote.

> The problem here is - how can we initialize a map that contains a "snapshot" of the pids in the process tree whose root is some given pid. The challenge here is to initialize this map and avoid race conditions of exited/new processes in this process tree

That's why I referred a userland way of following the process tree (by following forks, execs and exits from userland, using netlink, in parallel to BPF program startup).

A taskstats/conector go routine could monitor selected pid trees for forks an and exits and keep updating the map by removing the pid if that exited before the BPF program was loaded (or update the map with other needs)

It would still be racy - as there is no mutex locking mechanism in between kernel BPF program and a userland task - but the fact that pid numbers would keep growing, instead of reusing just exited ones, would protect us until BPF program is loaded, as the race window is small (not bigger than pid # reusable).

Of course this ads complexity to the initial logic.

Another possibility would be to timestamp each entry in the map and have some kind of prune mechanism for leftovers.

tracee-ebpf/tracee/tracee.bpf.c
grantseltzer: To be honest I don't understand what's going on here, the spacing is the same, they're just moved down a line becuase of the new CONFIG variable and the new BPF_HASH. git is just being verbose?

tracee-ebpf/tracee/proc.go
grantseltzer: @yanivagman A couple of unfortunate issues with [your design](https://github.com/aquasecurity/tracee/pull/927#discussion_r699629208) that I've run into. If we use `filepath.WalkDir()` we're not able to jump to different points in the process tree, it's iterative. That can be worked around, but the reason it's iterative is because it reads the entire directory upfront, taking a snapshot of it, and presenting the same data race as the current design. Even if not using `filepath.WalkDir` I run into the same issue with similar implementations that aren't iterative, we can only make the race window smaller if we can't get atomic state of the process tree (i.e. halt the OS :smile: ). 

@rafaeldtinoco I will read through that kernel doc you linked, i'm not familiar with that interface. 

Another simple idea (again, still has unavoidable race) is to use the current design except re-check the state of the whole process tree on every iteration of traversing PID/PPID relationship (i.e. call `gatherEntireProcessTree()`)  in every iteration of the loop in `populateProcessTreeFilterMap()`. I may need to explore this.

tracee-ebpf/tracee/proc.go
grantseltzer: That's a go vet error, but i'm still not entirely sure why it doesn't complain in this context.

tracee-ebpf/tracee/proc.go
rafaeldtinoco: @grantseltzer its likely we wont use this, as a curiosity though: https://bewareofgeek.livejournal.com/2945.html ... I used something like this in my previous life example =). Its a good example on how to get PROC_EVENT_FORK, PROC_EVENT_EXIT, PROC_EVENT_EXEC from userland (before we could do this with eBPF, for example). It has a netlink socket buffer and its a stream from kernel to userland, but.. yet, its fast. 

> But again, I think your discussion with Yaniv is in advanced state already.

tracee-ebpf/tracee/proc.go
grantseltzer: That's a go vet error, but i'm still not entirely sure why it doesn't complain in this context.

tracee-ebpf/tracee/tracee.bpf.c
rafaeldtinoco: check tabs versus spaces.. sometimes your editor has replaced tabs/spaces for keeping all // aligned.

grantseltzer: Nah that wasn't the issue. Just a combination of all of the lines moving down one and adding a line in the middle. No other difference in tabs/spaces or spaces between code and comment. I put the new BPF_HASH on the bottom to prove it but can revert that so all the filter BPF_HASH maps are next to each other

grantseltzer:
Ok, I'm satisfied with how I have this now. I've moved the bpf map update calls for the process tree map to be closer to when they're collected. This shortens the window for a data race to occur. Importantly, the bpf programs are installed before this map is populated, so the map is being updated before it's even finished being initialized. Another thing is that a data race should never be an issue when the filter is specified for long standing processes (which I expect it always will be). Ultimately there'll always be a tiny amount of risk of losing a data race unless we can halt the kernel, but the impact is negligible.

Another interesting thing I discovered is that libbpfgo's new API that uses `unsafe.Pointer` is a little unsafe. When we take pointers to booleans, false is typically interpreted as true in bpf code since anything other than 64 zero bits will be true. Just something to think about in the future.

grantseltzer:
My latest commit also makes use of FILTER_IN/FILTER_OUT, so there's less time spent updating bpf maps, that should cut down the race window as well.

yanivagman (approved):

PR #931, opened 2021-08-18
fix rego ext check

AsafEitani:
fixes the error getting when tracee-rules tries to load builtin.so as a rego signature.
the check was removed on this commit:
https://github.com/aquasecurity/tracee/commit/82a1289c12da2255682743f6d762bf6adde69a9b#diff-c56a5239a886de2c9586c32e19abf0abcc3ae40b97437bc5fd3a9a78538281b5L87


josedonizetti (commented):
@AsafEitani did you pull the latest? Because this should be fixed after the merge of https://github.com/aquasecurity/tracee/pull/906

PR #934, opened 2021-08-19
tracee-ebpf: restructure and split files

yanivagman:
This PR does **NOT** change any logic.
It splits tracee.go into several files as follows:
1. events_decoder.go - handles the decoding of events read from the events perf buffer
2. events_processor.go - handles processing of events
3. file_capture.go - handles events recieved on the file_writes perf buffer
4. net_events.go - handles events recieved on the net_events perf buffer
5. buckets_cache.go - bucketsCache functions and structure definition


itaysk (approved):
obviously a very welcome improvement. some thoughts:
1. in most cases the act of grouping a piece of code in it's own file is a good indication that it could be it's own package. for example: bucketsCache doesn't belong in package tracee. also good indication is where you used agent nouns to name the file (decoder /processor)
2. this brings up another discussion we had about renaming package tracee to something more relevant (could be an another PR just popped to my mind)
3. could we add some structure to the file naming convention:
file_capture - I see it's just the file writes so maybe write_capture? 
net_events.go - for consistency with file_capture maybe network_capture?
events_* - isn't everything in package tracee "event" handling? (remember this is all in the "tracee" directory which based on 2^ is more like "eventmonitor" directory, so that's already some context

yanivagman:
Yes, I also had in mind 1+2 - we can do it in a future PR.
About the names: The naming comes from the fact that we have 3 different perf buffers (maps):
- events map - used to send events which are shown in the output of tracee. All the events_* are related to this perf buffer.
- file_writes - which handles not only file_writes, but also capturing memory artifacts (of w+x regions) - so in this case, I agree that write_capture is a better name, and I'll change it
-  net_events - currently used for network capture, but also for network debug messages (which doesn't have to include capture data), and possibly for other network related stuff in the future

PR #937, opened 2021-08-19
tracee-rules: add GetSelectedEvents

### LINKED ISSUE #936: get event selectors from loaded signatures
Add an API on tracee-rules engine to `GetSelectedEvents() []string` for #935

itaysk:
close #936

### DIFF
--- a/tracee-rules/engine/engine.go
+++ b/tracee-rules/engine/engine.go
@@ -334,3 +334,12 @@ func analyzeEventOrigin(event tracee.Event) string {
 		return EVENT_HOST_ORIGIN
 	}
 }
+
+// GetSelectedEvents returns the event selectors that are relevant to the currently loaded signatures
+func (engine *Engine) GetSelectedEvents() []types.SignatureEventSelector {
+	res := make([]types.SignatureEventSelector, 0)
+	for k := range engine.signaturesIndex {
+		res = append(res, k)
+	}
+	return res
+}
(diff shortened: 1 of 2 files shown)


tracee-rules/engine/engine.go
```diff
@@ -334,3 +334,12 @@ func analyzeEventOrigin(event tracee.Event) string {
 		return EVENT_HOST_ORIGIN
 	}
 }
+
+// GetSelectedEvents returns the event selectors that are relevant to the currently loaded signatures
+func (engine *Engine) GetSelectedEvents() []types.SignatureEventSelector {
```
yanivagman, line 339: Don't we want the signature for this to be:
`GetSelectedEvents() []string`
As described in #936 ?

tracee-rules/engine/engine.go
```diff
@@ -334,3 +334,12 @@ func analyzeEventOrigin(event tracee.Event) string {
...
+	res := make([]types.SignatureEventSelector, 0)
+	for k := range engine.signaturesIndex {
+		res = append(res, k)
```
yanivagman, line 342: If the answer to the above question is yes, then I suggest:
1. Only save the Name field of the SignatureEventSelector to get the event name
2. To avoid duplications (different signatures that ask for the same events), you can use a map instead

tracee-rules/engine/engine.go
itaysk: I think that we'll need the entire event selector because potentially it would include more filtering data for tracee (fixed the example in the issue)

yanivagman: Ok. Anyways, I think duplications should be removed in the returned value

itaysk: they are removed (see the test)

yanivagman (approved):

PR #938, opened 2021-08-19
tracee-rules: update dependency	Masterminds/sprig

josedonizetti:
Not sure if there is a specific reason or not, but `Masterminds/sprig` v2.22.0 is a 2019 version, and incompatible with go mods. This PR is updating it to the latest version.


itaysk:
thanks. just making sure did you make a basic sanity test after this update? just to make sure the embedded templates still work

itaysk (approved):

PR #939, opened 2021-08-19
tracee-rules: refactor non used code

josedonizetti:
This code probably made sense at some point, but it doesn't anymore. The test for an empty event source returns `err`, but all the calls before check `if err != nil { return err}`, so if somehow the code entered inside the if it would `return nil`.  But it doesn't seem like that [inputs](https://github.com/aquasecurity/tracee/blob/main/tracee-rules/main.go#L91-L94) will be empty, it will either have a source or return [error](https://github.com/aquasecurity/tracee/blob/ceece80d347ef5e23777a4f8888cfd3fca9b4447/tracee-rules/input.go#L33)

### DIFF
--- a/tracee-rules/main.go
+++ b/tracee-rules/main.go
@@ -93,10 +93,6 @@ func main() {
 				return err
 			}
 
-			if inputs == (engine.EventSources{}) {
-				return err
-			}
-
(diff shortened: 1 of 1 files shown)


grantseltzer:
Makes sense, looks like it used to make more sense https://github.com/aquasecurity/tracee/commit/0e5c733cedfdd6422613643169c2a3c38a88627f

grantseltzer (approved):

PR #940, opened 2021-08-19
tracee-ebpf: fix error printing to be always text

josedonizetti:
Fixes https://github.com/aquasecurity/tracee/issues/752

### DIFF
--- a/tracee-ebpf/tracee/printer.go
+++ b/tracee-ebpf/tracee/printer.go
@@ -255,14 +255,12 @@ type gobEventPrinter struct {
 	out    io.WriteCloser
 	err    io.WriteCloser
 	outEnc *gob.Encoder
-	errEnc *gob.Encoder
 }
 
 func (p *gobEventPrinter) Init() error {
 	p.outEnc = gob.NewEncoder(p.out)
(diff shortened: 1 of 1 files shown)


itaysk:
this changes the format of the errors to get printed as text, but they are still printed to the printer's logger. For example, a gob printer would now print text to the gob error logger. I think we should eliminate the printer-specific error logger, and always print errors to stderr. as the issue suggests:

> we should always print errors to standard error in text format

also want to hear what others think

josedonizetti:
@itaysk Sorry, should have mentioned about it on the PR description. I considered doing the change for `stderr`, but opted not to do because it could generate a lot of confusion because of the `-o err-file:...` option, the printers are using `stderr` by default, unless a `-o err-file:..` was configured, to send the errors to  a specific file, which for me seems the best option here, or we drop the flag `err-file`. 

- https://github.com/aquasecurity/tracee/blob/main/tracee-ebpf/tracee/tracee.go#L320-L321
- https://github.com/aquasecurity/tracee/blob/main/tracee-ebpf/tracee/printer.go#L64

yanivagman (approved):
LGTM

Issue #762, opened 2021-06-10
update tracee-ebpf's gob test 


itaysk:
https://github.com/aquasecurity/tracee/tree/main/tracee-ebpf/test/gob
should use the external package

PR #941, opened 2021-08-20
tracee-ebpf: fix gob test

josedonizetti:
Fixes https://github.com/aquasecurity/tracee/issues/762

testing:
```
cd tracee-ebpf
make
sudo ./dist/tracee-ebpf -o gob | go run test/gob/gob.go
```


itaysk (approved):
awesome, thanks! while we are here, could you add that note about how to test in a readme in the same directory (just a few lines you wrote in this PR description)

Issue #930, opened 2021-08-17
Wait for tracee-ebpf to init before signaling readiness


yanivagman:
Currently, when starting the tracee container with the default values, we print:
`Loaded signature(s):  [TRC-1 TRC-2 TRC-3 TRC-4 TRC-5 TRC-6 TRC-7]`
This message is printed even when tracee-ebpf failed to load (or is still initializing) and may mislead the user which might think that tracee is up and running.
We should either wait with this message until tracee-ebpf finish its initialization phase or signal readiness with another message.

grantseltzer:
Moving this to milestone v0.6.3

PR #944, opened 2021-08-21
wait for tracee-ebpf to load

itaysk:
fix: #930


entrypoint.sh
```diff
@@ -25,5 +27,23 @@ for arg in "$@"; do
...
+rm -f $TRACEE_EBPF_READINESS
+echo "starting tracee-ebpf..."
+$TRACEE_EBPF_EXE --output=format:gob --security-alerts >pipe &
+# wait for tracee-ebpf to fully load
+i=0
+while [ ! -f $TRACEE_EBPF_READINESS ] && [ $i -le 10 ]
+do
+	i=$(( i + 1 ))
+	sleep 1
+done
```
rafaeldtinoco, line 43: By the time we write tracee-ebpf pid to the pid file, tracee-ebpf is not fully initialized as we may think, not fully solving issue #930 . Should we also move the time we write tracee.pid for after the pcap initialization and the StackAddressesMap := GetMap() call ? We could also solve a TODO at https://github.com/aquasecurity/tracee/blob/main/tracee-ebpf/tracee/tracee.go#L403 and it would all make sense to the same change. WDTY ?

rafaeldtinoco (changes requested):

entrypoint.sh
itaysk: Yes I already did that in my local branch, planned to send with the same or but couldn't be sent separately. Thanks for noticing

yanivagman:
Can this one be closed in favor of #967 ?

PR #945, opened 2021-08-21
Feature/fetch system info

AlonZivony:
Created an event at the start of tracee-ebpf run that collect system information for the use of signatures that need it.


rafaeldtinoco:
@AlonZivony , would you mind rebasing this interactively and drop the 2 commits you reverted, so history is not poluted ? Could you also rebase onto origin/main so the conflict is solved ?

rafaeldtinoco:
Alright, I have squashed all the commits into a single one so I could better understand what was being proposed. Maybe you should do the same ? Squash everything and explain in the commit log what is being done and why ?

I think this is the first time a 'userland' event is being emulated through the EventConfig / EventsIDToParams logic. I like the approach. @yanivagman are you okay with that approach (likely you are even more into this than I am) ?

Question: do we want to have this as a possible event driven interface for userland events ? Example: Another task 'requests' events to be printed from a new event channel (it could even be based in bpf maps since 2 userland tasks can 'talk' through the maps).

Question: Can we ellaborate a little further what problem this is solving ? Perhaps document it together with a signature example using it ?

AlonZivony:
About the squash, I usually want the history to represent exactly what was done so nothing would be missing if the approach taken is changed in the end. This is why I used 'revert' and not 'fixup' approach, so the history will remain linear. Squashing commits will make it harder to maintain the changes, but I saw that in large projects it is a common practice to squash commits when merging PRs. We can take this approach if you think it is the best practice for our case.
About the documentation I will do it, I think that this is a good idea. Currently the event brings information I need for signature I write, but maybe I can create some small example. I doesn't need to be a real signature after all, just example.

AlonZivony:
I noticed now that currently the event I creates cannot be unchosen by the -t flag.
What do you think, should I change it so the event will be invoked only if it was specified?

rafaeldtinoco:
> I noticed now that currently the event I creates cannot be unchosen by the -t flag.
> What do you think, should I change it so the event will be invoked only if it was specified?

If you ask me, its weird to get an initial event out of nothing. At the same time... the name is 'system_info_fetch', so it kind of gives us an idea that it was automatically generated for a specific purpose. 

**Now, thinking out loud:**

 > At the same time, this will be processed by tracee-rules so it can keep information about the running environment. 

1) Is there a way we can make tracee-rules to 'request' those type of events ? My idea:

- tracee-rules can request the event at anytime
- to request an event, tracee-rules can add the event name to an ebpf-map (a queue map with event names only)
- tracee would receive an event out of this queue and process its internal event, generating it at its output
- tracee-rules, from this event, would take the actions you need (for this logic)

**Thinking a bit more:**

- we can have a helper in libbpfgo to request an event from tracee, tracee-rules would simply use that helper
- this would make those internal events request-based AND they could be extended for other requests (during runtime)

**What I don't know:**

2) do you need those events only in the beginning of tracee-rules execution ?
3) will there ever be a situation where you might have to request information during tracee-rules runtime ?

AlonZivony:
I see the main issue in what you wrote is the question if we want these user-mode events to be triggered by tracee-rules.
This connects to the issue I talked about with @yanivagman about the independency of these 2 tools from one another.

**What I think**
I think that tracee-rules shouldn't talk directly with tracee-ebpf, but only receive information from it.
The reason for that is that I think that we want tracee-rules to be able to run on another machine than tracee-ebpf, so it means that tracee-ebpf runs alone and creates output that tracee-rules examine afterwards somewhere.
By creating bidirectional communication between the two, we might depend too much on the communication thatoperations we do that does not allow them to communicate might not be supported anymore or much less functional.

**Counter argument**
In opposite to my opinion, the question of the independency of the tools is now discussed with Yaniv and maybe the direction for the tools is to be united somehow, and by that making it possible to enable this communication.

**About the current PR**
- For now I think this PR is good because it creates an event that allow to collect useful information and does not break the current design of the tools. If more information will be needed it could be easily added to this event as another parameter, and if information during runtime will be needed in the future we will create an issue and discuss it.
- I do think that the current behavior that we can't cancel the event is unexpected and should be modified. I will look how I can do it.
- About the example test - I think writing a test that uses this event is quite trivial and not different from other state signatures. If you think it is needed I will write one and add it to the PR.

rafaeldtinoco:
> **What I think**
> \<snip\>
> **Counter argument**
> \<snip\>

I imagine that 'having both in the same tool' would be the option of running tracee as a library (idea that is already being worked on, right ?). I see pros and cons now, thank you.

> **About the current PR**
> 
> * For now I think this PR is good because it creates an event that allow to collect useful information and does not break the current design of the tools. If more information will be needed it could be easily added to this event as another parameter, and if information during runtime will be needed in the future we will create an issue and discuss it.

Yep, I like the idea of 'event emulation' of this merge. Just wanted to antecipate if we would have to make it generic for other events as well.

> * I do think that the current behavior that we can't cancel the event is unexpected and should be modified. I will look how I can do it.

Yep, +1 here. I'll be happy to +1 this after you push this with that change.

> * About the example test - I think writing a test that uses this event is quite trivial and not different from other state signatures. If you think it is needed I will write one and add it to the PR.

Nah, we're good. I was catching up with the entire need/idea. Thanks for all the info.

rafaeldtinoco:
Alright, the test looks good:

```
$ sudo ./dist/tracee-ebpf --trace uid=1000 --trace pid=new --trace event!=system_info_fetch

$ sudo ./dist/tracee-ebpf --trace uid=1000 --trace pid=new
10:45:56:900363  0      tracee-ebpf      0       0       0                system_info_fetch    initNamespaces: map[cgroup:4026531835 ipc:4026531839 mnt:4026531840 net:4026532008 pid:4026531836 pid_for_children:4026531836 time:4026531834 time_for_children:4026531834 user:4026531837 uts:4026531838]
```

and all the code seems like we've discussed previously. I would recommend that before merging this is squashed and the git log is similar to:

```
tracee-ebpf: fetch-system-info event

* Create function to fetch information about the system at initialization
* Save fetched system information into file system_info.json
* Add test for current system info fetched
* Fix bug with fetching symlinks contents
* Fix path of system_info save
* Create system info file read and retrieve functions
* Add global initialization according to flag
* Add documentation to system info helpers
* Chang system info fetch to invoke event instead of saving file
* Modify some system info event fields
* Add system info event to consts.go
* Add support for used map type in gob printer and reader
* Remove unused system info file save
* Change system info event invoking to use printer instead of tracee
* Update the system info tests to match current build
* Remove unused const
* Create documented sub-package for user mode events and moved system info into it
* Change created system info event name
* Change system_info_fetch event to be invoked only if specified

Signed-off-by: Alon Zivony <alon.zivony@aquasec.com>
```

> Note that in most of open source upstream projects - if not all - the verb is always in present tense, and the steps made during the development are usually either explained in a long git log - referring other commits so it can make sense - OR, like this one, listing each of the changes you'd like to be documented.

@itaysk you have subscribed as reviewer so I suppose you want to have the final word on this. I'm +1 as is with a single doubt: should we have the event default or make it available only by request ?

tracee-ebpf/tracee/tracee.go
```diff
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/aquasecurity/tracee/tracee-ebpf/tracee/userevents"
```
yanivagman, line 9: Please put it below `github.com/aquasecurity/tracee/tracee-ebpf/external`

tracee-ebpf/tracee/tracee.go
```diff
@@ -1425,3 +1427,10 @@ func (t *Tracee) prepareArgsForPrint(ctx *context, args map[argTag]interface{})
 
 	return nil
 }
+
+func (t *Tracee) invokeSystemInfoEvent() {
+	if t.eventsToTrace[SystemInfoEventID] {
+		systemInfoEvent, _ := userevents.CreateSystemInfoEvent()
+		t.printer.Print(systemInfoEvent)
```
yanivagman, line 1434: Please rebase. We now use a channel instead

tracee-ebpf/tracee/userevents/system_info.go
```diff
@@ -0,0 +1,62 @@
...
+	"path/filepath"
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
+)
+
+const INIT_PROC_NS_DIR = "/proc/1/ns"
+
+const INIT_NAMESPACES_KEY = "initNamespaces"
```
yanivagman, line 25: Go consts naming convention doesn't use '_'

tracee-ebpf/tracee/userevents/system_info.go
```diff
@@ -0,0 +1,62 @@
...
+
+// CreateSystemInfoEvent collect information of the running machine and create an event that includes this information.
+func CreateSystemInfoEvent() (external.Event, error) {
+	systemInfoArgs := fetchSystemInfo()
+	systemInfoEvent := external.Event{
+		Timestamp:   int(time.Now().UnixNano()),
+		ProcessName: "tracee-ebpf",
+		EventID:     0,
```
yanivagman, line 33: Why not to use SystemInfoEventID?

tracee-ebpf/tracee/userevents/system_info.go
```diff
@@ -0,0 +1,62 @@
...
+
+const INIT_PROC_NS_DIR = "/proc/1/ns"
+
+const INIT_NAMESPACES_KEY = "initNamespaces"
+
+// CreateSystemInfoEvent collect information of the running machine and create an event that includes this information.
+func CreateSystemInfoEvent() (external.Event, error) {
+	systemInfoArgs := fetchSystemInfo()
+	systemInfoEvent := external.Event{
+		Timestamp:   int(time.Now().UnixNano()),
```
yanivagman, line 31: This is not in the format we use for Timesptamp

tracee-ebpf/tracee/printer.go
```diff
@@ -261,6 +261,7 @@ func (p *gobEventPrinter) Init() error {
 	p.outEnc = gob.NewEncoder(p.out)
 	gob.Register(external.Event{})
 	gob.Register(external.SlimCred{})
+	gob.Register(map[string]int{})
```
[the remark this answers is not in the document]
AlonZivony: Yes, because the printer will fail to print it otherwise.
In the start I didn't put it and only received the event's value in json but not in gob.
All complicated types should be registered this way unless they are part of registered type.

tracee-ebpf/tracee/userevents/system_info.go
AlonZivony: It is weird because it is when I checked on the output

AlonZivony:
If you can enable the option for "squash merging" it would be great :)

tracee-ebpf/tracee/userevents/system_info.go
AlonZivony: image
You can see here that the first row is the system_info event, and the timestamp is identical to its following events

yanivagman (changes requested):
Please don't mix consts package creation with this PR. Let's discuss the creation of such package in a different PR

AlonZivony:
After a talk with Yaniv we decided to postpone the extraction of the consts package and as a result of the user-mode-events package that has to import it.

tracee-ebpf/tracee/consts.go
```diff
@@ -563,6 +564,7 @@ var EventsIDToEvent = map[int32]EventConfig{
 	SecurityBPFEventID:            {ID: SecurityBPFEventID, ID32Bit: sys32undefined, Name: "security_bpf", Probes: []probe{{event: "security_bpf", attach: kprobe, fn: "trace_security_bpf"}}, Sets: []string{"lsm_hooks"}},
 	SecurityBPFMapEventID:         {ID: SecurityBPFMapEventID, ID32Bit: sys32undefined, Name: "security_bpf_map", Probes: []probe{{event: "security_bpf_map", attach: kprobe, fn: "trace_security_bpf_map"}}, Sets: []string{"lsm_hooks"}},
 	SecurityKernelReadFileEventID: {ID: SecurityKernelReadFileEventID, ID32Bit: sys32undefined, Name: "security_kernel_read_file", Probes: []probe{{event: "security_kernel_read_file", attach: kprobe, fn: "trace_security_kernel_read_file"}}, Sets: []string{"lsm_hooks"}},
+	SystemInfoEventID:             {ID: SystemInfoEventID, ID32Bit: sys32undefined, Name: "system_info", Probes: []probe{}, Sets: []string{"default"}},
```
yanivagman, line 567: Please remove from the default set

tracee-ebpf/tracee/user_mode_events.go
```diff
@@ -0,0 +1,61 @@
...
+		Args:        systemInfoArgs,
+	}
+	return systemInfoEvent, nil
+}
+
+// fetchSystemInfo Fetch info from the system that might be significant for later use or for signatures.
+func fetchSystemInfo() []external.Argument {
+	systemInfo := make([]external.Argument, 1)
+	systemInfo[0] = external.Argument{
+		ArgMeta: external.ArgMeta{Name: "initNamespaces", Type: "map[string]int"},
```
yanivagman, line 44: You can use the value given by EventsIDToParams map instead (better in case we change the argument name or type)

yanivagman (changes requested):
Two small comments.
Also not sure about the name system_info - as @itaysk suggeted - maybe we should call it init_namespaces instead (and if there will be other system info in the future, make other events as required)

yanivagman (approved):
LGTM

yanivagman:
@AlonZivony  please resolve conflicts so we will be able to merge

PR #948, opened 2021-08-23
tracee-ebpf: add stats to external

yanivagman:
As a preparation to taking printer out of the tracee package and moving it into main package, we need to export Stats.
Doing this, we will also be able to export Stats to other consumers of tracee-ebpf.

The exported struct will expose statsStore as defined in tracee.go:
```
type statsStore struct {
	eventCounter  counter
	errorCounter  counter
	lostEvCounter counter
	lostWrCounter counter
	lostNtCounter counter
}
```

### DIFF
--- a/tracee-ebpf/external/external.go
+++ b/tracee-ebpf/external/external.go
@@ -30,6 +30,14 @@ type Event struct {
 	Args                []Argument `json:"args"` //Arguments are ordered according their appearance in the original event
 }
 
+type Stats struct {
+	EventCount  int
+	ErrorCount  int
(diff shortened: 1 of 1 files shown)


itaysk (approved):

grantseltzer:
Is the goal of this to keep statsStore unexported but have a matching definition to reflect the bytes into?

yanivagman:
@grantseltzer yes, statsStore includes counters which we don't need to export.
I already have the changes ready to take printer out of the tracee package, finalizing them now and will push soon

PR #949, opened 2021-08-23
tracee-ebpf: use replace for the external package

yanivagman:
Currently, when we make changes that also require changing the external package, we first have to commit the changes in external, wait for it to be merged, fetch the new package and update go.mod, and only then submit the other changes. 
This is annoying, and can be avoided by replacing external in go.mod with a relative path

### DIFF
--- a/tracee-ebpf/go.mod
+++ b/tracee-ebpf/go.mod
@@ -2,6 +2,8 @@ module github.com/aquasecurity/tracee/tracee-ebpf
 
 go 1.16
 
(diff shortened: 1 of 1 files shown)


grantseltzer (approved):
Works for me. I don't see a situation where we'd want to use a divergent version of the external package.

Just want to note that this will not affect external projects that import tracee-ebpf, it will affect building tracee-ebpf itself. Which is good :+1:

PR #950, opened 2021-08-23
tracee-ebpf: fix network capture with latest libbpf

### LINKED ISSUE #947: Network capture not working in main
Seems that we have a regression with network capture feature since we bumped the libbpf version (to support external btf files).
```
sudo ./dist/tracee-ebpf --capture net=wlp2s0 -t e=open --capture clear-dir
libbpf: Kernel error message: Exclusivity flag on, cannot modify
2021/08/23 13:33:50 error creating Tracee: tc hook create: file exists
```

yanivagman:
According to  https://lore.kernel.org/bpf/CAMy7=ZXTiaX9xzNi5aOavwsf+mziJ=w-EcHH2f=cJmCGr3EPQA@mail.gmail.com/T/#u, getting an EEXIST error can happen when a qdisc already exists, and we should explicitly check for it - this PR does exactly that.

Note: We will still get the printed error message from libbpf. To remove it, we will need to figure out how to ignore it from libbpfgo libbpf_print_fn function.

Fix #947

### DIFF
--- a/tracee-ebpf/tracee/tracee.go
+++ b/tracee-ebpf/tracee/tracee.go
@@ -940,7 +940,9 @@ func (t *Tracee) attachTcProg(ifaceName string, attachPoint bpf.TcAttachPoint, p
 	hook.SetAttachPoint(attachPoint)
 	err = hook.Create()
 	if err != nil {
-		return nil, fmt.Errorf("tc hook create: %v", err)
+		if errno, ok := err.(syscall.Errno); ok && errno != syscall.EEXIST {
+			return nil, fmt.Errorf("tc hook create: %v", err)
+		}
 	}
 	prog, _ := t.bpfModule.GetProgram(progName)
 	if prog == nil {


rafaeldtinoco:
@yanivagman let me handle the libbpf_print_fn. I want to add a debug for it as well, so we can have a dump from the libbpf relocations work and that will make my life easier to check CO-RE issues in older kernels. Are you ok with that ?

rafaeldtinoco (approved):
LGTM as described in upstream thread.

yanivagman:
> @yanivagman let me handle the libbpf_print_fn. I want to add a debug for it as well, so we can have a dump from the libbpf relocations work and that will make my life easier to check CO-RE issues in older kernels. Are you ok with that ?

Sure, no problem. Anyways, I'm not sure how we can ignore this error from libbpf_print_fn, so if you have an idea for how to do it - go for it

PR #951, opened 2021-08-23
tracee-ebpf: fix build error after libbpfgo linting fixes

rafaeldtinoco:


rafaeldtinoco:
We need this fix because after :

https://github.com/aquasecurity/libbpfgo/pull/54#pullrequestreview-736515470

we will get a build error. This is also important for one relying on local libbpfgo latest for development.

yanivagman (commented):
I don't get any build error - is it local on your machine (probably with the recent changes of libbpfgo)?

Edit: Ignore that. Just saw your second comment

yanivagman (approved):

PR #952, opened 2021-08-23
tracee-ebpf: move printer to main package

yanivagman:
This PR moves the printer logic out of the tracee package into the main package, and decouples events generation from printing.
This will also allow a consumer of tracee-ebpf as a library to use channels to get stdout,stderr and tracee stats on close.

### DIFF
--- a/tracee-ebpf/tracee/events_pipeline.go
+++ b/tracee-ebpf/tracee/events_pipeline.go
@@ -11,7 +11,7 @@ import (
 	"github.com/aquasecurity/tracee/tracee-ebpf/external"
 )
 
-func (t *Tracee) runEventPipeline(done <-chan struct{}) error {
+func (t *Tracee) runEventPipeline(done <-chan external.Stats) error {
 	var errcList []<-chan error
 
 	// Source pipeline stage.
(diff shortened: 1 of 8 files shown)


rafaeldtinoco (approved):
LGTM. I have reviewed not only the commit but also all the printer logic, so I could better understand it, and it looks all good. I have also tested output logic with different output formats.

tracee-ebpf/tracee/events_pipeline.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/aquasecurity/tracee/tracee-ebpf/external"
 )
 
-func (t *Tracee) runEventPipeline(done <-chan struct{}) error {
```
itaysk, line 14: I'm not sure it's a better design, to make the done channel communicate stats to the caller. done should trigger a graceful shutdown as quickly and cleanly as possible. now every place that wants to be able to gracefully shut down (this comment applies to the different stages of tracee) needs to have access to stats, and do the work to return them (will probably be more complex in the future). Instead, the caller can just call get stats after shutdown, which makes more sense IMO

tracee-ebpf/tracee/events_pipeline.go
```diff
@@ -222,26 +245,12 @@ func (t *Tracee) prepareEventForPrint(done <-chan struct{}, in <-chan RawEvent)
 				errc <- err
 				continue
 			}
+			t.stats.eventCounter.Increment()
```
itaysk, line 248: why increment in "prepare event for print"? "eventsCounter should report processed events

yanivagman: Not exactly.
Until now, the count was incremented in printEvent, which was even after this stage.
I think that we want to keep it that way, because sometimes there are events that weren't chosen by the user, but we still process them (for example, to capture executed files). In fact, any essential event that was not chosen by the user shuoldn't be accounted.

tracee-ebpf/tracee/events_pipeline.go
yanivagman: Good point. Let me see how I can improve that

yanivagman: See #953

PR #965, opened 2021-08-27
WIP: Improving performance for Tracee Rules

simar7:
This PR aims to take the learnings from the PoC and apply them to tracee-rules. The improvements include:

1. Using WASM as OPA Target
2. Enabling Partial Evaluation of Rego policies 
3. [WIP] Adding an AIO Rego policy with rule selector
4. Bumping up OPA to latest
5. Reverting PreparedEvent changes


danielpacak:
> > > 1. Reverting PreparedEvent changes
> > 
> > 
> > To my understanding the benefit of this approach is tied to goroutine architecture? meaning, previously we had one goroutine per signature, and processing the input in each signature was costly so the prepared input improved that by preparing once for all signatures, and now with AIO the benefit of prepared input is diminished since it's one gorouting/rego engine anyway so might as well let rego prepare the event. If that's correct, I think we can keep the code in external that prepares the event, and just revert the usage of this code in the engine (instead of blindly reverting the commit). wdyt?
> 
> Not quite. The benefit of this approach was to reduce invoking the OPA JSON round tripper by preparing the event ourselves. With WASM this is no longer the case as explained here https://openpolicyagent.slack.com/archives/C1H19LW4F/p1629870274210000?thread_ts=1629336330.165900&cid=C1H19LW4F

For single threaded event processor that's correct. Otherwise @itaysk point is valid. Also notice that if end user loads majority only signatures implemented in Go he will observe significant performance degradation. So I'm wondering if this could be implemented as an optional optimization? WDYT? /cc @AlonZivony

simar7:
I'll slice this PR to thinner slices with a PR for each improvement.

AlonZivony:
> > > > 1. Reverting PreparedEvent changes
> > > 
> > > 
> > > To my understanding the benefit of this approach is tied to goroutine architecture? meaning, previously we had one goroutine per signature, and processing the input in each signature was costly so the prepared input improved that by preparing once for all signatures, and now with AIO the benefit of prepared input is diminished since it's one gorouting/rego engine anyway so might as well let rego prepare the event. If that's correct, I think we can keep the code in external that prepares the event, and just revert the usage of this code in the engine (instead of blindly reverting the commit). wdyt?
> > 
> > 
> > Not quite. The benefit of this approach was to reduce invoking the OPA JSON round tripper by preparing the event ourselves. With WASM this is no longer the case as explained here https://openpolicyagent.slack.com/archives/C1H19LW4F/p1629870274210000?thread_ts=1629336330.165900&cid=C1H19LW4F
> 
> For a single threaded event processor that's correct. Otherwise @itaysk point is valid. Also notice that if end user loads only signatures implemented in Go he will observe significant performance degradation. So I'm wondering if this could be implemented as an optional optimization? WDYT? /cc @AlonZivony

I agree with Daniel on this, I did some research on current tracee-rules performance with many go signatures and the conclusion was that this functions takes about 50% of the CPU usage of all used signatures **together**.
In my opinion it should be removed in some way - either make it optional or remove it totally.

Issue #960, opened 2021-08-26
refactor tracee-ebpf (Go)


itaysk:
today, most of tracee-ebpf is two huge packages: main and tracee. 
split tracee-ebpf's Go code into multiple packages. each package is self-contained and testable.
think hard about the names and APIs of those packages.

Some ideas:
1. create new package `tracing` instead of the current `tracee` which will contain the core tracee-ebpf engine
1. rename the struct `tracee` to `tracer`

main package:
1. flags handling - stay in package `main` for now. but extract from `main.go` - #1248 
1. ebpf compilations - remove #1239 
1. check capabilities utility function - create package `capabilities` - 
1. printer.go - create new package `eventprinter`. Factory function moves to `main.go` - #1266

tracee package:
1. tracee package needs to eventually go away, broken down into other packages
1. external package needs to eventually go away, the structs will be moved to respective packages
1. buckets cache - create new package `bucketscache` - #1249
1. stats - create new package `stats`. `external.Stats` will move to this package - #1250
1. configurations - create package `config` that includes filter, captureConfig ,outputConfig and related functions. (also create an interface for config and implement validators) - #1264
1. `argprinters.go` - move to `eventprinter` package (see above). remove dependency on tracee (can be a static function). replace `parse` misnomer with `print` (as the file name suggests). - #1265
1. containers - create new package `containers`. The code is already independent of tracee - 
1. create package `events` - #1268
    1. `events_decoder.go` - create new package `events`. rename to `events_decoding.go`
    1. `events_processor.go` - create new package `events`. rename to `event_processing.go`. `CopyFileByPath` function should move to some utlis package or duplicated.
    1. `events.go` - create new package `events` processEvents` should be minimal as possible, there is some code here that should be in `processEvent`.
1.  `filters.go` - create new package `tracingfilters` (consider the usage of pointers vs values in `Filter`) - #1269

TBD - `tarcee.go` and `consts.go`

related: #934, #789, #1131, #1098

danielpacak:
This is also a good opportunity to move `main.go` of tracee-ebpf to `./cmd/tracee-ebpf/main.go` as described in #1217

itaysk:
@mtcherni95 perhaps a good start is to review the existing code and discuss here which new packages you would like to create?

simar7:
[...]

PR #967, opened 2021-08-28
wait for tracee-ebpf to load

itaysk:
fix: #930 

A different approach than #944 

1. take readiness probe out of the library, into main
2. now we can decouple initializing tracee and opening the output file
3. refactor prepareOutput which used to create both outputConfig and Printer, now it's only preparing (parsing) input but not opening the file
4. move the file open only after tracee has finished loading
5. this allows tracee to actually open the file, instead of the shell (#944)
6. now the entrypoint can leverage these changes by simply reading/writing to the fifo like a normal file

better to review one commit at a time


itaysk:
P.S since main.go is a mess, the code is not very elegant. As part of #960 we can improve that as well

tracee-ebpf/main.go
```diff
@@ -382,58 +408,62 @@ func prepareOutput(outputSlice []string, containerMode bool) (tracee.OutputConfi
...
+			}
+		} else {
+			dir := filepath.Dir(errPath)
+			os.MkdirAll(dir, 0755)
+			printcfg.errFile, err = os.Create(errPath)
+			if err != nil {
+				return outcfg, printcfg, fmt.Errorf("failed to create output path: %v", err)
+			}
 		}
 	}
```
yanivagman, line 464: If you already took newEventPrinter out of this function, I don't see a reason to keep all of the oufFile and errFile logic here - we can move it to a different function (preparePrinter or something like that)

itaysk: some `--output` options will be relevant for both OutputConfig and printerConfig. if we split parsing into 2 functions like you suggest we would be parsing the same cli options in different places. I don't mind doing it if you think it's better, just raising this concern.

yanivagman: I don't see any line above (lines 426-462) that changes something in outcfg, but only printcfg.
What I suggest is to split into two functions (I don't mind about the names, as long as they are meaningful):
1. Looping the user input and returning outcfg, outPath and errPath
2. Using outPath and errPath for the creation of ourFile and errFile used by the printer

grantseltzer:
Moving this to milestone v0.6.3

tracee-ebpf/main.go
itaysk: > Looping the user input and returning outcfg, outPath and errPath

and also `printerKind`. 

So `prepareOutput` ended up returning: `outcfg, outpath, errpath, printerkind`, which is basically all the ingredients for `printerConfig`. It might as well just return the printerconfig already, no? If you still think it's valid, I can make the change

itaysk: I've implemented (I think) what you asked in this commit, tell me if this is what you want, and if so I'll add to the PR https://github.com/itaysk/tracee/commit/124996f4011f22a153394e1ac514b78ef619e12f

yanivagman: > So prepareOutput ended up returning: outcfg, outpath, errpath, printerkind, which is basically all the ingredients for printerConfig. It might as well just return the printerconfig already, no? If you still think it's valid, I can make the change

This is the situation today - it returns eventPrinter and OutputConfig, but you changed it in your PR.

I think that there are two things in this PR that we can split:
1. resolving "wait for tracee-ebpf to load"
2. Splitting preparePrinter and prepareOutput

If possible, I would avoid having (2) in this PR

itaysk: it's in this PR for a reason, see #3 in the PR description

yanivagman: Ok, so please push the latest change so we can review it here

itaysk:
rebased

yanivagman:
@itaysk I think you accidently pushed an older version of libbpf into this PR

entrypoint.sh
```diff
@@ -25,6 +25,29 @@ for arg in "$@"; do
...
-# start, and pass all remaining flags to tracee-rules
+# create named pipe for communicating between tracee-ebpf and tracee-rules
+tracee_pipe=/tmp/tracee/pipe
+rm -f $tracee_pipe && mkfifo $tracee_pipe
+
+# start tracee-ebpf in the background
 EVENTS=$($TRACEE_RULES_EXE --list-events)
-$TRACEE_EBPF_EXE --output=format:gob --output=option:parse-arguments --trace event=$EVENTS,mem_prot_alert | $TRACEE_RULES_EXE --input-tracee=file:stdin --input-tracee=format:gob $@
+echo "starting tracee-ebpf..."
+$TRACEE_EBPF_EXE --output=format:gob --output=option:parse-arguments --trace event=$EVENTS --output=out-file:$tracee_pipe &
```
yanivagman, line 35: missing mem_prot_alert event

yanivagman (commented):
Two small comments that are related to changes since this PR was made.
Otherwise LGTM

entrypoint.sh
itaysk: I removed it on purpose. it's included in the `$EVENTS`

itaysk:
@yanivagman removed libbpf and also the entrypoint tests which were obsolete

yanivagman (approved):

PR #968, opened 2021-08-29
tracee-ebpf: fix sched_process_exec argument types

yanivagman:
This fix is required for proper json unmarshalling in tracee-rules of the sched_process_exec event


itaysk:
> This fix is required for proper json unmarshalling

To clarify, the current type `const char *` is wrong regardless of the json use case, right? I'm asking because if you think the current type is more fitting, it's better to fix the unmarshaller instead

yanivagman:
The current type is wrong

itaysk (approved):

PR #970, opened 2021-08-29
add mknod lsm hooks

roikol:


tracee-ebpf/tracee/consts.go
```diff
@@ -168,6 +168,8 @@ const (
 	SecurityBPFEventID
 	SecurityBPFMapEventID
 	SecurityKernelReadFileEventID
+	SecurityPathMknodEventID
```
yanivagman, line 171: This event is defined under CONFIG_SECURITY_PATH (https://elixir.bootlin.com/linux/v5.12.19/source/security/security.c#L1082) and will not work on all distros.
If you add SecurityInodeMknodEventID - no need for this one

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -4013,6 +4015,112 @@ int BPF_KPROBE(trace_security_kernel_read_file)
...
+    set_buf_off(SUBMIT_BUF_IDX, sizeof(context_t));
+
+    context_t context = init_and_save_context(ctx, submit_p, SECURITY_INODE_MKNOD, 4 /*argnum*/, 0 /*ret*/);
+
+    u64 *tags = bpf_map_lookup_elem(&params_names_map, &context.eventid);
+    if (!tags) {
+        return -1;
+    }
+
+    struct path* dir = (struct path*)PT_REGS_PARM1(ctx);
```
yanivagman, line 4092: The first argument is of type struct inode, and not struct path (https://elixir.bootlin.com/linux/v5.12.19/source/security/security.c#L1223)

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -4013,6 +4015,112 @@ int BPF_KPROBE(trace_security_kernel_read_file)
...
+    if (off == NULL) {
+        return -1;
+    }
+    save_str_to_buf(submit_p, (void *)&string_p->buf[*off], DEC_ARG(0, *tags));
+
+    save_dentry_path_to_str_buf(string_p, dentry);
+    off = get_buf_off(STRING_BUF_IDX);
+    if (off == NULL) {
+        return -1;
+    }
```
yanivagman, line 4114: You should probably reset the off as it is already defined above, no need to get_buf_off

yanivagman (changes requested):

tracee-ebpf/tracee/consts.go
roikol: your'e right, i missed it. i'll remove this event.

tracee-ebpf/tracee/tracee.bpf.c
roikol: i'll remove the usage of the first param - it is not adding any useful data.

tracee-ebpf/tracee/argprinters.go
```diff
@@ -43,3 +43,37 @@ func PrintAlert(alert alert) string {
 
 	return res
 }
+
+func ParseMknodMode(mode uint16) string {
```
yanivagman, line 47: It's more appropriate to put this functino in libbpfgo helpers (https://github.com/aquasecurity/libbpfgo/blob/main/helpers/argumentParsers.go) instead of here

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -4013,6 +4014,51 @@ int BPF_KPROBE(trace_security_kernel_read_file)
...
+    set_buf_off(SUBMIT_BUF_IDX, sizeof(context_t));
+
+    context_t context = init_and_save_context(ctx, submit_p, SECURITY_INODE_MKNOD, 3 /*argnum*/, 0 /*ret*/);
+
+    u64 *tags = bpf_map_lookup_elem(&params_names_map, &context.eventid);
+    if (!tags) {
+        return -1;
+    }
+
+    struct dentry* dentry = (struct dentry*)PT_REGS_PARM2(ctx);
```
yanivagman, line 4038: Don't we care about the first argument? It tells us the directory on which this dentry is created

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -4013,6 +4014,51 @@ int BPF_KPROBE(trace_security_kernel_read_file)
...
+        return -1;
+    }
+
+    save_dentry_path_to_str_buf(string_p, dentry);
+    u32 *off = get_buf_off(STRING_BUF_IDX);
+    if (off == NULL) {
+        return -1;
+    }
+    save_str_to_buf(submit_p, (void *)&string_p->buf[*off], DEC_ARG(0, *tags));
+
```
yanivagman, line 4054: Extra line

tracee-ebpf/tracee/tracee.bpf.c
roikol: i think we don't care for this argument because the second argument is the full path - it contains the directory path

tracee-ebpf/tracee/argprinters.go
roikol: opened a PR in libbpgo

yanivagman: Why not using the ParseInodeMode() already defined in the helpers pacakge of libbpfgo?

roikol: ParseInodeMode is not suitable because in this case the mode is different format.

the way to parse the mode in this case is like the kernel does it here:
https://elixir.bootlin.com/linux/latest/source/fs/namei.c#L3711

yanivagman: How is this a different format? This is mode_t, and the values are the same

yanivagman (approved):

tracee-ebpf/tracee/tracee.bpf.c
rainkin1993: Can I ask why prefer security_path_mknod to security_inode_mknod?
**security_path_mknod** provide struct path argument which can get absolute file path with mounted file system, while **security_inode_mknod** only provide struct dentry/inode argument which is not aware of mounted file system and can only get path relative to mount point of filesystem.

yanivagman: security_path_mknod (and all other path lsm hooks) are enabled by CONFIG_SECURITY_PATH which is not available on all distros: https://elixir.bootlin.com/linux/v5.14.8/source/include/linux/security.h#L1760

PR #972, opened 2021-08-30
tracee-ebpf: mitigate deadcode optimization issue for 5.4 and less

rafaeldtinoco:
Fixes: #851

This commit removes a conditional branch from the logic of
raw_syscalls:sys_enter tracepoint. This solves issues with
dead code detection and CO-RE optimizations (most likely).

Bug: https://github.com/aquasecurity/tracee/issues/851
Reference: https://github.com/aquasecurity/tracee/issues/851#issuecomment-903074596
Reference: https://x-lore.kernel.org/stable/20210821203108.215937-1-rafaeldtinoco@gmail.com/#t

### DIFF
--- a/tracee-ebpf/tracee/tracee.bpf.c
+++ b/tracee-ebpf/tracee/tracee.bpf.c
@@ -2028,14 +2028,9 @@ if (CONFIG_ARCH_HAS_SYSCALL_WRAPPER) {
         args_tmp.args[4] = PT_REGS_PARM5(&regs);
         args_tmp.args[5] = PT_REGS_PARM6(&regs);
     }
-} else { // NO CONFIG_ARCH_HAS_SYSCALL_WRAPPER
-    args_tmp.args[0] = ctx->args[0];
-    args_tmp.args[1] = ctx->args[1];
-    args_tmp.args[2] = ctx->args[2];
-    args_tmp.args[3] = ctx->args[3];
-    args_tmp.args[4] = ctx->args[4];
-    args_tmp.args[5] = ctx->args[5];
-} // END CONFIG_ARCH_HAS_SYSCALL_WRAPPER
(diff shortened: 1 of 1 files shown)


tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2029,14 +2032,7 @@ if (CONFIG_ARCH_HAS_SYSCALL_WRAPPER) {
...
     }
-} else { // NO CONFIG_ARCH_HAS_SYSCALL_WRAPPER
-    args_tmp.args[0] = ctx->args[0];
-    args_tmp.args[1] = ctx->args[1];
-    args_tmp.args[2] = ctx->args[2];
-    args_tmp.args[3] = ctx->args[3];
-    args_tmp.args[4] = ctx->args[4];
-    args_tmp.args[5] = ctx->args[5];
-} // END CONFIG_ARCH_HAS_SYSCALL_WRAPPER
+}
```
yanivagman, line 2035: You can move `bpf_probe_read(&args_tmp.args[0], sizeof(6 * sizeof(u64)), (void *)&ctx->args[0]); `here

rafaeldtinoco: I tried to be cautious here by avoiding the else branch (since inevitably args_tmp would be rewritten entirely if CONFIG_ARCH_HAS_SYSCALL_WRAPPER is defined).

rafaeldtinoco: Maybe its not a deal any longer when we use bpf_probe_read() for the verifier.. and moving it to else{} will avoid a copy every syscall wrap... let me change it and re-do some tests.

yanivagman (approved):

Issue #946, opened 2021-08-23
tracee-ebpf: Create sub-packages


AlonZivony:
Currently all the logic is located in the tracee package.
Now that the logic is separated from the tracee.go to other files, I think the next step is to create packages for each parent logic under tracee directory. This way the code structure will be more defined and each package will be more testable (because of less dependencies).

yanivagman:
Closing this in favor of #960

PR #974, opened 2021-08-30
Moved consts files to new module

AlonZivony:


yanivagman:
@AlonZivony as we discussed offline, we can keep the consts where they are today, and add a new event id enum to external that will map the different event ids to some global event ids. I opened https://github.com/aquasecurity/tracee/issues/1098 to track that.

Do you mind if we close this PR?

yanivagman:
I'm not saying that we shouldn't split tracee-ebpf into packages, and we even have an open issue for that (#946, opened by you). But spliting into packages is something we should discuss first - to which packages do we want to split and why? I also don't think that a "consts" package is something that we want to have - it is too wide.
So my suggestion is to close this PR (anyways it is out of date and needs to be rebased as many things changed since), and to open a discussion (in the discussions tab) so we can decide about the packages we want to split tracee-ebpf to.

AlonZivony:
I agree to what you say.
I close the PR.

PR #979, opened 2021-09-01
feat: Add flag for Partial Eval

simar7:
Adds a new flag `--partial-eval` that can be passed in for turning on partial evaluation of rego rules.

Signed-off-by: Simar <simar@linux.com>

PR #980, opened 2021-09-01
feat: Add flag for Runtime Target

simar7:
Adds a new flag `--rego-runtime-target` that can be specified with potential runtime targets for rego rules: `rego` or `wasm`.

Requires: https://github.com/aquasecurity/tracee/pull/979

Signed-off-by: Simar simar@linux.com

### DIFF
--- a/tracee-rules/main.go
+++ b/tracee-rules/main.go
@@ -9,8 +9,12 @@ import (
 	"net/http/pprof"
 	"os"
 	"os/signal"
+	"strings"
 	"syscall"
 	"time"
+
+	"github.com/open-policy-agent/opa/compile"
+	_ "github.com/open-policy-agent/opa/features/wasm"
 
 	"github.com/aquasecurity/tracee/tracee-rules/engine"
 	"github.com/aquasecurity/tracee/tracee-rules/types"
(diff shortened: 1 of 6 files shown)


itaysk:
Can you please prefix those rego related flags with "rego-"?

tracee-rules/signatures/rego/regosig/traceerego.go
```diff
@@ -35,7 +35,7 @@ const queryMetadata string = "data.%s.__rego_metadoc__"
 const packageNameRegex string = `package\s.*`
 
 // NewRegoSignature creates a new RegoSignature with the provided rego code string
-func NewRegoSignature(regoCodes ...string) (types.Signature, error) {
+func NewRegoSignature(target string, partialEval bool, regoCodes ...string) (types.Signature, error) {
```
danielpacak, line 38: NIT Not sure how many args is too much, but at some point we may consider [functional options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) pattern for friendly API

tracee-rules/main.go
```diff
@@ -59,7 +62,17 @@ func main() {
 				}()
 			}
 
-			sigs, err := getSignatures(c.String("rules-dir"), c.StringSlice("rules"))
+			var target string
+			switch strings.ToLower(c.String("target")) {
+			case "wasm":
```
danielpacak, line 67: I think that with OPA v0.32.0 we have to import `wasm` package for side effects. In previous version they used Go build tags. Did you add it somewhere?

```go
import (
	_ "github.com/open-policy-agent/opa/features/wasm"
)
```

danielpacak (changes requested):

simar7:
Let's merge https://github.com/aquasecurity/tracee/pull/979 first then I'll rebase this

tracee-rules/main.go
simar7: Yeah good catch - I had changed this in the benchmarks locally but forgot to do so here. I added it in now.

simar7:
Rebased on latest main

tracee-rules/signatures/rego/regosig/traceerego.go
simar7: IMO think 3 (considering regoCodes is a variadic) is still fine, although we could take the functional approach but could do as a separate PR.

danielpacak: Yup. We can refactor to use functional Opts later on.

danielpacak (approved):

PR #982, opened 2021-09-01
tracee-ebpf: fix process tree disabled

yanivagman:
After merging #836 , tracee-ebpf doesn't work without using the tree filter (always-on).
This PR fixes it.


grantseltzer (approved):
LGTM, thanks @yanivagman

Issue #986, opened 2021-09-02
json unmarshaling issues


itaysk:
reported by @idanr1986 

- [x] empty string array

```
panic: interface conversion: interface {} is nil, not []interface {}
```

- [ ] commit creds

```
commit_creds - panic: interface conversion: interface {} is map[string]interface {}, not external.SlimCred
```

- [ ] bytes

```
magic_write - panic: interface conversion: interface {} is string, not []uint8
```

itaysk:
we probably need to cover all of the `Argument.Type`s

itaysk:
```
❯ awk 'BEGIN {FS="\""} /Type/ { print $2 }' /home/vagrant/dev/tracee/tracee-ebpf/tracee/consts.go | sort | uniq

aio_context_t
alert_t
cap_user_header_t
char*
const char *
const char*
const clockid_t
const sigset_t*
const stack_t*
const struct timespec*
const struct timeval*
const void*
dev_t
gid_t
gid_t*
int
int*
int[2]
io_context_t
key_t
long
mode_t
mqd_t
pid_t
sigset_t*
slim_cred_t
struct __sysctl_args*
struct clone_args*
struct perf_event_attr*
struct pollfd*
struct robust_list_head*
struct rseq*
struct sysinfo*
struct timeval*
struct timex*
struct tms*
struct user_desc*
struct utsname*
time_t*
timer_t
u32
u64
uid_t
uid_t*
unsigned int
unsigned int*
unsigned long
void*
```

grantseltzer:
How can I reproduce these issues? I created a test signature that consumes commit_creds, built it, and ran it like so:

`sudo ../tracee-ebpf/dist/tracee-ebpf --trace event=commit_creds --output format:json | ./dist/tracee-rules --input-tracee file:stdin --input-tracee format:json`

I didn't run into this interface conversion panic

mtcherni95:
@grantseltzer any update regarding reproduction?

yanivagman:
> I'm not able to reproduce, no.

@grantseltzer have you seen #988? To me it seems as something that needs to be fixed.
But we can't merge it as is, and first need to handle #1029 to avoid regressions as @itaysk noted.

yanivagman:
last status was unreproducible

PR #988, opened 2021-09-02
improve json unmarshaling

itaysk:
partial fix to: #986

### DIFF
--- a/tracee-ebpf/external/external.go
+++ b/tracee-ebpf/external/external.go
@@ -181,6 +181,13 @@ func (arg *Argument) UnmarshalJSON(b []byte) error {
 			return err
 		}
 		arg.Value = tmp
+	case "slim_cred_t":
+		var tmp SlimCred
+		err := json.Unmarshal(partialArg.Value, &tmp)
+		if err != nil {
+			return err
+		}
+		arg.Value = tmp
 	default:
 		return fmt.Errorf("unrecognized argument type: %v", partialArg.Type)
 	}
--- a/tracee-ebpf/external/external.go
+++ b/tracee-ebpf/external/external.go
@@ -149,7 +149,6 @@ func (arg *Argument) UnmarshalJSON(b []byte) error {
 		if err != nil {
 			return err
 		}
-		arg.Value = int32(tmp)
 		arg.Value = tmp
 	case "unsigned int", "u32", "mode_t", "dev_t":
 		tmp, err := strconv.ParseUint(string(partialArg.Value), 10, 32)
--- a/tracee-ebpf/external/external.go
+++ b/tracee-ebpf/external/external.go
@@ -137,6 +137,7 @@ func (arg *Argument) UnmarshalJSON(b []byte) error {
 	if string(partialArg.Value) == "null" {
 		return nil
 	}
+	partialArg.Value = bytes.Trim(partialArg.Value, "\"")
 	switch partialArg.Type {
 	case "int", "pid_t", "uid_t", "gid_t", "mqd_t", "clockid_t", "const clockid_t", "key_t", "key_serial_t", "timer_t":
 		tmp, err := strconv.ParseInt(string(partialArg.Value), 10, 32)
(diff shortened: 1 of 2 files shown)


tracee-ebpf/external/external_test.go
```diff
@@ -20,14 +20,14 @@ func TestEventUnmarshalJSON(t *testing.T) {
...
+		// 	"hostprocessid":12434,"hostthreadid":12434,"hostparentprocessid":23921,"userid":1000,"mountnamespace":4026531840,
+		// 	"pidnamespace":4026531836,"processname":"strace","hostname":"ubuntu","eventid":"101","eventname":"ptrace",
+		// 	"argsnum":4,"returnvalue":0,"args":[{"name":"request","type":"long","value":"ptrace_seize"},
+		// 	{"name":"pid","type":"pid_t","value":12435},{"name":"addr","type":"void*","value":"0x0"},{"name":"data","type":"void*","value":"0x7f6f1eb44b83"}]}`,
+		// 	expect: Event{Timestamp: 26018249532, ProcessID: 12434, ThreadID: 12434, ParentProcessID: 23921, HostProcessID: 12434, HostThreadID: 12434, HostParentProcessID: 23921, UserID: 1000, MountNS: 4026531840, PIDNS: 4026531836, ProcessName: "strace", HostName: "ubuntu", EventID: 101, EventName: "ptrace", ArgsNum: 4, ReturnValue: 0, Args: []Argument{{ArgMeta: ArgMeta{Name: "request", Type: "long"}, Value: "ptrace_seize"}, {ArgMeta: ArgMeta{Name: "pid", Type: "pid_t"}, Value: int32(12435)}, {ArgMeta: ArgMeta{Name: "addr", Type: "void*"}, Value: "0x0"}, {ArgMeta: ArgMeta{Name: "data", Type: "void*"}, Value: "0x7f6f1eb44b83"}}},
+		// },
```
simar7, line 30: Should we add a case since we're here to handle `long` type?

tracee-ebpf/external/external.go
```diff
@@ -123,64 +124,75 @@ type ArgMeta struct {
...
-		for i, v := range argValue {
-			tmp[i] = fmt.Sprint(v)
+		arg.Value = int32(tmp)
+	case "long":
+		tmp, err := strconv.ParseInt(string(partialArg.Value), 10, 64)
+		if err != nil {
+			return err
+		}
+		arg.Value = int32(tmp)
+		arg.Value = tmp
```
yanivagman, line 153: I believe 152 should be removed?

tracee-ebpf/external/external_test.go
yanivagman: @itaysk can you please explain why you commented out this test?

itaysk: because the test case it wrong. I'll replace it

itaysk:
I've fixed those issues, but I should note that pretty-printed arg values (e.g `O_WRONLY`) will break. I think we need to communicate if it's raw or printed in the arg type

yanivagman:
> I've fixed those issues, but I should note that pretty-printed arg values (e.g O_WRONLY) will break. I think we need to communicate if it's raw or printed in the arg type

Here's another option - pass `argMetas` slice (from `processEvents()` func) as a pointer argument to `prepareArgs()` and change the argument type to the correct value when required

yanivagman (approved):

yanivagman:
> Here's another option - pass argMetas slice (from processEvents() func) as a pointer argument to prepareArgs() and change the argument type to the correct value when required

I opened #1029 to track this one

yanivagman:
@itaysk what's left for #986 after merging this one?

itaysk:
I don't think we can merge this before #1029 right? I've talked with @grantseltzer about looking into #986 generally

yanivagman:
> I don't think we can merge this before #1029 right?

Why not? It resolves current bugs that we have. If #1029 will require changes, it is the responsibility of this future PR to fix it

itaysk:
it creates a regression AFAIU

> pretty-printed arg values (e.g O_WRONLY) will break

yanivagman:
@itaysk This PR depends on completing #1029 , right?

yanivagman:
@itaysk can you please rebase your changes?

rafaeldtinoco (changes requested):
Needs rebasing (per last request). Reflecting status.

rafaeldtinoco:
After we have protobuf implemented (and a protobuf wrapper from tracee event to protobuf, so go-cel can work, can be found at https://github.com/aquasecurity/tracee/pull/1766/files#diff-af1a3c488c6c0d521bd322a92cdb79b2eaaf758ee7245db25b5742bb3d8aa32aR95) we will start sorting all the types when marshaling / un-marshaling tracee events since those will be strongly typed.

rafaeldtinoco:
Closing this due to inactivity.

PR #989, opened 2021-09-02
tracee-ebpf: move filters logic to a new file

yanivagman:


rafaeldtinoco:
Reviewing this now...

rafaeldtinoco (commented):
You will need to rebase to latest libbpfgo. Needed change (for now):

```
func (t *Tracee) initBPF() error {
	var err error

	t.bpfModule, err = bpf.NewModuleFromBufferArgs(bpf.NewModuleArgs{
		BTFObjPath: t.config.BTFObjPath,
		BPFObjBuff: t.config.BPFObjBytes,
		BPFObjName: t.config.BPFObjPath,
	})
	if err != nil {
		return err
	}
```

tracee-ebpf/tracee/filters.go
```diff
@@ -0,0 +1,563 @@
...
+			if err != nil {
+				return
+			}
+			if ppid == 1 {
+				return
+			}
+
+			if shouldBeTraced, ok := filter.PIDs[uint32(ppid)]; ok {
+				trace := boolToUInt32(shouldBeTraced)
+				processTreeBPFMap.Update(unsafe.Pointer(&pid), unsafe.Pointer(&trace))
```
rafaeldtinoco, line 549: Should we check for errors here ?

tracee-ebpf/tracee/filters.go
```diff
@@ -0,0 +1,563 @@
...
+				return
+			}
+			fn(uint32(ppid))
+		}
+		fn(uint32(pid))
+	}
+
+	for pid, shouldBeTraced := range filter.PIDs {
+		trace := boolToUInt32(shouldBeTraced)
+		processTreeBPFMap.Update(unsafe.Pointer(&pid), unsafe.Pointer(&trace))
```
rafaeldtinoco, line 559: and here...

rafaeldtinoco (approved):
There are some redundant type conversions (which is ok). A more complete linter complains about same struct functions receivers (function arguments of the same type) having different names in each "filters" function, if you care, if not then after a rebase you will be good.

yanivagman:
@rafaeldtinoco I didn't touch the code in initBPF(), and specifically not NewModuleFromBufferArgs().
Also, rebasing locally on main didn't show any conflicts

rafaeldtinoco:
> @rafaeldtinoco I didn't touch the code in initBPF(), and specifically not NewModuleFromBufferArgs().
> Also, rebasing locally on main didn't show any conflicts

That is perk of having libbpfgo being developed in parallel. We have to choose to keep developing tracee with an older libbpfgo version versus trying to keep up with latest libbpfgo version (which will have that problem).

yanivagman:
> That is perk of having libbpfgo being developed in parallel. We have to choose to keep developing tracee with an older libbpfgo version versus trying to keep up with latest libbpfgo version (which will have that problem)

I'm using libbpfgo version as defined in go.mod of the main branch. I don't think we should care now about what will happen when we will update to the most recent libbpfgo - this should be resolved in the PR which will update to the latest libbpfgo.

> Should we check for errors here ?

I only moved the code from one file to another. I didn't change anything accept changing some of the functions receivers. You probably right that we should check for errors, but it can wait for a future PR

PR #993, opened 2021-09-04
feat: Add AIO Rego rule selector rule

simar7:
This adds the ability to specify a `--rego-aio-enable` option to optionally add an AIO rego rule.


tracee-rules/signatures/rego/regosig/traceerego.go
```diff
@@ -62,13 +66,22 @@ func NewRegoSignature(regoCodes ...string) (types.Signature, error) {
...
-		rego.Query(fmt.Sprintf(queryMatch, pkgName)),
-	).PrepareForEval(context.TODO())
-	if err != nil {
-		return nil, err
+	var rerr error
+	if aioEnabled {
+		res.matchPQ, rerr = rego.New(
+			rego.Compiler(res.compiledRego),
+			rego.Query(queryMatchAll),
+		).PrepareForEval(context.TODO())
```
simar7, line 74: Note: I believe we need to rethink this as currently we create a Rego engine per signature and in the case of an AIO we need 1 engine with all rules with a selector (query) on the AIO rule. Thoughts and comments appreciated.

simar7: I fixed this here: https://github.com/aquasecurity/tracee/pull/993/commits/b81da5a8ac88666105a8eae88fcb7d102c49123b

danielpacak (changes requested):
As discussed lets make it a new implementation of `types.Signature` interface. This will allow us to apply the TDD methodology. Another advantage is that we won't increasing complexity of existing `regosig.RegoSignature`.

tracee-rules/main.go
```diff
@@ -168,6 +168,10 @@ func main() {
 				Name:  "rego-enable-parsed-events",
 				Usage: "enables pre parsing of input events to rego prior to evaluation",
 			},
+			&cli.BoolFlag{
+				Name:  "rego-aio-enable",
```
danielpacak, line 172: NIT: Maybe we can keep this flag shorter without loosing context by replacing `--rego-aio-enable` with `--rego-aio`. I was also thinking if `--rego-parse-events` can be used instead of `--rego-enable-parsed-events`. WDYT?

tracee-rules/main.go
```diff
@@ -168,6 +168,10 @@ func main() {
...
+				Usage: "enable rego AIO rule for rule indexing and evaluation",
```
danielpacak, line 173: I think rule indexing is not related to AIO but more to partial evaluation. Here AIO is about compiling all signatures once as a single (huge) module instead of compiling N (smaller) modules.

tracee-rules/signature.go
```diff
@@ -90,6 +89,10 @@ func findRegoSigs(target string, partialEval bool, dir string) ([]types.Signatur
 			return nil
 		}
 
+		if !aioEnabled && isAIO(d.Name()) {
```
danielpacak, line 92: What is this condition for? Don't we load aio.rego as embedded resource?

tracee-rules/signatures/rego/regosig/aio.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		return nil, err
+	}
+	res.selectedEvents, err = res.getSelectedEvents(pkgName)
+	if err != nil {
+		return nil, err
+	}
+	return &res, nil
+
+	return nil, nil
```
danielpacak, line 76: Unreachable code

tracee-rules/signatures/rego/regosig/aio.go
```diff
@@ -0,0 +1,77 @@
...
+			rego.Target(o.Target),
+			rego.Compiler(res.compiledRego),
+			rego.Query(queryMatchAll),
+		).PrepareForEval(ctx)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	res.metadata, err = res.getMetadata(pkgName)
```
danielpacak, line 66: Why don't we evaluate `__rego_metadoc_all__` rule here? This is also introducing tight coupling between the two implementations.

tracee-rules/signatures/rego/regosig/aio.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		return nil, err
+	}
+	res.selectedEvents, err = res.getSelectedEvents(pkgName)
```
danielpacak, line 70: DITTO: Why do we eval tracee_selected_events instead of tracee_selected_events_all ?

tracee-rules/signatures/rego/regosig/aio.rego
```diff
@@ -0,0 +1,40 @@
...
+    "description": "This rule indexes all loaded Rego rules via one."
+}
+
+__rego_metadoc_all__[id] = resp {
+	some i
+		resp := data.tracee[i].__rego_metadoc__
+		id := resp.id
+}
+
+tracee_selected_events[eventSelector] {
```
danielpacak, line 16: I think we should use `tracee_selected_events_all`. Otherwise we don't have any filtering after all.

tracee-rules/signatures/rego/regosig/aio.rego
```diff
@@ -0,0 +1,40 @@
...
+	some i
+		resp := data.tracee[i].tracee_selected_events
+		metadata := data.tracee[i].__rego_metadoc__
+		id := metadata.id
+}
+
+# Returns the map of signature identifiers to values matching the input event.
+tracee_match_all[id] = resp {
+	some i
+		# TODO Try adding event selector
```
danielpacak, line 34: Let's remove this TODO and comments below. What's suggested in the comment is not compatible with the existing implementation

tracee-rules/signatures/rego/regosig/traceerego.go
```diff
@@ -183,24 +190,52 @@ func (sig *RegoSignature) OnEvent(e types.Event) error {
...
 		case map[string]interface{}:
+			values, ok := results[0].Expressions[0].Value.(map[string]interface{})
+			if ok && len(values) <= 0 {
+				return nil
+			}
+		case bool:
+			sig.dispatch(results[0].Expressions[0].Value, ee)
+		}
+
+		// For AIO: result set can be an empty set of length=1, so we need to check value
```
danielpacak, line 204: I think our goal to have new implementation of types.Signature for AIO was to really don't touch existing RegoSignature.OnEvent method. Otherwise it's not different from adding aioEnabled arg to the NewRegoSignature constructor

tracee-rules/signatures/rego/regosig/traceerego.go
```diff
@@ -183,24 +190,52 @@ func (sig *RegoSignature) OnEvent(e types.Event) error {
...
+			return nil
+		}
+
+		sig.dispatch(values, ee)
+
+	}
+	return nil
+}
+
+func (sig *RegoSignature) dispatch(val interface{}, ee tracee.Event) {
```
danielpacak, line 216: The tracee_match_all rule returns a map from signature id to values. Where do we have logic to iterate through that map and create 0 to many findings?
Do we have tests for that in aio_test.go?

danielpacak (changes requested):

tracee-rules/main.go
simar7: Sure, updated.

tracee-rules/signature.go
simar7: Good catch. This is cruft from an older implementation I had. I've removed this and updated tests accordingly.

simar7: https://github.com/simar7/tracee/commit/f7bedd3265e437459106b059a65572002914032c

tracee-rules/signatures/rego/regosig/aio.go
simar7: thanks, cruft.

simar7:
Deprecated in favour of https://github.com/aquasecurity/tracee/pull/1015

PR #996, opened 2021-09-05
update docs

yanivagman:
This PR fixes the following in the documentation:
1. Added `pid=host` to the docker command. This is required by some of the features used by tracee (tree-filter and capture exec)
2. Added `/tmp` mount to the docker command. This is required for saving captured artifacts and to speed-up init time when running on a non-BTF enabled kernel (to save the compiled bpf file)
3. Explicitly write the docker command used to run Tracee on a non-BTF enabled kernels
4. Update the required capabilities with the correct values
5. Some other small fixes

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -22,20 +22,23 @@ The full documentation of Tracee is available at [https://aquasecurity.github.io
 
 Before you proceed, make sure you follow the [minimum requirements for running Tracee](https://aquasecurity.github.io/tracee/dev/install/prerequisites/).
 
-If running on BTF enabled kernel:
+If running on __BTF enabled kernel__:
 
 ```bash
-docker run --name tracee --rm --privileged -it aquasec/tracee:latest
+docker run --name tracee --rm --pid=host --privileged -v /tmp/tracee:/tmp/tracee -it aquasec/tracee:latest
 ```
 
-Note: __Running bpf requires access to the kernel configuration file. Depending on the linux distribution this can be in either `/proc/config.gz` (which docker mounts by default) or `/boot/config-$(uname -r)` (which must be mounted explicitly)._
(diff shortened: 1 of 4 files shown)


Readme.md
```diff
@@ -22,21 +22,24 @@ The full documentation of Tracee is available at [https://aquasecurity.github.io
 
 Before you proceed, make sure you follow the [minimum requirements for running Tracee](https://aquasecurity.github.io/tracee/dev/install/prerequisites/).
 
-If running on BTF enabled kernel:
+If running on __BTF enabled kernel__:
 
 ```bash
-docker run --name tracee --rm --privileged -it aquasec/tracee:latest
```
itaysk, line 28: /tmp might be too wide, why not /tmp/tracee?

docs/index.md
```diff
@@ -18,16 +18,23 @@ Check out the [Tracee video hub](https://info.aquasec.com/ebpf-runtime-security)
...
 ```
 
-Note: __Running bpf requires access to the kernel configuration file. Depending on the linux distribution this can be in either `/proc/config.gz` (which docker mounts by default) or `/boot/config-$(uname -r)` (which must be mounted explicitly)._
+> Note: Running on BTF enabled kernel requires access to the kernel configuration file. Depending on the linux distribution this can be in either `/proc/config.gz` (which docker mounts by default) or `/boot/config-$(uname -r)` (which must be mounted explicitly).
 
-This will run Tracee with the portable CO:RE bpf object and default settings. It will start reporting detections to standard output.  
-In order to simulate a suspicious behavior, you can run `strace ls` in another terminal, which will trigger the "Anti-Debugging" signature, which is loaded by default.
+If running on __BTF disabled kernel__:
+```bash
+docker run --name tracee --rm --pid=host --privileged -v /tmp:/tmp -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro -v /tmp/tracee:/tmp/tracee -it aquasec/tracee:latest
```
itaysk, line 31: no need for both /tmp and /tmp/tracee

grantseltzer (approved):
LGTM

PR #998, opened 2021-09-06
add argument 'type' to security_kernel_read_file event

roikol:


tracee-ebpf/tracee/argprinters.go
```diff
@@ -176,7 +176,68 @@ func (t *Tracee) prepareArgs(ctx *context, args map[string]interface{}) error {
...
+			1: "firmware",
+			2: "firmware",
+			3: "firmware",
+			4: "kernel-module",
+			5: "kexec-image",
+			6: "kexec-initramfs",
+			7: "security-policy",
+			8: "x509-certificate",
+		}
+	} else if osInfo.CompareOSBaseKernelRelease("5.8.18") == 0 || (osInfo.CompareOSBaseKernelRelease("4.17.19") == -1 && osInfo.CompareOSBaseKernelRelease("5.7.1") == 1) {
```
yanivagman, line 225: Tracee minimal kernel requirement is 4.18, so 4.17.19 check can be removed

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -4027,11 +4027,16 @@ int BPF_KPROBE(trace_security_kernel_read_file)
...
+    context_t context = init_and_save_context(ctx, submit_p, SECURITY_KERNEL_READ_FILE, 4 /*argnum*/, 0 /*ret*/);
 
     struct file* file = (struct file*)PT_REGS_PARM1(ctx);
     dev_t s_dev = get_dev_from_file(file);
     unsigned long inode_nr = get_inode_nr_from_file(file);
+    enum kernel_read_file_id type_id = (enum kernel_read_file_id)PT_REGS_PARM2(ctx);
+//    char type_id_str_buffer[50];
+//    const char *type_id_str = kernel_read_file_str[type_id];
+//    bpf_probe_read_str(type_id_str_buffer, 50, type_id_str);
+//    const char *type_id_str = kernel_read_file_id_str(type_id);
```
yanivagman, line 4039: Please remove comments

tracee-ebpf/tracee/argprinters.go
```diff
@@ -176,7 +176,68 @@ func (t *Tracee) prepareArgs(ctx *context, args map[string]interface{}) error {
...
+				return err
+			}
+			args["type"] = typeIdStr
+		}
 	}
 
 	return nil
 }
+
+func ParseKernelReadFileId(id int32) (string, error) {
```
yanivagman, line 192: This function reads the osInfo and initializes kernelReadFileIdStrs accordingly on every call, while osInfo shouldn't change between subsequent calls. Instead, kernelReadFileIdStrs can be initialized once during init.

roikol: i've added it as part of the Tracee struct. kernelReadFileIdStrs was too specific, so i've added a more generic argParsers field to the Tracee struct, and kernelReadFileIdStrs is part of it. 
now this is initialized when a new Tracee instance is created.

as there is only one argParser currently (kernelReadFileIdStrs), we can check that the event security_kernel_read_file was chosen, before initializing the kernelReadFileIdStrs.

let me if this is what you had in mind.

yanivagman: Actually, what I had in mind is using either `sync.Once` or `init()`

roikol: ohh okay.
i've updated it now to use `init()`.

tracee-ebpf/tracee/argprinters.go
```diff
@@ -176,7 +176,72 @@ func (t *Tracee) prepareArgs(ctx *context, args map[string]interface{}) error {
 		if cmd, isInt32 := args["cmd"].(int32); isInt32 {
 			args["cmd"] = helpers.ParseBPFCmd(cmd)
 		}
+	case SecurityKernelReadFileEventID:
+		if readFileId, isUint32 := args["type"].(uint32); isUint32 {
+			typeIdStr, err := ParseKernelReadFileId(int32(readFileId))
+			if err != nil {
+				return err
+			}
```
yanivagman, line 184: We currently don't return an error if an argument parsing failed (also see other argument parsers in this function) - let's just return an empty string in this case

roikol: i want to use the enum value if we couldn't translate to str (so this data won't be totally lost). 
so i changed it to be updated only if `err == nil`

yanivagman (approved):
LGTM

PR #1001, opened 2021-09-08
Check os-release file for rhel or centos string

### LINKED ISSUE #942: Broken makefile because of RHEL_RELEASE_CODE
I can no longer build tracee-ebpf with the makefile

```
fedo grant in ~> tracee-ebpf on ~> staticbuild
[*] make                                                                                130 ↵
/bin/sh: << 8 | : syntax error: operand expected (error token is "<< 8 | ")
clang \
	-I dist/libbpf/usr/include/bpf \
	-I 3rdparty/include \
	-I tracee/x86_64 \
	-DCORE \
	-D__TARGET_ARCH_x86 \
	-D__BPF_TRACING__ \
	-D__VM_LINUX_H__ \
	-D__KERNEL__ \
	-DLINUX_VERSION_CODE=331014 \
	-DRHEL_RELEASE_CODE= \
	-target bpf \
	-O2 -c -g -o dist/tracee.bpf.core.o tracee/tracee.bpf.c
tracee/tracee.bpf.c:323:24: error: invalid token at start of a preprocessor expression
#if (RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(8, 0))
                       ^
1 error generated.
make: *** [Makefile:129: dist/tracee.bpf.core.o] Error 1

```

The content of my /etc/redhat-release file is:

```
[*] cat /etc/redhat-release                                                              
Fedora release 33 (Thirty Three)
```

uname
```
[*] uname -a                                                                            130 ↵
Linux localhost.localdomain 5.13.6-100.fc33.x86_64 #1 SMP Wed Jul 28 15:30:10 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```

grantseltzer:
This attempts to fix #942 

My rationale here is that this is only to make centos8 and rhel8 work, so I normalize the content of /usr/lib/os-release (even this has variance from centos to rhel) and only set RHEL_RELEASE_CODE if the value of VERSION_ID is rhel or centos. 

This isn't necessary for fedora as all supported versions have kernel version 5.8+ 

### DIFF
--- a/tracee-ebpf/Makefile
+++ b/tracee-ebpf/Makefile
@@ -39,9 +39,12 @@ DOCKER_BUILDER_KERN_SRC ?= $(if $(shell readlink $(KERN_SRC_PATH)),$(shell readl
 DOCKER_BUILDER_KERN_SRC_MNT ?= $(dir $(DOCKER_BUILDER_KERN_SRC))
 LINUX_VERSION_CODE := $(shell uname -r | awk '{split($$0,a,"."); split(a[3], b, "-"); print lshift(a[1], 16) + lshift(a[2],8) + b[1];}')
 
-ifneq (,$(wildcard /etc/redhat-release))
-	RHEL_MAJOR := $(shell grep -Eoh -m 1 '[0-9]+\.[0-9]+' /etc/redhat-release | cut -d '.' -f 1)
-	RHEL_MINOR := $(shell grep -Eoh -m 1 '[0-9]+\.[0-9]+' /etc/redhat-release | cut -d '.' -f 2)
+DIST_ID := $(shell grep -m 1 'ID=' /usr/lib/os-release | cut -d '=' -f '2' | sed 's/\"//g')
+VERSION_ID := $(shell grep -m 1 'VERSION_ID=' /usr/lib/os-release | cut -d '=' -f '2' | sed 's/\"//g')
+CENTOS_MERGED_RHEL := $(shell echo $(DIST_ID) | sed 's/centos/rhel/')
+ifeq ($(CENTOS_MERGED_RHEL), rhel)
+	RHEL_MAJOR := $(shell echo $(VERSION_ID) | cut -d '.' -f 1)
+	RHEL_MAJOR := $(shell echo $(VERSION_ID) | cut -d '.' -f 2)
 	RHEL_RELEASE_CODE := $(shell echo $$(($(RHEL_MAJOR) << 8 | $(RHEL_MINOR))))
 else
 	RHEL_RELEASE_CODE := 0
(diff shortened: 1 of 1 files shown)


tracee-ebpf/Makefile
```diff
@@ -39,9 +39,12 @@ DOCKER_BUILDER_KERN_SRC ?= $(if $(shell readlink $(KERN_SRC_PATH)),$(shell readl
...
 
-ifneq (,$(wildcard /etc/redhat-release))
-	RHEL_MAJOR := $(shell grep -Eoh -m 1 '[0-9]+\.[0-9]+' /etc/redhat-release | cut -d '.' -f 1)
-	RHEL_MINOR := $(shell grep -Eoh -m 1 '[0-9]+\.[0-9]+' /etc/redhat-release | cut -d '.' -f 2)
+DIST_ID := $(shell grep -m 1 'ID=' /usr/lib/os-release | cut -d '=' -f '2' | sed 's/\"//g')
+VERSION_ID := $(shell grep -m 1 'VERSION_ID=' /usr/lib/os-release | cut -d '=' -f '2' | sed 's/\"//g')
+CENTOS_MERGED_RHEL := $(shell echo $(DIST_ID) | sed 's/centos/rhel/')
+ifeq ($(CENTOS_MERGED_RHEL), rhel)
+	RHEL_MAJOR := $(shell echo $(VERSION_ID) | cut -d '.' -f 1)
+	RHEL_MAJOR := $(shell echo $(VERSION_ID) | cut -d '.' -f 2)
```
rafaeldtinoco, line 47: you have major 2 times here, I suppose this is minor?

rafaeldtinoco:
There are some issues here I'm afraid:

* This change requires /usr/lib/os-release to exist and it does not exist by default in Fedora 33 and 34 cloud images, for example. That file is created by a package called **generic-release** while /etc/os-release is created by a package, automatically installed, called **fedora-release-common**. Checking all other distros, they all have a link from /etc/os-release to /usr/lib/os-release BUT fedora 33 and 34 (did not check 32, 31, 30). I think it is safer for us to use /etc/os-release instead of /usr/lib/os-release.

> The other issues are related to supported building environments outside docker. I'm able to compile tracee-ebpf in Ubuntu and Archlinux, for example, but not in Fedoras. I'll create an issue for us to enable all supported environments as building environments as well.

grantseltzer:
On Thu, Sep 9, 2021 at 3:34 PM Rafael David Tinoco ***@***.***>
wrote:

> There are some issues here I'm afraid:
>
>    - This change requires /usr/lib/os-release to exist and it does not
>    exist by default in Fedora 33 and 34 cloud images, for example. That file
>    is created by a package called *generic-release* while /etc/os-release
>    is created by a package, automatically installed, called
>    *fedora-release-common*. Checking all other distros, they all have a
>    link from /etc/os-release to /usr/lib/os-release BUT fedora 33 and 34 (did
>    not check 32, 31, 30). I think it is safer for us to use /etc/os-release
>    instead of /usr/lib/os-release.
>
> Ah ok, I figured it's safer to explicitly name the file that symlinks
point to but thanks for the insight. How do you typically check what
packages create that file?

rafaeldtinoco:
> Ah ok, I figured it's safer to explicitly name the file that symlinks point to but thanks for the insight. How do you typically check what packages create that file?

ubuntu/debian way:
 * apt-get install apt-file ; apt-file update ; apt-file search __
 
redhat/centos/alma/stream way:
* from dnf-utils / yum-utils : repoquery -f __ (includes non-installed packages, complete repo query tool)
* yum provides __ (includes non-installed packages)
* rpm -qf __ (for installed packages)

archlinux way:
* pacman -S pkgfile ; sudo pkgfile -s __

rafaeldtinoco (approved):
LGTM

PR #1002, opened 2021-09-09
tracee-ebpf: add sched_switch event

### LINKED ISSUE #983: Possibility to add new sched_* kernel function tracing
Hi, this issue is a feature request (if possible). I'm in need of tracing kernel functions like `sched_wakeup_new` and `sched_switch` to get the information of what process is running (also get to know the process if its clone/fork/exec event is not traced because the tracer may be running after the process is created), and the parent-child relationship ahead of time.

So I gently open this issue to ask if you can add support for these sched_* events. Thanks.

yanivagman:
close #983

### DIFF
--- a/tracee-ebpf/tracee/consts.go
+++ b/tracee-ebpf/tracee/consts.go
@@ -156,6 +156,7 @@ const (
 	SchedProcessForkEventID
 	SchedProcessExecEventID
 	SchedProcessExitEventID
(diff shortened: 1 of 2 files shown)


rafaeldtinoco (approved):
arguments of same type both sides, next argument is the task to be scheduled and 2nd argument of the trace point, everything else also seems sane. LGTM.

Issue #958, opened 2021-08-26
container_id initialization regex has to be fixed


rafaeldtinoco:
When running in an Ubuntu environment with LXD, the regex used to recognize running containers, and populate the map in charge for tracking existing containers + pids, will have false positives like:

```
./fs/cgroup/systemd/user.slice/user-1000.slice/user@1000.service/app.slice/snap.lxd.lxc.4b3296a2-c139-4da5-b782-006c431cca74.scope/tasks
```

```
$ sudo ./dist/tracee-ebpf --debug --trace event=mount
BTF: bpfenv = false, btfenv = false, vmlinux = true
BPF: using embedded BPF object
unpacked CO:RE bpf object file into memory
TIME             UID    COMM             PID     TID     RET              EVENT                ARGS
Running container = vice/app.slice/snap.lxd.lxc.4b3296a2-c139-4da5-b782-006c431cca74 pid = 2558259
```

We should not have "**UUID.scope/tasks**" regex as a 'generic' regex for now IMO.

Issue #884, opened 2021-08-06
Support other container engines


rafaeldtinoco:
Whenever tracing containers you might end up facing a new LXC/LXD containers spawned with broken container_id's:

```
TIME             UTS_NAME         CONTAINER_ID     UID    COMM             PID/host        TID/host        RET              EVENT                ARGS
01:09:10:417576  hirsute          blk-availabi     0      (activate)       2726   /113219  2726   /113219  0                execve               pathname: /sbin/blkdeactivate, argv: [/sbin/blkdeactivate -u -l wholevg -m disablequeueing -r wait]
01:09:10:420027  hirsute          apport.servi     0      (apport)         2727   /113220  2727   /113220  0                execve               pathname: /etc/init.d/apport, argv: [/etc/init.d/apport stop]
01:09:10:422200  hirsute          apport.servi     0      apport           2728   /113221  2728   /113221  0                execve               pathname: /usr/bin/run-parts, argv: [run-parts --lsbsysinit --list /lib/lsb/init-functions.d]
01:09:10:423332  hirsute          finalrd.serv     0      (finalrd)        2729   /113222  2729   /113222  0                execve               pathname: /usr/bin/finalrd, argv: [/usr/bin/finalrd]
01:09:10:424341  hirsute          apport.servi     0      apport           2730   /113223  2730   /113223  0                execve               pathname: /usr/bin/systemctl, argv: [systemctl -p LoadState --value show apport.service]
01:09:10:425165  hirsute          finalrd.serv     0      finalrd          2731   /113224  2731   /113224  0                execve               pathname: /usr/bin/mount, argv: [mount -o remount,exec /run]
01:09:10:429359  hirsute          blk-availabi     0      blkdeactivate    2732   /113225  2732   /113225  0                execve               pathname: /bin/umount, argv: [/bin/umount --help]
...
```

This is related to: https://github.com/aquasecurity/tracee/issues/862

This issue will try to summarise what container engines we should support:

### container engines
- docker
- containerd
- cri-o (oci)
- podman (oci)
- rkt
- lxd
- rkt (rocket, from CoreOS)

### container runtimes
- runc (oci)
- crun (oci)
- runv (oci)

Both, tracee.bpf.c AND tracee.go should better handle multiple container engines.

yanivagman:
@rafaeldtinoco after merging #1130 we now support:
- docker
- containerd
- crio
- podman

[...]

PR #1003, opened 2021-09-09
Fixed inconsistency in processes containerID value between startup and runtime

AlonZivony:
The change is incomplete, because the check in startup is if there is a 64
bytes of hexadecimal number in the cgroup dirname.
The reason for the incomplete change is that more accurate check will
be much heavier in performance.

### DIFF
--- a/tracee-ebpf/tracee/tracee.bpf.c
+++ b/tracee-ebpf/tracee/tracee.bpf.c
@@ -247,6 +247,7 @@ Copyright (C) Aqua Security inc.
 #define DEV_NULL_STR    0
 
 #define CONT_ID_LEN 12
+#define CONT_ID_MIN_FULL_LEN 64
 
 #ifndef CORE
 #if LINUX_VERSION_CODE < KERNEL_VERSION(5, 2, 0)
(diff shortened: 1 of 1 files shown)


tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2513,20 +2514,26 @@ SEC("raw_tracepoint/cgroup_attach_task")
...
     struct cgroup *dst_cgrp = (struct cgroup*)ctx->args[0];
     struct task_struct *task = (struct task_struct*)ctx->args[2];
     const char *cgrp_dirname = get_cgroup_dirname(dst_cgrp);
 
-    bpf_probe_read_str(&container_id.id, CONT_ID_LEN+1, cgrp_dirname);
+    // Only update container ID for names longer than 64-bytes.
+    // Container's cgroup dirname should contain 64-bytes of hexadecimal number, so a name smaller than 64-bytes
+    // can't be a real container's name (for example, "docker.service").
+    if (bpf_probe_read_str(&cgroup_dirname_buffer, CONT_ID_MIN_FULL_LEN, cgrp_dirname) == CONT_ID_MIN_FULL_LEN) {
+        bpf_probe_read_str(&container_id.id, CONT_ID_LEN+1, cgrp_dirname);
```
yanivagman, line 2527: Small optional optimization:

Instead of creating a buffer on the stuck, you can use a bigger id field for container_id_t:
```
typedef struct container_id {
    char id[CONT_ID_MIN_FULL_LEN+1];
} container_id_t;
```

Then reading one time instead of twice:
`int cgroup_len = bpf_probe_read_str(&container_id.id, CONT_ID_MIN_FULL_LEN, cgrp_dirname);`
and check if `cgroup_len == CONT_ID_MIN_FULL_LEN`

This should work fine, as we only take CONT_ID_LEN chars in `init_context()` and `sched_process_fork()` events

yanivagman: On a second thought, increasing container_id_t will also increase the container id map memory consumption.
We can keep it that way.

rafaeldtinoco:
Related (on the userland/initialization side): https://github.com/aquasecurity/tracee/issues/958 and https://github.com/aquasecurity/tracee/issues/884

yanivagman (approved):

PR #1010, opened 2021-09-15
tracee-ebpf: don't send argument type

yanivagman:
Instead of sending the type of every argument in an event,
we can use the argument index to figure out the type.
This saves sending an extra 1 byte per each argument of any given event.


grantseltzer (approved):
This looks good to me, it makes sense and works though I don't have any metrics to verify what the performance improvement is.

PR #1016, opened 2021-09-20
tracee-ebpf: simplify saving to buf

yanivagman:


tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1462,6 +1462,11 @@ static __always_inline int save_path_to_str_buf(buf_t *string_p, const struct pa
     unsigned int off;
     int sz;
 
+    // Get per-cpu string buffer
```
grantseltzer, line 1466: BPF programs can't be preempted and moved to a different CPU, right?

grantseltzer (commented):
Do I understand correctly that you're calculating the index of the string buffer when you originally create it, and then use it when accessing the buffer, instead of recalculating on each access?

tracee-ebpf/tracee/tracee.bpf.c
yanivagman: Right

grantseltzer (approved):
LGTM!

PR #1018, opened 2021-09-20
tracee-ebpf: remove events pipeline

yanivagman:
This PR removes the events pipeline that was created before tracee-rules existed.
The idea behind the pipeline was to create plugins between the different stages, and to improve performance, but neither is achieved by it today. As the amount of work done by each stage of the pipeline is very small, the channel communication time becomes more dominant, and the performance is degraded. With the pipeline, it takes ~75us (on my env) for an event to cross all of its stages (decode, process, emit). Without it, it now takes ~5us to perform the same logic.
Although it is more than x10 faster, this will not show a great improvement as the current bottleneck is the perf buffer polling as described in https://github.com/aquasecurity/libbpfgo/issues/80


tracee-ebpf/tracee/events.go
```diff
@@ -0,0 +1,184 @@
...
+			}
+
+			rawEvent.Args[argMeta.Name] = argVal
+			rawEvent.ArgMetas[i] = argMeta
+		}
+
+		if !t.shouldProcessEvent(rawEvent) {
+			continue
+		}
+		err = t.processEvent(&rawEvent.Ctx, rawEvent.Args)
```
rafaeldtinoco, line 80: If you keep the type creation (which I don't think is needed)... can't you pass &rawEvent only here ?

tracee-ebpf/tracee/events.go
```diff
@@ -0,0 +1,184 @@
...
+		if err != nil {
+			t.handleError(err)
+			continue
+		}
+
+		// Only emit events requested by the user
+		if !t.eventsToTrace[rawEvent.Ctx.EventID] {
+			continue
+		}
+		err = t.prepareArgs(&rawEvent.Ctx, rawEvent.Args)
```
rafaeldtinoco, line 90: same here

tracee-ebpf/tracee/events.go
```diff
@@ -0,0 +1,184 @@
...
+	for dataRaw := range t.eventsChannel {
+		dataBuff := bytes.NewBuffer(dataRaw)
+		var ctx context
+		err := binary.Read(dataBuff, binary.LittleEndian, &ctx)
+		if err != nil {
+			t.handleError(err)
+			continue
+		}
+
+		rawEvent := RawEvent{
```
rafaeldtinoco, line 54: Should we create a type for this ? It feels like we're adding a layer that is not necessarily needed as this type is only being used by shouldProcessEvent().

tracee-ebpf/tracee/events.go
```diff
@@ -0,0 +1,184 @@
...
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"strconv"
+	"unsafe"
+
+	"github.com/aquasecurity/tracee/tracee-ebpf/external"
+)
+
```
rafaeldtinoco, line 12: Because events_pipeline.go was deleted and events.go was created it was harder for me to know if some stuff already existed in the previous _pipeline.go file. For those changes, next time, maybe we should first move the file in one commit and change the code in the other - just thought about it actually, not something I ever worried - this way we have the diff 1:1 to compare existing code (just as a 2 cents note) and review.

tracee-ebpf/tracee/events.go
```diff
@@ -0,0 +1,184 @@
...
+			continue
+		}
+
+		rawEvent := RawEvent{
+			Ctx:      ctx,
+			Args:     make(map[string]interface{}, ctx.Argnum),
+			ArgMetas: make([]external.ArgMeta, ctx.Argnum),
+		}
+
+		params := EventsIDToParams[ctx.EventID]
```
rafaeldtinoco, line 60: so sometimes we're using ctx directly, sometimes using the rawevent type... like in line #87

rafaeldtinoco (commented):
Ill give another look tomorrow...

tracee-ebpf/tracee/events.go
yanivagman: What do you mean "keep the type creation"? Creating RawEvent? If so, then let me explain why it is created.
`RawEvent` (which shouldn't be with capital 'R' anymore - I'll fix it) is used to avoid iteration over the arguments.
The reason for that is that `RawEvent` contains an arguments map, which maps an argument name to its value. If we only created `external.Argument` slice (which is ordered by the argument number, unlike the arguments map), we had to iterate over it to get a specific argument value. So creating this `RawEvent` struct avoid this extra iteration, and simplifies the code (for example, in `prepareArgs()`).

> can't you pass &rawEvent only here

Yes, let me take care of that.

tracee-ebpf/tracee/events.go
yanivagman: See my previous comment

tracee-ebpf/tracee/events.go
yanivagman: Yes, I understand your point, and I actually wanted to split this into two commits at first for this exact reason.
The problem was that I also removed the `go func() {` lines which existed in the events_pipeline.go file, and which created an extra indentation. By removing it, it seemed like I changed almost every line of the file, and I think that it would have been much harder to review it like this. So I thought that creating a new file will make the review easier, as a reviewer can then open both files side by side and see the differences. Maybe I was wrong here, but my intention was to make the review easier.
Just for you to know - this patch mostly removed code, and there were only minimal changes in the logic.

tracee-ebpf/tracee/events.go
yanivagman: As I explained above, creating this type is necessary to avoid arguments iteration, which also simplifies the code in:
1. `shouldProcessEvent`
2. `processEvent`
3. `prepareArgs`

Have a look, for example, in `prepareArgs` where you can clearly see the advantage of using an arguments map

tracee-ebpf/tracee/events.go
yanivagman: I misunderstood you at first. You are right, we don't need it - removed.

tracee-ebpf/tracee/events.go
yanivagman: Ok, I've split the commit to few. The first commit has wrong indentation in purpose - it is now easier to review it one commit at a time. After review, we can merge all into one.

tracee-ebpf/tracee/events.go
yanivagman: We shouldn't, removed, thanks!

rafaeldtinoco (approved):
Thanks for the changes. LGTMN.

Issue #1023, opened 2021-09-23
Go modules are not published correctly


grantseltzer:
I realized when upgrading the external package in tracee-rules that there's something wrong with how we publish our packages.

We have a [github.com/aquasecurity/tracee](https://pkg.go.dev/github.com/aquasecurity/tracee) package, which we stopped having after v0.4.0 but still is shown in pkg.go.dev

We have a [github.com/aquasecurity/tracee/tracee-ebpf](https://pkg.go.dev/github.com/aquasecurity/tracee/tracee-ebpf) package, which doesn't have corresponding tags.

The [github.com/aquasecurity/tracee/tracee-ebpf/external](github.com/aquasecurity/tracee/tracee-ebpf/external) package does not have corresponding tags. As a result people can't do something like `go get github.com/aquasecurity/tracee/tracee-ebpf/external@v0.6.1`.

grantseltzer:
So i'm currently thinking that we should have one top level `github.com/aquasecurity/tracee` module. This would greatly simplify things, a downside being that we have to update usage of packages as we update the packages themselves.

I've tested a single module with my personal fork https://github.com/grantseltzer/tracee/tags with [this test consumer](https://github.com/grantseltzer/test-tracee-mod) and it appears to work how I envision.

When I did this, there's another small problem we run into where pkg.go.dev appears to take a couple hours to take effect:
 https://github.com/golang/go/issues/39007

itaysk:
closed by #1034

PR #1026, opened 2021-09-24
Remove multiple modules in favor of one top level one

grantseltzer:
The goal of this PR is to have a discussion about our use of Go modules in the tracee repo. I started investigating our module setup when I discovered #1023.

Based on what I've read so far modules are meant to be a collection of packages, which makes me think we should have a single tracee module, with various packages underneath, such as `tracee-ebpf/external`, `tracee-ebpf/tracee`, `tracee-rules/rego`, ... 

As a result all modules would be published with the same semantic version (same as we do now), the only difference is that dependencies inside the repository would always need to be synced. I'm in favor of this because why would we ever want tracee-rules to be using a different version of the external package from tracee-ebpf?

Particularly what i'm interested in hearing is why we decided to make `external` a separate module.

I've tested a single module with my personal fork https://github.com/grantseltzer/tracee/tags with [this test consumer](https://github.com/grantseltzer/test-tracee-mod) and it appears to work how I envision.

Need to verify that all versions of dependencies are the same before marking this as ready.

Should this get merged, follow up actions include cutting a new release and pushing to the go proxy.


simar7:
My 2c:

> The goal of this PR is to have a discussion about our use of Go modules in the tracee repo. I started investigating our module setup when I discovered #1023.
> 
> Based on what I've read so far modules are meant to be a collection of packages, which makes me think we should have a single tracee module, with various packages underneath, such as `tracee-ebpf/external`, `tracee-ebpf/tracee`, `tracee-rules/rego`, ...
> 
Packages are collections of modules. Modules are distinct versions that can be used independently as desired. In most cases, having one go.mod/go.sum per repo suffices. In our case we _chose_ to put all of our "packages" {package ϵ: tracee-rules, tracee-ebpf} into a single tracee repo. This complicates our situation a little by essential making our repo into a multi module repo. You can read more here: https://github.com/golang/go/wiki/Modules#faqs--multi-module-repositories

> As a result all modules would be published with the same semantic version (same as we do now), the only difference is that dependencies inside the repository would always need to be synced. I'm in favor of this because why would we ever want tracee-rules to be using a different version of the external package from tracee-ebpf?
> 
Semantic version is updated today for the Tracee repo as a whole. We don't seem to version individual modules themselves. This is something we can improve in order to better control our modules. See here how to control module version: https://golang.org/doc/modules/version-numbers#major-version 

It would be something like:
```
module github.com/aquasecurity/tracee/tracee-rules v1.2.3
```
If we went down this path, per module.

> Particularly what i'm interested in hearing is why we decided to make `external` a separate module.
> 

Today `external` is a package with a single module. My understanding of this is that without `external` being a standalone package, it would require the entire tracee package (lots of unnecessary stuff) for someone to use `external` as a library. Commit in reference: https://github.com/aquasecurity/tracee/pull/823

itaysk:
what @simar7 said. To add:
- a module is a unit of deployment and versioning. we don't necessarily want to deploy and version everything together
- external specifically is supposed to be consumable by external users, this is tracee-ebpf's API. the consumer shouldn't need to go get the entire tracee-ebpf module (with all of it's cgo complexity) just to import some type shared types dectalations.
- I like to think of it like "would this be a git repo" (as it usually go modules are, maybe why go pkg didn't pick it up). in our case we chose to group what should have been different repos in one for non-technical reasons and we might even break it apart some day.

grantseltzer:
Thanks for the info and context guys, this makes a lot more sense now. I do still want to tackle #1023 but this is not the way to do it.

Issue #1022, opened 2021-09-23
Supported Kubernetes Flavours


simar7:
This issue is keeping track of currently supported and unsupported Kubernetes flavours and setups. This list is a work in progress as we try out new Kubernetes flavours to test Tracee with.

**Currently Supported**
- minikube (v1.23.2+)

**Testing In Progress**
- k3s
- microk8s

**Needs Support**
- None so far, please feel free to add/update this if you find any.

Deployment YAMLs are available at: https://github.com/aquasecurity/tracee/tree/main/deploy/kubernetes

itaysk:
making note here that the interesting part of the test is for the OS the the Kubernetes platform ships with, not for k8s itself, since we want to make sure that ebpf works, and that BTF/kernel headers are available and that our ebpf probe successfully compiles and loads.

simar7:
Seems to be fine running on EKS (thanks to @krol3):
Version: v1.20.7-eks-135321
Container runtime: docker
```
$ tracee-ebpf --trace comm=strace --debug
KConfig: warning: could not check enabled kconfig features
(could not read /boot/config-5.4.141-67.229.amzn2.x86_64: stat /boot/config-5.4.141-67.229.amzn2.x86_64: no such file or directory)
OSInfo: KERNEL_RELEASE: 5.4.141-67.229.amzn2.x86_64
OSInfo: ID: alpine
OSInfo: VERSION_ID: 3.14.2
OSInfo: PRETTY_NAME: "Alpine Linux v3.14"
BTF: bpfenv = false, btfenv = false, vmlinux = true
BPF: using embedded BPF object
unpacked CO:RE bpf object file into memory
```

simar7:
On GKE:
Version: v1.20.9-gke.1001
Container runtime: docker 

```
# tracee-ebpf --debug
KConfig: warning: could not check enabled kconfig features
(could not read /boot/config-5.4.0-1049-gke: stat /boot/config-5.4.0-1049-gke: no such file or directory)
OSInfo: PRETTY_NAME: "Alpine Linux v3.14"
OSInfo: KERNEL_RELEASE: 5.4.0-1049-gke
OSInfo: ID: alpine
OSInfo: VERSION_ID: 3.14.2
BTF: bpfenv = false, btfenv = false, vmlinux = false
BPF: no BTF file was found or provided, building BPF object
found bpf object file at: /tmp/tracee/tracee.bpf.5_4_0-1049-gke.v0_6_1.o
```

[...]

PR #1027, opened 2021-09-24
doc: Add supported k8s info

simar7:
Addresses: https://github.com/aquasecurity/tracee/issues/1022

### DIFF
--- a/docs/install/kubernetes.md
+++ b/docs/install/kubernetes.md
@@ -12,3 +12,23 @@ This sample deploy use falcosidekick.tmpl and a config map for the falcosidekick
 
 > NOTE `See the complete config file in` [falcosidekick](https://github.com/falcosecurity/falcosidekick)
 
+# Supported Kubernetes Distributions
+
+Currently, the following Kubernetes distributions are supported out of the box:
+
+| Name | Version | Runtime
+| --- | --- | --- |
+GKE | v1.20.9-gke.1001 | docker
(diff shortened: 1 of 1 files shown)


yanivagman (commented):
I'm not sure I understand - what does it mean we support these k8s versions? Does it say we don't support any other k8s version?
Tracee shouldn't care about which k8s environment it runs on as long as the nodes have a supported kernel version.
If we had an interface with the k8s API (that we will probably add in the future for event enrichment) then we should have cared about k8s supported versions, but what us the meaning now?

itaysk:
Yeah I agree with Yaniv, I've sent another PR which is slightly less suggestive: https://github.com/aquasecurity/tracee/pull/1028

simar7:
> I'm not sure I understand - what does it mean we support these k8s versions? Does it say we don't support any other k8s version?
> Tracee shouldn't care about which k8s environment it runs on as long as the nodes have a supported kernel version.
> If we had an interface with the k8s API (that we will probably add in the future for event enrichment) then we should have cared about k8s supported versions, but what does it mean now?

Since the acceptance criteria wasn't clear for this story, I went ahead and tested a handful of k8s versions that I could find. If we're saying we support anything as long as the underlying nodes have the right setup, this PR can be scrapped. 

In other public projects, usually a "known set of working configurations" is provided, which was my goal with this PR.

PR #1030, opened 2021-09-28
tracee-ebpf: init context once

yanivagman:
Today we call `init_context()` twice per event: in `should_trace()` and in `init_and_save_context()`. In addition to this, we call other helpers to get the pid/tid/comm/etc, while this data was already fetched in `init_context()`. This is redundant, and adds some extra instructions which we can save.
This patch calls `init_context()` once, and uses the returned struct where needed.


rafaeldtinoco (approved):
Reviewed each change and they all look correct. Nice patch, much clearer now.

PR #1034, opened 2021-09-30
Individual module git tags

grantseltzer:
This makes it so that individual modules in the tracee repo
are properly tagged so that consumers can get versions of
the individual modules instead of only git commits.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@ BUILD_DATE = $(shell date -u +%Y-%m-%dT%H:%M:%SZ)
 VCS_REF="$(shell git rev-parse --short HEAD)"
 VCS_BRANCH="$(shell git rev-parse --abbrev-ref HEAD)"
 BUILD_ARG_DOCKER =--build-arg VERSION=$(VERSION) --build-arg BUILD_DATE=$(BUILD_DATE) --build-arg VCS_BRANCH=$(VCS_BRANCH) --build-arg VCS_REF=$(VCS_REF)
+MODULE_TAG_NAMES = tracee-ebpf/$(RELEASE_TAG) tracee-ebpf/external/$(RELEASE_TAG) tracee-rules/$(RELEASE_TAG)
 
 CMD_DOCKER ?= docker
 CMD_GIT ?= git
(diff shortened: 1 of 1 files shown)


grantseltzer:
> Just trying to piece it together: IIUC, this is the end result: https://github.com/grantseltzer/tracee/releases
> 

Correct

> The tags are essentially over the entire tracee repo but just include a subset of changes in their respective projects (e.g. tracee-rules, tracee-ebpf, etc?) Is my understanding correct?

The tags are exactly the same except for their names. If you navigate to the code in tracee-ebpf/v1.6.9 and tracee-rules/v1.6.9, it's the exact same git tree. The multiple tags are created so that `go get` recognizes the version for each individual module. 

> 
> If so how does the consumer pick them up? I see https://github.com/grantseltzer/test-tracee-mod/blob/main/go.mod#L5-L9 uses the versions of `v1.6.9` for tracee-ebpf for instance: https://github.com/grantseltzer/test-tracee-mod/blob/main/go.mod#L6 - Shouldn't that be `tracee-ebpf/v1.6.9`?

`go get github.com/aquasecurity/tracee-ebpf@v1.6.9` will result in https://github.com/grantseltzer/test-tracee-mod/blob/main/go.mod#L6

simar7:
> I'm sorry I messed that up, I meant to write `go get github.com/grantseltzer/tracee/tracee-ebpf@v1.6.9` (edited original post)

Oh yeah sorry about that I didn't notice that. Yeah it seems good to me, I could pull down that specific version. 

> The tags are exactly the same except for their names. If you navigate to the code in tracee-ebpf/v1.6.9 and tracee-rules/v1.6.9, it's the exact same git tree. The multiple tags are created so that go get recognizes the version for each individual module.

For my understanding, how does `go get` figure this out? I tried pulling down the repo and looking at the tags but I only saw `tracee-ebpf/v1.6.9` for instance. 
```
tracee-ebpf/external/v0.6.2
tracee-ebpf/external/v0.6.3
tracee-ebpf/external/v1.6.3
tracee-ebpf/external/v1.6.5
tracee-ebpf/external/v1.6.6
tracee-ebpf/external/v1.6.7
tracee-ebpf/external/v1.6.8
tracee-ebpf/external/v1.6.9
tracee-ebpf/v0.6.2
tracee-ebpf/v0.6.3
tracee-ebpf/v1.6.3
tracee-ebpf/v1.6.6
...
```

And for instance if we look at one of such commits:
```
commit b760f719e607f5d37c4f5505ed2c5b596e019ec2 (HEAD -> main, tag: tracee-rules/v1.6.9, tag: tracee-ebpf/v1.6.9, tag: tracee-ebpf/external/v1.6.9, origin/main, origin/HEAD)
Date:   Thu Sep 30 12:34:29 2021 -0400

    try future tag
    
```

Where are those multiple tags created? It's certainly a gap in my knowledge I'd like to supplement. But other than this lgtm.

PR #1036, opened 2021-10-03
tracee-ebpf: remove save_args_from_regs

yanivagman:
`save_args_from_regs()` is not used anymore for tracing syscalls. Remove it.


rafaeldtinoco (approved):
I assume is_x86_compat() isn't needed for the execution paths of trace_##name() calls (coming from the tail calls) thus the removal, right ? LGTM.

yanivagman:
> I assume is_x86_compat() isn't needed for the execution paths of trace_##name() calls (coming from the tail calls) thus the removal, right ? LGTM.

Right. This path is taken only when is_syscall is true, which never happens (today)

PR #1037, opened 2021-10-03
tracee-ebpf: update exec maps in sched_process_exec

### LINKED ISSUE #438: Use sched_process_exec and sched_process_fork tracepoints
Currently, we use the sys_enter and sys_exit tracepoints to update the traced_pids_map, new_pidns_map and new_pids_map maps during execve(at) and fork/clone/vfork syscalls.
This approach has two disadvantages:
1. It "wastes" instructions used for logic that is only relevant to these syscalls, which also wastes some cpu cycles in addition to that
2.  Failed execve/fork events will also update the maps with incorrect data - for example, if a porcess that is not traced is doing execve, we will add it in case pid=new was chosen, however, if the execve call failed, we will add it to the traced processes group althogh we shouldn't.

To resolve this, we can use two dedicated tracepoints (one for fork, one for execve) and move this logic there

yanivagman:
Today we check for newly created pids/containers in sys_enter and update internal maps when the system call is either execve/execveat.
We can avoid having this code in sys_enter (which is common to all syscalls) and move it to sched_process_exec instead.

Close #438


rafaeldtinoco (approved):
This looks correct, specially since we're updating traced_pids_map with followed pids only after should_trace() call. LGTM

Issue #1035, opened 2021-10-01
Suggestion - Add toleration to daemonset manifest


raesene:
In the Kubernetes [daemonset manifest](https://github.com/aquasecurity/tracee/blob/main/deploy/kubernetes/tracee.yaml) there's no tolerations for common taints. As a result the daemonset won't schedule pods to control plane nodes (or any other node with a taint on it).

Something like the below could be added to the daemonset to ensure that it tolerates all taints.

```
  tolerations:
  - operator: "Exists"
    effect: "NoExecute"
  - operator: "Exists"
    effect: "NoSchedule"
```

itaysk:
interesting point.. some questions:
1. what will happen in managed k8s where control plane nodes are inaccessible? would it fail or skip them?
2. how do we/user ensure there's room for Tracee on those nodes, which I presume they are usually tightly accounted?
3. what other scenarios may lead to these taints and might it be problematic?

the fact that you opened the issue tells you've considered that it's better to add it I suppose, so maybe just my lack of knowledge

 @danielpacak and @krol3 - any takes?

raesene:
Soo, 

 1. It shouldn't make any different, the control plane nodes are not running a kubelet in a managed cluster so it just won't deploy there.
 2. there's a number of daemonsets that will get deployed to control plane nodes in unmanaged clusters, logging, CNI etc. Ideally we should specify resource amounts for all manifests, but actually I'd often expect control plane nodes to have more available room than worker nodes, they don't have as much deployed to them and unless the operators are really being tight on resources they'll likely have space available.
 3. Taints are purely a mechanism to express a preference for things not being scheduled on a given set of nodes. They're not a security mechanism. Daemonsets typically will tolerate taints unless they're only targeting worker nodes.

Another option would be to provide one version with the toleration and another without. It's a tradeoff. The problem with not using a toleration is that Tracee loses sight of all activity on control plane nodes and realistically that's likely to be a key target for attackers, so it's quite a large blind spot to have.

danielpacak:
💯 Agree with @raesene We could also consider accepting all taints with:

```
tolerations:
  - operator: Exists
```

This is what usually kube-proxy or network plugin pods are doing.

And having resource requests and limits for incubating project is always good idea - we don't want to burn a data center because of a little typo in C code.

PR #1040, opened 2021-10-04
chore: k8s manifest - toleration and resources

krol3:
Related to the issue https://github.com/aquasecurity/tracee/issues/1035

### DIFF
--- a/deploy/kubernetes/tracee.yaml
+++ b/deploy/kubernetes/tracee.yaml
@@ -33,6 +33,18 @@ spec:
         - name: lib-modules
           mountPath: /lib/modules/
           readOnly: true
+        resources:
+          limits:
+            cpu: 500m
+            memory: 300Mi
+          requests:
+            cpu: 350m
+            memory: 50Mi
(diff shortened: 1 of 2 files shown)


deploy/kubernetes/tracee.yaml
```diff
@@ -33,6 +33,18 @@ spec:
...
+            cpu: 500m
+            memory: 300Mi
+          requests:
+            cpu: 350m
+            memory: 50Mi
+      tolerations:
+        - effect: NoSchedule
+          operator: Exists
+        - operator: "Exists"
+          effect: "NoSchedule"
```
danielpacak, line 47: Is there any difference between the two tolerations?? I guess you wanted to use NoExecute effect:

```suggestion
          effect: "NoExecute"
```

Also consider tolerating any taints with:

```yaml
tolerations:
  - operator: Exists
```

NIT: Use / don't use double quotes consistently 🤓

danielpacak (changes requested):

PR #1041, opened 2021-10-04
Add execution information flags to tracee-ebpf

AsafEitani:
Add process creation information to tracee-ebpf events.
The flag added is "-o option:exec-info" and calculates the file hash, file ctime and container ctime.
if the process was created on the host - it will get the mnt ctime.

Also changed the event for capturing to sched_process_exec instead of security_bprm_check.

There is a cache based on LRU for both container ctime and hash calculation.


itaysk:
Not a feedback for this PR, but this demonstrated the problem we have with our "custom" events which are undocumented and unintuitive.
No one would expect hashes to appear on `sched_process_exec`. We added it there for arbitrary/technical convenience, eventually we need to find a better way and when we do it'll be a breaking change.

PR #1042, opened 2021-10-04
fix: Ignore order of elements in engine_test.go

simar7:
This test will currently fail intermittently as seen here:
image

We need to ignore the order of elements received as its not important to assert. 


danielpacak (approved):
LGMT

PR #1052, opened 2021-10-07
Fix tracee-ebpf compilation on RHEL-likes

NeonSludge:
This PR fixes compilation of `tracee-ebpf` on RHEL-likes (AlmaLinux, Rocky Linux, etc.).
The detection is done by checking for the `rhel` substring in the `ID_LIKE` field of the `/etc/os-release` file instead of just using the `ID` field which contains the operating system identifier. This is necessary to avoid having to list all binary compatible RHEL derivatives in the Makefile.

For example in AlmaLinux these fields have the following values:
```
$ grep -m 1 'PRETTY_NAME=' /etc/os-release | cut -d '=' -f '2' | sed 's/\"//g'
AlmaLinux 8.4 (Electric Cheetah)

$ grep -m 1 'ID_LIKE=' /etc/os-release | cut -d '=' -f '2' | sed 's/\"//g'
rhel centos fedora

$ grep -m 1 'ID' /etc/os-release | cut -d '=' -f '2' | sed 's/\"//g'
almalinux
```


grantseltzer:
Apologies about the delay @NeonSludge, was out last week. I agree with your assessment, as I mentioned in my last comment it doesn't matter on Fedora. No need for me to bike shed on it. LGTM!

Issue #1059, opened 2021-10-09
Postee Deployment  Manifest needs Tolerations and Limits


raesene:
Looking at the new postee deployment manifest we need to make changes to it to mirror the changes in the falcosidekick manifest.

As we've gone for the approach of having a single manifest for each path, we'll need to make sure we mirror changes from one to the other where they're generic.

PR #1060, opened 2021-10-09
update postee manifest with tolerations and  resource limits

raesene:
Adds the resource limits and tolerations in to the Postee based manifest to bring it in line with the falco sidekick manifest. Fixes https://github.com/aquasecurity/tracee/issues/1059


danielpacak (approved):
LGMT

PR #1064, opened 2021-10-11
add socket_dup event

roikol:
create a new event to indicate a dup of socket.
in order to do that, we need to know that the file that is being 'dup'ed is of type socket.
we can get the 'struct file' of the duplicated FD, and we have to compare to file_operations of this file to the file_operations of a socket file. this is why i probe into the return of the 'socket' syscall. the return value of this syscall is the FD of the syscall in the current task. then i can go to the file descriptor table of the current task, and get the 'struct file' of this FD that i know to be a socket. i save the socket file_operations in a map, and then i use it in the 'dup*' events.


yanivagman (approved):

yanivagman:
LGTM.
Please resolve conflicts so we can merge.

PR #1081, opened 2021-10-14
feat: Switch base image to debian from alpine.

simar7:
This fixes the issue with not being able to run OPA/wasmas a rego runtime target.

The issue:
```
go: downloading github.com/bytecodealliance/wasmtime-go v0.30.0
go: downloading github.com/xeipuuv/gojsonpointer v0.0.0-20190905194746-02993c407bfb
# github.com/bytecodealliance/wasmtime-go
/usr/lib/gcc/x86_64-alpine-linux-musl/10.3.1/../../../../x86_64-alpine-linux-musl/bin/ld: /go/pkg/mod/github.com/bytecodealliance/wasmtime-go@v0.30.0/build/linux-x86_64/libwasmtime.a(std-008055cc7d873802.std.cf1c8f7e-cgu.0.rcgu.o): in function `std::sys::unix::net::on_resolver_failure':
/rustc/c8dfcfe046a7680554bf4eb612bad840e7631c4b//library/std/src/sys/unix/net.rs:450: undefined reference to `__res_init'
collect2: error: ld returned 1 exit status
make[1]: *** [Makefile:26: dist/tracee-rules] Error 2
make: *** [Makefile:43: dist/tracee-rules] Error 2
make[1]: Leaving directory '/tracee/tracee-rules'
The command '/bin/sh -c make' returned a non-zero code: 2
make: *** [Makefile:59: docker] Error 2
```

This is due to the issue: https://github.com/bytecodealliance/wasmtime-go/issues/21
More conversation: https://openpolicyagent.slack.com/archives/C01V5E6LQN7/p1634084728014500

The alternate idea to use alpine base image + prop. libc (https://github.com/sgerrand/alpine-pkg-glibc) does not work and runs into the same issue. Therefore I went ahead with this approach.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,9 @@
 ARG BASE=fat
 
-FROM golang:1.16-alpine as builder
-RUN apk --no-cache update && apk --no-cache add git clang llvm make gcc libc6-compat coreutils linux-headers musl-dev elfutils-dev libelf-static zlib-static
+FROM golang:1.17-buster as builder
+RUN echo "deb http://apt.llvm.org/buster/ llvm-toolchain-buster-12 main" >> /etc/apt/sources.list && apt-key adv --keyserver hkps://keyserver.ubuntu.com --recv-keys 15CF4D18AF4F7421 && \
+    DEBIAN_FRONTEND=noninteractive apt-get update -y && apt-get install -y --no-install-recommends gawk libelf-dev llvm-12 clang-12 && \
+    (for tool in "clang" "llc" "llvm-strip"; do path=$(which $tool-12) && ln -s $path ${path%-*}; done)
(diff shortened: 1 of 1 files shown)


yanivagman:
I understand the reasons you wrote for moving from alpine to debian.
Can you please have a look at https://github.com/aquasecurity/tracee/issues/335 and share your thoghts about it?

yanivagman:
Reducing size was not the only reason for this change.
Alpine images have a much smaller attack surface, which is also an important consideration.

I think this issue should be handled by the wasmtime project, and not by us introducing new dependencies to tracee (using glibc in that case).

itaysk:
If not by wasmtime than by us compiling and vendoring it (if we decide to go all in on wasm)

PR #1083, opened 2021-10-14
feat: Update tracee-ebpf base image to golang:1.17-alpine

simar7:


danielpacak:
To upgrade clang from 11 to 12 and fix compilation errors we replaced `golang:1.17-alpine` with `alpine:3.15` . See #1173 for more details.

PR #1084, opened 2021-10-14
feat: Update to go 1.17

simar7:
merge this prior to: https://github.com/aquasecurity/tracee/pull/1083 and https://github.com/aquasecurity/tracee/pull/1082

### DIFF
--- a/.github/actions/build-dependencies/action.yaml
+++ b/.github/actions/build-dependencies/action.yaml
@@ -5,7 +5,7 @@ inputs:
   go-version:
     description: Version of Go
     required: true
-    default: "1.16"
+    default: "1.17"
   opa-version:
     description: Version of OPA
     required: true
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -8,7 +8,7 @@ on:
     branches:
       - main
 env:
-  GO_VERSION: "1.16"
+  GO_VERSION: "1.17"
   OPA_VERSION: "v0.35.0"
 jobs:
   verify-code:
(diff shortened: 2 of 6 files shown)


simar7:
> Overall looks good, but the build is failing on GitHub runner and in my local env with the following error:
> 
> ```
> $ go build -o dist/tracee-rules
> go: updates to go.mod needed; to update it:
> 	go mod tidy
> ```

Hrm, yeah it's strange because locally it fine for me. Let me see what's missing here. But another way would also be to see what difference do you get if you run `go mod tidy` in your `go.mod` and `go.sum` files and push to this branch.

grantseltzer:
@simar7 Can you mark this as ready for review? As part of the next release I'd like to upgrade to 1.17

simar7:
> @simar7 Can you mark this as ready for review? As part of the next release I'd like to upgrade to 1.17

I've marked it ready for review, but please make sure all stakeholders are OK with this change. Last time that was not the case, as communicated offline.

yanivagman:
@mtcherni95

mtcherni95:
It will take some time in order for CNDR team to test if there is any issue with using an updated version of Go. 
Maybe, we could push it after 6.5 release?

grantseltzer:
@mtcherni95 Is there anything about [1.17](https://go.dev/doc/go1.17) you expect to be an issue? What would the testing consist of? Go 1.17 came out in August and I have been compiling/running tracee with it since then without any issue.

mtcherni95:
> maybe dump question, but do you really need to upgrade CNDR just because we now build Tracee with go 1.17?

that's the thing - I am not sure about it (I don't think so, but still wanted to test it first). It's in my backlog.

itaysk:
when do you think you'll test it?

yanivagman:
@simar7 can you please rebase so we can merge?

simar7:
> @simar7 can you please rebase so we can merge?

rebased

danielpacak (approved):

Issue #1075, opened 2021-10-13
Add ctime container creation to tracee-ebpf event


mtcherni95:

mtcherni95:
we would like tracee-ebpf to output also the ctime of the container creation once an event happens. this is because we would like to compare the time of container creation vs (for example) the time a file is created and executed inside the container.

mtcherni95:
we could be aware of container creation by maybe checking process id = 1 and execv to runc and save timestamp, however this would be done with a signature and at user level, not sure that it's what we want.

yanivagman:
Exactly. This is why we need to better define this task.
I understand the benefit of adding ctime, as well as other metadata of an executable.
But I don't think it should be related to "container creation", but maybe to process execution.
So, a good place for adding ctime might be sched_process_exec event, I think.
Then, tracee-rules can use this data however it wants, for example, when a "container created" signature is triggered.

PR #1089, opened 2021-10-17
adding ctime to sched-process-exec event

mtcherni95:


mtcherni95:
solves #1075

mtcherni95:
> We recently added ctime in #1041. This approach, however, is better as it is not exposed to race conditions. So, I suggest removing the ctime calculation made by "exec-info" flag, and change the name of this flag to exec-hash or something similar

alright, sounds good to me.

mtcherni95:
If no other comment/review, then we can promote IMO @itaysk @yanivagman

tracee-ebpf/tracee/events_processor.go
```diff
@@ -197,11 +197,6 @@ func (t *Tracee) processEvent(ctx *context, args map[string]interface{}, argMeta
 					*argMetas = append(*argMetas, hashMeta)
 					ctx.Argnum += 1
 					args["sha256"] = currentHash
-
-					ctimeMeta := external.ArgMeta{"ctime", "unsigned long"}
-					*argMetas = append(*argMetas, ctimeMeta)
-					ctx.Argnum += 1
-					args["ctime"] = sourceFileCtime
```
yanivagman, line 204: With this change in place, there is no more reason to have sourceFileCtime and even sourceFileStat is not required anymore, right? If so, can we just remove those and replace it by the ctime argument of the sched_process_exec event?

yanivagman (approved):

PR #1103, opened 2021-10-21
Feature/sort events by timestamp

AlonZivony:
Created a method to reorder events and passing them forward sorted by chronological order.
For more information about the issue, see #1113 .

## Algorithm
Events from the kernel are received one-by-one from the data channel.
We can rely on the fact that for each CPU, all the events received from it should be ordered by timestamps (except for syscalls which are not ordered because of the way we create the event in 2 steps).
So by adding to each event the source CPU, we can put it in its own CPU queue of events and get all the events to be almost ordered.

To avoid syscalls sorting problem, we can find the appropriate place to put the event in the queue from the end of it, and with maximum of 3 iterations we should find the matching place chronologically.

After we put all the events in queues according to the source CPU, we can extract each time the oldest first-in-queue event from all CPUs and send it forward.

To be able to promise that all events prior to oldest first-in-queue event arrived to other CPU's queues, we can use the fact that the CPUs' queues are ordered and be sure that if all queues sent events after given timestamp, all events prior to that timestamp have arrived. So, we can check for each CPU what is the most recent event it sent (for CPUs that sent new events). From those, we can check which one has the most ancient timestamp. Then, we can be sure that all events with older timestamps than that timestamp were received - and send them.

However, because of the other sorting problem cases (syscalls case and the vCPU case) we cannot send them right away, we need a time buffer. **This part is a bit complicated** - the vCPU case makes us wait at least 100ms to be sure that a vCPU didn't send a new event right after the previous (the other case is that it has no events to send). The syscalls case makes us wait about 3ms to be sure that there are no new events older than last event received in the CPU. Because the vCPU delayed event can be a syscall event (with older timestamp than the newest event in the CPU's queue), this make things event more complicated.
The solution to the 2 sorting problem cases is to wait at least 100ms until we send the events up to the decided timestamp. This way we can be sure that there won't be any new events received with older timestamp than the chosen one.

To summarize the algorithm - we have a CPU queue for each CPU. We insert new events to the matching CPU's queue, and follow which CPU was updated with new event. Each interval, we check from the most recent events from each CPU which has the oldest timestamp. After a delay of at least 100ms, we send all events from queues up to that timestamp in an ordered way. This way, we can be sure that all sent events are sorted.

### Queues
I implemented a queue structure myself for this PR, because I need an access to the internal queue to be able to insert new events not at the tail of the queue (in the case of syscalls events that are received in unsorted way).
### Pools
To reduce allocations and freeing amount, I introduced a Pool struct in this PR. The struct is used to make through it `alloc` and `free` of new event nodes which are used in the CPU queues. The pool save the freed nodes, and when `alloc` is required it return a saved free node if there is one saved, or alloc new one.
To avoid pooling large amount of nodes, the max pooled amount is the number of allocated nodes. If the number exceeds it, the Pool will free half of the amount pooled in it.

### DIFF
--- a/cmd/tracee-ebpf/internal/flags/flags-output.go
+++ b/cmd/tracee-ebpf/internal/flags/flags-output.go
@@ -20,7 +20,7 @@ Possible options:
 out-file:/path/to/file                             write the output to a specified file. create/trim the file if exists (default: stdout)
 err-file:/path/to/file                             write the errors to a specified file. create/trim the file if exists (default: stderr)
 none                                               ignore stream of events output, usually used with --capture
-option:{stack-addresses,detect-syscall,exec-env,relative-time,exec-hash,parse-arguments}
+option:{stack-addresses,detect-syscall,exec-env,relative-time,exec-hash,parse-arguments,sort-events}
                                                    augment output according to given options (default: none)
   stack-addresses                                  include stack memory addresses for each event
   detect-syscall                                   when tracing kernel functions which are not syscalls, detect and show the original syscall that called that function
@@ -28,6 +28,7 @@ option:{stack-addresses,detect-syscall,exec-env,relative-time,exec-hash,parse-ar
   relative-time                                    use relative timestamp instead of wall timestamp for events
   exec-hash                                        when tracing sched_process_exec, show the file hash(sha256) and ctime
   parse-arguments                                  do not show raw machine-readable values for event arguments, instead parse into human readable strings
+  sort-events                                      enable sorting events before passing to them output. This will decrease the overall program efficiency.
 Examples:
   --output json                                            | output as json
   --output gotemplate=/path/to/my.tmpl                     | output as the provided go template
@@ -89,6 +90,8 @@ func PrepareOutput(outputSlice []string) (tracee.OutputConfig, printerConfig, er
 				outcfg.ExecHash = true
 			case "parse-arguments":
 				outcfg.ParseArguments = true
+			case "sort-events":
+				outcfg.EventsSorting = true
 			default:
 				return outcfg, printcfg, fmt.Errorf("invalid output option: %s, use '--output help' for more info", outputParts[1])
 			}
--- a/pkg/events/sorting/pool.go
+++ b/pkg/events/sorting/pool.go
@@ -1,0 +1,100 @@
+package sorting
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/aquasecurity/tracee/pkg/external"
+)
+
+type eventNode struct {
+	event       *external.Event
+	previous    *eventNode
+	next        *eventNode
+	isAllocated bool
+}
+
+const poolFreeingPart = 2
+const minPoolFreeingSize = 200
+
+// eventsPool is a struct used to store empty eventNodes to prevent repeated allocation and freeing from the heap.
+// In details, the struct envelopes the allocation and freeing of the eventNode object, and store freed nodes
+// to to use them again when new allocation is required.
+// Pool is a known mechanism, so you can read more on it online.
+type eventsPool struct {
+	head             *eventNode
+	poolMutex        sync.Mutex
+	allocationsCount int
+	poolSize         int
+}
+
+// Alloc return an eventNode that contains the event given.
+// The function will try to use stored eventNode if there is one available.
+// If there isn't, it will allocate new one.
+func (p *eventsPool) Alloc(event *external.Event) *eventNode {
+	p.poolMutex.Lock()
+	p.allocationsCount += 1
+	defer p.poolMutex.Unlock()
+	node := p.head
+	if node != nil && node.isAllocated == false {
+		p.head = node.previous
+		node.event = event
+		node.isAllocated = true
+		node.previous = nil
+		p.poolSize -= 1
+		return node
+	}
+	return &eventNode{event: event, isAllocated: true}
+}
+
+// Free handle the eventNode after its usage has ended.
+// Free will try to store the node in the pool to be used by a future Alloc call.
+// To prevent the storage of all nodes event when there is no use for them (like after peaks of allocations), whenever
+// the amount of stored nodes pass the amount of nodes in use, a part of the pool will be removed from the pool (and
+// as a result will be garbage collected).
+func (p *eventsPool) Free(node *eventNode) error {
+	p.poolMutex.Lock()
+	defer p.poolMutex.Unlock()
+	// Prevent malicious use of free
+	if p.allocationsCount == 0 {
+		return fmt.Errorf("BUG: free called when no allocated node exist")
+	}
+
+	node.isAllocated = false
+	node.previous = nil
+	node.next = nil
+	p.allocationsCount -= 1
+	// Free memory in case of pooling too many nodes
+	if p.poolSize >= p.allocationsCount &&
+		p.poolSize >= minPoolFreeingSize {
+		freeingAmount := p.poolSize / poolFreeingPart
+		for i := 0; i < freeingAmount; i++ {
+			p.head = p.head.previous
+		}
+		p.poolSize -= freeingAmount
+	} else { // Add unused node to pool
+		if p.head == nil {
+			p.head = node
+		} else {
+			node.previous = p.head
+			p.head.next = node
+			p.head = node
+		}
+		p.poolSize += 1
+	}
+	return nil
+}
+
+func (p *eventsPool) Reset() {
+	p.poolMutex.Lock()
+	defer p.poolMutex.Unlock()
+	node := p.head
+	for node != nil {
+		next := node.previous
+		node.next = nil
+		node.previous = nil
+		node = next
+	}
+	p.allocationsCount = 0
+	p.poolSize = 0
+}
--- a/pkg/events/sorting/pool.go
+++ b/pkg/events/sorting/pool.go
@@ -7,13 +7,6 @@ import (
 	"github.com/aquasecurity/tracee/pkg/external"
 )
 
-type eventNode struct {
-	event       *external.Event
-	previous    *eventNode
-	next        *eventNode
-	isAllocated bool
-}
-
 const poolFreeingPart = 2
 const minPoolFreeingSize = 200
 
--- a/pkg/events/sorting/queue.go
+++ b/pkg/events/sorting/queue.go
@@ -1,0 +1,108 @@
+package sorting
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/aquasecurity/tracee/pkg/external"
+)
+
+type eventNode struct {
+	event       *external.Event
+	previous    *eventNode
+	next        *eventNode
+	isAllocated bool
+}
+
+// A double linked list used to store events in LIFO order
+type eventsQueue struct {
+	pool  eventsPool
+	tail  *eventNode
+	head  *eventNode
+	mutex sync.Mutex
+}
+
+// Put insert new event to the double linked list
+func (eq *eventsQueue) Put(newEvent *external.Event) {
+	newNode := eq.pool.Alloc(newEvent)
+	eq.mutex.Lock()
+	defer eq.mutex.Unlock()
+	eq.put(newNode)
+}
+
+// Get remove the node at the head of the queue and return it
+// Might return error with a valid event in case of internal pool error, for the user to know that an error occurred
+// but was contained
+func (eq *eventsQueue) Get() (*external.Event, error) {
+	eq.mutex.Lock()
+	defer eq.mutex.Unlock()
+	if eq.head == nil {
+		if eq.tail != nil {
+			return nil, fmt.Errorf("BUG: TAIL without a HEAD")
+		}
+		return nil, nil
+	}
+	headNode := eq.head
+	if headNode == eq.tail {
+		if headNode.next != nil || headNode.previous != nil {
+			return nil, fmt.Errorf("BUG: last existing node still conneced")
+		}
+		eq.tail = nil
+		eq.head = nil
+	} else {
+		if headNode.previous == nil {
+			return nil, fmt.Errorf("BUG: not TAIL lacking previous")
+		}
+		if headNode.next != nil {
+			return nil, fmt.Errorf("BUG: HEAD has next")
+		}
+		headNode.previous.next = nil
+		eq.head = headNode.previous
+	}
+	headNode.previous = nil
+	headNode.next = nil
+	extractedEvent := headNode.event
+	err := eq.pool.Free(headNode)
+	if err != nil {
+		eq.pool.Reset()
+		return extractedEvent, fmt.Errorf("error in queue's node freeing - %v", err)
+	}
+	return extractedEvent, nil
+}
+
+func (eq *eventsQueue) PeekHead() *external.Event {
+	eq.mutex.Lock()
+	defer eq.mutex.Unlock()
+	if eq.head == nil {
+		return nil
+	}
+	return eq.head.event
+}
+
+func (eq *eventsQueue) PeekTail() *external.Event {
+	eq.mutex.Lock()
+	defer eq.mutex.Unlock()
+	if eq.tail == nil {
+		return nil
+	}
+	return eq.tail.event
+}
+
+func (eq *eventsQueue) Empty() {
+	eq.mutex.Lock()
+	defer eq.mutex.Unlock()
+	eq.head = nil
+	eq.tail = nil
+}
+
+// Put insert new event to the double linked list
+func (eq *eventsQueue) put(newNode *eventNode) {
+	if eq.tail != nil {
+		newNode.next = eq.tail
+		eq.tail.previous = newNode
+	}
+	if eq.head == nil {
+		eq.head = newNode
+	}
+	eq.tail = newNode
+}
--- a/pkg/utils/environment/cpu_analyzing.go
+++ b/pkg/utils/environment/cpu_analyzing.go
@@ -1,0 +1,59 @@
+package environment
+
+import (
+	"fmt"
+	"os"
+	"strings"
+)
+
+const possibleCPUsFilePath = "/sys/devices/system/cpu/possible"
+
+func GetCPUAmount() (int, error) {
+	possibleCPUsFileContent, err := os.ReadFile(possibleCPUsFilePath)
+	if err == nil {
+		cpusAmount, err := parsePossibleCPUAmountFromCPUFileFormat(string(possibleCPUsFileContent))
+		if err == nil {
+			return cpusAmount, nil
+		}
+	}
+	return 0, err
+}
+
+const singleValue = 1
+const rangeValues = 2
+const rangeValuesWithGroups = 4
+
+// parsePossibleCPUAmountFromCPUFileFormat parse the format of /sys/devices/system/cpu files and return the amount of CPUs
+// specified. The format expected is according to the bitmap_parselist function in the linux kernel.
+// Notice that possible CPUs should be sequential values starting with 0, because the OS gives indexes from 0 onward to
+// all possible CPUs.
+//
+// Note: 'sysconf(_SC_NPROCESSORS_CONF)' is not used because it is found to be broken.
+// For more info see thread from libbpf - https://lore.kernel.org/bpf/ef0f23d0-456a-70b0-1ef9-2615a5528278@iogearbox.net/
+func parsePossibleCPUAmountFromCPUFileFormat(cpuFileContent string) (int, error) {
+	var rangeStart, rangeEnd int
+	var usedSize, groupSize int
+	bitmapsRegions := strings.Split(cpuFileContent, ",")
+	if len(bitmapsRegions) > 1 {
+		return 0, fmt.Errorf("possible cpus should be following indexes starting with 0, so multiple regions is not allowed")
+	}
+	cpusAmount := 0
+	n, _ := fmt.Sscanf(bitmapsRegions[0], "%d-%d:%d/%d", &rangeStart, &rangeEnd, &usedSize, &groupSize)
+	switch n {
+	case singleValue:
+		if rangeStart != 0 {
+			return 0, fmt.Errorf("possible cpus must start from the index 0, so single CPU index value must be 0")
+		}
+		cpusAmount = 1
+	case rangeValues:
+		if rangeStart != 0 {
+			return 0, fmt.Errorf("possible cpus should be following indexes range starting with 0")
+		}
+		cpusAmount = rangeEnd - rangeStart + 1
+	case rangeValuesWithGroups:
+		return 0, fmt.Errorf("possible cpus should be following indexes, but received groups format")
+	default:
+		return 0, fmt.Errorf("unkown possible cpu file format")
+	}
+	return cpusAmount, nil
+}
--- a/cmd/tracee-ebpf/main_test.go
+++ b/cmd/tracee-ebpf/main_test.go
@@ -1275,14 +1275,24 @@ func TestPrepareOutput(t *testing.T) {
 			expectedError: nil,
 		},
 		{
+			testName:    "option sort-events",
+			outputSlice: []string{"option:sort-events"},
+			expectedOutput: tracee.OutputConfig{
+				ParseArguments: true,
+				EventsSorting:  true,
+			},
+			expectedError: nil,
+		},
+		{
 			testName:    "all options",
-			outputSlice: []string{"option:stack-addresses", "option:detect-syscall", "option:exec-env", "option:exec-hash"},
+			outputSlice: []string{"option:stack-addresses", "option:detect-syscall", "option:exec-env", "option:exec-hash", "option:sort-events"},
 			expectedOutput: tracee.OutputConfig{
 				StackAddresses: true,
 				DetectSyscall:  true,
 				ExecEnv:        true,
 				ExecHash:       true,
 				ParseArguments: true,
+				EventsSorting:  true,
 			},
 			expectedError: nil,
 		},
--- a/pkg/events/sorting/cpu_queue.go
+++ b/pkg/events/sorting/cpu_queue.go
@@ -1,0 +1,56 @@
+package sorting
+
+import (
+	"fmt"
+
+	"github.com/aquasecurity/tracee/pkg/external"
+)
+
+// Events queue with the ability to follow if it was updated since last check and insertion by time specific for CPU
+// queue ordering
+type cpuEventsQueue struct {
+	eventsQueue
+	IsUpdated bool
+}
+
+// InsertByTimestamp insert new event to the queue in the right position according to its timestamp
+func (cq *cpuEventsQueue) InsertByTimestamp(newEvent *external.Event) error {
+	newNode := cq.pool.Alloc(newEvent)
+
+	cq.mutex.Lock()
+	defer cq.mutex.Unlock()
+	if cq.tail != nil &&
+		cq.tail.event.Timestamp > newEvent.Timestamp {
+		// We have a fresh event with a timestamp older than the last event received in this cpu's queue.
+		// This can only happen if this fresh event is a syscall event (for which we take the entry timestamp) which
+		// called some internal kernel functions (that are also traced). Insert the syscall event before these other
+		// events
+		insertLocation := cq.tail
+		for insertLocation.next != nil {
+			if insertLocation.next.event.Timestamp < newEvent.Timestamp {
+				break
+			}
+			if insertLocation.next == insertLocation {
+				return fmt.Errorf("encountered node with self reference at next")
+			}
+			insertLocation = insertLocation.next
+		}
+		cq.insertAfter(newNode, insertLocation)
+	} else {
+		cq.put(newNode)
+	}
+	return nil
+}
+
+// insertAfter insert new event to the queue after another node
+// This is useful if new node place is not at the end of the queue but before it
+func (cq *cpuEventsQueue) insertAfter(newNode *eventNode, baseEvent *eventNode) {
+	if baseEvent.next != nil {
+		baseEvent.next.previous = newNode
+	}
+	newNode.previous = baseEvent
+	newNode.next, baseEvent.next = baseEvent.next, newNode
+	if cq.head == baseEvent {
+		cq.head = newNode
+	}
+}
--- a/pkg/events/sorting/sorting.go
+++ b/pkg/events/sorting/sorting.go
@@ -1,0 +1,256 @@
+// Package sorting is responsible for sorting incoming events from the BPF programs chronologically.
+//
+// There are 3 known sources to events sorting issues:
+//   1. In perf buffer, events are read in round robing order from CPUs buffers (and not according to invocation time).
+//   2. Syscall events are invoked after internal events of the syscall (though the syscall happened before the
+//      internal events).
+//   3. Virtual CPUs might enter sleep mode by host machine scheduler and send events after some delay.
+//
+// To address the events perf buffers issue, the events are divided to queues according to the source CPU. This way
+// the events are almost ordered (except for syscalls). The syscall events are inserted to their right chronological
+// place manually.
+// This way, all events which occurred before the last event of the most delaying CPU could be sent forward with
+// guaranteed order.
+// To make sure syscall events are not missed when sending, a small delay is needed.
+// Lastly, to address the vCPU sleep issue (which might cause up to 2 events received in a delay), the events need to be
+// sent after a delay which is bigger than max possible vCPU sleep time (which is just an increase of the syscall events
+// delay sending).
+//
+// To summarize the algorithm main logic, here is textual simulation of the operation (assume that 2 scheduler ticks
+// are larger than max possible vCPU sleep time):
+//   -------------------------------------------------------------------
+// Tn = Timestamp (n == TOD)
+// #m = Event's Source CPU
+//
+// ### Initial State
+//
+//        [ CPU 0 ]    [ CPU 1 ]    [ CPU 2 ]
+//   HEAD    T1           T2           T4
+//           T3           T5
+//           T6
+//   TAIL    T8
+//
+// ### Scheduler Tick #1
+//
+// Incoming events: T9#1, T11#2, T13#1, T10#2, T12#2
+//
+// Queues state after insert:
+//        [ CPU 0 ]    [ CPU 1 ]    [ CPU 2 ]
+//   HEAD    T1           T2           T4
+//           T3           T5           T10 +
+//           T6           T9  +        T11 +
+//   TAIL    T8           T13 +        T12 +
+//
+//   - No event sent.
+//   - Oldest timestamp = T1.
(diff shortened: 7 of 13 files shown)


itaysk:
Can you please open an issue explaining what this is and what is the problem you were trying to solve

AlonZivony:
The current solution does not seem to solve the problem of disordered events from all sources, just from Perf Buffer CPUs union disordering. It seems that there is another reason for the disordering right now from tests I made, so this PR should stay as a draft until I will give global solution.

mtcherni95:
> The current solution does not seem to solve the problem of disordered events from all sources, just from Perf Buffer CPUs union disordering. It seems that there is another reason for the disordering right now from tests I made, so this PR should stay as a draft until I will give global solution.

Hi Alon, great work so far! maybe the 'disordering' between events is related to to multi core cpu? if so then that would be logical as the libbpf implements a buffer per core. If that's true then we shouldn't expect disordering in single core cpu. What do you think?

AlonZivony:
> > The current solution does not seem to solve the problem of disordered events from all sources, just from Perf Buffer CPUs union disordering. It seems that there is another reason for the disordering right now from tests I made, so this PR should stay as a draft until I will give global solution.
> 
> Hi Alon, great work so far! maybe the 'disordering' between events is related to multi core cpu? if so then that would be logical as the libbpf implements a buffer per core. If that's true then we shouldn't expect disordering in single core cpu. What do you think?

Hi Michael!
To you question, I just discovered with @yanivagman that the reason for the disordering is the result of the way we create events from syscalls. It occurs with single core CPU as well, so using one will not solve all the disordering issues, but will solve the more severe ones. You can take a look at the connected issue I opened to see all the disordering reasons I found.
Anyway, this PR will open soon for CR and merge, so I hope you will have the solution for your problem soon enough :)

AlonZivony:
After testing the performance of tracee-ebpf using my new feature, this is the results:
The test were taken in an isolated virtual machine, where only tracee-ebpf and the performance tool were used.
The results are the user-mode CPU usage in %, because the feature is user-mode feature.
The results are:

- Without the feature (main branch) - 20.0%
- With perf-buffer solution ordering - 23.2%
- With sort-guaranteed solution (for all 3 ordering issues) - 25.4%

The result is that the complete feature increase CPU usage by 27%, and the perf-buffer specific solution increase by 16%.
This is huge performance issue.
What do you think we should do?

tracee-ebpf/main.go
```diff
@@ -304,6 +305,11 @@ func main() {
 				Usage:       "when to build the bpf program. possible options: 'never'/'always'/'if-needed'",
 				Destination: &buildPolicy,
 			},
+			&cli.BoolFlag{
+				Name:  "sort-events",
+				Value: false,
```
mtcherni95, line 310: why false by default?

tracee-ebpf/tracee/events.go
```diff
@@ -15,27 +15,51 @@ import (
...
+	_              [3]byte //padding
 }
 
-func (t *Tracee) processEvents(done <-chan struct{}) error {
+// handleEvents is a high-level function that starts all operations related to events processing
+func (t *Tracee) handleEvents(done <-chan struct{}) {
+	var closeChannels []chan struct{}
+	processEventsOutputChan := t.config.ChanEvents
+	if t.config.Sorting {
+		sorterInputChan := make(chan external.Event, 100)
```
mtcherni95, line 44: is the channel size arbitrary? how do we know is it a good amount? maybe 100 is not enough. Just thinking out loud. What do you think?

tracee-ebpf/tracee/events_sorter_test.go
```diff
@@ -0,0 +1,176 @@
...
+			name: "Sorting not chronological order events complicated",
+		},
+	}
+	for _, testCase := range testCases {
+		inputChan := make(chan external.Event)
+		outputChan := make(chan external.Event)
+		errChan := make(chan error)
+		newSorter := InitEventSorter(inputChan, outputChan, errChan)
+		doneChannel := make(chan struct{})
+		go newSorter.Start(doneChannel)
```
mtcherni95, line 159: Looks like Start function isn't returning any type of error code. Don't we want to have it?

tracee-ebpf/tracee/events.go
AlonZivony: You are right,
I am not sure what size is good for the channel.
Do you have any advice on how to determine the channel size?
In the end I hope the writing and reading speeds will be similar, and there shouldn't be any large peaks of differences except maybe when sending the events forward from the sorter (because part of the time it is blocking operation because all pointers and lists should be updated).

tracee-ebpf/main.go
AlonZivony: So I thought it might be good at start to make it optional.
However, I think you are right and most users probably will want it.
I will change it.

itaysk: do we need this for the current open sourced signatures?

AlonZivony: We partly need it.
The "stdio_over_socket" signature might be unpredictable (will probably miss memory freeing part of the times because of unordered events).

tracee-ebpf/tracee/sorting/events_sorter.go
```diff
@@ -0,0 +1,332 @@
...
+
+// parseCPUAmountFromCPUFileFormat parse the format of /sys/devices/system/cpu files and return the amount of CPUs
+// specified. The format expected is according to the bitmap_parselist function in the linux kernel.
+func parseCPUAmountFromCPUFileFormat(cpuFileContent string) (int, error) {
+	bitmapsRegions := strings.Split(cpuFileContent, ",")
+	cpusAmount := 0
+	for _, region := range bitmapsRegions {
+		var rangeStart, rangeEnd int
+		var usedSize, groupSize int
+		n, err := fmt.Sscanf(region, "%d-%d:%d/%d", &rangeStart, &rangeEnd, &usedSize, &groupSize)
```
yanivagman, line 319: Please verify that this is indeed the format for possible cpu's file

yanivagman:
Also, let's not make this feature the default for now as it brings some overhead, and needs to be experimented first

tracee-ebpf/tracee/sorting/events_sorter_test.go
```diff
@@ -0,0 +1,269 @@
...
+			errChan := make(chan error)
+			newSorter, err := InitEventSorter(errChan)
+			require.NoError(t, err)
+			newSorter.cpuEventsQueues = make([]cpuEventsQueue, testCase.cpuAmount)
+			doneChannel := make(chan struct{})
+			go newSorter.Start(inputChan, outputChan, doneChannel, fatalErrorsChan)
+			go sendTestEvents(inputChan, testCase.eventsPools)
+			outputList, sorterErr := retrieveEventsFromSorter(testCase.eventsAmount, outputChan, fatalErrorsChan)
+			require.NoError(t, sorterErr)
+			assert.Empty(t, errChan)
```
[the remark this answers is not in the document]
AlonZivony: The check of the sorterErr does this logic (the retrieve function waits for the amount of received events)

yanivagman:
Please rebase on recent main which re-introduced the events pipeline

yanivagman (commented):
This is a big change, and I already reviewed most of it (will take another look today).
@itaysk / @grantseltzer  / @rafaeldtinoco  / @danielpacak can one of you guys also have a look?

rafaeldtinoco:
I'll review this today.

A initial feedback, already given before, is this: IMHO, we should propose PRs with commits containing a feature or a fix (like any other well maintained upstream project that I am aware of).

So, I'd rather have all this squashed (it could be at the end of reviews, ok for me) and the git log to be a complete description of everything that is changed (instead of things such as "Doc fix", "Changed files structure") OR this could be rebased into multiple commits, each of them being explicit about changes introduced and reasoning (@AlonZivony you can keep your git history in another local branch, if you'd like, like you said previously).

It's a good practice also to inform what part of the tree you're touching in your commit log ("tracee-ebpf: xxxx", "containers: yyyyy"). Not for the reviewer, but for future reference.

A big set of commits with small descriptions is not good if someone would backport this feature to an older version, for example (if we had a LTS tree).

Since we're not merging branches in our repository, it is hard to understand which commit is responsible for the feature introduction and which one is the end of it (just one example on why we should work like other upstream projects).

Let's say someone approves this PR, it would be "ready for merge" with a big set of commits that don't mean anything to the entire project.

Now, I'll get to the technical part (which looks very good from a first fast look).

yanivagman:
Agree with @rafaeldtinoco - the commits of a PR should be ready for review/merge, definitely not 30 commits like in this one

pkg/events/sorting/pool.go
```diff
@@ -0,0 +1,78 @@
...
+		return fmt.Errorf("a free operation on node from pool is requested when there are supposingly no allocated nodes")
+	}
+
+	node.isAllocated = false
+	node.previous = nil
+	node.next = nil
+	p.allocationsCount -= 1
+	// Free memory in case of pooling too many nodes
+	if p.poolSize >= p.allocationsCount {
+		freeingAmount := p.poolSize / poolFreeingPart
```
rafaeldtinoco, line 49: Should we set a minimum here ? I'm afraid of being stuck in lower values depending on alloc/dealloc sequence behavior. Something like if `p.poolSize > 200` && >= `p.allocationsCount`. WDYT ?

pkg/events/sorting/queue.go
```diff
@@ -0,0 +1,107 @@
...
+
+// Get remove the node at the head of the queue and return it
+// Might return error with a valid event in case of internal pool error, for the user to know that an error occured
+// but was contained
+func (eventsQueue *eventsQueue) Get() (*external.Event, error) {
+	eventsQueue.mutex.Lock()
+	defer eventsQueue.mutex.Unlock()
+	if eventsQueue.head == nil {
+		if eventsQueue.tail != nil {
+			return nil, fmt.Errorf("unmatching values of head and tail - one should be nil only if queue is empty")
```
rafaeldtinoco, line 40: Suggestion for less wide error message:

"BUG: TAIL without a HEAD".

pkg/events/sorting/queue.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+		return nil, nil
+	}
+	headNode := eventsQueue.head
+	if headNode == eventsQueue.tail {
+		if headNode.next != nil || headNode.previous != nil {
+			return nil, fmt.Errorf("trying to remove the only node in queue, but it is connected to another node")
```
rafaeldtinoco, line 47: "BUG: last existing node still connected".

pkg/events/sorting/queue.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+		eventsQueue.tail = nil
+		eventsQueue.head = nil
+	} else {
+		if headNode.previous == nil {
+			return nil, fmt.Errorf("a node which is not the tail without previous is given to remove")
```
rafaeldtinoco, line 53: "BUG: intermediate node missing previous".

pkg/events/sorting/queue.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+		if headNode.next != nil {
+			return nil, fmt.Errorf("head event to remove has next")
```
rafaeldtinoco, line 56: "BUG: HEAD points to next event".

pkg/events/sorting/pool.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	return &eventNode{event: event, isAllocated: true}
+}
+
+func (p *eventsPool) Free(node *eventNode) error {
+	p.poolMutex.Lock()
+	defer p.poolMutex.Unlock()
+	// Prevent malicious use of free
+	if p.allocationsCount == 0 {
+		return fmt.Errorf("a free operation on node from pool is requested when there are supposingly no allocated nodes")
```
rafaeldtinoco, line 40: "BUG: free called when no allocated nodes exist".

pkg/events/sorting/queue.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+		headNode.previous.next = nil
+		eventsQueue.head = headNode.previous
+	}
+	headNode.previous = nil
+	headNode.next = nil
+	extractedEvent := headNode.event
+	err := eventsQueue.pool.Free(headNode)
```
rafaeldtinoco, line 64: The only possible error here is a call to Free() when there are no allocated eventNode's. If that is the case, why not to simply deal with that silently, instead of propagating this error to upper calls ?

rafaeldtinoco:
I have reviewed `queue.go` and `pool.go` and parts of the rest of the changes. I'll continue reviewing the logic tomorrow (`cpu_queue.go` and `sorting.go`).

AlonZivony:
> I'll review this today.
> 
> A initial feedback, already given before, is this: IMHO, we should propose PRs with commits containing a feature or a fix (like any other well maintained upstream project that I am aware of).
> 
> So, I'd rather have all this squashed (it could be at the end of reviews, ok for me) and the git log to be a complete description of everything that is changed (instead of things such as "Doc fix", "Changed files structure") OR this could be rebased into multiple commits, each of them being explicit about changes introduced and reasoning (@AlonZivony you can keep your git history in another local branch, if you'd like, like you said previously).
> 
> It's a good practice also to inform what part of the tree you're touching in your commit log ("tracee-ebpf: xxxx", "containers: yyyyy"). Not for the reviewer, but for future reference.
> 
> A big set of commits with small descriptions is not good if someone would backport this feature to an older version, for example (if we had a LTS tree).
> 
> Since we're not merging branches in our repository, it is hard to understand which commit is responsible for the feature introduction and which one is the end of it (just one example on why we should work like other upstream projects).
> 
> Let's say someone approves this PR, it would be "ready for merge" with a big set of commits that don't mean anything to the entire project.
> 
> Now, I'll get to the technical part (which looks very good from a first fast look).

I agree with what you said here for the most parts.
The commits in the PR are not meant to be part of the `main` branch history, I agree with you that before the merge it should be squashed.
However, for the development stage, small and specific changes contained in each commit are very comfortable. For example, if I want to revert some change it is very handy to have a commit that contain it so I could just use git revert for the operation.
In the context of the PR I think the commit messages are understandable (correct me if I am wrong) and we need to remember that commit message should not be more than 50 characters for the subject (the internal could be in any length, but it is not visible for most view tools), so adding the parts I am touching is problematic. The small commits also helps to see what code was written when, and the commit message can help you understand to goal of the code and help you in the CR.

rafaeldtinoco:
About the git commits.. that's why can keep the commits in another local branch if you'd like but, by the time you propose it to be merged, they have to be ready to be merged. What if someone gives you a +1 and rebases/merges ? Then we would have a bunch of commits in the log that are related to a single change without proper description.

> In the context of the PR I think the commit messages are understandable (correct me if I am wrong) and we need to remember that commit message should not be more than 50 characters for the subject (the internal could be in any length, but it is not visible for most view tools), so adding the parts I am touching is problematic. The small commits also helps to see what code was written when, and the commit message can help you understand to goal of the code and help you in the CR.

You don't have to put the entire path in the git log commit. You have to refer to which parts of the code base you're touching in EACH commit, that is a best practice. You could create a prefix for your commits: "sorting: XXXX", for example. 

If you don't trust me for this, please trust other open source projects following the same guide lines:

0. [linux kernel](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/log/)
1. https://git.qemu.org/?p=qemu.git;a=log
2. https://libvirt.org/git/?p=libvirt.git
3. https://sourceware.org/git/?p=glibc.git
4. https://github.com/corosync/corosync/commits/main
5. https://github.com/ClusterLabs/pacemaker/commits/main
6. https://github.com/opensvc/multipath-tools/commits/master
7. https://github.com/systemd/systemd/commits/main

and ALL other successful open-source projects I'm aware of work this way, why shouldn't we ?

> However, if you think that managing small amount of big commits is easier I am OK with what you suggested - keeping a local branch with all the small commits. Tell me what you think better suits the project habits and I will follow it.

I would like us to follow best practices for open-source upstream projects but last word is not mine I'm afraid.

Let's see what others have to say.

danielpacak:
> About the git commits.. that's why can keep the commits in another local branch if you'd like but, by the time you propose it to be merged, they have to be ready to be merged. What if someone gives you a +1 and rebases/merges ? Then we would have a bunch of commits in the log that are related to a single change without proper description.
> 
> > In the context of the PR I think the commit messages are understandable (correct me if I am wrong) and we need to remember that commit message should not be more than 50 characters for the subject (the internal could be in any length, but it is not visible for most view tools), so adding the parts I am touching is problematic. The small commits also helps to see what code was written when, and the commit message can help you understand to goal of the code and help you in the CR.
> 
> You don't have to put the entire path in the git log commit. You have to refer to which parts of the code base you're touching in EACH commit, that is a best practice. You could create a prefix for your commits: "sorting: XXXX", for example.
> 
> If you don't trust me for this, please trust other open source projects following the same guide lines:
> 
> 0. [linux kernel](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/log/)
> 1. https://git.qemu.org/?p=qemu.git;a=log
> 2. https://libvirt.org/git/?p=libvirt.git
> 3. https://sourceware.org/git/?p=glibc.git
> 4. https://github.com/corosync/corosync/commits/main
> 5. https://github.com/ClusterLabs/pacemaker/commits/main
> 6. https://github.com/opensvc/multipath-tools/commits/master
> 7. https://github.com/systemd/systemd/commits/main
> 
> and ALL other successful open-source projects I'm aware of work this way, why shouldn't we ?
> 
> > However, if you think that managing small amount of big commits is easier I am OK with what you suggested - keeping a local branch with all the small commits. Tell me what you think better suits the project habits and I will follow it.
> 
> I would like us to follow best practices for open-source upstream projects but last word is not mine I'm afraid.
> 
> Let's see what others have to say.

I agree 💯 with @rafaeldtinoco with git commits logs to keep them tidy and follow some kind of conventions. I would never use it as a developer notebook, but rather as something that helps building release notes or changes log (automatically). More importantly, I'd expect the commit message to give me an idea of what should I test and focus on during review. Last not but least, a meaningful and concise commit message might attract other community member to join the effort.

rafaeldtinoco:
Alright, @AlonZivony and I have talked in a chat and agreed that he will receive a +1 to squash. After the squash we can do the official +1 on the review. He prefers working that way and, to be honest, its ok for me as long as we don't merge the non squashed version of the commits.

AlonZivony:
So to make the git commits situation clear:
I will continue with current commits until the code in this PR will pass the CRs.
Before the maintainers give me the option to merge I will be requested to squash the commits and give the better commit messages. After the log will pass your review I will get the +1 for the PR and merge it to the `main` branch.
Thanks @rafaeldtinoco and @danielpacak for you review about the Git best practices :)

yanivagman:
> So to make the git commits situation clear: I will continue with current commits until the code in this PR will pass the CRs. Before the maintainers give me the option to merge I will be requested to squash the commits and give the better commit messages. After the log will pass your review I will get the +1 for the PR and merge it to the `main` branch. Thanks @rafaeldtinoco and @danielpacak for you review about the Git best practices :)

TBH I don't think that such a PR should be squashed into 1 commit.
It will probably be better to split it into several. e.g.:
1. Introduce pool
2. Introduce sorter
3. Sort events

Of course, other possibilities exist.

I can understand that it might be easier for you to have all your commit history while developing, but this is not something that makes the life of a reviewer easier.

What works for me is the following:
1. Developing my own branch with many small commits
2. When I'm satisfied with the code, I checkout a new branch from the original development branch
3. I then craft commits that can be easily digestable by a reviewer and which make sense. Each logical part that is self contained gets its own commit.
4. When I get review comments, I'm adding new commits on top of the new branch.
5. I then use "rebase -i" to insert the changes to the relevant commits. Of course, before doing "rebase -i" you can open a new branch, and repeat the process described above.

rafaeldtinoco:
> What works for me is the following:
> 
> 1. Developing my own branch with many small commits
> 2. When I'm satisfied with the code, I checkout a new branch from the original development branch
> 3. I then craft commits that can be easily digestable by a reviewer and which make sense. Each logical part that is self contained gets its own commit.
> 4. When I get review comments, I'm adding new commits on top of the new branch.
> 5. I then use "rebase -i" to insert the changes to the relevant commits. Of course, before doing "rebase -i" you can open a new branch, and repeat the process described above.

This is what I do and see as a "best practice" for PRs as well.

AlonZivony:
> I can understand that it might be easier for you to have all your commit history while developing, but this is not something that makes the life of a reviewer easier.

I am not so sure that I understand why this is a best practice for a PR.
I divided the code to files and packages, so reviewing the Pool, Queue and Sorter separately is not an issue. You don't need them to be under a commit to do so. You can see all the changes in the diff tab, and the specific commit message give you even more information regarding the code, when calling it "Introduce Pool" doesn't give much.
In the end during the PR lifetime the commits are a tool for documenting the history and assisting solving CR comments.
As I said before, I totally agree that the current commits shouldn't get into the `main` branch history, because this specific commits are not helpful in managing the code and are only disturbing the log of the project.
If you can give me a good reason why this amount of commits disturb the reviewer I will rethink it, but currently I feel that squashing the commits before the PR is approved is not helpful, and only make the flow of giving CR and fixing the comments harder.

I think we should discuss the Git best practices on a proper discussion rather than in a PR, WDYT?

pkg/utils/environment/cpu_analyzing.go
```diff
@@ -0,0 +1,56 @@
+package environment
```
rafaeldtinoco, line 1: This entire file can be replaced by [sysconf implementation](https://pkg.go.dev/github.com/tklauser/go-sysconf) for golang OR something as simple as:

```go
// #cgo LDFLAGS:
// #include <unistd.h>
import "C"
import "fmt"
func main() {
    fmt.Printf("%d\n", C.sysconf(C._SC_NPROCESSORS_CONF))
}
```

AlonZivony: Nice!
Thanks for this comment, I didn't know it was implemented so easily to use.
We need to tell Datadog ;)

yanivagman: Please don't do that!

Also see this thread:
https://lore.kernel.org/bpf/CAEf4Bza6Bsee1i_ypbDogG5MsVFGW9pnatxHCn9PycW9eP2Gkw@mail.gmail.com/

rafaeldtinoco: Hum, okay, I've tested in multiple VMs attaching and detaching CPUs before suggesting it but, likely, I haven't tested it enough. Thanks @yanivagman

rafaeldtinoco: @AlonZivony could you add a comment with that link then ? So in the future we always remember why that logic exists instead ?

AlonZivony: No problem

rafaeldtinoco:
> To summarize the algorithm - we have a CPU queue for each CPU. We insert new
> events to the matching CPU's queue, and follow which CPU was updated with new
> event. Each interval, we check from the most recent events from each CPU
> which has the oldest timestamp. After a delay of at least 100ms, we send all
> events from queues up to that timestamp in an ordered way. This way, we can
> be sure that all sent events are sorted.

Could you confirm my understanding, then, described below, and, if right, add this as a comment to your EventsChronoOrder class.

```
Tn = Timestamp (n == TOD)

## Scheduler Tick #1

[ CPU 0 ]    [ CPU 1 ]    [ CPU 2 ]
  T1           T2           T4
  T3           T5           T9  +
  T6           T8  +        T10 +
  T7           T12 +        T11 +

    - oldest timestamp = T1.
    - T7 is oldest timestamp in most recent timestamps.
    - In 100ms from this tick we will send all events from T1 to T7.
    - Bigger timestamps than T7 (+) will be sent in next scheduling.

## Scheduler Tick #2

[ CPU 0 ]    [ CPU 1 ]    [ CPU 2 ]
  T20          T8  old      T9  old
  T21 +        T12 old      T10 old
               T22 +        T11 old
               T25 +        T23 +
                            T24 +

    - oldest timestamp = T8
    - T20 is oldest timestamp in most recent timestamps
    - In 100ms from this tick, send all events from T8 to T20
    - Bigger timestamps than T20 (+) will be sent in next scheduling.

## Scheduler Tick #3

[ CPU 0 ]    [ CPU 1 ]    [ CPU 2 ]
  T21 old      T22 old      T23 old
  T30          T25 old      T24 old
  T31          T34 +        T35 +
  T32          T50 +        T36 +
  T33          T51 +        T37 +
                            T38 +

    - oldest timestamp = T21
    - T33 is oldest timestamp in most recent timestamps
    - In 100ms from this tick, send all events from T21 to T33
    - Bigger timestamps than T33 (+) will be sent in next scheduling.
```

rafaeldtinoco:
I haven't investigated further, but... Something seems wrong when I trace all exiting events. 85 Events were count but I had no output but the initial one:

```
$ sudo ./dist/tracee-ebpf --trace event="*" --output option:sort-events
02:36:45:728831  0      tracee-ebpf      0       0       0                init_namespaces      cgroup: 4026531835, ipc: 4026531839, mnt: 4026531840, net: 4026532008, pid: 4026531836, pid_for_children: 4026531836, time: 4026531834, time_for_children: 4026531834, user: 4026531837, uts: 4026531838

End of events stream
Stats: {EventCount:85 ErrorCount:0 LostEvCount:0 LostWrCount:0 LostNtCount:0}
```

But for the default events, I get output:


```
$ sudo ./dist/tracee-ebpf --output option:sort-events
...
02:37:51:210640  113    nxserver.bin     1471543 1471554 29               openat               dirfd: -100, pathname: /proc/1471543/stat, flags: O_RDONLY, mode: 0
02:37:51:210682  113    nxserver.bin     1471543 1471554 0                security_file_open   pathname: /proc/1471543/stat, flags: O_RDONLY|O_LARGEFILE, dev: 22, inode: 4792353, ctime: 1642774452017546462
```

And, If I filter for a specific command, I also get output...

```
$ sudo ./dist/tracee-ebpf --output option:sort-events --trace event="*" --trace comm="bash" --trace follow
02:39:57:420761  0      tracee-ebpf      0       0       0                init_namespaces      cgroup: 4026531835, ipc: 4026531839, mnt: 4026531840, net: 4026532008, pid: 4026531836, pid_for_children: 4026531836, time: 4026531834, time_for_children: 4026531834, user: 4026531837, uts: 4026531838
02:39:58:781923  1000   bash             2783809 2783809 1                read                 fd: 0, buf: 0x7FFCEA935ABF, count: 1
02:39:58:781923  1000   bash             2783809 2783809 0                sys_enter            syscall: read
```

But not for all events.

AlonZivony:
> I haven't investigated further, but... Something seems wrong when I trace all exiting events. 85 Events were count but I had no output but the initial one:
> 
> ```
> $ sudo ./dist/tracee-ebpf --trace event="*" --output option:sort-events
> TIME             UID    COMM             PID     TID     RET              EVENT                ARGS
> 02:36:45:728831  0      tracee-ebpf      0       0       0                init_namespaces      cgroup: 4026531835, ipc: 4026531839, mnt: 4026531840, net: 4026532008, pid: 4026531836, pid_for_children: 4026531836, time: 4026531834, time_for_children: 4026531834, user: 4026531837, uts: 4026531838
> 
> End of events stream
> Stats: {EventCount:85 ErrorCount:0 LostEvCount:0 LostWrCount:0 LostNtCount:0}
> ```
> 
> But for the default events, I get output:
> 
> ```
> $ sudo ./dist/tracee-ebpf --output option:sort-events
> TIME             UID    COMM             PID     TID     RET              EVENT                ARGS
> ...
> 02:37:51:210640  113    nxserver.bin     1471543 1471554 29               openat               dirfd: -100, pathname: /proc/1471543/stat, flags: O_RDONLY, mode: 0
> 02:37:51:210682  113    nxserver.bin     1471543 1471554 0                security_file_open   pathname: /proc/1471543/stat, flags: O_RDONLY|O_LARGEFILE, dev: 22, inode: 4792353, ctime: 1642774452017546462
> ```
> 
> And, If I filter for a specific command, I also get output...
> 
> ```
> $ sudo ./dist/tracee-ebpf --output option:sort-events --trace event="*" --trace comm="bash" --trace follow
> TIME             UID    COMM             PID     TID     RET              EVENT                ARGS
> 02:39:57:420761  0      tracee-ebpf      0       0       0                init_namespaces      cgroup: 4026531835, ipc: 4026531839, mnt: 4026531840, net: 4026532008, pid: 4026531836, pid_for_children: 4026531836, time: 4026531834, time_for_children: 4026531834, user: 4026531837, uts: 4026531838
> 02:39:58:781923  1000   bash             2783809 2783809 1                read                 fd: 0, buf: 0x7FFCEA935ABF, count: 1
> 02:39:58:781923  1000   bash             2783809 2783809 0                sys_enter            syscall: read
> ```
> 
> TL;DR: all events with sorting is not giving me any output.

I will investigate it.
This is weird because I tested it manually myself couple of times but got all expected events each time.
Thanks for the check!

AlonZivony:
> > To summarize the algorithm - we have a CPU queue for each CPU. We insert new events to the matching CPU's queue, and follow which CPU was updated with new event. Each interval, we check from the most recent events from each CPU which has the oldest timestamp. After a delay of at least 100ms, we send all events from queues up to that timestamp in an ordered way. This way, we can be sure that all sent events are sorted.
> 
> Could you confirm my understanding, then, described below, and, if right, add this as a comment to your EventsChronoOrder class.
> 
> ```
> Tn = Timestamp (n == TOD)
> 
> ## Scheduler Tick #1
> 
>     [ CPU 0 ]    [ CPU 1 ]    [ CPU 2 ]
> HEAD  T1           T2           T4
>       T3           T5           T9  +
>       T6           T8  +        T10 +
> TAIL  T7           T12 +        T11 +
> 
>     - oldest timestamp = T1.
>     - T7 is oldest timestamp in most recent timestamps.
>     - In 2 ticks from now: send all events up to T7.
>     - Bigger timestamps than T7 (+) will be sent in next scheduling.
> 
> ## Scheduler Tick #2
> 
>     [ CPU 0 ]    [ CPU 1 ]    [ CPU 2 ]
> HEAD  T20          T8  old      T9  old
>       T21          T12 old      T10 old
>                    T22 +        T11 old
>                    T25 +        T23 +
> TAIL                            T24 +
> 
>     - oldest timestamp = T8
>     - T21 is oldest timestamp in most recent timestamps
>     - In 2 ticks from now: send all events up to T21.
>     - Bigger timestamps than T20 (+) will be sent in next scheduling.
> 
> ## Scheduler Tick #3
> 
>     [ CPU 0 ]    [ CPU 1 ]    [ CPU 2 ]
> HEAD  T30          T22 old      T23 old
>       T31          T25 old      T24 old
>       T32          T34 +        T35 +
>       T33          T50 +        T36 +
>                    T51 +        T37 +
> TAIL                            T38 +
> 
>     - oldest timestamp = T22
>     - T33 is oldest timestamp in most recent timestamps
>     - In 2 ticks from now: send all events up to T33.
>     - Bigger timestamps than T33 (+) will be sent in next scheduling.
> ```

You are correct - I will add it to the docs

AlonZivony:
So after short session I found a maybe unrelated bug - the pipeline close does not block the main exit.
As a result, the printer is closed before all events handling is proven to be done, which might cause events lost with delayed events send like with the sorter.
I will probably open an issue on it soon.

yanivagman:
related #1360 - maybe we can fix it there?

AlonZivony:
> related #1360 - maybe we can fix it there?

I am not sure that is the responsibility of that PR but it is close in the logic.
I will open an issue and if @liamg will think he can put it in the PR in an appropriate way he will fix it.

AlonZivony:
I will mark this PR as a draft until I am sure the bug is  solved (currently working on fixing it)

AlonZivony:
> @AlonZivony is this good for a new review ?

It's great.
I will address the CR soon, I just have few other important issues to address that are more prioritized for this moment but I will try to address the CR as soon as I can.

rafaeldtinoco:
Okay @AlonZivony 

I think we're good to go... I think PR could have the following commits:

```
1) main: external.Event passed as pointer                       # change to main.go, tracee.go and events_pipeline.go
2) main, external & tracee.bpf: introduce ProcessorID           # change to tracee.go, tracee.bpf.c and events_pipeline.go
3) utils/environment: parse the number of CPUs available        # cpu_analyzing.go
4) events/sorting: introduce event pool                         # pool.go
5) events/sorting: introduce queue & cpuqueue                   # queue.go & cpu_queue.go
6) events/sorting: introduce events sorter                      # sorting.go
7) events/sorting: introduce events sorter tests                # sorting_test.go
8) events_pipeline: sort events optionally                      # change to events_pipeline.go
9) main_test: add sort-events to tests                          # change to main_test.go
```

And, like showed in comments after suggested titles, most of them are single
file commits (or small changes to existing files) so I think that, with these
commits, your rebasing commits work will be minimal and, still, allow good
maintainability for cherry-picks and further fixes.

Also, are you addressing things I mentioned in comments ? Like adding the algorithm description to the source code, some changes to error messages, a missing link to next/previous for empty pool nodes (iirc). 

Looking forward to the final PR so I can +1 it officially.

cmd/tracee-ebpf/internal/flags/flags-output.go
```diff
@@ -20,14 +20,15 @@ Possible options:
...
+option:{stack-addresses,detect-syscall,exec-env,relative-time,exec-hash,parse-arguments,sort-events}
                                                    augment output according to given options (default: none)
   stack-addresses                                  include stack memory addresses for each event
   detect-syscall                                   when tracing kernel functions which are not syscalls, detect and show the original syscall that called that function
   exec-env                                         when tracing execve/execveat, show the environment variables that were used for execution
   relative-time                                    use relative timestamp instead of wall timestamp for events
   exec-hash                                        when tracing sched_process_exec, show the file hash(sha256) and ctime
   parse-arguments                                  do not show raw machine-readable values for event arguments, instead parse into human readable strings
+  sort-events                                      enable sorting events before passing to them output, which will result performance reduction
```
mtcherni95, line 31: nit: I would rephrase the last part to "this will decrease the overall program efficiency"

pkg/events/sorting/queue.go
```diff
@@ -0,0 +1,108 @@
...
+
+// Get remove the node at the head of the queue and return it
+// Might return error with a valid event in case of internal pool error, for the user to know that an error occurred
+// but was contained
+func (eq *eventsQueue) Get() (*external.Event, error) {
+	eq.mutex.Lock()
+	defer eq.mutex.Unlock()
+	if eq.head == nil {
+		if eq.tail != nil {
+			return nil, fmt.Errorf("BUG: TAIL without a HEAD")
```
mtcherni95, line 41: I would say "discrepancy: tail without head"

yanivagman (approved):
LGTM

rafaeldtinoco (approved):
LGTM @AlonZivony, awesome job.

AlonZivony:
Thanks!

PR #1108, opened 2021-10-24
separate pcaps by process context

roikol:
the purpose of this PR is to save a separate pcap for each process on the host.

we distinguish processes by their PID.
but as we all know - PIDs can be reassigned to new processes. this behavior would cause us to save different processes packets in the same pcap file. to overcome this problem, ideally, we would use the SCHED_PROCESS_EXIT event to close the pcap file and rename it with unique info (such as the timestamp of the SCHED_PROCESS_EXIT event) - so no more packets will be written to this pcap file.
after trying this method on my machine, i have noticed that in some cases we get the SCHED_PROCESS_EXIT event, and then we get some more packets from this process.
therefore I've decided to use the SCHED_PROCESS_EXEC event - when we receive this event for a PID we already have pcap for, we can rename the old pcap and avoid the problem described above.
this is not ideal as the pcapWriters are stored in memory and are never deleted (only deleted when SCHED_PROCESS_EXEC event for PID we already have pcap for, but then we create a new pcapWriter)

a better approach might be to tap into another tracepoint, one that is ensured that no packets will be sent/reveived after it.
i tried tracepoint:sched_process_free and kprobe:do_task_dead - both are not ensuring no packets will be sent/reveived after them.

meanwhile, i'm creating this PR as draft. 
if anyone has any insight about this i would love to hear.

also - 
do you think we should save separate pcaps only on user request (via a flag)?
or maybe we should keep a copy of the entire pcap of the host on top of the separated pcaps?

thanks,
Roi.

### DIFF
--- a/tracee-ebpf/main.go
+++ b/tracee-ebpf/main.go
@@ -345,7 +345,7 @@ err-file:/path/to/file                             write the errors to a specifi
 
 none                                               ignore stream of events output, usually used with --capture
 
-option:{stack-addresses,detect-syscall,exec-env,relative-time,exec-info,per-process-pcaps}
+option:{stack-addresses,detect-syscall,exec-env,relative-time,exec-info,per-process-pcaps,per-container-pcaps}
                                                    augment output according to given options (default: none)
   stack-addresses                                  include stack memory addresses for each event
   detect-syscall                                   when tracing kernel functions which are not syscalls, detect and show the original syscall that called that function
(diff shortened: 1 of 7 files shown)


yanivagman:
@roikol can we close this PR now that #1444 is merged?

roikol:
> @roikol can we close this PR now that #1444 is merged?

yes we can

PR #1110, opened 2021-10-25
tracee-ebpf: fix arm64 build

### LINKED ISSUE #1109: ARM64 build fails
There are several things that needs to be fixed to build tracee-ebpf in ARM64:

1. Add vmlinux.h for ARM64 arch
2. stack size exceeds 512 bytes in sys_enter tracepoint
3. Some macros are not defined correctly for CO-RE
4. sys_exit program tries to read orig_ax register which doesn't exists in ARM64 arch

yanivagman:
This PR fixes several things that caused ARM64 build to fail:
1. Add vmlinux.h for ARM64 arch
2. Don't read all `struct pt_regs` in sys_enter to save space on stack
3. Use `regs->syscallno` instead of `regs->orig_ax` in sys_enter
4. Define missing _TIF_32BIT macro
5. Use default value for PAGE_SHIFT (as a temporary workaround)

Fixes #1109


rafaeldtinoco (commented):
Some comments:

- my default arm64 image had mawk (instead of gawk) installed by default and compilation brakes (LINUX_VERSION_CODE Makefile macro needs gawk). Had to install gawk (not sure if that is documented).

- Im currently building with LLVM-13 and getting a bunch of warnings:

```
tracee/tracee.bpf.c:1910:28: warning: passing 'typeof ((((const volatile struct user_pt_regs
      *)(regs))->regs[0])) *' (aka 'const volatile unsigned long long *') to parameter of type 'void *' discards
      qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
        sys.args.args[0] = READ_KERN(PT_REGS_PARM1(regs));
                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

and clang did not like `-Wno-incompatible-pointer-types-discards-qualifiers` passed. I think you might need to cast the qualifier discard (unsure about LLVM12).

- Is tracee-ebpf loading for you ? I have verifier issues in 5.13.0-20-generic aarch64 kernel. Important part:

```
875: (7b) *(u64 *)(r10 -296) = r1
876: (bf) r1 = r10
877: (07) r1 += -296
; bpf_probe_read(&sockaddr, addr_len, (void *)address);
878: (bf) r2 = r7
879: (79) r3 = *(u64 *)(r10 -320)
880: (85) call bpf_probe_read#4
R2 min value is negative, either use unsigned or 'var &= const'
processed 875 insns (limit 1000000) max_states_per_insn 1 total_states 53 peak_states 53 mark_read 41
```

rafaeldtinoco:
> ```
> tracee/tracee.bpf.c:1910:28: warning: passing 'typeof ((((const volatile struct user_pt_regs
>       *)(regs))->regs[0])) *' (aka 'const volatile unsigned long long *') to parameter of type 'void *' discards
>       qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
>         sys.args.args[0] = READ_KERN(PT_REGS_PARM1(regs));
> ```
> 
> The problem is that the PT_REGS_PARMX helpers for arm64 define the variables as const, (unlike the x86 macros, which don't use an extra `user_pt_regs struct`): https://elixir.bootlin.com/linux/v5.14.17/source/tools/lib/bpf/bpf_tracing.h#L194. Then, the `READ_KERN` macro uses this type to define its `val` variable: `#define READ_KERN(ptr) ({ typeof(ptr) _val; `. This is then fed to `__builtin_memset` and `bpf_probe_read_kernel` which expects void*, thus discards the const qualifier. I don't think I can just do a regular cast here, as these types change between architectures.

```diff
From 59207e5c553dfbee4884e4411d85f15610f2625c Mon Sep 17 00:00:00 2001
From: Rafael David Tinoco <rafaeldtinoco@gmail.com>
Date: Tue, 9 Nov 2021 00:08:41 -0300
Subject: [PATCH] tracee.bpf: fix READ_KERN incompat ptr type discards

READ_KERN() is discarding "const" types. Cast macro functions arguments
to (void *) type in order to avoid the need of adding compiler overall
warning ignore for this type of error.

    Error example when compiling arm64:

    tracee/tracee.bpf.c:1955:28: warning: passing 'typeof ((((const volatile
    struct user_pt_regs *)(regs))->regs[0])) *' (aka 'const volatile
    unsigned long long *') to parameter of type 'void *' discards qualifiers
    [-Wincompatible-pointer-types-discards-qualifiers]

    sys.args.args[0] = READ_KERN(PT_REGS_PARM1(regs));
                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    tracee/tracee.bpf.c:304:41: note: expanded from macro 'READ_KERN'
    bpf_core_read(&_val, sizeof(_val), &ptr);
                  ^~~~~
---
 tracee-ebpf/tracee/tracee.bpf.c | 16 ++++++++--------
 1 file changed, 8 insertions(+), 8 deletions(-)

diff --git a/tracee-ebpf/tracee/tracee.bpf.c b/tracee-ebpf/tracee/tracee.bpf.c
index b0939b6..c77dfa8 100644
--- a/tracee-ebpf/tracee/tracee.bpf.c
+++ b/tracee-ebpf/tracee/tracee.bpf.c
@@ -290,19 +290,19 @@ Copyright (C) Aqua Security inc.
 #endif

 #ifndef CORE
-#define READ_KERN(ptr) ({ typeof(ptr) _val;                             \
-                          __builtin_memset(&_val, 0, sizeof(_val));     \
-                          bpf_probe_read(&_val, sizeof(_val), &ptr);    \
-                          _val;                                         \
+#define READ_KERN(ptr) ({ typeof(ptr) _val;                                             \
+                          __builtin_memset((void *)&_val, 0, sizeof(_val));             \
+                          bpf_probe_read((void *)&_val, sizeof(_val), (void *)&ptr);    \
+                          _val;                                                         \
                         })

 #define GET_FIELD_ADDR(field) &field
 #else
 // Try using READ_KERN here, just don't embed them in each other
-#define READ_KERN(ptr) ({ typeof(ptr) _val;                             \
-                          __builtin_memset(&_val, 0, sizeof(_val));     \
-                          bpf_core_read(&_val, sizeof(_val), &ptr);    \
-                          _val;                                         \
+#define READ_KERN(ptr) ({ typeof(ptr) _val;                                             \
+                          __builtin_memset((void *)&_val, 0, sizeof(_val));             \
+                          bpf_core_read((void *)&_val, sizeof(_val), (void *)&ptr);     \
+                          _val;                                                         \
                         })

 #define GET_FIELD_ADDR(field) __builtin_preserve_access_index(&field)
--
2.32.0
```

rafaeldtinoco:
> > my default arm64 image had mawk (instead of gawk) installed by default and compilation brakes (LINUX_VERSION_CODE Makefile macro needs gawk). Had to install gawk (not sure if that is documented).
> 
> I see that we use awk here: `LINUX_VERSION_CODE := $(shell uname -r | awk '{split($$0,a,".").....` But do we really need it? The only place that uses this variable (in the makefile) is the CO-RE build - but why would we want CO-RE to be aware of the linux kernel version it was built with? The only reason for that is the use of LINUX_VERSION_CODE in the bpf code, but then we should use runtime checks, and not compilation checks (so this is actually a bug!). I'll open an issue to track this one.

+1 on this.

yanivagman:
@rafaeldtinoco are you sure that it is ok to cast `(void *)&ptr` for `bpf_core_read()`?
bpf_core_read is eventually bpf_probe_read with __preserve_access_index applied on the third arguments (&ptr) - won't changing this make it useless?

rafaeldtinoco:
> @rafaeldtinoco are you sure that it is ok to cast `(void *)&ptr` for `bpf_core_read()`?
> `bpf_core_read()` is eventually `bpf_probe_read()` with `__preserve_access_index` applied on the third argument (&ptr) - won't changing this make it useless?

Wow, nice thinking. I'm not sure it would, but I would bet it could and might. So its safer not to cast it, I agree.

rafaeldtinoco (approved):
LGTM.

PR #1121, opened 2021-11-02
replacing exec-info with exec-hash and removing ctime calculation

mtcherni95:


yanivagman:
I think it will be more logical to merge this PR with #1089 - one commit to remove the old ctime mechanism and another commit to add it as an argument

mtcherni95:
cherry-picked this commit to pr #1089

PR #1122, opened 2021-11-03
init_module capture

AsafEitani:
Add init_module capture.
Split `send_bin` to helper and calling tail bpf programs for `raw_tracepoint` and `kprobe`.

The split was needed because a tail call cannot be called to a different type bpf program.

### DIFF
--- a/tracee-ebpf/tracee/consts.go
+++ b/tracee-ebpf/tracee/consts.go
@@ -90,7 +90,6 @@ const (
 	sendVfsWrite binType = iota + 1
 	sendMprotect
 	sendKernelModule
-	sendBufferModule
 )
 
 // argType is an enum that encodes the argument types that the BPF program may write to the shared buffer
--- a/tracee-ebpf/tracee/consts.go
+++ b/tracee-ebpf/tracee/consts.go
@@ -79,7 +79,7 @@ const (
 	tailVfsWrite uint32 = iota
 	tailVfsWritev
 	tailSendBin
-	tailSendBinSyscall
+	tailSendBinTP
 )
 
 // binType is an enum that specifies the type of binary data sent in the file perf map
--- a/tracee-ebpf/tracee/consts.go
+++ b/tracee-ebpf/tracee/consts.go
@@ -79,6 +79,7 @@ const (
 	tailVfsWrite uint32 = iota
 	tailVfsWritev
 	tailSendBin
+	tailSendBinSyscall
 )
 
 // binType is an enum that specifies the type of binary data sent in the file perf map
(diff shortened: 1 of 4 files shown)


tracee-ebpf/tracee/consts.go
```diff
@@ -79,6 +79,7 @@ const (
 	tailVfsWrite uint32 = iota
 	tailVfsWritev
 	tailSendBin
+	tailSendBinSyscall
```
yanivagman, line 82: It's not exactly related to syscall, but raw_tracepoint bpf program type.
The other one is for kprobe program type.
So maybe tailSendBinTP will be a better name?

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -967,6 +970,7 @@ static __always_inline struct sockaddr_un get_unix_sock_addr(struct unix_sock *s
     return sockaddr;
 }
 
+
```
yanivagman, line 973: nit: no need for this extra line

tracee-ebpf/tracee/write_capture.go
```diff
@@ -33,6 +33,11 @@ func (t *Tracee) processFileWrites() {
 		Size  uint64
 	}
 
+	type bufferModuleMeta struct {
+		Pid  uint32
+		Size uint64
+	}
```
yanivagman, line 39: Can't we just use kernelModuleMeta for this as well?
If you will set both DevID and Inode to zero (from bpf code), all of the below logic won't be needed

tracee-ebpf/tracee/consts.go
```diff
@@ -89,6 +90,7 @@ const (
 	sendVfsWrite binType = iota + 1
 	sendMprotect
 	sendKernelModule
+	sendBufferModule
```
yanivagman, line 93: Do we need this now?

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -3757,6 +3771,42 @@ int BPF_KPROBE(trace_mprotect_alert)
...
+    syscall_data_t *sys = bpf_map_lookup_elem(&syscall_data_map, &data.context.host_tid);
+    if (!sys)
+        return -1;
+
+    bin_args_t bin_args = {};
+
+    u32 pid = data.context.host_pid;
+    u32 dev = 0;
+    u64 inode = 0;
```
yanivagman, line 3789: No meaning for having two variables here, as these values don't change (constant zero).
You can have `u64 dummy = 0` and use it for both bpf_probe_read below

yanivagman (approved):

Issue #461, opened 2021-01-27
print raw arguments


yanivagman:
Currently, tracee-ebpf pretty-prints some of the events arguments (e.g. execve and clone flags).
We should enable the user to choose if he wants to print the arguments as raw.
To do that, we can add a new output option to the output flag.

itaysk:
we have recently discussed this and agreed to change tracee-ebpf to send only raw data (for argument values), and the consumer (tracee-rules or tracee-ebpf printer) will be able to present them differently. to support that, we will move "argPrinters" from Tracee to libbpfgo (#485).

itaysk:
now let's discuss the UX: I think that the by default, only table printer should print decoded args. all the rest should send raw args.
we can say this is controllable using an output "option" like `-o option:printargs`. setting this flag will take the extra step of printing the args into textual form, and by default this will be enabled only for table output. 
how does this sound?

PR #1123, opened 2021-11-03
Add option for raw arguments from various event flags

grantseltzer:
This handles #461 

Based on conversation with Yaniv/Rafael we decided to make raw arguments opt-in, so they're parsed into strings by default (for all output formats). Raw arguments can be received by specifying `--output option:raw-arguments`. 

### DIFF
--- a/docs/tracee-ebpf/output.md
+++ b/docs/tracee-ebpf/output.md
@@ -10,13 +10,13 @@ CLI Option | Description
 `none` | ignore stream of events output, usually used with `--capture`
 `out-file:/path/to/file` | write the output to a specified file. the path to the file will be created if not existing and the file will be deleted if existing (default: stdout)
 `err-file:/path/to/file` | write the errors to a specified file. the path to the file will be created if not existing and the file will be deleted if existing (default: stderr)
-`option:{stack-addresses,detect-syscall,exec-env,exec-info,relative-time,raw-arguments}` | augment output according to given options (default: none)
+`option:{stack-addresses,detect-syscall,exec-env,exec-info,relative-time,parse-arguments}` | augment output according to given options (default: none)
   stack-addresses | include stack memory addresses for each event
   detect-syscall | when tracing kernel functions which are not syscalls, detect and show the original syscall that called that function
   exec-env | when tracing execve/execveat, show the environment variables that were used for execution
 exec-info | when tracing sched_process_exec, show the file hash(sha256) and ctime
 relative-time | use relative timestamp instead of wall timestamp for events
-raw-arguments | do not show raw machine-readable values for event arguments, instead parse into human readable strings
+parse-arguments | do not show raw machine-readable values for event arguments, instead parse into human readable strings
 
 
 
--- a/docs/tracee-ebpf/output.md
+++ b/docs/tracee-ebpf/output.md
@@ -10,12 +10,14 @@ CLI Option | Description
 `none` | ignore stream of events output, usually used with `--capture`
 `out-file:/path/to/file` | write the output to a specified file. the path to the file will be created if not existing and the file will be deleted if existing (default: stdout)
 `err-file:/path/to/file` | write the errors to a specified file. the path to the file will be created if not existing and the file will be deleted if existing (default: stderr)
-`option:{stack-addresses,detect-syscall,exec-env,exec-info,relative-time}` | augment output according to given options (default: none)
+`option:{stack-addresses,detect-syscall,exec-env,exec-info,relative-time,raw-arguments}` | augment output according to given options (default: none)
   stack-addresses | include stack memory addresses for each event
   detect-syscall | when tracing kernel functions which are not syscalls, detect and show the original syscall that called that function
   exec-env | when tracing execve/execveat, show the environment variables that were used for execution
 exec-info | when tracing sched_process_exec, show the file hash(sha256) and ctime
 relative-time | use relative timestamp instead of wall timestamp for events
+raw-arguments | show raw values for event arguments that are otherwise parsed into strings, such as clone(2) flags
+
 
 
 (Use this flag multiple times to choose multiple capture options)
--- a/docs/tracee-ebpf/output.md
+++ b/docs/tracee-ebpf/output.md
@@ -16,7 +16,7 @@ CLI Option | Description
   exec-env | when tracing execve/execveat, show the environment variables that were used for execution
 exec-info | when tracing sched_process_exec, show the file hash(sha256) and ctime
 relative-time | use relative timestamp instead of wall timestamp for events
-raw-arguments | show raw values for event arguments that are otherwise parsed into strings, such as clone(2) flags
+raw-arguments | do not show raw machine-readable values for event arguments, instead parse into human readable strings
 
 
 
--- a/tracee-ebpf/main.go
+++ b/tracee-ebpf/main.go
@@ -335,13 +335,14 @@ err-file:/path/to/file                             write the errors to a specifi
 
 none                                               ignore stream of events output, usually used with --capture
 
-option:{stack-addresses,detect-syscall,exec-env,relative-time,exec-info}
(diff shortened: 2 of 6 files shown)


yanivagman (commented):
I think there are two places missing the tests for RawArguments:
`if event.Probes[0].attach == sysCall {`
and
`if readFileId, isInt32 := args["type"].(int32); isInt32 {`

other than that LGTM

grantseltzer:
Woops, added those. 

@itaysk Do you feel strongly about allowing different parsing behavior for table vs json/gob? We decided if someone is building a pipeline with tracee-ebpf then they would want to opt in to raw arguments.

itaysk:
thanks for checking with me @grantseltzer . can you further explain why opt-in vs out as originally planned, and why not set it as default behavior pipeline?

yanivagman:
Most users will probably want to have "human readable" output by default, so we agree that table printer should be opt-in.
About gob/json, two points:
1. It feels strange to have different behaviors for table vs gob/json - I think that all the printers should have a consistent behavior to avoid confusion and special handling for each printer.
2. Keeping this feature opt-in avoids breaking the current tracee API

tracee-ebpf/tracee/argprinters.go
```diff
@@ -57,47 +57,47 @@ func (t *Tracee) prepareArgs(ctx *context, args map[string]interface{}) error {
 	switch ctx.EventID {
 	case SysEnterEventID, SysExitEventID, CapCapableEventID, CommitCredsEventID, SecurityFileOpenEventID:
 		//show syscall name instead of id
-		if id, isInt32 := args["syscall"].(int32); isInt32 {
+		if id, isInt32 := args["syscall"].(int32); isInt32 && !t.config.Output.RawArguments {
```
yanivagman, line 60: Instead of adding this `!t.config.Output.RawArguments` for almost every case, I would group all these cases together, and use one `if` check to wrap them all.

This also makes me think about the two places where you didn't add this check:
1. Printing alerts
2. Socket address

Maybe we shuold handle these in a diferent place? e.g. event_decode?

grantseltzer: Should alerts be disabled if raw is specified? I think it would be the raw struct if not parsed, which I don't think is any more machine readable.

yanivagman: No, alerts should be kept as strings. There is no meaning (outside of tracee code) to the alerts enum

grantseltzer: So what would the check do relating to printing alerts?

itaysk:
I talked with Yaniv offline and we agreed to do it like described in the original issue, meaning raw output by default, controllable with a flag, and for table it'll be human readable.

itaysk:
another thing that needs to be in this PR is to turn OFF the raw output in tracee container, so that existing signatures don't breat

itaysk:
and for a future PR, we will need additional helpers for signatures to be able to work with the bitmask raw values (this should go in libbpfgo)

tracee-ebpf/tracee/argprinters.go
yanivagman: `So what would the check do relating to printing alerts?`

Not sure I understand what you ask here.
What I am saying is, that instead of checking `!t.config.Output.RawArguments` many times, just do it one time, and place all the relevant arguments (not alerts or socket address arguments) under this check.

grantseltzer: Ah - I was confused by your comment at the top of this thread. Makes sense now.

grantseltzer: I made changes where args were parsed in `readArgFromBuf()` so that only one check would have to happen for `t.config.Output.ParseArguments` but honestly I like the way it is in now better. It's clearer in the code to have all the argument types for each EventID checked in one place.

yanivagman: Wait, changes in `readArgFromBuf()`?

What I had in mind is something like the following:
switch ctx.EventID {
    case MemProtAlertEventID:
    ....
    // all the cases which are not changed come here (alerts and socket arguments)
}

if !t.config.Output.ParseArguments {
    return nil
}

switch ctx.EventID {
    case SysEnterEventID, SysExitEventID, CapCapableEventID, CommitCredsEventID, SecurityFileOpenEventID:
    ....
    // all other cases (which you changed in this commit) comes here, without any change

grantseltzer: I thought by `event_decode` you were referring to `events_decoder.go`, regardless miscommunication and I agree with your idea here. Just pushed these changes.

yanivagman: > I thought by event_decode you were referring to events_decoder.go

I was :-) but I also wrote:

> This also makes me think about...

So it was unrelated to this suggested change, just saying that we should consider handling alert and socket "printing" as part of the events decoder

tracee-ebpf/tracee/argprinters.go
```diff
@@ -54,6 +54,48 @@ func (t *Tracee) prepareArgs(ctx *context, args map[string]interface{}) error {
...
+			}
+			s = strings.TrimSuffix(s, ",")
+			s = fmt.Sprintf("{%s}", s)
+			args["remote_addr"] = s
+		}
+	case MemProtAlertEventID:
+		if alert, isAlert := args["alert"].(alert); isAlert {
+			args["alert"] = PrintAlert(alert)
+		}
+	}
```
yanivagman, line 93: As a future PR, we should think about taking this code to be part of the socket/alert decoding

docs/tracee-ebpf/output.md
```diff
@@ -10,12 +10,14 @@ CLI Option | Description
 `none` | ignore stream of events output, usually used with `--capture`
 `out-file:/path/to/file` | write the output to a specified file. the path to the file will be created if not existing and the file will be deleted if existing (default: stdout)
 `err-file:/path/to/file` | write the errors to a specified file. the path to the file will be created if not existing and the file will be deleted if existing (default: stderr)
-`option:{stack-addresses,detect-syscall,exec-env,exec-info,relative-time}` | augment output according to given options (default: none)
+`option:{stack-addresses,detect-syscall,exec-env,exec-info,relative-time,raw-arguments}` | augment output according to given options (default: none)
```
yanivagman, line 13: This should be changed to parse-arguments

docs/tracee-ebpf/output.md
```diff
@@ -10,12 +10,14 @@ CLI Option | Description
...
   stack-addresses | include stack memory addresses for each event
   detect-syscall | when tracing kernel functions which are not syscalls, detect and show the original syscall that called that function
   exec-env | when tracing execve/execveat, show the environment variables that were used for execution
 exec-info | when tracing sched_process_exec, show the file hash(sha256) and ctime
 relative-time | use relative timestamp instead of wall timestamp for events
+raw-arguments | do not show raw machine-readable values for event arguments, instead parse into human readable strings
```
yanivagman, line 19: This should be changed to parse-arguments

tracee-ebpf/main.go
```diff
@@ -404,6 +407,11 @@ func prepareOutput(outputSlice []string, containerMode bool) (tracee.OutputConfi
 		}
 	}
 
+	if printerKind == "table" ||
+		true { // XXX: Remove this line once tracee-rules signatures have been updated to take raw args by default into account (see #1123, libbpfgo#88)
```
yanivagman, line 411: With this line, there is no way to have raw arguments set (ParseArguments will always be true). Am I wrong?
I think this line should be removed, and the entrypoint should be changed (for now) to have parseArguments option set

grantseltzer: You're right - I've updated entrypoint.sh which makes a lot more sense.

yanivagman (approved):
LGTM

Issue #925, opened 2021-08-16
signature tests lack tests names


AlonZivony:
In the current `signaturetest.SigTest` struct there is no name member, which is quite odd because in a case of failure there is nothing to print to explain which sub-test failed.
In fact, it seems that currently tests just print all the `SigTest` struct, which is quite ugly way to show which sub-test failed.

PR #1128, opened 2021-11-08
refactor: tests for Go signatures

danielpacak:
The signaturetest.SigTest struct doesn't support meaningful assertions
because it has only boolean properties to indicate findings.

Use signaturetest.FindingHolder instead that keeps all findings
in memory and along with helper methods allows flexible
assertions (with testify).

Resolves: #925

### DIFF
--- a/tracee-rules/signatures/golang/kubernetes_api_connection_test.go
+++ b/tracee-rules/signatures/golang/kubernetes_api_connection_test.go
@@ -1,16 +1,23 @@
 package main
 
 import (
-	tracee "github.com/aquasecurity/tracee/tracee-ebpf/external"
 	"testing"
 
+	tracee "github.com/aquasecurity/tracee/tracee-ebpf/external"
 	"github.com/aquasecurity/tracee/tracee-rules/signatures/signaturestest"
 	"github.com/aquasecurity/tracee/tracee-rules/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestK8sApiConnection(t *testing.T) {
-	SigTests := []signaturestest.SigTest{
+	testCases := []struct {
+		Name     string
+		Events   []types.Event
+		Findings map[string]types.Finding
+	}{
 		{
+			Name: "should trigger detection",
 			Events: []types.Event{
 				tracee.Event{
 					EventName:   "execve",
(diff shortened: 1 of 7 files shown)


tracee-rules/signatures/golang/kubernetes_api_connection_test.go
```diff
@@ -38,14 +45,52 @@ func TestK8sApiConnection(t *testing.T) {
...
+							Value: map[string]string{
+								"sa_family": "AF_INET",
+								"sin_port":  "80",
+								"sin_addr":  "1.1.1.1",
+							},
+						},
+					},
+				},
+			},
+			Findings: map[string]types.Finding{
```
danielpacak, line 57: That's the main thing: Instead of asserting on true or false we assert on the whole Finding object, which includes event and data. Otherwise we're not catching regressions in the signature logic.

tracee-rules/signatures/signaturestest/signaturestest.go
```diff
@@ -1,32 +1,35 @@
...
+	"github.com/aquasecurity/tracee/tracee-rules/types"
+)
 
-// SigTest is a utility to test signatures
-type SigTest struct {
-	// Events are a series of events that the signature will be tested with (in the provided order)
-	Events []types.Event
-	// CB provide a callback for this test case only if you want to check the Finding result
-	// if providing your own callback, make sure to call SetStatus from it
-	CB types.SignatureHandler
```
danielpacak, line 11: Tests based on callbacks are kind of fragile. It's easy to forget to invoke the callback which might give false positive / negative test. It's simpler to assert to data (= Go structures) not functions.

tracee-rules/signatures/signaturestest/signaturestest.go
```diff
@@ -1,32 +1,35 @@
...
+	}
+	return r
+}
+
+func (h *FindingsHolder) FirstValue() *types.Finding {
+	if len(h.Values) == 0 {
+		return nil
 	}
+	return &h.Values[0]
 }
```
simar7, line 31: Do we need this?

tracee-rules/signatures/signaturestest/signaturestest.go
```diff
@@ -1,32 +1,35 @@
...
 
-func (st *SigTest) SetStatus(res types.Finding) {
-	st.Status = true
+func (h *FindingsHolder) Empty() bool {
+	return len(h.Values) == 0
```
simar7, line 34: ditto

simar7 (approved):
overall lgtm, 2 minor comments

tracee-rules/signatures/signaturestest/signaturestest.go
danielpacak: Actually the Empty method is not used anywhere so I'm removing it.

tracee-rules/signatures/signaturestest/signaturestest.go
danielpacak: We must check the length to avoid `panic: runtime error: index out of range [0] with length 0` in test cases where Values is not initialized

PR #1130, opened 2021-11-09
tracee-ebpf: use cgroup id for container id resolution

### LINKED ISSUE #473: Use cgroup id for containers

### LINKED ISSUE #958: container_id initialization regex has to be fixed

yanivagman:
This PR changes the way we extract container id for events.

The current solution has several problems:
1. Container id is saved per process, wasting memory.
2. Container id is only 12 chars long, while the full container id (64 chars long) might be required.
3. Bpf code is responsible for extracting container id from a cgroup name (we now check, for example, if cgroup name has "docker-" prefix, but this doesn't always work, e.g. with podman)
4. Regexes used during container id map initialization don't match the container ids we extract in runtime (as bpf code can't handle regexes)

To solve these problems, let's move to use cgroup id to get a container id:
1. In the context of every event, send task's cgroup id (instead of 12 chars container id)
2. Add a new map in userspace that maps cgroup id to container id
3. Update this map on init with existing containers by knowing that the lower 32 bits of the cgroup_id are actually the inode number in the cgroupfs entry
4. Add two new events (tracepoints) to track cgroup creation/removal: cgroup_mkdir and cgroup_rmdir
5. Using these two events, update cgroup to container id map in runtime if a match to a known container runtime is found
6. Use cgroup_id as an index to the map to get the container id of a given event

close #473
fix #958

### DIFF
--- a/tracee-ebpf/tracee/tracee.bpf.c
+++ b/tracee-ebpf/tracee/tracee.bpf.c
@@ -254,6 +254,7 @@ struct kernfs_node___old {
 #define CONFIG_DEBUG_NET            17
 #define CONFIG_PROC_TREE_FILTER     18
 #define CONFIG_CAPTURE_MODULES      19
+#define CONFIG_CGROUP_V1            20
 
 // get_config(CONFIG_XXX_FILTER) returns 0 if not enabled
 #define FILTER_IN  1
@@ -777,6 +778,14 @@ static __always_inline const u64 get_cgroup_id(struct cgroup *cgrp)
     return id;
 }
 
+static __always_inline const u64 get_cgroup_v1_subsys0_id(struct task_struct *task)
+{
+    struct css_set *cgroups = READ_KERN(task->cgroups);
+    struct cgroup_subsys_state *subsys0 = READ_KERN(cgroups->subsys[0]);
+    struct cgroup *cgroup = READ_KERN(subsys0->cgroup);
+    return get_cgroup_id(cgroup);
+}
+
 static __always_inline bool is_x86_compat(struct task_struct *task)
 {
 #if defined(bpf_target_x86)
@@ -1041,6 +1050,26 @@ static __inline int has_prefix(char *prefix, char *str, int n)
     return 0;
 }
 
+static __always_inline int get_config(u32 key)
+{
+    u32 *config = bpf_map_lookup_elem(&config_map, &key);
+
+    if (config == NULL)
+        return 0;
+
+    return *config;
+}
+
+static __always_inline int get_kconfig_val(u32 key)
+{
+    u32 *config = bpf_map_lookup_elem(&kconfig_map, &key);
+
+    if (config == NULL)
+        return 0;
+
+    return *config;
+}
+
 static __always_inline int init_context(context_t *context, struct task_struct *task)
 {
     u64 id = bpf_get_current_pid_tgid();
@@ -1057,7 +1086,12 @@ static __always_inline int init_context(context_t *context, struct task_struct *
     char * uts_name = get_task_uts_name(task);
     if (uts_name)
         bpf_probe_read_str(&context->uts_name, TASK_COMM_LEN, uts_name);
-    context->cgroup_id = bpf_get_current_cgroup_id();
+    if (get_config(CONFIG_CGROUP_V1)) {
+        context->cgroup_id = get_cgroup_v1_subsys0_id(task);
+    } else {
+        context->cgroup_id = bpf_get_current_cgroup_id();
+    }
+
     context->ts = bpf_ktime_get_ns();
     context->argnum = 0;
 
@@ -1079,26 +1113,6 @@ static __always_inline int init_event_data(event_data_t *data, void *ctx)
         return 0;
 
     return 1;
-}
-
-static __always_inline int get_config(u32 key)
-{
-    u32 *config = bpf_map_lookup_elem(&config_map, &key);
-
-    if (config == NULL)
-        return 0;
-
-    return *config;
-}
-
-static __always_inline int get_kconfig_val(u32 key)
-{
-    u32 *config = bpf_map_lookup_elem(&kconfig_map, &key);
-
-    if (config == NULL)
-        return 0;
-
-    return *config;
 }
 
 // returns 1 if you should trace based on uid, 0 if not
--- a/tracee-ebpf/tracee/tracee.bpf.c
+++ b/tracee-ebpf/tracee/tracee.bpf.c
@@ -66,6 +66,20 @@ Copyright (C) Aqua Security inc.
 //CO:RE is enabled
 #include <vmlinux.h>
 #include "co_re_missing_definitions.h"
+
+union kernfs_node_id {
+    struct {
+        u32     ino;
+        u32     generation;
+    };
+    u64         id;
+} __attribute__((preserve_access_index));
+
+struct kernfs_node___old {
+    const char            *name;
+    union kernfs_node_id  id;
+} __attribute__((preserve_access_index));
+
 #endif
 
 #undef container_of
@@ -745,7 +759,20 @@ static __always_inline const u64 get_cgroup_id(struct cgroup *cgrp)
     // id was not always of type u64 (but "union kernfs_node_id" before 5.5)
     // yet we can read it as u64 as the size of both types is the same
     u64 id;
-    bpf_probe_read(&id, sizeof(u64), GET_FIELD_ADDR(kn->id));
+
+#ifdef CORE
+    if (bpf_core_type_exists(union kernfs_node_id)) {
+        // recast pointer to capture kernfs_node___old type for compiler
+        struct kernfs_node___old *kn_old = (void *)kn;
+
+        // now use old "id", which is of type "union kernfs_node_id"
+        bpf_core_read(&id, sizeof(u64), &kn_old->id);
+    } else {
+        bpf_core_read(&id, sizeof(u64), &kn->id);
+    }
+#else
+    bpf_probe_read(&id, sizeof(u64), &kn->id);
+#endif
 
     return id;
 }
--- a/tracee-ebpf/tracee/tracee.bpf.c
+++ b/tracee-ebpf/tracee/tracee.bpf.c
@@ -191,22 +191,24 @@ Copyright (C) Aqua Security inc.
 #define SWITCH_TASK_NS              1012
 #define MAGIC_WRITE                 1013
 #define CGROUP_ATTACH_TASK          1014
-#define SECURITY_BPRM_CHECK         1015
-#define SECURITY_FILE_OPEN          1016
-#define SECURITY_INODE_UNLINK       1017
-#define SECURITY_SOCKET_CREATE      1018
-#define SECURITY_SOCKET_LISTEN      1019
-#define SECURITY_SOCKET_CONNECT     1020
-#define SECURITY_SOCKET_ACCEPT      1021
-#define SECURITY_SOCKET_BIND        1022
-#define SECURITY_SB_MOUNT           1023
-#define SECURITY_BPF                1024
-#define SECURITY_BPF_MAP            1025
-#define SECURITY_KERNEL_READ_FILE   1026
-#define SECURITY_INODE_MKNOD        1027
-#define SECURITY_POST_READ_FILE     1028
-#define SOCKET_DUP                  1029
-#define MAX_EVENT_ID                1030
+#define CGROUP_MKDIR                1015
+#define CGROUP_RMDIR                1016
+#define SECURITY_BPRM_CHECK         1017
+#define SECURITY_FILE_OPEN          1018
+#define SECURITY_INODE_UNLINK       1019
+#define SECURITY_SOCKET_CREATE      1020
+#define SECURITY_SOCKET_LISTEN      1021
+#define SECURITY_SOCKET_CONNECT     1022
+#define SECURITY_SOCKET_ACCEPT      1023
+#define SECURITY_SOCKET_BIND        1024
+#define SECURITY_SB_MOUNT           1025
+#define SECURITY_BPF                1026
+#define SECURITY_BPF_MAP            1027
+#define SECURITY_KERNEL_READ_FILE   1028
+#define SECURITY_INODE_MKNOD        1029
+#define SECURITY_POST_READ_FILE     1030
+#define SOCKET_DUP                  1031
+#define MAX_EVENT_ID                1032
 
 #define FILE_TYPE_SOCK      0
 
@@ -367,6 +369,7 @@ struct bpf_map_def SEC("maps") _name = { \
 
 typedef struct event_context {
     u64 ts;                     // Timestamp
+    u64 cgroup_id;
     u32 pid;                    // PID as in the userspace term
     u32 tid;                    // TID as in the userspace term
     u32 ppid;                   // Parent PID as in the userspace term
@@ -378,7 +381,6 @@ typedef struct event_context {
     u32 pid_id;
     char comm[TASK_COMM_LEN];
     char uts_name[TASK_COMM_LEN];
-    char cont_id[16];           // Container ID, padding to 16 to keep the context struct aligned
     u32 eventid;
     s64 retval;
     u32 stack_id;
@@ -426,10 +428,6 @@ typedef struct event_data {
     buf_t *submit_p;
     u32 buf_off;
 } event_data_t;
-
-typedef struct container_id {
-    char id[CONT_ID_LEN+1];
-} container_id_t;
 
 // For a good summary about capabilities, see https://lwn.net/Articles/636533/
 typedef struct slim_cred {
@@ -533,7 +531,6 @@ BPF_HASH(chosen_events_map, u32, u32);                  // Events chosen by the 
 BPF_HASH(traced_pids_map, u32, u32);                    // Keep track of traced pids
 BPF_HASH(new_pids_map, u32, u32);                       // Keep track of the processes of newly executed binaries
 BPF_HASH(new_pidns_map, u32, u32);                      // Keep track of new pid namespaces
-BPF_HASH(pid_to_cont_id_map, u32, container_id_t);      // Map pid to container id
 BPF_HASH(args_map, u64, args_t);                        // Persist args info between function entry and return
 BPF_HASH(syscall_data_map, u32, syscall_data_t);        // Persist data during syscall execution
 BPF_HASH(inequality_filter, u32, u64);                  // Used to filter events by some uint field either by < or >
@@ -736,6 +733,21 @@ static __always_inline const char * get_cgroup_dirname(struct cgroup *cgrp)
         return NULL;
 
     return READ_KERN(kn->name);
+}
+
+static __always_inline const u64 get_cgroup_id(struct cgroup *cgrp)
+{
+    struct kernfs_node *kn = READ_KERN(cgrp->kn);
+
+    if (kn == NULL)
+        return 0;
+
+    // id was not always of type u64 (but "union kernfs_node_id" before 5.5)
+    // yet we can read it as u64 as the size of both types is the same
+    u64 id;
+    bpf_probe_read(&id, sizeof(u64), GET_FIELD_ADDR(kn->id));
+
+    return id;
 }
 
 static __always_inline bool is_x86_compat(struct task_struct *task)
@@ -1018,10 +1030,7 @@ static __always_inline int init_context(context_t *context, struct task_struct *
     char * uts_name = get_task_uts_name(task);
     if (uts_name)
         bpf_probe_read_str(&context->uts_name, TASK_COMM_LEN, uts_name);
-    container_id_t *container_id = bpf_map_lookup_elem(&pid_to_cont_id_map, &context->host_tid);
-    if (container_id != NULL) {
-        __builtin_memcpy(context->cont_id, container_id->id, CONT_ID_LEN);
-    }
+    context->cgroup_id = bpf_get_current_cgroup_id();
     context->ts = bpf_ktime_get_ns();
     context->argnum = 0;
 
@@ -2295,6 +2304,10 @@ int sys_dup_exit_tail(void *ctx)
 SEC("raw_tracepoint/sched_process_fork")
 int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
 {
+    event_data_t data = {};
+    if (!init_event_data(&data, ctx))
+        return 0;
+
     // Note: we don't place should_trace() here, so we can keep track of the cgroups in the system
     struct task_struct *parent = (struct task_struct*)ctx->args[0];
     struct task_struct *child = (struct task_struct*)ctx->args[1];
@@ -2305,12 +2318,6 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
     int parent_tgid = get_task_host_tgid(parent);
     int child_tgid = get_task_host_tgid(child);
 
-    container_id_t *container_id = bpf_map_lookup_elem(&pid_to_cont_id_map, &parent_pid);
-    if (container_id != NULL) {
-        // copy the container id of the parent process to the child process
-        bpf_map_update_elem(&pid_to_cont_id_map, &child_pid, &container_id->id, BPF_ANY);
-    }
-
     // update process tree map if the parent has an entry
     int proc_tree_filter_set = get_config(CONFIG_PROC_TREE_FILTER);
     if (proc_tree_filter_set) {
@@ -2319,10 +2326,6 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
             bpf_map_update_elem(&process_tree_map, &child_tgid, tgid_filtered, BPF_ANY);
         }
     }
-
-    event_data_t data = {};
-    if (!init_event_data(&data, ctx))
-        return 0;
 
     if (!should_trace(&data.context))
         return 0;
@@ -2433,7 +2436,6 @@ int tracepoint__sched__sched_process_exit(struct bpf_raw_tracepoint_args *ctx)
         return 0;
 
     // Remove this pid from all maps
-    bpf_map_delete_elem(&pid_to_cont_id_map, &data.context.host_tid);
     bpf_map_delete_elem(&traced_pids_map, &data.context.host_tid);
     bpf_map_delete_elem(&new_pids_map, &data.context.host_tid);
     bpf_map_delete_elem(&syscall_data_map, &data.context.host_tid);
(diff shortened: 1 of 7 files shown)


rafaeldtinoco:
Good thing about having BTFHub embedded files now is that after rebasing with my dev branch Im able to quickly test it in multiple kernels... with that said, it looks like this code has issues loading in 5.4 kernel:

```
; struct cgroup *dst_cgrp = (struct cgroup*)ctx->args[0];
308: (79) r3 = *(u64 *)(r7 +0)
309: (b7) r1 = 288
310: (0f) r3 += r1
; char *path = (char*)ctx->args[1];
311: (79) r7 = *(u64 *)(r7 +8)
; struct kernfs_node *kn = READ_KERN(cgrp->kn);
312: (7b) *(u64 *)(r10 -72) = r6
313: (bf) r1 = r10
; struct cgroup *dst_cgrp = (struct cgroup*)ctx->args[0];
314: (07) r1 += -72
; struct kernfs_node *kn = READ_KERN(cgrp->kn);
315: (b7) r2 = 8
316: (85) call bpf_probe_read#4
last_idx 316 first_idx 294
regs=4 stack=0 before 315: (b7) r2 = 8
317: (79) r3 = *(u64 *)(r10 -72)
; if (kn == NULL)
318: (15) if r3 == 0x0 goto pc+7
 R0=inv(id=0) R3_w=inv(id=0) R6=invP0 R7=inv(id=0) R8=inv(id=0) R9=inv2344 R10=fp0 fp-8=???????m fp-16=mmmmmmmm fp-24=mmmmmmmm fp-32=mmmmmmmm fp-40=mmmmmmmm fp-48=mmmmmmmm fp-56=mmmmmmmm fp-64=mmmmmmmm fp-72=mmmmmmmm fp-80=0000mmmm fp-88=map_value fp-96=ctx fp-104=00000000 fp-112=00000000 fp-120=0000mmmm fp-128=mmmmmmmm fp-136=mmmmmmmm fp-144=mmmmmmmm fp-152=mmmmmmmm fp-160=mmmmmmmm fp-168=mmmmmmmm fp-176=mmmmmmmm fp-184=mmmmmmmm fp-192=mmmmmmmm fp-200=mmmmmmmm fp-208=mmmmmmmm fp-216=ctx fp-224=mmmmmmmm fp-232=mmmmmmmm
319: (85) call unknown#195896080
invalid func unknown#195896080
processed 319 insns (limit 1000000) max_states_per_insn 0 total_states 18 peak_states 18 mark_read 16

libbpf: -- END LOG --
libbpf: failed to load program 'tracepoint__cgroup__cgroup_mkdir'
2021/11/10 02:20:52 error creating Tracee: failed to load BPF object
```

It works fine in kernel 5.8.

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -732,6 +729,21 @@ static __always_inline const char * get_cgroup_dirname(struct cgroup *cgrp)
...
+{
+    struct kernfs_node *kn = READ_KERN(cgrp->kn);
+
+    if (kn == NULL)
+        return 0;
+
+    // id was not always of type u64 (but "union kernfs_node_id" before 5.5)
+    // yet we can read it as u64 as the size of both types is the same
+    u64 id;
+    bpf_probe_read(&id, sizeof(u64), GET_FIELD_ADDR(kn->id));
```
rafaeldtinoco, line 742: I think you will have to add type flavor for v5.4.y here so the field relocation is correctly calculated on these kernels.

libbpf was unable to relocate this when loading the object in an v5.4 kernel, it poisoned the instruction so the eBPF verifier would make it fail:

```
; if (kn == NULL)
318: (15) if r3 == 0x0 goto pc+7
 R0=inv(id=0) R3_w=inv(id=0) R6=invP0 R7=inv(id=0) R8=inv(id=0) R9=inv2344 R10=fp0 fp-8=???????m fp-16=mmmmmmmm fp-24=mmmmmmmm fp-32=mmmmmmmm fp-40=mmmmmmmm fp-48=mmmmmmmm fp-56=mmmmmmmm fp-64=mmmmmmmm fp-72=mmmmmmmm fp-80=0000mmmm fp-88=map_value fp-96=ctx fp-104=00000000 fp-112=00000000 fp-120=0000mmmm fp-128=mmmmmmmm fp-136=mmmmmmmm fp-144=mmmmmmmm fp-152=mmmmmmmm fp-160=mmmmmmmm fp-168=mmmmmmmm fp-176=mmmmmmmm fp-184=mmmmmmmm fp-192=mmmmmmmm fp-200=mmmmmmmm fp-208=mmmmmmmm fp-216=ctx fp-224=mmmmmmmm fp-232=mmmmmmmm
319: (85) call unknown#195896080
invalid func unknown#195896080
processed 319 insns (limit 1000000) max_states_per_insn 0 total_states 18 peak_states 18 mark_read 16
```

rafaeldtinoco (commented):
The overall change is VERY nice. I liked the approach of having cgroup_id <-> container_id relationship maintained by the cgroup events. After you deal with the 5.4 relocation error, I'm ready to +1.

tracee-ebpf/tracee/tracee.bpf.c
rafaeldtinoco: The 5.4 kernel has:

```
[1415] STRUCT 'kernfs_node' size=128 vlen=13
        'count' type_id=79 bits_offset=0
        'active' type_id=79 bits_offset=32
        'parent' type_id=1416 bits_offset=64
        'name' type_id=5 bits_offset=128
        'rb' type_id=559 bits_offset=192
        'ns' type_id=253 bits_offset=384
        'hash' type_id=9 bits_offset=448
        '(anon)' type_id=1424 bits_offset=512
        'priv' type_id=103 bits_offset=768
        'id' type_id=1423 bits_offset=832
        'flags' type_id=19 bits_offset=896
        'mode' type_id=59 bits_offset=912
        'iattr' type_id=1425 bits_offset=960

[1423] UNION 'kernfs_node_id' size=8 vlen=2
        '(anon)' type_id=1422 bits_offset=0
        'id' type_id=36 bits_offset=0
```

Latest kernels have:

```
[1418] STRUCT 'kernfs_node' size=128 vlen=13
        'count' type_id=71 bits_offset=0
        'active' type_id=71 bits_offset=32
        'parent' type_id=1419 bits_offset=64
        'name' type_id=4 bits_offset=128
        'rb' type_id=354 bits_offset=192
        'ns' type_id=86 bits_offset=384
        'hash' type_id=7 bits_offset=448
        '(anon)' type_id=1425 bits_offset=512
        'priv' type_id=85 bits_offset=768
        'id' type_id=30 bits_offset=832
        'flags' type_id=16 bits_offset=896
        'mode' type_id=53 bits_offset=912
        'iattr' type_id=1426 bits_offset=960

[30] TYPEDEF 'u64' type_id=22
```

I think what happened is that libbpf was unable to do this relocation because it had no other "flavor" of "struct kernfs_node" (containing the union kernfs_node_id) to work with (and generate the correct RELO code within .BTF.ext ELF section). WDYT ?

rafaeldtinoco: BTW, for this new struct flavor you will likely need: https://github.com/aquasecurity/tracee/issues/1133 work to be done. I'm working on that now.

yanivagman: @rafaeldtinoco can you have a look at https://github.com/aquasecurity/tracee/pull/1130/commits/10da2387c8065154baa5b18327d732feaaf6fcb9?
Hopefully, this will solve the problem for older kernels with BTF.

The problem is that I'm not sure btfgen knows how to handle type checks like I had to do here...

rafaeldtinoco: Still working on the verifier thing, but Id like to clarify this finding... 

kernel 5.13 ID value for a single container being started and stopped:
```
$ sudo bpftool prog tracelog
         systemd-1       [000] d... 630056.825108: bpf_trace_printk: BIGGER ID = 31687
         systemd-1       [007] d... 630060.909689: bpf_trace_printk: BIGGER ID = 31687
```

kernels 5.4 ID value for a single container being started:
```
$ sudo cat /sys/kernel/debug/tracing/trace_pipe
            runc-9391    [000] d... 292082.860162: 0: SMALLER ID = 4294967462
            runc-9391    [000] d... 292082.861525: 0: SMALLER ID = 4294967883
            runc-9391    [000] d... 292082.861934: 0: SMALLER ID = 4294969343
            runc-9391    [000] d... 292082.862313: 0: SMALLER ID = 4294968684
            runc-9391    [000] d... 292082.862822: 0: SMALLER ID = 4294968008
            runc-9391    [000] d... 292082.863143: 0: SMALLER ID = 4294968321
            runc-9391    [000] d... 292082.863479: 0: SMALLER ID = 4294967416
            runc-9391    [000] d... 292082.863840: 0: SMALLER ID = 4294967371
            runc-9391    [000] d... 292082.864355: 0: SMALLER ID = 4294967344
            runc-9391    [000] d... 292082.865515: 0: SMALLER ID = 4294967368
            runc-9391    [000] d... 292082.865784: 0: SMALLER ID = 4294967774
            runc-9466    [003] d... 292090.195132: 0: SMALLER ID = 4294968684
            runc-9466    [003] d... 292090.195269: 0: SMALLER ID = 4294967368
            runc-9466    [003] d... 292090.195348: 0: SMALLER ID = 4294967416
            runc-9466    [003] d... 292090.195419: 0: SMALLER ID = 4294967371
            runc-9466    [003] d... 292090.195489: 0: SMALLER ID = 4294968321
            runc-9466    [003] d... 292090.195590: 0: SMALLER ID = 4294967462
            runc-9466    [003] d... 292090.195663: 0: SMALLER ID = 4294967883
            runc-9466    [003] d... 292090.195749: 0: SMALLER ID = 4294969343
            runc-9466    [003] d... 292090.195826: 0: SMALLER ID = 4294967774
            runc-9466    [003] d... 292090.195942: 0: SMALLER ID = 4294967344
            runc-9466    [003] d... 292090.196026: 0: SMALLER ID = 4294968008
```

Are you sure that cgroup->kernfs_node_id->id has the same meaning in 5.4 than it has in kernels above ?

rafaeldtinoco: I think the logic is broken for kernels <= 5.4:

```
# kernel 5.13 and docker in impish

$ sudo bpftool prog tracelog
systemd-1       [007] d... 630955.196985: bpf_trace_printk: BIGGER ID=31844 (docker-f19788cebc98e4d5d88a945f80e45f0ecd7652f393ec24b7c25570c1c5ed3d76.scope)

$ sudo find /sys/fs/cgroup -type d -name '*f19788cebc98e4d5d88a945f80e45f0ecd7652f393ec24b7c25570c1c5ed3d76*'
/sys/fs/cgroup/system.slice/docker-f19788cebc98e4d5d88a945f80e45f0ecd7652f393ec24b7c25570c1c5ed3d76.scope

# kernel 5.4 and docker in bionic

$ sudo cat /sys/kernel/debug/tracing/trace_pipe
runc-10231   [001] d... 292996.043605: 0: SMALLER ID=4294967516 (82ef3a4e473f1b9b9d5be508a8a27988e78f8fde67c7bad16b5aadd27069432)
runc-10231   [001] d... 292996.045532: 0: SMALLER ID=4294967907 (82ef3a4e473f1b9b9d5be508a8a27988e78f8fde67c7bad16b5aadd27069432)
runc-10231   [001] d... 292996.046145: 0: SMALLER ID=4294969427 (82ef3a4e473f1b9b9d5be508a8a27988e78f8fde67c7bad16b5aadd27069432)
runc-10231   [001] d... 292996.046706: 0: SMALLER ID=4294968741 (82ef3a4e473f1b9b9d5be508a8a27988e78f8fde67c7bad16b5aadd27069432)
runc-10231   [001] d... 292996.047514: 0: SMALLER ID=4294968048 (82ef3a4e473f1b9b9d5be508a8a27988e78f8fde67c7bad16b5aadd27069432)
runc-10231   [001] d... 292996.047977: 0: SMALLER ID=4294968363 (82ef3a4e473f1b9b9d5be508a8a27988e78f8fde67c7bad16b5aadd27069432)
runc-10231   [001] d... 292996.048468: 0: SMALLER ID=4294967455 (82ef3a4e473f1b9b9d5be508a8a27988e78f8fde67c7bad16b5aadd27069432)
runc-10231   [001] d... 292996.049120: 0: SMALLER ID=4294967395 (82ef3a4e473f1b9b9d5be508a8a27988e78f8fde67c7bad16b5aadd27069432)
runc-10231   [001] d... 292996.050122: 0: SMALLER ID=4294967359 (82ef3a4e473f1b9b9d5be508a8a27988e78f8fde67c7bad16b5aadd27069432)
runc-10231   [001] d... 292996.050695: 0: SMALLER ID=4294967392 (82ef3a4e473f1b9b9d5be508a8a27988e78f8fde67c7bad16b5aadd27069432)
runc-10231   [001] d... 292996.051103: 0: SMALLER ID=4294967799 (82ef3a4e473f1b9b9d5be508a8a27988e78f8fde67c7bad16b5aadd27069432)

$ sudo find /sys/fs/cgroup -type d -name '*82ef3a4e473f1b9b9d5be508a8a27988e78f8fde67c7bad16b5aadd27069432*'
/sys/fs/cgroup/hugetlb/docker/82ef3a4e473f1b9b9d5be508a8a27988e78f8fde67c7bad16b5aadd27069432a
/sys/fs/cgroup/cpuset/docker/82ef3a4e473f1b9b9d5be508a8a27988e78f8fde67c7bad16b5aadd27069432a
/sys/fs/cgroup/memory/docker/82ef3a4e473f1b9b9d5be508a8a27988e78f8fde67c7bad16b5aadd27069432a
/sys/fs/cgroup/pids/docker/82ef3a4e473f1b9b9d5be508a8a27988e78f8fde67c7bad16b5aadd27069432a
/sys/fs/cgroup/blkio/docker/82ef3a4e473f1b9b9d5be508a8a27988e78f8fde67c7bad16b5aadd27069432a
/sys/fs/cgroup/cpu,cpuacct/docker/82ef3a4e473f1b9b9d5be508a8a27988e78f8fde67c7bad16b5aadd27069432a
/sys/fs/cgroup/devices/docker/82ef3a4e473f1b9b9d5be508a8a27988e78f8fde67c7bad16b5aadd27069432a
/sys/fs/cgroup/net_cls,net_prio/docker/82ef3a4e473f1b9b9d5be508a8a27988e78f8fde67c7bad16b5aadd27069432a
/sys/fs/cgroup/perf_event/docker/82ef3a4e473f1b9b9d5be508a8a27988e78f8fde67c7bad16b5aadd27069432a
/sys/fs/cgroup/freezer/docker/82ef3a4e473f1b9b9d5be508a8a27988e78f8fde67c7bad16b5aadd27069432a
/sys/fs/cgroup/systemd/docker/82ef3a4e473f1b9b9d5be508a8a27988e78f8fde67c7bad16b5aadd27069432a
```

yanivagman: I tested both on kernel 4.19 and kernel 5.10 - the logic should work.
Did you notice that I'm using the 32 LSB of the cgroup id and not the entire value?

rafaeldtinoco: Nevertheless,

I was NOT able to use kconf values accepted by the verifier. While your commit 7436832 works very well with full BTF files, any attempt of having something like:

```C
extern int LINUX_KERNEL_VERSION __kconfig;
...
#ifdef CORE
    //if (bpf_core_type_exists(union kernfs_node_id)) { // <= v5.0.0
    if (LINUX_KERNEL_VERSION < KERNEL_VERSION(5, 5, 0)) {
```

causes the infamous BADRELO error code for 5.4 kernels:

```
; bpf_core_read(&kn, sizeof(void *), (struct kernfs_node *) &cgrp->kn);
360: (b7) r2 = 8
361: (bf) r3 = r7
362: (85) call bpf_probe_read#4
last_idx 362 first_idx 316
regs=4 stack=0 before 361: (bf) r3 = r7
regs=4 stack=0 before 360: (b7) r2 = 8
363: (85) call unknown#195896080
invalid func unknown#195896080 <- BAD RELO code
```

rafaeldtinoco: If you keep this logic (linking cgroup ID with container_id), fixing, if needed, 5.4 scenario (if my assumption is right), then I think we should keep the type checks and work btfgen to do the relocations for us.

rafaeldtinoco: > I tested both on kernel 4.19 and kernel 5.10 - the logic should work. Did you notice that I'm using the 32 LSB of the cgroup id and not the entire value?

AH okay, it makes sense, you're right. I forgot about that. So all the IDs would have the same 32 LSB cgroup ID and it would work. Okay...

rafaeldtinoco: I have made btfgen to be capable of calculating the correct types in the tailored BTF file (with type based relocations feature). I'm "resolving" this conversation and starting a new one for something else.

rafaeldtinoco:
@yanivagman, Seems that I was able to add the type relocation feature in btfhub:

kernfs_node->id is an union in kernels <= 5.4:

```
$ sudo bpftool btf dump file ./5.4.0-87-generic.btf format raw  | less

[71] STRUCT 'kernfs_node' size=128 vlen=1
        'id' type_id=65 bits_offset=832

[65] UNION 'kernfs_node_id' size=8 vlen=2
        '(anon)' type_id=123 bits_offset=0
        'id' type_id=38 bits_offset=0

[123] STRUCT '(anon)' size=8 vlen=2
        'ino' type_id=1 bits_offset=0
        'generation' type_id=1 bits_offset=32

[1] TYPEDEF 'u32' type_id=30
[30] TYPEDEF '__u32' type_id=78
[78] INT 'unsigned int' size=4 bits_offset=0 nr_bits=32 encoding=(none)

[38] TYPEDEF 'u64' type_id=11
[11] TYPEDEF '__u64' type_id=98
[98] INT 'long long unsigned int' size=8 bits_offset=0 nr_bits=64 encoding=(none)
```

kernfs_node->id is an integer in kernels > 5.4:

```
$ sudo bpftool btf dump file ./5.13.0-20-generic.btf format raw | less

[59] STRUCT 'kernfs_node' size=128 vlen=1
        'id' type_id=82 bits_offset=832

[82] TYPEDEF 'u64' type_id=88
[88] TYPEDEF '__u64' type_id=34
[34] INT 'long long unsigned int' size=8 bits_offset=0 nr_bits=64 encoding=(none)
```

So let's consider that part "okay" (unless there is a bug in my code). Now, I would like to mention something else observed:

rafaeldtinoco:
could you try to use the FULL BTF file (from BTFHub) in a 5.4.0 kernel and see if you're getting the "container_id" value ?

```
$ sudo TRACEE_BTF_FILE=/home/rafaeldtinoco/work/sources/ebpf/btfgen/btfs/5.4.0-87-generic.btf ./dist/tracee-ebpf --debug --trace container -trace event=openat,openat2
OSInfo: VERSION: "18.04.6 LTS (Bionic Beaver)"
OSInfo: ID: ubuntu
OSInfo: ID_LIKE: debian
OSInfo: PRETTY_NAME: "Ubuntu 18.04.6 LTS"
OSInfo: VERSION_ID: "18.04"
OSInfo: VERSION_CODENAME: bionic
OSInfo: KERNEL_RELEASE: 5.4.0-87-generic
BTF: bpfenv = false, btfenv = true, vmlinux = false
BPF: using embedded BPF object
BTF: using BTF file from environment: /home/rafaeldtinoco/work/sources/ebpf/btfgen/btfs/5.4.0-87-generic.btf
unpacked CO:RE bpf object file into memory
TIME             CONTAINER_ID  UID    COMM             PID/host        TID/host        RET              EVENT                ARGS
07:26:09:565417                0      bash             1      /2019    1      /2019    3                openat               dirfd: -100, pathname: /etc/ld.so.cache, flags: O_RDONLY|O_CLOEXEC, mode: 0
07:26:09:565483                0      bash             1      /2019    1      /2019    3                openat               dirfd: -100, pathname: /lib/x86_64-linux-gnu/libtinfo.so.6, flags: O_RDONLY|O_CLOEXEC, mode: 0
07:26:09:565612                0      bash             1      /2019    1      /2019    3                openat               dirfd: -100, pathname: /lib/x86_64-linux-gnu/libc.so.6, flags: O_RDONLY|O_CLOEXEC, mode: 0
07:26:09:566305                0      bash             1      /2019    1      /2019    3                openat               dirfd: -100, pathname: /dev/tty, flags: O_RDWR|O_NONBLOCK, mode: 0
07:26:09:566897                0      bash             1      /2019    1      /2019    3                openat               dirfd: -100, pathname: /etc/nsswitch.conf, flags: O_RDONLY|O_CLOEXEC, mode: 0
07:26:09:567006                0      bash             1      /2019    1      /2019    3                openat               dirfd: -100, pathname: /etc/passwd, flags: O_RDONLY|O_CLOEXEC, mode: 0
07:26:09:567519                0      bash             1      /2019    1      /2019    3                openat               dirfd: -100, pathname: /etc/bash.bashrc, flags: O_RDONLY, mode: 0
07:26:09:569169                0      groups           9      /2079    9      /2079    3                openat               dirfd: -100, pathname: /etc/ld.so.cache, flags: O_RDONLY|O_CLOEXEC, mode: 0
```

rafaeldtinoco:
I'm able to get the container id if using the full BTF file for 5.13 kernel, for example:

```
$ sudo TRACEE_BTF_FILE=/home/rafaeldtinoco/work/sources/ebpf/btfgen/btfs/5.13.0-20-generic.btf ./dist/tracee-ebpf --debug --trace container -trace event=openat,openat2
OSInfo: KERNEL_RELEASE: 5.13.0-20-generic
OSInfo: PRETTY_NAME: "Ubuntu 21.10"
OSInfo: VERSION_ID: "21.10"
OSInfo: VERSION: "21.10 (Impish Indri)"
OSInfo: VERSION_CODENAME: impish
OSInfo: ID: ubuntu
OSInfo: ID_LIKE: debian
BTF: bpfenv = false, btfenv = true, vmlinux = true
BPF: using embedded BPF object
BTF: using BTF file from environment: /home/rafaeldtinoco/work/sources/ebpf/btfgen/btfs/5.13.0-20-generic.btf
unpacked CO:RE bpf object file into memory
TIME             CONTAINER_ID  UID    COMM             PID/host        TID/host        RET              EVENT                ARGS
04:52:27:510010  5040962671ac  0      bash             1      /3966413 1      /3966413 3                openat               dirfd: -100, pathname: /etc/ld.so.cache, flags: O_RDONLY|O_CLOEXEC, mode: 0
04:52:27:510301  5040962671ac  0      bash             1      /3966413 1      /3966413 3                openat               dirfd: -100, pathname: /lib/x86_64-linux-gnu/libtinfo.so.6, flags: O_RDONLY|O_CLOEXEC, mode: 0
04:52:27:510980  5040962671ac  0      bash             1      /3966413 1      /3966413 3                openat               dirfd: -100, pathname: /lib/x86_64-linux-gnu/libc.so.6, flags: O_RDONLY|O_CLOEXEC, mode: 0
04:52:27:516075  5040962671ac  0      bash             1      /3966413 1      /3966413 3                openat               dirfd: -100, pathname: /dev/tty, flags: O_RDWR|O_NONBLOCK, mode: 0
04:52:27:521596  5040962671ac  0      bash             1      /3966413 1      /3966413 3                openat               dirfd: -100, pathname: /etc/nsswitch.conf, flags: O_RDONLY|O_CLOEXEC, mode: 0
04:52:27:521795  5040962671ac  0      bash             1      /3966413 1      /3966413 3                openat               dirfd: -100, pathname: /etc/passwd, flags: O_RDONLY|O_CLOEXEC, mode: 0
```

yanivagman:
Thanks @rafaeldtinoco for pointing me to this issue!
It turns out that may solution doesn't handle cgroup v1 properly.
Ubuntu 18.04.6 uses cgroup v1 as the default, so you can see this problem. Luckily for us, it is easy to check if cgroup v2 is enabled or not in the system: https://docs.docker.com/config/containers/runmetrics/#enumerate-cgroups

The problem is that bpf_get_current_cgroup_id() helper returns cgroup v2 id, which doesn't match any of the cgroup ids found while walking cgroupfs (v1).

The solution I have in mind for this:
1. On tracee init - check if cgroup v2 is enabled or not (by checking for the existence of /sys/fs/cgroup/cgroup.controllers)
2. Init bpf config map with cgroup v1/v2 enabled flag
3. If cgroup v2 - use current logic
4. Otherwise, for each event context, get cgroup id of subsystem 0 (cpuset) (task->cgroups->subsys[0]->cgroup->kn->id)
5. As we use a specific cgroup subsystem (cpuset), only parse cpuset folders when iterating cgroup v1 on init and when cgroup_mkdir is received

WDYT?

yanivagman:
One more point that we should add to the docs:
If the user wants a correct enumeration of existing containers (whether cgroup v1/v2), he should run tracee in the host cgroup namespace (cgroupns=host in docker).
This point is also true today, when we parse cgroupfs on init.

yanivagman:
Note about my recent optimization in the last commit (cgroup v1 XOR v2, not both), taken from https://medium.com/nttlabs/cgroup-v2-596d035be4d7 : "cgroup v1 and v2 are incompatible and can’t be enabled simultaneously. Although there is “hybrid” configuration that allows mounting both v1 hierarchy and v2 hierarchy, the “hybrid” mode is underutilized for containers because you can’t enable v2 controllers that are already enabled for v1."

rafaeldtinoco:
Thanks @yanivagman, I'm reviewing this today and will let you know.

rafaeldtinoco:
> 5\. As we use a specific cgroup subsystem (cpuset), only parse cpuset folders when iterating cgroup v1 on init and when cgroup_mkdir is received

Yes, I think that is the secret. To rely only in a single cgroup subset and, yep, seems like cpuset is the one that should be taken.

rafaeldtinoco:
Sorry I accidentally closed the PR. Re-opened it.

PR #1139, opened 2021-11-14
tracee-ebpf: fix events sent in parallel to raw_sys_exit event

### LINKED ISSUE #1140: errors when raw_sys_exit is used in parallel with syscall events
Running tracee with: `sudo ./dist/tracee-ebpf -t e=sys_exit,poll` gives the following output:

```
13:24:30:322149  1000   xfce4-terminal   15797   15797   2                sys_exit             syscall: poll
failed to read argument 1 of event poll: invalid arg index 66
failed to read argument 2 of event poll: invalid arg index 143
failed to read argument 3 of event poll: invalid arg index 85
13:24:30:322147  1000   xfce4-terminal   15797   15797   2                poll                 fds: 0x9323F00000000007, : <nil>, : <nil>, : <nil>
```

Root cause of this error:

sys_exit tracepoint is responsible to sent two different kind of events:
1. raw sys_exit events (which only have syscall id argument)
2. syscall events (with full arguments data)

When used in parallel, the second event doesn't set the events buffer offset and argnum to 0 correctly

yanivagman:
fix #1140

### DIFF
--- a/tracee-ebpf/tracee/tracee.bpf.c
+++ b/tracee-ebpf/tracee/tracee.bpf.c
@@ -2064,6 +2064,8 @@ int tracepoint__raw_syscalls__sys_exit(struct bpf_raw_tracepoint_args *ctx)
             // We can't use saved args after execve syscall, as pointers are invalid
             // To avoid showing execve event both on entry and exit,
             // we only output failed execs
+            data.buf_off = sizeof(context_t);
+            data.context.argnum = 0;
             save_args_to_submit_buf(&data, types, &sys->args);
             data.context.ts = sys->ts;
             events_perf_submit(&data, id, ret);


tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2064,6 +2064,8 @@ int tracepoint__raw_syscalls__sys_exit(struct bpf_raw_tracepoint_args *ctx)
             // We can't use saved args after execve syscall, as pointers are invalid
             // To avoid showing execve event both on entry and exit,
             // we only output failed execs
+            data.buf_off = sizeof(context_t);
```
grantseltzer, line 2067: Why sizeof(context_t) and not 0 if this isn't being used?

grantseltzer (approved):
LGTM

PR #1141, opened 2021-11-14
use plain addr argument

roikol:
there is no need to convert addr argument to string of json. now using it as it comes: map[string]string


yanivagman (approved):
LGTM

PR #1142, opened 2021-11-15
tracee-ebpf: add MemProtAlert type to external package

yanivagman:
Not too long ago we had "security-alerts" as part of tracee-ebpf (until #1114 removed it).
"alerts" where messages sent by an eBPF event as an extra argument, to pass some information about a security alert.
Introduce MemProtAlert in the external package, so it can be used by tracee-ebpf consumers

This will also allow us to expose "alerts" as raw arguments (related #1123 )

### DIFF
--- a/tracee-ebpf/tracee/argprinters.go
+++ b/tracee-ebpf/tracee/argprinters.go
@@ -23,17 +23,13 @@ func Print16BytesSliceIP(in []byte) string {
 	return ip.String()
 }
 
-func (t *Tracee) prepareArgs(ctx *context, args map[string]interface{}) error {
+func (t *Tracee) parseArgs(ctx *context, args map[string]interface{}) error {
 	for key, arg := range args {
 		if ptr, isUintptr := arg.(uintptr); isUintptr {
 			args[key] = fmt.Sprintf("0x%X", ptr)
 		}
 	}
 
-	// EventIDs which are optionally parsed/enriched into human readable formats
-	if !t.config.Output.ParseArguments {
-		return nil
-	}
 	switch ctx.EventID {
 	case MemProtAlertEventID:
 		if alert, isUint32 := args["alert"].(uint32); isUint32 {
--- a/tracee-ebpf/tracee/argprinters.go
+++ b/tracee-ebpf/tracee/argprinters.go
@@ -5,7 +5,8 @@ import (
 	"fmt"
 	"github.com/aquasecurity/libbpfgo/helpers"
 	"net"
-	"strconv"
+
+	"github.com/aquasecurity/tracee/tracee-ebpf/external"
 )
 
 // PrintUint32IP prints the IP address encoded as a uint32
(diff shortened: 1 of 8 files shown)


tracee-ebpf/tracee/argprinters.go
```diff
@@ -53,17 +55,15 @@ func (t *Tracee) prepareArgs(ctx *context, args map[string]interface{}) error {
...
-		}
-	}
-
 	// EventIDs which are optionally parsed/enriched into human readable formats
 	if !t.config.Output.ParseArguments {
 		return nil
 	}
 	switch ctx.EventID {
+	case MemProtAlertEventID:
+		if message, isMessage := args["alert"].(external.Message); isMessage {
```
itaysk, line 64: message?

itaysk (commented):
Good that you raise this an fix it. I think it should be called event not message. We use event in our terminology to represent something tracee-ebpf emits.

itaysk:
related: https://github.com/aquasecurity/tracee/discussions/1105

yanivagman:
This is not an event. The "message" concept presented here is an argument inside an event.
The current use case is of the "mem_prot_alert" event, where the only argument is the alert.
But this concept can be used by any event, to decode different messages as an argument of the event

tracee-ebpf/tracee/argprinters.go
yanivagman: No. "alert" is the name of the argument, and I think we should keep this name

itaysk:
ok so I misunderstood your intention. it's just about encoding the strings passed as event arguments?
`mem_prot_alert` is not an event not message, but it's `alert` argument is a message?

yanivagman:
```
ok so I misunderstood your intention. it's just about encoding the strings passed as event arguments?
mem_prot_alert is not an event not message, but it's alert argument is a message?
```

Yes, it's about encoding the strings as event arguments. It then can be used by external consumer as a raw argument (without "stringifying" it), or given as a parsed string.

mem_prot_alert is a special event, which has only one argument called "alert" which is of message type. In the future, we can think of other abstractions, such as what you suggested in #1105

itaysk:
does all messages need `Ts` and `Payload`? It seems to me that instead of generalizing `security_alert`, what we needed was the opposite, to make it more specific no? (e.g `mem_prot_alert_message` or something like this)
Take a look at https://github.com/aquasecurity/libbpfgo/pull/89#discussion_r750007493 and tell me what you think

yanivagman:
> does all messages need Ts and Payload? It seems to me that instead of generalizing security_alert, what we needed was the opposite, to make it more specific no? (e.g mem_prot_alert_message or something like this) like commit creds
> Take a look at aquasecurity/libbpfgo#89 (comment) and tell me what you think

I took the extra step and removed Ts and Payload.
"message" is now an enum exposed by external, which can be translated to a string when parse-arguments is chosen

itaysk (approved):

PR #1144, opened 2021-11-16
fix: using exec-hash instead of exec-info

mtcherni95:


yanivagman (approved):
Thanks for the fix
LGTM

Issue #1150, opened 2021-11-18
tracee-ebpf should check that /proc/sys/kernel/ftrace_enabled is '1' (enabled) 


mtcherni95:

mtcherni95:
after chatting with @yanivagman and @rafaeldtinoco , maybe this issue should be in libbpfgo.

mtcherni95:
should we enable (set it to '1') it if at start-up we see that ftrace_enabled is '0'? or just report a warning? @yanivagman

yanivagman:
@mtcherni95 ftrace_enabled is set to 1 by default:
https://01.org/linuxgraphics/gfx-docs/drm/trace/ftrace.html#ftrace-enabled

If the user chose to set it to 0, there might be a reason for that (although I can't think of a reason why someone who is running tracing programs will ever set this value to 0) - let's warn in that case.

mtcherni95:
> @mtcherni95 ftrace_enabled is set to 1 by default:
> 
> https://01.org/linuxgraphics/gfx-docs/drm/trace/ftrace.html#ftrace-enabled
> 
> 
> 
> If the user chose to set it to 0, there might be a reason for that (although I can't think of a reason why someone who is running tracing programs will ever set this value to 0) - let's warn in that case.

Sounds good to me.

PR #1152, opened 2021-11-21
checking /proc/sys/kernel/ftrace_enabled

mtcherni95:
Resolves: #1150

### DIFF
--- a/tracee-ebpf/main.go
+++ b/tracee-ebpf/main.go
@@ -95,13 +95,19 @@ func main() {
 			if err != nil {
 				return err
 			}
-			err = checkRequiredCapabilities(selfCap)
-			if err != nil {
+			if err = checkRequiredCapabilities(selfCap); err != nil {
 				return err
 			}
 
-			// OS kconfig information
+			enabled, err := helpers.FtraceEnabled()
+			if err != nil {
+				return err
+			}
+			if !enabled {
+				fmt.Fprintf(os.Stderr, "ftrace_enabled: warning: ftrace is not enabled, kernel events won't be caught, make sure to enable it by executing echo 1 | sudo tee /proc/sys/kernel/ftrace_enabled")
+			}
 
+			// OS kconfig information
 			kernelConfig, err := helpers.InitKernelConfig()
 			if err == nil { // do not fail (yet ?) if we cannot init kconfig
 				kernelConfig.AddNeeded(helpers.CONFIG_BPF, helpers.BUILTIN)
--- a/tracee-ebpf/go.mod
+++ b/tracee-ebpf/go.mod
@@ -8,10 +8,10 @@ require (
 	github.com/aquasecurity/libbpfgo v0.2.2-libbpf-0.5.0
 	github.com/aquasecurity/tracee/tracee-ebpf/external v0.0.0-20210903145311-dfdb5d66613f
 	github.com/google/gopacket v1.1.19
+	github.com/hashicorp/golang-lru v0.5.4
 	github.com/stretchr/testify v1.7.0
 	github.com/syndtr/gocapability v0.0.0-20200815063812-42c35b437635
 	github.com/urfave/cli/v2 v2.3.0
 	golang.org/x/sys v0.0.0-20210903071746-97244b99971b
 	inet.af/netaddr v0.0.0-20210903134321-85fa6c94624e
-	github.com/hashicorp/golang-lru v0.5.4
 )


itaysk:
> Resolves: #1150

this goes in the PR body, not title :)

itaysk:
In the linked issue you say you will add the check to libbpfgo, what changed?

rafaeldtinoco:
Yes, @mtcherni95 , please check:

https://github.com/aquasecurity/libbpfgo/pull/94

mtcherni95:
I will update PR by using helper in libbpfgo - this will require update of go mod version of libbpgo

rafaeldtinoco:
> I will update PR by using helper in libbpfgo - this will require update of go mod version of libbpgo

Would you include:

```diff
commit eb264af (HEAD -> main)
Author: Rafael David Tinoco <rafaeldtinoco@gmail.com>
Date:   Fri Nov 5 20:01:05 2021

    tracee: new AttachTracepoint prototype

diff --git a/tracee-ebpf/tracee/tracee.go b/tracee-ebpf/tracee/tracee.go
index ee1d08e..72d3992 100644
--- a/tracee-ebpf/tracee/tracee.go
+++ b/tracee-ebpf/tracee/tracee.go
@@ -868,7 +868,8 @@ func (t *Tracee) initBPF() error {
                        case kretprobe:
                                _, err = prog.AttachKretprobe(probe.event)
                        case tracepoint:
-                               _, err = prog.AttachTracepoint(probe.event)
+                               tpEvent := strings.Split(probe.event, ":")
+                               _, err = prog.AttachTracepoint(tpEvent[0], tpEvent[1])
                        case rawTracepoint:
                                tpEvent := strings.Split(probe.event, ":")[1]
                                _, err = prog.AttachRawTracepoint(tpEvent)
```

to tracee as well ? You will need when bumping libbpfgo version.

grantseltzer:
@mtcherni95 Now that we've upgraded the libbpfgo dependency to the latest release (which includes this helper function) you can update this PR.

mtcherni95:
> @mtcherni95 Now that we've upgraded the libbpfgo dependency to the latest release (which includes this helper function) you can update this PR.

great, thanks a lot

tracee-ebpf/main.go
```diff
@@ -95,13 +95,19 @@ func main() {
...
+			if err = checkRequiredCapabilities(selfCap); err != nil {
+				return err
+			}
+
+			enabled, err := helpers.FtraceEnabled()
 			if err != nil {
 				return err
 			}
+			if !enabled {
+				fmt.Fprintf(os.Stderr, "ftrace_enabled: warning: ftrace is not enabled - kprobes won't be inserted\n")
```
grantseltzer, line 107: To a user, I don't know if having "kprobes won't be inserted" is very useful. Is it possible to list events that won't be traced? Are we sure that ftrace is absolutely required for all kprobes?

mtcherni95: I think so yeah. @yanivagman ?

rafaeldtinoco: Ftrace is needed for kprobes as they're built on top of ftrace base architecture (https://www.kernel.org/doc/Documentation/trace/ftrace-design.txt), like an extension to it. I don't think we should expand to much here, but I do think the message could be clearer for an average user that might not know what kprobes and ftrace are.

mtcherni95: I actually think that if a user is using Tracee then most likely he will have some technical background (and kprobes knowledge as well). I suggest something like `"ftrace_enabled: warning:  ftrace is not enabled - some events may not be caught (kprobes not enabled)"`.

rafaeldtinoco: I believe no events, at all, will be caught, if ftrace is disabled. 

And I disagree that users of tracee will know what a kprobe (or ftrace) is. 

I do agree that a warning like that could be helpful as long as it is rephrased to something like:

`ftrace_enabled: warning: ftrace is not enabled, kernel events won't be caught, make sure to enable it by executing echo 1 | sudo tee /proc/sys/kernel/ftrace_enabled`

mtcherni95: You do get events coming from tracepoints. 

Let's go with your message then :)

rafaeldtinoco: > You do get events coming from tracepoints.

I see what you mean. But our ebpf logic might be broken as there are maps manipulated by different eBPF programs (with different probe methods) in different times, I believe. 

I agree that if we consider tracee purely as an introspection tool, final user might know what we're talking about... but we're trying to shift that mindset as well (thus the explicit message about how to get it enabled).

Cool, thanks.

grantseltzer: So if this is the case should we just error out if ftrace is disabled? Is there any point in running tracee?

yanivagman: > So if this is the case should we just error out if ftrace is disabled? Is there any point in running tracee?

I don't think so - tracepoint will still work, which means that all the syscalls + some other events will still work (> 95% of the events are not kprobes) - so I think we should just warn

rafaeldtinoco: I believe you're right @grantseltzer. Its better to just fail instead of trying to rely in a tainted execution.

rafaeldtinoco: Alright then, works for me.

grantseltzer (approved):
LGTM

yanivagman (approved):

PR #1153, opened 2021-11-22
fix wrong type assertion

mtcherni95:


tracee-ebpf/tracee/events_processor.go
```diff
@@ -155,19 +156,19 @@ func (t *Tracee) processEvent(ctx *context, args map[string]interface{}, argMeta
...
-					if !ok || lastCtime != sourceFileCtime {
+					if !ok || lastCtime != castedSourceFileCtime {
 						//capture
 						err = CopyFileByPath(sourceFilePath, destinationFilePath)
 						if err != nil {
 							return err
 						}
 						//mark this file as captured
-						t.capturedFiles[capturedFileID] = sourceFileCtime
+						t.capturedFiles[capturedFileID] = castedSourceFileCtime
```
yanivagman, line 171: I think this is the only place that requires to cast the ctime, right?
If so, you can avoid the declaration and usage of castedSourceFileCtime in other places, and only cast here

mtcherni95: we use it also in line 164 - need to have the casted variable, no?

yanivagman: You can compare without a cast, knowing that timestamp will never be longer than 63 bits (thus you can compare the "signed" timestamp to unsigned)

mtcherni95: not sure you can actually "Invalid operation: lastCtime != sourceFileCtime (mismatched types int64 and uint64)"

Issue #676, opened 2021-04-05
Add container start/stop events


tomyweiss:

itaysk:
I'm splitting this into 2 separate issues as these are unrelated.

yanivagman:
@tomyweiss @itaysk maybe we can add this functionality with some helper function, but why do you think a signature is required here?

itaysk:
true, our signature library should be for detecting suspicious behaviors. if people want to add signatures for operational insights, that's fine but doesn't belong in tracee-rules

yanivagman:
I updated this issue so that we should add container start/stop events.
This can be done using higher level events

PR #1154, opened 2021-11-22
tracee-ebpf: filter containers using cgroup id

yanivagman:
Currently, to filter a container in the bpf code we use "pid != host_pid", and to identify a NEW container we look for pid 1 in a new pid namespace.
These checks are not always true (e.g. pid=host or containers that share pid namespace).

This PR introduces a new way to identify containers in the bpf code using cgroup ids.
To do this, the following algorithm is used:
1. Introduce a new (bpf) containers_map with a key of cgroup_id (lower 32 bits) and value of CONTAINER_EXISTED, CONTAINER_CREATED, CONTAINER_STARTED
2. Update containers_map on init with existing containers (CONTAINER_EXISTED)
3. Update containers_map when a new cgroup is created (CONTAINER_CREATED)
4. If a cgroup path is not of a new container (checked by userspace when a cgroup_mkdir event is received) - remove it from the bpf container_map
5. Update containers_map when a process of a newly created cgroup (CONTAINER_CREATED) has executed a binary (CONTAINER_STARTED)
6. Identifying a container process can then be done by using the cgroup id to check in containers_map if the state is either CONTAINER_EXISTED or CONTAINER_STARTED
7. Identifying a process of a new container can be done by checking if the state is CONTAINER_STARTED
8. We can then remove the check "pid != host_pid" that is used to identify a container with a check of existing cgroup_id in the bpf containers_map
9. Use cgroup_rmdir to delete cgroup_id from the bpf map

By doing this change, we will also be able to address the following issues in future PRs:
1. #676 - We will be able to have a new container_started EVENT which will also include the container runtime extracted from the cgroup name (e.g. docker/containerd/podman)
2. #255 - We will be able to create container id filters in kernel (by mapping container id to cgroup id)

### DIFF
--- a/tracee-ebpf/tracee/events.go
+++ b/tracee-ebpf/tracee/events.go
@@ -74,6 +74,22 @@ func (t *Tracee) processEvents(done <-chan struct{}) error {
 			continue
 		}
 
+		if !t.containers.CgroupExists(ctx.CgroupID) {
+			// Handle false container negatives (we should have identified this container id, but we didn't)
+			// This situation can happen as a race condition when updating the cgroup map.
+			// In that case, we can try to look for it in the cgroupfs and update the map if found.
+			t.containers.CgroupLookupUpdate(ctx.CgroupID)
+		}
+		containerId := t.containers.GetCgroupInfo(ctx.CgroupID).ContainerId
+		if (t.config.Filter.ContFilter.Enabled || t.config.Filter.NewContFilter.Enabled) && containerId == "" {
+			// Don't trace false container positives -
+			// a container filter is set by the user, but this event wasn't originated in a container.
+			// Although kernel filters shouldn't submit such events, we do this check to be on the safe side.
+			// For example, it might be that a new cgroup was created, and not by a container runtime,
+			// while we still didn't processed the cgroup_mkdir event and removed the cgroupid from the bpf container map.
+			continue
+		}
+
 		// Only emit events requested by the user
 		if !t.eventsToTrace[ctx.EventID] {
 			continue
@@ -117,7 +133,7 @@ func (t *Tracee) processEvents(done <-chan struct{}) error {
 			PIDNS:               int(ctx.PidID),
 			ProcessName:         string(bytes.TrimRight(ctx.Comm[:], "\x00")),
 			HostName:            string(bytes.TrimRight(ctx.UtsName[:], "\x00")),
-			ContainerID:         t.containers.GetContainerId(ctx.CgroupID),
+			ContainerID:         containerId,
 			EventID:             int(ctx.EventID),
 			EventName:           EventsIDToEvent[int32(ctx.EventID)].Name,
 			ArgsNum:             int(ctx.Argnum),
--- a/tracee-ebpf/tracee/events_processor.go
+++ b/tracee-ebpf/tracee/events_processor.go
@@ -212,7 +212,11 @@ func (t *Tracee) processEvent(ctx *context, args map[string]interface{}, argMeta
 		if !ok {
 			return fmt.Errorf("error parsing cgroup_mkdir args")
 		}
-		t.containers.CgroupUpdate(cgroupId, path)
+		info, err := t.containers.CgroupUpdate(cgroupId, path)
+		if err == nil && info.ContainerId == "" {
+			// If not a new container (no regex match) - remove from the bpf container_map
+			t.containers.RemoveFromBpfMap(t.bpfModule, cgroupId)
+		}
 
 	case CgroupRmdirEventID:
 		cgroupId, ok := args["cgroup_id"].(uint64)
--- a/tracee-ebpf/tracee/containers.go
+++ b/tracee-ebpf/tracee/containers.go
@@ -2,6 +2,7 @@ package tracee
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
 	"io/fs"
 	"os"
@@ -9,13 +10,22 @@ import (
 	"regexp"
 	"strings"
 	"syscall"
+	"unsafe"
+
+	bpf "github.com/aquasecurity/libbpfgo"
 )
 
 // Containers contain information about host running containers in the host.
 type Containers struct {
-	cgroupV1     bool
-	cgroupMP     string
-	cgroupContId map[uint32]string
+	cgroupV1 bool
+	cgroupMP string
+	cgroups  map[uint32]CgroupInfo
+}
+
+type CgroupInfo struct {
+	Path        string
+	ContainerId string
+	Runtime     string
 }
 
 // InitContainers initializes a Containers object and returns a pointer to it.
@@ -27,23 +37,14 @@ func InitContainers() *Containers {
 	}
 
 	return &Containers{
-		cgroupV1:     cgroupV1,
-		cgroupMP:     "",
-		cgroupContId: make(map[uint32]string),
+		cgroupV1: cgroupV1,
+		cgroupMP: "",
+		cgroups:  make(map[uint32]CgroupInfo),
 	}
 }
 
 func (c *Containers) IsCgroupV1() bool {
 	return c.cgroupV1
-}
-
-// GetContainers provides a list of all added containers by their uuid.
-func (c *Containers) GetContainers() []string {
-	var conts []string
-	for _, v := range c.cgroupContId {
-		conts = append(conts, v)
-	}
-	return conts
 }
 
 // Populate will populate all Containers information by reading mounted proc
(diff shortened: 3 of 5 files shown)


tracee-ebpf/tracee/events_processor.go
```diff
@@ -212,7 +212,11 @@ func (t *Tracee) processEvent(ctx *context, args map[string]interface{}, argMeta
 		if !ok {
 			return fmt.Errorf("error parsing cgroup_mkdir args")
 		}
-		t.containers.CgroupUpdate(cgroupId, path)
+		info, err := t.containers.CgroupUpdate(cgroupId, path)
+		if err == nil && info.ContainerId == "" {
+			// If not a new container (no regex match) - remove from the bpf container_map
+			t.containers.RemoveFromBpfMap(t.bpfModule, cgroupId)
```
rafaeldtinoco, line 218: Okay, so.. just to check if I understood, you remove it from the container_map but you keep it in the t.containers with empty CgroupInfo{} so it can be removed whenever a cgroup rmdir event happens. Is that it ?

rafaeldtinoco (commented):
Could you check the 2 comments I left ?

Also, could you please break lines in git log (as its the standard).

Overall change looks very good.

I'll do some testing now.

tracee-ebpf/tracee/events.go
```diff
@@ -74,6 +74,22 @@ func (t *Tracee) processEvents(done <-chan struct{}) error {
 			continue
 		}
 
+		if !t.containers.CgroupExists(ctx.CgroupID) {
+			// Handle false container negatives (we should have identified this container id, but we didn't)
+			// This situation can happen as a race condition when updating the cgroup map.
+			// In that case, we can try to look for it in the cgroupfs and update the map if found.
+			t.containers.CgroupLookupUpdate(ctx.CgroupID)
+		}
+		containerId := t.containers.GetCgroupInfo(ctx.CgroupID).ContainerId
```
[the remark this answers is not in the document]
yanivagman: You are right.
I assumed I handled that by always updating the map, even if no match is found - CgroupUpdate() was modified to do exactly that. The problem is that CgroupUpdate() is not called if err==nil (and not to ErrExist) - I'll that, thanks!

tracee-ebpf/tracee/events_processor.go
yanivagman: Yes.
Only the cgroup_mkdir and cgroup_rmdir events update the USERSPACE containers map (opposed to the bpf containers map that only keeps track of "real" containers). We keep "non-containers" cgroups so it will be found by CgroupExists() in processEvents (and will give an empty container id)

rafaeldtinoco:
For my current dev environment (5.13 and cgroupv2) it worked fine:

```
[rafaeldtinoco@fujitsu:~]$ sudo bpftool map dump name containers_map
Found 0 elements
[rafaeldtinoco@fujitsu:~]$ sudo bpftool map dump name containers_map
key: 33 7b 00 00  value: 03
key: 90 7b 00 00  value: 03
key: ed 7b 00 00  value: 03
Found 3 elements
[rafaeldtinoco@fujitsu:~]$ sudo bpftool map dump name containers_map
key: 33 7b 00 00  value: 03
key: 90 7b 00 00  value: 03
key: ed 7b 00 00  value: 03
Found 3 elements
[rafaeldtinoco@fujitsu:~]$ sudo bpftool map dump name containers_map
key: 33 7b 00 00  value: 03
key: 90 7b 00 00  value: 03
key: ed 7b 00 00  value: 03
Found 3 elements
[rafaeldtinoco@fujitsu:~]$ sudo bpftool map dump name containers_map
key: 33 7b 00 00  value: 03
key: 90 7b 00 00  value: 03
key: ed 7b 00 00  value: 03
Found 3 elements
[rafaeldtinoco@fujitsu:~]$ sudo bpftool map dump name containers_map
key: 33 7b 00 00  value: 03
key: 90 7b 00 00  value: 03
Found 2 elements
[rafaeldtinoco@fujitsu:~]$ sudo bpftool map dump name containers_map
Found 0 elements
```

Checking older kernels and cgroupv1 now...

rafaeldtinoco:
@yanivagman,

While for cgroups v2 I have a single entry in containers_map to each new spin-ed docker container, in cgroups v1 + kernel 5.4 I have many:

```
[rafaeldtinoco@bionic:~]$ sudo bpftool map dump id 51
key: ae 04 00 00  value: 02
key: 3a 00 00 00  value: 03
key: 1b 00 00 00  value: 02
key: 7a 01 00 00  value: 02
key: 48 02 00 00  value: 02
key: e7 06 00 00  value: 02
key: 2a 00 00 00  value: 02
key: 18 00 00 00  value: 02
key: 12 00 00 00  value: 02
key: fb 01 00 00  value: 02
key: 75 03 00 00  value: 02
Found 11 elements
```

I guess that is because we should be taking cpuset entry only ? (or something like it). Could you double check ?

rafaeldtinoco:
Same happens in a 5.13 kernel with cgroup v1:

```
$ sudo bpftool map dump id 8
key: d7 03 00 00  value: 02
key: 10 00 00 00  value: 02
key: 2d 00 00 00  value: 02
key: 28 00 00 00  value: 03
key: 0d 00 00 00  value: 02
key: 33 05 00 00  value: 02
key: a3 0a 00 00  value: 02
key: a8 02 00 00  value: 02
key: 13 00 00 00  value: 02
key: 33 02 00 00  value: 02
key: c5 01 00 00  value: 02
Found 11 elements
```

so its not related to kernel version, just wanted to confirm for you.

yanivagman:
> @yanivagman,
> 
> While for cgroups v2 I have a single entry in containers_map to each new spin-ed docker container, in cgroups v1 + kernel 5.4 I have many:

Yes, I'm aware of that. At first, I wanted to save only the cpuset controller cgroups in userspace (we do that already when populating the cgroups map at init), but in runtime cgroup_mkdir doesn't tell you which controller the cgroup belongs to (although we probably can find this information in the kernel structs and send it as another argument of the event). I didn't do that because:
1. Current logic works - cgroups map in userspace will be updated with any cgroup in runtime, and removed when the cgroup removed - so no functional problem
2. The extra space required to save these cgroups shuoldn't be a problem, as we are not saving it in the bpf map (where we should care about every entry we insert), but in userspace
3. Implementing the logic to send the controller name requires extra work, and poking with the kernel structs may require changes between kernel versions , which we better avoid

rafaeldtinoco:
> > @yanivagman,
> > While for cgroups v2 I have a single entry in containers_map to each new spin-ed docker container, in cgroups v1 + kernel 5.4 I have many:
> 
> Yes, I'm aware of that. At first, I wanted to save only the cpuset controller cgroups in userspace (we do that already when populating the cgroups map at init), 

Correct, if container already exists we have a single entry:

```
$ sudo bpftool map dump id 42
key: 28 00 00 00  value: 01
```

with value 1 and, after container exists, it is deleted correctly.

> but in runtime cgroup_mkdir doesn't tell you which controller the cgroup belongs to (although we probably can find this information in the kernel structs and send it as another argument of the event). 

```
$ sudo bpftool map dump id 42
key: f3 03 00 00  value: 02
key: 43 02 00 00  value: 02
key: 18 00 00 00  value: 02
key: 12 00 00 00  value: 02
key: d4 01 00 00  value: 02
key: 03 0b 00 00  value: 02
key: 42 00 00 00  value: 02
key: c0 02 00 00  value: 02
key: 3a 00 00 00  value: 03
key: 1b 00 00 00  value: 02
key: 59 05 00 00  value: 02
Found 11 elements
```

Okay, that explains why we have multiple entries of value 02 (created) and a single one as 03 (started). I'm just wondering if this 'random' pick of a cgroup_id to be '03' - based on which cgroup directory is created first - while others are 02 - will step into our way in any logic.

Do you mind explaining why that wouldn't happen ?

> I didn't do that because:
> 
> 1. Current logic works - cgroups map in userspace will be updated with any cgroup in runtime, and removed when the cgroup removed - so no functional problem
> 2. The extra space required to save these cgroups shuoldn't be a problem, as we are not saving it in the bpf map (where we should care about every entry we insert), but in userspace
> 3. Implementing the logic to send the controller name requires extra work, and poking with the kernel structs may require changes between kernel versions , which we better avoid

And it indeed works as all entries are cleaned at the container exit:

```
$ sudo bpftool map dump id 42
Found 0 elements
```

yanivagman:
> Okay, that explains why we have multiple entries of value 02 (created) and a single one as 03 (started). I'm just wondering if this 'random' pick of a cgroup_id to be '03' - based on which cgroup directory is created first - while others are 02 - will step into our way in any logic.
> 
> Do you mind explaining why that wouldn't happen ?

Yes. There is no problem with the logic because the cgroup_id we send per any event is always of the cpuset subsystem. This is also the cgroupid used (with cgroup v1) to filter events.

I don't like, however, the fact that the bpf map id "dirty" with all the cgroup controllers as you showed here. We should think of a way to clean these other cgroups

rafaeldtinoco:
> > Okay, that explains why we have multiple entries of value 02 (created) and a single one as 03 (started). I'm just wondering if this 'random' pick of a cgroup_id to be '03' - based on which cgroup directory is created first - while others are 02 - will step into our way in any logic.
> > Do you mind explaining why that wouldn't happen ?
> 
> Yes. There is no problem with the logic because the cgroup_id we send per any event is always of the cpuset subsystem. This is also the cgroupid used (with cgroup v1) to filter events.
> 
> I don't like, however, the fact that the bpf map is "dirty" with all the cgroup controllers as you showed here. We should think of a way to clean these other cgroups

Okay, let's open an issue to clean that up as this PR is already "reviewed" and complex enough. Sounds good ? I'm approving this one as the logic works.

yanivagman:
> I'm just wondering if this 'random' pick of a cgroup_id to be '03' - based on which cgroup directory is created first

So the answer is - the cgroup id that has the value '03' (started) is the cpuset cgroupid, and not the first directory to be created

rafaeldtinoco (approved):
LGTM after discussions.

yanivagman:
Maybe we will not have to clean the bpf map in userspace for the other controllers if we won't create the entry for other controllers (non cpuset controllers) at first in the cgroup_mkdir event.
I'm checking if this change is a 2-3 lines change or not (if not, we can keep it to a future PR)

yanivagman:
Ok, this seems to take more than 2-3 lines as it is only relevant for cgroup v1.
But this is what I have in mind:
If cgroup v1 {
    if cgroup->self->ss->id == 0
        update bpf map
} else {
    update_bpf_map
}

yanivagman:
> Ok, this seems to take more than 2-3 lines as it is only relevant for cgroup v1.
> But this is what I have in mind:
> 
> ```
> If cgroup v1 {
>     if cgroup->self->ss->id == 0
>         update bpf map
> } else {
>     update_bpf_map
> }
> ```

So this method doesn't work as `ss` is NULL (https://elixir.bootlin.com/linux/v5.15.4/source/include/linux/cgroup-defs.h#L361)

As this issue is not a functional issue (only keeping the bpf map small), in addition to the fact that it only happens with cgroup v1 systems (that are already starting to disappear), let's merge this now.

PR #1155, opened 2021-11-22
docs: add cgroupns=host docker option

yanivagman:
To identify existing containers, `--cgroupns=host` should be provided to the `docker run` command


rafaeldtinoco (approved):
LGTM

PR #1164, opened 2021-11-25
docs: use mkdocs macros plugin to specify version of tracee release artifacts

danielpacak:
To preview changes run `make mkdocs-serve` and open http://localhost:8000 in you web browser.

pin_version

### DIFF
--- a/docs/Dockerfile
+++ b/docs/Dockerfile
@@ -1,4 +1,4 @@
-FROM squidfunk/mkdocs-material:7.3.6
+FROM squidfunk/mkdocs-material:8.0.1
 
 RUN pip install mike
 RUN pip install mkdocs-macros-plugin
--- a/docs/index.md
+++ b/docs/index.md
@@ -21,17 +21,25 @@ Before you proceed, make sure you follow the [minimum requirements for running T
 If running on __BTF enabled kernel__:
 
 ```bash
-docker run --name tracee --rm --pid=host --cgroupns=host --privileged -v /tmp/tracee:/tmp/tracee -it aquasec/tracee:latest
+docker run --name tracee --rm -it --pid=host --cgroupns=host --privileged \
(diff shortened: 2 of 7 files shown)


mkdocs.yml
```diff
@@ -41,6 +41,13 @@ extra:
   version:
     method: mike
     provider: mike
+  # var defines custom variables that you can use throughout the documentation
+  # using the following syntax: {{ var.build.version }}.
+  var:
+    build:
+      # NOTE: We must update the current version before we publish the documentation.
+      # TODO: Alternatively, use {{ git.tag }} instead of {{ build.version }}.
```
itaysk, line 49: I looked into this today. `mkdocs-macros` declares `git.tag` as `git describe --tags` and then split by `-` taking the first part.
https://github.com/fralau/mkdocs_macros_plugin/blob/master/mkdocs_macros/context.py#L137
In Tracee, we tag each release with directory paths to all go modules, in order to satisfy go modules discovery. This doesn't align with mkdocs-macros' assumptions.
This is why `{{ git.tag }}` shows `tracee`.

I don't mind defining the version manually, what I mind (a little) is that we commit it to main. I thought about maybe defining it manually on the docs build command, so that it's usable, but not committed, and then I saw we already do this for `mike`. But I didn't find a way to use the mike version from macros.

So I created a simple `mkdocs-macros` plugin that gives us the correct git tag based on our use case. I'll send another PR to add that , and if you like it you can rebase this PR on top of the other one.

itaysk: https://github.com/aquasecurity/tracee/pull/1166

danielpacak: Updated to use `{{ git_tag_version }}` and `{{ git_tag_version[1:] }}` where applicable.

itaysk (approved):

PR #1167, opened 2021-11-28
add ctime to security_file_open and fix variable type

AsafEitani:
Add ctime of the opened file to `security_file_open`.
The type was fixed from `unsigned long` to `u64` because it resulted in parsing errors of following argument.
This issue wasn't noticed because there are no arguments after the `ctime` argument of `sched_process_exec`.

### DIFF
--- a/tracee-ebpf/tracee/tracee.bpf.c
+++ b/tracee-ebpf/tracee/tracee.bpf.c
@@ -892,7 +892,7 @@ static __always_inline unsigned long get_inode_nr_from_file(struct file *file)
     return READ_KERN(f_inode->i_ino);
 }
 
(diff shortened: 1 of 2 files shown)


tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2684,15 +2684,17 @@ int BPF_KPROBE(trace_security_file_open)
     dev_t s_dev = get_dev_from_file(file);
     unsigned long inode_nr = get_inode_nr_from_file(file);
     void *file_path = get_path_str(GET_FIELD_ADDR(file->f_path));
+    u64 ctime = get_ctime_nanosec_from_file(file);
```
mtcherni95, line 2687: where are you using it in user-space?

yanivagman (approved):

Issue #1029, opened 2021-09-26
Change argument type in parseArgs() when required


yanivagman:
`prepareArgs()` may change the type of a given argument to a different type (for example, `flags` argument of type long may be change to type string to reflect the selected flags). This type change may cause issues when marshalling/unmarshalling the arguments and counting on a correct value of the argument type (see https://github.com/aquasecurity/tracee/pull/988#issuecomment-927364548).

A possible fix for this issue may be to change the type in external.ArgMeta when the value is changed in `prepareArgs()`. This can be done by passing the argMetas slice (used by `processEvents()`) as a pointer argument to `prepareArgs()`

itaysk:
We've discussed in the past about always returning the raw values instead of the "pretty" values in all the output modes except table. I think that might be a better solution to this problem as well, WDYT? Now that the printers are available to the consumer via libbpfgo they can "print" them if they want.

yanivagman:
> We've discussed in the past about always returning the raw values instead of the "pretty" values in all the output modes except table. I think that might be a better solution to this problem as well, WDYT? Now that the printers are available to the consumer via libbpfgo they can "print" them if they want.

Yes, I thought about it too when I killed the events pipeline.
The thing is, that in the event pipeline we decode the arguments into an arguments map (maps argument name to value), which is also given as an argument to `prepareArgs()` to easily look and replace arguments of a given event. If we will take `prepareArgs()` out of `processEvents()` into main, then we will either need to recreate the arguments map, or use the `external.Argument` slice (which is ordered by the argument index) where we will have to iterate the slice everytime we want to find some specific argument.

itaysk:
> If we will take prepareArgs() out of processEvents() into main,

I didn't say we need to do this.. we just need to wrap the translation code in with a flag which is turned on only if output=table

yanivagman:
I think that table printer shouldn't be any different from other printers. Maybe we should have output option:pretty instead?

itaysk:
yes.. this is what I meant what I said to wrap the code with a flag which is tuned on when output=table

> we want pretty printing to be enabled by default (as this is the current state)

only for table. every other mode won't need this.
I think we can do option:pretty which will be turned on by default when format:table

rafaeldtinoco:
I'm +1 on having pretty print by default only for table (like we do) and keeping regular types for other printers (like json, fixing the unmarshaling).

yanivagman:
[...]

PR #1170, opened 2021-11-29
add stdin file type in sched_process_exec

roikol:
adding an argument to sched_process_exec to indicate which file type is associated with stdin. will be of use to us in the rules


tracee-ebpf/tracee/consts.go
```diff
@@ -938,7 +944,7 @@ var EventsIDToParams = map[int32][]external.ArgMeta{
 	SysEnterEventID:               {{Type: "int", Name: "syscall"}},
 	SysExitEventID:                {{Type: "int", Name: "syscall"}},
 	SchedProcessForkEventID:       {{Type: "int", Name: "parent_tid"}, {Type: "int", Name: "parent_ns_tid"}, {Type: "int", Name: "child_tid"}, {Type: "int", Name: "child_ns_tid"}},
-	SchedProcessExecEventID:       {{Type: "const char*", Name: "cmdpath"}, {Type: "const char*", Name: "pathname"}, {Type: "const char**", Name: "argv"}, {Type: "const char**", Name: "env"}, {Type: "dev_t", Name: "dev"}, {Type: "unsigned long", Name: "inode"}, {Type: "int", Name: "invoked_from_kernel"}, {Type: "unsigned long", Name: "ctime"}},
+	SchedProcessExecEventID:       {{Type: "const char*", Name: "cmdpath"}, {Type: "const char*", Name: "pathname"}, {Type: "const char**", Name: "argv"}, {Type: "const char**", Name: "env"}, {Type: "dev_t", Name: "dev"}, {Type: "unsigned long", Name: "inode"}, {Type: "int", Name: "invoked_from_kernel"}, {Type: "unsigned long", Name: "ctime"}, {Type: "unsigned int", Name: "stdin"}},
```
itaysk, line 947: `{Type: "unsigned int", Name: "stdin"}` doesn't mean anything to anyone. can we make it an enum? the work in https://github.com/aquasecurity/libbpfgo/pull/89 might be related

itaysk:
seems like `sched_process_exec` is slowly collecting too many unrelated data. I don't have a good solution for this now, just making an observation.

tracee-ebpf/tracee/consts.go
roikol: thanks @itaysk .
i agree, it doesn't mean a lot. i did add an enum of the supported file types in consts.go, and i've added the ParseStdinType() func in argprinters.go.
i can change the name of the argument to something like "stdin_file_type", do you think that would help?
regarding the type "unsigned int" - i can't change that because this indicated what is being passed to userSpace from the bpf code.

should the argprinters func, ParseStdinType(), be part of libbpfgo? 
it seems to me like it doesn't belong there because we don't support all file types in tracee currently - so it's not a generic kernel thing yet.
wdyt?

roikol:
> seems like `sched_process_exec` is slowly collecting too many unrelated data. I don't have a good solution for this now, just making an observation.

yes, this event is getting too packed.
i can create a new event (something like: "stdin_on_exec") that will rely on 'sched_process_exec'. this way the user can choose whether to see this data, and 'sched_process_exec' won't have an extra argument.
do you think it's the better approach? @itaysk

tracee-ebpf/tracee/argprinters.go
```diff
@@ -115,6 +115,14 @@ func (t *Tracee) parseArgs(ctx *context, args map[string]interface{}) error {
 				args["type"] = typeIdStr
 			}
 		}
+	case SchedProcessExecEventID:
+		if stdin, isUint32 := args["stdin"].(uint32); isUint32 {
```
mtcherni95, line 119: If for any reason it is not uint32, don't we want to maybe notify or log something (maybe at debug level). It might help us in future if something changed between kernel versions, or between distributions, not critical - just an idea. What do you think?

tracee-ebpf/tracee/events_processor.go
```diff
@@ -246,6 +250,32 @@ func (t *Tracee) removeSocketTail() {
 	}
 }
 
+func (t *Tracee) removePipeTail() {
+	// removing sys_pipe_exit_tail from sys_exit_tails.
+	bpfMap, err := t.bpfModule.GetMap("sys_exit_tails")
```
mtcherni95, line 255: I suggest here to check if err != nil and return then continue with your logic. By the way, maybe we can add error return to the function?

tracee-ebpf/tracee/argprinters.go
roikol: actually this shouldn't change as these values are something we define. (see #define FILE_TYPE_SOCK in tracee.bpf.c).

tracee-ebpf/tracee/events_processor.go
roikol: because we are using sync.Once - we can't return a value from the function (as far as i understand it). the way to do that is to set a global variable to hold the value we want to return.
but i guess in this case we can just use the t.handleError().

tracee-ebpf/tracee/consts.go
itaysk: > i did add an enum of the supported file types in consts.go

You added some int consts, not what I meant. Go doesn't have a proper enums so I linked that libbpfgo PR as example of "enums" in go (basically, type + stringer)

> i can't change that because this indicated what is being passed to userSpace from the bpf code

Actually that's not true anymore. it's a hint to help deserialize the arg value as accurately as possible, see https://github.com/aquasecurity/tracee/issues/1029 for another example.

> should the argprinters func, ParseStdinType(), be part of libbpfgo?

Only if this is something that some syscall accept/returns

yanivagman (commented):
I'm asking myself if we really need all this logic just to know if a file is a pipe:
1. Set tail call for pipe
2. Save pipe file ops
3. Check if file ops equal

Why not instead just check if file->f_inode->i_mode has the I_FIFO bit set?
We can do the same for socket (S_IFSOCK in i_mode).

If this can work, it will save us a lot of logic and make things much more elegant

PR #1177, opened 2021-11-30
tracee-ebpf: rename co-re headers

yanivagman:


rafaeldtinoco (commented):
LGTM (although I think I prefer using vmlinux terminology, because of historical reasons and public documentation). WDYT ? The `co_re_missing_definitions.h` change is definitely needed.

yanivagman:
> LGTM (although I think I prefer using vmlinux terminology, because of historical reasons and public documentation). WDYT ? The co_re_missing_definitions.h change is definitely needed.

I kept the file name vmlinux.h exactly for this reason.
IMO using co-re as the directory name fits better for having missing_definitions.h in it, don't you think?

rafaeldtinoco (approved):
Yep, agreed. Let's do it then.

PR #1184, opened 2021-11-30
tracee-ebpf: keep deleted containers

### LINKED ISSUE #1175: empty containerId field in some container events
containerId field is sometimes empty although it shouldn't.
The cause for this issue is a cgroup_rmdir event that is received before other events of the container that were originated by another cpu.
Although the timestamp of cgroup_rmdir is bigger (later) than any of the other container events, it is received BEFORE some of these events due to a race condition between different cpus that use different perf buffers.

related: #1113

yanivagman:
fix #1175

### DIFF
--- a/tracee-ebpf/tracee/containers.go
+++ b/tracee-ebpf/tracee/containers.go
@@ -10,6 +10,7 @@ import (
 	"regexp"
 	"strings"
 	"syscall"
+	"time"
 	"unsafe"
 
 	bpf "github.com/aquasecurity/libbpfgo"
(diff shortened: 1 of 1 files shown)


grantseltzer:
Would the same thing be accomplished by using bpf ringbuffers?

yanivagman:
> Would the same thing be accomplished by using bpf ringbuffers

Yes. Ring buffer should resolve such issues in the future

tracee-ebpf/tracee/containers.go
```diff
@@ -216,14 +217,31 @@ check:
 }
 
 func (c *Containers) CgroupRemove(cgroupId uint64) {
-	delete(c.cgroups, uint32(cgroupId))
+	now := time.Now()
+	// prune containers that have been removed more than 5 seconds ago
+	var deleted []uint64
+	for _, id := range c.deleted {
```
rafaeldtinoco, line 223: I believe its okay to cleanup only when the next event happens (for our use case, which will always contain containers being created/delete). The other idea would be to have a parallel routine doing this check (which I don't think its worth as we're probably solving this with ring buffer approach).

LGTM

rafaeldtinoco (approved):

PR #1187, opened 2021-12-01
added Hidden inodes event

itamarmaouda101:
added a event for alerting about hidden process.
in case of inode number that is 0 in dir file of process in /proc - Glibc will ignore it and wont show it

the event is checks if the inode is zeroed and it is - its alerting in tracee console (or the asked output)

example of alerting in json format about porcess hidden with pid 1:

{"timestamp":1638302644352381023,"processId":4393,"threadId":4393,"parentProcessId":4286,"hostProcessId":4393,"hostThreadId":4393,"hostParentProcessId":4286,"userId":1000,"mountNamespace":4026531840,"pidNamespace":4026531836,"processName":"top","hostName":"itamar-VirtualB","containerId":"","eventId":"1032","eventName":"hidden_inodes","argsNum":1,"returnValue":0,"stackAddresses":null,"args":[{"name":"process_hidden","type":"char*","value":"1"}]}

### DIFF
--- a/tracee-ebpf/tracee/tracee.bpf.c
+++ b/tracee-ebpf/tracee/tracee.bpf.c
@@ -187,7 +187,8 @@ Copyright (C) Aqua Security inc.
 #define SECURITY_INODE_MKNOD            1029
 #define SECURITY_POST_READ_FILE         1030
 #define SOCKET_DUP                      1031
-#define MAX_EVENT_ID                    1032
+#define HIDDEN_INODES                   1032
+#define MAX_EVENT_ID                    1033
 
 #define NET_PACKET                      0
 #define DEBUG_NET_SECURITY_BIND         1
@@ -2550,7 +2551,24 @@ int tracepoint__sched__sched_switch(struct bpf_raw_tracepoint_args *ctx)
 
     return events_perf_submit(&data, SCHED_SWITCH, 0);
 }
+SEC("kprobe/filldir64")
+int BPF_KPROBE(trace_filldir64)
+{
+    event_data_t data = {};
+    if (!init_event_data(&data, ctx))
+        return 0;
+    if (!should_trace((&data.context)))
+        return 0;
 
+    char * process_name = (char *)PT_REGS_PARM2(ctx);
+    unsigned long process_inode_number = (unsigned long) PT_REGS_PARM5(ctx);
+    if (process_inode_number == 0)
+    {
+        save_str_to_buf(&data, process_name, 0);
+        return events_perf_submit(&data, HIDDEN_INODES, 0);
+    }
+    return 0;
+}
 SEC("kprobe/do_exit")
 int BPF_KPROBE(trace_do_exit)
 {
--- a/tracee-ebpf/tracee/events_config.go
+++ b/tracee-ebpf/tracee/events_config.go
@@ -69,6 +69,7 @@ const (
 	SecurityInodeMknodEventID
 	SecurityPostReadFileEventID
 	SocketDupEventID
+	HiddenInodesEventID
 	MaxCommonEventID
 )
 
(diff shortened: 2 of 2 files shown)


tracee-ebpf/tracee/consts.go
```diff
@@ -183,7 +183,9 @@ const (
 	SecurityInodeMknodEventID
 	SecurityPostReadFileEventID
 	SocketDupEventID
+	HiddenInodesEventID
 	MaxCommonEventID
+	MaxEventID
```
yanivagman, line 188: Please remove

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1604,7 +1605,15 @@ static __always_inline void* get_path_str(struct path *path)
     set_buf_off(STRING_BUF_IDX, buf_off);
     return &string_p->buf[buf_off];
 }
-
+static __inline int local_strncmp(const char* str1, const char* str2, int len)
+{
+    for (int i=0; i< len ; i++)
```
yanivagman, line 1610: You are not unrolling the loop - this won't work on kernels < 5.2

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2588,7 +2619,6 @@ int tracepoint__cgroup__cgroup_attach_task(struct bpf_raw_tracepoint_args *ctx)
     save_str_to_buf(&data, comm, 1);
     save_to_submit_buf(&data, (void*)&pid, sizeof(int), 2);
     events_perf_submit(&data, CGROUP_ATTACH_TASK, 0);
-
```
yanivagman, line 2591: No need to remove this line

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2550,7 +2559,29 @@ int tracepoint__sched__sched_switch(struct bpf_raw_tracepoint_args *ctx)
...
+    event_data_t data = {};
+    if (!init_event_data(&data, ctx))
+        return 0;
+    if (!should_trace((&data.context)))
+        return 0;
 
+    char* top = "top";
+    char* ps = "ps";
+    //check's if the called process is top or ps (to list processes)
+    if ((!local_strncmp(top, data.context.comm, 3)) && (!local_strncmp(ps, data.context.comm, 2)))
```
yanivagman, line 2574: We already have the "has_prefix" function - can't it be used instead local_strncmp that you added here?

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2550,7 +2559,29 @@ int tracepoint__sched__sched_switch(struct bpf_raw_tracepoint_args *ctx)
...
+int BPF_KPROBE(trace_filldir64)
+{
+    event_data_t data = {};
+    if (!init_event_data(&data, ctx))
+        return 0;
+    if (!should_trace((&data.context)))
+        return 0;
 
+    char* top = "top";
+    char* ps = "ps";
```
yanivagman, line 2572: So, if a process will use getdents directly or using another executable than these two, this event won't show, right?

It is ok to assume that one of these two will be called, but maybe we can do better? Here is a suggestion:
We can check instead that the directory being read is /proc by getting the current syscall (getdents) fd argument, then using get_struct_file_from_fd() to get the file struct of the directory, and extract the dir name out of it.

If you think this is too complex or unnecessary, you can ignore this comment.

tracee-ebpf/tracee/consts.go
itamarmaouda101: removed

tracee-ebpf/tracee/tracee.bpf.c
itamarmaouda101: i removed the entire function- see my last comment

tracee-ebpf/tracee/tracee.bpf.c
itamarmaouda101: ok i just rebased so i think it solved that

tracee-ebpf/tracee/tracee.bpf.c
itamarmaouda101: i removed the entire verification- see my last comment

tracee-ebpf/tracee/tracee.bpf.c
itamarmaouda101: i did another research about that an i saw it unnecessary to check that , due that inode numbers aren't supposed to be zeroed.
also i remember that i tried to hide files with that technique (out of /proc) and i saw it doesn't work for files out /proc.
i want to recheck that, hope ill find time for that . 
but for now i think its fine because i don't think there any known technique that hiding files (not process) with that.

also about what you wrote , i check the filldir function args and the /proc isnt exist there to we cant test if it there (if you meant to take that path from getdents it can work but i think its a bit more complex for now due that technique is use only for hide process)

itamarmaouda101: after ill finish #1195 and #1222 ill check up if there is to hide regular files with that. 
im remember my trying to implement POC for that but it wasn't work

tracee-ebpf/tracee/tracee.bpf.c
yanivagman: still removed

yanivagman (approved):
LGTM!

PR #1196, opened 2021-12-02
Feature/support 32 bit syscalls

### LINKED ISSUE #880: Support missing 32bit syscalls
I noticed that msgsnd and msgrcv syscalls are not captured when I run tracee-ebpf.
The exact command I run is:
```bash
sudo docker run --name tracee --rm --privileged -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro -v /tmp/tracee:/tmp/tracee -it aquasec/tracee:latest trace --security-alerts -t event=switch_task_ns,msgsnd,msgget,msgrcv -o option:detect-syscall -o option:exec-env
```
I suspect that I miss other syscalls like setns in other context.

AlonZivony:
Added support for unsupported syscalls and fixed matching between 64bit to 32bit syscalls

close #880


yanivagman:
Please rebase on recent main - libbpf submodule has changed

tracee-ebpf/tracee/consts_amd64.go
```diff
@@ -798,5 +891,12 @@ const (
 	sys32faccessat2                   int32 = 439
 	sys32process_madvise              int32 = 440
 	sys32epoll_pwait2                 int32 = 441
+	sys32mount_setattr                int32 = 442
+	sys32quotactl_fd                  int32 = 443
+	sys32landlock_create_ruleset      int32 = 444
+	sys32landlock_add_rule            int32 = 445
+	sys32landlock_restrict_self       int32 = 446
+	sys32memfd_secret                 int32 = 447
+	sys32process_mrelease             int32 = 448
```
yanivagman, line 900: These syscalls are new!
This means we should add them to the "regular" 64bit event ids as well.
On recent kernels (not sure on which kernel version it started), all the architectures get the same syscall numbers for new syscalls. Also see this comment- https://github.com/torvalds/linux/blob/master/arch/x86/entry/syscalls/syscall_64.tbl#L346

yanivagman (commented):
Overall looks good!
Two small comments

tracee-ebpf/tracee/consts_amd64.go
AlonZivony: I will create a new PR for the new syscalls and put it there

tracee-ebpf/tracee/consts.go
```diff
@@ -183,7 +183,7 @@ const (
 	SecurityInodeMknodEventID
 	SecurityPostReadFileEventID
 	SocketDupEventID
-	MaxEventID
+	MaxNonSyscallEventID
```
yanivagman, line 186: Not sure about the name.
Maybe let's call it MaxCommonEventID instead?

yanivagman (approved):

Issue #1171, opened 2021-11-29
tracee-ebpf: CO-RE is failing for kernels < 5.0


rafaeldtinoco:
When developing the slim version of vmlinux.h, and the fix for LINUX_VERSION_CODE dependency, @yanivagman and I found: https://github.com/aquasecurity/tracee/pull/1160#discussion_r758318322

This issue is for two needed things:

1. to make sure tracee is CO-RE capable for 4.18 and 4.19 kernels
2. to add missing 4.18 kernel version handling for get_task_ns_pid and its twin functions.

rafaeldtinoco:
$ cat log.txt | grep -B2 "no matching targets" | sed 's:--:```\n--\n```:g'
```
libbpf: prog 'tracepoint__raw_syscalls__sys_enter': relo #6: non-matching candidate #0 [149] struct task_struct (0 @ offset 0)
libbpf: prog 'tracepoint__raw_syscalls__sys_enter': relo #6: non-matching candidate #1 [26120] struct task_struct (0 @ offset 0)
libbpf: prog 'tracepoint__raw_syscalls__sys_enter': relo #6: no matching targets found
```
--
```
libbpf: prog 'tracepoint__raw_syscalls__sys_enter': relo #12: non-matching candidate #0 [149] struct task_struct (0 @ offset 0)
libbpf: prog 'tracepoint__raw_syscalls__sys_enter': relo #12: non-matching candidate #1 [26120] struct task_struct (0 @ offset 0)
libbpf: prog 'tracepoint__raw_syscalls__sys_enter': relo #12: no matching targets found
```
--
```
libbpf: prog 'tracepoint__raw_syscalls__sys_enter': relo #32: non-matching candidate #0 [1423] struct kernfs_node (0:9 @ offset 104)
libbpf: prog 'tracepoint__raw_syscalls__sys_enter': relo #32: non-matching candidate #1 [26624] struct kernfs_node (0:9 @ offset 104)
libbpf: prog 'tracepoint__raw_syscalls__sys_enter': relo #32: no matching targets found
```
--
```
libbpf: prog 'tracepoint__raw_syscalls__sys_exit': relo #8: non-matching candidate #0 [149] struct task_struct (0 @ offset 0)
libbpf: prog 'tracepoint__raw_syscalls__sys_exit': relo #8: non-matching candidate #1 [26120] struct task_struct (0 @ offset 0)
[...]

PR #1200, opened 2021-12-04
418corefixes

rafaeldtinoco:
The following comment: https://github.com/aquasecurity/tracee/issues/1171#issuecomment-985966743 summarizes the whys of this PR.
---

commit bf6ff44 (HEAD -> 418corefixes, rafaeldtinoco/418corefixes)
Date:   Fri Dec 3 11:27:09 2021

    tracee-ebpf: turn CO-RE v4.18 and beyond compatible

    Fixes: #1171

    For the VANILLA v4.18 (and beyond) kernels, we need the following 3
    commits backported, together with this patch, in order to have tracee
    eBPF object loaded correctly.

    commit 0962590e5533
    Author: Daniel Borkmann <daniel@iogearbox.net>
    Date:   Wed Oct 31 23:05:52 2018

        bpf: fix partial copy of map_ptr when dst is scalar

    commit e4f8d81c738d
    Author: Steven Rostedt (VMware) <rostedt@goodmis.org>
    Date:   Mon Jul 9 21:48:54 2018

        cgroup/tracing: Move taking of spin lock out of trace event handlers

    commit f7cf25b2026d
    Author: Alexei Starovoitov <ast@kernel.org>
    Date:   Sat Jun 15 12:12:17 2019 -0700

        bpf: track spill/fill of constants

    There is no v4.18 stable tree, so there is no way to propose those
    patches to upstream. Nevertheless, it is important to mention here so
    someone needing v4.18 kernel is able to compile it with the 3 mentioned
    patches - and possibly other dependencies - and this patch to tracee.

    The v4.19 (stable tree: linux-4.19.y branch) kernel already contains the
    needed fixes and doesn't need changes to have tracee CO-RE object
    running.
    
---

commit 902f242
Date:   Fri Dec 3 11:05:54 2021

    tracee-ebpf: comments for co-re type flavors

    - add comments to flavored types
    - adjust spaces & tabs according to project source style

---

commit 646dcf2
Date:   Fri Dec 3 10:58:04 2021

    tracee-ebpf: fix kernfs_node CORE access in RHEL8

    Fixes: #1171

    RHEL8 4.18 kernels have a different kernfs_node struct. You can see the
    name rh_kabi_hiddenXXX which is part of the RH KABI abstraction header
    (an attempt of RH to keep KABI compatible between their back-ports):

    struct kernfs_node {
            atomic_t count;
            atomic_t active;
            struct kernfs_node *parent;
            const char *name;
            struct rb_node rb;
            const void *ns;
            unsigned int hash;
            union {
                    struct kernfs_elem_dir dir;
                    struct kernfs_elem_symlink symlink;
                    struct kernfs_elem_attr attr;
            };
            void *priv;
            union {
                    u64 id;
                    struct {
                            union kernfs_node_id id;
                    } rh_kabi_hidden_172;
                    union {         };
            };
            short unsigned int flags;
            umode_t mode;
            struct kernfs_iattrs *iattr;
    };

    while the regular v4.18 struct is:

    struct kernfs_node {
            atomic_t count;
            atomic_t active;
            struct kernfs_node *parent;
            const char *name;
            struct rb_node rb;
            const void *ns;
            unsigned int hash;
            union {
                    struct kernfs_elem_dir dir;
                    struct kernfs_elem_symlink symlink;
                    struct kernfs_elem_attr attr;
            };
            void *priv;
            union kernfs_node_id id;
            short unsigned int flags;
            umode_t mode;
            struct kernfs_iattrs *iattr;
    };

    One way of dealing with this would be to pass to tracee.bpf the current
    running OS distro and kernel version and use branches to decide which
    struct flavor to use.. but we cannot do that because of the dead branch
    bug of v5.4 kernel.

    RHEL8 4.18 kernel is also very different from vanilla 4.18 kernel.

    Another example is the

    struct pid_link {
            struct hlist_node node;
            struct pid *pid;
    };

    not existing in RHEL8 and the task_struct of RHEL8 having:

    struct task_struct {
    ...
        struct pid *thread_pid;
    ...
    }

    Based on that, type the type/field checks are better than version checks
    because most of the times it will do the right thing (if the type we
    want exists we declare it as a flavor).

### DIFF
--- a/tracee-ebpf/tracee/co-re/struct_flavors.h
+++ b/tracee-ebpf/tracee/co-re/struct_flavors.h
@@ -3,6 +3,8 @@
 
 #pragma clang attribute push (__attribute__((preserve_access_index)), apply_to = record)
 
+// (struct kernfs_node *)->id was union kernfs_node_id before 5.5
+
 union kernfs_node_id {
     struct {
         u32 ino;
@@ -27,17 +29,19 @@ struct kernfs_node___rh8 {
     };
 };
 
+// commit bf9765145b85 ("sock: Make sk_protocol a 16-bit value")
+
 struct sock___old {
-    struct sock_common      __sk_common;
-    unsigned int            __sk_flags_offset[0];
-    unsigned int            sk_padding : 1,
-                            sk_kern_sock : 1,
-                            sk_no_check_tx : 1,
-                            sk_no_check_rx : 1,
-                            sk_userlocks : 4,
-                            sk_protocol  : 8,
-                            sk_type      : 16;
-    u16                     sk_gso_max_segs;
+    struct sock_common  __sk_common;
+    unsigned int        __sk_flags_offset[0];
+    unsigned int        sk_padding : 1,
+                        sk_kern_sock : 1,
+                        sk_no_check_tx : 1,
+                        sk_no_check_rx : 1,
+                        sk_userlocks : 4,
+                        sk_protocol  : 8,
+                        sk_type      : 16;
+    u16                 sk_gso_max_segs;
 };
 
 #pragma clang attribute pop
--- a/tracee-ebpf/tracee/co-re/struct_flavors.h
+++ b/tracee-ebpf/tracee/co-re/struct_flavors.h
@@ -44,6 +44,27 @@ struct sock___old {
     u16                 sk_gso_max_segs;
 };
 
+// support bpf_core_type_exists((task struct)->pids) for kernels < 5.0
+
+enum pid_type
+{
+    PIDTYPE_PID,
+    PIDTYPE_PGID,
+    PIDTYPE_SID,
+    PIDTYPE_MAX,
+    __PIDTYPE_TGID
+};
+
+struct pid_link
+{
+    struct hlist_node node;
+    struct pid *pid;
+};
+
+struct task_struct___older_v50 {
+    struct pid_link pids[PIDTYPE_MAX];
+};
+
 #pragma clang attribute pop
 
 #endif
--- a/tracee-ebpf/tracee/co-re/struct_flavors.h
+++ b/tracee-ebpf/tracee/co-re/struct_flavors.h
@@ -5,15 +5,26 @@
 
 union kernfs_node_id {
     struct {
-        u32     ino;
-        u32     generation;
+        u32 ino;
+        u32 generation;
     };
-    u64         id;
+    u64 id;
 };
 
-struct kernfs_node___old {
-    const char            *name;
-    union kernfs_node_id  id;
+struct kernfs_node___older_v55 {
+    const char *name;
+    union kernfs_node_id id;
+};
+
+struct kernfs_node___rh8 {
+    const char *name;
+    union {
+        u64 id;
+        struct {
+            union kernfs_node_id id;
+        } rh_kabi_hidden_172;
+        union { };
+    };
 };
 
 struct sock___old {
--- a/tracee-ebpf/tracee/tracee.bpf.c
+++ b/tracee-ebpf/tracee/tracee.bpf.c
@@ -635,6 +635,7 @@ static __always_inline u32 get_task_cgroup_ns_id(struct task_struct *task)
 
 static __always_inline u32 get_task_ns_pid(struct task_struct *task)
 {
+    int nr = 0;
     struct nsproxy *namespaceproxy = READ_KERN(task->nsproxy);
     struct pid_namespace *pid_ns_children = READ_KERN(namespaceproxy->pid_ns_for_children);
     unsigned int level = READ_KERN(pid_ns_children->level);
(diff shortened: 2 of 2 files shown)


tracee-ebpf/tracee/co-re/struct_flavors.h
```diff
@@ -3,30 +3,68 @@
...
+#include <vmlinux.h> // linting purposes
+
+// (struct kernfs_node *)->id was union kernfs_node_id before 5.5
+
 union kernfs_node_id {
     struct {
-        u32     ino;
-        u32     generation;
+	u32 ino;
+	u32 generation;
```
yanivagman, line 13: Please use spaces instead of tabs

tracee-ebpf/tracee/co-re/struct_flavors.h
```diff
@@ -3,30 +3,68 @@
 
 #pragma clang attribute push (__attribute__((preserve_access_index)), apply_to = record)
 
+#include <vmlinux.h> // linting purposes
```
yanivagman, line 6: Why do we neeed vmlinux.h here?
Doesn't vmlinux.h includes this header?

tracee-ebpf/tracee/co-re/struct_flavors.h
```diff
@@ -3,30 +3,68 @@
...
 };
 
+struct kernfs_node___older_rh8 {
+    const char *name;
+    union {
+	u64 id;
+	struct {
+	    union kernfs_node_id id;
+	} rh_kabi_hidden_172;
+	union { };
```
yanivagman, line 30: Please use spaces instead of tabs

tracee-ebpf/tracee/co-re/struct_flavors.h
```diff
@@ -3,30 +3,68 @@
...
 
-struct kernfs_node___old {
-    const char            *name;
-    union kernfs_node_id  id;
+struct kernfs_node___older_v55 {
+    const char *name;
+    union kernfs_node_id id;
 };
 
+struct kernfs_node___older_rh8 {
```
yanivagman, line 23: I would change kernfs_node___older_rh8 to kernfs_node___rh8 to avoid confusion

tracee-ebpf/tracee/co-re/struct_flavors.h
```diff
@@ -3,30 +3,68 @@
...
+    struct sock_common	__sk_common;
+    unsigned int	__sk_flags_offset[0];
+    unsigned int	sk_padding : 1,
+			sk_kern_sock : 1,
+			sk_no_check_tx : 1,
+			sk_no_check_rx : 1,
+			sk_userlocks : 4,
+			sk_protocol  : 8,
+			sk_type      : 16;
+    u16			sk_gso_max_segs;
```
yanivagman, line 46: Same about tabs

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -771,9 +798,23 @@ static __always_inline const u64 get_cgroup_id(struct cgroup *cgrp)
...
+
+        if (bpf_core_field_exists(kn_old_rh8->id)) {
+            // RHEL8 has both types declared: union and u64:
+            //     kn->id
+            //     rh->rh_kabi_hidden_172->id
+            // pointing to the same data
+            bpf_core_read(&id, sizeof(u64), &kn_old_rh8->id);
+        } else {
+            // all other regular kernels bellow v5.5
+            struct kernfs_node___older_v55 *kn_old = (void *)kn;
```
yanivagman, line 812: Isn't it the same as kn_old defined above in line 801?

tracee-ebpf/tracee/co-re/struct_flavors.h
rafaeldtinoco: sorry, vimrc didnt have the new filenames set.

tracee-ebpf/tracee/co-re/struct_flavors.h
rafaeldtinoco: we dont, it was for my linting to stop complaining on errors. removing.

tracee-ebpf/tracee/tracee.bpf.c
rafaeldtinoco: yep, thanks, fixed.

tracee-ebpf/tracee/co-re/struct_flavors.h
```diff
@@ -3,17 +3,30 @@
...
-        u32     generation;
-    };
-    u64         id;
+	struct {
+		u32 ino;
+		u32 generation;
+	};
+	u64 id;
+};
+
```
rafaeldtinoco, line 15: I'll fix this.

yanivagman (approved):

Issue #286, opened 2020-11-05
Drop unnecessary capabilities on start


yanivagman:
As a security best practice, we should drop all unnecessary capabilities on start.
For now, this means ALL but CAP_SYS_ADMIN (need to verify this).
Starting from kernel 5.8, there are new CAP_BPF+CAP_PERFMON which should be used instead. (https://lwn.net/Articles/822362/)

Update: Also see required capabilties at https://aquasecurity.github.io/tracee/v0.6.4/install/prerequisites/
Update2: CAP_NET_ADMIN is also required for tc programs (packet capture feature in tracee)

yanivagman:
We should also c

AlonZivony:
For receiving the namespaces of the host (the init_namespaces event) tracee also should have the CAP_PTRACE capability.

PR #1202, opened 2021-12-05
Feature/drop unnecessary capabilities

AlonZivony:
Added the functionality of dropping capabilities so only the necessary ones documented at the "Prerequisites" doc page will be left.

### DIFF
--- a/cmd/tracee-ebpf/capabilities.go
+++ b/cmd/tracee-ebpf/capabilities.go
@@ -1,0 +1,110 @@
+package main
+
+import (
+	"fmt"
+	"github.com/aquasecurity/tracee/pkg/capabilities"
+	tracee "github.com/aquasecurity/tracee/pkg/ebpf"
+	"github.com/syndtr/gocapability/capability"
+)
+
+// IKernelVersionInfo is an interface to check kernel version
+type IKernelVersionInfo interface {
+	// CompareOSBaseKernelRelease compare given kernel version to current one.
+	// The return value is -1, 0 or 1 if given version is less,
+	// equal or bigger, respectively, than running one.
+	CompareOSBaseKernelRelease(string) int
+}
+
+const bpfCapabilitiesMinKernelVersion = "5.8"
+
+// ensureCapabilities makes sure program runs with required capabilities only
+func ensureCapabilities(OSInfo IKernelVersionInfo, cfg *tracee.Config) error {
+	selfCap, err := capabilities.Self()
+	if err != nil {
+		return err
+	}
+
+	rCaps, err := generateTraceeEbpfRequiredCapabilities(OSInfo, cfg, selfCap)
+	if err != nil {
+		return err
+	}
+
+	if err = capabilities.CheckRequired(selfCap, rCaps); err != nil {
+		return err
+	}
+	if err = capabilities.DropUnrequired(selfCap, rCaps); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Get all capabilities required to run tracee-ebpf for current run
+func generateTraceeEbpfRequiredCapabilities(OSInfo IKernelVersionInfo, cfg *tracee.Config, selfCap capability.Capabilities) (
+	[]capability.Cap, error) {
+	rCaps, err := getCapabilitiesRequiredByEBPF(selfCap, OSInfo)
+	if err != nil {
+		return nil, err
+	}
+	rCaps = append(rCaps, getCapabilitiesRequiredByTraceeEvents(cfg)...)
+
+	rCaps = removeDupCaps(rCaps)
+	return rCaps, nil
+}
+
+func getCapabilitiesRequiredByTraceeEvents(cfg *tracee.Config) []capability.Cap {
+	usedEvents := cfg.Filter.EventsToTrace
+	for eventID := range tracee.CreateEssentialEventsList(cfg) {
+		usedEvents = append(usedEvents, eventID)
+	}
+	for eventID := range tracee.GetCaptureEventsConfig(cfg) {
+		usedEvents = append(usedEvents, eventID)
+	}
+	caps := tracee.GetCapabilitiesRequiredByEvents(usedEvents)
+
+	return removeDupCaps(caps)
+}
+
+// Get all capabilities required for eBPF usage (including perf buffers maps management)
+func getCapabilitiesRequiredByEBPF(selfCap capability.Capabilities, OSInfo IKernelVersionInfo) ([]capability.Cap, error) {
+	caps := []capability.Cap{
+		capability.CAP_IPC_LOCK,
+		capability.CAP_SYS_RESOURCE,
+	}
+	var versCaps []capability.Cap
+	if OSInfo.CompareOSBaseKernelRelease(bpfCapabilitiesMinKernelVersion) <= 0 {
+		versCaps = []capability.Cap{
+			capability.CAP_BPF,
+			capability.CAP_PERFMON,
+		}
+		if err1 := capabilities.CheckRequired(selfCap, versCaps); err1 != nil {
+			versCaps = []capability.Cap{
+				capability.CAP_SYS_ADMIN,
+			}
+			if err2 := capabilities.CheckRequired(selfCap, versCaps); err2 != nil {
+				return nil, fmt.Errorf("missing capabilites required for eBPF program loading - either CAP_BPF + CAP_PERFMON or CAP_SYS_ADMIN")
+			}
+		}
+	} else {
+		versCaps = []capability.Cap{
+			capability.CAP_SYS_ADMIN,
+		}
+	}
+	caps = append(caps, versCaps...)
+	return caps, nil
+}
+
+func removeDupCaps(dupCaps []capability.Cap) []capability.Cap {
+	capsMap := make(map[capability.Cap]bool)
+	for _, c := range dupCaps {
+		capsMap[c] = true
+	}
+	caps := make([]capability.Cap, len(capsMap))
+	i := 0
+	for c := range capsMap {
+		caps[i] = c
+		i++
+	}
+
+	return caps
+}
--- a/cmd/tracee-ebpf/main.go
+++ b/cmd/tracee-ebpf/main.go
@@ -144,6 +144,13 @@ func main() {
 			}
 			cfg.Output = &output
 
+			// environment capabilities
+
+			err = ensureCapabilities(OSInfo)
+			if err != nil {
+				return err
+			}
+
 			// kernel lockdown check
 			lockdown, err := helpers.Lockdown()
 			if err == nil && lockdown == helpers.CONFIDENTIALITY {
@@ -151,15 +158,6 @@ func main() {
 			}
 			if debug {
 				fmt.Fprintf(os.Stdout, "OSInfo: Security Lockdown is '%v'\n", lockdown)
-			}
-
-			// environment capabilities
-			selfCap, err := capabilities.Self()
-			if err != nil {
-				return err
-			}
-			if err = capabilities.CheckRequired(selfCap, []capability.Cap{capability.CAP_IPC_LOCK, capability.CAP_SYS_ADMIN}); err != nil {
-				return err
 			}
 
 			enabled, err := helpers.FtraceEnabled()
@@ -503,6 +501,51 @@ func checkCommandIsHelp(s []string) bool {
 		return true
 	}
 	return false
+}
+
+const bpfCapabilitiesMinKernelVersion = "5.8"
+
+// ensureCapabilities makes sure the program has just the required capabilities to run
+func ensureCapabilities(OSInfo *helpers.OSInfo) error {
+	selfCap, err := capabilities.Self()
+	if err != nil {
+		return err
+	}
+
+	// Build the set of capabilities required to run
+	rCaps := []capability.Cap{
+		capability.CAP_IPC_LOCK,
+		capability.CAP_SYS_PTRACE,
+		capability.CAP_SYS_RESOURCE,
+		capability.CAP_NET_ADMIN,
+	}
+	if OSInfo.CompareOSBaseKernelRelease(bpfCapabilitiesMinKernelVersion) <= 0 {
+		bpfCaps := []capability.Cap{
+			capability.CAP_BPF,
+			capability.CAP_PERFMON,
+		}
+		if err1 := capabilities.CheckRequired(selfCap, bpfCaps); err1 != nil {
+			bpfCaps = []capability.Cap{
+				capability.CAP_SYS_ADMIN,
+			}
+			if err2 := capabilities.CheckRequired(selfCap, bpfCaps); err2 != nil {
+				return fmt.Errorf("missing capabilites required for eBPF program loading - either CAP_BPF + CAP_PERFMON or CAP_SYS_ADMIN")
+			}
+		}
+		rCaps = append(rCaps, bpfCaps...)
+	} else {
+		rCaps = append(rCaps, []capability.Cap{
+			capability.CAP_SYS_ADMIN,
+		}...)
+	}
+
+	if err = capabilities.CheckRequired(selfCap, rCaps); err != nil {
+		return err
+	}
+	if err = capabilities.DropUnrequired(selfCap, rCaps); err != nil {
+		return err
+	}
+	return nil
 }
 
 func getFormattedEventParams(eventID int32) string {
--- a/cmd/tracee-ebpf/main.go
+++ b/cmd/tracee-ebpf/main.go
@@ -18,12 +18,10 @@ import (
 	"github.com/aquasecurity/libbpfgo/helpers"
 	embed "github.com/aquasecurity/tracee"
 	"github.com/aquasecurity/tracee/cmd/tracee-ebpf/internal/flags"
-	"github.com/aquasecurity/tracee/pkg/capabilities"
 	tracee "github.com/aquasecurity/tracee/pkg/ebpf"
 	"github.com/aquasecurity/tracee/pkg/metrics"
 	"github.com/aquasecurity/tracee/types/trace"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
-	"github.com/syndtr/gocapability/capability"
 	cli "github.com/urfave/cli/v2"
 )
 
@@ -145,8 +143,7 @@ func main() {
 			cfg.Output = &output
 
 			// environment capabilities
-
-			err = ensureCapabilities(OSInfo)
+			err = ensureCapabilities(OSInfo, &cfg)
 			if err != nil {
 				return err
 			}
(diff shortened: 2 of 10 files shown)


AlonZivony:
Currently I return error from main if the read of the current kernel version fails, because I can't predict which capabilities are required.
However, I am not sure it worth crushing because we can just leave all capabilities and we just risk with security issues.
Tell me what you think :)

AlonZivony:
solving #286

AlonZivony:
Notice that dropping the `--privileged` flag isn't possible yet because AppArmor will block the read of /proc/1/ns symlinks

mtcherni95 (commented):
I think we should do the same for tracee-rules, no? Even more, in tracee-rules we want all the capabilities to be dropped IMO.

mtcherni95:
> I think we should do the same for tracee-rules, no? Even more, in tracee-rules we want all the capabilities to be dropped IMO.

Just saw you are actually doing it, great :)

tracee-ebpf/main.go
```diff
@@ -955,6 +958,20 @@ func prepareEventsToTrace(eventFilter *tracee.StringFilter, setFilter *tracee.St
 	return res, nil
 }
 
+var necessaryCapabilities = [...]capability.Cap{capability.CAP_SYS_ADMIN, capability.CAP_IPC_LOCK, capability.CAP_SYS_PTRACE, capability.CAP_BPF, capability.CAP_PERFMON}
```
rafaeldtinoco: what about `neededCapabilities` instead ?

tracee-ebpf/main.go
```diff
@@ -955,6 +958,20 @@ func prepareEventsToTrace(eventFilter *tracee.StringFilter, setFilter *tracee.St
...
+
+func dropUnnecessaryCapabilities(caps capability.Capabilities) error {
```
rafaeldtinoco: `dropUnneededCapabilities()` ?

rafaeldtinoco: Note: you should stash your changes to the same logic in the same PR. One commit should contain either one feature or one fix. If you have multiple commits doing the same thing you should stash them together before the PR. 

Reason: this way you remove the burden of making the reviewer to review something about to be changed over and over in the same PR.

tracee-ebpf/main.go
```diff
@@ -92,16 +92,10 @@ func main() {
...
 
-			selfCap, err := getSelfCapabilities()
+			err = ensureCapabilities()
 			if err != nil {
 				return err
 			}
-			if err = checkRequiredCapabilities(selfCap); err != nil {
-				return err
-			}
-			if err = dropUnnecessaryCapabilities(selfCap); err != nil {
```
rafaeldtinoco: See ? I just reviewed this part and now its gone to somewhere else. (Thus the reason for the request of stashing commits that are playing the same cards).

tracee-ebpf/main.go
```diff
@@ -958,42 +952,50 @@ func prepareEventsToTrace(eventFilter *tracee.StringFilter, setFilter *tracee.St
...
+	if err = dropUnnecessaryCapabilities(selfCap, requiredCapabilities); err != nil {
+		return err
 	}
 	return nil
 }
 
-func checkRequiredCapabilities(caps capability.Capabilities) error {
-	if !caps.Get(capability.EFFECTIVE, capability.CAP_SYS_ADMIN) {
-		return fmt.Errorf("insufficient privileges to run: missing CAP_SYS_ADMIN")
+func dropUnnecessaryCapabilities(selfCaps capability.Capabilities, necessaryCapabilities []capability.Cap) error {
```
rafaeldtinoco, line 983: can't you call neededCapabilities to callees and callers ? You're calling requireCapabilities in some places, necessaryCapabilities in some others. Feels confusing to me.

tracee-ebpf/main.go
```diff
@@ -90,9 +90,24 @@ func main() {
...
+			} else if debug {
+				for k, v := range OSInfo.GetOSReleaseAllFieldValues() {
+					fmt.Fprintf(os.Stdout, "OSInfo: %v: %v\n", k, v)
+				}
+			}
+
 			// environment capabilities
 
-			err = ensureCapabilities()
+			err = ensureCapabilities(OSInfo)
```
rafaeldtinoco, line 110: Yep, this is better.

rafaeldtinoco (commented):
Before your changes:

```
=== RUN   Test_checkClangVersion
--- PASS: Test_checkClangVersion (0.00s)
=== RUN   Test_checkRequiredCapabilites
=== RUN   Test_checkRequiredCapabilites/happy_path
=== RUN   Test_checkRequiredCapabilites/CAP_SYS_ADMIN_missing
=== RUN   Test_checkRequiredCapabilites/CAP_IPC_LOCK_missing
--- PASS: Test_checkRequiredCapabilites (0.00s)
    --- PASS: Test_checkRequiredCapabilites/happy_path (0.00s)
    --- PASS: Test_checkRequiredCapabilites/CAP_SYS_ADMIN_missing (0.00s)
    --- PASS: Test_checkRequiredCapabilites/CAP_IPC_LOCK_missing (0.00s)
=== RUN   Test_fetchFormattedEventParams
--- PASS: Test_fetchFormattedEventParams (0.00s)
PASS
ok  	github.com/aquasecurity/tracee/tracee-ebpf	0.029s
```

After your changes:

```
=== RUN   Test_checkRequiredCapabilites
=== RUN   Test_checkRequiredCapabilites/happy_path
=== RUN   Test_checkRequiredCapabilites/CAP_SYS_ADMIN_missing
=== RUN   Test_checkRequiredCapabilites/CAP_IPC_LOCK_missing
=== RUN   Test_checkRequiredCapabilites/un-required_capability_missing
--- PASS: Test_checkRequiredCapabilites (0.00s)
    --- PASS: Test_checkRequiredCapabilites/happy_path (0.00s)
    --- PASS: Test_checkRequiredCapabilites/CAP_SYS_ADMIN_missing (0.00s)
    --- PASS: Test_checkRequiredCapabilites/CAP_IPC_LOCK_missing (0.00s)
    --- PASS: Test_checkRequiredCapabilites/un-required_capability_missing (0.00s)
```

Looks good.

tracee-ebpf/main.go
AlonZivony: You are right, it is confusing.
I think I will use required, because I thinks it describes the best that this is a setup for the code requirements.

tracee-ebpf/main.go
```diff
@@ -955,13 +953,53 @@ func prepareEventsToTrace(eventFilter *tracee.StringFilter, setFilter *tracee.St
 	return res, nil
 }
 
-func checkRequiredCapabilities(caps capability.Capabilities) error {
-	if !caps.Get(capability.EFFECTIVE, capability.CAP_SYS_ADMIN) {
-		return fmt.Errorf("insufficient privileges to run: missing CAP_SYS_ADMIN")
+const bpfCapabilitiesMinKernelVersion = "5.8"
+
+// ensureCapabilities makes sure the program has just the required capabilities to run
+func ensureCapabilities(OSInfo *helpers.OSInfo) error {
```
rafaeldtinoco, line 959: How do you feel about making this function more readable and easy to maintain with something like:

```Go
func ensureCapabilities(OSInfo *helpers.OSInfo) error {
        selfCap, err := getSelfCapabilities()
        if err != nil {
                return err
        }

        r := []capability.Cap {
            capability.CAP_IPC_LOCK,
            capability.CAP_SYS_PTRACE,
            capability.CAP_SYS_RESOURCE,
        }

        if OSInfo.CompareOSBaseKernelRelease("5.8") <= 0 {
                r = append(r, []capability.Cap{
                    capability.CAP_BPF,
                    capability.CAP_PERFMON,
                }...)
        } else {
                r = append(r, []capability.Cap{
                    capability.CAP_SYS_ADMIN,
                }...)
        }

        s := append(r, capability.CAP_SETPCAP)

        if err = checkRequiredCapabilities(selfCap, s); err != nil {
                return err
        }
        if err = dropUnrequiredCapabilities(selfCap, r); err != nil {
                return err
        }

        return nil
```

Long variable names won't always help reading the source code, despite telling what the variables are for (which can be followed by the function names being called).

Having each capability on its own line helps maintaining the code. Future changes/patches to this code can add/remove capabilities by simply adding/removing lines from it (instead of changing the entire line).

AlonZivony: About the lines breaks - I think you are more than right, this is much more readable.
About the names of the variables - I think that the roles of each of the startup and runtime capabilities are not so clear from the code itself, so calling them just 's' and 'r' is too short. However, the names I have given might be too long.
Is it ok with you that I will call them 'startup' and 'runtime'?

rafaeldtinoco: yep, agreed.

AlonZivony:
We had a discussion about the requirement of the CAP_SETPCAP capability.
It is used to lower the bounding capabilities set, which in turn effect the capabilities of the process upon `execve`.
However, it is only relevant if the executed file has file capability set. This is possible only if a process with the CAP_SETFCAP capability has set its capabilities, which is not a likely situation.
This rise the question - do we require this capability or only try lowering the bounding set and if we don't have the capability we fail and move on?

AlonZivony:
I ended up changing the commit log back to many commits because I needed to revert some commit I had.
I think it is more logical to only squash commits using the UI of Github before merging, because there is use for commits using the CR cicle.

yanivagman:
Also need to add CAP_NET_ADMIN required for tc programs (packet capture feature in tracee)

cmd/tracee-ebpf/main.go
```diff
@@ -985,13 +983,61 @@ func prepareEventsToTrace(eventFilter *tracee.StringFilter, setFilter *tracee.St
...
+		return err
+	}
+	if err = dropUnrequiredCapabilities(selfCap, rCaps); err != nil {
+		return err
+	}
+	return nil
+}
+
+// dropUnrequiredCapabilities requires that all capabilities are already set, or SYS_SETPCAP capability is set
+// The function also tries to drop the capabilities bounding set, but it won't work if SYS_SETPCAP is not available.
```
rafaeldtinoco, line 1023: I believe you meant CAP_SETPCAP here, not SYS_, right ? As I'm reading, CAP_SETPCAP might not be available IF file capabilities is, correct ? Should we inform the user of one or another or selfCaps.Apply will handle the error ?

cmd/tracee-rules/main.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/aquasecurity/tracee/tracee-rules/engine"
```
rafaeldtinoco, line 15: needs check

cmd/tracee-rules/main.go
```diff
@@ -214,3 +219,31 @@ func sigHandler() chan bool {
...
+
+	selfCaps.Clear(capability.CAPS | capability.BOUNDS)
+	err = selfCaps.Apply(capability.CAPS | capability.BOUNDS)
+	if err != nil {
+		return fmt.Errorf("couldn't drop capabilities: %v", err)
+	}
+	return nil
+}
+
+func getSelfCapabilities() (capability.Capabilities, error) {
```
rafaeldtinoco, line 239: Now that we decided to refactor during development (method suggested by @danielpacak), shouldn't we unify the "getSelfCapabilities()" functions, declared in tracee-ebpf/main.go AND tracee-rules/main.go, in a single place ?

rafaeldtinoco (commented):
Some comments bellow.

cmd/tracee-ebpf/main.go
```diff
@@ -985,13 +983,61 @@ func prepareEventsToTrace(eventFilter *tracee.StringFilter, setFilter *tracee.St
...
-	if !caps.Get(capability.EFFECTIVE, capability.CAP_IPC_LOCK) {
-		return fmt.Errorf("insufficient privileges to run: missing CAP_IPC_LOCK")
+	// Build the set of capabilities required to run
+	rCaps := []capability.Cap{
+		capability.CAP_IPC_LOCK,
+		capability.CAP_SYS_PTRACE,
+		capability.CAP_SYS_RESOURCE,
+		capability.CAP_NET_ADMIN,
+	}
+	if OSInfo.CompareOSBaseKernelRelease(bpfCapabilitiesMinKernelVersion) <= 0 {
```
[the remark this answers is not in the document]
AlonZivony: I used it to not use a magic string.
I think that "5.8" doesn't mean much for a normal person who looks at the code.
However, the fact that this is the kernel version of bpf capabilities is meaningful to every reader, and he won't need to research to understand why this comparison is made.
Has a best practice this is the way I use magic numbers and strings, but if the practices for this project are different I will change it.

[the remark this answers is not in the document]
rafaeldtinoco: No hard opinions on keeping it.

cmd/tracee-ebpf/main.go
AlonZivony: The use I do with CAP_SETPCAP is just a little extra I added to take advantage of extra capabilities to make the process even more secured. This capability is not a must-have, rather a nice-to-have. I can print a log that the capability wasn't present, but I don't think that it is really important to have.

cmd/tracee-rules/main.go
rafaeldtinoco: Maybe you should coordinate this change with https://github.com/aquasecurity/tracee/pull/1256#issuecomment-1005850610 and have your changes rebased on top of that PR after its finished. WDYT ?

rafaeldtinoco: OR you can just refactor these duplicated functions and merge this PR, and ask for @danielpacak to re-check that PR. Up to you guys.

yanivagman: @AlonZivony @simar7 which PR should be reviewed first? This one or #1256 ?

AlonZivony: > @AlonZivony @simar7 which PR should be reviewed first? This one or #1256 ?

I think @simar7 's should be first because he is introducing more generic change and it will be easy for me to rebase.on him

danielpacak: 👋 @AlonZivony We've merged #1256 so please go ahead with rebasing your changes.

cmd/tracee-ebpf/main.go
AlonZivony: I didn't add message to the user if the capability is not preset because I fear it might be a spam.
Tell me if you want me to add it either way.

rafaeldtinoco (changes requested):
@AlonZivony I have retried failed tests your changes are breaking the smoke tests and they seem to be passing for other ones (so its likely this PR's fault). Would you mind checking and fixing it ? Thanks!

AlonZivony:
No problem, I will take a look at it.

rafaeldtinoco (changes requested):
@AlonZivony is checking if, with the kallsyms change, this is still good. After done he will re-request my review here. Thanks!

rafaeldtinoco:
@AlonZivony you should rebase here as well.

AlonZivony:
@rafaeldtinoco after our conversation I added the new commit of the PR - dynamically determine what are the capabilities required by tracee-ebpf to require them and drop the others.
It is a bit complicated but after a conversation with Yaniv we concluded it is mandatory to achieve the security issue we currently have with capabilities properly.

rafaeldtinoco:
> @rafaeldtinoco after our conversation I added the new commit of the PR - dynamically determine what are the capabilities required by tracee-ebpf to require them and drop the others. It is a bit complicated but after a conversation with Yaniv we concluded it is mandatory to achieve the security issue we currently have with capabilities properly.

Okay, I'll go through this new approach and provide feedback here. Thanks for letting me know.

rafaeldtinoco (changes requested):
@AlonZivony could you rebase and add last commit proposed at https://github.com/AlonZivony/tracee/pull/2 if you agree. If not we can discuss the proposed changes here/there (mostly naming changes).

rafaeldtinoco:
@AlonZivony this needs rebasing. @yanivagman are you good with this approach (in order to merge this) ? My understanding this is that this is ready/close to be merged.

rafaeldtinoco (changes requested):
tests failing. we've agreed we could move and merge this (last time we spoke). let me know once tests are good so I can review and merge. Thanks.

rafaeldtinoco (approved):
LGTM

rafaeldtinoco:
It is likely that this merge is causing: https://github.com/aquasecurity/tracee/issues/1803

cmd/tracee-ebpf/capabilities.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/syndtr/gocapability/capability"
+)
+
+// IKernelVersionInfo is an interface to check kernel version
+type IKernelVersionInfo interface {
+	// CompareOSBaseKernelRelease compare given kernel version to current one.
+	// The return value is -1, 0 or 1 if given version is less,
+	// equal or bigger, respectively, than running one.
+	CompareOSBaseKernelRelease(string) int
+}
```
yanivagman, line 16: I don't think this should be in this file.
If it is required for the tests (mockOSinfo), maybe put it there?

cmd/tracee-ebpf/capabilities.go
```diff
@@ -0,0 +1,110 @@
...
+func removeDupCaps(dupCaps []capability.Cap) []capability.Cap {
+	capsMap := make(map[capability.Cap]bool)
+	for _, c := range dupCaps {
+		capsMap[c] = true
+	}
+	caps := make([]capability.Cap, len(capsMap))
+	i := 0
+	for c := range capsMap {
+		caps[i] = c
+		i++
```
yanivagman, line 106: You can just append to caps instead of using index i

pkg/ebpf/events_capabilities.go
```diff
@@ -0,0 +1,37 @@
...
+	reqCapabilities := make(map[capability.Cap]bool)
+	for _, e := range events {
+		addEventAndDependenciesCapabilities(e, reqCapabilities)
+	}
+
+	capList := make([]capability.Cap, len(reqCapabilities))
+	i := 0
+	for reqCap := range reqCapabilities {
+		capList[i] = reqCap
+		i++
```
yanivagman, line 17: append

pkg/ebpf/events_definitions.go
```diff
@@ -5659,7 +5675,8 @@ var EventsDefinitions = map[int32]EventDefinition{
 		Internal: true,
 		Probes:   []probe{},
 		Dependencies: dependencies{
-			events: []eventDependency{{eventID: SecuritySocketBindEventID}},
+			events:       []eventDependency{{eventID: SecuritySocketBindEventID}},
+			capabilities: []capability.Cap{capability.CAP_SYS_ADMIN},
```
yanivagman, line 5679: This should be CAP_NET_ADMIN and not CAP_SYS_ADMIN, shouldn't it?

pkg/ebpf/tracee.go
```diff
@@ -1287,3 +1274,32 @@ func (t *Tracee) updateKallsyms() {
 		t.kernelSymbols = kernelSymbols
 	}
 }
+
+// GetCaptureEventsConfig set events used to capture data
+func GetCaptureEventsConfig(cfg *Config) map[int32]eventConfig {
+	essentialEvents := make(map[int32]eventConfig)
```
yanivagman, line 1280: essentialEvents?

pkg/ebpf/tracee.go
```diff
@@ -1287,3 +1274,32 @@ func (t *Tracee) updateKallsyms() {
 		t.kernelSymbols = kernelSymbols
 	}
 }
+
+// GetCaptureEventsConfig set events used to capture data
+func GetCaptureEventsConfig(cfg *Config) map[int32]eventConfig {
```
yanivagman, line 1279: For code readability (knowing which events are chosen), I think it would be better to put this function near CreateEssentialEventsList. Even better - keep this code in one part (combine CreateEssentialEventsList and GetCaptureEventsConfig) as done at https://github.com/aquasecurity/tracee/pull/1791/files#diff-607a60afe9e4f5447f69749836213a1ee925bd27961145aa924c50a037f58a85R211

yanivagman (commented):
Sorry for the late review, but I just saw this one was merged and I wanted to go over it as well.
Overall, the change looks good to me, but there are a few comments that I left.
I also think that some capabilities are missing in some cases. The one I can think about is CAP_SYS_PTRACE (I think this is the right one) for capture executables and capture profile (required for accessing root fs of other processes in /proc)

cmd/tracee-ebpf/capabilities.go
AlonZivony: But I have to use it in this file as the type received in the functions.
I don't think that a code that is used by the program should be in a test file.

cmd/tracee-ebpf/capabilities.go
AlonZivony: This is a method to reduce runtime, because I created an array in the specific size of the number of unique capabilities.
You can argue that this is less readable, and I will change it.

pkg/ebpf/events_capabilities.go
AlonZivony: Again, I use make with defined length to reduce runtime (so I need to address the index in the defined array).

pkg/ebpf/tracee.go
AlonZivony: This is resolved.

AlonZivony (commented):
@yanivagman I fixed all problematic comments.
I didn't agree with some, you can respond if you don't agree.

PR #1205, opened 2021-12-05
fix help on output flags

AsafEitani:
in the help string it is mentioned that several output flags can be provided in sequence, specifically the out-file and err-file flags:
```

Examples:
  --output json                                            | output as json
  --output gotemplate=/path/to/my.tmpl                     | output as the provided go template
  --output out-file:/my/out err-file:/my/err               | output to /my/out and errors to /my/err
  --output none                                            | ignore events output

Use this flag multiple times to choose multiple output options

```
Although it states that the flag should be provided multiple times the example shows differently.

### DIFF
--- a/docs/tracee-ebpf/output.md
+++ b/docs/tracee-ebpf/output.md
@@ -39,5 +39,5 @@ output as the provided go template
 output to `/my/out` and errors to `/my/err`
 
 ```
---output out-file:/my/out err-file:/my/err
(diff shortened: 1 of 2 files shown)


itaysk (commented):
Can you also fix the docs please?

mtcherni95 (approved):
LGTM

yanivagman (approved):

PR #1213, opened 2021-12-06
get file types from inode struct instead of file_operations

roikol:
use the i_mode field to check if socket dup

### DIFF
--- a/tracee-ebpf/tracee/tracee.bpf.c
+++ b/tracee-ebpf/tracee/tracee.bpf.c
@@ -2226,11 +2226,11 @@ int syscall__execveat(void *ctx)
     return events_perf_submit(&data, SYS_EXECVEAT, 0);
 }
 
-static __always_inline int check_fd_socket(u64 fd)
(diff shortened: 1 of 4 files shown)


tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2217,83 +2226,31 @@ int syscall__execveat(void *ctx)
     return events_perf_submit(&data, SYS_EXECVEAT, 0);
 }
 
-SEC("raw_tracepoint/sys_socket")
-int sys_socket_exit_tail(void *ctx)
+static __always_inline int check_fd_socket(u64 fd)
```
yanivagman, line 2229: You can make it a more generic function `check_fd_type(u64 fd, u16 type)` where you can pass the requested type as an argument (socket/fifo...)

yanivagman (commented):
Thanks @roikol 
Much better solution than the previous one.

yanivagman (approved):

Issue #1223, opened 2021-12-07
Introduce a single top-level Go module for tracee


danielpacak:
Part of #1217

PR #1224, opened 2021-12-07
chore: use top-level Go module

danielpacak:
This patch is the first step towards conventional Go project
structure where we have just a single top-level Go module
and all shared code goes under the pkg/ directory.

Resolves: #1223


.github/workflows/pr-checker.yaml
```diff
@@ -0,0 +1,37 @@
...
+          sudo apt-get install --yes build-essential pkgconf libelf-dev llvm-12 clang-12
+          for tool in "clang" "llc" "llvm-strip"
+          do
+            sudo rm -f /usr/bin/$tool
+            sudo ln -s /usr/bin/$tool-12 /usr/bin/$tool
+          done
+          sudo curl -L -o /usr/bin/opa https://github.com/open-policy-agent/opa/releases/download/v0.35.0/opa_linux_amd64 && sudo chmod 755 /usr/bin/opa
+      - name: Build
+        run: make all
+      - name: Test tracee-ebpf
```
danielpacak, line 33: We're changing CI workflow to test compilation while @rafaeldtinoco is working on Makefiles refactor.

rafaeldtinoco: Yep, this allows us to have CI tests while changing Makefiles (as the only thing we need to guarantee on further commits is that "make all" and "make test" works).

rafaeldtinoco (commented):
LGTM. I'll let @itaysk to +1 here if he agrees.

.github/workflows/pr-checker.yaml
```diff
@@ -0,0 +1,37 @@
...
+        run: make -C tracee-ebpf/ test
```
rafaeldtinoco, line 34: We need root permissions here:

https://github.com/aquasecurity/tracee/issues/1225

danielpacak: I see now ... Fundamentally it's no longer a unit test. It's kind of an integration tests that runs along with other unit tests.

tracee-ebpf/tracee/user_mode_events.go
```diff
@@ -36,27 +36,33 @@ func CreateInitNamespacesEvent() (external.Event, error) {
...
+	return initNamespacesArgs, nil
 }
 
 // fetchInitNamespaces fetch the namespaces values from the /proc/1/ns directory
-func fetchInitNamespaces() map[string]uint32 {
+func fetchInitNamespaces() (map[string]uint32, error) {
 	initNamespacesMap := make(map[string]uint32)
 	namespaceValueReg := regexp.MustCompile(":[[[:digit:]]*]")
-	namespacesLinks, _ := ioutil.ReadDir(InitProcNsDir)
+	namespacesLinks, err := ioutil.ReadDir(InitProcNsDir)
```
danielpacak, line 57: In case of permission errors and running this code along with other unit tests we don't see why it's failing

mtcherni95: I suggest to skip the error, and create a GitHub issue for it, in general this test should be part of integration test. @simar7 FYI

mtcherni95 (commented):
I think this is a great initiative. It will help as well to unblock working on a refactor of tracee-ebpf (which I currently designing :) ).

LGTM!

tracee-ebpf/tracee/user_mode_events.go
danielpacak: Skipped and created #1228

.github/workflows/pr-checker.yaml
danielpacak: I decided to step back from new workflows in favour or fixing build in container (with DOCKER=1)

.github/workflows/pr-checker.yaml
danielpacak: No longer relevant in scope of this PR. Anyway we have #1228 to follow up

rafaeldtinoco (approved):
LGTM. BTW, what about the workflow changes we've done ? I'll need those for the Makefile changes (likely).

yanivagman (approved):

danielpacak:
> LGTM. BTW, what about the workflow changes we've done ? I'll need those for the Makefile changes (likely).

I decided to not modify GitHub Actions workflows in this patch. But I noted down the one we were playing with if you need them.

rafaeldtinoco:
> > LGTM. BTW, what about the workflow changes we've done ? I'll need those for the Makefile changes (likely).
> 
> I decided to not modify GitHub Actions workflows in this patch. But I noted down the one we were playing with if you need them.

I decided not to brake things for now as well. I'm developing the entire new "Makefile" logic by using "Makefile.new" files. This way we can have both Makefile building schemas working.

Nevertheless, I believe we should move our tests to what we've done yesterday (perhaps in a new PR). At least until this refactoring effort finishes. WDYT ? If you agree, would you mind doing a PR with only the workflow changes ? If not, thats ok also.

danielpacak:
> > > LGTM. BTW, what about the workflow changes we've done ? I'll need those for the Makefile changes (likely).
> > 
> > 
> > I decided to not modify GitHub Actions workflows in this patch. But I noted down the one we were playing with if you need them.
> 
> I decided not to brake things for now as well. I'm developing the entire new "Makefile" logic by using "Makefile.new" files. This way we can have both Makefile building schemas working.
> 
> Nevertheless, I believe we should move our tests to what we've done yesterday (perhaps in a new PR). At least until this refactoring effort finishes. WDYT ? If you agree, would you mind doing a PR with only the workflow changes ? If not, thats ok also.

Sure, I will open a new PR.

PR #1227, opened 2021-12-08
tracee-ebpf: a new Makefile

rafaeldtinoco:
**tracee: a new makefile is born**

Original Makefile will be kept until all targets, release process and
docker container functionalities are in place. The new Makefile(s) are
meant to be used by:

    $ make -f Makefile.new <target>

> **Note:** DO NOT mix usage of ./Makefile, ./tracee-ebpf/Makefile &
> ./tracee-rules/Makefile with ./Makefile.one. Make sure to wipe all
> "dist" directories before using Makefile.one.

The new Makefile targets are: (fix this)

    $ make all                  # build tracee-ebpf, tracee-rules & rules

    $ make env                  # show makefile environment/variables
    $ make reqs                 # verify build requirements

    # build

    $ make core                 # build ./dist/tracee.bpf.core.o
    $ make non-core             # build ./dist/tracee.bpf.XXX.o
    $ make tracee-ebpf          # build ./dist/tracee-ebpf
    $ make tracee-rules         # build ./dist/tracee-rules
    $ make rules                # build ./dist/rules

    # clean

    $ make clean                # wipe ./dist/
    $ make clean-core           # wipe ./dist/tracee.bpf.core.o
    $ make clean-non-core       # wipe ./dist/tracee.bpf.XXX.o
    $ make clean-tracee-ebpf    # wipe ./dist/tracee-ebpf
    $ make clean-tracee-rules   # wipe ./dist/tracee-rules
    $ make clean-rules          # wipe ./dist/rules

    # test

    $ make test                 # run all go & opa tests
    $ make tracee-ebpf-test     # go test tracee-ebpf
    $ make tracee-rules-test    # go test tracee-rules
    $ make test-rules           # opa test (tracee-rules)

----

**REASON**

The advantage of new Makefiles is that all the build targets have correct dependencies and new compilations only occur if you remove the targets OR change related source files.

Another advantage is to split the DOCKER building logic from the OS building one (further changes).

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -52,6 +52,13 @@ jobs:
         run: |
           make -C tracee-ebpf/ all DOCKER=1
           make -C tracee-ebpf/ test DOCKER=1
+      - name: Test Single Makefile tracee-ebpf
+        run: |
+          rm -rf tracee-ebpf/dist
+          rm -rf dist
+          make -f Makefile.one tracee-ebpf
+          make -f Makefile.one clean
+          make -f Makefile.one test-tracee-ebpf
   test-tracee-rules:
     name: Test tracee-rules
     needs:
@@ -73,6 +80,17 @@ jobs:
       - name: Test tracee-rules in Docker
         run: |
           make -C tracee-rules/ test DOCKER=1
+      - name: Test Single Makefile tracee-rules
+        run: |
+          rm -rf tracee-rules/dist
+          rm -rf dist
+          make -f Makefile.one tracee-rules
+          make -f Makefile.one clean
+          make -f Makefile.one rules
+          make -f Makefile.one clean
+          make -f Makefile.one test-tracee-rules
+          make -f Makefile.one clean
+          make -f Makefile.one test-rules
   release-snapshot:
     name: Release Snapshot
     needs:
--- a/Makefile.one
+++ b/Makefile.one
@@ -1,0 +1,468 @@
+.PHONY: all
+all:
+	$(MAKE) env
+	$(MAKE) reqs
+	$(MAKE) tracee-ebpf tracee-rules rules
+
+# make
+
+PARALLEL = $(shell $(CMD_GREP) -c ^processor /proc/cpuinfo)
+
+NORM_MAKE = make -j$(PARALLEL)
+
+MAKE = make -f Makefile.one
+MAKEFLAGS += --no-print-directory
+MAKEFLAGS += -j$(PARALLEL)
+
+# tools
+
+CMD_GIT ?= git
+CMD_CLANG ?= clang
+CMD_LLC ?= llc
+CMD_STRIP ?= llvm-strip
+CMD_RM ?= rm
+CMD_INSTALL ?= install
+CMD_MKDIR ?= mkdir
+CMD_PKGCONFIG ?= pkg-config
+CMD_GO ?= go
+CMD_GREP ?= grep
+CMD_OPA ?= opa # https://github.com/open-policy-agent/opa/releases/download/v0.33.1/opa_linux_amd64
+
+check_%:
+	@command -v $* >/dev/null || (echo "missing required tool $*" ; exit 1)
+
+# libs
+
+LIB_ELF ?= libelf
+LIB_ZLIB ?= zlib
+
+checklib_%:
+	@$(CMD_PKGCONFIG) --silence-errors --validate $* 2>/dev/null || \
+		(echo "missing lib $*" ; exit 1)
+
+# version
+
+LAST_GIT_TAG ?= $(shell $(CMD_GIT) describe --tags --match 'v*' 2>/dev/null)
+VERSION ?= $(if $(RELEASE_TAG),$(RELEASE_TAG),$(LAST_GIT_TAG))
+
+# environment
+
+UNAME_M := $(shell uname -m)
+UNAME_R := $(shell uname -r)
+
+ifeq ($(UNAME_M),x86_64)
+   ARCH = x86_64
+   LINUX_ARCH = x86
+   GO_ARCH = amd64
+endif
+
+ifeq ($(UNAME_M),aarch64)
+   ARCH = arm64
+   LINUX_ARCH = arm64
+   GO_ARCH = arm64
+endif
+
+.PHONY: env
+env:
+	@echo ---------------------------------------
+	@echo "Makefile Environment:"
+	@echo ---------------------------------------
+	@echo "PARALLEL                 $(PARALLEL)"
+	@echo ---------------------------------------
+	@echo "CMD_CLANG                $(CMD_CLANG)"
+	@echo "CMD_GIT                  $(CMD_GIT)"
+	@echo "CMD_GO                   $(CMD_GO)"
+	@echo "CMD_GREP                 $(CMD_GREP)"
+	@echo "CMD_INSTALL              $(CMD_INSTALL)"
+	@echo "CMD_LLC                  $(CMD_LLC)"
+	@echo "CMD_MKDIR                $(CMD_MKDIR)"
+	@echo "CMD_OPA                  $(CMD_OPA)"
+	@echo "CMD_PKGCONFIG            $(CMD_PKGCONFIG)"
+	@echo "CMD_RM                   $(CMD_RM)"
+	@echo "CMD_STRIP                $(CMD_STRIP)"
+	@echo ---------------------------------------
+	@echo "LIB_ELF                  $(LIB_ELF)"
+	@echo "LIB_ZLIB                 $(LIB_ZLIB)"
+	@echo ---------------------------------------
+	@echo "VERSION                  $(VERSION)"
+	@echo "LAST_GIT_TAG             $(LAST_GIT_TAG)"
+	@echo "BPF_NOCORE_TAG           $(BPF_NOCORE_TAG)"
+	@echo ---------------------------------------
+	@echo "UNAME_M                  $(UNAME_M)"
+	@echo "UNAME_R                  $(UNAME_R)"
+	@echo "ARCH                     $(ARCH)"
+	@echo "LINUX_ARCH               $(LINUX_ARCH)"
+	@echo ---------------------------------------
+	@echo "OUTPUT_DIR               $(OUTPUT_DIR)"
+	@echo ---------------------------------------
+	@echo "KERN_RELEASE             $(KERN_RELEASE)"
+	@echo "KERN_BUILD_PATH          $(KERN_BUILD_PATH)"
+	@echo "KERN_SRC_PATH            $(KERN_SRC_PATH)"
+	@echo ---------------------------------------
+	@echo "LIBBPF_CFLAGS            $(LIBBPF_CFLAGS)"
+	@echo "LIBBPF_LDLAGS            $(LIBBPF_LDFLAGS)"
+	@echo "LIBBPF_SRC               $(LIBBPF_SRC)"
+	@echo ---------------------------------------
+	@echo "STATIC                   $(STATIC)"
+	@echo ---------------------------------------
+	@echo "BPF_VCPU                 $(BPF_VCPU)"
+	@echo "TRACEE_EBPF_OBJ_SRC      $(TRACEE_EBPF_OBJ_SRC)"
+	@echo ---------------------------------------
+	@echo "GO_ARCH                  $(GO_ARCH)"
+	@echo "GO_TAGS_EBPF             $(GO_TAGS_EBPF)"
+	@echo "GO_TAGS_RULES            $(GO_TAGS_RULES)"
+	@echo ---------------------------------------
+	@echo "CUSTOM_CGO_CFLAGS        $(CUSTOM_CGO_CFLAGS)"
+	@echo "CUSTOM_CGO_LDFLAGS       $(CUSTOM_CGO_LDFLAGS)"
+	@echo "CGO_EXT_LDFLAGS_EBPF     $(CGO_EXT_LDFLAGS_EBPF)"
+	@echo "CGO_EXT_LDFLAGS_RULES    $(CGO_EXT_LDFLAGS_RULES)"
+	@echo ---------------------------------------
+	@echo "GO_ENV_EBPF              $(GO_ENV_EBPF)"
+	@echo "GO_ENV_RULES             $(GO_ENV_RULES)"
+	@echo ---------------------------------------
+	@echo "TRACEE_EBPF_SRC          $(TRACEE_EBPF_SRC)"
+	@echo ---------------------------------------
+	@echo "GOSIGNATURES_DIR         $(GOSIGNATURES_DIR)"
+	@echo "GOSIGNATURES_SRC         $(GOSIGNATURES_SRC)"
+	@echo ---------------------------------------
+	@echo "REGO_SIGNATURES_DIR      $(REGO_SIGNATURES_DIR)"
+	@echo "REGO_SIGNATURES_SRC      $(REGO_SIGNATURES_SRC)"
+	@echo ---------------------------------------
+
+# usage
+
+.PHONY: help
+help:
+	@echo ""
+	@echo "$ make env                  # show makefile environment/variables"
+	@echo "$ make reqs                 # verify build requirements"
+	@echo ""
+	@echo "# build"
+	@echo ""
+	@echo "$ make all                  # build tracee-ebpf, tracee-rules & rules"
+	@echo "$ make bpf-core             # build ./dist/tracee.bpf.core.o"
+	@echo "$ make bpf-nocore           # build ./dist/tracee.bpf.XXX.o"
+	@echo "$ make tracee-ebpf          # build ./dist/tracee-ebpf"
+	@echo "$ make tracee-rules         # build ./dist/tracee-rules"
+	@echo "$ make rules                # build ./dist/rules"
+	@echo ""
+	@echo "# clean"
+	@echo ""
+	@echo "$ make clean                # wipe ./dist/"
+	@echo "$ make clean-bpf-core       # wipe ./dist/tracee.bpf.core.o"
+	@echo "$ make clean-bpf-nocore     # wipe ./dist/tracee.bpf.XXX.o"
+	@echo "$ make clean-tracee-ebpf    # wipe ./dist/tracee-ebpf"
+	@echo "$ make clean-tracee-rules   # wipe ./dist/tracee-rules"
+	@echo "$ make clean-rules          # wipe ./dist/rules"
+	@echo ""
+	@echo "# test"
+	@echo ""
+	@echo "$ make test                 # run all go & opa tests"
+	@echo "$ make test-tracee-ebpf     # go test tracee-ebpf"
+	@echo "$ make test-tracee-rules    # go test tracee-rules"
+	@echo "$ make test-rules           # opa test (tracee-rules)"
+	@echo ""
+
+# variables
+
+BPF_VCPU = v2
+
+# requirements
+
+.PHONY: reqs_cmds
+reqs_cmds: \
+	check_$(CMD_GIT) \
+	check_$(CMD_CLANG) \
+	check_$(CMD_LLC) \
+	check_$(CMD_STRIP) \
+	check_$(CMD_RM) \
+	check_$(CMD_GREP) \
+	check_$(CMD_INSTALL) \
+	check_$(CMD_MKDIR) \
+	check_$(CMD_PKGCONFIG) \
+	check_$(CMD_GO) \
+	check_$(CMD_OPA)
+
+.PHONY: reqs_libs
+reqs_libs: \
+	checklib_$(LIB_ELF) \
+	checklib_$(LIB_ZLIB)
+
+.PHONY: reqs
+reqs:
+	$(MAKE) reqs_cmds
+	$(MAKE) reqs_libs
+
+# output dir
+
+OUTPUT_DIR = ./dist
+
+$(OUTPUT_DIR):
+	$(CMD_MKDIR) -p $@
+	$(CMD_MKDIR) -p $@/libbpf
+	$(CMD_MKDIR) -p $@/libbpf/obj
+
+# bundle
+
+.PHONY: $(OUTPUT_DIR)/tracee.bpf
+$(OUTPUT_DIR)/tracee.bpf:
+	$(CMD_MKDIR) -p $@
+	$(CMD_INSTALL) -m 0640 ./tracee-ebpf/3rdparty/include/* $@
+	$(CMD_INSTALL) -m 0640 $(OUTPUT_DIR)/libbpf/bpf/*.h $@
+	$(CMD_INSTALL) -m 0640 $(TRACEE_EBPF_OBJ_SRC) $@
+
+#
+# libbpf
+#
+
+LIBBPF_CFLAGS =
+LIBBPF_LDLAGS =
+LIBBPF_SRC = ./tracee-ebpf/3rdparty/libbpf/src
+
+$(OUTPUT_DIR)/libbpf/libbpf.a: $(LIBBPF_SRC) $(wildcard $(LIBBPF_SRC)/*.[ch]) | $(OUTPUT_DIR)
+	CC="$(CMD_CLANG)" \
+		CFLAGS="$(LIBBPF_CFLAGS)" \
+		LD_FLAGS="$(LIBBPF_LDFLAGS)" \
+		$(NORM_MAKE) -C $(LIBBPF_SRC) \
+		BUILD_STATIC_ONLY=1 \
+		DESTDIR=$(abspath ./$(OUTPUT_DIR)/libbpf/) \
+		OBJDIR=$(abspath ./$(OUTPUT_DIR)/libbpf/obj) \
+		INCLUDEDIR= LIBDIR= UAPIDIR= prefix= libdir= \
+		install install_uapi_headers
+
+$(LIBBPF_SRC):
+ifeq ($(wildcard $@), )
+	$(CMD_GIT) submodule update --init --recursive
+endif
+
+#
+# non co-re ebpf
+#
+
+TRACEE_EBPF_OBJ_SRC = ./tracee-ebpf/tracee/tracee.bpf.c
+
+KERN_RELEASE ?= $(UNAME_R)
+KERN_BUILD_PATH ?= $(if $(KERN_HEADERS),$(KERN_HEADERS),/lib/modules/$(KERN_RELEASE)/build)
+KERN_SRC_PATH ?= $(if $(KERN_HEADERS),$(KERN_HEADERS),$(if $(wildcard /lib/modules/$(KERN_RELEASE)/source),/lib/modules/$(KERN_RELEASE)/source,$(KERN_BUILD_PATH)))
+
+BPF_NOCORE_TAG = $(subst .,_,$(KERN_RELEASE)).$(subst .,_,$(VERSION))
+
+.PHONY: bpf-nocore
+bpf-nocore: $(OUTPUT_DIR)/tracee.bpf.$(BPF_NOCORE_TAG).o
+
+$(OUTPUT_DIR)/tracee.bpf.$(BPF_NOCORE_TAG).o: $(OUTPUT_DIR)/libbpf/libbpf.a \
+	$(TRACEE_EBPF_OBJ_SRC)
+	$(MAKE) $(OUTPUT_DIR)/tracee.bpf
+	$(CMD_CLANG) -S -nostdinc \
+		-D__TARGET_ARCH_$(LINUX_ARCH) \
+		-D__BPF_TRACING__ \
+		-D__KERNEL__ \
+		-include $(KERN_SRC_PATH)/include/linux/kconfig.h \
+		-I $(KERN_SRC_PATH)/arch/$(LINUX_ARCH)/include \
+		-I $(KERN_SRC_PATH)/arch/$(LINUX_ARCH)/include/uapi \
+		-I $(KERN_BUILD_PATH)/arch/$(LINUX_ARCH)/include/generated \
+		-I $(KERN_BUILD_PATH)/arch/$(LINUX_ARCH)/include/generated/uapi \
+		-I $(KERN_SRC_PATH)/include \
+		-I $(KERN_BUILD_PATH)/include \
+		-I $(KERN_SRC_PATH)/include/uapi \
+		-I $(KERN_BUILD_PATH)/include/generated \
+		-I $(KERN_BUILD_PATH)/include/generated/uapi \
+		-I$(OUTPUT_DIR)/tracee.bpf \
+		-Wunused \
+		-Wall \
+		-Wno-frame-address \
+		-Wno-unused-value \
+		-Wno-unknown-warning-option \
+		-Wno-pragma-once-outside-header \
+		-Wno-pointer-sign \
+		-Wno-gnu-variable-sized-type-not-at-end \
+		-Wno-deprecated-declarations \
+		-Wno-compare-distinct-pointer-types \
+		-Wno-address-of-packed-member \
+		-fno-stack-protector \
+		-fno-jump-tables \
+		-fno-unwind-tables \
+		-fno-asynchronous-unwind-tables \
+		-xc -O2 -g -emit-llvm \
+		-c $(TRACEE_EBPF_OBJ_SRC) \
+		-o $(@:.o=.ll)
+	$(CMD_LLC) \
+		-march=bpf -mcpu=$(BPF_VCPU) \
+		-filetype=obj \
+		-o $@ \
+		$(@:.o=.ll)
+	$(CMD_RM) $(@:.o=.ll)
+
+.PHONY: clean-bpf-nocore
+clean-bpf-nocore:
+	$(CMD_RM) -rf $(OUTPUT_DIR)/tracee.bpf.$(BPF_NOCORE_TAG).o
+
+#
+# co-re ebpf
+#
+
+TRACEE_EBPF_OBJ_CORE_HEADERS = $(shell find tracee-ebpf -name *.h ! -iregex '.*3rdparty.*')
+
+.PHONY: bpf-core
+bpf-core: $(OUTPUT_DIR)/tracee.bpf.core.o
+
+$(OUTPUT_DIR)/tracee.bpf.core.o: $(OUTPUT_DIR)/libbpf/libbpf.a \
+	$(TRACEE_EBPF_OBJ_SRC) \
+	$(TRACEE_EBPF_OBJ_CORE_HEADERS)
+	$(MAKE) $(OUTPUT_DIR)/tracee.bpf
+	$(CMD_CLANG) \
+		-D__TARGET_ARCH_$(LINUX_ARCH) \
+		-D__BPF_TRACING__ \
+		-DCORE \
+		-I./tracee-ebpf/tracee/co-re/ \
+		-I$(OUTPUT_DIR)/tracee.bpf \
+		-target bpf \
+		-O2 -g \
+		-target bpf -march=bpf -mcpu=$(BPF_VCPU) \
+		-c $(TRACEE_EBPF_OBJ_SRC) \
+		-o $@
+
+.PHONY: clean-bpf-core
+clean-bpf-core:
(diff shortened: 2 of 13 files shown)


rafaeldtinoco:
@danielpacak if we use the simple workflow from your PR, then this Makefile change would pass tests. Could you check this new Makefile "idea" and see if there is any objection ? (The docker work will be done in another Makefile, the old Makefile still exists so we dont lose the ability of using old targets if needed).

itaysk:
Do you think that we need separate makefiles for tracee-ebpf and tracee-rules given we no longer going to have this separation between the projects in the repo?

rafaeldtinoco:
> Do you think that we need separate makefiles for tracee-ebpf and tracee-rules given we no longer going to have this separation between the projects in the repo?

No, I don't. My idea was "to follow" the dir structure changes. Would you prefer me to concentrate everything in a single Makefile in root dir from the beginning ? I can certainly do that.

itaysk:
whatever works for you :) follow the repo also sounds good

tracee-ebpf/Makefile
```diff
@@ -61,6 +61,10 @@ build: $(OUT_BIN)
 go_env := GOOS=linux GOARCH=$(ARCH:x86_64=amd64) CC=$(CMD_CLANG) CGO_CFLAGS="-I $(abspath $(LIBBPF_HEADERS))" CGO_LDFLAGS="$(abspath $(LIBBPF_OBJ))"
 ifndef DOCKER
 $(OUT_BIN): $(LIBBPF_HEADERS) $(LIBBPF_OBJ) $(OUT_BPF_CORE) $(filter-out *_test.go,$(GO_SRC)) $(BPF_BUNDLE) | $(OUT_DIR)
+	# workaround until Makefile.one becomes standard
```
rafaeldtinoco, line 64: This is needed because embedded.go is at the root directory, embedding dist there, and if we build with regular existing Makefiles, then we have to copy the entire content to it for the embed to work.

I don't want to touch existing Makefile very much, and, with this small workaround we can maintain both working until Makefile.one becomes the standard Makefile.

rafaeldtinoco:
@danielpacak

Okay. I'm satisfied with this one AND nothing was broken (so we can merge this back to main in the beginning of next week). I'm going to work in a "docker-make" now (automated make with docker) and then the release container generation.

rafaeldtinoco:
The Dockerfile.tracee-make is the base for the "tracee-make" command/container I'll use for the docker targets. It is able to use Makefile.one entirely (and tests).

Makefile.one
```diff
@@ -0,0 +1,442 @@
...
+	@echo "$ make clean-non-core       # wipe ./dist/tracee.bpf.XXX.o"
+	@echo "$ make clean-tracee-ebpf    # wipe ./dist/tracee-ebpf"
+	@echo "$ make clean-tracee-rules   # wipe ./dist/tracee-rules"
+	@echo "$ make clean-rules          # wipe ./dist/rules"
+	@echo ""
+	@echo "# test"
+	@echo ""
+	@echo "$ make test                 # run all go & opa tests"
+	@echo "$ make tracee-ebpf-test     # go test tracee-ebpf"
+	@echo "$ make tracee-rules-test    # go test tracee-rules"
```
danielpacak, line 160:
```suggestion
	@echo "$ make test-tracee-ebpf     # go test tracee-ebpf"
	@echo "$ make test-tracee-rules    # go test tracee-rules"
```

Makefile.one
```diff
@@ -0,0 +1,442 @@
...
+	check_$(CMD_RM) \
+	check_$(CMD_GREP) \
+	check_$(CMD_INSTALL) \
+	check_$(CMD_MKDIR) \
+	check_$(CMD_PKGCONFIG) \
+	check_$(CMD_GO) \
+	check_$(CMD_OPA)
+
+.PHONY: reqs_libs
+reqs_libs: \
```
danielpacak, line 185: When I run this target in the Vagrant VM I'm getting the following warning:

```
make -f Makefile.one reqs_libs
/usr/lib/x86_64-linux-gnu/pkgconfig/zlib.pc:9: warning: malformed version field with whitespace, trimming to [1.2.11]
/usr/lib/x86_64-linux-gnu/pkgconfig/libelf.pc:8: warning: malformed version field with whitespace, trimming to [0.176]
```

Is it expected?

Makefile.one
```diff
@@ -0,0 +1,442 @@
...
+
+# rules
+
+GOSIGNATURES_DIR ?= tracee-rules/signatures/golang
+GOSIGNATURES_SRC := $(shell find $(GOSIGNATURES_DIR) \
+		    -type f \
+		    -name '*.go' \
+		    ! -name '*_test.go' \
+		    ! -path '$(GOSIGNATURES_DIR)/examples/*' \
+		    )
```
danielpacak, line 401: NIT: We have tabs and spaces around these lines

danielpacak (changes requested):
👋 @rafaeldtinoco This is amazing work. I tested new Make targets in the Vagrant VM (Ubuntu 20.4) and everything works like a charm.

Before we merge we must rebase with the [next] branch. (Since we've already merged changes done by @simar7 to move rules to the top-level directory some paths in Makefile.one must be adjusted as well. I can help with that.)

Also, do you think that we should add `make -f Makefile.one all` and `make -f Makefile.one test` targets to the CI workflow?

[next]: https://github.com/aquasecurity/tracee/tree/next

Makefile.one
rafaeldtinoco: I'll add "--silence-errors" arguments to pkg-config. This error is from your "distro" not versioning libs right (as it seems).

Makefile.one
```diff
@@ -0,0 +1,442 @@
...
+
+LIB_ELF ?= libelf
+LIB_ZLIB ?= zlib
+
+checklib_%:
+	@$(CMD_PKGCONFIG) --silence-errors --validate $* 2>/dev/null || (echo "missing lib $*" ; exit 1)
+
+# version
+
+LAST_GIT_TAG ?= $(shell $(CMD_GIT) describe --tags --exclude 'tracee-rules/*' --exclude 'tracee-ebpf/*' 2>/dev/null)
```
itaysk, line 44: maybe a safer way is to `--include 'v*'`

Makefile.one
```diff
@@ -0,0 +1,443 @@
...
+endif
+
+ifeq ($(UNAME_M),aarch64)
+   ARCH = arm64
+   LINUX_ARCH = arm64
+   GO_ARCH = arm64
+endif
+
+.PHONY: env
+env:
```
itaysk, line 65: instead of manually printing every variable, we can use the `.VARIABLES` built-in variable that contains all variable names

embedded-ebpf.go
```diff
@@ -0,0 +1,12 @@
+//go:build ebpf
+// +build ebpf
+
+package tracee
+
+import (
+	"embed"
+)
+
+//go:embed "dist/tracee.bpf/*"
```
rafaeldtinoco, line 10: This is needed if we want to have independent compilations of "tracee-ebpf" and "tracee-rules" (and we do). That is why I created the "ebpf" go tag.

Makefile.one
```diff
@@ -0,0 +1,443 @@
...
+
+.PHONY: clean-non-core
+clean-non-core:
+	$(CMD_RM) -rf $(OUTPUT_DIR)/tracee.bpf.$(NONCORE_TAG).o
+
+#
+# co-re ebpf
+#
+
+TRACEE_EBPF_OBJ_SRC = ./tracee-ebpf/tracee/tracee.bpf.c
```
itaysk, line 297: `TRACEE_EBPF_SRC`? OBJ might be misleading here no?

Makefile.one
```diff
@@ -0,0 +1,443 @@
...
+OUTPUT_DIR = ./dist
+
+$(OUTPUT_DIR):
+	$(CMD_MKDIR) -p $@
+	$(CMD_MKDIR) -p $@/libbpf
+	$(CMD_MKDIR) -p $@/libbpf/obj
+
+# bundle
+
+$(OUTPUT_DIR)/tracee.bpf: $(OUTPUT_DIR)/libbpf/libbpf.a
```
itaysk, line 205: missing requirement `TRACEE_EBPF_OBJ_SRC`

rafaeldtinoco:
@danielpacak 
Alright, needed a few changes after rebase but I think we're all set now.

Makefile.one
itaysk: should we move this upper since it's used be previous target?

Makefile.one
```diff
@@ -0,0 +1,443 @@
...
+#
+
+KERN_RELEASE ?= $(UNAME_R)
+KERN_BUILD_PATH ?= $(if $(KERN_HEADERS),$(KERN_HEADERS),/lib/modules/$(KERN_RELEASE)/build)
+KERN_SRC_PATH ?= $(if $(KERN_HEADERS),$(KERN_HEADERS),$(if $(wildcard /lib/modules/$(KERN_RELEASE)/source),/lib/modules/$(KERN_RELEASE)/source,$(KERN_BUILD_PATH)))
+
+NONCORE_TAG = $(subst .,_,$(KERN_RELEASE)).$(subst .,_,$(VERSION))
+
+.PHONY: non-core
+non-core: $(OUTPUT_DIR)/tracee.bpf.$(NONCORE_TAG).o
```
itaysk, line 246: can we give this target a more meaningfull name like `bpf-nocore`

Makefile.one
```diff
@@ -0,0 +1,443 @@
...
+	$(CMD_RM) -rf $(OUTPUT_DIR)/tracee.bpf.$(NONCORE_TAG).o
+
+#
+# co-re ebpf
+#
+
+TRACEE_EBPF_OBJ_SRC = ./tracee-ebpf/tracee/tracee.bpf.c
+
+.PHONY: core
+core: $(OUTPUT_DIR)/tracee.bpf.core.o
```
itaysk, line 300: can we give this target a more meaningfull name like `bpf`?

Makefile.one
```diff
@@ -0,0 +1,443 @@
...
+
+$(OUTPUT_DIR)/tracee.bpf.core.o: $(OUTPUT_DIR)/tracee.bpf $(TRACEE_EBPF_OBJ_SRC)
+	$(CMD_CLANG) \
+		-D__TARGET_ARCH_$(LINUX_ARCH) \
+		-D__BPF_TRACING__ \
+		-DCORE \
+		-I./tracee-ebpf/tracee/co-re/ \
```
itaysk, line 307: should we include this in the `tracee.bpf` dir?

Makefile.one
```diff
@@ -0,0 +1,443 @@
...
+KERN_RELEASE ?= $(UNAME_R)
+KERN_BUILD_PATH ?= $(if $(KERN_HEADERS),$(KERN_HEADERS),/lib/modules/$(KERN_RELEASE)/build)
+KERN_SRC_PATH ?= $(if $(KERN_HEADERS),$(KERN_HEADERS),$(if $(wildcard /lib/modules/$(KERN_RELEASE)/source),/lib/modules/$(KERN_RELEASE)/source,$(KERN_BUILD_PATH)))
+
+NONCORE_TAG = $(subst .,_,$(KERN_RELEASE)).$(subst .,_,$(VERSION))
+
+.PHONY: non-core
+non-core: $(OUTPUT_DIR)/tracee.bpf.$(NONCORE_TAG).o
+
+$(OUTPUT_DIR)/tracee.bpf.$(NONCORE_TAG).o: $(OUTPUT_DIR)/tracee.bpf $(TRACEE_EBPF_OBJ_SRC)
```
itaysk, line 248: missing requirements for all the headers

Makefile.one
```diff
@@ -0,0 +1,443 @@
...
+			-X main.version=\"$(VERSION)\" \
+			" \
+		-v -o $@ ./tracee-ebpf
+
+.PHONY: clean-tracee-ebpf
+clean-tracee-ebpf:
+	$(CMD_RM) -rf $(OUTPUT_DIR)/tracee-ebpf
+
+.PHONY: test-tracee-ebpf
+test-tracee-ebpf: tracee-ebpf
```
itaysk, line 361: why would the test need the binary?

itaysk: missing requriement of test source files

Makefile.one
```diff
@@ -0,0 +1,443 @@
...
+#
+# tracee-rules
+#
+
+GO_ENV_RULES =
+GO_ENV_RULES += GOOS=linux
+GO_ENV_RULES += CC=$(CMD_CLANG)
+GO_ENV_RULES += GOARCH=$(GO_ARCH)
+
+TRACEE_RULES_SRC=$(shell find ./tracee-rules/ \
```
itaysk, line 375: can we make this similar to the ebpf source finder: `find ./tracee-ebpf/ -name '*.go' ! -name '*_test.go'`

Makefile.one
```diff
@@ -0,0 +1,443 @@
...
+
+$(OUTPUT_DIR)/tracee-rules: $(TRACEE_RULES_SRC) | $(OUTPUT_DIR)
+	$(GO_ENV_RULES) $(CMD_GO) build -v -o $@ ./tracee-rules
+
+.PHONY: clean-tracee-rules
+clean-tracee-rules:
+	$(CMD_RM) -rf $(OUTPUT_DIR)/tracee-rules
+
+.PHONY: test-tracee-rules
+test-tracee-rules: tracee-rules rules
```
itaysk, line 391: why would the test need the binary?

itaysk: given we've moved the rules to root, rules won't be tested here, so no need to require them. Either way the current code excludes test files from "rules" so it not useful here anyway

Makefile.one
rafaeldtinoco: TRACEE_EBPF_SRC is already all files that, if changed, will cause the target to need rebuild. Moving up to its closer to its first use, indeed.

Makefile.one
rafaeldtinoco: I'm using concatenated requirements to make sure "make" will always do things in order and have the target dependencies built (or not) only if needed (that was the main problem in the original Makefile).

The bundle depends in libbpf.a only (if it has changed, then the files within the bundle might have changed as well, then I need to copy them again).

Am I missing something ?

Makefile.one
rafaeldtinoco: bpf-nocore

Makefile.one
rafaeldtinoco: bpf-core

Makefile.one
rafaeldtinoco: We're not rebuilding CO-RE src from the bundle, during runtime, so I guess not. WDYT ?

Makefile.one
rafaeldtinoco: Yep, indeed. Nice catch.

Makefile.one
rafaeldtinoco: It does not necessarily need the binary, I guess I can remove that need. Will add the test source files as requirement.

Makefile.one
rafaeldtinoco: --match 'v*' works.

Makefile.one
rafaeldtinoco: I tried, and, if you check something like:
```
    $(foreach var,$(.VARIABLES),$(info $(var) = $($(var))))
```
it is **VERY chaotic**. The .VARIABLES built-in has a bunch of stuff we don't care.

I wanted something that would tell us, in a human consumable way, the major VARIABLES and their values in the workflows output.

Makefile.one
rafaeldtinoco: Actually, only for CORE headers. The non-CORE headers are system headers and I don't think we should list those as hard target build dependencies. Fixing CORE ones.

Dockerfile.tracee-make
```diff
@@ -0,0 +1,61 @@
...
+    https://github.com/open-policy-agent/opa/releases/download/v0.33.1/opa_linux_amd64_static && \
+    chmod 755 /usr/bin/opa
+
+RUN export uid=$uid gid=$gid && \
+    mkdir -p /home/tracee && \
+    echo "tracee:x:${uid}:${gid}:Tracee,,,:/home/tracee:/bin/bash" >> /etc/passwd && \
+    echo "tracee:x:${gid}:" >> /etc/group && \
+    echo "tracee ALL=(ALL) NOPASSWD: ALL" > /etc/sudoers.d/tracee && \
+    chmod 0440 /etc/sudoers.d/tracee && \
+    chown ${uid}:${gid} -R /home/tracee && \
```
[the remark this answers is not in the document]
rafaeldtinoco: Why would this brake the case you brought ? I'm just making sure that the tracee-make container uses the same UID as your host environment (because of generated files permissions). This way, all generated files will have same UID/GID as they would have in the host env (which is something we want).

Let me know if you find a use case we are breaking, I would like to consider it.

Dockerfile.tracee-make
```diff
@@ -0,0 +1,61 @@
...
+#
+# To use 'tracee-make' as replacement for the make tool:
+#
+# docker run --rm --pid=host --privileged \
+#       -v $(pwd):/tracee \
+#       -it tracee-make \
+#       make -f Makefile.one [clean|core|all|test|...]
+#
+
+FROM alpine:3.15 as tracee-make
```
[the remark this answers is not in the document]
rafaeldtinoco: You're right. I think I'll maintain both, as different build targets, in the same Dockerfile, and default to the debian one. This way we can always test/use both cases if/when needed.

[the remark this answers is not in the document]
rafaeldtinoco: I've added STATIC=1 to tracee-rules as well. With that, if one wants to use alpine to build binaries, they'll be able to generate universal static binaries:

```
$ ldd ./dist/tracee*
./dist/tracee-ebpf:
	not a dynamic executable
./dist/tracee-rules:
	not a dynamic executable
```

I have also added flavors to builders. One may chose to use ubuntu-tracee-make or alpine-tracee-make, with both, STATIC=1 and STATIC=0 env variables.

rafaeldtinoco:
@danielpacak and @itaysk 

Okay, Im satisfied again, feel free to continue the review.

rafaeldtinoco:
@danielpacak and @itaysk

Okay, Im satisfied again, feel free to continue the review. I might push a few fixes here and there if I see its needed (as I'm fully testing this approach with the 'tracee-make' container flavors (alpine and ubuntu, static and dynamic)). But I think I covered both cases fine now.

Makefile.one
itaysk: maybe I'm missing something as well :) this target copies the bpf source to the bundle dir. If the bpf source has changed, but libbpf hasn't, it won't update the bpf code, and that's not good. did I get it right?

Dockerfile.tracee-make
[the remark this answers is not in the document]
itaysk: maybe break was overreacting :) generating files owned by another user might create issues if the host user want's to manipulate the file (even just delete it). not a big deal, I'm resolving this

itaysk (approved):

Makefile.docker
```diff
@@ -0,0 +1,82 @@
...
+alpine-prepare:
+	docker build \
+		-f $(ALPINE_MAKE_DOCKERFILE) \
+		-t $(ALPINE_MAKE_CONTNAME):latest \
+		--build-arg UID=$(UID) \
+		--build-arg GID=$(GID) \
+		.
+
+UBUNTU_MAKE_CONTNAME = ubuntu-tracee-make
+UBUNTU_MAKE_DOCKERFILE = docker/Dockerfile.alpine-tracee-make
```
yanivagman, line 24: Shouldn't this be Dockerfile.ubuntu-tracee-make?

Makefile.docker
```diff
@@ -0,0 +1,82 @@
...
+		-v /usr/src:/usr/src:ro \
+		-v /tmp/tracee:/tmp/tracee \
+		-e LIBBPFGO_OSRELEASE_FILE=/etc/os-release-host \
+		$(DOCKER_RUN_ENV)
+
+.PHONY: alpine-shell
+alpine-shell:
+	docker $(DOCKER_RUN_ARGS) \
+		-it $(ALPINE_MAKE_CONTNAME) \
+		/bin/bash
```
yanivagman, line 54: I think `/bin/bash` is not available in an alpine container (but `/bin/sh`)

Makefile.one
```diff
@@ -0,0 +1,468 @@
...
+.PHONY: help
+help:
+	@echo ""
+	@echo "$ make all                  # build tracee-ebpf, tracee-rules & rules"
+	@echo ""
+	@echo "$ make env                  # show makefile environment/variables"
+	@echo "$ make reqs                 # verify build requirements"
+	@echo ""
+	@echo "# build"
+	@echo ""
```
yanivagman, line 145: You can move:
`@echo "$ make all # build tracee-ebpf, tracee-rules & rules"`
Here, which I think makes more sense.

Makefile.one
```diff
@@ -0,0 +1,468 @@
+.PHONY: all
+all:
+	$(foreach var,$(.VARIABLES),$(info $(var) = $($(var))))
+
+	#$(MAKE) env
+	#$(MAKE) reqs
+	#$(MAKE) tracee-ebpf tracee-rules rules
```
yanivagman, line 7: So, all (the default target?) doesn't include the bpf-noncore target anymore, right?
Although this is good for environments with BTF enabled, it means that on environments without BTF the user will need to build the bpf object sepaeately.
This is ok, as long as we understand that, but don't we want to inform the user in such cases?

embedded-ebpf.go
```diff
@@ -0,0 +1,12 @@
+//go:build ebpf
+// +build ebpf
+
+package tracee
```
yanivagman, line 4: Why not having a pkg "embed" instead?

yanivagman (commented):
This is a VERY nice change, thanks @rafaeldtinoco!
Some small comments inside

itaysk:
there's a documentation update that should go along these changes, but since this is going to next, not main, feel free to add them later.

docker/Dockerfile.alpine-tracee-make
```diff
@@ -0,0 +1,31 @@
+FROM alpine:3.15 as alpine-tracee-make
```
itaysk, line 1: I'm afraid the directory name `docker` might be confusing to some, as it will signal that this is where the dockerfile/artifact scripts are. maybe rename it to `docker-builder` or something? (not sure, if you think it's clear than leave it)

Makefile.one
itaysk: I think it's a good call, given us de-emphasizing non-core. if someone wants to build the bpf target, they will probably do not want to build all of tracee in the same place.

Makefile.docker
```diff
@@ -0,0 +1,82 @@
+.PHONY: all
```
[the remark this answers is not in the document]
rafaeldtinoco: Yes, I have thought it as well. I'd like to use docker to build OS packages as well (like build files able to generate OS packages for distros). So I'll change it to "builder" only and have all building logic inside it in an organized way. Let me know if you have another idea please.

Makefile.docker
rafaeldtinoco: I made bash exist and be the default for the builder shell CMD only (with a profile to make it a better experience). The default for docker and other targets is and will still be sh only.

Makefile.one
```diff
@@ -0,0 +1,468 @@
+.PHONY: all
+all:
+	$(foreach var,$(.VARIABLES),$(info $(var) = $($(var))))
+
+	#$(MAKE) env
```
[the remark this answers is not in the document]
rafaeldtinoco: nope, I have to fix that, likely a leftover from development.

Makefile.one
rafaeldtinoco: I can document it indeed. If you see the docker makefile, Ive made sure to have it working for non-core by including the kernel headers bind mounts (and I tested it with non BTF environments).

 I think I should mention this in building documentation indeed.

docker/Dockerfile.alpine-tracee-make
rafaeldtinoco: Yep. mentioned this in the other comment. Let's call it "builder" and have everything related to building (but the default Makefile) there.

Makefile.one
rafaeldtinoco: You're not. You're actually right. The bundle was not being recreated every time the BPF target needed rebuilding. I've fixed that now.

embedded-ebpf.go
rafaeldtinoco:
```
tracee-ebpf/main.go:19:2: found packages embed (embedded-ebpf.go) and tracee (embedded.go) in /home/rafaeldtinoco/work/sources/ebpf/tracee
```

This is a change being coordinated with @danielpacak and its part of the tree reorganization.

danielpacak: `embedded*.go` files are supposed to be in the root directory of the project because Go embed package does not support relative paths such as `../dist/some-file.o`. On the other hand, we named our single Go module, which has its go.mod in the root dir, `github.com/aquasecurity/tracee`. That's why I thought that importing the main package and giving it `embed` or `embedded` alias will do the job.

builder/README.md
```diff
@@ -0,0 +1,73 @@
+## Instructions on how to use **Makefile.docker**
+
+In order to have a controlled building environment for tracee, tracee provides
+a [Makefile.docker](../Makefile.docker) file that allows you to create and use
+docker controlled environments to build & test `tracee-ebpf` and
+`tracee-rules`.
+
+### Creating a builder container
```
danielpacak, line 8: Can we add one or two sentences explaining why we even have two builder containers? I know it has to do with musl libc and glibc, but for curious individuals it can be helpful.

danielpacak (commented):
Great progress @rafaeldtinoco I added few more comments / questions before we can wrap it up!

builder/README.md
```diff
@@ -0,0 +1,73 @@
...
+$ make -f builder/Makefile.docker ubuntu-shell
+```
+
+### Executing compiled binaries inside builder shell
+
+Execute build commands inside picked builder shell:
+
+```
+tracee@402fbaa94e7f[/tracee]$ make -f Makefile.one clean
+tracee@402fbaa94e7f[/tracee]$ make -f Makefile.one bpf-core
```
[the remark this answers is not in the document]
rafaeldtinoco: I was just giving examples and wanted to show you could use *any* build target. But maybe that hasn't become clear enough, I'll add more target examples to this item.

.github/workflows/pr.yaml
```diff
@@ -52,6 +52,13 @@ jobs:
         run: |
           make -C tracee-ebpf/ all DOCKER=1
           make -C tracee-ebpf/ test DOCKER=1
+      - name: Test Single Makefile tracee-ebpf
```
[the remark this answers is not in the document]
rafaeldtinoco: Mind if I drop this change from this PR then ? And we do one PR to "adjust" all new Makefile builds ?

builder/Dockerfile.alpine-tracee-make
```diff
@@ -0,0 +1,31 @@
+FROM alpine:3.15 as alpine-tracee-make
```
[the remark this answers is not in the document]
rafaeldtinoco: Removing the alias.

About the file renaming, if we list all Dockerfiles starting with the prefix, all Vagrant files starting with the same prefix, ... we get better results when listing the builder directory contents (IMO). If others prefer the change you mentioned I'm opened to make it.

.github/workflows/pr.yaml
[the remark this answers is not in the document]
danielpacak: Sure . Let's take it out of scope of this PR

builder/Makefile.docker
```diff
@@ -0,0 +1,133 @@
+.PHONY: all
```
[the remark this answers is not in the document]
rafaeldtinoco: That would NOT be the only Makefile there. I have to create a Makefile.release, in  the next days, and other Makefiles.XXX might appear (Makefile.vagrant, Makefile.docs, etc).

builder/README-containers.md
```diff
@@ -0,0 +1,79 @@
+## Instructions on how to use the 'tracee-make' docker container
```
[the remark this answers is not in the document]
rafaeldtinoco: I wanted README.md inside builder because when you're navigating github, you click in the "builder" directory and get instructions on how to build. I want to update the official documentation, though, to include information from this README* files there. Maybe to do a new PR to make the builder/ directory README simple and document things more inside docs/. WDYT ? Should I even wipe the README.md from builder ?

builder/README.md
rafaeldtinoco: Definitely. Good idea.

builder/README-containers.md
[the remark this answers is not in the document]
danielpacak: Let's keep these README files for now. We can do a follow up to redact and publish as docs.

Makefile.one
```diff
@@ -0,0 +1,468 @@
...
+tracee-ebpf: $(OUTPUT_DIR)/tracee-ebpf
+
+$(OUTPUT_DIR)/tracee-ebpf: $(OUTPUT_DIR)/tracee.bpf.core.o $(TRACEE_EBPF_SRC)
+	$(GO_ENV_EBPF) $(CMD_GO) build \
+		-tags $(GO_TAGS_EBPF) \
+		-ldflags="-w \
+			-extldflags \"$(CGO_EXT_LDFLAGS_EBPF)\" \
+			-X main.version=\"$(VERSION)\" \
+			" \
+		-v -o $@ ./tracee-ebpf
```
itaysk, line 364: shouldn't we include `embedded-ebpf.go` from the root as well?

itaysk (commented):
we also have some potions of the makefile implemented in the main.go, so that tracee can compile the ebpf probe at runtime. how do we want to handle this? since this goes to next and not main, we can fix in a future PR

rafaeldtinoco:
> we also have some potions of the makefile implemented in the main.go, so that tracee can compile the ebpf probe at runtime. how do we want to handle this? since this goes to next and not main, we can fix in a future PR

Yep! I think we should make sure main.go (runtime building) is 1:1 with Makefile.one whenever it becomes the standard. Documented the need here: https://github.com/aquasecurity/tracee/issues/1233#issuecomment-992460985

Makefile.one
rafaeldtinoco: Yep, embedded-ebpf.go was created later so it wasn't included. Adding it.

builder/README.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+Execute `tracee-ebpf` and `tracee-rules` from inside builder shell:
```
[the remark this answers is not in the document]
rafaeldtinoco: Yes, I should highlight an example on the correct execution together with tracee-rules. Doing that.

yanivagman:
> we also have some potions of the makefile implemented in the main.go, so that tracee can compile the ebpf probe at runtime. how do we want to handle this? since this goes to next and not main, we can fix in a future PR

I think we can start thinking about taking it out of tracee-ebpf binary after btfhub objects will be embedded into it. One option we can take in tracee-ebpf main, if it detects it should compile the bpf object file, is to tell the user that CO-RE is not supported on his environment (nor bpf/btf were provided by env-vars), so he should run the builder to make the bpf object and install it before tracee can proceed. WDYT?

builder/Dockerfile.alpine-tracee-make
```diff
@@ -0,0 +1,31 @@
...
+    apk --no-cache add musl-dev libc6-compat && \
+    apk --no-cache add llvm clang go make gcc && \
+    apk --no-cache add linux-headers && \
+    apk --no-cache add elfutils-dev && \
+    apk --no-cache add libelf-static && \
+    apk --no-cache add zlib-static && \
+    curl -L -o /usr/bin/opa https://github.com/open-policy-agent/opa/releases/download/v0.33.1/opa_linux_amd64_static && \
+    chmod 755 /usr/bin/opa
+
+RUN export uid=$uid gid=$gid && \
```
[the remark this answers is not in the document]
rafaeldtinoco: Yes, I don't know if the UID/GID picked by the builder user will exist in the distro picked to be our builder. By adding those lines to the end of passwd & group file I'm sure they will be the ones picked (instead of having to deal with the logic of checking if those UID/GID already exist).

yanivagman:
One more thing - I think we should add to dockerhub the builders images as part of a release. This way, a user will not have to build the image, and we can instruct him about how to build an ebpf object if necessary (also see my previous comment). WDYT?

Makefile.one
```diff
@@ -0,0 +1,468 @@
+.PHONY: all
```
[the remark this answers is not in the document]
rafaeldtinoco: I thought about having a Makefile.docs inside `builder` in next PRs. Idea is that this week I'll finish this Makefile by guaranteeing it has all targets covered by the old Makefile.

Makefile.one
rafaeldtinoco: https://github.com/aquasecurity/tracee/issues/1233#issuecomment-992487235

rafaeldtinoco:
> > we also have some potions of the makefile implemented in the main.go, so that tracee can compile the ebpf probe at runtime. how do we want to handle this? since this goes to next and not main, we can fix in a future PR
> 
> I think we can start thinking about taking it out of tracee-ebpf binary after btfhub objects will be embedded into it. One option we can take in tracee-ebpf main, if it detects it should compile the bpf object file, is to tell the user that CO-RE is not supported on his environment (nor bpf/btf were provided by env-vars), so he should run the builder to make the bpf object and install it before tracee can proceed. WDYT?

I like this approach of telling user to execute specific command. This removes the burden of us maintaining the building logic. But we would have to think of a way to extract embedded bundle to running env, so the user is able to compile it (just like we automatically do). Right ?

rafaeldtinoco:
> One more thing - I think we should add to dockerhub the builders images as part of a release. This way, a user will not have to build the image, and we can instruct him about how to build an ebpf object if necessary (also see my previous comment). WDYT?

So the builder image would contain the tracee-ebpf/ bundle directory ?

danielpacak (approved):

yanivagman:
> I like this approach of telling user to execute specific command. This removes the burden of us maintaining the building logic. But we would have to think of a way to extract embedded bundle to running env, so the user is able to compile it (just like we automatically do). Right ?

Actually, if we choose to remove compilation in tracee-ebpf we can take the embedded bundle out of it as well, and only keep the co-re object. The builder can provide all the necessary sources to build the bpf object file.

yanivagman:
> > we also have some potions of the makefile implemented in the main.go, so that tracee can compile the ebpf probe at runtime. how do we want to handle this? since this goes to next and not main, we can fix in a future PR
> 
> I think we can start thinking about taking it out of tracee-ebpf binary after btfhub objects will be embedded into it. One option we can take in tracee-ebpf main, if it detects it should compile the bpf object file, is to tell the user that CO-RE is not supported on his environment (nor bpf/btf were provided by env-vars), so he should run the builder to make the bpf object and install it before tracee can proceed. WDYT?

Opened #1226 to track this

Makefile.one
itaysk: wdym when rebuilding the entire project? if you build just this target, does it work? Unless I'm missing something I thing it'll generate a binary without the embeds

rafaeldtinoco: This target depends on `tracee.bpf.core.o`

`tracee.bpf.core.o` will execute a

`$(MAKE) $(OUTPUT_DIR)/tracee.bpf`

EACH time `libbpf.a` needs rebuild (or tracee ebpf obj src/headers). 

If you change embedded-ebpf.go, then "go build" will be executed again with either:

1. existing contents of the previous "tracee-ebpf" build (which is ok since they haven't been touched)
2. run all dependencies again, since `dist/tracee.ebpf` is always re-created (now) each time you have to build the CO-RE object (or each time you change libbpf, which also needs to rebuild CO-RE object).

itaysk: Maybe I didn't explain the question right: The command in the makefile is essentially go build ./tracee-ebpf . Then how does the go compiler finds the embedding file in the root /

itaysk: I think this answers my question (from `go help build`)

> Build compiles the packages named by the import paths,
along with their dependencies, but it does not install the results.
If the arguments to build are a list of .go files from a single directory,
build treats them as a list of source files specifying a single package.

Issue #1220, opened 2021-12-07
Move Rego signatures from ./tracee-rules/signatures/rego to ./signatures


danielpacak:
In scope of #1217 we should move Rego signatures from ./tracee-rules/signatures/rego to ./signatures  
At the same time we can fix #1199

simar7:
Can we close this now?

Issue #1199, opened 2021-12-03
aio.rego is copied to the dist/rules/ directory


danielpacak:
The `./tracee-rules/signatures/rego/regosig/aio.rego` is supposed to be embedded in Go binary and used internally by `aio.go`. We should not copy it to the `dist/rules/aio.rego`

simar7:
Fixed with https://github.com/aquasecurity/tracee/commit/11c1b21e8164c8ca523ac4e17d80b442609d4666

Issue #1229, opened 2021-12-08
Move Go signatures from ./tracee-rules/signatures/golang to ./signatures


danielpacak:

simar7:
Can we close this now?

PR #1231, opened 2021-12-08
chore: Move signatures into their own top level directory

simar7:
Fixes: https://github.com/aquasecurity/tracee/issues/1220
Fixes: https://github.com/aquasecurity/tracee/issues/1199


tracee-rules/signature.go
```diff
@@ -12,11 +14,10 @@ import (
 	"plugin"
 	"strings"
 
-	"github.com/aquasecurity/tracee/tracee-rules/signatures/rego/regosig"
 	"github.com/aquasecurity/tracee/tracee-rules/types"
 )
 
-//go:embed signatures/rego/helpers.rego
+//go:embed helpers.rego
```
simar7, line 20: This duplication is needed as `go embed` cannot work with relative paths (and symlinks).

itaysk:
also fixes: #1229 ? (if so please update so the issue is closed)

signatures/rego/regosig/traceerego_test.go
```diff
@@ -6,10 +6,12 @@ import (
 	"os"
 	"testing"
 
+	"github.com/aquasecurity/tracee/signatures/rego/regosig"
```
danielpacak, line 9: Are you using goimports to organise imports? I"m wondering why it's so different in terms of order and new lines added here and there?

tracee-rules/Makefile
```diff
@@ -8,11 +8,11 @@ OUT_DIR ?= dist
 OUT_BIN := $(OUT_DIR)/tracee-rules
 OUT_RULES := $(OUT_DIR)/rules
 GO_SRC := $(shell find . -type f -name '*.go')
-GOSIGNATURES_DIR ?= signatures/golang
+GOSIGNATURES_DIR ?= ../signatures/golang
 GOSIGNATURES_SRC := $(shell find $(GOSIGNATURES_DIR) -type f -name '*.go' ! -name '*_test.go' ! -path '$(GOSIGNATURES_DIR)/examples/*')
 OUT_GOSIGNATURES := $(OUT_RULES)/builtin.so
-REGO_SIGNATURES_DIR ?= signatures/rego
-REGO_SIGNATURES_SRC := $(shell find $(REGO_SIGNATURES_DIR) -type f -name '*.rego' ! -name '*_test.rego' ! -path '$(REGO_SIGNATURES_DIR)/examples/*')
+REGO_SIGNATURES_DIR ?= ../signatures/rego
```
danielpacak, line 14: I think we must also adjust the working directory for the `opa test` command. It seems that we no longer run any tests for Rego.

Instead of

```
opa test . --verbose --ignore="examples" --ignore="dist" --ignore="benchmark"
```

use something like?

```
opa test ./../signatures --verbose --ignore="examples" --ignore="dist" --ignore="benchmark"
```

danielpacak (changes requested):
Overall LGTM! There are tests failing though. In particular, `signature_test.go` where we refer to the relative path `signatures/rego`.

Quick fix might be to replace it with `./../signatures/rego` but preferably a unit test in the tracee-rules module should be agnostic to where we placed signatures after all.

I can reproduce the errors in my environment with the following commands so make sure they pass for you as well:

```
cd tracee-rules
make test
make test DOCKER=1
```

signatures/rego/regosig/aio_test.go
```diff
@@ -6,9 +6,11 @@ import (
 	"os"
 	"testing"
 
+	"github.com/aquasecurity/tracee/signatures/rego/regosig"
```
danielpacak, line 9: I'm not sure we should move aio.go and traceerego.go. These for me belong to tracee-rules core functionality and just rely on Rego file that we are trying to move to it's own location under ./signatures/rego

/cc @itaysk WDYT?

itaysk: agree!

itaysk: also aio.rego, and probably mapper.go? what is Mapper?

tracee-rules/signature.go
danielpacak: To get rid of duplicates in `signatures/rego/helpers.rego` and `tracee-rules/helpers.rego` you can created the `embedded.go` file in the root of the Git repo / Go module with the following content:

```
package tracee

import _ "embed"

//go:embed signatures/rego/helpers.rego
var RegoHelpersCode string
```

Then you. can use import and use the `RegoHelpersCode` variable in `signature.go`. I'm using this approach in https://github.com/aquasecurity/starboard/blob/main/embedded.go because of the limitations you mentioned.

signatures/rego/regosig/aio_test.go
danielpacak: - aio.rego is integral part of aio.go (actually Go-embedded)
- mapper.go is utility code shared between aio.go and traceerego.go (mapping of Rego eval results to tracee-rules structs)

danielpacak (approved):
LGTM!

PR #1214, opened 2021-12-06
add type of stdin in sched_process_exec

roikol:

PR #1235, opened 2021-12-10
Change args type metadata when parsed

grantseltzer:


itaysk (commented):
with this approach, the consumer needs to use both the name and type to understand what type is the value (e.g flags+string=parsed flags). This means that the consumer has responsibility not only to check the name (easy), but to know which arg names might be parsed by tracee (hard). I think it would be better to put all the necessary information in the type, so instead of string we say something like flags-string or something similar. wdyt?

tracee-ebpf/tracee/argprinters.go
```diff
@@ -124,6 +141,15 @@ func (t *Tracee) parseArgs(ctx *context, args map[string]interface{}) error {
 	return nil
 }
 
+func setArgumentType(name, typeName string, argMetas *[]external.ArgMeta) {
```
mtcherni95, line 144: Is there any use case where we pass to the function something that is not `"string"`?

tracee-ebpf/tracee/argprinters.go
```diff
@@ -23,7 +24,7 @@ func Print16BytesSliceIP(in []byte) string {
 	return ip.String()
 }
 
-func (t *Tracee) parseArgs(ctx *context, args map[string]interface{}) error {
+func (t *Tracee) parseArgs(ctx *context, args map[string]interface{}, argMetas []external.ArgMeta) error {
```
yanivagman, line 27: Shouldn't you pass argMetas as a pointer?

tracee-ebpf/tracee/argprinters.go
yanivagman: parseArgs() is currently used for pretty-printing, so argument type is always a string

yanivagman (commented):
> with this approach, the consumer needs to use both the name and type to understand what type is the value (e.g flags+string=parsed flags). This means that the consumer has responsibility not only to check the name (easy), but to know which arg names might be parsed by tracee (hard). I think it would be better to put all the necessary information in the type, so instead of string we say something like flags-string or something similar. wdyt?

Why do the consumer need to know the name of the argument?
1. With this change, the consumer has the correct type - whether its a string or an int, so he can easily parse it without knowing the name. "parsed flags" is not a type!
2. The consumer controls whether to pass "parse-arguments" or not, so he knows if he needs to expect a raw argument or a parsed argument

itaysk:
> Why do the consumer need to know the name of the argument?

currently you can't distinguish between an arg of type string, and an non-string arg that was "printed" to be a string. You need this distinction because the latter needs additional handling by the consumer. Look at the code that uses this, it has something like `if arg.Name == "whatever"` right?

> With this change, the consumer has the correct type - whether its a string or an int, so he can easily parse it without knowing the name. "parsed flags" is not a type!

The purpose of `arg.Type` is as a hint to the consumer how to process the `arg.Value` which might have lost its original type information in transit (such as in the JSON case). It doesn't have to be the C primitive type. In fact we might already treat it this way e.g https://github.com/aquasecurity/tracee/pull/1170

> The consumer controls whether to pass "parse-arguments" or not, so he knows if he needs to expect a raw argument or a parsed argument

Not necessarily. The "consumer" is the signature author which is written differently than the runtime configuration of tracee-ebpf.

itaysk:
> The purpose of arg.Type is as a hint to the consumer how to process the arg.Value which might have lost its original type information in transit (such as in the JSON case). It doesn't have to be the C primitive type. In fact we might already treat it this way e.g #1170

When I wrote this I thought we should have, when implementing raw arguments, just added the printed argument instead of augmenting it. This would give a more elegant solution to this problem and also to backwards compatibility.

yanivagman:
> > Why do the consumer need to know the name of the argument?
> 
> currently you can't distinguish between an arg of type string, and an non-string arg that was "printed" to be a string. You need this distinction because the latter needs additional handling by the consumer. Look at the code that uses this, it has something like `if arg.Name == "whatever"` right?
> 

Both an arg of type string and a non-string arg that was "printed" to be string should be treated as a string. The type information is only used to parse the argument correctly - in this case, as a string. The consumer of tracee-ebpf should know the sematics of the argument when he uses `arg.Name`. See my next comment.

> > With this change, the consumer has the correct type - whether its a string or an int, so he can easily parse it without knowing the name. "parsed flags" is not a type!
> 
> The purpose of `arg.Type` is as a hint to the consumer how to process the `arg.Value` which might have lost its original type information in transit (such as in the JSON case). It doesn't have to be the C primitive type. In fact we might already treat it this way e.g #1170
> 

#1170 didn't add any new types. Eventually #1214 was merged instead which only used umode_t.

> > The consumer controls whether to pass "parse-arguments" or not, so he knows if he needs to expect a raw argument or a parsed argument
> 
> Not necessarily. The "consumer" is the signature author which is written differently than the runtime configuration of tracee-ebpf.

The current signatures we have "assume" arguments are already parsed. We know that this what happens, because we pass parse-arguments to the entrypoint (https://github.com/aquasecurity/tracee/blob/main/entrypoint.sh#L35).
In the future, we may want to move to "raw" arguments (e.g. for performace reasons). In that case, we will need to change the signatures to use the new helpers @grantseltzer added in libbpfgo.
Anyways, the consumer is aware of the whether he uses "raw" arguments or not. Even the signature author.

itaysk:
I don't think it's reasonable to expect the consumer of the output to know how tracee-ebpf was configured like you suggest (raw-args/parse-args). but also I don't think it's critical for this PR so fine with me

grantseltzer:
I'm not sure we're all in agreement on how we expect this metadata change to be used. I could see this being useful for the sake of a tracee-ebpf consumer that is taking JSON output. The consumer could check the type and read the actual data accordingly. However the consumer is using JSON, couldn't they already check if the type is a number or a string? And isn't gob encoded with the data type as well? Who would actually be reading this "type" field, is it useful for the sake of determining how to read the raw data? Probably not.

So giving context about what the data actually represents (i.e. a string representation of the 'flag' argument, `flag (string)`) probably makes more sense than just "string" which is how I have it now. WDYT? @yanivagman @itaysk

itaysk:
> However the consumer is using JSON, couldn't they already check if the type is a number or a string?

No, because there are no primitive types in JSON, except `string`, `bool`, `float64`. This is why this "type" hint exists.

>  isn't gob encoded with the data type as well?

Yes. gob data is received exactly how it's sent. that's why we have gaps between data consumed via JSON/gob

> Who would actually be reading this "type" field

1. The `Event` JSON Unmarshaler is the primary consumer
2. Any user that reads the raw JSON output (and therefore doesn't use the Unmarshaler) (for example, you streamed Tracee-eBPF output to splunk.
3. Potentially also Gob users would need it, if we mangle with the output again like suggested in other PRs then it could be useful to know, but now we don't have such use case (and I'm fine with deciding we will never have).

> is it useful for the sake of determining how to read the raw data? Probably not.

I hope now it's clear that it is

itaysk:
To clarify, I'm ok with merging this after Yaniv's review

yanivagman (approved):

Issue #1193, opened 2021-12-01
Kernel types sizing mistakes when using sizeof with CO-RE


yanivagman:
We should be careful when using the sizeof operator on structs as our new slim version of vmlinux.h does not always contain all the fields of the struct.
As an example, sockaddr_in didn't contain the padding field (fixed in https://github.com/aquasecurity/tracee/pull/1190).
Another example is any place where we use READ_KERN, which uses sizeof to read the correct number of bytes of the requested field.

After going over our bpf code, I found two more structs that have incomplete definition but which are used with sizeof:
`struct tcphdr` and `struct udphdr`.

Other structs seem to have the complete definition.

To deal with such a problem, there are two helpers we can use (see "Sizing kernel types and fields" in https://nakryiko.com/posts/bpf-core-reference-guide/): `bpf_core_type_size()` and `bpf_core_field_size()`.
We might need to update our READ_KERN core macro to use `bpf_core_type_size()` to avoid future bugs.

Also, quoting from the bpf-core-reference-guide: "The general recommendation is to not read struct fields as a whole, if at all possible. Prefer reading just the primitive fields you are ultimately interested in."

In summary:
1. Add full definition of `struct tcphdr` and `struct udphdr`.
2. Update READ_KERN co-re macro to use  `bpf_core_type_size()` instead of `sizeof`
3. If possible, read the primitive fields instead of the full struct. Currently, we read the following structs in our bpf code:
    - path
    - iovec
    - sockaddr_un
    - sockaddr_in
    - sockaddr_in6
    - qstr
    - timespec64
    - in6_addr

yanivagman:
To deal with such problem, there are two helpers we can use (see "Sizing kernel types and fields" in https://nakryiko.com/posts/bpf-core-reference-guide/): `bpf_core_type_size()` and `bpf_core_field_size()`.

Also, quoting from the bpf-core-reference-guide: "The general recommendation is to not read struct fields as a whole, if at all possible. Prefer reading just the primitive fields you are ultimately interested in."

rafaeldtinoco:
I believe I have already addressed this with the new network code. The protocol headers are fully declared due to headers copying from skb into them (allocated in stack).

PR #1238, opened 2021-12-13
fix socket_dup event wrong port under core

### LINKED ISSUE #1237: socket_dup event wrong port under CORE
hey,

under CORE, socket_dup event shows wrong port number. for example:

ipv4:
15:15:04:334699  1000   python3          43338   43338   0                security_socket_connect sockfd: 3, remote_addr: map[sa_family:AF_INET sin_addr:8.8.8.8 sin_port:53]
15:15:04:378489  1000   python3          43338   43338   0                socket_dup           oldfd: 3, newfd: 0, remote_addr: map[sa_family:AF_INET sin_addr:8.8.8.8 sin_port:512]

port should be 53 - like in security_socket_connect.

ipv6:
15:15:09:925689  1000   python3          43340   43340   0                security_socket_connect sockfd: 3, remote_addr: map[sa_family:AF_INET6 sin6_addr:::1 sin6_flowinfo:1 sin6_port:8080 sin6_scopeid:16777216]
15:15:09:926380  1000   python3          43340   43340   0                socket_dup           oldfd: 3, newfd: 0, remote_addr: map[sa_family:AF_INET6 sin6_addr:::1 sin6_flowinfo:0 sin6_port:2560 sin6_scopeid:0]

port should be 8080 - like in security_socket_connect.

roikol:
fixes #1237 

use struct sock to get network details instead of using struct inet_sock. 
inet_sock's fields are macros, so it messes up the READ_KERN.

### DIFF
--- a/tracee-ebpf/tracee/co-re/vmlinux.h
+++ b/tracee-ebpf/tracee/co-re/vmlinux.h
@@ -418,23 +418,14 @@ struct in6_addr {
 	} in6_u;
 };
 
-typedef __u32 __portpair;
-typedef __u64 __addrpair;
-
 struct sock_common {
 	union {
-	    __addrpair	skc_addrpair;
 		struct {
 			__be32 skc_daddr;
 			__be32 skc_rcv_saddr;
 		};
 	};
 	union {
-        unsigned int	skc_hash;
-        __u16		skc_u16hashes[2];
-    };
-	union {
-	    __portpair	skc_portpair;
 		struct {
 			__be16 skc_dport;
 			__u16 skc_num;
--- a/tracee-ebpf/tracee/co-re/vmlinux.h
+++ b/tracee-ebpf/tracee/co-re/vmlinux.h
@@ -418,14 +418,23 @@ struct in6_addr {
 	} in6_u;
 };
 
+typedef __u32 __portpair;
+typedef __u64 __addrpair;
+
 struct sock_common {
 	union {
+	    __addrpair	skc_addrpair;
 		struct {
 			__be32 skc_daddr;
 			__be32 skc_rcv_saddr;
 		};
 	};
 	union {
+        unsigned int	skc_hash;
+        __u16		skc_u16hashes[2];
+    };
+	union {
+	    __portpair	skc_portpair;
 		struct {
 			__be16 skc_dport;
 			__u16 skc_num;
--- a/tracee-ebpf/tracee/tracee.bpf.c
+++ b/tracee-ebpf/tracee/tracee.bpf.c
@@ -983,7 +983,6 @@ static __always_inline struct sock_common * get_sock_sk_common(struct sock *sock
         bpf_probe_read((void *)sk_common, sizeof(struct sock_common), &sock->__sk_common);
     #endif
 
-    bpf_probe_read((void *)sk_common, sizeof(struct sock_common), &sock->__sk_common);
     return sk_common;
 }
 
--- a/tracee-ebpf/tracee/tracee.bpf.c
+++ b/tracee-ebpf/tracee/tracee.bpf.c
@@ -526,36 +526,37 @@ struct mount {
 
 /*================================= MAPS =====================================*/
 
-BPF_HASH(config_map, u32, u32);                         // various configurations
-BPF_HASH(kconfig_map, u32, u32);                        // kernel config variables
-BPF_HASH(chosen_events_map, u32, u32);                  // events chosen by the user
-BPF_HASH(traced_pids_map, u32, u32);                    // track traced pids
-BPF_HASH(new_pids_map, u32, u32);                       // track processes of newly executed binaries
-BPF_HASH(containers_map, u32, u8);                      // map cgroup id to container status {EXISTED, CREATED, STARTED}
-BPF_HASH(args_map, u64, args_t);                        // persist args between function entry and return
-BPF_HASH(syscall_data_map, u32, syscall_data_t);        // persist data during syscall execution
-BPF_HASH(inequality_filter, u32, u64);                  // filter events by some uint field either by < or >
-BPF_HASH(uid_filter, u32, u32);                         // filter events by UID, for specific UIDs either by == or !=
-BPF_HASH(pid_filter, u32, u32);                         // filter events by PID
-BPF_HASH(mnt_ns_filter, u64, u32);                      // filter events by mount namespace id
-BPF_HASH(pid_ns_filter, u64, u32);                      // filter events by pid namespace id
-BPF_HASH(uts_ns_filter, string_filter_t, u32);          // filter events by uts namespace name
-BPF_HASH(comm_filter, string_filter_t, u32);            // filter events by command name
-BPF_HASH(bin_args_map, u64, bin_args_t);                // persist args for send_bin funtion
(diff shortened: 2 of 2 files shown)


tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -958,39 +959,64 @@ static inline struct mount *real_mount(struct vfsmount *mnt)
...
+    if (sk_common == NULL)
+        return NULL;
+
+    #ifdef CORE
+        bpf_core_read((void *)sk_common, sizeof(struct sock_common), &sock->__sk_common);
+    #else
+        bpf_probe_read((void *)sk_common, sizeof(struct sock_common), &sock->__sk_common);
+    #endif
+
+    bpf_probe_read((void *)sk_common, sizeof(struct sock_common), &sock->__sk_common);
```
yanivagman, line 986: Why do we need this? (same as line 983 for non-core code)

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -958,39 +959,64 @@ static inline struct mount *real_mount(struct vfsmount *mnt)
...
 
-static __always_inline u32 get_inet_saddr(struct inet_sock *inet)
+static __always_inline struct sock* get_socket_sock(struct socket *socket)
 {
-    return READ_KERN(inet->inet_saddr);
+    return READ_KERN(socket->sk);
 }
 
-static __always_inline u32 get_inet_daddr(struct inet_sock *inet)
+static __always_inline struct sock_common * get_sock_sk_common(struct sock *sock)
```
yanivagman, line 972: Better to avoid reading the whole sock_common struct (#1193) + I don't see a reason to use a map here.

tracee-ebpf/tracee/co-re/vmlinux.h
```diff
@@ -418,14 +418,23 @@ struct in6_addr {
...
 			__be32 skc_daddr;
 			__be32 skc_rcv_saddr;
 		};
 	};
 	union {
+        unsigned int	skc_hash;
+        __u16		skc_u16hashes[2];
+    };
+	union {
+	    __portpair	skc_portpair;
```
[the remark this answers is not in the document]
roikol: you're right, i can remove these fields. at first i thought the problem was that the struct wasn't full.

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -526,36 +526,37 @@ struct mount {
...
+BPF_ARRAY(file_filter, path_filter_t, 3);                       // filter vfs_write events
+BPF_ARRAY(string_store, path_filter_t, 1);                      // store strings from userspace
+BPF_PERCPU_ARRAY(bufs, buf_t, MAX_BUFFERS);                     // percpu global buffer variables
+BPF_PERCPU_ARRAY(bufs_off, u32, MAX_BUFFERS);                   // holds offsets to bufs respectively
+BPF_PROG_ARRAY(prog_array, MAX_TAIL_CALL);                      // store programs for tail calls
+BPF_PROG_ARRAY(prog_array_tp, MAX_TAIL_CALL);                   // store programs for tail calls
+BPF_PROG_ARRAY(sys_enter_tails, MAX_EVENT_ID);                  // store programs for tail calls
+BPF_PROG_ARRAY(sys_exit_tails, MAX_EVENT_ID);                   // store programs for tail calls
+BPF_STACK_TRACE(stack_addresses, MAX_STACK_ADDRESSES);          // store stack traces
+BPF_PERCPU_ARRAY(sock_common_struct, struct sock_common, 1);    // struct sock_common holder
```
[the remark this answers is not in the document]
roikol: using a local variable reaches the BPF stack limit of 512 bytes on program tracepoint__inet_sock_set_state.
too many variables are saved on the stack. 
the way to avoid the limit is to use the ARRAY map, with the size of the struct, and read from there.

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -958,39 +959,64 @@ static inline struct mount *real_mount(struct vfsmount *mnt)
...
+    u32 idx = 0;
+    struct sock_common *sk_common = bpf_map_lookup_elem(&sock_common_struct, &idx);
+    if (sk_common == NULL)
+        return NULL;
+
+    #ifdef CORE
+        bpf_core_read((void *)sk_common, sizeof(struct sock_common), &sock->__sk_common);
+    #else
+        bpf_probe_read((void *)sk_common, sizeof(struct sock_common), &sock->__sk_common);
+    #endif
```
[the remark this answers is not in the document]
roikol: we can't because we have a pointer to struct sock_common - this is what we get from the map. this is related to my comment above

tracee-ebpf/tracee/tracee.bpf.c
roikol: the reason for the bug is using:
`READ_KERN(sk->__sk_common.skc_dport)`

it's just doesn't get the right field. (i've noticed it gets the skc_family field intead, which is 32 bits below)


using:
```
struct sock_common sk_common = READ_KERN(sk->__sk_common);
return sk_common.skc_dport;
```

solves it.

roikol: but then we reach the BPF stack limit, so struct sock_common needs to be saved in a map.

tracee-ebpf/tracee/tracee.bpf.c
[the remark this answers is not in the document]
yanivagman: If you will read the primitive fields and not the full struct, this should not be a problem

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -958,39 +959,63 @@ static inline struct mount *real_mount(struct vfsmount *mnt)
...
 {
-    return READ_KERN(inet->inet_daddr);
+    // get sock_common from map to avoid reaching bpf stack limit
+    u32 idx = 0;
+    struct sock_common *sk_common = bpf_map_lookup_elem(&sock_common_struct, &idx);
+    if (sk_common == NULL)
+        return NULL;
+
+    #ifdef CORE
+        bpf_core_read((void *)sk_common, sizeof(struct sock_common), &sock->__sk_common);
```
yanivagman, line 981: Note that now you have the WRONG size of struct sock_common (because you removed the extra fields)

tracee-ebpf/tracee/tracee.bpf.c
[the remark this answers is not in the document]
yanivagman: Ok, this will not be relevant if you will read the primitve values and get rid of the map

tracee-ebpf/tracee/tracee.bpf.c
yanivagman: > the reason for the bug is using: `READ_KERN(sk->__sk_common.skc_dport)`
> 
> it's just doesn't get the right field. (i've noticed it gets the skc_family field intead, which is 32 bits below)
> 
> using:
> 
> ```
> struct sock_common sk_common = READ_KERN(sk->__sk_common);
> return sk_common.skc_dport;
> ```
> 
> solves it.

You can read a pointer to sock_common instead of the whole struct, then use it to read the specific field

yanivagman:
Already fixed by #1281

PR #1244, opened 2021-12-14
tracee-ebpf: fix ppid

yanivagman:
This PR fixes two bugs:
1. table-verbose printer printed an incorrect value of the host ppid
2. host ppid should be `real_parent->tgid` (pid in userspace) and not `real_parent->pid` (tid in userspace)


rafaeldtinoco (approved):
Could you add the PR description to git log (since there is no issue opened for it). The change LGTM.

PR #1253, opened 2021-12-14
creating and using package capabilties

mtcherni95:
Solves #1254

### DIFF
--- a/cmd/tracee-ebpf/main.go
+++ b/cmd/tracee-ebpf/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"github.com/aquasecurity/tracee/pkg/capabilities"
 	"io"
 	"io/ioutil"
 	"log"
@@ -88,11 +89,11 @@ func main() {
 
 			// environment capabilities
 
-			selfCap, err := getSelfCapabilities()
+			selfCap, err := capabilities.SelfCapabilities()
 			if err != nil {
 				return err
 			}
-			if err = checkRequiredCapabilities(selfCap); err != nil {
+			if err = capabilities.CheckRequiredCapabilities(selfCap); err != nil {
 				return err
 			}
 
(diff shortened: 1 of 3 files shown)


cmd/tracee-ebpf/main.go
```diff
@@ -88,11 +89,11 @@ func main() {
 
 			// environment capabilities
 
-			selfCap, err := getSelfCapabilities()
+			selfCap, err := capabilities.SelfCapabilities()
```
danielpacak, line 92: I'm wondering if we could avoid repetition in package names and exported types / functions as explained in [Effective Go Package Names](https://go.dev/doc/effective_go#package-names) WDYT?

```suggestion
			selfCap, err := capabilities.Self()
```

cmd/tracee-ebpf/main.go
```diff
@@ -88,11 +89,11 @@ func main() {
...
 			if err != nil {
 				return err
 			}
-			if err = checkRequiredCapabilities(selfCap); err != nil {
+			if err = capabilities.CheckRequiredCapabilities(selfCap); err != nil {
```
danielpacak, line 96: DITTO
```suggestion
			if err = capabilities.CheckRequired(selfCap); err != nil {
```

pkg/capabilities/capabilities.go
```diff
@@ -0,0 +1,28 @@
+package capabilities
+
+import (
+	"fmt"
+	"github.com/syndtr/gocapability/capability"
```
danielpacak, line 5: Out of scope of this PR but I was wondering if we are okey with gocapability BSD 2-Clause "Simplified" License. (I assume that the fact it's not very actively maintained is because is feature complete?)

simar7:
We might be stepping on over on each others feet here https://github.com/aquasecurity/tracee/pull/1256

I've restructured the package layout and re-did tests for it. I think we should reconcile our PRs to avoid duplicate work.

PR #1255, opened 2021-12-14
creating and using buckets cache

mtcherni95:
Related to #960 (comment)


tracee-ebpf/tracee/tracee.go
```diff
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/aquasecurity/tracee/pkg/bucketscache"
```
yanivagman, line 9: nit: please put this import below the "standard" imports

mtcherni95: sure, thanks

yanivagman (approved):

Issue #1254, opened 2021-12-14
create capabilities package


mtcherni95:
frm #960 (comment)

PR #1256, opened 2021-12-15
feat: Introduce capabilities package

simar7:
This PR addresses the need for capabilities to be in its own testable package space.

Close https://github.com/aquasecurity/tracee/issues/1254


mtcherni95:
I think we both opened a PR on the same feature :) https://github.com/aquasecurity/tracee/pull/1253

simar7:
> I think we both opened a PR on the same feature :) https://github.com/aquasecurity/tracee/pull/1253

I'm fine keeping either one as long as we can reconcile our work together. If you would like to merge yours, please add the tests I've tried to add here.

mtcherni95:
> > I think we both opened a PR on the same feature :) #1253
> 
> I'm fine keeping either one as long as we can reconcile our work together. If you would like to merge yours, please add the tests I've tried to add here.

Since you already added tests, I'm more than ok to go with yours :)

cmd/tracee-ebpf/main.go
```diff
@@ -87,12 +88,15 @@ func main() {
 			cfg.Output = &output
 
 			// environment capabilities
-
-			selfCap, err := getSelfCapabilities()
+			selfCap, err := capability.NewPid2(0)
```
danielpacak, line 91: Why do we split capability.NewPid2(0) out of method that check capabilities of the current process? It seems that we always do capability.NewPid2 and then Load()

pkg/capabilities/capabilities.go
```diff
@@ -0,0 +1,25 @@
+package capabilities
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/syndtr/gocapability/capability"
+)
+
+func CheckRequiredCapabilities(caps capability.Capabilities, reqCaps []capability.Cap) error {
```
danielpacak, line 10: As commented on @michaelkatch PR, the capabilities packages has enough meaning to say that we're checking required capabilities.

```suggestion
func CheckRequired(caps capability.Capabilities, reqCaps []capability.Cap) error {
```

pkg/capabilities/capabilities.go
```diff
@@ -0,0 +1,25 @@
...
+	for _, c := range reqCaps {
+		if !caps.Get(capability.EFFECTIVE, c) {
+			return fmt.Errorf("insufficient privileges to run: missing CAP_%s", strings.ToUpper(c.String()))
+		}
+	}
+	return nil
+}
+
+func LoadSelfCapabilities(selfCap capability.Capabilities) (capability.Capabilities, error) {
```
danielpacak, line 19: I like more the scope of the "original" `getSelfCapabilities` method that inits the struct and loads capabilities.
```suggestion
func Self() (capability.Capabilities, error) {
```

pkg/capabilities/capabilities_test.go
```diff
@@ -0,0 +1,61 @@
...
+			assert.Equal(t, capability.EFFECTIVE, capType)
+			assert.Equal(t, capability.CAP_SYS_ADMIN, c)
+			return false
+		}}, []capability.Cap{capability.CAP_SYS_ADMIN})
+		assert.Equal(t, "insufficient privileges to run: missing CAP_SYS_ADMIN", err.Error())
+	})
+}
+
+func TestLoadSelfCapabilities(t *testing.T) {
+	t.Run("happy path", func(t *testing.T) {
```
danielpacak, line 48: Do you think we can really unit test this method? It seems that we're actually testing the implementation of the fakeCapability. If you comment out the line `err := selfCap.Load()` in the `LoadSelfCapabilities` method the test called "happy path" is still green.

BTW, what happy path means in this context?

danielpacak (changes requested):

cmd/tracee-ebpf/main.go
mtcherni95: I would keep the original implementation and usage of `getSelfCapabilities` and export it.

mtcherni95 (changes requested):

cmd/tracee-ebpf/main.go
simar7: I agree with you both. I did this to be able to dependency inject Load(). `NewPid2()` doesn't seem to be an a method defined through an interface so it's difficult to be able to dependency inject, unless we monkey patch it. 

Any other ways (besides the one suggested above) where we could keep `getSelfCapabilities` as it's today and also test the behaviour of both NewPid2() and Load() deterministically in a test? Thoughts?

pkg/capabilities/capabilities_test.go
simar7: Yeah that's a good point - I think another assertion that's missing is to assert if Load() was actually called or not. What do you think about that? Personally, I'm not fully satisfied with it either. Any ideas?

Happy path in this context would mean if the following items take place:
1. Load() is called and returns without errors.
2. Load() is called the required number of times that it's supposed to be (in this case once).

danielpacak:
> Have you guys seen #1202?
> This PR drops capabilities on start, and is related to this change.

Thank you for the heads up @yanivagman. I wasn't aware of the other patch. That's one of the reasons why I'm more and more inclined to [Opportunistic Refactoring](https://martinfowler.com/bliki/OpportunisticRefactoring.html) rather than refactoring for refactoring kind of task.

cmd/tracee-ebpf/main.go
simar7: Please see the alternative approach as described above demonstrated here https://github.com/aquasecurity/tracee/pull/1256/commits/60d0a30163a9e5258d4bb427a49a65c61b1462c3

pkg/capabilities/capabilities_test.go
simar7: Please see the idea mentioned above demonstrated here https://github.com/aquasecurity/tracee/pull/1256/commits/60d0a30163a9e5258d4bb427a49a65c61b1462c3

rafaeldtinoco:
@AlonZivony I think this overlaps to what I suggested to you in your PR #1202 (on selfCapabilities functions).

danielpacak (approved):

cmd/tracee-ebpf/main.go
```diff
@@ -87,12 +88,11 @@ func main() {
 			cfg.Output = &output
 
 			// environment capabilities
-
-			selfCap, err := getSelfCapabilities()
+			selfCap, err := capabilities.Self()
 			if err != nil {
 				return err
 			}
-			if err = checkRequiredCapabilities(selfCap); err != nil {
```
danielpacak, line 95: I think we can also remove `checkRequiredCapabilities` and `getSelfCapabilities` methods from main.go.

mtcherni95 (approved):
LGTM

PR #1273, opened 2021-12-16
Do not build eBPF non CO-RE object during tracee-ebpf execution

rafaeldtinoco:
With this commit we have 2 different Makefiles:

1. builder/Makefile.tracee (responsible for building **tracee** containers in 3 flavors: **core**, **core-btfhub** and **nocore**). We should have 3 official images:

- tracee-core
- tracee-core-btfhub
- tracee-nocore

In all of them you can opt to execute "tracee" (tracee-ebpf + tracee-rules) by default OR to export a TRACEE_EBPF_ONLY=1 variable and have them running as "tracee-ebpf" only.

```
$ make -f builder/Makefile.tracee help

To generate tracee container(s):

    $ make -f builder/Makefile.tracee build-alpine-tracee-core
    $ make -f builder/Makefile.tracee build-alpine-tracee-core-btfhub
    $ make -f builder/Makefile.tracee build-alpine-tracee-nocore

To execute tracee container(s):

    $ make -f builder/Makefile.tracee run-alpine-tracee-core
    $ make -f builder/Makefile.tracee run-alpine-tracee-nocore

To execute tracee-ebpf container(s):

    $ make -f builder/Makefile.tracee run-alpine-tracee-ebpf-core
    $ make -f builder/Makefile.tracee run-alpine-tracee-ebpf-nocore

Note:

You may provide arguments to executions using ARG variable

Example:

    $ make -f builder/Makefile.tracee build-alpine-tracee-core-btfhub

    $ make -f builder/Makefile.tracee \
          run-alpine-tracee-ebpf-core \
          ARG="--debug -trace event!='sched*'"
```

2. builder/Makefile.tracee-make (responsible for creating a building & execution controlled environment using 2 distributions: alpine & ubuntu).

```
$ make -f builder/Makefile.tracee-make help

To create an alpine-tracee-make container:

    $ make -f builder/Makefile.tracee-make alpine-prepare

To create an ubuntu-tracee-make container:

    $ make -f builder/Makefile.tracee-make ubuntu-prepare

To execute an alpine-tracee-make shell:

    $ make -f builder/Makefile.tracee-make alpine-shell

To execute an ubuntu-tracee-make shell:

    $ make -f builder/Makefile.tracee-make ubuntu-shell

Examples:

    $ make -f builder/Makefile.tracee-make ubuntu-make ARG="clean"
    $ make -f builder/Makefile.tracee-make ubuntu-make ARG="bpf-core"
[...]

PR #1276, opened 2021-12-16
Adjusting Makefile.one and btfhub.sh with fixes and improvements.

rafaeldtinoco:
bbf7da2 Makefile.one: build dependencies fixes
a8b8507 btfhub.sh: improvement and fixes
bcf796e tracee-make: fix dockerfiles sudoers and dependencies
6ffbad3 Makefile.one: default parallel jobs for libbpf build only

### DIFF
--- a/Makefile.one
+++ b/Makefile.one
@@ -17,6 +17,10 @@ MAKEFLAGS += --no-print-directory
 # tools
 #
 
+CMD_TR ?= tr
+CMD_CUT ?= cut
+CMD_AWK ?= awk
+CMD_SED ?= sed
 CMD_GIT ?= git
 CMD_CLANG ?= clang
 CMD_LLC ?= llc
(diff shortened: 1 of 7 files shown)


rafaeldtinoco:
@yanivagman this PR contain fixes to Makefile.one logic recently merged. Feel free to merge this, if you agree, or to steal this PR and do any adjustments it might need, as I'll be off for couple of weeks.

itaysk (commented):
IMO it's a bit too much that every recipe needs to declare all the commands it will use

rafaeldtinoco:
> IMO it's a bit too much that every recipe needs to declare all the commands it will use

Okay. I can remove checks for basic commands (such as CMD_RM, CMD_MKDIR, CMD_TOUCH, CMD_GREP, CMD_CAT). I'll keep the others, though, as they aren't "installed by default" such as those basic commands.

rafaeldtinoco:
@itaysk I have removed the basic checks now.

rafaeldtinoco:
@itaysk mind giving me a +1 on this one (if you think its good) ? Its blocking PR: #1273 for me (because of rebasing need).

itaysk:
Could someone else please take a look? Don't let me block you

rafaeldtinoco:
@yanivagman could you, please, review this one ? It's a bunch of needed fixes to Makefile.one and I would like to move on with it so I can generate the container images with the new Makefile and, finally, get rid of current building files. 

Thanks.

Makefile.one
```diff
@@ -31,24 +32,73 @@ CMD_CAT ?= cat
...
-	@command -v $* >/dev/null || (echo "missing required tool $*" ; exit 1)
+.check_%:
+#
+	@command -v $* >/dev/null
+	if [ $$? -ne 0 ]; then
+		echo "missing required tool $*"
+		exit 1
+	else
+		touch $@ # avoid target rebuilds due to inexistent file
+	fi
```
yanivagman, line 43: I'm not sure I understand why this is required.
If non-docker make is chosen, does this mean that we create these files on the system of the user?

Makefile.one
```diff
@@ -31,24 +32,73 @@ CMD_CAT ?= cat
...
+	touch $@ # avoid target rebuilds over and over due to inexistent file
+
+GO_VERSION = $(shell $(CMD_GO) version 2>/dev/null | awk '{print $$3}' | sed 's:go::g' | cut -d. -f1,2)
+
+.checkver_$(CMD_GO): \
+	| .check_$(CMD_GO)
+#
+	@if [ "${GO_VERSION}" != "1.16" ] && [ "${GO_VERSION}" != "1.17" ]; then
+		echo -n "you MUST use golang 1.16 or 1.17, "
```
yanivagman, line 86: This will break when a newer version of go will be out.
Maybe we should check if version is higher or equal to 1.16 instead?

itaysk:
@grantseltzer @danielpacak PTAL

Makefile.one
rafaeldtinoco: Those files are created during Makefile execution in current PWD. Without this files, each time the build target is triggered it will be executed. The files mean that the check was done and passed (or else they wouldn't exist). This is a well known workaround for Makefile if you need a logic to be triggered as a check, then you have to bind the logic to the existence of a file that is unmodified between runs. Or else each time you call "make -f Makefile.one tracee-ebpf" the build command will be executed (even if there are no modified files).

Makefile.one
rafaeldtinoco: I thought about it and decided to bind it to 1.16 or 1.17 because we know, for sure, they are compatible. We don't know if our code is compatible to 1.18 (or 1.19, ...) and the idea was to hard code the supported versions here. With that said, do you still want to make that change ?

yanivagman: So you think that go is not backwards compatible? I'm not sure this is true, but I'm also not 100% sure about that

yanivagman (approved):
LGTM
Please only check about go backwards compatibility

Makefile.one
rafaeldtinoco: @mtcherni95 any opinion here? Should we do specific versioning tests ? Or assume anything higher than 1.16 would be compatible ?

rafaeldtinoco:
> LGTM Please only check about go backwards compatibility

Thanks. I asked @mtcherni95 opinion before merging this.

Makefile.one
itaysk: I agree with Yaniv, that we need to check for minimum version (if we use a new language feature not available previously), but not for maximum version (if existing code will stop working), since Go is supposed to be forwards compatible, and in rare cases where it's not it would be a matter of bug fix

PR #1281, opened 2021-12-19
reorder variables on stack

roikol:
related to #1237
@yanivagman and i did some investigating on the bug.
we understood that the order of the variables on the stack solves the issue.
there's probably a bug somewhere with CORE+clang that creates some overflow or stack override.
changing the order of the variables avoids the override.

### DIFF
--- a/tracee-ebpf/tracee/tracee.bpf.c
+++ b/tracee-ebpf/tracee/tracee.bpf.c
@@ -2278,8 +2278,8 @@ static __always_inline int send_socket_dup(event_data_t *data, u64 oldfd, u64 ne
         save_to_submit_buf(data, &remote, sizeof(struct sockaddr_in), 2);
     }
     else if (family == AF_INET6) {
-        struct sockaddr_in6 remote;
         net_conn_v6_t net_details = {};
+        struct sockaddr_in6 remote;
 
         get_network_details_from_sock_v6(sk, &net_details, 0);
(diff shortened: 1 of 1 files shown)


rafaeldtinoco:
This is usually related to small amount of registers of BPF arch and the register pressure during compilation. That usually happens because of bound checks that, sometimes, use shadow registers to keep track of boundaries (afaict). When reorganizing the stack you mitigate the pressure and, sometimes, it helps indeed. CPU V3 is better at this afaik.

yanivagman:
> This might be related to small amount of registers of BPF arch and the register pressure during compilation. That usually happens because of bound checks that, sometimes, use shadow registers to keep track of boundaries (afaict). When reorganizing the stack you mitigate the pressure and, sometimes, it helps indeed. CPU V3 is better at this afaik. Even if that is the case, it should be handled, or failed, at compilation time (instead of generation a faulty binary).

Yet this shouldn't cause a bug in functionality (wrong output), but possible verifier issues. What we see here shouldn't happen. The fact that we change the order of the variables on the stack shouldn't cause the output to be different - there must be a bug somewhere related to CO-RE (in clang?), and it might be related to the registers allocation like you suggest. By the way, changing the optimization (0-3) didn't help with this issue

yanivagman:
I'm going to merge this, but let's keep #1237 open until we figure out why this fixes it.
We should understand the root cause to avoid such bugs in the future

yanivagman (approved):

PR #1285, opened 2021-12-20
tracee-ebpf: move filter consts to filters.go

yanivagman:


tracee-ebpf/tracee/filters.go
```diff
@@ -13,6 +13,36 @@ import (
...
+const (
+	filterIn  uint8 = 1
+	filterOut uint8 = 2
+)
+
+const (
+	uidLess uint32 = iota
+	uidGreater
+	pidLess
+	pidGreater
```
danielpacak, line 30: I see that uidGreater, pidGreater, mntNsGreater, and pidNsGreater constants are defined but not used in code. Is it expected?

tracee-ebpf/tracee/consts.go
```diff
@@ -44,34 +32,14 @@ const (
...
-// Set default inequality values
-// val<0 and val>math.MaxUint64 should never be used by the user as they give an empty set
+// Custom KernelConfigOption's to extend kernel_config helper support
+// Add here all kconfig variables used within tracee.bpf.c
 const (
-	LessNotSetUint    uint64 = 0
-	GreaterNotSetUint uint64 = math.MaxUint64
-	LessNotSetInt     int64  = math.MinInt64
-	GreaterNotSetInt  int64  = math.MaxInt64
+	CONFIG_ARCH_HAS_SYSCALL_WRAPPER helpers.KernelConfigOption = iota + helpers.CUSTOM_OPTION_START
```
danielpacak, line 42: NIT: Lets use camel case instead of snake case to make Go vet happy

danielpacak (approved):
LGTM. I just left 1 suggestion and 1 question.

tracee-ebpf/tracee/filters.go
yanivagman: It is used implicitly in line 170 (previously 140) to initialize the bpf map. It is always less+1 of theatching filter

Issue #1283, opened 2021-12-20
Save captured files using container id


yanivagman:
We currently save captured (written) files into directories which name is the mount namespace id.
This is problematic as mount namespace id might be reused when a container is exited, and a new container starts.
As we now extract the container id (using the cgroup id) per event, we can use it to name the directories used to save these captured artifacts.

itaysk:
is container ID guaranteed to be unique over time?

itaysk:
https://github.com/opencontainers/runtime-spec/blob/main/runtime.md#state

> id (string, REQUIRED) is the container's ID. This MUST be unique across all containers on this host. There is no requirement that it be unique across hosts.

I think it's supposed to be unique between the currently active containers, but not sure about over previous containers (which I think won't solve your issue?)

yanivagman:
> nk it's supposed to be unique between the currently active containers, but not sure about over previous containers (which I think won't solve your issue?)

In practice, every container has a unique id.

rafaeldtinoco:
> > is container ID guaranteed to be unique over time?
> 
> With a high probability, the answer is yes. In contrast, with mount namespace id, on some systems, with a probability of 1 you will get a collision

Isn't it a SHA-1 ? It would be 1/(2^80) or something like it (if we're using the full id).

itaysk:
It's a sha in docker maybe but other runtimes might choose otherwise.  Also people might create cgroups with any name. Just adding a consideration feel free to ignore, but maybe adding a random number or something will be safer

mtcherni95:
Do we want to change the path by using cgroup ns ID only for capture write or also for other artifacts?

PR #1291, opened 2021-12-21
using cgroup ns id instead of mnt ns id for path to captured written file

mtcherni95:
Solves #1283

### DIFF
--- a/tracee-ebpf/tracee/events_processor.go
+++ b/tracee-ebpf/tracee/events_processor.go
@@ -100,7 +100,11 @@ func (t *Tracee) processEvent(ctx *context, args map[string]interface{}, argMeta
 			}
 
 			// stop processing if write was already indexed
-			fileName := fmt.Sprintf("%d/write.dev-%d.inode-%d", ctx.MntID, dev, inode)
+			containerId := t.containers.GetCgroupInfo(ctx.CgroupID).ContainerId
+			if containerId == "" {
+				containerId = "host"
+			}
+			fileName := fmt.Sprintf("%s/write.dev-%d.inode-%d", containerId, dev, inode)
 			indexName, ok := t.writtenFiles[fileName]
 			if ok && indexName == filePath {
 				return nil
@@ -142,9 +146,13 @@ func (t *Tracee) processEvent(ctx *context, args map[string]interface{}, argMeta
 				}
 				castedSourceFileCtime := int64(sourceFileCtime)
 
-				capturedFileID := fmt.Sprintf("%d:%s", ctx.MntID, sourceFilePath)
+				containerId := t.containers.GetCgroupInfo(ctx.CgroupID).ContainerId
+				if containerId == "" {
+					containerId = "host"
+				}
+				capturedFileID := fmt.Sprintf("%s:%s", containerId, sourceFilePath)
 				if t.config.Capture.Exec {
-					destinationDirPath := filepath.Join(t.config.Capture.OutputPath, strconv.Itoa(int(ctx.MntID)))
+					destinationDirPath := filepath.Join(t.config.Capture.OutputPath, containerId)
 					if err := os.MkdirAll(destinationDirPath, 0755); err != nil {
 						return err
 					}
--- a/tracee-ebpf/tracee/tracee.bpf.c
+++ b/tracee-ebpf/tracee/tracee.bpf.c
@@ -3495,8 +3495,8 @@ static __always_inline u32 send_bin_helper(void* ctx, struct bpf_map_def *prog_a
     }
 
 #define F_SEND_TYPE   0
-#define F_MNT_NS      (F_SEND_TYPE + sizeof(u8))
-#define F_META_OFF    (F_MNT_NS + sizeof(u32))
+#define F_CGROUP_ID   (F_SEND_TYPE + sizeof(u8))
+#define F_META_OFF    (F_CGROUP_ID + sizeof(u64))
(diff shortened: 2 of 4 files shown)


tracee-ebpf/tracee/events_processor.go
```diff
@@ -100,7 +100,7 @@ func (t *Tracee) processEvent(ctx *context, args map[string]interface{}, argMeta
 			}
 
 			// stop processing if write was already indexed
-			fileName := fmt.Sprintf("%d/write.dev-%d.inode-%d", ctx.MntID, dev, inode)
+			fileName := fmt.Sprintf("%d/write.dev-%d.inode-%d", ctx.CgroupID, dev, inode)
```
yanivagman, line 103: Let's use container id here as well

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -3471,17 +3471,17 @@ static __always_inline u32 send_bin_helper(void* ctx, struct bpf_map_def *prog_a
     }
 
 #define F_SEND_TYPE   0
-#define F_MNT_NS      (F_SEND_TYPE + sizeof(u8))
-#define F_META_OFF    (F_MNT_NS + sizeof(u32))
+#define F_CGROUP_NS   (F_SEND_TYPE + sizeof(u8))
```
yanivagman, line 3474: We are not interested in cgroup namespace but cgroup id

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -3471,17 +3471,17 @@ static __always_inline u32 send_bin_helper(void* ctx, struct bpf_map_def *prog_a
...
 #define F_SZ_OFF      (F_META_OFF + SEND_META_SIZE)
 #define F_POS_OFF     (F_SZ_OFF + sizeof(unsigned int))
 #define F_CHUNK_OFF   (F_POS_OFF + sizeof(off_t))
 #define F_CHUNK_SIZE  (MAX_PERCPU_BUFSIZE >> 1)
 
     bpf_probe_read((void **)&(file_buf_p->buf[F_SEND_TYPE]), sizeof(u8), &bin_args->type);
 
-    u32 mnt_id = get_task_mnt_ns_id((struct task_struct *)bpf_get_current_task());
-    bpf_probe_read((void **)&(file_buf_p->buf[F_MNT_NS]), sizeof(u32), &mnt_id);
+    u32 cgroup_id = get_task_cgroup_ns_id((struct task_struct *)bpf_get_current_task());
```
yanivagman, line 3483: Wrong helper to get cgroup id.
cgroup namespace id != cgroup id

yanivagman (commented):
Other places that should be changed:
https://github.com/aquasecurity/tracee/blob/main/tracee-ebpf/tracee/events_processor.go#L103
https://github.com/aquasecurity/tracee/blob/main/tracee-ebpf/tracee/events_processor.go#L145
https://github.com/aquasecurity/tracee/blob/main/tracee-ebpf/tracee/events_processor.go#L147

tracee-ebpf/tracee/events_processor.go
mtcherni95: oops, right

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -3471,17 +3471,17 @@ static __always_inline u32 send_bin_helper(void* ctx, struct bpf_map_def *prog_a
...
 #define F_SZ_OFF      (F_META_OFF + SEND_META_SIZE)
 #define F_POS_OFF     (F_SZ_OFF + sizeof(unsigned int))
 #define F_CHUNK_OFF   (F_POS_OFF + sizeof(off_t))
 #define F_CHUNK_SIZE  (MAX_PERCPU_BUFSIZE >> 1)
 
     bpf_probe_read((void **)&(file_buf_p->buf[F_SEND_TYPE]), sizeof(u8), &bin_args->type);
 
-    u32 mnt_id = get_task_mnt_ns_id((struct task_struct *)bpf_get_current_task());
-    bpf_probe_read((void **)&(file_buf_p->buf[F_MNT_NS]), sizeof(u32), &mnt_id);
+    u64 cgroup_id = ((context_t *)ctx)->cgroup_id;
```
yanivagman, line 3483: This is NOT the context_t we define in tracee, but the bpf context

mtcherni95:
this PR should be promoted, only after this https://github.com/aquasecurity/tracee/pull/1301

tracee-ebpf/tracee/events_processor.go
```diff
@@ -142,9 +146,13 @@ func (t *Tracee) processEvent(ctx *context, args map[string]interface{}, argMeta
 				}
 				castedSourceFileCtime := int64(sourceFileCtime)
 
-				capturedFileID := fmt.Sprintf("%d:%s", ctx.MntID, sourceFilePath)
+				containerId := t.containers.GetCgroupInfo(ctx.CgroupID).ContainerId
```
yanivagman, line 149: You already have containderid initialized above in this function. Why doing it again?

mtcherni95: We do not need it for every case + every time we call the function only one case is matched. I'm okay with taking it apart at the beginning of the switch statement but I don't think is necessary.

mtcherni95:
cherry-picking to https://github.com/aquasecurity/tracee/pull/1301 as concurrency of containers cgroup is need for this current PR only.

PR #1292, opened 2021-12-21
tracee-ebpf: fix arm64 compilation due to inexisting events

rafaeldtinoco:
commit ce77ccc8470e ("Feature/support 32 bit syscalls (#1196)") added
events that did not exist for arm64 architecture. This commit adds those
events to the list of undefined, to arm64, syscalls list.

### DIFF
--- a/tracee-ebpf/tracee/events_arm64.go
+++ b/tracee-ebpf/tracee/events_arm64.go
@@ -350,6 +350,9 @@ const (
 	ProcessMreleaseEventID          int32 = 448
 )
 
+// TODO: There isnt a Unique32BitSyscalls list for arm64
+const Unique32BitSyscallsEndID int32 = Unique32BitSyscallsStartID
+
 // following syscalls are undefined on arm64
 const (
 	OpenEventID int32 = iota + 10000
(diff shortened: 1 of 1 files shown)


tracee-ebpf/tracee/events_arm64.go
```diff
@@ -352,7 +352,7 @@ const (
 
 // following syscalls are undefined on arm64
 const (
-	OpenEventID int32 = iota + 10000
+	OpenEventID int32 = iota + Unique32BitSyscallsStartID
```
yanivagman, line 355: These are not Unique32BitSyscallsStartID, but undefined values.
We should either keep 10000 or set a new macro for it

rafaeldtinoco: My problem is with this piece of "printList()":

```
for i := tracee.Unique32BitSyscallsStartID; i < int(tracee.Unique32BitSyscallsEndID); i++ {
        index := int32(i)
        event, ok := tracee.EventsIDToEvent[index]
        if !ok {
                continue
        }
        if event.Sets != nil {
                eventSets := fmt.Sprintf("%-22s %-40s %s\n",
                event.Name, fmt.Sprintf("%v", event.Sets),
                fetchFormattedEventParams(index))
                b.WriteString(eventSets)
        } else {
                b.WriteString(event.Name + "\n")
        }
}
```

that only works for x86 currently, right ?

rafaeldtinoco: I kept 10000 for now and created a Unique32BitSyscallsEndID with the same value of  Unique32BitSyscallsStartID for arm64. That fixes the issue for now, we should address this when "fixing" arm64 support.

yanivagman (approved):

PR #1296, opened 2021-12-22
update fileless_execution

chriskaliX:
add /run/shm to Fileless Execution, and here is my reference (https://linuxsecurity.com/features/fileless-malware-on-linux)


itaysk:
Thank you! can you please update the test in https://github.com/aquasecurity/tracee/blob/main/signatures/rego/fileless_execution_test.rego ?

chriskaliX:
> Thank you! can you please update the test in https://github.com/aquasecurity/tracee/blob/main/signatures/rego/fileless_execution_test.rego ?

My bad. it's done :-)

itaysk (approved):
thanks for contributing!

PR #1297, opened 2021-12-22
update cgroup lookup table if cgroup doesn't exists, when getting cgroup info

mtcherni95:


roikol:
looks good to me. allows us to avoid calling lookup&update everywhere we are trying to get containerID from cgroupID.

yanivagman (approved):

Issue #1290, opened 2021-12-21
process context for net events and process tree


itamarmaouda101:
hi 
currently im working to added feature of saving the context of every process in the system in map for give to the user more information about the incoming net packets and parse net work events. 
also it will help #1262 by giving the first process tree (of the currently running processes when tracee loads) to tracee rules

PR #1300, opened 2021-12-23
Process context map 

itamarmaouda101:
hi 
i created map that holds all of the processes in the computer. (see #1290)
the map key is the host_tid and the value is struct of the process context which holds the process data (the data is the same as the data in the event context data)- ill added the struct at the next comment

the entire map routines are based by 3 parts:
1. static upload - when tracee loads itseld , the initProcessTreeMap function captures every task in the Os (by iterating the /proc/<pid>/task/<task_id> )- and then captures the entire data to the struct and appends it to the map 
2. dynamic upload from user space- when tracee receive event of fork/exec/exit - it append/remove the event context data (of the process only) to/from the map,
3. dynamic upload from kernel space- when tracee receive event of fork/exec/exit - it append/remove the event context data (of the process only) to/from other map which his BPF_HASH map that will be available to the processTreeMap  in case of  tracee will receive an event of process before the creation of the process came to the user space parser

How to use: 
i created function of getProcessCtx() which gets a host_tid and checks in the two maps and return the entire struct

This PR will give tracee the ability to extract important data about the processes running in the system.
Because of that tracee eventually will have the ability to create net-events like the other events (and not just print in debug mode) by getting the the missed parts of the event from the process-tree map and then send it to the event channel

### DIFF
--- a/cmd/tracee-ebpf/printer.go
+++ b/cmd/tracee-ebpf/printer.go
@@ -233,6 +233,7 @@ func (p *gobEventPrinter) Init() error {
 	p.outEnc = gob.NewEncoder(p.out)
 	gob.Register(external.Event{})
 	gob.Register(external.SlimCred{})
+	gob.Register(external.ProcessCtx{})
 	gob.Register(make(map[string]string))
 	return nil
 }
--- a/cmd/tracee-ebpf/printer.go
+++ b/cmd/tracee-ebpf/printer.go
@@ -233,7 +233,6 @@ func (p *gobEventPrinter) Init() error {
 	p.outEnc = gob.NewEncoder(p.out)
 	gob.Register(external.Event{})
 	gob.Register(external.SlimCred{})
-	gob.Register(external.ProcessCtx{})
 	gob.Register(make(map[string]string))
 	return nil
 }
--- a/cmd/tracee-rules/input.go
+++ b/cmd/tracee-rules/input.go
@@ -47,7 +47,6 @@ func setupTraceeGobInputSource(opts *traceeInputOptions) (chan types.Event, erro
 	dec := gob.NewDecoder(opts.inputFile)
 	gob.Register(external.Event{})
 	gob.Register(external.SlimCred{})
-	gob.Register(external.ProcessCtx{})
 	gob.Register(make(map[string]string))
 	res := make(chan types.Event)
 	go func() {
--- a/cmd/tracee-rules/input.go
+++ b/cmd/tracee-rules/input.go
@@ -47,6 +47,7 @@ func setupTraceeGobInputSource(opts *traceeInputOptions) (chan types.Event, erro
 	dec := gob.NewDecoder(opts.inputFile)
 	gob.Register(external.Event{})
 	gob.Register(external.SlimCred{})
+	gob.Register(external.ProcessCtx{})
 	gob.Register(make(map[string]string))
 	res := make(chan types.Event)
 	go func() {
--- a/pkg/containers/containers.go
+++ b/pkg/containers/containers.go
@@ -90,6 +90,25 @@ func (c *Containers) procMountsCgroups() error {
 	return nil
 }
 
+// gets a containerID from a given task or process directory path
+func GetContainerIdFromTaskDir(taskPath string) (string, error) {
+	containerId := ""
+	taskPath = fmt.Sprintf("%s/cgroup", taskPath)
+	cgroupFile, err := os.Open(taskPath)
+	if err != nil {
+		return containerId, err
+	}
+	defer cgroupFile.Close()
+	scanner := bufio.NewScanner(cgroupFile)
+	for scanner.Scan() {
+		containerId, _ := getContainerIdFromCgroup(scanner.Text())
+		if containerId != "" {
+			break
+		}
+	}
+	return containerId, nil
+}
+
 // populate walks the cgroup fs informing CgroupUpdate() about existing dirs.
 func (c *Containers) populate() error {
 	if c.cgroupMP == "" {
--- a/pkg/external/external.go
+++ b/pkg/external/external.go
@@ -239,35 +239,32 @@ func (alert MemProtAlert) String() string {
 	}
 }
 
-type Process_ctx struct {
-	Ctime     uint64 // Ctime timestamp
-	Cgroup_id uint64
-	Pid       uint32 // PID as in the userspace term = nstgid
-	Tid       uint32 // TID as in the userspace term =
-	Ppid      uint32 // Parent PID as in the userspace term
-	Host_pid  uint32 // PID in host pid namespace
-	Host_tid  uint32 // TID in host pid namespace
-	Host_ppid uint32 // Parent PID in host pid namespace
-	Uid       uint32
-	Mnt_id    uint32
-	Pid_id    uint32
+type ProcessCtx struct {
+	Ctime       int
+	ContainerID string
+	Pid         int
+	Tid         int
+	Ppid        int
+	HostPid     int
+	HostTid     int
+	HostPpid    int
+	Uid         int
+	MntId       int
+	PidId       int
 }
 
-func SetProcessContext(ctx []byte) Process_ctx {
-	//if len(ctx) <52{
-	//	return
-	//}
-	var procCtx = Process_ctx{}
-	procCtx.Ctime = binary.LittleEndian.Uint64(ctx[0:8])
-	procCtx.Cgroup_id = binary.LittleEndian.Uint64(ctx[8:16])
-	procCtx.Pid = binary.LittleEndian.Uint32(ctx[16:20])
-	procCtx.Tid = binary.LittleEndian.Uint32(ctx[20:24])
-	procCtx.Ppid = binary.LittleEndian.Uint32(ctx[24:28])
-	procCtx.Host_pid = binary.LittleEndian.Uint32(ctx[28:32])
-	procCtx.Host_tid = binary.LittleEndian.Uint32(ctx[32:36])
-	procCtx.Host_ppid = binary.LittleEndian.Uint32(ctx[36:40])
-	procCtx.Uid = binary.LittleEndian.Uint32(ctx[40:44])
-	procCtx.Mnt_id = binary.LittleEndian.Uint32(ctx[44:48])
-	procCtx.Pid_id = binary.LittleEndian.Uint32(ctx[48:52])
+func SetProcessContext(ctx []byte) ProcessCtx {
+	var procCtx = ProcessCtx{}
+	procCtx.Ctime = int(binary.LittleEndian.Uint64(ctx[0:8]))
+	procCtx.ContainerID = string(binary.LittleEndian.Uint64(ctx[8:16]))
+	procCtx.Pid = int(binary.LittleEndian.Uint32(ctx[16:20]))
+	procCtx.Tid = int(binary.LittleEndian.Uint32(ctx[20:24]))
+	procCtx.Ppid = int(binary.LittleEndian.Uint32(ctx[24:28]))
+	procCtx.HostPid = int(binary.LittleEndian.Uint32(ctx[28:32]))
+	procCtx.HostTid = int(binary.LittleEndian.Uint32(ctx[32:36]))
+	procCtx.HostPpid = int(binary.LittleEndian.Uint32(ctx[36:40]))
+	procCtx.Uid = int(binary.LittleEndian.Uint32(ctx[40:44]))
+	procCtx.MntId = int(binary.LittleEndian.Uint32(ctx[44:48]))
+	procCtx.PidId = int(binary.LittleEndian.Uint32(ctx[48:52]))
 	return procCtx
 }
--- a/pkg/external/external.go
+++ b/pkg/external/external.go
@@ -2,6 +2,7 @@ package external
 
 import (
 	"bytes"
+	"encoding/binary"
 	"encoding/json"
 	"fmt"
 	"strconv"
@@ -236,4 +237,37 @@ func (alert MemProtAlert) String() string {
 	default:
 		return "Unknown alert"
 	}
+}
+
+type Process_ctx struct {
+	Ctime     uint64 // Ctime timestamp
+	Cgroup_id uint64
+	Pid       uint32 // PID as in the userspace term = nstgid
+	Tid       uint32 // TID as in the userspace term =
+	Ppid      uint32 // Parent PID as in the userspace term
+	Host_pid  uint32 // PID in host pid namespace
+	Host_tid  uint32 // TID in host pid namespace
+	Host_ppid uint32 // Parent PID in host pid namespace
+	Uid       uint32
+	Mnt_id    uint32
+	Pid_id    uint32
+}
+
+func SetProcessContext(ctx []byte) Process_ctx {
+	//if len(ctx) <52{
+	//	return
+	//}
+	var procCtx = Process_ctx{}
+	procCtx.Ctime = binary.LittleEndian.Uint64(ctx[0:8])
+	procCtx.Cgroup_id = binary.LittleEndian.Uint64(ctx[8:16])
+	procCtx.Pid = binary.LittleEndian.Uint32(ctx[16:20])
+	procCtx.Tid = binary.LittleEndian.Uint32(ctx[20:24])
+	procCtx.Ppid = binary.LittleEndian.Uint32(ctx[24:28])
+	procCtx.Host_pid = binary.LittleEndian.Uint32(ctx[28:32])
+	procCtx.Host_tid = binary.LittleEndian.Uint32(ctx[32:36])
+	procCtx.Host_ppid = binary.LittleEndian.Uint32(ctx[36:40])
+	procCtx.Uid = binary.LittleEndian.Uint32(ctx[40:44])
+	procCtx.Mnt_id = binary.LittleEndian.Uint32(ctx[44:48])
+	procCtx.Pid_id = binary.LittleEndian.Uint32(ctx[48:52])
+	return procCtx
 }
--- a/pkg/external/external.go
+++ b/pkg/external/external.go
@@ -2,7 +2,6 @@ package external
 
 import (
 	"bytes"
-	"encoding/binary"
 	"encoding/json"
 	"fmt"
 	"strconv"
@@ -237,34 +236,4 @@ func (alert MemProtAlert) String() string {
 	default:
 		return "Unknown alert"
 	}
-}
-
-type ProcessCtx struct {
-	Ctime       int
-	ContainerID string
-	Pid         int
-	Tid         int
-	Ppid        int
-	HostPid     int
-	HostTid     int
-	HostPpid    int
-	Uid         int
-	MntId       int
-	PidId       int
-}
-
-func SetProcessContext(ctx []byte) ProcessCtx {
-	var procCtx = ProcessCtx{}
-	procCtx.Ctime = int(binary.LittleEndian.Uint64(ctx[0:8]))
-	procCtx.ContainerID = string(binary.LittleEndian.Uint64(ctx[8:16]))
-	procCtx.Pid = int(binary.LittleEndian.Uint32(ctx[16:20]))
-	procCtx.Tid = int(binary.LittleEndian.Uint32(ctx[20:24]))
-	procCtx.Ppid = int(binary.LittleEndian.Uint32(ctx[24:28]))
-	procCtx.HostPid = int(binary.LittleEndian.Uint32(ctx[28:32]))
-	procCtx.HostTid = int(binary.LittleEndian.Uint32(ctx[32:36]))
-	procCtx.HostPpid = int(binary.LittleEndian.Uint32(ctx[36:40]))
-	procCtx.Uid = int(binary.LittleEndian.Uint32(ctx[40:44]))
-	procCtx.MntId = int(binary.LittleEndian.Uint32(ctx[44:48]))
-	procCtx.PidId = int(binary.LittleEndian.Uint32(ctx[48:52]))
-	return procCtx
 }
--- a/tracee-ebpf/tracee/co-re/vmlinux.h
+++ b/tracee-ebpf/tracee/co-re/vmlinux.h
@@ -714,8 +714,10 @@ struct tcphdr {
 };
 
 struct udphdr {
-	__be16 source;
-	__be16 dest;
+	__be16	source;
+    __be16	dest;
+    __be16	len;
+    __sum16	check;
 };
 
 enum kernel_read_file_id {
--- a/tracee-ebpf/tracee/co-re/vmlinux.h
+++ b/tracee-ebpf/tracee/co-re/vmlinux.h
@@ -716,8 +716,6 @@ struct tcphdr {
 struct udphdr {
 	__be16	source;
     __be16	dest;
-    __be16	len;
-    __sum16	check;
 };
 
 enum kernel_read_file_id {
--- a/tracee-ebpf/tracee/co-re/vmlinux.h
+++ b/tracee-ebpf/tracee/co-re/vmlinux.h
@@ -714,8 +714,8 @@ struct tcphdr {
 };
 
 struct udphdr {
-	__be16	source;
-    __be16	dest;
+	__be16 source;
+	__be16 dest;
 };
 
 enum kernel_read_file_id {
--- a/tracee-ebpf/tracee/events_processor.go
+++ b/tracee-ebpf/tracee/events_processor.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"unsafe"
 
 	"github.com/aquasecurity/tracee/pkg/containers"
 	"github.com/aquasecurity/tracee/pkg/external"
@@ -84,6 +85,10 @@ func (t *Tracee) shouldProcessEvent(ctx *context, args []external.Argument) bool
 	return true
 }
 
+var processTreeMap = make(map[uint32]external.Process_ctx)
+
+//proc[0] = 0
+
 func (t *Tracee) processEvent(event *external.Event) error {
 	switch int32(event.EventID) {
 
@@ -129,7 +134,9 @@ func (t *Tracee) processEvent(event *external.Event) error {
 		} else {
 			t.pidsInMntns.AddBucketItem(uint32(event.MountNS), uint32(event.HostProcessID))
 		}
-
+		processData := external.Process_ctx{ctx.Ts, ctx.CgroupID, ctx.Pid, ctx.Tid, ctx.Ppid, ctx.HostPid, ctx.HostTid, ctx.HostPpid, ctx.Uid, ctx.MntID, ctx.PidID}
+		processTreeMap[ctx.Tid] = processData
+		fmt.Println(processTreeMap)
 		//capture executed files
 		if t.config.Capture.Exec || t.config.Output.ExecHash {
 			filePath, err := getEventArgStringVal(event, "pathname")
@@ -213,7 +220,22 @@ func (t *Tracee) processEvent(event *external.Event) error {
 			}
 			return err
 		}
+	case SchedProcessExitEventID:
+		processContextMap, err := t.bpfModule.GetMap("process_context_map")
+		if err != nil {
+			t.Close()
+			return err
+		}
 
+		_ = processContextMap.DeleteKey(unsafe.Pointer(&ctx.Tid))
+		delete(processTreeMap, ctx.Tid)
+	case SchedProcessForkEventID:
+		childTid, ok := args["child_tid"].(uint32)
+		if !ok {
+			return fmt.Errorf("error parsing child_tid arg SchedProcessForkEventID")
+		}
+		processData := external.Process_ctx{ctx.Ts, ctx.CgroupID, ctx.Pid, ctx.Tid, ctx.Ppid, ctx.HostPid, ctx.HostTid, ctx.HostPpid, ctx.Uid, ctx.MntID, ctx.PidID}
+		processTreeMap[childTid] = processData
 	case CgroupMkdirEventID:
 		cgroupId, err := getEventArgUint64Val(event, "cgroup_id")
 		if err != nil {
--- a/tracee-ebpf/tracee/events_processor.go
+++ b/tracee-ebpf/tracee/events_processor.go
@@ -87,8 +87,6 @@ func (t *Tracee) shouldProcessEvent(ctx *context, args []external.Argument) bool
 
 var processTreeMap = make(map[uint32]external.Process_ctx)
 
-//proc[0] = 0
-
 func (t *Tracee) processEvent(event *external.Event) error {
 	switch int32(event.EventID) {
 
@@ -128,15 +126,15 @@ func (t *Tracee) processEvent(event *external.Event) error {
 		}
 
 	case SchedProcessExecEventID:
+		//update the process tree
+		processData := external.Process_ctx{ctx.Ts, ctx.CgroupID, ctx.Pid, ctx.Tid, ctx.Ppid, ctx.HostPid, ctx.HostTid, ctx.HostPpid, ctx.Uid, ctx.MntID, ctx.PidID}
+		processTreeMap[ctx.HostTid] = processData
 		//cache this pid by it's mnt ns
 		if event.ProcessID == 1 {
 			t.pidsInMntns.ForceAddBucketItem(uint32(event.MountNS), uint32(event.HostProcessID))
 		} else {
 			t.pidsInMntns.AddBucketItem(uint32(event.MountNS), uint32(event.HostProcessID))
 		}
-		processData := external.Process_ctx{ctx.Ts, ctx.CgroupID, ctx.Pid, ctx.Tid, ctx.Ppid, ctx.HostPid, ctx.HostTid, ctx.HostPpid, ctx.Uid, ctx.MntID, ctx.PidID}
-		processTreeMap[ctx.Tid] = processData
-		fmt.Println(processTreeMap)
 		//capture executed files
 		if t.config.Capture.Exec || t.config.Output.ExecHash {
 			filePath, err := getEventArgStringVal(event, "pathname")
@@ -221,14 +219,7 @@ func (t *Tracee) processEvent(event *external.Event) error {
 			return err
 		}
 	case SchedProcessExitEventID:
-		processContextMap, err := t.bpfModule.GetMap("process_context_map")
-		if err != nil {
-			t.Close()
-			return err
-		}
-
-		_ = processContextMap.DeleteKey(unsafe.Pointer(&ctx.Tid))
-		delete(processTreeMap, ctx.Tid)
+		delete(processTreeMap, ctx.HostTid)
 	case SchedProcessForkEventID:
 		childTid, ok := args["child_tid"].(uint32)
 		if !ok {
--- a/tracee-ebpf/tracee/events_processor.go
+++ b/tracee-ebpf/tracee/events_processor.go
@@ -7,7 +7,6 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
-	"unsafe"
 
 	"github.com/aquasecurity/tracee/pkg/containers"
 	"github.com/aquasecurity/tracee/pkg/external"
@@ -85,7 +84,6 @@ func (t *Tracee) shouldProcessEvent(ctx *context, args []external.Argument) bool
 	return true
 }
 
-
 func (t *Tracee) processEvent(event *external.Event) error {
 	switch int32(event.EventID) {
 
@@ -126,8 +124,8 @@ func (t *Tracee) processEvent(event *external.Event) error {
 
 	case SchedProcessExecEventID:
 		//update the process tree
-		processData := external.ProcessCtx{event.Timestamp, event.ContainerID, event.ProcessID, event.ThreadID, event.ParentProcessID, event.HostProcessID, event.HostThreadID, event.HostParentProcessID, event.UserID, event.MountNS, event.PIDNS}
-		processTreeMap[event.HostThreadID] = processData
+		processData := ProcessCtx{event.Timestamp, event.ContainerID, uint32(event.ProcessID), uint32(event.ThreadID), uint32(event.ParentProcessID), uint32(event.HostProcessID), uint32(event.HostThreadID), uint32(event.HostParentProcessID), uint32(event.UserID), uint32(event.MountNS), uint32(event.PIDNS)}
+		t.processTree.processTreeMap[event.HostThreadID] = processData
 		//cache this pid by it's mnt ns
 		if event.ProcessID == 1 {
 			t.pidsInMntns.ForceAddBucketItem(uint32(event.MountNS), uint32(event.HostProcessID))
@@ -218,16 +216,16 @@ func (t *Tracee) processEvent(event *external.Event) error {
 			return err
 		}
 	case SchedProcessExitEventID:
-		delete(processTreeMap, event.HostProcessID)
+		delete(t.processTree.processTreeMap, event.HostProcessID)
 	case SchedProcessForkEventID:
-		childTid, err := getEventArgUint64Val(event, "child_tid")
-		if err != nil {
-			return fmt.Errorf("error parsing child_tid arg SchedProcessForkEventID")
-		}
-		//processData := external.ProcessCtx{ctx.Ts, ctx.CgroupID, ctx.Pid, ctx.Tid, ctx.Ppid, ctx.HostPid, ctx.HostTid, ctx.HostPpid, ctx.Uid, ctx.MntID, ctx.PidID}
-		processData := external.ProcessCtx{event.Timestamp, event.ContainerID, event.ProcessID, event.ThreadID, event.ParentProcessID, event.HostProcessID, event.HostThreadID, event.HostParentProcessID, event.UserID, event.MountNS, event.ProcessID}
-		processTreeMap[event.HostThreadID] = processData
-		processTreeMap[int(childTid)] = processData
+		hostTid, _ := getEventArgUint32Val(event, "child_tid")
+		childPid, _ := getEventArgUint32Val(event, "child_pid")
+		pid, _ := getEventArgUint32Val(event, "child_ns_pid")
+		ppid, _ := getEventArgUint32Val(event, "parent_ns_pid")
+		hostPpid, _ := getEventArgUint32Val(event, "parent_pid")
+		childTid, _ := getEventArgUint32Val(event, "child_ns_tid")
+		processData := ProcessCtx{event.Timestamp, event.ContainerID, pid, childPid, ppid, hostTid, hostTid, hostPpid, uint32(event.UserID), uint32(event.MountNS), uint32(event.ProcessID)}
+		t.processTree.processTreeMap[int(childTid)] = processData
 	case CgroupMkdirEventID:
 		cgroupId, err := getEventArgUint64Val(event, "cgroup_id")
 		if err != nil {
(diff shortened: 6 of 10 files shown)


itamarmaouda101:
`Process_ctx struct {
	Ctime     uint64 
	Cgroup_id uint64
	Pid       uint32 
	Tid       uint32 
	Ppid      uint32 
	Host_pid  uint32
	Host_tid  uint32 
	Host_ppid uint32 
	Uid       uint32
	Mnt_id    uint32
	Pid_id    uint32
}
`
this is the process-data of every host_tid in the map

tracee-ebpf/tracee/tracee.go
```diff
@@ -905,6 +908,157 @@ func (t *Tracee) initBPF() error {
...
+		for _, task := range processTasks {
+			taskStatus := fmt.Sprintf("/proc/%d/task/%v/status", pid, task)
+			data, err := ioutil.ReadFile(taskStatus)
+			if err != nil {
+				t.handleError(err)
+			}
+			dataStatus := strings.Fields(string(data))
+			processStatus, err := parseProcStatus(dataStatus, taskStatus)
+			if err != nil {
+				t.handleError(err)
```
roikol, line 1054: also `continue`

tracee-ebpf/tracee/net_events.go
```diff
@@ -63,17 +62,43 @@ func (t *Tracee) processNetEvents() {
...
+							nsec,
+							comm,
+							hostTid,
+							pktLen,
+							netaddr.IPFrom16(pktMeta.SrcIP),
+							pktMeta.SrcPort,
+							netaddr.IPFrom16(pktMeta.DestIP),
+							pktMeta.DestPort,
+							pktMeta.Protocol)
+					} else {
```
roikol, line 83: are there any cases in which the `t.getProcessCtx(hostTid)` doesn't return the process context?

tracee-ebpf/tracee/co-re/vmlinux.h
```diff
@@ -705,8 +705,10 @@ struct tcphdr {
 };
 
 struct udphdr {
-	__be16 source;
-	__be16 dest;
+	__be16	source;
+    __be16	dest;
+    __be16	len;
```
roikol, line 710: are these fields used in this PR?

pkg/external/external.go
```diff
@@ -237,3 +238,36 @@ func (alert MemProtAlert) String() string {
 		return "Unknown alert"
 	}
 }
+
+type Process_ctx struct {
```
roikol, line 242: the convention is ThisKindOfCase

roikol (commented):
looks good!

tracee-ebpf/tracee/tracee.go
itamarmaouda101: fixed

tracee-ebpf/tracee/co-re/vmlinux.h
itamarmaouda101: nope i think its due the dns_eent's ill remove it

tracee-ebpf/tracee/net_events.go
itamarmaouda101: yep `t.getProcessCtx(hostTid)` returns the process context error

pkg/external/external.go
itamarmaouda101: u mean pascal-case? 
and no problem(:

roikol: yeah PascalCase ;)

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -507,6 +523,7 @@ typedef struct net_ctx_ext {
     __be16 local_port;
 } net_ctx_ext_t;
 
+
```
yanivagman, line 526: No need for this extra line

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1119,6 +1137,22 @@ static __always_inline int get_kconfig_val(u32 key)
...
+    context->host_pid = id >> 32;
+    context->host_ppid = get_task_ppid(task);
+    context->tid = get_task_ns_pid(task);
+    context->pid = get_task_ns_tgid(task);
+    context->ppid = get_task_ns_ppid(task);
+    context->mnt_id = get_task_mnt_ns_id(task);
+    context->pid_id = get_task_pid_ns_id(task);
+    context->uid = bpf_get_current_uid_gid();
+    context->ts = bpf_ktime_get_ns();
+    return 0;
```
yanivagman, line 1153: Why not using the already initialized context?

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1119,6 +1137,22 @@ static __always_inline int get_kconfig_val(u32 key)
...
+}
+
```
yanivagman, line 1155: What about the cgroup id (required for the resolution of container id)?

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2348,6 +2382,10 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
     struct task_struct *parent = (struct task_struct*)ctx->args[0];
     struct task_struct *child = (struct task_struct*)ctx->args[1];
 
+    process_context_t process = {};
+    init_process_context(&process, parent);
+    process.tid = get_task_ns_pid(child);
```
yanivagman, line 2387: The host tid is also different, but you don't update it here

tracee-ebpf/tracee/tracee.go
```diff
@@ -1,12 +1,15 @@
 package tracee
 
 import (
+	//"bytes"
```
yanivagman, line 4: Please remove if not used

tracee-ebpf/tracee/tracee.go
```diff
@@ -1,12 +1,15 @@
...
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
+	//"log"
```
yanivagman, line 12: Please remove if not used

tracee-ebpf/tracee/tracee.go
```diff
@@ -906,6 +909,157 @@ func (t *Tracee) initBPF() error {
 	return nil
 }
 
+func (t *Tracee) getProcessCtx(hostTid int) (external.ProcessCtx, error) {
+	processCtx, procExist := processTreeMap[hostTid]
+	if procExist {
+		//fmt.Printf("network process data: pid: %d, ctime: %v, ppid: %v, cgroup_id: %v\n",data.Pid, time.Unix(0, int64(data.Ctime+t.bootTime)), data.Ppid, data.Cgroup_id)
```
yanivagman, line 915: Please remove

tracee-ebpf/tracee/tracee.go
```diff
@@ -906,6 +909,157 @@ func (t *Tracee) initBPF() error {
...
+		return processCtx, nil
+	} else {
+
+		processContextMap, err := t.bpfModule.GetMap("process_context_map")
+		if err != nil {
+			t.Close()
```
yanivagman, line 921: Why calling Close here? This is not the responsibility of this function

tracee-ebpf/tracee/tracee.go
```diff
@@ -906,6 +909,157 @@ func (t *Tracee) initBPF() error {
...
+		}
+		processCtx = external.SetProcessContext(processCtxBpfMap)
+		return processCtx, nil
+	}
+}
+
+func timespecToTime(ts syscall.Timespec) time.Time {
+	return time.Unix(int64(ts.Sec), int64(ts.Nsec))
+}
+func getFileCtime(path string) (time.Time, error) {
```
yanivagman, line 936: Please add an empty line before

tracee-ebpf/tracee/tracee.go
```diff
@@ -906,6 +909,157 @@ func (t *Tracee) initBPF() error {
...
+	if err != nil {
+		return ctime, err
+	}
+	stat_t := fileInfo.Sys().(*syscall.Stat_t)
+	return timespecToTime(stat_t.Ctim), nil
+}
+
+func parseProcStatus(status []string, taskName string) (external.ProcessCtx, error) {
+	var process external.ProcessCtx
+	processFileds := []string{"Tgid:", "Pid:", "PPid:", "Uid:", "NStgid:", "NSpid:", "NSpgid:"}
```
yanivagman, line 948: Typo: Fields

tracee-ebpf/tracee/tracee.go
```diff
@@ -920,6 +1074,7 @@ func (t *Tracee) writeProfilerStats(wr io.Writer) error {
 
 // Run starts the trace. it will run until interrupted
 func (t *Tracee) Run() error {
+	t.initProcessTree()
```
yanivagman, line 1077: I think `New()` is a better place to call this function from

tracee-ebpf/tracee/tracee.go
```diff
@@ -906,6 +909,157 @@ func (t *Tracee) initBPF() error {
...
+		return ctime, err
+	}
+	stat_t := fileInfo.Sys().(*syscall.Stat_t)
+	return timespecToTime(stat_t.Ctim), nil
+}
+
+func parseProcStatus(status []string, taskName string) (external.ProcessCtx, error) {
+	var process external.ProcessCtx
+	processFileds := []string{"Tgid:", "Pid:", "PPid:", "Uid:", "NStgid:", "NSpid:", "NSpgid:"}
+	i := 0
```
yanivagman, line 949: You can initialize this varialbe as part of the for loop

tracee-ebpf/tracee/tracee.go
```diff
@@ -906,6 +909,157 @@ func (t *Tracee) initBPF() error {
...
+				return process, err
+			}
+			processVals = append(processVals, int(filed))
+		}
+		if i > len(processFileds)-1 {
+			break
+		}
+
+	}
+	processCtime, err := getFileCtime(taskName)
```
yanivagman, line 966: What does process ctime means? Start time?
If so, is this the real process start time as represnted by the kernel? I know this information exists in /proc/pid/stat

tracee-ebpf/tracee/tracee.go
```diff
@@ -906,6 +909,157 @@ func (t *Tracee) initBPF() error {
...
+		return process, err
+	}
+	process.Ctime = int(processCtime.Unix())
+	process.HostPid = processVals[0]
+	process.HostTid = processVals[1]
+	process.HostPpid = processVals[2]
+	process.Uid = processVals[3]
+	process.Pid = processVals[4]
+	process.Tid = processVals[5]
+	process.Ppid = processVals[6]
```
yanivagman, line 977: Instead of using numbers [0-6], it will be better to have names.
You can use a map instead of an array to save and get the values, which will be clearer

tracee-ebpf/tracee/tracee.go
```diff
@@ -906,6 +909,157 @@ func (t *Tracee) initBPF() error {
...
+	mntId, pidId, CgroupId, err := getNsIdData(taskName)
+	if err != nil {
+		return process, err
+	}
+	process.MntId = int(mntId)
+	process.PidId = int(pidId)
+	process.ContainerID = fmt.Sprint(CgroupId)
```
yanivagman, line 984: Cgroup Id is not the container id.
See containers.go for an API to get the container id from cgroup id

tracee-ebpf/tracee/tracee.go
```diff
@@ -906,6 +909,157 @@ func (t *Tracee) initBPF() error {
...
+	if err != nil {
+		return 0, 0, 0, err
+	}
+	processPidId = strings.TrimLeft(strings.TrimRight(processPidId, "]"), "pid:[")
+	pidId, err := strconv.ParseUint(processPidId, 10, 32)
+	if err != nil {
+		return 0, 0, 0, err
+	}
+	path = fmt.Sprintf("%s/ns/cgroup", taskName[:len(taskName)-7])
+	processCgroupId, err := os.Readlink(path)
```
yanivagman, line 1010: This is not the cgroup id, but the cgroup namespace id.

yanivagman:
In addition to the above comments - please move the new process tree logic into a new file instead of tracee.go

tracee-ebpf/tracee/net_events.go
```diff
@@ -63,17 +62,43 @@ func (t *Tracee) processNetEvents() {
...
+							comm,
+							networkProcess.Pid,
+							networkProcess.Tid,
+							0,
+							pktLen,
+							netaddr.IPFrom16(pktMeta.SrcIP),
+							pktMeta.SrcPort,
+							netaddr.IPFrom16(pktMeta.DestIP),
+							pktMeta.DestPort,
+							pktMeta.Protocol)
```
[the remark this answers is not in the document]
itamarmaouda101: yes sure!
i made another PR that will add it .
the reason for 2 different PRs is because this one only adding the process context map management.
the second PR will take some data from the process context map and will create a external.Event struct and send it to the events channel

cmd/tracee-ebpf/printer.go
```diff
@@ -233,6 +233,7 @@ func (p *gobEventPrinter) Init() error {
 	p.outEnc = gob.NewEncoder(p.out)
 	gob.Register(external.Event{})
 	gob.Register(external.SlimCred{})
+	gob.Register(external.ProcessCtx{})
```
yanivagman, line 236: Why do we need to register this struct?

cmd/tracee-rules/input.go
```diff
@@ -47,6 +47,7 @@ func setupTraceeGobInputSource(opts *traceeInputOptions) (chan types.Event, erro
 	dec := gob.NewDecoder(opts.inputFile)
 	gob.Register(external.Event{})
 	gob.Register(external.SlimCred{})
+	gob.Register(external.ProcessCtx{})
```
yanivagman, line 50: same

pkg/external/external.go
```diff
@@ -237,3 +237,17 @@ func (alert MemProtAlert) String() string {
...
+	Pid         int
+	Tid         int
+	Ppid        int
+	HostPid     int
+	HostTid     int
+	HostPpid    int
+	Uid         int
+	MntId       int
+	PidId       int
+}
```
yanivagman, line 253: Do we really need it in external? Is it exposed in anyway?

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2381,7 +2400,7 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
 
         save_to_submit_buf(&data, (void*)&parent_pid, sizeof(int), 0);
         save_to_submit_buf(&data, (void*)&parent_ns_pid, sizeof(int), 1);
-        save_to_submit_buf(&data, (void*)&parent_tgid, sizeof(int), 2);
+            save_to_submit_buf(&data, (void*)&parent_tgid, sizeof(int), 2);
```
yanivagman, line 2403: No need for this indentation

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,225 @@
...
+			pref, err := getCgroupDirPrefix()
+			if err != nil || pref == "unknown" {
+				return "unknown", err
+			}
+			path := fmt.Sprintf("%s%s", pref, cgroupData[2])
+			containerId, _ = containers.GetContainerIdFromCgroup(path)
+		}
+	}
+	return containerId, nil
+}
```
yanivagman, line 182: containers related logic should go into containers.go

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,225 @@
...
+)
+
+type ProcessTree struct {
+	processTreeMap map[int]external.ProcessCtx
+}
+
+func SetProcessContext(ctx []byte) external.ProcessCtx {
+	var procCtx = external.ProcessCtx{}
+	procCtx.StartTime = int(binary.LittleEndian.Uint64(ctx[0:8]))
+	procCtx.ContainerID = fmt.Sprintf("%v", binary.LittleEndian.Uint64(ctx[8:16]))
```
yanivagman, line 26: From bpf code you get the cgroup id, and not the container id

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,225 @@
+package tracee
+
+import (
+	"bufio"
+	"encoding/binary"
+	"fmt"
+	"github.com/aquasecurity/tracee/pkg/containers"
```
yanivagman, line 7: You already import this package below

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,225 @@
...
+	"io/ioutil"
+	"os"
+	"strconv"
+	"strings"
+	"syscall"
+	"time"
+
+	_ "github.com/aquasecurity/tracee/pkg/containers"
```
yanivagman, line 15: Why do you need to import with _?

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,225 @@
...
+	return process, nil
+}
+
+func getNsIdData(taskName string) (uint32, uint32, error) {
+	path := fmt.Sprintf("%s/ns/mnt", taskName[:len(taskName)-7])
+	processMntId, err := os.Readlink(path)
+	if err != nil {
+		return 0, 0, err
+	}
+	processMntId = strings.TrimLeft(strings.TrimRight(processMntId, "]"), "mnt:[")
```
yanivagman, line 99: You can also use TrimPrefix and TrimSuffix which I think will be clearer

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,225 @@
...
+		}
+	}
+	return "unknown", err
+}
+func cgroupV1() bool {
+	if _, err := os.Stat("/sys/fs/cgroup/cgroup.controllers"); os.IsNotExist(err) {
+		return true
+	}
+	return false
+}
```
yanivagman, line 145: Duplication of code from containers.go

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,225 @@
...
+		processTasks, err := taskDir.Readdirnames(-1)
+		for _, task := range processTasks {
+			taskDir := fmt.Sprintf("/proc/%d/task/%v", pid, task)
+			taskStatus := fmt.Sprintf("/proc/%d/task/%v/status", pid, task)
+			data, err := ioutil.ReadFile(taskStatus)
+			if err != nil {
+				continue
+			}
+			dataStatus := strings.Fields(string(data))
+			//task := fmt.Sprintf("/proc/%d/task/%v", pid, task)
```
yanivagman, line 211: Please remove comment

tracee-ebpf/tracee/co-re/vmlinux.h
```diff
@@ -705,8 +705,8 @@ struct tcphdr {
 };
 
 struct udphdr {
-	__be16 source;
-	__be16 dest;
+	__be16	source;
+    __be16	dest;
```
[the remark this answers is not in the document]
yanivagman: ?

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2402,6 +2421,8 @@ int tracepoint__sched__sched_process_exec(struct bpf_raw_tracepoint_args *ctx)
     event_data_t data = {};
     if (!init_event_data(&data, ctx))
         return 0;
+    process_context_t *process = (process_context_t*) &data.context;
+
```
yanivagman, line 2425: No need this extra line (maybe before the previous line?)

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,225 @@
...
+	procCtx.MntId = int(binary.LittleEndian.Uint32(ctx[44:48]))
+	procCtx.PidId = int(binary.LittleEndian.Uint32(ctx[48:52]))
+	return procCtx
+}
+
+func timespecToTime(ts syscall.Timespec) time.Time {
+	return time.Unix(int64(ts.Sec), int64(ts.Nsec))
+}
+
+func getFileStartTime(path string) (time.Time, error) {
```
yanivagman, line 43: `getFileCtime` was a better name. There is no such thing File start time.

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,225 @@
...
+	stat_t := fileInfo.Sys().(*syscall.Stat_t)
+	return timespecToTime(stat_t.Ctim), nil
+}
+
+func parseProcStatus(status []string, taskName string) (external.ProcessCtx, error) {
+	var process external.ProcessCtx
+	processFields := []string{"Tgid", "Pid", "PPid", "Uid", "NStgid", "NSpid", "NSpgid"}
+	processVals := make(map[string]int)
+	i := 0
+	//var processVals []int
```
yanivagman, line 58: please remove

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+	stat_t := fileInfo.Sys().(*syscall.Stat_t)
+	return timespecToTime(stat_t.Ctim), nil
+}
+
+func parseProcStatus(status []string, taskName string) (external.ProcessCtx, error) {
+	var process external.ProcessCtx
+	processFields := []string{"Tgid", "Pid", "PPid", "Uid", "NStgid", "NSpid", "NSpgid"}
+	processVals := make(map[string]int)
+	i := 0
```
yanivagman, line 57: You can initialize this variable as part of the loop

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,225 @@
...
+			if err != nil {
+				return process, err
+			}
+			processVals[processFields[i]] = int(field)
+		}
+		if i > len(processFields)-1 {
+			break
+		}
+	}
+	processCtime, err := getFileStartTime(taskName)
```
yanivagman, line 72: processCtime -> processStartTime
Also, is this the real process start time as represnted by the kernel?
You can get the real process start time from /proc/pid/stat

cmd/tracee-ebpf/printer.go
itamarmaouda101: yes 
it's for the passing the first "snapshot of process tree to tracee rules

itamarmaouda101: i cant remove it and we  will pass it with the json printer , if u willing to let me know (:

yanivagman: But I don't see such passing of the first snapshot.
If not part of this PR - let's add this in a future PR. Is that possible?

itamarmaouda101: yes sure!

tracee-ebpf/tracee/process_tree.go
itamarmaouda101: yes and no, i looks in /proc/pid/stat and /proc/pid/status for the process Ctime but i couldn't find it. 
so i can up with idea to take the directory Ctime instead due the reason its created in the task creation.
if u willing to get the exact Ctime i can give another look on that but when i tested it it seemed to work fine

itamarmaouda101: Btw afaik the kernel itself create this dirs in the process creation so i don't think there is to supposed to be a margin between the times

yanivagman: Look for starttime in https://man7.org/linux/man-pages/man5/proc.5.html

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,211 @@
...
+func getNsIdData(taskName string) (uint32, uint32, error) {
+	path := fmt.Sprintf("%s/ns/mnt", taskName[:len(taskName)-7])
+	processMntId, err := os.Readlink(path)
+	if err != nil {
+		return 0, 0, err
+	}
+
+	processMntId = strings.TrimSuffix(processMntId, "]")
+	processMntId = strings.TrimPrefix(processMntId, "mnt:[")
+	//processMntId = strings.TrimLeft(strings.TrimRight(processMntId, "]"), "mnt:[")
```
yanivagman, line 116: Please remove

pkg/containers/containers.go
```diff
@@ -254,6 +254,35 @@ func (c *Containers) GetContainers() []string {
 	}
 	return conts
 }
+func CgroupV1() bool {
+	if _, err := os.Stat("/sys/fs/cgroup/cgroup.controllers"); os.IsNotExist(err) {
+		return true
+	}
+	return false
+}
```
yanivagman, line 262: We already have IsCgroupV1 function in this file

pkg/containers/containers.go
```diff
@@ -254,6 +254,35 @@ func (c *Containers) GetContainers() []string {
...
+		if isCgroupV1 {
+			if strings.Contains(mountpoint, "cpuset") && fstype == "cgroup" {
+				return mountpoint, nil
+			}
+		} else if fstype == "cgroup2" {
+			return mountpoint, nil
+		}
+	}
+	return "unknown", err
+}
```
yanivagman, line 285: How is this function different from procMountsCgroups()?

tracee-ebpf/tracee/events_processor.go
```diff
@@ -2,14 +2,13 @@ package tracee
...
+	"github.com/aquasecurity/tracee/pkg/containers"
+	"github.com/aquasecurity/tracee/pkg/external"
 	"io"
 	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
-
-	"github.com/aquasecurity/tracee/pkg/containers"
-	"github.com/aquasecurity/tracee/pkg/external"
```
yanivagman, line 12: Let's keep those imports here

tracee-ebpf/tracee/process_tree.go
yanivagman: This comment is still valid

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,211 @@
...
+//returns the file creation time
+//Note: this is not the real process start time due its based on the file creation time, but it is a close approximation
+func processStartTime(path string) (time.Time, error) {
+	var ctime time.Time
+	fileInfo, err := os.Stat(path)
+	if err != nil {
+		return ctime, err
+	}
+	stat_t := fileInfo.Sys().(*syscall.Stat_t)
+	return timespecToTime(stat_t.Ctim), nil
```
yanivagman, line 64: You only call this (one line) function here, let's put it inline

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,211 @@
...
+	mntId, err := strconv.ParseUint(processMntId, 10, 32)
+	if err != nil {
+		return 0, 0, err
+	}
+	path = fmt.Sprintf("%s/ns/pid", taskName[:len(taskName)-7])
+	processPidId, err := os.Readlink(path)
+	if err != nil {
+		return 0, 0, err
+	}
+	processPidId = strings.TrimLeft(strings.TrimRight(processPidId, "]"), "pid:[")
```
yanivagman, line 126: If you chose to use TrimSuffix/Prefix above - why not doing the same here?

pkg/containers/containers.go
```diff
@@ -178,7 +178,7 @@ func (c *Containers) CgroupUpdate(cgroupId uint64, path string) (CgroupInfo, err
 }
 
 // extract container id and container runtime from path
-func getContainerIdFromCgroup(cgroupPath string) (string, string) {
+func GetContainerIdFromCgroup(cgroupPath string) (string, string) {
```
yanivagman, line 181: You can move getContainerID function to this file as well.
Then, you can avoid exporting this function (and any other) and only expose getContainerID, which can be renamed to GetTaskContainerID

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,174 @@
...
+	procCtx.HostTid = int(binary.LittleEndian.Uint32(ctx[32:36]))
+	procCtx.HostPpid = int(binary.LittleEndian.Uint32(ctx[36:40]))
+	procCtx.Uid = int(binary.LittleEndian.Uint32(ctx[40:44]))
+	procCtx.MntId = int(binary.LittleEndian.Uint32(ctx[44:48]))
+	procCtx.PidId = int(binary.LittleEndian.Uint32(ctx[48:52]))
+	return procCtx
+}
+
+//returns the file creation time
+//Note: this is not the real process start time due its based on the file creation time, but it is a close approximation
```
itamarmaouda101, line 52: due -> because
its -> it's

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,174 @@
...
+	procCtx.HostPid = int(binary.LittleEndian.Uint32(ctx[28:32]))
+	procCtx.HostTid = int(binary.LittleEndian.Uint32(ctx[32:36]))
+	procCtx.HostPpid = int(binary.LittleEndian.Uint32(ctx[36:40]))
+	procCtx.Uid = int(binary.LittleEndian.Uint32(ctx[40:44]))
+	procCtx.MntId = int(binary.LittleEndian.Uint32(ctx[44:48]))
+	procCtx.PidId = int(binary.LittleEndian.Uint32(ctx[48:52]))
+	return procCtx
+}
+
+//returns the file creation time
```
itamarmaouda101, line 51: This comment is unclear. What file?

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,174 @@
...
+//Note: this is not the real process start time due its based on the file creation time, but it is a close approximation
+func processStartTime(path string) (time.Time, error) {
```
itamarmaouda101, line 53: I think this should just be "getFileCreationTime", and just assign it to a variable named "processStartTime"

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,174 @@
...
+	return time.Unix(int64(ts.Sec), int64(ts.Nsec)), nil
+}
+
+//parse the status file of a process or given task and returns process context struct
+func parseProcStatus(status []byte, taskStatusPath string) (ProcessCtx, error) {
+	dataStatus := strings.Fields(string(status))
+	var process ProcessCtx
+	processFields := []string{"Tgid", "Pid", "PPid", "Uid", "NStgid", "NSpid", "NSpgid"}
+	processVals := make(map[string]int)
+	i := 0
```
itamarmaouda101, line 70:
```suggestion
	fieldIndex := 0
```

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,174 @@
...
+	for idx, val := range dataStatus {
+		if strings.Contains(val, processFields[i]) {
+			field, err := strconv.ParseUint(dataStatus[idx+1], 10, 32)
+			if err != nil {
+				return process, err
+			}
+			processVals[processFields[i]] = int(field)
+			i++
+			if strings.Contains(val, "NSpgid") {
```
itamarmaouda101, line 79: Comment why this is the break for the function

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,174 @@
...
+	stat_t := fileInfo.Sys().(*syscall.Stat_t)
+	ts := stat_t.Ctim
+	return time.Unix(int64(ts.Sec), int64(ts.Nsec)), nil
+}
+
+//parse the status file of a process or given task and returns process context struct
+func parseProcStatus(status []byte, taskStatusPath string) (ProcessCtx, error) {
+	dataStatus := strings.Fields(string(status))
+	var process ProcessCtx
+	processFields := []string{"Tgid", "Pid", "PPid", "Uid", "NStgid", "NSpid", "NSpgid"}
```
itamarmaouda101, line 68: Maybe use consts for these strings because they are used in other places

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,174 @@
...
+	processVals := make(map[string]int)
+	i := 0
+	for idx, val := range dataStatus {
+		if strings.Contains(val, processFields[i]) {
+			field, err := strconv.ParseUint(dataStatus[idx+1], 10, 32)
```
itamarmaouda101, line 73: Maybe check that idx+1 is not over the limits

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,174 @@
...
+		return process, err
+	}
+	process.MntId = int(mntId)
+	process.PidId = int(pidId)
+	return process, nil
+}
+
+//gets the namespace data for the process context struct by parsing the /proc/<Pid>/task directory
+func getNsIdData(taskStatusPath string) (uint32, uint32, error) {
+	path := fmt.Sprintf("%s/ns/mnt", taskStatusPath[:len(taskStatusPath)-7])
```
itamarmaouda101, line 107: Use a const instead of the magic '-7'

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,174 @@
...
+
+//initialize new process-tree
+func NewProcessTree() (*ProcessTree, error) {
+	p := ProcessTree{make(map[int]ProcessCtx)}
+	procDir, err := os.Open("/proc")
+	if err != nil {
+		return nil, err
+	}
+	defer procDir.Close()
+	entries, err := procDir.Readdirnames(-1)
```
itamarmaouda101, line 141: Magic

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	// Iterate over each pid
+	for _, procEntry := range entries {
+		pid, err := strconv.ParseUint(procEntry, 10, 32)
+		if err != nil {
+			continue
+		}
+		taskDir, err := os.Open(fmt.Sprintf("/proc/%d/task", pid))
+		processTasks, err := taskDir.Readdirnames(-1)
```
itamarmaouda101, line 153: Magic

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -547,6 +561,7 @@ BPF_HASH(bin_args_map, u64, bin_args_t);                // persist args for send
 BPF_HASH(sys_32_to_64_map, u32, u32);                   // map 32bit to 64bit syscalls
 BPF_HASH(params_types_map, u32, u64);                   // encoded parameters types for event
 BPF_HASH(process_tree_map, u32, u32);                   // filter events by the ancestry of the traced process
+BPF_HASH(process_context_map, u32, process_context_t);  // houlds the process_context data for every tid
```
itamarmaouda101, line 564: houlds -> holds

pkg/containers/containers.go
```diff
@@ -90,6 +90,47 @@ func (c *Containers) procMountsCgroups() error {
...
+func GetContainerIdFromTaskDir(taskPath string) (string, error) {
+	containerId := ""
+	taskPath = fmt.Sprintf("%s/cgroup", taskPath)
+	cgroupFile, err := os.Open(taskPath)
+	if err != nil {
+		return containerId, err
+	}
+	defer cgroupFile.Close()
+	scanner := bufio.NewScanner(cgroupFile)
+	taskContainer := InitContainers()
```
yanivagman, line 111: If you will make this a function of Containers struct, you won't have to initialize it again (calling it from the instance held by tracee).
On the other hand, that might not be neccessary and you can probably remove this line anyways - see my next comment

pkg/containers/containers.go
```diff
@@ -90,6 +90,47 @@ func (c *Containers) procMountsCgroups() error {
...
+		} else {
+			line := scanner.Text()
+			cgroupV2Suref := "0::/"
+			if strings.HasSuffix(line, cgroupV2Suref) {
+				continue
+			}
+			containerId = getContainerIdFromTaskContainer(taskContainer, cgroup)
+			break
+		}
+	}
```
yanivagman, line 130: You can just loop over the lines, sending each line to getContainerIdFromCgroup(), and break if container id is not empty. This should work for both cgroup v1 and cgroup v2, without even having to build the path string like you do in getContainerIdFromTaskContainer

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,177 @@
...
+func (t *Tracee) ParseProcessContext(ctx []byte) (ProcessCtx, error) {
+	var procCtx = ProcessCtx{}
+	procCtx.StartTime = int(binary.LittleEndian.Uint64(ctx[0:8]))
+	cgroupId := binary.LittleEndian.Uint64(ctx[8:16])
+	procCtx.ContainerID = t.containers.GetCgroupInfo(cgroupId).ContainerId
+	dataBuff := bytes.NewBuffer(ctx[16:])
+	err := binary.Read(dataBuff, binary.LittleEndian, &ctx)
+	if err != nil {
+		return procCtx, err
+	}
```
yanivagman, line 44: What's the meaning of reading ctx[16:] if you don't make use of it?
What about all the other ProcessCtx fields?

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,177 @@
...
+	mntId, pidId, err := getNsIdData(taskStatusPath)
+	if err != nil {
+		return process, err
+	}
+	process.MntId = uint32(mntId)
+	process.PidId = uint32(pidId)
+	return process, nil
+}
+
+const statusLen = len("/status")
```
yanivagman, line 104: Why do we need this as a constant?
I think it will be more readable inline

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,177 @@
...
+	processPidId = strings.TrimSuffix(processPidId, "]")
+	processPidId = strings.TrimPrefix(processPidId, "pid:[")
+	pidId, err := strconv.ParseUint(processPidId, 10, 32)
+	if err != nil {
+		return 0, 0, err
+	}
+	return uint32(mntId), uint32(pidId), nil
+}
+
+const getAsSingleSlice = -1
```
yanivagman, line 134: Again, I don't see a reason for having such a constant

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2504,6 +2526,9 @@ int tracepoint__sched__sched_process_exit(struct bpf_raw_tracepoint_args *ctx)
     bpf_map_delete_elem(&new_pids_map, &data.context.host_tid);
     bpf_map_delete_elem(&syscall_data_map, &data.context.host_tid);
 
+    //Remove this process from process tree map
```
yanivagman, line 2529: No need for these two lines

tracee-ebpf/tracee/tracee.go
```diff
@@ -918,6 +923,24 @@ func (t *Tracee) writeProfilerStats(wr io.Writer) error {
...
+	} else {
+		processContextMap, err := t.bpfModule.GetMap("process_context_map")
+		if err != nil {
+			return processCtx, err
+		}
+		processCtxBpfMap, err := processContextMap.GetValue(unsafe.Pointer(&hostTid))
+		if err != nil {
+			return processCtx, err
+		}
+		processCtx, err = t.ParseProcessContext(processCtxBpfMap)
```
yanivagman, line 939: Won't it be better to update processTreeMap before returning, so we will not have another cache miss?

tracee-ebpf/tracee/process_tree.go
itamarmaouda101: it's the offset in the byte array after the cgroup data.
i add a comment there that explain what is it (after ill pushed it updated )

tracee-ebpf/tracee/tracee.go
itamarmaouda101: yep i just added it

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,173 @@
...
+type ProcessTree struct {
+	processTreeMap map[int]ProcessCtx
+}
+
+func (t *Tracee) ParseProcessContext(ctx []byte) (ProcessCtx, error) {
+	var procCtx = ProcessCtx{}
+	procCtx.StartTime = int(binary.LittleEndian.Uint64(ctx[0:8]))
+	cgroupId := binary.LittleEndian.Uint64(ctx[8:16])
+	procCtx.ContainerID = t.containers.GetCgroupInfo(cgroupId).ContainerId
+	dataBuff := bytes.NewBuffer(ctx[16:]) // this is the offset after the cgroup and startTime in the ctx byte array
```
yanivagman, line 40: Can you please explain why do you read the rest of ctx?
And why you don't initialize the other fields of procCtx?

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,184 @@
...
+					return process, err
+				}
+				processVals[processFields[fieldIndex]] = uint32(field)
+				fieldIndex++
+				if strings.Contains(val, "NSpgid") {
+					break //for quick exit from the loop and evading looping on the other file content which is celery in that case
+				}
+			}
+		}
+	}
```
yanivagman, line 93:
```suggestion
	var process ProcessCtx
	processVals := make(map[string]uint32)
	for _, line := range strings.Split(string(status), "\n") {
		lineFields := strings.Split(line, ":")
		if len(lineFields) > 1 {
			valFields := strings.Fields(lineFields[1])
			val, err := strconv.ParseUint(strings.TrimSpace(valFields[0]), 10, 32)
			if err != nil {
				continue
			}
			processVals[lineFields[0]] = uint32(val)
		}
	}
```

tracee-ebpf/tracee/events_processor.go
```diff
@@ -213,7 +215,17 @@ func (t *Tracee) processEvent(event *external.Event) error {
...
+	case SchedProcessExitEventID:
+		delete(t.processTree.processTreeMap, event.HostProcessID)
+	case SchedProcessForkEventID:
+		hostTid, _ := getEventArgUint32Val(event, "child_tid")
+		childPid, _ := getEventArgUint32Val(event, "child_pid")
+		pid, _ := getEventArgUint32Val(event, "child_ns_pid")
+		ppid, _ := getEventArgUint32Val(event, "parent_ns_pid")
+		hostPpid, _ := getEventArgUint32Val(event, "parent_pid")
+		childTid, _ := getEventArgUint32Val(event, "child_ns_tid")
+		processData := ProcessCtx{event.Timestamp, event.ContainerID, uint32(pid), uint32(childTid), uint32(ppid), uint32(childPid), uint32(hostTid), uint32(hostPpid), uint32(event.UserID), uint32(event.MountNS), uint32(event.ProcessID)}
```
yanivagman, line 227: Why do you still cast the variables with uint32?

tracee-ebpf/tracee/process_tree.go
```diff
@@ -0,0 +1,174 @@
...
+	ts := stat_t.Ctim
+	return int(ts.Sec)*int(time.Second) + int(ts.Nsec)*int(time.Nanosecond), nil
+}
+
+//parse the status file of a process or given task and returns process context struct
+func parseProcStatus(status []byte, taskStatusPath string) (ProcessCtx, error) {
+	var process ProcessCtx
+	processVals := make(map[string]uint32)
+	for _, line := range strings.Split(string(status), "\n") {
+		lineFields := strings.Fields(line)
```
yanivagman, line 76: Why not using `lineFields := strings.Split(line, ":")` to have the keys without a colon?

yanivagman (approved):
LGTM

PR #1301, opened 2021-12-23
making thread safe cgroups map in Containers struct

mtcherni95:


mtcherni95:
This commit is required as we will get to `Containers` struct also from `processFileWrites` go routine and not only from `processEvents` (please see https://github.com/aquasecurity/tracee/pull/1291)

tracee-ebpf/tracee/containers.go
```diff
@@ -178,7 +182,9 @@ func (c *Containers) CgroupUpdate(cgroupId uint64, path string) (CgroupInfo, err
 		info.Runtime = runtime
 	}
```
mtcherni95, line 184: Note that here we always fall in the statement that updates the `c.cgroups `map, even if `info.ContainerId` and `info.Runtime` is not initialized. Is it intentional @yanivagman ?

yanivagman (approved):

PR #1304, opened 2021-12-23
fixed time printing syntax in net events

itamarmaouda101:
hi 
i saw that in the printing (as table in the terminal) of the net events there is a different style of the time syntax then the style in the regular events (that came from tracee- ebpf in the events buffer)

i changed it to the default style (which is only hour and not the full date)


itamarmaouda101:
Update: i closed that PR to changes that im working on in net_events.go - adding json format and adding data about the process that receives/sends the packet.
so ill upload all of that in one PR (:

Issue #1267, opened 2021-12-15
extract package `containers` from `containers.go`


itaysk:
from #960:

> containers - create new package `containers`. The code is already independent of tracee

yanivagman:
#1311

PR #1311, opened 2021-12-27
exporting containers.go to package

mtcherni95:


mtcherni95:
Solves: #1267

pkg/containers/containers.go
```diff
@@ -12,8 +13,6 @@ import (
 	"syscall"
 	"time"
 	"unsafe"
-
-	bpf "github.com/aquasecurity/libbpfgo"
```
yanivagman, line 16: Why removing it from here?

pkg/containers/containers.go
```diff
@@ -271,7 +270,7 @@ const (
 	containerStarted
 )
 
-func (c *Containers) PopulateBpfMap(bpfModule *bpf.Module) error {
+func (c *Containers) PopulateBpfMap(bpfModule *libbpfgo.Module) error {
 	containersMap, err := bpfModule.GetMap("containers_map")
```
yanivagman, line 274: containers_map shouldn't be hard-coded if we move this to a package.
Let's change this to get the map name as an additional arugment

pkg/containers/containers.go
```diff
@@ -287,7 +286,7 @@ func (c *Containers) PopulateBpfMap(bpfModule *bpf.Module) error {
 	return err
 }
 
-func RemoveFromBpfMap(bpfModule *bpf.Module, cgroupId uint64) error {
+func RemoveFromBpfMap(bpfModule *libbpfgo.Module, cgroupId uint64) error {
 	containersMap, err := bpfModule.GetMap("containers_map")
```
yanivagman, line 290: same

pkg/containers/containers.go
mtcherni95: Agree

yanivagman (approved):

PR #1317, opened 2022-01-02
tracee-ebpf: add container cgroup ctime

yanivagman:
This PR adds cgroup ctime information which can be used to track containers start time.
    
Instead of using the WalkDir function to find a specific cgroup, this PR implements a recursive function that is more efficient as it
only stats a directory if its name has a given suffix. By doing so, performance is 5 times better than the use of WalkDir for this specific case.

In addition, this PR fixes the following issue (thanks @rafaeldtinoco for spotting this):
When container filter is set, we might treat essential events such as cgroup_mkdir as "false container positives".
Such events should be processed before checking for false positives.
This PR fixes this by placing the check for false positives after processing the events

### DIFF
--- a/pkg/containers/containers.go
+++ b/pkg/containers/containers.go
@@ -2,7 +2,6 @@ package containers
 
 import (
 	"bufio"
-	"errors"
 	"fmt"
 	"io/fs"
 	"os"
@@ -30,6 +29,7 @@ type CgroupInfo struct {
 	Path        string
 	ContainerId string
 	Runtime     string
+	Ctime       time.Time
 	expiresAt   time.Time
 }
 
@@ -114,7 +114,7 @@ func (c *Containers) populate() error {
 			return nil
 		}
 
-		_, err = c.CgroupUpdate(stat.Ino, path)
+		_, err = c.CgroupUpdate(stat.Ino, path, time.Unix(stat.Ctim.Sec, stat.Ctim.Nsec))
 		return err
 	}
 
@@ -125,50 +125,10 @@ func (c *Containers) populate() error {
 	return filepath.WalkDir(c.cgroupMP, fn)
 }
 
-func (c *Containers) cgroupLookupUpdate(cgroupId uint64) error {
-	fn := func(path string, d fs.DirEntry, err error) error {
-		if !d.IsDir() {
-			return nil
-		}
-
-		if err != nil {
-			_, _ = fmt.Fprintln(os.Stderr, err)
-			return nil
-		}
-
-		// The lower 32 bits of the cgroup id are the inode number of the matching cgroupfs entry
-		var stat syscall.Stat_t
-		if err := syscall.Stat(path, &stat); err != nil {
-			return nil
-		}
-
-		// Check if this cgroup path belongs to cgroupId
-		if (stat.Ino & 0xFFFFFFFF) != (cgroupId & 0xFFFFFFFF) {
-			return nil
-		}
-
-		// If we reached this point, we found a match for cgroupId - update cgroups map and break search
-		c.CgroupUpdate(cgroupId, path)
-		return fs.ErrExist
-	}
-
-	err := filepath.WalkDir(c.cgroupMP, fn)
-	if errors.Is(err, fs.ErrExist) {
-		return nil
-	}
-	if err == nil {
-		// No match was found - update cgroup id with an empty entry
-		c.mtx.Lock()
-		c.cgroups[uint32(cgroupId)] = CgroupInfo{}
-		c.mtx.Unlock()
-	}
-	return err
-}
-
-// check if path belongs to a known container runtime and
-// add cgroupId with a matching container id, extracted from path
-func (c *Containers) CgroupUpdate(cgroupId uint64, path string) (CgroupInfo, error) {
-	info := CgroupInfo{Path: path}
+// CgroupUpdate checks if path belongs to a known container runtime and
+// updates cgroupId with a matching container id, extracted from path
+func (c *Containers) CgroupUpdate(cgroupId uint64, path string, ctime time.Time) (CgroupInfo, error) {
+	info := CgroupInfo{Path: path, Ctime: ctime}
 	info.ContainerId, info.Runtime = getContainerIdFromCgroup(path)
 
 	c.mtx.Lock()
@@ -177,7 +137,7 @@ func (c *Containers) CgroupUpdate(cgroupId uint64, path string) (CgroupInfo, err
 	return info, nil
 }
 
-// extract container id and container runtime from path
+// getContainerIdFromCgroup extracts container id and container runtime from path
 func getContainerIdFromCgroup(cgroupPath string) (string, string) {
 	prevPathComp := ""
 	for _, pc := range strings.Split(cgroupPath, "/") {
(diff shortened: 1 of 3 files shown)


itaysk:
Since this is not referencing any issue, can you please summarize what's the intention in this PR?

yanivagman:
> Since this is not referencing any issue, can you please summarize what's the intention in this PR?

Sorry, I just noticed this comment. With this in place, we will be able to create a container start event, which can be required by some signatures

pkg/containers/containers.go
```diff
@@ -246,6 +206,52 @@ func (c *Containers) CgroupRemove(cgroupId uint64) {
...
+	entries, err := os.ReadDir(rootDir)
+	if err != nil {
+		return "", err
+	}
+	for _, entry := range entries {
+		if !entry.IsDir() {
+			continue
+		}
+		entryPath := filepath.Join(rootDir, entry.Name())
+		if strings.HasSuffix(entryPath, subPath) {
```
rafaeldtinoco, line 236: In my system I'm getting empty subPath (lots of), making this to always be executed in such cases. So the logic works, but you should check why subPaths are empty to reduce execution overhead (and perhaps not run this logic if subPath is empty). 

Check it out:

```
subPath:
entryPath: /sys/fs/cgroup/system.slice/acpid.socket
subPath:
entryPath: /sys/fs/cgroup/system.slice/alsa-restore.service
subPath:
entryPath: /sys/fs/cgroup/system.slice/avahi-daemon.service
subPath:
entryPath: /sys/fs/cgroup/system.slice/avahi-daemon.socket
subPath:
entryPath: /sys/fs/cgroup/system.slice/binfmt-support.service
subPath:
entryPath: /sys/fs/cgroup/system.slice/blk-availability.service
subPath:
entryPath: /sys/fs/cgroup/system.slice/boot.mount
subPath:
entryPath: /sys/fs/cgroup/system.slice/console-setup.service
subPath:
entryPath: /sys/fs/cgroup/system.slice/containerd.service
subPath:
entryPath: /sys/fs/cgroup/system.slice/cpufrequtils.service
subPath:
entryPath: /sys/fs/cgroup/system.slice/cron.service
subPath:
entryPath: /sys/fs/cgroup/system.slice/dbus.service
subPath:
entryPath: /sys/fs/cgroup/system.slice/dbus.socket
subPath:
entryPath: /sys/fs/cgroup/system.slice/dm-event.socket
subPath:
entryPath: /sys/fs/cgroup/system.slice/docker-cd93f6d44215e34daa9688e7998beced9dae9424994bd2ca792c76c22288be6f.scope
subPath:
entryPath: /sys/fs/cgroup/system.slice/docker-cd93f6d44215e34daa9688e7998beced9dae9424994bd2ca792c76c22288be6f.scope
subPath: /system.slice/docker-cd93f6d44215e34daa9688e7998beced9dae9424994bd2ca792c76c22288be6f.scope
```

rafaeldtinoco:
I'm still checking the expiresAt logic with the new curTime approach. Will provide feedback soon...

rafaeldtinoco (commented):
I'm okay with this change as long as you fix the empty subPath issue and describe what is being done here in the commit log (which is not only adding ctime but also recursively trying to find the Path provided by the cgroup mkdir event). Perhaps describing why you are adding the ctime variable would also be good.

pkg/containers/containers.go
yanivagman: This function is also invoked from GetCgroupInfo as follows:
`getCgroupPath(c.cgroupMP, cgroupId, "")`
Which explains why you get the empty subPath - this is expected.

> `if subPath != "" && strings.HasSuffix(entryPath, subPath) {`

Will do

yanivagman: @rafaeldtinoco how often you see an empty subPath in your environment?
This shuold only happen when cgroup does not exist in the map, which shouldn't happen too often (Actually, almost never)

yanivagman: Ok, there was indeed an issue here.
cgroup_mkdir event wasn't processed in the case where container filter is set (I guess you did your checks with container filters set, right?)
The problem was that we filter for "false container poitives" and ignore the event in such cases. But cgroup_mkdir event is not a container event (but happens before the container starts), so we didn't get it.

I'll fix it in this PR, good that you found it

tracee-ebpf/tracee/events_pipeline.go
```diff
@@ -134,7 +124,7 @@ func (t *Tracee) decodeEvents(done <-chan struct{}) (<-chan *external.Event, <-c
 				PIDNS:               int(ctx.PidID),
 				ProcessName:         string(bytes.TrimRight(ctx.Comm[:], "\x00")),
 				HostName:            string(bytes.TrimRight(ctx.UtsName[:], "\x00")),
-				ContainerID:         containerId,
+				ContainerID:         t.containers.GetCgroupInfo(ctx.CgroupID).ContainerId,
```
rafaeldtinoco, line 127: for each received event, you're now calling GetCgroupInfo() and, even if cgroupInfo exists (most of the times), it still requires a RO mutex lock (as fyio, I'm not sure about performance hit here, when considering tons of events).

pkg/containers/containers.go
```diff
@@ -242,6 +202,52 @@ func (c *Containers) CgroupRemove(cgroupId uint64) {
...
+			return c.CgroupUpdate(cgroupId, path, time.Unix(stat.Ctim.Sec, stat.Ctim.Nsec))
+		}
+	}
+
+	// if no entry found, container may have already exited
+	// update map with existing data (ctime becomes an estimation only)
+	return c.CgroupUpdate(cgroupId, subPath, curTime)
+}
+
+func getCgroupPath(rootDir string, cgroupId uint64, subPath string) (string, error) {
```
rafaeldtinoco, line 222: This function is ingenious but I think it should be split into the 2 cases it represents. It looks overcomplicated the way it is.

1. **If subPath string is empty**, then it recurses the cgroup root dir (and subPath should be "" in the recursive call, because subPath will always be related to the cgroup fs root directory). Example:

```
subPath:
entryPath: /sys/fs/cgroup/system.slice/console-setup.service
subPath:
entryPath: /sys/fs/cgroup/system.slice/containerd.service
subPath:
entryPath: /sys/fs/cgroup/system.slice/cpufrequtils.service
subPath:
entryPath: /sys/fs/cgroup/system.slice/cron.service
subPath:
entryPath: /sys/fs/cgroup/system.slice/dbus.service
subPath:
entryPath: /sys/fs/cgroup/system.slice/dbus.socket
subPath:
entryPath: /sys/fs/cgroup/system.slice/dm-event.socket
subPath:
entryPath: /sys/fs/cgroup/system.slice/docker-4bed3dee01485506ad82b5c2d9d66d9daf58f94e8d982e7b6f080457a09a1dec.scope
subPath:
```

2. **If subPath isn't empty**, then there is no recursion and subPath is related to the cgroup fs root dir:

```
cgroupId = 33264, path = /sys/fs/cgroup/system.slice/docker-4bed3dee01485506ad82b5c2d9d66d9daf58f94e8d982e7b6f080457a09a1dec.scope, ContainerId = 4bed3dee01485506ad82b5c2d9d66d9daf58f94e8d982e7b6f080457a09a1dec, Runtime = docker

entryPath: /sys/fs/cgroup/system.slice/docker-4bed3dee01485506ad82b5c2d9d66d9daf58f94e8d982e7b6f080457a09a1dec.scope
subPath: /system.slice/docker-4bed3dee01485506ad82b5c2d9d66d9daf58f94e8d982e7b6f080457a09a1dec.scope

PATH: /sys/fs/cgroup/system.slice/docker-4bed3dee01485506ad82b5c2d9d66d9daf58f94e8d982e7b6f080457a09a1dec.scope
```

rafaeldtinoco (approved):
I'm approving this PR but I would like you to consider 2 things:

1. including the following commit I've prepared while I was reading and understanding the logic: it/79e3c413cf54a6910d84cbd03f3af5b3e1ff78a1 . It contains changes that might help a first time reader (as this logic isn't very straightforward).

2. consider my comment about getCgroupPath() function. I have tested, and it seems to be working fine. But, let me know what you think about my concern there.

pkg/containers/containers.go
yanivagman: Why do you say that there will be no recursion if subPath is not empty? There will be a recursion in both cases. The point is to find a path in cgroupfs that has subPath as a suffix. If subPath is empty, we will just look for cgroupid (walking cgroupfs recursively), and if subPath is not empty, we will still recurse, but will save stating to irrelevant paths that don't have subPath as a suffix

yanivagman:
> I'm approving this PR but I would like you to consider 2 things:
> 
> 1. including the following commit I've prepared while I was reading and understanding the logic: [rafaeldtinoco@79e3c41](https://github.com/rafaeldtinoco/tracee/commit/79e3c413cf54a6910d84cbd03f3af5b3e1ff78a1) . It contains changes that might help a first time reader (as this logic isn't very straightforward).
> 2. consider my comment about getCgroupPath() function. I have tested, and it seems to be working fine. But, let me know what you think about my concern there.

@rafaeldtinoco I left a few comments for you commit. Do you prefer to open a new PR for it?

rafaeldtinoco:
> @rafaeldtinoco I left a few comments for you commit. Do you prefer to open a new PR for it?

Sure, feel free to merge this if you'd like and I'll create a PR and ask your review. Thanks.

PR #1319, opened 2022-01-03
tracee-ebpf: fix non-systemd docker runtime

yanivagman:


pkg/containers/containers.go
```diff
@@ -169,58 +169,59 @@ func (c *Containers) cgroupLookupUpdate(cgroupId uint64) error {
...
+	for _, pc := range strings.Split(cgroupPath, "/") {
+		if len(pc) < 64 {
+			prevPathComp = pc
+			continue
+		}
+
+		runtime := "unknown"
+		id := strings.TrimSuffix(pc, ".scope")
+
+		if strings.HasPrefix(id, "docker-") {
```
mtcherni95, line 192: Maybe here we could `switch` on `strings.HasPrefix` without having to `goto check` at every case.
In effect, the break statement that is needed at the end of each case is provided automatically in Go. WDYT?

pkg/containers/containers.go
```diff
@@ -169,58 +169,54 @@ func (c *Containers) cgroupLookupUpdate(cgroupId uint64) error {
...
+		case strings.HasPrefix(id, "cri-containerd-"):
+			runtime = "containerd"
+			id = strings.TrimPrefix(id, "cri-containerd-")
+		case strings.HasPrefix(id, "libpod-"):
+			runtime = "podman"
+			id = strings.TrimPrefix(id, "libpod-")
+		}
+
+		if matched, _ := regexp.MatchString(`^[A-Fa-f0-9]{64}$`, id); matched {
+			if runtime == "unknown" && prevPathComp == "docker" {
```
mtcherni95, line 208: can we have a case with non systemd containerd or podman? meaning for example a path like `				/containerd/01adbaf26db7fac62f8b7ec5b116b2e60f4abbf230e29122e667996c6b5c04d4/`?
in that case, wouldn't we still have runtime `unknown`?

yanivagman: I am not aware of the existance of such cgroups paths for containerd/podman.
Are you?
Anyways, we can easilly add those in the future if it will be required

rafaeldtinoco (approved):
LGTM

Issue #1248, opened 2021-12-14
split flags parsing out of main.go


itaysk:
from https://github.com/aquasecurity/tracee/issues/960#issuecomment-992654757

> flags handling - stay in package main for now. but split in separate files: flags-filter.go, flags-output.go, flags-capture.go

yanivagman:
Done #1320

Issue #467, opened 2021-01-27
configuration from file


itaysk:
allow configuring tracee-rules and tracee-ebpf from a config file. this file should be given to the executable with a flag `--config` which will be parsed to fill the available flags options of each tool. this could be split in 2 issues, but most likely get implemented by the same person in one go.

itaysk:
notes from offline discussion with @mtcherni95 :

Our CLI flags are non-conventional. Usually CLI flags are key-value pairs, but in our case it's key-key-value. For example:
`--output format:json`, we have key1 which is `output`, key2 which is `format`, and value which is `json`. Ideally we would like this hierarchy to be natively represented in the configuration file, e.g:

```json
{
    "tracee-ebpf": {
        "output": {
            "format": "json"
        }
    }
}
```

BUT as an intermediate step, we are ok if the configuration file will be flat key-value, where the value contains key-value as string, e.g:

```json
{
    "tracee-ebpf": {
        "output": "format: json"
    }
}
```

(I use json to exemplify the point but I think we need to support yaml as well)

In the future we might want to change the CLI UX to be more conventional (simple k/v) and/or to generalize the inner key-value parsing code

itaysk:
related to #1288 and #1248

itaysk:
Hello :)
IMO we don't need more "prepare" or parsing functions, if anything we want to remove the ones we have. we should have just one way of parsing flags/config. If you want to use your (great) config format, that fine but I think you'll need to also change the flag parsing. If you don't want to touch the UX now, then change the file format to meet the current convention. This is essentially what's written here as discussed with @mtcherni95  https://github.com/aquasecurity/tracee/issues/467#issuecomment-1003112416

itaysk:
> I think that if we use the same function for both it will introduce coupling between the input (and thus UX) of the config file and of the cli

yes, and to me that's a good thing

> The CLI and the config files will practically need to have the same values which will create a conflict between creating a good UX for the cli and a good UX for the config file, possibly resulting in a bad/mediocore UX for both

This is what virtually every other CLI tool or framework does (source the same configuration from flags or files). I don't think anyone thinks it's mediocre.

> If some bug pops up in regards to parsing the config file, it might screw up the same parse in the CLI flags and vice versa.

Yes. but what about the glass half full :) now you have just one place to fix bugs

> adding a separate function ... won't take so much development time atm too

[...]

Issue #1288, opened 2021-12-20
Add "tracee" command


yanivagman:
The current entrypoint of the tracee container is entrypoint.sh.
This is a simple bash script that starts both tracee-ebpf and tracee-rules, and connects them with a pipe.
Although simple, this approach has several drawbacks:
1. It doesn't support passing capture flags (as well as other tracee-ebpf specific flags)
2. It is written in bash and not in go, making it less testable and impossible to use existing go packages (but binaries)
3. It is required to signal readiness of tracee-ebpf before starting tracee-rules (#930)

Now that we restructured the repo, we can start talking about adding a new (go) command that will be the entrypoint, called "tracee".
As a first step, this command can mimic entrypoint.sh functionality by spawning two new processes, and executing tracee-ebpf, then tracee-rules. Later, we can combine the flags of tracee-ebpf and tracee-rules, creating a more user friendly interface for interacting with tracee, while still giving the flexibility of using any flag of these two.

rafaeldtinoco:
I'm preparing a "mid-term" (that won't solve this issue) change allowing one to execute "Makefile.one" based containers for "tracee-ebpf" and "tracee" (tracee-ebpf + tracee-rules) using shell entrypoints. I believe this effort will be a continuation of changes made by https://github.com/aquasecurity/tracee/pull/1273.

itaysk:
This is a good opportunity to aggregate top-level management functionality such as configuration, and observability.
I was thinking to have some kind of "tracee-manager" to hold the HTTP server needed to expose stats, or receive commands, but actually this "tracee" executable is a better place to implement this.
related: #636 #887

josedonizetti:
@yanivagman @rafaeldtinoco This is still something on the plans? Having a single binary would better for the kubernetes deployment. Because to configure a liveness probe for kubernetes (so k8s knows tracee is healthy) checking two different process is a little odd. I was actually considering changing k8s kubernetes to instead of having one container, have two containers one for tracee-ebpf and another for tracee-rules and instead of a pipe, configure the communication to be unix sockets. So we would have one livenessprobe per container. But if moving into single binary is an option, I think i prefer a single binary.

yanivagman:
> @yanivagman @rafaeldtinoco This is still something on the plans?

Definitely.

rafaeldtinoco:
And, for now, we're also thinking on: https://github.com/aquasecurity/tracee/issues/2038 to make end user (docker container) life easier.

[...]

PR #1320, opened 2022-01-05
tracee-ebpf: seperate flag parsing function from main.go

NDStrahilevitz:
This is in preparation for #467.

The pull request simply moves the flag parsing functions for `trace`, `output`, and `capture` in tracee-ebpf to their own files, based on a similar PR by @mtcherni95.

solves #1248.


rafaeldtinoco:
@NDStrahilevitz I believe @yanivagman usually uses the tracee-ebpf flag whenever we do changes to `tracee.bpf.c` (and related) logic. I'll let him to confirm.

NDStrahilevitz:
> @NDStrahilevitz I believe @yanivagman usually uses the tracee-ebpf flag whenever we do changes to `tracee.bpf.c` (and related) logic. I'll let him to confirm.

I'll add him as a reviewer then as well. Which flag specifically did you mean? 
And just noting that this isn't changing any logic, just moves it, is the code in `tracee.bpf.c` dependent on the location of these functions somehow?

rafaeldtinoco:
> I'll add him as a reviewer then as well. Which flag specifically did you mean?

The PR label "tracee-ebpf" (I said flag before, sorry)

rafaeldtinoco (approved):
LGTM. Makes main look much cleaner.

yanivagman (commented):
Thanks @NDStrahilevitz , we definitely need some cleanups in main.
I wonder if we should put these files into a new internal package (https://dave.cheney.net/2019/10/06/use-internal-packages-to-reduce-your-public-api-surface).
This can help us in the future with #1288. WDYT?

NDStrahilevitz:
> Thanks @NDStrahilevitz , we definitely need some cleanups in main. I wonder if we should put these files into a new internal package (https://dave.cheney.net/2019/10/06/use-internal-packages-to-reduce-your-public-api-surface). This can help us in the future with #1288. WDYT?

I assume you mean as such `/cmd/tracee-ebpf/internal/flags...`? That sounds reasonable to me, i'll add that commit on your confirmation.

yanivagman:
> I assume you mean as such /cmd/tracee-ebpf/internal/flags...? That sounds reasonable to me, i'll add that commit on your confirmation.

Yes, this path is ok. This internal package can be called flags as the path suggests.

NDStrahilevitz:
> Yes, this path is ok. This internal package can be called flags as the path suggests.

Done.

Then I think this is ready to merge, right?

Issue #1310, opened 2021-12-27
Add higher level events to tracee-ebpf


AsafEitani:
For a user using tracee-ebpf as a monitoring or threat hunting tool to be able to understand the outputted events there is a need for a deep understanding of linux internals and syscalls.
Because the events are a reflection of the actual syscalls and lsm hooks it's really hard for a user to understand the meaning behind them.

Adding a higher level events which will symbolize common actions performed on the system will enable monitoring, threat hunting and writing tracee-rules rules much easier.
For example:

- Process executed - will be invoked when `sched_process_exec` is invoked.
- File opened
- File deleted
- Network connection invoked
- Network connection received
- etc

Those events can be part of a new set which will contain the events that are human (not linux experts) readable.

yanivagman:
#1105

yanivagman:
Pushing to v0.12.0 where we should at least merge #2186

Issue #975, opened 2021-08-30
Don't handle signals in tracee package, but in main


yanivagman:
Today we handle the SIGINT and SIGTERM signals in the tracee package (of tracee-ebpf).
A user that wants to use tracee-ebpf as a library may want to set its own handlers, and also be able to gracefully close the perf buffers and save the captured artifacts as done in the Run() function (and not in the Close() function).
A better approach to signal tracee to close can be done by using go context, which will then enable the caller (either main or a user that is using tracee-ebpf as a library) to call the context's cancelation function, and close tracee-ebpf gracefully.

PR #1323, opened 2022-01-06
tracee-ebpf: the return of the pipeline

yanivagman:
This PR separates events handling into two stages:
1. Events decoding
2. Events processing

This will allow us to introduce another stage to sort the events (#1103) as well as to add support for higher level events in the future (#1310)

NOTE: to avoid unnecessary copies, this PR breaks the current API of tracee by changing the events channel to use a **pointer** to external.Event.


itaysk:
what makes you want to re-introduce the pipeline now after you previously wanted to remove it? adding sorting and aggregating code would be possible also in a simple function call no?

tracee-ebpf/tracee/events_pipeline.go
```diff
@@ -39,25 +40,39 @@ type context struct {
 	_        [3]byte //padding
 }
 
-func (t *Tracee) processEvents(done <-chan struct{}) error {
+// handleEvents is a high-level function that starts all operations related to events processing
+func (t *Tracee) handleEvents(done <-chan struct{}) {
+	ctx, cancelFunc := context.WithCancel(context.Background())
```
itaysk, line 45: why another cancellation signal in addition to `done`?

tracee-ebpf/tracee/events_pipeline.go
```diff
@@ -84,69 +99,79 @@ func (t *Tracee) processEvents(done <-chan struct{}) error {
...
 			}
 		}
 
 		select {
-		case <-done:
-			return nil
-		case t.config.ChanEvents <- evt:
+		case <-ctx.Done():
+			return
+		case t.config.ChanEvents <- event:
```
itaysk, line 171: why can't we just write `*event` here and preserve the external interface?

tracee-ebpf/tracee/events_pipeline.go
```diff
@@ -39,25 +40,39 @@ type context struct {
...
 
-func (t *Tracee) processEvents(done <-chan struct{}) error {
+// handleEvents is a high-level function that starts all operations related to events processing
+func (t *Tracee) handleEvents(done <-chan struct{}) {
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	defer cancelFunc()
+
+	eventsChan := make(chan *external.Event)
+	go t.decodeEvents(ctx, eventsChan)
+	go t.processEvents(ctx, eventsChan)
```
itaysk, line 50: when you wanted to remove the pipeline you had a good case that it's too much for what we needed. Why do you want to bring it back now? more precisely, how goes calling the decode and process functions as goroutines help?

tracee-ebpf/tracee/events_pipeline.go
itaysk: I think this should happen in the pipeline, no in the pipeline step. it makes the pipeline less readable (you look at the handleevent and you don't know what's going to happen), and less testable

tracee-ebpf/tracee/events_pipeline.go
yanivagman: if we would have used done, we would have to send done signals as the number of pipeline stages (as once one of the stages consumes the event, another one won't get it).
Using context resolves this, as this is what it is intended to do - to signal cancellation.
In a future PR, I want to even move the context into main, and remove the done channel.

tracee-ebpf/tracee/events_pipeline.go
yanivagman: Using *event instead of event will do a copy of the event struct, which is not required.
By moving the pointer to the created event, we can avoid performing such copies, and work on a single instance of the event, which is preferrable.

yanivagman: > I think this should happen in the pipeline, no in the pipeline step. it makes the pipeline less readable (you look at the handleevent and you don't know what's going to happen), and less testable

This is the expected behavior of pipelines in go. A pipline has a source stage and a sink stage, as implemented here. I don't think we should implement it differently (https://go.dev/blog/pipelines)

tracee-ebpf/tracee/events_pipeline.go
yanivagman: When we removed the pipeline, it looked differently.
The code was much more complex (argTag and other stuff, which were removed since), and its original usage was not relevant anymore (as we added tracee-rules).
I think that the split introduced in this PR is much more readable, and makes a lot of sense.
The decode stage is responsible for creating the external.Event struct, which is the minimal struct to work with for any future work we want to do with it. 
For example, sorting events doesn't have to wait for their processing, and can reorder the events right after it has been decoded. In addition, such sorting will have to take place as another stage in the pipeline, as it needs to keep events until sync is achieved between the different cpu queues (#1103), so another channel is required, and we get back to pipeline anyways.

tracee-ebpf/tracee/events_pipeline.go
itaysk: > A pipline has a source stage and a sink stage, as implemented here

yes, a pipeline stage should have a source and a sink, but this code doesn't. What is the source and the sink here?

> I don't think we should implement it differently (https://go.dev/blog/pipelines)

I agree we should implement it like in that post but the code here is different

tracee-ebpf/tracee/events_pipeline.go
itaysk: > the split introduced in this PR is much more readable

I agree the split is good, but the question is why a pipeline. If you want to extract a function, that's always good, but if you want to make the function a go routine that's something else.

itaysk: > I think that the split introduced in this PR is much more readable, and makes a lot of sense.

I agree the split is good (the extracted functions). I just question the use of goroutines. 
If the motivation is supposed performance improvement thanks for parallelism - that's might be counter-productive, as @danielpacak had discovered when investigating tracee-rules pipeline. 
If the motivation is related to event sorting - ok, but I'm not clear how this pipelining this part helps

tracee-ebpf/tracee/events_pipeline.go
itaysk: the channel is not intended to be written to. It is intended to be `close`d when finished (this is a common go pattern). https://github.com/aquasecurity/tracee/blob/main/tracee-ebpf/tracee/tracee.go#L979
close is broadcasted to all consumers

itaysk: It's even discussed in the blog post you referred to:

> This approach has a problem: each downstream receiver needs to know the number of potentially blocked upstream senders and arrange to signal those senders on early return. Keeping track of these counts is tedious and error-prone.
We need a way to tell an unknown and unbounded number of goroutines to stop sending their values downstream. In Go, we can do this by closing a channel, because a receive operation on a closed channel can always proceed immediately, yielding the element type’s zero value.
This means that main can unblock all the senders simply by closing the done channel. This close is effectively a broadcast signal to the senders. We extend each of our pipeline functions to accept done as a parameter and arrange for the close to happen via a defer statement, so that all return paths from main will signal the pipeline stages to exit.

tracee-ebpf/tracee/events_pipeline.go
AlonZivony: Itay has a point.
The processing part doesn't have to be in its own go routine.
The sorting will have to be parallel, but the processing before and after it could be all one go routine (one for before and one for after) and by that reduce the performance problems of channels and context switching and avoid potential synchronization bugs.
The only thing that is bothering me is that we want to read as fast as we can from the data channel so the perf-buffers won't get full and result event loss. So for this reason, we might want the events decoding to be separated from other logics by different go routines.
Also, the pipeline design pattern make it very easy to expand the stages and add new features to events processing. So there are advantages.

itaysk: > we want to read as fast as we can from the data channel so the perf-buffers won't get full and result event loss. So for this reason, we might want the events decoding to be separated from other logics by different go routines.

Agree. I think this was the motivation for the original pipeline. but since the connecting channels are unbuffered, I don't this it'll make any difference. the reading gorouting would just block until the downstream goroutines would proceed

itaysk: > the pipeline design pattern make it very easy to expand the stages and add new features to events processing. So there are advantages.

definitely. but for that to happen I think we need to address https://github.com/aquasecurity/tracee/pull/1323#discussion_r779386915

tracee-ebpf/tracee/events_pipeline.go
yanivagman: > > A pipline has a source stage and a sink stage, as implemented here
> 
> yes, a pipeline stage should have a source and a sink, but this code doesn't. What is the source and the sink here?

As we only have two stages, this should be clear - the source is the decode events stage and the sink is the process events stage. You can argue that these stages have input and output channels, but these are not part of the pipeline. The input channel for decode events is for getting the events from the bpf code and the output channel is what has been configured for the events channel by the consumer of tracee-ebpf.

> 
> > I don't think we should implement it differently (https://go.dev/blog/pipelines)
> 
> I agree we should implement it like in that post but the code here is different

It is different by two things that I changed with the following reasons:
1. We don't return `outc`, `errc` and `err`: As we already have `handleError()` function which sends errors through a dedicated error channel, I didn't see a reason to add an extra error channel per each stage. Also, we never return an error which terminates the for loop, so there is no reason for having `err`. As we don't need errc and err, I decided to create the output channel outside the pipeline stages, which brings me to the second difference.
2. We don't call the `go` keyword from within the pipeline stage, but can call it from the outside, like I did here (which also avoids an extra indentation, and makes the PR easier to review).

tracee-ebpf/tracee/events_pipeline.go
yanivagman: > I agree the split is good (the extracted functions). I just question the use of goroutines.
> If the motivation is supposed performance improvement thanks for parallelism - that's might be counter-productive, as @danielpacak had discovered when investigating tracee-rules pipeline.
> If the motivation is related to event sorting - ok, but I'm not clear how this pipelining this part helps

I think you should have a look at #1103 and then you will have a better understanding for why this is required (and maybe you will have ideas for eliminating the need for the pipeline, who knows...).
About the performance part - I don't know what @danielpacak found in tracee-rules, but measuring the performance gain with a pipeline might be tricky. The time that a single event spends in the pipeline is longer than without the pipeline as the amount of computation is more or less the same, but an extra channel is added, which adds to the time the event stays in the pipeline. Yet, the throughput should be better, as things happen in parrallel, so more events can be processed in the same time window. Another thing to take into account, which might also be counter intuitive, is that the cpu usage might look higher with the pipeline when there is a high events load. The reason for this is that now two (or more) go routines are occupied with processing the events (instead of one without the pipeline), and if events keep coming in a high frequency, cpu load will be higher. Yet, the throughput should be higher, which is a good thing.

tracee-ebpf/tracee/events_pipeline.go
yanivagman: Fair enough. I'll remove the context part from this PR and will add it in a future PR when it will be also part of main (also related to #975)

tracee-ebpf/tracee/events_pipeline.go
AlonZivony: > We don't return outc, errc and err: As we already have handleError() function which sends errors through a dedicated error channel, I didn't see a reason to add an extra error channel per each stage.

In this matter I think the `handleError()` method is a problem, because it bounds you to create all go routines to be methods of `tracee`.
In my opinion, we should add an error channel which leads to a go routine that prints the errors and update the metrics. This way the new stages of the pipeline could be in different packages.

yanivagman: > > We don't return outc, errc and err: As we already have handleError() function which sends errors through a dedicated error channel, I didn't see a reason to add an extra error channel per each stage.
> 
> In this matter I think the `handleError()` method is a problem, because it bounds you to create all go routines to be methods of `tracee`. In my opinion, we should add an error channel which leads to a go routine that prints the errors and updates the metrics (in other words calling to `handleError` itself). This way the new stages of the pipeline could be in different packages.

Yes, you are right. We should handle errors in a different way so we can decouple the different stages. Currently, handleError is also called from net_events.go and write_capture.go, and I'm not going to handle this decoupling in this PR. I will update my PR though to include error channel and merging errors, like we had in the past.

tracee-ebpf/tracee/events_pipeline.go
danielpacak: IIRC, at the time when we looked into tracee-rules CPU utilisation the requirement was quite opposite. Our goal was to lower CPU from spikes even if we sacrifice throughput.

yanivagman: But is this what we really want?
Doing things in parallel will always mean having a greater cpu utilization for getting higher throughput. So if optimizing for no cpu spikes, the reason says that you need to kill all the parallelism. This makes no sense to me.

rafaeldtinoco: Isn't this a need for "tuning settings" feature ? 

Can't we implement tuning (though config options) to the amount of parallelism we would like to have ? 

Depth of channel buffering, amount of go routines we would like to run in parallel... We could even set the time window size we would like the events buffer to sort events.

Of course we can have a good "generic" number... But tuning things like that will permit us to maximize performance in real big machines, for example, where we can dedicate "tracee" to specific CPUs and considering amount of threads running.

This type of technique is used to maximize throughput in big cloud servers using DPDK and QEMU, for example. I'm just thinking that, with tunables, we would be able to pin tracee to specific CPUs dedicated to security in a big cloud server (for example), and other stuff (we could even recommend final users to use NOHZ, or disable RCU callbacks, so there are no uneeded context switches if we want a huge performance gain).

Just a thought that could help tracee being deployed and customized in big environments.

danielpacak: @yanivagman  TBH at that time the requirement to use as less CPU as possible did not make sense to me as well.

My conclusion was similar to what @rafaeldtinoco mentioned, i.e. that maybe we could control the number of goroutines used by tracee based on user config or number of available CPUs (e.g. respecting GOMAXPROCS env).

yanivagman: Throughput and cpu utilization are two different metrics to measure performance.
Deciding on the right metric to measure performance depends on the application and business needs.
For a server application, the throughtput might be the more important metric to measure.
However, for tracee we should probably balance between those two.
Having a higher throughput will enable tracee to handle bigger loads and peaks without loosing events (as the buffers have a limited amount of space) and respond quickly.
On the other hand, tracee is not the main business logic in a node, because it is used to monitor other applications in the system. This means that it competes with other applications on cpu resources, and cpu utilization then becomes important.
We can improve both of these metrics if we remove or optimize logic in hotspots, but having paralellism will benefit the throughput metric, while degrading cpu utilization.
Having the ability to balance these two metric with some kind of tuning is definitely a desired feature that we should consider in the future. For this PR, I think it is not a bad thing to have some parallelism and gain some throughput, even if that might affect cpu utilization, which can probably be solved with other measures, such as limiting the cpu resources with cgroups.

rafaeldtinoco (approved):
Are you still working on this ? I have +1'ed the change because its just an argument type change (pointer name) and file renaming basically.

yanivagman:
> Are you still working on this ? I have +1'ed the change because its just an argument type change (pointer name) and file renaming basically.

Thanks @rafaeldtinoco. The PR is ready. Other than what you mentioned, this PR also adds two go routines for the pipeline.
I'll wait to see if there are any more comments from @itaysk  @danielpacak @AlonZivony and merge it in Sunday

tracee-ebpf/tracee/events_pipeline.go
itaysk: > It is different by two things that I changed with the following reasons: We don't return outc, errc and err

That's what I'm trying to say. this implementation is not the pipeline that's described in the blog post. it's big difference since by not returning out and not receiving in you are coupling the stages (also with shared state)

Bottom line is that a proper pipeline should control the flow of execution, and the pipeline stages should not be involved. I don't think it's a huge deal but I think it would be changed to a proper pipeline at some point probably (it it would remain a pipeline)

tracee-ebpf/tracee/events_pipeline.go
itaysk: I have read the discussion and I correct me if I'm wrong but the question is still unanswered - what do we gain from calling those functions as goroutines?
as long as the channels are unbuffered we don't gain throughput or performance.
Also, just to address some of the questions, we should assume Tracee (not tracee-ebpf) is allocated 1 core

tracee-ebpf/tracee/events_pipeline.go
yanivagman: > > It is different by two things that I changed with the following reasons: We don't return outc, errc and err
> 
> That's what I'm trying to say. this implementation is not the pipeline that's described in the blog post. it's big difference since by not returning out and not receiving in you are coupling the stages (also with shared state)
> 
> Bottom line is that a proper pipeline should control the flow of execution, and the pipeline stages should not be involved. I don't think it's a huge deal but I think it would be changed to a proper pipeline at some point probably (it it would remain a pipeline)

Technically speaking, there is no real difference between what I implemented and what described in this blog. But let's not continue this discussion - have you seen the recent changes? I don't think so, because now it is implemented in the exact same way.

yanivagman: > > > why can't we just write *event here and preserve the external interface?
> 
> > Using *event instead of event will do a copy of the event struct, which is not required.
> > By moving the pointer to the created event, we can avoid performing such copies, and work on a single instance of the event, which is preferrable.
> 
> I don't think it matters if we do another copy or not (happy to be proven wrong here). As long as it's internal code then I wouldn't care if you wanted to share pointers but when it comes out of tracee, I think the API should remain a channel of struct and not a channel of pointer to struct.

Why?

tracee-ebpf/tracee/events_pipeline.go
yanivagman: > I have read the discussion and I correct me if I'm wrong but the question is still unanswered - what do we gain from calling those functions as goroutines? as long as the channels are unbuffered we don't gain throughput or performance. Also, just to address some of the questions, we should assume Tracee (not tracee-ebpf) is allocated 1 core

I don't think it's correct to say that with unbuffered channels we won't get performance gain. You have two go routines that work in parallel - then you get more throughput. Using buffered channel in a pipeline won't help, as the throughput will still be determined by the "weakest link" in the pipeline.
Other than that and like I wrote before - anyways we are going to need a pipeline for the sorting feature. I don't think there is a way around that.

PR #1324, opened 2022-01-06
Tracee config files

NDStrahilevitz:
Solves #467 

This PR adds the ability to read configuration files into tracee-rules and tracee-ebpf which mimic the cli flags.

The configuration file location can be read from an environment variable named `TRACEE-{RULES/EBPF}-CONFIG` or from a default location `./tracee-{rules/ebpf}-config.json` in the same folder as the executable.

Currently only json is supported.

There is some code duplication to be solved in regards to the config loading function and some debate could be had about how to load the configuration file (due to the way the cli package works, loading the filename from a flag won't work well).

I've [attached](https://github.com/aquasecurity/tracee/files/7822868/tracee-configs.zip) example config files.

### DIFF
--- a/cmd/tracee-ebpf/internal/config/config.go
+++ b/cmd/tracee-ebpf/internal/config/config.go
@@ -19,7 +19,7 @@ type TraceeEbpfCliConfig struct {
 	Capture      []string
 }
 
-func LoadConfig(fileName string) (config *TraceeEbpfCliConfig, err error) {
+func Load(fileName string) (config *TraceeEbpfCliConfig, err error) {
 	cfgFile, err := os.Open(fileName)
 
 	if err != nil {
--- a/cmd/tracee-ebpf/internal/config/config.go
+++ b/cmd/tracee-ebpf/internal/config/config.go
@@ -6,7 +6,7 @@ import (
 	"os"
 )
 
-type TraceeEbpfCliConfig struct {
+type CliConfig struct {
 	PerfRingBuffers struct {
 		EventSubmitBufferSize int
 		BlobBufferSize        int
@@ -19,7 +19,7 @@ type TraceeEbpfCliConfig struct {
 	Capture      []string
 }
 
-func Load(fileName string) (config *TraceeEbpfCliConfig, err error) {
+func Load(fileName string) (config *CliConfig, err error) {
 	cfgFile, err := os.Open(fileName)
 
 	if err != nil {
@@ -47,21 +47,21 @@ func DefaultConfigLocation() string {
 	return "./tracee-ebpf-config.json"
 }
 
-func (c *TraceeEbpfCliConfig) DefaultInstallPath() string {
+func (c *CliConfig) DefaultInstallPath() string {
 	if c.InstallPath != "" {
 		return c.InstallPath
 	}
 	return "/tmp/tracee"
 }
 
-func (c *TraceeEbpfCliConfig) DefaultBuildPolicy() string {
+func (c *CliConfig) DefaultBuildPolicy() string {
 	if c.BuildPolicy != "" {
 		return c.BuildPolicy
 	}
 	return "if-needed"
 }
 
-func (c *TraceeEbpfCliConfig) DefaultEventSubmitBufferSize() int {
+func (c *CliConfig) DefaultEventSubmitBufferSize() int {
 	bufferSize := c.PerfRingBuffers.EventSubmitBufferSize
 	if bufferSize != 0 {
 		return bufferSize
@@ -69,7 +69,7 @@ func (c *TraceeEbpfCliConfig) DefaultEventSubmitBufferSize() int {
 	return 1024
 }
 
-func (c *TraceeEbpfCliConfig) DefaultBlobBufferSize() int {
+func (c *CliConfig) DefaultBlobBufferSize() int {
 	bufferSize := c.PerfRingBuffers.BlobBufferSize
 	if bufferSize != 0 {
 		return bufferSize
--- a/cmd/tracee-ebpf/internal/config/config.go
+++ b/cmd/tracee-ebpf/internal/config/config.go
@@ -19,11 +19,11 @@ type CliConfig struct {
 	Capture      []string `json:"capture"`
 }
 
-func Load(fileName string) (config *CliConfig, err error) {
+func Load(fileName string) (config CliConfig, err error) {
 	cfgFile, err := os.Open(fileName)
 
 	if err != nil {
-		return nil, err
+		return CliConfig{}, err
 	}
 
 	defer cfgFile.Close()
@@ -31,7 +31,7 @@ func Load(fileName string) (config *CliConfig, err error) {
 	byteValue, err := ioutil.ReadAll(cfgFile)
 
 	if err != nil {
-		return nil, err
+		return CliConfig{}, err
 	}
 
 	err = json.Unmarshal(byteValue, &config)
--- a/cmd/tracee-ebpf/internal/config/config.go
+++ b/cmd/tracee-ebpf/internal/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"encoding/json"
+	"errors"
 	"io/ioutil"
 	"os"
 )
@@ -39,12 +40,12 @@ func Load(fileName string) (config CliConfig, err error) {
 	return config, err
 }
 
-func DefaultConfigLocation() string {
+func DefaultConfigLocation() (string, error) {
 	envLocation := os.Getenv("TRACEE_EBPF_CONFIG")
 	if envLocation != "" {
-		return envLocation
+		return envLocation, nil
 	}
-	return "./tracee-ebpf-config.json"
+	return "", errors.New("configuration location not found in TRACEE_EBPF_CONFIG")
 }
 
 func (c *CliConfig) DefaultInstallPath() string {
--- a/cmd/tracee-ebpf/internal/config/config.go
+++ b/cmd/tracee-ebpf/internal/config/config.go
@@ -1,0 +1,80 @@
+package config
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"os"
+)
+
+type TraceeEbpfCliConfig struct {
+	PerfRingBuffers struct {
+		EventSubmitBufferSize int
+		BlobBufferSize        int
+	}
+	VerboseDebug bool
+	InstallPath  string
+	BuildPolicy  string
+	Trace        []string
+	Output       []string
+	Capture      []string
+}
+
+var Config TraceeEbpfCliConfig
+
+func LoadConfig(fileName string) error {
+	cfgFile, err := os.Open(fileName)
+
+	if err != nil {
+		return err
+	}
+
+	defer cfgFile.Close()
+
+	byteValue, err := ioutil.ReadAll(cfgFile)
+
+	if err != nil {
+		return err
+	}
+
+	err = json.Unmarshal(byteValue, &Config)
+
+	return err
+}
+
+func DefaultConfigLocation() string {
+	envLocation := os.Getenv("TRACEE_EBPF_CONFIG")
+	if envLocation != "" {
+		return envLocation
+	}
+	return "./tracee-ebpf-config.json"
+}
+
+func (c *TraceeEbpfCliConfig) DefaultInstallPath() string {
+	if c.InstallPath != "" {
+		return c.InstallPath
+	}
+	return "/tmp/tracee"
+}
+
+func (c *TraceeEbpfCliConfig) DefaultBuildPolicy() string {
+	if c.BuildPolicy != "" {
+		return c.BuildPolicy
+	}
+	return "if-needed"
+}
+
+func (c *TraceeEbpfCliConfig) DefaultEventSubmitBufferSize() int {
+	bufferSize := c.PerfRingBuffers.EventSubmitBufferSize
+	if bufferSize != 0 {
+		return bufferSize
+	}
+	return 1024
+}
+
+func (c *TraceeEbpfCliConfig) DefaultBlobBufferSize() int {
+	bufferSize := c.PerfRingBuffers.BlobBufferSize
+	if bufferSize != 0 {
+		return bufferSize
+	}
+	return 1024
+}
--- a/cmd/tracee-ebpf/internal/config/config.go
+++ b/cmd/tracee-ebpf/internal/config/config.go
@@ -19,13 +19,11 @@ type TraceeEbpfCliConfig struct {
 	Capture      []string
 }
 
-var Config TraceeEbpfCliConfig
-
-func LoadConfig(fileName string) error {
+func LoadConfig(fileName string) (config *TraceeEbpfCliConfig, err error) {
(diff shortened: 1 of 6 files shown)


cmd/tracee-ebpf/internal/config/config.go
```diff
@@ -0,0 +1,80 @@
...
+
+	return err
+}
+
+func DefaultConfigLocation() string {
+	envLocation := os.Getenv("TRACEE_EBPF_CONFIG")
+	if envLocation != "" {
+		return envLocation
+	}
+	return "./tracee-ebpf-config.json"
```
NDStrahilevitz, line 49: This doesn't work as intended: the `.` refers to the working dir of the executing user instead of the location of the executable itself

cmd/tracee-rules/internal/config/config.go
```diff
@@ -0,0 +1,89 @@
...
+
+	return err
+}
+
+func DefaultConfigLocation() string {
+	envLocation := os.Getenv("TRACEE_RULES_CONFIG")
+	if envLocation != "" {
+		return envLocation
+	}
+	return "./tracee-rules-config.json"
```
NDStrahilevitz, line 67: Same as tracee-ebpf, this doesn't work as intended: the `.` refers to the working dir of the executing user instead of the location of the executable itself

cmd/tracee-ebpf/main.go
```diff
@@ -30,6 +32,12 @@ var buildPolicy string
 var version string
 
 func main() {
+	err := config.LoadConfig(config.DefaultConfigLocation())
+
+	if err != nil {
+		log.Println("No configuration file found in environment variable or default location, flags will be set from cli.")
```
mtcherni95, line 38: let's use fmt.Printf as in the rest of the file

cmd/tracee-rules/internal/config/config.go
```diff
@@ -0,0 +1,89 @@
...
+	Pprof struct {
+		Enable  bool
+		Address string
+	}
+	List bool
+}
+
+var Config TraceeRulesCliConfig
+
+func LoadConfig(fileName string) error {
```
mtcherni95, line 42: I would remove to just `Config`

cmd/tracee-ebpf/internal/config/config.go
```diff
@@ -0,0 +1,80 @@
...
+
+	byteValue, err := ioutil.ReadAll(cfgFile)
+
+	if err != nil {
+		return err
+	}
+
+	err = json.Unmarshal(byteValue, &Config)
+
+	return err
```
mtcherni95, line 41: `return json.Unmarshal(byteValue, &Config)`

cmd/tracee-ebpf/internal/config/config.go
```diff
@@ -0,0 +1,80 @@
+package config
+
```
mtcherni95, line 2: I would rename the `internal` dir to `input`

cmd/tracee-ebpf/internal/config/config.go
```diff
@@ -0,0 +1,80 @@
...
+	InstallPath  string
+	BuildPolicy  string
+	Trace        []string
+	Output       []string
+	Capture      []string
+}
+
+var Config TraceeEbpfCliConfig
+
+func LoadConfig(fileName string) error {
```
mtcherni95, line 24: Why not return a `TraceeEbpfCliConfig` instance instead of defining globally?

cmd/tracee-ebpf/internal/config/config.go
```diff
@@ -0,0 +1,80 @@
+package config
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"os"
+)
+
+type TraceeEbpfCliConfig struct {
```
mtcherni95, line 9: did we test it? I am not sure the unmarshal at line 39 will work if the struct is not provided with json labels.

cmd/tracee-rules/internal/config/config.go
```diff
@@ -0,0 +1,89 @@
+package config
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+)
+
+type TraceeRulesCliConfig struct {
```
mtcherni95, line 10: also here, don't we need json labels?

cmd/tracee-rules/internal/config/config.go
```diff
@@ -0,0 +1,89 @@
...
+		Template string
+	}
+	Pprof struct {
+		Enable  bool
+		Address string
+	}
+	List bool
+}
+
+var Config TraceeRulesCliConfig
```
mtcherni95, line 40: same as above, let's not use a global var, instead LoadConfig should return the configuration

cmd/tracee-ebpf/internal/config/config.go
itaysk, line 49: this won't fly (relative path). I don't think we need it, let's just stick with a cli flag, maybe env var

cmd/tracee-ebpf/internal/config/config.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = json.Unmarshal(byteValue, &Config)
+
+	return err
+}
+
+func DefaultConfigLocation() string {
```
itaysk, line 44: most importantly we need to be able to provide the config as a cli flag: `tracee-ebpf --config /path/to/file`

cmd/tracee-ebpf/internal/config/config.go
itaysk, line 9: I think `TraceeEbpf` is redundant here, since we are in `cmd/tracee-ebpf`

cmd/tracee-ebpf/main.go
```diff
@@ -209,54 +217,56 @@ func main() {
 			&cli.StringSliceFlag{
 				Name:    "trace",
 				Aliases: []string{"t"},
-				Value:   nil,
+				Value:   cli.NewStringSlice(config.Config.Trace...),
 				Usage:   "select events to trace by defining trace expressions. run '--trace help' for more info.",
 			},
```
itaysk, line 222: there's no point in having the CLI flags here and the config file there. If we're going with this approach than the Flag structs should move to the config package as well (created by the package and returned to be used here)

cmd/tracee-ebpf/main.go
```diff
@@ -209,54 +217,56 @@ func main() {
...
-				Value:       "if-needed",
+				Value:       config.Config.DefaultBuildPolicy(),
 				Usage:       "when to build the bpf program. possible options: 'never'/'always'/'if-needed'",
 				Destination: &buildPolicy,
 			},
 		},
 	}
 
-	err := app.Run(os.Args)
+	app.Before = altsrc.InitInputSourceWithContext(app.Flags, altsrc.NewJSONSourceFromFlagFunc("config"))
```
itaysk, line 267: I'm not entirely clear on why we need altsrc now if you're parsing the JSON manually

cmd/tracee-ebpf/internal/config/config.go
itaysk: I'm not sure how this relate to L276

```
app.Before = altsrc.InitInputSourceWithContext(app.Flags, altsrc.NewJSONSourceFromFlagFunc("config"))
```

itaysk (commented):
Thanks for working on this. I left some comments that need to be addressed, but honestly I'm still not sure this is the best approach though (but that's to be discussed in the issue)

itaysk:
WDYT about using https://github.com/spf13/viper  to manage and load the config file? the rest of the "glue" code you wrote can remain as is

cmd/tracee-ebpf/main.go
NDStrahilevitz: 1. Alright
2. I think all the error values here would be arbitrary strings (depends on the implementation of the std packages used to parse), and i'm not sure what we could do with them besides notifying the user differently. At first I logged a fatal here but I thought this shouldn't stop the user, would you say it should?

cmd/tracee-rules/internal/config/config.go
NDStrahilevitz: Agreed

cmd/tracee-ebpf/internal/config/config.go
NDStrahilevitz: I wanted one instance of the config so it can be watched for value changes later, but maybe that was jumping ahead of it and could be implemented better anyways.
There was the suggestion to use viper package for that, but for now i'll change it to a return value.

cmd/tracee-ebpf/internal/config/config.go
NDStrahilevitz: I tested it, it works, but it needs to correspond one to one (not case sensitive) with the json keys, so for the long term it would be a bit better with tags if we want to change the struct names (structure itself would still need to be mapped 1 to 1).

cmd/tracee-rules/internal/config/config.go
NDStrahilevitz: See above

cmd/tracee-rules/internal/config/config.go
NDStrahilevitz: Same as before, i'll change it to a return

cmd/tracee-ebpf/internal/config/config.go
NDStrahilevitz: Alright, i'll change it to return an error if not found instead.

cmd/tracee-ebpf/internal/config/config.go
NDStrahilevitz: That's fair, do you think we should even export the struct itself? So change it to `cliConfig`?

cmd/tracee-ebpf/internal/config/config.go
NDStrahilevitz: > most importantly we need to be able to provide the config as a cli flag: `tracee-ebpf --config /path/to/file`

I'll see if I can make it work with the `flag` package since from the `cli` it won't work because it parses after we need the location.

cmd/tracee-ebpf/main.go
NDStrahilevitz: @mtcherni95 suggested we don't use a global instance for the config, which for the current implementation I also think would be better. In that case we'd need to change the flags to be returned from a function which receives a config as a parameter (something like `GetCliFlags(config CliConfig) ([]cli.Flag)`. Either that or we keep them here and I pass the return value from the returned config as the default value instead.

cmd/tracee-ebpf/main.go
NDStrahilevitz: That's a leftover, i'll clean it up.

cmd/tracee-ebpf/internal/config/config.go
NDStrahilevitz: I think this should still stay in `internal` so it won't be able to be imported outside of `cmd/tracee-{}`. Would there be a reason we'd want it the other way?

NDStrahilevitz:
> WDYT about using https://github.com/spf13/viper to manage and load the config file? the rest of the "glue" code you wrote can remain as is

I'll do a check on how it looks, I think it reduce some boilerplate and be easier in the long run

cmd/tracee-rules/internal/config/config.go
NDStrahilevitz: Done

cmd/tracee-ebpf/internal/config/config.go
NDStrahilevitz: Now returns an instance so irrelevant

cmd/tracee-ebpf/internal/config/config.go
NDStrahilevitz: Done

cmd/tracee-ebpf/internal/config/config.go
NDStrahilevitz: Added json tags, done

cmd/tracee-ebpf/main.go
NDStrahilevitz: Done

cmd/tracee-ebpf/internal/config/config.go
NDStrahilevitz: Renamed to `CliConfig`

cmd/tracee-ebpf/internal/config/config.go
NDStrahilevitz: @itaysk 

After some testing with the `flag` package, it seems to me this feature can't really work well with the cli package used (`urfave/cli`):
1. The package doesn't seem to play well with another flag package (i've attempted to do so with the standard `flag` package), each will complain about the other flags not being defined
2. If we define the config flag using `urfave/cli` it will be parsed alongside the other flags, but the config flag must be loaded before the other flags so the config can be loaded and then allow the cli flags to override it. There isn't really a way to tell the difference between an undefined flag and an undefined config field so we must have the config as the flag's default value, but then we would have to load the `config` flag before to find the config location before the rest of the flags, which is isn't possible in this package.

I think we should stick to the environment variable, it's a simple and easy solution, and if running from a container wouldn't be that different from passing a `--config` flag (you'd probably pass the config in a volume regardless so only the way you declare the config file location changes when running the `docker` command)

cmd/tracee-ebpf/internal/config/config.go
NDStrahilevitz: irrelevant now

cmd/tracee-rules/internal/config/config.go
NDStrahilevitz: same as above

cmd/tracee-ebpf/main.go
NDStrahilevitz: logic changed a bit, new logs reflect that

cmd/tracee-ebpf/internal/config/config.go
NDStrahilevitz: Done

rafaeldtinoco:
Should we close this for now ? We're discussing how a centralized config file would look like for the major k8s architecture, I think any similar PR could use ideas from this one but this one wouldn't involve into those. @danielpacak ?

NDStrahilevitz:
> Should we close this for now ? We're discussing how a centralized config file would look like for the major k8s architecture, I think any similar PR could use ideas from this one but this one wouldn't involve into those. @danielpacak ?

I think this can be closed, but I'd also like to raise the point that cli users, or those who deploy tracee on non k8s envs, might need their own config file instead of using the k8s config structure. So a similar PR might still be needed in the future.

rafaeldtinoco:
Okay, I'm closing this for now only (and we can re-use, re-visit this later if needed). Yes @Nadav, also agree on the config file for cmdline users AND it is yet unclear if we're pursuing a single config file to do both: cmdline args AND signatures description OR we will have 2 files, or multiple. We are waiting the go-cel POC to be finished then we can decide. @danielpacak copied in here fyi.

PR #1327, opened 2022-01-08
Remove duplicate and sort the result of "tracee-rules --list-events"

vincent-pli:
There are duplicate items in the events list by `tracee-rules --list-events`, see the current result:
```
security_socket_connect,
dup,
dup2,
dup3,
close,
sched_process_exit,
execve,
security_socket_connect,
ptrace,
security_file_open,
ptrace,
security_file_open,
process_vm_writev,
security_sb_mount,
magic_write,
mem_prot_alert,
security_bprm_check,
security_bprm_check,
security_file_open,
init_module,
security_kernel_read_file,
security_file_open,
execve,
security_file_open  
```

To remove the duplicates and sort it will be more make sense.


cmd/tracee-rules/main.go
```diff
@@ -194,13 +195,20 @@ func listSigs(w io.Writer, sigs []types.Signature) error {
 }
 
 func listEvents(w io.Writer, sigs []types.Signature) {
-	var events []string
+	m := map[string]string{}
 	for _, sig := range sigs {
 		es, _ := sig.GetSelectedEvents()
 		for _, e := range es {
-			events = append(events, e.Name)
+			m[e.Name] = e.Name
```
mtcherni95, line 202: Hi, thank you for reporting! 
Why just not mapping between event name and empty struct? If element exists in map then continue.

vincent-pli: Done, thanks @mtcherni95

cmd/tracee-rules/main.go
```diff
@@ -194,13 +195,22 @@ func listSigs(w io.Writer, sigs []types.Signature) error {
...
+			}
 		}
 	}
+
+	var events []string
+	for k, _ := range m {
+		events = append(events, k)
+	}
+
+	sort.Slice(events, func(i, j int) bool { return events[i] < events[j] })
```
mtcherni95, line 213: I don't think this is needed any loner, as your map stores unique keys.

vincent-pli: Yes, just want to sort the keys to make the output more readable.

itaysk: yes, this is sorting not deduplicating @mtcherni95 . 
we should keep it not only for readability but also to make the output consistent. please keep it @vincent-pli

mtcherni95:
LGTM

mtcherni95 (approved):

PR #1328, opened 2022-01-09
Tracee-ebpf: Added exit of process group flag to sched_process_exit

AlonZivony:


tracee-ebpf/tracee/events_definitions.go
```diff
@@ -5722,6 +5722,7 @@ var EventsDefinitions = map[int32]EventDefinition{
 		Sets:           []string{"default", "proc", "proc_life"},
 		Params: []external.ArgMeta{
 			{Type: "long", Name: "exit_code"},
+			{Type: "int", Name: "is_process_group_exit"},
```
yanivagman, line 5725: Let's use Linux terminology, which is also shorter - group_dead

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2501,13 +2502,14 @@ int tracepoint__sched__sched_process_exit(struct bpf_raw_tracepoint_args *ctx)
...
-    if (proc_tree_filter_set) {
-        // Check number of threads in thread group to determine if this is last one
-        struct task_struct *task = (struct task_struct *)bpf_get_current_task();
-        struct task_struct *group_leader = READ_KERN(task->group_leader);
-        struct list_head thread_group_head = READ_KERN(group_leader->thread_group);
-        struct list_head *next = READ_KERN(thread_group_head.next);
-        if (GET_FIELD_ADDR(group_leader->thread_group) == next) {
+
+    int is_group_exit = 0;
+    struct task_struct *task = (struct task_struct *)bpf_get_current_task();
```
yanivagman, line 2507: You can use data.task instead

tracee-ebpf/tracee/events_definitions.go
yanivagman: Also, why not using bool?

AlonZivony: So I tried using bool but I got a wrong value.
I will try to look if I can do it.

tracee-ebpf/tracee/events_decoder.go
```diff
@@ -35,6 +35,7 @@ const (
 	credT
 	intArr2T
 	argsArrT
+	boolT
```
yanivagman, line 38: If not aligning this with bpf code, let's use other numbering for these types

tracee-ebpf/tracee/events_decoder.go
```diff
@@ -34,7 +34,13 @@ const (
 	u16T
 	credT
 	intArr2T
-	argsArrT
+)
+
+// This types are not matching the defined ones in the ebpf code because they are not types used by syscalls.
```
yanivagman, line 39:
```suggestion
// These types don't match the ones defined in the ebpf code since they are not being used by syscalls arguments.
```

yanivagman (approved):

PR #1330, opened 2022-01-10
Feature/process tree

### LINKED ISSUE #1262: Tracee-rules: Process tree
In tracee-rules there is a great need for process tree which will be available to signatures and their output.
A signature can use it to examine which process invoked the given event and to which processes it is connected (children and parents).
Also, using it will give much more information to the user upon a signature finding.

AlonZivony:
Created a dynamic process tree in tracee-rules to provide data to signatures and enrich signatures output.
Closes #1262 

### Intreduction
A need for access to the process tree from inside signatures has risen, along with a need of process tree view in the findings of signatures.
To answer this need, a process tree struct which is accessible to signatures by helpers is needed in tracee-rules.

### General Implementation details
The process tree is simply a map between a PID to the process with the given ID for simple and fast access.

Each process has a pointer to its parent process and children processes, which means that as long as they are referenced they won't be garbage-collected.

When a new process is observed, it will be added to the map. A reference to the parent will be added and a reference from the father to the child will be added.

When process exits, it will be removed from the map and from its father.
However, we want access to all of the living processes ancestors, which means that as long as a process has a living child, it won't be removed.

To create a unique tree for each container, the first process in each container will not have a father (could be easily recognized by PPID of 0 inside its namespace).

The process tree is a private global instance, which export some methods for the signature's helpers, for events processing and for findings enrichment.

### Problems
There are problems that makes the implementation of this feature to be much more complicated:
* Tracee-rules doesn't necessary run on the same machine as Tracee-ebpf, which means the only information regarding the process tree is given using events.
* Tracee-ebpf sends events in unordered manner. (A sorting feature is now in PR, but it is not mandatory)
* Some events which are expected could be lost between the kernel and the user-mode in Tracee-ebpf and never arrive to Tracee-rules
* There is a time difference between the processing of an event between the signatures and the process-tree, which might result unexpected status of the event at different stages.

#### Events Driven Information
Because all the information we can get is using the events received from Tracee-ebpf, this process tree is only a real-time tree with no information of all processes which are not followed using events.
However, there is not single event with all the information in need for the tree.

All events give general information about the process from the mandatory information in events.
This is a list of events with more information from other events and what information they give:
* `sched_process_fork`
    * process-fork - start time
    * thread-fork - thread count
* `sched_process_exec `- process name, execute binary information, thread count
* `sched_process_exit`
    * process-exit - process has died (remove if there are no child process)
    * thread-exit - reduce thread count

#### Events Status
To address the sorting and lost events issues, each process has a status field which is a bitmap to follow its information.
The statuses are as following:
* General Information - has information which is given from each possible event from a process
* Forked - has information from process-fork `sched_process_fork` event.
* Executed - has information from `sched_process_exec` event

Another status is created to be able in enrich the tree with information of processes with no event given on them - **Hollow Parent**.
This status means we only know the PID of the process (it is the PPID of another known process).
What this status give us is the ability to connect 2 process to the same parent even if we didn't receive any event from the parent process.

By addressing the statuses in each event handling, we can fill missing information about a process in unordered manner and there is no need to wait to an event which might have been lost (except for the process-exit event, which is crucial).

#### Delayed Delete of Processes
If a process is deleted upon `sched_process_exit`, it might be needed afterwards by a signature which sign on the exit event or because another events from the same process arrive after the exit event.
To make the process accessible even after exit event for a time buffer, each process is entered to a cache and only when a certain amount  of other processes are exited it will be deleted from the tree.
This is not a bullet proof solution, but a best-effort low-cost solution.


AlonZivony:
This PR is blocked by #1328

AlonZivony:
I think this PR is probably too big as is, I think maybe it should be divided to multiple PRs.

signatures/helpers/helpers.go
```diff
@@ -26,3 +28,11 @@ func IsFileWrite(flags string) bool {
 	}
 	return false
 }
+
```
AlonZivony, line 31: The functions here should be accessible from GOB signatures as well.

tracee-rules/types/process.go
```diff
@@ -0,0 +1,38 @@
+package types
+
+type ProcessInformationStatus uint32
+
+const (
```
AlonZivony, line 5: This values are not exported so they should not be under /types directory

rafaeldtinoco:
I'm closing this as this needs discussion, rework, rebase, etc etc.

yanivagman:
As far as I understand, the code in this PR was almost ready, but it was just too big, and the commits were not organized.
I think there is some good functionality implemented here, and I do want to see it merged in the future.

rafaeldtinoco:
I agree @yanivagman , dont want it to die, just wanted it better organized in other PRs etc.

Issue #295, opened 2020-11-10
Warn on kernel lockdown


yanivagman:
Kernel lockdown feature was added to the kenel since version 5.4:
https://www.zdnet.com/article/linux-to-get-kernel-lockdown-feature/
This is a new LSM, that effectively disables some BPF functionality as can be seet at:
https://lore.kernel.org/patchwork/patch/1055067/

More details:
https://lwn.net/Articles/791863/
https://github.com/iovisor/bcc/issues/2565

Detecting lockdown can be done using /sys/kernel/security/lockdown sysfs entry (Starting from kernel 5.4)
Lockdown can be set to integrity or confidentiality.

Example output:
$ sudo cat /sys/kernel/security/lockdown
[none] integrity confidentiality

When secureboot (and lockdown) is enabled:
$ sudo cat /sys/kernel/security/lockdown
none [integrity] confidentiality

We should add a descriptive warning if kernel lockdown is enabled and set to confidentiality - in which case we should also exit as kprobes, debugfs, perf, and bpf_read will not work

itaysk:
in bpftrace they opted for a different lockdown detection strategy:
https://github.com/iovisor/bpftrace/issues/925#issuecomment-586427173
https://github.com/iovisor/bpftrace/pull/1152

rafaeldtinoco:
The functionality has been merged into libbpfgo. I'm now waiting for @grantseltzer to bump libbpfgo version dependency from tracee so I can propose:

```C
// kernel lockdown check

lockdown, err := helpers.Lockdown()
if err != nil {
	return err
}
if lockdown == helpers.CONFIDENTIALITY {
	return fmt.Errorf("kernel lockdown is set to 'confidentiality', can't load eBPF programs.")
}
if debug {
	fmt.Fprintf(os.Stdout, "OSInfo: Security Lockdown is '%v'\n", lockdown)
}
```

to tracee.

PR #1333, opened 2022-01-11
Bump to libbpfgo v0.2.3-libbpf-0.6.1 with libbpf 0.6.1

rafaeldtinoco:
- This is needed bump to solve https://github.com/aquasecurity/tracee/issues/295
- It contains fixes to breaking changes related to parseArgs
- https://github.com/aquasecurity/libbpfgo/issues/108 was discovered during this PR (should be resolved in the next libbpfgo version)

### DIFF
--- a/tracee-ebpf/tracee/argprinters.go
+++ b/tracee-ebpf/tracee/argprinters.go
@@ -86,7 +86,6 @@ func (t *Tracee) parseArgs(event *external.Event) error {
 		if reqArg := getEventArg(event, "request"); reqArg != nil {
 			if req, isInt64 := reqArg.Value.(int64); isInt64 {
 				ptraceRequestArgument, err := helpers.ParsePtraceRequestArgument(uint64(req))
-				// TODO: https://github.com/aquasecurity/libbpfgo/issues/108
 				ParseOrEmptyString(reqArg, ptraceRequestArgument, err)
 			}
 		}
--- a/tracee-ebpf/tracee/argprinters.go
+++ b/tracee-ebpf/tracee/argprinters.go
@@ -30,6 +30,18 @@ func (t *Tracee) parseArgs(event *external.Event) error {
 		}
 	}
 
+	EmptyString := func(arg *external.Argument) {
+		arg.Type = "string"
+		arg.Value = ""
+	}
+
+	ParseOrEmptyString := func(arg *external.Argument, sysArg helpers.SystemFunctionArgument, err error) {
+		EmptyString(arg)
+		if err == nil {
+			arg.Value = sysArg.String()
+		}
+	}
+
 	switch int32(event.EventID) {
 	case MemProtAlertEventID:
 		if alertArg := getEventArg(event, "alert"); alertArg != nil {
@@ -39,7 +51,6 @@ func (t *Tracee) parseArgs(event *external.Event) error {
 		}
 	case SysEnterEventID, SysExitEventID, CapCapableEventID, CommitCredsEventID, SecurityFileOpenEventID:
 		if syscallArg := getEventArg(event, "syscall"); syscallArg != nil {
-			//show syscall name instead of id
 			if id, isInt32 := syscallArg.Value.(int32); isInt32 {
 				if event, isKnown := EventsDefinitions[id]; isKnown {
 					if event.Probes[0].attach == sysCall {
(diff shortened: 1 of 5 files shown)


tracee-ebpf/tracee/argprinters.go
```diff
@@ -51,130 +51,212 @@ func (t *Tracee) parseArgs(event *external.Event) error {
 		if int32(event.EventID) == CapCapableEventID {
 			if capArg := getEventArg(event, "cap"); capArg != nil {
 				if capability, isInt32 := capArg.Value.(int32); isInt32 {
-					capArg.Value = helpers.ParseCapability(capability)
 					capArg.Type = "string"
+					capArg.Value = "" // arg has to always exist (even if errors below)
+					capabilityFlagArgument, err := helpers.ParseCapability(uint64(capability))
+					if err != nil {
+						return err
```
yanivagman, line 58: returning an error will skip this event - I don't think this is the behavior we want

yanivagman (commented):
I have to say that I don't like the new helpers API exposed by libbpfgo.
1. Why should we cast the arguments? The helpers should expect the correct type of the arguments, as these are determined by the kernel (e.g. mode_t is of type unsigned it, why should the helpers accept uint64?)
2. We increased the lines count per call from 1 to 6
3. We need to stringify the returned value

tracee-ebpf/tracee/argprinters.go
```diff
@@ -51,130 +51,212 @@ func (t *Tracee) parseArgs(event *external.Event) error {
...
 	case PtraceEventID:
 		if reqArg := getEventArg(event, "request"); reqArg != nil {
 			if req, isInt64 := reqArg.Value.(int64); isInt64 {
-				reqArg.Value = helpers.ParsePtraceRequest(req)
 				reqArg.Type = "string"
+				reqArg.Value = "" // arg has to always exist (even if errors below)
+				ptraceRequestArgument, err := helpers.ParsePtraceRequestArgument(uint64(req))
+				if err == nil {
+					reqArg.Value = ptraceRequestArgument.String()
+				} // TODO: https://github.com/aquasecurity/libbpfgo/issues/108
```
grantseltzer, line 93: Opened to fix this: https://github.com/aquasecurity/libbpfgo/pull/110

This PR along with https://github.com/aquasecurity/libbpfgo/pull/109 can get merged and we can do another quick release of libbpfgo, wdyt?

tracee-ebpf/tracee/argprinters.go
rafaeldtinoco: Ok, based on your feedback:

1) @grantseltzer what was the reasoning behind this change ?

2) Should we revert the libbpfgo change and keep things as they were ?

3) I agree about the error handling. The fact that some argument might not be correctly parsed doesn't necessarily mean we should skip the entire event.
  - I wasn't sure about returning errors but, since the new API had them, I thought it was because you wanted this behavior (assumed this was discussed at somepoint).

tracee-ebpf/tracee/argprinters.go
rafaeldtinoco: Done. Let's hold another release for now, let me try to adequate this PR to something acceptable to @yanivagman first. I'll change the error handling and address his concerns.

tracee-ebpf/tracee/argprinters.go
```diff
@@ -51,130 +51,197 @@ func (t *Tracee) parseArgs(event *external.Event) error {
...
+				if err == nil {
+					domArg.Value = socketDomainArgument.String()
+				}
 			}
 		}
 		if typeArg := getEventArg(event, "type"); typeArg != nil {
 			if typ, isInt32 := typeArg.Value.(int32); isInt32 {
-				typeArg.Value = helpers.ParseSocketType(uint32(typ))
 				typeArg.Type = "string"
+				typeArg.Value = ""
```
rafaeldtinoco, line 119: @yanivagman 

There are 2 reasons why the size of each block passed from 1 to multiple lines:

1. Now we're making the argument to exist (even when there is an error). This is needed or else errors in tracee-rules (parsing errors) happen.
2. We are now checking for errors. Despite not returning them, we only give the Arg a value if the helpers parsing succeeds (which seems right).

tracee-ebpf/tracee/argprinters.go
```diff
@@ -51,130 +51,197 @@ func (t *Tracee) parseArgs(event *external.Event) error {
...
+				socketTypeArgument, err := helpers.ParseSocketType(uint64(typ))
```
rafaeldtinoco, line 120: We could pass argument as interface and cast it inside the helper function so the callee wouldn't have to worry about this. Can we postpone this change a bit just to "adjust" things now in between libbpfgo and tracee ? Maybe open an issue for this ?

yanivagman:
A possible solution to keep one line instead of several is to write a wrapper for the parsers. We can pass the parser function and argument to the wrapper, and return a string in any case (empty string in case of qn error). Wdyt?

tracee-ebpf/tracee/argprinters.go
```diff
@@ -51,130 +51,197 @@ func (t *Tracee) parseArgs(event *external.Event) error {
...
+				if err == nil {
+					typeArg.Value = socketTypeArgument.String()
```
rafaeldtinoco, line 122: About the "String"-ification discussion. Independently of what is right, IMHO, this follows recent changes to libbpfgo (about having a specific Type and a String() to it, so we can control the type's values and do 1:1 with real ones).

rafaeldtinoco:
> A possible solution to keep one line instead of several is to write a wrapper for the parsers. We can pass the parser function and argument to the wrapper, and return a string in any case (empty string in case of qn error). Wdyt?

Hum, I thought about this as well. Let me work on it.

tracee-ebpf/tracee/argprinters.go
rafaeldtinoco: Related to: https://github.com/aquasecurity/tracee/pull/1333#issuecomment-1010289310

tracee-ebpf/tracee/argprinters.go
rafaeldtinoco: Let's see how the parser wrapper idea fits into this and then we see the overall change again.

rafaeldtinoco:
> > A possible solution to keep one line instead of several is to write a wrapper for the parsers. We can pass the parser function and argument to the wrapper, and return a string in any case (empty string in case of qn error). Wdyt?
> 
> Hum, I thought about this as well. Let me work on it.

@yanivagman problem with that is that the wrapper would have to have a function pointer prototype generic, something like:

```go
func (t *Tracee) parseArgsHandler(arg *external.Argument, parser func(uint64)interface{}) {
}
```

but golang needs 

```
cannot use helpers.ParseMmapProt (value of type func(rawValue uint64) helpers.MmapProtArgument) as func(uint64) interface{} value in argument to t.parseArgsHandler
```

a non generic prototype... and each event would require its own wrapper (or at least a group of wrappers, one to each parser type).

rafaeldtinoco:
> A possible solution to this would be to have the function pointer as an unsafe pointer...

Nope, even if I use unsafe pointers to have the function pointer prototype generic, I would have to "cast" the function pointer again to call it:

```
func (t *Tracee) parseArgsHandler(arg *external.Argument, parser unsafe.Pointer) {
        functionToCall := (*func(arg *external.Argument)ReturnType)
}
```

grantseltzer:
All of the parsed arguments types implement the SystemFunctionArgument interface, theoretically you can do something like this: 

```
func ParseOrEmptyString(sysArg SystemFunctionArgument, err error) string {
	if err != nil {
		return ""
	} else {
		return sysArg.String()
	}
}
```

and then call it like:

```
stringVal := ParseOrEmptyString(ParseOpenFlagArgument(rawArg)
```

tracee-ebpf/tracee/argprinters.go
grantseltzer: Changes were for reducing code duplication, surfacing errors, and so we can define constants to avoid having redefined strings everywhere.

rafaeldtinoco:
@grantseltzer used your idea for https://github.com/aquasecurity/tracee/pull/1333/commits/59f176d9248d96335ac5f62640fb06b30347e4f1

grantseltzer:
> @yanivagman is this good for you ? It's not as good as having a generic wrapper function using function pointers BUT, like said in previous comments, this idea isn't straightforward and would require re-casting the function pointer (adding complexity as well).
> 
> The way things are now, they are back to minimal # of lines, and, yet, checking for errors coming from helpers.Parse* functions.
> 
> @grantseltzer ^

I'm happy with this solution. Regardless of Yaniv's thoughts on this, we do need to decide how to handle the libbpfgo fixes that we just merged. While the process for libbpfgo/tracee's relationship isn't working well right now, for the time being I think the best solution is that i'll cut yet another release of libbpfgo and upgrade it here. Wdyt?

rafaeldtinoco:
> I'm happy with this solution. Regardless of Yaniv's thoughts on this, we do need to decide how to handle the libbpfgo fixes that we just merged. While the process for libbpfgo/tracee's relationship isn't working well right now, for the time being I think the best solution is that i'll cut yet another release of libbpfgo and upgrade it here. Wdyt?

Agreed. Let's wait his feedback now, and then do that if that applies. Thanks @grantseltzer.

yanivagman:
> All of the parsed arguments types implement the SystemFunctionArgument interface, theoretically you can do something like this:
> 
> ```
> func ParseOrEmptyString(sysArg SystemFunctionArgument, err error) string {
> 	if err != nil {
> 		return ""
> 	} else {
> 		return sysArg.String()
> 	}
> }
> ```
> 
> and then call it like:
> 
> ```
> stringVal := ParseOrEmptyString(ParseOpenFlagArgument(rawArg)
> ```

I like this solution

> @grantseltzer used your idea for [0f2f821](https://github.com/aquasecurity/tracee/commit/0f2f821e094509cc1c98b85ffa394ca12eda21e9)

@rafaeldtinoco it doesn't seem to be exactly what @grantseltzer suggested (which shuold be more compact). Any reason for the change?

rafaeldtinoco:
> @rafaeldtinoco it doesn't seem to be exactly what @grantseltzer suggested (which shuold be more compact). Any reason for the change?

no particular reason, we were chatting on teams when did those comments, I can try to adjust.

yanivagman:
Should we link #295 to this issue?

rafaeldtinoco:
@yanivagman could you check now ? I've spoken to @grantseltzer about his idea, and changed the last commit. I have also included a last minute commit to fix formatting only (tracee-rules).

rafaeldtinoco:
Tests made:

```
$ sudo ./dist/tracee-ebpf -o format:json -o option:parse-arguments --trace comm=strace | ./dist/tracee-rules --input-tracee file:stdin --input-tracee format:json

*** Detection ***
Time: 2022-01-13T17:41:13Z
Signature ID: TRC-2
Signature: Anti-Debugging
Data: map[]
Command: strace
Hostname: fujitsu

$ sudo ./dist/tracee-ebpf -o format:gob -o option:parse-arguments --trace comm=strace | ./dist/tracee-rules --input-tracee file:stdin --input-tracee format:gob

*** Detection ***
Time: 2022-01-13T17:41:31Z
Signature ID: TRC-2
Signature: Anti-Debugging
Data: map[]
Command: strace
Hostname: fujitsu
```

rafaeldtinoco:
@yanivagman I'm waiting for your +1 on this one (for you not to loose track). Thanks!

tracee-ebpf/tracee/events_decoder.go
```diff
@@ -213,7 +213,11 @@ func readSockaddrFromBuff(buff io.Reader) (map[string]string, error) {
 	if err != nil {
 		return nil, err
 	}
-	res["sa_family"] = helpers.ParseSocketDomain(uint32(family))
+	socketDomainArg, err := helpers.ParseSocketDomainArgument(uint64(family))
+	if err != nil {
+		return nil, err
+	}
```
yanivagman, line 219: Do we really want to return an error in that case?
We can just stringify the family number instead

rafaeldtinoco:
@grantseltzer and @yanivagman 

I'm currently having a problem with a debugging printf() leftover at: https://github.com/aquasecurity/libbpfgo/issues/115 (for the socket type handling). I think we should wait libbpfgo to stabilize (proposed at https://github.com/aquasecurity/libbpfgo/pull/119) before moving forward with changes.

rafaeldtinoco:
@grantseltzer okay, after we stabilised libbpfgo with https://github.com/aquasecurity/libbpfgo/pull/119, this PR bumped libbpfgo to latest commit id. I'm now waiting a new release of libbpfgo (now with main fixed and libbpf bumped to 0.6.1) to add a new commit to this PR bumping to that release id.

grantseltzer (approved):
LGTM

PR #1334, opened 2022-01-11
entrypoint.sh: raise timeout to 60

rafaeldtinoco:
Fixes: 1325


yanivagman:
This should be a temporary fix until we will remove the bpf compilation from code, right?

rafaeldtinoco:
> This should be a temporary fix until we will remove the bpf compilation from code, right?

Or we can go directly with: https://github.com/aquasecurity/tracee/pull/1273 and this wouldn't be needed. I need your take there whenever you can, please.

yanivagman:
> > This should be a temporary fix until we will remove the bpf compilation from code, right?
> 
> Or we can go directly with: #1273 and this wouldn't be needed. I need your take there whenever you can, please.

Yes, we can wait for #1273. I don't think this change is needed

PR #1335, opened 2022-01-11
tracee-ebpf: fix missing sched_process_exit with pid=new

### LINKED ISSUE #1331: No "sched_process_exit" events if flag "--trace pid=new" is set
Is it a feature?

yanivagman:
fix #1331


rafaeldtinoco (approved):
makes sense. fixes the issue.

PR #1336, opened 2022-01-11
Bugfix: Add warning docs for group exit flag about race conditions

AlonZivony:
We noticed that the flag could have `true` value for multiple `sched_process_exit` events if they happen at almost the same because of concurrency issue (value used in the `do_exit` function in an atomic way, but we use the value afterwards so we are vulnerable to concurrent changes in the value).


tracee-ebpf/tracee/events_definitions.go
```diff
@@ -5722,6 +5722,9 @@ var EventsDefinitions = map[int32]EventDefinition{
 		Sets:           []string{"default", "proc", "proc_life"},
 		Params: []external.ArgMeta{
 			{Type: "long", Name: "exit_code"},
+			// The filed value represents that all threads exited at the event time.
```
yanivagman, line 5725: typo: filed -> field

yanivagman (approved):

PR #1342, opened 2022-01-13
getFileHash returning error and renaming it

mtcherni95:


tracee-ebpf/tracee/tracee.go
```diff
@@ -1007,23 +1007,26 @@ func boolToUInt32(b bool) uint32 {
...
+	return hex.EncodeToString(h.Sum(nil)), nil
 }
 
 func (t *Tracee) updateFileSHA() {
 	for k, v := range t.profiledFiles {
 		s := strings.Split(k, ".")
 		exeName := strings.Split(s[1], ":")[0]
 		filePath := fmt.Sprintf("%s.%d.%s", s[0], v.FirstExecutionTs, exeName)
-		fileSHA := getFileHash(filePath)
+		fileSHA, _ := computeFileHash(filePath)
```
danielpacak, line 1029: Is it intentional to skip error handling here? If not we may miss some logical errors. Shouldn't updateFileSHA() return error?

tracee-ebpf/tracee/write_capture.go
```diff
@@ -178,7 +178,7 @@ func (t *Tracee) processFileWrites() {
 			// Rename the file to add hash when last chunk was received
 			if meta.BinType == sendKernelModule {
 				if uint64(meta.Size)+meta.Off == kernelModuleMeta.Size {
-					fileHash := getFileHash(fullname)
+					fileHash, _ := computeFileHash(fullname)
```
danielpacak, line 181: What about errors? In other places we call `t.handleError(err)`

tracee-ebpf/tracee/tracee.go
mtcherni95: So yeah I agree with you. I left it without error check as it was before since I am not sure how we should handle error in this case. Maybe `updateFileSHA` should return error as well?

tracee-ebpf/tracee/write_capture.go
mtcherni95: Previous to this PR change if the value returned by `fileHash` was `""` then under the hood we had an error. The case in which `getFileHash` returns an empty string is not handled so far. I'm open to suggestions here.

tracee-ebpf/tracee/tracee.go
danielpacak: I'm always following this "fail fast" or "immediate and visible failure" principle. Unless there are good reasons for that. Here we should probably propagate this error to indicate abnormal behaviour of tracee. Normally computing hashes should just work

tracee-ebpf/tracee/write_capture.go
danielpacak: Looking at the next line it seems that a valid hash is expected to append it to the fullname:

```go
fileHash, _ := computeFileHash(fullname)
os.Rename(fullname, fullname+"."+fileHash)
```

mtcherni95: right, but if we have an error we would append an empty string, not sure this is something intended though.

tracee-ebpf/tracee/tracee.go
mtcherni95: right, so this function should be changed as well - updateFileSha should return an error as well.

danielpacak:
As discussed with @mtcherni95 we'll revisit error handling in other places. This PR is just focused on computeFileHash.

danielpacak (approved):
LGTM!

PR #1347, opened 2022-01-17
tracee-ebpf: remove mem_prot_alert dependencies

yanivagman:
mem_prot_alert event no longer depends on mmap and mprotect events as we now save syscall args anyway.
Remove these dependencies


grantseltzer (approved):
LGTM

Issue #1343, opened 2022-01-15
Drop events/rules with low priority on high load


yanivagman:
When system load is high, we might be required to drop some events/rules.
Currently we don't have a mechanism to prioritize events/rules, neither a mechanism to reduce load consumed by tracee-ebpf and tracee-rules.
To improve system performance on high loads, the following can be implemented:
1. Add priority field for each event
2. Add priority field for each rule
3. In tracee-ebpf, update a bpf map (use already existing chosen_events map?) with events to drop that are of low priority when required
4. On rules engine in tracee-rules, don't send events to rules with low priority when required
5. Implement load monitoring in tracee-ebpf
6. Implement load monitoring in tracee-rules
7. Expose an API to set events/rules priority
8. Expose an API to provide statistics of tracee-ebpf and tracee-rules dropped events/rules

itaysk:
> On events pipeline in tracee-ebpf, drop events with low priority when required

that might be too late, if we're defining the desired solution here, I think we want to drop is in eBPF (shoud_trace)

> On rules engine in tracee-rules, don't send events to rules with low priority when required

isn't this redundant if we're dropping the events in tracee-ebpf?

> Implement load monitoring in tracee-ebpf
Implement load monitoring in tracee-rules
Expose an API to provide statistics of tracee-ebpf and tracee-rules dropped events/rules

related to #887 

> Expose an API to set events/rules priority

related to #636

yanivagman:
> > On events pipeline in tracee-ebpf, drop events with low priority when required
> 
> that might be too late, if we're defining the desired solution here, I think we want to drop is in eBPF (shoud_trace)
> 

Yes, I was just about to update this issue with the following suggestion: set bpf map with the events to drop and drop in bpf code.

> > On rules engine in tracee-rules, don't send events to rules with low priority when required
> 
> isn't this redundant if we're dropping the events in tracee-ebpf?
> 

No. There might be rules that use events with high importance, for example execve, yet the rule itself might not be that important

> > Implement load monitoring in tracee-ebpf
> > Implement load monitoring in tracee-rules
> > Expose an API to provide statistics of tracee-ebpf and tracee-rules dropped events/rules
> 
> related to #887
> 
> > Expose an API to set events/rules priority
> 
> related to #636

mtcherni95:
> Yes, I was just about to update this issue with the following suggestion: set bpf map with the events to drop and drop in bpf code.

[...]

PR #1353, opened 2022-01-18
implementing load mitigation logics for tracee-ebpf

mtcherni95:


pkg/ebpf/events_pipeline.go
```diff
@@ -126,52 +128,75 @@ func (t *Tracee) decodeEvents(outerCtx gocontext.Context) (<-chan *trace.Event,
...
+					for _, derivative := range derivatives {
+						select {
+						case t.config.ChanEvents <- derivative:
+							t.stats.EventCount.Increment()
+						case <-ctx.Done():
+							return
+						}
+					}
+				}
+			case <-throughputTicker.C: // proportional integral derivative
```
NDStrahilevitz, line 186: I think we should separate sampling and decision making, so we should have a ticker for sampling which we can fine tune for accuracy, and a ticker for load control which looks at the current derivative sample (which looks to be our target here)

NDStrahilevitz, line 186: Also, as it is this isn't really PID, which is fine, but we shouldn't use the term unless we're actually using the thing.

pkg/ebpf/events_pipeline.go
```diff
@@ -126,52 +128,75 @@ func (t *Tracee) decodeEvents(outerCtx gocontext.Context) (<-chan *trace.Event,
...
+				{
+					lostEventsCurrent := t.stats.LostEvCount.Read()
+					delta := lostEventsCurrent - lostEvents
+					lostEvents = lostEventsCurrent
+					if delta > 0 {
+						if d, _ := t.DecreaseLoad(); len(d.ChangeInEvents.AffectedEvents) != 0 {
```
NDStrahilevitz, line 192: Couldn't this logic easily cause oscillations? I think if we separate the sampling as I said above we can reach a more clean solution which won't oscillate (in PID usually you use the D value to stop it, in that case it's f'' of Lost Events).

NDStrahilevitz:
Just a thought, we can export the current threshold as a metric in `Stats`, this way we can track the health probe's action in prometheus.

rafaeldtinoco:
@mtcherni95 and @NDStrahilevitz.. we've been discussing multiple things and ideas about fixing userland filtering (first by bringing the values from tracee-rules signatures to the current filtering mechanisms we have, and then, later, moving things to eBPF filtering). Should we close this for now and re-open once we have a clear picture ? Either way I think Nadav can use this as a base for his PRs on filtering improvements.

mtcherni95:
I believe we should either keep it or integrate this PR in another one. Anyways I think that the event pressure engine according to priorities is something we do want to have in Tracee. IMHO I wouldn't mind to keep this PR open, refinish it and push it. By default the feature could temporarily be set to off.

yanivagman:
can be linked to #1343

PR #1354, opened 2022-01-18
containers: optimize getContainerIdFromCgroup

NDStrahilevitz:
The function used to recompile a constant regex, now it is precompiled on module load and reused for matching.

### DIFF
--- a/pkg/containers/containers.go
+++ b/pkg/containers/containers.go
@@ -141,6 +141,10 @@ func (c *Containers) CgroupUpdate(cgroupId uint64, path string, ctime time.Time)
 	return info, nil
 }
 
+var (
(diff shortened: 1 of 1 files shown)


itaysk (approved):

pkg/containers/containers.go
```diff
@@ -169,7 +173,8 @@ func getContainerIdFromCgroup(cgroupPath string) (string, string) {
 			id = strings.TrimPrefix(id, "libpod-")
 		}
 
-		if matched, _ := regexp.MatchString(`^[A-Fa-f0-9]{64}$`, id); matched {
+		regex := containerIdFromCgroupRegex
+		if matched := regex.MatchString(id); matched {
```
mtcherni95, line 177: why not using `sync.Once` for `regexp.MustCompile` instead of defining a global var? I think it would be cleaner.

PR #1360, opened 2022-01-18
Move signal handing out of tracee-ebpf package

### LINKED ISSUE #975: Don't handle signals in tracee package, but in main

liamg:
Resolves #975

Other projects that wish to use tracee-ebpf as a library can now use the package with a `context.Context` to stop running: `Run() error` is now `Run(ctx context.Context) error`. Previously the `Run()` method could only be stopped using `SIGINT`/`SIGTERM`.

The tracee-ebpf binary now creates a context that is cancelled by these interrupts and passes it to the tracee-ebpf package so existing behaviour is preserved.

NOTE: I've imported `"context"` using the `gocontext` alias as `context` is already used elsewhere in the same package for another import. 

NOTE 2: This is my first tracee PR and I'm pretty unfamiliar with things atm, so feel free to rip this apart if necessary :)

### DIFF
--- a/tracee-ebpf/tracee/tracee.go
+++ b/tracee-ebpf/tracee/tracee.go
@@ -1,6 +1,7 @@
 package tracee
 
 import (
+	gocontext "context"
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
@@ -10,7 +11,6 @@ import (
 	"math"
 	"net"
 	"os"
-	"os/signal"
 	"path"
 	"path/filepath"
 	"strconv"
@@ -46,7 +46,6 @@ type Config struct {
 	KernelConfig       *helpers.KernelConfig
 	ChanEvents         chan external.Event
 	ChanErrors         chan error
-	ChanDone           chan struct{}
 }
 
 type CaptureConfig struct {
@@ -130,10 +129,6 @@ func (tc Config) Validate() error {
 
 	if tc.ChanErrors == nil {
 		return errors.New("nil errors channel")
-	}
-
-	if tc.ChanDone == nil {
-		return errors.New("nil done channel")
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


yanivagman:
Thanks @liamg !
This is definitely something that we need to take out of the tracee package into main.
@mtcherni95 I see that you assigned yourself for this task, do you mind taking a look at this PR?

mtcherni95:
> Thanks @liamg !
> This is definitely something that we need to take out of the tracee package into main.
> @mtcherni95 I see that you assigned yourself for this task, do you mind taking a look at this PR?

no problem - I'll review :)

tracee-ebpf/tracee/tracee.go
```diff
@@ -949,7 +947,7 @@ func (t *Tracee) Run() error {
 	go t.handleEvents(t.config.ChanDone)
 	go t.processFileWrites()
 	go t.processNetEvents()
-	<-sig
+	<-ctx.Done()
```
mtcherni95, line 950: I would add here a comment stating that this operation is actually blocking.

mtcherni95 (commented):
Looks good overall! I just gave a small comment. Please squash your commits into one.

yanivagman (commented):
A different approach to pass context (instead of `Run(ctx)`) might be through Tracee config (which is given to `tracee.New()`).
Although this will change Tracee's config API, we can then remove `cfg.ChanDone` and pass ctx to the events pipeline instead. I think this approach is cleaner than the `ChanDone` approach we have today to terminate the pipeline (which requires to close the channel to signal program termination). WDYT @liamg @mtcherni95 ?

liamg:
@mtcherni95 Thanks, I'll add the comment :+1: 

@yanivagman That sounds good, I'll take a look later on and see if I can do what you suggested :+1:

liamg:
@yanivagman I've made that change - WDYT?

mtcherni95:
> A different approach to pass context (instead of `Run(ctx)`) might be through Tracee config (which is given to `tracee.New()`). Although this will change Tracee's config API, we can then remove `cfg.ChanDone` and pass ctx to the events pipeline instead. I think this approach is cleaner than the `ChanDone` approach we have today to terminate the pipeline (which requires to close the channel to signal program termination). WDYT @liamg @mtcherni95 ?
> 
> **Edit: we might not even need to pass it through the config as the pipeline is started in Run(). So the comment is more about removing ChanDone from Tracee config and using ctx instead for the pipeline.**

yeah, that's a good idea.

yanivagman:
> @yanivagman I've made that change - WDYT?

Thanks @liamg.
LGTM
Seems that a rebase is now required

liamg:
@yanivagman Thanks, have rebased now :+1:

yanivagman (approved):

Issue #1359, opened 2022-01-18
capture network icmp packets


roikol:
hey,

right now we are only capturing packets of l4 protocols: tcp and udp.
we should add support for capturing packets of protocol: icmp.

itaysk:
Just out of curiosity can you share the use case?

roikol:
the use case is visibility (network wise) on the processes we monitor. right now if a process is communicating through icmp, we don't capture it in the resulted pcap.

security wise - icmp could be used to exfiltrate data between a process/malware on a victim machine to the C&C server of the attacker.

PR #1362, opened 2022-01-19
Capture icmp

roikol:
solves #1359.

using inet_sendmsg/inet6_sendmsg to save process context of icmp requests to network_map.
using icmp_rcv/icmpv6_rcv to delete from network_map after we got the response.
using icmp_send/icmp6_send to delete from network_map when there was no response.


tracee-ebpf/tracee/co-re/missing_definitions.h
```diff
@@ -49,6 +49,19 @@
 #define sk_flags          __sk_common.skc_flags
 #define sk_rxhash         __sk_common.skc_rxhash
 
+#define IPPROTO_HOPOPTS		0	/* IPv6 hop-by-hop options	*/
```
rafaeldtinoco: Could you squash changes to missing_definitions into a single commit and pay attention to numbers being aligned and separated from the macro by spaces ?

```C
#define IPPROTO_ICMPV6          58

#define ICMP_ECHO               8
#define ICMP_EXT_ECHO           42

#define ICMPV6_ECHO_REQUEST     128
```

tracee-ebpf/tracee/co-re/vmlinux.h
```diff
@@ -503,6 +503,24 @@ struct sockaddr_in6 {
 	__u32 sin6_scope_id;
 };
 
+struct msghdr {
```
rafaeldtinoco, line 506: Now, this file, all tabs are expanded in the beginning (to real tabs). I think we should change that to tabs being expanded to spaces but, before that, would you mind complying with the rest of the file and starting your structs indentation with real tabs ? (just so all the file is the same for now).

rafaeldtinoco (commented):
@roikol are you still working on this ? The git log says "capture icmp" but you're not sending it to `net_events` map yet nor changed processNetEvents() in userland.. just wondering if this is a preparation for something else that you would like to be merged first OR you're still working in this.

rafaeldtinoco:
> @roikol are you still working on this ? The git log says "capture icmp" but you're not changing processNetEvents() in userland.. just wondering if this is a preparation for something else that you would like to be merged first OR you're still working in this.

Also, shouldn't we be saving ICMP type & code into net_packet as well ? So processNetEvents could process it and save into pcap file ?

rafaeldtinoco:
I would also ask you to be more descriptive in your git logs about the changes you're making and why.

roikol:
> I would also ask you to be more descriptive in your git logs about the changes you're making and why.

@rafaeldtinoco no problem :)
should i squash the changes of the identation and tabs?

rafaeldtinoco:
> so this PR currently saves ICMP packets into the pcap. you can run it and open the resulted pcap to see packets (would be great actually to see that it works on a station other than my own :) ).

Yes, that will be my next step (sorry for not having done it yet, I wanted to confirm you were done). It's been a while I reviewed the pcap logic.

Let me give it a try then.

rafaeldtinoco:
> @rafaeldtinoco no problem :)
> should i squash the changes of the identation and tabs?

I would have all changes in a single commit and fix the indentation for sure.

rafaeldtinoco:
> there is actually no need to change anything in the userspace part. up until now we ignored any packet in `tc_probe()` which wasn't tcp or udp

Right, and you're including the `connect_id`:

`key: {port=0, proto=icmp, address=source}, value: {cmd, tid}`

into `network_map` so the ingress/egress probe can capture skbs from the "cmd".

(sorry took me sometime to read/remember the net capture code)

* I have some concerns here. And I'm not saying we should address them (if needed) in this PR, but at least document as TODOs (if they make sense).

----

0. The map key works because port is usually unique, per task, for existing TCP/UDP flows.. but 0 would make it not unique per existing task and ICMP packet.

> If I send a `echo request` and a `timestamp` before receiving anything, and I receive the `echo reply`.. I'll loose the `timestamp reply` because we haven't indexed map key using `icmp type & code`.

Does that make sense ?

----

1. What were the execution paths within kernel you considered to have:

=> `inet_sendmsg` to add a task to `network_map` (connect_id = (0, icmp, saddr))

=> `icmp_send` to remove the task from `network_map` (connect_id = (0, icmp, daddr))
=> `icmp_recv` to remove the task from `network_map` (connect_id = (0, icmp, saddr))

Is it guaranteed that we will always pull `connect_id` from `network_map` ?

What if I send an `echo request` and never get a "host unreachable" for the gateway ?

This is me pinging a unreachable host and receiving unreach icmp from my gateway (running in a kvm guest):

```
ping-2784731 [001] d... 114003.055236: bpf_trace_printk: INET_SENDMSG (ICMP)
ping-2784731 [001] d... 114004.079389: bpf_trace_printk: INET_SENDMSG (ICMP)
ping-2784731 [005] d... 114005.099339: bpf_trace_printk: INET_SENDMSG (ICMP)

qemu-system-x86-9383    [000] d... 114006.046856: bpf_trace_printk: ICMP_RECV
qemu-system-x86-9383    [000] d... 114006.046920: bpf_trace_printk: ICMP_RECV
qemu-system-x86-9383    [000] d... 114006.046973: bpf_trace_printk: ICMP_RECV
```

It works, it purges the `connect_id` from `network_map` but if I did not have received the unreachable icmp reply... then I suppose the map would have the entry forever (is that it ?).

Maybe we should purge all `network_map` entries for a task when it exits... something like that. I'm not sure this does not happen with TCP/UDP as well.

It's hard to dump the map contents with `bpftool` because this map has packet payload on each entry.

@yaniv does this make sense ? dumping `network_map` contents for a task when it exists ? just to make sure we don't have leaks on it ?

----

> so this PR currently saves ICMP packets into the pcap. you can run it and open the resulted pcap to see packets (would be great actually to see that it works on a station other than my own :) ).

It worked for me:

```
$ sudo tcpdump -n -r ./capture.pcap
reading from file ./capture.pcap, link-type EN10MB (Ethernet), snapshot length 65535
23:49:47.675929 IP 192.168.100.151 > 1.1.1.1: ICMP echo request, id 59, seq 1, length 64
23:49:47.679328 IP 1.1.1.1 > 192.168.100.151: ICMP echo reply, id 59, seq 1, length 64
23:49:48.677337 IP 192.168.100.151 > 1.1.1.1: ICMP echo request, id 59, seq 2, length 64
23:49:48.680781 IP 1.1.1.1 > 192.168.100.151: ICMP echo reply, id 59, seq 2, length 64
23:49:49.679275 IP 192.168.100.151 > 1.1.1.1: ICMP echo request, id 59, seq 3, length 64
23:49:49.683251 IP 1.1.1.1 > 192.168.100.151: ICMP echo reply, id 59, seq 3, length 64
23:49:50.681310 IP 192.168.100.151 > 1.1.1.1: ICMP echo request, id 59, seq 4, length 64
23:49:50.685414 IP 1.1.1.1 > 192.168.100.151: ICMP echo reply, id 59, seq 4, length 64
```

```
$ sudo tcpdump -n -r ./capture.pcap
reading from file ./capture.pcap, link-type EN10MB (Ethernet), snapshot length 65535
00:16:28.747589 IP 192.168.100.151 > 2.2.2.2: ICMP redirect 254.6.108.119 to host 0.0.0.0, length 56
```

I had issues with ordering of packets in the pcap file when capturing from
loopback device.. but its unrelated to this PR and only happened in lo device.

roikol:
thanks @rafaeldtinoco !

> 0. The map key works because port is usually unique, per task, for existing TCP/UDP flows.. but 0 would make it not unique per existing task and ICMP packet.
> 
> > If I send a `echo request` and a `timestamp` before receiving anything, and I receive the `echo reply`.. I'll loose the `timestamp reply` because we haven't indexed map key using `icmp type & code`.
> 
> Does that make sense ?

yes it is. sounds like a good idea to add the type and code. but this led me to think what happens when two different processes send an icmp to the same ip at the same time. we might have a mismatch of the packet process context. i'll have to figure out a way to handle this.



> 1. What were the execution paths within kernel you considered to have:
> 
> => `inet_sendmsg` to add a task to `network_map` (connect_id = (0, icmp, saddr))
> 
> => `icmp_send` to remove the task from `network_map` (connect_id = (0, icmp, daddr)) => `icmp_recv` to remove the task from `network_map` (connect_id = (0, icmp, saddr))
> 
> Is it guaranteed that we will always pull `connect_id` from `network_map` ?
> 
> What if I send an `echo request` and never get a "host unreachable" for the gateway ?

in these cases the kernel uses the `icmp_send` funcs to send an error back to the socket. so we should be covered in this regard.



> It works, it purges the `connect_id` from `network_map` but if I did not have received the unreachable icmp reply... then I suppose the map would have the entry forever (is that it ?).
> 
> Maybe we should purge all `network_map` entries for a task when it exits... something like that. I'm not sure this does not happen with TCP/UDP as well.
> 
> It's hard to dump the map contents with `bpftool` because this map has packet payload on each entry.
> 
> @Yaniv does this make sense ? dumping `network_map` contents for a task when it exists ? just to make sure we don't have leaks on it ?

we are using a map of type BPF_MAP_TYPE_LRU_HASH so we don't have leaks. if we'd want to delete from the map on process exit, then we'll have to have map from process context (host_tid) to the connect_id - which is the same as network_map, but the key and value switched. we can do that, let me know what you think.



> I had issues with ordering of packets in the pcap file when capturing from
> loopback device.. but its unrelated to this PR and only happened in lo device.

yes it happens on the loopback device. this is because each packet is captured both on the ingress and the egress - so every packet appears twice in the pcap. we can check in `tc_egress()` if the interface is a loopback device, and if it is - skip this packet (we captured it in the `tc_ingress()`). but i agree it should be in a different PR.

rafaeldtinoco:
> > If I send a `echo request` and a `timestamp` before receiving anything, and I receive the `echo reply`.. I'll loose the `timestamp reply` because we haven't indexed map key using `icmp type & code`.
 
> yes it is. sounds like a good idea to add the type and code. but this led me to think what happens when two different processes send an icmp to the same ip at the same time. we might have a mismatch of the packet process context. i'll have to figure out a way to handle this.

It seems that this problem exists for TCP and UDP as well (much harder to happen though), right ? It's hard to know skb context before its delivered to a task, I'm afraid =(. Conntrack code could give us some ideas, but it's likely much more complex than we what we need.

> > Maybe we should purge all `network_map` entries for a task when it exits... something like that. I'm not sure this does not happen with TCP/UDP as well.

> we are using a map of type BPF_MAP_TYPE_LRU_HASH so we don't have leaks. if we'd want to delete from the map on process exit, then we'll have to have map from process context (host_tid) to the connect_id - which is the same as network_map, but the key and value switched. we can do that, let me know what you think.

True, I failed to see it was a LRU_HASH. Nah, as long as we don't have leaks I'm fine.

> > I had issues with ordering of packets in the pcap file when capturing from
> > loopback device.. but its unrelated to this PR and only happened in lo device.
 
> yes it happens on the loopback device. this is because each packet is captured both on the ingress and the egress

Oh, yep, that makes sense. In lo devices we have to do whatever tcpdump does (for the lack of consistency):

```
$ sudo tcpdump -i lo src 127.0.0.1 or dst 127.0.0.1
tcpdump: verbose output suppressed, use -v[v]... for full protocol decode
listening on lo, link-type EN10MB (Ethernet), snapshot length 262144 bytes
11:52:41.839414 IP localhost > localhost: ICMP echo request, id 68, seq 1, length 64
11:52:41.839446 IP localhost > localhost: ICMP echo reply, id 68, seq 1, length 64
11:52:42.857376 IP localhost > localhost: ICMP echo request, id 68, seq 2, length 64
11:52:42.857407 IP localhost > localhost: ICMP echo reply, id 68, seq 2, length 64
11:52:43.881408 IP localhost > localhost: ICMP echo request, id 68, seq 3, length 64
11:52:43.881443 IP localhost > localhost: ICMP echo reply, id 68, seq 3, length 64
11:52:44.905366 IP localhost > localhost: ICMP echo request, id 68, seq 4, length 64
11:52:44.905396 IP localhost > localhost: ICMP echo reply, id 68, seq 4, length 64
```

Thanks!

rafaeldtinoco (changes requested):
I have requested changes during last iteration. Marking it to reflect current status.

roikol:
> I have requested changes during last iteration. Marking it to reflect current status.

yeah, i didn't forget! i changed my focus after we defined the milestone, but i should be back on it soon.

roikol:
hey @rafaeldtinoco,

it took me some time to get back to work on this PR.

we talked about adding the icmp type and code to avoid mismatch in the process context of the packets.
turns out that addind icmp type and code adds a lot of complexity and breaks the current way we capture packets - icmp type&code changes, by design, to indicate reply/response. therefore, adding it to the `connect_id` would mean to maintain the state of replies and responses.

after digging into the kernel code i've found that icmp has an identification value.
this value is set to be the source port of the socket, and it doesn't change between the packets. also, it is unique per icmp flow.
i used the `connect_id.port` to save this id (makes sense to me because this value derives from the source port), which means there's no need to change the `local_net_id_t` struct.

let me know what you think :)

rafaeldtinoco:
> after digging into the kernel code i've found that icmp has an identification value.
> this value is set to be the source port of the socket, and it doesn't change between the packets. also, it is unique per socket.
> i used the `connect_id.port` to save this id (makes sense to me because this value derives from the source port), which means there's no need to change the `local_net_id_t` struct.

It makes sense to me as well...

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3592,6 +3594,176 @@ int BPF_KPROBE(trace_tcp_connect)
...
+        connect_id.port = READ_KERN(icmph->un.echo.id);
+
+        if (send) {
+            connect_id.address.s6_addr32[3] = daddr;
+        }
+        else {
+            if (icmp_type == ICMP_ECHO) {
+                return 0;
+            }
+            #ifdef ICMP_EXT_ECHO
```
rafaeldtinoco, line 3644: We should remove this, right ? You're always defining it in `missing_definitions.h`, no ? And, considering this is an enum value defined in some kernels, and not others, the correct CO-RE way of using it would be to check for enum (like described at https://nakryiko.com/posts/bpf-core-reference-guide/#bpf-core-enum-value-exists).

rafaeldtinoco (changes requested):
Okay, I have re-tested this and was going to merge it but there is a minor ifdef you should revisit before we do so. Could you re-request my review once you figure best way to handle the enum value check ?

Thanks!

pkg/ebpf/c/tracee.bpf.c
roikol: but the thing is that it is not an enum, its a define (https://elixir.bootlin.com/linux/v5.17.1/source/include/uapi/linux/icmp.h#L72).
if i understand correctly, `bpf_core_enum_value_exists()` doesn't help in this case. or am i missing something?

rafaeldtinoco: No, you are not. I confused things when reviewing. I see it clearly now and it makes sense to always define it in CO-RE and to check type. Sorry for the noise.

rafaeldtinoco (approved):
Code looks good after changes regarding unique flow identification (port).

Issue #1363, opened 2022-01-19
tracee-ebpf :adding support to net events as regular events


itamarmaouda101:
Hey,

Right now we are only printing network events in the console by fmt.Printf() and not by the events channel ( because now they are only debug events).

We should add support for printing network events also in json and also add a support for adding more net -events like that other events.

Currently I'm working on creating that support by adding a `network_protocols` package and directory which will able us to add net events easily and also support tracing them using the command line

itamarmaouda101:
i think you right.
i hope we will support more net events  (for example protocols like dns, http, irc, etc) in the feature .
i created a draft PR which still prints the events only in debug mode (#1364 ) but  i think that later on it will be great idea to remove it from the `--debug`

yanivagman:
> And, IMO, we should remove the network debugging from "--debug".
> 
> @yanivagman WDYT about it ? IIRC we've discussed this in the past.
> 
> I'm doing VNC to the machine being observed/tested and its almost impossible to check "--debug" initial messages because of so many "network packets debugging" passing through my screen (even when --output none is given).

Yes, I agree. In the first place I wanted to create --debug-net flag, but @itaysk opposed to this idea, saying that --debug is enough.

In addition to this, I think this issue should be more general than just the json printer. Net events should be converted to an external.Event struct and pushed to the events channel, which then uses the regular printer interface

itamarmaouda101:
> In addition to this, I think this issue should be more general than just the json printer. Net events should be converted to an external.Event struct and pushed to the events channel, which then uses the regular printer interface

defiantly agree , i did exactly that in my PR.
 i created a NET_PACKET event handler that assign the  first argument to be that packet data (Src and Dst Ips and Ports ).
 now in the CR with @AlonZivony we thought about how can we  will support more then just Net packets (for example IRC, DNS, ICMP ,etc protocols)

itamarmaouda101:
I changed the issue subject to be more general because in my PR I will expend the network functionality as well

rafaeldtinoco:
This has been implemented.

PR #1364, opened 2022-01-19
tracee-ebpf :add support to net events  as regular events

itamarmaouda101:
solves #1363 
Add a  support for converting net events to a regular events in tracee-ebpf by sending the packets to the events channel.
also its add a network_protocols directory and package to support adding new net-events  as regular events.


itamarmaouda101:
the output should be like that:
{"timestamp":1302579175629,"processId":744,"threadId":744,"parentProcessId":744,"hostProcessId":744,"hostThreadId":744,"hostParentProcessId":1,"userId":101,"mountNamespace":4026532596,"pidNamespace":0,"processName":"systemd-resolve","hostName":"","containerId":"","eventId":"4000","eventName":"NetPacket","argsNum":5,"returnValue":0,"stackAddresses":null,"args":[{"name":"src_ip","type":"string","value":"127.0.0.53"},{"name":"dst_ip","type":"string","value":"127.0.0.53"},{"name":"src_port","type":"uint16","value":53},{"name":"dest_port","type":"uint16","value":37734},{"name":"protocol","type":"uint8","value":17}]}

and BTW there is 3 commits because i changed the way i print the network values.
frist i did with 5 args and then i changed it to one struct that hold the hole packet data (due tracee support up to 8 args)

tracee-ebpf/tracee/net_events.go
```diff
@@ -143,23 +190,23 @@ func (t *Tracee) processNetEvents() {
 				switch netEventId {
 				case DebugNetSecurityBind:
 					fmt.Printf("%v  %-16s  %-7d  debug_net/security_socket_bind LocalIP: %v, LocalPort: %d, Protocol: %d\n",
```
AlonZivony, line 192: All these printings print the same thing.
I think you should put them inside a const and maybe even create a print function for this message.

AlonZivony (changes requested):
The code is in good condition and you did some nice things.
My main complaints are about magics and code structure.
The only other main thing is that you don't send net events unless you are in a debug mode, which is not clear to me why. Can you explain it?

I do think that if you are able to remove the dependency on the `Tracee` struct it would be the best, because we want to divide the tracee package to sub-packages and it is not possible with all the dependencies.

tracee-ebpf/tracee/net_events.go
```diff
@@ -4,122 +4,169 @@ import (
...
-					continue
-				}
-
-				// todo: maybe we should not flush every packet?
-				err = t.pcapWriter.Flush()
-				if err != nil {
-					t.handleError(err)
-					continue
 				}
 			} else if t.config.Debug {
```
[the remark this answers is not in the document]
itamarmaouda101: i cant because of the `t.pcapWriter.WritePacket` will force me to mark that function as Tracee func

tracee-ebpf/tracee/net_events.go
itamarmaouda101: agreed , but i wondering if this is major change for one PR or its fine.
what you think ?

itamarmaouda101: also i think that all of that net events that arent NET_PACKET event will eventually need to move to other place and get a proper handling (also think about the IRC,DNS, HTTP,etc events that coming) i think its better to create a different PR for that change dont you ?

tracee-ebpf/tracee/net_events.go
```diff
@@ -4,122 +4,169 @@ import (
...
-				idx, ok := t.ngIfacesIndex[int(ifindex)]
-				if !ok {
-					t.handleError(err)
+				ctx, exist := t.processTree.processTreeMap[hostTid]
+				if !exist {
+					t.handleError(fmt.Errorf("couldn't find the process"))
 					continue
 				}
-
+				evt := createNetEvent(int(timeStamp), hostTid, processName, netEventId, "NetPacket", pktMeta, ctx)
```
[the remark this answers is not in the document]
itamarmaouda101: yeah agree , actually there is little talk about it in the issue #1363 . 
i think the end goal is to move it from debug mode and maybe give it a flag.
currently im following the  issue and i think after this CR we can talk with @yanivagman and talk about  what to do

tracee-ebpf/tracee/net_events.go
```diff
@@ -4,122 +4,169 @@ import (
...
+func isIpv6(ip [16]byte) bool {
+	zeroedPattern := make([]byte, 9, 9)
+	if bytes.Compare(ip[:9], zeroedPattern) == 0 {
+		return false
+	}
+	return true
+}
+
+//takes the packet metadata and create argument array with that data
+func createPacketMetadataArg(packetmeta PktMeta) []external.Argument {
```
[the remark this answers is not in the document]
itamarmaouda101: the function create the basic argument for the "net event " event . that is has now only one argument that is `external.PktMeta`.
the function takes a similar struct (`packetMeta`) and assigns some of it to the `extenal.PktMeta struct` 
the deafferents between the structs is firstly  the type of the Ip members, which has to be string in the extenal.PktMeta  (because of the print type we want to print ) struct and in the packetMeta has to byte array  (because of the binary. Read function) 

Secondly the  reason for the whole assigning in the `arg.SrcIp `and `arg.DstIp` is because im casing the ip from byte array to string and note that the functions `netaddr.IPFrom16` and  `netaddr.IPFrom4` expects for known size of byte  array which means i cant cast the whole byte array to size specific ( like to do something like that `netaddr.IPFrom4(packet.DestIP[:4]).String()' 

also the reason for having the `extenal.PktMeta` struct is for able Tracee rules to recognize that type

and BTW if you still thinks its great idea to move the ip checking and assign to another function, let me know (:

tracee-ebpf/tracee/net_events.go
```diff
@@ -4,122 +4,169 @@ import (
...
+	eventArgs := make([]external.Argument, 0, 0)
+	arg := external.PktMeta{}
+	if isIpv6(packetmeta.SrcIP) {
+		arg.SrcIP = netaddr.IPFrom16(packetmeta.SrcIP).String()
+	} else {
+		var ip [4]byte
+		copy(ip[:], packetmeta.SrcIP[12:])
```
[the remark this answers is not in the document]
itamarmaouda101: because `netaddr.IPFrom4` are expecting to known byte array size 
it tried earlier to do it with slicing but it throw error

tracee-ebpf/tracee/net_events.go
```diff
@@ -4,122 +4,169 @@ import (
...
+		arg.SrcIP = netaddr.IPFrom4(ip).String()
```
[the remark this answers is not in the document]
itamarmaouda101: because `netaddr.IPFrom4` are expecting to known byte array size 
it tried earlier to do it with slicing but it throw error

tracee-ebpf/tracee/net_events.go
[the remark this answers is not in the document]
AlonZivony: But I am confused about the slicing of `ip`, not of `packetmeta.SrcIP`.
The slicing just return `ip` as it is.

tracee-ebpf/tracee/net_events.go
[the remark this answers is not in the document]
AlonZivony: If you want a general function for creating all the net events args, create one named `createNetEventsArgs` and call in it to a `createPacketMetadataArg`.
This way `createPacketMetadataArg` will return an `external.Argument` as expected, and the enveloping `createNetEventsArgs` will return `[]external.Argument` as expected of it.

tracee-ebpf/tracee/net_events.go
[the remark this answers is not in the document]
AlonZivony: You can make it a method of `Tracee` for now, but this is too much code for debug specific logic

tracee-ebpf/tracee/net_events.go
AlonZivony: > agreed , but i wondering if this is major change for one PR or its fine. what you think ?

I don't think that putting it in this PR makes it too large. This is the required amount of effort for the size of this PR.



> also i think that all of that net events that arent NET_PACKET event will eventually need to move to other place and get a proper handling (also think about the IRC,DNS, HTTP,etc events that coming) i think its better to create a different PR for that change dont you ?

I think that when it comes, someone will create sub-function for it. But handling the debug printing for now should be done properly.

tracee-ebpf/tracee/net_events.go
```diff
@@ -4,122 +4,169 @@ import (
...
-				if !ok {
-					t.handleError(err)
+				ctx, exist := t.processTree.processTreeMap[hostTid]
+				if !exist {
+					t.handleError(fmt.Errorf("couldn't find the process"))
 					continue
 				}
-
+				evt := createNetEvent(int(timeStamp), hostTid, processName, netEventId, "NetPacket", pktMeta, ctx)
 				if t.config.Debug {
```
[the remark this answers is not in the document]
itamarmaouda101: Now we are only support packet tracing in debug mode and there isn't any flag that specify tracing a network event.
i assume there will be another PR that  will fix that by creating a specific flag and also a parser layer for other net event like all the net debug events that now we still printing without the `event cannel` 
**ill added it also to the docs

itamarmaouda101:
Update:
 the current PR include a `dns_request `net event which is a example of how to add a network event ill add here also a documentation like HOW-TO guide with the `dns_request `as example

AlonZivony (changes requested):
The code is in very high level - you did great job at dividing the code to files and functions.
I have really minor comments that you can address very quickly.
Another over-all comment I have is that there are still many debug names of things in the code - it seems you wanted to change them so I advise you go over all the code and make sure not deubg name is left before re-submitting.

tracee-ebpf/tracee/network_protocols/kernel_functions_based_net_events.go
```diff
@@ -0,0 +1,77 @@
...
+	var pktMetaData FunctionBasedPacket
+	err := binary.Read(payload, binary.LittleEndian, &pktMetaData)
+	if err != nil {
+		return pktMetaData, err
+	}
+
+	return pktMetaData, nil
+}
+
+func CreateDebugPacketMetadataArg(event *external.Event, packet FunctionBasedPacket) {
```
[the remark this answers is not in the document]
itamarmaouda101: I changed it  to `CreateFunctionBasedPacketMetadataArgs`

tracee-ebpf/tracee/net_events.go
```diff
@@ -16,162 +15,30 @@ func (t *Tracee) processNetEvents() {
...
-						DestPort uint16
-						Protocol uint8
-						_        [3]byte //padding
-					}
-					err = binary.Read(dataBuff, binary.LittleEndian, &pktMeta)
-					if err != nil {
+			if ShouldCapture {
+				interfaceIndex, ok := t.ngIfacesIndex[int(cap.InterfaceIndex)]
+				if ok {
+					if err := t.writePacket(cap.PacketLen, time.Unix(int64(evt.Timestamp), 0), interfaceIndex, dataBuff); err != nil {
```
roikol, line 38: can you explain why changing the time object? it used to be created like so: `time.Unix(0, int64(timeStamp+t.bootTime))`

tracee-ebpf/tracee/network_protocols/kernel_functions_based_net_events.go
```diff
@@ -0,0 +1,67 @@
...
+		return pktMetaData, err
+	}
+
+	return pktMetaData, nil
+}
+
+func CreateFunctionBasedPacketMetadataArgs(event *external.Event, packet FunctionBasedPacket) {
+	eventArgs := make([]external.Argument, 0, 0)
+	arg := external.FunctionBasedPacket{}
+	arg.LocalIP = parseIP(packet.LocalIP)
```
roikol, line 50: you can use `external.PktMeta` for the first arg (src ip, src port, dst ip, dst port, protocol). then you can break the rest into args:
tcp_new_state
tcp_old_state
sock_ptr

tracee-ebpf/tracee/network_protocols/kernel_functions_based_net_events.go
```diff
@@ -0,0 +1,67 @@
...
+	arg.LocalPort = packet.LocalPort
+	arg.RemoteIP = parseIP(packet.RemoteIP)
+	arg.Protocol = packet.Protocol
+	if event.EventID == int(NetInetSockSetState) {
+		arg.RemotePort = packet.RemotePort
+		arg.TcpNewState = packet.TcpNewState
+		arg.TcpOldState = packet.TcpOldState
+		arg.SockPtr = packet.SockPtr
+	}
+	evtArg := external.Argument{
```
roikol, line 60: maybe it'd be better to create a function for this under the `external` package like your `CreateStringArgument()` func. (maybe under external/debug.go that i suggested above).

tracee-ebpf/tracee/network_protocols/net_packet.go
```diff
@@ -0,0 +1,63 @@
...
+	if err != nil {
+		return pktMetaData, err
+	}
+	return pktMetaData, nil
+}
+
+//takes the packet metadata and create argument array with that data
+func createNetPacketMetadataArg(packet PacketMeta) []external.Argument {
+	eventArgs := make([]external.Argument, 0, 0)
+	arg := external.PktMeta{}
```
roikol, line 46: same here: you can use `external.PktMeta` for the first arg (src ip, src port, dst ip, dst port, protocol).

tracee-ebpf/tracee/tracee.go
```diff
@@ -920,8 +923,9 @@ func (t *Tracee) getProcessCtx(hostTid int) (ProcessCtx, error) {
 		if err != nil {
 			return processCtx, err
 		}
-		processCtx, err = t.ParseProcessContext(processCtxBpfMap)
-		t.processTree.processTreeMap[hostTid] = processCtx
+		cgroupId := t.containers.GetCgroupInfo(binary.LittleEndian.Uint64(processCtxBpfMap[8:16])).ContainerId
```
roikol, line 926: why not getting the containerID in the `processContext.ParseProcessContext()` func?
also - please change `cgroupId` var to `containerId` or something alike.

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -4148,6 +4155,24 @@ static __always_inline bool skb_revalidate_data(struct __sk_buff *skb, uint8_t *
...
+
+/* check_protocols is checking for potential network protocols based packets
+ * if it is finds a potential protocol based packet it sends it to the user-space with the eventID
+ * in the user-space we checking the packet data more deeply to verify it is the acual protocl,
+ * and if it is- we parsing it and alerting to the user space
+*/
+static __always_inline void check_protocols(struct __sk_buff *skb, uint8_t* head, uint8_t* tail, struct bpf_map_def *events_channel, net_packet_t* pkt, u64 flags){
+   if (is_dns_request(head, tail, skb, pkt) && event_chosen(NET_DNS_REQUEST)){
+        pkt->event_id = NET_DNS_REQUEST;
+        bpf_perf_event_output(skb, &net_events, flags, pkt, sizeof(net_packet_t));
```
roikol, line 4172: in this case we send the packet twice - once to capture the packet (save it in the pcap), and once to parse DNS values from it. 
you created generic handlers from network events, so i think we can send the packet only once.
if it is recognized as DNS packet here in the bpf, then change the `pkt.event_id` to `NET_DNS_REQUEST`, if not, then the `pkt.event_id` should stay as it is (`NET_PACKET`).

then - in `ProcessNetEvent()` you should return `true` in the case of `NetDnsRequest`.

let me know if that could work.

pkg/external/external.go
```diff
@@ -237,3 +244,31 @@ func (alert MemProtAlert) String() string {
...
+//for supporting networking events in tracee rules
+type PktMeta struct {
+	SrcIP    string `json:"srcIP"`
+	DestIP   string `json:"destIP"`
+	SrcPort  uint16 `json:"srcPort"`
+	DestPort uint16 `json:"destPort"`
+	Protocol uint8  `json:"protocol"`
+}
+
+type FunctionBasedPacket struct {
```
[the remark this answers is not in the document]
itamarmaouda101: I can create another file for that under the external directory.
we have to put the protocols under the external directory to let tracee-rules know the struct.
let me know what you think

pkg/external/external.go
```diff
@@ -237,3 +244,31 @@ func (alert MemProtAlert) String() string {
...
+	RemotePort  uint16  `json:"remotePort"`
+	Protocol    uint8   `json:"protocol"`
+	_           [3]byte //padding
+	TcpOldState uint32  `json:"tcpOldState"`
+	TcpNewState uint32  `json:"tcpNewState"`
+	_           [4]byte //padding
+	SockPtr     uint64  `json:"sockPtr"`
+}
+
+type DnsQueryData struct {
```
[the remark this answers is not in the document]
itamarmaouda101: It can be but I think it will be better as struct to save more place for other arguments.

Note that we save the `dns_request` event arguments as `DnsQueryData` for each arg.
In that case we support until 7 questions in event, so if ill expend it from struct to field it will be up to 2 questions

tracee-ebpf/tracee/events_definitions.go
```diff
@@ -6152,4 +6142,86 @@ var EventsDefinitions = map[int32]EventDefinition{
...
+		},
+	},
+	network_protocols.NetDnsRequest: {
+		ID32Bit: sys32undefined,
+		Name:    "net_dns_request",
+		Probes:  []probe{},
+		Sets:    []string{},
+		Params: []external.ArgMeta{
+			{Type: "external.PktMeta", Name: "PacketMetaData"},
+			{Type: "external.DnsQueryData", Name: "dnsQuestion"},
```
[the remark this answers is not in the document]
itamarmaouda101: Answered above, let me know what you think(:

tracee-ebpf/tracee/events_definitions.go
```diff
@@ -6152,4 +6142,86 @@ var EventsDefinitions = map[int32]EventDefinition{
 			{Type: "const char*", Name: "container_id"},
 		},
 	},
+	network_protocols.NetPacket: {
+		ID32Bit: sys32undefined,
+		Name:    "net_packet",
+		Probes:  []probe{},
+		Sets:    []string{},
```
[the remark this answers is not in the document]
itamarmaouda101: What you mean by set ? 
if you mean like a list of events.
I made somting similar at the [`tracee-ebpf/tracee/network_protocols/net_events_processor.go` ](https://github.com/aquasecurity/tracee/blob/4a2c8a32ea1642b448589bc443ee8b8b638c6a55/tracee-ebpf/tracee/network_protocols/net_events_processor.go#L9)

tracee-ebpf/tracee/net_events.go
itamarmaouda101: Because of  the time that came from the event is more accurate then the time in the user-space

tracee-ebpf/tracee/network_protocols/kernel_functions_based_net_events.go
itamarmaouda101: well yes and no, because of in the function based net events (formerly we called it `debug_net_events`) we used a similar struct with different name members.
Like `src ip `changed to `local ip`,  and `src port`, to `loacl port` etc...

I can change it if  you think it ok, let me know what you think

tracee-ebpf/tracee/network_protocols/kernel_functions_based_net_events.go
itamarmaouda101: I can move it but Why you think it should be under the external, isn't need to be under the  `tracee-ebpf/tracee/network_protocols/kernel_functions_based_net_events.go` ?

tracee-ebpf/tracee/tracee.bpf.c
itamarmaouda101: sound good, ill try it! (:

tracee-ebpf/tracee/tracee.go
itamarmaouda101: Because of that if we will want to fetch the `containerId` from the `processContext.ParseProcessContext()` we will need to use` Tracee Containers API` which will lead us to append this function as Tracee function and then in will make the `process context package`  be apart of Tracee.

I found out that to get he `containerID` as argument will free us to include that function as Tracee and make the code more compressed with Tracee functions

tracee-ebpf/tracee/network_protocols/net_packet.go
itamarmaouda101: That's what I do, don't I?

tracee-ebpf/tracee/tracee.bpf.c
itamarmaouda101: it Works, after ill fixed your other comments ill push it too

tracee-ebpf/tracee/network_protocols/net_packet.go
roikol: your'e right, my bad :)

tracee-ebpf/tracee/net_events.go
roikol: yes but the time that comes from the event is in nanoseconds since system boot time (it comes from `bpf_ktime_get_ns()`). or am i missing something here?

tracee-ebpf/tracee/network_protocols/kernel_functions_based_net_events.go
roikol: ohh i see.. ok so don't change it

tracee-ebpf/tracee/network_protocols/kernel_functions_based_net_events.go
roikol: you're right, keep it here.

yanivagman:
Thanks @itamarmaouda101.
I think this PR should be split to several (at least two) different PRs as it is hard to review in its current state.
The commits themselves are not self contained on their own, and it is just too hard to review the whole code at once. 
So, to start with, I think that the DNS logic with the new events should be in its own PR. Even without the DNS logic, you can probably break your PR to several self contained commits, so it will be easier to review.

Thanks.

PR #1367, opened 2022-01-19
New release workflow

grantseltzer:
This PR:
- Creates a new `Makefile.release` which has two targets:
    - Snapshot
        - Builds tracee-ebpf, tracee-rules, rules
        - Builds an archive of build artifacts along with license
        - Takes checksum of archive
        - Builds container images
    - Publish
        - Pushes container images to dockerhub
        - Creates github release with the build artifact archive 

For both of these targets the main environment variables to set are:

PUSH_DOCKER_REPO (default: aquasec/tracee)

SNAPSHOT_VERSION - the tag or SHA to label the release/snapshot as (default: latest git SHA)

This PR also does the following:

- Updates the github action for releasing to use this new makefile
- Fixes existing Makefile to cleanup intermediate images

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,9 +1,8 @@
 name: release
-on: 
-  workflow_dispatch:
-    inputs:
-      tag:
-        required: true
+on:
+  push:
+    tags:
+      - "v*"
 jobs:
   release:
     runs-on: ubuntu-20.04
@@ -17,4 +16,5 @@ jobs:
           gh auth login --with-token <<<'${{ secrets.GITHUB_TOKEN }}'
           docker login docker.io --username ${{ secrets.DOCKERHUB_USER }} --password-stdin <<<'${{ secrets.DOCKERHUB_TOKEN }}'
       - name: release
-        run: make release DOCKER=1 RELEASE_TAG=${{ github.event.inputs.tag }}
+        run: |
+          make -f builder/Makefile.release SNAPSHOT_VERSION=$(git describe --tags --abbrev=0) PUSH_DOCKER_REPO=aquasec/tracee snapshot publish
--- a/builder/Makefile.release
+++ b/builder/Makefile.release
@@ -158,9 +158,9 @@ snapshot: \
 #
 	# build binaries (tracee-ebpf (CO-RE + BTFHUB), tracee-rules, rules)
 	$(MAKE) -f builder/Makefile.tracee-make ubuntu-prepare ARG="all"
-	BTFHUB=1 $(MAKE) -f builder/Makefile.tracee-make ubuntu-make ARG="all"
+	BTFHUB=1 STATIC=1 $(MAKE) -f builder/Makefile.tracee-make ubuntu-make ARG="all"
 	# create the tar ball and checksum files
-	$(CMD_TAR) -czf $(OUT_ARCHIVE) $(RELEASE_FILES)
+	$(CMD_TAR) --exclude="*.so" -czf $(OUT_ARCHIVE) $(RELEASE_FILES)
 	$(CMD_CHECKSUM) $(OUT_ARCHIVE) > $(OUT_CHECKSUMS)
 	# build official slim container image (CO-RE + BTFHUB)
 	BTFHUB=1 $(MAKE) -f builder/Makefile.tracee-container build-tracee
--- a/builder/Makefile.release
+++ b/builder/Makefile.release
@@ -26,6 +26,7 @@ CMD_TAR ?= tar
 CMD_GIT ?= git
 CMD_RM ?= rm
 CMD_TOUCH ?= touch
+CMD_MKDIR ?= mkdir
 
 .ONESHELL:
 .check_%:
@@ -52,6 +53,7 @@ env:
 	@echo "CMD_TAR               $(CMD_TAR)"
 	@echo "CMD_TOUCH             $(CMD_TOUCH)"
 	@echo "CMD_RM                $(CMD_RM)"
+	@echo "CMD_MKDIR             $(CMD_MKDIR)"
 	@echo ---------------------------------------
 	@echo "PUSH_DOCKER_REPO      $(PUSH_DOCKER_REPO)"
 	@echo ---------------------------------------
@@ -125,7 +127,7 @@ RELEASE_NOTES ?= $(OUTPUT_DIR)/release_notes.txt
 
 $(OUTPUT_DIR):
 #
-	@$(CMD_MKDIR) -p $@
+	$(CMD_MKDIR) -p $@
 	$(CMD_TOUCH) $(RELEASE_NOTES)
 
 #
@@ -147,8 +149,8 @@ PUSH_DOCKER_REPO ?= aquasec/tracee
 
 .PHONY: snapshot
 snapshot: \
-	| $(OUT_DIR) \
-	.check_tree \
+	$(OUTPUT_DIR) \
+	| .check_tree \
 	.check_$(CMD_DOCKER) \
 	.check_$(CMD_TAR) \
 	.check_$(CMD_CHECKSUM) \
@@ -171,7 +173,7 @@ snapshot: \
 
 .PHONY: publish
 publish: \
-	$(OUT_DIR) \
+	$(OUTPUT_DIR) \
 	$(OUT_ARCHIVE) \
 	$(OUT_CHECKSUMS) \
 	| .check_tree \
--- a/builder/Makefile.release
+++ b/builder/Makefile.release
@@ -1,0 +1,196 @@
+#
+# Responsible for creating tracee snapshots for testing and releasing
+#
+
+.PHONY: all
+all: help
+release: snapshot publish
+
+#
+# make
+#
+
+.ONESHELL:
+SHELL = /bin/sh
+
+MAKEFLAGS += --no-print-directory
+
+#
+# tools
+#
+
+CMD_DOCKER ?= docker
+CMD_CHECKSUM ?= sha256sum
+CMD_GITHUB ?= gh
+CMD_TAR ?= tar
+CMD_GIT ?= git
+CMD_RM ?= rm
+CMD_TOUCH ?= touch
+
+.ONESHELL:
+.check_%:
+#
+	@command -v $* >/dev/null
+	if [ $$? -ne 0 ]; then
+		echo "missing required tool $*"
+		exit 1
+	else
+		touch $@ # avoid target rebuilds due to inexistent file
+	fi
+
+#
+# environment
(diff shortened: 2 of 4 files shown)


mtcherni95 (commented):
Looks very promising! shouldn't we also run tests maybe?

rafaeldtinoco:
I'm reviewing this now... sorry for the delay

builder/Makefile.tracee-container
```diff
@@ -93,7 +93,12 @@ build-alpine-tracee-core: \
 		-t $(ALPINE_TRACEE_CORE_CONTNAME):latest \
 		--build-arg=BTFHUB=0 \
 		--build-arg=FLAVOR=tracee-core \
+		--target tracee-core \
 		.
+		$(CMD_DOCKER) images \
+		--filter "label=AS=tracee-make" \
+		--format "{{.ID}}" \
+		| xargs docker rmi
```
rafaeldtinoco, line 101: nit: xargs $(CMD_DOCKER) for uniformity (and in all other "xargs docker rmi" places).

builder/Makefile.tracee-container
```diff
@@ -108,7 +113,12 @@ build-alpine-tracee-core-btfhub: \
 		-t $(ALPINE_TRACEE_CORE_BTFHUB_CONTNAME):latest \
 		--build-arg=BTFHUB=1 \
 		--build-arg=FLAVOR=tracee-core \
+		--target tracee-core \
 		.
+		$(CMD_DOCKER) images \
+		--filter "label=AS=tracee-make" \
+		--format "{{.ID}}" \
+		| xargs docker rmi
```
rafaeldtinoco, line 121: here.

builder/Makefile.tracee-container
```diff
@@ -123,7 +133,12 @@ build-alpine-tracee-nocore: \
 		-t $(ALPINE_TRACEE_NOCORE_CONTNAME):latest \
 		--build-arg=BTFHUB=0 \
 		--build-arg=FLAVOR=tracee-nocore \
+		--target tracee-nocore \
 		.
+	$(CMD_DOCKER) images \
+		--filter "label=AS=tracee-make" \
+		--format "{{.ID}}" \
+		| xargs docker rmi
```
rafaeldtinoco, line 141: here.

builder/Makefile.release
```diff
@@ -0,0 +1,109 @@
...
+.check_release_tag:
+#
+	@if [ -z "$(RELEASE_TAG)" ]; then
+		echo "missing required variable RELEASE_TAG"
+		exit 1
+	fi
+
+#
+# release
+#
```
rafaeldtinoco, line 46: Please jump lines after:

```

#
# title
# 

```

So all makefiles look the same.

builder/Makefile.release
```diff
@@ -0,0 +1,109 @@
...
+CMD_CP ?= cp
+CMD_GIT ?= git
+CMD_RM ?= rm
+CMD_MAKE ?= make
+
+TARGET_CORE_IMAGE_NAME = build-alpine-tracee-core
+TARGET_BTFHUB_IMAGE_NAME = build-alpine-tracee-core-btfhub
+TARGET_NOCORE_IMAGE_NAME = build-alpine-tracee-nocore
+
+MAKE_CONTAINER = $(CMD_MAKE) -f builder/Makefile.tracee-container
```
rafaeldtinoco, line 21: would you mind declaring the variables in the first (or the one that makes most sense) target that uses it ? I followed this convention so we didn't have a bunch of variables in the beginning of the Makefiles.

builder/Makefile.release
```diff
@@ -0,0 +1,109 @@
...
+	.check_$(CMD_CHECKSUM) \
+	.check_$(CMD_GITHUB) \
+	.check_$(CMD_TAR) \
+	.check_$(CMD_CP) \
+#
+	-rm $(release_notes)
+	echo '## Changelog' > $(release_notes)
+	$(CMD_GIT) log --pretty=oneline --abbrev=commit --no-decorate --no-color tags/$(shell $(CMD_GIT) describe --tags --abbrev=0)..HEAD >> $(release_notes)
+	echo '' >> $(release_notes)
+#
```
rafaeldtinoco, line 62: All makefiles now have a "help" target, that displays something like the comments you have in your makefile. Would you mind creating the help target with descriptions ?

builder/Makefile.release
```diff
@@ -0,0 +1,109 @@
...
+	echo '## Docker images' >> $(release_notes)
+	echo '- `docker pull docker.io/$(PUSH_DOCKER_REPO):latest (BTFHub enabled CO:RE tracee)`' >> $(release_notes);
+	echo '- `docker pull docker.io/$(PUSH_DOCKER_REPO):$(TARGET_BTFHUB_IMAGE_NAME)-$(RELEASE_TAG)`' >> $(release_notes);
+	echo '- `docker pull docker.io/$(PUSH_DOCKER_REPO):$(TARGET_CORE_IMAGE_NAME)-$(RELEASE_TAG)`' >> $(release_notes);
+	echo '- `docker pull docker.io/$(PUSH_DOCKER_REPO):$(TARGET_NOCORE_IMAGE_NAME)-$(RELEASE_TAG)`' >> $(release_notes);
+
+.PHONY: clean
+clean:
+#
+	$(CMD_RM) .check*
```
rafaeldtinoco, line 108: Hum, I need to add this to other makefiles (to clean .check files and not to rely in Makefile.one only).

builder/Makefile.release
rafaeldtinoco: The CMD_xxxx variables should be above .check_% target (without variables in between them), for uniformity.

rafaeldtinoco (commented):
The change looks very good. Most of the comments are related to formatting only (to keep uniform with other Makefiles we have). I haven't tested the release, just read it. I wonder how you're testing it as well.

grantseltzer:
> The change looks very good. Most of the comments are related to formatting only (to keep uniform with other Makefiles we have). I haven't tested the release, just read it. I wonder how you're testing it as well.

I'm working on testing this in my fork, with a separate container registry. 

Also want to note, that I want to be careful that we have consensus on the container image tags. We have 3 images now:

- tracee-core
- tracee-core-btfhub
- tracee-nocore

So  the images have tags like:

- docker.io/aquasec/tracee:tracee-core-0.6.1
- docker.io/aquasec/tracee:tracee-core-btfhub-0.6.1
- docker.io/aquasec/tracee:tracee-nocore-0.6.1

And then there may be a need to have `docker.io/aquasec/tracee:latest` which I set to default to the latest `tracee-core-btfhub`. Thoughts? @rafaeldtinoco @itaysk @danielpacak

danielpacak (commented):
Not sure if it's possible now, but I think we should also have a way to just run the release target and see that it works and builds release artifacts (executable binaries and container images) locally without creating Git tag and publishing to GitHub or DockerHub. In some other build systems, e.g. GoReleaser, it's called a release snapshot. Only when we pass PUBLISH=1 or similar explicit flag it would trigger an actual release. (Ideally such release snapshot would run in the PR validation workflow before we run integration tests.)

grantseltzer:
I'm going to add Daniel's suggestion but for now this works:

https://github.com/grantseltzer/tracee/releases/tag/v0.9.16

.github/workflows/release.yaml
```diff
@@ -17,4 +17,4 @@ jobs:
           gh auth login --with-token <<<'${{ secrets.GITHUB_TOKEN }}'
           docker login docker.io --username ${{ secrets.DOCKERHUB_USER }} --password-stdin <<<'${{ secrets.DOCKERHUB_TOKEN }}'
       - name: release
-        run: make release DOCKER=1 RELEASE_TAG=${{ github.event.inputs.tag }}
+        run: make -f builder/Makefile.release release RELEASE_TAG=${{ github.event.inputs.tag }} PUBLISH=y
```
danielpacak, line 20: I think that fundamentally the release script or release Make target **should not** create nor push Git tags to the upstream or your forks.

Creating and pushing an annotated Git tag to the upstream should trigger GitHub release workflow that eventually creates GitHub release and publish container images to DockerHub, ECR Gallery, etc.

```
git tag -a v0.7.0 -m 'Release v0.7.0' && git push upstream v0.7.0
````

What's more, we must have a way to test release artifacts that we're about to publish.

To do that release scripts may support two modes run in two different scenarios:
1. [PR Validation Workflow] Release snapshot to make sure that we're ready to release anytime from the main branch and use snapshot build artifacts (i.e. executable binaries, signatures, container images) for integration tests:
   ```
   make -f builder/Makefile.release SNAPSHOT=1
   ```
   The SNAPSHOT=1 flag is here to say that we build snapshot release artifacts with the current Git SHA-1 not the latest Git tag
2. [Release Workflow] Do actual release, i.e. publish changes log, artifacts and container images:
    ```
   make -f builder/Makefile.release PUBLISH=1
   ```

danielpacak (changes requested):

.github/workflows/release.yaml
grantseltzer: Why have our release script double as a builder for integration tests instead of just have a separate integration test script? We could run integration tests at the start of the release script, but seems like it'd be better to separate that.

danielpacak: Release snapshot won't double as a builder for integration or e2e test. It's actually a prerequisite to run e2e tests. The idea is to build (expensive) container images only once and reuse them for e2e tests. Otherwise our CI workflow won't be very efficient. More importantly, Make targets and other build scripts are error prone and fragile and probably the only opportunity to test them out (except the publish step) is to build such release snapshot in the PR validation workflow.

Since we plan to add basic smoke test to the PR validation workflow we could reuse container images cached in the previous steps and avoid duplicating build logic in the test runner script. I hope it explains better now.

grantseltzer:
As it stands now, this is working. I believe this is how you described it @danielpacak. Tested it on my personal fork: https://github.com/grantseltzer/tracee/runs/4992692333?check_suite_focus=true

rafaeldtinoco:
I believe you're still working on this @grantseltzer, right ? Please let me know once you're good (so I can review). Thanks!

grantseltzer:
> I believe you're still working on this @grantseltzer, right ? Please let me know once you're good (so I can review). Thanks!

Could have commented in here instead of messaging you offline, but regardless yes this is ready for review. The only thing I want to point out is asking your thoughts on the convention of 'aquasec/tracee:full-v0.6.6', as in having the version number in the tag.

rafaeldtinoco (approved):
LGTM. I'll squash this and merge (fixing git log). Thanks a lot for this change and all the tests you've made.

Issue #1366, opened 2022-01-19
Fail to start tracee under high load


rafaeldtinoco:
By accident I was testing tracee-ebpf in parallel to lots of server loads I have and I saw that:

```
$ sudo ./dist/tracee-ebpf --debug -o option:parse-arguments --trace comm=ping --trace event=* --trace event!='sched*' --capture net=home
OSInfo: ID: ubuntu
OSInfo: ID_LIKE: debian
OSInfo: KERNEL_RELEASE: 5.13.0-25-generic
OSInfo: ARCH: x86_64
OSInfo: PRETTY_NAME: "Ubuntu 21.10"
OSInfo: VERSION_ID: "21.10"
OSInfo: VERSION: "21.10 (Impish Indri)"
OSInfo: VERSION_CODENAME: impish
BTF: bpfenv = false, btfenv = false, vmlinux = true
BPF: using embedded BPF object
unpacked CO:RE bpf object file into memory
libbpf: Kernel error message: Exclusivity flag on, cannot modify
2022/01/19 16:34:14 error creating Tracee: error creating process tree: readlink /proc/2562484/task/2562484/ns/mnt: no such file or directory
```

There might be a race issue because of error handling when the task exits and we're trying to read procfs during the initial load.

yanivagman:
This error is from the new process tree pr merged earlier today.
@itamarmaouda101 can you have a look?

itamarmaouda101:
sure , on it!

itamarmaouda101:
i checked that  and i notice that in case that error occurs in the parsing of task file while initiate the process-tree- we return error and then it failed to create tracee. 
i created a PR that fixing that by changing the `return err' to `continue` in the task parsing phase of the process tree

PR #1368, opened 2022-01-20
change to conitine instead exit while error occurs

itamarmaouda101:
solves #1366 
by changing the return to continue while error occurs .
if you want i can also remove the error check in the creation of Tracee to verify it never crush on that case

### DIFF
--- a/tracee-ebpf/tracee/process_tree.go
+++ b/tracee-ebpf/tracee/process_tree.go
@@ -160,15 +160,16 @@ func NewProcessTree() (*ProcessTree, error) {
 			taskStatus := fmt.Sprintf("/proc/%d/task/%v/status", pid, task)
 			data, err := ioutil.ReadFile(taskStatus)
 			if err != nil {
+				// process might have exited - ignore this task
 				continue
 			}
 			processStatus, err := parseProcStatus(data, taskStatus)
 			if err != nil {
-				return nil, err
+				continue
 			}
 			containerId, err := containers.GetContainerIdFromTaskDir(taskDir)
 			if err != nil {
-				return nil, err
+				continue
 			}
 			processStatus.ContainerID = containerId
 			p.processTreeMap[int(processStatus.HostTid)] = processStatus


tracee-ebpf/tracee/process_tree.go
```diff
@@ -164,11 +164,11 @@ func NewProcessTree() (*ProcessTree, error) {
 			}
 			processStatus, err := parseProcStatus(data, taskStatus)
 			if err != nil {
-				return nil, err
+				continue
```
mtcherni95, line 167: shouldn't we handle the error somehow, maybe we could push it to the error channel or something similar. WDYT?

itamarmaouda101: yes i think it possible.
but actually i don't know if there is any error channel or something like that.
if you know send me a reference to the code and ill add it RN (:

rafaeldtinoco: I think that we should skip the error handling for now. Walking procfs will lead to many issues when the host is overloaded (pids appearing and disappearing as directories, files being created and deleted, files that can't be read, files in binary format, etc etc).

rafaeldtinoco: Printing errors to stderr is an alternative, but could make the final user confused about what is actually happening (while there isn't a real problem going on).

itamarmaouda101: agreed (:
BTW @rafaeldtinoco do you think i should remove the error check when the calling to `NewProcessTree()`  or verify it wont failed the whole Tracee creation ?

yanivagman: I agree with @rafaeldtinoco.
In addition, even when there are no errors, some processes may have been created/exited between the time we initialize the process tree to the time we start monitoring `sched_process_xxx` tracepoints. This mismatch should probably be resolved by reiterating procfs after bpf program started, removing/adding processes as required (we may even need to do this more than once, to the point when there are no more processes to add/remove)

yanivagman: > agreed (: BTW @rafaeldtinoco do you think i should remove the error check when the calling to `NewProcessTree()` or verify it wont failed the whole Tracee creation ?

I think that we shuold keep this error check as there might be more "severe" errors, such as not being able to open or read /proc

rafaeldtinoco (approved):
LGTM, would you mind opening an issue for the situation @yanivagman quoted ? (about having to re-read procfs over and over after maps are created and bpf programs are fully loaded and initialized ?).

yanivagman:
Please also ammend the commit message (has a typo in it). Can be something like:
`tracee-ebpf: don't exit on a failure to read a procfs entry`

itamarmaouda101:
> tracee-ebpf: don't exit on a failure to read a procfs entry

added, and i also wrote a little explain why we aren't exiting there when error occurs

tracee-ebpf/tracee/process_tree.go
```diff
@@ -155,6 +155,8 @@ func NewProcessTree() (*ProcessTree, error) {
 		}
 		taskDir, err := os.Open(fmt.Sprintf("/proc/%d/task", pid))
 		processTasks, err := taskDir.Readdirnames(-1)
+
+		// if it failed to parse the task dir we continue to avoid failing tracee
```
yanivagman, line 159: I would put such a comment above the continue statement below, saying that:
// process might have exited - ignore this task

PR #1373, opened 2022-01-20
tracee-ebpf: introduce event dependencies

yanivagman:
Some events depend on other events to function properly.
Add a new field to event definition to declare those dependencies.
When an event is chosen by the user, set its dependencies essential.

This change will also help with the introduction of higher level events (#1310),
that are derived from other events received in userspace.


rafaeldtinoco (approved):
LGTM

Issue #1378, opened 2022-01-23
Bug: Events counted twice in the stats.eventCounter


AlonZivony:
After dividing the processEvent function to decode and process, the event count should have increased after decoding the event.
For some reason, it increases in both.
The result is that the count of events is twice the actual amount.

The increment in the process function should be removed.

yanivagman:
closed by #1379

PR #1379, opened 2022-01-23
events_pipeline: Fixed bug of counting events twice

AlonZivony:
Solves #1378


tracee-ebpf/tracee/events_pipeline.go
```diff
@@ -180,7 +180,7 @@ func (t *Tracee) processEvents(done <-chan struct{}, in <-chan *external.Event)
 
 			select {
 			case t.config.ChanEvents <- *event:
-				t.stats.eventCounter.Increment()
+				break
```
yanivagman, line 183: I think we should remove from the decode stage and not from here

yanivagman (approved):

Issue #1386, opened 2022-01-24
Write tracee-ebpf stats to stderr when output is JSON or GOB


AsafEitani:
The problem:
When running tracee-ebpf and outputting the events as JSON or GOB it's impossible to know if there were dropped events due to the fact the the stats are only printed in table format output.
This makes it hard to trace detection problems of tracee-rules.

The solution:
Print stats to stderr if there were lost events.

yanivagman:
> The problem: When running tracee-ebpf and outputting the events as JSON or GOB it's impossible to know if there were dropped events due to the fact the the stats are only printed in table format output. This makes it hard to trace detection problems of tracee-rules.
> 
> The solution: Print stats to stderr.

I agree that this is a problem, but I don't think printing to stderr is the solution. Stats are not error.
We already expose `GetStats()` which returns `external.Stats` as an API of tracee-ebpf.
We should probably implement one of the following (or both):
1. A server that can take requests from the users (e.g. calling GetStats) - #636
2. Sending metrics periodically - #887

AsafEitani:
@yanivagman Thing is, I wish to know if we have lost events after we tracee, to be able to know if I can trust the events generated.
I saw #887 but I think that doesn't really solves the problem.
I updated the issue to only print the stats if any events were lost - therefore it deserves an error, to my opinion.

yanivagman:
> @yanivagman Thing is, I wish to know if we have lost events after we tracee, to be able to know if I can trust the events generated. I saw #887 but I think that doesn't really solves the problem. I updated the issue to only print the stats if any events were lost - therefore it deserves an error, to my opinion.

Ok, that makes more sense.
Yet, you should know that we have an "errors counter" and "lost events counter", and we shouldn't mix between the two. So, a possible implementation might be the following (I added the third line in this function):

```
func (t *Tracee) processLostEvents() {
	for {
		lost := <-t.lostEvChannel
		t.stats.lostEvCounter.Increment(int(lost))
                t.config.ChanErrors <- fmt.Errorf("lost %d events", lost)
	}
}
```

yanivagman:
You can do the same for lostWrCounter and lostNtCounter, and even state which kind (regular, write, net) of events were lost

AsafEitani:
Good idea that way we will know will running! :)
PR on the way

AsafEitani:
#1387

PR #1387, opened 2022-01-24
add stats printing to stderr in case of lost events

AsafEitani:
Addresses #1386

### DIFF
--- a/tracee-ebpf/tracee/events_processor.go
+++ b/tracee-ebpf/tracee/events_processor.go
@@ -15,8 +15,10 @@ import (
 func (t *Tracee) processLostEvents() {
 	for {
 		lost := <-t.lostEvChannel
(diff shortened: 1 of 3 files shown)


yanivagman:
See my comment here: https://github.com/aquasecurity/tracee/issues/1386#issuecomment-1020118159

yanivagman (changes requested):
See my comment here: [#1386 (comment)](https://github.com/aquasecurity/tracee/issues/1386#issuecomment-1020118159)

yanivagman (approved):

yanivagman (commented):
LGTM
Just please add a comment describing why we added the check for lost>0 (https://github.com/aquasecurity/libbpfgo/issues/122) so we know why it is there

Issue #1257, opened 2021-12-15
Detection for hooked syscalls


itamarmaouda101:
Hey 

Sometimes malware like to hook system calls in order to change the Os flow in important actions such as listing files, opening files, opening sockets, and more.

Currently, I'm working to add to Tracee 2 events that will detect those hooks.

The first event will give us the ability to print the syscall table entries, and the second event will be a derived event which checks if each syscall is hooked by boundary check with the memory regions of the kernel.

For every syscall that is hooked Tracee returns its syscall name and the module that hooked the syscall, this is the output of detection of diamorphine rootkit ( an open-source popular lkm rookit):

{"timestamp":1638891820305021152,"processId":53642,"threadId":53642,"parentProcessId":53641,"hostProcessId":53642,"hostThreadId":53642,"hostParentProcessId":53641,"userId":0,"mountNamespace":4026531840,"pidNamespace":4026531836,"processName":"tracee-ebpf","hostName":"itamar-VirtualB","containerId":"","eventId":"1032","eventName":"detect_hooked_syscalls","argsNum":0,"returnValue":0,"stackAddresses":null,"args":[{"name":"hooked_syscalls","type":"hookedSyscallData[]","value":[{"SyscallName":"kill","ModuleOwner":"diamorphine"},{"SyscallName":"getdents","ModuleOwner":"diamorphine"},{"SyscallName":"getdents64","ModuleOwner":"diamorphine"}]}]}

Notice that currently, the detection event is only check hardcoded syscall numbers, which are syscalls that malware like to hook

PR #1394, opened 2022-01-24
tracee-ebpf: add event to detect hooked syscalls

itamarmaouda101:
Hi
this is a fixed version of #1222 PR, 
Its solves #1257 
I put it now as Draft until it will pass a CR  (:

**how it works:**
A special ioctl cmd is executed during tracee initialization. This
special ioctl cmd tells eBPF kprobe function that it should look the
syscalls table for a list of given syscalls and their addresses.

By checking runtime syscall handlers addresses, it is possible to know
if the handlers were hijacked by a kernel module, for example. They way
way it works is this: if the syscall handler address is not in kernel
text segment boundaries, it means it is not part of running kernel
(likely a hook somewhere else).

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -201,7 +201,8 @@ Copyright (C) Aqua Security inc.
 #define CALL_USERMODE_HELPER            1037
 #define DIRTY_PIPE_SPLICE               1038
 #define DEBUGFS_CREATE_FILE             1039
-#define MAX_EVENT_ID                    1040
+#define PRINT_SYSCALL_TABLE             1040
+#define MAX_EVENT_ID                    1041
 
 #define NET_PACKET                      4000
 
@@ -299,6 +300,10 @@ Copyright (C) Aqua Security inc.
 #define MAX_PATH_COMPONENTS             20
 #define MAX_BIN_CHUNKS                  110
 #endif
+
+#define IOCTL_FETCH_SYSCALLS            65        // randomly picked number for ioctl cmd
+#define NUMBER_OF_SYSCALLS_X86          18
+#define NUMBER_OF_SYSCALLS_ARM          14
 
 /*================================ eBPF KCONFIGs =============================*/
 
@@ -628,6 +633,7 @@ BPF_HASH(process_tree_map, u32, u32);                   // filter events by the 
 BPF_HASH(process_context_map, u32, process_context_t);  // holds the process_context data for every tid
 BPF_HASH(network_config, u32, int);                     // holds the network config for each iface
 BPF_HASH(ksymbols_map, ksym_name_t, u64)            // holds the addresses of some kernel symbols
+BPF_HASH(syscalls_to_check_map, int, u64);                // syscalls to discover
 BPF_LRU_HASH(sock_ctx_map, u64, net_ctx_ext_t);         // socket address to process context
 BPF_LRU_HASH(network_map, local_net_id_t, net_ctx_t);   // network identifier to process context
 BPF_ARRAY(config_map, u32, 4);                          // various configurations
@@ -2270,6 +2276,50 @@ static __always_inline struct pipe_inode_info *get_file_pipe_info(struct file *f
     return pipe;
 }
 
+static __always_inline void fetch_syscalls_addresses(event_data_t *data){
+    int key = 0;
+    u64 *table_ptr = bpf_map_lookup_elem(&syscalls_to_check_map, (void *)&key);
+    if (table_ptr == NULL){
+        return ;
+    }
+
+    unsigned long *syscall_table_addr = (unsigned long*) *table_ptr;
+    u64 idx;
+    u64* syscall_num_p;                  // pointer to syscall_number
+    u64 syscall_num;
+    unsigned long syscall_addr = 0;
+    int monitored_syscalls_amount = 0;
+    #if defined(bpf_target_x86)
+        monitored_syscalls_amount = NUMBER_OF_SYSCALLS_X86;
+        u64 syscall_address[NUMBER_OF_SYSCALLS_X86];
+    #elif defined(bpf_target_arm64)
+        monitored_syscalls_amount = NUMBER_OF_SYSCALLS_ARM;
+        u64 syscall_address[NUMBER_OF_SYSCALLS_ARM];
+    #else
+
+        return
+    #endif
+
+    __builtin_memset(syscall_address, 0, sizeof(syscall_address));
+    // the map should look like [syscall_table][syscall number 1][syscall number 2][syscall number 3]...
+    #pragma unroll
+    for(int i =0; i < monitored_syscalls_amount; i++){
+        idx = i+1;
+        syscall_num_p = bpf_map_lookup_elem(&syscalls_to_check_map, (void *)&idx);
+        if (syscall_num_p == NULL){
+            continue;
+        }
+        syscall_num = (u64)*syscall_num_p;
+        syscall_addr = READ_KERN(syscall_table_addr[syscall_num]);
+        if (syscall_addr == 0){
+            return;
+        }
+        syscall_address[i] = syscall_addr;
+    }
+    save_u64_arr_to_buf(data, (const u64 *)syscall_address, monitored_syscalls_amount, 0);
+    events_perf_submit(data, PRINT_SYSCALL_TABLE, 0);
+}
+
 /*============================== SYSCALL HOOKS ===============================*/
 
 // include/trace/events/syscalls.h:
@@ -2875,6 +2925,22 @@ int BPF_KPROBE(trace_do_exit)
     long code = PT_REGS_PARM1(ctx);
 
     return events_perf_submit(&data, DO_EXIT, code);
+}
+
+SEC("kprobe/security_file_ioctl")
+int BPF_KPROBE(trace_tracee_trigger_event )
+{
+    event_data_t data = {};
+
+    if (!init_event_data(&data, ctx))
+        return 0;
+
+    unsigned int cmd = PT_REGS_PARM2(ctx);
+
+    if (cmd == IOCTL_FETCH_SYSCALLS && get_config(CONFIG_TRACEE_PID) == data.context.host_pid){
+        fetch_syscalls_addresses(&data);
+    }
+    return 0;
 }
 
 // include/trace/events/cgroup.h:
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -122,6 +122,7 @@ Copyright (C) Aqua Security inc.
 #define U16_T                           14UL
 #define CRED_T                          15UL
 #define INT_ARR_2_T                     16UL
+#define UINT64_ARR_T                    17UL
 #define TYPE_MAX                        255UL
 
 #if defined(bpf_target_x86)
@@ -1583,6 +1584,57 @@ static __always_inline int save_str_to_buf(event_data_t *data, void *ptr, u8 ind
     }
 
     return 0;
+}
+
+static __always_inline int save_u64_arr_to_buf(event_data_t *data, const u64 __user *ptr,int len , u8 index){
+
+    // Data saved to submit buf: [index][u64 count][u64 1][u64 2][u64 3]...
+
+
+    u8 elem_num = 0;
+    // Save argument index
+    data->submit_p->buf[(data->buf_off) & (MAX_PERCPU_BUFSIZE-1)] = index;
+
+    // Save space for number of elements (1 byte)
+    u32 orig_off = data->buf_off+1;
+    data->buf_off += 2;
+
+    #pragma unroll
+    for (int i = 0; i < len; i++) {
+        u64 element = 0;
+        int err = bpf_probe_read(&element, sizeof(u64), &ptr[i]);
+        if (err !=0)
+            goto out;
+        if (data->buf_off > MAX_PERCPU_BUFSIZE - sizeof(u64) )
+                // not enough space - return
+                goto out;
+
+        void *addr = &(data->submit_p->buf[data->buf_off ]);
+        int sz = bpf_probe_read(addr, sizeof(u64), (void *)&element);
+        if (sz == 0) {
+            elem_num++;
+            if (data->buf_off > MAX_PERCPU_BUFSIZE )
+                // Satisfy validator
+                goto out;
+
+            data->buf_off += sizeof(u64);
+            continue;
+        } else {
+            goto out;
+        }
+    }
+    if (data->buf_off > MAX_PERCPU_BUFSIZE - sizeof(u64) - sizeof(int))
+        // not enough space - return
+        goto out;
+
+    goto out;
+
+out:
+    // save number of elements in the array
+    data->submit_p->buf[orig_off & (MAX_PERCPU_BUFSIZE-1)] = elem_num;
+    data->context.argnum++;
+
+    return 1;
 }
 
 static __always_inline int save_str_arr_to_buf(event_data_t *data, const char __user *const __user *ptr, u8 index)
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -302,8 +302,8 @@ Copyright (C) Aqua Security inc.
 #endif
 
 #define IOCTL_FETCH_SYSCALLS            65        // randomly picked number for ioctl cmd
-#define NUMBER_OF_SYSCALLS_X86          18
-#define NUMBER_OF_SYSCALLS_ARM          14
+#define NUMBER_OF_SYSCALLS_TO_CHECK_X86 18
+#define NUMBER_OF_SYSCALLS_TO_CHECK_ARM 14
 
 /*================================ eBPF KCONFIGs =============================*/
 
@@ -1593,10 +1593,7 @@ static __always_inline int save_str_to_buf(event_data_t *data, void *ptr, u8 ind
 }
 
 static __always_inline int save_u64_arr_to_buf(event_data_t *data, const u64 __user *ptr,int len , u8 index){
-
     // Data saved to submit buf: [index][u64 count][u64 1][u64 2][u64 3]...
-
-
     u8 elem_num = 0;
     // Save argument index
     data->submit_p->buf[(data->buf_off) & (MAX_PERCPU_BUFSIZE-1)] = index;
@@ -1629,9 +1626,6 @@ static __always_inline int save_u64_arr_to_buf(event_data_t *data, const u64 __u
             goto out;
         }
     }
-    if (data->buf_off > MAX_PERCPU_BUFSIZE - sizeof(u64) - sizeof(int))
-        // not enough space - return
-        goto out;
 
     goto out;
 
@@ -2276,50 +2270,6 @@ static __always_inline struct pipe_inode_info *get_file_pipe_info(struct file *f
     return pipe;
 }
 
-static __always_inline void fetch_syscalls_addresses(event_data_t *data){
-    int key = 0;
-    u64 *table_ptr = bpf_map_lookup_elem(&syscalls_to_check_map, (void *)&key);
-    if (table_ptr == NULL){
-        return ;
-    }
-
-    unsigned long *syscall_table_addr = (unsigned long*) *table_ptr;
-    u64 idx;
-    u64* syscall_num_p;                  // pointer to syscall_number
-    u64 syscall_num;
-    unsigned long syscall_addr = 0;
-    int monitored_syscalls_amount = 0;
-    #if defined(bpf_target_x86)
-        monitored_syscalls_amount = NUMBER_OF_SYSCALLS_X86;
-        u64 syscall_address[NUMBER_OF_SYSCALLS_X86];
-    #elif defined(bpf_target_arm64)
-        monitored_syscalls_amount = NUMBER_OF_SYSCALLS_ARM;
-        u64 syscall_address[NUMBER_OF_SYSCALLS_ARM];
-    #else
-
-        return
-    #endif
-
-    __builtin_memset(syscall_address, 0, sizeof(syscall_address));
-    // the map should look like [syscall_table][syscall number 1][syscall number 2][syscall number 3]...
-    #pragma unroll
-    for(int i =0; i < monitored_syscalls_amount; i++){
-        idx = i+1;
-        syscall_num_p = bpf_map_lookup_elem(&syscalls_to_check_map, (void *)&idx);
-        if (syscall_num_p == NULL){
-            continue;
-        }
-        syscall_num = (u64)*syscall_num_p;
-        syscall_addr = READ_KERN(syscall_table_addr[syscall_num]);
-        if (syscall_addr == 0){
-            return;
-        }
-        syscall_address[i] = syscall_addr;
-    }
-    save_u64_arr_to_buf(data, (const u64 *)syscall_address, monitored_syscalls_amount, 0);
-    events_perf_submit(data, PRINT_SYSCALL_TABLE, 0);
-}
-
 /*============================== SYSCALL HOOKS ===============================*/
 
 // include/trace/events/syscalls.h:
@@ -2927,6 +2877,54 @@ int BPF_KPROBE(trace_do_exit)
     return events_perf_submit(&data, DO_EXIT, code);
 }
 
+/* invoke_print_syscall_table_event submit to the buff the syscalls function handlers address from the syscall table.
+ * the syscalls are strode in map which is syscalls_to_check_map and the syscall-table address is stored in the kernel_symbols map.
+ */
+static __always_inline void invoke_print_syscall_table_event(event_data_t *data){
+    int key = 0;
+    u64 *table_ptr = bpf_map_lookup_elem(&syscalls_to_check_map, (void *)&key);
+    if (table_ptr == NULL){
+        return ;
+    }
+
+    char syscall_table[15] = "sys_call_table";
+    unsigned long *syscall_table_addr = (unsigned long*) get_symbol_addr(syscall_table);
+    u64 idx;
+    u64* syscall_num_p;                  // pointer to syscall_number
+    u64 syscall_num;
+    unsigned long syscall_addr = 0;
+    int monitored_syscalls_amount = 0;
+    #if defined(bpf_target_x86)
+        monitored_syscalls_amount = NUMBER_OF_SYSCALLS_TO_CHECK_X86;
+        u64 syscall_address[NUMBER_OF_SYSCALLS_TO_CHECK_X86];
+    #elif defined(bpf_target_arm64)
+        monitored_syscalls_amount = NUMBER_OF_SYSCALLS_TO_CHECK_ARM;
+        u64 syscall_address[NUMBER_OF_SYSCALLS_TO_CHECK_ARM];
+    #else
+
+        return
+    #endif
+
+    __builtin_memset(syscall_address, 0, sizeof(syscall_address));
+    // the map should look like [syscall number 1][syscall number 2][syscall number 3]...
+    #pragma unroll
+    for(int i =0; i < monitored_syscalls_amount; i++){
+        idx = i;
+        syscall_num_p = bpf_map_lookup_elem(&syscalls_to_check_map, (void *)&idx);
+        if (syscall_num_p == NULL){
+            continue;
+        }
+        syscall_num = (u64)*syscall_num_p;
+        syscall_addr = READ_KERN(syscall_table_addr[syscall_num]);
+        if (syscall_addr == 0){
+            return;
+        }
+        syscall_address[i] = syscall_addr;
+    }
+    save_u64_arr_to_buf(data, (const u64 *)syscall_address, monitored_syscalls_amount, 0);
+    events_perf_submit(data, PRINT_SYSCALL_TABLE, 0);
+}
+
 SEC("kprobe/security_file_ioctl")
 int BPF_KPROBE(trace_tracee_trigger_event )
 {
@@ -2938,7 +2936,7 @@ int BPF_KPROBE(trace_tracee_trigger_event )
     unsigned int cmd = PT_REGS_PARM2(ctx);
 
     if (cmd == IOCTL_FETCH_SYSCALLS && get_config(CONFIG_TRACEE_PID) == data.context.host_pid){
-        fetch_syscalls_addresses(&data);
+        invoke_print_syscall_table_event(&data);
     }
     return 0;
 }
--- a/pkg/ebpf/events_derived.go
+++ b/pkg/ebpf/events_derived.go
@@ -2,11 +2,11 @@ package ebpf
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/aquasecurity/libbpfgo/helpers"
 	"github.com/aquasecurity/tracee/pkg/bufferdecoder"
 	"github.com/aquasecurity/tracee/types/trace"
-	"strings"
 )
 
 // deriveFn is a function prototype for a function that receives an event as
@@ -25,6 +25,9 @@ func (t *Tracee) initEventDerivationMap() error {
 		},
 		CgroupRmdirEventID: {
 			ContainerRemoveEventID: deriveContainerRemoved(t),
+		},
+		PrintSyscallTableEventID: {
+			DetectHookedSyscallsEventID: deriveDetectHookedSyscall(t),
 		},
 	}
 
@@ -121,12 +124,15 @@ func deriveContainerRemoved(t *Tracee) deriveFn {
 }
 
 func deriveDetectHookedSyscall(t *Tracee) deriveFn {
-	return func(event trace.Event)  (trace.Event, bool, error) {
+	return func(event trace.Event) (trace.Event, bool, error) {
 		syscallsAdresses, err := getEventArgUlongArrVal(&event, "syscalls_addresses")
 		if err != nil {
 			return trace.Event{}, false, fmt.Errorf("error parsing syscalls_numbers arg: %v", err)
 		}
-		hookedSyscallData := t.ParseHookedAddresses(syscallsAdresses)
+		hookedSyscallData, err := analyzeHookedAddresses(syscallsAdresses, t.config.OSInfo, t.kernelSymbols)
+		if err != nil {
+			return trace.Event{}, false, fmt.Errorf("error parsing analyzing hooked syscalls adresses arg: %v", err)
+		}
 		de := event
 		de.EventID = int(DetectHookedSyscallsEventID)
 		de.EventName = "hooked_syscalls"
@@ -139,43 +145,36 @@ func deriveDetectHookedSyscall(t *Tracee) deriveFn {
 	}
 }
 
-func (t *Tracee) ParseHookedAddresses(addresses []uint64) []bufferdecoder.HookedSyscallData {
+func analyzeHookedAddresses(addresses []uint64, OsConfig *helpers.OSInfo, kernelSymbols *helpers.KernelSymbolTable) ([]bufferdecoder.HookedSyscallData, error) {
 	hookedSyscallData := make([]bufferdecoder.HookedSyscallData, 0, 0)
 	for idx, syscallsAdress := range addresses {
-		InTextSegment, err := t.kernelSymbols.TextSegmentContains(syscallsAdress)
+		InTextSegment, err := kernelSymbols.TextSegmentContains(syscallsAdress)
 		if err != nil {
 			continue
 		}
 		if !InTextSegment {
-			hookingFunction := t.ParseSymbol(syscallsAdress)
-			arch := t.config.OsConfig.GetOSReleaseFieldValue(helpers.OS_ARCH)
+			hookingFunction := parseSymbol(syscallsAdress, kernelSymbols)
 			var syscallNumber int32
-			if strings.Compare(arch, "x86_64") == 0 {
-				if idx > len(syscallsToCheckX86) {
-					continue
-				}
-				syscallNumber = int32(syscallsToCheckX86[idx])
-			} else {
-				if idx > len(syscallsToCheckArm) {
-					continue
-				}
-				syscallNumber = int32(syscallsToCheckArm[idx])
+			if idx > len(syscallsToCheck) {
+				return nil, fmt.Errorf("syscall inedx out of the syscalls to check list %v", err)
 			}
+			syscallNumber = int32(syscallsToCheck[idx])
 			event, found := EventsDefinitions[syscallNumber]
-			var hookedSyscall bufferdecoder.HookedSyscallData
+			var hookedSyscallName string
 			if found {
-				hookedSyscall = bufferdecoder.HookedSyscallData{event.Name, hookingFunction.Owner}
+				hookedSyscallName = event.Name
 			} else {
-				hookedSyscall = bufferdecoder.HookedSyscallData{fmt.Sprint(syscallNumber), hookingFunction.Owner}
+				hookedSyscallName = fmt.Sprint(syscallNumber)
 			}
-			hookedSyscallData = append(hookedSyscallData, hookedSyscall)
+			hookedSyscallData = append(hookedSyscallData, bufferdecoder.HookedSyscallData{hookedSyscallName, hookingFunction.Owner})
+
 		}
 	}
-	return hookedSyscallData
+	return hookedSyscallData, nil
 }
 
-func (t *Tracee) ParseSymbol(address uint64) *helpers.KernelSymbol {
-	hookingFunction, err := t.kernelSymbols.GetSymbolByAddr(address)
+func parseSymbol(address uint64, table *helpers.KernelSymbolTable) *helpers.KernelSymbol {
+	hookingFunction, err := table.GetSymbolByAddr(address)
 	if err != nil {
 		hookingFunction = &helpers.KernelSymbol{}
 		hookingFunction.Owner = "hidden"
--- a/pkg/ebpf/events_derived.go
+++ b/pkg/ebpf/events_derived.go
@@ -3,7 +3,10 @@ package ebpf
 import (
 	"fmt"
 
+	"github.com/aquasecurity/libbpfgo/helpers"
+	"github.com/aquasecurity/tracee/pkg/bufferdecoder"
 	"github.com/aquasecurity/tracee/types/trace"
+	"strings"
 )
 
 // deriveFn is a function prototype for a function that receives an event as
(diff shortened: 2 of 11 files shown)


rafaeldtinoco:
I haven't reviewed this yet, but, Itamar, would you mind using the description (or a changed version of it) to your git commit log ?

https://github.com/rafaeldtinoco/tracee/commit/a499519e210b9bd123fa988cd28040f31be2528b

It said:

```
tracee-ebpf: add event to detect hooked syscalls

A special ioctl cmd is executed during tracee initialization. This
special ioctl cmd tells eBPF kprobe function that it should look the
syscalls table for a list of given syscalls and their addresses.

By checking runtime syscall handlers addresses, it is possible to know
if the handlers were hijacked by a kernel module, for example. They way
way it works is this: if the syscall handler address is not in kernel
text segment boundaries, it means it is not part of running kernel
(likely a hook somewhere else).
```

rafaeldtinoco:
You just removed 3rdparty/btfhub.sh file in https://github.com/aquasecurity/tracee/pull/1394/commits/c490c9f48e6078a3d626aaf80c26faf0c47aca85. Mind converting this to draft PR while you're working on it ? I can review it, even being a draft, don't worry.

itamarmaouda101:
> I haven't reviewed this yet, but, Itamar, would you mind using the description (or a changed version of it) to your git commit log ?
> 
> [rafaeldtinoco@a499519](https://github.com/rafaeldtinoco/tracee/commit/a499519e210b9bd123fa988cd28040f31be2528b)
> 
> It said:
> 
> ```
> tracee-ebpf: add event to detect hooked syscalls
> 
> A special ioctl cmd is executed during tracee initialization. This
> special ioctl cmd tells eBPF kprobe function that it should look the
> syscalls table for a list of given syscalls and their addresses.
> 
> By checking runtime syscall handlers addresses, it is possible to know
> if the handlers were hijacked by a kernel module, for example. They way
> way it works is this: if the syscall handler address is not in kernel
> text segment boundaries, it means it is not part of running kernel
> (likely a hook somewhere else).
> ```

sure !

itamarmaouda101:
> You just removed 3rdparty/btfhub.sh file in [c490c9f](https://github.com/aquasecurity/tracee/commit/c490c9f48e6078a3d626aaf80c26faf0c47aca85). Mind converting this to draft PR while you're working on it ? I can review it, even being a draft, don't worry.

Sure no problem (:

tracee-ebpf/tracee/tracee.go
```diff
@@ -1037,10 +1075,23 @@ func (t *Tracee) updateFileSHA() {
 	}
 }
 
-func (t *Tracee) invokeInitNamespacesEvent() {
```
itamarmaouda101, line 1040: @yanivagman 
Because the hooked syscalls detection also start at the user space (by trigger the IOCTL technique ) and the hooked file operations event also will be trigger by that I thought to change `t.invokeInitNamespacesEvent` to `t.invokeManualTriggeredEvents` and there include the whole events that we will start or create in the user space. 
 what you think about it ?

yanivagman: We should keep those decoupled:
1. Trigger ioctl events when required (doesn't have to be during init)
2. Trigger init related "system state probing" - this includes init namespaces event (which is not exactly an event) and other events in the future, such as process tree init state, container init state, etc.

itamarmaouda101: Ok, ​i will roll it back and add function for handling IOCTL driven event.
​I think just for now it will be fine to name it as `invokeSyscallDetection` and later on when the fops detection will be in ill rename it for handling the all IOCTL driven even

tracee-ebpf/tracee/events_processor.go
```diff
@@ -255,6 +261,39 @@ func (t *Tracee) processEvent(event *external.Event) error {
 			return fmt.Errorf("error parsing cgroup_rmdir args: %v", err)
 		}
 		t.containers.CgroupRemove(cgroupId)
+
+	case DetectHookedSyscallsEventID:
+		doOnce.Do(func() {
```
yanivagman, line 266: Why do we want to guard this with sync.Once?
What if we will want to trigger the event in runtime in the future?

tracee-ebpf/tracee/events_processor.go
```diff
@@ -255,6 +261,39 @@ func (t *Tracee) processEvent(event *external.Event) error {
...
+			hookedSyscallNumberMap, err := t.bpfModule.GetMap("hooked_syscalls_map")
+			if err == nil {
```
yanivagman, line 268: To avoid extra indentation and to handle the error:
```
if err != nil {
    return fmt.Errorf("error parsing detect_hooked_syscall args: %v", err)
}
```

tracee-ebpf/tracee/events_processor.go
```diff
@@ -255,6 +261,39 @@ func (t *Tracee) processEvent(event *external.Event) error {
...
+				hookedSyscallData := make([]kernelsyms.HookedSyscallData, 0, 0)
+				i := 0
+				for true {
```
yanivagman, line 271: Instead of looping with unknown bounds, you can pass the number of elements as an argument of the event

tracee-ebpf/tracee/events_processor.go
```diff
@@ -255,6 +261,39 @@ func (t *Tracee) processEvent(event *external.Event) error {
...
+					}
+					syscallNum := binary.LittleEndian.Uint32(data)
+					syscallAddr := binary.LittleEndian.Uint64(data[8:16])
+					hookingFunction, err := t.kernelSymbols.GetSymbolByAddr(syscallAddr)
+					if err != nil {
+						hookingFunction.Owner = "hidden"
+					}
+					hookingFunction.Owner = strings.TrimPrefix(hookingFunction.Owner, "[")
+					hookingFunction.Owner = strings.TrimSuffix(hookingFunction.Owner, "]")
+					event, _ := EventsDefinitions[int32(syscallNum)]
```
yanivagman, line 284: You should check that event exists (syscallNum may be invalid)

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -268,6 +269,8 @@ Copyright (C) Aqua Security inc.
 #define MAX_BIN_CHUNKS                  110
 #endif
 
+#define IOCTL_SYSCALL_HOOK              65
```
yanivagman, line 272: Any reason for using 65?

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2053,6 +2064,98 @@ static __always_inline unsigned short get_inode_mode_from_fd(u64 fd)
...
+            61,         // getdents64
+            280,        // bpf
+            281,        // execveat
+            56,         // openat
+            206,        // sendto
+            207,        // recvfrom
+            211,        // sendmsg
+            212	        // recvmsg
+
+        };
```
yanivagman, line 2129: This information can be configured from userspace (using a bpf map) instead of being hardcoded.
This will allow to extend this feature to other syscalls dynamically.

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2053,6 +2064,98 @@ static __always_inline unsigned short get_inode_mode_from_fd(u64 fd)
...
+            if (saddr == 0)
+                continue;
+
+            // syscall address MUST BE inside kernel text segment boundaries
+
+            if(!((saddr >= stext) && (saddr < etext))) {
+                hooked_syscall_t hooked_syscall = {snum, saddr};
+                bpf_map_update_elem(&hooked_syscalls_map, &j, &hooked_syscall, BPF_ANY);
+                j++;
+            }
```
yanivagman, line 2150: I would take this logic to userspace as anyways you get the stext and etext from there (using kallsysms).
This way we can remove the need for code_boundaries_map and just send the syscalls addresses for ALL the configured syscalls. The check will then be performed in userspace.

Other than that, we can push all the addresses as an integer array argument (we should define a new type for that) and remove the need for hooked_syscalls_map as well. This is better than reading the map in userspace after receiving the event.

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2618,6 +2721,27 @@ int BPF_KPROBE(trace_do_exit)
...
+
+    if (!init_event_data(&data, ctx))
+        return 0;
+
+    if (get_config(CONFIG_TRACEE_PID) != data.context.host_pid)
+        return 0;
+
+    unsigned int cmd = PT_REGS_PARM2(ctx);
+
+
```
yanivagman, line 2737: Please remove extra line

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2618,6 +2721,27 @@ int BPF_KPROBE(trace_do_exit)
...
+    if (cmd == IOCTL_SYSCALL_HOOK){
+        detect_hooked_syscalls(&data);
+    }
+    return 0;
+
```
yanivagman, line 2742: Please remove extra line (or put it before the `return 0`)

tracee-ebpf/tracee/tracee.go
```diff
@@ -651,6 +660,35 @@ func (t *Tracee) populateBPFMaps() error {
...
+			}
+			codeBoundariesMap, err := t.bpfModule.GetMap("code_boundaries_map") // u32, u32
+			if err != nil {
+				return err
+			}
+			errs := make([]error, 0)
+			key := int(0)
+			syscallTableSymbol, err := t.kernelSymbols.GetSymbolByName("system", "sys_call_table")
+			syscallTableAddress := syscallTableSymbol.Address
+			errs = append(errs, syscallTableAddrMap.Update(unsafe.Pointer(&key), unsafe.Pointer(&syscallTableAddress)))
```
yanivagman, line 676: This data can be passed through the already existing config_map.
No need for another map for that

tracee-ebpf/tracee/tracee.go
```diff
@@ -1037,10 +1075,23 @@ func (t *Tracee) updateFileSHA() {
...
+
+	//invoke InitNamespacesEvent
 	if t.eventsToTrace[InitNamespacesEventID] {
 		systemInfoEvent, _ := CreateInitNamespacesEvent()
 		t.config.ChanEvents <- systemInfoEvent
 		t.stats.eventCounter.Increment()
 	}
+	// invoke DetectHookedSyscallsEvent
+	if t.eventsToTrace[DetectHookedSyscallsEventID] {
+		path := "/tmp/tracee/out/tracee.pid"
```
yanivagman, line 1089: output path doesn't have to be /tmp/tracee/out.
Please use `cfg.Capture.OutputPath` instead

yanivagman (changes requested):
Nice feature!
A few changes are required/suggested.
Please also verify that your code works on 4.18/4.19 kernel as well

pkg/kernelsyms/kernel_symbols.go
```diff
@@ -0,0 +1,86 @@
...
+	Name    string
+	Type    string
+	Address uint64
+	Owner   string
+}
+
+type HookedSyscallData struct {
+	SyscallName string
+	ModuleOwner string
+}
```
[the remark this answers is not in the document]
itamarmaouda101: I think `HookedSyscallData` isn't  should be there but the `KernelSymbolTable` and `KernelSymbol` should be because the package  give's object of all the symbols  in the kernel

pkg/kernelsyms/kernel_symbols.go
```diff
@@ -0,0 +1,86 @@
...
+
+const IoctlSyscallHook int = 65
+
+/*
+ * this package is to support a kernel-malware detection and especially rootkits.
```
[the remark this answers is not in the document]
itamarmaouda101: I think ill remove it because now we aren't support a lot of detections.
if in the future we will create a file or package for malware detection ill added it (:

pkg/kernelsyms/kernel_symbols.go
```diff
@@ -0,0 +1,86 @@
...
+
+//GetSymbolByAddr return's a symbol by a given address
+func (t *KernelSymbolTable) GetSymbolByAddr(addr uint64) (KernelSymbol, error) {
+	for _, Symbol := range t.symbolMap {
+		if Symbol.Address == addr {
+			return Symbol, nil
+		}
+	}
+	return KernelSymbol{}, fmt.Errorf("couldent find the symbol")
+}
```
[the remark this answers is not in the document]
itamarmaouda101: Now its called  for every hooked syscall so I think it isn't supposed to to called so frequently because we check less then 20 syscalls and also I don't think any rootkit will hook more then 5 syscalls.
(for example diamorphine which is in if the most popular rootkits in GitHub hooks only 3 syscalls)

tracee-ebpf/tracee/events_processor.go
itamarmaouda101: Because now we only check this event when tracee start.
if in the future we will want to call this event in runtime we need to make some changes that can in clude that removing that guard, for example we will need to change the mechaniqe of the `ioctl trigger event` to something more dynamic (now its sonly triggers it when tracee starts).
if you still think it isn't necessary , ill remove it (:

tracee-ebpf/tracee/events_processor.go
itamarmaouda101: Ok , i changed it to `if err != nil {
				fmt.Errorf("error parsing detect_hooked_syscall args: %v", err)
				return
			}`

tracee-ebpf/tracee/tracee.bpf.c
itamarmaouda101: nope , it just a random number because it just for verification.
I can added a comment about it

tracee-ebpf/tracee/tracee.bpf.c
itamarmaouda101: Sounds good , ill change it that it will just "fetch" the asked syscalls to map and then we will parse it in the user-space.

About the integer array idea, I think its isn't will give us so much more then we already have because we still have to change the argument type tot a struct that will contain also the hooking module/object of of the syscall (see `HookedSyscallData`)

tracee-ebpf/tracee/events_processor.go
```diff
@@ -301,6 +304,44 @@ func (t *Tracee) processEvent(event *external.Event) error {
 			return fmt.Errorf("error parsing cgroup_rmdir args: %v", err)
 		}
 		t.containers.CgroupRemove(cgroupId)
+
+	case DetectHookedSyscallsEventID:
+		syscallsAdresses, _ := getEventArgUlongArrVal(event, "syscalls_numbers")
+		//if err != nil {
+		//	return fmt.Errorf("error parsing syscalls_numbers args: %v", err)
+		//}
+
```
itamarmaouda101, line 313: remove comment

tracee-ebpf/tracee/events_processor.go
```diff
@@ -301,6 +304,44 @@ func (t *Tracee) processEvent(event *external.Event) error {
 			return fmt.Errorf("error parsing cgroup_rmdir args: %v", err)
 		}
 		t.containers.CgroupRemove(cgroupId)
+
+	case DetectHookedSyscallsEventID:
+		syscallsAdresses, _ := getEventArgUlongArrVal(event, "syscalls_numbers")
```
itamarmaouda101, line 309: check the argument exists

tracee-ebpf/tracee/events_processor.go
```diff
@@ -331,6 +372,19 @@ func getEventArgUint64Val(event *external.Event, argName string) (uint64, error)
 	}
 	return 0, fmt.Errorf("argument %s not found", argName)
 }
+func getEventArgUlongArrVal(event *external.Event, argName string) ([]uint64, error) {
+	for _, arg := range event.Args {
+		if arg.Name == argName {
+			val, ok := arg.Value.([]uint64)
+			if !ok {
+				return nil, fmt.Errorf("argument %s is not of type ulong array", argName)
```
itamarmaouda101, line 380: ulong64*

tracee-ebpf/tracee/events_processor.go
```diff
@@ -331,6 +372,19 @@ func getEventArgUint64Val(event *external.Event, argName string) (uint64, error)
 	}
 	return 0, fmt.Errorf("argument %s not found", argName)
 }
+func getEventArgUlongArrVal(event *external.Event, argName string) ([]uint64, error) {
```
itamarmaouda101, line 375: ulong64*

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1441,6 +1448,63 @@ static __always_inline int save_str_to_buf(event_data_t *data, void *ptr, u8 ind
...
+                    bpf_trace_printk(error, sizeof(error), sz);
+            goto out;
+        }
+    }
+    if (data->buf_off > MAX_PERCPU_BUFSIZE - sizeof(u64) - sizeof(int))
+        // not enough space - return
+        goto out;
+     u64 padd = 0xffffffffffffffff;
+     void *addr = &(data->submit_p->buf[data->buf_off + sizeof(int)]);
+     bpf_probe_read(addr, sizeof(u64), &padd);
```
itamarmaouda101, line 1494: remove

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1441,6 +1448,63 @@ static __always_inline int save_str_to_buf(event_data_t *data, void *ptr, u8 ind
...
+
+    goto out;
+
+out:
+    // save number of elements in the array
+    data->submit_p->buf[orig_off & (MAX_PERCPU_BUFSIZE-1)] = elem_num;
+    data->context.argnum++;
+    const char fmt_str[] = "save %d args as u64!\n";
+    bpf_trace_printk(fmt_str, sizeof(fmt_str), elem_num);
```
itamarmaouda101, line 1503: remove

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1441,6 +1448,63 @@ static __always_inline int save_str_to_buf(event_data_t *data, void *ptr, u8 ind
...
+                // Satisfy validator
+                goto out;
+
+            data->buf_off += sizeof(u64);
+            const char fmt_str[] = "read arg idx: %d val is %lx\n";
+            bpf_trace_printk(fmt_str, sizeof(fmt_str), i, element);
+            continue;
+        } else {
+              const char error[] = "error!!!!!!!! %d\n";
+                    bpf_trace_printk(error, sizeof(error), sz);
```
itamarmaouda101, line 1485: remove

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -1441,6 +1448,63 @@ static __always_inline int save_str_to_buf(event_data_t *data, void *ptr, u8 ind
...
+        int sz = bpf_probe_read(addr, sizeof(u64), (void *)&element);
+        if (sz == 0) {
+            elem_num++;
+            if (data->buf_off > MAX_PERCPU_BUFSIZE )
+                // Satisfy validator
+                goto out;
+
+            data->buf_off += sizeof(u64);
+            const char fmt_str[] = "read arg idx: %d val is %lx\n";
+            bpf_trace_printk(fmt_str, sizeof(fmt_str), i, element);
```
itamarmaouda101, line 1481: remove

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2066,6 +2130,67 @@ static __always_inline unsigned short get_inode_mode_from_fd(u64 fd)
...
+    int number_of_hooked_syscalls = 0;
+    #if defined(bpf_target_x86)
+        number_of_hooked_syscalls = NUMBER_OF_SYSCALLS_X86;
+        u64 syscall_address[NUMBER_OF_SYSCALLS_X86];
+    #elif defined(bpf_target_arm64)
+        number_of_hooked_syscalls = NUMBER_OF_SYSCALLS_ARM;
+        u64 syscall_address[NUMBER_OF_SYSCALLS_ARM];
+    #else
+        const char err[] = "errorrrrr\n";
+        bpf_trace_printk(err, sizeof(err));
```
itamarmaouda101, line 2156: remove

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2066,6 +2130,67 @@ static __always_inline unsigned short get_inode_mode_from_fd(u64 fd)
...
+        return
+    #endif
+        const char beforeloop[] = "beforeloop\n";
+        bpf_trace_printk(beforeloop, sizeof(beforeloop));
```
itamarmaouda101, line 2160: remove

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2066,6 +2130,67 @@ static __always_inline unsigned short get_inode_mode_from_fd(u64 fd)
...
+    __builtin_memset(syscall_address, 0, sizeof(syscall_address));
+    #pragma unroll
+    for(int i =0; i < number_of_hooked_syscalls; i++){
+        idx = i+2;
+        snum_ptr = bpf_map_lookup_elem(&hooked_syscalls_map, (void *)&idx);
+        if (snum_ptr == NULL){
+            const char exit1[] = "exit1\n";
+            bpf_trace_printk(exit1, sizeof(exit1));
```
itamarmaouda101, line 2168: remove

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2066,6 +2130,67 @@ static __always_inline unsigned short get_inode_mode_from_fd(u64 fd)
...
+            return;
+        }
+        snum =(u64) *snum_ptr;
+        const char n[] = "sysccall number is %d\n";
+        bpf_trace_printk(n, sizeof(n), snum);
```
itamarmaouda101, line 2173: remove

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -2066,6 +2130,67 @@ static __always_inline unsigned short get_inode_mode_from_fd(u64 fd)
...
+        saddr = READ_KERN(syscall_table_addr[snum]);
+        if (saddr == 0){
+            const char exit2[] = "exit2\n";
+            bpf_trace_printk(exit2, sizeof(exit2));
+            return;
+        }
+//        bpf_map_update_elem(&hooked_syscalls_map, &key, &fill_zero, BPF_ANY);
+//        bpf_map_update_elem(&hooked_syscalls_map, &snum, &saddr, BPF_ANY);
```
itamarmaouda101, line 2181: remove

pkg/ebpf/events_processor.go
```diff
@@ -331,6 +367,19 @@ func getEventArgUint64Val(event *external.Event, argName string) (uint64, error)
 	}
 	return 0, fmt.Errorf("argument %s not found", argName)
 }
+func getEventArgUlongArrVal(event *external.Event, argName string) ([]uint64, error) {
+	for _, arg := range event.Args {
+		if arg.Name == argName {
+			val, ok := arg.Value.([]uint64)
+			if !ok {
+				return nil, fmt.Errorf("argument %s is not of type ulong array", argName)
+				fmt.Println("val is ", val)
```
itamarmaouda101, line 376: please remove debug print

rafaeldtinoco (commented):
leaving some comments, I see this is unfinished so I'll take a new look once you propose it.

pkg/ebpf/tracee.go
```diff
@@ -732,11 +740,83 @@ func (t *Tracee) populateBPFMaps() error {
...
+		if e == DetectHookedSyscallsEventID {
+			hookedSyscallsMap, err := t.bpfModule.GetMap("hooked_syscalls_map")
+			if err != nil {
+				return err
+			}
+			key := int(1)
+			syscallTableSymbol, err := t.kernelSymbols.GetSymbolByName("system", "sys_call_table")
+			syscallTableAddress := syscallTableSymbol.Address
+			errs = append(errs, hookedSyscallsMap.Update(unsafe.Pointer(&key), unsafe.Pointer(&syscallTableAddress)))
+			OSInfo, err := helpers.GetOSInfo()
```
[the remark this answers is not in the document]
itamarmaouda101: It's used also in main, and I saw that in the same function we are initialize the Tracee object, so I added to the `config` member the Os.info as `OsConfig`.

then we save one call to the `helpers.GetOsInfo`

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1485,6 +1492,54 @@ static __always_inline int save_str_to_buf(event_data_t *data, void *ptr, u8 ind
     return 0;
 }
 
+static __always_inline int save_u64_arr_to_buf(event_data_t *data, const u64 __user *ptr,int len , u8 index){
```
itamarmaouda101, line 1495: add DOCO

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2110,6 +2165,50 @@ static __always_inline unsigned short get_inode_mode_from_fd(u64 fd)
...
+            return;
+        }
+        snum =(u64) *snum_ptr;
+        saddr = READ_KERN(syscall_table_addr[snum]);
+        if (saddr == 0){
+            return;
+        }
+        syscall_address[i] = saddr;
+    }
+    int syscall_address_len = sizeof(syscall_address) / sizeof(u64);
```
AlonZivony, line 2207: isn't it `number_of_hooked_syscalls`?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2110,6 +2165,50 @@ static __always_inline unsigned short get_inode_mode_from_fd(u64 fd)
     return READ_KERN(f_inode->i_mode);
 }
 
+static __always_inline void detect_hooked_syscalls(event_data_t *data){
```
AlonZivony, line 2168: This function defiantly not detecting hooked syscalls, so this name is incorrect.
I suggest calling it `invoke_syscalls_addresses_event` which describes better the purpose here.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -190,7 +191,8 @@ Copyright (C) Aqua Security inc.
 #define SECURITY_INODE_SYMLINK          1031
 #define SOCKET_DUP                      1032
 #define HIDDEN_INODES                   1033
-#define MAX_EVENT_ID                    1034
+#define DETECT_HOOKED_SYSCALLS          1034
```
AlonZivony, line 194: The event (at least in the BPF part) isn't related to hooks, so you should change the name to suit its purpose better.
Maybe `SYSCALLS_INFO`? Think on a name that you think match the purpose here.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2693,6 +2792,22 @@ int BPF_KPROBE(trace_do_exit)
     return events_perf_submit(&data, DO_EXIT, code);
 }
 
+SEC("kprobe/security_file_ioctl")
+int BPF_KPROBE(trace_security_file_ioctl)
```
AlonZivony, line 2796: The name of the program is mutable, so I think you can call it `trace_tracee_trigger_event` or something similar.

pkg/ebpf/events_processor.go
```diff
@@ -329,6 +330,42 @@ func (t *Tracee) processEvent(event *trace.Event) error {
 			return fmt.Errorf("error parsing cgroup_rmdir args: %v", err)
 		}
 		t.containers.CgroupRemove(cgroupId)
+
+	case DetectHookedSyscallsEventID:
```
AlonZivony, line 334: The current name is problematic for the BPF name, but it does suit the user-mode event.
What I think you should do is to create 2 events - 1 for syscalls addresses and 1 for hooked syscalls.
I think it will match the logic of "derived events", because it is an event that is derived from another event.

pkg/ebpf/events_processor.go
```diff
@@ -423,3 +473,8 @@ func CopyFileByPath(src, dst string) error {
 	}
 	return nil
 }
+
+type HookedSyscallData struct {
```
AlonZivony, line 477: Shouldn't it be higher in the file? Maybe in another file (so it would be exported)?

pkg/ebpf/tracee.go
```diff
@@ -677,11 +685,83 @@ func (t *Tracee) populateBPFMaps() error {
...
+					return err
+				}
+			}
+		}
 	}
 
 	return nil
 }
 
+var syscallsTocheckX86 = []int{
```
AlonZivony, line 723:
```suggestion
var syscallsToCheckX86 = []int{
```

pkg/ebpf/tracee.go
```diff
@@ -677,11 +685,83 @@ func (t *Tracee) populateBPFMaps() error {
...
+	78,  // getdents
+	101, // ptrace
+	217, // getdents64
+	321, // bpf
+	322, // execveat
+	16,  // ioctl
+	257, // openat
+}
+
+var syscallsTocheckArm = []int{
```
AlonZivony, line 744:
```suggestion
var syscallsToCheckArm = []int{
```

pkg/ebpf/tracee.go
```diff
@@ -677,11 +685,83 @@ func (t *Tracee) populateBPFMaps() error {
...
+			errs = append(errs, hookedSyscallsMap.Update(unsafe.Pointer(&key), unsafe.Pointer(&syscallTableAddress)))
+			arch := t.config.OsConfig.GetOSReleaseFieldValue(helpers.OS_ARCH)
+			var syscallList []int
+			if strings.Compare(arch, "x86_64") == 0 {
+				syscallList = syscallsTocheckX86
+			} else {
+				syscallList = syscallsTocheckArm
+			}
+			for idx, val := range syscallList {
+				key = idx + 2
```
AlonZivony, line 709: Again, please explain why +2

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2110,6 +2165,50 @@ static __always_inline unsigned short get_inode_mode_from_fd(u64 fd)
...
+        return
+    #endif
+
+    __builtin_memset(syscall_address, 0, sizeof(syscall_address));
+    #pragma unroll
+    for(int i =0; i < number_of_hooked_syscalls; i++){
+        idx = i+2;
+        snum_ptr = bpf_map_lookup_elem(&hooked_syscalls_map, (void *)&idx);
+        if (snum_ptr == NULL){
+            return;
```
AlonZivony, line 2198: Why do you return here?
```suggestion
            continue;
```

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -585,6 +591,7 @@ BPF_HASH(sys_32_to_64_map, u32, u32);                   // map 32bit to 64bit sy
 BPF_HASH(params_types_map, u32, u64);                   // encoded parameters types for event
 BPF_HASH(process_tree_map, u32, u32);                   // filter events by the ancestry of the traced process
 BPF_HASH(process_context_map, u32, process_context_t);  // holds the process_context data for every tid
+BPF_HASH(hooked_syscalls_map, int, u64);                // syscalls to discover
```
AlonZivony, line 594: Think on a better name for the map.
Maybe `syscalls_in_hook_danger_map`

AlonZivony (changes requested):
Nice work!
Look at the comments, there are some naming to improve and some structure to change.
Also, create a commit for the Decoder apart from the new event. Maybe even do a commit for the syscalls addresses extraction and another for the hooked syscalls event.

pkg/ebpf/events_derived.go
```diff
@@ -69,6 +71,64 @@ func (t *Tracee) deriveEvent(event trace.Event) []trace.Event {
...
+			}
+			hookedSyscallData := make([]HookedSyscallData, 0, 0)
+			for idx, syscallsAdress := range syscallsAdresses {
+				InTextSegment, err := t.kernelSymbols.TextSegmentContains(syscallsAdress)
+				if err != nil {
+					fmt.Errorf(err.Error())
+					continue
+				}
+				if !InTextSegment {
+					hookingFunction, err := t.kernelSymbols.GetSymbolByAddr(syscallsAdress)
```
AlonZivony, line 88: I think you should export the building of the `HookedSyscallData` to a sub-function

pkg/ebpf/events_derived.go
```diff
@@ -69,6 +71,64 @@ func (t *Tracee) deriveEvent(event trace.Event) []trace.Event {
...
+						hookingFunction = &helpers.KernelSymbol{}
+						hookingFunction.Owner = "hidden"
+					}
+					hookingFunction.Owner = strings.TrimPrefix(hookingFunction.Owner, "[")
+					hookingFunction.Owner = strings.TrimSuffix(hookingFunction.Owner, "]")
+					arch := t.config.OsConfig.GetOSReleaseFieldValue(helpers.OS_ARCH)
+					var syscallNumber int32
+					if strings.Compare(arch, "x86_64") == 0 {
+						if idx < len(syscallsToCheckX86) {
+							syscallNumber = int32(syscallsToCheckX86[idx])
```
AlonZivony, line 99: What happens if `idx` is bigger than your list?

pkg/ebpf/tracee.go
```diff
@@ -1088,3 +1170,17 @@ func (t *Tracee) invokeInitEvents() {
 		}
 	}
 }
+
+const IoctlSyscallHook int = 65 // randomly picked number for ioctl cmd
```
AlonZivony, line 1174: Again, pick a name that shows this is a cmd name

pkg/ebpf/tracee.go
```diff
@@ -1088,3 +1170,17 @@ func (t *Tracee) invokeInitEvents() {
...
+
+func (t *Tracee) invokeIoctlTriggeredEvents() {
+	// invoke DetectHookedSyscallsEvent
+	if t.eventsToTrace[PrintSyscallTableEventID] || t.eventsToTrace[DetectHookedSyscallsEventID] {
+		ptmx, err := os.OpenFile(t.config.Capture.OutputPath, os.O_RDONLY, 444)
+		if err != nil {
+			return
```
AlonZivony, line 1181: Shouldn't you return error?

pkg/ebpf/tracee.go
```diff
@@ -1088,3 +1170,17 @@ func (t *Tracee) invokeInitEvents() {
...
+		}
+		syscall.Syscall(syscall.SYS_IOCTL, ptmx.Fd(), uintptr(IoctlSyscallHook), 0)
```
AlonZivony, line 1183: Shouldn't you check if you succeed?

pkg/ebpf/tracee.go
```diff
@@ -689,11 +697,84 @@ func (t *Tracee) populateBPFMaps() error {
 				return err
 			}
 		}
+		if e == PrintSyscallTableEventID {
```
AlonZivony, line 700: There is too much code here.
Find a way to export some of it to a sub-function.

AlonZivony (changes requested):
I liked the changes.
I still have some comments here.
I think you should also add tests for at least some of your code in uesrspace (like the array parsing, etc.)

pkg/ebpf/tracee.go
itamarmaouda101: I exported the `syscallList` check to a sub-function, it still needs to be reduced, after your PR will merge, ill change the place of the 'sys_call_table' address to your map. then its should be fine (:

pkg/ebpf/tracee.go
itamarmaouda101: Actually no, because we aren't really "talking" to char device (the IOCTL is triggering `/tmp/tracee.pid`), so every time we will try to send that ioctl we will get error that `/tmp/tracee.pid` isnt  char device (but the ioctl event will be triggered)

itamarmaouda101: Btw 
This is why in the previous comment, I didn't return error.

pkg/ebpf/events_derived.go
```diff
@@ -116,3 +122,68 @@ func deriveContainerRemoved(t *Tracee) deriveFn {
...
+			} else {
+				hookedSyscall = bufferdecoder.HookedSyscallData{fmt.Sprint(syscallNumber), hookingFunction.Owner}
+			}
+			hookedSyscallData = append(hookedSyscallData, hookedSyscall)
+		}
+	}
+	return hookedSyscallData
+}
+
+func (t *Tracee) ParseSymbol(address uint64) *helpers.KernelSymbol {
```
AlonZivony, line 180: This function shouldn't be a member of tracee, it should only receive the `t.kernelSymbols` as argument.

pkg/ebpf/events_derived.go
```diff
@@ -116,3 +122,68 @@ func deriveContainerRemoved(t *Tracee) deriveFn {
...
+		de.ReturnValue = 0
+		de.Args = []trace.Argument{
+			{ArgMeta: trace.ArgMeta{Name: "hooked_syscalls", Type: "hookedSyscallData[]"}, Value: hookedSyscallData},
+		}
+		de.ArgsNum = 1
+		return de, true, nil
+	}
+}
+
+func (t *Tracee) ParseHookedAddresses(addresses []uint64) []bufferdecoder.HookedSyscallData {
```
AlonZivony, line 145: This function shouln't be part of tracee, it should receive `t.config.OsConfig` and `t.kernelSymbols` as arguments

AlonZivony, line 145: I think the word _parse_ does not fit here well.
_Parse_ means that you take raw data and put it in the appropriate format.
I think the word _analyze_ will match better the case, though you might prefer something else.

pkg/ebpf/events_derived.go
```diff
@@ -116,3 +122,68 @@ func deriveContainerRemoved(t *Tracee) deriveFn {
...
+	for idx, syscallsAdress := range addresses {
+		InTextSegment, err := t.kernelSymbols.TextSegmentContains(syscallsAdress)
+		if err != nil {
+			continue
+		}
+		if !InTextSegment {
+			hookingFunction := t.ParseSymbol(syscallsAdress)
+			arch := t.config.OsConfig.GetOSReleaseFieldValue(helpers.OS_ARCH)
+			var syscallNumber int32
+			if strings.Compare(arch, "x86_64") == 0 {
```
AlonZivony, line 156: Suggestion - you can check it outside the loop and initialize a map according to it and prevent all this block.

pkg/ebpf/events_derived.go
```diff
@@ -116,3 +122,68 @@ func deriveContainerRemoved(t *Tracee) deriveFn {
...
+			} else {
+				if idx > len(syscallsToCheckArm) {
+					continue
+				}
+				syscallNumber = int32(syscallsToCheckArm[idx])
+			}
+			event, found := EventsDefinitions[syscallNumber]
+			var hookedSyscall bufferdecoder.HookedSyscallData
+			if found {
+				hookedSyscall = bufferdecoder.HookedSyscallData{event.Name, hookingFunction.Owner}
```
AlonZivony, line 170: The best practice would be to initialize `hookedSyscall` after the if-else statement, and just initialize the event name according according to if the event was found or not.
You want that only the parts that defer from one another will be in the if-else scope to avoid changes in one branch but not in the other.

pkg/ebpf/events_processor.go
```diff
@@ -333,6 +333,9 @@ func (t *Tracee) processEvent(event *trace.Event) error {
 			return fmt.Errorf("error parsing cgroup_mkdir args: %w", err)
 		}
 		t.containers.CgroupRemove(cgroupId, hId)
+
+	case FinitModuleEventID:
```
AlonZivony, line 337: Document this case

AlonZivony (changes requested):
Small fixes, mainly order and documentation.
Good work.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2732,6 +2834,22 @@ int BPF_KPROBE(trace_do_exit)
...
+int BPF_KPROBE(trace_tracee_trigger_event )
+{
+    event_data_t data = {};
+
+    if (!init_event_data(&data, ctx))
+        return 0;
+
+    unsigned int cmd = PT_REGS_PARM2(ctx);
+
+    if (cmd == IOCTL_FETCH_SYSCALLS && get_config(CONFIG_TRACEE_PID) == data.context.host_pid){
```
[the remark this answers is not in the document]
itamarmaouda101: I don't think so, because we submit an event only if it's driven by Tracee and its IOCTL command.

This means that only Tracee can submit this event in a specific scenario.

AlonZivony:
LGTM!

AlonZivony (approved):
LGTM!

pkg/ebpf/events_derived.go
```diff
@@ -116,3 +122,68 @@ func deriveContainerRemoved(t *Tracee) deriveFn {
...
+		InTextSegment, err := kernelSymbols.TextSegmentContains(syscallsAdress)
+		if err != nil {
+			continue
+		}
+		if !InTextSegment {
+			hookingFunction := parseSymbol(syscallsAdress, kernelSymbols)
+			var syscallNumber int32
+			if idx > len(syscallsToCheck) {
+				continue
+			}
```
yanivagman, line 164: Isn't it a bug if that happens?
If so, we should emit an error and break instead

pkg/ebpf/events_processor.go
```diff
@@ -333,6 +333,10 @@ func (t *Tracee) processEvent(event *trace.Event) error {
 			return fmt.Errorf("error parsing cgroup_mkdir args: %w", err)
 		}
 		t.containers.CgroupRemove(cgroupId, hId)
+
+	// in case FinitModule occurs it means that kernel module loaded, then we will want to check if it hooked the syscall table
+	case FinitModuleEventID:
```
yanivagman, line 338: InitModule as well?

pkg/ebpf/tracee.go
```diff
@@ -783,6 +784,29 @@ func (t *Tracee) populateBPFMaps() error {
...
```
yanivagman, line 787: This loop is for tail-calls/params map initialization (which is generic for all events).
Let's handle this outside of this loop (can be done before)

pkg/ebpf/tracee.go
```diff
@@ -783,6 +784,29 @@ func (t *Tracee) populateBPFMaps() error {
...
+			hookedSyscallsMap, err := t.bpfModule.GetMap("syscalls_to_check_map")
+			if err != nil {
+				return err
+			}
+			/* the hookedSyscallsMap store first the syscall table address then the syscall numbers, like that:
+			 * [sys_call_table symbol address][syscall num #1][syscall num #2][syscall num #3]...
+			 * with that, we can fetch the syscall address by accessing the syscall table in the syscall number index
```
yanivagman, line 794: This comment is not true anymore

pkg/ebpf/tracee.go
```diff
@@ -783,6 +784,29 @@ func (t *Tracee) populateBPFMaps() error {
...
+			 */
+			key := int(0)
```
yanivagman, line 796: I don't see a use for key here (but inside the next loop)

pkg/ebpf/tracee.go
```diff
@@ -783,6 +784,29 @@ func (t *Tracee) populateBPFMaps() error {
...
+			// the map should look like [syscall number 1][syscall number 2][syscall number 3]...
+			for idx, val := range syscallList {
+				key = idx
+				errs = append(errs, hookedSyscallsMap.Update(unsafe.Pointer(&(key)), unsafe.Pointer(&val)))
+			}
+			for _, err := range errs {
+				if err != nil {
+					return err
+				}
+			}
```
yanivagman, line 808: I don't think you need this errs map as you're in a loop, and can break anytime in case of an error.
Just do:
```
err = hookedSyscallsMap.Update(unsafe.Pointer(&(key)), unsafe.Pointer(&val))
if err != nil {
	return err
}
```

pkg/ebpf/tracee.go
```diff
@@ -802,6 +826,55 @@ func (t *Tracee) populateBPFMaps() error {
...
+	47,  // recvmsg
+	59,  // execve
+	62,  // kill
+	78,  // getdents
+	101, // ptrace
+	217, // getdents64
+	321, // bpf
+	322, // execveat
+	16,  // ioctl
+	257, // openat
```
yanivagman, line 847: Better to keep these sorted in the list (and sort arm64 list as well)

pkg/ebpf/tracee.go
```diff
@@ -802,6 +826,55 @@ func (t *Tracee) populateBPFMaps() error {
...
+	211, // sendmsg
+	212, // recvmsg
+}
+
+func getSyscallList(arch string) []int {
+	if strings.Compare(arch, "x86_64") == 0 {
+		return syscallsToCheckX86
+	}
+	return syscallsToCheckArm
+}
```
yanivagman, line 876: instead of having such a function, you can put the above two lists in a file that is conditionaly compiled for a given architecture, and use the same variable name for the two tables. For example, you can simply put the tables under events_amd64.go and events_amd64.go.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2111,6 +2169,50 @@ static __always_inline unsigned short get_inode_mode_from_fd(u64 fd)
...
+    #else
+
+        return
+    #endif
+
+    __builtin_memset(syscall_address, 0, sizeof(syscall_address));
+    // the map should look like [syscall_table][syscall number 1][syscall number 2][syscall number 3]...
+    #pragma unroll
+    for(int i =0; i < number_of_hooked_syscalls; i++){
+        idx = i+1;
```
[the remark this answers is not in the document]
itamarmaouda101: yep, the PR was Alon's loaded kallsyms_map that you merged already.
BTW, those last 3 comments are because I didn't delete the old function, the new `invoke_print_syscall_table_event ` has all the fixes (:

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -298,6 +300,10 @@ Copyright (C) Aqua Security inc.
 #define MAX_BIN_CHUNKS                  110
 #endif
 
+#define IOCTL_FETCH_SYSCALLS            65        // randomly picked number for ioctl cmd
+#define NUMBER_OF_SYSCALLS_X86          18
+#define NUMBER_OF_SYSCALLS_ARM          14
```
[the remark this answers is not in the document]
itamarmaouda101: those are the number of syscalls that we check for each arch, i  rename it to `NUMBER_OF_SYSCALLS_TO_CHECK_ARM` and `NUMBER_OF_SYSCALLS_TO_CHECK_X86`

yanivagman (approved):

chriskaliX:
awesome!

Issue #1369, opened 2022-01-20
Makefile.one tests on pkg


mtcherni95:
Hi, I was trying to run the tests under the `pkg/ `directory (e.g.`/pkg/capabilities/capabilities_test.go`). However, I see then when I run `make -f Makefile.one test` those tests are not triggered. If I am wrong then please correct me - otherwise, shouldn't we add it to Makefile.one? Now, those tests will be probably relevant to both tracee-ebpf and tracee-rules, so I am not sure under which target they should be. Maybe add a new target `pkg`?

rafaeldtinoco:
@michaelkatch 

That happened already (because of the dir structure refactoring) and I've added refactored tests to be tested to the target it was tested before.

Example:

```
.PHONY: test-tracee-ebpf
test-tracee-ebpf: \
        .checkver_$(CMD_GO) \
        $(OUTPUT_DIR)/tracee.bpf.core.o
#
        $(MAKE) $(OUTPUT_DIR)/btfhub
        $(GO_ENV_EBPF) $(CMD_GO) test \
                -tags $(GO_TAGS_EBPF) \
                -v \
                ./tracee-ebpf/... \
                ./cmd/tracee-ebpf/... <= HERE
```

I'd rather not to have more build targets in Makefile.one unless we're building something. Example: tracee-ebpf has its own binary, tracee-rules has its own binary, rules has builtin.so.

michaelkatch:
@rafaeldtinoco you got the wrong michael :) 
@mtcherni95

rafaeldtinoco:
I did not, read it right. :o) (sorry ?)

danielpacak:
You are right @mtcherni95 that tests in the ./pkg folder are not included when we run test-tracee-ebpf nor test-tracee-rules Make targets. Because `pkg/containers/container.go` refers to libbpfgo (and therefore requires CGO to compile) the easiest (and only solutions that would work now) is to add the `./pkg/...` to the test-tracee-ebpf target.

```
GOOS=linux CC=clang GOARCH=amd64 CGO_CFLAGS="-I/vagrant/dist/libbpf" CGO_LDFLAGS="-lelf -lz /vagrant/dist/libbpf/libbpf.a" go test \
	-tags core,ebpf \
	-v \
	./tracee-ebpf/... \
	./cmd/tracee-ebpf/... \
	./pkg/...
```

[...]

PR #1395, opened 2022-01-24
chore: run tests in the pkg subdirectories

danielpacak:
We already have code under pkg that requires CGO flags
to compile and run. Therefore, we run them in the
test-tracee-ebpf target.

Resolves: #1369


rafaeldtinoco (approved):
LGTM. I agree that if test needs CGO then it should be placed there (like argued in the issue).

PR #1397, opened 2022-01-25
tracee-ebpf: introduce container lifecycle events

### LINKED ISSUE #676: Add container start/stop events

yanivagman:
To support higher level events, add a mechanism to derive events from
other events.

Introduce two new higher level events:
1. container_create - derived from cgroup_mkdir
2. container_remove - derived from cgroup_rmdir

Close #676


tracee-ebpf/tracee/events_derived.go
```diff
@@ -0,0 +1,92 @@
...
+			if info := t.containers.GetCgroupInfo(cgroupId); info.ContainerId != "" {
+				de := event
+				de.EventID = int(ContainerCreateEventID)
+				de.EventName = "container_create"
+				de.ReturnValue = 0
+				de.StackAddresses = make([]uint64, 1)
+				de.Args = []external.Argument{
+					{ArgMeta: def.Params[0], Value: info.Runtime},
+					{ArgMeta: def.Params[1], Value: info.ContainerId},
+					{ArgMeta: def.Params[2], Value: info.Ctime},
```
roikol, line 33: i think it might be better that ctime will be of a numeric type (int64 probably), so we'll be able to use it later without convertions. here `info.Ctime` is `time.Time`.

tracee-ebpf/tracee/events_derived.go
```diff
@@ -0,0 +1,92 @@
...
+				derivatives = append(derivatives, de)
+			}
+
+			return nil
+		}
+		deriveFns = map[int32]deriveFn{
+			ContainerCreateEventID: containerCreateFn,
+		}
+	case CgroupRmdirEventID:
+		containerRemoveFn := func(def EventDefinition) error {
```
roikol, line 46: i think these inner funcs could be defined in the outer scope, so the `deriveEvent()` would be smaller - as i think we'll be adding more higher level events in the future. but it's your call :)

tracee-ebpf/tracee/events_pipeline.go
```diff
@@ -165,24 +165,34 @@ func (t *Tracee) processEvents(done <-chan struct{}, in <-chan *external.Event)
 				continue
 			}
 
+			// Derive event before parsing its arguments
+			derivatives := t.deriveEvent(*event)
```
roikol, line 169: just a question - why passing as value and not by reference?

yanivagman: deriveEvent shuold alter the original event.
In addition, there is no performance gain in golang to pass by pointer (even the opposite on small variables)

tracee-ebpf/tracee/events_derived.go
yanivagman: I thought about it too, but having it this way, we're saving passing the original event as an argument. In addition, the interface for `deriveFn` would have to be changed to return the derived event, which will have to add a check that the event is not empty (in the case that there is no error, but also no event has been derived, e.g. when `containerID == ""`)

rafaeldtinoco: Yep, I thought the same and then saw that Yaniv was using event from the outer function.

rafaeldtinoco (approved):
LGTM, seems to be working, the logic looks code.

Only missing thing I noticed was that printList() doesn't include those new events (shouldn't it ?). Likely because of the event ID numbering for user generated events ? needs checking.

I'm approving this, feel free to include a new commit adding it to printList() if you want.

Issue #1400, opened 2022-01-25
process type mismatch and storing bug in process context 


itamarmaouda101:
Hi 
I saw with @roikol  a bug in the #1300.
first is with the type of the arguments in the sched_process_fork_event 
and also in the assignment in the sched_process_fork_event to the process- tree map.

PR #1401, opened 2022-01-25
fixed type mismatch with the processCtx

itamarmaouda101:
Hi this PR solves #1400 by creating a new helper function that fix the type bug and change the process context assignment in the sched_process_fork_event and sched_process_exec_event

### DIFF
--- a/tracee-ebpf/tracee/events_processor.go
+++ b/tracee-ebpf/tracee/events_processor.go
@@ -216,7 +216,7 @@ func (t *Tracee) processEvent(event *external.Event) error {
 			return err
 		}
 	case SchedProcessExitEventID:
-		delete(t.processTree.processTreeMap, event.HostThreadID)
+		delete(t.processTree.processTreeMap, event.HostProcessID)
 	case SchedProcessForkEventID:
 		hostTid, _ := getEventArgInt32Val(event, "child_tid")
 		hostPid, _ := getEventArgInt32Val(event, "child_pid")
--- a/tracee-ebpf/tracee/events_processor.go
+++ b/tracee-ebpf/tracee/events_processor.go
@@ -124,7 +124,7 @@ func (t *Tracee) processEvent(event *external.Event) error {
 
 	case SchedProcessExecEventID:
 		//update the process tree
-		processData := ProcessCtx{event.Timestamp, event.ContainerID, uint32(event.ProcessID), uint32(event.ThreadID), uint32(event.ParentProcessID), uint32(event.HostProcessID), uint32(event.HostThreadID), uint32(event.HostParentProcessID), uint32(event.UserID), uint32(event.MountNS), uint32(event.PIDNS)}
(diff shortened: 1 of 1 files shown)


tracee-ebpf/tracee/events_processor.go
```diff
@@ -216,16 +216,16 @@ func (t *Tracee) processEvent(event *external.Event) error {
...
-		childTid, _ := getEventArgUint32Val(event, "child_ns_tid")
-		processData := ProcessCtx{event.Timestamp, event.ContainerID, pid, childPid, ppid, hostTid, hostTid, hostPpid, uint32(event.UserID), uint32(event.MountNS), uint32(event.ProcessID)}
-		t.processTree.processTreeMap[int(childTid)] = processData
+		hostTid, _ := getEventArgInt32Val(event, "child_tid")
+		hostPid, _ := getEventArgInt32Val(event, "child_pid")
+		pid, _ := getEventArgInt32Val(event, "child_ns_pid")
+		ppid, _ := getEventArgInt32Val(event, "parent_ns_pid")
+		hostPpid, _ := getEventArgInt32Val(event, "parent_pid")
+		tid, _ := getEventArgInt32Val(event, "child_ns_tid")
+		processData := ProcessCtx{event.Timestamp, event.ContainerID, uint32(pid), uint32(tid), uint32(ppid), uint32(hostTid), uint32(hostPid), uint32(hostPpid), uint32(event.UserID), uint32(event.MountNS), uint32(event.PIDNS)}
```
yanivagman, line 227: Although the original type of the argument is an int, there is no meaning to get the int32 value, then convert to uint32, as the size of the variable is 32 bits in both (so if theoretically we had a negative value, the conversion won't take it into account). so, you can just use getEventArgUint32Val like you did before and avoid this extra conversion

itamarmaouda101: I agree, but I talked with @roikol and he says that getEventArgUint32Val return's error to him because it is defined as int

roikol: the `getEventArgUint32Val()` func tries to assing the value as uint32: `val, ok := arg.Value.(uint32)`. it fails here because it is of type int32.
what we can do is use the new func intruduced here, but return a converted value: return `uint32(val)` or `return int(val)`.

itamarmaouda101:
I added a sleep on `sched_process_delete `to prevent case of race condition in case of receiving packet  and process exit on the same time (the map will be empty)

tracee-ebpf/tracee/events_processor.go
```diff
@@ -216,16 +217,17 @@ func (t *Tracee) processEvent(event *external.Event) error {
 			return err
 		}
 	case SchedProcessExitEventID:
-		delete(t.processTree.processTreeMap, event.HostProcessID)
+		time.Sleep(time.Second)
```
yanivagman, line 220: Please don't do that.
It blocks the processor thread for one second, thus no other events can be processed during this time.

yanivagman (approved):

Issue #1396, opened 2022-01-24
/sys/kernel/security/lockdown has to be bind mounted to container so tracee-ebpf is executed


rafaeldtinoco:
After commit a5bc030, tracee-make stopped being able to run tracee-ebpf as it does not bind mount "/sys/kernel/security" to the container:

```
$ sudo ./dist/tracee-ebpf --debug
2022/01/24 21:08:19 open /sys/kernel/security/lockdown: no such file or directory
```

rafaeldtinoco:
I'm fixing this soon, together with other PR I'm preparing (FYIO).

yanivagman:
Not only that, we should also make sure that with older kernels, were `/sys/kernel/security/lockdown` doesn't exist we don't fail tracee

PR #1402, opened 2022-01-25
main.go: fix missing security lockdown sysfs file

rafaeldtinoco:
After commit a5bc030, tracee-make stopped being able to run tracee-ebpf because the lack of '/sys/kernel/security' dir inside the container. That also happened to any other system not providing that file.

Exit only if lockdown mode is explicitly set to CONFIDENTIALITY, otherwise ignore it.

Fix: #1396

### DIFF
--- a/builder/Makefile.tracee-container
+++ b/builder/Makefile.tracee-container
@@ -129,14 +129,21 @@ build-alpine-tracee-nocore: \
 # run tracee (tracee-ebpf + tracee-rules)
 #
 
+DOCKER_RUN_ARGS = run --rm --pid=host --privileged \
+		-v /etc/os-release:/etc/os-release-host:ro \
+		-v $(PWD):/tracee \
+		-v /lib/modules:/lib/modules:ro \
+		-v /usr/src:/usr/src:ro \
+		-v /sys/kernel/security:/sys/kernel/security:ro \
+		-e LIBBPFGO_OSRELEASE_FILE=/etc/os-release-host
+
 .PHONY: run-alpine-tracee-core
 run-alpine-tracee-core: \
 	| .check_$(CMD_DOCKER) \
 	.check_tree
 #
-	$(CMD_DOCKER) run --privileged --pid=host \
-		-v /etc/os-release:/etc/os-release-host:ro \
-		-e LIBBPFGO_OSRELEASE_FILE=/etc/os-release-host \
+	$(CMD_DOCKER) \
+		$(DOCKER_RUN_ARGS) \
 		-e TRACEE_EBPF_ONLY=0 \
 		--rm -it $(ALPINE_TRACEE_CORE_CONTNAME) \
 		$(ARG)
@@ -146,9 +153,8 @@ run-alpine-tracee-core-btfhub: \
 	| .check_$(CMD_DOCKER) \
 	.check_tree
 #
-	$(CMD_DOCKER) run --privileged --pid=host \
-		-v /etc/os-release:/etc/os-release-host:ro \
-		-e LIBBPFGO_OSRELEASE_FILE=/etc/os-release-host \
+	$(CMD_DOCKER) \
+		$(DOCKER_RUN_ARGS) \
 		-e TRACEE_EBPF_ONLY=0 \
 		--rm -it $(ALPINE_TRACEE_CORE_BTFHUB_CONTNAME) \
 		$(ARG)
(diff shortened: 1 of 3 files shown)


rafaeldtinoco:
@simar7 would you mind checking this PR (to see if solves your issue) and +1'ing it so it can be merged ? Thanks!

builder/Makefile.tracee-container
```diff
@@ -129,14 +129,21 @@ build-alpine-tracee-nocore: \
 # run tracee (tracee-ebpf + tracee-rules)
 #
 
+DOCKER_RUN_ARGS = run --rm --pid=host --privileged \
+		-v /etc/os-release:/etc/os-release-host:ro \
+		-v $(PWD):/tracee \
+		-v /lib/modules:/lib/modules:ro \
+		-v /usr/src:/usr/src:ro \
+		-v /sys/kernel/security:/sys/kernel/security:ro \
```
simar7, line 137: is this path present on every relevant distro? What happens if it isn't?

cmd/tracee-ebpf/main.go
```diff
@@ -83,12 +83,8 @@ func main() {
 			cfg.Output = &output
 
 			// kernel lockdown check
-
 			lockdown, err := helpers.Lockdown()
-			if err != nil {
-				return err
-			}
-			if lockdown == helpers.CONFIDENTIALITY {
+			if err == nil && lockdown == helpers.CONFIDENTIALITY {
```
simar7, line 87: It sounds like an error can happen in two cases:
1. File was not found
2. File was found but had irrelevant content

In both cases it seems we want to proceed ahead. If so, does it make sense to return an error from this function?

simar7:
As discussed offline, we would need to add some unit tests for libbpfgo where this logic is present: https://github.com/aquasecurity/libbpfgo/blob/main/helpers/osinfo.go this would make it easier for the downstream users of libbpfgo to be able to safely use and guard themselves against regressions.

builder/Makefile.tracee-container
rafaeldtinoco: Distros have to have CONFIG_SECURITYFS kconfig option defined so they can have the "securityfs" filesystem mounted in that directory.

```
$ mount -t securityfs securityfs /sys/kernel/security
```

From all distros I checked, and I'm currently testing tracee with, **ALL** of them have that directory (because all of them have CONFIG_SECURITYFS enabled) and have the "securityfs" filesystem mounted on it.

Problem with Alpine Linux is that it doesn't have CONFIG_SECURITY_LOCKDOWN_LSM=y on its kconfig options, but it does have CONFIG_SECURITYFS set (so the dir exists).

PS: Please let me know if you have a distro that doesn't have this set (I would be interested to understand why).

cmd/tracee-ebpf/main.go
rafaeldtinoco: That was discussed at https://github.com/aquasecurity/libbpfgo/pull/103#discussion_r780512300 and returning the error seemed appropriate. Also, we might have other users of libbpfgo that might want to know if the file didn't exist or the error was on its contents (for example). Makes sense ?

rafaeldtinoco:
> As discussed offline, we would need to add some unit tests for libbpfgo where this logic is present: https://github.com/aquasecurity/libbpfgo/blob/main/helpers/osinfo.go this would make it easier for the downstream users of libbpfgo to be able to safely use and guard themselves against regressions.

I'm adding this as a TODO of https://github.com/aquasecurity/libbpfgo/pull/123 (because we're using that PR to stabilize the next libbpfgo release).

Let me know if you agree to my arguments on the comments you've made. I'll add the unit test to that PR so we're covered for it.

cmd/tracee-ebpf/main.go
simar7: You're right. The other users part convinced me to keep it. Thanks.

builder/Makefile.tracee-container
simar7: That's fair enough. I was just curious to know. Since we know all of our supported distros have this, it's good enough 👍

simar7:
> > As discussed offline, we would need to add some unit tests for libbpfgo where this logic is present: https://github.com/aquasecurity/libbpfgo/blob/main/helpers/osinfo.go this would make it easier for the downstream users of libbpfgo to be able to safely use and guard themselves against regressions.
> 
> 
> 
> I'm adding this as a TODO of https://github.com/aquasecurity/libbpfgo/pull/123 (because we're using that PR to stabilize the next libbpfgo release).
> 
> 
> 
> Let me know if you agree to my arguments on the comments you've made. I'll add the unit test to that PR so we're covered for it.
> 
> 
> 
> [Edit]
> 
> I see you opened https://github.com/aquasecurity/libbpfgo/issues/124 for the osinfo tests (which I agree would be a good initial case to someone wanting to contribute).

Yeah sounds good to me. I like to write the tests with the code but it is just one way to go about them.

simar7 (approved):

rafaeldtinoco:
@grantseltzer not sure if I need another approval (from you, for example) to get this merged. Could you +1 this based on Simars review ? Thanks!

grantseltzer (approved):
L G T M ! @rafaeldtinoco

rafaeldtinoco:
Still blocked from merge. @grantseltzer are you able to merge ? (Just curious). I might have loose my permissions somehow :o).

PR #1403, opened 2022-01-26
Packaging: Introduce a way to build Linux distro pkgs

rafaeldtinoco:
* Packaging: Introduce a way to build Linux distro pkgs

    - builder/Makefile.packaging is responsible for building Linux distros
      packages. Targets should be in form of "distro-bin-codename" and
      should use docker images to build packages and place them in dist/
      dir.

    - packaging/ directory contains all files needed to build the packages,
      including the Dockerfiles that are responsible for building the
      environments that will be used to generate those packages.

Idea behind this is simple: To build .deb files, to each existing/supported ubuntu release, and place them in `dist` directory so they can be included as release artifacts.

I'll expand this to CentOS/Stream, Debian, Fedora, SUSE and Alpine.


builder/Makefile.packaging
```diff
@@ -0,0 +1,136 @@
...
+		touch $@ # avoid target rebuilds due to inexistent file
+	fi
+
+#
+# usage
+#
+
+.PHONY: help
+help:
+	@echo ""
```
rafaeldtinoco, line 48: Ooops, missing help message here.

packaging/ubuntu/rules
```diff
@@ -0,0 +1,46 @@
...
+
+export DH_VERBOSE = 1
+
+.ONESHELL:
+include /etc/os-release
+
+STATIC := 0 # can't use golang tracee-rules if enabled
+BTFHUB := 1 # will include tailored BTF files from btfhub
+
+TRACEE_MAKE = BTFHUB=$(BTFHUB) STATIC=$(STATIC) make -f Makefile.one
```
grantseltzer, line 11: For all available options (versions of ubuntu) tracee-ebpf will be built with CO-RE and BTFHUB support, correct?

rafaeldtinoco: Yes it is always CORE, non static and with BTFHUB support. Shared libs according to the environment it was built on.

docs/building/packaging.md
```diff
@@ -0,0 +1,71 @@
+# Creating Tracee Linux packages
+
+> These instructions are meant to describe how to build Linux distributions
+> packages If you would like to have a local building building and execution
+> environment, [read this](./environment.md) instead.
+
+## Ubuntu Linux
+
+Tracee has a directory "packaging" with a skeleton of the needed "debian/"
+directory called "ubuntu". This skeleton is used by "packaging/ubuntu-pkgs.sh"
```
grantseltzer, line 10: Can you explain what this means? The needed debian directory called ubuntu?

builder/Makefile.packaging
```diff
@@ -0,0 +1,152 @@
...
+#
+
+.PHONY: ubuntu-bin-%
+ubuntu-bin-%: \
+	| .check_tree
+#
+	$(MAKE) .ubuntu-pkgs-make-$* ARG="packaging/ubuntu-pkgs.sh bin $*"
+
+#
+# TODO: fedora...
```
grantseltzer, line 142: Let's make a Github issue to track this instead

grantseltzer (approved):
I've tested this out on multiple ubuntu boxes and looks like it works well, nice work!

Couple questions:
- What's would the process be like to get ubuntu/debian to add tracee to their repos?
- Should we add the built .debs to the release artifacts?

builder/Makefile.packaging
rafaeldtinoco: https://github.com/aquasecurity/tracee/issues/1440

docs/building/packaging.md
rafaeldtinoco: Sure. 

This "ubuntu" directory is, indeed, a "src-tree/debian" directory, needed for debian package. The reason for it NOT to be called "debian" is because I'm reserving "debian" for the Debian set of packages.

The contents of this "ubuntu/" directory are the same as the "src/debian" directory of a debian/ubuntu package, but there are slight differences among debian & ubuntu that will be kept separately. 

So, at the end, we may have something like:

./builder/packaging/ubuntu => will be src_tree/debian during package builds

./builder/packaging/debian => will be src_tree/debian during package builds

./builder/packaging/fedora => will contain the RPMBUILD file

./builder/packaging/archlinux => will contain the PKGBUILD file

and so on...

Issue #1250, opened 2021-12-14
extract stats package from package tracee


itaysk:
from: https://github.com/aquasecurity/tracee/issues/960#issuecomment-992654757

> stats - create new package `stats`. `external.Stats` will move to this package

yanivagman:
Package metrics was introduced as part of #1404

Issue #887, opened 2021-08-07
Improve observability


josedonizetti:
Tracee-ebpf collect some [stats](https://github.com/aquasecurity/tracee/blob/aee7e8f04e08b2b869e8526dbff70e1562d016b5/tracee-ebpf/tracee/tracee.go#L255), which are printed as the epilogue when tracee-ebpf exits. For someone running tracee as a daemonset on a K8S cluster, those metrics (possibly even newer metrics in the future) would be scattered on the N nodes.  To allow a better observability, tracee could support prometheus (or open-telemetry?), which would allow users to have a central view of all tracee metrics on all pods running.

itaysk:
Yes! this was on my radar for a while but always de-prioritized. If you're interested in that, it'll be very much appreciated.

itaysk:
brainstorming here some potentially useful and not too hard to collect data. (this is of course incremental work, not for the initial PR)

metric | description | type | dimentions
--- | --- | --- | ---
tracee_ebpf_events_total | # events collected by Tracee-eBPF | counter | event id/name, container id
tracee_ebpf_lostevents_total | # events lost in the submission buffer | counter |
tracee_ebpf_write_lostevents_total | # events lost in the write buffer | counter |
tracee_ebpf_network_lostevents_total | # events lost in the network buffer | counter |
tracee_capture_$capturetype | # captured $capturetype files | counter |
tracee_capture_$capturetype_bytes_total | # bytes captured $capturetype files | counter |
tracee_rules_events_total | # events ingested by Tracee-Rules | counter | source (currently just Tracee)
tracee_signatures_total | # of loaded signatures | counter
tracee_detections_total | # detections made | counter | signature id, container id, severity
tracee_detections_latency_seconds | how long did it take for a detection to be made (finding.timestamp - finding.context.timestamp) | histogram

- hostname labels can be added during prometheus ingestion
- only if using the `--debug` flag, we can add Go's and OPA's default metrics (for troubleshooting)

NDStrahilevitz:
Since #1404 was merged, are there any other observability improvements we want to add before the milestone? If not we can probably remove this issue from the milestone and reintroduce it to a later one when relevant again.

PR #1404, opened 2022-01-26
Prometheus support

NDStrahilevitz:
* Adds prometheus metrics endpoint to tracee-ebpf cli enabled by flag (defaults to `:3366/metrics`)
* Adds prometheus metrics endpoint to tracee-rules cli enabled by flag (defaults to `:4466/metrics`)
* Adds engine stats
* Moves tracee-ebpf counter to it's own package (reused in engine stats)
* Adds metrics package which registers tracee-ebpf and tracee-rules stats to prometheus
* Move stats definition to metrics package (as part of stabilization #1250)

As part of #887

### DIFF
--- a/cmd/tracee-ebpf/main.go
+++ b/cmd/tracee-ebpf/main.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"io/ioutil"
 	"log"
+	"net/http"
 	"os"
 	"os/signal"
 	"path"
@@ -19,13 +20,17 @@ import (
 	"github.com/aquasecurity/tracee/cmd/tracee-ebpf/internal/flags"
 	"github.com/aquasecurity/tracee/pkg/capabilities"
 	"github.com/aquasecurity/tracee/pkg/external"
+	"github.com/aquasecurity/tracee/tracee-ebpf/metrics"
 	"github.com/aquasecurity/tracee/tracee-ebpf/tracee"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/syndtr/gocapability/capability"
 	cli "github.com/urfave/cli/v2"
 )
 
 var debug bool
 var traceeInstallPath string
+var listenMetrics bool
+var metricsAddr string
 
 var version string
 
@@ -159,6 +164,24 @@ func main() {
 				return fmt.Errorf("error creating Tracee: %v", err)
 			}
 
+			if listenMetrics {
+				err := metrics.RegisterPrometheus(t.Stats())
+				if err != nil {
+					fmt.Fprintf(os.Stderr, "Error registering prometheus metrics: %v\n", err)
+				} else {
+					mux := http.NewServeMux()
+					mux.Handle("/metrics", promhttp.Handler())
+
+					go func() {
+						fmt.Fprintf(os.Stdout, "Serving metrics endpoint at %s\n", metricsAddr)
+						if err := http.ListenAndServe(metricsAddr, mux); err != http.ErrServerClosed {
+							fmt.Fprintf(os.Stderr, "Error serving metrics endpoint: %v\n", err)
+						}
+					}()
+				}
+
+			}
+
 			if err := os.MkdirAll(cfg.Capture.OutputPath, 0755); err != nil {
 				t.Close()
 				return fmt.Errorf("error creating output path: %v", err)
@@ -220,8 +243,8 @@ func main() {
 
 			// always print stats before exiting
 			defer func() {
-				stats := t.GetStats()
-				printer.Epilogue(stats)
+				stats := t.Stats()
+				printer.Epilogue(*stats)
 				printer.Close()
 			}()
 
@@ -275,6 +298,18 @@ func main() {
 				Value:       "/tmp/tracee",
 				Usage:       "path where tracee will install or lookup it's resources",
 				Destination: &traceeInstallPath,
+			},
+			&cli.BoolFlag{
+				Name:        "metrics",
+				Usage:       "enable metrics endpoint",
+				Destination: &listenMetrics,
+				Value:       false,
+			},
+			&cli.StringFlag{
+				Name:        "metrics-addr",
+				Usage:       "listening address of the metrics endpoint server",
+				Value:       ":3366",
+				Destination: &metricsAddr,
 			},
 		},
 	}
--- a/cmd/tracee-rules/main.go
+++ b/cmd/tracee-rules/main.go
@@ -14,13 +14,17 @@ import (
 	"syscall"
 
 	"github.com/aquasecurity/tracee/tracee-rules/engine"
+	"github.com/aquasecurity/tracee/tracee-rules/metrics"
 	"github.com/aquasecurity/tracee/types"
 	"github.com/open-policy-agent/opa/compile"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/urfave/cli/v2"
 )
 
 const (
 	signatureBufferFlag = "sig-buffer"
+	metricsFlag         = "metrics"
+	metricsAddrFlag     = "metrics-addr"
 )
 
 func main() {
@@ -117,6 +121,26 @@ func main() {
 			if err != nil {
 				return fmt.Errorf("constructing engine: %w", err)
 			}
+
+			if c.Bool(metricsFlag) {
+				err := metrics.RegisterPrometheus(e.Stats())
+				if err != nil {
+					fmt.Fprintf(os.Stderr, "Error registering prometheus metrics: %v\n", err)
+				} else {
+					mux := http.NewServeMux()
+					mux.Handle("/metrics", promhttp.Handler())
+
+					go func() {
+						metricsAddr := c.String(metricsAddrFlag)
+						fmt.Fprintf(os.Stdout, "Serving metrics endpoint at %s\n", metricsAddr)
+						if err := http.ListenAndServe(metricsAddr, mux); err != http.ErrServerClosed {
+							fmt.Fprintf(os.Stderr, "Error serving metrics endpoint: %v\n", err)
+						}
+					}()
+				}
+
+			}
+
 			e.Start(sigHandler())
 			return nil
 		},
@@ -187,6 +211,16 @@ func main() {
 				Name:  signatureBufferFlag,
 				Usage: "size of the event channel's buffer consumed by signatures",
 				Value: 1000,
+			},
+			&cli.BoolFlag{
+				Name:  metricsFlag,
+				Usage: "enable metrics endpoint",
+				Value: false,
+			},
+			&cli.StringFlag{
+				Name:  metricsAddrFlag,
+				Usage: "listening address of the metrics endpoint server",
+				Value: ":4466",
 			},
 		},
 	}
--- a/tracee-ebpf/tracee/tracee.go
+++ b/tracee-ebpf/tracee/tracee.go
@@ -15,7 +15,6 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
-	"sync/atomic"
 	"syscall"
 	"time"
 	"unsafe"
@@ -28,6 +27,7 @@ import (
 	"github.com/aquasecurity/tracee/pkg/events/sorting"
 	"github.com/aquasecurity/tracee/pkg/external"
 	"github.com/aquasecurity/tracee/pkg/procinfo"
+	"github.com/aquasecurity/tracee/tracee-ebpf/metrics"
 	"github.com/google/gopacket/layers"
 	"github.com/google/gopacket/pcapgo"
 	lru "github.com/hashicorp/golang-lru"
@@ -165,7 +165,7 @@ type Tracee struct {
 	lostNetChannel    chan uint64
 	bootTime          uint64
 	startTime         uint64
-	stats             statsStore
+	stats             metrics.Stats
 	capturedFiles     map[string]int64
 	fileHashes        *lru.Cache
 	profiledFiles     map[string]profilerInfo
@@ -182,37 +182,8 @@ type Tracee struct {
 	eventsSorter      *sorting.EventsChronologicalSorter
 }
 
-type counter int32
-
-func (c *counter) Increment(amount ...int) {
-	sum := 1
-	if len(amount) > 0 {
-		sum = 0
-		for _, a := range amount {
-			sum = sum + a
-		}
-	}
-	atomic.AddInt32((*int32)(c), int32(sum))
-}
-
-type statsStore struct {
-	eventCounter  counter
-	errorCounter  counter
-	lostEvCounter counter
-	lostWrCounter counter
-	lostNtCounter counter
-}
-
-func (t *Tracee) GetStats() external.Stats {
-	var stats external.Stats
-
-	stats.EventCount = int(t.stats.eventCounter)
-	stats.ErrorCount = int(t.stats.errorCounter)
-	stats.LostEvCount = int(t.stats.lostEvCounter)
-	stats.LostWrCount = int(t.stats.lostWrCounter)
-	stats.LostNtCount = int(t.stats.lostNtCounter)
-
-	return stats
+func (t *Tracee) Stats() *metrics.Stats {
+	return &t.stats
 }
 
 // New creates a new Tracee instance based on a given valid Config
@@ -1046,6 +1017,6 @@ func (t *Tracee) invokeInitNamespacesEvent() {
 	if t.eventsToTrace[InitNamespacesEventID] {
 		systemInfoEvent, _ := CreateInitNamespacesEvent()
 		t.config.ChanEvents <- systemInfoEvent
-		t.stats.eventCounter.Increment()
+		t.stats.EventCount.Increment()
 	}
 }
--- a/tracee-rules/engine/engine.go
+++ b/tracee-rules/engine/engine.go
@@ -10,6 +10,7 @@ import (
 	"sync"
 
 	"github.com/aquasecurity/tracee/pkg/external"
+	"github.com/aquasecurity/tracee/tracee-rules/metrics"
 	"github.com/aquasecurity/tracee/types"
 	"github.com/open-policy-agent/opa/ast"
 )
@@ -35,11 +36,16 @@ type Engine struct {
 	output          chan types.Finding
 	waitGroup       sync.WaitGroup
 	config          Config
+	stats           metrics.Stats
 }
 
 //EventSources is a bundle of input sources used to configure the Engine
 type EventSources struct {
 	Tracee chan types.Event
+}
+
+func (e *Engine) Stats() *metrics.Stats {
+	return &e.stats
 }
 
 // NewEngine creates a new rules-engine with the given arguments
@@ -141,6 +147,7 @@ func (engine *Engine) unloadAllSignatures() {
 
 // matchHandler is a function that runs when a signature is matched
 func (engine *Engine) matchHandler(res types.Finding) {
+	engine.stats.Detections.Increment()
 	engine.output <- res
 }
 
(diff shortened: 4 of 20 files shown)


NDStrahilevitz:
I've seen an opportunity to solve #1250 here, but since both tracee and engine now expose stats, i'm not sure about a shared stats pkg, I think it could be done better @itaysk

pkg/stats/tracee_stats.go
```diff
@@ -0,0 +1,9 @@
+package stats
+
+type TraceeStats struct {
```
mtcherni95, line 3: I suggest to put TraceeStats and EngineStats in the same file, and rename the file to stats.go under package stats.
Also, maybe rename to something more meaningful to tracee-ebpf (as these counters are relevant to ebpf side)

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,95 @@
+package metrics
+
+import (
+	"github.com/aquasecurity/tracee/tracee-ebpf/tracee"
+	"github.com/aquasecurity/tracee/tracee-rules/engine"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func RegisterTraceeMetrics(t *tracee.Tracee) error {
+	err := prometheus.Register(prometheus.NewCounterFunc(prometheus.CounterOpts{
```
mtcherni95, line 10: why not using sugar "notation" `if err := prometheus.Register(...); err!=nil{}`

pkg/stats/tracee_stats.go
NDStrahilevitz: Renamed to EbpfStats

pkg/stats/engine_stats.go
```diff
@@ -0,0 +1,7 @@
+package stats
+
+type EngineStats struct {
```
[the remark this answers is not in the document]
NDStrahilevitz: Renamed to RulesStats

itaysk:
out of scope for this PR but in a later PR we should add labels to make this more usable

NDStrahilevitz:
> this is great, thanks! I only took a quick look but perhaps counter doesn't need to be a package and can be merged with stats?

I see the point of this, but the stats package has been split in the end between rules and ebpf into separate `stats` packages so it can't be merged with counter currently.

tracee-ebpf/pkg/stats/stats.go
```diff
@@ -0,0 +1,9 @@
+package stats
+
+type EbpfStats struct {
```
mtcherni95, line 3: I would merge `stats.go` in `metrics.go`, and I would place the package under `tracee-ebpf/metrics/` instead of `tracee-ebpf/pkg/metrics` (kind of what you did in tracee-rules)

tracee-ebpf/tracee/tracee.go
```diff
@@ -177,37 +178,22 @@ type Tracee struct {
...
-	var stats external.Stats
-
-	stats.EventCount = int(t.stats.eventCounter)
-	stats.ErrorCount = int(t.stats.errorCounter)
-	stats.LostEvCount = int(t.stats.lostEvCounter)
-	stats.LostWrCount = int(t.stats.lostWrCounter)
-	stats.LostNtCount = int(t.stats.lostNtCounter)
-
-	return stats
+func (t *Tracee) GetStats() stats.EbpfStats {
```
mtcherni95, line 189:
```suggestion
func (t *Tracee) Stats() stats.EbpfStats {
```

tracee-rules/metrics/metrics.go
```diff
@@ -0,0 +1,40 @@
+package metrics
+
+import (
+	"github.com/aquasecurity/tracee/tracee-rules/engine"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func RegisterRulesEngineMetrics(engine *engine.Engine) error {
+	err := prometheus.Register(prometheus.NewCounterFunc(prometheus.CounterOpts{
```
mtcherni95, line 9:
```suggestion
	err := prometheus.Register(prometheus.NewCounterFunc(prometheus.CounterOpts{...}; err != nil {.....}
```

tracee-rules/engine/stats.go
```diff
@@ -0,0 +1,11 @@
+package engine
+
+import "github.com/aquasecurity/tracee/tracee-rules/stats"
+
+func (e *Engine) GetStats() stats.RulesStats {
```
mtcherni95, line 5:
```suggestion
func (e *Engine) Stats() stats.RulesStats {
```

tracee-ebpf/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/aquasecurity/tracee/tracee-ebpf/tracee"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func RegisterEbpfEngineMetrics(t *tracee.Tracee) error {
+	err := prometheus.Register(prometheus.NewCounterFunc(prometheus.CounterOpts{
+		Namespace: "tracee_ebpf",
+		Name:      "events_total",
+		Help:      "events collected by tracee-ebpf",
+	}, func() float64 { return float64(t.GetStats().EventCount) }))
```
mtcherni95, line 13: You are now exposing prometheus endpoint - and requests can come in parallel.
`Increment()`, correctly, atomically increments the counter. However, read operations from the counter should be done atomically as well - in order to prevent race conditions.

tracee-ebpf/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,60 @@
+package metrics
+
+import (
+	"github.com/aquasecurity/tracee/tracee-ebpf/tracee"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func RegisterEbpfEngineMetrics(t *tracee.Tracee) error {
+	err := prometheus.Register(prometheus.NewCounterFunc(prometheus.CounterOpts{
```
mtcherni95, line 9:
```suggestion
	err := prometheus.Register(prometheus.NewCounterFunc(prometheus.CounterOpts{...}; err != nil {...}
```

tracee-ebpf/tracee/tracee.go
```diff
@@ -177,37 +178,22 @@ type Tracee struct {
...
+	return stats.EbpfStats{
+		EventCount:  int(t.stats.eventCounter),
```
mtcherni95, line 191: Read should be done atomically in order to prevent race condition with`Increment`-ing logics that can happen concurrently

mtcherni95 (changes requested):

tracee-ebpf/pkg/stats/stats.go
NDStrahilevitz: stats can't be merged into metrics since that causes circular dependencies between metrics (which needs tracee's object to register to prometheus) and tracee (which needs the stats protocol to output it). As such they need to be separate packages.

cmd/tracee-rules/main.go
```diff
@@ -174,6 +201,15 @@ func main() {
 				Name:  "list-events",
 				Usage: "print a list of events that currently loaded signatures require",
 			},
+			&cli.BoolFlag{
+				Name:  metricsFlag,
+				Usage: "enable metrics endpoint",
```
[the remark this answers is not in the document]
NDStrahilevitz: Done

cmd/tracee-ebpf/main.go
```diff
@@ -275,6 +300,17 @@ func main() {
 				Usage:       "path where tracee will install or lookup it's resources",
 				Destination: &traceeInstallPath,
 			},
+			&cli.BoolFlag{
```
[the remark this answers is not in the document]
NDStrahilevitz: Done

tracee-ebpf/tracee/tracee.go
NDStrahilevitz: Done

tracee-rules/engine/stats.go
NDStrahilevitz: Done

tracee-ebpf/tracee/tracee.go
NDStrahilevitz: Done through counter.Read()

tracee-ebpf/pkg/metrics/metrics.go
NDStrahilevitz: Done through counter.Read()

cmd/tracee-ebpf/main.go
```diff
@@ -158,6 +163,26 @@ func main() {
 				return fmt.Errorf("error creating Tracee: %v", err)
 			}
 
+			if listenMetrics {
+				err := metrics.RegisterEbpfEngineMetrics(t)
+				if err != nil {
+					if debug {
+						fmt.Fprintf(os.Stderr, "Error registering prometheus metrics: %v\n", err)
```
mtcherni95, line 170: I think, errors should always be printed.

cmd/tracee-ebpf/main.go
```diff
@@ -158,6 +163,26 @@ func main() {
...
+					}
+				} else {
+					mux := http.NewServeMux()
+					mux.Handle("/metrics", promhttp.Handler())
+
+					go func() {
+						fmt.Fprintf(os.Stdout, "Serving metrics endpoint at %s\n", metricsAddr)
```
mtcherni95, line 177: This can be under debug condition (same in tracee-rules)

mtcherni95 (changes requested):

cmd/tracee-ebpf/main.go
NDStrahilevitz: All other errors in the main.go file are only printed on debug mode

cmd/tracee-ebpf/main.go
NDStrahilevitz: There's no debug flag in tracee-rules, so this is based on the pprof endpoints there where this line is always logged.

mtcherni95 (commented):
One more note:
with this commit, if the user passes the `--metrics` flag, it will, by default, open a **Prometheus** endpoint by calling the `RegisterEbpfEngineMetrics` (for tracee-ebpf at least). 
Maybe in future we want to give the ability to open endpoint to other services, such as Datadog, Elastic and such. If so, then the current strategy is less effective. As a suggestion, the CLI could specify to which metrics services we could register for example
`--metrics=prometheus` or `--metrics=prometheus,datadog`
If so, then maybe package metrics (for both tracee-rules and tracee-ebpf), could have inside several files, one for each metrics service. For example: `tracee-rules/metrics/prometheus.go`.

It is all a matter to how much "flexible" we want to be, which is open to discussion.
What are your thoughts on it @itaysk , @danielpacak ?

pkg/metrics/metrics.go
NDStrahilevitz: conversation exists below.

mtcherni95:
> One more note: with this commit, if the user passes the `--metrics` flag, it will, by default, open a **Prometheus** endpoint by calling the `RegisterEbpfEngineMetrics` (for tracee-ebpf at least). Maybe in future we want to give the ability to open endpoint to other services, such as Datadog, Elastic and such. If so, then the current strategy is less effective. As a suggestion, the CLI could specify to which metrics services we could register for example `--metrics=prometheus` or `--metrics=prometheus,datadog` If so, then maybe package metrics (for both tracee-rules and tracee-ebpf), could have inside several files, one for each metrics service. For example: `tracee-rules/metrics/prometheus.go`.
> 
> It is all a matter to how much "flexible" we want to be, which is open to discussion. What are your thoughts on it @itaysk , @danielpacak ?

I think that what we could do is that once we add another data source that we can proceed with the above suggestion. @itaysk @danielpacak

tracee-rules/metrics/prometheus.go
```diff
@@ -0,0 +1,40 @@
+package metrics
+
+import (
+	"github.com/aquasecurity/tracee/tracee-rules/engine"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func RegisterRulesEnginePrometheus(engine *engine.Engine) error {
```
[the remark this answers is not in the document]
danielpacak: And I think all we need to pass to this method is stats.RulesStats . Passing the whole Engine introduce more dependencies between packages.

tracee-rules/stats/stats.go
```diff
@@ -0,0 +1,7 @@
+package stats
+
+type RulesStats struct {
```
danielpacak, line 3: These are metrics exposed by Tracee-rules right? Why don't we just call this struct Metrics? Also Rules prefix is somehow redundant because this code lives in the tracee-rules package

tracee-rules/engine/engine.go
```diff
@@ -29,6 +30,12 @@ type Engine struct {
 	output          chan types.Finding
 	waitGroup       sync.WaitGroup
 	parsedEvents    bool
+	statStore       statStore
+}
+
+type statStore struct {
```
danielpacak, line 36: Why we need both statStore and stats.RulesStats? Can we have one struct that represents Tracee-rules metrics and pass it to NewEngine constructor and to the method that actually registers metrics?

danielpacak (commented):
Overall LGTM! These are great insights exposed as metrics. I left few comments regarding structs and dependencies that we introduce between Go packages in this PR

tracee-rules/engine/stats.go
```diff
@@ -0,0 +1,11 @@
+package engine
+
+import "github.com/aquasecurity/tracee/tracee-rules/stats"
+
+func (e *Engine) Stats() stats.RulesStats {
```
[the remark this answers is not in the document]
NDStrahilevitz: They're in the same package, I wanted to separate them because I personally find it more readable that way (engine.go contains only functions relevant to it's own "real" function and this is more of a side export). I can move it back if that convention is important.

tracee-rules/engine/engine.go
NDStrahilevitz: That's an interesting proposal, i'll PoC it myself and if it works i'll probably introduce it.

danielpacak: Actually looking at the way we use Prometheus Go SDK I'm wondering why we use so many custom structs. Is it absolutely necessary? I think the simplest approach would be:

```go
// Define a single struct that groups tracee-rules metrics
type Metrics struct {
	EventsCount     prometheus.Counter
	DetectionsCount prometheus.Counter
	SignaturesCount prometheus.Counter
}

func registerMetrics() {
	// use promauto package to register counters
	metrics := Metrics{
		EventsCount: promauto.NewCounter(prometheus.CounterOpts{
			Namespace: "tracee_rules",
			Name:      "events_total",
			Help:      "events ingested by tracee-rules",
		}),
		DetectionsCount: promauto.NewCounter(prometheus.CounterOpts{
			Namespace: "tracee_rules",
			Name:      "detections_total",
			Help:      "detections made by tracee-rules",
		}),
		SignaturesCount: prometheus.NewCounter(prometheus.CounterOpts{
			Namespace: "tracee_rules",
			Name:      "signatures_total",
			Help:      "signatures loaded",
		}),
	}
	
	// Pass metrics to the engine; see my comment in another PR about functional options where metrics can be an optinal arg passed to engine constructor.
	NewEngine(metrics)
}

// matchHandler is a function that runs when a signature is matched
func (engine *Engine) matchHandler(res types.Finding) {
	// Increment counters whenever you like
	metrics.EventsCount.Inc()
	engine.output <- res
}

```

danielpacak: Another thing that came to my mind just now is that with dynamically loaded signatures the SignaturesCount may decrease so the counter is probably not appropriate in our case.

tracee-rules/metrics/prometheus.go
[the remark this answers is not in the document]
NDStrahilevitz: Solved in new commits

tracee-rules/engine/stats.go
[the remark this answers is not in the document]
danielpacak: TBH I have never seen methods with the same receiver split across multiple Go source files. Then if you look at my other comments I think it's easier to inject metrics to the Engine as optional arg at construction time.

NDStrahilevitz:
@danielpacak @mtcherni95 I've restructured the solution to decouple the metrics packages from Tracee and Engine, the register functions now take pointers to the Stats struct which is exported from Tracee/Engine. I've also taken into account some other changes you're requested so a re-review is probably required since the PR looks somewhat different now.

cmd/tracee-ebpf/main.go
```diff
@@ -159,6 +164,26 @@ func main() {
 				return fmt.Errorf("error creating Tracee: %v", err)
 			}
 
+			if listenMetrics {
+				err := metrics.RegisterEbpfEnginePrometheus(t)
+				if err != nil {
+					if debug {
```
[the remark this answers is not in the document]
NDStrahilevitz: Resolved

cmd/tracee-ebpf/main.go
```diff
@@ -159,6 +164,26 @@ func main() {
...
+						fmt.Fprintf(os.Stderr, "Error registering prometheus metrics: %v\n", err)
+					}
+				} else {
+					mux := http.NewServeMux()
+					mux.Handle("/metrics", promhttp.Handler())
+
+					go func() {
+						fmt.Fprintf(os.Stdout, "Serving metrics endpoint at %s\n", metricsAddr)
+						if err := http.ListenAndServe(metricsAddr, mux); err != http.ErrServerClosed && debug {
```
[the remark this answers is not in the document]
NDStrahilevitz: Resolved

tracee-rules/stats/stats.go
NDStrahilevitz: Resolved

tracee-rules/engine/engine.go
NDStrahilevitz: PR is now structured using you suggestion by registering the stat struct itself. I've also redefined the signatures count as a "gauge" instead of a counter in prometheus

tracee-rules/engine/stats.go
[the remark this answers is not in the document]
NDStrahilevitz: I went for an approach where the engine exposes it's stat struct pointer. To me splitting receiver functions seems fine but i'm not going to argue with the project's existing conventions, so I merged these functions back (also helps that they are smaller now)

itaysk:
@NDStrahilevitz @mtcherni95  some DoD items: tests, documentation, add o entrypoint

NDStrahilevitz:
> @NDStrahilevitz @mtcherni95 some DoD items: tests, documentation, add to entrypoint

What specifically should be done with regards to the entrypoint? I've added tests for the counter and documented how the metrics registering works near the relevant function.

mtcherni95 (approved):
LGTM! 
@itaysk I believe we can merge if you don't have anything else to add.

itaysk:
> What specifically should be done with regards to the entrypoint?

add it to the entrypoint.. I mean if you want someone to actually use this feature it should be enabled

NDStrahilevitz:
> > What specifically should be done with regards to the entrypoint?
> 
> add it to the entrypoint.. I mean if you want someone to actually use this feature it should be enabled

Sorry for the confusion but do you mean enabling the `--metrics` flag by default through hard coding in the entrypoint? Because I assumed that people would enable by passing the flag if they want (as is done currently)

itaysk:
yes that's what I meant. you should assume that users wouldn't manually configure Tracee, so if you want a feature to be available, you should enable it for them.

NDStrahilevitz:
> 

Done, also added relevant docs to this in `docs/integrations.md`.

danielpacak (approved):
For me stats and metrics are two different things, but except that the PR LGTM! Names are hard :)

tracee-rules/metrics/stats.go
```diff
@@ -0,0 +1,9 @@
+package metrics
+
+import "github.com/aquasecurity/tracee/pkg/counter"
+
+type Stats struct {
```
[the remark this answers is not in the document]
NDStrahilevitz: Since I had to split them from the common package and then merge the two packages (twice, in rules and ebpf), I think it just stayed that way in all that moving around. We could call it `metrics` now but the same thing was called `stats` before so would that be a blocker?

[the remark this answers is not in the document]
danielpacak: I don't think it's a blocker. Just a reminder to stick to well defined terms and data models whenever possible. Especially that Prometheus or OpenTelemetry have well defined domains documented on https://opentelemetry.io/docs/concepts/data-sources/ and https://prometheus.io/docs/concepts/data_model/

PR #1405, opened 2022-01-26
Introducing decoder pkg

mtcherni95:
Hi, after running some benchmarks on Tracee @NDStrahilevitz and I saw that the processEvents logic was using a lot of computational time in binary.Read() function. Please see flame graph: 

image

After exploring the internals of binary.Read() we saw it uses the `reflect` package which compromises program efficiency.
Moreover, from `binary` package description: "_[...] This package favors simplicity over efficiency. Clients that require high-performance serialization, especially for large data structures should look at more advanced solutions [...]_".

The intent of this PR is to **_maximize efficiency over simplicity_**.

A `decoder` package is introduced.

Similarly to `binary`, the `decoder` knows to read a sequence of bytes coming from the Tracee eBPF program to user-space and translate them to golang structs defined under the `protocol` package.

Here is a screenshot from a benchmark tests that compares the efficiency of the new `decoder` package vs `binary.Read` operations.
image

The `decoder` package translation scope is much smaller than the one of `binary.Read`, indeed `binary.Read` accept an interface while the `decoder` translate only to specific data structures. However `decoder` is much **(much!)** more efficient. 

Solves: #1407


tracee-ebpf/tracee/argprinters.go
```diff
@@ -1,28 +1,12 @@
...
-	"encoding/binary"
 	"fmt"
-	"net"
 
 	"github.com/aquasecurity/libbpfgo/helpers"
 	"github.com/aquasecurity/tracee/pkg/external"
 )
 
-// PrintUint32IP prints the IP address encoded as a uint32
-func PrintUint32IP(in uint32) string {
```
mtcherni95, line 13: please note: moved to eventsreader.go

tracee-ebpf/tracee/internal/bufferdecoder/decoder.go
```diff
@@ -0,0 +1,301 @@
...
+	ctx.Uid = binary.LittleEndian.Uint32(decoder.buffer[offset+40 : offset+44])
+	ctx.MntID = binary.LittleEndian.Uint32(decoder.buffer[offset+44 : offset+48])
+	ctx.PidID = binary.LittleEndian.Uint32(decoder.buffer[offset+48 : offset+52])
+	_ = copy(ctx.Comm[:], decoder.buffer[offset+52:offset+68])
+	_ = copy(ctx.UtsName[:], decoder.buffer[offset+68:offset+84])
+	ctx.EventID = int32(binary.LittleEndian.Uint32(decoder.buffer[offset+84 : offset+88]))
+	ctx.Retval = int64(binary.LittleEndian.Uint64(decoder.buffer[offset+88 : offset+96]))
+	ctx.StackID = binary.LittleEndian.Uint32(decoder.buffer[offset+96 : offset+100])
+	ctx.Argnum = uint8(binary.LittleEndian.Uint16(decoder.buffer[offset+100 : offset+102]))
+	decoder.cursor += ctx.GetSizeBytes()
```
yanivagman, line 65: Do we really need all of those GetSizeBytes() functions? To me it seems that they give us no real benefit as anyway the exact bytes offsets are hardcoded here. In addition, we will have to maintain those functions correct whenever we update one of the structs, increasing the chances of future programming mistakes.
If the problem it solves is padding (for example context is of size 104 while we read till byte 102 here), we can just add another field to the relevant structs of padding (for example, two more bytes for padding at the end of Context).

tracee-ebpf/tracee/internal/bufferdecoder/decoder.go
```diff
@@ -0,0 +1,301 @@
...
+import (
+	"encoding/binary"
+	"fmt"
+
+	"github.com/aquasecurity/tracee/pkg/external"
+)
+
+type EbpfDecoder struct {
+	buffer []byte
+	cursor uint32
```
yanivagman, line 18: setting cursor as an int type would save the extra castings to uint32 done below, won't it?

yanivagman (commented):
That's awesome @mtcherni95!
It seems like this will be a big improvement to performance

pkg/external/external.go
```diff
@@ -211,6 +211,10 @@ type SlimCred struct {
 	CapAmbient     uint64 /* Ambient capability set */
 }
 
+func (s SlimCred) GetSizeBytes() uint32 {
+	return 80
+}
+
```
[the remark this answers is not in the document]
mtcherni95: you mean instead of having `externals.SlimCred` object we will have `bufferdecoder.SlimCred` (under file protocol.go)? if so, then totally fine to me.

tracee-ebpf/tracee/process_tree.go
```diff
@@ -68,7 +68,7 @@ func getFileCtime(path string) (int, error) {
 	return int(ts.Sec)*int(time.Second) + int(ts.Nsec)*int(time.Nanosecond), nil
 }
 
-//parse the status file of a process or given task and returns process context struct
+//parse the status file of a process or given task and returns process Context struct
```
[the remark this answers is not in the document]
mtcherni95: oops, automatic find-replace caused it.

pkg/external/external.go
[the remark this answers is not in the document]
yanivagman: Not instead, in addition. We will still need SlimCred to be exposed in external so tracee-rules can use it

tracee-ebpf/tracee/internal/bufferdecoder/decoder.go
rafaeldtinoco: Yep, I agree with @yanivagman. I just left a comment about the hard coded values asking if we could get the arg type's sizes during initialization (using reflect type size or something like that) and have them constant... even if we need to adjusting padding on the affected structs...

rafaeldtinoco (commented):
@mtcherni95 Overall I think this is a very nice change and I like the way it has been done with minimal changes to the parsing logic, etc. Left some comments on top of Yaniv's review, which I think address same concerns I had mostly (and some other nit fixes).

pkg/external/external.go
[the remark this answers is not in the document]
mtcherni95: but then you will have code duplication and and it will make things prone to error, as every update on `externals.SlimCred` will require also an update to `bufferdecoder.SlimCred` (which people might forget to do so).
I can be kind of OK with code dup, but the coupling between these two structures bothers me. They are exactly the same but defined twice.

tracee-ebpf/tracee/internal/bufferdecoder/decoder.go
mtcherni95: @yanivagman , I'm fine with removing the `GetSizeBytes()` functions. I wouldn't add additional data to the structs just for expose the size of it. I would simply add it hard-coded.
@rafaeldtinoco I prefer not to use reflection as it negatively affects efficiency (`binary.Read` uses it, and that's the whole issue with it).

mtcherni95: we could still have GetSizeBytes() which returns an hardcoded value, and in the tests (https://github.com/aquasecurity/tracee/pull/1405/files#diff-242d9e577032010728b6349d850cb4bcb8c517923dd77d50ad684732b43c23efR31) we use reflection to get the size of the structs (`unsafe.Sizeof` doesn't help, I tried it already). If another developer adds/remove fields without updating the decoder then it will fail in tests.

tracee-ebpf/tracee/internal/bufferdecoder/protocol_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+func TestKernelModuleMetaSize(t *testing.T) {
+	var v KernelModuleMeta
+	assert.Equal(t, int(v.GetSizeBytes()), 24)
+}
+
+func TestMprotectWriteMetaSize(t *testing.T) {
+	var v MprotectWriteMeta
+	assert.Equal(t, int(v.GetSizeBytes()), 8)
+}
```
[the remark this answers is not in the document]
mtcherni95: @yanivagman please see this https://github.com/aquasecurity/tracee/pull/1405#discussion_r797514374
If agree, then in the tests I will use reflection to get exact size of the struct and assert it's equal to the hardcoded value.

tracee-ebpf/tracee/internal/bufferdecoder/decoder.go
rafaeldtinoco: > @rafaeldtinoco I prefer not to use reflection as it negatively affects efficiency (`binary.Read` uses it, and that's the whole issue with it).

Yep, it was just a way to say something else than hardcoded in here. Not necessarily needs to be reflect.

pkg/external/external.go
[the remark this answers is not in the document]
mtcherni95: alright, you have a point. done :)

tracee-ebpf/tracee/internal/bufferdecoder/decoder.go
rafaeldtinoco:
```
=== RUN   TestVfsWriteMetaSize
    protocol_test.go:31:
        	Error Trace:	protocol_test.go:31
        	Error:      	Not equal:
        	            	expected: 28
        	            	actual  : 20
        	Test:       	TestVfsWriteMetaSize
--- FAIL: TestVfsWriteMetaSize (0.00s)
=== RUN   TestKernelModuleMetaSize
--- PASS: TestKernelModuleMetaSize (0.00s)
=== RUN   TestMprotectWriteMetaSize
--- PASS: TestMprotectWriteMetaSize (0.00s)
FAIL
FAIL	github.com/aquasecurity/tracee/tracee-ebpf/tracee/internal/bufferdecoder	0.005s
```

After adding a variable to the arg type. I'm fine with that, since the PR checks will get the issue and allow the fix to happen (and I can understand why hard coding the size has better performance).

yanivagman (approved):
LGTM!

Issue #1191, opened 2021-12-01
signatures tests


itaysk:
add a collection of container images, each designed to trigger a specific signature. for the purpose of tests and demos.
each container should be as minimal as possible. exit code denotes the result.

- [ ] anti-debugging
- [ ] cgroup release agent
- [ ] code injection
- [ ] disk mount
- [ ] dropped executable
- [ ] dynamic code loading
- [ ] file-less execution
- [ ] illegitimate shell
- [ ] k8s SA token
- [ ] kernel module loading
- [ ] k8s cert theft
- [ ] ld preload
- [ ] reverse shell

simar7:
This is awesome!

itaysk:
Actually, it's probably easier to have a single docker image with all of the scripts in there (thanks @danielpacak for the suggestion).

Issue #1234, opened 2021-12-10
E2E sanity test


danielpacak:
In scope of this we should try running existing integration tests from tests/tracee_test.go in the GitHub Actions runner as a minimal e2e quality gate.

simar7:
See my old PR trying to work this in - https://github.com/aquasecurity/tracee/pull/1219 

From what I could gather, tracee no longer runs out of the box within GitHub Actions environment. We need to further investigate why this is the case.

grantseltzer:
@simar7 What do you mean that it no longer runs out of the box? I can investigate, and possibly retry having selfhosted runners.

simar7:
> @simar7 What do you mean that it no longer runs out of the box? I can investigate, and possibly retry having selfhosted runners.

Sorry I wasn't clear. What I noticed that Tracee was not running out of the box on GitHub actions with the default options. 

You can see the options I tried to run with in the sanity test PR I linked above.

rafaeldtinoco:
as a fyio on this... the following test, suggested by Simar on Teams:

```
$ sudo ./dist/tracee-ebpf -o format:json -o option:parse-arguments --trace comm=strace | ./dist/tracee-rules --input-tracee file:stdin --input-tracee format:json
```

has been a life saver when dealing with parseArgs and similar changes.

PR #1408, opened 2022-01-27
chore(ci): e2e signatures tests

danielpacak:
Run signatures tests as sanity check on each PR.

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -1,14 +1,2 @@
-build-trainer-docker:
-	docker build -t tracee-trainer:latest .
-
-build-tracee-docker-core:
-	cd .. && make -f builder/Makefile.tracee-container build-alpine-tracee-core
-
-build-tracee-docker-nocore:
-	cd .. && make -f builder/Makefile.tracee-container build-alpine-tracee-nocore
-
-build-tracee-docker-core-btfhub:
-	cd .. && make -f builder/Makefile.tracee-container build-alpine-tracee-core-btfhub
-
-run-tests: build-trainer-docker build-tracee-docker-core build-tracee-docker-nocore build-tracee-docker-core-btfhub
-	go test -v -run=TestTraceeSignatures
+docker-build-tracee-trainer:
+	docker build -t tracee-trainer:latest tests
--- a/tests/tracee_test.go
+++ b/tests/tracee_test.go
@@ -188,9 +188,21 @@ func TestTraceeSignatures(t *testing.T) {
 		os.RemoveAll(tempDir)
 	}()
 
-	for _, image := range []string{"tracee", "tracee-btfhub", "tracee-nocore"} {
-		for _, sigid := range []string{"TRC-3", "TRC-4", "TRC-9", "TRC-10", "TRC-11"} {
-			t.Run(fmt.Sprintf("%s/%s", image, sigid), func(t *testing.T) {
+	// Ubuntu 20.04 provided by GitHub Actions runner does not support CO:RE.
+	// Thus, we are running end-to-end signatures tests using tracee non CO:RE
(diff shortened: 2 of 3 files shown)


rafaeldtinoco (approved):
LGTM. Following TODOs I see: I really see a need for us to move Dockerfile and Makefile around with better namings for the tests. Or, at least, have them renamed to something else. I think we can do this soon, not needed in this PR. Just sharing a feeling.

tests/tracee_test.go
```diff
@@ -188,9 +188,21 @@ func TestTraceeSignatures(t *testing.T) {
...
-		for _, sigid := range []string{"TRC-3", "TRC-4", "TRC-9", "TRC-10", "TRC-11"} {
-			t.Run(fmt.Sprintf("%s/%s", image, sigid), func(t *testing.T) {
+	// Ubuntu 20.04 provided by GitHub Actions runner does not support CO:RE.
+	// Thus, we are running end-to-end signatures tests using tracee non CO:RE
+	// container image.
+
+	// FIXME Pass tracee container image flavor (tracee-nocore, tracee-core, etc.)
+	//       as input parameter to this test so we can set in the CI workflow
+	//       instead of hardcoding it here. The actual logic of the test should be
+	//       agnostic of tracee container flavor.
```
grantseltzer, line 198: Can we have an issue for this instead of a FIXME comment?

tests/tracee_test.go
```diff
@@ -188,9 +188,21 @@ func TestTraceeSignatures(t *testing.T) {
...
+	for _, image := range []string{"tracee-nocore"} {
+		// FIXME Pass signature identifiers (TRC-3, TRC-4, TRC-9, etc.) as input
+		//       parameter to this test so we can use it as smoke test in the
+		//       PR validation workflow (with TRC-2) only or as full-blown end-to-end
+		//       nightly test run.
```
grantseltzer, line 203: Same here

tests/tracee_test.go
danielpacak: I believe that to close #1191 and #1234 we must resolve these comments. And I'm actually going to jump right on it after we merge this PR.

tests/Makefile
```diff
@@ -1,14 +1,2 @@
...
-	cd .. && make -f builder/Makefile.tracee-container build-alpine-tracee-nocore
-
-build-tracee-docker-core-btfhub:
-	cd .. && make -f builder/Makefile.tracee-container build-alpine-tracee-core-btfhub
-
-run-tests: build-trainer-docker build-tracee-docker-core build-tracee-docker-nocore build-tracee-docker-core-btfhub
-	go test -v -run=TestTraceeSignatures
\ No newline at end of file
+docker-build-tracee-trainer:
+	docker build -t tracee-trainer:latest tests
```
simar7, line 2: Is this correct? 
```vagrant@ubuntu-impish:~/repos/tracee/tests$ make docker-build-tracee-trainer 
docker build -t tracee-trainer:latest tests
unable to prepare context: path "tests" not found
make: *** [Makefile:2: docker-build-tracee-trainer] Error 1
```

Maybe you meant: `docker build -t tracee-trainer:latest .`?

simar7: Sorry ignore that - I understood your motivation now and where it would be run from.

danielpacak: It's a temporary change to run it from the root directory. Relative paths are tricky. We are going to revisit makefile targets with @rafaeldtinoco to make it consistent with other build scripts

tests/tracee_test.go
danielpacak: See #1411 that resolves FIXME and TODOs

PR #1411, opened 2022-01-28
test: parameterize e2e signatures test and use it as sanity check

danielpacak:
Resolves: #1234
Resolves: #1191


rafaeldtinoco (approved):
LGTM

Issue #1332, opened 2022-01-10
TRC-11 support other devices in the mount


krol3:
I am testing the TRC-11 using a virtual machine.

It seems that we only support devices with sd name. https://github.com/aquasecurity/tracee/blob/main/signatures/rego/disk_mount.rego#L34

In my case, I tried using `mount /dev/vda1 /mnt01/` I couldnt see the alert TRC-11

```
Device      Start      End  Sectors  Size Type
/dev/vda1  227328 31457246 31229919 14.9G Linux filesystem
/dev/vda14   2048    10239     8192    4M BIOS boot
/dev/vda15  10240   227327   217088  106M EFI System
```

roikol:
hey ,
as the signature's description suggests - the signature only attempts to detect a mount of a container filesystem, not every mount in the system.
is that what you are trying to do?

in any case, we should consider renaming the file.

itaysk:
> the signature only attempts to detect a mount of a container filesystem

I think the feedback is that not all mount of container FS are detected

roikol:
i was actually incorrect in my comment. the signature attempts to detect a mount of the host filesystem into the container.

you're right, it wouldn't be detected in this case. 

we are thinking of the best way to solve this.

yanivagman:
Closing this as #1421 was merged

PR #1421, opened 2022-01-31
disk_mount signature alert on every device mount

roikol:
changed the signature to alert on every device mount from within the container.
changed the description accordingly.
related #1332.


rafaeldtinoco (approved):
LGTM.

Issue #1341, opened 2022-01-12
cap_capable doesn't monitor all capabilities with CO-RE


yanivagman:
After taking a look at #1340 I noticed that we don't monitor the capabilities correctly when CO-RE is enabled.
The reason for this is that CAP_OPT_NONE is not defined in CO-RE.

rafaeldtinoco:
Despite `CAP_OPT_NONE` not being defined... the logic is unfinished (as it seems), but it is currently working:

ubuntu 5.4.0-91-generic

```
$ sudo ./dist/tracee-ebpf -o format:json -o option:parse-arguments --trace event=cap_capable | jq -jc '"\nPID: ", .processId, " COMM: ",.processName," EVENT: ",.eventName," ARGS: ", .args, "\n"'

PID: 993 COMM: lxd-agent EVENT: cap_capable ARGS: [{"name":"cap","type":"string","value":"CAP_SETUID"}]
PID: 993 COMM: lxd-agent EVENT: cap_capable ARGS: [{"name":"cap","type":"string","value":"CAP_SYS_ADMIN"}]
PID: 993 COMM: sh EVENT: cap_capable ARGS: [{"name":"cap","type":"string","value":"CAP_SYS_ADMIN"}]
PID: 993 COMM: sh EVENT: cap_capable ARGS: [{"name":"cap","type":"string","value":"CAP_SYS_ADMIN"}]
PID: 993 COMM: sh EVENT: cap_capable ARGS: [{"name":"cap","type":"string","value":"CAP_SYS_ADMIN"}]
PID: 993 COMM: sh EVENT: cap_capable ARGS: [{"name":"cap","type":"string","value":"CAP_SYS_ADMIN"}]
PID: 993 COMM: sh EVENT: cap_capable ARGS: [{"name":"cap","type":"string","value":"CAP_SYS_ADMIN"}]
PID: 993 COMM: sh EVENT: cap_capable ARGS: [{"name":"cap","type":"string","value":"CAP_SYS_ADMIN"}]
PID: 993 COMM: sh EVENT: cap_capable ARGS: [{"name":"cap","type":"string","value":"CAP_SYS_ADMIN"}]
PID: 993 COMM: sh EVENT: cap_capable ARGS: [{"name":"cap","type":"string","value":"CAP_SYS_ADMIN"}]
PID: 993 COMM: sh EVENT: cap_capable ARGS: [{"name":"cap","type":"string","value":"CAP_SYS_ADMIN"}]
[...]

PR #1422, opened 2022-01-31
tracee-ebpf: cap_capable always submit requested cap

rafaeldtinoco:
Fix: #1341

The only execution path when cap_capable opts argument isn't CAP_OPT_NONE is is when being executed by has_cap_mac_admin() (selinux). In that case, it might be either CAP_OPT_NONE or CAP_OPT_NOAUDIT.

After https://github.com/aquasecurity/tracee/discussions/1340, let's always submit "cap_capable" event with the correct capability request.

### DIFF
--- a/tracee-ebpf/tracee/tracee.bpf.c
+++ b/tracee-ebpf/tracee/tracee.bpf.c
@@ -2942,32 +2942,21 @@ int BPF_KPROBE(trace_switch_task_namespaces)
 SEC("kprobe/cap_capable")
 int BPF_KPROBE(trace_cap_capable)
 {
-    int audit;
-
     event_data_t data = {};
+
     if (!init_event_data(&data, ctx))
         return 0;
 
     if (!should_trace(&data.context))
         return 0;
 
-    //const struct cred *cred = (const struct cred *)PT_REGS_PARM1(ctx);
-    //struct user_namespace *targ_ns = (struct user_namespace *)PT_REGS_PARM2(ctx);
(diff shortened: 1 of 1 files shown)


rafaeldtinoco:
And, tested in multiple CO-RE environments:

```
$ sudo ./dist/tracee-ebpf -o format:json -o option:parse-arguments -o option:detect-syscall --trace event=cap_capable | jq -jc '"\nPID: ", .processId, " COMM: ",.processName," EVENT: ",.eventName," ARGS: ", .args, "\n"'

PID: 3598912 COMM: lxcfs EVENT: cap_capable ARGS: [{"name":"cap","type":"string","value":"CAP_SYS_PTRACE"},{"name":"syscall","type":"int","value":"newstat"}]
PID: 3598912 COMM: lxcfs EVENT: cap_capable ARGS: [{"name":"cap","type":"string","value":"CAP_SYS_PTRACE"},{"name":"syscall","type":"int","value":"newstat"}]
PID: 1588503 COMM: udisksd EVENT: cap_capable ARGS: [{"name":"cap","type":"string","value":"CAP_SYS_NICE"},{"name":"syscall","type":"int","value":"sched_setattr"}]
```

Seems to be working fine.

yanivagman:
See my comment here:
https://github.com/aquasecurity/tracee/issues/1341#issuecomment-1026217705

rafaeldtinoco:
I'm closing this after brief discussion with Yaniv and  a change we agreed

PR #1424, opened 2022-02-01
tracee-ebpf: fix cap_capable when using CO-RE

### LINKED ISSUE #1341: cap_capable doesn't monitor all capabilities with CO-RE

yanivagman:
Add missing cap_opts definition in CO-RE to fix cap_capable event which
didn't monitor the requested capabilities correctly.

Fix #1341


rafaeldtinoco (approved):
LGTM, I agree we should only handle this case now that you confirmed 4.19 stable has backported the defines.

PR #1425, opened 2022-02-01
chore: rename tracee-trainer to tracee-tester

danielpacak:
Adds also manually dispatched GitHub Actions workflow
to publish the aquasec/tracee-tester:latest container
image to DockerHub.


danielpacak:
> What are your thoughts on having this image push to dockerhub with releases instead of manually? I can include it in #1367

We use it internally to test tracee before it's actually released so I'm not sure it's another release artifact that we want to publish. What's more, it has a different purpose and life-cycle, i.e. probably we won't change it that often. On the other hand, we may need new version of tracee-tester before we actually release tracee.

The main idea to publish tracee-tester to DockerHub was to speed up PR validation workflow and avoid rebuilding this image over and over again.

.github/workflows/publish-tracee-tester.yaml
```diff
@@ -0,0 +1,30 @@
...
+    name: Deploy the latest documentation
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Checkout main
+        uses: actions/checkout@v2
+      - name: Build image
+        run: |
+          docker image build --tag ${{ github.event.inputs.tag }} tests/tracee-tester
+      - name: Login to docker.io registry
+        uses: docker/login-action@v1
```
grantseltzer, line 24: I should be using this in the release workflow :+1:

grantseltzer (approved):
LGTM

PR #1428, opened 2022-02-02
chore(ci): skip building aquasec/tracee-tester:latest container image

danielpacak:
We've already published aquasec/tracee-tester:latest to [DockerHub](https://hub.docker.com/r/aquasec/tracee-tester) so we can just use it for testing.


rafaeldtinoco (approved):
LGTM

PR #1429, opened 2022-02-02
Buffered signature channels

NDStrahilevitz:
During the performance testing that @mtcherni95 and I are doing, we've found that one of tracee-rule's bottlenecks (especially when running with low cpu constraints) is blocked channels on sending and receiving. We've also measured high times spent in `runtime.futex`, which is related to the channel blocking as I will demonstrate.

This PR enables configuring the event channels buffers that stream to signatures in tracee-rules. The default value set in the CLI is 1000. This should decrease the channel blocking overhead when some signatures receive many events, and reduce time in context switching between the signature go routines.

The PR also introduces a `Config` struct for the tracee-rules engine.

First, the case of unbuffered channels: 

Currently, signatures in the rules engine are implemented as a map between signatures and channels of events they receive. These channels are hard coded to be unbuffered. Here i've attached a flame graph of this case and a `top` table of cpu time spent:
image
image

As we can see in this case much time is spent in `runtime.futex` and the chanrecv and chansend functions take 7.85% and 7.05% of cpu time respectively, cumulating in 14.95% of cpu time.

This isn't the case for buffered channels (set to 1000):
image
Here we can see a drop in chanrecv and chansend time with 4.77% and 3.82% of cpu time respectively adding up to 8.59% of cpu time.

And for the top graph a differential pprof between the cases:
image

As we can see here there was a flat drop of 10.73% of cpu time on context switching. This means buffering the channels, as done in this pr, overall freed up 19.32% of cpu time.

I've also attached the relevant pprofs here: 
[0buffer.pb.gz](https://github.com/aquasecurity/tracee/files/7985815/0buffer.pb.gz)
[1000buffer.pb.gz](https://github.com/aquasecurity/tracee/files/7985816/1000buffer.pb.gz)

### DIFF
--- a/cmd/tracee-rules/main.go
+++ b/cmd/tracee-rules/main.go
@@ -111,7 +111,7 @@ func main() {
 
 			config := engine.Config{
 				ParsedEvents:        c.Bool("rego-enable-parsed-events"),
-				SignatureBufferSize: c.Int(signatureBufferFlag),
+				SignatureBufferSize: c.Uint(signatureBufferFlag),
 			}
 			e, err := engine.NewEngine(sigs, inputs, output, os.Stderr, config)
 			if err != nil {
@@ -183,7 +183,7 @@ func main() {
 				Name:  "list-events",
 				Usage: "print a list of events that currently loaded signatures require",
 			},
-			&cli.IntFlag{
+			&cli.UintFlag{
 				Name:  signatureBufferFlag,
 				Usage: "size of the event channel's buffer consumed by signatures",
 				Value: 1000,
--- a/cmd/tracee-rules/main.go
+++ b/cmd/tracee-rules/main.go
@@ -19,6 +19,10 @@ import (
 	"github.com/urfave/cli/v2"
 )
 
+const (
+	signatureBufferFlag = "sig-buffer"
+)
+
 func main() {
 	app := &cli.App{
 		Name:  "tracee-rules",
@@ -106,7 +110,8 @@ func main() {
 			}
 
 			config := engine.Config{
-				ParsedEvents: c.Bool("rego-enable-parsed-events"),
+				ParsedEvents:        c.Bool("rego-enable-parsed-events"),
+				SignatureBufferSize: c.Int(signatureBufferFlag),
 			}
 			e, err := engine.NewEngine(sigs, inputs, output, os.Stderr, config)
 			if err != nil {
@@ -177,6 +182,11 @@ func main() {
 			&cli.BoolFlag{
 				Name:  "list-events",
 				Usage: "print a list of events that currently loaded signatures require",
+			},
+			&cli.IntFlag{
+				Name:  signatureBufferFlag,
+				Usage: "size of the event channel's buffer consumed by signatures",
+				Value: 1000,
 			},
 		},
 	}
--- a/cmd/tracee-rules/main.go
+++ b/cmd/tracee-rules/main.go
@@ -104,7 +104,11 @@ func main() {
 			if err != nil {
 				return err
 			}
-			e, err := engine.NewEngine(sigs, inputs, output, os.Stderr, c.Bool("rego-enable-parsed-events"))
+
+			config := engine.Config{
+				ParsedEvents: c.Bool("rego-enable-parsed-events"),
+			}
+			e, err := engine.NewEngine(sigs, inputs, output, os.Stderr, config)
 			if err != nil {
 				return fmt.Errorf("constructing engine: %w", err)
 			}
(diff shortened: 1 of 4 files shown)


cmd/tracee-rules/main.go
```diff
@@ -19,6 +19,10 @@ import (
 	"github.com/urfave/cli/v2"
 )
 
+const (
+	signatureBufferFlag = "sig-buffer"
```
mtcherni95, line 23: nit: not really required - I would keep the current standard without defining a const var for it

tracee-rules/engine/engine.go
```diff
@@ -47,7 +53,7 @@ func NewEngine(sigs []types.Signature, sources EventSources, output chan types.F
 	engine.logger = *log.New(logWriter, "", 0)
 	engine.inputs = sources
 	engine.output = output
-	engine.parsedEvents = parsedEvents
+	engine.config = config
```
mtcherni95, line 56: we should validate the configuration. what if the channel size is passed with a negative value?

cmd/tracee-rules/main.go
```diff
@@ -174,6 +183,11 @@ func main() {
 				Name:  "list-events",
 				Usage: "print a list of events that currently loaded signatures require",
 			},
+			&cli.IntFlag{
+				Name:  signatureBufferFlag,
+				Usage: "size of the event channel's buffer consumed by signatures",
+				Value: 1000,
```
danielpacak, line 189: Do we want to enable buffered channels by default? Setting the default value to 0 would fallback to the existing ubbuffered channels.

tracee-rules/engine/engine.go
```diff
@@ -238,7 +244,7 @@ func (engine *Engine) LoadSignature(signature types.Signature) (string, error) {
 		// signature already exists
 		return metadata.ID, nil
 	}
-	c := make(chan types.Event)
+	c := make(chan types.Event, engine.config.SignatureBufferSize)
```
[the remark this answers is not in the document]
danielpacak: BTW, what about the first channel that we instantiate in tracee-rules in the setupTraceeGobInputSource method? This one is still unbuffered

```go
func setupTraceeGobInputSource(opts *traceeInputOptions) (chan types.Event, error) {
	dec := gob.NewDecoder(opts.inputFile)
	gob.Register(make(map[string]string))
	res := make(chan types.Event) // Do we keep this one as unbuffered?
	go func() {
```

danielpacak (commented):
Overall LGTM. I just left a few comments and questions to get your feedback.

cmd/tracee-rules/main.go
mtcherni95: I think having buffer channels should be the default IMO . From a business logic point of view, it doesn't matter wether we are using buffer/unbuffer. However, buffered channel are more efficient as in case of high rate of writing operation to the channel, the logics **won't** be blocked.

danielpacak: Yup. That makes sense to me as well. I just wanted to confirm that it's intentional ;)

tracee-rules/engine/engine.go
NDStrahilevitz: Would changing the size to a uint be fine in this case?

tracee-rules/engine/engine.go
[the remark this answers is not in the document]
NDStrahilevitz: > Do we need to do the same in NewEngine constructor? In line 63 we also instantiate channels.

Yes thank you, it should go there as well, i'll upload a fix.

danielpacak (approved):

cmd/tracee-rules/main.go
```diff
@@ -104,7 +108,12 @@ func main() {
 			if err != nil {
 				return err
 			}
-			e, err := engine.NewEngine(sigs, inputs, output, os.Stderr, c.Bool("rego-enable-parsed-events"))
+
+			config := engine.Config{
+				ParsedEvents:        c.Bool("rego-enable-parsed-events"),
+				SignatureBufferSize: c.Uint(signatureBufferFlag),
```
mtcherni95, line 114: what if we pass -1? can we try to run with passing -1? the cli should output error.

NDStrahilevitz: From a test right now:
`2022/02/03 11:44:10 invalid value "-1" for flag -sig-buffer: parse error` so it should be safe

tracee-rules/engine/engine.go
NDStrahilevitz: Continued below

PR #1430, opened 2022-02-02
test: remove obsolete TestLaunchTracee test

danielpacak:
We have a new parameterized TestTraceeSignatures that does
the same thing (and more) and is actually run in the CI
pipeline.

### DIFF
--- a/tests/tracee_test.go
+++ b/tests/tracee_test.go
@@ -21,8 +21,6 @@ import (
 
 const (
 	waitTime                   = time.Second * 3
-	traceeDockerRunBTFEnabled  = `run --detach --name tracee --rm --pid=host --privileged -v /tmp/tracee:/tmp/tracee -t aquasec/tracee:latest`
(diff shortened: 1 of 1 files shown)


tests/tracee_test.go
```diff
@@ -46,58 +44,6 @@ func runCommand(t *testing.T, cmd string, args ...string) string {
 	return string(output)
 }
 
-// TestLaunchTracee tests the basic sanity workflow of running tracee
-// and detecting an attack by simulating a signature trigger
-func TestLaunchTracee(t *testing.T) {
```
danielpacak, line 51: I wasn't sure if we want to keep TestWebhookIntegration. @simar7  what was the idea behind it? You wanted to run it on each PR, nightly, weekly, before release? I think it's not very useful if we don't run such tests.

simar7 (approved):

tests/tracee_test.go
simar7: The idea was to test webhook :) but more importantly it was giving us a better way to test output by using an httptestserver. 

Maybe we have better tests now and it needs to go.

PR #1433, opened 2022-02-03
test: remove obsolete TestWebhookIntegration

danielpacak:
We have a new parameterized TestTraceeSignatures that does
the same thing and is run in the CI pipeline.


grantseltzer (approved):
LGTM

PR #1437, opened 2022-02-03
builder: container images: slim and full

rafaeldtinoco:
```
This commit is the result of a broader discussion regarding how we
should distribute tracee and how CO-RE and non CO-RE feature, and wording,
should be handled.

We, now, will have the following container images:

- tracee:latest
- tracee:full

And they are generated by:

- make -f builder/Makefile.tracee-container build-tracee
- make -f builder/Makefile.tracee-container build-tracee-full

And might be tested by:

- make -f builder/Makefile.tracee-container run-tracee
- make -f builder/Makefile.tracee-container run-tracee-full

  Obs: images need to be built before run targets are executed.

Both images can run:

- tracee-eBPF with CO-RE
- tracee-eBPF with CO-RE + BTFHub (tailored BTF files)
- tracee-eBPF non CO-RE (from /tmp/tracee non CO-RE eBPF obj)

So, one may built the non CO-RE object by using:

- make -f builder/Makefile.tracee-container run-tracee-full

and subsequent calls to the smaller image will work:

- make -f builder/Makefile.tracee-container run-tracee

  Obs: makefile automatically shares /tmp/tracee so this can be true.

The full image is different because:

- it contains tracee src code, tools and compilers
- it builds /tmp/tracee/xxx.bpf.o non CO-RE eBPF obj, if needed

  Obs: One can use the full image to generate needed non CO-RE obj and
       distribute it through different nodes, with same kernel, in order
       to run the smaller (slim) image on them.
```

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -139,7 +139,6 @@ jobs:
           severity: "CRITICAL"
           exit-code: "1"
 
-
   # smoke-test-noncore job is using TRC-2 (Anti-Debugging) signature and tracee
   # non CO-RE container image to run a quick smoke test on each PR.
   #
@@ -165,9 +164,9 @@ jobs:
         with:
           go-version: ${{ env.GO_VERSION }}
           opa-version: ${{ env.OPA_VERSION }}
-      - name: Build tracee-nocore
+      - name: Build tracee image (full w/out BTFHUB)
         run: |
-          make -f builder/Makefile.tracee-container build-alpine-tracee-nocore
+          BTFHUB=0 make -f builder/Makefile.tracee-container build-tracee-full
       - name: Install BPF
         run: |
           make -f Makefile.one install-bpf-nocore
@@ -175,7 +174,7 @@ jobs:
         run: |
           docker image pull aquasec/tracee-tester:latest
           go test -v -run "TestTraceeSignatures" ./tests/tracee_test.go \
-            -tracee-image-ref "tracee-nocore:latest" \
+            -tracee-image-ref "tracee:full" \
             -tracee-tester-image-ref "aquasec/tracee-tester:latest" \
             -tracee-signatures "TRC-2"
 
@@ -204,17 +203,16 @@ jobs:
       - name: Built tracee BPF
         run: |
           make -f Makefile.one bpf-core
-      - name: Build tracee-core-btfhub
+      - name: Build tracee image (slim w/ BTFHUB)
         run: |
-          make -f builder/Makefile.tracee-container build-alpine-tracee-core-btfhub
+          BTFHUB=1 make -f builder/Makefile.tracee-container build-tracee
       - name: Run tests
         run: |
           docker image pull aquasec/tracee-tester:latest
           go test -v -run "TestTraceeSignatures" ./tests/tracee_test.go \
-            -tracee-image-ref "tracee-btfhub:latest" \
+            -tracee-image-ref "tracee:latest" \
             -tracee-tester-image-ref "aquasec/tracee-tester:latest" \
             -tracee-signatures "TRC-2"
-
 
   # deprecated-verify-tracee-ebpf and deprecated-verify-tracee-rules jobs are
   # supposed to be removed once we fully migrate to Makefile.one and get rid of
--- a/docs/building/containers.md
+++ b/docs/building/containers.md
@@ -8,97 +8,71 @@
 
 ## Generating Tracee containers
 
-Tracee containers come in **3 flavors**:
+Tracee containers come in **2 flavors**:
 
-* `tracee`: default [CO-RE eBPF](https://nakryiko.com/posts/bpf-portability-and-co-re/) tracee (portable).
-
-  ```
-  $ make -f builder/Makefile.tracee-container build-alpine-tracee-core
-  ```
-
-* `tracee-btfhub`: CO-RE eBPF tracee with [older kernels support](https://github.com/aquasecurity/btfhub/blob/main/docs/supported-distros.md).
+* `tracee:latest`
 
   ```
-  $ make -f builder/Makefile.tracee-container build-alpine-tracee-core-btfhub
+  $ BTFHUB={0,1} make -f builder/Makefile.tracee-container build-tracee
   ```
 
-* `tracee-nocore`: non CO-RE eBPF binary (eBPF code is compiled before the run).
+* `tracee:full`
 
   ```
-  $ make -f builder/Makefile.tracee-container build-alpine-tracee-nocore
+  $ BTFHUB={0,1} make -f builder/Makefile.tracee-container build-tracee-full
   ```
 
-At the end you will end up with 3 docker images built. Example:
+> Note: BTFHUB=1 adds support to some [older kernels](https://github.com/aquasecurity/btfhub/blob/main/docs/supported-distros.md)
+> so user doesn't need to build specific non CO-RE eBPF objects to them.
 
-```
-$ docker image ls
-REPOSITORY           TAG       IMAGE ID       CREATED        SIZE
-tracee-btfhub        latest    1612003190b7   11 hours ago   73.6MB
-tracee-nocore        latest    cc8eba17fd76   11 hours ago   1.19GB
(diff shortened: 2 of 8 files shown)


rafaeldtinoco:
It is recommended that you test executing both container images in an environment that supports CO-RE and another environment that doesn't support it.

@danielpacak be aware that, BTFHUB=1 is NOT the default way of building both container images. This happens because I thought that, during the PR checks, we may want to chose whether we want or not BTFHUB files included, to make things faster to test. So, in order to generate the "official images", we have to prepend `BTFHUB=1` to the "make" command.

rafaeldtinoco:
Ready for review. Thanks!

.github/workflows/pr.yaml
```diff
@@ -168,16 +167,16 @@ jobs:
       - name: Buld tracee-tester image
         run: |
           make -f tests/Makefile docker-build-tracee-tester
-      - name: Build tracee-nocore
+      - name: Build tracee container image (full w/out BTFHUB)
```
rafaeldtinoco, line 170: Full is the old "fat" image, without BTFHUB support means it has to build it during its initialization and place in /tmp/tracee (unless it already exists there, then no build is needed).

.github/workflows/pr.yaml
```diff
@@ -206,20 +205,19 @@ jobs:
       - name: Built tracee BPF
         run: |
           make -f Makefile.one bpf-core
-      - name: Build tracee-core-btfhub
+      - name: Build tracee container image (slim w/ BTFHUB)
         run: |
-          make -f builder/Makefile.tracee-container build-alpine-tracee-core-btfhub
+          BTFHUB=1 make -f builder/Makefile.tracee-container build-tracee
```
rafaeldtinoco, line 210: @danielpacak could you check if BTFHUB=1 was passed to the docker build correctly AND the tracee:latest had, indeed, a tracee-ebpf binary with embedded tailored BTF files on it ? If that is true, we would be testing CO-RE.

I believe it worked because I told "go test" to use "tracee:latest" and it wouldn't have worked without BTFHUB support.

But it could be that the non core eBPF object was still in /tmp/tracee and then we would be testing the non CO-RE again here. Perhaps `rm -rf /tmp/tracee` in between the 2 tests ? Not sure.

rafaeldtinoco: @danielpacak, could you take a look at the failing tests for me ? With this new container images, the smoke tests seem flaky to me. They are passing locally:

```
$ go test -v -run "TestTraceeSignatures" ./tests/tracee_test.go -tracee-image-ref "tracee:latest" -tracee-tester-image-ref "aquasec/tracee-tester:latest"  -tracee-signatures "TRC-2,TRC-3"
=== RUN   TestTraceeSignatures
=== RUN   TestTraceeSignatures/tracee:latest/TRC-2
2022/02/04 01:29:42 Starting container id: e478f0c3163a image: testcontainers/ryuk:0.3.3
2022/02/04 01:29:43 Waiting for container id e478f0c3163a image: testcontainers/ryuk:0.3.3
2022/02/04 01:29:43 Container is ready id: e478f0c3163a image: testcontainers/ryuk:0.3.3
2022/02/04 01:29:43 Starting container id: 971112419f47 image: tracee:latest
2022/02/04 01:29:44 Waiting for container id 971112419f47 image: tracee:latest
2022/02/04 01:29:48 Container is ready id: 971112419f47 image: tracee:latest
2022/02/04 01:30:03 Starting container id: 79c7cd6872d1 image: aquasec/tracee-tester:latest
2022/02/04 01:30:03 Container is ready id: 79c7cd6872d1 image: aquasec/tracee-tester:latest
=== RUN   TestTraceeSignatures/tracee:latest/TRC-3
2022/02/04 01:30:24 Starting container id: 6e05b2a5e190 image: tracee:latest
2022/02/04 01:30:25 Waiting for container id 6e05b2a5e190 image: tracee:latest
2022/02/04 01:30:29 Container is ready id: 6e05b2a5e190 image: tracee:latest
2022/02/04 01:30:29 Starting container id: 5f763d8b7e91 image: aquasec/tracee-tester:latest
2022/02/04 01:30:30 Container is ready id: 5f763d8b7e91 image: aquasec/tracee-tester:latest
--- PASS: TestTraceeSignatures (73.66s)
    --- PASS: TestTraceeSignatures/tracee:latest/TRC-2 (47.16s)
    --- PASS: TestTraceeSignatures/tracee:latest/TRC-3 (26.48s)
PASS
ok  	command-line-arguments	73.673s
```

And looks like they're randomly failing in github actions. I tried raising timeout values, not sure if that helped.

Feel free to rebase this change in another PR of yours, with fixes, if you can find any issue.

grantseltzer (commented):
I've tested these out on multiple VMs across multiple distros and it seems to work well for me. I'll wait to let Daniel have a second look before approving, but nice work!

.github/workflows/pr.yaml
rafaeldtinoco: This has been resolved, hopefully, by https://github.com/aquasecurity/tracee/pull/1439 with the change of pulling the image before the initial container needed it.

@danielpacak and I discovered that docker background daemon was causing too many events when downloading image during the container initialization and the tests were flaky because of that (I'll explain in another case).

For now this was mitigated by making docker to pull the image (before it did not happen because we were building the container test image instead of pulling it during first container initialization).

more to come, for now this can be considered "mitigated" (or should be).

rafaeldtinoco:
@grantseltzer 

I have discussed the flaky tests with @danielpacak so we had a small merge before this PR to address (mitigate) the issue we were seeing (related to timing of events - container preparation - making the tests - TRC2 actually - to fail).

I think that this PR is "ready for review", so you can approve and work on top of it for your release PR. WDYT ? Let me know if you would like something to be changed, then we can merge and I can review the release PR made with these new images.

Thanks a lot for reviewing it.

grantseltzer:
I can't help but feel that "full" isn't descriptive of the difference between the two images. The documentation in `containers.md` doesn't explain that the full image contains the source code and compiler/tool chain either. I suppose `tracee:with-compiler` or `tracee:from-src` would be too long? WDYT? If i'm being pedantic feel free to tell me.

rafaeldtinoco:
> I can't help but feel that "full" isn't descriptive of the difference between the two images. The documentation in `containers.md` doesn't explain that the full image contains the source code and compiler/tool chain either. I suppose `tracee:with-compiler` or `tracee:from-src` would be too long? WDYT? If i'm being pedantic feel free to tell me.

You are definitely not being pedantic. Please do feel free to express yourself freely when reviewing. I used "full" because "fat" seemed awful #). I have no other ideas, but I agree that from-src/with-compiler sounds bad. I'm all in for suggestions.

docs/building/containers.md
```diff
@@ -8,97 +8,71 @@
 
 ## Generating Tracee containers
 
-Tracee containers come in **3 flavors**:
+Tracee containers come in **2 flavors**:
 
-* `tracee`: default [CO-RE eBPF](https://nakryiko.com/posts/bpf-portability-and-co-re/) tracee (portable).
+* `tracee:latest`
```
grantseltzer, line 13:
```suggestion
* `tracee:latest` (CO-RE enabled embedded tracee bpf object)
```

docs/building/containers.md
```diff
@@ -8,97 +8,71 @@
...
 
   ```
-  $ make -f builder/Makefile.tracee-container build-alpine-tracee-core
+  $ BTFHUB={0,1} make -f builder/Makefile.tracee-container build-tracee
   ```
 
-* `tracee-btfhub`: CO-RE eBPF tracee with [older kernels support](https://github.com/aquasecurity/btfhub/blob/main/docs/supported-distros.md).
+* `tracee:full`
```
grantseltzer, line 19:
```suggestion
* `tracee:full` (Contains tracee bpf source code and compiler tool chain) 
```

grantseltzer (approved):
After thinking about it _fully_ (pun intended) I accepted that I can't think of a better tag name. I just want to add documentation for it in the containers.md. Otherwise this is great and i'll have my release PR rebased and updated quickly.

PR #1438, opened 2022-02-03
tracee-ebpf: fix container filters

### LINKED ISSUE #1436: tracee-ebpf does not report events for echoing to file within a container
Hi,

While capturing with tracee-ebpf (-t c), running the following command within the container does not report any event:
`echo "aaa" > some_file.txt`

**IMPORTANT**: this only happens when running the container first, and only then running tracee with the **-t c** argument.
Furthermore, if running tracee without the container scope (-t c), before and after the container started, tracee **does** report the relevant events.

image

yanivagman:
Currently we don't trace existing container processes when a container
filter is set. This behavior is due to using the NewPidFilter when
a container filter is set. As we now use cgroups to identify
containers (and not by checking that pid != host_pid like we used to),
we can remove this extra filter to trace existing container processes
as expected by the filter.

While at it, fix the "!container" filter which currently doesn't work as the events are identified as false container positives.

Fix #1436


mtcherni95 (approved):
LGTM

PR #1444, opened 2022-02-06
Separate pcaps

roikol:
add options to separate captured pcap files by their process context.
two output options were added: 
 - separating by hostPid
 - separating by containerId

in tracee output folder, the captured pcaps would be captured under the containerId folder ("host" is used when containerId from process context is empty), in this format:
 - separating by hostPid:
        <tracee_out>/host/<process_name>_<process_id>_<process_start_time>.pcap
 - separating by containerId:
        <tracee_out>/<containerId>/captured.pcap
- user didn't choose separating the pcaps:
        <tracee_out>/host/captured.pcap

more details about how we separate:

instead of have one pcapWriter for tracee, we now have a map of pcapWriters, where the key is packetContext. packetContext is generated for each packet, and depends on whether the user chose to separate pcap by process_id, container_id, or not separate at all.  with every packet we get the hostTid from the kernel, and the packetContext is generated using the process context from the processTree map.
in order to maintain this map, we would want to delete on process_exit/cotainer_exit (when the user didn't choose to separate, there is only one value in this map, and it shouldn't be deleted). in order to do that, i have added two network event, which will go through the net_events channel - NetProcessExit and NetContainerExit.
note: we might have packets coming after the process had already exited - therefore i delete from the pcapWriters map with a delay of 1 second. also, for the same reason, i delete from processTree with delay of 1 second, as the net_events relay on this map to generate packetContext.

### DIFF
--- a/cmd/tracee-ebpf/internal/flags/flags-capture.go
+++ b/cmd/tracee-ebpf/internal/flags/flags-capture.go
@@ -21,16 +21,17 @@ Possible options:
 [artifact:]mem                     capture memory regions that had write+execute (w+x) protection, and then changed to execute (x) only.
 [artifact:]net=interface           capture network traffic of the given interface. Only TCP/UDP protocols are currently supported.
 
-dir:/path/to/dir        path where tracee will save produced artifacts. the artifact will be saved into an 'out' subdirectory. (default: /tmp/tracee).
-profile                 creates a runtime profile of program executions and their metadata for forensics use.
-clear-dir               clear the captured artifacts output dir before starting (default: false).
+dir:/path/to/dir        			path where tracee will save produced artifacts. the artifact will be saved into an 'out' subdirectory. (default: /tmp/tracee).
+profile                 			creates a runtime profile of program executions and their metadata for forensics use.
+clear-dir               			clear the captured artifacts output dir before starting (default: false).
+net:[per-container|per-process]		capture separate pcap file based on container/process context (default: none - saving one pcap for the entire host).
 
 Examples:
   --capture exec                                           | capture executed files into the default output directory
   --capture exec --capture dir:/my/dir --capture clear-dir | delete /my/dir/out and then capture executed files into it
   --capture write=/usr/bin/* --capture write=/etc/*        | capture files that were written into anywhere under /usr/bin/ or /etc/
   --capture profile                                        | capture executed files and create a runtime profile in the output directory
-  --capture net=eth0                                       | capture network traffic of eth0
+  --capture net=eth0 --capture net:per-container           | capture network traffic of eth0, and save pcap for each container
   --capture exec --output none                             | capture executed files into the default output directory not printing the stream of events
 
 Use this flag multiple times to choose multiple capture options
@@ -42,6 +43,9 @@ func PrepareCapture(captureSlice []string) (tracee.CaptureConfig, error) {
 
 	outDir := "/tmp/tracee"
 	clearDir := false
+
+	netCapturePerContainer := false
+	netCapturePerProcess := false
 
 	var filterFileWrite []string
 	for i := range captureSlice {
@@ -83,6 +87,15 @@ func PrepareCapture(captureSlice []string) (tracee.CaptureConfig, error) {
 			if !found {
 				capture.NetIfaces = append(capture.NetIfaces, iface)
 			}
+		} else if strings.HasPrefix(cap, "net:") {
+			netCaptureContext := strings.TrimPrefix(cap, "net:")
+			if netCaptureContext == "per-container" {
+				netCapturePerContainer = true
+			} else if netCaptureContext == "per-process" {
+				netCapturePerProcess = true
+			} else {
+				return tracee.CaptureConfig{}, fmt.Errorf("invalid network capture option: %s. accepted options - net:per-container or net:per-process", netCaptureContext)
+			}
 		} else if cap == "clear-dir" {
 			clearDir = true
 		} else if strings.HasPrefix(cap, "dir:") {
@@ -103,6 +116,12 @@ func PrepareCapture(captureSlice []string) (tracee.CaptureConfig, error) {
 	if clearDir {
 		os.RemoveAll(capture.OutputPath)
 	}
+
+	if netCapturePerContainer && netCapturePerProcess {
+		return tracee.CaptureConfig{}, fmt.Errorf("invalid capture flags: can't use both net:per-container and net:per-process capture options")
+	}
+	capture.NetPerContainer = netCapturePerContainer
+	capture.NetPerProcess = netCapturePerProcess
 
 	return capture, nil
 }
--- a/cmd/tracee-ebpf/internal/flags/flags-capture.go
+++ b/cmd/tracee-ebpf/internal/flags/flags-capture.go
@@ -24,7 +24,7 @@ Possible options:
 dir:/path/to/dir                    path where tracee will save produced artifacts. the artifact will be saved into an 'out' subdirectory. (default: /tmp/tracee).
 profile                             creates a runtime profile of program executions and their metadata for forensics use.
 clear-dir                           clear the captured artifacts output dir before starting (default: false).
-pcap:[per-container|per-process]     capture separate pcap file based on container/process context (default: none - saving one pcap for the entire host).
+pcap:[per-container|per-process]    capture separate pcap file based on container/process context (default: none - saving one pcap for the entire host).
 
 Examples:
   --capture exec                                           | capture executed files into the default output directory
--- a/cmd/tracee-ebpf/internal/flags/flags-capture.go
+++ b/cmd/tracee-ebpf/internal/flags/flags-capture.go
@@ -21,10 +21,10 @@ Possible options:
 [artifact:]mem                     capture memory regions that had write+execute (w+x) protection, and then changed to execute (x) only.
 [artifact:]net=interface           capture network traffic of the given interface. Only TCP/UDP protocols are currently supported.
 
-dir:/path/to/dir        			path where tracee will save produced artifacts. the artifact will be saved into an 'out' subdirectory. (default: /tmp/tracee).
-profile                 			creates a runtime profile of program executions and their metadata for forensics use.
-clear-dir               			clear the captured artifacts output dir before starting (default: false).
-net:[per-container|per-process]		capture separate pcap file based on container/process context (default: none - saving one pcap for the entire host).
+dir:/path/to/dir                    path where tracee will save produced artifacts. the artifact will be saved into an 'out' subdirectory. (default: /tmp/tracee).
+profile                             creates a runtime profile of program executions and their metadata for forensics use.
+clear-dir                           clear the captured artifacts output dir before starting (default: false).
+net:[per-container|per-process]     capture separate pcap file based on container/process context (default: none - saving one pcap for the entire host).
 
 Examples:
   --capture exec                                           | capture executed files into the default output directory
--- a/cmd/tracee-ebpf/internal/flags/flags-capture.go
+++ b/cmd/tracee-ebpf/internal/flags/flags-capture.go
@@ -24,14 +24,14 @@ Possible options:
 dir:/path/to/dir                    path where tracee will save produced artifacts. the artifact will be saved into an 'out' subdirectory. (default: /tmp/tracee).
 profile                             creates a runtime profile of program executions and their metadata for forensics use.
 clear-dir                           clear the captured artifacts output dir before starting (default: false).
-net:[per-container|per-process]     capture separate pcap file based on container/process context (default: none - saving one pcap for the entire host).
+pcap:[per-container|per-process]     capture separate pcap file based on container/process context (default: none - saving one pcap for the entire host).
 
 Examples:
   --capture exec                                           | capture executed files into the default output directory
   --capture exec --capture dir:/my/dir --capture clear-dir | delete /my/dir/out and then capture executed files into it
   --capture write=/usr/bin/* --capture write=/etc/*        | capture files that were written into anywhere under /usr/bin/ or /etc/
   --capture profile                                        | capture executed files and create a runtime profile in the output directory
-  --capture net=eth0 --capture net:per-container           | capture network traffic of eth0, and save pcap for each container
+  --capture net=eth0 --capture pcap:per-container           | capture network traffic of eth0, and save pcap for each container
   --capture exec --output none                             | capture executed files into the default output directory not printing the stream of events
 
 Use this flag multiple times to choose multiple capture options
@@ -87,14 +87,14 @@ func PrepareCapture(captureSlice []string) (tracee.CaptureConfig, error) {
 			if !found {
 				capture.NetIfaces = append(capture.NetIfaces, iface)
 			}
-		} else if strings.HasPrefix(cap, "net:") {
-			netCaptureContext := strings.TrimPrefix(cap, "net:")
+		} else if strings.HasPrefix(cap, "pcap:") {
+			netCaptureContext := strings.TrimPrefix(cap, "pcap:")
 			if netCaptureContext == "per-container" {
 				netCapturePerContainer = true
 			} else if netCaptureContext == "per-process" {
 				netCapturePerProcess = true
 			} else {
-				return tracee.CaptureConfig{}, fmt.Errorf("invalid network capture option: %s. accepted options - net:per-container or net:per-process", netCaptureContext)
+				return tracee.CaptureConfig{}, fmt.Errorf("invalid network capture option: %s. accepted options - pcap:per-container or pcap:per-process", netCaptureContext)
 			}
 		} else if cap == "clear-dir" {
 			clearDir = true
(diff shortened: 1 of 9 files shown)


tracee-ebpf/tracee/net_events.go
```diff
@@ -30,6 +133,22 @@ func (t *Tracee) processNetEvents() {
...
+			}
+
+			// get process start time
+			procStartTime := networkThread.StartTime
+			if networkThread.HostTid != networkThread.HostPid {
+				networkProcess, err := t.getProcessCtx(networkThread.HostPid)
+				if err == nil {
+					procStartTime = networkProcess.StartTime
+				}
+			}
```
yanivagman, line 148: Instead of performing this lookup for every packet, which is inefficient, we can keep the process start time in the process context of the thread

tracee-ebpf/tracee/net_events.go
```diff
@@ -63,11 +182,11 @@ func (t *Tracee) processNetEvents() {
 						t.handleError(err)
 						continue
 					}
-					networkProcess, err := t.getProcessCtx(int(hostTid))
+
 					hr, min, sec := timeStampObj.Clock()
 					nsec := uint16(timeStampObj.Nanosecond())
 
-					if err != nil {
+					if networkThreadErr != nil {
```
yanivagman, line 189: if networkThreadErr != nil you will never reach this point (as you continue in line 138 above)

tracee-ebpf/tracee/net_events.go
```diff
@@ -109,18 +228,32 @@ func (t *Tracee) processNetEvents() {
...
+				err = t.pcapWriters[packetContext].Flush()
 				if err != nil {
 					t.handleError(err)
 					continue
 				}
+			} else if (t.config.Output.PcapPerProcess && (netEventId == NetProcessExit || netEventId == NetContainerExit)) || (t.config.Output.PcapPerContainer && netEventId == NetContainerExit) {
+				_, pcapWriterExists := t.pcapWriters[packetContext]
+				if pcapWriterExists {
+					go t.netExit(packetContext)
+				}
```
yanivagman, line 256: You can take care of the cleanup while processing sched_process_exit event instead of doing it here, and avoid the extra events sent in the network perf buffer

cmd/tracee-ebpf/internal/flags/flags-output.go
```diff
@@ -92,6 +92,10 @@ func PrepareOutput(outputSlice []string) (tracee.OutputConfig, printerConfig, er
 				outcfg.ParseArguments = true
 			case "sort-events":
 				outcfg.EventsSorting = true
+			case "per-process-pcaps":
+				outcfg.PcapPerProcess = true
+			case "per-container-pcaps":
+				outcfg.PcapPerContainer = true
```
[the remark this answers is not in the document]
roikol: my original thought was that it is somewhat of an output option. but yeah, no problem, it fits under the capture flag

tracee-ebpf/tracee/net_events.go
roikol: agreed - i'll add it to the context.

tracee-ebpf/tracee/net_events.go
roikol: ok. but then this map will be accessed from two different go routines. should i use a lock on this map?

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -3257,6 +3274,14 @@ static __always_inline int net_map_update_or_delete_sock(void* ctx, int event_id
             get_local_net_id_from_network_details_v6(sk, &connect_id, &net_details, family);
     }
 
+    net_debug_t debug_event = {0};
+    debug_event.ts = data->context.ts;
+    debug_event.host_tid = data->context.host_tid;
+    __builtin_memcpy(&debug_event.comm, data->context.comm, TASK_COMM_LEN);
+    debug_event.local_addr = connect_id.address;
+    debug_event.local_port = __bpf_ntohs(connect_id.port);
+    debug_event.protocol = connect_id.protocol;
```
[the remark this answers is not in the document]
roikol: my bad, this are leftovers of somethink i eventually didn't include.

cmd/tracee-ebpf/internal/flags/flags-capture.go
```diff
@@ -21,16 +21,17 @@ Possible options:
...
+clear-dir                           clear the captured artifacts output dir before starting (default: false).
+net:[per-container|per-process]     capture separate pcap file based on container/process context (default: none - saving one pcap for the entire host).
 
 Examples:
   --capture exec                                           | capture executed files into the default output directory
   --capture exec --capture dir:/my/dir --capture clear-dir | delete /my/dir/out and then capture executed files into it
   --capture write=/usr/bin/* --capture write=/etc/*        | capture files that were written into anywhere under /usr/bin/ or /etc/
   --capture profile                                        | capture executed files and create a runtime profile in the output directory
-  --capture net=eth0                                       | capture network traffic of eth0
+  --capture net=eth0 --capture net:per-container           | capture network traffic of eth0, and save pcap for each container
```
OriGlassman, line 34: Sorry for jumping in on this - IMO the arguments for net options are rather confusing. Both have 'net' prefix, but one time it's with '=' for the network, and one time with ':' for the pcap.
Wouldn't the argument for the pcap should be 'pcap'? like so:
--capture net=eth0 --capture **pcap**=[per-container|per-process]

roikol: yes you're right its more accurate. i used `pcap:` instead of `pcap=` because this is more an option rather than a value (although i don't think it is consistent throughout the project)

yanivagman (changes requested):

cmd/tracee-ebpf/internal/flags/flags-capture.go
```diff
@@ -21,16 +21,18 @@ Possible options:
 [artifact:]mem                     capture memory regions that had write+execute (w+x) protection, and then changed to execute (x) only.
 [artifact:]net=interface           capture network traffic of the given interface. Only TCP/UDP protocols are currently supported.
 
-dir:/path/to/dir        path where tracee will save produced artifacts. the artifact will be saved into an 'out' subdirectory. (default: /tmp/tracee).
-profile                 creates a runtime profile of program executions and their metadata for forensics use.
-clear-dir               clear the captured artifacts output dir before starting (default: false).
+dir:/path/to/dir                    path where tracee will save produced artifacts. the artifact will be saved into an 'out' subdirectory. (default: /tmp/tracee).
+profile                             creates a runtime profile of program executions and their metadata for forensics use.
+clear-dir                           clear the captured artifacts output dir before starting (default: false).
+pcap:[per-container|per-process]     capture separate pcap file based on container/process context (default: none - saving one pcap for the entire host).
```
yanivagman, line 27: nit: extra space

pkg/bufferdecoder/decoder.go
```diff
@@ -40,27 +40,29 @@ func (decoder *EbpfDecoder) ReadAmountBytes() int {
...
-	ctx.MntID = binary.LittleEndian.Uint32(decoder.buffer[offset+44 : offset+48])
-	ctx.PidID = binary.LittleEndian.Uint32(decoder.buffer[offset+48 : offset+52])
-	_ = copy(ctx.Comm[:], decoder.buffer[offset+52:offset+68])
-	_ = copy(ctx.UtsName[:], decoder.buffer[offset+68:offset+84])
-	ctx.EventID = int32(binary.LittleEndian.Uint32(decoder.buffer[offset+84 : offset+88]))
-	ctx.Retval = int64(binary.LittleEndian.Uint64(decoder.buffer[offset+88 : offset+96]))
-	ctx.StackID = binary.LittleEndian.Uint32(decoder.buffer[offset+96 : offset+100])
-	ctx.ProcessorId = binary.LittleEndian.Uint16(decoder.buffer[offset+100 : offset+102])
-	ctx.Argnum = uint8(binary.LittleEndian.Uint16(decoder.buffer[offset+102 : offset+104]))
+	ctx.ThreadTs = binary.LittleEndian.Uint64(decoder.buffer[offset+8 : offset+16])
```
yanivagman, line 47: Instaed of Ts (shorthand for Timestamp) I would use StartTime which is more correct

pkg/bufferdecoder/protocol.go
```diff
@@ -20,6 +20,8 @@ const (
 // keep the 1-byte 'Argnum' as the final parameter before the padding (if padding is needed).
 type Context struct {
 	Ts          uint64
+	ThreadTs    uint64
+	ProcTs      uint64
```
yanivagman, line 24: Why do we need this data as part of the context of each event?
This is only required for new processes, isn't it?
If that so, adding it to sched_process_fork/sched_process_exec should have been enough, isn't that so?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2439,23 +2460,21 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
...
-        save_to_submit_buf(&data, (void*)&child_tgid, sizeof(int), 6);
-        save_to_submit_buf(&data, (void*)&child_ns_tgid, sizeof(int), 7);
+    save_to_submit_buf(&data, (void*)&parent_pid, sizeof(int), 0);
+    save_to_submit_buf(&data, (void*)&parent_ns_pid, sizeof(int), 1);
+    save_to_submit_buf(&data, (void*)&parent_tgid, sizeof(int), 2);
+    save_to_submit_buf(&data, (void*)&parent_ns_tgid, sizeof(int), 3);
+    save_to_submit_buf(&data, (void*)&child_pid, sizeof(int), 4);
+    save_to_submit_buf(&data, (void*)&child_ns_pid, sizeof(int), 5);
+    save_to_submit_buf(&data, (void*)&child_tgid, sizeof(int), 6);
+    save_to_submit_buf(&data, (void*)&child_ns_tgid, sizeof(int), 7);
```
yanivagman, line 2475: I'm not sure that we will need to send this event all the time.
@rafaeldtinoco is currently working on something called "janitor events" which should keep the process tree always up to date.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2491,11 +2508,16 @@ int tracepoint__sched__sched_process_exec(struct bpf_raw_tracepoint_args *ctx)
...
+    process_context_t *process = (process_context_t*) &data.context;
+    process->thread_ts = get_task_start_time(task);
+    process->proc_ts = get_process_start_time(task);
+
     // We passed all filters (in should_trace()) - add this pid to traced pids set
     bpf_map_update_elem(&traced_pids_map, &data.context.host_tid, &data.context.host_tid, BPF_ANY);
     bpf_map_update_elem(&process_context_map, &data.context.host_tid, process, BPF_ANY);
 
-    struct task_struct *task = (struct task_struct *)ctx->args[0];
     struct linux_binprm *bprm = (struct linux_binprm *)ctx->args[2];
```
yanivagman, line 2521: Let's move those two together

yanivagman (changes requested):
Looks like most of the new changes from the previous version of this PR is that you added the start time of the thread and process to the event context. This, however, might not be necessary per event and we better only add those to sched_process_fork/sched_process_exec events

pkg/ebpf/c/tracee.bpf.c
roikol: sounds awesome! i had the same thought for a while now

cmd/tracee-ebpf/internal/flags/flags-capture.go
yanivagman: Still here :-)

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2440,23 +2447,22 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
         bpf_map_update_elem(&new_pids_map, &child_pid, &child_pid, BPF_ANY);
     }
 
-    if (event_chosen(SCHED_PROCESS_FORK)) {
```
yanivagman, line 2443: I think we should keep this gate to avoid flooding userspace with process info events.
We can guard with a new config_option in addition to the existing SCHED_PROCESS_FORK, 
WDYT?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2494,7 +2498,12 @@ int tracepoint__sched__sched_process_exec(struct bpf_raw_tracepoint_args *ctx)
 
     // We passed all filters (in should_trace()) - add this pid to traced pids set
     bpf_map_update_elem(&traced_pids_map, &data.context.host_tid, &data.context.host_tid, BPF_ANY);
-    bpf_map_update_elem(&process_context_map, &data.context.host_tid, process, BPF_ANY);
+
+    process_context_t *process = bpf_map_lookup_elem(&process_context_map, &data.context.host_tid);
+    if (process != NULL) {
+        __builtin_memcpy(process->comm, data.context.comm, TASK_COMM_LEN);
+        bpf_map_update_elem(&process_context_map, &data.context.host_tid, process, BPF_ANY);
```
yanivagman, line 2505: I don't think that it is required to do this map update.
You have a pointer to process, which updates the entry without a required update.
Please check.

pkg/ebpf/events_processor.go
```diff
@@ -131,20 +140,13 @@ func (t *Tracee) processEvent(event *external.Event) error {
...
-			Ppid:        uint32(event.ParentProcessID),
-			HostTid:     uint32(event.HostThreadID),
-			HostPid:     uint32(event.HostProcessID),
-			HostPpid:    uint32(event.HostParentProcessID),
-			Uid:         uint32(event.UserID),
-			MntId:       uint32(event.MountNS),
-			PidId:       uint32(event.PIDNS)}
-		t.procInfo.UpdateElement(event.HostThreadID, processData)
+		//update the process tree with correct comm name
+		processData, err := t.procInfo.GetElement(event.HostThreadID)
```
yanivagman, line 144: For clarity, I think we should use HostProcessID instead (pid should be equal to tid at that point)

pkg/ebpf/events_processor.go
```diff
@@ -235,7 +237,14 @@ func (t *Tracee) processEvent(event *external.Event) error {
 			return err
 		}
 	case SchedProcessExitEventID:
-		t.procInfo.DeleteElement(event.HostThreadID)
+		if t.config.Capture.NetPerProcess {
```
yanivagman, line 240: You don't guard in all places with NetPerProcess, and in the future, we might have other reasons to have a process tree that are not related to net events. Let's remove this line, and guard in bpf code instead

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -404,7 +404,7 @@ typedef struct event_context {
 } context_t;
 
 typedef struct process_context {
-    u64 ts;                        // Timestamp
+    u64 start_time;                // Timestamp
```
yanivagman, line 407: The comment is now invalid

yanivagman (changes requested):

pkg/ebpf/events_processor.go
roikol: i only check for NetPerProcess to know if i should close the pcap. 
deleting from the process tree is done anyway for this event.
i'll guard in the bpf code, but i think this check for NetPerProcess should stay here.

pkg/ebpf/events_processor.go
roikol: yes they should be equal at that point. but the goal is to update comm for each thread in the process tree. every operation on the process tree is done on the context of the current thread (not process). don't you think?

yanivagman: If you only update comm in exec, it doesn't matter if you take thread or process, as they are equal at that point.
To update comm in thread resolution we should also monitor prctl event with command to change thread name

roikol: ok, i'll use pid for clarity

pkg/ebpf/net_events.go
```diff
@@ -25,8 +31,156 @@ type CaptureData struct {
...
+func (t *Tracee) getPcapContext(hostTid uint32) (processPcapId, procinfo.ProcessCtx, error) {
+
+	networkThread, err := t.getProcessCtx(hostTid)
+	if err != nil {
+		return processPcapId{}, procinfo.ProcessCtx{}, err
+	}
+	networkProcess, err := t.getProcessCtx(networkThread.HostPid)
+	if err != nil {
+		return processPcapId{}, procinfo.ProcessCtx{}, err
+	}
```
yanivagman, line 162: In case hostTid wasn't found, this error will repeat many times (for every packet), right?
This might cause error counter to have very big values, without understanding why exactly.
One thing I would change here is adding more context to the error, instead of just sending it as is, which might be unclear when trying to understand the problem.
Another thing to consider is to continue after sending the error (calling handleError directly instead of return)and just fill the missing fields with emty values. The advantage of doing that is that we save all the packets, even when there is an error of this kind, so we don't loose information about the network traffic

yanivagman (changes requested):

pkg/ebpf/net_events.go
```diff
@@ -5,6 +5,12 @@ import (
 	gocontext "context"
 	"encoding/binary"
 	"fmt"
+	"github.com/google/gopacket/layers"
+	"github.com/google/gopacket/pcapgo"
```
yanivagman, line 9: Please move those imports below

yanivagman (approved):
Congratulations @roikol, LGTM!

PR #1445, opened 2022-02-06
Bugfix/head with next upon lost events

### LINKED ISSUE #1441: CPU queueing has a BUG: HEAD with next detected
During some event loss tests, when enabling CPU queueing/sorting feature, I came across the message that @AlonZivony introduced during our CPU queueing feature review:

```
INFO: probing tracee-ebpf capabilities...
INFO: starting tracee-ebpf...
INFO: starting tracee-rules...

*** Detection ***
Time: 2022-02-04T19:33:42Z
Signature ID: TRC-2
Signature: Anti-Debugging
Data: map[]
Command: strace
Hostname: focal

lost 10052 events

*** Detection ***
Time: 2022-02-04T19:34:03Z
Signature ID: TRC-2
Signature: Anti-Debugging
Data: map[]
Command: strace
Hostname: focal

lost 6275 events
lost 27641 events
lost 116737 events
lost 8831 events

BUG: HEAD has next
BUG: HEAD has next
```

AlonZivony:
Should fix #1441.
@rafaeldtinoco I did not manage to reproduce your error message so if you can run this branch and check that it is not reoccurring it would be great.

### DIFF
--- a/pkg/events/sorting/cpu_queue.go
+++ b/pkg/events/sorting/cpu_queue.go
@@ -15,7 +15,10 @@ type cpuEventsQueue struct {
 
 // InsertByTimestamp insert new event to the queue in the right position according to its timestamp
 func (cq *cpuEventsQueue) InsertByTimestamp(newEvent *external.Event) error {
-	newNode := cq.pool.Alloc(newEvent)
+	newNode, allocErr := cq.pool.Alloc(newEvent)
+	if allocErr != nil {
+		cq.pool.Reset()
+	}
 
 	cq.mutex.Lock()
 	defer cq.mutex.Unlock()
@@ -31,7 +34,10 @@ func (cq *cpuEventsQueue) InsertByTimestamp(newEvent *external.Event) error {
 				break
 			}
 			if insertLocation.next == insertLocation {
-				return fmt.Errorf("encountered node with self reference at next")
+				err := fmt.Errorf("encountered node with self reference at next")
+				if allocErr != nil {
+					err = fmt.Errorf("%s\n%s", allocErr, err)
+				}
 			}
 			insertLocation = insertLocation.next
 		}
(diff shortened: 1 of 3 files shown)


AlonZivony:
Oh, need to rebase

mtcherni95 (commented):
Can we add add a test that catches the bug scenario?

AlonZivony:
So I did not manage to reproduce the bug.
If @rafaeldtinoco will be able to explain to me the full scenario of the bug maybe it will be possible.
However, from what it seems to me right now it is not a easy test to add (it won't be a UT), so it does not belong to this bugfix in my opinion.
Unless you want a UT that create the scenario? Because the BUG message here announce that unpredictable bug found. To me it seems unnecessary, but tell me what do you think.

rafaeldtinoco (approved):
Without the patch, I am able to reproduce the bug:

```
$ sudo ./dist/tracee-ebpf -o format:json -o option:parse-arguments -o option:sort-events --trace event=close,dup,dup2,dup3,execve,init_module,magic_write,mem_prot_alert,process_vm_writev,ptrace,sched_process_exec,sched_process_exit,security_bprm_check,security_file_open,security_kernel_read_file,security_sb_mount,security_socket_connect | pv -pterb | dd of=/dev/null
BUG: HEAD has next
```

in a 5.13.0-22-generic kernel (focal), 4 cpus/4gb ram, after a few seconds running that test case. 

Testing this PR seems to fix the issue. It looks like it was, indeed, a leftover of not clearing the bi-directional node connection.

Code looks good! I'm approving the fix.

BTW, as a feedback to you @AlonZivony, the throughput of your algorithm was very close to non sorting one (for my workload, 10MB/s with and without the sorting code). This is a good sign.

AlonZivony:
Thanks @rafaeldtinoco for checking.
I am glad to hear about the throughput :)

Issue #1447, opened 2022-02-07
tracee-ebpf : move the process context to different package


itamarmaouda101:
hi 
to able Tracee be more flexible we should extract the process Context feature to different package.
this will help to create to #1363 by giving access to the process tree more easily

yanivagman:
Done #1448

PR #1448, opened 2022-02-07
move the process context to different package

itamarmaouda101:
hi 
this  PR will solve #1447 by moving the process context to internal package under Tracee


tracee-ebpf/tracee/events_processor.go
```diff
@@ -2,6 +2,7 @@ package tracee
 
 import (
 	"fmt"
+	"github.com/aquasecurity/tracee/tracee-ebpf/tracee/internal/processContext"
```
yanivagman, line 5: Please move this import after the standard imports

tracee-ebpf/tracee/tracee.go
```diff
@@ -3,10 +3,12 @@ package tracee
 import (
 	gocontext "context"
 	"crypto/sha256"
+	"encoding/binary"
 	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/aquasecurity/tracee/tracee-ebpf/tracee/internal/processContext"
```
yanivagman, line 11: Please move this import below

tracee-ebpf/tracee/tracee.go
```diff
@@ -935,8 +937,9 @@ func (t *Tracee) getProcessCtx(hostTid int) (ProcessCtx, error) {
 		if err != nil {
 			return processCtx, err
 		}
-		processCtx, err = t.ParseProcessContext(processCtxBpfMap)
-		t.processTree.processTreeMap[hostTid] = processCtx
+		containerId := t.containers.GetCgroupInfo(binary.LittleEndian.Uint64(processCtxBpfMap[8:16])).ContainerId
+		processCtx, err = processContext.ParseProcessContext(processCtxBpfMap, containerId)
```
yanivagman, line 941: The fact that you now decode the container id here makes calling this parsing function irrelevant.
Instead you can pass the containers instance (t.containers) and let the function handle it internaly.

yanivagman (approved):
LGTM

yanivagman:
I think that this package should be moved to pkg instead of being internal.
As it uses the decoder package, I understand that this requires some change.
@mtcherni95 can you please have a look for how (and if) we can/should use the decoder out of the tracee (ebpf) package?

pkg/proctree/process_context.go
```diff
@@ -29,30 +28,23 @@ type ProcessCtx struct {
...
-	errs = append(errs, decoder.DecodeUint32(&procCtx.HostPpid))
-	errs = append(errs, decoder.DecodeUint32(&procCtx.Uid))
-	errs = append(errs, decoder.DecodeUint32(&procCtx.MntId))
-	errs = append(errs, decoder.DecodeUint32(&procCtx.PidId))
-	for _, e := range errs {
-		if e != nil {
-			return procCtx, e
-		}
-	}
+	procCtx.ContainerID = containers.GetCgroupInfo(cgroupId).ContainerId
```
yanivagman, line 38: Please verify that ctx byte array has at least 52 bytes. This is instead of the previous error checking we had

pkg/proctree/process_context.go
```diff
@@ -29,30 +28,26 @@ type ProcessCtx struct {
...
 type ProcessTree struct {
-	processTreeMap map[int]ProcessCtx
+	ProcessTreeMap map[int]ProcessCtx
 }
 
-func (t *Tracee) ParseProcessContext(ctx []byte) (ProcessCtx, error) {
+func ParseProcessContext(ctx []byte, containers *containers.Containers) (ProcessCtx, error) {
 	var procCtx = ProcessCtx{}
+	if len(ctx) < 52 {
+		return procCtx, fmt.Errorf("Error: ctx byte array to small %v", len(ctx))
```
yanivagman, line 37:
```suggestion
		return procCtx, fmt.Errorf("can't read process context: buffer too short")
```

pkg/proctree/process_context.go
```diff
@@ -29,30 +28,26 @@ type ProcessCtx struct {
 }
 
 type ProcessTree struct {
-	processTreeMap map[int]ProcessCtx
+	ProcessTreeMap map[int]ProcessCtx
 }
```
yanivagman, line 32: Let's not make this struct public (use uncapitalized struct and field names).
We should use getters/setters to access and update process info.
You can even take the changes introduced in #1452 into this PR and to make these getters/setters thread safe

yanivagman (approved):

Issue #1451, opened 2022-02-07
tracee-ebpf :  bug in the process ctx update 


itamarmaouda101:
Hi
I notice that in case we run Tracee with high amount of process creating and existing there is a bug that because Tracee to panic .
the bug I because of the race-conation in the process tree changes.
we should add mutex helpers to the process tree map

this is the output when the panic happen:
[err.txt](https://github.com/aquasecurity/tracee/files/8017154/err.txt)

PR #1452, opened 2022-02-07
Add safe update to process context

itamarmaouda101:
solves #1451 
adding 2 wrappers for changes the process context map entries


yanivagman:
Already done at #1448

PR #1453, opened 2022-02-08
tracee-ebpf: fix cgroup v1 multiple container_create event

### LINKED ISSUE #1432: container_create event received multiple times on cgroup v1 envs
When running tracee and choosing the new container_create event, multiple events are received on a cgroup v1 environment .
The reason for that is that cgroup_mkdir, from which container_create event is derived, is sent per each cgroup controller

yanivagman:
Fix #1432

### DIFF
--- a/pkg/containers/containers.go
+++ b/pkg/containers/containers.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"strconv"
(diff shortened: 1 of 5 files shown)


rafaeldtinoco (approved):
LGTM. Filtering a single cgroup controller to avoid duplicates seem like a good logic.

PR #1454, opened 2022-02-08
Merge external and types into a single module

AsafEitani:
* merge external and types into a single module - addresses #1312 
* change external.Event to external.TraceeEvent - as a preparation to address to #1431 

This PR will follow with another PR with the necessary modifications to `tracee/go.mod` and the changes to the code - mainly imports and change external.Event to external.TraceeEvent

Issue #1431, opened 2022-02-02
Generic types.Event overhead


NDStrahilevitz:
# The issue itself

In tracee-rules, the existing `Event` type is simply the generic `interface{}`. This has implications on the performance of tracee-rules and it's cpu time demand. During the performance analysis that @mtcherni95 and I are performing, we've noticed that convT2E can take from 10-20% of overall cpu time. 
In our case, since we're consuming tracee-rules and tracee-ebpf as libraries (single binary), that has the further effect of taking cpu time from tracee-ebpf, thus increasing the chance of event loss.

The simplest way to remove this overhead would be to replace `external.Event` with`types.Event` everywhere. This eliminates the problem entirely, and in our performance tests introduced performance gains (shown in the performance analysis below). However it did mean breaking the Signature interface API, which obviously means this change can't be taken too lightly.

If it's agreed to pursue this agenda I see two reasonable ways to go about it:

1. As said before - replacing `types.Event` with `external.Event` everywhere. This means loss of generality with the gain of efficiency, however since practically all signatures type check for `external.Event` anyway, and considering there is logic in tracee-rules tightly dependent on `external.Event` the loss of generality isn't very "real". This way can also be done as a first step towards a next more general solution
2. Immediately reimplementing the generality concept. In discussions I've had on this topic the idea of adding an `EventHeader` type was raised, which we could immediately supply along with tracee-ebpf events, and enrich other event sources if we want to add them to tracee-rules (unlike now where we would have to hard code their support). Signatures could then support events according to a header and not through type cast checking.
3. There are probably more ways to go about it, I'd like to use this issue to discuss them :)

This is a big issue to tackle but I think it's worth it in terms of performance and in establishing a better implementation of consuming events from various sources.

# Performance analysis:

There are 3 images showing the point of the optimization (running on the previously mentioned single binary tracee).

1. Flame graph with cpu usage of tracee.processEvents (buffered signature channels(see #1429) and no convT2E optimization).
![image](https://user-images.githubusercontent.com/22661609/152170479-a5241c94-f269-42f1-8cf2-b706ace82626.png)
2. Flame graph with cpu usage of tracee.processEvents (buffered signature channels(see #1429) and including convT2E optimization). As can be seen there was an ~8% cpu usage increase for tracee-ebpf.
![image](https://user-images.githubusercontent.com/22661609/152170555-79d06512-af40-4bc8-9398-8d289514f647.png)
[...]

Issue #1312, opened 2021-12-27
Go signatures aren't usable outside of Tracee tree


itaysk:
If someone wants to build their own Go signatures, following our instructions to create a standalone Go project and compile it with `-buildmode=plugin` the resulting artifact (.so) will not load in tracee-rules.
The reason is that the loader/linker checks for dependency version compatibility by comparing versions, but the hosting module (in our case tracee-rules) has no version. This is an upstream issue in Go: https://github.com/golang/go/issues/31354 .

The reason it worked for us so far is that the signatures are compiled within the same directory as the tracee, so the import is treated as local. A workaround to make Go treat the import as local is to `replace` the module path in go.mod. 

The GH issue indicates that the are no plans to address this for at least another year so the proposed solution for us is to extract tracee-rules/types into a separate Go module that both tracee-rules and the signatures would import.

thanks @AsafEitani for investigating

pomozoff:
Awesome!

danielpacak:
Before we extract tracee-rules/types somewhere I wanted to confirm that this solution really works and apparently it worked in this little POC

https://github.com/danielpacak/vulnerability-scanner
https://github.com/danielpacak/vulnerability-scanner-types
https://github.com/danielpacak/vulnerability-plugin-tomcat

@itaysk when you said "extract tracee-rules/types into a separate Go module" did you mean creating a new https://github.com/aquasecurity/tracee-types repository or Go module somewhere within the existing https://github.com/aquasecurity/tracee/types/go.mod repository?

yanivagman:
Shouldn't been closed yet

danielpacak:
I created [this](https://github.com/danielpacak/tracee-go-signatures) Git template repository to get started quickly with custom Go signatures and I'm still bumping into similar issue, this time with pkg/external:

```
udo ./dist/tracee-ebpf \
  --output=format:gob \
  --output=option:parse-arguments \
  | ./dist/tracee-rules \
  --rules-dir /home/vagrant/tracee-go-signatures/dist \
  --input-tracee=file:stdin \
  --input-tracee=format:gob
2022/01/26 09:01:12 error opening plugin /home/vagrant/tracee-go-signatures/dist/tracee-go-signatures.so: plugin.Open("/home/vagrant/tracee-go-signatures/dist/tracee-go-signatures"): plugin was built with a different version of package github.com/aquasecurity/tracee/pkg/external
Loaded 0 signature(s): []
```

[...]

PR #1455, opened 2022-02-08
Use the merged external and types module

AsafEitani:
completes #1454, solves #1312 and prepares for #1431

### DIFF
--- a/cmd/tracee-ebpf/main_test.go
+++ b/cmd/tracee-ebpf/main_test.go
@@ -3,12 +3,13 @@ package main
 import (
 	"errors"
 	"fmt"
+	"io/ioutil"
+	"testing"
+
 	"github.com/aquasecurity/tracee/cmd/tracee-ebpf/internal/flags"
 	"github.com/aquasecurity/tracee/tracee-ebpf/tracee"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"io/ioutil"
-	"testing"
 )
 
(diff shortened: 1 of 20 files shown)


NDStrahilevitz:
Personally, I think instead of going by this route:
#1460 (asaf pr) -> #1493 (my types pr) -> #1482 (event header refactor pr)

we could just do

#1493 -> #1482

or maybe even

1493 -> 1460 -> 1482

This is mostly so we can finalize the types for now. Either way, i'm fine with this, but I will note that in #1493 I've changed `trace.TraceeEvent` back to `trace.Event` (package name made me feel ok to do it).

AsafEitani:
IDK how we ended up doing the same thing twice, but it doesn't really matters to me.
I do think that we should keep `TraceeEvent` as it really represent that, and not any generic event that we will have in the future from other sources. It will make it easier to differentiate between detect and trace.

The only thing that matters to me is that either this PR or #1482 will be merged soon.
Too bad we did double the work.

PS. @grantseltzer just merged the final restructure and this PR will need a massive rebase, which is exactly why I wanted to merge both PRs one after the other.
@NDStrahilevitz We can go with whatever you choose as long as we will be able to compile signatures separately from tracee ASAP.

NDStrahilevitz:
About the name: how about `EbpfPayload`?

About the PR path we take I think let's just go with the shortest route which is #1493 -> #1482

NDStrahilevitz:
@AsafEitani since #1482 basically does the same thing and was just merged I think this can be closed.

PR #1458, opened 2022-02-08
fix usage of rwlock under procinfo package and moving bufferdecoder under pkg

mtcherni95:
This commit moves `bufferdecoder` under `pkg`. It also fixes usage of reader writer lock in `procinfo` package and other smaller fixes.


pkg/procinfo/procinfo.go
```diff
@@ -59,15 +64,34 @@ func ParseProcessContext(ctx []byte, containers *containers.Containers) (Process
...
+	}
+	if err := decoder.DecodeUint32(&procCtx.Uid); err != nil {
+		return ProcessCtx{}, err
+	}
+	if err := decoder.DecodeUint32(&procCtx.MntId); err != nil {
+		return ProcessCtx{}, err
+	}
+	if err := decoder.DecodeUint32(&procCtx.PidId); err != nil {
+		return ProcessCtx{}, err
+	}
```
yanivagman, line 94: What's the advantage in using the decoder here?
If you defined the struct in the decoder protocol, that would have been another case, but now the code is longer, and no real advantage here.

pkg/procinfo/procinfo.go
```diff
@@ -162,14 +186,20 @@ func NewProcessInfo() (*ProcInfo, error) {
...
-	// Iterate over each pid
+	// iterate over each pid
 	for _, procEntry := range entries {
 		pid, err := strconv.ParseUint(procEntry, 10, 32)
 		if err != nil {
 			continue
 		}
 		taskDir, err := os.Open(fmt.Sprintf("/proc/%d/task", pid))
+		if err != nil {
+			return nil, err
```
yanivagman, line 197: better use continue.
It might be that some task exited while looping procfs. We don't want to fail tracee in such a case.

pkg/procinfo/procinfo.go
```diff
@@ -162,14 +186,20 @@ func NewProcessInfo() (*ProcInfo, error) {
...
+		}
 		processTasks, err := taskDir.Readdirnames(-1)
+		if err != nil {
+			return nil, err
```
yanivagman, line 201: continue

yanivagman (changes requested):

pkg/procinfo/procinfo.go
yanivagman: no need to log. Just continue

yanivagman (approved):

PR #1460, opened 2022-02-08
Finish tracee repo restructure

grantseltzer:
This moves tracee-ebpf and tracee-rules directories into `pkg` and removes unneeded build files


yanivagman:
I suggest making the clean in one PR and another to move the tracee package under pkg. The first can be reviewed quickly while the second is a moving target

grantseltzer:
> I suggest making the clean in one PR and another to move the tracee package under pkg. The first can be reviewed quickly while the second is a moving target

You got it! https://github.com/aquasecurity/tracee/pull/1466

grantseltzer:
@yanivagman This is ready for review when you get a chance

yanivagman:
I didn't fully review this yet, but from a quick glance:
1. Maybe we should call the package ebpf instead of bpf (tracee-ebpf/tracee -> pkg/ebpf)?
2. I think we should create a dedicated folder for the C sources (which can be called, maybe, simply 'c'?). This folder can be placed under the ebpf package, and include tracee.bpf.c, as well as the files which are currently located at the co-re folder (we can flatten co-re and put its content into the 'c' folder). WDYT?

yanivagman:
Also, I see that `packaging/ubuntu/copyright` file needs to be upated as well as it contains:
`Files: tracee/tracee-ebpf/tracee.bpf.c`

packaging/ubuntu/copyright
```diff
@@ -1,8 +1,8 @@
-Files: tracee/tracee-ebpf/tracee.bpf.c
+Files: tracee/pkg/ebpf/c/tracee.bpf.c
 Copyright: 2022, Aqua Security
 License: GPL-2+
 
-Files: tracee/tracee-ebpf/co-re/*
+Files: tracee/tracee-ebpf/c/*
```
yanivagman, line 5: Should probably be `tracee/pkg/ebpf/c/*` and then the above `tracee/pkg/ebpf/c/tracee.bpf.c` copyright can be removed

Makefile
```diff
@@ -451,7 +451,7 @@ GO_ENV_EBPF += GOARCH=$(GO_ARCH)
 GO_ENV_EBPF += CGO_CFLAGS=$(CUSTOM_CGO_CFLAGS)
 GO_ENV_EBPF += CGO_LDFLAGS=$(CUSTOM_CGO_LDFLAGS)
 
-TRACEE_EBPF_SRC_DIRS = ./cmd/tracee-ebpf/ ./tracee-ebpf/ ./pkg/
+TRACEE_EBPF_SRC_DIRS = ./cmd/tracee-ebpf/ ./pkg/ebpf
```
yanivagman, line 454: Are you sure it is ok to remove `./pkg`?

embedded-ebpf.go
```diff
@@ -1,7 +1,7 @@
 //go:build ebpf
 // +build ebpf
 
-package tracee
+package ebpf
```
yanivagman, line 4: can you please remind me why we have to keep the embedded* files in the root dir and not move it under pkg/ebpf?

embedded.go
```diff
@@ -1,4 +1,4 @@
-package tracee
+package ebpf
 
 import _ "embed"
```
yanivagman, line 3: Why do we even need this file?
It only contains embed for rego helpers, which are not used in the ebpf (tracee) packge, are they?

embedded-ebpf.go
danielpacak: Go embed package does not support relative paths like './../somebpf.o'

yanivagman: Yes, I just saw the issue saying exactly that.
But it seems that there is a workaround for this (which is not portable to windows, but that is not a problem for us): https://github.com/golang/go/issues/46056#issuecomment-938251415

danielpacak: Do we need yet another workaround? :)

yanivagman: TBH I don't know...
It just looks ugly to have this go file in the root directory

embedded.go
rafaeldtinoco: we needed a file that would only embed things if "ebpf" build flag is given. The other will always embed things. This has been a result of the merge of tracee-ebpf and tracee-rules. Please make sure that files are correctly embedded if this is changed (BTFHUB=1 and check if btfhub is being embedded).

embedded-ebpf.go
yanivagman: > because of ./dist/btfhub path. it needs to be in root because embed does not accept "../../" path.

Yes, @danielpacak already mentioned this - see my reply above

embedded.go
yanivagman: > we needed a file that would only embed things if "ebpf" build flag is given. The other will always embed things. This has been a result of the merge of tracee-ebpf and tracee-rules. Please make sure that files are correctly embedded if this is changed (BTFHUB=1 and check if btfhub is being embedded).

But what do you embed here? currently nothing (maybe it was used when we have to embed the C sources before we removed it from main.go? If so, this file can be removed as well)

grantseltzer: Yea i'm also a bit confused about the use of `embedded.go` in the root directory. I changed it to 'package ebpf' with search/replace, but now that I think about it, it's not imported anywhere, is it? It embeds the signautres/rego/helpers.rego when built but I'm not sure it's actually being built.

rafaeldtinoco: Related commits:

commit 82a1289c
Date:   Mon Aug 9 17:32:35 2021

    tracee-ebpf: add support to custom rego helpers

and

commit c22f59cd
Author: Simarpreet Singh <simar@linux.com>
Date:   Tue Mar 9 19:14:40 2021

    feat: Use //go:embed to bundle artifacts (#596)

grantseltzer: Ok that makes sense. After testing it I reverted back to keeping the embedded go files as the 'github.com/aquasecurity/tracee' package

yanivagman: I don't think I follow...
Why do we need this file?

itaysk:
I know I'm late for this party, but just one minor suggestion to consider renaming `package ebpf` to something like `package trace`. 1) it's not really about ebpf interface, which is in libbpfgo, it's more about how to use this for tracing. see for example cilium/ebpf. 2) ebpf doesn't signal the intent of the package, it just signals which technology is bring used to implement this intent.
If you find it too hard/controversial just ignore this comment since merging this is more important :)

grantseltzer:
> I know I'm late for this party, but just one minor suggestion to consider renaming `package ebpf` to something like `package trace`. 1) it's not really about ebpf interface, which is in libbpfgo, it's more about how to use this for tracing. see for example cilium/ebpf. 2) ebpf doesn't signal the intent of the package, it just signals which technology is bring used to implement this intent. If you find it too hard/controversial just ignore this comment since merging this is more important :)

I agree with your sentiment that `ebpf` isn't a good name because it's a package for interfacing with ebpf. Though I think `trace` can be confused with `tracee` easily. What do you think about `tracer` or `tracing`?

yanivagman:
> I know I'm late for this party, but just one minor suggestion to consider renaming `package ebpf` to something like `package trace`. 1) it's not really about ebpf interface, which is in libbpfgo, it's more about how to use this for tracing. see for example cilium/ebpf. 2) ebpf doesn't signal the intent of the package, it just signals which technology is bring used to implement this intent. If you find it too hard/controversial just ignore this comment since merging this is more important :)

The way I see it, package ebpf will serve both tracing and enforcement in the future. For this reason, I think that a more neutral name such as eBPF is a better fit. Such a package will be responsible for:
1. eBPF programs and maps initialization
2. updates to ebpf maps and programs
3. Perf buffer initialization

These stuff are already most of what tracee.go is composed of, with some exceptions. After creating the eBPF package, we can take out other stuff from there into packages

grantseltzer:
> > I know I'm late for this party, but just one minor suggestion to consider renaming `package ebpf` to something like `package trace`. 1) it's not really about ebpf interface, which is in libbpfgo, it's more about how to use this for tracing. see for example cilium/ebpf. 2) ebpf doesn't signal the intent of the package, it just signals which technology is bring used to implement this intent. If you find it too hard/controversial just ignore this comment since merging this is more important :)
> 
> The way I see it, package ebpf will serve both tracing and enforcement in the future. For this reason, I think that a more neutral name such as eBPF is a better fit. Such a package will be responsible for:
> 
> 1. eBPF programs and maps initialization
> 2. updates to ebpf maps and programs
> 3. Perf buffer initialization
> 
> These stuff are already most of what tracee.go is composed of, with some exceptions. After creating the eBPF package, we can take out other stuff from there into packages

I suppose it's also named by `github.com/aquasecurity/tracee/pkg/ebpf` so it's clear it's within the context of tracee. If we're in agreement, is there anything preventing this from being approved/merged?

yanivagman:
LGTM

PR #1469, opened 2022-02-10
Introduce net_packet as regular event

itamarmaouda101:
Hi 
This PR Introduce a firstly support to convert incoming network events to regular events by sending them to the `event channel`.

Note: this Introduce will added that support only to the `net_packet` event, later on ill create another PR that add this ability to the other network events


yanivagman (changes requested):

tracee-ebpf/tracee/net_events.go
```diff
@@ -130,36 +96,36 @@ func (t *Tracee) processNetEvents() {
 
 				switch evtMeta.NetEventId {
 				case DebugNetSecurityBind:
-					fmt.Printf("%v  %-16s  %-7d  debug_net/security_socket_bind LocalIP: %v, LocalPort: %d, Protocol: %d\n",
-						timeStampObj, evtMeta.ProcessName, evtMeta.HostTid, netaddr.IPFrom16(pkt.LocalIP), pkt.LocalPort, pkt.Protocol)
+					fmt.Printf("%v  %-16s  %-7d  debug_net/security_socket_bind LocalIP: %s, LocalPort: %d, Protocol: %d\n",
+						timeStampObj, evtMeta.ProcessName, evtMeta.HostTid, parseIP(pkt.LocalIP), pkt.LocalPort, pkt.Protocol)
```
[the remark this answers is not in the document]
itamarmaouda101: the previous `netaddr.IPFrom16()` used the `netaddr` pkg function which is good, but I thought to replace it with my implantation that using also the net utilities helpers (in the next PR will use them with all the net events when we will parse ip

tracee-ebpf/tracee/net_events.go
```diff
@@ -175,15 +141,19 @@ func (t *Tracee) processNetEvents() {
 	}
 }
 
-func (t *Tracee) writePacket(packetLen uint32, timeStamp time.Time, interfaceIndex int, dataBuff *bytes.Buffer) error {
+func (t *Tracee) writePacket(capData CaptureData, timeStamp time.Time, dataBuff *bytes.Buffer) error {
```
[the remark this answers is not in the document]
itamarmaouda101: I think it will be more efficient to use struct because it firstly reduce the number of args in the function and also  in the next PR ill add a parsing to that struct by `binary.read`.
let me know what you think

tracee-ebpf/tracee/net_events.go
```diff
@@ -34,79 +52,27 @@ func (t *Tracee) processNetEvents() {
...
+					evt, packet := netPacketProtocolHandler(dataBuff, evtMeta, networkProcess, "net_packet", t.bootTime)
+					captureData.PacketLen = packet.PktLen
+					captureData.InterfaceIndex = packet.IfIndex
+					if err := t.writePacket(captureData, time.Unix(int64(evtMeta.TimeStamp), 0), dataBuff); err != nil {
+						t.handleError(err)
+						continue
+					}
+					select {
+					case t.config.ChanEvents <- evt:
+						t.stats.eventCounter.Increment()
```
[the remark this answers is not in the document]
itamarmaouda101: sounds good , you want ill add it in his PR or in another one ?

tracee-ebpf/tracee/net_events.go
```diff
@@ -206,3 +176,48 @@ func parseEventMetaData(payloadBytes []byte) (EventMeta, *bytes.Buffer) {
...
+	CreateNetPacketMetaArgs(&evt, packet)
+	evt.EventName = eventName
+	evt.Timestamp = int(evtMeta.TimeStamp + bootTime)
+	return evt, packet
+}
+
+// parsing the PacketMeta struct from bytes.buffer
+func ParseNetPacketMetaData(payload *bytes.Buffer) (PacketMeta, error) {
+	var pktMetaData PacketMeta
+	err := binary.Read(payload, binary.LittleEndian, &pktMetaData)
```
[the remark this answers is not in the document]
mtcherni95: I would add `PackMeta` as part of `bufferdecoder` protocol, and add the relevant decoding function (with unit tests and benchamarks). WDYT?

tracee-ebpf/tracee/net_events.go
```diff
@@ -206,3 +176,48 @@ func parseEventMetaData(payloadBytes []byte) (EventMeta, *bytes.Buffer) {
...
+		ArgMeta: external.ArgMeta{"metadata", "external.PktMeta"},
+		Value:   arg,
+	}
+	eventArgs = append(eventArgs, evtArg)
+	return eventArgs
+}
+
+func CreateNetPacketMetaArgs(event *external.Event, NetPacket PacketMeta) {
+	event.Args = createNetPacketMetadataArg(NetPacket)
+	event.ArgsNum = len(event.Args)
```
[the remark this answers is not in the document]
itamarmaouda101: Now it will be always 1, but after the dns will be it will be more then 1.
Because of that in dans there the `pktMeta` and after that the dns info (requests and responses)

tracee-ebpf/tracee/net_events.go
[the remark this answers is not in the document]
itamarmaouda101: TBH IDK , because in the following PR we will created package that handles the all network events and also adding new events (such detection for http , irc,  dns and icmp ) so for every protocol will have a specific parser (that  parse form byte array to the asked protocol struct).
WDYT?

pkg/procinfo/procinfo.go
```diff
@@ -55,6 +56,22 @@ func (p *ProcInfo) GetElement(hostTid int) (ProcessCtx, error) {
 	return processCtx, nil
 }
 
+func (ctx *ProcessCtx) GetEventByProcessCtx() external.Event {
+	var event external.Event
```
mtcherni95, line 60:
```suggestion
	return external.Event {
	ContainerID: ctx.ContainerId,
	ProcessId : int(ctx.Pid),
	// and so on...
	}
```

pkg/external/network_external.go
```diff
@@ -0,0 +1,9 @@
+package external
+
+type PktMeta struct {
+	SrcIP    string `json:"src_ip"`
+	DestIP   string `json:"dest_ip"`
+	SrcPort  uint16 `json:"src_port"`
+	DestPort uint16 `json:"dest_port"`
```
itamarmaouda101, line 7: dst*

tracee-ebpf/tracee/net_events.go
```diff
@@ -34,79 +52,33 @@ func (t *Tracee) processNetEvents() {
 			timeStampObj := time.Unix(0, int64(evtMeta.TimeStamp+t.bootTime))
 
 			if evtMeta.NetEventId == NetPacket {
-				var pktLen uint32
-				err := binary.Read(dataBuff, binary.LittleEndian, &pktLen)
+				networkProcess, err := t.getProcessCtx(int(evtMeta.HostTid))
```
itamarmaouda101, line 55: move it after the packet capture data resolved

tracee-ebpf/tracee/net_events.go
```diff
@@ -34,79 +52,33 @@ func (t *Tracee) processNetEvents() {
...
+				} else {
+					// To get the current ("wall") time, we add the boot time into it.
+					evtMeta.TimeStamp += t.bootTime
 				}
-				idx, ok := t.ngIfacesIndex[int(ifindex)]
-				if !ok {
+				evt, packet := netPacketProtocolHandler(dataBuff, evtMeta, networkProcess, "net_packet")
+				var captureData CaptureData
+				captureData.PacketLen = packet.PktLen
+				captureData.InterfaceIndex = packet.IfIndex
```
itamarmaouda101, line 70: add function to the capture data and add err in the return

tracee-ebpf/tracee/net_events.go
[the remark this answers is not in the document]
itamarmaouda101: remove

tracee-ebpf/tracee/net_events.go
```diff
@@ -206,3 +182,54 @@ func parseEventMetaData(payloadBytes []byte) (EventMeta, *bytes.Buffer) {
...
+	evt := ctx.GetEventByProcessCtx()
+	evt.Timestamp = int(eventMeta.TimeStamp)
+	evt.ProcessName = eventMeta.ProcessName
+	evt.EventID = int(eventMeta.NetEventId)
+	evt.EventName = eventName
+	evt.Timestamp = ts
+	return evt
+}
+
+func netPacketProtocolHandler(buffer *bytes.Buffer, evtMeta EventMeta, ctx procinfo.ProcessCtx, eventName string) (external.Event, PacketMeta) {
```
itamarmaouda101, line 196: add DOCO

pkg/procinfo/procinfo.go
```diff
@@ -55,6 +56,22 @@ func (p *ProcInfo) GetElement(hostTid int) (ProcessCtx, error) {
...
+		ThreadID:            int(ctx.Tid),
+		ParentProcessID:     int(ctx.Ppid),
+		HostProcessID:       int(ctx.HostPid),
+		HostThreadID:        int(ctx.HostTid),
+		HostParentProcessID: int(ctx.HostPpid),
+		UserID:              int(ctx.Uid),
+		MountNS:             int(ctx.MntId),
+		PIDNS:               int(ctx.PidId),
+	}
+
```
yanivagman, line 72: nit: extra line

tracee-ebpf/tracee/net_events.go
```diff
@@ -30,83 +38,39 @@ func (t *Tracee) processNetEvents() {
...
 			evtMeta, dataBuff := parseEventMetaData(in)
 
-			// timeStamp is nanoseconds since system boot time
 			timeStampObj := time.Unix(0, int64(evtMeta.TimeStamp+t.bootTime))
 
-			if evtMeta.NetEventId == NetPacket {
-				var pktLen uint32
-				err := binary.Read(dataBuff, binary.LittleEndian, &pktLen)
-				if err != nil {
+			captureData, err := parseCaptureData(dataBuff)
```
yanivagman, line 43: Capture data is not available when only `t.config.Debug` is set and not `--capure net` - reading it here will break this flow

yanivagman (changes requested):

tracee-ebpf/tracee/net_events.go
```diff
@@ -196,13 +165,77 @@ func (t *Tracee) writePacket(packetLen uint32, timeStamp time.Time, interfaceInd
...
+		Value: external.PktMeta{
+			SrcIP:    NetPacket.SrcIP,
+			DstIP:    NetPacket.DstIP,
+			SrcPort:  NetPacket.SrcPort,
+			DstPort:  NetPacket.DstPort,
+			Protocol: NetPacket.Protocol,
+		},
+	}}
+	event.ArgsNum = 1
 }
```
yanivagman, line 241: Consider combining these two functions for clarity

yanivagman (changes requested):

tracee-ebpf/tracee/net_events.go
```diff
@@ -30,83 +38,40 @@ func (t *Tracee) processNetEvents() {
...
-					t.handleError(err)
-					continue
-				}
-				idx, ok := t.ngIfacesIndex[int(ifindex)]
-				if !ok {
-					t.handleError(err)
-					continue
+				captureData, err := parseCaptureData(dataBuff)
+				if err == nil && len(t.config.Capture.NetIfaces) > 0 {
+					if err := t.writePacket(captureData, time.Unix(int64(evtMeta.TimeStamp), 0), bytes.NewBuffer(dataBuff)); err != nil {
```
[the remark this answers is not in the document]
itamarmaouda101: Ok
Firstly, I changed it to `payloadBytes`.
Secondly, the reason I return a byte array is because now we aren't need the `bytes.NewBuffer()` anymore (besides in the writePacket and in the debug events which in the next PR will be use the byte array instead the `newBuffer`).
BTW another reason to use a byte array instead of the `newBuffer `is because if in the packet parsing we now using the the `binary.littleendian.uint` function which use a byte array and not a `newBuffer`

tracee-ebpf/tracee/net_events.go
```diff
@@ -28,85 +36,44 @@ func (t *Tracee) processNetEvents() {
...
-						DestIP   [16]byte
-						SrcPort  uint16
-						DestPort uint16
-						Protocol uint8
-						_        [3]byte //padding
-					}
-					err = binary.Read(dataBuff, binary.LittleEndian, &pktMeta)
+					networkProcess, err := t.getProcessCtx(int(evtMeta.HostTid))
 					if err != nil {
 						t.handleError(err)
```
yanivagman, line 65: Do we really want to treat this as an error in case process context wasn't found? It's different from the previous behavior were we just omit some fields in such a case.

yanivagman (approved):

PR #1470, opened 2022-02-11
events_pipeline: add queueEvents to cache pipeline events

rafaeldtinoco:
Currently tracee-rules has a throughput ratio of 1:4 (events consumed versus events being produced by tracee-ebpf). This requires tracee-ebpf to deal with this lag but, at the same time, the perf buffer events consumption cannot be left behind (or we might lose important events coming from kernel and miss signatures detections).

One way to solve this would be to increase the ChanEvents channel buffer size so, if events being printed weren't consumed as fast as they were being produced, the channel would cache them until they can be consumed, releasing the pressure over the perf buffer.

Unfortunately, only increasing the ChanEvents buffer wasn't enough to stop causing event losses. There is a need for a FIFO QUEUE in the pipeline, so events are cached and released according to speed of the end-of-pipeline consumption.

For now, this patch introduces in the pipeline a queue of up to 1 million events. This has proved to significantly reduce event losses during host load bursts, avoid signature detection losses.

It has been tested in a 2GB environment (and higher): memory allocation and de-allocation behaved well (~1GB consumption until the buffer was full, when events started being lost again, something that is expected).

New tracee-ebpf requirement: 2GB of ram minimum.

Until we know good default cache sizes, based on performance benchmarks of different host configurations, this option won't be enabled by default. Idea is, in near future, to give end user the power of controlling the size of this queue. It will be a game of how much memory you are willing to spend versus how many event loss you are able to accept.

Also, Future work might include queue watermarks advertisement for the eBPF programs. This way, based on the queue usage, eBPF programs might chose to send events (or not) based on their priorities. With that, the pressure over the perf buffer would be reduced as the cache is consumed.

### DIFF
--- a/builder/entrypoint.sh
+++ b/builder/entrypoint.sh
@@ -70,7 +70,13 @@ run_tracee_rules() {
     events=$($TRACEE_RULES_EXE --list-events)
 
     echo "INFO: starting tracee-ebpf..."
-    ${TRACEE_EBPF_EXE} --metrics --output=format:gob --output=option:parse-arguments --trace event=${events} --output=out-file:${TRACEE_PIPE} &
+    ${TRACEE_EBPF_EXE} \
+        --metrics \
+        --output=format:gob \
+        --output=option:parse-arguments \
+        --output=option:cache-events \
+        --trace event=${events} \
+        --output=out-file:${TRACEE_PIPE} &
     tracee_ebpf_pid=$!
 
     # start tracee-rules
--- a/cmd/tracee-ebpf/internal/flags/flags-output.go
+++ b/cmd/tracee-ebpf/internal/flags/flags-output.go
@@ -29,6 +29,7 @@ option:{stack-addresses,detect-syscall,exec-env,relative-time,exec-hash,parse-ar
   exec-hash                                        when tracing sched_process_exec, show the file hash(sha256) and ctime
   parse-arguments                                  do not show raw machine-readable values for event arguments, instead parse into human readable strings
   sort-events                                      enable sorting events before passing to them output. This will decrease the overall program efficiency.
+  cache-events                                     enable caching events to release perf-buffer pressure. This will decrease amount of event loss until cache is full.
 Examples:
   --output json                                            | output as json
   --output gotemplate=/path/to/my.tmpl                     | output as the provided go template
@@ -92,6 +93,8 @@ func PrepareOutput(outputSlice []string) (tracee.OutputConfig, printerConfig, er
 				outcfg.ParseArguments = true
 			case "sort-events":
 				outcfg.EventsSorting = true
+			case "cache-events":
+				outcfg.EventsCaching = true
 			default:
 				return outcfg, printcfg, fmt.Errorf("invalid output option: %s, use '--output help' for more info", outputParts[1])
 			}
--- a/cmd/tracee-ebpf/main.go
+++ b/cmd/tracee-ebpf/main.go
@@ -156,7 +156,7 @@ func main() {
 				return fmt.Errorf("failed preparing BPF object: %w", err)
 			}
 
-			cfg.ChanEvents = make(chan external.Event)
+			cfg.ChanEvents = make(chan external.Event, 10000) // TODO: sync buffer size with queueEventsBufferSize setup
 			cfg.ChanErrors = make(chan error)
 
 			t, err := tracee.New(cfg)
--- a/cmd/tracee-ebpf/main_test.go
+++ b/cmd/tracee-ebpf/main_test.go
@@ -1278,8 +1278,17 @@ func TestPrepareOutput(t *testing.T) {
 			expectedError: nil,
 		},
 		{
+			testName:    "option cache-events",
+			outputSlice: []string{"option:cache-events"},
+			expectedOutput: tracee.OutputConfig{
+				ParseArguments: true,
+				EventsCaching:  true,
+			},
+			expectedError: nil,
+		},
+		{
 			testName:    "all options",
-			outputSlice: []string{"option:stack-addresses", "option:detect-syscall", "option:exec-env", "option:exec-hash", "option:sort-events"},
+			outputSlice: []string{"option:stack-addresses", "option:detect-syscall", "option:exec-env", "option:exec-hash", "option:sort-events", "option:cache-events"},
 			expectedOutput: tracee.OutputConfig{
 				StackAddresses: true,
 				DetectSyscall:  true,
@@ -1287,6 +1296,7 @@ func TestPrepareOutput(t *testing.T) {
 				ExecHash:       true,
 				ParseArguments: true,
 				EventsSorting:  true,
+				EventsCaching:  true,
 			},
 			expectedError: nil,
 		},
--- a/tracee-ebpf/tracee/events_pipeline.go
+++ b/tracee-ebpf/tracee/events_pipeline.go
@@ -2,6 +2,7 @@ package tracee
 
 import (
 	"bytes"
+	"container/list"
 	gocontext "context"
 	"encoding/binary"
 	"fmt"
@@ -21,9 +22,19 @@ const maxStackDepth int = 20
 func (t *Tracee) handleEvents(ctx gocontext.Context) {
 	var errcList []<-chan error
 
+	// cache in here could be []bytes
+
 	// Source pipeline stage.
 	eventsChan, errc := t.decodeEvents(ctx)
 	errcList = append(errcList, errc)
+
+	if t.config.Output.EventsCaching {
+		// Event asynchronous queue. Placed after decodeEvents so events that
+		// does not require processing aren't kept in queue, reducing the memory
+		// footprint. TODO: turn queueEventsBufferSize configurable.
+		eventsChan, errc = t.queueEvents(ctx, eventsChan)
+		errcList = append(errcList, errc)
+	}
 
 	if t.config.Output.EventsSorting {
 		eventsChan, errc = t.eventsSorter.StartPipeline(ctx, eventsChan)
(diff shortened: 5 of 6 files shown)


rafaeldtinoco:
The test case to prove this works is the following... In a virtual machine with 4 vCPUs and 2GB ram, run the commands:

1. In one terminal run tracee-ebpf (with and without "cache-events" option). Without having it you will start receiving lost events from perf buffer soon. With cache events you will realize RSS of trace-ebpf grows to ~1.1GB before start losing events.

```
$ sudo ./dist/tracee-ebpf -o format:json -o option:parse-arguments -o option:cache-events --trace event=close,dup,dup2,dup3,execve,init_module,magic_write,mem_prot_alert,process_vm_writev,ptrace,sched_process_exec,sched_process_exit,security_bprm_check,security_file_open,security_kernel_read_file,security_sb_mount,security_socket_connect | ./dist/tracee-rules --input-tracee file:stdin --input-tracee format:json
```

2. In a second terminal execute:
```
$ while true; do ps -ef; done
```
You may stop once events are being lost.

3. You may execute:
```
$ strace ls
```
Before the buffer gets full and see that TRC-2 ALWAYS work. It won't work if the cache is full (it might or might not, depending on the events being lost).

rafaeldtinoco:
@danielpacak 

BTW, this will make smoke tests not to be flaky (option:cache-events is being added to entrypoint.sh as well). With the default cache/buffer, it is very likely that workload happening in github actions nodes won't make perf events to be lost, allowing TRC-2 detection to happen each time there is a PR (and a smoke test).

yanivagman:
> New tracee-ebpf requirement: 2GB of ram minimum.

Isn't it too much to request a mimimum of 2GB ram?
Tracee is supposed to run inside a container, which might have limited resources, including memory

mtcherni95 (commented):
Great initiative! Some questions:
1. you said the by increasing ChanEvents we would get the desired effect. Why so? How does it differ from what you implemented here? a Go channel respects the FIFO behaviour, so I don't see any difference between increasing the size of ChanEvents and adding a cache layer. Actually as a gut feeling, the current implementation should add computational complexity as it requires other two go routines and:
2. lock contention. We are using a lock which is used by two go routines and both are trying to write on the defended data structure. Having lock per se is not an issue, I'm more worried about the contention on it.
3. I see todos like benchmarking, I think if we go with this solution then would be better to add them now. WDYT?

mtcherni95 (commented):
Also, I am bit concerned with the memory requirement. As a first thought, and probably not the best solution, maybe instead of saving the events in a channel/queue like data structure (indeed memory), we could save to disk. From the other side we would have higher IO. This could be solved by a miscellaneous solution:
1. save in memory X MB of data (much lower than 2 GB)
2. if we got over X of data in cache, then dump to file.

rafaeldtinoco:
> 1. you said the by increasing ChanEvents we would get the desired effect. Why so? How does it differ from what you implemented here? a Go channel respects the FIFO behaviour, so I don't see any difference between increasing the size of ChanEvents and adding a cache layer. 

I have not digged into channels implementation to see why it did not provided desired throughout, what I did was to experiment it and see it caused event losses. 

My feeling is that channel buffering is based on poll events or similar logic and scheduler time slice for so many descriptors selection isn't big enough to let them to be handled in a satisfying manner (we can go into that direction if you prefer but I don't see why right now, as we need to fix something pontual first. This, btw, is a rough guess on the why.

> Actually as a gut feeling, the current implementation should add computational complexity as it requires other two go routines and:

Like you said in our last talk, logic couldn't be slower than what we currently have in tracee-rules.

> 2. lock contention. We are using a lock which is used by two go routines and both are trying to write on the defended data structure. Having lock per se is not an issue, I'm more worried about the contention on it.

To be able to control memory increase and decrease I had to dereference the event pointer and turn it into nil, then add it to the buffer and turn it into new after printed, again based on experimentation (keeping only the obj pointer in the queue made memory behavior erratic, after this the memory would grow and shrink as expected)

> 3. I see todos like benchmarking, I think if we go with this solution then would be better to add them now. WDYT? 

I'd like to add a small cache now, before the benchmark, as we are loosing very simple detections in the smoke tests. Don't have to be 1GB, but we need some.

rafaeldtinoco:
> Also, I am bit concerned with the memory requirement. As a first thought, and probably not the best solution, maybe instead of saving the events in a channel/queue like data structure (indeed memory), we could save to disk.

Okay but then you have to have in mind you need sustained 8MB/s of throughout (12MBs of production minus 4 MBs of consumption) to the disk (not difficult to have) and that you are still using page cache for the recent events (kernel would commit the page cache to the disk based in page cache dirty ratio watermark and timings). 

What I mean is that, in a low memory system, if disk cant catch up with sustained throughput, the dirty ratio would trigger sync of the page cache causing sys time of tracee-ebpf to be used by the pdflush thread to commit the I/O causing much higher event looses.

I don't dislike the idea, but I think it should be an addend to the cache idea (done in a second step), as it have more implications to be measured.

 From the other side we would have higher IO. This could be solved by a miscellaneous solution:
> 1. save in memory X MB of data (much lower than 2 GB)
> 2. if we got over X of data in cache, then dump to file.

Yep, that is what I meant, but with potential big implications to smaller environments as well.

rafaeldtinoco:
@yanivagman, @michaelkatch, @itaysk 

> > 2\. if we got over X of data in cache, then dump to file.

BTW, for this case, instead of reinventing the wheel, we should rely in an already existing queue mechanism (kafka, rabbitmq, etc). The queue mechanism would be the one keeping LRU events cached in-memory (or persistent in disk), etc. 

I'm not saying this should be the default... I'm saying that we could have:

1. No cache (default, and needs improvement)
2. Small cache (256MB, could be default or not, would allow some host cpu bursts)
3. Medium cache (in-memory cache, sized by benchmarks) directly in tracee-ebpf AND could be tied to ebpf event priorities.
4. Big cache (in-memory cache, sized by benchmarks), in tracee-ebpf, but consumed by  a routine that would persist events to an external queue... to be consumed by tracee-rules.

The (4) option would allow us to do remote signatures processing.

> Note: I know that tracee-ebpf and tracee-rules were designed to be executed with low CPU/MEM footprint AND that events could/should be missed. I just think that it is not realistic, currently, not to have a big cache in between tracee-ebpf and tracee-rules if we want detections to happen in a trustful manner (even assuming we could miss some).

rafaeldtinoco:
> @yanivagman, @michaelkatch, @itaysk
> 
> > > 2. if we got over X of data in cache, then dump to file.
> 
> BTW, for this case, instead of reinventing the wheel, we should rely in an already existing queue mechanism (kafka, rabbitmq, etc). The queue mechanism would be the one keeping LRU events cached in-memory (or persistent in disk), etc.
> 
> I'm not saying this should be the default... I'm saying that we could have:
> 
> 1. No cache (default, and needs improvement)
> 2. Small cache (256MB, could be default or not, would allow some host cpu bursts)
> 3. Medium cache (in-memory cache, sized by benchmarks) directly in tracee-ebpf AND could be tied to ebpf event priorities (to perf buffer pressure by reducing amount of events sent).
> 4. Big cache (in-memory cache, sized by benchmarks), in tracee-ebpf, but consumed by  a routine that would persist events to an external queue... to be consumed by tracee-rules.
> 
> The (4) option would allow us to do remote signatures processing in a centralized environment.
> 
> > Note: I know that tracee-ebpf and tracee-rules were designed to be executed with low CPU/MEM footprint AND that events could/should be missed. I just think that it is not realistic, currently, not to have a big cache in between tracee-ebpf and tracee-rules if we want detections to happen in a trustful manner (even assuming we could miss some).

rafaeldtinoco:
Rebased on top of recent changes. Pushed again.

rafaeldtinoco:
After syncing with @itaysk, I have tested GOB output only, with 1M events being cached, and it turns out that memory footprint is pretty much the same, as we're keeping external.Event object in memory (independently of the way it is printed).

The ration of 4:1 was kept by using GOB (though the throughput was 6MB/s production to 1.5MB/s consumption, smaller than with JSON when they were 12MB/s to 3MB/s).

image

And, after the queue is consumed, memory consumption is reduced, like expected:

image

So, independently of the output format, the main issue (of ratio being 4:1) is still there. Just as FYIO.

rafaeldtinoco:
@mtcherni95 and @NDStrahilevitz 

This is the SUMMARY:

# Tracee scalability test:

For a 8 CPU, 2GB environment test: Throughput scalability for tracee-ebpf. 

> Note: Workload confined to a single CPU (#7).
> Using json as output format (spits out more data, easier to observe)
> Using 1M events buffer

## tracee-ebpf only (piped to /dev/null):

* Confining tracee-ebpf to 1 x CPU (0): average of **14MB/s**
* Confining tracee-ebpf to 2 x CPUs (0,1): average of **14MB/s**
* Confining tracee-ebpf to 4 x CPUs (0,1,2,3): average of **13MB/s**
* Confining tracee-ebpf to 6 x CPUs (0...5): average of **13MB/s**

> tracee-ebpf does not seem to be held back by CPU availability.
> throughput is reduced a bit by context switching in more than 2 CPUs.

## tracee-ebpf piped to tracee-rules

Picking best scenario for tracee-ebpf: confined to CPU 0

* Confining tracee-rules to 1 x CPU 1 (0): average **1.20MB/s** (lost events at 1m and 5 seconds)
* Confining tracee-rules to 2 x CPUs (1,2): average **2.4MB/s** (lost events at 1m and 10 seconds)
* Confining tracee-rules to 3 x CPUs (1,2,3): average **2.85MB/s** (lost events at 1m and 10 seconds)
* Confining tracee-rules to 4 x CPUs (1..4): average **2.85MB/s** (lost events at 1m and 20 seconds)

> lost events happen at the same time because the buffer is filled in the same speed, it is drained faster or slower (that is what gives us the average throughput, measuring the speed the events come out of tracee-ebpf because of tracee-rules consumption).
> tracee-rules need 3 CPUs to reach its maximum throughput, and its throughput is 4x slower than what tracee-ebpf is able to consume.

Summary of what this means:

If, in an 8 CPU node, we have a sustained workload of 1 CPU utilization (13% average utilization of this same node), we will fill the 1GB buffer in 1 minute and some seconds). After that, we will start loosing events and detections.

> OBS: GOB output format shows similar results

rafaeldtinoco:
# GOLANG signatures only

When testing golang signatures only (107 of them), the ratio
production:consumption is different. Let's go with numbers...

This is the SUMMARY:

# Tracee scalability test:

For a 8 CPU, 2GB environment test: Throughput scalability for tracee-ebpf. 

> Note: Workload confined to a single CPU (7).
> Using json as output format (spits out more data, easier to observe)
> Using 1 million events buffer
> NOW: Using GOLANG signatures ONLY.
> NOW: -o option:detect-syscall

## tracee-ebpf only (piped to /dev/null):

* Confining tracee-ebpf to 1 x CPU (0): average of **14MB/s**

> no changes from previous test (as expected)

## tracee-ebpf piped to tracee-rules

> tracee-rules now has now 107 golang (only) signatures in place: [TRC-2 TRC-109 TRC-9 TRC-42 TRC-36 TRC-43 TRC-47 TRC-59 TRC-26 TRC-95 TRC-12 TRC-3 TRC-111 TRC-112 TRC-18 TRC-20 TRC-10 TRC-34 TRC-86 TRC-28 TRC-54 TRC-51 TRC-41 TRC-99 TRC-24 TRC-25 TRC-80 TRC-38 TRC-49 TRC-60 TRC-27 TRC-19 TRC-103 TRC-22 TRC-71 TRC-101 TRC-4 TRC-94 TRC-17 TRC-102 TRC-11 TRC-13 TRC-5 TRC-74 TRC-62 TRC-67 TRC-76 TRC-106 TRC-8 TRC-21 TRC-40 TRC-23 TRC-15 TRC-39 TRC-16 TRC-117 TRC-122 TRC-66 TRC-35 TRC-105 TRC-7 TRC-107 TRC-63 TRC-58 TRC-108 TRC-64 TRC-57 TRC-91 TRC-48 TRC-104 TRC-88 TRC-89 TRC-90 TRC-14 TRC-29 TRC-45 TRC-82 TRC-84 TRC-100 TRC-55 TRC-37 TRC-110 TRC-83 TRC-33 TRC-78 TRC-6 TRC-61 TRC-30 TRC-96 TRC-97 TRC-98 TRC-113 TRC-114 TRC-115 TRC-116 TRC-121 TRC-118 TRC-119 TRC-120 TRC-1 TRC-124 TRC-123 TRC-125 TRC-127 TRC-128 TRC-129 TRC-130]

Picking best scenario for tracee-ebpf: confined to CPU 0

* Confining tracee-rules to 1 x CPU 1 (1): average **7 MB/s** (lost events at 1m and 45 seconds)
* Confining tracee-rules to 2 x CPUs (1,2): average **6.80 MB/s** (lost events at 1m and 45 seconds)
* Confining tracee-rules to 3 x CPUs (1,2,3): average **6.50 MB/s** (lost events at 1m and 50 seconds ...)

> We improved the throughput of tracee-rules by using golang signatures only (from **2.85 MB/s** to **7 MB/s**)
> We have now a ratio of 14:7 (2:1) and we are using 1 CPU to tracee-ebpf and 1 CPU to tracee-rules
> It is likely that, in this scenario, by paralelizing tracee-rules signature processing, we may increase the consumption throughput and make it to have a 1:1 ratio.

Summary of what this means:

If, in an 8 CPU node, we have a sustained workload of 1 CPU utilization (13% average utilization of this same node), we will fill the 1GB buffer in 1 minute and 45 seconds). Tracee will need 2 full CPUs (1 for event generation, 1 for event consumption). After that time, we will start loosing events and detections. (For this particular workload. For workloads using more USER %CPU time, this might not be true, but serves as a guidance).

> OBS: GOB output format shows similar results

rafaeldtinoco:
I believe we still need a cache in tracee-ebpf, but, for now, while we are not offloading events to an external queue, for example, we may do something as simple as suggested at: https://github.com/aquasecurity/tracee/pull/1484

PR #1474, opened 2022-02-12
makefile: remove leftovers from root dir

rafaeldtinoco:
- remove old Dockerfile
- remove old entrypoint.sh
- move Vagrantfile to builder/


rafaeldtinoco:
Smoke tests are failing because of we are missing a cache being implemented at:

https://github.com/aquasecurity/tracee/pull/1470

yanivagman (approved):

yanivagman:
Just saw that we missed .dockerignore as well

rafaeldtinoco:
> Just saw that we missed .dockerignore as well

No it is needed! It ignores btfhub files when creating the build container.

Issue #1314, opened 2021-12-28
Incorporate tutorials into docs


itaysk:
Currently we have just one type of docs which are "reference" documentation. We should add a "how-to/tutorials" section with goal-driven content. We can start by drawing already available content:

https://blog.aquasec.com/kubernetes-runtime-security-ebpf
https://blog.aquasec.com/tracee-runtime-malware-alerts-aqua-postee
https://blog.aquasec.com/runtime-security-tracee-rules
https://github.com/krol3/workshop-cloud-native-security/blob/main/runtime.md

yanivagman:
We can probably add these two as well:
https://blog.aquasec.com/linux-vulnerabilitie-tracee
https://blog.aquasec.com/ebpf-container-tracing-malware-detection

krol3:
Nice issue! I would like to help!

PR #1475, opened 2022-02-13
feat: add docs further-reading

krol3:
Related the issue https://github.com/aquasecurity/tracee/issues/1314, adding a section "further Reading", like the other projects starboard and trivy.


rafaeldtinoco:
@AnaisUrlichs would you mind reviewing this PR ? I think @krol3 is proposing something you recently mentioned to us, but I'm not sure this is what you had in mind (regarding place, formatting, etc). Thanks!

mkdocs.yml
```diff
@@ -21,14 +21,15 @@ nav:
...
-  - FAQ: faq.md
   - Architecture: architecture.md
   - Tracing:
     - Getting Started: tracee-ebpf/index.md
     - Output Options: tracee-ebpf/output.md
     - Trace Options: tracee-ebpf/trace-options.md
     - Capturing Artifacts: tracee-ebpf/capture.md
     - Overriding OS needed files: tracee-ebpf/override-os-needed-files
+  - FAQ: faq.md
+  - Further Reading: further-reading.md
```
AnaisUrlichs, line 32: Could we put "Further Reading" within a section called "References"? We are trying to follow this framework and this would be a great first step to implementing it 💯 https://diataxis.fr/

itaysk: I really like the framework but I think reference documentation related to something else, also in the framework explanation. 

I think they don't mean "references" like extra things you could check out. Rather "the reference documentation" like the complete description of all the functionalities and options.

In the doc they say that reference documentation is "dry information" like "encyclopedia article"

AnaisUrlichs: Yes, you are right.
What do you think of having a community section in each docs?
E.g. the structure could look something like this

docs

itaysk: personally I  think that the best option is to add these external link in the tutorials or how-to sections. If someone wrote a good tutorial, we should make it discoverable in the docs right where all other tutorials are.

itaysk: I think that most tutorials are written in blogs, not in the docs, so the docs' tutorials section will be almost exclusively external links anyway (also we do this, see the aqua blog)

AnaisUrlichs: @itaysk tutorials in the tutorial section in the docs have a different purpose to tutorials in a blog -- usually you provide additional explanations in blog posts that should not be part of a tutorial in the docs 
Separately, I like the suggestion of highlighting external tutorials in the tutorials section

krol3:
@itaysk @AnaisUrlichs for a first release about more documentation in Tracee. Do you have a request change in this PR? or do you prefer to close this?

AnaisUrlichs (approved):

rafaeldtinoco:
@itaysk I'll let you merge this one (based on your discussions with @AnaisUrlichs).

AnaisUrlichs:
I think we should merge it and then I can reorganise the docs a little afterwards @itaysk thoughts?

itaysk:
LGTM! My only comment is that "reference" might be confusing heading. I am starting to think we should have just 2 types of docs : reference and tutorials. (Reference docs is like the Trivy manual and tutorial are how to solve specific scenarios). If we don't want to make the switch in this PR that's fine

rafaeldtinoco:
@krol3 can you rebase this and address comments ? I need to either merge or close this! Thanks!

rafaeldtinoco:
Closing this PR as it is pretty much dead right now. Please open a new one if really needed.

PR #1478, opened 2022-02-13
tracee-ebpf: remove string store map

yanivagman:
string_store map is no longer necessary (used as a workaround for compilation errors on older clang/libbpf versions)
Remove it.


rafaeldtinoco (approved):
LGTM

PR #1480, opened 2022-02-15
Use HTTPS link instead of SSH link for git submodules

Phat3:
Changed `libbpf` submodule link to use HTTPS instead of SSH so the `tracee` git repository can be cloned recursively even on machines that do not have an SSH public key registered on Github.


mtcherni95:
I remember you faced also a similar issue, right @YuviGold :)

rafaeldtinoco (approved):
LGTM. Thanks for the contribution.

PR #1482, opened 2022-02-16
Event header refactor

NDStrahilevitz:
Solves #1431 

# The solution

This PR solves two issues with one stone:
1. Lack of a proper message structure for `tracee-rules`.
2. The runtime overhead of directly sending `interface{}` to `tracee-rules`

In addition, this PR removes the parsed events logic (@danielpacak from our talk you mentioned it is unnecessary at this point so I took the liberty of removing it in this scope).

We do it by redefining the `Event` type to this:
```
//EventHeaders are headers attached to the Event struct, used to send metadata about the payload
type EventHeaders struct {
	// ContentType indicates the content of the Event payload.
	ContentType string
	// Origin indicates whether Event originates from host or container.
	Origin string
	// Custom additional custom headers, nil most of the time
	Custom map[string]string
}

//Event is a generic event that the Engine can process
type Event struct {
	Headers EventHeaders
	Payload interface{}
}
```

Doing this somehow avoids the convT2E allocation (pprof flame graph below to demonstrate)

Obviously changing the type requires further changes throughout `tracee-rules` which this PR also provides.
When the PR will be removed from draft I will introduce a prerequisite PR for the `types` changes and remove the reference to my fork from `go.mod`. Alternatively @AsafEitani could possibly introduce my changes in his PR merging external and types.

# Current issues with the PR

This PR is marked as a draft because it currently "plasters" over the previous solution. I believe that moving to this new Event struct will allow us to possibly further streamline the `SignatureEventSelector` system, possibly removing it and replacing it with simple header checks.

In addition, neither @mtcherni95 or me is currently very sure of why this solves the convT2E issue. 
My theory is that by simply assigning to the struct, we avoid the later allocations that were necessary since we are simply moving a struct with some pointer value, but i'm not 100% convinced this is actually the case, as such I would like to hear yours.

# Flame graphs

The problem of convT2E is more thoroughly demonstrated in the performance analysis in the issue here (#1431). The next flame graphs will only demonstrate that this pr solves it.

image

And seemingly the convT2E moved here (with significantly reduced cpu time)

image


go.mod
```diff
@@ -78,9 +78,12 @@ require (
...
 	golang.org/x/text v0.3.6 // indirect
+	golang.org/x/time v0.0.0-20210723032227-1f47c861a9ac // indirect
 	google.golang.org/genproto v0.0.0-20210602131652-f16073e35f0c // indirect
 	google.golang.org/grpc v1.38.0 // indirect
 	google.golang.org/protobuf v1.27.1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 )
+
+replace github.com/aquasecurity/tracee/types => github.com/NDStrahilevitz/tracee/types v0.0.0-20220215182747-608b834f9ebe
```
danielpacak, line 89: This I guess is just for testing?

types/types.go
```diff
@@ -49,6 +58,6 @@ type SignalSourceComplete string
 //Finding is the main output of a signature. It represents a match result for the signature business logic
 type Finding struct {
 	Data        map[string]interface{}
-	Context     Event
+	Context     interface{}
```
danielpacak, line 61: Maybe it's a good time to rename Context to Event or Payload? IMO Context is quite confusing.

types/types.go
```diff
@@ -38,7 +38,16 @@ type SignatureEventSelector struct {
 type SignatureHandler func(found Finding)
 
 //Event is a generic event that the Engine can process
-type Event interface{}
+type EventHeader struct {
```
danielpacak, line 41: I'm okey with the current structure, but maybe we could rename EventHeader to EventHeaders to indicate that we can have multiple headers. I'd make it even more generic with helper methods to retrieve Content Type and Origin:

```go
type EventHeaders map[string]string

type Event struct {
	Headers  EventHeaders
	Payload interface{}
}

func (e Event) ContentType() string {
	return e.Headers["Content-Type"]
}

func (e Event) Origin() string {
	return e.Headers["Origin"]
}
```

pkg/external/external.go
```diff
@@ -5,10 +5,12 @@ import (
...
 	"fmt"
 	"strconv"
+
+	"github.com/aquasecurity/tracee/types"
 )
 
-// Event is a user facing data structure representing a single event
-type Event struct {
+// TraceeEvent is a user facing data structure representing a single event
+type TraceeEvent struct {
```
danielpacak, line 13: Just to make sure I'm following, TraceeEvent is a struc that represent a payload of of event generated by tracee-ebpf?

pkg/external/external.go
```diff
@@ -31,13 +33,46 @@ type Event struct {
...
+}
+
 // ToUnstructured returns a JSON compatible map with string, float, int, bool,
 // []interface{}, or map[string]interface{} children.
 //
 // It allows this Event to be manipulated generically. For example, it can be
 // used as a parsed input with OPA SDK to avoid relatively expensive JSON
 // encoding round trip.
-func (e Event) ToUnstructured() (map[string]interface{}, error) {
+func (e TraceeEvent) ToUnstructured() (map[string]interface{}, error) {
```
danielpacak, line 75: I think we can safely remove ToUnstructured as it's no longer used by the rules engine (see ToParsedEvent method in engine.go) Anyway this method doesn't make much sense anymore. We introduced it some time back hoping to improve performance of OPA evaluations.

danielpacak (commented):
Hey @NDStrahilevitz Overall LGTM! I left a few comments to discuss.

go.mod
NDStrahilevitz: Yes, as I wrote i'll introduce a preparation PR when this is out of draft.

types/types.go
NDStrahilevitz: Maybe, throughout the project we usually assign the event payload here so I guess `Payload` would be more appropriate.
I was also considering a more radical approach: we could report Findings through the new `Event` type, for example by setting the content type to `tracee/finding`.
This way we could have tracee responds to it's own findings organically (in terms of use cases - perhaps some signatures could depend on another signature triggering it, etc.) WDYT?

types/types.go
NDStrahilevitz: Perhaps we could have the headers fulfill such an interface?
```
type EventHeaders interface {
    ContentType() string
    Origin()             string
}
```

Like this we could indicate "mandatory" header results.

pkg/external/external.go
NDStrahilevitz: Yes. This is a name I agreed upon with @AsafEitani previously, however i'm open to suggestions for better names (I have my doubts about it after implementing this PR).

types/types.go
danielpacak: In my code snippet the Event stuct fulfils such interface. Structs and methods is probably all we need for now.

types/types.go
danielpacak: That's also a good idea. If we're not sure I'd leave it as is for now. We can maybe just document that TraceeEvents represents a payload of the event generated by tracee-ebpf to distinguish it from Event envelope, i.e. headers + payload

types/types.go
NDStrahilevitz: I think CloudEvents do the same thing with defining their headers as an interface (They do that so they can version their header structs). 
However, I am also slightly worried with using `map[string]string` for headers because we will need to allocate a map for each event pushed to tracee-rules, this can get expensive overtime (and performance is one of the reasons for this refactor as well from my PoV).

danielpacak: Agree. So maybe what you proposed is a good tradeoff:

```go
type EventHeaders struct {
	// ContentType indicates the content of the Event payload.
	ContentType string
	// Origin indicates whether Event originates from host or container.
	Origin string
	
	// Headers additional custom headers, nil most of the time
	Headers map[string]string
}

type Event struct {
	Headers EventHeaders
	
	Payload interface{}
}

func (e Event) ContentType() string {
	return e.Headers.ContentType
}

func (e Event) Origin() string {
	return e.Headers.Origin
}
```

pkg/external/external.go
NDStrahilevitz: Done

pkg/external/external.go
NDStrahilevitz: I changed it back to `trace.Event` with @AsafEitani's refactor I thought the package name kinda fulfills the role instead of using `TraceeEvent` WDYT

types/types.go
NDStrahilevitz: Done

types/types.go
NDStrahilevitz: I think i'll leave as is, maybe after this is merged we can open an issue about using these new capabilities more thoroughly.
For now - I'll document what you requested.

NDStrahilevitz:
I've uploaded a preparation PR for the types changes (#1493).

I suggest we move protocol discussions there and after merging it I will mark as ready for review.

pkg/external/external.go
NDStrahilevitz: Resolved in #1493

types/types.go
NDStrahilevitz: Resolved in #1493

pkg/rules/engine/engine.go
```diff
@@ -324,41 +307,23 @@ func (engine *Engine) UnloadSignature(signatureId string) error {
...
-	u, err := e.ToUnstructured()
-	if err != nil {
-		return ParsedEvent{}, fmt.Errorf("unstructuring event: %w", err)
-	}
-	// TODO In OPA >= v0.30.0 we can try passing tracee.Event directly to get rid of ToUnstructured call.
-	v, err := ast.InterfaceToValue(u)
-	if err != nil {
-		return ParsedEvent{}, fmt.Errorf("converting unstructured event to OPA ast.Value: %w", err)
+func eventSignatureSelector(event protocol.Event) (detect.SignatureEventSelector, error) {
+	origin := event.Origin()
```
danielpacak, line 311: Is is still in scope of event header refactor? Looks to me like we are changing the logic and way event selectors are used.

danielpacak (commented):
Overall LGTM. I'm wondering, are we also changing the logic of event selectors in trace rules in scope of this PR?

cmd/tracee-rules/input.go
```diff
@@ -61,7 +61,7 @@ func setupTraceeGobInputSource(opts *traceeInputOptions) (chan types.Event, erro
 					log.Printf("error while decoding event: %v", err)
 				}
 			} else {
-				res <- event
+				res <- event.ToProtocol()
```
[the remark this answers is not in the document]
NDStrahilevitz: In the mid to long term I also think conversions to protocol should be done at the producer level. 
However, in the case of tracee-ebpf that means changing the output scheme, which can currently output to json/table etc, at the `Payload` level (`trace.Event`). I think that with this perhaps a rethink to the output is required, and that's not in the scope of this PR IMO. I suggest we immediately open an issue about this after this PR.

pkg/procinfo/procinfo.go
```diff
@@ -75,7 +75,7 @@ func ParseProcessContext(ctx []byte, containers *containers.Containers) (Process
 	var procCtx = ProcessCtx{}
 	const ctxSize = 52
 	if len(ctx) < ctxSize {
-		return procCtx, fmt.Errorf("can't read process context: buffer too short")
+		return procCtx, fmt.Errorf("can't read process Payload: buffer too short")
```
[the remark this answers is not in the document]
NDStrahilevitz: That's by mistake i'll revert it.

pkg/rules/engine/engine.go
```diff
@@ -324,41 +307,23 @@ func (engine *Engine) UnloadSignature(signatureId string) error {
...
+	contentType := event.ContentType()
+	if !(strings.HasPrefix(origin, trace.EventSource) && strings.HasPrefix(event.ContentType(), trace.EventContentType)) {
```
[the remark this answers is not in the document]
NDStrahilevitz: By mistake.

pkg/rules/engine/engine.go
```diff
@@ -220,27 +217,13 @@ func (engine *Engine) consumeSources(done <-chan bool) {
...
-			}
-			engine.signatures[s] <- pe
-		} else {
-			engine.signatures[s] <- event
-		}
-	default:
-		engine.signatures[s] <- event
-	}
+func (engine *Engine) dispatchEvent(s detect.Signature, event protocol.Event) {
+	engine.signatures[s] <- event
```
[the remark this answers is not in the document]
NDStrahilevitz: @danielpacak said that the parsed event stuff is simply unnecessary for rego to work, so I just removed it, no special logic required instead.

pkg/rules/benchmark/signature/wasm/signatures.go
```diff
@@ -94,21 +96,26 @@ func NewSignature(metadata types.SignatureMetadata, selector []types.SignatureEv
...
 	return s.selector, nil
 }
 
-func (s *signature) OnEvent(event types.Event) error {
-	var input interface{} = event
+func (s *signature) OnEvent(event protocol.Event) error {
+	ee, ok := event.Payload.(trace.Event)
+
+	if !ok {
+		return fmt.Errorf("invalid event")
```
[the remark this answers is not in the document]
NDStrahilevitz: Agreed, that's just what it previously was.

NDStrahilevitz:
> In addition:
> 
> 1. in protocol.go we are exposing functions like `Origin`, `ContentType`, `Header` but the fields of `Event` and `EventHeaders` are also exported. What am I missing?
> 2. in particular, `EventHeaders.Custom` is exported and a user might use the library wrongly and addressing to a `nil` `Custom` field causing a program crash/panic.
> 3. let's add a documentation to package `trace` and `protocol` under folder `types`

1. The functions are there to ensure an abstracted access to those headers. IMO I'm perfectly fine with unexposing the headers and providing access only through receivers. However, that would complicate event building so maybe we can suffice using just the suggestion below (2)
2. Then as part of what I wrote above, at a minimum we could unexpose the`Custom` field and expose a `SetHeader` method
3. Yeah I see the packages themselves are missing GDocs, I can add them in this scope.

pkg/rules/engine/engine.go
mtcherni95: this conversation can be resolved IMO @danielpacak

cmd/tracee-rules/input.go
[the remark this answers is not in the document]
mtcherni95: So let's open an issue right after this PR. We want the Protocol event to be sent from client (aka tracee-ebpf). can be done after merging this one IMO.

pkg/rules/engine/engine.go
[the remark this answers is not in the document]
mtcherni95: @danielpacak could please just clarify why the event stuff is unnecessary for rego? I might be missing something here. thanks!

mtcherni95:
> > In addition:
> > 
> > 1. in protocol.go we are exposing functions like `Origin`, `ContentType`, `Header` but the fields of `Event` and `EventHeaders` are also exported. What am I missing?
> > 2. in particular, `EventHeaders.Custom` is exported and a user might use the library wrongly and addressing to a `nil` `Custom` field causing a program crash/panic.
> > 3. let's add a documentation to package `trace` and `protocol` under folder `types`
> 
> 1. The functions are there to ensure an abstracted access to those headers. IMO I'm perfectly fine with unexposing the headers and providing access only through receivers. However, that would complicate event building so maybe we can suffice using just the suggestion below (2)
> 2. Then as part of what I wrote above, at a minimum we could unexpose the`Custom` field and expose a `SetHeader` method
> 3. Yeah I see the packages themselves are missing GDocs, I can add them in this scope.

ok so let's unexpose `Custom` and provide its setter function.

NDStrahilevitz:
> > > In addition:
> > > 
> > > 1. in protocol.go we are exposing functions like `Origin`, `ContentType`, `Header` but the fields of `Event` and `EventHeaders` are also exported. What am I missing?
> > > 2. in particular, `EventHeaders.Custom` is exported and a user might use the library wrongly and addressing to a `nil` `Custom` field causing a program crash/panic.
> > > 3. let's add a documentation to package `trace` and `protocol` under folder `types`
> > 
> > 
> > 
> > 1. The functions are there to ensure an abstracted access to those headers. IMO I'm perfectly fine with unexposing the headers and providing access only through receivers. However, that would complicate event building so maybe we can suffice using just the suggestion below (2)
> > 2. Then as part of what I wrote above, at a minimum we could unexpose the`Custom` field and expose a `SetHeader` method
> > 3. Yeah I see the packages themselves are missing GDocs, I can add them in this scope.
> 
> ok so let's unexpose `Custom` and provide its setter function.

I need to upload a separate PR for that (since I need to refer to an existing commit here of `types`), i'll link it once ready.

mtcherni95:
> 
lets

mtcherni95:
> > > > In addition:
> > > > 
> > > > 1. in protocol.go we are exposing functions like `Origin`, `ContentType`, `Header` but the fields of `Event` and `EventHeaders` are also exported. What am I missing?
> > > > 2. in particular, `EventHeaders.Custom` is exported and a user might use the library wrongly and addressing to a `nil` `Custom` field causing a program crash/panic.
> > > > 3. let's add a documentation to package `trace` and `protocol` under folder `types`
> > > 
> > > 
> > > 
> > > 1. The functions are there to ensure an abstracted access to those headers. IMO I'm perfectly fine with unexposing the headers and providing access only through receivers. However, that would complicate event building so maybe we can suffice using just the suggestion below (2)
> > > 2. Then as part of what I wrote above, at a minimum we could unexpose the`Custom` field and expose a `SetHeader` method
> > > 3. Yeah I see the packages themselves are missing GDocs, I can add them in this scope.
> > 
> > 
> > ok so let's unexpose `Custom` and provide its setter function.
> 
> Ok, I can do that in a separate PR since it shouldn't affect this (ideally we would have caught this in the prep PR)

fair enough :)

pkg/rules/engine/engine.go
danielpacak: Okey. Thank you for the confirmation.

pkg/rules/engine/engine.go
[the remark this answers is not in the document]
danielpacak: There are a few reasons:
1. It doesn't solve fundamental performance issue with OPA Rego signatures.
2. We introduced it as a micro optimisation,  which also has broken abstractions of rules.Engine. For example, the rules.Engine is aware of OPA-specific options.
3. Event parsing is unnecessary for Go signatures, and overall might decrease performance if you load only Go signatures

types/protocol/protocol.go
```diff
@@ -1,3 +1,4 @@
+// Package protocol includes the "API" of events the rule-engine can consume. All producers who intend to add support for tracee need to support this protocol.
```
danielpacak, line 1: Usually you document packages in Go by adding the `doc.go` file? See https://go.dev/blog/godoc

danielpacak (approved):
LGTM. I only left one more comment regarding doc.go and documenting packages.

types/protocol/protocol.go
NDStrahilevitz: From what I read here, it seems `doc.go` is for large packages (for example the `tracee` package, they use `encoding/gob` as theirs). For small ones it's suggested to do short comments like the one above so I think this is fine.

NDStrahilevitz:
@danielpacak since you approved #1516 could we merge it before so I can point to it here?

types/protocol/protocol.go
danielpacak: Okey, I guess protocol won't become huge package and we can resolve this conversation for now.

pkg/rules/engine/engine.go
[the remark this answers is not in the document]
NDStrahilevitz: @mtcherni95 Can we resolve this?

mtcherni95:
> > > > In addition:
> > > > 
> > > > 1. in protocol.go we are exposing functions like `Origin`, `ContentType`, `Header` but the fields of `Event` and `EventHeaders` are also exported. What am I missing?
> > > > 2. in particular, `EventHeaders.Custom` is exported and a user might use the library wrongly and addressing to a `nil` `Custom` field causing a program crash/panic.
> > > > 3. let's add a documentation to package `trace` and `protocol` under folder `types`
> > > 
> > > 
> > > 
> > > 1. The functions are there to ensure an abstracted access to those headers. IMO I'm perfectly fine with unexposing the headers and providing access only through receivers. However, that would complicate event building so maybe we can suffice using just the suggestion below (2)
> > > 2. Then as part of what I wrote above, at a minimum we could unexpose the`Custom` field and expose a `SetHeader` method
> > > 3. Yeah I see the packages themselves are missing GDocs, I can add them in this scope.
> > 
> > 
> > ok so let's unexpose `Custom` and provide its setter function.
> 
> Ok, I can do that in a separate PR since it shouldn't affect this (ideally we would have caught this in the prep PR)

done here https://github.com/aquasecurity/tracee/pull/1516

mtcherni95 (approved):
LGTM!

Issue #1456, opened 2022-02-08
Flaky smoke tests


rafaeldtinoco:
### Discussed in https://github.com/aquasecurity/tracee/discussions/1442

<div type='discussions-op-text'>

<sup>Originally posted by **rafaeldtinoco** February  5, 2022</sup>
## Performance Observations (env: 5.15 kernel, 4 cpus, 4gb)

@danielpacak and I investigated some flaky smoke tests today. Smoke tests started being VERY flaky after we caused docker daemon to download and instantiate a container image, right before the test execution, instead of pulling the image first and then initializing the container.

What happened was that the workload of the docker daemon, in the docker hub node, was enough to make a very simple test (TRC-2) to fail significantly. I'm enumerating reproducers in multiple steps, here, isolating the variables:

1. multiple events + multiple rules
2. multiple events + single rule
3. single event + single rule

The test is the following:

- Run `tracee-ebpf`
- Run `while true; do ps -ef; done` in a 4 cpu box
- Run `strace ls` multiple times (wait a bit so pipeline gets full)

The tests:

----

1. multiple events + multiple rules

```
$ sudo ./dist/tracee-ebpf -o format:json -o option:parse-arguments --trace event=close,dup,dup2,dup3,execve,init_module,magic_write,mem_prot_alert,process_vm_writev,ptrace,sched_process_exec,sched_process_exit,security_bprm_check,security_file_open,security_kernel_read_file,security_sb_mount,security_socket_connect | ./dist/tracee-rules --input-tracee file:stdin --input-tracee format:json
Loaded 12 signature(s): [TRC-2 TRC-14 TRC-3 TRC-11 TRC-9 TRC-4 TRC-5 TRC-12 TRC-8 TRC-6 TRC-10 TRC-7]

*** Detection ***
Time: 2022-02-05T05:08:23Z
Signature ID: TRC-2
Signature: Anti-Debugging
Data: map[]
Command: strace
Hostname: debian

*** Detection ***
Time: 2022-02-05T05:08:43Z
Signature ID: TRC-2
Signature: Anti-Debugging
Data: map[]
Command: strace
Hostname: debian

lost 5891 events
lost 100578 events
```

Executed: 5 `strace` commands
Got: 2 detections

----

2. multiple events + single rule

```
[...]

PR #1484, opened 2022-02-17
tracee-ebpf: perf-buffer-size set based on host size

rafaeldtinoco:
An intermediate cache, in between the events production (tracee-ebpf)
and events consumption (tracee-rules), is needed, in order to allow
bursts of events not to cause event loss.

Fixes: #1456

This cache might be an internal (to tracee-ebpf) buffer, releasing the
pressure in the perf-buffer, OR it can be the perf-buffer itself (since
it is an anonymous pages mmap done by libbpf).

Having an internal asynchronous cache (and not perf-buffer only) is good
if we want to offload events to an external queue, for example. For now,
we don't need it, as we are not offloading the events. With that, we can
increase the perf-buffer size only and increase detection rates for
higher load bursts.

This commit sets default perf-buffer sizes based on the amount of memory
host has:

- host: up to 1  GB, perf-buffer-size = 256 MB
- host: up to 4  GB, perf-buffer-size = 512 MB
- host: up to 8  GB, perf-buffer-size = 1   GB
- host: up to 16 GB, perf-buffer size = 2   GB

For bigger hosts, the perf-buffer-size is 4 GB. The perf-buffer size
might be overriden by the cmd option:

--perf-buffer-size X

where X is the amount of memory pages reserved for the perf buffer (just
like it was before).


cmd/tracee-ebpf/main.go
```diff
@@ -279,7 +284,7 @@ func main() {
 			&cli.IntFlag{
 				Name:    "perf-buffer-size",
 				Aliases: []string{"b"},
-				Value:   1024,
+				Value:   0,
```
rafaeldtinoco, line 287: Oops I believe I can replace 0 by the function directly here, will do.

PR #1485, opened 2022-02-17
parseArgs: replace Sprintf with string concat

NDStrahilevitz:
# The fix

Simply change the way in which pointer arguments are parsed into a hex string in `parseArgs`. `fmt.Sprintf`, due to it's need to support many types of operations, can be less performant than using more specific operations. In this case, simply replacing `Sprintf` with a string concatenation conversion functions removes this code path bottleneck.

# Performance analysis

First a flame graph comparison:
 
With Sprintf:
image
Without:
image

Note the cpu percentage decrease in processEvents at the top of the screenshots (which should translate to gains in other places, hopefully in more time for event processing and signature handling).

In addition i've performed a benchmark between these two functions:

```
package tohex

import (
	"fmt"
	"strconv"
)

func SprintFToHex(val uintptr) string {
	return fmt.Sprintf("0x%X", val)
}

func ConcatToHex(val uintptr) string {
	return "0x" + strconv.FormatUint(uint64(val), 16)
}
```

And the results show a x2 efficiency increase in the second function:

```
il.ar.tal.gr@ILNadavSt tohex % go test -bench=.
goos: darwin
goarch: arm64
pkg: tohex
BenchmarkSprintf-8      19122786                62.52 ns/op
BenchmarkConcat-8       37810806                31.38 ns/op
PASS
ok      tohex   3.864s
il.ar.tal.gr@ILNadavSt tohex % go test -bench=.
goos: darwin
goarch: arm64
pkg: tohex
BenchmarkSprintf-8      17953947                62.43 ns/op
BenchmarkConcat-8       37373716                31.63 ns/op
PASS
ok      tohex   2.522s
```

With these two demonstrations I think it is clear that this PR will increase performance of this code path (which I assume is pretty frequent)

### DIFF
--- a/tracee-ebpf/tracee/argprinters.go
+++ b/tracee-ebpf/tracee/argprinters.go
@@ -2,6 +2,7 @@ package tracee
 
 import (
 	"fmt"
+	"strconv"
 
 	"github.com/aquasecurity/libbpfgo/helpers"
 	"github.com/aquasecurity/tracee/pkg/external"
@@ -10,7 +11,7 @@ import (
 func (t *Tracee) parseArgs(event *external.Event) error {
 	for i := range event.Args {
 		if ptr, isUintptr := event.Args[i].Value.(uintptr); isUintptr {
-			event.Args[i].Value = fmt.Sprintf("0x%X", ptr)
+			event.Args[i].Value = "0x" + strconv.FormatUint(uint64(ptr), 16)
 		}
 	}
 


tracee-ebpf/tracee/argprinters.go
```diff
@@ -10,7 +11,7 @@ import (
 func (t *Tracee) parseArgs(event *external.Event) error {
 	for i := range event.Args {
 		if ptr, isUintptr := event.Args[i].Value.(uintptr); isUintptr {
-			event.Args[i].Value = fmt.Sprintf("0x%X", ptr)
+			event.Args[i].Value = "0x" + strconv.FormatUint(uint64(ptr), 16)
```
mtcherni95, line 14: might be a silly question, but are we sure this won't case a regression? meaning is `strconv.FormatUint(uint64(ptr), 16)` outputting the same as `%X`?

yanivagman: Not silly at all, as there is a difference - digits won't be capitalized as before.
Anyway, I took this into consideration before I merged it. Argprinters are used for pretty printing, and eventually we will want to get rid of this extra parsting step and send "raw" events to tracee-rules. For now, even though we still have the argprinters set, no signature expects pointers, so that shouldn't cause any regressions

PR #1486, opened 2022-02-17
tracee-ebpf: minimize config map lookups in bpf code

yanivagman:
By grouping filter and option configuration, we can save many map
lookups per every traced event.
For example, if no filter is set by the user, there is no point to
perform extra map lookups just to get the configuration of each filter.
Instead of doing 14 map lookups in that case, only 2 are actually
needed.

This commit performs the required changes to reduce config map lookups
to the minimum required.


rafaeldtinoco (approved):
LGTM. Nice change btw!

PR #1488, opened 2022-02-17
events_pipeline: add queueEvents to cache pipeline events

rafaeldtinoco:
events_pipeline: add queueEvents to cache pipeline events

Under some circumstances, tracee-rules might be slower to consume events
than tracee-ebpf is capable of generating them. This requires tracee-
ebpf to deal with this possible lag, but, at the same, perf-buffer
consumption can't be left behind (or important events coming from the
kernel might be loss, causing detection misses).

There are 3 variables connected to this issue:

1) perf buffer could be increased to hold very big amount of memory
   pages: The problem with this approach is that the requested space,
   to perf-buffer, through libbpf, has to be contiguous and it is almost
   impossible to get very big contiguous allocations through mmap after
   a node is running for some time.

2) raising the events channel buffer to hold a very big amount of
   events: The problem is this approach is that the overhead of dealing
   with that amount of buffers, in a golang channel, causes event losses
   as well. It means this is not enough to relief the pressure from
   kernel events into perf-buffer.

3) create an internal, to tracee-ebpf, buffer based on the node size.

This commit introduces (3): a FIFO queue done in the pipeline.

OBS: IF the ratio of producer:consume is bigger than 1:1, and sustained
for more time than this cache can handle, then event loss will start to
happen. This is expected.

The second patch makes the caching engine an interface to be implemented.
The only backend implemented is mem caching:

The cmdline argument is "--cache cache-type=mem --cache mem-cache-size=X".
If mem-cache-size isn't provided, then automatic numbers for cache size
are calculated based on the amount of ram available.


rafaeldtinoco:
Alright, sorry for the forced pushes. It is clean for reviews now. 

@yanivagman I'm introducing the cache but:

1. It is not set by default in tracee-ebpf itself (only in entrypoint.sh, to help with flaky tests AND overall experience for users).

I believe the default cache size is conservative (based on our measurements) AND the default values can be changed if they are not adequate to real scenarios (My expectation is that values wouldn't go down if we have to change, only up).

2. User may be able to set the cache size with the cmdline. Imagining that the container image user finds that the default cache isn't good enough, for any reason, then user may provide cmdline argument to set cache bigger or smaller.

cmd/tracee-ebpf/main.go
```diff
@@ -91,6 +92,11 @@ func main() {
 			}
 			cfg.Output = &output
 
+			// enable option:cache-events if an events cache size was given
+			if !cfg.Output.EventsCache && cfg.EventsCacheSize > 0 {
+				cfg.Output.EventsCache = true
+			}
+
```
rafaeldtinoco, line 99: Should this move to the end of `func New(cfg Config)...` ?

NDStrahilevitz:
I have yet to fully test the performance impact this has myself, but could we maybe write the queueEvent mechanism as an interface? Currently we are limited to this implementation, but perhaps we would like to allow or test (I for sure would like to test other mechanisms as well) queueing through different mechanisms in the future (cache in filesystem, queue through kafka, etc).

Maybe something as simple as this as this could work:
```
type EventQueue interface {
    Queue(Event) error
    Dequeue()      error
}
```
We'd add that to tracee `Config` and we could simply use both as goroutines in the `queueEvents` parts according to the injected queue in the config. And obviously we would use your queue implementation by default.

tracee-ebpf/tracee/events_pipeline.go
```diff
@@ -38,6 +45,140 @@ func (t *Tracee) handleEvents(ctx gocontext.Context) {
...
+// 2) raising the events channel buffer to hold a very big amount of
+//    events: The problem with this approach is that the overhead of
+//    dealing with that amount of buffers, in a golang channel, causes
+//    event losses as well. It means this is not enough to relief the
+//    pressure from kernel events into perf-buffer.
+//
+// 3) create an internal, to tracee-ebpf, buffer based on the node size.
+
+// queueEvents implements an internal FIFO queue for caching events
+func (t *Tracee) queueEvents(ctx gocontext.Context, in <-chan *external.Event) (chan *external.Event, chan error) {
```
NDStrahilevitz, line 71: As mentioned in my comment, maybe we could use a Queue interface here instead of hardcoding this mechanism?

rafaeldtinoco:
> I have yet to fully test the performance impact this has myself, but could we maybe write the queueEvent mechanism as an interface? Currently we are limited to this implementation, but perhaps we would like to allow or test (I for sure would like to test other mechanisms as well) queueing through different mechanisms in the future (cache in filesystem, queue through kafka, etc).
> 
> Maybe something as simple as this as this could work:
> ```
> type EventQueue interface {
>     Queue(Event) error
>     Dequeue()      error
> }
> ```
> We'd add that to tracee `Config` and we could simply use both as goroutines in the `queueEvents` parts according to the injected queue in the config. And obviously we would use your queue implementation by default.

That's a good idea indeed. We can then implement the interface by other backends if/when appropriate. I'll work on that.

rafaeldtinoco:
@NDStrahilevitz could you please check if commit:ffe0f26b is close to what you had in mind ?

tracee-ebpf/tracee/events_pipeline.go
```diff
@@ -38,6 +44,79 @@ func (t *Tracee) handleEvents(ctx gocontext.Context) {
...
+	out := make(chan *external.Event, 10000)
+	errc := make(chan error, 1)
+	done := make(chan bool, 1)
+
+	// idea here is to have multiple queue backends. for now we currently have
+	// a single backend keeping events in memory based on config events cache
+	// size (in MBs for this backend)
+	evtQueue := queue.EventQueueMem{}
```
NDStrahilevitz, line 78: Could we pass this through tracee's `Config`? This way as a library consumer it would be easier to switch a backend, currently there would still be a need to manually change tracee's code.

So the config would have a field like this:
```
type Config struct {
    ...
    EventQueue EventQueue
}
```

NDStrahilevitz:
> @NDStrahilevitz could you please check if commit [8f5c41e](https://github.com/aquasecurity/tracee/commit/8f5c41e2fa4b35c99c11cad5eb8d986250acedb7) is close to what you had in mind ?

It's great, I've added just one more comment about it.

tracee-ebpf/tracee/events_pipeline.go
rafaeldtinoco: Done.

rafaeldtinoco:
@NDStrahilevitz mind reviewing this again ? I have implemented a generic interface for cache types. Everything can be configured through --cache cmdline argument OR through its internal CacheConfig struct. That will allow you to implement other queueing backends if you'd like.

rafaeldtinoco:
@danielpacak would you mind reviewing this for me as well ? It is likely a fast review (from the changes point of view). The "--cache cache-type=none" is the default option (so there are no changes if not enabled).

cmd/tracee-ebpf/internal/flags/flags-cache.go
```diff
@@ -0,0 +1,61 @@
...
+Use this flag multiple times to choose multiple output options
+`
+}
+
+func PrepareCache(cacheSlice []string) (queue.CacheConfig, error) {
+	cache := queue.CacheConfig{}
+	var err error
+
+	if strings.Contains(cacheSlice[0], "none") {
+		goto end
```
NDStrahilevitz, line 28: nit: Can we just return here instead of goto?

pkg/events/queue/queue.go
```diff
@@ -0,0 +1,33 @@
...
+type CacheType int
+
+const (
+	EventQueueTypeNone CacheType = iota
+	EventQueueTypeMem
+)
+
+type CacheConfig struct {
+	EventsCache        bool
+	EventsCacheType    CacheType
```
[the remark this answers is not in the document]
rafaeldtinoco: Okay so.. instead of having a config struct, we now configure the EventQueue implementation directly based on its type. AND there is Setup() function does not require any argument (check flags-cache.go). Also, I'm implementing a String() function to describe the cache type being loaded in debug.

Apart from that, I **think** that I implemented what you suggested, let me know if not, please. Thanks.

pkg/events/queue/queue_mem_list.go
```diff
@@ -0,0 +1,109 @@
...
+
+import (
+	"container/list"
+	"github.com/aquasecurity/tracee/pkg/external"
+	"sync"
+)
+
+type EventQueueMem struct {
+	amountOfEvents     int // number of cached events
+	mutex              sync.Mutex
```
mtcherni95, line 11: why not using `sync.RWMutex`?

pkg/events/queue/queue_mem_list.go
```diff
@@ -0,0 +1,109 @@
...
+	cache              *list.List
+	EventsCacheMemSize int
```
mtcherni95, line 13: Let's explicit the unit of measure here as well

pkg/events/queue/queue_mem_list.go
```diff
@@ -0,0 +1,109 @@
...
+func (q *EventQueueMem) Setup() error {
+	q.amountOfEvents = q.getQueueSizeInEvents()
+	q.cache = new(list.List)
+
+	return nil
+}
+
+// available returns when queue is available for enqueuing events
+func (q *EventQueueMem) available() {
+	for {
```
mtcherni95, line 29: I think what we are trying to achieve here can be done with condition variables. The variable state is changed by `Dequeue` function (once `q.cache.Len() < q.amountOfEvents`).
While Enqueue is waiting on the condition that `q.cache.Len() < q.amountOfEvents`

mtcherni95 (changes requested):

cmd/tracee-ebpf/internal/flags/flags-cache.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/aquasecurity/tracee/pkg/events/queue"
+	"strconv"
+	"strings"
+)
+
+func CacheHelp() string {
+	return `Select different cache types for the event pipeline queueing.
+Possible options:
+cache-type={none,mem}                              pick the appropriate cache type.
+mem-cache-size=256                                 set memory cache size. only works for cache-type=mem.
```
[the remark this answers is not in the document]
NDStrahilevitz: I think it's good to indicate unit of measure, but maybe we could document it in the help text instead. `mem-cache-size-bytes` is a bit of a mouthful IMO.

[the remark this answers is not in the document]
rafaeldtinoco: Yep, I agree to Nadav here. mem-cache-size-bytes is too long. Documenting might be a better approach.

cmd/tracee-ebpf/internal/flags/flags-cache.go
```diff
@@ -0,0 +1,67 @@
...
+			switch value {
+			case "mem":
+				cache = &queue.EventQueueMem{}
+				set = true
+			default:
+				return nil, fmt.Errorf("unrecognized cache-mem option: %s (valid options are: none,mem)", o)
+			}
+		case "mem-cache-size":
+			if v, ok := cache.(*queue.EventQueueMem); ok {
+				v.EventsCacheMemSize, err = strconv.Atoi(value)
```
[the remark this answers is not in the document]
rafaeldtinoco: Hum.. I thought about it.. I'm saying the type of cache being used.. but not its internal config options (because they might vary from backend to backend). Maybe I can try adding the debug into the cache backend itself, let me c..

cmd/tracee-ebpf/main.go
```diff
@@ -57,6 +57,19 @@ func main() {
 				Debug:              c.Bool("debug"),
 			}
 
+			if checkCommandIsHelp(c.StringSlice("cache")) {
```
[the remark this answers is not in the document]
rafaeldtinoco: Yep, but I have to do that for all other options as well. I'll add a commit to the PR changing that.

pkg/events/queue/queue_mem_list.go
rafaeldtinoco: Enqueue needs list write access, dequeue also needs list write access.. Len() from `available` is read-only, so I guess we could use RO mutex there.

cmd/tracee-ebpf/internal/flags/flags-cache.go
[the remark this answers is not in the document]
mtcherni95: I would prefer a longer name that provides all the needed information, instead of a shorter that requires me to read the documentation. Maybe we can make the name shorter and adding the unit of measure in it.

pkg/events/queue/queue_mem_list.go
rafaeldtinoco: I'm calling this function outside lock scope from Enqueue so I can lock and unlock while looping here (to allow dequeueing, while still checking the length).

In between one check and another there might be more than 1 dequeues, because of mutex fights, thus the loop. If I do 1 dequeue per 1 enqueue, I'll make enqueue/dequeueing synchronous whenever the buffer is full.. its a different approach then this one.

Now, if there is a better implementation inside Mutex class, like with lighter locking primitives, watching a variable state/change and waking up a callback, or something like a read-copy-update approach, then I can use here I suppose. 

One clear improvement for this would be to use RO locks for checking the length, as you said. Or we could yield sometime so the Dequeue would have more time to do chunks of dequeuing (this would reduce CPU consumption, but would impact throughput).

Quick note: when looping here, I'm still holding the event, instead of returning and loosing the event.

pkg/events/queue/queue_mem_list.go
rafaeldtinoco: fair enough.

pkg/events/queue/queue_mem_list.go
rafaeldtinoco: NM, I just read what you wrote correctly. I believe you are suggesting having a standalone variable that won't compete with the locking, I think that is good, will do it.

cmd/tracee-ebpf/internal/flags/flags-cache.go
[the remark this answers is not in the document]
rafaeldtinoco: Help command describes it is in MB. About having MB in the comm argument itself, I know no other unix tool that does that so I preferred to keep as is. I made it clear to be MB in the help AND changed variable name as requested (which, for the programmer, I think might be better).

pkg/events/queue/queue_mem_list.go
rafaeldtinoco: Explained at: https://github.com/aquasecurity/tracee/pull/1488#discussion_r813920308

cmd/tracee-ebpf/internal/flags/flags-cache.go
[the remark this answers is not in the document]
rafaeldtinoco: Yep, a good place to add this was in the queue String() function (which was already printed in --debug). Now we see this:

```
$ sudo taskset -c 0 ./dist/tracee-ebpf --debug -o option:parse-arguments -o option:detect-syscall --cache cache-type=mem --cache mem-cache-size=512...

Cache: cache type is "In-Memory Event Queue (Size = 512 MB)"
...
```

rafaeldtinoco:
@mtcherni95 I have done changes based on your comments. I have also squashed those changes into major commits ready to be merged. I have added 2 "fix" commits at the end of PR which are minor and addressing small fixes. Asking for your review again.

rafaeldtinoco:
Okay, I have tested this multiple times and it looks like the amount of memory being is very close to what is asked for AND the queue seems to be working fine (lose events when it reaches its maximum, able to detect events after catching up with workloads, etc). @mtcherni95 feel free to review at your convenience. Thanks!

pkg/events/queue/queue_mem_list.go
```diff
@@ -0,0 +1,114 @@
...
+			if !blocking {
+				return nil // return nil if non blocking
+			}
+			continue // continue until there is an event in the queue if blocking
+		}
+		break
+	}
+
+	e := q.cache.Front()
+	event := e.Value.(external.Event)
```
mtcherni95, line 66: I know we will be safe as we are enqueueing always objects of type `external.Event`, nevertheless I am always a fan of assertion checking: 
```
event, ok := e.Value.(external.Event)
if !ok {
...
}
```

pkg/events/queue/queue_mem_list.go
```diff
@@ -0,0 +1,114 @@
...
+
+// Dequeue pops an event from the queue
+func (q *EventQueueMem) Dequeue(blocking bool) *external.Event {
+
+	// "spin" until cache length is bigger than 0
+	for {
+		q.mutex.Lock() // can't be RO as Remove() will update the list
+		if q.cache.Len() == 0 {
+			q.mutex.Unlock() // release mutex to allow concurrency
+			if !blocking {
```
mtcherni95, line 57: if `Dequeue()` is called with `blocking` input being `true`, and length of the queue is 0 then we will have a go routine running in a while true style, causing high cpu usage and locking operation. 
Moreover, even when we are not blocking, the outer loop (the one calling `Dequeue()`) will also run in a while true loop as long as len of queue = 0. 
I am not sure spinning is a good approach here.
What am I missing?
This can be solved with condition variables, in this way we do not have to spin - which is expensive. If we want to be blocked we can sleep on the condition variable (len of queue != 0)

In general, we can `RLock` when reading the queue length and `Lock` when calling Remove()

pkg/ebpf/events_pipeline.go
```diff
@@ -38,6 +43,74 @@ func (t *Tracee) handleEvents(ctx gocontext.Context) {
...
+//    event losses as well. It means this is not enough to relief the
+//    pressure from kernel events into perf-buffer.
+//
+// 3) create an internal, to tracee-ebpf, buffer based on the node size.
+
+// queueEvents implements an internal FIFO queue for caching events
+func (t *Tracee) queueEvents(ctx gocontext.Context, in <-chan *external.Event) (chan *external.Event, chan error) {
+	out := make(chan *external.Event, 10000)
+	errc := make(chan error, 1)
+	done := make(chan bool, 1)
```
mtcherni95, line 72: I suggest using `done := make(chan struct{},1 )`
Go will optimize empty structs (struct {}), so it does not need any memory.

pkg/events/queue/queue_mem_list.go
```diff
@@ -0,0 +1,114 @@
...
+	return nil
+}
+
+// Enqueue pushes an event into the queue (may block until queue is available)
+func (q *EventQueueMem) Enqueue(evt *external.Event) {
+	if evt == nil {
+		return
+	}
+
+	// "spin" until cache length is smaller then maxAmountOfEvents (blocking)
```
mtcherni95, line 34: spinning will cause high cpu usage. I am not sure why we ain't using condition variable. wdyt?

mtcherni95:
looks good! I am still though a little bit concerned regarding the spinning approach causing high cpu usage. As said above in my comments, condition variables will fit best for this use-case. Looking forward for your thoughts @rafaeldtinoco

pkg/events/queue/queue_mem_list.go
rafaeldtinoco: We are not "spinning" (like spin locks) since mutexes allow the task to sleep/yield (so fair scheduler in Linux will be smart not to let this to step into other job if the preemption is needed).

Condition variables cannot be locked in a different "transaction", they would have to be locked while the "updates" are happening or else you always have a potential race in between the lock, the conditional variable and the data being changed.

My proposal to that concern is for us to sleep in this loop, and each time we loop we sleep a bit more until a threshold is reached. WDYT ?

This way if we have data to process we would be consuming CPU (obviously) but, if queue is empty for sometime, we would reduce CPU consumption.

pkg/ebpf/events_pipeline.go
rafaeldtinoco: Hum, nice, didn't know that.

pkg/events/queue/queue_mem_list.go
rafaeldtinoco: Like I said, splitting the condition variable from the transaction in charge for changing the actual data is racy (my previous comment). I suggested a change to reduce CPU consumption, let me know what you think.

rafaeldtinoco:
@mtcherni95 provided feedback on the condition variable idea, and another idea to reduce CPU consumption. if you agree I'll provide the changes.

pkg/events/queue/queue_mem_list.go
rafaeldtinoco: Okay as a feedback to this discussion, Michael and I just spoke on Teams and I was missing knowledge of the sync package functions he was proposing me to use. Now I'll change the loop to make use of wakeup like calls based on conditions (which was what I was missing).

pkg/events/queue/queue_mem_list.go
rafaeldtinoco: Per last comment in the other comment.

pkg/events/queue/queue.go
```diff
@@ -0,0 +1,14 @@
...
+	"github.com/aquasecurity/tracee/pkg/external"
+)
+
+type CacheConfig EventQueue
+
+type EventQueue interface {
+	String() string
+	Setup() error
+	Enqueue(*external.Event)
+	Dequeue(bool) *external.Event // bool is for blocking (true) or non-blocking (false)
```
NDStrahilevitz, line 13: This requires other Queue interfaces to implement Dequeue with a bool which might be unnecessary. I think this should move to be an implementation detail of `EventQueueMem`.

pkg/events/queue/queue_mem_list.go
```diff
@@ -0,0 +1,114 @@
+package queue
+
+import (
+	"container/list"
+	"fmt"
+	"github.com/aquasecurity/tracee/pkg/external"
+	"sync"
+)
+
+type EventQueueMem struct {
```
NDStrahilevitz, line 10: In continuation of above I think the `blocking` field should be defined here and initialized as pleased on queue construction.

rafaeldtinoco: Same as previous comment. Thanks.

rafaeldtinoco:
@mtcherni95

Thanks for reviewing it and demoting me from the "spinlock" like idea. It made me read more about golang sync package and make parallels to C/posix synchronization methods I knew.

A the end, I think that enqueue/dequeue functions will always consume maximum of CPU they can (because of the nature of pipeline and queueing) BUT it was the wrong approach (yielding CPUs until a condition is met is the right thing to do).

With that said, I have just pushed a rebased PR (since external.Event is now trace.Event) with the changes we've discussed. I also made sure to check if memory consumption was still the same as observed before AND if queue was reaching its limits and coming back to empty state (able to process new detections after that) and it seems to me it has worked fine.

I believe I have addressed all concerns so far, so I'm asking for a new review before we can merge it.

pkg/ebpf/events_pipeline.go
```diff
@@ -38,6 +43,76 @@ func (t *Tracee) handleEvents(ctx gocontext.Context) {
...
+	// receive and cache events (release pressure in the pipeline)
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				done <- true
+				return
+			case event := <-in:
+				if event != nil {
+					t.config.Cache.Enqueue(event) // may block if queue is full
```
rafaeldtinoco, line 89: we may block because of cond.Wait() inside the memory queue implementation BUT that is not a big deal since events will always happen and they will always unblock this (if done has to be processed).

pkg/ebpf/events_pipeline.go
```diff
@@ -38,6 +43,76 @@ func (t *Tracee) handleEvents(ctx gocontext.Context) {
...
+	go func() {
+		defer close(out)
+		defer close(errc)
+
+		for {
+			select {
+			case <-done:
+				return
+			default:
+				event := t.config.Cache.Dequeue() // may block if queue is empty
```
rafaeldtinoco, line 105: same here

pkg/events/queue/queue.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"github.com/aquasecurity/tracee/types/trace"
+)
+
+type CacheConfig EventQueue
+
+type EventQueue interface {
+	String() string
+	Setup() error
```
mtcherni95, line 11: what if the user doesn't call Setup? I usually tend to not have functions forcing the user to keep a state of their struct. I am suggesting to remove the `Setup` function from the interface and instead all the initialization/setup operation to be in a `NewEventQueueMem` function.

rafaeldtinoco: Yes, I thought about it as well and it is fine for me. I'll ask you to clone my repository and add fixes on top of it (I'm gone today) if you are ok with that.

pkg/events/queue/queue_mem_list.go
```diff
@@ -0,0 +1,110 @@
...
+	mutex                *sync.Mutex
+	cond                 *sync.Cond
+	cache                *list.List
+	maxAmountOfEvents    int // max number of cached events possible
+	EventsCacheMemSizeMB int
+}
+
+func (q *EventQueueMem) String() string {
+	return fmt.Sprintf("In-Memory Event Queue (Size = %d MB)", q.EventsCacheMemSizeMB)
```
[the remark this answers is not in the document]
rafaeldtinoco: Fine for me but we just call it once at the beginning of the debug iirc.

pkg/events/queue/queue.go
mtcherni95: sure, I already have a branch that takes your commits and on top of them has some more that should address my comments: https://github.com/mtcherni95/tracee/tree/michael-default-caching
I can close this PR and open a PR with the branch above, as we agreed. Let me know if confirmed then I'll proceed.

mtcherni95:
As agreed with @rafaeldtinoco I am gonna close this PR and open a new one based on his commits + some other commits that should address my comments. The reason behind this is that Rafael will be out for some days.

PR #1490, opened 2022-02-20
tracee-ebpf: add security_path_symlink event

OriGlassman:


AlonZivony (approved):
Seems good to me.
@yanivagman I think you can approve this quickly for @OriGlassman

tracee-ebpf/tracee/tracee.bpf.c
```diff
@@ -3035,6 +3036,29 @@ int BPF_KPROBE(trace_security_socket_create)
     return events_perf_submit(&data, SECURITY_SOCKET_CREATE, 0);
 }
 
+SEC("kprobe/security_path_symlink")
```
yanivagman, line 3039: security_path_* is guarded with CONFIG_SECURITY_PATH:
https://elixir.bootlin.com/linux/latest/source/include/linux/security.h#L1767
This config is not enabled on some distros, so we better not use it, and definitely not as a default event, which will fail tracee to start.

security_inode_symlink can be used instead

yanivagman (changes requested):

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -187,9 +187,10 @@ Copyright (C) Aqua Security inc.
 #define SECURITY_KERNEL_READ_FILE       1028
 #define SECURITY_INODE_MKNOD            1029
 #define SECURITY_POST_READ_FILE         1030
-#define SOCKET_DUP                      1031
-#define HIDDEN_INODES                   1032
-#define MAX_EVENT_ID                    1033
+#define SECURITY_INODE_SYMLINK           1031
```
yanivagman, line 190: please fix number indentation

pkg/ebpf/events_definitions.go
```diff
@@ -6082,6 +6083,18 @@ var EventsDefinitions = map[int32]EventDefinition{
 			{Type: "dev_t", Name: "dev"},
 		},
 	},
+	SecurityInodeSymlinkEventId: {
+		ID32Bit: sys32undefined,
+		Name:    "security_inode_symlink",
+		Probes: []probe{
+			{event: "security_inode_symlink", attach: kprobe, fn: "trace_security_inode_symlink"},
+		},
+		Sets: []string{"default", "lsm_hooks", "fs", "fs_file_ops"},
```
yanivagman, line 6092: If you don't mind, let's remove this from the default set for now

tracee-ebpf/tracee/tracee.bpf.c
yanivagman: > Fixed - There is one downside now though: 2 events are generated when a 'ln -s' is happening inside a container. One with the paths relevant for the container, and the other for the overlay fs. Unlike for when hooking security_path_symlink, which happens only in the container scope.

There might be an easy fix for this.
By looking at inode->i_sb->s_magic and comparing it to overlayfs magic number, you can know if the path is within the container or not

yanivagman (approved):

PR #1493, opened 2022-02-21
types: Add event headers

NDStrahilevitz:
Adds event headers to `detect`.
Adds new `protocol` package to convert between `trace.Event` and `detect.Event`

### DIFF
--- a/types/detect/detect.go
+++ b/types/detect/detect.go
@@ -1,6 +1,8 @@
 // Package detect includes the "API" of the rule-engine and includes public facing types that consumers of the rule engine should work with
 package detect
 
+import "github.com/aquasecurity/tracee/types/protocol"
+
 // Signature is the basic unit of business logic for the rule-engine
 type Signature interface {
 	//GetMetadata allows the signature to declare information about itself
@@ -12,7 +14,7 @@ type Signature interface {
 	//Close cleans the signature after Init operation
 	Close()
 	//OnEvent allows the signature to process events passed by the Engine. this is the business logic of the signature
-	OnEvent(event Event) error
+	OnEvent(event protocol.Event) error
 	//OnSignal allows the signature to handle lifecycle events of the signature
 	OnSignal(signal Signal) error
 }
@@ -37,9 +39,6 @@ type SignatureEventSelector struct {
 //SignatureHandler is a callback function that reports a finding
 type SignatureHandler func(found Finding)
 
-//Event is a generic event that the Engine can process
-type Event interface{}
-
 //Signal is a generic lifecycle event for a signature
 type Signal interface{}
 
@@ -49,6 +48,6 @@ type SignalSourceComplete string
 //Finding is the main output of a signature. It represents a match result for the signature business logic
 type Finding struct {
 	Data        map[string]interface{}
-	Context     Event
+	Event       protocol.Event //Event is the causal event of the Finding
 	SigMetadata SignatureMetadata
 }
--- a/types/protocol/protocol.go
+++ b/types/protocol/protocol.go
@@ -1,0 +1,32 @@
+package protocol
+
+//EventHeaders are headers attached to the Event struct, used to send metadata about the payload
+type EventHeaders struct {
+	// ContentType indicates the content of the Event payload.
+	ContentType string
+	// Origin indicates whether Event originates from host or container.
+	Origin string
+	// Custom additional custom headers, nil most of the time
+	Custom map[string]string
+}
+
+//Event is a generic event that the Engine can process
+type Event struct {
+	Headers EventHeaders
+	Payload interface{}
+}
+
+func (e Event) ContentType() string {
+	return e.Headers.ContentType
+}
+
+func (e Event) Origin() string {
+	return e.Headers.Origin
+}
+
+func (e Event) Header(header string) string {
+	if e.Headers.Custom != nil {
+		return e.Headers.Custom[header]
+	}
+	return ""
+}
--- a/types/trace/trace.go
+++ b/types/trace/trace.go
@@ -5,10 +5,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"strconv"
+
+	"github.com/aquasecurity/tracee/types/protocol"
 )
 
-// TraceeEvent is a user facing data structure representing a single event
-type TraceeEvent struct {
+// Event is a single result of an ebpf event process. It is used as a payload later delivered to tracee-rules.
+type Event struct {
 	Timestamp           int        `json:"timestamp"`
 	ProcessorID         int        `json:"processorId"`
 	ProcessID           int        `json:"processId"`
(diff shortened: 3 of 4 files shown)


types/detect/detect.go
```diff
@@ -49,6 +76,6 @@ type SignalSourceComplete string
 //Finding is the main output of a signature. It represents a match result for the signature business logic
 type Finding struct {
 	Data        map[string]interface{}
-	Context     Event
+	Context     interface{}
```
danielpacak, line 79: I think we should at least GDoc what we expect here. Event with envelope or just payload?

types/detect/detect.go
```diff
@@ -38,7 +38,34 @@ type SignatureEventSelector struct {
 type SignatureHandler func(found Finding)
 
 //Event is a generic event that the Engine can process
-type Event interface{}
+type EventHeaders struct {
```
danielpacak, line 41: We must fix GDoc, which refers to Event instead of EventHeaders

types/trace/trace.go
```diff
@@ -31,84 +31,19 @@ type TraceeEvent struct {
...
+	HostOrigin      EventOrigin = "host"
+)
 
-	var u interface{}
-	d := json.NewDecoder(buf)
-	d.UseNumber()
-	err = d.Decode(&u)
-	if err != nil {
-		return nil, err
+func (e Event) EventOrigin() EventOrigin {
```
danielpacak, line 41: This method has Event receiver so we can probably name it Origin()
```suggestion
func (e Event) Origin() EventOrigin {
```

types/protocol/protocol.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+const (
+	TraceeEventSrc         = "tracee"
+	TraceeEventContentType = "tracee/event"
+)
+
+func ToEvent(e trace.Event) detect.Event {
+	origin := TraceeEventSrc + "/" + string(e.EventOrigin())
+	contentType := TraceeEventContentType + "-" + e.EventName
```
danielpacak, line 15: Why do we add event name to the content type string? Is tracee-ebpf generating different kinds of events?

types/detect/detect.go
```diff
@@ -38,7 +38,34 @@ type SignatureEventSelector struct {
...
+	// ContentType indicates the content of the Event payload.
+	ContentType string
+	// Origin indicates whether Event originates from host or container.
+	Origin string
+	// Headers additional custom headers, nil most of the time
+	Headers map[string]string
```
danielpacak, line 47: Just an idea to call these headers `Custom` to distinguish from mandatory ContentType and Origin. WDYT?

```go
event.Headers.CotentType
event.Headers.Custom["Content-Length"]
```

danielpacak (changes requested):
Overall LGTM. I left few minor comments and questions

types/detect/detect.go
NDStrahilevitz: I like it, i'll add that change.

types/protocol/protocol.go
NDStrahilevitz: First of all, tracee-ebpf can also output other types of payloads but that's besides the point.
As far as the signatures see the events they are different, since the selectors filter by the event's `Name`, by including them in the content type I can avoid unpacking them and keep it a bit more generic. 
In the future, I think we could move from event selectors to simple header comparisons between sigs and events, however I think including the event name would likely stay in that case.
Maybe we could add a different header for it?

types/trace/trace.go
NDStrahilevitz: Agreed, will do.

types/detect/detect.go
NDStrahilevitz: I'm not sure either, we used to pass the payload here directly since we've sent the payload directly to the engine. I'm not sure what would be better...

NDStrahilevitz:
@AsafEitani I've renamed here `TraceeEvent` to `EbpfEvent` WDYT?

types/detect/detect.go
NDStrahilevitz: I've decided to keep it as the payload for now, i've renamed it as such and GDoc'd it.

types/protocol/protocol.go
danielpacak: Usually ContentType indicates how to decode the Payload, e.g. distinguish Tracee-ebpf event sent as GOB or JSON from Kubernetes audit event sent as JSON or YAML.

If we concatenate ContentType with event name we're probably enriching original event with pre-calculated metadata that's only useful for processing events down the pipeline, but I'd not call it ContentType.

NDStrahilevitz: Then sounds like a different header. I could use Customs to pass it, but again i'd like not to allocate maps so frequently, so I wonder if there is a hard coded header we could add here.

danielpacak: Regarding ContentType its usually defined as serialisation format + API version + Embedded format. See, for example, how Kafka defines it https://docs.confluent.io/platform/current/kafka-rest/api.html

danielpacak: But if tracee-rules is the only reader of the ContentType then we can maybe leave it with what you proposed in this PR. I don't have strong opinions until I see how we eventually use such ContentType header.

types/detect/detect.go
```diff
@@ -37,8 +37,36 @@ type SignatureEventSelector struct {
 //SignatureHandler is a callback function that reports a finding
 type SignatureHandler func(found Finding)
 
+//EventHeaders are headers attached to the Event struct, used to send metadata about the payload
+type EventHeaders struct {
+	// ContentType indicates the content of the Event payload.
+	ContentType string
```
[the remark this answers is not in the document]
danielpacak: Do we have benchmarks that show the difference? If yes and enums are indeed faster, we may consider enums. 

But in general, I'm in favour of Rob Pike's rules:
1. You can't tell where a program will spend its time
2. Don't optimise until you've measured
3. Fancy algorithms are slow for small N, and N is usually small
4. Fancy algorithms are buggier than simple ones
5. Data structures, not algorithms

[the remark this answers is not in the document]
NDStrahilevitz: We aren't using the `ContentType` to search inside `Custom` so there's no map lookups here. I think `ContentType` will mostly be used in simple string comparisons, not a very expensive operation. Also hard coding it as an enum will really limit flexibility in terms of new signatures and event sources in the future, which is one of the things we're aiming to keep and improve in this PR.

types/protocol/protocol.go
NDStrahilevitz: I think we should open an issue about header rigor and stuff like that after we merge this. I do think there's room for improvement. But yes, atm tracee-rules is the only consumer of this ContentType.

Specifically, I think these headers will allow us, for example, to streamline the `SignatureSelector` mechanism significantly if not entirely remove it. When we get to that we should debate the rigor of this header.

types/trace/trace.go
```diff
@@ -7,8 +7,8 @@ import (
 	"strconv"
 )
 
-// TraceeEvent is a user facing data structure representing a single event
-type TraceeEvent struct {
+// EbpfEvent is a single result of an ebpf event process. It is used as a payload later delivered to tracee-rules.
+type EbpfEvent struct {
```
danielpacak, line 11: Nit: Not sure what's the best name for the event payload generated by tracee-ebpf, but if we use acronyms in Go they're usually capital letters, i.e. `EBPFEvent`. At least that's what I've seen in Go SDK

danielpacak (approved):

types/trace/trace.go
NDStrahilevitz: I'm not so sure what name is best Fine by me, I'll change it to `EBPFEvent`.
So since you've approved, anything else before we merge this?

yanivagman: Why not just call it Event?

NDStrahilevitz: > Why not just call it Event?

I initially suggested this in the PR but @AsafEitani and @danielpacak had their reservations about it. I sort of understand, in a talk I had with @AsafEitani we kept confusing between the two, however I do think the `trace` package solves that issue (`external` was much less meaningful)

types/detect/detect.go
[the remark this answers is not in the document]
NDStrahilevitz: Resolved offline.

types/trace/trace.go
```diff
@@ -31,84 +31,19 @@ type TraceeEvent struct {
 	Args                []Argument `json:"args"` //Arguments are ordered according their appearance in the original event
 }
 
-type Stats struct {
-	EventCount  int
-	ErrorCount  int
-	LostEvCount int
-	LostWrCount int
-	LostNtCount int
-}
```
yanivagman, line 40: Why removing Stats from here?

NDStrahilevitz: Because it was left here by mistake, the actual stats are exported in the `metrics` packages.

types/trace/trace.go
NDStrahilevitz: I initially had reservations about it because `external.Event` and `types.Event` were confusing to discuss in conversation.
Now that it is split to `detect` and `trace` in `types` - I have no problems with calling them both `Event`. In fact, that was the initial state of the PR until @AsafEitani posted his reservations about it in #1455.
If @danielpacak approves calling it `trace.Event` I have no problem changing it back.

danielpacak: Let's call it trace.Event.

danielpacak: I'm just wondering if `detect.Event` that represent generic event with any payload, mandatory and custom headers belongs to the `detect` package. It's something shared between tracing and detection capabilities.

NDStrahilevitz: Maybe we could move `detect.Event` to `protocol.Event`?

NDStrahilevitz: I've updated the PR to reflect the above suggestion (including moving back to `trace.Event`)

types/protocol/protocol.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+	return ""
+}
+
+const (
+	TraceeEventSrc         = "tracee"
+	TraceeEventContentType = "tracee/event"
+)
+
+func ToEvent(e trace.Event) Event {
```
NDStrahilevitz, line 43: Not sure if `trace` should provide this conversion or `protocol`

@danielpacak WDYT?

NDStrahilevitz: Agreed, i'll move it to `trace`. Then, guess we can merge?

danielpacak: Yup

types/detect/detect.go
```diff
@@ -49,6 +48,6 @@ type SignalSourceComplete string
 //Finding is the main output of a signature. It represents a match result for the signature business logic
 type Finding struct {
 	Data        map[string]interface{}
-	Context     Event
+	Payload     interface{} //Payload is the event's payload that caused the Finding
```
danielpacak, line 51: Sorry for catching it late, but I think we may add protocol.Event here. Otherwise someone who's processing findings won't know the ContentType of the Payload.

types/protocol/protocol.go
danielpacak: Agreed. For now we have ContentType that serves our purpose to avoid Go type casing in so many places. We can further decide what would be the best value for ContentType.

PR #1498, opened 2022-02-22
introduce netproto package

itamarmaouda101:
Hi 

This PR will help #1363 by crating the `netproto package`, that will hold the network events parses.

For  example:
to add dns_request event, we will need to only add the parsing script to a new file under the package directory, and in the `net_events.go` do the event check of the event (like in the net_packet) and then send it to the dns_request handler

BTW: the only change in this PR is to move the `net_packet` event functions (and the `evtMeta struct`) to the net package under `net_packet.go`.

Note: 
After the we will separate the debug events from the network events (which is my next PR) ill create PR that add `network_event_processor`

### DIFF
--- a/pkg/ebpf/net_events.go
+++ b/pkg/ebpf/net_events.go
@@ -5,27 +5,19 @@ import (
 	gocontext "context"
 	"encoding/binary"
 	"fmt"
+	"github.com/aquasecurity/tracee/pkg/procinfo"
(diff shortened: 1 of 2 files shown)


rafaeldtinoco:
This should be on hold until https://github.com/aquasecurity/tracee/pull/1500 is merged (then it has to be rebased because of conflicts). FYIO.

Issue #1499, opened 2022-02-22
broken pcap


roikol:
currently captured pcap is broken: 
![image](https://user-images.githubusercontent.com/17714090/155147568-ea86a72a-db68-49cd-afc7-434257637468.png)

this is because the packets are saved from the wrong offset in `payloadBytes`, and the timestampts are wrong

yanivagman:
@itamarmaouda101 can it be related to your latest changes?

yanivagman:
Fixed in main

PR #1500, opened 2022-02-22
capture pcap with correct packet data

roikol:
solves #1499

### DIFF
--- a/pkg/ebpf/net_events.go
+++ b/pkg/ebpf/net_events.go
@@ -203,31 +203,9 @@ func (t *Tracee) processNetEvents(ctx gocontext.Context) {
 					continue
 				}
 
-				if !t.config.Debug {
-					packetBytes, err := getPacketBytes(netDecoder, netCaptureData.PacketLength)
-					if err != nil {
-						t.handleError(err)
-						continue
-					}
-
-					if err := t.writePacket(netCaptureData, time.Unix(0, int64(netEventMetadata.TimeStamp)), packetContext, packetBytes); err != nil {
-						t.handleError(err)
-						continue
-					}
-				} else {
+				if t.config.Debug {
 					evt, err := netPacketProtocolHandler(netDecoder, netEventMetadata, networkThread, "net_packet")
 					if err != nil {
-						t.handleError(err)
-						continue
-					}
-
-					// capture the packet
-					packetBytes, err := getPacketBytes(netDecoder, netCaptureData.PacketLength)
-					if err != nil {
-						t.handleError(err)
-						continue
-					}
-					if err := t.writePacket(netCaptureData, time.Unix(0, int64(netEventMetadata.TimeStamp)), packetContext, packetBytes); err != nil {
 						t.handleError(err)
 						continue
 					}
@@ -241,6 +219,16 @@ func (t *Tracee) processNetEvents(ctx gocontext.Context) {
 					}
 				}
 
+				// capture the packet
+				packetBytes, err := getPacketBytes(netDecoder, netCaptureData.PacketLength)
+				if err != nil {
+					t.handleError(err)
+					continue
+				}
+				if err := t.writePacket(netCaptureData, time.Unix(0, int64(netEventMetadata.TimeStamp)), packetContext, packetBytes); err != nil {
+					t.handleError(err)
+					continue
+				}
 			} else if t.config.Debug {
 				var netDebugEvent bufferdecoder.NetDebugEvent
 				err = netDecoder.DecodeNetDebugEvent(&netDebugEvent)
@@ -334,10 +322,7 @@ func (t *Tracee) writePacket(capData bufferdecoder.NetCaptureData, timeStamp tim
 func getPacketBytes(netDecoder *bufferdecoder.EbpfDecoder, packetLength uint32) ([]byte, error) {
 	packetBytes := make([]byte, packetLength)
 	err := netDecoder.DecodeBytes(packetBytes[:], packetLength)
-	if err != nil {
-		return packetBytes, err
-	}
-	return packetBytes, nil
+	return packetBytes, err
 }
 
 // netPacketProtocolHandler parse a given a packet bytes buffer to packetMeta and event
--- a/pkg/ebpf/net_events.go
+++ b/pkg/ebpf/net_events.go
@@ -28,9 +28,8 @@ type EventMeta struct {
 }
 
 type CaptureData struct {
-	PacketLength      uint32 `json:"pkt_len"`
-	InterfaceIndex    uint32 `json:"if_index"`
-	PacketStartOffset uint32 `json:"pkt_start_off"`
+	PacketLength   uint32 `json:"pkt_len"`
+	InterfaceIndex uint32 `json:"if_index"`
 }
 
 type netInfo struct {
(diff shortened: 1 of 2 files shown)


pkg/ebpf/net_events.go
```diff
@@ -186,14 +186,14 @@ func netPacketProtocolHandler(buffer []byte, evtMeta EventMeta, ctx procinfo.Pro
 	return evt, nil
 }
 
-func parseCaptureData(payload []byte) (CaptureData, error) {
+func parseCaptureData(payload []byte) (CaptureData, int, error) {
```
grantseltzer: Can you just add a comment here that the int represents the packet start offset?

pkg/ebpf/net_events.go
```diff
@@ -186,14 +189,18 @@ func netPacketProtocolHandler(buffer []byte, evtMeta EventMeta, ctx procinfo.Pro
...
-		return capData, 0, fmt.Errorf("payload too short\n")
+		return capData, fmt.Errorf("payload too short\n")
 	}
-	capData.PacketLen = binary.LittleEndian.Uint32(payload[0:4])
+	capData.PacketLength = binary.LittleEndian.Uint32(payload[0:4])
 	capData.InterfaceIndex = binary.LittleEndian.Uint32(payload[4:8])
-	return capData, 8, nil
+	capData.PacketStartOffset = 8
+	if debug {
+		capData.PacketStartOffset = 48
```
grantseltzer, line 201: Out of curiosity, what is it about debug mode that pushes the packet start offset ahead? What is it exactly that's 'in debug mode'?

grantseltzer (approved):
I tested this and confirms it fixes the issue. I tested it for capturing tcp traffic on the loopback interface, but if there's something else I should test on my machine let me know.

Just left a question comment that i'm very curious about and suggestion for adding a comment.

pkg/ebpf/net_events.go
```diff
@@ -138,19 +139,21 @@ func (t *Tracee) processNetEvents(ctx gocontext.Context) {
...
 		Timestamp:      timeStamp,
-		CaptureLength:  int(capData.PacketLen),
-		Length:         int(capData.PacketLen),
+		CaptureLength:  int(capData.PacketLength),
+		Length:         int(capData.PacketLength),
 		InterfaceIndex: int(idx),
 	}
 
-	err := t.pcapWriter.WritePacket(info, dataBuff.Bytes()[:capData.PacketLen])
+	fmt.Printf("====================== packet_len: %d, start_offset: %d\n", capData.PacketLength, capData.PacketStartOffset)
```
yanivagman, line 154: Please remove this debug message

pkg/ebpf/net_events.go
yanivagman: This shouldn't happen here, see my previous comments.
In addition, we shuoldn't use magic numbers, as these might change in the future if and when we will change the data passed from bpf code.

Answer to @grantseltzer question: The packet payload comes after the metadata.
If debug is not enabled, only the packet length and interface index are passed.
Otherwise, there is other metadata (source ip, port, etc) that is passed

pkg/ebpf/net_events.go
```diff
@@ -50,12 +51,12 @@ func (t *Tracee) processNetEvents(ctx gocontext.Context) {
 			}
 
 			if evtMeta.NetEventId == NetPacket {
-				captureData, err := parseCaptureData(payloadBytes)
+				captureData, err := parseCaptureData(payloadBytes, t.config.Debug)
```
yanivagman, line 54: We shouldn't pass t.config.Debug to this helper, as it has nothing to do with the capture data bytes which are always the same

pkg/ebpf/net_events.go
```diff
@@ -21,8 +21,9 @@ type EventMeta struct {
 }
 
 type CaptureData struct {
-	PacketLen      uint32 `json:"pkt_len"`
-	InterfaceIndex uint32 `json:"if_index"`
+	PacketLength      uint32 `json:"pkt_len"`
+	InterfaceIndex    uint32 `json:"if_index"`
+	PacketStartOffset uint32 `json:"pkt_start_off"`
```
yanivagman, line 26: Your'e mixing the data received from bpf code with variables used to parse it.
Please don't do that.

yanivagman (changes requested):

roikol:
> I tested this and confirms it fixes the issue. I tested it for capturing tcp traffic on the loopback interface, but if there's something else I should test on my machine let me know.
> 
> Just left a question comment that i'm very curious about and suggestion for adding a comment.

thank you @grantseltzer for your review and testing of the code!

pkg/ebpf/net_events.go
roikol: just to elaborate: 

in debug mode we send extra data for debug.

for each network event, the first bytes are:
"Timestamp:NetworkEventId:HostTid:ProcessName".

if the NetworkEventId indicates a packet, we add:
"PacketLen:InterfaceIndex"

and for debug mode we add:
"SrcIp:DstIP:SrcPort:DstPort:Protocol:fullPacket"

so packet with no debug:
"Timestamp:NetworkEventId:HostTid:ProcessName:PacketLen:InterfaceIndex:fullPacket"

packet with debug mode:
"Timestamp:NetworkEventId:HostTid:ProcessName:PacketLen:InterfaceIndex:SrcIp:DstIP:SrcPort:DstPort:Protocol:fullPacket"

roikol:
thanks @yanivagman for your review.
i changed the approach, and now i'm returning the offsets from the parsing functions. let me know what you think

yanivagman (commented):
I have to say that I don't like the approach of keeping track of the current offset in the buffer.
We better move the structs used here into the bufferdecoder package, which also defines the protocol between tracee-ebpf and tracee-rules.

roikol:
> I have to say that I don't like the approach of keeping track of the current offset in the buffer. We better move the structs used here into the bufferdecoder package, which also defines the protocol between tracee-ebpf and tracee-rules.

actually i agree, seems like `bufferdecoder` is perfect for this use case.
should i create a new file under the `bufferdecoder` package? something like net_decoder.go?

itamarmaouda101:
> > I have to say that I don't like the approach of keeping track of the current offset in the buffer. We better move the structs used here into the bufferdecoder package, which also defines the protocol between tracee-ebpf and tracee-rules.
> 
> actually i agree, seems like `bufferdecoder` is perfect for this use case. should i create a new file under the `bufferdecoder` package? something like net_decoder.go?

Hi (:
I think because we are starting now to add more network events, maybe we should create a new package for the network events.
TBH I created a PR last week that do it: #1498 
now it only has the `net_packet `event, but later ill add `DNS protocol` events (it's already ready for opening PR)

pkg/ebpf/net_events.go
```diff
@@ -324,79 +331,50 @@ func (t *Tracee) writePacket(capData CaptureData, timeStamp time.Time, packetCon
...
+	packetBytes := make([]byte, packetLength)
+	err := netDecoder.DecodeBytes(packetBytes[:], packetLength)
 	if err != nil {
-		return evt, err
+		return packetBytes, err
 	}
-	evt = CreateNetEvent(evtMeta, ctx, eventName, int(evtMeta.TimeStamp))
-	appendPktMetaArg(&evt, packet)
-	return evt, nil
+	return packetBytes, nil
```
yanivagman, line 340: just `return packetBytes, err` will have the same effect

pkg/ebpf/net_events.go
```diff
@@ -185,95 +172,115 @@ func (t *Tracee) processNetEvents(ctx gocontext.Context) {
...
+						t.handleError(err)
+						continue
+					}
+
+					// output the event
+					select {
+					case t.config.ChanEvents <- evt:
+						t.stats.NetEvCount.Increment()
+					case <-ctx.Done():
+						return
```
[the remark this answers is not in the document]
itamarmaouda101: Yes (:
#1515 will add the ability to select `net_packet` as event and also remove it from the debug events

yanivagman (approved):

Issue #1357, opened 2022-01-18
magic_write: event doen't catch all file writes operations


AlonZivony:
## Original Issue
The syscall `sendfile` which is another method for writing to files does not trigger the `magic_write` event.

## Expanded Issue
After examining the file write operations in the kernel, it seems that there are 2 ways to write into a file:

1. Using one of the following file operations (of the generic `struct file_operations`) - `write`, `write_iter`, `splice_write`, `copy_file_range`
2. Writing to a mapped memory to a file (which created using the `mmap` file operation).

Our goal is to catch all cases in the event.
**This issue focus on the first one - unsupported file operations**

## Catching File Operations
To succeed catching file operations, we need to hook global-use functions which are the closest to the use of the operations.
Unfortunately, even though each file operation has a envelope function (like `do_splice_from` for the `splice_from` file operation), these envelope functions are inline most of the times so they are not hookable.

One of the biggest problems of doing `magic_write` is that we need to extract the data written. However, all file operations except `write` work with files, not with buffer. To optimize their operations, they implement their logic by coping pages most of the times - not chunks of data. This means that the only interface we have with the data is through pages struct, so we need the functionality of translating page struct to its virtual address to be able to read the information.
The implementation of translating `struct page` to its virtual address is a very complicated one, which depends on the kernel version, architecture and physical memory model. This means that we have to load the **kconfig** values to determine which physical memory is used. The code also need to be very flexible according to different kernel versions. Also, all algorithms use global variables in the kernel, so receiving their address is also mandatory.

## Feature Implementation Roadmap
1. Load environment values:
    * Kernel symbols #1505  
    * Kconfigs #1504  
    * Calculate consts/defines of the kernel (for CORE)
2. Implement `struct page` to virtual address #1507 
3. Putting a hook for each file operation:
    * `splice_from`
        - `do_splice` #1551 
        - `direct_splice_actor` #1552 
    * `write_iter`
        - `__kernel_write` #1550 
        - `do_iter_readv_writev`
    * `copy_file_range`
        - `vfs_copy_file_range`

PR #1508, opened 2022-02-23
Load kallsyms values

### LINKED ISSUE #1505: Missing global symbols in BPF programs
Some global symbols are useful in BPF programs to enable access to kernel data and for understanding kernel build.
For example some functions existence mark that the kernel was built with some kconfig enabled, and there are structures in the kernel to save offsets of data and addresses.
This is a CO-RE issue because in non CO-RE the globals are available to use.

AlonZivony:
Load symbols from /proc/kallsyms to BPF map.
The first commit is a patch until the loading of kallsyms will be merged into libbpfgo, so please only review the second commit.
This is a first step needed for #1357.

Fix #1505

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -79,6 +79,7 @@ Copyright (C) Aqua Security inc.
 #define MAX_STR_FILTER_SIZE             16        // bounded to size of the compared values (comm)
 #define FILE_MAGIC_HDR_SIZE             32        // magic_write: bytes to save from a file's header
 #define FILE_MAGIC_MASK                 31        // magic_write: mask used for verifier boundaries
+#define MAX_KSYM_NAME_SIZE              64
 
 #define SUBMIT_BUF_IDX                  0
 #define STRING_BUF_IDX                  1
@@ -466,6 +467,10 @@ typedef struct string_filter {
     char str[MAX_STR_FILTER_SIZE];
 } string_filter_t;
 
+typedef struct ksym_name {
+    char str[MAX_KSYM_NAME_SIZE];
+} ksym_name_t;
+
 typedef struct event_data {
     struct task_struct *task;
     context_t context;
@@ -593,6 +598,7 @@ BPF_HASH(params_types_map, u32, u64);                   // encoded parameters ty
 BPF_HASH(process_tree_map, u32, u32);                   // filter events by the ancestry of the traced process
 BPF_HASH(process_context_map, u32, process_context_t);  // holds the process_context data for every tid
 BPF_HASH(network_config, u32, int);                     // holds the network config for each iface
+BPF_HASH(ksymbols_map, ksym_name_t, u64)            // holds the addresses of some kernel symbols
 BPF_LRU_HASH(sock_ctx_map, u64, net_ctx_ext_t);         // socket address to process context
 BPF_LRU_HASH(network_map, local_net_id_t, net_ctx_t);   // network identifier to process context
 BPF_ARRAY(config_map, u32, 4);                          // various configurations
@@ -1175,6 +1181,18 @@ static __always_inline int get_kconfig_val(u32 key)
         return 0;
 
     return *config;
+}
+
+static __always_inline void* get_symbol_addr(char *symbol_name)
+{
+    char new_ksym_name[MAX_KSYM_NAME_SIZE] = {};
+    bpf_probe_read_str(new_ksym_name, MAX_KSYM_NAME_SIZE, symbol_name);
+    void **sym = bpf_map_lookup_elem(&ksymbols_map, (void *)&new_ksym_name);
+
+    if (sym == NULL)
+        return 0;
+
+    return *sym;
 }
 
 static __always_inline int init_context(context_t *context, struct task_struct *task, u32 options)
--- a/pkg/ebpf/events_definitions.go
+++ b/pkg/ebpf/events_definitions.go
@@ -22,8 +22,12 @@ type probe struct {
 	fn     string
 }
 
-type dependency struct {
+type eventDependency struct {
 	eventID int32
+}
+
+type dependencies struct {
+	events   []eventDependency // Events required to be loaded and/or submitted for the event to happen
 }
 
 // EventDefinition is a struct describing an event configuration
@@ -31,7 +35,7 @@ type EventDefinition struct {
 	ID32Bit        int32
 	Name           string
 	Probes         []probe
-	Dependencies   []dependency
+	Dependencies   dependencies
 	EssentialEvent bool
 	Sets           []string
 	Params         []trace.ArgMeta
@@ -5860,9 +5864,8 @@ var EventsDefinitions = map[int32]EventDefinition{
 		ID32Bit: sys32undefined,
 		Name:    "magic_write",
 		Probes:  []probe{}, Sets: []string{},
-		Dependencies: []dependency{
-			{eventID: VfsWriteEventID},
-			{eventID: VfsWritevEventID},
+		Dependencies: dependencies{
+			events: []eventDependency{{VfsWriteEventID}, {VfsWritevEventID}},
 		},
 		Params: []trace.ArgMeta{
 			{Type: "const char*", Name: "pathname"},
@@ -6127,10 +6130,8 @@ var EventsDefinitions = map[int32]EventDefinition{
 		ID32Bit: sys32undefined,
 		Name:    "socket_dup",
 		Probes:  []probe{},
-		Dependencies: []dependency{
-			{eventID: DupEventID},
-			{eventID: Dup2EventID},
-			{eventID: Dup3EventID},
+		Dependencies: dependencies{
+			events: []eventDependency{{DupEventID}, {Dup2EventID}, {Dup3EventID}},
 		},
 		Sets: []string{},
 		Params: []trace.ArgMeta{
@@ -6154,8 +6155,8 @@ var EventsDefinitions = map[int32]EventDefinition{
 		ID32Bit: sys32undefined,
 		Name:    "container_create",
 		Probes:  []probe{},
-		Dependencies: []dependency{
-			{eventID: CgroupMkdirEventID},
+		Dependencies: dependencies{
+			events: []eventDependency{{CgroupMkdirEventID}},
 		},
 		Sets: []string{},
 		Params: []trace.ArgMeta{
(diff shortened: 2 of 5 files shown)


rafaeldtinoco:
@AlonZivony Im converting this to draft so I can review Itamar's PR first, then Ill get back to this one (after you rebase and demote from being draft). I hope you dont mind. Just trying to organize the reviews.

AlonZivony:
Updated to use new libbpfgo @rafaeldtinoco

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -579,6 +579,7 @@ BPF_HASH(sys_32_to_64_map, u32, u32);                   // map 32bit to 64bit sy
 BPF_HASH(params_types_map, u32, u64);                   // encoded parameters types for event
 BPF_HASH(process_tree_map, u32, u32);                   // filter events by the ancestry of the traced process
 BPF_HASH(process_context_map, u32, process_context_t);  // holds the process_context data for every tid
+BPF_HASH(symbols_map, u32, u64)                         // holds the addresses of kernel symbols
```
yanivagman, line 582: let's call it ksymbols_map or kernel_symbols_map to signify the origin of the symbols (we might have userspace symbols in the future as well).
Also, maybe change the comment to:
"holds the addresses of **some** kernel symbols"

pkg/ebpf/tracee.go
```diff
@@ -267,6 +269,12 @@ func New(cfg Config) (*Tracee, error) {
 	}
 	t.containers = c
 
+	t.kernelSymbols, err = helpers.NewKernelSymbolsMap()
+	if err != nil {
+		t.Close()
+		return nil, fmt.Errorf("error creating symbols map: %v", err)
+	}
```
yanivagman, line 276: I think we should only initialize this map if there's an event that actually uses it.
This way, even if there is no access to kallsysms for some reason, we will not totally fail tracee.

pkg/ebpf/tracee.go
```diff
@@ -555,6 +563,24 @@ func (t *Tracee) populateBPFMaps() error {
...
+	kallsymsValues := initialization.LoadKallsymsValues(t.kernelSymbols)
+
+	for key, value := range kallsymsValues {
+		keyU32 := uint32(key)
+		address := value.Address
+		err = bpfKallsymsMap.Update(unsafe.Pointer(&keyU32), unsafe.Pointer(&address))
+		if err != nil {
+			return err
+		}
+	}
```
yanivagman, line 582: ditto

yanivagman (changes requested):
I think this commit should be part of a larger PR that actually uses this symbols map.
Otherwise, we just add unused code into tracee.

pkg/ebpf/tracee.go
AlonZivony: I can argue with it, because using the symbols is not only relevant for the eBPF programs.
For example, if the kconfig file is not available in the distro we want to use it to determine the kconfigs used in compilation. This is done in the usermode, though you can argue that in that case we can request the data sepcifically.

If you want me to implement the check if some program use it, it will also add the overhead of changing the events definitions interface to add dependencies (we can add the use of the kconfig map as well). I am not rejecting the idea, just asking if this is what you want me to implement.

AlonZivony: More argument to do it is to enable more flexibility of eBPF programs loading. If in the future we would like to load an BPF program on the fly, we would like to have all information needed beforehand (or to load it on the fly). It doesn't mean that we have to load it beforehand, but it means that if you want us not to load it for removing some capabilities for example it will make it difficult to support dynamic loading, so I don't think it will help from the security aspect too much.

cmd/tracee-ebpf/main.go
```diff
@@ -176,6 +176,8 @@ func main() {
 			cfg.ChanEvents = make(chan trace.Event)
 			cfg.ChanErrors = make(chan error)
 
+			cfg.RequiredInits = &tracee.RequiredInitConfig{}
```
yanivagman, line 179: I don't understand why do we need that for.
If an event requires a kconfig, isn't it enough to declare it in its dependencies?

pkg/ebpf/events_definitions.go
```diff
@@ -22,16 +22,20 @@ type probe struct {
 	fn     string
 }
 
-type dependency struct {
-	eventID int32
+// Requirement is a type for information required by the event to operate, which might require some inializations
```
yanivagman, line 25: typo: inializations

pkg/ebpf/events_definitions.go
```diff
@@ -99,6 +103,10 @@ const (
 
 const Unique32BitSyscallsStartID = 3000
 
+const (
+	Ksymbols Requirement = iota
```
yanivagman, line 107: When defining an enum, the convention is to put the type definition (`type Requirement uint8`) just beofre the consts

pkg/ebpf/tracee.go
```diff
@@ -179,12 +185,40 @@ type Tracee struct {
...
 	procInfo          *procinfo.ProcInfo
 	eventsSorter      *sorting.EventsChronologicalSorter
+	kernelSymbols     *helpers.KernelSymbolTable
 }
 
 func (t *Tracee) Stats() *metrics.Stats {
 	return &t.stats
 }
 
+func isEssentials(id int32) bool {
```
yanivagman, line 195: nit: isEssential (not plural)

pkg/ebpf/tracee.go
```diff
@@ -271,6 +302,14 @@ func New(cfg Config) (*Tracee, error) {
 	}
 	t.containers = c
 
+	if t.config.RequiredInits.kallsyms {
+		t.kernelSymbols, err = helpers.NewKernelSymbolsMap()
```
yanivagman, line 306: I would take this init before handling the event dependencies

pkg/ebpf/tracee.go
```diff
@@ -271,6 +302,14 @@ func New(cfg Config) (*Tracee, error) {
...
+		if err != nil {
+			t.Close()
```
yanivagman, line 308: Instead of failing tracee in case of an error initilializing the symbols map, let's just print an error for all the events that have such dependencies, and continue as usual with any other event

pkg/ebpf/tracee.go
```diff
@@ -179,12 +185,40 @@ type Tracee struct {
...
+	event.EssentialEvent = true
+	EventsDefinitions[id] = event
+}
+
+func (t *Tracee) handleEventsDependencies(e int32) {
+	eDependencies := EventsDefinitions[e].Dependencies
+	for _, req := range eDependencies.requirements {
+		switch req {
+		case Ksymbols:
+			t.config.RequiredInits.kallsyms = true
```
yanivagman, line 210: I think that instead of setting this config (which is actually intended for use by whoever uses the ebpf package, and not for internal use inside the package itself), you should aggregate these symbols to initialize the bpf map later.

yanivagman (changes requested):

cmd/tracee-ebpf/main.go
AlonZivony: We need it to determine for the whole program if it should load the kallsyms file or not.
I created a category in the configuration for things that require initialization if are required by the code, so when the question if to initialize them is given we have a single bool flag to determine it.

pkg/ebpf/tracee.go
AlonZivony: I think that the current situation is not as you describe.
If the events channel, error channel and the kconfigs are part of the configuration, it means that the use of the config is no longer to represent the configuration of the user, but to represent the state of tracee or to be its way of communication outside.
Anyway, I can do as you say. It is not so important to use config here (just making it more organized - a part is responsible for handling events info and another to load information like kallsyms). I do think that if this is how you think of the configuration maybe we should divide it to environment and configuration.

pkg/ebpf/tracee.go
AlonZivony: As we discussed, this will be answered in another PR

pkg/ebpf/tracee.go
```diff
@@ -185,12 +190,37 @@ type Tracee struct {
...
+	event := EventsDefinitions[id]
+	event.EssentialEvent = true
+	EventsDefinitions[id] = event
+}
+
+func handleEventsDependencies(e int32, initReq *RequiredInitValues) {
+	eDependencies := EventsDefinitions[e].Dependencies
+	if len(eDependencies.ksymbols) > 0 {
+		initReq.kallsyms = true
+	}
```
yanivagman, line 214: Instead of initReq, you can populate a list with all the symbols required by the different events.
You can later use this list to initialize the bpf map

pkg/ebpf/tracee.go
```diff
@@ -580,6 +619,25 @@ func (t *Tracee) populateBPFMaps() error {
...
+		if err != nil {
+			return err
+		}
+
+		var reqKsyms []string
+		for id, event := range EventsDefinitions {
+			if event.EssentialEvent || t.eventsToTrace[id] {
+				reqKsyms = append(reqKsyms, event.Dependencies.ksymbols...)
+			}
+		}
```
yanivagman, line 634: This looks like something you could have achieved in the EventsDefinitions iteration you did in New.
See my previous comment

pkg/ebpf/tracee.go
```diff
@@ -79,6 +79,11 @@ type netProbe struct {
 	egressHook  *bpf.TcHook
 }
 
+// RequiredInitValues determines if to initialize values that might be needed by eBPF programs
+type RequiredInitValues struct {
+	kallsyms bool
+}
+
```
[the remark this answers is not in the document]
AlonZivony: In my opinion - it is important.
It is important because it is expandable. If there will be more requirements for initialization, you have a very convenient way to store them without changing the code. Its use will be the greatest if this value is also determined from iterations on the events, which will mean you won't need to change any function interface to add it, just add some code to the `handleDependencies` function.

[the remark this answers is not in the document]
AlonZivony: It is possible to transform this value to a list of requested symbols like you requested in other comment.

pkg/ebpf/events_definitions.go
```diff
@@ -22,16 +22,17 @@ type probe struct {
 	fn     string
 }
 
-type dependency struct {
-	eventID int32
+type dependencies struct {
+	eventIDs []int32
```
AlonZivony, line 26: Change to slice of struct named `eventsDependenceis` that include the event's ID

yanivagman (approved):
LGTM!

Issue #1507, opened 2022-02-23
Translate struct page to virtual address in BPF programs


AlonZivony:
In my work on the issue of missing `magic_write` event during `sendfile` syscall (#1357) I concluded that given the fact that we can't read from files directly the data, the best thing we have in the flow of the code in the kernel is access to pages that include the data (through the `pipe_inode_info` struct).
So, to be able to read the data from a page, a method to convert `struct page` to its matching  virtual address is required.
It is a bit complicated to implement because of different physical memory models available and different architecture implementations, but it should be possible using defines, kconfig and globals (#1506  #1504 #1505).

yanivagman:
This is more an implementation discussion than an issue in tracee.
Converting to discussion

PR #1512, opened 2022-02-23
Load expened kconfig to BPF

### LINKED ISSUE #1504: Missing kconfig values in BPF programs
Kconfig change the behavior of the kernel in different cases, which is critical for some implementations of BPF programs to know.
Currently there is a single value we pass to the BPF programs, which is not expandable.

AlonZivony:
Changed kconfig loading to be expandable, and added kconfigs which are required in the future to calculate page address.
Fix #1504

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -288,7 +288,17 @@ Copyright (C) Aqua Security inc.
 #define get_kconfig(x) CONFIG_##x
 #endif
 
+#ifdef CORE
+
 #define ARCH_HAS_SYSCALL_WRAPPER        1000U
+
+#else
+
+#ifndef CONFIG_ARCH_HAS_SYSCALL_WRAPPER
+#define CONFIG_ARCH_HAS_SYSCALL_WRAPPER 0
+#endif
+
+#endif // CORE
 
 /*================================ eBPF MAPS =================================*/
 
--- a/pkg/ebpf/initialization/kconfig.go
+++ b/pkg/ebpf/initialization/kconfig.go
@@ -1,0 +1,44 @@
+package initialization
+
+import (
+	"fmt"
+	"github.com/aquasecurity/libbpfgo/helpers"
+	"os"
+)
+
+// Custom KernelConfigOption's to extend kernel_config helper support
+// Add here all kconfig variables used within tracee.bpf.c
+const (
+	CONFIG_ARCH_HAS_SYSCALL_WRAPPER helpers.KernelConfigOption = iota + helpers.CUSTOM_OPTION_START
+)
+
+var kconfigUsed = map[helpers.KernelConfigOption]string{
+	CONFIG_ARCH_HAS_SYSCALL_WRAPPER: "CONFIG_ARCH_HAS_SYSCALL_WRAPPER",
+}
+
+// LoadKconfigValues load all kconfig variables used within tracee.bpf.c
+func LoadKconfigValues(kc *helpers.KernelConfig, isDebug bool) (map[helpers.KernelConfigOption]helpers.KernelConfigOptionValue, error) {
+	values := make(map[helpers.KernelConfigOption]helpers.KernelConfigOptionValue)
+	var err error
+	for key, keyString := range kconfigUsed {
+		if err = kc.AddCustomKernelConfig(key, keyString); err != nil {
+			return nil, err
+		}
+	}
+
+	// re-load kconfig and get just added kconfig option values
+	if err = kc.LoadKernelConfig(); err != nil { // invalid kconfig file: assume values then
+		if isDebug {
+			fmt.Fprintf(os.Stderr, "KConfig: warning: assuming kconfig values, might have unexpected behavior\n")
+		}
+		for key, _ := range kconfigUsed {
+			values[key] = helpers.UNDEFINED
+		}
+		values[CONFIG_ARCH_HAS_SYSCALL_WRAPPER] = helpers.BUILTIN // assume CONFIG_ARCH_HAS_SYSCALL_WRAPPER is a BUILTIN option
+	} else {
+		for key, _ := range kconfigUsed {
+			values[key] = kc.GetValue(key) // undefined, builtin OR module
+		}
+	}
+	return values, err
+}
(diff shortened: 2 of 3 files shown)


pkg/ebpf/initialization/kconfig.go
```diff
@@ -0,0 +1,60 @@
...
+
+	// re-load kconfig and get just added kconfig option values
+	if err = kc.LoadKernelConfig(); err != nil { // invalid kconfig file: assume values then
+		if isDebug {
+			fmt.Fprintf(os.Stderr, "KConfig: warning: assuming kconfig values, might have unexpected behavior\n")
+		}
+		for key, _ := range kconfigUsed {
+			values[key] = helpers.UNDEFINED
+		}
+		values[CONFIG_ARCH_HAS_SYSCALL_WRAPPER] = helpers.BUILTIN // assume CONFIG_ARCH_HAS_SYSCALL_WRAPPER is a BUILTIN option
```
rafaeldtinoco, line 53: Would you like to assume other values if KCONFIG is not present ? This is the case for alpine containers IIRC (not /proc/kconfig nor /boot/config-$(uname -r) files are available, and we're assuming the most common used values, like CONFIG_ARCH_HAS_SYSCALL_WRAPPER).

rafaeldtinoco (commented):
This change looks good. I have 1 question (made bellow) and also, I would like to confirm the "pkg/ebpf/initialization/kconfig.go" with @yanivagman. I'm not opposing to it, just want to know if we will start doing that for other parts as well.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -288,7 +288,49 @@ Copyright (C) Aqua Security inc.
...
+#define	SPARSEMEM_EXTREME               1005U
+#define	DYNAMIC_MEMORY_LAYOUT           1006U
+#define	X86_32                          1007U
+#define	X86_PAE                         1008U
+
+#else
+
+#ifndef CONFIG_ARCH_HAS_SYSCALL_WRAPPER
+#define CONFIG_ARCH_HAS_SYSCALL_WRAPPER 0
+#endif
```
grantseltzer, line 307: How will these be used? How are defining them here different than loading it into the `kconfig_map` BPF_HASH map?

pkg/ebpf/initialization/kconfig.go
```diff
@@ -0,0 +1,60 @@
...
+const (
+	CONFIG_ARCH_HAS_SYSCALL_WRAPPER helpers.KernelConfigOption = iota + helpers.CUSTOM_OPTION_START
+	CONFIG_FLATMEM
+	CONFIG_DISCONTMEM
+	CONFIG_SPARSEMEM
+	CONFIG_SPARSEMEM_VMEMMAP
+	CONFIG_SPARSEMEM_EXTREME
+	CONFIG_DYNAMIC_MEMORY_LAYOUT
+	CONFIG_X86_32
+	CONFIG_X86_PAE
```
grantseltzer, line 20: I understand that `helpers.CUSTOM_OPTION_START` allows for expansion of config options, but all of these can be added to the libbpfgo helpers package as well right? When this merges could you create an issue so that we can put them in a more permanent spot?

grantseltzer (commented):
Just some questions on this, i'm struggling to picture the second half of this solution.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -288,7 +288,49 @@ Copyright (C) Aqua Security inc.
...
+
 #define ARCH_HAS_SYSCALL_WRAPPER        1000U
+#define FLATMEM                         1001U
+#define	DISCONTMEM                      1002U
+#define	SPARSEMEM                       1003U
+#define	SPARSEMEM_VMEMMAP               1004U
+#define	SPARSEMEM_EXTREME               1005U
+#define	DYNAMIC_MEMORY_LAYOUT           1006U
+#define	X86_32                          1007U
+#define	X86_PAE                         1008U
```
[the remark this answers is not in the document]
AlonZivony: So these values are the keys for the map initialized in the user-mode in the `initialization/kconfig.go` file.

They are defined only for CORE because in no-CORE we have all defines in the header files we import (the `#include <linux/kconfig.h>` we do). So, to use them we just concat the name to `CONFIG_` and we have the values.
This mechanism was established before this PR.

pkg/ebpf/c/tracee.bpf.c
AlonZivony: When using no-CORE, we have the issue that the kconfig values might not be defined.
To bypass it, I defined them in the case they are not defined with the value of 0 to be able to create a uniform behavior using no-CORE and CORE with the `get_kconfig` macro.
If you have better solution to this issue I will gladly hear :)

pkg/ebpf/initialization/kconfig.go
AlonZivony: I don't think they should be added to libbpfgo, because they are specific to the future uses I have for them in translating pages to addresses. I don't think any bpf program should have them.
However, you can say that adding them in this PR is out of scope.
In that case I can add the configs in the big PR of my feature solving #1507, but I felt that putting the different parts in different PRs might be easier to review.

AlonZivony:
> Responses made sense, you have a couple of failing tests that seem related to reading in config, any thoughts on what's causing it?

The tests that failed are the smoke tests, which has a problem right now.
All tests passed.

AlonZivony:
@yanivagman can you merge the PR for me?

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: Thanks @grantseltzer for checking this and @AlonZivony to keep using what we came up with. The reasoning behind this is the following:

If you ever use a kconfig relocation (done by libbpf when you declare an external symbol coming from the kconfig file) by using the following syntax: extern KCONFIG_OPTION __kconfig, then libbpf will, FOR SURE, require a kconfig file existence (/boot/kconfig-$(uname -r) OR /proc/kconfig.gz). 

@yanivagman and I decided that, in order not to be forced to have those 2 files, and being able to assume kconfig values, we would pass the values through kconfig map. With that, we can "mock" kconfig values to a container, for example, without requiring the container to have a kconfig file (which libbpf would require).

There are pros and cons with that approach, for sure, and we may revisit this in the near future. For now, it was the right thing to do (Alon using the same approach), IMO.

Thanks!

pkg/ebpf/initialization/kconfig.go
rafaeldtinoco: I agree with @AlonZivony here, that is exactly why the helpers had this option of being extended (because libbpfgo might not need the kconfigs tracee will need). Thanks!

Issue #1514, opened 2022-02-24
enable tracing network events


itamarmaouda101:
Hi

As part of the network change we currently do, we need to able tracing network events from the terminal.
For example in the `net_packet` event, we should enable the option to trace it with giving to the user the ability to select that event and also add a interface name to listen on.

currently I'm working on 2 main things that will enable that:
  1. separate the network events from the debug events event numbering (maybe in the fetrue we will take them out from the net_event.go file)
  2. adding a `net` option to the `-trace` flag to give the user the ability to ask for specific interface to listen to.

rafaeldtinoco:
This has already been done.

PR #1515, opened 2022-02-24
enable tracing net_packet event

itamarmaouda101:
Hi 
this PR solves #1514 
by separate the `net_packet` to new numbering of events which will be only for network events and adding the `net` option flag to the `-trace`  flag

### DIFF
--- a/cmd/tracee-ebpf/internal/flags/flags-capture.go
+++ b/cmd/tracee-ebpf/internal/flags/flags-capture.go
@@ -72,7 +72,7 @@ func PrepareCapture(captureSlice []string) (tracee.CaptureConfig, error) {
 		} else if cap == "mem" {
 			capture.Mem = true
 		} else if strings.HasPrefix(cap, "net=") {
-			err := tracee.ParseIface(strings.TrimPrefix(cap, "net"), &capture.NetIfaces)
+			err := tracee.ParseIface(strings.TrimPrefix(cap, "net="), &capture.NetIfaces)
 			if err != nil {
 				return tracee.CaptureConfig{}, err
 			}
--- a/cmd/tracee-ebpf/internal/flags/flags-capture.go
+++ b/cmd/tracee-ebpf/internal/flags/flags-capture.go
@@ -2,7 +2,6 @@ package flags
 
 import (
 	"fmt"
-	"net"
 	"os"
 	"path/filepath"
 	"strings"
@@ -73,20 +72,9 @@ func PrepareCapture(captureSlice []string) (tracee.CaptureConfig, error) {
 		} else if cap == "mem" {
 			capture.Mem = true
 		} else if strings.HasPrefix(cap, "net=") {
-			iface := strings.TrimPrefix(cap, "net=")
-			if _, err := net.InterfaceByName(iface); err != nil {
-				return tracee.CaptureConfig{}, fmt.Errorf("invalid network interface: %s", iface)
-			}
-			found := false
-			// Check if we already have this interface
-			for _, item := range capture.NetIfaces {
-				if iface == item {
-					found = true
-					break
-				}
-			}
-			if !found {
-				capture.NetIfaces = append(capture.NetIfaces, iface)
+			err := tracee.ParseIface(strings.TrimPrefix(cap, "net"), &capture.NetIfaces)
+			if err != nil {
+				return tracee.CaptureConfig{}, err
 			}
 		} else if strings.HasPrefix(cap, "pcap:") {
 			netCaptureContext := strings.TrimPrefix(cap, "pcap:")
--- a/cmd/tracee-ebpf/internal/flags/flags-filter.go
+++ b/cmd/tracee-ebpf/internal/flags/flags-filter.go
@@ -2,6 +2,7 @@ package flags
 
 import (
 	"fmt"
+	"net"
 	"strings"
 
 	tracee "github.com/aquasecurity/tracee/pkg/ebpf"
@@ -73,6 +74,28 @@ To 'escape' those operators, please use single quotes, e.g.: 'uid>0'
 `
 }
 
+func parseNetFields(comm string, interfaceArray *[]string) (bool, error) {
+	if strings.HasPrefix(comm, "net=") {
+		iface := strings.TrimPrefix(comm, "net=")
+		if _, err := net.InterfaceByName(iface); err != nil {
+			return false, fmt.Errorf("invalid network interface: %s", iface)
+		}
+		found := false
+		// Check if we already have this interface
+		for _, item := range *interfaceArray {
+			if iface == item {
+				found = true
+				break
+			}
+		}
+		if !found {
+			*interfaceArray = append(*interfaceArray, iface)
+		}
+		return true, nil
+	}
+	return false, nil
+}
+
 func PrepareFilter(filters []string) (tracee.Filter, error) {
 	filter := tracee.Filter{
 		UIDFilter: &tracee.UintFilter{
@@ -125,7 +148,8 @@ func PrepareFilter(filters []string) (tracee.Filter, error) {
 		ProcessTreeFilter: &tracee.ProcessTreeFilter{
 			PIDs: make(map[uint32]bool),
 		},
-		EventsToTrace: []int32{},
+		EventsToTrace:    []int32{},
+		InterfaceToTrace: []string{},
 	}
 
 	eventFilter := &tracee.StringFilter{Equal: []string{}, NotEqual: []string{}}
@@ -278,7 +302,13 @@ func PrepareFilter(filters []string) (tracee.Filter, error) {
 			filter.Follow = true
 			continue
 		}
-
+		found, err := parseNetFields(f, &filter.InterfaceToTrace)
+		if err != nil {
+			return filter, err
+		}
+		if found {
+			continue
+		}
 		return tracee.Filter{}, fmt.Errorf("invalid filter option specified, use '--trace help' for more info")
 	}
 
--- a/cmd/tracee-ebpf/internal/flags/flags-filter.go
+++ b/cmd/tracee-ebpf/internal/flags/flags-filter.go
@@ -130,7 +130,7 @@ func PrepareFilter(filters []string) (tracee.Filter, error) {
 			PIDs: make(map[uint32]bool),
 		},
 		EventsToTrace: []int32{},
-		NetFilter: &tracee.NetFilter{
+		NetFilter: &tracee.IfaceFilter{
 			InterfacesToTrace: []string{},
 		},
 	}
@@ -216,7 +216,7 @@ func PrepareFilter(filters []string) (tracee.Filter, error) {
 		}
 
 		if strings.HasPrefix(filterName, "net") {
-			err := filter.NetFilter.Parse(operatorAndValues)
+			err := filter.NetFilter.Parse(strings.TrimPrefix(operatorAndValues, "="))
 			if err != nil {
 				return tracee.Filter{}, err
 			}
--- a/cmd/tracee-ebpf/internal/flags/flags-filter.go
+++ b/cmd/tracee-ebpf/internal/flags/flags-filter.go
@@ -2,7 +2,6 @@ package flags
 
 import (
 	"fmt"
-	"net"
 	"strings"
 
 	tracee "github.com/aquasecurity/tracee/pkg/ebpf"
@@ -38,6 +37,9 @@ The field 'set' selects a set of events to trace according to predefined sets, w
 
 The special 'follow' expression declares that not only processes that match the criteria will be traced, but also their descendants.
 
+The field 'net' specifies which interfaces to monitor when tracing network events.
+Notice that the 'net' field is mandatory when tracing network events.
+
 Examples:
   --trace pid=new                                              | only trace events from new processes
   --trace pid=510,1709                                         | only trace events from pid 510 or pid 1709
@@ -67,33 +69,12 @@ Examples:
   --trace openat.pathname=/tmp*                                | only trace 'openat' events that have 'pathname' prefixed by "/tmp"
   --trace openat.pathname!=/tmp/1,/bin/ls                      | don't trace 'openat' events that have 'pathname' equals /tmp/1 or /bin/ls
   --trace comm=bash --trace follow                             | trace all events that originated from bash or from one of the processes spawned by bash
+  --trace net=docker0 			                       | trace the net events over docker0 interface
 
 
 Note: some of the above operators have special meanings in different shells.
 To 'escape' those operators, please use single quotes, e.g.: 'uid>0'
 `
-}
-
-func parseNetFields(comm string, interfaceArray *[]string) (bool, error) {
-	if strings.HasPrefix(comm, "net=") {
-		iface := strings.TrimPrefix(comm, "net=")
-		if _, err := net.InterfaceByName(iface); err != nil {
-			return false, fmt.Errorf("invalid network interface: %s", iface)
-		}
-		found := false
-		// Check if we already have this interface
-		for _, item := range *interfaceArray {
-			if iface == item {
-				found = true
-				break
-			}
-		}
-		if !found {
-			*interfaceArray = append(*interfaceArray, iface)
-		}
-		return true, nil
-	}
-	return false, nil
 }
 
 func PrepareFilter(filters []string) (tracee.Filter, error) {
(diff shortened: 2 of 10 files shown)


pkg/ebpf/net_events.go
```diff
@@ -203,13 +203,12 @@ func (t *Tracee) processNetEvents(ctx gocontext.Context) {
 					continue
 				}
 
-				if t.config.Debug {
+				if t.eventsToTrace[NetPacket] {
```
itamarmaouda101, line 206: add verification of the packet interface with the asked interface in `-trace` flag

pkg/ebpf/tracee.go
```diff
@@ -266,6 +281,16 @@ func New(cfg Config) (*Tracee, error) {
 	}
 	t.containers = c
 
+	t.networkInfo.ifacesIndexList = make(map[int]*net.Interface)
+	for _, iface := range appendStr2Str(t.config.Filter.InterfaceToTrace, t.config.Capture.NetIfaces) {
```
roikol, line 285: maybe `appendStr2Str` is redundent now. you can just loop over these two lists and check whether the `netIface.Index` exists in the `t.networkInfo.ifacesIndexList` map (or even just update the map without checking). 
the question is what costs more - calling `net.InterfaceByName(iface)` (maximum of 2 calls per interface), or calling `appendStr2Str`.

pkg/ebpf/tracee.go
```diff
@@ -825,20 +841,14 @@ func (t *Tracee) initBPF() error {
...
-
-	err = t.populateBPFMaps()
-	if err != nil {
-		return err
-	}
-
-	if t.config.Capture.NetIfaces != nil || t.config.Debug {
-		for _, iface := range t.config.Capture.NetIfaces {
-			ingressHook, err := t.attachTcProg(iface, bpf.BPFTcIngress, "tc_ingress")
+	if t.config.Capture.NetIfaces != nil || t.config.Filter.InterfaceToTrace != nil {
```
roikol, line 844: this if statememt is no longer needed - you can just loop over the `t.networkInfo.ifacesIndexList` list.

pkg/ebpf/tracee.go
```diff
@@ -1070,3 +1081,19 @@ func (t *Tracee) invokeInitNamespacesEvent() {
 		t.stats.EventCount.Increment()
 	}
 }
+func (f *Filter) getTracedIfaceIdx(ifaceName string) int {
```
yanivagman, line 1084: Unlike `getCapturedIfaceIdx` which has a meaning for the index (which is the index of the NgInterface), I don't think there is a meaning for the traced interface index. Instead, you can just have a function that checks if `ifaceName` is valid  (exists in InterfaceToTrace slirce) or not, and return a bool

yanivagman (changes requested):
Few comments follow.
As this code changes multiple interfaces logic, please test that when capturing more than one interface, the correct interface name appears for each in the captured pcap file.

pkg/bufferdecoder/net_decoder.go
```diff
@@ -8,7 +8,7 @@ import (
 
 type NetEventMetadata struct {
 	TimeStamp   uint64   `json:"timeStamp"`
-	NetEventId  uint32   `json:"netEventId"`
+	NetEventId  int32    `json:"netEventId"`
```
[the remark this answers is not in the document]
itamarmaouda101: Because now the type of the net events will be int32 (as the regular events type)
So if we just parse it at the first time as int32 , we will save later the cast in the `netEventsProcessor`

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4361,7 +4363,7 @@ static __always_inline int tc_probe(struct __sk_buff *skb, bool ingress) {
     // See bpf_skb_event_output in net/core/filter.c.
     u64 flags = BPF_F_CURRENT_CPU;
     flags |= (u64)skb->len << 32;
```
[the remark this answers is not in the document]
itamarmaouda101: It depends, in the case of `net_packets` it's fine to send only the packet metadata (Src/Dst IP, port, and protocol)
but in the case of the other net events like DNS/IRC/HTTP etc.
we will need to send the whole packet because we will want to parse it in the userspace.

I can send only the  metadata in case of `net_packet` event and capture flag wasn't chosen, WYT?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4360,19 +4363,22 @@ static __always_inline int tc_probe(struct __sk_buff *skb, bool ingress) {
...
-        bpf_perf_event_output(skb, &net_events, flags, &pkt, sizeof(pkt));
-    }
-    else {
-        // If not debugging, only send the minimal required data to save the
-        // packet. This will be the timestamp (u64), net event_id (u32),
-        // host_tid (u32), comm (16 bytes), packet len (u32), and ifindex (u32)
-        bpf_perf_event_output(skb, &net_events, flags, &pkt, 40);
     }
 
+    if ((get_config(CONFIG_OPTIONS) & OPT_CAPTURE_PACKETS)){
```
yanivagman, line 4378: Doing this is not enough I'm afraid.
In the case of multiple interfaces receieved as an input, how do you know if this specific interface was chosen for capture or not?

cmd/tracee-ebpf/internal/flags/flags-capture.go
```diff
@@ -73,21 +72,12 @@ func PrepareCapture(captureSlice []string) (tracee.CaptureConfig, error) {
...
-				}
-			}
-			if !found {
-				capture.NetIfaces = append(capture.NetIfaces, iface)
+
+			err := tracee.ParseIface(strings.TrimPrefix(cap, "net"), &capture.NetIfaces)
+			if err != nil {
+				return tracee.CaptureConfig{}, err
 			}
+
```
yanivagman, line 80: extra line

cmd/tracee-ebpf/internal/flags/flags-capture.go
```diff
@@ -73,21 +72,12 @@ func PrepareCapture(captureSlice []string) (tracee.CaptureConfig, error) {
 		} else if cap == "mem" {
 			capture.Mem = true
 		} else if strings.HasPrefix(cap, "net=") {
-			iface := strings.TrimPrefix(cap, "net=")
```
yanivagman, line 76: I think it is better to keep this in a sperate line, and trim "net=" and not just "net". Anyways you then trim "=" in `PaseIface` as well.
If doing so, don't forget to update the net filter as well to trim "net="

pkg/ebpf/filters.go
```diff
@@ -643,3 +645,29 @@ func (filter *ContIDFilter) FilterOut() bool {
...
+		return fmt.Errorf("invalid network interface: %s", iface)
+	}
+	found := false
+	// Check if we already have this interface
+	for _, item := range *ifaces {
+		if iface == item {
+			found = true
+			break
+		}
+	}
```
yanivagman, line 668: I think you can use the findInList function youv'e added instead

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4360,19 +4363,22 @@ static __always_inline int tc_probe(struct __sk_buff *skb, bool ingress) {
     //
     // See bpf_skb_event_output in net/core/filter.c.
     u64 flags = BPF_F_CURRENT_CPU;
-    flags |= (u64)skb->len << 32;
-    if (get_config(CONFIG_OPTIONS) & OPT_DEBUG_NET){
+
+    // If not capture, only send the minimal required data to save the
```
yanivagman, line 4367: Why do you write "If not capture"? Shouldn't it be "if net_packet event wasn't chosen"?

pkg/ebpf/tracee.go
```diff
@@ -1076,3 +1091,17 @@ func (t *Tracee) invokeInitEvents() {
...
+	return findInList(ifaceName, &t.config.Capture.NetIfaces)
+}
+func findInList(element string, list *[]string) int {
+	for idx, name := range *list {
+		if name == element {
+			return idx
+		}
+	}
+	return -1
+}
```
yanivagman, line 1107: As we're in go and not in C, it would be better to return an error instead of -1.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -214,6 +218,7 @@ Copyright (C) Aqua Security inc.
 #define OPT_CAPTURE_MODULES             (1 << 6)
 #define OPT_CGROUP_V1                   (1 << 7)
 #define OPT_PROCESS_INFO                (1 << 8)
+#define OPT_CAPTURE_PACKETS             (1 << 9)
```
itamarmaouda101, line 221: please remove

pkg/ebpf/tracee.go
```diff
@@ -284,6 +308,7 @@ func New(cfg Config) (*Tracee, error) {
 	if t.config.maxPidsCache == 0 {
 		t.config.maxPidsCache = 5
 	}
+
```
itamarmaouda101, line 311: nit: extra line

pkg/ebpf/tracee.go
```diff
@@ -370,6 +385,7 @@ const (
 	optCaptureModules
 	optCgroupV1
 	optProcessInfo
+	optCaptureNet
```
itamarmaouda101, line 388: please remove

pkg/ebpf/tracee.go
```diff
@@ -428,6 +444,7 @@ func (t *Tracee) getOptionsConfig() uint32 {
 	}
 	if t.config.Capture.NetIfaces != nil || t.config.Debug {
 		cOptVal = cOptVal | optProcessInfo
+		cOptVal = cOptVal | optCaptureNet
```
itamarmaouda101, line 447: remove

pkg/ebpf/net_events.go
```diff
@@ -230,40 +241,50 @@ func (t *Tracee) processNetEvents(ctx gocontext.Context) {
...
-					// output the event
-					select {
-					case t.config.ChanEvents <- evt:
-						t.stats.NetEvCount.Increment()
-					case <-ctx.Done():
-						return
+				ifaceName := t.netInfo.ifaces[int(netCaptureData.InterfaceIndex)].Name
+				ifaceIdx, err := t.getTracedIfaceIdx(ifaceName)
+				if err == nil {
+					if t.eventsToTrace[netEventMetadata.NetEventId] && ifaceIdx >= 0 {
```
AlonZivony, line 247: I am a bit confused from the check of `ifaceIdx >= 0`.
Isn't it given if there is no error?

pkg/ebpf/net_events.go
```diff
@@ -230,40 +241,50 @@ func (t *Tracee) processNetEvents(ctx gocontext.Context) {
...
+						case <-ctx.Done():
+							return
+						}
 					}
 				}
 
-				// capture the packet
-				packetBytes, err := getPacketBytes(netDecoder, netCaptureData.PacketLength)
+				ifaceName = t.netInfo.ifaces[int(netCaptureData.InterfaceIndex)].Name
+				ifaceIdx, err = t.getCapturedIfaceIdx(ifaceName)
```
AlonZivony, line 264: You use here the same names too much.
There is 2 indexes here and the difference is not clear.
I suggest to call this one `configCaptureIfaceIndex`

pkg/ebpf/net_events.go
```diff
@@ -230,40 +241,50 @@ func (t *Tracee) processNetEvents(ctx gocontext.Context) {
...
 				if err != nil {
-					t.handleError(err)
 					continue
 				}
-				if err := t.writePacket(netCaptureData, time.Unix(0, int64(netEventMetadata.TimeStamp)), packetContext, packetBytes); err != nil {
-					t.handleError(err)
-					continue
+				if ifaceIdx >= 0 {
```
AlonZivony, line 268: Again, it is weird to check it if there was not error (unless you expect to get -1 somehow?)

cmd/tracee-ebpf/internal/flags/flags-filter.go
```diff
@@ -208,6 +211,14 @@ func PrepareFilter(filters []string) (tracee.Filter, error) {
 			continue
 		}
 
+		if strings.HasPrefix(filterName, "net") {
```
AlonZivony, line 214: It is nice to add a filter, but if you don't write in the help message that the user needs to write `net=<iface>` no one will know how to use it...

pkg/ebpf/tracee.go
```diff
@@ -831,28 +859,29 @@ func (t *Tracee) initBPF() error {
...
-				return err
-			}
-
-			tcProbe := netProbe{
-				ingressHook: ingressHook,
-				egressHook:  egressHook,
-			}
-			t.tcProbe = append(t.tcProbe, tcProbe)
+	if t.eventsToTrace[NetPacket] && len(t.config.Filter.NetFilter.InterfacesToTrace) == 0 {
+		return fmt.Errorf("couldn't trace %s event, missing interface", EventsDefinitions[NetPacket].Name)
```
AlonZivony, line 863: This is the error message I get if I try to trace the `net_event` event without using the `-t net=<iface`.
This is not a clear message to get, because I don't get from it that I should add a tracing interface.
Try to catch the case of missing tracing interfaces before this point and giving a self clear error message for the user to understand what he should do to solve the problem.

AlonZivony (changes requested):
The PR seems really good to me, only a few touches left to do:
1. Rearrange the commits to more logical separation. There is a commit with 3 sub issues in it, which is weird when commit should do 1 thing, and there are the commits that are called "Fixed CR"
2. Make the UI more user friendly. The user don't really know how to use the new `net_event` event, and you never print an explanation to him. It is also very unclear what is the new `net_event` - which net packets are captured???

pkg/ebpf/events_definitions.go
```diff
@@ -6178,4 +6187,14 @@ var EventsDefinitions = map[int32]EventDefinition{
 			{Type: "unsigned long", Name: "ctime"},
 		},
 	},
+	NetPacket: {
+		ID32Bit:      sys32undefined,
+		Name:         "net_packet",
+		Probes:       []probe{},
+		Dependencies: []dependency{},
+		Sets:         []string{},
```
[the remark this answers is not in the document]
itamarmaouda101: I talked with @yanivagman that we will add the `network_events` set after we will have more net events (DNS, IRC, etc.)

cmd/tracee-ebpf/internal/flags/flags-filter.go
```diff
@@ -37,6 +37,8 @@ The field 'set' selects a set of events to trace according to predefined sets, w
 
 The special 'follow' expression declares that not only processes that match the criteria will be traced, but also their descendants.
 
+The field 'net' is required when tracing network events. without it, tracee wouldn't know which interface to sniff over
```
AlonZivony, line 40: The documentation here should specify what the `net` filter is, not just that it is a mandatory field for net events.
I recommend writing the following:
```suggestion
The field 'net' specifies which interfaces to monitor when tracing network events.
Notice that the 'net' field is mandatory when tracing network events.
````

File: docs/tracee-ebpf/trace-options.md [hunk 24]
```diff
@@ -196,3 +196,8 @@ trace all events that originated from bash or from one of the processes spawned
 ```
 --trace comm=bash --trace follow
 ```
+trace the network events over a given interface name
```
AlonZivony, line 199: The convention is to put a blank line before each doc line, so follow the convention.

pkg/bufferdecoder/net_decoder.go
```diff
@@ -33,8 +33,8 @@ func (decoder *EbpfDecoder) DecodeNetEventMetadata(eventMetaData *NetEventMetada
 }
 
 type NetCaptureData struct {
-	PacketLength   uint32 `json:"pkt_len"`
-	InterfaceIndex uint32 `json:"if_index"`
+	PacketLength            uint32 `json:"pkt_len"`
+	CaptureConfigIfaceIndex uint32 `json:"if_index"`
```
AlonZivony, line 37: The "Capture" part is not necessary in my opinion, because it is part of the struct name

AlonZivony (changes requested):
Great job!
Some really minor textual stuff left and I think it is good to go.
Remember to fix the commits (or do you prefer waiting for finale approve before it?)

pkg/ebpf/tracee.go
```diff
@@ -87,7 +87,7 @@ func (tc Config) Validate() error {
 		if _, ok := EventsDefinitions[e]; !ok {
 			return fmt.Errorf("invalid event to trace: %d", e)
 		}
-		if strings.Compare(EventsDefinitions[e].Name, EventsDefinitions[NetPacket].Name) == 0 {
+		if strings.Compare("net_packet", EventsDefinitions[e].Name) == 0 {
```
yanivagman, line 90: You can simply compare e to NetPacket instead

yanivagman (approved):
Ok, LGTM!
Nice work on this one @itamarmaouda101

PR #1516, opened 2022-02-27
protocol: encapsulate custom event headers

NDStrahilevitz:
In continuation of this from #1482:

> > In addition:
> > 
> > 1. in protocol.go we are exposing functions like `Origin`, `ContentType`, `Header` but the fields of `Event` and `EventHeaders` are also exported. What am I missing?
> > 2. in particular, `EventHeaders.Custom` is exported and a user might use the library wrongly and addressing to a `nil` `Custom` field causing a program crash/panic.
> 
> 1. The functions are there to ensure an abstracted access to those headers. IMO I'm perfectly fine with unexposing the headers and providing access only through receivers. However, that would complicate event building so maybe we can suffice using just the suggestion below (2)
> 2. Then as part of what I wrote above, at a minimum we could unexpose the`Custom` field and expose a `SetHeader` method

This PR unexports the new `Custom` field in `protocol.EventHeaders` and adds a method to set a custom header. This is done to ensure safe access to custom event headers.

### DIFF
--- a/types/protocol/protocol.go
+++ b/types/protocol/protocol.go
@@ -7,7 +7,7 @@ type EventHeaders struct {
 	// Origin indicates whether Event originates from host or container.
 	Origin string
 	// Custom additional custom headers, nil most of the time
-	Custom map[string]string
+	custom map[string]string
 }
 
 //Event is a generic event that the Engine can process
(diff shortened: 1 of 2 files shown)


mtcherni95 (commented):
let's put the pr in draft mode and promote only after https://github.com/aquasecurity/tracee/pull/1482

NDStrahilevitz:
> let's put the pr in draft mode and promote only after #1482

No need to put to draft, this doesn't affect #1482 and we could easily rebase it if needed. If anything, it would be better if this gets merged before so I can point #1482 to this version of `types`, but it's not critical.

danielpacak (approved):

Issue #1509, opened 2022-02-23
Add types test to `make test`


NDStrahilevitz:
Currently the test script in the Makefile only tests tracee-rules, tracee-ebpf and the signatures. Since #1491, the tests that were previously in `external` have moved to `types/trace`. As such we need to add the `types` package to the default test script.

PR #1517, opened 2022-02-27
Makefile: add types module testing

NDStrahilevitz:
Solves #1509

This PR:
1. Adds a script(`test-types`) to test the `types` module since it was missing from tests.
2. Adds `test-types` to `make test`.
3. Adds a job to verify types in the PR github workflow.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -572,7 +572,8 @@ test-tracee-rules: \
 		-v \
 		./cmd/tracee-rules/... \
 		./pkg/rules/...
-
+	cd ./types; $(CMD_GO) test -v ./...
+	cd ..
 #
 # rules
 #


Makefile
```diff
@@ -621,12 +621,24 @@ test-rules: \
 #
 	$(CMD_OPA) test $(REGO_SIGNATURES_DIR) --verbose
 
+
+#
+# types
+#
+
+.PHONY: test-types
```
mtcherni95, line 629: you need to update also the `help`

mtcherni95 (changes requested):

Makefile
```diff
@@ -621,12 +621,24 @@ test-rules: \
...
+test-types: \
+	.checkver_$(CMD_GO)
+#
+	cd ./types; $(CMD_GO) test -v ./trace/...; $(CMD_GO) test -v ./protocol/...; $(CMD_GO) test -v ./detect/...
```
mtcherni95, line 634:
```suggestion
	cd ./types; $(CMD_GO) test -v ./...
```

NDStrahilevitz: Thanks I couldn't find a script like that which worked haha

Makefile
NDStrahilevitz: done

Makefile
```diff
@@ -621,12 +622,24 @@ test-rules: \
...
 
+
+#
+# types
+#
+
+.PHONY: test-types
+test-types: \
+	.checkver_$(CMD_GO)
+#
```
mtcherni95, line 633: empty line

NDStrahilevitz: That empty line with # is there in almost all other scripts.

rafaeldtinoco: Yep, this is following the format I proposed to Makefiles. I'll review this tomorrow morning.

mtcherni95: ohh didn't know that, my bad :) thanks @rafaeldtinoco

Makefile
```diff
@@ -621,12 +622,24 @@ test-rules: \
 #
 	$(CMD_OPA) test $(REGO_SIGNATURES_DIR) --verbose
 
+
+#
+# types
```
rafaeldtinoco, line 627: Despite this change being functional and working good, I have concerns on "test-types". What does it mean for the end user ? What is "types" ?  We had, so far, a clear distinction between "tracee-ebpf", "tracee-rules" and "rules", thus the 3 different tests. 

Shouldn't we add the types test, for now, to "test-tracee-ebpf" OR "test-tracee-rules" ? And, after we're done with making "tracee-ebpf + tracee-rules" a single program, then we can have all tests in a single "test target".

WDYT ?

NDStrahilevitz: I'm not sure it fits either programs, `types` defines their protocol. We could split the inner packages between testing ebpf and rules but that seems ugly.

I agree we can find a better solution once we merge the two into a single program, but rn I can't really thing of a "prettier" solution, because there are certainly tests here that need to run.

rafaeldtinoco: I don't want to block the change, just change it to make more sense. What if we change test-types for something more generic. Check "test-tracee-ebpf"... it contains:

```
./cmd/tracee-ebpf/...
./pkg/...
```

@danielpacak had suggested we included ./pkg/ tests in tracee-ebpf for now and later organized the pkg tests into a single "test" target. @danielpacak WDYT here ?

rafaeldtinoco: BTW, isn't this already being tested by that target ? with "./pkg/..." ?

NDStrahilevitz: > BTW, isn't this already being tested by that target ? with "./pkg/..." ?

No since this is in a different module.
I guess i'd slightly lean on this being in tracee-rules tests, but since `trace` is relevant to tracee-ebpf that doesn't make too much sense either. I can't really see this not being tested seperately for now.

Maybe we need to split tests to ebpf cmd rules cmd and pkg and we can add this under pkg? But that's a larger scope than this PR.

rafaeldtinoco: yes, let's do the following: to add the test into an existing target (tracee-ebpf or tracee-rules) and open an issue to either:

1. create a single test target containing all needed tests
2. or to create multiple targets to each pkg (with a broader one executing them all)

WDYT ?

rafaeldtinoco: (then you can get rid of the workflow changes as well)

NDStrahilevitz: 1. That exists in `make test` which just composes all the other test targets.

I think we just missed restructuring the tests in terms of the new latest structure. Again that would be outside this scope. In terms of the current makefile structure I think how it's done here makes the most sense (at least short term). A PR fixing the entire structure could use the changes done here and reimplement them but I fear if we don't merge this now it will be missed along the way later on.

If we have to include this in one of the other test targets... I guess it would be tracee rules, but I think it should be kept in it's own target.

rafaeldtinoco: Okay, we've synced on Teams about this. We agreed you would include the test in `test-tracee-rules` target and I'm going to open an issue to reformulate the tests targets with the new project structure (since `test-tracee-ebpf` and `test-tracee-rules` are not reflecting the project structure we have today).

rafaeldtinoco: https://github.com/aquasecurity/tracee/issues/1522

PR #1525, opened 2022-03-02
decoding: cast SlimCred to trace.SlimCred

NDStrahilevitz:
This fixes an issue with gob encoding where `trace.SlimCred` was registered but when we parse the SlimCred argument it's parsed as `bufferdecoder.SlimCred`.

The fix simply adds a cast after the decoding, which should break on compile time if either of the contracts change.

### DIFF
--- a/pkg/bufferdecoder/eventsreader.go
+++ b/pkg/bufferdecoder/eventsreader.go
@@ -93,7 +93,7 @@ func ReadArgFromBuff(ebpfMsgDecoder *EbpfDecoder, params []trace.ArgMeta) (trace
 	case credT:
 		var data SlimCred
 		err = ebpfMsgDecoder.DecodeSlimCred(&data)
-		res = data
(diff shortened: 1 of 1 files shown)


AsafEitani:
Damn that's a small fix XD
LGTM

AsafEitani (approved):

yanivagman (approved):
LGTM

pkg/bufferdecoder/eventsreader.go
```diff
@@ -93,7 +93,7 @@ func ReadArgFromBuff(ebpfMsgDecoder *EbpfDecoder, params []trace.ArgMeta) (trace
 	case credT:
 		var data SlimCred
 		err = ebpfMsgDecoder.DecodeSlimCred(&data)
-		res = data
+		res = trace.SlimCred(data)
```
mtcherni95, line 96: let's add a comment explaining why

PR #1527, opened 2022-03-03
Michael default caching

mtcherni95:


pkg/events/queue/queue_mem_list.go
```diff
@@ -0,0 +1,118 @@
...
+		return amountInKB * 1024
+	}
+	MBtoKB := func(amountInMB int) int {
+		return amountInMB * 1024
+	}
+	GBtoMB := func(amountInGB int) int {
+		return amountInGB * 1024
+	}
+	AmountOfEvents := func(amountInMB int) int {
+		return MBtoKB(KBtoB(amountInMB)) / eventSize
```
mtcherni95, line 97: @rafaeldtinoco FYI shouldn't this be`KbtoB(MBtoKB(amountInMB))?`

mtcherni95: btw, asking you as was part of your commits.

pkg/events/queue/queue_mem_list.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	// EventsCacheMemSize was provided, return exact amount of events for it
+	if q.eventsCacheMemSizeMB > 0 {
+		return AmountOfEvents(q.eventsCacheMemSizeMB)
+	}
+
+	switch {
```
mtcherni95, line 105: we get here only if `q.eventsCacheMemSizeMB < = 0` and the only relevant switch case will always be the first one (`case q.eventsCacheMemSizeMB <= GBtoMB(1)`) . what am I missing? again, asking it here as this was part of your commits @rafaeldtinoco and maybe I am loosing something important in the logics.

rafaeldtinoco: If you provide a cache size ("mem-cache-size" cmdline option), then you know what you are doing... q.eventsCacheMemSizeMB won't be 0 (default) and I'll return the amount of events for the given cache size (no automatic cache size calculation is needed).

Now... if you ONLY provided "cache-type", but no "mem-cache-size", then q.eventsCacheMemSizeMB will be 0 AND tracee calculates the amount of events for a specific cache size. Cache size depends on the amount of existing host/node memory.

pkg/events/queue/queue_mem_list.go
rafaeldtinoco: could be if eventSize was 1 and not 1024 (if I got what you mean correctly). 

If I did that, then eventSize would always have to be 1024 multiple (1k, 2k, 4k). The way it is, number can be 768 or 1280 (which makes a big difference at the end).

The most correct thing to do here would be to calculate the eventSize based on an existing number (trace.Event) BUT I'm afraid that it is more complex than just that (considering garbage collection and many other overhead from runtime). 

I measured different cache sizes and found out 1024 was something close to a constant for calculating cache sizes VS amount of events (this won't be true if we start having other event types in the future).

pkg/events/queue/queue_mem_list_test.go
```diff
@@ -0,0 +1,34 @@
+package queue
```
rafaeldtinoco, line 1: well done, thanks for that =).

rafaeldtinoco (approved):
Your changes look good to me. If you're satisfied with my comments then feel free to merge. If not, then we can discuss things further.

rafaeldtinoco:
BTW, @mtcherni95, whenever you do git logs/commits, always use imperative wording:

1. Limit the subject line to 50 characters.
2. Capitalize only the first letter in the subject line.
3. Don't put a period at the end of the subject line.
4. Put a blank line between the subject line and the body.
5. Wrap the body at 72 characters.
6. Use the imperative mood.
7. Describe what was done and why, but not how.

Cheers!

pkg/events/queue/queue_mem_list.go
mtcherni95: what I am saying is that your are using function `KBtoB(amountInMB)` and accepting something which is in units of MB. I don't see any constellation of input where this could be correct. I think should simply be the opposite: 
`KbtoB(MBtoKB(amountInMB))`

pkg/events/queue/queue_mem_list.go
mtcherni95: not sure we understood each other. the logics before checks 
```
if q.eventsCacheMemSizeMB > 0 {
		return AmountOfEvents(q.eventsCacheMemSizeMB)
	}
```
if it doesn't enter in the if statement will ALWAYS enter in this case case `q.eventsCacheMemSizeMB <= GBtoMB(1):` (as q.eventsCacheMemSizeMB will ALWAYS be <= 0).

so what's the point of the other switch conditions?

grantseltzer:
@mtcherni95 Is this ready to merge? It seems like there's still discussion.

mtcherni95:
> @mtcherni95 Is this ready to merge? It seems like there's still discussion. 

Hi, I'd prefer to wait for Rafael's answer first.

Issue #1529, opened 2022-03-03
pcap files are not being closed


roikol:
thanks @OriGlassman.

pcap files are not being close on netExit, just deleted from the mapping. after some time of capturing network traffic, this could result in error of "too many files open".

yanivagman:
Closed by #1530

PR #1530, opened 2022-03-03
close pcap files instead of just deleting from mapping

roikol:
solves #1529.

introducing new struct `netPcap` to hold both the writer, and the file object. this way, on netExit, we can access the file object and close it (note: flushing the ngWriter is adviced before closing the underline file).

also, changed name of Tracee struct field: netPcap to netCapture.

### DIFF
--- a/pkg/ebpf/net_events.go
+++ b/pkg/ebpf/net_events.go
@@ -19,13 +19,18 @@ import (
 	"github.com/google/gopacket/pcapgo"
 )
 
+type netPcap struct {
+	FileObj *os.File
+	Writer  *pcapgo.NgWriter
+}
+
 type netInfo struct {
 	mtx           sync.Mutex
-	pcapWriters   map[processPcapId]*pcapgo.NgWriter
+	pcapWriters   map[processPcapId]netPcap
 	ngIfacesIndex map[int]int
 }
 
-func (ni *netInfo) GetPcapWriter(id processPcapId) (*pcapgo.NgWriter, bool) {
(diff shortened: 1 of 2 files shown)


pkg/ebpf/net_events.go
```diff
@@ -137,15 +150,15 @@ func (t *Tracee) createPcapFile(pcapContext processPcapId) error {
 	if err != nil {
 		return err
 	}
-	t.netPcap.SetPcapWriter(pcapContext, pcapWriter)
+	t.netCapture.SetPcapWriter(pcapContext, netPcap{pcapFile, pcapWriter})
```
mtcherni95, line 153: Better to create structs with explicit labeling:
```
netPcap {
    FileObj: pcapFile, 
    Writer: pcapWriter,
}
```

pkg/ebpf/net_events.go
```diff
@@ -302,22 +315,22 @@ func (t *Tracee) writePacket(capData CaptureData, timeStamp time.Time, packetCon
...
 	if !pcapWriterExists {
 		err := t.createPcapFile(packetContext)
 		if err != nil {
 			return err
 		}
 	}
 
-	writer, _ := t.netPcap.GetPcapWriter(packetContext)
-	err := writer.WritePacket(info, dataBuff.Bytes()[:capData.PacketLen])
+	pcap, _ := t.netCapture.GetPcapWriter(packetContext)
```
mtcherni95, line 326: we are calling twice `t.netCapture.GetPcapWriter(packetContext)`, why?

yanivagman (approved):
LGTM

PR #1531, opened 2022-03-06
Produce protocol events in tracee-ebpf

### LINKED ISSUE #1266: extract package `eventprinter` from `printer.go`
from #960 :

> printer.go - create new package `eventprinter`. Factory function moves to `main.go`

### LINKED ISSUE #1524: Produce `protocol.Event` in tracee-ebpf
Currently, `trace.Event`s are being converted to `protocol.Event` on consuming inside `tracee-rules` itself. As was already mentioned in #1482 this logic is backwards, but was a continuation of what was done before. Now this should be fixed.

Fixing this has two parts:
1. Requiring that events are already sent as `protocol.Event` on arrival to `tracee-rules` (so no conversions should be required at all).
2. Implement sending `protocol.Event`s from `tracee-ebpf`. I think this could be done with a new `output` flag, however there could other better ways which is why i'm opening the issue.

NDStrahilevitz:
This PR fixes a technical debt from #1482 where events where converted to `protocol` in tracee-rules's input instead of tracee-ebpf's output.

As part of the PR:
1. Separates the `EventPrinter` to it's own package (solves #1266) and adds tests to it.
2. Solves #1524
3. Fix a bug with a leftover template string in `output.go`

Close #1524 
Close #1266


cmd/tracee-rules/input.go
```diff
@@ -52,7 +52,7 @@ func setupTraceeGobInputSource(opts *traceeInputOptions) (chan protocol.Event, e
 	res := make(chan protocol.Event)
 	go func() {
 		for {
-			var event trace.Event
+			var event protocol.Event
```
mtcherni95, line 55: I might be missing something. If event is not sent in "protocol format", will we fail in decoding? is this intentional?

cmd/tracee-ebpf/internal/printer/printer_test.go
```diff
@@ -0,0 +1,86 @@
+package printer_test
```
mtcherni95, line 1: 1. package name should be lowercase. Don't use snake_case or camelCase in package names.
2. does this code compile? in the same folder you two different packages (!?)
3. the package name in this case should simply be `printer`

cmd/tracee-ebpf/internal/printer/printer.go
```diff
@@ -29,51 +29,64 @@ type eventPrinter interface {
 	Close()
 }
 
-func newEventPrinter(kind string, containerMode bool, relativeTS bool, out io.WriteCloser, err io.WriteCloser) (eventPrinter, error) {
+type Config struct {
```
mtcherni95, line 32: great

cmd/tracee-ebpf/internal/printer/printer.go
```diff
@@ -29,51 +29,64 @@ type eventPrinter interface {
...
+	ContainerMode bool
+	RelativeTS    bool
+	Protocol      bool
+}
+
+func New(config Config) (eventPrinter, error) {
 	var res eventPrinter
-	var initError error
+	kind := config.Kind
 	switch {
```
mtcherni95, line 46:
```suggestion
	switch config.Kind {
	case "ignore":
	....
```

cmd/tracee-ebpf/internal/flags/flags-output.go
```diff
@@ -29,6 +30,7 @@ option:{stack-addresses,detect-syscall,exec-env,relative-time,exec-hash,parse-ar
   exec-hash                                        when tracing sched_process_exec, show the file hash(sha256) and ctime
   parse-arguments                                  do not show raw machine-readable values for event arguments, instead parse into human readable strings
   sort-events                                      enable sorting events before passing to them output. This will decrease the overall program efficiency.
+  protocol                                         write output in compatible format for tracee-rules (relevant only when sending as json or gob)
```
mtcherni95, line 33: ugly-ish json or gob only limitation... thinking if we can do something better...

cmd/tracee-ebpf/internal/printer/printer.go
```diff
@@ -29,51 +29,64 @@ type eventPrinter interface {
...
+	OutPath       string
+	OutFile       io.WriteCloser
+	ErrPath       string
+	ErrFile       io.WriteCloser
+	ContainerMode bool
+	RelativeTS    bool
+	Protocol      bool
+}
+
+func New(config Config) (eventPrinter, error) {
```
mtcherni95, line 43: like 👍

cmd/tracee-rules/input.go
NDStrahilevitz: Yes. Notice that before we would fail on any event that wasn't a `trace.Event` so this is an improvement.

cmd/tracee-ebpf/internal/printer/printer_test.go
NDStrahilevitz: 1. This is what the vscode plugin suggested by default, I have no issue with changing it.
2. Yes, it compiles and is included in tests.
3. I had to separate it due to some circular dependency issue that would happen otherwise. The test code works as expected, I can change the package name if it's important.

cmd/tracee-ebpf/internal/printer/printer.go
NDStrahilevitz: I would still like to keep it in a variable for later use in printing and such, so this doesn't really make the code cleaner later IMO.

cmd/tracee-ebpf/internal/flags/flags-output.go
NDStrahilevitz: It wouldn't be relevant in any other case (in Open Source), since messaging is done only in those formats. I understand not liking the lack of genericity but so far I haven't come with a better idea (without actually using generics which don't yet exist).

cmd/tracee-ebpf/internal/printer/printer_test.go
NDStrahilevitz: > yeah let's use `printer`

We can't though, that's my point, we need to use something else otherwise we get a circular dependency (this is because the printer config is parsed from the output flag).

If we want to keep the tests in the same package, I can move the config back to `flags` but I think it makes less sense than just keeping the test package separate.

cmd/tracee-ebpf/internal/printer/printer.go
NDStrahilevitz: If we use the switch the way you suggest I need to check for the gotemplate path first since the `strings.HasPrefix` can't be compared to an actual string, so I would have to check it in an `if` statement first, and then have the actual `kind` switch in an else statement... 
I understand wanting to clean up the switch but there's no way to actually do so without getting rid of the gotemplate path (which would basically be like getting rid of the feature for no reason), otherwise we get even worse code IMO.

builder/entrypoint.sh
```diff
@@ -70,7 +70,7 @@ run_tracee_rules() {
     events=$($TRACEE_RULES_EXE --list-events)
 
     echo "INFO: starting tracee-ebpf..."
-    ${TRACEE_EBPF_EXE} --metrics --output=format:gob --output=option:parse-arguments --trace event=${events} --output=out-file:${TRACEE_PIPE} &
+    ${TRACEE_EBPF_EXE} --metrics --output=format:gob --output option:protocol --output=option:parse-arguments --trace event=${events} --output=out-file:${TRACEE_PIPE} &
```
danielpacak, line 73: Do we need this new option after all? What would be a use case to not passing the `--ouput=option:protocol` flag to tracee-ebpf? I though that tracee-ebpf is going to send generic event and tracee-rules is going to consume generic event.

```suggestion
    ${TRACEE_EBPF_EXE} --metrics --output=format:gob --output=option:protocol --output=option:parse-arguments --trace event=${events} --output=out-file:${TRACEE_PIPE} &
```

danielpacak (changes requested):
👋 @NDStrahilevitz I believe we must rebase this PR to fix tests (mount /boot path in integration tests). Also, when I run tracee and trigger detection I'm getting the following error:

```console
$ sudo ./dist/tracee-ebpf \
  --output=format:gob \
  --output=option:protocol \
  --output=option:parse-arguments \
  | ./dist/tracee-rules \
  --input-tracee=file:stdin \
  --input-tracee=format:gob

*** Detection ***
Time: 2022-03-10T09:04:24Z
Signature ID: TRC-2
Signature: Anti-Debugging
Data: map[]
Command: 2022/03/10 09:04:24 error writing to output: template: default:7:20: executing "default" at <.Context.ProcessName>: can't evaluate field Context in type detect.Finding
```

builder/entrypoint.sh
NDStrahilevitz: I thought that it would be the outputting client's responsibility to send a protocolized event, not the producer itself (in this case i'm making the distinction between the cmd/tracee-ebpf program and the Tracee struct itself). 

I would like the protocol toggle so I could for example dump the events into a file, without all the Headers metadata around it (doubly so if and when we add more headers). Maybe it would be more appropriate to send as protocol by default and disable it if wanted, but I think it would make even less sense since it would mean by default Table and JSON/GOB formats have different behaviours (table would in any case output the raw event).

Personally, I think the best solution would have been to separate formatting and type output decision, but there's no clean way to do it without generics (not yet in Go) or `interface{}` which would be inefficient.

NDStrahilevitz:
> 👋 @NDStrahilevitz I believe we must rebase this PR to fix tests (mount /boot path in integration tests). Also, when I run tracee and triggered detection I got the following error:
> 
> ```
> $ sudo ./dist/tracee-ebpf \
>   --output=format:gob \
>   --output=option:protocol \
>   --output=option:parse-arguments \
>   | ./dist/tracee-rules \
>   --input-tracee=file:stdin \
>   --input-tracee=format:gob
> Loaded 14 signature(s): [TRC-1 TRC-13 TRC-2 TRC-14 TRC-3 TRC-11 TRC-9 TRC-4 TRC-5 TRC-12 TRC-8 TRC-6 TRC-10 TRC-7]
> 
> *** Detection ***
> Time: 2022-03-10T09:04:24Z
> Signature ID: TRC-2
> Signature: Anti-Debugging
> Data: map[]
> Command: 2022/03/10 09:04:24 error writing to output: template: default:7:20: executing "default" at <.Context.ProcessName>: can't evaluate field Context in type detect.Finding
> ```

Ugh, I know how I missed this in the previous headers PR, there's a code embedded default GoTemplate in `output.go` that wasn't updated to use the new `Detection.Event.Payload` field... Speaking of, there's some technical debt with regards to only support `trace.Event` here, so we should fix that, i'll open a PR/issue about it later.

Anyway i'll commit a fix to that here.

cmd/tracee-ebpf/internal/printer/printer.go
```diff
@@ -198,16 +211,23 @@ func (p templateEventPrinter) Close() {
...
 func (p jsonEventPrinter) Init() error { return nil }
 
 func (p jsonEventPrinter) Preamble() {}
 
 func (p jsonEventPrinter) Print(event trace.Event) {
-	eBytes, err := json.Marshal(event)
+	var err error
+	var eBytes []byte
+	if p.protocol {
+		eBytes, err = json.Marshal(event.ToProtocol())
```
danielpacak, line 227: This still looks to me like a workaround. Why do we perform conversion from a specific to generic event type so late at print time? Shouldn't we update the eventPrinter interface and adapt the upstream code as well? In particular, send protocol.events to the `cfg.ChanEvents`? 

```go
type eventPrinter interface {
	// Print prints a single event
	Print(event protocol.Event)

	// ... other methods trimmed ... //
```

Now passing generic event to printer implementors, will allows us to keep GOB or JSON printers intact (by just taking the Payload), but at the same time will allow us to enhance them with event metadata if we like. And still I think that the protocol flag passed to tracee-ebpf in unnecessary. We have tens of flags already.

danielpacak (commented):
I'm still not sure this is the right way to pass events around tracee-ebpf and tracee-rules. Please see my comment about the eventPrinter interface.

cmd/tracee-ebpf/internal/printer/printer.go
NDStrahilevitz: The problem with this way is that json or gob printers won't just print out the payload, we would have to explicitly tell them to do so inside their implementations (since the event would be passed with metadata), basically just doing the opposite of the protocol flag - so we would still have to define a way to allow them to be printed without the headers.

Also by passing `protocol.Event` instead of `trace.Event` how will it affect all the other printers? They would have to check if the payload is actually a `trace.Event` and only then print (for example the table printer), are we okay with that?

About having `cfg.ChanEvents` send the events as protocol already, that would mean any other consumer would have to either support our message format, or convert to their own protocol with a middleware, not so bad, but if we only output a single event it seems redundant to me. Maybe if we started outputting different payloads it would make sense instead of emitting from multiple channels.

NDStrahilevitz:
@danielpacak and I just synced over this, we've agreed to see how this PR would look with tracee-ebpf outputting `protocol.Event` directly.

cmd/tracee-ebpf/internal/printer/printer.go
```diff
@@ -1,4 +1,4 @@
-package main
+package printer
```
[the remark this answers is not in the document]
NDStrahilevitz: I think it's a bit of a use case specific thing for the cmd, don't see why we would need to import it somewhere else. I don't have much of a strong opinion on this though.

cmd/tracee-ebpf/internal/printer/printer.go
```diff
@@ -198,16 +211,23 @@ func (p templateEventPrinter) Close() {
...
 
 func (p jsonEventPrinter) Print(event trace.Event) {
-	eBytes, err := json.Marshal(event)
+	var err error
+	var eBytes []byte
+	if p.protocol {
+		eBytes, err = json.Marshal(event.ToProtocol())
+	} else {
+		eBytes, err = json.Marshal(event)
+	}
```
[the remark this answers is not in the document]
NDStrahilevitz: I personally thought it would be useful for the case where we want to dump the only the payload itself, unrelated to sending to tracee-rules. Currently it seems though that the interface will change to receiving `protocol.Event` from the get go, so maybe we can just always print with the protocol.

cmd/tracee-ebpf/internal/flags/flags-output.go
NDStrahilevitz: I've removed it as an option in a new commit (`Print` now takes a `protocol.Event`), please take a look.

cmd/tracee-ebpf/internal/printer/printer.go
NDStrahilevitz: I've added a commit so that Print takes a `protocol.Event` instead of a `trace.Event` please take a look. I still have reservations about outputting `protocol.Event` directly from tracee-ebpf `ChanEvents`, I suggest we take it to this week's sync.

cmd/tracee-ebpf/internal/printer/printer.go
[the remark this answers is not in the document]
NDStrahilevitz: Resolved in latest commit

danielpacak (changes requested):
Good progress @NDStrahilevitz I'm glad that we remove the protocol flag :) I left few more comment to look at.

cmd/tracee-ebpf/internal/printer/printer.go
```diff
@@ -22,58 +23,71 @@ type eventPrinter interface {
...
+	OutPath       string
+	OutFile       io.WriteCloser
+	ErrPath       string
+	ErrFile       io.WriteCloser
+	ContainerMode bool
+	RelativeTS    bool
+	Debug         bool
+}
+
+func New(config Config) (eventPrinter, error) {
```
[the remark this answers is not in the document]
NDStrahilevitz: I generally don't like exporting stuff unless I know i'll use them outside of package scope, but I don't mind exporting an interface, I can do so if it's wanted.

cmd/tracee-ebpf/internal/printer/printer.go
```diff
@@ -105,32 +120,40 @@ func (p tableEventPrinter) Preamble() {
 	fmt.Fprintln(p.out)
 }
 
-func (p tableEventPrinter) Print(event trace.Event) {
-	ut := time.Unix(0, int64(event.Timestamp))
+func (p tableEventPrinter) Print(event protocol.Event) {
+	evt, ok := event.Payload.(trace.Event)
+
+	if !ok {
+		if p.debug {
```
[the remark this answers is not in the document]
NDStrahilevitz: I'm outputting to p.out here because this isn't really an error IMO. It's perfectly valid to output an event unrelated to the printer type, it's the printer's fault for not supporting it. As such I also guarded it under debug.

cmd/tracee-ebpf/internal/printer/printer.go
```diff
@@ -206,8 +229,10 @@ func (p jsonEventPrinter) Init() error { return nil }
 
 func (p jsonEventPrinter) Preamble() {}
 
-func (p jsonEventPrinter) Print(event trace.Event) {
-	eBytes, err := json.Marshal(event)
+func (p jsonEventPrinter) Print(event protocol.Event) {
```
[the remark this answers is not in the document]
NDStrahilevitz: You're right I forgot to change that code path, i'll fix it.

cmd/tracee-ebpf/internal/printer/printer.go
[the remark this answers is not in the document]
NDStrahilevitz: internal is useful to keep it unexposed outside of `cmd`. I've put this here for the same reason I moved the flags to `internal` before, it's usecase specific to the command.

Again, I don't have much of a strong opinion on this so I don't mind moving it, but just explaining my reasoning here.

cmd/tracee-ebpf/internal/printer/printer.go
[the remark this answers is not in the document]
danielpacak: But currently it's a warning. You either export both constructor and returned interface or keep both private.

cmd/tracee-ebpf/internal/printer/printer.go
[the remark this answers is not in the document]
danielpacak: It's like hiding logical errors that shouldn't normally happen and making them visible only if I enable the debug flag. Isn't it?

cmd/tracee-ebpf/internal/printer/printer.go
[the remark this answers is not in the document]
danielpacak: The pkg indicates any code that we intend to export and use somewhere else as Go module / library. Everything under cmd is implicitly "private" and only used to compile tracee-ebpf and tracee-rules executable binaries.

cmd/tracee-ebpf/internal/printer/printer.go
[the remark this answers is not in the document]
NDStrahilevitz: Why is it a logical error though? Theoretically `protocol.Event`s can have any kind of payload, it's the printers job to either be completely agnostic to it or report if it's unable to print the type. As far as I see it it's a completely valid code path. 

This is partially why I was hesitant on changing the `Print` signature.

cmd/tracee-ebpf/internal/printer/printer.go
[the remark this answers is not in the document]
NDStrahilevitz: Then I think we can move `flags` out of internal as well. Would that be in scope for this? I can do it later in a chore PR as well.

builder/entrypoint.sh
NDStrahilevitz: Resolving since the option was removed.

cmd/tracee-ebpf/internal/printer/printer.go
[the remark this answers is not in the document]
danielpacak: Theoretically it's protocol.Event, but in practice tracee-ebpf is supposed to emit trace.Event. If that's not the case we should fail fast and catch such errors.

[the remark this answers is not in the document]
NDStrahilevitz: Even so I don't think the printer is the place to fail on this (if it's something to fail on at all). Especially since the "table" usecase is just for data dumping. Other printers are perfectly fine not knowing the payload, for example the json and gob ones.
If we would have changed `cfg.ChanEvents` to `chan protocol.Event`, would you still be of the same opinion? Or if there were multiple output channels?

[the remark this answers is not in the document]
danielpacak: Getting this error in printer is extremely unlikely to happen. And it's extremely hard to understand why we show error only if the debug flag is enabled. I'll also stop here and let another reviewer resolve this conversation.

cmd/tracee-ebpf/internal/printer/printer.go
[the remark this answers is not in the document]
danielpacak: Yes, the same is applicable to internal/printer and internal/flags packages. The internal/flags was already there so maybe we can take it out of scope and have separate discussion. I don't want to make it my own preference,  but usually Go packages, including Go SDK, are quite flat.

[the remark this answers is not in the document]
NDStrahilevitz: So can we agree that for the sake of consistency in this PR i'll keep it here and later we can take it out? I'd prefer that over having one here and one there.

cmd/tracee-ebpf/internal/printer/printer.go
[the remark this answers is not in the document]
NDStrahilevitz: It's not an error it's just a limitation of the table printer. Imagine we would be outputting to table and in the meanwhile we would have background events which aren't `trace.Event`. Obviously we wouldn't want to print to table but there's nothing erroneous about the events happening, they are simply irrelevant to the logging method we chose. As such notifying the user they happened is, IMO, relevant only if he's debugging, otherwise the user probably wouldn't care at all about those events, and printing the error would just be a visual nuisance. 

Anyways, I think we both made our cases clear so I will also stop now and wait for another reviewer to resolve.

danielpacak (changes requested):
I had another round and the changes look good to me. Please update the branch with main so I can approve it.

yanivagman (approved):
LGTM!

NDStrahilevitz:
I've moved this to draft since this now fails unit tests. This is because type information can be lost in the case of json. In GOB i'm not sure what's the cause. Regardless it seems we need some kind of type registration here, which will probably require some new header. I'll move this back to review once a solution is found.

NDStrahilevitz:
I think this will be solved in the near future with the move to protobufs @rafaeldtinoco so i'm closing this.

PR #1532, opened 2022-03-06
tests: mount /boot for smoke tests

yanivagman:
Tests always fail with:
`error creating Tracee: could not read /boot/config-5.11.0-1028-azure: stat /boot/config-5.11.0-1028-azure: no such file or directory\n`

Fix this


mtcherni95 (approved):
LGTM

PR #1535, opened 2022-03-06
tracee-ebpf: fix container removal in cgroup v1

### LINKED ISSUE #1534: Missing container ids
Container removal didn't take into account cgroup hierarchy id, unlike cgroupMkdir (container creation).
Under cgroup v1 environments, this might cause running containers to be removed before they are actually removed. In case another container has the same cgroup id, but in another cgroup hierarchy, the removal of such a container will also remove the first one.

yanivagman:
Container removal didn't take into account cgroup hierarchy id, unlike
cgroupMkdir (container creation).
Under cgroup v1 environments, this might cause running containers to be
removed before they are actually removed. In case another container has
the same cgroup id, but in another cgroup hierarchy, the removal of such
a container will also remove the first one.

Fix this by updating the container maps for only one cgroup hierarchy (cpuset).

Close #1534

### DIFF
--- a/pkg/containers/containers.go
+++ b/pkg/containers/containers.go
@@ -41,11 +41,13 @@ type CgroupInfo struct {
 
 // InitContainers initializes a Containers object and returns a pointer to it.
 // User should further call "Populate" and iterate with Containers data.
-func InitContainers() *Containers {
+func InitContainers() (*Containers, error) {
 	cgroupV1 := false
 	if _, err := os.Stat("/sys/fs/cgroup/cgroup.controllers"); os.IsNotExist(err) {
 		cgroupV1 = true
-		initCgroupV1()
+		if err := initCgroupV1(); err != nil {
+			return nil, err
+		}
 	}
 
 	return &Containers{
@@ -53,11 +55,15 @@ func InitContainers() *Containers {
 		cgroupMP: "",
 		cgroups:  make(map[uint32]CgroupInfo),
 		mtx:      sync.RWMutex{},
-	}
+	}, nil
 }
 
 func (c *Containers) IsCgroupV1() bool {
 	return c.cgroupV1
+}
+
+func (c *Containers) GetCgroupV1HID() int {
+	return cgroupV1HierarchyID
 }
 
 // Populate populates Containers struct by reading mounted proc and cgroups fs.
@@ -71,14 +77,11 @@ func (c *Containers) Populate() error {
 }
 
 // getCgroupV1SSID finds cgroup v1 hierarchy ID.
-func initCgroupV1() {
-	// Set hierarchy ID to 1 as a fallback in case of an error
-	cgroupV1HierarchyID = 1
-
+func initCgroupV1() error {
 	cgroupsFile := "/proc/cgroups"
 	file, err := os.Open(cgroupsFile)
 	if err != nil {
-		return
+		return fmt.Errorf("could not open cgroups file %s: %w", cgroupsFile, err)
 	}
 	defer file.Close()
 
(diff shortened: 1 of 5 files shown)


pkg/containers/containers.go
```diff
@@ -41,25 +41,31 @@ type CgroupInfo struct {
...
 // InitContainers initializes a Containers object and returns a pointer to it.
 // User should further call "Populate" and iterate with Containers data.
-func InitContainers() *Containers {
+func InitContainers() (*Containers, error) {
 	cgroupV1 := false
 	if _, err := os.Stat("/sys/fs/cgroup/cgroup.controllers"); os.IsNotExist(err) {
 		cgroupV1 = true
-		initCgroupV1()
+		if err := initCgroupV1(); err != nil {
+			return nil, nil
```
grantseltzer, line 49: Shouldn't this return an error? If not perhaps a debug statement?

pkg/containers/containers.go
```diff
@@ -71,14 +77,11 @@ func (c *Containers) Populate() error {
...
-func initCgroupV1() {
-	// Set hierarchy ID to 1 as a fallback in case of an error
-	cgroupV1HierarchyID = 1
-
+func initCgroupV1() error {
 	cgroupsFile := "/proc/cgroups"
 	file, err := os.Open(cgroupsFile)
 	if err != nil {
-		return
+		return err
```
grantseltzer, line 84:
```suggestion
		return fmt.Errorf("could not open cgroups file %s: %w", cgroupsFile, err)
```

pkg/containers/containers.go
```diff
@@ -90,10 +93,13 @@ func initCgroupV1() {
 		}
 		intID, err := strconv.Atoi(sline[1])
 		if err != nil || intID < 0 {
-			return
+			return fmt.Errorf("error parsing /proc/cgroups")
```
grantseltzer, line 96:
```suggestion
			return fmt.Errorf("error parsing /proc/cgroups: %w", err)
```

pkg/ebpf/events_processor.go
```diff
@@ -328,7 +327,11 @@ func (t *Tracee) processEvent(event *trace.Event) error {
 		if err != nil {
 			return fmt.Errorf("error parsing cgroup_rmdir args: %v", err)
 		}
-		t.containers.CgroupRemove(cgroupId)
+		hId, err := getEventArgUint32Val(event, "hierarchy_id")
+		if err != nil {
+			return fmt.Errorf("error parsing cgroup_mkdir args: %v", err)
```
grantseltzer, line 332:
```suggestion
			return fmt.Errorf("error parsing cgroup_mkdir args: %w", err)
```

pkg/ebpf/tracee.go
```diff
@@ -260,7 +260,10 @@ func New(cfg Config) (*Tracee, error) {
 		}
 	}
 
-	c := containers.InitContainers()
+	c, err := containers.InitContainers()
+	if err != nil {
+		return nil, fmt.Errorf("error initializing containers: %v", err)
```
grantseltzer, line 265:
```suggestion
		return nil, fmt.Errorf("error initializing containers: %w", err)
```

grantseltzer (commented):
Since I can't reproduce this (Yaniv and I spoke offline about how it can be produced in a specific sandbox), can someone who can confirm this fixes the issue approve this? I just had a couple comments about errors.

pkg/containers/containers.go
yanivagman: Yes, you are right. Good catch

OriGlassman:
> Since I can't reproduce this (Yaniv and I spoke offline about how it can be produced in a specific sandbox), can someone who can confirm this fixes the issue approve this? I just had a couple comments about errors.

Based on an environment that this bug was detected on, it no longer occurs and indeed seems to address the issue.

rafaeldtinoco:
Reviewing this now...

rafaeldtinoco (approved):
That is a nice catch! This way, if another cgroup controller directory is being removed we don't remove the container until the cpuset controller directory is removed (and then we still wait sometime on the deletion logic we already had).

PR #1537, opened 2022-03-09
fix compilation, using trace instead of external package

mtcherni95:
this commit fixes Tracee compilation.

FYI @yanivagman


yanivagman (approved):
LGTM

PR #1540, opened 2022-03-10
test(smoke): wait for tracee pid file before sigs are triggered

danielpacak:
To mitigate loosing detections when tracee-ebpf is not
ready when we trigger detection.

### DIFF
--- a/cmd/tracee-rules/output.go
+++ b/cmd/tracee-rules/output.go
@@ -22,8 +22,8 @@ Time: {{ dateInZone "2006-01-02T15:04:05Z" (now) "UTC" }}
 Signature ID: {{ .SigMetadata.ID }}
 Signature: {{ .SigMetadata.Name }}
 Data: {{ .Data }}
-Command: {{ .Context.ProcessName }}
-Hostname: {{ .Context.HostName }}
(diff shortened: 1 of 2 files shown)


yanivagman:
@danielpacak @rafaeldtinoco if that's the case, then this is actually a regression. We used to do the wait in entrypoint.sh, and only start tracee-rules after tracee-ebpf is ready. If that's not the case now, then we need to fix it in entrypoint.sh, and not in the tests themselves

danielpacak:
This PR doesn't make sense if tracee-ebpf readiness check is supposed to be part of the default entrypoint.sh. It used to work like that in v0.6.5 and it seems that we have kind of regression now.

danielpacak:
We spoke with @rafaeldtinoco about logic that we had in entrypoint.sh and it was applicable to scenario when BPF compilation was part of tracee-ebpf. So it's not exactly regression, but it happened to mitigate the problem that we are trying to solve with this PR.

I created #1548 to revisit the purpose of tracee.pid and when we should write it. This PR is just a temporary solution for flaky tests to be less likely by waiting for /tmp/tracee/out/tracee.pid and adding delay of ten seconds.

yanivagman (approved):
LGTM

PR #1541, opened 2022-03-12
Auto-generated Makefile's help message

YuviGold:
Instead of maintaining an help message, it is now created
automatically according to comments through the Makefile file.
Additionally, the help message is now colorized.

- `##@` - marks a new category
- `##` - marks a new entry

The Makefile was reorganized in order for the generation script's output to keep the same order as before.

Before

Screenshot from 2022-03-12 14-52-32

After

Screenshot from 2022-03-12 14-44-39

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,4 @@
-.PHONY: all | env
-all: tracee-ebpf tracee-rules rules
+.DEFAULT_GOAL := all
 
 #
 # make
@@ -104,7 +103,7 @@ LAST_GIT_TAG ?= $(shell $(CMD_GIT) describe --tags --match 'v*' 2>/dev/null)
 VERSION ?= $(if $(RELEASE_TAG),$(RELEASE_TAG),$(LAST_GIT_TAG))
 
 #
-# environment
+##@ environment
 #
 
 UNAME_M := $(shell uname -m)
@@ -123,7 +122,7 @@ ifeq ($(UNAME_M),aarch64)
 endif
 
 .PHONY: env
-env:
+env:  ## show makefile environment/variables
 	@echo ---------------------------------------
 	@echo "Makefile Environment:"
 	@echo ---------------------------------------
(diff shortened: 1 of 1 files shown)


mtcherni95 (commented):
Nice!
LGTM - @rafaeldtinoco what do you think?

rafaeldtinoco:
Not bad, I haven't gone through the changes yet, but I do like the idea of having it automatically generated. @YuviGold do you plan to change all other Makefiles as well ? (under builder/ directory we have some others).

Makefile
```diff
@@ -198,47 +197,22 @@ env:
...
-	@echo "    $$ make test                 	# run all go & opa tests"
-	@echo "    $$ make test-tracee-ebpf     	# go test tracee-ebpf"
-	@echo "    $$ make test-tracee-rules    	# go test tracee-rules"
-	@echo "    $$ make test-rules           	# opa test (tracee-rules)"
-	@echo ""
-	@echo "# flags"
-	@echo ""
-	@echo "    $$ STATIC=1 make ...                 # build static binaries"
-	@echo "    $$ BTFHUB=1 STATIC=1 make ...        # build static binaries, embed BTF"
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-30s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
```
rafaeldtinoco, line 212: $(CMD_AWK) here instead (so user might be able to override de command (gawk, mawk, etc)).

rafaeldtinoco: I find hard to understand/maintain a oneliner here. Is it possible for us to have something structured ? Even if with \ at the end lines (just so its easier to read).

Like:

```awk
BEGIN {
	FS = ":.*##";
	printf "\nUsage:\n  make \033[36m<target>\033[0m\n"
}

/^[a-zA-Z_0-9-]+:.*?##/ {
	printf "  \033[36m%-30s\033[0m %s\n", $1, $2
}

/^##@/ {
	printf "\n\033[1m%s\033[0m\n", substr($0, 5)
}
```

Makefile
```diff
@@ -198,47 +197,22 @@ env:
 # usage
 #
 
+define help_category
+	printf "\n\033[1m%s\033[0m\n" $(1)
```
rafaeldtinoco, line 201: In this case (mixing awk and makefile functions, see other comment) it would be better if we have variables for the colors so we could configure them without having to change the logic (error prone).

rafaeldtinoco (commented):
I like the idea, and I love making nice building logic. Could you check my comments and tell me if you're able to address them ? I'm only willing to +1 this if we also make this to other makefiles (to make things uniform). 

Idea for new makefiles would be to copy your logic lines and structure Makefile with the right comments (so the help is automatically built).

Another idea (for the future) would be to have ENV automatically built as well.

Looking forward to reading your comments. Cheers

YuviGold:
@rafaeldtinoco Hey thanks for your review! All the comments look legitimate and i am going to address them all.

Regarding the alignment of other Makefiles... Which ones did you mean? In any case, I think it can be a part of a separate PR.

Makefile
```diff
@@ -198,47 +197,22 @@ env:
...
-	@echo "    $$ make test-rules           	# opa test (tracee-rules)"
-	@echo ""
-	@echo "# flags"
-	@echo ""
-	@echo "    $$ STATIC=1 make ...                 # build static binaries"
-	@echo "    $$ BTFHUB=1 STATIC=1 make ...        # build static binaries, embed BTF"
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-30s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
+	$(call help_category,flags)
+	$(call help_entry,"STATIC=1 make ..."," build static binaries")
+	$(call help_entry,"BTFHUB=1 STATIC=1 make ..."," build static binaries$(comma) embed BTF")
```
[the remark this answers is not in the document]
YuviGold: Theoretically, that's possible. But in order to have a flexible output as you had (which was my initial intention) it makes it more problematic.
It depends on the type of the variable (e.g. boolean, str) or whether you'd like to support a combination (such as `BTFHUB=1 STATIC=1`) in the output.
For now, I just migrated that code to be only awk on the `END` section

YuviGold:
@rafaeldtinoco I fixed the code according to your requests. Please take a look :)

rafaeldtinoco:
I'll revisit all makefiles in a near future, but not now. I'll reuse this suggestion (with proper credits) when the time comes. I'll probably not use colors, but the auto generated idea is worth.

PR #1545, opened 2022-03-13
fix container pcaps not being closed

### LINKED ISSUE #1544: pcaps are not being closed when separating per-container
when using `--capture pcap:per-container`, the pcap files are not being closed when the container exits.
this is because we use the `cgroup_rmdir` event to identify when the container exits, but we assumed the `containerId` field of the event will hold the container id. this is not the case - `cgroup_rmdir` events occur in the context of the host, hence `containerId` field is empty.

roikol:
fixes #1544.

in `cgroup_rmdir` event, get the container id from the tracee Containers map. use this container id to get the pcap context, and close properly.

this approach forces getting the pcap context using the `container id`, and not the `hostTid` - therefore, created `getContainerPcapContext` func (and also `getProcessPcapContext`, `getHostPcapContext`) and generalized `getPcapContext`.

### DIFF
--- a/pkg/ebpf/events_processor.go
+++ b/pkg/ebpf/events_processor.go
@@ -241,7 +241,7 @@ func (t *Tracee) processEvent(event *trace.Event) error {
 	case SchedProcessExitEventID:
 		if t.config.ProcessInfo {
 			if t.config.Capture.NetPerProcess {
-				pcapContext, _, err := t.getPcapContext(uint32(event.HostThreadID))
+				pcapContext, _, err := t.getPcapContextFromTid(uint32(event.HostThreadID))
 				if err == nil {
 					go t.netExit(pcapContext)
 				}
@@ -317,17 +317,18 @@ func (t *Tracee) processEvent(event *trace.Event) error {
 		}
 
 	case CgroupRmdirEventID:
-		if t.config.Capture.NetPerContainer {
-			pcapContext, _, err := t.getPcapContext(uint32(event.HostThreadID))
-			if err == nil {
-				go t.netExit(pcapContext)
-			}
-		}
-
 		cgroupId, err := getEventArgUint64Val(event, "cgroup_id")
 		if err != nil {
 			return fmt.Errorf("error parsing cgroup_rmdir args: %v", err)
 		}
+
+		if t.config.Capture.NetPerContainer {
+			if info := t.containers.GetCgroupInfo(cgroupId); info.ContainerId != "" {
+				pcapContext := t.getContainerPcapContext(info.ContainerId)
+				go t.netExit(pcapContext)
+			}
+		}
+
 		t.containers.CgroupRemove(cgroupId)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/ebpf/net_events.go
```diff
@@ -169,12 +177,18 @@ func (t *Tracee) getPcapContext(hostTid uint32) (processPcapId, procinfo.Process
 	}
 
 	if t.config.Capture.NetPerProcess {
-		packetContext = processPcapId{hostPid: networkProcess.HostPid, comm: networkProcess.Comm, procStartTime: uint64(networkProcess.StartTime), contID: contID}
+		networkProcess, err := t.getProcessCtx(networkThread.HostPid)
+		if err != nil {
+			return pcapContext, procinfo.ProcessCtx{}, fmt.Errorf("unable to get ProcessCtx of hostTid %d to generate pcap context: %v", networkThread.HostPid, err)
+		}
+		pcapContext = t.getProcessPcapContext(networkProcess.HostPid, networkProcess.Comm, uint64(networkProcess.StartTime), contID)
+		return pcapContext, networkThread, nil
```
yanivagman, line 185: Is this return needed? It's the same as below

pkg/ebpf/net_events.go
```diff
@@ -169,12 +177,18 @@ func (t *Tracee) getPcapContext(hostTid uint32) (processPcapId, procinfo.Process
...
 	} else if t.config.Capture.NetPerContainer {
-		packetContext = processPcapId{contID: contID}
+		pcapContext = t.getContainerPcapContext(contID)
+		return pcapContext, networkThread, nil
```
yanivagman, line 188: Is this return needed? It's the same as below

pkg/ebpf/events_processor.go
```diff
@@ -317,17 +317,18 @@ func (t *Tracee) processEvent(event *trace.Event) error {
 		}
 
 	case CgroupRmdirEventID:
+		cgroupId, err := getEventArgUint64Val(event, "cgroup_id")
+		if err != nil {
+			return fmt.Errorf("error parsing cgroup_rmdir args: %v", err)
+		}
+
 		if t.config.Capture.NetPerContainer {
-			pcapContext, _, err := t.getPcapContext(uint32(event.HostThreadID))
```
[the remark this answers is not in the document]
roikol: the problem was that the `cgroup_rmdir` event doesn't happen in the context of the container (`containerId == ""`).
this means that the process (`event.HostThreadID`) triggering this event also have `containerId == ""`. trying to get the pcap context of this `HostThreadID` would look for the `containerId` field, realize that it's empty, and assume that the pcap context is of the host (TRACEE_OUT/out/host/capture.pcap). therefore it will close the pcap of the host, instead of the pcap of the container. 

the fix: use the `cgroup_id` argument of `cgroup_rmdir` event, to get the `container id` which is being closed. get the pcap context using this `container id`, and close the correct pcap file.

yanivagman (approved):

PR #1547, opened 2022-03-14
templates: change rawjson.tmpl to be backwards compatible

NDStrahilevitz:
Some users have reported a breaking change when we changed `Finding.Context` to `Finding.Event.Payload`.

This PR updates the `rawjson.tmpl` to be backwards compatible with to the previous struct(from 0.6.5). Later json templates need to be versioned appropriately (A discussion on this will be opened in the near future).

### DIFF
--- a/cmd/tracee-rules/output_test.go
+++ b/cmd/tracee-rules/output_test.go
@@ -250,3 +250,70 @@ HostName: foobar.local
 
 	}
 }
+
+func TestJsonTemplateV1(t *testing.T) {
+	testCases := []struct {
+		testName     string
+		finding      detect.Finding
+		expectedJson string
+	}{
+		{
+			testName: "convert finding to v1",
+			finding: detect.Finding{
+				Data: map[string]interface{}{
+					"a": 123,
+					"b": "c",
+					"d": true,
+					"f": map[string]string{
+						"123": "456",
+						"foo": "bar",
+					},
+				},
+				Event: trace.Event{
+					ProcessID:   21312,
+					Timestamp:   1321321,
+					UserID:      0,
+					ContainerID: "abbc123",
+					EventName:   "execve",
+				}.ToProtocol(),
+				SigMetadata: detect.SignatureMetadata{
+					ID:          "TRC-1",
+					Version:     "0.1.0",
+					Name:        "Standard Input/Output Over Socket",
+					Description: "Redirection of process's standard input/output to socket",
+					Tags:        []string{"linux", "container"},
+					Properties: map[string]interface{}{
+						"Severity":     3,
+						"MITRE ATT&CK": "Persistence: Server Software Component",
+					},
+				},
+			},
+			expectedJson: `{
+				"Data": {
+					"a":123,"b":"c","d":true,"f":{"123":"456","foo":"bar"}
+				},
+				"Context":{
+					"timestamp":1321321,"processorId":0,"processId":21312,"threadId":0,"parentProcessId":0,"hostProcessId":0,"hostThreadId":0,"hostParentProcessId":0,"userId":0,"mountNamespace":0,"pidNamespace":0,"processName":"","hostName":"","containerId":"abbc123","eventId":"0","eventName":"execve","argsNum":0,"returnValue":0,"stackAddresses":null,"args":null
+				},
+				"SigMetadata":{
+					"ID":"TRC-1","Version":"0.1.0","Name":"Standard Input/Output Over Socket","Description":"Redirection of process's standard input/output to socket","Tags":["linux","container"],"Properties":{"MITRE ATT\u0026CK":"Persistence: Server Software Component","Severity":3}
+				}
+			}`,
+		},
+	}
+
+	jsonTemplate, err := setupTemplate("templates/rawjson.tmpl")
+
+	require.NoError(t, err)
+
+	for _, tc := range testCases {
+		t.Run(tc.testName, func(t *testing.T) {
+			//pass finding through template and compare to a v1 json
+			var buf bytes.Buffer
+			jsonTemplate.Execute(&buf, tc.finding)
+
+			assert.JSONEq(t, tc.expectedJson, buf.String())
+		})
+	}
+
+}
--- a/cmd/tracee-rules/templates/rawjson.tmpl
+++ b/cmd/tracee-rules/templates/rawjson.tmpl
@@ -1,1 +1,1 @@
-{{ toJson . }}
+{"Data":{{ toJson .Data }},"Context":{{ toJson .Event.Payload }},"SigMetadata":{{ toJson .SigMetadata }}}


mtcherni95 (commented):
LGTM
@AsafEitani I suggest you to check it with your apps :)

cmd/tracee-rules/templates/0.6.5json.tmpl
```diff
@@ -0,0 +1 @@
+{"Data": {{ toJson .Data }}\n"Context": {{ toJson .Event.Payload }}\n"SigMetadata": {{ toJson .SigMetadata }}\n}
```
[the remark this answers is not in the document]
NDStrahilevitz: My template was missing commas so it wasn't even a valid json 😅, thanks for noticing. I've pushed an updated template, have a look.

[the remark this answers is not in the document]
NDStrahilevitz: Updated to what was in your initial comment.

danielpacak:
Just for my understanding. Is this PR related to the [rawjson.tmpl](https://github.com/aquasecurity/tracee/blob/v0.6.5/tracee-rules/templates/rawjson.tmpl) that will produce different output in the upcoming v0.7.0?

If yes, I'm wondering why don't we patch the rawjson.tmpl directly? Git revision will differentiate v0.6.5 and v0.7.0 versions so I'm not sure we need to append version numbers to template filenames:

https://github.com/aquasecurity/tracee/blob/v0.6.5/tracee-rules/templates/rawjson.tmpl
https://github.com/aquasecurity/tracee/blob/v0.7.0/tracee-rules/templates/rawjson.tmpl

NDStrahilevitz:
> Just for my understanding. Is this PR related to the [rawjson.tmpl](https://github.com/aquasecurity/tracee/blob/v0.6.5/tracee-rules/templates/rawjson.tmpl) that will produce different output in the upcoming v0.7.0?
> 
> If yes, I'm wondering why don't we patch the rawjson.tmpl directly? Git revision will differentiate between v0.6.5 and v0.7.0 so I'm not sure we need to append version numbers to template filenames:
> 
> https://github.com/aquasecurity/tracee/blob/v0.6.5/tracee-rules/templates/rawjson.tmpl https://github.com/aquasecurity/tracee/blob/v0.7.0/tracee-rules/templates/rawjson.tmpl

The usecase for this is for users of 0.7.0 that want their findings outputted in the format of 0.6.5 (their tools depend on the previous format). Since rawjson does a straight conversion to json, using that template from 0.6.5 on a 0.7.0 finding won't convert to the needed format.

danielpacak:
> > Just for my understanding. Is this PR related to the [rawjson.tmpl](https://github.com/aquasecurity/tracee/blob/v0.6.5/tracee-rules/templates/rawjson.tmpl) that will produce different output in the upcoming v0.7.0?
> > If yes, I'm wondering why don't we patch the rawjson.tmpl directly? Git revision will differentiate between v0.6.5 and v0.7.0 so I'm not sure we need to append version numbers to template filenames:
> > https://github.com/aquasecurity/tracee/blob/v0.6.5/tracee-rules/templates/rawjson.tmpl https://github.com/aquasecurity/tracee/blob/v0.7.0/tracee-rules/templates/rawjson.tmpl
> 
> The usecase for this is for users of 0.7.0 that want their findings outputted in the format of 0.6.5 (their tools depend on the previous format). Since rawjson does a straight conversion to json, using that template from 0.6.5 on a 0.7.0 finding won't convert to the needed format.
> 
> If someone wants to just output the current `Finding` struct to a json as is, then `rawjson.tmpl` will suffice, but if a previous version of the json is needed this new template will do that conversion.

Got it. If we want to maintain output templates for backward compatibility do you think it makes sense to unit test them? I saw in your previous comments that every single character matters ;)

NDStrahilevitz:
As discussed with @danielpacak, i've updated this PR to update `rawjson.tmpl` itself to be backwards compatible (by using @AsafEitani's tmpl code), and added a unit test to catch regressions in the future.

cmd/tracee-rules/output_test.go
```diff
@@ -250,3 +250,58 @@ HostName: foobar.local
...
+
+	if assert.NoError(t, err) {
+		for _, tc := range testCases {
+			//test if the raw json of findingv065 is equivalent to the 065 template result
+			var buf bytes.Buffer
+			jsonTemplate.Execute(&buf, tc.finding)
+
+			assert.JSONEq(t, tc.expectedJson, buf.String())
+		}
+	}
```
danielpacak, line 305: Usually we'd use require.NoError() to stop executing tests in case of critical failures. Also, for table driven tests using subtests with t.Run() will show you which one has actually failed in case of assertion errors:

```suggestion
	jsonTemplate, err := setupTemplate("templates/rawjson.tmpl")
	require.NoError(t, err)

	for _, tc := range testCases {
		t.Run(tc.testName, func(t *testing.T) {
			var buf bytes.Buffer
			err := jsonTemplate.Execute(&buf, tc.finding)
			require.NoError(t, err)
			assert.JSONEq(t, tc.expectedJson, buf.String())
		})
	}
```

cmd/tracee-rules/output_test.go
```diff
@@ -250,3 +250,58 @@ HostName: foobar.local
...
+					Name:        "Standard Input/Output Over Socket",
+					Description: "Redirection of process's standard input/output to socket",
+					Tags:        []string{"linux", "container"},
+					Properties: map[string]interface{}{
+						"Severity":     3,
+						"MITRE ATT&CK": "Persistence: Server Software Component",
+					},
+				},
+			},
+			expectedJson: "{\"Data\":{\"a\":123,\"b\":\"c\",\"d\":true,\"f\":{\"123\":\"456\",\"foo\":\"bar\"}},\"Context\":{\"timestamp\":1321321,\"processorId\":0,\"processId\":21312,\"threadId\":0,\"parentProcessId\":0,\"hostProcessId\":0,\"hostThreadId\":0,\"hostParentProcessId\":0,\"userId\":0,\"mountNamespace\":0,\"pidNamespace\":0,\"processName\":\"\",\"hostName\":\"\",\"containerId\":\"abbc123\",\"eventId\":\"0\",\"eventName\":\"execve\",\"argsNum\":0,\"returnValue\":0,\"stackAddresses\":null,\"args\":null},\"SigMetadata\":{\"ID\":\"TRC-1\",\"Version\":\"0.1.0\",\"Name\":\"Standard Input/Output Over Socket\",\"Description\":\"Redirection of process's standard input/output to socket\",\"Tags\":[\"linux\",\"container\"],\"Properties\":{\"MITRE ATT&CK\":\"Persistence: Server Software Component\",\"Severity\":3}}}",
```
danielpacak, line 291: I believe this can be more readable with Go multiline strings, wdyt?

```suggestion
			expectedJson: `{
	"Data": {
		"a": 123
	}
}`,
```

danielpacak (changes requested):

cmd/tracee-rules/output_test.go
NDStrahilevitz: Right on both, i'll add it.

cmd/tracee-rules/output_test.go
NDStrahilevitz: Done

danielpacak (approved):

itaysk:
came here to ask about tests and and saw this was already discussed. I agree it should have been fixed with this PR. @NDStrahilevitz are you working on it?

NDStrahilevitz:
> came here to ask about tests and and saw this was already discussed. I agree it should have been fixed with this PR. @NDStrahilevitz are you working on it?

Do you mean tests for the rest of the templates, or for the one modified in this PR? If the former, i'm not currently working on it (but I can open an issue on it), if the latter then a test was added as part of this PR.

PR #1553, opened 2022-03-14
pkg/ebpf: add hook for magic_write event upon __kernel_write function

### LINKED ISSUE #1550: Missing hook in __kernel_write for magic_write
In the effort to expand the `magic_write` event to catch all file operations, the `__kernel_write` function is targeted as one of the functions that envelope the use of the file operation `write_iter` for kernel uses.

This is sub-issue of the #1357 issue.

AlonZivony:
fix #1550


pkg/ebpf/tracee.go
```diff
@@ -619,6 +620,7 @@ func (t *Tracee) populateBPFMaps() error {
 	errs = append(errs, t.initTailCall(tailVfsWritev, "prog_array", "trace_ret_vfs_writev_tail"))
 	errs = append(errs, t.initTailCall(tailSendBin, "prog_array", "send_bin"))
 	errs = append(errs, t.initTailCall(tailSendBinTP, "prog_array_tp", "send_bin_tp"))
+	errs = append(errs, t.initTailCall(tailKernelWrite, "prog_array", "trace_ret_vfs_write_tail"))
```
roikol, line 623: if i understand correctly, this should be `trace_kernel_write_tail`

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3960,6 +3964,21 @@ int BPF_KPROBE(trace_ret_vfs_writev_tail)
     return do_vfs_write_writev_tail(ctx, VFS_WRITEV);
 }
 
+SEC("kprobe/__kernel_write")
+TRACE_ENT_FUNC(kernel_write, KERNEL_WRITE);
+
+SEC("kretprobe/__kernel_write")
+int BPF_KPROBE(trace_ret_kernel_write)
+{
+    return do_vfs_write_writev(ctx, KERNEL_WRITE, TAIL_KERNEL_WRITE);
```
roikol, line 3973: the only difference i see between the `__kernel_write` function and the `vfs_write` one, is that the `buf` argument is a user pointer in vfs_write. this shouldn't cause a problem because of the use of `bpf_probe_read` in `do_vfs_write_writev` func.

i guess maybe just renaming `do_vfs_write_writev` to something more general would make it clearer.

pkg/ebpf/events_definitions.go
```diff
@@ -5850,7 +5850,11 @@ var EventsDefinitions = map[int32]EventDefinition{
 	MagicWriteEventID: {
 		ID32Bit: sys32undefined,
 		Name:    "magic_write",
-		Probes:  []probe{}, Sets: []string{},
+		Probes: []probe{
```
roikol, line 5853: you've added `magic_write` event to be able to capture by itself. therefore, you should set it as essential if capture write was chosen by the user (https://github.com/aquasecurity/tracee/blob/main/pkg/ebpf/tracee.go#L203).
also keep track of the capture here https://github.com/aquasecurity/tracee/blob/main/pkg/ebpf/events_processor.go#L107

pkg/ebpf/tracee.go
AlonZivony: I must have mistyped it xD

roikol:
@AlonZivony 

looking good now. i like the approach of adding the new event, it goes well with how magic_write works today :)

roikol:
@AlonZivony 

looking good now. i like the approach of adding the new event, it goes well with how magic_write works today :)

rafaeldtinoco:
Reviewing this one now.

pkg/ebpf/events_definitions.go
rafaeldtinoco: nice catches @roikol

rafaeldtinoco (approved):
LGTM

rafaeldtinoco:
This merge had an issue and is being fixed by: https://github.com/aquasecurity/tracee/pull/1632

rafaeldtinoco:
When reviewing this I missed:

https://github.com/aquasecurity/tracee/issues/1721

PR #1555, opened 2022-03-15
limit number of open pcap using lru.Cache

roikol:
using lru.Cache as the map of the opened pcapWriters. a callback is called whenever a key is removed from this cache (due to LRU, or specifically remove due to process/container exit), enabling us to close the FD.

i set the limit to be 512 - just an arbitrary number that seems to me as good enough to let tracee use 1024-512 FDs, and also not having the overhead of opening an already existing pcap too many times.

if there are any other things to consider here - please let me know!

### DIFF
--- a/pkg/ebpf/net_events.go
+++ b/pkg/ebpf/net_events.go
@@ -3,6 +3,7 @@ package ebpf
 import (
 	gocontext "context"
 	"fmt"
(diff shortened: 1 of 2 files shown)


rafaeldtinoco (changes requested):
LGTM, would you mind rebasing to origin/main and resolving conflicts so I can merge it ?

Issue #1522, opened 2022-02-28
Re-create Makefile test targets


rafaeldtinoco:
After https://github.com/aquasecurity/tracee/pull/1517 we have realized that the test makefile targets lost their initial organization with the project re-structure. Now, `test-tracee-ebpf`, `test-tracee-rules` and `test-rules` don't seem to be reflecting what we need.

Comments about it: https://github.com/aquasecurity/tracee/pull/1517#discussion_r815849662

I spoke to @danielpacak about this and he agreed he could help with this reorganization of tests. @NDStrahilevitz FYI

PR #1559, opened 2022-03-16
chore(make): re-create test targets

danielpacak:
Instead of grouping tests by tracee-ebpf and tracee-rules
we will group them by scope, i.e. unit, integration, and
end-to-end tests.

```
               .
              / \\
             /   \ \
            / E2E \' \
           / TESTS \'' \
          /---------\ ' '\
         /           \''   \
        / INTEGRATION \ ' '  \
       /     TESTS     \   ''  \
      /-----------------\' ' ' ' \
     /                   \' ' '  ' \
    /                     \' '   ' ' \
   /      UNIT TESTS       \  '   ' /
  /                         \   ' /
 /                           \'  /
/_____________________________\/
```

NB There are still some tests that do not fall into category
of unit, integration, nor end-to-end tests, but fixing them
is out of scope of this patch.

Resolves: #1522


danielpacak:
Good read on practical testing pyramid on https://martinfowler.com/articles/practical-test-pyramid.html

tests/integration/integration_test.go
```diff
@@ -212,8 +212,11 @@ func getAllSyscallsInSet(set string) []string {
 	return syscallsInSet
 }
 
+// TODO This test is nice, but very hard to run in CI pipeline because it requires
+//      root privileges. I'll keep it for now so we can run it manually with sudo.
 func Test_Events(t *testing.T) {
-	t.Skip("This is an integration test")
+	t.Skip("This test requires root privileges")
+
```
grantseltzer, line 219: How can this be manually run?

danielpacak: That's a very good question to the author of this test ;) I've added GoDoc with instructions on how to run in manually based on reverse engineering and taking into account new Make targets.

danielpacak:
> I think this looks great, all for the change. What do you think about:
> 
> * Running all tests at the begining of the release github actions workflow
> * Adding a make target to run all tests

We run all tests as part PR validation workflow, which at the end does what we called release snapshot. IMO, if that is okey and did not catch any failures we can create an annotated Git tag and start release workflow without running all these tests again.

Regarding make target that runs all kinds of tests I'm not sure about it. My biggest concern is that if some make targets are run in parallel these tests may become flaky. Currently, for the PR validation workflow, we run them sequentially and that's what I wanted to keep in scope of #1522. But definitely, that's something we can refine in subsequent PRs.

grantseltzer (approved):

Issue #1560, opened 2022-03-17
Catch file operations hooks of the /proc


itamarmaouda101:
Hi 
we should add detection for hooking `/proc` file_operations function handlers.

sometimes malware hooks those functions to hide processes

PR #1561, opened 2022-03-17
hooks detections:  file operations hooking under /proc

itamarmaouda101:
Hi 
This PR solves #1560 .

It is adding two new events.
1. `fetch_proc_fops` that print the user the addresses of the members of /proc file_opertions
2.  `detect_hooked_proc_fops` user-driven event that checks the memory region of the function address and alerts if it is out the range

Note:
The events now are just checking the addresses of the struct itself and the `iterate_shared` function because AFAIK this the function and this struct are commonly hooked by malware

### DIFF
--- a/pkg/bufferdecoder/eventsreader.go
+++ b/pkg/bufferdecoder/eventsreader.go
@@ -211,7 +211,7 @@ func GetParamType(paramType string) ArgType {
 		return credT
 	case "umode_t":
 		return u16T
-	case "unsigned long[]":
+	case "unsigned long[]", "HookedFunctionData[]":
 		return uint64ArrT
 	default:
 		// Default to pointer (printed as hex) for unsupported types
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -202,7 +202,8 @@ Copyright (C) Aqua Security inc.
 #define DIRTY_PIPE_SPLICE               1038
 #define DEBUGFS_CREATE_FILE             1039
 #define PRINT_SYSCALL_TABLE             1040
-#define MAX_EVENT_ID                    1041
+#define FETCH_PROC_FOPS                 1035
+#define MAX_EVENT_ID                    1036
 
 #define NET_PACKET                      4000
 
@@ -1215,6 +1216,11 @@ static __always_inline struct pipe_buffer *get_last_write_pipe_buffer(struct pip
 
     struct pipe_buffer *current_buffer = get_node_addr(bufs, curbuf);
     return current_buffer;
+}
+
+static __always_inline struct inode * get_inode_from_file(struct file *file)
+{
+    return READ_KERN(file->f_inode);
 }
 
 /*============================ HELPER FUNCTIONS ==============================*/
@@ -2268,6 +2274,32 @@ static __always_inline struct pipe_inode_info *get_file_pipe_info(struct file *f
         return NULL;
     }
     return pipe;
+}
+
+static __always_inline void send_file_operations_struct(event_data_t *data, struct file * called_file)
+{
+    char* top = "top";
+    char* ps = "ps";
+    //check's if the called process is top or ps (to list processes)
+    if ((!has_prefix(data->context.comm, "top ", 4)) && (!has_prefix(data->context.comm, "ps ", 3)))
+        return ;
+    struct inode *file_inode = get_inode_from_file(called_file);
+    struct file_operations *fops = (struct file_operations *)READ_KERN(file_inode->i_fop);
+    if (fops == NULL)
+        return ;
+
+    unsigned long iterate_shared_addr = (unsigned long) READ_KERN(fops->iterate_shared);
+    if (iterate_shared_addr == 0)
+        return ;
+
+    unsigned long fops_addresses[2] ={
+        iterate_shared_addr,
+        (unsigned long) fops
+    };
+
+    save_u64_arr_to_buf(data, (const u64 *)fops_addresses, 2, 0);
+    events_perf_submit(data, FETCH_PROC_FOPS, 0);
+
 }
 
 /*============================== SYSCALL HOOKS ===============================*/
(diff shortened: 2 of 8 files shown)


yanivagman (changes requested):
Why do we need to events and not one?
I think that a single DetectHookedProcFopsEventID event would be enough.
This also means there is no need to define a new "derived event" here.

yanivagman:
> > Why do we need two events and not one? I think that a single DetectHookedProcFopsEventID event would be enough. This also means there is no need to define a new "derived event" here.
> 
> @yanivagman We can merge it into one event, but in that case, we need to mess a bit with the events Definitions table, because the `DetectHookedProcFopsEventID` event only alerts on hooked fops so its argument needs to be only the `struct HookedFunctionData` and if will merge them we will need to make its argument include the list of fop arguments.
> 
> That can mislead the user, what do you think?

I don't see a reason to merge the list of arguments. Just use the argument of DetectHookedProcFopsEventID in the event definition and remove the other event definition. Am I missing something?

pkg/bufferdecoder/eventsreader.go
```diff
@@ -211,7 +211,7 @@ func GetParamType(paramType string) ArgType {
 		return credT
 	case "umode_t":
 		return u16T
-	case "unsigned long[]":
+	case "unsigned long[]", "HookedFunctionData[]":
```
rafaeldtinoco, line 214: Why can't you simply stay with "unsigned long[]" for your case ? I dont think we need to create another type here (as we're sticking with basic types unless we need to differentiate them somewhere else, not your case).

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4819,6 +4825,59 @@ int BPF_KPROBE(trace_ret_do_splice)
...
+    if (!init_event_data(&data, ctx))
+       return 0;
+
+    if (!should_trace(&data.context))
+        return 0;
+    struct file *f = (struct file *)PT_REGS_PARM1(ctx);
+    if (f == NULL)
+        return 0;
+
+    get_path_str(GET_FIELD_ADDR(f->f_path));
```
rafaeldtinoco, line 4863: @yanivagman, Itamar told me he couldn't use:

```C
buf_t *string_p = get_path_str(GET_FIELD_ADDR(f->f_path));
```

not sure the reason. Is there any specific reason he had to use get_buf() and get_buf_off() ?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4819,6 +4825,59 @@ int BPF_KPROBE(trace_ret_do_splice)
...
+    // Get per-cpu string buffer
+    buf_t *string_p = get_buf(STRING_BUF_IDX);
+    if (string_p == NULL)
+        return -1;
+    u32 *off = get_buf_off(STRING_BUF_IDX);
+    if (off == NULL)
+        return 0;
+
+    if (*off > MAX_PERCPU_BUFSIZE - MAX_STRING_SIZE)
```
rafaeldtinoco, line 4872: I believe `get_path_str()` takes care of checking if d_name of dentry fits the buffer already:

```c
// Is string buffer big enough for dentry name?
sz = 0;
if (off <= buf_off) { // verify no wrap occurred
    len = len & ((MAX_PERCPU_BUFSIZE >> 1)-1);
    sz = bpf_probe_read_str(&(string_p->buf[off & ((MAX_PERCPU_BUFSIZE >> 1)-1)]), len, (void *)d_name.name);
}
else
    break;
```

no ?

pkg/ebpf/events_processor.go
```diff
@@ -340,8 +345,42 @@ func (t *Tracee) processEvent(event *trace.Event) error {
...
+			}
+			if !inTextSeg {
+				hookingFunction, err := t.kernelSymbols.GetSymbolByAddr(addr)
+				if hookingFunction.Owner == "system" && err == nil {
+					continue
+				}
+				if err != nil {
+					hookingFunction.Owner = "hidden"
+				} else {
+					hookingFunction.Owner = strings.TrimPrefix(hookingFunction.Owner, "[")
```
rafaeldtinoco, line 367: it feels to me that [ and ] should be removed in GetSymbolByAddr function, no ?

rafaeldtinoco (changes requested):
@yanivagman should take a look at my comments as well. Thanks @itamarmaouda101 !!

pkg/ebpf/events_processor.go
```diff
@@ -340,8 +345,42 @@ func (t *Tracee) processEvent(event *trace.Event) error {
...
+		}
+		hookedFops := make([]bufferdecoder.HookedSymbolData, 0, 0)
+		for idx, addr := range fopsAddresses {
+			inTextSeg, err := t.kernelSymbols.TextSegmentContains(addr)
+			if err != nil {
+				return fmt.Errorf("error checking kernel address: %v", err)
+			}
+			if !inTextSeg {
+				hookingFunction, err := t.kernelSymbols.GetSymbolByAddr(addr)
+				if hookingFunction.Owner == "system" && err == nil {
```
AlonZivony, line 361: Just saw it - this is a bug.
Change to:
```suggestion
				if err == nil && hookingFunction.Owner == "system" {
```

AlonZivony, line 361: BTW,
I think that this if-if-else form is not a best practice.
if recommend you to change it to be:
```suggestion
if err == nil {
    if hookingFunction.Owner == "system" {
    } else {}
} else {}
```

pkg/bufferdecoder/eventsreader.go
itamarmaouda101: I agree, but I think we have no other choice because in the event defections map this event argument is `HookedFunctionData[]`  which is because after the process of the event it casts its argument to be HookedFunctionData.

That is done because we submit to bpf array of addresses that each one is a function address that we want to check, and only after the check (if it hooked or not), I assign it to the HookedFunctionData struct with the name of the symbol of the address and the module owner

pkg/ebpf/events_processor.go
itamarmaouda101: yep , ill open PR on that in libbpfgo to change it

itamarmaouda101: Hi i, updating that ill tried to create a fix for that in libbpfgo, for some reason i couldn't run my libpfgo commit with the PR, I'd love to do it in the following days but I'm supposed to be offline for the couple weeks so i open the PR as draft there but ill be available for any Q in the git (:

pkg/ebpf/c/tracee.bpf.c
yanivagman: I need to see the verifier error first to try to answer that

AsafEitani:
#1718 continues and replaces this PR.
This can be closed.

Issue #1563, opened 2022-03-20
Missing container detections with `containerd` runtime


NDStrahilevitz:
There are two cases with the containerd runtime where tracee will currently not detect new containers and relevant events as happening in a container:
1. In some k8s configurations, containerd pods will be created under cpuset as `/.../cpuset/...:cri-containerd:{containerId}`. The current containerId parsing function doesn't apply to such paths, and so events happening under these cgroups will not be enriched with the relevant containerId.
2. In non k8s environments, where containerd containers can be created through the `ctr` tool, containers will be created under `/...cpuset/{namespace}/{containerId}`. The current parsing function doesn't suffice here as well but it is harder to allow for such containers because:
a. These containers can be created with an arbitrary containerId (`ctr` receives the id as a parameter)
b. We have no support for namespaces at this moment

The first case is easily fixable and I hope to upload a PR for it soon. 
The second case is more relevant for personal usage, and is harder to implement, since we need to determine a way to track namespaces (this might also be relevant to other container runtimes where namespacing is used).

NDStrahilevitz:
Case no. 1 was solved in #1564

PR #1564, opened 2022-03-20
containers: add missing detection for containerd

NDStrahilevitz:
This PR fixes the first case described in #1563

I've confirmed this works in a relevant local environment by tracing `cgroup_mkdir` and `container_create` events.

### DIFF
--- a/pkg/containers/containers.go
+++ b/pkg/containers/containers.go
@@ -225,6 +225,9 @@ func getContainerIdFromCgroup(cgroupPath string) (string, string) {
 		case strings.HasPrefix(id, "cri-containerd-"):
 			runtime = "containerd"
 			id = strings.TrimPrefix(id, "cri-containerd-")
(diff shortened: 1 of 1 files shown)


NDStrahilevitz:
Would adding tests for this function be in the scope of this PR? If so i'd like to add that too.

pkg/containers/containers.go
```diff
@@ -225,6 +225,9 @@ func getContainerIdFromCgroup(cgroupPath string) (string, string) {
 		case strings.HasPrefix(id, "cri-containerd-"):
 			runtime = "containerd"
 			id = strings.TrimPrefix(id, "cri-containerd-")
+		case strings.Contains(pc, ":cri-containerd:"):
+			runtime = "containerd"
```
mtcherni95, line 229: I think we should start having runtime names under const vars for multiple usage, as it is happening here with `"containerd"`.

yanivagman (approved):

Issue #1565, opened 2022-03-22
tracing call_usermode_helper function


itamarmaouda101:
Hi 

We should add an event that traces calls of `call_usermode_helper()`.
This function is used commonly by malware to interact with the user-space from the kernel.

Currently, I'm working to add kprobe event to trace it (: 

BTW this is the manual of the function:
[https://www.kernel.org/doc/htmldocs/kernel-api/API-call-usermodehelper.html](https://www.kernel.org/doc/htmldocs/kernel-api/API-call-usermodehelper.html
)

Also, here is an example of kernel module that using this function to escape a container: 
[https://xcellerator.github.io/posts/docker_escape/](https://xcellerator.github.io/posts/docker_escape/)

PR #1566, opened 2022-03-22
Adding call_usermode_helper event

itamarmaouda101:
Hello 

This PR solves #1565 by adding kprobe to `call_usermode_helper()`.

`call_usermode_helper` is a function that is commonly used by kernel rootkits to interact with the user space from the kernel.

Here is the function man page:
[https://www.kernel.org/doc/htmldocs/kernel-api/API-call-usermodehelper.html](https://www.kernel.org/doc/htmldocs/kernel-api/API-call-usermodehelper.html
)

And This is an example of using this function to get container escape:
[https://xcellerator.github.io/posts/docker_escape/](https://xcellerator.github.io/posts/docker_escape/ )

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -197,7 +197,8 @@ Copyright (C) Aqua Security inc.
 #define __KERNEL_WRITE                  1034
 #define PROC_CREATE                     1035
 #define KPROBE_ATTACH                   1036
-#define MAX_EVENT_ID                    1037
+#define CALL_USERMODE_HELPER            1037
+#define MAX_EVENT_ID                    1038
 
 #define NET_PACKET                      4000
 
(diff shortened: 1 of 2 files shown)


pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2717,6 +2718,29 @@ int BPF_KPROBE(trace_filldir64)
...
+
+SEC("kprobe/call_usermodehelper")
+int BPF_KPROBE(trace_call_usermodehelper)
+{
+    event_data_t data = {};
+    if (!init_event_data(&data, ctx))
+        return 0;
+
+    if (!should_trace(&data.context))
+        return 0;
```
rafaeldtinoco, line 2730: enter here

rafaeldtinoco (changes requested):
let me know once you rebase asking for a new review. thanks!

pkg/ebpf/events_definitions.go
```diff
@@ -6248,4 +6249,18 @@ var EventsDefinitions = map[int32]EventDefinition{
...
+		ID32Bit: sys32undefined,
+		Name:    "call_usermodehelper",
+		Probes: []probe{
+			{event: "call_usermodehelper", attach: kprobe, fn: "trace_call_usermodehelper"},
+		},
+		Sets: []string{},
+		Params: []trace.ArgMeta{
+			{Type: "const char*", Name: "pathname"},
+			{Type: "const char*const*", Name: "argv"},
+			{Type: "const char*const*", Name: "envp"},
```
rafaeldtinoco, line 6262: Differently than execve probe, in here we always want to get the env variables I suppose. Mostly because it is not a question of tracing executed commands, but a specific security probe. Spoke to @itamarmaouda101 who confirmed it.

rafaeldtinoco (approved):
LGTM

PR #1569, opened 2022-03-22
Makefile: add "-fPIC" to libbpf compilation flags

NDStrahilevitz:
@rafaeldtinoco 

This is in preparation for container enrichment, the issue arised when importing `containerd`'s runtime SDK, compilation would fail due to relocations, as I understand it most go modules are compiled with position independence by default but clang doesn't do so, and so conflicts sometimes happens with libbpf.

This flag should solve those issues.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -285,7 +285,7 @@ $(OUTPUT_DIR)/tracee.bpf: \
 # libbpf
 #
 
-LIBBPF_CFLAGS =
+LIBBPF_CFLAGS = "-fPIC"
 LIBBPF_LDLAGS =
 LIBBPF_SRC = ./3rdparty/libbpf/src
 


rafaeldtinoco:
Okay we just spoke let's see what this brakes then we can merge it.

mtcherni95:
this won't be part of v0.7.0 version. right?

rafaeldtinoco:
BTW, if we only needed our static libs into our final binary it would be ok not to have fPIC but, since we are mixing (static linking for libbpf and possible dynamic linking for OS libs) then we definitely need fPIC added to the static lib. So this change is correct.

rafaeldtinoco (approved):
LGTM per my last comment.

PR #1571, opened 2022-03-22
Remove interactive tty flags from docker run when building, RELEASING.md

grantseltzer:
The new release workflow has a flaw where the tar file asset does not contain the tracee-ebpf and tracee-rules binaries. This appears to be because the actual build is done in a docker container but that container is invoked with the `-it` flag, causing the build to fail. This PR removes those flags.

See failing command here: 
https://github.com/aquasecurity/tracee/runs/5631481033?check_suite_focus=true#step:4:1277

This also adds a RELEASING markdown file consistent with other Aqua projects.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -17,6 +17,9 @@
 1. Prepare release by creating the PR with the following changes
    1. Update the libbpfgo module
    1. Update the types module
+   1. Update the container image tag in the following files:
+      1. `deploy/kubernetes/tracee-falcosidekick/falcosidekick.yaml`
+      1. `deploy/kubernetes/tracee-postee/tracee.yaml`
 1. Run tests and checks
    1. Check that there are no verifier issues when choosing all events in tracee-ebpf (using `--trace e=*`)
    1. Check both CO-RE and non CO-RE builds
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,0 +1,46 @@
+# Releasing
+
+1. Checkout your fork and make sure it's up-to-date with the `upstream`
+   ```console
+   $ git remote -v
+   origin     git@github.com:<your account>/tracee.git (fetch)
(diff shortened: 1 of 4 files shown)


rafaeldtinoco (approved):
LGTM

RELEASING.md
```diff
@@ -0,0 +1,46 @@
...
+   upstream   git@github.com:aquasecurity/tracee.git (fetch)
+   upstream   git@github.com:aquasecurity/tracee.git (push)
+   ```
+   ```
+   git pull -r
+   git fetch upstream
+   git merge upstream/main
+   git push
+   ```
+1. Prepare release by creating the PR with the following changes
```
danielpacak, line 17: I'm wondering if we want to keep using the latest tag in K8s deployment descriptors here: https://github.com/aquasecurity/tracee/blob/v0.7.0-rc-1/deploy/kubernetes/tracee-postee/tracee.yaml#L20

If there is a non backward compatible change in tracee image and we don't not update YAMLs it will fail in a way that's hard to troubleshoot without going through changes log.

danielpacak (approved):
LGTM! Just left question about K8s deployment descriptors and latest image tag.

RELEASING.md
danielpacak: Using latest tag works fine with the latest release of tracee. However, if we we introduce new environment variable that must be set in K8s container image or extra volume mount, previous versions will stop working. That's why I'm usually pinning versions in K8s deployment descriptors. This also helps troubleshooting bugs when you know which exact version was running in a cluster. Last but not least, using the latest tag is not considered as good practice and many tools (including our own products) will warn about that.

grantseltzer: I added a note about this.

danielpacak (approved):

PR #1572, opened 2022-03-22
Enrich container events

NDStrahilevitz:
# Changes

Adds container data enrichment from runtime APIs to events.

Currently the data added is:
1. Container name (if exists)
2. Container's image
3. Pod metadata (if in k8s)

Supported runtimes are:
1. Containerd
2. Docker
4. CRI-O

Podman will be done in a later PR, since it requires implementing our own API client (the provided library isn't very good IMO) but integration should be simple through the interfaces introduced here.

Architecture diagram (thanks @mtcherni95):

image


simar7:
hi @NDStrahilevitz - to keep you in the loop: We had an impromptu discussion about including Pod information in the events Tracee emits for consumers. One such consumer could be Postee - taking in events from Tracee and then being able to "act" upon them.

A use case would be to label a pod that was reported in a Tracee event. For this we could use `kubectl` along with a label selector that matches the relevant pod with its name, as reported by the Tracee event.

NDStrahilevitz:
> hi @NDStrahilevitz - to keep you in the loop: We had an impromptu discussion about including Pod information in the events Tracee emits for consumers. One such consumer could be Postee - taking in events from Tracee and then being able to "act" upon them.
> 
> A use case would be to label a pod that was reported in a Tracee event. For this we could use `kubectl` along with a label selector that matches the relevant pod with its name, as reported by the Tracee event.

Thanks for the heads up. I think we should be in sync about what fields we want to add as part of this PR (for example i'm not sure the `containerName` field will be very useful), since we will need to add them to `trace.Event`, which is a separate PR.

pkg/containers/containers.go
```diff
@@ -28,20 +30,21 @@ type Containers struct {
...
+	Container runtime.ContainerMetadata
+	Runtime   string
+	Ctime     time.Time
+	expiresAt time.Time
 }
 
 // InitContainers initializes a Containers object and returns a pointer to it.
 // User should further call "Populate" and iterate with Containers data.
-func InitContainers() (*Containers, error) {
+func InitContainers(sockets runtime.Sockets) (*Containers, error) {
```
mtcherni95, line 47: the stauts of `sockets` is supposed to change? if yes why is it passed by value? shouldn't it be passed by reference/pointer?

pkg/containers/containers.go
```diff
@@ -50,14 +53,28 @@ func InitContainers() (*Containers, error) {
 		}
 	}
 
+	runtimeService := RuntimeInfoService()
+
+	registerEnricher(&runtimeService, sockets, "containerd", runtime.ContainerdEnricher)
```
mtcherni95, line 58: error checking, also I am not sure i understand the logics here. Can you please describe it?

pkg/containers/containers.go
NDStrahilevitz: I'm passing it by value because it's not supposed to change here, this is intentional. Sockets are initialized in CLI config time and injected downwards.

cmd/tracee-ebpf/main.go
```diff
@@ -299,6 +311,11 @@ func main() {
 				Value:   cli.NewStringSlice("none"),
 				Usage:   "Control event caching queues. run '--cache help' for more info.",
 			},
+			&cli.StringSliceFlag{
+				Name:  "containers",
```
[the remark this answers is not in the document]
NDStrahilevitz: Yeah i've had the same feeling, but container-runtime is also a bit of a mouthful. I thought of `sockets` as an alternative but it's also a bit unexpressive of the actual intention.
Maybe a third opinion can decide, because I don't really like any of these options.

cmd/tracee-ebpf/internal/flags/flags-containers.go
```diff
@@ -0,0 +1,81 @@
...
+func autodiscoverSockets() runtime.Sockets {
+	sockets := runtime.Sockets{}
+	const (
+		defaultContainerd = "/var/run/containerd/containerd.sock"
+		defaultDocker     = "/var/run/docker.sock"
+		defaultCrio       = "/var/run/crio/crio.sock"
+		defaultPodman     = "/var/run/podman/podman.sock"
+	)
+
+	sockets.Register("containerd", defaultContainerd)
```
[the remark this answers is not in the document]
NDStrahilevitz: I don't want to fail on this, maybe printing to stderr would be appropriate though WDYT?

pkg/containers/runtime/runtime.go
```diff
@@ -0,0 +1,44 @@
...
+	if s.sockets == nil {
+		s.sockets = make(map[string]string)
+	}
+
+	_, err := os.Stat(socket)
+	if err == nil {
+		s.sockets[runtime] = socket
+	}
+
+	return err
```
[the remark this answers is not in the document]
NDStrahilevitz: The error will be returned correctly here, if it's nil the register will happen and nil will be returned, otherwise we return the os.Stat error and not do anything.
However I think adding the "invalid socket" would be better so i'll do it like this.

pkg/containers/runtime/runtime.go
```diff
@@ -0,0 +1,44 @@
...
+	Name        string
+	Image       string
+}
+
+type ContainerEnricher interface {
+	Get(containerId string, ctx context.Context) (ContainerMetadata, error)
+}
+
+type Sockets struct {
+	sockets map[string]string
```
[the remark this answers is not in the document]
NDStrahilevitz: I'm probably going to make this an enum/const, I just wanted something working first.

cmd/tracee-ebpf/internal/flags/flags-containers.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+//check default paths for all supported container runtimes and aggregate them
+func autodiscoverSockets() runtime.Sockets {
+	sockets := runtime.Sockets{}
+	const (
+		defaultContainerd = "/var/run/containerd/containerd.sock"
+		defaultDocker     = "/var/run/docker.sock"
+		defaultCrio       = "/var/run/crio/crio.sock"
+		defaultPodman     = "/var/run/podman/podman.sock"
```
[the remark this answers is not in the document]
NDStrahilevitz: Not yet but i'm keeping the socket for future use, because we will want to support podman. Maybe we will get it in this PR as well. I can remove it in the end if we won't.

cmd/tracee-ebpf/internal/flags/flags-containers.go
```diff
@@ -0,0 +1,81 @@
...
+
+	for _, flag := range containerFlags {
+		parts := strings.Split(flag, ":")
+		if len(parts) != 2 {
+			return sockets, fmt.Errorf("failed to parse container flags (must be of format {runtime}:{socket path})")
+		}
+		containerRuntime := parts[0]
+
+		if !contains(supportedRuntimes, containerRuntime) {
+			return sockets, fmt.Errorf("provided unsupported container runtime (see --containers help for supported runtimes)")
```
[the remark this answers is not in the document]
NDStrahilevitz: Do we support case insensitive flags in other places? I personally don't find it so important, i think it's pretty easy to understand from the help what the supported values are, but if we support it in other flags i'll do it here.

go.mod
```diff
@@ -77,12 +90,15 @@ require (
...
+	golang.org/x/text v0.3.7 // indirect
+	google.golang.org/genproto v0.0.0-20211208223120-3a66f561d7aa // indirect
 	google.golang.org/protobuf v1.27.1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 )
+
+replace github.com/kubernetes/cri-api => k8s.io/cri-api v0.23.5-rc.0
+
+replace github.com/aquasecurity/tracee/types => github.com/NDStrahilevitz/tracee/types v0.0.0-20220317134116-17003f78d7a4
```
yanivagman, line 104: I didn't do a CR yet on this one, but from a quick look, this line should probably be removed

pkg/containers/containers.go
```diff
@@ -183,12 +200,18 @@ func (c *Containers) populate() error {
 // NOTE: ALL given cgroup dir paths are stored in CgroupInfo map.
 func (c *Containers) CgroupUpdate(cgroupId uint64, path string, ctime time.Time) (CgroupInfo, error) {
 	containerId, runtime := getContainerIdFromCgroup(path)
+	metadata, err := c.enricher.Get(containerId, runtime, context.Background())
```
yanivagman, line 203: if containerid is empty, there is no point in making this call

NDStrahilevitz: You're right, i'll fix that.

cmd/tracee-ebpf/internal/flags/flags-containers.go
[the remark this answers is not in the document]
mtcherni95: ok sounds good. Let's just remember to actually remove it if not used, as you said.

cmd/tracee-ebpf/internal/flags/flags-containers.go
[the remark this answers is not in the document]
mtcherni95: up to you - not so critical for me

pkg/containers/runtime/runtime.go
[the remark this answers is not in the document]
mtcherni95: ok - let's make sure it will be included in this PR (maybe you can do the change once the PR is upgraded from draft)

cmd/tracee-ebpf/internal/flags/flags-containers.go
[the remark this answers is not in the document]
mtcherni95: yeah let's add to the error channel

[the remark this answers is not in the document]
NDStrahilevitz: sounds like a good solution to me

pkg/containers/containers.go
NDStrahilevitz: Done, let me know what you think

cmd/tracee-ebpf/internal/flags/flags-containers.go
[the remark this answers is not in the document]
NDStrahilevitz: done

pkg/containers/runtime/runtime.go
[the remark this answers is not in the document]
NDStrahilevitz: done

pkg/containers/service.go
```diff
@@ -0,0 +1,35 @@
...
+	enrichers map[string]runtime.ContainerEnricher
+}
+
+func RuntimeInfoService() runtimeInfoService {
+	return runtimeInfoService{
+		enrichers: make(map[string]runtime.ContainerEnricher),
+	}
+}
+
+func (e *runtimeInfoService) Register(runtime string, socket string, enricherBuilder func(socket string) (runtime.ContainerEnricher, error)) error {
```
mtcherni95: what if a user calls more than once register? we should protect it inside the function (not allowing multiple registering)

cmd/tracee-ebpf/internal/flags/flags-containers.go
```diff
@@ -0,0 +1,89 @@
...
+		if err != nil {
+			return sockets, fmt.Errorf("failed to register runtime socket, %s", err.Error())
+		}
+	}
+
+	return sockets, nil
+}
+
+//check default paths for all supported container runtimes and aggregate them
+func autodiscoverSockets() runtime.Sockets {
```
mtcherni95, line 67: `autoDiscoverSocket`

pkg/containers/runtime/runtime.go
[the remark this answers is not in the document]
NDStrahilevitz: Done as an enum in latest commit, WDYT?

pkg/containers/runtime/sockets.go
```diff
@@ -0,0 +1,34 @@
+package runtime
```
mtcherni95, line 1: containersruntime?

pkg/containers/runtime/sockets.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"fmt"
+	"os"
+)
+
+type Sockets struct {
+	sockets map[RuntimeId]string
```
mtcherni95, line 9: I am wondering if there will be any scenario where sockets are registered on runtime...

pkg/containers/service.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/aquasecurity/tracee/pkg/containers/runtime"
+)
+
+type runtimeInfoService struct {
+	enrichers map[runtime.RuntimeId]runtime.ContainerEnricher
```
mtcherni95, line 11: also here, will we have any scenario in which enricher is added(/deleted) on the fly (dynamically)

mtcherni95:
Adding this diagram to quickly get an overall view of the additions/changes:
image

pkg/containers/service.go
NDStrahilevitz: Sounds like an override? I don't think it's an inappropriate use of the API.

pkg/containers/runtime/sockets.go
NDStrahilevitz: Maybe if we detect the socket has been moved? If we wouldn't allow runtime registering, what would you suggest the API be instead?

pkg/containers/service.go
NDStrahilevitz: Theoretically if we can detect the installation of a runtime while tracee is running this could be useful.

pkg/containers/containers.go
```diff
@@ -50,14 +53,28 @@ func InitContainers() (*Containers, error) {
 		}
 	}
 
+	runtimeService := RuntimeInfoService()
```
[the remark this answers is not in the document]
NDStrahilevitz: The only reason I did this was because maps aren't automatically allocated on struct construction, so I had to include some constructor. I don't think we should call this function New due to the package scope but we can prefix this function with a `New`, so it will be `NewRuntimeInfoService` instead.

I do think that what we have now expresses the intent clearly enough though.

pkg/containers/runtime/runtime.go
```diff
@@ -0,0 +1,44 @@
+package runtime
```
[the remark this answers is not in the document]
NDStrahilevitz: While I agree the name is more indicative, I think we would be renaming this a lot in consuming packages, and the name are already getting a bit long in some places anyway.

NDStrahilevitz:
@yanivagman I've added pod metadata through labels as we discussed

NDStrahilevitz:
@rafaeldtinoco I did some tests to see how this feature works under stress (running tracee through a container with limited cpu), and found out that we will need to mount the runtime sockets to the container, usually they are found in /var/run.

Should we mount these by default? I'd like your opinion on tackling this.

pkg/containers/containers.go
```diff
@@ -182,62 +200,87 @@ func (c *Containers) populate() error {
...
 	c.mtx.Lock()
 	c.cgroups[uint32(cgroupId)] = info
 	c.mtx.Unlock()
 
+	c.enrichCgroupInfo(cgroupId)
+
 	return info, nil
 }
 
+func (c *Containers) enrichCgroupInfo(cgroupId uint64) {
```
rafaeldtinoco, line 224: Could you add function documentation here and in most of other functions you created ? When yaniv created this code I did a PR on top just to document each function, as the cgroup id <-> container id mapping was not very straightforward to understand. Same happens here.

rafaeldtinoco (commented):
Nice job. I have not gone through the runtime specific enrichers yet, will do when the PR is ready. I thought it was better to have a general interface check first, and left some comments.

rafaeldtinoco:
> @rafaeldtinoco I did some tests to see how this feature works under stress (running tracee through a container with limited cpu), and besides confirming this will work, I also found out that we will need to mount the runtime sockets to the container, usually they are found in /var/run.

Yep =(

> Should we mount these by default? I'd like your opinion on tackling this (with regards to documentation and deployment scrips).

Problem is that, like your code predicts, there might be more than 1 runtime simultaneously and we would have to bind mount each socket (as they might be placed in different places, correct ?).

More and more I see need for a "wrapper" execution script I'm afraid (calling docker run with appropriate arguments depending on the execution environment).

@danielpacak thoughts ? We've discussed in the past, in a tracee meeting, about user experience we came up with the idea of having a wrapper script (curl http://awesome-tracee-url/wrapper | bash) that would take decisions based on running environment (like docker arguments and so).

Of course, in parallel, we have to document the bind mount needs (for container enrich feature)...

pkg/containers/containers.go
NDStrahilevitz: Yes, i'll work on adding more documentation around this PR.

pkg/containers/containers.go
```diff
@@ -50,14 +54,28 @@ func InitContainers() (*Containers, error) {
...
 	return &Containers{
 		cgroupV1: cgroupV1,
 		cgroupMP: "",
 		cgroups:  make(map[uint32]CgroupInfo),
 		mtx:      sync.RWMutex{},
+		enricher: runtimeService,
 	}, nil
 }
 
+func registerEnricher(runtimeService *runtimeInfoService, sockets runtime.Sockets, runtime cruntime.RuntimeId, enricher func(socket string) (runtime.ContainerEnricher, error)) error {
```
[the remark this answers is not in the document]
NDStrahilevitz: I think i'm going to change the RuntimeInfoService to receive Sockets as it's input, so there won't be a need for this function.
If it doesn't look better that way, we can move registration to `tracee.go` how does that sound?

pkg/containers/containers.go
```diff
@@ -15,6 +16,8 @@ import (
 	"unsafe"
 
 	"github.com/aquasecurity/libbpfgo"
+	"github.com/aquasecurity/tracee/pkg/containers/runtime"
+	cruntime "github.com/aquasecurity/tracee/pkg/containers/runtime"
```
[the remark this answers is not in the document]
NDStrahilevitz: Haha, no idea how gofmt let that stay, i'll remove the first one. (second one was to resolve conflicts with parameters called `runtime`).

cmd/tracee-ebpf/internal/flags/flags-containers.go
```diff
@@ -0,0 +1,106 @@
...
+func autoDiscoverSockets() runtime.Sockets {
+	sockets := runtime.Sockets{}
+	const (
+		defaultContainerd = "/var/run/containerd/containerd.sock"
+		defaultDocker     = "/var/run/docker.sock"
+		defaultCrio       = "/var/run/crio/crio.sock"
+		defaultPodman     = "/var/run/podman/podman.sock"
+	)
+
+	registerSocket(&sockets, "containerd", defaultContainerd)
```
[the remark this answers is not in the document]
NDStrahilevitz: Yes, I think adding the autodiscover mechanism to the help is relevant, I'll do that.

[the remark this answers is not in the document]
NDStrahilevitz: With regards to debug messages, I went with the opposite way where failures on autodiscovery would be printed to stderr, but maybe we can just print the whole autodiscover process to stdout on debug, errors (which really aren't proper errors) included, WDYT?

pkg/containers/containers.go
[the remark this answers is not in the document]
NDStrahilevitz: I've opted for nesting this function like you said.

cmd/tracee-ebpf/internal/flags/flags-containers.go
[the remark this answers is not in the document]
rafaeldtinoco: Id like to have the output in debug. Just like the OSINFO discoveries are printed. And errors should be printed as well. Thanks!

pkg/containers/containers.go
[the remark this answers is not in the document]
rafaeldtinoco: Have you thought about this one ?

rafaeldtinoco (commented):
It all looks good to me. A small reservation I have is the InitContainers() calling cruntime.XXXEnricher() functions instead of calling Object->Enricher() interface implementation (or something similar).

Feel free to think about ^ and yes, or no, suggest a final PR in non-draft mode so we can review it once more and go for a merge. Cheers!

cmd/tracee-ebpf/internal/flags/flags-containers.go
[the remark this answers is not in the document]
NDStrahilevitz: Alright, i'm passing debug to the prepare function now - on debug it will print successful and failed autodiscovered socket registers to stdout and stderr respectively. Does that sound reasonable? I think it emulates what the OSInfo does well enough (of course that's all done in main.go unlike this flag).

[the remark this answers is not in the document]
rafaeldtinoco: Yep. And add a prefix to the debug message, something like ENRICH: or something you think its meaningful (just like OSInfo: prefix we currently have). Up to you.

pkg/containers/containers.go
rafaeldtinoco: Thanks. Please mark this done once you're finished with this.

pkg/containers/containers.go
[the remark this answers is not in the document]
NDStrahilevitz: Are you talking about the name or where we're registering?
On the former, i'm still of the same opinion that it shouldn't be New(), I'm open to other suggestions (for the struct as well as the corresponding constructor).
On the latter, I'm not really sure where else we would register them, initially all the enrichers were hardcoded in the service and a switch was used to derive which one of them to use. So really the only other way to register that I see would be even more inwards. Externally from containers, I'm not sure where it would be appropriate to do, since the containers package to me seems to be the root of this feature, initializing it there made sense to me.

pkg/containers/containers.go
NDStrahilevitz: Resolving since I removed the function

pkg/containers/containers.go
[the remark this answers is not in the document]
NDStrahilevitz: I've refactored the registration process by having the service receive the `Sockets` as an arguments before. WDYT?

cmd/tracee-ebpf/internal/flags/flags-containers.go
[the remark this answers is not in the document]
NDStrahilevitz: Done.

cmd/tracee-ebpf/main.go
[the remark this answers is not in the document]
yanivagman: I think runtime[s] might be a better fit than containers. One reason for this is that containers can be confused with the containers filter. WDYT?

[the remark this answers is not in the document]
NDStrahilevitz: No problem with that, just to make sure you mean `--runtimes` right?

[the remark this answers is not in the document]
yanivagman: yes

pkg/containers/containers.go
```diff
@@ -182,62 +202,98 @@ func (c *Containers) populate() error {
...
 	return info, nil
 }
 
+// enrichCgroupInfo checks for a given cgroupId if it is relevant to some running container
+// it then calls the runtime info service to gather additional data from the container's runtime
+func (c *Containers) enrichCgroupInfo(cgroupId uint64) error {
+	c.mtx.RLock()
+	info := c.cgroups[uint32(cgroupId)]
+	c.mtx.RUnlock()
+	containerId := info.Container.ContainerId
```
yanivagman, line 237: You make an assumption here that info was initialized, which is true now, but might change in the future for whatever reason.

rafaeldtinoco:
> @rafaeldtinoco @yanivagman I'm currently blocked from moving this from draft until the relevant types PR #1601 is merged. Can we get it reviewed so this can go forward?

Im leaving a comment there then.

pkg/containers/containers.go
```diff
@@ -182,62 +202,98 @@ func (c *Containers) populate() error {
...
+	runtime := info.Runtime
+
+	if containerId != "" && runtime != cruntime.Unknown {
+		metadata, err := c.enricher.Get(containerId, runtime, context.Background())
```
rafaeldtinoco, line 241: You're always using context.Background(), is it really needed to have it as the 3rd argument then ?

Besides, shouldn't we create a timed-out context in here so we don't let ContainerEnricher implementing the interface to be stuck in Get() forever and ever ? (I'm not sure all APIs talking to container daemons will have timed-out attempts, etc).

pkg/containers/containers.go
```diff
@@ -182,62 +202,98 @@ func (c *Containers) populate() error {
...
+
+		//if enrichment fails, just return early
+		if err != nil {
+			return err
+		}
+
+		info.Container = metadata
+		c.mtx.Lock()
+		c.cgroups[uint32(cgroupId)] = info
```
rafaeldtinoco, line 250: Shouldn't you check if `c.cgroups[uint32(cgroupdId)]` still exists when holding the mutex again ? It could have been removed after you read it for the first time AND you would be reintroducing it here (unsure if it could happen in practice, but there is a possibility in theory).

rafaeldtinoco (changes requested):
I haven't dig into other runtimes other than docker for now (which is okay as they are self-contained). I have provided some feedback that I would like clarification.

I would like to have @yanivagman's input as well.

I think we're close to merging this.. Thanks!

pkg/containers/runtime/containerd.go
```diff
@@ -0,0 +1,85 @@
...
+func ContainerdEnricher(socket string) (ContainerEnricher, error) {
+	enricher := containerdEnricher{}
+	unixSocket := "unix://" + strings.TrimPrefix(socket, "unix://")
+
+	client, err := containerd.New(socket)
+	if err != nil {
+		return nil, err
+	}
+
+	conn, err := grpc.Dial(unixSocket, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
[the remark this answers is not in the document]
rafaeldtinoco: @NDStrahilevitz is there something to be done here ?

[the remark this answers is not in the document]
NDStrahilevitz: There's probably a way to make this blocking with a timeout, is this what we want?

pkg/containers/containers.go
NDStrahilevitz: Then we should just use the provided Context argument here, seems like an oversight from me. Would you prefer if we hardcoded the context timeout logic?

pkg/containers/containers.go
NDStrahilevitz: Fair point, I can add that check. Should we return an error in such a case?

rafaeldtinoco: Perhaps just ignore ? Its like "we don't need enrichment any longer".. unless you can think of something else.

pkg/containers/runtime/containerd.go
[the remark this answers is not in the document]
rafaeldtinoco: I think that might be related to the time out I thought about in https://github.com/aquasecurity/tracee/pull/1572#discussion_r852831740 (hadn't seen this before). We would guarantee enrichers timeout if they're doing something odd.

pkg/containers/containers.go
NDStrahilevitz: Reasonable to me, ignoring was also what I had in mind.

pkg/containers/containers.go
rafaeldtinoco: Sorry, I don't follow what you said. I thought either we remove the ctx argument OR created a timed-out context so enricher would not hang forever.

NDStrahilevitz: I misread where this code comment was.
We have the the context as an argument exactly so such a change (like adding a timeout) could be done once here, instead of being done over and over in each implementation.
Since I had no decision on what to do in here before, I just passed the background context. Now I can simply add a timeout logic here.

pkg/containers/runtime/sockets.go
NDStrahilevitz: I personally find that name too long. 
I had a conversation with Yaniv where I was considering `cri` but we agreed that may mislead to think this relates to the k8s cri (although there are similarities).

pkg/containers/containers.go
NDStrahilevitz: Fixed this assumption to return early with an error if the cgroup doesn't exist anymore.

pkg/containers/containers.go
NDStrahilevitz: Done, please resolve if it's fine by you now.

pkg/containers/containers.go
NDStrahilevitz: Done.

cmd/tracee-ebpf/internal/flags/flags-containers.go
```diff
@@ -0,0 +1,116 @@
...
+
+func ContainersHelp() string {
+	return `Select which container runtimes to connect to for container events enrichment.
+By default, if no flag is passed, tracee will automatically detect installed runtimes by going through known runtime socket paths.
+
+Tracee will look for the following paths:
+1. Docker:     /var/run/docker.sock
+2. Containerd: /var/run/containerd/containerd.sock
+3. CRI-O:      /var/run/crio/crio.sock
+
```
yanivagman, line 19: Don't we want to add podman as well? What is missing in order to add such a support?

cmd/tracee-ebpf/internal/flags/flags-containers.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func contains(s []string, val string) bool {
+	for _, str := range s {
+		if str == val {
+			return true
+		}
+	}
+	return false
+}
```
yanivagman, line 39: I've seem this function repeated in previous PRs (although I don't remember where, and it might had another name), maybe we should put it somewhere under "utils" to avoid code duplication?

cmd/tracee-ebpf/main.go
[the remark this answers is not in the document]
yanivagman: This is still something we might want to change.
Maybe we should call it "crs" (container runtime socket)? It is both short and more correct than just "container" or "runtime"

pkg/containers/containers.go
```diff
@@ -182,62 +202,112 @@ func (c *Containers) populate() error {
...
+		return fmt.Errorf("no cgroup to enrich")
+	}
+
+	containerId := info.Container.ContainerId
+	runtime := info.Runtime
+
+	if containerId != "" && runtime != cruntime.Unknown {
+		//There might be a performance overhead with the cancel
+		//But, I think it will be negligable since this code path shouldn't be reached too frequently
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
yanivagman, line 249: Does this mean that the pipeline is stuck until we get the response (possibly for a few seconds)?
Won't this cause us loosing many events?
Maybe we should consider processing other events somehow while we're waiting here

yanivagman (dismissed):
I left a few more comments.
Other than that, I think that we should add some of the new fields into the table printer (for sure we want it to be in the table-verbose printer).
In addition, a few words in the documentation might also be helpful for such a change

cmd/tracee-ebpf/internal/flags/flags-containers.go
NDStrahilevitz: Yes, but I've decided to leave it to a later PR, since it requires implementing our own client since I couldn't find a good client library (their own library is pretty bad IMO, it's version tied and has a ton of unrelated dependencies).

cmd/tracee-ebpf/internal/flags/flags-containers.go
NDStrahilevitz: We should add a generic version once we move to 1.18 probably.

pkg/containers/containers.go
NDStrahilevitz: I think this code path is in a separate goroutine than event processing, so no? 
Maybe we should make sure enrichment always gets it's own goroutine and then we'll be fine, but I'm not sure it's necessary.

NDStrahilevitz:
> I think that we should add some of the new fields into the table printer (for sure we want it to be in the table-verbose printer).

Definitely, i'll work on that.

> In addition, a few words in the documentation might also be helpful for such a change

Do you have any pointers where it might be relevant? I've added a comment here about it before with regards to socket mounting. Any other idea?

pkg/containers/containers.go
yanivagman: I actually don't think it's in a separate goroutine, so this is why I'm asking.
If adding a new go routine, you should note that we might have event ordering issues.
In addition, if another event comes from the same container, you need to make sure that you don't query the runtime again.
So a proper solution might require:
1. Marking that there's an ongoing query for a given container (can be saved as a bool in the cgroup info)
2. Queue events from that container while there's an ongoing query

yanivagman:
> Do you have any pointers where it might be relevant? I've added a comment here about it before with regards to socket mounting. Any other idea?

I thought we have a detailed description of the fields in the documentation, but I don't see any other than this quickstart: https://github.com/aquasecurity/tracee/blob/main/docs/tracee-ebpf/index.md#quickstart
So maybe this can wait for another PR

pkg/containers/containers.go
yanivagman: > 1. Marking that there's an ongoing query for a given container (can be saved as a bool in the cgroup info)

In addition to that, if we already tried to query a given container id and failed in the past (for whatever reason that the runtime isn't aware of the container, or we might have a wrong container id parsed from the cgroup name), I think we should not try to query again. We can keep this information in the cgroupinfo as well

NDStrahilevitz: I think the solution is refactoring this to be a part of the event pipeline, WDYT?

yanivagman: There might be other ways to solve this, but this is a possible solution.
We can have another "enrichment" pipeline stage.

rafaeldtinoco: Nice catch @yanivagman. So, the idea here would be to cache ALL events from a specific container while it is being enriched ? And then release them once the enrichment succeeds or fails ? If that is the case, I would place this as a last step in the pipeline (after sorting and caching). 

Be aware of the following as well: When we release all the events... we might have held up to 10 seconds of events (timeout for the enrichment to fail ?). This means that we could be releasing a big amount of data (max throughput * 10, could be something like 40MB in worst cases) and this could also cause latencies big latency in the pipeline as well.

NDStrahilevitz: There are probably other ways to do it, but i'm going for a strategy similar to what @rafaeldtinoco described currently as a stage in the pipeline.

rafaeldtinoco: @NDStrahilevitz maybe we should reduce the timeout to the absolute needed time. And mark the timedout ones not to re-try them over and over like @yanivagman suggested.

NDStrahilevitz: I'm wondering if we should let the timeout(s) be configurable under the `crs` (that will be the new flag name) flag.
In addition, I am currently marking falied tries in the pipeline stage.

Currently, there's no good stage in the pipeline to add this logic, since I need the CgroupMkdir logic to run first, which is in the `processEvents` part. So I will need a PR to split `processEvents` from the pipeline sink stage (they are currently comined), and then I can put this enrich stage in between. In general, I think we could benefit from refactoring the pipeline to some degree, see #1684 that I opened.

NDStrahilevitz: Updating that I have solved this through #1689. I need it merged so I can rebase this PR on top of it.

rafaeldtinoco: PR #1689 has been merged. Please rebase this one and highlight the specific changes about:

1. enqueueing events when a specific event+containerid is being enriched
2. marking enrichment has failed and wont be tried again

so we can move on with this PR as well. Thanks!

PS:

Be aware that the enrichment process will cause latencies in the pipeline (unfortunately) as we can't queue events that need enrichment AND still process the subsequent ones, that don't need, because of chronological/ordering needs. 

IMO, we should "enforce" caching in the pipeline when using container enrichment (this would allow enrichment to take sometime while reducing the pressure being put by the perfbuffer in the origin).

NDStrahilevitz:
I've rebased the PR on top of the pipeline changes from #1689 and enrichment now happens as part of the pipeline, so it should be non blocking now.

rafaeldtinoco:
Okay, going through a final review before merging this so we can release tracee soon.

pkg/ebpf/events_enrich.go
```diff
@@ -0,0 +1,124 @@
...
+				}
+			}
+		}
+	}()
+
+	go func() {
+		for enrich := range enriches {
+			cgroupId := enrich.cgroupId
+			//mark the query as finished, it should not be attempted again and no new events should enter it's queue
+			attempted[cgroupId] = true
```
rafaeldtinoco, line 67: If you do this here.. and an event for this cgroupid arrives in the other routine, it will be sent to the out channel BEFORE the buffered events, making them out of order, no ?

rafaeldtinoco: Ok so we need to block events from entering the queue when we are processing it AND, at the same time, we can't allow that event to be sent to "out" channel just yet.

rafaeldtinoco (approved):
LGTM.. but I'm waiting @NDStrahilevitz to clarify one thing (mentioned in my last comment). After that is clarified (and fixed, if needed), we may merge this one.

rafaeldtinoco:
I spoke to @NDStrahilevitz and I'll be merging this after the fix I mentioned earlier and perhaps addressing this:

```
$ sudo ./dist/tracee-ebpf -o format:json -o option:parse-arguments -o option:detect-syscall -trace container -trace comm=bash -trace follow
failed to register enricher: error registering enricher: unsupported runtime crio
```

as well (shouldn't be printed without --debug iirc).

rafaeldtinoco:
Enrichment from Docker (at least) contains a leading "/":

image

rafaeldtinoco:
> Enrichment from Docker (at least) contains a leading "/":
> 
> image

This has been answered here: https://github.com/moby/moby/issues/6705 and the leading slash is part of the container name to tell its a local container. I'm going to keep it for now, as it is provided by docker API, but we should discuss if we would like to keep it or not.

rafaeldtinoco:
I have made some changes, which I'll be merging in another PR, using all commits from this one. The changes are just addressing the synchronization issue I brought to discussion (and agreed with Nadav).

I have ran race detection and it seems good. I have also done a very small test to see if events would be enqueued while being enriched, and if after enriched they would follow other execution path, and it seems to be working as expected:

```go
// non container event and not cgroup_mkdir (or the event already enriched) skip per cgroupId caching
if (event.ContainerID == "" && event.EventID != int(CgroupMkdirEventID)) || event.ContainerImage != "" {
	// enrichment attempt is being done, or has failed, so do not try it again
	attemptedMutex.RLock()
	if _, ok := attempted[cgroupId]; ok {
		// wait transaction (cache flush) to finish so cached events and this one are ordered
		attempted[cgroupId].Lock()
		fmt.Println("enriched, or not")
		out <- event
		attempted[cgroupId].Unlock()
		attemptedMutex.RUnlock()
		continue
	}
	attemptedMutex.RUnlock()
	out <- event // some events might not have an attempted map
	continue
} else {
	//if the container queue has not been created - create the container queue and invoke to enrich query
	if _, ok := queues[cgroupId]; !ok {
		queues[cgroupId] = make(chan *trace.Event, 1000)

		go func() {
			metadata, err := t.containers.EnrichCgroupInfo(cgroupId)
			enriches <- enrichResult{cgroupId, metadata, err}
		}()
	}
	fmt.Println("enqueued")
	queues[cgroupId] <- event
}			
```

```
$ sudo ./dist/tracee-ebpf -o none -o option:parse-arguments -o option:detect-syscall -trace container -trace comm=bash -trace follow --crs docker:/var/run/docker.sock
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enqueued
enriched, or not
enriched, or not
enriched, or not
enriched, or not
enriched, or not
enriched, or not
enriched, or not
enriched, or not
enriched, or not
enriched, or not
enriched, or not
enriched, or not
```

rafaeldtinoco:
@NDStrahilevitz, @mtcherni95

After merging this one, I have tried to address some concerns through:

https://github.com/aquasecurity/tracee/pull/1764

after a quick talk with Nadav. And then I have faced lots of failing tests and decided to revert this merge at:

https://github.com/aquasecurity/tracee/pull/1768

There are some concerns related to 

I liked the per containerId (cgroupId) caching made, I just think it needs more testing regarding synchronization (`queues` map should be protected, but protecting it might hold the pipeline for long time if cached events is too big (10 seconds), etc). I think we should revisit it and do some more testing. I'm not cutting a new release before we merge it again (fyi).

Issue #1573, opened 2022-03-23
adding uprobes to Tracee


itamarmaouda101:
Hi 

We should add support to uprobes events in Tracee. in order to add the ability to trace user-space events.

Currently, I'm working on PR that adds that to Tracee (:

itaysk:
this sounds very interesting! please describe the use case - why you need it and how you are seeing the user uses it. Also if you have a design in mind please discuss it here before you start implementing it. We also planned something similar so please coordinate with us before you start working on this, to make sure you spend your time effectively.

itaysk:
Also it's a dup of #363 so please let's discuss is there

PR #1574, opened 2022-03-23
Introduce uprobes to Tracee

itamarmaouda101:
Hi 
This  PR solves #1573 .

Its still a draft because I'm working to add an event that will be example of adding an uprobes event


rafaeldtinoco:
Closing this PR (we might revisit it in the future but is should be closed for now).

Issue #1538, opened 2022-03-09
Compiling tracee using docker containers fails


AsafEitani:
I came upon an issue with the new way to compile tracee with docker:
```
make -f builder/Makefile.tracee-make alpine-prepare
make -f builder/Makefile.tracee-make alpine-make ARG="clean all"
```
This results the following error:
`the input device is not a TTY`

The full make log:
```
[2022-03-09T12:14:31.407Z] Sending build context to Docker daemon  32.45MB
[2022-03-09T12:14:31.407Z] Step 1/9 : FROM alpine:3.15
[2022-03-09T12:14:31.407Z]  ---> c059bfaa849c
[2022-03-09T12:14:31.407Z] Step 2/9 : ARG uid=1000
[2022-03-09T12:14:31.407Z]  ---> Using cache
[2022-03-09T12:14:31.407Z]  ---> 18607f300744
[2022-03-09T12:14:31.407Z] Step 3/9 : ARG gid=1000
[2022-03-09T12:14:31.407Z]  ---> Using cache
[2022-03-09T12:14:31.407Z]  ---> f36978b5ee1f
[2022-03-09T12:14:31.407Z] Step 4/9 : RUN apk --no-cache update &&     apk --no-cache add sudo coreutils findutils &&     apk --no-cache add bash git curl rsync &&     apk --no-cache add musl-dev libc6-compat &&     apk --no-cache add llvm clang go make gcc &&     apk --no-cache add linux-headers &&     apk --no-cache add elfutils-dev &&     apk --no-cache add libelf-static &&     apk --no-cache add zlib-static &&     curl -L -o /usr/bin/opa https://github.com/open-policy-agent/opa/releases/download/v0.33.1/opa_linux_amd64_static &&     chmod 755 /usr/bin/opa
[2022-03-09T12:14:31.407Z]  ---> Using cache
[2022-03-09T12:14:31.407Z]  ---> 50450734164e
[...]

PR #1575, opened 2022-03-23
fix: compile tracee in Docker container without interactive mode

danielpacak:
Resolves: #1538


AsafEitani:
@rafaeldtinoco raised the concern that if the the container execution is not interactive (thus not blocking) it may result that the compilation process carries on without the container execution finished.
Maybe keeping the -i will work? I think the issue was with the -t.
Anyway apparently the docker compiled binary is intended to be used by the other docker images - which doesn't meet my need. so the issue is no longer a "issue" for me.

PR #1576, opened 2022-03-24
chore(docs): preview mkdocs on macOS

danielpacak:
Running make -f builder/Makefile.mkdocs on macOS doesn't really work.
With this patch to Vagrantfile we can run MkDocs server in Vagrant
guest machine and forward it to macOS host on http://localhost:8000.


rafaeldtinoco (approved):
LGTM.

Issue #1577, opened 2022-03-24
The default tracee-rules templates are not added to container image under /tracee/templates/*.tmpl


danielpacak:
The default tracee-rules templates are not added to the container image. We used to have them in 0.6.5

```console
$ docker run -it --rm --entrypoint "" aquasec/tracee:0.6.5 ls -l /tracee
total 33352
-rwxr-xr-x 1 root root     1370 Dec  6 16:06 entrypoint.sh
drwxr-xr-x 2 root root     4096 Dec  6 16:07 rules
drwxr-xr-x 2 root root     4096 Dec  6 16:08 templates
-rwxr-xr-x 1 root root 12080280 Dec  6 16:07 tracee-ebpf
-rwxr-xr-x 1 root root 22055184 Dec  6 16:07 tracee-rules
```

```console
$ docker run -it --rm --entrypoint "" aquasec/tracee:v0.7.0-rc-1 ls -l /tracee
total 43772
-rwxr-xr-x 1 root root     4891 Mar 21 17:10 entrypoint.sh
drwxr-xr-x 2 root root     4096 Mar 21 17:16 rules
-rwxr-xr-x 1 root root 18240304 Mar 21 17:15 tracee-ebpf
-rwxr-xr-x 1 root root 26562752 Mar 21 17:15 tracee-rules
```

The templates are referred to by the deploy/kubernetes/tracee-postee/tracee.yaml deployment descriptor. Here's the error you'll get after deploying in Kubernetes cluster:

```
INFO: probing tracee-ebpf capabilities...
INFO: starting tracee-ebpf...
INFO: starting tracee-rules...
Loaded 14 signature(s): [TRC-1 TRC-13 TRC-2 TRC-14 TRC-3 TRC-11 TRC-9 TRC-4 TRC-5 TRC-12 TRC-8 TRC-6 TRC-10 TRC-7]
Serving metrics endpoint at :3366
2022/03/24 11:06:39 error preparing webhook template: open ./templates/rawjson.tmpl: no such file or directory
```

PR #1578, opened 2022-03-24
fix(deploy): update DaemonSet config to deploy tracee in K8s

danielpacak:

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -9,6 +9,18 @@ Vagrant.configure("2") do |config|
   # Forward MkDocs dev server to preview documentation on the host at http://localhost:8000/tracee
   config.vm.network :forwarded_port, guest: 8000, host: 8000
 
+  # Forward MicroK8s dashboard to access it on the host at https://localhost:10443
+  #
+  # To access the Kubernetes dashboard from the host run the following command:
+  #     kubectl port-forward --address 0.0.0.0 -n kube-system service/kubernetes-dashboard 10443:443
+  #
+  # To sing in use the token retrieved with
+  #     token=$(microk8s kubectl -n kube-system get secret | grep default-token | cut -d " " -f1)
+  #     kubectl -n kube-system describe secret $token
+  #
+  # TIP For Google Chrome you may allow insecure TLS connections at chrome://flags/#allow-insecure-localhost
+  config.vm.network :forwarded_port, guest: 10443, host: 10443
+
   config.vm.provider "virtualbox" do |vb|
     vb.gui = false
     vb.memory = "2048"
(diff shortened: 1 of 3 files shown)


deploy/kubernetes/tracee-falcosidekick/tracee.yaml
```diff
@@ -17,8 +17,8 @@ spec:
       name: tracee
     spec:
       containers:
-      - image: aquasec/tracee:latest
-        imagePullPolicy: Always
+      - image: docker.io/aquasec/tracee:v0.7.0-rc-1
+        imagePullPolicy: IfNotPresent
         args:
           - --webhook http://tracee-webhook:2801 --webhook-template ./templates/falcosidekick.tmpl --webhook-content-type application/json
```
danielpacak, line 23: Note that due to #1577 the Pod will enter the crash loop.

Vagrantfile
```diff
@@ -9,6 +9,18 @@ Vagrant.configure("2") do |config|
   # Forward MkDocs dev server to preview documentation on the host at http://localhost:8000/tracee
   config.vm.network :forwarded_port, guest: 8000, host: 8000
 
+  # Forward MicroK8s dashboard to access it on the host at https://localhost:10443
```
danielpacak, line 12: tracee_on_k8s

deploy/kubernetes/tracee-falcosidekick/tracee.yaml
```diff
@@ -27,11 +27,8 @@ spec:
         volumeMounts:
         - name: tmp-tracee
           mountPath: /tmp/tracee
-        - name: usr-src
```
danielpacak, line 30: I wasn't sure if we still need these volume mounts?

itaysk (approved):
LGTM
about the docker image versions - sure we want to change latest to pinned version and maintain it?
Also, just a reminder that the `v` in the image name needs to be fixed either before this PR is merged or with the same PR that fixes it.

danielpacak:
> LGTM
> about the docker image versions - sure we want to change latest to pinned version and maintain it?
> Also, just a reminder that the `v` in the image tag needs to be fixed either before this PR is merged or with the same PR that fixes it.

IMHO using pinned tag instead of latest adds some effort to maintain it, but at least we won't break installation experience for previous versions of tracee. For example, if next time we publish the latest image that requires more environment variables to be set, but previous deployment YAML doesn't set these envs tracee may not even star.

Regarding the v prefix we need to figure out if v0.7.0-rc-1 was a bug or a feature. I'll open another PR in any case to make sure we use the valid image tag. /cc @grantseltzer

itaysk:
> Regarding the v prefix we need to figure out if v0.7.0-rc-1 was a bug or a feature.

I think in docker it's conveniently omitted. Not sure why though

danielpacak:
> > Regarding the v prefix we need to figure out if v0.7.0-rc-1 was a bug or a feature.
> 
> I think in docker it's conveniently omitted. Not sure why though

We've merged #1586 . So I'm going to open another PR with image tags without the "v" prefix.

Issue #1558, opened 2022-03-16
Catch "dirty pipe" explotation


AlonZivony:
The new vulnerability in pipe and page cache is a specific use which could not be observed by our current events.
To catch it we can use 2 methods I thought about:
1. Getting the flags upon `splice` syscall of the `pipe_buffer` first in line in the output (similar to what [DataDog did](https://github.com/DataDog/datadog-agent/pull/11235/files)).
2. Creating a new event for `pipe_write` - check when writing to a page in a pipe what are the flags of the `page_buffer` and who owns the page cache, and by that be able to identify if the page is part of a file + has merge flag active.

itaysk:
we always try to do 1 over 2. Can you please add what's missing from the output today?

itaysk:
I'm removing the label since this issue captures the requirement expected from Tracee

AlonZivony:
> we always try to do 1 over 2. Can you please add what's missing from the output today?

The exploit uses the existing flags of the `pipe_buffer` in use by the `pipe_inode_info` of the output file (which is a pipe in this case) to get invalid flags to the page from the file's page cache.
Currently we have no way to know which flags are given to the `pipe_buffer` in use before (or after) the splice, so it is impossible to know if the flags are legitimate or not.
There might be other info we can get to monitor this exploit, I explore it right now.

rafaeldtinoco:
Alright, PR was rebased and merged so I guess this can be closed now. Thanks.

PR #1580, opened 2022-03-24
pkg/ebpf: created dirty_pipe_splice event

AlonZivony:
Created an event to catch splice to a pipe which results that the last `pipe_buffer` struct has the `PIPE_BUF_CAN_MERGE` flag on. 
This will enable to catch vulnerable pipes used.
To have visibility of using the vulnerable pipe, an event that catches `pipe_write` function to a vulnerable pipe is needed.

Partly solves #1558

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -195,7 +195,8 @@ Copyright (C) Aqua Security inc.
 #define SOCKET_DUP                      1032
 #define HIDDEN_INODES                   1033
 #define __KERNEL_WRITE                  1034
-#define MAX_EVENT_ID                    1035
+#define DIRTY_PIPE_SPLICE               1035
+#define MAX_EVENT_ID                    1036
 
 #define NET_PACKET                      4000
 
@@ -311,6 +312,19 @@ Copyright (C) Aqua Security inc.
 #ifndef CONFIG_ARCH_HAS_SYSCALL_WRAPPER
 #define CONFIG_ARCH_HAS_SYSCALL_WRAPPER 0
 #endif
+
+#endif // CORE
+
+/*================================ eBPF CO-RE SUPPORT =================================*/
+
+#ifdef CORE
+#define get_type_size(x) bpf_core_type_size(x)
+#define __get_node_addr(array, node_type, index) ((node_type *)((void *)array + (index * get_type_size(node_type))))
+#define get_node_addr(array, index) __get_node_addr(array, typeof(*array), index)
+
+#else
+#define get_type_size(x) sizeof(x)
+#define get_node_addr(array, index) &array[index]
 
 #endif // CORE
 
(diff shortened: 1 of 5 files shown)


AlonZivony:
Checked that is working properly on v5.10, v5.8 and v5.4.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4455,6 +4506,83 @@ int BPF_KPROBE(trace_security_inode_mknod)
...
+
+    struct pipe_buffer *last_write_page_buffer = get_last_write_pipe_buffer(out_pipe);
+    unsigned int out_pipe_last_buffer_flags = READ_KERN(last_write_page_buffer->flags);
+    // This is not a dirty-pipe splice if CAN_MERGE flag is off.
+    if ((out_pipe_last_buffer_flags & PIPE_BUF_FLAG_CAN_MERGE) == 0) {
+        return 0;
+    }
+
+    u64 out_inode_number = READ_KERN(out_inode->i_ino);
+    unsigned int out_pipe_last_buffer_next_data_offset = READ_KERN(last_write_page_buffer->offset) + READ_KERN(last_write_page_buffer->len);
```
rafaeldtinoco, line 4546: I assume you're informing last buffer `offset + len` here so you know what data was exposed (or could have been) during the pipe buf merge attempt (dividing file contents / page_size and moving cursor to `offset + len`).

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4455,6 +4506,83 @@ int BPF_KPROBE(trace_security_inode_mknod)
...
+    // In kernel v5.10 the pointer passed was no longer of the user, so flexibility is needed to read it
+    loff_t off_in;
+    #ifndef CORE
+    #if LINUX_VERSION_CODE < KERNEL_VERSION(5, 10, 0)
+    off_in = READ_USER(*off_in_addr);
+    #else
+    off_in = READ_KERN(*off_in_addr);
+    #endif
+    #else // CORE
+    struct public_key_signature *check;
```
rafaeldtinoco, line 4565: ok, this DEFINITELY needs more explanation (saying its a mitigation for checking kernel versions based on diff types) =D or else unadvertised readers might be totally crazy =D.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4455,6 +4506,83 @@ int BPF_KPROBE(trace_security_inode_mknod)
...
+
+    save_to_submit_buf(&data, &in_inode_number, sizeof(u64), 0);
+    save_to_submit_buf(&data, &in_file_type, sizeof(unsigned short), 1);
+    save_str_to_buf(&data, in_file_path, 2);
+    save_to_submit_buf(&data, &off_in, sizeof(loff_t), 3);
+    save_to_submit_buf(&data, &write_len, sizeof(size_t), 4);
+    save_to_submit_buf(&data, &out_inode_number, sizeof(u64), 5);
+    save_to_submit_buf(&data, &out_pipe_last_buffer_flags, sizeof(unsigned int), 6);
+    save_to_submit_buf(&data, &out_pipe_last_buffer_next_data_offset, sizeof(unsigned int), 7);
+    return events_perf_submit(&data, DIRTY_PIPE_SPLICE, 0);
```
rafaeldtinoco, line 4582: nit, jump a line to return, remove extra line break.

rafaeldtinoco (changes requested):
Okay, code looks very good and congrats for the dirty work of finding all corner cases for CO-RE and things alike.

I have requested some changes and, tbh, I think we should start considering moving the eBPF business logic (of CVEs, like this one) to a new file. I have discussed this with Yaniv and I can work after this is merged.

I think we're close to merge this. I haven't gone through non CO-RE OR the old kernels type checks/case (can do it as a last step).

Thanks!

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4455,6 +4506,83 @@ int BPF_KPROBE(trace_security_inode_mknod)
...
+    // dirty_pipe_splice is a splice to a pipe which results that the last page copied could be modified (the
+    // PIPE_BUF_CAN_MERGE flag is on in the pipe_buffer struct).
+    if (!event_chosen(DIRTY_PIPE_SPLICE))
+        return 0;
+
+    struct file *out_file = (struct file*)saved_args.args[2];
+    struct inode *out_inode = READ_KERN(out_file->f_inode);
+    struct pipe_inode_info *out_pipe = READ_KERN(out_inode->i_pipe);
+    // Check that output is a pipe
+    if (!out_pipe) {
```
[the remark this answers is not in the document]
AlonZivony: You are very right.
It crossed my mind when writing it but I forgot to test if it might prove to be problematic. There is a more hermetic way to do this check, but I was a bit lazy to implement it (it requires to use kallsyms which wasn't available until now).
Anyways, I will implement the other way (as implemented by the [get_pipe_info](https://elixir.bootlin.com/linux/latest/source/fs/pipe.c#L1367) function).

pkg/ebpf/c/tracee.bpf.c
AlonZivony: I will think of a better name to describe it, so it won't be hard to understand why I calculate this value.

AlonZivony:
I changed the output arguments to be more relevant to the vulnerability.
No major change except the change of the check if the file is a pipe or not.

rafaeldtinoco (changes requested):
@AlonZivony,

I have rebased your PR and added a PR to https://github.com/AlonZivony/tracee/pull/1, AND included another commit on top (splitting definitions like they already were. Ill improve this code definition later, just wanted to be the same as we had for other code).

Could you pick the LAST 2 commits in that PR, force push into this PR and re-ask for my review ? Then I think we can merge it.

OR, even better, you can close this PR and approve https://github.com/aquasecurity/tracee/pull/1647 directly so I can merge that one once you test it.

Note: Commits will have same credits. Cheers!

Thank you!

rafaeldtinoco:
@AlonZivony NM, I just merged the other PR to make things easier (for other PRs needing rebase as well).

Issue #1582, opened 2022-03-24
catch sequence operations hooks 


itamarmaouda101:
Hi 

Currently, I'm working to add an event that will detect hooks of sequence operations under /proc/net directory.

That event will give Tracee the ability to verify that there isn't hooks under those sequence operations.

The reason for that detection is  that sometimes malware hooks tcp4_seq_show in order to hide TCP connection

PR #1583, opened 2022-03-24
adding detection for sequence operations hooks

itamarmaouda101:
Hi 
This PR solves #1582 , by adding two events. 

The first event fetch addresses of the network interfaces sequence operations functions, and the second one is higher-level event that used to analyze if those function address are hooked


AsafEitani:
#1730 replaces this PR

PR #1589, opened 2022-03-25
tracee-ebpf: debug protect `--metrics` stdout prints

### LINKED ISSUE #1581: tracee-rules fails when piping `metrics` enabled tracee-ebpf through stdin
First i'll note this won't affect the docker images or tests since we're using a separate named pipe in these cases.

In short the bug is reproduced as such:
```
tracee-ebpf --metrics --output gob --output option:parse-arguments | tracee-rules --input-tracee=format:gob --input-tracee file:stdin
2022/03/24 13:02:27 error while decoding event: gob: duplicate type received
2022/03/24 13:02:27 error while decoding event: gob: duplicate type received
2022/03/24 13:02:27 error while decoding event: gob: unknown type id or corrupted data
2022/03/24 13:02:27 error while decoding event: gob: unknown type id or corrupted data
2022/03/24 13:02:27 error while decoding event: gob: type mismatch in decoder: want struct type trace.Event; got non-struct
```

With json format this isn't crashing but a similar error occurs, I guess this is because the gob module panics on some errors. I'm not sure what's the best way to solve this and if we should at all, but we should at least document it in case someone has this problem.

NDStrahilevitz:
This PR:
* Document the potential breaking behavior if enabling the debug flag
* Debug protects the `--metrics` init flag print

Close #1581

### DIFF
--- a/cmd/tracee-ebpf/main.go
+++ b/cmd/tracee-ebpf/main.go
@@ -190,7 +190,9 @@ func main() {
 					mux.Handle("/metrics", promhttp.Handler())
 
 					go func() {
-						fmt.Fprintf(os.Stdout, "Serving metrics endpoint at %s\n", metricsAddr)
+						if debug {
+							fmt.Fprintf(os.Stdout, "Serving metrics endpoint at %s\n", metricsAddr)
+						}
 						if err := http.ListenAndServe(metricsAddr, mux); err != http.ErrServerClosed {
 							fmt.Fprintf(os.Stderr, "Error serving metrics endpoint: %v\n", err)
 						}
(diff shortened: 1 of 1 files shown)


cmd/tracee-ebpf/main.go
```diff
@@ -313,7 +315,7 @@ func main() {
 			&cli.BoolFlag{
 				Name:        "debug",
 				Value:       false,
-				Usage:       "write verbose debug messages to standard output and retain intermediate artifacts",
+				Usage:       "write verbose debug messages to standard output and retain intermediate artifacts. this may break the event stream if --output:out-file is set to stdin, so a different file should be provided",
```
itaysk, line 318: I think you meant stdout, but anyway I don't think this is accurate.. the promise for the output of tracee-ebpf to be just events holds regardless if the output is sent to stdout or to a file. I would suggest: `enabling debug printing will pollute the output with informative messages which will likely to break consumers that expected the output to be machine-readable event stream`

NDStrahilevitz: Your version misses the fact that this is only relevant if both events and debug messages go to stdout (if the streams are separate there will be no problem), but it does reflect the message more clearly, So I've edited to: `enabling will output debug messages to stdout, which will likely break consumers which expect to receive machine-readable events from stdout` WDYT?

itaysk: Ah I see now what you meant initially. I didn't realize we were println-ing straight away. This wording sgtm

Issue #1590, opened 2022-03-26
error occurs while building release snapshot


grantseltzer:
There's an error when building that the directory `tracee/dist/libbpf/obj`does not exist. This happens when the libbpf.a static file is being built. It works the second time it's run because at that point the file is actually created. 

It appears to happen because the libbpf makefile tries to mkdir the passed argument "OBJDIR" which we pass here: https://github.com/aquasecurity/tracee/blob/main/Makefile#L308

```
...
Successfully built 393e75e4ad77
Successfully tagged ubuntu-tracee-make:latest
docker \
	run --rm --pid=host --privileged -v /etc/os-release:/etc/os-release-host:ro -v /home/rotscale/go/src/github.com/aquasecurity/tracee:/tracee -v /lib/modules:/lib/modules:ro -v /usr/src:/usr/src:ro -v /sys/kernel/security:/sys/kernel/security:ro -e LIBBPFGO_OSRELEASE_FILE=/etc/os-release-host -e STATIC=1 -e BTFHUB=1 \
	ubuntu-tracee-make \
	make all
CC="clang" \
	CFLAGS=""-fPIC"" \
	LD_FLAGS="" \
	make -j8 \
	-C ./3rdparty/libbpf/src \
	BUILD_STATIC_ONLY=1 \
	DESTDIR=/tracee/dist/libbpf \
	OBJDIR=/tracee/dist/libbpf/obj \
	INCLUDEDIR= LIBDIR= UAPIDIR= prefix= libdir= \
	install install_uapi_headers
  MKDIR    /tracee/dist/libbpf/obj/staticobjs
/bin/sh: 1: cannot create /tracee/dist/libbpf/obj/libbpf.pc: Directory nonexistent
make[1]: *** [Makefile:103: /tracee/dist/libbpf/obj/libbpf.pc] Error 2
make[1]: *** Waiting for unfinished jobs....
  INSTALL  bpf.h libbpf.h btf.h libbpf_common.h libbpf_legacy.h xsk.h bpf_helpers.h bpf_helper_defs.h bpf_tracing.h bpf_endian.h bpf_core_read.h skel_internal.h libbpf_version.h
  INSTALL  ../include/uapi/linux/bpf.h ../include/uapi/linux/bpf_common.h ../include/uapi/linux/btf.h
make: *** [Makefile:301: dist/libbpf/libbpf.a] Error 2
rm .check_pkg-config
[...]

PR #1591, opened 2022-03-26
Remove OBJDIR argument

grantseltzer:
This should handle #1590 


grantseltzer:
Just removing the OBJDIR argument isn't going to work either, would need to update usage of it, or figure out why the directory is not created before the make target is executed. It's difficult to debug as it's a makefile running another makefile which starts a docker container and runs a makefile in that.

PR #1594, opened 2022-03-27
Create directory that causes issue for not existing

grantseltzer:
This solves https://github.com/aquasecurity/tracee/issues/1590


rafaeldtinoco:
Please see https://github.com/aquasecurity/tracee/issues/1590#issuecomment-1080434345. I don't like changing this without having a reproducer and understanding why it fails for you.

grantseltzer:
@rafaeldtinoco  `make -f builder/Makefile.release snapshot`, strangely it only happens when building this way. I've been unable to pin down the exact cause.

rafaeldtinoco:
> @rafaeldtinoco `make -f builder/Makefile.release snapshot`, strangely it only happens when building this way. I've been unable to pin down the exact cause.

Alright, was able to reproduce:

```
Successfully tagged ubuntu-tracee-make:latest
docker \
	run --rm --pid=host --privileged -v /etc/os-release:/etc/os-release-host:ro -v /home/rafaeldtinoco/temp/tracee:/tracee -v /lib/modules:/lib/modules:ro -v /usr/src:/usr/src:ro -v /sys/kernel/security:/sys/kernel/security:ro -e LIBBPFGO_OSRELEASE_FILE=/etc/os-release-host -e STATIC=1 -e BTFHUB=1 \
	ubuntu-tracee-make \
	make all
CC="clang" \
	CFLAGS=""-fPIC"" \
	LD_FLAGS="" \
	make -j24 \
	-C ./3rdparty/libbpf/src \
	BUILD_STATIC_ONLY=1 \
	DESTDIR=/tracee/dist/libbpf \
	OBJDIR=/tracee/dist/libbpf/obj \
	INCLUDEDIR= LIBDIR= UAPIDIR= prefix= libdir= \
	install install_uapi_headers
  MKDIR    /tracee/dist/libbpf/obj/staticobjs
/bin/sh: 1: cannot create /tracee/dist/libbpf/obj/libbpf.pc: Directory nonexistent
make[1]: *** [Makefile:103: /tracee/dist/libbpf/obj/libbpf.pc] Error 2
make[1]: *** Waiting for unfinished jobs....
  INSTALL  bpf.h libbpf.h btf.h libbpf_common.h libbpf_legacy.h xsk.h bpf_helpers.h bpf_helper_defs.h bpf_tracing.h bpf_endian.h bpf_core_read.h skel_internal.h libbpf_version.h
  INSTALL  ../include/uapi/linux/bpf.h ../include/uapi/linux/bpf_common.h ../include/uapi/linux/btf.h
rm .check_pkg-config
make: *** [Makefile:301: dist/libbpf/libbpf.a] Error 2
make[1]: *** [builder/Makefile.tracee-make:189: ubuntu-make] Error 2
tar: ./dist/tracee-ebpf: Cannot stat: No such file or directory
tar: ./dist/tracee-rules: Cannot stat: No such file or directory
tar: ./dist/rules: Cannot stat: No such file or directory
tar: ./dist/tracee.bpf.core.o: Cannot stat: No such file or directory
tar: Exiting with failure status due to previous errors
docker build \
	-f builder/Dockerfile.alpine-tracee-container \
	-t tracee:latest \
	--build-arg=BTFHUB=1 \
	--build-arg=FLAVOR=tracee-core \
	--target tracee-core \
```

Let's see..

danielpacak:
I cannot reproduce in Vagrant machine as well and it seems to be working fine in the actual release workflow, isn't it? https://github.com/aquasecurity/tracee/actions/runs/2017607127

rafaeldtinoco:
> I cannot reproduce in Vagrant machine as well and it seems to be working fine in the actual release workflow, isn't it? https://github.com/aquasecurity/tracee/actions/runs/2017607127

Because we do a regular compilation first, and it creates the directory... then a call to the makefile release won't fail (as the libbpf directory is already built for the first time).

Check my previous comment... a `make -C 3rdparty/libbpf/src clean` and `make -f builder/Makefile.release snapshot` will cause the error. Removing "dist" from dockerfile ignore file will mitigate the error.

PR #1596, opened 2022-03-27
ebpf: fix string filter

### LINKED ISSUE #1595: String filters might not always work
In some cases, string filters (e.g. comm=xxx) might not work as expected.
The reason for this seems to be related to the fact that we don't explicitly declare the string size, and just give a pointer to the go string struct. By doing this, there might be extra garbage bytes, as the bpf map expects a fixed string size of 16.

yanivagman:
Fix #1595

### DIFF
--- a/cmd/tracee-ebpf/internal/flags/flags-filter.go
+++ b/cmd/tracee-ebpf/internal/flags/flags-filter.go
@@ -7,6 +7,9 @@ import (
 	tracee "github.com/aquasecurity/tracee/pkg/ebpf"
 )
 
+// MaxBpfStrFilterSize value should match MAX_STR_FILTER_SIZE defined in BPF code
+const MaxBpfStrFilterSize = 16
+
 func FilterHelp() string {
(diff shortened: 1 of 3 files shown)


cmd/tracee-ebpf/internal/flags/flags-filter.go
```diff
@@ -7,6 +7,9 @@ import (
 	tracee "github.com/aquasecurity/tracee/pkg/ebpf"
 )
 
+// MaxBpfStrFilterSize value should match MAX_STR_FILTER_SIZE defined in BPF code
+const MaxBpfStrFilterSize = 16
```
mtcherni95, line 11: what's the unite of measure? I guess bytes. I suggest to add it to the var name.

yanivagman: In this case, it's 16 bytes, but also 16 chars

rafaeldtinoco (approved):
LGTM

PR #1598, opened 2022-03-27
Refactor event derivation mechanism

NDStrahilevitz:
This PR refactors the `deriveEvent` function and mechanism so that it is:

1. More declarative in syntax
2. Takes less cpu time (reduced by 3.91% of overall time in code path)

# The changes

The `Tracee` object now initializes a map on startup of type `map[int32]map[int32]deriveFn`, this represents a map of a derived EventID to its possible derivations and their corresponding deriving functions.

Then `deriveEvents` can simply check if the current event has any derivations to be done and get the map of all it's derivative functions, instead of calculating that map on each `deriveEvents` call which causes unnecessary GC work due to the map and function variable assignments.

This improves CPU usage in runtime and allows a more declarative approach to defining event derivatives through the map initialization function (`initEventDerivationMap` in `events_derived.go`)

# Performance analysis

In terms of benchmarking I did not see any regression in throughput.
I have generated pprofs of before and after to check if the bottleneck was removed, the graph output shows this:

image

As can be seen we have reduced time in mallocgc through a reduction from `deriveEvents`. As expected we have gained some time back in `mapaccess` however it's easily seen that the difference in time is negative (a gain in cpu time).
I have not seen any new bottleneck arising from this, and I have attached the pprofs so this can be confirmed by others.

pprofs: 
1. [before](https://github.com/aquasecurity/tracee/files/8357725/070.pb.gz)
2. [after](https://github.com/aquasecurity/tracee/files/8357728/refactor.pb.gz)

A diff pprof can be generated as such:
`go tool pprof -http=":4441" -diff_base 070.pb.gz refactor.pb.gz`

### DIFF
--- a/pkg/ebpf/events_derived.go
+++ b/pkg/ebpf/events_derived.go
@@ -6,87 +6,113 @@ import (
 	"github.com/aquasecurity/tracee/types/trace"
 )
 
-type deriveFn func(EventDefinition) error
+// deriveFn is a function prototype for a function that receives an event as
+// argument and may produce a new event if relevant.
+// It returns the a derived or empty event, depending on succesful derivation,
+// a bool indicating if an event was derived, and an error if one occured.
+type deriveFn func(trace.Event) (trace.Event, bool, error)
 
-func (t *Tracee) deriveEvent(event trace.Event) []trace.Event {
-	var deriveFns map[int32]deriveFn
-	var derivatives []trace.Event
-
-	eventID := int32(event.EventID)
-	switch eventID {
-	case CgroupMkdirEventID:
-		containerCreateFn := func(def EventDefinition) error {
-			cgroupId, err := getEventArgUint64Val(&event, "cgroup_id")
-			if err != nil {
-				return err
-			}
-
-			if info := t.containers.GetCgroupInfo(cgroupId); info.ContainerId != "" {
-				de := event
-				de.EventID = int(ContainerCreateEventID)
-				de.EventName = "container_create"
-				de.ReturnValue = 0
-				de.StackAddresses = make([]uint64, 1)
-				de.Args = []trace.Argument{
-					{ArgMeta: def.Params[0], Value: info.Runtime},
-					{ArgMeta: def.Params[1], Value: info.ContainerId},
-					{ArgMeta: def.Params[2], Value: info.Ctime.UnixNano()},
-				}
-				de.ArgsNum = 3
-
-				derivatives = append(derivatives, de)
-			}
-
-			return nil
-		}
-		deriveFns = map[int32]deriveFn{
-			ContainerCreateEventID: containerCreateFn,
-		}
-	case CgroupRmdirEventID:
-		containerRemoveFn := func(def EventDefinition) error {
-			cgroupId, err := getEventArgUint64Val(&event, "cgroup_id")
-			if err != nil {
-				return err
-			}
-
-			if info := t.containers.GetCgroupInfo(cgroupId); info.ContainerId != "" {
-				de := event
-				de.EventID = int(ContainerRemoveEventID)
-				de.EventName = "container_remove"
-				de.ReturnValue = 0
-				de.StackAddresses = make([]uint64, 1)
-				de.Args = []trace.Argument{
-					{ArgMeta: def.Params[0], Value: info.Runtime},
-					{ArgMeta: def.Params[1], Value: info.ContainerId},
-				}
-				de.ArgsNum = 2
-
-				derivatives = append(derivatives, de)
-			}
-
-			return nil
-		}
-		deriveFns = map[int32]deriveFn{
-			ContainerRemoveEventID: containerRemoveFn,
-		}
+// Initialize the eventDerivations map.
+// Here we declare for each Event (represented through it's ID)
+// to which other Events it can be derived and the corresponding function to derive into that Event.
+func (t *Tracee) initEventDerivationMap() error {
+	t.eventDerivations = map[int32]map[int32]deriveFn{
+		CgroupMkdirEventID: {
+			ContainerCreateEventID: deriveContainerCreate(t),
+		},
+		CgroupRmdirEventID: {
+			ContainerRemoveEventID: deriveContainerRemoved(t),
+		},
 	}
 
+	return nil
+}
+
+// deriveEvent takes a trace.Event and checks if it can derive additional events from it
+// as defined by tracee's eventDerivations map.
+// The map is initialized in the above function
+func (t *Tracee) deriveEvent(event trace.Event) []trace.Event {
+	derivatives := []trace.Event{}
+	deriveFns := t.eventDerivations[int32(event.EventID)]
 	for id, deriveFn := range deriveFns {
 		// Don't derive events which were not requested by the user
 		if !t.eventsToTrace[id] {
 			continue
 		}
 
-		def, ok := EventsDefinitions[id]
-		if !ok {
-			t.handleError(fmt.Errorf("failed to get configuration of event %d", id))
-			continue
-		}
-
-		if err := deriveFn(def); err != nil {
+		derivative, derived, err := deriveFn(event)
+		if err != nil {
 			t.handleError(fmt.Errorf("failed to derive event %d: %v", id, err))
+		} else if derived {
+			derivatives = append(derivatives, derivative)
 		}
 	}
 
 	return derivatives
+}
+
+/*
+* Derivation functions:
+* Most derivation functions take tracee as a closure argument to track it's runtime state
+* Tracee builds it's derivation map from these functions and injects itself as an argument to the closures
+* The derivation map is then built with the returned deriveFn functions, which is used in deriveEvents
+ */
+
+//Receives a tracee object as a closure argument to track it's containers
+//If it receives a cgroup_mkdir event, it can derive a container_create event from it
+func deriveContainerCreate(t *Tracee) deriveFn {
+	return func(event trace.Event) (trace.Event, bool, error) {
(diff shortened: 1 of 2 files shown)


pkg/ebpf/derivations.go
```diff
@@ -0,0 +1,61 @@
+package ebpf
+
+import (
+	"github.com/aquasecurity/tracee/types/trace"
+)
+
+//returns the derived event, wether an event was derived, and a relevant error if failure was due to one
```
mtcherni95, line 7: typo in `whether`

pkg/ebpf/derivations.go
```diff
@@ -0,0 +1,61 @@
...
+type deriveFn func(trace.Event, EventDefinition) (trace.Event, bool, error)
+
+func deriveContainerCreate(t *Tracee) deriveFn {
```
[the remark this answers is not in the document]
NDStrahilevitz: I see why we would want that but that can easily cause circular dependencies since most derivation function actually depend on it's tracee's object current state, so it's difficult to separate.
I'll see if it's possible since it would be better if we can do it.

pkg/ebpf/derivations.go
```diff
@@ -0,0 +1,61 @@
...
+	return func(event trace.Event, def EventDefinition) (trace.Event, bool, error) {
```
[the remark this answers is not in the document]
NDStrahilevitz: Like a `Derivation` struct? I can't think of a case where we wouldn't immediately want all 3 values so having them returned unpacked is convenient.

pkg/ebpf/derivations.go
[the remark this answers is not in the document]
NDStrahilevitz: Actually, these aren't exported so there's not much of reason to be worried about usage outside of `ebpf`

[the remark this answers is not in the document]
mtcherni95: it is still more correct to have it under internal, you do not what anyways to have them used outside (no one should expose them)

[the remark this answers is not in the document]
NDStrahilevitz: Resolved offline, this will be done in a later PR to keep it out of this one's scope.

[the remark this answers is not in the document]
yanivagman: I would wait with putting it under "internal" of the ebpf package.
The ebpf package will eventually handle only ebpf related stuff - bpf maps population, program loading/attachment, bpf filters and perf buffer polling and decoding.
In the future, we can potentially split some of this logic to a new "events" package, and there we can have such "internal" packages

pkg/ebpf/derivations.go
```diff
@@ -0,0 +1,72 @@
+package ebpf
```
yanivagman, line 1: I don't see a reason to split these functions to a new file.
It is short enough, and will be easier to read if we will keep it in the same file

pkg/ebpf/tracee.go
```diff
@@ -18,6 +17,8 @@ import (
 	"time"
 	"unsafe"
 
+	"github.com/aquasecurity/tracee/pkg/events/queue"
```
yanivagman, line 20: If you move this, let's put it below just before "github.com/aquasecurity/tracee/pkg/events/sorting"

pkg/ebpf/events_derived.go
```diff
@@ -6,85 +6,33 @@ import (
...
+	t.eventDerivations = map[int32]map[int32]deriveFn{
+		CgroupMkdirEventID: {
+			ContainerCreateEventID: deriveContainerCreate(t),
+		},
+		CgroupRmdirEventID: {
+			ContainerRemoveEventID: deriveContainerRemoved(t),
+		},
 	}
 
+	return nil
```
yanivagman, line 19: If you already made this step to initialize these functions here, you can also save some other runtime steps by adding the derived event "template", which will contain the event id, event name, return value, stack addresses, args, and args num of the event. Then, when deriving the event in runtime, there will be no need to get the event definition and set these fields again

pkg/ebpf/tracee.go
NDStrahilevitz: I think this moved due to automatic formatting, I didn't touch it myself.

pkg/ebpf/derivations.go
yanivagman: #1561 and #1583 shouldn't be derived events. I still didn't have the time to review these PRs yet, but from a quick look, the logic there should be moved to process_events.
Even if we will have more derived events in the future, the deriveEvent function is the only generic function there, and all the other deriveFns are the specific ones. By placing the generic deriveEvent function first in the file, I think this should be easier to read and maintain

pkg/ebpf/tracee.go
NDStrahilevitz: done

pkg/ebpf/derivations.go
NDStrahilevitz: Done

pkg/ebpf/events_derived.go
yanivagman: Indeed this is for optimizations purposes.
If you approve that there is no much of performance gain by doing that, then I'm ok with keeping it that way

NDStrahilevitz: There might be a performance gain only if we had an event that is VERY frequently derived (this might be workload dependent as well), and even so a map access is a pretty cheap operation (notice that we access the derivation map for every event and it takes only 0.5% cpu, it is a small map however), so I think this can be considered a premature optimization.

yanivagman (approved):
Nice refactor, LGTM!

rafaeldtinoco:
I'll be taking a look at this and merging (looks good so far). Commits should be squashed and git log should have proper line breaks.. I can do it after review.

pkg/ebpf/events_derived.go
```diff
@@ -6,87 +6,102 @@ import (
...
-				return err
-			}
+func (t *Tracee) initEventDerivationMap() error {
+	t.eventDerivations = map[int32]map[int32]deriveFn{
+		CgroupMkdirEventID: {
+			ContainerCreateEventID: deriveContainerCreate(t),
+		},
+		CgroupRmdirEventID: {
+			ContainerRemoveEventID: deriveContainerRemoved(t),
+		},
```
rafaeldtinoco, line 21: Single liners here are easier to read/understand:

```go
	t.eventDerivations = map[int32]map[int32]deriveFn{
		CgroupMkdirEventID: {ContainerCreateEventID: deriveContainerCreate(t)},
		CgroupRmdirEventID: {ContainerRemoveEventID: deriveContainerRemoved(t)},
	}
```

pkg/ebpf/events_derived.go
```diff
@@ -6,87 +6,102 @@ import (
 	"github.com/aquasecurity/tracee/types/trace"
 )
 
-type deriveFn func(EventDefinition) error
+//deriveFn represents a function which takes an event and derives whether another event should result from it
+//
+//returns a derived (or empty) event, whether an event was derived, and an error if relevant
```
rafaeldtinoco, line 11: You could improve this comment, it is not very clear. Suggestion:

```go
// deriveFn is a function prototype for a function that receives an event as
// argument and produces a new event, depending on the event type received. It
// returns the derived event, a bool indicating if there was indeed a derived
// event created, and an error/nil.
```

pkg/ebpf/events_derived.go
```diff
@@ -6,87 +6,102 @@ import (
...
-
-				derivatives = append(derivatives, de)
-			}
+	return derivatives
+}
 
-			return nil
+/*
+* Derivation functions:
+* Each derivation function initially takes it's runtime dependencies (usually tracee to track its state)
```
rafaeldtinoco, line 49: Could you clarify what you meant here ?

pkg/ebpf/events_derived.go
```diff
@@ -6,87 +6,102 @@ import (
...
+* It then returns the actual deriveFn which is stored in tracee's derivation map (where it initializes the closure with its own state)
+* From the derivation map we call the relevant scoped deriveFn
```
rafaeldtinoco, line 51: And we're limited to a single derived function per event type (which is okay as we're currently using one). I wonder if this comment is needed or it would be better to add description to each derive function AND `deriveEvent()` one, with a generic explanation.

rafaeldtinoco (changes requested):
Change looks very good and, tbh, I think there might be other places where we could benefit of same technique (not to create/re-create function definitions over and over.

I'd like comments to be improved (gave simple suggestion) and each derive function to have its own description (what it is doing and why). This way next derived functions will follow the same guideline.

pkg/ebpf/events_derived.go
NDStrahilevitz: I've rewritten to this:

```
// deriveFn is a function prototype for a function that receives an event as
// argument and may produce a new event if relevant.
// It returns the a derived or empty event, depending on succesful derivation,
// a bool indicating if an event was derived, and an error if one occured.
```

NDStrahilevitz:
> Change looks very good and, tbh, I think there might be other places where we could benefit of same technique (not to create/re-create function definitions over and over.
> 
> I'd like comments to be improved (gave simple suggestion) and each derive function to have its own description (what it is doing and why). This way next derived functions will follow the same guideline.

Thanks for the feedback, I've squashed the commits, added comments and rewritten the others, take a look when you have the time :)

pkg/ebpf/events_derived.go
rafaeldtinoco: I see your point. Kind of makes sense. Thanks.

rafaeldtinoco (approved):
Thanks for the changes. LGTM now.

PR #1601, opened 2022-03-28
trace: add container and pod data to event

NDStrahilevitz:
This is in preparation for #1572

Also a good opportunity to discuss the new enrichment fields and if more are required for initial implementation.

### DIFF
--- a/types/trace/trace.go
+++ b/types/trace/trace.go
@@ -26,6 +26,11 @@ type Event struct {
 	ProcessName         string     `json:"processName"`
 	HostName            string     `json:"hostName"`
 	ContainerID         string     `json:"containerId"`
+	ContainerImage      string     `json:"containerImage"`
+	ContainerName       string     `json:"containerName"`
+	PodName             string     `json:"podName"`
+	PodNamespace        string     `json:"podNamespace"`
+	PodUID              string     `json:"podUID"`
 	EventID             int        `json:"eventId,string"`
 	EventName           string     `json:"eventName"`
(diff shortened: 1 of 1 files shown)


rafaeldtinoco (commented):
I haven't replied yet because I believe there was a meaningful reason for this struct to be flat (regarding possible padding and the fact this data comes from perf buffer shared memory).

@yanivagman could you confirm this thought here ?

yanivagman:
> I haven't replied yet because I believe there was a meaningful reason for this struct to be flat (regarding possible padding and the fact this data comes from perf buffer shared memory).
> 
> @yanivagman could you confirm this thought here ?

Actually we now have the bufferdecoder that takes care of reading the bytes from the buffer (using the protocol's Context struct in that case), and this struct is what tracee-ebpf generates as an output.

LGTM!

yanivagman (approved):

NDStrahilevitz:
I've flattened the struct to simplify usage later down the line (anonymous structs don't play nicely in go sometimes). We can merge this now @yanivagman

yanivagman:
> I've flattened the struct to simplify usage later down the line (anonymous structs don't play nicely in go sometimes). We can merge this now @yanivagman

Ok, let's wait for @rafaeldtinoco to approve and merge as he requested a review on this

rafaeldtinoco:
> Ok, let's wait for @rafaeldtinoco to approve and merge as he requested a review on this

I was being pro-active only =) sorry. I was waiting for the final checks to merge it.

PR #1603, opened 2022-03-28
engine: unify signature loading logic

NDStrahilevitz:
This refactor unifies the signature loading logic.
`NewEngine` and `LoadSignature` both implemented their own internal signature map storing logic, with the difference being that `LoadSignature` would immediately start the signature and `NewEngine` wouldn't (Start does that instead).

This refactor adds a new internal `loadSignature` function which is used both by `NewEngine` and the exported `LoadSignature`.

This also solves the bug where signatures stat wasn't incremented on the initial load from `NewEngine`.

### DIFF
--- a/pkg/rules/engine/engine.go
+++ b/pkg/rules/engine/engine.go
@@ -1,7 +1,6 @@
 package engine
 
 import (
-	"errors"
 	"fmt"
 	"io"
 	"log"
@@ -63,45 +62,10 @@ func NewEngine(sigs []detect.Signature, sources EventSources, output chan detect
 	engine.signaturesIndex = make(map[detect.SignatureEventSelector][]detect.Signature)
 	engine.signaturesMutex.Unlock()
 	for _, sig := range sigs {
-		engine.signaturesMutex.Lock()
-		engine.signatures[sig] = make(chan protocol.Event, engine.config.SignatureBufferSize)
-		engine.signaturesMutex.Unlock()
-		meta, err := sig.GetMetadata()
+		_, err := engine.loadSignature(sig)
 		if err != nil {
-			engine.logger.Printf("error getting metadata: %v", err)
-			continue
+			engine.logger.Printf("error loading signature: %v", err)
 		}
-		se, err := sig.GetSelectedEvents()
-		if err != nil {
-			engine.logger.Printf("error getting selected events for signature %s: %v", meta.Name, err)
-			continue
-		}
-		for _, es := range se {
-			if es.Name == "" {
-				es.Name = ALL_EVENT_TYPES
-			}
-			if es.Origin == "" {
-				es.Origin = ALL_EVENT_ORIGINS
-			}
-			if es.Source == "" {
-				engine.logger.Printf("signature %s doesn't declare an input source", meta.Name)
-			} else {
-				engine.signaturesMutex.Lock()
-				engine.signaturesIndex[es] = append(engine.signaturesIndex[es], sig)
-				engine.signaturesMutex.Unlock()
-			}
-		}
-		err = sig.Init(engine.matchHandler)
-		if err != nil {
-			engine.logger.Printf("error initializing signature %s: %v", meta.Name, err)
-			continue
-		}
-	}
-	engine.signaturesMutex.RLock()
-	lenSigs := len(engine.signatures)
-	engine.signaturesMutex.RUnlock()
-	if len(sigs) != lenSigs {
-		return nil, errors.New("one or more signatures are not uniquely identifiable")
 	}
 	return &engine, nil
 }
@@ -221,23 +185,47 @@ func (engine *Engine) dispatchEvent(s detect.Signature, event protocol.Event) {
 	engine.signatures[s] <- event
 }
 
-//LoadSignature will store in Engine data structures the given signature and activate its handling business logics.
+//LoadSignature will call the internal signature loading logic and activate its handling business logics.
 // It will return the signature ID as well as error.
 func (engine *Engine) LoadSignature(signature detect.Signature) (string, error) {
+	id, err := engine.loadSignature(signature)
+	if err != nil {
+		return id, err
+	}
+	engine.signaturesMutex.RLock()
+	go signatureStart(signature, engine.signatures[signature], &engine.waitGroup)
+	engine.signaturesMutex.RUnlock()
+
+	return id, nil
+}
+
+//loadSignature handles storing a signature in the Engine data structures
+// It will return the signature ID as well as error.
+func (engine *Engine) loadSignature(signature detect.Signature) (string, error) {
+	metadata, err := signature.GetMetadata()
+	if err != nil {
+		return "", fmt.Errorf("error getting metadata: %w", err)
+	}
 	selectedEvents, err := signature.GetSelectedEvents()
 	if err != nil {
-		return "", fmt.Errorf("failed to store signature: %w", err)
+		return "", fmt.Errorf("error getting selected events for signature %s: %w", metadata.Name, err)
 	}
-	metadata, _ := signature.GetMetadata()
 	// insert in engine.signatures map
-	engine.signaturesMutex.Lock()
-	defer engine.signaturesMutex.Unlock()
+	engine.signaturesMutex.RLock()
 	if engine.signatures[signature] != nil {
+		engine.signaturesMutex.RUnlock()
 		// signature already exists
-		return metadata.ID, nil
+		return "", fmt.Errorf("failed to store signature: signature \"%s\" already loaded", metadata.Name)
+	}
+	engine.signaturesMutex.RUnlock()
+	if err := signature.Init(engine.matchHandler); err != nil {
+		//failed to initialize
+		return "", fmt.Errorf("error initializing signature %s: %w", metadata.Name, err)
 	}
 	c := make(chan protocol.Event, engine.config.SignatureBufferSize)
+	engine.signaturesMutex.Lock()
 	engine.signatures[signature] = c
+	engine.signaturesMutex.Unlock()
 
 	// insert in engine.signaturesIndex map
 	for _, selectedEvent := range selectedEvents {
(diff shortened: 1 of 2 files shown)


pkg/rules/engine/engine.go
```diff
@@ -96,6 +96,7 @@ func NewEngine(sigs []detect.Signature, sources EventSources, output chan detect
 			engine.logger.Printf("error initializing signature %s: %v", meta.Name, err)
 			continue
 		}
+		engine.stats.Signatures.Increment()
```
danielpacak, line 99: Nice catch! Can we have a unit test so we won't delete this line accidentally in the future?

danielpacak (changes requested):

pkg/rules/engine/engine.go
NDStrahilevitz: Thanks, but I caught myself because I'm the one who added these stats and forgot this ;)

I was actually thinking maybe we should use LoadSignature here instead, I'm not sure why there is this duplicate logic (do you know why). WDYT? We could still unit test it obviously.

NDStrahilevitz: I uploaded a refactor which includes a test for signature loading, which also tests this increment, check it out.

pkg/rules/engine/engine_test.go
```diff
@@ -512,3 +514,24 @@ func TestGetSelectedEvents(t *testing.T) {
...
+	source := EventSources{
+		Tracee: input,
+	}
+	output := make(chan detect.Finding)
+	engine, err := NewEngine(sigs, source, output, os.Stdout, Config{})
+	assert.NoError(t, err)
+
+	sig := regoFakeSignature{}
+	metadata, err := sig.GetMetadata()
+	assert.NoError(t, err)
```
danielpacak, line 529:
```suggestion
	require.NoError(t, err)
```

pkg/rules/engine/engine_test.go
```diff
@@ -512,3 +514,24 @@ func TestGetSelectedEvents(t *testing.T) {
 	}
 	assert.ElementsMatch(t, expected, se)
 }
+func TestLoadSignature(t *testing.T) {
```
danielpacak, line 517:
```suggestion
func TestEngine_LoadSignature(t *testing.T) {
```

pkg/rules/engine/engine.go
```diff
@@ -221,23 +185,47 @@ func (engine *Engine) dispatchEvent(s detect.Signature, event protocol.Event) {
...
+		return id, err
+	}
+	engine.signaturesMutex.RLock()
+	go signatureStart(signature, engine.signatures[signature], &engine.waitGroup)
+	engine.signaturesMutex.RUnlock()
+
+	return id, nil
+}
+
+//loadSignature handles storing a signature in the Engine data structures
```
danielpacak, line 202:
```suggestion
// loadSignature handles storing a signature in the Engine data structures.
```

pkg/rules/engine/engine.go
```diff
@@ -221,23 +185,47 @@ func (engine *Engine) dispatchEvent(s detect.Signature, event protocol.Event) {
...
+// It will return the signature ID as well as error.
```
danielpacak, line 203: I assume we don't return signature od and error at the same time.
```suggestion
// It will return the signature ID or error.
```

pkg/rules/engine/engine.go
```diff
@@ -63,45 +62,10 @@ func NewEngine(sigs []detect.Signature, sources EventSources, output chan detect
...
 	for _, sig := range sigs {
-		engine.signaturesMutex.Lock()
-		engine.signatures[sig] = make(chan protocol.Event, engine.config.SignatureBufferSize)
-		engine.signaturesMutex.Unlock()
-		meta, err := sig.GetMetadata()
+		_, err := engine.loadSignature(sig)
 		if err != nil {
-			engine.logger.Printf("error getting metadata: %v", err)
-			continue
+			engine.logger.Printf("%v", err)
```
danielpacak, line 67: Add prefix to give and idea what went wrong in case of error

pkg/rules/engine/engine.go
```diff
@@ -63,45 +62,10 @@ func NewEngine(sigs []detect.Signature, sources EventSources, output chan detect
...
-		if err != nil {
-			engine.logger.Printf("error initializing signature %s: %v", meta.Name, err)
-			continue
-		}
-	}
-	engine.signaturesMutex.RLock()
-	lenSigs := len(engine.signatures)
-	engine.signaturesMutex.RUnlock()
-	if len(sigs) != lenSigs {
-		return nil, errors.New("one or more signatures are not uniquely identifiable")
```
danielpacak, line 104: Is this check removed intentionally? I think it's useful to catch issue when someone accidentally used the same identifier for two different signatures.

pkg/rules/engine/engine.go
```diff
@@ -221,23 +185,47 @@ func (engine *Engine) dispatchEvent(s detect.Signature, event protocol.Event) {
...
+//LoadSignature will call the internal signature loading logic and activate its handling business logics.
 // It will return the signature ID as well as error.
 func (engine *Engine) LoadSignature(signature detect.Signature) (string, error) {
+	id, err := engine.loadSignature(signature)
+	if err != nil {
+		return id, err
+	}
+	engine.signaturesMutex.RLock()
+	go signatureStart(signature, engine.signatures[signature], &engine.waitGroup)
+	engine.signaturesMutex.RUnlock()
```
danielpacak, line 197: Why is that critical section needed here when we create goroutine?

pkg/rules/engine/engine.go
```diff
@@ -221,23 +185,47 @@ func (engine *Engine) dispatchEvent(s detect.Signature, event protocol.Event) {
...
 	selectedEvents, err := signature.GetSelectedEvents()
 	if err != nil {
-		return "", fmt.Errorf("failed to store signature: %w", err)
+		return "", fmt.Errorf("error getting selected events for signature %s: %w", metadata.Name, err)
 	}
-	metadata, _ := signature.GetMetadata()
 	// insert in engine.signatures map
-	engine.signaturesMutex.Lock()
-	defer engine.signaturesMutex.Unlock()
+	engine.signaturesMutex.RLock()
```
danielpacak, line 214: Not in scope of this PR, but I'm just wondering why haven't used sync.Map for signatures index. It would probably make this code much easier to reason about.

pkg/rules/engine/engine_test.go
```diff
@@ -512,3 +514,24 @@ func TestGetSelectedEvents(t *testing.T) {
...
+	sig := regoFakeSignature{}
+	metadata, err := sig.GetMetadata()
+	assert.NoError(t, err)
+
+	id, err := engine.LoadSignature(&sig)
+	assert.NoError(t, err)
+	assert.Equal(t, metadata.ID, id)
+
+	//check that signature stats were correctly incremented
+	assert.Equal(t, 1, int(engine.Stats().Signatures.Read()))
```
danielpacak, line 536: Are you confident with this test case? Maybe we could check that it works as expected for more than one signature as well?

danielpacak (changes requested):
Nice work! I left a few suggestions and questions to fully understand this implementation.

pkg/rules/engine/engine_test.go
```diff
@@ -512,3 +514,24 @@ func TestGetSelectedEvents(t *testing.T) {
...
 }
+func TestLoadSignature(t *testing.T) {
+	var sigs []detect.Signature
+	input := make(chan protocol.Event)
+	source := EventSources{
+		Tracee: input,
+	}
+	output := make(chan detect.Finding)
+	engine, err := NewEngine(sigs, source, output, os.Stdout, Config{})
+	assert.NoError(t, err)
```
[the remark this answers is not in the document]
NDStrahilevitz: So `require` makes it so the test stops there? If so makes sense, i'll change it around

pkg/rules/engine/engine.go
NDStrahilevitz: We do the same in `Start()` where before the loop we RLock and then RUnlock

pkg/rules/engine/engine.go
NDStrahilevitz: I did remove this on purpose, mostly because it was screwing tests, but I also think it's more correct this way since duplicate signatures shouldn't stop construction IMO, and this case will be logged due to this error in line 218:
```
// signature already exists
return "", fmt.Errorf("failed to store signature: signature already exists")
```

WDYT? Should we bring this back anyway?

pkg/rules/engine/engine.go
NDStrahilevitz: Sync.Map uses `any` as it's types though right? I think it would make code cleaner but there might be overhead in accessing the signatures that way, maybe we should attempt this with generics, if and when that's supported in `sync.Map` or implement our own sync-safe `SignatureMap`, WDYT?

NDStrahilevitz:
@danielpacak I think i've addressed all of what you raised in code or answers, WDYT?

pkg/rules/engine/engine_test.go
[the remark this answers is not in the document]
danielpacak: If you check the code it will fail the test and stop executing it further.

pkg/rules/engine/engine.go
danielpacak: Personally, I would fail fast rather than troubleshoot why one signature was overwritten accidentally. Such errors shows issues in input signatures that must be fixed anyway.

pkg/rules/engine/engine.go
danielpacak: Ah okey. My bad. I missed the second arg to the signatureStart method.

pkg/rules/engine/engine.go
NDStrahilevitz: There wouldn't be an overwrite, the signature already exists and so the load fails.

pkg/rules/engine/engine.go
danielpacak: Get started with SitnagureMap seems to me like a good idea. To see if it improves the readability of code. For Go 1.18 with generics I'm not familiar with performance implications, which might be similar to sync.Map.

pkg/rules/engine/engine_test.go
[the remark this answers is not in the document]
NDStrahilevitz: Alright, I made the change.

pkg/rules/engine/engine.go
NDStrahilevitz: Alright, can we do it in another PR though? I'll open a relevant issue if so (and maybe we should add a "refactor" label to issues?)

pkg/rules/engine/engine.go
danielpacak: Got it. Would be nice to add test scenario for such case one day.

pkg/rules/engine/engine.go
danielpacak: Like I said in my initial commit this is out of scope of this PR.

danielpacak (approved):

Issue #1613, opened 2022-03-29
Tracing file creation by kernel modules


itamarmaouda101:
Hi 

Sometimes, kernel rootkits try to commutate with the user-space.

To accomplish that they create files with special file operations that handle their requests.

To create those files from the kernel they can use one of these functions:

- [x] https://github.com/aquasecurity/tracee/pull/1614 (proc_create - to create file under /proc)
- [x] https://github.com/aquasecurity/tracee/pull/1648 (debugfs_create_file - to create file under /sys/kernel/debug)
- [ ] https://github.com/aquasecurity/tracee/pull/1649 (debugfs_create_dir - to create dir under /sys/kernel/debug)
- [ ] https://github.com/aquasecurity/tracee/pull/1661(add register_chrdev_region event )
- [ ] (add trace to create_device function)

We should add an event on those functions to alert if some kernel try to communicate with the user space. 

Here's an example that uses this function to complete a  container escape in a privileged container:
[https://xcellerator.github.io/posts/docker_escape/](https://xcellerator.github.io/posts/docker_escape/)

PR #1614, opened 2022-03-29
add proc_create event

itamarmaouda101:
Hi 

Sometimes malware try to create a secret backdoor to communicate with the user-space, which could be done by creating a file under /proc with a special file operation that handles that the malware tasks.

This PR on of 2 that will solve #1613 which is the about this problem around the entire the Linux kernel (not just under /proc).

Its added an event that trace calls of `proc_create` that communally used by malwares to deploy files under /proc in order to create a commination between the kernel space to the user space.


itamarmaouda101:
> @itamarmaouda101 could you describe further the need to have that event, if it is something you need for some other feature you are not mentioning, what are the plans to use it, etc, etc ? We will try, from now on, to be very descriptive on why we're adding changes to tracee as this is something we lacked behind so far.

Yes sure, ill update the desperation and the issue (:

rafaeldtinoco (approved):
LGTM; needed minor style fix, Ill squash after tests are good.

PR #1615, opened 2022-03-29
Substitute out 'v' in docker image tags

grantseltzer:

### DIFF
--- a/builder/Makefile.release
+++ b/builder/Makefile.release
@@ -135,7 +135,7 @@ $(OUTPUT_DIR):
 #
 
 SNAPSHOT_VERSION ?= $(shell git rev-parse HEAD)
-SNAPSHOT_VERSION := $(shell echo $(SNAPSHOT_VERSION) | sed -e "s/^v//")
+IMAGE_TAG := $(subst v,,$(SNAPSHOT_VERSION))
 
 RELEASE_FILES = LICENSE
 RELEASE_FILES += $(OUTPUT_DIR)/tracee-ebpf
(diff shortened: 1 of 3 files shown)


builder/Makefile.release
```diff
@@ -135,7 +135,7 @@ $(OUTPUT_DIR):
 #
 
 SNAPSHOT_VERSION ?= $(shell git rev-parse HEAD)
```
danielpacak, line 137: The changes to code look good. However, I realized that we have two places where we resolve Git revision and it may be confusing / hard to read. First place is GitHub Actions workflow YAML for release where we use `git describe --tags --abbrev=0` command and the default value in Makefile, i.e. `git rev-parse HEAD`.

This "interface" can become simpler if `SNAPSHOT` was a boolean arg with the default value `0`:
1. Setting `SNAPSHOT=1` would imply that we build artifacts using the current commit using `git rev-parse HEAD` as version, and we skip publishing.
2. Setting `SNAPSHOT=0` or leaving it with the default value of 0 would use `git describe--tags --abbrev=0` as version, and perform publishing

To release snapshot (in PR validation workflow or in my local computer) :

```
make -f builder/Makefile.release SNAPSHOT=1 PUSH_DOCKER_REPO=aquasec/tracee
```

To perform release (in Release workflow):

```
make -f builder/Makefile.release PUSH_DOCKER_REPO=aquasec/tracee
```

This would be similar to GoReleaser that we use for other projects:

```console
$ goreleaser release -h
Releases the current project

Usage:
  goreleaser release [flags]

Aliases:
  release, r

Flags:
      --auto-snapshot                Automatically sets --snapshot if the repo is dirty
  -f, --config string                Load configuration from file
  -h, --help                         help for release
  -p, --parallelism int              Amount tasks to run concurrently (default: number of CPUs)
      --release-footer string        Load custom release notes footer from a markdown file
      --release-footer-tmpl string   Load custom release notes footer from a templated markdown file (overrides --release-footer)
      --release-header string        Load custom release notes header from a markdown file
      --release-header-tmpl string   Load custom release notes header from a templated markdown file (overrides --release-header)
      --release-notes string         Load custom release notes from a markdown file (will skip GoReleaser changelog generation)
      --release-notes-tmpl string    Load custom release notes from a templated markdown file (overrides --release-notes)
      --rm-dist                      Removes the dist folder
      --skip-announce                Skips announcing releases (implies --skip-validate)
      --skip-publish                 Skips publishing artifacts
      --skip-sbom                    Skips cataloging artifacts
      --skip-sign                    Skips signing artifacts
      --skip-validate                Skips git checks
      --snapshot                     Generate an unversioned snapshot release, skipping all validations and without publishing any artifacts (implies --skip-publish, --skip-announce and --skip-validate)

grantseltzer: I like this approach, i've pushed these changes.

danielpacak (approved):
🚀

Issue #623, opened 2021-03-16
Add race detector check in make test targets, and in github actions


grantseltzer:
All tests should run with Go's builtin race detector, same with selftests in libbpfgo

yanivagman:
related: #756

Issue #756, opened 2021-06-08
Data race while tracing events


simar7:
Building tracee with the race detector and running it as so:

```
$ sudo ./dist/tracee-ebpf

WARNING: DATA RACE
Read at 0x00c0002de098 by main goroutine:
  github.com/aquasecurity/tracee/tracee-ebpf/tracee.(*Tracee).Run()
      /home/vagrant/tracee/tracee-ebpf/tracee/tracee.go:902 +0x2e4
  main.main.func1()
      /home/vagrant/tracee/tracee-ebpf/main.go:83 +0x7be
  github.com/urfave/cli/v2.(*App).RunContext()
      /home/vagrant/go/pkg/mod/github.com/urfave/cli/v2@v2.1.1/app.go:311 +0xabe
  github.com/urfave/cli/v2.(*App).Run()
      /home/vagrant/go/pkg/mod/github.com/urfave/cli/v2@v2.1.1/app.go:211 +0x1434
  main.main()
      /home/vagrant/tracee/tracee-ebpf/main.go:147 +0x13ad

Previous write at 0x00c0002de098 by goroutine 18:
  sync/atomic.AddInt32()
      /usr/local/go/src/runtime/race_amd64.s:292 +0xb
  github.com/aquasecurity/tracee/tracee-ebpf/tracee.(*counter).Increment()
      /home/vagrant/tracee/tracee-ebpf/tracee/tracee.go:220 +0x96
  github.com/aquasecurity/tracee/tracee-ebpf/tracee.(*Tracee).printEvent.func1()
      /home/vagrant/tracee/tracee-ebpf/tracee/pipeline.go:203 +0x144

Goroutine 18 (running) created at:
  github.com/aquasecurity/tracee/tracee-ebpf/tracee.(*Tracee).printEvent()
      /home/vagrant/tracee/tracee-ebpf/tracee/pipeline.go:200 +0x7d
  github.com/aquasecurity/tracee/tracee-ebpf/tracee.(*Tracee).runEventPipeline()
      /home/vagrant/tracee/tracee-ebpf/tracee/pipeline.go:35 +0x2c4
==================
==================
WARNING: DATA RACE
Read at 0x00c0002de0a0 by main goroutine:
  github.com/aquasecurity/tracee/tracee-ebpf/tracee.(*Tracee).Run()
      /home/vagrant/tracee/tracee-ebpf/tracee/tracee.go:902 +0x2e4
  main.main.func1()
[...]

PR #1627, opened 2022-04-01
test: run unit tests with -race flag

danielpacak:
Adds quality gate to test that code and tests are written
without race conditions. This allows catching issues similar
to #1629 and #1630.

Resolves: #623

### DIFF
--- a/pkg/rules/engine/engine_test.go
+++ b/pkg/rules/engine/engine_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"os"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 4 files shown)


yanivagman:
related: #623, #756

pkg/rules/engine/engine_test.go
```diff
@@ -71,38 +72,6 @@ func TestConsumeSources(t *testing.T) {
 		expectedEvent     interface{}
 		config            Config
 	}{
-		{
-			name: "happy path - with one matching selector, parsed event enabled",
```
danielpacak, line 75: We don't use parsed events anymore, and it seems that this case is the same as "happy path - with one matching selector"

rafaeldtinoco (approved):
LGTM but it is still in draft @danielpacak. How would you like to move forward ?

Issue #1630, opened 2022-04-04
fix data race in TestEventsChronologicalSorter_Start


danielpacak:
```console
$ go test -v -race -run="TestEventsChronologicalSorter_Start" ./pkg/events/sorting/...
=== RUN   TestEventsChronologicalSorter_Start
=== RUN   TestEventsChronologicalSorter_Start/Sorting_chronological_order_events_from_1_CPU
==================
WARNING: DATA RACE
Read at 0x00c0001262a8 by goroutine 9:
  github.com/aquasecurity/tracee/pkg/events/sorting.(*EventsChronologicalSorter).addEvent()
      /vagrant/pkg/events/sorting/sorting.go:176 +0x5d
  github.com/aquasecurity/tracee/pkg/events/sorting.(*EventsChronologicalSorter).Start()
      /vagrant/pkg/events/sorting/sorting.go:158 +0x2d6
  github.com/aquasecurity/tracee/pkg/events/sorting.TestEventsChronologicalSorter_Start.func1·dwrap·14()
      /vagrant/pkg/events/sorting/sorting_test.go:266 +0x8b

Previous write at 0x00c0001262a8 by goroutine 10:
  github.com/aquasecurity/tracee/pkg/events/sorting.sendTestEvents()
      /vagrant/pkg/events/sorting/sorting_test.go:305 +0x1a4
  github.com/aquasecurity/tracee/pkg/events/sorting.TestEventsChronologicalSorter_Start.func1·dwrap·15()
      /vagrant/pkg/events/sorting/sorting_test.go:267 +0x64

Goroutine 9 (running) created at:
  github.com/aquasecurity/tracee/pkg/events/sorting.TestEventsChronologicalSorter_Start.func1()
      /vagrant/pkg/events/sorting/sorting_test.go:266 +0x384
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1259 +0x22f
  testing.(*T).Run·dwrap·21()
      /usr/local/go/src/testing/testing.go:1306 +0x47

Goroutine 10 (finished) created at:
  github.com/aquasecurity/tracee/pkg/events/sorting.TestEventsChronologicalSorter_Start.func1()
      /vagrant/pkg/events/sorting/sorting_test.go:267 +0x48b
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1259 +0x22f
  testing.(*T).Run·dwrap·21()
      /usr/local/go/src/testing/testing.go:1306 +0x47
==================
```

PR #1634, opened 2022-04-04
fix(test): data race in TestEventsChronologicalSorter_Start

danielpacak:
Resolves: #1630

### DIFF
--- a/pkg/events/sorting/sorting_test.go
+++ b/pkg/events/sorting/sorting_test.go
@@ -1,12 +1,11 @@
 package sorting
 
 import (
+	"context"
 	"fmt"
 	"sort"
 	"testing"
 	"time"
-
-	"golang.org/x/net/context"
 
 	"github.com/aquasecurity/tracee/types/trace"
 	"github.com/stretchr/testify/assert"
@@ -129,7 +128,6 @@ type sorterTestCase struct {
 }
 
 func TestEventsChronologicalSorter_Start(t *testing.T) {
-	t.Skip("It's causing DATA RACE, see https://github.com/aquasecurity/tracee/issues/1630")
 	testCases := []sorterTestCase{
 		{
 			eventsPools: []eventsIteration{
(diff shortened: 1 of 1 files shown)


pkg/events/sorting/sorting_test.go
```diff
@@ -297,13 +299,19 @@ func retrieveEventsFromSorter(expectedEventsAmount int, sorterOutputChan chan *t
...
 }
 
-func sendTestEvents(sorterInputChannel chan *trace.Event, eventPool []eventsIteration) {
-	startTime := int(time.Now().UnixNano())
-	for _, eventsIteration := range eventPool {
-		time.Sleep(eventsIteration.delay)
-		for _, event := range eventsIteration.events {
-			event.Timestamp = startTime + (event.Timestamp * int(time.Millisecond.Nanoseconds()))
-			sorterInputChannel <- &event
+func sendTestEvents(eventPool []eventsIteration) <-chan *trace.Event {
```
danielpacak, line 302: Make sendTestEvents the owner of the events channel. This is a common pattern to reason about Go concurrent programs. 
Beyond that, close the channel explicitly and use non buffered channel it's easier to catch data races.

pkg/events/sorting/sorting_test.go
```diff
@@ -273,13 +272,16 @@ func TestEventsChronologicalSorter_Start(t *testing.T) {
...
 
-func retrieveEventsFromSorter(expectedEventsAmount int, sorterOutputChan chan *trace.Event, fatalErrorsChan chan error) ([]trace.Event, error) {
+func retrieveEventsFromSorter(expectedEventsAmount int, sorterOutputChan <-chan *trace.Event, fatalErrorsChan chan error) ([]trace.Event, error) {
 	ticker := time.NewTicker(time.Second)
 	outputList := make([]trace.Event, 0)
 	eventsReceived := 0
 	for {
 		select {
-		case event := <-sorterOutputChan:
+		case event, ok := <-sorterOutputChan:
```
danielpacak, line 281: I think in the sorter itself we're not checking for the closed channel, but it's not relevant to fix the DATA RACE warnings.

AlonZivony: I will check it

rafaeldtinoco: @AlonZivony any feedback here ?

PR #1639, opened 2022-04-05
fix(test): data race in Test_setupOutput

danielpacak:
Resolves: #1629

### DIFF
--- a/cmd/tracee-rules/output_test.go
+++ b/cmd/tracee-rules/output_test.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


cmd/tracee-rules/output_test.go
```diff
@@ -112,26 +113,30 @@ HostName: foobar.local
 	}
 
 	for _, tc := range testCases {
-		var actualOutput bytes.Buffer
-		findingCh, err := setupOutput(&actualOutput, "", "", "", tc.outputFormat)
-		require.NoError(t, err, tc.name)
+		t.Run(tc.name, func(t *testing.T) {
+			actualOutput := NewSyncBuffer([]byte{})
```
danielpacak, line 117: That was the root cause of DATA RACE warning it test code. bytes.Buffer is not thread safe, whereas it was access by two goroutines.

cmd/tracee-rules/output_test.go
```diff
@@ -112,26 +113,30 @@ HostName: foobar.local
...
+					"foo1": "bar1, baz1",
+					"foo2": []string{"bar2", "baz2"},
+				},
+				Event:       tc.inputEvent,
+				SigMetadata: sm,
+			}
 
-		time.Sleep(time.Millisecond)
-		checkOutput(t, tc.name, actualOutput, tc.expectedOutput)
+			time.Sleep(time.Millisecond)
```
danielpacak, line 133: We should get rid of this sleepy code one day, but I keep it out of scope to enable -race flag ASAP.

rafaeldtinoco (approved):
Elegant fix.

PR #1642, opened 2022-04-05
refactor: move debug setting to flags package

NDStrahilevitz:
## Changes

The debug settings is now defined in the `flags` package.
The PR also uses Tracee's internal debug setting instead of the previous global var where it was needed.

## Why?

This was done so flags can access the debug status without passing it as an arg.


rafaeldtinoco (approved):
LGTM based on our previous conversation.

PR #1643, opened 2022-04-06
fix: add containerId to derived container events

NDStrahilevitz:
This happens because the containerId is derived from the cgroup at the end of the pipeline.
However the process creating the container will be running from a different cgroup.


NDStrahilevitz:
I'm not sure if this is by design or not, so we can close this if it is.

NDStrahilevitz:
Now that I think of it this was a mistake since a container can create another one, so i'm closing this.

Issue #1644, opened 2022-04-06
Decouple tracee-rules and engine from `trace.Event`


NDStrahilevitz:
When the `protocol` package for events was implemented, the headers were defined according to a common but not necessarily fitting way for tracee-rules.

The headers then had to enrich payload info into themselves so for example the ContentType header would look like: "tracee/event-execve".

This by itself wouldn't be a problem, except the engine matches events to signatures through a different system which defines three fields:
1. Name
2. Origin
3. Source

While these may be slightly domain specific to tracee-ebpf, they can easily fit other sources, however with our current headers a conversion between them and the signature selector was needed (see in `engine.go` line:296). 

I propose matching the headers at the minimum to the SignatureSelectors, this will keep backwards compatibility and easily allow new and generic `protocol.Event`s from other sources.
There are 4 independent PRs required to reach this goal:
1. Producing protocol events from tracee-ebpf (#1531)
2. Matching the `EventHeaders` struct (#1645)
3. Removing the coupling in engine.go (#1646)
4. Allowing any payload from protocol in REGO signatures (#1654)

I'm opening the issue as a main thread for for opinions on the overall proposal, so it won't be divided between the PRs.

Once all of these are merged, another PR should be opened to update the Signature Writing documentation to more accurately reflect the way the protocol is used.

NDStrahilevitz:
@rafaeldtinoco @danielpacak Tagging you as Rafael requested.

danielpacak:
I think it make sense. Could you provide sample values for each header type? I'm not quite sure if I understand the difference between Origin and Source headers in particular.

NDStrahilevitz:
@danielpacak  So, as I explained these headers are a bit problematic because they were designed with the ebpf domain specifically in mind,  but now we need to extrapolate them for generic usage.
So classically as an example from tracee-ebpf it would look like so:
```
{
    Source: "tracee",
    Origin: "host",
    Name: "execve"
}
```

So I assume Source would indicate other systems (for example k8s, CNDR,  etc.), Origin could maybe be a subsystem of Source (so maybe in k8s that could be audit_log) and Name is an arbitrary field. Eventually, what decides these headers will be the conventions signatures writers use... Best we can do is document it. I would ideally change names in the Selectors but that would be a breaking change. @AsafEitani I think your feedback would be appreciated for this.

NDStrahilevitz:
To include REGO parity in this goal, I've uploaded #1654 and added it to the original post.

[...]

PR #1646, opened 2022-04-06
engine: use the matching headers for signature selection

NDStrahilevitz:
As described in #1644 

This implements the header changes in #1645 

With this PR the engine signature selection will be decoupled from tracee-ebpf's events, and will be able to process any `protocol.Event` which existing signatures can support.

This will be in draft until #1645 is merged.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.17
 require (
 	github.com/Masterminds/sprig/v3 v3.2.2
 	github.com/aquasecurity/libbpfgo v0.2.5-libbpf-0.7.0
-	github.com/aquasecurity/tracee/types v0.0.0-20220228102148-dffb469aed94
+	github.com/aquasecurity/tracee/types v0.0.0-20220412160215-ccc81e5e59cd
 	github.com/google/gopacket v1.1.19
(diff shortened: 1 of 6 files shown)


danielpacak:
👋 @NDStrahilevitz We've merged #1645 so I think you can rebase and test your changes.

go.mod
```diff
@@ -86,3 +86,5 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 )
+
+replace github.com/aquasecurity/tracee/types => github.com/NDStrahilevitz/tracee/types v0.0.0-20220412091018-d793c93b2ebc
```
danielpacak, line 90: Except this replace statement other changes look good to me.

danielpacak (changes requested):

go.mod
NDStrahilevitz: Yes this was for testing, now that the types commit is up I will update this.

NDStrahilevitz: Done

PR #1648, opened 2022-04-07
added debugfs_create_file event

itamarmaouda101:
Hi 

This PR is the second PR (2nd out of 3) that will solve #1613 .
It adds to Tracee a kprobe event that traces calls to 'debugfs_create_file' which is a kernel function that is used to create a file under the debug filesystem (under /sys/fs/debug).

BTW this is the main page of the function:
[https://www.kernel.org/doc/htmldocs/filesystems/API-debugfs-create-file.html](https://www.kernel.org/doc/htmldocs/filesystems/API-debugfs-create-file.html
)

The event's arguments will be the file name, its path, mode, and the file operations pointer in the memory

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -3262,13 +3262,13 @@ int BPF_KPROBE(trace_debugfs_create_file)
     if (!should_trace((&data.context)))
         return 0;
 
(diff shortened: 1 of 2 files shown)


AlonZivony (approved):
LGTM

rafaeldtinoco (approved):
LGTM, will squash and merge.

PR #1649, opened 2022-04-07
add debugfs_create_dir event

itamarmaouda101:
Hi 

This PR is the third of 3 PR that solve #1613 .

It adds to Tracee a kprobe event that trace calls to `debugfs_create_dir`, which is a kernel function that is used by the kernel module to create directories under /sys/kernel/debug.

Here is the man page of the function from the kernel docs:
[https://www.kernel.org/doc/htmldocs/filesystems/API-debugfs-create-dir.html](https://www.kernel.org/doc/htmldocs/filesystems/API-debugfs-create-dir.html
)

The event will print the name of the dir and its path, here's s an example for output:
`04:33:27:509621  0      insmod           3426    3426    0                debugfs_create_dir   name: malware_poc, path: /sys/kernel/debug/
`

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -202,7 +202,8 @@ Copyright (C) Aqua Security inc.
 #define DIRTY_PIPE_SPLICE               1038
 #define DEBUGFS_CREATE_FILE             1039
 #define PRINT_SYSCALL_TABLE             1040
-#define MAX_EVENT_ID                    1041
+#define DEBUGFS_CREATE_DIR              1041
+#define MAX_EVENT_ID                    1042
 
 #define NET_PACKET                      4000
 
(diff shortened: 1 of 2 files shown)


AlonZivony (approved):
I liked the fact you concat the relative debugfs path to the normal mount point of the fs :)
LGTM

rafaeldtinoco:
@itamarmaouda101 this one needs rebasing now.

pkg/ebpf/events_processor.go
```diff
@@ -333,6 +333,12 @@ func (t *Tracee) processEvent(event *trace.Event) error {
 			return fmt.Errorf("error parsing cgroup_mkdir args: %w", err)
 		}
 		t.containers.CgroupRemove(cgroupId, hId)
+	case DebugfsCreateDir:
+		path, err := getEventArgStringVal(event, "path")
+		if err != nil {
+			return fmt.Errorf("error parsing debufs_create_dir args: %w", err)
+		}
+		event.Args[1].Value = fmt.Sprintf("/sys/kernel/debug/%s", path)
```
rafaeldtinoco, line 341: Could you re-check this ? Based on my test in `debugfs_create_file`:

```
"args":[{"name":"file_name","type":"const char*","value":"testing123"},{"name":"path","type":"const char*","value":"/usb"}
```

You might get "//dirname" here. Also, you're assuming debugfs is always mounted in /sys/kernel/debug (not sure that is the case everywhere, needs checking.

rafaeldtinoco (changes requested):
Needs rebase and some comments given.

pkg/ebpf/events_processor.go
yanivagman: As this event is already called "debugfs_create_file", I don't think that this code is even needed

yanivagman (changes requested):

pkg/ebpf/events_processor.go
AlonZivony: I think that if it is possible to pass the full path to the file instead of only its path inside the FS it would be much more user friendly.
It will also match the way the `path` argument is used in other events.

itamarmaouda101:
Hi 
I rebased and removed the lines you talked about them.

let me know what you think, BTW the event is debugfs_create_dir

yanivagman (approved):
LGTM

rafaeldtinoco (approved):
LGTM

PR #1654, opened 2022-04-10
rules: allow generic protocol.Events in rego signatures

NDStrahilevitz:
## Changes

1. Removes the validation for trace.Event casting
2. In case of an irrelevant payload to a rego signature, no error will be returned, the document will simply not match
3. Adds relevant tests for invalid event payloads, where the finding is nil

### DIFF
--- a/pkg/rules/regosig/traceerego.go
+++ b/pkg/rules/regosig/traceerego.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/aquasecurity/tracee/types/detect"
 	"github.com/aquasecurity/tracee/types/protocol"
-	"github.com/aquasecurity/tracee/types/trace"
 	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/rego"
 )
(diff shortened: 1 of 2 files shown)


pkg/rules/regosig/traceerego.go
```diff
@@ -159,14 +158,7 @@ func (sig *RegoSignature) getSelectedEvents(pkgName string) ([]detect.SignatureE
 // if bool is "returned", a true evaluation will generate a Finding with no data
 // if document is "returned", any non-empty evaluation will generate a Finding with the document as the Finding's "Data"
 func (sig *RegoSignature) OnEvent(event protocol.Event) error {
-	var input rego.EvalOption
-	eventObj, ok := event.Payload.(trace.Event)
```
mtcherni95, line 163: cool the fact we do not need to assert anymore.

pkg/rules/regosig/traceerego_test.go
```diff
@@ -347,22 +356,6 @@ func OnEventSpec(t *testing.T, target string, partial bool) {
 			}
 		})
 	}
-
-	t.Run("Should return error when event has unrecognized type", func(t *testing.T) {
```
mtcherni95, line 351: not sure I understand where this piece of code went, if needed

NDStrahilevitz: It's not needed anymore since there's no "unrecognized type" anymore. If the type is irrelevant for the rule, then evaluation will simply not detect anything. I've added a  new relevant test case with a `nil` finding instead.

danielpacak (changes requested):
It seems that we must resolve conflicts in tests.

danielpacak (approved):

PR #1661, opened 2022-04-10
add register_chrdev_region event

itamarmaouda101:
Hi

This PR is part of #1613 

It adds a kprobe event that traces calls to `register_chardev_region `which is used to create character devices.

Char devices are usually used to create a communication between the user space and the kernel.

Here's the DOCO of the function:
https://www.kernel.org/doc/htmldocs/kernel-api/API-register-chrdev-region.html

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -202,7 +202,8 @@ Copyright (C) Aqua Security inc.
 #define DIRTY_PIPE_SPLICE               1038
 #define DEBUGFS_CREATE_FILE             1039
 #define PRINT_SYSCALL_TABLE             1040
-#define MAX_EVENT_ID                    1041
+#define REGISTER_CHARDEV_REGION         1041
+#define MAX_EVENT_ID                    1042
 
 #define NET_PACKET                      4000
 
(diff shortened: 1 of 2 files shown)


pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4728,6 +4729,25 @@ int BPF_KPROBE(trace_security_inode_mknod)
...
+int BPF_KPROBE(trace_register_chrdev_region)
+{
+    event_data_t data = {};
+    if (!init_event_data(&data, ctx))
+        return 0;
+
+    if (!should_trace(&data.context))
+        return 0;
+
+    u32 major_number = PT_REGS_PARM1(ctx);
```
yanivagman, line 4742: Seems like the second parameter is not the major number but the device count:
https://elixir.bootlin.com/linux/v5.17.3/source/fs/char_dev.c#L200

yanivagman (changes requested):
I wonder why you chose to use this function that is being called to register a range of device numbers (and which is not being called in every path taken to register a char device) and not `__register_chrdev` instead?

pkg/ebpf/c/tracee.bpf.c
itamarmaouda101: Yes u right, i changed to first trace `__register_chrdev` and fixed the prams.
BTW , I saw that the real major number is the return of the function, do you have any idea how we can get the return value of the event?

pkg/ebpf/events_definitions.go
```diff
@@ -6325,4 +6326,17 @@ var EventsDefinitions = map[int32]EventDefinition{
...
+		ID32Bit: sys32undefined,
+		Name:    "register_chrdev_region",
+		Probes: []probe{
+			{event: "__register_chrdev", attach: kprobe, fn: "trace___register_chrdev"},
+		},
+		Sets: []string{},
+		Params: []trace.ArgMeta{
+			{Type: "unsigned int", Name: "major_number"},
+			{Type: "const char*", Name: "char_device_name"},
+			{Type: "u64", Name: "char_device_fops"},
```
AlonZivony, line 6339: Pass this argument as a `struct file_operations*` - it would make more sense (and all pointers are treated as `u64` behind the scenes anyway)

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4728,6 +4729,27 @@ int BPF_KPROBE(trace_security_inode_mknod)
...
+int BPF_KPROBE(trace___register_chrdev)
+{
+    event_data_t data = {};
+    if (!init_event_data(&data, ctx))
+        return 0;
+
+    if (!should_trace(&data.context))
+        return 0;
+
+    unsigned int major_number = PT_REGS_PARM1(ctx);
```
AlonZivony, line 4742: This is problematic as @yanivagman said because the major might be received by the return value of the function.
In my opinion, you can make this program a `kretprobe` and than analyze the return value as the major in the case `major==0`.

AlonZivony (changes requested):
I continue with what said before me

yanivagman:
Replaced by #1693

Issue #1651, opened 2022-04-08
eBPF CO-RE improvements: remove non CO-RE source code


grantseltzer:
Continuing to maintain non-CORE bpf code adds more complexity to the bpf code, build system, and logistics for installing/running. With the addition of BTFHub support we should be able to remove support for non-CORE code. This would involve removing the IFDEF macros in bpf code, and build file recipes

yanivagman:
This seems like a logical step for the future of Tracee, however, as we already have such a support implemented, I think we should wait for a while with removing it, at least until BTF will become more common on existing kernels. I would say a year from now would be a better time to remove support for non -core bpf code

rafaeldtinoco:
> This seems like a logical step for the future of Tracee, however, as we already have such a support implemented, I think we should wait for a while with removing it, at least until BTF will become more common on existing kernels. I would say a year from now would be a better time to remove support for non -core bpf code

We have a list of supported environments for Tracee and we can add support for any environment that does not have BTF enabled in BTFHUB. That, per se, is already something that enables us to drop non CO-RE code.

What is the supported environment you currently see not supported in BTFHUB ? Should we add support there ?

yanivagman:
As long as non-core environments are still common, I don't think BTFhub is enough. When you spot an environment for which you need to genereate a new BTF file, you can immediately start using Tracee on that environment as it takes time to report  this missing environment in BTFHUB and generate a BTF for it, not to mention that the already embedded BTFs we have in Tracee won't be enough, and the user will have to wait for the next version of Tracee to be out before he will be able to use it (assuming that he reported on the missing BTF to BTFHub, and that it was added in time before the next release).

There are few cases where this can happen today: 
1. LTS kernels without BTF support, where new patches appear from time to time
2. Distros that are not currently in BTFHub - there are so many distros out there, and only a portion of it is available in BTFHub
3. Custom compiled kernels

For this reason, I think we should keep this support for now

rafaeldtinoco:
> As long as non-core environments are still common, I don't think BTFhub is enough. 

I don't think they're common, that is the point. And we can't guarantee supportability in *any existing environment* IMHO. BTFhub was meant to add support to LTS distributions without BTF support that we currently TEST and SUPPORT (or the requirements would be to use a kernel that supports BTF).

> When you spot an environment for which you need to generate a new BTF file, you can't immediately start using Tracee on that environment as it takes time to report this missing environment in BTFHUB and generate a BTF for it, 

[...]

PR #1663, opened 2022-04-11
bugfix: support dirty_pipe_splice in unexploited kernels

### LINKED ISSUE #1659: Build of non-core bpf object fails on kernels < 5.8
On main branch, tracee non-core bpf object doesn't compile with kernels before 5.8:
 `error: use of undeclared identifier 'PIPE_BUF_FLAG_CAN_MERGE' if ((out_pipe_last_buffer_flags & PIPE_BUF_FLAG_CAN_MERGE) == 0) {
...`

### LINKED ISSUE #1691: do_splice eBPF program is not affected by filters
## Prerequisites

- [ ] There isn't an issue describing the bug.

## Bug description

The `do_splice` eBPF program does not check if it should be traced, but if it should be submitted.
This means that filters won't work on the event.

AlonZivony:
Event implementation uses a symbol which was introduced only in kernel version 5.8.
Fixed the bug using it and a bug of using the `get_symbol_addr` helper (global values are not supported in older kernels),
Also prevent event analyzing if splice failed, just to reduce overhead and wrong analyze of kernel's data.

This PR has a problem that old Manajro kernels don't have the `do_splice` symbol for some reason in `kallsyms` file, which make the event unloadable in these versions. Ubuntu 4.18 was fine though. I don't see a way to bypass it so it will be suffice for now.

fix #1659 
fix #1691

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -2300,7 +2300,8 @@ static __always_inline unsigned short get_inode_mode_from_fd(u64 fd)
 
 static __always_inline struct pipe_inode_info *get_file_pipe_info(struct file *file) {
     struct pipe_inode_info *pipe = READ_KERN(file->private_data);
-    if (READ_KERN(file->f_op) != get_symbol_addr("pipefifo_fops")) {
+    char pipe_fops_sym[14] = "pipefifo_fops";
+    if (READ_KERN(file->f_op) != get_symbol_addr(pipe_fops_sym)) {
         return NULL;
     }
     return pipe;
@@ -4918,6 +4919,27 @@ TRACE_ENT_FUNC(do_splice, DIRTY_PIPE_SPLICE);
 SEC("kretprobe/do_splice")
 int BPF_KPROBE(trace_ret_do_splice)
 {
+    if (!should_submit(DIRTY_PIPE_SPLICE))
+        return 0;
+
+    // The Dirty Pipe vulnerability exist in the kernel since version 5.8, so there is not use to do logic if version
+    // is too old. In non-CORE, it will even mean using defines which are not available in the kernel headers, which
+    // will cause bugs.
+    #if !defined(CORE) && (LINUX_VERSION_CODE < KERNEL_VERSION(5, 8, 0))
+    return 0;
+    #else
+    #ifdef CORE
+    // Check if field of struct exist to determine kernel version - some fields
+    // change between versions.
+    // In version 5.8 of the kernel, the field "high_zoneidx" changed its name to "highest_zoneidx".
+    // This means that the existence of the field "high_zoneidx" can indicate that the kernel version is lower
+    // than v5.8
+    struct alloc_context *check_508;
+    if (bpf_core_field_exists(check_508->high_zoneidx)) {
+        return 0;
+    }
+    #endif // CORE
+
     args_t saved_args;
     if (load_args(&saved_args, DIRTY_PIPE_SPLICE) != 0) {
         // missed entry or not traced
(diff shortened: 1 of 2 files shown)


rafaeldtinoco:
> This PR has a problem that old Manajro kernels don't have the `do_splice` symbol for some reason in `kallsyms` file, which make the event unloadable in these versions. Ubuntu 4.18 was fine though. I don't see a way to bypass it so it will be suffice for now.

I find it hard that do_splice was optimized as inline (because of its size).

That leads me to think what if functions are optimized and symbols don't exist. This would be similar to the probe not existing (and not making tracee to fail when loading eBPF programs).

Also, related to do_splice, I couldn't find a kconfig to opt out of it. That is weird. It is defined as a syscall and implemented in linux/fs/splice.c without any opt outs.

AlonZivony:
> I find it hard that do_splice was optimized as inline (because of its size).

I find it hard to believe myself. Unfortunately it really is missing from my `kallsyms` file, I don't have a concrete explanation to it.
Because it is not an exported function maybe it is possible for it to go missing in some kernel compilation.

yanivagman:
I think that instead of adding this ZONE stuff for kernel version detection, we should handle it in userspace code and just not load the probe if the kernel < 5.8 (this will also solve the problem you might have with old manjaro versions)

AlonZivony:
I agree with you, though it is hard to do it currently.
I think the right thing to do is when we support not loading a program we will add a dependency of kernel version to check if it is supported in the version or not.
I can add the dependency now for the event but the support for the not loading part in user-mode is out of scope for this PR IMO.

yanivagman:
> I agree with you, though it is hard to do it currently. I think the right thing to do is when we support not loading a program we will add a dependency of kernel version to check if it is supported in the version or not. I can add the dependency now for the event but the support for the not loading part in user-mode is out of scope for this PR IMO.

Yes, I agree. Then we can wait for #1657 to be merged, on which @NDStrahilevitz will rebase and add the ability of not loading an event's programs when something fails. After that, adding a kernel version dependency to an event should be straight forward as we already have the "dependencies" field for each event where we can add the kernel version, and we also extract the running kernel version on tracee start, and pass it as part of the configuration.

AlonZivony:
@yanivagman I think for now we should merge this PR so Tracee will be stable until the dependency PRs are merged.
If you have an issue with it please tell me.

rafaeldtinoco:
> @yanivagman I think for now we should merge this PR so Tracee will be stable until the dependency PRs are merged. If you have an issue with it please tell me.

I agree with @AlonZivony. We should merge this instead of holding a feature focusing on specific distro (manjaro) older kernels. This is the type of burden I mentioned at https://github.com/aquasecurity/tracee/issues/1651#issuecomment-1099058753.

yanivagman:
IMO we shouldn't merge this hacky way (using high_zoneidx) of finding the kernel version as adding kernel version event dependency is just around the corner (within days).
If you want to merge this and remove it once the new mechanism is here, go ahead

rafaeldtinoco:
> IMO we shouldn't merge this hacky way (using high_zoneidx) of finding the kernel version as adding kernel version event dependency is just around the corner (within days).

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4734,6 +4735,27 @@ TRACE_ENT_FUNC(do_splice, DIRTY_PIPE_SPLICE);
...
+    return 0;
+    #else
+    #ifdef CORE
+    // Check if field of struct exist to determine kernel version - some fields
+    // change between versions.
+    // In version 5.8 of the kernel, the field "high_zoneidx" changed its name to "highest_zoneidx".
+    // This means that the existence of the field "high_zoneidx" can indicate that the kernel version is lower
+    // than v5.8
+    struct alloc_context *check_508;
+    if (bpf_core_field_exists(check_508->high_zoneidx)) {
```
rafaeldtinoco, line 4754: @yanivagman I agree with you this is "hacky". If you think we can wait until the "per event" support I'm okay. Please also, take a look at recent merges using the same technique (this goes together with that older discussion os us about not being able to branch condition kconfig relocations, thus the use of this "technique"). Maybe we should revisit those as well.

rafaeldtinoco:
@AlonZivony PR: #1657 has been merged now and I believe this was waiting on it. Feel free to rebase and make needed changes to use the "handle events" PR changes.

AlonZivony:
Also fix #1691

yanivagman:
> I'm merging this, even being a `hacky` approach. I've opened the bug: #1743 and assigned to Alon so, when we define the kernel versioning feature, in event dependencies, Alon will have to re-fix it based on that.
> 
> PS: It is better to have a hacky fix than a broken test =D

Yes, I agree. It took longer than I thought

Issue #1672, opened 2022-04-17
Create a signature for syscall detection in tracee-rules


itamarmaouda101:
## Prerequisites

- [x] I'm going to create a PR to solve this (assign to yourself)

## Feature description
We should add to tracee-rules a signature to the detect_hooked_syscalls event.

## Context

* Linux version: Ubuntu 20.04.3 LTS
* Linux kernel version: 5.13.0-39-generic
* Tracee version (or commit id of your tree): de2749f27f2a0bf891018f547012ec2263fdf08f
* LLVM version: llc-12
* Golang version: go1.17.6

AsafEitani:
#1705

PR #1679, opened 2022-04-18
add syscall detection signature

itamarmaouda101:
## Description

Hi this PR adds to tracee-rules a signeture for detecting hooking of system calls in the system call table.

I added it as an extra to the `detect_hooked_syscalls` event in tracee-ebpf
The changes i added is the signature and also a little mistake (n the derived event of `detect_hooked_syscalls`, the mistake is that the name of the event we print wasn't like in the events definition map (in the map it was `detect_hooked_syscalls` and in the drived-event it was `hooked_syscalls`)

Fixes: #1672 

Please pick one and delete the others:

- New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

I add to the signature test file and test manually the `detect_hooked_syscalls` event

## Final Checklist:

- [x] I have added tests that prove my fix is effective or that my feature works.
- [ ] Any dependent changes have been merged and published already.

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/signatures/rego/syscall_table_hooking.rego
+++ b/signatures/rego/syscall_table_hooking.rego
@@ -1,0 +1,35 @@
+package tracee.TRC_15
+
+import data.tracee.helpers
+
+__rego_metadoc__ := {
+    "id": "TRC-15",
+    "version": "0.1.0",
+    "name": "Hooking system calls by override the system call table entires",
+    "description": "Usage of kernel module to hooks system calls",
+    "tags": ["linux"],
+    "properties": {
+        "Severity": 5,
+        "MITRE ATT&CK": "Persistence: Hooking system calls entires in the system-call table",
+    }
+}
+
+eventSelectors := [
+    {
+        "source": "tracee",
+        "name": "detect_hooked_syscalls"
+    },
+]
+
+tracee_selected_events[eventSelector] {
+	eventSelector := eventSelectors[_]
+}
+
+tracee_match = res {
+    input.eventName == "detect_hooked_syscalls"
+    hooked_syscalls_arr := helpers.get_tracee_argument("hooked_syscalls")
+    c := count(hooked_syscalls_arr)
+    c > 0
+    res :={"hooked syscall": hooked_syscalls_arr}
+
+}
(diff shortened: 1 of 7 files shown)


pkg/ebpf/events_definitions.go
```diff
@@ -6322,7 +6322,7 @@ var EventsDefinitions = map[int32]EventDefinition{
 		},
 		Sets: []string{},
 		Params: []trace.ArgMeta{
-			{Type: "HookedSyscallData[]", Name: "hooked_syscalls"},
+			{Type: "bufferdecoder.HookedSyscallData[]", Name: "hooked_syscalls"},
```
AlonZivony, line 6325: I can't agree with what you have done here.
If you want to export a type through signature, it should be under the `types` package.

signatures/rego/syscall_table_hooking.rego
```diff
@@ -0,0 +1,35 @@
+package tracee.TRC_15
+
+import data.tracee.helpers
+
+__rego_metadoc__ := {
+    "id": "TRC-15",
+    "version": "0.1.0",
+    "name": "Hooking System calls by override The System call table entires",
```
AlonZivony, line 8: Capitalize only the first letter.

signatures/rego/syscall_table_hooking_test.rego
```diff
@@ -0,0 +1,40 @@
...
+        "args": [
+            {
+                "name": "hooked_syscalls",
+                "value": [{"SyscallName":"kill","ModuleOwner":"diamorphine"},{"SyscallName":"getdents","ModuleOwner":"diamorphine"},{"SyscallName":"getdents64","ModuleOwner":"diamorphine"}]
+            }
+        ]
+    }
+}
+
+test_match_2 {
```
AlonZivony, line 16: What is the difference between test_match_1 and test_match_2?
You need to have unique tests, there is no real use to doing the same test twice.
You can test for example the behavior with 1 hooked syscall and multiple hooked syscalls as different senerios.

AlonZivony (changes requested):
Looks good.
Change what I wrote and I don't see a reason for it not to be merged quickly.

rafaeldtinoco:
@AsafEitani 

I believe this one will be replaced by: PR: #1705 ? If that is so, could you close this one please ?

Issue #1684, opened 2022-04-25
Streamline events pipeline


NDStrahilevitz:
## Prerequisites

- [x] Is this an EPIC issue describing other issues ? (add label: EPIC)
- [ ] I'm going to create a PR to solve this (assign to yourself)
- [ ] Someone else should solve this.

## Feature description

This an epic which may include bug fixes and features, but probably mostly refactors.

## Context

Currently, the ebpf package is structured to use a pipeline, where channels are passed to pipeline function which return other channels:

<img width="1458" alt="image" src="https://user-images.githubusercontent.com/7395852/166618159-38baa8a8-7066-4d38-904b-f42c2f2d4756.png">

However, there is no formal abstraction for this besides a convention set by the code. Some problems that arise from this, for example:

1. The output buffers may vary in size between the pipeline functions.
2. Pipeline function signatures may change between pipeline stages
3. The pipeline passes a `trace.Event` and a gocontext, however we may require additional **defined** context between pipeline steps

I'm suggesting to refactor this part as part of the ongoing "stabilization" effort, and possibly move the pipeline to it's own `pipeline` package, which can then be used alongside a separated `events` package.

In addition, the pipeline should be able to easily process events from various eBPF buffers (for example the network event buffer)

## Additional

PR: #1689 (pipeline: split process and sink stages)

yanivagman:
This issue can be closed in favor of the following 2 issues:

1. #3254 is about extracting pipeline to its own package
2. As part of #2870 we will create an internal pipeline struct that will embed trace.Event in it and some extra internal fields (e.g. timeInPipeline). Also see this comment https://github.com/aquasecurity/tracee/issues/2870#issuecomment-1570919351

PR #1689, opened 2022-04-26
pipeline: split process and sink stages

NDStrahilevitz:
## Description

This PR:
1. Removes the sinking and derivation of events from processEvents
2. Adds a new event sinking pipeline stage
3. Adds a new event derivation pipeline stage
4. Makes the out channel buffer uniform across the pipeline stages (10000)

Relevant to #1684

This was done in order to have an enrichment stage (see #1572) in the pipeline, after the process stage but before deriving and sinking.

## Type of change

- Refactor (non-breaking change which adds functionality)

## How Has This Been Tested?

Unit tests were ran locally.

Manual testing:
In order to check that event derivation was working correctly I have done the following test:

1. Build tracee
2. Run it as such:`./dist/tracee-ebpf --trace event=cgroup_mkdir,container_create`
3. In parallel, start a container as such `docker run --rm redis`
4. A container_create event should be outputted

## Final Checklist:

- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] Any dependent changes have been merged and published already.

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/ebpf/events_derived.go
+++ b/pkg/ebpf/events_derived.go
@@ -1,6 +1,7 @@
 package ebpf
 
 import (
+	"context"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


rafaeldtinoco:
Reviewing this now...

rafaeldtinoco:
Alright, from what I can understand the pipeline looks like this now:

image

And I think it looks good after this change. I'm approving it, merging, and will run the weekly tests on top of it just to double check.

rafaeldtinoco (approved):
LGTM. I'm running: https://github.com/aquasecurity/tracee/actions/runs/2267515129 after this change to make sure we're good (I couldn't find a reason why we wouldn't be, just double checking).

yanivagman:
I don't understand - why do we need a buffered channel if there is only one thread per each stage in the pipeline?

rafaeldtinoco:
> I don't understand - why do we need a buffered channel if there is only one thread per each stage in the pipeline?

AFAICT channel buffers are queues for events being picked by poll/select like mechanism:

You can "queue" more than 1 event to those mechanisms each time the task pushing events is scheduled. As soon as the task pushing events is preempted by the scheduler, or the buffer gets full, it stops pushing events. When the channel reader is scheduled, it will block until all events queued are read (read hungry approach).

This way you won't have a 1-event-scheduled per 1-time-event-pusher is scheduled, like if we would by having a non buffered channel.

yanivagman:
So if I understand correctly, "the read hungry" approach is the reason for using a buffered channel?
If so, why do we need it to be of size 10000? Won't it be enough to have it of size 1/10/100?
Eventually, bigger buffers mean more memory, right?

Issue #1688, opened 2022-04-26
empty process context for net_packet event


roikol:
## Prerequisites

- [ ] This affects latest released version.
- [x] This affects current development tree (origin/HEAD).
- [x] There isn't an issue describing the bug.
- [x] I'm going to create a PR to solve this (assign to yourself).
- [ ] Someone else should solve this.

## Bug description

When tracing net events (currently the only net event is net_packet), and not capturing net, the process context of the net_packet event is empty.

## Steps to reproduce

cmdline:
`./dist/tracee-ebpf --trace event=net_packet --capture net=enp0s3`

result:
![image](https://user-images.githubusercontent.com/17714090/165305796-419e47e9-1619-4e64-af59-36e86985ae53.png)

## Context

* Linux version: Ubuntu 20.04.2 LTS
* Linux kernel version: 5.13.0-39-generic
* Tracee version (or commit id of your tree): 2a606cbfab417942132e9dc73a146aa2ddef249e
* LLVM version: 12
* Golang version: 1.18.1

## Additional Information

When also capturing net, the process context is as expected. Also, for processes running before tracee is started, the process context is as expected.

This is because the process tree is updated only if `OPT_PROCESS_INFO` (`optProcessInfo`) is set, and it set only on this condition: 

`t.config.Capture.NetIfaces != nil || t.config.Debug` 

https://github.com/aquasecurity/tracee/blob/main/pkg/ebpf/tracee.go#L504

rafaeldtinoco:
@roikol 

Thinking more about this one, I believe we should have checked for length instead of != nil:

```diff
commit c6f382e6 (HEAD -> int-to-trace-fix)
Author: Rafael David Tinoco <rafaeldtinoco@gmail.com>
Date:   Tue Apr 26 22:49:22 2022

    network: InterfacesToTrace should be checked by length

    Commit 82b4ec3e ("fix empty process context for net_packet event
    (#1692)") checks if InterfacesToTrace is not null but it should check
    for length being bigger than 0 instead.

diff --git a/pkg/ebpf/tracee.go b/pkg/ebpf/tracee.go
index d22222db..66d6a666 100644
--- a/pkg/ebpf/tracee.go
+++ b/pkg/ebpf/tracee.go
@@ -281,7 +281,7 @@ func New(cfg Config) (*Tracee, error) {
                t.events[MprotectEventID] = eventConfig{}
                t.events[MemProtAlertEventID] = eventConfig{}
        }
[...]

PR #1692, opened 2022-04-26
set network related options for traced net events instead of just cap…

roikol:
…ture net

## Description

Up until now the process tree was only used for the net events - network events from the kernel to the `netChannel` come only with `hostTid`, so in order to have the full process context, we retrieve it from the process tree (with `hostTid` as key).

The configuration to maintain process tree (`optProcessInfo`) was only set if user asked to capture net.

Once the `net_packet` event was introduced, user can also ask to trace net, and so we need to set the configuration to maintain process tree for trace net, which is what this PR is doing.

Also, for the same reason, setting security_socket_bind as essential event (it is not related to this bug, but it should be set in the same manner).

Fixes: #1688

## Type of change

Please pick one and delete the others:

- Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

ran 
`./dist/tracee-ebpf -t e=net_packet -t net=enp0s3`

and got the expected results - process context not empty.

## Final Checklist:

- [ ] I have added tests that prove my fix is effective or that my feature works.
- [x] Any dependent changes have been merged and published already.

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/ebpf/tracee.go
+++ b/pkg/ebpf/tracee.go
@@ -281,7 +281,7 @@ func New(cfg Config) (*Tracee, error) {
 		t.events[MprotectEventID] = eventConfig{}
 		t.events[MemProtAlertEventID] = eventConfig{}
 	}
-	if t.config.Capture.NetIfaces != nil || cfg.Debug {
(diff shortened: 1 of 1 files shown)


rafaeldtinoco (approved):
LGTM. @roikol please follow the git log guidelines next time, I'm fixing it for you this time (git log subject length is too big, etc).

Issue #1622, opened 2022-03-30
Dynamic loading of eBPF programs


AlonZivony:
As there are more and more reasons for an eBPF program to fail while loading,
it seems that a more flexible approach to a failure is needed - instead of exiting, tracee should log that it failed to load the event and move on.
This dynamic loading should be both for failure in the loading itself, or because of some missing dependency.
It is critical to unload/prevent the loading of eBPF programs which depends on failed one, and also unload/prevent the loading of programs which the failed one depends on would be beneficial (for reducing performance for unneeded programs).

Examples for failing reasons because of dependencies:
* Failed loading kernel symbols
* Failed loading critical kconfigs

rafaeldtinoco:
Related: https://github.com/aquasecurity/tracee/issues/1588

PR #1707, opened 2022-04-28
ebpf: add a new required dependency field.

NDStrahilevitz:
## Initial Checklist

## Description (git log)

Marking an event as required will make it so failing to attach it's probe will fail tracee.
Otherwise, tracee will continue to load as usual and simply log the failure.
Now events can be loaded with partial dependencies.

Fixes: #1622

## Type of change

New feature (non-breaking change adding functionality).

I've marked event dependencies that I think should be required as such, but I'm not 100% confident that I got it right on the first go. So anyone reviewing should pay extra attention to that :)

In addition, the previously named "essential" events have been marked as required, and a function for handling all initial required events was extracted out of `New` into `initRequiredEvents`. There may also be edge cases there to watch for.

## How Has This Been Tested?

This hasn't been tested yet. It will be tested on a reported GKE environment where tracee failed to load the `VfsWritev` event. After testing that `magic_write` can be loaded there I will report back here.

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/ebpf/events_definitions.go
+++ b/pkg/ebpf/events_definitions.go
@@ -25,6 +25,7 @@ type probe struct {
 type eventDependency struct {
 	eventID      int32
 	shouldSubmit bool
+	required     bool
 }
 
 type dependencies struct {
@@ -6202,7 +6203,7 @@ var EventsDefinitions = map[int32]EventDefinition{
 		Name:    "container_create",
 		Probes:  []probe{},
 		Dependencies: dependencies{
-			events: []eventDependency{{eventID: CgroupMkdirEventID, shouldSubmit: true}},
+			events: []eventDependency{{eventID: CgroupMkdirEventID, shouldSubmit: true, required: true}},
 		},
 		Sets: []string{},
 		Params: []trace.ArgMeta{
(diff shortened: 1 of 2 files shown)


yanivagman:
I think that what should be marked as required or not is the probe, and not the event.
If an event was chosen as a dependency of another event, the meaning is that we must successfully load and attach that event, otherwise, it shouldn't be marked as a dependency.
On the other hand, some probes of an event may be required, while other not. One example is vfs_writev probe for magic_write, and another example is when we will have probes that depend on some bpf features that only exist on newer kernels, we might choose if it is a requirement or not for the event.

NDStrahilevitz:
> I think that what should be marked as required or not is the probe, and not the event. If an event was chosen as a dependency of another event, the meaning is that we must successfully load and attach that event, otherwise, it shouldn't be marked as a dependency. On the other hand, some probes of an event may be required, while other not. One example is vfs_writev probe for magic_write, and another example is when we will have probes that depend on some bpf features that only exist on newer kernels, we might choose if it is a requirement or not for the event.

Agreed, so I'll post here what we discussed offline.
You plan to flatten the probe dependencies away from the event dependencies. 
Then this PR (or yours) can introduce a `probeDependency` struct where we can mark a probes as require instead, so we could have a probe as required for one event and unrequired for another (for example vfs_writev and magic_write). 

So this PR will wait on that.

rafaeldtinoco (approved):
I'm approving but left couple of comments so you can chose to accept if you think they're worth. Overall change looks good and I like the fact that we're auto-loading all programs (we will errors independently of given cmdline).

I see that you are mentioning we should load (or not) programs based on existing kernel features (and chose to attach or not based on current config, that might do reconfigs). Check also tracee.go for network programs "autoload" set to false. Maybe you want to get rid of that as well (so network programs are also always loaded).

yanivagman:
@NDStrahilevitz this PR needs to be rebased on main as probes are now flatten (like we previously discussed)

NDStrahilevitz:
I am closing this in favor of the fix i've done in #1791

PR #1731, opened 2022-05-03
tracee-ebpf: add tail calls to event dependencies

yanivagman:
## Initial Checklist

## Description (git log)

Build a generic mechanism to declare the required tail-call for the proper functionality of an event.
In addition, Introudce capture events to take care of capture dependencies (probes and tail-calls) automatically.

## Type of change

New feature (non-breaking change adding functionality).

## How Has This Been Tested?

Manually run and verified the proper functionality of the following events:
- exexcve
- execveat
- socket_dup

Manually run and verified the proper functionality of the following capture features:
- capture write
- capture mem

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/ebpf/events_definitions.go
+++ b/pkg/ebpf/events_definitions.go
@@ -21,13 +21,20 @@ type probe struct {
 	fn     string
 }
 
+type dependencies struct {
+	events    []eventDependency // Events required to be loaded and/or submitted for the event to happen
+	ksymbols  []string
+	tailCalls []tailCall
+}
+
 type eventDependency struct {
 	eventID int32
 }
 
-type dependencies struct {
-	events   []eventDependency // Events required to be loaded and/or submitted for the event to happen
-	ksymbols []string
+type tailCall struct {
+	mapName  string
+	mapIdx   uint32
+	progName string
 }
 
 // EventDefinition is a struct describing an event configuration
@@ -764,6 +771,11 @@ var EventsDefinitions = map[int32]EventDefinition{
 		ID32Bit: sys32execve,
 		Name:    "execve",
 		Syscall: true,
+		Dependencies: dependencies{
+			tailCalls: []tailCall{
+				{mapName: "sys_enter_tails", mapIdx: uint32(ExecveEventID), progName: "syscall__execve"},
+			},
+		},
 		Sets:    []string{"default", "syscalls", "proc", "proc_life"},
 		Params: []trace.ArgMeta{
 			{Type: "const char*", Name: "pathname"},
@@ -3514,6 +3526,11 @@ var EventsDefinitions = map[int32]EventDefinition{
 		ID32Bit: sys32execveat,
 		Name:    "execveat",
 		Syscall: true,
+		Dependencies: dependencies{
+			tailCalls: []tailCall{
+				{mapName: "sys_enter_tails", mapIdx: uint32(ExecveatEventID), progName: "syscall__execveat"},
+			},
+		},
 		Sets:    []string{"default", "syscalls", "proc", "proc_life"},
 		Params: []trace.ArgMeta{
 			{Type: "int", Name: "dirfd"},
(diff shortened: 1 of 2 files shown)


rafaeldtinoco (changes requested):
This is a very nice change. I have proposed a small change on top of it so it is easier for us to understand tail calls types, event dependencies and which tail call types would be enabled per configuration option. WDYT ?

https://github.com/yanivagman/tracee/pull/1

yanivagman:
> This is a very nice change. I have proposed a small change on top of it so it is easier for us to understand tail calls types, event dependencies and which tail call types would be enabled per configuration option. WDYT ?
> 
> [yanivagman#1](https://github.com/yanivagman/tracee/pull/1)

Capture stuff shouldn't be a dependency of an event. For a matter of fact, in a future PR I want to decouple events and capture artifacts, meaning that when some artifact needs to be captured, it has its own dependencies of tail calls (like in this PR) and bpf programs (probes), so no event dependencies should exist then

rafaeldtinoco:
> Capture stuff shouldn't be a dependency of an event. For a matter of fact, in a future PR I want to decouple events and capture artifacts, meaning that when some artifact needs to be captured, it has its own dependencies of tail calls (like in this PR) and bpf programs (probes), so no event dependencies should exist then

Well I liked your change exactly because it associated events with tail calls BUT I don't like having tail calls disclosed somewhere else because I think it is hard to associate which probes depend on which tail call (without having to go through tracee.bpf.c file and finding them out one by one). It will be even harder when we start splitting that file.

I think you should associate all tail calls with the **probes** that depend on them, in the same place, just like you've done with event dependencies. It looks to me that a tailcall is both: a probe AND a capture-option dependency (and not an event dependency only, which I agree).

I can understand some tail calls won't be used if capturing wasn't selected (multiple types of capturing). Then it is the case of not registering them, like I proposed, based on the types of captures.

Feel free to merge mine and make change, change yours, etc. If you still want to use yours, feel free to ask for another review for the conflict resolution.

Thanks

yanivagman:
> Well I liked your change exactly because it associated events with tail calls BUT I don't like having tail calls disclosed somewhere else because I think it is hard to associate which probes depend on which tail call (without having to go through tracee.bpf.c file and finding them out one by one). It will be even harder when we start splitting that file.

I agree with that, and this was my intention in creating this PR. There are two steps for this to happen - associating tail calls with events (done in this PR), and associating tail calls with the different capture feature.
I didn't implement the second step in this PR, as it will require introducing capture definition like we have for events definition, and I think it should be done in a different PR.

So the question is - do you think that this PR takes a step closer to achieving this goal, or is there anything here that I CHANGED that takes us a step back?

yanivagman:
Also note that I wrote that capture stuff shouldn't be a dependency of an event. I didn't say that tail calls shouldn't be a dependency of the capture stuff. Like you said:

> It looks to me that a tailcall is both: a probe AND a capture-option dependency (and not an event dependency only, which I agree).

yanivagman:
Ok @rafaeldtinoco, I combined the two PRs into one for completeness as you suggested.
This should answer your second concern:

> I don't like having tail calls disclosed somewhere else because
> I think it is hard to associate which probes depend on which tail call
> (without having to go through tracee.bpf.c file and finding them out one
> by one). It will be even harder when we start splitting that file.

An event is now self contained (from tail-calls dependencies perspective), and using the event dependencies infrastructure for the capture features avoid code duplication, and makes things more elegant.

rafaeldtinoco:
> An event is now self contained (from tail-calls dependencies perspective), and using the event dependencies infrastructure for the capture features avoids code duplication, and makes things more elegant.

Thanks a lot. I'm reviewing it now.

rafaeldtinoco (approved):
> Ok @rafaeldtinoco, I combined the two PRs into one for completeness as you suggested. This should answer your second concern:
> 
> > I don't like having tail calls disclosed somewhere else because
> > I think it is hard to associate which probes depend on which tail call
> > (without having to go through tracee.bpf.c file and finding them out one
> > by one). It will be even harder when we start splitting that file.
> 
> An event is now self contained (from tail-calls dependencies perspective), and using the event dependencies infrastructure for the capture features avoids code duplication, and makes things more elegant.

Okay.. definitely the Capture events are not "events" (like all other events purpose) but, tbh, I found the solution elegant. Approving it and thanks for addressing my concern.

PR #1766, opened 2022-05-17
tracee-rules: introduce CEL signatures

danielpacak:
## Initial Checklist

## Description (git log)

This patch introduces new type of Tracee-Rules signatures that can be
written as Common Expression Language (CEL).

This feature is enabled by placing CEL signature definition files
(.cel, .yaml, .yml) in the  the --rules-dir directory. Sample definition
files can be found in the pkg/rules/celsig/testdata/rules/ directory.

The following snippet shows how you define the Anti-Debugging signature using CEL:

    # pkg/rules/celsig/testdata/rules/anti_debugging_ptraceme.yml
    kind: SignaturesConfig
    apiVersion: tracee.aquasecurity.github.io/v1alpha1
    signatures:
     - metadata:
          id: "CEL-2"
          version: "0.1.0"
          name: "Anti-Debugging"
          tags:
            - "linux"
            - "containers"
          properties:
            "Severity": 3
            "MITRE ATT&CK": "Defense Evasion: Execution Guardrails"
        eventSelectors:
          - source: tracee
            name: ptrace
        expression: |-
          input.eventName == 'ptrace' && input.stringArg('request') == 'PTRACE_TRACEME'

You can use the following commands to compile Tracee and load CEL-2, CEL-5, and
CEL-12 signatures from pkg/rules/celsig/testdata/rules/anti_debuggin_ptraceme.yml
and pkg/rules/celsig/testdata/rules/signatures.yaml.

    vagrant up && vagrant ssh
    make
    cp -v pkg/rules/celsig/testdata/rules/*.y*ml dist/rules
    sudo ./dist/tracee-ebpf \
      --output=format:gob \
      --output=option:parse-arguments \
      | ./dist/tracee-rules \
      --input-tracee=file:stdin \
      --input-tracee=format:gob

Notice that CEL-2, CEL-5, and CEL-12 correspond to TRC-2, TRC-5, and TRC-12 signaturese
written as OPA Rego policies.

With Helm installer you can define CEL signatures in Helm values.yaml as follows:

    # deploy/helm/tracee/values.yaml
    signatures:
      # config defines Common Expression Language (CEL) signatures that are loaded
      # by Tracee-Rules from the --rules-dir directory. If the config object is not
      # empty, the configuration file is automatically mounted at
      # /tracee/rules/signatures.yaml.
      config:
        kind: SignaturesConfig
        apiVersion: tracee.aquasecurity.github.io/v1alpha1
        signatures:
          - metadata:
              id: "CEL-2"
              version: "0.1.0"
              name: "Anti-Debugging"
              tags:
                - "linux"
                - "containers"
              properties:
                Severity: 3
                "MITRE ATT&CK": "Defense Evasion: Execution Guardrails"
            eventSelectors:
              - source: tracee
                name: ptrace
            expression: |-
              input.eventName == 'ptrace' && input.stringArg('request') == 'PTRACE_TRACEME'

## Type of change

## How Has This Been Tested?

You can use the following commands to compile Tracee and load CEL-2, CEL-5, and
CEL-12 signatures from pkg/config/testdata/tracee-config.yml.
 
    vagrant up && vagrant ssh
    make
    sudo ./dist/tracee-ebpf \
      --output=format:gob \
      --output=option:parse-arguments \
      | ./dist/tracee-rules \
      --input-tracee=file:stdin \
      --input-tracee=format:gob \
      --traceeconfig=/vagrant/pkg/config/testdata/tracee-config.yml

Notice that CEL-2, CEL-5, and CEL-12 correspond to TRC-2, TRC-5, and TRC-12 signaturese
written as OPA Rego policies.

### Benchmarks OPA Rego vs CEL

```console
$ vagrant@ubuntu-impish:/vagrant/pkg/rules/regosig$ go test -run=XXX -bench=. -benchmem -cpu=1
goos: linux
goarch: amd64
pkg: github.com/aquasecurity/tracee/pkg/rules/regosig
cpu: Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz
BenchmarkSignature_OnEvent/Anti-Debugging         	   20691	     66679 ns/op	   20757 B/op	     390 allocs/op
BenchmarkSignature_OnEvent/Illegitimate_Shell     	   15615	     70510 ns/op	   27896 B/op	     552 allocs/op
BenchmarkSignature_OnEvent/Fileless_Execution     	   16000	     87947 ns/op	   32527 B/op	     573 allocs/op
BenchmarkSignature_OnEvent/SockAddr               	   18630	     58711 ns/op	   24206 B/op	     442 allocs/op
PASS
ok  	github.com/aquasecurity/tracee/pkg/rules/regosig	7.799s
```

```console
$ vagrant@ubuntu-impish:/vagrant/pkg/rules/celsig$ go test -run=XXX -bench=. -benchmem -cpu=1
goos: linux
goarch: amd64
pkg: github.com/aquasecurity/tracee/pkg/rules/celsig
cpu: Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz
BenchmarkSignature_OnEvent/Anti-Debugging         	  485341	      4086 ns/op	    2185 B/op	      14 allocs/op
BenchmarkSignature_OnEvent/Illegitimate_Shell     	  164449	      6744 ns/op	    3410 B/op	      57 allocs/op
BenchmarkSignature_OnEvent/Fileless_Execution     	  369415	      3560 ns/op	    2267 B/op	      28 allocs/op
BenchmarkSignature_OnEvent/SockAddr               	  246528	      4602 ns/op	    3060 B/op	      32 allocs/op
PASS
ok  	github.com/aquasecurity/tracee/pkg/rules/celsig	5.788s
```

## Final Checklist:

## Git Log Checklist:

My commits logs have:


AsafEitani:
Getting this error when trying to run:
```
eitani@ubuntu:~/projects/tracee$ make
GOOS=linux CC=clang GOARCH=amd64 CGO_CFLAGS= CGO_LDFLAGS= go build \
        -tags  \
        -ldflags="-w \
                -extldflags \"\" \
                " \
        -v -o dist/tracee-rules \
        ./cmd/tracee-rules
github.com/aquasecurity/tracee/cmd/tracee-rules
# github.com/aquasecurity/tracee/cmd/tracee-rules
cmd/tracee-rules/main.go:168:32: e.Signatures undefined (type *engine.Engine has no field or method Signatures)
make: *** [Makefile:545: dist/tracee-rules] Error 2
```

danielpacak:
> Getting this error when trying to compile:
> 
> ```
> eitani@ubuntu:~/projects/tracee$ make
> GOOS=linux CC=clang GOARCH=amd64 CGO_CFLAGS= CGO_LDFLAGS= go build \
>         -tags  \
>         -ldflags="-w \
>                 -extldflags \"\" \
>                 " \
>         -v -o dist/tracee-rules \
>         ./cmd/tracee-rules
> github.com/aquasecurity/tracee/cmd/tracee-rules
> # github.com/aquasecurity/tracee/cmd/tracee-rules
> cmd/tracee-rules/main.go:168:32: e.Signatures undefined (type *engine.Engine has no field or method Signatures)
> make: *** [Makefile:545: dist/tracee-rules] Error 2
> ```

Ah sorry. I've removed the clutter and it should compile now.

pkg/config/testdata/tracee-config.yml
```diff
@@ -0,0 +1,57 @@
+kind: TraceeConfig
```
AsafEitani, line 1: What about supporting multiple config files? This file can very quickly turn hugh and unorganized like the falco signatures - which is best avoided in my opinion

pkg/config/testdata/tracee-config.yml
```diff
@@ -0,0 +1,57 @@
...
+apiVersion: tracee.aquasecurity.github.io/v1alpha1
+signatures:
+  - metadata:
```
AsafEitani, line 4: What about other information like mitre technique?

pkg/config/testdata/tracee-config.yml
```diff
@@ -0,0 +1,57 @@
...
+      id: "CEL-12"
+      name: "Illegitimate Shell"
+    eventSelectors:
+      - source: tracee
+        name: security_bprm_check
+        origin: "*"
+    expression: |
+      input.EventName == 'security_bprm_check' and
+      (
+        (arg(input, 'pathname') endsWith '/ash') or
```
AsafEitani, line 22: Can we add support for multiple values comparison? Will make our lives easier.

pkg/config/testdata/tracee-config.yml
AsafEitani: BTW I think it will be better to call it something like rules-config or just rules.yml because we will have an actual config for tracee.

pkg/rules/celsig/signature.go
```diff
@@ -0,0 +1,123 @@
...
+		return err
+	}
+	matches, ok := out.(bool)
+	if !ok {
+		return fmt.Errorf("unrecognized value: %v", out)
+	}
+	if matches {
+		s.cb(detect.Finding{
+			Event:       event,
+			SigMetadata: s.metadata,
```
AsafEitani, line 82: Can we add a feature to get a certain argument or value as the Data field? This could be defined in the signature

cmd/tracee-rules/main.go
```diff
@@ -140,10 +142,52 @@ func main() {
...
+				go func() {
+					for {
+						select {
+						case event, ok := <-watcher.Events:
+							if !ok {
+								return
+							}
+							if event.Op&fsnotify.Write == fsnotify.Write {
+								fmt.Printf("\n*** Configuration Changed ***\nFile: %v\n\n", event)
+								_, err := celsig.LoadFromConfigFile(configPath)
```
AsafEitani, line 160: Why don't we update the signatures?

AsafEitani (commented):
Nice!
Had some questions and requests but the general idea seems really cool.

cmd/tracee-rules/main.go
```diff
@@ -73,22 +75,21 @@ func main() {
 				return err
 			}
 
-			var loadedSigIDs []string
-			for _, s := range sigs {
-				m, err := s.GetMetadata()
+			configPath := c.String("traceeconfig")
+			if configPath != "" {
```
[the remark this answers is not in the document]
danielpacak: We declared traceeconfig flag as optional so the value might be blank string.

cmd/tracee-rules/main.go
danielpacak: I tried but Engine.UnloadSignature followed by Engine.LoadSitnagure has some indeterministic behaviour. Taking it out of scope of this PR, but eventually we may want to reload CEL signatures whenever config file has changes.

pkg/config/testdata/tracee-config.yml
danielpacak: My idea was to define a single config that's read by tracee-ebpf and tracee-rules to keep it comprehensive and in one place. In particular, event selectors will probably be read by tracee-epbf to implement in kernel filtering. The kind and apiVersion are common metadata attributes to version configuration schema.

pkg/config/testdata/tracee-config.yml
danielpacak: We'll add all properties defined in SigMetadata like we do for Rego and Go signatures. I was just lazy to type them in :)

pkg/config/testdata/tracee-config.yml
```diff
@@ -0,0 +1,57 @@
+kind: TraceeConfig
+apiVersion: tracee.aquasecurity.github.io/v1alpha1
```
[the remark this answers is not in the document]
danielpacak: This is common approach in Kubernetes, Kind and other projects where configuration is a first class citizen, i.e. it has schema to validate it and it's versioned from the get go. It facilitates deprecations and extension of configuration properties.

pkg/config/testdata/tracee-config.yml
danielpacak: Certainly, we'll look into existing signatures and add macros that make expression readable.

pkg/rules/celsig/wrapper/wrapper.go
```diff
@@ -0,0 +1,59 @@
...
+	"fmt"
+	"time"
+
+	"github.com/aquasecurity/tracee/types/protocol"
+	"github.com/aquasecurity/tracee/types/trace"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+// Wrap the specified protocol.Event as Event so we can use cel-go without implementing custom ref.TypeProvider.
+func Wrap(envelope protocol.Event) *Event {
```
NDStrahilevitz, line 13: If we pass a `protocol.Event` here and not a `trace.Event`, I think we should just support the whole protocol structure inside the cel sigs (we had a similar discussion once for rego).
Otherwise we may consider giving up the protocol as a whole, and just using `trace.Event`s everywhere.

pkg/rules/celsig/wrapper/event.proto
```diff
@@ -0,0 +1,50 @@
...
+
+syntax = "proto3";
+
+import "google/protobuf/timestamp.proto";
+
+package wrapper;
+
+option go_package = "pkg/rules/celsig/wrapper";
+
+message Event {
```
NDStrahilevitz, line 16: So we will not support header access here?

danielpacak: TBD. This proto message is just a quick win and shortest path to evaluate cel-go implementation and benchmark it. How we model Events and how we use Protocol Buffer messages will probably be a separate PR.

pkg/config/config_test.go
```diff
@@ -0,0 +1,28 @@
...
+			Name: "Anti-Debugging",
+		},
+		EventSelectors: []config.EventSelector{
+			{
+				Source: "tracee",
+				Name:   "ptrace",
+			},
+		},
+		Expression: `input.EventName == 'ptrace' and argument(input, 'request') == 'PTRACE_TRACEME'
+`,
```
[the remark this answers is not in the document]
danielpacak: YAML tricks: I used `|-` instead of `|` in config and the newline is no longer need to assert equality.

pkg/config/testdata/tracee-config.yml
danielpacak: Anyway, I don't have strong opinion on a single (potentially huge) file vs multiple smaller YAML chunks. To be discussed.

pkg/rules/celsig/wrapper/wrapper.go
danielpacak: This wrapper was introduced only for the POC. Not the final solution. Here's why https://github.com/google/cel-go/issues/408

NDStrahilevitz: Ok, I understand none of this is final but wanted to know if this design is planned to change.
I remember I did the same thing in rego and we also both agreed at the time that it wasn't ideal.

danielpacak: I'm not aware of such plans. So long signature authors don't use message headers the current implementation is enough. If this requirement changed we'll change code.

danielpacak: We could always define `header` as input variable and keep input as is of type  protocol.Event.Payload to avoid rewriting all signatures:

```
input.eventName == 'ptrace' && headers['X-Max'] == 'Gift'
```

NDStrahilevitz: I actually like that first style better. Although that looks like it means we wouldn't support different payload types but as you said that doesn't seem to a goal currently going forward.

danielpacak: I think we still could support different payload types. Especially in Cel-go it's legit to write the following expression:

```
type(input) == 'trace.Event' && input.eventName == 'ptrace'
```

But let's wait for a "real" use case before we complicate things ;)

pkg/config/testdata/tracee-config.yml
danielpacak: The current behaviour implemented in this PR is as follows: Tracee-Rules loads 0 to many *.yaml files from the --rules-dir directory. Each file can define 0 to many CEL signatures. In other words, you can have 1 files with N signatures or N files with 1 signature or, in general, N file with M signatures.

pkg/config/testdata/tracee-config.yml
danielpacak: Actually, CEL already defines a handful of macros to support iteration. We can write similar expression using the `exists` macro as follows:

```
['/ash', '/bash', '/csh', '/ksh', '/sh', '/tcsh', '/zsh', '/dash'].exists(e, input.stringArg('pathname').endsWith(e))
```

rafaeldtinoco (approved):
LGTM, had many sessions with @danielpacak about this PR and I'm thankful for all his help on this feature. I'll document this as a tech preview for the release (not doing it on purpose now so I can come up with correct wording and thanks to Daniel.).

PR #1771, opened 2022-05-19
[FEAT] Reintroduce container event enrichment

NDStrahilevitz:
# Changes

Adds container data enrichment from runtime APIs to events.

Currently the data added is:
1. Container name (if exists)
2. Container's image
3. Pod metadata (if in k8s)

Supported runtimes are:
1. Containerd
2. Docker
4. CRI-O

Podman will be done in a later PR, since it requires implementing our own API client (the provided library isn't very good IMO) but integration should be simple through the interfaces introduced here.

Architecture diagram (thanks @mtcherni95):

image


rafaeldtinoco (approved):
After merging this, rolling it back because of race issues, discussing with @NDStrahilevitz, running tests (full tests) multiple times, I think we can merge this now. LGTM.

Let's remove the full tests trigger in workflows file so we can merge this!

rafaeldtinoco:
Im closing this because Im removing the weekly tests (I asked Nadav to do) and merging (as the tests went fine in multiple scenarios). Thanks @NDStrahilevitz

PR #1779, opened 2022-05-31
[MAINT] checkers: use clang-format-13 as a base for eBPF formatting

rafaeldtinoco:


rafaeldtinoco:
nope, moving on by changing .clang-format file and making clang-format-12 and newer compatible.

Issue #1787, opened 2022-06-01
[FEAT] Logger


AlonZivony:
## Prerequisites

- [ ] This issue is an EPIC issue (add label: EPIC).
- [ ] This issue is an EPIC TASK (add issue to EPIC description).

Select one OR another:

- [ ] I'll create a PR to implement this feature (assign to yourself).
- [ ] Someone else should implement this (describe it well).

## Feature description

This feature is about creating a logger class used by tracee, which could be passed to all subclasses.

The feature is needed because there is no graceful way to handle errors from outside the `Tracee` class.
This way each error, alert of some problem or just debug information could be handled using one interface in all places.

Requirements from https://github.com/aquasecurity/tracee/pull/2093?ts=8#issuecomment-1227662401:

1. log errors and debug cases
2. debug should be a no-op when not running in debug mode
3. caching the errors to avoid repetitive prints

## Additional Information (feature drawings, files, logs, etc)

I think of creating an interface for logger which all classes will use as a member.
In initialization the logger will be given (originated from `tracee`).

danielpacak:
Have you considered using existing logging frameworks? For example, https://github.com/sirupsen/logrus or https://github.com/kubernetes/klog They are pretty common and somehow define log level and structured message format so logs can be consumed by other systems for observability.

geyslan:
The current proposal is a zap interface https://github.com/geyslan/logger-poc/blob/main/logger/zap/logger.go which can filter golang packages out and count, when debugging, log triggers by filename and line.

PR #1788, opened 2022-06-01
pkg/logger: created a logger for tracee

AlonZivony:
## Initial Checklist

## Description (git log)

Create a logger class for tracee.

Fixes: #1787

## Type of change

## Final Checklist:

## Git Log Checklist:

My commits logs have:


AlonZivony:
@rafaeldtinoco this a draft of an idea to create a logger, because I think we really need one.
Tell me what you think.
The implementations of the `Alert` and `Info` was just on a whim, we can change them (maybe print `Alert` if in debug instead of passing to error channel?).

rafaeldtinoco:
I like it.

And, yes, having a logger would be awesome. I think cmd/tracee-{ebpf,rules} should have their logger implementations as well (stdout/stderr errors during initialization, etc).

I believe we should create levels for the logger (like other projects do). Higher the log levels are, higher amount of info you get. Something like:

1. ERRORS
2. WARNINGS / ALERT
3. INFO
4. DEBUG

Similar to your proposal actually.
And user can chose logging level.

AlonZivony:
To change tracee-rules to work with the logger interface would be harder, because it uses the `log` package.
I am currently more concerned about the logger in `tracee-ebpf` because log doesn't exist currently.

It might be a good idea to use a global logger with the `ILogger` interface, and expose the methods of it as functions to other packages with `Register` function to determine the logger to be used.
The flow I see in this case would be:
1. Initialize a logger
2. Register the logger as the global logger (the logger itself won't be exposed)
3. Use global `Error`, `Warning`, etc. functions to log (which behind the scenes will call the methods of the global logger).
What do you think would be preferable from these 2 approaches? Global logger or passing the logger to each class?

AlonZivony:
@yanivagman @rafaeldtinoco what do you prefer?
Global functions exposing global unexported logger, or passing the logger from one package to others as a member?
If you have other ideas please write to me.
I want to push forward this PR so we can start using it in new packages in the future :)

rafaeldtinoco:
@AlonZivony are you abandoning this in favor of PR: #1837 ?

AlonZivony:
I think we do

PR #1801, opened 2022-06-06
[MAINT] fix qemu runs for GKE kernels (not supporting FS labels)

rafaeldtinoco:
Maintenance PR regarding distro testing.


rafaeldtinoco:
Included in https://github.com/aquasecurity/tracee/pull/1800

PR #1817, opened 2022-06-09
[MAINT] Fix noisy Test_getSignatures that was trying to load *_test.rego files

danielpacak:
The test output was cluttered with the following messages:

2022/06/09 13:17:11 error creating rego signature with: package tracee.TRC_7: 4 errors occurred:
tracee.TRC_7:4: rego_unsafe_var_error: var tracee_match is unsafe


simar7 (approved):
lgtm, thanks for fixing the noise!

Issue #1813, opened 2022-06-08
[BUG] Capabilities enforcement apply only for 1 thread


AlonZivony:
## Prerequisites

- [ ] This affects latest released version.
- [x] This affects current development tree (origin/HEAD).
- [x] There isn't an issue describing the bug.

Select one OR another:

- [x] I'm going to create a PR to solve this (assign to yourself).
- [ ] Someone else should solve this.

## Bug description

This bug was discovered while investigating #1803 
We noticed that the error message we received isn't received on each run, only for part of them.
When forcing only 1 goroutine, the error occurs in much more runs.
@yanivagman suggested that the current capabilities package change only thread's capabilities, which might not be the thread with reduced capabilities. See the discussion in #1803 for more details.

This is how I'm executing tracee (cmdline):
`sudo ./dist/tracee-ebpf -o none -o option:parse-arguments -o option:detect-syscall-trace comm-bash -trace follow`

This is the error I'm getting:
Inconsistently, get the following error message:
![image](https://user-images.githubusercontent.com/87267148/172633837-2cf7ca58-f772-4358-8e00-ec758f9edf27.png)

## Context

Relevant information about my setup:

* Linux version: Ubuntu
* Linux kernel version: 5.13
* Tracee version (or commit id of your tree):  727e6

rafaeldtinoco:
I believe we've dropped this from this milestone, right ? Just wanted to make sure @AlonZivony (based on our previous meeting).

yanivagman:
> I believe we've dropped this from this milestone, right ? Just wanted to make sure @AlonZivony (based on our previous meeting).

I think we should keep it as part of the milestone as we already merged the dropped capabilities PR, but it's implementation is incorrect, which might cause hard to debug issues on some environments due to it being statistically working.

rafaeldtinoco:
Sure, no objections. I just wanted to confirm that is what we wanted! Thanks.

PR #1819, opened 2022-06-12
capabilities: change capabilities package to libcap

AlonZivony:
## Initial Checklist

## Description (git log)

Change capabilities package to `libcap`.

Fixes: #1813 

## Type of change

## How Has This Been Tested?

- /pkg/capabilities/capabilities_test.go
- /cmd/tracee-ebpf/capabilities_test.go

- make test-unit

## Final Checklist:

## Git Log Checklist:

My commits logs have:


AlonZivony:
I didn't check if this change solves the problem if inconsistent capabilities of the process, though I think it should.
I will need to give an answer on this in the future (I hope I will be able tomorrow).

rafaeldtinoco:
I added a "tests-failed" label just so we are sure you checked if it solves that issue or not (then you can remove it). Just to keep track.

rafaeldtinoco:
@AlonZivony just to make sure I understood, this last change of yours is targeting the fix of [#1813](https://github.com/aquasecurity/tracee/issues/1813) as well, correct ?

AlonZivony:
I managed to run a test to see if the change of the capability package really made the results of capabilities dropping to be deterministic.
I have ran the tests we have on both versions - `cap` package and `capability` package.
From past experience, the `jammy` and `focal` distribution tests should have failed (because of lacking capabilities).
From the results, we can see that the `capabilities` package only failed in 1 test - which shows that it statistically fails/pass.
However, the `cap` package failed on both distros tests - both for CORE and non-CORE.
I believe it should be enough to prove that we should move to the `cap` package.

For reference, here are the tests pipelines for both packages:
* `capability` - https://github.com/aquasecurity/tracee/runs/7092814673?check_suite_focus=true
* `cap` - https://github.com/aquasecurity/tracee/runs/7087911590?check_suite_focus=true

pkg/capabilities/capabilities.go
```diff
@@ -4,47 +4,73 @@ import (
...
 	// therefore it must be patched in for testability
-	NewPid2 = capability.NewPid2
+	GetPID = func(pid int) (ICapabilitiesSet, error) {
+		return cap.GetPID(pid)
+	}
 )
 
-func CheckRequired(caps capability.Capabilities, reqCaps []capability.Cap) error {
+// ICapabilitiesSet is an interface for the cap.Set type, created for testability.
+type ICapabilitiesSet interface {
```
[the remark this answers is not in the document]
grantseltzer: Is there not an implementation of this yet? Just the mock one?

pkg/capabilities/capabilities.go
```diff
@@ -4,47 +4,73 @@ import (
...
 )
 
 var (
-	// NewPid2 is not defined on an interface
+	// GetPID is not defined on an interface
 	// therefore it must be patched in for testability
-	NewPid2 = capability.NewPid2
+	GetPID = func(pid int) (ICapabilitiesSet, error) {
+		return cap.GetPID(pid)
+	}
```
grantseltzer, line 15: Can you describe to me the reason you've designed it this way? What's the purpose of the interface?

pkg/capabilities/capabilities.go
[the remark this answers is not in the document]
AlonZivony: So the 'I' comes from conventions I know from C++, and it is a short for "interface".
I need to check what is the convention for Go

pkg/capabilities/capabilities.go
AlonZivony: The purpose here is to be able to test the functions in this file.
Because the `cap` library doesn't export any interface, to be able to mock its functionality I needed to use a variable that points to the function. This method is used broadly in the project, you can see it in other files.
About the interface, I just need a way to mock the capabilities used. Again, it's all for testability reasons.

cmd/tracee-ebpf/capabilities_test.go
```diff
@@ -20,87 +20,82 @@ func (mOSInfo mockOSInfo) CompareOSBaseKernelRelease(version string) int {
 }
 
 type mockCapabilities struct {
-	missingCaps []capability.Cap
+	missingCaps []cap.Value
 }
 
-func (mockCaps *mockCapabilities) Get(which capability.CapType, what capability.Cap) bool {
+func (mockCaps *mockCapabilities) GetFlag(vec cap.Flag, val cap.Value) (bool, error) {
```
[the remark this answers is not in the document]
AlonZivony: There is no point, it is just the signature for the function in the new library.

grantseltzer:
Defining an interface for testability in Go is smart, but I find that the way this package is designed uses anti-patterns that Go code should not use. In particular, functions should almost never return interfaces, defining an interface which another package implements but isn't designed to, and having a package defined global variable (in this case a function).  It's a bit hard to understand. 

The code can be simplified if you have a defined implementation of the `Set` interface. Just have a struct that all of these functions in `pkg/capabilities` can be methods of.

AlonZivony:
> defining an interface which another package implements but isn't designed to

I don't agree with you on this one. We had couple of discussions on this issue - Go packages never expose an interface for exported types. The common practice is for a package to define interfaces for the API it expects to receive from the type it uses, and by that allow mocking of the type. It was different if the package was under our project, but this is the only way to mock types from another project.

>  having a package defined global variable (in this case a function)

Using a global unexported variable is a very common practice. It is used both for implementing the Singleton design pattern, but also for other uses like testability (like we have here). This is because there is no  built-in option in Go to mock dependencies. 
However, it helps only to test the `Self` function which don't really need testing (it just wrap the `cap.GetPID` function). So I will remove it.

> The code can be simplified if you have a defined implementation of the `Set` interface. Just have a struct that all of these functions in `pkg/capabilities` can be methods of.

If I thought that we might change capabilities package in the future, I would have wrapped the `cap` package API with local functions. However, it seems to me like a waste of time because it is really unlikely.
About creating a type for all the functions - it seems to me a bit odd to do so because there is no state between the functions. They change the process capabilities, which is the real state. I don't think there is a reason to go so far with the implementation here.

yanivagman (approved):
Two small comments, but otherwise LGTM.

pkg/capabilities/capabilities.go
```diff
@@ -4,47 +4,73 @@ import (
...
+}
+
+func CheckRequired(caps Set, reqCaps []cap.Value) error {
 	for _, c := range reqCaps {
-		if !caps.Get(capability.EFFECTIVE, c) {
-			return fmt.Errorf("insufficient privileges to run: missing CAP_%s", strings.ToUpper(c.String()))
+		exist, err := caps.GetFlag(cap.Effective, c)
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
AlonZivony: So you are right, it might not be fatal.
I change it so it will return missing capabilities and error (if one is created from the `cap` package).
The reason this might not be fatal is that the drop capabilities function will ensure the capabilities anyway, just won't make any meaningful error message to the user.

grantseltzer (approved):
LGTM!

rafaeldtinoco:
Ok, Grant gave him +1, lets get a +1 from @yanivagman, @NDStrahilevitz and/or I as well (as we're currently working now) (Yaniv didnt check the latest changes from what @AlonZivony told me).

pkg/capabilities/errors.go
```diff
@@ -0,0 +1,44 @@
...
+
+type MissingCapabilitiesError struct {
+	missingCaps []cap.Value
+}
+
+type DropCapabilitiesError struct {
+	Err error
+}
+
+func newErrMissingCapabilities(missingCaps []cap.Value) *MissingCapabilitiesError {
```
NDStrahilevitz, line 18: Since there's no special logic to creating the error I think the new functions are overkill, just export the struct and fields. No need for encapsulation here, especially since these should be returned as `error`s anyway so the interface will encapsulate naturally.

pkg/capabilities/errors.go
```diff
@@ -0,0 +1,44 @@
...
+
+func (missCapErr *MissingCapabilitiesError) Error() string {
+	var missingCapsNames []string
+	for _, missingCap := range missCapErr.missingCaps {
+		missingCapsNames = append(missingCapsNames, strings.ToUpper(missingCap.String()))
+	}
+	return fmt.Sprintf("insufficient capabilities to run: missing %s", missingCapsNames)
+}
+
+func newErrDropCapabilities(err error) *DropCapabilitiesError {
```
NDStrahilevitz, line 32: ditto

cmd/tracee-ebpf/capabilities.go
```diff
@@ -20,20 +20,22 @@ type IKernelVersionInfo interface {
 
 // ensureCapabilities makes sure program runs with required capabilities only
 func ensureCapabilities(OSInfo IKernelVersionInfo, cfg *tracee.Config, allowHighCapabilities bool) error {
```
NDStrahilevitz, line 22: nit: Not sure why you prefixed the `I` in the interface, it's very much just a Java/C# thing. I'd take the opportunity to just rename into `KernelVersionInfo`.

NDStrahilevitz (changes requested):
Overall it seems fine since it's just a library change. Left some "refactor"-ish comments.

NDStrahilevitz (approved):
LGTM!

Issue #1676, opened 2022-04-18
TC error message when using net events


AlonZivony:
## Prerequisites

- [x] I checked the documentation and found no answer.
- [x] There isn't an issue describing the bug.
- [ ] I can reproduce the issue in latest released version.
- [x] I can reproduce the issue in development version.

## Bug description

When tracing a net event, the following error is received for each interface monitored:
_libbpf: Kernel error message: Exclusivity flag on, cannot modify_

This is the normal flow of the program, so it shouldn't create any error.
This error should be suppressed. 

## Steps to reproduce

```
$ sudo ./dist/tracee-ebpf --capture net=<interface> --output none
libbpf: Kernel error message: Exclusivity flag on, cannot modify
```

## Context

* Linux version: Manjaro
* Linux kernel version: 5.10.105
* Tracee version (or commit id of your tree): de2749f2

yanivagman:
Also see this: https://www.spinics.net/lists/bpf/msg44842.html

rafaeldtinoco:
This has been mitigated by:

https://github.com/aquasecurity/libbpfgo/pull/174

and I'll bump libbpfgo dependency version and it will fade away.

PR #1820, opened 2022-06-13
tc and net probes refactoring

rafaeldtinoco:
## Initial Checklist

## Description (git log)

probes: refactor tc and network probes

- unify probes package for different probes types
- allow probes be attached/detached during runtime
- allow tc probes to be attached/detached from specific interface

This work will help in the future when dynamic attachment/detachment is
neded, but the current focus is to cleanup tracee.go initialization as
well (going forward with probes refactoring).

Commit adds:

- Probes interface:

  - Using a traceProbe Handle (types: rawTracepoint, kprobe, kretprobe)

      Syntax: Attach(EventHandle)
      Syntax: Detach(EventHandle)

  - Using a tcProbe Handle:

      Syntax: Attach(EventHandle, *net.Interface)
      Syntax: Detach(EventHandle, *net.Interface)

  - to detach all probes:

      DetachAll()

- Probe interface:

 - traceProbe (attach/detach/autoload)
 - tcProbe (attach/detach/autoload)

## Type of change

## How Has This Been Tested?

```
$ sudo ./dist/tracee-ebpf -o format:json -o option:parse-arguments -o option:detect-syscall -trace comm=bash -trace follow -trace event=net_packet -trace net=docker0 -capture net=docker0
```

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/ebpf/probes/probes.go
+++ b/pkg/ebpf/probes/probes.go
@@ -2,174 +2,442 @@ package probes
 
 import (
 	"fmt"
+	"net"
 	"strings"
+	"syscall"
 
 	bpf "github.com/aquasecurity/libbpfgo"
 )
 
+//
+// Probes:
+//
+
+// when attaching a traceProbe, by handle, to its eBPF program:
+//
+//   Handle == traceProbe (types: rawTracepoint, kprobe, kretprobe)
+//
+//     Attach(EventHandle)
+//     Detach(EventHandle)
+//
+// when attaching a tcProbe, by handle, to its eBPF program:
+//
+//   Handle == tcProbe
+//
+//     Attach(EventHandle, *net.Interface)
+//     Detach(EventHandle, *net.Interface)
+//
+// to detach all probes:
+//
+//     DetachAll()
+//
+// NOTE: keeping both interfaces with variadic args on **purpose** until we
+//       define real use cases, by extending supported "probes" types (trace,
+//       tc, socket, xdp, tunnel, cgroup, ...) **
+//
+
 type Probes interface {
-	Attach(probe Handle) error
-	Detach(probe Handle) error
+	Attach(handle Handle, args ...interface{}) error
+	Detach(handle Handle, args ...interface{}) error
+	DetachAll() error
 }
 
 type probes struct {
-	probes map[Handle]*probe
 	module *bpf.Module
+	probes map[Handle]Probe
 }
 
-type probe struct {
-	event     string
-	probeType probeType
-	fn        string
-	link      *bpf.BPFLink
-}
+// Init initializes a Probes interface
+func Init(module *bpf.Module, netEnabled bool) (Probes, error) {
 
-//Initialize tracee's probe map
-func Init(module *bpf.Module) (Probes, error) {
+	allProbes := map[Handle]Probe{
+		SysEnter:                   &traceProbe{eventName: "raw_syscalls:sys_enter", probeType: rawTracepoint, programName: "tracepoint__raw_syscalls__sys_enter"},
+		SysExit:                    &traceProbe{eventName: "raw_syscalls:sys_exit", probeType: rawTracepoint, programName: "tracepoint__raw_syscalls__sys_exit"},
+		SchedProcessFork:           &traceProbe{eventName: "sched:sched_process_fork", probeType: rawTracepoint, programName: "tracepoint__sched__sched_process_fork"},
+		SchedProcessExec:           &traceProbe{eventName: "sched:sched_process_exec", probeType: rawTracepoint, programName: "tracepoint__sched__sched_process_exec"},
+		SchedProcessExit:           &traceProbe{eventName: "sched:sched_process_exit", probeType: rawTracepoint, programName: "tracepoint__sched__sched_process_exit"},
+		SchedSwitch:                &traceProbe{eventName: "sched:sched_switch", probeType: rawTracepoint, programName: "tracepoint__sched__sched_switch"},
+		DoExit:                     &traceProbe{eventName: "do_exit", probeType: kprobe, programName: "trace_do_exit"},
+		CapCapable:                 &traceProbe{eventName: "cap_capable", probeType: kprobe, programName: "trace_cap_capable"},
+		VfsWrite:                   &traceProbe{eventName: "vfs_write", probeType: kprobe, programName: "trace_vfs_write"},
+		VfsWriteRet:                &traceProbe{eventName: "vfs_write", probeType: kretprobe, programName: "trace_ret_vfs_write"},
+		VfsWriteV:                  &traceProbe{eventName: "vfs_writev", probeType: kprobe, programName: "trace_vfs_writev"},
+		VfsWriteVRet:               &traceProbe{eventName: "vfs_writev", probeType: kretprobe, programName: "trace_ret_vfs_writev"},
+		KernelWrite:                &traceProbe{eventName: "__kernel_write", probeType: kprobe, programName: "trace_kernel_write"},
+		KernelWriteRet:             &traceProbe{eventName: "__kernel_write", probeType: kretprobe, programName: "trace_ret_kernel_write"},
+		CgroupAttachTask:           &traceProbe{eventName: "cgroup:cgroup_attach_task", probeType: rawTracepoint, programName: "tracepoint__cgroup__cgroup_attach_task"},
+		CgroupMkdir:                &traceProbe{eventName: "cgroup:cgroup_mkdir", probeType: rawTracepoint, programName: "tracepoint__cgroup__cgroup_mkdir"},
+		CgroupRmdir:                &traceProbe{eventName: "cgroup:cgroup_rmdir", probeType: rawTracepoint, programName: "tracepoint__cgroup__cgroup_rmdir"},
+		SecurityBPRMCheck:          &traceProbe{eventName: "security_bprm_check", probeType: kprobe, programName: "trace_security_bprm_check"},
+		SecurityFileOpen:           &traceProbe{eventName: "security_file_open", probeType: kprobe, programName: "trace_security_file_open"},
+		SecurityFileIoctl:          &traceProbe{eventName: "security_file_ioctl", probeType: kprobe, programName: "trace_tracee_trigger_event"},
+		SecurityFilePermission:     &traceProbe{eventName: "security_file_permission", probeType: kprobe, programName: "trace_security_file_permission"},
+		SecuritySocketCreate:       &traceProbe{eventName: "security_socket_create", probeType: kprobe, programName: "trace_security_socket_create"},
+		SecuritySocketListen:       &traceProbe{eventName: "security_socket_listen", probeType: kprobe, programName: "trace_security_socket_listen"},
+		SecuritySocketConnect:      &traceProbe{eventName: "security_socket_connect", probeType: kprobe, programName: "trace_security_socket_connect"},
+		SecuritySocketAccept:       &traceProbe{eventName: "security_socket_accept", probeType: kprobe, programName: "trace_security_socket_accept"},
+		SecuritySocketBind:         &traceProbe{eventName: "security_socket_bind", probeType: kprobe, programName: "trace_security_socket_bind"},
+		SecuritySocketSetsockopt:   &traceProbe{eventName: "security_socket_setsockopt", probeType: kprobe, programName: "trace_security_socket_setsockopt"},
+		SecuritySbMount:            &traceProbe{eventName: "security_sb_mount", probeType: kprobe, programName: "trace_security_sb_mount"},
+		SecurityBPF:                &traceProbe{eventName: "security_bpf", probeType: kprobe, programName: "trace_security_bpf"},
+		SecurityBPFMap:             &traceProbe{eventName: "security_bpf_map", probeType: kprobe, programName: "trace_security_bpf_map"},
+		SecurityKernelReadFile:     &traceProbe{eventName: "security_kernel_read_file", probeType: kprobe, programName: "trace_security_kernel_read_file"},
+		SecurityKernelPostReadFile: &traceProbe{eventName: "security_kernel_post_read_file", probeType: kprobe, programName: "trace_security_kernel_post_read_file"},
+		SecurityInodeMknod:         &traceProbe{eventName: "security_inode_mknod", probeType: kprobe, programName: "trace_security_inode_mknod"},
+		SecurityInodeSymlink:       &traceProbe{eventName: "security_inode_symlink", probeType: kprobe, programName: "trace_security_inode_symlink"},
+		SecurityInodeUnlink:        &traceProbe{eventName: "security_inode_unlink", probeType: kprobe, programName: "trace_security_inode_unlink"},
+		SecurityMmapAddr:           &traceProbe{eventName: "security_mmap_addr", probeType: kprobe, programName: "trace_mmap_alert"},
+		SecurityMmapFile:           &traceProbe{eventName: "security_mmap_file", probeType: kprobe, programName: "trace_security_mmap_file"},
+		DoSplice:                   &traceProbe{eventName: "do_splice", probeType: kprobe, programName: "trace_do_splice"},
+		DoSpliceRet:                &traceProbe{eventName: "do_splice", probeType: kretprobe, programName: "trace_ret_do_splice"},
+		ProcCreate:                 &traceProbe{eventName: "proc_create", probeType: kprobe, programName: "trace_proc_create"},
+		SecurityFileMProtect:       &traceProbe{eventName: "security_file_mprotect", probeType: kprobe, programName: "trace_security_file_mprotect"},
+		CommitCreds:                &traceProbe{eventName: "commit_creds", probeType: kprobe, programName: "trace_commit_creds"},
+		SwitchTaskNS:               &traceProbe{eventName: "switch_task_namespaces", probeType: kprobe, programName: "trace_switch_task_namespaces"},
+		ARMKprobe:                  &traceProbe{eventName: "arm_kprobe", probeType: kprobe, programName: "trace_arm_kprobe"},
+		CallUsermodeHelper:         &traceProbe{eventName: "call_usermodehelper", probeType: kprobe, programName: "trace_call_usermodehelper"},
+		DebugfsCreateFile:          &traceProbe{eventName: "debugfs_create_file", probeType: kprobe, programName: "trace_debugfs_create_file"},
+		DebugfsCreateDir:           &traceProbe{eventName: "debugfs_create_dir", probeType: kprobe, programName: "trace_debugfs_create_dir"},
+		DeviceAdd:                  &traceProbe{eventName: "device_add", probeType: kprobe, programName: "trace_device_add"},
+		RegisterChrdev:             &traceProbe{eventName: "__register_chrdev", probeType: kprobe, programName: "trace___register_chrdev"},
+		RegisterChrdevRet:          &traceProbe{eventName: "__register_chrdev", probeType: kretprobe, programName: "trace_ret__register_chrdev"},
+		DoInitModule:               &traceProbe{eventName: "do_init_module", probeType: kprobe, programName: "trace_do_init_module"},
+		DoInitModuleRet:            &traceProbe{eventName: "do_init_module", probeType: kretprobe, programName: "trace_ret_do_init_module"},
+		LoadElfPhdrs:               &traceProbe{eventName: "load_elf_phdrs", probeType: kprobe, programName: "trace_load_elf_phdrs"},
+		Filldir64:                  &traceProbe{eventName: "filldir64", probeType: kprobe, programName: "trace_filldir64"},
+		TaskRename:                 &traceProbe{eventName: "task:task_rename", probeType: rawTracepoint, programName: "tracepoint__task__task_rename"},
+		UDPSendmsg:                 &traceProbe{eventName: "udp_sendmsg", probeType: kprobe, programName: "trace_udp_sendmsg"},
+		UDPDisconnect:              &traceProbe{eventName: "__udp_disconnect", probeType: kprobe, programName: "trace_udp_disconnect"},
+		UDPDestroySock:             &traceProbe{eventName: "udp_destroy_sock", probeType: kprobe, programName: "trace_udp_destroy_sock"},
+		UDPv6DestroySock:           &traceProbe{eventName: "udpv6_destroy_sock", probeType: kprobe, programName: "trace_udpv6_destroy_sock"},
+		InetSockSetState:           &traceProbe{eventName: "sock:inet_sock_set_state", probeType: rawTracepoint, programName: "tracepoint__inet_sock_set_state"},
+		TCPConnect:                 &traceProbe{eventName: "tcp_connect", probeType: kprobe, programName: "trace_tcp_connect"},
+		ICMPRecv:                   &traceProbe{eventName: "icmp_rcv", probeType: kprobe, programName: "trace_icmp_rcv"},
+		ICMPSend:                   &traceProbe{eventName: "__icmp_send", probeType: kprobe, programName: "trace_icmp_send"},
+		ICMPv6Recv:                 &traceProbe{eventName: "icmpv6_rcv", probeType: kprobe, programName: "trace_icmpv6_rcv"},
+		ICMPv6Send:                 &traceProbe{eventName: "icmp6_send", probeType: kprobe, programName: "trace_icmp6_send"},
+		Pingv4Sendmsg:              &traceProbe{eventName: "ping_v4_sendmsg", probeType: kprobe, programName: "trace_ping_v4_sendmsg"},
+		Pingv6Sendmsg:              &traceProbe{eventName: "ping_v6_sendmsg", probeType: kprobe, programName: "trace_ping_v6_sendmsg"},
+		DefaultTcIngress:           &tcProbe{programName: "tc_ingress", tcAttachPoint: bpf.BPFTcIngress},
+		DefaultTcEgress:            &tcProbe{programName: "tc_egress", tcAttachPoint: bpf.BPFTcEgress, skipLoopback: true},
+	}
+
+	// disable autoload for network related eBPF programs in network is disabled
+	if !netEnabled {
+		for _, p := range allProbes {
+			if tc, ok := p.(*tcProbe); ok {
+				tc.autoload(module, false)
+			}
+		}
+	}
+
 	return &probes{
-		probes: map[Handle]*probe{
-			SysEnter:                   {event: "raw_syscalls:sys_enter", probeType: rawTracepoint, fn: "tracepoint__raw_syscalls__sys_enter"},
-			SysExit:                    {event: "raw_syscalls:sys_exit", probeType: rawTracepoint, fn: "tracepoint__raw_syscalls__sys_exit"},
-			SchedProcessFork:           {event: "sched:sched_process_fork", probeType: rawTracepoint, fn: "tracepoint__sched__sched_process_fork"},
-			SchedProcessExec:           {event: "sched:sched_process_exec", probeType: rawTracepoint, fn: "tracepoint__sched__sched_process_exec"},
-			SchedProcessExit:           {event: "sched:sched_process_exit", probeType: rawTracepoint, fn: "tracepoint__sched__sched_process_exit"},
-			SchedSwitch:                {event: "sched:sched_switch", probeType: rawTracepoint, fn: "tracepoint__sched__sched_switch"},
-			DoExit:                     {event: "do_exit", probeType: kprobe, fn: "trace_do_exit"},
-			CapCapable:                 {event: "cap_capable", probeType: kprobe, fn: "trace_cap_capable"},
-			VfsWrite:                   {event: "vfs_write", probeType: kprobe, fn: "trace_vfs_write"},
-			VfsWriteRet:                {event: "vfs_write", probeType: kretprobe, fn: "trace_ret_vfs_write"},
-			VfsWriteV:                  {event: "vfs_writev", probeType: kprobe, fn: "trace_vfs_writev"},
-			VfsWriteVRet:               {event: "vfs_writev", probeType: kretprobe, fn: "trace_ret_vfs_writev"},
-			KernelWrite:                {event: "__kernel_write", probeType: kprobe, fn: "trace_kernel_write"},
-			KernelWriteRet:             {event: "__kernel_write", probeType: kretprobe, fn: "trace_ret_kernel_write"},
-			CgroupAttachTask:           {event: "cgroup:cgroup_attach_task", probeType: rawTracepoint, fn: "tracepoint__cgroup__cgroup_attach_task"},
-			CgroupMkdir:                {event: "cgroup:cgroup_mkdir", probeType: rawTracepoint, fn: "tracepoint__cgroup__cgroup_mkdir"},
-			CgroupRmdir:                {event: "cgroup:cgroup_rmdir", probeType: rawTracepoint, fn: "tracepoint__cgroup__cgroup_rmdir"},
-			SecurityBPRMCheck:          {event: "security_bprm_check", probeType: kprobe, fn: "trace_security_bprm_check"},
-			SecurityFileOpen:           {event: "security_file_open", probeType: kprobe, fn: "trace_security_file_open"},
-			SecurityFileIoctl:          {event: "security_file_ioctl", probeType: kprobe, fn: "trace_tracee_trigger_event"},
-			SecurityFilePermission:     {event: "security_file_permission", probeType: kprobe, fn: "trace_security_file_permission"},
-			SecuritySocketCreate:       {event: "security_socket_create", probeType: kprobe, fn: "trace_security_socket_create"},
-			SecuritySocketListen:       {event: "security_socket_listen", probeType: kprobe, fn: "trace_security_socket_listen"},
-			SecuritySocketConnect:      {event: "security_socket_connect", probeType: kprobe, fn: "trace_security_socket_connect"},
-			SecuritySocketAccept:       {event: "security_socket_accept", probeType: kprobe, fn: "trace_security_socket_accept"},
-			SecuritySocketBind:         {event: "security_socket_bind", probeType: kprobe, fn: "trace_security_socket_bind"},
-			SecuritySocketSetsockopt:   {event: "security_socket_setsockopt", probeType: kprobe, fn: "trace_security_socket_setsockopt"},
-			SecuritySbMount:            {event: "security_sb_mount", probeType: kprobe, fn: "trace_security_sb_mount"},
-			SecurityBPF:                {event: "security_bpf", probeType: kprobe, fn: "trace_security_bpf"},
-			SecurityBPFMap:             {event: "security_bpf_map", probeType: kprobe, fn: "trace_security_bpf_map"},
-			SecurityKernelReadFile:     {event: "security_kernel_read_file", probeType: kprobe, fn: "trace_security_kernel_read_file"},
-			SecurityKernelPostReadFile: {event: "security_kernel_post_read_file", probeType: kprobe, fn: "trace_security_kernel_post_read_file"},
-			SecurityInodeMknod:         {event: "security_inode_mknod", probeType: kprobe, fn: "trace_security_inode_mknod"},
-			SecurityInodeSymlink:       {event: "security_inode_symlink", probeType: kprobe, fn: "trace_security_inode_symlink"},
-			SecurityInodeUnlink:        {event: "security_inode_unlink", probeType: kprobe, fn: "trace_security_inode_unlink"},
-			SecurityMmapAddr:           {event: "security_mmap_addr", probeType: kprobe, fn: "trace_mmap_alert"},
-			SecurityMmapFile:           {event: "security_mmap_file", probeType: kprobe, fn: "trace_security_mmap_file"},
-			DoSplice:                   {event: "do_splice", probeType: kprobe, fn: "trace_do_splice"},
-			DoSpliceRet:                {event: "do_splice", probeType: kretprobe, fn: "trace_ret_do_splice"},
-			ProcCreate:                 {event: "proc_create", probeType: kprobe, fn: "trace_proc_create"},
-			SecurityFileMProtect:       {event: "security_file_mprotect", probeType: kprobe, fn: "trace_security_file_mprotect"},
-			CommitCreds:                {event: "commit_creds", probeType: kprobe, fn: "trace_commit_creds"},
-			SwitchTaskNS:               {event: "switch_task_namespaces", probeType: kprobe, fn: "trace_switch_task_namespaces"},
-			ARMKprobe:                  {event: "arm_kprobe", probeType: kprobe, fn: "trace_arm_kprobe"},
-			CallUsermodeHelper:         {event: "call_usermodehelper", probeType: kprobe, fn: "trace_call_usermodehelper"},
-			DebugfsCreateFile:          {event: "debugfs_create_file", probeType: kprobe, fn: "trace_debugfs_create_file"},
-			DebugfsCreateDir:           {event: "debugfs_create_dir", probeType: kprobe, fn: "trace_debugfs_create_dir"},
-			DeviceAdd:                  {event: "device_add", probeType: kprobe, fn: "trace_device_add"},
-			RegisterChrdev:             {event: "__register_chrdev", probeType: kprobe, fn: "trace___register_chrdev"},
-			RegisterChrdevRet:          {event: "__register_chrdev", probeType: kretprobe, fn: "trace_ret__register_chrdev"},
-			DoInitModule:               {event: "do_init_module", probeType: kprobe, fn: "trace_do_init_module"},
-			DoInitModuleRet:            {event: "do_init_module", probeType: kretprobe, fn: "trace_ret_do_init_module"},
-			LoadElfPhdrs:               {event: "load_elf_phdrs", probeType: kprobe, fn: "trace_load_elf_phdrs"},
-			Filldir64:                  {event: "filldir64", probeType: kprobe, fn: "trace_filldir64"},
-			TaskRename:                 {event: "task:task_rename", probeType: rawTracepoint, fn: "tracepoint__task__task_rename"},
-		},
+		probes: allProbes,
 		module: module,
 	}, nil
 }
 
-//Attach the probe corresponding to the probe's handle
-func (p *probes) Attach(probe Handle) error {
-	if p.probes[probe] == nil {
-		return fmt.Errorf("failed to attach probe: doesn't exist")
+// Attach attaches given handle's program to its hook
+func (p *probes) Attach(handle Handle, args ...interface{}) error {
+	if _, ok := p.probes[handle]; !ok {
+		return fmt.Errorf("probe handle (%d) does not exist", handle)
 	}
 
-	return p.probes[probe].attach(p.module)
+	return p.probes[handle].attach(p.module, args...)
 }
 
-//Detach the probe corresponding to the probe's handle
-func (p *probes) Detach(probe Handle) error {
-	if p.probes[probe] == nil {
-		return fmt.Errorf("failed to detach probe: doesn't exist")
+// Attach detaches given handle's program from its hook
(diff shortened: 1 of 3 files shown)


rafaeldtinoco:
@NDStrahilevitz I'll start using you more for reviewing code if you don't mind.

pkg/ebpf/probes/probes.go
```diff
@@ -82,51 +84,117 @@ func Init(module *bpf.Module) (Probes, error) {
 			LoadElfPhdrs:               {event: "load_elf_phdrs", probeType: kprobe, fn: "trace_load_elf_phdrs"},
 			Filldir64:                  {event: "filldir64", probeType: kprobe, fn: "trace_filldir64"},
 		},
+		netProbes: []*probe{
```
NDStrahilevitz, line 87: Why not write these as standard probes?

pkg/ebpf/probes/probes.go
```diff
@@ -7,26 +7,28 @@ import (
 	bpf "github.com/aquasecurity/libbpfgo"
 )
 
+// Probes
+
 type Probes interface {
+	// attach
 	Attach(probe Handle) error
+	AttachNetProbes() error
```
NDStrahilevitz, line 15: Can't we handle the net probes the same way we handle probes and just add them as event dependencies? Then we would load them using the same logic we do for standard probes. It would also help streamline the net events code in tracee.go

pkg/ebpf/probes/probes.go
```diff
@@ -7,26 +7,28 @@ import (
...
+	// detach
 	Detach(probe Handle) error
+	DetachNetProbes() error
+	// no autoload
+	DisableNetProbesAutoload() error
```
NDStrahilevitz, line 20: It's reasonable to place this function in this package, but I don't think as part of the interface. I think this will occur naturally if we just treat the net probes the same way we treat the regular probes.
And if we do make the net probes part of the standard probes, this logic could possibly fit better in tracee itself (more of a stylistic decision but I do think there is some merit to keeping the interface minimal).

pkg/ebpf/probes/tc_probes.go
```diff
@@ -0,0 +1,243 @@
...
+// tcProbes
+
+type TcProbes interface {
+	// attach
+	AttachToInterface(iface *net.Interface) error
+	AttachToInterfaceByName(ifaceName string) error
+	AttachToInterfacesByName(ifaceNames []string) error
+	// detach
+	DetachFromInterface(iface *net.Interface) error
+	DetachFromInterfaceByName(ifaceName string) error
```
NDStrahilevitz, line 31: Same goes for these two as the above comment.

NDStrahilevitz (commented):
I think this refactor definitely does the job of cleaning up this code and making it more manageable, so I already like it just for that.

But I do think (at least for the net probes, for the tc probes it might be a bit more complex) it can be better integrated into the probe logics we have, which would make the net code more standardized to the other stuff we have and fit more naturally into the overall flow of code.

pkg/ebpf/probes/probes.go
rafaeldtinoco: Yep, I thought about it.. but it felt weird to me we would have to add all the handles to each existing network event, something like:

image

of course we can optimize it, just explaining where I went that direction, let me see what I can do.

rafaeldtinoco: Anyway, I think that is the price we have to pay. Events such as DNS event wont have ICMP probes, for example. So I think we are good.

pkg/ebpf/probes/tc_probes.go
```diff
@@ -0,0 +1,243 @@
...
+	// add here tcProbes you would like attached to loopback devices only
+	{attachPoint: bpf.BPFTcIngress, programName: "tc_ingress"},
+}
+
+// tcProbes
+
+type TcProbes interface {
+	// attach
+	AttachToInterface(iface *net.Interface) error
+	AttachToInterfaceByName(ifaceName string) error
```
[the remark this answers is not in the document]
rafaeldtinoco: And there is also something to be added to what you said: we have to always index interfaces by ID only, and not name. I think I'll enforce the `iface *net.Interface` argument in the interface for now and solve altogether later (on were to do the interface name <-> *net.Interface conversion).

rafaeldtinoco:
I'm going to bump libbpfgo again because of:

https://github.com/aquasecurity/libbpfgo/pull/174

Which mitigates the following issue:

https://github.com/aquasecurity/tracee/issues/1676

So no exclusivity errors appear in tracee's output.

pkg/ebpf/events_definitions.go
```diff
@@ -5428,6 +5442,15 @@ var EventsDefinitions = map[int32]EventDefinition{
 	DnsRequest: {
 		ID32Bit: sys32undefined,
 		Name:    "dns_request",
+		Probes: []probeDependency{
```
NDStrahilevitz, line 5445: I actually think here we can keep the probes empty, same we do with container_create, since this is a derived event from NetPacket, it should declare the NetPacket event as a dependency.

pkg/ebpf/events_definitions.go
```diff
@@ -5440,6 +5463,15 @@ var EventsDefinitions = map[int32]EventDefinition{
 	DnsResponse: {
 		ID32Bit: sys32undefined,
 		Name:    "dns_response",
+		Probes: []probeDependency{
```
NDStrahilevitz, line 5466: Same as above here.

pkg/ebpf/events_definitions.go
rafaeldtinoco: Example with dns_request only:

```
$ sudo ./dist/tracee-ebpf -o format:json -o option:parse-arguments -o option:detect-syscall -trace comm=bash -trace follow -trace event=dns_request -trace net=docker0 -capture net=docker0

{"timestamp":1655230060646770986,"processorId":0,"processId":13,"cgroupId":0,"threadId":13,"parentProcessId":9,"hostProcessId":3782544,"hostThreadId":3782544,"hostParentProcessId":3782540,"userId":0,"mountNamespace":4026533391,"pidNamespace":4026533430,"processName":"http","hostName":"","containerId":"c1ab59ed5848dee2c17cf334cf6c6a4bbfea73a9ead52bf58ef5f3f394a35501","containerImage":"","containerName":"","podName":"","podNamespace":"","podUID":"","eventId":"4001","eventName":"dns_request","argsNum":2,"returnValue":0,"stackAddresses":null,"args":[{"name":"metadata","type":"trace.PktMeta","value":{"src_ip":"172.17.0.2","dst_ip":"1.1.1.1","src_port":33486,"dst_port":53,"protocol":17,"packet_len":78,"iface":"docker0"}},{"name":"dns_questions","type":"[]bufferdecoder.DnsQueryData","value":[{"query":"archive.ubuntu.com","queryType":"A","queryClass":"IN"}]}]}

{"timestamp":1655230060647083330,"processorId":0,"processId":12,"cgroupId":0,"threadId":12,"parentProcessId":9,"hostProcessId":3782543,"hostThreadId":3782543,"hostParentProcessId":3782540,"userId":0,"mountNamespace":4026533391,"pidNamespace":4026533430,"processName":"http","hostName":"","containerId":"c1ab59ed5848dee2c17cf334cf6c6a4bbfea73a9ead52bf58ef5f3f394a35501","containerImage":"","containerName":"","podName":"","podNamespace":"","podUID":"","eventId":"4001","eventName":"dns_request","argsNum":2,"returnValue":0,"stackAddresses":null,"args":[{"name":"metadata","type":"trace.PktMeta","value":{"src_ip":"172.17.0.2","dst_ip":"1.1.1.1","src_port":34317,"dst_port":53,"protocol":17,"packet_len":79,"iface":"docker0"}},{"name":"dns_questions","type":"[]bufferdecoder.DnsQueryData","value":[{"query":"security.ubuntu.com","queryType":"A","queryClass":"IN"}]}]}
```

and with both:

```
$ sudo ./dist/tracee-ebpf -o format:json -o option:parse-arguments -o option:detect-syscall -trace comm=bash -trace follow -trace event=dns_request -trace event=net_packet -trace net=docker0 -capture net=docker0

{"timestamp":1655230109736500168,"processorId":0,"processId":215,"cgroupId":0,"threadId":215,"parentProcessId":211,"hostProcessId":3795107,"hostThreadId":3795107,"hostParentProcessId":3795103,"userId":0,"mountNamespace":4026533391,"pidNamespace":4026533430,"processName":"http","hostName":"","containerId":"c1ab59ed5848dee2c17cf334cf6c6a4bbfea73a9ead52bf58ef5f3f394a35501","containerImage":"","containerName":"","podName":"","podNamespace":"","podUID":"","eventId":"4000","eventName":"net_packet","argsNum":1,"returnValue":0,"stackAddresses":null,"args":[{"name":"metadata","type":"trace.PktMeta","value":{"src_ip":"172.17.0.2","dst_ip":"1.1.1.1","src_port":33780,"dst_port":53,"protocol":17,"packet_len":78,"iface":"docker0"}}]}

{"timestamp":1655230109736500168,"processorId":0,"processId":215,"cgroupId":0,"threadId":215,"parentProcessId":211,"hostProcessId":3795107,"hostThreadId":3795107,"hostParentProcessId":3795103,"userId":0,"mountNamespace":4026533391,"pidNamespace":4026533430,"processName":"http","hostName":"","containerId":"c1ab59ed5848dee2c17cf334cf6c6a4bbfea73a9ead52bf58ef5f3f394a35501","containerImage":"","containerName":"","podName":"","podNamespace":"","podUID":"","eventId":"4001","eventName":"dns_request","argsNum":2,"returnValue":0,"stackAddresses":null,"args":[{"name":"metadata","type":"trace.PktMeta","value":{"src_ip":"172.17.0.2","dst_ip":"1.1.1.1","src_port":33780,"dst_port":53,"protocol":17,"packet_len":78,"iface":"docker0"}},{"name":"dns_questions","type":"[]bufferdecoder.DnsQueryData","value":[{"query":"archive.ubuntu.com","queryType":"A","queryClass":"IN"}]}]}
```

NDStrahilevitz: You're right I misread the code. 
It does make me confused about the net_packet event now, since it seems the ebpf code can generate it, but it can also be derived inside the ebpf code into a dns_request for example. But then why would we derive it back? Obviously not a question for the scope of this PR though.

pkg/ebpf/probes/tc_probes.go
```diff
@@ -0,0 +1,198 @@
...
+	if module == nil {
+		fmt.Printf("eh null\n")
+	}
+
+	prog, _ := module.GetProgram(p.programName)
+	if prog == nil {
+		return fmt.Errorf("could not find tc program %s", p.programName)
+	}
+
+	var tcOpts bpf.TcOpts
```
NDStrahilevitz, line 161: style nit: why not `tcOpts := bpf.TcOpts{ ProgFd: int(prog.GetFd()) }`?

NDStrahilevitz (changes requested):
LGTM! 
But I did find one place where I'm almost sure there should be an error returned and not a print. Other than that I just have one style nit.

pkg/ebpf/events_definitions.go
rafaeldtinoco: Yep, it was done like that so the DNS event would also be capable of generating a net_event that generated it. Other way of approaching it would be to depend on net_event and to generate a DNS event from it (and others depending on the protocol handlers). Problem with that would be that every network event would depend on net_event (which has many probes). Let's talk about it, definitely.

pkg/ebpf/probes/tc_probes.go
rafaeldtinoco: No special reason, I believe I just moved this code and did not touch it. I'll change it.

pkg/ebpf/probes/tc_probes.go
```diff
@@ -0,0 +1,198 @@
...
+
+	err = p.tcHook.Create()
+	if err != nil {
+		if errno, ok := err.(syscall.Errno); ok && errno != syscall.EEXIST {
+			return fmt.Errorf("tc hook create: %v", err)
+		}
+	}
+
+	if module == nil {
+		fmt.Printf("eh null\n")
```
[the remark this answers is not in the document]
rafaeldtinoco: Just a leftover from dev. Removing it.

pkg/ebpf/probes/tc_probes.go
```diff
@@ -0,0 +1,198 @@
...
+
+// detach detaches a tcProbe from a given Interface
+func (p *tcProbe) detach() error {
+	var err error
+
+	if p.tcHook != nil {
+		err = p.tcHook.Destroy()
+	}
+
+	// TODO: check this
```
rafaeldtinoco, line 179: I still need to check this:

image

I don't want to leave a hanging program/filter just because we're afraid of deleting the qdisc. There is an upstream discussion about this that Yaniv started but, still, there are many BPF programs being left hanging there if you execute tracee multiple times.

rafaeldtinoco: I'm opening the following issue for this: https://github.com/aquasecurity/tracee/issues/1828 AND will continue the exact same behavior in this refactoring.

rafaeldtinoco:
> LGTM! But I did find one place where I'm almost sure there should be an error returned and not a print. Other than that I just have one style nit.

Okay, I've done 2 fixes that were needed:

1. Tcprobe wasn't a pointer so detach had argument as nil.
2. Cleanup wasn't working correctly because of (1), made sure it is.
3. Opened a bug to address existing behavior issues

rafaeldtinoco:
@yanivagman I have requested @NDStrahilevitz to review this one for me (as he played a big part in the probe refactoring). I believe we are in good shape and I'm going to merge this soon, but wanted to highlight you first.

yanivagman:
@rafaeldtinoco if you don't mind, I would like to review this before you merge it.
Hope that I'll be able to do this today

rafaeldtinoco:
> @rafaeldtinoco if you don't mind, I would like to review this before you merge it. Hope that I'll be able to do this today

Nope, take your time. I highlighted you because of that. Thanks!

yanivagman (changes requested):
Indeed there's a lot of refactoring to be done in the network related code.
It was hard for me to review this PR as there is only one commit that does two different things:
1. Refactoring in the sense of moving code into new packages and between files
2. Changing logic - e.g. the way we destory hooks

I'm also in favor of making the tc probes "citizens with equal rights" and combine it with the regular probes by introducing a new attach type ("tc"). The specialized tc functions that deal with netwrok interfaces can be kept in a different file as it is done here though.

pkg/ebpf/events_definitions.go
```diff
@@ -5428,6 +5442,15 @@ var EventsDefinitions = map[int32]EventDefinition{
...
 		ID32Bit: sys32undefined,
 		Name:    "dns_request",
+		Probes: []probeDependency{
+			{handle: probes.UDPSendmsg, required: true},
+			{handle: probes.UDPDisconnect, required: true},
+			{handle: probes.UDPDestroySock, required: true},
+			{handle: probes.UDPv6DestroySock, required: true},
+			{handle: probes.InetSockSetState, required: true},
+			{handle: probes.Pingv4Sendmsg, required: true},
+			{handle: probes.Pingv6Sendmsg, required: true},
```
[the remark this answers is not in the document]
rafaeldtinoco: Nope, I should remove them, missed it.

pkg/ebpf/probes/probes.go
```diff
@@ -7,26 +7,24 @@ import (
 	bpf "github.com/aquasecurity/libbpfgo"
 )
 
+// Probes
+
 type Probes interface {
 	Attach(probe Handle) error
 	Detach(probe Handle) error
+	DetachAll() error
+	DisableNetAutoload() error
```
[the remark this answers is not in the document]
rafaeldtinoco: I thought about it as well. Let me see what I can do there.

pkg/ebpf/probes/probes.go
```diff
@@ -7,26 +7,24 @@ import (
...
-}
-
-type probe struct {
-	event     string
-	probeType probeType
-	fn        string
-	link      *bpf.BPFLink
+	module      *bpf.Module
+	probes      map[Handle]*probe
+	loadNetwork bool
```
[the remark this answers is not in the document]
rafaeldtinoco: image
image

pkg/ebpf/probes/probes.go
```diff
@@ -81,52 +79,108 @@ func Init(module *bpf.Module) (Probes, error) {
...
 		return fmt.Errorf("failed to attach probe: doesn't exist")
 	}
 
 	return p.probes[probe].attach(p.module)
 }
 
-//Detach the probe corresponding to the probe's handle
+// Detach detaches the probe corresponding to the probe's handle.
 func (p *probes) Detach(probe Handle) error {
+
```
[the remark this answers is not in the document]
rafaeldtinoco: I did some research about golang styling enforcement and could not find anything that called my attention (like clang-format). Let me know if you have any idea for style enforcement!

pkg/ebpf/probes/probes.go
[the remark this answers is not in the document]
rafaeldtinoco: Im marking this as solved because the existence or not of this variable will depend on what we conclude in the previous comment (about autoloading network probes by default, during initialization, or not).

pkg/ebpf/probes/probes.go
```diff
@@ -81,52 +79,108 @@ func Init(module *bpf.Module) (Probes, error) {
...
+			UDPDestroySock:             {event: "udp_destroy_sock", probeType: kprobe, fn: "trace_udp_destroy_sock"},
+			UDPv6DestroySock:           {event: "udpv6_destroy_sock", probeType: kprobe, fn: "trace_udpv6_destroy_sock"},
+			InetSockSetState:           {event: "sock:inet_sock_set_state", probeType: rawTracepoint, fn: "tracepoint__inet_sock_set_state"},
+			TCPConnect:                 {event: "tcp_connect", probeType: kprobe, fn: "trace_tcp_connect"},
+			ICMPRecv:                   {event: "icmp_rcv", probeType: kprobe, fn: "trace_icmp_rcv"},
+			ICMPSend:                   {event: "__icmp_send", probeType: kprobe, fn: "trace_icmp_send"},
+			ICMPv6Recv:                 {event: "icmpv6_rcv", probeType: kprobe, fn: "trace_icmpv6_rcv"},
+			ICMPv6Send:                 {event: "icmp6_send", probeType: kprobe, fn: "trace_icmp6_send"},
+			Pingv4Sendmsg:              {event: "ping_v4_sendmsg", probeType: kprobe, fn: "trace_ping_v4_sendmsg"},
+			Pingv6Sendmsg:              {event: "ping_v6_sendmsg", probeType: kprobe, fn: "trace_ping_v6_sendmsg"},
```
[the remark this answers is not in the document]
rafaeldtinoco: I agree, that is the part that Nadav also proposed (and I also felt when developing this). My comment to nadav was to merge this as is (better than what we currently have) and then bring tcProbes to Probes. If you think it is mandatory to have this now, we can try to have it now then (or to break into a next PR doing it). But I do agree.

pkg/ebpf/probes/probes.go
```diff
@@ -81,52 +79,108 @@ func Init(module *bpf.Module) (Probes, error) {
...
 		},
-		module: module,
-	}, nil
+		module:      module,
+		loadNetwork: true,
```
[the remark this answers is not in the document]
rafaeldtinoco: part of previous discussion.

pkg/ebpf/probes/probes.go
```diff
@@ -81,52 +79,108 @@ func Init(module *bpf.Module) (Probes, error) {
...
 	}
 
 	return p.probes[probe].detach()
 }
 
-// ProbeType is an enum that describes the mechanism used to attach the event
-// Kprobes are explained here: https://github.com/iovisor/bcc/blob/master/docs/reference_guide.md#1-kp
-// Tracepoints are explained here: https://github.com/iovisor/bcc/blob/master/docs/reference_guide.md#3-tracep
-// Raw tracepoints are explained here: https://github.com/iovisor/bcc/blob/master/docs/reference_guide.md#7-raw-tracep
+// DetachAll detaches the probe corresponding to the probe's handle.
```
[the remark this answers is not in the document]
rafaeldtinoco: Yep, thanks. It was left like that during development. Fixing.

pkg/ebpf/probes/probes.go
```diff
@@ -81,52 +79,108 @@ func Init(module *bpf.Module) (Probes, error) {
...
+		err := probe.detach()
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// DisableNetAutoLoad disables eBPF prog auto load (and attach) for net probes
```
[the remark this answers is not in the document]
rafaeldtinoco: true.

pkg/ebpf/probes/probes.go
```diff
@@ -81,52 +79,108 @@ func Init(module *bpf.Module) (Probes, error) {
...
+func (p *probes) DisableNetAutoload() error {
+	p.loadNetwork = false
+
+	for handle, probe := range p.probes {
+		if int32(handle) >= 1000 { // network probes only
```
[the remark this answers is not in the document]
rafaeldtinoco: > We needed to disable network probes because some of the programs will require CAP_NET_ADMIN, and we don't want to force the user to have this capability if he doesn't need to use any network feature of tracee.
> Thinking about it, only the two tc programs we have require this capability, as the others are just kprobes.

But we are loading all eBPF programs by default now, no ? I know we can change that, I was just reflecting what we are doing in a refactor.

We're attaching them based on the dependencies, but we're loading them all. That is why I had the disable autoload for network probes (to accommodate what you said here). To avoid loading the network probes by default.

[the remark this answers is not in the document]
rafaeldtinoco: > * I think that instead of looping all the probes, we can just disable the autoload for these two programs.

Ah, ok. I hope CAP_NET_ADMIN is not needed then.

> * If you choose to keep this code, I think it will be more clear to define "1000" as a constant. Also, it is prone to future bugs to have ">= 1000", as in the future someone may add handles starting from, e.g. 2000, and it will be disabled as well.

It sounds good to me if CAP_NET_ADMIN isn't needed. Actually it makes the network probes behave like regular probes, which I'm all in for.

pkg/ebpf/probes/probes.go
```diff
@@ -147,28 +201,48 @@ func (p *probe) attach(module *bpf.Module) error {
...
+
 	return err
 }
 
+// disableAutoload disables eBPF program automatic loading for given probe
+func (p *probe) disableAutoload(module *bpf.Module) error {
+
+	if p.link != nil { // already attached
+		return nil
+	}
```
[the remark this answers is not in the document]
rafaeldtinoco: true =D.

pkg/ebpf/probes/probes.go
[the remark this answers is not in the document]
rafaeldtinoco: I found this: https://github.com/uber-go/guide/blob/master/style.md but its huge and full of things we don't need. At the end they stick with golangci linter (which I already explored), I'll check that later. Let me know if there is a big blocker for you and I'll change styling.

pkg/ebpf/probes/probes.go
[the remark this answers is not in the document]
rafaeldtinoco: Okay, your idea was explained bellow saying that we could only autoload them all (but the tcProbes). I'll try to propose something else then.

pkg/ebpf/probes/probes.go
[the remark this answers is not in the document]
rafaeldtinoco: Yep, so you're right: Dropping net_admin capability I'm still able to load all probes (but not tcProbes):

If network filters are used:

image

If not (but all regular network probes are still loaded):

image

pkg/ebpf/probes/probes.go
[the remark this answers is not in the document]
rafaeldtinoco: Ill propose something in this PR.

pkg/ebpf/probes/tc_probes.go
```diff
@@ -0,0 +1,222 @@
...
+func (p *tcProbe) detach() error {
+	var err error
+
+	if p.tcHook == nil {
+		return fmt.Errorf("tcProbe isn't attached")
+	}
+
+	// TODO: https://github.com/aquasecurity/tracee/issues/1828
+	// 1. entire clsact qdisc will be destroyed after this
+	// 2. Tc detach does not work and destroy needs to be used
```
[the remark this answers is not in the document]
rafaeldtinoco: It wasn't clear to me what it meant ("this check" wording, and 2 destroy
functions called made me confused). I had to re-read the mailing list
thread, and took sometime to find your email and answers related to it
(and see that your last email did not even had an answer upstream).

I prefer keeping like this if you don't mind, but I can add missing
information to that list if you think its appropriate.

IMO I would add anything else we judge important for the bug resolution,
including the huge mailing list URLs, in the issue I created (actually I would
even say its better to always create issues for things like that, and have a
quick explanation in the code pointing the URL).

pkg/ebpf/probes/tc_probes.go
```diff
@@ -0,0 +1,222 @@
...
+	// tcOpts := bpf.TcOpts{}
+	// err = p.tcHook.Query(&tcOpts)
+	// if err != nil {
+	// 	return err
+	// }
+
+	// err = p.tcHook.Detach(&tcOpts)
+	// if err != nil {
+	// 	return err
+	// }
```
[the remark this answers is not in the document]
rafaeldtinoco: I wanted to show we should query tcopts, detach and destroy (if the API was fully working). I'll
move this to the case (and I do know that libbpf calls detach when destroy is called, either way
I tried detach before destroying to see if API was working.

pkg/ebpf/probes/tc_probes.go
```diff
@@ -0,0 +1,222 @@
...
+// disableAutoload disables a tcProbe's program autoload when eBPF obj is loaded
+func (p *tcProbe) disableAutoload(module *bpf.Module) error {
+
+	prog, _ := module.GetProgram(p.programName)
+	if prog == nil {
+		return fmt.Errorf("could not find %s program", p.programName)
+	}
+
+	return prog.SetAutoload(false)
+}
```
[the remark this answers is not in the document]
rafaeldtinoco: It is, I'll move tcProbes into probes only like suggested.

pkg/ebpf/probes/tc_probes.go
```diff
@@ -0,0 +1,222 @@
...
+
+	p.addInterface(netIface)
+
+	for _, tcProbe := range p.tcProbes[netIface] {
+		err := tcProbe.attach(p.module, netIface)
+		if err != nil {
+			return err
+		}
+		if tcProbe.tcHook == nil {
+			fmt.Printf("HELL NO\n")
```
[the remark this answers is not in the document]
rafaeldtinoco: well, clearly a leftover from debugging =). fixing it.

rafaeldtinoco:
Moved to draft, still working in bringing tcProbes into Probes interface and declaring tcProbes as regular probe Dependencies (To all selected interfaces in that case).

rafaeldtinoco:
@NDStrahilevitz and I think probe should be an interface and both tracing probe and tc probe should implement it (it will be much cleaner than what it became by having both together). I had done that before this attempt but had doubts but after the result of not having an interface, and after a ping of him saying the same thing, I'll do that approach again).

rafaeldtinoco:
@NDStrahilevitz ok, so probes and tcprobes are part of the same package now. I have created 2 interfaces (Probes and Probe) and implemented Probe with "traceProbe" and "tcProbe".

pkg/ebpf/probes/probes.go
```diff
@@ -2,178 +2,503 @@ package probes
...
+		probes: allProbes,
 		module: module,
 	}, nil
 }
 
-//Attach the probe corresponding to the probe's handle
-func (p *probes) Attach(probe Handle) error {
-	if p.probes[probe] == nil {
-		return fmt.Errorf("failed to attach probe: doesn't exist")
+func (p *probes) Attach(args ...interface{}) error {
```
rafaeldtinoco, line 139: image

image

image

pkg/ebpf/probes/probes.go
```diff
@@ -2,178 +2,503 @@ package probes
...
+	return p.attachOrDetach(true, args...)
+}
+
+func (p *probes) Detach(args ...interface{}) error {
+	return p.attachOrDetach(false, args...)
+}
+
+func (p *probes) attachOrDetach(attach bool, args ...interface{}) error {
```
rafaeldtinoco, line 147: There was a discussion between Nadav and I about variadic functions being used here but I used it on purpose so the API could accept any argument (depending on existing and future probe types we may have).

I explained how to use the API here: https://github.com/aquasecurity/tracee/pull/1820/files#diff-a4767f3b581014c47b886c59c79fcf8374aee0edece50c13d6e9758df78abf53R12 and, despite preferences on implementation, the caller won't be affected by this decision.

For now user will attach a probe or detach it. User will also attach a tcProbe to an interface or detach it (and the API supports both, which is good enough and much better than what we currently have).

I'll the one using this interface to implement dynamic interface attachment/detachment and, if it does not serve its purpose, I could rethink it but, for now, I think it is good.

rafaeldtinoco:
I believe this is good for now, considering I'll be the one using this API for future work. Please let me know if there is a blocker or if we can go ahead and merge this for 0.7.1. Thanks!

pkg/ebpf/probes/probes.go
```diff
@@ -2,178 +2,503 @@ package probes
...
+	kretprobe            // github.com/iovisor/bcc/blob/master/docs/reference_guide.md#1-kp
+	tracepoint           // github.com/iovisor/bcc/blob/master/docs/reference_guide.md#3-tracep
+	rawTracepoint        // github.com/iovisor/bcc/blob/master/docs/reference_guide.md#7-raw-tracep
 )
 
-//internal attach function - takes *bpf.Module the from the owning probes struct as a parameter
-func (p *probe) attach(module *bpf.Module) error {
-	if p.link != nil {
-		return nil
+type Probe interface {
```
NDStrahilevitz, line 218: So this interface isn't the one you actually used for `traceProbe` and `tcProbe`, which DO have the same function signatures, just not the ones in the interface here.
I suggest we change the interface to fit what's actually used so we can use it in the `Probes` map, which could also remove a lot of the type casts (i've marked one such place with a review comment for example) making the code more readable.

So it should be (in pseudocode):
```
attach(args...)
detach(args...)
autoload(bool)
```

pkg/ebpf/probes/probes.go
```diff
@@ -2,178 +2,503 @@ package probes
...
 
-//Detach the probe corresponding to the probe's handle
-func (p *probes) Detach(probe Handle) error {
-	if p.probes[probe] == nil {
-		return fmt.Errorf("failed to detach probe: doesn't exist")
+func (p *probes) DetachAll() error {
+	var err error
+
+	for _, in := range p.probes {
+		switch i := in.(type) {
```
NDStrahilevitz, line 188: For example: this type cast could be avoided if both implemented the interface

pkg/ebpf/probes/probes.go
```diff
@@ -2,178 +2,503 @@ package probes
...
-	Detach(probe Handle) error
+	Attach(args ...interface{}) error
+	Detach(args ...interface{}) error
+	DetachAll() error
 }
 
 type probes struct {
-	probes map[Handle]*probe
 	module *bpf.Module
+	probes map[EventHandle]interface{}
```
NDStrahilevitz, line 43: If we fix the interface this could be `map[EventHandle]probe`

pkg/ebpf/probes/probes.go
```diff
@@ -2,178 +2,503 @@ package probes
...
+type traceProbe struct {
+	probeType   probeType
+	eventName   string
+	programName string
+	netProbe    bool
+	bpfLink     *bpf.BPFLink
+}
+
+// attach attaches an eBPF program to its probe
+func (p *traceProbe) attach(args ...interface{}) error {
```
NDStrahilevitz, line 237: so the in interface it should be:
`attach(args ...interface{})` and not `attach(module)`, same for detach and autoload.

pkg/ebpf/probes/probes.go
```diff
@@ -2,178 +2,503 @@ package probes
...
+
+type tcProbe struct {
+	programName   string
+	tcHooks       map[*net.Interface]*bpf.TcHook
+	tcAttachPoint bpf.TcAttachPoint
+	skipLoopback  bool
+}
+
+// attach attaches an eBPF program to its probe
+func (p *tcProbe) attach(args ...interface{}) error {
```
NDStrahilevitz, line 321: ditto

pkg/ebpf/probes/probes.go
```diff
@@ -2,178 +2,503 @@ package probes
...
+		if err != nil {
+			return err
+		}
 	}
 
-	p.link = nil
 	return err
 }
 
-type Handle int32
```
NDStrahilevitz, line 173: Not sure why the name change to EventHandle. 
Asking because the package explicitly removes itself from the events, so this was a deliberate naming choice (as in probeHandle -> probes.Handle).

NDStrahilevitz (commented):
So ideas for how to implement aside, I think there is a contradiction in the implementation here - that the singular probes don't implement the `probe` interface but a separate non existing code interface. So the written interface should be what's actually done, then it can be used in the package (i've listed examples).

To make it clear, i'm not suggesting to change how this is implemented now (contrary to what i've done in my PR to you), but to refine how it's done here.

Other than that, LGTM.

pkg/ebpf/tracee.go
```diff
@@ -937,89 +838,41 @@ func (t *Tracee) initBPF() error {
 		BPFObjName:      t.config.BPFObjPath,
 	}
 
+	// Open the eBPF object file (create a new module)
+
```
yanivagman, line 842: nit: why do we keep these extra lines between the comment and the code relevant to the comment? (here and in many other places)

pkg/ebpf/probes/probes.go
yanivagman: Yes, I wonder the same

pkg/ebpf/probes/probes.go
```diff
@@ -2,178 +2,503 @@ package probes
...
+		TCPConnect:                 &traceProbe{eventName: "tcp_connect", probeType: kprobe, programName: "trace_tcp_connect", netProbe: true},
+		ICMPRecv:                   &traceProbe{eventName: "icmp_rcv", probeType: kprobe, programName: "trace_icmp_rcv", netProbe: true},
+		ICMPSend:                   &traceProbe{eventName: "__icmp_send", probeType: kprobe, programName: "trace_icmp_send", netProbe: true},
+		ICMPv6Recv:                 &traceProbe{eventName: "icmpv6_rcv", probeType: kprobe, programName: "trace_icmpv6_rcv", netProbe: true},
+		ICMPv6Send:                 &traceProbe{eventName: "icmp6_send", probeType: kprobe, programName: "trace_icmp6_send", netProbe: true},
+		Pingv4Sendmsg:              &traceProbe{eventName: "ping_v4_sendmsg", probeType: kprobe, programName: "trace_ping_v4_sendmsg"},
+		Pingv6Sendmsg:              &traceProbe{eventName: "ping_v6_sendmsg", probeType: kprobe, programName: "trace_ping_v6_sendmsg"},
+		DefaultTcIngress:           &tcProbe{programName: "tc_ingress", tcAttachPoint: bpf.BPFTcIngress},
```
yanivagman, line 115: I think these two can be combined into one "probe".
Just use the same fields as we used to have for probes and probeType will be either tcIngress or tcEgress for tc programs.

yanivagman (commented):
See my comment about the fields usage. I think traceProbe and tcProbe should be combined into one "probe" struct as we used to have, and then the interface can be removed. WDYT?

pkg/ebpf/probes/probes.go
yanivagman: I mean to combine traceProbe and tcProbe to probe, as it was before. Meaning, to adjust the tc probes to fit the existing probe struct.
No need for direction field. Can simply be probeType (the field name we currently have for probe), with either tcIngress or tcEgress for the tc programs.

yanivagman:
> > See my comment about the fields usage. I think traceProbe and tcProbe should be combined into one "probe" struct as we used to have, and then the interface can be removed. WDYT?
> 
> Why remove the interface? The code splitting is completely natural here with same function signature and different internal logic. The interface should be made usable (simple line changes in the interface as i've shown), not discarded. In fact the traceProbes should probably be split as well in the future to enable particular capabilities for them (for example the uprobe trigger).

As a future step, I agree, but for now, we already have kprobe, tracepoint, and syscall with different logic to attach the probes, so the natural thing to do as a first step is to simply add the tc probes logic to the existing switch case. Like you say, if we create an interface, then the logical thing to do is to make kprobe, tracepoint and syscall part of this interface as well, but this can be done in a future PR together with tc, or otherwise I think it is confusing

rafaeldtinoco:
> > > See my comment about the fields usage. I think traceProbe and tcProbe should be combined into one "probe" struct as we used to have, and then the interface can be removed. WDYT?
> > 
> > 
> > Why remove the interface? The code splitting is completely natural here with same function signature and different internal logic. The interface should be made usable (simple line changes in the interface as i've shown), not discarded. In fact the traceProbes should probably be split as well in the future to enable particular capabilities for them (for example the uprobe trigger).
> 
> As a future step, I agree, but for now, we already have kprobe, tracepoint, and syscall with different logic to attach the probes, so the natural thing to do as a first step is to simply add the tc probes logic to the existing switch case. Like you say, if we create an interface, then the logical thing to do is to make kprobe, tracepoint and syscall part of this interface as well, but this can be done in a future PR together with tc, or otherwise I think it is confusing

I like having them split, like first suggested by Nadav, because tcProbes and tracing Probes are different in nature, but both are "Probes" (i dont like this name by the way, it is misleading since "tc" programs have attach points and hooks). 

The tcProbes need to be attached to interface, need to be detached and attached on-demand, etc (which is something I've considered for my current code I believe).

When coding, tracing probes are very similar, and different than tcprobes, thus the classification in 2 different "Probe" implementations. If we think ahead, we will have the following "Probe" types:

- traceProbes (TYPE_KPROBE, TYPE_TRACEPOINT,, RAW_TRACEPOINT)
- tcProbes (QDISC CLASSIFIERS)
- socketProbes (SOCKET_FILTER, SOCK_OPS, SK_SKB, SK_MSG)
- xdpProbes (express data path)
- tunnelProbes (LWT_IN, LWT_OUT, LWT_XMIT)
- cgroupProbes (CGROUP_SKB, CGROUP_SOCK, CGROUP_DEVICE, CGROUP_SOCKADDR, CGROUP_SYSCTL, CGROUP_SOCKOPT)

and more...

I don't want to go back now to go forward in a near future. I believe addressing concerns from Nadav related to interface implementation are the best next step

If you allow me, I think we should address Nadavs concerns on interface, merge this, allow me to do some more coding regarding network improvements, see if this has a fit or need improvements, etc.

Many of our concerns on classification (for what the interface should look like) will take shape after other probe types appear and we have to implement specific (to their types) functionalities.

pkg/ebpf/tracee.go
rafaeldtinoco: styling preferences. we should start enforcing one, clearly my preferences are different than yours, and different from Nadavs (who does comments like //something). =(

pkg/ebpf/probes/probes.go
rafaeldtinoco: image

rafaeldtinoco:
@NDStrahilevitz and I just talked about this one:

- I'm going to fix Probe interface implementation (so it is actually implemented by traceProbe and tcProbe)

- I'll add a fixed "Handle" argument in the methods so it is not part of the variadic "args" approach.

- I'll keep the variadic approach until we have automatic attachment/detachment from network interfaces implemented: this will give us a better idea if we should implement specific arguments or keep the interface generic for future probe types (xdp, cgroups, etc). Then we can make it less generic (variadic) with real usage.

pkg/ebpf/probes/probes.go
rafaeldtinoco: https://github.com/aquasecurity/tracee/pull/1820#issuecomment-1167207980

pkg/ebpf/probes/probes.go
```diff
@@ -2,174 +2,486 @@ package probes
...
+			netIface = a
+		}
+	}
+
+	if _, ok := p.probes[handle]; !ok {
+		return fmt.Errorf("incorrect event handle provided")
 	}
 
-	return p.probes[probe].attach(p.module)
+	switch b := p.probes[handle].(type) {
```
NDStrahilevitz, line 162: So since we use (args ...interface{}) Do we actually need this switch? Internally the functions should interpret if and what args are relevant no? I can see leaving it as a safety measure though so up to you.

pkg/ebpf/probes/probes.go
```diff
@@ -2,174 +2,486 @@ package probes
...
+		Pingv4Sendmsg:              &traceProbe{eventName: "ping_v4_sendmsg", probeType: kprobe, programName: "trace_ping_v4_sendmsg"},
+		Pingv6Sendmsg:              &traceProbe{eventName: "ping_v6_sendmsg", probeType: kprobe, programName: "trace_ping_v6_sendmsg"},
+		DefaultTcIngress:           &tcProbe{programName: "tc_ingress", tcAttachPoint: bpf.BPFTcIngress},
+		DefaultTcEgress:            &tcProbe{programName: "tc_egress", tcAttachPoint: bpf.BPFTcEgress, skipLoopback: true},
+	}
+
+	// disable autoload for network related eBPF programs in network is disabled
+	if !netEnabled {
+		for _, in := range allProbes {
+			switch i := in.(type) {
```
NDStrahilevitz, line 127: Did you remove the part of disabling the network traceProbes on purpose? I remember it being here before.

pkg/ebpf/probes/probes.go
```diff
@@ -2,174 +2,486 @@ package probes
...
+}
+
+func (p *probes) Detach(handle Handle, args ...interface{}) error {
+	return p.attachOrDetach(handle, false, args...)
+}
+
+func (p *probes) attachOrDetach(handle Handle, attach bool, args ...interface{}) error {
+	var netIface *net.Interface
+
+	for _, arg := range args {
```
NDStrahilevitz, line 151: So if we go by just sending the args forward regardless of concrete probe type, there's no reason to do these conversions, just send the args forward and interpretation will occur internally in the probe.

NDStrahilevitz (approved):
Left some final style questions anyway, but overall LGTM!

pkg/ebpf/probes/probes.go
rafaeldtinoco: I did and there was a comment from Yaniv to have only tc disabled by default in the beginning.

rafaeldtinoco:
> Left some final style questions anyway, but overall LGTM!

Yep, I have addressed yours and Yaniv's last comments, I think its much better and cleaner now. A final +1 before merging if nothing else. Thanks for all the reviews.

pkg/ebpf/probes/probes.go
```diff
@@ -2,175 +2,470 @@ package probes
...
+
+	return err
+}
+
+//
+// common function(s) to Probe implementations
+//
+
+// enableDisableAutoload enables or disables an eBPF program autoload setting
+func enableDisableAutoload(args ...interface{}) error {
```
NDStrahilevitz, line 438: Last minute catch, but from reading this function it doesn't currently need the variadic args, maybe we can go back to static types here? Or will there be future probes/hooks where it will be required?

NDStrahilevitz (commented):
So I caught one thing last minute, will +1 after, but i'm fine with going ahead with this anyway.

pkg/ebpf/probes/probes.go
rafaeldtinoco: This one could have static types, I think it will be better!

rafaeldtinoco:
@yanivagman could I get from you a +1 to merge ? Thanks!

yanivagman (approved):
LGTM after the changes. Thanks

Issue #1821, opened 2022-06-14
[REFAC]  Export derived event implementations to a package


AlonZivony:
## Prerequisites

Select one OR another:

- [ ] I have discussed the refactoring idea with one (or another) maintainer.
- [ ] I'll create a PR to implement this refactoring idea (assign to yourself).
- [ ] Someone else should implement this (describe it well).

## Refactoring description

This refactoring is about creating more freedom and making the derived events more manageable.
Currently all derived events are implemented inside the _ebpf_ package, in a single file.
Because the use in derived events is only growing, creating a way to manage them in separate files will be crucial.

PR #1822, opened 2022-06-14
Refactor/derived events package

AlonZivony:
## Initial Checklist

## Description (git log)

Create a new package for implementing new derived events.

df18d24e pkg/ebpf: extract events params parsing functions to package
87ed207e pkg/epbf: generate multiple derived events at once
75f982c7 pkg/ebpf: create interface for derived events in new package

Fixes: #1821

## Type of change

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/ebpf/events_derived.go
+++ b/pkg/ebpf/events_derived.go
@@ -3,11 +3,9 @@ package ebpf
 import (
 	"context"
 	"fmt"
-	"strings"
 
-	"github.com/aquasecurity/libbpfgo/helpers"
-	"github.com/aquasecurity/tracee/pkg/containers"
 	"github.com/aquasecurity/tracee/pkg/events"
+	"github.com/aquasecurity/tracee/pkg/events/derive"
 	"github.com/aquasecurity/tracee/types/trace"
 )
 
@@ -24,37 +22,37 @@ func (t *Tracee) initDerivationTable() error {
 		events.CgroupMkdir: {
 			events.ContainerCreate: {
 				Enabled:  t.events[events.ContainerCreate].submit,
-				Function: deriveContainerCreate(t.containers),
+				Function: derive.ContainerCreate(t.containers),
 			},
 		},
 		events.CgroupRmdir: {
 			events.ContainerRemove: {
 				Enabled:  t.events[events.ContainerRemove].submit,
-				Function: deriveContainerRemoved(t.containers),
+				Function: derive.ContainerRemoved(t.containers),
 			},
 		},
 		events.PrintSyscallTable: {
 			events.HookedSyscalls: {
 				Enabled:  t.events[events.PrintSyscallTable].submit,
-				Function: deriveDetectHookedSyscall(t.kernelSymbols),
+				Function: derive.DetectHookedSyscall(t.kernelSymbols),
 			},
 		},
 		events.DnsRequest: {
 			events.NetPacket: {
 				Enabled:  t.events[events.NetPacket].submit,
-				Function: deriveNetPacket(),
+				Function: derive.NetPacket(),
 			},
 		},
 		events.DnsResponse: {
 			events.NetPacket: {
 				Enabled:  t.events[events.NetPacket].submit,
-				Function: deriveNetPacket(),
+				Function: derive.NetPacket(),
 			},
 		},
 		events.PrintNetSeqOps: {
 			events.HookedSeqOps: {
 				Enabled:  t.events[events.HookedSeqOps].submit,
-				Function: deriveHookedSeqOps(t.kernelSymbols),
+				Function: derive.HookedSeqOps(t.kernelSymbols),
 			},
 		},
 	}
@@ -76,7 +74,7 @@ func (t *Tracee) deriveEvents(ctx context.Context, in <-chan *trace.Event) (<-ch
 			case event := <-in:
 				out <- event
 
-				// Derive event before parsing its arguments
+				// Derive event before parse its arguments
 				derivatives, errors := events.Derive(*event, t.eventDerivations)
 
 				for _, err := range errors {
@@ -94,199 +92,4 @@ func (t *Tracee) deriveEvents(ctx context.Context, in <-chan *trace.Event) (<-ch
 	}()
 
 	return out, errc
-}
-
-/*
-* Derivation functions:
-* Most derivation functions take tracee as a closure argument to track it's runtime state
-* Tracee builds its derivation map from these functions and injects itself as an argument to the closures
-* The derivation map is then built with the returned events.DeriveFunction functions, which is used in deriveEvents
- */
-
-// deriveContainerCreate receives a containers as a closure argument to track it's containers.
-// If it receives a cgroup_mkdir event, it can derive a container_create event from it.
-func deriveContainerCreate(containers *containers.Containers) events.DeriveFunction {
-	return func(event trace.Event) (trace.Event, bool, error) {
-		cgroupId, err := getEventArgUint64Val(&event, "cgroup_id")
-		if err != nil {
-			return trace.Event{}, false, err
-		}
-
-		def := events.Definitions.Get(events.ContainerCreate)
-		if info := containers.GetCgroupInfo(cgroupId); info.Container.ContainerId != "" {
-			de := event
-			de.EventID = int(events.ContainerCreate)
-			de.EventName = def.Name
-			de.ReturnValue = 0
-			de.StackAddresses = make([]uint64, 1)
-			de.Args = []trace.Argument{
-				{ArgMeta: def.Params[0], Value: info.Runtime.String()},
-				{ArgMeta: def.Params[1], Value: info.Container.ContainerId},
-				{ArgMeta: def.Params[2], Value: info.Ctime.UnixNano()},
-				{ArgMeta: def.Params[3], Value: info.Container.Image},
-				{ArgMeta: def.Params[4], Value: info.Container.Name},
-				{ArgMeta: def.Params[5], Value: info.Container.Pod.Name},
-				{ArgMeta: def.Params[6], Value: info.Container.Pod.Namespace},
-				{ArgMeta: def.Params[7], Value: info.Container.Pod.UID},
-			}
-			de.ArgsNum = len(de.Args)
-
-			return de, true, nil
-		}
-
-		return trace.Event{}, false, nil
-	}
-}
-
-// deriveContainerRemoved receives a containers.Containers object as a closure argument to track it's containers.
-// If it receives a cgroup_rmdir event, it can derive a container_remove event from it.
-func deriveContainerRemoved(containers *containers.Containers) events.DeriveFunction {
-	return func(event trace.Event) (trace.Event, bool, error) {
-		cgroupId, err := getEventArgUint64Val(&event, "cgroup_id")
-		if err != nil {
-			return trace.Event{}, false, err
-		}
-
-		def := events.Definitions.Get(events.ContainerRemove)
-		if info := containers.GetCgroupInfo(cgroupId); info.Container.ContainerId != "" {
-			de := event
-			de.EventID = int(events.ContainerRemove)
-			de.EventName = def.Name
-			de.ReturnValue = 0
-			de.StackAddresses = make([]uint64, 1)
-			de.Args = []trace.Argument{
-				{ArgMeta: def.Params[0], Value: info.Runtime.String()},
-				{ArgMeta: def.Params[1], Value: info.Container.ContainerId},
-			}
-			de.ArgsNum = len(de.Args)
-
-			return de, true, nil
-		}
-
-		return trace.Event{}, false, nil
-	}
-}
-
-func deriveDetectHookedSyscall(kernelSymbols *helpers.KernelSymbolTable) events.DeriveFunction {
-	return func(event trace.Event) (trace.Event, bool, error) {
-		syscallAddresses, err := getEventArgUlongArrVal(&event, "syscalls_addresses")
-		if err != nil {
-			return trace.Event{}, false, fmt.Errorf("error parsing syscalls_numbers arg: %v", err)
-		}
-		hookedSyscall, err := analyzeHookedAddresses(syscallAddresses, kernelSymbols)
-		if err != nil {
-			return trace.Event{}, false, fmt.Errorf("error parsing analyzing hooked syscalls addresses arg: %v", err)
-		}
-		de := event
-		de.EventID = int(events.HookedSyscalls)
-		de.EventName = "hooked_syscalls"
-		de.ReturnValue = 0
-		de.Args = []trace.Argument{
-			{ArgMeta: trace.ArgMeta{Name: "hooked_syscalls", Type: "[]trace.HookedSymbolData"}, Value: hookedSyscall},
-		}
-		de.ArgsNum = 1
-		return de, true, nil
-	}
-}
-
-// deriveNetPacket derives net_packet from net events with 'metadata' arg
-func deriveNetPacket() events.DeriveFunction {
-	return func(event trace.Event) (trace.Event, bool, error) {
-		metadataArg := events.GetArg(&event, "metadata")
-		if metadataArg == nil {
-			return trace.Event{}, false, fmt.Errorf("couldn't find argument name metadata in event %s", event.EventName)
-		}
-
-		def := events.Definitions.Get(events.NetPacket)
-		de := event
-		de.EventID = int(events.NetPacket)
-		de.EventName = def.Name
-		de.ReturnValue = 0
-		de.Args = []trace.Argument{
-			*metadataArg,
-		}
-		de.ArgsNum = 1
-		return de, true, nil
-	}
-}
-
-func analyzeHookedAddresses(addresses []uint64, kernelSymbols *helpers.KernelSymbolTable) ([]trace.HookedSymbolData, error) {
-	hookedSyscalls := make([]trace.HookedSymbolData, 0)
-	for idx, syscallsAddress := range addresses {
-		InTextSegment, err := kernelSymbols.TextSegmentContains(syscallsAddress)
-		if err != nil {
-			continue
-		}
-		if !InTextSegment {
-			syscallsToCheck := events.SyscallsToCheck()
-			hookingFunction := parseSymbol(syscallsAddress, kernelSymbols)
-			if idx > len(syscallsToCheck) {
-				return nil, fmt.Errorf("syscall inedx out of the syscalls to check list %v", err)
-			}
-			syscallNumber := syscallsToCheck[idx]
-			event, found := events.Definitions.GetSafe(syscallNumber)
-			var hookedSyscallName string
-			if found {
-				hookedSyscallName = event.Name
-			} else {
-				hookedSyscallName = fmt.Sprint(syscallNumber)
-			}
-			hookedSyscalls = append(hookedSyscalls, trace.HookedSymbolData{SymbolName: hookedSyscallName, ModuleOwner: hookingFunction.Owner})
-
-		}
-	}
-	return hookedSyscalls, nil
-}
-
-func parseSymbol(address uint64, table *helpers.KernelSymbolTable) *helpers.KernelSymbol {
-	hookingFunction, err := table.GetSymbolByAddr(address)
-	if err != nil {
-		hookingFunction = &helpers.KernelSymbol{}
-		hookingFunction.Owner = "hidden"
-	}
-	hookingFunction.Owner = strings.TrimPrefix(hookingFunction.Owner, "[")
-	hookingFunction.Owner = strings.TrimSuffix(hookingFunction.Owner, "]")
-	return hookingFunction
-}
-
-var seq_ops_functions = [4]string{
-	"seq_show",
-	"seq_open",
-	"seq_close",
-	"seq_next",
-}
-
-func deriveHookedSeqOps(kernelSymbols *helpers.KernelSymbolTable) events.DeriveFunction {
-	return func(event trace.Event) (trace.Event, bool, error) {
-		seqOpsArr, err := getEventArgUlongArrVal(&event, "net_seq_ops")
-		if err != nil || len(seqOpsArr) < 1 {
-			return trace.Event{}, false, err
-		}
-		seqOpsName := parseSymbol(seqOpsArr[0], kernelSymbols).Name
-		hookedSeqOps := make([]trace.HookedSymbolData, 0)
-		for _, addr := range seqOpsArr[1:] {
-			inTextSegment, err := kernelSymbols.TextSegmentContains(addr)
-			if err != nil {
-				continue
-			}
-			if !inTextSegment {
-				hookingFunction := parseSymbol(addr, kernelSymbols)
-				hookedSeqOps = append(hookedSeqOps, trace.HookedSymbolData{SymbolName: hookingFunction.Name, ModuleOwner: hookingFunction.Owner})
-			}
-		}
-		def := events.Definitions.Get(events.HookedSeqOps)
-		de := event
-		de.EventID = int(events.HookedSeqOps)
-		de.EventName = "hooked_seq_ops"
-		de.ReturnValue = 0
-		de.StackAddresses = make([]uint64, 1)
-		de.Args = []trace.Argument{
-			{ArgMeta: def.Params[0], Value: seqOpsName},
-			{ArgMeta: def.Params[1], Value: hookedSeqOps},
-		}
-		de.ArgsNum = 1
-		return de, true, nil
-
-	}
-
 }
--- a/pkg/events/derive/container_removed.go
+++ b/pkg/events/derive/container_removed.go
@@ -1,0 +1,27 @@
+package derive
+
+import (
+	"github.com/aquasecurity/tracee/pkg/containers"
+	"github.com/aquasecurity/tracee/pkg/events"
+	"github.com/aquasecurity/tracee/pkg/events/parse"
+	"github.com/aquasecurity/tracee/types/trace"
+)
+
+// ContainerRemoved receives a containers.Containers object as a closure argument to track it's containers.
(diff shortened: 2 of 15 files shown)


NDStrahilevitz:
Jumping on this PR just because i've also done this work in #1818, so there may be room for combining the two, WDYT?

AlonZivony:
> Jumping on this PR just because i've also done this work in #1818, so there may be room for combining the two, WDYT?

Though the two PRs share some changes which are required for extracting things to new packages, I think that derived events should have its own package under your new `events` package.
If you agree with me that this is the case, than it means that both of the changes are independent (though they affect each other because the new derived package will move outside `ebpf` with your PR).
I also think that your PR is already big, so I don't think it will make it easier to merge if we combine them...

NDStrahilevitz:
> Though the two PRs share some changes which are required for extracting things to new packages, I think that derived events should have its own package under your new `events` package. If you agree with me that this is the case, than it means that both of the changes are independent (though they affect each other because the new derived package will move outside `ebpf` with your PR). I also think that your PR is already big, so I don't think it will make it easier to merge if we combine them...

By combining them I mean one of the approaches goes to the other PR (or a combination) and then either I drop this part from my PR and rebase on top (which i'm ok with), or that I add your part to my PR and we close this one.
With regards to splitting into a subpackage, I like it, but currently it looks unjustified, since there's actually barely anything inside the subpackage, and it's not used for any significant purpose (ie testing, but maybe you planned on adding that?). I'll elaborate further in a review.

pkg/ebpf/events/derived/derived.go
```diff
@@ -0,0 +1,29 @@
+package derived
+
+import "github.com/aquasecurity/tracee/types/trace"
+
+// DeriveFn is a function prototype for a function that receives an event as
+// argument and may produce a new event if relevant.
+// It returns list of derived events, which might be empty depending on successful derivation,
+// a bool indicating if an event was derived, and an error if one occurred.
+type DeriveFn func(trace.Event) ([]trace.Event, bool, error)
```
NDStrahilevitz, line 9: Why would we return a list of derived events? I assume it's for the ability to derive multiple events from one, but currently our mechanism for deriving one events into multiple is declaring multiple derivable events in the derivation table (events_derived.go:18).

So we could drop the map to map idea and use a list of derived events instead, as you do here, but we don't currently have an event that needs multiple derivation, as far as I can tell from the code, and secondly, why would it be better than the solution we currently have (performance, readability, etc.)?

I do see how we could use this to return multiple events of the same derivation, but again, do we actually need such a feature (Is it related to #1823)?

pkg/ebpf/events/derived/derived.go
```diff
@@ -0,0 +1,29 @@
...
+
+// IDerivedEventGenerator is an interface for derived events generator class.
+// The main purpose of it is to create derived events given an event.
+// To generate derived event, the GenerateDerivedFn function should be used with initialized generator.
+type IDerivedEventGenerator interface {
```
NDStrahilevitz, line 14: Do we use this anywhere besides GenerateDerivedFn? If not, where would it be used?

pkg/ebpf/events/derived/derived.go
```diff
@@ -0,0 +1,29 @@
...
+	GenerateEvents(event trace.Event) ([]trace.Event, bool, error)
+}
+
+func GenerateDerivedFn(generator IDerivedEventGenerator) DeriveFn {
```
NDStrahilevitz, line 18: I don't see this being used anywhere, so what is the purpose of this function? Do you plan on using it in the future?

pkg/ebpf/events/derived/derived.go
```diff
@@ -0,0 +1,29 @@
...
+	return func(event trace.Event) ([]trace.Event, bool, error) {
+		return generator.GenerateEvents(event)
+	}
+}
+
+// EventSkeleton is a struct for the necessary information from an event definition to create the derived event
+type EventSkeleton struct {
```
NDStrahilevitz, line 25: Likewise, is this used anywhere, or will be used in the near future?

pkg/events/parsing/params.go
```diff
@@ -0,0 +1,85 @@
+package parsing
```
NDStrahilevitz, line 1: I like this as a subpackage!
Here, i'd only like to suggest that you could slightly change the package's name, and then make the exported function names shorter, for example:
`parseargs.Int32Val` or `parse.EventArgInt32Val`, or anything like that. But either way it's a nice change.

pkg/ebpf/events_derived.go
```diff
@@ -6,20 +6,16 @@ import (
...
-// deriveFn is a function prototype for a function that receives an event as
-// argument and may produce a new event if relevant.
-// It returns a derived or empty event, depending on successful derivation,
-// a bool indicating if an event was derived, and an error if one occurred.
-type deriveFn func(trace.Event) (trace.Event, bool, error)
-
 // Initialize the eventDerivations map.
 // Here we declare for each Event (represented through it's ID)
 // to which other Events it can be derived and the corresponding function to derive into that Event.
 func (t *Tracee) initEventDerivationMap() error {
```
NDStrahilevitz, line 17: So I don't see here anywhere where the change to return multiple events from a derivation was used. So i'm not sure what the point of the change was.

pkg/ebpf/events/derived/derived.go
```diff
@@ -0,0 +1,29 @@
+package derived
```
NDStrahilevitz, line 1: So I like having it as a subpackage, but there's barely anything here. 
The only exported symbol that's used from here is the DeriveFn, I don't see how that's worthy of it's own package. If more stuff was moved here, or the other exported symbols would be actually used, then it would be much more justified (which it could be).

NDStrahilevitz (changes requested):
I like the subpackage splitting, but currently, I feel some of the mechanism changes in the PR aren't actually used for anything, which is my main gripe with it. 
Unless it's shown we actually use the interfaces, generate function and event list from deriving somewhere, I'm not sure why these changes were included in the PR.

In addition, I think adding tests to the PR would improve it, and possibly justify some of the new stuff (although I think it can be tested on it's own without any of them).

AlonZivony:
>I like the subpackage splitting, but currently, I feel some of the mechanism changes in the PR aren't actually used for anything, which is my main gripe with it. Unless it's shown we actually use the interfaces, generate function and event list from deriving somewhere, I'm not sure why these changes were included in the PR.

So if you will look at the issue #1823 you will see that this PR is part of the roadmap for new derived events I want to implement, which has much more logic than current ones so I wanted to put them in different files (so as a result - I created a new package for it)

pkg/ebpf/events/derived/derived.go
AlonZivony: The current structure enable me to create a single derived event from a single event.
I can create multiple events from multiple events, but I can't create for a single `cgroup_makedir` event more than 1 `container_create` event. In my case, I want to create for a single event of SO loading multiple events of collisions with previous loaded SOs. The current structure just won't allow it, and for no good reason.

pkg/ebpf/events/derived/derived.go
AlonZivony: No, I didn't plan for it to be used anywhere except for `GenerateDerivedFn`.
But this is the idea, I want 1 structure for creating derivation function to make the code using it more readable.

However, I think we can also make the `eventsDerivation` map use the interface instead of the function.
This will make it possible to have state of single derived event generator that use multiple base events for its derivation.
So I propose using a map of the following structure - `map[int32]map[int32]IDerivedEventGenerator`.
Afterwards we will call the `GenerateDerivedFn` function for each struct and create the function with state between different triggering event. WDYT?

pkg/ebpf/events/derived/derived.go
AlonZivony: Yea, I plan that each derived event implementation will use it for creating the current `DerivedFn`.

NDStrahilevitz:
> So if you will look at the issue #1823 you will see that this PR is part of the roadmap for new derived events I want to implement, which has much more logic than current ones so I wanted to put them in different files (so as a result - I created a new package for it). I will add that I don't think that creating a good infrastructure require that it will be used immediately. If you agree that the future derived events should be implemented using the interface I created in a different package I think it proves the PR reasoning.

I assumed the infrastructure has reasons to exist, but without seeing them used, I, or any other reviewer for that matter, can't really review if it's a good approach or if it's actually needed infrastructure. 
For example, I don't see the purpose of the interface, maybe in later PRs for #1823 it will be useful, but then you can just add that in those PRs, until then it's cluttering noise, that could potentially be used "wrong" beforehands. Same goes for returning a list instead of a single event.
In addition, if the subpackage was created to host new derived functions (which I think is a good idea, I had a PR previously where I split the derived functions into their own file, Yaniv rejected that idea because there weren't many of those functions  at the time), why are the deriving functions not in the package now? 
So we could have a package with the deriving infrastructure, the deriving functions, and tests for all of the above! But currently, that isn't the case, so the package is unjustified, besides cleaning up the `ebpf` package which is obviously a good thing.

pkg/ebpf/events/derived/derived.go
NDStrahilevitz: Alright, so it is related to #1823, I can go with that, but could you also document at the comment that a DeriveFn should return events of a singular type?

pkg/ebpf/events/derived/derived.go
NDStrahilevitz: Well, that's the purpose of the DeriveFn, to be the interface. 
If you need some external dependency, we have an existing pattern to support that (we use that with tracee for example for the container events).

It could be that what you suggest would be more readable, but we can't actually see that if we don't write it out. So if you could maybe rewrite one of the derivation functions to use the IDerivedEventGenerator, that may give an idea if this would make it actually more readable.

pkg/ebpf/events/derived/derived.go
NDStrahilevitz: So ditto to above, if we want to know if this will be more readable you need to write out an example to compare against.

pkg/ebpf/events/derived/derived.go
NDStrahilevitz: Also, I think we could use this to reduce the api from ([]trace.Event, bool, err) to just ([]trace.Event, []error). I initially included the bool to indicate that while no error occurred in the derivation, the event was irrelevant and empty, so it shouldn't be propagated.
But now we could just check for `len(derivatives) != 0`, or even just loop over the slice without such a check (if no error occured).
However, because we can derive multiple events at once now, this should now return a `[]error`, since we can have multiple derivation errors.

pkg/ebpf/events/derive/derive.go
```diff
@@ -0,0 +1,61 @@
+package derive
```
NDStrahilevitz, line 1: Shouldn't this be in `events/derive` not `ebpf/events/derive`?

pkg/ebpf/events/derive/derive.go
```diff
@@ -0,0 +1,61 @@
...
+			derivatives = append(derivatives, derivative...)
+		}
+	}
+
+	return derivatives, errors
+}
+
+// NewEvent create a new derived event from given event values, adjusted by the derived event skeleton meta-data.
+// This method enables using the context of the base event, but with the new arguments and meta-data of the derived one.
+func NewEvent(baseEvent *trace.Event, skeleton EventSkeleton, argsValues []interface{}) (trace.Event, error) {
```
NDStrahilevitz, line 46: Nice! But why pass the event by reference and not value?

pkg/ebpf/events_derived.go
```diff
@@ -99,16 +79,16 @@ func (t *Tracee) deriveEvents(ctx context.Context, in <-chan *trace.Event) (<-ch
...
-		cgroupId, err := getEventArgUint64Val(&event, "cgroup_id")
+func deriveContainerCreate(t *Tracee) derive.DeriveFunction {
+	return func(event trace.Event) ([]trace.Event, error) {
+		cgroupId, err := parsing.GetEventArgUint64Val(&event, "cgroup_id")
 		if err != nil {
-			return trace.Event{}, false, err
+			return []trace.Event{}, err
 		}
 
 		def := EventsDefinitions[ContainerCreateEventID]
```
NDStrahilevitz, line 94: Shouldn't it be `generateEventSkeleton(ContainerCreateEventID)` instead?

pkg/events/parsing/params.go
AlonZivony: The convention I know we agreed on is that packages names should be the verb of what they are doing, and the files names should be the noun of the operation (for example, we have the `sorting` package and within the `sorter.go` file which export the a sorter type).
I agree with you that the 'GetEvent' part is redundant, because the package name is events/parsing.

pkg/ebpf/events/derive/derive.go
AlonZivony: Just to avoid futile copying of the event, because it is a large struct.

pkg/events/parsing/params.go
NDStrahilevitz: Fair enough. Shortening the names is the main benefit either way.

NDStrahilevitz (approved):
LGTM! @rafaeldtinoco @yanivagman can one of you please confirm?

AlonZivony:
Didn't change anything with the push, tried to give Nadav a co-author for the last commit but have a bug with it for some reason.

NDStrahilevitz:
> Didn't change anything with the push, tried to give Nadav a co-author for the last commit but have a bug with it for some reason.

Thanks ☺️! Maybe i'm misinterpreting the docs but I think you need to add a <> around my email.

rafaeldtinoco:
@NDStrahilevitz so what is the take about the changes @AlonZivony did to derived events ? Did you include his needs in https://github.com/aquasecurity/tracee/pull/1818 ?

I ask because that PR is already +1, just trying to avoid re-work and to close PRs if they were reworked.

Thanks!

rafaeldtinoco (changes requested):
I'm requesting changes just so the review is -1 (check previous comment please).

NDStrahilevitz:
> @NDStrahilevitz so what is the take about the changes @AlonZivony did to derived events ? Did you include his needs in #1818 ? Is there anything left to be done in this PR once that one gets merged (rebasing this and adding other needs) ?
> 
> I ask because that PR is already +1, just trying to avoid re-work and to close PRs if they were reworked.
> 
> Thanks!

So there are some functional changes in this PR that were not included in #1818, mainly the option to derive multiple events at once, and the EventSkeleton mechanism (although with the changes to EventDefinitions this may not be necessary now). 
It may be that this PR will now need to do less, or still do most of what it has before. @AlonZivony you should rebase and see what you still need to add, but I believe the PR can probably be much smaller. Or, just include the changes with the relevant features, your call though.

pkg/ebpf/events_derived.go
```diff
@@ -77,16 +79,19 @@ func (t *Tracee) deriveEvents(ctx context.Context, in <-chan *trace.Event) (<-ch
...
 				// Derive event before parsing its arguments
-				derivatives, errors := events.Derive(*event, t.eventDerivations)
+				derivatives, errors := derive.Derive(*event, t.eventDerivations)
 
 				for _, err := range errors {
 					t.handleError(err)
 				}
 
 				for _, derivative := range derivatives {
+					// Drop derived events which were not requested by the user
```
NDStrahilevitz, line 89: This is not needed anymore, the `Enabled` field in the derivation table takes care of that.

pkg/ebpf/net_events.go
```diff
@@ -255,13 +256,17 @@ func (t *Tracee) processNetEvents(ctx gocontext.Context) {
...
 					// derive events chosen by the user
-					derivatives, errors := events.Derive(evt, t.eventDerivations)
+					derivatives, errors := derive.Derive(evt, t.eventDerivations)
 
 					for _, err := range errors {
 						t.handleError(err)
 					}
 
 					for _, derivative := range derivatives {
+						// Drop derived events which were not requested by the user
```
NDStrahilevitz, line 266: ditto

pkg/ebpf/net_events.go
```diff
@@ -270,6 +275,9 @@ func (t *Tracee) processNetEvents(ctx gocontext.Context) {
 							return
 						}
 					}
+					for _, err := range errors {
```
NDStrahilevitz, line 278: duplicate of line 261 here

pkg/events/derive/derive.go
```diff
@@ -0,0 +1,78 @@
...
+)
+
+// DeriveFunction is a function prototype for a function that receives an event as
+// argument and may produce a new event if relevant.
+// It returns a derived or empty event, depending on successful derivation,
+// and an error if one occurred.
+type DeriveFunction func(trace.Event) ([]trace.Event, []error)
+
+// EventSkeleton is a struct for the necessary information from an event definition to create the derived event
+type EventSkeleton struct {
```
NDStrahilevitz, line 16: So if you just keep `derive` inside the events package I don't think you need all of this.
Just use the `event` struct instead as the skeleton, because that literally is what the event struct is.

pkg/events/derive/derive.go
```diff
@@ -0,0 +1,78 @@
...
+// as defined by tracee's eventDerivations map.
+// The map is initialized in the above function
+func Derive(event trace.Event, derivationTable DerivationTable) ([]trace.Event, []error) {
+	derivatives := []trace.Event{}
+	errors := []error{}
+	deriveFns := derivationTable[events.ID(event.EventID)]
+	for id, deriveFn := range deriveFns {
+		if deriveFn.Enabled {
+			derivative, errs := deriveFn.Function(event)
+			if len(errs) > 0 {
```
NDStrahilevitz, line 38: no need to length check, if 0 the loop will skip

pkg/events/derive/derive.go
```diff
@@ -0,0 +1,78 @@
...
+				for _, err := range errs {
+					errors = append(errors, fmt.Errorf("failed to derive event %d: %v", id, err))
+				}
+			}
+			if len(derivative) > 0 {
```
NDStrahilevitz, line 43: ditto, append will just not do anything if the spread in on an empty slice.

pkg/events/derive/derive.go
```diff
@@ -0,0 +1,78 @@
+package derive
```
NDStrahilevitz, line 1: Again, I suggest to rethink if this actually needs to be a package now.
I'll suggest an alternative: add a package for the deriving functions themselves inside ebpf where the pipeline is. There you can move all the derivation functions from `events_derived.go`. That will also accomplish some needed cleanup, and that can consume the events package for the derivation interfaces.

pkg/events/derive/derive_test.go
```diff
@@ -0,0 +1,79 @@
...
+	return fmt.Errorf("failed to derive event %d: %s", id, errMsg)
+}
+
+func Test_DeriveEvent(t *testing.T) {
+	alwaysDeriveError := func() DeriveFunction {
+		return func(e trace.Event) ([]trace.Event, []error) {
+			return []trace.Event{}, []error{fmt.Errorf("derive error")}
+		}
+	}
+	testEventID := events.ID(1)
```
NDStrahilevitz, line 22: nit: declare these as consts at the top?

But I like this change to the test.

pkg/events/parsing/params.go
NDStrahilevitz: I actually want to go back on this. `parse` is also a verb, and, you named the other package `derive` on the same scale, not `deriving`. Let's go for the simple word.

pkg/events/derive/derive.go
```diff
@@ -0,0 +1,78 @@
...
+	"fmt"
+	"github.com/aquasecurity/tracee/pkg/events"
+	"github.com/aquasecurity/tracee/types/trace"
+)
+
+// DeriveFunction is a function prototype for a function that receives an event as
+// argument and may produce a new event if relevant.
+// It returns a derived or empty event, depending on successful derivation,
+// and an error if one occurred.
+type DeriveFunction func(trace.Event) ([]trace.Event, []error)
```
NDStrahilevitz, line 13: But of course we still need this change.

NDStrahilevitz (changes requested):
So I think a lot of this PR stayed the way it is instead of rethinking what it actually needs to accomplish now as outlined in the relevant issue.

I'm thinking of the following:
1. Is the EventSkeleton mechanism still necessary? (i've elaborated down in the review)
2. So derive be a package of it's own? Does it actually make the code more readable and appropriately siloed? I remind that derivation is uniquely an event related mechanism
3. The derivation functions are still clustered in one place. What if we use the derive package to store them, instead of the events mechanism? Since it's a pipeline specific implementation, that could be an ebpf subpackage.
4. There are some left overs from before #1818 here (dropping derived events) that need to be cleaned, i'm mentioning this to show that it might be best to rethink this PR's additions.

So, mostly I think this PR can still accomplish code improvement, but it has some clutter now again. I think this can be solved by instead of just rebasing, soft reverting and seeing what actually needs to be added now (and surely there will be, just maybe not all of what was initially included).

AlonZivony:
I feel that the new structure I created has testability issues.
I think maybe it would be better to create a standalone `argsGenerator` function for each derived event which could be tested, and the main `DeriveFunction` generator will only call it and after the `generateSingelEvenetDeriveFunction`.
This way a simple test of "given event x, are the params are as expected?" will be possible.
WDYT?

NDStrahilevitz:
> I feel that the new structure for derived events I created has testability issues. I think maybe it would be better to create a standalone `argsGenerator` function for each derived event which could be tested, and the main `DeriveFunction` generator will only call it and after the `generateSingelEvenetDeriveFunction`. This way a simple test of "given event x, are the params are as expected?" will be possible. WDYT?

So this is a black box - white box thing.
So the problem with testing the derivation functions mostly comes down to the external dependencies they sometime need (for example, containers or a kernelSymbolTable). If you want to take the black box testing approach, you will need to create mocks, and then test the resulting functions.
Alternatively you could take the white box approach and only test the args generators. It would probably be easier TBH and may be just as beneficial since that is the main logic of deriving.
So if we also test `generateSingleEventDeriveFunction` in addition to that, that basically covers our ground as long as people stick to your template. I'm fine with either choices.

pkg/events/derive/derive.go
```diff
@@ -0,0 +1,64 @@
...
+	Params []trace.ArgMeta
+}
+
+type argsGenerator func(event trace.Event) ([]interface{}, error)
+
+// generateSingleEventDeriveFunction create an events.DeriveFunction to generate a single derive trace.Event.
+// The event will be created using the original event information, the ID given and the arguments given.
+// The order of the arguments given will match the order in the event definition, so make sure the order match
+// the order of the params in the events.event struct of the event under events.Definitions.
+func generateSingleEventDeriveFunction(id events.ID, argsGen argsGenerator) events.DeriveFunction {
```
NDStrahilevitz, line 23: This is a nice abstraction IMO.
But maybe just `singleEventDerive` or `singleEventDeriveFunc`? Generate usually implies to me long term generator functions (the kinds that in python for example would `yield`).

pkg/events/derive/derive.go
```diff
@@ -0,0 +1,64 @@
...
+	de.StackAddresses = make([]uint64, 1)
+	de.Args = make([]trace.Argument, len(skeleton.Params))
+	for i, value := range argsValues {
+		de.Args[i] = trace.Argument{ArgMeta: skeleton.Params[i], Value: value}
+	}
+	de.ArgsNum = len(de.Args)
+	return de, nil
+}
+
+func generateEventSkeleton(eventID events.ID) eventSkeleton {
```
NDStrahilevitz, line 57: Same here, maybe just `makeEventSkeleton`? Simpler language.

pkg/events/derive/container_removed.go
```diff
@@ -0,0 +1,24 @@
...
+		cgroupId, err := parse.ArgUint64Val(&event, "cgroup_id")
+		if err != nil {
+			return nil, err
+		}
+		if info := containers.GetCgroupInfo(cgroupId); info.Container.ContainerId != "" {
+			return []interface{}{cgroupId}, nil
+		}
+		return []interface{}{}, nil
+	}
+	return generateSingleEventDeriveFunction(events.ContainerRemove, argsGen)
```
NDStrahilevitz, line 23: Just a thought, you can do an an anonymous function here instead so it would read:

```
return generateSingleEventDeriveFunction(events.ContainerRemove, func(event trace.Event) ([]interface{}, error) {
  cgroupId, err := parse.ArgUint64Val(&event, "cgroup_id")
  if err != nil {
	  return nil, err
  }
  if info := containers.GetCgroupInfo(cgroupId); info.Container.ContainerId != "" {
	  return []interface{}{cgroupId}, nil
  }
  return []interface{}{}, nil
})
```

This pattern is very common in javascript actually, and usually that's the syntax used.
Up to you what you think looks better though.

NDStrahilevitz (commented):
Like this a lot now! I only left some style suggestions.
Maybe also squash some of the commits.

So only the testing is left, I answered my thoughts in a previous comments and besides that it's good!

pkg/utils/utils.go
```diff
@@ -0,0 +1,18 @@
+package utils
+
+import (
+	"strings"
+
+	"github.com/aquasecurity/libbpfgo/helpers"
+)
+
+func ParseSymbol(address uint64, table *helpers.KernelSymbolTable) *helpers.KernelSymbol {
```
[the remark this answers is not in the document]
AlonZivony: I agree, but for now this is what I could do.

pkg/events/derive/container_removed.go
AlonZivony: So after thinking about it more I decided that the function here should be testable.
So I decided that it would be better to create her in a sub-function.

NDStrahilevitz (approved):
LGTM!

AlonZivony:
We have some issue with the new event I pushed, fixing it and than Nadav will give his final +1

Issue #1729, opened 2022-05-03
Tracee-Rules filter mediation


NDStrahilevitz:
## Prerequisites

- [ ] This issue is an EPIC issue (add label: EPIC).
- [ ] This issue is an EPIC TASK (add issue to EPIC description).

Select one OR another:

- [x] I'll create a PR to implement this feature (assign to yourself).

## Feature description

In order to configure tracee-ebpf filters during runtime according to signature needs, the metadata described in #1728 should be used to send requests to the API in #1726.
This request mediation should happen in the tracee-rules `cmd`, in order to decouple from the engine itself.

## Additional Information (feature drawings, files, logs, etc)

<img width="656" alt="image" src="https://user-images.githubusercontent.com/22661609/166468493-a8a9ebf7-2b3a-423d-8ef9-c34fe3c475b5.png">

yanivagman:
With the new tracee binary, this becomes irrelevant

Issue #1728, opened 2022-05-03
Filter Signature Metadata


NDStrahilevitz:
## Prerequisites

- [ ] This issue is an EPIC issue (add label: EPIC).
- [ ] This issue is an EPIC TASK (add issue to EPIC description).

Select one OR another:

- [x] I'll create a PR to implement this feature (assign to yourself).
- [ ] Someone else should implement this (describe it well).

## Feature description

In order to allow configuring tracee-ebpf filters according to the needs of tracee-rules signatures, signatures should expose a new `Filters` metadata field. This can be used by tracee-rules or any other mediator to determine filter requests to be sent to tracee-ebpf (API in #1726).
If a priority system will be needed, the signature's severity field will be used to determine that priority.
Per event filters will be supported through #1830.

## Additional Information (feature drawings, files, logs, etc)

Draft of the metadata struct:

```
type Operator uint //enum for evaluator types (for example: ==, !=, >, etc.)

type Filter struct {
    field         string
    value        interface{} //this should match the filterType
    operator  Operator
}
```

An example for a filter struct:
`Filter{field: "execve.container", value: true, operator: Equal}`

mtcherni95:
@AsafEitani need to think how we should add the filtering metadata to signatures.

AsafEitani:
Cool!
I got some concerns though:

- Once we move to that method it's all or nothing. You can't have a rule that is using this feature and another that isn't because that will mean either we get nothing from it (all the events will have to get to userspace) or we will lose relevant events for the later signature (because by not using the feature it doesn't have a `Filter` field - so only event relevant to other signature will get to it)
- The suggested struct for it is **very** not comfortable and is unreadable. We should try to come up with something more elegant in my opinion.
- I assume that rego rules won't support this feature - but I'm OK with that because is seems that we are moving away from rego - see #1732 
- There are some filters that are not that easy to set a filter for.
- The conversion of the rules will be manual and hard, some filters are complicated and we will need to find a simple representation for them to be checked in the kernel =(eg. regex filter will have to have a contains, starts or ends filter of some part of the regex filter - `\.+?foo\d+\s\w.+?bar` will have to have some sort of a filter like `contains('foo')\ends('bar')`)

NDStrahilevitz:
@AsafEitani These are valid concerns, i'd like to try and address them because I think they are mostly solvable:

[...]

PR #1750, opened 2022-05-11
filters: integrate new filters protocol

NDStrahilevitz:
## Initial Checklist

- [x] There is an issue describing the need for this PR.
- [ ] Git log contains summary of the change.
- [ ] Git log contains motivation and context of the change.
- [ ] If part of an EPIC, PR git log contains EPIC number.
- [ ] If part of an EPIC, PR was added to EPIC description.

## Description (git log)

This PR currently adds a new filters package, intended to replace the existing filters in `pkg/tracee/filters.go`.
These filters should work exactly the same as the existing ones with the added ability of being able to be changed during runtime.

Additionally, it makes use of the new `protocol.Filter` to create a later common interface between parsing the cli into the main Filter struct, and a gRPC service we will add later to receive mediated filters from tracee-rules.
It refactors filter parsing from the CLI - switching to parse into `[]protocol.Filter` which is down the line parsed into a `tracee.Filter`.

While this tries to maintain a high degree of similarity to the previous filtering logic, this PR probably breaks previous behavior.

Fixes: #1727 

## Type of change

- [x] Code refactor (code improvement and/or code removal)
- [x] New feature (non-breaking change adding functionality).
- [x] Breaking change (cause existing functionality not to work as expected) (possibly?).

## How Has This Been Tested?

New unit tests were added to the new filters package.
Tests were moved and changed according to the new flow of filter info.

Reproduce the test by running:

`make test-unit`

## Final Checklist:

- [x] I have made corresponding changes to the documentation.
- [x] My code follows the style guidelines (C and Go) of this project.
- [ ] I have performed a self-review of my own code.
- [x] I have commented all functions/methods created explaining what they do.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] My changes generate no new warnings.
- [x] I have added tests that prove my fix, or feature, is effective.
- [ ] New and existing unit tests pass locally with my changes.
- [ ] Any dependent changes have been merged and published before.

## Git Log Checklist:

My commits logs have:

- [x] Subject starts with "subsystem|file: description".
- [x] Do not end the subject line with a period.
- [x] Limit the subject line to 50 characters.
- [x] Separate subject from body with a blank line.
- [x] Use the imperative mood in the subject line.
- [x] Wrap the body at 72 characters.
- [x] Use the body to explain what and why instead of how.

PR #1832, opened 2022-06-16
Signature Filter Metadata

NDStrahilevitz:
## Initial Checklist

## Description (git log)

44cfa746e5cc88e8646035186b4bd8a6e31eef1c engine: add mediated filters method
f0b33000f091e76d0d2c3a585bdc9dd9badcc4f6 signatures: add filters to signatures
11086f905cc007156717183213e5503f91460617 types: add filter protocol for signatures

Commit 11086f905cc007156717183213e5503f91460617 will be added later as a separate perquisite PR.

Fixes: #1728
Fixes: #1729

Depends on #1842

## Type of change

## How Has This Been Tested?

New tests were added to engine_test.go for filter mediation.

In addition, filters were added to all existing signatures to generate the following mediated output:
```
filter: {Field:args.request Operator:0 Value:[PTRACE_TRACEME]}
filter: {Field:event Operator:0 Value:[execve security_socket_connect ptrace security_file_open magic_write sched_process_exec dup dup2 dup3 close security_bprm_check hooked_syscalls process_vm_writev mem_prot_alert init_module security_kernel_read_file security_sb_mount]}
filter: {Field:execve.args.envp Operator:0 Value:[*LD_PRELOAD* *LD_LIBRARY_PATH]}
filter: {Field:execve.container Operator:0 Value:[true]}
filter: {Field:hooked_syscalls.args.hooked_syscalls Operator:1 Value:[[]]}
filter: {Field:magic_write.container Operator:0 Value:[true]}
filter: {Field:ptrace.args.request Operator:0 Value:[PTRACE_POKETEXT]}
filter: {Field:sched_process_exec.args.pathname Operator:0 Value:[memfd:* /dev/shm* /run/shm*]}
filter: {Field:security_bprm_check.args.pathname Operator:0 Value:[*/ash */bash */csh */ksh */sh */tcsh */zsh */dash]}
filter: {Field:security_bprm_check.processName Operator:0 Value:[nginx httpd httpd-foregroun lighttpd apache apache2]}
filter: {Field:security_file_open.args.pathname Operator:0 Value:[*/release_agent /etc/kubernetes/pki/* *secrets/kubernetes.io/serviceaccount* *token /etc/ld.so.preload]}
filter: {Field:security_file_open.container Operator:0 Value:[true]}
filter: {Field:security_file_open.processName Operator:1 Value:[flanneld kube-proxy etcd kube-apiserver coredns kube-controller kubectl]}
filter: {Field:security_file_open.processName Operator:0 Value:[kube-apiserver kubelet kube-controller etcd]}
filter: {Field:security_kernel_read_file.args.type Operator:0 Value:[kernel-module]}
filter: {Field:security_sb_mount.args.dev_name Operator:0 Value:[/dev/*]}
filter: {Field:security_sb_mount.container Operator:0 Value:[true]}
filter: {Field:security_socket_connect.container Operator:0 Value:[true]}
```

This output will be used to test #1750

## Final Checklist:

- [ ] Any dependent changes (#1842) have been merged and published before.

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/rules/engine/engine.go
+++ b/pkg/rules/engine/engine.go
@@ -306,3 +306,102 @@ func (engine *Engine) GetSelectedEvents() []detect.SignatureEventSelector {
 	}
 	return res
 }
+
+// GetFilters returns a mediated array of all signature's filters
+func GetFilters(signatures []detect.Signature) ([]detect.Filter, error) {
+	filters := make([]detect.Filter, 0)
+	for _, sig := range signatures {
+		sigFilters, err := getSignatureFilters(sig)
+		if err != nil {
+			return []detect.Filter{}, fmt.Errorf("error getting filters from signature: %s", err)
+		}
+		filters = append(filters, sigFilters...)
+	}
+	res := mediateFilters(filters)
+	return res, nil
+}
+
+// GetFilters returns a mediated array of all engine loaded signatures filters
+func (engine *Engine) GetFilters() ([]detect.Filter, error) {
+	sigs := []detect.Signature{}
+	engine.signaturesMutex.RLock()
+	defer engine.signaturesMutex.RUnlock()
+	for sig := range engine.signatures {
+		sigs = append(sigs, sig)
+	}
+	return GetFilters(sigs)
+}
+
+// getSignatureFilters get all filters from a signature, including base filters from selectors
+func getSignatureFilters(sig detect.Signature) ([]detect.Filter, error) {
+	res := []detect.Filter{}
+	metadata, _ := sig.GetMetadata()
+	selectors, err := sig.GetSelectedEvents()
+	if err != nil {
+		return []detect.Filter{}, fmt.Errorf("missing selectors for base filters from signature %s", metadata.Name)
+	}
+	for _, s := range selectors {
+		res = append(res, selectorToFilter(s)...)
+	}
+	filters, err := sig.GetFilters()
+	if err != nil {
+		return []detect.Filter{}, fmt.Errorf("error getting filter from signature %s: %s", metadata.Name, err)
+	}
+	res = append(res, filters...)
+
+	return res, nil
+}
+
+// selectorToFilter generates default initial filters from the signature's event selectors
+func selectorToFilter(selector detect.SignatureEventSelector) []detect.Filter {
+	eventFilter := detect.EqualFilter("event", []interface{}{selector.Name})
+	res := []detect.Filter{eventFilter}
+	if selector.Origin != "" && selector.Origin != "*" {
+		originFilter := detect.EqualFilter(fmt.Sprintf("%s.context.%s", selector.Name, selector.Origin), []interface{}{true})
+		res = append(res, originFilter)
+	}
+	return res
+}
+
+// mediateFilters performs mediation logics on a list of filters
+func mediateFilters(filters []detect.Filter) []detect.Filter {
+	type filterHeader struct {
+		field    string
+		operator detect.FilterOperator
+	}
+
+	filterGroups := make(map[filterHeader]*detect.Filter)
+	// For now - mediation should be like a SQL GROUP BY clause, with no value duplications
+	for _, filter := range filters {
+		// Get the filter "header"
+		header := filterHeader{field: filter.Field, operator: filter.Operator}
+
+		// create the group if needed
+		if filterGroups[header] == nil {
+			filterGroups[header] = &detect.Filter{
+				Field:    header.field,
+				Operator: header.operator,
+			}
+		}
+
+		// Append non existing filter values to the group
+		for _, filterVal := range filter.Value {
+			found := false
+			for _, fgVal := range filterGroups[header].Value {
+				if filterVal == fgVal {
+					found = true
+					break
+				}
+			}
+			if !found {
+				filterGroups[header].Value = append(filterGroups[header].Value, filterVal)
+			}
+		}
+	}
+
+	res := make([]detect.Filter, 0)
+	for _, filter := range filterGroups {
+		res = append(res, *filter)
+	}
+	return res
+}
--- a/signatures/golang/kubernetes_api_connection.go
+++ b/signatures/golang/kubernetes_api_connection.go
@@ -43,6 +43,12 @@ func (sig *K8sApiConnection) GetSelectedEvents() ([]detect.SignatureEventSelecto
 	}, nil
 }
 
+func (sig *K8sApiConnection) GetFilters() ([]detect.Filter, error) {
+	return []detect.Filter{
+		detect.ContainsFilter("security_socket_connect.args.remote_addr", "sa_family:AF_INET", "sa_family:AF_INET6"), //maps parse as strings to "map[key1:value1 key2:value2 ...]"
+	}, nil
+}
+
 func (sig *K8sApiConnection) OnEvent(event protocol.Event) error {
 
 	eventObj, ok := event.Payload.(trace.Event)
--- a/signatures/golang/stdio_over_socket.go
+++ b/signatures/golang/stdio_over_socket.go
@@ -44,9 +44,15 @@ func (sig *StdioOverSocket) GetSelectedEvents() ([]detect.SignatureEventSelector
 	}, nil
 }
 
-func (sig *StdioOverSocket) OnEvent(event protocol.Event) error {
+func (sig *StdioOverSocket) GetFilters() ([]detect.Filter, error) {
+	return []detect.Filter{
+		detect.ContainsFilter("security_socket_connect.args.remote_addr", "sa_family:AF_INET", "sa_family:AF_INET6"), //maps parse as strings to "map[key1:value1 key2:value2 ...]"
+	}, nil
+}
 
+func (sig *StdioOverSocket) OnEvent(event protocol.Event) error {
 	eventObj, ok := event.Payload.(trace.Event)
+
 	if !ok {
 		return fmt.Errorf("invalid event")
 	}
--- a/signatures/rego/anti_debugging_ptraceme.rego
+++ b/signatures/rego/anti_debugging_ptraceme.rego
@@ -1,5 +1,7 @@
 package tracee.TRC_2
 
+import data.tracee.helpers
+
 __rego_metadoc__ := {
 	"id": "TRC-2",
 	"version": "0.1.0",
@@ -18,6 +20,14 @@ tracee_selected_events[eventSelector] {
 		"name": "ptrace",
 	}
 }
+
+signature_filters := [
+	{
+		"field": "ptrace.args.request",
+		"operator": helpers.filter_equal,
+		"value": ["PTRACE_TRACEME"]
+	}
+]
 
 tracee_match {
 	input.eventName == "ptrace"
--- a/cmd/tracee-rules/main.go
+++ b/cmd/tracee-rules/main.go
@@ -15,6 +15,7 @@ import (
 	"github.com/aquasecurity/tracee/pkg/rules/engine"
 	"github.com/aquasecurity/tracee/pkg/server"
 	"github.com/aquasecurity/tracee/types/detect"
+	"github.com/aquasecurity/tracee/types/protocol"
 
 	"github.com/open-policy-agent/opa/compile"
 	"github.com/urfave/cli/v2"
@@ -38,6 +39,7 @@ func init() {
 
 const (
 	signatureBufferFlag = "sig-buffer"
+	listFilterFlag      = "list-filters"
 )
 
 func main() {
@@ -47,7 +49,6 @@ func main() {
 		Action: func(c *cli.Context) error {
 
 			// Capabilities command line flags
-
 			err := capabilities.Initialize(c.Bool("allcaps"))
 			if err != nil {
 				return err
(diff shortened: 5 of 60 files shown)


pkg/rules/engine/engine.go
```diff
@@ -307,3 +307,57 @@ func (engine *Engine) GetSelectedEvents() []detect.SignatureEventSelector {
 	}
 	return res
 }
+
+// GetFilters returns a mediated array of all signature's filters
+func (engine *Engine) GetFilters() ([]protocol.Filter, error) {
+	allFilters := make([]protocol.Filter, 0)
+	engine.signaturesMutex.RLock()
```
mtcherni95, line 314: I don't see where the lock is released. What am I missing? If it is not released then a deadlock will be created. Let's `defer` lock release :)

pkg/rules/engine/engine.go
```diff
@@ -307,3 +307,57 @@ func (engine *Engine) GetSelectedEvents() []detect.SignatureEventSelector {
...
+	for sig := range engine.signatures {
+		filters, err := sig.GetFilters()
```
mtcherni95, line 316: why we return if only one signature fails in giving back its filters? why not just pushing error to error channel and continue with the iteration over the signatures?

pkg/rules/engine/engine.go
```diff
@@ -307,3 +307,57 @@ func (engine *Engine) GetSelectedEvents() []detect.SignatureEventSelector {
...
+		if err != nil {
+			return []protocol.Filter{}, fmt.Errorf("error getting filters: %v", err)
+		}
+		if len(filters) == 0 {
+			metadata, _ := sig.GetMetadata()
```
mtcherni95, line 321: ditto

pkg/rules/engine/engine.go
```diff
@@ -307,3 +307,57 @@ func (engine *Engine) GetSelectedEvents() []detect.SignatureEventSelector {
 	}
 	return res
 }
+
+// GetFilters returns a mediated array of all signature's filters
```
mtcherni95, line 311: might be helpful to give a short explanation on what mediated array is.

pkg/rules/engine/engine.go
```diff
@@ -307,3 +307,57 @@ func (engine *Engine) GetSelectedEvents() []detect.SignatureEventSelector {
...
+	}
+	remove := func(s []interface{}, i int) []interface{} {
+		s[i] = s[len(s)-1]
+		return s[:len(s)-1]
+	}
+	filterGroups := make(map[filterHeader]*protocol.Filter)
+	for _, filter := range filters {
+		header := filterHeader{field: filter.Field, operator: filter.Operator}
+		if filterGroups[header] == nil {
+			filterGroups[header] = &protocol.Filter{}
```
mtcherni95, line 344: why not setting `Field` and `Operator` in object creation? This instead of accessing twice to the map.

pkg/rules/engine/engine.go
```diff
@@ -307,3 +307,57 @@ func (engine *Engine) GetSelectedEvents() []detect.SignatureEventSelector {
...
+			filterGroups[header].Field = header.field
+			filterGroups[header].Operator = header.operator
+		}
+		for _, fgVal := range filterGroups[header].Value {
```
mtcherni95, line 348: not sure what we are trying to achieve here

pkg/rules/engine/engine.go
```diff
@@ -307,3 +307,57 @@ func (engine *Engine) GetSelectedEvents() []detect.SignatureEventSelector {
...
+			for i := 0; i < len(filter.Value); i++ {
+				if filter.Value[i] == fgVal {
+					filter.Value = remove(filter.Value, i)
+					i--
+				}
+			}
+		}
+		filterGroups[header].Value = append(filterGroups[header].Value, filter.Value...)
+	}
+	res := make([]protocol.Filter, 0)
```
mtcherni95, line 358: why not size zero and not `len(filterGroups)`

pkg/rules/engine/engine.go
NDStrahilevitz: Not missing anything I just forgot to add it back while I was rewriting stuff, thanks for catching.

pkg/rules/engine/engine.go
NDStrahilevitz: I remember adding a comment about that, looks like it went missing, i'll add it back too.

But basically, if even one signature has no filters, if we go ahead with filtering anyway, that signature will most likely miss a lot, if not all, relevant events, so filtering opts out the moment not all signatures have it enabled.

pkg/rules/engine/engine.go
NDStrahilevitz: ditto ditto

pkg/rules/engine/engine.go
NDStrahilevitz: Well mediation is a very general concept. But mediated has a meaning, how the mediation itself is done could be explained later in the function. But for mediation examples, one can look at the tests.

pkg/rules/engine/engine.go
NDStrahilevitz: Because then I would need to do an indexed loop instead of a range. This function isn't performance critical so I don't think it matters that much to save the append calls.

pkg/rules/engine/engine.go
NDStrahilevitz: This is the mediation itself, it's basically an SQL `GROUP BY` on Field and Operator. I'll add a comment explaining that too.

signatures/golang/kubernetes_api_connection.go
```diff
@@ -43,6 +43,14 @@ func (sig *K8sApiConnection) GetSelectedEvents() ([]detect.SignatureEventSelecto
 	}, nil
 }
 
+func (sig *K8sApiConnection) GetFilters() ([]protocol.Filter, error) {
+	return []protocol.Filter{
+		{Field: "execve.container", Operator: protocol.Equal, Value: []interface{}{true}},
+		{Field: "security_socket_connect.container", Operator: protocol.Equal, Value: []interface{}{true}},
+		{Field: "event", Operator: protocol.Equal, Value: []interface{}{"execve", "security_socket_connect"}},
```
AsafEitani, line 50: Would we need to add filter to state the source type and events? seems really redundant due to the fact that we already state it in the event selector...
Couldn't you get that from the event selectors automatically?

signatures/golang/stdio_over_socket.go
```diff
@@ -51,6 +51,12 @@ func (sig *stdioOverSocket) GetSelectedEvents() ([]detect.SignatureEventSelector
 	}, nil
 }
 
+func (sig *stdioOverSocket) GetFilters() ([]protocol.Filter, error) {
+	return []protocol.Filter{
+		{Field: "event", Operator: protocol.Equal, Value: []interface{}{"security_socket_connect", "dup", "dup2", "dup3", "close", "sched_process_exec"}},
```
AsafEitani, line 56: we could possibly add a `equals` or `compare` function to types like addresses and then be able to compare them.

signatures/rego/anti_debugging_ptraceme.rego
```diff
@@ -19,6 +19,23 @@ tracee_selected_events[eventSelector] {
 	}
 }
 
+filters := [
+	{
+		"field": "event",
+		"operator": 0, #Equal
```
AsafEitani, line 25: I would recommend adding an alias to that in the helpers.rego so we could use `equal` instead of `0`

AsafEitani:
Could we add an example to a go signature? they are all empty.

NDStrahilevitz:
> Could we add an example to a go signature? they are all empty.

You're looking at the benchmark ones, look at `signatures/golang`, they are defined.

signatures/golang/kubernetes_api_connection.go
AsafEitani: Also can we add a helper for each operator? that way we can use:
`helpers.equals("event", ["execve", "security_socket_connect"])` and it will return the appropriate filter object.
This will be way more readable and will help us convert our signatures.

NDStrahilevitz: I'm thinking maybe separating the Filter from protocol and adding a filters package to types. Then we could define these helpers there WDYT?

signatures/golang/stdio_over_socket.go
NDStrahilevitz: Not sure I understand, could you please elaborate and provide example syntax?

signatures/rego/anti_debugging_ptraceme.rego
NDStrahilevitz: image
WDYT?

signatures/golang/kubernetes_api_connection.go
AsafEitani: Sounds good

signatures/golang/stdio_over_socket.go
AsafEitani: If we have for example the slimCreds struct we could define a function named compare that receives another variable and compares them using some logic
`slimCred1.Compare(bitmaskIntFromFilter)`

So we could compare more complex types and not only strings and ints.

signatures/rego/anti_debugging_ptraceme.rego
AsafEitani: sounds good although I would even just call it equal but it's really a preference at this point.

signatures/golang/stdio_over_socket.go
NDStrahilevitz: So SlimCred for example are usually translated to args right? You could set filters on the corresponding args created from it I think. So you could create your own helpers to make filters for those structs.
It would really help if I got a code example, or you can make a PR to this branch to show an example.

signatures/golang/kubernetes_api_connection.go
NDStrahilevitz: Done.

types/detect/detect.go
```diff
@@ -36,6 +42,16 @@ type SignatureEventSelector struct {
 	Origin string
 }
 
+func (selector SignatureEventSelector) ToFilters() []protocol.Filter {
```
rafaeldtinoco, line 45: So you're basically generating an initial filter from the existing event selector right ? Should we just mandate event selector to be a regular filter and you just get it together with all others ? No hard opinion, thinking out loud.

rafaeldtinoco (commented):
Nice work.

We already knew that the filters would look very much like re-creating the existing signatures, that is a price to pay for now (until we create filters automatically, after having other things merged, like CEL support, etc). 

Thanks for already creating the filters for existing signatures (including rego).

pkg/rules/engine/engine.go
```diff
@@ -307,3 +307,77 @@ func (engine *Engine) GetSelectedEvents() []detect.SignatureEventSelector {
...
+	type filterHeader struct {
+		field    string
+		operator protocol.FilterOperator
+	}
+	// remove := func(s []interface{}, i int) []interface{} {
+	// 	s[i] = s[len(s)-1]
+	// 	return s[:len(s)-1]
+	// }
+	filterGroups := make(map[filterHeader]*protocol.Filter)
+	// For now - mediation should be like a SQL GROUP BY clause, with no value duplications
```
rafaeldtinoco, line 351: So we're not mediating among different operators ? I think "it is fine" if we consider the current filtering mechanism also does not mediate that (giving mutual exclusive operators would be problematic with what we currently have now as well). I believe we will have to handle that eventually ? (likely outside the scope of this POC ?).

types/detect/detect.go
NDStrahilevitz: @AsafEitani and I talked about the similarity between the two, which is why he wanted this function to exist, but I think there is merit to separate definitions because they do separate things, the selectors are used to route inside the engine, a functionality still needed even after the filters. But yes, the selectors do allow us to have the most simple filters for all signatures.
I'm actually considering moving this function to `engine.go` to avoid this implied understanding that these two structs are the same.
And defining the selectors as filters would require large code changes in other places, so even if we go that direction, it might be best to think of it later down the line.

pkg/rules/engine/engine.go
NDStrahilevitz: So the filters in tracee-ebpf sort of already mediate between their operators. I have kept that logic in my filters refactor (#1750). I also believe we will face that problem if and when this filter system gets more complex, but I don't want to get ahead of myself with that.

NDStrahilevitz:
I have uploaded #1842 as a base types dependency for all the filter changes. For this purpose I will remove those commits which I combined from this PR.

types/detect/detect.go
rafaeldtinoco: > I'm actually considering moving this function to engine.go to avoid this implied understanding that these two structs are the same.

I like that as well.

pkg/rules/engine/engine.go
rafaeldtinoco: > but I don't want to get ahead of myself with that.

agreed, we should have it simple and working for most cases now.

signatures/golang/stdio_over_socket.go
NDStrahilevitz: I just added an example of relevant filters here @AsafEitani could you check?

yanivagman:
Agreed to discuss this later if needed

PR #1837, opened 2022-06-19
ebpf: changed to use log package

AlonZivony:
## Initial Checklist

## Description (git log)

Change prints and error handling to use logging logic instead.
This way tracee-epbf can support multiple logging levels without moving
context between packages.
This is an alternative to #1788, using a known package for logging.

Fixes: #1787 

## Type of change

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/containers/containers.go
+++ b/pkg/containers/containers.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/aquasecurity/libbpfgo"
 	cruntime "github.com/aquasecurity/tracee/pkg/containers/runtime"
+	log "github.com/sirupsen/logrus"
 )
 
 var cgroupV1HierarchyID int
@@ -44,7 +45,7 @@ type CgroupInfo struct {
 
 // InitContainers initializes a Containers object and returns a pointer to it.
 // User should further call "Populate" and iterate with Containers data.
-func InitContainers(sockets cruntime.Sockets, debug bool) (*Containers, error) {
+func InitContainers(sockets cruntime.Sockets) (*Containers, error) {
 	cgroupV1 := false
 	if _, err := os.Stat("/sys/fs/cgroup/cgroup.controllers"); os.IsNotExist(err) {
 		cgroupV1 = true
@@ -57,16 +58,16 @@ func InitContainers(sockets cruntime.Sockets, debug bool) (*Containers, error) {
 
 	//attempt to register for all supported runtimes
 	err := runtimeService.Register(cruntime.Containerd, cruntime.ContainerdEnricher)
-	if err != nil && debug {
-		fmt.Fprintf(os.Stderr, "Enricher: %v\n", err)
+	if err != nil {
+		log.Debugf("Enricher: %v\n", err)
 	}
 	err = runtimeService.Register(cruntime.Crio, cruntime.CrioEnricher)
-	if err != nil && debug {
-		fmt.Fprintf(os.Stderr, "Enricher: %v\n", err)
+	if err != nil {
+		log.Debugf("Enricher: %v\n", err)
 	}
 	err = runtimeService.Register(cruntime.Docker, cruntime.DockerEnricher)
-	if err != nil && debug {
-		fmt.Fprintf(os.Stderr, "Enricher: %v\n", err)
+	if err != nil {
+		log.Debugf("Enricher: %v\n", err)
 	}
 
 	return &Containers{
--- a/pkg/ebpf/log.go
+++ b/pkg/ebpf/log.go
@@ -1,0 +1,55 @@
+package ebpf
+
+import (
+	"fmt"
+	"github.com/aquasecurity/tracee/pkg/metrics"
+	log "github.com/sirupsen/logrus"
+	"github.com/sirupsen/logrus/hooks/writer"
+	"io/ioutil"
+	"os"
+)
+
+// ErrorChanWriter is a io.Writer implementation that writes to error channel, while updating program stats.
+type ErrorChanWriter struct {
+	c     chan<- error
+	stats *metrics.Stats
+}
+
+func InitErrorChanWriter(c chan<- error, stats *metrics.Stats) ErrorChanWriter {
+	return ErrorChanWriter{
+		c:     c,
+		stats: stats,
+	}
+}
+
+func (ec ErrorChanWriter) Write(message []byte) (n int, err error) {
+	ec.stats.ErrorCount.Increment()
+	ec.c <- fmt.Errorf(string(message))
+	return n, nil
+}
+
+// ConfigureLogger configure the errors created by tracee to be directed into the error channel, and normal
+// logs to stderr.
+// Notice that the logger log level must enable the log for it to be written.
+// For example, if the followed log level is Error and above, Warnings won't be written to the errors channel.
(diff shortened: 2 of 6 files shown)


AlonZivony:
@NDStrahilevitz @rafaeldtinoco WDYT?

pkg/ebpf/log.go
```diff
@@ -0,0 +1,55 @@
...
+	"fmt"
+	"github.com/aquasecurity/tracee/pkg/metrics"
+	log "github.com/sirupsen/logrus"
+	"github.com/sirupsen/logrus/hooks/writer"
+	"io/ioutil"
+	"os"
+)
+
+// ErrorChanWriter is a io.Writer implementation that writes to error channel, while updating program stats.
+type ErrorChanWriter struct {
```
NDStrahilevitz, line 13: Currently this doesn't need to be exported since it's only used here.

pkg/ebpf/log.go
```diff
@@ -0,0 +1,55 @@
+package ebpf
```
NDStrahilevitz, line 1: Maybe we could separate this into an `errorwriter` package?

pkg/ebpf/log.go
```diff
@@ -0,0 +1,55 @@
...
+func (ec ErrorChanWriter) Write(message []byte) (n int, err error) {
+	ec.stats.ErrorCount.Increment()
+	ec.c <- fmt.Errorf(string(message))
+	return n, nil
+}
+
+// ConfigureLogger configure the errors created by tracee to be directed into the error channel, and normal
+// logs to stderr.
+// Notice that the logger log level must enable the log for it to be written.
+// For example, if the followed log level is Error and above, Warnings won't be written to the errors channel.
```
NDStrahilevitz, line 34: Just to make sure, does the ordering of the levels in `LogLevels` matter, or does it just declare the relevant levels and their order is static?

pkg/ebpf/log.go
```diff
@@ -0,0 +1,55 @@
...
+func InitErrorChanWriter(c chan<- error, stats *metrics.Stats) ErrorChanWriter {
+	return ErrorChanWriter{
+		c:     c,
+		stats: stats,
+	}
+}
+
+func (ec ErrorChanWriter) Write(message []byte) (n int, err error) {
+	ec.stats.ErrorCount.Increment()
+	ec.c <- fmt.Errorf(string(message))
```
NDStrahilevitz, line 27: Maybe use `errors.New` here instead, since there's no formatting used.

pkg/containers/containers.go
```diff
@@ -17,6 +17,7 @@ import (
 
 	"github.com/aquasecurity/libbpfgo"
 	cruntime "github.com/aquasecurity/tracee/pkg/containers/runtime"
+	log "github.com/sirupsen/logrus"
```
NDStrahilevitz, line 20: The `log` import for example in respect to my 2nd point in the review comment.

NDStrahilevitz (commented):
I personally like this better than the interface proposal, and I think this already adds a lot in terms of simplifying the code.

However here is what I am unsure of:
1. Instead of importing directly as log, since it seems we have decided to bake this into tracee, maybe tracee should own a logger from the package? Then we use t.Loglevelf(...). I'm suggesting this because otherwise we have to remember to import the package everywhere and rename it to `log`.
2. If we don't want to bake this into tracee, maybe we should go back to the interface suggestion, or define a simple messaging struct for the existing ErrorChannel. Then we would use this logger externally (at the cmd level for us) and other consumers could make that decision on their own.
3. Irrespectively of how we handle the previous 2 comments, I think we should allow the error logging to be structured as json for consumers. This could either be done automatically if `--output json` is set, or maybe through a different flag.
4. In this PR you mostly replaced the existing `fmt.Fprintf` calls with log calls. But I think there are a lot of error returns which probably fit better to be just logged to debug instead (In particular this was raised in the container enrichment, but there's possibly more relevant features). The question is: should this be fixed in the scope of this PR, or maybe handle it later?

So to conclude, I like this but I think we should answer the above stuff before deciding if to go ahead with this or the other option, and change what we choose according to the reasons why.

yanivagman:
I didn't review the code, but to refer to parts of the comment by @NDStrahilevitz, I think that:
1. Logger should be in its own package, where it can wrap the logrus package
2. Logger can be kept in the tracee instance, like Nadav suggested
3. A logger instnace can be initialized by main, and set for every package used by it (today we have the main tracee struct in the ebpf package, but this will change in the future)

pkg/ebpf/log.go
AlonZivony: No, the order is insignificant here.

pkg/ebpf/log.go
AlonZivony: So I thought about it too,
but we use a very specific implementation here and I feel this struct is too small and basic for its own package.
Maybe in the future if we will develop it more than it would be exported.

pkg/ebpf/tracee.go
```diff
@@ -333,7 +335,7 @@ func New(cfg Config) (*Tracee, error) {
 		return nil, fmt.Errorf("error intitalizing event derivation map: %w", err)
 	}
 
-	c, err := containers.InitContainers(t.config.Sockets, t.config.Debug)
```
rafaeldtinoco, line 336: Much better =D

pkg/ebpf/log.go
```diff
@@ -0,0 +1,55 @@
...
+		Writer: InitErrorChanWriter(t.config.ChanErrors, &t.stats),
+		LogLevels: []log.Level{
+			log.WarnLevel,
+			log.ErrorLevel,
+			log.FatalLevel,
+			log.PanicLevel,
+		},
+	})
+	log.AddHook(&writer.Hook{
+		Writer: os.Stderr,
```
rafaeldtinoco, line 48: Shouldn't we respect the configuration option for output ? If you run:

```
./dist/tracee-ebpf --output help
...
out-file:/path/to/file                             write the output to a specified file. create/trim the file if exists (default: stdout)
err-file:/path/to/file                             write the errors to a specified file. create/trim the file if exists (default: stderr)
...
```

Instead of hardcoding stdout/stderr ?

rafaeldtinoco:
> I didn't review the code, but to refer to parts of the comment by @NDStrahilevitz, I think that:
> 
> 1. Logger should be in its own package (which can simlpy be named `logger`), where it can wrap the logrus package
> 2. A logger instnace can be initialized by main, and set for every package used by it (today we have the main tracee struct in the ebpf package, but this will change in the future)
> 3. Logger can be kept in the tracee instance, like Nadav suggested (`t.logger`)

I agree to all this PLUS, let's make sure we configure logger from the cmdline output options as well (out-file, err-file options).

AlonZivony:
I wonder if we might want to implement the `Warning`, `Error`, `Fatal` and `Panic` with `error` argument instead of `string`.
It will allow tests to use the `errors.As` and `errors.Is` methods to check the workflows.
WDYT?

rafaeldtinoco:
I believe this is currently being worked by @geyslan now, for 0.9.0, right @AlonZivony ? IF that is so, mind if we close this one ?

AlonZivony:
Not at all

PR #1839, opened 2022-06-19
Cache should trace and introduce task_info map

yanivagman:
## Initial Checklist

## Description (git log)

This PR introduces two changes that improve the bpf programs performance.
The first change avoids performing scope caclculations (should_trace) on every event, by caching the last result of should_trace() in a new per-task map called task_info_map.

In addition, it adds task's start_time into the event context of every event so we can identify a specific task even when a pid is being reused (due to wrapping).

The second change reduces the amount of map lookups required for the different events by removing new_pids, trace_pids and syscall_data maps and introducing the matching fields in task_info map.

Fixes: #issue_number

## Type of change

## How Has This Been Tested?

Tested manually with the different filters that exist in tracee to verify nothing was broken.

- command 01
- command 02

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -467,28 +467,8 @@ typedef __u64 stack_trace_t[MAX_STACK_DEPTH];
 
 // INTERNAL STRUCTS --------------------------------------------------------------------------------
 
-typedef struct event_context {
-    u64 ts; // Timestamp
-    u64 cgroup_id;
-    u32 pid;       // PID as in the userspace term
-    u32 tid;       // TID as in the userspace term
-    u32 ppid;      // Parent PID as in the userspace term
-    u32 host_pid;  // PID in host pid namespace
-    u32 host_tid;  // TID in host pid namespace
-    u32 host_ppid; // Parent PID in host pid namespace
-    u32 uid;
-    u32 mnt_id;
-    u32 pid_id;
-    char comm[TASK_COMM_LEN];
-    char uts_name[TASK_COMM_LEN];
-    u32 eventid;
-    s64 retval;
-    u32 stack_id;
-    u16 processor_id; // The ID of the processor which processed the event
-    u8 argnum;
-} context_t;
-
-typedef struct process_context {
+// todo: check network context after this change
+typedef struct task_context {
     u64 start_time; // thread's start time
     u64 cgroup_id;
     u32 pid;       // PID as in the userspace term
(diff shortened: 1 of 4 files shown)


pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1432,12 +1453,42 @@ static __always_inline int init_event_data(event_data_t *data, void *ctx)
...
+        // check if we need to recompute scope due to context change
+        task_context_t *old_context = &data->task_info->context;
+        task_context_t *new_context = &data->context.task;
+
+        if ((old_context->cgroup_id == new_context->cgroup_id) &&
+            old_context->uid == new_context->uid && old_context->mnt_id == new_context->mnt_id &&
+            old_context->pid_id == new_context->pid_id &&
+            has_prefix(old_context->comm, new_context->comm, TASK_COMM_LEN) &&
+            has_prefix(old_context->uts_name, new_context->uts_name, TASK_COMM_LEN))
+            data->task_info->recompute_scope = true;
```
rafaeldtinoco, line 1486: Shouldn't this be false ? Thought: if nothing has changed, recompute_scope should be false then should_trace() wouldn't call do_should_trace() again (at least until context has changed).

rafaeldtinoco (commented):
70% way through, I'll give another look tomorrow. I like the caching mechanism from what I read so far. Left a small question and a clang-14 error (likely unrelated to this change) for now.

pkg/ebpf/c/tracee.bpf.c
yanivagman: Wow, good catch.
I'll change this to '!=' and "||" instead

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1391,28 +1392,48 @@ static __always_inline int get_iface_config(int ifindex)
     return *config;
 }
 
+// INTERNAL: STRINGS -------------------------------------------------------------------------------
+
+static __inline int has_prefix(char *prefix, char *str, int n)
+{
+    int i;
+#pragma unroll
```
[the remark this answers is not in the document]
yanivagman: I don't know why this happens in clang-14, but it actually gave me an idea.
Instead of comparing char by char like has_prefix does, we can make the comparison by casting to u64*, and compare the 16 bytes by two u64 comparisons.

rafaeldtinoco (approved):
LGTM now.

Issue #1852, opened 2022-06-23
[BUG] Error when no event derive


AlonZivony:
## Prerequisites

- [ ] This affects latest released version.
- [x] This affects current development tree (origin/HEAD).
- [x] There isn't an issue describing the bug.

Select one OR another:

- [x] I'm going to create a PR to solve this (assign to yourself).
- [ ] Someone else should solve this.

## Bug description

After #1822, derive function that decides to not derive an event is not supported unintentionally. 
The reason is because the returned value of the `deriveArgsFunction` is not checked to be nil.

This is the error I'm getting:
"failed to derive event 2001: error while building derived event 'container_create' - expected 8 arguments but given 0"

PR #1853, opened 2022-06-23
Bugfix: pkg/events/derive: add support to no derivation

AlonZivony:
## Initial Checklist

## Description

Fix bug that if derive event function does not need to derive event it
generates an error that it failed.
The reason is that there was no given way to not derive function.

Fixes: #1852 

## Type of change

## How Has This Been Tested?

- pkg/events/derive/derive_test - No derivation test for `singleEventDeriveFunc`.

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/events/derive/derive.go
+++ b/pkg/events/derive/derive.go
@@ -17,18 +17,25 @@ type eventSkeleton struct {
 	Params []trace.ArgMeta
 }
 
+// deriveArgsFunction is the main logic of the derived event.
+// It checks the event and produce the arguments of the derived event.
+// If no event is derived, then the returned args should equal `nil`.
 type deriveArgsFunction func(event trace.Event) ([]interface{}, error)
 
 // singleEventDeriveFunc create an events.DeriveFunction to generate a single derive trace.Event.
 // The event will be created using the original event information, the ID given and the arguments given.
 // The order of the arguments given will match the order in the event definition, so make sure the order match
 // the order of the params in the events.event struct of the event under events.Definitions.
+// If the arguments given is nil, than no event will be derived.
 func singleEventDeriveFunc(id events.ID, deriveArgsFunc deriveArgsFunction) events.DeriveFunction {
 	skeleton := makeEventSkeleton(id)
 	return func(event trace.Event) ([]trace.Event, []error) {
 		args, err := deriveArgsFunc(event)
 		if err != nil {
 			return nil, []error{err}
+		}
+		if args == nil {
+			return []trace.Event{}, nil
 		}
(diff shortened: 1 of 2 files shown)


pkg/events/derive/derive.go
```diff
@@ -17,19 +17,26 @@ type eventSkeleton struct {
...
+// deriveArgsFunction is the main logic of the derived event.
+// It checks the event and produce the arguments of the derived event.
+// If no event is derived, than the returned args should equal `nil`.
 type deriveArgsFunction func(event trace.Event) ([]interface{}, error)
 
 // singleEventDeriveFunc create an events.DeriveFunction to generate a single derive trace.Event.
 // The event will be created using the original event information, the ID given and the arguments given.
 // The order of the arguments given will match the order in the event definition, so make sure the order match
 // the order of the params in the events.event struct of the event under events.Definitions.
+// If the arguments given is nil, than no event will be derived.
```
mtcherni95, line 29: small typo: "then" :)

pkg/events/derive/derive.go
```diff
@@ -17,19 +17,26 @@ type eventSkeleton struct {
...
 func singleEventDeriveFunc(id events.ID, deriveArgsFunc deriveArgsFunction) events.DeriveFunction {
 	skeleton := makeEventSkeleton(id)
 	return func(event trace.Event) ([]trace.Event, []error) {
 		args, err := deriveArgsFunc(event)
 		if err != nil {
 			return nil, []error{err}
 		}
+		if args == nil {
+			return []trace.Event{}, nil
```
NDStrahilevitz, line 38: But now - `events.Derive` will receive an empty event! So the length of the returned events will be > 0, implying an event _was_ derived, despite what this is trying to achieve.
So I think this should return `nil, nil`,  and then it's good :)

NDStrahilevitz (changes requested):
So I think there may be some faulty logic in this fix regarding what will happen once this returns back to `events.Derive`. 
Can you please add a test there as well to make sure?

pkg/events/derive/derive.go
AlonZivony: You are wrong here - `events.Derive` will receive an empty list.
If you look at the implementation, it does `append`, which will result no appending because it is an empty list.

NDStrahilevitz: Ahh I missed the `[]` at the front you're right. So yeah should be good.

NDStrahilevitz (approved):
LGTM!

yanivagman (approved):
LGTM

Issue #1874, opened 2022-06-28
[BUG] No option to pass drop capabilities failure


AlonZivony:
## Prerequisites

- [x] This affects latest released version.
- [x] This affects current development tree (origin/HEAD).
- [x] There isn't an issue describing the bug.

Select one OR another:

- [x] I'm going to create a PR to solve this (assign to yourself).
- [ ] Someone else should solve this.

## Bug description

In some environments the operation of dropping capabilities is not allowed (for example, AWS lambdas).
Currently, we are not allowing tracee to run with high capabilities for security reasons, but it means that tracee is unusable in the mentioned environments.
Tracee should allow the user a way to run it knowing the security risk of it.

PR #1876, opened 2022-06-28
Bugfix: cmd: enable running with high capabilities by flag

AlonZivony:
## Initial Checklist

## Description (git log)

Enable tracee-ebpf and tracee-rules to keep running with high
capabilities if dropping capabilities failed, if configured to do so
by the user.
This will allow tracee to run in environment that don't allow
capabilities dropping.

Fixes: #1874 

## Type of change

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/builder/entrypoint.sh
+++ b/builder/entrypoint.sh
@@ -77,6 +77,7 @@ run_tracee_rules() {
         --cache cache-type=mem \
         --cache mem-cache-size=512 \
         --containers=${CONTAINERS_ENRICHMENT:="0"}\
+        --allow-high-capabilities=${ALLOW_HIGH_CAPABILITIES:="0"}\
         --trace event=${events} \
         --output=out-file:${TRACEE_PIPE} &
     tracee_ebpf_pid=$!
@@ -84,7 +85,11 @@ run_tracee_rules() {
     # start tracee-rules
 
     echo "INFO: starting tracee-rules..."
-    $TRACEE_RULES_EXE --metrics --input-tracee=file:${TRACEE_PIPE} --input-tracee=format:gob $@
+    $TRACEE_RULES_EXE\
+        --metrics --input-tracee=file:${TRACEE_PIPE}\
+        --input-tracee=format:gob\
+        --allow-high-capabilities=${ALLOW_HIGH_CAPABILITIES:="0"}\
+        $@
     TRACEE_RET=$?
 }
 
--- a/cmd/tracee-ebpf/main.go
+++ b/cmd/tracee-ebpf/main.go
@@ -34,6 +34,10 @@ var enrich bool
 
 var version string
 
+const (
+	allowHighCapabilitiesFlag = "allow-high-capabilities"
+)
+
 func main() {
 	app := &cli.App{
 		Name:    "Tracee",
@@ -149,7 +153,7 @@ func main() {
 			cfg.Output = &output
 
 			// environment capabilities
-			err = ensureCapabilities(OSInfo, &cfg)
+			err = ensureCapabilities(OSInfo, &cfg, c.Bool(allowHighCapabilitiesFlag))
 			if err != nil {
 				return err
 			}
@@ -368,6 +372,12 @@ func main() {
 				Name:        "containers",
 				Usage:       "enable container info enrichment to events. this feature is experimental and may cause unexpected behavior in the pipeline",
 				Destination: &enrich,
+			},
+			&cli.BoolFlag{
+				Name:    allowHighCapabilitiesFlag,
+				Aliases: []string{"ahc"},
+				Usage:   "allow tracee-ebpf to run with high capabilities, in case that capabilities dropping fails",
+				Value:   false,
 			},
 		},
 	}
--- a/cmd/tracee-rules/main.go
+++ b/cmd/tracee-rules/main.go
@@ -23,9 +23,10 @@ import (
 )
 
 const (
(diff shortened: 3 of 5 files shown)


NDStrahilevitz:
> Elaborate in the drop capabilities error message that the user can use the new flag to run with high capabilities. Also, add support for the flag in the docker entry-point (look at @NDStrahilevitz's PR).

It's #1886. But you could also just use the EnvVars field in the Flag struct, inside of using the wrapping technique I did, up to you.

AlonZivony:
> It's #1886. But you could also just use the EnvVars field in the Flag struct, inside of using the wrapping technique I did, up to you.

I chose to implement as you to keep on uniformed API.

cmd/tracee-rules/main.go
```diff
@@ -222,6 +228,11 @@ func main() {
 				Usage: "listening address of the metrics endpoint server",
 				Value: ":4466",
 			},
+			&cli.BoolFlag{
```
NDStrahilevitz, line 231: Maybe just add a shorter alias for the flag (like --ahc or something) for usability? Because it's a bit verbose rn.

rafaeldtinoco:
> Actually, we need to document the `EBPF_ALLOW_HIGH_CAPABILITIES` env flag somewhere, could you add docs for it? You correctly advise to use the flag in the cli but docker users won't be able to do that.

I think this is a "corner case" and we should have one session in "Prerequisites" session under installation.

image

cmd/tracee-ebpf/main.go
```diff
@@ -369,6 +373,12 @@ func main() {
 				Usage:       "enable container info enrichment to events. this feature is experimental and may cause unexpected behavior in the pipeline",
 				Destination: &enrich,
 			},
+			&cli.BoolFlag{
+				Name:    allowHighCapabilitiesFlag,
+				Aliases: []string{"ahc"},
```
NDStrahilevitz, line 378: Since you've added the alias here, maybe also in tracee-rules?

docs/install/prerequisites.md
```diff
@@ -39,3 +39,13 @@ capabilities:
...
+
+# Corner Cases
+
+## Drop Capabilities Error
+
+Tracee-ebpf and tracee-rules both try to reduce capabilities upon startup.
+Some environments won't allow capabilities dropping because of permission issues.
+Failure in capabilities dropping will result tracee's crash with a matching error, to guarantee that tracee isn't running with excess capabilities without the user agreement.
+To allow tracee to run with high capabilities and prevent crushing, the `--allow-high-capabilities` flag can be used.
+For docker users, to allow tracee-ebpf high capabilities the environment variable ` EBPF_ALLOW_HIGH_CAPABILITIES=1` should be used.
```
NDStrahilevitz, line 51: I think the env flag should affect both programs, so maybe just `ALLOW_HIGH_CAPABILITIES`, otherwise you would need to run the image both with the env flag and the `--ahc` flag itself following which isn't very intuitive UX.
And a nit: there's a space at the start of `EBPF_...`

builder/entrypoint.sh
```diff
@@ -77,6 +77,7 @@ run_tracee_rules() {
         --cache cache-type=mem \
         --cache mem-cache-size=512 \
         --containers=${CONTAINERS_ENRICHMENT:="0"}\
+        --allow-high-capabilities=${EBPF_ALLOW_HIGH_CAPABILITIES:="0"}\
```
NDStrahilevitz, line 80: Yep, so just add that to tracee-rules too and it's good, the env flag should enable both for a good UX.

NDStrahilevitz (commented):
Left some comments on improving the UX of the env flag, besides that the docs are good IMO.

docs/install/prerequisites.md
AlonZivony: The problem is, you won't be able to use the `--allow-high-capabilities` flag in the container, which reduce usabilitiy.
However, I think it might be better either way.
I need to check if I can make it so that it will use the flag with the env var only if it is defined.

NDStrahilevitz (approved):
LGTM!

docs/install/prerequisites.md
```diff
@@ -39,3 +39,13 @@ capabilities:
 
 [libbpf CO-RE documentation]: https://github.com/libbpf/libbpf#bpf-co-re-compile-once--run-everywhere
 [BTFHUB]: https://github.com/aquasecurity/btfhub-archive
+
+# Corner Cases
+
+## Drop Capabilities Error
+
+Tracee-ebpf and tracee-rules both try to reduce capabilities upon startup.
+Some environments won't allow capabilities dropping because of permission issues.
```
rafaeldtinoco, line 48: Could you better describe permission issues here ?

docs/install/prerequisites.md
```diff
@@ -39,3 +39,13 @@ capabilities:
...
+Failure in capabilities dropping will result tracee's crash with a matching error, to guarantee that tracee isn't running with excess capabilities without the user agreement.
```
rafaeldtinoco, line 49: Could we change "crash" wording here ? Its an error exit, right ? Not a crash ?

cmd/tracee-rules/main.go
rafaeldtinoco: Yep, this is very big.. maybe we should address this now ?

rafaeldtinoco (approved):
LGTM

Issue #1873, opened 2022-06-28
[BUG] Container enrichment feature causes lockup in the pipeline


rafaeldtinoco:
## Bug description

In order to test container enrichment impact in performance, and with a bit of my `chaos engineering hat`, I have created a **simple testcase** close to what we used to test **initial performance numbers** (when we found out ratio from producer->consumer).

The idea is simple:

- Before: we used to run `while true; do ps -ef; done`
- After: add containers running **same test case** for some time.

By starting the new test, I'm able to get up to 14MB/s with JSON output, a good indication that I'm doing what I wanted. After a very quick time, the pipeline stops ending events out and seems locked/stuck.

The process stack trace will be attached so it can be analyzed. I have also compiled tracee-ebpf with `-race` but it showed no apparent root cause for the lockup.

By manually removing `enrichContainerEvents` from the pipeline I'm able to do what I was supposed to: run `stress.sh`, stop it, see pipeline draining down, run `stress.sh` test again, see new events coming, pipeline being drained, etc.

## Steps to reproduce

Run tracee with:

```
$ sudo ./dist/tracee-ebpf -trace container -o format:json -o option:parse-arguments -o option:detect-syscall
```

Create a docker container with:

```Dockerfile
FROM ubuntu:jammy

ARG uid=1000
ARG gid=1000

# install needed environment

RUN export DEBIAN_FRONTEND=noninteractive && \
    apt-get update && \
    apt-get install -y sudo coreutils findutils && \
    apt-get install -y bash git curl rsync && \
    apt-get install -y llvm clang golang make gcc && \
    apt-get install -y linux-headers-generic && \
    apt-get install -y libelf-dev && \
    apt-get install -y zlib1g-dev

RUN export DEBIAN_FRONTEND=noninteractive && \
    apt-get update && \
    apt-get install -y man bash-completion && \
    apt-get install -y net-tools iproute2 wget vim && \
    apt-get install -y stress

# allow TRACEE* and LIBBPFGO* environment variables through sudo

RUN echo "Defaults env_keep += \"LANG LC_* HOME EDITOR PAGER GIT_PAGER MAN_PAGER\"" > /etc/sudoers && \
    echo "Defaults env_keep += \"LIBBPFGO* TRACEE*\"" >> /etc/sudoers && \
    echo "root ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers && \
[...]

PR #1877, opened 2022-06-28
enrich: fix enrichment queue locks

NDStrahilevitz:
## Initial Checklist

## Description (git log)

This PR solves two potential mutex locks that can occur in the enrichment pipeline. It also increases the queue sizes as a backup buffer.

```
Date:   Tue Jun 28 09:26:54 2022 +0000

    enrich: read queue only once at enrich start
    
    This commit includes two fixes:
    1. In the enrichment receiving part of this pipeline stage,
    the relevant queue is read with a mutex acquire at multiple points.
    To avoid unnecessary points of failure with that I've moved this
    queue acquisition to the top of the loop.
    2. In the enrich sending routine, the enrich is sent to a queue while
    a lock is acquired. If the queue is full the send will be stuck and
    the lock will not be unlocked. The queue is now read before sending the
    enrichment so the lock will always be available.
  
    This commit also increases the size of the enrichment queue buffers from
    1000 to 2500.
```

It also solves an edge case where an event could be sent to a closed enrichment queue:
```
Date:   Thu Jun 30 14:38:09 2022 +0000

    enrich: fix send on closed channel
    
    On some edge cases an event could be semnt to an enrich queue which
    was closed.
    Now, a map of stale queues exists to hold these queues so the sending
    routine can close them.
```

Fixes: #1873

## Type of change

## How Has This Been Tested?

This has been tested according to the procedure described in #1873. With this fix, the pipeline may get stuck for a while since the queues may be full, but no lock will occur, so removing the stress test will cause the pipeline to drain, unlike the noticed effect in #1873. 

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/ebpf/events_enrich.go
+++ b/pkg/ebpf/events_enrich.go
@@ -69,7 +69,7 @@ func (t *Tracee) enrichContainerEvents(ctx gocontext.Context, in <-chan *trace.E
 					queuesMutex.RUnlock()
 					if !exists {
 						queuesMutex.Lock()
-						queues[cgroupId] = make(chan *trace.Event, 1000)
+						queues[cgroupId] = make(chan *trace.Event, 2500)
 						queuesMutex.Unlock()
 
 						go func() {
@@ -78,8 +78,9 @@ func (t *Tracee) enrichContainerEvents(ctx gocontext.Context, in <-chan *trace.E
 						}()
 					}
 					queuesMutex.RLock()
-					queues[cgroupId] <- event
+					queue := queues[cgroupId]
 					queuesMutex.RUnlock()
+					queue <- event
 				}
 			}
 		}
@@ -93,15 +94,16 @@ func (t *Tracee) enrichContainerEvents(ctx gocontext.Context, in <-chan *trace.E
 			attemptedMutex.Lock()
 			attempted[cgroupId] = mutex // now attempted[cgroupId] is a transaction, only happens once and has begun
 			attemptedMutex.Unlock()
+			queuesMutex.RLock()
+			queue := queues[cgroupId]
+			queuesMutex.RUnlock()
 			if enrich.err != nil {
 				//only send error if it's not a non-existing cgroup error
 				if enrich.err.Error() != "no cgroup to enrich" {
 					t.handleError(enrich.err)
 				}
-				queuesMutex.RLock()
-				queue := queues[cgroupId]
-				queuesMutex.RUnlock()
-				for evt := range queues[cgroupId] {
+
+				for evt := range queue {
 					select {
 					case out <- evt:
 					case <-done:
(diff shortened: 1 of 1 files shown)


NDStrahilevitz:
@rafaeldtinoco I've reproduced your test from the issue with no problems this time running on this branch. If you can confirm as well I think it's GTG.

NDStrahilevitz:
@rafaeldtinoco @yanivagman Up to discussion: should we include #1892 in this?

rafaeldtinoco:
> > @rafaeldtinoco I've reproduced your test from the issue with no problems this time running on this branch. If you can confirm as well I think it's GTG.
> 
> I'm still able to reproduce using this PR code.

I believe we spoke it is "acceptable" to start blocking when resources reached a threshold (that is actually wanted I would say). So I'm not sure about that one.

We recently spoke: we enqueue events for a container being enriched until the enrichment happens. if that container's queue becomes full, because talking to the socket took too long, then it might be that we block the entire pipeline.

We could:

1) Add a bigger buffer to each container queue (if ever needed). If you think, the "caching" mechanism is exactly that, but for all the pipeline. We could either use something like that OR just rely in the channel buffering (remember that the caching in memory was needed because overhead of having a very big channel buffer wasn't as good as it).

2) Make sure that whatever we pick as a buffer is as big as the max throughput possibility during the enrichment timeout. Example: if we have 5 seconds for the timeout, and a single container being enriched, then the maximum theoretical throughput would be lets say 10MB/s. If the timeout is 5 seconds, the buffer has to have 50MB at least (with a margin) so the pipeline wouldn't be blocked until timeout happens.

Nevertheless, I think that those could be future enhancements and we could only do (2) using channel buffers for now (WDYT ?).

rafaeldtinoco:
@NDStrahilevitz BTW what is the status on this one ? Are we still locking ? Im asking because of a possible RC1 (we could merge something better than what we have and then fix it entirely in RC2). Or, if you think this is fully ready, I can test and merge if good.

rafaeldtinoco:
Ok, but be aware you're missing big rebase (--containers dont even exist in this version, and container name still had / on its name). I'm testing it now.

rafaeldtinoco:
> Ok, but be aware you're missing big rebase (--containers dont even exist in this version, and container name still had / on its name). I'm testing it now.

Ok I have spoken to Nadav and Im rebasing my branch for testing purposes. I was able to cause a very big latency in the pipeline, many enrichments failed (for X reason, probably runtime overloaded) and many delays in delivering events in the pipeline (events happening each couple of seconds and lots of lost events from perf buffer).

After couple of minutes without workload the pipeline went back to normal, CPU consumption reduced to 0% and memory footprint was reduced.

This might be good enough for RC1 and we did not have any lockup and /or clear race condition as well. We might need to reduce channel buffer (instead of increasing) so we fail faster but also are able to recover faster (independently of amount of lost events happening, which we can use caching mechanism to try to deal with bursts).

Nevertheless, we also have a feature enablement/disablement flag to help in future issues (disabling the feature if something goes south).

rafaeldtinoco:
@NDStrahilevitz like we spoke, we're going with this one in RC1 and discussing/improving/using https://github.com/aquasecurity/tracee/pull/1904 for RC2. I saw you changed buffers sizes recently. Are we good here ? Let me know when you feel we can merge this (we're releasing RC1 later today). Thanks.

PR #1885, opened 2022-06-28
Revert "fix leak"

NDStrahilevitz:
Reverts aquasecurity/tracee#1875.

It wasn't a leak it was tracee's event pipe so closing it caused e2e tests to fail.

### DIFF
--- a/cmd/tracee-rules/input.go
+++ b/cmd/tracee-rules/input.go
@@ -145,7 +145,6 @@ func parseTraceeInputFile(option *traceeInputOptions, fileOpt string) error {
 	if err != nil {
 		return fmt.Errorf("invalid file: %s", fileOpt)
 	}
-	defer f.Close()
 	option.inputFile = f
 	return nil
 }


cmd/tracee-rules/input.go
```diff
@@ -145,7 +145,6 @@ func parseTraceeInputFile(option *traceeInputOptions, fileOpt string) error {
 	if err != nil {
 		return fmt.Errorf("invalid file: %s", fileOpt)
 	}
-	defer f.Close()
```
mtcherni95, line 148: I see. When this will be closed then?

NDStrahilevitz: That's the entrypoint's job.

mtcherni95: also, why the e2e didn't fail once `defer f.Close()` was added?

NDStrahilevitz: For some reason the workflows on the PR didn't run, not sure why.

rafaeldtinoco: @mtcherni95 I saw it was a simple "fix" and the merge was green (I assumed tests had passed, which made me feel safe, but, for some reason we did not have tests in that PR, not sure why).

PR #1904, opened 2022-07-02
events_enrich: a new atomic approach with single lock

rafaeldtinoco:
## Initial Checklist

## Description (git log)

```
events_enrich: a new atomic approach with single lock

Producer:

Each cgroupId gets its own channel of events. Events are enqueued and only
de-dequed once the cgroupId was enriched OR has failed enrichment.

[cgroupId #0]: queue of trace events for the host
[cgroupId #1]: queue of trace events for container described by cgroupId #1
[cgroupId #2]: queue of trace events for container described by cgroupId #2
[cgroupId #3]: queue of trace events for container described by cgroupId #3
...

If the cgroupId channel does not exist it is created and an enrichment phase
is triggered for that cgroupId. If the cgroup channel is not used after a
specific period, it is removed.

Consumer:

In order to de-queue an event from its queue there is a scheduled operation
to each queued event.

queueReady: it is a simple scheduler of events using their cgroupId as index

[cgroupId #1][cgroupId #2][cgroupId #3][#cgroupId #2][cgroupId #3]...

One event is de-queued from queueReady only if its respective cgroupId has
been enriched OR has failed enrichment.

- - - -

Observations:

* With this model, the pipeline will only block when one of these
  channels is full. In this case, pipeline will be blocked until this
  channel's cgroupId is enriched and its enqueued events are de-queued.

* A good way to test this is to run the following:

  sudo ./dist/tracee-ebpf \
         --cache cache-type=mem \
         --cache mem-cache-size=1024 \
         --containers \
         -o format:json \
         -o option:parse-arguments \
         -o option:detect-syscall \
         -trace container \
         -trace event=openat \
         --crs docker:/var/run/docker.sock \
         | jq -c '. | {containerName}'

  during a stress test for container creations/deletions.

```

Fixes: https://github.com/aquasecurity/tracee/issues/1873

## Type of change

## How Has This Been Tested?

Described at: https://github.com/aquasecurity/tracee/issues/1873#issue-1286727048

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/ebpf/events_enrich.go
+++ b/pkg/ebpf/events_enrich.go
@@ -10,170 +10,158 @@ import (
 	"github.com/aquasecurity/tracee/types/trace"
 )
 
-const (
-	enrichResultQueueSize = 1000
-	enrichmentQueueSize   = 10000
-)
+//
+// Producer:
+//
+// Each cgroupId gets its own channel of events. Events are enqueued and only
+// de-dequed once the cgroupId was enriched OR has failed enrichment.
+//
+// [cgroupId #A]: queue of trace events for container described by cgroupId #A
+// [cgroupId #B]: queue of trace events for container described by cgroupId #B
+// [cgroupId #C]: queue of trace events for container described by cgroupId #C
+// ...
+//
+// If the cgroupId channel does not exist it is created and an enrichment phase
+// is triggered for that cgroupId. If the cgroup channel is not used after a
+// specific period, it is removed.
+//
+// Consumer:
+//
+// In order to de-queue an event from its queue there is a scheduled operation
+// to each queued event.
+//
+// queueReady: it is a simple scheduler of events using their cgroupId as index
+//
+// [cgroupId #A][cgroupId #B][cgroupId #C][#cgroupId #B][cgroupId #C]...
+//
+// One event is de-queued from queueReady only if its respective cgroupId has
+// been enriched OR has failed enrichment.
+//
+// Observation:
+//
+// With this model, the pipeline will only block when one of these channels is
+// full. In this case, pipeline will be blocked until this channel's cgroupId
+// is enriched and its enqueued events are de-queued.
+//
 
 func (t *Tracee) enrichContainerEvents(ctx gocontext.Context, in <-chan *trace.Event) (chan *trace.Event, chan error) {
+	const (
+		contQueueSize  = 10000  // max num of events queued per container
+		queueReadySize = 100000 // max num of events queued in total
+	)
+
 	type enrichResult struct {
-		cgroupId uint64
-		result   runtime.ContainerMetadata
-		err      error
+		result runtime.ContainerMetadata
+		err    error
 	}
 
-	queues := make(map[uint64]chan *trace.Event) // cgroupId and queues
-	queuesMutex := sync.RWMutex{}
-	attempted := make(map[uint64]*sync.Mutex)                  // cgroupId and enrichment attempt expressed as transaction lock
-	attemptedMutex := sync.RWMutex{}                           // big lock for attempted map
-	enriches := make(chan enrichResult, enrichResultQueueSize) // small buffer to reduce chance for blocking
-	//queues that are no longer used. we don't want to close on receiving side so it will mark queues as stale and sender can close
-	staleQueues := make(map[uint64]chan *trace.Event)
+	// big lock
+	bLock := sync.RWMutex{}
+	// pipeline channels
 	out := make(chan *trace.Event, 10000)
 	errc := make(chan error, 1)
-	done := make(chan struct{}, 1)
+	// state machine for enrichment
+	enrichDone := make(map[uint64]bool)
+	enrichInfo := make(map[uint64]*enrichResult)
+	// 1 queue per cgroupId
+	queues := make(map[uint64]chan *trace.Event)
+	// scheduler queue
+	queueReady := make(chan uint64, queueReadySize)
 
+	// queues map writer
 	go func() {
 		defer close(out)
 		defer close(errc)
-		for {
+		for { // enqueue events
 			select {
-			case <-ctx.Done():
-				done <- struct{}{}
-				return
 			case event := <-in:
+				// send out irrelevant events
+				if event.ContainerID == "" {
+					out <- event
+					continue
+				}
 				cgroupId := uint64(event.CgroupID)
-				// cgroup_mkdir event: need the cgroupId from its argument
+				// CgroupMkdir: pick EventID from the event itself
 				if event.EventID == int(events.CgroupMkdir) {
 					cgroupId, _ = parse.ArgUint64Val(event, "cgroup_id")
 				}
-				// cgroup_rmdir: need to clean attempt so cgroupId can be reused
+				// CgroupRmdir: clean up remaining events and maps
 				if event.EventID == int(events.CgroupRmdir) {
+					var eInfo *enrichResult
 					cgroupId, _ = parse.ArgUint64Val(event, "cgroup_id")
-					attemptedMutex.Lock()
-					delete(attempted, cgroupId)
-					attemptedMutex.Unlock()
-				}
-				// non container event and not cgroup_mkdir (or the event already enriched) skip per cgroupId caching
-				if (event.ContainerID == "" && event.EventID != int(events.CgroupMkdir)) || event.ContainerImage != "" {
-					// enrichment attempt is being done, or has failed, so do not try it again
-					attemptedMutex.RLock()
-					if _, ok := attempted[cgroupId]; ok {
-						// wait transaction (cache flush) to finish so cached events and this one are ordered
-						attempted[cgroupId].Lock()
-						out <- event
-						attempted[cgroupId].Unlock()
-						attemptedMutex.RUnlock()
-						continue
+					bLock.RLock()
+					if i, ok := enrichInfo[cgroupId]; ok {
+						eInfo = i
 					}
-					attemptedMutex.RUnlock()
-					out <- event // some events might not have an attempted map
-					continue
-				} else {
-					//if the container queue has not been created - create the container queue and invoke to enrich query
-					queuesMutex.RLock()
-					_, exists := queues[cgroupId]
-					queuesMutex.RUnlock()
-					if !exists {
-						queuesMutex.Lock()
-						queues[cgroupId] = make(chan *trace.Event, enrichmentQueueSize)
-						queuesMutex.Unlock()
-
-						go func() {
-							metadata, err := t.containers.EnrichCgroupInfo(cgroupId)
-							enriches <- enrichResult{cgroupId, metadata, err}
-						}()
-					}
-					queuesMutex.RLock()
-					queue, ok := queues[cgroupId]
-					queuesMutex.RUnlock()
-					if ok {
-						//if queue still exists send as usual
-						queue <- event
-					} else {
-						//if not, get the enriched data and handle the stale queue
-						info := t.containers.GetCgroupInfo(cgroupId)
-						queuesMutex.RLock()
-						queue = staleQueues[cgroupId]
-						queuesMutex.RUnlock()
-
-						//probably empty but just to make sure
-						for evt := range queue {
-							enrichEvent(evt, info.Container)
-							out <- evt
+					for evt := range queues[cgroupId] {
+						if evt.ContainerImage == "" && eInfo != nil {
+							bLock.RUnlock()
+							enrichEvent(evt, eInfo.result)
+							bLock.RLock()
 						}
-
-						enrichEvent(event, info.Container)
-
-						close(queue)
-						queuesMutex.Lock()
-						delete(staleQueues, cgroupId)
-						queuesMutex.Unlock()
-						out <- event
+						out <- evt
 					}
+					bLock.RUnlock()
+					bLock.Lock()
+					delete(enrichDone, cgroupId)
+					delete(enrichInfo, cgroupId)
+					delete(queues, cgroupId)
+					bLock.Unlock()
+					continue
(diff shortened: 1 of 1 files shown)


rafaeldtinoco:
@NDStrahilevitz this is based on your work but I wanted to try another approach with only atomic operations and channels (easier to maintain and less susceptible to races).

At the end, I had to add a single lock to guarantee map reads/writes operations, but everything else is either atomic compare/exchange (primitives) or channel ops.

I have done a few tests and recorded so you could check [HERE](https://drive.google.com/file/d/1qnW19IAMwX9tZ_7lVEbhxXQZPBtbo5yS/view?usp=sharing)

rafaeldtinoco:
@NDStrahilevitz tests look good. I was able to "recover" the pipeline after stressed a lot (up to 1.5GB of ram because of caching events). 

@yanivagman will also take a look at this.

rafaeldtinoco:
image

While still caching events is above.

Then when we started loosing events due to full cache:

image


Fully blocked by full cache and full channel buffers:

image

NDStrahilevitz (approved):
LGTM!

OriGlassman:
> @yanivagman May I get your input as well ? I know the perf tests can be tricky, but a general look and observation if you can observe some corner case we haven't, etc.
> 
> @OriGlassman are you able to test this and check if it mitigates/solves #1906 ?

Hi, as I wrote in #1906, it did not mitigate the issue.

rafaeldtinoco:
> > @yanivagman May I get your input as well ? I know the perf tests can be tricky, but a general look and observation if you can observe some corner case we haven't, etc.
> > @OriGlassman are you able to test this and check if it mitigates/solves #1906 ?
> 
> Hi, as I wrote in #1906, it did not mitigate the issue.

Thanks @OriGlassman , I failed to see that update. Thanks a lot.

pkg/ebpf/events_enrich.go
```diff
@@ -8,172 +8,157 @@ import (
...
-						info := t.containers.GetCgroupInfo(cgroupId)
-						queuesMutex.RLock()
-						queue = staleQueues[cgroupId]
-						queuesMutex.RUnlock()
-
-						//probably empty but just to make sure
-						for evt := range queue {
-							enrichEvent(evt, info.Container)
-							out <- evt
+					i := enrichInfo[cgroupId]
```
yanivagman, line 94: what if `enrichInfo[cgroupId]` entry doesn't exist?
This might happen, for example, when starting tracee with pre-existing containers which are then removed

One more point - the (read) lock is not taken for this operation - shouldn't it be?

pkg/ebpf/events_enrich.go
```diff
@@ -8,172 +8,157 @@ import (
...
+					delete(enrichDone, cgroupId)
+					delete(enrichInfo, cgroupId)
+					delete(queues, cgroupId)
+					bLock.Unlock()
+					continue
 				}
+				// make sure a queue channel exists for this cgroupId
+				if _, ok := queues[cgroupId]; !ok {
+					queueSize := contQueueSize
+					bLock.Lock()
```
yanivagman, line 111: I think you should take the lock before the check
`if _, ok := queues[cgroupId]; !ok {`
or otherwise, you might perform the above two lines more than once for the same cgroupId

pkg/ebpf/events_enrich.go
```diff
@@ -8,172 +8,157 @@ import (
...
+					queues[cgroupId] = make(chan *trace.Event, queueSize)
+					enrichDone[cgroupId] = &uberAtom.Bool{}
+					bLock.Unlock()
+					go func(cgroupId uint64) {
+						metadata, err := t.containers.EnrichCgroupInfo(cgroupId)
+						bLock.Lock()
+						enrichInfo[cgroupId] = &enrichResult{metadata, err}
+						enrichDone[cgroupId].Toggle()
```
yanivagman, line 119: You probably want to Store a `true` value, and not toggle, which negates the previous value (double negation will cause it to be false again)

Also - Not necessarily for this PR, but we might want to try again in case of a failure (the container runtime daemon might not be availalbe once, but maybe after for whatever reason - need to check the returned error code)

pkg/ebpf/events_enrich.go
```diff
@@ -8,172 +8,157 @@ import (
...
+					bLock.Lock()
+					delete(enrichDone, cgroupId)
+					delete(enrichInfo, cgroupId)
+					delete(queues, cgroupId)
+					bLock.Unlock()
+					continue
 				}
+				// make sure a queue channel exists for this cgroupId
+				if _, ok := queues[cgroupId]; !ok {
+					queueSize := contQueueSize
```
yanivagman, line 110: any reason for having this variable and not directly using contQueueSize?

pkg/ebpf/events_enrich.go
```diff
@@ -8,172 +8,157 @@ import (
...
-						queuesMutex.RLock()
-						queue = staleQueues[cgroupId]
-						queuesMutex.RUnlock()
-
-						//probably empty but just to make sure
-						for evt := range queue {
-							enrichEvent(evt, info.Container)
-							out <- evt
+					i := enrichInfo[cgroupId]
+					for evt := range queues[cgroupId] {
```
yanivagman, line 95: shouldn't a (read) lock be taken before accessing queues?

pkg/ebpf/events_enrich.go
```diff
@@ -8,172 +8,157 @@ import (
...
 					}
+					bLock.Lock()
+					delete(enrichDone, cgroupId)
+					delete(enrichInfo, cgroupId)
+					delete(queues, cgroupId)
+					bLock.Unlock()
+					continue
 				}
+				// make sure a queue channel exists for this cgroupId
+				if _, ok := queues[cgroupId]; !ok {
```
yanivagman, line 109: same question about the read lock.
Also - I see that you access queues[cgroupId] twice, so you can probably just save the channel once, and use it here again without the need to lock again.

pkg/ebpf/events_enrich.go
```diff
@@ -8,172 +8,157 @@ import (
...
-					if len(queue) < 1 {
-						break
+		for { // de-queue events
+			select {
+			case cgroupId := <-queueReady: // queue for received cgroupId is ready
+				bLock.RLock()
+				if !enrichDone[cgroupId].Load() {
+					bLock.RUnlock()
+					// re-schedule the operation if queue is not enriched
+					queueReady <- cgroupId
```
yanivagman, line 142: So if I understand correctly, this will busy loop until container is enriched because it is wrapped by the above:
`case cgroupId := <-queueReady:`
If that's the case, is it really what we want (busy-looping)?

pkg/ebpf/events_enrich.go
```diff
@@ -8,172 +8,157 @@ import (
...
+				} else {
+					// de-queue event if queue is enriched
+					event := <-queues[cgroupId]
+					bLock.RUnlock()
+					if event.ContainerImage == "" {
+						// event is not enriched: enrich if enrichment worked
+						bLock.RLock()
+						i := enrichInfo[cgroupId]
```
yanivagman, line 150: shouldn't happen at that point, but better checking that `enrichInfo[cgroupId]` is valid before accessing i.result

yanivagman (commented):
I like this change very much.
Few comments follow

NDStrahilevitz:
@yanivagman About the read lock comments - in general on the sending side, there's no need for the read locks because the consuming side doesn't do any writes, so that side gets to do the read locks. 
Might be better to add them back just in case in the future we would want to to do writes in the consuming side, but currently I think we can do without them.

pkg/ebpf/events_enrich.go
NDStrahilevitz: The other option is doing what I previously did which was queueing on enrichments, and then emptying the queues, but that requires coordinating stale queues etc. which is what we wanted to avoid.

pkg/ebpf/events_enrich.go
NDStrahilevitz: Just noting - it's a leftover from before my PR to this.

pkg/ebpf/events_enrich.go
NDStrahilevitz: Just noting we also agreed to not do retries in the original PR, but I do think we should open an issue on it.

pkg/ebpf/events_enrich.go
rafaeldtinoco: Yep, added a read lock and addressed the map key existence.

pkg/ebpf/events_enrich.go
rafaeldtinoco: yep.

pkg/ebpf/events_enrich.go
rafaeldtinoco: fixed.

rafaeldtinoco:
Okay, so, with the assumption:

- concurrent map reads are ok
- map read can't occur while map is being written

We have 3 maps:

1. enrichDone
2. enrichInfo
3. queueReady

I added read and write locks to EACH map reference.

I have then joined duplicate locks.

I have then removed LockUnlockLockUnlock approaches.

I believe it addresses all the mentioned cases.

pkg/ebpf/events_enrich.go
rafaeldtinoco: Plus (to what Nadav said) we're not busy looping during enrichment, we're busy looping in a "simple scheduler" iteration. And it mixes events from multiple containers, if they're coming through channel, giving chance for all containers events to be processed.

rafaeldtinoco:
BTW, the fact that we have events triggering enrichment together with all other events is bad because it causes jittering in cgroup mkdir/rmdir events, causing the enrichment to fail because some containers might have already died.

That is why I had the "idea" before of having a dedicated janitor events channel (even if we duplicate the event through the 2 channels, 1 for janitor events only, the other containing all selected events). 

We could enrich based on the "janitor events channel", causing the enrichment to happen as soon as that event happens (and not jittered by many events coming in the same pipe).

NDStrahilevitz:
> As a side note: having everything in the same pipeline (events AND service-type events) is sub-optimal in cases like this. The events being stuffed into the same ring is causing jittering in the service-type events, such as cgroup mkdir/rmdir, that are triggering the enrichment.
> 
> Example: if we have many container creations/deletions and the perf buffer is full, we might try to enrich a container that is long gone.
> 
> For the container feature it might not be _that_ problematic, because the containers might not be _that_ short-lived (such as the stress test being used by this issue).

We could (maybe even in the next release) fan out the channel into the processEvents stage to parallelize it in the pipeline. Currently all our stages are sequential but it doesn't have to be that way. We may have to revise this code again afterwards but I believe it could be even simpler if it's done.

rafaeldtinoco:
@yanivagman I'm afraid we need to merge this today, I believe I have addressed your concerns, was a bit more cautious and conservative regarding locks (and tested it all again).

@NDStrahilevitz if you could +1 it again (no big changes from your last reviews).

Thanks

pkg/ebpf/events_enrich.go
```diff
@@ -8,172 +8,163 @@ import (
...
+				if _, ok := queues[cgroupId]; !ok {
+					queues[cgroupId] = make(chan *trace.Event, contQueueSize)
+					enrichDone[cgroupId] = &uberAtom.Bool{}
+
+					go func(cgroupId uint64) {
+						metadata, err := t.containers.EnrichCgroupInfo(cgroupId)
+						bLock.Lock()
+						enrichInfo[cgroupId] = &enrichResult{metadata, err}
+						enrichDone[cgroupId].Store(true)
+						bLock.Unlock()
```
NDStrahilevitz, line 126: I'm pretty sure there's a race condition that can happen between this line 132. And this race condition would panic tracee if i'm right that it can happen.
I think you should just lock 118-119 and only read lock the rest.

pkg/ebpf/events_enrich.go
```diff
@@ -8,172 +8,163 @@ import (
 	"github.com/aquasecurity/tracee/pkg/events"
 	"github.com/aquasecurity/tracee/pkg/events/parse"
 	"github.com/aquasecurity/tracee/types/trace"
-)
 
-const (
-	enrichResultQueueSize = 1000
-	enrichmentQueueSize   = 10000
+	uberAtom "go.uber.org/atomic"
```
NDStrahilevitz, line 12: nit: maybe just use the default `atomic` here?

NDStrahilevitz (changes requested):
Pretty sure there's a race condition here. I don't think we should be afraid of just RLocking.

pkg/ebpf/events_enrich.go
yanivagman: If I understand correctly the semantics of this rwmutex,
This is a different go routine.
So from what I understand, this code won't run until the current go routine will release it's lock in 132.

NDStrahilevitz: You're right just went over this with Rafael, it will probably be fine. I'd still fine grain that lock just to be sure, but we'll probably be fine without.

rafaeldtinoco: Yep, its a diff routine and it can Lock/Unlock. I could Unlock()/Lock() before 130 but we're fine (same numbers from previous test).

rafaeldtinoco:
Already, addressed the "no need for atomic because of new locking approach", and made locking fine grained after calling the parallel routine, and fixed the ctx.Done extra channel unneeded.

NDStrahilevitz (approved):
LGTM!

yanivagman (approved):
LGTM as well!

rafaeldtinoco:
Thank you, squashed and reworded gitlog to fix it. Merging it now.

Issue #1900, opened 2022-06-30
[BUG] Cgroup 1 doesn't exist when running from container


NDStrahilevitz:
## Prerequisites

- [x] This affects latest released version.
- [x] This affects current development tree (origin/HEAD).
- [ ] There isn't an issue describing the bug.

Select one OR another:

- [x] I'm going to create a PR to solve this (assign to yourself).
- [ ] Someone else should solve this.

## Bug description

During performance tests for the close release I have noticed that when running tracee in a container I get this pprof:

<img width="1377" alt="image" src="https://user-images.githubusercontent.com/22661609/176721743-eb4e5d74-d3b1-4d56-be3f-518e8454af69.png">

I am not sure why, but adding a log here (containers.go:441):
<img width="622" alt="image" src="https://user-images.githubusercontent.com/22661609/176721988-a425d874-b844-4680-8382-b0c52596e090.png">

Gave the following results when running from a container:
<img width="262" alt="image" src="https://user-images.githubusercontent.com/22661609/176722120-93eb477a-eed2-4294-a415-b35cda60400b.png">

Since cgroup 1 should always be the host afaik, we should mitigate these checks for that id.

## Steps to reproduce

1. Add the log described above
2. Run tracee in a container
3. Observe the logs

## Context

Relevant information about my setup:

* Linux version: Ubuntu 21.04
* Linux kernel version: 5.11.0-1027-aws
* Tracee version (or commit id of your tree): HEAD
* LLVM version:
* Golang version: 1.17

yanivagman:
Related issue, other symptoms: https://github.com/aquasecurity/tracee/issues/1338#issuecomment-1027955397

yanivagman:
Mounting cgroup v1 cpuset controller in the entrypoint (in case of cgroup v1 environment) should resolve both this issue and #1338

Issue #1338, opened 2022-01-12
Sub command "trace" can not work with option --trace c


vincent-pli:
I try sub command `trace` with
```
docker run --name tracee --rm --pid=host --cgroupns=host --privileged -v /tmp/tracee:/tmp/tracee -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro -it aquasec/tracee:0.6.5 trace --trace c
```
But no event collected:
```
TIME             CONTAINER_ID  UID    COMM             PID/host        TID/host        RET              EVENT                ARGS
```

I expected it will be many event since I run it in a kubernetes node.
Checked some code, with option `--trace c`, `--trace container` `--trace !container` .etc,  the code will go through folders in path `/sys/fs/cgroup/cpuset`(cgroupv1) to collect `containerID`,  that's make sense, but we do not mount host's `/sys/fs/cgroup/` to the `tracee`, the code go through the file system of the tracee container rather than the host.

To change the command like this, it will work:
```
docker run --name tracee --rm --pid=host --cgroupns=host --privileged -v /tmp/tracee:/tmp/tracee -v /lib/modules/:/lib/modules/:ro -v /usr/src:/usr/src:ro  -v /sys/fs/cgroup/:/sys/fs/cgroup/:ro  -it  aquasec/tracee:0.6.5 trace --trace c
```

yanivagman:
Well that's strange.
`cgoupns=host` should have taken care of that as it then uses the same cgroupfs as the one on the host.
Can you please exec into the container after it is started and `ls /sys/fs/cgroup/cpuset`?

PR #1915, opened 2022-07-04
containers: mount cpuset on cgroupsv1

NDStrahilevitz:
## Initial Checklist

## Description (git log)

```
Date:   Mon Jul 4 15:29:33 2022 +0000

    containers: mount cpuset on cgroupsv1
    
    When cgroupsV1 is used, the cgroup controller is located at the host
    /sys/fs/cgroup/cpuset. However, this fs is duplicated to the container's
    filesystem and used instead of it when running tracee-ebpf
    from a container. As a result, a different root cgroupId is given,
    and existing cgroups are not detected for population on startup.
    
    This fix makes tracee create a local mountpoint for the host's cpuset
    if it detects usage of cgroupsV1.
```

Fixes: #1900
Fixes: #1338

## Type of change

## How Has This Been Tested?

You can test with either of these ways:
1. Add the log described in #1900, build tracee to a container and run. The log should be empty of the repeated cgroup 1 checks.
2. I. Start a redis container `docker run -d --rm redis`
    II. Run tracee as a container with `trace -t event=existing_container` and confirm you get a result 

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/containers/containers.go
+++ b/pkg/containers/containers.go
@@ -8,7 +8,6 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
-	"strconv"
 	"strings"
 	"sync"
 	"syscall"
@@ -22,6 +21,8 @@ import (
 var cgroupV1HierarchyID int
 
 const cgroupV1Controller = "cpuset"
+const cgroupV1FsType = "cgroup"
+const cgroupV2FsType = "cgroup2"
 
 // Containers contains information about running containers in the host.
 type Containers struct {
@@ -45,10 +46,15 @@ type CgroupInfo struct {
 // InitContainers initializes a Containers object and returns a pointer to it.
 // User should further call "Populate" and iterate with Containers data.
 func InitContainers(sockets cruntime.Sockets, debug bool) (*Containers, error) {
-	cgroupV1 := false
+	containers := &Containers{
+		cgroupV1: false,
+		cgroupMP: "",
+		cgroups:  make(map[uint32]CgroupInfo),
+		mtx:      sync.RWMutex{},
+	}
+
 	if _, err := os.Stat("/sys/fs/cgroup/cgroup.controllers"); os.IsNotExist(err) {
-		cgroupV1 = true
-		if err := initCgroupV1(); err != nil {
+		if err := containers.initCgroupV1(); err != nil {
 			return nil, err
 		}
 	}
@@ -69,13 +75,9 @@ func InitContainers(sockets cruntime.Sockets, debug bool) (*Containers, error) {
 		fmt.Fprintf(os.Stderr, "Enricher: %v\n", err)
 	}
 
-	return &Containers{
-		cgroupV1: cgroupV1,
-		cgroupMP: "",
-		cgroups:  make(map[uint32]CgroupInfo),
-		mtx:      sync.RWMutex{},
-		enricher: runtimeService,
-	}, nil
+	containers.enricher = runtimeService
+
+	return containers, nil
 }
 
 func (c *Containers) IsCgroupV1() bool {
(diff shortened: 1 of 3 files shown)


pkg/containers/utils.go
```diff
@@ -0,0 +1,83 @@
...
+	"strconv"
+	"strings"
+	"syscall"
+)
+
+// runsOnContainerV1 tests if process is running on a container in cgroupsv1
+// it tests by checking for existance of a release_agent file in cpuset
+// from https://www.kernel.org/doc/Documentation/cgroup-v1/cgroups.txt:
+//
+// - release_agent: the path to use for release notifications (this file exists in the top cgroup only)
```
yanivagman, line 17:
```suggestion
// - release_agent: ... this file exists in the top cgroup only ...
```

pkg/containers/utils.go
```diff
@@ -0,0 +1,83 @@
...
+	if os.IsNotExist(err) {
+		return true, nil
+	}
+	if err != nil {
+		return false, err
+	}
+	return false, nil
+}
+
+// mountsCpuset mounts /sys/fs/cgroup/cpuset to the container's cpuset fs
```
yanivagman, line 32:
```suggestion
// mountsCpuset mounts cpuset cgroup controller to the container's current working directory
```

pkg/containers/utils.go
```diff
@@ -0,0 +1,83 @@
...
+// this is done in order to detect host cgroup data when running tracee from a container in cgroupsv1
+func mountCpuset(mountPath string) error {
+	isMp, _ := isMountpoint(mountPath)
+
+	if isMp {
+		return nil
+	}
+
+	// equivalent to mount -t cgroup -o cpuset cgroup /sys/fs/cgroup/cpuset
```
yanivagman, line 41:
```suggestion
	// equivalent to mount -t cgroup -o cpuset cgroup cpuset
```

pkg/containers/utils.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	return nil
+}
+
+func isMountpoint(path string) (bool, error) {
+	mountpointCmd := fmt.Sprintf("mountpoint %s | grep \"is a mountpoint\"", path)
+	out, err := exec.Command("sh", "-c", mountpointCmd).Output()
+	if err != nil {
+		return false, err
+	}
```
yanivagman, line 53: Let's avoid invoking external commands.
Instead, you can look for the path in `/proc/mounts`

pkg/containers/utils.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// mountsCpuset mounts /sys/fs/cgroup/cpuset to the container's cpuset fs
+// this is done in order to detect host cgroup data when running tracee from a container in cgroupsv1
+func mountCpuset(mountPath string) error {
+	isMp, _ := isMountpoint(mountPath)
+
+	if isMp {
+		return nil
+	}
```
yanivagman, line 39:
```suggestion
	if isMp, _ := isMountpoint(mountPath); isMp {
	        // already mounted
		return nil
	}
```

pkg/containers/containers.go
```diff
@@ -98,28 +97,24 @@ func (c *Containers) Populate() error {
...
+	inContainer, err := runsOnContainerV1()
+	if err != nil {
+		return err
 	}
 
-	return fmt.Errorf("couldn't determine cgroup v1 hierarchy id")
+	if inContainer {
+		os.Mkdir("./cpuset", 0644)
+		err = mountCpuset("./cpuset")
+		return err
```
yanivagman, line 114: same as above - shouldn't be a fatal error

yanivagman (changes requested):
I didn't see in the code that the newly mounted dir is used by containers.go (specifically by the `InitContainers()` function).
Does this code really resolves the issues at its current state?

pkg/containers/utils.go
NDStrahilevitz: actually should be `mountPath` in the end.

pkg/containers/utils.go
yanivagman: Invoking external commands is not a good security practice, although you are right that we control the image files.
Yet it is more perfomant to simply read the /proc/mounts file, and we already do that in procMountsCgroups() so you can simply use the same code, without invoking external commands.

yanivagman:
> > I didn't see in the code that the newly mounted dir is used by containers.go (specifically by the `InitContainers()` function). Does this code really resolves the issues at its current state?
> 
> It does, later down the line when populating existing cgroups - the code searches for the cgroups mountpoints in `/proc/mounts` (the call happens at containers.go:90).

But I see in this function that the last match in /proc/mounts "wins", shouldn't we change that in some way to be more deterministic now that there is more than one mountpoint from cpuset cgroup mountpoint?

NDStrahilevitz:
> But I see in this function that the last match in /proc/mounts "wins", shouldn't we change that in some way to be more deterministic now that there is more than one mountpoint from cpuset cgroup mountpoint?

I was under the assumption the mount point is replaced, seems like that's not the case so I agree, i'm thinking explicit search targets is the solution.

pkg/containers/containers.go
```diff
@@ -88,63 +90,64 @@ func (c *Containers) GetCgroupV1HID() int {
...
-	for i := 1; scanner.Scan(); i++ {
-		sline := strings.Fields(scanner.Text())
-		if len(sline) < 2 || sline[0] != cgroupV1Controller {
-			continue
+	inContainer, err := runsOnContainerV1()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "containers: failed to detect if running on a cgroupv1 container: %s", err)
+	}
+
+	if inContainer {
```
yanivagman, line 117: if `err != nil` above, the value of this variable is invalid, isn't it?

pkg/containers/containers.go
```diff
@@ -88,63 +90,64 @@ func (c *Containers) GetCgroupV1HID() int {
...
+			fmt.Fprintf(os.Stderr, "failed to get absolute path of cputset mountpoint %s\n", mountPath)
+			return nil
 		}
-		intID, err := strconv.Atoi(sline[1])
-		if err != nil || intID < 0 {
-			return fmt.Errorf("error parsing %s: %w", cgroupsFile, err)
+
+		err = mountCpuset(mountPath)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "containers: failed to mount cgroupv1 controller: %s\n", err)
```
yanivagman, line 127: if you continue and set `c.cgroupMP = mountPath` below althogh `err != nil` here it will be invalid

pkg/containers/utils.go
```diff
@@ -0,0 +1,119 @@
...
+		if err != nil || intID < 0 {
+			return -1, fmt.Errorf("error parsing %s: %w", cgroupsFile, err)
+		}
+		return intID, nil
+	}
+	return -1, fmt.Errorf("couldn't determine cgroup v1 hierarchy id")
+}
+
+// searchMountpoint searches if path is a mountpoint for fstype
+func searchMountpoint(fstype string, path string) (bool, error) {
```
yanivagman, line 76: `searchMountpoint()` has the same meaning as `findMountpoint()` defined below.
I would rename the function to either `isMountpoint` or `checkMountpoint`.

pkg/containers/containers.go
NDStrahilevitz: You're right i've added a return above inside the if.

pkg/containers/containers.go
NDStrahilevitz: By convention the bool returned in such a case would be false, which is also the case here. I can add a return there anyway if you want.

NDStrahilevitz: Added the early return.

yanivagman (approved):
LGTM
@rafaeldtinoco please take a look as well

NDStrahilevitz:
@yanivagman Talked with Rafael offline, he said he will skip reviewing this one, checked it on a cgroup v2 machine now as well and it's working good so i'm merging.

Issue #1483, opened 2022-02-16
TRC-8 - how add process allowlist


krol3:
I am testing tracee in AKS, I would like to add process to the allow list. Similar like this: https://github.com/aquasecurity/tracee/blob/main/signatures/rego/k8s_service_account_token.rego#L40

Could you guide me, how to set this rule?

Samples of detection, that I would like to add to the allow list in the rule TRC-8

```
*** Detection ***
Time: 2022-02-16T19:39:21Z
Signature ID: TRC-8
Signature: K8S Service Account Token Use Detected
Data: map[]
Command: telegraf
Hostname: omsagent-frlw6

*** Detection ***
Time: 2022-02-16T19:38:56Z
Signature ID: TRC-8
Signature: K8S Service Account Token Use Detected
Data: map[]
Command: kubectl
Hostname: tunnelfront-f84

*** Detection ***
Time: 2022-02-16T19:38:28Z
Signature ID: TRC-8
Signature: K8S Service Account Token Use Detected
Data: map[]
Command: telegraf
Hostname: omsagent-x5vdg

*** Detection ***
Time: 2022-02-16T19:38:30Z
Signature ID: TRC-8
Signature: K8S Service Account Token Use Detected
Data: map[]
Command: starboard-opera
Hostname: starboard-opera
```

roikol:
hey @krol3 ,
in order to add process names to the allowlist, you need to manually add the process names you want to the allowlist here: https://github.com/aquasecurity/tracee/blob/main/signatures/rego/k8s_service_account_token.rego#L40
and then compile and run.

otherwise, you can change the signature after tracee is compiled - the path from which this signature is taken from after compilation is TRACEE_DIR/dist/rules/k8s_service_account_token.rego

itaysk:
related: https://github.com/aquasecurity/tracee/issues/1423

itaysk:
@roikol should we add these to the existing allowlist?

roikol:
@itaysk 
i'm only comfortable with adding `kubectl`. the other processes are not part of the Kuberenetes infrastructure.
i'll push it now.

itaysk:
Is it though? Feels to me that kuebctl in the server is an IOC @raesene wdyt?

raesene:
[...]

PR #1916, opened 2022-07-04
Disable noisy signatures

rafaeldtinoco:
## Initial Checklist

## Description (git log)

signatures/rego: disable TRC-8 for being too noisy

TRC-8 signature needs better tuning, depending on user's environment.
Unfortunately, until GO-CEL is fully supported, we can't configure
signatures values when tracee is distributed as a docker container.

For now, it is better to keep TRC-8 disabled until that is sorted out.

## Type of change

- Issue mitigation (not a fix) for issue: #1483.

## How Has This Been Tested?

https://github.com/aquasecurity/tracee/issues/1483#issuecomment-1144556600

## Final Checklist:

## Git Log Checklist:

My commits logs have:


rafaeldtinoco:
@AsafEitani mind giving me a +1 here if you agree ? This signature is too noisy and requires manual changes that can't be achieved the way we distribute the signatures currently (go-cel signatures will be capable of allowing such customizations as you already know and can be seen in the merged POC).

AsafEitani (approved):
Agreed
I don't know which logic we can introduce with gocel that can solve this, it's a problem that we don't know all the 3rd party binaries of kubernetes that might do that.
Possibly should remove this signature altogether if it's not giving us the value we want.

rafaeldtinoco:
> Agreed I don't know which logic we can introduce with gocel that can solve this, it's a problem that we don't know all the 3rd party binaries of kubernetes that might do that. Possibly should remove this signature altogether if it's not giving us the value we want.

go-cel signatures will allow end user to change values in a signature (CHANGE-HERE approach). Currently there is no way end user could change the signature values without re-generating tracee container, for example.

I'll keep the files for now until signature is migrated, then we can decide which signatures to keep/add.

Thanks for checking this.

Issue #1655, opened 2022-04-10
Get loaded shared object exported symbols


AlonZivony:
By reading the symbols of the exported shared object, a user can find hooks of normal libraries.
This could be implemented as a derived event of `shard_object_loaded` event (#1631)

PR #1956, opened 2022-07-11
Feature/so loader

AlonZivony:
## Initial Checklist

#### pkg/utils/shared_objects: load dynamic symbols
Created a struct using Tracee's cache for mount ns process to resolve
absolute path in the host of a path inside a container.
This enables the host to access files inside a container, without
the need to access the container mount point.

#### pkg/containers: resolve host absolute container path
Add the functionality to load dynamic symbols of ELF files.
This will enable analysis of shared object files loaded or ELF file
that loads shared objects.

Fixes: #1655 

## Type of change

## How Has This Been Tested?

- [ ] pkg/containers/path_resolver_test.go
- [ ] pkg/utils/shared_objects/host_symbols_loader_test.go

## Final Checklist:

## Git Log Checklist:

My commits logs have:


yanivagman (commented):
I actually considered removing this buckets cache.
Keeping a container's entrypoint host pid should be enough for the entire lifetime of a container, and I don't see any advantage of having a bucket for all pids seen in a given mount namespace.
Instead, we can add the container's entrypoint pid to the cgroup info and simply extract it from there.

yanivagman (commented):
Overall looks good.
There are a few naming related comments. In general, I think function/variable names should be shorter to make the code cleaner. I left a few comments about this, but it may be relevant to other names used here as well.

pkg/utils/shared_objects/host_symbols_loader.go
```diff
@@ -0,0 +1,131 @@
...
+	syms, err := soLoader.loadSOSymbols(soInfo)
+	if err != nil {
+		return nil, err
+	}
+	return copyMap(syms.ExportedSymbols), nil
+}
+
+// GetImportedSymbols try to get shared objects imported symbols from lru, and if fails read needed information
+// from ELF file.
+func (soLoader *HostSharedObjectSymbolsLoader) GetImportedSymbols(soInfo SoExaminationInfo) (map[string]bool, error) {
```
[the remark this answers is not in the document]
NDStrahilevitz: To bring another perspective to this: I very much like immutability, it reduces surface for misuse of code and bugs later down the line. 
But we could be in a performance sensitive context here. If these functions will used in hot paths of the code, we should probably avoid the copy, but that should be benchmarked later on, when we know how this code will be used. Otherwise I think it's a premature optimization.

pkg/utils/shared_objects/host_symbols_loader.go
```diff
@@ -0,0 +1,131 @@
...
+	syms, err := soLoader.loadSOSymbols(soInfo)
+	if err != nil {
+		return nil, err
+	}
+	return copyMap(syms.ImportedSymbols), nil
```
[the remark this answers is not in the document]
NDStrahilevitz: see my thoughts above, we don't know if this is required performance.

pkg/utils/shared_objects/container_symbols_loader.go
```diff
@@ -0,0 +1,46 @@
+package shared_objects
+
+import (
+	"github.com/aquasecurity/tracee/pkg/containers"
+)
+
+// ContainersPathSOSymsLoaderDecorator is a decorator for SO loaders that resolves containers-relative paths to
+// absolute host paths.
+type ContainersPathSOSymsLoaderDecorator struct {
```
NDStrahilevitz, line 9: The fact that this is a decorator seems irrelevant in name to what looks to be the intended usage (can't tell since there's no consuming code). Intent is the most important, let's keep it simple, something like `ContainerSymbolsLoader` as opposed to the `HostSymbolLoader`.

If later down the line this decorator is needed elsewhere, it should be split out and then given the decorator name, but currently this is just another `SODynamicSymbolsLoader` implementation, let's name it as such.

pkg/utils/shared_objects/host_symbols_loader.go
```diff
@@ -0,0 +1,131 @@
...
+)
+
+// HostSharedObjectSymbolsLoader is responsible for efficient reading of shared object's symbols.
+// The logic of the loader here is used on absolute paths, so container relative paths won't work here.
+type HostSharedObjectSymbolsLoader struct {
+	loadingFunc func(path string) (*soDynamicSymbols, error)
+	soCache     soDynamicSymbolsCache
+}
+
+func InitSODynamicSymbolsLoader(cacheSize int) *HostSharedObjectSymbolsLoader {
```
NDStrahilevitz, line 15: I assume this may be used alongside the container symbols loader? So this should probably just be `HostSymbolsLoader` leave the interface name to the interface, IMO concrete implementations should present themselves as such.

pkg/containers/path_resolver.go
```diff
@@ -0,0 +1,43 @@
...
+	"fmt"
+	"github.com/aquasecurity/tracee/pkg/bucketscache"
+	"os"
+)
+
+// PathResolver generate an accessible absolute path from the root mount namespace to a a relative path
+// in a container.
+// Notice - to resolve host mount namespace, tracee reads from /proc/1/ns which requires CAP_SYS_PTRACE capability.
+type PathResolver struct {
+	statFunc         func(string) (os.FileInfo, error)
```
NDStrahilevitz, line 13: Existing interfaces could be used here see https://pkg.go.dev/io/fs

NDStrahilevitz (commented):
Looks good, but I agree with Yaniv, this needs some simpler names, added some suggestions of my own for that.

Also left a comment about an alternative way to write the path resolving (with existing golang interfaces for file systems, up to you if you prefer to use it)

pkg/utils/sharedobjs/host_symbols_loader.go
```diff
@@ -0,0 +1,132 @@
...
+
+// HostSymbolsLoader is responsible for efficient reading of shared object's symbols.
+// The logic of the loader here is used on absolute paths, so container relative paths won't work here.
+// This object operation requires the CAP_DAC_OVERRIDE to access files across the system.
+type HostSymbolsLoader struct {
+	loadingFunc func(path string) (*dynamicSymbols, error)
+	soCache     soDynamicSymbolsCache
+}
+
+func InitHostDynamicSymbolsLoader(cacheSize int) *HostSymbolsLoader {
```
yanivagman, line 16:
```suggestion
func InitHostSymbolsLoader(cacheSize int) *HostSymbolsLoader {
```

yanivagman (approved):
LGTM

rafaeldtinoco:
Reviewing this one. Will merge it myself if no objections/comments.

PR #1958, opened 2022-07-11
capabilities: mount cpuset in tmpfs when in a container

### LINKED ISSUE #1945: [BUG] Clean up the manual cpuset mount in cgroupv1 environments
## Prerequisites

- [x] There isn't an issue describing the bug.

Select one OR another:

- [ ] I'm going to create a PR to solve this (assign to yourself).

## Bug description

Screen Shot 2022-07-07 at 16 11 16

1. The cpuset diretory is mounted in the same directory as the tracee-ebpf binary.

2. The cpuset directory is never unmounted and/or cleaned up.

Screen Shot 2022-07-07 at 16 15 31

3. The syscall.Mount() call fails because of missing ./cpuset directory in some cases.

Screen Shot 2022-07-07 at 16 20 43

## Steps to reproduce

See pictures above.

```
$ make -f builder/Makefile.tracee-make alpine-prepare

$ make -f builder/Makefile.tracee-make alpine-shell
...

tracee@54ab0cbb4872[/tracee]$ make all
...

$ sudo ./dist/tracee-ebpf -o none
containers: failed to mount cgroupv1 controller: failed to mount no such file or directory
```

## Context

Relevant information about my setup:

* Linux version: Ubuntu Jammy
* Linux kernel version: 5.15.0-33-generic
* Tracee version (or commit id of your tree): v0.8.0-rc-1-24-g07c8af7
* LLVM version: 13
* Golang version: 1.17

grantseltzer:
This fixes the logic of creating the cgroup mount when in a container using cgroupv1. It adds an error check and creates the mount in tmpfs instead of the src directory. 

## Initial Checklist

## Description (git log)

commit 04af9297f788782231e58cc831148f4f275a30e1 (HEAD -> breakup-capabilities-check-and-drop, pfork/breakup-capabilities-check-and-drop)
Date:   Wed Jul 13 18:17:08 2022 -0400

    containers: move cpuset cgroup mount to tmpfs
    
commit d10100f36c7c0a64f762d7f5619d87436d5ecb45
Date:   Wed Jul 13 10:33:55 2022 -0400

    Add error check to cpuset mkdir call
    
Fixes  #1945 

## Type of change

## How Has This Been Tested?

#1945 contains reproduction instructions

### DIFF
--- a/pkg/containers/containers.go
+++ b/pkg/containers/containers.go
@@ -116,7 +116,12 @@ func (c *Containers) initCgroupV1() error {
 	}
 
 	if inContainer {
-		os.Mkdir("./cpuset", 0644)
+		err = os.Mkdir("./cpuset", 0644)
+		if err != nil && !os.IsExist(err) {
+			fmt.Fprintf(os.Stderr, "failed to create cpuset directory %s\n", err)
+			return nil
+		}
+
 		mountPath, err := filepath.Abs("./cpuset")
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "failed to get absolute path of cputset mountpoint %s\n", mountPath)
--- a/pkg/containers/containers.go
+++ b/pkg/containers/containers.go
@@ -116,13 +116,13 @@ func (c *Containers) initCgroupV1() error {
 	}
 
 	if inContainer {
-		err = os.Mkdir("./cpuset", 0644)
-		if err != nil && !os.IsExist(err) {
+		path, err := os.MkdirTemp("/tmp", "tracee-cpuset")
+		if err != nil {
 			fmt.Fprintf(os.Stderr, "failed to create cpuset directory %s\n", err)
 			return nil
 		}
 
-		mountPath, err := filepath.Abs("./cpuset")
+		mountPath, err := filepath.Abs(path)
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "failed to get absolute path of cputset mountpoint %s\n", mountPath)
 			return nil


yanivagman:
@grantseltzer do you have an environment where the issue is reproduced? I think that maybe we should find which capability is missing and add it as a temporary workaround instead of doing this split.

grantseltzer:
> @grantseltzer do you have an environment where the issue is reproduced? I think that maybe we should find which capability is missing and add it as a temporary workaround instead of doing this split.

Yes I do, i'll try doing this!

grantseltzer:
I've updated this, removing CAP_IPC_LOCK from the list of required capabilities fixes this on my test machine. @rafaeldtinoco can you confirm?

rafaeldtinoco:
I believe you're missing a commit in the PR ?

https://github.com/aquasecurity/tracee/commit/777202a30cf43bfb38c590ae74212a4f50c2a9b2

Could you update the PR ?

NDStrahilevitz (approved):
LGTM!

yanivagman:
> > @grantseltzer do you have an environment where the issue is reproduced? I think that maybe we should find which capability is missing and add it as a temporary workaround instead of doing this split.
> 
> Yes I do, i'll try doing this!

Anything with that? I think that if we can simply add the missing capability it will be better than doing this change as a workaround

grantseltzer:
I've tried finding a missing capability but i'm finding inconsistent
results in my testing environment so I haven't come to any definitive
conclusions yet. Cap sys admin should have everything, so what could be
missing? Is it possible to be unrelated?

Also do we know what unwanted behavior tracee experiences besides this
error message?

NDStrahilevitz:
> I've tried finding a missing capability but i'm finding inconsistent
results in my testing environment so I haven't come to any definitive
conclusions yet. Cap sys admin should have everything, so what could be
missing? Is it possible to be unrelated?
>
> Also do we know what unwanted behavior tracee experiences besides this
error message?

Just to note, when this error message occurs tracee behaves just like it did _before_ we've added this mounting solution. Even though it needs improvement, this error message should be treated more as warning rather than a bug IMO.

I also don't think this is related to CAP_SYS_ADMIN since it seems the issue is related to the directory not existing. So we need to check what the error in creating the directory is (my bad that i've forgot to add an error check to the `Mkdir` call).
Since I can't replicate this issue on my env (otherwise I would've caught it in development), can you maybe add an error check there to investigate, since you seem to be able to reproduce?

grantseltzer:
Ok I found the root of the issue. We need to map the volume `/sys/fs/cgroup` into the containers (where this error message occurs). We do not need to change the ordering of checking/dropping capabilities! I also added that error check for mkdir.

NDStrahilevitz:
> Ok I found the root of the issue. We need to map the volume `/sys/fs/cgroup` into the containers (where this error message occurs). We do not need to change the ordering of checking/dropping capabilities! I also added that error check for mkdir.

This is actually exactly what we tried to avoid with this solution. I think what you're getting is simply bypassing the mechanism we added, not actually solving the issue. 
@yanivagman if you're ok with just using this instead we should remove the code from #1915.

yanivagman:
> Ok I found the root of the issue. We need to map the volume `/sys/fs/cgroup` into the containers (where this error message occurs). We do not need to change the ordering of checking/dropping capabilities! I also added that error check for mkdir.

Wait, is it the root cause? I still don't understand why mkdir fails when dropping the capabilities.
I think we should avoid mounting cpuset from the host - it is only relevant for cgroup v1 and makes the command more cumbersome

grantseltzer:
> > Ok I found the root of the issue. We need to map the volume `/sys/fs/cgroup` into the containers (where this error message occurs). We do not need to change the ordering of checking/dropping capabilities! I also added that error check for mkdir.
> 
> Wait, is it the root cause? I still don't understand why mkdir fails when dropping the capabilities. I think we should avoid mounting cpuset from the host - it is only relevant for cgroup v1 and makes the command more cumbersome

No I suppose not, I didn't realize the usecase of the feature was to avoid the mount.

NDStrahilevitz:
Maybe it's because the directory already exists from previous creation in the "probe" execution of tracee-ebpf? We had similar trouble with that execution mounting before the actual tracee-ebpf so it would make sene.
I would think that it should then give an "already exists" error but maybe it's worth checking? Maybe adding an os.Stat call first.

grantseltzer:
Alright! Adding `CAP_DAC_OVERRIDE` does temporarily fix this issue. tracee-ebpf does not have permission (even with cap_sys_admin) to create the `cpuset` directory. If it already exists everything works fines. 

So @NDStrahilevitz as you suggested, I suspect if I add a check to see if the directory already exists, this can be resolved.

grantseltzer:
Yes this sounds correct to me. Working on the fix at the moment.

rafaeldtinoco:
> If so - we simply need to fix the permissions before the makedir or create the directory elsewhere where we have the right permissions. In addition, we need to add the error check for makedir (already done, I think, in this PR)

I believe we should mount the cgroup elsewhere. If Im using the "development container" (tracee-make) then I get the cgroup directory mounted with the source directory. It is not a problem if we're executing the container with an entrypoint, but without it it is bad IMO.

rafaeldtinoco:
image

Okay, it works now, turned out to be an easy fix after all, not bad.

Just to document: we do know there are leftovers (cgroup filesystem bind mounted in a temporary directory in case of cgroup v1) and we opted not to clean them (not for this release anyway).

Merging this.

rafaeldtinoco (approved):
LGTM

grantseltzer:
Ah I see.

Since the behavior falls back to previous versions in the case of this
error (if I understand your comment about this correctly) perhaps its best
to just think of it as a known bug, add the error check, and move on? I
don't think it has to do with a missing capability.

PR #1966, opened 2022-07-14
ebpf: simplify filters logic

yanivagman:
## Initial Checklist

## Description (git log)

Refactor the filtering logic and make it simpler.
Avoid unnecessary map lookups where possible.

Fixes: #issue_number

## Type of change

## How Has This Been Tested?

Manually, running all filters

- [ ] Test File A
- [ ] Test File B

- command 01
- command 02

## Final Checklist:

- [ x New and existing unit tests pass locally with my changes.

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -299,20 +299,8 @@ enum event_id_e
 #define FILTER_CGROUP_ID_ENABLED (1 << 21)
 #define FILTER_CGROUP_ID_OUT     (1 << 22)
 
-enum filter_options_e
-{
-    UID_LESS,
-    UID_GREATER,
-    PID_LESS,
-    PID_GREATER,
-    MNTNS_LESS,
-    MNTNS_GREATER,
-    PIDNS_LESS,
-    PIDNS_GREATER
-};
-
-#define LESS_NOT_SET    0
-#define GREATER_NOT_SET ULLONG_MAX
+#define FILTER_MAX_NOT_SET 0
+#define FILTER_MIN_NOT_SET ULLONG_MAX
 
 #define DEV_NULL_STR 0
 
@@ -552,6 +540,14 @@ typedef struct config_entry {
     u32 options;
     u32 filters;
     u32 cgroup_v1_hid;
+    u64 uid_max;
+    u64 uid_min;
+    u64 pid_max;
+    u64 pid_min;
+    u64 mnt_ns_max;
+    u64 mnt_ns_min;
+    u64 pid_ns_max;
+    u64 pid_ns_min;
     u8 events_to_submit[128]; // use 8*128 bits to describe up to 1024 events
 } config_entry_t;
 
@@ -707,7 +703,6 @@ BPF_HASH(kconfig_map, u32, u32, 10240);                 // kernel config variabl
 BPF_HASH(interpreter_map, u32, file_id_t, 10240);       // interpreter file used for each process
 BPF_HASH(containers_map, u32, u8, 10240);               // map cgroup id to container status {EXISTED, CREATED, STARTED}
 BPF_HASH(args_map, u64, args_t, 1024);                  // persist args between function entry and return
-BPF_HASH(inequality_filter, u32, u64, 256);             // filter events by some uint field either by < or >
 BPF_HASH(uid_filter, u32, u32, 256);                    // filter events by UID, for specific UIDs either by == or !=
 BPF_HASH(pid_filter, u32, u32, 256);                    // filter events by PID
 BPF_HASH(mnt_ns_filter, u64, u32, 256);                 // filter events by mount namespace id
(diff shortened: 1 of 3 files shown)


NDStrahilevitz:
If this isn't suddenly required for 0.8.0, could you open a PR to #1750 and include this there instead? I just want to avoid the likely conflicts down the line.

pkg/ebpf/tracee.go
```diff
@@ -667,18 +675,18 @@ func (t *Tracee) populateBPFMaps() error {
...
 	if err = bpfConfigMap.Update(unsafe.Pointer(&cZero), unsafe.Pointer(&configVal[0])); err != nil {
 		return err
 	}
 
 	errmap := make(map[string]error, 0)
-	errmap["uid_filter"] = t.config.Filter.UIDFilter.Set(t.bpfModule, "uid_filter", uidLess)
-	errmap["pid_filter"] = t.config.Filter.PIDFilter.Set(t.bpfModule, "pid_filter", pidLess)
-	errmap["mnt_ns_filter"] = t.config.Filter.MntNSFilter.Set(t.bpfModule, "mnt_ns_filter", mntNsLess)
-	errmap["pid_ns_filter"] = t.config.Filter.PidNSFilter.Set(t.bpfModule, "pid_ns_filter", pidNsLess)
+	errmap["uid_filter"] = t.config.Filter.UIDFilter.Set(t.bpfModule, "uid_filter")
```
rafaeldtinoco, line 686: Unrelated to your PR, but we should start the filterMap name once `t.config.Filter.UIDFilter` is created (so we don't have to give the 2nd argument whenever calling it).

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1454,30 +1449,19 @@ static __always_inline int init_event_data(event_data_t *data, void *ctx)
...
 
-    if (!filter_out)
-        return 0;
-
-    u64 *lessThan = bpf_map_lookup_elem(&inequality_filter, &less_idx);
-    if (lessThan == NULL)
-        return 1;
-
-    if ((*lessThan != LESS_NOT_SET) && (key >= *lessThan))
+    if ((lower_than != LOWER_NOT_SET) && (key >= lower_than))
```
NDStrahilevitz, line 1458: This would break backwards compatibility, but just want to note, we technically set this behavior for the `<` operation, not `<=`. Maybe it's worth changing to be in line (note that in #1750 I allow for both operations).

pkg/ebpf/filters.go
```diff
@@ -116,7 +105,7 @@ func (filter *UintFilter) Parse(operatorAndValues string) error {
 	return nil
 }
 
-func (filter *UintFilter) Set(bpfModule *bpf.Module, filterMapName string, lessIdx uint32) error {
+func (filter *UintFilter) Set(bpfModule *bpf.Module, filterMapName string) error {
```
NDStrahilevitz, line 108: Awesome that we got rid of this!

pkg/ebpf/tracee.go
```diff
@@ -652,11 +652,19 @@ func (t *Tracee) populateBPFMaps() error {
 	}
 
 	cZero := uint32(0)
-	configVal := make([]byte, 144)
+	configVal := make([]byte, 208)
 	binary.LittleEndian.PutUint32(configVal[0:4], uint32(os.Getpid()))
 	binary.LittleEndian.PutUint32(configVal[4:8], t.getOptionsConfig())
 	binary.LittleEndian.PutUint32(configVal[8:12], t.getFiltersConfig())
 	binary.LittleEndian.PutUint32(configVal[12:16], uint32(t.containers.GetCgroupV1HID()))
+	binary.LittleEndian.PutUint64(configVal[16:24], t.config.Filter.UIDFilter.Less)
```
NDStrahilevitz, line 660: Maybe not for this PR but when I integreate these changes in #1750, but we should probably encapsulate this in the `Set` method.
Probably pass the buffers for the max and min values to the BPF Filter struct. 

As preparation, while I do like the cleaner `Set` maybe we should pass the config or buffer pointers to the Set method for now so it won't get lost, WDYT?

NDStrahilevitz (changes requested):
I like this change!
So both from reading and some testing, doesn't seem to break anything so we're good on that.

I do have some comments I would like to resolve before approving but it will probably amount at most some variable name changes, and mainly to see how to fit this with #1750 after we merge this.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1454,30 +1449,19 @@ static __always_inline int init_event_data(event_data_t *data, void *ctx)
 // INTERNAL: FILTERING -----------------------------------------------------------------------------
 
 static __always_inline int
-uint_filter_matches(bool filter_out, void *filter_map, u64 key, u32 less_idx, u32 greater_idx)
+uint_filter_matches(bool filter_out, void *filter_map, u64 key, u64 lower_than, u64 greater_than)
```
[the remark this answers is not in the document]
yanivagman: Which names do you suggest then?

pkg/ebpf/tracee.go
yanivagman: The Set method will possibly be removed in a future PR I'm working on for multiple scopes. The reason for that is the need to set values that are a combination of multiple filters. For this reason, ai think we should wait with such changes for now

NDStrahilevitz: Interesting. So we would have a function similar to `Set` that would take multiple Filters as parameters? Either way, we will have it until then, but I'll defer handling this to my PR.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1454,30 +1449,19 @@ static __always_inline int init_event_data(event_data_t *data, void *ctx)
...
-    if ((*lessThan != LESS_NOT_SET) && (key >= *lessThan))
+    if ((lower_than != LOWER_NOT_SET) && (key >= lower_than))
         return 0;
 
-    u64 *greaterThan = bpf_map_lookup_elem(&inequality_filter, &greater_idx);
-    if (greaterThan == NULL)
-        return 1;
-
-    if ((*greaterThan != GREATER_NOT_SET) && (key <= *greaterThan))
+    if ((greater_than != GREATER_NOT_SET) && (key <= greater_than))
```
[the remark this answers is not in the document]
yanivagman: ditto

NDStrahilevitz (approved):
LGTM!

Issue #1816, opened 2022-06-09
[RFE] Translate File Descriptors to File Names


mthbrown:
## Prerequisites

- [ X] There isn't an issue describing the feature I need.
- [ X] I don't think opening a discussion thread first is relevant.
- [ X] I have a use case for the feature I would like to request.

## Feature description

This feature is about being able to easily determine what file some syscalls are related to. Currently, if I have tracee filter by something like the `write` syscall, I might see something like this:

```json
{
  "timestamp": 1654734049345486000,
  "processorId": 0,
  "processId": 7,
  "threadId": 7,
  "parentProcessId": 1,
  "hostProcessId": 233076,
  "hostThreadId": 233076,
  "hostParentProcessId": 233036,
  "userId": 0,
  "mountNamespace": 4026532349,
  "pidNamespace": 4026532352,
  "processName": "uname",
  "hostName": "55ab21515f70",
  "containerId": "55ab21515f707ca9a071f78b6efbac13d4d1ac9072fbf981a80989035ef943ee",
  "eventId": "1",
  "eventName": "write",
  "argsNum": 3,
  "returnValue": 6,
  "stackAddresses": null,
  "args": [
    {
      "name": "fd",
      "type": "int",
      "value": 1
    },
    {
      "name": "buf",
      "type": "void*",
      "value": 140337780654080
    },
    {
      "name": "count",
      "type": "size_t",
      "value": 6
    }
  ]
}
```

I have no idea what file this refers to and just have a file descriptor to work with. I'd have to try to go through previous syscalls to do the mapping. It would be great if there was an option to show the actual filename + path as well, similar to strace's `-y` option

The feature is useful because currently working with the output isn't easy

## Context

My current environment, where I would like the feature to work, is:

* Linux version: Ubuntu 20.04
* Linux kernel version: Linux ubuntu2004.localdomain 5.4.0-105-generic
* Tracee version (or commit id of your tree): Tracee version "v0.7.0"
* LLVM version:
* Golang version:

geyslan:
https://github.com/aquasecurity/tracee/blob/260570df25844a7e05a1440d448082cec2447823/pkg/ebpf/events_pipeline.go#L161

[// Argument holds the information for one argument
type Argument struct {
	ArgMeta
	Value interface{} `json:"value"`
}](https://github.com/aquasecurity/tracee/blob/260570df25844a7e05a1440d448082cec2447823/types/trace/trace.go#L79-L83)

[...]

PR #1971, opened 2022-07-14
[FEAT] Translate File Descriptors to File Names

geyslan:
## Initial Checklist

## Description (git log)

- ebpf: standardize and reorder syscall names
- ebpf: initial approach to translate fd
- tracee: events: docs: fd arg enrichment

Fixes: #1816 

## Type of change

## How Has This Been Tested?

- [ ] Test File A
- [ ] Test File B

- `sudo ./dist/tracee-ebpf --trace event=write,read --trace pid=new --trace comm=cat -o option:parse-arguments-fds`
- `cat /etc/passwd > out`
```bash
23:56:34:092239  1000   cat              335031  335031  832              read                 fd: 3=/usr/lib/x86_64-linux-gnu/libc.so.6, buf: 0x7fff2e6aeb08, count: 832
23:56:34:092532  1000   cat              335031  335031  2996             read                 fd: 3=/etc/locale.alias, buf: 0x557910d914a0, count: 4096
23:56:34:092550  1000   cat              335031  335031  0                read                 fd: 3=/etc/locale.alias, buf: 0x557910d914a0, count: 4096
23:56:34:092878  1000   cat              335031  335031  1867             read                 fd: 3=/etc/passwd, buf: 0x7ff0efbd4000, count: 131072
23:56:34:092888  1000   cat              335031  335031  1867             write                fd: 1=/vagrant/out, buf: 0x7ff0efbd4000, count: 1867
23:56:34:092947  1000   cat              335031  335031  0                read                 fd: 3=/etc/passwd, buf: 0x7ff0efbd4000, count: 131072
```

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -146,47 +146,220 @@ enum argument_type_e
 #define UNDEFINED_SYSCALL 1000
 
 #if defined(bpf_target_x86)
-    #define SYS_MMAP         9
-    #define SYS_MPROTECT     10
-    #define SYS_RT_SIGRETURN 15
-    #define SYS_EXECVE       59
-    #define SYS_EXIT         60
-    #define SYS_EXIT_GROUP   231
-    #define SYS_EXECVEAT     322
-    #define SYSCALL_CONNECT  42
-    #define SYSCALL_ACCEPT   43
-    #define SYSCALL_ACCEPT4  288
-    #define SYSCALL_LISTEN   50
-    #define SYSCALL_BIND     49
-    #define SYSCALL_SOCKET   41
-    #define SYSC_SETSOCKOPT  54
-    #define SYS_DUP          32
-    #define SYS_DUP2         33
-    #define SYS_DUP3         292
-    #define SYS_OPEN         2
-    #define SYS_OPENAT       257
-    #define SYS_OPENAT2      437
+    #define SYSCALL_READ                   0
+    #define SYSCALL_WRITE                  1
+    #define SYSCALL_OPEN                   2
+    #define SYSCALL_CLOSE                  3
+    #define SYSCALL_FSTAT                  5
+    #define SYSCALL_LSEEK                  8
+    #define SYSCALL_MMAP                   9
+    #define SYSCALL_MPROTECT               10
+    #define SYSCALL_RT_SIGRETURN           15
+    #define SYSCALL_IOCTL                  16
+    #define SYSCALL_PREAD64                17
+    #define SYSCALL_PWRITE64               18
+    #define SYSCALL_READV                  19
+    #define SYSCALL_WRITEV                 20
+    #define SYSCALL_DUP                    32
+    #define SYSCALL_DUP2                   33
+    #define SYSCALL_SOCKET                 41
+    #define SYSCALL_CONNECT                42
+    #define SYSCALL_ACCEPT                 43
+    #define SYSCALL_SENDTO                 44
+    #define SYSCALL_RECVFROM               45
+    #define SYSCALL_SENDMSG                46
+    #define SYSCALL_RECVMSG                47
+    #define SYSCALL_SHUTDOWN               48
+    #define SYSCALL_BIND                   49
+    #define SYSCALL_LISTEN                 50
+    #define SYSCALL_GETSOCKNAME            51
+    #define SYSCALL_GETPEERNAME            52
+    #define SYSCALL_SETSOCKOPT             54
+    #define SYSCALL_GETSOCKOPT             55
+    #define SYSCALL_EXECVE                 59
+    #define SYSCALL_EXIT                   60
+    #define SYSCALL_FCNTL                  72
+    #define SYSCALL_FLOCK                  73
+    #define SYSCALL_FSYNC                  74
+    #define SYSCALL_FDATASYNC              75
+    #define SYSCALL_FTRUNCATE              77
+    #define SYSCALL_GETDENTS               78
+    #define SYSCALL_FCHDIR                 81
+    #define SYSCALL_FCHMOD                 91
+    #define SYSCALL_FCHOWN                 93
+    #define SYSCALL_FSTATFS                138
+    #define SYSCALL_READAHEAD              187
+    #define SYSCALL_FSETXATTR              190
+    #define SYSCALL_FGETXATTR              193
+    #define SYSCALL_FLISTXATTR             196
+    #define SYSCALL_FREMOVEXATTR           199
+    #define SYSCALL_GETDENTS64             217
+    #define SYSCALL_FADVISE64              221
+    #define SYSCALL_EXIT_GROUP             231
+    #define SYSCALL_EPOLL_WAIT             232
+    #define SYSCALL_EPOLL_CTL              233
+    #define SYSCALL_INOTIFY_ADD_WATCH      254
+    #define SYSCALL_INOTIFY_RM_WATCH       255
+    #define SYSCALL_OPENAT                 257
+    #define SYSCALL_MKDIRAT                258
+    #define SYSCALL_MKNODAT                259
+    #define SYSCALL_FCHOWNAT               260
+    #define SYSCALL_FUTIMESAT              261
+    #define SYSCALL_NEWFSTATAT             262
+    #define SYSCALL_UNLINKAT               263
+    #define SYSCALL_SYMLINKAT              266
+    #define SYSCALL_READLINKAT             267
+    #define SYSCALL_FCHMODAT               268
+    #define SYSCALL_FACCESSAT              269
+    #define SYSCALL_SYNC_FILE_RANGE        277
+    #define SYSCALL_VMSPLICE               278
+    #define SYSCALL_UTIMENSAT              280
+    #define SYSCALL_EPOLL_PWAIT            281
+    #define SYSCALL_SIGNALFD               282
+    #define SYSCALL_FALLOCATE              285
+    #define SYSCALL_TIMERFD_SETTIME        286
+    #define SYSCALL_TIMERFD_GETTIME        287
+    #define SYSCALL_ACCEPT4                288
+    #define SYSCALL_SIGNALFD4              289
+    #define SYSCALL_DUP3                   292
+    #define SYSCALL_PREADV                 295
+    #define SYSCALL_PWRITEV                296
+    #define SYSCALL_PERF_EVENT_OPEN        298
+    #define SYSCALL_RECVMMSG               299
+    #define SYSCALL_NAME_TO_HANDLE_AT      303
+    #define SYSCALL_OPEN_BY_HANDLE_AT      304
+    #define SYSCALL_SYNCFS                 306
+    #define SYSCALL_SENDMMSG               307
+    #define SYSCALL_SETNS                  308
+    #define SYSCALL_FINIT_MODULE           313
+    #define SYSCALL_EXECVEAT               322
+    #define SYSCALL_PREADV2                327
+    #define SYSCALL_PWRITEV2               328
+    #define SYSCALL_STATX                  332
+    #define SYSCALL_PIDFD_SEND_SIGNAL      424
+    #define SYSCALL_IO_URING_ENTER         426
+    #define SYSCALL_IO_URING_REGISTER      427
+    #define SYSCALL_OPEN_TREE              428
+    #define SYSCALL_FSCONFIG               431
+    #define SYSCALL_FSMOUNT                432
+    #define SYSCALL_FSPICK                 433
+    #define SYSCALL_OPENAT2                437
+    #define SYSCALL_FACCESSAT2             439
+    #define SYSCALL_PROCESS_MADVISE        440
+    #define SYSCALL_EPOLL_PWAIT2           441
+    #define SYSCALL_MOUNT_SETATTR          442
+    #define SYSCALL_QUOTACTL_FD            443
+    #define SYSCALL_LANDLOCK_ADD_RULE      445
+    #define SYSCALL_LANDLOCK_RESTRICT_SELF 446
+    #define SYSCALL_PROCESS_MRELEASE       448
+
 #elif defined(bpf_target_arm64)
-    #define SYS_MMAP         222
-    #define SYS_MPROTECT     226
-    #define SYS_RT_SIGRETURN 139
-    #define SYS_EXECVE       221
-    #define SYS_EXIT         93
-    #define SYS_EXIT_GROUP   94
-    #define SYS_EXECVEAT     281
-    #define SYSCALL_CONNECT  203
-    #define SYSCALL_ACCEPT   202
-    #define SYSCALL_ACCEPT4  242
-    #define SYSCALL_LISTEN   201
-    #define SYSCALL_BIND     200
-    #define SYSCALL_SOCKET   198
-    #define SYSC_SETSOCKOPT  208
-    #define SYS_DUP          23
-    #define SYS_DUP2         UNDEFINED_SYSCALL
-    #define SYS_DUP3         24
-    #define SYS_OPEN         UNDEFINED_SYSCALL
-    #define SYS_OPENAT       56
-    #define SYS_OPENAT2      437
+    #define SYSCALL_READ                   63
+    #define SYSCALL_WRITE                  64
+    #define SYSCALL_OPEN                   UNDEFINED_SYSCALL
+    #define SYSCALL_CLOSE                  57
+    #define SYSCALL_FSTAT                  80
+    #define SYSCALL_LSEEK                  62
+    #define SYSCALL_MMAP                   222
+    #define SYSCALL_MPROTECT               226
+    #define SYSCALL_RT_SIGRETURN           139
+    #define SYSCALL_IOCTL                  29
+    #define SYSCALL_PREAD64                67
+    #define SYSCALL_PWRITE64               68
+    #define SYSCALL_READV                  65
+    #define SYSCALL_WRITEV                 66
+    #define SYSCALL_DUP                    23
+    #define SYSCALL_DUP2                   UNDEFINED_SYSCALL
+    #define SYSCALL_SOCKET                 198
+    #define SYSCALL_CONNECT                203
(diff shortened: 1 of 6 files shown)


geyslan:
I bring you three issues for discussion.

**1. Path length limitation**

Currently it's working in vagrant env with `MAX_CACHED_PATH_SIZE` (64). Do you suggest another length for this buffer?

**2. Syscalls with more than one file descriptor arg**

Both sides (user/kernel) logic are dealing only with syscalls that have a unique arg as file descriptor. Changes are needed to both sides for improving it to treat syscalls like:

```
-------------------------------------------
syscall            fd arg type   nth args
-------------------------------------------
pipe               int[2]
pipe2              int[2]

epoll_ctl             int          1,3
renameat              int          1,3
renameat2             int          1,3
linkat                int          1,3
move_mount            int          1,3

dup2                  int          1,2
dup3                  int          1,2
sendfile              int          1,2
kexec_file_load       int          1,2
pidfd_getfd           int          1,2
socket_dup            int          1,2

fanotify_mark         int          1,4
```

**3. Timing for captured path map entry deletion.**

Knowing the safe time to delete the path entry from map is not so crystal. Perhaps to use a lock control? Certainly it wouldn't be optimal. [It's not deleting so far](https://github.com/aquasecurity/tracee/blob/b8b4860da84f4c8719bc1e2c26c09855b3969c49/pkg/events/parse_args.go#L58-L60).

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -146,47 +146,220 @@ enum argument_type_e
...
 
 #if defined(bpf_target_x86)
-    #define SYS_MMAP         9
-    #define SYS_MPROTECT     10
-    #define SYS_RT_SIGRETURN 15
-    #define SYS_EXECVE       59
-    #define SYS_EXIT         60
-    #define SYS_EXIT_GROUP   231
-    #define SYS_EXECVEAT     322
-    #define SYSCALL_CONNECT  42
```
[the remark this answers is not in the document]
rafaeldtinoco: Maybe we should normalize names by using something not used in kernel headers ?

pkg/events/parse_args.go
```diff
@@ -27,6 +39,28 @@ func ParseArgs(event *trace.Event) error {
 		}
 	}
 
+	if fdArg := GetArg(event, "fd"); fdArg != nil {
```
rafaeldtinoco, line 42: A map lookup to each call to ParseArgs will slow down the pipeline for all cases when we have "option:parse-arguments". That is definitely not what we want. Maybe we should have another option "option:parse-arguments-fds" instead, that would only do the fds translations if explicitly requested (with a warning that translating file descriptors might cause slow dows in the pipeline).

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2539,6 +2846,28 @@ int tracepoint__raw_syscalls__sys_enter(struct bpf_raw_tracepoint_args *ctx)
         sys->id = *id_64;
     }
 
+    if (should_submit(sys->id, data.config) &&
+        has_syscall_fd_arg(sys->id) &&
+        data.config->options & OPT_CAPTURE_FD_FILEPATH) {
```
rafaeldtinoco, line 2851: I believe OPT_CAPTURE_FD_FILEPATH should be checked first, then should_submit, then has_syscall_fd_arg.

rafaeldtinoco (commented):
Thanks for your contribution Geyslan. I left some comments on your PR. Let me know what you think!

rafaeldtinoco:
> I bring you three issues for discussion.
> 
> **1. Path length limitation**
> 
> Currently it's working in vagrant env with `MAX_CACHED_PATH_SIZE` (64). Do you suggest another length for this buffer?

I believe it is ok for now.


> **2. Syscalls with more than one file descriptor arg**
> 
> Both sides (user/kernel) logic are dealing only with syscalls that have a unique arg as file descriptor. Changes are needed to both sides for improving it to treat syscalls like:
> 
> ```
> -------------------------------------------
> syscall            fd arg type   nth args
> -------------------------------------------
> pipe               int[2]
> pipe2              int[2]
> 
> epoll_ctl             int          1,3
> renameat              int          1,3
> renameat2             int          1,3
> linkat                int          1,3
> move_mount            int          1,3
> 
> dup2                  int          1,2
> dup3                  int          1,2
> sendfile              int          1,2
> kexec_file_load       int          1,2
> pidfd_getfd           int          1,2
> socket_dup            int          1,2
> 
> fanotify_mark         int          1,4
> ```

Let's focus in the simplest cases for now, as you did (good enough for the exercise).

> **3. Timing for captured path map entry deletion.**
> 
> Knowing the safe time to delete the path entry from map is not so crystal. Perhaps to use a lock control? Certainly it wouldn't be optimal. [It's not deleting so far](https://github.com/aquasecurity/tracee/blob/b8b4860da84f4c8719bc1e2c26c09855b3969c49/pkg/events/parse_args.go#L58-L60).

Maybe you can cleanup all entries when the task exits only ? This will improve performance of parseArgs call as well (check my comment about specific option for this feature).

I would clear the FD entry whenever "close()" is called OR clear ALL the task's entries when the task exits, WDYT ?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -735,6 +919,7 @@ BPF_PROG_ARRAY(sys_enter_tails, MAX_EVENT_ID);          // store programs for ta
 BPF_PROG_ARRAY(sys_exit_tails, MAX_EVENT_ID);           // store programs for tail calls
 BPF_STACK_TRACE(stack_addresses, MAX_STACK_ADDRESSES);  // store stack traces
 BPF_HASH(module_init_map, u32, kmod_data_t, 256);       // holds module information between
+BPF_HASH(fd_arg_path_map, fd_arg_task_t, fd_arg_path_t, 1024); // store fds paths by task
```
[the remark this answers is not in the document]
geyslan: I was trying to not touch other lines. But, ok, I'll realign the others to same column. :+1:

pkg/events/parse_args.go
geyslan: I agree. For sure `option:parse-arguments-fds` can be created.

pkg/ebpf/c/tracee.bpf.c
geyslan: Limiting the checks only for `option:parse-arguments-fds`, yeah, we can avoid a lot of processing reading OPT_CAPTURE_FD_FILEPATH first.

geyslan:
> I believe it is ok for now.
> ...
> Let's focus in the simplest cases for now, as you did (good enough for the exercise).

:+1: 

> Maybe you can cleanup all entries when the task exits only ? This will improve performance of parseArgs call as well (check my comment about specific option for this feature).

> I would clear the FD entry whenever "close()" is called OR clear ALL the task's entries when the task exits, WDYT ?

It could be a way to do that. But the map can get full as soon as a long living task calls a bunch of syscalls with different fd arguments. And other misleading references could happen by unordered flow of data (submission vs consumption) as I mentioned in https://github.com/aquasecurity/tracee/issues/1816?ts=8#issuecomment-1183227274.

The current design is:

```c
typedef struct fd_arg_task {
    u32 pid;
    u32 tid;
    int fd;
} fd_arg_task_t;
```

I was pondering to change it to this:

```c
typedef struct fd_arg_event {
    unsigned long ts; // Timestamp of syscall entry
    u16 processor_id; // The ID of the processor which processed the event
    int fd;
} fd_arg_event_t;
```
With that, we could safely assume that the filepath is exclusively referenced to a specific event, being able to delete it without reservations after translating it. WDYT?

```go
	if fdArg := GetArg(event, "fd"); fdArg != nil {
		if fd, isInt32 := fdArg.Value.(int32); isInt32 {
			fdArgTask := &FDArgTask{
				Timestamp: uint32(event.Timestamp),
				Processor: uint16(event.ProcessorID),
				FD:        fd,
			}
			bs, err := fdArgPathMap.GetValue(unsafe.Pointer(fdArgTask))
			if err != nil {
				return err
			}

			fdArgPath := &FDArgPath{}
			copy(fdArgPath.Path[:], bs[:len(fdArgPath.Path)])

			fdArg.Value = fmt.Sprintf("%d<%s>", fd, fdArgPath.Path)
			if err := fdArgPathMap.DeleteKey(unsafe.Pointer(fdArgTask)); err != nil {
				return err
			}
		}
	}
```

rafaeldtinoco:
> > I believe it is ok for now.
> > ...
> > Let's focus in the simplest cases for now, as you did (good enough for the exercise).
> 
> 👍
> 
> > Maybe you can cleanup all entries when the task exits only ? This will improve performance of parseArgs call as well (check my comment about specific option for this feature).
> 
> > I would clear the FD entry whenever "close()" is called OR clear ALL the task's entries when the task exits, WDYT ?
> 
> It could be a way to do that. But the map can get full as soon as a long living task calls a bunch of syscalls with different fd arguments. And other misleading references could happen by unordered flow of data (submission vs consumption) as I mentioned in #1816.

```
[root@fujitsu:/proc]# for dir in $(find . -type d -iregex '.*task/[0-9]*/fd'); do find $dir/ | grep "fd/[0-9]" | grep -v "fd/[0|1|2]$" | wc -l; done | xargs | sed 's: :+:g' | bc
71338
```

If my test is right, 71k file descriptors in my server, 64 bytes per fd would give us ~4MB, not sure that is "too much" for BPF map. Maybe a LRU_HASH would be enough in a regular workload, guaranteeing file descriptors being used to have its path populated.

I guess it would be ok not to enrich the FD if we couldn't, as the parse-args option isn't used for the signatures/detections (the event pathname argument is).

> 
> The current design is:
> 
> ```c
> typedef struct fd_arg_task {
>     u32 pid;
>     u32 tid;
>     int fd;
> } fd_arg_task_t;
> ```
> 
> I was pondering to change it to this:
> 
> ```c
> typedef struct fd_arg_event {
>     unsigned long ts; // Timestamp of syscall entry
>     u16 processor_id; // The ID of the processor which processed the event
>     int fd;
> } fd_arg_event_t;
> ```
> 
> With that, we could safely assume that the filepath is exclusively referenced to a specific event, being able to delete it without reservations after translating it. WDYT?
> 
> ```go
> 	if fdArg := GetArg(event, "fd"); fdArg != nil {
> 		if fd, isInt32 := fdArg.Value.(int32); isInt32 {
> 			fdArgTask := &FDArgTask{
> 				Timestamp: uint32(event.Timestamp),
> 				Processor: uint16(event.ProcessorID),
> 				FD:        fd,
> 			}
> 			bs, err := fdArgPathMap.GetValue(unsafe.Pointer(fdArgTask))
> 			if err != nil {
> 				return err
> 			}
> 
> 			fdArgPath := &FDArgPath{}
> 			copy(fdArgPath.Path[:], bs[:len(fdArgPath.Path)])
> 
> 			fdArg.Value = fmt.Sprintf("%d<%s>", fd, fdArgPath.Path)
> 			if err := fdArgPathMap.DeleteKey(unsafe.Pointer(fdArgTask)); err != nil {
> 				return err
> 			}
> 		}
> 	}
> ```

Adding to map and deleting with correspondent enrichment seems too much. Let's go with the LRU map and see how that goes (and if we need to improve the logic to something more complex later).

rafaeldtinoco:
Also:

image

Make sure to "fix format" your code for the next push (there is a Makefile target for it). This way the tests will pass the code analysis phase and we can check the functional tests results.

Thanks again!

geyslan:
@rafaeldtinoco 

> If my test is right, 71k file descriptors in my server, 64 bytes per fd would give us ~4MB, not sure that is "too much" for BPF map. Maybe a LRU_HASH would be enough in a regular workload, guaranteeing file descriptors being used to have its path populated.

In my workstation (not a server), diregarding [0-2] defaults, I got ~63k fds at normal conditions.

```bash
[hb proc]# for fds in '[0-9]*/task/[0-9]*/fd/*'; do echo $fds; done | tr ' ' '\n' | grep -c -v "fd/[0|1|2]$"
63851
```

> I guess it would be ok not to enrich the FD if we couldn't, as the parse-args option isn't used for the signatures/detections (the event pathname argument is).

All right. So if `option:parse-arguments-fds` is set and the file path of (e.g.) 4 FD is no longer found in the map, we can enrich it as `"4<missing>"`.

> Adding to map and deleting with correspondent enrichment seems too much. Let's go with the LRU map and see how that goes (and if we need to improve the logic to something more complex later).

Got it. So let's keep with current map design, changing it to LRU type and see how things behave. Thanks!

geyslan:
> All right. So if option:parse-arguments-fds is set and the file path of (e.g.) 4 FD is no longer found in the map, we can enrich it as "4<missing>".

I decided not to follow that. Currently if the option is set and a map entry is not found, it's erring. Surely it can be changed to out differently.

pkg/ebpf/tracee.go
```diff
@@ -454,6 +464,7 @@ const (
 	optCaptureModules
 	optCgroupV1
 	optProcessInfo
+	optCaptureFDFilePath
```
rafaeldtinoco, line 467: We should change "Capture" wording here. We use "capture" to "capture artifacts" and it might be a bit confusing. Anything else, no hard opinion.

rafaeldtinoco (commented):
Nice work!

I believe we should only parse FD args when parse-args is also set (or assume both together). Something like:

```C
$ git diff -p
diff --git a/cmd/tracee-ebpf/internal/flags/output.go b/cmd/tracee-ebpf/internal/flags/output.go
index 25716ed0..9ce26120 100644
--- a/cmd/tracee-ebpf/internal/flags/output.go
+++ b/cmd/tracee-ebpf/internal/flags/output.go
@@ -88,6 +88,7 @@ func PrepareOutput(outputSlice []string) (tracee.OutputConfig, printer.Config, e
                                outcfg.ParseArguments = true
                        case "parse-arguments-fds":
                                outcfg.ParseArgumentsFDs = true
+                               outcfg.ParseArguments = true // no point in parsing file descriptor args only
                        case "sort-events":
                                outcfg.EventsSorting = true
                        default:
diff --git a/pkg/ebpf/events_pipeline.go b/pkg/ebpf/events_pipeline.go
index 0cd30030..06c75ae6 100644
--- a/pkg/ebpf/events_pipeline.go
+++ b/pkg/ebpf/events_pipeline.go
@@ -272,12 +272,12 @@ func (t *Tracee) sinkEvents(ctx gocontext.Context, in <-chan *trace.Event) <-cha
                                                t.handleError(err)
                                                continue
                                        }
-                               }
-                               if t.config.Output.ParseArgumentsFDs {
-                                       err := events.ParseArgsFDs(event, t.FDArgPathMap)
-                                       if err != nil {
-                                               t.handleError(err)
-                                               continue
+                                       if t.config.Output.ParseArgumentsFDs {
+                                               err := events.ParseArgsFDs(event, t.FDArgPathMap)
+                                               if err != nil {
+                                                       t.handleError(err)
+                                                       continue
+                                               }
                                        }
                                }
                                select {
```

Some small comments, but overall looks good. I'll merge once those are addressed! Thanks again!

rafaeldtinoco:
BTW @geyslan, I took so much time to review this that it has a conflict. Would you mind rebasing it to origin/main again, and address the minor comments ? I'll merge once you force push the changes (highlight me please). Thanks!

pkg/ebpf/tracee.go
geyslan: Perhaps `optTranslateFDFilePath`.

rafaeldtinoco:
Adding @yanivagman for a review of this feature (might be controversial because of enriching delays and all, but I think because it is optional and disabled by default might be good enough). If Yaniv feels like having that feature, I also added the correct milestone (since the release 0.8.1 is this week and we won't be adding this one before that).

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2205,6 +2391,128 @@ static __always_inline int save_args_to_submit_buf(event_data_t *data, u64 types
     return arg_num;
 }
 
+// INTERNAL: SYSCALLS ------------------------------------------------------------------------------
+
+static __always_inline bool has_syscall_fd_arg(uint syscall_id)
+{
+    // Only syscalls with one fd argument so far
+    switch (syscall_id) {
```
rafaeldtinoco, line 2399: Only way to "fix this long list" I can think of (because values are not sequential) would be to transform the syscall defines into a feature table (syscall, archnumber, features bit flags), but we can't have global variables so... its tricky to improve this :\

geyslan: I think we can't run away entirely from `#define`s. A draf suggestion would be to create a `get_syscall()`, to be called in sys_enter, which returns a bit flagged syscall structure based on arch and syscall_id?

```c
static __always_inline struct syscall get_syscall(uint syscall_id, uint arch)
{
	struct syscall s = {
		.syscall_id = syscall_id,
		.arch = arch,
	};

	switch (s.syscall_id) {
		case SYSCALL_READ:
			s.flags = FD_ARG1
			break;
		...
		case SYSCALL_DUP2:
		case SYSCALL_DUP3:
			s.flags = FD_ARG_1 | FD_ARG_2;
			break;
	}

	return s;
}
```

It would be a initial step for checking flags from specific use cases.

rafaeldtinoco (approved):
Alright, this looks okay to me for now considering:

1. it won´t waste time in sys_enter() as its a simple check for config option (but I do think we could replace that code in sys_enter() with a tail call for all syscalls that would need the fd name translation).
 
2. pipeline won´t be delayed as this is an option apart from regular parse-args (so it won't run unless explicitly asked) but, yet, a nice introspection parsing option (to understand which fd belongs to what).

I think we should think about the tail call option, and to enrich the file descriptor parsing with other types of file descriptors (sockets ?). I would check what strace provides:

image

socket file descriptors are as important as regular file path descriptors when we want to understand what file is "described" by it.

rafaeldtinoco:
I'm going to merge this from another PR, solving the conflicts, adding some TODOs and opening an issue to address what I think should be addressed.

rafaeldtinoco:
Closing this to merge https://github.com/aquasecurity/tracee/pull/2086 instead (solving conflicts).

Issue #1967, opened 2022-07-14
[BUG] tracee-ebpf and tracee-rules mismatch of interpreting which events belong to container/host


OriGlassman:
## Prerequisites

- [x] This affects latest released version.
- [x] This affects current development tree (origin/HEAD).
- [x] There isn't an issue describing the bug.

Select one OR another:

- [ ] I'm going to create a PR to solve this (assign to yourself).
- [x] Someone else should solve this.

## Bug description

tracee-ebpf may regard events that contain containerId to the **host** (upon creation of a new container), while tracee-rules will regard them to a **container** (since the event has a containerId). 

## Context

Relevant information about my setup:

* Linux version:
* Linux kernel version:
* Tracee version (or commit id of your tree):
* LLVM version:
* Golang version:

NDStrahilevitz:
Can you please elaborate with an example? Are you talking about `container_create` events or any event with a containerId?

rafaeldtinoco:
Yes @OriGlassman, please try to always provide us a reproducer and the output of the effects you're observing, so we can reproduce and fix ourselves. Thank you!

OriGlassman:
When container is created, at some point tracee-ebpf will populate the containerId field but still treat these events as host events (with container filter on, you won't see those events) as tracee-ebpf waits for the first execve insde the container and then attributes all the events afterwards to the container.
tracee-rules attribution to host or container is based on containerId field. If the field is populated, it attributes the event to container.
So what do we have? we have a few events with containerId populated, where tracee-ebpf relates them to the host, while tracee-rules relates those same events to a container. Mismatch.

AlonZivony:
The bugfix implementation:
* Modify the `padding` member of the `context_t` to be a flags bitfield, with flag for `container_started`
* When doing `sched_process_exec` change the state of the container to started.
* Create in `trace.Event` a member for flags received from the kernel, and pass it on as bools.
* Change check in tracee-rules to check the new flag instead the `ContainerID`

NDStrahilevitz:
Could you provide an effective example of a problem this bug creates? As far as I can tell what was described here is more of a "semantic" difference between the programs, but not a behavioral one. 
The only possible problem I understand may occur is that we may miss submission of some early container events from the kernel? If so maybe we should try to mark them as container events earlier, that's a tracee-ebpf only problem though.

AlonZivony:
The problem here is that tracee-rules and tracee-ebpf are not matching in their definition of when a container really starts.
[...]

PR #1984, opened 2022-07-19
Bugfix: container created flag

AlonZivony:
## Initial Checklist

## Description (git log)

Add a flag to all events marking whether the process is a part of a
started container.
This way tracee-rules will be able to use the logic of tracee-ebpf to
determine the origin of the event.

Fixes: #1967

## Type of change

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -307,9 +307,15 @@ enum event_id_e
 #define CONT_ID_LEN          12
 #define CONT_ID_MIN_FULL_LEN 64
 
+enum context_flags_e
+{
+    CONTAINER_STARTED_FLAG = (1 << 0)
+};
+
 enum container_state_e
 {
-    CONTAINER_EXISTED = 1, // container existed before tracee was started
+    CONTAINER_UNKNOWN = 0, // mark that container state is unknown
+    CONTAINER_EXISTED,     // container existed before tracee was started
     CONTAINER_CREATED,     // new cgroup path created
     CONTAINER_STARTED      // a process in the cgroup executed a new binary
 };
@@ -473,7 +479,7 @@ typedef struct task_context {
     u32 pid_id;
     char comm[TASK_COMM_LEN];
     char uts_name[TASK_COMM_LEN];
-    u32 padding;
+    u32 flags;
 } task_context_t;
 
 typedef struct event_context {
@@ -506,6 +512,7 @@ typedef struct task_info {
     bool new_task;        // set if this task was started after tracee. Used with new_pid filter
     bool follow;          // set if this task was traced before. Used with the follow filter
     int should_trace;     // last decision of should_trace()
+    u8 container_state;   // the state of the container the task resides in
 } task_info_t;
 
 typedef struct bin_args {
@@ -1380,6 +1387,7 @@ init_context(event_context_t *context, struct task_struct *task, u32 options)
     context->task.mnt_id = get_task_mnt_ns_id(task);
     context->task.pid_id = get_task_pid_ns_id(task);
     context->task.uid = bpf_get_current_uid_gid();
+    context->task.flags = 0;
     bpf_get_current_comm(&context->task.comm, sizeof(context->task.comm));
     char *uts_name = get_task_uts_name(task);
     if (uts_name)
@@ -1426,11 +1434,13 @@ static __always_inline int init_event_data(event_data_t *data, void *ctx)
     if (data->task_info == NULL) {
         return 0;
     }
+    bool container_lookup_required = true;
     if (!exist) {
         data->task_info->syscall_traced = false;
         data->task_info->new_task = false;
         data->task_info->follow = false;
         data->task_info->recompute_scope = true;
+        data->task_info->container_state = CONTAINER_UNKNOWN;
     } else {
         // check if we need to recompute scope due to context change
         task_context_t *old_context = &data->task_info->context;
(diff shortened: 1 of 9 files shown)


pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1444,6 +1453,27 @@ static __always_inline int init_event_data(event_data_t *data, void *ctx)
...
             *(u64 *) &old_context->uts_name[8] != *(u64 *) &new_context->uts_name[8])
             data->task_info->recompute_scope = true;
+
+        if ((old_context->flags & CONTAINER_STARTED_FLAG) != 0) {
+            new_context->flags |= CONTAINER_STARTED_FLAG;
+            should_lookup_container_state = false;
+        }
+    }
+
+    if (should_lookup_container_state) {
```
rafaeldtinoco, line 1463: Okay, so ... if the task exists, `should_lookup_container_state` will **99% of the times** be **TRUE**, no ? It will only be **FALSE** if `flags` does NOT contain `CONTAINER_STARTED_FLAG`., which happens ONLY when a `cgroup_mkdir` is fired, and a sched happens (due to the other state machine) is that right ?

If that is right, why should we even cache the state, since we're doing the bpf map lookup almost every time anyway ?

It also feels weird to me that we have 2 state machines for the same information, I do understand the reasoning (trying to cache, just like the context) though.

rafaeldtinoco (commented):
Left some comments for discussion.

types/trace/trace.go
```diff
@@ -12,50 +12,59 @@ import (
...
+// Origin derive the EventOrigin of a trace.Event
 func (e Event) Origin() EventOrigin {
-	if e.ContainerID != "" || e.ProcessID != e.HostProcessID {
+	if e.ContextFlags.ContainerStarted {
 		return ContainerOrigin
-	} else {
-		return HostOrigin
 	}
+	if e.ContainerID != "" || e.ProcessID != e.HostProcessID {
+		return ContainerInitOrigin
```
yanivagman, line 65: `e.ProcessID != e.HostProcessID` is unnecessarily container init. pid namespace might be used without what we define as a "container"

yanivagman (changes requested):

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: > So to your question - it's not true to say that 99% of the time it will be true, as it will be false for any container task.

I'm missing something I guess, just thought it was worth clarifying (even if I'm wrong):

We set the `task_info->context.flags` to `CONTAINER_STARTED_FLAG` at `sched_process_exec` IF the `task_info->container_state` is `CONTAINER_CREATED`.

`task_info->container_state` will only be `CONTAINER_CREATED` if a previous `cgroup_mkdir` happened. So the flag `CONTAINER_STARTED_FLAG` will only exist after a `cgroup_mkdir` happened and the next `sched_process_exec`.

Then...

```C
bool should_lookup_container_state = true;

if (!exist) {
 ...
    data->task_info->container_state = CONTAINER_UNKNOWN;

} else {

    // check if we need to recompute scope due to context change
    task_context_t *old_context = &data->task_info->context;
    task_context_t *new_context = &data->context.task;

...
    if ((old_context->flags & CONTAINER_STARTED_FLAG) != 0) {
        new_context->flags |= CONTAINER_STARTED_FLAG;
        should_lookup_container_state = false;
    }
}
```

`CONTAINER_STARTED_FLAG` will exist in previous context. We are going to clear the flag for the new context, about to update the old context. The next time this logic runs, the flag won't exist and `should_lookup_container_state` will be always true, no ?

yanivagman: > We set the `task_info->context.flags` to `CONTAINER_STARTED_FLAG` at `sched_process_exec` IF the `task_info->container_state` is `CONTAINER_CREATED`

No, we set it to `CONTAINER_STARTED` (no extra `_FLAG`)


> `task_info->container_state` will only be `CONTAINER_CREATED` if a previous `cgroup_mkdir` happened.

correct

> So the flag `CONTAINER_STARTED_FLAG` will only exist after a `cgroup_mkdir` happened and the next `sched_process_exec`.

The container state will be `CONTAINER_STARTED` (again. without `FLAG`) and the flag is also set for `CONTAINER_EXISTED` as can be seen in line 1469 below

> We are going to clear the flag for the new context, about to update the old context.

Why is that? We just set it again for the new_context

rafaeldtinoco: Well... Do the following change:

```patch
index c1d3cc98..030a6222 100644
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -1460,8 +1460,11 @@ static __always_inline int init_event_data(event_data_t *data, void *ctx)
         }
     }

+    task_context_t *new_context = &data->context.task;
+
     if (should_lookup_container_state) {
-        task_context_t *new_context = &data->context.task;
+        bpf_printk("MAPLOOKUP: %s (pid=%u)", new_context->comm, new_context->pid);
+
         u32 cgroup_id_lsb = new_context->cgroup_id;
         u8 *state = bpf_map_lookup_elem(&containers_map, &cgroup_id_lsb);
         if (state != NULL) {
@@ -1473,6 +1476,7 @@ static __always_inline int init_event_data(event_data_t *data, void *ctx)
             data->task_info->container_state = CONTAINER_UNKNOWN;
         }
     } else {
+        bpf_printk("NOMAPLOOKUP %s (pid=%u)", new_context->comm, new_context->pid);
         data->task_info->container_state = CONTAINER_STARTED;
     }
```

and observe for yourself:

image

yanivagman: Observe what? Are these container processes? If they are, something is wrong. Else, this is expected.

rafaeldtinoco: Alright then. I thought you wanted to avoid map lookups if the task context did not change. Now you do have another map lookup every time init_event_data() is called. If that is what you want then yep, everything is fine.

yanivagman: So in this case, yes. we indeed have an extra map lookup. Ideally we can remove it, but for now it is required in order to get the container state of a given cgroup id. While the cgroup id of the task doesn't have to change (and most usually won't), the container state for that cgroup might. Like I wrote above, we might do optimizations for this to avoid this extra lookup most of the time:

> A possible future optimization might mark the task as a host task (by comparing to the host cgroup id), and updating the field in cgroup_attach_task (which might happen, for example, when a container is started).

This requires initializing the host cgroup id in the configuration map, and we can possibly make it part of the scope of this PR if we think it is necessary

rafaeldtinoco: I understand that comment (optimization) now. I had in my head we wanted to avoid the lookups from the beginning (that is why I started this thread).

yanivagman (approved):
LGTM

rafaeldtinoco:
Im reopening this because of https://github.com/aquasecurity/tracee/pull/2029 and regressions observed when we added more tests to each PR. I have reverted the following PRs:

https://github.com/aquasecurity/tracee/pull/1984
https://github.com/aquasecurity/tracee/pull/2007

rafaeldtinoco:
Im reopening this because of https://github.com/aquasecurity/tracee/pull/2029 and regressions observed when we added more tests to each PR. I have reverted the following PRs:

https://github.com/aquasecurity/tracee/pull/1984
https://github.com/aquasecurity/tracee/pull/2007

PR #1998, opened 2022-07-20
containers: containers_map set by package initialization

rafaeldtinoco:
## Initial Checklist

## Description (git log)

There is no need to have the containers_map bpf name given as argument.

## Type of change

## How Has This Been Tested?

Rely in PR workflows.

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/containers/containers.go
+++ b/pkg/containers/containers.go
@@ -26,12 +26,13 @@ const cgroupV2FsType = "cgroup2"
 
 // Containers contains information about running containers in the host.
 type Containers struct {
-	cgroupV1 bool
-	cgroupMP string
-	cgroups  map[uint32]CgroupInfo
-	deleted  []uint64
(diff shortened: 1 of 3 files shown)


NDStrahilevitz (commented):
I think it might be better if we do pass the map name somewhere, possibly in `New` instead? WDYT? I will probably do the same in the filters refactor as you suggested (pass the map name as a filter parameter).

rafaeldtinoco:
> I think it might be better if we do pass the map name somewhere more external, possibly in `New` instead? WDYT? I will probably do the same in the filters refactor as you suggested (pass the map name as a filter parameter).

Yes, lets have this as "default": to give the names during Init call, externally. I'm cool with that.

NDStrahilevitz (approved):
LGTM!

Issue #1727, opened 2022-05-03
[FEAT] Tracee internal Filters API


NDStrahilevitz:
## Prerequisites

- [ ] This issue is an EPIC issue (add label: EPIC).
- [x] This issue is an EPIC TASK (add issue to EPIC description).

Select one OR another:

- [x] I'll create a PR to implement this feature (assign to yourself).

## Feature description

In order to enable runtime configuration of filters, the capabilities of current filtering logic in tracee must be extended to allow adding new filters on the go.

As such a new `FilterGroup` interface should be defined which will describe the logical steps needed to add a filter.
This API will be implemented by the existing filters `StringFilter` `UintFilter`, etc.
These new logics will be in `pkg/filters`.

## Additional Information (feature drawings, files, logs, etc)

Interface draft:
```
type FilterGroup interface {
    Add(val, evaluator) err
    Reconcile() err
    Remove(val, evaluator) err
}
```

PR #2009, opened 2022-07-24
Feature/filter protocol ebpf infrastructure

NDStrahilevitz:
## Initial Checklist

## Description (git log)

This PR refactors the existing filters package, preparing it for using protocol filters and then integrating them.

Additionally, it makes use of the new protocol.Filter to create a later common interface between parsing the cli into the main Filter struct, and a gRPC service we will add later to receive mediated filters from tracee-rules.
It refactors filter parsing from the CLI - switching to parse into []protocol.Filter which is down the line parsed into a tracee.Filter.

While this tries to maintain a high degree of similarity to the previous filtering logic, this PR possibly breaks previous behavior.

Opened instead of #1750

Fixes: #1727

## Type of change

## How Has This Been Tested?

New unit tests were added to the new filters package.
New unit test was added to check errors in parsing protocol.Filters in `pkg/ebpf/filter_test.go`
Tests were moved and changed according to the new flow of filter info.

`make test-unit`

## Final Checklist:

## Git Log Checklist:

My commits logs have:


pkg/filters/sets/suffixset.go
```diff
@@ -0,0 +1,63 @@
...
+	if lenVal < set.minLen {
+		return false
+	}
+
+	for suffixLen := set.minLen; suffixLen <= set.maxLen; suffixLen++ {
+		if lenVal < suffixLen {
+			return false
+		}
+
+		check := val[lenVal-suffixLen : lenVal]
```
yanivagman, line 53: This is the only line that is different between prefixset and suffix set.
Seems to me that those two can be combined to avoid code duplication.

cmd/tracee-ebpf/flags/filter.go
```diff
@@ -132,7 +132,7 @@ func PrepareFilter(filtersArr []string) (tracee.Filter, error) {
 			Filters: make(map[events.ID]filters.IntFilter),
 		},
 		ArgFilter: &filters.ArgFilter{
-			Filters: make(map[events.ID]map[string]filters.ArgFilterVal),
+			Filters: make(map[events.ID]map[string]filters.StringFilter),
```
yanivagman: Today we do argument comparisons by converting to string first, but in the future, we might perform comparisons of other types (e.g. int), which will also be more efficient. For this reason the original implementation defined a map value of its own (ArgFilterVal), so why changing this here?

pkg/filters/sets/suffixset.go
NDStrahilevitz: True, I personally find the code simpler with separate responsibilities between the sets, but maybe there's a performance argument to be made here, so I will check that option.

pkg/events/events.go
```diff
@@ -86,6 +86,15 @@ func (e *eventDefinitions) NamesToIDs() map[string]ID {
 	return namesToIds
 }
 
+func (e *eventDefinitions) GetID(eventName string) (ID, bool) {
```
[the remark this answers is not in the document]
NDStrahilevitz: I use it later in the context filter to avoid a repeated map build.
I also think it's a useful helper to have, we could definitely use it to simplify code in other places, instead of doing the map build.

cmd/tracee-ebpf/flags/filter.go
NDStrahilevitz: Well, we're moving it to ebpf later anyway. I didn't see a use in having them separated since they are currently just the same, if we decide for some reason not to move this logic to ebpf anyway, and further improve it in userspace we would probably need to create a custom struct for it anyway. Otherwise it would just be duplication from the current StringFilter.

As a compromise I can change it to a separate struct which will embed the StringFilter like so:
```
type eventArgFilter struct {
    StringFilter
}
```

pkg/filters/args.go
```diff
@@ -8,10 +8,17 @@ import (
...
 type ArgFilter struct {
-	Filters map[events.ID]map[string]StringFilter // key to the first map is event id, and to the second map the argument name
+	Filters map[events.ID]map[string]*StringFilter // key to the first map is event id, and to the second map the argument name
 	Enabled bool
 }
 
+func NewArgFilter() *ArgFilter {
+	return &ArgFilter{
+		Filters: map[events.ID]map[string]*StringFilter{},
+		Enabled: false,
```
yanivagman: Default boolean value in go is false, so no need to initialize enabled fields in the constructors

pkg/filters/containers.go
```diff
@@ -9,50 +9,24 @@ import (
...
-
-	filter.Equal = strFilter.Equal
-	filter.NotEqual = strFilter.NotEqual
-
-	return nil
-}
-
 func (filter *ContainerFilter) InitBPF(bpfModule *bpf.Module, conts *containers.Containers) error {
-	if !filter.Enabled {
+	if !filter.Enabled() {
```
yanivagman: I don't see an Enabled() function of ContainerFilter, does this compile?

pkg/filters/containers.go
```diff
@@ -84,11 +58,3 @@ func (filter *ContainerFilter) InitBPF(bpfModule *bpf.Module, conts *containers.
 
 	return nil
 }
-
-func (filter *ContainerFilter) FilterOut() bool {
```
yanivagman: I don't see that you've updated references to this function, does this compile?

pkg/filters/containers.go
```diff
@@ -9,50 +9,24 @@ import (
...
 	return &ContainerFilter{
-		Equal:      []string{},
-		NotEqual:   []string{},
-		Enabled:    false,
-		bpfMapName: mapName,
+		BPFStringFilter: NewBPFStringFilter(mapName),
 	}
 }
 
-func (filter *ContainerFilter) Parse(operatorAndValues string) error {
```
yanivagman: I don't see that you updated references to this function. Does it compile?

pkg/filters/processtree.go
```diff
@@ -17,10 +17,11 @@ type ProcessTreeFilter struct {
 	Enabled bool
 }
 
-func NewProcessTreeFilter() *ProcessTreeFilter {
+func NewProcessTreeFilter(mapName string) *ProcessTreeFilter {
 	return &ProcessTreeFilter{
 		PIDs:    map[uint32]bool{},
-		Enabled: false,
+		enabled: false,
```
yanivagman, line 24: this field is not exposed now, but no API to get it is given, does this compile?

pkg/filters/containers.go
NDStrahilevitz: It compiles, this struct embeds BPFStringFilter _in this commit_. Do note I change this filter in one of the last commits.

cmd/tracee-ebpf/flags/filter.go
```diff
@@ -82,19 +82,19 @@ To 'escape' those operators, please use single quotes, e.g.: 'uid>0'
...
+		UTSFilter:         filters.NewBPFStringFilter(tracee.UTSFilterMap),
+		CommFilter:        filters.NewBPFStringFilter(tracee.CommFilterMap),
 		ContFilter:        filters.NewBoolFilter(),
 		NewContFilter:     filters.NewBoolFilter(),
-		ContIDFilter:      filters.NewContainerFilter("cgroup_id_filter"),
+		ContIDFilter:      filters.NewContainerFilter(tracee.CgroupIdFilterMap),
 		RetFilter:         filters.NewRetFilter(),
 		ArgFilter:         filters.NewArgFilter(),
-		ProcessTreeFilter: filters.NewProcessTreeFilter(),
+		ProcessTreeFilter: filters.NewProcessTreeFilter(tracee.ProcessTreeFilterMap),
```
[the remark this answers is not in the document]
NDStrahilevitz: They're feature specific filters, obviously used for ebpf. The string filters can be used in a non-ebpf context which is why the separation exists

pkg/filters/args.go
NDStrahilevitz: It was done to be explicit, but I can remove it.

cmd/tracee-ebpf/flags/filter.go
NDStrahilevitz: Not sure if you thumbs upped for embedding or keeping as is

pkg/filters/containers.go
NDStrahilevitz: ditto, embedded struct inherits this from StringFilter, and ditto that I change this filter later

pkg/filters/processtree.go
NDStrahilevitz: fixed

pkg/filters/containers.go
NDStrahilevitz: fixed

pkg/filters/int.go
```diff
@@ -28,8 +33,8 @@ func newIntFilter(is32Bit bool) *IntFilter {
 	return &IntFilter{
 		Equal:    []int64{},
 		NotEqual: []int64{},
-		Greater:  GreaterNotSetInt,
-		Less:     LessNotSetInt,
+		Greater:  maxIntVal,
+		Less:     minIntVal,
```
yanivagman: In BPF code we now call Greater(Than) Min (to signify that this is the minimal allowed value), and Less(LowerThan) Max. I think we should use the same terminology here to avoid confusions.

pkg/filters/int.go
```diff
@@ -67,11 +72,11 @@ func (filter *IntFilter) Parse(operatorAndValues string) error {
 		case "!=":
 			filter.NotEqual = append(filter.NotEqual, val)
 		case ">":
-			if (filter.Greater == GreaterNotSetInt) || (val > filter.Greater) {
+			if (filter.Greater == maxIntVal) || (val > filter.Greater) {
```
yanivagman: with the terminology used in out BPF code, this line would look like:
`if (filter.Min == minNotSet) || (val > filter.Min) {`

pkg/filters/args.go
```diff
@@ -19,6 +20,55 @@ func NewArgFilter() *ArgFilter {
...
+			(suffixCheck && !prefixCheck && strings.HasSuffix(argValStr, f)) ||
+			(prefixCheck && suffixCheck && strings.Contains(argValStr, f)) {
+			return true
+		}
+	}
+	return false
+}
+
+func (filter *ArgFilter) Filter(eventID events.ID, args []trace.Argument) bool {
+	if filter.Enabled {
```
yanivagman: to avoid extra indentation, you can use:
```
if !filter.Enabled {
    return true
}
```

pkg/filters/retval.go
```diff
@@ -19,6 +19,35 @@ func NewRetFilter() *RetFilter {
 	}
 }
 
+func (filter *RetFilter) Filter(eventID events.ID, retVal int64) bool {
+	if filter.Enabled {
```
yanivagman: to avoid extra indentation, you can use:
```
if !filter.Enabled {
    return true
}
```

pkg/filters/retval.go
```diff
@@ -19,6 +19,35 @@ func NewRetFilter() *RetFilter {
...
+		if filter, ok := filter.Filters[eventID]; ok {
```
yanivagman, line 38: ditto

pkg/filters/args.go
```diff
@@ -130,5 +129,29 @@ func (filter *ArgFilter) Parse(filterName string, operatorAndValues string, even
...
+}
+
+func (filter *ArgFilter) Disable() {
+	filter.enabled = false
+	for _, filterMap := range filter.Filters {
+		for _, f := range filterMap {
+			f.Disable()
+		}
+	}
+}
```
yanivagman, line 126: Just a thought for a possible future PR -
I'm not sure how useful it will be to have a generic Enable/Disable methods for this (argument) filter, but letting these methods also accept an event id and argument names, will allow to enable/disable a specific argument filter, which might be useful for changing filters dynamically.

pkg/filters/int.go
```diff
@@ -13,12 +13,12 @@ const (
...
-	Equal    []int64
-	NotEqual []int64
-	Greater  int64
-	Less     int64
-	Is32Bit  bool
-	enabled  bool
+	Equal       []int64
+	NotEqual    []int64
+	GreaterThan int64
+	LessThan    int64
```
yanivagman: See my comment in the previous commit about using Min/Max instead of these

pkg/filters/args.go
NDStrahilevitz: We could have a separate function to enable disable an Arg, that's a good idea, but I also think a general Enable Disable is good for feature parity and having a common Filter interface.

pkg/filters/args.go
NDStrahilevitz: Sure, that's how I did it in #1750

pkg/filters/int.go
NDStrahilevitz: I export these as `Minimum()` and `Maximum()` for usage in the ebpf context, in the context of the internal logic of the filter I find it clearer with LessThan and GreaterThan (because there's also the "ongoing" logic of setting these values while parsing, userspace filters have further context to them than ebpf).

pkg/filters/int.go
NDStrahilevitz: This logic is completely changed further down the line anyway, but I think `defaultGreatest` or something might be a better name in the userspace context.

pkg/filters/int.go
```diff
@@ -7,6 +7,11 @@ import (
 	"strings"
 )
 
+const (
+	minIntVal int64 = math.MinInt64
+	maxIntVal int64 = math.MaxInt64
+)
+
```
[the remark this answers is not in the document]
NDStrahilevitz: I agree that we can choose better names, `defaultMin` or `defaultGreaterThan` would fit best for the userspace context in my opinion, wdyt?

pkg/filters/int.go
```diff
@@ -53,57 +67,89 @@ func (f *IntFilter) Enabled() bool {
...
-func (filter *IntFilter) Parse(operatorAndValues string) error {
-	if len(operatorAndValues) < 2 {
-		return fmt.Errorf("invalid operator and/or values given to filter: %s", operatorAndValues)
+// priority goes by (from most significant):
+// 1. equality
+// 2. greater
+// 3. lesser
+// 4. non equality
+func (f *IntFilter) Filter(val int64) bool {
+	result := !f.enabled || f.equal[val] || val > f.greaterThan || val < f.lessThan
```
yanivagman, line 83: What about the case that the user want `val > x && val` < y ?

pkg/filters/int.go
```diff
@@ -53,57 +67,89 @@ func (f *IntFilter) Enabled() bool {
...
 
-	if operatorString == "!" {
-		if len(operatorAndValues) < 3 {
-			return fmt.Errorf("invalid operator and/or values given to filter: %s", operatorAndValues)
-		}
-		operatorString = operatorAndValues[0:2]
-		valuesString = operatorAndValues[2:]
+func (f *IntFilter) validate(val int64) bool {
+	const maxIntVal32Bit = math.MaxInt32
+	const minIntVal32Bit = math.MinInt32
```
yanivagman, line 92: I don't see the point of redeclaring these consts

pkg/filters/string.go
```diff
@@ -6,55 +6,174 @@ import (
...
+	vals := make([]string, len(filterReq.Value))
+	for i := 0; i < len(arr); i++ {
+		val := arr[i]
+		valStr, ok := val.(string)
+		if !ok {
+			return fmt.Errorf("failed to add to filter: invalid value: %s", val)
+		}
+		vals[i] = valStr
+	}
+	for _, val := range vals {
```
yanivagman, line 121: Again, not sure why we need this loop

pkg/filters/string.go
```diff
@@ -6,55 +6,174 @@ import (
...
+		return UnsupportedOperator(operator)
+	}
 	return nil
 }
 
+func (f *StringFilter) addEqual(val string) {
+	prefixSet := val[len(val)-1] == '*'
+	suffixSet := val[0] == '*'
+	if prefixSet && suffixSet && len(val) > 1 {
+		f.contains[val[1:len(val)-1]] = true
```
yanivagman, line 146: Will val with an input `**` pass here?

pkg/filters/string.go
```diff
@@ -6,55 +6,174 @@ import (
...
+		return
+	}
+	if prefixSet {
+		f.prefixes.Put(val[:len(val)-1])
```
yanivagman, line 150: Will val of `*` pass here?

pkg/filters/string.go
```diff
@@ -6,55 +6,174 @@ import (
...
+	if !prefixSet && !suffixSet {
+		f.equal[val] = true
+		return
+	}
+}
+
+func (f *StringFilter) addNotEqual(val string) {
+	prefixSet := val[len(val)-1] == '*'
+	suffixSet := val[0] == '*'
+	if prefixSet {
```
yanivagman, line 166: We should probably check that either `*` and `**` are not given here as val

yanivagman (changes requested):
This set of comments is relevant for the main commit - "integrate protocol into filters parsing"

go.mod
```diff
@@ -105,3 +105,5 @@ require (
 )
 
 replace github.com/kubernetes/cri-api => k8s.io/cri-api v0.23.5-rc.0
+
+replace github.com/aquasecurity/tracee/types => github.com/NDStrahilevitz/tracee/types v0.0.0-20220714132244-d08bd36cfbde
```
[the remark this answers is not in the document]
NDStrahilevitz: When we merge this types commit I will fix this commit to change the types version.

cmd/tracee-ebpf/flags/filter.go
```diff
@@ -72,283 +71,175 @@ Examples:
...
+		return protocol.Filter{}, EmptyValueFilter()
 	}
-	if len(eventsToTrace) == 0 && len(setsToTrace) == 0 {
-		setsToTrace = append(setsToTrace, "default")
+	splitVals := strings.Split(valuesStr, ",")
+	values := []interface{}{}
+
+	//parse vals to actual types
+	for _, val := range splitVals {
+		values = append(values, val)
```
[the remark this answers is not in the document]
NDStrahilevitz: I should fix the comment to interface{}

pkg/ebpf/filter.go
```diff
@@ -0,0 +1,429 @@
...
+
+	for _, filterReq := range filterRequests {
+		field := filterReq.Field
+		if strings.HasSuffix(field, ".retval") {
+			filterMap[retval] = append(filterMap[retval], filterReq)
+			continue
+		}
+
+		if strings.Contains(field, ".") {
+			filterMap[args] = append(filterMap[retval], filterReq)
```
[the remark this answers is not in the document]
NDStrahilevitz: Possibly a missing test for these cases, will fix and add

pkg/ebpf/tracee.go
```diff
@@ -105,26 +105,6 @@ func (tc Config) Validate() error {
...
-				if eventParams[i].Name == argName {
-					argFound = true
-					break
-				}
-			}
-			if !argFound {
-				return fmt.Errorf("invalid argument filter argument name: %s", argName)
-			}
-		}
-	}
```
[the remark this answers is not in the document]
NDStrahilevitz: We can't validate these anymore as the Filters map is private. Also, validation is done while parsing.

pkg/filters/containers.go
```diff
@@ -6,52 +6,60 @@ import (
...
+	filter := &ContainerFilter{
+		BPFStringFilter: stringFilter,
 	}
+
+	return filter, nil
 }
 
-func (filter *ContainerFilter) InitBPF(bpfModule *bpf.Module, conts *containers.Containers) error {
-	if !filter.Enabled() {
+func (f *ContainerFilter) InitBPF(bpfModule *bpf.Module, containers *containers.Containers) error {
```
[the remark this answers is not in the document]
NDStrahilevitz: It's a scoped rename inside the function. I can change it back but it shouldn't affect external code.

pkg/filters/int.go
NDStrahilevitz: Should we support that? If `x < y` then we're good anyway, on the other hand it's just a conflict, which means we either break the filter or, as is done here, and I remember adding a test demonstrating it, that case is done as `val > x || val < y`.
As we've talked, these filters are meant to include as much as possible for all requests, so a conflicting request won't and shouldn't override or break the filter.

pkg/ebpf/filter.go
```diff
@@ -0,0 +1,429 @@
...
+	vals := make([]string, len(filterReq.Value))
+	for i := 0; i < len(arr); i++ {
+		val := arr[i]
+		valStr, ok := val.(string)
+		if !ok {
+			return fmt.Errorf("failed to add to filter: invalid value")
+		}
+		vals[i] = valStr
+	}
+	for _, val := range vals {
```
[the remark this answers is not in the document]
NDStrahilevitz: First loop is converting the type to string, then second one is adding. We don't actually need them separate, when i've written it I just thought of the steps sequentially, we can combine.

pkg/filters/string.go
NDStrahilevitz: ditto

pkg/filters/string.go
NDStrahilevitz: I'll check that

cmd/tracee-ebpf/flags/filter.go
```diff
@@ -72,283 +71,162 @@ Examples:
...
-		}
+// expandFilterField allows filter field aliases
+func expandFilterField(field string) string {
+	switch field {
+	case "e":
+		return "event"
+	case "c":
+		return "container"
+	default:
+		return field
```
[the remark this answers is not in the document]
NDStrahilevitz: outdated

cmd/tracee-ebpf/flags/filter.go
```diff
@@ -72,283 +71,162 @@ Examples:
   --trace openat.pathname=/tmp*                                | only trace 'openat' events that have 'pathname' prefixed by "/tmp"
   --trace openat.pathname!=/tmp/1,/bin/ls                      | don't trace 'openat' events that have 'pathname' equals /tmp/1 or /bin/ls
   --trace comm=bash --trace follow                             | trace all events that originated from bash or from one of the processes spawned by bash
-  --trace net=docker0 			                       | trace the net events over docker0 interface
+  --trace net=docker0 			                               | trace the net events over docker0 interface
```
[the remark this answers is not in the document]
NDStrahilevitz: outdated

pkg/filters/filters.go
```diff
@@ -1,5 +1,56 @@
...
+
+	"github.com/aquasecurity/tracee/types/protocol"
+)
+
+type Operator uint
+
+const (
+	Equal Operator = iota
+	NotEqual
+	Lesser
```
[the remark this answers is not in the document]
NDStrahilevitz: Agree, when we approve the relevant types PR, I will rename it to Lower then. Then as part of integrating the actual commit i'll also rename it here, so there won't be a mismatch in between.

pkg/filters/int.go
NDStrahilevitz: Already addressed this comment, I mostly did it as a mnemonic for myself while writing the code. And technically if we ever wanted different values it would be a one time change here. 
I think it's more appropriate to just change the names as we've also discussed.

pkg/filters/args.go
```diff
@@ -49,45 +43,38 @@ func (filter *ArgFilter) Filter(eventID events.ID, args []trace.Argument) bool {
...
+	field := filterReq.Field
+	parts := strings.Split(field, ".")
+	if len(parts) != 2 {
+		return fmt.Errorf("invalid format for arg filter: %s", field)
 	}
+	eventName := parts[0]
+	argName := parts[1]
 
-	eventDefinition, ok := events.Definitions.GetSafe(id)
+	id, ok := events.Definitions.GetID(eventName)
```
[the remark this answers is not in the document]
NDStrahilevitz: Is it critical? This happens on the initial parse, it's not a performance critical path. When and if we make the move to runtime modification of filters, we will probably change stuff here again.

cmd/tracee-ebpf/flags/filter.go
```diff
@@ -72,283 +71,175 @@ Examples:
...
+		filterName = expandFilterField(filterName)
+		err := validateFilterName(filterName)
+		if err != nil {
+			return protocol.Filter{}, err
+		}
+		return protocol.EqualFilter(filterName, false), nil
+	}
+	// if operator is last char then the value is missing
+	if operatorIndex+1 == len(filterStr) {
+		return protocol.Filter{}, EmptyValueFilter()
```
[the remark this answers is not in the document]
NDStrahilevitz: Added a test now to check for "uid>=" and it returns an expected error, so it should be ok.

pkg/filters/string.go
NDStrahilevitz: They did pass with no errors, added tests as well.

pkg/filters/string.go
NDStrahilevitz: Did that and added tests

pkg/ebpf/filter.go
```diff
@@ -0,0 +1,429 @@
...
+	if len(eventsToTrace) == 0 && len(setsToTrace) == 0 {
+		setsToTrace = append(setsToTrace, "default")
+	}
+
+	for _, set := range setsToTrace {
+		setEvents, ok := setsToEvents[set]
+		if !ok {
+			return nil, fmt.Errorf("invalid set to trace: %s", set)
+		}
+		res = append(res, setEvents...)
```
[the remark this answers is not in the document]
NDStrahilevitz: Fixed.

pkg/ebpf/filter.go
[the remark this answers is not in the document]
NDStrahilevitz: Tests didn't catch it because we test only adding one argfilter, if so the retval map is empty so the test passes since the input is fine. Case missing is adding both a retval and an arg filter

pkg/filters/sets/suffixset.go
NDStrahilevitz: So actually, this probably wouldn't really work, because we would need to track both the prefix and suffix lengths separately.
So it would still be different loops, and the code would be less readable IMO.

yanivagman: What I mean is, that you can simply pass the filter's constructor a boolean argument to describe prefix/suffix, and keep two filters for prefix or suffix. But the you can also remove one of the files and leave only one, as they are almost identical.

cmd/tracee-ebpf/flags/filter.go
yanivagman: Thumbs up for the compromise :-)

pkg/filters/containers.go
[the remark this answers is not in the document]
yanivagman: Yes, I understand that it's scoped, just saying it is confusing.

pkg/filters/string.go
yanivagman: But should it pass? ** effectively means all strings, same for * - they both shouldn't pass

pkg/filters/string.go
yanivagman: ditto

pkg/filters/int.go
```diff
@@ -104,15 +103,17 @@ func (f *IntFilter) addNotEqual(val int64) {
 	f.notEqual[val] = true
 }
 
-func (f *IntFilter) addLesser(val int64) {
-	if val > f.lessThan {
-		f.lessThan = val
+func (f *IntFilter) addLesserThan(val int64) {
```
yanivagman, line 106: You can replace Less->Lower in this commit as well

pkg/filters/string.go
NDStrahilevitz: Didn't clarify, no they shouldn't pass, I fixed that, and added tests.

pkg/filters/int.go
NDStrahilevitz: Critical? I'd rather not deal with cropping up conflict errors in the rebase again, that's why I did the styling changes in one final commit.

yanivagman: But this is the last commit - you can just ammend it without rebasing

NDStrahilevitz: Oh you mean change LesserThan to LowerThan? I'd rather wait on the types change for that.

pkg/filters/int.go
NDStrahilevitz: Fixed

pkg/filters/sets/suffixset.go
NDStrahilevitz: I don't think it would make the code more readable, especially if internally I would need to use something like `prefixOrSuffix` everywhere, would just make it more verbose. 
I think the duplication, in this specific case, makes the codes intent more clear (this is code for suffix comparison, this is for prefix comparison).

NDStrahilevitz:
Closing this (again) in favor of #2063

Issue #2005, opened 2022-07-24
[BUG]  Fix integration tests


NDStrahilevitz:
## Prerequisites

- [x] This affects latest released version.
- [x] This affects current development tree (origin/HEAD).
- [x] There isn't an issue describing the bug.

Select one OR another:

- [x] I'm going to create a PR to solve this (assign to yourself).
- [ ] Someone else should solve this.

## Bug description

Currently integrations tests are:
1. Being skipped in `make test-integration`
2. Broken and failing when attempting to execute
3. Not straightforward to run

## Steps to reproduce

1. Remove the `t.Skip()` line in `tests/integration.go`
2. run `make test-integration`

## Additional Information (files, logs, etc)

In order to fix this, I will slightly rewrite the tests to run "locally", as in, setup a tracee instance and run separately a triggering mechanism. This will also allow us to test other features later.
A further cleanup of functions in `tracee-ebpf/main.go` is required in order to allow reuse in the integration tests, that will be a separate refactor PR.

Issue #1949, opened 2022-07-10
[BUG] cdk_linux_amd64 binary is recognized as malware


roikol:
## Prerequisites

- [ ] This affects latest released version.
- [x] This affects current development tree (origin/HEAD).
- [x] There isn't an issue describing the bug.

Select one OR another:

- [ ] I'm going to create a PR to solve this (assign to yourself).
- [x] Someone else should solve this.

## Bug description

file `tracee/tests/tracee-tester/cdk_linux_amd64` is recognized as trojan by ESET antivirus.
so when cloning the project into my local station, this file is quarantined by the AV.

![image](https://user-images.githubusercontent.com/17714090/178139476-60f65249-1e34-411d-8043-bf8fac8aa113.png)

![image](https://user-images.githubusercontent.com/17714090/178139575-877bd9e9-ce4c-4731-88d4-1d32b05e45eb.png)

## Steps to reproduce

clone tracee on a station with ESET

PR #2017, opened 2022-07-25
recomission integration tests

NDStrahilevitz:
## Initial Checklist

## Description (git log)

```
integration: recomission the tests
    
    Integration tests were  previously skipped and as such unmaintained
    and were broken over time.
    Fix the tests by redesigning them to locally run tracee.
```

Fixes: #2005 

## Type of change

## How Has This Been Tested?

Run tests locally by running:
`sudo env "PATH=$PATH" make test-integration`

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/.github/actions/build-dependencies/action.yaml
+++ b/.github/actions/build-dependencies/action.yaml
@@ -41,3 +41,11 @@ runs:
       run: |
         go install honnef.co/go/tools/cmd/staticcheck@latest
       shell: bash
+    - name: Install docker
+      run: |
+        sudo apt-get install --yes ca-certificates curl gnupg lsb-release
+        sudo mkdir -p /etc/apt/keyrings
+        curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg
+        echo "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.gpg] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+        sudo apt-get update
+        sudo apt-get install docker-ce docker-ce-cli containerd.io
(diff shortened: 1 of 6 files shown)


rafaeldtinoco:
@grantseltzer fyi

rafaeldtinoco (commented):
1. I think we should move tests/tracee-tester to an external branch (due to https://github.com/aquasecurity/tracee/issues/1949) and have an action in that new branch that creates "tracee-tester" docker image and publishes it. This way we can add more tests to it (there) according to the things we want to this in tracee (here).

2. I have also played a bit with reorganizing Makefile and adding some flags to the test targets (https://gist.github.com/rafaeldtinoco/468998d513c9a61426d239527c6b4287) and got the following **several** data races for `integration_test_.go`:

image

I believe that is because of parallel calls of `startTracee` for the multiple testCases.

NDStrahilevitz:
> 1. I think we should move tests/tracee-tester to an external branch (due to #1949) and have an action in that new branch that creates "tracee-tester" docker image and publishes it. This way we can add more tests to it (there) according to the things we want to this in tracee (here).
> 2. I have also played a bit with reorganizing Makefile and adding some flags to the test targets (https://gist.github.com/rafaeldtinoco/468998d513c9a61426d239527c6b4287) and got the following **several** data races for `integration_test_.go`:
> 
> image
> 
> I believe that is because of parallel calls of `startTracee` for the multiple testCases.

1. Is it in the scope of this PR or just a comment? Because we don't really use tracee-tester here, but it could be an alternative approach
2. Yeah the makefile command I had to tailor specifically, and I also noticed at some point that parallelism was causing issues, which makes sense since a tracee instance would "steal" events from another (this was theoretically a problem also with the previous version, which could make tests flaky). That's why I added `-p 1`, and why testing with `-race` breaks it.

rafaeldtinoco:
> * Is it in the scope of this PR or just a comment? Because we don't really use tracee-tester here, but it could be an alternative approach

I thought about saying it because of the "tests" directory. Didnt mean for this PR, sorry for not mentioning it.

> * Yeah the makefile command I had to tailor specifically, and I also noticed at some point that parallelism was causing issues, which makes sense since a tracee instance would "steal" events from another (this was theoretically a problem also with the previous version, which could make tests flaky). That's why I added `-p 1`, and why testing with `-race` breaks it.

Hum, ok. If -p1 fixes that I'm fine then =). Would you mind applying: https://gist.github.com/rafaeldtinoco/468998d513c9a61426d239527c6b4287 (without the race for this test) then ? Quick code replacement only.

rafaeldtinoco (approved):
LGTM

PR #2027, opened 2022-07-28
Revert container started

rafaeldtinoco:


rafaeldtinoco:
@yanivagman @NDStrahilevitz I think this will revert tests to a better shape again. Then we can try to understand what is happening.

rafaeldtinoco:
Im closing this because I might revert more than 1 commit altogether. I want to do more testing first.

Issue #2019, opened 2022-07-26
[REFAC] Change way to invoke events to uprobe and save context of trigger event


AsafEitani:
## Prerequisites

Select one OR another:

- [ ] I have discussed the refactoring idea with one (or another) maintainer.
- [x] I'll create a PR to implement this refactoring idea (assign to yourself).
- [ ] Someone else should implement this (describe it well).

## Refactoring description

This refactoring is about changing the way tracee-ebpf triggers eBPF programs (for events like `hooked_syscalls`).
It's also about saving the context of a triggering event to the trigger event so it will be within the execution context (kernel module loading triggers `hooked_syscalls` so we want to know which program caused the `hooked_syscalls`)

The refactoring is needed because we want to retain the context of events, and not to rely on ioctl syscall to trigger events so we can control the arguments.

PR #2031, opened 2022-07-30
convert trigger events from ioctl syscall to a uprobe

AsafEitani:
## Initial Checklist

## Description (git log)

commit 17b4c94551824780e6213d7bf7eacc82506324f6 (HEAD -> uprobe, origin/uprobe)
Date:   Sat Jul 30 15:57:10 2022 +0300

    events: add context of triggering event to the triggered event
    
    Retains the event context of the triggering event (e.g. hooked_syscall gets the context of the insmod that preceded it)

commit c62d19cfb0055ea1adac9abf4fdf4bb9e2fe3fdb
Date:   Mon Jun 6 16:00:37 2022 +0300

    events: convert trigger events to uprobe
    
    Change the way tracee-ebpf is triggering events.
    Solves #2019

Fixes: #2019

## Type of change

## How Has This Been Tested?

Tested both on x86 and ARM.

`tracee-ebpf -t e=hooked_syscalls,hooked_seq_ops` and load a rootkit like diamorphine that hooks syscalls or seq ops.

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -1723,10 +1723,9 @@ static __always_inline int save_str_to_buf(event_data_t *data, void *ptr, u8 ind
 }
 
 static __always_inline int
-add_u64_elements_to_buf(event_data_t *data, const u64 __user *ptr, int len, u32 count_off)
+add_u64_elements_to_buf(event_data_t *data, const u64 __user *ptr, int len, volatile u32 count_off)
 {
     u8 elem_num = 0;
-    u8 current_elem_num;
 #pragma unroll
     for (int i = 0; i < len; i++) {
         void *addr = &(data->submit_p->buf[data->buf_off]);
@@ -1740,7 +1739,11 @@ add_u64_elements_to_buf(event_data_t *data, const u64 __user *ptr, int len, u32 
     }
 out:
     // save number of elements in the array
-    current_elem_num = data->submit_p->buf[count_off & (MAX_PERCPU_BUFSIZE - 1)];
+    if (count_off > (MAX_PERCPU_BUFSIZE - 1))
+        return 0;
+    if (count_off < 0)
+        return 0;
+    u8 current_elem_num = data->submit_p->buf[count_off & (MAX_PERCPU_BUFSIZE - 1)];
     data->submit_p->buf[count_off & (MAX_PERCPU_BUFSIZE - 1)] = current_elem_num + elem_num;
 
     return 1;
@@ -1755,7 +1758,7 @@ save_u64_arr_to_buf(event_data_t *data, const u64 __user *ptr, int len, u8 index
 
     // Save space for number of elements (1 byte)
     data->buf_off += 1;
-    u32 orig_off = data->buf_off;
+    volatile u32 orig_off = data->buf_off;
     data->submit_p->buf[(data->buf_off) & (MAX_PERCPU_BUFSIZE - 1)] = 0;
     data->buf_off += 1;
     data->context.argnum++;
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -1727,7 +1727,7 @@ add_u64_elements_to_buf(event_data_t *data, const u64 __user *ptr, int len, u32 
 {
     u8 elem_num = 0;
     u8 current_elem_num;
-    #pragma unroll
+#pragma unroll
     for (int i = 0; i < len; i++) {
         void *addr = &(data->submit_p->buf[data->buf_off]);
         if (data->buf_off > MAX_PERCPU_BUFSIZE - sizeof(u64))
@@ -3148,44 +3148,11 @@ int uprobe_syscall_trigger(struct pt_regs *ctx)
     return events_perf_submit(&data, PRINT_SYSCALL_TABLE, 0);
 }
 
-static __always_inline void invoke_fetch_network_seq_operations_event(struct pt_regs *ctx,
-                                                                      unsigned long struct_address)
-{
-    event_data_t data = {};
-    if (!init_event_data(&data, ctx))
-        return;
-
-    struct seq_operations *seq_ops = (struct seq_operations *) struct_address;
-    u64 show_addr = (u64) READ_KERN(seq_ops->show);
-    if (show_addr == 0) {
-        return;
-    }
-
-    u64 start_addr = (u64) READ_KERN(seq_ops->start);
-    if (start_addr == 0) {
-        return;
-    }
-
-    u64 next_addr = (u64) READ_KERN(seq_ops->next);
-    if (next_addr == 0) {
-        return;
-    }
-
-    u64 stop_addr = (u64) READ_KERN(seq_ops->stop);
-    if (stop_addr == 0) {
-        return;
-    }
-    u64 seq_ops_addresses[NET_SEQ_OPS_SIZE + 1] = {
-        (u64) seq_ops, show_addr, start_addr, next_addr, stop_addr};
-    save_u64_arr_to_buf(&data, (const u64 *) seq_ops_addresses, 5, 0);
-    events_perf_submit(&data, PRINT_NET_SEQ_OPS, 0);
-}
-
 SEC("uprobe/trigger_seq_ops_event")
 int uprobe_seq_ops_trigger(struct pt_regs *ctx)
 {
 #if defined(bpf_target_x86)
-    uint64_t *address_array = ((void *) ctx->sp);
+    uint64_t *address_array = ((void *) ctx->sp + 8);
 #elif defined(bpf_target_arm64)
     uint64_t *address_array = ((void *) ctx->sp) + 16;
 #else
@@ -3193,10 +3160,40 @@ int uprobe_seq_ops_trigger(struct pt_regs *ctx)
 #endif
 
     uint64_t struct_address;
+    event_data_t data = {};
+    if (!init_event_data(&data, ctx))
+        return 0;
+    u32 count_off = data.buf_off + 1;
+    // Init u64 arr with size 0 before adding elements in loop
+    save_u64_arr_to_buf(&data, NULL, 0, 0);
+#pragma unroll
     for (int i = 0; i < NET_SEQ_OPS_TYPES; i++) {
         bpf_probe_read(&struct_address, 8, (address_array + i));
-        invoke_fetch_network_seq_operations_event(ctx, struct_address);
+        struct seq_operations *seq_ops = (struct seq_operations *) struct_address;
+        u64 show_addr = (u64) READ_KERN(seq_ops->show);
+        if (show_addr == 0) {
+            return 0;
+        }
+
+        u64 start_addr = (u64) READ_KERN(seq_ops->start);
+        if (start_addr == 0) {
+            return 0;
+        }
+
+        u64 next_addr = (u64) READ_KERN(seq_ops->next);
+        if (next_addr == 0) {
+            return 0;
+        }
+
+        u64 stop_addr = (u64) READ_KERN(seq_ops->stop);
+        if (stop_addr == 0) {
+            return 0;
+        }
+        u64 seq_ops_addresses[NET_SEQ_OPS_SIZE + 1] = {show_addr, start_addr, next_addr, stop_addr};
+
+        add_u64_elements_to_buf(&data, (const u64 *) seq_ops_addresses, 4, count_off);
     }
+    events_perf_submit(&data, PRINT_NET_SEQ_OPS, 0);
     return 0;
 }
 
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -83,6 +83,7 @@
 #define FILE_MAGIC_HDR_SIZE 32        // magic_write: bytes to save from a file's header
 #define FILE_MAGIC_MASK     31        // magic_write: mask used for verifier boundaries
 #define NET_SEQ_OPS_SIZE    4         // print_net_seq_ops: struct size
+#define NET_SEQ_OPS_TYPES   6         // print_net_seq_ops: argument size
 #define MAX_KSYM_NAME_SIZE  64
 
 enum buf_idx_e
@@ -339,8 +340,6 @@ enum container_state_e
     #define MAX_BIN_CHUNKS        110
 #endif
 
-#define IOCTL_FETCH_SYSCALLS            (1 << 0) // bit wise flags
-#define IOCTL_HOOKED_SEQ_OPS            (1 << 1)
 #define NUMBER_OF_SYSCALLS_TO_CHECK_X86 18
 #define NUMBER_OF_SYSCALLS_TO_CHECK_ARM 14
 
@@ -2399,36 +2398,6 @@ static __always_inline int get_local_net_id_from_network_details_v6(struct sock 
     return 0;
 }
 
-static __always_inline void invoke_fetch_network_seq_operations_event(event_data_t *data,
-                                                                      unsigned long struct_address)
-{
-    struct seq_operations *seq_ops = (struct seq_operations *) struct_address;
-
-    u64 show_addr = (u64) READ_KERN(seq_ops->show);
-    if (show_addr == 0) {
-        return;
-    }
-
-    u64 start_addr = (u64) READ_KERN(seq_ops->start);
-    if (start_addr == 0) {
-        return;
-    }
-
-    u64 next_addr = (u64) READ_KERN(seq_ops->next);
-    if (next_addr == 0) {
-        return;
-    }
-
-    u64 stop_addr = (u64) READ_KERN(seq_ops->stop);
-    if (stop_addr == 0) {
-        return;
-    }
-    u64 seq_ops_addresses[NET_SEQ_OPS_SIZE + 1] = {
-        (u64) seq_ops, show_addr, start_addr, next_addr, stop_addr};
-    save_u64_arr_to_buf(data, (const u64 *) seq_ops_addresses, 5, 0);
-    events_perf_submit(data, PRINT_NET_SEQ_OPS, 0);
-}
-
 static __always_inline int save_sockaddr_to_buf(event_data_t *data, struct socket *sock, u8 index)
 {
     struct sock *sk = get_socket_sock(sock);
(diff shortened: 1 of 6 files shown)


pkg/ebpf/events_processor.go
```diff
@@ -363,6 +364,8 @@ func (t *Tracee) processEvent(event *trace.Event) error {
 	// in case FinitModule and InitModule occurs it means that a kernel module was loaded
 	// and we will want to check if it hooked the syscall table and seq_ops
 	case events.DoInitModule:
+		contextID := uint64(len(derive.InvokedContext) + 1)
+		derive.InvokedContext[contextID] = *event
```
NDStrahilevitz, line 368: This map is shared between two separate goroutines, it needs a mutex.

pkg/ebpf/tracee.go
```diff
@@ -913,6 +912,15 @@ func (t *Tracee) initBPF() error {
 		return err
 	}
 
+	err = t.attachSelfUprobe("uprobe_syscall_trigger", "github.com/aquasecurity/tracee/pkg/ebpf.(*Tracee).invokeTriggeredSyscalls")
```
NDStrahilevitz, line 915: These probes should be integrated into the probes package.

pkg/ebpf/tracee.go
```diff
@@ -913,6 +912,15 @@ func (t *Tracee) initBPF() error {
...
+	if err != nil {
+		return err
+	}
+	err = t.attachSelfUprobe("uprobe_seq_ops_trigger", "github.com/aquasecurity/tracee/pkg/ebpf.(*Tracee).invokeTriggeredSeqOpsCall")
```
NDStrahilevitz, line 919: ditto

pkg/ebpf/tracee.go
```diff
@@ -1165,3 +1155,27 @@ func (t *Tracee) updateKallsyms() error {
 	t.kernelSymbols = kernelSymbols
 	return nil
 }
+
+// attachSelfUprobe attaches a Uprobe to itself to invoke an eBPF program from userspace
+func (t *Tracee) attachSelfUprobe(progFn string, symbolName string) error {
```
NDStrahilevitz, line 1160: This logic can be used in probes.go as part of integration.

pkg/events/derive/derive.go
```diff
@@ -10,6 +11,9 @@ import (
 // Use as static variable for testability reasons
 var getEventDefFunc = events.Definitions.Get
 
+// InvokedContext is a map used to store the triggering events' context
+var InvokedContext = make(map[uint64]trace.Event)
```
NDStrahilevitz, line 15: ditto on the mutex.
I'm also not sure if this is the best place to have the map, Maybe it should get it's own pacakge like most other caches, and then it can be managed there, with memory safety and all. It can be given as a dependency to the deriving functions.

NDStrahilevitz (changes requested):
I like this for being cleaner than the previous solution, but:
1. Uprobes aren't integrated into the existing probes code.
2. The invoking map, imo, doesn't belong in the deriving code, I believe this is why a mutex was also more easily missed.

rafaeldtinoco:
Please don't forget to fully update our PR template. I self-requested review because I want to follow the uprobe work if possible! Thanks.

pkg/ebpf/tracee.go
```diff
@@ -1117,40 +1121,38 @@ var netSeqOps = [6]string{
...
-			}
-			ptmx.Close()
+func (t *Tracee) invokeTriggeredSeqOps(contextMapID uint64) {
+	_, ok := t.events[events.HookedSeqOps]
+	if !ok {
+		return
+	}
+	var seqOpsPointers [len(netSeqOps)]uint64
+	for i, seq_name := range netSeqOps {
+		seqOpsStruct, err := t.kernelSymbols.GetSymbolByName("system", seq_name)
```
rafaeldtinoco, line 1144: Ingenious =D

pkg/ebpf/tracee.go
rafaeldtinoco: https://github.com/rafaeldtinoco/tracee/commit/19f112bdc5d6f7694a61025980edeebcfcbd2fba

Since I recently touched that code, I created the needed change if you want to pick that commit and merge into your PR and push it.

pkg/ebpf/tracee.go
rafaeldtinoco: https://github.com/aquasecurity/tracee/pull/2031#discussion_r936178427

rafaeldtinoco:
commit 23bbc9b2 needs "rewording" (git rebase -i -> reword) for line breaks.

pkg/events/derive/derive.go
```diff
@@ -45,6 +49,29 @@ func singleEventDeriveFunc(id events.ID, deriveArgsFunc deriveArgsFunction) even
...
+		}
+		if args == nil {
+			return []trace.Event{}, nil
+		}
+		de, err := newTriggeredEvent(&event, skeleton, args)
+		if err != nil {
+			return []trace.Event{}, []error{err}
+		}
+		return []trace.Event{de}, nil
+	}
```
rafaeldtinoco, line 72: Shouldn't we just add an argument to this function (isTriggered true/false) instead of duplicating the previous function ?

pkg/events/derive/derive.go
```diff
@@ -64,6 +91,28 @@ func newEvent(baseEvent *trace.Event, skeleton eventSkeleton, argsValues []inter
 	return de, nil
 }
 
+// newTriggeredEvent create a new derived event from stored triggering event , adjusted by the derived event skeleton meta-data.
+// This method enables using the context of the triggering event, but with the new arguments and meta-data of the derived one.
+func newTriggeredEvent(baseEvent *trace.Event, skeleton eventSkeleton, argsValues []interface{}) (trace.Event, error) {
```
rafaeldtinoco, line 96: Same for this one.

pkg/events/derive/derive.go
NDStrahilevitz: I think it would make more sense to do something similar to the `deriveArgsFunction` and pass an `eventBuilder` function. Then we reduce to a single function receiving both (probably the existing `singleEventDeriveFunc`).

pkg/events/derive/derive.go
```diff
@@ -72,3 +82,25 @@ func makeEventSkeleton(eventID events.ID) eventSkeleton {
 		Params: def.Params,
 	}
 }
+
+func getInvokingEventContext(event *trace.Event) (trace.Event, error) {
```
NDStrahilevitz, line 86: Maybe something simpler like `withInvokingContext` similar to how golang contexts do it.

pkg/events/derive/derive.go
```diff
@@ -72,3 +82,25 @@ func makeEventSkeleton(eventID events.ID) eventSkeleton {
...
+		if !ok {
+			return trace.Event{}, fmt.Errorf("error parsing caller_context_id arg: %v", err)
+		}
+		return contextEvent, nil
+	} else {
+		return *event, nil
+	}
+}
+
+func BaseEvent(event *trace.Event) (trace.Event, error) {
```
NDStrahilevitz, line 104: This can be left unexported I believe so `baseEvent`.

pkg/events/derive/derive.go
```diff
@@ -10,6 +12,10 @@ import (
 // Use as static variable for testability reasons
 var getEventDefFunc = events.Definitions.Get
 
+// InvokedContext is a map used to store the triggering events' context
+var InvokedContext = make(map[uint64]trace.Event)
```
NDStrahilevitz, line 16: I think we shouldn't leave this as a global variable, and now we have a global mutex too.
It could very simply be fit into a struct like the `procinfo` cache.

NDStrahilevitz (changes requested):

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3172,34 +3147,61 @@ static __always_inline void invoke_print_syscall_table_event(event_data_t *data)
...
+    events_perf_submit(&data, PRINT_NET_SEQ_OPS, 0);
+}
+
+SEC("uprobe/trigger_seq_ops_event")
+int uprobe_seq_ops_trigger(struct pt_regs *ctx)
+{
+#if defined(__TARGET_ARCH_x86)
+    uint64_t *address_array = ((void*)ctx->sp);
+#else
+    uint64_t *address_array = ((void*)ctx->sp)+16;
```
yanivagman: Where does this 16 comes from? Is it something specific to ARM64?
If so, better if we add if defined(bpf_target_arm64)

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -83,6 +83,7 @@
 #define FILE_MAGIC_HDR_SIZE 32        // magic_write: bytes to save from a file's header
 #define FILE_MAGIC_MASK     31        // magic_write: mask used for verifier boundaries
 #define NET_SEQ_OPS_SIZE    4         // print_net_seq_ops: struct size
+#define NET_SEQ_OPS_TYPES   6         // print_net_seq_ops: argument size
```
yanivagman, line 86: This can be passed as an argument to the ioctl, which also has the advantage of removing the need to sync it with userspace struct size.

yanivagman (dismissed):
That's a good change. Need to check that it works well when tracee is deployed inside a container as well, so we won't have surprises on release.

pkg/ebpf/events_processor.go
```diff
@@ -371,10 +371,8 @@ func (t *Tracee) processEvent(event *trace.Event) error {
 			if err != nil {
 				return err
 			}
-			err = t.invokeIoctlTriggeredEvents(IoctlFetchSyscalls | IoctlHookedSeqOps)
-			if err != nil {
-				return err
-			}
+			t.invokeTriggeredSyscalls()
```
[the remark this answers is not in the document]
rafaeldtinoco: Yes, I thought the same in previous review, forgot to mention.

pkg/ebpf/events_processor.go
```diff
@@ -371,10 +371,8 @@ func (t *Tracee) processEvent(event *trace.Event) error {
...
+			t.invokeTriggeredSeqOps()
```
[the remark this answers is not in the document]
rafaeldtinoco: Also, all the SeqOps should contain "NetSeqOps" as there are multiple seqops structures in kernel (if you agree).

rafaeldtinoco (commented):
LGTM after Yaniv's points are addressed.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3130,16 +3099,21 @@ int BPF_KPROBE(trace_do_exit)
...
-static __always_inline void invoke_print_syscall_table_event(event_data_t *data)
+SEC("uprobe/trigger_syscall_event")
+int uprobe_syscall_trigger(struct pt_regs *ctx)
 {
+    event_data_t data = {};
+    if (!init_event_data(&data, ctx))
+        return 0;
+
     int key = 0;
     u64 *table_ptr = bpf_map_lookup_elem(&syscalls_to_check_map, (void *) &key);
```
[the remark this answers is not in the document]
AsafEitani: Can we do this in a different PR? this will take some testing

pkg/ebpf/c/tracee.bpf.c
AsafEitani: Can we do this in a different PR? this will take some testing

pkg/ebpf/probes/probes.go
```diff
@@ -266,6 +274,73 @@ func (p *traceProbe) autoload(module *bpf.Module, autoload bool) error {
 	return enableDisableAutoload(module, p.programName, autoload)
 }
 
+//
+// uProbe
+//
+
+type uProbe struct {
```
NDStrahilevitz, line 281: Idea for the future, I think we can have this struct have a `Trigger` method of it's own, then instead of manually building the `symbolName` we can also generate it (of course matching to an existing bpf uprobe code).

This could (possibly) also help with a more concrete method of event invocation, an invoked event from a trigger would have its context stored "by definition" through this `Trigger`.

rafaeldtinoco: I would wait for more uprobe usage needs (specially those unrelated to tracee binary itself, like these ones) and then define something else if needed.

pkg/ebpf/tracee.go
```diff
@@ -1124,39 +1119,38 @@ var netSeqOps = [6]string{
...
-		_, ok := t.events[events.HookedSyscalls]
-		if ok {
-			ptmx, err := os.OpenFile(t.config.Capture.OutputPath, os.O_RDONLY, 0444)
-			if err != nil {
-				return err
-			}
-			syscall.Syscall(syscall.SYS_IOCTL, ptmx.Fd(), uintptr(IoctlFetchSyscalls), 0)
-			ptmx.Close()
-		}
+func (t *Tracee) triggerSyscallsIntegrityCheck() {
```
rafaeldtinoco, line 1112: Alright, let's be honest here.. the previous function `invokeIoctlTriggeredEvents()` already existed as a function for `*Tracee` so I'm willing to accept/agree having these functions here for now, but this should be **temporary** as the functions are unrelated to `*Tracee` and related to specific events.

If we compare all other `(*Tracee).Functions` we have "initBPF", "getOptionsConfig", "generateInitValues", ... and they're all generic to functioning of Tracee (not to specific events), or they should be.

We should probably create an interface for anything relying in similar approach(es) (anything that needs a trigger + derived events + re-trigger, based in uprobes and things like that). @NDStrahilevitz for the record here (as he is our refactoring guy =D).

I'm willing to move forward as we're basically replacing the ioctl() mechanism for uprobe, but I think it needs refactor in the next version (and an issue opened).

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: Mind opening an issue and putting a TODO in the code so we don't forget ?

pkg/ebpf/c/tracee.bpf.c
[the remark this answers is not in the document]
rafaeldtinoco: Mind opening an issue and putting a TODO in the code so we don't forget ?

rafaeldtinoco (commented):
@AsafEitani

I believe we're half way through for the "final solution", probably with a "functional enough" code for this release. I'm willing to merge (for 0.8.1) as I know its important for now... but there are some issues to be opened (and TODOs to be added to the code) and addressed for 0.8.2 (related to uprobe arguments and the uprobe symbol/functions placed somewhere else). 

I'll give a second look at the 5.4 verifier changes patches before merging, and I would like to ask @NDStrahilevitz your input here as well, so we can merge this asap (9/9/22 ?) for the 0.8.1 release.

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: Yep, is this a calling convention for the uprobe eBPF programs contexts in both arches ? Did you find it by increasing the offset or is it documented somewhere ? A comment here would be nice as well.

pkg/ebpf/tracee.go
NDStrahilevitz: Yep, please add a TODO to move away from this.

pkg/ebpf/tracee.go
```diff
@@ -1108,55 +1109,38 @@ func (t *Tracee) getCapturedIfaceIdx(ifaceName string) (int, bool) {
...
+			continue
 		}
+		seqOpsPointers[i] = seqOpsStruct.Address
 	}
+	t.triggerSeqOpsIntegrityCheckCall(seqOpsPointers)
+}
+
+// triggerSeqOpsIntegrityCheck is used by a Uprobe to trigger an eBPF program that prints the seq ops pointers
+//go:noinline
+func (t *Tracee) triggerSeqOpsIntegrityCheckCall(seqOpsStruct [len(derive.NetSeqOps)]uint64) error {
```
NDStrahilevitz, line 1143: Main challenge for moving the triggers might be defining the "probe arguments" more generically, but we'll tackle that when refactoring this.

pkg/events/derive/hooked_seq_ops.go
```diff
@@ -19,18 +37,18 @@ func deriveHookedSeqOpsArgs(kernelSymbols *helpers.KernelSymbolTable) deriveArgs
...
+		hookedSeqOps := make(map[string]trace.HookedSymbolData, 0)
+		for i, addr := range seqOpsArr {
 			inTextSegment, err := kernelSymbols.TextSegmentContains(addr)
 			if err != nil {
 				continue
 			}
 			if !inTextSegment {
 				hookingFunction := utils.ParseSymbol(addr, kernelSymbols)
-				hookedSeqOps = append(hookedSeqOps, trace.HookedSymbolData{SymbolName: hookingFunction.Name, ModuleOwner: hookingFunction.Owner})
+				hookedSeqOps[NetSeqOps[i/4]+"_"+NetSeqOpsFuncs[i%4]] =
```
NDStrahilevitz, line 48: Could you make this indexing more concise/human-readable (maybe a "translation" function)

NDStrahilevitz (commented):
So, for this part of this change "package" (including #2049) i'm basically ok with this one, my main gripe overall is the way we retain context, which i've addressed over there.

So for this one I think we're good on merging for now, just add the comments and TODOs requested and open an issue for refactoring appropriately in 0.8.2, mainly on where and how we trigger the probes, this could be solved overall in the context of an "event invocation" package, similar to how we have a package for event derivation.

pkg/ebpf/c/tracee.bpf.c
AsafEitani: So it is documented but I had to do some testing to figure which argument goes where.
This is the documentation:
https://go.googlesource.com/go/+/refs/heads/dev.regabi/src/cmd/compile/internal-abi.md

I'll add a comment

pkg/ebpf/c/tracee.bpf.c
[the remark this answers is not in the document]
AsafEitani: #2055 
Will add a TODO in the code

pkg/ebpf/tracee.go
AsafEitani: Sure I agree, I'll add a TODO and add that to #2055

pkg/ebpf/tracee.go
rafaeldtinoco: https://github.com/aquasecurity/tracee/issues/2055

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: Thanks for the pointers for the golang calling convention (as a comment).

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3130,16 +3098,23 @@ int BPF_KPROBE(trace_do_exit)
...
 }
 
-/* invoke_print_syscall_table_event submit to the buff the syscalls function handlers address from
- * the syscall table. the syscalls are strode in map which is syscalls_to_check_map and the
+/* uprobe_syscall_trigger submit to the buff the syscalls function handlers address from
+ * the syscall table. the syscalls are stored in map which is syscalls_to_check_map and the
  * syscall-table address is stored in the kernel_symbols map.
  */
-static __always_inline void invoke_print_syscall_table_event(event_data_t *data)
+SEC("uprobe/trigger_syscall_event")
```
rafaeldtinoco, line 3105: I left a suggestion for changes at: https://github.com/aquasecurity/tracee/pull/2049#discussion_r941866347 after this was merged.

Issue #1815, opened 2022-06-08
[BUG] Use elevated capabilities in new kernels


AlonZivony:
## Prerequisites

- [ ] This affects latest released version.
- [ ] This affects current development tree (origin/HEAD).
- [x] There isn't an issue describing the bug.

Select one OR another:

- [ ] I'm going to create a PR to solve this (assign to yourself).
- [x] Someone else should solve this.

## Bug description

To mitigate the #1803 bug, new kernels capabilities changed in tracee-ebpf to include `CAP_SYS_ADMIN` instead of `CAP_BPF`+`CAP_PERFMON`.
This mitigation resulted that tracee-ebpf in new kernels have much higher capabilities than required, which make it much more exposed to security issues.
To remove the `CAP_SYS_ADMIN` capability, we need a solution so kprobes could be created without any errors.

yanivagman:
More info from @AlonZivony mentioned in #747 :
"The issue here was that some distros didn't work without CAP_SYS_ADMIN event though it should work with their kernel version.
To do this bugfix, the reason for the error in https://github.com/aquasecurity/tracee/issues/1803 should be resolved.
I don't have the time for it right now, so if @cdelzotti want to solve it, it would be great :)"

cdelzotti:
Can I have indications about which distro doesn't work without CAP_SYS_ADMIN in order to reproduce the bug ?

AlonZivony:
From the tests it seems that the `jammy` and `focal` releases didn't work with the new capabilities.
You can try to look more into the tests that failed.

rafaeldtinoco:
Please be aware that, if because of a kernel bug, those distributions aren't able to cope with correct capabilities, then it is what is is. We have to make sure we work in all supported environments with LATEST kernel/env updates (and, hopefully, try to mitigate released environments. This is exactly what BTFHUB is about, for example, as latest kernels provide BTF information, but we wanted to maximize our chances of working in all envs).

cdelzotti:
Okay, I recompiled the kernel on a default Ubuntu Focal with the same kernel version and the bug disappeared. I deducted it was related to Ubuntu itself and bingo : https://patchwork.ozlabs.org/project/ubuntu-kernel/patch/20201126205206.2463398-2-cascardo@canonical.com/

[...]

PR #2033, opened 2022-07-31
Add `perf_event_paranoid` dependent capability support

cdelzotti:
## Initial Checklist

## Description (git log)

Capabilities required by Tracee depend on the `perf_event_paranoid` kernel value.
Before this commit, Tracee required the `CAP_SYS_ADMIN` capability to be certain
to run on every configuration. This commit checks the `perf_event_paranoid` to ensure that
Tracee is running with required capabilities only, therefore reducing the exposition to security
issues.

Fixes: #1815

## Type of change

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/cmd/tracee-ebpf/capabilities.go
+++ b/cmd/tracee-ebpf/capabilities.go
@@ -4,6 +4,8 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strconv"
+	"strings"
 
 	"github.com/aquasecurity/tracee/pkg/capabilities"
 	tracee "github.com/aquasecurity/tracee/pkg/ebpf"
@@ -50,7 +52,7 @@ func ensureCapabilities(OSInfo KernelVersionInfo, cfg *tracee.Config, allowHighC
 // Get all capabilities required to run tracee-ebpf for current run
 func generateTraceeEbpfRequiredCapabilities(OSInfo KernelVersionInfo, cfg *tracee.Config) (
 	[]cap.Value, error) {
-	rCaps, err := getCapabilitiesRequiredByEBPF(OSInfo)
+	rCaps, err := getCapabilitiesRequiredByEBPF(OSInfo, cfg.Debug)
 	if err != nil {
 		return nil, err
 	}
@@ -73,18 +75,69 @@ func getCapabilitiesRequiredByTraceeEvents(cfg *tracee.Config) []cap.Value {
 	return removeDupCaps(caps)
 }
 
-// Get all capabilities required for eBPF usage (including perf buffers maps management)
-func getCapabilitiesRequiredByEBPF(OSInfo KernelVersionInfo) ([]cap.Value, error) {
-	// In kernel 5.8, CAP_BPF and CAP_PERFMON capabilities were introduced in order to replace CAP_SYS_ADMIN when
-	// loading eBPF programs.
-	// For some reasons, some distributions using new kernels still need CAP_SYS_ADMIN,
-	// so tracee still use it instead of the new capabilities.
-	caps := []cap.Value{
+// Retrieve the value of the kernel parameter perf_event_paranoid
+func getKernelPerfEventParanoidValue() (int, error) {
+	/*
+	* perf event paranoia level:
+	*  -1 - not paranoid at all
+	*   0 - disallow raw tracepoint access for unpriv
+	*   1 - disallow cpu events for unpriv
+	*   2 - disallow kernel profiling for unpriv
+	*   4 - disallow all unpriv perf event use (Only on some particular distributions)
+	 */
+	const MaxParanoiaLevel = 4
+	value, err := os.ReadFile("/proc/sys/kernel/perf_event_paranoid")
+	if err != nil {
+		return MaxParanoiaLevel, errors.New("cannot read kernel paranoia")
+	}
+	intVal, err := strconv.ParseInt(strings.TrimSuffix(string(value), "\n"), 0, 16)
+	if err != nil {
+		return MaxParanoiaLevel, errors.New("cannot handle kernel paranoia")
+	}
+	return int(intVal), nil
+}
+
+// getCapabilitiesRequiredByEBPF gets all capabilities required for eBPF usage (including perf buffers maps management)
+func getCapabilitiesRequiredByEBPF(OSInfo KernelVersionInfo, debug bool) ([]cap.Value, error) {
+	// In kernel 5.8, CAP_BPF and CAP_PERFMON capabilities were introduced in
+	// order to replace CAP_SYS_ADMIN when loading eBPF programs.
+	privilegedCaps := []cap.Value{
 		cap.IPC_LOCK,
 		cap.SYS_RESOURCE,
 		cap.SYS_ADMIN,
 	}
-	return caps, nil
+	unprivilegedCaps := []cap.Value{
+		cap.IPC_LOCK,
+		cap.SYS_RESOURCE,
+		cap.BPF,
+		cap.PERFMON,
+	}
+	kernelParanoidValue, err := getKernelPerfEventParanoidValue()
+	if err != nil {
+		if debug {
+			fmt.Println("Paranoid: could not read /proc/sys/kernel/perf_event_paranoid: ", err.Error())
+		}
+		return privilegedCaps, nil
+	}
+	const BpfCapabilitiesMinKernelVersion = "5.8"
+	if OSInfo.CompareOSBaseKernelRelease(BpfCapabilitiesMinKernelVersion) < 0 {
+		// if kernelParanoidValue is too high, CAP_SYS_ADMIN is required
+		if kernelParanoidValue > 3 {
+			if debug {
+				fmt.Println("Paranoid: Value in /proc/sys/kernel/perf_event_paranoid is > 3")
+				fmt.Println("Paranoid: Tracee needs CAP_SYS_ADMIN instead of CAP_BPF + CAP_PERFMON")
+				fmt.Println("Paranoid: To change that behavior set perf_event_paranoid to 3 or less.")
+			}
+			return privilegedCaps, nil
+		} else {
+			err := capabilities.CheckRequired(unprivilegedCaps)
+			if err != nil {
+				return privilegedCaps, nil
+			}
+			return unprivilegedCaps, nil
+		}
+	}
+	return privilegedCaps, nil
 }
 
 func removeDupCaps(dupCaps []cap.Value) []cap.Value {
--- a/docs/installing/prerequisites.md
+++ b/docs/installing/prerequisites.md
@@ -28,8 +28,8 @@ capabilities:
 
 * Manage eBPF maps limits (`CAP_SYS_RESOURCE`)
 * Load and Attach eBPF programs:
-    1. `CAP_BPF`+`CAP_PERFMON` for recent kernels (>=5.8)
-    2. or `CAP_SYS_ADMIN` for older kernels
+    1. `CAP_BPF`+`CAP_PERFMON` for recent kernels (>=5.8) where the kernel perf paranoid value in `/proc/sys/kernel/perf_event_paranoid` is equal to 3 or less
+    2. or `CAP_SYS_ADMIN` otherwise
 * `CAP_SYS_PTRACE` (to collect information about processes upon startup)
 * `CAP_NET_ADMIN` (to use tc for packets capture)
 * `CAP_SETPCAP` (if given - used to reduce bounding set capabilities)


cmd/tracee-ebpf/capabilities.go
```diff
@@ -73,16 +74,47 @@ func getCapabilitiesRequiredByTraceeEvents(cfg *tracee.Config) []cap.Value {
...
 
+// Retrieve the value of the kernel parameter perf_event_paranoid
+func getKernelPerfEventParanoidValue() int {
+	// Read the kernel parameter perf_event_paranoid
+	value, err := os.ReadFile("/proc/sys/kernel/perf_event_paranoid")
+	// If file read failed, returns the most restrictive value
+	if err != nil {
+		return 4
+	}
+	// Strip the newline character
```
rafaeldtinoco, line 85: too descriptive.

cmd/tracee-ebpf/capabilities.go
```diff
@@ -73,16 +74,47 @@ func getCapabilitiesRequiredByTraceeEvents(cfg *tracee.Config) []cap.Value {
...
+	value = value[:len(value)-1]
+	// Convert the value to an integer
```
rafaeldtinoco, line 87: ditto

cmd/tracee-ebpf/capabilities.go
```diff
@@ -73,16 +74,47 @@ func getCapabilitiesRequiredByTraceeEvents(cfg *tracee.Config) []cap.Value {
...
+	intVal, err := strconv.ParseInt(string(value), 0, 16)
+	// If conversion failed, returns the most restrictive value
```
rafaeldtinoco, line 89: ditto

cmd/tracee-ebpf/capabilities.go
```diff
@@ -73,16 +74,47 @@ func getCapabilitiesRequiredByTraceeEvents(cfg *tracee.Config) []cap.Value {
...
 func getCapabilitiesRequiredByEBPF(OSInfo KernelVersionInfo) ([]cap.Value, error) {
 	// In kernel 5.8, CAP_BPF and CAP_PERFMON capabilities were introduced in order to replace CAP_SYS_ADMIN when
 	// loading eBPF programs.
-	// For some reasons, some distributions using new kernels still need CAP_SYS_ADMIN,
-	// so tracee still use it instead of the new capabilities.
-	caps := []cap.Value{
-		cap.IPC_LOCK,
-		cap.SYS_RESOURCE,
-		cap.SYS_ADMIN,
+	kernelParanoidValue := getKernelPerfEventParanoidValue()
```
rafaeldtinoco, line 100: Hum, thats a nice catch =). I would add comment from `kernel/events/core.c` as a comment here (or in that function):

```
/*
 * perf event paranoia level:
 *  -1 - not paranoid at all
 *   0 - disallow raw tracepoint access for unpriv
 *   1 - disallow cpu events for unpriv
 *   2 - disallow kernel profiling for unpriv
 *   4 - disallow all unpriv perf event use
 */
```

or something similar.

cmd/tracee-ebpf/capabilities.go
```diff
@@ -73,16 +74,47 @@ func getCapabilitiesRequiredByTraceeEvents(cfg *tracee.Config) []cap.Value {
...
+	var caps []cap.Value
+	// if kernelParanoidValue is too high, CAP_SYS_ADMIN is required
+	if kernelParanoidValue > 1 {
+		fmt.Println("Kernel perf_event_paranoid value is greater than 1, thus Tracee needs to use the CAP_SYS_ADMIN capability instead of CAP_BPF + CAP_PERFMON.")
```
rafaeldtinoco, line 104: We can't write messages to stdout in non fatal errors I'm afraid. We should write to stderr OR, even better, to stdout if debug flag is given.

rafaeldtinoco:
@AlonZivony mind taking a look ?

rafaeldtinoco:
The errors are very likely related to the stdout message (FYIO) @cdelzotti

cmd/tracee-ebpf/capabilities.go
```diff
@@ -73,16 +74,47 @@ func getCapabilitiesRequiredByTraceeEvents(cfg *tracee.Config) []cap.Value {
...
+// Retrieve the value of the kernel parameter perf_event_paranoid
+func getKernelPerfEventParanoidValue() int {
+	// Read the kernel parameter perf_event_paranoid
+	value, err := os.ReadFile("/proc/sys/kernel/perf_event_paranoid")
+	// If file read failed, returns the most restrictive value
+	if err != nil {
+		return 4
+	}
+	// Strip the newline character
+	value = value[:len(value)-1]
```
AlonZivony, line 94:
```suggestion
	value = strings.TrimSuffix(value, '\n')
```

cmd/tracee-ebpf/capabilities.go
```diff
@@ -73,16 +74,47 @@ func getCapabilitiesRequiredByTraceeEvents(cfg *tracee.Config) []cap.Value {
...
+	// Convert the value to an integer
+	intVal, err := strconv.ParseInt(string(value), 0, 16)
+	// If conversion failed, returns the most restrictive value
+	if err != nil {
+		return 4
```
AlonZivony, line 97: I think it would be better to use consts for these values,
it will make it self-explanatory.

cmd/tracee-ebpf/capabilities.go
```diff
@@ -73,16 +74,47 @@ func getCapabilitiesRequiredByTraceeEvents(cfg *tracee.Config) []cap.Value {
...
-	// For some reasons, some distributions using new kernels still need CAP_SYS_ADMIN,
-	// so tracee still use it instead of the new capabilities.
-	caps := []cap.Value{
-		cap.IPC_LOCK,
-		cap.SYS_RESOURCE,
-		cap.SYS_ADMIN,
+	kernelParanoidValue := getKernelPerfEventParanoidValue()
+	var caps []cap.Value
+	// if kernelParanoidValue is too high, CAP_SYS_ADMIN is required
+	if kernelParanoidValue > 1 {
```
AlonZivony, line 109: This one is weird.
You wrote in the issue that you saw that values up to 3 enable the use of `CAP_BPF` and `CAP_PERFMON`.
Why did you change it to be 1?
In my system I have the value of 2, though it works for me with  `CAP_BPF` and `CAP_PERFMON`. Is this a bug?

cmd/tracee-ebpf/capabilities.go
AlonZivony: I think it should probably be printed if in debug mode.
I hope that in the near future we will have a logger to handle this logic, but you can just change it so you get the debug flag as a argument to this function.

AlonZivony (changes requested):
The main issue I see here is with the maximum level of the paranoid which enables the new capabilities.
I think @rafaeldtinoco is right with his assumption that the prints make the tests fail.

cmd/tracee-ebpf/capabilities.go
cdelzotti: I took the liberty to add the config to the function parameters to check the Debug flag, let me know if I shouldn't.

AlonZivony: Hmm I think we should only pass the flag for now, to make it less dependent and more testable.

cmd/tracee-ebpf/capabilities.go
cdelzotti: Just wanted to be safe as the documentation mentions  `2 - disallow kernel profiling for unpriv`. But it does work qith values up to 3 on my machine.

AlonZivony:
Sorry for the delayed CR, it made you do some work twice :(

cmd/tracee-ebpf/capabilities.go
AlonZivony: I think we can have the assumption that tracee run as a privileged process and user.

cdelzotti:
Okay, I think I took everything into account.

rafaeldtinoco:
@AlonZivony I'll merge this once you think is good, just let me know. Thanks!

cmd/tracee-ebpf/capabilities.go
```diff
@@ -73,16 +75,53 @@ func getCapabilitiesRequiredByTraceeEvents(cfg *tracee.Config) []cap.Value {
...
-	// For some reasons, some distributions using new kernels still need CAP_SYS_ADMIN,
-	// so tracee still use it instead of the new capabilities.
-	caps := []cap.Value{
-		cap.IPC_LOCK,
-		cap.SYS_RESOURCE,
-		cap.SYS_ADMIN,
+	kernelParanoidValue := getKernelPerfEventParanoidValue()
+	var caps []cap.Value
+	// if kernelParanoidValue is too high, CAP_SYS_ADMIN is required
+	if kernelParanoidValue > 3 {
```
AlonZivony, line 107: You miss here the check of the kernel version, which should be at least 5.8 to use `CAP_BPF` and `CAP_PERFMON` (they don't exist before this version).

Should probably look the following:
```suggestion
        const bpfCapabilitiesMinKernelVersion = "5.8"
        if OSInfo.CompareOSBaseKernelRelease(bpfCapabilitiesMinKernelVersion) <= 0 {
	    if kernelParanoidValue > 3 {
```

AlonZivony (changes requested):
Just noticed the missing check of kernel versions (because of the failed tests).
Add it and it should be good.
Notice you will need to change the tests to check all these scenarios.

cmd/tracee-ebpf/capabilities.go
AlonZivony: I will add that when I implemented it, I checked when using `CAP_BPF` and `CAP_PERFMON` if they are available.
If they aren't, I reverted to using `CAP_SYS_ADMIN`.
I think we try at all cost to make tracee work when used, so if there is a way to run it with even higher capabilities then we will use it.
You can see my implementation in commit 6b0cad48.

AlonZivony (approved):
LGTM

cmd/tracee-ebpf/capabilities.go
cdelzotti: Thank you, how can I modify the tests ?

docs/installing/prerequisites.md
```diff
@@ -28,8 +28,8 @@ capabilities:
 
 * Manage eBPF maps limits (`CAP_SYS_RESOURCE`)
 * Load and Attach eBPF programs:
-    1. `CAP_BPF`+`CAP_PERFMON` for recent kernels (>=5.8)
-    2. or `CAP_SYS_ADMIN` for older kernels
+    1. `CAP_BPF`+`CAP_PERFMON` for recent kernels (>=5.8) where the kernel perf paranoid value in `/proc/sys/kernel/perf_event_paranoid` is equal to 1 or less
```
yanivagman, line 31: Above it seems that you check if paranoid level is more than 3, not 1 like written here. Which one is true?

cdelzotti: 1, forgot to update the documentation, sorry

cmd/tracee-ebpf/capabilities.go
```diff
@@ -73,17 +75,66 @@ func getCapabilitiesRequiredByTraceeEvents(cfg *tracee.Config) []cap.Value {
...
+	const bpfCapabilitiesMinKernelVersion = "5.8"
+	if OSInfo.CompareOSBaseKernelRelease(bpfCapabilitiesMinKernelVersion) > 0 {
+		caps = privilegedCaps
+	} else {
+		// if kernelParanoidValue is too high, CAP_SYS_ADMIN is required
+		if kernelParanoidValue > 3 {
+			if debug {
+				fmt.Print("Kernel perf_event_paranoid value is greater than 3, ")
+				fmt.Print("requiring Tracee to use the CAP_SYS_ADMIN capability instead of CAP_BPF + CAP_PERFMON\n")
+				fmt.Println("To remove this message, set the kernel parameter perf_event_paranoid to 3 or less.")
```
yanivagman, line 126: So I wonder if this is actually a bug (in Ubuntu?) since the kernel documentation says that CAP_PERFMON should be enough for every operation used with perf: https://www.kernel.org/doc/html/latest/admin-guide/perf-security.html#perf-events-access-control 
Actually, the kernel sources don't even mention levels 3 or 4 as can be seen in the comment here:
https://elixir.bootlin.com/linux/v5.19/source/kernel/events/core.c#L416
So if I understand it right, it's probably an addition of Ubuntu, which is not part of mainline kernel, and they "forgot" to allow the operation with CAP_PERFMON enabled, am I right?

yanivagman (commented):
Thanks @cdelzotti 
A few comments/questions and I think we're set

cmd/tracee-ebpf/capabilities.go
rafaeldtinoco: They extracted from grsecurity patches:

image

rafaeldtinoco: I think we should assume that there is a deviation in between upstream acceptance of >=2 and grscurity/ubuntu (and possibly other distros ?) having > 2 values.

yanivagman: > They extracted from grsecurity patches:
> 
> image

From when is this commit dated? Maybe they required CAP_SYS_ADMIN before CAP_PERFMON was available (5.8), and simply forgot to add the option for this capability sinve these kernels

cmd/tracee-ebpf/capabilities.go
```diff
@@ -73,17 +75,66 @@ func getCapabilitiesRequiredByTraceeEvents(cfg *tracee.Config) []cap.Value {
...
+	*   0 - disallow raw tracepoint access for unpriv
+	*   1 - disallow cpu events for unpriv
+	*   2 - disallow kernel profiling for unpriv
+	*   4 - disallow all unpriv perf event use
+	 */
+	const MAX_PARANOIA_LEVEL = 4
+	value, err := os.ReadFile("/proc/sys/kernel/perf_event_paranoid")
+	if err != nil {
+		return MAX_PARANOIA_LEVEL
+	}
```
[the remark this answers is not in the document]
cdelzotti: Hmm, why not, this would allow the caller to print a more precise debug message I guess

cmd/tracee-ebpf/capabilities.go
cdelzotti: I think you might be right, as in this discussion (https://patchwork.kernel.org/project/linux-hardening/patch/1469630746-32279-1-git-send-email-jeffv@google.com/) they seem to wish for a `CAP_PERF` that would handle this kind of things.

cdelzotti: > I think we should assume that there is a deviation in between upstream acceptance of >=2 and grscurity/ubuntu (and possibly other distros ?) having > 2 values.

That really looks like it as I noticed that Tracee runs fine on a recompiled kernel with a paranoid value > 3 without `CAP_SYS_ADMIN`

rafaeldtinoco: Yep and I can talk to their kernel team and ask them to address that, and will, but we will have to work in many already releases broken kernels even if they fix that in latest releases I'm afraid.

cdelzotti: Yeah, so I don't really think it changes something to the way this PR addresses the `perf_event_paranoid` value problem

yanivagman (approved):
I'll let @rafaeldtinoco have another look before merging this, but LGTM, thanks @cdelzotti

rafaeldtinoco (commented):
Thanks @yanivagman,

@cdelzotti how do you feel about making the change proposed at:
https://gist.github.com/rafaeldtinoco/9d913fb93a03dd8697b9b678f199baa4

Reasons:

1. We dont have to follow `caps` value (return already says what we are returning each time.
2. Simplifies 2 "if/elses" (no need for else if we already return before.
3. Debug messages are usually prefixed with the session we're currently debugging for and are easier to follow (simpler to know what is the file we tried to read).

feel free to include this change if you agree, or parts of it, and we can merge this. could you also make sure I did not do anything wrong ? I haven't tested the change.

rafaeldtinoco:
Thanks @cdelzotti for your contribution!

Issue #2036, opened 2022-08-01
[BUG] 'symbols_loaded' event errors are too noisy


AlonZivony:
## Prerequisites

- [ ] This affects latest released version.
- [x] This affects current development tree (origin/HEAD).
- [x] There isn't an issue describing the bug.

Select one OR another:

- [x] I'm going to create a PR to solve this (assign to yourself).
- [ ] Someone else should solve this.

## Bug description

The `symbols_loaded` event have errors resulting from the loader of symbols.
These errors happens frequently in some environments and from certain loads, and passed many times for each time there is error.

Example output:
failed to derive event 758: has no access to container fs - no recorded living process of mount namespace 4026532273
failed to derive event 758: has no access to container fs - no recorded living process of mount namespace 4026532273
failed to derive event 758: has no access to container fs - no recorded living process of mount namespace 4026532273
failed to derive event 758: has no access to container fs - no recorded living process of mount namespace 4026532273
failed to derive event 758: has no access to container fs - no recorded living process of mount namespace 4026532273
failed to derive event 758: has no access to container fs - no recorded living process of mount namespace 4026532273
failed to derive event 758: has no access to container fs - no recorded living process of mount namespace 4026532273
failed to derive event 758: has no access to container fs - no recorded living process of mount namespace 4026532273

## Steps to reproduce
Run the `symbols_loaded` event in environment with container already running.

NDStrahilevitz:
This event is supposed to work with existing containers too right? Let's try to solve the root cause before just debug blocking the error, this could be a legit bug.

AlonZivony:
Currently it doesn't suppose to work with existing containers.
The reason for that is that we use the `bucketcache` of tracee named `pidsInMntns` to get to the container path.
This cache is using the `sched_process_exec` event, so it is fed only with the creation of new containers.
@yanivagman wants to change this cache, so when it will be changed it should be better.

Anyways, there are many good reasons for this event to fail.
Because it is originated from user-space, the shared object or even the entire container might not be accessible anymore when the event is triggered.
Because I don't think there is a way to make sure we get rid of these errors, I think we should silence them (or at least give them lower logging level when we have a logger).

NDStrahilevitz:
I understood the cause is the bucket cache, which, as far as i've checked, is populated when starting tracee. So it should work with old containers too, unless the populate skips them. So shouldn't the fix be pre-populating the cache correctly?

[...]

PR #2037, opened 2022-08-01
pkg/events/derive: prevent spam errors with symbols_loaded

AlonZivony:
## Initial Checklist

## Description (git log)

Symbols loader make many errors, and by returning them from the
derive function it causes too much noise.
Change it so excess errors will be visible only while debugging.

Fixes: #2036

## Type of change

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/events/derive/symbols_loaded.go
+++ b/pkg/events/derive/symbols_loaded.go
@@ -9,8 +9,8 @@ import (
 	"strings"
 )
 
-func SymbolsLoaded(soLoader sharedobjs.DynamicSymbolsLoader, watchedSymbols []string, whitelistedLibsPrefixes []string) events.DeriveFunction {
-	gen := initSymbolsLoadedEventGenerator(soLoader, watchedSymbols, whitelistedLibsPrefixes)
+func SymbolsLoaded(soLoader sharedobjs.DynamicSymbolsLoader, watchedSymbols []string, whitelistedLibsPrefixes []string, isDebug bool) events.DeriveFunction {
+	gen := initSymbolsLoadedEventGenerator(soLoader, watchedSymbols, whitelistedLibsPrefixes, isDebug)
 	return singleEventDeriveFunc(events.SymbolsLoaded, gen.deriveArgs, withOriginalContext)
 }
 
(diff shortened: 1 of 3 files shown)


rafaeldtinoco:
@AlonZivony we're releasing 0.8.1 today/tomorrow after PRs from @AsafEitani (ioctl->uprobe changes) are merged. Should I include this fix in this release? I ask you because this is a "fix" and might be needed for 0.8.1.

pkg/events/derive/symbols_loaded.go
```diff
@@ -75,7 +78,12 @@ func (symbsLoadedGen *symbolsLoadedEventGenerator) deriveArgs(event trace.Event)
 
 	soSyms, err := symbsLoadedGen.soLoader.GetExportedSymbols(loadingObjectInfo)
 	if err != nil {
-		return nil, err
+		// This error happens frequently in some environments, so we need to silence it to reduce spam.
+		if symbsLoadedGen.isDebug {
```
rafaeldtinoco, line 82: https://github.com/aquasecurity/tracee/issues/2036#issuecomment-1211524984

rafaeldtinoco (commented):
Based on the discussion at: https://github.com/aquasecurity/tracee/issues/2036#issuecomment-1211524984

I'm okay in accepting this PR for the 0.8.1 release without the initial buckets cache population. I would request that the derived event is not generated and a counter is kept until we implement a better debug printer mechanism.

NDStrahilevitz:
My concern is that the event should be visible to the users and we shouldn't try to "hide" it with the debug flag.
If we're going as far as to count error occurrences, I would suggest we cache their existence in a counter map (`map[string]int` or `map[error]int`) and simply just report them once. That way the error is reported to the user, and is not spammed.

Even further that could even be a mechanism in tracee itself and not just this derived event (possibly a `reportedErrors` map).

And either way open an issue on pre-populating this bucketcache.

AlonZivony:
>  I would suggest we cache their existence in a counter map (map[string]int or map[error]int) and simply just report them on their first X occurrences (so for each missing mount namespace we could limit the amount of prints for example).

This sounds to me like a great idea!
I think we should only pass the first error for each namespace, because returning it more than once is futile (it is not likely that tracee will cache a mount ns after it was executed, at least not in a reliable way).
I will change my code to do so.

rafaeldtinoco:
I like the idea, a similar idea occurred to me when writing the comment but I didn't want to add burden to Alon, I'm glad you both agree on that.

pkg/events/derive/symbols_loaded.go
NDStrahilevitz: Now that we have the error counting method, can we remove the debug mode check here?

rafaeldtinoco: I thought we would give the error in debug mode only (which I think it is the correct thing).

NDStrahilevitz: IIRC we report runtime errors to stderr when they occur without checking for debug mode, exception are sometimes made for initialization code errors (for example setting up the runtime enrichers, if a runtime is "missing" will emit a false positive error).
It may be better in the future that errors will only be emitted if debug mode is set (better defined through log levels), but we should normalize that in one go, not break standards in the middle. We can, and should, do that alongside the logger PR.

Since the code won't spam errors now, I think it's safe to remove the debug condition.

rafaeldtinoco: https://github.com/aquasecurity/tracee/issues/2036#issuecomment-1212033996

rafaeldtinoco (approved):
LGTM.

PR #2041, opened 2022-08-02
types: add context flags with container flag to event

AlonZivony:
## Initial Checklist

## Description (git log)

Flags will enable to pass boolean information about the context from
the kernel.
Add the container context from the kernel in the new flags.

## Type of change

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/types/trace/trace.go
+++ b/types/trace/trace.go
@@ -12,50 +12,59 @@ import (
 
 // Event is a single result of an ebpf event process. It is used as a payload later delivered to tracee-rules.
 type Event struct {
-	Timestamp           int        `json:"timestamp"`
-	ThreadStartTime     int        `json:"threadStartTime"`
-	ProcessorID         int        `json:"processorId"`
-	ProcessID           int        `json:"processId"`
-	CgroupID            uint       `json:"cgroupId"`
-	ThreadID            int        `json:"threadId"`
-	ParentProcessID     int        `json:"parentProcessId"`
-	HostProcessID       int        `json:"hostProcessId"`
-	HostThreadID        int        `json:"hostThreadId"`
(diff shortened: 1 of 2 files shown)


types/trace/trace.go
```diff
@@ -12,50 +12,59 @@ import (
...
+	if e.ContextFlags.ContainerStarted {
 		return ContainerOrigin
-	} else {
-		return HostOrigin
 	}
+	if e.ContainerID != "" {
+		return ContainerInitOrigin
+	}
+	return HostOrigin
 }
```
yanivagman, line 68: This function will break until we merge the container flags PR.
We should consider moving it out of types package.

AlonZivony: So me and @rafaeldtinoco saw that the bug wasn't from the container flag PR, but from unreliable tests.
The problem was fixed in the cb65c6ac commit.
This means we can safely merge now the 2 PRs for the container flag, so removing the `Origin` change isn't a problem.

yanivagman: Cool! What is this commit (cb65c6ac) exactly? I can't find it looking at main

NDStrahilevitz: > Cool! What is this commit (cb65c6ac) exactly? I can't find it looking at main

It's [this](https://github.com/aquasecurity/tracee/commit/cb56c6ace4cdb9b8a3f4b6aa74426deece708662).

Just to note though I think this change would further break old recordings on new `tracee-rules` versions.

yanivagman (approved):
This PR is actually the same as #2007 which was later reverted in #2027, but now discovered not to be causing the issue we thought it was (https://github.com/aquasecurity/tracee/pull/2041/files#r935748225).
So shouldn't be a problem to merge it again now.

Issue #2047, opened 2022-08-04
[BUG] 'symbols_loaded' event libraries path missing


AlonZivony:
## Prerequisites

- [x] This affects latest released version.
- [x] This affects current development tree (origin/HEAD).
- [x] There isn't an issue describing the bug.

Select one OR another:

- [x] I'm going to create a PR to solve this (assign to yourself).
- [ ] Someone else should solve this.

## Bug description

The `symbols_loaded` event uses a whitelist of libraries, using known libraries directories.
We noticed its missing multiple known directories:
* /lib/x86_64-linux-gnu
* /usr/lib/aarch64-linux-gnu/libc.so.6
* /lib/aarch64-linux-gnu/libc.so.6

From documentation, we noticed that some distros use directories matching the architecture (like the case of `aarch64` and `x86-64` I mentioned). This would apply to x86 32bit libraries belonging to the `i386` or `i686` architecture directory.
Should probably include any combination of lib directory with architecture.

## Steps to reproduce

Steps to reproduce the issue:

1. step 1
2. step 2

## Context

Relevant information about my setup:

* Linux version:
* Linux kernel version:
* Tracee version (or commit id of your tree):
* LLVM version:
* Golang version:

PR #2048, opened 2022-08-04
pkg/events/derive: fix libs whitelist of symbols_loaded

AlonZivony:
## Initial Checklist

## Description (git log)

Add whitelist for architectures libraries directories.

Fixes: #2047

## Type of change

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/events/derive/symbols_loaded.go
+++ b/pkg/events/derive/symbols_loaded.go
@@ -16,11 +16,18 @@ func SymbolsLoaded(soLoader sharedobjs.DynamicSymbolsLoader, watchedSymbols []st
 
 // Most specific paths should be at the top, to prevent bugs with iterations over the list
 var knownLibrariesDirs = []string{
-	"/usr/lib/x86_64-linux-gnu/",
 	"/usr/lib64/",
 	"/usr/lib/",
 	"/lib64/",
 	"/lib/",
+}
+
+var knownArchitectureDirs = []string{
+	"x86_64-linux-gnu",
+	"aarch64-linux-gnu",
+	"i386-linux-gnu",
+	"i686-linux-gnu",
+	"", // non-specific architecture dir
 }
 
 // symbolsLoadedEventGenerator is responsible of generating event if shared object loaded to a process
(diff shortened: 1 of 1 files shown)


pkg/events/derive/symbols_loaded.go
```diff
@@ -16,13 +16,20 @@ func SymbolsLoaded(soLoader sharedobjs.DynamicSymbolsLoader, watchedSymbols []st
...
 var knownLibrariesDirs = []string{
-	"/usr/lib/x86_64-linux-gnu/",
 	"/usr/lib64/",
 	"/usr/lib/",
 	"/lib64/",
 	"/lib/",
 }
 
+var knownArchitectureDirs = []string{
+	"", // non-specific architecture dir
```
NDStrahilevitz, line 26: Would this usually be skipped or not? If so maybe it should be last as range will stop before?

AlonZivony: Usually the libs are in the general directory, not in the arch specific directory.
The arch directories are not relevant to all distributions from what I have read.
In my opinion (which I didn't test) it is probably more efficient to keep it in this order.

rafaeldtinoco: I'm afraid it is the opposite nowadays. If you do a:

```
$ for lib in $(pkg-config --list-all | awk '{print $1}'); do find /usr/lib | grep $lib; done
```

you will see most of the libs placed in `/usr/lib/$(uname -m)-linux-gnu/`, no ? This is because most of the distros (big ones) want to be able to cross compile things and they split libs per arch.

AlonZivony: Interesting.
OK, I will change it so the empty string will be the last one to be checked.
Thanks @NDStrahilevitz @rafaeldtinoco

NDStrahilevitz:
Maybe instead of giving all possible arch directories, we should autodiscover out of the list which exists and then search only those? WDYT?

AlonZivony:
> Maybe instead of giving all possible arch directories, we should autodiscover out of the list which exists and then search only those? WDYT?

Good idea theoretically.
However, can you guarantee that the arch libraries will be the same in containers environments?
I need to check it and think about it.

NDStrahilevitz:
> > Maybe instead of giving all possible arch directories, we should autodiscover out of the list which exists and then search only those? WDYT?
> 
> Good idea theoretically. However, can you guarantee that the arch libraries will be the same in containers environments? I need to check it and think about it.

Good point, then it might be too much complexity for what would probably be a very minimal optimization.

AlonZivony:
From what I have seem each mounting point can have its own architecture libraries directories.
It can also be changed during runtime (if for example, the user installs `arm` libraries to run using `qemu`).

It seems to me we will need to check it for each mounting point and it will still be problematic.
I don't know if this will be cost-effective.

NDStrahilevitz (approved):
LGTM!

PR #2049, opened 2022-08-04
retain context of triggering event to the triggered event

AsafEitani:
Retains the event context of the triggering event
(e.g. hooked_syscall gets the context of the insmod that preceded it)

## Initial Checklist

Fixes: #2019

## Type of change

## How Has This Been Tested?

- `tracee-ebpf -t e=hooked_syscalls`
- load a kernel module with `insmod`
- the `hooked_syscall` event should have `insmod` as it's process name

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -82,11 +82,9 @@
 #define MAX_STR_FILTER_SIZE 16        // bounded to size of the compared values (comm)
 #define FILE_MAGIC_HDR_SIZE 32        // magic_write: bytes to save from a file's header
 #define FILE_MAGIC_MASK     31        // magic_write: mask used for verifier boundaries
-#define NET_SEQ_OPS_SIZE                                                                           \
-    4 // print_net_seq_ops: struct size - TODO: pass this variable as uprobe argument
-#define NET_SEQ_OPS_TYPES                                                                          \
-    6 // print_net_seq_ops: argument size - TODO: pass this variable as uprobe argument
-#define MAX_KSYM_NAME_SIZE 64
+#define NET_SEQ_OPS_SIZE    4 // print_net_seq_ops: struct size - TODO: replace with uprobe argument
+#define NET_SEQ_OPS_TYPES   6 // print_net_seq_ops: argument size - TODO: replace with uprobe argument
+#define MAX_KSYM_NAME_SIZE  64
 
 enum buf_idx_e
 {
@@ -348,8 +346,13 @@ enum container_state_e
     #define MAX_BIN_CHUNKS        110
 #endif
 
-#define NUMBER_OF_SYSCALLS_TO_CHECK_X86 18
-#define NUMBER_OF_SYSCALLS_TO_CHECK_ARM 14
+#if defined(bpf_target_x86)
+    #define NUMBER_OF_SYSCALLS_TO_CHECK 18
+#elif defined(bpf_target_arm64)
+    #define NUMBER_OF_SYSCALLS_TO_CHECK 14
+#else
+    #define NUMBER_OF_SYSCALLS_TO_CHECK 0
+#endif
 
 #define MAX_CACHED_PATH_SIZE 64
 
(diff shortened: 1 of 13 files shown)


yanivagman:
Please rebase on #2031.
I can also see you have here code from Alon's PR.
This should be removed after rebasing.

pkg/ebpf/events_processor.go
```diff
@@ -2,6 +2,7 @@ package ebpf
 
 import (
 	"fmt"
+	"github.com/aquasecurity/tracee/pkg/events/derive"
```
yanivagman, line 5: nit: put this import below with the other github.com/.... imports

pkg/ebpf/events_processor.go
```diff
@@ -363,6 +364,10 @@ func (t *Tracee) processEvent(event *trace.Event) error {
 	// in case FinitModule and InitModule occurs it means that a kernel module was loaded
 	// and we will want to check if it hooked the syscall table and seq_ops
 	case events.DoInitModule:
+		derive.InvokedContextMutex.Lock()
+		contextID := uint64(len(derive.InvokedContext) + 1)
```
yanivagman, line 368: So you use the length of the map as the context ID? Are the events who enter the map never get deleted? If they are deleted, then this method is error prone, and if they are not deleted you have a memory leak.

You can have a GetContextID method in the derive package instead of having the mutex here. In addition, as this is a number that is simply incremented you can use the counter package we have in Tracee, which already takes care of atomicity.

pkg/events/derive/derive.go
```diff
@@ -2,6 +2,8 @@ package derive
 
 import (
 	"fmt"
+	"github.com/aquasecurity/tracee/pkg/events/parse"
```
yanivagman, line 5: nit: move import below

pkg/events/events.go
```diff
@@ -5620,6 +5620,7 @@ var Definitions = eventDefinitions{
 			Sets:         []string{},
 			Params: []trace.ArgMeta{
 				{Type: "unsigned long[]", Name: "syscalls_addresses"},
+				{Type: "unsigned long", Name: "caller_context_id"},
```
yanivagman, line 5623: Does this id has any meaning outside the scope of tracee-ebpf?

pkg/events/derive/derive.go
```diff
@@ -72,3 +82,25 @@ func makeEventSkeleton(eventID events.ID) eventSkeleton {
...
 }
+
+func withInvokingContext(event *trace.Event) (trace.Event, error) {
+	contextID, err := parse.ArgUint64Val(event, "caller_context_id")
+	if err != nil {
+		return trace.Event{}, fmt.Errorf("error parsing caller_context_id arg: %v", err)
+	}
+	if contextID > 0 {
+		InvokedContextMutex.Lock()
+		contextEvent, ok := InvokedContext[contextID]
```
yanivagman, line 93: You don't delete the map entry after it has been used. This is a memory leak.

pkg/ebpf/events_processor.go
```diff
@@ -363,6 +364,10 @@ func (t *Tracee) processEvent(event *trace.Event) error {
...
+		derive.InvokedContext[contextID] = *event
```
yanivagman, line 369: You save all the event, including the arguments which are irrelevant for deriving the event later on.

yanivagman (changes requested):
I'm not sure this is the best way to save the context of the triggering event.
Here is an idea which will remove the need of having an event context map in userspace (that as you will see next in the comments, has some drawbacks, including the need to save the data safely and remove it when it is not necessary):
As you now have the uprobes mechanism, you can simply pass the event's context as an argument to the bpf code, which can then initialize the event with the correct information. WDYT?

AsafEitani:
> I'm not sure this is the best way to save the context of the triggering event. Here is an idea which will remove the need of having an event context map in userspace (that as you will see next in the comments, has some drawbacks, including the need to save the data safely and remove it when it is not necessary): As you now have the uprobes mechanism, you can simply pass the event's context as an argument to the bpf code, which can then initialize the event with the correct information. WDYT?

I wanted to do that but you said there is no need to pass the whole context to the bpf program XD
Do you want me to do that instead?

pkg/events/events.go
AsafEitani: No but that's an internal event so the user won't get it.

pkg/events/derive/derive.go
```diff
@@ -2,14 +2,39 @@ package derive
...
 	"github.com/aquasecurity/tracee/pkg/events"
+	"github.com/aquasecurity/tracee/pkg/events/parse"
 	"github.com/aquasecurity/tracee/types/trace"
 )
 
 // Use as static variable for testability reasons
 var getEventDefFunc = events.Definitions.Get
 
+// Context is a struct used to store the triggering events' context
+type Context struct {
```
NDStrahilevitz, line 17: I think this mechanism can get it's own file.

pkg/events/derive/derive.go
```diff
@@ -2,14 +2,39 @@ package derive
...
+	}
+	contextMapID := uint64(InvokedContext.Counter.Read())
+	InvokedContext.Counter.Increment(1)
+	InvokedContext.Mutex.Lock()
+	InvokedContext.Map[contextMapID] = event
+	InvokedContext.Mutex.Unlock()
+	return contextMapID
+}
+
+var InvokedContext = Context{make(map[uint64]trace.Event), sync.Mutex{}, counter.Counter(1)}
```
NDStrahilevitz, line 36: As a compromise, Instead of exporting this global instance (basically a singleton), I think we can have an upper `AddToMap` which is exported instead of the whole struct. This is then shared internally with the module.

However, as mentioned, the best solution would be to not have shared state though, this should get it's own module and injected from tracee into the deriving module. Then we can easily add tests for this mechanism, which are currently missing.

pkg/events/derive/derive.go
```diff
@@ -2,14 +2,39 @@ package derive
...
 	"github.com/aquasecurity/tracee/types/trace"
 )
 
 // Use as static variable for testability reasons
 var getEventDefFunc = events.Definitions.Get
 
+// Context is a struct used to store the triggering events' context
+type Context struct {
+	Map     map[uint64]trace.Event
+	Mutex   sync.Mutex
```
NDStrahilevitz, line 19: Can't this be a RWMutex?

NDStrahilevitz (changes requested):
Regardless of where this is implemented (userspace or maybe a bpf map?), I think the struct responsible for this context retaining shouldn't be a global instance, since it has state which isn't easily tested otherwise.

I like that it was movied into a struct as that allows us to easily change the mechanism later so i'm not very concerned about the initial userspace implementation.

Going even further, I think the event invoking mechanism could be separated from the deriving mechanism, but that's for a future PR.

pkg/events/derive/derive.go
AsafEitani: Agreed, this will be included in the `triggeredEvents` package as mentioned in #2055

pkg/events/derive/derive.go
AsafEitani: As we will change this mechanism in the next version, is that really important that we change this and only export the `AddToMap` function? we will change it anyway once we create the `triggeredEvents` package.

pkg/events/derive/derive.go
NDStrahilevitz: We can move it already inside the current package, then just move it later.

pkg/events/derive/derive.go
NDStrahilevitz: Global exported variables are just a problem waiting to happen a lot of the times. On that chance I don't want to leave it like this before merging.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3137,6 +3137,12 @@ int BPF_KPROBE(trace_do_exit)
 SEC("uprobe/trigger_syscall_event")
 int uprobe_syscall_trigger(struct pt_regs *ctx)
 {
+#if defined(__TARGET_ARCH_x86)
```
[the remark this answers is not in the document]
rafaeldtinoco: BTW I used `u64 *syscall_table_addr` instead of a `unsigned long *` because I think the value from `get_symboladdr` will be a `* (void *)`, which is u64 (size of a pointer). No ? Just checking..

rafaeldtinoco:
I haven't finished reviewing this but I think the bpf code needs some more polishing. Suggested an example, have not tested it.

pkg/events/derive/derive.go
```diff
@@ -72,3 +75,7 @@ func makeEventSkeleton(eventID events.ID) eventSkeleton {
 		Params: def.Params,
 	}
 }
+
+func withOriginalContext(event *trace.Event) (trace.Event, error) {
```
rafaeldtinoco, line 79: You should place a comment here explaining the difference between Original and Invoking contexts.

pkg/events/derive/trigger.go
```diff
@@ -0,0 +1,55 @@
...
+	invokedContext.Mutex.RUnlock()
+	if !ok {
+		return trace.Event{}, fmt.Errorf("caller_context_id arg not in context map")
+	}
+	// Remove from map to avoid memory leak
+	delete(invokedContext.Map, contextID)
+	return contextEvent, nil
+}
+
+func withInvokingContext(event *trace.Event) (trace.Event, error) {
```
rafaeldtinoco, line 45: This function is targeting a specific event (hooked syscall) and, yet, it is implemented like a generic function to the package. Should we do a switch case here ?

pseudo:
```
switch (event.ID) {
case hookedsyscall:
do what you did
default:
return original context (or error: you should not call this)
```

rafaeldtinoco:
NIT: Could you fix formatting for the previously added TODOs ?

```diff
diff --git a/pkg/ebpf/c/tracee.bpf.c b/pkg/ebpf/c/tracee.bpf.c
index aaf73533..5f9fd4ad 100644
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -74,6 +74,7 @@

 // INTERNAL ----------------------------------------------------------------------------------------

+// clang-format off
 #define MAX_PERCPU_BUFSIZE  (1 << 15) // set by the kernel as an upper bound
 #define MAX_STRING_SIZE     4096      // same as PATH_MAX
 #define MAX_BYTES_ARR_SIZE  4096      // max size of bytes array (arbitrarily chosen)
@@ -82,11 +83,10 @@
 #define MAX_STR_FILTER_SIZE 16        // bounded to size of the compared values (comm)
 #define FILE_MAGIC_HDR_SIZE 32        // magic_write: bytes to save from a file's header
 #define FILE_MAGIC_MASK     31        // magic_write: mask used for verifier boundaries
-#define NET_SEQ_OPS_SIZE                                                                           \
-    4 // print_net_seq_ops: struct size - TODO: pass this variable as uprobe argument
-#define NET_SEQ_OPS_TYPES                                                                          \
-    6 // print_net_seq_ops: argument size - TODO: pass this variable as uprobe argument
-#define MAX_KSYM_NAME_SIZE 64
+#define NET_SEQ_OPS_SIZE    4         // print_net_seq_ops: struct size - TODO: pass this variable as uprobe argument
+#define NET_SEQ_OPS_TYPES   6         // print_net_seq_ops: argument size - TODO: pass this variable as uprobe argument
+#define MAX_KSYM_NAME_SIZE  64
+// clang-format on

 enum buf_idx_e
 {
```

pkg/ebpf/c/tracee.bpf.c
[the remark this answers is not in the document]
rafaeldtinoco: Despite styling changes (to something way less verbose) I think the initial "ifdef" to SYSCALLS_TO_CHECK will save you some ifdefs inside the function.

rafaeldtinoco:
Alright, after addressing my comments, and making sure functionality is good (as you probably are doing already), I think we can merge this and prepare/do the release (considering we've opened a new issue for addressing some of our current concerns).

Note: I do think the derived event package/logic is WAY overcomplicated everytime I read it for reviews =). @NDStrahilevitz (future refactoring work ?).

pkg/events/derive/trigger.go
NDStrahilevitz: This is basically the current logic for defining an invoked event, which other events could be as well, so leaving it generic is the right call I think. 
With that said, the mechanism should be moved, simplifying the derivation package would go naturally with that.

AsafEitani: It's a generic function because it's used both in `hooked_syscalls` and in `hooked_seq_ops`.
Also I think it's best that we keep it generic for future usages (in any way we need to pass this to `singleEventDeriveFunc` so it's the developers' choice if to use the original or this.

rafaeldtinoco (approved):
LGTM.

Issue #1753, opened 2022-05-12
Create an event for collision of SO symbols


AlonZivony:
One known way to create hook and running malicious code is using the LD_PRELOAD method.
In this way, an attacker can export the same function as `libc` or other used shared library, and by loading the malicious shared object before the known one the loader will resolve the imported symbol to be of the malicious shared object instead of the normal one.
To locate this issue, we need event that can check if a shared object loaded to a process has the same symbols as another one.

rafaeldtinoco:
Solved by https://github.com/aquasecurity/tracee/pull/2053 and https://github.com/aquasecurity/tracee/pull/2740

PR #2053, opened 2022-08-08
pkg/ebpf: add derived events for ld SO symbols collision

AlonZivony:
## Initial Checklist

## Description (git log)

Generate a derived event in the case that a shared object loaded
to a process export the same symbols as a loaded shared object.
This helps to detect cases of hooking shared library functions.

Fixes: #1753 

## Type of change

## Final Checklist:

## Git Log Checklist:

My commits logs have:


pkg/events/events.go
```diff
@@ -5767,6 +5768,27 @@ var Definitions = eventDefinitions{
...
+			ID32Bit: sys32undefined,
+			Name:    "symbols_collision",
+			Probes:  []probeDependency{},
+			Dependencies: dependencies{
+				Events: []eventDependency{
+					{EventID: SharedObjectLoaded},
+					{EventID: SchedProcessExec}, // Used to get mount namespace cache
+				},
+				Capabilities: []cap.Value{
+					cap.SYS_PTRACE,   // Used to get host mount NS for bucket cache
```
rafaeldtinoco, line 5781: I'm afraid you will have to check the run with: `--capabilities bypass=false` and, likely ?, raise privileges (ring1) in the places where the privilege is needed (that's a good question for the other bucket cache PR as well.

pkg/events/derive/derive.go
```diff
@@ -71,6 +76,31 @@ func deriveSingleEvent(id events.ID, deriveArgs deriveArgsFunction) deriveFuncti
 	}
 }
 
+// deriveMultipleEvents create an deriveFunction to generate multiple derive trace.Events.
+// The events will be created using the original event information, the ID given and the arguments given.
+// The order of the arguments given will match the order in the event definition, so make sure the order match
+// the order of the params in the events.event struct of the event under events.Definitions.
+// If the arguments given is nil, then no event will be derived.
+func deriveMultipleEvents(id events.ID, multiDeriveArgsFunc multiDeriveArgsFunction) deriveFunction {
```
rafaeldtinoco, line 84: This change and its test should go in another commit with proper git log message (add derive multiple events logic).

docs/events/security_alerts/symbols_collision.md
```diff
@@ -0,0 +1,52 @@
...
+## Intro
+symbols_collision - a shared object loaded to a process has collisions in exported symbols with another loaded shared object.
+
+## Description
+An event marking that a shared object loaded to current process, and have collisions of exported symbols
+with other shared object already loaded to the process. This event can help in inform on the 
+occasion that a shared object tries to override some symbol of another library.
+
+### Configuring the event
+The event is configured using arguments filtering.
```
rafaeldtinoco, line 12: The other symbol events had an example of filtering and what to expect. Shouldn't we have the same here? Okay if you prefer not because of X reasons.

rafaeldtinoco (commented):
This is a big one, I'm still reviewing it (despite the few comments I left).

pkg/events/events.go
rafaeldtinoco: I have created https://github.com/aquasecurity/tracee/pull/2380 to fix a few procfs bad permissions we were having.

pkg/events/derive/derive.go
rafaeldtinoco: Not for this PR, but, It feels to me that we could have only the deriveMultipleEvents now, right ? We can keep a deriveSingleEvent as a wrapper to a single event returned deriveMultipleEvents (just to keep compatibility).

pkg/events/derive/symbols_collision.go
```diff
@@ -0,0 +1,288 @@
...
+
+// addObjCollisions add to cache of one shared object collisions the collision with a second shared object.
+func (socCache sharedObjectsCollisionsCache) addObjCollisions(objID sharedobjs.ObjID,
+	collidedObjID sharedobjs.ObjID,
+	collisions []string) {
+	var objCollisions map[sharedobjs.ObjID][]string
+	collisionsIface, ok := socCache.cache.Get(objID)
+	if ok {
+		objCollisions = collisionsIface.(map[sharedobjs.ObjID][]string)
+		objCollisions[collidedObjID] = collisions
```
rafaeldtinoco, line 228: can't this be removed as you're doing the same after if/else block ?

pkg/events/derive/symbols_collision.go
```diff
@@ -0,0 +1,288 @@
...
+// Get from the cache the collisions check result of the object with other shared objects.
+// Return the collisions per other shared objects, and if the given shared object was monitored in the cache.
+func (socCache sharedObjectsCollisionsCache) getObjCollisions(objID sharedobjs.ObjID) (
+	map[sharedobjs.ObjID][]string, bool) {
+	var collisions map[sharedobjs.ObjID][]string
+	collisionsIface, ok := socCache.cache.Get(objID)
+	if ok {
+		collisions = collisionsIface.(map[sharedobjs.ObjID][]string)
+		return collisions, true
+	} else {
```
rafaeldtinoco, line 210: this is same as return nil, false, right ? And there is no need for else {}

pkg/events/derive/symbols_collision.go
```diff
@@ -0,0 +1,288 @@
...
+// AddCollisions add the collisions between 2 shared objects to the cache.
+func (socCache sharedObjectsCollisionsCache) AddCollisions(obj1 sharedobjs.ObjID,
+	obj2 sharedobjs.ObjID,
+	collisions []string) {
+	socCache.addObjCollisions(obj1, obj2, collisions)
+	socCache.addObjCollisions(obj2, obj1, collisions)
+}
+
+// GetCollision return the symbols collided between 2 shared objects from the cache if exist.
+// If the the collisions are cached for one object, will update it in the seconde object cache.
```
rafaeldtinoco, line 178: nit: second instead of seconde

pkg/events/derive/symbols_collision.go
```diff
@@ -0,0 +1,288 @@
...
+// Return bool indication if the collision existed in the cache.
+func (socCache sharedObjectsCollisionsCache) GetCollision(obj1 sharedobjs.ObjID,
+	obj2 sharedobjs.ObjID) ([]string, bool) {
+	obj1Collisions, ok := socCache.getObjCollisions(obj1)
+	if ok {
+		collisions, ok := obj1Collisions[obj2]
+		if ok {
+			socCache.addObjCollisions(obj2, obj1, collisions)
```
rafaeldtinoco, line 186: So your cache is always duplicated (also in mem footprint) and you're workarounding the fact that one of the duplicates might have vanish because of the LRU. Couldn't you have made a:

map[struct{obj1,obj2}][]string

and do lookups with {obj1,obj2} and {obj2,obj1} (or even sort obj1, obj2 for a single lookup) ?

pkg/events/derive/symbols_collision.go
```diff
@@ -0,0 +1,288 @@
...
+}
+
+// GetProcessLoadedObjects get the shared SOs loaded to the process, and if the process existed in the cache.
+func (pcache *processesLoadedObjectsCache) GetProcessLoadedObjects(pid int) ([]sharedobjs.ObjInfo, bool) {
+	var loadedObjects []sharedobjs.ObjInfo
+	loadedObjectsIface, ok := pcache.cache.Get(pid)
+	if ok {
+		loadedObjects = loadedObjectsIface.([]sharedobjs.ObjInfo)
+		return loadedObjects, true
+	} else {
```
rafaeldtinoco, line 153: same as nil, false ? no need for else ?

pkg/events/derive/symbols_collision.go
```diff
@@ -0,0 +1,288 @@
...
+		if len(collisions) > 0 {
+			collisionEventsArgs = append(collisionEventsArgs, []interface{}{loadingObjectInfo.Path, lsoInfo.Path, collisions})
+		}
+	}
+
+	return collisionEventsArgs, errs
+}
+
+// handleExec delete saved process loaded objects in case of execve, because entire memory is overwritten in this case
+func (soColGen *SymbolsCollisionArgsGenerator) handleExec(event trace.Event) ([][]interface{}, []error) {
```
rafaeldtinoco, line 109: Do we really need the handle Exec?

You're caching the []sharedobjs.ObjInfo per PID number in the cache. Linux will have 4 Million PIDs, which is much higher than the size of the LRU.

I guess you're thinking if we're using filters then the LRU cache might still have an entry for that PID ?

pkg/events/derive/symbols_collision.go
```diff
@@ -0,0 +1,288 @@
...
+func SymbolsCollision(
+	soLoader sharedobjs.DynamicSymbolsLoader,
+	symbolsBlackList []string,
+	symbolsWhiteList []string) deriveFunction {
+	gen := initSOCollisionsEventGenerator(soLoader, symbolsBlackList, symbolsWhiteList)
+	return deriveMultipleEvents(events.SymbolsCollision, gen.deriveArgs)
+}
+
+// SymbolsCollisionArgsGenerator is the struct responsible to create the SO symbols collisions derived event.
+// To do so, it uses multiple caches to reduce performance and chances for failure.
```
rafaeldtinoco, line 26: nit: to accelerate performance (instead of to reduce performance)

rafaeldtinoco (commented):
I left some comments for you with some questions and nits. I'm going to need the deriveMultipleEvents for my simple DNS events, do you mind splitting that logic in a new PR so I can merge it (then we can rebase this one on top) ?

For me this all looks ok, probably will give another look.

I'm concerned about 1024 pids cached only, is it enough to pick collisions (considering an amd64 linux has 4M possible pids) ?

And the memory footprint might be big also, BUT I can understand this is an event that wouldn't be enabled by default (mostly for research in the beginning, likely).

Either way, lets have the "derive multiple events" logic merged in another PR, then I can revisit this one and we can merge it as well.

Thanks!

pkg/events/derive/derive.go
AlonZivony: So you are right, you don't really need the `deriveSingleEvent` function from the abstraction point-of-view.
However, I think it is much easier to add new derive events with it if multiple derived events aren't required. We can adjust the single event to use the multi-events as a specific use-case.

pkg/events/derive/symbols_collision.go
AlonZivony: You are right, this is a lot smarter :)

pkg/events/derive/symbols_collision.go
AlonZivony: What I am thinking is that someone will execute a binary twice in a row using the same process.
This way the cache will maintain the first execution SOs, but even though they are not relevant because the process memory was overwritten.
This should reduce false-positives.

pkg/events/derive/symbols_collision.go
AlonZivony: I need to review my logic here, I think I have a bug.

AlonZivony:
> I left some comments for you with some questions and nits. I'm going to need the deriveMultipleEvents for my simple DNS events, do you mind splitting that logic in a new PR so I can merge it (then we can rebase this one on top) ?
> 
> For me this all looks ok, probably will give another look.
> 
> I'm concerned about 1024 pids cached only, is it enough to pick collisions (considering an amd64 linux has 4M possible pids) ?
> 
> And the memory footprint might be big also, BUT I can understand this is an event that wouldn't be enabled by default (mostly for research in the beginning, likely).
> 
> Either way, lets have the "derive multiple events" logic merged in another PR, then I can revisit this one and we can merge it as well.
> 
> Thanks!

I will push the multi derive logic as a PR tomorrow.
Thanks for the review and hard work!

pkg/events/derive/derive.go
AlonZivony: Opened a PR for this - #2383

rafaeldtinoco: Merged! Thank you! You can rebase this one on top of the merged PR.

AlonZivony:
> I'm concerned about 1024 pids cached only, is it enough to pick collisions (considering an amd64 linux has 4M possible pids) ?

This LRU is mainly used for keeping the loaded symbols while the ELF's dependencies are loaded by the loader.
New shared objects can be loaded afterwards, but I am not sure that it is achievable to keep track of all the processes until they load library dynamically sometime during their runtime.
If you think still increasing this LRU size has positive impact I will increase it.

docs/contributing/events/security_alerts/symbols_collision.md
```diff
@@ -0,0 +1,52 @@
...
+## Dependency Events
+### shared_object_loaded
+The event of shared object loading triggers this event, and supplies the information on the
+shared object necessary to examine its shared objects.
+
+### sched_process_exec
+Used by tracee to maintain mount NS cache, used in this event to get to processes file system.
+Also, used to maintain the cache used by the event for performance improvment.
+
+## Example Use Case
```
roikol, line 33: i think a cmdline example of how to use the event with the filters is very important here

pkg/events/derive/symbols_collision.go
```diff
@@ -0,0 +1,278 @@
...
+import (
+	"fmt"
+	"github.com/aquasecurity/tracee/pkg/events"
+
+	"github.com/aquasecurity/tracee/pkg/utils/sharedobjs"
+	"github.com/aquasecurity/tracee/types/trace"
+	"github.com/hashicorp/golang-lru/simplelru"
+)
+
+// SymbolsCollision generate an events for collisions between symbols exported by shared objects loaded to the same
```
roikol, line 12: nit: generate an events -> generate events

pkg/events/events.go
```diff
@@ -5845,6 +5846,27 @@ var Definitions = eventDefinitions{
 				{Type: "const char*const*", Name: "symbols"},
 			},
 		},
+		SymbolsCollision: {
```
roikol, line 5849: you should add the `DocPath` pointing to the MD file

docs/contributing/events/security_alerts/symbols_collision.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Arguments
+* `loaded_path`:`const char*`[K] - the path of the file loaded.
+* `collision_path`:`const char*`[K,TOCTOU] - the path of the file already loaded, which has collision with the new loaded one.
+* `symbols`:`const char*const*`[U,TOCTOU] - list of symbols collided between the files.
+
+## Dependency Events
+### shared_object_loaded
+The event of shared object loading triggers this event, and supplies the information on the
+shared object necessary to examine its shared objects.
```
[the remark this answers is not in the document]
AlonZivony: Lol you are right
Nice catch

pkg/ebpf/tracee.go
```diff
@@ -589,20 +589,41 @@ func (t *Tracee) initDerivationTable() error {
...
 	}
 
+	symbolsCollisionFilters := map[string]filters.Filter{}
+	for filterScope := range t.config.FilterScopes.Map() {
+		maps.Copy(symbolsCollisionFilters, filterScope.ArgFilter.GetEventFilters(events.SymbolsCollision))
+	}
+	collisionAllowListSymbols := []string{}
+	collisionDenyListSymbols := []string{}
+
+	if symbolsCollisionFilters != nil {
```
rafaeldtinoco, line 613: this is always true

pkg/ebpf/tracee.go
```diff
@@ -589,20 +589,41 @@ func (t *Tracee) initDerivationTable() error {
...
+		collisionSymbolsFilter, ok := symbolsCollisionFilters["symbols"].(*filters.StringFilter)
+		if collisionSymbolsFilter != nil && ok {
+			collisionAllowListSymbols = collisionSymbolsFilter.Equal()
+			collisionDenyListSymbols = collisionSymbolsFilter.NotEqual()
+		}
+	}
+
+	soSymbolsCollisionsDeriveFn := derive.SymbolsCollision(
```
rafaeldtinoco, line 621: I believe this should go inside the `events.SharedObjectLoaded { events.SymbolsCollision {`, the same way you did to `SharedObjectLoaded`.

pkg/ebpf/tracee.go
```diff
@@ -589,20 +589,41 @@ func (t *Tracee) initDerivationTable() error {
...
+		soLoader,
+		collisionAllowListSymbols,
+		collisionDenyListSymbols,
+	)
+
```
rafaeldtinoco, line 626: This got too big to be here IMO.

rafaeldtinoco: I refactored filterscopes into its own package so I can move this initialization to each derive package file and have this cleaned up. I know you're out so if all the rest of the code is good I'll be able to merge with some quick reviews from the OSS team. If not we can sync later.

rafaeldtinoco:
Rebased and with a minor fix at: https://github.com/aquasecurity/tracee/pull/2740

rafaeldtinoco:
I have rebased and merged this one from https://github.com/aquasecurity/tracee/pull/2740. Will move initialization from derivation table function to appropriate place in a new PR.

pkg/events/derive/symbols_collision.go
```diff
@@ -0,0 +1,299 @@
...
+					soColGen.returnedErrors[err.Error()] = true
+					logger.Warn("symbols_loaded", "object loaded", so.ObjInfo, "error", err.Error())
+				} else {
+					logger.Debug("symbols_loaded", "object loaded", so.ObjInfo, "error", err.Error())
+				}
+				return nil, nil
+			}
+			so.FilterSymbols(soColGen.symbolsBlackList)
+			so.FilterOutSymbols(soColGen.symbolsWhiteList)
+			if err != nil {
```
rafaeldtinoco, line 137: It would already have returned here. I'm changing (in a new PR where I'm adding more documentation and removing the initialization from the initDerivationTable function) this. Removing this err check and returning `nil, err` in the check above.

PR #2080, opened 2022-08-21
ebpf: optimize SysEnter and SysExit probes

NDStrahilevitz:
## Initial Checklist

## Description (git log)

```
Date:   Sun Aug 28 13:31:21 2022 +0000

    tracee.bpf.c: optimize sys_enter and sys_exit
    
    Optimize these hot path probes by moving the logic of event init and
    submit to run conditionally in a tail call.
    Only the following logic will run if no event submission is needed:
    1. Storing task's current syscall
    2. Storing syscalls arguments in task
```

Fixes: #issue_number

## Type of change

## How Has This Been Tested?

Currently tested with this script (credits to @rafaeldtinoco):
```
#!/bin/bash
# sysctl -w kernel.bpf_stats_enabled=1
bpftool prog show > /tmp/progshow.$$
checkfile() {
    prognum=$(echo $line | cut -d' ' -f1)
    progname=$(echo $line | cut -d' ' -f2)
    runtime=$(cat /tmp/progshow.$$ | grep ^$prognum | awk '{print $9}')
    amount=$(cat /tmp/progshow.$$ | grep ^$prognum | awk '{print $11}')
    if [[ $runtime -eq 0 ]]; then runtime=1; fi
    if [[ $amount -eq 0 ]]; then amount=1; fi
    average=$((runtime/amount))
    echo "PROGRAM: $progname (runtime: $runtime ns, amount: $amount times, average: $average ns)"
}
# kprobes
bpftool perf | grep kprobe | awk '{print $6" "$9}' | \
while read line; do
    checkfile
done
# traces
bpftool perf | grep trace | awk '{print $6" "$8}' | \
while read line; do
    checkfile
done
rm -f /tmp/progshow.$$
```

1. Run `tracee-ebpf -o none`
2. Let it run for about 30 seconds
3. Run the above script, observe the average of sys_enter and sys_exit (should be around ~1000ns and ~3000ns respectively)
4. Rerun tracee-ebpf with `tracee-ebpf -o none -t e=task_rename,container_create,security_file_open` (this will cause skipping of the event logic)
5. Let it run for 30 seconds
6. Run the above script - average for sys_enter and sys_exit should now be lower.

In addition: no events should break their previous behavior, this can be verified as such:
1. Run `tracee-ebpf -o option:detect-syscall -t e=task_rename`
2. Syscall arguments should be parsed successfully

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -922,8 +922,10 @@ BPF_ARRAY(file_filter, path_filter_t, 3);                          // filter vfs
 BPF_PERCPU_ARRAY(bufs, buf_t, MAX_BUFFERS);                        // percpu global buffer variables
 BPF_PROG_ARRAY(prog_array, MAX_TAIL_CALL);                         // store programs for tail calls
 BPF_PROG_ARRAY(prog_array_tp, MAX_TAIL_CALL);                      // store programs for tail calls
-BPF_PROG_ARRAY(sys_enter_tails, MAX_EVENT_ID);                     // store programs for tail calls
-BPF_PROG_ARRAY(sys_exit_tails, MAX_EVENT_ID);                      // store programs for tail calls
+BPF_PROG_ARRAY(sys_enter_tails, MAX_EVENT_ID);                     // store syscall specific programs for tail calls from sys_enter
+BPF_PROG_ARRAY(sys_exit_tails, MAX_EVENT_ID);                      // store syscall specific programs for tail calls from sys_exit
+BPF_PROG_ARRAY(sys_enter_submit_tail, 1);                          // store program for submitting syscalls from sys_enter
+BPF_PROG_ARRAY(sys_exit_submit_tail, 1);                           // store program for submitting syscalls from sys_exit
 BPF_STACK_TRACE(stack_addresses, MAX_STACK_ADDRESSES);             // store stack traces
 BPF_HASH(module_init_map, u32, kmod_data_t, 256);                  // holds module information between
 BPF_LRU_HASH(fd_arg_path_map, fd_arg_task_t, fd_arg_path_t, 1024); // store fds paths by task
@@ -1597,6 +1599,38 @@ init_context(event_context_t *context, struct task_struct *task, u32 options)
     return 0;
 }
 
+static __always_inline task_info_t *init_task_info(u32 key, bool *initialized)
+{
+    task_info_t *task_info = bpf_map_lookup_elem(&task_info_map, &key);
+    if (initialized != NULL) {
+        *initialized = task_info == NULL;
+    }
+    if (task_info == NULL) {
+        // unlikely code path - possibly optimize with unlikely macro later
+
+        // get the submit buffer to fill the task_info_t in the map
+        // this is done because allocating the stack space for task_info_t usually
+        // crosses the verifier limit.
+        int buf_idx = SUBMIT_BUF_IDX;
+        void *submit_buffer = bpf_map_lookup_elem(&bufs, &buf_idx);
+        if (submit_buffer == NULL)
+            return NULL;
+
+        bpf_map_update_elem(&task_info_map, &key, submit_buffer, BPF_NOEXIST);
+        task_info = bpf_map_lookup_elem(&task_info_map, &key);
+        // appease the verifier
+        if (task_info == NULL) {
+            return NULL;
+        }
+        task_info->syscall_traced = false;
+        task_info->new_task = false;
+        task_info->follow = false;
+        task_info->recompute_scope = true;
+        task_info->container_state = CONTAINER_UNKNOWN;
+    }
+    return task_info;
+}
+
 static __always_inline int init_event_data(event_data_t *data, void *ctx)
 {
     int zero = 0;
@@ -1613,23 +1647,14 @@ static __always_inline int init_event_data(event_data_t *data, void *ctx)
     if (data->submit_p == NULL)
         return 0;
 
-    // try updating task_info_map with noexist flag. If successful, we will need to initialize it
-    // later. to save stack space (of size task_info_t) update the map with garbage data using the
-    // submit buffer.
-    int exist = bpf_map_update_elem(
-        &task_info_map, &data->context.task.host_tid, data->submit_p, BPF_NOEXIST);
-    data->task_info = bpf_map_lookup_elem(&task_info_map, &data->context.task.host_tid);
+    // check if task_info was initialized in this call
+    bool task_info_initalized = false;
+    data->task_info = init_task_info(data->context.task.host_tid, &task_info_initalized);
     if (data->task_info == NULL) {
         return 0;
     }
     bool container_lookup_required = true;
-    if (!exist) {
-        data->task_info->syscall_traced = false;
-        data->task_info->new_task = false;
-        data->task_info->follow = false;
-        data->task_info->recompute_scope = true;
-        data->task_info->container_state = CONTAINER_UNKNOWN;
-    } else {
+    if (!task_info_initalized) {
         // check if we need to recompute scope due to context change
         task_context_t *old_context = &data->task_info->context;
         task_context_t *new_context = &data->context.task;
@@ -2770,20 +2795,22 @@ static __always_inline struct pipe_inode_info *get_file_pipe_info(struct file *f
 SEC("raw_tracepoint/sys_enter")
 int tracepoint__raw_syscalls__sys_enter(struct bpf_raw_tracepoint_args *ctx)
 {
-    event_data_t data = {};
-    if (!init_event_data(&data, ctx))
-        return 0;
+    struct task_struct *task = (struct task_struct *) bpf_get_current_task();
 
-    if (!should_trace(&data))
+    u32 task_id = bpf_get_current_pid_tgid(); // get the tid only
+    // use config as garbage value here
+    task_info_t *task_info = init_task_info(task_id, NULL);
+    if (task_info == NULL) {
         return 0;
+    }
 
-    syscall_data_t *sys = &data.task_info->syscall_data;
+    syscall_data_t *sys = &(task_info->syscall_data);
     sys->id = ctx->args[1];
 
     if (get_kconfig(ARCH_HAS_SYSCALL_WRAPPER)) {
         struct pt_regs *regs = (struct pt_regs *) ctx->args[0];
 
-        if (is_x86_compat(data.task)) {
+        if (is_x86_compat(task)) {
 #if defined(bpf_target_x86)
             sys->args.args[0] = READ_KERN(regs->bx);
             sys->args.args[1] = READ_KERN(regs->cx);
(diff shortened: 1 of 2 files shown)


rafaeldtinoco (approved):
It looks good to time. As another optimization I would skip all timing related syscalls right away (without even doing map lookups). @yanivagman mind taking a look at this change ? I think it is solid, but the code is sensitive and getting a second look would be good.

NDStrahilevitz:
> It looks good to time. As another optimization I would skip all timing related syscalls:
> 
> image
> 
> right away (without even doing map lookups). @yanivagman mind taking a look at this change ? I think it is solid, but the code is sensitive and getting a second look would be good.

Sounds like a good additional optimization, maybe for another PR rather than in this one though. I'll add it to the the issue tracking optimization.

NDStrahilevitz:
> > It looks good to time. As another optimization I would skip all timing related syscalls

Sounds like a good additional optimization, maybe for another PR rather than in this one though. I'll add it to the the issue tracking optimization.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2770,20 +2783,28 @@ static __always_inline struct pipe_inode_info *get_file_pipe_info(struct file *f
...
+    u64 task_id = bpf_get_current_pid_tgid(); // get the tgid only
+    task_info_t *task_info = NULL;
+
+    int zero = 0;
+    config_entry_t *config = (config_entry_t *) bpf_map_lookup_elem(&config_map, &zero);
+    if (config == NULL)
         return 0;
 
-    if (!should_trace(&data))
+    // use config as garbage value here
```
yanivagman, line 2795: This is tricky, because we need the size of the config map value to be greater than the size of the task_info map value. Now it works (task_info map value is ~170 bytes in size, while the config map is ~210 bytes), but we may need to pay attention to this in the future.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2818,43 +2839,67 @@ int tracepoint__raw_syscalls__sys_enter(struct bpf_raw_tracepoint_args *ctx)
...
-            void *file_path = get_path_str(GET_FIELD_ADDR(file->f_path));
+    bool submitEnter = should_submit(RAW_SYS_ENTER, config);
+    bool submitSyscall = should_submit(sys->id, config);
 
-            bpf_probe_read_str(&fd_arg_path.path, sizeof(fd_arg_path.path), file_path);
-            bpf_map_update_elem(&fd_arg_path_map, &fd_arg_task, &fd_arg_path, BPF_ANY);
-        }
+    if (submitEnter || submitSyscall) {
+        bpf_tail_call(ctx, &sys_enter_submit_tail, 0);
     }
```
yanivagman, line 2854: This (and the calculation of the submitEnter and submitSyscall) is already done in the tail call, so no need for it here. Also, see my next comment for another reason why we need to remove this.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2818,43 +2839,67 @@ int tracepoint__raw_syscalls__sys_enter(struct bpf_raw_tracepoint_args *ctx)
...
 
+    // call syscall handler, if exists
+    bpf_tail_call(ctx, &sys_enter_tails, sys->id);
```
yanivagman, line 2857: This will never happen now, as the above tail call never returns.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2898,34 +2952,55 @@ int tracepoint__raw_syscalls__sys_exit(struct bpf_raw_tracepoint_args *ctx)
...
         return 0;
 
+    sys->ret = ret;
+
+    bool submitExit = should_submit(RAW_SYS_EXIT, config);
+    bool submitSyscall = should_submit(sys->id, config);
+
+    if (submitExit || submitSyscall) {
+        bpf_tail_call(ctx, &sys_exit_submit_tail, 0);
+    }
```
yanivagman, line 2962: This should be removed - same reasons as in sys_enter

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2898,34 +2952,55 @@ int tracepoint__raw_syscalls__sys_exit(struct bpf_raw_tracepoint_args *ctx)
...
+    uint id = sys->id;
+    long ret = ctx->args[1];
+
     if (should_submit(RAW_SYS_EXIT, data.config)) {
         save_to_submit_buf(&data, (void *) &id, sizeof(int), 0);
         events_perf_submit(&data, RAW_SYS_EXIT, ret);
     }
 
-    if (should_submit(id, data.config)) {
+    if (should_submit(sys->id, data.config) && (id != SYSCALL_EXECVE && id != SYSCALL_EXECVEAT) ||
```
yanivagman, line 2986: Either use `sys->id` or `id`

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2898,34 +2952,55 @@ int tracepoint__raw_syscalls__sys_exit(struct bpf_raw_tracepoint_args *ctx)
...
+        ((id == SYSCALL_EXECVE || id == SYSCALL_EXECVEAT) && (ret != 0))) {
+        // We can't use saved args after execve syscall, as pointers are
+        // invalid To avoid showing execve event both on entry and exit, we
+        // only output failed execs
         u64 types = 0;
         u64 *saved_types = bpf_map_lookup_elem(&params_types_map, &id);
         if (!saved_types) {
-            goto out;
```
yanivagman, line 2910: We may still want to call an exit tail call, so let's keep the `goto out`

yanivagman (changes requested):

yanivagman:
> It looks good to time. As another optimization I would skip all timing related syscalls

What if the user chose to trace one of these syscalls?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2770,20 +2783,28 @@ static __always_inline struct pipe_inode_info *get_file_pipe_info(struct file *f
...
 int tracepoint__raw_syscalls__sys_enter(struct bpf_raw_tracepoint_args *ctx)
 {
-    event_data_t data = {};
-    if (!init_event_data(&data, ctx))
+    struct task_struct *task = (struct task_struct *) bpf_get_current_task();
+    u64 task_id = bpf_get_current_pid_tgid(); // get the tgid only
+    task_info_t *task_info = NULL;
+
+    int zero = 0;
+    config_entry_t *config = (config_entry_t *) bpf_map_lookup_elem(&config_map, &zero);
```
yanivagman, line 2791: This lookup might not be necessary

pkg/ebpf/c/tracee.bpf.c
NDStrahilevitz: Yep, I also had the problem with the stack space, why I cotinued with this method.
If you believe config can be a volatile struct for this, maybe we should use another one?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1597,6 +1599,27 @@ init_context(event_context_t *context, struct task_struct *task, u32 options)
...
 }
 
+// we pass a pointer to a pointer here so we can move from a NULL pointer
+static __always_inline int init_task_info(task_info_t **task_info, u32 key, void *dump)
+{
+    // try updating task_info_map with noexist flag. If successful, we will need to initialize it
+    // later. to save stack space (of size task_info_t) update the map with garbage data using the
+    // dump argument.
+    int exist = bpf_map_update_elem(&task_info_map, &key, dump, BPF_NOEXIST);
+    *task_info = bpf_map_lookup_elem(&task_info_map, &key);
```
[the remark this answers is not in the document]
NDStrahilevitz: Added this optimization.

pkg/events/events.go
```diff
@@ -4885,6 +4890,11 @@ var Definitions = eventDefinitions{
 			Probes: []probeDependency{
 				{Handle: probes.SysExit, Required: true},
 			},
+			Dependencies: dependencies{
+				TailCalls: []TailCall{
+					{MapName: "sys_exit_submit_tail", MapIdx: 0, ProgName: "sys_exit_submit"},
```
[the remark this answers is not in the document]
NDStrahilevitz: They are essential events atm.

pkg/ebpf/c/tracee.bpf.c
NDStrahilevitz: Tried writing it out, and it makes the code worse off overall, since in exit we don't have the syscall id from the get go. In sys enter it doens't make much of a difference in terms of performance atm.

pkg/ebpf/c/tracee.bpf.c
NDStrahilevitz: We have this tail call at the end of the submit tailcall if it's reached.

pkg/ebpf/c/tracee.bpf.c
NDStrahilevitz: see my comment [here](https://github.com/aquasecurity/tracee/pull/2080#discussion_r958348112) on that

pkg/ebpf/c/tracee.bpf.c
NDStrahilevitz: needed for submit checks, see [here](https://github.com/aquasecurity/tracee/pull/2080#discussion_r958348112)

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2770,20 +2783,28 @@ static __always_inline struct pipe_inode_info *get_file_pipe_info(struct file *f
 SEC("raw_tracepoint/sys_enter")
 int tracepoint__raw_syscalls__sys_enter(struct bpf_raw_tracepoint_args *ctx)
 {
-    event_data_t data = {};
-    if (!init_event_data(&data, ctx))
+    struct task_struct *task = (struct task_struct *) bpf_get_current_task();
+    u64 task_id = bpf_get_current_pid_tgid(); // get the tgid only
```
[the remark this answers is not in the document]
NDStrahilevitz: fixed

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1597,6 +1599,36 @@ init_context(event_context_t *context, struct task_struct *task, u32 options)
...
+        *initialized = task_info == NULL;
+    }
+    if (task_info == NULL) {
+        // unlikely code path - possibly optimize with unlikely macro later
+
+        // get the submit buffer to fill the task_info if needed
+        int buf_idx = SUBMIT_BUF_IDX;
+        void *config = bpf_map_lookup_elem(&bufs, &buf_idx);
+        if (config == NULL)
+            return 0;
```
yanivagman, line 1615:
```suggestion
            return NULL;
```

rafaeldtinoco:
> > It looks good to time. As another optimization I would skip all timing related syscalls
> 
> What if the user chose to trace one of these syscalls?

Well I did not think we had that use case but I see your point. And thinking a bit more about the suggestion, I believe a big optimization for cases relying in timing syscalls a lot is to use vDSO (what glibc and musl might be already doing), so... nevermind. =)

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1597,6 +1599,36 @@ init_context(event_context_t *context, struct task_struct *task, u32 options)
...
+    task_info_t *task_info = bpf_map_lookup_elem(&task_info_map, &key);
+    if (initialized != NULL) {
+        *initialized = task_info == NULL;
+    }
+    if (task_info == NULL) {
+        // unlikely code path - possibly optimize with unlikely macro later
+
+        // get the submit buffer to fill the task_info if needed
+        int buf_idx = SUBMIT_BUF_IDX;
+        void *config = bpf_map_lookup_elem(&bufs, &buf_idx);
```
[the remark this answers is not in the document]
NDStrahilevitz: name is a left over, will fix both.

yanivagman (approved):
LGTM!

PR #2086, opened 2022-08-22
[REBASE ONLY] for the PR: #1971

rafaeldtinoco:
This is a rebase for PR: #1971 , squashing 2 commits and solving minor conflicts.

### DIFF
--- a/cmd/tracee-ebpf/flags/output.go
+++ b/cmd/tracee-ebpf/flags/output.go
@@ -29,6 +29,7 @@ option:{stack-addresses,detect-syscall,exec-env,relative-time,exec-hash,parse-ar
   relative-time                                    use relative timestamp instead of wall timestamp for events
   exec-hash                                        when tracing sched_process_exec, show the file hash(sha256) and ctime
   parse-arguments                                  do not show raw machine-readable values for event arguments, instead parse into human readable strings
+  parse-arguments-fds                              enable parse-arguments and enrich fd with its file path translation. This can cause pipeline slowdowns.
   sort-events                                      enable sorting events before passing to them output. This will decrease the overall program efficiency.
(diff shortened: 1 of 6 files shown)


yanivagman:
To avoid an extra context switch while reading the map value in userspace, wouldn't it have been better to read the map value in sys_exit and send it as part of the event?

Also, shouldn't we change the fd argument type to string now?

rafaeldtinoco:
@yanivagman I'm sorry Yaniv, I thought you were out next couple of weeks or else I would have pinged you before merging it. About the fd argument change, if no signatures are relying in fd being an integer we could. I thought about this feature being "an extra" to introspection (more than anything else) as it adds some extra processing that we don't want when dealing with signatures. WDYT ?

geyslan:
> Also, shouldn't we change the fd argument type to string now?

@yanivagman if `parse-arguments-fds` is set, yeah, userspace outputs json as `fdArg.Value.(String)` even though it is initialized as `fdArg.Value.(int32)`.

yanivagman:
> > Also, shouldn't we change the fd argument type to string now?
> 
> @yanivagman if `parse-arguments-fds` is set, yeah, userspace outputs json as `fdArg.Value.(String)` even though it is initialized as `fdArg.Value.(int32)`.

If this is the output:
`[{"name":"fd","type":"int","value":"3=/etc/passwd"},`
Then type should be string then

yanivagman:
> @yanivagman I'm sorry Yaniv, I thought you were out next couple of weeks or else I would have pinged you before merging it. About the fd argument change, if no signatures are relying in fd being an integer we could. I thought about this feature being "an extra" to introspection (more than anything else) as it adds some extra processing that we don't want when dealing with signatures. WDYT ?

Yes, makes sense

rafaeldtinoco:
@geyslan could you move the event generation to sys_exit() then, based in Yaniv's feedback ? Thanks!

PR #2093, opened 2022-08-24
log: introduce logger package

geyslan:
## Initial Checklist

## Description (git log)

Introduce logger package and make sample changes to cmd/tracee-ebpf/main.go.

Fixes: #1787 

## Type of change

## How Has This Been Tested?

- [ ] Test File A
- [ ] Test File B

- command 01
- command 02

## Final Checklist:

## Git Log Checklist:

My commits logs have:


pkg/logger/logger.go
```diff
@@ -0,0 +1,328 @@
...
+
+	var (
+		cfg     zap.Config
+		encoder zapcore.Encoder
+		metrics bool
+	)
+
+	if lvl == DebugLevel {
+		cfg = zap.NewDevelopmentConfig()
+		encoder = zapcore.NewConsoleEncoder(cfg.EncoderConfig)
```
AlonZivony, line 304: Don't you think the encoding should be decided by the user?

pkg/logger/logger.go
```diff
@@ -0,0 +1,328 @@
...
+
+type LogCounter struct {
+	rwMutex sync.RWMutex
+	data    map[LogOrigin]uint32
+}
+
+func (lc *LogCounter) update(lo LogOrigin) {
+	lc.rwMutex.Lock()
+	lc.data[lo]++
+	lc.rwMutex.Unlock()
```
AlonZivony, line 36: For best practice, I think you should use the `defer` option for the unlock, to avoid cases in the future while modifying the function that the unlock won't be reached for some reason.
This is true for the following 2 functions as well in my opinion.

AlonZivony (commented):
Just looked at the code.
I had few comments that you can change as I suggested or tell me if you think I am wrong.
My main issue here is the following - how do you plan to give all packages the logger?
I guess you didn't yet do this change, but I think this is the main issue with creating a logger instance.

geyslan:
> Just looked at the code. I had few comments that you can change as I suggested or tell me if you think I am wrong.

@AlonZivony, first, thanks for your review. :+1: 

> My main issue here is the following - how do you plan to give all packages the logger? I guess you didn't yet do this change, but I think this is the main issue with creating a logger instance.

Currently it's being designed to be used at [package level](https://github.com/aquasecurity/tracee/pull/2093/files#diff-c1dc2eb1e149a1dc7ecd5cb474e167fb456eb7ad0ea0256cf62fba693a89af79R288-R289) `var defaultLogger = New(os.Stdout, InfoLevel)` as well through user-defined instantiations `logger.New()`. Both usages (for Debug) are [here](https://github.com/aquasecurity/tracee/pull/2093/files#diff-c1dc2eb1e149a1dc7ecd5cb474e167fb456eb7ad0ea0256cf62fba693a89af79R168-R174).

In this way, a project can always rely on the package-level `logger.Default()` in conjunction with the user-defined ones:

```go
import "github.com/aquasecurity/tracee/pkg/logger"

logger.Debug(...)
logger.Info(...)

logFile = logger.New(fileWriter, logger.DebugLevel, "enrichment", "ebpf") // only logs when log calls are made from packages ending with "enrichment" or "ebpf".
logFile.Debug(...)
```
The experience can be improved if we create a `logger.SetDefault(l *Logger)` to overwrite defaultLogger with a user-defined logger. WDYT?

cmd/tracee-ebpf/main.go
```diff
@@ -26,6 +30,8 @@ var traceeInstallPath string
 var listenMetrics bool
 var metricsAddr string
 var enrich bool
+var logStdout = logger.Default()
```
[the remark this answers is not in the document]
geyslan: > In my opinion we don't need log for both stdout and stderr.

It would demand a way to set multi writers for an unique logger. I think it can be done using `zapcore.NewMultiWriteSyncer`. 

What if the user require a unique logger with a single writer? It seems the way that tracee is doing this [currently](https://github.com/aquasecurity/tracee/blob/main/cmd/tracee-ebpf/main.go#L65-L66).

> The logger configuration and levels should indicate to where the log should go.

I agree with you that this is a well tied approach. But as I mentioned, if tracee or other project requires to log to different outputs and/or to use different levels for that matter, it would preferable to use a loose approach, letting the user to set it on its own.

pkg/logger/logger.go
geyslan: Yeah, it can be changed.

pkg/logger/logger.go
NDStrahilevitz: Just to note: if this becomes a hot path, defer has a performance overhead. But since this is currently only used in debug mode it should be fine.

cmd/tracee-ebpf/main.go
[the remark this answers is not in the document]
NDStrahilevitz: I think we're already overly flexible and non standard in how we log. 
IMO this PR should be used to solve that problem, not add more flexibility. I think Alon's approach would be better for that.
We also need to remove the debug drilling we used throughout the code, as it's clutter in a lot of cases.
Calling logger.Debug(...) should be enough and should output when running tracee in debug mode.

pkg/logger/logger.go
```diff
@@ -0,0 +1,328 @@
...
+	return l.l.Sync()
+}
+
+// Default returns default logger
+func Default() *Logger {
+	return defaultLogger
+}
+
+// package level Logger
+var defaultLogger = New(os.Stdout, InfoLevel)
```
NDStrahilevitz, line 289: By default we should probably log to stderr, since currently we use stdout as the output of events.

pkg/logger/logger.go
```diff
@@ -0,0 +1,328 @@
...
+		encoder = zapcore.NewJSONEncoder(cfg.EncoderConfig)
+	}
+
+	core := zapcore.NewCore(
+		encoder,
+		zapcore.AddSync(w),
+		zapcore.Level(lvl),
+	)
+
+	pkgs := map[string]bool{}
```
NDStrahilevitz, line 317: Can you please explain how this is used? Maybe add usage to another package to show this mechanism off.

NDStrahilevitz (commented):
So the main reason I thought we need a logger is:
1. Standardizing the way we log throught the project
2. Solving the "debug drilling" everywhere

Number 1 I think is sort of solved here, although I'm not sure in that case why we don't just use a direct zap logger, or any other logger for that matter. Maybe i'm missing something clever done here.

Debug drilling is left as is, which I personally would like to see gone.

In the logger itself, i'm not 100% sure what the metrics and pkgs are supposed to be or solve, some docs on that would be appreciated.

geyslan:
@NDStrahilevitz, thank you for your review.

> Number 1 I think is sort of solved here, although I'm not sure in that case why we don't just use a direct zap logger, or any other logger for that matter. Maybe i'm missing something clever done here.

Using zap logger directly is totally possible but we won't be able to put in one place the required features such as filtering packages and counting logs.

> Debug drilling is left as is, which I personally would like to see gone, since it heavily clutters the code in a lot of places.

Using the package level `logger.Debug()` should suffix that, am I right?

> In the logger itself, i'm not 100% sure what the metrics and pkgs are supposed to be or solve, some docs on that would be appreciated.

I've demonstrated it in this repo: https://github.com/geyslan/logger-poc/blob/main/cmd/zap/main.go

Summarizing:

outerpackage.go
```go
package outerpackage

import logger "github.com/geyslan/logger-poc/logger/zap"

type empty struct{}

func (e empty) fun(l *logger.Logger) {
	l.Debug("4 Debbuging from outer package")
}

func Log(l *logger.Logger) {
	l.Debug("3 Debbuging from outer package")
	empty{}.fun(l)
}
```

main.go
```go
func generateLogsOuter(log *logger.Logger) {
	fmt.Println()
	fmt.Println(strings.Repeat("-", 5), "generateLogs()")

	log.Debug("1 Debbuging from main package")
	log.Debug("2 Debbuging from main package")
	outerpackage.Log(log)
}

func main() {
...
	filteredLogger := logger.New(os.Stdout, logger.DebugLevel, "outerpackage")
	generateLogsOuter(filteredLogger)
	dumpMetrics(filteredLogger)
}
```

`filteredLogger` only logs and updates metrics if logger calls are made from "outerpackage" go package. It only filters calls out if packages are specified on the logger creation.

For that case, we can see in output only the "outerpackage" logs.

```
----- generateLogs()
2022-08-25T15:39:33.615-0300	DEBUG	3 Debbuging from outer package
2022-08-25T15:39:33.615-0300	DEBUG	4 Debbuging from outer package

>>>>> Dump LogCount
key: {File:/home/gg/code/logger-poc/outerpackage/outerpackage.go Line:12} - value: 1
key: {File:/home/gg/code/logger-poc/outerpackage/outerpackage.go Line:8} - value: 1

```
These calls from main were just ignored:

```
	log.Debug("1 Debbuging from main package")
	log.Debug("2 Debbuging from main package")
```

pkg/logger/logger.go
geyslan: I demonstrated in https://github.com/aquasecurity/tracee/pull/2093?ts=8#issuecomment-1227636610 comment. In tracee I think it could be useful when passing a flag like `tracee -debug=ebpf`.

cmd/tracee-ebpf/main.go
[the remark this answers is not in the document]
geyslan: > We also need to remove the debug drilling we used throughout the code, as it's clutter in a lot of cases. Calling logger.Debug(...) should be enough and should output when running tracee in debug mode.

I think this PR is already removing the drilling if the package importer decide to use the package level functions. But for sure we can remove the possibility to create new user-defined loggers letting all importers relying on a singleton.

geyslan:
These are the main points to be solved at the moment, based on the conversation that @rafaeldtinoco had with @NDStrahilevitz:

1. log errors and debug cases
2. debug should be a no-op when not running in debug mode
3. caching the errors to avoid repetitive prints

I will update https://github.com/aquasecurity/tracee/issues/1787 with this so anyone can know the proper requirements.

geyslan:
Closing this due to the new PR https://github.com/aquasecurity/tracee/pull/2110 and to keep this proposal unchanged.

PR #2108, opened 2022-08-26
[MAINT] Latest gofmt

rafaeldtinoco:
This is a maintenance PR.


rafaeldtinoco:
We can't do this change yet: looks like gofmt from golang 1.19 has new rules that will cause some problems if we mix using old and newer formatters.

PR #2110, opened 2022-08-26
log: introduce simple logger package

geyslan:
## Initial Checklist

## Description (git log)

Introduce logger package and make sample changes to cmd/tracee-ebpf/main.go.

This is a proposal based on the discussion in #2093 PR. 

```
commit 4db97d9fdfdd0d26c60c82b7ddec2d2d499f4dd3 (HEAD -> 1787-logger-simple, origin/1787-logger-simple)
Author: Geyslan Gregório <geyslan@gmail.com>
Date:   Tue Aug 30 08:51:33 2022 -0300

    log: kickstarting logger use

commit a055ca72d6bc4c05183542235d8cc701212421bc
Author: Geyslan Gregório <geyslan@gmail.com>
Date:   Tue Aug 30 08:43:04 2022 -0300

    log: introduce simple logger package
    
    This logger can be used through all runtime directly from its
    package-level instance as well as from instances maintained by the
    consumer itself.
    
    Its base is a zap logger with the addition of log (origin) counting.
    
    Environment variables can be used to set the base logger:
    TRACEE_LOGGER_LVL, TRACEE_LOGGER_ENCODER and TRACEE_LOGGER_AGGREGATE.
```

Fixes: #1787 

## Type of change

## How Has This Been Tested?

- [ ] Test File A
- [ ] Test File B

- command 01
- command 02

## Final Checklist:

## Git Log Checklist:

My commits logs have:


cmd/tracee-ebpf/main.go
```diff
@@ -57,17 +58,33 @@ func main() {
...
+						}
+						return logger.InfoLevel
+
+					}(debug),
+					Encoder: logger.NewJSONEncoder(
+						logger.NewProductionConfig().EncoderConfig,
+					),
+				},
+			)
+
```
geyslan, line 78: This is just to show the Initialization capability. The logger (at package level) is already initialized to its [default](https://github.com/aquasecurity/tracee/pull/2110/files#diff-c1dc2eb1e149a1dc7ecd5cb474e167fb456eb7ad0ea0256cf62fba693a89af79R243).

cmd/tracee-ebpf/main.go
```diff
@@ -201,17 +216,15 @@ func main() {
 			if listenMetrics {
 				err := t.Stats().RegisterPrometheus()
 				if err != nil {
-					fmt.Fprintf(os.Stderr, "Error registering prometheus metrics: %v\n", err)
+					logger.Error("Registering prometheus metrics", logger.ErrorField(err))
```
NDStrahilevitz, line 219: Maybe we should have a more structured string here instead like "prometheus_metrics" WDYT? Otherwise I think the previous "free" string would fit better.

cmd/tracee-ebpf/main.go
```diff
@@ -201,17 +216,15 @@ func main() {
...
 					mux.Handle("/metrics", promhttp.Handler())
 
 					go func() {
-						if debug {
-							fmt.Fprintf(os.Stdout, "Serving metrics endpoint at %s\n", metricsAddr)
-						}
+						logger.Debug("Serving metrics endpoint at", logger.String("metricsAddr", metricsAddr))
 						if err := http.ListenAndServe(metricsAddr, mux); err != http.ErrServerClosed {
-							fmt.Fprintf(os.Stderr, "Error serving metrics endpoint: %v\n", err)
+							logger.Error("Serving metrics endpoint", logger.ErrorField(err))
```
NDStrahilevitz, line 227: ditto

pkg/logger/logger.go
```diff
@@ -0,0 +1,269 @@
...
+		panic("Logger cannot be nil")
+	}
+
+	defaultLogger = l
+}
+
+// Init initializes the package level default logger using given config.
+// It's not thread safe so if required use it always at the beggining.
+// Log count of the current default logger will be lost.
+func Init(cfg *LoggerConfig) {
```
NDStrahilevitz, line 263: So I checked for the case where an Init call is missing and messages were simply gone from output. While this is better than panicking, it's important that at the very least it's documented that an Init call must be made, otherwise consuming apps may miss it and miss output logs.
Maybe even an `init` method here would be appropriate, and then `Init` can override the default set logger.

NDStrahilevitz (changes requested):
This is much better IMO. I would like a way to set the log level externally from the CLI, like an environment variable (this is necessary for apps consuming tracee as a library, which currently can't set the logging level and could also differ between pacakges).

Still need to check if the error aggregation works correctly (symbols_loaded.go is the place to do so as described in #2036).

pkg/logger/logger.go
geyslan: Actually if Init() call is missing the logs won't be gone, since `defaultLogger` is [initialized](https://github.com/aquasecurity/tracee/pull/2110/files#diff-c1dc2eb1e149a1dc7ecd5cb474e167fb456eb7ad0ea0256cf62fba693a89af79R243) first. Perhaps, to avoid confusion, Init() should be named `Reset()` and SetDefault() made private `setDefault()`. Does it make sense?

Regarding init(), I agree with you.

Something like:

```go
var defaultLogger *Logger

func init() {
	NewLogger(NewDefaultLoggerConfig())
}
```

cmd/tracee-ebpf/main.go
geyslan: I would go with structured `logger.Error("prometheus_metrics", logger.ErrorField(err))`.

pkg/logger/logger.go
NDStrahilevitz: Strange, since logs were gone when I checked on my machine, but maybe I did something wrong.
Anyways let's go with your `init`.

geyslan: I'll be pushing changes soon. :+1: 

> Strange, since logs were gone when I checked on my machine, but maybe I did something wrong.

Could it be related to avoidance of output when a origin had already been triggered? Currently this approach only writes to output when the log origin is [new](https://github.com/aquasecurity/tracee/pull/2110/files#diff-c1dc2eb1e149a1dc7ecd5cb474e167fb456eb7ad0ea0256cf62fba693a89af79R192-R194):

```go
	_, file, line := getCallerInfo(skip + 1)
	if new := l.updateCounter(file, line); new {
		l.l.Debug(msg, fields...)
	}
```

If the origin has already been triggered, it just updates the counter.

pkg/logger/logger.go
```diff
@@ -0,0 +1,358 @@
...
+	fatal(1, l, msg, fields...)
+}
+
+// Sync
+func (l *Logger) Sync() error {
+	return l.l.Sync()
+}
+
+const (
+	LOGGER_LVL = "LOGGER_LVL"
```
NDStrahilevitz, line 244: Let's prefix this with TRACEE just to avoid any potential conflicts.

pkg/logger/logger.go
```diff
@@ -0,0 +1,358 @@
...
+	LOGGER_ENC = "LOGGER_ENC"
```
NDStrahilevitz, line 245: ditto

pkg/logger/logger.go
```diff
@@ -0,0 +1,358 @@
...
+)
+
+func getLoggerLevelFromEnv() (lvl Level) {
```
NDStrahilevitz, line 248: note: We will need docs on this.

cmd/tracee-ebpf/main.go
```diff
@@ -57,17 +58,35 @@ func main() {
...
 			debug := debug.Enabled()
 
+			// Logger Setup
+			if !logger.IsSetFromEnv() {
+				logger.Init(
+					&logger.LoggerConfig{
+						Writer: os.Stderr,
+						Level: func(debug bool) logger.Level {
+
+							if debug {
```
NDStrahilevitz, line 68: So this mean an environment variable will win over the debug flag. Also needs to be documented.

cmd/tracee-ebpf/main.go
```diff
@@ -57,17 +58,35 @@ func main() {
 			// for the rest of execution, use this debug mode value
 			debug := debug.Enabled()
```
NDStrahilevitz, line 59: note: eventually we want to move on from this.

NDStrahilevitz (commented):
Ok, practically this looks good to go ahead, please note my comments below.
Do you think we should replace all logging with this in the scope of this PR or move that to a later one? @rafaeldtinoco would also like your opinion on that.

josedonizetti:
Cool! Log FTW :)

Couple questions:

1 - can't we use SugaredLogger? It has good performance, with better syntax. Eg:
It doesn't seem like log performance is critical to use the strongly-typed approach. 

```
# this 

logger.Debug("os_info", "os_realease_field", helpers.OS_KERNEL_RELEASE, "OS_KERNEL_RELEASE", OSInfo.GetOSReleaseFieldValue(helpers.OS_KERNEL_RELEASE))

# intead of 

logger.Debug("os_info", logger.Uint32("os_realease_field", uint32(helpers.OS_KERNEL_RELEASE)), logger.String("OS_KERNEL_RELEASE", OSInfo.GetOSReleaseFieldValue(helpers.OS_KERNEL_RELEASE)))
```

2 - Why are we adding complexity of log counting? It was pointed to me this issue https://github.com/aquasecurity/tracee/issues/2036, and it seemed the reasoning was because the same error message getting logged multiple times?

geyslan:
> Ok, practically this looks good to go ahead, please note my comments below.

:+1: 

> Do you think we should replace all logging with this in the scope of this PR or move that to a later one?

I think we can make this PR only an introduction to logger package, demonstrating how to use it with changes to the main file. As it is right now (but I would rebase it and commit accordingly). But let me know.

Just more a couple questions.

- Currently, by default, it aggregates log origins (not the full logs) inside the LogCount. WDYT of changing it to optional making possible it to be set from env `TRACEE_LOGGER_AGREGATE=true` and through `LoggerConfig`?
- `logger.Dump()` is the way of get counts of log origins (aggregated). How do you plan on using this eventually? Triggered via signal?

geyslan:
> 1 - can't we use SugaredLogger? It has good performance, with better syntax. Eg: It doesn't seem like log performance is critical to use the strongly-typed approach.

Yeah, I suppose it's not an issue to use the sugared version.

In any case the sugared zap is in second place in [performance](https://github.com/uber-go/zap#performance).

image


> 2 - Why are we adding complexity of log counting? It was pointed to me this issue #2036, and it seemed the reasoning was because the same error message getting logged multiple times?

I guess @NDStrahilevitz and @AlonZivony can answer you about this more precisely.

NDStrahilevitz:
So about sugared logs - by design we shouldn't have noisy logs, so i'm ok with moving there.

@josedonizetti We want to keep the log counting to silence the noisy stuff and track how many occurred.

@geyslan I would like this PR at least to encompass the tracee-ebpf and tracee-rules cmd directories if that's not too much. We could do the packages in a later PR. However do note that by skipping that we may miss bugs in the logger now (especially with regard to the log counting).

josedonizetti:
@geyslan @NDStrahilevitz @rafaeldtinoco Would if would be okay if we simplified the scope here? 

1 - add a basic logger, and change both tracee-ebpf and tracee-rules to use it
2 - change the whole project to use a logger
3 - add log counting

It seems to me it is easier to review, understand the scope, and test things. We would also get a logger quicker to the project, allowing new PRs to already use it. WDYT?

NDStrahilevitz:
> @geyslan @NDStrahilevitz @rafaeldtinoco Wonder if it would be okay to simplifyy the scope here?
> 
> 1 - add a basic logger, and change both tracee-ebpf and tracee-rules main.go to use it 2 - change the whole project to use a logger 3 - add log counting
> 
> It seems to me it is easier to review, understand the scope, and test things. We would also get a logger quicker to the project, allowing new PRs to already use it. WDYT?

Good plan, let's keep this PR for point 1.

geyslan:
> Good plan, let's keep this PR for point 1.

Ok, as point 3 is almost 100% I'll do this embrace it as well.

note:@geyslan When returning to this PR, implement `TRACEE_LOGGER_AGGREGATE=true` and `LoggerConfig` logic, making aggregation (couting) not default.

rafaeldtinoco:
@geyslan would you mind rebasing this ?
@josedonizetti would you mind reviewing this ?
@NDStrahilevitz would you mind giving a +1 after those ^ ?

Thanks!

josedonizetti:
@geyslan Perhaps some of the design decisions here are due to requirements I don't know, let me know:

1 - why do support env variables to configure the logger? We are not using env variables for anything else on tracee. 
2 - why only env vars and not command line flags? 
3 - it is weird to me to couple the env variables under the log package, calling methods on `init`. This hides implementation details that I don't think are related to logging, the env vars value should be used to initialize the logger configuration which is passed as a dependency to the logger initialization once. 

```
main.go

/// config := getLogConfig() // checks for the env vars if it is not there, uses the default values
// log.Init(config) 
```

This gives the package a single initialization point and also decouples it from the env vars, which should only be used for configuration. Also avoids duplication, eg, every method checking for an env var must check and set `setFromEnv`,  and removes the need to check `IsSetFromEnv` whenever one is initializing the logger. 

WDYT?

geyslan:
> 1 - why do support env variables to configure the logger? We are not using env variables for anything else on tracee.

Tracee can be consumed as a library. https://github.com/aquasecurity/tracee/pull/2110#pullrequestreview-1088405734

> 2 - why only env vars and not command line flags?

Ditto.

> 3 - it is weird to me to couple the env variables under the log package, calling methods on init. This hides implementation details that I don't think are related to logging, the env vars value should be used to initialize the logger configuration which is passed as a dependency to the logger initialization once.

Foundations about the `init()` use are here https://github.com/aquasecurity/tracee/pull/2110#discussion_r957140226.

> This gives the package a single initialization point...

Yes, but doesn't guarantee that's the unique point calling logger.Init() and doesn't make it so crystal. About that topic I really don't know how to enforce it. The more we provide options we make it decentralized.

How about instead of this:

```go
			// Avoiding to override package-level logger
			// when it's already set by logger environment variables
			if !logger.IsSetFromEnv() {
				// Logger Setup
				logger.Init(
					&logger.LoggerConfig{
						Writer:    os.Stderr,
						Level:     logger.InfoLevel,
						Encoder:   logger.NewJSONEncoder(logger.NewProductionConfig().EncoderConfig),
						Aggregate: false,
					},
				)
			}
```

Do something like this?

```go
			logger.Init(
				&logger.LoggerConfig{
					Writer:    os.Stderr,
					Level:     logger.InfoLevel,
					Encoder:   logger.NewJSONEncoder(logger.NewProductionConfig().EncoderConfig),
					Aggregate: false,
					// Don't override package-level logger config
					// when it's already set by logger environment variables
					OverrideEnv: false,
				},
			)
```

Then, through a single flag we can set the logger (at any point) for reusing or not env vars.

josedonizetti:
> Tracee can be consumed as a library. [#2110 (review)](https://github.com/aquasecurity/tracee/pull/2110#pullrequestreview-1088405734)

Hmm, at least for me, if you consuming Tracee as a lib, you should set your own logger `logger.SetBase()` it will respect whatever your application is already using. One can bring their own env vars and config files. Parse them as they want not related to tracee library. Their app should handle their log initialization if they want a library. Because one consuming as a library could for example forget to set the env vars.

> > 2 - why only env vars and not command line flags?
> 
> Ditto.

But one doesn't cancel the other. A user now passes every configuration he wants tracee to do via command line, why for logs he would have to use an env variable? (but perhaps this should be another PR?)
 
> Yes, but doesn't guarantee that's the unique point calling logger.Init() and doesn't make it so crystal. About that topic I really don't know how to enforce it. The more we provide options we make it decentralized.
> 

We could if we want, for example, panic if the logger is already set for example. 

Well, it seems most of the requirements came from the product, which @NDStrahilevitz is more aware of. Probably best to get his review, besides the different options on the design, the code seems LGMT for me. :+1:

yanivagman:
> Hmm, at least for me, if you consuming Tracee as a lib, you should set your own logger `logger.SetBase()` it will respect whatever your application is already using. One can bring their own env vars and config files. Parse them as they want not related to tracee library. Their app should handle their log initialization. Because one consuming it as a library should be changing it programmatically. What if a library user forgets to set the env vars?

Agree with @josedonizetti here. Maybe @NDStrahilevitz can explain what he meant in his comment https://github.com/aquasecurity/tracee/pull/2110#pullrequestreview-1088405734?

NDStrahilevitz:
So about the flags or arguments "debate", we can have both, and i've noted in previous reviews that the flag isn't working anymore, perhaps I didn't make it clear that I _do_ want it to still function, just have the env flag taking priority. 

So about the initialization, my reasoning was as such:
1. Tracee has a whole bunch of required initialization logic, adding more adds difficulty in consuming
2. To make the logging simpler it's solution should implicitly initializing the logger as it is part of tracee's mechanism of reporting problems
3. If a consuming app is aware of this and wants to redirect the logs and change configuration behavior, they could still use something like `AddBase` but otherwise would not be concerned with it or have to even know of it.

I'm usually for not having implicit behavior, but in the case of logging I thought making an exception would be best so tracee can be initialized more smoothly, especially since we will likely use the global logger in other packages, and wanted it to work for other apps "out of the box".

I do agree we can make the initial logic "prettier" though, if we set implicitly via env variables on `logger.init()` we can just have a new config with cli arguments overriding it if they exist, with something like this similar to @josedonizetti's suggestion:
```
func main(c *cli.Ctx) {
    // ...logger is implicitly ready by this point
    config := getCliConfig(c)
    log.SetConfig(config)
}
```

josedonizetti (approved):

geyslan:
@NDStrahilevitz this PR version expired. I'm going to rebase it for you. Do you need something else?

NDStrahilevitz:
@geyslan I think this is good enough for now, i'll approve and let's merge after the rebase. About all the init logic and how I think it will become clearer after some usage what would be best. 
One nit that I have is that I think we should just do `Debug` instead of `Debugw` since it's not as if we have both to differentiate between having a wide version and not, but up to you.

NDStrahilevitz (approved):

Issue #2133, opened 2022-08-31
[REFAC] Decouple detect-syscall supporting events from syscall_data_t


NDStrahilevitz:
## Prerequisites

Select one OR another:

- [x] I have discussed the refactoring idea with one (or another) maintainer.
- [x] I'll create a PR to implement this refactoring idea (assign to yourself).
- [ ] Someone else should implement this (describe it well).

## Refactoring description

In order to further optimize sys_enter and sys_exit, we will want to disable it's logic running entirely for irrelevant syscalls.
However, some events depend on the syscall_data struct where the id is stored, which is done in the sys_enter probe.

In order to minimize runtime of these probes, we will move from getting the syscall id from the struct, instead getting it from the current task registries.

PR #2134, opened 2022-08-31
ebpf: add get_task_syscall_id helper

NDStrahilevitz:
## Initial Checklist

- [x] I have discussed this change with another maintainer.

## Description (git log)

```
Date:   Wed Aug 31 12:49:38 2022 +0000

    ebpf: add get_task_syscall_id helper
    
    This is used in events supporting the detect-syscall option, in order to
    decouple them from the syscall_data_t struct tracking mechanism.
    This decoupling is required for these events to function in the future
    without calling sys_enter and sys_exit logic for every syscall.
    security_file_open is intentionally left as is, because it will have
    syscall tail call dependencies in the near future.
```

Fixes: #2133

## Type of change

## How Has This Been Tested?

Test by running:
`./dist/tracee-ebpf -o option:detect-syscall -t e=task_rename`

Wait for an event output and confirm a valid syscall is printed.

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -1254,6 +1254,33 @@ static __always_inline bool is_compat(struct task_struct *task)
 #endif
 }
 
+static __always_inline int get_syscall_id_from_regs(struct pt_regs *regs)
+{
+#if defined(bpf_target_x86)
+    int id = READ_KERN(regs->orig_ax);
+#elif defined(bpf_target_arm64)
+    int id = READ_KERN(regs->syscallno);
+#endif
+    return id;
+}
+
+static __always_inline struct pt_regs *get_task_pt_regs(struct task_struct *task)
+{
+// THREAD_SIZE here is statistically defined and assumed to work for 4k page sizes.
+#if defined(bpf_target_x86)
+    void *__ptr = READ_KERN(task->stack) + THREAD_SIZE - TOP_OF_KERNEL_STACK_PADDING;
+    return ((struct pt_regs *) __ptr) - 1;
+#elif defined(bpf_target_arm64)
+    return ((struct pt_regs *) (THREAD_SIZE + READ_KERN(task->stack)) - 1);
+#endif
+}
+
+static __always_inline int get_task_syscall_id(struct task_struct *task)
+{
+    struct pt_regs *regs = get_task_pt_regs(task);
+    return get_syscall_id_from_regs(regs);
+}
+
 // HELPERS: VFS ------------------------------------------------------------------------------------
 
 static __always_inline struct dentry *get_mnt_root_ptr_from_vfsmnt(struct vfsmount *vfsmnt)
@@ -2803,7 +2830,6 @@ int tracepoint__raw_syscalls__sys_enter(struct bpf_raw_tracepoint_args *ctx)
     struct task_struct *task = (struct task_struct *) bpf_get_current_task();
 
     u32 task_id = bpf_get_current_pid_tgid(); // get the tid only
-    // use config as garbage value here
     task_info_t *task_info = init_task_info(task_id, NULL);
     if (task_info == NULL) {
         return 0;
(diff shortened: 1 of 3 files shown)


pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1254,6 +1254,32 @@ static __always_inline bool is_compat(struct task_struct *task)
 #endif
 }
 
+static __always_inline int get_syscall_id_from_regs(struct pt_regs *regs)
```
yanivagman, line 1257: Let's use this function in sys_exit as well, since we have the same code there too

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1254,6 +1254,32 @@ static __always_inline bool is_compat(struct task_struct *task)
...
+{
+#if defined(bpf_target_x86)
+    int id = READ_KERN(regs->orig_ax);
+#elif defined(bpf_target_arm64)
+    int id = READ_KERN(regs->syscallno);
+#endif
+    return id;
+}
+
+static __always_inline struct pt_regs *get_task_pt_regs(struct task_struct *task)
```
rafaeldtinoco, line 1267: Thats a very nice approach. I think it needs better documentation (I see this comes from task_pt_regs() from in kernel). To document why you're rounding up current kernel stack (I believe you're going after thread_info->task_struct and picking up pt_regs from task_struct, right ?).

We currently don't support other page sizes (other than 4k) but, if and when we do, we will need to calculate THREAD_SIZE dynamically.

rafaeldtinoco: @yanivagman thats actually a good thing, should we worry about different page sizes ? There are many servers with hugepages around there (specially those trying to avoid TLB lookup bottlenecks for data passing etc). Of course the majority is 4k page sized (tends to change in arm64 world at least).

yanivagman: Actually, this code (x86 part) isn't new. We had this code up until tracee v0.6.3, and removed it after adding syscall data stashing.
We need to return it now that we want to do some performance optimizations, like @NDStrahilevitz mentioned.



> I see this comes from task_pt_regs() from in kernel

That's correct

> We currently don't support other page sizes (other than 4k) but, if and when we do, we will need to calculate THREAD_SIZE dynamically.

Agree. We should document this here

yanivagman: > @yanivagman thats actually a good thing, should we worry about different page sizes ? There are many servers with hugepages around there (specially those trying to avoid TLB lookup bottlenecks for data passing etc). Of course the majority is 4k page sized (tends to change in arm64 world at least).

We can use https://pkg.go.dev/os#Getpagesize in userspace to initialize a corresponding value in the config map, then use it here. WDYT?

pkg/ebpf/c/tracee.bpf.c
NDStrahilevitz: Done.

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: Hum, that is a nice approach... hopefully verifiers will accept that. Im okay in having the 4k default for 0.8.2 (for performance improvements) and have an issue to get the page size for this (and possibly other sensitive points).

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1254,6 +1254,32 @@ static __always_inline bool is_compat(struct task_struct *task)
...
+{
+#if defined(bpf_target_x86)
+    void *__ptr = READ_KERN(task->stack) + THREAD_SIZE - TOP_OF_KERNEL_STACK_PADDING;
+    return ((struct pt_regs *) __ptr) - 1;
+#elif defined(bpf_target_arm64)
+    return ((struct pt_regs *) (THREAD_SIZE + READ_KERN(task->stack)) - 1);
```
yanivagman, line 1273: I think this won't compile with arm64 CO-RE as THREAD_SIZE is not defined in missing_definitions.h

yanivagman (changes requested):

pkg/ebpf/c/tracee.bpf.c
NDStrahilevitz: You're right it doesn't. Working backwards through the ARM definitions right now (with help from @rafaeldtinoco).

yanivagman (approved):
LGTM

Issue #2120, opened 2022-08-28
[REFAC] Optimize 


NDStrahilevitz:
## Prerequisites

Select one OR another:

- [x] I have discussed the refactoring idea with one (or another) maintainer.
- [x] I'll create PRs to implement this refactoring idea (assign to yourself).
- [ ] Someone else should implement this (describe it well).

## Refactoring description

`tracee-ebpf` has a bpf code hot path in it's `sys_enter` and `sys_exit` probes. These are used for:
1. Submitting syscall events
2. Tracking syscall in task information
3. Saving syscall args across bpf programs

This logic can take up over time a lot of kernel cpu time and as such, we can and should optimize it.
These can be done in the following incremental steps:

1. Making syscall event initialization conditional (done in #2080)
2. Remove map lookups where possible (kconfig map in `sys_enter` for example)
3. Moving syscall event initialization to a tailcall (possible optimization)
4. Optimize `init_event_data`
5.  Two steps here:
      a. Create separate syscall tracepoints for relevant events which need their argument stored
      b. Remove `sys_enter` and `sys_exit` as essential events, making them explicit dependencies for most syscall events

yanivagman:
`sys_enter` and `sys_exit` are no longer essential events in tracee, and we have seperate probes for some specific syscalls.

PR #2143, opened 2022-09-01
Refactor/full tail call enter exit

NDStrahilevitz:
## Initial Checklist

- [x] There is an issue describing the need for this PR (#2120).

## Description (git log)
Optimize sys_enter and sys_exit by making code parts conditional through tail calls.
Benchmarking main against this branch after a 5m runtime resulted in:

HEAD:
```
PROGRAM: sys_enter (runtime: 135407119 ns, amount: 306451 times, average: 441 ns)
PROGRAM: sys_exit (runtime: 127016300 ns, amount: 300287 times, average: 422 ns)
```

Branch:
```
PROGRAM: sys_enter (runtime: 65267123 ns, amount: 355574 times, average: 183 ns)
PROGRAM: sys_exit (runtime: 81635824 ns, amount: 355797 times, average: 229 ns)
```

further optimize by making entire sys_enter/exit logic conditional on a tail call
```
    ebpf: move sys_enter/exit logic to tail call
    
    This tail call is initialized from traced syscalls and required syscalls
    for other events to function.
    This is done to further reduce the time spent in these syscall probes.
```

optimize bpf code with indexed submit tailcall
```
    ebpf: index submit tailcall on syscall id
    
    Change the submit tailcall to index on id, and call it depending on the
    current syscall.
    Set the tailcall indices dynamically based on syscall events traced.
```

introduce tail call multi indexing
```
    events: change tailcall index to list
    
    This is done so tail calls can have indexes added to them dynamically.
```

## Type of change

## How Has This Been Tested?

Need to add tests for getTailCalls.

Tested manually with running tracee-ebpf without filters and checking that syscalls are outputted.

## Final Checklist:

## Git Log Checklist:

My commits logs have:


yanivagman (approved):
Nice work!
Overall LGTM, some nit comments

pkg/ebpf/tracee.go
```diff
@@ -819,17 +822,12 @@ func (t *Tracee) populateBPFMaps() error {
...
-			tailCalls[tailCall] = true
-		}
+	tailCalls, err := t.GetTailCalls()
+	if err != nil {
+		return err
 	}
-	for tailCall := range tailCalls {
-		err := t.initTailCall(tailCall.MapName, tailCall.MapIdx, tailCall.ProgName)
+	for _, tailCall := range tailCalls {
+		err := t.initTailCall(tailCall.MapName, tailCall.MapIndexes, tailCall.ProgName)
```
rafaeldtinoco, line 830: For near future: feels to me that, just like we did with probes, creating a package/interface for it, we could also have a small package for "tailcalls" and remove "initTailCall" from tracee object.

This would avoid having to have []events.TailCalls constructed and passed as an argument to a tracee object method.

pkg/ebpf/tracee.go
```diff
@@ -864,6 +905,16 @@ func (t *Tracee) attachProbes() error {
 		if !ok {
 			continue
 		}
+		if event.Syscall {
+			err := t.probes.Attach(probes.SyscallEnter__Internal)
```
rafaeldtinoco, line 909: I think its ok to call this over and over (as the implementations allow that, checking if link is attached before trying to attacha gain).

pkg/events/events.go
```diff
@@ -5034,6 +5028,12 @@ var Definitions = eventDefinitions{
 			Probes: []probeDependency{
 				{Handle: probes.SecurityMmapAddr, Required: true},
 				{Handle: probes.SecurityFileMProtect, Required: true},
+				{Handle: probes.SyscallEnter__Internal, Required: true},
+			},
+			Dependencies: dependencies{
+				TailCalls: []TailCall{
+					{MapName: "sys_enter_init_tail", MapIndexes: []uint32{uint32(Mmap), uint32(Mprotect)}, ProgName: "sys_enter_init"},
```
rafaeldtinoco, line 5035: So, apart from dynamically adding indexes for the syscalls being traced (done in initialization), we declare a "sys_enter_init" tailcall dependency to each event that REALLY DEPENDS on the syscall entry data (so if they are traced, but not their dependent syscall, we have the syscall tailcall as well).

And we can't depend on the syscall event (and add the tailcall there) because it would always include the tail call.. if I got this right.

Events that would depend on the init_tail are, for now: mem_prot_alert, most of, if not all, security events (file open, socket listen, socket connect, socket accept, socket bind, setsockopt, file mprotect)...

pkg/ebpf/tracee.go
```diff
@@ -864,6 +905,16 @@ func (t *Tracee) attachProbes() error {
...
+			if err != nil {
+				return err
+			}
+			err = t.probes.Attach(probes.SyscallExit__Internal)
+			if err != nil {
+				return err
+			}
+		}
```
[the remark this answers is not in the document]
rafaeldtinoco: https://github.com/aquasecurity/tracee/pull/2143#discussion_r964393875

Could be BUT Its ok to call multiple times, implementations check if link is attached.

pkg/ebpf/tracee.go
```diff
@@ -853,6 +851,49 @@ func (t *Tracee) populateBPFMaps() error {
 	return nil
 }
 
+func getTailCalls(eventConfigs map[events.ID]eventConfig) ([]events.TailCall, error) {
+	// Since tail calls with an array map index is incomprable type, it can't be the map key.
+	// Therefore we use a map of the tail call struct string forms and only append a tail call
```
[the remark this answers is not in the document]
rafaeldtinoco: yep, unclear to me as well

rafaeldtinoco:
This is a sensitive area to be touched and maintained. I would move the vanilla sys_enter()/sys_exit() tracing functions bellow all others touched and document its for a single sys_enter/sys_exit event (highlight that). I would also describe better what is being done on top of each tailcall function. In the tailcalls initialization function I would describe the overall idea of having indexes per syscall, describe the tailcall dependency logic for events depending on the syscall entry, etc etc (So others can maintain this without an extensive code analysis).

pkg/ebpf/tracee.go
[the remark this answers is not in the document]
NDStrahilevitz: I'll clarify it, basically explaining here why I changed the map key in the implementation of the function.

pkg/ebpf/tracee.go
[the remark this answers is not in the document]
NDStrahilevitz: Yep, Attach is agnostic to multiple calls as we implemeted it.

pkg/events/events.go
NDStrahilevitz: You got it right exactly, think we should add this as a comment before merging, where do you think it would fit?

If we depend on the event it would include the submit tail call too (because it would mark the event for full dependency loading), that's why we add the tail call dependency directly, avoiding the syscall event.

pkg/ebpf/tracee.go
NDStrahilevitz: I actually think tail calls can even be probes themselves since they technically satisfy the interface of Attach and Detach.
There's certainly room for thought here.

NDStrahilevitz:
@yanivagman @rafaeldtinoco just pushed with further comments explaining the changes, thanks for noting what wasn't clear. Please go through and confirm we're good with what's written and then we can merge :)

PR #2167, opened 2022-09-14
docs: add 'tracee-system' namespace to the manual installation

vtrduque:
Just for standardization, since the preferred way to install Tracee is using Helm I think the manual way could be the same


josedonizetti:
Duque, need you to sign  the CLA? :D

josedonizetti (approved):
Thanks!

PR #2170, opened 2022-09-19
[RFC][DRAFT] new network code

rafaeldtinoco:
This is a draft. There are some issues I'm already aware, so no full review should be done, please (not yet). BUT, feel free to get familiarized with this if you need/want.

```
$ sudo ./dist/tracee-ebpf --output format:json -t event=new_net_packet,new_dns_request -t comm=nc,ping
```

PS: DNS event is not actually a DNS event (yet), just testing things around.


pkg/ebpf/c/tracee.bpf.c
```diff
@@ -6278,4 +6253,786 @@ SEC("tc")
...
+CGROUP_SKB_HANDLE(proto_icmpv6);
+// clang-format on
+
+//
+// Network event submission functions
+//
+
+// clang-format off
+#define CGROUP_SKB_SUBMIT(name)                                                \
+static __always_inline u32 cgroup_skb_submit_##name(                           \
```
rafaeldtinoco, line 6779: I'll probably just pack these 4 variables (used in many hanlers) into a struct and pass the struct alongside them (instead of using macros to keep code more readable).

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -6278,4 +6253,786 @@ SEC("tc")
...
+    if (!sk) {
+        bpf_printk("ERROR: could not get full bpf_sock");
+        return 1;
+    }
+
+    iphdrs iphdrs = {0};
+    protohdrs protohdrs = {0};
+
+    // process protocols and needed events
+    return cgroup_skb_handle_family(ctx, neteventctx, &iphdrs, &protohdrs);
```
rafaeldtinoco, line 6858: I might use tail calls if needed (protocol numbers would be the index for the function/handler to call). Not needed right now (stack size is good and logic is simple).

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -6278,4 +6253,786 @@ SEC("tc")
...
+typedef union protohdrs_t {
+    struct tcphdr tcphdr;
+    struct udphdr udphdr;
+    struct icmphdr icmphdr;
+    struct icmp6hdr icmp6hdr;
+} protohdrs;
+
+// clang-format off
+#define CGROUP_SKB_HANDLE(name)                                                \
+static __always_inline u32 cgroup_skb_handle_##name(                           \
```
rafaeldtinoco, line 6756: ditto (about macros and, instead, just having a struct here to pack all variables passed alongside handlers).

rafaeldtinoco:
Full tracing (all processes) network probes do not seem to be causing slowdowns in a simple Gbit network:

image



image

rafaeldtinoco:
I'm closing this for a new cleaned/rebased PR.

Issue #2173, opened 2022-09-20
[BUG] Memory leak in the args_map


AlonZivony:
## Prerequisites

- [x] This affects latest released version.
- [x] This affects current development tree (origin/HEAD).
- [x] There isn't an issue describing the bug.

Select one OR another:

- [ ] I'm going to create a PR to solve this (assign to yourself).
- [x] Someone else should solve this.

## Bug description

I used the `register_chrdev` event and noticed that I don't get it sometimes, even though other events bounded to it in my code happen in deterministic way.
After some research I managed to print logs from various places in the event's hook implementation and noticed that it fails at loading the arguments:
![image](https://user-images.githubusercontent.com/87267148/191266902-854443b9-3f93-47cf-b239-8d9b8c015c9e.png)
The `-7` value returned from the `save_args` function  (`E2BIG` probably returned from [`array_map_update_elem`](https://elixir.bootlin.com/linux/v4.9/source/kernel/bpf/arraymap.c#L191) in the kernel) indicates that the `args_map` is full (which contains 1024 nodes).

I added some logs - on the `save_args` function and the `del_args` function - and checked if the values saved are deleted afterwards.
I was spammed with the `save_args` function logs, so I turned it off and checked if the `del_args` is called.
It isn't.

After going over the code of the main user of these functions, the `vfs_write` event, I saw the problem:
* We save the arguments upon entering the function (using a `kprobe`)
* We load them upon the function return (in the `kretprobe` hook)
* We do a lot of logic which might result returning
* We call the tail-call
* The tail call load the arguments
* The tail call delete the arguments from the map

It is easy to see the problem here - we don't clean properly the `args_map` in this hook, and maybe in other places.
We should think of a way to clean properly our maps.
Moreover, I got no indication about this error. We have no way to properly log errors occurred in the kernel space.

AlonZivony:
I will add that many `vfs_write` calls I saw got this error too, which resulted their failure and exit.
〣( º Δº )〣

PR #2181, opened 2022-09-21
pkg/ebpf: quick fix for args_map memory leak

AlonZivony:
## Initial Checklist

## Description (git log)

Make sure that all save_args paths call in the end to the del_args function to clean up the args_map BPF map.

Fixes: #2173

## Type of change

## How Has This Been Tested?

Using the same test that I found the bug with - great load of events and check if the `register_chrdev` is passed.

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -3517,6 +3517,7 @@ int syscall__accept4(void *ctx)
         // missed entry or not traced
         return 0;
     }
+    del_args(SOCKET_ACCEPT);
 
     event_data_t data = {};
     if (!init_event_data(&data, ctx))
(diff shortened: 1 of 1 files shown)


AlonZivony:
This is the working fix for the current problem.

I was delayed because of another unexpected leak - caused by failed calls to the `bpf_tail_call`.
This is interesting because we might have another places we don't expect the case in which the tail call is not called (here it happens because the `trace_ret_vfs_write_tail` function is called only if the capture write is selected).

Anyways my local tests which failed originally all pass now and my logs show that we aren't loosing any arguments now (which caused the program failure).

AlonZivony:
@yanivagman @rafaeldtinoco do you think we should also change the `args_map` to be a LRU?
It will reduce the change that a reoccurrence of this bug will make programs crash, but it will cost memory and performance.
It is not likely for the `args_map` to be full in real-life scenario without bugs. But it is possible.
WDYT?

rafaeldtinoco (approved):
LGTM

PR #2233, opened 2022-10-13
[REFACTOR] Cgroup Interface (cgroupv1 and cgroupv2 initialization)

rafaeldtinoco:
## Description (git log)

This is a refactor triggered by comment https://github.com/aquasecurity/tracee/pull/2200#discussion_r990781388. Tracee needs to always initialize cgroupv1 and/or groupv2 when needed. Cgroupv2 is always needed for cgroup ebpf programs.

## Type of change

## How Has This Been Tested?

```
sudo ./dist/tracee-ebpf --install-path /tmp/tracee --cache cache-type=mem --cache mem-cache-size=512 --output format:json --output option:parse-arguments --output option:detect-syscall --containers --trace container --trace comm=bash --trace follow --capabilities bypass=false | jq -c
```

Note: Set bypass to false so capabilities are applied:

```
# cat /proc/$(pidof tracee-ebpf)/status | grep Cap
CapInh: 0000000000000000
CapPrm: 000001ffffffffff
CapEff: 0000000000000000
CapBnd: 0000000000000000
CapAmb: 0000000000000000
```

### DIFF
--- a/pkg/cgroup/cgroup.go
+++ b/pkg/cgroup/cgroup.go
@@ -2,8 +2,6 @@ package cgroup
 
 import (
 	"bufio"
-	"errors"
-	"fmt"
 	"io/fs"
 	"os"
 	"path/filepath"
@@ -357,69 +355,6 @@ func GetCgroupControllerHierarchy(subsys string) (int, error) {
 	}
 
 	return value, nil
-}
-
-// RunsOnContainerV1 tests if process is running on a container in cgroupsv1 it
-// tests by checking for existance of a release_agent file in cpuset from
-// https://www.kernel.org/doc/Documentation/cgroup-v1/cgroups.txt:
-//
-// - release_agent: ...         (this file exists in the top cgroup only)
-//
-// WARNING: IF THIS WASN'T CLEAR THIS IS ONLY VALID FOR CGROUPSV1 MACHINES
-func RunsOnContainerV1() (bool, error) {
-	_, err := os.Stat(CgroupReleaseAgent)
-	if os.IsNotExist(err) {
-		return true, nil
-	}
-	if err != nil {
-		return false, err
-	}
-	return false, nil
-}
-
-// GetCgroupV1HierarchyId returns the hierarchyId of cgroupsv1
-func GetCgroupV1HierarchyId() (int, error) {
-	const cgroupsFile = "/proc/cgroups"
-	file, err := os.Open(cgroupsFile)
-	if err != nil {
-		return -1, fmt.Errorf("could not open cgroups file %s: %w", cgroupsFile, err)
-	}
-	defer file.Close()
-
-	scanner := bufio.NewScanner(file)
-	for i := 1; scanner.Scan(); i++ {
-		sline := strings.Fields(scanner.Text())
-		if len(sline) < 2 || sline[0] != CgroupDefaultController {
-			continue
-		}
-		intID, err := strconv.Atoi(sline[1])
-		if err != nil || intID < 0 {
-			return -1, fmt.Errorf("error parsing %s: %w", cgroupsFile, err)
-		}
-		return intID, nil
-	}
-	return -1, fmt.Errorf("couldn't determine cgroup v1 hierarchy id")
-}
-
-// MountCgroupV1Cpuset mounts cpuset cgroup controller to the container's
-// current working directory this is done in order to detect host cgroup data
-// when running tracee from a container in cgroupsv1 since the function uses the
-// mount syscall, it requires CAP_SYS_ADMIN
-func MountCgroupV1Cpuset(mountPath string) error {
-	mp, _ := mount.IsMountpoint(mountPath, CgroupV1FsType)
-
-	if mp {
-		// already mounted
-		return nil
-	}
-
-	// equivalent to mount -t cgroup -o cpuset cgroup mountPath
-	// we ignore EBUSY because it means we already mounted the cpuset
-	err := syscall.Mount("cgroup", mountPath, "cgroup", 0, "cpuset")
-	if err != nil && !errors.Is(err, syscall.EBUSY) {
-		return fmt.Errorf("failed to mount %s", err)
-	}
-	return nil
 }
 
 // GetCgroupPath walks the cgroup fs and provides the cgroup directory path of
--- a/pkg/cgroup/cgroup.go
+++ b/pkg/cgroup/cgroup.go
@@ -14,21 +14,360 @@ import (
 	"github.com/aquasecurity/tracee/pkg/mount"
 )
 
-const CgroupV1Controller = "cpuset"
-const CgroupV1FsType = "cgroup"
-const CgroupV2FsType = "cgroup2"
+// Constants
 
-const CgroupCpusetReleaseAgent = "/sys/fs/cgroup/cpuset/release_agent"
+const (
+	CgroupV1FsType          = "cgroup"
+	CgroupV2FsType          = "cgroup2"
+	CgroupDefaultController = "cpuset"
+	CgroupControllersFile   = "/sys/fs/cgroup/cgroup.controllers"
+	procCgroups             = "/proc/cgroups"
+	sysFsCgroup             = "/sys/fs/cgroup"
+)
+
+// Versions
+
+type CgroupVersion int
+
+func (v CgroupVersion) String() string {
+	switch v {
+	case CgroupVersion1:
+		return CgroupV1FsType
+	case CgroupVersion2:
+		return CgroupV2FsType
+	}
+
+	return ""
+}
+
+const (
+	CgroupVersion1 CgroupVersion = iota
+	CgroupVersion2
+)
+
+//
+// Cgroups
+//
+
+type Cgroups struct {
+	cgroupv1 Cgroup
+	cgroupv2 Cgroup
+	cgroup   *Cgroup // pointer to default cgroup version
+	hid      int     // default cgroup controller hiearchy ID
+}
+
+func NewCgroups() (*Cgroups, error) {
+	var err error
+	var cgrp *Cgroup
+	var cgroupv1, cgroupv2 Cgroup
+
+	// discover the default cgroup being used
+	defaultVersion, err := GetCgroupDefaultVersion()
+	if err != nil {
+		return nil, err
+	}
+
+	// only start cgroupv1 if it is the OS default (orelse it isn't needed)
+	if defaultVersion == CgroupVersion1 {
+		cgroupv1, err = NewCgroup(CgroupVersion1)
+		if err != nil {
+			if _, ok := err.(*VersionNotSupported); !ok {
+				return nil, err
+			}
+		}
+	}
+
+	// start cgroupv2 (if supported)
+	cgroupv2, err = NewCgroup(CgroupVersion2)
+	if err != nil {
+		if _, ok := err.(*VersionNotSupported); !ok {
+			return nil, err
+		}
+	}
+
+	// at least one (or both) has to be supported
+	if cgroupv1 == nil && cgroupv2 == nil {
+		return nil, NoCgroupSupport()
+
+	}
+
+	hid := 0
+
+	// adjust pointer to the default cgroup version
+	switch defaultVersion {
+	case CgroupVersion1:
+		if cgroupv1 == nil {
+			return nil, CouldNotFindOrMountDefaultCgroup(CgroupVersion1)
+		}
+		cgrp = &cgroupv1
+
+		// discover default cgroup controller hierarchy id for cgroupv1
+		hid, err = GetCgroupControllerHierarchy(CgroupDefaultController)
+		if err != nil {
+			return nil, err
+		}
+
+	case CgroupVersion2:
+		if cgroupv2 == nil {
+			return nil, CouldNotFindOrMountDefaultCgroup(CgroupVersion2)
+		}
+		cgrp = &cgroupv2
+	}
+
+	cs := &Cgroups{
+		cgroupv1: cgroupv1,
+		cgroupv2: cgroupv2,
+		cgroup:   cgrp,
+		hid:      hid,
+	}
+
+	return cs, nil
+}
+
+func (cs *Cgroups) Destroy() error {
+	if cs.cgroupv1 != nil {
+		err := cs.cgroupv1.destroy()
+		if err != nil {
+			return err
+		}
+	}
+	if cs.cgroupv2 != nil {
+		return cs.cgroupv2.destroy()
+	}
+
+	return nil
+}
+
+func (cs *Cgroups) GetDefaultCgroupHierarchyID() int {
+	return cs.hid
+}
+
+func (cs *Cgroups) GetDefaultCgroup() Cgroup {
+	return *cs.cgroup
+}
+
+//
+// Cgroup
+//
+
+type Cgroup interface {
+	init() error
+	destroy() error
+	GetMountPoint() string
+	getDefaultHierarchyID() int
+}
+
+func NewCgroup(ver CgroupVersion) (Cgroup, error) {
+	var c Cgroup
+
+	switch ver {
+	case CgroupVersion1:
+		c = &CgroupV1{}
+	case CgroupVersion2:
+		c = &CgroupV2{}
+	}
+
+	return c, c.init()
+}
+
+// cgroupv1
+
+type CgroupV1 struct {
+	mounted    *mount.MountOnce
+	mountpoint string
+	hid        int
+}
+
+func (c *CgroupV1) init() error {
+
+	// 0. check if cgroup type is supported
+	supported, err := mount.IsFileSystemSupported(CgroupVersion1.String())
+	if err != nil {
+		return err
+	}
+	if !supported {
+		return &VersionNotSupported{}
+	}
+
+	// 1. mount cgroup (if needed)
+	c.mounted, err = mount.NewMountOnce(
+		CgroupV1FsType,
+		CgroupV1FsType,
+		CgroupDefaultController,
+		sysFsCgroup, // where to check for already mounted cgroupfs
+	)
+	if err != nil {
+		return err
+	}
+
+	// 2. discover where cgroup is mounted
+	c.mountpoint = c.mounted.GetMountpoint()
+
+	return nil
+}
+
+func (c *CgroupV1) destroy() error {
+	return c.mounted.Umount()
+}
+
+func (c *CgroupV1) GetMountPoint() string {
+	return c.mountpoint
+}
+
+func (c *CgroupV1) getDefaultHierarchyID() int {
+	return c.hid
+}
+
+// cgroupv2
+
+type CgroupV2 struct {
+	mounted    *mount.MountOnce
+	mountpoint string
+	hid        int
+}
+
+func (c *CgroupV2) init() error {
+	// 0. check if cgroup type is supported
+	supported, err := mount.IsFileSystemSupported(CgroupVersion2.String())
+	if err != nil {
+		return err
+	}
+	if !supported {
+		return &VersionNotSupported{}
+	}
+
+	// 1. mount cgroup (if needed)
+	c.mounted, err = mount.NewMountOnce(
+		CgroupV2FsType,
+		CgroupV2FsType,
+		"",          // cgroupv2 has no default controller
+		sysFsCgroup, // where to check for already mounted cgroupfs
+	)
+	if err != nil {
+		return err
+	}
+
+	// 2. discover where cgroup is mounted
+	c.mountpoint = c.mounted.GetMountpoint()
+
+	return nil
+}
+
+func (c *CgroupV2) destroy() error {
+	return c.mounted.Umount()
+}
+
+func (c *CgroupV2) GetMountPoint() string {
+	return c.mountpoint
+}
+
+func (c *CgroupV2) getDefaultHierarchyID() int {
+	return c.hid
+}
+
+//
+// General
+//
+
+func GetCgroupDefaultVersion() (CgroupVersion, error) {
+	// 1st Method: already mounted cgroupv1 filesystem
+
+	if ok, _ := IsCgroupV2MountedAndDefault(); ok {
+		return CgroupVersion2, nil
+	}
+
+	//
+	// 2nd Method: From cgroup man page:
+	// ...
+	// 2. The unique ID of the cgroup hierarchy on which this
+	//    controller is mounted. If multiple cgroups v1
+	//    controllers are bound to the same hierarchy, then each
+	//    will show the same hierarchy ID in this field.  The
+	//    value in this field will be 0 if:
+	//
+	//    a) the controller is not mounted on a cgroups v1
+	//       hierarchy;
+	//    b) the controller is bound to the cgroups v2 single
+	//       unified hierarchy; or
+	//    c) the controller is disabled (see below).
+	// ...
+
+	var value int
+
+	file, err := os.Open(procCgroups)
+	if err != nil {
+		return -1, CouldNotOpenFile(procCgroups, err)
+	}
+	defer file.Close()
+
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := strings.Fields(scanner.Text())
+		if line[0] != CgroupDefaultController {
+			continue
+		}
+		value, err = strconv.Atoi(line[1])
+		if err != nil || value < 0 {
+			return -1, ErrorParsingFile(procCgroups, err)
+		}
+	}
+
+	if value == 0 { // == (a), (b) or (c)
+		return CgroupVersion2, nil
+	}
+
+	return CgroupVersion1, nil
+}
+
+// IsCgroupV2MountedAndDefault tests if cgroup2 is mounted and is the default
+// cgroup version being used by the running environment. It does so by checking
+// the existance of a "cgroup.controllers" file in default cgroupfs mountpoint.
+func IsCgroupV2MountedAndDefault() (bool, error) {
+	_, err := os.Stat(CgroupControllersFile)
+	if os.IsNotExist(err) {
+		return false, nil
+	}
+	if err != nil {
+		return false, CouldNotOpenFile(CgroupControllersFile, err)
+	}
+
+	return true, nil
+}
+
+// Returns a cgroup controller hierarchy value
+func GetCgroupControllerHierarchy(subsys string) (int, error) {
+	var value int
+
+	file, err := os.Open(procCgroups)
+	if err != nil {
+		return -1, CouldNotOpenFile(procCgroups, err)
+	}
+	defer file.Close()
+
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := strings.Fields(scanner.Text())
+		if len(line) < 2 || line[0] != subsys {
+			continue
+		}
+		value, err = strconv.Atoi(line[1])
+		if err != nil || value < 0 {
+			return -1, ErrorParsingFile(procCgroups, err)
+		}
+	}
+
+	return value, nil
+}
 
 // RunsOnContainerV1 tests if process is running on a container in cgroupsv1 it
 // tests by checking for existance of a release_agent file in cpuset from
 // https://www.kernel.org/doc/Documentation/cgroup-v1/cgroups.txt:
 //
-// - release_agent: ...	 (this file exists in the top cgroup only)
+// - release_agent: ...         (this file exists in the top cgroup only)
 //
 // WARNING: IF THIS WASN'T CLEAR THIS IS ONLY VALID FOR CGROUPSV1 MACHINES
 func RunsOnContainerV1() (bool, error) {
-	_, err := os.Stat(CgroupCpusetReleaseAgent)
+	_, err := os.Stat(CgroupReleaseAgent)
 	if os.IsNotExist(err) {
 		return true, nil
 	}
@@ -50,7 +389,7 @@ func GetCgroupV1HierarchyId() (int, error) {
 	scanner := bufio.NewScanner(file)
 	for i := 1; scanner.Scan(); i++ {
 		sline := strings.Fields(scanner.Text())
-		if len(sline) < 2 || sline[0] != CgroupV1Controller {
+		if len(sline) < 2 || sline[0] != CgroupDefaultController {
 			continue
 		}
 		intID, err := strconv.Atoi(sline[1])
--- a/pkg/cgroup/errors.go
+++ b/pkg/cgroup/errors.go
@@ -1,0 +1,26 @@
+package cgroup
+
+import "fmt"
+
+type VersionNotSupported struct{}
+
+func (c *VersionNotSupported) Error() string {
+	return "unsupported cgroup version"
+}
+
+func NoCgroupSupport() error {
+	return fmt.Errorf("could not find cgroup support")
+}
+
+func CouldNotFindOrMountDefaultCgroup(ver CgroupVersion) error {
+	return fmt.Errorf("could not find/mount default %v support", ver.String())
+}
+
+func ErrorParsingFile(file string, err error) error {
+	return fmt.Errorf("error parsing %s: %w", file, err)
+}
+
+func CouldNotOpenFile(file string, err error) error {
+	return fmt.Errorf("could not open %s: %w", file, err)
+
+}
--- a/pkg/containers/containers.go
+++ b/pkg/containers/containers.go
@@ -17,16 +17,12 @@ import (
 	"github.com/aquasecurity/libbpfgo"
 	"github.com/aquasecurity/tracee/pkg/cgroup"
 	cruntime "github.com/aquasecurity/tracee/pkg/containers/runtime"
-	"github.com/aquasecurity/tracee/pkg/mount"
 )
-
-var cgroupV1HierarchyID int
 
 // Containers contains information about running containers in the host.
 type Containers struct {
-	cgroupV1   bool
-	cgroupMP   string
-	cgroups    map[uint32]CgroupInfo
+	cgroups    *cgroup.Cgroups
+	cgroupsMap map[uint32]CgroupInfo
 	deleted    []uint64
 	mtx        sync.RWMutex // protecting both cgroups and deleted fields
 	enricher   runtimeInfoService
@@ -44,19 +40,20 @@ type CgroupInfo struct {
 
 // New initializes a Containers object and returns a pointer to it.
 // User should further call "Populate" and iterate with Containers data.
-func New(sockets cruntime.Sockets, mapName string, debug bool) (*Containers, error) {
+func New(
+	cgroups *cgroup.Cgroups,
+	sockets cruntime.Sockets,
+	mapName string,
+	debug bool,
+) (
+	*Containers,
+	error,
+) {
 	containers := &Containers{
-		cgroupV1:   false,
-		cgroupMP:   "",
-		cgroups:    make(map[uint32]CgroupInfo),
+		cgroups:    cgroups,
+		cgroupsMap: make(map[uint32]CgroupInfo),
 		mtx:        sync.RWMutex{},
 		bpfMapName: mapName,
-	}
-
-	if _, err := os.Stat("/sys/fs/cgroup/cgroup.controllers"); os.IsNotExist(err) {
-		if err := containers.initCgroupV1(); err != nil {
-			return nil, err
-		}
 	}
 
 	runtimeService := RuntimeInfoService(sockets)
@@ -82,105 +79,20 @@ func New(sockets cruntime.Sockets, mapName string, debug bool) (*Containers, err
 
 // Close executes cleanup logic for Containers object
 func (c *Containers) Close() error {
-	// if we are on cgroupv1 and previously executed cpuset mounting logic (see function initCgroupV1)
-	if c.IsCgroupV1() && strings.HasPrefix(c.cgroupMP, "/tmp") {
-		// first unmount the temporary cpuset mount
-		err := syscall.Unmount(c.cgroupMP, 0)
-		if err != nil {
-			return err
-		}
-
-		// then remove the dir
-		err = os.RemoveAll(c.cgroupMP)
-		if err != nil {
-			return err
-		}
-	}
-
 	return nil
 }
 
-func (c *Containers) IsCgroupV1() bool {
-	return c.cgroupV1
-}
-
-func (c *Containers) GetCgroupV1HID() int {
-	return cgroupV1HierarchyID
+func (c *Containers) GetDefaultCgroupHierarchyID() int {
+	return c.cgroups.GetDefaultCgroupHierarchyID()
 }
 
 // Populate populates Containers struct by reading mounted proc and cgroups fs.
 func (c *Containers) Populate() error {
-	if c.cgroupMP == "" {
-		err := c.findCgroupMounts()
-		if err != nil {
-			return err
-		}
-	}
-
 	return c.populate()
 }
 
-// getCgroupV1SSID finds cgroup v1 hierarchy ID.
-func (c *Containers) initCgroupV1() error {
-	c.cgroupV1 = true
-	hierarchyId, err := cgroup.GetCgroupV1HierarchyId()
-	if err != nil {
-		return err
-	}
-	cgroupV1HierarchyID = hierarchyId
-
-	inContainer, err := cgroup.RunsOnContainerV1()
-	if err != nil {
-		fmt.Fprintf(os.Stderr, "containers: failed to detect if running on a cgroupv1 container: %s", err)
-		return nil
-	}
-
-	if inContainer {
-		path, err := os.MkdirTemp("/tmp", "tracee-cpuset")
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "failed to create cpuset directory %s\n", err)
-			return nil
-		}
-
-		mountPath, err := filepath.Abs(path)
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "failed to get absolute path of cputset mountpoint %s\n", mountPath)
-			return nil
-		}
-
-		err = cgroup.MountCgroupV1Cpuset(mountPath)
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "containers: failed to mount cgroupv1 controller: %s\n", err)
-			return nil
-		}
-
-		// now that we know what the mountPath is we can just set it
-		c.cgroupMP = mountPath
-	}
-
-	return nil
-}
-
-// findCgroupMounts finds cgroups v1 and v2 mountpoints.
-func (c *Containers) findCgroupMounts() error {
-	fsType := cgroup.CgroupV2FsType
-	search := ""
-	if c.cgroupV1 {
-		fsType = cgroup.CgroupV1FsType
-		search = "cpuset"
-	}
-
-	mp, err := mount.SearchMountpoint(fsType, search)
-
-	if err != nil {
-		return err
-	}
-
-	c.cgroupMP = mp
-	return nil
-}
-
-// gets a containerID from a given task or process directory path
+// GetContainerIdFromTaskDir gets a containerID from a given task or process
+// directory path
 func GetContainerIdFromTaskDir(taskPath string) (string, error) {
 	containerId := ""
 	taskPath = fmt.Sprintf("%s/cgroup", taskPath)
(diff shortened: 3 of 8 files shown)


rafaeldtinoco:
@NDStrahilevitz enrichment is still failing (I missed something but likely easy to fix) but I would like your opinion in this change (about the refactoring), considering comments from @yanivagman saying we should mount cgroupv2 (discussion at: https://github.com/aquasecurity/tracee/pull/2200#discussion_r990781388).

pkg/ebpf/tracee.go
```diff
@@ -328,14 +335,57 @@ func (t *Tracee) Init() error {
...
+			return err // cgroupv1 is unavailable when it should be
+		}
+		defaultCgroup = t.cgroupv1 // default is cgroupv1
+	}
+
+	isCgroupV2, err := cgroup.IsCgroupV2Supported()
+	if err != nil {
+		return err
+	}
+	if !isCgroupV2 {
```
NDStrahilevitz, line 368: Why should we fail here?
Are we simply dropping support for all non cgroupv1 systems?

pkg/ebpf/tracee.go
```diff
@@ -328,14 +335,57 @@ func (t *Tracee) Init() error {
...
+		return fmt.Errorf("cgroupv2 is not supported in this environment")
+	}
+
+	t.cgroupv2 = cgroup.InitCgroupV2()
+	err = t.cgroupv2.Initialize()
+	if err != nil {
+		return err // cgroupv2 is unavailable (it needs to be for cgroup progs)
```
NDStrahilevitz, line 375: I don't think we should fail here if we don't use the cgroup progs. It's one thing not to support network events on those systems, another entirely to just drop tracee as a whole.

pkg/ebpf/tracee.go
```diff
@@ -1251,6 +1322,20 @@ func (t *Tracee) Close() {
 			fmt.Fprintf(os.Stderr, "failed to clean containers module when closing tracee: %s", err)
 		}
 	}
+
+	if t.cgroupv1 != nil {
+		err := t.cgroupv1.Destroy()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "failed to clean cgroupv1: %s", err)
```
NDStrahilevitz, line 1329: There's a logger now, let's use it for new code. If I missed other prints in the code it goes for that as well.

pkg/ebpf/tracee.go
```diff
@@ -328,14 +335,57 @@ func (t *Tracee) Init() error {
...
-	t.containers, err = containers.New(t.config.Sockets, "containers_map", t.config.Debug)
+	// Initialize cgroup filesystems
+
+	var defaultCgroup cgroup.Cgroup
+
+	isCgroupV1, err := cgroup.IsCgroupV1Default()
+	if err != nil {
+		return err
+	}
+	t.cgroupv1 = nil
```
NDStrahilevitz, line 354: No need, this should start as nil by default.

pkg/ebpf/tracee.go
```diff
@@ -328,14 +335,57 @@ func (t *Tracee) Init() error {
...
+	if t.cgroupv1 == nil {
+		defaultCgroup = t.cgroupv2 // default is cgroupv2
+	}
+
+	// Initialize containers
+
+	t.containers, err = containers.New(
+		defaultCgroup, // cgroupv1 or cgroupv2
+		t.config.Sockets,
+		"containers_map",
```
NDStrahilevitz, line 386: I think there's a defined const for this now once you rebase on main.

pkg/cgroup/cgroup.go
```diff
@@ -13,36 +13,167 @@ import (
...
 	if err != nil {
-		return -1, fmt.Errorf("could not open cgroups file %s: %w", procCgroups, err)
+		return fmt.Errorf("cgroup: failed to create temporary dir: %s", err)
+	}
+	mountPath, err := filepath.Abs(path)
+	if err != nil {
+		return fmt.Errorf("cgroup: failed to get absolute path: %s", path)
+	}
+	mp, _ := mount.IsMountpointOf(mountPath, CgroupV1FsType)
+	if mp {
```
NDStrahilevitz, line 46: If we're already mounted I think we can just return successful no?

pkg/cgroup/cgroup.go
```diff
@@ -13,36 +13,167 @@ import (
...
+	path, err := os.MkdirTemp(TmpCgroup, TmpCgroupV2Path)
+	if err != nil {
+		return fmt.Errorf("cgroup: failed to create temporary dir: %s", err)
+	}
+	mountPath, err := filepath.Abs(path)
+	if err != nil {
+		return fmt.Errorf("cgroup: failed to get absolute path: %s", path)
+	}
+	mp, _ := mount.IsMountpointOf(mountPath, CgroupV2FsType)
+	if mp {
```
NDStrahilevitz, line 111: ditto

pkg/cgroup/cgroup.go
```diff
@@ -80,63 +211,81 @@ func GetCgroupPath(rootDir string, cgroupId uint64, subPath string) (string, err
...
-	_, err := os.Stat(releseAgentPath)
-	if os.IsNotExist(err) {
-		return true, nil
-	}
-	if err != nil {
-		return false, err
-	}
-	return false, nil
-}
+// // UmountCgroupV1Cpuset umounts cpuset cgroup controller from a temporary place
```
NDStrahilevitz, line 214: Needs cleanup I assume.

pkg/ebpf/probes/probes.go
```diff
@@ -2,48 +2,17 @@ package probes
```
NDStrahilevitz, line 2: Love the split, thanks for taking care of that.

NDStrahilevitz (commented):
Awesome refactor Rafael, I really like the new cgroup package split.
Left some questions and notes, specifically the cgroupv2 requirement concerns me since this complete drop of support may leave some customers in a problem (this is relevant to the networking PR as well since we should fail the probes on unsupported systems).

pkg/ebpf/tracee.go
rafaeldtinoco: Yep, I had doubts if I should just start using it for all the fprintfs around when refactoring, I believe we should as well.

pkg/ebpf/tracee.go
rafaeldtinoco: I wanted to make sure (when testing =D).

pkg/cgroup/cgroup.go
rafaeldtinoco: It should be mounted only once per temporary tmp directory (with no recurrent callings IMO). Lets see the final version and we can change this if needed.

rafaeldtinoco:
> I do think it would be a good idea to just merge this before the networking PR, cherry-pick them on top of main, and then rebase the network PR, because while I tried going through it commit by commit, it would be easier if I had it clean from the netowrk PR diffs.

Yes, I know it got polluted, as the network code is a bit big. Will work in merging this one first (right on it).

rafaeldtinoco:
The test is failing in RHEL8 because I believe we currently have a bug at the capabilities logic.

So we are keeping CAP_SYS_ADMIN in NEW kernels (shouldn't it be dropped ?):

image

AND for OLD kernels (that seems correct):

image

and dropping it for RHEL8 only (this is correct, but by accident):

image

I believe the capabilities feature is currently broken.

rafaeldtinoco:
This PR is held until https://github.com/aquasecurity/tracee/pull/2243 is merged.

pkg/ebpf/tracee.go
rafaeldtinoco: Can't find...

pkg/ebpf/tracee.go
rafaeldtinoco: I would like this change to be made for all fprintfs and outside the scope of this PR.

rafaeldtinoco:
This PR is built on top of https://github.com/aquasecurity/tracee/pull/2243 (which needs to be merged first). That PR takes care of the "new capabilities" feature (privilege rings) and this one uses it for mounting and unmounting cgroupv1 and cgroupv2 during tracee startup/shutdown time.

rafaeldtinoco:
This should be reviewed before https://github.com/aquasecurity/tracee/pull/2200 because that PR needs this one first (and contains the same commits for the time being).

pkg/cgroup/cgroup.go
```diff
@@ -0,0 +1,292 @@
...
+	"path/filepath"
+	"strconv"
+	"strings"
+	"syscall"
+
+	"github.com/aquasecurity/tracee/pkg/mount"
+)
+
+type Cgroup interface {
+	initialize() error
```
NDStrahilevitz, line 17: Since interfaces deal with what is publicly available, it's a bit unnecessary to have `initalize` here, and i'll expand further about something that may be simpler to do.

pkg/containers/containers.go
```diff
@@ -15,19 +15,15 @@ import (
...
-const cgroupV1Controller = "cpuset"
-const cgroupV1FsType = "cgroup"
-const cgroupV2FsType = "cgroup2"
+var cgroupDefaultControllerHierarchyID int
 
 // Containers contains information about running containers in the host.
 type Containers struct {
-	cgroupV1   bool
-	cgroupMP   string
+	defCgroup  cgroup.Cgroup
```
NDStrahilevitz, line 26: nit: think just `cgroup` is enough

pkg/containers/containers.go
```diff
@@ -46,20 +42,23 @@ type CgroupInfo struct {
 
 // New initializes a Containers object and returns a pointer to it.
 // User should further call "Populate" and iterate with Containers data.
-func New(sockets cruntime.Sockets, mapName string, debug bool) (*Containers, error) {
+func New(
+	defCgroup cgroup.Cgroup,
```
NDStrahilevitz, line 46: ditto, the "default" concept should be irrelevant for the reading here (and we can make the "default" logic a bit simpler IMO i'll expand upon it further in the review).

pkg/ebpf/tracee.go
```diff
@@ -408,17 +413,62 @@ func (t *Tracee) Init() error {
...
 	}
 
-	// Initialize containers enrichment logic
+	// Initialize cgroup filesystems
 
-	capabilities.GetInstance().Requested(func() error { // TODO: workaround until PR: #2233 is in place
+	var defaultCgroup cgroup.Cgroup
 
-		t.containers, err = containers.New(t.config.Sockets, "containers_map", t.config.Debug)
+	err = caps.Requested(func() error { // ring2
```
NDStrahilevitz, line 420: So I think all the nested logic here can go into a function in the new `cgroup` package of `cgroup.New()` that will return an instance of the `Cgroup` interface, that way we may be able to not have two instances of cgroup at the same time on tracee's struct level.
It might be necessary to add a "combined" v1+v2 cgroup struct impl for this, in which case it may be better to just leave it be.

pkg/cgroup/cgroup.go
```diff
@@ -0,0 +1,292 @@
...
+	Destroy() error
+	GetMountPoint() string
+	GetDefaultHierarchy() int
+	IsV1() bool
+}
+
+// cgroupv1
+
+func InitCgroupV1() (Cgroup, error) {
```
NDStrahilevitz, line 26: So I think it would be best if we had here some `func New() (Cgroup, error)` and it would be responsible for decision making of which cgroup implementation to return, and that would be the preferred way to construct a cgroup object.

pkg/cgroup/cgroup.go
```diff
@@ -0,0 +1,292 @@
...
+
+	"github.com/aquasecurity/tracee/pkg/mount"
+)
+
+type Cgroup interface {
+	initialize() error
+	Destroy() error
+	GetMountPoint() string
+	GetDefaultHierarchy() int
+	IsV1() bool
```
NDStrahilevitz, line 21: If we go for the approach I suggested, we may need to also add `IsV2() bool` for the combined implementation (if it's a necessary one).

pkg/cgroup/cgroup.go
NDStrahilevitz: That way in tracee.go itself there will be minimal exposure of internal decision making of which cgroups we need, just `t.cgroup, err := cgroup.New()` (obviously inside a `capabilities.Required` callback).

NDStrahilevitz (approved):
I don't see any obvious issues here, only substantial comments I have are about some naming and perhaps a different "construction" structure for the cgroup module, up to you if you find it better, but LGTM for merging, not going to block on style.

pkg/cgroup/cgroup.go
```diff
@@ -0,0 +1,292 @@
...
+
+	// mount -t cgroup -o cpuset cgroup mountPath
+	err = syscall.Mount(CgroupV1FsType, mountPath, CgroupV1FsType, 0, CgroupV1Controller)
+	if err != nil {
+		return failMountCgroup(1, err)
+	}
+
+	c.mountPath = mountPath
+
+	c.hierarchy, err = GetCgroupControllerHierarchy("cpuset")
```
yanivagman, line 62: Let's not make the hierarchy "cpuset" hardcoded

pkg/cgroup/cgroup.go
```diff
@@ -0,0 +1,292 @@
...
+	err := c.initialize()
+	return c, err
+}
+
+type cgroupV2 struct {
+	mountPath string
+	hierarchy int
+}
+
+func (c *cgroupV2) initialize() error {
```
yanivagman, line 109: This function looks the same as the cgroup v1 initialize function. I think it can be generalized to take relevant parameters and handle both to avoid code duplication

pkg/cgroup/cgroup.go
```diff
@@ -0,0 +1,292 @@
...
+
+	// equivalent to mount -t cgroup2 cgroup2 mountPath
+	err = syscall.Mount(CgroupV2FsType, mountPath, CgroupV2FsType, 0, "")
+	if err != nil {
+		return failMountCgroup(2, err)
+	}
+
+	c.mountPath = mountPath
+
+	c.hierarchy, err = GetCgroupControllerHierarchy("cpuset")
```
yanivagman, line 134: Let's avoid hardcoding "cpuset"

pkg/cgroup/cgroup.go
```diff
@@ -0,0 +1,292 @@
...
+	if err != nil {
+		return failGetCgroupHierarchy(2, err)
+	}
+
+	return nil
+}
+
+func (c *cgroupV2) Destroy() error {
```
yanivagman, line 142: Again, this function is unnecessarily duplicated of cgroup v1

pkg/cgroup/cgroup.go
```diff
@@ -0,0 +1,292 @@
...
+// Checks if cgroupv1 is the default cgroup filesystem
+func IsCgroupV1Default() (bool, error) {
+	var err [3]error
+	var cpu, cpuset, memory int
+
+	cpu, err[0] = GetCgroupControllerHierarchy("cpu")
+	cpuset, err[1] = GetCgroupControllerHierarchy("cpuset")
+	memory, err[2] = GetCgroupControllerHierarchy("memory")
+
+	for _, e := range err {
```
yanivagman, line 262: It can be (although less common) that not all cgroup hierarchies are compiled into the kernel, but here you assume they always are.

yanivagman (changes requested):
I will finish reviewing this one later, but wanted to post some comments I already have

pkg/cgroup/cgroup.go
rafaeldtinoco: Yes, but I'm assuming the 3 most important ones are always available. I believe they are in all environments we currently test tracee,

rafaeldtinoco:
@NDStrahilevitz could you check the changes I made ? I believe I have addressed what you suggested. I preferred to use a "Cgroups" object to contain both "Cgroup" objects (cgroupv1 and cgroupv2). This way the initialization from the caller perspective is easier (and the cleanup as well).

Let me know if you have any other comments pls.

Note: I believe I address all Yaniv's comments in this version as well (mostly were about code duplication, which was removed in this new commit).

pkg/cgroup/cgroup.go
```diff
@@ -0,0 +1,312 @@
...
+		if e != nil {
+			return false, e
+		}
+	}
+
+	if cpu == cpuset && cpu == memory {
```
josedonizetti, line 288: @rafaeldtinoco how can `cpu` be `cpuset` and `memory` at the same time? 
```suggestion
	if cpu == cpuset && cpu == memory {
```

pkg/cgroup/cgroup.go
```diff
@@ -0,0 +1,312 @@
...
+const procFilesystems = "/proc/filesystems"
+
+const CgroupV1FsType = "cgroup"
+const CgroupV2FsType = "cgroup2"
+
+const DefaultCgroupController = "cpuset"
+
+const TmpCgroup = "/tmp"
+const TmpCgroupV1Path = "tracee-cgroupv1"
+const TmpCgroupV2Path = "tracee-cgroupv2"
```
josedonizetti, line 29: Nit

```suggestion
const (
  procCgroups = "/proc/cgroups"
  procFilesystems = "/proc/filesystems"
  
  CgroupV1FsType = "cgroup"
  CgroupV2FsType = "cgroup2"
  
  DefaultCgroupController = "cpuset"
  
  TmpCgroup = "/tmp"
  TmpCgroupV1Path = "tracee-cgroupv1"
  TmpCgroupV2Path = "tracee-cgroupv2"
)
```

pkg/cgroup/cgroup.go
```diff
@@ -0,0 +1,312 @@
...
+}
+
+// NewCgroups tries to initialize both: cgroupv1 and cgroupv2, setting one as
+// the default depending on the running environment
+func NewCgroups() (*Cgroups, error) {
+	cs := &Cgroups{}
+
+	isCgroupV1, err := IsCgroupV1Default() // initialize cgroupv1
+	if err != nil {
+		return cs, err
```
josedonizetti, line 56: considering we are returning a pointer, best to return nil if there is an error, no?
if you agree, it would make sense to change all returns of errors on this method

```suggestion
		return nil, err
```

pkg/cgroup/cgroup.go
```diff
@@ -0,0 +1,312 @@
...
+}
+
+func (cs *Cgroups) GetV2() *Cgroup {
+	return cs.cgroupv2
+}
+
+// Destroy unmounts cgroupfs that were temporarily mounted
+func (cs *Cgroups) Destroy() error {
+	if cs.cgroupv1 != nil {
+		err := cs.cgroupv1.Destroy()
```
josedonizetti, line 102: nit
```suggestion
		return cs.cgroupv1.Destroy()
```

pkg/cgroup/cgroup.go
```diff
@@ -0,0 +1,312 @@
...
+		if err != nil {
+			return err
+		}
+	}
+	if cs.cgroupv2 != nil {
+		err := cs.cgroupv2.Destroy()
```
josedonizetti, line 108: nit

```suggestion
		return cs.cgroupv2.Destroy()
```

pkg/cgroup/cgroup.go
```diff
@@ -0,0 +1,312 @@
...
+	case CgroupV1:
+		tmpPath = TmpCgroupV1Path
+		fsType = CgroupV1FsType
+		controller = DefaultCgroupController
+	case CgroupV2:
+		tmpPath = TmpCgroupV2Path
+		fsType = CgroupV2FsType
+		controller = "" // no default controller for cgroupv2
+	default:
+		return c, failWrongCgroupVersion()
```
josedonizetti, line 136: considering we are returning a pointer, best to return nil if there is an error, no?
if you agree, it would make sense to change all returns of errors on this method

```suggestion
		return nil, failWrongCgroupVersion()
```

pkg/cgroup/cgroup.go
```diff
@@ -0,0 +1,312 @@
...
+	mountPath string
+	hierarchy int
+}
+
+func (c *Cgroup) Destroy() error {
+	err := syscall.Unmount(c.mountPath, 0)
+	if err != nil {
+		return err
+	}
+	err = os.RemoveAll(c.mountPath)
```
josedonizetti, line 187: nit
```suggestion
	return os.RemoveAll(c.mountPath)
```

pkg/cgroup/cgroup.go
```diff
@@ -0,0 +1,312 @@
...
+// Checks if cgroupv2 filesystem is supported
+func IsCgroupV2Supported() (bool, error) {
+	file, err := os.Open(procFilesystems)
+	if err != nil {
+		return false, failCouldNotOpenFile(procFilesystems, err)
+	}
+	defer file.Close()
+
+	scanner := bufio.NewScanner(file)
+	for i := 1; scanner.Scan(); i++ {
```
josedonizetti, line 304: Nit
```suggestion
	for scanner.Scan() {
```

pkg/mount/mount.go
```diff
@@ -0,0 +1,55 @@
...
+func SearchMountpoint(fstype string, search string) (string, error) {
+	file, err := os.Open(procMounts)
+	if err != nil {
+		return "", err
+	}
+	defer file.Close()
+
+	mp := ""
+	scanner := bufio.NewScanner(file)
+	for i := 1; scanner.Scan(); i++ {
```
josedonizetti, line 45: nit
```suggestion
	for scanner.Scan() {
```

pkg/mount/mount.go
```diff
@@ -0,0 +1,55 @@
...
+// IsMountpointOf searches if path is a mountpoint for fstype
+func IsMountpointOf(path string, fstype string) (bool, error) {
+	file, err := os.Open(procMounts)
+	if err != nil {
+		return false, err
+	}
+	defer file.Close()
+
+	scanner := bufio.NewScanner(file)
+	for i := 1; scanner.Scan(); i++ {
```
josedonizetti, line 20: nit
```suggestion
	for scanner.Scan() {
```

pkg/ebpf/probes/common.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"fmt"
+
+	bpf "github.com/aquasecurity/libbpfgo"
+)
+
+// enableDisableAutoload enables or disables an eBPF program autoload setting
+func enableDisableAutoload(module *bpf.Module, programName string, autoload bool) error {
+	var err error
```
josedonizetti, line 11: nit, we can remove this line

josedonizetti (commented):
@rafaeldtinoco Added a question about an if that seems would never work, and a couple of nits, feel free to ignore them if you don't agree.

pkg/cgroup/cgroup.go
rafaeldtinoco: Because in cgroupv2 they are all the same Id

josedonizetti: Ahhh! Maybe add a comment about it?

pkg/cgroup/cgroup.go
rafaeldtinoco: We tend to return an empty object in some other places... and I had the same doubt when I did it. Thinking a bit more, after your comment, since I'm setting t.cgroups directly, maybe it is better to have it as nil if not initialized correctly.

pkg/cgroup/cgroup.go
rafaeldtinoco: Then I would not destroy cgroupv2 if cgroupv1 does not exist (I might have cgroupv1 + cgroupv2 or cgroupv2 only).

pkg/cgroup/cgroup.go
rafaeldtinoco: I could indeed return here directly, preferred not to just "because" (so both calls were the same =D).

pkg/cgroup/cgroup.go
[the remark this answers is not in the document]
rafaeldtinoco: Yes, same case as the other one. I agree.

rafaeldtinoco:
@josedonizetti thanks for the review Jose, I have addressed most of your nits and suggestions and pushed a new commit. I'm using new commits because I'm rebasing the new-network PR on top of this one.

rafaeldtinoco:
Okay @yanivagman, I guess it is on you now. Let me know if you have any suggestions/concerns, please. Thanks.

pkg/cgroup/cgroup.go
```diff
@@ -0,0 +1,311 @@
...
+//
+// Cgroups
+//
+
+// Cgroups handles cgroups initialization (v1+v2 or v2 only)
+type Cgroups struct {
+	cgroupv1 *Cgroup // used for cgroup filesystem readings
+	cgroupv2 *Cgroup // used for cgroup prog attachment (and readings, if v2 is default)
+	cgroup   *Cgroup // default cgroup (might point to one or another)
+}
```
yanivagman, line 49: Why do we need to keep both cgroupv1 and cgroupv2? Won't having the deafult cgroup be enough?

pkg/ebpf/tracee.go
```diff
@@ -408,17 +412,31 @@ func (t *Tracee) Init() error {
...
-	// Initialize containers enrichment logic
+	// Initialize cgroups filesystems
 
-	capabilities.GetInstance().Requested(func() error { // TODO: workaround until PR: #2233 is in place
+	err = caps.Requested(func() error { // ring2
+		t.cgroups, err = cgroup.NewCgroups()
+		return err
+	}, cap.SYS_ADMIN)
+	if err != nil {
+		return err
```
yanivagman, line 422: This will fail tracee from running. But is that what we really want? What if the user don't need cgroup programs?

pkg/cgroup/cgroup.go
```diff
@@ -0,0 +1,311 @@
...
+
+	switch ver {
+	case CgroupV1:
+		tmpPath = TmpCgroupV1Path
+		fsType = CgroupV1FsType
+		controller = DefaultCgroupController
+	case CgroupV2:
+		tmpPath = TmpCgroupV2Path
+		fsType = CgroupV2FsType
+		controller = "" // no default controller for cgroupv2
```
yanivagman, line 136: nit: No need to initialize the variable to an empty string. This is the default value of a string variable.

pkg/containers/containers.go
```diff
@@ -229,12 +138,12 @@ func (c *Containers) populate() error {
...
 
-	return filepath.WalkDir(c.cgroupMP, fn)
+	return filepath.WalkDir(c.cgroup.GetMountPoint(), fn)
 }
 
 // CgroupUpdate checks if given path belongs to a known container runtime,
-// saving container information in Containers CgroupInfo map.
-// NOTE: ALL given cgroup dir paths are stored in CgroupInfo map.
+// saving container information in Containers CgroupInfo map.  NOTE: ALL given
+// cgroup dir paths are stored in CgroupInfo map.
```
yanivagman, line 146: This doesn't make the comment more readable

yanivagman (changes requested):
Ok, I finished reviewing this one.
I like the refactor you've done, but there are a few points that might cause tracee to work on less environments. Please see my comments below.

Also, I think it would have been easier if you would have made one refactor PR (with no logic changes at all), and another one with your new suggested changes. We can merge the refactor quickly, and then discuss the required changes for the cgroup stuff. Your call if you want to do this split or not.

pkg/cgroup/cgroup.go
NDStrahilevitz: I think this one takes care of the case where you have a hybrid system with both.

pkg/ebpf/tracee.go
NDStrahilevitz: We need the mounting logic anyway when we run tracee inside a container (see from a while back #1338 #1900)

yanivagman: But now it is always required, which makes the requirement for CAP_BPF and CAP_PERFMON on newer kernels irrelevant.
CAP_SYS_ADMIN is something we should avoid. We need to have the ability to let the user run tracee without requiring this capability - either by running the binary directly or by mounting the cgroup themselves.

NDStrahilevitz: We only need to run this in a capabilities ring once during initialization and that's it, other than that we can keep running only with CAP_BPF and CAP_PERFMON.
I do think we should move this caps call to the cgroups method though, that way we can scope it only for the mount call itself.

yanivagman: The thing is that there are environments where CAP_SYS_ADMIN won't be available for us, and we need to try to run tracee even without it (of course, if we can't continue - we can return an error)

rafaeldtinoco: I'll do like what we've discussed and mount only when needed. For mounting, I'll raise to ring2 with CAP_SYSADMIN only (and will have to unmounted at the cleanup).

rafaeldtinoco:
@yanivagman, I'm reworking this PR entirely, after refactoring the code in other already merged PRs, but, about:


> I think we should opt to use the previous method for deciding about group v1/v2 since it is the standard way to do this check and also runc uses it (and we need to be consistent with runc): https://github.com/opencontainers/runc/blob/main/libcontainer/cgroups/fs2/create.go#L14

> This is important for correctly resolving the container ID from the cgroup path. Also, it will allow not to return an error from this function, and if the file doesn't exist we can simply assume cgroup v1 (as we did with the current implementation)

About hierarchyID, the cgroups man page states:

```
    2. The unique ID of the cgroup hierarchy on which this
       controller is mounted.  If multiple cgroups v1
       controllers are bound to the same hierarchy, then each
       will show the same hierarchy ID in this field.  The
       value in this field will be 0 if:

       a) the controller is not mounted on a cgroups v1
          hierarchy;

       b) the controller is bound to the cgroups v2 single
          unified hierarchy; or

       c) the controller is disabled (see below).
```

So, the previous method (currently in use) always assumes that cgroupv1 is mounted in /sys/fs/cgroup tmpfs and that might not always be the case. We could have unmounted cgroupv1 but, still, being the default, for example.

Kubernetes documentation (https://kubernetes.io/docs/concepts/architecture/cgroups/) is also misleading as they tell user to identify if "/sys/fs/cgroup" is a tmpfs or a cgroup2 filesystem (but in our case the filesystems might not even be mounted).

With that said, I believe the correct way of detecting the default cgroup version being used is through hiearchyID of cpuset from /proc/cgroups.

rafaeldtinoco:
About last comment, reading: https://github.com/systemd/systemd/blob/69c0807432fa4fbfbf507a53872664cd26715559/src/basic/cgroup-util.c#L1986 we see that systemd relies in executing stat on "/sys/fs/cgroup" to check if it is a cgroup2 filesystem (same approach documented by kubernetes). The fact is that we might run tracee inside a container with no group mounted (let's say). Then, the /proc/cgroups is the best alternative to understand which group is the default and to mount one (or both) cgroup versions.

pkg/cgroup/cgroup.go
rafaeldtinoco: @yanivagman, I'm reworking this PR entirely, after refactoring the code in other already merged PRs, but, about:


> I think we should opt to use the previous method for deciding about group v1/v2 since it is the standard way to do this check and also runc uses it (and we need to be consistent with runc): https://github.com/opencontainers/runc/blob/main/libcontainer/cgroups/fs2/create.go#L14

> This is important for correctly resolving the container ID from the cgroup path. Also, it will allow not to return an error from this function, and if the file doesn't exist we can simply assume cgroup v1 (as we did with the current implementation)

About hierarchyID, the cgroups man page states:

```
    2. The unique ID of the cgroup hierarchy on which this
       controller is mounted.  If multiple cgroups v1
       controllers are bound to the same hierarchy, then each
       will show the same hierarchy ID in this field.  The
       value in this field will be 0 if:

       a) the controller is not mounted on a cgroups v1
          hierarchy;

       b) the controller is bound to the cgroups v2 single
          unified hierarchy; or

       c) the controller is disabled (see below).
```

So, the previous method (currently in use) always assumes that cgroupv1 is mounted in /sys/fs/cgroup tmpfs and that might not always be the case. We could have unmounted cgroupv1 but, still, being the default, for example.

Kubernetes documentation (https://kubernetes.io/docs/concepts/architecture/cgroups/) is also misleading as they tell user to identify if "/sys/fs/cgroup" is a tmpfs or a cgroup2 filesystem (but in our case the filesystems might not even be mounted).

With that said, I believe the correct way of detecting the default cgroup version being used is through hiearchyID of cpuset from /proc/cgroups.

rafaeldtinoco: About last comment, reading: https://github.com/systemd/systemd/blob/69c0807432fa4fbfbf507a53872664cd26715559/src/basic/cgroup-util.c#L1986 we see that systemd relies in executing stat on "/sys/fs/cgroup" to check if it is a cgroup2 filesystem (same approach documented by kubernetes). The fact is that we might run tracee inside a container with no group mounted (let's say). Then, the /proc/cgroups is the best alternative to understand which group is the default and to mount one (or both) cgroup versions.

pkg/cgroup/cgroup.go
```diff
@@ -0,0 +1,311 @@
...
+const (
+	procCgroups     = "/proc/cgroups"
+	procFilesystems = "/proc/filesystems"
+
+	CgroupV1FsType = "cgroup"
+	CgroupV2FsType = "cgroup2"
+
+	DefaultCgroupController = "cpuset"
+
+	TmpCgroup       = "/tmp"
```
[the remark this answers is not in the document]
rafaeldtinoco: Actually this brings to a good topic... we should use default OS temp directory for temporary files (like these and PID files) AND, perhaps, rename "tmpdir" to something else (it is not a temporary directory, it is an artifacts output directory). No ?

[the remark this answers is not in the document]
rafaeldtinoco: Also, if we want to use a "global tmpdir", then we should have global variable to it (instead of having it as part of Tracee object). Or else we will have to pass it as an argument for all packages using it.

pkg/cgroup/cgroup.go
```diff
@@ -0,0 +1,311 @@
...
+	if mp {
+		return nil, cgroupAlreadyMounted(ver, mountPath)
+	}
+
+	// equivalent to:
+	//
+	// cgroupv1: mount -t cgroup -o cpuset cgroup mountPath
+	// cgroupv2: mount -t cgroup2 cgroup2 mountPath
+
+	err = syscall.Mount(fsType, mountPath, fsType, 0, controller) // mount
```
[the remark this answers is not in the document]
rafaeldtinoco: Being solved by new code.

pkg/cgroup/cgroup.go
rafaeldtinoco: I believe this is solved (Nadav explained the reason).

pkg/ebpf/tracee.go
rafaeldtinoco: This is being solved by new code.

pkg/cgroup/cgroup.go
```diff
@@ -14,73 +12,332 @@ import (
...
+type Cgroup interface {
+	init() error
+	destroy() error
+	GetMountPoint() string
+	getDefaultHierarchyID() int
+}
+
+func NewCgroup(ver CgroupVersion) (Cgroup, error) {
+	switch ver {
+	case CgroupVersion1:
```
NDStrahilevitz, line 158: this can be shorter:
```
var c Cgroup
switch ver {
case CgroupVersion1:
     c = &Cgroup
...
}
err := c.init()
return c, err
```

pkg/mount/mount.go
```diff
@@ -2,54 +2,214 @@ package mount
...
+	m.managed = false // proforma
+
+	return true, nil
+}
+
+//
+// General
+//
+
+func IsDirEmpty(pathname string) (bool, error) {
```
NDStrahilevitz, line 158: I think this can move to `pkg/utils/files.go`

pkg/containers/containers.go
```diff
@@ -17,16 +17,12 @@ import (
...
 
-var cgroupV1HierarchyID int
-
 // Containers contains information about running containers in the host.
 type Containers struct {
-	cgroupV1   bool
-	cgroupMP   string
-	cgroups    map[uint32]CgroupInfo
+	cgroups    *cgroup.Cgroups
+	cgroupsMap map[uint32]CgroupInfo
```
NDStrahilevitz, line 25: suggestion: maybe `cgroupsInfo` or `cgroupsCache`? WDYT?

pkg/mount/mount.go
```diff
@@ -2,54 +2,214 @@ package mount
...
+}
+
+func (m *MountOnce) GetMountpoint() string {
+	return m.target
+}
+
+// private
+
+func (m *MountOnce) isMountedByOS() (bool, error) {
+	mp, err := SearchMountpoint(m.fsType, m.data)
```
yanivagman, line 139: I would only look at standard mountpoints (e.g. /sys/fs/cgroup or /tmp/tracee/*) and not on any mountpoint.
It can be that some other userspace program has mounted cgroup fs (e.g. Cilium), and we will find it first. But then, if this other program stops running, it might also unmount the fs.

pkg/cgroup/cgroup.go
```diff
@@ -14,73 +12,332 @@ import (
...
+	}
+
+	return ""
+}
+
+const (
+	CgroupVersionNone CgroupVersion = 1 << iota
+	CgroupVersion1
+	CgroupVersion2
+	CgroupVersionAll CgroupVersion = CgroupVersion1 & CgroupVersion2
```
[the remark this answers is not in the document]
rafaeldtinoco: Its a leftover from a previous idea, making it iota.

pkg/ebpf/tracee.go
```diff
@@ -408,17 +410,30 @@ func (t *Tracee) Init() error {
...
 
-		t.containers, err = containers.New(t.config.Sockets, "containers_map", t.config.Debug)
-		if err != nil {
-			return fmt.Errorf("error initializing containers: %w", err)
-		}
-		return nil
+	// Initialize containers enrichment logic
 
-	}, cap.SYS_ADMIN)
+	capabilities.GetInstance().Requested(
```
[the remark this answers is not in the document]
rafaeldtinoco: No, actually you're right. Nice catch! Thanks

pkg/containers/containers.go
rafaeldtinoco: Nope, Info is already used and it is not a cache =).

pkg/cgroup/errors.go
```diff
@@ -0,0 +1,21 @@
+package cgroup
+
+import "fmt"
+
+type VersionNotSupported struct{}
```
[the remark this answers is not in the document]
rafaeldtinoco: Because of:

image

I'm actually checking the type of the error to take decisions (unsupported version is an accepted error).

pkg/cgroup/cgroup.go
```diff
@@ -14,73 +12,332 @@ import (
...
+type Cgroups struct {
+	cgroupv1 Cgroup
+	cgroupv2 Cgroup
+	cgroup   *Cgroup // pointer to default cgroup version
+	hid      int     // default cgroup controller hiearchy ID
+}
+
+func NewCgroups() (*Cgroups, error) {
+	var err error
+	var cgrp *Cgroup
```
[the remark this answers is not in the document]
rafaeldtinoco: This is a pointer of a pointer (its the default cgroup, pointing to either cgroupv1 or cgroupv2).

pkg/cgroup/cgroup.go
```diff
@@ -14,73 +12,332 @@ import (
...
-		sline := strings.Fields(scanner.Text())
-		if len(sline) < 2 || sline[0] != CgroupV1Controller {
+	for scanner.Scan() {
+		line := strings.Fields(scanner.Text())
+		if line[0] != CgroupDefaultController {
 			continue
 		}
-		intID, err := strconv.Atoi(sline[1])
-		if err != nil || intID < 0 {
-			return -1, fmt.Errorf("error parsing %s: %w", cgroupsFile, err)
```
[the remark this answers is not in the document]
rafaeldtinoco: Nothing, returning it.

pkg/cgroup/cgroup.go
```diff
@@ -14,73 +12,332 @@ import (
...
+	//    b) the controller is bound to the cgroups v2 single
+	//       unified hierarchy; or
+	//    c) the controller is disabled (see below).
+	// ...
+
+	var value int
+
+	file, err := os.Open(procCgroups)
 	if err != nil {
-		return -1, fmt.Errorf("could not open cgroups file %s: %w", cgroupsFile, err)
```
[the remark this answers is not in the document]
rafaeldtinoco: nothing, returning it.

pkg/mount/mount.go
rafaeldtinoco: Fair enough.

pkg/cgroup/cgroup.go
```diff
@@ -14,73 +12,332 @@ import (
...
+
+func NewCgroups() (*Cgroups, error) {
+	var err error
+	var cgrp *Cgroup
+	var cgroupv1, cgroupv2 Cgroup
+
+	// discover the default cgroup being used
+	defaultVersion, err := GetCgroupDefaultVersion()
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
rafaeldtinoco: I believe for not to fail in the inexistence of a cgroupfs we should review containers.go as it might depend on its existence (the change would have to be coordinated among the two). I have opened https://github.com/aquasecurity/tracee/issues/2371 to be checked after this PR is merged (it is blocking the new network code for sometime now).

pkg/cgroup/cgroup.go
```diff
@@ -14,73 +12,352 @@ import (
...
+//
+// General
+//
+
+func GetCgroupDefaultVersion() (CgroupVersion, error) {
+	// 1st Method: already mounted cgroupv1 filesystem
+
+	ok, err := IsCgroupV1MountedAndDefault()
+	if err != nil {
+		return -1, err
```
yanivagman, line 272: maybe we should continue with the second method instead of failing here?

pkg/cgroup/cgroup.go
```diff
@@ -14,73 +12,352 @@ import (
...
-// mount syscall, it requires CAP_SYS_ADMIN
-func MountCgroupV1Cpuset(mountPath string) error {
-	mp, _ := mount.IsMountpoint(mountPath, CgroupV1FsType)
+// IsCgroupV1MountedAndDefault tests if cgroupv1 is mounted and is the default
+// cgroup version being used by the running environment. It does so by checking
+// the existance of a "release_agent" file in the cpuset cgroup controller.
+//
+// https://www.kernel.org/doc/Documentation/cgroup-v1/cgroups.txt
+func IsCgroupV1MountedAndDefault() (bool, error) {
+	_, err := os.Stat(CgroupReleaseAgent)
```
yanivagman, line 327: the method we used was to check for the existance of cgroup.controller (this is also used by runc) - shouldn't we continue with that?

pkg/mount/mount.go
```diff
@@ -3,53 +3,202 @@ package mount
...
 	if err != nil {
 		return false, err
 	}
+	// Check for mounted cgroupfs in sysfs only (other mountpoints might be
+	// being managed by other softwares, and they could unmount the cgroupfs
+	// unexpectedly)
+	if mp == "" || !strings.Contains(mp, sysFsCgroupPrefix) {
```
yanivagman, line 147: This package should be a generic mount package, no?
If so, we should get this as a parameter and not hardcode the cgroup prefix

pkg/cgroup/cgroup.go
rafaeldtinoco: Actually I almost copied the function we had and it was:
```diff
-func RunsOnContainerV1() (bool, error) {
-       _, err := os.Stat(CgroupReleaseAgent)
-       if os.IsNotExist(err) {
-               return true, nil
-       }
-       if err != nil {
-               return false, err
-       }
-       return false, nil
-}
```

rafaeldtinoco: image

yanivagman: Like discussed offline, this was used to check if we run in a container or not once we already know that we are on a cgroup v1 environment (by checking for cgroup.controller existance)

yanivagman (approved):
Thanks for addressing all the comments @rafaeldtinoco 
LGTM now

Issue #2236, opened 2022-10-17
[FEAT]: add tracee binary with collect and rules subcommand


josedonizetti:
## Prerequisites

- [ ] This issue is an EPIC issue (add label: EPIC).
- [x] This issue is an EPIC TASK (add issue to EPIC description).

Select one OR another:

- [z] I'll create a PR to implement this feature (assign to yourself).
- [ ] Someone else should implement this (describe it well).

## Feature description

Add tracee binary with subcommand for collecting and rules.

josedonizetti:
Considering the chat we had, about the one binary after the caps change @rafaeldtinoco is doing, and rules detection becoming event, this issue no longer makes sense.

PR #2237, opened 2022-10-17
feat: add cmd/tracee

josedonizetti:
Add binary with subcommands for collecting (ebpf), and rules.

Fixes https://github.com/aquasecurity/tracee/issues/2236

This PR extract two new packages `cmd/tracee/collect` and `cmd/tracee/rules` from `cmd/tracee-ebpf` and `cmd/tracee-rules` to create a new binary that support the same behavior but with subcomands. This is one of the refactorings towards a single binary, which I'm breaking in several PRs to help with reviewing and multiple conflicts due to main changes. 

The idea is to keep supporting `tracee-ebpf` and `tracee-rules` while we are not done with the one binary.

```
NAME:
   Tracee - Tracee cloud native runtime security

USAGE:
   tracee [global options] command [command options] [arguments...]

COMMANDS:
   collect  Collect OS events and syscalls using eBPF
   rules    A rule engine for runtime security
   help, h  Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --help, -h  show help (default: false)

```


josedonizetti:
Considering our chat today. This PR no longer makes sense.

PR #2217, opened 2022-10-08
[FEAT] Multi scopes (set) of filters

geyslan:
## Initial Checklist

- [x] There is an issue describing the need for this PR.
- [x] Git log contains summary of the change.
- [x] Git log contains motivation and context of the change.

## Description (git log)

commit f98189d4671800f3c1c615a4eea8d10c51e67561 (HEAD -> greg-multi-scopes, origin/greg-multi-scopes)
Author: Geyslan Gregório <geyslan@gmail.com>
Date:   Sat Sep 24 12:09:19 2022 -0300

    multi-scopes user space implementation

commit c84214300a8df11f05bc8df0f54cc1716f0b094f
Author: Yaniv Agman <yanivagman@gmail.com>
Date:   Tue Jun 28 22:00:55 2022 +0300

    multi-scopes initial work

Fixes: #2176

## Type of change

- [x] New feature (non-breaking change adding functionality).
- [x] Breaking change (cause existing functionality not to work as expected).

Issue #2176, opened 2022-09-20
Multi scopes (set) of filters


geyslan:
Currently Tracee only supports one set of filters (in the CLI) given by the user to decide about the scope of events to emit as part of the output. 

In some cases, the user might want to be able to have different “streams” of events. Each stream can be described by a different scope, which is in fact a different set of filters.

# Technicalities

Covered issues:

- https://github.com/aquasecurity/tracee/issues/2301
- https://github.com/aquasecurity/tracee/issues/2337

rafaeldtinoco:
Closed by https://github.com/aquasecurity/tracee/pull/2626

PR #2238, opened 2022-10-17
feat: multi scopes (set) of filters

geyslan:
Due to https://github.com/aquasecurity/tracee/pull/2217#issuecomment-1275130338, closed #2217 and pushed this one.

8738fe4f addressing minor issues
efa5acf9 ebpf: fix network capture using multiscopes
7feb73be flags/filter: better help message for multiscopes
9923b011 filters: change var names for better understand
07c48e9a scopes: add all capture events to all 64 scopes
7f9db05b ebpf: lazy submit checks for net events
3422889b e2e: silence dns test prints
efff3762 fix: signature events + common events
80f1ade6 ebpf: check event submit by scope with map
f2a7ff25 pipeline: increment events filtered (derived)
f7674f85 fix: false container logic needs filter scopes out
af49f90d network: adjust multi scope for ingress packets
9b28fc0d feat: multi scopes (set) of filters

Fix: #2176

## Type of change

## Steps

- [x] https://github.com/aquasecurity/tracee/issues/2301
  - [x] PR https://github.com/aquasecurity/tracee/pull/2302 raised 
- [x] Increase scopes to 64
- [x] Improve shouldProcessEvent() to only iterate through a reduced number scopes (enabled and filtered by type)
- [x] Tackle `// TODO:gg`
  - [x] Implement the new cli syntax parsing (under discussion)
- [x] Make sure that all flags are being checked in `Tracee.shouldProcessEvent()`
  - [x] Fix Greater and Less which turned bugged after rebases
  - [x] Check set flag (net is not available)
  - [x] Create/change tests
- [x] Change printer output column to the matchedScopes flag only (without fanciness)
- [x] Create json output field for matchedScopes.
- [x] Create tests for new trace flag syntax
- [x] Create tests for multi-scopes
- [x] Change scope flag syntax to `<scope>:filtername=`
- [x] Changes after reviews
- [x] Comment magical parts of the code
- [x] Network filtering
- Update documentation .md (postponed to after merge)

### DIFF
--- a/pkg/cmd/flags/filter.go
+++ b/pkg/cmd/flags/filter.go
@@ -2,6 +2,7 @@ package flags
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	tracee "github.com/aquasecurity/tracee/pkg/ebpf"
@@ -53,7 +54,7 @@ Examples:
   --trace pid=510,1709                                         | only trace events from pid 510 or pid 1709
   --trace p=510 --trace p=1709                                 | only trace events from pid 510 or pid 1709 (same as above)
   --trace container=new                                        | only trace events from newly created containers
-  --trace container_id=ab356bc4dd554                           | only trace events from container id ab356bc4dd554
+  --trace container=ab356bc4dd554                              | only trace events from container id ab356bc4dd554
   --trace container                                            | only trace events from containers
   --trace c                                                    | only trace events from containers (same as above)
   --trace '!container'                                         | only trace events from the host
@@ -76,49 +77,134 @@ Examples:
   --trace binary=/usr/bin/ls                                   | only trace events from /usr/bin/ls binary
   --trace binary=host:/usr/bin/ls                              | only trace events from /usr/bin/ls binary in the host mount namespace
   --trace binary=4026532448:/usr/bin/ls                        | only trace events from /usr/bin/ls binary in 4026532448 mount namespace
-  --trace close.fd=5                                           | only trace 'close' events that have 'fd' equals 5
+  --trace close.args.fd=5                                      | only trace 'close' events that have 'fd' equals 5
   --trace openat.args.pathname=/tmp*                           | only trace 'openat' events that have 'pathname' prefixed by "/tmp"
   --trace openat.args.pathname!=/tmp/1,/bin/ls                 | don't trace 'openat' events that have 'pathname' equals /tmp/1 or /bin/ls
   --trace openat.context.processName=ls                        | only trace 'openat' events that have 'processName' equal to 'ls'
   --trace security_file_open.context.container                 | only trace 'security_file_open' events coming from a container
   --trace comm=bash --trace follow                             | trace all events that originated from bash or from one of the processes spawned by bash
-  --trace net=docker0                                          | trace the net events over docker0 interface
 
+Filters can also be configured within up to 64 scopes (workloads).
+Events that match all trace expressions within a single scope will be traced.
+To find out which scopes an event is related to, read the bitmask in one of these ways:
+
+- using '-o format:json', matchedScopes JSON field (in decimal)
+- using '-o format:table-verbose', SCOPES collumn (in hexadecimal)
+
+Examples:
+  -t 42:event=sched_process_exec -t 42:binary=/usr/bin/ls      | trace in scope 42 sched_process_exec event from /usr/bin/ls binary
+  -t 3:event=openat -t 3:comm=id -t 9:event=close -t 9:comm=ls | trace in scope 3 only openat event from id command
+                                                                 and
+                                                                 trace in scope 9 only close event from ls command
+  -t 6:event=openat -t 6:comm=id -t 7:event=close -t 7:comm=id | trace in scope 6 only openat event from id command
+                                                                 and
+                                                                 trace in scope 7 only close event from id command
+  -t 3:event=openat -t 3:comm=id -t 9:event=close              | trace in scope 3 only openat event from id command
+                                                                 and
+                                                                 trace in scope 9 only close event from all
 
 Note: some of the above operators have special meanings in different shells.
 To 'escape' those operators, please use single quotes, e.g.: 'uid>0'
 `
 }
 
-func PrepareFilter(filtersArr []string) (tracee.Filter, error) {
-	filter := tracee.Filter{
-		UIDFilter:         filters.NewBPFUInt32Filter(tracee.UIDFilterMap),
-		PIDFilter:         filters.NewBPFUInt32Filter(tracee.PIDFilterMap),
-		NewPidFilter:      filters.NewBoolFilter(),
-		MntNSFilter:       filters.NewBPFUIntFilter(tracee.MntNSFilterMap),
-		PidNSFilter:       filters.NewBPFUIntFilter(tracee.PidNSFilterMap),
-		UTSFilter:         filters.NewBPFStringFilter(tracee.UTSFilterMap),
-		CommFilter:        filters.NewBPFStringFilter(tracee.CommFilterMap),
-		BinaryFilter:      filters.NewBPFBinaryFilter(tracee.BinaryFilterMap),
-		ContFilter:        filters.NewBoolFilter(),
-		NewContFilter:     filters.NewBoolFilter(),
-		ContIDFilter:      filters.NewContainerFilter(tracee.CgroupIdFilterMap),
-		RetFilter:         filters.NewRetFilter(),
-		ArgFilter:         filters.NewArgFilter(),
-		ContextFilter:     filters.NewContextFilter(),
-		ProcessTreeFilter: filters.NewProcessTreeFilter(tracee.ProcessTreeFilterMap),
-		EventsToTrace:     []events.ID{},
+// filterFlag holds pre-parsed filter flag fields
+type filterFlag struct {
+	full              string
+	filterName        string
+	operatorAndValues string
+	scopeIdx          int
+}
+
+func parseFilterFlag(flag string) (*filterFlag, error) {
+	var (
+		scopeID           int // stores the parsed scope index, not its flag position
+		filterName        string
+		operatorAndValues string
+
+		scopeEndIdx      int // stores ':' flag index (end of the scope value)
+		filterNameIdx    int
+		filterNameEndIdx int
+		operatorIdx      int
+		err              error
+	)
+
+	scopeEndIdx = strings.Index(flag, ":")
+	operatorIdx = strings.IndexAny(flag, "=!<>")
+
+	if scopeEndIdx == -1 && operatorIdx == -1 {
+		return &filterFlag{
+			full:              flag,
+			filterName:        flag,
+			operatorAndValues: "",
+			scopeIdx:          scopeID,
+		}, nil
 	}
 
-	eventFilter := cliFilter{
-		Equal:    []string{},
-		NotEqual: []string{},
-	}
-	setFilter := cliFilter{
-		Equal:    []string{},
-		NotEqual: []string{},
+	if operatorIdx != -1 {
+		operatorAndValues = flag[operatorIdx:]
+		filterNameEndIdx = operatorIdx
+	} else {
+		operatorIdx = len(flag) - 1
+		filterNameEndIdx = len(flag)
 	}
 
+	// check operators
+	if len(operatorAndValues) == 1 ||
+		operatorAndValues == "!=" ||
+		operatorAndValues == "<=" ||
+		operatorAndValues == ">=" {
+
+		return nil, filters.InvalidExpression(flag)
+	}
+
+	if scopeEndIdx != -1 && scopeEndIdx < operatorIdx {
+		// parse its ID
+		scopeID, err = strconv.Atoi(flag[:scopeEndIdx])
+		if err != nil {
+			return nil, filters.InvalidScope(fmt.Sprintf("%s - %s", flag, err))
+		}
+
+		// now consider it as a scope index
+		scopeID--
+		if scopeID < 0 || scopeID > tracee.MaxFilterScopes-1 {
+			return nil, filters.InvalidScope(fmt.Sprintf("%s - scopes must be between 1 and %d", flag, tracee.MaxFilterScopes))
+		}
+
+		filterNameIdx = scopeEndIdx + 1
+	}
+
+	if len(operatorAndValues) >= 2 &&
+		operatorAndValues[0] == '!' &&
+		operatorAndValues[1] != '=' {
+
+		filterName = flag[filterNameIdx:]
+		if strings.HasSuffix(filterName, "follow") ||
+			strings.HasSuffix(filterName, "container") {
+
+			return &filterFlag{
+				full:              flag,
+				filterName:        filterName,
+				operatorAndValues: "",
+				scopeIdx:          scopeID,
+			}, nil
+		}
+
+		return nil, filters.InvalidExpression(flag)
+	}
+
+	// parse filter name
+	filterName = flag[filterNameIdx:filterNameEndIdx]
+
+	return &filterFlag{
+		full:              flag,
+		filterName:        filterName,
+		operatorAndValues: operatorAndValues,
+		scopeIdx:          scopeID,
+	}, nil
+}
+
+func PrepareFilterScopes(filtersArr []string) (*tracee.FilterScopes, error) {
 	eventsNameToID := events.Definitions.NamesToIDs()
 	// remove internal events since they shouldn't be accesible by users
 	for event, id := range eventsNameToID {
@@ -127,203 +213,245 @@ func PrepareFilter(filtersArr []string) (tracee.Filter, error) {
 		}
 	}
 
-	for _, filterFlag := range filtersArr {
-		filterName := filterFlag
-		operatorAndValues := ""
-		operatorIndex := strings.IndexAny(filterFlag, "=!<>")
-		if operatorIndex > 0 {
-			filterName = filterFlag[0:operatorIndex]
-			operatorAndValues = filterFlag[operatorIndex:]
+	// parse and store filters by scope
+	parsedMap := map[int][]*filterFlag{}
+	for _, filter := range filtersArr {
+		parsed, err := parseFilterFlag(filter)
+		if err != nil {
+			return nil, err
 		}
 
-		if len(operatorAndValues) == 1 || operatorAndValues == "!=" || operatorAndValues == "<=" || operatorAndValues == ">=" {
-			return tracee.Filter{}, filters.InvalidExpression(filterFlag)
-		}
-
-		if strings.Contains(filterFlag, ".retval") {
-			err := filter.RetFilter.Parse(filterName, operatorAndValues, eventsNameToID)
-			if err != nil {
-				return tracee.Filter{}, err
-			}
-			continue
-		}
-
-		if strings.Contains(filterFlag, ".context") {
-			err := filter.ContextFilter.Parse(filterName, operatorAndValues)
-			if err != nil {
-				return tracee.Filter{}, err
-			}
-			continue
-		}
-
-		if strings.Contains(filterFlag, ".args") {
-			err := filter.ArgFilter.Parse(filterName, operatorAndValues, eventsNameToID)
-			if err != nil {
-				return tracee.Filter{}, err
-			}
-			continue
-		}
-
-		// The filters which are more common (container, event, pid, set, uid) can be given using a prefix of them.
-		// Other filters should be given using their full name.
-		// To avoid collisions between filters that share the same prefix, put the filters which should have an exact match first!
-		if filterName == "comm" {
-			err := filter.CommFilter.Parse(operatorAndValues)
-			if err != nil {
-				return tracee.Filter{}, err
-			}
-			continue
-		}
-
-		if filterName == "binary" || filterName == "bin" {
-			err := filter.BinaryFilter.Parse(operatorAndValues)
-			if err != nil {
-				return tracee.Filter{}, err
-			}
-			continue
-		}
-
-		if strings.HasPrefix("container", filterName) {
-			if operatorAndValues == "=new" {
-				err := filter.NewContFilter.Parse("new")
-				if err != nil {
-					return tracee.Filter{}, err
-				}
-				continue
-			}
-			if operatorAndValues == "!=new" {
-				err := filter.ContFilter.Parse(filterName)
-				if err != nil {
-					return tracee.Filter{}, err
-				}
-				err = filter.NewContFilter.Parse("!new")
-				if err != nil {
-					return tracee.Filter{}, err
-				}
-				continue
-			}
-			if strings.Contains(operatorAndValues, "=") {
-				err := filter.ContIDFilter.Parse(operatorAndValues)
-				if err != nil {
-					return tracee.Filter{}, err
-				}
-				continue
-			}
-			err := filter.ContFilter.Parse(filterName)
-			if err != nil {
-				return tracee.Filter{}, err
-			}
-			continue
-		}
-
-		if strings.HasPrefix("!container", filterName) {
-			err := filter.ContFilter.Parse(filterName)
-			if err != nil {
-				return tracee.Filter{}, err
-			}
-			continue
-		}
-
-		if strings.HasPrefix("event", filterName) {
-			err := eventFilter.Parse(operatorAndValues)
-			if err != nil {
-				return tracee.Filter{}, err
-			}
-			continue
-		}
-
-		if filterName == "mntns" {
-			if strings.ContainsAny(operatorAndValues, "<>") {
-				return tracee.Filter{}, filters.InvalidExpression(operatorAndValues)
-			}
-			err := filter.MntNSFilter.Parse(operatorAndValues)
-			if err != nil {
-				return tracee.Filter{}, err
-			}
-			continue
-		}
-
-		if filterName == "pidns" {
-			if strings.ContainsAny(operatorAndValues, "<>") {
-				return tracee.Filter{}, filters.InvalidExpression(operatorAndValues)
-			}
-			err := filter.PidNSFilter.Parse(operatorAndValues)
-			if err != nil {
-				return tracee.Filter{}, err
-			}
-			continue
-		}
-
-		if filterName == "tree" {
-			err := filter.ProcessTreeFilter.Parse(operatorAndValues)
-			if err != nil {
-				return tracee.Filter{}, err
-			}
-			continue
-		}
-
-		if strings.HasPrefix("pid", filterName) {
-			if operatorAndValues == "=new" {
-				filter.NewPidFilter.Parse("new")
-				continue
-			}
-			if operatorAndValues == "!=new" {
-				filter.NewPidFilter.Parse("!new")
-				continue
-			}
-			err := filter.PIDFilter.Parse(operatorAndValues)
-			if err != nil {
-				return tracee.Filter{}, err
-			}
-			continue
-		}
-
-		if strings.HasPrefix("set", filterName) {
-			err := setFilter.Parse(operatorAndValues)
-			if err != nil {
-				return tracee.Filter{}, err
-			}
-			continue
-		}
-
-		if filterName == "uts" {
-			err := filter.UTSFilter.Parse(operatorAndValues)
-			if err != nil {
-				return tracee.Filter{}, err
-			}
-			continue
-		}
-
-		if strings.HasPrefix("uid", filterName) {
-			err := filter.UIDFilter.Parse(operatorAndValues)
-			if err != nil {
-				return tracee.Filter{}, err
-			}
-			continue
-		}
-
-		if strings.HasPrefix("follow", filterFlag) {
-			filter.Follow = true
-			continue
-		}
-		return tracee.Filter{}, InvalidFilterOptionError(filterFlag)
+		scopeIdx := parsed.scopeIdx
+		parsedMap[scopeIdx] = append(
+			parsedMap[scopeIdx],
+			parsed,
+		)
 	}
 
-	var err error
-	filter.EventsToTrace, err = prepareEventsToTrace(eventFilter, setFilter, eventsNameToID)
-	if err != nil {
-		return tracee.Filter{}, err
+	filterScopes := tracee.NewFilterScopes()
+	for scopeIdx, fsFlags := range parsedMap {
+		filterScope := tracee.NewFilterScope()
+		eventFilter := cliFilter{
+			Equal:    []string{},
+			NotEqual: []string{},
+		}
+		setFilter := cliFilter{
+			Equal:    []string{},
+			NotEqual: []string{},
+		}
+
+		for _, filterFlag := range fsFlags {
+			if strings.Contains(filterFlag.full, ".retval") {
+				err := filterScope.RetFilter.Parse(filterFlag.filterName, filterFlag.operatorAndValues, eventsNameToID)
+				if err != nil {
+					return nil, err
+				}
+				continue
+			}
+
+			if strings.Contains(filterFlag.full, ".context") {
+				err := filterScope.ContextFilter.Parse(filterFlag.filterName, filterFlag.operatorAndValues)
+				if err != nil {
+					return nil, err
+				}
+				continue
+			}
+
+			if strings.Contains(filterFlag.full, ".args") {
+				err := filterScope.ArgFilter.Parse(filterFlag.filterName, filterFlag.operatorAndValues, eventsNameToID)
+				if err != nil {
+					return nil, err
+				}
+				continue
+			}
+
+			// The filters which are more common (container, event, pid, set, uid) can be given using a prefix of them.
+			// Other filters should be given using their full name.
+			// To avoid collisions between filters that share the same prefix, put the filters which should have an exact match first!
+			if filterFlag.filterName == "comm" {
+				err := filterScope.CommFilter.Parse(filterFlag.operatorAndValues)
+				if err != nil {
+					return nil, err
+				}
+				continue
+			}
+
+			if filterFlag.filterName == "binary" || filterFlag.filterName == "bin" {
+				err := filterScope.BinaryFilter.Parse(filterFlag.operatorAndValues)
+				if err != nil {
+					return nil, err
+				}
+				continue
+			}
+
+			if strings.HasPrefix("container", filterFlag.filterName) {
+				if filterFlag.operatorAndValues == "=new" {
+					err := filterScope.NewContFilter.Parse("new")
+					if err != nil {
+						return nil, err
+					}
+					continue
+				}
+				if filterFlag.operatorAndValues == "!=new" {
+					err := filterScope.ContFilter.Parse(filterFlag.filterName)
+					if err != nil {
+						return nil, err
+					}
+					err = filterScope.NewContFilter.Parse("!new")
+					if err != nil {
+						return nil, err
+					}
+					continue
+				}
+				if strings.Contains(filterFlag.operatorAndValues, "=") {
+					err := filterScope.ContIDFilter.Parse(filterFlag.operatorAndValues)
+					if err != nil {
+						return nil, err
+					}
+					continue
+				}
+				err := filterScope.ContFilter.Parse(filterFlag.filterName)
+				if err != nil {
+					return nil, err
+				}
+				continue
+			}
+
+			if strings.HasPrefix("!container", filterFlag.filterName) {
+				err := filterScope.ContFilter.Parse(filterFlag.filterName)
+				if err != nil {
+					return nil, err
+				}
+				continue
+			}
+
+			if strings.HasPrefix("event", filterFlag.filterName) {
+				err := eventFilter.Parse(filterFlag.operatorAndValues)
+				if err != nil {
+					return nil, err
+				}
+				continue
+			}
+
+			if filterFlag.filterName == "mntns" {
+				if strings.ContainsAny(filterFlag.operatorAndValues, "<>") {
+					return nil, filters.InvalidExpression(filterFlag.operatorAndValues)
+				}
+				err := filterScope.MntNSFilter.Parse(filterFlag.operatorAndValues)
+				if err != nil {
+					return nil, err
+				}
+				continue
+			}
+
+			if filterFlag.filterName == "pidns" {
+				if strings.ContainsAny(filterFlag.operatorAndValues, "<>") {
+					return nil, filters.InvalidExpression(filterFlag.operatorAndValues)
+				}
+				err := filterScope.PidNSFilter.Parse(filterFlag.operatorAndValues)
+				if err != nil {
+					return nil, err
+				}
+				continue
+			}
+
+			if filterFlag.filterName == "tree" {
+				err := filterScope.ProcessTreeFilter.Parse(filterFlag.operatorAndValues)
+				if err != nil {
+					return nil, err
+				}
+				continue
+			}
+
+			if strings.HasPrefix("pid", filterFlag.filterName) {
+				if filterFlag.operatorAndValues == "=new" {
+					filterScope.NewPidFilter.Parse("new")
+					continue
+				}
+				if filterFlag.operatorAndValues == "!=new" {
+					filterScope.NewPidFilter.Parse("!new")
+					continue
+				}
+				err := filterScope.PIDFilter.Parse(filterFlag.operatorAndValues)
+				if err != nil {
+					return nil, err
+				}
+				continue
+			}
+
+			if strings.HasPrefix("set", filterFlag.filterName) {
+				err := setFilter.Parse(filterFlag.operatorAndValues)
+				if err != nil {
+					return nil, err
+				}
+				continue
+			}
+
+			if filterFlag.filterName == "uts" {
+				err := filterScope.UTSFilter.Parse(filterFlag.operatorAndValues)
+				if err != nil {
+					return nil, err
+				}
+				continue
+			}
+
+			if strings.HasPrefix("uid", filterFlag.filterName) {
+				err := filterScope.UIDFilter.Parse(filterFlag.operatorAndValues)
+				if err != nil {
+					return nil, err
+				}
+				continue
+			}
+
+			if strings.HasPrefix("follow", filterFlag.filterName) {
+				filterScope.Follow = true
+				continue
+			}
+
+			return nil, InvalidFilterOptionError(filterFlag.full)
+		}
+
+		var err error
+		filterScope.EventsToTrace, err = prepareEventsToTrace(eventFilter, setFilter, eventsNameToID)
+		if err != nil {
+			return nil, err
+		}
+
+		if err := filterScopes.Set(scopeIdx, filterScope); err != nil {
+			return nil, err
+		}
 	}
 
-	return filter, nil
+	if len(filterScopes.Map()) == 0 {
+		// If nothing was set, let us consider it as a single default scope
+		eventFilter := cliFilter{
+			Equal:    []string{},
+			NotEqual: []string{},
+		}
+		setFilter := cliFilter{
+			Equal:    []string{},
+			NotEqual: []string{},
+		}
+
+		var err error
+		newScope := tracee.NewFilterScope()
+		newScope.EventsToTrace, err = prepareEventsToTrace(eventFilter, setFilter, eventsNameToID)
+		if err != nil {
+			return nil, err
+		}
+
+		filterScopes.Add(newScope)
+	}
+
+	return filterScopes, nil
 }
 
-func prepareEventsToTrace(eventFilter cliFilter, setFilter cliFilter, eventsNameToID map[string]events.ID) ([]events.ID, error) {
+func prepareEventsToTrace(eventFilter cliFilter, setFilter cliFilter, eventsNameToID map[string]events.ID) (map[events.ID]string, error) {
 	eventsToTrace := eventFilter.Equal
 	excludeEvents := eventFilter.NotEqual
 	setsToTrace := setFilter.Equal
 
-	var res []events.ID
+	var res map[events.ID]string
 	setsToEvents := make(map[string][]events.ID)
 	isExcluded := make(map[events.ID]bool)
 	for id, event := range events.Definitions.Events() {
@@ -357,29 +485,27 @@ func prepareEventsToTrace(eventFilter cliFilter, setFilter cliFilter, eventsName
 		setsToTrace = append(setsToTrace, "default")
 	}
 
-	res = make([]events.ID, 0, events.Definitions.Length())
+	res = make(map[events.ID]string, events.Definitions.Length())
 	for _, name := range eventsToTrace {
 		// Handle event prefixes with wildcards
 		if strings.HasSuffix(name, "*") {
-			var ids []events.ID
 			found := false
 			prefix := name[:len(name)-1]
 			for event, id := range eventsNameToID {
 				if strings.HasPrefix(event, prefix) && !isExcluded[id] {
-					ids = append(ids, id)
+					res[id] = event
 					found = true
 				}
 			}
 			if !found {
 				return nil, InvalidEventError(name)
 			}
-			res = append(res, ids...)
 		} else {
 			id, ok := eventsNameToID[name]
 			if !ok {
 				return nil, InvalidEventError(name)
 			}
-			res = append(res, id)
+			res[id] = name
 		}
 	}
 	for _, set := range setsToTrace {
@@ -389,7 +515,7 @@ func prepareEventsToTrace(eventFilter cliFilter, setFilter cliFilter, eventsName
 		}
 		for _, id := range setEvents {
 			if !isExcluded[id] {
-				res = append(res, id)
+				res[id] = events.Definitions.Get(id).Name
 			}
 		}
 	}
--- a/pkg/cmd/flags/filter.go
+++ b/pkg/cmd/flags/filter.go
@@ -92,16 +92,20 @@ To find out which scopes an event is related to, read the bitmask in one of thes
 - using '-o format:table-verbose', SCOPES collumn (in hexadecimal)
 
 Examples:
+
   -t 42:event=sched_process_exec -t 42:binary=/usr/bin/ls      | trace in scope 42 sched_process_exec event from /usr/bin/ls binary
-  -t 3:event=openat -t 3:comm=id -t 9:event=close -t 9:comm=ls | trace in scope 3 only openat event from id command
-                                                                 and
-                                                                 trace in scope 9 only close event from ls command
-  -t 6:event=openat -t 6:comm=id -t 7:event=close -t 7:comm=id | trace in scope 6 only openat event from id command
-                                                                 and
-                                                                 trace in scope 7 only close event from id command
-  -t 3:event=openat -t 3:comm=id -t 9:event=close              | trace in scope 3 only openat event from id command
-                                                                 and
-                                                                 trace in scope 9 only close event from all
+
+  -t 3:event=openat -t 3:comm=id -t 9:event=close -t 9:comm=ls - trace in scope 3 only openat event from id command
+                                                               | and
+                                                               - trace in scope 9 only close event from ls command
+
+  -t 6:event=openat -t 6:comm=id -t 7:event=close -t 7:comm=id - trace in scope 6 only openat event from id command
+                                                               | and
+                                                               _ trace in scope 7 only close event from id command
+
+  -t 3:event=openat -t 3:comm=id -t 9:event=close              - trace in scope 3 only openat event from id command
+                                                               | and
+                                                               - trace in scope 9 only close event from all
 
 Note: some of the above operators have special meanings in different shells.
 To 'escape' those operators, please use single quotes, e.g.: 'uid>0'
--- a/pkg/cmd/flags/filter.go
+++ b/pkg/cmd/flags/filter.go
@@ -8,6 +8,7 @@ import (
 	tracee "github.com/aquasecurity/tracee/pkg/ebpf"
 	"github.com/aquasecurity/tracee/pkg/events"
 	"github.com/aquasecurity/tracee/pkg/filters"
+	"github.com/aquasecurity/tracee/pkg/logger"
 )
 
 func filterHelp() string {
@@ -421,8 +422,9 @@ func PrepareFilterScopes(filtersArr []string) (*tracee.FilterScopes, error) {
 			return nil, err
 		}
 
-		if err := filterScopes.Set(scopeIdx, filterScope); err != nil {
-			return nil, err
+		err = filterScopes.Set(scopeIdx, filterScope)
+		if err != nil {
+			logger.Warn("Setting scope", "error", err)
 		}
 	}
 
@@ -444,7 +446,10 @@ func PrepareFilterScopes(filtersArr []string) (*tracee.FilterScopes, error) {
 			return nil, err
 		}
 
-		filterScopes.Add(newScope)
+		err = filterScopes.Add(newScope)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	return filterScopes, nil
--- a/pkg/cmd/printer/printer.go
+++ b/pkg/cmd/printer/printer.go
@@ -103,20 +103,102 @@ func (p tableEventPrinter) Preamble() {
 	if p.verbose {
 		switch p.containerMode {
 		case ContainerModeDisabled:
-			fmt.Fprintf(p.out, "%-16s %-16s %-13s %-12s %-12s %-6s %-16s %-7s %-7s %-7s %-16s %-20s %s", "TIME", "UTS_NAME", "CONTAINER_ID", "MNT_NS", "PID_NS", "UID", "COMM", "PID", "TID", "PPID", "RET", "EVENT", "ARGS")
+			fmt.Fprintf(p.out,
+				"%-16s %-17s %-17s %-13s %-12s %-12s %-6s %-16s %-7s %-7s %-7s %-16s %-20s %s",
+				"TIME",
+				"SCOPES",
+				"UTS_NAME",
+				"CONTAINER_ID",
+				"MNT_NS",
+				"PID_NS",
+				"UID",
+				"COMM",
+				"PID",
+				"TID",
+				"PPID",
+				"RET",
+				"EVENT",
+				"ARGS",
+			)
 		case ContainerModeEnabled:
-			fmt.Fprintf(p.out, "%-16s %-16s %-13s %-12s %-12s %-6s %-16s %-15s %-15s %-15s %-16s %-20s %s", "TIME", "UTS_NAME", "CONTAINER_ID", "MNT_NS", "PID_NS", "UID", "COMM", "PID/host", "TID/host", "PPID/host", "RET", "EVENT", "ARGS")
+			fmt.Fprintf(p.out,
+				"%-16s %-17s %-17s %-13s %-12s %-12s %-6s %-16s %-15s %-15s %-15s %-16s %-20s %s",
+				"TIME",
+				"SCOPES",
+				"UTS_NAME",
+				"CONTAINER_ID",
+				"MNT_NS",
+				"PID_NS",
+				"UID",
+				"COMM",
+				"PID/host",
+				"TID/host",
+				"PPID/host",
+				"RET",
+				"EVENT",
+				"ARGS",
+			)
 		case ContainerModeEnriched:
-			fmt.Fprintf(p.out, "%-16s %-16s %-13s %-16s %-12s %-12s %-6s %-16s %-15s %-15s %-15s %-16s %-20s %s", "TIME", "UTS_NAME", "CONTAINER_ID", "IMAGE", "MNT_NS", "PID_NS", "UID", "COMM", "PID/host", "TID/host", "PPID/host", "RET", "EVENT", "ARGS")
+			fmt.Fprintf(p.out,
+				"%-16s %-17s %-17s %-13s %-16s %-12s %-12s %-6s %-16s %-15s %-15s %-15s %-16s %-20s %s",
+				"TIME",
+				"SCOPES",
+				"UTS_NAME",
+				"CONTAINER_ID",
+				"IMAGE",
+				"MNT_NS",
+				"PID_NS",
+				"UID",
+				"COMM",
+				"PID/host",
+				"TID/host",
+				"PPID/host",
+				"RET",
+				"EVENT",
+				"ARGS",
+			)
 		}
 	} else {
 		switch p.containerMode {
 		case ContainerModeDisabled:
-			fmt.Fprintf(p.out, "%-16s %-6s %-16s %-7s %-7s %-16s %-20s %s", "TIME", "UID", "COMM", "PID", "TID", "RET", "EVENT", "ARGS")
+			fmt.Fprintf(p.out,
+				"%-16s %-6s %-16s %-7s %-7s %-16s %-20s %s",
+				"TIME",
+				"UID",
+				"COMM",
+				"PID",
+				"TID",
+				"RET",
+				"EVENT",
+				"ARGS",
+			)
 		case ContainerModeEnabled:
-			fmt.Fprintf(p.out, "%-16s %-13s %-6s %-16s %-15s %-15s %-16s %-20s %s", "TIME", "CONTAINER_ID", "UID", "COMM", "PID/host", "TID/host", "RET", "EVENT", "ARGS")
+			fmt.Fprintf(p.out,
+				"%-16s %-13s %-6s %-16s %-15s %-15s %-16s %-20s %s",
+				"TIME",
+				"CONTAINER_ID",
+				"UID",
+				"COMM",
+				"PID/host",
+				"TID/host",
+				"RET",
+				"EVENT",
+				"ARGS",
+			)
 		case ContainerModeEnriched:
-			fmt.Fprintf(p.out, "%-16s %-13s %-16s %-6s %-16s %-15s %-15s %-16s %-20s %s", "TIME", "CONTAINER_ID", "IMAGE", "UID", "COMM", "PID/host", "TID/host", "RET", "EVENT", "ARGS")
+			fmt.Fprintf(p.out,
+				"%-16s %-13s %-16s %-6s %-16s %-15s %-15s %-16s %-20s %s",
+				"TIME",
+				"CONTAINER_ID",
+				"IMAGE",
+				"UID",
+				"COMM",
+				"PID/host",
+				"TID/host",
+				"RET",
+				"EVENT",
+				"ARGS",
+			)
 		}
 	}
 	fmt.Fprintln(p.out)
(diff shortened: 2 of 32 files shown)


geyslan:
For testing:

```shell
sudo ./dist/tracee-ebpf \
-o format:table-verbose \
-t "comm=who comm=ls" \
-t "comm=who" \
-t "event=close comm=grep comm=ls" \
-t "comm=whois" \
-t "comm=who pid>1" \
-t "comm=who mntns=4026531841" \
-t "comm=ls pid>502000 pid<505000" \
-t "comm=uname pid>505000 pid<506000" \
-t "comm=who pid!=1" \
-t "event=close,security_file_open comm=grep"
```

geyslan:
There are comments like "TODO:gg" as next steps.

geyslan:
> For testing:
> 
> ```shell
> sudo ./dist/tracee-ebpf \
> -o format:table-verbose \
> -t "comm=who comm=ls" \
> -t "comm=who" \
> -t "event=close comm=grep comm=ls" \
> -t "comm=whois" \
> -t "comm=who pid>1" \
> -t "comm=who mntns=4026531841" \
> -t "comm=ls pid>502000 pid<505000" \
> -t "comm=uname pid>505000 pid<506000" \
> -t "comm=who pid!=1" \
> -t "event=close,security_file_open comm=grep"
> ```

BTW the changes in progress, @josedonizetti @rafaeldtinoco how about this syntax?

```shell
sudo ./dist/tracee-ebpf \
-t comm=whois comm=uname \
-t comm-2=who comm-2=ls \
-t event-3=close comm-3=grep comm-3=ls
```

The -t flag would remain unchanged. All filters would have to be parsed with additional dash and scope number (i.e. comm-2) but for a single scope or for the first one in the case of multiple definitions.

geyslan:
The current syntax is:

```shell
-o format:table-verbose \
-t comm=who \
-t comm-2=ls \
-t comm-5=whois \
-t comm-3=uname -t comm-3=who \
-t comm-64=who \
-t event-63=close -t comm-63=grep \
-t comm-39=whois
```

geyslan:
@rafaeldtinoco, @yanivagman shouldProcess() fixed. Ready for review.

Certainly new issues will arise in the process, but the next steps are:
- create new syntax tests
- update doc

pkg/ebpf/events_processor.go
```diff
@@ -31,10 +31,122 @@ func (t *Tracee) processLostEvents() {
...
+		return false
+	}
+	containsSet := func(setsFromCtxEvent []string, setsToTrace map[string]bool) bool {
+		for _, s := range setsFromCtxEvent {
+			if _, found := setsToTrace[s]; found {
+				return true
+			}
+		}
+		return false
+	}
```
geyslan, line 49: I'll move that logic to outer functions.

pkg/ebpf/filters.go
```diff
@@ -38,6 +41,327 @@ type Filter struct {
...
+	filterScopesMap          map[*FilterScope]int
+	filterScopesUserSpaceMap map[*FilterScope]int
+
+	uidFilterMin             uint64
+	uidFilterMax             uint64
+	pidFilterMin             uint64
+	pidFilterMax             uint64
+	uidFilterableInUserSpace bool
+	pidFilterableInUserSpace bool
+}
```
geyslan, line 91: Insert a comment regarding the future need for locking as scopes will change at runtime.

pkg/ebpf/tracee.go
```diff
@@ -609,11 +616,16 @@ func (t *Tracee) initDerivationTable() error {
 	pathResolver := containers.InitPathResolver(&t.pidsInMntns)
 	soLoader := sharedobjs.InitContainersSymbolsLoader(&pathResolver, 1024)
 
-	symbolsLoadedFilters := t.config.Filter.ArgFilter.GetEventFilters(events.SymbolsLoaded)
+	symbolsLoadedFilters := map[string]filters.Filter{}
+	for filterScope := range t.config.FilterScopes.UserSpaceMap() {
+		for k, v := range filterScope.ArgFilter.GetEventFilters(events.SymbolsLoaded) {
+			symbolsLoadedFilters[k] = v
```
geyslan, line 622: As we have already reached go 1.18, this can be replaced by `maps.Copy(symbolsLoadedFilters, filterScope.ArgFilter.GetEventFilters(events.SymbolsLoaded))`.

geyslan:
https://github.com/aquasecurity/tracee/pull/2238/commits/d06bf927487ef50808903abd2c11df8b61b43f2a fixed the issue of lost events on VMs starting from the jammy version.

Before that the event security_sb_mount (730) was being submitted and discarded (treated as a false container positive) in tests (i.e. TRC-1014).

@rafaeldtinoco @NDStrahilevitz @yanivagman

pkg/ebpf/filters.go
```diff
@@ -39,6 +42,348 @@ type Filter struct {
...
+
+// FillUserSpaceMap fills user space map
+// It must be called after CalculateGlobalMinMax()
+func (fs *FilterScopes) FillUserSpaceMap() {
+	userSpaceMap := make(map[*FilterScope]int)
+
+	for filterScope := range fs.filterScopesMap {
+		if filterScope.CheckEventsInUserSpace ||
+			filterScope.CheckSetsInUserSpace ||
+			filterScope.ArgFilter.Enabled() ||
```
geyslan, line 273: Add the Context filter.

pkg/ebpf/filters.go
geyslan: Add fields with pre-calculated values to reduce processing in hot paths, i.e.: `ContainerFilterEnabled()`, `IsFilterNetSet()`.

yanivagman:
> [d06bf92](https://github.com/aquasecurity/tracee/commit/d06bf927487ef50808903abd2c11df8b61b43f2a) fixed the event loss issue on VMs starting from the jammy version.
> 
> Before that the event `security_sb_mount` (730) was being submitted and discarded (treated as a false container positive) in tests (i.e. `TRC-1014`).
> 
> @rafaeldtinoco @NDStrahilevitz @yanivagman

These "false positives" are only relevant when the container filter is set, but I don't think that we set it in the test. So I wonder - how can it be that we entered the "if containerfilter.enabled" statement assuming it is not?

geyslan:
> > [d06bf92](https://github.com/aquasecurity/tracee/commit/d06bf927487ef50808903abd2c11df8b61b43f2a) fixed the event loss issue on VMs starting from the jammy version.
> > Before that the event `security_sb_mount` (730) was being submitted and discarded (treated as a false container positive) in tests (i.e. `TRC-1014`).
> > @rafaeldtinoco @NDStrahilevitz @yanivagman
> 
> These "false positives" are only relevant when the container filter is set, but I don't think that we set it in the test. So I wonder - how can it be that we entered the "if containerfilter.enabled" statement assuming it is not?

@yanivagman, ./tests/kerneltest.sh sets `container=new` flag: https://github.com/aquasecurity/tracee/blob/main/tests/kerneltest.sh#L103

Analyzing compared data from that test scenario (with just TRC-1014), config_entry `options` and `cgroup_v1_hid` are the unique fields with divergent values. @rafaeldtinoco 

```shell
~/code 
❯ diff tracee-jammy/pahole-jammy tracee-focal/pahole-focal

~/code 
❯ diff tracee-jammy/pahole-jammy tracee-focal/pahole-focal-noncore 

~/code 
❯ diff tracee-jammy/pahole-jammy-noncore tracee-focal/pahole-focal-noncore

~/code 
❯ diff tracee-jammy/config_map-jammy-noncore tracee-focal/config_map-focal-noncore 
4,6c4,6
<             "tracee_pid": 20002,
<             "options": 1,
<             "cgroup_v1_hid": 0,
---
>             "tracee_pid": 33827,
>             "options": 65,
>             "cgroup_v1_hid": 6,

```

The same difference in main branch:

```shell
❯ diff tracee-jammy/config_map-jammy-main-noncore tracee-focal/config_map-focal-main-noncore
4,5c4,5
<             "tracee_pid": 86100,
<             "options": 1,
---
>             "tracee_pid": 96846,
>             "options": 65,
7c7
<             "cgroup_v1_hid": 0,
---
>             "cgroup_v1_hid": 6,

```

**Focal**:

`getOptionsConfig()` is returning 65 (optDetectOrigSyscall | optCgroupV1)
`GetDefaultCgroupHierarchyID()` is returning 6

```shell
vagrant@ubuntu-focal:/vagrant$ head -2 /proc/cgroups 
#subsys_name	hierarchy	num_cgroups	enabled
cpuset	6	25	1

```

**Jammy**:

`getOptionsConfig()` is returning 1 (optDetectOrigSyscall)
`GetDefaultCgroupHierarchyID()` is returning 0

```shell
vagrant@ubuntu-jammy:/vagrant$ head -2 /proc/cgroups 
#subsys_name	hierarchy	num_cgroups	enabled
cpuset	0	96	1
```

pkg/cmd/flags/filter.go
```diff
@@ -88,38 +89,7 @@ To 'escape' those operators, please use single quotes, e.g.: 'uid>0'
 `
 }
 
-func PrepareFilter(filtersArr []string) (tracee.Filter, error) {
```
rafaeldtinoco, line 91: Don't forget to add examples in the `filterHelp()` string output AND to explain what it is.

pkg/cmd/flags/filter.go
```diff
@@ -128,203 +98,271 @@ func PrepareFilter(filtersArr []string) (tracee.Filter, error) {
...
-		// To avoid collisions between filters that share the same prefix, put the filters which should have an exact match first!
-		if filterName == "comm" {
-			err := filter.CommFilter.Parse(operatorAndValues)
-			if err != nil {
-				return tracee.Filter{}, err
-			}
-			continue
+		filterName = filterFlag[0:dashIndex]
+		operatorAndValues = filterFlag[operatorIdx:]
+		scopeIdx--
```
rafaeldtinoco, line 126: I see that no scope and scope -1 is "scope 0" actually. Why not to have "filter-0", "filter-1", "filter-2" and make "filter" or "filter-0" the base ones ?

rafaeldtinoco: Probably because the end user would be confused about adding a "filter=X" then a "filter-1=X", right? (Since it would be the "second" filter scope). Alright.

rafaeldtinoco (commented):
Still reviewing, leaving some comments for now.

NDStrahilevitz (commented):
Not a full review but IMO, the `filtername-<scope>=....` syntax is very ugly.
Isn't something like `<scope>:filtername` possible? 
Importantly I think scope should come before the filter name, to me at least it looks more intuitive (and of course omitting the scope should default to scope 1).
Not -1ing because if it's only me then it's only me.

pkg/filters/uint.go
```diff
@@ -203,14 +221,11 @@ func NewBPFUInt32Filter(mapName string) *BPFUIntFilter[uint32] {
...
-func (filter *BPFUIntFilter[T]) InitBPF(bpfModule *bpf.Module) error {
+func (filter *BPFUIntFilter[T]) InitBPF(bpfModule *bpf.Module, filterScopeID uint) error {
 	if !filter.Enabled() {
 		return nil
 	}
 
-	bpfFilterEqual := uint32(filterEqual) // const need local var for bpfMap.Update()
-	bpfFilterNotEqual := uint32(filterNotEqual)
-
 	// equalityFilter filters events for given maps:
```
rafaeldtinoco, line 229: this comment no longer applies.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -936,6 +961,10 @@ typedef struct kmod_data {
 typedef struct bpf_attach {
     enum bpf_write_user_e write_user;
 } bpf_attach_t;
+typedef struct equality {
+    u64 bitmask;
```
rafaeldtinoco, line 965: need to document better how bitmask and valid_bits work.

geyslan:
> Not a full review but IMO, the `filtername-<scope>=....` syntax is very ugly. Isn't something like `<scope>:filtername` possible? Importantly I think scope should come before the filter name, to me at least it looks more intuitive (and of course omitting the scope should default to scope 1). Not -1ing because if it's only me then it's only me.

`<scope>:filtername=...` is a solution. @NDStrahilevitz @josedonizetti what are your thoughts?

geyslan:
> > Not a full review but IMO, the `filtername-<scope>=....` syntax is very ugly. Isn't something like `<scope>:filtername` possible? Importantly I think scope should come before the filter name, to me at least it looks more intuitive (and of course omitting the scope should default to scope 1). Not -1ing because if it's only me then it's only me.
> 
> `<scope>:filtername=...` is a solution. @yanivagman @josedonizetti what are your thoughts?

In today's discussion we have decided for the @NDStrahilevitz suggestion as a temporary flag syntax. 👍🏼 

`<scope>:filtername=...`
`2:comm=who 42:event=openat`

pkg/utils/utils.go
```diff
@@ -16,3 +16,29 @@ func ParseSymbol(address uint64, table *helpers.KernelSymbolTable) *helpers.Kern
 	hookingFunction.Owner = strings.TrimSuffix(hookingFunction.Owner, "]")
 	return hookingFunction
 }
+
+func HasBit(n uint64, offset uint) bool {
```
NDStrahilevitz, line 20: Probably should make all/most of these generic for any numeric type

pkg/utils/utils.go
```diff
@@ -16,3 +16,29 @@ func ParseSymbol(address uint64, table *helpers.KernelSymbolTable) *helpers.Kern
...
+
+func ClearBit(n *uint64, offset uint) {
+	*n &= ^(1 << offset)
+}
+
+func SetBit(n *uint64, offset uint) {
+	*n |= (1 << offset)
+}
+
+func Min(x, y uint64) uint64 {
```
NDStrahilevitz, line 32: Especially this and `Max` could be useful as generic shorthands.

pkg/cmd/flags/filter.go
```diff
@@ -128,203 +98,271 @@ func PrepareFilter(filtersArr []string) (tracee.Filter, error) {
...
+			NotEqual: []string{},
+		}
+
+		for _, procFilterFlag := range scopeFilterFlags {
+			filterFlag := procFilterFlag.filterFlag[:procFilterFlag.operatorIdx]
 
-		if strings.HasPrefix("container", filterName) {
-			if operatorAndValues == "=new" {
-				err := filter.NewContFilter.Parse("new")
+			if strings.Contains(filterFlag, ".retval") {
```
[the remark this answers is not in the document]
NDStrahilevitz: I think with the new `<scope>:filter` syntax this becomes more reasonable like so:
```
CURRENT: --trace comm-2=ping --trace event-2=openat --trace openat.relval=0 (counter intuitive)
NEW: --trace 2:comm=ping --trace 2:event=openat --trace 2:openat.retval=0
```

tests/integration/integration_test.go
```diff
@@ -186,6 +186,52 @@ func checkSecurityFileOpenExecve(t *testing.T, gotOutput *[]trace.Event) {
 		assert.Equal(t, events.Execve, syscall)
 	}
 }
+
+func checkScope42SecurityFileOpenLs(t *testing.T, gotOutput *[]trace.Event) {
```
NDStrahilevitz, line 190: Great of you to add new tests!

NDStrahilevitz (commented):
Mostly I have comments about readability and adding more comments.
Nice work!

pkg/utils/utils.go
geyslan: I'll leave generic improvements for further work.

pkg/utils/utils.go
geyslan: ditto

pkg/ebpf/events_pipeline.go
```diff
@@ -236,13 +238,115 @@ func (t *Tracee) decodeEvents(outerCtx context.Context) (<-chan *trace.Event, <-
...
+	matchedScopes := event.MatchedScopes
+
 	eventID := events.ID(event.EventID)
 
-	// short circuit by most performant and critical
-	return t.config.Filter.ContextFilter.Filter(event) &&
-		t.config.Filter.RetFilter.Filter(eventID, int64(event.ReturnValue)) &&
-		t.config.Filter.ArgFilter.Filter(eventID, event.Args)
+	for filterScope := range t.config.FilterScopes.UserSpaceMap() {
+		bitOffset := uint(filterScope.ID)
```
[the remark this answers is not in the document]
geyslan: I suppose it's better then.

image

image

pkg/cmd/flags/filter.go
```diff
@@ -128,203 +98,259 @@ func PrepareFilter(filtersArr []string) (tracee.Filter, error) {
...
-		if filterName == "comm" {
-			err := filter.CommFilter.Parse(operatorAndValues)
-			if err != nil {
-				return tracee.Filter{}, err
-			}
-			continue
+		return
+	}
+
+	type processedFilterFlag struct {
```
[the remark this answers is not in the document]
geyslan: Agree. Let's do this in further milestones.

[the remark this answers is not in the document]
geyslan: Right now, I'm not able to visualize scopes separated of filters.

pkg/ebpf/events_pipeline.go
```diff
@@ -237,14 +239,109 @@ func (t *Tracee) decodeEvents(outerCtx context.Context, sourceChan chan []byte)
...
+	matchedScopes := event.MatchedScopes
+
+	for filterScope := range t.config.FilterScopes.UserSpaceMap() {
+		bitOffset := uint(filterScope.ID)
+
+		// Events submitted with matching scopes.
+		// The scope must have its bit cleared when it does not match.
+		if origMatchedScopes != 0 && utils.HasBit(origMatchedScopes, bitOffset) {
+			if filterScope.CheckEventsInUserSpace &&
+				!filterScope.IsEventTraceable(eventID) {
```
yanivagman, line 257: In a future PR, checking if an event is traceable for some scope can be done using a BPF map

pkg/ebpf/tracee.go
```diff
@@ -950,49 +1002,28 @@ func (t *Tracee) populateBPFMaps() error {
...
-		if v != nil {
-			return fmt.Errorf("error setting %v filter: %v", k, v)
+		errMap[UIDFilterMap] = filterScope.UIDFilter.InitBPF(t.bpfModule, scopeID)
+		errMap[PIDFilterMap] = filterScope.PIDFilter.InitBPF(t.bpfModule, scopeID)
+		errMap[MntNSFilterMap] = filterScope.MntNSFilter.InitBPF(t.bpfModule, scopeID)
+		errMap[PidNSFilterMap] = filterScope.PidNSFilter.InitBPF(t.bpfModule, scopeID)
+		errMap[UTSFilterMap] = filterScope.UTSFilter.InitBPF(t.bpfModule, scopeID)
+		errMap[CommFilterMap] = filterScope.CommFilter.InitBPF(t.bpfModule, scopeID)
+		errMap[ContIdFilter] = filterScope.ContIDFilter.InitBPF(t.bpfModule, t.containers, scopeID)
+		errMap[BinaryFilterMap] = filterScope.BinaryFilter.InitBPF(t.bpfModule, scopeID)
```
yanivagman, line 1021: I think we should avoid setting the BPF maps for every scope (get value + set value operations).
Instead, the bitmaps can be calculated in userspace, then initialize the maps once.

pkg/filters/processtree.go
```diff
@@ -132,8 +150,24 @@ func (filter *ProcessTreeFilter) InitBPF(bpfModule *bpf.Module) error {
...
+		if shouldBeTraced {
+			utils.SetBit(&equalInScopes, filterScopeID)
+		} else {
+			utils.ClearBit(&equalInScopes, filterScopeID)
+		}
+		utils.SetBit(&setInScopes, filterScopeID)
+
+		binary.LittleEndian.PutUint64(filterVal[0:8], equalInScopes)
+		binary.LittleEndian.PutUint64(filterVal[8:16], setInScopes)
+		processTreeBPFMap.Update(unsafe.Pointer(&pid), unsafe.Pointer(&filterVal[0]))
```
yanivagman, line 170: Dup code

yanivagman (changes requested):

pkg/ebpf/filters.go
```diff
@@ -41,6 +44,330 @@ type Filter struct {
...
+
+func (fs FilterScopes) Slice() []*FilterScope {
+	var slice []*FilterScope
+
+	for filterScope := range fs.filterScopesMap {
+		slice = append(slice, filterScope)
+	}
+
+	return slice
+}
```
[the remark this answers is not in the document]
geyslan: Nope. It's an implementation first stage leftover.

pkg/ebpf/filters.go
```diff
@@ -41,6 +44,330 @@ type Filter struct {
...
+
+func (fs FilterScopes) IsFilterNetSet() bool {
+	for filterScope := range fs.filterScopesMap {
+		if len(filterScope.NetFilter.Interfaces()) > 0 {
+			return true
+		}
+	}
+
+	return false
+}
```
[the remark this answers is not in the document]
geyslan: Probably, but going to wait @rafaeldtinoco change since this is now based on #2569.

pkg/ebpf/events_pipeline.go
```diff
@@ -237,14 +239,109 @@ func (t *Tracee) decodeEvents(outerCtx context.Context, sourceChan chan []byte)
...
+	origMatchedScopes := event.MatchedScopes
+	matchedScopes := event.MatchedScopes
+
+	for filterScope := range t.config.FilterScopes.UserSpaceMap() {
+		bitOffset := uint(filterScope.ID)
+
+		// Events submitted with matching scopes.
+		// The scope must have its bit cleared when it does not match.
+		if origMatchedScopes != 0 && utils.HasBit(origMatchedScopes, bitOffset) {
+			if filterScope.CheckEventsInUserSpace &&
```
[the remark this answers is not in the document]
geyslan: Whenever we have an event filter. Here: 

image

pkg/ebpf/events_pipeline.go
```diff
@@ -271,15 +368,19 @@ func (t *Tracee) processEvents(ctx context.Context, in <-chan *trace.Event) (<-c
...
 				// For example, it might be that a new cgroup was created, and not by a container runtime,
 				// while we still didn't processed the cgroup_mkdir event and removed the cgroupid from the bpf container map.
-				id := events.ID(event.EventID)
-				// don't skip cgroup_mkdir and cgroup_rmdir so we can derive container_create and container_remove events
-				if id != events.CgroupMkdir && id != events.CgroupRmdir {
+				eventId := events.ID(event.EventID)
+				logger.Debug("false container positive - possible", "event.Timestamp", event.Timestamp, "eventId", eventId)
+
+				// don't skip cgroup_mkdir, cgroup_rmdir and security_sb_mount so we can derive container_create and container_remove events
+				if eventId == events.CgroupMkdir || eventId == events.CgroupRmdir || eventId == events.SecuritySbMount {
```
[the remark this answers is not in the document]
geyslan: It used to be a commit (now squashed) related to an possible issue:

```
pipeline: processEvents check for SecuritySbMount

From ubuntu jammy the SecuritySbMount check is required to treat
container false positives.
```

kerneltests were failing in jammy above and I found out that in jammy that was the cause. It's a workaround to let the kerneltests pass and, for sure, need more investigation.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2009,128 +2116,117 @@ static __always_inline int do_should_trace(program_data_t *p)
...
+        u64 mask = ~p->config->bin_path_filter_enabled_scopes;
+        res &= binary_filter_matches(filter_out_scopes, proc_info) | mask;
+    }
 
-        if (!filter_out)
-            return 0;
+    if (p->config->follow_filter_enabled_scopes) {
+        // trace this task anyway if follow was set by a scope
+        u64 mask = ~p->config->follow_filter_enabled_scopes;
+        res |= proc_info->follow_in_scopes | mask;
```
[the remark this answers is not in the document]
geyslan: Need to test this.

pkg/ebpf/events_pipeline.go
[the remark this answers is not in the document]
geyslan: This will be removed.

pkg/ebpf/events_pipeline.go
```diff
@@ -237,14 +239,109 @@ func (t *Tracee) decodeEvents(outerCtx context.Context, sourceChan chan []byte)
...
+			filterFollow := filterScope.Follow
+
+			// short circuit by most performant and critical
+			if !filterContext || !filterRetVal || !filterArg ||
+				filterUID || filterPID || filterUTS || filterComm ||
+				filterCont || filterContID || filterNewCont || filterMntNS ||
+				filterPidNS || filterNewPid || filterProcTree || filterFollow /* || filterCgroupID */ {
+
+				continue
+			}
```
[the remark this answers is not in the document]
geyslan: There was a lot of changes in the c (kernel land) that improved all logic and I'll get this removed soon.

pkg/ebpf/filters.go
```diff
@@ -22,8 +23,10 @@ const (
 	BinaryFilterMap      = "binary_filter"
 )
 
-type Filter struct {
+type FilterScope struct {
+	ID                int
 	EventsToTrace     []events.ID
+	SetsToTrace       map[string]bool
```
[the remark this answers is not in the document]
geyslan: Ok. It's fixed now.

pkg/ebpf/filters.go
```diff
@@ -41,6 +44,330 @@ type Filter struct {
...
+}
+
+func (fs FilterScopes) ContainerFilterEnabled() bool {
+	for filterScope := range fs.Map() {
+		if filterScope.ContainerFilterEnabled() {
+			return true
+		}
+	}
+
+	return false
```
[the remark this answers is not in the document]
geyslan: It's called once in the setup prior stage. In the future we can improve it thinking in the runtime changes.

pkg/ebpf/events_pipeline.go
```diff
@@ -271,15 +368,19 @@ func (t *Tracee) processEvents(ctx context.Context, in <-chan *trace.Event) (<-c
 				continue
 			}
 
-			if (t.config.Filter.ContFilter.Value() || t.config.Filter.NewContFilter.Enabled()) && event.ContainerID == "" {
+			if t.config.FilterScopes.ContainerFilterEnabled() && event.ContainerID == "" {
```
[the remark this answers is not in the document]
geyslan: I've introduced Recompute() to solve this.

pkg/ebpf/filters.go
```diff
@@ -41,6 +44,330 @@ type Filter struct {
...
+		ContFilter:             filters.NewBoolFilter(),
+		NewContFilter:          filters.NewBoolFilter(),
+		ContIDFilter:           filters.NewContainerFilter(CgroupIdFilterMap),
+		RetFilter:              filters.NewRetFilter(),
+		ArgFilter:              filters.NewArgFilter(),
+		ContextFilter:          filters.NewContextFilter(),
+		ProcessTreeFilter:      filters.NewProcessTreeFilter(ProcessTreeFilterMap),
+		BinaryFilter:           filters.NewBPFBinaryFilter(BinaryFilterMap),
+		Follow:                 false,
+		NetFilter:              &NetIfaces{Ifaces: []string{}},
```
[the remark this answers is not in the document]
geyslan: Waiting https://github.com/aquasecurity/tracee/pull/2569.

pkg/ebpf/filters.go
```diff
@@ -41,6 +44,330 @@ type Filter struct {
...
+	pidFilterMax             uint64
+	uidFilterableInUserSpace bool
+	pidFilterableInUserSpace bool
+}
+
+func NewFilterScopes() *FilterScopes {
+	return &FilterScopes{
+		filterScopesArray:        [MaxFilterScopes]*FilterScope{},
+		filterScopesMap:          map[*FilterScope]int{},
+		filterScopesUserSpaceMap: map[*FilterScope]int{},
```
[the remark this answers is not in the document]
geyslan: I've put comments in the code.

pkg/ebpf/tracee.go
geyslan: I agree. Currently all the logic for updating filter maps is inside each filter type; depending on the type the logic is different: https://github.com/aquasecurity/tracee/issues/2589. For now, I suppose changing `InitBPF` to `UpdateBPF` would make it clearer.

geyslan:
Update:

Most review comments were addressed or at least responded to. Tomorrow's work will be directed towards documenting magical parts of the code.

@NDStrahilevitz I've changed the filter flag parsing. I would appreciate your review on this as well.

@rafaeldtinoco we're synced with the net part.

@yanivagman pr is ready for a new review round.

Tks you all.

pkg/ebpf/events_pipeline.go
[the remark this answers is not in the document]
rafaeldtinoco: I believe this could be related to a recent fix made by @NDStrahilevitz (https://github.com/aquasecurity/tracee/commit/de1bee930fe161c16b5f034bc1999eff96dbb2e3)

rafaeldtinoco:
> @rafaeldtinoco we're synced with the net part.

Working on it now.

pkg/ebpf/events_pipeline.go
```diff
@@ -241,14 +243,69 @@ func (t *Tracee) decodeEvents(outerCtx context.Context, sourceChan chan []byte)
...
+			if !filterScope.RetFilter.Filter(eventID, int64(event.ReturnValue)) {
+				utils.ClearBit(&matchedScopes, bitOffset)
+				continue
+			}
+
+			if !filterScope.ArgFilter.Filter(eventID, event.Args) {
+				utils.ClearBit(&matchedScopes, bitOffset)
+				continue
+			}
+
```
rafaeldtinoco, line 281: could you add a comment explaining why you're running UID and PID filter in here as well now ? Since last shouldProcessEvent did not run them.

pkg/ebpf/events_pipeline.go
```diff
@@ -242,14 +244,82 @@ func (t *Tracee) decodeEvents(outerCtx context.Context, sourceChan chan []byte)
...
+	matchedScopes := event.MatchedScopes
+
+	for filterScope := range t.config.FilterScopes.Map() {
+		bitOffset := uint(filterScope.ID)
+
+		// Events submitted with matching scopes.
+		// The scope must have its bit cleared when it does not match.
+		if utils.HasBit(origMatchedScopes, bitOffset) {
+			// TODO: check if the event is traceable to some scope using a BPF map
+			if !filterScope.IsEventTraceable(eventID) {
```
rafaeldtinoco, line 262: nit: there is an extra line break here.

This logic did not exist in the previous shouldProcessEvent() version.

What happens is the following: this function is called 2 times in the pipeline:

1. decodeEvents
1. deriveEvents

For derived events, let's pick "net_packet_ipv4" for example, the way it works is this:

1. it receives a "net_packet_ip_base" real event (coming from the kernel)
1. this event goes through decodeEvents
1. decodeEvents calls shouldProcessEvent() and it accepts the event.
1. this event goes through all the other pipeline stages
1. this event goes into deriveEvents and can or cannot be outputted (depends if the event is being traced) 
1. the derived event "net_packet_ipv4" will go through shouldProcessEvent() and, if not filtered, will be outputted

Now... in YOUR version.. what happens is the following:

1. it receives a "net_packet_ip_base" real event (coming from the kernel)
1. this event goes through decodeEvents
1. decodeEvents calls shouldProcessEvent() and it gets filtered: the reason why it gets filtered is that you added a logic "IsEventTraceable" that takes into consideration if the event was ever given to a scope. In the case of the derived event, the event is a dependency on the events to be filtered (since derivation did not occur yet). So, at this stage, you have the "net_packet_ip_base" event and you are filtering it because it wasn't selected in any scope.

You should go through all events given to any scope, loop through their event dependencies, and also accept those events, so they can be further derived. This will allow "net_packet_ip_base" to continue in the pipeline until it gets derived and filtered (since it wasn't picked in any scope).

There might be some gotchas in the middle (with the bit cleaning logic)

rafaeldtinoco: Please include this patch: https://github.com/geyslan/tracee/pull/8 to this PR. It solves both issues: 

- derived events can now be filtered by scopes
- network events have matched_scopes correctly set

Example of usage:

```txt
$ sudo ./dist/tracee-ebpf --install-path /home/rafaeldtinoco/tracee/tmp --cache cache-type=mem --cache mem-cache-size=512 --output option:parse-arguments --capabilities bypass=false --output format:json --trace 1:event=net_packet_ipv4 --trace 1:net_packet_ipv4.args.dst="8.8.8.8" --trace 2:event=net_packet_ipv4 --trace 2:net_packet_ipv4.args.dst="1.1.1.1" --trace 3:event=openat --trace 3:openat.args.pathname="/etc/passwd" --trace 4:event=openat --trace 4:openat.args.pathname="/etc/shadow"
{"timestamp":1674015345045854703,"threadStartTime":1674015345043785470,"processorId":16,"processId":1576483,"cgroupId":6539,"threadId":1576483,"parentProcessId":1459289,"hostProcessId":1576483,"hostThreadId":1576483,"hostParentProcessId":1459289,"userId":1000,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"ping","hostName":"fujitsu","containerId":"","containerImage":"","containerName":"","podName":"","podNamespace":"","podUID":"","podSandbox":false,"eventId":"2000","eventName":"net_packet_ipv4","matchedScopes":1,"argsNum":3,"returnValue":0,"stackAddresses":[0],"contextFlags":{"containerStarted":false},"args":[{"name":"src","type":"const char*","value":"192.168.100.2"},{"name":"dst","type":"const char*","value":"8.8.8.8"},{"name":"proto_ipv4","type":"trace.ProtoIPv4","value":{"version":4,"IHL":5,"TOS":0,"length":84,"id":51397,"flags":2,"fragOffset":0,"TTL":64,"protocol":"ICMPv4","checksum":15657,"srcIP":"192.168.100.2","dstIP":"8.8.8.8"}}]}
{"timestamp":1674015346047464836,"threadStartTime":1674015345043785470,"processorId":16,"processId":1576483,"cgroupId":6539,"threadId":1576483,"parentProcessId":1459289,"hostProcessId":1576483,"hostThreadId":1576483,"hostParentProcessId":1459289,"userId":1000,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"ping","hostName":"fujitsu","containerId":"","containerImage":"","containerName":"","podName":"","podNamespace":"","podUID":"","podSandbox":false,"eventId":"2000","eventName":"net_packet_ipv4","matchedScopes":1,"argsNum":3,"returnValue":0,"stackAddresses":[0],"contextFlags":{"containerStarted":false},"args":[{"name":"src","type":"const char*","value":"192.168.100.2"},{"name":"dst","type":"const char*","value":"8.8.8.8"},{"name":"proto_ipv4","type":"trace.ProtoIPv4","value":{"version":4,"IHL":5,"TOS":0,"length":84,"id":51534,"flags":2,"fragOffset":0,"TTL":64,"protocol":"ICMPv4","checksum":15520,"srcIP":"192.168.100.2","dstIP":"8.8.8.8"}}]}
{"timestamp":1674015349234380912,"threadStartTime":1674015349232781700,"processorId":16,"processId":1576533,"cgroupId":6539,"threadId":1576533,"parentProcessId":1459289,"hostProcessId":1576533,"hostThreadId":1576533,"hostParentProcessId":1459289,"userId":1000,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"cat","hostName":"fujitsu","containerId":"","containerImage":"","containerName":"","podName":"","podNamespace":"","podUID":"","podSandbox":false,"eventId":"257","eventName":"openat","matchedScopes":4,"argsNum":4,"returnValue":3,"stackAddresses":null,"contextFlags":{"containerStarted":false},"args":[{"name":"dirfd","type":"int","value":-100},{"name":"pathname","type":"const char*","value":"/etc/passwd"},{"name":"flags","type":"string","value":""},{"name":"mode","type":"mode_t","value":0}]}
```

geyslan: Awesome diving. Thanks @rafaeldtinoco.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -833,16 +836,38 @@ typedef struct ksym_name {
 typedef struct config_entry {
```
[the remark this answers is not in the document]
geyslan: Done.

geyslan:
Tackling those just arrived conflicts now.

pkg/ebpf/events_pipeline.go
geyslan: @rafaeldtinoco https://github.com/aquasecurity/tracee/actions/runs/3951089425

rafaeldtinoco:
Okay, so I tested it entirely using filters and, unfortunately, that did not test the "ingress" (as I had a filter for "dst", which is egress. I discovered ingress did not have the correct matched_scope, making it to be filtered by multi scopes userland filtering.

```diff
diff --git a/pkg/ebpf/c/tracee.bpf.c b/pkg/ebpf/c/tracee.bpf.c
index 49c1a005..179f992f 100644
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -6939,6 +6939,7 @@ struct {
 typedef struct net_task_context {
     struct task_struct *task;
     task_context_t taskctx;
+    u64 matched_scopes;
 } net_task_context_t;
 
 struct {
@@ -6984,6 +6985,7 @@ static __always_inline u64 sizeof_net_event_context_t(void)
 static __always_inline void set_net_task_context(event_data_t *event, net_task_context_t *netctx)
 {
     netctx->task = event->task;
+    netctx->matched_scopes = event->context.matched_scopes;
     __builtin_memset(&netctx->taskctx, 0, sizeof(task_context_t));
     __builtin_memcpy(&netctx->taskctx, &event->context.task, sizeof(task_context_t));
 }
@@ -7195,7 +7197,7 @@ int BPF_KPROBE(cgroup_bpf_run_filter_skb)
     u32 host_tid = p.event->context.task.host_tid;
     bpf_map_update_elem(&entrymap, &host_tid, &entry, BPF_ANY);
 
-    // TODO: https://github.com/aquasecurity/tracee/issues/2221
+    // pick task context from the inode map
     net_task_context_t *netctx = bpf_map_lookup_elem(&inodemap, &inode);
     if (!netctx)
         return 0;
@@ -7218,6 +7220,8 @@ int BPF_KPROBE(cgroup_bpf_run_filter_skb)
     eventctx->eventid = NET_PACKET_IP;                      // will be changed in skb program
     eventctx->stack_id = 0;                                 // no stack trace
     eventctx->processor_id = p.event->context.processor_id; // copy from current ctx
+    eventctx->matched_scopes = netctx->matched_scopes;      // pick matched-scopes from net ctx
+
 
     // inform userland about protocol family (for correct L3 header parsing)...
     struct sock_common *common = (void *) sk;
```

this patc does the trick. I'm sending together with previous patch.

rafaeldtinoco:
https://github.com/geyslan/tracee/pull/9

josedonizetti (commented):
@geyslan This PRs breaks the new `tracee` binary.  Not sure if you want to fix it here or I can take it after you merge. But there are two problems:

1 - https://github.com/aquasecurity/tracee/blob/main/cmd/tracee/main.go#L259 -> This needs to consider the scope, probably using a regex matching so it gets the events for `-t event=anti_debugging` and also `- t 1:event=anti_debuggig`

2 - https://github.com/aquasecurity/tracee/pull/2238/files#diff-70859fceede6c73f664e1a1f54783ef93c29fccefbf64230aaaf76071843cf55R72 -> `IsEventTraceable` is only considering the events passed in the command line, but the problem is that for signatures, they are like derived events, they don't come from the ebpf. So if you pass `-t event=anti_debugging` the ebpf will collect a `ptrace` event, which down the pipe would be converted to a `anti_debugging` event, but because `ptrace` is not on the list, it is not traceable and gets ignored.

rafaeldtinoco:
> So if you pass `-t event=anti_debugging` the ebpf will collect a `ptrace` event, which down the pipe would be converted to a `anti_debugging` event, but because `ptrace` is not on the list, it is not traceable and gets ignored.

Yep, it is the problem I spot recently for derived events, in the case of derived events we're allowing them to pass the decode stage (checking the derivation table, if the event is an event that will generate a derived event) and only filtering at the derivation stage.

We should do similar for rules events.

geyslan:
> @geyslan This PRs breaks the new `tracee` binary. Not sure if you want to fix it here or I can take it after you merge. But there are two problems:
> 
> 1 - https://github.com/aquasecurity/tracee/blob/main/cmd/tracee/main.go#L259 -> This needs to consider the scope, probably using a regex matching so it gets the events for `-t event=anti_debugging` and also `- t 1:event=anti_debuggig`
> 
> 2 - https://github.com/aquasecurity/tracee/pull/2238/files#diff-70859fceede6c73f664e1a1f54783ef93c29fccefbf64230aaaf76071843cf55R72 -> `IsEventTraceable` is only considering the events passed in the command line, but the problem is that for signatures, they are like derived events, they don't come from the ebpf. So if you pass `-t event=anti_debugging` the ebpf will collect a `ptrace` event, which down the pipe would be converted to a `anti_debugging` event, but because `ptrace` is not on the list, it is not traceable and gets ignored.

I've solved the item 1. Item 2 is under analysis.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2008,128 +2162,210 @@ static __always_inline int do_should_trace(program_data_t *p)
...
+        //
+        // considering an event not from a container
+        //
+        // bool_filter_matches(0, false) = 0000 0000
+        //
+        // res                           = 1111 1111 &
+        //                                 0000 0000 |
+        //                                 1110 1111
+        //                                 ---------
+        //                                 1110 1111 (scope 5 unmatched)
```
yanivagman, line 2193: I think this comment is too much and repeats itself.
The comment should mostly explain about mask, which we didn't explain about in the filtering functions. Something like:
`For scopes which has this filter disabled we want to set the matching bits using 'mask'`
I would also put this comment right before computing `res` below

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2008,128 +2162,210 @@ static __always_inline int do_should_trace(program_data_t *p)
...
+        // considering an event from "ping" command
+        //
+        // res                         = 1111 1111 <- ~0ULL
+        // equality_filter_matches()   = 0000 0100
+        //
+        // res                         = 1111 1111 &
+        //                               0000 0100 |
+        //                               1111 0001
+        //                               ---------
+        //                               1111 0101 (scope 2 and 4 unmatched)
```
yanivagman, line 2288: This comment repeats and looks redudant to me

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2008,128 +2162,210 @@ static __always_inline int do_should_trace(program_data_t *p)
...
+    //       0000 1110
+    //       ---------
+    //       0000 0010 (scope 2 match)
+    //
+    // considering an event from "ping" command
+    //
+    // res = 1111 0101 &
+    //       0000 1110
+    //       ---------
+    //       0000 0100 (scope 3 match)
```
yanivagman, line 2353: Again, I don't see a reason for repeating all of this.
The new thing in this line is ANDing with enabled_scopes so I would just explain about that in one sentence:
`Make sure only enabled scopes are set in the bitmask (other bits are invalid)`

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2008,128 +2162,210 @@ static __always_inline int do_should_trace(program_data_t *p)
...
+    if (p->config->bin_path_filter_enabled_scopes) {
+        u64 filter_out_scopes = p->config->bin_path_filter_out_scopes;
+        u64 mask = ~p->config->bin_path_filter_enabled_scopes;
+        res &= binary_filter_matches(filter_out_scopes, proc_info) | mask;
+    }
+
+    if (p->config->follow_filter_enabled_scopes) {
+        // trace this proc anyway if follow was set by a scope
+        u64 mask = p->config->follow_filter_enabled_scopes;
+        res |= proc_info->follow_in_scopes & mask;
```
yanivagman, line 2318: I think we can oneline it

pkg/ebpf/events_pipeline.go
```diff
@@ -276,15 +346,20 @@ func (t *Tracee) processEvents(ctx context.Context, in <-chan *trace.Event) (<-c
...
 				// while we still didn't processed the cgroup_mkdir event and removed the cgroupid from the bpf container map.
-				id := events.ID(event.EventID)
-				// don't skip cgroup_mkdir and cgroup_rmdir so we can derive container_create and container_remove events
-				if id != events.CgroupMkdir && id != events.CgroupRmdir {
+				eventId := events.ID(event.EventID)
+
+				// don't skip cgroup_mkdir, cgroup_rmdir and security_sb_mount so we can derive container_create and container_remove events
+				if eventId != events.CgroupMkdir &&
+					eventId != events.CgroupRmdir &&
+					eventId != events.SecuritySbMount {
```
yanivagman, line 360: I still don't understand why this is needed

pkg/ebpf/tracee.go
```diff
@@ -942,49 +994,28 @@ func (t *Tracee) populateBPFMaps() error {
...
-		if v != nil {
-			return fmt.Errorf("error setting %v filter: %v", k, v)
+		errMap[UIDFilterMap] = filterScope.UIDFilter.UpdateBPF(t.bpfModule, scopeID)
+		errMap[PIDFilterMap] = filterScope.PIDFilter.UpdateBPF(t.bpfModule, scopeID)
+		errMap[MntNSFilterMap] = filterScope.MntNSFilter.UpdateBPF(t.bpfModule, scopeID)
+		errMap[PidNSFilterMap] = filterScope.PidNSFilter.UpdateBPF(t.bpfModule, scopeID)
+		errMap[UTSFilterMap] = filterScope.UTSFilter.UpdateBPF(t.bpfModule, scopeID)
+		errMap[CommFilterMap] = filterScope.CommFilter.UpdateBPF(t.bpfModule, scopeID)
+		errMap[ContIdFilter] = filterScope.ContIDFilter.UpdateBPF(t.bpfModule, t.containers, scopeID)
+		errMap[BinaryFilterMap] = filterScope.BinaryFilter.UpdateBPF(t.bpfModule, scopeID)
```
yanivagman, line 1013: I still see that you loop over all scopes and update the bpf map for every scope.
The value to be update can be calculated once, without the need to get the value from the bpf map, modify, and the update the bpf map again for each scope.

yanivagman (changes requested):

pkg/cmd/flags/filter.go
```diff
@@ -76,49 +77,134 @@ Examples:
...
+
+		filterNameIdx = scopeEndIdx + 1
 	}
-	setFilter := cliFilter{
-		Equal:    []string{},
-		NotEqual: []string{},
+
+	if len(operatorAndValues) >= 2 &&
+		operatorAndValues[0:1] == "!" &&
+		operatorAndValues[1:2] != "=" {
```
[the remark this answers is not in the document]
geyslan: Indeed. Done.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -934,6 +962,14 @@ typedef struct bpf_attach {
     enum bpf_write_user_e write_user;
 } bpf_attach_t;
 
+typedef struct equality {
+    // bitmask with scopes on which a equal '=' filter is set
+    // its offset value will depend on the filter's equality precedence order
```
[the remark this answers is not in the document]
geyslan: As a bitmask, I referred to the specific scope bit as offset. What about?

```suggestion
    // its bit value will depend on the filter's equality precedence order
```

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -934,6 +962,14 @@ typedef struct bpf_attach {
...
+    u64 equal_in_scopes;
+    // bitmask with scopes on which a filter is set
+    u64 set_in_scopes;
```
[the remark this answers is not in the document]
geyslan: I agree. And it goes along with all other fields naming like `config_entry.*filter_enabled_scopes` and `config_entry.enabled_scopes`.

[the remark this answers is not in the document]
geyslan: Done.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1965,40 +2005,154 @@ static __always_inline int init_tailcall_program_data(program_data_t *p, void *c
...
+
+    u64 equal_in_scopes = 0;
+    u64 set_in_scopes = 0;
+    eq_t *equality = bpf_map_lookup_elem(filter_map, &value);
+    if (equality != NULL) {
+        equal_in_scopes = equality->equal_in_scopes;
+        set_in_scopes = equality->set_in_scopes;
+    }
 
+    // there will always be a valid (scope) bitmap in addition to the returned value
```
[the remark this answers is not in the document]
geyslan: This is a leftover from the initial work. Will be removed.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1965,40 +2005,154 @@ static __always_inline int init_tailcall_program_data(program_data_t *p, void *c
...
+    u64 equal_in_scopes = 0;
+    u64 set_in_scopes = 0;
+    eq_t *equality = bpf_map_lookup_elem(filter_map, &value);
+    if (equality != NULL) {
+        equal_in_scopes = equality->equal_in_scopes;
+        set_in_scopes = equality->set_in_scopes;
+    }
+
+    return equal_in_scopes | (filter_out_scopes & ~set_in_scopes);
+}
```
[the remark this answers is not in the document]
geyslan: I agree. Sample calling it for mnt_ns:

```c
  u32 mnt_id = context->mnt_id;
  res &= equality_filter_matches(filter_out_scopes, &mnt_ns_filter, &mnt_id) | mask;
```

Changed. Thanks.

pkg/ebpf/events_pipeline.go
geyslan: The goal is to remove this. I'm going to test it right now.

geyslan: Okay, it got resolved down the road at some point. https://github.com/aquasecurity/tracee/pull/2238/checks

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1965,40 +2005,154 @@ static __always_inline int init_tailcall_program_data(program_data_t *p, void *c
...
-    return filter_out;
+    u64 equal_in_scopes = 0;
+    u64 set_in_scopes = 0;
+    eq_t *equality = bpf_map_lookup_elem(filter_map, key);
+    if (equality != NULL) {
+        equal_in_scopes = equality->equal_in_scopes;
+        set_in_scopes = equality->set_in_scopes;
+    }
+
+    return equal_in_scopes | (filter_out_scopes & ~set_in_scopes);
```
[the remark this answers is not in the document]
geyslan: > Need to check when we use the != operator (since we usually test the = operator)

**-t 2:comm=ls**

config_entry:

```
"comm_filter_enabled_scopes": 2
"comm_filter_out_scopes": 0
```

comm_filter:

```
[{
        "key": {
            "str": "ls"
        },
        "value": {
            "equal_in_scopes": 2,
            "enabled_in_scopes": 2
        }
    }
]
```

```c
return equal_in_scopes | (filter_out_scopes & ~enabled_in_scopes);
 2 | (0 & ~2) = 2 (0010)
```

**-t 2:comm!=ls**

config_entry:

```
"comm_filter_enabled_scopes": 2
"comm_filter_out_scopes": 2
```

comm_filter:

```
[{
        "key": {
            "str": "ls"
        },
        "value": {
            "equal_in_scopes": 0,
            "enabled_in_scopes": 2
        }
    }
]
```

```c
return equal_in_scopes | (filter_out_scopes & ~enabled_in_scopes);
 0 | (2 & ~2) = ..FD (1101)
```

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1965,40 +2005,154 @@ static __always_inline int init_tailcall_program_data(program_data_t *p, void *c
...
+static __always_inline u64 bool_filter_matches(u64 filter_out_scopes, bool val)
 {
-    return filter_out ^ val;
+    // check compute_scopes() for initial info
+    //
+    // e.g.: cmdline: -t 5:container
+    //
+    // considering an event from a container
+    //
+    //   filter_out_scopes = 0000 0000
```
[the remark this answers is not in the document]
geyslan: Based on tests, yes.

`-t 5:container`

config_entry:

```
"cont_filter_enabled_scopes": 16,
"cont_filter_out_scopes": 0,
```

val = **true** (from a container)
```c
filter_out_scopes ^ (val ? ~0ULL : 0)
 0 ^ -1 = -1
```

val = **false** (not from a container)
```c
filter_out_scopes ^ (val ? ~0ULL : 0)
 0 ^ 0 = 0
```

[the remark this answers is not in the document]
geyslan: `bool_filter_matches()` could return just `filter_out_scopes ^ -val;` without any prejudice.

geyslan:
> > 2 - https://github.com/aquasecurity/tracee/pull/2238/files#diff-70859fceede6c73f664e1a1f54783ef93c29fccefbf64230aaaf76071843cf55R72 -> `IsEventTraceable` is only considering the events passed in the command line, but the problem is that for signatures, they are like derived events, they don't come from the ebpf. So if you pass `-t event=anti_debugging` the ebpf will collect a `ptrace` event, which down the pipe would be converted to a `anti_debugging` event, but because `ptrace` is not on the list, it is not traceable and gets ignored.
> 
> I've solved the item 1. Item 2 is under analysis.

For item 2 there's a workaround available in https://github.com/geyslan/tracee/commit/bba37398fb6a3cb3ea1c7ca97cccba06f2dba08a being tested by @josedonizetti

pkg/ebpf/tracee.go
geyslan: > The value to be update can be calculated once, without the need to get the value from the bpf map, modify, and the update the bpf map again for each scope.

Unfortunately it can't, at least with the current design, if I'm not wrong.

Each filterScope has a bunch of filters like UIDFilterMap, PIDFilterMap, CommFilterMap etc. And each one of those can have different keys and values.

The cmdline:

`-t 2:comm=shell.sh -t 4:comm=script.sh,zsh`

builds two scope each with its own set of filters, but with only comm filters enabled and set.

Each iteration works on updating (initializing) the comm map for the respective scope, see:

image

image

So the rest of the work in the UpdateBPF() is to set or clear the scope bit, based on the equality precedence, which must be done for each equality/non-equality. In our example:

comm=shell.sh in the scope 2
comm=script.sh and comm=zsh in the scope 4

Which results in the comm_filter map populated as:

image

To avoid iterating through the filterScope, we must redesign where we keep information about scopes. For now I see no other practical way.

pkg/ebpf/c/tracee.bpf.c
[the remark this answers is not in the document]
yanivagman: So what's the conclusion? To me it seems that we indeed need to change to `filter_out_scopes & enabled_in_scopes);` so we will get `2 & 2` in the last example

pkg/ebpf/c/tracee.bpf.c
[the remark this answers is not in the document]
yanivagman:
```
filter_out_scopes ^ (val ? ~0ULL : 0)
 0 ^ -1 = -1
```

> ... since otherwise all scopes will match, which is wrong

Isn't it indeed what we see here?

pkg/ebpf/tracee.go
yanivagman: > To avoid iterating through the filterScope...

This is not what I meant. We do need to iterate it. What I meant is that we should do the calculation not using the bpf map itself, but using user-defined maps. Get, modify, update can be done on these user maps, and then update the bpf map only once

geyslan: After discussion we've decided that it can be tackled in a future PR.

pkg/ebpf/c/tracee.bpf.c
geyslan: @rafaeldtinoco @NDStrahilevitz @josedonizetti I would appreciate your reply on this matter, since you're not so acquainted to the multiscope logic.

pkg/ebpf/c/tracee.bpf.c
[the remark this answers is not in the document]
geyslan: After tests with `return equal_in_scopes | (filter_out_scopes & enabled_in_scopes);` we got no regressions:

with equal and non equal to the same event filter 

`1:event!=accept4 -t 1:pid=1`

0000000000000001 scope 1 for all events (but accept4) from pid 1

`-t 42:event=accept4 -t 42:pid=1 -t 10:event=accept4 -t 10:pid=1`
 
0000020000000200 scope 42 and 10 for accept4 events from pid 1

image

Then, changing it to `return equal_in_scopes | (filter_out_scopes & enabled_in_scopes);`.

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: I'm okay with removing the redundant comments. I do think having a full example is very important, though.

pkg/ebpf/c/tracee.bpf.c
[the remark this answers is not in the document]
geyslan: > we got no regressions

This statement is wrong. In other test cases, we caught errors, which conclude that the current proposed logic is correct. No changes required.

pkg/ebpf/events_pipeline.go
```diff
@@ -220,16 +221,20 @@ func (t *Tracee) decodeEvents(outerCtx context.Context, sourceChan chan []byte)
...
 				ContextFlags:        parseContextFlags(ctx.Flags),
 			}
 
-			if !t.shouldProcessEvent(evt) {
-				t.stats.EventsFiltered.Increment()
-				continue
+			// base events for derived ones should be filtered in later stage
+			if _, ok := t.eventDerivations[eventId]; !ok {
+				if !t.shouldProcessEvent(&evt) {
+					t.stats.EventsFiltered.Increment()
```
yanivagman, line 235: Isn't it a problem that we don't increment this counter later when calculating derived events (in case it shouldn't be processed)?

pkg/ebpf/events_pipeline.go
```diff
@@ -242,14 +247,79 @@ func (t *Tracee) decodeEvents(outerCtx context.Context, sourceChan chan []byte)
...
 	eventID := events.ID(event.EventID)
+	origMatchedScopes := event.MatchedScopes
+	matchedScopes := event.MatchedScopes
+
+	for filterScope := range t.config.FilterScopes.Map() {
+		bitOffset := uint(filterScope.ID)
+
+		// Events submitted with matching scopes.
+		// The scope must have its bit cleared when it does not match.
+		if utils.HasBit(origMatchedScopes, bitOffset) {
```
yanivagman, line 263: To prevent the extra indentation, we can continue if the scope's bit isn't set.

pkg/ebpf/filters.go
```diff
@@ -37,3 +41,318 @@ type Filter struct {
...
+	}
+
+	return false
+}
+
+// TODO: add locking mechanism as scopes will change at runtime
+type FilterScopes struct {
+	filterScopesArray        [MaxFilterScopes]*FilterScope // underlying filter scopes array
+	filterScopesMap          map[*FilterScope]int          // stores only enabled scopes
+	filterScopesUserSpaceMap map[*FilterScope]int          // stores a reduced map with only user space filtered scopes
```
yanivagman, line 86: Any advantage of setting the key-value like you did here over having a key which is the scope ID and a value which is *FilterScope?
Because I think it is more intuitive and can save a few lines of codes if we use scope ID as the key

tests/integration/integration_test.go
```diff
@@ -186,6 +186,52 @@ func checkSecurityFileOpenExecve(t *testing.T, gotOutput *[]trace.Event) {
...
+		// ls - scope 42
+		assert.Equal(t, "ls", evt.ProcessName)
+		assert.Equal(t, uint64(1<<41), evt.MatchedScopes)
+		arg, err := helpers.GetTraceeArgumentByName(evt, "pathname")
+		require.NoError(t, err)
+		assert.Contains(t, arg.Value, "integration")
+	}
+}
+
+func checkScopes4n2Execve(t *testing.T, gotOutput *[]trace.Event) {
```
yanivagman, line 206: typo in function name?

yanivagman:
Ok, I think we're getting close to the point where we can merge this PR (after addressing the open comments).
I would also like someone else (@NDStrahilevitz/ @rafaeldtinoco ) to review it before we merge since I was part of its writing and I might be missing things that fresh eyes can catch

tests/integration/integration_test.go
geyslan: not quite, but it can be clearer:

```suggestion
func checkExecveOnScopes4And2(t *testing.T, gotOutput *[]trace.Event) {
```

pkg/ebpf/filters.go
geyslan: > I think this is basically using the map as a "set" to avoid duplication (a technique we use in other places too usually done with `map[T]bool/struct{}`).

Yes. This is the motivation.

> I'm not sure if we should actually care about duplicate scopes but I assume that's the reason, as there is an error about such a case.

Since we will be changing scopes at runtime, I assumed that by having pointers as keys we could make the management easier.

pkg/ebpf/events_pipeline.go
geyslan: Raised a commit for that. 👍🏼

pkg/ebpf/events_pipeline.go
```diff
@@ -427,6 +427,7 @@ func (t *Tracee) sinkEvents(ctx context.Context, in <-chan *trace.Event) <-chan
 		for event := range in {
 			// Only emit events requested by the user
 			id := events.ID(event.EventID)
+			event.MatchedScopes &= t.events[id].emit
 			if (t.events[id].emit & event.MatchedScopes) == 0 {
```
yanivagman, line 431: You can now simply check:
`if (event.MatchedScopes) == 0`

geyslan: Good catch. Done.

yanivagman (approved):
I think we're finally there.
All the comments I still have can be handled in a future PR.
LGTM!
Good job @geyslan

pkg/ebpf/tracee.go
```diff
@@ -1534,64 +1573,76 @@ func (t *Tracee) triggerMemDump(event trace.Event) error {
...
-	if printMemDumpFilters == nil {
-		return fmt.Errorf("no address or symbols were provided to print_mem_dump event. " +
-			"please provide it via -t print_mem_dump.args.address=<hex address>" +
-			", -t print_mem_dump.args.symbol_name=<owner>:<symbol> or " +
-			"-t print_mem_dump.args.symbol_name=<symbol> if specifying a system owned symbol")
-	}
+	errArgFilter := make(map[int]error, 0)
 
-	eventHandle := t.triggerContexts.Store(event)
+	for filterScope := range t.config.FilterScopes.Map() {
```
NDStrahilevitz, line 1578: @AsafEitani we need to test the uprobe events with the new scope mechanisms (in particular this call I guess but all of them in general).

pkg/filters/string.go
```diff
@@ -245,12 +244,28 @@ func (filter *BPFStringFilter) InitBPF(bpfModule *bpf.Module) error {
...
 	//Initialize the associated bpfMap
 	//First initialize notEqual values since equality should take precedence
 	for str := range filter.notEqual {
 		byteStr := make([]byte, maxBpfStrFilterSize)
 		copy(byteStr, str)
-		if err := bpfMap.Update(unsafe.Pointer(&byteStr[0]), unsafe.Pointer(&bpfFilterNotEqual)); err != nil {
+
+		var equalInScopes, equalitySetInScopes uint64
+		curVal, err := bpfMap.GetValue(unsafe.Pointer(&byteStr[0]))
+		if err == nil {
```
NDStrahilevitz, line 257: What if `err != nil`? There's no check for that.

pkg/filters/processtree.go
```diff
@@ -96,6 +98,27 @@ func (filter *ProcessTreeFilter) InitBPF(bpfModule *bpf.Module) error {
 		return fmt.Errorf("could not read proc dir: %v", err)
 	}
 
+	updateBPF := func(shouldBeTraced bool, pid uint64) {
+		filterVal := make([]byte, 16)
+		var equalInScopes, equalitySetInScopes uint64
+		curVal, err := processTreeBPFMap.GetValue(unsafe.Pointer(&pid))
+		if err == nil {
```
NDStrahilevitz, line 105: ditto, and I guess in all other filters.

NDStrahilevitz (commented):
So couple general notes:
1. `FilterScope` is a bit of a mouthful and can probably be called just `Scope` and `Scopes`. So a `Scope` is a group of `Filters`. Then there is a clear relation between the two with no chance of misnaming.,
2. There's a lot of heuristic changing of mechanisms depending on filters like the uprobe stuff. @AsafEitani or anyone else from research developing these should test them before merging (currently looks like args from `Diamorphine` are empty also in main, but the initial integrity check is missing when running `4:e=hooked_syscalls`).
3. Potential error check miss in eBPF filters.

pkg/cmd/printer/printer.go
```diff
@@ -146,20 +228,106 @@ func (p tableEventPrinter) Print(event trace.Event) {
 	if p.verbose {
 		switch p.containerMode {
 		case ContainerModeDisabled:
-			fmt.Fprintf(p.out, "%-16s %-16s %-13s %-12d %-12d %-6d %-16s %-7d %-7d %-7d %-16d %-20s ", timestamp, event.HostName, containerId, event.MountNS, event.PIDNS, event.UserID, event.ProcessName, event.ProcessID, event.ThreadID, event.ParentProcessID, event.ReturnValue, eventName)
+			fmt.Fprintf(p.out,
+				"%-16s %016x  %-16s %-13s %-12d %-12d %-6d %-16s %-7d %-7d %-7d %-16d %-20s ",
+				timestamp,
+				event.MatchedScopes,
```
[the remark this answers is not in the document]
geyslan: I'm not opinionated about this, but I think that with line breaks we can identify all the fields and their formats without going back and forth across the screen.

pkg/ebpf/errors.go
```diff
@@ -0,0 +1,23 @@
...
+
+func FilterScopesMaxExceededError() error {
+	return fmt.Errorf("filter scopes maximum exceeded [%d]", MaxFilterScopes)
+}
+
+func FilterScopesOutOfRangeError(idx int) error {
+	return fmt.Errorf("filter scopes index [%d] out-of-range [0-%d]", idx, MaxFilterScopes-1)
+}
+
+func FilterScopeAlreadySetWithDifferentIDError(scope *FilterScope, id int) error {
```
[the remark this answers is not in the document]
geyslan: I'll rename it, thanks.

There's a special case that we need to abort, when we can't set any scope. On the other hand, in the general case, for adding, we can Warn.

pkg/ebpf/filters.go
```diff
@@ -37,3 +41,318 @@ type Filter struct {
...
+}
+
+const MaxFilterScopes = 64
+
+func isIDInRange(id int) bool {
+	return id >= 0 && id < MaxFilterScopes
+}
+
+func (fs FilterScope) IsEventTraceable(event events.ID) bool {
+	if _, found := fs.EventsToTrace[event]; found {
```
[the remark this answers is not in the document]
geyslan: `IsEventTraceable()` helper is a leftover. Now we can match scopes for events in the bpf land. Removing it.

pkg/ebpf/filters.go
```diff
@@ -37,3 +41,318 @@ type Filter struct {
...
+}
+
+// ContainerFilterEnabled returns a bitmask of scopes that have at least one container filter type enabled
+func (fs FilterScopes) ContainerFilterEnabled() uint64 {
+	return atomic.LoadUint64(&fs.containerFiltersEnabled)
+}
+
+// Compute recalculates values, updates flags and fills the reduced user space map
+// It must be called at initialization and at every runtime scopes changes
+func (fs *FilterScopes) Compute() {
```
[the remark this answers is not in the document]
geyslan: You're right.

pkg/ebpf/tracee.go
```diff
@@ -942,49 +1000,28 @@ func (t *Tracee) populateBPFMaps() error {
...
-	errmap[ContIdFilter] = t.config.Filter.ContIDFilter.InitBPF(t.bpfModule, t.containers)
-	errmap[BinaryFilterMap] = t.config.Filter.BinaryFilter.InitBPF(t.bpfModule)
+	for filterScope := range t.config.FilterScopes.Map() {
+		scopeID := uint(filterScope.ID)
+		errMap := make(map[string]error, 0)
 
-	for k, v := range errmap {
-		if v != nil {
-			return fmt.Errorf("error setting %v filter: %v", k, v)
+		errMap[UIDFilterMap] = filterScope.UIDFilter.UpdateBPF(t.bpfModule, scopeID)
```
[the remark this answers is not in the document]
geyslan: Let's also handle this.

pkg/filters/string.go
geyslan: Failure is the right result on the first key lookup as we haven't stored it yet. So we could check curVal == nil instead, but it would rule out the other check, logically speaking, because if err != nil, then curVal == nil, and vice versa.

That's why I opted to check err == nil only; we'll always get an err at the first lookup of some key, and it's alright.

But it will be replaced soon with the pre-computed approach as we discussed.

pkg/filters/processtree.go
geyslan: ditto.

geyslan:
> 1. `FilterScope` is a bit of a mouthful and can probably be called just `Scope` and `Scopes`. So a `Scope` is a group of `Filters`. Then there is a clear relation between the two with no chance of misnaming.,

Agree. We can handle that for sure after this. But any further thoughts are welcome.

pkg/ebpf/filters.go
```diff
@@ -37,3 +41,318 @@ type Filter struct {
...
+		return true
+	}
+
+	return false
+}
+
+// TODO: add locking mechanism as scopes will change at runtime
+type FilterScopes struct {
+	filterScopesArray        [MaxFilterScopes]*FilterScope // underlying filter scopes array
+	filterScopesMap          map[*FilterScope]int          // stores only enabled scopes
```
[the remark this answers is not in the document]
geyslan: > it prevents the same scope object (as in the actual object in memory, since this is a pointer), being used in two separate places.

Indeed. That's the point of our use here, two maps and one array, all linked through scope pointers.

> As such I could construct the same exact `FilterScope` in different pointers and it should work fine (try it with `-t 4:comm=ls -t 5:comm=ls`).

Yes, they are the same in logic. As we didn't draw the next step completely, I preferred to leave it as it is, a different object.

> I assume you are aware of this but it does make the AlreadySet error a bit unnecessary, and should be probably documented for clarity sake.

We still don't know if this type of error should be disregarded. It will depend on the type of API we want to provide for, let's say, a management web interface. Could this interface to pick scopes and reuse them? Would it be clever to reuse the object or just rebuild it? Open to discuss.

rafaeldtinoco:
image

Okay was able to make DNS and HTTP events to work (related to recently added CAPTURE code/rebase). Checking reasons.

rafaeldtinoco:
@geyslan @NDStrahilevitz 

I discovered a capture issue (unrelated to this PR) that would break it. I'm fixing both MAIN and this one, so if it gets merged after this is merged then I have a fix as well. FYIO

rafaeldtinoco:
This is going to be merged at https://github.com/aquasecurity/tracee/pull/2238 with reworded commits. I don't think squashing commits will be beneficial as it contains history about important fixes related to multi scopes code.

rafaeldtinoco:
Merged at https://github.com/aquasecurity/tracee/pull/2626. Congrats @geyslan, @yanivagman, @NDStrahilevitz. Nice team work =D. Super awesome feature.

geyslan:
> Merged at #2626. Congrats @geyslan, @yanivagman, @NDStrahilevitz. Nice team work =D. Super awesome feature.

My sincere thanks to everyone.

Issue #2242, opened 2022-10-18
[BUG] capabilities are being set wrong for all supported environments


rafaeldtinoco:
## Prerequisites

- [x] This affects current development tree (origin/HEAD).

Select one OR another:

- [x] I'm going to create a PR to solve this (assign to yourself).
- [ ] Someone else should solve this.

## Bug description

When working at https://github.com/aquasecurity/tracee/pull/2233, I have discovered that our capabilities feature is/was broken. After discussions with @yanivagman, we have agreed in some other way of doing capabilities enforcement (that will unlock some current on going PRs).

The test is failing in RHEL8 because I believe we currently have a bug at the capabilities logic.

So we are keeping CAP_SYS_ADMIN in NEW kernels (shouldn't it be dropped ?):

![image](https://user-images.githubusercontent.com/7395852/195763263-8981d145-6b8f-4fe1-a428-d8faf2eab691.png)

AND for OLD kernels (that seems correct):

![image](https://user-images.githubusercontent.com/7395852/195763282-a98e809c-53b3-4d69-b06b-956c71bac5f0.png)

and dropping it for RHEL8 only (this is correct, but by accident):

![image](https://user-images.githubusercontent.com/7395852/195763313-7623bc3c-59bb-4eba-ac1e-f442cc593381.png)

PR #2243, opened 2022-10-18
[FEATURE][FIX] New capabilities with rings (replacing old capabilities feature)

rafaeldtinoco:
## Initial Checklist

## Description (git log)

Reduce tracee capabilities during its execution. Do it through
"protection rings":

* Privileged   (ring 0): all capabilities are Effective (almost never)
* Required     (ring 1): only required (by events) capabilities are Effective (during config)
* Requested    (ring 2): single time requested capabilities are Effective (special needs)
* Unprivileged (ring 3): no capabilities are Effective at all (most of the time)

This way, tracee will be most of its time (99%) without any Effective
capability (Unprivileged ring). The Required() ring will set
capabilities needed by Events as Effective until Unprivileged ring is
set.

This change is needed to allow tracee to raise its capabilities for
certain operations (like mounting/umounting cgroupv2 filesystems, for
example).

Documentation provided at docs/deep-drive/capabilities.md.

Fixes: #2242

## Type of change

## How Has This Been Tested?

Multiple executions, in different environments and kernel versions, with command lines similar to:

```
sudo ./dist/tracee-ebpf --containers --output format:json --output option:parse-arguments --output option:detect-syscall --trace container --trace comm=ls --capabilities add="cap_kill" --capabilities add="cap_syslog" --capabilities add="cap_lease" --capabilities drop="cap_bpf"
```

You may add, remove or skip capabilities dropping feature. By not enabling the cmdline, tracee successfully drops the capabilities as needed.

## Final Checklist:

## Git Log Checklist:

My commits logs have:


rafaeldtinoco:
There are some opportunistic style/format fixes/improvements together, but I won't split into new commits just because of that (I believe the opportunistic refactoring is a good practice and want to use more in my PRs).

rafaeldtinoco:
I believe the integration tests are flaky:

image

I have seen this before (and remember someone mentioned it). Was it you @grantseltzer ?

rafaeldtinoco:
@yanivagman per our recent discussion, this PR will unblock:

https://github.com/aquasecurity/tracee/pull/2233

which will be capable of mounting/unmounting cgroupv1/cgroupv2 during startup/exit.

And then unblock:

https://github.com/aquasecurity/tracee/pull/2200

- it fails because of capabilities bug
- it needs cgroupv1/v2 mounting/unmounting

rafaeldtinoco:
I'll fix the permission errors for focalhwe and jammy (recent kernels) tomorrow. Might be missing a corner case for a needed capability in those kernels/envs.

cmd/tracee-ebpf/flags/capabilities.go
```diff
@@ -5,116 +5,71 @@ import (
...
-	var exampleUsesString string
-	for cmd, exp := range exampleUses {
-		cmd = fmt.Sprintf(cmd, CapsMainFlag, AllowFailedDropFlag, CancelDropFlag, AddReqCapsFlag)
-		exampleUsesString += fmt.Sprintf("  %-65s | %s\n", cmd, exp)
-	}
-
-	return fmt.Sprintf("%s\n\nExample uses:\n%s", mainHelp, exampleUsesString)
-}
+Possible options:
+  --capabilities bypass={0,1}               | keep all capabilities during execution time.
```
josedonizetti, line 19: can we use `bypass=[true|false]`? At least for me I will bug every time I have to think which enables, which doesn't.

rafaeldtinoco:
> @rafaeldtinoco Didn't review it yet but took a quick pick, and something popped into my mind, let me know what you think. I really would prefer that moving-in/moving-out from a ring would be done in a block, without the need to remember to run unprivileged again.

I can give a try, lets see...

rafaeldtinoco:
Yep, I like it:

image

I'll make that change @josedonizetti

cmd/tracee-ebpf/flags/capabilities_test.go
```diff
@@ -1,88 +0,0 @@
...
-
-import (
-	"testing"
-
-	"github.com/aquasecurity/tracee/cmd/tracee-ebpf/flags"
-	"github.com/stretchr/testify/assert"
-	"kernel.org/pub/linux/libs/security/libcap/cap"
-)
-
-func TestPrepareCapsConfig(t *testing.T) {
```
NDStrahilevitz, line 11: I think this test was pretty easy to adjust to the new logic which is even simpler. 
At the very least let's open an issue to add it back, can be marked as a "good first issue" too.

pkg/capabilities/capabilities.go
```diff
@@ -1,88 +1,295 @@
...
+
+func (c *Capabilities) getProc() {
+	c.have = cap.GetProc()
+}
+
+func (c *Capabilities) setProc() {
+	err := c.have.SetProc()
 	if err != nil {
-		return err
+		panic(err)
```
NDStrahilevitz, line 223: Why panic here?

pkg/capabilities/capabilities.go
```diff
@@ -1,88 +1,295 @@
...
+	bypass bool
+}
+
+func NewCapabilities(bypass bool) *Capabilities {
+	c := &Capabilities{}
+	c.initialize(bypass)
+	return c
+}
+
+func (c *Capabilities) initialize(bypass bool) {
```
NDStrahilevitz, line 42: If this can fail (internally it panics if yes so it can), I think it would be better to keep to our conventions and return an error here. The constructor `NewCapabilities` would likewise have an error which we would return from tracee.

pkg/ebpf/tracee.go
```diff
@@ -264,6 +274,11 @@ func New(cfg Config) (*Tracee, error) {
 		events:        GetEssentialEventsList(),
 	}
 
+	// Start capabilities rings
+	t.capabilities = capabilities.NewCapabilities(t.config.Capabilities.BypassCaps)
+	t.capabilities.Unprivileged() // move to ring3 right away
```
NDStrahilevitz, line 279: Maybe this should just be the default behavior, so `NewCapabilities` returns an unprivileged `capabilities` object. Then it is a defined behavior that privileged actions is only opt in inside their scope and will not occur anywhere else.

pkg/capabilities/capabilities.go
```diff
@@ -1,88 +1,295 @@
...
+const (
+	Privileged   ringType = iota // ring0 (all capabilities enabled, startup/shutdown)
+	Required                     // ring1 (needed capabilities only: config time)
+	Requested                    // ring2 (temporary specific capabilities)
+	Unprivileged                 // ring3 (no capabilities: runtime)
+)
+
+type Capabilities struct {
+	have   *cap.Set
+	all    map[cap.Value]map[ringType]bool
```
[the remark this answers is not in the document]
rafaeldtinoco: Nice catch Nadav, you are right.

pkg/capabilities/capabilities.go
```diff
@@ -1,88 +1,295 @@
...
+	c.set(Required, values...) // populate ring1 (Required)
 }
 
-func checkRequired(caps set, reqCaps []cap.Value) error {
-	var missingCaps []cap.Value
-	for _, c := range reqCaps {
-		exist, err := caps.GetFlag(cap.Effective, c)
-		if err != nil {
-			return err
+func (c *Capabilities) RequireByString(values ...string) error {
```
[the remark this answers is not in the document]
rafaeldtinoco: Yep, I had doubts about the "FromString" methods, since they're not part of the object's core. I'll change it.

pkg/capabilities/capabilities.go
rafaeldtinoco: We can handle the errors up to last scope and then exit with a Clean() but I thought... no capabilities might get me into a situation where I can't clean =D.

pkg/capabilities/capabilities.go
rafaeldtinoco: I'm ok with that change.

pkg/ebpf/tracee.go
rafaeldtinoco: True. This is a leftover from my last change (adding scopes) so.. having "Unprivileged" by default is definitly the way to go.

rafaeldtinoco:
Thanks for the review @NDStrahilevitz, I'll change things, and we can merge next week (as we're releasing a "special" release, with changes coming tomorrow, this Friday).

cmd/tracee-ebpf/flags/capabilities_test.go
rafaeldtinoco: https://github.com/aquasecurity/tracee/issues/2258

pkg/capabilities/capabilities.go
rafaeldtinoco: I don't think not being able to handle capabilities is a recoverable error, so I haven't returned any error and panic'ed. Closing this comment and keeping the other opened if you have more comments on this.

pkg/capabilities/capabilities.go
NDStrahilevitz: We can always panic on the failure outside (in `tracee.Init()`), but there's also merit to the idea of not letting any attack surface at all here. @yanivagman WDYT?

rafaeldtinoco:
BTW, test showing fork()/exec() Without any privileges after bound dropping (independently of inheritable):

image


image

pkg/capabilities/capabilities.go
yanivagman: That's a good question.
In the future, this won't necessarily be called only from Init, right?
Anyway, I don't like much the option of panicing Tracee. It can happen that something will fail due to some bug, and then we crash Tracee although we could have possibly continue the execution with the existing capabilties.
I'm in favor of logging an error here.

rafaeldtinoco:
I'll rebase and address comments from @yanivagman tomorrow, so we can merge this before the next release (if others agree).

pkg/capabilities/capabilities.go
rafaeldtinoco: Alright! Lets error instead!

docs/deep-dive/capabilities.md
```diff
@@ -0,0 +1,65 @@
+# Special: Environment Capabilities
+
+**tracee-ebpf** and **tracee-rules** both try to reduce their capabilities
+during their execution. The way they do is through "protection rings":
+
+* Privileged   (ring 0): all capabilities are Effective (almost never)
```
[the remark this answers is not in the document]
rafaeldtinoco: Added an intro quoting parts of kernel docs and explaining the meanings.

cmd/tracee-ebpf/main.go
```diff
@@ -155,28 +166,32 @@ func main() {
...
 			if err != nil {
 				return fmt.Errorf("failed preparing BPF object: %w", err)
 			}
 
 			cfg.ChanEvents = make(chan trace.Event, 1000)
-			// We buffer the error channel because we may want to publish errors before we start flusing this channel
-			cfg.ChanErrors = make(chan error, 10)
+			cfg.ChanErrors = make(chan error, 10) // buffer to allow next errors without blocking
+
+			// Createe Tracee Singleton
```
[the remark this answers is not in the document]
rafaeldtinoco: Tracee Createes :P

pkg/capabilities/capabilities.go
```diff
@@ -1,88 +1,290 @@
...
-		if err != nil {
-			return err
-		}
-		if !exist {
-			missingCaps = append(missingCaps, c)
-		}
+// Public Methods
+
+// Privileged is a protection ring with all caps set as Effective.
+func (c *Capabilities) Privileged(cb func() error) error {
```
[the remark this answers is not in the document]
rafaeldtinoco: We should. I missed adding because of a context switch (likely) and never came back to adding it there. I believe you are right regarding libcap behavior!!! Lets do it!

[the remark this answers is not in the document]
rafaeldtinoco: Im fixing this.

josedonizetti (commented):
@rafaeldtinoco when booting tracee without any capabilities (not using sudo), we get a segmentation error that is misleading:
```
./dist/tracee-ebpf 
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x1 addr=0xffffffc0 pc=0x7fb55469455b]

runtime stack:
runtime.throw({0x171a7f5?, 0x0?})
	/snap/go/9951/src/runtime/panic.go:1047 +0x5d fp=0x7fb5067fb8c8 sp=0x7fb5067fb898 pc=0x94ce1d
runtime.sigpanic()
	/snap/go/9951/src/runtime/signal_unix.go:819 +0x369 fp=0x7fb5067fb918 sp=0x7fb5067fb8c8 pc=0x963e49
...
```

can we double check tracee is getting called with enough caps, and if not return a nicer error?

rafaeldtinoco:
> can we double check tracee is getting called with enough caps, and if not return a nicer error?

```
$ TRACEE_LOGGER_LVL=debug ./dist/tracee-ebpf --containers --output format:json --output option:parse-arguments --output option:detect-syscall --trace container --trace comm=ls --capabilities add="cap_kill" --capabilities add="cap_lease,,," --capabilities drop="cap_syslog" --capabilities bypass=0
{"level":"debug","ts":1666375394.1403742,"msg":"osinfo","OSReleaseField":"ID","OS_KERNEL_RELEASE":"manjaro"}
{"level":"debug","ts":1666375394.1404386,"msg":"osinfo","OSReleaseField":"ID_LIKE","OS_KERNEL_RELEASE":"arch"}
{"level":"debug","ts":1666375394.1404557,"msg":"osinfo","OSReleaseField":"BUILD_ID","OS_KERNEL_RELEASE":"rolling"}
{"level":"debug","ts":1666375394.1404705,"msg":"osinfo","OSReleaseField":"PRETTY_NAME","OS_KERNEL_RELEASE":"\"Manjaro Linux\""}
{"level":"debug","ts":1666375394.1404862,"msg":"osinfo","OSReleaseField":"KERNEL_RELEASE","OS_KERNEL_RELEASE":"5.15.74-3-MANJARO"}
{"level":"debug","ts":1666375394.1405168,"msg":"osinfo","OSReleaseField":"ARCH","OS_KERNEL_RELEASE":"x86_64"}
{"level":"debug","ts":1666375394.1415193,"msg":"osinfo","security_lockdown":"none"}
{"level":"debug","ts":1666375394.1503236,"msg":"capabilities change","pkg":"capabilities"}
{"level":"debug","ts":1666375394.1505644,"msg":"capabilities change","pkg":"capabilities"}
{"level":"debug","ts":1666375394.150603,"msg":"enabling","pkg":"capabilities","cap":"cap_syslog"}
{"level":"debug","ts":1666375394.150662,"msg":"capabilities change","pkg":"capabilities"}
{"level":"debug","ts":1666375394.3082733,"msg":"capabilities change","pkg":"capabilities"}
{"level":"debug","ts":1666375394.3083105,"msg":"enabling","pkg":"capabilities","cap":"cap_ipc_lock"}
{"level":"debug","ts":1666375394.3083243,"msg":"enabling","pkg":"capabilities","cap":"cap_sys_resource"}
{"level":"debug","ts":1666375394.3083296,"msg":"enabling","pkg":"capabilities","cap":"cap_lease"}
{"level":"debug","ts":1666375394.308336,"msg":"enabling","pkg":"capabilities","cap":"cap_sys_admin"}
{"level":"debug","ts":1666375394.3083417,"msg":"enabling","pkg":"capabilities","cap":"cap_kill"}
libbpf: map 'fd_arg_path_map': failed to create: Operation not permitted(-1)
{"level":"debug","ts":1666375394.3091176,"msg":"capabilities change","pkg":"capabilities"}
{"level":"fatal","ts":1666375394.3094344,"msg":"app","error":"error initializing Tracee: failed to load BPF object: operation not permitted"}
```
Weird, checking...

rafaeldtinoco:
Got it:

image

josedonizetti (approved):

rafaeldtinoco:
@josedonizetti and I paired remotely to close this for the release (and tested together in multiple envs).

Issue #2073, opened 2022-08-16
[BUG] Flaky integration test


grantseltzer:
## Prerequisites

- [ ] This affects latest released version.
- [x] This affects current development tree (origin/HEAD).
- [x] There isn't an issue describing the bug. (Partially, other issues with this test)

Select one OR another:

- [x] I'm going to create a PR to solve this (assign to yourself).
- [ ] Someone else should solve this.

## Bug description

Running the `do_a_file_write` integration test fails. When I run it on my machine there's a value mismatch because it just reads in the first event for `magic_write` and compares it to a hard coded slice of bytes. This doesn't work if another magic_write event occurs elsewhere on the system. We need to filter for the exact event properly.

## Steps to reproduce

Steps to reproduce the issue:

1. `sudo make test-integration`

grantseltzer:
This is fixed by #2253

yanivagman:
@grantseltzer what's the status of this one?

PR #2250, opened 2022-10-18
[MAINT] integration: fix integration tests

rafaeldtinoco:
This is a mitigation until https://github.com/aquasecurity/tracee/issues/2073 is solved.


rafaeldtinoco:
@grantseltzer FYI, I'm merging this once tests pass (so other PRs depending on the test fix don't fail because of it).

Issue #2177, opened 2022-09-20
Remove need for parse-arguments


NDStrahilevitz:
Currently we require `-o option:parse-arguments` for signatures to function properly. This makes both the signatures and the event pipeline less efficient.

As a first step, let's remove argument parsing and use the raw arguments values in signatures.
This means we should update the golang signatures to use the libbpfgo consts and remove `parse-arguments` from being enabled by default with the pipeline. libbpfgo already have argument parsing helpers in a separate module so they can be used as enum values in signatures.

Next steps (not part of this issue) include:

- [ ] ~Find a way to export libbpfgo constants to rego and cel-go~ Add a way to keep parsed arguments for cases where REGO signatures are wanted (#3008).
- [ ] #3019

PR #2251, opened 2022-10-19
[FEAT] Args syscall filter

NDStrahilevitz:
## Initial Checklist

- [x] There is an issue describing the need for this PR (#2177).

## Description (git log)

commit 2182908a7ff0fd8c8153624e02f7c591a5249e08 (HEAD -> args_syscall_filter, origin/args_syscall_filter)
```
    filters: parse syscall argument filters by name
    
    This allows generic usage of syscall event names in relevant argument
    filters.
    
    Without this change, syscall ids may differ between architectures.
```

commit 7b2a68f3e06e434663db2c526f2c3be091fd7725
```
    filters: add generic Filter method to interface
    
    1. Add the generic Filter method, this takes any value to the filter.
    In practice, the external method must confirm the type, then pass to an
    internal type-safe method.
    
    2. Add generic type constraints to int and uint filters. This enables
    later down the line to avoid correct type casting for arguments.
```

## Type of change

## How Has This Been Tested?

1) Run `tracee-ebpf -t security_file_open.syscall=execve -o option:detect-syscall -t e=security_file_open`
Confirm that events are outputted only with "execve" syscalls and that most events are filtered:
`Stats: {EventCount:46 EventsFiltered:1322 NetEvCount:0 ErrorCount:0 LostEvCount:0 LostWrCount:0 LostNtCount:0}`

2) Run `tracee-ebpf -t task_rename.syscall=execnotve`
Confirm that tracee fails running.

## Final Checklist:

## Git Log Checklist:

My commits logs have:


NDStrahilevitz:
@rafaeldtinoco I think the check-vet fails here because of a golang version mismatch on your runner machines, could you confirm if they support go.18?

geyslan (commented):
LGTM

rafaeldtinoco:
Starting to review this now.

pkg/filters/args.go
```diff
@@ -92,13 +95,33 @@ func (filter *ArgFilter) Parse(filterName string, operatorAndValues string, even
 		return InvalidEventArgument(argName)
 	}
 
+	var err error
+	if argName == "syscall" {
+		err = filter.parseSyscallFilter(id, operatorAndValues)
+	} else {
+		err = filter.parseFilter(id, argName, operatorAndValues, func() Filter {
+			return NewStringFilter()
```
rafaeldtinoco, line 103: Do you expect to have other types here ? is that the reason why you want a constructor pointer ?

rafaeldtinoco (approved):
@NDStrahilevitz sorry for taking so long for reviewing this but I went back up to when you changed all the filters (needed to catch up with your code for this and further PRs).

So, it all looks good to me (your already done filter work is very good). I left some small questions.

Also, it feels to me that "generics" would be better used if we joined Int and Uint filters into a single object using "generics" (supporting s8, s16, s32, s64, u8, u16, u32, u64), rather than using it to differentiate among 32bit and 64bit versions only.  

Of course I'm not proposing this to be done in this PR, so I'm approving this one but I'm a bit skeptical in the generics usage (at least for this PR, maybe you're expecting to be use in a different way in further PRs).

Cheers!

pkg/filters/uint.go
```diff
@@ -24,16 +25,18 @@ type UIntFilter struct {
...
-func NewUInt32Filter() *UIntFilter {
-	return newUIntFilter(true)
+func NewUInt32Filter() *UIntFilter[uint32] {
+	return newUIntFilter[uint32](true)
 }
 
-func newUIntFilter(is32Bit bool) *UIntFilter {
-	return &UIntFilter{
+func newUIntFilter[T constraints.Unsigned](is32Bit bool) *UIntFilter[T] {
+	return &UIntFilter[T]{
```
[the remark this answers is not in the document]
NDStrahilevitz: Basically if you take a look at the external `Filter` method it does the following:
`filterable, ok := val.(T)` this is basically the magic of the generic type parameter, as long as we construct the correct Filter for the argument type, this type cast should always succeed.

I believe, as you said, that we can merge the Int and Uint filters as well, I haven't thought of that before, nice catch. Let's do that in another PR.

pkg/filters/args.go
NDStrahilevitz: In the future when we better map out argument types (i'm prototyping that in my spare time), we can easily map an argument to a filter constructor.

PR #2253, opened 2022-10-19
Allow tracee-ebpf to trace itself

grantseltzer:
## Initial Checklist

## Description

This adds a configuration option in the tracee filter config to allow tracee-ebpf to trace itself. This does not yet expose a flag at the command line for doing so but I can add it. The original motivation is for the sake of the second commit, allowing the integration tests to generate events, but it could make sense to want to run tracee this way. 

## git log

commit ce3fc067c7feef047aeb77c434ed0dbae045db20 (HEAD -> allow-self-trace, pfork/allow-self-trace)
Date:   Tue Oct 18 19:27:58 2022 -0400

    Fix flaky magic_write test (wasn't triggering self)
    
commit 1d3c06aac5e26069dc6ed2c253f9e19b9fcbc59d
Date:   Tue Oct 18 19:27:33 2022 -0400

    Add an option to Tracee config for tracing own PID
    
Fixes: #issue_number

## Type of change

## How Has This Been Tested?

- [ ] integration_test.go

- `make test-integration`

### DIFF
--- a/cmd/tracee-ebpf/flags/filter.go
+++ b/cmd/tracee-ebpf/flags/filter.go
@@ -71,7 +71,8 @@ Examples:
   --trace openat.pathname=/tmp*                                | only trace 'openat' events that have 'pathname' prefixed by "/tmp"
   --trace openat.pathname!=/tmp/1,/bin/ls                      | don't trace 'openat' events that have 'pathname' equals /tmp/1 or /bin/ls
   --trace comm=bash --trace follow                             | trace all events that originated from bash or from one of the processes spawned by bash
-  --trace net=docker0 			                       | trace the net events over docker0 interface
+  --trace self                                                 | trace all events that originated from the running tracee process
+  --trace net=docker0                                          | trace the net events over docker0 interface
 
 
 Note: some of the above operators have special meanings in different shells.
@@ -274,6 +275,11 @@ func PrepareFilter(filtersArr []string) (tracee.Filter, error) {
 
 		if strings.HasPrefix("follow", filterFlag) {
 			filter.Follow = true
+			continue
+		}
+
+		if strings.HasPrefix("self", filterFlag) {
+			filter.TraceSelf = true
 			continue
 		}
 		return tracee.Filter{}, InvalidFilterOptionError(filterFlag)
--- a/pkg/ebpf/tracee.go
+++ b/pkg/ebpf/tracee.go
@@ -778,7 +778,11 @@ func (t *Tracee) populateBPFMaps() error {
 
 	cZero := uint32(0)
 	configVal := make([]byte, 208)
-	binary.LittleEndian.PutUint32(configVal[0:4], uint32(os.Getpid()))
+	selfPid := 0
+	if !t.config.Filter.TraceSelf {
+		selfPid = os.Getpid()
+	}
+	binary.LittleEndian.PutUint32(configVal[0:4], uint32(selfPid))
 	binary.LittleEndian.PutUint32(configVal[4:8], t.getOptionsConfig())
 	binary.LittleEndian.PutUint32(configVal[8:12], t.getFiltersConfig())
 	binary.LittleEndian.PutUint32(configVal[12:16], uint32(t.containers.GetCgroupV1HID()))
--- a/tests/integration/integration_test.go
+++ b/tests/integration/integration_test.go
@@ -2,8 +2,11 @@ package integration
 
 import (
 	"context"
+	"fmt"
+	"io/ioutil"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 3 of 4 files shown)


tests/integration/integration_test.go
```diff
@@ -264,7 +262,7 @@ func Test_EventFilters(t *testing.T) {
 			ctx, cancel := context.WithCancel(context.Background())
 			filter, err := flags.PrepareFilter(tc.filterArgs)
 			require.NoError(t, err)
-
+			filter.TraceSelf = true
```
NDStrahilevitz, line 265: Instead of setting this by default, shouldn't we have a cli option to enable this?
Similar to `-t follow` we can have `-t self`.

grantseltzer: Yea makes sense, I added this.

cmd/tracee-ebpf/flags/filter.go
```diff
@@ -276,6 +278,12 @@ func PrepareFilter(filtersArr []string) (tracee.Filter, error) {
 			filter.Follow = true
 			continue
 		}
+
+		if strings.HasPrefix("self", filterFlag) {
+			fmt.Println("WILL TRACE SELF!")
```
NDStrahilevitz, line 283: Leftover

pkg/ebpf/tracee.go
```diff
@@ -778,7 +778,11 @@ func (t *Tracee) populateBPFMaps() error {
 
 	cZero := uint32(0)
 	configVal := make([]byte, 208)
-	binary.LittleEndian.PutUint32(configVal[0:4], uint32(os.Getpid()))
+	selfPid := 0
```
NDStrahilevitz, line 781: @rafaeldtinoco @yanivagman If I'm not mistaken this is the scheduler's PID right? so this might be a problem.

pkg/ebpf/tracee.go
```diff
@@ -778,7 +778,11 @@ func (t *Tracee) populateBPFMaps() error {
...
+	if t.config.Filter.TraceSelf {
+		selfPid = os.Getpid()
```
NDStrahilevitz, line 783: So the logic [here](https://github.com/aquasecurity/tracee/blob/main/pkg/ebpf/c/tracee.bpf.c#L1822)  is what determines self tracing.
So if we want to enable self tracing we actually need to pass some impossible PID here (-1 or the MAX_PID+1 on the system).
Am I missing something in this assessment?

NDStrahilevitz (changes requested):

pkg/ebpf/tracee.go
grantseltzer: I'm very curious to learn if that's true

pkg/ebpf/tracee.go
grantseltzer: That would be the case if you're correct about the scheduler being an invisible pid 0. But it appears to work just fine as locally, so I imagine it's not a concern?

cmd/tracee-ebpf/flags/filter.go
grantseltzer: woops

pkg/ebpf/tracee.go
NDStrahilevitz: I'm curious how it works right now then, because it appears that you send the correct ID (tracee's actual ID) if you want it to trace itself, which should actually block the tracing instead of allowing it.
Changing the `==` to `!=` in the bpf code would change that though.

pkg/ebpf/tracee.go
NDStrahilevitz: Based on this https://www.baeldung.com/linux/process-pid-0

pkg/ebpf/tracee.go
grantseltzer: You are right, I didn't confirm it was working correctly, thanks for catching that.

pkg/ebpf/tracee.go
rafaeldtinoco: I believe that is not currently (modern kernels) accurate. If you follow x86_64 arch (at least) logic:

1. x86_64_start_kernel()
2. x86_64_start_reservations()
3. start_kernel()

at the very end of start_kernel() you will see:

image

Ending up with: rest_init() function.

If you read rest_init() logic, you will see that:

1. The initial task (PID=1) is created, and it will run kernel_init() when scheduled, which will call the init process:
```C
	pid = user_mode_thread(kernel_init, NULL, CLONE_FS);
```

2. Then a kernel thread, called kthreadadd, is created and that is a task that will always be in charge of creating more kernel threads, as needed, in an infinite loop (scheduled in and out), as you can see in the function "kthreadadd".
```C
	pid = kernel_thread(kthreadd, NULL, CLONE_FS | CLONE_FILES);
```

3. The "kernel_init" function waits for a completion at the beginning, that only happens after "kthreadadd" task exists:
```C
	complete(&kthreadd_done);
```
This way the user task will have that needed kernel task already running if needed (and kernel threads available).

4. At the end, that initial task (to be considered PID=0, being Parent of PID=1 and Parent ID of the kthreadadd task) will enter in an ininite idling loop:
```C
	cpu_startup_entry(CPUHP_ONLINE);
```
which is basically "do nothing":
```C
void cpu_startup_entry(enum cpuhp_state state)
{
	arch_cpu_idle_prepare();
	cpuhp_online_idle(state);
	while (1)
		do_idle();
}
```

pkg/ebpf/tracee.go
rafaeldtinoco: Yep, I would use MAX_PID value instead of a logic with "0", because zero might also mean the "host_tid" value inside eBPF program is not initialized (yet), and lose the last PID tracing.

IMO: DO NOT overflow the unsigned int with -1 or MAX_PID+1, that is implementation dependent and might not work as expected.

rafaeldtinoco: Nah, I think it is fine to use 0 for the eBPF logic... so I'm +1 here.

rafaeldtinoco (dismissed):
LGTM (check comments)

rafaeldtinoco:
@NDStrahilevitz since you began the review, I'll let you merge if you agree to my comments (I'm +1). If you don't have time or don't see this, I can merge this tomorrow.

cmd/tracee-ebpf/flags/filter.go
```diff
@@ -276,6 +277,11 @@ func PrepareFilter(filtersArr []string) (tracee.Filter, error) {
 			filter.Follow = true
 			continue
 		}
+
+		if strings.HasPrefix("self", filterFlag) {
```
NDStrahilevitz, line 281: We also have the `set` filter check which uses a prefix so I think this should come before that one and have a full equality check.

tests/integration/integration_test.go
NDStrahilevitz, line 265: Now we need to remove this and add the `-t self`.

tests/integration/integration_test.go
```diff
@@ -202,11 +200,11 @@ func Test_EventFilters(t *testing.T) {
...
 		eventFunc  func(*testing.T, *[]trace.Event)
 	}{
-		// {
-		// 	name:       "do a file write",
-		// 	filterArgs: []string{"event=magic_write"},
-		// 	eventFunc:  checkMagicwrite,
-		// },
+		{
+			name:       "do a file write",
+			filterArgs: []string{"event=magic_write"},
```
NDStrahilevitz, line 205: `-t self` goes here.

NDStrahilevitz (commented):
Final fixes before I +1, otherwise LGTM.

yanivagman (changes requested):
1. I'm not sure I understand the motivation for tracing self. Tracing self can cause infinite loops in some cases (depends on the chosen events), and we should avoid this. If someone has a usecase to trace tracee he can run another instance of tracee.

2. Tracee PID is being used in other places in the bpf code. For example, when a uprobe is being trigger from Tracee and we want to catch it in the bpf code. You can't simply set it to 0

Can we solve the test issue in another way? Maybe forking a new process?

rafaeldtinoco:
> 2\. Tracee PID is being used in other places in the bpf code. For example, when a uprobe is being trigger from Tracee and we want to catch it in the bpf code. You can't simply set it to 0

Ohhh, sh**, forgot about that. Nice catch.

NDStrahilevitz:
> 1. I'm not sure I understand the motivation for tracing self. Tracing self can cause infinite loops in some cases (depends on the chosen events), and we should avoid this. If someone has a usecase to trace tracee he can run another instance of tracee.
> 2. Tracee PID is being used in other places in the bpf code. For example, when a uprobe is being trigger from Tracee and we want to catch it in the bpf code. You can't simply set it to 0
> 
> Can we solve the test issue in another way? Maybe forking a new process?

(2) can be solved by adding some config bit of `TRACE_SELF` so the check become `host_pid == tracee_pid && config & TRACE_SELF`. 
But I think if forking can be used to solve the test, that would be better, or we could write some specialized trigger binary (I think we have one already for the e2e tests), and we trace that one with `-t comm=binary_name`.

If not, then I think we can keep self tracing with a warning (both at the docs and log level) that it can cause infinite loops and is to be used at the user's discretion.

grantseltzer:
Ok I will put this on the back-burner on until after release since the test is commented out and wasn't providing any value before anyway. I think forking for the event triggering logic (and tracing that PID) will be better.

PR #2256, opened 2022-10-19
Update tracee tags in deployments

grantseltzer:
## Initial Checklist

- [ ] There is an issue describing the need for this PR.
- [x] Git log contains summary of the change.
- [ ] Git log contains motivation and context of the change.
- [ ] If part of an EPIC, PR git log contains EPIC number.
- [ ] If part of an EPIC, PR was added to EPIC description.

## Description (git log)

This updates the git tag for tracee in the deployment scripts in advance of us cutting tracee v0.8.4. This will allow for the correct version to be part of the release and fix helm installations.

Signed-off-by: grantseltzer <grantseltzer@gmail.com>

## Type of change

- [ ] Bug fix (non-breaking change fixing an issue, preferable).
- [x] Quick fix (minor non-breaking change requiring no issue, use with care)
- [ ] Code refactor (code improvement and/or code removal)
- [ ] New feature (non-breaking change adding functionality).
- [ ] Breaking change (cause existing functionality not to work as expected).

PR #2257, opened 2022-10-19
Start documentation changes for kubernetes

grantseltzer:
## Initial Checklist

## Description (git log)

This is the start of changes to documentation in tracee with a particular focus
on changing the entry to be towards the user and less so the contributor. The
idea is also to push the use case of tracee for kubernetes and less on running
locally with docker.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -11,11 +11,17 @@
 
 # Tracee: Runtime Security and Forensics using eBPF
 
-Tracee is a Runtime Security and forensics tool for Linux. It uses Linux **eBPF
-technology** to trace your system and applications **at runtime**, and analyzes
-collected events in order to detect **suspicious behavioral patterns**. It is
-usually delivered as a docker container, but there are other ways you can use
-it (even create your own customized tracee container).
+Tracee is a runtime security and forensics tool for Linux based cloud deployments.
+It uses **eBPF** to trace the host OS and applications **at runtime**, and analyzes
+collected events in order to detect **suspicious behavioral patterns**. It can be
+run as a daemon-set in your kubernetes environment, but is flexible to be run for
+many purposes any Linux based hosts. It can be delivered via helm, as a docker
+container, or as a small set of static binaries.
+
+The goal of Tracee is to serve as an easy to use and effective solution for learning 
+when cloud native attacks occur in your environment. By leveraging Aqua's advanced 
+security research, performant eBPF based detection, and cloud native first
+approach Tracee makes runtime detection accesible, powerful, and effective.
 
 Watch a quick video demo of Tracee:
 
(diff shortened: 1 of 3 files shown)


docs/index.md
```diff
@@ -122,6 +130,21 @@ $ docker run \
...
+1. Install tracee via helm.
+
+See the kubernetes [installation guide].
+
+2. FIXME:
+
+- How are users interacting with Tracee via Kubernetes?
+  - Are signature alerts printed to stderr and then collected via logs?
+  - Is Postee enabled by default and if so how is it configured?
+  - How are signatures configured?
```
grantseltzer, line 146: @josedonizetti @simar7 Perhaps you two can help me with this?

docs/installing/kubernetes.md
```diff
@@ -29,8 +29,8 @@ the detections in your preferred way (e.g. Slack, E-mail, JIRA and more).
 
         ```text
         $ helm repo add aqua-charts https://aquasecurity.github.io/helm-charts
-        $ helm dependency update ./deploy/helm/tracee
-        $ helm install tracee ./deploy/helm/tracee \
+        $ helm dependency update
+        $ helm install tracee aqua/tracee \
```
grantseltzer, line 33: @josedonizetti Can you confirm this is correct (for once #2256 merges)?

grantseltzer:
To discuss:

- Would it be helpful to have a tutorial of installing tracee to a minikube cluster? or something similar?
- There's a video in the README, it would be great if we can record a new demo at the beginning of Kubecon to replace this.

docs/index.md
simar7: sorry don't have a context but I'll share what I know about tracee + k8s:
- tracee is deployed via a helm chart https://github.com/aquasecurity/tracee/blob/main/deploy/helm/tracee/values.yaml
- this helm chart has an included dependency on postee, which is how today people can see their tracee rules events showing up in the form of an UI.
- there's nothing special in terms of configuring signatures, the default ones are loaded to be run.

josedonizetti: @grantseltzer @simar7 I think we should make postee not enabled by default. The first interaction of a user with tracee would be through stdout logs. Than using postee+tracee should be a tutorial, where we can dive more into postee usage. 

Yes, currently all signatures will be loaded by default, and there isn't a way to configure it.

docs/installing/kubernetes.md
josedonizetti:
```
helm repo add aqua https://aquasecurity.github.io/helm-charts/
helm repo update
helm dependency update aqua/tracee
helm install tracee aqua/tracee \
  --namespace tracee-system \
  --create-namespace
```

josedonizetti:
1+ to have a getting started with minikube, or kind

rafaeldtinoco:
> 1+ to have a tutorial with minikube, or kind

Be aware that kind needs to exclude other tracee-ebpf processes from the tracing (as all nodes are containers and not vms or real nodes).

grantseltzer:
Closing in favor of #2272

Issue #2267, opened 2022-10-20
[FEAT] make postee optional, not default


josedonizetti:
## Prerequisites

- [x] This issue is an EPIC issue (add label: EPIC).
- [ ] This issue is an EPIC TASK (add issue to EPIC description).

Select one OR another:

- [x] I'll create a PR to implement this feature (assign to yourself).
- [ ] Someone else should implement this (describe it well).

## Feature description

Postee is enabled by default, which means a user testing tracee for the first time needs to understand how to use postee, and our current documentation is not converting it. Let's make it optional now, and if later we decide to make it enabled by default we should have tutorials converting it on the docs.

PR #2268, opened 2022-10-20
k8s: make postee optional

josedonizetti:
## Initial Checklist

Fix https://github.com/aquasecurity/tracee/issues/2267

- postee optional on helm
- k8 tracee.yaml without posted or falco


rafaeldtinoco (approved):
LGTM

PR #2275, opened 2022-10-20
docs: add overview to docs, contributing sections

josedonizetti:
Add basic overview for sections docs, and contributing. Let me know if you want to add something, or we can keep it simple for now, just to make the release with some information, instead of blank pages.


AnaisUrlichs (approved):
added one minor comment :D

Issue #2219, opened 2022-10-09
[DOCS] The Helm installation guide is confusing


AnaisUrlichs:
## Prerequisites

- [x] I can reproduce the issue in latest released version.
- [ ] I can reproduce the issue in development version (from source tree).
- [x] I checked the documentation and found no answer.
- [x] There isn't an issue describing the bug.

## Bug description 

TLDR I deleted most of the issue template.

I am trying to install tracee in my local Kubernetes cluster through the Helm Chart. 
I have followed the documentation. However, it seems to be a mixture between installing Helm Chart from a registry and from the tracee source code.

When I don't use the Helm Chart in the tracee source repo but the Chart in the offiical Aqua Registry, I do not have access to the following folder 'deploy/helm/tracee'. Thus, I cannot update the dependencies.
Now, I am not sure if this is a must before installing the chart or if just means that I do not have the latest version of e.g. postee being installed alongside Tracee. I.e. what is the reason that we don't ask people to use a Chart in the Aqua Registry but clone the repo instead?
![image](https://user-images.githubusercontent.com/33576047/194759915-b94959ec-570d-4417-8f74-47c9f9c5a265.png?)
However, if this means that I might be installing a newer version of tracee and an older version of postee (referenced in the helm dependencies) and those two might not work together, then that would be an issue. 

So updating the dependencies makes only sense when I install from the source Helm Chart directly as described in the docs but why do I then need to add the Aqua Chart registry before? i.e. I don't think this step is required in the docs:
```
helm repo add aqua-charts https://aquasecurity.github.io/helm-charts
```

josedonizetti:
@AnaisUrlichs The tracee helm chart isn't published on the aqua repository yet. I'm finalizing a few things to have it published today/tmr. As soon as it is there, I'll update the documentation to use the helm repo, instead of installing it from the source code. 

https://github.com/aquasecurity/tracee/issues/2102

josedonizetti:
@AnaisUrlichs chart published, this will work now:

```
```
helm repo add aqua https://aquasecurity.github.io/helm-charts/
helm repo update
helm install tracee aqua/tracee  --namespace tracee-system --create-namespace   --set hostPID=true --version 0.8.3
```
```

rafaeldtinoco:
Close by PR: #2276

PR #2276, opened 2022-10-21
docs: fix helm install

josedonizetti:
Fix for https://github.com/aquasecurity/tracee/issues/2219

### DIFF
--- a/docs/getting-started/installing/kubernetes.md
+++ b/docs/getting-started/installing/kubernetes.md
@@ -18,23 +18,38 @@ the detections in your preferred way (e.g. Slack, E-mail, JIRA and more).
 
 1. Install **Tracee** using **Helm**
 
-    1. Clone the Helm chart:
+	1. Add Aqua chart repository:
 
-        ```text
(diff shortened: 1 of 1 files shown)


docs/getting-started/installing/kubernetes.md
```diff
@@ -18,23 +18,38 @@ the detections in your preferred way (e.g. Slack, E-mail, JIRA and more).
...
+		```
+		$ helm install tracee aqua/tracee \
+				--namespace tracee-system --create-namespace \
+				--set hostPID=true
+		```
+  
+		or install the Helm chart from a local directory:
+
+		```text
+		$ helm repo add aqua https://aquasecurity.github.io/helm-charts/
```
AnaisUrlichs, line 47: do you nee to have the tracee gh repo cloned before after this step? Why does the user need to add the aqua helm registry in this case?

josedonizetti: Because of the postee dependency. It needs to update the dependency of postee, which uses the aqua helm reo

AnaisUrlichs (approved):

PR #2290, opened 2022-10-23
caps: raise privileges for cgroupv1 mount

rafaeldtinoco:
## Description (git log)

This is a temporary mitigation, so cgroupv1 directory can be mounted during docker container initialization. The full code, mounting cgroupv1, cgroupv2 and unmounting them at the end of the execution is
being worked at PR: #2233.

### DIFF
--- a/pkg/ebpf/tracee.go
+++ b/pkg/ebpf/tracee.go
@@ -405,10 +405,15 @@ func (t *Tracee) Init() error {
 
 	// Initialize containers enrichment logic
 
-	t.containers, err = containers.New(t.config.Sockets, "containers_map", t.config.Debug)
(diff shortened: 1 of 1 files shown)


josedonizetti (approved):
Tested on both Digital Ocean, and Google Kubernetes Engine, and the fix works on both!

NDStrahilevitz:
Seems like libbpf was moved back to 1.0.0 here was that on purpose?

rafaeldtinoco:
> Seems like libbpf was moved back to 1.0.0 here was that on purpose?

https://github.com/aquasecurity/tracee/pull/2298

No it wasn't.

PR #2292, opened 2022-10-24
Raise capabilities for privileged operations. DO not drop by default.

rafaeldtinoco:
This is needed for the next release and I need asap, so I'm merging without further reviews. Both, local and e2e tests have passed with this version (and that is why I will merge asap).

### DIFF
--- a/pkg/capabilities/capabilities.go
+++ b/pkg/capabilities/capabilities.go
@@ -11,7 +11,7 @@ import (
 	"kernel.org/pub/linux/libs/security/libcap/cap"
 )
 
-var initialized bool
+var Caps Capabilities // singleton for all packages
 
 const pkgName = "capabilities"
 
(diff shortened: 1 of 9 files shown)


josedonizetti (approved):

pkg/capabilities/capabilities.go
```diff
@@ -11,7 +11,7 @@ import (
 	"kernel.org/pub/linux/libs/security/libcap/cap"
 )
 
-var initialized bool
+var Caps Capabilities // singleton for all packages
```
yanivagman, line 14: This shouldn't be exported

pkg/capabilities/capabilities.go
```diff
@@ -31,28 +31,27 @@ const (
...
+	initialized bool
+	lock        *sync.Mutex // big lock to guarantee all threads are on the same ring
 }
 
-func NewCapabilities(bypass bool) (*Capabilities, error) {
-	c := &Capabilities{}
-	err := c.initialize(bypass)
-	return c, err
+func NewCapabilities(bypass bool) error {
+	Caps = Capabilities{}
```
yanivagman, line 42: Doesn't this break the idea of it being a singleton? (I think this will create a new instance of Capabilities every time NewCapabilities is called, and after the first time, it won't be initialized at all.)

Also, you can use `sync.Once` instead of having the initialized bool.
Something like:

```
var once sync.Once
var caps *Capabilities // singleton for all packages

func NewCapabilities(bypass bool) (*Capabilities, error) {
        var err error
	once.Do(func() {
                caps = &Capabilities{}
                err = caps.initialize(bypass)
	})

	return caps, err
}
```

rafaeldtinoco: > Doesn't this break the idea of it being a singleton?

Clearly I missed returning the reference. Fixing.

rafaeldtinoco: Being fixed at https://github.com/aquasecurity/tracee/pull/2298

PR #2298, opened 2022-10-25
Last touches before v0.9.2 (kubecon)

rafaeldtinoco:
## Description (git log)

libbpf was mistakenly moved to another version during a rushed release.

## Type of change

### DIFF
--- a/cmd/tracee-ebpf/flags/capabilities.go
+++ b/cmd/tracee-ebpf/flags/capabilities.go
@@ -26,19 +26,15 @@ Available capabilities:
 
 func PrepareCapabilities(capsSlice []string) (tracee.CapabilitiesConfig, error) {
 	capsConfig := tracee.CapabilitiesConfig{
-		BypassCaps: false,
+		BypassCaps: true, // bypass capabilities by default
 	}
 
 	for _, slice := range capsSlice {
 		if strings.Contains(slice, "bypass=") {
 			b := strings.TrimPrefix(slice, "bypass=")
-			if b == "1" || b == "true" {
-				capsConfig.BypassCaps = true
-				capsConfig.AddCaps = nil
-				capsConfig.DropCaps = nil
-				return capsConfig, nil
-			}
-			if b != "0" && b != "false" {
+			if b == "0" || b == "false" {
+				capsConfig.BypassCaps = false
+			} else if b != "1" && b != "true" {
 				return capsConfig, fmt.Errorf("bypass should either be true or false")
 			}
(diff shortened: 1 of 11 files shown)


pkg/capabilities/capabilities.go
```diff
@@ -11,7 +11,8 @@ import (
 	"kernel.org/pub/linux/libs/security/libcap/cap"
 )
 
-var Caps Capabilities // singleton for all packages
+var once sync.Once
+var Caps *Capabilities // singleton for all packages
```
yanivagman, line 15: Let's keep Caps unexported and rename NewCapabilities to GetInstance as we already discussed

yanivagman (commented):
I think that GetInstance should not return an error, only the instance (if not initialized, panic). Then you will be able to remove all the error checking, and use the instance on the same line you get it. The only change required for this is to have an Initialize() function that will be called once (which internally calls sync.Once)

rafaeldtinoco:
> I think that GetInstance should not return an error, only the instance (if not initialized, panic). Then you will be able to remove all the error checking, and use the instance on the same line you get it. The only change required for this is to have an Initialize() function that will be called once (which internally calls sync.Once)

I thought about this as well, I'm glad I said it. Will make the change soon and test things over for the release.

cmd/tracee-ebpf/flags/capabilities.go
```diff
@@ -26,19 +26,15 @@ Available capabilities:
...
 			b := strings.TrimPrefix(slice, "bypass=")
-			if b == "1" || b == "true" {
-				capsConfig.BypassCaps = true
-				capsConfig.AddCaps = nil
-				capsConfig.DropCaps = nil
-				return capsConfig, nil
-			}
-			if b != "0" && b != "false" {
+			if b == "0" || b == "false" {
+				capsConfig.BypassCaps = false
```
geyslan, line 36: Would be safe to break here?

pkg/capabilities/capabilities.go
```diff
@@ -31,26 +32,35 @@ const (
...
-	Caps = Capabilities{}
-	return Caps.initialize(bypass)
+func Initialize(bypass bool) error {
+	var err error
+
+	once.Do(func() {
+		caps = &Capabilities{}
+		err = caps.initialize(bypass)
+	})
+	if err != nil {
```
yanivagman, line 48: I think you can simply return err at that point

yanivagman (approved):
LGTM!

PR #2299, opened 2022-10-25
k8s: update version to 0.9.2

josedonizetti:
Preparing for release 0.9.2, updating k8s files to  tag 0.9.2

### DIFF
--- a/deploy/helm/tracee/Chart.yaml
+++ b/deploy/helm/tracee/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: tracee
 description: Linux Runtime Security and Forensics using eBPF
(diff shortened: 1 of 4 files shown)


deploy/helm/tracee/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: tracee
 description: Linux Runtime Security and Forensics using eBPF
-home: https://aquasecurity.github.io/tracee/v0.9.1/
+home: https://aquasecurity.github.io/tracee/v0.9/
```
josedonizetti, line 4: docs doesn't support patch version anymore, so I've removed the patch number

PR #2304, opened 2022-10-25
rules: disable TRC-1022

josedonizetti:
disable TRC-1022, new executable dropped, because it is triggering thousands of false positives on a kubernetes cluster every time a new pod is created. 

I'll create an issue to investigate this further before we reenable it. 

@roikol  ^^


rafaeldtinoco:
@josedonizetti you need to remove from the test list as well I suppose.

rafaeldtinoco (approved):
LGTM

PR #2323, opened 2022-11-01
Run integration test triggers in own PID

### LINKED ISSUE #2073: [BUG] Flaky integration test

grantseltzer:
## Initial Checklist

## Description (git log)

This is a change to the integration tests so that the commands which are meant to trigger events that the tests look for, are actually picked up by tracee. In order for this to happen, the events can't be the same PID as tracee. This is accomplished using syscall.ForkExec() which runs functions in a shell script.

Fixes #2073 

## Type of change

## How Has This Been Tested?

`sudo make test-integration`

### DIFF
--- a/tests/integration/integration_test.go
+++ b/tests/integration/integration_test.go
@@ -2,9 +2,10 @@ package integration
 
 import (
 	"context"
+	"io/ioutil"
 	"os"
-	"os/exec"
 	"strings"
+	"syscall"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


tests/integration/integration_test.go
```diff
@@ -16,49 +17,30 @@ import (
...
+
+	_, err := forkAndExecFunction(doMagicWrite)
+	require.NoError(t, err)
+
+	waitForTraceeOutput(t, gotOutput, time.Now(), true)
+
+	// check tracee output
+	fail := true
+	for _, evt := range *gotOutput {
+		ok := assert.ElementsMatch(t, []byte(evt.EventName), []byte("magic_write"))
```
NDStrahilevitz, line 31: Why are you comparing this way and not just
```suggestion
		ok := assert.Equal(t, evt.Eventname, "magic_write")
```
Also am I reading it correct that if you find magic_write then the test fails? Shouldn't it be the opposite? If not we can also break on the `if ok` check.

tests/integration/integration_test.go
```diff
@@ -291,3 +270,35 @@ func Test_EventFilters(t *testing.T) {
...
+const (
+	doMagicWrite = "do_magic_write"
+	doLs         = "do_ls"
+	doDockerRun  = "do_docker_run"
+)
+
+// forkAndExecFunction runs a function in `tester.sh` in it's own system process.
+// This is so Tracee running in the current pid can pick the command up.
+// It returns the output of the process and a possible error.
+func forkAndExecFunction(funcName string) ([]byte, error) {
```
NDStrahilevitz, line 283: Small suggestion, we can do
`type testFunc string`
and then define the consts above as such and `funcName` as such, which would prevent passing something arbitrary here in the future by mistake.

NDStrahilevitz (approved):
I like this, think the testing code is also cleaner this way, would like an answer on comment 1 before giving a +1, but otherwise LGTM.

tests/integration/integration_test.go
grantseltzer: It is the opposite because fail is set to false. It is a bit backwards but it makes sense here to fail unless all criteria marks it as success.

tests/integration/integration_test.go
grantseltzer: Sounds good!

yanivagman (approved):
LGTM

tests/integration/integration_test.go
grantseltzer: Good point, I cleaned it up

Issue #2174, opened 2022-09-20
[FEAT] create an error perfbuffer so userland logger can log ebpf code errors


rafaeldtinoco:
Select one OR another:

- [ ] I'll create a PR to implement this feature (assign to yourself).
- [x] Someone else should implement this (describe it well).

## Feature description

Currently, we miss errors within eBPF programs because we're not sending them to userland (through bpf_printks()). This feature is about creating a perfbuffer meant to serve error logs (from if(error) {} code from eBPF programs) to userland so the logger (About to be merged) can log them somewhere.

This will help with issues such as:

https://github.com/aquasecurity/tracee/issues/2173

And it is related to:

https://github.com/aquasecurity/tracee/issues/1787

PR #2340, opened 2022-11-08
[FEAT] log ebpf errors

geyslan:
## Description (git log)

568d48b4 log ebpf errors

    Introduce the submit_error() - EBPF helper.
    EBPF errors are processed and logged by processEBPFErrors() go routine.

Fixes: #2174 

## Type of change

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -901,6 +901,13 @@ typedef struct bpf_attach {
     enum bpf_write_user_e write_user;
 } bpf_attach_t;
 
+#define MAX_ERROR_MSG_LEN 64
+
+typedef struct err {
+    u64 ret;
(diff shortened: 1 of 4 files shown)


geyslan:
Due to the ebpf limitations, stack size can easily exceed when using such a structure for errors.

```c
#define MAX_ERROR_MSG_LEN 64

typedef struct err {
    u64 ret;
    char msg[MAX_ERROR_MSG_LEN];
} err_t;
```

Two possible ways to circumvent that:

1. create a map with predefined errors (key: error_id, value: "error message") filled from userspace), so submit_error would take care of receive the return error, lookup the error map and send the whole error entry through the perf buffer.
2. not create a map, just define errors constants in ebpf to be used for submit_error. User space would take care of eat the perfbuffer and lookup a go map for the respective error.

geyslan:
Closed due to this proposal: https://github.com/aquasecurity/tracee/pull/2352

PR #2352, opened 2022-11-10
[FEAT] log ebpf errors

geyslan:
## Description (git log)

commit cfd8611d

    pkg/ebpf: make string copy work in focal 4.19

commit ea4cac9b

    logger: warn event loss
    
    Due ChanErrors removal #2403, warn about it.

commit ff4c077b

    logger: ebpf errors
    
    Introduce EBPF error control mechanism - submission via perf buffer.
    EBPF errors are aggregated before being submitted and are later
    processed and logged by the processBPFErrors() go routine.
    
    This starts the use of bpf error mechanism through some type checking
    like BPF_ERR_MAP_UPDATE_ELEM, BPF_ERR_GET_CURRENT_COMM and
    BPF_ERR_INIT_CONTEXT.

Fixes: #2174 

## Type of change

## How Has This Been Tested?

Toggled a checking (falsed an error) in the ebpf side to be able to send the error and get it in user space.

```shell
❯ TRACEE_LOGGER_LVL=debug sudo -E ./dist/tracee-ebpf -t comm=who
[sudo] password for gg: 
{"level":"debug","ts":1670529435.1558275,"msg":"osinfo","pkg":"urfave","file":"urfave.go","line":46,"OSReleaseField":"ID","OS_KERNEL_RELEASE":"manjaro"}
{"level":"debug","ts":1670529435.1558466,"msg":"osinfo","pkg":"urfave","file":"urfave.go","line":46,"OSReleaseField":"ID_LIKE","OS_KERNEL_RELEASE":"arch"}
{"level":"debug","ts":1670529435.155852,"msg":"osinfo","pkg":"urfave","file":"urfave.go","line":46,"OSReleaseField":"BUILD_ID","OS_KERNEL_RELEASE":"rolling"}
{"level":"debug","ts":1670529435.1558602,"msg":"osinfo","pkg":"urfave","file":"urfave.go","line":46,"OSReleaseField":"KERNEL_RELEASE","OS_KERNEL_RELEASE":"5.15.81-1-MANJARO"}
{"level":"debug","ts":1670529435.1558657,"msg":"osinfo","pkg":"urfave","file":"urfave.go","line":46,"OSReleaseField":"ARCH","OS_KERNEL_RELEASE":"x86_64"}
{"level":"debug","ts":1670529435.1558702,"msg":"osinfo","pkg":"urfave","file":"urfave.go","line":46,"OSReleaseField":"PRETTY_NAME","OS_KERNEL_RELEASE":"\"Manjaro Linux\""}
{"level":"debug","ts":1670529435.1560128,"msg":"osinfo","pkg":"urfave","file":"urfave.go","line":117,"security_lockdown":"none"}


{"level":"debug","ts":1670529436.1772962,"msg":"Failed to add or update a map element","id":2,"type":"BPF_ERR_MAP_UPDATE_ELEM","ret":0,"cpu":12,"file":"./pkg/ebpf/c/tracee.bpf.c","line":3514,"count":1}
{"level":"debug","ts":1670529436.1775048,"msg":"Failed to add or update a map element","id":2,"type":"BPF_ERR_MAP_UPDATE_ELEM","ret":0,"cpu":14,"file":"./pkg/ebpf/c/tracee.bpf.c","line":3514,"count":1}
...
{"level":"debug","ts":1670529440.1198132,"msg":"Failed to add or update a map element","id":2,"type":"BPF_ERR_MAP_UPDATE_ELEM","ret":0,"cpu":0,"file":"./pkg/ebpf/c/tracee.bpf.c","line":3514,"count":4}
^C
End of events stream
Stats: {EventCount:0 EventsFiltered:0 NetEvCount:0 BPFErrorsCount:15 ErrorCount:0 LostEvCount:0 LostWrCount:0 LostNtCount:0 LostBPFErrCount:0}
```

### DIFF
--- a/pkg/ebpf/bpf_error.go
+++ b/pkg/ebpf/bpf_error.go
@@ -1,0 +1,190 @@
+package ebpf
+
+import (
+	"encoding/binary"
+	"fmt"
+	"strings"
+	"unsafe"
+
+	"github.com/aquasecurity/tracee/pkg/logger"
+)
+
+const BPFMaxErrFileLen = 72 // BPF_MAX_ERR_FILE_LEN
+
+type BPFErrorType uint32
+
+const (
+	BPFErrUnspec BPFErrorType = iota // BPF_ERR_UNSPEC
+
+	// tracee functions
+	BPFErrInitContext // BPF_ERR_INIT_CONTEXT
+
+	// bpf helpers functions
+	BPFErrMapUpdateElem  // BPF_ERR_MAP_UPDATE_ELEM
+	BPFErrMapDeleteElem  // BPF_ERR_MAP_DELETE_ELEM
+	BPFErrGetCurrentComm // BPF_ERR_GET_CURRENT_COMM
+	BPFErrTailCall       // BPF_ERR_TAIL_CALL
+)
+
+var stringMap = map[BPFErrorType]string{
+	BPFErrUnspec: "BPF_ERR_UNSPEC",
+
+	// tracee functions
+	BPFErrInitContext: "BPF_ERR_INIT_CONTEXT",
+
+	// bpf helpers functions
+	BPFErrMapUpdateElem:  "BPF_ERR_MAP_UPDATE_ELEM",
+	BPFErrMapDeleteElem:  "BPF_ERR_MAP_DELETE_ELEM",
+	BPFErrGetCurrentComm: "BPF_ERR_GET_CURRENT_COMM",
+	BPFErrTailCall:       "BPF_ERR_TAIL_CALL",
+}
+
+var errorMap = map[BPFErrorType]string{
+	BPFErrUnspec: "Unspecifed BPF error",
+
+	// tracee functions
+	BPFErrInitContext: "Failed to init context",
+
+	// bpf helpers functions
+	BPFErrMapUpdateElem:  "Failed to add or update a map element",
+	BPFErrMapDeleteElem:  "Failed to delete a map element",
+	BPFErrGetCurrentComm: "Failed to get current command",
+	BPFErrTailCall:       "Failed to tail call",
+}
+
+func (b BPFErrorType) String() string {
+	if bpfErr, found := stringMap[b]; found {
+		return bpfErr
+	}
+
+	return stringMap[BPFErrUnspec]
+}
+
+// BPFError struct contains aggregated data about a bpf error origin
+type BPFError struct {
+	id       BPFErrorType
+	logLevel logger.Level
+	count    uint32                 // percpu error occurrences
+	padding  uint32                 // empty space
+	ret      int64                  // return value
+	cpu      uint32                 // cpu number
+	line     uint32                 // line number
+	file     [BPFMaxErrFileLen]byte // filename
+}
+
+func (b BPFError) Error() string {
+	if bpfErr, found := errorMap[BPFErrorType(b.id)]; found {
+		return bpfErr
+	}
+
+	return errorMap[BPFErrUnspec]
+}
+
+func (b BPFError) ID() uint32 {
+	return uint32(b.id)
+}
+
+func (b BPFError) Type() BPFErrorType {
+	return BPFErrorType(b.id)
+}
+
+func (b BPFError) LogLevel() logger.Level {
+	return b.logLevel
+}
+
+func (b BPFError) Count() uint32 {
+	return b.count
+}
+
+func (b BPFError) Return() int64 {
+	return int64(b.ret)
+}
+
+func (b BPFError) CPU() uint32 {
+	return b.cpu
+}
+
+func (b BPFError) Line() uint32 {
+	return b.line
+}
+
+func (b BPFError) File() []byte {
+	return b.file[:]
+}
+
+func (b BPFError) FileAsString() string {
+	errFile := string(b.file[:])
+	errFileEnd := strings.Index(errFile, "\x00")
+
+	return errFile[:errFileEnd]
+}
+
+func (b BPFError) Size() int {
+	return int(unsafe.Sizeof(b.id)) +
+		int(unsafe.Sizeof(b.logLevel)) +
+		int(unsafe.Sizeof(b.count)) +
+		int(unsafe.Sizeof(b.padding)) +
+		int(unsafe.Sizeof(b.ret)) +
+		int(unsafe.Sizeof(b.id)) +
+		int(unsafe.Sizeof(b.cpu)) +
+		int(unsafe.Sizeof(b.line)) +
+		int(len(b.file))
+}
+
+func (b *BPFError) Decode(rawBuffer []byte) error {
+	if len(rawBuffer) < b.Size() {
+		return fmt.Errorf("can't decode error raw data - buffer of %d should have at least %d bytes", len(rawBuffer), b.Size())
+	}
+
+	b.id = BPFErrorType(binary.LittleEndian.Uint32(rawBuffer[0:4]))
+	b.logLevel = logger.Level(binary.LittleEndian.Uint32(rawBuffer[4:8])) // this is an int8 (zapcore.Level)
+	b.count = binary.LittleEndian.Uint32(rawBuffer[8:12])
+	b.padding = binary.LittleEndian.Uint32(rawBuffer[12:16])
+
+	// bpf_error
+	b.ret = int64(binary.LittleEndian.Uint64(rawBuffer[16:24]))
+	b.cpu = binary.LittleEndian.Uint32(rawBuffer[24:28])
+	b.line = binary.LittleEndian.Uint32(rawBuffer[28:32])
+	copy(b.file[:], rawBuffer[32:104])
+	// bpf_error
+
+	return nil
+}
+
+func (t *Tracee) processBPFErrors() {
+	for {
+		select {
+		case rawData := <-t.bpfErrorsChannel:
+			if len(rawData) == 0 {
+				continue
+			}
+
+			bpfErr := BPFError{}
+			if err := bpfErr.Decode(rawData); err != nil {
+				t.handleError(fmt.Errorf("consumeBPFErrorsChannel: decode - %v", err))
+				continue
+			}
+			// This logger timestamp in no way reflects the ebpf error original time
+			// since bpf errors channel is populated with aggregated logs
+			logger.BPFError(bpfErr.Error(), logger.Level(bpfErr.LogLevel()),
+				"id", bpfErr.ID(),
+				"type", bpfErr.Type().String(),
+				"ret", bpfErr.Return(),
+				"cpu", bpfErr.CPU(),
+				"file", bpfErr.FileAsString(),
+				"line", bpfErr.Line(),
+				"count", bpfErr.Count(),
+			)
+			t.stats.BPFErrorsCount.Increment(uint64(bpfErr.Count()))
+
+		case lost := <-t.lostBPFErrChannel:
+			// When terminating tracee-ebpf the lost channel receives multiple "0 lost events" events.
+			// This check prevents those 0 lost events messages to be written to stderr until the bug is fixed:
+			// https://github.com/aquasecurity/libbpfgo/issues/122
+			if lost > 0 {
+				t.stats.LostBPFErrCount.Increment(lost)
+				t.config.ChanErrors <- fmt.Errorf("lost %d ebpf error events", lost)
+			}
+		}
+	}
+}
--- a/pkg/ebpf/bpf_error.go
+++ b/pkg/ebpf/bpf_error.go
@@ -183,7 +183,7 @@ func (t *Tracee) processBPFErrors() {
 			// https://github.com/aquasecurity/libbpfgo/issues/122
 			if lost > 0 {
 				t.stats.LostBPFErrCount.Increment(lost)
-				t.config.ChanErrors <- fmt.Errorf("lost %d ebpf error events", lost)
+				logger.Warn(fmt.Sprintf("lost %d ebpf error events", lost))
 			}
 		}
 	}
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -627,6 +627,8 @@ enum kconfig_key_e
             _val;                                                                                  \
         })
 
+    #define READ_KERN_STR_INTO(dst, src) bpf_probe_read_str((void *) &dst, sizeof(dst), src)
+
     #define READ_USER(ptr)                                                                         \
         ({                                                                                         \
             typeof(ptr) _val;                                                                      \
@@ -653,6 +655,8 @@ enum kconfig_key_e
             bpf_core_read((void *) &_val, sizeof(_val), &ptr);                                     \
             _val;                                                                                  \
         })
+
+    #define READ_KERN_STR_INTO(dst, src) bpf_core_read_str((void *) &dst, sizeof(dst), src)
 
     #define READ_USER(ptr)                                                                         \
         ({                                                                                         \
@@ -927,6 +931,48 @@ typedef struct bpf_attach {
     enum bpf_write_user_e write_user;
 } bpf_attach_t;
 
+enum bpf_error_level
+{
+    BPF_ERR_LOG_LVL_DEBUG = -1,
+    BPF_ERR_LOG_LVL_INFO,
+    BPF_ERR_LOG_LVL_WARN,
+    BPF_ERR_LOG_LVL_ERROR,
+};
+
+enum bpf_error_id
+{
+    BPF_ERR_UNSPEC = 0U, // enforce enum to u32
+
+    BPF_ERR_INIT_CONTEXT,
+
+    BPF_ERR_MAP_UPDATE_ELEM,
+    BPF_ERR_MAP_DELETE_ELEM,
+    BPF_ERR_GET_CURRENT_COMM,
+    BPF_ERR_TAIL_CALL,
+};
+
+#define BPF_MAX_ERR_FILE_LEN 72
+
+typedef struct bpf_error {
+    s64 ret; // return value
+    u32 cpu;
+    u32 line;                        // line number
+    char file[BPF_MAX_ERR_FILE_LEN]; // filename
+} bpf_error_t;
+
+typedef struct bpf_error_count {
+    u32 count;
+    u64 ts; // timestamp
+} bpf_error_count_t;
+
+typedef struct bpf_error_output {
+    enum bpf_error_id id; // type
+    enum bpf_error_level level;
+    u32 count;
+    u32 padding;
+    struct bpf_error err;
+} bpf_error_output_t;
+
 // KERNEL STRUCTS ----------------------------------------------------------------------------------
 
 #ifndef CORE
@@ -991,11 +1037,13 @@ BPF_LRU_HASH(fd_arg_path_map, fd_arg_task_t, fd_arg_path_t, 1024); // store fds 
 BPF_LRU_HASH(bpf_attach_map, u32, bpf_attach_t, 1024);             // holds bpf prog info
 BPF_LRU_HASH(bpf_attach_tmp_map, u32, bpf_attach_t, 1024);         // temporarily hold bpf_attach_t
 BPF_PERCPU_ARRAY(cached_event_data_map, event_data_t, 1);          // cached event data between chained tail calls
-
+BPF_HASH(errors_count, bpf_error_t, bpf_error_count_t, 4096);      // errors count
+BPF_PERCPU_ARRAY(error_output_scratch, bpf_error_output_t, 1);     // error output scratch
 // clang-format on
 
 // EBPF PERF BUFFERS -------------------------------------------------------------------------------
 
+BPF_PERF_OUTPUT(errors, 1024);      // errors submission
 BPF_PERF_OUTPUT(events, 1024);      // events submission
 BPF_PERF_OUTPUT(file_writes, 1024); // file writes events submission
 BPF_PERF_OUTPUT(net_events, 1024);  // network events submission
@@ -1003,6 +1051,53 @@ BPF_PERF_OUTPUT(net_events, 1024);  // network events submission
 // FUNCTIONS DECLARATIONS --------------------------------------------------------------------------
 
 static __always_inline void *get_path_str(struct path *path);
+
+// HELPERS: ERRORS ---------------------------------------------------------------------------------
+
+static __always_inline void do_tracee_error(
+    void *ctx, enum bpf_error_level level, enum bpf_error_id id, s64 ret, u32 line, void *file)
+{
+    if (!ctx || !file)
+        return;
+
+    u32 zero = 0;
+    bpf_error_output_t *err_output = bpf_map_lookup_elem(&error_output_scratch, &zero);
+    if (unlikely(err_output == NULL))
+        return;
+
+    err_output->level = level;
+    err_output->id = id;
+
+    err_output->err.ret = ret;
+    err_output->err.cpu = bpf_get_smp_processor_id();
+    err_output->err.line = line;
+    READ_KERN_STR_INTO(err_output->err.file, file);
+
+    bpf_error_count_t counter_buf = {};
+    counter_buf.count = 1;
+    counter_buf.ts = bpf_ktime_get_ns(); // store the current ts
+    u64 ts_prev = 0;
+
+    bpf_error_count_t *counter = bpf_map_lookup_elem(&errors_count, &err_output->err);
+    if (likely(counter != NULL)) {
+        ts_prev = counter->ts; // store previous ts
+
+        counter->count += 1;
+        counter->ts = counter_buf.ts; // set to current ts
+    } else {
+        counter = &counter_buf;
+        bpf_map_update_elem(&errors_count, &err_output->err, counter, BPF_ANY);
+    }
+
+    // submit error when its cpu occurrence time diff is greater than 2s
+    if ((counter->ts - ts_prev) > (u64) 2000000000) {
+        err_output->count = counter->count;
+        bpf_perf_event_output(ctx, &errors, BPF_F_CURRENT_CPU, err_output, sizeof(*err_output));
+        counter->count = 0; // reset, assuming that the consumer is incrementing
+    }
+}
+
+#define tracee_error(ctx, level, id, ret) do_tracee_error(ctx, level, id, ret, __LINE__, __FILE__);
 
 // HELPERS: DEVICES --------------------------------------------------------------------------------
 
@@ -1694,8 +1789,9 @@ static __always_inline int get_iface_config(int ifindex)
 // INTERNAL: CONTEXT -------------------------------------------------------------------------------
 
 static __always_inline int
-init_context(event_context_t *context, struct task_struct *task, u32 options)
+init_context(void *ctx, event_context_t *context, struct task_struct *task, u32 options)
 {
+    long ret = 0;
     u64 id = bpf_get_current_pid_tgid();
     context->task.start_time = get_task_start_time(task);
     context->task.host_tid = id;
@@ -1708,7 +1804,12 @@ init_context(event_context_t *context, struct task_struct *task, u32 options)
     context->task.pid_id = get_task_pid_ns_id(task);
     context->task.uid = bpf_get_current_uid_gid();
     context->task.flags = 0;
-    bpf_get_current_comm(&context->task.comm, sizeof(context->task.comm));
+    ret = bpf_get_current_comm(&context->task.comm, sizeof(context->task.comm));
+    if (unlikely(ret < 0)) {
+        tracee_error(ctx, BPF_ERR_LOG_LVL_ERROR, BPF_ERR_GET_CURRENT_COMM, ret);
+        return -1;
+    }
+
     char *uts_name = get_task_uts_name(task);
     if (uts_name)
         bpf_probe_read_str(&context->task.uts_name, TASK_COMM_LEN, uts_name);
@@ -1763,15 +1864,18 @@ static __always_inline task_info_t *init_task_info(u32 key, bool *initialized)
 
 static __always_inline int init_event_data(event_data_t *data, void *ctx)
 {
+    long ret = 0;
     int zero = 0;
     data->config = bpf_map_lookup_elem(&config_map, &zero);
     if (unlikely(data->config == NULL))
         return 0;
 
     data->task = (struct task_struct *) bpf_get_current_task();
-
-    init_context(&data->context, data->task, data->config->options);
-
+    ret = init_context(ctx, &data->context, data->task, data->config->options);
+    if (unlikely(ret < 0)) {
+        tracee_error(ctx, BPF_ERR_LOG_LVL_ERROR, BPF_ERR_INIT_CONTEXT, ret);
+        return 0;
+    }
     data->ctx = ctx;
     data->buf_off = sizeof(event_context_t);
     int buf_idx = SUBMIT_BUF_IDX;
@@ -3432,6 +3536,7 @@ int sys_dup_exit_tail(void *ctx)
 SEC("raw_tracepoint/sched_process_fork")
 int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
 {
+    long ret = 0;
     event_data_t data = {};
     if (!init_event_data(&data, ctx))
         return 0;
@@ -3448,7 +3553,9 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
     task.context.tid = get_task_ns_pid(child);
     task.context.host_tid = get_task_host_pid(child);
     task.context.start_time = start_time;
-    bpf_map_update_elem(&task_info_map, &task.context.host_tid, &task, BPF_ANY);
+    ret = bpf_map_update_elem(&task_info_map, &task.context.host_tid, &task, BPF_ANY);
+    if (ret < 0)
+        tracee_error(ctx, BPF_ERR_LOG_LVL_DEBUG, BPF_ERR_MAP_UPDATE_ELEM, ret);
 
     int parent_pid = get_task_host_pid(parent);
     int child_pid = get_task_host_pid(child);
@@ -3460,7 +3567,9 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
     if (data.config->filters & FILTER_PROC_TREE_ENABLED) {
         u32 *tgid_filtered = bpf_map_lookup_elem(&process_tree_map, &parent_tgid);
         if (tgid_filtered) {
-            bpf_map_update_elem(&process_tree_map, &child_tgid, tgid_filtered, BPF_ANY);
+            ret = bpf_map_update_elem(&process_tree_map, &child_tgid, tgid_filtered, BPF_ANY);
+            if (ret < 0)
+                tracee_error(ctx, BPF_ERR_LOG_LVL_DEBUG, BPF_ERR_MAP_UPDATE_ELEM, ret);
         }
     }
 
@@ -3471,7 +3580,9 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
     // perform this check after should_trace() to only add forked childs of a traced parent
     task.follow = true;
     task.new_task = true;
-    bpf_map_update_elem(&task_info_map, &child_pid, &task, BPF_ANY);
+    ret = bpf_map_update_elem(&task_info_map, &child_pid, &task, BPF_ANY);
+    if (ret < 0)
+        tracee_error(ctx, BPF_ERR_LOG_LVL_DEBUG, BPF_ERR_MAP_UPDATE_ELEM, ret);
 
     if (should_submit(SCHED_PROCESS_FORK, data.config) || data.config->options & OPT_PROCESS_INFO) {
         int parent_ns_pid = get_task_ns_pid(parent);
@@ -4895,6 +5006,7 @@ int BPF_KPROBE(trace_security_socket_setsockopt)
 // To delete socket from net map use tid==0, otherwise, update
 static __always_inline int net_map_update_or_delete_sock(void *ctx, struct sock *sk, u32 tid)
 {
+    long ret = 0;
     net_id_t connect_id = {0};
     u16 family = get_sock_family(sk);
 
@@ -4914,7 +5026,10 @@ static __always_inline int net_map_update_or_delete_sock(void *ctx, struct sock 
         if (tid != 0) {
             net_ctx_t net_ctx;
             net_ctx.host_tid = tid;
-            bpf_get_current_comm(&net_ctx.comm, sizeof(net_ctx.comm));
+            ret = bpf_get_current_comm(&net_ctx.comm, sizeof(net_ctx.comm));
+            if (ret < 0)
+                tracee_error(ctx, BPF_ERR_LOG_LVL_DEBUG, BPF_ERR_GET_CURRENT_COMM, ret);
+
             bpf_map_update_elem(&network_map, &connect_id, &net_ctx, BPF_ANY);
         } else {
             bpf_map_delete_elem(&network_map, &connect_id);
@@ -4993,6 +5108,7 @@ int BPF_KPROBE(trace_udpv6_destroy_sock)
 SEC("raw_tracepoint/inet_sock_set_state")
 int tracepoint__inet_sock_set_state(struct bpf_raw_tracepoint_args *ctx)
 {
+    long ret = 0;
     net_id_t connect_id = {0};
     net_ctx_ext_t net_ctx_ext = {0};
 
@@ -5036,7 +5152,10 @@ int tracepoint__inet_sock_set_state(struct bpf_raw_tracepoint_args *ctx)
             if (connect_id.port) {
                 if (!sock_ctx_p) {
                     net_ctx_ext.host_tid = data.context.task.host_tid;
-                    bpf_get_current_comm(&net_ctx_ext.comm, sizeof(net_ctx_ext.comm));
+                    ret = bpf_get_current_comm(&net_ctx_ext.comm, sizeof(net_ctx_ext.comm));
+                    if (ret < 0)
+                        tracee_error(ctx, BPF_ERR_LOG_LVL_DEBUG, BPF_ERR_GET_CURRENT_COMM, ret);
+
                     net_ctx_ext.local_port = connect_id.port;
                     bpf_map_update_elem(&sock_ctx_map, &sk, &net_ctx_ext, BPF_ANY);
                     bpf_map_update_elem(&network_map, &connect_id, &net_ctx_ext, BPF_ANY);
@@ -5063,6 +5182,7 @@ int tracepoint__inet_sock_set_state(struct bpf_raw_tracepoint_args *ctx)
 SEC("kprobe/tcp_connect")
 int BPF_KPROBE(trace_tcp_connect)
 {
+    long ret = 0;
     event_data_t data = {};
     if (!init_event_data(&data, ctx))
         return 0;
@@ -5089,7 +5209,10 @@ int BPF_KPROBE(trace_tcp_connect)
     }
 
     net_ctx_ext.host_tid = data.context.task.host_tid;
-    bpf_get_current_comm(&net_ctx_ext.comm, sizeof(net_ctx_ext.comm));
+    ret = bpf_get_current_comm(&net_ctx_ext.comm, sizeof(net_ctx_ext.comm));
+    if (ret < 0)
+        tracee_error(ctx, BPF_ERR_LOG_LVL_DEBUG, BPF_ERR_GET_CURRENT_COMM, ret);
+
     net_ctx_ext.local_port = connect_id.port;
     bpf_map_update_elem(&sock_ctx_map, &sk, &net_ctx_ext, BPF_ANY);
 
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -1071,7 +1071,12 @@ static __always_inline void do_tracee_error(
     err_output->err.ret = ret;
     err_output->err.cpu = bpf_get_smp_processor_id();
     err_output->err.line = line;
-    READ_KERN_STR_INTO(err_output->err.file, file);
+
+    u64 fsize = __builtin_strlen(file);
+    if (unlikely(fsize >= BPF_MAX_ERR_FILE_LEN))
+        fsize = BPF_MAX_ERR_FILE_LEN - 1;
+    __builtin_memcpy(err_output->err.file, file, fsize);
+    err_output->err.file[fsize] = '\0';
 
     bpf_error_count_t counter_buf = {};
     counter_buf.count = 1;
--- a/pkg/ebpf/tracee.go
+++ b/pkg/ebpf/tracee.go
@@ -177,12 +177,15 @@ type Tracee struct {
 	eventsPerfMap     *bpf.PerfBuffer
 	fileWrPerfMap     *bpf.PerfBuffer
 	netPerfMap        *bpf.PerfBuffer
+	bpfErrorsPerfMap  *bpf.PerfBuffer
 	eventsChannel     chan []byte
 	fileWrChannel     chan []byte
 	netChannel        chan []byte
+	bpfErrorsChannel  chan []byte
 	lostEvChannel     chan uint64
 	lostWrChannel     chan uint64
 	lostNetChannel    chan uint64
+	lostBPFErrChannel chan uint64
 	bootTime          uint64
 	startTime         uint64
 	stats             metrics.Stats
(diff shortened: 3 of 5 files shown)


pkg/metrics/stats.go
```diff
@@ -7,13 +7,15 @@ import (
 
 // When updating this struct, please make sure to update the relevant exporting functions
 type Stats struct {
```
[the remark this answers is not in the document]
geyslan: `RegisterPrometheus()` updated.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -967,10 +986,40 @@ BPF_LRU_HASH(bpf_attach_tmp_map, u32, bpf_attach_t, 1024);         // temporaril
...
+    err->_func[MAX_ERROR_FUNC_LEN - 1] = '\0';
+    err->_line = line;
+
+    bpf_perf_event_output(ctx, &errors, BPF_F_CURRENT_CPU, err, sizeof(*err));
+}
+
+#define submit_error(ctx, err)                                                                     \
+    do {                                                                                           \
+        do_submit_error(ctx, err, (void *) __FUNCTION__, __LINE__);                                \
+    } while (0);
```
NDStrahilevitz, line 1011: Why is this in a `do while` and not just a simple statement?

pkg/ebpf/errors.go
```diff
@@ -0,0 +1,226 @@
...
+	BPF_ERR_MAP_UPDATE_ELEM BPFErrorType = iota + 1
+	BPF_ERR_MAP_DELETE_ELEM
+	BPF_ERR_GET_CURRENT_COMM
+	BPF_ERR_TAIL_CALL
+
+	BPF_ERR_UNSPEC BPFErrorType = math.MaxUint32
+)
+
+func (b BPFErrorType) String() string {
+	stringMap := map[BPFErrorType]string{
```
NDStrahilevitz, line 25: In general these kind of maps should stay out of functions since they are reallocated per call.

pkg/ebpf/errors.go
```diff
@@ -0,0 +1,226 @@
...
+
+	if bpfErr, found := stringMap[b]; found {
+		return bpfErr
+	}
+
+	return stringMap[BPF_ERR_UNSPEC]
+}
+
+func (b BPFError) Error() string {
+	errorMap := map[BPFErrorType]string{
```
NDStrahilevitz, line 42: ditto

pkg/ebpf/c/tracee.bpf.c
geyslan: Ah sure. Remainder of the submit_new_error that has multiple lines. Getting rid of the do while wrapper for this. Tks.

geyslan:
I made some declaration reordering and moved errors.go to bpf/kernel/errors.go, not pushed yet.

@NDStrahilevitz @yanivagman @rafaeldtinoco

About naming I would like to know from you what would we pick: `BPFError` or `KernelError`?

geyslan:
> I made some declaration reordering and moved errors.go to bpf/kernel/errors.go, not pushed yet.
> 
> @NDStrahilevitz @yanivagman @rafaeldtinoco
> 
> About naming I would like to know from you what we would pick: `BPFError` or `KernelError`?
> 
> That choice would imply conformity to:
> 
> * BPF_ERR_MAP_UPDATE_ELEM vs KERNEL_ERR_MAP_UPDATE_ELEM
> * struct bpf_error vs struct kernel_error
> * ...

The `BPFError` was chosen. Making changes accordingly.

pkg/ebpf/bpf_errors.go
```diff
@@ -0,0 +1,228 @@
...
+
+func (t *Tracee) processBPFErrors() {
+	for {
+		select {
+		case rawData := <-t.bpfErrorsChannel:
+			if len(rawData) == 0 {
+				continue
+			}
+
+			bpfErr := NewBPFError()
```
NDStrahilevitz, line 203: Since we export the struct anyway and only use this in decoding, I think the constructor is a bit redundant.

NDStrahilevitz (changes requested):

pkg/ebpf/bpf_errors.go
geyslan: I agree. At start this would ensure that the error id is set to BPF_ERR_UNSPEC, which is math.MaxUint32. Perhaps changing BPF_ERR_UNSPEC to 0 is the best option here.

NDStrahilevitz: Yep, usually when you do these kind of golang "enums" you set the unspecified value to iota exactly to have it as the default value.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -967,10 +986,37 @@ BPF_LRU_HASH(bpf_attach_tmp_map, u32, bpf_attach_t, 1024);         // temporaril
...
+        return;
+
+    bpf_probe_read_kernel_str((void *) &err->_func, sizeof(err->_func), func);
+    err->_func[BPF_MAX_ERR_FUNC_LEN - 1] = '\0';
+    err->_line = line;
+
+    bpf_perf_event_output(ctx, &errors, BPF_F_CURRENT_CPU, err, sizeof(*err));
+}
+
+#define submit_error(ctx, err) do_submit_error(ctx, err, (void *) __FUNCTION__, __LINE__);
```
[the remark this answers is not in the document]
geyslan: I implemented these two macros for different cases. The bpf_error structure resides on the stack, so for functions with a single error check and submission we can rely on submit_new_error() which takes care of the instantiation of bpf_error. For cases where we need to submit errors in more than one line in the function, the ideal is to reuse a single pre-allocated bpf_error structure. And for that cases we can use submit_error().

I'm changing the signature of the submit_error() to `submit_error(ctx, err, err_id, err_ret)`, so its usage can be as easy as:

```c
    struct bpf_error err;
    long ret;
...
    ret = bpf_map_update_elem(&task_info_map, &task.context.host_tid, &task, BPF_ANY);
    if (ret < 0)
        submit_error(ctx, err, BPF_ERR_MAP_UPDATE_ELEM, ret);
...
            ret = bpf_map_update_elem(&process_tree_map, &child_tgid, tgid_filtered, BPF_ANY);
            if (ret < 0)
                submit_error(ctx, err, BPF_ERR_MAP_UPDATE_ELEM, ret);

```

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -967,10 +986,37 @@ BPF_LRU_HASH(bpf_attach_tmp_map, u32, bpf_attach_t, 1024);         // temporaril
...
+
+#define submit_new_error(ctx, err_id, err_ret)                                                     \
```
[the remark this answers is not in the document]
geyslan: Using `do { } while (0);` to make a macro function always do what it means to do.

pkg/ebpf/c/tracee.bpf.c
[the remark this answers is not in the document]
NDStrahilevitz: Got it, please include a comment explaining the usage difference between the two.

pkg/ebpf/bpf_errors.go
```diff
@@ -0,0 +1,228 @@
...
+
+func (b BPFError) lineSize() int {
+	return int(unsafe.Sizeof(b.line))
+}
+
+func (b BPFError) functionSize() int {
+	return int(len(b.function))
+}
+
+func (b BPFError) Size() int {
```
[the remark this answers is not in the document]
geyslan: I suppose that Size() returns a compile-time calculated constant, but I need to be sure.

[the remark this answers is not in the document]
NDStrahilevitz: Because it uses dynamic unsafe.Sizeof checks i'm not sure it does but either way we may see bottlenecks in other areas (and this part may be insignificant), this just put the idea in my mind.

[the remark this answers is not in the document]
geyslan: > Because it uses dynamic unsafe.Sizeof checks i'm not sure it does but either way we may see bottlenecks in other areas (and this part may be insignificant), this just put the idea in my mind.

I got your point and thought about that as well.

It seems that the compiler takes care and calculates the return for us: https://go.godbolt.org/z/Kc3jbWanM

```asm
        TEXT    main.BPFError.Size(SB), NOSPLIT|ABIInternal, $0-64
        FUNCDATA        $0, gclocals·g2BeySu+wFnoycgXfElmcg==(SB)
        FUNCDATA        $1, gclocals·g2BeySu+wFnoycgXfElmcg==(SB)
        FUNCDATA        $5, main.BPFError.Size.arginfo1(SB)
        MOVL    $64, AX
        RET
        TEXT    main.BPFError.SizeStatic(SB), NOSPLIT|ABIInternal, $0-64
        FUNCDATA        $0, gclocals·g2BeySu+wFnoycgXfElmcg==(SB)
        FUNCDATA        $1, gclocals·g2BeySu+wFnoycgXfElmcg==(SB)
        FUNCDATA        $5, main.BPFError.SizeStatic.arginfo1(SB)
        MOVL    $64, AX
        RET
```

pkg/ebpf/bpf_errors.go
```diff
@@ -0,0 +1,228 @@
...
+
+type BPFErrorType uint32
+
+const (
+	BPF_ERR_MAP_UPDATE_ELEM BPFErrorType = iota + 1
+	BPF_ERR_MAP_DELETE_ELEM
+	BPF_ERR_GET_CURRENT_COMM
+	BPF_ERR_TAIL_CALL
+
+	BPF_ERR_UNSPEC BPFErrorType = math.MaxUint32
```
yanivagman, line 23: nit: Since 0 is the deafult value used in many cases, you can set unspec to be 0 (iota) as @NDStrahilevitz suggested in one of his comments.

yanivagman (approved):
LGTM

geyslan:
I had to reduce `BPF_MAX_ERR_FUNC_LEN` the to 40 bytes to pass `noncore` tests. The error was related to stack size exceeded (512 bytes).

yanivagman:
One thing that we should be aware of when adding a new bpf error is the following:
If an error happens too frequently (e.g. map_update_element fails because map is full in some hot path) it will cause the errors perf buffer to get filled quickly and may consume a lot of cpu resources when consuming these events in userspace.

We should consider adding a mechanism that will turn off the submission and logging of such errors when they happen too frequently (e.g. more than 10 errors per second), and re enable it some time later

NDStrahilevitz:
> One thing that we should be aware of when adding a new bpf error is the following: If an error happens too frequently (e.g. map_update_element fails because map is full in some hot path) it will cause the errors perf buffer to get filled quickly and may consume a lot of cpu resources when processing these events in userspace.
> 
> We should consider adding a mechanism that will turn off the submission and logging of such errors when they happen too frequently (e.g. more than 10 errors per second), and re enable it some time later

Maybe we should have log levels set (possibly with separate perf buffers too?), and then we set a log level from userspace and in ebpf we have some method `LOG_EBPF_ERROR(level, msg)` which internally will do a log level check and if relevant submit to the perf buffer.
So things like `map_update_element` failures may fit DEBUG level better than an error you always want.

yanivagman:
> Maybe we should have log levels set (possibly with separate perf buffers too?), and then we set a log level from userspace and in ebpf we have some method `LOG_EBPF_ERROR(level, msg)` which internally will do a log level check and if relevant submit to the perf buffer. So things like `map_update_element` failures may fit DEBUG level better than an error you always want.

map_update_element error can be critical if a map is full, since things will stop working correctly. The problem is that after we submitted this error once, it can repeat too frequently, and we should find a mechanism to reduce the amount of repeated errors once they appear too frequently

geyslan:
> map_update_element error can be critical if a map is full, since things will stop working correctly. The problem is that after we submitted the error once, it can repeat too frequently, and we should find a way to reduce the amount of repeated errors once they happen

Just brainstorming.

We can create a map like `BPF_HASH(bpf_error_threshold, u32, settings, 1024);` populated from user space with the line of the error (key) and settings as `{ u32 threshold; u32 count; }` (value); count would be updated from bpf right after the submission to the perfbuffer.

In bpf before submission it would lookup for that threshold and count. The other requirement to that proposal is to decrement the count after the perfbuffer consumption in user space.

Overall processing increase:

- only one initialization of bpf_error_threshold
- for all error submission:
  - (ebpf) one bpf lookup to get the value, one bpf update to set the count (if threshold is not reached).
  - (user space) one bpf lookup and one bpf update decrementing count after the error receiving.

Besides that bpf lookups and updates, it would demand bpf map locking... so I think it would not be the right path.

pkg/ebpf/bpf_error_decoder.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	d.delimiter = newDelimiter
+
+	return nil
+}
+
+// decodeBuffer decodes rawBuffer updating fields
+// - Fields must be passed as pointers
+// - The fields order matters
+func decodeBuffer(rawBuffer []byte, fields ...interface{}) error {
```
NDStrahilevitz, line 39: Ok, so this is a bit similar to how the `binary.Read` method works minus the reflection option, which we've used in the past but moved from.
We don't really need the switch case type versatility here since we know ahead of time what our field and type order is in the buffer, so we should do the same here as we do in bufferdecoder in that regard, without passing in `fields ...interface{}`, it's unnecessary flexibility.

geyslan: I simplified the logic in bpf_error.go, getting rid of the outer (file) decoder. The decoder test was removed as well, as it's trivial now.

pkg/ebpf/bpf_error.go
```diff
@@ -0,0 +1,159 @@
...
+				continue
+			}
+
+			logger.Error(bpfErr.Error(),
+				"id", bpfErr.ID(),
+				"type", bpfErr.Type().String(),
+				"ret", bpfErr.Return(),
+				"func", bpfErr.FunctionAsString(),
+				"line", bpfErr.Line(),
+			)
```
geyslan, line 146: @NDStrahilevitz @yanivagman I suppose we don't want bpf error being printing to the standard output, right? Unless any occurrence of it is a case of panic. Should we logger.Debug() here or create another logger to output through Error() to stderr (since the pkg one default is to stdout)?

geyslan: Ignore that. I assume that the user must redirect the error to not taint the proper output.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1663,7 +1738,10 @@ init_context(event_context_t *context, struct task_struct *task, u32 options)
     context->task.pid_id = get_task_pid_ns_id(task);
     context->task.uid = bpf_get_current_uid_gid();
     context->task.flags = 0;
-    bpf_get_current_comm(&context->task.comm, sizeof(context->task.comm));
+    ret = bpf_get_current_comm(&context->task.comm, sizeof(context->task.comm));
+    if (ret < 0)
+        tracee_error(ctx, options, BPF_ERR_GET_CURRENT_COMM, ret);
```
NDStrahilevitz, line 1743: Should we fail here?

geyslan: Agree; then failing `init_event_data()` in sequence.

geyslan:
This https://github.com/aquasecurity/tracee/pull/2352/commits/0c42e3ccb472657e71280ff17a7a8f0d79172da2 solved the issue under discussion related to insns limit exceeded. Tomorrow I'll install a focal64 vm and make sure simple assignment works as expected.

@yanivagman @NDStrahilevitz @AlonZivony

geyslan:
> This [0c42e3c](https://github.com/aquasecurity/tracee/commit/0c42e3ccb472657e71280ff17a7a8f0d79172da2) solved the issue under discussion related to insns limit exceeded. Tomorrow I'll install a focal64 vm and make sure simple assignment works as expected.
> 
> @yanivagman @NDStrahilevitz @AlonZivony

Everything seems to work fine in focal (5.4.0) too.

```shell
vagrant@ubuntu-focal:/vagrant$ uname -a
Linux ubuntu-focal 5.4.0-132-generic #148-Ubuntu SMP Mon Oct 17 16:02:06 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

```shell
vagrant@ubuntu-focal:/vagrant$ sudo ./dist/tracee-ebpf --install-path=/tmp/tracee -t comm=who -o option:aggregate-bpf-errors
{"level":"error","ts":1669721710.0727417,"msg":"Failed to get current command","id":4,"type":"BPF_ERR_GET_CURRENT_COMM","ret":0,"file":"./pkg/ebpf/c/tracee.bpf.c","line":1746,"count":19214}
{"level":"error","ts":1669721710.0728297,"msg":"Failed to init context","id":1,"type":"BPF_ERR_INIT_CONTEXT","ret":-1,"file":"./pkg/ebpf/c/tracee.bpf.c","line":1813,"count":19220}
{"level":"error","ts":1669721712.0722592,"msg":"Failed to get current command","id":4,"type":"BPF_ERR_GET_CURRENT_COMM","ret":0,"file":"./pkg/ebpf/c/tracee.bpf.c","line":1746,"count":30121}
{"level":"error","ts":1669721712.072304,"msg":"Failed to init context","id":1,"type":"BPF_ERR_INIT_CONTEXT","ret":-1,"file":"./pkg/ebpf/c/tracee.bpf.c","line":1813,"count":30127}
{"level":"error","ts":1669721714.085526,"msg":"Failed to get current command","id":4,"type":"BPF_ERR_GET_CURRENT_COMM","ret":0,"file":"./pkg/ebpf/c/tracee.bpf.c","line":1746,"count":52742}
{"level":"error","ts":1669721714.0855682,"msg":"Failed to init context","id":1,"type":"BPF_ERR_INIT_CONTEXT","ret":-1,"file":"./pkg/ebpf/c/tracee.bpf.c","line":1813,"count":52748}
{"level":"error","ts":1669721716.073009,"msg":"Failed to get current command","id":4,"type":"BPF_ERR_GET_CURRENT_COMM","ret":0,"file":"./pkg/ebpf/c/tracee.bpf.c","line":1746,"count":58021}
{"level":"error","ts":1669721716.0730498,"msg":"Failed to init context","id":1,"type":"BPF_ERR_INIT_CONTEXT","ret":-1,"file":"./pkg/ebpf/c/tracee.bpf.c","line":1813,"count":58027}
^C
End of events stream
Stats: {EventCount:0 EventsFiltered:0 NetEvCount:0 BPFErrorsCount:116048 ErrorCount:0 LostEvCount:0 LostWrCount:0 LostNtCount:0 LostBPFErrCount:0}
```

Change only for testing:

```diff
vagrant@ubuntu-focal:/vagrant$ git diff
diff --git a/pkg/ebpf/c/tracee.bpf.c b/pkg/ebpf/c/tracee.bpf.c
index 38205ba..340ad78 100644
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -1742,7 +1742,7 @@ init_context(void *ctx, event_context_t *context, struct task_struct *task, u32
     context->task.uid = bpf_get_current_uid_gid();
     context->task.flags = 0;
     ret = bpf_get_current_comm(&context->task.comm, sizeof(context->task.comm));
-    if (unlikely(ret < 0)) {
+    if (unlikely(ret >= 0)) {
         tracee_error(ctx, options, BPF_ERR_GET_CURRENT_COMM, ret);
         return -1;
     }
```

NDStrahilevitz (approved):
LGTM, awesome work @geyslan

pkg/ebpf/bpf_error.go
```diff
@@ -0,0 +1,223 @@
...
+
+func (b BPFErrorType) String() string {
+	if bpfErr, found := stringMap[b]; found {
+		return bpfErr
+	}
+
+	return stringMap[BPF_ERR_UNSPEC]
+}
+
+type BPFError struct {
```
AlonZivony, line 59: I think it would be good if you documented this struct and its members

pkg/ebpf/bpf_error.go
```diff
@@ -0,0 +1,223 @@
...
+	"fmt"
+	"math"
+	"strings"
+	"time"
+	"unsafe"
+
+	"github.com/aquasecurity/tracee/pkg/logger"
+)
+
+const BPF_MAX_ERR_FILE_LEN = 72
```
AlonZivony, line 14: We use CamleCase for consts in golang, even when they have matching value in the BPF code.
You should write it as CamleCase, and write comment that it matches the value in the BPF code.

pkg/ebpf/bpf_error.go
```diff
@@ -0,0 +1,223 @@
...
+
+type BPFErrorType uint32
+
+const (
+	BPF_ERR_UNSPEC BPFErrorType = iota
```
AlonZivony, line 19: Again, use CamleCase and comment that it matches BPF code values.

pkg/ebpf/bpf_error.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	return stringMap[BPF_ERR_UNSPEC]
+}
+
+type BPFError struct {
+	ret  int64
+	id   uint32
+	line uint32
+	file [BPF_MAX_ERR_FILE_LEN]byte
```
AlonZivony, line 63: Why file?

pkg/ebpf/bpf_error.go
```diff
@@ -0,0 +1,223 @@
...
+	ticker := time.NewTicker(time.Second * t.config.Output.AggregateBPFErrorsInterval)
+	for range ticker.C {
+		totalCount := uint32(0)
+		var addErr error
+		iter := t.BPFErrorsCountMap.Iterator()
+		for iter.Next() {
+			keyBytes := iter.Key()
+			bpfErr := BPFError{}
+			if err := bpfErr.Decode(keyBytes); err != nil {
+				t.handleError(fmt.Errorf("dumpErrorCounts - %v", err))
```
AlonZivony, line 141: All the errors in this function have the same error message, so it will be hard to distinguish between them if they happen (at least distinguish from which place the rose from).
You should probably add a short description of what failed.

pkg/ebpf/tracee.go
```diff
@@ -904,6 +924,17 @@ func (t *Tracee) validateKallsymsDependencies() {
...
 func (t *Tracee) populateBPFMaps() error {
+	// Prepare percpu error scratch map as soon as possible to make bpf tracee_error() available
+	bpfErrorsScratchMap, err := t.bpfModule.GetMap("error_scratch")
+	if err != nil {
+		t.Close()
+		return fmt.Errorf("error getting access to 'error_scratch' eBPF Map %v", err)
+	}
+	zeroKey := uint32(0)
+	// cpu (4 bytes) + bpf_error (88 bytes)
+	errorScratchBuffer := make([]byte, goruntime.NumCPU()*4+goruntime.NumCPU()*88)
```
AlonZivony, line 935: Use constant instead of 88

AlonZivony (changes requested):
Few things that I saw that should be changed, mainly semantics.
Amazing work

pkg/ebpf/bpf_error.go
geyslan: In the future we'll have bpf code split in more files.

cmd/tracee-ebpf/flags/output.go
```diff
@@ -30,6 +32,8 @@ option:{stack-addresses,detect-syscall,exec-env,relative-time,exec-hash,parse-ar
   exec-hash                                        when tracing sched_process_exec, show the file hash(sha256) and ctime
   parse-arguments                                  do not show raw machine-readable values for event arguments, instead parse into human readable strings
   parse-arguments-fds                              enable parse-arguments and enrich fd with its file path translation. This can cause pipeline slowdowns.
+  aggregate-bpf-errors                             enable bpf errors aggregation avoiding a noisy output (dump every 2 seconds)
```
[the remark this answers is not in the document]
NDStrahilevitz: I agree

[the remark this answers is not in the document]
geyslan: We discussed picking only aggregation if the insns limit issue was not solved. As it was, I continued with the initial issue requirement (perfbuf).

But sure we can remove perfbuf and make aggregation the only err output - how about a flag to disable dumping?

Let's just note that without perfbuf we cannot have approximate timestamps.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -963,14 +992,62 @@ BPF_HASH(module_init_map, u32, kmod_data_t, 256);                  // holds modu
...
 BPF_PERF_OUTPUT(file_writes, 1024); // file writes events submission
 BPF_PERF_OUTPUT(net_events, 1024);  // network events submission
 
+// HELPERS: ERRORS ---------------------------------------------------------------------------------
+
+// inc_error_count returns count after increment
+static __always_inline u32 inc_error_count(bpf_error_t *err)
+{
+    u32 one = 1;
+    u32 *count = bpf_map_lookup_elem(&errors_count, err);
```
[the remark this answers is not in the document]
geyslan: Alright. So a perfbuf submission (conditioned on a timestamp check) with errors count (aggregation).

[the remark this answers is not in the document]
geyslan: @yanivagman @NDStrahilevitz @AlonZivony @rafaeldtinoco @josedonizetti can we have an amen for that?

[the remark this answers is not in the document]
NDStrahilevitz: I'm good with only aggregation but i'm not sure why we shouldn't use the map only then? As I saw we don't lose any info that way, what would be the pros of the perf buffer here?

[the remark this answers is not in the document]
yanivagman: 1. Having a map requires polling and code in userspace. Here we don't need it. In general, perf buffers are the preferred way for kernel to userspace communication.
2. Using perf buffers we can add context - pid, event name and so on that caused the error.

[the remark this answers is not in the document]
geyslan: @yanivagman as agreed.

```shell
tracee-focal on  2174-err-perfbuffer-2 [$!?] via 🐹 v1.19.3 via ⍱ v2.3.2 
❯ sudo ./dist/tracee-ebpf -t comm=who
{"level":"error","ts":1669900985.2415924,"msg":"Failed to get current command","id":4,"type":"BPF_ERR_GET_CURRENT_COMM","ret":0,"cpu":7,"file":"./pkg/ebpf/c/tracee.bpf.c","line":1772,"count":1}
{"level":"error","ts":1669900985.2417276,"msg":"Failed to init context","id":1,"type":"BPF_ERR_INIT_CONTEXT","ret":-1,"cpu":7,"file":"./pkg/ebpf/c/tracee.bpf.c","line":1839,"count":1}
{"level":"error","ts":1669900985.308129,"msg":"Failed to get current command","id":4,"type":"BPF_ERR_GET_CURRENT_COMM","ret":0,"cpu":1,"file":"./pkg/ebpf/c/tracee.bpf.c","line":1772,"count":1}
...
^C{"level":"error","ts":1669901010.5678182,"msg":"Failed to get current command","id":4,"type":"BPF_ERR_GET_CURRENT_COMM","ret":0,"cpu":0,"file":"./pkg/ebpf/c/tracee.bpf.c","line":1772,"count":11318}
{"level":"error","ts":1669901010.5678468,"msg":"Failed to init context","id":1,"type":"BPF_ERR_INIT_CONTEXT","ret":-1,"cpu":0,"file":"./pkg/ebpf/c/tracee.bpf.c","line":1839,"count":11318}

End of events stream
Stats: {EventCount:0 EventsFiltered:0 NetEvCount:0 BPFErrorsCountPerCPU:[11318 1 55 1 120 505 163 253 21 0 93 16 4 73 21 0] ErrorCount:0 LostEvCount:0 LostWrCount:0 LostNtCount:0 LostBPFErrCount:0}

```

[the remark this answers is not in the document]
NDStrahilevitz: I think we should aggregate the BPFError count in the printer's epilogue, the division per CPU is a bit too verbose IMO,
but maybe the verbose table printer can keep it that way?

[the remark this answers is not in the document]
geyslan: Yeah, I agree with you. I already did that aggregation to prometheus: https://github.com/aquasecurity/tracee/pull/2352/files#diff-187e709c653b0ef28d702e5b1c2481017d62f306a9b97f661006a1a08f27698aR83-R93

Let me see it on printer.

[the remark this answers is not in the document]
geyslan: @NDStrahilevitz take a look at the result:

```shell
^C
End of events stream
Stats: {EventCount:0 EventsFiltered:0 NetEvCount:0 BPFErrorsCount:30 ErrorCount:0 LostEvCount:0 LostWrCount:0 LostNtCount:0 LostBPFErrCount:0}
```

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -963,14 +1000,76 @@ BPF_HASH(module_init_map, u32, kmod_data_t, 256);                  // holds modu
...
+
+    err->id = id;
+    err->ret = ret;
+    err->cpu = bpf_get_smp_processor_id();
+    err->line = line;
+    READ_KERN_STR_INTO(err->file, file);
+
+    bpf_error_count_t *counter_buf = bpf_map_lookup_elem(&error_count_scratch, &zero);
+    if (unlikely(counter_buf == NULL))
+        return;
```
yanivagman, line 1037: I don't see a use for this counter_buf.
You can simply save current timestamp as a local variable.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -963,14 +1000,76 @@ BPF_HASH(module_init_map, u32, kmod_data_t, 256);                  // holds modu
...
+    } else {
+        counter = counter_buf;
+    }
+    bpf_map_update_elem(&errors_count, err, counter, BPF_ANY);
+
+    // submit error when its cpu occurrence time diff is greater than 2s
+    if ((counter_buf->ts_prev - ts_prev) > 2000000000) {
+        bpf_error_output_t *err_output = bpf_map_lookup_elem(&error_output_scratch, &zero);
+        if (unlikely(err_output == NULL))
+            return;
```
yanivagman, line 1058: since bpf_error_output_t embed bpf_error in it, we can avoid having both error_scratch and error_output_scratch maps by having only error_output_scratch and moving it up instead of error_scratch.

pkg/ebpf/bpf_error.go
```diff
@@ -0,0 +1,186 @@
...
+				"type", bpfErr.Type().String(),
+				"ret", bpfErr.Return(),
+				"cpu", bpfErr.CPU(),
+				"file", bpfErr.FileAsString(),
+				"line", bpfErr.Line(),
+				"count", bpfErr.Count(),
+			)
+			increment := bpfErr.Count() - uint32(t.stats.BPFErrorsCount[bpfErr.CPU()])
+			// overflow check can be moved to Counter if required
+			if uint32(math.MaxInt32-t.stats.BPFErrorsCount[bpfErr.CPU()]) < increment {
```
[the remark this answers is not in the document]
geyslan: > 1. Count can be 64bit in bpf code as well, then no overflow should occur
> 2. If doing the above, you can simply take the bpf value of count and override the value of the userspace counter, saving the need to calculate increment

I'm implementing changes to the Counter based on this issue: https://github.com/aquasecurity/tracee/issues/2410

pkg/ebpf/c/tracee.bpf.c
geyslan: Ok. I agree, considering it isn't a macro, so the memory footprint is low. I'll change it.

pkg/ebpf/c/tracee.bpf.c
geyslan: There are remaining logic from the previous implementation. I'll ponder about what your statement.

pkg/ebpf/c/tracee.bpf.c
geyslan: Ah, the scratch is unneeded indeed. But the current timestamp can reside in a stack `counter_buf` as it needs to be used in the first error occurrence (map update).

pkg/ebpf/c/tracee.bpf.c
geyslan: image

As discussed above, the `error_count_scratch` can be removed as we're no longer using macro functions.

But the `error_scratch` is required since it's the key of `errors_count`. For being a key, it only needs to have the unique fields to fully identify the error, which are in this case:

```c
typedef struct bpf_error {
    s64 ret;              // return value
    enum bpf_error_id id; // type
    u32 cpu;
    u32 line;                        // line number
    char file[BPF_MAX_ERR_FILE_LEN]; // filename
} bpf_error_t;
```
On the other hand, `bpf_error_output` is composed that way just to hold the output buffer with more fields (count and a required padding).

```c
typedef struct bpf_error_output {
    u32 count;
    u32 padding;
    struct bpf_error err;
} bpf_error_output_t;
```

How about the two scratches? I'm using them to diminish the stack footprint.

geyslan: Hold on. Had another coffe.

geyslan: @yanivagman you nailed it. Thanks.

image

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -963,14 +1000,61 @@ BPF_HASH(module_init_map, u32, kmod_data_t, 256);                  // holds modu
...
+    } else {
+        counter = &counter_buf;
+    }
+    bpf_map_update_elem(&errors_count, &err_output->err, counter, BPF_ANY);
+
+    // submit error when its cpu occurrence time diff is greater than 2s
+    if ((counter_buf.ts_prev - ts_prev) > 2000000000) {
+        err_output->count = counter->count;
+        bpf_perf_event_output(ctx, &errors, BPF_F_CURRENT_CPU, err_output, sizeof(*err_output));
+    }
```
yanivagman, line 1053:
```suggestion
    err_output->count = 1;
    u64 ts_curr = bpf_ktime_get_ns();
    u64 ts_prev = 0;

    bpf_error_count_t *counter = bpf_map_lookup_elem(&errors_count, &err_output->err);
    if (likely(counter != NULL)) {
        ts_prev = counter->ts_prev; // store previous ts

        counter->count += 1;
        err_output->count = counter->count;
        counter->ts_prev = ts_curr; // set ts_prev as current ts
    } else {
        bpf_error_count_t counter_buf = { .count = 1, .ts_prev = ts_curr };
        bpf_map_update_elem(&errors_count, &err_output->err, counter_buf, BPF_ANY);
    }

    // submit error when its cpu occurrence time diff is greater than 2s
    if ((ts_curr - ts_prev) > 2000000000) {
        bpf_perf_event_output(ctx, &errors, BPF_F_CURRENT_CPU, err_output, sizeof(*err_output));
    }
```

yanivagman: In a second look, your code may be better when considering the stack size.
So I just suggest that you rename to curr_ts which is more clear

yanivagman: And `bpf_map_update_elem(&errors_count, &err_output->err, counter, BPF_ANY);` only needs to been in the else case

geyslan: I'm avoiding initializations like this `bpf_error_count_t counter_buf = { .count = 1, .ts_prev = ts_curr };` since it can cause:

```
invalid indirect read from stack R3 off -16+4 size 16
```

geyslan: > In a second look, your code may be better when considering the stack size.
So I just suggest that you rename to curr_ts which is more clear

Good. I've changed it to just `ts` letting the comments explain its purpose.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -963,14 +1000,61 @@ BPF_HASH(module_init_map, u32, kmod_data_t, 256);                  // holds modu
...
+    counter_buf.count = 1;
+    counter_buf.ts = bpf_ktime_get_ns(); // store the current ts
+    u64 ts_prev = 0;
+
+    bpf_error_count_t *counter = bpf_map_lookup_elem(&errors_count, &err_output->err);
+    if (likely(counter != NULL)) {
+        ts_prev = counter->ts; // store previous ts
+
+        counter->count += 1;
+        counter->ts = counter_buf.ts; // set to current ts
```
geyslan, line 1043: @yanivagman I overlooked this part. Probably it's racy. WDYT about reverting it to my previous `BPF_PERCPU_HASH(errors_count, bpf_error_t, bpf_error_count_t, 4096);`?

yanivagman: Can you please explain?

geyslan: Ignore this. errors_count key has the cpu id so each cpu has its own error count per origin and return value.

pkg/ebpf/bpf_error.go
```diff
@@ -0,0 +1,207 @@
...
+				"count", bpfErr.Count(),
+			)
+			// Create a minor key to keep track of the bpf error origin,
+			// as the BPFError type has count which is volatile
+			orig := getBPFErrorOrigin(bpfErr)
+			if cnt, found := t.stats.BPFErrorsCount[orig]; !found {
+				t.stats.BPFErrorsCount[orig] = counter.NewCounter(uint64(bpfErr.Count()))
+			} else {
+				cnt.Set(uint64(bpfErr.Count()))
+			}
```
geyslan, line 174: @yanivagman I returned to my previous understanding. The perfbuf is aggregated with origin (which already contains cpu) and its count. To correctly set the count, we must track the origin. If we do it percpu only we'll override count when errors arrive from different origins (ret/id/cpu/line/file). So this CounterMap comes as a solution.

geyslan: After further analysis I decided to simplify this by [resetting](https://github.com/aquasecurity/tracee/pull/2352/files#diff-65cb2098ddddc611ef7711cd72ebd7fd5f6bb0e9bfddef3b5600305e8bdf3f80R1083) the count in the ebpf after submission. In the future if we need to keep percpu counts in user space, we can revisit this for improvement.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -993,15 +1030,62 @@ BPF_HASH(module_init_map, u32, kmod_data_t, 256);                  // holds modu
 BPF_LRU_HASH(fd_arg_path_map, fd_arg_task_t, fd_arg_path_t, 1024); // store fds paths by task
 BPF_LRU_HASH(bpf_attach_map, u32, bpf_attach_t, 1024);             // holds bpf prog info
 BPF_LRU_HASH(bpf_attach_tmp_map, u32, bpf_attach_t, 1024);         // temporarily hold bpf_attach_t
-
+BPF_HASH(errors_count, bpf_error_t, bpf_error_count_t, 4096);      // errors count
```
yanivagman, line 1033: shouldn't this be a percpu map? or otherwise there can be race conditions, no?

geyslan: I think that my previous design was using percpu, but take a look in the errors_count entry:

```shell
❯ sudo bpftool map dump name errors_count | head -n13
[{
        "key": {
            "ret": 0,
            "id": "BPF_ERR_MAP_UPDATE_ELEM",
            "cpu": 2,
            "line": 3503,
            "file": "./pkg/ebpf/c/tracee.bpf.c"
        },
        "value": {
            "count": 3,
            "ts": 10421895290386
        }
    },{

```

Its key contains cpu, file and line, which I think make the entry to be accessed uniquely percpu; considering also that bpf doesn't preempt and that this map is not accessed from userspace. WDYT?

pkg/ebpf/bpf_error.go
```diff
@@ -0,0 +1,186 @@
...
+
+			bpfErr := BPFError{}
+			if err := bpfErr.Decode(rawData); err != nil {
+				t.handleError(fmt.Errorf("consumeBPFErrorsChannel: decode - %v", err))
+				continue
+			}
+			// This logger timestamp in no way reflects the ebpf error original time
+			// since bpf errors channel is populated with aggregated logs
+			logger.Error(bpfErr.Error(),
```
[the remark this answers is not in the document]
geyslan: Done.

pkg/logger/logger.go
```diff
@@ -188,6 +188,25 @@ func isAggregateSetAndIsLogNotNew(skip int, l *Logger) bool {
 
 // Log functions
 
+// BPFError logs errors from bpf programs
+// It does not use the aggregation logic as the bpf errors are already aggregated on the bpf side
+func BPFError(msg string, logLevel Level, keysAndValues ...interface{}) {
```
NDStrahilevitz, line 193: I'd keep this function near `processBPFErrors` rather than here, since it's more of a readability helper rather than something a consumer of the logger package would call from anywhere else.

pkg/ebpf/bpf_error.go
```diff
@@ -0,0 +1,194 @@
...
+		return bpfErr
+	}
+
+	return stringMap[BPFErrUnspec]
+}
+
+type BPFErrorLogLevel int32
+
+const (
+	BPFErrLogLvlDebug BPFErrorLogLevel = iota + BPFErrorLogLevel(logger.DebugLevel) // BPF_ERR_LOG_LVL_DEBUG
```
NDStrahilevitz, line 60: This (and the threesubsequent constants) are unused and as such should be removed.

pkg/ebpf/bpf_error.go
```diff
@@ -0,0 +1,194 @@
...
+
+func (b BPFErrorType) String() string {
+	if bpfErr, found := stringMap[b]; found {
+		return bpfErr
+	}
+
+	return stringMap[BPFErrUnspec]
+}
+
+type BPFErrorLogLevel int32
```
NDStrahilevitz, line 57: ditto to the type

pkg/ebpf/bpf_error.go
```diff
@@ -0,0 +1,194 @@
...
+	BPFErrLogLvlDebug BPFErrorLogLevel = iota + BPFErrorLogLevel(logger.DebugLevel) // BPF_ERR_LOG_LVL_DEBUG
+	BPFErrLogLvlInfo                                                                // BPF_ERR_LOG_LVL_INFO
+	BPFErrLogLvlWarn                                                                // BPF_ERR_LOG_LVL_WARN
+	BPFErrLogLvlError                                                               // BPF_ERR_LOG_LVL_ERROR
+)
+
+// BPFError struct contains aggregated data about a bpf error origin
+type BPFError struct {
+	id       BPFErrorType
+	logLevel BPFErrorLogLevel
```
NDStrahilevitz, line 69: you could just use the `logger.LogLevel` type directly

pkg/metrics/stats.go
```diff
@@ -78,6 +80,16 @@ func (stats *Stats) RegisterPrometheus() error {
 		return err
 	}
 
+	err = prometheus.Register(prometheus.NewCounterFunc(prometheus.CounterOpts{
+		Namespace: "tracee_ebpf",
+		Name:      "bpf_errors_total",
+		Help:      "bpf errors collected by tracee-ebpf",
```
NDStrahilevitz, line 86:
```suggestion
		Help:      "errors collected by tracee-ebpf during ebpf execution",
```

pkg/logger/logger.go
geyslan: This is proposed inside the logger package because the default behaviour of this package is to aggregate calls if logger aggregation is on. For the case of the ebpf error coming from the perfbuf it doesn't make sense, so we need a logger helper that skips the inner aggregation. Its naming `logger.BPFError()` can be changed for sure. WDYT?

NDStrahilevitz: As far as I see all this does is switch on the log level and call the appropriate public logger function, with a certain field enrichment on the default case.
No aggregation logic is done in this helper function so I'm not sure what I'm missing.

Does the location of the function affect it's behavior?

pkg/ebpf/bpf_error.go
geyslan: Ok. I got it.

pkg/logger/logger.go
geyslan: > Does the location of the function affect it's behavior?

Yep. Inside the pkg logger we can avoid the inner aggregation. Calling it from outside if the pkg aggregation is on, it will only aggregate.

See that we're bypassing the default behaviour calling it directly: `pkgLogger.l.Debugw(msg, keysAndValues...)`. If we call it from outside we'll face the pkg logger aggregation:

image

Besides that, the default behaviour of the debugw is append the origin from the user space debug call. We don't want that for ebpf errors.

geyslan: Regarding this, if necessary we can change the `logger.BPFError()` to something more generic like `logger.Log(aggregate=false, ...)` or rethink places for that. But for this release, I assume we're ready.

NDStrahilevitz (approved):
LGTM

pkg/logger/logger.go
geyslan: Anything related to this, please let's take to https://github.com/aquasecurity/tracee/issues/2479.

yanivagman (approved):
LGTM!

rafaeldtinoco:
Finally going through this now (and will merge it right after).

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1004,6 +1052,57 @@ BPF_PERF_OUTPUT(net_events, 1024);  // network events submission
...
+    bpf_error_count_t counter_buf = {};
+    counter_buf.count = 1;
+    counter_buf.ts = bpf_ktime_get_ns(); // store the current ts
+    u64 ts_prev = 0;
+
+    bpf_error_count_t *counter = bpf_map_lookup_elem(&errors_count, &err_output->err);
+    if (likely(counter != NULL)) {
+        ts_prev = counter->ts; // store previous ts
+
+        counter->count += 1;
```
rafaeldtinoco, line 1089: you're changing a pointee value on the map without an atomic operation. I believe this is racy among different CPUs (as the "errors_count" map isn't a per CPU map). So, you might need either to copy the counter to the stack, increment and update the key with the new stack values (the key update is atomic) OR use atomic_xchg() for the pointer (not sure if that works for a map pointer, need to check).

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: I left some comments about atomicity in this map operations at https://github.com/aquasecurity/tracee/pull/2352#discussion_r1049162722. Yaniv is right, its either a PERCPU HASH or you have to use atomic operations for incrementing and zeroing.

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: I believe, due to the nature of this counter, there is no problem if we loose a few increments (due to 2 cpus copying the values to the stack at the same time and both updating the key values with their stack values). With that, it would be just a question of memcping the pointer to the stack, incrementing and updating the key.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1004,6 +1052,57 @@ BPF_PERF_OUTPUT(net_events, 1024);  // network events submission
...
+    } else {
+        counter = &counter_buf;
+        bpf_map_update_elem(&errors_count, &err_output->err, counter, BPF_ANY);
+    }
+
+    // submit error when its cpu occurrence time diff is greater than 2s
+    if ((counter->ts - ts_prev) > (u64) 2000000000) {
+        err_output->count = counter->count;
+        bpf_perf_event_output(ctx, &errors, BPF_F_CURRENT_CPU, err_output, sizeof(*err_output));
+        counter->count = 0; // reset, assuming that the consumer is incrementing
```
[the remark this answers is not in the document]
rafaeldtinoco: Same thing as https://github.com/aquasecurity/tracee/pull/2352#discussion_r1049167603.

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: Not doing that might get us into a situation where count is in diff cachelines, and it gets badly corrupted, not that would impact too much, but, still...

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1004,6 +1052,57 @@ BPF_PERF_OUTPUT(net_events, 1024);  // network events submission
...
+        counter->count += 1;
+        counter->ts = counter_buf.ts; // set to current ts
+    } else {
+        counter = &counter_buf;
+        bpf_map_update_elem(&errors_count, &err_output->err, counter, BPF_ANY);
+    }
+
+    // submit error when its cpu occurrence time diff is greater than 2s
+    if ((counter->ts - ts_prev) > (u64) 2000000000) {
+        err_output->count = counter->count;
```
rafaeldtinoco, line 1098: In this case, being a per-CPU map makes this atomic (as no other CPU will be changing the value pointed by the err_output pointer.

rafaeldtinoco:
@geyslan, I left a commit at https://github.com/geyslan/tracee/pull/7 changing very few things (basically formatting AND a suggestion to the map update concurrency issue). I have not tested, etc... so feel free to merge (as you got LGTM) after you check the suggestions and reasoning. Cheers!

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: https://github.com/geyslan/tracee/pull/7

yanivagman: Geyslan made a good point that part of the key to this map is the cpu id.
In that case, it is not possible that two different CPUs will update the same counter

rafaeldtinoco: Indeed it is a good point (haven't payed attention to it). Nice work @geyslan !!

PR #2354, opened 2022-11-12
ebpf: fix error handling

josedonizetti:
```commit 691dd00ed10eed01146a2c4cc5ce3fe051a87f61 (HEAD -> fix-ignored-error, origin/fix-ignored-error)
Date:   Sat Nov 12 14:35:35 2022 -0300

    ebpf: fix error handling
    
    The error returned by kernel.Lockdown() was ignored.
```

ps: in case where we actually want to ignore the error, we should instead do `lockdown, _ := kernel.Lockdown()`

### DIFF
--- a/pkg/cmd/urfave/urfave.go
+++ b/pkg/cmd/urfave/urfave.go
@@ -105,10 +105,15 @@ func GetTraceeRunner(c *cli.Context, version string) (cmd.Runner, error) {
 	// Check kernel lockdown
 
 	lockdown, err := helpers.Lockdown()
+	if err != nil {
+		logger.Error("osinfo", "error", err)
+	}
+
 	if err == nil && lockdown == helpers.CONFIDENTIALITY {
 		return runner, fmt.Errorf("kernel lockdown is set to 'confidentiality', can't load eBPF programs")
 
 	}
+
 	logger.Debug("osinfo", "security_lockdown", lockdown)
 
 	// Check if ftrace is enabled


geyslan (approved):
LGTM, good catch.

cmd/tracee-ebpf/main.go
```diff
@@ -153,7 +153,10 @@ func main() {
 			// Check kernel lockdown
 
 			lockdown, err := helpers.Lockdown()
-			if err == nil && lockdown == helpers.CONFIDENTIALITY {
+			if err != nil {
+				return err
```
yanivagman, line 157: The question is - do we really want to fail tracee if we can't read lockdown status?
I'm always in favor of trying to be opportunistic and continue with the execution, failing later if we really can't continue. We can log the error instead.

yanivagman (changes requested):

cmd/tracee-ebpf/main.go
geyslan: Since Lockdown() is an external helper, I wouldn't be comfortable continuing with execution.

In this specific case, could tracee continue to work without a lockdown mode? If so, when it gives an error, I would check these two returns:

`return NOVALUE, fmt.Errorf("could not get lockdown mode")`

Continuing if the error contains "could not get lockdown mode".

yanivagman: It is possible that the lockdown file doesn't exists (e.g. older kernel) and tracee will be able to continue.

> Continuing if the error contains "could not get lockdown mode".

Yes, that can be an option

rafaeldtinoco: Make sure to return a specific error type you can check (instead of checking if error contains). Example: https://github.com/aquasecurity/tracee/blob/main/pkg/cgroup/cgroup.go#L72

josedonizetti:
@yanivagman @geyslan I've changed the code to only log the error, instead of returning the error and stopping tracee. This way the user knows if something odd happens, and can follow up with an issue.

pkg/cmd/urfave/urfave.go
```diff
@@ -105,10 +105,15 @@ func GetTraceeRunner(c *cli.Context, version string) (cmd.Runner, error) {
 	// Check kernel lockdown
 
 	lockdown, err := helpers.Lockdown()
-	if err == nil && lockdown == helpers.CONFIDENTIALITY {
+	if err != nil {
+		logger.Error("kernel lockdown", "error", err)
+	}
+
+	if lockdown == helpers.CONFIDENTIALITY {
```
yanivagman, line 112: To be on the safe side, better check error is nil

josedonizetti: Oh! That ia true, because we are not returning!

pkg/cmd/urfave/urfave.go
```diff
@@ -105,10 +105,15 @@ func GetTraceeRunner(c *cli.Context, version string) (cmd.Runner, error) {
 	// Check kernel lockdown
 
 	lockdown, err := helpers.Lockdown()
+	if err != nil {
+		logger.Error("kernel lockdown", "error", err)
```
geyslan, line 109: Perhaps changing 117 debug msg to the same pattern?

Issue #2360, opened 2022-11-16
Capture write's`written_files` file is not updating


AlonZivony:
## Prerequisites

- [x] This affects latest released version.
- [x] This affects current development tree (origin/HEAD).
- [ ] There isn't an issue describing the bug.

Select one OR another:

- [x] I'm going to create a PR to solve this (assign to yourself).
- [ ] Someone else should solve this.

## Bug description

When running Tracee with capture of written files, a file named `written_files` should be created.
This file maps the files device and inode to the file path.
However, the creation of this file depends on the submission of the `vfs_write`, `vfs_writev` and `kernel_write` events.
Since we divided the probes and the events dependency, the capture write feature does not require the submission of the events, which means the `written_files` files does not update.

## Steps to reproduce

Steps to reproduce the issue:

Run `./tracee-ebpf -c write -t e=init_namespaces` should ensure that the file is not updated.

yanivagman:
A possible solution here might be adding a bpf map (possibly LRU) to map between device id + inode to file path, then reading the id+inode in userspace (in case it is not already populated) whenever a capture event is received.

PR #2361, opened 2022-11-16
pkg/events: Add event dependencies to capture write

AlonZivony:
## Initial Checklist

## Description (git log)

Capture write need the submission of the write events to manage updating the 'written_files' file in the out directory. To submit these events, they are now added as event dependencies.

Fixes: #2360 

## Type of change

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/events/events.go
+++ b/pkg/events/events.go
@@ -5815,6 +5815,11 @@ var Definitions = eventDefinitions{
 					{MapName: "prog_array", MapIndexes: []uint32{tailKernelWrite}, ProgName: "trace_ret_kernel_write_tail"},
 					{MapName: "prog_array", MapIndexes: []uint32{tailSendBin}, ProgName: "send_bin"},
 				},
+				Events: []eventDependency{
+					{VfsWrite},
+					{VfsWritev},
+					{KernelWrite},
+				},
 			},
 		},
 		CaptureExec: {


pkg/events/events.go
```diff
@@ -5815,6 +5815,11 @@ var Definitions = eventDefinitions{
 					{MapName: "prog_array", MapIndexes: []uint32{tailKernelWrite}, ProgName: "trace_ret_kernel_write_tail"},
 					{MapName: "prog_array", MapIndexes: []uint32{tailSendBin}, ProgName: "send_bin"},
 				},
+				Events: []eventDependency{
+					{VfsWrite},
+					{VfsWritev},
```
yanivagman, line 5820: 1. There are environments (GKE if I remember correctly) where this event will fail to attach since the vfs_writev symbol is not exposed through kallsyms - won't this make capture unavailable on such environments?
2. vfs_write and vfs_writev can be very verbose, and sending them through the events perf buffer may cause lost events. I'm not sure this is the best way to go (maybe we should consider sending the path when we capture in the bpf code instead of adding these dependencies)

AlonZivony: > 1. There are environments (GKE if I remember correctly) where this event will fail to attach since the vfs_writev symbol is not exposed through kallsyms - won't this make capture unavailable on such environments?

You are right that this is a problem, but it is somewhat impossible to really use symbols that are available for all the kernels (`vfs_writev` is a very basic symbol, so I think we should be able to expect it to be available most of the time). The only solution is to tell somehow that it is okay for the dependency to fail (maybe add require flag?)

> 2. vfs_write and vfs_writev can be very verbose, and sending them through the events perf buffer may cause lost events. I'm not sure this is the best way to go (maybe we should consider sending the path when we capture in the bpf code instead of adding these dependencies)

I agree with you here, and I think this solution is the best. But are we okay with letting a bug with one of our features to stay?

yanivagman:
As explained in the comment above - adding these dependencies is unacceptable (mostly due to performance hit introduced by submitting vfs_write(v) events to userspace).
Need to find another solution to fix #2360 or rethink written_files feature.

Issue #2362, opened 2022-11-17
[FEAT] add webhook support to tracee-ebpf


josedonizetti:
Add webhook support to `tracee-ebpf` as part of the everything is an event task.

yanivagman:
Part of #2355

PR #2363, opened 2022-11-17
[DO NOT MERGE] ebpf: add webhook support

josedonizetti:
## Initial Checklist

This PR adds webhook support to `tracee-ebpf`, which is part of the task to have everything as an event, by merging `tracee-ebpf` and `tracee-rules`. The `--webhook` flag allows multiple values (currently limited) so we can extend it further later, also it is not a printer, because we should be able to print/send webhooks at the same time, plus webhook should evolve to much more complex than a printer later, eg:

- other formats (template? gob?)
- secret header value
- tls
- mutual tls (spiffe integration, etc)
- circuit break
- event filtering

## Description (git log)

```
commit 28ad426144834803a0093dcd67c0b57a34ca4d14 (HEAD -> add-webhook-tracee-ebpf, origin/add-webhook-tracee-ebpf)
Date:   Wed Nov 16 21:54:52 2022 -0300

    ebpf: add webhook support
```

Copy and paste the git log here.

Fixes: https://github.com/aquasecurity/tracee/issues/2362

## Type of change

## How Has This Been Tested?

```
sudo tracee-ebpf --webhook url=http://localhost:8080 --webhook timeout=5s --webhook format=json --trace event=execve
```

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/cmd/tracee-ebpf/main.go
+++ b/cmd/tracee-ebpf/main.go
@@ -150,6 +150,11 @@ func main() {
 			printerConfig.ContainerMode = containerMode
 			cfg.Output = &output
 
+			webhook, err := flags.PrepareWebhook(c.StringSlice("webhook"))
+			if err != nil {
+				return err
+			}
+
 			// Check kernel lockdown
 
 			lockdown, err := helpers.Lockdown()
@@ -247,6 +252,9 @@ func main() {
 					select {
 					case event := <-cfg.ChanEvents:
 						printer.Print(event)
+						if err := webhook.Send(event); err != nil {
+							printer.Error(err)
+						}
 					case err := <-cfg.ChanErrors:
 						printer.Error(err)
 					case <-ctx.Done():
@@ -360,6 +368,11 @@ func main() {
 				Name:        "containers",
 				Usage:       "enable container info enrichment to events. this feature is experimental and may cause unexpected behavior in the pipeline",
 				Destination: &enrich,
+			},
+			&cli.StringSliceFlag{
+				Name:  "webhook",
+				Value: nil,
+				Usage: "HTTP endpoint to call for every event. Run '--webhook help' for more info.",
 			},
 		},
 	}
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -1,0 +1,56 @@
+package webhook
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/aquasecurity/tracee/pkg/logger"
+	"github.com/aquasecurity/tracee/types/trace"
+)
+
+type Webhook struct {
+	URL     string
+	Format  string
+	Timeout time.Duration
+}
+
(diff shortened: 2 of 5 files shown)


pkg/webhook/webhook.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/aquasecurity/tracee/types/trace"
+)
+
+type Webhook struct {
+	URL     string
+	Format  string
+	Timeout time.Duration
+}
+
+func (w *Webhook) Send(event trace.Event) error {
```
geyslan, line 20: Will this method make changes in the receiver in the future? If not, I suppose it can be a value receiver.

pkg/webhook/webhook.go
```diff
@@ -0,0 +1,56 @@
...
+	if resp.StatusCode != http.StatusOK {
+		logger.Error(fmt.Sprintf("error sending webhook, http status: %d", resp.StatusCode))
+	}
+
+	_ = resp.Body.Close()
+
+	return nil
+}
+
+func (w *Webhook) getContentType() string {
```
geyslan, line 54: ditto

geyslan (approved):
Tested with `for ((i=0;i<5;i++)); do echo -e "\n\nnc INSTANCE $i\n"; nc -l -p 8080; done` and worked.

cmd/tracee-ebpf/main.go
```diff
@@ -247,6 +252,9 @@ func main() {
 					select {
 					case event := <-cfg.ChanEvents:
 						printer.Print(event)
+						if err := webhook.Send(event); err != nil {
+							printer.Error(err)
```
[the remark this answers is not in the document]
josedonizetti: :+1: I'll add a task to look into this, once we mege this PR. k?

josedonizetti:
I've added a DO NOT MERGE to the PR title after speaking with @NDStrahilevitz. This PR is dangerous to the user because it does not filter which events go on the webhook. Tracee collects lots and lots of data, and the current data doesn't make sense to go down a webhook, webhooks intentions are not to take data out, for this other strategies should be used, GRPC streaming sever, fluentd, etc

This PR is part of the task `everything is an event`, which would bring signatures are events, and those events make sense to be sent down the webhook because they are supposed to generate alerts. Let's wait until others PRs for the everything is an event task is done, before merging this one.

cmd/tracee-ebpf/main.go
```diff
@@ -247,6 +252,9 @@ func main() {
 					select {
 					case event := <-cfg.ChanEvents:
 						printer.Print(event)
+						if err := webhook.Send(event); err != nil {
```
yanivagman, line 255: Adding this here means that for every event we will create and send an http packet, is that right?
If so, it has very bad implications on performance which we should avoid.
Maybe as a first step we should simply make the webhook another printer? WDYT?

NDStrahilevitz:
@josedonizetti can we move this to draft to prevent accidental merging until you have the other relevant infrastructure merged?

PR #2364, opened 2022-11-17
Change tracee objects to support multiple modules

grantseltzer:
## Initial Checklist

## Description (git log)

commit 877d5bfca46906d89b1477f4e92a131b938042e8 (HEAD -> modular-bpf-objects, pfork/modular-bpf-objects)
Date:   Thu Nov 17 12:11:32 2022 -0500

    Change BPF object parsing logic to support multiple objects
    
    This changes the fields in Tracee's config and central
    data structure so that there are multiple possible BPF
    compiled BPF objects with their own paths and contents.
    
    This PR should not change any logic or behavior, but
    instead set tracee up for modular BPF objects where
    current 'main' BPF object is replaced by a series of
    modules such as for individual programs and maps.
    
## Type of change

## How Has This Been Tested?

- [ ] make test
- [ ] make test-integration

### DIFF
--- a/pkg/ebpf/tracee.go
+++ b/pkg/ebpf/tracee.go
@@ -62,8 +62,7 @@ type Config struct {
 	Debug              bool
 	maxPidsCache       int // maximum number of pids to cache per mnt ns (in Tracee.pidsInMntns)
 	BTFObjPath         string
-	BPFObjPath         string
-	BPFObjBytes        []byte
+	BPFObjConfigs      map[TraceeBPFModuleID]BPFObjConfig
 	KernelConfig       *helpers.KernelConfig
 	ChanEvents         chan trace.Event
 	ChanErrors         chan error
@@ -72,6 +71,17 @@ type Config struct {
 	Sockets            runtime.Sockets
 	ContainersEnrich   bool
 }
+
+type BPFObjConfig struct {
+	BPFObjPath  string
+	BPFObjBytes []byte
+}
+
+type TraceeBPFModuleID uint32
+
+const (
+	BPFModuleMain TraceeBPFModuleID = iota + 1
+)
 
 type CaptureConfig struct {
 	OutputPath      string
@@ -140,8 +150,14 @@ func (tc Config) Validate() error {
 		}
 	}
 
-	if tc.BPFObjBytes == nil {
-		return errors.New("nil bpf object in memory")
+	if len(tc.BPFObjConfigs) == 0 {
+		return errors.New("no bpf objects configured")
+	}
+
+	for i := range tc.BPFObjConfigs {
+		if tc.BPFObjConfigs[i].BPFObjBytes == nil {
+			return errors.New("nil bpf object in memory")
+		}
 	}
 
 	if tc.ChanEvents == nil {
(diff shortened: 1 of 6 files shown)


grantseltzer:
#2365 is the follow up to this

pkg/ebpf/tracee.go
```diff
@@ -1252,11 +1290,17 @@ func (t *Tracee) writeProfilerStats(wr io.Writer) error {
 }
 
 func (t *Tracee) getProcessCtx(hostTid uint32) (procinfo.ProcessCtx, error) {
+
+	//FIXME: is this access going to affect performance?
+	mainBpfModule, ok := t.bpfModules[BPFModuleMain]
```
yanivagman, line 1295: nit: in some places you call this bpfMainModule and in some others mainBpfModule - let's try to stick with only one of them

pkg/ebpf/tracee.go
```diff
@@ -71,6 +70,17 @@ type Config struct {
 	ContainersEnrich   bool
 }
 
+type BPFObjConfig struct {
```
NDStrahilevitz, line 73: I propose we move these 3 into a `bpfplugin` package or something of the sorts because it's really a start of that.
Later we can have plugins using it to register themselves etc. but for now it would be a good place for the types.

pkg/ebpf/tracee.go
```diff
@@ -1252,11 +1290,17 @@ func (t *Tracee) writeProfilerStats(wr io.Writer) error {
 }
 
 func (t *Tracee) getProcessCtx(hostTid uint32) (procinfo.ProcessCtx, error) {
+
+	//FIXME: is this access going to affect performance?
```
NDStrahilevitz, line 1294: This function is invoked as a hot path only in the network events which are getting replaced.
So I think this path is getting deleted anyway in the near future (@yanivagman @rafaeldtinoco  correct me if i'm wrong).

grantseltzer: @yanivagman I see the main bpf module as a thing we'll eventually deprecate as all logic will eventually be separated into different modules.

pkg/ebpf/tracee.go
grantseltzer: Does it make sense to have this as part of libbpfgo instead?

grantseltzer: I think having logic for plugins in their own package makes sense. However since we don't know what the interface looks like I don't want to split it up too early.

pkg/ebpf/tracee.go
grantseltzer: Have what I think is a better approach started here: https://github.com/grantseltzer/tracee/tree/modular-tracee-bpf-register

i'll open that in a PR tomorrow.

grantseltzer: @yanivagman What do you think of the current state of this PR? The next step after this is moving maps from the main module to a separate map one.

rafaeldtinoco:
@yanivagman this is outdated (needs rebasing) and won't be merged soon. I'm closing this but will keep as a reference for future related work. Hope you agree.

Issue #2367, opened 2022-11-20
[BUG] Data segment symbols missing on some kernels


AlonZivony:
## Prerequisites

- [x] This affects latest released version.
- [x] This affects current development tree (origin/HEAD).
- [x] There isn't an issue describing the bug.

Select one OR another:

- [x] I'm going to create a PR to solve this (assign to yourself).
- [ ] Someone else should solve this.

## Bug description

When compiling the kernel, you can configure which symbols will be included in the `kallsyms`, if any will be included at all.
The `CONFIG_KALLSYMS` determines if to have `kallsyms`, which includes the `text` and `inittext` segments symbols.
The `CONFIG_KALLSYMS_ALL` determines if to include all symbols in the kernel in the `kallsyms`.
It means that kernels without `CONFIG_KALLSYMS_ALL` won't support using of `data` and `rodata` symbols.

Currently we use the `pipefifo_fops` struct symbol, which should be in the `rodata` segment.
This means that in kernels with partial symbols, the `get_file_pipe_info` will not work properly.

PR #2370, opened 2022-11-21
pkg/ebpf: cancel event with missing symbols dependency

AlonZivony:
## Initial Checklist

## Description (git log)

When an event has a missing symbols dependency which it defines as crucial for its operation, it will be canceled with matching error message.

Fixes: #2367

## Type of change

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/ebpf/tracee.go
+++ b/pkg/ebpf/tracee.go
@@ -401,6 +401,9 @@ func (t *Tracee) Init() error {
 		return err
 	}
 
+	// Canceling events missing kernel symbols
+	t.validateKallsymsDependencies()
+
 	// Initialize process tree before receiving events (will use it)
 
 	t.procInfo, err = procinfo.NewProcessInfo()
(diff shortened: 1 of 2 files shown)


pkg/ebpf/tracee.go
```diff
@@ -827,6 +830,54 @@ func (t *Tracee) getFiltersConfig() uint32 {
...
+// validateKallsymsDependencies load all symbols required by events' dependencies from the kallsyms file to check for missing symbols.
+// If some symbols are missing, it will cancel their event with informative error message.
+func (t *Tracee) validateKallsymsDependencies() {
+	var reqKsyms []string
+	symsToDependentEvents := make(map[string][]events.ID)
+	for id := range t.events {
+		event := events.Definitions.Get(id)
+		for _, symDep := range event.Dependencies.KSymbols {
+			reqKsyms = append(reqKsyms, symDep.Symbol)
+			if symDep.Required {
```
rafaeldtinoco, line 842: I wonder when symbols wouldn't be required (looks like kSymboldependency has required always set to true in all symbols declared).

rafaeldtinoco (approved):
LGTM. Couldn't find anything wrong with the logic.

pkg/ebpf/tracee.go
AlonZivony: So I have a place I want to do it in already, this is why I added the required flag.
I thought that we can try using symbols that are not guaranteed to exist like symbols in the `data` or `rodata` segments, and have a fail-safe use of `text` segment symbols in the case they don't exist.
For example, with the `pipefifo_fops` we can try loading it in the eBPF program, and if it doesn't exist we can try matching some of the file operations pointers to expected functions symbols to verify it. You can see that the best use-case is with the struct, but it will just mean more complicated check if it is missing.

PR #2374, opened 2022-11-23
tracee: add cmd/tracee

josedonizetti:
## Initial Checklist

```
commit 841cb75c8062ad25bb977f6e13c53c173d79c09a (HEAD -> add-tracee-cmd, origin/add-tracee-cmd)
Date:   Sat Nov 12 14:34:02 2022 -0300

    tracee: add cmd/tracee
    
    The new tracee binary combines both tracee-ebpf, and tracee-rules to a
    new UX of tracee. Where every thing is treated as an event.
    
    The rule engine becomes part of the tracee events pipeline.
```

Fixes: 

## How Has This Been Tested?

```
# compiles
make

# test single rule event
./dist/tracee --trace event=anti_debugging_ptraceme 

# test event with output json
./dist/tracee --trace event=anti_debugging_ptraceme --output format:json

# test single event with extra non rule event
./dist/tracee --trace event=anti_debugging_ptraceme,execve

# test single event with non rule dependency event
./dist/tracee --trace event=ptrace,anti_debugging_ptraceme

# test all events
./dist/tracee  --trace event=anti_debugging_ptraceme,aslr_inspection,cgroup_notify_on_release_modification,cgroup_release_agent_modification,core_pattern_modification,default_loader_modification,disk_mount,docker_abuse,dynamic_code_loading,fileless_execution,hidden_file_created,illegitimate_shell,kernel_module_loading,kubernetes_api_connection,kubernetes_certificate_theft_attempt,ld_preload,process_vm_write_code_injection,proc_fops_hooking,proc_kcore_read,proc_mem_access,proc_mem_code_injection,ptrace_code_injection,rcd_modification,sched_debug_recon,scheduled_task_modification,stdio_over_socket,sudoers_modification,syscall_table_hooking,system_request_key_config_modification

# test rule event non exported 
./dist/tracee --output format:json --trace event=dropped_executable
```

Notes:

- This is planned for the current release, though due to the whole concept change, we will not make it default yet, users will be able to use the new experience, by customizing their installation. So, it is important to consider we are keeping backward compatibility for now. So not every refactor was done here, but the idea is to do it for the next release until we only have the single binary and new experience. 

- The PR adds the rule engine as part of the tracee events pipeline, where all rules (signatures) become an event. The events definitions table is changed by the signatures, new events are added using the signatures name, metadata, event selectors and a new filed event name. 

follow up prs:
- adding to docker, kubernetes will be done once this PR is merged
- the rego rules were not changed yet
- move packages inside `cmd/tracee-ebpg/` to `pkg/cmd`
- documentation, so it can become the default option


go.mod
```diff
@@ -103,3 +103,5 @@ require (
 )
 
 replace github.com/kubernetes/cri-api => k8s.io/cri-api v0.23.5-rc.0
+
+replace github.com/aquasecurity/tracee/types => ./types
```
josedonizetti, line 107: Once everyone agrees with the implementation I will create a separate PR in order to fix the go.mod here.

pkg/ebpf/finding.go
```diff
@@ -0,0 +1,60 @@
...
+func FindingToEvent(f detect.Finding) (*trace.Event, error) {
+	s, ok := f.Event.Payload.(trace.Event)
+
+	if !ok {
+		return nil, fmt.Errorf("error converting finding to event: %s", f.SigMetadata.ID)
+	}
+
+	eventID, found := events.Definitions.GetID(f.SigMetadata.EventName)
+	if !found {
+		return nil, fmt.Errorf("error event not found: %s", f.SigMetadata.EventName)
```
NDStrahilevitz, line 22:
```suggestion
		return nil, fmt.Errorf("error finding event not found: %s", f.SigMetadata.EventName)
```

pkg/rules/engine/engine.go
```diff
@@ -67,6 +72,25 @@ func NewEngine(sigs []detect.Signature, sources EventSources, output chan detect
 	return &engine, nil
 }
 
+// StartPipeline receives an input chanel, and returns an output channel
+// allowing the rule engine to be used in the events pipeline
+func StartPipeline(cfg Config, input chan protocol.Event, done chan bool) <-chan detect.Finding {
+	output := make(chan detect.Finding)
+
+	source := EventSources{Tracee: input}
+	engine, err := NewEngine(cfg.Signatures, source, output, cfg)
```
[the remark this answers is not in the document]
josedonizetti: Yup, it is definitely weird, and not what we want going forward, but for this PR it was chosen because I didn't want to change even more things on this PR, and we will want to support the old experience for now, this will change in the upcoming weeks, allowing us to consolidate in one solution.

types/detect/detect.go
```diff
@@ -24,6 +24,7 @@ type SignatureMetadata struct {
 	ID          string
 	Version     string
 	Name        string
+	EventName   string `json:"event-name"`
```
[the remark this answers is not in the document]
josedonizetti: yeah, I used the dash because of the rego signatures metadata, but agree need to stick with the convention.

pkg/ebpf/events_engine.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/aquasecurity/tracee/types/protocol"
+	"github.com/aquasecurity/tracee/types/trace"
+)
+
+// engineEvents stage in the pipeline allows rules detection to be executed in the pipeline
+func (t *Tracee) engineEvents(ctx context.Context, in <-chan *trace.Event) (<-chan *trace.Event, <-chan error) {
+	out := make(chan *trace.Event)
+	errc := make(chan error, 1)
+	done := make(chan bool)
+	engineInput := make(chan protocol.Event)
```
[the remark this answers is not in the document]
josedonizetti: @NDStrahilevitz That is what I was planning, but wanted more feedback, specially because we want a soft transition or to have an adapter not to break aqua applications using tracee.

cmd/tracee/main.go
```diff
@@ -0,0 +1,258 @@
...
+
+			traceeConfig.EngineConfig = engine.Config{
+				Enabled:             true,
+				Signatures:          sigs,
+				SignatureBufferSize: c.Uint("sig-buffer"),
+			}
+
+			return cmd.Run(ctx, c, traceeConfig, printerConfig)
+		},
+		Flags: []cli.Flag{
```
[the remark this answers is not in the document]
josedonizetti: Also considered this refactor, and opted not to change now, because if we decide to drop the previous experience to have only `cmd/tracee` this won't be necessary either way, the PR was getting to big and this refactor along a couple of others for the `cmd/tracee-ebpg/*` was purposely not done.

pkg/events/events.go
```diff
@@ -69,10 +69,36 @@ type Event struct {
...
+
+	return evt
+}
+
 type eventDefinitions struct {
 	events map[ID]Event
 }
 
+// Add add event to defitions
+func (e *eventDefinitions) Add(eventId ID, evt Event) {
```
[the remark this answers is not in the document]
josedonizetti: @NDStrahilevitz 100%, didn't think about the option of overriding an event, will fix it!

pkg/events/events.go
```diff
@@ -69,10 +69,36 @@ type Event struct {
 	Params       []trace.ArgMeta
 }
 
+// NewEvent creates a new event
+func NewEvent(name string, sets []string, depsID []ID) Event {
```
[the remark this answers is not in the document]
josedonizetti: Also, I like the idea of having `event_definitions.go` and migrating the logic. But it would be for a next PR.

pkg/events/events.go
```diff
@@ -205,6 +231,12 @@ const (
 	CapturePcap
 )
 
+// Rules events
+const (
+	StartRulesID ID = 6000
```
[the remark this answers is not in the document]
josedonizetti: hmm, I thought that considering we want a specific boundary for the id of the rule, enforcing a range, it would be best to have constants here, and perhaps validate them. But yeah, this code seems lost here.

cmd/tracee/main.go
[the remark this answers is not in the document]
geyslan: I think we shouldn't mess with the old tracee-ebpf too much in the first PRs. New things must reside only here. This will facilitate parallel PRs from other tasks.

cmd/tracee/main.go
```diff
@@ -0,0 +1,258 @@
...
+
+		events.Definitions.Add(events.ID(id), event)
+		id++
+	}
+
+}
+
+func getEventNames(c *cli.Context) []string {
+	// TODO: --trace event= will become --events
+	traceArgs := c.StringSlice("trace")
```
geyslan, line 249: As this is a new binary (and for the transition we're keeping the old), how about making it `--events` already?

pkg/rules/signature/signature.go
```diff
@@ -49,7 +49,8 @@ func getSignatures(target string, partialEval bool, rulesDir string, rules []str
 	} else {
 		for _, s := range sigs {
 			for _, r := range rules {
-				if m, err := s.GetMetadata(); err == nil && m.ID == r {
+				if m, err := s.GetMetadata(); err == nil &&
+					(m.ID == r || m.EventName == r) {
```
geyslan, line 53: Isn't ID check sufficient? Have ID and EventName the same value?

josedonizetti: hmm, how would we do it? The ids are `TRC-1` for example, events are more meaningful, like `anti_debugging_ptraceme`, making eventName `TRC-1` have no meaning, and changing ID for eventName will break the compatibility of the signatures.

cmd/tracee/main.go
josedonizetti: Maybe, but can be a separate PR pre-release, instead of changing this one? I feel like this PR is way too big already.

geyslan: Makes sense. 👍🏼

pkg/rules/signature/signature.go
geyslan: I got it. Agreeing with your implementation.

NDStrahilevitz: I don't find it very critical myself, but my two cents on the issue is that i'm not sure if we should allow selecting signatures through their event names in the same context, but again, not too attached to it either way.

pkg/rules/engine/engine.go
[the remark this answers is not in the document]
NDStrahilevitz: Not saying we shouldn't support the old implementation, just saying we should refactor all the argument to be passed in one struct. The `tracee-rules` binary shouldn't be broken for this, just some code needs to move around. I think this is the kind of moving around which fits to the scope of this PR.

pkg/events/events.go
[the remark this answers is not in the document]
NDStrahilevitz: I don't know if it's reasonable to reach 1000 rules, and maybe it's even wise to set this hard limit, but if we do it should at least be documented.
I think we can keep the consts here as long as we validate for them.

pkg/ebpf/events_engine.go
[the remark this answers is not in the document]
NDStrahilevitz: I'd argue a major version is the time, but we would need to communicate this to other teams and products. Either way, it probably deserves it's own PR, just though it's a good point to raise the issue.

Makefile
```diff
@@ -229,6 +229,7 @@ help:
 	@echo "    $$ make tracee-rules         	# build ./dist/tracee-rules"
 	@echo "    $$ make tracee-bench         	# build ./dist/tracee-bench"
 	@echo "    $$ make rules                	# build ./dist/rules"
+	@echo "    $$ make tracee                 # build ./dist/tracee"
```
rafaeldtinoco, line 232: nit: unaligned

Makefile
```diff
@@ -244,6 +245,7 @@ help:
 	@echo "    $$ make clean-tracee-rules   	# wipe ./dist/tracee-rules"
 	@echo "    $$ make clean-tracee-bench   	# wipe ./dist/tracee-bench"
 	@echo "    $$ make clean-rules          	# wipe ./dist/rules"
+	@echo "    $$ make clean-tracee 					# wipe ./dist/tracee"
```
rafaeldtinoco, line 248: nit: unaligned

signatures/golang/anti_debugging_ptraceme.go
```diff
@@ -25,6 +25,7 @@ func (sig *AntiDebuggingPtraceme) GetMetadata() (detect.SignatureMetadata, error
 		ID:          "TRC-102",
 		Version:     "1",
 		Name:        "Anti-Debugging detected",
+		EventName:   "anti_debugging_ptraceme",
```
rafaeldtinoco, line 28: all these simple EventName changes should be on its own commit (to clean this up a bit).

josedonizetti: @rafaeldtinoco Yes, just want to be sure everyone agreed with the new UX, as a whole PR before breaking it on refactoring without a context. The update of types+event_name can be separated on a different PR, will do it in now.

cmd/tracee/main.go
```diff
@@ -0,0 +1,258 @@
...
+			// parse arguments must be enabled if the rule engine is part of the pipeline
+			traceeConfig.Output.ParseArguments = true
+
+			traceeConfig.EngineConfig = engine.Config{
+				Enabled:             true,
+				Signatures:          sigs,
+				SignatureBufferSize: c.Uint("sig-buffer"),
+			}
+
+			return cmd.Run(ctx, c, traceeConfig, printerConfig)
```
rafaeldtinoco, line 95: So you're keeping it like tracee-ebpf for now, and then change it during next PRs ?

josedonizetti: hmm, didn't understand the question

rafaeldtinoco: nm, you already clarified on teams! thanks.

pkg/ebpf/events_engine.go
```diff
@@ -0,0 +1,81 @@
...
+		defer close(out)
+		defer close(errc)
+		defer close(engineInput)
+
+		for {
+			select {
+			case event := <-in:
+				id := events.ID(event.EventID)
+
+				// if the even is marked as submit is pass it to the engine
```
rafaeldtinoco, line 35: this comment is wrongly written

pkg/ebpf/events_engine.go
```diff
@@ -0,0 +1,81 @@
...
+func (t *Tracee) engineEvents(ctx context.Context, in <-chan *trace.Event) (<-chan *trace.Event, <-chan error) {
+	out := make(chan *trace.Event)
+	errc := make(chan error, 1)
+	done := make(chan bool)
+	engineInput := make(chan protocol.Event)
+
+	engineOutput := engine.StartPipeline(t.config.EngineConfig, engineInput, done)
+
+	// TODO: in the upcoming releases, the rule engine should be changed to receive trace.Event,
+	// and return an trace.Event, which should remove the necessity of converting trace.Event to protocol.Event,
```
rafaeldtinoco, line 22: nit: "a trace.Event"

pkg/events/events.go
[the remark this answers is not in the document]
rafaeldtinoco: Yep, agreed.

pkg/events/events.go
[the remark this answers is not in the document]
rafaeldtinoco: +1 on keeping it here, together with other limits.

pkg/rules/engine/engine.go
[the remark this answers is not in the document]
rafaeldtinoco: Agreed. Instead of cfg.Signatures and cfg, right ? Its an easy change.

pkg/ebpf/events_engine.go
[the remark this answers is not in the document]
rafaeldtinoco: > 2\. Remove the `protocol.Event` entirely and just use `trace.Event` directly in signatures.

That would change all signatures in a simple way (and remove the need for checking `event.Payload.(trace.Event)` all the time, so agreed (but for another milestone/PR).

rafaeldtinoco (commented):
It looks good to me as a 1st step towards what we want (with many things to be changed in subsequent PRs, but good enough for this begin). I'll give another look after you address the current comments (and split the changes into multiple commits).

josedonizetti:
I'm currently breaking this PR down into several others, to simplify the reviewing process, and be able to discuss things deeper. 

DONE: 

1 - add EventName to types
2 - Add event name to signatures
3 - Extract pkg/rules/signature
4 - Extract pkg/cmd

TODO:

5 - Add cmd/tracee
6 - Add rules to events_pipeline

PR #2399, opened 2022-11-28
[MAINT] workflows/pr: add kinetic60 and focal419

rafaeldtinoco:
- Add an Ubuntu with upstream v6.0 kernel
- Add an Ubuntu Focal with upstream v4.19 kernel (non-core)

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -119,7 +119,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


geyslan (approved):
Even with 4.19 not doing smoothly.

rafaeldtinoco:
As suspected, we're broken in v4.19 (likely due to libbpf 1.0+ and the functions being wrapped to maintain compatibility with older libbpfs). Still, this PR is good (and I'll open a bug for v4.19 non CO-RE support fix).

PR #2403, opened 2022-11-29
Streamline error logging

NDStrahilevitz:
## Initial Checklist

## Description (git log)
commit 4fcc4735089633ae623a3115aa0f9b3bcb6026b1
```
    tracee-ebpf: enable log redirection
    
    Change the err-file flag to log-file, which handles redirecting
    tracee-ebpf's log output.
```
commit 94efde27687d7eadc7b3b0653feb5e368557f85d
```
    errors: remove error channel and printer
    
    tracee.handleError redirects errors to the logger instead of an error
    channel.
    Additionally, the printer's error method is also removed, using the
    logger for printing errors too.
```

Fixes: #2404

## Type of change

## How Has This Been Tested?

tracee-ebpf -o log-file:test

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/docs/docs/tracing/output-formats.md
+++ b/docs/docs/tracing/output-formats.md
@@ -92,8 +92,8 @@ output and standard error:
 
 2. Error file
 
-    Redirect errors to your log files if needed:
+    Redirect logs to a file if needed:
 
     ```text
-    $ sudo TRACEE_BPF_FILE=do-not-exist ./dist/tracee-ebpf --output json --trace comm=bash --trace follow --trace event=openat --output out-file:/tmp/tracee.log --output err-file:/tmp/tracee.err
+    $ sudo TRACEE_BPF_FILE=do-not-exist ./dist/tracee-ebpf --output json --trace comm=bash --trace follow --trace event=openat --output out-file:/tmp/tracee.events --output log-file:/tmp/tracee.log
     ```
--- a/pkg/cmd/flags/output.go
+++ b/pkg/cmd/flags/output.go
@@ -19,7 +19,7 @@ Possible options:
 [format:]gob                                       output events in gob format
 [format:]gotemplate=/path/to/template              output events formatted using a given gotemplate file
 out-file:/path/to/file                             write the output to a specified file. create/trim the file if exists (default: stdout)
-err-file:/path/to/file                             write the errors to a specified file. create/trim the file if exists (default: stderr)
+log-file:/path/to/file                             write the logs to a specified file. create/trim the file if exists (default: stderr)
 none                                               ignore stream of events output, usually used with --capture
 option:{stack-addresses,detect-syscall,exec-env,relative-time,exec-hash,parse-arguments,sort-events}
                                                    augment output according to given options (default: none)
@@ -35,14 +35,20 @@ option:{stack-addresses,detect-syscall,exec-env,relative-time,exec-hash,parse-ar
 Examples:
   --output json                                            | output as json
   --output gotemplate=/path/to/my.tmpl                     | output as the provided go template
-  --output out-file:/my/out --output err-file:/my/err      | output to /my/out and errors to /my/err
+  --output out-file:/my/out --output log-file:/my/log      | output to /my/out and logs to /my/log
   --output none                                            | ignore events output
 Use this flag multiple times to choose multiple output options
 `
 }
 
-func PrepareOutput(outputSlice []string) (tracee.OutputConfig, printer.Config, error) {
-	outcfg := tracee.OutputConfig{}
+type OutputConfig struct {
+	tracee.OutputConfig
+	LogPath string
+	LogFile *os.File
+}
+
+func PrepareOutput(outputSlice []string) (OutputConfig, printer.Config, error) {
+	outcfg := OutputConfig{}
 	printcfg := printer.Config{}
 	printerKind := "table"
 	outPath := ""
@@ -68,6 +74,8 @@ func PrepareOutput(outputSlice []string) (tracee.OutputConfig, printer.Config, e
 			}
 		case "out-file":
 			outPath = outputParts[1]
+		case "log-file":
+			outcfg.LogPath = outputParts[1]
 		case "option":
 			switch outputParts[1] {
 			case "stack-addresses":
@@ -113,6 +121,22 @@ func PrepareOutput(outputSlice []string) (tracee.OutputConfig, printer.Config, e
 		dir := filepath.Dir(outPath)
 		os.MkdirAll(dir, 0755)
 		printcfg.OutFile, err = os.Create(outPath)
+		if err != nil {
+			return outcfg, printcfg, fmt.Errorf("failed to create output path: %v", err)
+		}
+	}
+
+	if outcfg.LogPath == "" {
+		outcfg.LogFile = os.Stderr
+	} else {
+		errPath := outcfg.LogPath
+		fileInfo, err := os.Stat(errPath)
+		if err == nil && fileInfo.IsDir() {
+			return outcfg, printcfg, fmt.Errorf("cannot use a path of existing directory %s", errPath)
+		}
+		dir := filepath.Dir(errPath)
+		os.MkdirAll(dir, 0755)
+		outcfg.LogFile, err = os.Create(errPath)
 		if err != nil {
 			return outcfg, printcfg, fmt.Errorf("failed to create output path: %v", err)
 		}
--- a/pkg/cmd/flags/output.go
+++ b/pkg/cmd/flags/output.go
@@ -46,7 +46,6 @@ func PrepareOutput(outputSlice []string) (tracee.OutputConfig, printer.Config, e
 	printcfg := printer.Config{}
 	printerKind := "table"
 	outPath := ""
-	errPath := ""
 	for _, o := range outputSlice {
 		outputParts := strings.SplitN(o, ":", 2)
 		numParts := len(outputParts)
@@ -69,8 +68,6 @@ func PrepareOutput(outputSlice []string) (tracee.OutputConfig, printer.Config, e
 			}
 		case "out-file":
 			outPath = outputParts[1]
-		case "err-file":
-			errPath = outputParts[1]
 		case "option":
 			switch outputParts[1] {
 			case "stack-addresses":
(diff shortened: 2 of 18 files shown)


cmd/tracee-ebpf/internal/printer/printer.go
```diff
@@ -195,15 +178,11 @@ func (p *templateEventPrinter) Init() error {
...
-func (p templateEventPrinter) Error(err error) {
-	fmt.Fprintf(p.err, "%v\n", err)
-}
-
 func (p templateEventPrinter) Print(event trace.Event) {
 	if p.templateObj != nil {
 		err := (*p.templateObj).Execute(p.out, event)
 		if err != nil {
-			p.Error(err)
+			logger.Error("error executing tempalte", err)
```
geyslan, line 185:
```suggestion
			logger.Error("error executing template", err)
```

cmd/tracee-ebpf/internal/printer/printer.go
```diff
@@ -227,15 +206,11 @@ func (p jsonEventPrinter) Preamble() {}
 func (p jsonEventPrinter) Print(event trace.Event) {
 	eBytes, err := json.Marshal(event)
 	if err != nil {
-		p.Error(err)
+		logger.Error("error marshaling event to json", err)
```
geyslan, line 209: ditto

cmd/tracee-ebpf/internal/printer/printer.go
```diff
@@ -266,14 +241,10 @@ func (p *gobEventPrinter) Preamble() {}
 func (p *gobEventPrinter) Print(event trace.Event) {
 	err := p.outEnc.Encode(event)
 	if err != nil {
-		p.Error(err)
+		logger.Error("error encoding event to gob", err)
```
geyslan, line 244: ditto

pkg/logger/logger.go
```diff
@@ -283,6 +283,21 @@ func (l *Logger) Sync() error {
 	return l.l.Sync()
 }
 
+// Setters
+// TODO: add SetWriter and Setlogger to the struct level?
```
geyslan, line 287: logger offers two ways of logging... by the package level and by receiver level. In the tracee we are only using the pkg level. For the sake of compliance I would add the option for receivers. If in the future we are not going to use different loggers (instantiations), we can keep the pkg level by just removing the rest.

pkg/ebpf/write_capture.go
```diff
@@ -155,7 +156,7 @@ func (t *Tracee) processFileWrites() {
 			// https://github.com/aquasecurity/libbpfgo/issues/122
 			if lost > 0 {
 				t.stats.LostWrCount.Increment(int(lost))
-				t.config.ChanErrors <- fmt.Errorf("lost %d write events", lost)
+				logger.Warn(fmt.Sprintf("lost %d write events", lost))
```
geyslan, line 159: ditto

pkg/ebpf/net_events.go
```diff
@@ -303,7 +304,7 @@ func (t *Tracee) processNetEvents(ctx gocontext.Context) {
 			// https://github.com/aquasecurity/libbpfgo/issues/122
 			if lost > 0 {
 				t.stats.LostNtCount.Increment(int(lost))
-				t.config.ChanErrors <- fmt.Errorf("lost %d network events", lost)
+				logger.Warn(fmt.Sprintf("lost %d network events", lost))
```
geyslan, line 307: ditto

pkg/ebpf/events_pipeline.go
```diff
@@ -425,5 +426,5 @@ func MergeErrors(cs ...<-chan error) <-chan error {
 
 func (t *Tracee) handleError(err error) {
 	t.stats.ErrorCount.Increment()
-	t.config.ChanErrors <- err
+	logger.Error("tracee encountered an error", err)
```
geyslan, line 429: ditto

pkg/ebpf/events_processor.go
```diff
@@ -23,7 +24,7 @@ func (t *Tracee) processLostEvents() {
 		// https://github.com/aquasecurity/libbpfgo/issues/122
 		if lost > 0 {
 			t.stats.LostEvCount.Increment(int(lost))
-			t.config.ChanErrors <- fmt.Errorf("lost %d events", lost)
+			logger.Warn(fmt.Sprintf("lost %d events", lost))
```
geyslan, line 27: ditto

cmd/tracee-ebpf/internal/printer/printer.go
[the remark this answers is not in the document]
geyslan: But we have to standardize it for our own good.

AlonZivony:
@NDStrahilevitz made an issue for this - #2404.
Please add to the description as fixed by this PR.

pkg/logger/logger.go
NDStrahilevitz: My initial solution was first implementing the receiver then using it globally, however it didn't work as expected. I don't think we actually use the receiver level at all right? Maybe we should consider removing it entirely.
(note: trivy also uses a similar strategy with only a global logger they initialize once with an InitGlobal like method. It would simplify our logger a bit to be similar to that).

cmd/tracee-ebpf/main.go
```diff
@@ -148,7 +148,8 @@ func main() {
 			}
 
 			printerConfig.ContainerMode = containerMode
-			cfg.Output = &output
+			cfg.Output = &output.OutputConfig
+			logger.SetWriter(output.LogFile)
```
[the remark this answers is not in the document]
NDStrahilevitz: Programmatically we can set it through SetBase. 
From my experience working  on this package now I think we should consider moving away from all the implicit env vars, since it made logic much more complex with a single global Init() point of entry.

cmd/tracee-ebpf/internal/printer/printer.go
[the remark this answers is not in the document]
NDStrahilevitz: I'll fix it, but I must say it's a bit unintuitive to work like this with implicit key values, since the usage of the function does not imply it at all.

NDStrahilevitz:
@rafaeldtinoco had to run format fixes since lint wasn't passing for some reason.

pkg/logger/logger.go
geyslan: > I don't think we actually use the receiver level at all right? Maybe we should consider removing it entirely.

I agree. I just implemented it as a option since we didn't have a specific goal that time. So let's remove the receiver option.

> (note: trivy also uses a similar strategy with only a global logger they initialize once with an InitGlobal like method. It would simplify our logger a bit to be similar to that).

Alright.

cmd/tracee-ebpf/main.go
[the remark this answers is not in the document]
geyslan: > Programmatically we can set it through SetBase. From my experience working on this package now I think we should consider moving away from all the implicit env vars, since it made logic much more complex with a single global Init() point of entry.

Should env vars be checked before the logger.Init() (outside the logger)?

cmd/tracee-ebpf/internal/printer/printer.go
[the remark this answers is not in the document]
geyslan: It's the way structured logs like zap work.

But logrus has a different API:

image

---

Anyway our API could easily be changed to:

```go
logger.Error("executing template", err)

...

func Error(msg string, err error, keysAndValues ...interface{}) {
	errorw(1, pkgLogger, fmt.Sprintf("%s - %s", msg, err.Error()), keysAndValues...)
}
```

In this case, if there's no error, we must use logger.Info()/Warn() etc.

rafaeldtinoco:
@NDStrahilevitz patch:

```diff
diff --git a/pkg/cmd/flags/flags_test.go b/pkg/cmd/flags/flags_test.go
index e8197cae..e5b87a0e 100644
--- a/pkg/cmd/flags/flags_test.go
+++ b/pkg/cmd/flags/flags_test.go
@@ -103,8 +103,8 @@ func TestPrepareFilter(t *testing.T) {
                        expectedError: filters.InvalidValue("-1"),
                },
                {
-                       testName:      "invalid uid 1",
-                       filters:       []string{"uid=   "},
+                       testName: "invalid uid 1",
+                       filters: []string{"uid= "},
                        expectedError: filters.InvalidValue("\t"),
                },
                {
```

NDStrahilevitz:
@yanivagman @rafaeldtinoco style tests (and normal) are passing now, your reviews would be appreciated

cmd/tracee-ebpf/main.go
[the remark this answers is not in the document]
NDStrahilevitz: I think we should drop the env vars, it wasn't a good design decision in hindsight, library consumers should Init the logger on their own if they want non default behavior (this is what we do in other aqua products too).

cmd/tracee-ebpf/internal/printer/printer.go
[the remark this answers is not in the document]
geyslan: It (typed pair) was initially proposed but the use of unspecified types (through interfaces) was picked in the merged PR. I think that were comments asking to make the API cleaner (simple).

> Anyway I fixed the logs to align with it.

👍🏼

docs/docs/tracing/output-formats.md
```diff
@@ -92,8 +92,8 @@ output and standard error:
 
 2. Error file
 
-    Redirect errors to your log files if needed:
+    Redirect logs to a file if needed:
 
     ```text
-    $ sudo TRACEE_BPF_FILE=do-not-exist ./dist/tracee-ebpf --output json --trace comm=bash --trace follow --trace event=openat --output out-file:/tmp/tracee.log --output err-file:/tmp/tracee.err
+    $ sudo TRACEE_BPF_FILE=do-not-exist ./dist/tracee-ebpf --output json --trace comm=bash --trace follow --trace event=openat --output out-file:/tmp/tracee.log --output log-file:/tmp/tracee.log
```
yanivagman, line 98: tracee.log is used in this example both for out-file and log-file. Better change one of them to something else

cmd/tracee-ebpf/internal/printer/printer.go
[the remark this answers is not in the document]
geyslan: @NDStrahilevitz Do you think it's worth opening a discussion about the logger api?

pkg/cmd/flags/output.go
```diff
@@ -19,7 +19,7 @@ Possible options:
 [format:]gob                                       output events in gob format
 [format:]gotemplate=/path/to/template              output events formatted using a given gotemplate file
 out-file:/path/to/file                             write the output to a specified file. create/trim the file if exists (default: stdout)
-err-file:/path/to/file                             write the errors to a specified file. create/trim the file if exists (default: stderr)
+log-file:/path/to/file                             write the errors to a specified file. create/trim the file if exists (default: stderr)
```
yanivagman, line 22: comment should be changed from `write the errors...` to `write the logs...`

docs/docs/tracing/output-formats.md
NDStrahilevitz: `tracee.events` and `tracee.log` respectively?

yanivagman: > `tracee.events` and `tracee.log` respectively?

Can be. The main point is to have different files

NDStrahilevitz:
@yanivagman fixed the docs issues you had, are we good to go with this?

yanivagman (approved):
LGTM

Issue #2407, opened 2022-11-29
[FEAT] add EventName to SignatureMetadata


josedonizetti:
## Feature description

Add `EventName` to `detect.SignatureMetadata`. The new field will be used to definite the event name used when loading rules to the `events.Definitions` table. This issue is part of the EPIC: #2355

More context for this issue can be found on the POC implementation: https://github.com/aquasecurity/tracee/pull/2374

PR #2408, opened 2022-11-29
types: add EventName to SignatureMetadata

josedonizetti:
Fix for https://github.com/aquasecurity/tracee/issues/2407

## Feature description

Add `EventName` to `detect.SignatureMetadata`. The new field will be used to definite the event name used when loading rules to the `events.Definitions` table.

More context for this issue can be found on the POC implementation: https://github.com/aquasecurity/tracee/pull/2374


rafaeldtinoco (approved):
LGTM

Issue #2409, opened 2022-11-29
[FEAT] Add EventName to golang signatures


josedonizetti:
## Feature description

Add `EventName` to golang signatures. The new field was introduced on the PR https://github.com/aquasecurity/tracee/pull/2408. This issue is part of EPIC: #2355

The usage of the event name will be for loading rules as events when booting tracee:

```
sudo tracee --event=anti_debugging_ptraceme
```

More context for this issue can be found on the POC implementation: https://github.com/aquasecurity/tracee/pull/2374

PR #2412, opened 2022-11-29
signatures: add event name to golang sigs

josedonizetti:
Fix https://github.com/aquasecurity/tracee/issues/2409

### DIFF
--- a/cmd/tracee-rules/output_test.go
+++ b/cmd/tracee-rules/output_test.go
@@ -290,6 +290,7 @@ func TestOutputTemplates(t *testing.T) {
 					ID:          "TRC-1",
 					Version:     "0.1.0",
 					Name:        "Standard Input/Output Over Socket",
+					EventName:   "stdio",
 					Description: "Redirection of process's standard input/output to socket",
 					Tags:        []string{"linux", "container"},
 					Properties: map[string]interface{}{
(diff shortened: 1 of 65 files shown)


yanivagman (commented):
Some event names are very long, e.g. cgroup_notify_on_release_modification.
This can be inconventient for the user to spefcify such a long event name, and will also cause the table printer to have wrong identations (event name is expected to be up to 15 chars)

josedonizetti:
@yanivagman Agree names too big might be inconvenient. But event names also must be meaningful, and 15 charts might be too short for it. So table output should expect a higher limit for indentation. Maybe 30 charts?

I'll refactor a few of the existing ones.

yanivagman:
> @yanivagman Agree names too big might be inconvenient. But event names also must be meaningful, and 15 charts might be too short for it. So table output should expect a higher limit for indentation. Maybe 30 charts?
> 
> I'll refactor a few of the existing ones.

30 chars may be too wide - see the output of the table printer, this will cause events to be split into two lines.
Anyway, we can also trim the event name from the table printer, I think this is also an option.

josedonizetti:
@yanivagman I don't think we should limit names due to table output which is something used for quick tests only, anyone using tracee seriously in dev/staging/production will be using JSON, or another format, but not a table. Or do you see use cases specifically for table?

yanivagman:
> @yanivagman I don't think we should limit names due to table output which is something used for quick tests only, anyone using tracee seriously in dev/staging/production will be using JSON, or another format, but not a table. Or do you see use cases specifically for table?

This is why I suggested to trim the event name for table output

yanivagman (approved):
LGTM

Issue #2415, opened 2022-11-29
[REFAC] extract cmd/tracee-ebpf logic


josedonizetti:
The logic from `cmd/tracee-ebpf` will be called from two places with the introduction of a new binary `cmd/tracee`. Let's extract the logic to a new package `pkg/cmd` so we don't have to repeat ourselves on both binaries.

PR #2416, opened 2022-11-29
tracee-ebpf: extract logic into pkg/cmd

josedonizetti:
Fix https://github.com/aquasecurity/tracee/issues/2415

This is a refactoring, nothing should have changed besides restructuring. 

This PR extracts the `cmd/tracee-ebpf` login into a new package `pkg/cmd`. This is necessary because we will be using this logic in multiple binaries (perhaps even inside aqua apps). It also breaks the booting into two steps:

1 - parsing flags and returning configurations
2 - running tracee

This is done because we want to decouple the flags library from running tracee, allowing us to change the library in the future, but also others using tracee as a lib. 

For extra context check the PR https://github.com/aquasecurity/tracee/pull/2374


cmd/tracee-rules/main.go
```diff
@@ -136,26 +136,18 @@ func main() {
...
+			)
 
-				if c.Bool(server.PProfEndpointFlag) {
-					httpServer.EnablePProfEndpoint()
-				}
+			if err != nil {
+				return err
+			}
 
+			if httpServer != nil {
```
geyslan, line 150: Should we have a log if httpServer is nil?

josedonizetti: hmm...why? if the user didn't pass any flag to enable it, why log?

geyslan: I got it. I was already rethinking my question, since rules are going to die.

geyslan (approved):
LGTM

rafaeldtinoco (approved):
LGTM as well.

Issue #2410, opened 2022-11-29
[BUG] counter.Counter can overflow


geyslan:
## Prerequisites

- [x] This affects latest released version.
- [x] This affects current development tree (origin/HEAD).
- [x] There isn't an issue describing the bug.

Select one OR another:

- [x] I'm going to create a PR to solve this (assign to yourself).

## Bug description

`counter.Counter` can overflow. As Tracee doesn't make use of counters with negative values, changing it to uint64 could facilitate verification (in this case only if Counter wrap around).

PR #2427, opened 2022-12-01
pkg/counter: change Counter type

geyslan:
3a4a95a

    pkg/counter: change Counter type
    
    - change count type to uint64
    - encapsulate the count avoiding non-atomic override
      - atomic override can be done with Set()
    - Increment() and Decrement() now return error if wrap occurs

## Initial Checklist

Fixes: #2410 #2428

## Type of change

### DIFF
--- a/pkg/counter/counter.go
+++ b/pkg/counter/counter.go
@@ -1,31 +1,94 @@
 package counter
 
-import "sync/atomic"
+import (
+	"fmt"
+	"strconv"
+	"sync/atomic"
+)
 
-type Counter int32
-
-func (c *Counter) Increment(amount ...int) {
-	sum := 1
-	if len(amount) > 0 {
-		sum = 0
-		for _, a := range amount {
-			sum = sum + a
-		}
-	}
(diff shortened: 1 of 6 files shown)


geyslan:
More about the reasons for this PR: https://github.com/aquasecurity/tracee/pull/2352#discussion_r1037471397

pkg/counter/counter.go
```diff
@@ -1,31 +1,94 @@
...
+
+	return nil
+}
+
+func (c *Counter) decAtomic(v uint64) error {
+	v = ^uint64(v - 1)
+	if (v + c.count) > c.count {
+		return errorCounterWrapAround()
+	}
+	atomic.AddUint64(&c.count, v)
```
yanivagman, line 71: v is of type uint, so I don't understand how this code works (will always increment and not decrement)

geyslan: atomic pkg only provides Add (no Sub). To get around this is just to do a two's complement calculation: v = ^uint64(v - 1).

So the adder will subtract instead.

Issue #2419, opened 2022-12-01
[BUG] sched_process_exec program too big


AlonZivony:
## Prerequisites

- [ ] This affects latest released version.
- [x] This affects current development tree (origin/HEAD).
- [x] There isn't an issue describing the bug.

Select one OR another:

- [x] I'm going to create a PR to solve this (assign to yourself).
- [ ] Someone else should solve this.

## Bug description

`sched_process_exec` implementation became too big of event to work in old supported kernels.
Execution error in 4.19:
```libbpf: prog 'tracepoint__sched__sched_process_exec': BPF program load failed: Argument list too long```
We need to reduce its size so it will be compatible to all supported versions.

PR #2434, opened 2022-12-06
Feature/reduce sched exec instruction

AlonZivony:
## Initial Checklist

## Description (git log)

Copy and paste the git log here.

Fixes: #2419

## Type of change

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -146,6 +146,7 @@ enum tail_call_id_e
     TAIL_SEND_BIN,
     TAIL_SEND_BIN_TP,
     TAIL_KERNEL_WRITE,
+    TAIL_SCHED_PROCESS_EXEC_EVENT_SUBMIT,
     MAX_TAIL_CALL
 };
 
@@ -612,10 +613,6 @@ enum kconfig_key_e
 
 #endif // CORE
 
-// FUNCTIONS DECLARATIONS --------------------------------------------------------------------------
-
-static __always_inline void *get_path_str(struct path *path);
-
 // EBPF MACRO HELPERS ------------------------------------------------------------------------------
 
 #ifndef CORE
@@ -993,6 +990,7 @@ BPF_HASH(module_init_map, u32, kmod_data_t, 256);                  // holds modu
 BPF_LRU_HASH(fd_arg_path_map, fd_arg_task_t, fd_arg_path_t, 1024); // store fds paths by task
 BPF_LRU_HASH(bpf_attach_map, u32, bpf_attach_t, 1024);             // holds bpf prog info
 BPF_LRU_HASH(bpf_attach_tmp_map, u32, bpf_attach_t, 1024);         // temporarily hold bpf_attach_t
+BPF_PERCPU_ARRAY(cached_event_data_map, event_data_t, 1);          // cached event data between chained tail calls
 
 // clang-format on
 
@@ -1001,6 +999,10 @@ BPF_LRU_HASH(bpf_attach_tmp_map, u32, bpf_attach_t, 1024);         // temporaril
 BPF_PERF_OUTPUT(events, 1024);      // events submission
 BPF_PERF_OUTPUT(file_writes, 1024); // file writes events submission
 BPF_PERF_OUTPUT(net_events, 1024);  // network events submission
+
+// FUNCTIONS DECLARATIONS --------------------------------------------------------------------------
+
+static __always_inline void *get_path_str(struct path *path);
 
 // HELPERS: DEVICES --------------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


yanivagman (commented):
I see that in this PR you've added 4 new tail calls to sched_process_exec.
It is preferable to only add one since the max tail calls allowed is limited to 32, and we have future plans to use it (e.g. for filtering)

AlonZivony:
> I see that in this PR you've added 4 new tail calls to sched_process_exec. It is preferable to only add one since the max tail calls allowed is limited to 32, and we have future plans to use it (e.g. for filtering)

The problem is that we have the cap of 4096 instructions (if I remember correctly).
When  I compile the program on my machine, it is about 12.5k instructions.
It means that we don't have a choice but to divide it to at least 3 functions, preferably 4.
The reason I have 5 is for logic reasons - I tried to divide to functions where each contain understandable and describable logic.
If you have better division for it I will more then welcome the idea.

AlonZivony:
BTW, it seems that the `uprobe_syscall_trigger` program now crashes Tracee after we solved the instructions issues.
Is it something we know about from before?

Another issue - in this PR I save only the submit buffer state between programs.
I am debating with myself if we might want to save the entire `event_data_t` struct, only initializing pointers to other map members, so we won't need to recalculate all the members each time. It will increase the size of this PR and make it harder to finish approving before version release, and it will also add more memory for the new map and copying operations.
I am opening a debate here - do you think we should keep it this way or change to caching the entire data struct?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1771,12 +1782,20 @@ static __always_inline int init_event_data(event_data_t *data, void *ctx)
...
 
+    submit_buf_state_t *preserved_buffer = get_preserved_submit_buffer_data();
+    if (preserved_buffer != NULL && preserved_buffer->is_initialized == true) {
+        data->buf_off = preserved_buffer->buf_off;
+        data->context.argnum = preserved_buffer->argnum;
+        delete_preserved_submit_buffer_data();
+    } else {
+        data->buf_off = sizeof(event_context_t);
+    }
+
```
yanivagman, line 1798: A different approach will be to save the entire event_data_t in a percpu map before calling a tail call, then getting it in the tail call and initializing the pointers to other maps in a dedicated function, not related to init_event_data

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3486,27 +3529,71 @@ int tracepoint__sched__sched_process_exec(struct bpf_raw_tracepoint_args *ctx)
     // We passed all filters (in should_trace()) - add this pid to traced pids set
     data.task_info->follow = true;
 
-    struct task_struct *task = (struct task_struct *) ctx->args[0];
-    struct linux_binprm *bprm = (struct linux_binprm *) ctx->args[2];
+    if (should_submit(SCHED_PROCESS_EXEC, data.config) || data.config->options & OPT_PROCESS_INFO) {
```
yanivagman, line 3532: Now that you moved everything under that `if`, it will make more sense to check the negation of this condition and return if true. This will save the extra indentation.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3486,27 +3529,71 @@ int tracepoint__sched__sched_process_exec(struct bpf_raw_tracepoint_args *ctx)
...
+        // and for the interpreter. Because the interpreter is loaded only after the executed elf is
+        // loaded, the map value of the executed elf should be overridden by the interpreter.
+        file_info_t *elf_interpreter =
+            bpf_map_lookup_elem(&interpreter_map, &data.context.task.host_tid);
+
+        save_str_to_buf(&data, (void *) filename, 0);
+        save_str_to_buf(&data, file_path, 1);
+        save_to_submit_buf(&data, &s_dev, sizeof(dev_t), 4);
+        save_to_submit_buf(&data, &inode_nr, sizeof(unsigned long), 5);
+        save_to_submit_buf(&data, &ctime, sizeof(u64), 7);
```
yanivagman, line 3563: I see that some of the fields are not ordered according to the event's parameters.
I think we should either change the order in events.go as well to fit what we have here, or split this program and the tailcall so the indexes and order will be kept as they were.

yanivagman (changes requested):

rafaeldtinoco:
Alright I merged https://github.com/aquasecurity/tracee/pull/2461 which will fix (hopefully) the issue with the network tests in v4.19 vanilla kernel. Feel free to rebase. I have re-run the failed test but I'm not sure if the test is rebased automatically (I don't think so).

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1978,6 +1980,54 @@ static __always_inline int should_submit(u32 event_id, config_entry_t *config)
...
+    if (unlikely(data->config == NULL))
+        return -4;
+
+    int buf_idx = SUBMIT_BUF_IDX;
+    data->submit_p = bpf_map_lookup_elem(&bufs, &buf_idx);
+    if (unlikely(data->submit_p == NULL))
+        return -5;
+
+    bool task_info_initalized = false;
+    data->task_info = init_task_info(data->context.task.host_tid, &task_info_initalized);
```
yanivagman, line 2018: Is there a reason to do this again in the tail call? Won't it be the same as what we've loaded?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1978,6 +1980,54 @@ static __always_inline int should_submit(u32 event_id, config_entry_t *config)
...
+    if (unlikely(data->task_info == NULL)) {
+        return -6;
+    }
+
+    u32 cgroup_id_lsb = data->context.task.cgroup_id;
+    u8 *state = bpf_map_lookup_elem(&containers_map, &cgroup_id_lsb);
```
yanivagman, line 2024: Ditto

yanivagman (approved):
Tests finally pass in 4.19!
LGTM

PR #2439, opened 2022-12-06
Add rules to the pipeline

josedonizetti:
## Initial Checklist

This PRs adds the rule engine to the events pipeline.

Fixes: https://github.com/aquasecurity/tracee/issues/2441

## How Has This Been Tested?

```
# compiles
make

# test single rule event
./dist/tracee --trace event=anti_debugging
# test event with output json
./dist/tracee --trace event=anti_debugging --output format:json

# test single event with extra non rule event
./dist/tracee --trace event=anti_debugging,execve

# test single event with non rule dependency event
./dist/tracee --trace event=ptrace,anti_debugging

# test all events
./dist/tracee  --trace event=anti_debugging,aslr_inspection,cgroup_notify_on_release,cgroup_release_agent,core_pattern_modification,default_loader_mod,disk_mount,docker_abuse,dynamic_code_loading,fileless_execution,hidden_file_created,illegitimate_shell,k8s_api_connection,k8s_cert_theft,kernel_module_loading,ld_preload,process_vm_write_inject,proc_fops_hooking,proc_kcore_read,proc_mem_access,proc_mem_code_injection,ptrace_code_injection,rcd_modification,sched_debug_recon,scheduled_task_mod,stdio_over_socket,sudoers_modification,syscall_hooking,system_request_key_mod

# test rule event non exported, won't work because the event isn't exported
./dist/tracee --output format:json --trace event=dropped_executable
{"level":"fatal","ts":1670495492.8543997,"msg":"app","error":"invalid event to trace: dropped_executable"}

# list all events, including those that are rules
./dist/tracee --list
```

### DIFF
--- a/cmd/tracee/main.go
+++ b/cmd/tracee/main.go
@@ -3,12 +3,17 @@ package main
 import (
 	"context"
 	"os"
+	"strings"
 
 	"github.com/aquasecurity/tracee/pkg/cmd"
 	"github.com/aquasecurity/tracee/pkg/cmd/flags"
 	"github.com/aquasecurity/tracee/pkg/cmd/flags/server"
 	"github.com/aquasecurity/tracee/pkg/cmd/urfave"
+	"github.com/aquasecurity/tracee/pkg/events"
 	"github.com/aquasecurity/tracee/pkg/logger"
+	"github.com/aquasecurity/tracee/pkg/rules/engine"
+	"github.com/aquasecurity/tracee/pkg/rules/signature"
+	"github.com/aquasecurity/tracee/types/detect"
 
 	cli "github.com/urfave/cli/v2"
 )
@@ -37,12 +42,35 @@ func main() {
 		Usage:   "Trace OS events and syscalls using eBPF",
 		Version: version,
 		Action: func(c *cli.Context) error {
-
 			if c.NArg() > 0 {
 				return cli.ShowAppHelp(c) // no args, only flags supported
 			}
 
 			flags.PrintAndExitIfHelp(c)
+
+			// Rego command line flags
+
+			rego, err := flags.PrepareRego(c.StringSlice("rego"))
+			if err != nil {
+				return err
+			}
+
+			// event names are parsed here because we need to know which signatures to load
+			eventNames := getEventNames(c)
+
+			sigs, err := signature.Find(
+				rego.RuntimeTarget,
+				rego.PartialEval,
+				c.String("rules-dir"),
+				eventNames,
+				rego.AIO,
+			)
+
+			if err != nil {
+				return err
+			}
+
+			createEventsFromSignatures(events.StartRulesID, sigs)
 
 			if c.Bool("list") {
 				cmd.PrintEventList() // list events
@@ -52,6 +80,15 @@ func main() {
 			runner, err := urfave.GetTraceeRunner(c, version)
 			if err != nil {
 				return err
+			}
+
+			// parse arguments must be enabled if the rule engine is part of the pipeline
+			runner.TraceeConfig.Output.ParseArguments = true
+
+			runner.TraceeConfig.EngineConfig = engine.Config{
+				Enabled:             true,
+				Signatures:          sigs,
+				SignatureBufferSize: c.Uint("sig-buffer"),
 			}
 
 			ctx := context.Background()
@@ -146,6 +183,24 @@ func main() {
 				Name:  "containers",
 				Usage: "enable container info enrichment to events. this feature is experimental and may cause unexpected behavior in the pipeline",
 			},
+
+			// TODO: add webhook
+
+			// rules
+			&cli.StringFlag{
+				Name:  "rules-dir",
+				Usage: "directory where to search for rules in CEL (.yaml), OPA (.rego), and Go plugin (.so) formats",
+			},
+			&cli.StringSliceFlag{
+				Name:  "rego",
+				Usage: "Control event rego settings. run '--rego help' for more info.",
+				Value: cli.NewStringSlice(),
+			},
+			&cli.UintFlag{
+				Name:  "sig-buffer", // TODO: rename to rules-buffer?
+				Usage: "size of the event channel's buffer consumed by signatures",
+				Value: 1000,
+			},
 		},
 	}
 
@@ -153,4 +208,58 @@ func main() {
 	if err != nil {
 		logger.Fatal("app", "error", err)
 	}
+}
+
+func createEventsFromSignatures(startId events.ID, sigs []detect.Signature) {
+	id := startId
+
+	for _, s := range sigs {
+		m, err := s.GetMetadata()
+		if err != nil {
+			logger.Error("failed to load event", "error", err)
+			continue
+		}
+
+		selectedEvents, err := s.GetSelectedEvents()
+		if err != nil {
+			logger.Error("failed to load event", "error", err)
+			continue
+		}
+
+		dependencies := make([]events.ID, 0)
+
+		for _, s := range selectedEvents {
+			eventID, found := events.Definitions.GetID(s.Name)
+			if !found {
+				logger.Error("failed to load event depedency", "event", s.Name)
+				continue
+			}
+
+			dependencies = append(dependencies, eventID)
+		}
+
+		event := events.NewEventDefinition(m.EventName, []string{"rules"}, dependencies)
+
+		err = events.Definitions.Add(events.ID(id), event)
+		if err != nil {
+			logger.Error("failed to add event definition", "error", err)
+			continue
+		}
+
+		id++
+	}
+
+}
+
+func getEventNames(c *cli.Context) []string {
+	// TODO: --trace event= will become --events
+	traceArgs := c.StringSlice("trace")
+	for _, optValue := range traceArgs {
+		if strings.HasPrefix(optValue, "event=") {
+			values := strings.Split(optValue, "=")
+			return strings.Split(values[1], ",")
+		}
+	}
+
+	return []string{}
 }
--- a/cmd/tracee/main.go
+++ b/cmd/tracee/main.go
@@ -58,6 +58,12 @@ func main() {
 			// event names are parsed here because we need to know which signatures to load
 			eventNames := getEventNames(c)
 
+			// if list is enabled we set eventNames to nil, so every rule is loaded,
+			// and displayed on the list
+			if c.Bool("list") {
+				eventNames = nil
+			}
+
 			sigs, err := signature.Find(
 				rego.RuntimeTarget,
 				rego.PartialEval,
@@ -65,7 +71,6 @@ func main() {
 				eventNames,
 				rego.AIO,
 			)
-
 			if err != nil {
 				return err
 			}
@@ -73,7 +78,7 @@ func main() {
 			createEventsFromSignatures(events.StartRulesID, sigs)
 
 			if c.Bool("list") {
-				cmd.PrintEventList() // list events
+				cmd.PrintEventList(true) // list events
 				return nil
 			}
 
@@ -94,7 +99,6 @@ func main() {
 			ctx := context.Background()
 
 			return runner.Run(ctx)
-
 		},
 		Flags: []cli.Flag{
 			&cli.BoolFlag{
@@ -240,7 +244,7 @@ func createEventsFromSignatures(startId events.ID, sigs []detect.Signature) {
 
 		event := events.NewEventDefinition(m.EventName, []string{"rules"}, dependencies)
 
-		err = events.Definitions.Add(events.ID(id), event)
+		err = events.Definitions.Add(id, event)
 		if err != nil {
 			logger.Error("failed to add event definition", "error", err)
 			continue
@@ -248,7 +252,6 @@ func createEventsFromSignatures(startId events.ID, sigs []detect.Signature) {
 
 		id++
 	}
-
 }
 
 func getEventNames(c *cli.Context) []string {
--- a/cmd/tracee/main.go
+++ b/cmd/tracee/main.go
@@ -91,9 +91,11 @@ func main() {
 			runner.TraceeConfig.Output.ParseArguments = true
 
 			runner.TraceeConfig.EngineConfig = engine.Config{
-				Enabled:             true,
-				Signatures:          sigs,
-				SignatureBufferSize: c.Uint("sig-buffer"),
+				Enabled:    true,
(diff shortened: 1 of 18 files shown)


josedonizetti:
@NDStrahilevitz I tried to address all your comments from the previous PR https://github.com/aquasecurity/tracee/pull/2374, let me know if anything passed by.

pkg/rules/engine/engine.go
```diff
@@ -68,6 +70,25 @@ func NewEngine(config Config, sources EventSources, output chan detect.Finding)
...
+// StartPipeline receives an input channel, and returns an output channel
+// allowing the rule engine to be used in the events pipeline
+func StartPipeline(cfg Config, input chan protocol.Event, done chan bool) <-chan detect.Finding {
+	output := make(chan detect.Finding)
+
+	source := EventSources{Tracee: input}
+	engine, err := NewEngine(cfg, source, output)
+	if err != nil {
+		logger.Error("error creating engine: " + err.Error())
+	}
```
geyslan, line 82: If NewEngine errs, engine is nil. We need to abort.

pkg/events/events.go
```diff
@@ -74,10 +76,46 @@ type Event struct {
...
+	evt.Dependencies = d
+
+	return evt
+}
+
 type eventDefinitions struct {
 	events map[ID]Event
 }
 
+// Add add event to defitions
```
geyslan, line 104: Perhaps a comment on `var Definitions = eventDefinitions{` to make it clear that this is map that changes at runtime?

pkg/ebpf/events_pipeline.go
```diff
@@ -323,31 +330,28 @@ func (t *Tracee) sinkEvents(ctx context.Context, in <-chan *trace.Event) <-chan
 	go func() {
 		defer close(errc)
 		for event := range in {
-			// Only emit events requested by the user
+			//Only emit events requested by the user
```
geyslan, line 333:
```suggestion
			// Only emit events requested by the user
```

pkg/ebpf/events_pipeline.go
```diff
@@ -428,3 +432,21 @@ func (t *Tracee) handleError(err error) {
 	t.stats.ErrorCount.Increment()
 	logger.Error("tracee encountered an error", "error", err)
 }
+
+// parseArguments is required by the rules engine,
+// if the rule engine staged is enabled, parseArguments will happen there,
+// before sending an event to be processed, if the rule engine is disabled,
+// parseArgumnets happens on the sink stage (because anyone pipeing the output to tracee-rules will need it)
```
geyslan, line 439: I tried to rewrite this but got confused. Could you elaborate on separate sentences?

geyslan (changes requested):
First review pass: put some comments.

pkg/ebpf/finding.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"fmt"
+
+	"github.com/aquasecurity/tracee/pkg/events"
+	"github.com/aquasecurity/tracee/types/detect"
+	"github.com/aquasecurity/tracee/types/trace"
+)
+
+// FindingToEvent converts a detect.Fcingin into a trace.Event
```
geyslan, line 11:
```suggestion
// FindingToEvent converts a detect.Finding into a trace.Event
```

geyslan (approved):
Tested and LGTM. Great work!

rafaeldtinoco:
image

You're missing alignment in the "list" command. Also, a line jump in between the last signature events and System Calls events.

Also: maybe you should change "Events" to "Signature Events" so they are not confused. At least for now?

pkg/ebpf/finding.go
```diff
@@ -0,0 +1,60 @@
...
+	eventID, found := events.Definitions.GetID(f.SigMetadata.EventName)
+	if !found {
+		return nil, fmt.Errorf("error finding event not found: %s", f.SigMetadata.EventName)
+	}
+
+	return newEvent(int(eventID), f.SigMetadata.EventName, s), nil
+}
+
+func newEvent(id int, name string, s trace.Event) *trace.Event {
+	return &trace.Event{
```
rafaeldtinoco, line 29: any idea about including detect.Finding data into the event ? As Args of the event being created ?

pkg/ebpf/finding.go
```diff
@@ -0,0 +1,60 @@
...
+		ContainerImage:      s.ContainerImage,
+		ContainerName:       s.ContainerName,
+		PodName:             s.PodName,
+		PodNamespace:        s.PodNamespace,
+		PodUID:              s.PodUID,
+		ArgsNum:             s.ArgsNum,
+		ReturnValue:         s.ReturnValue,
+		StackAddresses:      s.StackAddresses,
+		ContextFlags:        s.ContextFlags,
+		Args:                s.Args,
```
rafaeldtinoco, line 57: maybe this should be the finding data instead of args from the original event ?

rafaeldtinoco (changes requested):
I left some comments, doubts and nits for you. Nice work. Let's do another round after some answers and talks. Cheers!

pkg/ebpf/events_engine.go
```diff
@@ -0,0 +1,81 @@
...
+						t.handleError(err)
+						continue
+					}
+
+					// pass the event to the sink stage, if the event is also marked as emit
+					// it will be sent to print by the sink stage
+					out <- event
+
+					// send the event to the rule event
+					engineInput <- event.ToProtocol()
```
rafaeldtinoco, line 48: Check for the bug I found for https://github.com/aquasecurity/tracee/pull/2463. You have to call `t.shouldProcessEvent()` for the event, before sending it to the rules engine. Or else, if we are filtering through the cmdline, ALL the events will be passed to the rules engine (and not only those being filtered, which is what we want).

pkg/ebpf/events_engine.go
```diff
@@ -0,0 +1,81 @@
...
+				return
+			}
+		}
+	}()
+
+	go func() {
+		for {
+			select {
+			case finding := <-engineOutput:
+				event, err := FindingToEvent(finding)
```
rafaeldtinoco, line 60: Be aware that eventually you will have to call:
```
if !t.shouldProcessEvent(event) {
    continue
}
```
in here before sending event to "out" channel. This will happen because the user might have choosen the event (Which is a signature) but they might be filtering the output of such event through args, for example. For now you're using the original event args, but eventually it will be the "finding data", right ? Then you might "filter" out detections as well, just like other events.

yanivagman (commented):
Nice work @josedonizetti!

pkg/ebpf/finding.go
NDStrahilevitz: Agreed

cmd/tracee/main.go
```diff
@@ -154,3 +215,57 @@ func main() {
...
+		id++
+	}
+
+}
+
+func getEventNames(c *cli.Context) []string {
+	// TODO: --trace event= will become --events
+	traceArgs := c.StringSlice("trace")
+	for _, optValue := range traceArgs {
+		if strings.HasPrefix(optValue, "event=") {
```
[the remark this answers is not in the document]
NDStrahilevitz: How it works is that first equality events are collected. If there are none(from event=... or set=...) it will pick the default set.
Then it will exclude all matching events picked up before.
So this technically won't have any problem since none of the signature events are included in the default set atm.

Since this function is practically only used to collect the signature names the name could maybe use a change, or at least a comment.

cmd/tracee/main_test.go
```diff
@@ -0,0 +1,107 @@
...
+				eventSelector := detect.SignatureEventSelector{Name: d}
+				selectedEvents = append(selectedEvents, eventSelector)
+			}
+
+			return selectedEvents, nil
+		},
+	}
+}
+
+type fakeSignature struct {
```
NDStrahilevitz, line 83: I think we have a lot of duplicates of this mock signature, maybe we should export one at the `signature` package?

pkg/cmd/tracee.go
```diff
@@ -97,12 +97,19 @@ func (r Runner) Run(ctx context.Context) error {
...
 	padChar, firstPadLen, secondPadLen := " ", 9, 36
 	titleHeaderPadFirst := getPad(padChar, firstPadLen)
 	titleHeaderPadSecond := getPad(padChar, secondPadLen)
 
 	var b strings.Builder
+
+	if printRulesSet {
+		b.WriteString("Events: " + titleHeaderPadFirst + "Sets:" + titleHeaderPadSecond + "Arguments:\n")
+		b.WriteString("____________  " + titleHeaderPadFirst + "____ " + titleHeaderPadSecond + "_________" + "\n\n")
+		printEventGroup(&b, events.StartRulesID, events.MaxRulesID)
```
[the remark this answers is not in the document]
NDStrahilevitz: This is only relevant if `printRulesSet` is true.
I think we should take out these small blocks of
```
b.WriteString
b.WriteString
printEventGroup
```
into their own small function then we can use these blocks to create a PrintList function for each of these cases.

pkg/cmd/tracee.go
```diff
@@ -97,12 +97,19 @@ func (r Runner) Run(ctx context.Context) error {
...
-func PrintEventList() {
+func PrintEventList(printRulesSet bool) {
 	padChar, firstPadLen, secondPadLen := " ", 9, 36
 	titleHeaderPadFirst := getPad(padChar, firstPadLen)
 	titleHeaderPadSecond := getPad(padChar, secondPadLen)
 
 	var b strings.Builder
+
+	if printRulesSet {
+		b.WriteString("Events: " + titleHeaderPadFirst + "Sets:" + titleHeaderPadSecond + "Arguments:\n")
```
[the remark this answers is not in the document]
josedonizetti: I've renamed it only to `Rules`, think it doesn't look bad, considering they are rules. About the position, up to you, want first? last? middle?

pkg/cmd/tracee.go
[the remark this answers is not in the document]
josedonizetti: I wasn't worrying about it because the idea is to stop supporting `tracee-ebpf` binary, so soon we will have only a version of it. Though I don't mind the refactoring if you think it is important, I can create a refactor issue as a follow-up.

signatures/rego/anti_debugging_ptraceme.rego
```diff
@@ -4,6 +4,7 @@ __rego_metadoc__ := {
 	"id": "TRC-2",
 	"version": "0.1.0",
 	"name": "Anti-Debugging",
+	"eventName": "anti_debugging",
```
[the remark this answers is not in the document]
josedonizetti: This change was required to have all tests pass. The whole rename was done as a separate PR, https://github.com/aquasecurity/tracee/pull/2457, now that I've rebased all regos are renamed properly already.

pkg/ebpf/events_engine.go
```diff
@@ -0,0 +1,81 @@
...
+			}
+		}
+	}()
+
+	go func() {
+		defer close(done)
+
+		<-ctx.Done()
+		done <- true
+	}()
```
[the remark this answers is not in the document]
josedonizetti: Yup, I also want to do it, didn't do it here because I want to avoid changing tracee-rules. I'll create a follow-up task as a refactor. Ok?

cmd/tracee/main_test.go
josedonizetti: LIke the idea, will create a task as a follow-up refactor, ok?

cmd/tracee/main.go
```diff
@@ -146,6 +189,24 @@ func main() {
...
+			&cli.StringSliceFlag{
+				Name:  "rego",
+				Usage: "Control event rego settings. run '--rego help' for more info.",
+				Value: cli.NewStringSlice(),
+			},
+			&cli.UintFlag{
+				Name:  "sig-buffer", // TODO: rename to rules-buffer?
+				Usage: "size of the event channel's buffer consumed by signatures",
+				Value: 1000,
+			},
```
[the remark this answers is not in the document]
yanivagman: Do we want to expose this in the new tracee binary?
What advantage does it have, except confusing the user with an additional flag?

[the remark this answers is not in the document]
josedonizetti: How about removing, setting it to 1000, and getting feedback from users, I don't think we should drop it from everywhere now, but we could experiment without it on the new binary.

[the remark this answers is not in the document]
yanivagman: > How about removing, setting it to 1000, and getting feedback from users, I don't think we should drop it from everywhere now, but we could experiment without it on the new binary.

That what I had in mind as well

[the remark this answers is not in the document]
josedonizetti: done!

cmd/tracee/main.go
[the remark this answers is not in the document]
rafaeldtinoco: What if you define a set and then use != ?

pkg/cmd/tracee.go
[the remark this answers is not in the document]
rafaeldtinoco: Not sure.. lets keep first then =)

pkg/ebpf/events_engine.go
rafaeldtinoco: If the feature will be a "tech preview" I think it is fine do have an issue and mark it for the next release (considering we are out of time for this one).

cmd/tracee/main.go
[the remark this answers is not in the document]
josedonizetti: Example?

rafaeldtinoco (approved):
All good on my side. Will let you handle the rest of the comments with others, but +1'ing on my side.

pkg/ebpf/events_engine.go
josedonizetti: It won't be the default solution, for now, the only way to use it will be purposely choose it, also will make sure to note it as experimental on the release notes/docs.

cmd/tracee/main_test.go
NDStrahilevitz: Yep

cmd/tracee/main.go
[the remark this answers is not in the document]
NDStrahilevitz: for example `set=container` `event!=container_remove` will add all container events but remove container_remove. this can potentially clash with this mechanism.
But it looks like you're using this to collect the signature events only so the function should reflect that as well as docs.

Issue #2430, opened 2022-12-05
[FEAT] Trim event name for the table output


josedonizetti:
Some event names might be too big to the table output, let's trim them to a max of 20 chars before printing.

context: https://github.com/aquasecurity/tracee/pull/2412

PR #2440, opened 2022-12-06
tracee: trim event name for table output

josedonizetti:
Fixes https://github.com/aquasecurity/tracee/issues/2430


josedonizetti:
@yanivagman is it what you had in mind?

yanivagman (approved):
LGTM

Issue #2444, opened 2022-12-07
[FEAT] re-enable the signature `dropped_executable`


josedonizetti:
The `dropped_executable` signature was disabled when we added the golang rules due to false positives. Let's enable it back again.

Issue #2353, opened 2022-11-11
Update rules examples


yanivagman:
We currently give gdb (which uses ptrace) as an example to trigger the Anti Debugging rule. This example doesn't represent the real use of Tracee to detect attacks and we should replace it with a real attack scenario. A good example can be the detection of fileless execution.

josedonizetti:
This was implemented on the PR https://github.com/aquasecurity/tracee/pull/2797

-> https://aquasecurity.github.io/tracee/v0.16/getting-started/kubernetes-quickstart/#exercising-a-security-event

PR #2445, opened 2022-12-07
rules: reenable dropped_executable

josedonizetti:
Fixes https://github.com/aquasecurity/tracee/issues/2444

Configures the Origin to be the same we used for rego: https://github.com/aquasecurity/tracee/blob/main/signatures/rego/dropped_executable.rego#L20

I've tested it on GKE and DO, adding the context to be containers fixes the false positives.
Note, I want to use this signatures+nginxrat malware to do the new tracee get started tutorial example -> https://github.com/aquasecurity/tracee/issues/2353


yanivagman (approved):
LGTM

PR #2457, opened 2022-12-08
rules: add event name to rego signatures

josedonizetti:
Fixes https://github.com/aquasecurity/tracee/issues/2411. This is not a mandatory change, because we are not using the rego signatures, though I would like to keep it consistent with the changes done on the golang signatures here -> https://github.com/aquasecurity/tracee/issues/2409


geyslan (approved):
LGTM

PR #2461, opened 2022-12-08
network: e2e-net-test v419 skip should return 0

rafaeldtinoco:
## Description (git log)

... or else the v419 non CORE tests will fail.

Fixes: commit 8c67067b

Fixes: PR: #2456 (it had a bug)

## Type of change

### DIFF
--- a/tests/e2e-net-test.sh
+++ b/tests/e2e-net-test.sh
@@ -4,18 +4,17 @@
 # This test is executed by github workflows inside the action runners
 #
 
-error_info() {
+info_exit() {
     echo -n "INFO: "
     echo $@
-    exit 1
+    exit 0
 }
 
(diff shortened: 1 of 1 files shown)


tests/e2e-net-test.sh
```diff
@@ -7,15 +7,14 @@
 error_info() {
     echo -n "INFO: "
     echo $@
-    exit 1
+    exit 0
```
yanivagman, line 10: This function is called error_info, so don't you want to return a value != 0

rafaeldtinoco: it should be called exit_info, fixing

rafaeldtinoco: fixed.

yanivagman (approved):
LGTM

Issue #2462, opened 2022-12-08
[BUG] derived events should also be filtered


rafaeldtinoco:
## Prerequisites

- [ ] This affects latest released version.
- [x] This affects current development tree (origin/HEAD).
- [ ] There isn't an issue describing the bug.

Select one OR another:

- [x] I'm going to create a PR to solve this (assign to yourself).
- [ ] Someone else should solve this.

## Bug description

This is how I'm executing tracee (cmdline):

```
sudo ./dist/tracee-ebpf --install-path /tmp/tracee --cache cache-type=mem --cache mem-cache-size=512 --output format:json --output option:parse-arguments --capabilities bypass=false --trace comm=ping --trace event=net_packet_ipv4 --trace net_packet_ipv4.args.dst=8.8.8.8,1.1.1.1
```

This is the error I'm getting: Filters are not applied, I get all network events.

```
$ sudo ./dist/tracee-ebpf --install-path /tmp/tracee --cache cache-type=mem --cache mem-cache-size=512 --output format:json --output option:parse-arguments --capabilities bypass=false --trace comm=ping --trace event=net_packet_ipv4 --trace net_packet_ipv4.args.dst=8.8.8.8,1.1.1.1 
new string filter
new string filter
new string filter
new arg filter
string parsing
values: [ping]
equal, val: ping
equal
enabling string filter: &{map[ping:true] map[] {map[] map[] [] 9223372036854775807} {map[] map[] [] 9223372036854775807} map[] {map[] map[] [] 9223372036854775807} {map[] map[] [] 9223372036854775807} map[] false}
id: 2007
argName: dst
operatorAndValues: =8.8.8.8,1.1.1.1
setting filter for id: 2007
constructing filter for id (2007) for argName: dst
new string filter
1 x filter.filters[id] = map[dst:0xc0001ee5a0]
1 x filter.filters[id][argName] = &{map[] map[] {map[] map[] [] 9223372036854775807} {map[] map[] [] 9223372036854775807} map[] {map[] map[] [] 9223372036854775807} {map[] map[] [] 9223372036854775807} map[] false}
string parsing
values: [8.8.8.8 1.1.1.1]
equal, val: 8.8.8.8
equal
equal, val: 1.1.1.1
equal
[...]

PR #2463, opened 2022-12-08
events_pipeline: run filters for derived events

rafaeldtinoco:
## Description (git log)

When the cmdline has arg filters for derived events, for example, they are never filtered (Equal, NotEqual, Contains, NotContains) as their Event ID comes from the functions DeriveSingleEvent & DeriveMultiple Events, and not from the perfbuffer event, like regular events.

Their filtering stage should be done in a later stage in the pipeline, when derivatives (derived events) are obtained, and not in the decodeEvents phase. This patch fixes the issue by running all filters for the derived event being processed.

Fixes: #2462

## Type of change

## How Has This Been Tested?

After the patch is applied, the network events can be filtered by "src" and "dst" (at least, need to check other args). This makes it easier to check the events as network events sometimes can be too noisy.

```
$ sudo ./dist/tracee-ebpf --install-path /tmp/tracee --cache cache-type=mem --cache mem-cache-size=512 --output format:json --output option:parse-arguments --capabilities bypass=false --trace comm=ping --trace event=net_packet_ipv4 --trace net_packet_ipv4.args.dst=8.8.8.8
{"timestamp":1670531130757400611,"threadStartTime":266674446820342,"processorId":16,"processId":2393710,"cgroupId":155653,"threadId":2393710,"parentProcessId":2393482,"hostProcessId":2393710,"hostThreadId":2393710,"hostParentProcessId":2393482,"userId":1000,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"ping","hostName":"fujitsu","containerId":"","containerImage":"","containerName":"","podName":"","podNamespace":"","podUID":"","eventId":"2007","eventName":"net_packet_ipv4","argsNum":3,"returnValue":0,"stackAddresses":[0],"contextFlags":{"containerStarted":false},"args":[{"name":"src","type":"const char*","value":"192.168.100.2"},{"name":"dst","type":"const char*","value":"8.8.8.8"},{"name":"proto_ipv4","type":"trace.ProtoIPv4","value":{"version":4,"IHL":5,"TOS":0,"length":84,"id":7740,"flags":2,"fragOffset":0,"TTL":64,"protocol":"ICMPv4","checksum":59314,"srcIP":"192.168.100.2","dstIP":"8.8.8.8"}}]}
{"timestamp":1670531131758246337,"threadStartTime":266674446820342,"processorId":16,"processId":2393710,"cgroupId":155653,"threadId":2393710,"parentProcessId":2393482,"hostProcessId":2393710,"hostThreadId":2393710,"hostParentProcessId":2393482,"userId":1000,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"ping","hostName":"fujitsu","containerId":"","containerImage":"","containerName":"","podName":"","podNamespace":"","podUID":"","eventId":"2007","eventName":"net_packet_ipv4","argsNum":3,"returnValue":0,"stackAddresses":[0],"contextFlags":{"containerStarted":false},"args":[{"name":"src","type":"const char*","value":"192.168.100.2"},{"name":"dst","type":"const char*","value":"8.8.8.8"},{"name":"proto_ipv4","type":"trace.ProtoIPv4","value":{"version":4,"IHL":5,"TOS":0,"length":84,"id":7927,"flags":2,"fragOffset":0,"TTL":64,"protocol":"ICMPv4","checksum":59127,"srcIP":"192.168.100.2","dstIP":"8.8.8.8"}}]}
```

@NDStrahilevitz FYI, and I think that I'll need to create network filters as each network event has its own Type, right ? (So I can filter by ProtoIPv4, ProtoDNS, etc, arguments).

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/ebpf/events_pipeline.go
+++ b/pkg/ebpf/events_pipeline.go
@@ -305,6 +305,10 @@ func (t *Tracee) deriveEvents(ctx context.Context, in <-chan *trace.Event) (<-ch
 				}
 
 				for _, derivative := range derivatives {
+					// derived events also might have filters
+					if !t.shouldProcessEvent(derivative) {
+						continue
+					}
 					out <- &derivative
 				}
 


NDStrahilevitz (commented):
This fix by itself will work but it's not necessarily optimal:
1. On the context filtering side, we could always pre-filter this by applying the context filter on the base event. There is an edge case here though if the user wants the base event fully but filter on the derived.
2. We may be able to pre-predict argument filtering through the base event, but this is more complicated than point 1.

I think we may want to include the first optimization at least for now, WDYT?

rafaeldtinoco:
> * On the context filtering side, we could always pre-filter this by applying the context filter on the base event. There is an edge case here though if the user wants the base event fully but filters on the derived.

But my base event has packet payload, I would have to do the same processing (packet header data) before and after derivation if I wanted to filter the net arguments, no? I mean, if you take a look at my events, they all have a "Base" event and the only argument is the packet data.

I want to include a "net packet filter type" as well, so the user would be able o filter based on the packet header fields.

I also thought this step could be done in an earlier stage but thought that for this release this was enough (and maybe then we can think of something more optimal).

WDYT?

NDStrahilevitz:
> > * On the context filtering side, we could always pre-filter this by applying the context filter on the base event. There is an edge case here though if the user wants the base event fully but filters on the derived.
> 
> But my base event has packet payload, I would have to do the same processing (packet header data) before and after derivation if I wanted to filter the net arguments, no? I mean, if you take a look at my events, they all have a "Base" event and the only argument is the packet data.

Context filtering doesn't refer to the packets, since those are arguments. I meant stuff like `dnsevent.context.processName` which we do in the context filter.

rafaeldtinoco:
> > > * On the context filtering side, we could always pre-filter this by applying the context filter on the base event. There is an edge case here though if the user wants the base event fully but filters on the derived.
> > 
> > 
> > But my base event has packet payload, I would have to do the same processing (packet header data) before and after derivation if I wanted to filter the net arguments, no? I mean, if you take a look at my events, they all have a "Base" event and the only argument is the packet data.
> 
> Context filtering doesn't refer to the packets, since those are arguments. I meant stuff like `dnsevent.context.processName` which we do in the context filter.

I think its better for me to create an issue for that optimization instead of adding code (and possibly bugs) to the pipeline right now (1 or 2 days before the release). For now, I'm pretty sure the code I added works for derived events, so its a "fix" for now.

WDYT ?

NDStrahilevitz:
> > * On the context filtering side, we could always pre-filter this by applying the context filter on the base event. There is an edge case here though if the user wants the base event fully but filter on the derived.
> 
> the base events are internal so filters won't apply, I believe you want to pick the context filters from the event (derived) given to the cmdline and apply to the base events earlier in the pipeline, is that correct ?

The base events are internal so not up for emission, but they will be submitted and exist in the pipeline at the decode stage.
As you said, I would like to apply the context filters on the base events at that stage.
I'm ok with adding that issue for later, just open link it and I will approve.

NDStrahilevitz (approved):

rafaeldtinoco:
> I'm ok with adding that issue for later, just open link it and I will approve.

https://github.com/aquasecurity/tracee/issues/2468

Done! Thanks for the review.

Issue #2481, opened 2022-12-13
Add print_mem_dump event


AsafEitani:
This feature is about adding an event to dump the content of memory addresses or symbols.

AsafEitani:
#2483

PR #2483, opened 2022-12-13
events: add print_mem_dump event

AsafEitani:
The event is used to dump the content of memory addresses or symbols.
The symbol or address to be dumped is provided via the event filter:
provide it via `-t print_mem_dump.args.address=<hex address>`, `-t print_mem_dump.args.symbol_name=<owner>:<symbol>` or `-t print_mem_dump.args.symbol_name=<symbol>` if specifying a system owned symbol
## Initial Checklist

## Description (git log)

commit 7685dd1ea6afb5c52ef7191eb7ad71b241301dd1 (HEAD -> mem_dump, origin/mem_dump)
Date:   Sun Dec 4 17:59:45 2022 +0200

    events: add print_mem_dump event
    
    The event is used to dump the content of memory addresses or symbols.

Fixes: #2481

## Type of change

## How Has This Been Tested?
dumping both address and symbols based values from the memory and viewing as it printed after kernel module loading.

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -106,6 +106,7 @@ int KERNEL_VERSION SEC("version") = LINUX_VERSION_CODE;
 #define MAX_KSYM_NAME_SIZE  64
 #define UPROBE_MAGIC_NUMBER 20220829
 #define ARGS_BUF_SIZE       32000
+#define MEM_DUMP_SIZE       100
 // clang-format on
 
 // helper macros for branch prediction
@@ -470,6 +471,7 @@ enum event_id_e
     NET_PACKET_ICMPV6,
     NET_PACKET_DNS,
     DO_MMAP,
+    PRINT_MEM_DUMP,
     MAX_EVENT_ID,
 };
 
@@ -4241,6 +4243,36 @@ int uprobe_seq_ops_trigger(struct pt_regs *ctx)
     save_to_submit_buf(p.event, (void *) &caller_ctx_id, sizeof(uint64_t), 1);
     events_perf_submit(&p, PRINT_NET_SEQ_OPS, 0);
     return 0;
+}
+
+SEC("uprobe/trigger_mem_dump_event")
+int uprobe_mem_dump_trigger(struct pt_regs *ctx)
+{
+    u64 address = 0;
+    u32 trigger_pid = bpf_get_current_pid_tgid() >> 32;
+
+#if defined(bpf_target_x86)
+    address = ctx->bx; // 1nd arg
+    char arch[] = "x86_64";
+#elif defined(bpf_target_arm64)
+    address = ctx->user_regs.regs[1]; // 1nd arg
+    char arch[] = "arm64";
+#else
+    return 0;
+#endif
+
+    event_data_t data = {};
+    if (!init_event_data(&data, ctx))
+        return 0;
+
+    // uprobe was triggered from other tracee instance
+    if (data.config->tracee_pid != trigger_pid)
+        return 0;
+
+    save_to_submit_buf(&data, (void *) &address, sizeof(void *), 0);
+    save_bytes_to_buf(&data, (void *) address, MEM_DUMP_SIZE, 1);
+    save_str_to_buf(&data, (void *) arch, 2);
+    return events_perf_submit(&data, PRINT_MEM_DUMP, 0);
 }
 
 static __always_inline struct trace_kprobe *get_trace_kprobe_from_trace_probe(void *tracep)
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -4274,13 +4274,13 @@ int uprobe_mem_dump_trigger(struct pt_regs *ctx)
     if (size <= 0)
         return 0;
 
-    int ret = save_bytes_to_buf(&data, (void *) address, size & MAX_MEM_DUMP_SIZE, 1);
+    int ret = save_bytes_to_buf(&data, (void *) address, size & MAX_MEM_DUMP_SIZE, 0);
     // return in case of failed pointer read
     if (ret == 0) {
         tracee_log(ctx, BPF_LOG_LVL_ERROR, BPF_LOG_ID_MEM_READ, ret);
         return 0;
     }
-    save_to_submit_buf(&data, (void *) &address, sizeof(void *), 0);
+    save_to_submit_buf(&data, (void *) &address, sizeof(void *), 1);
     save_to_submit_buf(&data, &size, sizeof(u64), 2);
 
     return events_perf_submit(&data, PRINT_MEM_DUMP, 0);
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -106,7 +106,7 @@ int KERNEL_VERSION SEC("version") = LINUX_VERSION_CODE;
 #define MAX_KSYM_NAME_SIZE  64
 #define UPROBE_MAGIC_NUMBER 20220829
 #define ARGS_BUF_SIZE       32000
-#define MEM_DUMP_SIZE       127
+#define MAX_MEM_DUMP_SIZE   127
 // clang-format on
 
 // helper macros for branch prediction
@@ -4251,39 +4251,43 @@ int uprobe_mem_dump_trigger(struct pt_regs *ctx)
 {
     u64 address = 0;
     u64 size = 0;
+    u64 caller_ctx_id = 0;
     u32 trigger_pid = bpf_get_current_pid_tgid() >> 32;
 
 #if defined(bpf_target_x86)
-    address = ctx->bx; // 1st arg
-    size = ctx->cx;    // 2nd arg
+    address = ctx->bx;       // 1st arg
+    size = ctx->cx;          // 2nd arg
+    caller_ctx_id = ctx->di; // 3rd arg
 #elif defined(bpf_target_arm64)
-    address = ctx->user_regs.regs[1]; // 1st arg
-    size = ctx->user_regs.regs[2];    // 2nd arg
+    address = ctx->user_regs.regs[1];       // 1st arg
+    size = ctx->user_regs.regs[2];          // 2nd arg
+    caller_ctx_id = ctx->user_regs.regs[3]; // 3rd arg
 #else
     return 0;
 #endif
 
-    event_data_t data = {};
-    if (!init_event_data(&data, ctx))
+    program_data_t p = {};
+    if (!init_program_data(&p, ctx))
         return 0;
 
     // uprobe was triggered from other tracee instance
-    if (data.config->tracee_pid != trigger_pid)
+    if (p.config->tracee_pid != trigger_pid)
         return 0;
 
     if (size <= 0)
         return 0;
 
-    int ret = save_bytes_to_buf(&data, (void *) address, size & MAX_MEM_DUMP_SIZE, 0);
+    int ret = save_bytes_to_buf(p.event, (void *) address, size & MAX_MEM_DUMP_SIZE, 0);
     // return in case of failed pointer read
     if (ret == 0) {
         tracee_log(ctx, BPF_LOG_LVL_ERROR, BPF_LOG_ID_MEM_READ, ret);
         return 0;
     }
-    save_to_submit_buf(&data, (void *) &address, sizeof(void *), 1);
-    save_to_submit_buf(&data, &size, sizeof(u64), 2);
+    save_to_submit_buf(p.event, (void *) &address, sizeof(void *), 1);
+    save_to_submit_buf(p.event, &size, sizeof(u64), 2);
+    save_to_submit_buf(p.event, &caller_ctx_id, sizeof(u64), 3);
 
-    return events_perf_submit(&data, PRINT_MEM_DUMP, 0);
+    return events_perf_submit(&p, PRINT_MEM_DUMP, 0);
 }
 
 static __always_inline struct trace_kprobe *get_trace_kprobe_from_trace_probe(void *tracep)
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -106,7 +106,7 @@ int KERNEL_VERSION SEC("version") = LINUX_VERSION_CODE;
 #define MAX_KSYM_NAME_SIZE  64
 #define UPROBE_MAGIC_NUMBER 20220829
 #define ARGS_BUF_SIZE       32000
-#define MEM_DUMP_SIZE       100
+#define MEM_DUMP_SIZE       127
 // clang-format on
 
 // helper macros for branch prediction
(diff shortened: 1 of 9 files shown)


pkg/ebpf/tracee.go
```diff
@@ -1626,6 +1630,65 @@ func (t *Tracee) triggerSeqOpsIntegrityCheckCall(
...
+// that prints the seq ops pointers
+func (t *Tracee) triggerMemDump() error {
+	_, ok := t.events[events.PrintMemDump]
+	if !ok {
+		return nil
+	}
+
+	printMemDumpFilters := t.config.Filter.ArgFilter.GetEventFilters(events.PrintMemDump)
+	if printMemDumpFilters == nil {
+		return fmt.Errorf("no address or symbols were provided to print_mem_dump event. " +
```
NDStrahilevitz, line 1643: @rafaeldtinoco looks like this event will also be configured through filters.
In the context of this PR we need to document it, and exclude it from the filtering at the derivation stage later.

NDStrahilevitz: I also think this is more support for separating event configuration and filtering

rafaeldtinoco: Okay, so, for this PR, it should be rebased to include the same skip as commit 12460c13 AND I do agree with Nadav, we should create an interface for providing config options to signatures (instead of using the filters themselves).

AlonZivony: I suggest we open an issue on this and create this interface in a new PR.
For now @AsafEitani you should use the skip mechanism that @rafaeldtinoco did for previous similar events.

pkg/ebpf/tracee.go
```diff
@@ -1626,6 +1630,65 @@ func (t *Tracee) triggerSeqOpsIntegrityCheckCall(
...
+			}
+			t.triggerMemDumpCall(symbol.Address)
+		}
+	}
+
+	return nil
+}
+
+//go:noinline
+func (t *Tracee) triggerMemDumpCall(address uint64) error {
```
AlonZivony, line 1688: This is the function that trigger the uprobe, so you should document it in here instead in `triggerMemDump`.

AlonZivony (changes requested):
Overall look good.
You need to rebase on main, so the change in the `parse` package is not necessary because we use generics now there.
I think we can wrap it quickly and enter it, changing the infrastructure in the future to be better for configured events and triggered events.

pkg/ebpf/tracee.go
AsafEitani: Done

pkg/ebpf/tracee.go
AsafEitani: But all other uprobe functions are documented at the function that calls them

pkg/events/events.go
```diff
@@ -6401,5 +6402,28 @@ var Definitions = eventDefinitions{
...
+			},
+			Probes: []probeDependency{
+				{Handle: probes.PrintMemDump, Required: true},
+			},
+			Dependencies: dependencies{
+				Events: []eventDependency{
+					{EventID: DoInitModule},
+				},
+				KSymbols: []kSymbolDependency{
+					{Symbol: "sys_call_table", Required: true},
```
NDStrahilevitz, line 6424: Is this needed? I don't see the symbol used anywhere else in the PR.

AsafEitani: We need the kernel symbols to be able to translate the provided symbol to address.
This is how I managed to tell tracee-ebpf that there is a dependancy.
Now that I think of it there will be a problem with the lazy load...
@NDStrahilevitz what do you recommend?

NDStrahilevitz: Oh I see, this forces tracee to construct the kernelSymbols object.
Lazy version shouldn't work different actually, but we need some way to declare a dependency for the event to require symbol access in general.
Maybe `KSymbols` should be a struct of:
```
     RequireDynamicSymbols bool
     StaticSymbols                   []kSymbolDependency
```

Then for events that only need dynamic access we mark the bool. If we need some preloaded static symbol, we add it to the nested array.

AsafEitani: @yanivagman What are your thoughts? I'm fine with it if you say it's ok - I'll implement

AlonZivony: It seems good to me too

yanivagman: Dependencies section of an Event should well define the needs of the event to work correctly. If it is missing a way to define that symbols are required in general, then I think the approach @NDStrahilevitz suggested is good

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -105,6 +105,7 @@ int KERNEL_VERSION SEC("version") = LINUX_VERSION_CODE;
 #define NET_SEQ_OPS_TYPES   6         // print_net_seq_ops: argument size - TODO: replace with uprobe argument
 #define MAX_KSYM_NAME_SIZE  64
 #define UPROBE_MAGIC_NUMBER 20220829
+#define MEM_DUMP_SIZE       100
```
yanivagman, line 108: Why 100? Who says that you will be able to read that amount of bytes from a given address?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4219,6 +4221,36 @@ int uprobe_seq_ops_trigger(struct pt_regs *ctx)
...
 }
 
+SEC("uprobe/trigger_mem_dump_event")
+int uprobe_mem_dump_trigger(struct pt_regs *ctx)
+{
+    u64 address = 0;
+    u32 trigger_pid = bpf_get_current_pid_tgid() >> 32;
+
+#if defined(bpf_target_x86)
+    address = ctx->bx; // 1nd arg
```
yanivagman, line 4231: nit: 1st

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4219,6 +4221,36 @@ int uprobe_seq_ops_trigger(struct pt_regs *ctx)
...
+    char arch[] = "x86_64";
+#elif defined(bpf_target_arm64)
+    address = ctx->user_regs.regs[1]; // 1nd arg
```
yanivagman, line 4234: same

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4219,6 +4221,36 @@ int uprobe_seq_ops_trigger(struct pt_regs *ctx)
...
 
+SEC("uprobe/trigger_mem_dump_event")
+int uprobe_mem_dump_trigger(struct pt_regs *ctx)
+{
+    u64 address = 0;
+    u32 trigger_pid = bpf_get_current_pid_tgid() >> 32;
+
+#if defined(bpf_target_x86)
+    address = ctx->bx; // 1nd arg
+    char arch[] = "x86_64";
```
yanivagman, line 4232: Why is arch required? It is something that doesn't change on a given machine and you can get it in other ways

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4219,6 +4221,36 @@ int uprobe_seq_ops_trigger(struct pt_regs *ctx)
...
+    event_data_t data = {};
+    if (!init_event_data(&data, ctx))
+        return 0;
+
+    // uprobe was triggered from other tracee instance
+    if (data.config->tracee_pid != trigger_pid)
+        return 0;
+
+    save_to_submit_buf(&data, (void *) &address, sizeof(void *), 0);
+    save_bytes_to_buf(&data, (void *) address, MEM_DUMP_SIZE, 1);
```
yanivagman, line 4249: Better if you get MEM_DUMP_SIZE as an argument and AND it with some max allowed size.
Also, since the whole point of the event is to read these bytes and this operation can fail (e.g. memory not mapped), I think you should check for error here and return if failed.

AsafEitani:
Fixed #2546

pkg/ebpf/tracee.go
```diff
@@ -1404,6 +1404,10 @@ func (t *Tracee) Run(ctx gocontext.Context) error {
 	t.invokeInitEvents()
 	t.triggerSyscallsIntegrityCheck(trace.Event{})
 	t.triggerSeqOpsIntegrityCheck(trace.Event{})
+	err := t.triggerMemDump()
+	if err != nil {
+		return err
```
yanivagman, line 1409: Should this fail the whole run?

pkg/ebpf/tracee.go
```diff
@@ -1639,3 +1643,62 @@ func (t *Tracee) triggerSeqOpsIntegrityCheckCall(
 	seqOpsStruct [len(derive.NetSeqOps)]uint64) error {
 	return nil
 }
+
+// triggerMemDump is used by a Uprobe to trigger an eBPF program
+// that prints the seq ops pointers
```
yanivagman, line 1648: seq ops pointers?

pkg/ebpf/tracee.go
```diff
@@ -1639,3 +1643,62 @@ func (t *Tracee) triggerSeqOpsIntegrityCheckCall(
...
+func (t *Tracee) triggerMemDump() error {
+	_, ok := t.events[events.PrintMemDump]
+	if !ok {
```
yanivagman, line 1651: nit - you can do instead
`if _, ok := t.events[events.PrintMemDump]; !ok {`
which is shorter. This is used when the return variable (ok in this case) is not required after the check.

I saw this pattern repeated in several places in this PR, so if you want you can fix as well

pkg/filters/args.go
```diff
@@ -42,7 +42,8 @@ func (filter *ArgFilter) Filter(eventID events.ID, args []trace.Argument) bool {
 			}
 		}
 		if !found {
-			return false // always filter if argument does not exist
+			// TODO: remove once events arguments are introduced
+			return eventID == events.PrintMemDump // filter if argument does not exist
```
yanivagman, line 46: What does this do?
If it is a workaround for some bug, please add a comment that explains it.

yanivagman (changes requested):

pkg/ebpf/c/tracee.bpf.c
AsafEitani: how?

pkg/ebpf/tracee.go
AsafEitani: yes as this means that the symbol provided doesn't exist for instance.

pkg/ebpf/c/tracee.bpf.c
yanivagman: One option - use uname. Many others exist.

pkg/ebpf/tracee.go
yanivagman:
```
	t.invokeInitEvents()
	t.triggerSyscallsIntegrityCheck(trace.Event{})
	t.triggerSeqOpsIntegrityCheck(trace.Event{})
```
All these don't fail the whole run. Why the inconsistency?
Since this event is triggered once (and not attached to any kprobe/tracepoint) - I think logging an error would be enough.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4219,6 +4222,46 @@ int uprobe_seq_ops_trigger(struct pt_regs *ctx)
...
+    if (size <= 0)
+        return 0;
+
+    int ret = save_bytes_to_buf(&data, (void *) address, size & MAX_MEM_DUMP_SIZE, 1);
+    // return in case of failed pointer read
+    if (ret == 0) {
+        tracee_log(ctx, BPF_LOG_LVL_ERROR, BPF_LOG_ID_MEM_READ, ret);
+        return 0;
+    }
+    save_to_submit_buf(&data, (void *) &address, sizeof(void *), 0);
```
yanivagman, line 4259: Putting index 0 after index 1 will reverse the order of these args in the table printer.
Better put it in the correct order (or change the order in the event definition)

NDStrahilevitz:
Needs a rebase since we've changed the event processing structure.

pkg/ebpf/tracee.go
AsafEitani: So I fixed it for now - warning if there was an issue.
In the future when we implement argument reception from the user we will fail if the one of the arguments wasn't what we expected (for instance - symbol doesn't exists)
This is the issue: #2559

pkg/ebpf/events_processor.go
```diff
@@ -440,5 +453,23 @@ func (t *Tracee) processTriggeredEvent(event *trace.Event) error {
 	// will be moved back as such we apply the value internally and not
 	// through a referene switch
 	(*event) = withInvokingContext
+	if event.EventID == int(events.PrintMemDump) {
```
NDStrahilevitz, line 456: suggestion: You should do this in a separate processor function. You can now register multiple processors for one event :)

NDStrahilevitz (approved):
Left a small rewrite suggestion (should also be a good demonstration of the new processing capabilities).
+1 for this anyway, but I assume we're waiting for the big eBPF PRs first iirc.

pkg/events/events.go
```diff
@@ -21,6 +21,11 @@ type probeDependency struct {
 	Required bool // should tracee fail if probe fails to attach
 }
 
+type kSymbolsDependencies struct {
+	RequireDynamicSymbols bool
+	StaticSymbols         []kSymbolDependency
```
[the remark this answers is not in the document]
NDStrahilevitz: They're not, for example `hooked_proc_fops` needs both.
And if we go by the nil suggestion then all events will have a dynamic symbols requirement, which isn't true. As such tracee's memory footprint will include the ksymbols even if no ksymbols related events are traced.

pkg/filters/args.go
```diff
@@ -42,7 +42,11 @@ func (filter *ArgFilter) Filter(eventID events.ID, args []trace.Argument) bool {
 			}
 		}
 		if !found {
-			return false // always filter if argument does not exist
+			// TODO: remove once events arguments are introduced
+			// filter if argument does not exist unless the event is originating
+			// from the kernel and uses filters as arguments
+			// issue #2546
```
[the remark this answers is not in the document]
NDStrahilevitz: +1
We've already discussed multiple times in various forums that we need a separate interface for event configuration.
This is a byproduct of us not fixing that yet.

pkg/events/events.go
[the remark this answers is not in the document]
yanivagman: > They're not, for example hooked_proc_fops needs both.

RequireDynamicSymbols is set to true only in the new mem_dump event, and not for hooked_proc_fops.

> And if we go by the nil suggestion then all events will have a dynamic symbols requirement, which isn't true. As such tracee's memory footprint will include the ksymbols even if no ksymbols related events are traced.

I suggested differentiating between three options:
1. No dependencis -> nil (then we should NOT ask for dynamic symbols)
2. Static dependencies -> list is not empty, ask only for those symbol dependencies
3. Dynamic dependencies -> list is empty -> set the dependency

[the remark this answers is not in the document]
NDStrahilevitz: > RequireDynamicSymbols is set to true only in the new mem_dump event, and not for hooked_proc_fops.
It should have both but it's practically not needed because by setting the static symbols the kernel symbols object is loaded and dynamic querying is enabled.

> I suggested differentiating between three options:
You're right, i've misread what you wrote then, that can work as well.

pkg/events/events.go
```diff
@@ -11,7 +11,7 @@ import (
 
 type dependencies struct {
 	Events       []eventDependency // Events required to be loaded and/or submitted for the event to happen
-	KSymbols     []kSymbolDependency
+	KSymbols     *[]kSymbolDependency
```
NDStrahilevitz, line 14: This needs documentation for the difference between nil empty and values.

yanivagman (approved):
LGTM

Issue #2487, opened 2022-12-14
[BUG] arg filters don't work when argument is missing from event


rafaeldtinoco:
## Prerequisites

- [ ] This affects latest released version.
- [x] This affects current development tree (origin/HEAD).
- [ ] There isn't an issue describing the bug.

Select one OR another:

- [x] I'm going to create a PR to solve this (assign to yourself).
- [ ] Someone else should solve this.

## Bug description

When running tracee like this:

```
sudo ./dist/tracee-ebpf --output format:json --output option:parse-arguments --trace event=openat --trace openat.args.pathname=/etc/passwd 
```

you will realize that the events that don't contain `pathname` argument will be displayed (when they should not):

```
not found: pathname
not found: pathname
not found: pathname
not found: pathname
not found: pathname
not found: pathname
not found: pathname
not found: pathname
not found: pathname
not found: pathname
not found: pathname
not found: pathname
EH AQUI: /proc/cpuinfo
EH AQUI: /proc/self/mountinfo
{"timestamp":1670988030860120046,"threadStartTime":288978540525770,"processorId":16,"processId":2022681,"cgroupId":18637,"threadId":2022681,"parentProcessId":1791387,"hostProcessId":2022681,"hostThreadId":2022681,"hostParentProcessId":1731629,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"runc","hostName":"fujitsu","containerId":"","containerImage":"","containerName":"","podName":"","podNamespace":"","podUID":"","eventId":"257","eventName":"openat","argsNum":3,"returnValue":-2,"stackAddresses":null,"contextFlags":{"containerStarted":false},"args":[{"name":"dirfd","type":"int","value":-100},{"name":"flags","type":"string","value":"O_RDONLY|O_CLOEXEC"},{"name":"mode","type":"mode_t","value":0}]}
[...]

Issue #2486, opened 2022-12-14
[BUG] symbols loaded args is a char ** and ArgFilter.Filter() does not consider this type


rafaeldtinoco:
## Prerequisites

- [ ] This affects latest released version.
- [x] This affects current development tree (origin/HEAD).
- [ ] There isn't an issue describing the bug.

Select one OR another:

- [x] I'm going to create a PR to solve this (assign to yourself).
- [ ] Someone else should solve this.

## Bug description

symbols loaded event arg is a char ** and ArgFilter.Filter() does not consider this type:

```
sudo ./dist/tracee-ebpf --output option:parse-arguments --trace event=symbols_loaded --trace symbols_loaded.args.symbols=open,fopen --trace symbols_loaded.args.library_path=libc

argName: symbols
argVal type: []string
contents: [open fopen]

argName: library_path
argVal type: string
contents: /usr/lib/x86_64-linux-gnu/libc.so.6
```

rafaeldtinoco:
To check ideas from @NDStrahilevitz at https://github.com/aquasecurity/tracee/pull/2488#discussion_r1048506587. He already had that in mind and was working some possible solutions.

PR #2488, opened 2022-12-14
Arg filter fixes

rafaeldtinoco:
## Initial Checklist

## Description (git log)

commit 4c047682 (HEAD -> arg-filter-fixes, rafaeldtinoco/arg-filter-fixes, main)
Date:   Wed Dec 14 03:45:45 2022

    filters: if arg is missing do not process it
    
    When filtering an event for an argument that might not exist (because
    the eBPF code might not send it), the events missing the argument
    being filtered were being sent.
    
    One exampe is when openat filters for openat.args.pathname of
    /etc/passwd. Some openat events missing pathname might be sent out
    through printer.
    
    Do not process the event when an argument being filtered for does not
    exist in the event.
    
Fixes: #2487

commit cbc79bba
Date:   Wed Dec 14 03:44:14 2022

    filters: add []string type to args filters
    
    SymbolsLoaded event has a []string argument in tracee.Arguments and
    if any of those values has a filter, it must be considered.
    
Fixes: #2486

## Type of change

## How Has This Been Tested?

Tested by executing:

```
sudo ./dist/tracee-ebpf --output format:json --output option:parse-arguments --trace event=openat --trace openat.args.pathname=/etc/passwd 
```

and

```
sudo ./dist/tracee-ebpf --output option:parse-arguments --trace event=symbols_loaded --trace symbols_loaded.args.symbols=readdir,readdir64,open,link,rmdir,fopen,open64,fopen64 --trace symbols_loaded.args.library_path="libc"
```

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/filters/args.go
+++ b/pkg/filters/args.go
@@ -42,7 +42,7 @@ func (filter *ArgFilter) Filter(eventID events.ID, args []trace.Argument) bool {
 			}
 		}
 		if !found {
-			return true
+			return false // always filter if argument does not exist
 		}
 		// TODO: use type assertion instead of string conversion
 		if argName != "syscall" {
--- a/pkg/ebpf/events_pipeline.go
+++ b/pkg/ebpf/events_pipeline.go
@@ -312,9 +312,17 @@ func (t *Tracee) deriveEvents(ctx context.Context, in <-chan *trace.Event) (<-ch
 				}
 
 				for _, derivative := range derivatives {
-					// derived events also might have filters
-					if !t.shouldProcessEvent(derivative) {
-						continue
+					// Skip events that don't work well with filtering due
+					// to missing types being handled and similar reasons.
+					// https://github.com/aquasecurity/tracee/issues/2486
+					switch events.ID(derivative.EventID) {
+					case events.SymbolsLoaded:
+					case events.SharedObjectLoaded:
+					default:
+						// Derived events might need filtering as well
+						if !t.shouldProcessEvent(derivative) {
+							continue
+						}
 					}
 					out <- &derivative
 				}


rafaeldtinoco:
I still need to test e2e tests with this fix, but I think it will fix it (problem was symbols loaded event being totally filtered because ShouldProcessEvent() is now being called from the derivative event pipeline stage).

pkg/filters/args.go
```diff
@@ -44,12 +44,28 @@ func (filter *ArgFilter) Filter(eventID events.ID, args []trace.Argument) bool {
...
+					// check all string filters for argName (e.g.: SymbolsLoaded)
+					if !filter.Filter(argV) {
+						return false
+					}
+				}
+				return true
+			default:
+				// TODO: Expand other tracee.Argument types here
+				//       ArgFilter should not be used by unimplemented types
+				return true // never filter unknown types
```
rafaeldtinoco, line 66: My understanding is that we should ALWAYS filter unknown types, but that brakes e2e tests (meaning that those tests are relying in arg types we currently don't filter, or might be filtering wrongly).

NDStrahilevitz: This will break most behavior right now, let'skeep as is

pkg/filters/args.go
```diff
@@ -42,14 +42,30 @@ func (filter *ArgFilter) Filter(eventID events.ID, args []trace.Argument) bool {
 			}
 		}
 		if !found {
-			return true
+			return false // always filter if argument does not exist
```
NDStrahilevitz, line 45: Can we test only witrh this change and see that nothing breaks?

pkg/filters/args.go
```diff
@@ -42,14 +42,30 @@ func (filter *ArgFilter) Filter(eventID events.ID, args []trace.Argument) bool {
...
-		}
-		res := filter.Filter(argVal)
-		if !res {
+			switch argValue := argVal.(type) {
+			case string:
+				if !filter.Filter(argValue) {
+					return false
+				}
+				return true
+			case []string:
```
NDStrahilevitz, line 55: We can implement this case much more efficiently with a FilterObject of its own. Internally just a map and a loop doing hash checks.

NDStrahilevitz (changes requested):

pkg/filters/args.go
rafaeldtinoco: Sure, let me check!

pkg/filters/args.go
rafaeldtinoco: Definitely. I was just exploring what was going on and checking what would brake what.

NDStrahilevitz: If we do it as a generic object we could even cover other primitive array type (structs are more complex as can be seen in the SockAddr PR).
The problem is registering the correct filter at parse time since we have so many type names for arguments (I was working on resolving it but it's really hard to resolve to the decode type we use in `bufferdecoder`, since we have differing types for decoding and final representation)

pkg/filters/args.go
rafaeldtinoco: Well I'm only mitigating the issue by skipping object related events from the derivation filtering (for now). Will keep issue https://github.com/aquasecurity/tracee/issues/2486 opened for next milestone.

pkg/filters/args.go
rafaeldtinoco: Well only this change still brakes e2e because of symbols loaded event. Adding a mitigation for that, plus this change, is the way to go for this release (IMO).

pkg/filters/args.go
rafaeldtinoco: Keeping https://github.com/aquasecurity/tracee/issues/2486 opened and linking this comment there so we can discuss for the next release. For now I'll just skip those events we know for sure are broken. E2E tests are passing with the latest push, check if you agree pls. Then we can have at least a mitigation for this release.

yanivagman (approved):
LGTM

PR #2493, opened 2022-12-14
[FEAT] builder: add custom-rules arg opt to entrypoint.sh

geyslan:
## Description (git log)

commit ca346d16e273f8cdc53e24034a02718c2d8d82ab

    builder: add custom-rules arg opt to entrypoint.sh
    
    For docker must be used like
    docker run -v /your/custom/rules/dir:/custom-rules:ro ... custom-rules
    
    Context: #2402

Fixes: #2402

## Type of change


builder/entrypoint.sh
```diff
@@ -113,6 +118,17 @@ fi
 if [ "${1}" == "trace" ]; then
     TRACEE_EBPF_ONLY=1
     shift
+elif [ "${1}" == "custom-rules" ]; then
+    TRACEE_RULES="/custom-rules"
```
rafaeldtinoco, line 122: don't u need to call "shift" to shift the arguments, if the user also wants to feed tracee-rules with arguments ?

geyslan: There's at 131.

josedonizetti (approved):
awesome, it works, tested on minkube.

PR #2507, opened 2022-12-17
ebpf: remove interpreter map

yanivagman:
interpreter_map is used to save interpreter file info per process. Now that we have the proc_info_map, we can save it there and remove the interpreter map.

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -782,10 +782,21 @@ typedef struct task_info {
     u8 container_state;   // the state of the container the task resides in
 } task_info_t;
 
+typedef struct file_info {
+    union {
+        char pathname[MAX_CACHED_PATH_SIZE];
+        char *pathname_p;
+    };
+    dev_t device;
+    unsigned long inode;
+    u64 ctime;
+} file_info_t;
+
 typedef struct proc_info {
     bool new_proc; // set if this process was started after tracee. Used with new_pid filter
     bool follow;   // set if this process was traced before. Used with the follow filter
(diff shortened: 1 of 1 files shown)


pkg/ebpf/c/tracee.bpf.c
```diff
@@ -6762,20 +6753,29 @@ int BPF_KPROBE(trace_load_elf_phdrs)
     if (!should_trace((&data)))
         return 0;
 
-    file_info_t elf = {};
+    proc_info_t *proc_info = bpf_map_lookup_elem(&proc_info_map, &data.context.task.host_pid);
+    if (proc_info == NULL) {
```
NDStrahilevitz, line 6757: we can use `unlikely` here considering the later comment.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3828,9 +3821,7 @@ int tracepoint__sched__sched_process_exit(struct bpf_raw_tracepoint_args *ctx)
     // evaluate should_trace before removing this pid from the maps
     bool traced = should_trace(&data);
 
-    // Remove this pid from all maps
     bpf_map_delete_elem(&task_info_map, &data.context.task.host_tid);
```
NDStrahilevitz, line 3824: Just making sure, this stays here and shouldn't move to the `sched_process_free` program?

NDStrahilevitz (approved):
LGTM, needs a rebase anyway so I left a suggestion and a question.

pkg/ebpf/c/tracee.bpf.c
yanivagman: Yes, this should stay here since it is tid and not pid

PR #2509, opened 2022-12-19
network: fix net_packet_dns_request and response

rafaeldtinoco:
## Description (git log)

Fixes:

1. need to memset 0 the event struct in stack not to have garbage in comm

2. need to pass the object instead of a pointer as argument, the same as dns_request and dns_response events had.

3. only send dns_request event if there are no responses.

## Type of change

----

## How Has This Been Tested?

IN A

image

image

IN MX

(the old event dns_request has trashed command names, I solved that in the new one as well)

image

(looks like the old dns_response event did not pick the correct MX response, while the new one did)

image

IN NS

image

image

----

PS: For the "IN A" tests I have emulated 2 things:

1. pktlen from the new events is always 14 bytes shorter (that is because the new events only consider L3 layer and beyond, being agnostic to L2 type). 

image

2. the interface name was used as "docker0"
 
just to check if old and new events were the same.

----

@roikol FYI, could you please check this PR with the same tests you did and see if there are any differences from old and new events for your signatures ? Thank you!

@yanivagman if this fixes Roi's issues, then we should consider a new minor release for this fix (for CNDRs existing DNS signatures to work, if that is what they need/want).

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -7343,6 +7343,7 @@ static __always_inline u64 sizeof_net_event_context_t(void)
 static __always_inline void set_net_task_context(event_data_t *data, net_task_context_t *netctx)
 {
     netctx->task = data->task;
+    __builtin_memset(&netctx->taskctx, 0, sizeof(task_context_t));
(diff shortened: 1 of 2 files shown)


roikol:
@rafaeldtinoco 

looks good! also tested it locally and it works as expected

roikol (approved):

Issue #2355, opened 2022-11-12
Everything is an event


josedonizetti:
## Feature description

This epic represents the new tracee experience, where we will have a single binary, every thing will be an event. The idea is to implement this feature incrementally to avoid a HUGE pr complicating the review process. 

Please add as an item to the list for any bug, refactor, change found for the new tracee binary.

## TODO

- [x] fix filtering (eg: `tracee --trace set=rules`)
- [x] fix errors when loading same signature in more than one scope
- [x] add shortcuts (eg: `tracee --trace e=fileless_execution`)
- [x] rules events not aligned
- [x] add signature events metada (#2541)
- [x] signature events should be multiscoped ( ./dist/tracee --trace 1:event=anti_debugging --trace 2:event=anti_debugging), should print twice, once per scope
- [x] finding.Data should an arguments of signature events
- [x] multiple printers
- [x] make tracee binary the default for tracee image
- [x] make signature events load by default
- [x] add webhook support
- [x] fix k8s yamls to use tracee binary
- [x] remove helm flag everythingIsAnEvent
- [x] make sure postee integration works
- [x] make sure falcosidekiq integration works
- [x] documentation

A few TODOs from the original PR that we agreed on implementing on the next release 0.11.0
- [x] https://github.com/aquasecurity/tracee/pull/2439#discussion_r1045165063
- [x] https://github.com/aquasecurity/tracee/pull/2439#discussion_r1045227985
- [x] https://github.com/aquasecurity/tracee/pull/2439#discussion_r1045164237
- [x] https://github.com/aquasecurity/tracee/pull/2439#discussion_r1045144380
- [x] https://github.com/aquasecurity/tracee/pull/2439#discussion_r1043547108

josedonizetti:
1 - extract tracee-rules logic
2 - add signatures as events
3 - add template as a printer option
4 - add webhook as a printer oprion

josedonizetti:
- add webhook support
- fix filtering
- add shortcuts (eg: `tracee --tracee e=fileless_execution`)
- refactor events printer
- rules events no aligned
-

itaysk:
another TODO here: 
if I create a "signature" (now just an event), I'm expecting to filter it just like every event, including argument filters. For example, I created a `file_write` signature based on `security_file_open`, and I wanted to filter `file_write.args.pathname=/something` which didn't work.

josedonizetti:
Problem 1:
---
[...]

PR #2511, opened 2022-12-19
Add colorizaion of rules events to the table printer

AsafEitani:
## Initial Checklist

## Description (git log)
Adds colorization to the table printer on rules events.

commit bec6f3670a4f34c9e41b2b686a548332721cf43e (HEAD -> color_table, origin/color_table)
Date:   Mon Dec 19 20:35:31 2022 +0200

    printer: add colorizaion of rules events

## Type of change

### DIFF
--- a/pkg/cmd/printer/printer.go
+++ b/pkg/cmd/printer/printer.go
@@ -10,6 +10,7 @@ import (
 	"text/template"
 	"time"
 
+	"github.com/aquasecurity/tracee/pkg/events"
 	"github.com/aquasecurity/tracee/pkg/logger"
 	"github.com/aquasecurity/tracee/pkg/metrics"
 	"github.com/aquasecurity/tracee/types/trace"
@@ -127,6 +128,9 @@ func (p tableEventPrinter) Print(event trace.Event) {
 	if p.relativeTS {
 		ut = ut.UTC()
 	}
+
+	isRule := event.EventID >= int(events.StartRulesID) && event.EventID < int(events.MaxRulesID)
+
 	timestamp := fmt.Sprintf("%02d:%02d:%02d:%06d", ut.Hour(), ut.Minute(), ut.Second(), ut.Nanosecond()/1000)
 
 	containerId := event.ContainerID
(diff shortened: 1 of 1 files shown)


pkg/cmd/printer/printer.go
```diff
@@ -127,6 +128,14 @@ func (p tableEventPrinter) Print(event trace.Event) {
 	if p.relativeTS {
 		ut = ut.UTC()
 	}
+
+	isRule := false
+	for _, v := range events.Definitions.Get(events.ID(event.EventID)).Sets {
+		if v == "rules" {
+			isRule = true
+		}
+	}
```
yanivagman, line 137: Instead, you can simply check if ID is between StartRulesID to MaxRulesID

rafaeldtinoco:
A change like this should be optional (set by a cmdline argument). There might be other tools parsing tracee output... having ANSI escape codes will disturb those. GNU tools usually have something like "--color=auto" or "--color=always" (or =no).

AsafEitani:
> A change like this should be optional (set by a cmdline argument). There might be other tools parsing tracee output... having ANSI escape codes will disturb those. GNU tools usually have something like "--color=auto" or "--color=always" (or =no).

Do you think that users parse the table printer? not using JSON?

rafaeldtinoco:
> Do you think that users parse the table printer? not using JSON?

TBH, I don't. But it is also a bad practice to colorize without explicitly requesting, or checking the TERMCAP support (imagine the user is running this in a vt100 terminal, or in any other terminal type that does not support colors).

josedonizetti:
@AsafEitani I like the idea of colors, and think it should be enabled by default. But also agree with @rafaeldtinoco and the user should be able to disable it. Can we go with the same solution `trivy` uses? The library `https://github.com/fatih/color` allows us to use color, and the user can set a env var to disable it, eg: `NO_COLOR=true` and unset it, if he wants color back. 

WDYT?

If you want to see an example on trivy -> https://github.com/aquasecurity/trivy/blob/main/pkg/report/table/table.go#L22

rafaeldtinoco (changes requested):
Per last comments.

AsafEitani:
Waiting for #2541 to implement the color by severity

rafaeldtinoco:
@AsafEitani, https://github.com/aquasecurity/tracee/issues/2355 (took care of #2541) is finished. Do you really think we should colorize the output ? @josedonizetti opinions ? Maybe to have another printer (now that we support multiple printers) for the colorized output and derived from the regular table printer ? Just thinking out loud.

Converting this to draft for now.

AsafEitani:
I don't have a strong opinion, what ever you choose.

rafaeldtinoco:
Lets close this for now then. We can re-discuss if we ever think there is a need for it.

Issue #2516, opened 2022-12-22
Expand binary filter to support mount ns


yanivagman:
Currently the binary filter can not differentiate between binaries that have the same path but reside in different mount namespaces. Having the mount namespace is required to identify a specific binary to filter by.

PR #2517, opened 2022-12-22
filters: expand binary filter to support mount ns

yanivagman:
Currently the binary filter can not differentiate between binaries that have the same path but reside in different mount namespaces. Having the mount namespace is required to identify a specific binary to filter by.

Enable the user to provide the mount namespace in addition to the binary path and keep support of providing a binary without specifying a mount namespace.

To filter by path, the following syntax is used::
    --trace binary=/some/binary/path

To filter by path and a specific mount namespace (e.g. 4026532448), use:
    --trace binary=4026532448:/some/binary/path

Specifying the host mount namespace can be done using the "host" alias:
    --trace binary=host:/some/binary/path

Fixes: #2516

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -793,10 +793,16 @@ typedef struct file_info {
     u64 ctime;
 } file_info_t;
 
+typedef struct binary {
+    u32 mnt_id;
+    char path[MAX_BIN_PATH_SIZE];
+} binary_t;
+
 typedef struct proc_info {
     bool new_proc; // set if this process was started after tracee. Used with new_pid filter
     bool follow;   // set if this process was traced before. Used with the follow filter
-    char binary[MAX_BIN_PATH_SIZE];
+    struct binary binary;
+    u32 binary_no_mnt; // used in binary lookup when we don't care about mount ns. always 0.
     file_info_t interpreter;
 } proc_info_t;
 
@@ -822,10 +828,6 @@ typedef struct path_filter {
 typedef struct string_filter {
     char str[MAX_STR_FILTER_SIZE];
 } string_filter_t;
-
-typedef struct binary_filter {
-    char str[MAX_BIN_PATH_SIZE];
-} binary_filter_t;
 
 typedef struct ksym_name {
     char str[MAX_KSYM_NAME_SIZE];
@@ -1025,7 +1027,7 @@ BPF_HASH(pid_ns_filter, u64, u32, 256);                            // filter eve
 BPF_HASH(uts_ns_filter, string_filter_t, u32, 256);                // filter events by uts namespace name
 BPF_HASH(comm_filter, string_filter_t, u32, 256);                  // filter events by command name
 BPF_HASH(cgroup_id_filter, u32, u32, 256);                         // filter events by cgroup id
-BPF_HASH(binary_filter, binary_filter_t, u32, 256);                // filter events by binary path
+BPF_HASH(binary_filter, binary_t, u32, 256);                       // filter events by binary path and mount namespace
 BPF_HASH(bin_args_map, u64, bin_args_t, 256);                      // persist args for send_bin funtion
 BPF_HASH(sys_32_to_64_map, u32, u32, 1024);                        // map 32bit to 64bit syscalls
 BPF_HASH(params_types_map, u32, u64, 1024);                        // encoded parameters types for event
@@ -1876,7 +1878,9 @@ static __always_inline task_info_t *init_task_info(u32 tid, u32 pid)
     if (proc_info == NULL) {
         scratch->proc_info.new_proc = false;
         scratch->proc_info.follow = false;
-        __builtin_memset(scratch->proc_info.binary, 0, MAX_BIN_PATH_SIZE);
+        scratch->proc_info.binary.mnt_id = 0;
+        scratch->proc_info.binary_no_mnt = 0;
+        __builtin_memset(scratch->proc_info.binary.path, 0, MAX_BIN_PATH_SIZE);
         bpf_map_update_elem(&proc_info_map, &pid, &scratch->proc_info, BPF_NOEXIST);
     }
 
@@ -2101,7 +2105,16 @@ static __always_inline int do_should_trace(event_data_t *data)
 
     if (config & FILTER_BIN_PATH_ENABLED) {
         bool filter_out = (config & FILTER_BIN_PATH_OUT) == FILTER_BIN_PATH_OUT;
-        if (!equality_filter_matches(filter_out, &binary_filter, proc_info->binary))
+        // lookup by binary path only
+        u32 *equality = bpf_map_lookup_elem(&binary_filter, proc_info->binary.path);
+        if (equality != NULL)
+            return *equality;
+        // lookup by binary path and mount namespace
+        equality = bpf_map_lookup_elem(&binary_filter, &proc_info->binary);
+        if (equality != NULL)
+            return *equality;
+
+        if (!filter_out)
             return 0;
     }
 
@@ -3726,8 +3739,9 @@ int tracepoint__sched__sched_process_exec(struct bpf_raw_tracepoint_args *ctx)
     proc_info->new_proc = true;
 
     // extract the binary name to be used in should_trace
-    __builtin_memset(proc_info->binary, 0, MAX_BIN_PATH_SIZE);
-    bpf_probe_read_str(proc_info->binary, MAX_BIN_PATH_SIZE, file_path);
+    __builtin_memset(proc_info->binary.path, 0, MAX_BIN_PATH_SIZE);
+    bpf_probe_read_str(proc_info->binary.path, MAX_BIN_PATH_SIZE, file_path);
+    proc_info->binary.mnt_id = data.context.task.mnt_id;
 
     if (!should_trace(&data))
         return 0;
--- a/pkg/filters/binary.go
+++ b/pkg/filters/binary.go
@@ -1,0 +1,203 @@
+package filters
+
+import (
+	"encoding/binary"
+	"strconv"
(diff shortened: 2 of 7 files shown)


pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2100,7 +2104,16 @@ static __always_inline int do_should_trace(event_data_t *data)
 
     if (config & FILTER_BIN_PATH_ENABLED) {
         bool filter_out = (config & FILTER_BIN_PATH_OUT) == FILTER_BIN_PATH_OUT;
-        if (!equality_filter_matches(filter_out, &binary_filter, proc_info->binary))
+        // lookup by binary path only
+        u32 *equality = bpf_map_lookup_elem(&binary_filter, proc_info->binary.path);
```
NDStrahilevitz, line 2108: Shouldn't there be a check for `binary_no_mnt` instead of doing two lookups?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -783,10 +783,16 @@ typedef struct task_info {
     u8 container_state;   // the state of the container the task resides in
 } task_info_t;
 
+typedef struct binary {
+    u64 mnt_id;
```
NDStrahilevitz, line 787: On my machine the max mnt_id is `127292` (from `/proc/sys/user/max_mnt_namespaces`).
If this is uniform across systems, then u64 max is much larger.
I wonder if we can use int64 here instead, then when `no_mnt_id=1` we set `mnt_id=-1`.
Then we could later use only one map lookup:
`equality = bpf_map_lookup_elem(&binary_filter, &proc_info->binary);`
Which would compare both the binary and the mnt_id which in a "neutral mode" would always be -1.
Could this work?

pkg/filters/binary.go
```diff
@@ -0,0 +1,182 @@
...
+
+func (f *BinaryFilter) FilterOut() bool {
+	if len(f.equal) > 0 && len(f.notEqual) == 0 {
+		return false
+	} else {
+		return true
+	}
+}
+
+const (
```
NDStrahilevitz, line 126: nit: unless we reuse these later it would be better to keep it in the using function(`InitBPF`) so it won't pollute the module's namespace

pkg/ebpf/c/tracee.bpf.c
yanivagman: ns_id is 32 bit indeed. I will change it. Anyway, no_mnt_id is always zero. What do you suggest exactly?

pkg/ebpf/c/tracee.bpf.c
NDStrahilevitz: Maybe you meant to rename but didn't, i'll mark the field in a review

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -783,10 +783,16 @@ typedef struct task_info {
...
+    u32 mnt_id;
+    char path[MAX_BIN_PATH_SIZE];
+} binary_t;
+
 typedef struct proc_info {
     bool new_proc; // set if this process was started after tracee. Used with new_pid filter
     bool follow;   // set if this process was traced before. Used with the follow filter
-    char binary[MAX_BIN_PATH_SIZE];
+    struct binary binary;
+    u32 binary_no_mnt; // used in binary lookup when we don't care about mount ns. always 0.
```
NDStrahilevitz, line 795: this one.

NDStrahilevitz: Is this meant to be set when the binary filter ignores the mount namespace?

pkg/ebpf/c/tracee.bpf.c
NDStrahilevitz: If the field is supposed to be set as I thought it was, then we can use it to determine which lookup to do. 
But better is if we know how to set it we can default the mount ns itself to -1 and then the hashing will always match without caring for the namespace.

pkg/ebpf/c/tracee.bpf.c
yanivagman: No. It is meant to be always 0.

pkg/ebpf/c/tracee.bpf.c
yanivagman: It is meant to be always 0.
It is used for the lookup, when we give the key starting from the binary path and not from mnt_id of the binary. Then, padding with zero at the end with `binary_no_mnt` is used to look for matches with no mount namespace.

pkg/filters/binary.go
```diff
@@ -0,0 +1,182 @@
...
+	fn := func(bin nsBinary, eqVal uint32) error {
+		if len(bin.path) > maxBpfBinPathSize {
+			return InvalidValue(bin.path)
+		}
+		binBytes := make([]byte, bpfBinFilterSize)
+		if bin.mntNS == 0 {
+			copy(binBytes, bin.path)
+		} else {
+			binary.LittleEndian.PutUint32(binBytes, bin.mntNS)
+			copy(binBytes[4:], bin.path)
```
rafaeldtinoco, line 162: I think you should comment here on the need to use the initial 32bit when no mountNs is given (due to map indexing). It did not occur to me naturally when reading.

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: I asked this to be documented (also had doubts when first read, then when reading the bpf code it was clear).

rafaeldtinoco (changes requested):
I believe you need to raise privileges for procfs reading AND to comment on the initial 32bit from the binary map entry. Other than that I'm +1.

pkg/filters/binary.go
```diff
@@ -0,0 +1,182 @@
...
+	hostMntNS uint32
+	enabled   bool
+}
+
+func getHostMntNS() uint32 {
+	if ns, err := proc.GetProcNS(1, "mnt"); err == nil {
+		return uint32(ns)
+	}
+
+	return 0
```
NDStrahilevitz, line 30: If there was an error, is this a valid value? There is no later logic taking this value into concern, so if it's not necessarily correct we should probably add an error return value here.

pkg/filters/binary.go
```diff
@@ -0,0 +1,182 @@
...
+
+type BinaryFilter struct {
+	equal     map[nsBinary]bool
+	notEqual  map[nsBinary]bool
+	hostMntNS uint32
+	enabled   bool
+}
+
+func getHostMntNS() uint32 {
+	if ns, err := proc.GetProcNS(1, "mnt"); err == nil {
```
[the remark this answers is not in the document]
NDStrahilevitz: True. This could also be another reason for the function to fail and a reason to add an `error` return value.

pkg/filters/binary.go
```diff
@@ -0,0 +1,182 @@
...
+func NewBPFBinaryFilter(binaryMapName string) *BPFBinaryFilter {
+	return &BPFBinaryFilter{
+		BinaryFilter:  *NewBinaryFilter(),
+		binaryMapName: binaryMapName,
+	}
+}
+
+func (f *BPFBinaryFilter) InitBPF(bpfModule *bpf.Module) error {
+	const (
+		maxBpfBinPathSize = 256 // maximum binary path size supported by BPF
```
NDStrahilevitz, line 140: Suggestion, add a reference to the const name in the eBPF code if anyone should need reference for it later.

NDStrahilevitz (commented):
Overall LGTM, minor queries before I +1.

pkg/ebpf/c/tracee.bpf.c
yanivagman: changed to u32

pkg/filters/binary.go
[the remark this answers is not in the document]
yanivagman: Done

pkg/filters/binary.go
yanivagman: Done

pkg/filters/binary.go
```diff
@@ -0,0 +1,203 @@
...
+
+	for _, val := range values {
+		mntAndPath := strings.Split(val, ":")
+		var bin nsBinary
+		if len(mntAndPath) == 1 {
+			bin.path = val
+		} else if len(mntAndPath) == 2 {
+			bin.path = mntAndPath[1]
+			if mntAndPath[0] == "host" {
+				if hostMntNS == 0 {
```
NDStrahilevitz, line 79: nit: We can probably cache this for the first time then reuse later and short circuit if necessary.

NDStrahilevitz (approved):
LGTM, added one optimization suggestion but I don't think it's critical for merging or not.

pkg/filters/binary.go
yanivagman: This is exactly what I'm already doing here.
hostMntNS is defined above, and once it becomes non-zero, it is considered cached

NDStrahilevitz: No I mean you can do it above the parsing loop. Store it in the struct like before, do the check then fail even earlier. This is mostly relevant if we ever want to do runtime parsing of this filter.
Maybe even do it in the constructor and add an error value there (but it will make the filter initialization code slightly more annoying(.

NDStrahilevitz: No I mean you can do it above the parsing loop. Store it in the struct like before, do the check then fail even earlier. This is mostly relevant if we ever want to do runtime parsing of this filter.
Maybe even do it in the constructor and add an error value there (but it will make the filter initialization code slightly more annoying).

yanivagman: So other than failing earlier it has no real benefit since caching is already imlpemented here.
I think it is ok to continue as is.

rafaeldtinoco (approved):
LGTM

Issue #2526, opened 2022-12-27
kernelSymbols not updated for derived functions


roikol:
## Description

some derived functions need to use `t.kernelSymbols`.
but the derived functions are in the `derive` package, therefore when `t.kernelSymbols` is updated, it is not updated for the derived functions, and they use outdated data.

this is caused because we "create" the derivation functions when tracee begins, and they receive the `t.kernelSymbols` as an argument.

```
events.PrintSyscallTable: {
	events.HookedSyscalls: {
		Enabled:        t.events[events.PrintSyscallTable].submit,
		DeriveFunction: derive.DetectHookedSyscall(t.kernelSymbols),
	},
},
...
events.PrintNetSeqOps: {
	events.HookedSeqOps: {
		Enabled:        t.events[events.HookedSeqOps].submit,
		DeriveFunction: derive.HookedSeqOps(t.kernelSymbols),
	},
},
```

for exapmle, when a new kernel module that hooks seq operations is loaded, we get an event about the hooks, but because kernelSymbols are not updated for the derivation func, we don't know the target of the hook, and the owner of it is presumed "hidden":
`12:46:22:310426  0      insmod           9669    9669    0                hooked_seq_ops       hooked_seq_ops: map[tcp4_seq_ops_show:{ hidden}]`

instead, we should get this:
`12:46:22:310426  0      insmod           9669    9669    0                hooked_seq_ops       hooked_seq_ops: map[tcp4_seq_ops_show:{hook_tcp4_seq_show main}]`

roikol:
related https://github.com/aquasecurity/libbpfgo/pull/282

PR #2527, opened 2022-12-27
pkg/derive: use up-to-date kernelSymbols

roikol:
use global variable of KernelSymbols in derive package to ensure using the updated kernel symbols at all times

## Initial Checklist

## Description (git log)

```
Date:   Tue Dec 27 13:27:38 2022 +0200

    pkg/derive: use up-to-date kernelSymbols
    
    use global variable of KernelSymbols in derive package to ensure
    using the updated kernel symbols at all times
```

Fixes: #2526

## Type of change

## How Has This Been Tested?

- [ ] Test File A
- [ ] Test File B

- command 01
- command 02

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/ebpf/tracee.go
+++ b/pkg/ebpf/tracee.go
@@ -395,6 +395,7 @@ func (t *Tracee) Init() error {
 			if !initialization.ValidateKsymbolsTable(t.kernelSymbols) {
 				return fmt.Errorf("kernel symbols were not loaded currectly")
 			}
+			derive.KernelSymbols = t.kernelSymbols
 			return nil
 
 		}, cap.SYSLOG)
@@ -667,7 +668,7 @@ func (t *Tracee) initDerivationTable() error {
 		events.PrintSyscallTable: {
 			events.HookedSyscalls: {
 				Enabled:        t.events[events.PrintSyscallTable].submit,
-				DeriveFunction: derive.DetectHookedSyscall(t.kernelSymbols),
+				DeriveFunction: derive.DetectHookedSyscall(),
 			},
 		},
 		events.DnsRequest: {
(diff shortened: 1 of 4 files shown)


NDStrahilevitz:
In the next version of libbpfgo we've added a `Refresh` method for the `KernelSymbolsTable`. As such I think a global variable won't be needed since the injected object can be refreshed and updated.
Since we will likely also move to a lazy version of the `KernelSymbolsTable` (see https://github.com/aquasecurity/libbpfgo/pull/282), it's possible you won't have to do refreshes (up to your decision).

roikol:
> In the next version of libbpfgo we've added a `Refresh` method for the `KernelSymbolsTable`. As such I think a global variable won't be needed since the injected object can be refreshed and updated. Since we will likely also move to a lazy version of the `KernelSymbolsTable` (see [aquasecurity/libbpfgo#282](https://github.com/aquasecurity/libbpfgo/pull/282)), it's possible you won't have to do refreshes (up to your decision).

ok, so when we start using the new libbpfgo version, i'll use the `Refresh` method in the derivation func.
thanks @NDStrahilevitz !

NDStrahilevitz:
> ok, so when we start using the new libbpfgo version, i'll use the `Refresh` method in the derivation func. thanks @NDStrahilevitz !

You already have a mechanism for updating the ksymbols, in `tracee.go/updateKallsyms`. You should use `Refresh` there.
What you were trying to fix is probably related to the fact that the dependncy injection of the derivation would keep the old object, but since you will always have the same reference now, doing the `Refresh` somewhere external like the `updateKallsyms` will be better.

roikol:
> the

oh i see, the reference wouldn't change once we'll use the new version, so the problem should not exists anymore

Issue #1385, opened 2022-01-24
Add HTTP request/response events


yanivagman:
Add HTTP request event ( can include the command, headers, host, uri, protocol...)
Add HTTP response event ( can include status code, headers, protocol...)

rafaeldtinoco:
Closed by https://github.com/aquasecurity/tracee/pull/2571 and https://github.com/aquasecurity/tracee/pull/2538

PR #2538, opened 2023-01-01
network: add http events

roikol:
this is a draft PR for now - it depends on changes to the `trace` package: #2535.
i'm opening it for reference.

## Initial Checklist

## Description (git log)

```
Date:   Sun Jan 1 18:53:05 2023 +0200

    network: add HTTP events
    
    introduce new HTTP request and HTTP response events
```

Fixes: #1385

## Type of change

## How Has This Been Tested?

- [ ] Test File A
- [ ] Test File B

- command 01: `./dist/tracee-ebpf -t e=net_packet_http_request,net_packet_http_response`
- command 02: `curl google.com`

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -470,6 +470,7 @@ enum event_id_e
     NET_PACKET_ICMPV6,
     NET_PACKET_DNS,
     DO_MMAP,
+    NET_PACKET_HTTP,
     MAX_EVENT_ID,
 };
 
@@ -7369,11 +7370,12 @@ static __always_inline bool is_socket_supported(struct socket *sock)
 #define SUB_NET_PACKET_ICMP    (1 << 3) // submit net_packet_icmp_base
 #define SUB_NET_PACKET_ICMPV6  (1 << 4) // submit net_packet_icmpv6_base
 #define SUB_NET_PACKET_DNS     (1 << 5) // submit net_packet_dns_base
+#define SUB_NET_PACKET_HTTP    (1 << 6) // submit net_packet_http_base
 
 #define SUB_NET_PACKET_L3      (SUB_NET_PACKET_IP)
 #define SUB_NET_PACKET_L4      (SUB_NET_PACKET_TCP | SUB_NET_PACKET_UDP | \
                                 SUB_NET_PACKET_ICMP | SUB_NET_PACKET_ICMPV6)
-#define SUB_NET_PACKET_L7      (SUB_NET_PACKET_DNS)
+#define SUB_NET_PACKET_L7      (SUB_NET_PACKET_DNS | SUB_NET_PACKET_HTTP)
 
 typedef struct net_event_contextmd {
     u8 submit;
@@ -7466,6 +7468,9 @@ static __always_inline void should_submit_net_event(net_event_context_t *neteven
 
     if (should_submit(NET_PACKET_DNS, config))
         neteventctx->md.submit |= SUB_NET_PACKET_DNS;
+
+    if (should_submit(NET_PACKET_HTTP, config))
+        neteventctx->md.submit |= SUB_NET_PACKET_HTTP;
 }
 
 //
@@ -7600,6 +7605,12 @@ int BPF_KPROBE(trace_security_socket_sendmsg)
     return security_socket_send_recv_msg(sock, &data);
 }
 
+    // network retval values
+    #define family_ipv4     (1 << 0)
+    #define family_ipv6     (1 << 1)
+    #define proto_http_req  (1 << 2)
+    #define proto_http_resp (1 << 3)
+
 //
 // Socket Ingress/Egress eBPF program loader (right before and right after eBPF)
 //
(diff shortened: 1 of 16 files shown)


pkg/ebpf/c/tracee.bpf.c
```diff
@@ -7311,17 +7313,19 @@ static __always_inline bool is_socket_supported(struct socket *sock)
...
-#define SUB_NET_PACKET_ICMP    (1 << 3) // submit net_packet_icmp_base
-#define SUB_NET_PACKET_ICMPV6  (1 << 4) // submit net_packet_icmpv6_base
-#define SUB_NET_PACKET_DNS     (1 << 5) // submit net_packet_dns_base
+#define SUB_NET_PACKET_IP               (1 << 0) // submit net_packet_ip_base
+#define SUB_NET_PACKET_TCP              (1 << 1) // submit net_packet_tcp_base
+#define SUB_NET_PACKET_UDP              (1 << 2) // submit net_packet_udp_base
+#define SUB_NET_PACKET_ICMP             (1 << 3) // submit net_packet_icmp_base
+#define SUB_NET_PACKET_ICMPV6           (1 << 4) // submit net_packet_icmpv6_base
+#define SUB_NET_PACKET_DNS              (1 << 5) // submit net_packet_dns_base
+#define SUB_NET_PACKET_HTTP_REQUEST     (1 << 6) // submit net_packet_http_request_base
```
rafaeldtinoco, line 7322: Don't you think we should have SUB_NET_PACKET_HTTP here only ? Just like the DNS, the distinction between REQUEST/RESPONSE would be done in the derivation layer. WDYT ?

roikol: yes i wanted it to be like the DNS - the problem is that the `http` package i'm using in the derivation layer expects me to know whether it's a request or a reponse. 
i'm checking if its a request or response in the kernel. so i think we can solve it by adding an argument of the direction of the packet to the HTTP base event, and then i'll know how to derive. sounds good?

roikol:
> Very nice that you're creating this!
> 
> I have plans for adding SSL support for all current events (by uprobing libssl and tying the reads/writes(), before encryption, to a traced socket). This way we could have the "real" content (before encryption) and use it as if it was the packet payload.
> 
> I thought we could "automagically" decrypt SSL for both: events and pcap data.

sounds awesome! it will be a very strong capability in tracee

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: Hum... I did something similar to differentiate IPv4 and IPv6.. if you look at `cgroup_bpf_run_filter_skb` kprobe, I'm using "eventctx->retval = family", so userland knows what to expect (v4 or v6) in the derivation phase without adding anything to the context.

We could expand that idea (have an enum with multiple values for retval ?)... 

The idea of having more arguments to the base network event is also ok, I considered it in the past but wanted to avoid it as much as we could (to make uniform to all network base events just to have a single "payload" argument).

Adding more arguments to the event will require:

```C
typedef struct net_event_context {
    event_context_t eventctx;
    struct { // event arguments (needs packing), use anonymous struct to ...
        u8 index0;
        u32 bytes;
        u8 index1; // == 1, set by default
        u32 bytes;
        // ... (payload sent by bpf_perf_event_output)
    } __attribute__((__packed__)); // ... avoid address-of-packed-member warns
    // members bellow this point are metadata (not part of event to be sent)
    net_event_contextmd_t md;
} net_event_context_t;
```

and padding recalculation (need to confirm sizeof_net_event_context_t() function is good). 

I think I'd rather try to use regular existing fields (like the retval) with different enums... we could have something like:

#define XXX_ipv4 (1 << 0)
#define XXX_ipv6 (1 << 1)
#define XXX_HTTP_req (1 << 2)
#define XXX_HTTP_resp (1 << 3)

and set the retval to it. WDYT ?

roikol: that's a nice approach. for now i think retval is good enough for our needs (maybe in the future we can add another argument).

i've updated this PR - it now checks whether HTTP request/response both in BPF part and in userpsace part.

once the types PR (#2535) will be merged i'll update this one to use the retval instead of having the same check twice.
(this one is currently for reference).

rafaeldtinoco (commented):
Nice job @roikol. I have made a few changes (comments and styles mostly) and merged them with the following PR: https://github.com/aquasecurity/tracee/pull/2571

Please could you create a test for HTTP events, using the same structure I've done for the other network tests? This way we can test HTTP on each PR. Thanks!

Question: I wonder if we should include PktMeta to all other network packet events (instead of having src/dst src_port/dst_port single arguments). My doubt is about filtering, we would have to create a specific filter type for "PkgMeta", but, still, might be worth. WDYT ?

roikol:
> Nice job @roikol. I have made a few changes (comments and styles mostly) and merged them with the following PR: #2571
> 
> Please could you create a test for HTTP events, using the same structure I've done for the other network tests? This way we can test HTTP on each PR. Thanks!
> 
> Question: I wonder if we should include PktMeta to all other network packet events (instead of having src/dst src_port/dst_port single arguments). My doubt is about filtering, we would have to create a specific filter type for "PkgMeta", but, still, might be worth. WDYT ?

thanks @rafaeldtinoco !

i'll work on adding the tests.

regarding the `PktMeta` - since these arguments (IPs, ports, etc.) are part of any packet related event, i think it's a good idea to keep it in a structure. otherwise every packet related event will have 7 argument before adding the specific arguments for the event, which is quite a lot.
we can keep this in mind and see if in future network events this assumption is still true - if not, we can change it.
also - are we going to fill the `PktMeta.Iface` with the real interface? if not, we can remove it.
WDYT?

rafaeldtinoco:
> are we going to fill the `PktMeta.Iface` with the real interface? if not, we can remove it.

I don't know yet. I wonder if that info would be important for you guys. I think I can get it from the skb/ctx from cgroup programs IIRC. If you don't think its worth we can remove it for now and re-add if ever needed!

rafaeldtinoco:
> i think it's a good idea to keep it in a structure. otherwise every packet related event will have 7 argument before adding the specific arguments for the event, which is quite a lot.

agreed!

Issue #2126, opened 2022-08-30
Container folder is not getting created under /tmp/tracee 


OriGlassman:
## Bug description

This is how I'm executing tracee (cmdline):
tracee-ebpf -t c -c net=docker0 -c pcap:per-container -c mem -c write=memfd:* -c write=/*  -o option:exec-env -o option:exec-hash -o option:parse-arguments -o format:json

This is the error I'm getting:
For a specific container, the container folder (and thus the pcap) is not getting created /tmp/tracee/out/<cont_id>. 
The folder is getting created after a write inside the container (but even then no pcap is generated afterwards).
By looking at the network_map, there is a value corresponding to the container:
`        "key": {
            "address": {
                "in6_u": {
                    "u6_addr8": [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0
                    ],
                    "u6_addr16": [0,0,0,0,0,0,0,0
                    ],
                    "u6_addr32": [0,0,0,0
                    ]
                }
            },
            "port": 34835,
            "protocol": 6
        },
        "value": {
            "host_tid": 43028,
            "comm": "registry"
        }
`
 

## Context

Relevant information about my setup:

* Linux version: Ubuntu 20.04.4
* Linux kernel version: 5.15.0-1017-aws
* Tracee version (or commit id of your tree): v0.8.1
* LLVM version:
* Golang version:

rafaeldtinoco:
Alright, after speaking to @OriGlassman, he said he was reproducing 100% of the times (differently than before) and I was given a cmdline containing:

```
-t net=docker0 -c pcap:per-container
```

but missing

```
-c net=docker0
```

Unfortunately, the "current" network code needs for user to specify the net=XXX multiple times (for tracing and capturing).

With that said, I believe there is still a minor race condition when the directory isn't created and the pcap file is tried to be written (which is what this issue is about).

Issue #2096, opened 2022-08-25
Don't require specifying a device for network pcap


OriGlassman:
As of today tracee-ebpf has two pcap related arguments: '-c net=<interface> and -c pcap:<host|per-container>'.
This puts the burden of capturing pcaps on the user to supply an interface..
I suggest to have a 'pcap:<host,containers,all>' (default all) option, and leave tracee to detect the relevant network interface.

rafaeldtinoco:
The `-c net=` will vanish as soon as we drop tc progs attached to interfaces (being currently worked on). I'm about to create all issues for the network improvements about to come. Will label this as "networking" (and all subsequent/related ones).

Issue #2152, opened 2022-09-06
Redesign network events and capture features


rafaeldtinoco:
## Redesign network events and capture features

This issue is about redesigning the old network model, based on tc hooks and defining a specific interface to be monitored, into cgroup ebpf program(s) that are capable of generating network packets (including for different layers and protocols). This issue also defines some related posterior (to needed changes) work.

The new network code requirements are:

- be capable of relating egress/ingress packets to specific host tasks
- not to rely on specific network interfaces for attachments
- be close to processes (after mangling and translations)
- allow near future flow enforcements through eBPF
- work in older kernels (not only in recent ones)

## Technicalities

Covered Issues:

1. Remove kprobes used to keep context among networking flows
  #2569
1. Improve performance and better skb programs in newer kernels
  #2578
1. To have a single pcap file by default for `--capture network`
  #2587

Related Bugs:

1. add MPTCP support
  #2068

Finished Work (old Issues model, one issue per item)

1. Recalculate the socket's task context  
  #2221 
1. Entire clsact qdisc is purged when tc hook is destroyed
  #1828
1. network events should also be enriched
  #1922
1. container folder is not getting created under
  #2126
1. pcap capturing options
  #2096
1  Split network eBPF logic into multiple files
   #2545

# Done by others

1.  HTTP request/response events 
  #1385 
1. add e2e test for http events
  #2574

PR #2547, opened 2023-01-05
network: add cgroup based pcap capture code

rafaeldtinoco:
## Description (git log)

**network: add cgroup based pcap capture code**

The new cgroup based pcap capture code is able to capture ANY TCP, UDP or ICMP packet for both IPv4 and IPv6 protocols from any physical (L1) and link (L2) layers.

Differently than the current existing pcap capture code, this new approach allows tracee to capture packets from any process on the host or in a container, independently of the interfaces they are using to communicate. There is no need to bind a tc hook program to a specific interface now.

```
$ tracee-ebpf --capture network
```

There is also a mid layer code responsible for managing opened pcap files using a LRU cache (so tracee has a limit on amount of opened files being used).

Another feature is the capability of selecting one or multiple pcap file types to have captured:

- per process   (processes/{host,container_id}/process_TID_TS.pcap)
- per container (containers/container_id.pcap)
- per command   (commands/{host,container_id}/command.pcap)

```
$ tracee-ebpf --capture network --capture netpcap:process,container,command
```

NOTE: This allows the previous tc hook based network code to
      be retired and, with it, the need for specifying an
      interface to be hooked to.

Fixes: #2126
Fixes: #2096
Related: #2152

## Type of change

## How Has This Been Tested?

Execute the commands described in the git log message and check for the created pcap files under `/tmp/tracee/out/pcap/` directory.

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/docs/docs/capturing/index.md
+++ b/docs/docs/capturing/index.md
@@ -116,6 +116,120 @@ Tracee can capture the following types of artifacts:
 
 1. **Network PCAP Files**
 
+     Anytime a **network packet** is delivered to a process, traced by tracee,
+     this packet might be captured into one or multiple pcap files.
+
+     A good way to test this behavior is to execute:
+
+     ```text
+     $ sudo ./dist/tracee-ebpf \
+         --capture network
+     ```
+
+     and observe **pcap files** for all processes being created:
+
+     ```text
+     $ cd /tmp/tracee/out
+     $ find pcap
+     pcap
+     pcap/processes
+     pcap/processes/host
+     pcap/processes/host/node_186708_1573567360495399.pcap
+     pcap/processes/host/node_1196826_1662656211119567.pcap
+     pcap/processes/host/zerotier-one_7882_137007714376.pcap
+     pcap/processes/host/sshd_1196773_1662654999660718.pcap
+     ```
+
+     You can also select how the pcap files will be divided:
+
+     1. per process
+     1. per container
+     1. per-command
+
+     and you can even have multiple divisions at the same time. Example: a ping
+     command is executed inside a container. You want to summarize captured
+     traffic per container and per command. You will find the same captured
+     data for that ping command inside `commands/container_id/ping.pcap` and
+     inside `containers/container_id.pcap`.
+
+     ```text
+     $ sudo ./dist/tracee-ebpf \
+         --capture network \
+         --capture netpcap:process,container,command
+     ```
+
+     ```text
+     $ cd /tmp/tracee/out
+     $ find pcap
+     pcap
+     pcap/commands
+     pcap/commands/b86533d11f3
+     pcap/commands/b86533d11f3/ping.pcap
+     pcap/commands/host
+     pcap/commands/host/sshd.pcap
+     pcap/commands/host/zerotier-one.pcap
+     pcap/commands/host/node.pcap
+     pcap/commands/fd95a035ce5
+     pcap/commands/fd95a035ce5/ping.pcap
+     pcap/processes
+     pcap/processes/b86533d11f3
+     pcap/processes/b86533d11f3/ping_1261180_1663772450241192.pcap
+     pcap/processes/host
+     pcap/processes/host/node_186708_1573567360495399.pcap
+     pcap/processes/host/node_1196826_1662656211119567.pcap
+     pcap/processes/host/zerotier-one_7882_137007714376.pcap
+     pcap/processes/host/sshd_1196773_1662654999660718.pcap
+     pcap/processes/fd95a035ce5
+     pcap/processes/fd95a035ce5/ping_1261163_1663769383806467.pcap
+     pcap/containers
+     pcap/containers/host.pcap
+     pcap/containers/b86533d11f3.pcap
+     pcap/containers/fd95a035ce5.pcap
+     ```
+
+     you can see the packets by executing tcpdump on any pcap file:
+
+     ```text
+     $ tcpdump -r pcap/containers/b86533d11f3.pcap
+     reading from file pcap/containers/b86533d11f3.pcap, link-type NULL (BSD loopback), snapshot length 65535
+     02:52:00.524035 IP 172.17.0.3 > dns.google: ICMP echo request, id 5, seq 476, length 64
+     02:52:00.533145 IP dns.google > 172.17.0.3: ICMP echo reply, id 5, seq 476, length 64
+     02:52:01.525455 IP 172.17.0.3 > dns.google: ICMP echo request, id 5, seq 477, length 64
+     02:52:01.535414 IP dns.google > 172.17.0.3: ICMP echo reply, id 5, seq 477, length 64
+     02:52:02.526715 IP 172.17.0.3 > dns.google: ICMP echo request, id 5, seq 478, length 64
+     02:52:02.536444 IP dns.google > 172.17.0.3: ICMP echo reply, id 5, seq 478, length 64
+     02:52:03.528739 IP 172.17.0.3 > dns.google: ICMP echo request, id 5, seq 479, length 64
+     02:52:03.538622 IP dns.google > 172.17.0.3: ICMP echo reply, id 5, seq 479, length 64
+
+     $ tcpdump -r pcap/commands/b86533d11f3/ping.pcap
+     reading from file pcap/commands/b86533d11f3/ping.pcap, link-type NULL (BSD loopback), snapshot length 65535
+     02:52:00.524035 IP 172.17.0.3 > dns.google: ICMP echo request, id 5, seq 476, length 64
+     02:52:00.533145 IP dns.google > 172.17.0.3: ICMP echo reply, id 5, seq 476, length 64
+     02:52:01.525455 IP 172.17.0.3 > dns.google: ICMP echo request, id 5, seq 477, length 64
+     02:52:01.535414 IP dns.google > 172.17.0.3: ICMP echo reply, id 5, seq 477, length 64
+     02:52:02.526715 IP 172.17.0.3 > dns.google: ICMP echo request, id 5, seq 478, length 64
+     02:52:02.536444 IP dns.google > 172.17.0.3: ICMP echo reply, id 5, seq 478, length 64
+     02:52:03.528739 IP 172.17.0.3 > dns.google: ICMP echo request, id 5, seq 479, length 64
+     02:52:03.538622 IP dns.google > 172.17.0.3: ICMP echo reply, id 5, seq 479, length 64
+     ```
+
+    !!! Note
+        Note that the same packets were written to 2 different pcap files: the
+        pcap file describing the container `b86533d11f3` (because it was
+        executing a single process: ping) and the pcap file describing ANY ping
+        command executed in that container (commands/b86533d11f3/ping.pcap).
+
+    The format for the pcap filenames inside `output_dir` is the following:
+
+    1. **processes**:  
+       ./pcap/processes/`container_id`/`process_comm`/`host_tid`/`task_starttime`.pcap
+    1. **containers**:  
+       ./pcap/containers/`container_id`.pcap
+    1. **commands**:  
+       ./pcap/commands/`container_id`/`process_comm`.pcap
+
+1. **Network PCAP Files (deprecated)**
+
      Anytime a **packet** goes through the **network interface**, the **packet
      is captured into the pcap file**. Only packets that are generated by traced
      processes are being captured.
@@ -132,16 +246,16 @@ Tracee can capture the following types of artifacts:
      ```
 
      and execute on the host:
-     
+
      ```text
-     $ ping 127.0.0.1
+      $ ping 127.0.0.1
      ```
-     
+
      and observe **pcap file**:
-     
+
      ```text
      $ tcpdump -n -r /tmp/tracee/out/host/capture.pcap
-     
+
      15:48:33.109392 IP 127.0.0.1 > 127.0.0.1: ICMP echo request, id 74, seq 1, length 64
      15:48:33.109440 IP 127.0.0.1 > 127.0.0.1: ICMP echo reply, id 74, seq 1, length 64
      15:48:34.138680 IP 127.0.0.1 > 127.0.0.1: ICMP echo request, id 74, seq 2, length 64
--- a/pkg/bufferdecoder/eventsreader.go
+++ b/pkg/bufferdecoder/eventsreader.go
@@ -149,7 +149,9 @@ func ReadArgFromBuff(ebpfMsgDecoder *EbpfDecoder, params []trace.ArgMeta) (trace
 		if err != nil {
 			return argMeta, nil, fmt.Errorf("error reading byte array size: %v", err)
 		}
-		if size > 4096 {
+		// error if byte buffer is too big (and not a network event)
+		// TODO: check for network event IDs range isntead of arg name
+		if size > 4096 && argMeta.Name != "payload" {
 			return argMeta, nil, fmt.Errorf("byte array size too big: %d", size)
 		}
 		res, err = ReadByteSliceFromBuff(ebpfMsgDecoder, int(size))
--- a/pkg/cmd/flags/capture.go
+++ b/pkg/cmd/flags/capture.go
@@ -19,11 +19,13 @@ Possible options:
 [artifact:]module                  capture loaded kernel modules.
 [artifact:]mem                     capture memory regions that had write+execute (w+x) protection, and then changed to execute (x) only.
 [artifact:]net=interface           capture network traffic of the given interface. Only TCP/UDP/ICMP protocols are currently supported.
+[artifact:]network                 capture network traffic. Only TCP/UDP/ICMP protocols are currently supported.
 
 dir:/path/to/dir                    path where tracee will save produced artifacts. the artifact will be saved into an 'out' subdirectory. (default: /tmp/tracee).
 profile                             creates a runtime profile of program executions and their metadata for forensics use.
 clear-dir                           clear the captured artifacts output dir before starting (default: false).
 pcap:[per-container|per-process]    capture separate pcap file based on container/process context (default: none - saving one pcap for the entire host).
+netpcap:[process,container,command] capture separate pcap files organized by processes, containers and/or commands
 
 Examples:
   --capture exec                                           | capture executed files into the default output directory
@@ -31,7 +33,10 @@ Examples:
   --capture write=/usr/bin/* --capture write=/etc/*        | capture files that were written into anywhere under /usr/bin/ or /etc/
   --capture profile                                        | capture executed files and create a runtime profile in the output directory
   --capture net=eth0                                       | capture network traffic of eth0
+  --capture network                                        | capture network traffic. default: pcap for processes only (check netpcap options)
   --capture net=eth0 --capture pcap:per-container          | capture network traffic of eth0, and save pcap for each container
+  --capture network --capture netpcap:process,command      | capture network traffic, save pcap files for processes and commands
+  --capture network --capture netpcap:container,command    | capture network traffic, save pcap files for containers and commands
   --capture exec --output none                             | capture executed files into the default output directory not printing the stream of events
 
 Use this flag multiple times to choose multiple capture options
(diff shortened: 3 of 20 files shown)


rafaeldtinoco:
I'll retire the old network capture code once this is merged (and likely the tc hooks logic as well, since we now have the DNS events and PCAP capturing using cgroup probes).

pkg/pcap/common.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	// create the string that will hold the pcap filename
+
+	var format string
+
+	switch t {
+	case process:
+		format = fmt.Sprintf(
+			"pcap/processes/%v/%v_%v_%v.pcap",
```
NDStrahilevitz, line 100: example where the constants also fit

pkg/pcap/cache.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"github.com/aquasecurity/tracee/types/trace"
+)
+
+//
+// Multiple Pcap Implementation Caches
+//
+
+type pcapCache interface {
```
NDStrahilevitz, line 11: Now for something more substantive.
The interface here is redundant and can be replaced with one `struct` using a generic type argument whose constraint is the `pcap` interface.
Unless I missed something all the caches functionally do exactly the same thing except for type checking the pcap object stored. Using a generic implementation of Cache you can use the `T` type to do the type checking according to the type provided when constructing the cache. This way we reduce the 3 separate implementations to one.

pkg/pcap/pcaps.go
```diff
@@ -0,0 +1,156 @@
...
+// All Pcap Implementations at Once
+//
+
+type Pcaps struct {
+	processes  pcapCache // ProcessPcap Cache (indexed by host_tid)
+	containers pcapCache // ContainersPcap Cache (indexed by container_id)
+	commands   pcapCache // CommandsPcap Cache (indexed by comm)
+}
+
+func NewPcaps(config int, output *os.File) (*Pcaps, error) {
```
NDStrahilevitz, line 39: using the int as a config with flags isn't very readable. This easily could have been a struct with three bool flags and it would be much more readable.

pkg/pcap/pcaps.go
```diff
@@ -0,0 +1,156 @@
...
+	// 7 (111): process + container + command
+	//
+
+	initializeGlobalVars(output)
+
+	var processes pcapCache
+	var containers pcapCache
+	var commands pcapCache
+
+	if config&0x1 == 0x1 {
```
NDStrahilevitz, line 62: Then you could have `if config.captureProcesses` etc. which is much more immediately readable.

pkg/pcap/pcaps.go
```diff
@@ -0,0 +1,156 @@
...
+// NOTE: Pcaps is not thread safe, should be called from a single routine.
+// TODO: might replace cache_XXX and pcap_XXX files with generics.
+//
+
+//
+// All Pcap Implementations at Once
+//
+
+type Pcaps struct {
+	processes  pcapCache // ProcessPcap Cache (indexed by host_tid)
```
NDStrahilevitz, line 34: We can possibly reduce this to one field of `[]pcapCache` then all operations simply range on the caches.
Later if we want to add another cache type it would only require adding a new config flag and appending a cache (of course if you think future caches won't be added this doesn't really matter).

pkg/pcap/common.go
```diff
@@ -0,0 +1,159 @@
...
+	"sync"
+
+	"github.com/aquasecurity/tracee/pkg/logger"
+	"github.com/aquasecurity/tracee/pkg/utils"
+	"github.com/aquasecurity/tracee/types/trace"
+	"github.com/google/gopacket/layers"
+	"github.com/google/gopacket/pcapgo"
+)
+
+var outputDirectory *os.File
```
NDStrahilevitz, line 16: This should probably be configurable.

pkg/metrics/stats.go
```diff
@@ -15,6 +15,7 @@ type Stats struct {
 	LostEvCount      counter.Counter
 	LostWrCount      counter.Counter
 	LostNtCount      counter.Counter
+	LostNtCapCount   counter.Counter
```
NDStrahilevitz, line 18: Thanks for adding this.
Maybe we should add a capture counter as well? WDYT?

NDStrahilevitz (changes requested):

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -7370,6 +7372,15 @@ struct {
     __type(value, struct entry);     // ... linked to entry ctx->args
 } entrymap SEC(".maps");             // can't use args_map (indexed by existing events only)
 
+// network capture events
+
+struct {
```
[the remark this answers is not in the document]
rafaeldtinoco: I'm not using the map macros for the network code (needed better visibility and control of the structures, and commenting key, value fields, etc).

pkg/bufferdecoder/eventsreader.go
```diff
@@ -149,9 +149,7 @@ func ReadArgFromBuff(ebpfMsgDecoder *EbpfDecoder, params []trace.ArgMeta) (trace
 		if err != nil {
 			return argMeta, nil, fmt.Errorf("error reading byte array size: %v", err)
 		}
-		if size > 4096 {
-			return argMeta, nil, fmt.Errorf("byte array size too big: %d", size)
-		}
+		// do not limit byte array size to 4k (as network packets might be bigger)
```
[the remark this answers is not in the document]
rafaeldtinoco: Maybe set to 4k to all other events but the network ones ? I'll be able to do that once I move all the new network events to the "network event" range (need to remove the legacy code first).

pkg/pcap/common.go
```diff
@@ -0,0 +1,159 @@
...
+	o := outputDirectory // pointer to global var (readability)
+
+	s = "pcap"
+	pcapOnce.Do(func() {
+		utils.MkdirAtExist(o, s, os.ModePerm)
+	})
+
+	switch t {
+	case process:
+		s = "pcap/processes"
```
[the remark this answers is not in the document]
rafaeldtinoco: Yep, you are right, I thought about it and never did.

pkg/pcap/cache.go
rafaeldtinoco: There is a TODO for that in the code. I'll do it right after, in a new PR, if you don't mind.

NDStrahilevitz: FWIW I think doing this for the cache right now would be worth it, it should be pretty easy and quick, for the pcaps maybe less so.

pkg/bufferdecoder/eventsreader.go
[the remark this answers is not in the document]
NDStrahilevitz: Sounds good.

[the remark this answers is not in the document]
yanivagman: You can't simply remove it. MAX_BYTES_ARR_SIZE is defined to be 4096 in the bpf code.

roikol (commented):
overall looks good!

pkg/pcap/cache.go
rafaeldtinoco: I'll give it a try.

pkg/pcap/common.go
rafaeldtinoco: I prefer to keep as is for now.

pkg/pcap/pcap.go
```diff
@@ -0,0 +1,22 @@
...
+
+// Multiple Pcap Implementations
+
+type pcap interface {
+	write(*trace.Event, []byte) error
+	flush() error
+	close() error
+}
+
+const (
```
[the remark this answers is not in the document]
rafaeldtinoco: Yep!

pkg/pcap/pcap_command.go
```diff
@@ -0,0 +1,57 @@
...
+// Per Command Pcap:
+// creates/opens a {host,container}/comm/all.pcap file for each ever execed cmd
+
+type commandPcap struct {
+	packets int // amount of packets written (sync each X times)
+	file    *os.File
+	writer  *pcapgo.NgWriter
+}
+
+func newCommandPcap(event *trace.Event) (*commandPcap, error) {
```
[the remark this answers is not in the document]
rafaeldtinoco: Yep, I've tried that and result wasn't confusing, I can give it a try after all other comments, but not sure as they're small and very clear to read right now (that was my bet for keeping it like this).

pkg/pcap/pcaps.go
rafaeldtinoco: Yep, could be. let's see what generics will look like (if they don't compromise readability too much) and then check if its worth it or keep it as is for now. Will get back to this in the next round.

pkg/pcap/pcaps.go
rafaeldtinoco: I answered Roi on this already. Made it better with const for the flags (and it helps when checking t.config.capture.network != 0 in many parts to see if at least one is enabled, instead of checking multiple bools or another bool that would say something is set).

pkg/pcap/pcaps.go
rafaeldtinoco: I believe the const improved readability and is enough for now.

pkg/pcap/cache.go
rafaeldtinoco: Actually, you spotted a bug. The caches should be indexed differently:

- process cache entry using host_tid
- container cache entry using container_id
- command cache entry using process_name (or comm)

so... they wouldn't be "exactly" the same but still ok for generics, although I don't like the complexity added in a first attempt. I think I'll stick with common functions and minor implementations (easier to understand and maintain).

pkg/bufferdecoder/eventsreader.go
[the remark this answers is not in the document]
rafaeldtinoco: turning this into:

```C
		// error if byte buffer is too big (and not a network event)
		if size > 4096 && argMeta.Name != "payload" {
			return argMeta, nil, fmt.Errorf("byte array size too big: %d", size)
		}
```

pkg/pcap/common.go
```diff
@@ -0,0 +1,185 @@
...
+// TODO: flush per interval as well (so we can grow flushAtPackets higher)
+
+const AF_INET = 2
+const AF_INET6 = 10
+
+//
+// Functions
+//
+
+func initializeGlobalVars(output *os.File) {
```
NDStrahilevitz, line 39: This can just be the `init()` method of the package.
With regards to the output file, i'm rather adamant about moving to a regular config struct, I will expand further near the relevant code.

pkg/utils/utils.go
```diff
@@ -16,3 +18,17 @@ func ParseSymbol(address uint64, table *helpers.KernelSymbolTable) *helpers.Kern
 	hookingFunction.Owner = strings.TrimSuffix(hookingFunction.Owner, "]")
 	return hookingFunction
 }
+
+// ErrorFuncName displays a given error prefixed by the current (caller) fn name
+func ErrorFuncName(e error) error {
```
NDStrahilevitz, line 23: Nice addition, but I think it might be better centralized at the logger level. @geyslan I believe we have something similar there right?

pkg/pcap/pcaps.go
```diff
@@ -0,0 +1,95 @@
+package pcap
```
NDStrahilevitz, line 1: nit: since the "main" module struct is here rather than in `pcap.go` maybe the package should be `pcaps` instead.
Then `NewPcaps` can just be `New` as importers will use it as `pcaps.New()`.

pkg/pcap/pcaps.go
```diff
@@ -0,0 +1,95 @@
...
+// NOTE: Pcaps is not thread safe, should be called from a single routine.
+// TODO: might replace cache_XXX and pcap_XXX files with generics.
+//
+
+// Pcaps holds all Pcap for different PcapTypes
+type Pcaps struct {
+	pcapCaches map[PcapType]*PcapCache
+}
+
+func NewPcaps(config PcapType, output *os.File) (*Pcaps, error) {
```
NDStrahilevitz, line 34: I'm going to stress again that this should be a regular struct config.
I'll lay out my argument again:
1. This is the exception in the project
2. It's not as intuitive for use later in debugging and reading internally.
3. More importantly, it is externally harder to use, especially for non cli users who will have to configure this by hand. Letting them do bit flags to configure is not ok from a consumer standpoint.
4. In addition since we have other configuration options, like the file, which you allow configuring in this function, it shows that it should be open to non-cli users too. It would be easiest if a struct was parsed for the output directory location.
5. If we ever want to add new configuration options, rather than breaking the function signature we simply add a new **non-breaking** struct field.
I think these are enough important point to move from the flag approach. If it's that important, maybe convert it to a  bit flag internally for use, but at least externally, it should be simple bool flags, with the other configuration options (for now the output dir, in the future potentially more).

pkg/bufferdecoder/eventsreader.go
```diff
@@ -149,7 +149,8 @@ func ReadArgFromBuff(ebpfMsgDecoder *EbpfDecoder, params []trace.ArgMeta) (trace
 		if err != nil {
 			return argMeta, nil, fmt.Errorf("error reading byte array size: %v", err)
 		}
-		if size > 4096 {
+		// error if byte buffer is too big (and not a network event)
+		if size > 4096 && argMeta.Name != "payload" {
```
NDStrahilevitz, line 153: How can we be sure `payload` will only ever be used in net events? It's a rather generic name and I can see it being used in other events in the future (especially when we add plugins and users can add their own events).

NDStrahilevitz (changes requested):
Main stress points for me are:
1. The package's configuration scheme.
2. The validation in the eventsreader for the payload argument - I see it backfiring in the future and elaborated in the comment.

pkg/utils/utils.go
rafaeldtinoco: Yep, I suggest @geyslan picks this and move to the logger in another PR, meanwhile this PR can use it as its good for debugging (saved me 2 times).

pkg/bufferdecoder/eventsreader.go
rafaeldtinoco: We can't, but I did not want to pass the eventID as an argument for the function. I could rename the payload arg to something that would only be used by networking code as well OR we can change the function prototype, I'm good with either.

pkg/pcap/pcaps.go
rafaeldtinoco: Yep, I thought about it before, will do.

pkg/pcap/pcaps.go
rafaeldtinoco: Alright, the external user reason (using it as a lib) convinced me.

pkg/utils/utils.go
rafaeldtinoco: Reason (for @geyslan): AFAIK, Logger provides file and function and line but it does not concatenate errors from frame to frame as this does, which is good so I can just return "err" when it happens and it is still useful.

pkg/ebpf/net_capture.go
```diff
@@ -0,0 +1,181 @@
...
+const AF_INET6 = 10
+
+func (t *Tracee) processNetCaptureEvents(ctx gocontext.Context) {
+	var err error
+	var eventsChan <-chan *trace.Event
+	var errChan <-chan error
+	var errChanList []<-chan error
+
+	// create and manage all needed pcap files
+	t.netCapturePcap, err = pcap.NewPcaps(t.config.Capture.Network, t.outDir)
```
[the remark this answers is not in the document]
rafaeldtinoco: Yep, it does and you are right.

pkg/bufferdecoder/eventsreader.go
NDStrahilevitz: Maybe we should mark the argument type as something else? like `NetworkBytes`, then we can have a special handler for that.
@yanivagman WDYT?

rafaeldtinoco: Okay, I have found a way to deal with this, but in the next PR once I remove the old network events and move the new ones to the "network event ID range", then I'll pass an argument to ReadArgFromBuff with eventID and check if we are in the network event range before checking this limit. For now, keep this as is with a TODO for me to do in another PR in this same milestone.

pkg/utils/utils.go
geyslan: I think this function is already in the right place (utils). Perhaps naming it `EnrichErr()` can be better.

NDStrahilevitz (approved):

yanivagman (changes requested):
I'm still in the middle of the review, but placing some comments to be handled before merging this one

pkg/pcaps/common.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+// getContainerID returns the container string to be used in pcap files or dirs
+func getContainerID(contID string) string {
+	if contID == "" {
+		return "host"
+	}
+	if len(contID) > 4 {
+		contID = contID[0:11]
+	}
```
yanivagman, line 90: Don't you need to check for `len>12`?

yanivagman (commented):
Ok, just one more small comment from my side and we're good to go.
Nice work @rafaeldtinoco !

pkg/cmd/flags/capture.go
```diff
@@ -19,19 +19,24 @@ Possible options:
 [artifact:]module                  capture loaded kernel modules.
 [artifact:]mem                     capture memory regions that had write+execute (w+x) protection, and then changed to execute (x) only.
 [artifact:]net=interface           capture network traffic of the given interface. Only TCP/UDP/ICMP protocols are currently supported.
+[artifact:]network                 capture network traffic. Only TCP/UDP/ICMP protocols are currently supported.
 
 dir:/path/to/dir                    path where tracee will save produced artifacts. the artifact will be saved into an 'out' subdirectory. (default: /tmp/tracee).
 profile                             creates a runtime profile of program executions and their metadata for forensics use.
 clear-dir                           clear the captured artifacts output dir before starting (default: false).
 pcap:[per-container|per-process]    capture separate pcap file based on container/process context (default: none - saving one pcap for the entire host).
+netpcap:[process,container,command] capture separate pcap files organized by processes, containers and/or commands
```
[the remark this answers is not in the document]
rafaeldtinoco: I'll update https://github.com/aquasecurity/tracee/pull/2569 with pcap then.

pkg/pcaps/pcaps.go
```diff
@@ -0,0 +1,103 @@
...
+// trace tons of processes, containers or commands).
+//
+// That is what the PcapCache struct is for. It keeps Pcap files and maintain
+// those Pcap files cached while they're still being used.
+//
+// At the end we have the Pcap struct itself. It describes a pcap file being
+// kept opened on behalf of a process, a container or a command.
+//
+// NOTE: Pcaps is not thread safe, should be called from a single routine.
+// TODO: might replace cache_XXX and pcap_XXX files with generics.
```
rafaeldtinoco, line 26: Remove this comment (won't do that).

docs/docs/capturing/index.md
```diff
@@ -116,6 +116,103 @@ Tracee can capture the following types of artifacts:
...
+    !!! Note
+        Note that the same packets were written to 2 different pcap files: the
+        pcap file describing the container `b86533d11f3` (because it was
+        executing a single process: ping) and the pcap file describing ANY ping
+        command executed in that container (commands/b86533d11f3/ping.pcap).
+
+1. **Network PCAP Files (deprecated)**
+
      Anytime a **packet** goes through the **network interface**, the **packet
      is captured into the pcap file**. Only packets that are generated by traced
```
rafaeldtinoco, line 217: This is not true. All packets are being captured, no matter the traced events. Will need some coordination with multiple scope changes IF we want to limit the packets being written to pcap files using the filters.

Need to clarify that here.

pkg/pcaps/cache.go
```diff
@@ -0,0 +1,79 @@
...
+	var i interface{}
+
+	i, ok = p.itemCache.Get(getEventField(event, p.itemType))
+	if !ok {
+		// create an item and return it
+		new, err := NewPcap(event, p.itemType)
+		if err != nil {
+			return nil, utils.ErrorFuncName(err)
+		}
+		p.itemCache.Add(getEventField(event, p.itemType), new)
```
rafaeldtinoco, line 53: command pcap indexes should be `containerID + comm` instead of just `comm` or else only "pcap/commands/host" directory will ever exist (since those are created first and indexed with that name only).

pkg/cmd/flags/capture.go
```diff
@@ -71,12 +76,33 @@ func PrepareCapture(captureSlice []string) (tracee.CaptureConfig, error) {
...
+		} else if cap == "network" {
+			capture.Net.CaptureProcess = true
 		} else if strings.HasPrefix(cap, "net=") {
 			capture.NetIfaces = &tracee.NetIfaces{}
 			err := capture.NetIfaces.Parse(strings.TrimPrefix(cap, "net="))
 			if err != nil {
 				return tracee.CaptureConfig{}, err
 			}
+		} else if strings.HasPrefix(cap, "netpcap:") {
+			capture.Net.CaptureProcess = true // process is default capture mode
```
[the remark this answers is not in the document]
rafaeldtinoco: This was done so the capture=network is considered enabled if the user only provides the netpcap argument. So if --capture network is given only pcap process is enabled. If --capture netpcap:command is given, for example, --capture network isn't even needed.

pkg/pcaps/common.go
rafaeldtinoco: Fixed! Nice catch!

docs/docs/capturing/index.md
```diff
@@ -116,6 +116,120 @@ Tracee can capture the following types of artifacts:
...
+     pcap
+     pcap/processes
+     pcap/processes/host
+     pcap/processes/host/node_186708_1573567360495399.pcap
+     pcap/processes/host/node_1196826_1662656211119567.pcap
+     pcap/processes/host/zerotier-one_7882_137007714376.pcap
+     pcap/processes/host/sshd_1196773_1662654999660718.pcap
+     ```
+
+     You can also elect how the pcap files will be divided:
```
yanivagman, line 143: elect->select

yanivagman (approved):
LGTM

PR #2550, opened 2023-01-06
tracee: refactor signals done channel

josedonizetti:
To process signals we use a done channel which is passed down as an argument. But it would be best to use a Context. To fix it we are using `signal.NotifyContext` which returns a context that will be closed in case the expected signals are sent.

This refactor was discussed on https://github.com/aquasecurity/tracee/issues/2355

Task: https://github.com/aquasecurity/tracee/issues/2355

### DIFF
--- a/cmd/tracee-ebpf/main.go
+++ b/cmd/tracee-ebpf/main.go
@@ -3,6 +3,8 @@ package main
 import (
 	"context"
 	"os"
+	"os/signal"
+	"syscall"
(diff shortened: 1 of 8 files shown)


NDStrahilevitz (approved):
LGTM, we should run nautilus e2e before merging.

NDStrahilevitz:
Merging after confirming this worked in our e2e.

Issue #2477, opened 2022-12-13
Tracee code cleanup


geyslan:
- [x] I'll create a PR to implement this refactoring idea (assign to yourself).

## Refactoring description

- [x] Capitalize the first letter of log messages
- [x] Remove capitalization of the first letter of errors
- [x] Adjust import blocks order
- [x] Make sure errors are being checked

---

- [x] Split #2555, moving error return logic to its own PR.
  - [x] #2812
  - [x] #2815
  - [x] #2818

PR #2555, opened 2023-01-09
code cleanup

geyslan:
## Initial Checklist

## Description (git log)

65d29bf5 cleanup: check error returns
310d5f94 tool: add goimports-reviser
be1e8207 cleanup: import package by groups
8e826109 tool: add errcheck
0b8ac034 cleanup: check relevant return errors
dba95eb4 cleanup: standardize log and error first letter

Fixes: #2477 

## Type of change

### DIFF
--- a/pkg/bufferdecoder/eventsreader.go
+++ b/pkg/bufferdecoder/eventsreader.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/aquasecurity/libbpfgo/helpers"
 	"github.com/aquasecurity/tracee/pkg/events"
+	"github.com/aquasecurity/tracee/pkg/logger"
 	"github.com/aquasecurity/tracee/types/trace"
 )
 
@@ -351,7 +352,10 @@ func readStringFromBuff(ebpfMsgDecoder *EbpfDecoder) (string, error) {
 	res, err := ReadByteSliceFromBuff(ebpfMsgDecoder, int(size-1)) //last byte is string terminating null
 	defer func() {
 		var dummy int8
-		ebpfMsgDecoder.DecodeInt8(&dummy) // discard last byte which is string terminating null
+		err := ebpfMsgDecoder.DecodeInt8(&dummy) // discard last byte which is string terminating null
+		if err != nil {
+			logger.Warn("Trying to discard last byte", "error", err)
+		}
 	}()
 	if err != nil {
 		return "", fmt.Errorf("error reading string arg: %v", err)
--- a/pkg/bufferdecoder/eventsreader.go
+++ b/pkg/bufferdecoder/eventsreader.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/aquasecurity/libbpfgo/helpers"
+
 	"github.com/aquasecurity/tracee/pkg/events"
 	"github.com/aquasecurity/tracee/pkg/logger"
 	"github.com/aquasecurity/tracee/types/trace"
--- a/pkg/capabilities/capabilities.go
+++ b/pkg/capabilities/capabilities.go
@@ -103,13 +103,13 @@ func (c *Capabilities) initialize(bypass bool) error {
 
 	paranoid, err := getKernelPerfEventParanoidValue()
 	if err != nil {
-		logger.Debug("could not get perf_event_paranoid, assuming highest")
+		logger.Debug("Could not get perf_event_paranoid, assuming highest")
 	}
 
 	if paranoid > 2 {
-		logger.Debug("paranoid: Value in /proc/sys/kernel/perf_event_paranoid is > 2")
-		logger.Debug("paranoid: Tracee needs CAP_SYS_ADMIN instead of CAP_BPF + CAP_PERFMON")
-		logger.Debug("paranoid: To change that behavior set perf_event_paranoid to 2 or less.")
+		logger.Debug("Paranoid: Value in /proc/sys/kernel/perf_event_paranoid is > 2")
+		logger.Debug("Paranoid: Tracee needs CAP_SYS_ADMIN instead of CAP_BPF + CAP_PERFMON")
+		logger.Debug("Paranoid: To change that behavior set perf_event_paranoid to 2 or less.")
 		c.Require(cap.SYS_ADMIN)
 	}
 
@@ -210,7 +210,7 @@ func (c *Capabilities) Requested(cb func() error, values ...cap.Value) error {
 
 	errCb := cb()
 	if errCb != nil {
-		logger.Debug("caps Requested cb func", "error", errCb)
+		logger.Debug("Capabilities Requested cb func", "error", errCb)
 	}
 
 	if !c.bypass {
@@ -306,11 +306,11 @@ func (c *Capabilities) apply(t ringType) error {
 		return err
 	}
 
-	logger.Debug("capabilities change")
+	logger.Debug("Capabilities change")
 
 	for k, v := range c.all {
 		if v[t] {
-			logger.Debug("enabling cap", "cap", k)
+			logger.Debug("Enabling cap", "cap", k)
 		}
 		err = c.have.SetFlag(cap.Effective, v[t], k)
 		if err != nil {
--- a/pkg/capabilities/capabilities.go
+++ b/pkg/capabilities/capabilities.go
@@ -82,7 +82,10 @@ func (c *Capabilities) initialize(bypass bool) error {
 	}
 
 	for c := range c.all {
-		cap.DropBound(c) // drop all capabilities from bound
+		err = cap.DropBound(c) // drop all capabilities from bound
+		if err != nil {
+			logger.Debug("Dropping capability from bound", "cap", c, "error", err)
+		}
 	}
 
 	err = c.setProc()
@@ -92,10 +95,13 @@ func (c *Capabilities) initialize(bypass bool) error {
 
 	// The base for required capabilities (ring1) depends on the following:
 
-	c.Require(
+	err = c.Require(
 		cap.IPC_LOCK,
 		cap.SYS_RESOURCE,
 	)
+	if err != nil {
+		logger.Fatal("Requiring capabilities", "error", err)
+	}
 
 	// Kernels bellow v5.8 do not support cap.BPF + cap.PERFMON (instead of
 	// having to have cap.SYS_ADMIN), nevertheless, some kernels, like RHEL8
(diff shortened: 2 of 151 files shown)


geyslan:
@rafaeldtinoco @yanivagman @AlonZivony @NDStrahilevitz

I've inserted new logger calls, please check if they should be Error or Debug for each case. Thanks!

pkg/bufferdecoder/eventsreader.go
```diff
@@ -346,7 +348,10 @@ func readStringFromBuff(ebpfMsgDecoder *EbpfDecoder) (string, error) {
 	res, err := ReadByteSliceFromBuff(ebpfMsgDecoder, int(size-1)) //last byte is string terminating null
 	defer func() {
 		var dummy int8
-		ebpfMsgDecoder.DecodeInt8(&dummy) // discard last byte which is string terminating null
+		err := ebpfMsgDecoder.DecodeInt8(&dummy) // discard last byte which is string terminating null
+		if err != nil {
+			logger.Error("Trying to discard last byte", "error", err)
```
AlonZivony, line 353: I think this should be a debug or warn level, as it is not breaking the string.
If there is a real error here then it will come up with next argument or event.

pkg/capabilities/capabilities.go
```diff
@@ -82,7 +83,10 @@ func (c *Capabilities) initialize(bypass bool) error {
 	}
 
 	for c := range c.all {
-		cap.DropBound(c) // drop all capabilities from bound
+		err = cap.DropBound(c) // drop all capabilities from bound
+		if err != nil {
+			logger.Error("Dropping capability from bound", "cap", c, "error", err)
```
AlonZivony, line 88: This is Debug or even Info level.
It is not critical and expected to fail in many runs.

pkg/capabilities/capabilities.go
```diff
@@ -92,37 +96,52 @@ func (c *Capabilities) initialize(bypass bool) error {
 
 	// The base for required capabilities (ring1) depends on the following:
 
-	c.Require(
+	err = c.Require(
 		cap.IPC_LOCK,
 		cap.SYS_RESOURCE,
 	)
+	if err != nil {
+		logger.Error("Requiring capabilities", "error", err)
```
AlonZivony, line 104: Should be fatal

pkg/capabilities/capabilities.go
```diff
@@ -92,37 +96,52 @@ func (c *Capabilities) initialize(bypass bool) error {
...
-		logger.Debug("paranoid: Value in /proc/sys/kernel/perf_event_paranoid is > 2")
-		logger.Debug("paranoid: Tracee needs CAP_SYS_ADMIN instead of CAP_BPF + CAP_PERFMON")
-		logger.Debug("paranoid: To change that behavior set perf_event_paranoid to 2 or less.")
-		c.Require(cap.SYS_ADMIN)
+		logger.Debug("Paranoid: Value in /proc/sys/kernel/perf_event_paranoid is > 2")
+		logger.Debug("Paranoid: Tracee needs CAP_SYS_ADMIN instead of CAP_BPF + CAP_PERFMON")
+		logger.Debug("Paranoid: To change that behavior set perf_event_paranoid to 2 or less.")
+		err = c.Require(cap.SYS_ADMIN)
+		if err != nil {
+			logger.Error("Requiring capabilities", "error", err)
```
AlonZivony, line 122: Should be fatal

pkg/capabilities/capabilities.go
```diff
@@ -92,37 +96,52 @@ func (c *Capabilities) initialize(bypass bool) error {
...
+		logger.Error("Getting capability flag", "error", err)
+	}
 	if hasBPF {
-		c.Require(
+		err = c.Require(
 			cap.BPF,
 			cap.PERFMON,
 		)
+		if err != nil {
+			logger.Error("Requiring capabilities", "error", err)
```
AlonZivony, line 136: Should be fatal

pkg/capabilities/capabilities.go
```diff
@@ -92,37 +96,52 @@ func (c *Capabilities) initialize(bypass bool) error {
...
+		}
 	} else {
-		c.Require(
+		err = c.Require(
 			cap.SYS_ADMIN,
 		)
+		if err != nil {
+			logger.Error("Requiring capabilities", "error", err)
```
AlonZivony, line 143: Should be fatal.

pkg/cmd/flags/capture.go
```diff
@@ -104,7 +105,10 @@ func PrepareCapture(captureSlice []string) (tracee.CaptureConfig, error) {
 
 	capture.OutputPath = filepath.Join(outDir, "out")
 	if clearDir {
-		os.RemoveAll(capture.OutputPath)
+		err := os.RemoveAll(capture.OutputPath)
+		if err != nil {
+			logger.Error("Removing all", "path", "error", err)
```
AlonZivony, line 110: This might not be a real problem if the files failed to be removed are not colliding with capture ones.
I think it should be a Warn level.
If the capture open file fails it should be a real error.

pkg/ebpf/tracee.go
```diff
@@ -312,7 +314,10 @@ func New(cfg Config) (*Tracee, error) {
 			return t, fmt.Errorf("could not get event")
 		}
 		for _, capArray := range evt.Dependencies.Capabilities {
-			caps.Require(capArray)
+			err = caps.Require(capArray)
+			if err != nil {
+				logger.Debug("Requiring capabilities", "error", err)
```
AlonZivony, line 319: This should be error or even fatal, as it will prevent tracee from behaving as required by the user.

AlonZivony (changes requested):
Didn't review all as this is a large change.
I gave my opinion on code I am more familiar with.
I will try to review what is left soon.

pkg/bufferdecoder/eventsreader.go
geyslan: Warn then.

pkg/capabilities/capabilities.go
geyslan: As expected to fail, Debug.

pkg/ebpf/tracee.go
geyslan: I'm changing them to Fatal but I'm not aborting the logic.

pkg/ebpf/tracee.go
```diff
@@ -478,7 +483,7 @@ func (t *Tracee) Init() error {
 			t.pidsInMntns.AddBucketItem(uint32(mountNS), uint32(pid))
 		}
 	} else {
-		logger.Debug("caps Requested", "error", err)
+		logger.Debug("Requesting capabilities", "error", err)
```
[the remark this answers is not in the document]
geyslan: Good catch.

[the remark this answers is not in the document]
geyslan: @AlonZivony But the `Requested()` can fail before calling `cb` func.

image

So, I think it must be a generic error related to Requested function. The error field should have enough information related to the error.

pkg/ebpf/tracee.go
```diff
@@ -1554,17 +1580,26 @@ func (t *Tracee) updateFileSHA() {
...
 	if t.events[events.InitNamespaces].emit {
-		capabilities.GetInstance().Requested(func() error { // ring2
+		err := capabilities.GetInstance().Requested(func() error { // ring2
 			systemInfoEvent := events.InitNamespacesEvent()
 			t.config.ChanEvents <- systemInfoEvent
 			return nil
 		}, cap.SYS_PTRACE)
-		t.stats.EventCount.Increment()
+		if err != nil {
+			logger.Debug("Requesting capabilities", "error", err)
```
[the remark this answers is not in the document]
geyslan: ditto https://github.com/aquasecurity/tracee/pull/2555#discussion_r1065918297

pkg/ebpf/tracee.go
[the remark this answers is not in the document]
geyslan: Analysing it again, the Requested logic is confuse. It logs the cb error and return it only if !c.bypass doesn't err. I think this can be made clearer in the future. WDYT @rafaeldtinoco?

pkg/ebpf/events_pipeline.go
```diff
@@ -363,7 +369,10 @@ func (t *Tracee) sinkEvents(ctx context.Context, in <-chan *trace.Event) <-chan
 
 			select {
 			case t.config.ChanEvents <- *event:
-				t.stats.EventCount.Increment()
+				err := t.stats.EventCount.Increment()
+				if err != nil {
+					logger.Error("Incrementing events count", "error", err)
```
NDStrahilevitz, line 374: So I did a quick check and running at ~15k events/sec (which is a lot), it would take us ~19498080 years to overflow this, which can be the only error iirc.
So considering the hotpath, i'm wondering if there should:
1. Even be an error here
2. If we keep 1, an error check for a log

And considering this case, which is our quickest rising counter besides `FilteredEvents`, I would say it's even more irrelevant in other counters.

NDStrahilevitz (commented):
LGTM but I think we should get rid of the error counters, they really don't make practical sense.

pkg/ebpf/events_pipeline.go
geyslan: @NDStrahilevitz thanks for digging into this. Besides Increment() and Decrement(), Counter has `Set()`, which was implemented during the process of the aggregated bpf logs PR, in which it would be used just to set the count instead of calculating it per cpu.

`Set()` could easily make a value near to a possible wrap around, but that function is not being used.

Another code that can circumvent straight Increment is `t.stats.BPFLogsCount.Increment(uint64(bpfLog.Count()))` or others that make the Increment using runtime values.

I agree with you that odds are low here and considering the hot paths I would change them to `_ = *.Increment()/Decrement()`, making explicit that it can err and for the  BPFLogsCount I must keep the checking, since received values are out of direct control.

geyslan: Done. 👍🏼

NDStrahilevitz:
@AlonZivony @AsafEitani No practical changes to logging logic here but we should check in the e2e before.

AsafEitani:
> @AlonZivony @AsafEitani No practical changes to logging logic here but we should check in the e2e before.

Sounds good, @geyslan please run the e2e just to make sure, if you need help with that feel free to ping me

geyslan:
> > @AlonZivony @AsafEitani No practical changes to logging logic here but we should check in the e2e before.
> 
> Sounds good, @geyslan please run the e2e just to make sure, if you need help with that feel free to ping me

@AsafEitani all green in the 453 run.

rafaeldtinoco:
I want to review this one before it gets merged.

geyslan:
This will be splitted in two PRs due to possible regressions related to error checking changes. So converting it to Draft.

geyslan:
Closing because a rebase here can be more difficult than a redo.

Issue #2508, opened 2022-12-19
[BUG] tracee release missing go signatures 


idanr1986:
the tracee release does not contain the compiled go signatures as  builtin.so In dist/rules folder

yanivagman:
Yep. we need to include builtin.so in the produced tar.gz artifact

PR #2557, opened 2023-01-10
chore: add go signatures to the tar release

josedonizetti:
Fixes https://github.com/aquasecurity/tracee/issues/2508

### DIFF
--- a/builder/Makefile.release
+++ b/builder/Makefile.release
@@ -157,7 +157,7 @@ release: \
 	$(MAKE) -f builder/Makefile.tracee-make ubuntu-prepare
 	BTFHUB=1 STATIC=1 $(MAKE) -f builder/Makefile.tracee-make ubuntu-make ARG="all"
 	# create the tar ball and checksum files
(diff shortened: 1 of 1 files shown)


builder/Makefile.release
```diff
@@ -157,7 +157,7 @@ release: \
 	$(MAKE) -f builder/Makefile.tracee-make ubuntu-prepare
 	BTFHUB=1 STATIC=1 $(MAKE) -f builder/Makefile.tracee-make ubuntu-make ARG="all"
 	# create the tar ball and checksum files
-	$(CMD_TAR) --exclude="*.so" -czf $(OUT_ARCHIVE) $(RELEASE_FILES)
```
yanivagman, line 160: @rafaeldtinoco maybe you remember - did we have a reason to exclude the .so in first place?

rafaeldtinoco: Yes, we did not include the golang signatures, only the rego ones, before. After we moved from rego to golang signatures by default, we never changed the Makefile.release to reflect that!

rafaeldtinoco (approved):
LGTM.

Issue #2562, opened 2023-01-11
add command arguments and env to captured profile


itaysk:
when we capture execution profile we should add the command arguments and environment to the captured profile. for example:

run 1:
```
ls x
ls x
```

run 2:
```
ls x
ls y
```

these runs should have different profiles but currently produce the same one

yanivagman:
Moved to https://github.com/aquasecurity/tracee-action
Closing this one

PR #2563, opened 2023-01-11
feat: add args, env to profile

itaysk:
Close: #2562


itaysk:
completed in https://github.com/aquasecurity/tracee-action/pull/13

PR #2565, opened 2023-01-11
types/trace: add kernel read type

NDStrahilevitz:


rafaeldtinoco (approved):
I'm okay with the change, but could you reword the git log and add a longer comment as to why the type is being added, etc? Thanks!

Issue #1922, opened 2022-07-05
Network events should also be enriched


rafaeldtinoco:
## Bug description

Enrichment works with:

![image](https://user-images.githubusercontent.com/7395852/177399077-99a4c4b1-dc54-4737-9f30-68e03ad3d3c1.png)

But fails to work when picking "net_packet" only event:

![image](https://user-images.githubusercontent.com/7395852/177399349-7d277098-735b-4c71-9c12-7dbfaccc6297.png)

## Context

Relevant information about my setup:

* Linux version: any
* Linux kernel version: any
* Tracee version (or commit id of your tree): e1a28fe
* LLVM version: any
* Golang version: any

NDStrahilevitz:
I'm not sure this is because of the enrichment events not being enabled (cgroup_mkdir etc should be enabled by default). I think net events just aren't getting enriched since they don't go through the enrichment pipeline. Can you maybe confirm with another singular event.

rafaeldtinoco:
> I'm not sure this is because of the enrichment events not being enabled (cgroup_mkdir etc should be enabled by default). I think net events just aren't getting enriched since they don't go through the enrichment pipeline. Can you maybe confirm with another singular event.
> 
> We can try and add enrichment to the network events for next release, but it's a bit messy because it has its own separate "pipeline".

You're right.

yanivagman:
IMO solution should be to use the same pipeline as a regular event.
We should send network events through the events perf buffer, and modify the pipeline to handle them (need to build the proper context from procinfo, then use the regular pipeline)

yanivagman:
I just want to be clear that when I suggest to use the same perf buffer for network events, I only refer to events, and not the full net packets (pcap) which will continue to use the net perf buffer.

> The advantage to your would be less memory taken for another buffer, but we would have a smaller buffer for more events, so (correct me if i'm wrong) it might cause more event loss.

Another benefit for using the events buffer is event ordering. Different perf buffers (events VS net-events) might have a big offset in their timestamp if there is a big load in one of them. Now we also have different buffers for each CPU, so it's not that big advantage, but as we will move to use ring-buffer, this advantage will become more relevant.

About lost events - there can be more lost events in the events buffer, but then we can just make the buffer bigger. If we keep having two different buffers for network events and "regular" events, then if the rate at which events are produced is too high events will be lost anyway (either network events or "regular" events).

rafaeldtinoco:
[...]

Issue #1828, opened 2022-06-15
Entire clsact qdisc is purged when tc hook is destroyed


rafaeldtinoco:
When refactoring network probes from https://github.com/aquasecurity/tracee/pull/1820 I have realized that we should better document clsact qdisc issue on tc probe destruction. I have documented the behavior observed in the current version to show that refactor is not changing the behavior.

## Prerequisites

- [x] This affects latest released version.
- [x] This affects current development tree (origin/HEAD).
- [x] There isn't an issue describing the bug.

It depends on https://lore.kernel.org/bpf/CAMy7=ZULTCoSCcjxw=MdhaKmNM9DXKc=t7QScf9smKKUB+L_fQ@mail.gmail.com/T/#t already observed by @yanivagman in a near past.

## Bug description

1. Entire clsact qdisc is purged when tc hook is destroyed: it might destroy existing filters (other eBPF programs) unrelated to tracee.
2. eBPF TC detach does not work (destroy needs to be used)

This is how I'm executing tracee (cmdline):

```
$ sudo ./dist/tracee-ebpf -o format:json -o option:parse-arguments -o option:detect-syscall -trace comm=bash -trace follow -trace event=net_packet -trace net=docker0
```

This is the error I'm getting:

In the current version of tracee, we are deleting the qdisc clsact entirely when destroying our tc hook:

<img width="563" alt="image" src="https://user-images.githubusercontent.com/7395852/173924276-2742f156-b1f5-4af5-b627-f7c446b72547.png">

First there is no eBPF program attached to docker0 interface. I leave on purpose 2 eBPF programs attached to it. I execute tracee and, when tc hook is destroyed, it deletes all 4 eBPF programs (and not only the 2 that tracee has created) likely because it destroyed clsact qdisc.

Again, but showing qdiscs now:

<img width="1423" alt="image" src="https://user-images.githubusercontent.com/7395852/173929139-d0039cf4-75b9-4552-a949-6f27c85ad8b5.png">

The second picture is the same thing, but showed in tc filter level.

## Steps to reproduce

Steps to reproduce the issue:

1. Have any other eBPF program hooked into an interface clsact.
2. Execute tracee to trace `event=net_packet` and hook to same network interface with `net=<interface>`
4. Observe all clsact being purged (existing filters) once tracee exits.

## Context

Relevant information about my setup:

* Linux version: any
* Linux kernel version: any
* Tracee version (or commit id of your tree): 6499f067
* LLVM version: any
[...]

PR #2387, opened 2022-11-27
pkg/ebpf: change fork thread start time to be since epoch

AlonZivony:
## Initial Checklist

- [x] There is an issue describing the need for this PR.
- [x] Git log contains summary of the change.
- [x] Git log contains motivation and context of the change.
- [ ] If part of an EPIC, PR git log contains EPIC number.
- [ ] If part of an EPIC, PR was added to EPIC description.

## Description (git log)

The thread start time of the `sched_process_fork` was in monotonic time until now. Change it to time since epoch, so it will be comparable to other times, like what is done for the event timestamp.

Fixes: #2386

## Type of change

- [x] Bug fix (non-breaking change fixing an issue, preferable).
- [ ] Quick fix (minor non-breaking change requiring no issue, use with care)
- [ ] Code refactor (code improvement and/or code removal)
- [ ] New feature (non-breaking change adding functionality).
- [ ] Breaking change (cause existing functionality not to work as expected).

## Final Checklist:

Pick "Bug Fix" or "Feature", delete the other and mark appropriate checks.

- [ ] I have made corresponding changes to the documentation.
- [ ] My code follows the style guidelines (C and Go) of this project.
- [ ] I have performed a self-review of my own code.
- [ ] I have commented all functions/methods created explaining what they do.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] My changes generate no new warnings.
- [ ] I have added tests that prove my fix, or feature, is effective.
- [ ] New and existing unit tests pass locally with my changes.
- [ ] Any dependent changes have been merged and published before.

## Git Log Checklist:

My commits logs have:

- [ ] Subject starts with "subsystem|file: description".
- [ ] Do not end the subject line with a period.
- [ ] Limit the subject line to 50 characters.
- [ ] Separate subject from body with a blank line.
- [ ] Use the imperative mood in the subject line.
- [ ] Wrap the body at 72 characters.
- [ ] Use the body to explain what and why instead of how.

PR #2569, opened 2023-01-11
network: remove old network code based on tc hooks

rafaeldtinoco:
## NOTE: needs https://github.com/aquasecurity/tracee/pull/2547 merged first.

network: remove old network code based on tc hooks

Userland code removal:

1. NetPacket event (now net_packet_xxx events)
1. DnsRequest, DnsResponse events (now net_packet_dns_xxx)
1. capture_pcap event (now Pcaps pkg)
1. NetPacket derivation (now net_packet_xxx derivations)

> Replacing old network logic and events with the new one.
> New logic does not use tc hooks so no interface needs to be specified.

Userland files removal:

1. net_decoder
1. net_proto
1. net_proto_handlers

> net_capture is a parallel pipeline receiving network packets events.
> net_capture now uses the same decodeEvents as the main pipeline.
> net_capture processes received network packets using gopacket.

Userland code actions:

1. move net_packet events to network event id ranges
1. split net_packet events in userland and kernel ranges

> Needed to split network events from the rest.

BPF maps removal:

1. sock_ctx_map
1. network_map

> These maps were responsible for saving network context between
> kprobes, tracepoints and the tc hooks.

BPF probes/tracepoints removal:

1. udp_sendmsg (kprobe)
1. __udp_disconnect (kprobe)
1. udp_destroy_sock (kprobe)
1. udpv6_destroy_sock (kprobe)
1. inet_sock_set_state (tracepoint)
1. tcp_connect (kprobe)
1. icmp_send (kprobe)
1. icmp6_send (kprobe)
1. icmp_recv (kprobe)
1. icmpv6_recv (kprobe)
1. ping_v4_sendmsg (kprobe)
1. ping_v6_sendmsg (kprobe)

> These probes were responsible for updating network maps to work
> in conjunction with tc hooks.

BPF functions removal:

1. net_map_update_or_delete_sock
1. icmp_delete_network_map

> Helper BPF functions

Fixes: #1828
Fixes: #2152
Fixes: #1922
Fixes: #2096
Fixes: #2126


rafaeldtinoco:
@yanivagman I believe @roikol's input is the most important one for this removal. Also, it will likely brake e2e tests because of:

1. cmdline syntax: "--trace net=XXX" does not exist anymore
2. DnsRequest and DnsResponse events do not exist anymore.
3. NetPacket event does not exist anymore.

rafaeldtinoco:
> it will likely brake e2e tests because of

@AsafEitani FYIO

pkg/cmd/flags/capture.go
```diff
@@ -18,20 +18,21 @@ Possible options:
...
 [artifact:]module                  capture loaded kernel modules.
 [artifact:]mem                     capture memory regions that had write+execute (w+x) protection, and then changed to execute (x) only.
-[artifact:]net=interface           capture network traffic of the given interface. Only TCP/UDP/ICMP protocols are currently supported.
+[artifact:]network                 capture network traffic. Only TCP/UDP/ICMP protocols are currently supported.
 
 dir:/path/to/dir                    path where tracee will save produced artifacts. the artifact will be saved into an 'out' subdirectory. (default: /tmp/tracee).
 profile                             creates a runtime profile of program executions and their metadata for forensics use.
 clear-dir                           clear the captured artifacts output dir before starting (default: false).
-pcap:[per-container|per-process]    capture separate pcap file based on container/process context (default: none - saving one pcap for the entire host).
+netpcap:[process,container,command] capture separate pcap files organized by processes, containers and/or commands
```
rafaeldtinoco, line 26: https://github.com/aquasecurity/tracee/pull/2547/files#r1067386220

rafaeldtinoco: also change documentation when this is done (http://localhost:8000/tracee/docs/capturing/)

rafaeldtinoco:
TODO: remove deprecated documentation for capturing pcap.

pkg/cmd/flags/capture.go
```diff
@@ -18,25 +18,21 @@ Possible options:
 [artifact:]exec                    capture executed files.
 [artifact:]module                  capture loaded kernel modules.
 [artifact:]mem                     capture memory regions that had write+execute (w+x) protection, and then changed to execute (x) only.
-[artifact:]net=interface           capture network traffic of the given interface. Only TCP/UDP/ICMP protocols are currently supported.
 [artifact:]network                 capture network traffic. Only TCP/UDP/ICMP protocols are currently supported.
 
 dir:/path/to/dir                    path where tracee will save produced artifacts. the artifact will be saved into an 'out' subdirectory. (default: /tmp/tracee).
 profile                             creates a runtime profile of program executions and their metadata for forensics use.
 clear-dir                           clear the captured artifacts output dir before starting (default: false).
-pcap:[per-container|per-process]    capture separate pcap file based on container/process context (default: none - saving one pcap for the entire host).
```
yanivagman, line 27: Default was previously none, meaning no pcap separation was made when this option wasn't chosen. I think we should stick with that.

pkg/cmd/tracee.go.orig
```diff
@@ -0,0 +1,166 @@
+package cmd
```
yanivagman, line 1: What is this new file for?

pkg/ebpf/events_processor.go
```diff
@@ -13,7 +12,6 @@ import (
 
 	"github.com/aquasecurity/tracee/pkg/events"
 	"github.com/aquasecurity/tracee/pkg/events/parse"
-	"github.com/aquasecurity/tracee/pkg/procinfo"
```
yanivagman, line 16: Procinfo is actually the basis for future enrichments. I think we should keep it for now

pkg/ebpf/events_processor.go.orig
```diff
@@ -0,0 +1,538 @@
+package ebpf
```
yanivagman, line 1: What is this new file used for?

pkg/ebpf/probes/probes.go
```diff
@@ -124,16 +124,6 @@ func Init(module *bpf.Module, netEnabled bool) (Probes, error) {
 		DoMmapRet:                  &traceProbe{eventName: "do_mmap", probeType: kretprobe, programName: "trace_ret_do_mmap"},
 	}
 
-	// disable autoload for tcProbes if network is disabled
```
yanivagman, line 127: We should disable autoload for the new cgroup programs or otherwise CAP_NET_ADMIN will be required even if we don't choose network events.

pkg/ebpf/tracee.go
```diff
@@ -1263,33 +1159,12 @@ func (t *Tracee) attachProbes() error {
...
 }
 
 func (t *Tracee) initBPF() error {
 	var err error
 
-	if t.netEnabled() {
-		// TODO: NET_ADMIN is needed by tcProbes, which are not declared as
-		// events so they don't have required capabilities. Once tcProbes gone
-		// we can remove this from the required capabilities set.
-		capabilities.GetInstance().Require(cap.NET_ADMIN)
```
yanivagman, line 1290: Isn't this required for cgroup programs as well?

yanivagman (changes requested):

pkg/ebpf/events_processor.go.orig
rafaeldtinoco: Leftover from "patch", removing it.

pkg/cmd/tracee.go.orig
rafaeldtinoco: Leftover from "patch", removing it.

pkg/bufferdecoder/eventsreader.go
```diff
@@ -149,7 +149,9 @@ func ReadArgFromBuff(ebpfMsgDecoder *EbpfDecoder, params []trace.ArgMeta) (trace
 		if err != nil {
 			return argMeta, nil, fmt.Errorf("error reading byte array size: %v", err)
 		}
-		if size > 4096 {
+		// error if byte buffer is too big (and not a network event)
+		// TODO: check for network event IDs range isntead of arg name
```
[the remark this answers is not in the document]
rafaeldtinoco: I kept the 2 if you don't mind (--trace net and --trace network do the same thing).

pkg/cmd/flags/capture.go
```diff
@@ -18,23 +18,19 @@ Possible options:
 [artifact:]exec                    capture executed files.
 [artifact:]module                  capture loaded kernel modules.
 [artifact:]mem                     capture memory regions that had write+execute (w+x) protection, and then changed to execute (x) only.
-[artifact:]net=interface           capture network traffic of the given interface. Only TCP/UDP/ICMP protocols are currently supported.
```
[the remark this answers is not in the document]
rafaeldtinoco: I kept the 2 if you don't mind (--trace net and --trace network do the same thing).

pkg/cmd/flags/capture.go
rafaeldtinoco: Yes, it is a TODO (number 3) at https://github.com/aquasecurity/tracee/issues/2152. I'll do in a new PR (since this one is about removing old code).

pkg/ebpf/events_processor.go
rafaeldtinoco: Yep, I had doubts about removing it.

pkg/cmd/flags/capture.go
rafaeldtinoco: added a comment with a TODO (to be soon resolved by me)

rafaeldtinoco: https://github.com/aquasecurity/tracee/issues/2587

pkg/ebpf/tracee.go
rafaeldtinoco: All the network-based events are dependent of a single "network base event":

image

and it sets the NET_ADMIN as a required capability.

The problem we had with tc hooks was that they were not used as "probe dependencies" for the network events so the logic of raising capabilities was not enough.

Now we can use the dependency like it should be used.

pkg/cmd/tracee.go
```diff
@@ -101,7 +101,7 @@ func PrintEventList(printRulesSet bool) {
 	printEventGroup(&b, events.InitNamespaces, events.MaxUserSpace)
 	b.WriteString("\n\nNetwork Events: " + titleHeaderPadFirst + "Sets:" + titleHeaderPadSecond + "Arguments:\n")
 	b.WriteString("____________  " + titleHeaderPadFirst + "____ " + titleHeaderPadSecond + "_________\n\n")
```
rafaeldtinoco, line 103: BTW I'll fix these header pads as they are broken. There is only a single headerpad for multiple headers and the pads don't align. I'll create a very simple table with the appropriate module.

pkg/ebpf/probes/probes.go
rafaeldtinoco: My idea was to get rid of specifics and make things generic enough so that network probes or regular tracing probes were configured the same, specially after the introduction of privilege rings (which even having CAP_NET_ADMIN, that capability won't be effective after initialization either way). 

I guess I can keep the same "if netEnabled() logic disable autoload" and then enable them at tracee probes.Init().

pkg/bufferdecoder/eventsreader.go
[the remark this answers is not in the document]
rafaeldtinoco: Yes, now I have the network event ID range to use, thanks, will update this part.

pkg/ebpf/probes/probes.go
rafaeldtinoco: Okay so CAP_NET_ADMIN is only enabled if user is tracing net events OR capturing network:

```
rafaeldtinoco@fujitsu:~/.../tracee-mycode$ sudo ./dist/tracee-ebpf --log debug --install-path /home/rafaeldtinoco/tracee/tmp --cache cache-type=mem --cache mem-cache-size=512 --output format:json --output option:parse-arguments --output option:detect-syscall --trace comm=ping --capture net --capabilities bypass=false 2>&1 | grep net_admin
{"level":"debug","ts":1673872318.1034772,"msg":"enabling cap","cap":"cap_net_admin","pkg":"capabilities","file":"capabilities.go","line":313}
```
```
rafaeldtinoco@fujitsu:~/.../tracee-mycode$ sudo ./dist/tracee-ebpf --log debug --install-path /home/rafaeldtinoco/tracee/tmp --cache cache-type=mem --cache mem-cache-size=512 --output format:json --output option:parse-arguments --output option:detect-syscall --trace comm=ping --trace event=net_packet_ipv4 --capabilities bypass=false 2>&1 | grep net_admin
{"level":"debug","ts":1673872340.8947747,"msg":"enabling cap","cap":"cap_net_admin","pkg":"capabilities","file":"capabilities.go","line":313}
```
```
rafaeldtinoco@fujitsu:~/.../tracee-mycode$ sudo ./dist/tracee-ebpf --log debug --install-path /home/rafaeldtinoco/tracee/tmp --cache cache-type=mem --cache mem-cache-size=512 --output format:json --output option:parse-arguments --output option:detect-syscall --trace comm=ping --capabilities bypass=false 2>&1 | grep net_admin
```

pkg/cmd/tracee.go
rafaeldtinoco: https://github.com/aquasecurity/tracee/issues/2590

pkg/ebpf/probes/probes.go
yanivagman: > `rafaeldtinoco@fujitsu:~/.../tracee-mycode$ sudo ./dist/tracee-ebpf --log debug --install-path /home/rafaeldtinoco/tracee/tmp --cache cache-type=mem --cache mem-cache-size=512 --output format:json --output option:parse-arguments --output option:detect-syscall --trace comm=ping --capabilities bypass=false 2>&1 | grep net_admin`

I don't think this says that CAP_NET_ADMIN is not used to load the BPF programs, but just that it is not requested.
Try to run tracee without CAP_NET_ADMIN privilege and see if you get an error (I think it will fail while trying to load the cgroup programs)

pkg/cmd/flags/capture.go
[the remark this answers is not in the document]
yanivagman: If you keep both network and net, this should be document here and in other places that refer to that.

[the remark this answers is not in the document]
yanivagman: > If you keep both network and net, this should be documented here and in other places that refer to that.

Ok, I see that this place only lists possible options, so not a must

pkg/ebpf/events_processor.go
```diff
@@ -282,75 +247,6 @@ func (t *Tracee) processSchedProcessExec(event *trace.Event) error {
...
-
-	go t.deleteProcInfoDelayed(event.HostThreadID)
-	return nil
-}
-
-func (t *Tracee) processSchedProcessFork(event *trace.Event) error {
-	err := t.convertArgMonotonicToEpochTime(event, "start_time")
-	if err != nil {
-		return err
-	}
```
yanivagman, line 304: This code is required, unrelated to procinfo

pkg/ebpf/probes/probes.go
yanivagman: Ok, I see that you reintroduced it below, so that should be ok

pkg/bufferdecoder/eventsreader.go
```diff
@@ -150,8 +154,7 @@ func ReadArgFromBuff(ebpfMsgDecoder *EbpfDecoder, params []trace.ArgMeta) (trace
 			return argMeta, nil, fmt.Errorf("error reading byte array size: %v", err)
 		}
 		// error if byte buffer is too big (and not a network event)
-		// TODO: check for network event IDs range isntead of arg name
-		if size > 4096 && argMeta.Name != "payload" {
+		if size > 4096 && id > events.NetPacketBase && id < events.MaxNetID {
```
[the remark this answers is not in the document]
rafaeldtinoco: You're right. Sorry, too many things in parallel, stupid mistake.

pkg/ebpf/events_processor.go
rafaeldtinoco: The only place it was used was within the procinfo.ProcessCtx{} (will be worked together with https://github.com/aquasecurity/tracee/pull/2569#discussion_r1070620792)

rafaeldtinoco: image

rafaeldtinoco: Added a reminder to bring logic back in with the refactor.

rafaeldtinoco:
@geyslan I'm assuming the merge of this one is good as you're rebasing your work on top of this PR, right ? So after @yanivagman gives his +1 I'll go ahead and merge! be advised =D

pkg/ebpf/events_processor.go
yanivagman: This logic was added in this PR: https://github.com/aquasecurity/tracee/pull/2387 and is required. Why do you want to remove it?

pkg/bufferdecoder/eventsreader.go
```diff
@@ -150,8 +154,7 @@ func ReadArgFromBuff(ebpfMsgDecoder *EbpfDecoder, params []trace.ArgMeta) (trace
 			return argMeta, nil, fmt.Errorf("error reading byte array size: %v", err)
 		}
 		// error if byte buffer is too big (and not a network event)
-		// TODO: check for network event IDs range isntead of arg name
-		if size > 4096 && argMeta.Name != "payload" {
+		if size > 4096 && id > events.MaxNetID && id < events.MaxCommonID {
```
yanivagman, line 157: I still think this isn't right.
The condition should probably be:
if size > 4096 && (id < events.NetPacketBase || id > events.MaxNetID) {

pkg/ebpf/events_processor.go
rafaeldtinoco: I missed seeing it had a pointer to Event as an argument, and that it would update the ts of the given argument (thought it was only for the proc info work being done bellow). I have put it back in place.

yanivagman (approved):
LGTM!

PR #2580, opened 2023-01-15
rules: raise privileges when needed if user root

rafaeldtinoco:
## Description (git log)

Until now, after dropping the effective capabilities, tracee-rules wouldn't raise privileges when needed, always erroring when executed as root (instead of doing the correct: dropping all privileges).

This patch fixes this issue.

## How to test

Run tracee-rules with "sudo" and see permission issues when trying to read ./dist/rules directory, for example. After the patch you can run the same command and it will be able to read the directory because of added CAP_DAC_OVERRIDE capability that was raised when needed.

## Type of change

### DIFF
--- a/cmd/tracee-rules/main.go
+++ b/cmd/tracee-rules/main.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/open-policy-agent/opa/compile"
 	"github.com/urfave/cli/v2"
+	"kernel.org/pub/linux/libs/security/libcap/cap"
 )
 
 func init() {
@@ -82,15 +83,20 @@ func main() {
 			}
 
 			var loadedSigIDs []string
-			for _, s := range sigs {
-				m, err := s.GetMetadata()
-				if err != nil {
-					logger.Error("failed to load signature", "error", err)
-					continue
-				}
-				loadedSigIDs = append(loadedSigIDs, m.ID)
-			}
-
+			capabilities.GetInstance().Requested(
+				func() error {
+					for _, s := range sigs {
+						m, err := s.GetMetadata()
+						if err != nil {
+							logger.Error("Failed to load signature", "error", err)
+							continue
+						}
+						loadedSigIDs = append(loadedSigIDs, m.ID)
+					}
+					return nil
+				},
+				cap.DAC_OVERRIDE,
+			)
 			if c.Bool("list-events") {
 				listEvents(os.Stdout, sigs)
 				return nil
--- a/pkg/rules/celsig/config.go
+++ b/pkg/rules/celsig/config.go
@@ -7,8 +7,10 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/aquasecurity/tracee/pkg/capabilities"
 	"github.com/aquasecurity/tracee/types/detect"
 	"gopkg.in/yaml.v2"
+	"kernel.org/pub/linux/libs/security/libcap/cap"
 )
 
 var (
@@ -96,23 +98,33 @@ func NewConfigsFromDir(dirPath string) ([]SignaturesConfig, error) {
 // paths of files with the specified extensions.
 func walkFilesWithExtensions(rootDir string, extensions []string) ([]string, error) {
 	var files []string
-	err := filepath.WalkDir(rootDir, func(path string, d fs.DirEntry, err error) error {
-		if err != nil {
+
+	err := capabilities.GetInstance().Requested(
+		func() error {
+			err := filepath.WalkDir(rootDir,
+				func(path string, d fs.DirEntry, err error) error {
+					if err != nil {
+						return err
+					}
+
+					if d.IsDir() {
+						return nil
+					}
+
+					for _, s := range extensions {
+						if strings.HasSuffix(strings.ToLower(path), "."+s) {
+							files = append(files, path)
+							return nil
+						}
+					}
+
+					return nil
+				})
 			return err
-		}
+		},
+		cap.DAC_OVERRIDE,
+	)
 
-		if d.IsDir() {
-			return nil
-		}
-
-		for _, s := range extensions {
-			if strings.HasSuffix(strings.ToLower(path), "."+s) {
-				files = append(files, path)
-				return nil
-			}
-		}
-
-		return nil
-	})
 	return files, err
+
 }
--- a/pkg/rules/signature/signature.go
+++ b/pkg/rules/signature/signature.go
@@ -12,12 +12,14 @@ import (
 	"plugin"
 	"strings"
 
+	"github.com/aquasecurity/tracee/pkg/capabilities"
 	"github.com/aquasecurity/tracee/pkg/logger"
 
 	embedded "github.com/aquasecurity/tracee"
 	"github.com/aquasecurity/tracee/pkg/rules/celsig"
 	"github.com/aquasecurity/tracee/pkg/rules/regosig"
 	"github.com/aquasecurity/tracee/types/detect"
+	"kernel.org/pub/linux/libs/security/libcap/cap"
 )
 
 func Find(target string, partialEval bool, rulesDir string, rules []string, aioEnabled bool) ([]detect.Signature, error) {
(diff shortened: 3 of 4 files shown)


geyslan:
@geyslan (me) review this.

rafaeldtinoco:
This is a fix and orthogonal to @geyslan multi scopes work. Can be merged after a review.

pkg/rules/signature/signature.go
```diff
@@ -12,7 +12,9 @@ import (
 	"plugin"
 	"strings"
 
+	"github.com/aquasecurity/tracee/pkg/capabilities"
 	"github.com/aquasecurity/tracee/pkg/logger"
+	"kernel.org/pub/linux/libs/security/libcap/cap"
```
geyslan, line 17: Ditto.

pkg/rules/signature/signature.go
```diff
@@ -61,97 +63,109 @@ func Find(target string, partialEval bool, rulesDir string, rules []string, aioE
...
+					if d.IsDir() || d.Name() == "helpers.rego" {
+						return nil
+					}
+					if !isHelper(d.Name()) {
+						return nil
+					}
+					helperCode, err := ioutil.ReadFile(path)
+					if err != nil {
+						logger.Error("reading file " + path + ": " + err.Error())
+						return nil
```
geyslan, line 124: What about returning the error in those WalkDir calls? It already does return in case of err not nil and different of SkipDir.

https://cs.opensource.google/go/go/+/refs/tags/go1.19.5:src/path/filepath/path.go;l=486

pkg/rules/signature/signature.go
```diff
@@ -61,97 +63,109 @@ func Find(target string, partialEval bool, rulesDir string, rules []string, aioE
...
 				}
-			}
-			logger.Error("creating rego signature with: " + string(regoCode[0:newlineOffset]) + ": " + err.Error())
+				if d.IsDir() || !isRegoFile(d.Name()) || isHelper(d.Name()) {
+					return nil
+				}
+				regoCode, err := ioutil.ReadFile(path)
+				if err != nil {
+					logger.Error("reading file " + path + ": " + err.Error())
+					return nil
```
geyslan, line 142: Ditto about the WalkDir question.

pkg/rules/signature/signature.go
```diff
@@ -61,97 +63,109 @@ func Find(target string, partialEval bool, rulesDir string, rules []string, aioE
...
+						codeLength := len(regoCode)
+						if codeLength < 22 {
+							newlineOffset = codeLength
+						} else {
+							newlineOffset = 22
+						}
+					}
+					logger.Error("creating rego signature with: " + string(regoCode[0:newlineOffset]) + ": " + err.Error())
+					return nil
+				}
```
geyslan, line 161: Ditto.

geyslan (changes requested):

cmd/tracee-rules/main.go
```diff
@@ -82,15 +83,20 @@ func main() {
...
+				func() error {
+					for _, s := range sigs {
+						m, err := s.GetMetadata()
+						if err != nil {
+							logger.Error("failed to load signature", "error", err)
+							continue
+						}
+						loadedSigIDs = append(loadedSigIDs, m.ID)
+					}
+					return nil
```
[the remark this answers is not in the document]
rafaeldtinoco: Yep, its right. We can fail to load one signature but be able to load the other, thus the logging error instead of fully failing. Note this was already the behavior, I didn't change it:

```go
for _, s := range sigs {
        m, err := s.GetMetadata()
        if err != nil {
                logger.Error("failed to load signature", "error", err)
                continue
        }
        loadedSigIDs = append(loadedSigIDs, m.ID)
}
```

pkg/rules/celsig/config.go
```diff
@@ -7,8 +7,10 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/aquasecurity/tracee/pkg/capabilities"
 	"github.com/aquasecurity/tracee/types/detect"
 	"gopkg.in/yaml.v2"
+	"kernel.org/pub/linux/libs/security/libcap/cap"
```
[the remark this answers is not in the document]
rafaeldtinoco: Is this a standard? My formatting tools don't do that. I'd rather you to create a way to enforce than me changing it now and saving the file and getting it always changed.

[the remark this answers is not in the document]
rafaeldtinoco:
```go
	"github.com/aquasecurity/tracee/pkg/capabilities"
	"github.com/aquasecurity/tracee/types/detect"
	"gopkg.in/yaml.v2"
```

that is what I automatically get from my linters.

pkg/rules/signature/signature.go
rafaeldtinoco: I did not change that code, I have just wrapped it into a raised privilege callback. I would guess the author didn't want to fully fail if not able to read one rego signature... others would be attempted.

Still, the focus in the PR should be about raising the privilege of already existing code I suppose. We should do other PR for changing existing code IMO.

cmd/tracee-rules/main.go
```diff
@@ -82,15 +83,20 @@ func main() {
...
-				}
-				loadedSigIDs = append(loadedSigIDs, m.ID)
-			}
-
+			capabilities.GetInstance().Requested(
+				func() error {
+					for _, s := range sigs {
+						m, err := s.GetMetadata()
+						if err != nil {
+							logger.Error("failed to load signature", "error", err)
```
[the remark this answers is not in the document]
rafaeldtinoco: done.

cmd/tracee-rules/main.go
```diff
@@ -17,6 +17,7 @@ import (
 	"github.com/aquasecurity/tracee/pkg/rules/engine"
 	"github.com/aquasecurity/tracee/pkg/rules/signature"
 	"github.com/aquasecurity/tracee/types/detect"
+	"kernel.org/pub/linux/libs/security/libcap/cap"
```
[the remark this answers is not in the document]
rafaeldtinoco: done.

cmd/tracee-rules/main.go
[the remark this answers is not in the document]
rafaeldtinoco: done ?

pkg/rules/celsig/config.go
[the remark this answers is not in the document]
rafaeldtinoco: done ?

pkg/rules/signature/signature.go
rafaeldtinoco: out of scope.

rafaeldtinoco: done.

pkg/rules/signature/signature.go
rafaeldtinoco: out of scope.

pkg/rules/signature/signature.go
rafaeldtinoco: out of scope.

pkg/rules/signature/signature.go
rafaeldtinoco: out of scope.

pkg/rules/celsig/config.go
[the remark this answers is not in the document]
NDStrahilevitz: I think normal formatting tools will go a-z and not enforce separate blocks, but that's only based on intuition.

pkg/rules/signature/signature.go
```diff
@@ -61,97 +63,109 @@ func Find(target string, partialEval bool, rulesDir string, rules []string, aioE
...
-				if codeLength < 22 {
-					newlineOffset = codeLength
-				} else {
-					newlineOffset = 22
+	capabilities.GetInstance().Requested(
+		func() error {
+			filepath.WalkDir(dir,
+				func(path string, d fs.DirEntry, err error) error {
+					if err != nil {
+						return err
```
NDStrahilevitz, line 113: Maybe add an error log here as well?

pkg/rules/signature/signature.go
```diff
@@ -61,97 +63,109 @@ func Find(target string, partialEval bool, rulesDir string, rules []string, aioE
...
 func findGoSigs(dir string) ([]detect.Signature, error) {
 	var res []detect.Signature
-	filepath.WalkDir(dir, func(path string, d fs.DirEntry, err error) error {
-		if err != nil {
+	capabilities.GetInstance().Requested(
+		func() error {
+			err := filepath.WalkDir(dir,
+				func(path string, d fs.DirEntry, err error) error {
+					if err != nil {
+						return err
```
NDStrahilevitz, line 71: ditto

NDStrahilevitz (approved):
LGTM, I think you might want to add error logs where I marked but +1 otherwise.

pkg/rules/celsig/config.go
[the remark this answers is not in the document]
geyslan: > ```go
> 	"github.com/aquasecurity/tracee/pkg/capabilities"
> 	"github.com/aquasecurity/tracee/types/detect"
> 	"gopkg.in/yaml.v2"
> 	"kernel.org/pub/linux/libs/security/libcap/cap"
> ```
> 
> that is what I automatically get from my linters.

Yeah, I know. The same happens here. I think `goimports` is the tool we need to tackle this.

geyslan (approved):
LGTM

pkg/rules/signature/signature.go
rafaeldtinoco: Won't hurt (considering we're not changing the behavior).

pkg/rules/signature/signature.go
rafaeldtinoco: Same.

PR #2581, opened 2023-01-15
ebpf: turn has_prefix() to macro using compound expr

rafaeldtinoco:
## Description (git log)

Newer LLVM versions might fail to optimize the has_string() loop unrolling with the following error:

./pkg/ebpf/c/tracee.bpf.c:2487:5: warning: loop not unrolled: the optimizer was unable to perform the requested transformation; the transformation might be disabled or specified as part of an unsupported transformation ordering [-Wpass-failed=transform-warning]

By replacing the has_string() inline function with a macro compound statement, it is possible to mitigate that issue and make the eBPF binary properly optimized.

## Type of change

## How Has This Been Tested?

Compile Tracee with clang-12, clang-13, and clang-14. Check if all tests pass and no compiler warnings happen.

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -2480,10 +2480,40 @@ static __always_inline int events_perf_submit(program_data_t *p, u32 id, long re
 
 // INTERNAL: STRINGS -------------------------------------------------------------------------------
 
+// Workaround: Newer LLVM versions might fail to optimize has_prefix()
+// loop unrolling with the following error:
+//
+//     warning: loop not unrolled: the optimizer was unable to perform
+//     the requested transformation; the transformation might be
+//     disabled or specified as part of an unsupported transformation
+//     ordering
+//
+
+#if defined(__clang__) && __clang_major__ > 13
+
+    #define has_prefix(p, s, n)                                                                    \
+        ({                                                                                         \
+            int rc = 0;                                                                            \
+            char *pre = p, *str = s;                                                               \
+            _Pragma("unroll") for (int z = 0; z < n; pre++, str++, z++)                            \
+            {                                                                                      \
+                if (!*pre) {                                                                       \
+                    rc = 1;                                                                        \
+                    break;                                                                         \
+                } else if (*pre != *str) {                                                         \
+                    rc = 0;                                                                        \
+                    break;                                                                         \
+                }                                                                                  \
+            }                                                                                      \
+            rc;                                                                                    \
+        })
+
+#else
+
 static __inline int has_prefix(char *prefix, char *str, int n)
 {
     int i;
-#pragma unroll
+    #pragma unroll
     for (i = 0; i < n; prefix++, str++, i++) {
         if (!*prefix)
             return 1;
@@ -2495,6 +2525,8 @@ static __inline int has_prefix(char *prefix, char *str, int n)
     // prefix is too long
     return 0;
 }
+
+#endif
 
 // HELPERS: VFS ------------------------------------------------------------------------------------
 


rafaeldtinoco:
@yanivagman 

I was able to make the eBPF code to optimize (unrolled loops) the has_string() function in both, clang-13 and clang-14.

clang-14 had problems understanding an inline function argument as being a defined scalar number (unroll would not happen). The alternative was to create a macro using a compound statement (like if it was a function).

clang-13 did not like the compound statement-based macro with "break" (it would make stack > 512b). Without a "break," it worked.

With that, I preferred to keep things "as they were" for <= clang-13 AND to provide an alternative for fixing clang-14 compilation (a macro that works the same way and does not cause compiler optimization problems).

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2480,10 +2480,38 @@ static __always_inline int events_perf_submit(program_data_t *p, u32 id, long re
 
 // INTERNAL: STRINGS -------------------------------------------------------------------------------
 
+// Workaround: Newer LLVM versions might fail to optimize has_string()
```
yanivagman, line 2483: has_string() -> has_prefix()

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2480,10 +2480,38 @@ static __always_inline int events_perf_submit(program_data_t *p, u32 id, long re
...
+#if defined(__clang__) && __clang_major__ > 13
+
+    #define has_prefix(p, s, n)                                                                    \
+        ({                                                                                         \
+            int rc = 0;                                                                            \
+            char *pre = p, *str = s;                                                               \
+            _Pragma("unroll") for (int z = 0; z < n; pre++, str++, z++)                            \
+            {                                                                                      \
+                if (!*pre) {                                                                       \
+                    rc = 1;                                                                        \
```
yanivagman, line 2501: since you don't return here, won't this cause the loop to continue?
If so, then this won't make what it supposed to do (eventualy the next statement may be true and rc=0 will be returned)

rafaeldtinoco: I missed the break statement. Ended up sending the "test" version (break caused issues for clang < 14). Sending it with breaks now.

yanivagman (approved):
LGTM.
Just rename the commit message to has_prefix as well.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2480,10 +2480,40 @@ static __always_inline int events_perf_submit(program_data_t *p, u32 id, long re
...
+                    break; \
```
yanivagman, line 2502: The "\" char is not aligned

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2480,10 +2480,40 @@ static __always_inline int events_perf_submit(program_data_t *p, u32 id, long re
...
+                } else if (*pre != *str) {                                                         \
+                    rc = 0;                                                                        \
+                    break; \
```
yanivagman, line 2505: here as well

geyslan:
@rafaeldtinoco testing main branch using

```
❯ clang --version
clang version 15.0.7
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

I've faced the same warning.

```shell
./pkg/ebpf/c/tracee.bpf.c:5747:13: warning: loop not unrolled: the optimizer was unable to perform the requested transformation; the transformation might be disabled or specified as part of an unsupported transformation ordering [-Wpass-failed=transform-warning]
        if (has_prefix(
            ^
```

Trying to use the inlined function for the **clang 15**, resulted the same:

`#if defined(__clang__) && __clang_major__ > 13 && __clang_major__ != 15`

```shell
./pkg/ebpf/c/tracee.bpf.c:2653:5: warning: loop not unrolled: the optimizer was unable to perform the requested transformation; the transformation might be disabled or specified as part of an unsupported transformation ordering [-Wpass-failed=transform-warning]
    for (i = 0; i < n; prefix++, str++, i++) {
    ^
```

rafaeldtinoco:
Opened: https://github.com/aquasecurity/tracee/issues/2624 for the issue.

Issue #2392, opened 2022-11-28
Add the `syscall` argument to all LSM hooks


AlonZivony:
Passing the originated syscall of a hook is not very hard to do, so we agreed that we should pass it for all LSM hooks and other internal hook events.

Issue #2389, opened 2022-11-28
Change missing arguments in events to default values


AlonZivony:
Currently, when an event has a missing argument for some reason (error while getting it, configuration, etc.), we pass it without the argument. This cause inconsistency in the event structure, and also make it harder to store the events in DBs.
To fix it, we decided to pass a default value in the argument instead of passing the event without it.
This change will require the following changes:
* https://github.com/aquasecurity/tracee/issues/2261#issuecomment-1369143881
* #2391
* #2392

NDStrahilevitz:
Copying my comment from #2390

> I've had some thoughts about the nil value. If we go with this approach then every argument check (in a helper or directly in 
> signature) will need to include some arg == nil check. This can get rather cumbersome and may not be intuitive.
> It may be more appropriate to go for a strategy like the io.EOF error we have in Read methods, rather than trying to define some alternative value, we should have an explicit method at the struct receiver level that attempts to:
> 
> 1. Query the argument
> 2. Parse it to the correct type (we can use generics for this I believe)
>
> Then we define a constant error that signifies a missing argument which we document near the function so users will know to check for it. This is more explicit and likely to be simpler as users will already check for an error value anyway when extracting an argument from an event.

PR #2575, opened 2023-01-12
events: always detect originated syscall

AlonZivony:
## Initial Checklist

- [x] There is an issue describing the need for this PR.
- [x] Git log contains summary of the change.
- [x] Git log contains motivation and context of the change.
- [ ] If part of an EPIC, PR git log contains EPIC number.
- [ ] If part of an EPIC, PR was added to EPIC description.

## Description (git log)

Always evict the originated syscall argument for internal hooks events.
Remove the `detect-syscall` output option that was used to make it available.
Fix documentation to match new format.

Fixes: #2391 

## Type of change

- [ ] Bug fix (non-breaking change fixing an issue, preferable).
- [ ] Quick fix (minor non-breaking change requiring no issue, use with care)
- [x] Code refactor (code improvement and/or code removal)
- [ ] New feature (non-breaking change adding functionality).
- [x] Breaking change (cause existing functionality not to work as expected).

## Final Checklist:

Pick "Bug Fix" or "Feature", delete the other and mark appropriate checks.

- [ ] I have made corresponding changes to the documentation.
- [ ] My code follows the style guidelines (C and Go) of this project.
- [ ] I have performed a self-review of my own code.
- [ ] I have commented all functions/methods created explaining what they do.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] My changes generate no new warnings.
- [ ] I have added tests that prove my fix, or feature, is effective.
- [ ] New and existing unit tests pass locally with my changes.
- [ ] Any dependent changes have been merged and published before.

## Git Log Checklist:

My commits logs have:

- [ ] Subject starts with "subsystem|file: description".
- [ ] Do not end the subject line with a period.
- [ ] Limit the subject line to 50 characters.
- [ ] Separate subject from body with a blank line.
- [ ] Use the imperative mood in the subject line.
- [ ] Wrap the body at 72 characters.
- [ ] Use the body to explain what and why instead of how.

PR #2591, opened 2023-01-16
Feature/add originated syscall to all events

AlonZivony:
This PR needs https://github.com/aquasecurity/tracee/pull/2575/ merged first.

## Description (git log)

Add to all eBPF events the originated syscall for the event as a
member of the context.
This will allow to always k now what triggered each internal event, and
may allow for syscall events ID to not be connected to syscalls IDs.

This PR is rebased on #2575

Fixes: #2392
Close: #2389


AlonZivony:
This won't pass the e2e test as it breaks the current events API

types/trace/trace.go
```diff
@@ -40,6 +40,7 @@ type Event struct {
 	ReturnValue         int          `json:"returnValue"`
 	StackAddresses      []uint64     `json:"stackAddresses"`
 	ContextFlags        ContextFlags `json:"contextFlags"`
+	OriginatedSyscall   int          `json:"originatedSyscall"`
```
NDStrahilevitz, line 43: I'm conflicted about moving this here but if we go with this then:
1. Please fix the syscall argument filtering and move it to the context filter
2. style: I think this fits slightly better above `ReturnValue` and that just `Syscall` is enough.
3. Previously when `parse-arguments` was used, the syscall would be translated to it's string form, making it readable. This is no longer the case, and will not be present in the table printer.
4. To solve the above pointer, and keep syscalls normalized, I think we should change this to `string` and process it to a syscall name in `processEvent`.

NDStrahilevitz (changes requested):

types/trace/trace.go
rafaeldtinoco: 1. I prefer something else than "OriginatedSyscall"
2. I agree that this should be normalized to a string.

rafaeldtinoco (changes requested):
Per last comment.

AlonZivony:
I put this as a draft for now as I haven't finished all the changes requestd.

pkg/bufferdecoder/decoder.go
```diff
@@ -61,7 +61,7 @@ func (decoder *EbpfDecoder) DecodeContext(ctx *Context) error {
 	_ = copy(ctx.UtsName[:], decoder.buffer[offset+76:offset+92])
 	ctx.Flags = binary.LittleEndian.Uint32(decoder.buffer[offset+92 : offset+96])
 	ctx.EventID = events.ID(int32(binary.LittleEndian.Uint32(decoder.buffer[offset+96 : offset+100])))
-	// offset 100:103 is used for padding
+	ctx.OrigSyscall = binary.LittleEndian.Uint32(decoder.buffer[offset+100 : offset+104])
```
rafaeldtinoco, line 64: This conflicts with @geyslan multiscope work (both are touching the padding). @yanivagman FYIO. Since Geyslan work has priority, I believe you will have to change the approach here.

rafaeldtinoco:
@yanivagman may we move this to next milestone ?

pkg/ebpf/events_pipeline.go
```diff
@@ -227,6 +235,7 @@ func (t *Tracee) decodeEvents(outerCtx context.Context, sourceChan chan []byte)
 				Args:                args,
 				StackAddresses:      StackAddresses,
 				ContextFlags:        parseContextFlags(ctx.Flags),
```
yanivagman, line 237: No need to call `parseContextFlags` again

pkg/events/events.go
```diff
@@ -151,6 +151,15 @@ func (e *eventDefinitions) NamesToIDs() map[string]ID {
 	return namesToIds
 }
 
+func (e *eventDefinitions) IDs32ToIDs() map[ID]ID {
+	IDs32ToIDs := make(map[ID]ID, len(e.events))
+
+	for id, evt := range e.events {
+		IDs32ToIDs[evt.ID32Bit] = id
```
yanivagman, line 158: Should handle the case of undefined ID 32bit

pkg/events/events.go
```diff
@@ -4965,21 +4974,17 @@ var Definitions = eventDefinitions{
...
 		},
 		SysExit: {
 			ID32Bit: sys32undefined,
 			Name:    "sys_exit",
 			Probes: []probeDependency{
 				{Handle: probes.SysExit, Required: true},
 			},
-			Sets: []string{},
-			Params: []trace.ArgMeta{
-				{Type: "int", Name: "syscall"},
```
yanivagman, line 4981: let's keep the syscall argument (the only argument) for sys_enter/exit

yanivagman (changes requested):

yanivagman (approved):
LGTM.
@NDStrahilevitz we can merge after your approval

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -752,7 +752,7 @@ typedef struct event_context {
     u64 ts; // Timestamp
     task_context_t task;
     u32 eventid;
-    u32 padding; // free for further use
+    u32 syscall; // The syscall which triggered the
```
yanivagman, line 755: Triggered the event

pkg/events/parse_args.go
```diff
@@ -75,20 +75,11 @@ func ParseArgs(event *trace.Event) error {
 				}
 			}
 		}
-		if ID(event.EventID) == SecurityFileOpen {
-			if flagsArg := GetArg(event, "flags"); flagsArg != nil {
```
yanivagman, line 79: Why removing this?

AlonZivony: Moved to the case of `open` and such

pkg/ebpf/c/tracee.bpf.c
NDStrahilevitz: Still missing the end there
```suggestion
    u32 syscall; // The syscall which triggered the event
```

pkg/ebpf/events_pipeline.go
```diff
@@ -325,6 +334,30 @@ func parseContextFlags(flags uint32) trace.ContextFlags {
 	}
 }
 
+// Get the syscall name from its ID, taking into account architecture and 32bit/64bit modes
+func parseSyscallID(syscallID int, isCompat bool, compatTranslationMap map[events.ID]events.ID) (string, error) {
```
NDStrahilevitz, line 338: I don't see a case here for event which will not have an originated syscall (for example uprobe events I believe would fit this category), and that it is fine for them not to have such an origin. In that case the return should be `"", nil"`.
I assume when the syscall id passed here is 0 that would be the case, but we have to make sure the bpf code sets it as such.

pkg/events/parse_args.go
```diff
@@ -56,7 +56,7 @@ func ParseArgs(event *trace.Event) error {
 				ParseOrEmptyString(prevProtArg, mmapProtArgument, nil)
 			}
 		}
-	case SysEnter, SysExit, CapCapable, CommitCreds, SecurityFileOpen, TaskRename, SecurityMmapFile, KallsymsLookupName, DoMmap:
+	case SysEnter, SysExit:
```
NDStrahilevitz, line 59: Just a question:
So these events will have `context.Syscall="enter"/"exit"` and their argument will be the actual called syscall right?

pkg/filters/context.go
```diff
@@ -167,7 +169,8 @@ func (filter *eventCtxFilter) Filter(evt trace.Event) bool {
 		filter.podNSFilter.Filter(evt.PodNamespace) &&
 		filter.podUIDFilter.Filter(evt.PodUID) &&
 		filter.tidFilter.Filter(int64(evt.ThreadID)) &&
-		filter.uidFilter.Filter(int64(evt.UserID))
+		filter.uidFilter.Filter(int64(evt.UserID)) &&
+		filter.syscallFilter.Filter(evt.Syscall)
```
NDStrahilevitz, line 173: This should probably be higher in the chain so it will get short circuited faster.

pkg/events/parse_args.go
AlonZivony: No, the argument and the context value will be matching.
We kept it as is because we want them to be printed in the Table Printer.

NDStrahilevitz: Wouldn't it be better to add a Syscall column instead? Actually, now all events with syscall arguments will be missing the syscall in the table printer.

AlonZivony: You are correct, but the `syscall` argument is used in more complicated scenarios so in those cases I think using `json` or `gob` is acceptable.
I think we didn't want to make the Table View more packed as it has too many values to print as it is.

NDStrahilevitz:
@AlonZivony Needs a rebase.

NDStrahilevitz (approved):
LGTM, I will merge once tests are done

PR #2592, opened 2023-01-16
types/trace: add PodSandbox field

NDStrahilevitz:
Add a PodSandbox field to indicate wether the event is coming from a kubernetes sandbox container or not.


NDStrahilevitz:
@rafaeldtinoco @geyslan Can one of you approve this as well (once you're both ok with #2593)

geyslan (approved):
LGTM

Issue #2599, opened 2023-01-17
change pkg/logger to be a standalone package  


geyslan:
Being a standalone package, logger can be used by libbpfgo.

Context:
- https://github.com/aquasecurity/libbpfgo/pull/284#discussion_r1068238401

yanivagman:
@geyslan I believe this is part of the effort for #2417 , right?
If so, no need for a new issue

geyslan:
Could #2417 be an EPIC?

yanivagman:
You mean that you want to keep this issue and link #2417 back here?
I don't think #2417 is that big to make it an epic, but if you prefer keeping this issue, we shuold remove the kind/feature label

geyslan:
Yes, I do prefer to keep this, since after this merge I need to raise other PR to remove the replace in go.mod.

rafaeldtinoco:
> Yes, I do prefer to keep this, since after this merge I need to raise other PR to remove the replace in go.mod.

IMO we should avoid so many issues for so simple things (like replacing go.mod) if you don't mind. We're trying to reduce maintainership burden. Would that be ok ?

geyslan:
> 

Ok, no worries. Let's close this.

Issue #2417, opened 2022-11-30
Print libbpf and libbpfgo logs to logger


AlonZivony:
As part of our effort to get all output of Tracee to be through the new logger,
we also need to think about logs received from the `libbpfgo` library we use, which in turn use `libbpf`.
This library prints its output to `stdout` currently, meaning we miss them when reading the log file of the logger.
There should be a callback to handle `libbpf` prints, so we should probably do the following:
1. Use logger in `libbpfgo`
2. Add callback to `libbpf` that pass logs to the logger

geyslan:
@AlonZivony as soon as I get info about the involved logic I'll return to this.

geyslan:
Deferring this to v0.12.0 as, as discussed, it would be better to change pkg/logger to be a standalone project under tracee or to be a project repository.

- https://github.com/aquasecurity/libbpfgo/pull/284#discussion_r1068238401

PR #2600, opened 2023-01-17
feat: make logger a standalone package

geyslan:
## Initial Checklist

## Description (git log)

commit 992c6091403b4408f4e7aad7fbf4edbba3477aef 

    logger: make logger a standalone package

Fixes:

- #2599

Necessary for:

- #2417

## Type of change

## How Has This Been Tested?

```shell
sudo ./dist/tracee-ebpf -t uid=1000 -t comm=who --log debug
{"level":"debug","ts":1673991173.4891648,"msg":"osinfo","KERNEL_RELEASE":"5.15.85-1-MANJARO","ARCH":"x86_64","PRETTY_NAME":"\"Manjaro Linux\"","ID":"manjaro","ID_LIKE":"arch","BUILD_ID":"rolling","pkg":"urfave","file":"urfave.go","line":53}
{"level":"debug","ts":1673991173.489193,"msg":"RuntimeSockets: failed to register default","socket":"containerd","error":"failed to register runtime socket stat /var/run/containerd/containerd.sock: no such file or directory","pkg":"flags","file":"containers.go","line":45}
{"level":"debug","ts":1673991173.489201,"msg":"RuntimeSockets: failed to register default","socket":"crio","error":"failed to register runtime socket stat /var/run/crio/crio.sock: no such file or directory","pkg":"flags","file":"containers.go","line":45}
{"level":"debug","ts":1673991173.4892054,"msg":"RuntimeSockets: failed to register default","socket":"podman","error":"failed to register runtime socket stat /var/run/podman/podman.sock: no such file or directory","pkg":"flags","file":"containers.go","line":45}
{"level":"debug","ts":1673991173.4893782,"msg":"osinfo","security_lockdown":"none","pkg":"urfave","file":"urfave.go","line":116}
{"level":"debug","ts":1673991173.4915483,"msg":"BTF","bpfenv":false,"btfenv":false,"vmlinux":true,"pkg":"initialize","file":"bpfobject.go","line":40}
{"level":"debug","ts":1673991173.4915576,"msg":"BPF: using embedded BPF object","pkg":"initialize","file":"bpfobject.go","line":69}
{"level":"debug","ts":1673991173.4931793,"msg":"unpacked CO:RE bpf object file into memory","pkg":"initialize","file":"bpfobject.go","line":144}
{"level":"debug","ts":1673991173.4935548,"msg":"Enricher","error":"error registering enricher: unsupported runtime containerd","pkg":"containers","file":"containers.go","line":64}
{"level":"debug","ts":1673991173.4935637,"msg":"Enricher","error":"error registering enricher: unsupported runtime crio","pkg":"containers","file":"containers.go","line":68}
```


go.mod
```diff
@@ -101,3 +103,5 @@ require (
 )
 
 replace github.com/kubernetes/cri-api => k8s.io/cri-api v0.23.5-rc.0
+
+replace github.com/aquasecurity/tracee/logger => ./logger
```
geyslan, line 107: To be removed after merge and `go get`.

rafaeldtinoco (dismissed):
LGTM after a rebase.

NDStrahilevitz (changes requested):
I've discussed with @geyslan that a solution with a logger interface on libbpfgo's side should be attempted before going with the go.mod solution for these reasons (IMO):
1. There are other consumers of libbpfgo besides us. Tying them to our logger from tracee isn't very ok from our end IMO, and I believe it's likely they would set their own logger for libbpfgo if possible (an interface solution would enable that).
2. This would practically introduce a circular dependency of tracee -> libbpfgo -> tracee logger, where they are not circular only _technically_ by the go.mod. The logger we have in tracee is very much intended for tracee's own particular uses (the log aggregation for example).
3. Having another go.mod will make development more burdensome whenever we do a change to the logger, as we all know the `types` module already does. In addition any change in our logger would indirectly affect libbfpgo and may cause unintended behavior.

rafaeldtinoco:
> * There are other consumers of libbpfgo besides us. Tying them to our logger from tracee isn't very ok from our end IMO, and I believe it's likely they would set their own logger for libbpfgo if possible (an interface solution would enable that).

+1

> * This would practically introduce a circular dependency of tracee -> libbpfgo -> tracee logger, where they are not circular only _technically_ by the go.mod. The logger we have in tracee is very much intended for tracee's own particular uses (the log aggregation for example).

+1

> * Having another go.mod will make development more burdensome whenever we do a change to the logger, as we all know the `types` module already does. In addition any change in our logger would indirectly affect libbfpgo and may cause unintended behavior.

+1

geyslan:
As @NDStrahilevitz illuminatingly explained and I agreed, closing this.

Issue #2587, opened 2023-01-16
To have a single pcap file by default for --capture network


rafaeldtinoco:
After the new network code replaced the pcap functionality, the default behavior was changed to having process-related pcap files only captured by default. After quick discussion at https://github.com/aquasecurity/tracee/pull/2569#discussion_r1070542516 it was settled that the default behavior should be a "single pcap file" with everything.

PR #2604, opened 2023-01-19
network: pcap default cap mode to a single file

rafaeldtinoco:
network: pcap default cap mode to a single file

Fixes: #2587

### DIFF
--- a/pkg/cmd/flags/capture.go
+++ b/pkg/cmd/flags/capture.go
@@ -14,24 +14,23 @@ func captureHelp() string {
 Captured artifacts will appear in the 'output-path' directory.
 Possible options:
 
-[artifact:]write[=/path/prefix*]   capture written files. A filter can be given to only capture file writes whose path starts with some prefix (up to 50 characters). Up to 3 filters can be given.
-[artifact:]exec                    capture executed files.
-[artifact:]module                  capture loaded kernel modules.
-[artifact:]mem                     capture memory regions that had write+execute (w+x) protection, and then changed to execute (x) only.
-[artifact:]network                 capture network traffic. Only TCP/UDP/ICMP protocols are currently supported.
+[artifact:]write[=/path/prefix*]         capture written files. A filter can be given to only capture file writes whose path starts with some prefix (up to 50 characters). Up to 3 filters can be given.
+[artifact:]exec                          capture executed files.
+[artifact:]module                        capture loaded kernel modules.
+[artifact:]mem                           capture memory regions that had write+execute (w+x) protection, and then changed to execute (x) only.
+[artifact:]network                       capture network traffic. Only TCP/UDP/ICMP protocols are currently supported.
 
-dir:/path/to/dir                    path where tracee will save produced artifacts. the artifact will be saved into an 'out' subdirectory. (default: /tmp/tracee).
-profile                             creates a runtime profile of program executions and their metadata for forensics use.
-clear-dir                           clear the captured artifacts output dir before starting (default: false).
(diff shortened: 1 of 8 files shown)


pkg/cmd/flags/capture.go
```diff
@@ -14,24 +14,23 @@ func captureHelp() string {
...
+pcap:[single,process,container,command]  capture separate pcap files organized by single file, files per processes, containers and/or commands
 
 Examples:
   --capture exec                                           | capture executed files into the default output directory
   --capture exec --capture dir:/my/dir --capture clear-dir | delete /my/dir/out and then capture executed files into it
   --capture write=/usr/bin/* --capture write=/etc/*        | capture files that were written into anywhere under /usr/bin/ or /etc/
   --capture profile                                        | capture executed files and create a runtime profile in the output directory
-  --capture net                                            | capture network traffic. default: pcap for processes only (check pcap options)
-  --capture network                                        | capture network traffic. default: pcap for processes only (check pcap options)
+  --capture net (or network)                               | capture network traffic. default: single pcap file for every traced packet
```
yanivagman, line 33: for every packet -> containing all traced packets

pkg/cmd/flags/capture.go
```diff
@@ -71,16 +70,20 @@ func PrepareCapture(captureSlice []string) (tracee.CaptureConfig, error) {
...
-			capture.Net.CaptureProcess = true // process is default capture mode
+			found := false
+			capture.Net.CaptureSingle = true // single is set even without --capture net
 			context := strings.TrimPrefix(cap, "pcap:")
 			fields := strings.Split(context, ",")
-			found := false
 			for _, field := range fields {
+				capture.Net.CaptureSingle = false // remove default single pcap mode
+				if field == "single" {
+					found = true // single was requested, enable it at the end
```
yanivagman, line 83: Why do we need this found variable? Can't we simply set `capture.Net.CaptureSingle = true` here?

pkg/pcaps/pcap.go
```diff
@@ -67,6 +73,8 @@ func NewPcap(e *trace.Event, t PcapType) (*Pcap, error) {
 }
 
 func (p *Pcap) write(event *trace.Event, payload []byte) error {
+	p.fileLock.Lock()
```
yanivagman, line 76: Is there more than one writer?

pkg/cmd/flags/capture.go
rafaeldtinoco: This is so if the user executes:

```
tracee-ebpf --capture pcap:XXX
```

without the `--capture network` argument, it will activate the single mode by default, like if `--capture network` argument was given, and only remove the mode if it hasn't been chosen in `XXX`.

But the logic is wrong =D, fixing it.

rafaeldtinoco: Re-checking the logic and testing it.

pkg/pcaps/pcap.go
rafaeldtinoco: I need more tests. I was able to corrupt the pcap file without it. With the locking, I wasn't (ever after transferring 3GB of data and capturing a huge pcap). I wonder if the primitives used for the mutex (barriers) are avoiding a race of some sort (but this code isn't supposed to be parallel).

Checking...

rafaeldtinoco: image

Looking good now. I think I got some situation with a partial flush to the file that might have been done in a later moment due to workload. Will keep an eye on pcap files and check if they're always consistent. Removing the lock for now.

yanivagman (approved):
LGTM

PR #2606, opened 2023-01-19
types: omit container fields if empty

josedonizetti:
Working on the tracee-action, we want to show the user's signatures triggered by the pipeline in a PR comment in JSON. But the JSON gets really noisy with a lot of empty fields related to containers for events that are not container related, so I thought we could omit those fields if empty, not sure though if this would affect other things.

### DIFF
--- a/types/trace/trace.go
+++ b/types/trace/trace.go
@@ -28,13 +28,13 @@ type Event struct {
 	PIDNS               int          `json:"pidNamespace"`
 	ProcessName         string       `json:"processName"`
 	HostName            string       `json:"hostName"`
-	ContainerID         string       `json:"containerId"`
-	ContainerImage      string       `json:"containerImage"`
-	ContainerName       string       `json:"containerName"`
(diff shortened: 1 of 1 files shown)


josedonizetti:
@NDStrahilevitz @yanivagman  WDYT?

NDStrahilevitz (changes requested):
To see if this works you need to point types at the commit or do
`replace github.com/aquasecurity/tracee/types => ./types` in `go.mod`.
Do that, run with our e2e then we can decide.
I'd also rather keep `containerId` even if empty at least so filtering for container/non-container based on that alone would be enough.

rafaeldtinoco:
> Do that, run with our e2e then we can decide.

I thought the e2e tests would mangle the types by themselves (not replacing for the latest, but replacing because of compatibility with builtin.so from the internal rules.

NDStrahilevitz:
I'm doing this change now in #2760, can we close in favor?

Issue #2115, opened 2022-08-27
Sign artifacts with cosign


josedonizetti:
sign artifacts with cosign

PR #2607, opened 2023-01-19
sign container images with cosign

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>

This PR aims to sign tracee container images with cosign using the keyless approach, which avoids the hassle of securely keeping our private keys.

PTAL @itaysk @dlorenc

> Thanks to @marcofranssen for helping me to fix the makefile; I've appealed to his wisdom about makefiles ! 

Fixes: #2115

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -12,6 +12,9 @@ jobs:
   release:
     name: Release
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      id-token: write
     steps:
       - name: Checkout Code
         uses: actions/checkout@v2
@@ -19,7 +22,11 @@ jobs:
           submodules: true
       - name: Install Dependencies
         uses: ./.github/actions/build-dependencies
-      - name: Login to docker.io and github
+      - name: Install Cosign
+        uses: sigstore/cosign-installer@main
+        with:
+          cosign-release: 'v1.13.1'
+      - name: authenticate
         run: |
           gh auth login --with-token <<<'${{ secrets.GITHUB_TOKEN }}'
           docker login docker.io --username ${{ secrets.DOCKERHUB_USER }} --password-stdin <<<'${{ secrets.DOCKERHUB_TOKEN }}'
--- a/builder/Makefile.release
+++ b/builder/Makefile.release
@@ -26,6 +26,7 @@ CMD_GIT ?= git
 CMD_RM ?= rm
 CMD_TOUCH ?= touch
 CMD_MKDIR ?= mkdir
+CMD_COSIGN ?= cosign
 
 .ONESHELL:
 .check_%:
@@ -181,8 +182,12 @@ ifneq ("$(SNAPSHOT)", "1")
 	# push container images to docker hub
 	#
 	$(CMD_DOCKER) push $(PUSH_DOCKER_REPO):latest
+
+	COSIGN_EXPERIMENTAL=1 $(CMD_COSIGN) sign $(PUSH_DOCKER_REPO)@$(shell docker inspect --format='{{index .RepoDigests 0}}' $(PUSH_DOCKER_REPO):latest | awk -F'@' '{print $2}')
 	$(CMD_DOCKER) push $(PUSH_DOCKER_REPO):$(IMAGE_TAG)
+	COSIGN_EXPERIMENTAL=1 $(CMD_COSIGN) sign $(PUSH_DOCKER_REPO)@$(shell docker inspect --format='{{index .RepoDigests 0}}' $(PUSH_DOCKER_REPO):full | awk -F'@' '{print $2}')
 	$(CMD_DOCKER) push $(PUSH_DOCKER_REPO):full
+	
 	$(CMD_DOCKER) push $(PUSH_DOCKER_REPO):full-$(IMAGE_TAG)
 	#
 	# create release notes


itaysk (commented):
Thanks for taking a stab a this! just a preliminary proof read now

.github/workflows/release.yaml
```diff
@@ -6,16 +6,30 @@ on:
...
           gh auth login --with-token <<<'${{ secrets.GITHUB_TOKEN }}'
           docker login docker.io --username ${{ secrets.DOCKERHUB_USER }} --password-stdin <<<'${{ secrets.DOCKERHUB_TOKEN }}'
       - name: release
+        env:
+          COSIGN_EXPERIMENTAL: "true"
         run: |
-          make -f builder/Makefile.release PUSH_DOCKER_REPO=aquasec/tracee
+          #!/bin/sh
+          set -euo pipefail
+          make -f builder/Makefile.release PUSH_DOCKER_REPO=developerguy/tracee
```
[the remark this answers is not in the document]
developer-guy: yeah, I'm going to change it once the pipeline succeed in my fork

.github/workflows/release.yaml
```diff
@@ -6,16 +6,25 @@ on:
...
+      - name: Install crane
+        uses: imjasonh/setup-crane@v0.1
       - name: authenticate
         run: |
           gh auth login --with-token <<<'${{ secrets.GITHUB_TOKEN }}'
           docker login docker.io --username ${{ secrets.DOCKERHUB_USER }} --password-stdin <<<'${{ secrets.DOCKERHUB_TOKEN }}'
       - name: release
         run: |
-          make -f builder/Makefile.release PUSH_DOCKER_REPO=aquasec/tracee
+          make -f builder/Makefile.release PUSH_DOCKER_REPO=developerguy/tracee
```
rafaeldtinoco, line 30: I guess you're testing, right ? Could you move this to draft if you're only testing ? Thanks!

developer-guy: we are ready, https://github.com/developer-guy/tracee/actions/runs/4026850058/jobs/6921861268

developer-guy: we are ready https://github.com/developer-guy/tracee/actions/runs/4027024574/jobs/6922256638

developer-guy:
```shell
$ for i in latest 999.999.999 full full-999.999.999
  do
    COSIGN_EXPERIMENTAL=1 cosign verify developerguy/tracee:$i
  done
```

builder/Makefile.release
```diff
@@ -186,3 +187,20 @@ endif
 clean:
 #
 	$(MAKE) clean
+
+TAGS := latest $(IMAGE_TAG) full full-$(IMAGE_TAG)
+
+define sign_rule
+.PHONY: $1-sign
+$1-sign:
+	COSIGN_EXPERIMENTAL=1 $(CMD_COSIGN) sign $(PUSH_DOCKER_REPO)@$(shell crane digest $(PUSH_DOCKER_REPO):$1)
```
itaysk, line 196: can we sign the tag instead of the digest? this way we don't need crane dependency and it's simpler

developer-guy: We should sign the digest, not the tag itself.
> https://github.com/sigstore/cosign#sign-a-container-and-store-the-signature-in-the-registry

builder/Makefile.release
```diff
@@ -186,3 +187,20 @@ endif
 clean:
 #
 	$(MAKE) clean
+
+TAGS := latest $(IMAGE_TAG) full full-$(IMAGE_TAG)
```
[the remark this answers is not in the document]
developer-guy: Yep, you are right. I preferred this because I don't want to duplicate the `cosign sign` command for each build. But if you are okay with that, I can add the `cosign sign` command right after each build.

builder/Makefile.release
itaysk: I understand the recommendation, but not sure it applies in our case, since we are using the tag to find the digest anyway. If there's ambiguity in the tag, we will sign the wrong image in both cases (please correct me if I'm wrong). So in this case I think we can just sign the tag in the first place

developer-guy: A container image tag is a human-readable string that is associated with a specific version of the image. It is typically in the format of "repository:tag" and is used to identify the image in a registry. For example, an image could have a tag of "nginx:1.19.2" to indicate that it is version 1.19.2 of the nginx image.

A container image digest, on the other hand, is a unique and immutable identifier generated from the image's content. It is a cryptographic hash of the image's metadata and layers. The digest is used to verify the integrity of the image and to ensure that it has not been tampered with. For example, an image could have a digest of "sha256:abcdef123456789".

In general, using the image digest instead of the image tag when pulling images from a registry is recommended, the same applies to signing images, as it provides a stronger guarantee of the image's content. The tag can change for a particular image, for example, when a new version is pushed to the registry, whereas the digest remains the same for a specific version of the image.

> _https://cloud.google.com/kubernetes-engine/docs/archive/using-container-images_

yanivagman: I also think we should use image digest and not tag.
@itaysk is your only concern is the crane dependancy?
Can't we then simply use docker inspect to get the image digest instead of crane?

developer-guy: This would work yeah, and I updated the code by removing the crane and using the docker inspect

itaysk: I just didn't see the value of signing digest of tag in this specific case (i understand the value generally). anyway, it shouldn't block the pr

developer-guy: we are ready for merge then 🙉

josedonizetti:
@developer-guy Thank you for working on this! ❤️

PR #2609, opened 2023-01-20
chore: add MatchedScopes to Event

geyslan:
MatchedScopes is a requirement for the multiscope effort #2238.

Since types is a different module this needs to enter first to make e2e happy.


yanivagman (approved):
LGTM

PR #2616, opened 2023-01-23
ebpf: fix progs from kthread ctx might preempt work

rafaeldtinoco:
commit 4ef3b231
Date:   Tue Jan 24 01:23:28 2023

    ebpf: fix network events corruption
    
    Do not share scratch buffer (from array maps) between network and regular
    probes. This solves problems such as:
    
      {"level":"error","ts":1674523484.3078973,"msg":"tracee encountered an
      error","error":"failed to read argument 0 of event net_packet_capture: error
      reading byte array: can't read context from buffer: buffer too short"}
      {"level":"debug","ts":1674523484.3081043,"msg":"network capture: no payload
      packet","pkg":"ebpf","file":"net_capture.go","line":89}
    
    Note: This change also allows init_program_data() to use event_data and scratch
    area from the stack of the caller, not only from an array item from a global
    map.
    
    Note: It is not clear yet why this change solves the network events corruptions.
    It seems that sharing array maps with programs from other probes was causing the
    issue, even if using another index from the array maps.

commit 59c90082
Date:   Mon Jan 23 13:43:12 2023

    ebpf: fix progs from kthread ctx might preempt work
    
    Research team has found out that ebpf programs executed from hooks that might
    have kthread context could preempt regular probes, causing singled item buffers
    (used as scratch area to reduce stack usage) to be corrupted.
    
    Adding a second array item to the per cpu maps mitigate the issue: kthread
    ctx'ed programs can now use another item in the array map as its scratch area.
    
    This fixes issues such as:
    
      {"level":"error","ts":1674451314.9671354,"msg":"tracee encountered an
      error","error":"failed to read argument 1 of event openat: string size too
      big: 33554436"}
    
      {"level":"error","ts":1674451315.9101224,"msg":"tracee encountered an
      error","error":"failed to read argument 1 of event openat: string size too
      big: 1668248176"}
    
      ...

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -873,6 +873,7 @@ typedef struct program_data {
     task_info_t *task_info;
     event_data_t *event;
     void *ctx;
+    bool kthread; // this is kernel thread (bottom halves: e.g. network ctx)
 } program_data_t;
 
 // For a good summary about capabilities, see https://lwn.net/Articles/636533/
@@ -1057,9 +1058,9 @@ BPF_HASH(module_init_map, u32, kmod_data_t, 256);                  // holds modu
 BPF_LRU_HASH(fd_arg_path_map, fd_arg_task_t, fd_arg_path_t, 1024); // store fds paths by task
 BPF_LRU_HASH(bpf_attach_map, u32, bpf_attach_t, 1024);             // holds bpf prog info
 BPF_LRU_HASH(bpf_attach_tmp_map, u32, bpf_attach_t, 1024);         // temporarily hold bpf_attach_t
-BPF_PERCPU_ARRAY(event_data_map, event_data_t, 1);                 // persist event related data
+BPF_PERCPU_ARRAY(event_data_map, event_data_t, 2);                 // persist event related data
 BPF_HASH(logs_count, bpf_log_t, bpf_log_count_t, 4096);            // logs count
-BPF_PERCPU_ARRAY(scratch_map, scratch_t, 1);                       // scratch space to avoid allocating stuff on the stack
+BPF_PERCPU_ARRAY(scratch_map, scratch_t, 2);                       // scratch space to avoid allocating stuff on the stack
 // clang-format on
 
 // EBPF PERF BUFFERS -------------------------------------------------------------------------------
@@ -1853,10 +1854,11 @@ init_context(void *ctx, event_context_t *context, struct task_struct *task, u32 
     return 0;
 }
 
-static __always_inline task_info_t *init_task_info(u32 tid, u32 pid)
+static __always_inline task_info_t *init_task_info(u32 tid, u32 pid, bool kthread)
 {
-    int zero = 0;
-    scratch_t *scratch = bpf_map_lookup_elem(&scratch_map, &zero);
+    int scratch_map_idx = kthread ? 1 : 0; // second buffer for kthread ctx
+
+    scratch_t *scratch = bpf_map_lookup_elem(&scratch_map, &scratch_map_idx);
     if (unlikely(scratch == NULL))
         return NULL;
 
@@ -1888,12 +1890,14 @@ static __always_inline bool context_changed(task_context_t *old, task_context_t 
            *(u64 *) &old->uts_name[8] != *(u64 *) &new->uts_name[8];
 }
 
+// clang-format off
 static __always_inline int init_program_data(program_data_t *p, void *ctx)
 {
     long ret = 0;
     int zero = 0;
+    int evt_buf_idx = p->kthread ? 1 : 0; // second buffer for kthread ctx
 
-    p->event = bpf_map_lookup_elem(&event_data_map, &zero);
+    p->event = bpf_map_lookup_elem(&event_data_map, &evt_buf_idx);
     if (unlikely(p->event == NULL))
         return 0;
 
@@ -1913,14 +1917,18 @@ static __always_inline int init_program_data(program_data_t *p, void *ctx)
     p->event->buf_off = 0;
 
     bool container_lookup_required = true;
+
     p->task_info = bpf_map_lookup_elem(&task_info_map, &p->event->context.task.host_tid);
     if (unlikely(p->task_info == NULL)) {
-        p->task_info =
-            init_task_info(p->event->context.task.host_tid, p->event->context.task.host_pid);
+        p->task_info = init_task_info(
+            p->event->context.task.host_tid,
+            p->event->context.task.host_pid,
+            p->kthread
+        );
         if (unlikely(p->task_info == NULL)) {
             return 0;
         }
-        // we just initialized task info so we know that recompute_scope is already set to true
+        // just initialized task info: recompute_scope is already set to true
         goto out;
     }
 
@@ -1928,21 +1936,27 @@ static __always_inline int init_program_data(program_data_t *p, void *ctx)
     if (context_changed(&p->task_info->context, &p->event->context.task))
         p->task_info->recompute_scope = true;
 
-    // If task is already part of a container, there is no need to check if state changed
     u8 container_state = p->task_info->container_state;
-    if (container_state == CONTAINER_STARTED || container_state == CONTAINER_EXISTED) {
-        p->event->context.task.flags |= CONTAINER_STARTED_FLAG;
-        container_lookup_required = false;
+
+    // if task is already part of a container: no need to check if state changed
+    switch (container_state) {
+        case CONTAINER_STARTED:
+        case CONTAINER_EXISTED:
+            p->event->context.task.flags |= CONTAINER_STARTED_FLAG;
+            container_lookup_required = false;
     }
 
 out:
     if (container_lookup_required) {
         u32 cgroup_id_lsb = p->event->context.task.cgroup_id;
         u8 *state = bpf_map_lookup_elem(&containers_map, &cgroup_id_lsb);
+
         if (state != NULL) {
             p->task_info->container_state = *state;
-            if (*state == CONTAINER_STARTED || *state == CONTAINER_EXISTED) {
-                p->event->context.task.flags |= CONTAINER_STARTED_FLAG;
+            switch (*state) {
+                case CONTAINER_STARTED:
+                case CONTAINER_EXISTED:
+                    p->event->context.task.flags |= CONTAINER_STARTED_FLAG;
             }
         }
     }
@@ -1952,6 +1966,7 @@ out:
 
     return 1;
 }
+// clang-format on
 
 static __always_inline int init_tailcall_program_data(program_data_t *p, void *ctx)
 {
@@ -3207,7 +3222,7 @@ int sys_enter_init(struct bpf_raw_tracepoint_args *ctx)
     task_info_t *task_info = bpf_map_lookup_elem(&task_info_map, &tid);
     if (unlikely(task_info == NULL)) {
         u32 pid = pid_tgid >> 32;
-        task_info = init_task_info(tid, pid);
+        task_info = init_task_info(tid, pid, false);
         if (unlikely(task_info == NULL)) {
             return 0;
         }
@@ -3351,7 +3366,7 @@ int sys_exit_init(struct bpf_raw_tracepoint_args *ctx)
     task_info_t *task_info = bpf_map_lookup_elem(&task_info_map, &tid);
     if (unlikely(task_info == NULL)) {
         u32 pid = pid_tgid >> 32;
-        task_info = init_task_info(tid, pid);
+        task_info = init_task_info(tid, pid, false);
         if (unlikely(task_info == NULL)) {
             return 0;
         }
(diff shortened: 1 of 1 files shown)


pkg/ebpf/c/tracee.bpf.c
```diff
@@ -6884,7 +6904,7 @@ typedef struct net_event_context {
 
 struct {
     __uint(type, BPF_MAP_TYPE_LRU_HASH);
-    __uint(max_entries, 1024);          // simultaneous cgroup/skb ingress/eggress progs
+    __uint(max_entries, 10240);         // simultaneous cgroup/skb ingress/eggress progs
```
rafaeldtinoco, line 6907: Not sure if raising this is needed (theoretically this is # of CPUs).

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1888,12 +1891,18 @@ static __always_inline bool context_changed(task_context_t *old, task_context_t
...
+static __always_inline int init_program_data(program_data_t *p, void *ctx, bool kthread)
 {
     long ret = 0;
     int zero = 0;
 
-    p->event = bpf_map_lookup_elem(&event_data_map, &zero);
+    // kthreads can preempt probe handlers & corrupt maps: use another buffer
+    int evt_buf_idx = 0;
+    if (kthread)
+        evt_buf_idx = 1;
```
rafaeldtinoco, line 1903: int evt_buf_idx = kthread ? 1 : 0; // more elegant.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1853,10 +1853,13 @@ init_context(void *ctx, event_context_t *context, struct task_struct *task, u32
     return 0;
 }
 
-static __always_inline task_info_t *init_task_info(u32 tid, u32 pid)
+static __always_inline task_info_t *init_task_info(u32 tid, u32 pid, bool kthread)
 {
-    int zero = 0;
-    scratch_t *scratch = bpf_map_lookup_elem(&scratch_map, &zero);
+    int scratch_map_idx = 0;
```
rafaeldtinoco, line 1858: ditto

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1057,9 +1057,9 @@ BPF_HASH(module_init_map, u32, kmod_data_t, 256);                  // holds modu
 BPF_LRU_HASH(fd_arg_path_map, fd_arg_task_t, fd_arg_path_t, 1024); // store fds paths by task
 BPF_LRU_HASH(bpf_attach_map, u32, bpf_attach_t, 1024);             // holds bpf prog info
 BPF_LRU_HASH(bpf_attach_tmp_map, u32, bpf_attach_t, 1024);         // temporarily hold bpf_attach_t
-BPF_PERCPU_ARRAY(event_data_map, event_data_t, 1);                 // persist event related data
+BPF_PERCPU_ARRAY(event_data_map, event_data_t, 2);                 // persist event related data
```
AlonZivony, line 1060: We will need to think of a solution for the `bufs` map as well, so path strings for example will not collide

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1904,7 +1907,7 @@ static __always_inline int init_program_data(program_data_t *p, void *ctx)
     p->event->task = (struct task_struct *) bpf_get_current_task();
     ret = init_context(ctx, &p->event->context, p->event->task, p->config->options);
     if (unlikely(ret < 0)) {
-        // disable logging as a workaround for instruction limit verifier error on kernel 4.19
+        // disable logging: workaround for inst limit verifier error on kernel 4.19
```
AlonZivony, line 1910: I don't think shortening 'instruction' to 'inst' make it more readable, but less.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1913,36 +1916,46 @@ static __always_inline int init_program_data(program_data_t *p, void *ctx)
...
         p->task_info->recompute_scope = true;
 
-    // If task is already part of a container, there is no need to check if state changed
     u8 container_state = p->task_info->container_state;
-    if (container_state == CONTAINER_STARTED || container_state == CONTAINER_EXISTED) {
-        p->event->context.task.flags |= CONTAINER_STARTED_FLAG;
-        container_lookup_required = false;
+
+    // if task is already part of a container: no need to check if state changed
+    switch (container_state) {
```
AlonZivony, line 1941: Are you sure that this is better?
I don't think we will have another case in the future, and I think switch is a bit less performant efficient than if-else.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1913,36 +1916,46 @@ static __always_inline int init_program_data(program_data_t *p, void *ctx)
...
 out:
     if (container_lookup_required) {
         u32 cgroup_id_lsb = p->event->context.task.cgroup_id;
         u8 *state = bpf_map_lookup_elem(&containers_map, &cgroup_id_lsb);
+
         if (state != NULL) {
             p->task_info->container_state = *state;
-            if (*state == CONTAINER_STARTED || *state == CONTAINER_EXISTED) {
-                p->event->context.task.flags |= CONTAINER_STARTED_FLAG;
+            switch (*state) {
```
AlonZivony, line 1955: ditto

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1952,6 +1965,7 @@ static __always_inline int init_program_data(program_data_t *p, void *ctx)
 
     return 1;
 }
+// clang-format on
```
AlonZivony, line 1968: ?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -6884,7 +6898,7 @@ typedef struct net_event_context {
 
 struct {
     __uint(type, BPF_MAP_TYPE_LRU_HASH);
-    __uint(max_entries, 1024);          // simultaneous cgroup/skb ingress/eggress progs
+    __uint(max_entries, 1240);          // simultaneous cgroup/skb ingress/eggress progs
```
AlonZivony, line 6901: Why are you increasing the size here?

AlonZivony (changes requested):
Seems good overall.
I don't really get why you use the term `kthread` here.
We want to identify probable interrupt-driven events, which might interrupt other events and overwrite per-CPU buffers. I think a better term would be `irq_event` or something similar.

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: keep functions lesser than 100 cols without splitting in an unreadable way (use the stanza when you're sure the formatting is better than the auto formatted one).

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: nope, typo.

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: image

same optimization for both cases, switch causes shorter lines and it is usually better for readability (based on my linters and the picture above).

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1853,10 +1853,11 @@ init_context(void *ctx, event_context_t *context, struct task_struct *task, u32
     return 0;
 }
 
-static __always_inline task_info_t *init_task_info(u32 tid, u32 pid)
+static __always_inline task_info_t *init_task_info(u32 tid, u32 pid, bool kthread)
 {
-    int zero = 0;
-    scratch_t *scratch = bpf_map_lookup_elem(&scratch_map, &zero);
+    int scratch_map_idx = kthread ? 1 : 0; // second buffer for kthread ctx
```
yanivagman, line 1858: Not necessarily a kthread context I believe

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1888,12 +1889,14 @@ static __always_inline bool context_changed(task_context_t *old, task_context_t
            *(u64 *) &old->uts_name[8] != *(u64 *) &new->uts_name[8];
 }
 
-static __always_inline int init_program_data(program_data_t *p, void *ctx)
+// clang-format off
+static __always_inline int init_program_data(program_data_t *p, void *ctx, bool kthread)
```
yanivagman, line 1893: Instead of adding a new argument to this function and changing all references, you can add a new field to program_data_t which will tell the buffer index to use

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: discussing in the previous comment.

yanivagman (changes requested):

rafaeldtinoco:
> I don't really get why you use the term `kthread` here.

In newer kernels, skb ingress programs (and its callers, which I'm probing to) comes from the "ksoftirqd" kthread, [ksoftirqd/cpu#] (or sometimes [idle/cpu#]). It is the kthread responsible for running delayed work from the irq context (top halves) I believe.

If wording isn't good I can change for sure.

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: https://github.com/aquasecurity/tracee/pull/2616#issuecomment-1400636143

rafaeldtinoco:
> In newer kernels, skb ingress programs (and its callers, which I'm probing to) comes from the "ksoftirqd" kthread, [ksoftirqd/cpu#] (or sometimes [idle/cpu#]). It is the kthread responsible for running delayed work from the irq context (top halves) I believe.

That leads to other question also... ksoftirqs run outside the interrupt context, as regular kernel threads, as bottom halves... so they shouldn't be causing the corruptions. I wonder which execution flow to __cgroup_bpf_run_filter_skb() came from an interrupt handler... couldn't find.

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: I'm adding a TODO (since the network code, the only one we know for sure is causing eBPF code preemption) does not use bufs map.

https://github.com/aquasecurity/tracee/issues/2618

rafaeldtinoco:
Okay so here is one thing (that initially led me to think of having a simplified `init_program_data` function for network probes that might come from softirqs (instead of the multiple strach area buffers).

With my reproducer, without this fix:

image

It is clear that work from ksoftirqd, sharing same event context scratch buffer with regular probes, causes regular events contexts to be corrupted. 

Now, with this fix:

image

We only fix the "regular events" contexts, but not the network events contexts.

This means that the ksoftirq are preempting each other as well (as expected). This might require that the network code uses local event context (instead just another simple entry in the map array).

Fixing it.

AlonZivony:
Wow, I was sure that softirq should not preemt each other.
This is gonna be a tough one...

rafaeldtinoco:
> This is gonna be a tough one...

I think I got it. I'm allowing callers of init_program_data to specify their own "event_data_t" and "scratch_t", so they can specify from their own stack (or from another map). If not null, use the ones provided, if null, use the existing maps (single entry &zero). Lets see..

rafaeldtinoco:
> Wow, I was sure that softirq should not preemt each other.

Right ? Its weird to me as well, Im still testing and investigating.

yanivagman:
> > Wow, I was sure that softirq should not preemt each other.
> 
> Right ? Its weird to me as well, Im still testing and investigating.

Actually it makes sense. We saw that perf_event_output() causes softirqs to be called after its invocation. Possibly other bpf helpers also cause softirqs to be called. If that's the case, this code can be reentrant.

rafaeldtinoco:
Okay, I was able to succesfully capture tons of GBs in pcap files without having issues with either the regular events or network events. It seems that this last was good enough without bigger changes to already stable code.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -7148,6 +7164,21 @@ int BPF_KPROBE(cgroup_bpf_run_filter_skb)
             return 0; // wrong attachment type, return fast
     }
 
+    u32 zero = 0;
+    event_data_t *e = bpf_map_lookup_elem(&net_heap_event, &zero);
```
rafaeldtinoco, line 7168: Can't use stack due to size limits, unfortunately.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -6940,6 +6945,22 @@ struct {
     __type(value, u32);
 } net_cap_events SEC(".maps");
 
+// scratch area
+
+struct {
+    __uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+    __uint(max_entries, 1);                    // simultaneous softirqs running per CPU (?)
```
rafaeldtinoco, line 6952: I'm not sure if a softirq, running right after a top half, that just preempted the bpf program, can be preempted by another softirq. For now, simply using another map for the network events scratch, was enough to avoid corruption.

yanivagman (approved):
LGTM!

Issue #2261, opened 2022-10-20
[BUG] Arguments may be missing if submission fails


NDStrahilevitz:
## Bug description

This is the general case of #2260.
In the bpf helper functions `save_to_submit_buf`, `save_bytes_to_buf`, `save_str_to_buf`, etc. the field `data->context.argnum` is not incremented on failure cases.
As such arguments can fail to be submitted and will be overriden, causing them to not exist in the event itself.

## Steps to reproduce

Simple reproduction can be done by doing this modificaiton
<img width="532" alt="image" src="https://user-images.githubusercontent.com/22661609/196957641-58c4ede6-2d66-4020-8f6d-ff2d68edfc81.png">
Then run `tracee-ebpf -t e=security_file_open -o option:detect-syscall` and note how both `pathname` and `syscall_pathname` are entirely missing from the arguments.

## Context

Relevant information about my setup:

* Linux version:
* Linux kernel version:
* Tracee version (or commit id of your tree):
* LLVM version:
* Golang version:

NDStrahilevitz:
@AsafEitani @yanivagman @rafaeldtinoco I think we need this before our CNDR release as it's causing issues with signature testing.

NDStrahilevitz:
> Do you know the sentence "it's not a bug, it's a feature"? So this is the case here :-)

Definitely not an intuitive reasoning here, since the event "contract" is broken. while I understand the reasoning, i'd say this is more a designed bug than a feature.

> We don't want to increment the argument number if we failed to save them in bpf code for any reason. This helps us to send the event to userspace even in cases where something fails (e.g. when failing to read some string from userspace memory, which is controllable by the userspace, thus can be made to fail on purpose).

If we fail to save something that means we need to treat a failure not ignore that it happened. wether that's by marking the argument for fixing or setting a default value then and there, it's a "breach of contract" of the event.

> Anyway, we may consider adding some default value to these arguments instead of simply removing them, but this will probably need to happen in userspace and not in the bpf code.

I think we will need to do this, AND mark optional arguments as such, for example the "syscall" arguments which are appended only for some CLI settings.

> Rules need to be ready to expect a non existing argument, and not fail tracee-rules in any case.
I don't really agree here, but we already have this issue with the optional "detect-syscall" which is practically required in our signatures. This expects rules to be aware of how tracee-ebpf may internally fail or succeed in argument logic.

Overall I think this issue exposes a design fault in how we treat sending incomplete events.

[...]

PR #2617, opened 2023-01-23
events: pass uninitialized arguments

AlonZivony:
## Initial Checklist

## Description (git log)

Not passing arguments when they are missing from the eBPF payload proved to be problematic, because of parsing errors and database tables building.
Now we always pass all arguments, with default `nil` value if they are missing.

Fixes: #2261

## Type of change

## Final Checklist:

## Git Log Checklist:

My commits logs have:

### DIFF
--- a/docs/docs/detecting/golang.md
+++ b/docs/docs/detecting/golang.md
@@ -70,7 +70,7 @@ There are 2 ways you can get your own golang signatures working with tracee.
         
         		switch e.EventName {
         		case "openat", "execve":
-        			arg, err := helpers.GetTraceeArgumentByName(e, "pathname")
+        			arg, err := helpers.GetTraceeArgumentByName(e, "pathname", helpers.GetArgOps{DefaultArgs: false})
         			if err != nil {
         				return err
         			}
--- a/pkg/ebpf/events_pipeline.go
+++ b/pkg/ebpf/events_pipeline.go
@@ -158,10 +158,9 @@ func (t *Tracee) decodeEvents(outerCtx context.Context, sourceChan chan []byte) 
 				continue
 			}
 
-			args := make([]trace.Argument, 0, ctx.Argnum)
-
+			args := make([]trace.Argument, len(eventDefinition.Params))
 			for i := 0; i < int(ctx.Argnum); i++ {
-				argMeta, argVal, err := bufferdecoder.ReadArgFromBuff(
+				idx, arg, err := bufferdecoder.ReadArgFromBuff(
 					eventId,
 					ebpfMsgDecoder,
 					eventDefinition.Params,
@@ -170,8 +169,16 @@ func (t *Tracee) decodeEvents(outerCtx context.Context, sourceChan chan []byte) 
 					t.handleError(fmt.Errorf("failed to read argument %d of event %s: %v", i, eventDefinition.Name, err))
 					continue
 				}
-
-				args = append(args, trace.Argument{ArgMeta: argMeta, Value: argVal})
+				if args[idx].Value != nil {
+					t.handleError(fmt.Errorf("read more than one instance of argument %s of event %s. Saved value: %v. New value: %v", arg.Name, eventDefinition.Name, args[idx].Value, arg.Value))
+				}
+				args[idx] = arg
+			}
+			// Fill missing arguments metadata
+			for i := 0; i < len(eventDefinition.Params); i++ {
+				if args[i].Value == nil {
+					args[i].ArgMeta = eventDefinition.Params[i]
+				}
 			}
 
 			// Add stack trace if needed
--- a/signatures/helpers/arguments_helpers.go
+++ b/signatures/helpers/arguments_helpers.go
@@ -7,10 +7,18 @@ import (
 	"github.com/aquasecurity/tracee/types/trace"
 )
 
+// GetArgOps represents options for arguments getters
+type GetArgOps struct {
+	DefaultArgs bool // Receive default args value (value equals 'nil'). If set to false, will return error if arg not initialized.
+}
+
 // GetTraceeArgumentByName fetches the argument in event with `Name` that matches argName
-func GetTraceeArgumentByName(event trace.Event, argName string) (trace.Argument, error) {
+func GetTraceeArgumentByName(event trace.Event, argName string, opts GetArgOps) (trace.Argument, error) {
 	for _, arg := range event.Args {
 		if arg.Name == argName {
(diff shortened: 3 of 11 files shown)


tests/e2e-net-rules/helpers.go
```diff
@@ -6,7 +6,7 @@ import (
 	"github.com/aquasecurity/tracee/types/trace"
 )
 
-func GetTraceeArgumentByName(event trace.Event, argName string) (trace.Argument, error) {
+func GetInitializedTraceeArgumentByName(event trace.Event, argName string) (trace.Argument, error) {
 	for _, arg := range event.Args {
 		if arg.Name == argName {
 			return arg, nil
```
roikol, line 12: shouldn't you check here as well? `arg.Value != nil`

roikol (approved):
LGTM!
just one question here

signatures/helpers/arguments_helpers.go
```diff
@@ -17,9 +17,23 @@ func GetTraceeArgumentByName(event trace.Event, argName string) (trace.Argument,
 	return trace.Argument{}, fmt.Errorf("argument %s not found", argName)
 }
 
+// GetInitializedTraceeArgumentByName is an envelope function for GetTraceeArgumentByName that
+// return error if the argument is not initialized.
+func GetInitializedTraceeArgumentByName(event trace.Event, argName string) (trace.Argument, error) {
```
NDStrahilevitz, line 22: I think it might be better to actually use a "query options struct" in the previous function to change its behavior rather than exporting a new function (internally we can keep these as two `private` separate functions.

So we would have one function `GetTraceeArgument(event, name, opts)` where opts is a struct that could for now look like `{ DefaultValue bool //document this option` (we could perhaps add other query options later), and it would internally call either this function (which we would make private) or the old function.

NDStrahilevitz (changes requested):
I think we need to make the usage difference more explicit and clear to signature writers. A documentation section or page should also be written to make signature users aware of the possibility of uninitialized arguments and how they can choose to handle it.

Also I think it would be good to move the signature helpers to the `types` module so signature writers can reuse them directly, and so we wouldn't have to copy paste them in all sort of different places (we have them now both in `signature/helpers` and `tests/e2e-net-rules` and maybe more places, and we have another copies in internal projects as well).

signatures/helpers/arguments_helpers.go
AlonZivony: So if we follow the Clean Code rules, then:
`Don't use flag arguments. Split method into several independent methods that can be called from the client without the flag.`

I tried to follow this rule, but if you think it is more useable this way I have no problem with changing it.

NDStrahilevitz: > So if we follow the Clean Code rules, then:
Don't use flag arguments. Split method into several independent methods that can be called from the client without the flag.

I would usually also follow this rule, however we now have to take into account that we have to make sure that non tracee devs can, will and should use these helpers. As such we have to make their correct use as clear as possible, and an options struct would encode it well IMO.
With the current naming schema I think if I were not a tracee dev it wouldn't be clear to me what's the difference between the two and why there is a difference as well (this is also a documentation issue).

docs/docs/detecting/golang.md
```diff
@@ -70,7 +70,7 @@ There are 2 ways you can get your own golang signatures working with tracee.
         
         		switch e.EventName {
         		case "openat", "execve":
-        			arg, err := helpers.GetTraceeArgumentByName(e, "pathname")
+        			arg, err := helpers.GetInitializedTraceeArgumentByName(e, "pathname")
```
yanivagman, line 73: This function name is too long.
Maybe rename to `GetInitializedEvtArg` and the existing `GetTraceeArgumentByName` function name to `GetEvtArg`

pkg/ebpf/events_pipeline.go
```diff
@@ -170,8 +169,16 @@ func (t *Tracee) decodeEvents(outerCtx context.Context, sourceChan chan []byte)
 					t.handleError(fmt.Errorf("failed to read argument %d of event %s: %v", i, eventDefinition.Name, err))
 					continue
 				}
-
-				args = append(args, trace.Argument{ArgMeta: argMeta, Value: argVal})
+				if args[idx].Value != nil {
+					t.handleError(fmt.Errorf("read more than one instance of argument %s of event %s", arg.Name, eventDefinition.Name))
```
yanivagman, line 173: It might be helpful to also include the previous and current argument values in the message

tests/e2e-net-rules/helpers.go
```diff
@@ -14,3 +14,17 @@ func GetTraceeArgumentByName(event trace.Event, argName string) (trace.Argument,
...
+	arg, err := GetTraceeArgumentByName(event, argName)
+	// Return find error if received, as the argument won't be initialized
+	if err != nil {
+		return arg, err
+	}
+	if arg.Value == nil {
+		return arg, fmt.Errorf("argument %s is not initialized", argName)
+	}
+	return arg, nil
+}
```
yanivagman, line 30: Let's remove this to avoid code duplication

docs/docs/detecting/golang.md
```diff
@@ -69,8 +69,7 @@ There are 2 ways you can get your own golang signatures working with tracee.
         		}
         
         		switch e.EventName {
-        		case "openat", "execve":
-        			arg, err := helpers.GetTraceeArgumentByName(e, "pathname")
+        		case "openat", "execve":TraceeArgumentByName(e, "pathname", helpers.GetArgOps{DefaultArgs: false})
```
yanivagman, line 72: Missing "Get" prefix in function name

yanivagman (approved):
Small typo in documentation, but LGTM

docs/docs/detecting/golang.md
NDStrahilevitz: Is this valid syntax?

docs/docs/detecting/golang.md
```diff
@@ -70,7 +70,7 @@ There are 2 ways you can get your own golang signatures working with tracee.
         
         		switch e.EventName {
         		case "openat", "execve":
-        			arg, err := helpers.GetTraceeArgumentByName(e, "pathname")
+                    arg, err := helpers.GetTraceeArgumentByName(e, "pathname", helpers.GetArgOps{DefaultArgs: false})
```
NDStrahilevitz, line 73: Indentation looks wrong, but we can fix this later I guess. Is it rendered ok and just showing wrong here?

PR #2621, opened 2023-01-24
network: adjust capturing packets in many fronts

rafaeldtinoco:
1. Tracee would capture any network packet independently of given event filters. That wouldn't allow users to specify a single protocol to be captured, for example. Now, consider the following command:

    tracee-ebpf --output none \
        --trace event=net_packet_dns,net_packet_icmp \
        --capture network

   Tracee is able to capture ICMP and DNS packets only, for example.

2. Tracee would capture all PCAP files using the entire packet payloads. That would lead a situation where the user wouldn't be able to run tracee, with the capturing option, for too long in a higher data transfer environment, as the PCAP file would grow in a very fast pace.

   Considering the following command:

    tracee-ebpf --output none \
        --trace event=net_packet_dns,net_packet_icmp \
        --capture network
        --capture pcap:command
        --capture pcap-snaplen:{headers,max,128b,256b,512b,1k,2k,4k,...}

   The user is now capable of selecting how much data they want to include on each captured packet:

   - headers: capture only up to layer 4 headers and no payload
   - max: capture the entire packet payload (heavy)
   - size + b: size in bytes of the amount of data after the L4 header to capture
   - size + kb: size in KBS of the amount of data after the L4 header to capture

   There are still some improvements to be made:

   - If capturing net_packet_ipv4, the IPv6 packets will also be captured. Some userland filtering will have to take its place in userland for derived network packet events.

   - After multi scopes are merged, each scope might have its own PCAP files, this way each scope might have a set of its own artifacts files for introspection.

### DIFF
--- a/pkg/cmd/flags/capture.go
+++ b/pkg/cmd/flags/capture.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	tracee "github.com/aquasecurity/tracee/pkg/ebpf"
@@ -14,16 +15,17 @@ func captureHelp() string {
 Captured artifacts will appear in the 'output-path' directory.
 Possible options:
 
-[artifact:]write[=/path/prefix*]         capture written files. A filter can be given to only capture file writes whose path starts with some prefix (up to 50 characters). Up to 3 filters can be given.
-[artifact:]exec                          capture executed files.
-[artifact:]module                        capture loaded kernel modules.
-[artifact:]mem                           capture memory regions that had write+execute (w+x) protection, and then changed to execute (x) only.
-[artifact:]network                       capture network traffic. Only TCP/UDP/ICMP protocols are currently supported.
+[artifact:]write[=/path/prefix*]              capture written files. A filter can be given to only capture file writes whose path starts with some prefix (up to 50 characters). Up to 3 filters can be given.
+[artifact:]exec                               capture executed files.
+[artifact:]module                             capture loaded kernel modules.
+[artifact:]mem                                capture memory regions that had write+execute (w+x) protection, and then changed to execute (x) only.
+[artifact:]network                            capture network traffic. Only TCP/UDP/ICMP protocols are currently supported.
 
-dir:/path/to/dir                         path where tracee will save produced artifacts. the artifact will be saved into an 'out' subdirectory. (default: /tmp/tracee).
-profile                                  creates a runtime profile of program executions and their metadata for forensics use.
-clear-dir                                clear the captured artifacts output dir before starting (default: false).
-pcap:[single,process,container,command]  capture separate pcap files organized by single file, files per processes, containers and/or commands
+dir:/path/to/dir                              path where tracee will save produced artifacts. the artifact will be saved into an 'out' subdirectory. (default: /tmp/tracee).
+profile                                       creates a runtime profile of program executions and their metadata for forensics use.
+clear-dir                                     clear the captured artifacts output dir before starting (default: false).
+pcap:[single,process,container,command]       capture separate pcap files organized by single file, files per processes, containers and/or commands
+pcap-snaplen:[default, headers, max or SIZE]  sets captured payload from each packet (default=96b, headers, max, or 512b, 1kb, ...)
 
 Examples:
   --capture exec                                           | capture executed files into the default output directory
@@ -32,6 +34,10 @@ Examples:
   --capture profile                                        | capture executed files and create a runtime profile in the output directory
   --capture net (or network)                               | capture network traffic. default: single pcap file containing all traced packets
   --capture net --capture pcap:process,command             | capture network traffic, save pcap files for processes and commands
+  --capture net --capture pcap-snaplen:1kb                 | capture network traffic, single pcap file (default), capture 1KB from each packet
+  --capture net --capture pcap-snaplen:max                 | capture network traffic, single pcap file (default), capture full sized packets
+  --capture net --capture pcap-snaplen:headers             | capture network traffic, single pcap file (default), capture headers only from each packet
+  --capture net --capture pcap-snaplen:default             | capture network traffic, single pcap file (default), capture headers + 96 bytes from each packet
   --capture network --capture pcap:container,command       | capture network traffic, save pcap files for containers and commands
   --capture exec --output none                             | capture executed files into the default output directory not printing the stream of events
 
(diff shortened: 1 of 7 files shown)


pkg/cmd/flags/capture.go
```diff
@@ -24,6 +25,7 @@ dir:/path/to/dir                         path where tracee will save produced ar
 profile                                  creates a runtime profile of program executions and their metadata for forensics use.
 clear-dir                                clear the captured artifacts output dir before starting (default: false).
 pcap:[single,process,container,command]  capture separate pcap files organized by single file, files per processes, containers and/or commands
+pcap-snaplen:[size, headers or max]      sets length to be captured from each packet's payload (100b, 512b, 1kb, max or headers). 64kb, max, or higher, makes entire payload available.
```
rafaeldtinoco, line 28: snaplen is the same term used by tcpdump, I wanted to keep it because people already knowing tcpdump will understand more easily.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -859,6 +859,10 @@ typedef struct config_entry {
     u8 events_to_submit[128]; // use 8*128 bits to describe up to 1024 events
 } config_entry_t;
 
+typedef struct netconfig_entry {
+    int capture_length; // amount of network packet payload to capture (pcap)
```
rafaeldtinoco, line 863: I might need more fields here in near future, so I guessed having a map for the network config options would be ok.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -7337,28 +7343,34 @@ CGROUP_SKB_HANDLE_FUNCTION(proto_icmpv6);
 // Network submission functions
 //
 
-#define FULL    1
-#define HEADERS 0
+#define FULL    65536 // 1 << 16
```
rafaeldtinoco, line 7346: 1 << 16 is actually the max size of an IP packet. The sk_buff contains multiple IP packets (in the same TCP segment, when TCP is being used) but its max size is 64k as well. When setting this to 64k it would overflow the IPv4 packet length field, so I simply use the number to tell cgroup_skb_submit to send the ENTIRE packet (whenever 64k is used). This is done through flags argument of that function.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -7594,6 +7626,14 @@ CGROUP_SKB_HANDLE_FUNCTION(proto)
             return CGROUP_SKB_HANDLE(proto_icmpv6);
     }
 
+    // TODO: If cmdline is tracing net_packet_ipv6 only, then the the ipv4
```
rafaeldtinoco, line 7629: I also quoted this in the git log, and will adjust this once we implement capturing different artifacts per given scope (and when I start to run filtering in the network events pipeline, from its perfbuffer).

pkg/cmd/flags/capture.go
```diff
@@ -90,6 +96,37 @@ func PrepareCapture(captureSlice []string) (tracee.CaptureConfig, error) {
...
+			context = strings.ToLower(context) // normalize
+			if context == "max" {
+				amount = (1 << 16) // sets full packet length for capturing
+			} else if context == "headers" {
+				amount = 0 // sets headers only length for capturing (default)
+			} else if strings.HasSuffix(context, "kb") ||
+				strings.HasSuffix(context, "k") {
+				context = strings.TrimSuffix(context, "kb")
+				context = strings.TrimSuffix(context, "k")
+				amount, err = strconv.ParseInt(context, 10, 64)
```
geyslan, line 113: As the final amount is treated as an uint, how about using `strconv.ParseUint()`?

pkg/cmd/flags/capture.go
```diff
@@ -90,6 +96,37 @@ func PrepareCapture(captureSlice []string) (tracee.CaptureConfig, error) {
...
+				amount *= 1024 // result in bytes
+			} else if strings.HasSuffix(context, "b") {
+				context = strings.TrimSuffix(context, "b")
+				amount, err = strconv.ParseInt(context, 10, 64)
```
geyslan, line 117: ditto.

pkg/pcaps/pcaps.go
```diff
@@ -30,6 +30,7 @@ type Config struct {
 	CaptureProcess   bool
 	CaptureContainer bool
 	CaptureCommand   bool
+	CaptureLength    int64
```
geyslan, line 33: `uint32`?

pkg/ebpf/c/tracee.bpf.c
geyslan, line 863: `u32`?

geyslan (approved):
overall it LGTM

pkg/pcaps/pcaps.go
rafaeldtinoco: yep, fixed (will push soon)

rafaeldtinoco:
I'm finishing up some changes (regarding mangling packet headers with lengths) and will force push as soon as its ready (very soon).

roikol (commented):
Looks good @rafaeldtinoco !!

pkg/cmd/flags/capture.go
```diff
@@ -90,6 +96,37 @@ func PrepareCapture(captureSlice []string) (tracee.CaptureConfig, error) {
...
+			} else {
+				return tracee.CaptureConfig{}, fmt.Errorf("coud not parse pcap snaplen: missing b or kb ?")
+			}
+			if err != nil {
+				return tracee.CaptureConfig{}, fmt.Errorf("coud not parse pcap snaplen: %v", err)
+			}
+			if amount == (1 << 16) {
+				amount -= 1
+			} else if amount > (1 << 16) {
+				return tracee.CaptureConfig{}, fmt.Errorf("maximum pcap snaplen is 64kb")
```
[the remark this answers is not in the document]
rafaeldtinoco: ok

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -6888,7 +6893,8 @@ static __always_inline bool is_socket_supported(struct socket *sock)
 typedef struct net_event_contextmd {
     u8 submit;
     u32 header_size;
-    u16 padding;
+    u8 captured;
+    u8 padding;
```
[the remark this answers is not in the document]
rafaeldtinoco: Yes, I want to control the size of this struct because of `sizeof_net_event_context_t`, so the metadata is always aligned.

Issue #2631, opened 2023-01-27
Changing the default set of events


rafaeldtinoco:
### Discussed in https://github.com/aquasecurity/tracee/discussions/2611

<div type='discussions-op-text'>

<sup>Originally posted by **yanivagman** January 22, 2023</sup>
We got feedback from some users saying that today's default set is too noisy, and a suggestion for a new default set that we can use.

After discussion, these were defined as default:

`sudo ./dist/tracee --list | grep '\[default'`

```shell
creat                          [default syscalls fs fs_file_ops]
chmod                          [default syscalls fs fs_file_attr]
fchmod                         [default syscalls fs fs_file_attr]
chown                          [default syscalls fs fs_file_attr]
fchown                         [default syscalls fs fs_file_attr]
lchown                         [default syscalls fs fs_file_attr]
ptrace                         [default syscalls proc]
setuid                         [default syscalls proc proc_ids]
setgid                         [default syscalls proc proc_ids]
setpgid                        [default syscalls proc proc_ids]
setsid                         [default syscalls proc proc_ids]
setreuid                       [default syscalls proc proc_ids]
setregid                       [default syscalls proc proc_ids]
setresuid                      [default syscalls proc proc_ids]
setresgid                      [default syscalls proc proc_ids]
setfsuid                       [default syscalls proc proc_ids]
setfsgid                       [default syscalls proc proc_ids]
mount                          [default syscalls fs]
init_module                    [default syscalls system system_module]
fchownat                       [default syscalls fs fs_file_attr]
fchmodat                       [default syscalls fs fs_file_attr]
setns                          [default syscalls proc]
process_vm_readv               [default syscalls proc]
process_vm_writev              [default syscalls proc]
finit_module                   [default syscalls system system_module]
memfd_create                   [default syscalls fs fs_file_ops]
move_mount                     [default syscalls fs]
sched_process_exec             [default proc]
security_inode_unlink          [default lsm_hooks fs fs_file_ops]
security_socket_connect        [default lsm_hooks net net_sock]
security_socket_accept         [default lsm_hooks net net_sock]
security_socket_bind           [default lsm_hooks net net_sock]
security_sb_mount              [default lsm_hooks fs]
net_packet_icmp                [default network_events]
net_packet_icmpv6              [default network_events]
net_packet_dns_request         [default network_events]
net_packet_dns_response        [default network_events]
net_packet_http_request        [default network_events]
net_packet_http_response       [default network_events]
```

[...]

PR #2636, opened 2023-01-27
chore: new default set of events

geyslan:
7f860ae2 chore: new default set of events

Fixes:

- #2631 

## Type of change

- [x] Breaking change (cause existing functionality not to work as expected).

### DIFF
--- a/pkg/events/events.go
+++ b/pkg/events/events.go
@@ -309,7 +309,7 @@ var Definitions = eventDefinitions{
 			ID32Bit: sys32open,
 			Name:    "open",
 			Syscall: true,
-			Sets:    []string{"default", "syscalls", "fs", "fs_file_ops"},
+			Sets:    []string{"syscalls", "fs", "fs_file_ops"},
 			Params: []trace.ArgMeta{
 				{Type: "const char*", Name: "pathname"},
 				{Type: "int", Name: "flags"},
@@ -320,7 +320,7 @@ var Definitions = eventDefinitions{
 			ID32Bit: sys32close,
 			Name:    "close",
 			Syscall: true,
-			Sets:    []string{"default", "syscalls", "fs", "fs_file_ops"},
+			Sets:    []string{"syscalls", "fs", "fs_file_ops"},
 			Params: []trace.ArgMeta{
 				{Type: "int", Name: "fd"},
 			},
@@ -329,7 +329,7 @@ var Definitions = eventDefinitions{
 			ID32Bit: sys32stat,
 			Name:    "stat",
 			Syscall: true,
-			Sets:    []string{"default", "syscalls", "fs", "fs_file_attr"},
+			Sets:    []string{"syscalls", "fs", "fs_file_attr"},
 			Params: []trace.ArgMeta{
 				{Type: "const char*", Name: "pathname"},
 				{Type: "struct stat*", Name: "statbuf"},
@@ -339,7 +339,7 @@ var Definitions = eventDefinitions{
 			ID32Bit: sys32fstat,
 			Name:    "fstat",
 			Syscall: true,
-			Sets:    []string{"default", "syscalls", "fs", "fs_file_attr"},
+			Sets:    []string{"syscalls", "fs", "fs_file_attr"},
 			Params: []trace.ArgMeta{
 				{Type: "int", Name: "fd"},
 				{Type: "struct stat*", Name: "statbuf"},
@@ -349,7 +349,7 @@ var Definitions = eventDefinitions{
 			ID32Bit: sys32lstat,
 			Name:    "lstat",
 			Syscall: true,
-			Sets:    []string{"default", "syscalls", "fs", "fs_file_attr"},
+			Sets:    []string{"syscalls", "fs", "fs_file_attr"},
 			Params: []trace.ArgMeta{
 				{Type: "const char*", Name: "pathname"},
 				{Type: "struct stat*", Name: "statbuf"},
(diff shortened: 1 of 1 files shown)


rafaeldtinoco:
@yanivagman this one is yours, I believe it needs a decision more than a technical view.

pkg/events/events.go
```diff
@@ -1387,7 +1387,7 @@ var Definitions = eventDefinitions{
 			ID32Bit: sys32setuid32,
 			Name:    "setuid",
 			Syscall: true,
-			Sets:    []string{"default", "syscalls", "proc", "proc_ids"},
+			Sets:    []string{"syscalls", "proc", "proc_ids"},
```
yanivagman, line 1390: Are the other setXXXid syscalls noisy? Maybe we should include them after all?

pkg/events/events.go
```diff
@@ -1960,7 +1960,7 @@ var Definitions = eventDefinitions{
 			ID32Bit: sys32mount,
 			Name:    "mount",
 			Syscall: true,
-			Sets:    []string{"default", "syscalls", "fs"},
+			Sets:    []string{"syscalls", "fs"},
```
yanivagman, line 1963: I think we should keep this one. It is important enough and not too frequent

pkg/events/events.go
```diff
@@ -2061,7 +2061,7 @@ var Definitions = eventDefinitions{
 			ID32Bit: sys32init_module,
 			Name:    "init_module",
 			Syscall: true,
-			Sets:    []string{"default", "syscalls", "system", "system_module"},
+			Sets:    []string{"syscalls", "system", "system_module"},
```
yanivagman, line 2064: init_module is important enought to be included

pkg/events/events.go
```diff
@@ -2950,7 +2950,7 @@ var Definitions = eventDefinitions{
 			ID32Bit: sys32fchownat,
 			Name:    "fchownat",
 			Syscall: true,
-			Sets:    []string{"default", "syscalls", "fs", "fs_file_attr"},
+			Sets:    []string{"syscalls", "fs", "fs_file_attr"},
```
yanivagman, line 2953: fchownat should be kept if we are adding chownat
if we keep chown then fchownat should be kept as well

pkg/events/events.go
```diff
@@ -3529,7 +3529,7 @@ var Definitions = eventDefinitions{
 			ID32Bit: sys32process_vm_readv,
 			Name:    "process_vm_readv",
 			Syscall: true,
-			Sets:    []string{"default", "syscalls", "proc"},
+			Sets:    []string{"syscalls", "proc"},
```
yanivagman, line 3532: let's keept this one too

pkg/events/events.go
```diff
@@ -3570,7 +3570,7 @@ var Definitions = eventDefinitions{
 			ID32Bit: sys32finit_module,
 			Name:    "finit_module",
 			Syscall: true,
-			Sets:    []string{"default", "syscalls", "system", "system_module"},
+			Sets:    []string{"syscalls", "system", "system_module"},
```
yanivagman, line 3573: and this

pkg/events/events.go
```diff
@@ -1190,7 +1190,7 @@ var Definitions = eventDefinitions{
 			ID32Bit: sys32creat,
 			Name:    "creat",
 			Syscall: true,
-			Sets:    []string{"default", "syscalls", "fs", "fs_file_ops"},
+			Sets:    []string{"syscalls", "fs", "fs_file_ops"},
```
yanivagman, line 1193: let's keep this to track the creation of new files

pkg/events/events.go
```diff
@@ -1282,7 +1282,7 @@ var Definitions = eventDefinitions{
 			ID32Bit: sys32lchown32,
 			Name:    "lchown",
 			Syscall: true,
-			Sets:    []string{"default", "syscalls", "fs", "fs_file_attr"},
+			Sets:    []string{"syscalls", "fs", "fs_file_attr"},
```
yanivagman, line 1285: if we keep chown then we should keep this as well

pkg/events/events.go
```diff
@@ -3639,7 +3639,7 @@ var Definitions = eventDefinitions{
 			ID32Bit: sys32memfd_create,
 			Name:    "memfd_create",
 			Syscall: true,
-			Sets:    []string{"default", "syscalls", "fs", "fs_file_ops"},
+			Sets:    []string{"syscalls", "fs", "fs_file_ops"},
```
yanivagman, line 3642: we should also keep this one

pkg/events/events.go
```diff
@@ -3891,7 +3891,7 @@ var Definitions = eventDefinitions{
 			ID32Bit: sys32move_mount,
 			Name:    "move_mount",
 			Syscall: true,
-			Sets:    []string{"default", "syscalls", "fs"},
+			Sets:    []string{"syscalls", "fs"},
```
yanivagman, line 3894: let's keep it

pkg/events/events.go
```diff
@@ -5398,7 +5398,7 @@ var Definitions = eventDefinitions{
 			Probes: []probeDependency{
 				{Handle: probes.SecuritySbMount, Required: true},
 			},
-			Sets: []string{"default", "lsm_hooks", "fs"},
+			Sets: []string{"lsm_hooks", "fs"},
```
yanivagman, line 5401: let's keep it

yanivagman (changes requested):
Some events we should reconsider adding. WDYT?

pkg/events/events.go
geyslan: `sudo nano /etc/shadow` generated 22 events with all setXXXid, so it seems ok. Including them.

yanivagman (approved):
LGTM

PR #2639, opened 2023-01-28
docs: cli: general fixes and cosmetics

geyslan:
Quote cli option values with single quotation mark when required.

## Type of change


yanivagman (approved):
LGTM

PR #2641, opened 2023-01-29
docs: add tracee-action

itaysk:
(also removed the explanation about tracee-rules/tracee-ebpf as it's not that interesting going forward)


yanivagman (approved):
LGTM

PR #2646, opened 2023-01-30
ebpf: initialize scopes in trigger events

geyslan:
ebpf: initialize scopes in trigger events

Addendum to #2644


yanivagman (approved):
LGTM

PR #2645, opened 2023-01-30
chore: fix: default set container conflict

geyslan:
Some events

```
ContainerCreate
ContainerRemove
ExistingContainer
```

can't be mixed with `container=new`.

This fixes integration tests failures like:

`--- FAIL: Test_EventFilters/trace_only_events_from_new_containers`

Context:
- #2636

PR #2647, opened 2023-01-30
events: bring back container create/remove events

NDStrahilevitz:
#2645 removed all container set events from the default set because they broke integration tests (in particular `existing_container`).
This PR brings back `container_create` and `container_remove`.


geyslan (approved):
LGTM

PR #2663, opened 2023-01-31
logger: set libbpfgo logger callback

geyslan:
## Initial Checklist

## Description (git log)

commit f965b789dde7a42f92919ae241a14e5434bc66e3

    libbpfgo: bump to 8f83f25
    
    From this hash, libbpfgo publicizes its log callback API.

commit f14050d2e9ec81013afd2388d27111729be14d54

    chore: set libbpfgo logger callback

## Depends on

- https://github.com/aquasecurity/libbpfgo/pull/284

## Type of change

### Context

- https://github.com/aquasecurity/tracee/pull/2600

Fixes: #2417

### DIFF
--- a/cmd/tracee-ebpf/main.go
+++ b/cmd/tracee-ebpf/main.go
@@ -6,6 +6,7 @@ import (
 	"os/signal"
 	"syscall"
 
+	"github.com/aquasecurity/libbpfgo"
 	"github.com/aquasecurity/tracee/pkg/cmd"
 	"github.com/aquasecurity/tracee/pkg/cmd/flags"
 	"github.com/aquasecurity/tracee/pkg/cmd/flags/server"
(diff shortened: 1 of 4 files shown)


geyslan:
> After merging this and https://github.com/aquasecurity/tracee/pull/2663, I'll open an issue in both projects to tackle the hard-coded filters move.

https://github.com/aquasecurity/libbpfgo/pull/284#issuecomment-1437170456

cmd/tracee-ebpf/main.go
```diff
@@ -29,6 +30,28 @@ func init() {
...
+				lvl = logger.InfoLevel
+			case libbpfgo.LibbpfDebugLevel:
+				lvl = logger.DebugLevel
+			}
+
+			logger.Log(lvl, false, msg, keyValues...)
+		},
+		LogFilters: []func(libLevel int, msg string) bool{
+			libbpfgo.LogFilterLevel,
+			libbpfgo.LogFilterOutput,
```
rafaeldtinoco, line 52: So you're planning on using the default LogFilter function, then making it generic in libbpfgo, then bumping libbpfgo again in Tracee (that time with a function filtering the output). Is that so ?

geyslan: If I got you right, it's not my intent. This PR only closes the cycle of the backlog in libbpfgo without changing behaviour of both projects.

The next step must be to remove the consumer logic from libbfpgo and place it, in our case, in tracee. See https://github.com/aquasecurity/libbpfgo/issues/293.

After this merge I'm opening this, in tracee, as well:

image

rafaeldtinoco: Cool that is exactly what I meant =D. Sorry for the miss interpretation.

rafaeldtinoco (approved):
LGTM

Issue #2660, opened 2023-01-31
Running tracee from image shows lockdown error log


yanivagman:
## Description

```
docker run   --name tracee --rm -it   --pid=host --cgroupns=host --privileged   -v /etc/os-release:/etc/os-release-host:ro   -e LIBBPFGO_OSRELEASE_FILE=/etc/os-release-host   aquasec/tracee:latest   trace
{"level":"error","ts":1675174347.176406,"msg":"osinfo","error":"open /sys/kernel/security/lockdown: no such file or directory"}
^CTIME             UID    COMM             PID     TID     RET              EVENT                     ARGS
```

## Output of `tracee -v`:

```
v0.11.0
```

we should silence this log message instead by changing to info level

rafaeldtinoco:
Info level is the default and the message can still be seen. Moving it to debug level then.

PR #2684, opened 2023-02-07
lockdown: fix: turn error message into a debug one

rafaeldtinoco:
Fixes: #2660


yanivagman:
You wrote in the issue turning the message to debug, but you changed to info.
Which one is correct?

pkg/cmd/urfave/urfave.go
```diff
@@ -107,9 +107,8 @@ func GetTraceeRunner(c *cli.Context, version string) (cmd.Runner, error) {
 
 	lockdown, err := helpers.Lockdown()
 	if err != nil {
-		logger.Error("osinfo", "error", err)
+		logger.Info("osinfo", "lockdown", err)
```
NDStrahilevitz, line 110: PR title says it should be a debug message but this is info. Intentional?

rafaeldtinoco:
Ooops, I pushed the testing version (with Info) instead of correct one (with Debug). Changing it.

yanivagman (approved):
LGTM

Issue #2681, opened 2023-02-06
make integration tests more robust


geyslan:
## Description

- We only have a few tests for multi-scope. Since it touches many layers, we need to create a robust test suite.
- Integration tests lack support for testing more event filter types. Building a more specialized test suite is a must.

rafaeldtinoco:
For the integration tests, please be aware of: https://github.com/aquasecurity/tracee/issues/1681#issuecomment-1495598079 when adding tests that rely in PID=1. Cheers!

geyslan:
> For the integration tests, please be aware of: [#1681 (comment)](https://github.com/aquasecurity/tracee/issues/1681#issuecomment-1495598079) when adding tests that rely in PID=1. Cheers!

OK. But if we run the integration test in a container just to not touch the host pid 1, that process will be a shell (for example)... so I think running tests just to filter the pid 1 that way doesn't make sense. It's the same of test against pid=whatever.

rafaeldtinoco:
> so I think running tests just to filter the pid 1 that way doesn't make sense. It's the same of test against pid=whatever.

Sorry, I did not make myself clear, I just wanted to bring awareness to the issue, not to run tests inside a container as a suggestion!

PR #2688, opened 2023-02-07
make integration tests more robust

geyslan:
chore: make integration tests more robust
    
    - Add generic functions to check received events in different manners:
      - ExpectAllInOrder
      - ExpectAllEqualTo
      - ExpectAtLeastOneOfEach
      - ExpectAnyOfEach
    - Add waitForTraceeOutputEvents function to wait for tracee output
      events until buffer fills with certain number of events or
      timeout occurs.
    - Add waitforTraceeStop function to wait for tracee to stop or until
      timeout occurs.
    - Remove tester.sh concentrating all event filter tests in
      event_filters_test.go.
    - Create syscaller tool to easily generate syscalls from the command
      line (strictly for integration tests). It reuses the tracee events
      package.
    - Fix goroutine leaks, deadlocks and race conditions.

Fixes: #2681

## Initial Checklist

## Type of change

## How Has This Been Tested?

`make test-unit`
`sudo make test-integration`


geyslan:
This is a WIP but any review is welcome.

.gitignore
```diff
@@ -25,4 +25,4 @@ compile_commands.json
 .vagrant
 
 # binaries and build files
-dist
\ No newline at end of file
+dist
```
geyslan, line 28: Touched this file when created the syscaller tool in the implementation first stage. So, just a not related formatting fix here.

tests/integration/integration_test.go
```diff
@@ -259,112 +68,22 @@ func checkDockerdBinaryFilter(t *testing.T, gotOutput *[]trace.Event) {
...
-		{
-			name:       "trace only execve events that contains l",
-			filterArgs: []string{"event=execve", "execve.args.pathname=*l*"},
-			eventFunc:  checkExecve,
-		},
 		{
 			name:       "trace only events from new containers",
 			filterArgs: []string{"container=new", "event!=container_create,container_remove"},
 			eventFunc:  checkNewContainers,
 		},
```
geyslan, line 81: These remaining tests will be migrated.

tests/integration/integration_test.go
```diff
@@ -259,112 +68,22 @@ func checkDockerdBinaryFilter(t *testing.T, gotOutput *[]trace.Event) {
...
-		{
-			name:       "trace only security_file_open from \"execve\" syscall",
-			filterArgs: []string{"event=security_file_open", "security_file_open.context.syscall=execve"},
-			eventFunc:  checkSecurityFileOpenExecve,
-		},
 		{
 			name:       "trace only events from \"/usr/bin/dockerd\" binary and contain it's pid",
 			filterArgs: []string{"bin=/usr/bin/dockerd"},
 			eventFunc:  checkDockerdBinaryFilter,
 		},
```
geyslan, line 86: ditto.

tests/integration/exec.go
```diff
@@ -0,0 +1,190 @@
...
+	err = syscall.Close(int(null.Fd()))
+	if err != nil {
+		return fmt.Errorf("close: %w", err)
+	}
+
+	return nil
+}
+
+// waitChild waits for a child process to exit
+func waitChild(pid int, timeout time.Duration) (int, error) {
```
geyslan, line 84: This timeout ensures we don't hang but can cause other issues like:

https://github.com/aquasecurity/tracee/actions/runs/4195065537/jobs/7274176239

```
=== RUN   Test_EventFilters/container:_event:_trace_only_events_from_new_containers
    tracee.go:81: started tracee...
	executing: /usr/bin/docker [docker run -d --rm alpine]
Unable to find image 'alpine:latest' locally
latest: Pulling from library/alpine
63b65145d645: Pulling fs layer
make: *** wait: No child processes.  Stop.
make: *** Waiting for unfinished jobs....
make: *** wait: No child processes.  Stop.
Error: Process completed with exit code 143.
```

geyslan: Fixed. Now each exec has up to 10s to finish or will be killed.

tests/integration/exec.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	return cmd, args, nil
+}
+
+// handleFDs closes all fds except for stdin, stdout, stderr
+// It also redirects stdout to /dev/null
+func handleFDs() error {
+	// close range of fds
+	err := unix.CloseRange(3, ^uint(0), unix.CLOSE_RANGE_UNSHARE)
```
rafaeldtinoco, line 66: This syscall was added in 5.9 and glibc 2.34. Not sure about clang support. Golang uses Syscall(SYS_CLOSE_RANGE). I wonder how this passed 5.4 tests and if it passes vanilla 5.4 kernels. Just curious.

geyslan: Really curious. SYS_CLOSE_RANGE is 463, so it should have returned an error.

NDStrahilevitz: Integration tests don't run per kernel so I guess that's why (probably the machine running it is one of the latest kernels).

geyslan: You're right. That's the cause.

tests/integration/exec.go
```diff
@@ -0,0 +1,197 @@
...
+		return -1, syscall.Errno(errno)
+	}
+
+	return waitChild(int(pid), timeout)
+}
+
+// forkAndCallSys forks and calls the syscalls specified in evts, changing the child's comm to newComm
+// It returns the pid of the child process
+func forkAndCallSys(newComm string, evts []trace.Event, timeout time.Duration) (int, error) {
+	pid, _, errno := syscall.RawSyscall(syscall.SYS_FORK, 0, 0, 0)
```
rafaeldtinoco, line 158: I don't think this is as simple as that. Your stack frame might be shared among different runtime threads (even if you're running on a single CPU).

Have you read the implementation of `unix.forkExec()`? The path I'm suggesting you to read, if you haven't, is:

```
forkExec() -> forkAndExecInChild() -> forkAndExecInChild1()
```

You will notice that there are runtime internal functions called:

1. runtime_BeforeFork() (right before vfork() is called)
2. runtime_AfterFork() (right after 
3. runtime_AfterForkInChild()

The clone() syscall is executed by vfork() and not fork(). It means that the parent's page tables won't be copied (likely due to performance, so we would be good in that aspect). The main difference is that the caller is suspended until the child calls execve(). This might make a big difference for runtime (that could be doing a bunch of stuff and preempting the runtime routine calling fork).

There are MANY other things managed by Golang's implementation of forkAndExec(). Many things related to process control group, terminal controlling processing, etc, etc.

It is very likely impossible to have a "forkAndFunction()" approach without the runtime having the capability to do so. A forkAndExec() exists just because runtime allows it to exist (with many mitigations done in the implementation I quoted).

tests/integration/exec.go
```diff
@@ -0,0 +1,197 @@
...
+			return -1, fmt.Errorf("timed out waiting for child process to exit")
+		}
+
+		time.Sleep(100 * time.Millisecond) // poll every 100ms
+	}
+}
+
+// forkAndExec forks and execs a command
+// It returns the pid of the child process
+func forkAndExec(cmd string, timeout time.Duration) (int, error) {
```
rafaeldtinoco, line 122: ditto below

rafaeldtinoco (changes requested):
Needs discussion regarding vfork() and clone() And all process creation related mitigations by the syscall package and the golang runtime internal functions.

rafaeldtinoco:
I think we can have our own version of forkAndExec() setting cpumask and changing command name, etc... but I don't think the forkAndSys would work correctly. Either way, I'll finish the review after this points are clarified.

tests/integration/exec.go
geyslan: > I don't think this is as simple as that. Your stack frame might be shared among different runtime threads (even if you're running on a single CPU).

Even in the case of an ordinary `fork()`? Note that I'm not using `vfork()`.

See below from `man fork 2`:

> The child process and the parent process run in separate memory spaces.  At the time of fork() both memory spaces have  the  same content.   Memory writes, file mappings (mmap(2)), and unmappings (munmap(2)) performed by one of the processes do not affect the other.

rafaeldtinoco: I meant the "parent" stack frame (shared among multiple runtime threads). This is likely the reason why there are internal (to runtime) functions that need to be called.

image

rafaeldtinoco: There are runtime locks that will guarantee the stack doesn't grow before an exec takes place, for example, and things like that.

geyslan: I understand that mixing Go runtime with fork and syscalls is not that trivial, especially not using the provided Go API.

My initial tests used a [syscaller.go binary](https://github.com/aquasecurity/tracee/blob/88ac808171c52231ced7bdb82f495ddd4d078f3d/tests/integration/syscaller/syscaller.go) that was detached from the tests runtime and could work nicely. But I noticed that if we wanted to improve the syscalls maps with information brought in from tracee event type it would be hard to maintain (due to code duplication).

Running this PR with runtime.GOMAXPROCS(2), reserving core 0 for tests executions and cores 1..maxcpu for the tests threads (tracee), it seems to work ok.

Instead, what would your proposal be?

geyslan:
Raising fixes related to race conditions detected.

tests/integration/exec.go
rafaeldtinoco: My suggestion was at the end of comments:

> I think we can have our own version of forkAndExec() setting cpumask and changing command name, etc... but I don't think the forkAndSys would work correctly.

I'm afraid of creating something that "works most of the times" but might introduce races and issues that are only present in the testing environment (because of things we could predict before creating the test environment).

About this:

> But I noticed that if we wanted to improve the syscalls maps with information brought in from tracee event type ...

Could you clarify what would be the info brought from the event type and how you would feed the syscall testing?

geyslan:
Tests failed due to possible time ticker deadlocks, I'll look into it TOM.

tests/integration/exec.go
geyslan: > I'm afraid of creating something that "works most of the times" but might introduce races and issues that are only present in the testing environment (because of things we could predict before creating the test environment).

I agree with you. So surely my initial detached sycaller.go binary would solve this. Having an external execution, through the ForkAndExec provided by the golang API, would work just fine. If the sycaller broke something related to its runtime, which is not expected anyway, as we will create (chain) the specific calls to be reproducible in each case.

> Could you clarify what would be the info brought from the event type and how you would feed the syscall testing? I ask you this because if its only a syscalls table with a bunch of arguments to be tested, then it might be okay to have something like your syscaller proposal.

```go
var syscallMap = map[events.ID]sysArgs{
	events.Read: {0, 0, 0, 0, 0, 0},
}
```

That is the current implementation, just a raw table with events and args, for simple tests as shown in test table. We can make sure that this simple map can be set with safe events args that won't harm runtime.

Having a foundation working for that, we can add and make use of:

```go
var chainedCallsMap = map[events.ID][]syscall{
	events.Read: {
		events.Open: {pathname, flags, 0, 0, 0, 0},
		events.Read: {fd, buf, count, 0, 0, 0},
		events.Close: {fd, 0, 0, 0, 0, 0},
	},
	// Signatures
	events.AntiDebugging: {
		...,
		events.Ptrace: {whatever, 0, 0, 0, 0, 0},
		...,
	},
}
```

> (which is actually close to having the shell scripts, no ?)

With only shell scripts or ready third-party binaries, we can't have the freedom to specify events in the order and way we want.

> Don't get me wrong, I think its a great idea to have what you are proposing, I just don't think its "safe" enough when it comes to running it over and over and over (and picking a runtime race because of something broke by design).

With a detached syscaller, as I mentioned, we can have a safe runtime.

Regarding the race conditions and the likely go routine leak (test cases loop), I have to dig into it as this behaviour seems to come from the main branch.

geyslan: _An addendum:_ to avoid getting stuck with the processorID test, which doesn't make much sense, we can certainly let the threads run free without blocking them. But we still need to make sure that calls (forks) should only run on the same CPU to avoid unordered events.

geyslan:
I'm postponing this to v0.13 due to:

- #2762
- #2763

tests/integration/exec.go
rafaeldtinoco: > With only shell scripts or ready third-party binaries, we can't have the freedom to specify events in the order and way we want.

I understand you want more control of what is being executed and the results it provides (like calling a specific syscall and checking the event it generated before moving on). If that is the case, I agree that having that would be very nice indeed.

rafaeldtinoco: About running on the same CPU to guarantee ordering, now we have network events and they run from softIRQ context... so that might not be enough for the network event testing... as the runtime might get preempted (if you're thinking about parallelizing the tests and checking results in parallel).

Maybe you will have to stick with a single routine doing event, checking, doing event, checking... but you will find out during implementation.

rafaeldtinoco:
Converting to draft as this will be broken multiple times until v0.13.0 is released.

rafaeldtinoco:
Converted this to draft as it will be broken multiple times before until v0.13.0 is released.

tests/integration/exec.go
geyslan: The current proposed solution makes use of an external tool, so I consider this solved.

josedonizetti:
I want to review this one

josedonizetti (approved):

rafaeldtinoco:
TODO: We should merge this one at the end of the milestone (no matter if its on the bug fix window) just so we don't change tests while doing the release.

josedonizetti (approved):

rafaeldtinoco:
image

@geyslan shouldn't integration tests binary be inside ./dist/ like all other binaries (including tracee-gpt, for example). Its a convention, I wonder if we could make the same for this binary.

geyslan:
> @geyslan shouldn't integration tests binary be inside ./dist/ like all other binaries (including tracee-gpt, for example). Its a convention, I wonder if we could make the same for this binary.

Currently the syscaller is used only for the golang integration tests that lie under `./tests/integration/syscaller`. But if you have plans to run it from other scripts and so, I think the move makes sense.

josedonizetti:
> > @geyslan shouldn't integration tests binary be inside ./dist/ like all other binaries (including tracee-gpt, for example). Its a convention, I wonder if we could make the same for this binary.
> 
> Currently the syscaller is used only for the golang integration tests that lie under `./tests/integration/syscaller`. But if you have plans to run it from other scripts and so, I think the move makes sense.

+1, also think that if are not using it elsewhere, as it was created specifically for e2e, we should have it under `./tests/integrations`

Issue #2683, opened 2023-02-06
capture network won't work in some cases


yanivagman:
## Description

When no network event is chosen but capture network is, pcap won't be created.
To reproduce:
`./dist/tracee -c net -t event=sched_process_exec`

## Output of `tracee -v`:

```
(paste your output here)
```

## Output of `uname -a`:

```
(paste your output here)
```

PR #2689, opened 2023-02-07
network: default to capturing all network traffic

rafaeldtinoco:
network: default to capturing all network traffic

- Before this change, only network traffic that was explicitly being traced by the user was captured.

- This change makes it so that all network traffic is captured by default and adds a command line option:

  --capture pcap-options:[none,filtered]

  to allow the old behavior, of capturing only traffic from the given filters, to still exist if "filtered" option is given.

Fixes: #2683

### DIFF
--- a/pkg/cmd/flags/capture.go
+++ b/pkg/cmd/flags/capture.go
@@ -27,27 +27,49 @@ clear-dir                                     clear the captured artifacts outpu
 Network:
 
 pcap:[single,process,container,command]       capture separate pcap files organized by single file, files per processes, containers and/or commands
+pcap-options:[none,filtered]                  network capturing options:
+                                              - none (default): pcap files containing all packets (traced/filtered or not)
+                                              - filtered: pcap files containing only traced/filtered packets
+                                                          (user needs at least 1 net_packet event to be traced)
 pcap-snaplen:[default, headers, max or SIZE]  sets captured payload from each packet:
                                               - default=96b (up to 96 bytes of payload if payload exists)
-											  - headers (up to layer 4, icmp & dns have full headers)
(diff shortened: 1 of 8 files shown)


pkg/cmd/flags/capture.go
```diff
@@ -27,27 +27,49 @@ clear-dir                                     clear the captured artifacts outpu
...
+                                              - none (default): pcap files containing all packets (traced/filtered or not)
+                                              - filtered: pcap files containing only traced/filtered packets
+                                                          (user needs at least 1 net_packet event to be traced)
 pcap-snaplen:[default, headers, max or SIZE]  sets captured payload from each packet:
                                               - default=96b (up to 96 bytes of payload if payload exists)
-											  - headers (up to layer 4, icmp & dns have full headers)
-											  - sizes ended in 'b' or 'kb' (for ipv4, ipv6, tcp, udp):
-											    256b, 512b, 1kb, 2kb, 4kb, ... (up to requested size)
-											  - max (entire packet)
+                                              - headers (up to layer 4, icmp & dns have full headers)
```
rafaeldtinoco, line 36: tabs for spaces.

pkg/ebpf/c/types.h
```diff
@@ -307,8 +307,14 @@ typedef struct config_entry {
     u64 pid_min;
 } config_entry_t;
 
+enum capture_options_e
+{
+    NET_CAP_OPT_FILTERED = (1 << 0), // pcap should obey event filters
```
rafaeldtinoco, line 312: I might enter NET_CAP_OPT_SCOPED here, for example. And maybe other future options.

geyslan (commented):
user land LGTM, soon I'll look into the bpf land. 👍🏼

geyslan (approved):

Issue #2692, opened 2023-02-08
Write pid file only at the end of the start process


josedonizetti:
The tracee-action started to lose some events after the multiscope release, debugging it, I noticed that is is because we are writing the pid file too early, and probably with multiscope the start time is taking more time than before.

PR #2693, opened 2023-02-08
tracee: fix pid file initialization

josedonizetti:
Fix https://github.com/aquasecurity/tracee/issues/2692

The pid initialization should be the last step before tracee blocks. I would actually prefer tracee lib itself didn't write this file, but instead, the binary did it, but for this we would have to refactor `Run` not to clear resources, moving the clearing to `Close`. 
So tracee would be executed like:

```
t.Run()
defer t.Close()
... (write pid, other things)
```

### DIFF
--- a/pkg/ebpf/tracee.go
+++ b/pkg/ebpf/tracee.go
@@ -460,20 +460,6 @@ func (t *Tracee) Init() error {
 		return fmt.Errorf("error initializing network capture: %w", err)
 	}
 
-	// Initialize PID file
-
-	pidFile, err := utils.OpenAt(t.outDir, "tracee.pid", syscall.O_WRONLY|syscall.O_CREAT, 0640)
-	if err != nil {
-		t.Close()
-		return fmt.Errorf("error creating readiness file: %w", err)
-	}
-
-	_, err = pidFile.Write([]byte(strconv.Itoa(os.Getpid()) + "\n"))
-	if err != nil {
-		t.Close()
-		return fmt.Errorf("error writing to readiness file: %w", err)
-	}
-
 	// Get reference to stack trace addresses map
 
 	stackAddressesMap, err := t.bpfModule.GetMap("stack_addresses")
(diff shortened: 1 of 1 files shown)


pkg/ebpf/tracee.go
```diff
@@ -1370,6 +1356,13 @@ func (t *Tracee) Run(ctx gocontext.Context) error {
 	t.bpfLogsPerfMap.Start()
 	go t.processBPFLogs()
 	t.running = true
+
+	// write pid file
+	err = t.writePid()
+	if err != nil {
+		logger.Warn("memory dump", "error", err)
```
NDStrahilevitz, line 1363: Error message seems wrong? Also I think we should log error and return here (with `t.Close()` called), not continue without the pid file.

geyslan: Yep, we need to abort.

geyslan (dismissed):

pkg/ebpf/tracee.go
josedonizetti: If CNDR depends on it, probably best to abort in case it was not able to write?

pkg/ebpf/tracee.go
```diff
@@ -1370,8 +1356,17 @@ func (t *Tracee) Run(ctx gocontext.Context) error {
...
 	t.running = true
-	// block until ctx is cancelled elsewhere
-	<-ctx.Done()
+
+	// write pid file
+	err = t.writePid()
+	if err != nil {
+		// not able to write pid, abort
+		return err
+	} else {
```
NDStrahilevitz, line 1365: nit: no need to `else` this as return will hit from `if`

rafaeldtinoco (approved):
LGTM

NDStrahilevitz (approved):
LGTM

itaysk:
> I would actually prefer tracee lib itself didn't write this file, but instead, the binary did it, but for this we would have to refactor Run not to clear resources, moving the clearing to Close.
So tracee would be executed like:
t.Run()
defer t.Close()
... (write pid, other things)

I'm surprised this isn't the case today. should we open an issue for that?

josedonizetti:
@itaysk yes, good point, the current strategy is required by CNDR, but we can find I different way of doing it that works for both places. I'll create the issue.

PR #2697, opened 2023-02-08
k8s: bump tag to 0.11.1

josedonizetti:
bump tags for the upcoming bugfix release


rafaeldtinoco (approved):
LGTM (we should automate this soon, so the release process does it for us).

PR #2703, opened 2023-02-09
chore: generate k8s statics

josedonizetti:

There was a lot of duplication on the kubernetes deploy yamls, because we support different types of installations (vanilla, with postee, with falcosidekiq).

This PR is removing this duplication by using `kubectl kustomize`, which allows us to have a base file, and specific templates to update the base and generate different yamls. This simplifies a lot the work for upcoming changes like updating tags, and some upcoming changes to kubernetes yamls (enrichment by default, liveness probe, requests/limits).

One only needs to:1. update the base file (`deploy/kubernetes/kustomize/base/tracee.yaml`)
2. execute the script to generate all statics

### 2. Explain how to test it

1. update the base file (`deploy/kubernetes/kustomize/base/tracee.yaml`) (eg: change the docker image tag)
2. `cd deploy/kubernetes/ && ./update-static.sh` (needs kubectl installed)
3. `git status`

### 3. Other comments

This is also a first step towards automating tags bump, so we don't have to do it manually every release.


rafaeldtinoco (approved):
LGTM

rafaeldtinoco:
> This is also a first step towards automating tags bump, so we don't have to do it manually every release.

Thanks for that! =D

Issue #2705, opened 2023-02-10
When filtering for 2 derived events from the same base event, only the last derived is printed (twice)


rafaeldtinoco:
Network events heavily use derived events. 

First, the base event is sent to sink -> printer and then it is used in the deriving logic. The derive logic might have more than 1 type of derived event to be sent (for the networking case, the DNS event might generate up to 3 event types (DNS, DNS Request, DNS Response), with 1 or more events per type, if the derivation generates an array of events).

```go
case event := <-in:
	// Get a copy of our event before sending it down the pipeline.
	// This is needed because later modification of the event (in
	// particular of the matched scopes) can affect the derivation
	// and later pipeline logic acting on the derived event.
	eventCopy := *event
	out <- event

	// Derive event before parse its arguments
	derivatives, errors := t.eventDerivations.DeriveEvent(eventCopy)
```` 

The problem happens in the loop:

```
for _, derivative := range derivatives {
```

at the end of it, we pass the pointer of `derivative` to the out channel:

```
out <- &derivative
```

But the pointer of the derivative variable will point to:

1. `derivatives[0]`
2. `derivatives[1]`
3. `derivatives[2]`

By the time the out channel is "printed out", the pointer will have the last address and the last derived event will be the one printed out (3 times in this case).

This can be observed with the following command line:

```
$ sudo ./dist/tracee --install-path /tmp/tracee --output option:parse-arguments --capabilities bypass=false --trace binary=/usr/bin/dig --output format:json --trace event=net_packet_dns,net_packet_dns_request 

$ dig @1.1.1.1 rafaeldtinoco.com
```

The way it is now:

```
[...]

PR #2706, opened 2023-02-10
pipeline: fix multiple derived events logic

rafaeldtinoco:

Currently, if one event derives multiple event types, like:

```go
events.NetPacketDNSBase: {
	events.NetPacketDNS: {
		Enabled:...
		DeriveFunction:...
	},
	events.NetPacketDNSRequest: {
		Enabled:...
		DeriveFunction:...
	},
	events.NetPacketDNSResponse: {
		Enabled:...
		DeriveFunction:...
	},
},
```

And more than 1 type of derived event is being filtered:

--trace event=net_packet_dns,net_packet_dns_request

Then only the last one of them will be printed (and twice in this case).

This patch fixes that behavior by giving the correct pointer to the *trace.Event channel (instead of giving one pointer that gets *overwritten).

Fixes: #2705

### 2. Explain how to test it

```
$ sudo ./dist/tracee --install-path /tmp/tracee --output option:parse-arguments --capabilities bypass=false --trace binary=/usr/bin/dig --output format:json --trace event=net_packet_dns,net_packet_dns_request 

$ dig @1.1.1.1 rafaeldtinoco.com
```

Only `net_packet_dns_request` is printed out (and two times) instead of `net_packet_dns` AND `net_packet_dns_request`.

### DIFF
--- a/pkg/ebpf/events_pipeline.go
+++ b/pkg/ebpf/events_pipeline.go
@@ -419,7 +419,9 @@ func (t *Tracee) processEvents(ctx context.Context, in <-chan *trace.Event) (<-c
 }
 
 // deriveEvents is the derivation pipeline stage
-func (t *Tracee) deriveEvents(ctx context.Context, in <-chan *trace.Event) (<-chan *trace.Event, <-chan error) {
+func (t *Tracee) deriveEvents(ctx context.Context, in <-chan *trace.Event) (
+	<-chan *trace.Event, <-chan error,
+) {
 	out := make(chan *trace.Event)
 	errc := make(chan error, 1)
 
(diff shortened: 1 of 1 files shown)


pkg/ebpf/events_pipeline.go
```diff
@@ -462,9 +461,14 @@ func (t *Tracee) deriveEvents(ctx context.Context, in <-chan *trace.Event) (<-ch
...
 					}
-					out <- &derivative
-				}
 
+					// Passing "derivative" variable here will make the ptr
+					// address always be the same as the last item. This makes
+					// the printer to print 2 or 3 times the last event, instead
+					// of printing all derived events (when there are more than
+					// one).
+					out <- &derivatives[i]
```
rafaeldtinoco, line 470: This is the change, the rest is opportunistic styling.

geyslan (approved):
LGTM

PR #2709, opened 2023-02-10
Multiple printers

josedonizetti:

The code here is still a draft, but I want to discuss if you all agree with the syntax. This PR adds support for multiple printers, for example, a user can have `table` to `stdout`, and `json` to a file, or/and `gob` to a file.

Now, when you use `tracee` you are able to send the output to either `stdout`, or an output file. For example:

```
tracee  # prints table to stdout
tracee --output table # prints table to stdout
tracee --output json # prints json to stdout
tracee --output out-file:/tmp/test # prints table to /tmp/test
tracee --output json --output out-file:/tmp/test # prints json to /tmp/test
```

So, you can pick one format in which the default is `table` and one output path in which the default is `stdout`.

The PR allow you to have multiple printer with a new syntax. For example:

```
tracee # prints table to stdout
tracee --output format:table # prints table to stdout
tracee --output json # prints json to stdout

tracee --output out-file:/tmp/test # prints json to /tmp/test
tracee --output out-file:/tmp/test,format:table # prints table to /tmp/test
tracee --output out-file:/tmp/test,gob # prints gob to /tmp/test

tracee --output table --output out-file:/tmp/test # prints table to stdout, and prints json to /tmp/test
tracee --output table --output out-file:/tmp/test,gob # prints table to sdout, and prints gob to /tmp/test
tracee --output table --output out-file:/tmp/test,gob  --output out-file:/tmp/test2,format:json # prints table to sdout, and prints gob to /tmp/test, and prints json to /tmp/test2
``` 

For the new syntax, when you pass a format directly to output, it only reflects `stdout`, for `out-file` they have their own default which is json, and allow customization of this format with the syntax `out-file:FILE[,format]`

### DIFF
--- a/cmd/tracee-ebpf/main.go
+++ b/cmd/tracee-ebpf/main.go
@@ -44,14 +44,14 @@ func main() {
 				return cli.ShowAppHelp(c) // no args, only flags supported
 			}
 
-			flags.PrintAndExitIfHelp(c)
+			flags.PrintAndExitIfHelp(c, false)
 
 			if c.Bool("list") {
 				cmd.PrintEventList(false) // list events
 				return nil
 			}
 
-			runner, err := urfave.GetTraceeRunner(c, version)
+			runner, err := urfave.GetTraceeRunner(c, version, false)
 			if err != nil {
 				return err
 			}
--- a/cmd/tracee/main.go
+++ b/cmd/tracee/main.go
@@ -47,7 +47,7 @@ func main() {
 				return cli.ShowAppHelp(c) // no args, only flags supported
 			}
 
-			flags.PrintAndExitIfHelp(c)
+			flags.PrintAndExitIfHelp(c, true)
 
 			// Rego command line flags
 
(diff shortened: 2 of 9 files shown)


geyslan:
I think this a great idea. 👍🏼 

My only concern is the non-standard syntax that we're managing across the entire project.

> tracee --output table --output out-file:/tmp/test,gob # prints table to sdout, and prints gob to /tmp/test

In the first we have `--output table` which implies the default stdout.
In the second we have `--output out-file:/tmp/test,gob` which implies an output file (not stdout) and a format.

I think mixing format and output end on the left part of the flag is a bit confusing.

What do you think of standardizing like:

```shell
--output table:stdout   # default behaviour
--output json:stdout
--output gob:stdout
```

```shell
--output table:/file
--output json:/file
--output gob:/file
```

Doing that we can use `--output json:{files}` to output to multiple outpus separating the files with comma:

```
--output json:/file1,/file2
```

or ask for different outputs in different formats like

```
--output json:/file1,/file2 --output table:stdout,/file3 --output gob:/file4,./file5
```

Thus, we force the user to ask specifically and avoid to mix format and output type, what makes the parsing easier.

josedonizetti:
@geyslan cool, I like the new syntax. But it is a breaking change, right?  I was trying to be compatible with what we had, but we could implement this only for the new binary.

PR #2710, opened 2023-02-10
Latest fixes

rafaeldtinoco:

commit 6da420d8 (HEAD -> net-corrupt-fix, rafaeldtinoco/latest-fixes)
Date:   Sun Feb 12 04:09:15 2023

    network: fix network packets corruption
    
    Before this patch, the skb timestamp was used as the key for a map
    shared between cgroup/skp program and the kprobe program responsible
    for calling it. The reason why this is needed is because there is
    absolute no helper available inside the cgroup/skb context in kernels
    below 5.6. The technique used here is to pre-process all the event
    in a kprobe, before the ingress/egress cgroup/skb programs, and simply
    submit the result (together with the patcket headers/data) from the
    ingress/egress programs.
    
    Unfortunately preemption is enabled in between the kprobe and ingress
    or egress programs. This means that the map can't be a per cpu map,
    since the task can migrate cpus, and the key that indexes the
    pre-processed work has to be known inside the cgroup/skb program, which
    is very hard without having bpf helpers to use things such as host tid,
    and similar keys.
    
    Unfortunately only skb timestamp as the key is also not enough. There
    are times that the timestamp is zeroed and it would cause conflicts and
    make tracee to generate errors such as:
    
        {"L":"ERROR","T":"2023-02-10T16:31:57.883-0300","M":"tracee
            encountered an error","error":"failed to derive event 2006:
            resource record length exceeds data"}
    
        {"L":"ERROR","T":"2023-02-10T16:31:57.883-0300","M":"tracee
            encountered an error","error":"failed to derive event 2007:
            resource record length exceeds data"}
    
        {"L":"ERROR","T":"2023-02-10T16:31:57.883-0300","M":"tracee
            encountered an error","error":"failed to derive event 2008:
            resource record length exceeds data"}
    
    The only values shared in between the kprobe and the cgroup/skb programs
    are either the skb timestamp or skb data pointer value. So, instead of
    relying only on the skb timestamp, we also use the skb data pointer now
    to find the pre-processed work.
    
    Results filtering for net_packet_ipv4 seem satisfactory:
    
    [  5] local 127.0.0.1 port 33388 connected to 127.0.0.1 port 5201
    [ ID] Interval           Transfer     Bitrate         Retr  Cwnd
    [  5]   0.00-1.00   sec  1.25 GBytes  10.8 Gbits/sec    0    895 KBytes
    [  5]   1.00-2.00   sec  1.23 GBytes  10.6 Gbits/sec    0   1.75 MBytes
    [  5]   2.00-3.00   sec  1.26 GBytes  10.9 Gbits/sec    0   1.75 MBytes
    [  5]   3.00-4.00   sec  1.25 GBytes  10.8 Gbits/sec    0   1.75 MBytes
    [  5]   4.00-5.00   sec  1.31 GBytes  11.2 Gbits/sec    0   1.75 MBytes
    [  5]   5.00-6.00   sec   998 MBytes  8.38 Gbits/sec    0   1.75 MBytes
    [  5]   6.00-7.00   sec  1.08 GBytes  9.26 Gbits/sec    0   1.75 MBytes
    [  5]   7.00-8.00   sec  1.01 GBytes  8.68 Gbits/sec    0   1.75 MBytes
    [  5]   8.00-9.00   sec  1.33 GBytes  11.5 Gbits/sec    0   1.75 MBytes
    [  5]   9.00-10.00  sec  1.09 GBytes  9.32 Gbits/sec    0   1.81 MBytes
    - - - - - - - - - - - - - - - - - - - - - - - - -
    [ ID] Interval           Transfer     Bitrate         Retr
    [  5]   0.00-10.00  sec  11.8 GBytes  10.1 Gbits/sec    0   sender
    [  5]   0.00-10.00  sec  11.8 GBytes  10.1 Gbits/sec        receiver
    
    In regards to throughput and amount of packets processed (and lost due
    to this technique).
    
    This patch also removes the kretprobe after the cgroup/skb program runs,
    since it is not needed anymore.

commit 761bea50
Date:   Fri Feb 10 15:33:26 2023

    network: fix segfault in DNS parser
    
    A DNS packet might not have a question section, in cases where the
    event gets corrupted (because of N reasons). This patch fixes the
    issue.

### 2. Explain how to test it

Terminal 1, execute tracee:

```
$ sudo ./dist/tracee --pprof --metrics --log debug --install-path /tmp/tracee --output option:parse-arguments --capabilities bypass=false --output none --trace comm!=node --trace event=net_packet_dns,net_packet_dns_request,net_packet_dns_response,hooked_proc_fops,hooked_syscalls,init_module,magic_write,mem_prot_alert,process_vm_writev,ptrace,sched_process_exec,security_bprm_check,security_file_open,security_inode_rename,security_kernel_read_file,security_sb_mount,security_socket_connect,socket_dup
```

Terminal 2, 3, 4 and 5, execute a dig loop:

```
$ while true; do dig rafaeldtinoco.com; done
```

> make sure resolutions are made locally to systemd-resolved (to its fast).

Terminal 6 and 7, execute a "ps -ef" loop:

```
$ while true; do ps -ef; done
```

Let the workload run for some time and observe tracee output some errors like:

```
    {"L":"ERROR","T":"2023-02-10T16:31:57.883-0300","M":"tracee
        encountered an error","error":"failed to derive event 2006:
        resource record length exceeds data"}
```

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -3891,6 +3891,7 @@ int BPF_KPROBE(trace_do_sigaction)
 //
 
 // clang-format off
+
 static __always_inline bool is_socket_supported(struct socket *sock)
 {
     struct sock *sk = (void *) BPF_READ(sock, sk);
@@ -3913,13 +3914,10 @@ static __always_inline bool is_socket_supported(struct socket *sock)
 
     return 1; // supported
 }
-// clang-format on
 
 //
 // network related maps
 //
-
-// clang-format off
 
 // cgroupctxmap
 
@@ -3958,8 +3956,8 @@ typedef struct net_event_context {
 
 struct {
     __uint(type, BPF_MAP_TYPE_LRU_HASH);
-    __uint(max_entries, 1024);          // simultaneous cgroup/skb ingress/eggress progs
-    __type(key, u64);                   // sk_buff timestamp
+    __uint(max_entries, 10240);         // simultaneous cgroup/skb ingress/eggress progs
+    __type(key, u64);                   // sk_buff timestamp AND skb data pointer
     __type(value, net_event_context_t); // event context built so cgroup/skb can use
 } cgrpctxmap SEC(".maps");              // saved info between SKB caller and SKB program
 
@@ -3974,7 +3972,7 @@ typedef struct net_task_context {
 
 struct {
     __uint(type, BPF_MAP_TYPE_LRU_HASH);
-    __uint(max_entries, 10240);             // simultaneous sockets being traced
+    __uint(max_entries, 65535);             // simultaneous sockets being traced
     __type(key, u64);                       // socket inode number ...
     __type(value, struct net_task_context); // ... linked to a task context
 } inodemap SEC(".maps");                    // relate sockets and tasks
@@ -3987,7 +3985,7 @@ typedef struct entry {
 
 struct {
     __uint(type, BPF_MAP_TYPE_LRU_HASH);
-    __uint(max_entries, 1024);       // simultaneous tasks being traced for entry/exit
+    __uint(max_entries, 10240);      // simultaneous tasks being traced for entry/exit
     __type(key, u32);                // host thread group id (tgid or tid) ...
     __type(value, struct entry);     // ... linked to entry ctx->args
 } entrymap SEC(".maps");             // can't use args_map (indexed by existing events only)
@@ -4016,8 +4014,6 @@ struct {
     __type(key, u32);                          // per cpu index ... (always zero)
     __type(value, event_data_t);               // ... linked to a scratch area
 } net_heap_event SEC(".maps");
-
-// clang-format on
 
 //
 // support functions for network code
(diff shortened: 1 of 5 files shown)


pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3957,8 +3957,8 @@ typedef struct net_event_context {
 } net_event_context_t;
 
 struct {
-    __uint(type, BPF_MAP_TYPE_LRU_HASH);
-    __uint(max_entries, 1024);          // simultaneous cgroup/skb ingress/eggress progs
```
rafaeldtinoco, line 3961: Raising to be cautious.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3986,8 +3986,8 @@ typedef struct entry {
 } entry_t;
 
 struct {
-    __uint(type, BPF_MAP_TYPE_LRU_HASH);
```
rafaeldtinoco, line 3989: ditto

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3974,7 +3974,7 @@ typedef struct net_task_context {
 
 struct {
     __uint(type, BPF_MAP_TYPE_LRU_HASH);
-    __uint(max_entries, 10240);             // simultaneous sockets being traced
```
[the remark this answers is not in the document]
yanivagman: What is the increase in memory?
net_task_context is quite big so seems to be more than a few MB, no?

yanivagman (approved):
LGTM
We need to calculate our maps memory footprint.
At some point, this may become a problem

geyslan (approved):
LGTM

I ran it for a while and got no error.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4315,51 +4303,54 @@ int BPF_KPROBE(cgroup_bpf_run_filter_skb)
...
-    bpf_map_delete_elem(&entrymap, &host_tid);
-
-    // use skb timestamp as the key for cgroup/skb
-    u64 skbts = BPF_READ(skb, tstamp);
+    if (skbts) // first attempt with skb timestamp
+        bpf_map_update_elem(&cgrpctxmap, &skbts, &neteventctx, BPF_NOEXIST);
 
-    // delete netctx after cgroup ebpf program runs
-    bpf_map_delete_elem(&cgrpctxmap, &skbts);
+    if (data) // second attempt will be with skb data pointer
```
NDStrahilevitz, line 4352: There is no termination after the first attempt, by design? Or are the field mutually exclusive on being NULL (if `data` is NULL then `skbts` necessarily isn't and vice versa)?

pkg/ebpf/c/tracee.bpf.c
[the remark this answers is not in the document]
NDStrahilevitz: 1. What is the 65k number based on (maybe we should make it configurable)?
2. With regards to memory: would it be possible (probably in a future PR) to use a `*task_context_t` instead?

rafaeldtinoco:
This is all wrong. I'm closing this and opening a new one.

PR #2713, opened 2023-02-13
tracee: fix args on signatures events

josedonizetti:

This PR changes signature events to use `finding.Data` as arguments instead of the original event arguments. 
Also, it fix the filtering based on such arguments for signature events.

This PR is part of epic: https://github.com/aquasecurity/tracee/issues/2355

### 2. Explain how to test it

The simplest way to test is to change an existing signature to include same [output data](https://github.com/aquasecurity/tracee/blob/main/signatures/golang/anti_debugging_ptraceme.go#L70), recompile tracee, and test out the feature, eg:

```
tracee --trace event=anti_debugging --trace anti_debugging.args.arg1=test
```

### DIFF
--- a/pkg/ebpf/finding.go
+++ b/pkg/ebpf/finding.go
@@ -22,15 +22,16 @@ func FindingToEvent(f detect.Finding) (*trace.Event, error) {
 		return nil, fmt.Errorf("error finding event not found: %s", f.SigMetadata.EventName)
 	}
 
-	return newEvent(int(eventID), f.SigMetadata, s), nil
+	return newEvent(int(eventID), f, s), nil
 }
 
-func newEvent(id int, sigMetadata detect.SignatureMetadata, s trace.Event) *trace.Event {
-	metadata := getMetadataFromSignatureMetadata(sigMetadata)
+func newEvent(id int, f detect.Finding, s trace.Event) *trace.Event {
+	arguments := getArguments(f)
+	metadata := getMetadataFromSignatureMetadata(f.SigMetadata)
 
 	return &trace.Event{
 		EventID:             id,
-		EventName:           sigMetadata.Name,
+		EventName:           f.SigMetadata.Name,
 		Timestamp:           s.Timestamp,
 		ThreadStartTime:     s.ThreadStartTime,
 		ProcessorID:         s.ProcessorID,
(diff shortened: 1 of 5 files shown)


pkg/ebpf/finding.go
```diff
@@ -50,11 +53,29 @@ func newEvent(id int, name string, s trace.Event) *trace.Event {
...
 }
+
+func getArguments(f detect.Finding) []trace.Argument {
+	arguments := make([]trace.Argument, 0, len(f.Data))
+
+	for k, v := range f.Data {
+		arg := trace.Argument{
+			ArgMeta: trace.ArgMeta{
+				Name: k,
+				Type: reflect.TypeOf(v).Name(),
```
geyslan, line 72: I think we have specific C types names. Take a look at `GetParamType()` and `ReadArgFromBuff()`.

pkg/ebpf/finding_test.go
```diff
@@ -0,0 +1,113 @@
...
+		PodNamespace:        "namespace",
+		PodUID:              "uid",
+		ReturnValue:         0,
+		MatchedScopes:       1,
+		ArgsNum:             2,
+		Args: []trace.Argument{
+			{
+				ArgMeta: trace.ArgMeta{
+					Name: "arg1",
+					Type: "string",
```
geyslan, line 43: `char*`?

rafaeldtinoco (changes requested):
This needs rebase. 

@josedonizetti is this still valid ?

With the discussion at https://github.com/aquasecurity/tracee/issues/2355#issuecomment-1424770772 and the about to be merged (https://github.com/aquasecurity/tracee/pull/2752) I wasn't sure (since the the Event type is being changed there to have metadata, as the discussion suggests).

josedonizetti:
@rafaeldtinoco yes, still valid. The Metadata PR will only reflect `SignatureMetadata`, the user still needs access to `finding.Data` from a signature detection, which will be passed as arguments, and different than `metadata` it will be filterable.

pkg/ebpf/finding.go
```diff
@@ -50,11 +52,59 @@ func newEvent(id int, name string, s trace.Event) *trace.Event {
...
+			Value: v,
+		}
+
+		arguments = append(arguments, arg)
+	}
+
+	return arguments
+}
+
+// TODO: we probably should have internal types instead of using kernel, or golang types
```
geyslan, line 82: +1

pkg/events/events.go
```diff
@@ -80,6 +80,17 @@ type Event struct {
...
+func (e *Event) IsASignatureEvent() bool {
+
+	for _, s := range e.Sets {
+		if s == "signatures" {
+			return true
+		}
+	}
+
+	return false
+}
```
geyslan, line 92: This isn't yet being used. Will it be checked in a hot path? If so, could it be a preset boolean flag?

pkg/ebpf/finding.go
```diff
@@ -50,11 +52,59 @@ func newEvent(id int, name string, s trace.Event) *trace.Event {
...
+		return "unsigned long long"
+	case string:
+		return "const char *"
+	case bool:
+		return "bool"
+	case float32:
+		return "float"
+	case float64:
+		return "long double"
+	default: // TODO: how to implement int8, uint8, pointers, slices, and maps
```
geyslan, line 107: I guess that's trivial by using our own internal types.

geyslan (approved):
LGTM

pkg/events/events.go
josedonizetti: @geyslan This is used to avoid the argument name validation when booting tracee, for example, if you do `-t security_open_file.args.pathname!=X`, there is a validation to check if the `pathaname` is a paramenter declared on the event when tracee is starting. 

As signature events the parameters are dynamic, we are skipping the validation so we don't fail the boot: https://github.com/aquasecurity/tracee/pull/2713/files#diff-a15ccb695c12a7219d07ca60dd20a3e26df991cb65d2313dcd593b5b5295a55cR104

yanivagman (approved):
LGTM

PR #2714, opened 2023-02-13
Add hidden linux kernel module event

OriGlassman:


rafaeldtinoco (changes requested):
You're missing test results and a better PR description.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -655,6 +658,234 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
     return 0;
 }
 
+void update_hidden_modules(u64 mod_addr, char *mod_name)
```
roikol, line 661: i think that the method that was used to detect the hidden module is valuable.
for example, adding an argument here that states the method (procfs / kset / tree) and add it to the `kernel_module_t` key.

then in user space you can add it as an argument to the event, or create 3 different derived events.

WDYT?

note: if you decide to do this, i think that `already_reported_hidden_modules_map` map key should also contain the method. i think it would be valuable to see that kernel module is hidden from procfs, and mod_tree, but not module_kset for example.

pkg/ebpf/c/maps.h
```diff
@@ -114,6 +124,12 @@ BPF_LRU_HASH(bpf_attach_tmp_map, u32, bpf_attach_t, 1024);         // temporaril
 BPF_PERCPU_ARRAY(event_data_map, event_data_t, 1);                 // persist event related data
 BPF_HASH(logs_count, bpf_log_t, bpf_log_count_t, 4096);            // logs count
 BPF_PERCPU_ARRAY(scratch_map, scratch_t, 1);                       // scratch space to avoid allocating stuff on the stack
+BPF_QUEUE(found_hidden_modules_queue, kernel_module_t, 1024);      // modules that were found to be hidden and to be reported to userspace
+BPF_HASH(modules_from_procfs, u64, kernel_module_name_t, 1024);    // modules returned by reading /proc
+BPF_QUEUE(walk_mod_tree_queue, rb_node_t, 1024);                   // used to walk a rb tree
+BPF_HASH(unhidden_modules_map, u64, u64, 1024);                    // modules that are shown (not hidden)
+BPF_HASH(already_reported_hidden_modules_map, u64, u64, 1024);     // the hidden modules that we already reported
```
roikol, line 131: i believe this map should be LRU because there's no place to remove from it.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -101,6 +101,9 @@ char LICENSE[] SEC("license") = "GPL";
 int KERNEL_VERSION SEC("version") = LINUX_VERSION_CODE;
 #endif
 
+#define MAX_HIDDEN_MODS 3
+#define MAX_NUM_MODS    1000 / 5
```
roikol, line 105: can you add a comment saying where these numbers come from?

pkg/ebpf/c/vmlinux.h
```diff
@@ -233,6 +233,33 @@ struct list_head {
...
 };
 
+#define list_entry(ptr, type, member) container_of(ptr, type, member)
+
+#define list_next_entry(pos, member)                                                               \
+    list_entry(READ_KERN((pos)->member.next), typeof(*(pos)), member)
+
+#define list_first_entry(ptr, type, member) list_entry(READ_KERN((ptr)->next), type, member)
+
+#define list_entry_is_head(pos, head, member) (&pos->member == (head))
```
roikol, line 243: are these macros used?

list_entry_is_head
list_for_each_entry
end_list_for_each_entry

pkg/ebpf/c/vmlinux.h
```diff
@@ -661,11 +689,66 @@ struct dentry {
...
+} seqcount_latch_t;
+
+struct latch_tree_root {
+    seqcount_latch_t seq;
+    struct rb_root tree[2];
+};
+
+struct mod_tree_root {
+    struct latch_tree_root root;
+    unsigned long addr_min;
```
roikol, line 745: i think addr fields are not used?

pkg/ebpf/c/vmlinux.h
```diff
@@ -661,11 +689,66 @@ struct dentry {
...
     const char *version;
     const char *srcversion;
+    struct module_kobject mkobj;
+    struct module_layout core_layout;
+};
+
+#define rb_entry(ptr, type, member) container_of(ptr, type, member)
+
+struct rb_node {
+    unsigned long __rb_parent_color;
```
roikol, line 717: i think this field is not used?

pkg/ebpf/c/vmlinux.h
```diff
@@ -661,11 +689,66 @@ struct dentry {
...
 struct module {
     struct list_head list;
     char name[MODULE_NAME_LEN];
     const char *version;
     const char *srcversion;
+    struct module_kobject mkobj;
+    struct module_layout core_layout;
+};
+
+#define rb_entry(ptr, type, member) container_of(ptr, type, member)
```
roikol, line 714: i think this macro is not used?

types/trace/trace.go
```diff
@@ -212,6 +212,11 @@ type HookedSymbolData struct {
 	ModuleOwner string
 }
 
+type HiddenKernelModule struct {
```
roikol, line 215: since this is on package `trace`, this should be added in a different PR and merged before the reset of the PR.

so one PR with this new type.
another PR with all the rest, and also update the `trace` package in go.mod

roikol (changes requested):
hey @OriGlassman 

good work here!
i've left some comments for you to consider.

please also use clang-format in the .c and .h files. and use go-fmt on the to .go files.

roikol:
depends on: #2742

once it merges, `trace` package should be updated in go.mod

OriGlassman:
@roikol @rafaeldtinoco Ready for merge from my side, thanks!

pkg/events/events.go
```diff
@@ -5759,6 +5765,46 @@ var Definitions = eventDefinitions{
...
 			},
 		},
+		HiddenKernelModule: {
+			ID32Bit: sys32undefined,
+			Name:    "hidden_kernel_module",
+			Dependencies: dependencies{
+				Events: []eventDependency{
+					{EventID: HiddenKernelModuleSeeker},
+				},
+				TailCalls: []TailCall{
```
roikol, line 5775: i believe those tail calls should be under the `HiddenKernelModuleSeeker` event

roikol (approved):
LGTM!
we need to run the tests though

rafaeldtinoco:
Alright, thanks @OriGlassman, I'll review this soon!

pkg/ebpf/c/maps.h
```diff
@@ -25,6 +25,13 @@
         __type(value, _value_type);                                                                \
     } _name SEC(".maps");
 
+#define BPF_MAP_QUEUE(_name, _type, _value_type, _max_entries)                                     \
+    struct {                                                                                       \
+        __uint(type, _type);                                                                       \
+        __uint(max_entries, _max_entries);                                                         \
+        __type(value, _value_type);                                                                \
+    } _name SEC(".maps");
+
```
yanivagman, line 34: No need for this macro - `BPF_MAP` macro should suffice

pkg/ebpf/c/maps.h
```diff
@@ -106,6 +120,7 @@ BPF_PROG_ARRAY(sys_enter_submit_tail, MAX_EVENT_ID);               // store prog
 BPF_PROG_ARRAY(sys_exit_submit_tail, MAX_EVENT_ID);                // store program for submitting syscalls from sys_exit
 BPF_PROG_ARRAY(sys_enter_init_tail, MAX_EVENT_ID);                 // store program for performing syscall tracking logic in sys_enter
 BPF_PROG_ARRAY(sys_exit_init_tail, MAX_EVENT_ID);                  // store program for performing syscall tracking logic in sys_exits
+BPF_PROG_ARRAY(hidden_kernel_mod_prog_arr, MAX_TAIL_CALL);         // store program for hidden kernel module logic
```
yanivagman, line 123: is this map required? why not using prog_array map?

pkg/ebpf/c/maps.h
```diff
@@ -114,6 +129,12 @@ BPF_LRU_HASH(bpf_attach_tmp_map, u32, bpf_attach_t, 1024);         // temporaril
 BPF_PERCPU_ARRAY(event_data_map, event_data_t, 1);                 // persist event related data
 BPF_HASH(logs_count, bpf_log_t, bpf_log_count_t, 4096);            // logs count
 BPF_PERCPU_ARRAY(scratch_map, scratch_t, 1);                       // scratch space to avoid allocating stuff on the stack
+BPF_QUEUE(found_hidden_modules_queue, kernel_module_t, 1024);      // modules that were found to be hidden and to be reported to userspace
+BPF_HASH(modules_from_procfs, u64, kernel_module_name_t, 2048);    // modules returned by reading /proc
+BPF_QUEUE(walk_mod_tree_queue, rb_node_t, 2048);                   // used to walk a rb tree
+BPF_HASH(unhidden_modules_map, u64, u64, 2048);                    // modules that are shown (not hidden)
+BPF_LRU_HASH(already_reported_hidden_modules_map, u64, u64, 1024); // the hidden modules that we already reported
```
yanivagman, line 136: Maps we have today are generic, meaning that they are used by more than one event.
Here you add 5 new maps just for 1 event, which will be created on any invocation of tracee, even if the event was not chosen. Few things to consider:
1. Reduce the amount of used maps
2. Move these maps declaration near to the event that use them
3. We need to find a way to load these maps only when this event is requested by the user

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -656,6 +656,306 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
     return 0;
 }
 
+void __always_inline update_hidden_modules(u64 mod_addr, char *mod_name)
+{
+    if (bpf_map_lookup_elem(&already_reported_hidden_modules_map, &mod_addr) != NULL) {
```
yanivagman, line 661: Such check can be done in userspace.
This event is not too frequent, so shouldn't be a problem to send to userspace and check there. This will allow us to remove the already_reported_hidden_modules_map

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -656,6 +656,306 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
...
+        return; // already reported this module - no need to report again
+    }
+
+    kernel_module_t hidden_mod = {.addr = mod_addr};
+    bpf_probe_read_str(hidden_mod.name, MODULE_NAME_LEN, mod_name);
+
+    bpf_map_push_elem(&found_hidden_modules_queue, &hidden_mod, BPF_EXIST);
```
yanivagman, line 668: Why do we need this queue? Can't you just send an event per module?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -656,6 +656,306 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
...
+{
+    struct list_head *head = (struct list_head *) get_symbol_addr("modules");
+    bool iterated_all_modules = false;
+    struct module *pos, *n;
+
+    pos = list_first_entry(head, typeof(*pos), list);
+    n = list_next_entry(pos, list);
+
+#pragma unroll
+    for (int i = 0; i < MAX_NUM_MODS; i++) {
```
yanivagman, line 692: No way this loop works for kernel version 4.19

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -656,6 +656,306 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
...
+        }
+
+        pos = n;
+        n = list_next_entry(n, entry);
+    }
+
+    return !finished_iterating;
+}
+
+static __always_inline u32 prepare_hidden_mods_output(u64 *addrs,
```
yanivagman, line 753: by sending an event per each discovered hidden module, this (and other) step can be removed

yanivagman (changes requested):

pkg/ebpf/c/maps.h
OriGlassman: Tried that at first - supplying a key for BPF_MAP_TYPE_QUEUE maps seems to throw an error:
{"level":"warn","ts":1677430230.0956712,"msg":"libbpf: map 'found_hidden_modules_queue': failed to create: Invalid argument(-22)"}

pkg/ebpf/c/maps.h
OriGlassman: Not required - switched to prog_array

pkg/ebpf/c/tracee.bpf.c
OriGlassman: Yep, I think I can change it to be an event per hidden module

rafaeldtinoco:
These needs rebase AND addressing comments from Yaniv. Also, after the rebase, new testers will be used (https://github.com/aquasecurity/tracee/pull/2776) to test code.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -656,6 +664,382 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
...
+}
+
+// Populate the queue with the current modules in the list of modules.
+// We can't read it once and then hook on do_init_module and free_module since a hidden module will
+// remove itself from the list directly and we wouldn't know (hence from our perspective the module
+// will reside in the modules list, which could be false). So on every trigger, we go over the
+// modules list and populate the queue. This is a queue since we need to delete all elements so the
+// next run will be clean. Since this mechanism is suppose to be triggered every once in a while,
+// this should be ok.
+static __always_inline bool init_shown_modules_queue()
```
roikol, line 688: you can rename the func as it is not a queue anymore

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -656,6 +664,382 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
...
+            break;
+        }
+
+        bpf_map_update_elem(&unhidden_modules_map, &pos, &empty, BPF_ANY);
+
+        pos = n;
+        n = list_next_entry_ebpf(n, list);
+    }
+
+    return !iterated_all_modules; // false is valid value
```
roikol, line 712: i think it would be clearer to just return 0 instead of breaking, and return 1 at the end of func. your choice though, its not a big deal..

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -656,6 +664,382 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
...
+        }
+
+        pos = n;
+        n = list_next_entry_ebpf(n, entry);
+    }
+
+    return !finished_iterating;
+}
+
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 20, 0)
```
roikol, line 761: must check for CORE as well

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -656,6 +664,382 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
...
+                Now, it's right subtree's turn */
+                curr = READ_KERN(curr->rb_right);
+            }
+        }
+    }
+
+    return !finished_iterating;
+}
+
+    #ifdef CORE
```
roikol, line 802: why only in CORE? what happens if you are not in CORE and kernel version higher the 4.20?

pkg/ebpf/c/vmlinux.h
```diff
@@ -669,13 +670,64 @@ struct dentry {
     struct inode *d_inode;
 };
 
+enum bpf_func_id
```
roikol, line 673: could you also move BPF_FUNC_probe_write_user and BPF_FUNC_override_return here from missing_definitions.h?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -656,6 +664,382 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
     return 0;
 }
 
+// If we won't create these maps, the userspace needs to be kernel version aware as well,
+// otherwise it'll try to interact with the maps and report an error
+BPF_HASH(unhidden_modules_map, u64, u64, 2048); // modules that are shown (not hidden)
```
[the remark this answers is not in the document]
OriGlassman: Right you are!

rafaeldtinoco:

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -102,6 +102,14 @@ char LICENSE[] SEC("license") = "GPL";
...
 #endif
 
+#define list_entry_ebpf(ptr, type, member) container_of(ptr, type, member)
+
+#define list_next_entry_ebpf(pos, member)                                                          \
+    list_entry_ebpf(READ_KERN((pos)->member.next), typeof(*(pos)), member)
+
+#define list_first_entry_ebpf(ptr, type, member)                                                   \
+    list_entry_ebpf(READ_KERN((ptr)->next), type, member)
+
```
yanivagman, line 112: Such helpers should be defined in one of the headers and not here

pkg/ebpf/tracee.go
```diff
@@ -1355,6 +1484,9 @@ func (t *Tracee) Run(ctx gocontext.Context) error {
 	if err != nil {
 		logger.Warn("Memory dump", "error", err)
 	}
+
+	go t.lkmSeekerRoutine()
```
yanivagman, line 1488: This should only run if the hidden kernel module event is selected

pkg/ebpf/tracee.go
```diff
@@ -1346,6 +1355,126 @@ func (t *Tracee) initBPF() error {
...
+		}
+
+		t.triggerKernelModuleSeeker()
+
+		r := rand.Intn(300) + 10 // min 10 seconds sleep
+		time.Sleep(time.Duration(r) * time.Second)
+	}
+}
+
+func (t *Tracee) fillModulesFromProcFs(modulesFromProcFs *bpf.BPFMap) error {
```
yanivagman, line 1418: It seems that some operations here require bumping the effective capabilites.
For that, you should use the capabilities Requested() func and test when capabilitiies bypass=false (soon this will become the default)

pkg/events/derive/hidden_kernel_module.go
```diff
@@ -0,0 +1,53 @@
+package derive
```
yanivagman, line 1: Why does this event needs to be derived?
If all what you need is a place to process the arguments recieved from bpf code, a more proper place to do that will be the events processor.

Derived event mechanism should be used mostly for events we can't produce in bpf code since we need some extra data from userspace (e.g. parsing symbols or procfs)

pkg/ebpf/tracee.go
```diff
@@ -1346,6 +1355,126 @@ func (t *Tracee) initBPF() error {
 	return errfmt.WrapError(err)
 }
 
+func clearMap(bpfMap *bpf.BPFMap) error {
```
yanivagman, line 1358: To better organize code, we should place event specific code (hidden modules) in its own file and not under tracee.go

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -656,6 +664,345 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
     return 0;
 }
 
+// If we won't create these maps, the userspace needs to be kernel version aware as well,
+// otherwise it'll try to interact with the maps and report an error
```
yanivagman, line 668: I don't understand this comment. Can you please explain?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -656,6 +664,345 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
...
+BPF_HASH(modules_from_procfs,
+         u64,
+         kernel_module_name_t,
+         MAX_NUM_MODS); // modules returned by reading /proc
+
+void __always_inline send_hidden_module(u64 mod_addr, char *mod_name, program_data_t *p)
+{
+    reset_event_args(p);
+    save_to_submit_buf(p->event, &mod_addr, sizeof(u64), 0);
+    save_bytes_to_buf(p->event, (void *) mod_name, MODULE_NAME_LEN & MAX_MEM_DUMP_SIZE, 1);
```
yanivagman, line 679: It will be more clear to use `save_str_to_buf()` instead

yanivagman (changes requested):
Some comments follow.
In addition, since the BPF logic here is not trivial, I think you should add some here to validate the code works correctly on different environment and assure we will know if something gets broken in the future.

pkg/ebpf/c/maps.h
OriGlassman: u32, u64...
image

pkg/ebpf/tracee.go
OriGlassman: First thing the routine does is return if the event is not selected. Can move it out and not fire the goroutine at all, if it is preferred (performance-wise obviously, readability not sure).

pkg/events/derive/hidden_kernel_module.go
OriGlassman: Need a place to only send the events for relevant hidden modules (by using the lru.Cache).
If I can "drop" the events without setting an error, then sure. Is it possible to do so via the events processor mechanism?

pkg/ebpf/c/tracee.bpf.c
OriGlassman: Unfortunately used the bytes helper while playing 1on1 with the verifier.. Perhaps I'll add a comment to clarify though

pkg/ebpf/c/tracee.bpf.c
OriGlassman: If I won't create the maps, then the userspace will try to access them and receive an error.
Now, the question is whether this error is "ok" (kernel versions below 5.2 etc). Or perhaps we're above 5.2 but it just failed fetching the map.

rafaeldtinoco:
I want to read this one.

pkg/ebpf/c/common/consts.h
```diff
@@ -18,6 +18,7 @@
 #define SEND_META_SIZE      28
 
 #define MAX_MEM_DUMP_SIZE   127
+#define MAX_NUM_MODS        600       // number of iterations - value that the verifier was seen to cope with - the higher, the better.
```
yanivagman, line 21: This macro name is not clear enough - does MODS imply modules?
Also - why should we put this under common? isn't it only used for your event?

pkg/ebpf/c/maps.h
yanivagman: key size must be zero - https://docs.kernel.org/bpf/map_queue_stack.html

pkg/ebpf/c/tracee.bpf.c
yanivagman: Where does userspace try to access them?

pkg/ebpf/c/common/consts.h
OriGlassman: Changed it to MAX_NUM_MODULES.

Where do you recommend to put it?

pkg/ebpf/c/tracee.bpf.c
OriGlassman: unhidden_modules_map -  https://github.com/OriGlassman/tracee/blob/lkm_finder/pkg/ebpf/tracee.go#L1382 - userspace clears the map after each run.. I'm not familiar with any way to clear a (hash) map from kernelspace, prior 5.13 via this helper: https://github.com/torvalds/linux/commit/69c087ba6225b574afb6e505b72cb75242a3d844

modules_from_procfs -  https://github.com/OriGlassman/tracee/blob/lkm_finder/pkg/ebpf/tracee.go#L1388 - userspace clears the map and fill entries from /proc/modules before each kernelspace run.

pkg/events/derive/hidden_kernel_module.go
yanivagman: It doesn't matter if you place the lru cache here or there.
The advantage of having this event as a non derived event will become clear when we will add actions in bpf code. Then it will be possible to block the operation in bpf code, and not wait for the event to be derived, where it is too late to block. Anyway, it is ok if you prefer to keep it here for now.

yanivagman (changes requested):

pkg/ebpf/c/tracee.bpf.c
OriGlassman: Just to point out that the maps must be created and can't be under 
"#if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 20, 0) || defined(CORE)", unlike the queue used for example:
image

pkg/ebpf/c/maps.h
OriGlassman: So the BPF_MAP_QUEUE macro shouldn't really be used by the user, rather the BPF_QUEUE which does not need the type.
I've defined the BPF_MAP_QUEUE just as a mean to support other use cases, for example a stack. Perhaps I'll rename it to something more relevant though, and add support for stack.

pkg/ebpf/c/tracee.bpf.c
yanivagman: Ok, now I understand. Since the new logic is only intended for kernel 5.2 and above, one can think that we don't need to create these maps for lower kernel versions, but anyway we should create them to avoid getting errors when trying to access these maps from userspace (which is not aware of the kernel version). Is that correct?
If so, better to update this comment to reflect the fact that the logic (and maps) will only be used in kernels above 5.2

yanivagman (approved):
LGTM.
Nice work!

rafaeldtinoco (approved):

PR #2715, opened 2023-02-13
refactor: rules renamed to signatures

josedonizetti:

This PR renames `rules`to `signatures`where possible. The places where it didn't rename are because they are still `tracee-rules` which would be removed once `tracee` becomes the default solution and we drop support for `tracee-ebpf` and `tracee-rules`. 

The flags renamed were only done for `tracee`, `tracee-ebpf`is still using `rules-dir`and `rules`flags as we don't want to introduce a breaking change there. 

Some points of attention:

1 - make files targets are now using `signatures`
2 - we generate `./dist/signatures`instead of `./dist/rules`
3 - `pkg/rules` became `pkg/signatures`
4 - packaging had to be changed `docker, `ubuntu`, etc

### 2. Explain how to test it

```
make clean && make test-unit
make clean && make signatures
make clean && make e2e-net-signatures
make clean && make -f builder/Makefile.tracee-container build-tracee
# docker run ...
make clean && sudo ./tests/e2e-net-test.sh
```

### 3. Other comments

With the unified binary it was decide to use the name `signature events` for signatures instead of `rules`, also `rules` will have a different meaning once we introduce policies.


josedonizetti:
@NDStrahilevitz  would this refactor affect CNDR?

NDStrahilevitz:
> @NDStrahilevitz would this refactor affect CNDR?

Only in the sense that it will break import links. Kind of a headache but not going to block for it.

yanivagman (approved):
LGTM

rafaeldtinoco:
Could you pls hold the merge just until I give a fast look into this PR ? We can merge later today I believe.

rafaeldtinoco:
Reviewing this one now.

docs/contributing/building/building.md
```diff
@@ -58,14 +58,14 @@
...
-        $ make clean-rules          	# wipe ./dist/rules
+        $ make clean-signatures       # wipe ./dist/signatures
 
     # test
 
         $ make test-types           	# run unit tests for types module
         $ make test-unit            	# run unit tests
         $ make test-integration     	# run integration tests
-        $ make test-rules           	# opa test (tracee-rules)
+        $ make test-signatures        # opa test (tracee-rules)
```
rafaeldtinoco, line 68: nit: unaligned, there are tabs and spaces and that might lead to unalignment. I would replace all tabs for spaces (in between the target name and the # comment).

docs/docs/tracing/network-events.md
```diff
@@ -263,7 +263,7 @@ arguments).
 It is possible to create Golang (or Rego) signatures for the network events. If you haven't read about how to create signatures, do it [HERE](../../detecting/golang/).
 
 !!! Examples Note
-    Bellow is an example of how to create a signature for the `net_packet_dns` event. This same example is used by Tracee CI/CD tests and can be found at the [GitHub repository](https://github.com/aquasecurity/tracee/tree/main/tests/e2e-net-rules), together with some other signatures for the network events.
+    Bellow is an example of how to create a signature for the `net_packet_dns` event. This same example is used by Tracee CI/CD tests and can be found at the [GitHub repository](https://github.com/aquasecurity/tracee/tree/main/tests/e2e-net-signatures), together with some other signatures for the network events.
```
rafaeldtinoco, line 266: nit: Mind fixing "Bellow" to "Below" ? It was my mistake previously, but, since you're touching this =).

packaging/ubuntu/rules
```diff
@@ -45,9 +45,9 @@ override_dh_auto_install:
 	mkdir -m 0755 -p ./debian/tracee-rules/usr/lib/tracee/
 	install -m 0755 ./dist/tracee-rules ./debian/tracee-rules/usr/lib/tracee/
 	ln -s /usr/lib/tracee/tracee-rules ./debian/tracee-rules/usr/bin/tracee-rules
-	# rules
-	mkdir -m 0755 -p ./debian/tracee-rules/usr/lib/tracee/rules/
-	install -m 0644 ./dist/rules/* ./debian/tracee-rules/usr/lib/tracee/rules/
+	# signatures
```
rafaeldtinoco, line 48: You need to change packaging/ubuntu/control file as well. The package named `tracee-rules` should be named `tracee-signatures` now I believe. You can test generating packages for at least 1 distro (make -f builder/Makefile.packaging ubuntu-bin-focal) to make sure it works.

packaging/fedora/tracee.spec
```diff
@@ -75,21 +75,21 @@ ln -s %{_libdir}/tracee/tracee-rules $RPM_BUILD_ROOT/%{_bindir}/tracee-rules
...
 %files
 %{_bindir}/tracee
 %{_libdir}/tracee/tracee
-%{_libdir}/tracee/rules*
+%{_libdir}/tracee/signatures*
 
 %files -n tracee-ebpf
 %{_bindir}/tracee-ebpf
 
 %files -n tracee-rules
```
rafaeldtinoco, line 92: this should be named tracee-signatures (it will generate a tracee-signatures) package.. and all "tracee-rules" lines should be replaced by "tracee-signatures" as well. You can test the package generation by executing `make -f builder/Makefile.packaging fedora-bin-37` to make sure it works and generates the correct packages.

Makefile
```diff
@@ -222,14 +222,14 @@ help:
 	@echo ""
 	@echo "# build"
 	@echo ""
-	@echo "    $$ make all                      # build tracee-ebpf, tracee-rules & rules"
+	@echo "    $$ make all                      # build tracee-ebpf, tracee-rules & signatures"
 	@echo "    $$ make bpf-core                 # build ./dist/tracee.bpf.core.o"
 	@echo "    $$ make bpf-nocore               # build ./dist/tracee.bpf.XXX.o"
 	@echo "    $$ make tracee-ebpf              # build ./dist/tracee-ebpf"
 	@echo "    $$ make tracee-rules             # build ./dist/tracee-rules"
```
rafaeldtinoco, line 229: I believe you are keeping tracee rules because it will get deprecated with the single binary. That might be the reason why you didn't change tracee-rules packaging, let me go back there =).

packaging/fedora/tracee.spec
rafaeldtinoco: Ok I see now the point you want to keep tracee-rules (since it will get deprecated). Lets check if packages are good..

packaging/ubuntu/rules
rafaeldtinoco: Ok I see now the point you want to keep tracee-rules (since it will get deprecated). Lets check if packages are good..

rafaeldtinoco: ok nm, works good.

rafaeldtinoco (approved):
LGTM, very minor nits on spelling.

packaging/fedora/tracee.spec
rafaeldtinoco: nm, looks ok, got the idea.

Makefile
josedonizetti: 👍

packaging/fedora/tracee.spec
josedonizetti: 👍

PR #2716, opened 2023-02-14
network: fix network packets corruption

rafaeldtinoco:

network: fix network packets corruption

Before this patch, the skb timestamp was used as the key for a map
shared between the cgroup/skp program and the kprobe program responsible
for calling it. The reason why this was needed is that there is
absolute no helper available inside the cgroup/skb context in kernels
below 5.6. The technique used here is to pre-process all the events in
a kprobe, before the ingress/egress cgroup/skb programs, and simply
submit the result (together with the packet headers/data) from the
ingress/egress programs.

Unfortunately, preemption is enabled in between the kprobe and ingress
or egress programs. This means that the map can't be a per cpu map,
since the task can migrate cpus, and the key that indexes the
pre-processed work has to be known inside the cgroup/skb program,
which is very hard without having bpf helpers to use things such as
host tid, and similar keys.

Unfortunately, only skb timestamp as the key is also not enough. There
are times that the timestamp is zeroed and it would cause conflicts
and make tracee to generate errors such as:

    {"L":"ERROR","T":"2023-02-10T16:31:57.883-0300","M":"tracee
        encountered an error","error":"failed to derive event 2006:
        resource record length exceeds data"}

    {"L":"ERROR","T":"2023-02-10T16:31:57.883-0300","M":"tracee
        encountered an error","error":"failed to derive event 2007:
        resource record length exceeds data"}

    {"L":"ERROR","T":"2023-02-10T16:31:57.883-0300","M":"tracee
        encountered an error","error":"failed to derive event 2008:
        resource record length exceeds data"}

There were other issues as well like:

1. Logic should ignore IPv4/IPv6 dual stack (unsupported for now)
2. Logic should ignore IPIP and other tunneled protocols (for now)
3. Logic should ignore other protocols (than TCP/UDP/ICMP/ICMPv6)

The only values shared in between the kprobe and the cgroup/skb
programs are either the skb timestamp, skb data pointer value OR the
ctx pointer itself (the kprobe skb argument and ctx argument from the
cgroup/skb program, although the skb program pointer points to the VM
memory).

And there is more problem: the "sk_buff" structure available to the
cgroup/skb program can't be used as a bpf map key or value if it
contains arithmetic (like creating a hash using 2 or 3 fields from the
"sk_buff" passed as the ctx argument to the cgroup/skb program).

So, instead of relying only on the skb timestamp, this patch makes the
kprobe logic to parse sb_buff data, read the L3 header contents and
use the IP/IPv6 header fields to index the event context temporarily
in a map until the cgroup/skb program runs.

Throughput results with net_packet_ipv4 filtering (the worst case),
with "--output none", seem satisfactory:

```
[  5] local 127.0.0.1 port 33388 connected to 127.0.0.1 port 5201
[ ID] Interval           Transfer     Bitrate         Retr  Cwnd
[  5]   0.00-1.00   sec  1.25 GBytes  10.8 Gbits/sec    0    895 KBytes
[  5]   1.00-2.00   sec  1.23 GBytes  10.6 Gbits/sec    0   1.75 MBytes
[  5]   2.00-3.00   sec  1.26 GBytes  10.9 Gbits/sec    0   1.75 MBytes
[  5]   3.00-4.00   sec  1.25 GBytes  10.8 Gbits/sec    0   1.75 MBytes
[  5]   4.00-5.00   sec  1.31 GBytes  11.2 Gbits/sec    0   1.75 MBytes
[  5]   5.00-6.00   sec   998 MBytes  8.38 Gbits/sec    0   1.75 MBytes
[  5]   6.00-7.00   sec  1.08 GBytes  9.26 Gbits/sec    0   1.75 MBytes
[  5]   7.00-8.00   sec  1.01 GBytes  8.68 Gbits/sec    0   1.75 MBytes
[  5]   8.00-9.00   sec  1.33 GBytes  11.5 Gbits/sec    0   1.75 MBytes
[  5]   9.00-10.00  sec  1.09 GBytes  9.32 Gbits/sec    0   1.81 MBytes
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.00  sec  11.8 GBytes  10.1 Gbits/sec    0   sender
[  5]   0.00-10.00  sec  11.8 GBytes  10.1 Gbits/sec        receiver
```

This patch also removes the kretprobe after the cgroup/skb program runs,
since it is not needed anymore.

### 2. Explain how to test it

Terminal 1, execute tracee:

```
$ sudo ./dist/tracee --pprof --metrics --log debug --install-path /tmp/tracee --output option:parse-arguments --capabilities bypass=false --output none --trace comm!=node --trace event=net_packet_dns,net_packet_dns_request,net_packet_dns_response,hooked_proc_fops,hooked_syscalls,init_module,magic_write,mem_prot_alert,process_vm_writev,ptrace,sched_process_exec,security_bprm_check,security_file_open,security_inode_rename,security_kernel_read_file,security_sb_mount,security_socket_connect,socket_dup
```

Terminal 2, 3, 4 and 5, execute a dig loop:

```
$ while true; do dig rafaeldtinoco.com; done
```

> make sure resolutions are made locally to systemd-resolved (to its fast).

Terminal 6 and 7, execute a "ps -ef" loop:

```
$ while true; do ps -ef; done
```

Let the workload run for some time and observe tracee output some errors like:

```
    {"L":"ERROR","T":"2023-02-10T16:31:57.883-0300","M":"tracee
        encountered an error","error":"failed to derive event 2006:
        resource record length exceeds data"}
```

### 3. Other comments

Previous (and broken) PR: https://github.com/aquasecurity/tracee/pull/2710


pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3956,12 +3976,22 @@ typedef struct net_event_context {
...
+    u16 ip_csum;
+    struct in6_addr ip_saddr;
+    struct in6_addr ip_daddr;
+} indexer_t;
+
+typedef struct {
     __uint(type, BPF_MAP_TYPE_LRU_HASH);
-    __uint(max_entries, 1024);          // simultaneous cgroup/skb ingress/eggress progs
-    __type(key, u64);                   // sk_buff timestamp
+    __uint(max_entries, 10240);         // simultaneous cgroup/skb ingress/eggress progs
```
rafaeldtinoco, line 3988: I'll measure this before merging. Need to make sure number here is good for production environments with tons of sockets and on going flows/connections.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3974,7 +4004,7 @@ typedef struct net_task_context {
 
 struct {
     __uint(type, BPF_MAP_TYPE_LRU_HASH);
-    __uint(max_entries, 10240);             // simultaneous sockets being traced
+    __uint(max_entries, 65535);             // simultaneous sockets being traced
```
rafaeldtinoco, line 4007: ditto

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3987,7 +4017,7 @@ typedef struct entry {
 
 struct {
     __uint(type, BPF_MAP_TYPE_LRU_HASH);
-    __uint(max_entries, 1024);       // simultaneous tasks being traced for entry/exit
+    __uint(max_entries, 10240);      // simultaneous tasks being traced for entry/exit
```
rafaeldtinoco, line 4020: ditoo

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3996,7 +4026,7 @@ struct {
 
 struct {
     __uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
-    __uint(max_entries, 1024);
+    __uint(max_entries, 10240);
```
rafaeldtinoco, line 4029: ditto

rafaeldtinoco:
The code isn't very different from previous code, despite the patch being longer. Idea is that I'm parsing L3 header within the kprobe before the cgroup/skb is called, so I can generate the key that will be used inside the cgroup/skb program.

With this code I was able to track all on going connections and sockets in multiple tested scenarios (but will test it again, with smaller maps and more performance in mind, before merging).

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4106,8 +4229,8 @@ int BPF_KPROBE(trace_sock_alloc_file)
     entry.args[0] = PT_REGS_PARM1(ctx); // struct socket *sock
     struct socket *sock = (void *) PT_REGS_PARM1(ctx);
 
-    if (!is_socket_supported(sock))
-        return 0;
+    // if (!is_socket_supported(sock))
```
rafaeldtinoco, line 4232: I'll remove this before merging.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4295,202 +4416,122 @@ int BPF_KPROBE(cgroup_bpf_run_filter_skb)
...
-CGROUP_SKB_HANDLE_FUNCTION(proto_udp_dns);
-CGROUP_SKB_HANDLE_FUNCTION(proto_icmp);
-CGROUP_SKB_HANDLE_FUNCTION(proto_icmpv6);
+           if (nethdrs->iphdrs.iphdr.ihl > 5) { // re-read IP header if needed
+                l3_size -= get_type_size(struct iphdr);
+                l3_size += nethdrs->iphdrs.iphdr.ihl * 4;
+                bpf_core_read(nethdrs, l3_size, data_ptr);
+            }
 
-#define CGROUP_SKB_HANDLE(name) cgroup_skb_handle_##name(ctx, neteventctx, nethdrs);
```
rafaeldtinoco, line 4412: This is just a positional change. I'll eventually spinoff the network code into its own file anyway.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4295,202 +4416,122 @@ int BPF_KPROBE(cgroup_bpf_run_filter_skb)
...
+                    return 1; // ignore other protocols
+            }
 
-//
-// Network submission functions
-//
+            // add IPv4 header items to indexer
+            indexer.ip_csum = nethdrs->iphdrs.iphdr.check;
+            indexer.ip_saddr.in6_u.u6_addr32[0] = nethdrs->iphdrs.iphdr.saddr;
+            indexer.ip_daddr.in6_u.u6_addr32[0]  = nethdrs->iphdrs.iphdr.daddr;
```
rafaeldtinoco, line 4490: nit: two spaces here.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4295,202 +4416,122 @@ int BPF_KPROBE(cgroup_bpf_run_filter_skb)
...
-    return cgroup_skb_submit(&net_cap_events,
-                             ctx,
-                             neteventctx,
-                             event_type,
-                             nc->capture_length);
-}
+    neteventctx.bytes = 0; // event arg size: no payload by default (changed inside skb prog)
 
-// capture packet a single time (if passing through multiple protocols being submitted to userland)
+    // TODO: log collisions
```
rafaeldtinoco, line 4524: I'll have to implement logs for network errors. If I had that placed before, I would have understood much sooner what was going on with the code (map key conflicts and missing logic).

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4518,70 +4555,8 @@ static __always_inline u32 cgroup_skb_generic(struct __sk_buff *ctx)
...
-        has_prefix("POST ", http_min_str, 6)   ||
-        has_prefix("PUT ", http_min_str, 5)    ||
-        has_prefix("DELETE ", http_min_str, 8) ||
-        has_prefix("HEAD ", http_min_str, 6)) {
-        return proto_http_req;
-    }
-    // clang-format on
-
-    return 0;
-}
```
rafaeldtinoco, line 4576: Positional change.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4518,70 +4555,8 @@ static __always_inline u32 cgroup_skb_generic(struct __sk_buff *ctx)
...
-
-//
-// SUPPORTED SOCKET FAMILY TYPES (inet, inet6)
-//
-
-CGROUP_SKB_HANDLE_FUNCTION(family)
```
rafaeldtinoco, line 4582: The family handle function was added to the initial cgroup/skb function because we needed this logic to access the bpf map and pick the event context for the packet being processed.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4754,7 +4797,45 @@ CGROUP_SKB_HANDLE_FUNCTION(proto)
...
+    if (has_prefix("GET ", http_min_str, 5)    ||
+        has_prefix("POST ", http_min_str, 6)   ||
+        has_prefix("PUT ", http_min_str, 5)    ||
+        has_prefix("DELETE ", http_min_str, 8) ||
+        has_prefix("HEAD ", http_min_str, 6)) {
+        return proto_http_req;
+    }
+
+    return 0;
+}
```
rafaeldtinoco, line 4838: Positional change.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4085,7 +4113,102 @@ static __always_inline int should_capture_net_event(net_event_context_t *neteven
...
+
+// capture packet a single time (if passing through multiple protocols being submitted to userland)
+#define cgroup_skb_capture() {                                                                     \
+    if (should_submit_net_event(neteventctx, CAP_NET_PACKET)) {                                    \
+        if (neteventctx->md.captured == 0) { /* do not capture the same packet twice */            \
+            cgroup_skb_capture_event(ctx, neteventctx, NET_PACKET_CAP_BASE);                       \
+            neteventctx->md.captured = 1;                                                          \
+        }                                                                                          \
+    }                                                                                              \
+}
```
rafaeldtinoco, line 4208: Positional change.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3913,13 +3911,36 @@ static __always_inline bool is_socket_supported(struct socket *sock)
...
+    struct icmphdr icmphdr;
+    struct icmp6hdr icmp6hdr;
+    union {
+        u8 tcp_extra[40]; // data offset might set it up to 60 bytes
+    };
+} protohdrs;
+
+typedef struct nethdrs_t {
+    iphdrs iphdrs;
+    protohdrs protohdrs;
```
rafaeldtinoco, line 3942: Positional change.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4295,202 +4416,122 @@ int BPF_KPROBE(cgroup_bpf_run_filter_skb)
...
-//
-
-// NOTE: proto header structs need full type in vmlinux.h (for correct skb copy)
-
-typedef union iphdrs_t {
-    struct iphdr iphdr;
-    struct ipv6hdr ipv6hdr;
-} iphdrs;
+    void *data_ptr = NULL;
+    u16 mac_len = BPF_READ(skb, mac_len);
```
rafaeldtinoco, line 4449: split into 2 maps: ingress/egress netctx because the packet header would be the same for both (causing a conflict in the map). By having 2 cgrpctxmaps I can avoid that happening and still have the 2 different packets inside the map(s).

rafaeldtinoco (commented):
Left comments with explanations and some other nits/notes to address before merging (tomorrow).

pkg/ebpf/c/tracee.bpf.c
yanivagman: All struct definitions are now under types.h, so better move it there or to network.h

yanivagman (approved):
A minor comment about location of declared types,
LGTM

geyslan (changes requested):
Overall it LGTM. Just need relocate some code as @yanivagman mentioned.

rafaeldtinoco:
Opened https://github.com/aquasecurity/tracee/issues/2721 during some tests (to investigate the behavior in near future).

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: @rafaeldtinoco

PR #2717, opened 2023-02-14
tracee-rules: bypass capabilities if user not root

josedonizetti:

Fix https://github.com/aquasecurity/tracee/issues/2718

The [PR](https://github.com/aquasecurity/tracee/pull/2580) fixed a bug when `tracee-rules` was executed with `sudo`, though it introduced a bug for when `tracee-rules` is executed without `sudo`, it fails because it can't raise capabilities.

This PR fixes this bug. 

### 2. Explain how to test it

```
make
./dist/tracee-rules --list-events
sudo ./dist/tracee-rules --list-events
```

### DIFF
--- a/cmd/tracee-rules/main.go
+++ b/cmd/tracee-rules/main.go
@@ -51,7 +51,10 @@ func main() {
 
 			// Capabilities command line flags
 
-			err := capabilities.Initialize(c.Bool("allcaps"))
+			// if the user is not root, we will bypass capabilities raising,
+			// which means default capabilities will be used
+			bypass := c.Bool("allcaps") || !isRoot()
+			err := capabilities.Initialize(bypass)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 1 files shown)


cmd/tracee-rules/main.go
```diff
@@ -285,3 +288,7 @@ func listEvents(w io.Writer, sigs []detect.Signature) {
 	sort.Slice(events, func(i, j int) bool { return events[i] < events[j] })
 	fmt.Fprintln(w, strings.Join(events, ","))
 }
+
+func isRoot() bool {
+	return os.Getegid() == 0
```
geyslan, line 293:
```suggestion
	return os.Geteuid() == 0
```

geyslan (changes requested):

geyslan (approved):
LGTM

rafaeldtinoco (approved):
LGTM! Tku!

PR #2720, opened 2023-02-14
tracee: fix filters

josedonizetti:

Fixes TODO for https://github.com/aquasecurity/tracee/issues/2355

The unified binary create events for all signatures by default (add events to Definitions). The enabling of such events is handled by `tracee` default filtering logic. With this strategy we can support use cases like: 

--trace e=anti_debugging
--trace set=signatures
--trace set=signatures --trace e!=anti_debugging

### 2. Explain how to test it

```
tracee --trace e=anti_debugging
tracee --trace set=signatures
tracee --trace set=signatures --trace e!=anti_debugging
```

### DIFF
--- a/cmd/tracee/main.go
+++ b/cmd/tracee/main.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"os"
 	"os/signal"
-	"strings"
 	"syscall"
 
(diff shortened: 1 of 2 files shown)


rafaeldtinoco (approved):
LGTM

Issue #2589, opened 2023-01-16
filters equality mismatches


geyslan:
## Description

There are mismatches in the equality evaluation order for `uint` and `containers` filters

### _equal, then not equal_

**containers**
https://github.com/aquasecurity/tracee/blob/main/pkg/filters/containers.go#L34-L58

**uint**
https://github.com/aquasecurity/tracee/blob/main/pkg/filters/uint.go#L224-L245

when compared to string and binary

### _not equal, then equal_

**string**
https://github.com/aquasecurity/tracee/blob/main/pkg/filters/string.go#L248-L265

**binary**
https://github.com/aquasecurity/tracee/blob/main/pkg/filters/binary.go#L188-L200

## Additional details 

Equality - equal - should always be evaluated later by overriding the _not equal_ filters with the same key.

geyslan:
@NDStrahilevitz @yanivagman

PR #2726, opened 2023-02-15
swap uint and containers equality order

geyslan:
Fixes: #2589

    swap uint and containers equality order
    
    Make them have the same equality precedence as string and binary.


NDStrahilevitz (approved):
LGTM

PR #2729, opened 2023-02-15
flags: remove cache-events from output help

josedonizetti:

`tracee --output help` list an option for `cache-events` which the code ignores, this probably is an old flag that was replaced by higher level flag `tracee --cache help`

### 2. Explain how to test it

```
tracee --output help
```


geyslan (approved):
LGTM

Issue #2734, opened 2023-02-16
not finding containers_map entry from cgroup_mkdir event


geyslan:
## Description

Running proposed integration tests PR https://github.com/aquasecurity/tracee/pull/2688, I've faced these errors:

```
=== RUN   Test_EventFilters/pid:_event:_args:_trace_event_sched_switch_with_args_from_pid_0
    tracee.go:81: started tracee...
	executing: /usr/bin/kill [kill -SIGHUP 1]
    tracee.go:118: timed out on output
=== RUN   Test_EventFilters/pid:_trace_events_from_pid_1
{"level":"error","ts":1676561576.2194765,"msg":"error processing event","event":"cgroup_mkdir","error":"failed to get lookup key 824719350456 from map containers_map: no such file or directory"}
{"level":"error","ts":1676561576.2195385,"msg":"tracee encountered an error","error":"failed to get lookup key 824719350456 from map containers_map: no such file or directory"}
    tracee.go:81: started tracee...
	executing: /usr/bin/kill [kill -SIGHUP 1]
{"level":"error","ts":1676561578.1471684,"msg":"error processing event","event":"cgroup_mkdir","error":"failed to get lookup key 824722541736 from map containers_map: no such file or directory"}
{"level":"error","ts":1676561578.1472602,"msg":"tracee encountered an error","error":"failed to get lookup key 824722541736 from map containers_map: no such file or directory"}
{"level":"error","ts":1676561578.3838687,"msg":"error processing event","event":"cgroup_mkdir","error":"failed to get lookup key 824722969368 from map containers_map: no such file or directory"}
{"level":"error","ts":1676561578.3838956,"msg":"tracee encountered an error","error":"failed to get lookup key 824722969368 from map containers_map: no such file or directory"}
```

https://github.com/aquasecurity/tracee/actions/runs/4195470616/jobs/7275189121

yanivagman:
Ok, looking at the code - this means that a cgroup was created in the system, but it is not a container's cgroup (can be systemd cgroup or something like that). Then, we try to delete it from the containers_map which is a bpf map, but the entry doesn't exist so we fail here: https://github.com/aquasecurity/libbpfgo/blob/15ae9ff99817997b95110096eda30c4d3824c948/libbpfgo.go#L926

This can happen if a cgroup was created and removed very quickly and it is not part of a container.
Not sure we have something to do with it.

[...]

PR #2737, opened 2023-02-16
log when not a container cgroup instead of err

geyslan:

log when not a container cgroup instead of err

Fixes: #2734

- #2734

### DIFF
--- a/pkg/ebpf/events_processor.go
+++ b/pkg/ebpf/events_processor.go
@@ -270,7 +270,13 @@ func (t *Tracee) processCgroupMkdir(event *trace.Event) error {
 		// If cgroupId is from a regular cgroup directory, and not the
 		// container base directory (from known runtimes), it should be
 		// removed from the containers bpf map.
-		err = t.containers.RemoveFromBpfMap(t.bpfModule, cgroupId, hId)
+		if err := t.containers.RemoveFromBpfMap(t.bpfModule, cgroupId, hId); err != nil {
+			// If the cgroupId was not found in bpf map, this could mean that
+			// it is not a container cgroup and, as a systemd cgroup, could have been
+			// created and removed very quickly.
+			// In this case, we don't want to return an error.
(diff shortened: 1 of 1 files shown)


pkg/ebpf/events_processor.go
```diff
@@ -271,6 +271,14 @@ func (t *Tracee) processCgroupMkdir(event *trace.Event) error {
 		// container base directory (from known runtimes), it should be
 		// removed from the containers bpf map.
 		err = t.containers.RemoveFromBpfMap(t.bpfModule, cgroupId, hId)
+		if err != nil {
+			// If the cgroupId was not found in bpf map, this could mean that
+			// it is not a container cgroup and, as a systemd cgroup, could have been
+			// created and removed very quickly.
+			// In this case, we don't want to return an error.
+			logger.Debug("error removing map entry, certainly cgroup is not from a container", "error", err)
+			err = nil
```
yanivagman, line 280: You can just use `err := ...` instead of `err = ...` above to avoid this

pkg/ebpf/events_processor.go
```diff
@@ -271,6 +271,14 @@ func (t *Tracee) processCgroupMkdir(event *trace.Event) error {
 		// container base directory (from known runtimes), it should be
 		// removed from the containers bpf map.
 		err = t.containers.RemoveFromBpfMap(t.bpfModule, cgroupId, hId)
+		if err != nil {
+			// If the cgroupId was not found in bpf map, this could mean that
+			// it is not a container cgroup and, as a systemd cgroup, could have been
+			// created and removed very quickly.
+			// In this case, we don't want to return an error.
+			logger.Debug("error removing map entry, certainly cgroup is not from a container", "error", err)
```
yanivagman, line 279:
```suggestion
			logger.Debug("failed to remove entry from containers bpf map", "error", err)
```

yanivagman (approved):
LGTM

Issue #2665, opened 2023-02-01
Tracee Policies


yanivagman:
- [x] Implementing policies in Tracee, as discussed in #2640
    - [x] https://github.com/aquasecurity/tracee/pull/2892
    - [x] https://github.com/aquasecurity/tracee/pull/2922
    - [x] https://github.com/aquasecurity/tracee/pull/2923
    - [x] https://github.com/aquasecurity/tracee/pull/2845
    - [x] https://github.com/aquasecurity/tracee/pull/2927 
    - [ ] work on docs

Since policies are based on multi-scopes, a few extra changes to the UX and internal meaning are required:

- [x] https://github.com/aquasecurity/tracee/issues/2751 
- [x] https://github.com/aquasecurity/tracee/pull/2919 ~Multiscopes~ policies numbers feature won't be available through the CLI anymore (adding a number to filters will be removed), and it will become the infrastructure for policies. A user that wants to use it, will need to define a policy. 
- [x] remove policies table column as discussed in https://github.com/aquasecurity/tracee/pull/2923#issuecomment-1485965210

geyslan:
I tested this on an ongoing PR and it worked great.

![image](https://user-images.githubusercontent.com/3372117/226683347-a0618a34-f66f-43b2-98ec-405cb089b0fb.png)

PR #2738, opened 2023-02-16
filterscopes: create a filterscopes pkg

rafaeldtinoco:

filterscopes: create a filterscopes pkg

A filterscopes pkg is needed so that the it can be used outside of the ebpf pkg. This is needed for the upcoming refactors such as the one that will allow the filterscopes to be used by the "load objects" and "symbols collision" events.

### 3. Other comments

@geyslan please check if this brakes your current work too much, and, if you can accept it and rebase your code, I'll be able to continue a refactor proposal to "ld SO symbols collision" (https://github.com/aquasecurity/tracee/pull/2053/) PR so the derivation table initialization does not get specific event initialization at its top.

### DIFF
--- a/pkg/cmd/flags/filter.go
+++ b/pkg/cmd/flags/filter.go
@@ -5,9 +5,9 @@ import (
 	"strconv"
 	"strings"
 
-	tracee "github.com/aquasecurity/tracee/pkg/ebpf"
 	"github.com/aquasecurity/tracee/pkg/events"
 	"github.com/aquasecurity/tracee/pkg/filters"
+	"github.com/aquasecurity/tracee/pkg/filterscope"
 	"github.com/aquasecurity/tracee/pkg/logger"
 )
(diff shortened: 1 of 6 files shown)


yanivagman:
filterscopes name may be confusing. Perhaps policies package?

rafaeldtinoco:
> filterscopes name may be confusing. Perhaps policies package?

I agreed with @geyslan that he will be renaming in his PR (so this is a refactor only).

geyslan:
> > filterscopes name may be confusing. Perhaps policies package?
> 
> I agreed with @geyslan that he will be renaming in his PR (so this is a refactor only), because he is already renaming all terms (as he agreed with you, from what he said).

I'll tackle #2665 renaming after this, as Rafael needs the scopes decoupled.

geyslan (approved):
LGTM

Issue #2231, opened 2022-10-12
Support multiple outputs for tracee-ebpf


AsafEitani:
## Prerequisites

- [x] This issue is an EPIC issue (add label: EPIC).
- [ ] This issue is an EPIC TASK (add issue to EPIC description).

Select one OR another:

- [ ] I'll create a PR to implement this feature (assign to yourself).
- [x] Someone else should implement this (describe it well).

## Feature description

This feature is about adding support for multiple outputs of tracee-ebpf.
An example use case is if someone wants to display the table output on screen but save the json output to disk for later investigation.
Another use case is outputting both json and gob formats for signature testing purposes.

patrick-stephens:
I've submitted a PR to integrate with Fluent Bit which potentially also resolves this: #2155 
You can use Fluent Bit then to send to one/many differing endpoints including a file, S3, etc. without any change to Tracee required: https://docs.fluentbit.io/manual/pipeline/outputs

PR #2746, opened 2023-02-20
Add multiple printers

josedonizetti:

This PR add support for multiple printers to new `tracee` binary. The old binary `tracee-ebpf` maintain the previous syntax, and the change is only implemented for the new binary, for such I have first created a commit where I rename the previous `output` flag in order to simplify the deletion once we deprecate and remove `tracee-ebpf`. 

The new syntax was discussed on a previous PR https://github.com/aquasecurity/tracee/pull/2709
This PR is part of https://github.com/aquasecurity/tracee/issues/2355
Close https://github.com/aquasecurity/tracee/issues/2231

### 2. Explain how to test it

Please check the help commands:
```
tracee-ebpf --output help
tracee --output help
```

### 3. Other comments

Multiple printers will be the base of the webhook implementation for the new binary, where webhooks will be a printer like stdout, etc.

### DIFF
--- a/pkg/cmd/flags/output.go
+++ b/pkg/cmd/flags/output.go
@@ -1,0 +1,279 @@
+package flags
+
+import (
+	"errors"
+	"fmt"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/aquasecurity/tracee/pkg/cmd/printer"
+	tracee "github.com/aquasecurity/tracee/pkg/ebpf"
+	"github.com/aquasecurity/tracee/pkg/logger"
+)
+
+func outputHelp() string {
+	return `Control how and where output is printed.
+Format options:
+table[:/path/to/file,...]                          output events in table format (default). The default path to file is stdout.
+table-verbose[:/path/to/file,...]                  output events in table format with extra fields per event. The default path to file is stdout.
+json[:/path/to/file,...]                           output events in json format. The default path to file is stdout.
+gob[:/path/to/file,...]                            output events in gob format. The default path to file is stdout.
+gotemplate=/path/to/template[:/path/to/file,...]   output events formatted using a given gotemplate file. The default path to file is stdout.
+none                                               ignore stream of events output, usually used with --capture
+
+Fluent Forward options:
+forward:url                                        send events in json format using the Forward protocol to a Fluent receiver
+
+Log options:
+log-file:/path/to/file                             write the logs to a specified file. create/trim the file if exists (default: stderr)
+
+Other options:
+option:{stack-addresses,exec-env,relative-time,exec-hash,parse-arguments,sort-events}
+                                                   augment output according to given options (default: none)
+  stack-addresses                                  include stack memory addresses for each event
+  exec-env                                         when tracing execve/execveat, show the environment variables that were used for execution
+  relative-time                                    use relative timestamp instead of wall timestamp for events
+  exec-hash                                        when tracing sched_process_exec, show the file hash(sha256) and ctime
+  parse-arguments                                  do not show raw machine-readable values for event arguments, instead parse into human readable strings
+  parse-arguments-fds                              enable parse-arguments and enrich fd with its file path translation. This can cause pipeline slowdowns.
+  sort-events                                      enable sorting events before passing to them output. This will decrease the overall program efficiency.
+
+Examples:
+  --output json                                                  | output as json to stdout
+  --output json:/my/out                                          | output as json to /my/out
+  --output gotemplate=/path/to/my.tmpl                           | output as the provided go template to stdout
+  --output gob:/my/out --output log-file:/my/log                 | output gob to /my/out and logs to /my/log
+  --output json --output gob:/my/out                             | output as json to stdout and as gob to /my/out
+  --output json:/my/out1,/my/out2                                | output as json to both /my/out and /my/out2
+  --output none                                                  | ignore events output
+  --output table --output option:stack-addresses                 | output as table with stack addresses
+  --output forward:tcp://user:pass@127.0.0.1:24224?tag=tracee    | output via the Forward protocol to 127.0.0.1 on port 24224 with the tag 'tracee' using TCP
+
+Use this flag multiple times to choose multiple output options
+`
+}
+
+type OutputConfig struct {
+	TraceeConfig   *tracee.OutputConfig
+	PrinterConfigs []printer.Config
+	LogFile        *os.File
+}
+
+func PrepareOutput(outputSlice []string) (OutputConfig, error) {
+	outConfig := OutputConfig{}
+	traceeConfig := &tracee.OutputConfig{}
+
+	var logPath string
+
+	//outpath:format
+	printerMap := make(map[string]string)
+
+	for _, o := range outputSlice {
+		outputParts := strings.SplitN(o, ":", 2)
+
+		if strings.HasPrefix(outputParts[0], "gotemplate=") {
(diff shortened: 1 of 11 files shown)


pkg/cmd/flags/output.go
```diff
@@ -30,92 +34,82 @@ option:{stack-addresses,exec-env,relative-time,exec-hash,parse-arguments,sort-ev
...
   sort-events                                      enable sorting events before passing to them output. This will decrease the overall program efficiency.
+
 Examples:
   --output json                                            | output as json to stdout
-  --output gotemplate=/path/to/my.tmpl                     | output as the provided go template
-  --output out-file:/my/out --output log-file:/my/log      | output to /my/out and logs to /my/log
+  --output json:/my/out                                    | output as json to /my/out
+  --output gotemplate=/path/to/my.tmpl                     | output as the provided go template to stdout
+  --output gob:/my/out --output log-file:/my/log      	   | output gob to /my/out and logs to /my/log
+  --output json --output gob:/my/out                       | output as json to /my/out
```
geyslan, line 43: json and gob?

pkg/cmd/flags/output.go
```diff
@@ -30,92 +34,82 @@ option:{stack-addresses,exec-env,relative-time,exec-hash,parse-arguments,sort-ev
...
+  --output json:/my/out1,/my/out2                          | output as json to both /my/out and /my/out2
```
geyslan, line 44: Tested as:

`--output json:./out1,./out2`
`--output json:stdout,./output`
`--output json:./output --output json`

And one of them doesn't outputs; it's outputting to one end only.

pkg/cmd/urfave/urfave.go
```diff
@@ -101,9 +111,17 @@ func GetTraceeRunner(c *cli.Context, version string) (cmd.Runner, error) {
 
 	// Container information printer flag
 
-	printerConfig := output.PrinterConfig
-	printerConfig.ContainerMode = cmd.GetContainerMode(cfg)
-	cfg.Output = output.TraceeConfig
+	printers := make([]printer.EventPrinter, 0, len(output.PrinterConfigs))
+	for _, pConfig := range output.PrinterConfigs {
+		pConfig.ContainerMode = getContainerMode(cfg)
```
geyslan, line 116: I think you can use here the `(fs *FilterScopes) ContainerFilterEnabled()` helper.

geyslan (commented):
I put some comments to be addressed.

pkg/cmd/flags/output.go
josedonizetti: @geyslan I've tested the samples above and they all worked. 🤔

pkg/cmd/urfave/urfave.go
josedonizetti: @geyslan The `getContainerMode`is using `ContainerFilterEnabled()`helper, right?

pkg/cmd/flags/output.go
```diff
@@ -13,16 +14,19 @@ import (
...
+Format options:
+table[:/path/to/file,...]                          output events in table format (default). The default path to file is stdout.
+table-verbose[:/path/to/file,...]                  output events in table format with extra fields per event. The default path to file is stdout.
+json[:/path/to/file,...]                           output events in json format. The default path to file is stdout.
+gob[:/path/to/file,...]                            output events in gob format. The default path to file is stdout.
+gotemplate=/path/to/template[:/path/to/file,...]   output events formatted using a given gotemplate file. The default path to file is stdout.
 none                                               ignore stream of events output, usually used with --capture
+
+Log options:
+log-file:/path/to/file                             write the logs to a specified file. create/trim the file if exists (default: stderr)
```
yanivagman, line 26: log options should probably move to the `--log` flag

pkg/cmd/flags/output.go
```diff
@@ -13,16 +14,19 @@ import (
...
+forward:url                                        send events in json format using the Forward protocol to a Fluent receiver
```
yanivagman, line 27: is this a log option (since it is located here) or an output printer?

josedonizetti: @yanivagman output printer.

pkg/cmd/flags/output.go
josedonizetti: I like it, will create an issue for it.

pkg/cmd/urfave/urfave.go
geyslan: Actually you're using the FilterScope helper, I've mentioned the FilterScope**S** one, which already provides the needed information based on a bitmask filled through (fs *FilterScopes) Compute().

pkg/cmd/printer/broadcast.go
```diff
@@ -0,0 +1,56 @@
...
+func (b *Broadcast) Print(event trace.Event) {
+	for _, c := range b.eventsChan {
+		// we use select to avoid blocking the print loop if one of the printers is slow
+		// in case the event can't be sent because the buffered channel is full, we drop it
+		// and log an error
+		select {
+		case c <- event:
+		default:
+			logger.Error("dropping event due to slow printer", "event", event)
+		}
```
geyslan, line 40: I got more error logs than events, running:

sudo ./dist/tracee -t e='*' --output table:stdout

This logger Error polutes table output, WDYT of changing it to debug level and to increment errors channel?

geyslan (changes requested):

pkg/cmd/printer/broadcast.go
josedonizetti: @geyslan Done. Though if you test with "e='*'" we will lose events just like we do on the main branch, and possibly the  files generated with two prints will have different event on it.

geyslan (approved):
LGTM. Great job with the printer broadcast.

pkg/cmd/printer/broadcast.go
```diff
@@ -0,0 +1,58 @@
...
+func (b *Broadcast) Print(event trace.Event) {
+	for _, c := range b.eventsChan {
+		// we use select to avoid blocking the print loop if one of the printers is slow
+		// in case the event can't be sent because the buffered channel is full, we drop it
+		// and log an warning
+		select {
+		case c <- event:
+		default:
+			// TODO: add metrics about not printed events
+			logger.Debug("dropping event due to slow printer", "event", event)
```
yanivagman, line 41: I think this deserves at least info level log message

pkg/cmd/urfave/urfave.go
yanivagman, line 116: `getContainerMode(cfg)` can be called once outside of the loop

yanivagman (approved):
Nice work @josedonizetti !
Overall LGTM, some nit comments ahead

pkg/cmd/printer/broadcast.go
geyslan: Info is default level, so depending on the volume of events it will output more logs than events.

yanivagman: I think this should happen too frequently (at least I hope so), and if something is dropped here, there should be a way for the user to be aware of that. So as long as we don't have metrics for events dropped here, we should expose this information somehow

rafaeldtinoco:
Make sure to rebase so https://github.com/aquasecurity/tracee/pull/2776 is included and new testers are used.

rafaeldtinoco:
Thanks for waiting on my review/read. Nice job!

rafaeldtinoco (commented):
LGTM

PR #2744, opened 2023-02-19
tracee: add signature metadata to events

josedonizetti:
### 1. Explain what the PR does

This PR adds support to Metadata to `trace.Event`. At first this will be used to add `SignatureMetadata` when a signature event is created from a detection. Instead of adding `SignatureMetadata` directly, we have a more generic approach allowing for future use by events that are not signatures (eg: derivations, ebpf)

The `Metadata` field is marked as `omitempty` so we only print it if there is metadata, below an example of a signaure event, and a ebpf event. 

signature event, example:
```
{"timestamp":1676897645064612171,"threadStartTime":1676897645064527338,"processorId":5,"processId":1120432,"cgroupId":695068,"threadId":1120432,"parentProcessId":1120430,"hostProcessId":1120432,"hostThreadId":1120432,"hostParentProcessId":1120430,"userId":1000,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"strace","hostName":"josedonizetti-x","containerId":"","containerImage":"","containerName":"","podName":"","podNamespace":"","podUID":"","podSandbox":false,"eventId":"6018","eventName":"Anti-Debugging detected","matchedScopes":1,"argsNum":4,"returnValue":0,"syscall":"","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"args":[{"name":"request","type":"string","value":"PTRACE_TRACEME"},{"name":"pid","type":"pid_t","value":0},{"name":"addr","type":"void*","value":"0x0"},{"name":"data","type":"void*","value":"0x0"}],"metadata":{"SignatureDescription":"A process used anti-debugging techniques to block a debugger. Malware use anti-debugging to stay invisible and inhibit analysis of their behavior.","SignatureID":"TRC-102","SignatureName":"Anti-Debugging detected","SignatureProperties":{"Category":"defense-evasion","Kubernetes_Technique":"","Severity":1,"Technique":"Debugger Evasion","external_id":"T1622","id":"attack-pattern--e4dc8c01-417f-458d-9ee0-bb0617c1b391"},"SignatureTags":[],"SignatureVersion":"1"}}
```

ebpf event, example:
```
[...]

PR #2752, opened 2023-02-20
types: add event metadata

josedonizetti:

This PR adds support to Metadata to `trace.Event`. At first this will be used to add `SignatureMetadata` when a signature event is created from a detection. Instead of adding `SignatureMetadata` directly, we have a more generic approach allowing for future use by events that are not signatures (eg: derivations, ebpf)

The `Metadata` field is a pointer and marked as `omitempty` so we only print it if there are metadata. 

This PR is part of https://github.com/aquasecurity/tracee/issues/2355

### 2. Explain how to test it

Test the PR using this change to add signature metadata to signature events https://github.com/aquasecurity/tracee/pull/2753

### 3. Other comments

The [first implementation](https://github.com/aquasecurity/tracee/pull/2744) was using a more dynamic approach, but after discussion we decide to have a structure for the specific fields  `Version`, `Description`, `Tags`, `Properties`. This is based on `SignatureMetadata`, but different because we don't need name, nor id, as events already have name and id of themselves.

### DIFF
--- a/types/trace/trace.go
+++ b/types/trace/trace.go
@@ -44,6 +44,15 @@ type Event struct {
 	StackAddresses      []uint64     `json:"stackAddresses"`
 	ContextFlags        ContextFlags `json:"contextFlags"`
 	Args                []Argument   `json:"args"` //Arguments are ordered according their appearance in the original event
+	Metadata            *Metadata    `json:"metadata,omitempty"`
+}
+
+// Metadata is a struct that holds metadata about an event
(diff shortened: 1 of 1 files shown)


rafaeldtinoco (approved):
LGTM. I'll let you merge @josedonizetti because I want to make sure this type is like you want for https://github.com/aquasecurity/tracee/pull/2753. Either way we can change the type later if needed. Cheers!

Issue #2397, opened 2022-11-28
Tracee should drop capabilities by default and all signatures should work


rafaeldtinoco:
Tracee should be able to drop capabilities and have all its signatures fully working.

> By the time the feature was added, the E2E tests still failed because of some signatures not having proper capabilities in the pipeline event processing phases.

rafaeldtinoco:
We agreed to move this one to the next milestone.

rafaeldtinoco:
I closed https://github.com/aquasecurity/tracee/issues/2637#issue-1560438400 as a duplicate of this one (but it is actually one of the possible errors).

NDStrahilevitz:
Should we drop the bypass in integration tests as well?

PR #2759, opened 2023-02-22
capabilities: drop capabilities by default

rafaeldtinoco:
Fixes: #2397

This PR changes the default behavior from NOT dropping capabilities by default to dropping. 

### 2. Explain how to test it

Make sure you have `/proc/sys/kernel/perf_event_paranoid` set to 0.

Then run tracee with multiple events and capture options and, at the end, chose to add or drop capabilities at will:

```
sudo rm -rf /tmp/tracee/out/pcap && sudo tracee-ebpf --pprof --metrics --log debug --install-path /tmp/tracee --output option:parse-arguments --capabilities bypass=false --output format:json --trace event=net_packet_dns_request --capture network --capture pcap:single,process,command,container --capture pcap-options:filtered --capture pcap-snaplen:max --capabilities bypass=false --capabilities add=cap_lease --capabilities drop=cap_sys_admin
```

Note: If you drop cap_bpf, for example, tracee won't be able to work (as it tries to use cap_bpf by default in newer kernels, instead of cap_sys_admin). Now, if you drop cap_bpf but add cap_sys_admin, then it will work <- This might be a good test to play with.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -100,4 +100,6 @@ require (
 	kernel.org/pub/linux/libs/security/libcap/psx v1.2.64 // indirect
 )
 
+replace github.com/aquasecurity/tracee/types => ./types
+
 replace github.com/kubernetes/cri-api => k8s.io/cri-api v0.23.5-rc.0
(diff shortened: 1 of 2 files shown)


rafaeldtinoco:
Closing this for now.

PR #2760, opened 2023-02-22
Enrich image digest

### LINKED ISSUE #2691: Tracee to print unique image digest id for container events
Currently when running container commands like:
`./tracee-ebpf -t e=container_create -o option:exec-hash -o option:parse-arguments -o json --containers`

The output is:
`{"timestamp":1675856825493291836,"threadStartTime":254582557,"processorId":0,"processId":1,"cgroupId":22,"threadId":1,"parentProcessId":0,"hostProcessId":1,"hostThreadId":1,"hostParentProcessId":0,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"systemd","hostName":"tomer-ThinkPad","containerId":"","containerImage":"","containerName":"","podName":"","podNamespace":"","podUID":"","eventId":"2001","eventName":"container_create","argsNum":8,"returnValue":0,"stackAddresses":[0],"contextFlags":{"containerStarted":false},"args":[{"name":"runtime","type":"const char*","value":"docker"},{"name":"container_id","type":"const char*","value":"bc09696d0d49576575c5b13bb62dc3afb66282865a8732fc61d124e71894e283"},{"name":"ctime","type":"unsigned long","value":1675856825494541529},{"name":"container_image","type":"const char*","value":"alpine:3.17.1"},{"name":"container_name","type":"const char*","value":"cranky_jang"},{"name":"pod_name","type":"const char*","value":""},{"name":"pod_namespace","type":"const char*","value":""},{"name":"pod_uid","type":"const char*","value":""}]}`

Please add the container image digest id to the printed output.
reference:
https://docs.docker.com/engine/reference/commandline/pull/#pull-an-image-by-digest-immutable-identifier
image

NDStrahilevitz:

1. Separate the container enrich data in `trace.go` to it's own struct
2. Add an image digest field to the new data struct
3. Enrich the new field in pipeline
4. (Opportunistically) add new fields to context filter + fix bug (pod namespace filtered container image due to typo)

### 2. Explain how to test it

`tracee-ebpf --containers -t e=container_create,existing_container,sched_process_exec --json`
`docker run --rm redis`
Confirm any running container, new redis container and exec events include all enrich fields

### 3. Other comments

Fixes #2691


pkg/containers/runtime/crio.go
```diff
@@ -55,6 +55,7 @@ func (e *crioEnricher) Get(containerId string, ctx context.Context) (ContainerMe
 	}
 	metadata.Name = resp.Status.Metadata.Name
 	metadata.Image = resp.Status.Image.Image
+	metadata.ImageDigest = strings.TrimPrefix(resp.Status.ImageRef, "sha256:")
```
danielciuraru86, line 58: why do you need to trim the prefix? its part of the digest

NDStrahilevitz: Fixed.

rafaeldtinoco:

pkg/ebpf/events_pipeline.go
```diff
@@ -233,12 +241,7 @@ func (t *Tracee) decodeEvents(outerCtx context.Context, sourceChan chan []byte)
 				HostName:            string(bytes.TrimRight(ctx.UtsName[:], "\x00")),
 				CgroupID:            uint(ctx.CgroupID),
 				ContainerID:         containerInfo.ContainerId,
```
yanivagman, line 243: Shouldn't we put ContainerId under ContainerData?

NDStrahilevitz: It definitely could fit.
But we probably have a lot of users who would need to migrate the field, and it's a good indication for wether the fields inside `ContainerData` will be set.
`ContainerData` is only filled up with enrichment so it's easier to switch around (less existing consumers).

yanivagman: If you are already changing this, better to do it once in accordance with #2870 and put it under event.context.container

yanivagman: Yes.
Anyway we are going to break it, so better break it once

rafaeldtinoco:
I want to read this one.

pkg/ebpf/events_pipeline.go
NDStrahilevitz: Done

pkg/ebpf/events_enrich.go
```diff
@@ -78,7 +78,7 @@ func (t *Tracee) enrichContainerEvents(ctx gocontext.Context, in <-chan *trace.E
 			case event := <-in:
 				eventID := events.ID(event.EventID)
 				// send out irrelevant events (non container or already enriched), don't skip the cgroup lifecycle events
-				if (event.ContainerID == "" || event.ContainerImage != "") && eventID != events.CgroupMkdir && eventID != events.CgroupRmdir {
+				if (event.Container.ID == "" || event.Container.Name != "") && eventID != events.CgroupMkdir && eventID != events.CgroupRmdir {
```
yanivagman, line 81: `ContainerImage` -> `Container.Name`
is that correct or should have it been `Container.ImageName`?

yanivagman (commented):
Since this is a breaking change, we need to make sure to add a note in the release notes

rafaeldtinoco (approved):
LGTM, there is a doubt from Yaniv about `events_enrich` check, other than that I'm good.

pkg/ebpf/events_enrich.go
NDStrahilevitz: @rafaeldtinoco got it right.
As for the name changes:
`ContainerName` -> `Container.Name`
`ContainerImage` -> `Container.ImageName`

yanivagman (approved):

PR #2749, opened 2023-02-20
feat: add do_truncate event

roikol:
issue: #2748

### 1. Explain what the PR does

```
Author: RoiKol <roi.kol@aquasec.com>
Date:   Mon Feb 20 14:48:00 2023 +0200

    feat: add do_truncate event
    
    add event indicating a file truncate.
```

### 2. Explain how to test it

command 1: `./dist/tracee-ebpf -t e=do_truncate`
command 2: `echo test >> test.txt && truncate -s 1 test.txt`

PR #2769, opened 2023-02-26
Revert "feat: add do_truncate event (#2749)"

yanivagman:
Reverting latest commits to check 4.19 tests failure


rafaeldtinoco:
@yanivagman v4.19 is broken after all the runners got corrupted (due to a PR containing build-dependencies inclusion in the wrong jobs) and had to be fixed manually (last Friday).

I still haven't figured what is happening and why 4.19 worked before and now it is not working. Will have to figure it out.

rafaeldtinoco:
Related: https://github.com/aquasecurity/tracee/pull/2771

Issue #1602, opened 2022-03-28
tracee: non CO-RE execution in v4.19 kernels have libbpf errors/warnings


rafaeldtinoco:
Here is the output of a simple non CO-RE test:

```
$ ~/test-tracee.sh
 _   _  ___  _   _    ____ ___        ____  _____
| \ | |/ _ \| \ | |  / ___/ _ \      |  _ \| ____|
|  \| | | | |  \| | | |  | | | |_____| |_) |  _|
| |\  | |_| | |\  | | |__| |_| |_____|  _ <| |___
|_| \_|\___/|_| \_|  \____\___/      |_| \_\_____|

OSInfo: Security Lockdown is ''
OSInfo: ID: debian
OSInfo: KERNEL_RELEASE: 4.19.219
OSInfo: ARCH: x86_64
OSInfo: PRETTY_NAME: "Debian GNU/Linux bookworm/sid"
BTF: bpfenv = true, btfenv = false, vmlinux = false
BPF: using BPF object from environment: /tmp/tracee/tracee.bpf.4_19_219.v0_7_0-rc1-21-g9f3dd903.o
libbpf: Error in bpf_create_map_xattr(sys_enter_tails):ERROR: strerror_r(-524)=22(-524). Retrying without BTF.
libbpf: Error in bpf_create_map_xattr(sys_exit_tails):ERROR: strerror_r(-524)=22(-524). Retrying without BTF.
libbpf: Error in bpf_create_map_xattr(prog_array_tp):ERROR: strerror_r(-524)=22(-524). Retrying without BTF.
libbpf: Error in bpf_create_map_xattr(prog_array):ERROR: strerror_r(-524)=22(-524). Retrying without BTF.
Loaded 14 signature(s): [TRC-1 TRC-13 TRC-2 TRC-14 TRC-3 TRC-11 TRC-9 TRC-4 TRC-5 TRC-12 TRC-8 TRC-6 TRC-10 TRC-7]
libbpf: Error in bpf_create_map_xattr(sys_enter_tails):ERROR: strerror_r(-524)=22(-524). Retrying without BTF.
libbpf: Error in bpf_create_map_xattr(sys_exit_tails):ERROR: strerror_r(-524)=22(-524). Retrying without BTF.
libbpf: Error in bpf_create_map_xattr(prog_array_tp):ERROR: strerror_r(-524)=22(-524). Retrying without BTF.
[...]

Issue #2762, opened 2023-02-22
Integration Tests race condition


geyslan:
## Description

Running `go test -race` one can detect the conditions.

```shell
sudo make test-integration
GOOS=linux CC=clang GOARCH=amd64 CGO_CFLAGS="-I/home/gg/code/tracee/dist/libbpf" CGO_LDFLAGS="-lelf  -lz  /home/gg/code/tracee/dist/libbpf/libbpf.a" \
go test -race \
        -tags core,ebpf \
        -ldflags="-w \
                -extldflags \"\" \
                -X main.version=\"v0.9.2-330-g7e4ff9b1\" \
                " \
        -v \
        -p 1 \
        ./tests/integration/... \

=== RUN   TestInitNamespacesEvent
--- PASS: TestInitNamespacesEvent (0.00s)
=== RUN   Test_EventFilters
=== RUN   Test_EventFilters/do_a_file_write
    tracee.go:58: started tracee...
==================
WARNING: DATA RACE
Write at 0x00c00081a0e0 by goroutine 30:
  github.com/aquasecurity/tracee/pkg/ebpf.(*Tracee).Run()
      /home/gg/code/tracee/pkg/ebpf/tracee.go:1369 +0x6f3
  github.com/aquasecurity/tracee/tests/integration.startTracee.func2()
      /home/gg/code/tracee/tests/integration/tracee.go:59 +0x58

Previous read at 0x00c00081a0e0 by goroutine 27:
  github.com/aquasecurity/tracee/pkg/ebpf.(*Tracee).Running()
      /home/gg/code/tracee/pkg/ebpf/tracee.go:1454 +0x5a
  github.com/aquasecurity/tracee/tests/integration.waitforTraceeStart()
      /home/gg/code/tracee/tests/integration/tracee.go:93 +0x67
  github.com/aquasecurity/tracee/tests/integration.Test_EventFilters.func1()
      /home/gg/code/tracee/tests/integration/integration_test.go:394 +0x344
  testing.tRunner()
      /usr/lib/go/src/testing/testing.go:1576 +0x216
  testing.(*T).Run.func1()
      /usr/lib/go/src/testing/testing.go:1629 +0x47

Goroutine 30 (running) created at:
  github.com/aquasecurity/tracee/tests/integration.startTracee()
      /home/gg/code/tracee/tests/integration/tracee.go:59 +0x55d
  github.com/aquasecurity/tracee/tests/integration.Test_EventFilters.func1()
[...]

PR #2771, opened 2023-02-27
sets: default set can't have network events v419

rafaeldtinoco:
## Description

Commit 0601f099 added network events to the default set but, since the default set is used for all versions, this broke v419 runs, as that version does not support network events.

commit a8a0d377 (HEAD -> v419-fix, rafaeldtinoco/v419-fix)
Date:   Wed Mar 1 01:59:53 2023

    filter: mitigate v4.19 lack of network events
    
    The default event set includes some net_packet events, but they're
    not supported in vanilla v4.19 kernels. This patch mitigates this
    issue by adding all the network related events to an exclusion list
    during the filter events preparation phase.
    
    Fixes: #2771
    
    Note: This mitigation should be removed as soon as tracee has the
          feature probe mechanism in place.

commit e2d52ad5
Date:   Wed Mar 1 02:30:42 2023

    libbpfgo error callbacks: mitigate v4.19 map xattr
    
    Fixes: #1602

commit 03edd7be
Date:   Wed Mar 1 01:58:30 2023

    chore: events: keep network related events at the end

### DIFF
--- a/pkg/cmd/flags/filter.go
+++ b/pkg/cmd/flags/filter.go
@@ -5,6 +5,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/aquasecurity/libbpfgo/helpers"
 	"github.com/aquasecurity/tracee/pkg/events"
 	"github.com/aquasecurity/tracee/pkg/filters"
 	"github.com/aquasecurity/tracee/pkg/filterscope"
@@ -456,7 +457,11 @@ func PrepareFilterScopes(filtersArr []string) (*filterscope.FilterScopes, error)
 	return filterScopes, nil
 }
 
-func prepareEventsToTrace(eventFilter cliFilter, setFilter cliFilter, eventsNameToID map[string]events.ID) (map[events.ID]string, error) {
+func prepareEventsToTrace(
+	eventFilter cliFilter, setFilter cliFilter, eventsNameToID map[string]events.ID,
+) (
+	map[events.ID]string, error,
+) {
 	eventsToTrace := eventFilter.Equal
 	excludeEvents := eventFilter.NotEqual
 	setsToTrace := setFilter.Equal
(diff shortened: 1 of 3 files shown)


rafaeldtinoco:
The focal419 runner is running with clang13 and golang1.18 for these last tests. Will have to update it to clang14 and golang1.19 so this runner is the same as all others (and re-check to see if tests still pass). If tests pass, then we're good for v419 (for the release).

rafaeldtinoco:
Yes, so.. besides this fix, which tests didn't catch because they don't run the default set of events, there is another problem that raised when bumping go to 1.19 and clang to 14:

image

This wan't caught before because the runner, despite having go1.19, was still running tests using go1.18 (due to bad GOHOME env variable).

So, the summary is:

- this fix is needed for v4.19
- v4.19 only works with go1.18 and clang13
- v4.19 with go1.19 and clang14 has the errors of the latest test.

yanivagman:
> * commit [0601f09](https://github.com/aquasecurity/tracee/commit/0601f099b971402518eeba254b7103c6c17d3cfa) added network events to the default set but, since the default set is used for all versions, this broke v419 runs, as that version does not support network events.

This fix is a valid workaround, but for the long term we will need to identify if network events are supported on the environment where we run, and according to that choose which events can be enabled and which can't

yanivagman:
> ACTION: will have to understand why clang14 and go1.19 brake v4.19.

My guess is that clang is causing this, and not go1.19.
For now, can we use clang13 and go1.19?

rafaeldtinoco:
> > ACTION: will have to understand why clang14 and go1.19 brake v4.19.
> 
> My guess is that clang is causing this, and not go1.19. For now, can we use clang13 and go1.19?

@geyslan can we ? I remember you said you needed 1.19 for your latest changes.

geyslan:
> > > ACTION: will have to understand why clang14 and go1.19 brake v4.19.
> > 
> > 
> > My guess is that clang is causing this, and not go1.19. For now, can we use clang13 and go1.19?
> 
> @geyslan can we ? I remember you said you needed 1.19 for your latest changes.

For sure, `atomic.Bool` comes in 1.19. But if not possible I can re-implement it directly in tracee.

rafaeldtinoco:
> For sure, `atomic.Bool` comes in 1.19. But if not possible I can re-implement it directly in tracee.

Alright, I'll give it another hour to see if I can discover what is going on, then if not possible we can try getting a workaround from you (for the release) if you don't mind. Thank you!

geyslan:
> > For sure, `atomic.Bool` comes in 1.19. But if not possible I can re-implement it directly in tracee.
> 
> Alright, I'll give it another hour to see if I can discover what is going on, then if not possible we can try getting a workaround from you (for the release) if you don't mind. Thank you!

Just don't bother rushing this, since I've pulled #2762 from v0.13. It's not a priority.

rafaeldtinoco:
I have opened https://github.com/aquasecurity/tracee/issues/2772 for the clang14 + v4.19 issue.

Will try to get network events out of the default set when running in a v4.19 kernel (that is not RHEL or similar).

pkg/cmd/flags/filter.go
```diff
@@ -456,22 +457,53 @@ func PrepareFilterScopes(filtersArr []string) (*filterscope.FilterScopes, error)
...
 		}
 	}
+
+	// Exclude network events from the default set if kernel v4.19.
+	// Issue: https://github.com/aquasecurity/tracee/pull/2771
+	// TODO: workaround until we don't have fetures probe feature.
+	if osInfo, err := helpers.GetOSInfo(); err == nil {
+		kernel51ComparedToRunningKernel, err := osInfo.CompareOSBaseKernelRelease("5.1.0")
+		if err != nil {
+			logger.Error("failed to compare kernel version", "error", err)
```
geyslan, line 486: Only logging? What about to return error?

pkg/cmd/flags/filter.go
```diff
@@ -456,22 +457,53 @@ func PrepareFilterScopes(filtersArr []string) (*filterscope.FilterScopes, error)
...
+		}
+		if kernel51ComparedToRunningKernel == helpers.KernelVersionNewer {
```
geyslan, line 488: It is being compared even when `osInfo.CompareOSBaseKernelRelease()` may have returned an error.

geyslan (approved):
Overall LGTM.

pkg/cmd/flags/filter.go
```diff
@@ -456,22 +457,53 @@ func PrepareFilterScopes(filtersArr []string) (*filterscope.FilterScopes, error)
...
+
+	// build a map: k:set, v:eventID
 	for id, event := range events.Definitions.Events() {
 		for _, set := range event.Sets {
 			setsToEvents[set] = append(setsToEvents[set], id)
 		}
 	}
+
+	// Exclude network events from the default set if kernel v4.19.
+	// Issue: https://github.com/aquasecurity/tracee/pull/2771
```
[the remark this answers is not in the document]
rafaeldtinoco: Ooops

pkg/cmd/flags/filter.go
rafaeldtinoco: Yaniv prefers to continue in cases like this, to fail only in last case. It might fail later on because of existing network events in the default set, so there is no problem not to fail here.

pkg/cmd/flags/filter.go
rafaeldtinoco: true. fixed.

geyslan (approved):
LGTM

rafaeldtinoco:
@geyslan please merge this once you're okay with e2e tests (at least to consider another merge).

PR #2773, opened 2023-02-27
errors: improve error output

rafaeldtinoco:

errors: improve error output

It is very common to receive errors such as:

  "operation failed"

due to error being propagated to upper layers without context. If the error is related to a syscall, for example, it might be very generic and impossible to know its origins.

The logger package has solved that by putting context frames in both logger and debug messages, but it is not enough when tracee is being exited due to an error.

This patch:

- provides context for all existing errors.
- move error context functions to logger package.

By doing this change, it was possible to understand better all errors coming from dropping privileges, for example. Without it, it was very time consuming to understand what error was coming from which part of the code.

### DIFF
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -568,3 +568,28 @@ func GetLevel() Level {
 func HasDebugLevel() bool {
 	return pkgLogger.cfg.Level == DebugLevel
 }
+
+// ErrorFuncName displays a given error prefixed by the current (caller) fn name
+// and a given string in a given format (just like fmt.Sprintf)
+func NewErrorf(format string, a ...interface{}) error {
+	sentence := fmt.Sprintf(format, a...)
+
+	pc, _, _, _ := runtime.Caller(1)
+	funcName := runtime.FuncForPC(pc).Name()
+	index := strings.LastIndex(funcName, "/") + 1
+
+	return fmt.Errorf("%v: %v", funcName[index:], sentence)
+}
+
+// ErrorFunc displays a given error prefixed by the current (caller) fn name
+func ErrorFunc(e error) error {
+	if e != nil {
+		pc, _, _, _ := runtime.Caller(1)
+		funcName := runtime.FuncForPC(pc).Name()
+		index := strings.LastIndex(funcName, "/") + 1
+
+		return fmt.Errorf("%v: %v", funcName[index:], e)
+	}
+
+	return nil
+}
--- a/cmd/tracee-rules/input.go
+++ b/cmd/tracee-rules/input.go
@@ -4,7 +4,6 @@ import (
 	"bufio"
 	"encoding/gob"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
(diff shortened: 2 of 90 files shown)


pkg/logger/logger.go
```diff
@@ -568,3 +568,28 @@ func GetLevel() Level {
 func HasDebugLevel() bool {
 	return pkgLogger.cfg.Level == DebugLevel
 }
+
+// ErrorFuncName displays a given error prefixed by the current (caller) fn name
+// and a given string in a given format (just like fmt.Sprintf)
+func Errorf(format string, a ...interface{}) error {
```
geyslan, line 574: comment with different name

geyslan: I worry about name confusion with (proper) logger function like:

`func Error(msg string, keysAndValues ...interface{}) {`

In the future, we might want to provide, I don't know, `Errorf()` that logs using formats (verbs) instead of key/value pairs. Zap already provides these signatures, we're just ignoring them for now.

How about calling it `NewErrorf()`?

rafaeldtinoco: Sounds good to me. I'm fixing the testing as well.

pkg/logger/logger.go
```diff
@@ -568,3 +568,28 @@ func GetLevel() Level {
...
+	if e != nil {
+		pc, _, _, _ := runtime.Caller(1)
+		funcName := runtime.FuncForPC(pc).Name()
+		index := strings.LastIndex(funcName, "/") + 1
+
+		return fmt.Errorf("%v: %v", funcName[index:], e)
+	}
+
+	return nil
+}
```
geyslan, line 595: This is a great improvement for debugging without doubts. It's awesome, thanks @rafaeldtinoco.

I'm ok with this PR and already approving it, but I still have some concerns about the naming and where these helpers lie. Reading the returns with the enriched errors, I got a bit confused about the meaning and origin. It looks like we are returning a logger function to be managed/called from the caller.

What about to move them to `pkg/{err,er,e,te,whatever}/errors.go` and use them like:

`return e.NewErrorf("%s blah", str)`

what could make possible to return it to the previous naming:

`return e.Errorf("%s blah", str)`

I think that could make the purpose clear.

The final suggestion would change `ErrorFunc()` to something like `WrapError(err)`, `EnrichError(err)` or `FormatError(err)`.

Or as it would always be used directly from the error pkg, something as `e.Func(err)` would suffix as well.

WDYT?

geyslan (approved):
LGTM, put final comments. Thanks!

pkg/logger/logger.go
rafaeldtinoco: I want to have this released so we can better track issues (much better than what we have today). I'll open an issue to track what you're proposing because I think it is a good idea, but it would take too much time for this release I'm afraid.

rafaeldtinoco: https://github.com/aquasecurity/tracee/pull/2773#discussion_r1119379506

PR #2776, opened 2023-02-27
workflow: move runners to jenkins

rafaeldtinoco:
workflow: move runners to jenkins.


rafaeldtinoco:
Alright, so... Stream8 is still failing due to "an internal error". That might be because of something wrong in the AMI image or something else in Jenkins. I have already warned devops team about it but, for now, I'll merge the workflow (so we have testers) and fix this tomorrow.

PR #2779, opened 2023-02-28
signatures: load sigs as default events

josedonizetti:

Signatures are loaded by default. This is part of https://github.com/aquasecurity/tracee/issues/2355

### 2. Explain how to test it

Try triggering a simple signature like`anti_debugging`


geyslan (approved):
LGTM.

Tested:

image

PR #2785, opened 2023-03-01
k8s: remove flag everythingIsAnEvent from helm

josedonizetti:
This flag was used for users who wanted to try the new experience when it was beta, but the new experience will now be the default option.

This PR removes the helm flag `everythingIsAnEvent` which was used to enable the new experience while it was beta. With the next release this experience will be the default option. 

This is part of https://github.com/aquasecurity/tracee/issues/2355

### 2. Explain how to test it

```
# have a k8s cluster

helm repo add aqua https://aquasecurity.github.io/helm-charts/
helm dependency update ./deploy/helm/tracee
helm install tracee ./deploy/helm/tracee \
        --namespace tracee-system --create-namespace \
        --set hostPID=true

```

### DIFF
--- a/deploy/helm/tracee/templates/daemonset.yaml
+++ b/deploy/helm/tracee/templates/daemonset.yaml
@@ -30,14 +30,6 @@ spec:
         - name: tracee
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
-          {{- if .Values.everythingIsAnEvent }}
-          command: ["/tracee/tracee"]
-          args:
-            - --output
-            - json
-            - --filter
-            - event={{ .Values.events }}
-          {{- end }}
           {{- if .Values.postee.enabled }}
           args:
             - --webhook={{ .Values.postee.webhook }}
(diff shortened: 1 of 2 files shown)


deploy/helm/tracee/templates/daemonset.yaml
```diff
@@ -30,14 +30,6 @@ spec:
         - name: tracee
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
-          {{- if .Values.everythingIsAnEvent }}
-          command: ["/tracee/tracee"]
-          args:
-            - --output
-            - json
```
yanivagman, line 37: Don't we still need to set the output as json?

josedonizetti: This was all under the flag, which we are removing here. For the new binary, several things will change, and they are handled on the PR https://github.com/aquasecurity/tracee/pull/2786

rafaeldtinoco (approved):
LGTM

yanivagman (approved):

geyslan (commented):
LGTM

PR #2789, opened 2023-03-01
printer: block instead of drop events for broadcast

josedonizetti:

This PR is changing the printer broadcast to block instead of drop events. The use of nonblocking here is dropping a lot of events, we could use a timeout to give some or increase the size of the buffer, but this needs to be investigated deeper. 

The previous solution was a blocking one, based on the fact that tracee core is ready to hold events, with a buffer 10 times bigger than the buffers for the printers -> https://github.com/aquasecurity/tracee/blob/main/pkg/ebpf/events_pipeline.go#L102

### 2. Explain how to test it

we will test it on the e2e to validate it fixes the flaky tests created

### 3. Other comments

I will create a issue to investigate it further, but considering we have a release this week, blocking is the best solution

### DIFF
--- a/pkg/cmd/printer/broadcast.go
+++ b/pkg/cmd/printer/broadcast.go
@@ -3,7 +3,6 @@ package printer
 import (
 	"context"
 
-	"github.com/aquasecurity/tracee/pkg/logger"
 	"github.com/aquasecurity/tracee/types/trace"
 )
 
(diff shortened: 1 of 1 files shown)


geyslan:
Testing this.

geyslan:
> we will test it on the e2e to validate it fixes the flaky tests created

Tested locally against e2e and seems to reduce raised errors, and it makes sense as we avoid dropping events by blocking them until consumption.

geyslan (approved):
LGTM

yanivagman (approved):
LGTM

Issue #2793, opened 2023-03-02
add event to indicate usage of inotify


roikol:
an event indicating an Inotify watch added on a file

Issue #1653, opened 2022-04-08
eBPF CO-RE improvements: Load BPF programs based on available kernel features


grantseltzer:
From @yanivagman: 
We recently added the new "dependencies" field into the event definition. This concept of dependencies can be expanded to describe which BPF features are required for an event to work, such as BTF support, LSM bpf programs support, ring buffer support, or a minimal kernel version. We can query which bpf features are available on tracee init (we already do that for BTF support and the kernel version), and verify that an event chosen by the user satisfies these dependencies. To support "legacy" kernels for already existing events, we can have two (or more) different "flavors" of bpf programs for the SAME event. We can then automatically choose the program that best fits the environment where tracee runs by adding a new dependency field to the probe struct (same concept of events dependencies, but this time for probes and not for events).

rafaeldtinoco:
I liked this approach a lot!

grantseltzer:
@rafaeldtinoco @yanivagman What do you think about shelling out to `bpftool feature` in order to detect kernel features? I was hoping there'd be some large bpf file that bpftool uses to load and detect errors, but the programs are actually inlined as BPF byte code within bpftool. We can recreate this functionality but that will require adding API for activating every single bpf feature first. Additionally we'd have to maintain this, and it'd be updated often. 

The downside is that we'd require bpftool as a runtime dependency. Perhaps we can embed it at compile time?

rafaeldtinoco:
I don't like the idea of relying in a specific userland tool output (even if we embed it into tracee binary and/or tracee container). I think we should approach this by attempting to load very simple eBPF programs, split by different objects, and take conclusions on that (even if not complete as bpftool does)). @yanivagman thoughts ?

yanivagman:
I agree with @rafaeldtinoco.
We should fibd the way to look for features in libbpfgo

PR #2794, opened 2023-03-02
feat: add inotify_find_inode event

roikol:
issue: #2793

```
Date:   Thu Mar 2 11:58:33 2023 +0200

    feat: add inotify_find_inode event
    
    add the inotify_find_inode event to indicate an inotify watch being
    added to a file
```

### 2. Explain how to test it

command 1: `./dist/tracee-ebpf -f e=inotify_find_inode`
command 2: `inotifywatch /tmp/test.txt`

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -4132,6 +4132,36 @@ int BPF_KPROBE(trace_ret_file_modified)
     return 0;
 }
 
+SEC("kprobe/inotify_find_inode")
+TRACE_ENT_FUNC(inotify_find_inode, INOTIFY_WATCH);
+
+SEC("kretprobe/inotify_find_inode")
+int BPF_KPROBE(trace_ret_inotify_find_inode)
+{
+    program_data_t p = {};
+    if (!init_program_data(&p, ctx))
+        return 0;
+
+    args_t saved_args;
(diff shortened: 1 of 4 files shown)


yanivagman:
Looks like kernel 5.4 introduced the security_path_noitfy which would have been a better choice:
https://elixir.bootlin.com/linux/v5.4.234/source/security/security.c#L920
However, since we want to keep support for 4.18 kernels, we can't use it as is. So this is just a note for after we will implement a mechanism to selectively load probes according to the environment.
In the meanwhile, I think you should consider renaming this event to something more abstract (so we can change the probe without renaming the event in the future)

roikol:
> Looks like kernel 5.4 introduced the security_path_noitfy which would have been a better choice: https://elixir.bootlin.com/linux/v5.4.234/source/security/security.c#L920 However, since we want to keep support for 4.18 kernels, we can't use it as is. So this is just a note for after we will implement a mechanism to selectively load probes according to the environment (#1653 ). In the meanwhile, I think you should consider renaming this event to something more abstract (so we can change the probe without renaming the event in the future)

yeah i saw this lsm but decided to go with `inotify_find_inode` to be compatible with all supported kernel versions.
i'll rename to event to `inotify_watch`, sounds good?

rafaeldtinoco (approved):
LGTM.

rafaeldtinoco:
Im merging now as this is a simple event and comment from Yaniv has already been noticed. Thanks!

PR #2795, opened 2023-03-02
k8s: remove falcosidekiq yaml

josedonizetti:

This PR removes the falcosidekick yaml from our repo. Instead we will write a tutorial on how to use falcosidekick with tracee, this will simplify maintenance as the user can install falcosidekick directly from its repository.


geyslan (approved):
LGTM

Is there already an issue for the next step (tutorial)?

josedonizetti:
@geyslan https://github.com/aquasecurity/tracee/issues/2796

PR #2800, opened 2023-03-02
gptdocs: add option to generate docs for a list of events

rafaeldtinoco:
This is a fix to tracee-gptdocs (currently dead locked due to a last min addiction).

### DIFF
--- a/pkg/cmd/gptdocs.go
+++ b/pkg/cmd/gptdocs.go
@@ -26,15 +26,11 @@ const (
 	outputDirectory = "./dist/gptdocs"
 )
 
-type handleInfo struct {
-	eventName   string
-	programName string
-}
-
 type GPTDocsRunner struct {
 	OpenAIKey         string
 	OpenAITemperature float64
 	OpenAIMaxTokens   int
+	GivenEvents       []string
 }
 
 type WorkRet struct {
@@ -58,6 +54,8 @@ func (r GPTDocsRunner) Run(ctx context.Context) error {
 
 	var wg sync.WaitGroup
 
+	// Go routines pool to handle work
+
 	for i := 0; i < amountOfWorkers; i++ {
 		wg.Add(1)
 		go func() {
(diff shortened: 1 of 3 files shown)


pkg/cmd/urfave/urfave.go
```diff
@@ -198,10 +198,13 @@ func getContainerMode(cfg tracee.Config) printer.ContainerMode {
 	return containerMode
 }
 
-func GetGPTDocsRunner(c *cli.Context, k string, t float64, m int) (*cmd.GPTDocsRunner, error) {
+func GetGPTDocsRunner(c *cli.Context, k string, t float64, m int, e []string) (
```
geyslan, line 201: context is not being used?

pkg/cmd/gptdocs.go
```diff
@@ -98,39 +98,71 @@ func (r GPTDocsRunner) Run(ctx context.Context) error {
...
 
+		// Check if the event is in the given list of events and skip if it not
+
+		if len(r.GivenEvents) > 0 {
+			found := false
+			for _, given := range r.GivenEvents {
+				if strings.Contains(evt.Name, given) {
+					found = true
+				}
+			}
```
geyslan, line 151: should break when found?

pkg/cmd/gptdocs.go
```diff
@@ -98,39 +98,71 @@ func (r GPTDocsRunner) Run(ctx context.Context) error {
...
 				return nil
 			default:
+				logger.Debug("file already exists", "file", fileName)
 				continue
 			}
 		}
 
+		// Check if the event is in the given list of events and skip if it not
+
+		if len(r.GivenEvents) > 0 {
```
geyslan, line 145: It's safe to loop over slice (which can be empty), so we can get rid of this check.

geyslan (approved):
LGTM

pkg/cmd/urfave/urfave.go
rafaeldtinoco: true, removing

pkg/cmd/gptdocs.go
rafaeldtinoco: nope, this is correct. continue has to be outside this range loop.

pkg/cmd/gptdocs.go
rafaeldtinoco: I want the debug and if I remove the check the debug would always trigger.

PR #2812, opened 2023-03-05
standardize error/log first letter

geyslan:

    cleanup: standardize error first letter
    
    - errors first letter must be in lower case, as the errors are returned
      and can be formatted, merged or wrapped in the process.

    cleanup: standardize log first letter
    
    - log first letter must be in upper case, as it's from the first
      word being written.

Part of effort:
- #2477


rafaeldtinoco (approved):
LGTM

PR #2813, opened 2023-03-05
ebpf: remove bin_args_map

yanivagman:

Use event_data_map to store bin_args_t and remove bin_args_map. This also has the benefit of reducing instruction count and complexity of the send_bin_helper() function.

### 2. Explain how to test it

Use the `--capture` write flag to verify functionality still works as expected

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -2211,6 +2211,22 @@ enum bin_type_e
     SEND_KERNEL_MODULE,
 };
 
+static __always_inline u32 tail_call_send_bin(void *ctx,
+                                              program_data_t *p,
+                                              bin_args_t *bin_args,
+                                              int tail_call)
+{
+    if (p->event->buf_off < ARGS_BUF_SIZE - sizeof(bin_args_t)) {
+        bpf_probe_read(&(p->event->args[p->event->buf_off]), sizeof(bin_args_t), bin_args);
+        if (tail_call == TAIL_SEND_BIN)
+            bpf_tail_call(ctx, &prog_array, tail_call);
+        else if (tail_call == TAIL_SEND_BIN_TP)
+            bpf_tail_call(ctx, &prog_array_tp, tail_call);
+    }
+
+    return 0;
+}
+
 static __always_inline u32 send_bin_helper(void *ctx, void *prog_array, int tail_call)
 {
     // Note: sending the data to the userspace have the following constraints:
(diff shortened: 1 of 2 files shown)


rafaeldtinoco:

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2599,10 +2583,14 @@ static __always_inline int capture_file_write(struct pt_regs *ctx, u32 event_id)
             bin_args.full_size = io_vec.iov_len;
         }
     }
-    bpf_map_update_elem(&bin_args_map, &id, &bin_args, BPF_ANY);
 
-    // Send file data
-    bpf_tail_call(ctx, &prog_array, TAIL_SEND_BIN);
+    if (p.event->buf_off < ARGS_BUF_SIZE - sizeof(bin_args_t)) {
```
NDStrahilevitz, line 2587: I see this logic repeating itself throughout the PR, we should probably make it a util function

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2233,13 +2233,13 @@ static __always_inline u32 send_bin_helper(void *ctx, void *prog_array, int tail
...
+    u32 zero = 0;
 
-    bin_args_t *bin_args = bpf_map_lookup_elem(&bin_args_map, &id);
-    if (bin_args == 0) {
-        // missed entry or not traced
-        return 0;
-    }
+    event_data_t *event = bpf_map_lookup_elem(&event_data_map, &zero);
+    if (!event || (event->buf_off > ARGS_BUF_SIZE - sizeof(bin_args_t)))
+        return -1;
```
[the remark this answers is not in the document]
yanivagman: Actually should have returned 0. Will fix, thanks.

pkg/ebpf/c/tracee.bpf.c
yanivagman: I see what you mean but I don't think it is a good idea to hide a tail call behind a function and anyway its only 3 lines of code, so I prefer keeping it as is

NDStrahilevitz: I think we could name it something like `tail_call_send_bin` and it would make the side effect explicit in it's name, but I also understand the caution taken.
Perhaps a macro would fit here better, something like a `send_bin_probe`, which would work like the `BPF_KPROBE` macro but embed this logic in addition?

NDStrahilevitz (approved):
LGTM overall, left a last comment for consideration on the tail call procedure matter.

pkg/ebpf/c/tracee.bpf.c
yanivagman: done

PR #2814, opened 2023-03-05
fix: signing problem by using crane to find digest 

### LINKED ISSUE #2804: container image signing doesn't work
https://github.com/aquasecurity/tracee/pull/2607 added cosign to the release pipeline
https://github.com/aquasecurity/tracee/actions/runs/4326096380/jobs/7553096886 was released with it
but it seems something went wrong:
image

@developer-guy do you have any idea?

while someone is looking it this, it's good to add in the `release-snapshot` workflow so it can be tested

developer-guy:
### 3. Other comments

Fixes #2804 

`docker image inspect` throws an error like the following:

```shell
 at <index .RepoDigests 0>: error calling index: reflect: slice index out of range
```

because we need to re-pull the image to be able to make the `.RepoDigests` field to be filled, so, to avoid pulling the image, we can use the crane tool.

PTAL @itaysk


josedonizetti (commented):
@developer-guy Any idea why locally I can't emulate the error. After a pull I have the digest, since it is printed after the push?

```
docker image push josedonizetti/test1:1
The push refers to repository [docker.io/josedonizetti/test1]
6b34e7279374: Pushed
202fe64c3ce3: Pushed
1: digest: sha256:d92276b79031b703ffc65c710805d97512948bf7eaa3602c7d84fa84ea6cecab size: 736
```

and if I do an inspect I can get it?

developer-guy:
In the pipeline, we build the image and we push it to the registry then we look at the digest of it but to be able to do that we need to pull back the image from the registry, since we didn't do that, we couldn't get the digest information, so, to avoid pulling image, we can use crane since it doesn't require you to pull images to get its digest.

rafaeldtinoco:
Heads up: I'm changing that file for the arm64 support being added. I might do this change already in my PR and give credits to @developer-guy there. Please hold the merge.

rafaeldtinoco:
@developer-guy I left to you to update this PR and add appropriate cosigns into the container images being currently built (check Makefile.release for both x64 and arm64 images now). 

https://github.com/aquasecurity/tracee/blob/eb2d969ea576873043f295a4a3b2fac316e6cf28/builder/Makefile.release#L196-L225

Thank you for your work!

developer-guy:
Cool, I'm going to do this ASAP

developer-guy:
I rebased from main, PTAL @rafaeldtinoco

rafaeldtinoco (approved):
LGTM

rafaeldtinoco:
Merged at: https://github.com/aquasecurity/tracee/pull/2869 with same credits. Thanks for your contribution.

PR #2824, opened 2023-03-07
Fix make error when output for xargs is empty

pomozoff:

Fix make error when output for xargs is empty

### 2. Explain how to test it

Run `BTFHUB=0 make -f builder/Makefile.tracee-container build-tracee` on a system without any docker images.

### 3. Other comments

The next command should return an empty result:
```bash
	$(CMD_DOCKER) images \
		--filter "label=AS=tracee-make" \
		--format "{{.ID}}"
```

### DIFF
--- a/builder/Makefile.tracee-container
+++ b/builder/Makefile.tracee-container
@@ -105,7 +105,7 @@ build-tracee: \
 	$(CMD_DOCKER) images \
 		--filter "label=AS=tracee-make" \
 		--format "{{.ID}}" \
-		| xargs $(CMD_DOCKER) rmi
+		| xargs -r $(CMD_DOCKER) rmi
 
 .PHONY: build-tracee-full
 build-tracee-full: \
@@ -122,7 +122,7 @@ build-tracee-full: \
 	$(CMD_DOCKER) images \
 		--filter "label=AS=tracee-make" \
 		--format "{{.ID}}" \
-		| xargs $(CMD_DOCKER) rmi
+		| xargs -r $(CMD_DOCKER) rmi
 
 #
 # run tracee and tracee-full


builder/Makefile.tracee-container
```diff
@@ -105,7 +105,7 @@ build-tracee: \
 	$(CMD_DOCKER) images \
 		--filter "label=AS=tracee-make" \
 		--format "{{.ID}}" \
-		| xargs $(CMD_DOCKER) rmi
+		| xargs -r $(CMD_DOCKER) rmi
```
rafaeldtinoco, line 108: Actually the idea here was to save space:

image

by removing the "tracee-make" container image. But the image is named "ubuntu-tracee-make" and the docker filter is not picking it.

rafaeldtinoco (changes requested):
The `-r` is a good addition but we should also make sure tracee-make container image is removed.

builder/Makefile.tracee-container
pomozoff: Actually, the `Makefile.tracee-make" script generates this container. I believe, the `Makefile.tracee-container` script should not be responsible to clean up after the first one.

rafaeldtinoco: But this tracee-make container used for the container building process and shouldn't be used by anything else. Why do you think we shouldn't clean it ?

You can have the "tracee-make" container using:

```
$ make -f builder/Makefile.tracee-make help

To create an alpine-tracee-make container:

    $ make -f builder/Makefile.tracee-make alpine-prepare

To create an ubuntu-tracee-make container:

    $ make -f builder/Makefile.tracee-make ubuntu-prepare

To execute an alpine-tracee-make shell:

    $ make -f builder/Makefile.tracee-make alpine-shell

To execute an ubuntu-tracee-make shell:

    $ make -f builder/Makefile.tracee-make ubuntu-shell

Examples:

    $ make -f builder/Makefile.tracee-make ubuntu-make ARG="clean"
    $ make -f builder/Makefile.tracee-make ubuntu-make ARG="bpf-core"
    $ make -f builder/Makefile.tracee-make ubuntu-make ARG="tracee-ebpf"

Tell tracee-make to do STATIC builds:

    $ STATIC=0 make -f builder/Makefile.tracee-make alpine-make ARG="tracee-ebpf"
    $ STATIC=1 make -f builder/Makefile.tracee-make alpine-make ARG="tracee-ebpf"
```

rafaeldtinoco:
This has been fixed by commit 3442a928.

PR #2825, opened 2023-03-07
ebpf: remove params_type_map and use events_map instead

yanivagman:

The recently introduced events_map should carry all the configuration data required for the event. Currently it has only one value specifying for which scopes the event should be submitted. However, we can also use it to store the encoded event's parameter types which is currently being done using the params_type_map. This has two advantages - First, it removes the params_type_map which reduces memory footprint and complexity. Second, it saves an extra map lookup in the params_type_map, which is located in syscalls submit hotpath.


rafaeldtinoco:

NDStrahilevitz (approved):
LGTM, nice optimization 
(also needs a rebase)

Issue #2783, opened 2023-02-28
Refactor logger.{NewErrorf,ErrorFunc}


rafaeldtinoco:
## Description

According to the discussion at https://github.com/aquasecurity/tracee/pull/2773#discussion_r1119379506, the functions dealing with error messages printing should be refactored into an errors package or something similar containing both functions created/modified by that PR:

- newErrorf
- ErrorFunc

PR #2842, opened 2023-03-09
errfmt: introduce new package for error formatting

geyslan:

commit 69d1a2ac61418a0e336ca077f68688634173c2b3

    errfmt: refactor to use pkg/errfmt

commit 52d176ce8f17888cbf3cf02ab95e91bc72a9e4f8

    errfmt: introduce new package for error formatting
    
    This commit introduces a new error-formatting package, moving and
    re-factoring existing error-formatting code from the logger package
    into it.

Fixes: #2783

- #2783

### DIFF
--- a/cmd/tracee-rules/input.go
+++ b/cmd/tracee-rules/input.go
@@ -12,12 +12,13 @@ import (
 	"kernel.org/pub/linux/libs/security/libcap/cap"
 
 	"github.com/aquasecurity/tracee/pkg/capabilities"
+	"github.com/aquasecurity/tracee/pkg/errfmt"
 	"github.com/aquasecurity/tracee/pkg/logger"
 	"github.com/aquasecurity/tracee/types/protocol"
 	"github.com/aquasecurity/tracee/types/trace"
 )
 
-var errHelp = logger.NewErrorf("user has requested help text")
+var errHelp = errfmt.Errorf("user has requested help text")
 
 type inputFormat uint8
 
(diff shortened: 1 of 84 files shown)


geyslan:
> [ERROR] GitHub self host service failed connecting to GitHub. Infrastrcuture team has been updated as wel, please hold until investigation.

Those spawned when I cancelled checks workflow.

geyslan:
> [ERROR] GitHub self host service failed connecting to GitHub. Infrastrcuture team has been updated as wel, please hold until investigation.

@rafaeldtinoco could this message be related to this warning?

image

rafaeldtinoco:
> @rafaeldtinoco could this message be related to this warning?

Nope. I have warned devops already about the spam (independently of an error, there shouldn't be a spam like that in public issues). They always check the other errors, so its just a question of retrying from our side (and let them review the errors whenever they can).

pkg/errs/errors.go
```diff
@@ -0,0 +1,47 @@
+package errs
```
NDStrahilevitz, line 1: The name of the package doesn't really fit imo, since this isn't a generic errors package (to contrast see the standard package).
IIRC these wrappers are specifically used for debug errors right? So I think `debug` or `debugerr` would be a more fitting name.

rafaeldtinoco: Yep, it is not debug as well. Lets call it errfmt or something like it. Not a big deal IMO.

rafaeldtinoco (changes requested):
@geyslan could u rename the pkg then ? Tks!

geyslan:
> @geyslan could u rename the pkg then ? Tks!

Done.

rafaeldtinoco (approved):

PR #2845, opened 2023-03-10
rename scopes related to policies

geyslan:

commit a0fdbb68cc8db95effb5195319e3b33506ab0d2b

    go.mod: bump types

commit c774757442f9aad2d84439a98f530ae597f5af6a

    user: rename scope references to policy
    
    This continues the renaming of scope to policy, in user land.

commit b3611d7a5420dc2e8643253590df6666b1c67cb2

    bpf: rename fields related to policies
    
    event_context_t, net_task_context_t:
      matched_scopes -> matched_policies
    
    This also renames other parts that are related to policies.

### 3. Other comments

This is part of:

- #2665


yanivagman:
Making this change also requires changes in userspace, i.e Context of bufferdecoder and Event matchedScopes

geyslan:
> Making this change also requires changes in userspace, i.e Context of bufferdecoder and Event matchedScopes

I was leaving the user land changes to one single PR, since we'll have to bump types and so. Let me check it again. 👍🏻

geyslan:
The checks will fail as we have a [change](https://github.com/aquasecurity/tracee/pull/2845/commits/9e8ef3ce925cd8c51441cb5e01628118177a8e20) in types pkg. After this is merged, we need to bump the go.mod.

rafaeldtinoco:

pkg/cmd/flags/filter.go
```diff
@@ -121,31 +121,31 @@ type filterFlag struct {
 	full              string
 	filterName        string
 	operatorAndValues string
-	scopeIdx          int
+	policyIdx         int
```
yanivagman, line 124: and this will be removed as well (and all following changes to filterFlag)?

pkg/cmd/flags/filter.go
```diff
@@ -88,28 +88,28 @@ Examples:
...
+  -f 6:event=openat -f 6:comm=id -f 7:event=close -f 7:comm=id - trace in policy 6 only openat event from id command
                                                                | and
-                                                               _ trace in scope 7 only close event from id command
+                                                               _ trace in policy 7 only close event from id command
 
-  -f 3:event=openat -f 3:comm=id -f 9:event=close              - trace in scope 3 only openat event from id command
+  -f 3:event=openat -f 3:comm=id -f 9:event=close              - trace in policy 3 only openat event from id command
                                                                | and
-                                                               - trace in scope 9 only close event from all
+                                                               - trace in policy 9 only close event from all
```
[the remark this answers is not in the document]
geyslan: Definitely. Just renaming it accordingly. Once we receive the policies being loaded, this CLI will be removed, as that part of documentation.

pkg/cmd/flags/filter.go
geyslan: We need to think since it's part of the infrastructure and not exposed. It should be the part or not of the policy name mapping. We'll see.

josedonizetti: @yanivagman why should we remove the index?

geyslan:
> The checks will fail as we have a [change](https://github.com/aquasecurity/tracee/pull/2845/commits/9e8ef3ce925cd8c51441cb5e01628118177a8e20) in types pkg. After this is merged, we need to bump the go.mod.

Please, after approval, don't merge this before I remove this temp https://github.com/aquasecurity/tracee/pull/2845/commits/e71c43d38c1570fae17c1b163d64949d4cdb6a70 commit.

pkg/cmd/flags/filter.go
josedonizetti: But we are using it for policies. Policies are parsed and generate an []filterFlag

yanivagman: So at least we better rename this struct and not call it filterFlag.

yanivagman: Sure. This should probably be part of the removal of multiscopes from the CLI

types/trace/trace.go
```diff
@@ -37,7 +37,7 @@ type Event struct {
 	PodSandbox          bool         `json:"podSandbox"`
 	EventID             int          `json:"eventId,string"`
 	EventName           string       `json:"eventName"`
-	MatchedScopes       uint64       `json:"matchedScopes"`
+	MatchedPolicies     uint64       `json:"matchedPolicies"`
```
geyslan, line 40: @yanivagman a question for a next PR:

Does it make sense to expose in `Event` type a field like `MatchedPoliciesNames []string`?

geyslan: Right. After this merge, I'm going to tackle the ID<=>Name on top of https://github.com/aquasecurity/tracee/pull/2873.

go.mod
```diff
@@ -133,3 +120,5 @@ require (
 )
 
 replace github.com/kubernetes/cri-api => k8s.io/cri-api v0.23.5-rc.0
+
+replace github.com/aquasecurity/tracee/types => ./types
```
NDStrahilevitz, line 124: Need to split into a types PR

geyslan: Doing that while solving conflicts. 👍🏻 +1

PR #2853, opened 2023-03-13
ebpf: align execve enter and exit timestamps

yanivagman:

For almost all of the syscalls, we submit an event at sys_exit hook point. Execve, however, is a special case since we can't use the provided pointers at exit time as they are invalid. For this reason, we currently have two hookpoints for execve syscall - one at syscall entry, and the other one at syscall exit. The latter is called only when execve fails and returns an error.

Align the timestamp used by a failed execve call to the timestamp used by the syscall entry hook.

### 2. Explain how to test it

Simply start tracee with only execve event and watch for some failed execve call:
`./dist/tracee -f e=execve`
The timestamp should now be aligned.

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -440,6 +440,7 @@ int syscall__execve(void *ctx)
     if (!p.task_info->syscall_traced)
         return -1;
     syscall_data_t *sys = &p.task_info->syscall_data;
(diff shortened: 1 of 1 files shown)


rafaeldtinoco:

rafaeldtinoco (approved):
LGTM

rafaeldtinoco:
Im merging this as it is a simple change and I doubt you won't merge it tomorrow like it is.

Issue #2854, opened 2023-03-13
add list of helpers to bpf_attach event


roikol:
`bpf_attach` event currently contains two expicitly chosen helpers - `probe_write_user` and `override_return`.
instead of having them as arguments, we should have the list of all helpers used by the bpf program, and then use that the list in the signatures as we see fit

PR #2855, opened 2023-03-13
feat: add helpers list to bpf_attach

roikol:
issue: #2854

```
Date:   Wed Mar 15 17:01:14 2023 +0200

    feat: add security_bpf_prog event

    add security_bpf_prog event with details about the program

Date:   Mon Mar 13 14:57:11 2023 +0200

    feat: add helpers list to bpf_attach
    
    add full list of helpers used in the bpf program to the bpf_attach 
    event instead of prog_write_user and prog_override_return explicit args
```

### 2. Explain how to test it

command: `./dist/tracee-ebpf -f e=bpf_attach`

### DIFF
--- a/docs/docs/events/builtin/extra/security_bpf_prog.md
+++ b/docs/docs/events/builtin/extra/security_bpf_prog.md
@@ -1,0 +1,49 @@
+# security_bpf_prog
+
+## Intro
+security_bpf_prog - Do a check when the kernel generate and return a file descriptor for BPF programs.
+
+## Description
+This event marks the act of getting a file descriptor of a BPF program. It is triggered when the BPF program is being 
+loaded, or when the user asks for it explicitly. This event is of type LSM hook.
+The event contains data about the BPF program, and whether it is currently being loaded or not. 
+
+
+## Arguments
+* `type`:`int`[K] - the BPF program type.
+* `name`:`const char*`[K] - the BPF program name (first 16 bytes only, as this is how it is saved in the kernel).
+* `helpers`:`unsigned long[]`[K] - list of all BPF helpers being used by the BPF program.
+* `id`:`u32`[K] - the BPF program ID as set by the kernel.
+* `load`:`bool`[K] - whether this BPF program is currently being loaded.
+
+## Hooks
+### security_bpf_prog
+#### Type
+kprobe
+#### Purpose
+The LSM hook of getting a file descriptor of a BPF program. This hook triggers the event. 
+
+### bpf_check
+#### Type
+kprobe
+#### Purpose
+Save data of whether this BPF program is currently being loaded.
+
+### check_helper_call
+#### Type
+kprobe
+#### Purpose
+get information about which helper functions are used by the BPF program
+
+### check_map_func_compatibility
+#### Type
+kprobe
+#### Purpose
+get information about which helper functions are used by the BPF program
+
+## Example Use Case
+./tracee -f e=security_bpf_prog
+
+## Issues
+
+## Related Events
--- a/pkg/events/events.go
+++ b/pkg/events/events.go
@@ -263,6 +263,7 @@ const (
 	DoTruncate
 	FileModification
 	InotifyWatch
+	SecurityBpfProg
 	MaxCommonID
 )
 
@@ -6110,7 +6111,7 @@ var Definitions = eventDefinitions{
 		BpfAttach: {
 			ID32Bit: sys32undefined,
 			Name:    "bpf_attach",
-			DocPath: "security_alerts/bpf_attach.md",
+			DocPath: "docs/events/builtin/extra/bpf_attach.md",
 			Probes: []probeDependency{
 				{Handle: probes.SecurityFileIoctl, Required: true},
 				{Handle: probes.SecurityBpfProg, Required: true},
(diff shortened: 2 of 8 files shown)


rafaeldtinoco:

pkg/events/events.go
```diff
@@ -6263,6 +6264,24 @@ var Definitions = eventDefinitions{
...
 			},
 		},
+		SecurityBpfProg: {
+			ID32Bit: sys32undefined,
+			Name:    "security_bpf_prog",
+			Probes: []probeDependency{
+				{Handle: probes.SecurityBpfProg, Required: true},
+				{Handle: probes.BpfCheck, Required: true},
+				{Handle: probes.CheckHelperCall, Required: false},
+				{Handle: probes.CheckMapFuncCompatibility, Required: false},
```
yanivagman, line 6274: Why are these two not required? Do these functions exist in all kernel versions?

roikol: thanks @NDStrahilevitz ! you're right.

it is documented here: https://github.com/aquasecurity/tracee/blob/main/docs/docs/events/builtin/extra/bpf_attach.md#issues

just to elaborate:
not in all kernels these symbols exists. in some kernels these symbols could be optimized by the compiler, and it'll change the symbol name accordingly (e.g. check_helper_call.isra.0). therefore we use both of them - to cover all cases.

this behavior can result in error message from libbpf.
but we can skip this error using the logger, can't we? (i though we already did that: https://github.com/aquasecurity/tracee/blob/main/pkg/cmd/initialize/callbacks.go#L24)

yanivagman (changes requested):
Please document the changes

yanivagman (approved):
LGTM

docs/docs/events/builtin/extra/security_bpf_prog.md
```diff
@@ -0,0 +1,51 @@
...
+#### Type
+kprobe
+#### Purpose
+check whether the BPF program uses helper functions of interest
+
+## Example Use Case
+./tracee -f e=security_bpf_prog
+
+## Issues
+The 'check_helper_call' and 'check_map_func_compatibility' serves the same purpose.
```
rafaeldtinoco, line 48: This isn't true anymore. Now we have the `SetLibbpfgoCallbacks()` and `check_map_func_compatibility()` error isn't displayed. You should check if the `check_helper_call` is displayed and filter it out. We're expecting those errors so we should filter them out for now (until we can probe for the kprobes existence with feature probing feature.

The code is at:

https://github.com/aquasecurity/tracee/blob/c9ec790e1ddd43853531fa5fdee1e316238495cf/pkg/cmd/initialize/callbacks.go#L17-L24

rafaeldtinoco (changes requested):
Please change documentation stating that there will be an error and make sure that the kprobes `check_helper_call` and `check_map_func_compatibility` do not generate errors in older kernels (or whenever one, or the other, do not exist). Thanks!

pkg/events/parse_args.go
```diff
@@ -314,17 +323,232 @@ func (arg CustomFunctionArgument) Value() uint64 {
...
+	for idx, helperName := range BpfHelpers {
+		if (helperList[idx/64] & (1 << (idx % 64))) > 0 {
+			usedHelpers = append(usedHelpers, helperName)
+		}
 	}
+
+	return usedHelpers, nil
+}
+
+var BpfHelpers = []string{
```
rafaeldtinoco, line 339: We currently have 2 types in libbpfgo describing:

`type BPFProgType`
`type MapType`

With all descriptions about existing program and map types. I wonder if we should have this list there, in another type, as helper types. Or else we will start to split definitions related to libbpf internals. WDYT @yanivagman ?

yanivagman: libbpf has func_id enum (https://github.com/libbpf/libbpf/blob/b5e9722ec268702b2ec62863222277148e4fe656/include/uapi/linux/bpf.h#L5751), so libbpfgo better have it too.

This string array, however, is a kind of "toString()" function, which we will need to implement in addition to the enum itself. So I'm also in favor of having it in libbpf, but it's ok if we will have it here for now until we a PR is made in libbpfgo, then we can update here to use the libbpfgo version.

@NDStrahilevitz it is unclear if you meant here or there:

> We should probably have it **there** (or at least some form of enums **here**) also as part of the effort to not require `parse-arguments`.

docs/docs/events/builtin/extra/security_bpf_prog.md
roikol: i went over the output of all the test envs, and no warning/error was generated in regards to these probes. so we're good to remove this comment.
i'll update it now

pkg/events/parse_args.go
roikol: i've created a PR in libbpfgo to parse BPF helpers:
https://github.com/aquasecurity/libbpfgo/pull/302

rafaeldtinoco, line 339: I still think we should have a GetAll() method from libbpfgo so you can range all helpers in `parseBpfAttachHelperUsage()`. I'm merging like it is now based on what Yaniv said, but I believe we should remove this slice from here (we should maintain it in libbpfgo because this list will likely suffer 1:1 updates with libbpf).

rafaeldtinoco (approved):
I'm ok in merging this but I still think we should get rid of the slice (as described at https://github.com/aquasecurity/tracee/pull/2855/files#r1144160815).

PR #2858, opened 2023-03-13
events: add exec_binprm event

OriGlassman:

Added exec_binprm event, for the cases where execve fails, and we need to access the arguments without the risk of TOCTOU.

### 2. Explain how to test it

echo "ls" > ./bla
chmod +x ./bla
./bla


rafaeldtinoco:

yanivagman:
We already have:
1. execve(at) syscall events
2. sched_process_exec
3. security_bprm_check
4. and soon also process_execution (https://github.com/AsafEitani/tracee/blob/4e9ea832b973b78c82c2c54e84ee6cb0f1099877/docs/events/user_friendly/process_execution.md)

Why do we need to have yet another exec event?

OriGlassman:
> 

1. path from execve is unreliable (toctou)
2. only happens when execve is successful 
3. can't know if the execve is going to fail at the stage of the bprm check and no relevant return code here
4. same reasons as described above

yanivagman:
I'm still not convinced we need to introduce a fifth event for exec.
Better to try and improve one of the existing events.
One example - save the data from the probe you added here into a percpu buffer and enrich the failed execve event, if that happens.

rafaeldtinoco:

rafaeldtinoco:
Adding it to v0.13.0 milestone because there is no issue associated with the PR.

yanivagman (changes requested):
Let's not add another exec event but enrich one of the existing

pkg/events/events.go
```diff
@@ -6263,6 +6264,21 @@ var Definitions = eventDefinitions{
 				{Type: "dev_t", Name: "dev"},
 			},
 		},
+		FailedExecution: {
+			ID32Bit: sys32undefined,
+			Name:    "failed_execution",
```
AlonZivony, line 6269: All events should have the `DocPath` field pointing to the relevant documentation path.
Please create a documentation file for this event and put its path here.

AlonZivony (changes requested):

AlonZivony (commented):
I suggested some semantic changes.
I will add that as I know this event is not supported in kernel 4.19, you should make sure it won't crash in that case. I am not sure if it does, but in the case that it does crash I want you to make it give a normal error message and just run without the event.

OriGlassman:
> I suggested some semantic changes. I will add that as I know this event is not supported in kernel 4.19, you should make sure it won't crash in that case. I am not sure if it does, but in the case that it does crash I want you to make it give a normal error message and just run without the event.

So you recommend catching the failure for the specific probe? This needs to be in userspace where the programs are loaded, unless the probe is marked not required - but then it may not get loaded when it should actually get loaded, and we won't know - which is worse. 
It will only crash in 4.19 if someone selects this specific event.

pkg/ebpf/c/vmlinux.h
```diff
@@ -598,6 +598,7 @@ struct sk_buff {
 };
 
 struct linux_binprm {
+    struct mm_struct *mm;
```
roikol, line 601: i think this is not used ?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4233,6 +4233,105 @@ int BPF_KPROBE(trace_ret_inotify_find_inode)
     return events_perf_submit(&p, INOTIFY_WATCH, 0);
 }
 
+SEC("kprobe/exec_binprm")
+TRACE_ENT_FUNC(exec_binprm, EXEC_BINPRM);
```
roikol, line 4237: i think you can just use `PROCESS_EXECUTION_FAILED` here instead of `EXEC_BINPRM` because this is the event you're dealing with here.
you can remove the `EXEC_BINPRM` all together IMO

roikol (approved):
nice work!
i've left two minor comments.

pkg/ebpf/c/tracee.bpf.c
OriGlassman: In the future there'll probably be more hooks for this event. If they all use PROCESS_EXECUTION_FAILED, they may clash with each other.
If more than 1 kprobe happens (before the kretprobe) and they use the same index, it will get override.
For example:
(1) kprobe <somewhere_before_exec_binprm>
(2) kprobe exec_binprm // **overrides args of above function**
(3) kretprobe exec_binprm
(4) kretprobe <somewhere_before_exec_binprm>

OriGlassman:
@yanivagman 
> Didn't review this yet, but documentation for the new event is missing

Added documentation for the event, thanks.

AlonZivony (changes requested):
Few comments on the documenation

rafaeldtinoco:
Please rebase the PR on top of HEAD (we reverted the libbpf bump that causes the 5.4 issues you're facing).

yanivagman (approved):
LGTM

PR #2874, opened 2023-03-15
add bpf byte code capture

### LINKED ISSUE #1274: BPF program capturing
Capture BPF programs and their source code when loaded to the kernel.

AsafEitani:
fix #1274

Adds BPF programs bytecode (JIT) capture.

### 2. Explain how to test it

`tracee-ebpf -c bpf`

### DIFF
--- a/docs/classic-docs/docs/capturing/index.md
+++ b/docs/classic-docs/docs/capturing/index.md
@@ -325,3 +325,29 @@ Tracee can capture the following types of artifacts:
         Example kernel module taken from [this blog]
 
 [this blog]: https://blog.sourcerer.io/writing-a-simple-linux-kernel-module-d9dc3762c234
+
+1. **BPF programs**
+
+    Anytime a **BPF program** is loaded, the BPF bytecode will be captured.
+    The bytecode is either a JIT or assembly BPF bytecode. This is the BPF program as it was loaded by the loading process.
+    The bytecode is basically the BPF section of the compiled ELF that loads the BPF program, which contains the program instructions themselves.
+    You can disassemble the bytecode with the help of `binutils-bpf` package and the following commandline:
+    `objdump -D -b binary -m bpf <path>`
+
+     ```text
+     $ sudo ./dist/tracee-ebpf \
+        --output none \
+        --filter comm=bash \
+        --filter follow \
+        --capture clear-dir \
+        --capture bpf
+     ```
+
+    Captured bpf bytecode will be found in tracee destination directory, just like
+    any other captured file would be:
+
+     ```text
+     $ sudo ls /tmp/tracee/out/host
+       bpf.name-test_prog.pid-3668786.c8b62228208f4bdbf21df09c01046b73dd44733841675bf3c0ff969fbedab616
+     ```
+   The hex value after the last "." is the hash of the bpf bytecode.
--- a/pkg/ebpf/c/common/consts.h
+++ b/pkg/ebpf/c/common/consts.h
@@ -14,7 +14,9 @@
 #define MAX_KSYM_NAME_SIZE  64
 #define UPROBE_MAGIC_NUMBER 20220829
 #define ARGS_BUF_SIZE       32000
-#define SEND_META_SIZE      24
+#define SEND_BPF_OBJECT     4
+#define SEND_META_SIZE      28
+
 #define MAX_MEM_DUMP_SIZE   127
 
 // clang-format on
@@ -53,6 +55,7 @@
 #define OPT_CGROUP_V1             (1 << 5)
 #define OPT_PROCESS_INFO          (1 << 6)
 #define OPT_TRANSLATE_FD_FILEPATH (1 << 7)
+#define OPT_CAPTURE_BPF           (1 << 8)
 
 enum buf_idx_e
 {
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -96,6 +96,7 @@
 #include "common/memory.h"
 #include "common/network.h"
 #include "common/probes.h"
+#include "common/bpf_prog.h"
 
 char LICENSE[] SEC("license") = "GPL";
 #ifndef CORE
@@ -2997,15 +2998,39 @@ int BPF_KPROBE(trace_security_bpf)
     if (should_submit(SECURITY_BPF, p.event)) {
         // 1st argument == cmd (int)
         save_to_submit_buf(p.event, (void *) &cmd, sizeof(int), 0);
-
         events_perf_submit(&p, SECURITY_BPF, 0);
     }
-
     union bpf_attr *attr = (union bpf_attr *) PT_REGS_PARM2(ctx);
 
     reset_event_args(&p);
     check_bpf_link(&p, attr, cmd);
 
+    // Capture BPF object loaded
+    if (cmd == BPF_PROG_LOAD && p.config->options & OPT_CAPTURE_BPF) {
+        bin_args_t bin_args = {};
+        u32 pid = p.task_info->context.host_pid;
+
+        u32 insn_cnt = get_attr_insn_cnt(attr);
+        const struct bpf_insn *insns = get_attr_insns(attr);
+        unsigned int insn_size = (unsigned int) (sizeof(struct bpf_insn) * insn_cnt);
+
+        bin_args.type = SEND_BPF_OBJECT;
+        char prog_name[16] = {0};
+        long sz = bpf_probe_read_str(prog_name, 16, attr->prog_name);
+        if (sz > 0) {
+            sz = bpf_probe_read_str(bin_args.metadata, sz, prog_name);
+        }
+
+        u32 rand = bpf_get_prandom_u32();
+        bpf_probe_read(&bin_args.metadata[16], 4, &rand);
+        bpf_probe_read(&bin_args.metadata[20], 4, &pid);
+        bpf_probe_read(&bin_args.metadata[24], 4, &insn_size);
+        bin_args.ptr = (char *) insns;
+        bin_args.start_off = 0;
+        bin_args.full_size = insn_size;
+
+        tail_call_send_bin(ctx, &p, &bin_args, TAIL_SEND_BIN);
+    }
     return 0;
 }
 
@@ -3251,6 +3276,10 @@ int BPF_KPROBE(trace_security_kernel_post_read_file)
     }
 
     if (p.config->options & OPT_CAPTURE_MODULES) {
+        // Do not extract files greater than 4GB
+        if (size >= (u64) 1 << 32) {
+            return 0;
+        }
         // Extract device id, inode number for file name
         dev_t s_dev = get_dev_from_file(file);
         unsigned long inode_nr = get_inode_nr_from_file(file);
@@ -3261,7 +3290,7 @@ int BPF_KPROBE(trace_security_kernel_post_read_file)
         bpf_probe_read(bin_args.metadata, 4, &s_dev);
         bpf_probe_read(&bin_args.metadata[4], 8, &inode_nr);
         bpf_probe_read(&bin_args.metadata[12], 4, &pid);
-        bpf_probe_read(&bin_args.metadata[16], 8, &size);
+        bpf_probe_read(&bin_args.metadata[16], 4, &size);
         bin_args.start_off = 0;
         bin_args.ptr = buf;
         bin_args.full_size = size;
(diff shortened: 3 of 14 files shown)


pkg/ebpf/c/common/consts.h
```diff
@@ -61,4 +64,7 @@ enum buf_idx_e
     MAX_BUFFERS
 };
 
+#define BPF_PROG_LOAD 5
```
NDStrahilevitz, line 67: I originally aggregated all consts here as a convenience, but I believe we should overtime migrate the consts to more domain appropriate files (I think some consts already are).
These fit better in the new `bpf.h` file IMO, WDYT?

pkg/ebpf/c/common/bpf.h
```diff
@@ -0,0 +1,14 @@
+#ifndef __TRACEE_BPF_H__
```
NDStrahilevitz, line 1: 🚀

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3011,8 +3012,37 @@ int BPF_KPROBE(trace_security_bpf)
...
+            bpf_probe_read(&bin_args.metadata[16], 4, &rand);
+            bpf_probe_read(&bin_args.metadata[20], 4, &pid);
+            bpf_probe_read(&bin_args.metadata[24], 4, &insn_size);
+            bin_args.ptr = (char *) insns;
+            bin_args.start_off = 0;
+            bin_args.full_size = insn_size;
+
+            u64 id = bpf_get_current_pid_tgid();
+            bpf_map_update_elem(&bin_args_map, &id, &bin_args, BPF_ANY);
+            bpf_tail_call(ctx, &prog_array, TAIL_SEND_BIN);
```
NDStrahilevitz, line 3044: The tail call here means we skip over the following lines:
```
union bpf_attr *attr = (union bpf_attr *) PT_REGS_PARM2(ctx);

reset_event_args(&p);
check_bpf_link(&p, attr, cmd);
```

Are you sure this is ok?

pkg/ebpf/tracee.go
```diff
@@ -326,7 +332,7 @@ func New(cfg Config) (*Tracee, error) {
 	for id := range t.events {
 		evt, ok := events.Definitions.GetSafe(id)
 		if !ok {
-			return t, errfmt.Errorf("could not get event")
+			return t, errfmt.Errorf("could not get event %d", id)
```
NDStrahilevitz, line 335: Nice catch

pkg/events/events.go
```diff
@@ -6263,6 +6264,19 @@ var Definitions = eventDefinitions{
 				{Type: "dev_t", Name: "dev"},
 			},
 		},
+		CaptureBpf: {
```
NDStrahilevitz, line 6267: nit:  would better fit near the other Capture events rather than down here.

pkg/ebpf/c/tracee.bpf.c
AsafEitani: fixed, thanks

pkg/ebpf/c/vmlinux.h
```diff
@@ -1149,6 +1194,30 @@ struct icmp6hdr {
...
+    struct btf_type **types;
+    u32 *resolved_ids;
+    u32 *resolved_sizes;
+    const char *strings;
+    void *nohdr_data;
+    struct btf_header hdr;
+    u32 nr_types; /* includes VOID for base BTF */
+    // ...
+};
+
```
yanivagman, line 1220: Are these structs used in your code?

pkg/ebpf/write_capture.go
```diff
@@ -148,10 +163,20 @@ func (t *Tracee) processFileWrites(ctx context.Context) {
 				continue
 			}
 			// Rename the file to add hash when last chunk was received
-			if meta.BinType == bufferdecoder.SendKernelModule {
-				if uint64(meta.Size)+meta.Off == kernelModuleMeta.Size {
-					fileHash, _ := t.computeOutFileHash(fullname)
-					utils.RenameAt(t.outDir, fullname, t.outDir, fullname+"."+fileHash)
+			if meta.BinType == bufferdecoder.SendKernelModule || meta.BinType == bufferdecoder.SendBpfObject {
+				switch meta.BinType {
```
yanivagman, line 167: Do we really need both switch case and if? seems to be redundant

pkg/ebpf/c/vmlinux.h
```diff
@@ -5,6 +5,8 @@
 ; // don't remove: clangd parsing bug https://github.com/clangd/clangd/issues/1167
 #pragma clang attribute push(__attribute__((preserve_access_index)), apply_to = record)
 
+#define BPF_OBJ_NAME_LEN 16U
```
yanivagman, line 8: nit: this should be placed under missing_definitions.h and not here

yanivagman (commented):
one last comment and we can merge

yanivagman (approved):
LGTM

docs/classic-docs/docs/capturing/index.md
```diff
@@ -325,3 +325,25 @@ Tracee can capture the following types of artifacts:
...
+     ```text
+     $ sudo ./dist/tracee-ebpf \
+        --output none \
+        --filter comm=bash \
+        --filter follow \
+        --capture clear-dir \
+        --capture bpf
+     ```
+
+    Captured module will be found in tracee destination directory, just like
```
rafaeldtinoco, line 342: module ? Maybe a leftover from a copy paste from above ?

docs/classic-docs/docs/capturing/index.md
```diff
@@ -325,3 +325,25 @@ Tracee can capture the following types of artifacts:
...
+    any other captured file would be:
+
+     ```text
+     $ sudo ls /tmp/tracee/out/host
+       bpf.name-test_prog.pid-3668786.c8b62228208f4bdbf21df09c01046b73dd44733841675bf3c0ff969fbedab616
```
rafaeldtinoco, line 347: What is this file format ? Is it an ELF without the ELF header ? How can I use, read and/or disasemble it ? What would it be good for ? Note that the kernel module (from the above capture) is loadable, and a proper object file. This file is just "data" for me when loading a very simple eBPF object.

rafaeldtinoco:
Quick note, we pick also the libbpf `nametest` eBPF probe programs. So each capture will have the same eBPF object (probes) over and over. Not that this is a problem. Just wanted to highlight it.

docs/classic-docs/docs/capturing/index.md
rafaeldtinoco: Alright, so... if you install `binutils-bpf` in `ubuntu`, you're able to disassemble the raw binary by setting the specific arch. There is also a way of renaming/creating the ELF sections with llvm-objcopy but I wasn't able to.

This is the bytecode I got from the object that was dumped:

```
# /usr/lib/bpf/bin/objdump -D -b binary -m bpf ./bpf.name-kprobe__sys_mma.pid-86815.3c670bbf686f
fa2bb1fa1e2f2b2c21175ffa4bfd612703f5f7c4dfbc73c61d87

./bpf.name-kprobe__sys_mma.pid-86815.3c670bbf686ffa2bb1fa1e2f2b2c21175ffa4bfd612703f5f7c4dfbc73c61d87:     file format binary


Disassembly of section .data:

0000000000000000 <.data>:
   0:   b7 02 00 00 e5 07 00 00         mov %r2,0x7e5
   8:   63 2a fc ff 00 00 00 00         stxw [%fp+-4],%r2
  10:   bf a4 00 00 00 00 00 00         mov %r4,%fp
  18:   07 04 00 00 fc ff ff ff         add %r4,-4
  20:   18 12 00 00 08 00 00 00         lddw %r2,8
  28:   00 00 00 00 00 00 00 00
  30:   18 03 00 00 ff ff ff ff         lddw %r3,0xffffffff
  38:   00 00 00 00 00 00 00 00
  40:   b7 05 00 00 04 00 00 00         mov %r5,4
  48:   85 00 00 00 19 00 00 00         call 25
  50:   b7 00 00 00 00 00 00 00         mov %r0,0
  58:   95 00 00 00 00 00 00 00         exit
```

This is the bytecode from the binary that was loaded:

```
$ /usr/lib/bpf/bin/objdump -j kprobe/sys_mmap -D -m bpf ./main.bpf.o

./main.bpf.o:     file format elf64-bpfle

Disassembly of section kprobe/sys_mmap:

0000000000000000 <kprobe__sys_mmap>:
   0:   b7 02 00 00 e5 07 00 00         mov %r2,0x7e5
   8:   63 2a fc ff 00 00 00 00         stxw [%fp+-4],%r2
  10:   bf a4 00 00 00 00 00 00         mov %r4,%fp
  18:   07 04 00 00 fc ff ff ff         add %r4,-4
  20:   18 02 00 00 00 00 00 00         lddw %r2,0
  28:   00 00 00 00 00 00 00 00
  30:   18 03 00 00 ff ff ff ff         lddw %r3,0xffffffff
  38:   00 00 00 00 00 00 00 00
  40:   b7 05 00 00 04 00 00 00         mov %r5,4
  48:   85 00 00 00 19 00 00 00         call 25
  50:   b7 00 00 00 00 00 00 00         mov %r0,0
  58:   95 00 00 00 00 00 00 00         exit
```

confirming the raw eBPF bytecode saved from the code is the same.

pkg/ebpf/c/vmlinux.h
```diff
@@ -910,7 +910,52 @@ enum bpf_cmd
...
+    };
+
+    struct { /* anonymous struct for BPF_BTF_LOAD */
+        __u64 btf;
+        __u64 btf_log_buf;
+        __u32 btf_size;
+        __u32 btf_log_size;
+        __u32 btf_log_level;
+    };
+
```
rafaeldtinoco, line 958: You have added the entire type here, not only the fields/members you're using, right ? If that is the case, is this definition somewhere in the system headers? If not, then the correct place for this definition would be in `missing_noncore_definitions.h` for the full type and in `vmlinux.h` you would declare only the members you're using for the CO-RE code.

pkg/ebpf/c/common/bpf.h
rafaeldtinoco, line 1: I would not call this file `bpf.h`. There are numerous `bpf.h` files, for kernel uapi (userland headers), internal files to the kernel, and libbpf `bpf.h` (multiple: userland and kernel). I would call something else IMHO.

rafaeldtinoco (changes requested):
I'll finish the review tomorrow morning. I left some comments regarding documentation and minor stuff (headers placement, etc). Going through the rest of the code (more to get to know now, than anything else). Cheers!

pkg/ebpf/c/vmlinux.h
AsafEitani: Changed it - thanks for the catch. TBH I have little clue to what I'm doing regarding CORE and non CORE and I don't know how you divided it - I just try to make it work XD

pkg/ebpf/c/common/bpf.h
AsafEitani: changed to bpf_prog.h
=

docs/classic-docs/docs/capturing/index.md
AsafEitani: fixed - thanks

pkg/ebpf/c/vmlinux.h
rafaeldtinoco: No worries at all, we can help you find out where things go as long as it works as it did ;). Thanks, Asaf!

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2997,15 +2998,39 @@ int BPF_KPROBE(trace_security_bpf)
...
     check_bpf_link(&p, attr, cmd);
 
+    // Capture BPF object loaded
+    if (cmd == BPF_PROG_LOAD && p.config->options & OPT_CAPTURE_BPF) {
+        bin_args_t bin_args = {};
+        u32 pid = p.task_info->context.host_pid;
+
+        u32 insn_cnt = get_attr_insn_cnt(attr);
+        const struct bpf_insn *insns = get_attr_insns(attr);
+        unsigned int insn_size = (unsigned int) (sizeof(struct bpf_insn) * insn_cnt);
```
rafaeldtinoco, line 3015: @AsafEitani FTR only... We don't have a problem on using `sizeof(struct bpf_insn)` here because the `struct bpf_insn` won't ever have its size changed.

BUT, the correct approach for any other kernel types, whenever you want to get its size, is to use the bpf type helper `bpf_core_type_size(type)` for CO-RE and sizeof() for NON-CO-RE. This is major because the type might change from one kernel to another and for portability, the helper will consider that (So you don't need to declare the FULL type in the vmlinux.h file, only the types you really use).

rafaeldtinoco (approved):
LGTM

PR #2864, opened 2023-03-14
Logger in signatures

NDStrahilevitz:
### 1. Explain what the PR does

1.  logger: change method names to .Levelw
    
    Previously logger methods were of format .Level, which didn't reflect the
    internal usage of wide printing (.Levelw).
    Change is done to reflect that and later allow addition of the other
    formats (.Level and .Levelf)

2. types: add SignatureContext type for init
    
    Add a Context type for signatures in order to have a centralized struct
    for adding context injections to signatures.
    This change is included with the addition of a logger to the context.

3. signatures: signatures: use new SignatureContext
    
    Update all included signatures to use the SignatureContext type.
    Also introduce logger example in docs.

### 2. Explain how to test it

Add a logger usage in one of the provided signatures compile and trigger it (Easiest one to test is ptrace_code_injection).

PR #2880, opened 2023-03-16
types: add SignatureContext type for init

NDStrahilevitz:
Add a Context type for signatures in order to have a centralized struct for adding context injections to signatures.
This change is included with the addition of a logger to the context.

Required for #2864


geyslan (approved):
LGTM

PR #2884, opened 2023-03-17
tests: disable cache for integration tests

geyslan:

Disabling the cache allows us to run the cache multiple times and detect non-deterministic anomalies.

### 3. Other comments

Run `sudo make test-integration` more than one time and see that it no longer caches.


rafaeldtinoco (approved):
LGTM

PR #2891, opened 2023-03-21
Feature/improve symbols loaded performance

### LINKED ISSUE #2466: Improve symbols loaded map copy performance
## Prerequisites

- [x] This affects latest released version.
- [x] There isn't an issue describing the bug.

Select one OR another:

## Bug description

We noticed while running Tracee that the `symbols_loaded` event has a big performance problems.
You can see in the following graph its influence on the overall performance (true to 24th to October):
d239b5fe-b04b-4656-b979-97522cf77aee

The performance issue is the `copyMap` function used, to make the maps returned from the `HostSymbolsLoader` safe to use.
However, it seems to be a bit much for this.
We should probably find a more efficient to move the information, or even pass an unsafe map pointer.

AlonZivony:

events: avoid copying cached map in symbols loader
Copying of the cached maps in the loader proved to be heavy on the
performance because its wide use.
By using the cached map, we avoid excessive copy of the maps.

events: introduce cache to symbols_loaded
Introduce caching mechanism for watched symbols check in the
symbols_loaded derive event function.
This should further improve the performance of this heavy event.

Fix #2466

### DIFF
--- a/pkg/events/derive/symbols_loaded.go
+++ b/pkg/events/derive/symbols_loaded.go
@@ -95,7 +95,7 @@ func initSymbolsLoadedEventGenerator(
 		}
 	}
 
-	cacheLRU, _ := lru.New(1024)
+	cacheLRU, _ := lru.New(10240)
 
 	return &symbolsLoadedEventGenerator{
 		soLoader:            soLoader,
--- a/pkg/events/derive/symbols_loaded.go
+++ b/pkg/events/derive/symbols_loaded.go
@@ -4,6 +4,7 @@ import (
 	"path"
 	"strings"
 
+	lru "github.com/hashicorp/golang-lru"
 	"golang.org/x/exp/maps"
 
 	"github.com/aquasecurity/tracee/pkg/errfmt"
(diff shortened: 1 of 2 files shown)


NDStrahilevitz:
Have you tested this with pprof?

AlonZivony:
> Have you tested this with pprof?

So after testing it for a very short time with some load I received 75% decrease in CPU time.
Realistically this test is not optimal, because in a long run the cache will get full and the CPU time will increase.
I believe that we can say that it will improve the performance of the event.

rafaeldtinoco:
@NDStrahilevitz will let you merge since you asked for profiling info.

NDStrahilevitz:
> So after testing it for a very short time with some load I received 75% decrease in CPU time (running only with the `symbols_loaded` event). Realistically this test is not optimal, because in a long run the cache will get full and the CPU time will increase. I believe that we can say that it will improve the performance of the event.

Could you please include a screenshot of the before and after? I trust that you tested it right but I would like to see the difference documented here.

AlonZivony:
Sorry for the delay.
This is the flame graph of the main:
cpu_main

This is the flame graph of the new feature:
cpu

I also add the pprof files this was created from.
I will say again that I did only **30 seconds record** under stress of few steps of Linux kernel compilation (made sure that both compilations started at the same point and ended after the same amount of steps).
[pprofs.zip](https://github.com/aquasecurity/tracee/files/11090415/pprofs.zip)

rafaeldtinoco:
Lovely job @AlonZivony.

PR #2892, opened 2023-03-21
Policies

josedonizetti:

This PR adds policies support (https://github.com/aquasecurity/tracee/issues/2665). 

We have an intermediate structure (FilterMap) which is generated from policy files or flags, this structure than is used to create the internal policies tracee uses to filter events. There are a few examples under a directory which can be used for testing and also to get a grasp of how policies work, also take a look at https://github.com/aquasecurity/tracee/discussions/2640

For now, we are not supporting any actions other than `log`, `webhook` and `fluentd` are coming in a separate PR, and also not supporting actions override under the events structure.

### 2. Explain how to test it

```
tracee -p examples/new_containers.yaml # only one policy
tracee -p examples # directory with several policies
tracee -p examples/dig.yaml -p examples/new_containers.yaml # specific policies
```

### DIFF
--- a/pkg/cmd/flags/policy.go
+++ b/pkg/cmd/flags/policy.go
@@ -1,0 +1,300 @@
+package flags
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/aquasecurity/tracee/pkg/errfmt"
+	"github.com/aquasecurity/tracee/pkg/events"
+)
+
+// PolicyFile is the structure of the policy file
+type PolicyFile struct {
+	Name          string   `yaml:"name"`
+	Description   string   `yaml:"description"`
+	Scope         []string `yaml:"scope"`
+	DefaultAction string   `yaml:"defaultAction"`
+	Rules         []Rule   `yaml:"rules"`
+}
+
+// Rule is the structure of the rule in the policy file
+type Rule struct {
+	Event  string   `yaml:"event"`
+	Filter []string `yaml:"filter"`
+	Action string   `yaml:"action"`
+}
+
+// PrepareFilterMapForPolicies prepares the FilterMap for the policies
+func PrepareFilterMapFromPolicies(policies []PolicyFile) (FilterMap, error) {
+	filterMap := make(FilterMap)
+
+	if len(policies) == 0 {
+		return nil, errfmt.Errorf("no policies provided")
+	}
+
+	if len(policies) > 64 {
+		return nil, errfmt.Errorf("too many policies provided, there is a limit of 64 policies")
+	}
+
+	policyNames := make(map[string]bool)
+
+	for i, p := range policies {
+		err := validatePolicy(p)
+		if err != nil {
+			return nil, err
+		}
+
+		if _, ok := policyNames[p.Name]; ok {
+			return nil, errfmt.Errorf("policy %s already exist", p.Name)
+		}
+		policyNames[p.Name] = true
+
+		filterFlags := make([]*filterFlag, 0)
+
+		err = validateAction(p.Name, p.DefaultAction)
+		if err != nil {
+			return nil, err
+		}
+
+		// scope
+		for _, s := range p.Scope {
+			s = strings.ReplaceAll(s, " ", "")
+
+			if s == "global" && len(p.Scope) > 1 {
+				return nil, errfmt.Errorf("policy %s, global scope must be unique", p.Name)
+			}
+
+			if s == "global" {
+				break
+			}
+
+			var scope, filterName, operatorAndValues string
+
+			switch s {
+			case "follow", "!container":
+				scope = s
+				filterName = s
+				operatorAndValues = ""
+			default:
+				operatorIdx := strings.IndexAny(s, "=!<>")
+
+				if operatorIdx == -1 {
+					return nil, errfmt.Errorf("policy %s, scope %s is not valid", p.Name, s)
+				}
+
+				scope = s[:operatorIdx]
+				filterName = s[:operatorIdx]
+				operatorAndValues = s[operatorIdx:]
+			}
+
+			err := validateScope(p.Name, scope)
+			if err != nil {
+				return nil, err
+			}
+
+			filterFlags = append(filterFlags, &filterFlag{
+				full:              s,
+				filterName:        filterName,
+				operatorAndValues: operatorAndValues,
+				policyIdx:         i,
+			})
+		}
+
+		events := make(map[string]bool)
+
+		for _, r := range p.Rules {
+			err := validateEvent(p.Name, r.Event)
+			if err != nil {
+				return nil, err
+			}
+
+			// currently an event can only be used once in the policy,
+			// this might change in the future to allow multiple times the same event with different filters
+			if _, ok := events[r.Event]; ok {
+				return nil, errfmt.Errorf("policy %s, event %s is duplicated", p.Name, r.Event)
+			}
+
+			events[r.Event] = true
+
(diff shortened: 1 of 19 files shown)


examples/uid_higher_or_equal_than_zero.yaml
```diff
@@ -0,0 +1,8 @@
+name: uid_higher_or_equal_than_zero
+description: This policy traces read and write syscall for uids higher or equal than 0
```
geyslan, line 2:
```suggestion
name: uid_equal_to_or_higher_than_zero
description: This policy traces read and write syscall for uids equal to or higher than 0
```
rename the file name as well.

pkg/cmd/urfave/urfave.go
```diff
@@ -210,3 +235,74 @@ func GetGPTDocsRunner(k string, t float64, m int, e []string) (
 		GivenEvents:       e,
 	}, nil
 }
+
+func getPolicies(paths []string) ([]flags.PolicyFile, error) {
+	policies := make([]flags.PolicyFile, 0)
+
+	for _, path := range paths {
+		if path == "" {
+			return nil, errfmt.Errorf("policy path cannot be empty: %s", path)
```
geyslan, line 244: path is empty, so the formatting is not necessary. Perhaps %d (index)?

pkg/cmd/urfave/urfave.go
```diff
@@ -103,10 +108,30 @@ func GetTraceeRunner(c *cli.Context, version string, newBinary bool) (cmd.Runner
...
-	policies, err := flags.PreparePolicies(c.StringSlice("filter"))
+	var filterMap flags.FilterMap
+
+	if len(c.StringSlice("policy")) > 0 {
+		policies, err := getPolicies(c.StringSlice("policy"))
+		if err != nil {
+			return runner, err
+		}
+
+		filterMap, err = flags.PrepareFilterMapForPolicies(policies)
```
geyslan, line 119:
```suggestion
		filterMap, err = flags.PrepareFilterMapFromPolicies(policies)
```

WDYT?

pkg/cmd/urfave/urfave.go
```diff
@@ -103,10 +108,30 @@ func GetTraceeRunner(c *cli.Context, version string, newBinary bool) (cmd.Runner
...
+		if err != nil {
+			return runner, err
+		}
+	} else {
+		filterMap, err = flags.PrepareFilterMapForFlags(c.StringSlice("filter"))
```
geyslan, line 124: ditto

```suggestion
		filterMap, err = flags.PrepareFilterMapFromFlags(c.StringSlice("filter"))
```

geyslan:
Ran a few tests and all seemed good.

examples/uid_higher_or_equal_than_zero.yaml
josedonizetti: why? if the signal is `>=` higher or equal

geyslan (approved):
LGTM

yanivagman:
I want to review this one before merge

geyslan:
> I want to review this one before merge

On top of this we have this other PR:

- https://github.com/josedonizetti/tracee/pull/2

examples/context_processname.yaml
```diff
@@ -0,0 +1,10 @@
+name: context_processname
+description: This policy traces openat done by ls
+scope: 
+  - global
+defaultAction: log
+rules:
+  - event: openat
+    filter:
+      - processName=ls
```
yanivagman, line 9: let's keep the context filters uniform with the scope context

examples/pid_lower_than_1000.yaml
```diff
@@ -0,0 +1,13 @@
+name: pid_lower_than_1000
+description: This policy traces read and write syscalls for pids lower than 1000
```
yanivagman, line 2: actually pids can wrap, so better use equality (`pid=1000`) instead of inequality in this example

examples/pid_lower_than_1000.yaml
```diff
@@ -0,0 +1,13 @@
...
+scope: 
+  - global
+defaultAction: log
+rules:
+  - event: read
+    filter: 
+      - pid<1000
+  - event: write
+    filter:
+      - pid<1000
```
yanivagman, line 12: just to justify the use of event context filter and not a scope, better have different pid filters for the read and write events

examples/uid_higher_than_or_equal_than_zero.yaml
```diff
@@ -0,0 +1,9 @@
+name: uid_higher_than_or_equal_than_zero
+description: This policy traces read and write syscall for uids higher than or equal than 0
```
yanivagman, line 2: nit: greater than, equal to

examples/uid_higher_than_or_equal_than_zero.yaml
```diff
@@ -0,0 +1,9 @@
...
+scope: 
+  - uid>=1000
```
yanivagman, line 4: the file name and description say 0, but the filter has 1000

pkg/cmd/urfave/urfave.go
```diff
@@ -210,3 +235,74 @@ func GetGPTDocsRunner(k string, t float64, m int, e []string) (
...
+			return nil, err
+		}
+
+		for _, file := range files {
+			if file.IsDir() {
+				continue
+			}
+
+			// TODO: support json
+			if !strings.HasSuffix(file.Name(), ".yaml") {
```
yanivagman, line 278: `.yml` should also be an option

pkg/cmd/flags/policy.go
```diff
@@ -0,0 +1,291 @@
...
+		events := make(map[string]bool)
+
+		for _, r := range p.Rules {
+			err := validateEvent(p.Name, r.Event)
+			if err != nil {
+				return nil, err
+			}
+
+			if _, ok := events[r.Event]; ok {
+				return nil, errfmt.Errorf("policy %s, event %s is duplicated", p.Name, r.Event)
```
yanivagman, line 105: we can add a note that currently only one rule per event is supported (we may add support for more than one rule per event in the future)

pkg/cmd/flags/policy.go
```diff
@@ -0,0 +1,291 @@
...
+	if len(policies) == 0 {
+		return nil, errfmt.Errorf("no policies provided")
+	}
+
+	if len(policies) > 64 {
+		return nil, errfmt.Errorf("too many policies provided, there is a limit of 64 policies")
+	}
+
+	for i, p := range policies {
+		err := validatePolicy(p)
```
yanivagman, line 40: I think we are missing validation that the same policy name is not used more than once

yanivagman (changes requested):
Can't wait for this to be merged :-)
A few comments inline

pkg/cmd/flags/policy.go
yanivagman: Actually I meant that the error message should indicate that, something like:
"policy x uses event y in more than one rule which is not currently supported"

yanivagman (approved):
Nit comment, otherwise LGTM

pkg/cmd/flags/policy.go
josedonizetti: AHHHHH! sorry, fixing it

geyslan:
Since we now have this merged, I have josedonizetti#2 and #2919 to rebase, but depending on the last step of the #2760 types changes and that github is back up and running.

PR #2898, opened 2023-03-21
docs: add policies documentation

josedonizetti:

### DIFF
--- a/docs/docs/policies/actions.md
+++ b/docs/docs/policies/actions.md
@@ -1,0 +1,40 @@
+# Actions
+
+Policies support global actions and specific actions per event. The supported actions now are:
+
+- log
+- webhook
+
+## Default Action
+
+The default action is the action executed for every event in the policy:
+
+```
+name: rules events example
+description: This policy shows multiple events tracing
+scope:
+  - global
+defaultAction: log
+rules:
+  - event: dropped_executable
+  - event: security_file_open
+  - event: sched_process_exec
+  - event: close
+```
+
+## Event Action
+
+```
+name: rules events example
+description: This policy shows multiple events tracing
+scope:
+  - global
+defaultAction: log
+rules:
+  - event: dropped_executable
+    action:
+        - webhook:http://webhook:8080?timeout=5s
+  - event: security_file_open
+  - event: sched_process_exec
+  - event: close
+```
--- a/docs/docs/policies/index.md
+++ b/docs/docs/policies/index.md
@@ -1,0 +1,66 @@
+# Getting Started with Policies
+
+Tracee's policies are how you describe what tracee should collect and what actions it should take.
+
+For example, we would like to collect all `sched_process_exec` events from the whole host:
+
+```
+name: get started policy
+description: This policy is a get started example
+scope:
+  - global
+defaultAction: log
+rules:
+  - event: sched_process_exec
+```
+
+Here, we create a basic policy with the scope `global` (for the whole host) and the action of logging the events that match the policy.
+
+We can execute the policy and see its result:
+```
+tracee -p get_started_policy.yaml
+```
+
+Once we execute, we notice we also want to collect the event `sched_process_exit`, so we can change the policy accordingly:
+
+```
+name: get started policy
+description: This policy is a get started example
+scope:
+  - global
+defaultAction: log
+rules:
+  - event: sched_process_exec
+  - event: sched_process_exit
+```
+
+After testing, we see that for our usecase we only care about `sched_process_exit` when the user id is zero:
+
+```
+name: get started policy
+description: This policy is a get started example
+scope:
+  - global
+defaultAction: log
+rules:
+  - event: sched_process_exec
+  - event: sched_process_exit
+    filter:
+        - uid=0 
+```
+
+And, the last change is that we don't care about the whole host, only new containers, so we change our scope from global
+to `container=new`:
+
+```
+name: get started policy
+description: This policy is a get started example
+scope:
+  - container=new
+defaultAction: log
+rules:
+  - event: sched_process_exec
+  - event: sched_process_exit
+    filter:
+        - uid=0 
+```
--- a/docs/docs/policies/rules.md
+++ b/docs/docs/policies/rules.md
@@ -1,0 +1,116 @@
+# Rules
+
+Rules determine which events a policy should tracee. You can see all supported events on tracee with:
+
+```
+tracee -l 
+```
+
+## Events
+
+I want to tracee events for `dropped_executable` which is a signature event, `security_file_open`,
+`sched_process_exec` and the syscall `close`. 
+
+```
+name: rules events example
+description: This policy shows multiple events tracing
+scope:
+  - global
+defaultAction: log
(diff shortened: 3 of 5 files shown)


docs/docs/policies/rules.md
```diff
@@ -0,0 +1,114 @@
...
+rules:
+  - event: dropped_executable
+  - event: security_file_open
+  - event: sched_process_exec
+  - event: close
+```
+
+# Filters
+
+Events support filters, let's take a look at the 3 types of filters supported:
```
geyslan, line 29:
```suggestion
Events in Tracee support filters. Let's take a closer look at the 3 (three) types of filters currently supported:
```

geyslan (changes requested):

docs/docs/policies/actions.md
```diff
@@ -0,0 +1,40 @@
+# Actions
+
+Policies support global actions and specific actions per event. The supported actions now are:
```
AnaisUrlichs, line 3: Could we give a definition of what an action is? I mean people will be able to understand as they start reading the first sentences but I think it is better if we don't let any user guess anything but for us to make the statements

docs/docs/policies/actions.md
```diff
@@ -0,0 +1,40 @@
...
+
+- log
+- webhook
+
+## Default Action
+
+The default action is the action executed for every event in the policy:
```
AnaisUrlichs, line 10: Could we rewrite this sentence as:
"Every policy has a default action that will be executed for every event in same the policy"

docs/docs/policies/actions.md
```diff
@@ -0,0 +1,40 @@
...
+  - global
+defaultAction: log
+rules:
+  - event: dropped_executable
+  - event: security_file_open
+  - event: sched_process_exec
+  - event: close
+```
+
+## Event Action
```
AnaisUrlichs, line 25: after this new title, we should introduce the next YAML
"In addition to a default action, it is also possible to specify a custom action for each event within the policy. If no custom action is specified, an event will trigger the default action."

Not sure if we want to call them 'custom' action tho

docs/docs/policies/index.md
```diff
@@ -0,0 +1,66 @@
+# Getting Started with Policies
+
+Tracee's policies are how you describe what tracee should collect and what actions it should take.
```
AnaisUrlichs, line 3: "Tracee's policies describe what **<here is a word missing e.g. information>** tracee should collect and what actions it should take **in response to an event**."

Feel free to not add my suggestions

docs/docs/policies/index.md
```diff
@@ -0,0 +1,66 @@
...
+
+For example, we would like to collect all `sched_process_exec` events from the whole host:
```
AnaisUrlichs, line 5: Instead of "we" and similar pronouns, try to express it more generic ☺️ 
"For example, the following policy collects all `sched_process_exec` events from the whole host:"

AnaisUrlichs: well given you continue with "we" let's keep it:
"For example, the following policy we specify to collect all `sched_process_exec` events from the whole host:"

docs/docs/policies/index.md
```diff
@@ -0,0 +1,66 @@
...
+```
+
+Here, we create a basic policy with the scope `global` (for the whole host) and the action of logging the events that match the policy.
+
+We can execute the policy and see its result:
+```
+tracee -p get_started_policy.yaml
+```
+
+Once we execute, we notice we also want to collect the event `sched_process_exit`, so we can change the policy accordingly:
```
AnaisUrlichs, line 24: hmm, here is some information missing, why do we notice that? maybe if you just want to also add that event, we could express it differently, otherwise, people will ask what you know that they don't.
"Once we execute the policy, we decide that we also want to collect the event `sched_process_exit`. We can change the policy to collect the event `sched_process_exit`:"

docs/docs/policies/index.md
```diff
@@ -0,0 +1,66 @@
...
+description: This policy is a get started example
+scope:
+  - global
+defaultAction: log
+rules:
+  - event: sched_process_exec
+  - event: sched_process_exit
+```
+
+After testing, we see that for our usecase we only care about `sched_process_exit` when the user id is zero:
```
AnaisUrlichs, line 37: Why?

AnaisUrlichs: maybe provide further information for people who are new to user id's and the reasoning that you made in this sentence 🙃

docs/docs/policies/index.md
```diff
@@ -0,0 +1,66 @@
...
+  - global
+defaultAction: log
+rules:
+  - event: sched_process_exec
+  - event: sched_process_exit
+    filter:
+        - uid=0 
+```
+
+And, the last change is that we don't care about the whole host, only new containers, so we change our scope from global
```
AnaisUrlichs, line 52: This is an example where people can reason by themselves as to why they only care about new containers (given they are familiar with containers) -- BUT if you want to make 100% sure users understand the way you think while writing this, I would add your reasoning

docs/docs/policies/scopes.md
```diff
@@ -0,0 +1,72 @@
+# Scopes
+
+Scopes define the workload this policy cares about. For example,
+we want to know all dns events executed by the tool `dig`:
```
AnaisUrlichs, line 4: Can you provide more information here

docs/docs/policies/scopes.md
```diff
@@ -0,0 +1,72 @@
...
+description: This policy traces all processes executed on the host
+scope:
+  - !container
+defaultAction: log
+rules:
+  - event: sched_process_exec
+```
+
+
+   !!! Note
```
AnaisUrlichs, line 57: can you either provide a sentence about each -- what it is, when to use it, or a link where people can read up about it?

docs/docs/policies/rules.md
```diff
@@ -0,0 +1,116 @@
+# Rules
+
+Rules determine which events a policy should tracee. You can see all supported events on tracee with:
```
AnaisUrlichs, line 3: you mean 'trace' instead of 'tracee' right? :D

docs/docs/policies/rules.md
```diff
@@ -0,0 +1,116 @@
...
+        16) "containerId"  
+        17) "containerImage"  
+        18) "containerName"  
+        19) "podName"  
+        20) "podNamespace"  
+        21) "podUid"  
+
+### Argument Filters
+
+In addition, I want to filter `security_open_file` events to only track those occurring in `/tmp/*`. Here's how to set the policy:
```
AnaisUrlichs, line 76: but then here it reads like a guide too

docs/docs/policies/scopes.md
```diff
@@ -0,0 +1,72 @@
...
+        The following is a list of available context fields:
```
AnaisUrlichs, line 58: same here, take these lists out and put them into a docs page 
somewhere here https://aquasecurity.github.io/tracee/v0.12/docs/overview/

Issue #2902, opened 2023-03-22
print_mem_dump not working


AsafEitani:
## Description

run `tracee-ebpf -f e=print_mem_dump -e print_mem_dump.args.symbol_name=system:security_file_open
nothing pops up

The cause is this commit:
https://github.com/aquasecurity/tracee/commit/0459ae5f937628f4d4749fc3b40a83ffff9a561a

## Output of `tracee -v`:

```
v0.10.0-73-gc8850253
```

## Output of `uname -a`:

```
Linux ubuntu 5.15.0-50-generic #56~20.04.1-Ubuntu SMP Tue Sep 27 15:51:29 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

## Additional details 
The `print_mem_dump` has a workaround to allow it to be submitted although the filter doesn't match:
https://github.com/aquasecurity/tracee/blob/main/pkg/filters/args.go#L50

Something happened when the multiscope was introduced that causes the event to not be printed.
The BPF code that is submitting the event is running.
The event is received by the user land process.

rafaeldtinoco:
@geyslan u think u can check this one ?

AsafEitani:
> There is also a special case for uprobe-related events at:
> 
> ![image](https://user-images.githubusercontent.com/7395852/226924372-715c6b36-1cb5-4d45-a2fe-3091daf24997.png)
> 
> Because the `shouldProcess()` logic could not deal with them correctly, but it makes derivative events to be sent down the pipeline (instead of trying to filter them).
> 

The print memory event is not derived, could that be the reason for the bug?

geyslan:
Testing c88502537b3ecbf7e72069e91e97d2126a2b57eb, we have the right result:

![image](https://user-images.githubusercontent.com/3372117/227057059-0a51e1aa-be91-45f3-a274-69de602232a6.png)

In main, commenting out the args filter check, I made it print:

![image](https://user-images.githubusercontent.com/3372117/227057774-ced544d1-dedd-4f2d-b160-57d50127197e.png)

As you can see, the arguments are duplicated and the first ones are nil. It's the cause of filter failure.

geyslan:
![image](https://user-images.githubusercontent.com/3372117/227059134-9268d6d1-29b7-48b2-8189-84581b241083.png)

For testing:

```diff
diff --git a/pkg/ebpf/events_pipeline.go b/pkg/ebpf/events_pipeline.go
[...]

PR #2914, opened 2023-03-23
skip arg filtering for PrintMemDump

geyslan:

commit 2d3094ce22e11186c96e3eea808a3f378b8114d8

    filters: args: skip arg filtering for PrintMemDump
    
    PrintMemDump event must not be filtered out by arguments filter, as it
    a special case that uses "args" as a way to pass "params" to the event.

commit 595ebeaca3a515a6cf1a891f0b42ae6ca45615d6

    ebpf: assign processPrintMemDump args values

    Instead of appending new args to processPrintMemDump event, assign
    values to its existing args. More on that in #2617.

### 2. Explain how to test it

```shell
sudo ./dist/tracee -f e=print_mem_dump -f print_mem_dump.args.symbol_name=system:security_file_open
```

image

### 3. Other comments

Fixes: #2902


yanivagman (approved):
LGTM

PR #2921, opened 2023-03-27
types: add container and kubernetes context fields

NDStrahilevitz:
Split the Container and Pod fields into separate Containers and Kubernetes struct fields.
Add an ImageDigest field to the Container struct which represents the container image's SHA digest.

Types PR required for #2760


rafaeldtinoco (approved):
LGTM

PR #2922, opened 2023-03-27
types: event policy name

geyslan:

This adds the MatchedPoliciesNames field to the event struct. This field will be used to store the name of the policy that matched the event.

It also omits the MatchedPolicies bitmask from the event struct marshalling.

### 3. Other comments

This is required by #2923


yanivagman (approved):
LGTM

josedonizetti (approved):

geyslan:
Waiting e2e fix due to recent type changes, as reported in e2e test 705.

PR #2925, opened 2023-03-28
containers: parse ContainerID by inner cgroup

### LINKED ISSUE #2896: Use internal container ids when parsing containers inside containers
Currently container id parsing extracts the first detected container in the cgroup path.
When evaluating container ids of containers inside containers this evaluates in the internal containers being associated to the external container (in terms of id, the cgroup id is differentiated).
This can result in awkward results in signatures relying on data keyed by container id, since the signature would receive multiple `container_create` events of the same id, or even multiple `existing_container` events on startup for the same id.

As such when parsing for the container id in the `container` package we should extract the last id in the cgroup path and associate it to the container.

Potential future work - parent container:
* Add an indication for parent container
* Should it be a `parent_container_id` string field or an `internal_container` boolean field?
* Only in event context and/or `container_create`/`existing_container` argument

NDStrahilevitz:
Following changes are introduced:
1. New CgroupInfo field `ContainerRoot` is introduced to signify the directory being the root cgroup directory of a container (this solves podman container_create event duplication and is used to ensure no duplication in existing_container)
2. ContainerID from a cgroup path is now parsed by the last container id and runtime found.
This means for example that kind, a development kubernetes cluster deployed through docker and using containerd internally, will report its node containers as docker containers with their own ID and their internal pod containers as containerd containers with their ID, whereas before all containers would share the node container's runtime and ID.

How to test:

1. `tracee-ebpf -f e=container_create -o json`
2. `docker run --privileged --name dind -d -it --rm docker:stable-dind`
3. `docker exec -it dind /bin/ash`
4. `docker run --rm -d ubuntu`

The first `container_create` should have the first `container_id` in it's argument and no `container.id` in it's context.
The second `container_create` should have the internal `container_id` in it's argument and the first `container_id` argument in the `container.id` context field.

Fix #2896

### DIFF
--- a/pkg/containers/containers.go
+++ b/pkg/containers/containers.go
@@ -34,11 +34,12 @@ type Containers struct {
 
 // CgroupInfo represents a cgroup dir (might describe a container cgroup dir).
 type CgroupInfo struct {
-	Path      string
-	Container cruntime.ContainerMetadata
-	Runtime   cruntime.RuntimeId
-	Ctime     time.Time
-	expiresAt time.Time
+	Path          string
+	Container     cruntime.ContainerMetadata
+	Runtime       cruntime.RuntimeId
+	ContainerRoot bool // is the cgroup directory the root of its container
+	Ctime         time.Time
+	expiresAt     time.Time
 }
 
 // New initializes a Containers object and returns a pointer to it.
@@ -119,7 +120,7 @@ func GetContainerIdFromTaskDir(taskPath string) (string, error) {
 	}()
 	scanner := bufio.NewScanner(cgroupFile)
 	for scanner.Scan() {
-		containerId, _ := getContainerIdFromCgroup(scanner.Text())
+		containerId, _, _ := getContainerIdFromCgroup(scanner.Text())
 		if containerId != "" {
 			break
 		}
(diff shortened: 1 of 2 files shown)


yanivagman (approved):
LGTM

rafaeldtinoco:
Regarding:

```
This means for example that kind, a development kubernetes cluster
deployed through docker and using containerd internally, will report its
node containers as docker containers with their own ID and their
internal pod containers as containerd containers with their ID, whereas
before all containers would share the node container's runtime and ID.
```

When the cgroup parsing logic was added there was a discussion about supporting nested containers and the decision was not to include this scenario (to keep the 1st level containers tracked only).

Now that we decided to include, based on this PR, are we sure it works with all possible nested scenarios without regressions ?

- docker shim inside docker containerd
- docker containerd inside docker shim
- crio inside docker containerd
- docker containerd inside crio
- runc inside docker
- docker inside runc
- ...

Because we tested multiple engines/runtimes when the container enrichment was added and I'm afraid this PR addresses "kind" scenario but not all possible combinations and introduce regressions.

NDStrahilevitz:
@rafaeldtinoco Firstly, I would like to know if you have encountered any such regressions.
Second, I believe the change is relatively safe overall, as I think all `container in container` scenarios would fall into either of these two cases: 

1. All the scenarios provided will create an initial cgroup for the parent container and all containers internal to it will have its cgroup path relative to it. In that case the logic here should work fine for those, as that's exactly the scenarios i've seen generated, any scenario that generates an equivalent cgroup paths "configuration" should work the same.
2. The scenarios provided create a cgroup path for the parent container and then the internal containers are created in cgroup paths external to it. The internal containers would be parsed correctly (in terms of runtime and ID), since they are equivalent to parsing a parent container, but their parent container might be missing in the `context.container` field (TBD). This is equivalent to how we would parse the container beforehands, so no regression.

In terms of possible regressions that could be introduced here I would think it's most likely that if any error we will have errors regarding the correct runtime parsing.
As for triple nested containers (or any N nested containers), I can't make any promises but I don't think the cases for those would fall far if not neatly into the two cases i've described above.

And regardless I don't think we've tested all those previous scenarios before.
Specifically for the `runc inside docker` and `docker inside runc`, i'm not sure we handled these 100% even before, we don't have any runtime declaration for `runc` at the very least, but I assume we did parse container ids just fine.

Specifically the kind usecase does test `containerd inside docker containerd/docker shim`, for what it's worth.

PR #2930, opened 2023-03-29
events: prevent symbols map cache corruption

AlonZivony:

The current `events_collisions` event might corrupt in some cases the cache map, as it is now returned from the loader class as a reference. Change the event to copy on change the map.

Fix #2930


NDStrahilevitz (approved):
LGTM

Issue #2823, opened 2023-03-07
Tracee should have a logging option for libbpf relocation debug log


rafaeldtinoco:
Today when executing tracee we can't see the relocation messaged done by libbpf when loading the eBPF object. This was reported by @OriGlassman and it is an important feature for when someone is developing eBPF code and needs to know libbpf errors directly.

Currently a workaround is to return false in the SetLibbpfgoCallbacks() function, but we should have an official way of doing that.

Workaround:

![image](https://user-images.githubusercontent.com/7395852/223429941-34abef5d-cf67-467e-8ba6-df30f1c09f55.png)

geyslan:
Perhaps to raise a flag like `-log libbpf` which overrides all `LogFilters`. WDYT?

geyslan:
Currently we're using only single logger at package level, so a libbpf output level option should force/override the logger level as well. Remember that we're passing libbpf output directly to logger.Log().

So, `-log libbpf:debug` would implicitly be the same as `-log debug`. What can be a bit confuse.

Considering that we want to output all to the same logger end, I assume that we don't need a `-log libbpf<warn,info,debug>` option.

Moreover, I suppose that the exposed issue

> Today when executing tracee we can't see the relocation messaged done by libbpf when loading the eBPF object.


is being resolved via https://github.com/aquasecurity/libbpfgo/pull/312.

geyslan:
As the team discussed today, we agreed to use a single base logger, as it already is, but implement --log filter options, such as for packages or raw strings. It would answer all needs. 

So, I'm going to close this and open a proper issue soon.

rafaeldtinoco:
Should this be closed by now ?

PR #2933, opened 2023-03-29
flags: logger: add libbpf log level flag

geyslan:
### 3. Other comments

Fixes: #3035

~~Currently, due to a libbpfgo [issue](https://github.com/aquasecurity/libbpfgo/issues/311), all libbpf output is being set as level 0 and tests won't work accordingly.~~

### DIFF
--- a/cmd/tracee-ebpf/main.go
+++ b/cmd/tracee-ebpf/main.go
@@ -11,7 +11,6 @@ import (
 	"github.com/aquasecurity/tracee/pkg/cmd"
 	"github.com/aquasecurity/tracee/pkg/cmd/flags"
 	"github.com/aquasecurity/tracee/pkg/cmd/flags/server"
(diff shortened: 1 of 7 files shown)


NDStrahilevitz:
Is there a reason we would want to split off the libbpf log level from our own? I think it should just inherit tracee's log level instead of introducing the complexity of two separate log levels.

geyslan:
> Is there a reason we would want to split off the libbpf log level from our own? I think it should just inherit tracee's log level instead of introducing the complexity of two separate log levels.

There's no reason, I agree with you. This PR will be reworked. Please take a look at https://github.com/aquasecurity/tracee/issues/2823#issuecomment-1492398582.

geyslan:
The rework is ready for review.

geyslan:
Closing since a lot changed. https://github.com/aquasecurity/tracee/issues/3035 will be tackled in another PR.

Issue #2171, opened 2022-09-19
Wait for perf buffer to empty before exiting tracee-ebpf


AsafEitani:
## Prerequisites

- [x] This issue is an EPIC issue (add label: EPIC).
- [ ] This issue is an EPIC TASK (add issue to EPIC description).

Select one OR another:

- [ ] I'll create a PR to implement this feature (assign to yourself).
- [x] Someone else should implement this (describe it well).

## Feature description

After sending a termination signal to tracee-ebpf (Ctrl+C), tracee-ebpf should first kill the bpf programs, and continue to consume events from the perf buffer in the go userspace. This will empty the perf buffer and no event will be lost due to premature exit.

danielciuraru86:
We noticed this issue when running tracee in CI.
We created a workflow where we generate a few system events (network, process_exec, filesystem) and the last one before the tracee termination signal was not always present in the output. (the type of the last event did not matter)

Snippet from our workflow

```
      - name: Start Tracee public
        uses: aquasecurity/tracee-action@v0.3.0-start

      - name: Create Profile Data
        run: |

          echo "Creating detection data"
          date
          nmap -sV -p 80 localhost

          echo "Creating container data"
          sudo docker run -d alpine:3.17.1 sleep 1

          echo "Creating filesystem data"
          echo "console.log('Addition');" > testdata/files/to-add.js
          echo "console.log('Modification');" >> testdata/files/to-modify.js
          rm -f testdata/files/to-delete.js

          echo "Creating network call data"
          curl -sL https://facebook.com > /dev/null
          
      - name: Stop Tracee
        uses: aquasecurity/tracee-action@v0.3.0-stop
```

The last curl to facebook.com sometimes is missed

rafaeldtinoco:
@danielciuraru86 and @tamirkiviti13 be aware that the network code needs the work being done at https://github.com/aquasecurity/tracee/pull/2716 to be WAY more reliable (and don't miss network flows).

For your awareness, you might be experiencing the results of that bug (and not this one).

yanivagman:
> @danielciuraru86 and @tamirkiviti13 be aware that the network code needs the work being done at #2716 to be WAY more reliable (and don't miss network flows).
> 
> For your awareness, you might be experiencing the results of that bug (and not this one).

They mentioned to me that adding delay at the end of the run resolved the problem, so I think it is indeed related to this one.

danielciuraru86:
[...]

Issue #1377, opened 2022-01-23
Bug: Sync issues between Tracee.Run() return and its goroutines finish


AlonZivony:
Currently, when tracee-ebpf receive the exit signal it finishes the `Tracee.Run()` function and return to `main`.
In `main`, the `printer.Close()` method is called which disable future writing to the printer.

However, the events pipeline isn't necessary closed when the printer is closed.
This might lead to events loss from the output, and sync problem between the printer stats and the actual printed values.

Other similar issues might happen with other goroutines of the function.

AlonZivony:
Probably the solution to this issue will be to wait for all goroutines of `Tracee.Run()` to finish before exiting.

AlonZivony:
I am not sure if I have the time to take this in.
If someone else think he can do it he can assign himself.

mtcherni95:
is this relevant?

rafaeldtinoco:
@NDStrahilevitz feel free to close this if you think there is nothing to be done.

AlonZivony:
@NDStrahilevitz I ping you on it again

rafaeldtinoco:
Isn't this a duplicate of https://github.com/aquasecurity/tracee/issues/2171 ?

AsafEitani:
Not quite but it makes sense to solve those two together

geyslan:
I'm tackling https://github.com/aquasecurity/tracee/pull/2688 and facing race issues, one solution is being to change Tracee.running bool flag to atomic.Bool.

PR #2969, opened 2023-04-03
consume remaining events after ctx is done

geyslan:

    consume remaining tracee events after ctx is done
    
    This also extracts the dispatch logic into the function processEvent
    to reuse it in the drain loop.
    
    Tests has also been updated reordering object creation and had the race
    detector, "-race" flag, disabled. It must be re-enabled in the
    one binary test suite effort.

---

    drain and print events after tracee ctx is done

### 3. Other comments

Fixes: #1377
Fixes: #2171

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -720,6 +720,7 @@ test-unit: \
 	$(CMD_GO) test \
 		-tags ebpf \
 		-short \
+		-race \
 		-v \
 		-coverprofile=coverage.txt \
 		./cmd/... \
@@ -733,6 +734,7 @@ test-types: \
 	# Note that we must changed the directory here because types is a standalone Go module.
 	cd ./types && $(CMD_GO) test \
 		-short \
+		-race \
 		-v \
 		-coverprofile=coverage.txt \
 		./...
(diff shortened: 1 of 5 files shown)


AsafEitani (approved):
LGTM

geyslan:
> LGTM I think there is something similar needed in tracee-rules

Yes. I added other commit tackling tracee-rules part. This one is breaking unit-tests due to data-race detection. So I run an e2e test (740) against this PR and it passed.

As we already have issues with tracee-rules tests,

https://github.com/aquasecurity/tracee/blob/a907a350ddac27edb0251a5c9af0c842d39b6a67/pkg/signatures/engine/engine_test.go#L340-L345

I ask you for an in-depth review and wonder if we wouldn't need to fix rules before to tackle this.

geyslan:
> Yes. I added other commit tackling tracee-rules part. This one is breaking unit-tests due to data-race detection. So I run an e2e test (740) against this PR and it passed.

I removed `-race` flag to make tests happy (for while).

rafaeldtinoco:
I want to review this one before it gets merged.

geyslan:
Re-enabled race detection to debug via workflow errors since locally it's running without erring.

image

yanivagman (approved):
LGTM

PR #2971, opened 2023-04-03
logger: refactor logger configuration

### LINKED ISSUE #2475: Logger refactors
Some of the implemented functionality of the logger is not required anymore, therefore needs to be removed for clarity:

- [x] logger pkg was implemented to provide a base logger package as default. But it also provides a mechanism to instantiate new loggers. The latter wasn't used and won't be.

- [x] logger pkg was initially implemented to provide setup through env vars. It wasn't used and won't be, specially after the [creation](https://github.com/aquasecurity/tracee/commit/7fe1f4cfaa052e087ee79608112503ff28365a77) of the --output log-file flag which can work in conjunction with the new SetWrite() and SetLevel().

- [x] logger lacks a mechanism to allow external loggers to be injected instead of it. This is relevant for consumers of tracee as a library which have their own loggers.

- [x] Logger must comply with common logging interfaces, `zap` being ideal here (since we already use it as well and it's common in other aqua products).

NDStrahilevitz:
Removed the option for configuring the logger from environemnt variables in favor of programmatic configuration.
This further exposes the option of supplying your own logger implementation and replacing tracee's default implemenation for library consumers of tracee.

Clean up the package to separate files for the utilities (log counters and caller info).

Fix #2475

### DIFF
--- a/cmd/tracee-ebpf/main.go
+++ b/cmd/tracee-ebpf/main.go
@@ -16,24 +16,6 @@ import (
 	"github.com/aquasecurity/tracee/pkg/logger"
 )
 
-func init() {
-	// Avoiding to override package-level logger
-	// when it's already set by logger environment variables
-	if !logger.IsSetFromEnv() {
-		// Logger Setup
-		logger.Init(
-			&logger.LoggerConfig{
-				Writer:    os.Stderr,
-				Level:     logger.InfoLevel,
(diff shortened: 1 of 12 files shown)


geyslan (approved):
LGTM

rafaeldtinoco:
If I don't take a look at this today feel free to merge with +1 from Geyslan.

rafaeldtinoco (approved):
LGTM

Issue #2739, opened 2023-02-17
network: events might be lost in already listening sockets


rohitburke:
## Description

In the current Tracee network events if we see in the [network_trace.go](https://github.com/aquasecurity/tracee/blob/main/types/trace/network_trace.go) file. In the ProtoIPv4, ProtoIPv6, and ProtoUDP we have a length-named parameter that I thought will give me the length of string data I send but it passes me only the header length, not a **_full packet length._** 

It can see in the below image
 
![pll](https://user-images.githubusercontent.com/30924877/219571914-cba74cf8-1005-4d98-b224-02c5b887ab43.png)

To get a full packet length is possible right now only in pcap files by adding this in the cmd line
"--capture pcap-snaplen:byte" It accepts byte, KB, max, and headers.

But these things are not possible on the event page of tracee. It would be really good if it print the exact full packet length in the output.

 
## Output of `tracee -v`:

```
v0.11.1
```

## Output of `uname -a`:

```
Linux kali 5.18.0-kali7-amd64
```

rafaeldtinoco:
Thanks for opening the bug, will try to address this next release (by end of month, lets see, will keep you posted).

rafaeldtinoco:
Ok, I had some time to come back to this one now, sorry for the delay. I believe in your example, the package containing that length was actually not capture (for SOME reason). 

> I see you're monitoring the same SRC and DST IPs, but have you started the daemon (and the client) after tracee was started or before ? That is a good answer for me to try to understand what happened (to which direction was the flow ? can you test with different IPs so it is easier to understand ?).

I have tested a few examples in my machine and, besides fixing something else I found, I believe I have reproduce the same behavior:

Let me show you what I did... I'm running 2 netcats, in 2 different IPs on the same interface:

1. 10.10.69.1/24
2. 10.10.69.2/24

And one netcat is called "othernc" and the second just "nc" (so I can differentiate both binaries and command names).

Okay, so, by executing:

- `othernc -k -4 -l -s 10.10.69.1 -p 8080 > /dev/null`

and

- `dd if=/dev/zero bs=XXX count=1 | nc -q 3 -s 10.10.69.2 10.10
.69.1 8080`

> where XXX is the number of bytes being transferred in the TCP payload (all the rest is header overhead).

## 1024 bytes

[...]

PR #2982, opened 2023-04-06
Important Network Fixes for Tracee

rafaeldtinoco:
## Issue description

The exact problem is described at:

https://github.com/aquasecurity/tracee/issues/2739#issuecomment-1503605814

And a "prove" that this PR fixes the issue is at:

https://github.com/aquasecurity/tracee/issues/2739#issuecomment-1507142356

## How to test this

```
$ sudo cp /usr/bin/nc /usr/bin/othernc

# screen 1

$ sudo ./dist/tracee-ebpf --output format:json --filter event=net_packet_ipv4 --filter comm=othernc

# screen 2

$ othernc -4 -l 8080 &

$ dd if=/dev/zero bs=1024 count=1 | nc -i 5 -q 2 localhost 8080
```

And observe the pacotes in both: wireshark and tracee output. Each existing packet in wireshark should be described as one net_packet_ipv4 event in tracee output.

## PR description

commit 5b082b40 (HEAD -> rafaelchanges, rafaeldtinoco/rafaelchanges)
Date:   Tue Apr 11 14:57:57 2023

    network: fix tracing for initial pakets on accepted socket
    
    When a "sock" is cloned because of a SYN packet, a new "sock" is created
    and the return value is the new "sock" (not the original one).
    
    There is a problem though, the "sock" does not contain a valid "socket"
    associated to it yet (sk_socket is NULL as this is running with SoftIRQ
    context). Without a "socket" we also don't have a "file" associated to
    it, nor an inode associated to that file. This is the way tracee links
    a network flow (packets) to a task.
    
    The only way we can relate this new "sock", just cloned by a kernel
    thread, to a task, is through the existence of the old "sock" struct,
    describing the listening socket (one accept() was called for).
    
    Then, by knowing the old "sock" (with an existing socket, an existing
    file, an existing inode), we're able to link this new "sock" to the task
    we're tracing for the old "sock".
    
    This patch fixes the issue. In bullets:
    
    - tracing a process that has a socket listening for connections.
    - it receives a SYN packet and a new socket can be created (accept).
    - a sock (socket descriptor) is created for the socket to be created.
    - no socket/inode exists yet (sock->sk_socket is NULL).
    - accept() traces are too late for initial pkts (socked does not exist).
    - by linking old "sock" to the new "sock" we can relate the task.
    - some of the initial packets, sometimes with big length, are traced now.
    
    Fixes: #2739

commit 9bfa2e0f
Date:   Wed Apr 5 22:15:15 2023

    network: fix net_event_context_t wrong padding
    
    When network code was added, the net_event_context_t was being padded
    in userland, instead of being correctly packed in eBPF code.
    
    This happened because I thought the initial 4 bytes (u32) of the
    []bytes length was together with the payload, when the events were
    being derived, but it wasn't. The initial 4 bytes was actually a
    padding I got right "by accident".
    
    NOTE: This is an important fix to all the network code, although the
    padding was being "somehow" done, and the wrong userland code was
    making it to work, side effects of this mistake are unknown.

commit 8cdfcb00
Date:   Wed Apr 5 15:47:42 2023

    derivation table: quick refactor for readability

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -4868,6 +4868,8 @@ CGROUP_SKB_HANDLE_FUNCTION(proto_icmpv6);
 // Network submission functions
 //
 
+// TODO: check if TCP needs a LRU map of sent events by pkt ID (avoid dups)
+
 static __always_inline u32 cgroup_skb_submit(void *map,
                                              struct __sk_buff *ctx,
                                              net_event_context_t *neteventctx,
@@ -4904,6 +4906,8 @@ static __always_inline u32 cgroup_skb_submit(void *map,
 }
 
 #define cgroup_skb_submit_event(a,b,c,d) cgroup_skb_submit(&events,a,b,c,d)
+
+// TODO: check if TCP needs a LRU map of sent events by pkt ID (avoid dups)
 
 static __always_inline u32 cgroup_skb_capture_event(struct __sk_buff *ctx,
                                                     net_event_context_t *neteventctx,
@@ -5050,9 +5054,17 @@ int BPF_KPROBE(trace_security_sk_clone)
     struct sock *osock = (void *) PT_REGS_PARM1(ctx);
     struct sock *nsock = (void *) PT_REGS_PARM2(ctx);
 
+    struct socket *osocket = BPF_READ(osock, sk_socket);
+    if (!osocket)
+        return 0;
+
     // obtain old socket inode
-    u64 inode = BPF_READ(osock, sk_socket, file, f_inode, i_ino);
+    u64 inode = BPF_READ(osocket, file, f_inode, i_ino);
     if (inode == 0)
+        return 0;
+
+    // check if old socket family is supported
+    if (!is_family_supported(osocket))
         return 0;
 
     // if the original socket isn't linked to a task, then the newly cloned
@@ -5221,7 +5233,6 @@ int BPF_KPROBE(cgroup_bpf_run_filter_skb)
         u64 oinode = *o;
 
         // with the old inode, find the netctx for the task
-
         netctx = bpf_map_lookup_elem(&inodemap, &oinode);
         if (!netctx)
             return 0; // old inode wasn't being traced as well
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -4810,6 +4810,12 @@ static __always_inline enum event_id_e net_packet_to_net_event(net_packet_t pack
     return MAX_EVENT_ID;
 }
 
+// The address of &neteventctx->eventctx will be aligned as eventctx is the
+// first member of that packed struct. This is a false positive as we do need
+// the neteventctx struct to be all packed.
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Waddress-of-packed-member"
+
 static __always_inline int should_submit_net_event(net_event_context_t *neteventctx,
                                                    net_packet_t packet_type)
 {
@@ -4824,6 +4830,8 @@ static __always_inline int should_submit_net_event(net_event_context_t *netevent
     }
     return 0;
 }
+
+#pragma clang diagnostic pop
 
 static __always_inline int should_capture_net_event(net_event_context_t *neteventctx,
                                                     net_packet_t packet_type)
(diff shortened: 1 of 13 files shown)


pkg/ebpf/c/tracee.bpf.c
```diff
@@ -5008,7 +5015,64 @@ int BPF_KRETPROBE(trace_ret_sock_alloc_file)
...
+    // - it receives a SYN packet and a new socket can be created (accept).
+    // - a sock (socket descriptor) is created for the socket to be created.
+    // - no socket/inode exists yet (sock->sk_socket is NULL).
+    // - accept() traces are too late for initial pkts (socked does not exist).
+    // - by linking old "sock" to the new "sock" we can relate the task.
+    // - some of the initial packets, sometimes with big length, are traced now.
+    //
+    // More at: https://github.com/aquasecurity/tracee/issues/2739
+
+    struct sock *osock = (void *) PT_REGS_PARM1(ctx);
```
roikol, line 5050: i see that `security_sk_clone` is called from few places in the kernel (and not just for the case of TCP accept). are we interested in updating the map for every family/protocol?
if not, you can check here the family/protocol of the socket.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -5118,15 +5187,52 @@ int BPF_KPROBE(cgroup_bpf_run_filter_skb)
...
-        return 0; // e.g. task isn't being traced
+    struct net_task_context *netctx;
+
+    // obtain the task ctx using the obtained socket inode
+
+    if (!mightbecloned) {
+        // pick network context from the inodemap (inode <=> task)
+        netctx = bpf_map_lookup_elem(&inodemap, &inode);
+        if (!netctx)
+            mightbecloned = true; // e.g. task isn't being traced
```
roikol, line 5206: if we get here, is there a reason to continue looking for the `net_task_context`? 
correct me if i'm wrong, but if we had this info saved in the map it would be pointed by the inode we already got, isn't it?

roikol (approved):
looking good!
there is a combination of unrelated changes in this PR IMO, but the primary fix looks solid.
i just left a couple of questions.. approving.

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: This is what I mapped when I was doing this work:

image

I could only find this execution path:

```
sk_set_socket()
  sk_clone_lock() (where security_sk_clone() is called)
    inet_csk_clone_lock()
      tcp_create_openrq_child()
        tcp_v4_sync_recv_sock (also called by tcp_ipv6)
          syn_recv_sock (from ipv4 sock_af_ops)
```

But with a "search" now (instead of a callee function search) I could see it is called by:

- af_alg_accept() (PF_ALG type)
- bluetooth (PF_BLUETOOTH)
- iucv (z/VM comms, PF_IUCV)
- tipc (cluster domain sockets, PF_TIPC)
- vsock (VM / Hypervisor sockets, PF_VSOCK)

So I guess my "kconfig" file (used to index my code) is too small (not including all the other execution paths). So you're right, checking the socket type is likely better (despite those socket types being uncommon or not).

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: Not sure I follow your question :\. Lets see..

1. I check if the "sock" struct has an associated inode with it. For most of the cases, of existing sockets already (on going communication after socket was created) this is the case (no matter if code was called by either task or kernel context).

2. If inode is zero I know it "could be" a clone socket (or something else, sometimes, through observation).

3. IF the inode isn't zero, I try to obtain the task context from a saved inode <=> task relation. That is true for most of the cases actually (for existing/on-going flows from existing sockets). 

So, at (3) if I find the task context from the current inode, I' m good and will continue the existing logic (before this PR). Which is the default path and mechanism.

4. BUT if the inode was ZERO OR the task context could not be found (either because the task isn't being traced OR because this is a socket that was JUST created, by a kernel context, so no tracks exist)... then I need another check: check if I have this "sock" created by a "syn_recv_sock" execution path, or, in other words, check if this "unseen socket" came from the tcp handshake. IF IT CAME, then check if there is a task associated with it (which would be currently a task associated with the socket that created it). IF IT DID NOT COME from that path, then its safe to ignore the new socket (either unrelated OR not being traced).

rafaeldtinoco:
Okay, with a +1 from Roi, with bigger context from the network code, I'll move on and merge this (I want it tested from more people before the release). About the unrelated commits in this PR: Its a minor opportunistic refactor and no logic change.

Issue #2970, opened 2023-04-03
dynamic_code_loading filtered with no filter arguments


NDStrahilevitz:
## Description

1. in `signature_engine.go` under line 66 add the following: `logger.Debugw("shouldn't process from engine", "event", event.EventName, "sig_id", finding.SigMetadata.ID)`
2. run `sudo ./dist/tracee --log file:./bruh.log --log debug`
3. let it run for a while
4. notice that there are filtered events
5. in the log file you should see that the filtered events are `dynamic_code_loading`

## Output of `tracee -v`:

Tracee version "v0.12.0-105-g5dd68dce"

## Output of `uname -a`:

Linux ip-172-31-75-171 5.15.0-1031-aws #35-Ubuntu SMP Fri Feb 10 02:07:18 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

yanivagman:
I think the thing here is that we don't copy the event in `engineEvents() `as we do in `derivedEvents()` https://github.com/aquasecurity/tracee/blob/bc31a5d1aa081b35f11a8ffdc895a145f64835d0/pkg/ebpf/events_pipeline.go#L455 Then the base event matchedPolicies is zeroed and we don't see it in the output

geyslan:
> I think the thing here is that we don't copy the event in `engineEvents() `as we do in `derivedEvents()`
> 
> https://github.com/aquasecurity/tracee/blob/bc31a5d1aa081b35f11a8ffdc895a145f64835d0/pkg/ebpf/events_pipeline.go#L455
> 
> Then the base event matchedPolicies is zeroed and we don't see it in the output

@yanivagman I think what @NDStrahilevitz reported is related to "although" `dynamic_code_loading` is not set, it is being submitted and emitted.

So, I listed all values that is being set in `events_map` via a default `sudo ./dist/tracee`:

```diff
diff --git a/pkg/ebpf/tracee.go b/pkg/ebpf/tracee.go
index f8151b2f..49caf797 100644
--- a/pkg/ebpf/tracee.go
+++ b/pkg/ebpf/tracee.go
@@ -995,6 +995,7 @@ func (t *Tracee) populateBPFMaps() error {
                }
                binary.LittleEndian.PutUint64(eventConfigVal[8:16], paramTypes)
 
[...]

PR #2984, opened 2023-04-06
sig: engine: copy event before engine processing

geyslan:

This is needed because later modification of the event (in particular of the matched policies) can affect engine stage.

### 3. Other comments

Fixes: #2970


yanivagman (approved):
LGTM

josedonizetti (approved):

Issue #2992, opened 2023-04-10
errfmt: prefixFunc only need to prefix at debug level


geyslan:
`prefixFunc` always prefix the msg error with the function name, but it's an information required only for debug level.

https://github.com/aquasecurity/tracee/blob/90f56d07ee7e6bb17acf861997aefe2dc327985c/pkg/errfmt/errfmt.go#L22-L26

rafaeldtinoco:
> but it's an information required only for debug level.

Is it ? What about an error you can't reproduce and user is willing to help by saying "I got this error" and you don't have a clue where the origin is (this is what was happening).

Maybe we can have smaller, but, still, position informative information. Software vendors use error codes usually (Error: XXXXXX (Code: 432432)). Maybe we can have something like that so users, who don't run tracee in debug mode, that can't reproduce an issue at will, are still able to inform us where the error came from.

Just thinking out loud. I think the change of having the error origin is very good to get rid of it (even in non debug mode). For me, debug is for "tracing purposes" as well. Errors should always be informative about what it means and its origins.

geyslan:
Sorry, I didn't make the issue so clear. The intent is to prefix the function name with whatever `logger.{Debug,Error,Warn}w` helper is used, but only when the `LoggerConfig.Level` is Debug. So, in default level - Info, for instance, any log output, even if the errfmt is used, won't prefix anything. But any use of the errfmt helpers will behave like it behaves today only when debug level is set.

See my current implementation (with a false additional logger.Warnw to exemplify):

```shell
tracee on  2992-prefix-debug [$!] via 🐹 v1.19.7 via ⍱ v2.3.4 took 39s 
❯ sudo ./dist/tracee -f comm=uname --log warn 
{"level":"warn","ts":1681144933.57907,"msg":"Enricher","error":"error registering enricher: unsupported runtime crio"}
TIME             UID    COMM             PID     TID     RET              EVENT                     ARGS
```

```shell

tracee on  2992-prefix-debug [$!] via 🐹 v1.19.7 via ⍱ v2.3.4 took 4s 
❯ sudo ./dist/tracee -f comm=uname --log debug
Place your finger on the fingerprint reader
[...]

PR #2994, opened 2023-04-10
logger: errfmt: flags: prefix debug level only

geyslan:

    logger: errfmt: flags: prefix debug level only
    
    Make errfmt prefix only when debug level is set. This is to avoid
    noise in the output when debug level is not set and to facilitate
    unit testing, where there's a comparison between the output and
    the expected output.

### 3. Other comments

Fixes: #2992

### DIFF
--- a/pkg/cmd/flags/flags_test.go
+++ b/pkg/cmd/flags/flags_test.go
@@ -896,22 +896,22 @@ func TestPrepareOutput(t *testing.T) {
 		{
 			testName:      "empty option flag",
 			outputSlice:   []string{"option"},
-			expectedError: errors.New("parseOption: option flag can't be empty, use '--output help' for more info"),
+			expectedError: errors.New("option flag can't be empty, use '--output help' for more info"),
 		},
 		{
 			testName:      "empty option flag 2",
 			outputSlice:   []string{"option:"},
-			expectedError: errors.New("parseOption: option flag can't be empty, use '--output help' for more info"),
+			expectedError: errors.New("option flag can't be empty, use '--output help' for more info"),
 		},
 		{
 			testName:      "invalid option value",
 			outputSlice:   []string{"option:foo"},
-			expectedError: errors.New("setOption: invalid output option: foo, use '--output help' for more info"),
+			expectedError: errors.New("invalid output option: foo, use '--output help' for more info"),
 		},
 		{
 			testName:      "empty file for format",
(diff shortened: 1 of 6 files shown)


rafaeldtinoco:
I'm against this change without an error code in the default error level that tells us where the error came from (like being discussed in the issue).

geyslan:
> I'm against this change without an error code in the default error level that tells us where the error came from (like being discussed in the issue).

As this was requested in today's 1:1 meeting, just pinging everyone that relates to the topic as well. @NDStrahilevitz @yanivagman @josedonizetti

yanivagman:
The idea is that the average user (who doesn't run in debug level) should not see the full verbose error message that also includes all the stack tracee and everything.
I'm ok with adding some indication of where the error came from though.

geyslan:
> The idea is that the average user (who doesn't run in debug level) should not see the full verbose error message that also includes all the stack trace and everything.

The current main Debugw outputs (which prints the stack trace) only if logger Level is Debug, so that should not be a noise problem.

> I'm ok with adding some indication of where the error came from though.

This is how the current main code outputs logs that aren't Debugw (sample warn):

```json
{"level":"warn","ts":1681216908.6991243,"msg":"Enricher","error":"containers.(*runtimeInfoService).Register: error registering enricher: unsupported runtime podman"}
```

It has an error with a prefixed origin `containers.(*runtimeInfoService).Register:`, which depends on the use of the `errfmt` helpers to achieve that format.

On the other hand, Debugw outputs the origin as part of the structured log:

```json
{"L":"DEBUG","T":"2023-04-11T09:45:43.031-0300","M":"Enricher","error":"containers.(*runtimeInfoService).Register: error registering enricher: unsupported runtime podman","origin":"containers:pkg/containers/containers.go:82","calls":"New() < (*Tracee).Init() < Runner.Run() < main.func1() < (*App).RunContext() < (*App).Run() < main()"}
```

In this case, we have the prefixed error (the same formatted error), the origin field with pkg:file:line and the calls field (stack trace).

Said that, we already have a way of pinpoint the log origin in both cases, what, perhaps, makes this PR worthless.

yanivagman:
> Said that, we already have a way of pinpoint the log origin in both cases, what, perhaps, makes this PR worthless.

Yep. If that's the case, I don't see a point to change it

geyslan:
As discussed, not planned.

PR #2937, opened 2023-03-30
feat: support unmarshalling trace types

roikol:
this change will make unmarshalling for signatures much straight forward

### 1. Explain what the PR does

```
Author: RoiKol <roi.kol@aquasec.com>
Date:   Thu Mar 30 18:19:20 2023 +0300

    feat: support unmarshalling trace types
    
    unmarshal trace structs to the correct type
```

PR #2996, opened 2023-04-13
feat: unmarshall trace types correctly

roikol:
the idea is to unmarshal types from the `trace` package to the correct type, instead of `interface` (which is the case now).
it'll make writing signatures easier as there will be no need to parse types in the signature itself (or in the helpers).

in practice, i moved the parsing logic from `helpers.go` (https://github.com/aquasecurity/tracee/blob/main/signatures/helpers/helpers.go) to the `UnmarshalJSON()` func in the `trace` package.

in the future, after this PR merges, any new types in the `trace` package should be parsed there as well.

this PR has the same purpose as #2937 which didn't solve the problem.

```
Date:   Sun Apr 2 17:50:23 2023 +0300

    feat: unmarshall trace types correctly
    
    unmarshal trace types to their real types instead of interface
```

### DIFF
--- a/types/trace/trace.go
+++ b/types/trace/trace.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


rafaeldtinoco:
@roikol I'll review this and unblock you today. Sorry for the delay.

rafaeldtinoco (approved):
LGTM. As I see, at the end, simply unmarshalling to the type wasn't enough and you ended up doing the parsing in the correct place (JSON unmarshalling function). Nice change!

PR #2999, opened 2023-04-14
Enable a Debug Shell for Test Runners

rafaeldtinoco:
commit 8913fc09 (HEAD -> debug-shell, rafaeldtinoco/debug-shell)
Date:   Tue Apr 18 10:15:10 2023

    workflow: introduce distro image debug shell
    
    This commit introduces a workflow that makes possible triggering a VM
    debug shell, to be ssh'ed to, that allows a developer to debug
    development issues happening on specific kernel versions. With that,
    the developer has the same environment used on PR tests, provisioned
    so they can fix issues more easily.
    
    The workflow shall be triggered manually whenever a debug shell is
    needed, and the provisioned VM should be shutdown at the end of its
    usage. Provisioning VMs require repository write access so, engineers
    that have only triage permissions should ask maintainers for debug
    shells.

### DIFF
--- a/.github/workflows/debug-shell.yaml
+++ b/.github/workflows/debug-shell.yaml
@@ -1,0 +1,262 @@
+name: Debug Distro Shell
+on:
+  workflow_dispatch:
+    inputs:
+      distro:
+        type: choice
+        description: Distro Image to Debug
+        required: true
+        options:
+          - alma418
+          - alma418-arm64
+          - amzn2510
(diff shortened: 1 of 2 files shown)


rafaeldtinoco:
Tested using https://github.com/rafaeldtinoco/tracee/actions/runs/4733040111. Needs merge to be tested on MAIN. Merging it (since its chore) so I can test and use it.

Issue #2900, opened 2023-03-22
Global configuration from file


yanivagman:
With policies in place, events and filters will be configured through policy files.
Yet there are some configuration which are global, and we should allow the user to set them through a file.

- [ ] Extract config package from cmd and tracee packages
- [x] Add a configuration file that can be used to set global options, such as install path, perf buffer size, caching, etc.

Previous attempt was made to create a configuration file, and can be used as a reference: 
#1324 (issue #467)

Issue #1264, opened 2021-12-15
extract package `config` from `tracee`


itaysk:
from https://github.com/aquasecurity/tracee/issues/960:

> configurations - create package `config` that includes filter, captureConfig ,outputConfig and related functions. (also create an interface for config and implement validators)

yanivagman:
The following consts can also be moved into this package:
```
type bpfConfig uint32

const (
	configDetectOrigSyscall bpfConfig = iota + 1
	configExecEnv
	configCaptureFiles
	configExtractDynCode
	configTraceePid
	configStackAddresses
	configUIDFilter
	configMntNsFilter
	configPidNsFilter
	configUTSNsFilter
	configCommFilter
	configPidFilter
	configContFilter
	configFollowFilter
	configNewPidFilter
	configNewContFilter
	configDebugNet
	configProcTreeFilter
	configCaptureModules
	configCgroupV1
)
```

yanivagman:
> The following consts can also be moved into this package:
> 
> ```
> type bpfConfig uint32
> 
> const (
> 	configDetectOrigSyscall bpfConfig = iota + 1
> 	configExecEnv
> 	configCaptureFiles
> 	configExtractDynCode
> 	configTraceePid
> 	configStackAddresses
> 	configUIDFilter
> 	configMntNsFilter
> 	configPidNsFilter
> 	configUTSNsFilter
> 	configCommFilter
> 	configPidFilter
> 	configContFilter
> 	configFollowFilter
> 	configNewPidFilter
> 	configNewContFilter
> 	configDebugNet
> 	configProcTreeFilter
> 	configCaptureModules
> 	configCgroupV1
> )
> ```

In a second thought, there are some consts here that are more related to filters and other stuff not directly related to what we might want to have in the `config` package

PR #3000, opened 2023-04-15
tracee one binary cli migration to cobra/viper

geyslan:

Migrates tracee (new binary) cli to cobra - tracee-ebpf continues using urfave -, making it possible to load global config via viper.

That is a **breaking change**, as `--list` and `--version` have been promoted to command status `list` `l` and `version` `v`. The root `--help` has been kept but now we have a `help` for commands. To print the help of a flag, we can use the --help or -h flag followed
by the flag name, e.g. --help filter.

`--capabilities` shorthand has been changed from `--caps` to `-C` since cobra doesn't allow more than one character for that.

```
❯ sudo ./dist/tracee help         
Tracee uses eBPF technology to tap into your system and give you
access to hundreds of events that help you understand how your system behaves.

Usage:
  tracee [flags]
  tracee [command]

Available Commands:
  completion  Generate the autocompletion script for the specified shell
  help        Help about any command
  list        List traceable events
  version     Print the version number of Tracee

Flags:
      --blob-perf-buffer-size int   Size, in pages, of the internal perf ring buffer used to send blobs from the kernel (default 1024)
  -a, --cache strings               Control event caching queues (default [none])
  -C, --capabilities strings        Define capabilities for tracee to run with
  -c, --capture strings             Capture artifacts that were written, executed or found to be suspicious
      --config string               Global config file (yaml, json between others - see documentation)
      --containers                  Enable container info enrichment to events. This feature is experimental and may cause unexpected behavior in the pipeline
      --crs strings                 Define connected container runtimes
  -f, --filter strings              Select events to trace by defining filter expressions
      --healthz                     Enable healthz endpoint
  -h, --help                        help for tracee
      --install-path string         Path where tracee will install or lookup it's resources (default "/tmp/tracee")
      --listen-addr string          Listening address of the metrics endpoint server (default ":3366")
  -l, --log strings                 Logger options (default [info])
      --metrics                     Enable metrics endpoint
  -o, --output strings              Control how and where output is printed (default [table])
  -b, --perf-buffer-size int        Size, in pages, of the internal perf ring buffer used to submit events from the kernel (default 1024)
  -p, --policy strings              Path to a policy or directory with policies
      --pprof                       Enable pprof endpoints
      --pyroscope                   Enable pyroscope agent
      --rego strings                Control event rego settings
      --signatures-dir string       Directory where to search for signatures in CEL (.yaml), OPA (.rego), and Go plugin (.so) formats

Use "tracee [command] --help" for more information about a command.
```

```
❯ ./dist/tracee --help config # -h config
The --config flag allows you to define global configuration options (flags)
for tracee, by providing a file in YAML or JSON format between others (see documentation).

All flags can be set in the config file, except for the following, which are reserved only
for the CLI:
  --config (this flag)
  --capture
  --policy

The --filter flag also cannot be set in the config file since it's reserved for the CLI
and the policy file loading mechanism (via the --policy flag).
```

---

It also introduces `--config` flag for loading global configuration (yaml). It ignores policy, filter, config and capture flags.

### 2. Explain how to test it

Run tracee (one binary) with your usual flags considering changes explained in section [one](#1.-Explain-what-the-PR-does).

Run tracee (one binary) with `--config` flag using `./examples/global_config.yaml`.

`sudo ./dist/tracee --config ./examples/global_config.yaml --filter comm=uname,comm=who`

### 3. Other comments

Fixes: #2900


geyslan:
@rafaeldtinoco and @NDStrahilevitz, JFYI.

examples/global_config.yaml
```diff
@@ -0,0 +1,19 @@
+blob-perf-buffer-size: 1024
```
NDStrahilevitz, line 1: Let's add a json example as well

cmd/tracee/cmd/help_flags.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/spf13/cobra"
+
+	"github.com/aquasecurity/tracee/pkg/cmd/flags"
+)
+
+func init() {
+	rootCmd.AddCommand(helpFlagsCmd)
+}
+
+var helpFlagsCmd = &cobra.Command{
```
NDStrahilevitz, line 16: This is currently not working for me with `tracee help filter`, or any of the options listed in `flags/help.go`.

NDStrahilevitz (commented):
I have not managed to successfully:
1. Run the help flags
2. Run tracee with a config file (it was ignored when set through the flag)

Not -1ing due to the possibility something went wrong in my process of testing but currently this doesn't seem to work.

cmd/tracee/cmd/help_flags.go
geyslan: Ok. `tracee help` only work for commands. Cobra has it as a built-in command help.

Running it we can see:

```
... 
Available Commands:
  completion  Generate the autocompletion script for the specified shell
  help        Help about any command
  help-flags  Help about any flags
  list        List traceable events
  version     Print the version number of Tracee
...
```

So it works for list as list is a command:

```
./dist/tracee help list
List traceable events

Usage:
  tracee list

Aliases:
  list, l

Flags:
  -h, --help   help for list
```

Cobra distinguishes between command, sub-commands and flags. `--filter` in our case is a flag, so I created the help-flags to output help for them.

Then help is for commands and help-flags is for the flags.

`tracee help`

```
...
Flags:
      --blob-perf-buffer-size int   Size, in pages, of the internal perf ring buffer used to send blobs from the kernel (default 1024)
  -a, --cache strings               Control event caching queues (default [none])
  -C, --capabilities strings        Define capabilities for tracee to run with
  -c, --capture strings             Capture artifacts that were written, executed or found to be suspicious
      --config string               Global config file (yaml)
      --containers                  Enable container info enrichment to events. This feature is experimental and may cause unexpected behavior in the pipeline
      --crs strings                 Define connected container runtimes
  -f, --filter strings              Select events to trace by defining filter expressions
      --healthz                     Enable healthz endpoint
  -h, --help                        help for tracee
      --install-path string         Path where tracee will install or lookup it's resources (default "/tmp/tracee")
      --listen-addr string          Listening address of the metrics endpoint server (default ":3366")
  -l, --log strings                 Logger options (default [info])
      --metrics                     Enable metrics endpoint
  -o, --output strings              Control how and where output is printed (default [table])
  -b, --perf-buffer-size int        Size, in pages, of the internal perf ring buffer used to submit events from the kernel (default 1024)
  -p, --policy strings              Path to a policy or directory with policies
      --pprof                       Enable pprof endpoints
      --pyroscope                   Enable pyroscope agent
      --rego strings                Control event rego settings
      --signatures-dir string       Directory where to search for signatures in CEL (.yaml), OPA (.rego), and Go plugin (.so) formats
```

And `tracee help-flags log` (log is a flag) would do the magic:

```
❯ ./dist/tracee help-flags log   
Control logger options - aggregation and level priority.

Possible options:
  --log aggregate[:interval]          | turns log aggregation on, delaying output optional interval (s, m) (default is off)
  --log <debug|info|warn|error|panic> | set log level, info is the default
  --log file:/path/to/file            | write the logs to a specified file. create/trim the file if exists (default: stderr)

Examples:
  --log debug                              | outputs debug level logs
  --log debug --log aggregate              | outputs aggregated debug level logs every 3 seconds (default)
  --log aggregate:5s                       | outputs aggregated logs every 5 seconds
  --log debug --log file:/tmp/tracee.log   | outputs debug level logs to /tmp/tracee.log
```

geyslan: > Run tracee with a config file (it was ignored when set through the flag)

A flag override the respective value set in the config file.

If a config file has log set to debug and I run `sudo ./dist/tracee --config cfg.yaml --log warn`, all cfg.yaml flags are set but log, which is set to warn.

examples/global_config.yaml
geyslan: Done.

cmd/tracee/cmd/help_flags.go
NDStrahilevitz: Ok, so what failed for me in the config was the filter field which I set as such:
```
filter:
  - e=ptrace
  - comm=strace
```
Perhaps I set it wrong.

About `help-flags`, i've missed that, thanks for pointing out. Is it impossible to extend the help command for this? It would be more intuitive for use.
Furthermore trying the "old school way" goes like this right now
```
sudo ./dist/tracee --filter help
{"level":"fatal","ts":1681996946.1647704,"msg":"Execution","error":"invalid filter option specified (help), use '--filter help' for more info"}
```

geyslan: > Ok, so what failed for me in the config was the filter field which I set as such:
> 
> ```
> filter:
>   - e=ptrace
>   - comm=strace
> ```
> 
> Perhaps I set it wrong.

It happens that filter, policy and capture cobra flags aren't bind to viper config file. It was a design option. For example, policy has its own logic for file loading and filter use must reside only via CLI.
 
> About `help-flags`, i've missed that, thanks for pointing out. Is it impossible to extend the help command for this? It would be more intuitive for use.

Initially I haven't succeeded in expand the built-in help command to accommodate both command and flags (possibly flag and command conflict or premature parsing of an arg with `--`), there's the reason for a new command to parse only flags without `--`. 

> Furthermore trying the "old school way" goes like this right now
> 
> ```
> sudo ./dist/tracee --filter help
> {"level":"fatal","ts":1681996946.1647704,"msg":"Execution","error":"invalid filter option specified (help), use '--filter help' for more info"}
> ```

Thanks for pointing out this. I'm going to fix the messages for the new binary.

pkg/cmd/flags/config.go
```diff
@@ -0,0 +1,14 @@
+package flags
+
+func configHelp() string {
+	return `The --config flag allows you to define global configuration options (flags)
+for tracee, by providing a file in YAML or JSON format or other viper supported formats.
+
+All flags can be set in the config file, except for the following, which are reserved
+for the CLI:
```
yanivagman, line 8: --filter (and in the future --capture as well) can also be set through policies

yanivagman (changes requested):
Looks good @geyslan,
The matching issue also lists a sub-issue for extracting "config" package. Don't you think this is still required? If so, we better open a new issue to track that

cmd/tracee/cmd/root.go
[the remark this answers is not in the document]
geyslan: It's kind of a [convention](https://github.com/spf13/cobra/blob/main/user_guide.md#create-rootcmd) to name it root using cobra. This is where we have the logic of root commands and flags - mainly since we mostly use flags instead of commands. Other commands are in separate file as `cmd/tracee/cmd/help_flag.go`, `cmd/tracee/cmd/list.go` and `cmd/tracee/cmd/version.go`.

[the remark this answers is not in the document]
geyslan: But we can change it to tracee or other name. LMK.

cmd/tracee/cmd/root.go
```diff
@@ -0,0 +1,285 @@
...
+	cobra.OnInitialize(initConfig)
+
+	// Flags
+
+	// config is not bound to viper
+	rootCmd.Flags().StringVar(
+		&cfgFile,
+		"config",
+		"",
+		"Global config file (yaml, json between others supported by viper)",
```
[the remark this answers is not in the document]
geyslan: Viper supports these file types:

image

I opted to mention "supported by viper" to summarize.

Perhaps, `"Global config file (yaml, json between others - check documentation)",`, WDYT?

cmd/tracee/cmd/root.go
```diff
@@ -0,0 +1,285 @@
...
+func initCmd() error {
+	rootCmd.SetOut(os.Stdout)
+	rootCmd.SetErr(os.Stderr)
+
+	cobra.OnInitialize(initConfig)
+
+	// Flags
+
+	// config is not bound to viper
+	rootCmd.Flags().StringVar(
```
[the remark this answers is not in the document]
geyslan: `list`, `version` and `help-flag` are commands so they have their own file under `cmd/tracee/cmd/`.

I'm going to reorder this way following your suggestion:

```
        // Filter/Policy flags

	"filter"
	"policy"

	// Output flags

	"output"
	"capture"

	// Config flag

	"config"

	// Container flags

	"containers"
	"crs"

	// Signature flags

	"signatures-dir"
	"rego"

	// Buffer/Cache flags

	"perf-buffer-size"
	"blob-perf-buffer-size"
	"cache"

	// Server flags

	server.MetricsEndpointFlag
	server.HealthzEndpointFlag
	server.PProfEndpointFlag
	server.PyroscopeAgentFlag
	server.ListenEndpointFlag

	// Other flags

	"capabilities",
	"install-path",
	"log"
```

geyslan:
> The matching issue also lists a sub-issue for extracting "config" package. Don't you think this is still required? If so, we better open a new issue to track that

@yanivagman it became not so crystal clear for me. Much more after bringing up config loading through viper. For example:

```go
	rootCmd.Flags().StringSliceP(
		"log",
		"l",
		[]string{"info"},
		"Logger options",
	)
	err = viper.BindPFlag("log", rootCmd.Flags().Lookup("log"))
	if err != nil {
		return errfmt.WrapError(err)
	}
```

That code has cobra flag (for cli) and viper bind (for file loading). If we split them into different packages or files I suppose it will be more difficult to track them down.

Anyway, if the goal is to encapsulate third-party library logic in our own config package with our own API, then yes, I agree to open a new issue for that.

cmd/tracee/cmd/root.go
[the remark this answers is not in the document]
yanivagman: I see. I'm fine with keeping it root then

cmd/tracee/cmd/root.go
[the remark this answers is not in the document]
yanivagman: Will it be listed in our documentation? Because I don't remember seeing it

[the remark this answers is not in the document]
geyslan: Can be. The documentation change will be tackled after these breaking changes approval.

yanivagman:
> @yanivagman it became not so crystal clear for me. Much more after bringing up config loading through viper. For example:

So actually what I had in mind is to extract all tracee config structs into their own package, unrelated to the CLI interface. Similar to what we had in this issue https://github.com/aquasecurity/tracee/issues/1264 minus the filter related stuff which has its own package now

geyslan:
> > @yanivagman it became not so crystal clear for me. Much more after bringing up config loading through viper. For example:
> 
> So actually what I had in mind is to extract all tracee config structs into their own package, unrelated to the CLI interface. Similar to what we had in this issue #1264 minus the filter related stuff which has its own package now

I got it. Since we have tracee-ebpf and tracee-rule using urfave, let's open an issue to resolve config extraction after their drop.

geyslan:
Once this is approved, I will work on the documentation changes.

rafaeldtinoco:
Minor rebase needs because of "revive" PR. Its good that a rebase will have a "revive" pass as well. Thanks for reviewing that PR Geyslan. I believe the branch is only missing this PR now for the release (And fixes if needed).

Cheers!

cmd/tracee/cmd/help_flags.go
geyslan: > About help-flags, i've missed that, thanks for pointing out. Is it impossible to extend the help command for this? It would be more intuitive for use.
Furthermore trying the "old school way" goes like this right now

Check https://github.com/aquasecurity/tracee/pull/3000#discussion_r1182805909

cmd/tracee/cmd/list.go
```diff
@@ -0,0 +1,22 @@
...
+	rootCmd.AddCommand(listCmd)
+}
+
+var listCmd = &cobra.Command{
+	Use:     "list",
+	Aliases: []string{"l"},
+	Short:   "List traceable events",
+	Long:    ``,
+	Run: func(cmd *cobra.Command, args []string) {
+		tcmd.PrintEventList(true) // list events
```
josedonizetti, line 19: This is a problem for signature events. They need to be loaded before listing or else we will always see an empty list. 

There was an idea to make the default signature events part of tracee instead of a plugin, which would fix this for tracee signature events, for users' events we still need to ask a path and load the events before.

geyslan: Thanks @josedonizetti. fixed (workaround) via c55bc45d71522ee267d86cb55fde3848a67bb87d.

cmd/tracee/cmd/list.go
```diff
@@ -0,0 +1,43 @@
...
+	Use:     "list",
+	Aliases: []string{"l"},
+	Short:   "List traceable events",
+	Long:    ``,
+	Run: func(cmd *cobra.Command, args []string) {
+		// Get signatures to update event list
+		sigs, err := signature.Find(
+			compile.TargetRego,
+			false,
+			"",
```
yanivagman, line 30: Not a must, but we can potentially add `viper.GetString("signatures-dir")` here by adding the `signatures-dir` flag

yanivagman (approved):
LGTM, thanks @geyslan 
Just a reminder to open a separate issue for extracting the config package.

cmd/tracee/cmd/list.go
geyslan: Done. Inserted the same `signaturs-dir` root flag to the list command.

geyslan:
402c36e2c95c18b8efa43a2cea7d64f077f6caac

image

josedonizetti (approved):

PR #3001, opened 2023-04-16
events: improve hidden_kernel_module

OriGlassman:
Add improvements:
- Run after a kernel module was loaded (in addition to periodically)
- Address possible race condition that my cause false-positive

### 2. Explain how to test it

./tracee -f e=hidden_kernel_module

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -583,21 +583,54 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
 // number of iterations - value that the verifier was seen to cope with - the higher, the better
 #define MAX_NUM_MODULES 600
 
-// This map is only used from kernels 5.2 and above.
-// If we won't create the map, the userspace will receive an error when interacting with it,
-// and since the userspace is not kernel version aware, it won't know whether it's because
-// the map simply wasn't created due to kernel version, or something unexpected failed.
-BPF_HASH(modules_map, u64, kernel_module_t, MAX_NUM_MODULES);
+enum
+{
+    PROC_MODULES = 1 << 0,
+    KSET = 1 << 1,
+    MOD_TREE = 1 << 2,
+    NEW_MOD = 1 << 3,
+    FULL_SCAN = 1 << 30,
+    HIDDEN_MODULE = 1 << 31,
+};
 
-void __always_inline send_hidden_module(u64 mod_addr, char *mod_name, program_data_t *p)
+// Forcibly create the map in all kernels, even when not needed, due to lack of
+// support for kernel version awareness about map loading errors.
+
+BPF_HASH(modules_map, u64, kernel_module_t, MAX_NUM_MODULES);
+BPF_HASH(new_module_map, u64, kernel_new_mod_t, MAX_NUM_MODULES);
+
+// We only care for modules that got deleted or inserted between our scan and if
+// we detected something suspicious. Since it's a very small time frame, it's
+// not likely that a large amount of modules will be deleted. Instead of saving
+// a map of deleted modules, we could have saved the last deleted module
+// timestamp and, if we detected something suspicious, verify that no modules
+// got deleted between our check. This is preferable space-wise (u64 instead of
+// a map), but an attacker might start unloading modules in the background and
+// race with the check in order to abort reporting for hidden modules.
+
+BPF_LRU_HASH(recent_deleted_module_map, u64, kernel_deleted_mod_t, 50);
+BPF_LRU_HASH(recent_inserted_module_map,
+             u64,
+             kernel_new_mod_t,
+             50); // Likewise for module insertion
+
+void __always_inline lkm_seeker_send_to_userspace(struct module *mod, u32 *flags, program_data_t *p)
 {
     reset_event_args(p);
+    u64 mod_addr = (u64) mod;
+    char *mod_name = mod->name;
+    const char *mod_srcversion = BPF_CORE_READ(mod, srcversion);
+
     save_to_submit_buf(p->event, &mod_addr, sizeof(u64), 0);
     save_bytes_to_buf(p->event,
                       (void *) mod_name,
                       MODULE_NAME_LEN & MAX_MEM_DUMP_SIZE,
-                      1); // this is actually a string, the argument is saved as bytes since the
-                          // verifier didn't like it as str
+                      1); // string saved as bytes (verifier issues).
+    save_to_submit_buf(p->event, flags, sizeof(u32), 2);
+    save_bytes_to_buf(p->event,
+                      (void *) mod_srcversion,
+                      MODULE_SRCVERSION_MAX_LENGTH & MAX_MEM_DUMP_SIZE,
+                      3); // string saved as bytes (verifier issues).
 
     events_perf_submit(p, HIDDEN_KERNEL_MODULE_SEEKER, 0);
 }
@@ -638,17 +671,37 @@ statfunc bool init_shown_modules()
         }
     }
 
-    return !iterated_all_modules; // false is valid value
+    return !iterated_all_modules; // False is valid value
 }
+
+static u64 start_scan_time_init_shown_mods = 0;
+static u64 last_module_insert_time = 0;
 
 statfunc bool is_hidden(u64 mod)
 {
     kernel_module_t *mod_from_map = bpf_map_lookup_elem(&modules_map, &mod);
-    if (mod_from_map == NULL) {
-        return true; // if we don't find the module in our map, it means the module is hidden
-    } else {
-        return !mod_from_map->seen_modules_list; // if it wasn't seen in modules list, it's hidden
+    if (mod_from_map != NULL && mod_from_map->seen_modules_list) {
+        return false;
     }
+
+    // Verify that this module wasn't removed after we initialized modules_map
+    kernel_deleted_mod_t *deleted_mod = bpf_map_lookup_elem(&recent_deleted_module_map, &mod);
+    if (deleted_mod && deleted_mod->deleted_time > start_scan_time_init_shown_mods) {
+        // This module got deleted after the start of the scan time.. So there
+        // was a valid remove, and it's not hidden.
+        return false;
+    }
+
+    // Verify that this module wasn't inserted after we initialized modules_map
+    kernel_new_mod_t *recent_insert_mod = bpf_map_lookup_elem(&recent_inserted_module_map, &mod);
+    if (!recent_insert_mod ||
+        (recent_insert_mod->insert_time != 0 && // If insert_time is 0, it means the module is
+                                                // currently being inserted - it's not hidden.
+         recent_insert_mod->insert_time <= start_scan_time_init_shown_mods)) {
+        return true;
+    }
+
+    return false;
 }
 
 statfunc bool find_modules_from_module_kset_list(program_data_t *p)
@@ -659,6 +712,7 @@ statfunc bool find_modules_from_module_kset_list(program_data_t *p)
     struct kobject *pos = list_first_entry_ebpf(head, typeof(*pos), entry);
     struct kobject *n = list_next_entry_ebpf(pos, entry);
     bool finished_iterating = false;
+    u32 flags = KSET | HIDDEN_MODULE;
 
 #pragma unroll
     for (int i = 0; i < MAX_NUM_MODULES; i++) {
@@ -675,7 +729,7 @@ statfunc bool find_modules_from_module_kset_list(program_data_t *p)
             struct module *mod = BPF_CORE_READ(mod_kobj, mod);
             if (mod) {
                 if (is_hidden((u64) mod)) {
-                    send_hidden_module((u64) mod, mod->name, p);
+                    lkm_seeker_send_to_userspace(mod, &flags, p);
                 }
             }
         }
@@ -687,7 +741,7 @@ statfunc bool find_modules_from_module_kset_list(program_data_t *p)
     return !finished_iterating;
 }
 
-BPF_QUEUE(walk_mod_tree_queue, rb_node_t, 2048); // used to walk a rb tree
+BPF_QUEUE(walk_mod_tree_queue, rb_node_t, MAX_NUM_MODULES); // used to walk a rb tree
 
 statfunc struct latch_tree_node *__lt_from_rb(struct rb_node *node, int idx)
 {
@@ -700,6 +754,7 @@ statfunc bool walk_mod_tree(program_data_t *p, struct rb_node *root, int idx)
     struct module *mod;
     bool finished_iterating = false;
     struct rb_node *curr = root;
+    u32 flags = MOD_TREE | HIDDEN_MODULE;
 
 #pragma unroll
     for (int i = 0; i < MAX_NUM_MODULES; i++) {
@@ -719,7 +774,7 @@ statfunc bool walk_mod_tree(program_data_t *p, struct rb_node *root, int idx)
                 mod = BPF_CORE_READ(container_of(ltn, struct mod_tree_node, node), mod);
 
                 if (is_hidden((u64) mod)) {
-                    send_hidden_module((u64) mod, mod->name, p);
+                    lkm_seeker_send_to_userspace(mod, &flags, p);
                 }
 
                 /* We have visited the node and its left subtree.
@@ -753,6 +808,36 @@ statfunc bool find_modules_from_mod_tree(program_data_t *p)
     return walk_mod_tree(p, node, seq & 1);
 }
 
+static __always_inline u64 check_new_mods_only(program_data_t *p)
+{
+    struct module *pos, *n;
+    u64 start_scan_time = bpf_ktime_get_ns();
+    char modules_sym[8] = "modules";
+    kernel_new_mod_t *new_mod;
+    u64 mod_addr;
+    struct list_head *head = (struct list_head *) get_symbol_addr(modules_sym);
+
+    pos = list_first_entry_ebpf(head, typeof(*pos), list);
+    n = pos;
+
+#pragma unroll
+    for (int i = 0; i < MAX_NUM_MODULES; i++) {
+        pos = n;
+        n = list_next_entry_ebpf(n, list);
+        if (&pos->list == head) {
+            return start_scan_time; // To be used in userspace
+        }
+
+        mod_addr = (u64) pos;
+        new_mod = bpf_map_lookup_elem(&new_module_map, &mod_addr);
+        if (new_mod) {
+            new_mod->last_seen_time = bpf_ktime_get_ns();
+        }
+    }
+
+    return 0;
+}
+
 statfunc bool check_is_proc_modules_hooked(program_data_t *p)
 {
     struct module *pos, *n;
@@ -760,6 +845,7 @@ statfunc bool check_is_proc_modules_hooked(program_data_t *p)
     char modules_sym[8] = "modules";
     kernel_module_t *mod_from_map;
     struct list_head *head = (struct list_head *) get_symbol_addr(modules_sym);
+    u32 flags = PROC_MODULES | HIDDEN_MODULE;
 
     pos = list_first_entry_ebpf(head, typeof(*pos), list);
     n = pos;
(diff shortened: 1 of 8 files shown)


pkg/events/derive/hidden_kernel_module.go
```diff
@@ -25,36 +26,83 @@ import (
 
 var (
 	foundHiddenKernModsCache *lru.Cache
+	wakeupChannel            = make(chan ScanRequest)
```
NDStrahilevitz, line 29: This introduces pretty severe coupling between the big tracee object and the derive module.
Any problem with the channel now like being full and somehow missing the read from it, would cause a lock in the channel and potentially a lock in tracee as a whole.
However I see why this happened, the logic of scanning would, in a regular ebpf event, likely be handled in the `processEvents` part of the pipeline.
I think this shows we need another post-derive process stage in the pipeline to accommodate derived event processing.

pkg/events/derive/hidden_kernel_module.go
```diff
@@ -25,36 +26,83 @@ import (
...
+			return nil, err
+		}
+		srcversion := string(srcversionBytes[:bytes.IndexByte(srcversionBytes[:], 0)]) // Remove the trailing terminating characters
+
+		flags, err := parse.ArgVal[uint32](&event, "flags")
+		if err != nil {
+			return nil, err
 		}
 
+		if flags&procModules != 0 {
```
NDStrahilevitz, line 78:
```suggestion
		if flags&procModules {
```

NDStrahilevitz: ditto for the next `if`s.

pkg/events/derive/hidden_kernel_module.go
```diff
@@ -81,13 +129,16 @@ func clearMap(bpfMap *bpf.BPFMap) error {
 	return nil
 }
 
+func GetWakeupChannelRead() <-chan ScanRequest {
```
NDStrahilevitz, line 132: See, now anything can read from this module's messaging. So another module calls this and starts extracting messaging and now there's a conflict between it and `tracee.go`.
There's likely a way to avoid cross module messaging here.

NDStrahilevitz (dismissed):
The design of channel messaging between `derive` and `tracee.go` causes direct coupling between the implementation of the derived event and tracee itself.
As in: If we wanted to separate the event out into a plugin for example, it would be difficult, if not impossible to do so.
As i've written in the first comment, I think most of the logic here should move to a new pipeline stage happening after the derive stage.
Alternatively: maybe a separate `modscan` package should be added with parallel module scanning logic that can be initialized by the tracee object (although this will require some future development to how tracee registers such objects).

pkg/events/derive/hidden_kernel_module.go
OriGlassman: "Any problem with the channel now like being full and somehow missing the read from it, would cause a lock in the channel and potentially a lock in tracee as a whole."

The receiving end runs in a goroutine so can't block tracee. I could publish to the channel in a goroutine as well, if the thread that runs the derived event code is the same one who does all the other stuff.

NDStrahilevitz: It's also in its own pipeline thread. But the problem of extendability remains (the code can't be separated out from tracee as a whole).

pkg/events/derive/hidden_kernel_module.go
OriGlassman: I don't see any reason why it would be called elsewhere (I'm guessing the design is what you're worried about), but I could move everything to hidden_kernel_module.go and make the channel "private". Assuming I can move the uprobed function (triggerKernelModuleSeeker) from tracee.go to my derived event, my only problem is the kernelSymbols which I need to read periodically, that is under Tracee object.

pkg/events/derive/hidden_kernel_module.go
OriGlassman: Without it the compiler yells:
pkg/events/derive/hidden_kernel_module.go:78:6: non-boolean condition in if statement

rafaeldtinoco:
I also want to review this one. One thing I spotted in code already merged:

```go
func clearMap(bpfMap *bpf.BPFMap) error {
	var err error

	var iter = bpfMap.Iterator()
	for iter.Next() {
		addr := binary.LittleEndian.Uint64(iter.Key())
		err = bpfMap.DeleteKey(unsafe.Pointer(&addr))

		if err != nil {
			logger.Errorw("err occurred DeleteKey: " + err.Error())
			return err
		}
	}

	return nil
}
```

There is a batch libbpf function for that, way more performatic I believe. Check:

https://github.com/aquasecurity/libbpfgo/blob/main/selftest/map-batch/main.go/#L112-L118

rafaeldtinoco:
Algo, you forgot to raise capabilities at:

https://github.com/aquasecurity/tracee/blob/main/pkg/ebpf/tracee.go/#L1455-L1459

```go
	modsMap, err := t.bpfModule.GetMap("modules_map")
	if err != nil {
		logger.Errorw("err occurred GetMap: " + err.Error())
		return
	}
```

So, now that E2E tests are dropping capabilities by default, they're failing with:

```
[2023-04-21T03:47:51.930Z] {"level":"error","ts":1682048788.2915356,"msg":"Ignored key without a value.","ignored":"failed to update map modules_map: operation not permitted"}
```

pkg/events/derive/hidden_kernel_module.go
```diff
@@ -25,69 +26,188 @@ import (
...
 
-		return []interface{}{addrHex, string(name)}, nil
+		foundHiddenKernModsCache.Add(address, struct{}{}) // So we won't report multiple times
+
+		return []interface{}{addrHex, name, srcversion}, nil
 	}
 }
 
-func InitFoundHiddenModulesCache() error {
+func newModsCheckForHidden(startScanTime uint64, flags uint32) error {
```
rafaeldtinoco, line 130: Could you please "godoc" all the functions with a brief description of what they do and why ?

pkg/events/derive/hidden_kernel_module.go
rafaeldtinoco: > the logic of scanning would, in a regular ebpf event, likely be handled in the processEvents part of the pipeline.

> I think this shows we need another post-derive process stage in the pipeline to accommodate derived event processing.

I agree with @NDStrahilevitz here. We need a `processEvents` for the derived events. I would say this "wakeupChannel" should be global - for the derived pkgs - for all derived events scheduling callbacks to it, and not particular to a single event derived logic implementation. The callback would have a pointer for a function to be executed, with arguments (just like any other work queue implementation).

BUUUT I also think its too much for this PR. I think @OriGlassman should finish this work, and we should stop with all the uprobe triggering stuff until we DECOUPLE the uprobe/hooking functions/logic from the ebpf package. I prefer the opportunistic refactor approach but it did not happen since the first hooking logic was added, so...

Also, please note some things that CAN be done now:

1. spin off the hidden_kernel_module related functions to a new file in the golang ebpf package
2. spin off all the eBPF functions related to hidden kernel modules to a eBPF source file
3. improve overall documentation (not only function descriptions and summaries, but also the full logic).

About (3):

Logic to be documented for first code readers: from time to time trigger the hidden module investigation, multiple techniques using tailcalls, sending to userland but not reporting, investigating again, sending back and forth until finally user is warned.

Note: I know there is some description to this already, but it does not correspond to the entire logic and chronology of what happens when. Also, the code is already open source, so by fully documenting it we are only helping ourselves to maintain it in a near future (and not others, that will INEVITABLY copy the logic like done before).

pkg/events/derive/hidden_kernel_module.go
```diff
@@ -25,69 +26,188 @@ import (
...
+			return nil, err
+		}
+
+		if nameBytes == nil { // If we didn't parse it in procModules condition above, parse it now
+			nameBytes, err := parse.ArgVal[[]byte](&event, "name")
+			if err != nil {
+				return nil, err
+			}
+			name = string(nameBytes[:bytes.IndexByte(nameBytes[:], 0)]) // Remove the trailing terminating characters
+		}
```
rafaeldtinoco, line 111: This is the part I'm referring to as "the logic better documented", so when we have to support a bug here, for example, we don't have to re-read it all over (and can gain couple of hours).

Example: "we will receive this event once again with flags that will case it to be sent to user". What flags ? When will be sent to user ? How ? When will it come back ? Etc...

pkg/events/derive/hidden_kernel_module.go
rafaeldtinoco: IMO: lets have the logic working like it is, with no known issues after review, and then we should move this channel to something "generic" for all the event derivation logic in another PR (something like a "workqueue" logic for a derived events "processEvents" step (I mentioned this in a previous comment, based on a feedback already given by Nadav)).

The code is already too complex to include something like this now, IMHO: we should concentrate in having this work the best way, spin off code from big source files to their own files (so we can see all triggering related functions in a single place) and document it better.

I would move all the uprobe hook functions and methods from tracee object:

image

into its own file under ebpf package directory, for example.

rafaeldtinoco (changes requested):
@OriGlassman its a very ingenuous work (sorry it took me sometime to read your initial PR already merged and this one improving it). With that said, I left some comments on better documenting it, a few simple changes that might be good, positional changes into new files AND some capabilities raising fixes needed.

I did not take the time to check if the execution path of probes you used, for the race windows u mentioned, between the probes and event processing, could be mitigated by other type of probes, etc.

I know its more of a "research" thing, but I was wondering if this "flag based state machine" approach could be simplified by eBPF only logic making sure there are no races.

Either way, at the end, talking is cheaper than doing.. so good job =D

yanivagman:
@OriGlassman not sure what's the status here, 
both @NDStrahilevitz @rafaeldtinoco requested changes - is the PR ready for review?

OriGlassman:
> @OriGlassman not sure what's the status here, both @NDStrahilevitz @rafaeldtinoco requested changes - is the PR ready for review?

Was OOO for a couple of weeks, but I'm back now... PR should be ready in the next couple of days.

OriGlassman:
> I also want to review this one. One thing I spotted in code already merged:
> 
> ```go
> func clearMap(bpfMap *bpf.BPFMap) error {
> 	var err error
> 
> 	var iter = bpfMap.Iterator()
> 	for iter.Next() {
> 		addr := binary.LittleEndian.Uint64(iter.Key())
> 		err = bpfMap.DeleteKey(unsafe.Pointer(&addr))
> 
> 		if err != nil {
> 			logger.Errorw("err occurred DeleteKey: " + err.Error())
> 			return err
> 		}
> 	}
> 
> 	return nil
> }
> ```
> 
> There is a batch libbpf function for that, way more performatic I believe. Check:
> 
> https://github.com/aquasecurity/libbpfgo/blob/main/selftest/map-batch/main.go/#L112-L118

For that I need to know the keys in advance, no?

OriGlassman:
Ready for review from my side guys.. Thanks!
@rafaeldtinoco @yanivagman @NDStrahilevitz

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -688,21 +743,25 @@ static __always_inline bool find_modules_from_module_kset_list(program_data_t *p
     return !finished_iterating;
 }
 
-BPF_QUEUE(walk_mod_tree_queue, rb_node_t, 2048); // used to walk a rb tree
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 20, 0) || defined(CORE)
```
rafaeldtinoco, line 746: 4,20 ? Either way, we do not support non CORE any longer. So you can remove this ifdef here.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -688,21 +743,25 @@ static __always_inline bool find_modules_from_module_kset_list(program_data_t *p
...
+BPF_QUEUE(walk_mod_tree_queue, rb_node_t, MAX_NUM_MODULES); // used to walk a rb tree
 
+    #ifdef CORE // in non CORE builds it's already defined
```
rafaeldtinoco, line 749: no need (only CORE supported now)

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -585,21 +585,55 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
...
+             u64,
+             kernel_new_mod_t,
+             50); // Likewise for module insertion
+
+void __always_inline lkm_seeker_send_to_userspace(struct module *mod, u32 *flags, program_data_t *p)
 {
     reset_event_args(p);
+    u64 mod_addr = (u64) mod;
+    char *mod_name = mod->name;
+    const char *mod_srcversion = READ_KERN(mod->srcversion);
```
rafaeldtinoco, line 623: READ_KERN was dropped, should use BPF_CORE_READ (and similar) macros (directly from libbpf UAPI headers).

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -676,7 +731,7 @@ static __always_inline bool find_modules_from_module_kset_list(program_data_t *p
             struct module *mod = READ_KERN(mod_kobj->mod);
```
rafaeldtinoco, line 731: DITTO

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -774,20 +865,32 @@ static __always_inline bool check_is_proc_modules_hooked(program_data_t *p)
...
+            if (start_scan_time_init_shown_mods - (2 * 1000000000) < last_module_insert_time) {
+                continue;
+            }
+
             // Check again with the address being the start of the memory area, since
             // there's a chance the module is in /proc/modules but not in /proc/kallsyms (since the
             // file can be hooked).
-            key = (u64) READ_KERN(pos->core_layout.base);
-            mod_from_map = bpf_map_lookup_elem(&modules_map, &key);
+            mod = (u64) READ_KERN(pos->core_layout.base);
```
rafaeldtinoco, line 886: ditto.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3838,68 +4041,116 @@ int BPF_KPROBE(trace_do_init_module)
...
+        last_module_insert_time = insert_time;
 
-    // save module_data for kretprobe
-    bpf_map_update_elem(&module_init_map, &p.event->context.task.host_tid, &module_data, BPF_ANY);
+        if (!should_submit_module_load)
+            return 0;
+    }
 
-    return 0;
+    const char *version = READ_KERN(mod->version);
```
rafaeldtinoco, line 4067: ditto

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3838,68 +4041,116 @@ int BPF_KPROBE(trace_do_init_module)
...
+            return 0;
     }
 
-    // get next of original previous
-    struct list_head *orig_prev_ptr = (struct list_head *) (orig_module_data->prev);
-    u64 orig_prev_next_addr = (u64) READ_KERN(orig_prev_ptr->next);
-    // get previous of original next
-    struct list_head *orig_next_ptr = (struct list_head *) (orig_module_data->next);
-    u64 orig_next_prev_addr = (u64) READ_KERN(orig_next_ptr->prev);
+    const char *version = READ_KERN(mod->version);
```
rafaeldtinoco, line 4105: ditto

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3838,68 +4041,116 @@ int BPF_KPROBE(trace_do_init_module)
...
+        u64 addr = (u64) mod;
+        u32 flags = FULL_SCAN;
+        lkm_seeker_send_to_userspace((struct module *) addr, &flags, &p);
+        reset_event_args(&p); // Do not corrupt the buffer for the do_init_module event
+        if (!should_submit_do_init_module)
+            return 0;
+    }
+
+    // save strings to buf
+    const char *version = READ_KERN(mod->version);
```
rafaeldtinoco, line 4146: ditto

pkg/events/derive/hidden_kernel_module.go
```diff
@@ -38,34 +59,121 @@ func deriveHiddenKernelModulesArgs() deriveArgsFunction {
...
+// and reports if they are hidden
+func newModsCheckForHidden(startScanTime uint64, flags uint32) error {
+	// Since in old kernels it is not possible to iterate on a hashmap, the job is done here (userspace).
+	// It goes over a map that is maintained in the eBPF code (on each module insertion/deletion the map is updated),
+	// and checks if each module was seen (in modules list), and when was it inserted (to address race conditions).
+	// Based on this information, it determines whether it's hidden or not.
+	// If found hidden, it sends a message in the channel, which causes the address and the flags to get passed
+	// to the lkm submitter program (eBPF), which sends it back to userspace,
+	// this time with flags that will cause it to get submitted to the user as an event,
+	err := capabilities.GetInstance().Specific(
```
rafaeldtinoco, line 123: Should use the `EBPF` capabilities RING (it already contains the base line for all capabilities needed by picked events PLUS the ones needed for eBPF functions).

pkg/events/derive/hidden_kernel_module.go
```diff
@@ -38,34 +59,121 @@ func deriveHiddenKernelModulesArgs() deriveArgsFunction {
...
+	if err != nil {
+		logger.Errorw("Error occurred initializing kernel hidden modules: " + err.Error())
+	}
 	return err
 }
 
+// clearMap a utility to clear a map
 func clearMap(bpfMap *bpf.BPFMap) error {
-	err := capabilities.GetInstance().EBPF(
+	err := capabilities.GetInstance().Specific(
```
rafaeldtinoco, line 176: Should use the `EBPF` capabilities RING (it already contains the base line for all capabilities needed by picked events PLUS the ones needed for eBPF functions).

pkg/events/derive/hidden_kernel_module.go
```diff
@@ -74,94 +182,100 @@ func clearMap(bpfMap *bpf.BPFMap) error {
...
+func ClearModulesState() {
+	_ = clearMap(allModsMap)
+	_ = clearMap(recentDeletedModulesMap)  // We only care for modules that got deleted in the midst of a scan (and since we're not scanning now, clear the map)
+	_ = clearMap(recentInsertedModulesMap) // We only care for modules that got inserted in the midst of a scan (and since we're not scanning now, clear the map)
+}
 
-	err = capabilities.GetInstance().Specific(
+// FillModulesFromProcFs fills a map with modules from /proc/modules, to be checked in kernel-space for inconsistencies.
+func FillModulesFromProcFs(kernelSymbols helpers.KernelSymbolTable) error {
+	err := capabilities.GetInstance().Specific(
```
rafaeldtinoco, line 217: Should use the `EBPF` capabilities RING (it already contains the base line for all capabilities needed by picked events PLUS the ones needed for eBPF functions).

rafaeldtinoco: IF reading /proc/modules here needs another capability, then you can either add the capability to the base line (by adding as a requirement in the events.go event) OR, if this code is not **that frequent** you can use `Specific()` here, but with the capability event needed to read /proc/ fs only, not CAP_SYS_ADMIN).

rafaeldtinoco (commented):
I have added a few comments. The PR needs rebasing (and changing READ_KERN macro to BPF_CORE_READ). The rest, I have proposed a chore commit to your personal branch at https://github.com/OriGlassman/tracee/pull/1).

I believe we're close to merge, and I really want to spin this code out of tracee.bpf.c to something more "event specific" (I'll do it as chore actions later).

Nice work! I'll rebase myself tomorrow if you're out (Friday for you), lets see if we can get this merged before freezing the tree for the release. Cheers!

OriGlassman:
> I have added a few comments. The PR needs rebasing (and changing READ_KERN macro to BPF_CORE_READ). The rest, I have proposed a chore commit to your personal branch at [OriGlassman#1](https://github.com/OriGlassman/tracee/pull/1)).
> 
> I believe we're close to merge, and I really want to spin this code out of tracee.bpf.c to something more "event specific" (I'll do it as chore actions later).
> 
> Nice work! I'll rebase myself tomorrow if you're out (Friday for you), lets see if we can get this merged before freezing the tree for the release. Cheers!

Done

rafaeldtinoco (approved):
LGTM

Issue #3004, opened 2023-04-19
Single binary tracee isn't dropping capabilities correctly


rafaeldtinoco:
## Description

During investigation of #2998, I discovered that, when executing `tracee-ebpf`, the capabilities are dropped correctly, but when executing the single binary `tracee`, they are not.

That happens because `tracee` single binary calls a function to find signatures in a directory and that function raised capabilities. Since capabilities were not initialized yet (by tracee.New()) then they were initialized incorrectly (GetInstace() initializes capabilities if it isn't initialized yet).

## How to test this

```
while true; do for cap in $(sudo cat /proc/$(pidof tracee-ebpf)/status | grep CapEff | awk '{print $2}'); do capsh --decode=$cap; done; done
```

```
while true; do for cap in $(sudo cat /proc/$(pidof tracee)/status | grep CapEff | awk '{print $2}'); do capsh --decode=$cap; done; done
```

And compare effective capabilities on both.

Issue #2998, opened 2023-04-13
Legacy cgroup kprobe detachment fails when capabilities are dropped by default


rafaeldtinoco:
## Description

After:

commit 9e46a18bc411b347d7041ef01995f76d9190a36d
Author: Rafael David Tinoco <rafaeldtinoco@gmail.com>
Date:   Tue Mar 28 20:14:11 2023 +0000

    capabilities: change ring names and create base cap

Tracee started failing at its exist in kernels 5.4 (legacy kprobe detachment):

## Output of `tracee -v`:

```
# with capabilities drop:

ubuntu@ip-172-31-28-34:~/tracee$ sudo ./dist/tracee-ebpf --output format:json --filter event=net_packet_ipv4 --filter comm=othernc
{"level":"error","ts":1681416128.3713608,"msg":"failed to detach probes when closing tracee","err":"probes.(*probes).DetachAll: probes.(*cgroupProbe).detach: failed to detach program: cgroup_skb_egress (failed to detach (legacy) program cgroup_skb_egress from cgroupv2 /sys/fs/cgroup/unified)"}

# bypassing capabilities drop:

ubuntu@ip-172-31-28-34:~/tracee$ sudo ./dist/tracee-ebpf --output format:json --filter event=net_packet_ipv4 --filter comm=othernc --capabilities bypass=true
```

## Output of `uname -a`:

```
Linux ip-172-31-28-34 5.4.0-137-generic #154-Ubuntu SMP Thu Jan 5 17:03:22 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```

PR #3006, opened 2023-04-19
Capabilities fixes

rafaeldtinoco:
commit cb71563f (HEAD -> capabilities-fixes, origin/capabilities-fixes)
Date:   Wed Apr 19 04:01:45 2023

    capabilities: fix legacy detachment for older kernels
    
    Legacy attachments (for now: cgroupv2 prog attachments under older
    kernels) might not be done using BpfLink logic. Without a file
    descriptor for the link, tracee needs to raise its capabilities
    in order to call bpf() syscall for the legacy detachment.
    
    This fixes error happening in older kernels when exiting tracee:
    
      {"level":"error","ts":1681876470.9877832,"msg":"failed to detach
      probes when closing tracee","err":"probes.(*probes).DetachAll:
      probes.(*cgroupProbe).detach: failed to detach program:
      cgroup_skb_ingress (failed to detach (legacy) program
      cgroup_skb_ingress from cgroupv2 /sys/fs/cgroup/unified)"}
    
    Fixes: #2998

commit b063b4b0
Date:   Wed Apr 19 03:16:46 2023

    capabilities: fix early call to cap raise
    
    When executing tracee-ebpf, the capabilities are dropped correctly, but
    when executing the single binary, they are not.
    
    That happens because the single binary calls a function to find
    signatures in a directory, and that function raised capabilities. Since
    the capabilities gwere not initialized at that moment (by tracee.New()),
    then they were initialized incorrectly (since GetInstace() initializes
    capabilities if it isn't yet).
    
    This commit removes capabilities raising from early stage functions,
    in tracee rules logic, and postpones a bit the full capabilities setup
    This way, both tracee-ebpf and tracee binaries can coexist with
    a working initialization logic.
    
    Fixes: #3004

### DIFF
--- a/pkg/signatures/celsig/config.go
+++ b/pkg/signatures/celsig/config.go
@@ -8,9 +8,7 @@ import (
 	"strings"
 
 	"gopkg.in/yaml.v2"
-	"kernel.org/pub/linux/libs/security/libcap/cap"
 
-	"github.com/aquasecurity/tracee/pkg/capabilities"
 	"github.com/aquasecurity/tracee/pkg/logger"
 	"github.com/aquasecurity/tracee/types/detect"
 )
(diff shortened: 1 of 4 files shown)


rafaeldtinoco:
@josedonizetti I would like you to take a look at the changes related to "tracee-rules" (and tracee single binary) initialization, if possible. I postponed the privilege dropping so both tracee-ebpf and tracee could use the same "capabilities initialization" under Tracee.Init().

pkg/signatures/celsig/config.go
```diff
@@ -105,32 +103,26 @@ func NewConfigsFromDir(dirPath string) ([]SignaturesConfig, error) {
 func walkFilesWithExtensions(rootDir string, extensions []string) ([]string, error) {
 	var files []string
 
-	err := capabilities.GetInstance().Specific(
```
yanivagman, line 108: Why do we remove this request?

yanivagman: Ok, I see why from you commit comment.
I'm just trying to think what should be done in the future if we need to call this function in runtime (maybe we will support dynamically loading new signatures?)
Can't we move the capabilities initialization part to be before calling this func?

rafaeldtinoco: > I'm just trying to think what should be done in the future if we need to call this function in runtime (maybe we will support dynamically loading new signatures?)

Yes, that will be the "next step", together with the "mapping all things that need to be done for the dynamic unload/load of signatures" (for the API). For now I thought it would be ok to keep it like this and fix the bugs we currently have (For the release). WDYT ?

rafaeldtinoco:
@yanivagman I believe this is ready for merging (for the release).

yanivagman (approved):
LGTM

PR #3012, opened 2023-04-20
types: update tracee to use latest types

rafaeldtinoco:
This is needed after Tracee types JSON unmarshalling fixes.


rafaeldtinoco:
Alright, this fixes E2E test and it is chore, merging.

PR #3014, opened 2023-04-24
events: fix missing capability hidden_kernel_module

OriGlassman:
On older kernels, CAP_SYS_ADMIN is required to interact with eBPF maps. For newer kernels, either CAP_SYS_ADMIN or CAP_BPF is needed. tracee is running with CAP_BPF by default, hence on new kernels it works but not on old kernels (so need to elevate capability explicitly).

### 2. Explain how to test it

./tracee -f e=hidden_kernel_module


rafaeldtinoco:
EBPF ring deals with more stuff than kernel version:

image

to understand what capabilities are needed for eBPF kernel calls.

OriGlassman:
> @OriGlassman raising capabilities to EBPF ring should take care of needed capabilities (for older and newer kernels) so you don't have to raise to cap.SYS_ADMIN. Isn't that the case ? It works for all other map interactions. Example:
> 
> ```go
>   293  	if err == nil && info.Container.ContainerId == "" {
>   294  		// If cgroupId is from a regular cgroup directory, and not the
>   295  		// container base directory (from known runtimes), it should be
>   296  		// removed from the containers bpf map.
>   297: 		err := capabilities.GetInstance().EBPF(
>   298  			func() error {
>   299  				return t.containers.RemoveFromBpfMap(t.bpfModule, cgroupId, hId)
>   300  			},
>   301  		)
> ```

Wasn't aware of EBPF ring. Updated now, thanks.

rafaeldtinoco (approved):
LGTM

Issue #2624, opened 2023-01-26
Tracee can't be compiled with clang 15


rafaeldtinoco:
## Description

PR: #2581 has created a macro for `has_prefix()` to make clang-14 unroll loops from the eBPF code correctly. It was reported by @geyslan that clang-15 won't unroll loops correctly even with the 2 approaches we currently have. So, for now, tracee can only be built correctly by clang-12, clang-13, and clang-14.

Current commit: c8850253

In addition to this, @geyslan reported a verification error when loading the bpf program (compiled with clang-15) into the kernel (might be related to the above loop unrolling issue):

```
❯ sudo ./dist/tracee -f comm=uname
Place your finger on the fingerprint reader
{"level":"warn","ts":1685366140.1138046,"msg":"libbpf: prog 'trace_ret_vfs_write_tail': BPF program load failed: Permission denied"}
libbpf: prog 'trace_ret_vfs_write_tail': -- BEGIN PROG LOAD LOG --
0: R1=ctx(off=0,imm=0) R10=fp0
; int BPF_KPROBE(trace_ret_vfs_write_tail)
0: (7b) *(u64 *)(r10 -184) = r1       ; R1=ctx(off=0,imm=0) R10=fp0 fp-184_w=ctx
1: (b7) r1 = 0                        ; R1_w=0
; int zero = 0;
2: (63) *(u32 *)(r10 -80) = r1        ; R1_w=0 R10=fp0 fp-80=????0000
...
; if (path == NULL) {
3311: (15) if r9 == 0x0 goto pc+2     ; R9_w=map_value(off=0,ks=24,vs=4096,imm=0)
3312: (79) r6 = *(u64 *)(r10 -160)    ; R6_w=map_value(off=0,ks=4,vs=32152,imm=0) R10=fp0
3313: (05) goto pc+1945
; if (path_buf != NULL && !has_prefix("/dev/null", (char *) &path_buf, 10)) {
5259: (bf) r1 = r9                    ; R1_w=map_value(off=0,ks=24,vs=4096,imm=0) R9_w=map_value(off=0,ks=24,vs=4096,imm=0)
5260: (57) r1 &= 255
R1 bitwise operator &= on pointer prohibited
processed 1019 insns (limit 1000000) max_states_per_insn 0 total_states 61 peak_states 61 mark_read 35
-- END PROG LOAD LOG --
[...]

PR #3016, opened 2023-04-25
clang: fix nested loop unrolling

rafaeldtinoco:
It turns out that nested loop unrolling makes the unrolling transformation not to work in clang (at least for eBPF architecture):

./pkg/ebpf/c/common/common.h:83:5: warning: loop not unrolled: the optimizer was unable to perform the requested transformation; the transformation might be disabled or specified as part of an unsupported transformation ordering [-Wpass-failed=transform-warning]

    for (i = 0; i < n; prefix++, str++, i++) {
    ^

By having two "has_prefix" function versions, depending if the caller is already unrolling the loop or not, clang is able to transform the loop in multiple versions.

Fixes: #2624

### DIFF
--- a/pkg/ebpf/c/common/common.h
+++ b/pkg/ebpf/c/common/common.h
@@ -83,40 +83,11 @@ static __always_inline const char *get_device_name(struct device *dev)
 
 // INTERNAL: STRINGS -------------------------------------------------------------------------------
 
-// Workaround: Newer LLVM versions might fail to optimize has_prefix()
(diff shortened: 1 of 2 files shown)


rafaeldtinoco:
Vanilla 4.19 kernels (non CORE only) don't like this fix BUT they are going to be dropped with CORE only Tracee (https://github.com/aquasecurity/tracee/issues/1651) so I'm closing this and will have this commit together with the non CORE drop.

PR #3018, opened 2023-04-27
Disable argument parsing by default

### LINKED ISSUE #2177: Remove need for parse-arguments

### LINKED ISSUE #3008: Parse argument in a separate argument instead of inline
Currently when running `tracee --o option:parse-arguments` event arguments will be switched out to a string form where possible.
In order to support #2177 in REGO, it's required that rego signatures will be able to access parsed arguments, despite golang signatures being written with the assumption of non parsed arguments.

As such the logic of the flag should change to one of these two:
1. Add a new event `ParsedArgument` field which will include a parsed copy of the arguments.
2. For each argument append a new argument with the name `argname_parsed` which will have the parsed values.

REGO and cel-go signatures will then implicitly ALWAYS access those instead of the default arguments.

NDStrahilevitz:

1. types: add separete ParsedArguments list
    
    Prepare for removing inline argument parsing.
    
    This was done in a separate list in order to make the choice between
    querying the parsed and non parsed list as seamless as possible.
    Since the list is a copy with inlined parsed arguments, it can be used
    just like the previous arguments list was just with a name change.
5.  events: parse arguments into separate list
    
    Populate the ParsedArguments list previously introduced.
    Use this list in the table printer for argument printing.

6. signatures: work without argument parsing
    
    Refactor the golang signatures and helpers to work without enabling
    -o option:parse-arguments in the CLI.
    
    This included changing the signatures themselves to use the libbpfgo
    helper package kernel constants where relevant and refactoring our own
    helpers package to work with non parsed arguments (for example flags).
    
    Additionally added in the GetTraceeArgumentByName helper an option to
    query the ParsedArguments list instead of the regular one.

7. signatures: make rego sigs work with parsed args
    
    Refactored the rego helpers and tests to use parsed arguments access.
    This means when running rego signatures and/or using our helpers tracee
    must be run with -o option:parse-arguments.

### 2. Explain how to test it

Tested by sanity testing some signatures for example
1. `anti_debugging` -> `strace ls`
2. `ld_preload` -> `sudo touch /etc/ld.so.preload`
3. `proc_fops_hooking` -> `phide rootkit`

For rego i've just updated the signatures and tests and made sure they pass.

Resolve #2177
Resolve #3008

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -172,3 +172,5 @@ require (
 )
 
 replace github.com/kubernetes/cri-api => k8s.io/cri-api v0.23.5-rc.0
+
+replace github.com/aquasecurity/tracee/types => ./types
--- a/pkg/events/parse_args.go
+++ b/pkg/events/parse_args.go
@@ -23,8 +23,8 @@ func ParseArgs(event *trace.Event) error {
 	event.ParsedArgs = make([]trace.Argument, len(event.Args))
 	copy(event.ParsedArgs, event.Args)
 	for i := range event.ParsedArgs {
-		if ptr, isUintptr := event.Args[i].Value.(uintptr); isUintptr {
-			event.Args[i].Value = "0x" + strconv.FormatUint(uint64(ptr), 16)
+		if ptr, isUintptr := event.ParsedArgs[i].Value.(uintptr); isUintptr {
+			event.ParsedArgs[i].Value = "0x" + strconv.FormatUint(uint64(ptr), 16)
 		}
 	}
 
--- a/pkg/events/parse_args.go
+++ b/pkg/events/parse_args.go
@@ -20,19 +20,21 @@ type fdArgTask struct {
 }
 
 func ParseArgs(event *trace.Event) error {
-	for i := range event.Args {
+	event.ParsedArgs = make([]trace.Argument, len(event.Args))
+	copy(event.ParsedArgs, event.Args)
+	for i := range event.ParsedArgs {
 		if ptr, isUintptr := event.Args[i].Value.(uintptr); isUintptr {
 			event.Args[i].Value = "0x" + strconv.FormatUint(uint64(ptr), 16)
 		}
 	}
 
-	emptyString := func(arg *trace.Argument) {
+	emptyStringArg := func(arg *trace.Argument) {
 		arg.Type = "string"
 		arg.Value = ""
 	}
 
 	parseOrEmptyString := func(arg *trace.Argument, sysArg helpers.SystemFunctionArgument, err error) {
-		emptyString(arg)
+		emptyStringArg(arg)
 		if err == nil {
 			arg.Value = sysArg.String()
 		}
@@ -40,26 +42,26 @@ func ParseArgs(event *trace.Event) error {
 
 	switch ID(event.EventID) {
 	case MemProtAlert:
-		if alertArg := GetArg(event, "alert"); alertArg != nil {
+		if alertArg := getParsedArg(event, "alert"); alertArg != nil {
 			if alert, isUint32 := alertArg.Value.(uint32); isUint32 {
 				alertArg.Value = trace.MemProtAlert(alert).String()
 				alertArg.Type = "string"
 			}
 		}
-		if protArg := GetArg(event, "prot"); protArg != nil {
+		if protArg := getParsedArg(event, "prot"); protArg != nil {
 			if prot, isInt32 := protArg.Value.(int32); isInt32 {
 				mmapProtArgument := helpers.ParseMmapProt(uint64(prot))
 				parseOrEmptyString(protArg, mmapProtArgument, nil)
 			}
 		}
-		if prevProtArg := GetArg(event, "prev_prot"); prevProtArg != nil {
+		if prevProtArg := getParsedArg(event, "prev_prot"); prevProtArg != nil {
 			if prevProt, isInt32 := prevProtArg.Value.(int32); isInt32 {
 				mmapProtArgument := helpers.ParseMmapProt(uint64(prevProt))
 				parseOrEmptyString(prevProtArg, mmapProtArgument, nil)
 			}
 		}
 	case SysEnter, SysExit:
-		if syscallArg := GetArg(event, "syscall"); syscallArg != nil {
+		if syscallArg := getParsedArg(event, "syscall"); syscallArg != nil {
 			if id, isInt32 := syscallArg.Value.(int32); isInt32 {
 				if Core.IsDefined(ID(id)) {
 					eventDefinition := Core.GetDefinitionByID(ID(id))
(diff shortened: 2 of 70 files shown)


go.mod
```diff
@@ -154,3 +154,5 @@ require (
 )
 
 replace github.com/kubernetes/cri-api => k8s.io/cri-api v0.23.5-rc.0
+
+replace github.com/aquasecurity/tracee/types => ./types
```
NDStrahilevitz, line 158: Create a types PR once approved.

cmd/tracee/main.go
```diff
@@ -74,7 +74,7 @@ func main() {
 			}
 
 			// parse arguments must be enabled if the rule engine is part of the pipeline
-			runner.TraceeConfig.Output.ParseArguments = true
+			// runner.TraceeConfig.Output.ParseArguments = true
```
AlonZivony, line 77: If you plan to remove this line, then remove it.
Don't leave a commented code.

types/trace/trace.go
```diff
@@ -41,6 +41,7 @@ type Event struct {
 	StackAddresses       []uint64     `json:"stackAddresses"`
 	ContextFlags         ContextFlags `json:"contextFlags"`
 	Args                 []Argument   `json:"args"` //Arguments are ordered according their appearance in the original event
+	ParsedArgs           []Argument   `json:"parsedArgs,omitempty"`
```
AlonZivony, line 44: The current convention is that we pass Nil values if arguments are missing.
Shouldn't we apply it on the parsed args too?
Open for discussion here.

pkg/signatures/benchmark/signature/golang/anti_debugging.go
```diff
@@ -52,21 +54,22 @@ func (sig *antiDebugging) OnEvent(event protocol.Event) error {
...
-	requestString := request.Value.(string)
-	if requestString != "PTRACE_TRACEME" {
-		return nil
+
+	if requestArg == int(libbpfgo.PTRACE_TRACEME.Value()) {
+		metadata, err := sig.GetMetadata()
+		if err != nil {
+			return err
+		}
+		sig.cb(detect.Finding{
```
AlonZivony, line 67: I will write it here but it is true for all other signatures.
Should we really omit the data we pass here in the `detect.Finding`?
I think you should parse the argument if the signature is triggered, and pass it through the data member as before.
This way you parse the argument only upon triggering, instead of each event.

signatures/helpers/arguments_helpers.go
```diff
@@ -45,12 +53,34 @@ func GetTraceeIntArgumentByName(event trace.Event, argName string) (int, error)
...
+		return uint(argUint32), nil
+	}
+	argUint64, ok := arg.Value.(uint64)
 	if ok {
-		return int(argInt), nil
+		return uint(argUint64), nil
 	}
 
-	return 0, fmt.Errorf("can't convert argument %v to int", argName)
+	return 0, fmt.Errorf("can't convert argument %v to uint (argument is of type %T)", argName, arg.Value)
```
AlonZivony, line 83: Nice, much better error messages!

AlonZivony (changes requested):
Few thoughts on the PR (didn't read each signature)

types/trace/trace.go
NDStrahilevitz: I'm not sure if you're suggesting that:
1. we should drop the `omitempty`
2. Parsed arguments should go inside Args and be nil (with a different name)
3. Something else

Could you elaborate please?

pkg/signatures/benchmark/signature/golang/anti_debugging.go
NDStrahilevitz: The most elegant solution I can think of is adding this logic into the detection callback passed in tracee-rules.
However outside of the single binary solution this would mean importing the `events` package into tracee-rules AND tracee-rules needing awareness of parse-arguments configuration (from `OutputConfig`) which is _currently_ a bit of an issue, but when we drop the old arch we can add that.

Otherwise we would need to supply a robust argument parsing solution like we have in `parge_args.go` that is external to tracee's go.mod so that signatures can import it without breaking their plugin compilation (because currently they would have to import `pkg/events`), or let the signature authors take responsibility for the parsing if it matters for their usecase.

AlonZivony: I agree with you, but as you change the signatures here you are the signature author.
I suggest you use the `libbpfgo` helpers to parse the argument and put the parsed argument in `Data` map.

types/trace/trace.go
NDStrahilevitz: > So I liked the idea of separating the parsed arguments from the arguments, as we don't plan to support them for long.

Actually I had no intention of removing the parsed arguments, just keeping them separated and not overriding the originals.

> I suggested here to remove the `omitempty`

This is just so when users run with `-o json` but not with `parse-arguments`, they won't see the empty field each time.

AlonZivony: > This is just so when users run with `-o json` but not with `parse-arguments`, they won't see the empty field each time.

So we talked previously that breaking the expected structure of events with missing arguments might sabotage efforts to put the events in databases which aren't flexible.
I ask here if it won't be a problem here as well.

pkg/events/parse_args.go
```diff
@@ -20,46 +20,48 @@ type fdArgTask struct {
 }
 
 func ParseArgs(event *trace.Event) error {
-	for i := range event.Args {
+	event.ParsedArgs = make([]trace.Argument, len(event.Args))
+	copy(event.ParsedArgs, event.Args)
+	for i := range event.ParsedArgs {
 		if ptr, isUintptr := event.Args[i].Value.(uintptr); isUintptr {
 			event.Args[i].Value = "0x" + strconv.FormatUint(uint64(ptr), 16)
 		}
```
yanivagman, line 28: Looks like you override `event.Args` instead of `event.ParsedArgs`

NDStrahilevitz:
@yanivagman and I have decided to give up on the double lists and keep the inlined parsing. As such I will open a similar PR updating only the signatures and documentation in the near future.

PR #3021, opened 2023-05-01
fix: detect container id from cgroup in  GitHub Action

ShiraCohen33:

Fix a bug related to detect container id when run docker on GitHub Action

### 2. Explain how to test it

run in Github Action `dist/tracee-ebpf -f e=container_create,existing_container,cgroup_mkdir -o json --containers`, run docker and check if the event contains the container id and other data.


yanivagman (approved):
LGTM

PR #3022, opened 2023-05-01
fix: kernel version comparison

roikol:
in `security_kernel_read_file` parsing of the argument `type` has an error in one condition - making it never true. thus the parsing fails in some kernel versions.

```
Date:   Mon May 1 13:54:53 2023 +0300

    fix: kernel version comparison

    fix never-true if condition when comparing kernel versions
    in initKernelReadFileTypes func
```


yanivagman (approved):
LGTM

PR #3025, opened 2023-05-02
events: fix hidden_kernel_module derivation

rafaeldtinoco:
- missed a lockup in capabilities ring raising during review, fix it
- fixed logger errors (needs key/value) and used correct error func


rafaeldtinoco:
@OriGlassman FYI

PR #3026, opened 2023-05-02
Signatures data sourcing

### LINKED ISSUE #1423: Signatures configuration / external data
Hello!
We would like the Go Signatures to get content from an external information source. To give a more practical use case, you can consider the following: a signature might want to know if a request over the network to a suspicious IP is performed.
We do not want the signature to have hard-coded list of black IPs, as this list might change and get updated. We would like instead that, on demand, the signature could access to a "DB Layer" which is feeded by tracee-rules. 
In particular, tracee-rules will expose an API that its consumers can use in order to push data to the DB Layer. The signatures, when required will then be able to pull data from it. Schematically, the proposal would look like something like that:
image
From a more "code level" point of view I would suggest to have:
1. a new package `tracee/pkg/dblayer` which will define an interface to the DB layer: `type DbLayer interface` with Push and Pull operations 
2. tracee-rules engine will store as one of its field an instance to the DB layer
3. the Signature interface will expose a new method `SetExternalData(dbLayer DbLayer) error`
4. tracee-rules will expose a new function `PushExternalData` 
5. obviously the DB layer will protect its data with reader/writer lock

So far I see the following open questions:
1. what exactly will be the input for `PushExternalData`? a json? if so, how will the json be stored within the DB layer implementation?
2. what about data persistency? Eventually the DB layer is an in memory layer. If the process, for some reason, fails then we want to restart and reload the data we were previously storing.
3. I can see how this can work out with Go signatures. Not sure though regarding Rego. Is this something we want to support?

One more note: all the above is a suggestion and feedbacks/ideas are more than welcome! :)

### LINKED ISSUE #2561: Add state saving mechanism to tracee-rules
# The problem:
Rules which are based on state saving needs to manage their own state - this can lead to duplication in state saving for similar rules.
If a global is used there are no duplicates but because rules aren't ordered we can't know which rules will get the event first and save it to the global.

# The solution:
A state saving mechanism which will be first in the event consumption pipeline of tracee-rules - and will later be relocated to the single binary to the end of the ebpf pipeline.

current state:
tracee-ebpf ---pipe---> tracee-rules parse event ----> send to each rule if was selected

suggested state:
tracee-ebpf ---pipe---> tracee-rules parse event ----> run state saving functions if defined for this event ----> send to each rule if was selected

This way tracee itself maintains a single copy of any LRU for state saving and updates it prior to the event getting to any rule.
The proposed addition to a rule will be adding function:
```
func (sig *<signame>) GetStateSavingFunc() (string, func(string, trace.Event)) {
	return "sched_process_exec", "process_executed", helpers.ProcessCreatedStateSaving
}
```

In this example:
- "sched_process_exec" is the event to be sent to the helper function
- "process_executed" is the LRU name
- `helpers.ProcessCreatedStateSaving` function is a shared function that does some logic and saves to the LRU name provided to it according to the logic

Tracee-rules will iterate over all the rules before starting them and will create a map of unique functions to be run and LRUs to be created, so if multiple rules use the same LRU name and run the same function - only one instance of it will run.

NDStrahilevitz:
The following changes were made

1. Add the capability to register data sources for signatures and request them in their Init() stage.
2. In "everything is an event mode" register a containers data source from tracee's `containers` package.
3. Update the e2e instrumentation tests to work with the single binary architecture. This was done in order to test the containers data source feature.

Resolve #2561
Resolve #1423


go.mod
```diff
@@ -154,3 +155,5 @@ require (
 )
 
 replace github.com/kubernetes/cri-api => k8s.io/cri-api v0.23.5-rc.0
+
+replace github.com/aquasecurity/tracee/types => ./types
```
NDStrahilevitz, line 159: Move to a types PR once approved.

AlonZivony (approved):

types/detect/detect.go
```diff
@@ -67,3 +68,10 @@ type Logger interface {
 	Warnw(format string, v ...interface{})
 	Errorw(format string, v ...interface{})
 }
+
+type DataSource interface {
+	Get(string) (interface{}, bool)
+	ValueType() string
+	Namespace() string
```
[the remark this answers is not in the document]
NDStrahilevitz: Namespace is to avoid ID collisions when we allow external users to plug their own data sources from plugins (in CNDR we already have such a usecase).
ID is basically a data source name.
They practically make a joint key for a DataSource.

tests/e2e-instrumentation-signatures/e2e-containers_data_source.go
```diff
@@ -0,0 +1,95 @@
...
+
+type e2eContainersDataSource struct {
+	cb             detect.SignatureHandler
+	containersData detect.DataSource
+}
+
+func (sig *e2eContainersDataSource) Init(ctx detect.SignatureContext) error {
+	sig.cb = ctx.Callback
+	containersData, ok := ctx.GetDataSource("tracee", "containers")
+	if !ok {
```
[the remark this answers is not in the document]
NDStrahilevitz: I've thought of adding a method like the `GetEventSelectors` but I didn't want to break all signatures and require them to declare mostly empty selectors.
I've already faced errors in terms of ordering (a signature would be loaded before registering the data source and then fail, so i've had to add data sources into the config like signatures), but i'm not sure if selectors would solve that specifically.

Either way i'm open to changing this implementation so i'm keeping the discussion open.

types/detect/detect.go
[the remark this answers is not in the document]
AlonZivony: So can you put it as a comment on each of them and maybe on the interface so it will be clear?

types/detect/detect.go
```diff
@@ -22,8 +22,9 @@ type Signature interface {
 }
 
 type SignatureContext struct {
-	Callback SignatureHandler
-	Logger   Logger
+	Callback      SignatureHandler
+	Logger        Logger
+	GetDataSource func(namespace string, id string) (DataSource, bool)
```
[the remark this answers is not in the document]
AlonZivony: I thought putting a struct that implements some defined interface.
It is still possible to pass here a functions of an object to change the behavior in the future.
I think it is better to use defined structs rather than unnamed functions as the interface here.

tests/e2e-instrumentation-signatures/e2e-containers_data_source.go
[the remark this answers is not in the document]
AlonZivony: Maybe we can add a new method like `GetEventSelectors` to a new signature interface, and just assume by default that it is an empty list if the function is not implemented.

AlonZivony (changes requested):
Fix status of my CR

tests/e2e-instrumentation-signatures/e2e-containers_data_source.go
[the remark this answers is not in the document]
NDStrahilevitz: It's a good approach since it's basically versioning, but because it's versioning it would require us to tackle that whole beast for this.

types/detect/detect.go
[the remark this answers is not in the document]
NDStrahilevitz: What's the interface we need here besides requesting the `DataSource` the signature needs?
As I see it, from the signature's PoV the function _is_ it's interface for requesting the DataSource.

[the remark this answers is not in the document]
AlonZivony: So I agree that in Go it is possible to look at this function as an interface.
However, you encourage the usage of hardcoded function instead of more robust approach.
This distinction is not so important, so I won't be persistent on my opinion here. Function is good enough interface.

AlonZivony, line 27: Maybe create a struct for the namespace and ID.
It would reduce the amount of functions needed to implement in the interface and make it easier to change names and such in the future.

NDStrahilevitz: How would it reduce the number of functions?
However I can see the merits if this function changes using a struct will reduce the chance for breaking changes on the calling side.
On the implementing side this is an interface we implement in the engine's side, the signature author only consumes the interface as a user not as an implementor.

go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/google/cel-go v0.15.1
 	github.com/google/gopacket v1.1.19
 	github.com/hashicorp/golang-lru v0.5.4
+	github.com/hashicorp/golang-lru/v2 v2.0.2
```
josedonizetti, line 17: can we relace all use of the `golang-lru` to `golang-lru/v2`? so we have everything on the latest version, instead of part using v2 and part using the older version?

pkg/ebpf/signature_engine.go
```diff
@@ -19,6 +20,9 @@ func (t *Tracee) engineEvents(ctx context.Context, in <-chan *trace.Event) (<-ch
 	engineOutput := make(chan detect.Finding, 100)
 	engineInput := make(chan protocol.Event)
 	source := engine.EventSources{Tracee: engineInput}
+
+	t.config.EngineConfig.DataSources = append(t.config.EngineConfig.DataSources, containers.NewDataSource(t.containers))
```
josedonizetti, line 24: Why initialize it here? I feel like this should be done at the creation of the engine.Config, and not hidden inside the pipeline.

go.mod
NDStrahilevitz: I think I did move all usages of `golang-lru` to `v2` but for some reason when `go mod tidy`ied afterwards it kept it there.

pkg/ebpf/signature_engine.go
NDStrahilevitz: I'll see if I can make this move work.

rafaeldtinoco:
I want to review this one and will do tomorrow (as we're approaching freeze date).

pkg/containers/datasource.go
```diff
@@ -0,0 +1,72 @@
...
+		"k8s_pod_name":      "string",
+		"k8s_pod_namespace": "string",
+		"k8s_pod_sandbox":   "bool",
+	}
+	schema, _ := json.Marshal(schemaMap)
+	return string(schema)
+}
+
+func (ctx SignaturesDataSource) Version() uint {
+	return 1
```
rafaeldtinoco, line 63: WDYT about standartazing YYYYMMDDHHMM (or anything that can be monotonically increased whenever changed) ? This idea comes from the DNS SOA serial number (there is a RFC describing that each time the zone config changes the serial number should be increased to that format, or similar, so you are always increasing numbers AND, at the same time, you know the last time of change (in our case of the "format").

pkg/signatures/engine/engine.go
```diff
@@ -344,3 +367,36 @@ func (engine *Engine) GetSelectedEvents() []detect.SignatureEventSelector {
 	}
 	return res
 }
+
+func (engine *Engine) RegisterDataSource(dataSource detect.DataSource) error {
```
rafaeldtinoco, line 371: Please improve godocs for exported methods (All of them if possible).

pkg/ebpf/signature_engine.go
rafaeldtinoco: I came here for the same reason =D.

pkg/containers/datasource.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func (ctx SignaturesDataSource) Get(key interface{}) (map[string]interface{}, error) {
+	containerId, ok := key.(string)
+	if !ok {
+		return nil, detect.ErrKeyNotSupported
+	}
+	ctx.containers.mtx.RLock()
+	defer ctx.containers.mtx.RUnlock()
+	for _, cgroup := range ctx.containers.cgroupsMap {
```
rafaeldtinoco, line 26: Thinking out loud: should we worry about this iteration performance ? should we have a map indexed by containerID (and updated during CgroupInfo updates) ? etc...

pkg/events/derive/hidden_kernel_module.go
```diff
@@ -24,7 +24,7 @@ import (
 )
 
 var (
-	foundHiddenKernModsCache *lru.Cache
+	foundHiddenKernModsCache *lru.Cache[uint64, struct{}]
```
rafaeldtinoco, line 27: Why is this file being changed from v0 to v2 ? Just curious, as we might have many other places using v0 LRU non generics way.

rafaeldtinoco: Actually, we have "simplelru" instead of pure v0 I believe (symbols_collision, symbols_loaded, host_symbols_loader).

rafaeldtinoco: I wonder if we should uniformize this. Another thing that occurred to me as well, we have protected generic based maps helper funcions (I believe @AlonZivony created, was it in a non merged PR ?). 

I think we will eventually have to make it uniform:

- mutex protected maps (which is what Alon did IIRC) helper
- same LRU approach in all places (sake of maintainability)
- probably more atomic map like helpers

pkg/pcaps/common.go
```diff
@@ -71,7 +71,7 @@ func getItemIndexFromEvent(event *trace.Event, itemType PcapType) interface{} {
 		return ret
 	}
 
-	return new(interface{})
+	return ""
```
rafaeldtinoco, line 74: Oh shit, I don't have a test for pcap code yet (there is an issue opened for it). Will have to test to make sure files are being kept opened while LRU entry exists (and being evicted correctly). Will need to test this tomorrow :\ just to make sure.

pkg/signatures/engine/engine.go
```diff
@@ -22,18 +22,21 @@ type Config struct {
...
+	signaturesIndex  map[detect.SignatureEventSelector][]detect.Signature
+	signaturesMutex  sync.RWMutex
+	inputs           EventSources
+	output           chan detect.Finding
+	waitGroup        sync.WaitGroup
+	config           Config
+	stats            metrics.Stats
+	dataSources      map[string]map[string]detect.DataSource
+	dataSourcesMutex sync.RWMutex
 }
```
rafaeldtinoco, line 40: I know we did not have comments or the existing maps of maps (example: I needed to refresh memory going to older code for the singaturesIndex map of arrays). Could we add some comments here (why map[string]map[string], since string is too generic to understand at a first glance).

pkg/signatures/engine/engine.go
```diff
@@ -344,3 +367,36 @@ func (engine *Engine) GetSelectedEvents() []detect.SignatureEventSelector {
...
+func (engine *Engine) RegisterDataSource(dataSource detect.DataSource) error {
+	engine.dataSourcesMutex.Lock()
+	defer engine.dataSourcesMutex.Unlock()
+
+	namespace := dataSource.Namespace()
+	id := dataSource.ID()
+
+	if _, ok := engine.dataSources[namespace]; !ok {
+		engine.dataSources[namespace] = map[string]detect.DataSource{}
+	}
```
rafaeldtinoco, line 380: So what happens if we register same data source twice ? Or a diff data source with the same namespace ? Could you document the namespace directive, what it means, obligations of callers, etc ? Thanks!

tests/e2e-instrumentation-test.sh
```diff
@@ -139,33 +133,30 @@ for TEST in $TESTS; do
     ## cleanup at EXIT
 
     found=0
-    cat $SCRIPT_TMP_DIR/build-$$ | grep "Signature ID: $TEST" -B2 | head -3 | grep -q "\*\*\* Detection" && found=1
+    cat $SCRIPT_TMP_DIR/build-$$ | grep $TEST && found=1
```
rafaeldtinoco, line 136: I have not executed this. Is this really "all needed" ? Should we grep for something more complex ? Just worried about an error in stdout with the name of a signature (dunno, cannot parse event XXXX, and we consider it was triggered, or something crazier). Trying to explore what could go wrong, not criticizing.

types/detect/detect.go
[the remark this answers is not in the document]
rafaeldtinoco: Hum, I left a comment saying that the map[string]map[string]XX line should also mention it (sorry for the duplicate request).

tests/e2e-instrumentation-signatures/e2e-containers_data_source.go
```diff
@@ -0,0 +1,99 @@
...
+		return fmt.Errorf("containers data source not registered")
+	}
+	if containersData.Version() > 1 {
```
rafaeldtinoco, line 23: good reason for the version number suggestion (ditto in other comment).

types/detect/detect.go
```diff
@@ -67,3 +70,27 @@ type Logger interface {
...
+	//	- ErrKeyNotSupported - When the key used does not match to a support key
+	//	- Otherwise errors may vary.
+	Get(interface{}) (map[string]interface{}, error)
+	// Version of the data fetched. Whenever the schema has a breaking change the version should be incremented.
+	// Consumers of the data source should verify they are running against a support version before using it.
+	Version() uint
+	// The types of keys the data source supports.
+	Keys() []string
+	// JSON Schema of the data source's result. All Get results should conform to the schema described.
+	Schema() string
```
rafaeldtinoco, line 88: Last question I guess. When are you planning on checking for conformity ? (to check if schema has changed, why wouldn't the version be enough and "a contract" between user and the data source) ? Not saying it is a bad idea to have the Schema() method, just trying to understand your thoughts!

yanivagman:
Going to review this soon, but one comment about the naming of "data source" - I think it is not exactly a source of data, but more a store of data, so a more appropriate name might be "data store"

pkg/containers/datasource.go
NDStrahilevitz: So instead of the `int` we would return a `time.Time` and then signature consumers would compare to that date? The versioning is a bit more descriptive but not sure if it would be as user friendly. On the fence for this.

pkg/containers/datasource.go
NDStrahilevitz: This is an example implementation. I'm not sure if it's optimal performance wise (likely not), I initially toyed with improving it but that led me to playing with the cgroups package internals, which I felt was out of scope. Probably adding caching at the datasource layer could help.

pkg/events/derive/hidden_kernel_module.go
NDStrahilevitz: I initially wanted to implement data sources using LRUs as the backend always, but that attempt didn't work out due to generics usage issues.
As part of that commit I checked all places (in tracee open source itself) we use the hashicorp lru and updated it, so I don't think there are any left overs. As I already finished the commit thought I might as well leave it in the PR.

pkg/pcaps/common.go
NDStrahilevitz: Might be best I drop the LRU commit then, maybe add to a separate PR, as it's not strictly needed functionality wise here. If you want you could even add pcaps test + this in one PR, might be best.

pkg/signatures/engine/engine.go
NDStrahilevitz: Sure, i'll add for the other while i'm at it too.

pkg/signatures/engine/engine.go
NDStrahilevitz: diff data source in the same namespace gets registered to that namespace. ID collision in a namespace returns an error.
I'll document it in the godoc.

tests/e2e-instrumentation-test.sh
NDStrahilevitz: Consider that we're using "everything is an event" single binary here, so the previous string was no longer viable. Not sure if there is a better query I could do. Maybe if there was a stripped down printer only for event names it would be safer.

NDStrahilevitz: Actually I should just use json output and query the json with eventName. That would likely be best.

tests/e2e-instrumentation-signatures/e2e-containers_data_source.go
NDStrahilevitz: Could you give a pseudo code example to explain how it would be an improvement?

types/detect/detect.go
NDStrahilevitz: I'm not, this is more of a documentation in code thing that data source creators should add. I'm toying with adding a "DataSourceManager" which would proxy calls through it to data sources (this may be needed for "tracee analyze" compatibility later so we can export data source history), and then we could validate there, but not sure if it's a good idea performance-wise.

pkg/containers/datasource.go
rafaeldtinoco: Ah gotcha, sorry, did not mean to rush things!

tests/e2e-instrumentation-signatures/e2e-containers_data_source.go
rafaeldtinoco: no pseudocode needed, sorry, its literally just an integer value being manually monotonically increased each time the datasource value "interface{}" is altered (like in the given example, from the DNS RFC document, it is basically it).

rafaeldtinoco:
> but more a store of data, so a more appropriate name might be "data store"

+1 for naming.

pkg/events/derive/hidden_kernel_module.go
AlonZivony: > * mutex protected maps (which is what Alon did IIRC) helper

I implemented in my process tree PR (#3054) a RWMap safe to use from multiple go routines.
If you want you can use it (just copy the code to this PR), or use the `sync.map` (which use interfaces instead of generics).
[You can see the code here](https://github.com/aquasecurity/tracee/pull/3054/files#diff-d3fe810a7fe6849ae66f7c3f4e68fc48e877a5cd97a7b8ca726353052d4b70f8)

pkg/ebpf/signature_engine.go
NDStrahilevitz: So the data source implementation uses tracee's internal container tracker `(*tracee).containers`.
We create the EngineConfig in `GetTraceeRunner`, where the `tracee` object isn't created yet. As such I can't create them at the same time.
Not sure if a refactor to make this possible is worth it, at least right now.

pkg/containers/datasource.go
NDStrahilevitz: So instead we would check `Version > 1684936308 => fail` instead of `Version > 1 => fail`? I guess we could except we would have to document it a bit more specifically. 
Still on the fence as this requires _slightly_ more effort to parse and when updating sources, instead of just incrementing the counter. So what you suggest has more info encoded but we could encode that info somewhere else in the data source as well.

pkg/events/derive/hidden_kernel_module.go
NDStrahilevitz: Well a safe RWMap and a LRU cache are not exactly the same thing, I think LRUs are mainly useful when we want to limit the number of items. RWMap has no such syntax. So we likely have a use for both.

NDStrahilevitz: Either way I will separate this out to a PR for v0.16.0 and @rafaeldtinoco you can commit on top of it.

go.mod
rafaeldtinoco: https://github.com/aquasecurity/tracee/pull/3141

rafaeldtinoco: I'm doing it. Creating another PR (rebase of this without the types change) as well.

rafaeldtinoco (approved):
LGTM (and I believe we already had many feedback and back and forth, since the freeze is tomorrow I believe I can merge this as is and ask for minor changes/fixes the next week before the release, hope that is okay for everybody or else it won't make it).

tests/e2e-instrumentation-signatures/e2e-containers_data_source.go
rafaeldtinoco: TODO: check if you want to address this after freeze (freezing tomorrow). This will be merged as is by https://github.com/aquasecurity/tracee/pull/3142

pkg/signatures/engine/engine.go
rafaeldtinoco: TODO: check if you want to address this after freeze (freezing tomorrow). This will be merged as is by https://github.com/aquasecurity/tracee/pull/3142

pkg/signatures/engine/engine.go
rafaeldtinoco: TODO: check if you want to address this after freeze (freezing tomorrow). This will be merged as is by https://github.com/aquasecurity/tracee/pull/3142

pkg/containers/datasource.go
rafaeldtinoco: TODO: check if you want to address this after freeze (freezing tomorrow). This will be merged as is by https://github.com/aquasecurity/tracee/pull/3142

pkg/ebpf/signature_engine.go
rafaeldtinoco: Fair enough.

rafaeldtinoco:
Will be merged as https://github.com/aquasecurity/tracee/pull/3142 keeping appropriate credits. TODOs within the comments for the release (to be done next week). Cheers!

pkg/containers/datasource.go
yanivagman: I thought about semantic versining (major-minor) where major is increased in breaking changes (schema changed) and minor is increased in addition of new fields. Then, a signature keeps working as long as the data source has the same major, and when it adds new fields, only signatures that look for a new field will also check if the minor is big enough.

pkg/containers/datasource.go
yanivagman: I left the same comment in the other PR that was merged.
In the future, we will probably want to add such a map to be O(1) for getting a container.

yanivagman:
> > but more a store of data, so a more appropriate name might be "data store"
> 
> +1 for naming.

So I guess we're staying with the data source naming after all?

rafaeldtinoco:
> So I guess we're staying with the data source naming after all?

I thought we could rename it next week, as it's not a logical change!

PR #3029, opened 2023-05-02
docs: fix docs for cobra/viper flags

josedonizetti:


geyslan:
> @yanivagman This was only for @geyslan to cherry pick, the PR will have the docs as well

It's in the oven.

Issue #3036, opened 2023-05-03
triggered events are not triggered when running on WSL


roikol:
when running tracee in WSL, triggered events (i.e. print_mem_dump, hooked_syscalls, etc.) are not triggered.

the reason: 
at the start of the uprobe func in tracee.bpf.c, there's a check if the triggering process ID equals to the tracee PID from the config (https://github.com/aquasecurity/tracee/blob/main/pkg/ebpf/c/tracee.bpf.c#L1504).
WSL works by having one VM per windows user, and every distro ran by that user is namespaced within this one VM. meaning - WSL gives namespaced access.

so when running tracee in WSL - tracee uses the namespaces PID instead of hostPID, thus the events are not triggered.

yanivagman:
Closed by #3037

PR #3037, opened 2023-05-03
feat: support running tracee from pid ns

roikol:
issue: #3036

```
Date:   Wed May 3 17:33:53 2023 +0300

    feat: support running tracee from pid ns

    in triggered events, check the pid and host_pid of the triggering
    process against tracee_pid to allow running tracee from non-host pid ns
```

### 2. Explain how to test it

run tracee inside WSL
command: `./dist/tracee-ebpf -f e=print_mem_dump -f print_mem_dump.args.symbol_name=security_bprm_check`

upon tracee startup you should get `print_mem_dump` event.

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -835,9 +835,9 @@ int uprobe_lkm_seeker(struct pt_regs *ctx)
     p.event->context.syscall = NO_SYSCALL;
     p.event->context.matched_policies = ULLONG_MAX;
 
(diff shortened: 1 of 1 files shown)


rafaeldtinoco:
Maybe create some little named var pointers for the comparison ? So lines aren't so big ? either way "fix-fmt" as well for the test passing.

yanivagman (approved):
LGTM

rafaeldtinoco:
@yanivagman could this be considered a fix for the .1 version ?

Issue #2950, opened 2023-03-31
libbpf print output is raw and escaped


geyslan:
## Description

As reported in aquasecurity/libbpfgo#313 and discussed in aquasecurity/libbpfgo#312, the output of libbpf is an escaped string.

Tracee logger is based on zap which uses a json structure format which cannot interpret escaped strings uber-go/zap#963.

After discussion we raised the issue https://github.com/aquasecurity/tracee/issues/3035 to be the focal point of the new logger feature `--log filter`.

For now, as we have bumped the libbpfgo to `v0.4.8-libbpf-1.2.0`, this issue can be solved with:

- [x] remove suffix "\n" from libbpf msg before logging.

## Output of `tracee -v`:

```
Tracee version "v0.8.3-596-ge1018ba3"
```

## Output of `uname -a`:

```
Linux hb 6.2.7-2-MANJARO #1 SMP PREEMPT_DYNAMIC Sat Mar 18 17:15:16 UTC 2023 x86_64 GNU/Linux
```

## Additional details 

Next step is https://github.com/aquasecurity/tracee/issues/3035.

rafaeldtinoco:
Couldn't play with this one before the release. Maybe we can include it in the buf fix release right after (1st week).

Issue #3035, opened 2023-05-03
Create `--log filter`, `--log filter-out` options


geyslan:
> As the team discussed today, we agreed to use a single base logger, as it already is, but implement --log filter options, such as for packages or raw strings. It would answer all needs.

Possible options:

- `--log filter:msg=somemsg1` (outputs logs containing somemsg1)
- `--log filter:msg=somemsg1,somemsg2` (outputs logs _containing_ somemsg1 or somemsg2)
- `--log filter:pkg=somepkg2,somepkg2` (outputs logs originated in somepkg1 or somepkg2)
- `--log filter:file=somefile1,somefile2` (outputs logs originated in somefile1 or somefile2)
- `--log filter:libbpf` (outputs logs originated in libbpf - tackle verifier formatted output as well)

`filter-out` should have the same options. 

Context: https://github.com/aquasecurity/tracee/issues/2823 #2950

geyslan:
@yanivagman @rafaeldtinoco @josedonizetti What do you think of the options proposed above?

rafaeldtinoco:
Do we have to pick one ? Or you want to implement all of them ? I would go for "all of them" then I consider it very very cool.

rafaeldtinoco:
I like the syntax. Can I mix things ?

--log filter:pkg=somepkg1,msg=somemsg1 ?

geyslan:
> I like the syntax. Can I mix things ?
> 
> --log filter:pkg=somepkg1,msg=somemsg1 ?

Yep. Mixing is doable.

rafaeldtinoco:
I am +1 then, looks very helpful IMO.

geyslan:
> Am I correct in understanding that the filters declare which logs _will_ appear and not which will be filtered out? If so then I think it should be the opposite. Simply put, currently I have to know which logs I want _ahead_ of time, which is odd since logs are usually discovered and not known ahead. Let's say I run in some log level, it's reasonable to assume that I would like to see all messages of that level and not have to know which to select and which not to select. If there is some message in particular that spams my system for example I would like to filter out only that one in particular.

`"filter:pkg=somepkg1,pkg3;msg=somemsg1,msg2;file=ola,ole;libbpf;file=oie,oia;pkg=pkg5"`

would show only what is set.

`"filter-out:pkg=somepkg1,pkg3;msg=somemsg1,msg2;file=ola,ole;libbpf;file=oie,oia;pkg=pkg5"`

would show all but what is set.

[...]

PR #3038, opened 2023-05-03
libbpfgo: trim suffix newline from libbpf message

geyslan:

After https://github.com/aquasecurity/libbpfgo/commit/6d17d1ac71918fcf13379ffecfbccff607204b9b, the libbpf message is passed as it is to the callback function. This causes the message to be printed with an additional newline.

So this commit trims the suffix newline from the message.

### 3. Other comments

Fix: #2950

### DIFF
--- a/pkg/cmd/initialize/callbacks.go
+++ b/pkg/cmd/initialize/callbacks.go
@@ -2,6 +2,7 @@ package initialize
 
 import (
 	"regexp"
+	"strings"
 
 	"github.com/aquasecurity/libbpfgo"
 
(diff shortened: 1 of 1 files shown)


yanivagman:
How a verifier log output would look like after this change? Will it be possible to differentiate between the different lines of the verifier error message?

geyslan:
> How a verifier log output would look like after this change? Will it be possible to differentiate between the different lines of the verifier error message?

I split the initial issue into two. This PR just fix the newline issue. The check for a verifier error msg, between other filters, will be tackled on https://github.com/aquasecurity/tracee/issues/3035.

yanivagman (approved):
LGTM

PR #3040, opened 2023-05-03
ebpf: use new libbpf syscall macros

yanivagman:
A recent libbpf version introduced syscall param macros. Let's use them.

A recent libbpf version introduced syscall param macros.
Let's use them.

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -96,17 +96,12 @@ int sys_enter_init(struct bpf_raw_tracepoint_args *ctx)
             sys->args.args[5] = READ_KERN(regs->bp);
 #endif // bpf_target_x86
         } else {
-            sys->args.args[0] = READ_KERN(PT_REGS_PARM1(regs));
-            sys->args.args[1] = READ_KERN(PT_REGS_PARM2(regs));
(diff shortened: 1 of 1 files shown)


pkg/ebpf/c/tracee.bpf.c
```diff
@@ -96,17 +96,12 @@ int sys_enter_init(struct bpf_raw_tracepoint_args *ctx)
...
-            sys->args.args[2] = READ_KERN(PT_REGS_PARM3(regs));
-#if defined(bpf_target_x86)
-            // x86-64: r10 used instead of rcx (4th param to a syscall)
-            sys->args.args[3] = READ_KERN(regs->r10);
-#else
-            sys->args.args[3] = READ_KERN(PT_REGS_PARM4(regs));
-#endif
-            sys->args.args[4] = READ_KERN(PT_REGS_PARM5(regs));
-            sys->args.args[5] = READ_KERN(PT_REGS_PARM6(regs));
+            sys->args.args[0] = PT_REGS_PARM1_CORE_SYSCALL(regs);
```
NDStrahilevitz, line 99: Does this macros take care of the x86 case above as well? If so we could get rid of that block too.

NDStrahilevitz (approved):
LGTM, let's wait for @rafaeldtinoco since I see he wanted to review too

PR #3042, opened 2023-05-04
cli: don't sort flags in help/usage messages

yanivagman:

After moving to cobra, flags are automatically sorted alphabetically. This, however, is not the desired behavior since we want the most useful flags to show first in the help message.
Disable this automatic sorting behavior.

### 2. Explain how to test it

`./dist/tracee -h`


geyslan (approved):
LGTM

PR #3047, opened 2023-05-04
add pid to memory capture and change the timestamp to epoch or relative

### LINKED ISSUE #3046: memory capture name should be pid + timestamp in epoch
Currently the name of the captured mem is the kernel nano seconds from boot.
This should be a combination of the PID of the process and the actual timestamp of the event.

AsafEitani:
fix #3046

commit 6e648284c82d8c190cb7bffdea4dbd4c41d7544d (HEAD -> mem_cap_fix)
Date:   Thu May 4 13:59:07 2023 +0300

    add pid to memory capture and change the timestamp to epoch or relative
    
    fix #3046

### 2. Explain how to test it

capture a memory region (packed file) and see the file name.


pkg/ebpf/write_capture.go
```diff
@@ -92,7 +92,14 @@ func (t *Tracee) processFileWrites(ctx context.Context) {
 					continue
 				}
 				// note: size of buffer will determine maximum extracted file size! (as writes from kernel are immediate)
-				filename = fmt.Sprintf("bin.%d", mprotectMeta.Ts)
+				if t.config.Output.RelativeTime {
```
NDStrahilevitz, line 95: Do we use this config option anywhere else besides the event's timestamp field right now?
Anyway if this is userfacing it should probably be documented.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3157,8 +3157,10 @@ int BPF_KPROBE(trace_security_file_mprotect)
             submit_mem_prot_alert_event(p.event, alert, addr, len, reqprot, prev_prot, file_info);
         }
         if (should_extract_code) {
+            u32 pid = p.event->context.task.host_pid;
             bin_args.type = SEND_MPROTECT;
-            bpf_probe_read(bin_args.metadata, sizeof(u64), &p.event->context.ts);
+            bpf_probe_read(bin_args.metadata, 4, &pid);
+            bpf_probe_read(&bin_args.metadata[4], sizeof(u64), &p.event->context.ts);
```
yanivagman, line 3163: timestamp is 8 bytes long, thus it would be better to align it by saving it before the pid and not after

pkg/ebpf/write_capture.go
yanivagman: I think that's exactly the idea. In order to correlate the event and the captured files by timestamp, they need to have the same value.
Agree that we should document this option, though not directly related to this PR

NDStrahilevitz: Not at all against the change, but if the scope of this flag is changed for this, maybe it should change for other timestamps as well (for example timestamps in arguments?).
Whatever we choose the scope of the flag needs to be clearly defined.

yanivagman: I think in that case the scope of the flag is still the same since the only reason to have the timestamp in the filename is to correlate it to the event that generated it. Not sure that's the case for timestamps in arguments.

pkg/bufferdecoder/protocol_test.go
```diff
@@ -46,5 +46,5 @@ func TestBpfObjectMetaSize(t *testing.T) {
 func TestMprotectWriteMetaSize(t *testing.T) {
 	var v MprotectWriteMeta
 	size := int(unsafe.Sizeof(v))
-	assert.Equal(t, size, int(v.GetSizeBytes()))
+	assert.Equal(t, size-4, int(v.GetSizeBytes()))
```
yanivagman, line 49: can you please explain this change?

AsafEitani: There is alignment so we fix that, you can see above on other examples that are not 8byte aligned.

yanivagman (approved):
LGTM

PR #3054, opened 2023-05-04
Feature/process tree data source

### LINKED ISSUE #2586: Process Tree: Create a process tree fed by events that works as a datasource
Process tree

1. Populated by Fork, Exec and Exit within the controller events
2. Should provide a datasource to signatures about processes/threads status.
3. Should also be populated by regular events
4. Should

Information that should be available:

1. Process related information (pid, ppid, tid, nspid, nstid, ...)
2. Information about the binary being executed
3. Information about the binary interpreter
4. Information about any real interpreter (shell, python, perl, ...)

## Issues that still need work (updates go here as they are done)

1. Process Tree should parse the signal events arguments in a generic way (issue: #3477)

2. Process Tree needs to deal with exec() called by non leader threads (issue: #3478)

3. Process Tree allows regular SchedFork,Exec,Exit events to be disabled (issue: #3479)

4. Signal events should be spinned out of regular events definitions (issue: #3480)

5. Control Plane needs refactoring for initialization arguments (issue: #3481)

6. There are some useful values that are missing from the current process tree Process struct (issue #3489):

 Execution time - the executable file information is changed in the change log during execution feed, but the time of the change is impossible to get in the current API
 Container ID
 Execution CMD
Maybe we even want the environment variables?
 Files hashes in the FileInfo struct

## PRs (related to process tree)

- https://github.com/aquasecurity/tracee/pull/3364
- https://github.com/aquasecurity/tracee/pull/3468

## Old Description (of this issue)

- A referece implementation of the process tree in tracee-rules can be found here: #1330
- Also consider the comments in #1374 for the implementation

AlonZivony:

This PR first purpose is to add data source for signatures, giving information about processes and threads.
To achieve it, user space process tree is reintroduced.

Fix #2586

### 2. Explain how to test it

The PR include unit testing for all of the logic of the process tree.

### 3. Other comments

The current process tree is only fed upon events.
We need to initialize it with the information of the system using procfs.
I don't think it has to be in the same PR, as after short amount of runtime all relevant information will be gathered.
After this will be approved I will do the effort to initialize it.


go.mod
```diff
@@ -169,4 +169,5 @@ require (
 	kernel.org/pub/linux/libs/security/libcap/psx v1.2.64 // indirect
 )
 
+replace github.com/aquasecurity/tracee/types => ./types
```
NDStrahilevitz, line 172: note: add a types PR later

yanivagman:
+1 for review

rafaeldtinoco:
I guess everyone wants to review this one, including me. =D

pkg/proctree/events_processing.go
```diff
@@ -0,0 +1,39 @@
...
+// ProcessEvent update the process tree according to arriving event
+func (tree *ProcessTree) ProcessEvent(traceeEvent *trace.Event) error {
+	switch traceeEvent.EventName {
+	case "sched_process_fork":
+		return tree.processForkEvent(traceeEvent)
+	case "sched_process_exec":
+		return tree.processExecEvent(traceeEvent)
+	case "sched_process_exit":
+		return tree.processExitEvent(traceeEvent)
+	case "exit", "init_namespaces":
```
AlonZivony, line 16: Add to excluded list all the user-mode events.
Even better - think of a way to identify non-kernel events and prevent their processing.

pkg/ebpf/events_pipeline.go
```diff
@@ -46,6 +46,12 @@ func (t *Tracee) handleEvents(ctx context.Context) {
 	eventsChan, errc = t.processEvents(ctx, eventsChan)
 	errcList = append(errcList, errc)
 
+	eventsChan, errc = t.processTree.StartProcessingPipeline(ctx, eventsChan)
```
AlonZivony, line 49: Add this stages to the events processing stage.
This will require to create a solution to get all events in the processing - to be able to fill default process data in the tree.

pkg/ebpf/events_pipeline.go
```diff
@@ -46,6 +46,12 @@ func (t *Tracee) handleEvents(ctx context.Context) {
...
+	errcList = append(errcList, errc)
+
+	eventsChan, errc = t.enrichProcess(ctx, eventsChan)
```
AlonZivony, line 52: @NDStrahilevitz do we want to move this to the processing stage or keep it where the container enrichment is?

pkg/proctree/types.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/RoaringBitmap/roaring"
+
+	"github.com/aquasecurity/tracee/pkg/utils/types"
+)
+
+// ProcessTree is a struct which follow the state of processes during runtime in container contexts.
+// The process tree is updated through Tracee's events, and is designed to overcome problems that may arise because
+// of events consumption (like handling lost events).
+type ProcessTree struct {
+	containers         types.RWMap[string, *containerProcessTree]
```
AlonZivony, line 15: The `containers` member is currently not used.
The idea is to keep track on the process trees of each container.
This is not mandatory now, so we can remove it.

pkg/ebpf/events_pipeline.go
```diff
@@ -555,6 +561,55 @@ func (t *Tracee) getStackAddresses(stackID uint32) ([]uint64, error) {
...
+	errc := make(chan error, 1)
+
+	go func() {
+		defer close(out)
+		defer close(errc)
+
+		for {
+			select {
+			case event := <-in:
+				currentProcess, err := t.processTree.GetProcessInfo(event.HostProcessID, event.Timestamp)
```
AlonZivony, line 575: We can get only partial information here, as events order is not guaranteed and the exec event might come after current one.
In the container enrichment, we wait for the information about the container to be fulfilled and only then pass the event onward.
We should consider if we want the same behavior here as well.

AlonZivony (commented):
Some thoughts and self review.

pkg/proctree/events_processing.go
```diff
@@ -0,0 +1,37 @@
...
+		return nil
+	default:
+		return tree.processDefaultEvent(traceeEvent)
+	}
+}
+
+// processDefaultEvent tries to expand the process tree in case of lost events or missing general information
```
rafaeldtinoco, line 23: It is also called by processForkEvent, processExecEvent and processExitEvent, as a "baseline" for the process creation, right ? Maybe change the function doc to this "baseline" concept ? Maybe you meant it in the "missing general information" but it wasn't clear for me.

pkg/proctree/exit_processing.go
```diff
@@ -0,0 +1,88 @@
+package proctree
+
+import (
+	"github.com/aquasecurity/tracee/signatures/helpers"
+	"github.com/aquasecurity/tracee/types/trace"
+)
+
+// processExitEvent remove references of processes from the tree when the corresponding process exit without children, or
+// if the last child process of a process exits.
+// Notice that there is a danger of memory leak if there are lost events of sched_process_exit (but is limited to the
```
rafaeldtinoco, line 12: I would say it is quite certain there will be leaks because its quite certain there will be lost events for process creation/exits. 

Could you clarify for me what happens if:

- lost a process exit for a child of a process
- parent exists and isn't cleared because of a fake existing child
- parent PID is re-used by the OS and `processDefaultEvent` would pick the "wrong process from the tree" (from the old existing process, with a different parent and outdated metadata).

I bet this assumption is wrong, maybe I haven't read the code enough, but could you clarify how is this solved ?

rafaeldtinoco: I bet this is solved at `addNewForkdProcess` (it states that "this will delete old instance if its exit was missing") but I miss how.

1. processForkEvent processes a fork event
2. it calls processDefaultEvent (it will find the old event and add a thread to it, possibly with other existing threads to already exited process)
3. processMainThreadFork or processThreadFork will call getProcess

there is a comment saying that getProcess will have errors if the existing process is a result of lost exit event. I don't see how.

pkg/proctree/events_processing.go
```diff
@@ -0,0 +1,37 @@
...
+func (tree *ProcessTree) processDefaultEvent(event *trace.Event) error {
+	process, err := tree.getProcess(event.HostProcessID)
+	if err != nil {
+		process = tree.addGeneralEventProcess(event)
+		process.addThread(event.HostThreadID)
+	} else if process.Status.Contains(uint32(hollowParent)) {
```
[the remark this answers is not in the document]
rafaeldtinoco: Also, status is a transaction flag in this case, right ? You should use a lock/mutex acquire() before status flag check AND a release() after the operation you do (fillHollowParentProcess...) because of the flag existence.

Possibly the same in all other .Status.Contains situations...

If not, there might be a race in between the conditional logic execution and the flag status existence. I searched for that in the commit adding the RW map protection but couldn't find.

rafaeldtinoco (commented):
I still need to read this again. Its lot of stuff, but I left lots of comments regarding better documentation (when we're writing things look a lot clearer than first time readers, that's why) AND asked some questions on some doubts that occurred to me.

Overall code looks nice but.. can it pass a fork bombing ? I'm skeptical about the child arrays, the missing events and tree node leftovers, and things like that, aren't you ?

I asked myself if keeping a "per thread" enriched info map and just a tree of PIDs and TIDs per namespace wouldn't be enough. Even if, sometimes, the info might not be totally accurate (in the middle of a task being created, the tree being updated and the info being acquired/updated).

Either way its a nice job, but Im a bit skeptical tbh. Not because of the code, because of the nature of the data we're keeping (too volatile with lots of interconnected references).

pkg/proctree/exit_processing.go
AlonZivony: The idea is the following:
* We can't know if we missed the exit of a process or one of its children.
* The only way to know this has happened, is if we get an event that tells us that an existing process in the tree was forked (means that its PID didn't exist before, and now the new fork process takes its place)
* So, when we get fork event we check if the process in the process tree is found as forked already.
* If it was forked, we overwrite it with the new fork information.

Now, I see your point here and thought about few things:
* We can also determine that we missed exit of a thread if we see a thread event with different PID. So this should also increase our accuracy in the data we have.
* Currently I only override the current process information, leaving its parents and children with false reference. This is really not good. What we can do here is just call the delete function to remove its information from the tree.

In the end, it is impossible to create a process tree which is safe from lost events, because you can't expect to have an accurate data source created by inaccurate data. We do a best effort here to notice if the information we get does not agree with the information we had until now.

pkg/proctree/types.go
```diff
@@ -0,0 +1,108 @@
...
+
+import (
+	"github.com/RoaringBitmap/roaring"
+)
+
+// ProcessTree is a struct which follow the state of processes during runtime in container contexts.
+// The process tree is updated through Tracee's events, and is designed to overcome problems that may arise because
+// of events consumption (like handling lost events).
+type ProcessTree struct {
+	containers         map[string]*containerProcessTree
```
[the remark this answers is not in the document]
AlonZivony: So the original idea was that we will save the links to the pid 0 of each container to have quick access to its process tree. This is currently not implemented, so I already suggested to remove this member and maybe return it in the future.

About the fact that the containers can be just a normal process tree - it is possible. Tracee currently only see one layer of containers, so it seems to me that a container process tree doesn't need to track containers, but we can maybe create this layer ourselves. Anyway, this is not implemented and I will remove it.

pkg/proctree/types.go
```diff
@@ -0,0 +1,108 @@
...
+	executed
+	generalCreated // Information that resides in every event
+	hollowParent   // Information of parent process that resides in every event
+)
+
+type timestamp int
+
+type processNode struct {
+	// TODO: Add information about the processes like opened files, network activities (like TCP connections), argv, environment variables, loader and interpreter
+	InContainerIDs  ProcessIDs
```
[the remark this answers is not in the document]
AlonZivony: I thought that it is clear - we have the PIDs in the host in the PID namespace. 
The reason that it is called IDs and not PIDs is that we are in the context of the process struct. As PID stands for "process ID", I think it makes more sense to call it ID here.
You are right that saying in container might be confusing, and I can change it.
As I wanted the process tree to monitor containers (but didn't implement it in this PR), I thought naming it container will make more sense.

pkg/proctree/events_processing.go
```diff
@@ -0,0 +1,37 @@
+package proctree
+
+import (
+	"github.com/aquasecurity/tracee/types/trace"
+)
+
+// ProcessEvent update the process tree according to arriving event
+func (tree *ProcessTree) ProcessEvent(traceeEvent *trace.Event) error {
+	switch traceeEvent.EventName {
```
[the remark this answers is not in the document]
AlonZivony: So you are right, using IDs will be better.
The idea @NDStrahilevitz suggested is to put the process tree under the `ProcessEvent` stage in the pipeline, so it will be saved in a map for callbacks there.

pkg/proctree/exit_processing.go
rafaeldtinoco: > * This is really not good. What we can do here is just call the delete function to remove its information from the tree.

Yep, even if we do more flushes than we wanted, I believe it will be "safer" considering the complexity of the updates and the possibility that we miss events (unfortunately). Its tricky, indeed.

>  We do a best effort here to notice if the information we get does not agree with the information we had until now.

YES, that was my point about being skeptical. Maybe we should do 2 things to make sure we are "good enough":

1. do some full cleanups from time to time (like a garbage collector approach, either cleaning things up if needed OR even flushing lots of references just for the sake of keeping things right -> we need to confirm if this is even needed).
2. a way to confirm this is needed would be to stress test (functional tests, not unit) the tree with lots of process creation/deletion/forks/pthreads creation. then take pictures of the tree and check how bad (or good) it is.

that would give me a good sensation about this, and unfortunately I know its a bit more of work... sorry for asking that.. I just wanted to find a way to be confident that we can trust in the tree enough (after hours/days and not only for 1 or 2 hours).

pkg/proctree/fork_processing.go
```diff
@@ -0,0 +1,155 @@
...
+		return err
+	}
+
+	newProcessInHostIDs, err := parseForkInHostIDs(event)
+	if err != nil {
+		return err
+	}
+
+	if newProcessInHostIDs.Pid == newProcessInHostIDs.Tid {
+		err = tree.processMainThreadFork(event, newProcessInHostIDs)
```
NDStrahilevitz, line 31: nit: drop the `err = ...`  in the clauses, change to a `return` and you can drop the else afterwards.

pkg/proctree/types.go
```diff
@@ -0,0 +1,114 @@
...
+	Cmd             []string
+	ExecutionBinary BinaryInfo
+	StartTime       timestamp
+	ExecTime        timestamp
+	ExitTime        timestamp
+	ParentProcess   *processNode
+	ChildProcesses  []*processNode
+	Threads         types.RWMap[int, *threadInfo]
+	IsAlive         bool
+	Status          roaring.Bitmap // Values type are processInformationStatus
```
NDStrahilevitz, line 103: Interesting, can we use this in other parts of our code? Maybe to represent exec,open flags etc.?

pkg/proctree/exit_processing.go
```diff
@@ -0,0 +1,96 @@
...
+// possible number of PIDs - 32768)
+func (tree *ProcessTree) processExitEvent(event *trace.Event) error {
+	err := tree.processDefaultEvent(event)
```
NDStrahilevitz, line 15: Just to be sure, we're using this as an opportunity to fill out the process tree from the terminating process which generated the event?

pkg/proctree/exit_processing.go
```diff
@@ -0,0 +1,96 @@
...
+
+	processGroupExit, err := helpers.GetTraceeBoolArgumentByName(*event, "process_group_exit")
+	if err != nil {
+		return err
+	}
+
+	if processGroupExit {
+		process.IsAlive = false
+		process.ExitTime = timestamp(event.Timestamp)
+		for _, tid := range process.Threads.Keys() {
```
NDStrahilevitz, line 40: Wouldn't it be better to range on `process.Threads.Values()`? Or are those copies? If the latter maybe we should rethink having the `Values()` function or add a helper which gets an array with the references directly for modification, because this seems to be about roundabout currently.

pkg/proctree/exit_processing.go
```diff
@@ -0,0 +1,96 @@
...
+			if ok && info.exitTime == 0 {
+				info.exitTime = timestamp(event.Timestamp)
+			}
+		}
+		tree.cachedDeleteProcess(process.InHostIDs.Pid)
+	}
+	return nil
+}
+
+const cachedDeadEvents = 100
```
NDStrahilevitz, line 51: Why did you chose this value specifically?

signatures/helpers/arguments_helpers.go
```diff
@@ -39,6 +39,20 @@ func GetTraceeStringArgumentByName(event trace.Event, argName string) (string, e
 	return "", fmt.Errorf("can't convert argument %v to string", argName)
 }
 
+// GetTraceeBoolArgumentByName gets the argument matching the "argName" given from the event "argv" field, casted as bool.
```
NDStrahilevitz, line 42: nit: I think it would be better, instead of expanding these helpers to reduce them to one `GetArgValueByName[T]`.
And since this was added in the signatures package but it's intended use was in the proctree, maybe the above function should be given also in the `events` package (similar to how we have `GetArg` in `parse_args.go` (which i'm not sure is the correct place for actually)).

pkg/ebpf/events_pipeline.go
NDStrahilevitz: @rafaeldtinoco WDYT, since we worked on the container enrichment together.
There is a key difference here where the latency of the network call for enrichment doesn't exist, BUT, process tree processing may not finish in time for enrichment here which is sort of similar (although we would likely reduce the chance of this by moving the processing to the process stage which is earlier in the pipeline).

pkg/proctree/events_processing.go
```diff
@@ -0,0 +1,39 @@
...
+	case "sched_process_fork":
+		return tree.processForkEvent(traceeEvent)
+	case "sched_process_exec":
+		return tree.processExecEvent(traceeEvent)
+	case "sched_process_exit":
+		return tree.processExitEvent(traceeEvent)
+	case "exit", "init_namespaces":
+		return nil
+	default:
```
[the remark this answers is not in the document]
rafaeldtinoco: Discussing about having "high priority events" for internal processing (so we can stop trying to queue events and hold specific queues of the pipeline because of a parallel action): https://github.com/aquasecurity/tracee/issues/3086#issuecomment-1558482744

At the same time, the container enriching queueing seems to be working fine (although it took us some time to get there, and many stress tests).

pkg/proctree/types.go
```diff
@@ -0,0 +1,114 @@
...
+	forked processInformationStatus = iota
+	executed
+	generalCreated // Information that resides in every event
+	hollowParent   // Information of parent process that resides in every event
+)
+
+type timestamp int
+
+type processNode struct {
+	// TODO: Add information about the processes like opened files, network activities (like TCP connections), argv, environment variables, loader and interpreter
```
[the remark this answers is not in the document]
rafaeldtinoco: Yep, I had a talk with @AlonZivony about an idea: to keep the proctree as a versioned tree (snapshotted) each time there is a delta on it. This was also an idea to always have a historical "point of truth" of the entire tree (so the changes in proc tree would also be auditable AND we could use it as an artifact).

With that in mind, we expanded the idea saying that everything related to a process could be a "snapshot" of the proctree node (opened files, opened sockets, DNS queried, etc).

Good discussion, etc, to have more ideas for the "artifact saving per context triggering" feature we once imagined.

Either way, without too much bikeshedding, an idea of a "versioned proc tree" (with snapshots per node added/removed) is a nice idea that allows you to reconstruct the tree at any point in time (and would make locking much simpler).

pkg/proctree/types.go
rafaeldtinoco: I liked the roaring bitmap package as well =D.

rafaeldtinoco:
We're branching v0.16.0 today. I'm afraid this change won't make it. Scheduling for v0.17.0 now.

pkg/proctree/events_processing.go
```diff
@@ -0,0 +1,37 @@
+package proctree
+
+import (
+	"github.com/aquasecurity/tracee/types/trace"
+)
+
+// ProcessEvent update the process tree according to arriving event
+func (tree *ProcessTree) ProcessEvent(traceeEvent *trace.Event) error {
```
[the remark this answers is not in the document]
AlonZivony: So I think in the end that this is the processing of the process tree.
It is not `tracee.ProcessEvent`, but `procTree.ProcessEvent`.
I think they are both doing the same role, each on its own package.
I don't think this is really confusing as they are methods.

pkg/proctree/types.go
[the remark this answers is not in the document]
AlonZivony: So the InContainerIDs were meant to be just namespace IDs.
This is not relevant in the implementation anymore, but I though on clarifying this.

pkg/ebpf/events_pipeline.go
AlonZivony: So I resolve this as this will be solved when we move the control plane events

pkg/proctree/exit_processing.go
AlonZivony: This was solved now using a capped size LRU for garbage collecting nodes.
We can improve it to also limit the amount of dead ancestors we keep for a node, but this could come later on.

pkg/proctree/exit_processing.go
AlonZivony: So I have no reason, just an hunch.
The whole size of the cache is a problematic issue, as we can't predict the load.
The number here is just me trying to put a limit to the amount of memory used for dead nodes.

pkg/proctree/process.go
```diff
@@ -0,0 +1,375 @@
...
+		execInfo:       types.NewTimeSeries[procExecInfo](procExecInfo{}),
+	}, nil
+}
+
+// setGenInfoFromEventProtected fill the general info of the process (information
+// of the event given by every process received from it) in a way that it is filled
+// only once in an efficient way to reduce performance penalty.
+// This method uses the process lock to reduce unnecessary locking if not needed. Make sure to unlock
+// the lock before using this.
+func (p *processNode) setGenInfoFromEventProtected(event *trace.Event) {
```
yanivagman, line 66: nit: usually Gen is used in place of generate, which may be confusing in this case.
Consider using general instead

pkg/proctree/types.go
```diff
@@ -0,0 +1,50 @@
...
+	Ppid            int
+	NsPpid          int // TODO: is this redundant?
+	UserId          int
+	ContainerId     string
+	Cmd             []string
+	ExecutionBinary FileInfo
+	StartTime       int // First thread fork time. TODO: Can we use the mian thread start time instead?
+	ExecTime        int // Last execve call time
+	ExitTime        int
+	ExistingThreads []int
```
yanivagman: I think calling it `Threads` is enough, or if you want to be consistent with ChildrenIds then call it `ThreadIds`

pkg/proctree/types.go
```diff
@@ -0,0 +1,50 @@
...
+	ChildrenIds     []int
```
yanivagman: It may be confusing that children refer to child processes. I would be more explicit here - `childProcessIDs`

pkg/proctree/types.go
```diff
@@ -0,0 +1,50 @@
...
+	IsAlive         bool
+}
+
+// ThreadInfo is the user facing representation of a thread data at a specific time.
+type ThreadInfo struct {
+	HostId        int
```
yanivagman: No need for Host if you call the below field NsId. Making it Id will also be more consistent with Id in the ProcessInfo

pkg/proctree/types.go
```diff
@@ -0,0 +1,50 @@
...
+	NsId          int
+	HostProcessID int
```
yanivagman: same here

yanivagman (changes requested):
Nice work @AlonZivony !
This is a first pass I did on the first four commits.
I will do another pass after pushing the fixes

pkg/ebpf/tracee.go
```diff
@@ -130,9 +132,9 @@ func (t *Tracee) Stats() *metrics.Stats {
 func GetEssentialEventsList() map[events.ID]eventConfig {
 	// Set essential events
 	return map[events.ID]eventConfig{
-		events.SchedProcessExec: {},
-		events.SchedProcessExit: {},
-		events.SchedProcessFork: {},
+		events.SchedProcessExec: {submit: 0xFFFFFFFFFFFFFFFF},
+		events.SchedProcessExit: {submit: 0xFFFFFFFFFFFFFFFF},
+		events.SchedProcessFork: {submit: 0xFFFFFFFFFFFFFFFF},
```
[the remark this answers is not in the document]
rafaeldtinoco: Agreed. Should be moved to the control plane.

pkg/utils/types/rwmaps.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"sync"
+
+	"golang.org/x/exp/maps"
+)
+
+// RWMap is a map envelopment for concurrent map use
+type RWMap[K comparable, V any] struct {
+	m     map[K]V
+	mutex *sync.RWMutex
```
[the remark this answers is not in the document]
rafaeldtinoco: IMO, We should follow object-oriented encapsulation principle, instead of exporting mutexes for external (to interface) users (and making a consumer responsbility for the object correctness). If, in the future, we want to change the object to have another synchronization mechanism, we would have to change all code that uses it (that is exactly what we are trying to avoid with encapsulation). The correct approach for the case you describe would be to provide a method that has a callback argument. The method would call the function while maintaining the safety of the context.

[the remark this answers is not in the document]
rafaeldtinoco: Resuming, its either a "read-copy-update" approach (like the way its implemented NOW, where the user picks an item, works on it, and sets it back) OR a "transactional" approach (with atomicity, consistency, isolation) with a callback.

We have to think of the "types" being used in the map (its not the same depending on the types):

1. If the types are "simple" types (atomic uint, atomic int, pointer to a struct that has its internal synchronization) then its ok to be like that. Lets assume you're doing a map of an `example Struct{}`. You would get an instance of Example (one instantiated object) and rely on its own methods to change it (and it would have its own internal synchronization mechanism).

2. Now, if the types "complex" types (like a map of map, or map of array, non instantiated objects), then you are in trouble and you would require the callback approach, for example.

@yanivagman,

With that said, now you get why I proposed all the event definitions to be instantiated: to guarantee encapsulation, isolation and atomicity. By having everything instantiated we can always encapsulate synchronization mechanisms and make the users agnostic on how its done. I'm sorry @NDStrahilevitz did not like that as well, but the alternative to that model isn't great (like you can see in this simple example).

pkg/utils/types/time_series.go
```diff
@@ -0,0 +1,83 @@
...
+	return co.states[relevantStateIndex]
+}
+
+// ChangeDefault change the value assumed before first recorded change
+func (co *TimeSeries[T]) ChangeDefault(defaultVal T) {
+	co.mu.Lock()
+	defer co.mu.Unlock()
+	co.states[0].Val = defaultVal
+}
+
```
rafaeldtinoco, line 66: // getRelevantStateIndex searches for the index of the state in the states slice of the TimeSeries object that has the closest start time to the given time parameter. It uses binary search to find the index.

pkg/utils/types/time_series.go
```diff
@@ -0,0 +1,83 @@
...
+	return &TimeSeries[T]{
+		states: []State[T]{defaultState},
+	}
+}
+
+// AddState add a change in the value - from given time until next change, the value is the new one.
+func (co *TimeSeries[T]) AddState(newState State[T]) {
+	co.mu.Lock()
+	defer co.mu.Unlock()
+	insertIndex := co.getRelevantStateIndex(newState.StartTime) + 1
```
rafaeldtinoco, line 38: You're always assuming that the `StartTime` of the new state is bigger than the of the last state. Should you try to guarantee that (and possibly return an error ?).

pkg/utils/types/time_series.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (co *TimeSeries[T]) getRelevantStateIndex(time Timestamp) int {
+	low := 0
+	high := len(co.states) - 1
+	closestIndex := 0
+
+	for low <= high {
+		mid := low + (high-low)/2
+
+		if co.states[mid].StartTime <= time {
```
rafaeldtinoco, line 75: You're assuming that `states []State[T]` is always sorted here. That isn't guaranteed by golang.

pkg/utils/types/time_series.go
```diff
@@ -0,0 +1,83 @@
...
+		states: []State[T]{defaultState},
+	}
+}
+
+// AddState add a change in the value - from given time until next change, the value is the new one.
+func (co *TimeSeries[T]) AddState(newState State[T]) {
+	co.mu.Lock()
+	defer co.mu.Unlock()
+	insertIndex := co.getRelevantStateIndex(newState.StartTime) + 1
+	stateList := make([]State[T], len(co.states)+1)
```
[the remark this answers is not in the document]
rafaeldtinoco: TODO: This might not be very efficient if called frequently. We could use a linked list OR, at least, allocate a bigger slice and only creating a new only when needed.

pkg/utils/types/time_series.go
```diff
@@ -0,0 +1,83 @@
...
+	defer co.mu.RUnlock()
+	relevantStateIndex := co.getRelevantStateIndex(time)
+	return co.states[relevantStateIndex]
+}
+
+// ChangeDefault change the value assumed before first recorded change
+func (co *TimeSeries[T]) ChangeDefault(defaultVal T) {
+	co.mu.Lock()
+	defer co.mu.Unlock()
+	co.states[0].Val = defaultVal
```
rafaeldtinoco, line 64: Should you check if there is no state recorded yet before changing the value ?

types/datasource/proctree.go
```diff
@@ -0,0 +1,66 @@
...
+	Pid   int
+	Mount int
+}
+
+// ProcessLineage is a representation of a process and its ancestors until the oldest ancestor
+// known in the tree.
+// The lineage is only relevant for the container the process resides in.
+type ProcessLineage []ProcessInfo
+
+type ProcKey struct {
```
rafaeldtinoco, line 52: nit: add a comment in these types: key for what ?

pkg/proctree/tree.go
```diff
@@ -0,0 +1,297 @@
...
+	// You might even want more than 32768 to allow relevant information of dead processes to be
+	// available after their exit.
+	MaxProcesses int
+	// Max amount of threads nodes to allow in memory.
+	// Too small value here might result missing information in the tree and inconsistency.
+	// Default recommended value is 32768, as this is the max amount of TIDs in the system.
+	// You might even want more than 32768 to allow relevant information of dead threads to be
+	// available after their exit.
+	MaxThreads int
+	// Max size of cache for processes removing after exit - translate to the delay between process
```
rafaeldtinoco, line 43: What is the metric here ? Max size in items ? Because it is said to be "translated to the delay between process exit to its removal from the tree", so is it seconds (as its a delay ?). I think its better to clarify this comment.

pkg/proctree/tree.go
```diff
@@ -0,0 +1,297 @@
...
+
+// ProcessTree is a struct which follow the state of processes during runtime of a system.
+// The process tree is updated through Tracee's events, and is designed to overcome problems that
+// may arise because of events consumption (like handling lost events).
+// For more information on the logic, please go to the package documentation.
+type ProcessTree struct {
+	processes          types.RWMap[int, *processNode]
+	threads            types.RWMap[int, *threadNode]
+	processesGC        *lru.Cache[nodeUniqueId, *processNode]
+	threadsGC          *lru.Cache[nodeUniqueId, *threadNode]
```
rafaeldtinoco, line 59: GC stands for ? Add a comment to the end of the declaration explaining ? Wasn't clear to me at a first glance.

pkg/ebpf/tracee.go
```diff
@@ -279,6 +281,19 @@ func New(cfg config.Config) (*Tracee, error) {
 		}
 	}
```
rafaeldtinoco, line 283: // Configure a new process tree to ... if ...

pkg/proctree/tree.go
```diff
@@ -0,0 +1,297 @@
...
+	return tree, nil
+}
+
+// getProcess get the process node from the process tree if exists, and return error if not.
+func (tree *ProcessTree) getProcess(hostProcessID int) (*processNode, error) {
+	process, ok := tree.processes.Get(hostProcessID)
+	if !ok {
+		return nil, fmt.Errorf("no process with given Id is recorded")
+	}
+	_, _ = tree.processesGC.Get(process.GetUniqueId())
```
rafaeldtinoco, line 112: // keep cache hot ?

pkg/proctree/tree.go
```diff
@@ -0,0 +1,297 @@
...
+	// Too small value here might result processes missing from tree after their exit, before
+	// users were able to query the process information because of the delay between the feeding
+	// of the tree and the consumption of its information.
+	MaxCacheDelete int
+}
+
+// ProcessTree is a struct which follow the state of processes during runtime of a system.
+// The process tree is updated through Tracee's events, and is designed to overcome problems that
+// may arise because of events consumption (like handling lost events).
+// For more information on the logic, please go to the package documentation.
```
rafaeldtinoco, line 54: Where is the package documentation ? I believe the logic should be explained interleaved with the code, so its easier to understand for first readers, IMO.

pkg/proctree/tree.go
```diff
@@ -0,0 +1,297 @@
...
+	if err != nil {
+		return nil, err
+	}
+	threadsGC, err := lru.NewWithEvict[nodeUniqueId, *threadNode](
+		config.MaxThreads,
+		func(id nodeUniqueId, node *threadNode) {
+			err = tree.deleteThread(node)
+			if err != nil {
+				logger.Errorw("delete thread from tree", "err", err.Error())
+			}
```
rafaeldtinoco, line 94: Could you explain what each cache does and why they exist ? If this is explained somewhere else, then I believe it should be moved here IMO.

pkg/proctree/tree.go
```diff
@@ -0,0 +1,297 @@
...
+			if err != nil {
+				return nil, err
+			}
+		}
+		process.mutex.Lock()
+		process.connectParent(parentProcess)
+		process.mutex.Unlock()
+		parentProcess.mutex.Lock()
+		parentProcess.connectChild(process)
+		parentProcess.mutex.Unlock()
```
rafaeldtinoco, line 261: Isn't this a clear case of a transaction ?

 Where parent and child both have dependency on each other ? I would assume intrinsically, in a first read, you would have to:

1. lock parent
2. lock process
3. connect process to parent
4. conenct parent to process
5. unlock process
6. unlock parent

this way if other childs are to be added to the parent process they would only be able to once parent is "consitent" (thats what defines a transaction consistency).

pkg/proctree/export.go
```diff
@@ -0,0 +1,214 @@
+package proctree
+
+/*
```
rafaeldtinoco, line 3: nit: keep // in all comments in the project please (styling checks)

pkg/proctree/export.go
```diff
@@ -0,0 +1,214 @@
...
+Package proctree is used to create and maintain a process tree using Tracee's events.
+
+The process tree exports to the user only read operations - the user can query for a processes and
```
rafaeldtinoco, line 6: nit: query for a "processes" (plural), nits fix.

pkg/proctree/export.go
```diff
@@ -0,0 +1,214 @@
...
+*	Limit the amount of nodes it keeps, to avoid leakage when missing exit events.
+*	Take information from all events, even those which are not related to process and thread life
+	cycle.
+*	Try cleaning processes with PID which is found to be reused by new fork events.
+*	Initializing parent processes nodes, even with almost no information, to establish connections
+	between known nodes.
+*	The tree is *not* trying to create one big tree, but instead has the forest architecture.
+	As it doesn't have all the history, it can't connect all nodes, and so it connects only nodes
+	it knows are connected.
+*/
```
rafaeldtinoco, line 39:
```suggestion
// Process Tree Documentation
//
// The process tree only exports read operations to the user - the user can query for processes and
// threads information. However, the process tree itself is rather complex and should not be changed
// by unsupported methods.
//
// Currently, the only supported way to update and change the process tree is via events. The
// process tree is aware of Tracee's events and knows how to update itself using these events. This
// can be done either using a pipeline or directly using the appropriate method.
//
// # Synchronization
//
// The tree's usage is not synchronized. Process information may be requested some time after it has
// been updated or even deleted. To address this issue, the tree:
//
// 1. Makes processes available even after they have exited and should have been deleted for some
//    time.
// 2. Exports processes and threads information according to a specific time, instead of providing
//    all the information and leaving the user to process it.
//
// # Events feeding and edge case handling
//
// The overall design of the tree takes into consideration that events are not guaranteed to be
// ordered, nor are they guaranteed to arrive at all. Moreover, the tree needs to account for the
// fact that it starts running in the middle of the system runtime. As a result, it misses some
// process fork, exec, and exit events.
//
// To withstand these issues, the tree does the following:
//
// 1. Limits the number of nodes it keeps to avoid leakage when missing exit events.
// 2. Gathers information from all events, even those that are not related to process and thread
//    life cycles.
// 3. Attempts to clean processes with a PID that is found to be reused by new fork events.
// 4. Initializes parent process nodes, even with minimal information, to establish connections
//    between known nodes.
// 5. The tree does *not* attempt to create one large tree, but instead adopts a forest
//    architecture. Since it does not have all the history, it can't connect all nodes, so it only
//    connects nodes it knows are connected.
//
```

pkg/proctree/export.go
```diff
@@ -0,0 +1,214 @@
...
+
+import (
+	gocontext "context"
+
+	"github.com/aquasecurity/tracee/pkg/logger"
+	"github.com/aquasecurity/tracee/pkg/utils/types"
+	"github.com/aquasecurity/tracee/types/datasource"
+	"github.com/aquasecurity/tracee/types/trace"
+)
+
```
rafaeldtinoco, line 49: ProcessTree struct should be here, together with `NewProcessTree()` as they are exported. Also, it might be better to have the type declaration closer to the documentation above.

rafaeldtinoco: Also, I'm not sure I like the tree.go + export.go. I would keep all in the same file (its easier for readers to understand what is going on, as you already have the ProcessTree methods also split into `events_processing.go`, `exec_processing.go` and ` exit_processing.go` (those do make sense to have).

pkg/proctree/process.go
```diff
@@ -0,0 +1,383 @@
...
+
+	"github.com/aquasecurity/tracee/pkg/utils/types"
+	"github.com/aquasecurity/tracee/types/trace"
+)
+
+// processNode is a node in the process tree representing a process.
+// Its purpose is to keep the information of the process, and the connections to other processes
+// and threads nodes.
+// The current implementation is using mutex, so all internal members are assumed to be protected
+// by it (so are not necessarily thread-safe types)
```
rafaeldtinoco, line 17: You want to say here you're using mutex as a big lock for all fields within this type.

pkg/proctree/process.go
```diff
@@ -0,0 +1,383 @@
...
+type processNode struct {
+	// TODO: Add information about the processes like opened files,
+	//  network activities (like TCP connections), argv, environment variables,
+	//  loader and interpreter
+	id             int
```
rafaeldtinoco, line 22: is this the PID (process id) or some other id ? maybe rename it to "pid" ?

pkg/proctree/process.go
```diff
@@ -0,0 +1,375 @@
...
+			p.mutex.Unlock()
+		},
+	)
+}
+
+// setGeneralInfoProtected is used to fill general information of a process
+// only once, but when the information cannot be retrieved from an event like
+// with setGenInfoFromEventProtected.
+// This method is not protected by locks.
+func (p *processNode) setGeneralInfoProtected(
```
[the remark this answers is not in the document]
rafaeldtinoco: It does not make any sense to me to have protected and unprotected versions of the same change to internal fields. Also, if you already have a "Once" guaranteeing that the operation will be done just once, why not to have only a protected version ? There is no free lunch when it comes to concurrency (its either an atomic change or a protected (mutex,futex,spins) transactional change). We can't mix both.

pkg/proctree/process.go
```diff
@@ -0,0 +1,383 @@
...
+	if checkTime >= exitTime {
+		return false
+	}
+	forkTime := p.getForkTime()
+	return checkTime >= forkTime
+}
+
+// setExitTime sets the process's exit time
+func (p *processNode) setExitTime(exitTime types.Timestamp) {
+	p.exitTime = exitTime
```
rafaeldtinoco, line 150: So, your object is either thread-safe or it is not. Like in this case, for example. What happens if exitTime is being written while another thread is reading the integer? Lots of things might happen:

- if the integer (hardly the case) is split among 2 CPU cache lines, you might read just half of it (while the other half hasn't been updated, which would truncate the value)
- some threads might read one value, some others might read another (golang's memory model won't guarantee ordering)

If you want to keep getters that aren't protected by your big lock (mutex) than, at least, you've got to make them atomic values (with atomic loads/stores) so they can be used concurrently by simultaneous threads (OS or runtime's threads).

rafaeldtinoco: The same happens with all other cases where you aren't protecting values by holding the mutex. The mutex is a "primitive" and will guarantee that all CPUs are coherent (because of barriers used by locking primitives) and all values are atomically changed (or transactions, when changing more than one field, are atomically started/ended). If you don't use a mutex, you *have* to use atomic Loads/Stores for thread-safety.

pkg/proctree/process.go
```diff
@@ -0,0 +1,383 @@
...
+		StartTime: execTime,
+		Val:       info,
+	}
+	p.execInfo.AddState(execState)
+}
+
+// setDefaultExecInfo change the execution information assumed for the process before its first
+// execution received.
+func (p *processNode) setDefaultExecInfo(info procExecInfo) {
+	p.execInfo.ChangeDefault(info)
```
rafaeldtinoco, line 213: Like here, for example. You have the internal to execInfo mutex, protecting the execInfo internal fields, but you still have a Load() operation for the pointer when you de-reference `p.execInfo`. So the correct thread-safety operation here would either use the mutex or use an atomic Load for a `*Struct{}` (*procExecInfo in this case). Just like an int64 should have an atomic Load()/Store(), etc.

rafaeldtinoco: Then we can argue on fields that are changed runtime and the fields that aren't changed runtime. If that is the case (fields that are never changed runtime), we should make it very explicity they don't need to be atomic because they're **immutable**.

pkg/proctree/process.go
rafaeldtinoco: I have added other comment somewhere saying that if the values are **immutable** then its ok to have getters without mutexes or atomic loads (but then there couldn't exist a setter for these values, they would have to be set in the New() function only and never changed.

pkg/proctree/process.go
```diff
@@ -0,0 +1,383 @@
...
+
+// amountOfChildren return the amount of processes registered as children of current process
+func (p *processNode) amountOfChildren() int {
+	return len(p.childProcesses)
+}
+
+// hasChildren return if the current process has children registered to it
+func (p *processNode) hasChildren() bool {
+	return p.amountOfChildren() != 0
+}
```
rafaeldtinoco, line 332: In here.. the len(p.childProcesses) read the metadata for the map (to get its length). At the same time, you have `maps.Values` walking the map to construct a slice (and its not thread-safe). Those 2 aren't thread-safety among them (lets say you can `amountOfChildre()` in 2 different threads). Now the question: "Oh but I won't be calling methods from 2 different threads because its always coming from the control plane". Then there is no need for thread-safety at all. I strongly think all our structs/classes should be thread-safe for all its operations to all its fields/transactions... arguable and I know others disagree.

pkg/proctree/thread.go
```diff
@@ -0,0 +1,230 @@
...
+			t.mutex.Unlock()
+		},
+	)
+}
+
+// setGeneralInfoProtected is used to fill general information of a thread
+// only once, but when the information cannot be retrieved from an event like
+// with setGenInfoFromEventProtected.
+// This method is not protected by locks.
+func (t *threadNode) setGeneralInfoProtected(
```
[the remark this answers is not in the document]
rafaeldtinoco: Ditto as well. And all other related (to thread-safety) comments made.

pkg/proctree/fork_processing.go
```diff
@@ -0,0 +1,176 @@
...
+			mount: event.MountNS,
+		},
+		processExitTime,
+	)
+	newThread.mutex.Unlock()
+	return nil
+}
+
+// parseForkHostIds gets the new forked process Ids in the host PId namespace
+func parseForkHostIds(event *trace.Event) (taskIds, error) {
```
rafaeldtinoco, line 104: I believe everything that is "PID" should be named PID and not ID. It makes things confusing reading ID when we are able to know if its the PID/TID right away by reading the function/variable name.

pkg/proctree/fork_processing.go
```diff
@@ -0,0 +1,176 @@
...
+	// event
+	if err != nil ||
+		newProcess.getForkTime() != 0 {
+		tree.immediateRemoveProcessFromTree(newInHostIds.Pid)
+		newProcess, err = tree.newForkedProcessNode(newInHostIds, newInNsIds)
+		if err != nil {
+			return err
+		}
+	}
+	newProcess.mutex.Lock()
```
rafaeldtinoco, line 58: This is a no for me. The synchronization mechanism has to be controled by the interface/class/struct that contains it. Or else things will get out of control in no time (and the encapsulation characteristic of object orientation vanishes).

pkg/proctree/fork_processing.go
```diff
@@ -0,0 +1,176 @@
...
+	tree.copyParentBinaryInfo(types.Timestamp(event.Timestamp), newProcess)
+
+	newProcess.setGeneralInfoProtected(
```
rafaeldtinoco, line 61: This is also a no for me. The struct/class is either thread-safe or not. User can't chose what method to pick (I could use the same struct else where and use an unprotected method and I would brake your code).

pkg/proctree/fork_processing.go
```diff
@@ -0,0 +1,176 @@
...
+// copyParentBinaryInfo copies the binary information of the parent node information if exist to
+// the given process node.
+// This is useful for forked processes, as they have the same binary as parent process until exec
+// is invoked.
+func (tree *ProcessTree) copyParentBinaryInfo(time types.Timestamp, p *processNode) {
+	fatherProcess := p.getParent()
+	if fatherProcess == nil {
+		return
+	}
+	fatherProcess.mutex.RLock()
```
rafaeldtinoco, line 172: ditto on external (to type) synchronization control and encapsulation.

rafaeldtinoco:
Have gone through 50% of the changes, there are many comments to be discussed, etc... so I stopped and will resume once discussions are done. Cheers!

pkg/utils/types/rwmaps.go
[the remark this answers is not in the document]
AlonZivony: So I agree with Rafael that I use this map to handle instanced objects, which are responsible to manage themselves.
We can have the map expose a callback function, which will be protected by `Lock()` for its entire operation.
@rafaeldtinoco maybe we should have 2 different types for it though, to make sure that the user is aware of what its use-case is and what best suit it.
I think that either way further PR can introduce the new callback method or the new type, so this conversation does not affect this PR.

pkg/utils/types/time_series.go
```diff
@@ -0,0 +1,83 @@
...
+import (
+	"sync"
+)
+
+// Timestamp is a regulation of time types
+type Timestamp int
+
+// State represent a change of value - the new value and the time of the change
+type State[T any] struct {
+	StartTime Timestamp
```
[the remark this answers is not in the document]
AlonZivony: Good point.
I wanted to standardize the use of timestamp, as I am not sure how in the long term we will want to handle it.
I will crate a get and set functions.

pkg/proctree/events_processing.go
```diff
@@ -0,0 +1,50 @@
...
+	parent := process.getParent()
+	process.mutex.RUnlock()
+	if parent == nil {
+		_, err = tree.generateParentProcess(event.HostParentProcessID, event.ParentProcessID, process)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
AlonZivony: So if you will go into details, all the functions here are very basic and almost lockless.
For the get function of the processes and thread, we have to read-lock the mutexes of the map, but all the other things are optimized.
We use the `setGenInfoFromEventProtected` to initialized process information only if missing, without using locks (uses `sync.Once` object).
The `addGeneralEventThread` only create new thread if the thread is missing, else does nothing.
And again, we create a parent only if it is missing.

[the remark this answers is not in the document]
AlonZivony: @yanivagman notice that we currently only use the three basich `schedX` events, so this is not happening for all Tracee's events.

pkg/proctree/exit_processing.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/aquasecurity/tracee/signatures/helpers"
+	"github.com/aquasecurity/tracee/types/trace"
+)
+
+// ProcessExitEvent remove references of processes from the tree when the corresponding process
+// exit without children, or if the last child process of a process exits.
+// Notice that there is a danger of memory leak if there are lost events of sched_process_exit
+// (but is limited to the possible number of PIds - 32768).
+func (tree *ProcessTree) ProcessExitEvent(event *trace.Event) error {
+	err := tree.processGeneralEvent(event)
```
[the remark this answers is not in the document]
AlonZivony: I wanted to enable using the specific events processing functions directly, which is the current use-case.
The general `ProcessEvent` is not used for performance reasons.
Another problem with doing what you propose is the fact that there are events which should not be processes, for example the `init_namespaces` event. Their data is not matching the real thread and process data, and will corrupt the tree.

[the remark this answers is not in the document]
AlonZivony: If you agree that this is not a concern please resolve the conversation

pkg/proctree/exit_processing.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		return err
+	}
+	process, err := tree.getProcess(event.HostProcessID)
+	if err != nil {
+		return fmt.Errorf("process was inserted to the treee but is missing right after")
```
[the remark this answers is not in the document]
AlonZivony: Why not?
It must be in the tree in this point, but it is missing now.
This is a case that should not happen, and I am clarifying what is odd here.

pkg/proctree/exit_processing.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	thread, err := tree.newProcessThreadNode(process, event.HostThreadID)
```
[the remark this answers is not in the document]
AlonZivony: Hmm you are incorrect, but the name is misleading.
The function will not create a new process if one already exist, but will return the existing one.
I will need to clarify this, and also do some minor optimization to the implementation as I see something that can be implemented better.

pkg/proctree/tree.go
```diff
@@ -0,0 +1,297 @@
...
+}
+
+// newProcessThreadNode add a new thread to a process, and also sign the thread in the tree.
+// It will return existing one if there is one with the same ID.
+func (tree *ProcessTree) newProcessThreadNode(process *processNode, tid int) (*threadNode, error) {
+	newThread, err := process.addThreadBasic(tid)
+	if err != nil {
+		return nil, err
+	}
+	err = tree.setThread(newThread)
```
AlonZivony, line 295: We should probably avoid setting the thread in the map if it exists already...

pkg/proctree/tree.go
```diff
@@ -0,0 +1,297 @@
...
+		return nil, err
+	}
+	tree.threads.Set(tid, thread)
+	tree.threadsGC.Add(thread.GetUniqueId(), thread)
+	return thread, nil
+}
+
+// newProcessThreadNode add a new thread to a process, and also sign the thread in the tree.
+// It will return existing one if there is one with the same ID.
+func (tree *ProcessTree) newProcessThreadNode(process *processNode, tid int) (*threadNode, error) {
```
AlonZivony, line 290: The name is misleading - it seams like the thread will always be created, not only if it doesn't exist.
I think a better name would be `getOrCreateNewProcessThreadNode` or `createIfMissingProcessThreadNode`.

pkg/proctree/export.go
```diff
@@ -0,0 +1,200 @@
...
+	relevantTime := time
+	for i, p := range pList {
+		lineage[i] = p.export(types.Timestamp(relevantTime))
+		relevantTime = lineage[i].StartTime
+	}
+	return lineage, nil
+}
+
+// StartProcessingPipeline create a goroutine that feeds the process tree from the incoming events.
+func (tree *ProcessTree) StartProcessingPipeline(
```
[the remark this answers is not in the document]
AlonZivony: I will just remove this function as it is not used now

pkg/proctree/export.go
```diff
@@ -0,0 +1,200 @@
...
+		ChildrenIds:     childrenIds,
+		IsAlive:         p.isAlive(time),
+	}
+}
+
+func (t *threadNode) export(time types.Timestamp) ThreadInfo {
+	t.mutex.RLock()
+	defer t.mutex.RUnlock()
+
+	process := t.getProcess()
```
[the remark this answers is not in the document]
AlonZivony: Hmm I can, but it seems to me like a waste.
Getting the process is inexpensive, and getting its PID is also lockless.
So why should we save the PID twice in this case?

pkg/proctree/process.go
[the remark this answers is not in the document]
AlonZivony: So the reason here is that I need to lock the node when applying the changes.
I wanted to lock the node from inside the `Once` function, to avoid locking the node for nothing after the first call.
However, when processing fork event, I lock the mutex already.
I don't want to lock it twice, so I created another function to fill the data if the mutex is locked from outside.
What do you propose to do here to avoid unnecessary locking twice in fork?

pkg/utils/types/time_series.go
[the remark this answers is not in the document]
AlonZivony: > Shouldn't this be []*State[T] ?

I thought no, because the `val` can be just a reference so doing double dereference seems weird to me.
However, if the timestamp will be developed to be a more unique time then you are right.

> TODO: This might not be very efficient if called frequently. We could use a linked list OR, at least, allocate a bigger slice and only creating a new only when needed.

However, you will pay for having a linked list by having inefficient search. You won't be able to do binary search in that case for example.
You might be right that we should allocate bigger sized array, I need to think about it.

pkg/utils/types/time_series.go
AlonZivony: I doesn't get what you mean.
I guarantee it.

pkg/utils/types/time_series.go
AlonZivony: No, because the object will always be created with a default state as its first state.

pkg/proctree/tree.go
AlonZivony: I want here to update its place in the LRU.
I will add a comment on it.

pkg/proctree/tree.go
AlonZivony: The package documentation is written currently in the `export.go` file.
I don't want to have duplicate of documentation, because it always leads to one of them is not up-to-date with the other.

pkg/proctree/tree.go
AlonZivony: So I wanted to avoid locking both locks at once, to avoid deadlocks in the long term.
However, you are right here that this is crucial for them to be consistent with each other.
I will change it.

pkg/proctree/process.go
AlonZivony: So it is the PID, as PID stands for Process ID and this is a process node.
@yanivagman do you think this should be changed to `pid`?

pkg/proctree/process.go
AlonZivony: The current design puts the responsibility for locking and unlocking nodes on the Tree.
No method of the nodes lock the mutex, except for the `setGenInfoFromEventProtected` method.
To conclude - the process and thread nodes are not thread-safe. So, the comments here are not relevant in this context.
If you feel that some method of the process tree is not using mutexes as needed, than please comment there.

pkg/proctree/process.go
AlonZivony: Again, the problem you point here is not relevant because this is not a thread-safe operation.
I think that maybe you are confused because the `TimeSeries` object is thread-safe, and I use it. I implemented it that way so it will be safe to use in all places, not specifically in the current process tree nodes which are not thread-safe.
The face I use a thread-safe object in my nodes doesn't mean the nodes are threa-safe.
I will add a documentation for it on the nodes so it will be clear that they are not thread-safe.

pkg/proctree/fork_processing.go
AlonZivony: However, the `processNode` and `threadNode` are both unexported structs.
They wouldn't be used by users outside of the tree scope.
The are not thread-safe.
The tree itself is thread-safe, and implement all of the operations needed to make it possible.

pkg/utils/types/time_series.go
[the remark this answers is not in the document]
rafaeldtinoco: > I thought no, because the val can be just a reference so doing double dereference seems weird to me. However, if the timestamp will be developed to be a more unique time then you are right.

If we have:

```go
type State[T any] struct {
    StartTime Timestamp
    Val *T
}
```
we would enforce Val to always be a pointer (or get a compile-time error). This way you would guarantee no shallow copies are done (which is likely the case you need as Val is always a pointer to some instance).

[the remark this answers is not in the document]
rafaeldtinoco: > However, you will pay for having a linked list by having inefficient search. 

Sure, a binary tree then (considering StartTime can always be sorted and used as the node's keys). Might be faster than creating slices over and over (or having to manage a bigger one) and will allow faster searches (search will continue being binary and same or better complexity).

pkg/utils/types/time_series.go
rafaeldtinoco: You are correct. the Starttime is always increasing number over time and you're adding at specific indexes.

pkg/utils/types/time_series.go
rafaeldtinoco: nm, I see `[]State[T]` is always sorted due to startime be always increasing number (hopefully =D).

pkg/utils/types/time_series.go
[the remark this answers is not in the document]
AlonZivony: Do you think I should just use the `time.Time` type, which is the standard of Go?
We can initialize it pretty easily using `time.Unix(0, event.Timestamp)` and return it to nano-seconds using `t.UnixNano()`.
The only drawback here is that it stores x3 more information than the current type, but I guess it is not such a big deal.

pkg/utils/types/time_series.go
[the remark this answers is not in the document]
AlonZivony: I agree with you that binary tree is better for the long term usage of this object.
However, currently I don't anticipate that I will need more than 10 states in an object.
Thus, the current implementation is probably better in performance than binary tree or AVL tree.
The current exposed API will support change in the implementation in the future, so I don't see a specific reason to change the implementation right now.
I will add a TODO for it in the code, as in the long term when other places will adopt this type it will increase performance.

pkg/proctree/events_processing.go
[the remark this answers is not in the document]
yanivagman: > @yanivagman notice that we currently only use the three basich `schedX` events, so this is not happening for all Tracee's events.

But I see that you call this function in the default case as well (for any other event), isn't it right?

pkg/proctree/exit_processing.go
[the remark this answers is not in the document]
yanivagman: The `right after` doesn't fit here

pkg/proctree/exit_processing.go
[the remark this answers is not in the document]
AlonZivony: But I currently don't use the `ProcessEvent` function, rather call the specific functions instead.
The `ProcessEvent` is useful for generic handling of events, which I use for example in the tests.
However, if I want to support calling directly to `ProcessExecEvent` then I have to use this function from there and not from the `ProcessEvent` function.

pkg/proctree/events_processing.go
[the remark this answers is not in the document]
AlonZivony: I do, but it is not called in the pipeline right now.
When and if we put it in the pipeline for other events, than we can optimize it more and maybe filter some of the events.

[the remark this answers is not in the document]
yanivagman: Ok I see. I thought the function is being used, but now I understand it is not.
So right now if you missed one of exec/fork/exit events then process tree state will never be updated for the relevant process, right? I thought this mechanism was used to resolve such cases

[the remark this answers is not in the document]
AlonZivony: So that was the general idea.
However, as we want to move control plane events we will receive much less lost events.
In addition, to support all processes in the process event stage is a bit of a trouble, as we use a map right now and I only signed `sched_process_exec`, `sched_process_fork` and `sched_process_exit`. If we sign each event, than we can just sign all non-problematic events to the general processing function. Maybe we should export it?
Anyways, I propose that we start like this, and in later PR we can add more events to the process tree processing.

rafaeldtinoco:
I'm working on this together with Alon and we will have a new PR soon.

Issue #3033, opened 2023-05-03
golang updates: updating dependencies to latest brakes --filter event=aaa,bbb (comma usage)


rafaeldtinoco:
## Description

```
$ go get -u ./...
```
then
```
$ sudo ./dist/tracee-ebpf --filter event=$(./dist/tracee-rules --list-events TRC-107)
{"level":"fatal","ts":1683088413.7106009,"msg":"App","error":"invalid filter option specified (hooked_syscalls), use '--filter help' for more info"}
```

I'm currently releasing v0.14.0 (was trying to update dependencies but couldn't because of this).

yanivagman:
Not sure it's related to the golang deps. I'm seeing this issue in main while using the unified binary

rafaeldtinoco:
There might be two issues, as I was able to reproduce that issue by using `tracee-ebpf` and you just reported a similar (or the same) issue for `tracee` single binary.

For the `trace` single binary situation, it happens because of cobra. Check the difference for the cmdline parsing:

![image](https://user-images.githubusercontent.com/7395852/236379622-14372766-3a17-48fc-8a95-ec0fdbd30483.png)

I'm checking a "good" way to keep things working as well as keeping compatibility with existing urfave code.

rafaeldtinoco:
@josedonizetti and/or @geyslan, what did you have in mind when doing the cobra filter parsing ? I can see that if you give the following command line:

```
--filter event=one,two
```

It won't concatenate `one,two` but it will have two strings: `event=one,two`. Did you expect the user to set one filter per operator and per event ? Something like:

```
--filter event=one --filter event=two --filter event!=three
```

Is that it ? Because I can "concatenate" the next arguments into the previous event + operator, but I'm not sure that is what you had planned.

rafaeldtinoco:
Also, I'm not sure the single binary is behaving like it should for filters... I created a function to deal with the issue I described above.. but the filters don't look like they're being applied correctly:

tracee-ebpf: no events (comm=cmd01 does not exist)

```
$ sudo ./dist/tracee-ebpf --filter event=openat --filt
er comm=cmd01
filter: event=openat
filter: comm=cmd01
TIME             UID    COMM             PID     TID     RET              EVENT                     ARGS
```

tracee: events (comm=cmd01 is ignored ?)

```
$ sudo ./dist/tracee --filter event=openat --filter co
mm=cmd01
filter: event=openat
[...]

PR #3057, opened 2023-05-05
cmd: adjust cobra cmdline parsing for comma

rafaeldtinoco:

Create PrepareFilterMapFromCobraFlags() in order to concatenate values from the same filter type, per operator index, so it can be used by the same filter parsering logic.

Fixes: #3033

### 2. Explain how to test it

```
sudo ./dist/tracee --filter event=one,two --filter event=three --filter comm=cmd01 --filter event!=four,five,six --filter comm=cmd02,cmd03
```

### DIFF
--- a/cmd/tracee/cmd/root.go
+++ b/cmd/tracee/cmd/root.go
@@ -24,8 +24,7 @@ var (
 	rootCmd = &cobra.Command{
 		Use:   "tracee",
 		Short: "Trace OS events and syscalls using eBPF",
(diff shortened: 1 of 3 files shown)


geyslan:
I raised a cobra solution: https://github.com/aquasecurity/tracee/pull/3058 LMK

rafaeldtinoco:
Hehehe I'm not very into cobra so I implemented the same thing myself. But we should go with yours. Closing this one.

rafaeldtinoco:
Fixed by #3058

Issue #3055, opened 2023-05-04
add raw_tracepoint attachment to bpf_attach event


roikol:
currently bpf_attach event is not triggered when a BPF program is attached to raw_tracepoint.
we would like to see it in as part of bpf_attach event

PR #3060, opened 2023-05-07
feat: support raw_tracepoint in bpf_attach event

roikol:
close: #3055

```
Date:   Thu May 4 18:52:17 2023 +0300

    feat: support raw_tracepoint in bpf_attach event

    catch raw_tracepoint attachment to BPF program as part of bpf_attach
    event
```

### 2. Explain how to test it

terminal 1: `./dist/tracee-ebpf -f e=bpf_attach`
terminal 2: `./dist/tracee-ebpf -f e=sched_process_exec`

in terminal 1 you should see `bpf_attach` event with `attach_type` raw_tracepoint and `symbol_name` sched_process_exec


pkg/ebpf/probes/probes.go
```diff
@@ -33,104 +33,105 @@ func Init(module *bpf.Module, netEnabled bool) (Probes, error) {
...
+		FdInstall:                           &traceProbe{eventName: "fd_install", probeType: kprobe, programName: "trace_fd_install"},
+		FilpClose:                           &traceProbe{eventName: "filp_close", probeType: kprobe, programName: "trace_filp_close"},
+		InotifyFindInode:                    &traceProbe{eventName: "inotify_find_inode", probeType: kprobe, programName: "trace_inotify_find_inode"},
+		InotifyFindInodeRet:                 &traceProbe{eventName: "inotify_find_inode", probeType: kretprobe, programName: "trace_ret_inotify_find_inode"},
+		BpfCheck:                            &traceProbe{eventName: "bpf_check", probeType: kprobe, programName: "trace_bpf_check"},
+		ExecBinprm:                          &traceProbe{eventName: "exec_binprm", probeType: kprobe, programName: "trace_exec_binprm"},
+		ExecBinprmRet:                       &traceProbe{eventName: "exec_binprm", probeType: kretprobe, programName: "trace_ret_exec_binprm"},
+		TracepointProbeRegisterPrioMayExist: &traceProbe{eventName: "tracepoint_probe_register_prio_may_exist", probeType: kprobe, programName: "trace_tracepoint_probe_register_prio_may_exist"},
```
rafaeldtinoco, line 134: can we make this smaller and avoid such a big code shift ?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1719,12 +1731,34 @@ int BPF_KPROBE(trace_security_file_ioctl)
...
 
-        send_bpf_attach(&p, bpf_prog_file, perf_event_file);
+        send_bpf_perf_attach(&p, bpf_prog_file, perf_event_file);
     }
 
     return 0;
 }
 
+SEC("kprobe/tracepoint_probe_register_prio_may_exist")
+int BPF_KPROBE(trace_tracepoint_probe_register_prio_may_exist)
```
yanivagman, line 1741: 1. This function only exists after kernel v5.10, meaning it will fail in any earlier version (and as well fail tracee I believe)
2. There is another kernel function named [tracepoint_probe_register_prio](https://elixir.bootlin.com/linux/v5.10.178/C/ident/tracepoint_probe_register_prio) - what about it?
3. Both of the above functions call tracepoint_add_func](https://elixir.bootlin.com/linux/v5.10.178/C/ident/tracepoint_add_func) which is also available in earlier kernel version - won't it be better to have a kprobe on it instead?

rafaeldtinoco: There is also `tracepoint_probe_register` (vanilla), which is the one I can see being called by all `register_trace_##name` functions in DECLARE_TRACE macro. This one was added back in 2015.

By the way, a good point was made. Without an instrumentation test for the bpf_attach (which might be pretty easy to be added), we wouldn't get the error Yaniv mentions, maybe we should add a quick instrumentation test (to make sure logic is tested among all kernel versions and arches).

rafaeldtinoco (changes requested):
Per comments.

pkg/ebpf/c/tracee.bpf.c
roikol: the `tracepoint_probe_register_prio_may_exist` exists in kernels v4.19+ https://elixir.bootlin.com/linux/v4.19.282/source/kernel/tracepoint.c#L364
(there was something weird yesterday with Elixir, it couldn't find the func in versions smaller than v5.10 like Yaniv said, but now it can find it).

if i understand correctly:

tracepoint_add_func is called from two paths - 
bpf + raw_tracepoint
user enabled raw_tracepoint under tracefs

we are only interested in the first path which uses `tracepoint_probe_register_prio_may_exist`.

the other path uses the `tracepoint_probe_register` and `tracepoint_probe_register_prio`.

the bpf + raw_tracepoint path starts in the bpf() syscall.
the only functions we can trace in this path:
bpf() -> bpf_raw_tracepoint_open() -> ... -> tracepoint_probe_register_prio_may_exist()

and only `tracepoint_probe_register_prio_may_exist` holds the information we need (bpf program + tracepoint info) and is not vulnerable to TOCTOU.


there may be another option of combining data from `tracepoint_add_func` with a BPF LSMhook (e.g. `security_bpf_prog`). this will remove the need of using `tracepoint_probe_register_prio_may_exist`, but it still add a kprobe, and also add more comlexity.
WDYT?


about the tests - agreed, we should have an instrumentation test for `bpf_attach` event.
i'll try to add the tests today/tomorrow.

rafaeldtinoco:
```
bpf() -> bpf_raw_tracepoint_open() -> ... -> tracepoint_probe_register_prio_may_exist()
```

If this path exists in previous versions (and it looks it does), then I believe we're good =D.

rafaeldtinoco:
I'll review, rebase and, possibly, merge this before the freeze tomorrow (FYIO).

yanivagman (approved):
LGTM
nit: indentation of all probes because of long name `TpProbeRegisterPrioMayExist`

tests/e2e-instrumentation-test.sh
```diff
@@ -86,7 +86,7 @@ for TEST in $TESTS; do
         --cache mem-cache-size=512 \
         --output format:json \
         --output option:parse-arguments \
-        --filter comm=echo,mv \
+        --filter comm=echo,mv,bpftrace \
```
rafaeldtinoco, line 89: I guess you're doing the test with 'tracee' and not 'bpftrace', will remove this in the rebase, check if tests are good etc.

rafaeldtinoco:
LGTM as well. Im rebasing on https://github.com/aquasecurity/tracee/pull/3148 and merging.

Issue #3063, opened 2023-05-07
empty helpers list in bpf_attach and security_bpf_prog


roikol:
## Description

in events `bpf_attach` and `security_bpf_prog` - list of bpf helpers that are used by a bpf program is empty on kernels greater than 5.13.

## Output of `tracee -v`:

```
(paste your output here)
```

`Tracee version "v0.14.0-12-g2f7f65cc"`

## Output of `uname -r`:

```
(paste your output here)
```

`5.15.0-microsoft-standard-WSL2`

PR #3064, opened 2023-05-07
fix: correct list of bpf helpers

roikol:
fixes: #3063

```
Date:   Sun May 7 14:39:10 2023 +0300

    fix: correct list of bpf helpers

    get correct kernel version when getting bpf helpers list for
    bpf_attach and security_bpf_prog events
```

### 2. Explain how to test it

in kernel version >= 5.13 : 

terminal 1: `./dist/tracee-ebpf -f e=bpf_attach`
terminal 2: `./dist/tracee-ebpf -f e=security_socket_bind`

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -3513,12 +3513,13 @@ int BPF_KPROBE(trace_check_helper_call)
 
     int func_id;
 
-#if (LINUX_VERSION_CODE < KERNEL_VERSION(5, 13, 0))
-    func_id = (int) PT_REGS_PARM2(ctx);
-#else
-    struct bpf_insn *insn = (struct bpf_insn *) PT_REGS_PARM2(ctx);
-    func_id = READ_KERN(insn->imm);
-#endif
+    if (!bpf_core_enum_value_exists(enum bpf_func_id, BPF_FUNC_for_each_map_elem)) {
+        // if BPF_FUNC_for_each_map_elem doesn't exist under bpf_func_id - kernel version < 5.13
+        func_id = (int) PT_REGS_PARM2(ctx);
+    } else {
+        struct bpf_insn *insn = (struct bpf_insn *) PT_REGS_PARM2(ctx);
+        func_id = READ_KERN(insn->imm);
+    }
 
     return handle_bpf_helper_func_id(p.event->context.task.host_tid, func_id);
 }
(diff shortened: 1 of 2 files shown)


pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3513,12 +3513,13 @@ int BPF_KPROBE(trace_check_helper_call)
 
     int func_id;
 
-#if (LINUX_VERSION_CODE < KERNEL_VERSION(5, 13, 0))
```
geyslan, line 3516: > Sometimes the only way to detect the presence of the necessary feature is through checking the Linux kernel version. Libbpf allows to do that from BPF program code using a special extern variable:

`extern int LINUX_KERNEL_VERSION __kconfig;`

https://nakryiko.com/posts/bpf-core-reference-guide/#linux-kernel-version

yanivagman: Actually it seems that the commit that introduced the parameters change of this function also introduced the helper addition (https://github.com/torvalds/linux/commit/cc0f83530934dda0ce1dd01990d2f37f8c3f0d92#diff-edbb57adf10d1ce1fbb830a34fa92712fd01db1fbd9b6f2504001eb7bcc7b9d0R5666) so I think it is good enough. Checking for kernel version doesn't seem to have an advantage here since code can be backported.

yanivagman (approved):
LGTM

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: verifier <= 5.4 had issues when using conditional branch and scalar values coming from kconfig relocations IIRC. we should revisit that, but I remember we couldn't do that that is why we started using type checks for version checks.

rafaeldtinoco: image

ok libbpf treats that as a special kconfig file now and treats different distro versioning. IF we can get passed the 5.4 verifier then we would be good to start using that schema for linux kernel versions checking (BUT remember that things are backported to RHEL/SUSE etc, so it would be better to check for types being used).

geyslan: What about a schema using both `LINUX_KERNEL_VERSION` and `bpf_core_*_exists()`? Maybe setting bits for enabled features in a bitmask (or map) populated by that schema.

PR #3067, opened 2023-05-08
v0.14.1 dependencies updates (security, ...) and an urfave quick fix

rafaeldtinoco:
- tracee 0.14.0 should have updated packages but didn't because of an urfave issue. this is solved by this PR and the dependencies should be updated (security issues, etc).


cmd/tracee-ebpf/main.go
```diff
@@ -49,6 +49,7 @@ func main() {
 
 			return runner.Run(ctx)
 		},
+		SliceFlagSeparator: " ",
```
rafaeldtinoco, line 52: needed to make the urfave to parse cmdline strings the same way it did

yanivagman: From what I understand, the default separator for string slice is comma - so why don't we use it as is? We can simply remove the parsing we do for comma seperation ourseleves, can't we?
Also, from what I understand, we can simply disable the separation using 
https://github.com/urfave/cli/pull/1588 - maybe we should use that instead of a space separator?

geyslan: I'm not against to remove our own parsing in the future since we have both urfave and cobra parsers which can achieve that, but for this issue I consider `SliceFlagSeparator` enough.

geyslan (approved):
LGTM

cmd/tracee-ebpf/main.go
rafaeldtinoco: > From what I understand, the default separator for string slice is comma - so why don't we use it as is? We can simply remove the parsing we do for comma seperation ourseleves, can't we?

we should do that after dropping urfave. For now we should use comma for both (after latest fix for cobra) and make sure command line is good, then we can change the parsing behavior (IMO).

yanivagman (approved):
LGTM

PR #3069, opened 2023-05-08
workflow: add 6.1 and 6.3 kernels

rafaeldtinoco:
@yanivagman FYI, IMO would be good to have those included (and any issue fixed, like https://github.com/aquasecurity/tracee/pull/3067, before the 0.14.1).

These are 6.1 and 6.3 kernels from the upstream stable tree, being used in Ubuntu Jammy, and using the same kconfig values as Ubuntu does for 5.19 kernels (this way we can test upstream trees before Ubuntu releases them).

### DIFF
--- a/.github/workflows/debug-shell.yaml
+++ b/.github/workflows/debug-shell.yaml
@@ -25,6 +25,8 @@ on:
           - jammy519-arm64
           - focal419
           - focal419-arm64
+          - jammy6127
+          - jammy6301
 jobs:
   alma418:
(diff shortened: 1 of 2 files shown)


rafaeldtinoco:
Im only including the DEBUG SHELL for 6.1 and 6.3 kernels in this PR so I can further investigate why tracee is failing in those kernels (and enable the PR workflow for them once they're fixed).

I need to merge this orelse the debug shell workflow won't work and I won't be able to investigate.

PR #3092, opened 2023-05-10
Fix logger

josedonizetti:

The logger should be the first thing to initialize, as we have packages using it already, like: https://github.com/aquasecurity/tracee/blob/main/pkg/signatures/signature/signature.go#L23

There is also two simple commits to remove an old TODO, and clean up an `if` not used.


rafaeldtinoco (approved):
LGTM

PR #3099, opened 2023-05-12
docs: remove classic-docs

josedonizetti:

Remove classic-docs, with the upcoming issue to add `analyze` command, we will fully drop the support for tracee-rules and tracee-ebpf.


yanivagman (approved):
LGTM

Issue #2736, opened 2023-02-16
tracee: add subcomand to analyze events in a file


josedonizetti:
A user should be able to store events in a file, and then run the signatures events on this data to see if trigger any signature.

eg:
```
tracee --output json:/events.json
tracee --trace event=anti_debugging --input json:/events.json
```

This is supported by the binaries `tracee-ebpf` and `tracee-rules`.

itaysk:
tracee-rules was just a rule engine so giving it input had a clear meaning - this is the input for the rules.
tracee now is not a rules engine, it's a runtime security tool that generates security observability events for the user. the primary use case of tracee now is more about generating events than consuming events, so I'm not sure `tracee --input` is the best option.

itaysk:
I would be interested to hear more about the use case of this requirement. I am guessing it comes from researchers wanting to reproduce/test signature events?

yanivagman:
> I would be interested to hear more about the use case of this requirement. I am guessing it comes from researchers wanting to reproduce/test signature events?

With the current architecture of two binaries, it was possible to generate the events using tracee-ebpf in one node, and analyze it with tracee-rules in another. It was also possible to keep the events in a log file for later analysis on the same node. Moving to one binary architecture, we need to keep this option viable.

josedonizetti:
> the primary use case of tracee now is more about generating events than consuming events, so I'm not sure `tracee --input` is the best option.

Agree, `--input` is not the best option, my plan was more aligned with the issue description, a specific subcommand for it, or maybe even a different binary.  
```
tracee analyze
#or
tracee-analyze 
```

The use case might be worthwhile for users outside of research too, with policies and an API the user will be able to enable/disable the collection of specific events, one might be debugging a specific incident, and enable a specific collection, which he can write some quick rego and see if it triggers outside of the main cluster, without the need to deploy anything.

Though he could also do it in the cluster, add a yaml with rego, and enable the event for just a context.

itaysk:
ok go it. it's not in any milestone so I presume it's just a reminder, not a immediate work item right?

josedonizetti:
@itaysk Yes, I have extracted issues from #2355 that aren't expected to be on this milestone, for this milestone we aim to finish all TODOS, and close #2355.

PR #3101, opened 2023-05-13
tracee: add analyze cmd

josedonizetti:

Fix https://github.com/aquasecurity/tracee/issues/2736

Adds analyze subcommand. The subcommand only accepts json as input, and only output json.

### 2. Explain how to test it

```
tracee --filter event=ptrace --output=json:events.json

tracee analyze events.json
tracee analyze --event=anti_debugging events.json
tracee analyze --event=dropped_executable events.json
```

### 3. Other comments

This is marked as EXPERIMENT, because the implementation will later change to support datasources. For now this is equivalent to `tracee-rules`.

### DIFF
--- a/cmd/tracee/cmd/analyze.go
+++ b/cmd/tracee/cmd/analyze.go
@@ -1,0 +1,225 @@
+package cmd
+
+import (
+	"bufio"
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+
+	"github.com/aquasecurity/tracee/pkg/cmd/flags"
+	"github.com/aquasecurity/tracee/pkg/cmd/initialize"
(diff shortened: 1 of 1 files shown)


josedonizetti:
@yanivagman 90% of the work is done here, but before finishing documentation, and testing, I would like to know if the implementation covers what we need. Who from research can we ping to review this?

rafaeldtinoco (approved):
LGTM. 

NOTE: I think its ok to have it already included as experimental in v0.16.0 release as its orthogonal to any other code and couldn't cause any harm.

NDStrahilevitz:
We should add a warning not to rely on data sources when running `tracee analyze` for now, just to cover that edge case until we resolve it.

josedonizetti:
> We should add a warning not to rely on data sources when running `tracee analyze` for now, just to cover that edge case until we resolve it.

That makes sense, as it is marked as experimental I expected people to know it will not support everything, but the warning will also help. I'll add it today!

PR #3113, opened 2023-05-18
types: add context information to the same struct

josedonizetti:
trace.Event now group all contextual information under the Context type


josedonizetti:
Closing for now while we still discuss the idea.

PR #3114, opened 2023-05-18
tracee: bump types with new trace.Event

josedonizetti:
The new trace.Event groups all contextual information under the same struct.


josedonizetti:
Closing for now while we still discuss the idea.

Issue #3110, opened 2023-05-16
problem filtering args of type 'bytes'


roikol:
## Description

in previous versions of tracee i was able to filter arguments of type `bytes` like so:

`./dist/tracee-ebpf -f e=magic_write -f magic_write.args.bytes="[11 11*,[22 22*,[33 33*"`

when trying to do the same thing using policies, the event is not triggered at all.

policies i've tried (difference is the `""`):

```
name: test policy
description: test magic_write filter
scope:
  - global
defaultAction: log
rules:
  - event: magic_write
    filter:
      - args.bytes="[11 11*,[22 22*,[33 33*"
```

```
name: test policy
description: test magic_write filter
scope:
  - global
defaultAction: log
rules:
  - event: magic_write
    filter:
      - args.bytes=[11 11*,[22 22*,[33 33*
```

* for the sake the example i gave a simple psuedo filter (`[11 11*,[22 22*,[33 33*`). 
  i can provide the real filter we use if needed

## Output of `tracee version`:

```
(paste your output here)
```

Tracee version: "v0.14.2"

PR #3115, opened 2023-05-18
policies: fix args filtering

josedonizetti:

Fix https://github.com/aquasecurity/tracee/issues/3110

### 2. Explain how to test it

Save policy to p.yaml ->
```
name: test
description: test
scope:
  - global
defaultAction: log
rules:
  - event: magic_write
    filter:
      - args.bytes=[108 10]
```

```
sudo tracee -p p.yaml
```

```
echo "l" > /tmp/test.txt
```


yanivagman (approved):
LGTM

Issue #2975, opened 2023-04-04
Add webhook to global default policy action


yanivagman:
Currently only log action is supported in policies.
Add support for webhook as well.

PR #3126, opened 2023-05-21
Add policy printer

josedonizetti:

Adds a printer for policy actions. The printer creates a map where a policy event can point to a specific printer, if it was overriding on the policy declaration, or if not, uses the default action. 

Fix https://github.com/aquasecurity/tracee/issues/2975

### 2. Explain how to test it

```
name: dig
scope:
  - global
defaultAction: log
rules:
  - event: ptrace
    action: webhook:http://localhost:8080
    action: webhook:http://localhost:8081
```

### DIFF
--- a/docs/docs/policies/actions.md
+++ b/docs/docs/policies/actions.md
@@ -1,13 +1,47 @@
 # Actions
 
-Actions are taken when there is an event match. Currently the only action supported is `log`. 
+## DefaultAction
+Every policy must have a `defaultAction`. Actions are taken when there is match on some rule declared at the policy. 
+The following actions are currently supported:
+
+- `log` - output events in json format. The default path to file is stdout.
+- `forward:http://url/fluent` - send events in json format using the Forward protocol to a Fluent receiver
+- `webhook:http://url/webhook` - send events in json format to the webhook url
+
+## Examples
+
+### Log action
 
 ```yaml
-name: overview policy
-description: sample overview policy
+name: log_sample_policy
+description: log sample policy
 scope:
   - global
 defaultAction: log
+rules:
+  - event: dropped_executable
+```
+
+### Webhook action
+
+```yaml
+name: webhook_sample_policy
+description: webhook_sample_policy
+scope:
+  - global
+defaultAction: webhook:http://localhost:8080
+rules:
+  - event: dropped_executable
+```
+
+### Fluentd action
+
+```yaml
+name: webhook_sample_policy
+description: webhook_sample_policy
+scope:
+  - global
+defaultAction: forward:tcp://localhost:24224
 rules:
   - event: dropped_executable
 ```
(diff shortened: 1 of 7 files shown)


josedonizetti:
The PR is working, and good for this release. But I'm starting to wonder it should have been different because right now a policy can either have one default action, or the user can override the action on an event with one action. The per-event declaration should actually allow for multiple actions, because if the user wants to do multiple things? 

For example, he wants to send an event to the webhook, but he also wants to log it?

I think we can ship it as is and change the behaviour on a next release, but I'm okay with holding if you don't agree.

geyslan (approved):
Overall, LGTM.

geyslan:
> For example, he wants to send an event to the webhook, but he also wants to log it?
> 
> I think we can ship it as is and change the behaviour on a next release...

Yep, I agree with that.

docs/docs/policies/actions.md
```diff
@@ -1,13 +1,47 @@
 # Actions
 
-Actions are taken when there is an event match. Currently the only action supported is `log`. 
+## DefaultAction
+Every policy must have a `defaultAction`, actions are taken when there is match on some rule declared at the policy.
```
geyslan, line 4:
```suggestion
Every policy must have a `defaultAction`. Actions are taken when there is match on some rule declared at the policy. 
```

docs/docs/policies/actions.md
```diff
@@ -1,13 +1,47 @@
...
+
+```yaml
+name: webhook_sample_policy
+description: webhook_sample_policy
+scope:
+  - global
+defaultAction: forward:tcp://localhost:24224
+rules:
+  - event: dropped_executable
+```
```
geyslan, line 47: Should these also be `examples/` files?

geyslan (changes requested):
If you take the effort of creating the example files, I wonder if it's a good idea to place all policies under ./examples/policies and global config under ./examples/config. WDYT?

docs/docs/policies/actions.md
josedonizetti: don't think so, in the example we use logs because the user can apply at will and see the result, for webhook/fluentd the user needs to do other steps, so only docs is enough.

josedonizetti:
> If you take the effort of creating the example files, I wonder if it's a good idea to place all policies under ./examples/policies and global config under ./examples/config. WDYT?

Yup, makes sense to split policies examples from config examples, but in another PR

geyslan (approved):
LGTM

docs/docs/policies/actions.md
```diff
@@ -1,13 +1,47 @@
...
+defaultAction: webhook:http://localhost:8080
+rules:
+  - event: dropped_executable
+```
+
+### Fluentd action
+
+```yaml
+name: webhook_sample_policy
+description: webhook_sample_policy
```
yanivagman, line 41: should probably be renamed to fluentd/forward

Issue #2114, opened 2022-08-27
Add readinessProbe


josedonizetti:
## Feature description

Once we have `/healthz` endpoints on tracee-ebpf and tracee-rules, let's add a readiness probe on k8s so a pod is only marked as `Running` if both process are healthy.

PR #3128, opened 2023-05-21
k8s: add readiness probe

josedonizetti:

Fix https://github.com/aquasecurity/tracee/issues/2114

The readiness probe will be useful once we create services for the metrics endpoint and also the API which is upcoming.

### DIFF
--- a/deploy/kubernetes/tracee/tracee.yaml
+++ b/deploy/kubernetes/tracee/tracee.yaml
@@ -14,7 +14,7 @@ data:
       - mem-cache-size=512
     perf-buffer-size: 1024
     containers: true
-    healthz: false
+    healthz: true
     metrics: true
     pprof: false
     pyroscope: false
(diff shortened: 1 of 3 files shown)


deploy/kubernetes/tracee/tracee.yaml
```diff
@@ -57,6 +57,10 @@ spec:
         env:
           - name: LIBBPFGO_OSRELEASE_FILE
             value: /etc/os-release-host
+        readinessProbe:
+          httpGet:
+            path: /healthz
+            port: 3366
```
yanivagman, line 63: Should this port be fixed?

josedonizetti: Yes, for the static installation (under deploy/kubernetes/tracee) it is always fixed, and matches the configmap we provide on the same file https://github.com/aquasecurity/tracee/blob/main/deploy/kubernetes/tracee/tracee.yaml#L21.

For helm, it is the default and is updated based on what input the user gives.

yanivagman (approved):
LGTM

Issue #3129, opened 2023-05-21
k8s: add base policy with config map


josedonizetti:
kubernetes should have a default base policy installed as a configmap, users should be able to add their own policies to this to load it on the tracee.

PR #3130, opened 2023-05-21
k8s: add default signatures policy

josedonizetti:

Fix https://github.com/aquasecurity/tracee/issues/3129

This PR adds a default policy for signatures events on kubernetes. It introduces a configmap that can be extended with newer policies, or changes to the default policy depending on the user needs. It also includes a tutorial on how to add a new policy.


josedonizetti:
@AnaisUrlichs can you review the tutorial I wrote?

rafaeldtinoco (approved):
LGTM. Will let others to review as you requested.

PR #3137, opened 2023-05-24
logger: flags: add filter and filter-out

geyslan:

callbacks: print formatted libbpf verifier errors
    
    This filters out logs originated from libbpf which are related to bpf
    loading failure. It outputs that error message in a formatted way
    via stderr.

logger: flags: add filter and filter-out
    
    This commit adds two new options to log flag: filter and filter-out.
    
    The filter option is used to filter out logs that do not match the
    option values. The filter-out option is used to filter out logs that
    match the option values.
    
    The filter and filter-out flags can be used multiple times to filter
    out or filter, and can be used together.
    
    Filter options:
      msg=<value,...>    | Filters logs that message contains a value.
      regex=<value,...>  | Filters logs that a regex matches the message.
      pkg=<value,...>    | Filters logs that originate from a package.
      file=<value,...>   | Filters logs that originate from a file.
      lvl=<value,...>    | Filters logs that are of a specific level.
      libbpf             | Filters logs that originate from libbpf.
    
    Some examples:
    
    --log filter:'msg=foo,bar;pkg=core;lvl=error'
    
    would filter in logs that have either foo or bar in the message, are
    from the core package, and are of error level. All other logs would be
    filtered out.
    
    --log filter-out:'msg=foo,bar;pkg=core;lvl=error'
    
    would filter out log messages that have either foo or bar in the
    message, are from the core package, and are of error level. All other
    logs would be output.
    
    --log filter:msg=foo,bar --log filter-out:pkg=core
    
    would filter in logs that have either foo or bar in the message, and
    based on that result, filter out logs that are from the core package.
    All other logs would be filtered out.

### 2. Explain how to test it

For log filter flag, see above.

For verifier errors test, apply the patch below, compile, run and see.

```diff
diff --git a/pkg/ebpf/c/tracee.bpf.c b/pkg/ebpf/c/tracee.bpf.c
index 184f23f3..837578ce 100644
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -2773,7 +2773,7 @@ static __always_inline int capture_file_write(struct pt_regs *ctx, u32 event_id)
 
 // Check if capture write was requested for this path
 #pragma unroll
-    for (int i = 0; i < 3; i++) {
+    for (int i = 0; i < 3000; i++) {
         int idx = i;
         path_filter_t *filter_p = bpf_map_lookup_elem(&file_filter, &idx);
         if (filter_p == NULL) 
```

### 3. Other comments

Fix: #3035

### DIFF
--- a/docs/docs/outputs/logging.md
+++ b/docs/docs/outputs/logging.md
@@ -16,4 +16,40 @@ Logs can be aggregated for a given interval to delay its output:
 sudo ./dist/tracee --log debug --log aggregate:5s
 ```
 
+Filter logs which message contains specified words:
 
+```console
+sudo ./dist/tracee --log filter:msg=foo,bar
+```
+
+Filter logs using regular expressions against messages:
+
+```console
+sudo ./dist/tracee --log filter:regex='^foo'
+```
+
+Filter logs originating from a specific package:
+
+```console
+sudo ./dist/tracee --log filter:pkg=core
+```
+
+Filter logs originating from a specific file:
+
+```console
+sudo ./dist/tracee --log filter:file=/pkg/cmd/flags/logger.go
+```
+
+Filter logs based on their severity level:
+
+```console
+sudo ./dist/tracee --log filter:lvl=error
+```
+
+Filter logs originating from libbpf:
+
+```console
+sudo ./dist/tracee --log filter:libbpf
+```
+
+All `--log filter` options can also be used with `--log filter-out` for the opposite behavior. For more information, please refer to the `--log` help in the CLI.
--- a/pkg/cmd/flags/logger.go
+++ b/pkg/cmd/flags/logger.go
@@ -1,6 +1,7 @@
 package flags
 
 import (
+	"errors"
 	"os"
 	"strings"
 	"time"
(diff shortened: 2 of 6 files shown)


pkg/cmd/flags/logger.go
```diff
@@ -64,40 +130,96 @@ func PrepareLogger(logOptions []string, newBinary bool) (logger.LoggingConfig, e
...
-		case "info":
-			lvl = logger.InfoLevel
-		case "warn":
-			lvl = logger.WarnLevel
-		case "error":
-			lvl = logger.ErrorLevel
-		case "fatal":
-			lvl = logger.FatalLevel
-		default:
+		// parse filter option
```
josedonizetti, line 156: can we extract this section to be its own method?

pkg/logger/filter.go
```diff
@@ -0,0 +1,337 @@
...
+}
+
+func newFilterString() *filterString {
+	return &filterString{
+		vals: map[string]FilterKind{},
+	}
+}
+
+func (fs *filterString) add(s string, t FilterKind) {
+	if _, ok := fs.vals[s]; ok && t == FilterIn {
```
josedonizetti, line 103: if I understood correctly the idea here is that we avoid overriding an existing filter with a filter-out? if so, maybe log a debug message whenever an override happens? could help a user identify a mistake

docs/docs/outputs/logging.md
```diff
@@ -16,4 +16,40 @@ Logs can be aggregated for a given interval to delay its output:
...
+Filter logs originating from a specific file:
+
+```console
+sudo ./dist/tracee --log filter:file=/pkg/cmd/flags/logger.go
+```
+
+Filter logs based on their severity level:
+
+```console
+sudo ./dist/tracee --log filter:lvl=error
```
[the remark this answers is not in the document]
geyslan: Well, isn't the same behaviour.

The default severity level being Info would result in logging all equals higher than that, so Info, Warn, Error, Panic.

image

However by setting `--log filter:lvl=error`, whatever severity level is set, all logs will be filtered out except those with Error level.

pkg/cmd/flags/logger.go
```diff
@@ -34,21 +50,71 @@ func InvalidLogOption(opt string, newBinary bool) error {
...
+		return nil
+	case strings.HasPrefix(opt, "aggregate"):
+		return nil
+	case strings.HasPrefix(opt, "filter-out"):
+		return nil
+	case strings.HasPrefix(opt, "filter"):
+		return nil
+	}
+
+	if _, err := parseLevel(opt); err == nil {
```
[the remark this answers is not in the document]
geyslan: Yes. It's intentional. I'm changing parseLevel() return so it would be clearer.

pkg/cmd/flags/logger.go
```diff
@@ -34,21 +50,71 @@ func InvalidLogOption(opt string, newBinary bool) error {
 	return errfmt.Errorf("invalid log option: %s, use '--log help' for more info", opt)
 }
 
+func InvalidLogOptionValue(opt string, newBinary bool) error {
```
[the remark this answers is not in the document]
geyslan: Solved, as we discussed offline. Changes arising soon.

pkg/logger/filter.go
geyslan: > if I understood correctly the idea here is that we avoid overriding an existing filter with a filter-out? 

Exactly. Since filter-out has precedence.

> If so, maybe log a debug message whenever an override would happen? could help a user identify a mistake.

Good idea. 👍🏼

geyslan: Oh, wait... we are still at first stage building the logger. Trying to log something would work correctly only if via stderr. `fmt.Fprintf(os.Stderr...)` WDYT?

josedonizetti: oh! true. Maybe leave as is than

pkg/cmd/flags/logger.go
geyslan: I'll take a look at this later. 👍🏼

pkg/logger/filter.go
geyslan: I'll think about it more.

pkg/cmd/flags/logger.go
geyslan: Well, we can, but it would demand some back and forth of data and state. So let's leave it rest for now.

pkg/logger/filter.go
geyslan: Ok, I created the `ErrFilterOutExistsForKey` error to be checked and logged instead of being considered a failure.

josedonizetti (approved):

PR #3142, opened 2023-05-25
3026 rebased

rafaeldtinoco:
This is https://github.com/aquasecurity/tracee/pull/3026 rebased and using types merged by: https://github.com/aquasecurity/tracee/pull/3141

Resolve https://github.com/aquasecurity/tracee/issues/2561
Resolve https://github.com/aquasecurity/tracee/issues/1423

@NDStrahilevitz : FYI (kept commits and credits).


pkg/containers/datasource.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func (ctx SignaturesDataSource) Get(key interface{}) (map[string]interface{}, error) {
+	containerId, ok := key.(string)
+	if !ok {
+		return nil, detect.ErrKeyNotSupported
+	}
+	ctx.containers.mtx.RLock()
+	defer ctx.containers.mtx.RUnlock()
+	for _, cgroup := range ctx.containers.cgroupsMap {
```
yanivagman, line 26: So every Get(container_id is O(n) - won't it be better to add another cgroupInfo map where the key is the container id and the value is a pointer to cgroup info so lookup will be O(1)?

pkg/containers/datasource.go
```diff
@@ -0,0 +1,72 @@
...
+		"k8s_pod_name":      "string",
+		"k8s_pod_namespace": "string",
+		"k8s_pod_sandbox":   "bool",
+	}
+	schema, _ := json.Marshal(schemaMap)
+	return string(schema)
+}
+
+func (ctx SignaturesDataSource) Version() uint {
+	return 1
```
yanivagman, line 63: We're missing the major/minor concept here (major for breaking changes in the schema and minor for additions of new fields to the schema)

yanivagman (commented):
Sorry for reviewing after merge, I believe the following comments can be managed as follow ups

yanivagman (commented):
Sorry for reviewing after merge, I believe the following comments can be managed as follow ups

Issue #3143, opened 2023-05-25
policy: error processing policy with arguments


josedonizetti:
## Description

```
sudo ./dist/tracee -p test.yaml
Error: policy.validateContext: policy openat_args_pathname, filter args.pathname is not valid
```

```
name: openat_args_pathname
description: traces openat under /tmp/*
scope:
  - global
defaultAction: log
rules:
  - event: openat
    filter:
      - args.pathname=/tmp*
```
## Output of `tracee version`:

```
(paste your output here)
```

## Output of `uname -a`:

```
(paste your output here)
```

PR #3144, opened 2023-05-25
policy: fix validation of args and retval

josedonizetti:

Fix https://github.com/aquasecurity/tracee/issues/3143

Fixes validation for filter `args` and `retval` on policies.

### 2. Explain how to test it

create file p.yaml with policy below 
```
name: openat_args_pathname
description: traces openat under /tmp/*
scope:
  - global
defaultAction: log
rules:
  - event: openat
    filter:
      - args.pathname=/tmp*
```

Pass policy to tracee
```
sudo tracee -p p.yaml
```

### DIFF
--- a/pkg/policy/policy_file_test.go
+++ b/pkg/policy/policy_file_test.go
@@ -5,9 +5,16 @@ import (
 	"testing"
 
 	"gotest.tools/assert"
+
+	"github.com/aquasecurity/tracee/pkg/events"
 )
 
 func TestPolicyValidate(t *testing.T) {
+	fakeSignatureEvent := events.NewEventDefinition("fake_signature", []string{"signatures", "default"}, nil)
+
+	err := events.Definitions.Add(9000, fakeSignatureEvent)
+	assert.NilError(t, err)
+
 	tests := []struct {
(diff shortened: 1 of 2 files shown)


pkg/policy/policy_file_test.go
```diff
@@ -188,15 +195,177 @@ func TestPolicyValidate(t *testing.T) {
...
+				Rules: []Rule{
+					{
+						Event: "write",
+						Filter: []string{
+							"retval",
+						},
+					},
+				},
+			},
+			expectedError: errors.New("policy.PolicyFile.validateRules: policy invalid_retval, invalid filter operator: retval"),
```
geyslan, line 214: Just a note for ourselves of the future: We should use error variables to be possible to compare them in tests and further code path. I've put a comment about this in https://github.com/aquasecurity/tracee/pull/3137/files#diff-f70250a062f09eebb2d5b94837022723ff90b7175790d0d7326da75f1ade0222R360-R361

The unique concern is how to make it comparable when the error is a formatted one.

geyslan (approved):
LGTM

PR #3154, opened 2023-05-28
chore: restore old ContainerID field

NDStrahilevitz:
## Change 

The PR reintroduces the `ContainerID` field into the `trace.Event` type.
It should be equivalent to using the new `.Container.ID` field.
This was done because the field change back in v0.13 has broken some json consumers of tracee.
We've decided to duplicate this field in it's original format until we make the event type overhaul in v0.16 to let consumers have more time to prepare migrations.

## Testing

Run `tracee -o json -f c`
Start a container `docker run --rm ubuntu`
Confirm that both the `"containerId"` field and `"container": "id"` field are populated the same way.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/Masterminds/sprig/v3 v3.2.3
 	github.com/aquasecurity/libbpfgo v0.4.8-libbpf-1.2.0.0.20230509162948-80f41e18e690
 	github.com/aquasecurity/libbpfgo/helpers v0.4.6-0.20230321190037-f591a2c5734f
-	github.com/aquasecurity/tracee/types v0.0.0-20230525195705-4707c7022ac4
+	github.com/aquasecurity/tracee/types v0.0.0-20230529103208-c134a2b7369a
 	github.com/containerd/containerd v1.7.0
 	github.com/docker/docker v23.0.5+incompatible
 	github.com/golang/protobuf v1.5.3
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -162,8 +162,10 @@ require (
...
 	golang.org/x/sync v0.2.0 // indirect
 	golang.org/x/text v0.9.0 // indirect
-	gopkg.in/yaml.v3 v3.0.1
+	gopkg.in/yaml.v3 v3.0.1 // indirect
 	kernel.org/pub/linux/libs/security/libcap/psx v1.2.68 // indirect
 )
 
 replace github.com/kubernetes/cri-api => k8s.io/cri-api v0.23.5-rc.0
+
+replace github.com/aquasecurity/tracee/types => ./types
```
NDStrahilevitz, line 171: Move to types PR once approved

AsafEitani (approved):
LGTM

josedonizetti (approved):
LGTM

tests/integration/integration_test.go
```diff
@@ -186,7 +186,7 @@ func checkNewContainers(t *testing.T, gotOutput *eventOutput) {
 	containerIds := []string{}
 	output := gotOutput.getEventsCopy()
 	for _, evt := range output {
-		containerIds = append(containerIds, evt.Container.ID)
+		containerIds = append(containerIds, evt.Container.ID, evt.ContainerID)
```
yanivagman, line 189: Why do we add (the same) container id twice here?

NDStrahilevitz: Just confirming the field duplication works as intended.

NDStrahilevitz: Once we move to the new type I assume the test will break and change again.

PR #3176, opened 2023-05-31
policy: actions printers configured with -output

josedonizetti:


geyslan:
Starting to review this.

docs/docs/policies/actions.md
```diff
@@ -22,26 +22,98 @@ rules:
   - event: dropped_executable
 ```
 
-### Webhook action
+if we start tracee it will log table to stdout:
+
+```
```
geyslan, line 27: I don't know if this is really changing the output on the doc build, but when it's related to _running commands_ we started to tag the block with `console`, to differentiate it from `shell` scripts (which have its own highlight).

geyslan: It's the same for the below blocks.

docs/docs/policies/actions.md
```diff
@@ -22,26 +22,98 @@ rules:
...
+tracee --policies policy.yaml
+```
+
+To customize it, we use the [`--output`](../../outputs/output-formats) flag, for example, instead of table I would like to log a json to stdout:
```
geyslan, line 31:
```suggestion
To customize it, we use the [`--output`](../../outputs/output-formats) flag. For example, instead of table I would like to log a json to stdout:
```

docs/docs/policies/actions.md
```diff
@@ -22,26 +22,98 @@ rules:
...
+
+```
+tracee --policies policy.yaml --output json
+```
+
+or I would like to send json logs to a file:
```
geyslan, line 37:
```suggestion
Or I would like to send json logs to a file:
```

docs/docs/policies/actions.md
```diff
@@ -22,26 +22,98 @@ rules:
   - event: dropped_executable
 ```
 
-### Webhook action
+if we start tracee it will log table to stdout:
+
+```
+tracee --policies policy.yaml
```
geyslan, line 28:
```suggestion
tracee --policy policy.yaml
```

Same to below repetitions.

docs/docs/policies/actions.md
```diff
@@ -22,26 +22,98 @@ rules:
...
+```
+tracee --policies policy.yaml --output forward:tcp://localhost:24224
+```
+
+## Multiple outputs
+
+Tracee supports multiple outputs, and when used with policies, we have a broadcast by type.
+
+For example, if you declared an output to json stdout, and an output table to a file, the log action
+will broadcast the event matched to both outputs.
```
geyslan, line 84: remove the line break?

docs/docs/policies/actions.md
```diff
@@ -22,26 +22,98 @@ rules:
...
+  - event: dropped_executable
+```
+
+```
+tracee --policies policy.yaml --output json:stdout --output table:/path/to/file
+```
+
+## Rule Actions
+
+A Rule can override default action of a policy, for example, if you create a policy which by default will log, we can change a rule to send a webhook if a match happens.
```
geyslan, line 102:
```suggestion
A Rule can override default action of a policy. For example, if you create a policy which by default will log, we can change a rule to send a webhook if a match happens.
```

geyslan (changes requested):
1st pass.

pkg/cmd/printer/policy.go
```diff
@@ -108,21 +118,28 @@ func (pp *PolicyEventPrinter) Print(event trace.Event) {
...
 		// avoid executing the same printer twice
-		if _, ok := executed[c]; ok {
+		if _, ok := executed[action]; ok {
 			continue
 		}
 
-		c <- event
+		eventChans, ok := pp.printerMap[action]
+		if !ok {
+			logger.Fatalw("printer not found", "printerName", action)
```
geyslan, line 134: If it's fatal, should we call panic() in this case?

geyslan (approved):
LGTM

yanivagman (approved):
LGTM

Issue #3171, opened 2023-05-30
matchedPoliciesNames has policy without the event 


josedonizetti:
## Description

The `matchedPoliciesNames` has a policy that didn't declare the event.

test1:
```
name: test1
description: test1
scope:
  - global
defaultAction: log
rules:
  - event: anti_debugging
```

test2:
```
name: test2
description: test2
scope:
  - global
defaultAction: log
rules:
  - event: ptrace
```

```
strace ls 
```

```
{"timestamp":1685463372089515478,"threadStartTime":1685463372074238271,"processorId":6,"processId":389929,"cgroupId":1719180,"threadId":389929,"parentProcessId":215443,"hostProcessId":389929,"hostThreadId":389929,"hostParentProcessId":215443,"userId":1000,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"strace","hostName":"josedonizetti-x","containerId":"","container":{},"kubernetes":{},"eventId":"101","eventName":"ptrace","matchedPolicies":["test2","test1"],"argsNum":4,"returnValue":0,"syscall":"ptrace","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"args":[{"name":"request","type":"string","value":"PTRACE_SYSCALL"},{"name":"pid","type":"pid_t","value":389932},{"name":"addr","type":"void*","value":"0x0"},{"name":"data","type":"void*","value":"0x0"}]}
```

rafaeldtinoco:
I think it might be even worse than that: https://github.com/aquasecurity/tracee/issues/3184

PR #3183, opened 2023-06-01
ebpf: Pipeline and BPF Network policy matching

### LINKED ISSUE #3184: network events are not working correctly with policies
## Description

Use the following test as a base line (ping -4 and ping -6):
```
$ ping -4 -c 1 -I 172.16.254.1 172.16.254.2 ; ping -6 -c 1 -I fd6e:a63d:71f:2f4::ff:1 fd6e:a63d:71f:2f4::ff:2
```

Execute tracee with a single scope in the command line:

```
$ sudo ./dist/tracee --filter event=net_packet_ipv4 --filter event=net_packet_icmp
 --filter event=net_packet_icmpv6 --filter comm=ping
02:42:40:518141  1000   ping             31358   31358   0                net_packet_ipv4           src: 172.16.254.1, dst: 172.16.254.2, proto_ipv4: {4 5 0 84 19254 2 0 64 ICMPv4 39757 172.16.254.1 172.16.254.2}
02:42:40:518141  1000   ping             31358   31358   0                net_packet_icmp           src: 172.16.254.1, dst: 172.16.254.2, proto_icmp: {EchoRequest 14503 51 1}
02:42:40:518190  1000   ping             31358   31358   0                net_packet_ipv4           src: 172.16.254.2, dst: 172.16.254.1, proto_ipv4: {4 5 0 84 17765 0 0 64 ICMPv4 57630 172.16.254.2 172.16.254.1}
02:42:40:518190  1000   ping             31358   31358   0                net_packet_icmp           src: 172.16.254.2, dst: 172.16.254.1, proto_icmp: {EchoReply 16551 51 1}
02:42:40:520708  1000   ping             31361   31361   0                net_packet_icmpv6         src: fd6e:a63d:71f:2f4::ff:1, dst: fd6e:a63d:71f:2f4::ff:2, proto_icmpv6: {EchoRequest 21409}
02:42:40:520754  1000   ping             31361   31361   0                net_packet_icmpv6         src: fd6e:a63d:71f:2f4::ff:2, dst: fd6e:a63d:71f:2f4::ff:1, proto_icmpv6: {EchoReply 21153}

End of events stream
Stats: {EventCount:6 EventsFiltered:0 NetCapCount:0 BPFLogsCount:0 ErrorCount:0 LostEvCount:0 LostWrCount:0 LostNtCapCount:0 LostBPFLogsCount:0}
```

WORKS AS EXPECTED.

----

Now do the same but using 2 policy files instead:

`net_packet_icmp.yaml`
```
name: net_packet_icmp
description: net_packet_icmp test
scope:
  - comm=ping
defaultAction: log
rules:
  - event: net_packet_icmp
```
`net_packet_icmpv6.yaml`
```
name: net_packet_icmpv6
description: net_packet_icmpv6 test
scope:
  - comm=ping
defaultAction: log
rules:
  - event: net_packet_icmpv6
```

And run tracee using those 2 policies:

```
$ sudo ./dist/tracee --config ~/policies/config.yaml --policy ~/policies/network/ 
02:40:11:421982  1000   ping             30984   30984   0                net_packet_icmp           src: 172.16.254.1, dst: 172.16.254.2, proto_icmp: {EchoRequest 29987 47 1}
02:40:11:422041  1000   ping             30984   30984   0                net_packet_icmp           src: 172.16.254.2, dst: 172.16.254.1, proto_icmp: {EchoReply 32035 47 1}
02:40:11:424594  1000   ping             30985   30985   0                net_packet_icmpv6         src: fd6e:a63d:71f:2f4::ff:1, dst: fd6e:a63d:71f:2f4::ff:2, proto_icmpv6: {EchoRequest 23837}
02:40:11:424634  1000   ping             30985   30985   0                net_packet_icmpv6         src: fd6e:a63d:71f:2f4::ff:2, dst: fd6e:a63d:71f:2f4::ff:1, proto_icmpv6: {EchoReply 23581}

End of events stream
Stats: {EventCount:4 EventsFiltered:0 NetCapCount:0 BPFLogsCount:0 ErrorCount:0 LostEvCount:0 LostWrCount:0 LostNtCapCount:0 LostBPFLogsCount:0}
```

It works..

BUUUUUUUUUT, add another policy called `net_packet_ipv4.yaml`:

```
name: net_packet_ipv4
description: net_packet_ipv4 test
scope:
  - comm=ping
defaultAction: log
rules:
  - event: net_packet_ipv4
```

And run tracee again with the same command (using all policies):

```
$ sudo ./dist/tracee --config ~/policies/config.yaml --policy ~/policies/network/ 
02:41:45:467755  1000   ping             31221   31221   0                net_packet_ipv4           src: 172.16.254.1, dst: 172.16.254.2, proto_ipv4: {4 5 0 84 13929 2 0 64 ICMPv4 45082 172.16.254.1 172.16.254.2}
02:41:45:467802  1000   ping             31221   31221   0                net_packet_ipv4           src: 172.16.254.2, dst: 172.16.254.1, proto_ipv4: {4 5 0 84 14237 0 0 64 ICMPv4 61158 172.16.254.2 172.16.254.1}

End of events stream
Stats: {EventCount:2 EventsFiltered:0 NetCapCount:0 BPFLogsCount:0 ErrorCount:0 LostEvCount:0 LostWrCount:0 LostNtCapCount:0 LostBPFLogsCount:0}
```

It only picks `net_packet_ipv4` now.

geyslan:

commit 9332b938f55eb81bcbb7efd4dc9dbaaeb2d9ebba (HEAD -> 3171-event-skip, origin/3171-event-skip)
Author: Geyslan Gregório <geyslan@gmail.com>
Date:   Fri Jun 2 10:12:36 2023 -0300

    ebpf: pipeline: add comment to matchPolicies

commit 52edc8df29ba8dc3be1261610cc1b21b06c5997f
Date:   Fri Jun 2 06:49:11 2023 +0000

    network: send network events if any policy is matched
    
    Check if network event has to be sumitted: if any of the policies matched,
    submit. All network events are base events (to be derived in userland).
    This means that if any of the policies need a network event, we can
    submit it and let userland deal with it.
    
    Fixes: #3184
    
    Co-authored-by: Jose Donizetti <jdbjunior@gmail.com>

commit d233ade13f65e8f712ed4670cf988b5317981c02
Date:   Fri Jun 2 05:51:55 2023 +0000

    network: fix net_packet_icmpv6 event dependency

commit 1129fbc3fc43eed8b67c6ca528f21b42976d8c3c
Date:   Fri Jun 2 04:29:48 2023 +0000

    policies: improve comments and add couple of TODOs
    
    TODOs added:
    
    1. check if base events should set kernel policy bitmask if the events
       to be derived from them.
    
    2. make load and stores atomic for ContainerFilterEnabled() function

commit 15c67ff0713d2b001311b49b19054ecac3d8b6e0
Date:   Fri Jun 2 03:26:49 2023 +0000

    policies: document matchPolicies for a first time reader

commit aed8b155c5e718a4c61930740e7b2df58f86e387
Date:   Fri Jun 2 02:40:19 2023 +0000

    policies: turn setMatchedPolicies into a nested func
    
    - removes it from ebpf package as a method.
    - add godoc to the invokeInitEvents() method.

commit 52d8fab9b094bbe3a92f338344da2cb4cef7dce9
Author: Geyslan Gregório <geyslan@gmail.com>
Date:   Wed May 31 21:34:36 2023 -0300

    ebpf: pipeline: use MatchedPoliciesUser
    
    This fix the problem of matched policies not being filtered out in base
    and derived events by propagating the MatchedPoliciesKernel through the
    pipeline without changes - as it came from the kernel -, and by making
    use of the new MatchedPoliciesUser field to store the filtered matched
    policies which must be used in the sink stage.
    
    It also moves the matched scopes name fetching to the sink stage,
    improving the performance of the pipeline.
    
    As well, it removes the shouldProcessEvent() wrapper function as it is
    kind of redundant.
    
    Finally, it does some minor refactoring in the pipeline code.
    
    Co-authored-by: Jose Donizetti <jdbjunior@gmail.com>

## Fix: #3171

```yaml
name: pol1
description: general policy
scope:
  - global
defaultAction: log
rules:
  - event: anti_debugging
```

```yaml
name: pol2
description: general policy
scope:
  - global
defaultAction: log
rules:
  - event: ptrace
    filter:
      - retval!=0
```

```yaml
name: pol3
description: general policy
scope:
  - global
defaultAction: log
rules:
  - event: ptrace
  - event: anti_debugging
```

```yaml
name: pol4
description: general policy
scope:
  - global
defaultAction: log
rules:
  - event: ptrace
  - event: anti_debugging
    filter:
      - retval!=0
```

```yaml
name: pol5
description: general policy
scope:
  - global
defaultAction: log
rules:
  - event: anti_debugging
    filter:
      - retval=0
```

With some debugging outputs to show matched fields.

`sudo ./dist/tracee -p pol1.yaml -p pol2.yaml -p pol3.yaml -p pol4.yaml -p pol5.yaml -o json 1>/dev/null`

```
SINK: ptrace, [pol3 pol4], 11111, 1100
SINK: ptrace, [pol4 pol3], 11111, 1100
SINK: anti_debugging, [pol5 pol1 pol3], 11111, 10101
```

---

`sudo ./dist/tracee -p pol1.yaml -p pol2.yaml -p pol3.yaml -p pol4.yaml -p pol5.yaml -o json`

```json
{"timestamp":1685628313720348881,"threadStartTime":1685628313717928337,"processorId":8,"processId":391890,"cgroupId":22422,"threadId":391890,"parentProcessId":314166,"hostProcessId":391890,"hostThreadId":391890,"hostParentProcessId":314166,"userId":1000,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"strace","hostName":"hb","containerId":"","container":{},"kubernetes":{},"eventId":"101","eventName":"ptrace","matchedPolicies":["pol3","pol4"],"argsNum":4,"returnValue":0,"syscall":"ptrace","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"args":[{"name":"request","type":"string","value":"PTRACE_SEIZE"},{"name":"pid","type":"pid_t","value":391891},{"name":"addr","type":"void*","value":"0x0"},{"name":"data","type":"void*","value":"0x0"}]}
{"timestamp":1685628313720813154,"threadStartTime":1685628313720738423,"processorId":0,"processId":391892,"cgroupId":22422,"threadId":391892,"parentProcessId":391890,"hostProcessId":391892,"hostThreadId":391892,"hostParentProcessId":391890,"userId":1000,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"strace","hostName":"hb","containerId":"","container":{},"kubernetes":{},"eventId":"101","eventName":"ptrace","matchedPolicies":["pol3","pol4"],"argsNum":4,"returnValue":0,"syscall":"ptrace","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"args":[{"name":"request","type":"string","value":"PTRACE_TRACEME"},{"name":"pid","type":"pid_t","value":0},{"name":"addr","type":"void*","value":"0x0"},{"name":"data","type":"void*","value":"0x0"}]}
{"timestamp":1685628313720813154,"threadStartTime":1685628313720738423,"processorId":0,"processId":391892,"cgroupId":22422,"threadId":391892,"parentProcessId":391890,"hostProcessId":391892,"hostThreadId":391892,"hostParentProcessId":391890,"userId":1000,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"strace","hostName":"hb","containerId":"","container":{},"kubernetes":{},"eventId":"6018","eventName":"anti_debugging","matchedPolicies":["pol5","pol1","pol3"],"argsNum":0,"returnValue":0,"syscall":"","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"args":[],"metadata":{"Version":"1","Description":"A process used anti-debugging techniques to block a debugger. Malware use anti-debugging to stay invisible and inhibit analysis of their behavior.","Tags":null,"Properties":{"Category":"defense-evasion","Kubernetes_Technique":"","Severity":1,"Technique":"Debugger Evasion","external_id":"T1622","id":"attack-pattern--e4dc8c01-417f-458d-9ee0-bb0617c1b391","signatureID":"TRC-102","signatureName":"Anti-Debugging detected"}}}
```

---

```yaml
name: test1
description: test1
scope:
  - global
defaultAction: log
rules:
  - event: anti_debugging
```

```
name: test2
description: test2
scope:
  - global
defaultAction: log
rules:
  - event: ptrace
```

`sudo ./dist/tracee -p test1.yaml -p test2.yaml -o json`

```json
{"timestamp":1685582076882541151,"threadStartTime":1685582076868546156,"processorId":0,"processId":286555,"cgroupId":18028,"threadId":286555,"parentProcessId":126951,"hostProcessId":286555,"hostThreadId":286555,"hostParentProcessId":126951,"userId":1000,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"strace","hostName":"hb","containerId":"","container":{},"kubernetes":{},"eventId":"101","eventName":"ptrace","matchedPolicies":["test2"],"argsNum":4,"returnValue":0,"syscall":"ptrace","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"args":[{"name":"request","type":"string","value":"PTRACE_SEIZE"},{"name":"pid","type":"pid_t","value":286556},{"name":"addr","type":"void*","value":"0x0"},{"name":"data","type":"void*","value":"0x0"}]}
{"timestamp":1685582076883117796,"threadStartTime":1685582076882936399,"processorId":4,"processId":286557,"cgroupId":18028,"threadId":286557,"parentProcessId":286555,"hostProcessId":286557,"hostThreadId":286557,"hostParentProcessId":286555,"userId":1000,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"strace","hostName":"hb","containerId":"","container":{},"kubernetes":{},"eventId":"101","eventName":"ptrace","matchedPolicies":["test2"],"argsNum":4,"returnValue":0,"syscall":"ptrace","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"args":[{"name":"request","type":"string","value":"PTRACE_TRACEME"},{"name":"pid","type":"pid_t","value":0},{"name":"addr","type":"void*","value":"0x0"},{"name":"data","type":"void*","value":"0x0"}]}
{"timestamp":1685582076883117796,"threadStartTime":1685582076882936399,"processorId":4,"processId":286557,"cgroupId":18028,"threadId":286557,"parentProcessId":286555,"hostProcessId":286557,"hostThreadId":286557,"hostParentProcessId":286555,"userId":1000,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"strace","hostName":"hb","containerId":"","container":{},"kubernetes":{},"eventId":"6018","eventName":"anti_debugging","matchedPolicies":["test1"],"argsNum":0,"returnValue":0,"syscall":"","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"args":[],"metadata":{"Version":"1","Description":"A process used anti-debugging techniques to block a debugger. Malware use anti-debugging to stay invisible and inhibit analysis of their behavior.","Tags":null,"Properties":{"Category":"defense-evasion","Kubernetes_Technique":"","Severity":1,"Technique":"Debugger Evasion","external_id":"T1622","id":"attack-pattern--e4dc8c01-417f-458d-9ee0-bb0617c1b391","signatureID":"TRC-102","signatureName":"Anti-Debugging detected"}}}
```

## Fix #3184

`net_packet_icmp.yaml`

```
name: net_packet_icmp
description: net_packet_icmp test
scope:
  - comm=ping
defaultAction: log
rules:
  - event: net_packet_icmp
```

`net_packet_icmpv6.yaml`

```
name: net_packet_icmpv6
description: net_packet_icmpv6 test
scope:
  - comm=ping
defaultAction: log
rules:
  - event: net_packet_icmpv6
```

`sudo ./dist/tracee -p net_packet_icmp.yaml -p net_packet_icmpv6.yaml -o json`

`ping -4 -c 1 127.0.0.1 ; ping -6 -c 1 ::`

```json
{"timestamp":1685713410424513412,"threadStartTime":1685713410408173831,"processorId":1,"processId":1106232,"cgroupId":13238,"threadId":1106232,"parentProcessId":1105989,"hostProcessId":1106232,"hostThreadId":1106232,"hostParentProcessId":1105989,"userId":1000,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"ping","hostName":"ubuntu-jammy","containerId":"","container":{},"kubernetes":{},"eventId":"2004","eventName":"net_packet_icmp","matchedPolicies":["net_packet_icmp"],"argsNum":3,"returnValue":0,"syscall":"sendto","stackAddresses":[0],"contextFlags":{"containerStarted":false,"isCompat":false},"args":[{"name":"src","type":"const char*","value":"127.0.0.1"},{"name":"dst","type":"const char*","value":"127.0.0.1"},{"name":"proto_icmp","type":"trace.ProtoICMP","value":{"typeCode":"EchoRequest","checksum":40537,"id":2,"seq":1}}]}
{"timestamp":1685713410424561408,"threadStartTime":1685713410408173831,"processorId":1,"processId":1106232,"cgroupId":13238,"threadId":1106232,"parentProcessId":1105989,"hostProcessId":1106232,"hostThreadId":1106232,"hostParentProcessId":1105989,"userId":1000,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"ping","hostName":"ubuntu-jammy","containerId":"","container":{},"kubernetes":{},"eventId":"2004","eventName":"net_packet_icmp","matchedPolicies":["net_packet_icmp"],"argsNum":3,"returnValue":0,"syscall":"","stackAddresses":[0],"contextFlags":{"containerStarted":false,"isCompat":false},"args":[{"name":"src","type":"const char*","value":"127.0.0.1"},{"name":"dst","type":"const char*","value":"127.0.0.1"},{"name":"proto_icmp","type":"trace.ProtoICMP","value":{"typeCode":"EchoReply","checksum":42585,"id":2,"seq":1}}]}
{"timestamp":1685713410427638029,"threadStartTime":1685713410425449403,"processorId":1,"processId":1106233,"cgroupId":13238,"threadId":1106233,"parentProcessId":1105989,"hostProcessId":1106233,"hostThreadId":1106233,"hostParentProcessId":1105989,"userId":1000,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"ping","hostName":"ubuntu-jammy","containerId":"","container":{},"kubernetes":{},"eventId":"2005","eventName":"net_packet_icmpv6","matchedPolicies":["net_packet_icmpv6"],"argsNum":3,"returnValue":0,"syscall":"sendto","stackAddresses":[0],"contextFlags":{"containerStarted":false,"isCompat":false},"args":[{"name":"src","type":"const char*","value":"::1"},{"name":"dst","type":"const char*","value":"::1"},{"name":"proto_icmpv6","type":"trace.ProtoICMPv6","value":{"typeCode":"EchoRequest","checksum":62159}}]}
{"timestamp":1685713410427664007,"threadStartTime":1685713410425449403,"processorId":1,"processId":1106233,"cgroupId":13238,"threadId":1106233,"parentProcessId":1105989,"hostProcessId":1106233,"hostThreadId":1106233,"hostParentProcessId":1105989,"userId":1000,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"ping","hostName":"ubuntu-jammy","containerId":"","container":{},"kubernetes":{},"eventId":"2005","eventName":"net_packet_icmpv6","matchedPolicies":["net_packet_icmpv6"],"argsNum":3,"returnValue":0,"syscall":"","stackAddresses":[0],"contextFlags":{"containerStarted":false,"isCompat":false},"args":[{"name":"src","type":"const char*","value":"::1"},{"name":"dst","type":"const char*","value":"::1"},{"name":"proto_icmpv6","type":"trace.ProtoICMPv6","value":{"typeCode":"EchoReply","checksum":61903}}]}
```

### 3. Other comments

Fix: #3171
Fix: #3184


geyslan:
These (base | derived) cases will be able to be tested more accurately as we improve the new integration tests that will be merged soon.

NDStrahilevitz (commented):
Not sure why you removed the wrapping `shouldProcessEvent`, as the `if` check done has not seemed to change.

geyslan:
> Not sure why you removed the wrapping `shouldProcessEvent`, as the `if` check done has not seemed to change.

I'm trying to make the involved logic clearer from the pipeline code view. The wrap was masking the assignments to the event even the computing function isn't changing it at all.

@rafaeldtinoco found another issue #3184 that probably can change the code we are discussing even more.

geyslan:
> Not sure why you removed the wrapping `shouldProcessEvent`, as the `if` check done has not seemed to change.

I'm trying to make the involved logic clearer from the pipeline code view. The wrap was masking the assignments to the event even the computing function isn't changing it at all.

@rafaeldtinoco found another issue #3184 that probably can change the code we are discussing even more.

pkg/ebpf/events_pipeline.go
```diff
@@ -281,35 +282,47 @@ func (t *Tracee) decodeEvents(outerCtx context.Context, sourceChan chan []byte)
...
+		_, ok := p.EventsToTrace[eventID]
+		if !ok {
+			// The remaining filter checks can be skipped, allowing to continue to the next policy.
+			// This treats the current policy as unmatched, even though this event could potentially
+			// match as a derived event in later stages. That is why it is crucial to propagate
+			// the event matching bit as-is, enabling checks for possible derived events.
+
+			// Therefore, the field to be used for actual policy matching checks is the policy name.
 			continue
 		}
```
yanivagman, line 312: In the sink stage we do the following:
```
id := events.ID(event.EventID)
event.MatchedPolicies &= t.events[id].emit
if event.MatchedPolicies == 0 {
    continue
}
```

1. If we would have done the mapping to policy names only at the sink stage (after the lines I pasted above) I believe this would have also solve the problem, no?
2. Let's assume we only requested for a derived event in some policy. So now the policy bitmap (MatchedPolicies) is not zero, but the matchedPoliciesNames is empty for the original event, right? Won't this cause the original event, which should not be printed, to be emitted (with an empty matched policy list)?

yanivagman: > 2\. Let's assume we only requested for a derived event in some policy. So now the policy bitmap (MatchedPolicies) is not zero, but the matchedPoliciesNames is empty for the original event, right? Won't this cause the original event, which should not be printed, to be emitted (with an empty matched policy list)?

Actually the lines I pasted above should resolve this I believe.

And about the first point I wrote - is there a reason not to do the bitmap to names mapping in the sink stage (even if only for performance reasons)

yanivagman: I think that eventually we should do the following:
1. Keep the original matched policies bitmap we got from the kernel untouched, to be used by derived events as well
2. Make a copy of this bitmap, which will be used to check for matched policies of the current events
3. Continue with the pipeline in any case if the event has derived events (like we do now)
4. In the sink stage, only emit the event if the copy of the bitmap is non zero and event should be emitted
5. Map the matchedPolicies bitmap to policy name in the sink stage if we passed all above

WDYT?

geyslan: @yanivagman, I'm doing some tests and what you proposed makes perfect sense:

image

I was actually going to the same direction after @rafaeldtinoco's findings in https://github.com/aquasecurity/tracee/issues/3184#issuecomment-1571725177:

>  I'm really considering that we have to propagate the MatchedScopes bitmap without bit clearing at all.

@josedonizetti and @NDStrahilevitz FYI

geyslan:
> Not sure why you removed the wrapping `shouldProcessEvent`, as the `if` check done has not seemed to change.

Returned it renaming to `shouldEmitEvent()`.

pkg/ebpf/events_pipeline.go
```diff
@@ -262,7 +262,7 @@ func (t *Tracee) decodeEvents(outerCtx context.Context, sourceChan chan []byte)
 				Syscall:             syscall,
 			}
 
-			if !t.shouldProcessEvent(&evt) {
+			if !t.shouldEmitEvent(&evt) {
```
yanivagman, line 265: emit only happens at the last stage of the pipeline.
I think that removing this wrapper and calling it directly like you did was better, but if keeping the wrapper consider renaming it.

pkg/ebpf/events_pipeline.go
```diff
@@ -281,35 +281,34 @@ func (t *Tracee) decodeEvents(outerCtx context.Context, sourceChan chan []byte)
...
 
 	for p := range t.config.Policies.Map() {
 		bitOffset := uint(p.ID)
 
 		// Events submitted with matching policies.
 		// The policy must have its bit cleared when it does not match.
-		if !utils.HasBit(matchedPolicies, bitOffset) {
+		if !utils.HasBit(bitmap, bitOffset) {
 			continue
 		}
```
yanivagman, line 299: If the event was not chosen by this policy, you can remove the matchedPolicies bit of this policy and continue (like you did before, but also clear the bit of the policy before continue)

pkg/ebpf/events_pipeline.go
```diff
@@ -281,35 +281,34 @@ func (t *Tracee) decodeEvents(outerCtx context.Context, sourceChan chan []byte)
...
 }
 
-// computePolicies iterates through the policies that do the filtering in user space, checking whether an event should be considered.
-// If it should not, it sets the respective offset to 0.
-// Finally it returns computed policies and the names of the policies that matched.
-func (t *Tracee) computePolicies(event *trace.Event) (uint64, []string) {
+// computePoliciesMatches iterates through the policies that do the filtering in user space.
+// If the filter doesn't match, the policy bit is cleared from the matched policies bitmap.
+// Finally, the filtered bitmap is stored in the event and returned.
+func (t *Tracee) computePoliciesMatches(event *trace.Event) uint64 {
```
yanivagman, line 287: I would rename to matchPolicies which is more concise

pkg/events/trigger/context.go
```diff
@@ -74,6 +74,7 @@ func (store *context) Apply(event trace.Event) (trace.Event, error) {
 	invoking.ReturnValue = 0
 	invoking.Args = make([]trace.Argument, len(event.Args))
 	invoking.MatchedPolicies = event.MatchedPolicies
+	invoking.MatchedPoliciesFiltered = event.MatchedPoliciesFiltered
```
geyslan, line 77: Added this, as this part seems to be creating a new derived based on the event.

types/trace/trace.go
```diff
@@ -14,35 +14,36 @@ import (
...
+	PIDNS                   int          `json:"pidNamespace"`
+	ProcessName             string       `json:"processName"`
+	HostName                string       `json:"hostName"`
+	ContainerID             string       `json:"containerId"`
+	Container               Container    `json:"container,omitempty"`
+	Kubernetes              Kubernetes   `json:"kubernetes,omitempty"`
+	EventID                 int          `json:"eventId,string"`
+	EventName               string       `json:"eventName"`
+	MatchedPolicies         uint64       `json:"-"`
+	MatchedPoliciesFiltered uint64       `json:"-"`
```
geyslan, line 38:
```suggestion
	MatchedPoliciesKernel   uint64       `json:"-"`
	MatchedPoliciesUserland uint64       `json:"-"`
```

Perhaps? Suggestions?

yanivagman: remove the land and I'm +1 on the rename
`MatchedPoliciesKernel`
`MatchedPoliciesUser`

pkg/bufferdecoder/protocol.go
```diff
@@ -23,29 +23,29 @@ const (
...
+	HostPpid              uint32
+	Uid                   uint32
+	MntID                 uint32
+	PidID                 uint32
+	Comm                  [16]byte
+	UtsName               [16]byte
+	Flags                 uint32
+	EventID               events.ID // int32
+	Syscall               int32
+	MatchedPoliciesKernel uint64
```
yanivagman, line 43: I think for the protocol this can still be called matchedPolicies without kernel

types/trace/trace.go
```diff
@@ -14,35 +14,36 @@ import (
...
+	ProcessName           string       `json:"processName"`
+	HostName              string       `json:"hostName"`
+	ContainerID           string       `json:"containerId"`
+	Container             Container    `json:"container,omitempty"`
+	Kubernetes            Kubernetes   `json:"kubernetes,omitempty"`
+	EventID               int          `json:"eventId,string"`
+	EventName             string       `json:"eventName"`
+	MatchedPoliciesKernel uint64       `json:"-"`
+	MatchedPoliciesUser   uint64       `json:"-"`
+	MatchedPoliciesNames  []string     `json:"matchedPolicies,omitempty"`
```
yanivagman, line 39: Now this can simply be called MatchedPolicies

yanivagman (approved):
LGTM, nit naming comments

go.mod
```diff
@@ -167,3 +167,5 @@ require (
 )
 
 replace github.com/kubernetes/cri-api => k8s.io/cri-api v0.23.5-rc.0
+
+replace github.com/aquasecurity/tracee/types => ./types
```
geyslan, line 171: After approval this commit must be cherry-picked, merged into main and
removed from this branch.

pkg/ebpf/finding.go
```diff
@@ -29,32 +29,33 @@ func newEvent(id int, f detect.Finding, s trace.Event) *trace.Event {
...
+		PIDNS:                 s.PIDNS,
+		ProcessName:           s.ProcessName,
+		HostName:              s.HostName,
+		Container:             s.Container,
+		Kubernetes:            s.Kubernetes,
+		ReturnValue:           s.ReturnValue,
+		StackAddresses:        s.StackAddresses,
+		ContextFlags:          s.ContextFlags,
+		MatchedPoliciesKernel: s.MatchedPoliciesKernel,
+		MatchedPoliciesUser:   s.MatchedPoliciesUser,
```
geyslan, line 55: Reflected the changes to finding.

types/trace/trace.go
```diff
@@ -14,35 +14,36 @@ import (
...
+	ProcessName           string       `json:"processName"`
+	HostName              string       `json:"hostName"`
+	ContainerID           string       `json:"containerId"`
+	Container             Container    `json:"container,omitempty"`
+	Kubernetes            Kubernetes   `json:"kubernetes,omitempty"`
+	EventID               int          `json:"eventId,string"`
+	EventName             string       `json:"eventName"`
+	MatchedPoliciesKernel uint64       `json:"-"`
+	MatchedPoliciesUser   uint64       `json:"-"`
+	MatchedPolicies       []string     `json:"matchedPolicies,omitempty"`
```
geyslan, line 39: Actual changes.

pkg/ebpf/events_pipeline.go
```diff
@@ -413,8 +411,9 @@ func (t *Tracee) processEvents(ctx context.Context, in <-chan *trace.Event) (<-c
 					logger.Debugw("False container positive", "event.Timestamp", event.Timestamp, "eventId", eventId)
 
 					// filter container policies out
-					utils.ClearBits(&event.MatchedPolicies, policiesWithContainerFilter)
-					if event.MatchedPolicies == 0 {
+					utils.ClearBits(&event.MatchedPoliciesKernel, policiesWithContainerFilter)
```
rafaeldtinoco, line 414: I have sent you an offline message about atomic store/loads for the policiesWithContainerFilter (I'm adding a TODO in a comment for the ContainerFilterEnabled() Function). Also, note that protecting the policiesWithContainerFilter with atomics won't be enough when we grow
from 64 policies (u64) to 128 (2x the CPU word size). You will require
a mutex internal to ContainerFilterEnabled() logic. Either way, check
my offline message on that!

pkg/ebpf/events_pipeline.go
```diff
@@ -472,7 +471,7 @@ func (t *Tracee) deriveEvents(ctx context.Context, in <-chan *trace.Event) (
 					case events.PrintMemDump:
 					default:
 						// Derived events might need filtering as well
-						if !t.shouldProcessEvent(&derivative) {
+						if t.matchPolicies(&derivative) == 0 {
```
rafaeldtinoco, line 474: Ok, but a question that raises now: is copying the event really necessary now that we do the matched policies copy and differentiation in between kernel and userland ? Not copying/duplicating a high frequency of derived events (like net_packet_XXX) could be good for the pipeline (and us =D).

yanivagman: Good question, probably no need to copy.
Just to avoid regression surprises, let's do it (remove the copy) right after the release

pkg/ebpf/events_pipeline.go
geyslan: > You will require a mutex internal to ContainerFilterEnabled() logic. Either way, check my offline message on that!

Indeed. This must be part of the API discussion as: embed and lock all policies computation when changes are required.

pkg/ebpf/events_pipeline.go
geyslan: Your point makes sense and got me thinking that since we're ignoring events not set in matchPolicies() gets its bit intact for later stages of the pipeline (i.e. derivation).

So, perhaps this check of derivation can be removed:

image

I didn't check it yet and won't before get the fixes merged. As Yaniv suggested, let's move on and try those changes in the next release.

yanivagman: This check of derivation is here for performance reasons - if the event doesn't have derived events and no matched policies then there is no point to push it to the next pipeline stage

pkg/ebpf/events_pipeline.go
```diff
@@ -233,40 +231,47 @@ func (t *Tracee) decodeEvents(outerCtx context.Context, sourceChan chan []byte)
...
+			// If there aren't any policies that need filtering in userland, tracee **may** skip
+			// this event, as long as there aren't any derivatives that depend on it. Some base
+			// events (for derivative ones) might not have set related policy bit, thus the need
+			// to continue with those within the pipeline.
+			//
+			// TODO: should the base event set policy bit for policies that depend on their
+			//       derivative events ? (check matchPolicies)
+			//
+
+			if t.matchPolicies(&evt) == 0 {
```
josedonizetti, line 272: This logic will break signatures. Because just like derivatives the signatures are not the events coming from kernel, they expect a base event to be parsed in a later stage. For `anti_debugging` for example, we need to allow ptrace to pass because it is a base event, but ptrace itself wasn't matched in any policy.

josedonizetti (changes requested):

pkg/ebpf/events_pipeline.go
josedonizetti: updating the branch worked, the bug was in a previous version

josedonizetti: but let's update the comment above to mention signatures too

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4969,25 +4969,32 @@ statfunc enum event_id_e net_packet_to_net_event(net_packet_t packet_type)
...
+    // Check if event has to be sumitted: if any of the policies matched, submit. All network events
+    // are base events (to be derived in userland). This means that if any of the policies need a
+    // network event, kernel can submit and let userland deal with it.
+
+    event_config_t *evt_config = bpf_map_lookup_elem(&events_map, &evt_id);
+    if (evt_config == NULL)
+        return 0;
+
+    // Any policy matched is enough to submit the net event
+    return evt_config->submit_for_policies;
```
yanivagman, line 4997: This is the configuration of the event, seems that no matching is taken into account now

rafaeldtinoco: It should whenever I start applying kernel filters to network context (Which isn't the case YET). But it will.

rafaeldtinoco: I'll probably create the same logic we do have for should_filter() but targeting network filters (to be done in kernel). Then I should consider the masking because the in-kernel filtering might have cleaned the ctx->matched_policies (which is not the case yet).

pkg/ebpf/events_pipeline.go
rafaeldtinoco: Yes, @geyslan is testing the "inner IF" need (after his fixes @josedonizetti and I think the inner IF isn't needed). He tested E2E without it and it works fine (and our E2E as well). If you know any particular corner case that still needs the inner IF:

```go
if _, ok := t.eventDerivations[eventId]; !ok {
```

let us know, but it looks like matchPolicies() takes care of everything needed now.

rafaeldtinoco: But this condition won't happen. All base events are marked to the policies that have derived events as that policies needed event as well. So the base events should never have the policy bitmask zeroed and matchPolicies() will take care of everything (I believe @geyslan is testing that now or has tested it after discussion with @josedonizetti and I).

rafaeldtinoco: Actually, if filters are applied to the base event, the bitmask might have been cleaned and then we would have to check the base event for derived events. So this check might be right @geyslan told me so, and I agreed. I asked for a comment there explicitly saying that (maybe giving one example).

pkg/ebpf/c/tracee.bpf.c
yanivagman: But is it correct?
The scope was checked for this task and saved in matched scopes.
Here you don't take the scope into account only the events. What if the scope only wants events for e.g. the ping process and nothing else? Does it work as is now?

yanivagman: Let me also give an example. 
Say that current process is 'ls' and you have a global policy that traces some network event, e.g. net_packet_ipv4, so match_policies check pass.
Also say you have a policy for ping that wants icmp events.
Now the icmp event is not part of the global policy, yet you will provide matched policy for the ping process. Am I wrong?

pkg/ebpf/events_pipeline.go
geyslan: Consider:

A policy with:

- 1 base event (with an arg filter)
- 1 event derived from that same base (without any filter - i.e. any base sent must arrive at the printer)

The base (first to be checked) policy has a unmatch, so it has the policy bit cleared.

When returning from `matchPolicies()`, as it is just one policy, this return has the bit cleared and it's 0, so if we don't check, as a requirement, that this event has derivatives from it, the event that should derivate goes to /dev/null.

geyslan:
Folks, I'm merging this, but please continue to discuss.

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: Yes you are correct indeed. We should do what you're saying. Thanks for the attention call for this, you're correct.

rafaeldtinoco: @yanivagman spoke to Geyslan, he will add a patch AND'ing the matched policies from "network context" (which is 1:1 to the task context using the socket). I'll test with the network profiles I have and we can merge that.

rafaeldtinoco: We're not updating the event matched policies for the network events as we don't have (yet) ni-kernel filters for those. Once I add, then I'll change this function to reflect the in kernel filters in the mask as well.

geyslan: https://github.com/aquasecurity/tracee/pull/3198

Issue #3186, opened 2023-06-01
config yaml file can't set output to json


rafaeldtinoco:
## Description

When using the following config file:

```yaml
blob-perf-buffer-size: 1024
install-path: /tmp/tracee
cache:
    - cache-type=mem
    - mem-cache-size=1024
capabilities: []
containers: true
crs: []
log:
    - info
output:
    - json
perf-buffer-size: 1024
pyroscope: false
rego: []
signatures-dir: ""
metrics: true
healthz: true
pprof: true
listen-addr: :3366
```

and executing

```
sudo ./dist/tracee --config ~/policies/config.yaml 
```

Tracee output isn't json as expected (should it be ?)

## Output of `tracee version`:

```
commit: 234b9d0d
```

## Output of `uname -a`:

```
v6.3.0
```

## Additional details 

![image](https://github.com/aquasecurity/tracee/assets/7395852/a29e3f57-f52e-4854-b95f-ebab6d33ff2d)

PR #3187, opened 2023-06-01
tracee: fix output configuration via configfile

josedonizetti:

Fix https://github.com/aquasecurity/tracee/issues/3186

The `c.Flags()` opinion should only be used for command line only arguments, anything else must use `viper` because that is how we process the config file https://github.com/aquasecurity/tracee/blob/main/cmd/tracee/cmd/root.go#L313

### 2. Explain how to test it

create the cfg.yaml below
```
blob-perf-buffer-size: 1024
install-path: /tmp/tracee
cache:
    - cache-type=mem
    - mem-cache-size=1024
capabilities: []
containers: true
crs: []
log:
    - info
output:
    - json
perf-buffer-size: 1024
pyroscope: false
rego: []
signatures-dir: ""
metrics: true
healthz: true
pprof: true
listen-addr: :3366
```

start tracee

```
sudo ./dist/tracee --config cfg.yaml 
```

### DIFF
--- a/pkg/cmd/cobra/cobra.go
+++ b/pkg/cmd/cobra/cobra.go
@@ -138,22 +138,17 @@ func GetTraceeRunner(c *cobra.Command, version string) (cmd.Runner, error) {
 		return runner, errors.New("policy and filter flags cannot be used together")
 	}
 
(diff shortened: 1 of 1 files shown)


geyslan (approved):
LGTM

Issue #3185, opened 2023-06-01
tracee: panic during initialization


rafaeldtinoco:
## Description

```
$ sudo ./dist/tracee --config ~/policies/config.yaml --policy ~/policies/network/ 
TIME             UID    COMM             PID     TID     RET              EVENT                     ARGS
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x110 pc=0x18506b8]

goroutine 28 [running]:
github.com/aquasecurity/tracee/pkg/ebpf.(*Tracee).enrichContainerEvents.func1()
        /home/rafaeldtinoco/work/ebpf/tracee-libbpfgo/pkg/ebpf/events_enrich.go:79 +0x128
created by github.com/aquasecurity/tracee/pkg/ebpf.(*Tracee).enrichContainerEvents
        /home/rafaeldtinoco/work/ebpf/tracee-libbpfgo/pkg/ebpf/events_enrich.go:73 +0x1f4
```

## Output of `tracee version`:

```
v0.14.2-56-g234b9d0d
```

## Output of `uname -a`:

```
Linux azure 6.3.2 #5 SMP Tue May 16 04:48:36 UTC 2023 aarch64 aarch64 aarch64 GNU/Linux
```

rafaeldtinoco:
Since we're passing a channel of pointers, it is likely that the *tracee.Event gets NULL in the pipeline in the moment of shutdown, or similar logic, I believe.

In the pipeline, the caching step checks if event is null before moving on in both 'in/out' channels. 

![image](https://github.com/aquasecurity/tracee/assets/7395852/afd87112-5875-44de-8bb6-ef012313c7fc)

I believe the same should happen in the `enrich` pipeline stage:

![image](https://github.com/aquasecurity/tracee/assets/7395852/7ef31396-fbae-4eb6-b467-e0044c0d2a82)

PR #3188, opened 2023-06-01
tracee: fix panic when ctrl-c after the boot

josedonizetti:

Fix https://github.com/aquasecurity/tracee/issues/3185

check if the event is nil before continuing on all states of the pipeline, to avoid panic when a `(ctrl+c seg faults)` is send just after the initialization.

stages of the pipeline:

- decode -> creates the event
- queue -> already checks for nil in both stages [queue](https://github.com/aquasecurity/tracee/blob/main/pkg/ebpf/events_pipeline.go#L120) and [dequeue](https://github.com/aquasecurity/tracee/blob/main/pkg/ebpf/events_pipeline.go#L138)
- sort -> [already checks for nil](https://github.com/aquasecurity/tracee/blob/main/pkg/events/sorting/sorting.go#L165)
- process -> fixed in this pr
- enrich -> fixed in this pr
- derive -> [already checks for nil](https://github.com/aquasecurity/tracee/blob/main/pkg/ebpf/events_pipeline.go#L447)
- signatures -> [check on the first part](https://github.com/aquasecurity/tracee/blob/main/pkg/ebpf/signature_engine.go#L54) but second stage of finding fixed in this pr
- sink  -> fixed in this pr


rafaeldtinoco (approved):
LGTM. Thanks for the fix!

PR #3195, opened 2023-06-01
fix: clean up untraced signatures

### LINKED ISSUE #3070: tracee single binary should not load plugin if traced events is not a signature event
## Description

After discussions at https://github.com/aquasecurity/tracee/discussions/3065#discussioncomment-5837732 (not necessiraly related to the issue being described here) and to check the following bug: https://github.com/aquasecurity/tracee/issues/2778 we should, perhaps, seek a way of NOT loading the plugin mechanism if events being traced aren't signature events.

This could avoid errors when trying to execute the `tracee` single binary when compiled statically, like described at https://github.com/aquasecurity/tracee/discussions/3065#discussioncomment-5837634

### LINKED ISSUE #3190: Signatures loaded into engine despite not being traced
I found this issue while accidentally breaking tracee, so it's not really visible when tracee is working as intended.
Running tracee (while the ebpf decoding was broken) as following gave the following logs:
```
tracee -f e=sched_process_exec
ctx {6662557490042479 6662557488064774 298763 768186 768186 704670 768186 768186 704670 1000 4026531841 4026531836 [115 104 0 0 0 0 0 0 0 0 0 0 0 0 0 0] [105 112 45 49 55 50 45 51 49 45 55 53 45 49 55 0] 0 713 59 1 0 0 1 0 [0]}
{"level":"error","ts":1685630571.0000381,"msg":"Handling event by signature Fileless execution detected: argument pathname is not initialized"}
14:42:50:999388  1000   sh               768186  768186  0                sched_process_exec        cmdpath: <nil>, pathname: <nil>, dev: <nil>, inode: <nil>, ctime: <nil>, inode_mode: <nil>, interpreter_pathname: <nil>{"level":"error","ts":1685630571.0001652,"msg":"Handling event by signature Rcd modification detected: argument pathname is not initialized"}
, interpreter_dev: <nil>, interpreter_inode: <nil>, interpreter_ctime: <nil>, argv: <nil>, interp: <nil>, stdin_type: <nil>, stdin_path: <nil>, invoked_from_kernel: <nil>, env: <nil>
{"level":"error","ts":1685630571.0002441,"msg":"Handling event by signature Scheduled tasks modification detected: argument pathname is not initialized"}
ctx {6662557491171189 6662557490777762 298763 768187 768187 768186 768187 768187 768186 1000 4026531841 4026531836 [119 104 105 99 104 0 0 0 0 0 0 0 0 0 0 0] [105 112 45 49 55 50 45 51 49 45 55 53 45 49 55 0] 0 713 59 1 0 0 2 0 [0]}
{"level":"error","ts":1685630571.0007372,"msg":"Handling event by signature Fileless execution detected: argument pathname is not initialized"}
```
In particular the last log indicates that the signature was working despite not being requested for tracing at all. I think when loading the signature event definition and then loading the signatures on start up, there needs to be a check of tracing those events in addition.

NDStrahilevitz:
```
Date:   Thu Jun 1 15:21:15 2023 +0000

    tracee: clean up untraced signatures
    
    Up to this commit, all signatures would be loaded to the engine even if
    tracing them was not requested. This would consume runtime resources.
    
    Fix this by verifying they were actually requested for tracing and
    removing those that weren't.
    
    In addition, if no signatures are traced post cleanup, disable the
    engine event pipeline stage.
```

FIx #3190 
Fix #3070

### DIFF
--- a/pkg/ebpf/tracee.go
+++ b/pkg/ebpf/tracee.go
@@ -240,6 +240,30 @@ func New(cfg Config) (*Tracee, error) {
 		}
 	}
 
+	// Clean up signatures not chosen for tracing
+	// Looping from the back to avoid bounds checking
+	for i := len(t.config.EngineConfig.Signatures) - 1; i > 0; i-- {
+		sig := t.config.EngineConfig.Signatures[i]
+		md, _ := sig.GetMetadata()
+		sigEventName := md.EventName
+		sigEventId, ok := events.Definitions.GetID(sigEventName)
+		if !ok {
+			logger.Debugw("removing signature from events to trace", "signature", sigEventName, "reason", "ID does not exist in definitions")
+			t.config.EngineConfig.Signatures = append(t.config.EngineConfig.Signatures[:i], t.config.EngineConfig.Signatures[i+1:]...)
+			continue
+		}
+		_, ok = t.events[sigEventId]
+		if !ok {
+			logger.Debugw("removing signature from events to trace", "signature", sigEventName, "reason", "was not requested for tracing")
+			t.config.EngineConfig.Signatures = append(t.config.EngineConfig.Signatures[:i], t.config.EngineConfig.Signatures[i+1:]...)
+		}
+	}
+
+	// Disable engine pipeline if no signatures are traced post cleanup
+	if len(t.config.EngineConfig.Signatures) == 0 {
+		t.config.EngineConfig.Enabled = false
+	}
+
 	// Handle all essential events dependencies
 
 	for id, evt := range t.events {


pkg/ebpf/tracee.go
```diff
@@ -240,6 +240,30 @@ func New(cfg Config) (*Tracee, error) {
 		}
 	}
 
+	// Clean up signatures not chosen for tracing
```
rafaeldtinoco, line 243: image

Styling adustments needed please. I believe @josedonizetti wanted to change this approach (not to load the engine from the cobra pkg already).

pkg/ebpf/tracee.go
```diff
@@ -240,6 +240,30 @@ func New(cfg Config) (*Tracee, error) {
...
+	// Looping from the back to avoid bounds checking
+	for i := len(t.config.EngineConfig.Signatures) - 1; i > 0; i-- {
+		sig := t.config.EngineConfig.Signatures[i]
```
yanivagman, line 246: You can use range instead, build the signatures list and assign to `t.config.EngineConfig.Signatures` at the end.

pkg/ebpf/tracee.go
```diff
@@ -240,6 +240,30 @@ func New(cfg Config) (*Tracee, error) {
...
+		md, _ := sig.GetMetadata()
+		sigEventName := md.EventName
+		sigEventId, ok := events.Definitions.GetID(sigEventName)
+		if !ok {
+			logger.Debugw("removing signature from events to trace", "signature", sigEventName, "reason", "ID does not exist in definitions")
+			t.config.EngineConfig.Signatures = append(t.config.EngineConfig.Signatures[:i], t.config.EngineConfig.Signatures[i+1:]...)
```
yanivagman, line 252: I don't think it's a good idea to remove signatures from the engine.
Soon we won't to be able to update the selected events in runtime, and removing them from here will require to reiterate the directories to find signatures and initilization.

NDStrahilevitz: Then this means we need two lists, one which we track on the side with all currently available signatures, which will allow us to load them later into the engine if they were unloaded. Then there is this list which i'm updating which is the initial signature list going on.
But that external list should probably be added alongside the API work on Loading and Unloading events.

pkg/ebpf/tracee.go
```diff
@@ -240,6 +240,30 @@ func New(cfg Config) (*Tracee, error) {
...
+		}
+		_, ok = t.events[sigEventId]
+		if !ok {
+			logger.Debugw("removing signature from events to trace", "signature", sigEventName, "reason", "was not requested for tracing")
+			t.config.EngineConfig.Signatures = append(t.config.EngineConfig.Signatures[:i], t.config.EngineConfig.Signatures[i+1:]...)
+		}
+	}
+
+	// Disable engine pipeline if no signatures are traced post cleanup
+	if len(t.config.EngineConfig.Signatures) == 0 {
```
NDStrahilevitz, line 263: This is probably not correct @josedonizetti.
When we will have Load/Unload event (in v0.16?) then we can possibly run 0 sigs initially then load one later. But if we disable the pipeline stage from the get go, that won't be possible (without adding code to reconfigure and modify the pipeline stages on the go).
So the correct thing is likely keeping the pipeline stage on, but adding a check for sig length before inputting the event to the engine, if no sigs are running we can optimize and skip it into the next one.

NDStrahilevitz:
@yanivagman Are we aiming to include event loading/unloading in this milestone? If not, we should add this until then. WDYT?

NDStrahilevitz:
@yanivagman @rafaeldtinoco @josedonizetti  Since dynamic loading of events isn't likely in this milestone, do we want to merge this PR or should I close it?
I think it's better just so we have less idle go routines running at least until we change the whole mechanism.

rafaeldtinoco:
> I think it's better just so we have less idle go routines running at least until we change the whole mechanism.

I'm in favor of not changing it since we will change the mechanism soon and it adds more code/complexity (don't think the idling is impacting as the scheduler is smart). If I'm wrong, then I'm happy to merge it.

NDStrahilevitz:
Closed as not planned

PR #3200, opened 2023-06-05
Bugfix: path_resolver: check multiple pids in mount ns

AlonZivony:

Don't return error upon error while accessing process root directory when resolving path to mount ns.
Instead, keep searching for accessible process root directory. This solves failure in accessing containers fs because the first process dies after switching to new mount ns.

### DIFF
--- a/pkg/containers/path_resolver.go
+++ b/pkg/containers/path_resolver.go
@@ -2,6 +2,7 @@ package containers
 
 import (
 	"fmt"
+	"github.com/aquasecurity/tracee/pkg/logger"
 	"io/fs"
 	"os"
 	"strings"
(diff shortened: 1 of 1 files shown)


josedonizetti (approved):

pkg/containers/path_resolver.go
```diff
@@ -56,7 +55,9 @@ func (cPathRes *ContainerPathResolver) GetHostAbsPath(mountNSAbsolutePath string
 		// fs.FS interface requires relative paths, so the '/' prefix should be trimmed.
 		entries, err := fs.ReadDir(cPathRes.fs, strings.TrimPrefix(procRootPath, "/"))
 		if err != nil {
-			return "", errfmt.WrapError(err)
+			// This process is either not alive or we don't have permissions to access.
+			// Try next pid in mount ns to find accessible path to mount ns files.
+			continue
```
yanivagman, line 60: Maybe log the error as well?

josedonizetti:
merged with the PR https://github.com/aquasecurity/tracee/pull/3205

PR #3209, opened 2023-06-07
ebpf: pipeline: reduce iteration over policies

geyslan:

5bce4837 **policies: replace "userspace" with "userland"** _(2023/jun/08) <Geyslan Gregório>_

```
```

0a015810 **docs: change "bitmask" mentions to "bitmap"** _(2023/jun/07) <Geyslan Gregório>_

```
```

229ce1c9 **ebpf: pipeline: reduce iteration over policies** _(2023/jun/07) <Geyslan Gregório>_

```
This creates a policy bitmap with filters that need to be checked in the
userland. This bitmap is used to avoid iterating the entire map when
there is no policy match in the kernel that requires userland filtering.

When there is a policy that requires userland filtering, the iteration
must be done using FilterableInUserSpaceMap(), which returns a possible
reduced map of policies.
```

### DIFF
--- a/pkg/ebpf/events_pipeline.go
+++ b/pkg/ebpf/events_pipeline.go
@@ -282,7 +282,7 @@ func (t *Tracee) decodeEvents(outerCtx context.Context, sourceChan chan []byte) 
 }
 
 // matchPolicies does the userland filtering (policy matching) for events. It iterates through all
-// existing policies, that were set by the kernel in the event bitmask. Some of those policies might
+// existing policies, that were set by the kernel in the event bitmap. Some of those policies might
 // not match the event after userland filters are applied. In those cases, the policy bit is cleared
 // (so the event is "filtered" for that policy).
 // This may be called in different stages of the pipeline (decode, derive, engine).
@@ -435,7 +435,7 @@ func (t *Tracee) processEvents(ctx context.Context, in <-chan *trace.Event) (
 				continue
 			}
 
-			// Get a bitmask with all policies containing container filters
+			// Get a bitmap with all policies containing container filters
 			policiesWithContainerFilter := t.config.Policies.ContainerFilterEnabled()
 
 			// Filter out events that don't have a container ID from all the policies that have
--- a/pkg/ebpf/events_pipeline.go
+++ b/pkg/ebpf/events_pipeline.go
@@ -290,7 +290,13 @@ func (t *Tracee) matchPolicies(event *trace.Event) uint64 {
 	eventID := events.ID(event.EventID)
 	bitmap := event.MatchedPoliciesKernel
 
-	for p := range t.config.Policies.Map() { // range through each existing policy
+	// Short circuit if there are no policies in userland that need filtering.
+	if bitmap&t.config.Policies.FilterableInUserSpace() == 0 {
+		event.MatchedPoliciesUser = bitmap // store untoched bitmap to be used in sink stage
+		return bitmap
+	}
+
+	for p := range t.config.Policies.FilterableInUserSpaceMap() { // range through each userland filterable policy
 		// Policy ID is the bit offset in the bitmap.
 		bitOffset := uint(p.ID)
 
@@ -369,7 +375,7 @@ func (t *Tracee) matchPolicies(event *trace.Event) uint64 {
 		}
 	}
 
-	event.MatchedPoliciesUser = bitmap // store filtered map to be used in sink stage
+	event.MatchedPoliciesUser = bitmap // store filtered bitmap to be used in sink stage
 
 	return bitmap
 }
--- a/pkg/ebpf/events_pipeline.go
+++ b/pkg/ebpf/events_pipeline.go
@@ -291,12 +291,12 @@ func (t *Tracee) matchPolicies(event *trace.Event) uint64 {
 	bitmap := event.MatchedPoliciesKernel
 
 	// Short circuit if there are no policies in userland that need filtering.
-	if bitmap&t.config.Policies.FilterableInUserSpace() == 0 {
+	if bitmap&t.config.Policies.FilterableInUserland() == 0 {
 		event.MatchedPoliciesUser = bitmap // store untoched bitmap to be used in sink stage
 		return bitmap
 	}
 
-	for p := range t.config.Policies.FilterableInUserSpaceMap() { // range through each userland filterable policy
+	for p := range t.config.Policies.FilterableInUserlandMap() { // range through each userland filterable policy
 		// Policy ID is the bit offset in the bitmap.
 		bitOffset := uint(p.ID)
 
--- a/pkg/policy/policies.go
+++ b/pkg/policy/policies.go
@@ -20,7 +20,8 @@ type Policies struct {
 	uidFilterableInUserSpace bool
 	pidFilterableInUserSpace bool
 
-	containerFiltersEnabled uint64
+	filterableInUserSpace   uint64 // bitmap of policies that must be filtered in user space
+	containerFiltersEnabled uint64 // bitmap of policies that have at least one container filter type enabled
 }
 
 func NewPolicies() *Policies {
@@ -34,6 +35,7 @@ func NewPolicies() *Policies {
 		pidFilterMax:               filters.MaxNotSetUInt,
 		uidFilterableInUserSpace:   false,
 		pidFilterableInUserSpace:   false,
+		filterableInUserSpace:      0,
 		containerFiltersEnabled:    0,
 	}
 }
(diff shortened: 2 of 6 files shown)


rafaeldtinoco:
This:

c715a276 **policies: Replace "userspace" with "userland"** _(2023/jun/08) <Geyslan Gregório>_

```
This commit updates the codebase in events_pipeline.go and policies.go
to replace occurrences of "userspace" with "userland" in the context of
policies filtering.

By using the term "userland", the code becomes more accessible to
developers familiar with standard operating system concepts and promotes
better comprehension of the code's functionality related to policies
filtering in the user-level environment.
```

is WAYYY to descriptive for such a simple change. Git logs should be extensive if mandatory for the understanding of what patch is doing. Otherwise they should be oneliner.

rafaeldtinoco:
Third PR git log seems right for me. I'll check it out locally once Im done with some of my coding, and give a +1 soon. Cheers!

geyslan:
> no need for more reasoning (the issue can give more in-depth if ever needed) IMO.

We don't have an issue for this, as it's a chore and it has been mentioned in the past that chores shouldn't have open issues.

So, without more information to reason the changes, I believe that well-written comments when changing the naming in code are valuable for a few reasons. First, they act as documentation and help future contributors understand the purpose and reasoning behind the name change. They also provide context, reducing confusion and making the changes easier to understand. Well-written comments reduce cognitive load by providing additional context, making it easier for developers to work with the code in the future. Lastly, reasoned comments prevent further discussions such as "why name it this way" or "I think it's better to name it that way." Do you agree?

rafaeldtinoco:
> Lastly, reasoned comments prevent further discussions such as "why name it this way" or "I think it's better to name it that way." Do you agree?

Nope, I think this is bike shedding (both, the git log and your answer). Its a simple change, it should be stated in a simple oneliner git log.

pkg/ebpf/events_pipeline.go
```diff
@@ -282,15 +282,21 @@ func (t *Tracee) decodeEvents(outerCtx context.Context, sourceChan chan []byte)
...
 func (t *Tracee) matchPolicies(event *trace.Event) uint64 {
 	eventID := events.ID(event.EventID)
 	bitmap := event.MatchedPoliciesKernel
 
-	for p := range t.config.Policies.Map() { // range through each existing policy
+	// Short circuit if there are no policies in userland that need filtering.
+	if bitmap&t.config.Policies.FilterableInUserland() == 0 {
+		event.MatchedPoliciesUser = bitmap // store untoched bitmap to be used in sink stage
+		return bitmap
+	}
```
rafaeldtinoco, line 297: You can place this logic in FilterableInUserlandMap() and have it iterate 0 items (instead of doing extra logic outside the object).

rafaeldtinoco (dismissed):
make filterableinuserlandmap iterate 0 items (and check bitmap to be faster when returning) instead of doing things outside the class.

pkg/policy/policies.go
```diff
@@ -58,43 +60,58 @@ func (ps *Policies) PIDFilterMax() uint64 {
...
 	return atomic.LoadUint64(&ps.containerFiltersEnabled)
 }
 
-// Compute recalculates values, updates flags and fills the reduced user space
-// map. It must be called at initialization and at every runtime policies changes
+// FilterableInUserland returns a bitmap of policies that must be filtered in userland
+// (ArgFilter, RetFilter, ContextFilter, UIDFilter and PIDFilter).
+//
+// TODO: make sure the stores are also atomic (an atomic load is only protecting
+// the read from context switches, not from CPU cache coherency issues).
```
rafaeldtinoco, line 84: Not that simple (about atomics). If you did something like this:

```
diff --git a/pkg/policy/policies.go b/pkg/policy/policies.go
index 3b363a7d..0f72a4d3 100644
--- a/pkg/policy/policies.go
+++ b/pkg/policy/policies.go
@@ -13,30 +13,49 @@ type Policies struct {
        filterEnabledPoliciesMap  map[*Policy]int      // stores only enabled policies
        filterUserlandPoliciesMap map[*Policy]int      // stores a reduced map with only userland filterable policies
 
-       uidFilterMin            uint64
-       uidFilterMax            uint64
-       pidFilterMin            uint64
-       pidFilterMax            uint64
-       uidFilterableInUserland bool
-       pidFilterableInUserland bool
-
-       filterableInUserland    uint64 // bitmap of policies that must be filtered in userland
-       containerFiltersEnabled uint64 // bitmap of policies that have at least one container filter type enabled
+       uidFilterMin *atomic.Uint64
+       uidFilterMax *atomic.Uint64
+       pidFilterMin *atomic.Uint64
+       pidFilterMax *atomic.Uint64
+
+       uidFilterableInUserland *atomic.Bool
+       pidFilterableInUserland *atomic.Bool
+
+       filterableInUserland    *atomic.Uint64 // bitmap of policies that must be filtered in userland
+       containerFiltersEnabled *atomic.Uint64 // bitmap of policies that have at least one container filter type enabled
 }
 
 func NewPolicies() *Policies {
+       uidFilterMin := &atomic.Uint64{}
+       uidFilterMax := &atomic.Uint64{}
+       pidFilterMin := &atomic.Uint64{}
+       pidFilterMax := &atomic.Uint64{}
+       uidFilterableInUserland := &atomic.Bool{}
+       pidFilterableInUserland := &atomic.Bool{}
+       FilterableInUserland := &atomic.Uint64{}
+       containerFiltersEnabled := &atomic.Uint64{}
+
+       uidFilterMin.Store(filters.MinNotSetUInt)
+       uidFilterMax.Store(filters.MaxNotSetUInt)
+       pidFilterMin.Store(filters.MinNotSetUInt)
+       pidFilterMax.Store(filters.MaxNotSetUInt)
+       uidFilterableInUserland.Store(false)
+       pidFilterableInUserland.Store(false)
+       FilterableInUserland.Store(0)
+       containerFiltersEnabled.Store()
+
        return &Policies{
                policiesArray:             [MaxPolicies]*Policy{},
                filterEnabledPoliciesMap:  map[*Policy]int{},
                filterUserlandPoliciesMap: map[*Policy]int{},
-               uidFilterMin:              filters.MinNotSetUInt,
-               uidFilterMax:              filters.MaxNotSetUInt,
-               pidFilterMin:              filters.MinNotSetUInt,
-               pidFilterMax:              filters.MaxNotSetUInt,
-               uidFilterableInUserland:   false,
-               pidFilterableInUserland:   false,
-               filterableInUserland:      0,
-               containerFiltersEnabled:   0,
+               uidFilterMin:              uidFilterMin,
+               uidFilterMax:              uidFilterMax,
+               pidFilterMin:              pidFilterMin,
+               pidFilterMax:              pidFilterMax,
+               uidFilterableInUserland:   uidFilterableInUserland,
+               pidFilterableInUserland:   pidFilterableInUserland,
+               filterableInUserland:      FilterableInUserland,
+               containerFiltersEnabled:   containerFiltersEnabled,
        }
 }
 
@@ -45,45 +64,39 @@ func (ps *Policies) Count() int {
 }
 
 func (ps *Policies) UIDFilterMin() uint64 {
-       return ps.uidFilterMin
+       return ps.uidFilterMin.Load()
 }
 
 func (ps *Policies) UIDFilterMax() uint64 {
-       return ps.uidFilterMax
+       return ps.uidFilterMax.Load()
 }
 
 func (ps *Policies) PIDFilterMin() uint64 {
-       return ps.pidFilterMin
+       return ps.pidFilterMin.Load()
 }
 
 func (ps *Policies) PIDFilterMax() uint64 {
-       return ps.pidFilterMax
+       return ps.pidFilterMax.Load()
 }
 
 func (ps *Policies) UIDFilterableInUserland() bool {
-       return ps.uidFilterableInUserland
+       return ps.uidFilterableInUserland.Load()
 }
 
 func (ps *Policies) PIDFilterableInUserland() bool {
-       return ps.pidFilterableInUserland
+       return ps.pidFilterableInUserland.Load()
 }
 
-// ContainerFilterEnabled returns a bitmap of policies that have at least
-// one container filter type enabled.
-//
-// TODO: make sure the stores are also atomic (an atomic load is only protecting
-// the read from context switches, not from CPU cache coherency issues).
+// ContainerFilterEnabled returns a bitmap of policies that have at least one container filter type
+// enabled.
 func (ps *Policies) ContainerFilterEnabled() uint64 {
-       return atomic.LoadUint64(&ps.containerFiltersEnabled)
+       return ps.containerFiltersEnabled.Load()
 }
 
-// FilterableInUserland returns a bitmap of policies that must be filtered in userland
-// (ArgFilter, RetFilter, ContextFilter, UIDFilter and PIDFilter).
-//
-// TODO: make sure the stores are also atomic (an atomic load is only protecting
-// the read from context switches, not from CPU cache coherency issues).
+// FilterableInUserland returns a bitmap of policies that must be filtered in userland (ArgFilter,
+// RetFilter, ContextFilter, UIDFilter and PIDFilter).
 func (ps *Policies) FilterableInUserland() uint64 {
-       return atomic.LoadUint64(&ps.filterableInUserland)
+       return ps.filterableInUserland.Load()
 }
 
 // Compute recalculates values, updates flags, fills the reduced userland map,
@@ -98,7 +111,9 @@ func (ps *Policies) Compute() {
        ps.updateContainerFilterEnabled()
 
        userlandMap := make(map[*Policy]int)
-       ps.filterableInUserland = 0
+
+       ps.filterableInUserland.Store(0)
+
        for p := range ps.filterEnabledPoliciesMap {
                if p.ArgFilter.Enabled() ||
                        p.RetFilter.Enabled() ||
```

And encapsulated the Load/Stores for the uint64s, you would still have a problem requiring a mutex protecting those uint64s (and not only atomics). So atomics might not even be needed for them (if they're protected by a mutex).

That happens because you use "utils.SetBit()" in that variable, and the SetBit does:

- load
- shift
- store

Which are 3 operations that can't be protected by a single underlaying HW atomic operation. So you need a mutex (with an atomic primitive under the hood) to protect the load/change/store (as you can't do a "compare and exchange" atomic HW operation for bitwise).

rafaeldtinoco: Its better if you add a "big lock" (single mutex) and hold that mutex for all operations on your uint64s for now (it will be safer and less error prone). For bools you can use atomics with SetTrue() SetFalse() methods, for uint64s that are simply Load/Store you can also have Set/Get methods, but for uint64s with bits being shifted, you will need a lock if no underlaying atomic instruction/method/operation exists (which I believe is the case here).

rafaeldtinoco (approved):
Not from this PR, the entire interface will need sync work at https://github.com/aquasecurity/tracee/issues/3239. Thanks for opening that @geyslan! LGTM this one

josedonizetti (approved):

PR #3210, opened 2023-06-07
docs: fix tag creation example relasing.md

josedonizetti:

Fix typo on tag creation

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -29,7 +29,7 @@
    [`.github/workflows/release.yaml`] workflow.
 
    ```console
-   git tag -v0.99.0
+   git tag v0.99.0
    git push upstream v0.99.0
    ```
 


rafaeldtinoco:
Despite this fix, which is right, I believe this entire RELEASING file should be re-worked in this patch. Also, the git log title has a typo.

josedonizetti:
> Despite this fix, which is right, I believe this entire RELEASING file should be re-worked in this patch. Also, the git log title has a typo.

sure, what else do you want to change on releasing.md?

rafaeldtinoco:
> @rafaeldtinoco sure, what else do you want to change on releasing.md?

Sorry, I should have mentioned it before. I would say some things are out of order mostly (or slightly inaccurate). I would have a single list with the exact ordering of things to be done: 

1. check libbpfgo and libbpf
2. update types to latest
3. update container tags (should be automated in the release process)
4. run e2e tests
5. clean clone of the repository for tag bump
6. tag bump
7. check generated images for both arches
8. publish helm chart

WDYT?

rafaeldtinoco:
Waiting feedback about the new release model so we can close this (or not), after Jose's update.

PR #3222, opened 2023-06-11
flags: refactor FilterMap

yanivagman:

Currently the filterFlag struct includes information about the policy - it's index and name. Since FilterMap is already indexed by policy, this information is redundant and can be set once per policy. Removing the policy information from filterFlag makes code simpler and easier to modify in future PR (e.g. splitting filter flag to event and scope flags).


josedonizetti (approved):
LGTM

Issue #3031, opened 2023-05-02
Extract config package from cmd and tracee packages


geyslan:
As discussed in https://github.com/aquasecurity/tracee/pull/3000#issuecomment-1527872736, we have decided to wait urfave to be dropped with tracee-ebpf and tracee-rule to tackle this.

Earlier discussion was done on extracting logic from cmd and tracee cli and can be used as a reference:

https://github.com/aquasecurity/tracee/issues/960:

> configurations - create package config that includes filter, captureConfig ,outputConfig and related functions. (also create an interface for config and implement validators) - https://github.com/aquasecurity/tracee/issues/1264

> Do we want "filters" and "containers" packages to be exposed to external consumers? Or is it just the configuration of those (for which I agree we should have a config package)

> The config package will implement the methods associated to the Config struct, such as func (tc Config) Validate() error - this reference should give an idea as well https://github.com/mtcherni95/tracee/blob/tracee-library/tracee-ebpf/engine/config/config.go

PR #3228, opened 2023-06-12
config: extract config structs to its own pkg

geyslan:

Fix: #3031

1f91b802 **config: move pcaps Config struct to pkg/config** _<sub>(2023/jun/13) Geyslan Gregório \<geyslan@gmail.com\></sub>_

```
It names the struct `PcapsConfig` and moves it to pkg/config.
```

ec36fe95 **flags: rename OutputConfig to PrepareOutputResult** _<sub>(2023/jun/13) Geyslan Gregório \<geyslan@gmail.com\></sub>_

```
```

0cb13954 **config: extract config structs to its own pkg** _<sub>(2023/jun/13) Geyslan Gregório \<geyslan@gmail.com\></sub>_

```
This extracts the config file from the ebpf package to its own.
It also moves into the config package PrinterConfig and
flags/OutputConfig renaming the latter to PrinterOutputConfig.
```


pkg/config/config.go
```diff
@@ -92,53 +134,48 @@ const (
...
+
+type PrinterConfig struct {
+	Kind          string
+	OutPath       string
+	OutFile       io.WriteCloser
+	ContainerMode ContainerMode
+	RelativeTS    bool
+}
+
+type PrinterOutputConfig struct {
```
josedonizetti, line 178: This is the only one I'm not sure. Because it is not actually a config, it was created just to simplify the return from `PrepareOutput` to return two values only, instead of three which makes the code look very ugly. Maybe it should be a structure there, for it only still? `type PrepareOutputResult struct{}`? WDYT?

```suggestion
type PrinterOutputConfig struct {
```

geyslan: > This is the only one I'm not sure. Because it is not actually a config, it was created just to simplify the return from `PrepareOutput` to return two values only, instead of three which makes the code look very ugly.

Oh I understand. In this case I'll leave it as it is.

> Maybe it should be a structure there, for it only still? `type PrepareOutputResult struct{}`? WDYT?

Can it be internal to the pkg? prepareOutputResult?

josedonizetti: @geyslan yup, it can be a internal one

geyslan: Actually it can't since it's used from cobra and urfave, so I've just renamed as you suggested. 👍🏼

josedonizetti (commented):
@geyslan why didn't we also bring the [engine.Config](https://github.com/aquasecurity/tracee/blob/main/pkg/signatures/engine/engine.go#L20), since all configs will be under `pkg/config` why not the engine as well?

geyslan:
> @geyslan why didn't we also bring the [engine.Config](https://github.com/aquasecurity/tracee/blob/main/pkg/signatures/engine/engine.go#L20), since all configs will be under `pkg/config` why not the engine as well?

I think it makes sense. Besides engine we have also pcaps.

geyslan:
Bringing the engine configuration into pkg/config made the tracee-rules now dependent on pkg/config. So this build change is required:

https://github.com/aquasecurity/tracee/pull/3228/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R498-R503

rafaeldtinoco (approved):
I agree on not making tracee-rules depending on libbpf just now (add a TODO for when we have the single binary if some other change is needed). LGTM

josedonizetti (approved):

Issue #3230, opened 2023-06-13
tracee-rules panic


OriGlassman:
Command: ./tracee-rules --allcaps

Error (hard to reproduce):
anic: sync: WaitGroup is reused before previous Wait has returned

 

goroutine 1 [running]:
sync.(*WaitGroup).Wait(0xc0001fc780?)
    /usr/local/go/src/sync/waitgroup.go:118 +0x7f
github.com/aquasecurity/tracee/pkg/signatures/engine.(*Engine).checkCompletion(0xc0001fc780)
    /pkg/signatures/engine/engine.go:137 +0x3f
github.com/aquasecurity/tracee/pkg/signatures/engine.(*Engine).consumeSources(0xc0001fc780, {0x1793000, 0xc0002fecc0})
    /pkg/signatures/engine/engine.go:215 +0x34a
github.com/aquasecurity/tracee/pkg/signatures/engine.(*Engine).Start(0xc0001fc780, {0x1793000, 0xc0002fecc0})
    /pkg/signatures/engine/engine.go:112 +0x1e7
main.main.func1(0xc0002fe640)
   /cmd/tracee-rules/main.go:166 +0xb17
github.com/urfave/cli/v2.(*App).RunContext(0xc0002971e0, {0x1792270?, 0xc0001280e8}, {0xc000124000, 0x7, 0x7})
    /home/ubuntu/go/pkg/mod/github.com/urfave/cli/v2@v2.3.0/app.go:322 +0x97c
github.com/urfave/cli/v2.(*App).Run(...)
    /home/ubuntu/go/pkg/mod/github.com/urfave/cli/v2@v2.3.0/app.go:224
main.main()
    /cmd/tracee-rules/main.go:257 +0x97a

(tracee version 0.14.2)

josedonizetti:
This is the same bug we have at -> https://github.com/aquasecurity/tracee/issues/3151

I'll be fixing it today.

Issue #3151, opened 2023-05-27
races when loading/unloading tracee signatures (ctrl+c during initialization).


rafaeldtinoco:
## Description

```
$ sudo ./dist/tracee -f e=security_file_open
^CTIME             UID    COMM             PID     TID     RET              EVENT                     ARGS
panic: sync: WaitGroup is reused before previous Wait has returned

goroutine 26 [running]:
sync.(*WaitGroup).Wait(0xc000030fc0?)
        /usr/local/go/src/sync/waitgroup.go:141 +0x85
github.com/aquasecurity/tracee/pkg/signatures/engine.(*Engine).checkCompletion(0xc000030fc0)
        /home/ubuntu/tracee-rafael/pkg/signatures/engine/engine.go:137 +0x3f
github.com/aquasecurity/tracee/pkg/signatures/engine.(*Engine).consumeSources(0xc000030fc0, {0x2b92448, 0xc0001fc4c0})
        /home/ubuntu/tracee-rafael/pkg/signatures/engine/engine.go:215 +0x3ea
github.com/aquasecurity/tracee/pkg/signatures/engine.(*Engine).Start(0xc000030fc0, {0x2b92448, 0xc0001fc4c0})
        /home/ubuntu/tracee-rafael/pkg/signatures/engine/engine.go:112 +0x1e7
created by github.com/aquasecurity/tracee/pkg/ebpf.(*Tracee).engineEvents
        /home/ubuntu/tracee-rafael/pkg/ebpf/signature_engine.go:39 +0x473
```

## Output of `tracee version`:

Pre v0.15 (freeze date)

## Output of `uname -a`:

```
v6.3.1
```

geyslan:
@rafaeldtinoco is this still relevant? I still haven't been able to reproduce.

rafaeldtinoco:
This wait group is the signatures engine wait group and it is used do synchronize loading/unloading the signatures. I believe the ctrl+c made both logic to run concurrently and cause the issue.

I have not seen any recent changes in that code path, so its relevant I suppose (although not "high priority", but, still...).

rafaeldtinoco:
Setting API server label because this will be sorted out when we go the dynamic loading/unloading of signatures.

geyslan:
I've also assigned @josedonizetti since it can be connected to https://github.com/aquasecurity/tracee/pull/3188 solving.

josedonizetti:
Looking into it today!

PR #3233, opened 2023-06-13
engine: fix panic on waitgroup

josedonizetti:

[Wait must be synchronized with the first Add.](https://go.dev/src/sync/waitgroup.go?s=1857:1892#L42) 

This seems to be the case here, as we call Add from inside the goroutine, for a short lived `tracee-rules` process there might  be a race between `checkCompletion` and `signatureStart` triggering the error reported on the issues https://github.com/aquasecurity/tracee/issues/3230 and https://github.com/aquasecurity/tracee/issues/3151

### DIFF
--- a/pkg/signatures/engine/engine.go
+++ b/pkg/signatures/engine/engine.go
@@ -88,7 +88,6 @@ func NewEngine(config Config, sources EventSources, output chan detect.Finding) 
 
 // signatureStart is the signature handling business logics.
 func signatureStart(signature detect.Signature, c chan protocol.Event, wg *sync.WaitGroup) {
-	wg.Add(1)
 	for e := range c {
 		if err := signature.OnEvent(e); err != nil {
 			meta, _ := signature.GetMetadata()
@@ -106,6 +105,7 @@ func (engine *Engine) Start(ctx context.Context) {
 	defer engine.unloadAllSignatures()
 	engine.signaturesMutex.RLock()
 	for s, c := range engine.signatures {
+		engine.waitGroup.Add(1)
 		go signatureStart(s, c, &engine.waitGroup)
 	}
 	engine.signaturesMutex.RUnlock()
@@ -242,6 +242,7 @@ func (engine *Engine) dispatchEvent(s detect.Signature, event protocol.Event) {
 	engine.signatures[s] <- event
 }
 
+// TODO: This method seems not to be used, let's confirm inside the team and remove it if not needed
 // LoadSignature will call the internal signature loading logic and activate its handling business logics.
 // It will return the signature ID as well as error.
 func (engine *Engine) LoadSignature(signature detect.Signature) (string, error) {
@@ -250,6 +251,7 @@ func (engine *Engine) LoadSignature(signature detect.Signature) (string, error) 
 		return id, err
 	}
 	engine.signaturesMutex.RLock()
+	engine.waitGroup.Add(1)
 	go signatureStart(signature, engine.signatures[signature], &engine.waitGroup)
 	engine.signaturesMutex.RUnlock()
 


geyslan (commented):
Makes sense. LGTM.

yanivagman:
Didn't we want to merge this before the v0.15.1 release?

OriGlassman:
Are you guys sure this can only happen by sending a SIGINT? This error was observed on a lambda (where it shouldn't have received a SIGINT)

josedonizetti:
@OriGlassman It is not because of SIGINT, but a short-lived `tracee-rules`, which might happen due to different reasons, a signal as one of the issues, or if you are using it to analyze events from a file, maybe the file doesn't have enough events. 

The race seems to happen because we close the [`inputs.Tracee`](https://github.com/aquasecurity/tracee/blob/main/cmd/tracee-rules/input.go#L129) channel to quickly, due to the lack of events. Triggering the stop of the engine [`consumeSources`](https://github.com/aquasecurity/tracee/blob/main/pkg/signatures/engine/engine.go#L197) loop, which starts a [`Wait`](https://github.com/aquasecurity/tracee/blob/main/pkg/signatures/engine/engine.go#L137) op on the waitgroup, but the go routines for the signatures still didn't had the time to setup the [waitgroup](https://github.com/aquasecurity/tracee/blob/main/pkg/signatures/engine/engine.go#L137).


A simple way to simulate it is by creating a file with just one event:

test.json
```
{"timestamp":1686775048087416430,"threadStartTime":1686754169760481353,"processorId":7,"processId":1,"cgroupId":10999,"threadId":12,"parentProcessId":3293,"hostProcessId":3374,"hostThreadId":3480,"hostParentProcessId":3293,"userId":65534,"mountNamespace":4026533123,"pidNamespace":4026533127,"processName":"prometheus","hostName":"josedonizetti-x","containerId":"632b2b487644f2627ffd0fc148b0c83b1efd43a8a2236975071fb85ae932d922","container":{"id":"632b2b487644f2627ffd0fc148b0c83b1efd43a8a2236975071fb85ae932d922"},"kubernetes":{},"eventId":"2010","eventName":"net_packet_http_request","matchedPolicies":[""],"argsNum":2,"returnValue":0,"syscall":"write","stackAddresses":[0],"contextFlags":{"containerStarted":true,"isCompat":false},"args":[{"name":"metadata","type":"trace.PktMeta","value":{"src_ip":"127.0.0.1","dst_ip":"127.0.0.1","src_port":46122,"dst_port":9100,"protocol":6,"packet_len":336,"iface":"any"}},{"name":"http_request","type":"trace.ProtoHTTPRequest","value":{"method":"GET","protocol":"HTTP/1.1","host":"localhost:9100","uri_path":"/metrics","headers":{"Accept":["application/openmetrics-text;version=1.0.0,application/openmetrics-text;version=0.0.1;q=0.75,text/plain;version=0.0.4;q=0.5,*/*;q=0.1"],"Accept-Encoding":["gzip"],"User-Agent":["Prometheus/2.44.0"],"X-Prometheus-Scrape-Timeout-Seconds":["5"]},"content_length":0}}]}
```

Then, starting `tracee-rules` to analyze it in a loop:

```
while true; do sudo ./dist/tracee-rules --allcaps --input-tracee format:json --input-tracee file:./test.log 2>&1 | grep -v info; done
```

It should hit the issue every X time that we run `tracee-rules`. 

After several tests, this PR indeed fixes the bug.

josedonizetti:
@yanivagman @rafaeldtinoco @geyslan as per the comment above ˆˆ, the bug is indeed fixed by this change.

geyslan (approved):

rafaeldtinoco (approved):
LGTM

Issue #3181, opened 2023-05-31
policies: rename  scope, filter and action to use plural as they are arrays


josedonizetti:
```
name: sample
description: sample
scopes: 
  - global
rules:
  - event: name
     filters:
        - retval=0
        - args.pathname!=/tmp
     actions:
        - webhook
        - log
```

PR #3242, opened 2023-06-14
policies: rename list fields to be plural

josedonizetti:

Fix https://github.com/aquasecurity/tracee/issues/3241
Fix https://github.com/aquasecurity/tracee/issues/3181

### 2. Explain how to test it

This PR does two things:

1 - renames a few fields of the policy, and fix the code where appropriate.
     filter -> filters
     action -> actions

2 - add support to multiple default actions
      defaultAction -> defaultActions

### 3. Other comments

This PR will break the internal tests, because of the policies change.

### DIFF
--- a/deploy/helm/tracee/templates/tracee-policies.yaml
+++ b/deploy/helm/tracee/templates/tracee-policies.yaml
@@ -9,7 +9,8 @@ data:
   signatures.yaml: |-
     name: signature_events
     description: traces all signature events
-    defaultAction: log
+    defaultActions:
+      - log
     scope:
       - global
     rules:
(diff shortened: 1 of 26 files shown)


deploy/helm/tracee/templates/tracee-policies.yaml
```diff
@@ -9,8 +9,9 @@ data:
   signatures.yaml: |-
     name: signature_events
     description: traces all signature events
-    defaultAction: log
-    scope:
+    defaultActions: 
+      - log
+    scopes:
```
yanivagman, line 14: This should be kept as scope and not scopes (we have one scope, which is defined by several filters)

josedonizetti: ok

yanivagman:
Still a draft?

rafaeldtinoco (approved):
LGTM

PR #3243, opened 2023-06-14
Update packaging.md

pimvh:
Hi!  I've tried building a package on Ubuntu Jammy Jellyfish with the command specified in this docs (with `sudo`, as my user was not in the docker group). I ran into the following issue however (nearing the end of the build):

```
....
....
....
Successfully tagged ubuntu-pkgs-make:jammy
docker \
        run --rm --pid=host --network=host --privileged -v /etc/os-release:/etc/os-release-host:ro -v /home/superpim/tracee:/tracee/tracee -v /lib/modules:/lib/modules:ro -v /usr/src:/usr/src:ro -v /sys/kernel/security:/sys/kernel/security:ro -e LIBBPFGO_OSRELEASE_FILE=/etc/os-release-host \
        -it ubuntu-pkgs-make:jammy \
        packaging/ubuntu-pkgs.sh bin jammy


fatal: detected dubious ownership in repository at '/tracee/tracee'
To add an exception for this directory, call:

        git config --global --add safe.directory /tracee/tracee
dpkg-buildpackage: warning:     debian/changelog(l1): version '~22.04~2306141456-' is invalid: revision cannot be empty
LINE: tracee (~22.04~2306141456-) jammy; urgency=medium
dpkg-buildpackage: info: source package tracee
dpkg-buildpackage: info: source version unknown
dpkg-buildpackage: error: version number does not start with digit
mv: cannot stat '../*.deb': No such file or directory
rm .ubuntu-pkgs-prepare-jammy .check_docker

```
Seems to be a permission issue.

I think adding this or something similar to the docs could be beneficial for the read-and-run documentation readers.

It adds a little bit to the Ubuntu installation docs, to make this issue avoidable for other people.

### 2. Explain how to test it

You can run the commands to build a package on a Ubuntu machine, on which your user is not part of the docker group (you will have to `sudo make ...`, and if you did not clone the repo as root, you will get this issue.

### DIFF
--- a/docs/contributing/building/packaging.md
+++ b/docs/contributing/building/packaging.md
@@ -10,6 +10,12 @@ You may use `builder/Makefile.packaging` to generate Ubuntu deb packages. It
 will use docker containers to generate appropriate packages, so you don't need
 to install build dependencies in your OS.
 
+Before running these commands make sure your user is added to the docker group:
+
+```
+sudo usermod -aG docker $USER
+```
+
 !!! Ubuntu Note
 
     !!! Building Tip


docs/contributing/building/packaging.md
```diff
@@ -10,6 +10,12 @@ You may use `builder/Makefile.packaging` to generate Ubuntu deb packages. It
 will use docker containers to generate appropriate packages, so you don't need
 to install build dependencies in your OS.
 
+Before running these commands make sure your user is added to the docker group:
+
+```
+sudo usermod -aG docker $USER
+```
+
```
rafaeldtinoco, line 18: I assume this instruction is also true for the Fedora packages, no ? We should add it below the introduction then. I think its a good info.

pimvh: Yes, I think so. I've added the same snippet to the Fedora intro.

rafaeldtinoco:
Thanks for the contribution.

Issue #2778, opened 2023-02-27
tracee can't load signature produced binary artifact tar.gz


josedonizetti:
## Description

```
fatal error: runtime: no plugin module data

goroutine 1 [running]:
runtime.throw({0x183394e?, 0x46ec1b?})
        /usr/local/go/src/runtime/panic.go:1047 +0x5d fp=0xc0008a6ea0 sp=0xc0008a6e70 pc=0x43e07d
plugin.lastmoduleinit()
        /usr/local/go/src/runtime/plugin.go:20 +0x8ce fp=0xc0008a6f88 sp=0xc0008a6ea0 pc=0x46ea6e
plugin.open({0xc000170340, 0x33})
        /usr/local/go/src/plugin/plugin_dlopen.go:77 +0x3ef fp=0xc0008a71f8 sp=0xc0008a6f88 pc=0xbbaa0f
plugin.Open(...)
        /usr/local/go/src/plugin/plugin.go:32
github.com/aquasecurity/tracee/pkg/rules/signature.findGoSigs.func1.1({0xc000170340, 0x33}, {0x216aba8, 0xc00021eb00}, {0x0?, 0x0?})
        /tracee/pkg/rules/signature/signature.go:78 +0x1c9 fp=0xc0008a72d8 sp=0xc0008a71f8 pc=0x132c2e9
path/filepath.walkDir({0xc000170340, 0x33}, {0x216aba8, 0xc00021eb00}, 0xc0008a74d0)
        /usr/local/go/src/path/filepath/path.go:398 +0x5c fp=0xc0008a7398 sp=0xc0008a72d8 pc=0x581f5c
path/filepath.walkDir({0xc00016a750, 0x28}, {0x216ac18, 0xc0005f6570}, 0xc0008a74d0)
        /usr/local/go/src/path/filepath/path.go:420 +0x2aa fp=0xc0008a7458 sp=0xc0008a7398 pc=0x5821aa
path/filepath.WalkDir({0xc00016a750, 0x28}, 0xc0008a74d0)
        /usr/local/go/src/path/filepath/path.go:484 +0xb0 fp=0xc0008a74b8 sp=0xc0008a7458 pc=0x582690
github.com/aquasecurity/tracee/pkg/rules/signature.findGoSigs.func1()
        /tracee/pkg/rules/signature/signature.go:67 +0x45 fp=0xc0008a74f0 sp=0xc0008a74b8 pc=0x132c105
[...]

PR #3244, opened 2023-06-14
tracee: skip golang plugin for static binaries

josedonizetti:

Fix https://github.com/aquasecurity/tracee/issues/2778

If the binary is static, we will skip the golang signatures in order to avoid panics, so tracee-static can still be used for builtin events and rego signatures.

### 2. Explain how to test it

 on main:
```
make clean && make
rm dist/tracee
BTFHUB=0 STATIC=1 make tracee

sudo ./dist/tracee # should trigger the error
```

on this PR, the same test print it is skipping the golang plugin.

### DIFF
--- a/pkg/signatures/signature/signature.go
+++ b/pkg/signatures/signature/signature.go
@@ -2,6 +2,7 @@ package signature
 
 import (
 	"bytes"
+	"debug/elf"
 	"fmt"
 	"io/fs"
 	"os"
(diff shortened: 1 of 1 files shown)


josedonizetti:
Thanks for the tip on checking the symbols @rafaeldtinoco :D

rafaeldtinoco (approved):
LGTM.

I'll ask for @yanivagman input on this one as it was a direct request from him (I'm in doubt about the Warn versus Debug, it's just a question of usability).

Issue #3251, opened 2023-06-18
containerId and containerStarted flag does not match


yanivagman:
## Description

It was reported that an event had and empty containerId while containerStarted flag was set to true.
This should be, of course, impossible since it can't be that we know about a container that was started without knowing its ID.
Since this is an edge case (a possible race condition that may explain this is described next), **a possible solution is to set containerStarted to false in the decode stage whenever we get an empty container id**.

Following is a possible explanation of the observed behavior:

The containerStarted flag is being set in the eBPF code. In order for it to be true, three things should happen:

1. The cgroup state (in the BPF map) was first CONTAINER_CREATED (meaning the cgroup was created after tracee started)
2. Then a process was executed in this cgroup (while CONTAINER_CREATED state is set in the bpf map)
3. This executed process has a mount namespace which is different than its parent mount namespace
 
Now the question is - how do we have an empty container id while containerStarted field is set to true?

To explain that, I will quickly describe how container id is being extracted:

1. Cgroup id is being extracted from the kernel as part of the event
2. Cgroup id is mapped in userspace to container id (if we already have this mapping)
3. If the mapping is not cached, it will perform a lookup in cgroupfs to find which cgroup path has a matching cgroup id and extract the container id from the cgroup path

Few points we can make from the above:

1. missing (or missing data in) cgroup_mkdir event is irrelevant since then tracee will simply perform a lookup in cgroup fs and fill the missing mapping.
2. If the process that triggered the event is indeed on the host, it is a "special" process since it has a cgroup which is different from the host cgroup and created in a new mount namespace.
3. If the process that triggered the event is actually in a container, then it is possible that the cgroup directory was deleted before performing the mapping between cgroup id to container id, but it is very unlikely that this happened AND that we didn't get the cgroup_mkdir event, so I don't think that's the case
4. Assuming the process is not in a container and the cgroup directory exist, tracee should have deleted the cgroup entry from the bpf map (happens once, usually at cgroup_mkdir event). It is possible that the cgroup was created and immediately a process in that cgroup executed before we had the chance to delete the cgroup entry from the bpf map (thus we get an empty container id with containerStarted flag set).

## Output of `tracee version`:

```
v0.15.0
```

## Output of `uname -a`:

```
(paste your output here)
```

## Additional details 

Event example:

     {
        "timestamp": 1685741186280775000,
        "threadStartTime": 6789227257225423,
        "processorId": 0,
        "processId": 17485,
        "cgroupId": 467826,
[...]

PR #3253, opened 2023-06-19
fix: fix container edge case in events pipeline

geyslan:

Fix: #3251

045f08b8 **fix: fix container edge case in events pipeline** _<sub>(2023/jun/19) Geyslan Gregório \<geyslan@gmail.com\></sub>_

```
See #3251 for more details.
```


pkg/ebpf/events_pipeline.go
```diff
@@ -380,15 +380,19 @@ func (t *Tracee) matchPolicies(event *trace.Event) uint64 {
...
 }
 
-func parseContextFlags(flags uint32) trace.ContextFlags {
+func parseContextFlags(containerId string, flags uint32) trace.ContextFlags {
 	const (
 		contStartFlag = 1 << iota
 		IsCompatFlag
 	)
-	return trace.ContextFlags{
-		ContainerStarted: (flags & contStartFlag) != 0,
```
NDStrahilevitz, line 389: Why not just do
`ContainerStarted: (flags & contStartFlag) != 0 && containerId != ""` here instead? Simple one line change.

NDStrahilevitz (commented):
I think we could add a short description in the commit instead referring to the issue. While the race condition case is rather complex to describe, the additional condition here is very simple to explain as a sanity check.

pkg/ebpf/events_pipeline.go
geyslan: Firstly I did as you suggested but I thought that any change in any field related to that return would change the other lines in the future. So I extracted their assignment to single lines (which won't be automatically realigned in a single change).

image

yanivagman (approved):
LGTM

Issue #3264, opened 2023-06-24
context.go: not thread-safe


rafaeldtinoco:
## Description

The following code:

```go
func (store *context) Store(event trace.Event) uint64 {
	id := uint64(store.counter.Read())
	_ = store.counter.Increment()
	store.mutex.Lock()
	store.store[id] = event
	store.mutex.Unlock()
	return id
}
```

Mixes uses more than 1 atomic instruction for the same value and mixes atomics and mutexes. The counter increment should return the new value added (per underlying HW atomic instruction and sync package methods as well).

The locking happening after the atomics Load might make parallel runs to increase the counter multiple times before the lock is acquired by concurrent threads (not a big deal since not being sequencial is not a problem for this particular package (store).

The following code:

```go
func (store *context) Get(id uint64) (trace.Event, bool) {
	store.mutex.RLock()
	contextEvent, ok := store.store[id]
	store.mutex.RUnlock()
	if !ok {
		return trace.Event{}, false
	}
	store.mutex.Lock()
	delete(store.store, id)
	store.mutex.Unlock()
	return contextEvent, true
}
```

Has a race: the mutex is freed after a value LOAD. This is a transaction dependent of ID, should hold the lock for the entire transaction: parallel threads could have added/deleted same ID (replacing the store with new info) in between RUnlock() and delete() would delete the new one and return the old info, for example. This is conceptual problem that would be caught in an unit test (but hardly in the real situation we're using as the event type using Store package aren't that frequent.

Issue #3263, opened 2023-06-24
counter.go: not thread-safe


rafaeldtinoco:
## Description

The current counter.go code is not thread-safe, despite the use of atomics: There can't be load/stores in parallel without holding a mutex (thus checking overflow is tricky).

By adding a simple threaded unit test you can see the error:

```go
func TestIncrement_MultipleThreads(t *testing.T) {
	expected := uint64(100000)
	c := NewCounter(0)

	wg := sync.WaitGroup{}

	for i := 0; i < 100; i++ {
		wg.Add(1)
		go func() {
			for j := 0; j < 1000; j++ {
				err := c.Increment()
				if err != nil {
					t.Error(err)
				}
			}
			wg.Done()
		}()
	}

	wg.Wait()

	if c.Read() != expected {
		t.Errorf("Increment was incorrect, got %d, expected %d", c.Read(), expected)
	}
}
```

You will see the following errors:

```
=== RUN   TestIncrement_MultipleThreads
==================
WARNING: DATA RACE
Read at 0x00c0000b8588 by goroutine 15:
  github.com/aquasecurity/tracee/pkg/counter.(*Counter).incAtomic()
      /home/rafaeldtinoco/work/ebpf/tracee/pkg/counter/counter.go:60 +0x37
  github.com/aquasecurity/tracee/pkg/counter.(*Counter).Increment()
      /home/rafaeldtinoco/work/ebpf/tracee/pkg/counter/counter.go:23 +0xc7
  github.com/aquasecurity/tracee/pkg/counter.TestIncrement_MultipleThreads.func1()
      /home/rafaeldtinoco/work/ebpf/tracee/pkg/counter/counter_test.go:34 +0x6b

Previous write at 0x00c0000b8588 by goroutine 12:
  sync/atomic.AddInt64()
      /usr/lib/go-1.19/src/runtime/race_amd64.s:289 +0xb
  sync/atomic.AddUint64()
      <autogenerated>:1 +0x1b
  github.com/aquasecurity/tracee/pkg/counter.(*Counter).Increment()
      /home/rafaeldtinoco/work/ebpf/tracee/pkg/counter/counter.go:23 +0xc7
  github.com/aquasecurity/tracee/pkg/counter.TestIncrement_MultipleThreads.func1()
      /home/rafaeldtinoco/work/ebpf/tracee/pkg/counter/counter_test.go:34 +0x6b

Goroutine 15 (running) created at:
[...]

PR #3265, opened 2023-06-25
thread-safety issues fix

rafaeldtinoco:
commit 0c41d9bc9 (HEAD -> counter-fixes, rafaeldtinoco/counter-fixes)
Date:   Sun Jun 25 05:26:41 2023

    pkg/events/trigger: fix context concurrency issues
    
    Fixes: #3264

Fixes: #3264

commit 3147f27be
Date:   Sun Jun 25 05:26:01 2023

    pkg/counter: method for increase and read atomically

commit a1bf09c4d
Date:   Sun Jun 25 03:15:56 2023

    pkg/counter: fix concurrency issues with atomics
    
    Fixes: #3263
    
    Rename increment/decrement to increase/decrease: Terms increment &/
    ecrement are technicaly more correct terms, but increase and decrease
    are more widely used, and more common terms, to most audiences.

Fixes: #3263

### DIFF
--- a/pkg/counter/counter.go
+++ b/pkg/counter/counter.go
@@ -19,35 +19,53 @@ func NewCounter(initialValue uint64) Counter {
 
 // Increase increases the counter by the given value (thread-safe).
 func (c *Counter) Increase(values ...uint64) error {
+	_, err := c.IncreaseAndRead(values...)
+	return err
+}
+
+// IncreaseAndRead increases the counter by the given value (thread-safe) and returns the new value.
+func (c *Counter) IncreaseAndRead(values ...uint64) (uint64, error) {
 	if len(values) == 0 {
 		values = append(values, 1)
 	}
 
+	var n uint64
+
 	for _, value := range values {
-		if atomic.AddUint64(&c.value, value) < value {
-			return errors.New("counter overflow")
+		n = atomic.AddUint64(&c.value, value)
+		if n < value {
+			return n, errors.New("counter overflow")
 		}
 	}
 
-	return nil
+	return n, nil // return last known value (atomically incremented)
 }
 
 // Decrease decreases the counter by the given value (thread-safe).
 func (c *Counter) Decrease(values ...uint64) error {
+	_, err := c.DecreaseAndRead(values...)
+	return err
+}
+
+// DecreaseAndRead decreases the counter by the given value (thread-safe) and returns the new value.
+func (c *Counter) DecreaseAndRead(values ...uint64) (uint64, error) {
 	if len(values) == 0 {
 		values = append(values, 1)
 	}
+
+	var n uint64
 
 	for _, value := range values {
 		if value == 0 {
 			continue
 		}
-		if atomic.AddUint64(&c.value, ^uint64(value-1)) == math.MaxUint64 {
-			return errors.New("counter underflow")
+		n = atomic.AddUint64(&c.value, ^uint64(value-1))
+		if n == math.MaxUint64 {
+			return n, errors.New("counter underflow")
 		}
 	}
 
-	return nil
+	return n, nil // return last known value (atomically decremented)
 }
 
 // Read returns the current value of the counter (thread-safe).
--- a/pkg/counter/counter.go
+++ b/pkg/counter/counter.go
@@ -1,96 +1,66 @@
 package counter
 
 import (
+	"errors"
 	"fmt"
+	"math"
 	"strconv"
 	"sync/atomic"
-
-	"github.com/aquasecurity/tracee/pkg/errfmt"
 )
 
 type Counter struct {
-	count uint64
+	value uint64
 }
 
-func NewCounter(v uint64) Counter {
-	return Counter{
-		count: v,
-	}
+// NewCounter creates a new counter with the given initial value.
+func NewCounter(initialValue uint64) Counter {
+	return Counter{value: initialValue}
 }
 
-func (c *Counter) Increment(amount ...uint64) error {
-	if len(amount) == 0 {
-		return c.incAtomic(1)
-	}
-
-	sum, err := sumUint64(amount...)
-	if err != nil {
-		return errfmt.WrapError(err)
+// Increase increases the counter by the given value (thread-safe).
+func (c *Counter) Increase(values ...uint64) error {
+	if len(values) == 0 {
+		values = append(values, 1)
 	}
 
-	return c.incAtomic(sum)
-}
-
-func (c *Counter) Decrement(amount ...uint64) error {
-	if len(amount) == 0 {
-		return c.decAtomic(1)
-	}
-
-	sum, err := sumUint64(amount...)
-	if err != nil {
-		return errfmt.WrapError(err)
-	}
-
-	return c.decAtomic(sum)
-}
-
-func (c *Counter) Read() uint64 {
-	return atomic.LoadUint64(&c.count)
-}
-
-func (c *Counter) Set(val uint64) {
-	atomic.StoreUint64(&c.count, val)
-}
-
-func (c Counter) Format(f fmt.State, r rune) {
-	f.Write([]byte(strconv.FormatUint(c.count, 10)))
-}
-
-func (c *Counter) incAtomic(v uint64) error {
-	if (v + c.count) < c.count {
-		return errorCounterWrapAround()
-	}
-	atomic.AddUint64(&c.count, v)
-
-	return nil
-}
-
-func (c *Counter) decAtomic(v uint64) error {
-	v = ^uint64(v - 1)
-	if (v + c.count) > c.count {
-		return errorCounterWrapAround()
-	}
-	atomic.AddUint64(&c.count, v)
-
-	return nil
-}
-
-func sumUint64(values ...uint64) (uint64, error) {
-	sum := uint64(0)
-	for _, v := range values {
-		sum += v
-		if sum < v {
-			return 0, errorCounterSumWrapAround()
+	for _, value := range values {
+		if atomic.AddUint64(&c.value, value) < value {
+			return errors.New("counter overflow")
 		}
 	}
 
-	return sum, nil
+	return nil
 }
 
-func errorCounterWrapAround() error {
-	return errfmt.Errorf("counter: wrap around")
+// Decrease decreases the counter by the given value (thread-safe).
+func (c *Counter) Decrease(values ...uint64) error {
+	if len(values) == 0 {
+		values = append(values, 1)
+	}
+
+	for _, value := range values {
+		if value == 0 {
+			continue
+		}
+		if atomic.AddUint64(&c.value, ^uint64(value-1)) == math.MaxUint64 {
+			return errors.New("counter underflow")
+		}
+	}
+
+	return nil
 }
 
-func errorCounterSumWrapAround() error {
-	return errfmt.Errorf("counter sum: wrap around")
+// Read returns the current value of the counter (thread-safe).
+func (c *Counter) Read() uint64 {
+	return atomic.LoadUint64(&c.value)
+}
+
+// Set sets the counter to the given value (thread-safe).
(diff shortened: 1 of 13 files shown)


pkg/events/trigger/context.go
```diff
@@ -1,72 +1,83 @@
...
+	mutex   *sync.Mutex
 	counter counter.Counter
 }
 
+// NewContext creates a new context store.
 func NewContext() *context {
 	return &context{
 		store:   make(map[uint64]trace.Event),
-		mutex:   sync.RWMutex{},
-		counter: counter.NewCounter(1),
```
NDStrahilevitz, line 31: FYI, pretty sure I left it at one for a reason (my bad for not documenting it).

rafaeldtinoco: Thats because there was a "read and increment":

```diff
        id := uint64(store.counter.Read())
-       _ = store.counter.Increment()
+       _ = store.counter.Increase()
```

so you would pick 1 and increment to two. Now, its 0 but at the first Load() it is 1 (because of the atomic increment returning the new value). So its the same thing as before, but it starts at 0 due to a single atomic operation and not two.

pkg/counter/counter.go
```diff
@@ -1,96 +1,66 @@
...
-func (c *Counter) Decrement(amount ...uint64) error {
-	if len(amount) == 0 {
-		return c.decAtomic(1)
-	}
-
-	sum, err := sumUint64(amount...)
-	if err != nil {
-		return errfmt.WrapError(err)
+	for _, value := range values {
+		if atomic.AddUint64(&c.value, value) < value {
```
NDStrahilevitz: So now the overflow will be detected in the middle of increasing and not ahead of time. Which means that once an error is returned the counter won't stay in its initial count, meaning there could be a "partial increase".

Not sure if this behavior is better or not, but pointing it out.

pkg/events/trigger/context.go
```diff
@@ -1,72 +1,83 @@
...
-	Get(uint64) (trace.Event, bool)         // get an invoked event context
+	Load(uint64) (trace.Event, bool)        // loads an invoked event context
 	Apply(trace.Event) (trace.Event, error) // apply an invoked event context (implicitly gets the event)
 }
 
-// context is a struct used to store the triggering events context
 type context struct {
 	store   map[uint64]trace.Event
-	mutex   sync.RWMutex
+	mutex   *sync.Mutex
```
NDStrahilevitz, line 25: I guess this is ok because we don't have many frequent user triggered events... But when we do couldn't this be a pretty contentious lock?

NDStrahilevitz (approved):
Changes work and are safe so I'm adding +1. Feel free to resolve the questions too, but since we're near the milestone we might want to delay resolving them (again, if you think it's correct) to the next one (or to a fix to the RC).

pkg/counter/counter.go
rafaeldtinoco: Yes, I'm aware. The only way to guarantee multiple increases is to holding a mutex/lock, which is something we don't want I suppose.

pkg/events/trigger/context.go
rafaeldtinoco: The Store() involves copying trace.Event (needs locking, can't be atomic), the Load involves another copy (also needs locking). We could return a pointer but then the map deletion from the Load() would also require locking. No free lunch :\. We can improve this in the future if needed. I think we should first make sure all locks work correctly and then profile lock contention to optimize those.

rafaeldtinoco (commented):
Thanks for reviewing this.

pkg/counter/counter.go
NDStrahilevitz: Ok, I understand the choice now.
I wonder if we should even allow the API of allowing more then one value increase at a time (I don't think we use it except for the shorthand of `.Increase() => increase 1`).

pkg/events/trigger/context.go
NDStrahilevitz: Ok I see that we really only used the read lock for one happy path anyway... Resolving.

josedonizetti (approved):

pkg/counter/counter.go
```diff
@@ -1,96 +1,84 @@
...
-	if (v + c.count) > c.count {
-		return errorCounterWrapAround()
+	for _, value := range values {
+		if value == 0 {
+			continue
+		}
+		n = atomic.AddUint64(&c.value, ^uint64(value-1))
+		if n == math.MaxUint64 {
+			return n, errors.New("counter underflow")
+		}
```
geyslan, line 65: Decrementing is tricky indeed. As it is now it only considers a wrap of 1 unity. When value is larger than 1, and a wrap occurs, n can be any value greater than the previous value. So this check won't fail even a wrap occurring.

Checking like this won't be a solution either, since we're not locking the whole arithmetic.

```go
	for _, value := range values {
		if value == 0 {
			continue
		}

		originalValue = atomic.LoadUint64(&c.value)
		n = atomic.AddUint64(&c.value, ^uint64(value-1))

		// Detecting underflow by comparing the new value with the original value
		if n > originalValue {
			// Reset the counter to its original value before returning the error.
			atomic.StoreUint64(&c.value, originalValue)
			return n, errors.New("counter underflow")
		}
	}
```

I suspect that case for that we need a mutex.

geyslan (commented):
Sorry for the late review.

pkg/counter/counter.go
```diff
@@ -1,96 +1,84 @@
...
+func (c *Counter) Increase(values ...uint64) error {
+	_, err := c.IncreaseAndRead(values...)
+	return err
 }
 
-func (c *Counter) Decrement(amount ...uint64) error {
-	if len(amount) == 0 {
-		return c.decAtomic(1)
+// IncreaseAndRead increases the counter by the given value (thread-safe) and returns the new value.
+func (c *Counter) IncreaseAndRead(values ...uint64) (uint64, error) {
```
yanivagman, line 27: I think increment/decrement was a better naming. Not sure why you say increase/decrease is more widely used - if you google for "increase counter" you will get most results which say that the operation is called increment/decrement and it causes the value of the counter to be increased/decreased

pkg/counter/counter.go
```diff
@@ -1,96 +1,84 @@
...
-
-func (c *Counter) Read() uint64 {
-	return atomic.LoadUint64(&c.count)
-}
+	for _, value := range values {
+		n = atomic.AddUint64(&c.value, value)
+		if n < value {
+			return n, errors.New("counter overflow")
+		}
+	}
```
yanivagman, line 39: It is better to first sum the values (as was done previously) and then increment than incrementing for every value since:
1. sum is a local variable, thus not exposed to concurrency issues
2. keeping things simple and not using atomic for every value
3. theoretically there is a race condition in the overflow check when two callers have multiple values - if the first caller overflows and we context switch before checking `n < value` then the second caller increments and context switch back to the first caller, where the overflow condition is not necessarily true anymore

pkg/counter/counter.go
yanivagman: Continue with my previous comment - if we first sum the values and then decrement we can avoid having a mutex. I think the previous implementation was good.

yanivagman (commented):
Sorry for the late review, I planned to do it before branching out

pkg/counter/counter.go
geyslan: In this case I suspect we need to check the values sum against wrapping as well.

pkg/counter/counter.go
rafaeldtinoco: Not against a re-re-naming.

pkg/counter/counter.go
rafaeldtinoco: You still have a race there (the same you describe below about context switching). You're reading c.count without an underlaying atomic instruction and the CPU cache coherence might not have updated all CPUs caches (thus the value will be wrong). There is no free lunch, without holding a mutex there can't be LOAD and STORES (multiple operations). That is why a single atomic operation needs to change the value and give previous value.

pkg/counter/counter.go
rafaeldtinoco: You can't "not use atomics" and still be race free, because of HW cache coherence. You are right about the possible race (and it was known when I write). Its better to have 2 errors than 2 increase/decreases.

rafaeldtinoco:
> I think the previous implementation was good.

```
=== RUN   TestIncrement_MultipleThreads
==================
WARNING: DATA RACE
Read at 0x00c0000b8588 by goroutine 15:
  github.com/aquasecurity/tracee/pkg/counter.(*Counter).incAtomic()
      /home/rafaeldtinoco/work/ebpf/tracee/pkg/counter/counter.go:60 +0x37
  github.com/aquasecurity/tracee/pkg/counter.(*Counter).Increment()
      /home/rafaeldtinoco/work/ebpf/tracee/pkg/counter/counter.go:23 +0xc7
  github.com/aquasecurity/tracee/pkg/counter.TestIncrement_MultipleThreads.func1()
      /home/rafaeldtinoco/work/ebpf/tracee/pkg/counter/counter_test.go:34 +0x6b

Previous write at 0x00c0000b8588 by goroutine 12:
  sync/atomic.AddInt64()
      /usr/lib/go-1.19/src/runtime/race_amd64.s:289 +0xb
  sync/atomic.AddUint64()
      <autogenerated>:1 +0x1b
  github.com/aquasecurity/tracee/pkg/counter.(*Counter).Increment()
      /home/rafaeldtinoco/work/ebpf/tracee/pkg/counter/counter.go:23 +0xc7
  github.com/aquasecurity/tracee/pkg/counter.TestIncrement_MultipleThreads.func1()
      /home/rafaeldtinoco/work/ebpf/tracee/pkg/counter/counter_test.go:34 +0x6b
```

We need more tests to check our assumptions. I had many assumptions that proved wrong when creating multi-threaded unit tests.

rafaeldtinoco:
Okay, we all spoke "offline" and agreed to use a mutex for the "overflow" check (as we can't have LOAD/STORE simultaneously AND we want to have multiple values added by some callers). A simple mutex will be a single primitive atomics for multiple values added in those cases (so performance wise it will be fine).

NDStrahilevitz:
If I understand correctly, the reason we need a mutex is because we allow sum increments of multiple numbers in one `Increase()` call.

As we do not ever in the code use something like `Increase(a,b)`, and only, `Increase()` or `Increase(a)`, why not just reduce the API to `Increase(uint)`, use one atomic addition operation and resolve it that way? For the case we want the "syntax sugar" of increasing one as `Increase()`, we can add a separate method of `Increment()` (which would be `Increase(1)` under the hood).

rafaeldtinoco:
> As we do not ever in the code use something like `Increase(a,b)`, and only, `Increase()` or `Increase(a)`, why not just reduce the API to `Increase(uint)`, use one atomic addition operation and resolve it that way?

Yes, we all reached that point of conclusion today and I think you are right in all that you just said. We would have to loop externally or in a new method like you said (and the overflow check would be per atomic INC operation, as its the only way to get a value and increment in the same operation).

We also thought that a mutex could be beneficial: a single atomic/primitive for anything else we want (checking overflow AND multiple values increase). Thus my last comment.

pkg/counter/counter.go
rafaeldtinoco: There is no usecase (1) in Tracee, so we are safe to use atomics only and check for overflow after atomic operation happened. Will put a PR together.

rafaeldtinoco:
@NDStrahilevitz https://github.com/aquasecurity/tracee/pull/3276 I believe it address all of our concerns and is race-free without mutexes. Suggestions from everybody =).

pkg/counter/counter.go
rafaeldtinoco: https://github.com/aquasecurity/tracee/pull/3276 I believe it address all of our concerns and is race-free without mutexes. Suggestions from everybody =).

PR #3271, opened 2023-06-26
build: remove signing from snapshot

josedonizetti:

Now that we have cosign for our release, we should remove it from the snapshot. The snapshot is generated daily, which generates a lot of signatures not used in our docker hub.


rafaeldtinoco (approved):
LGTM

PR #3276, opened 2023-06-27
fix(pkg/counter): finish making counter atomic

rafaeldtinoco:
a6817c18d **fix(pkg/counter): finish making counter atomic** _<sub>(2023/Jun/27) Rafael David Tinoco \<raf>

```
Last changes renamed Increment/Decrement to Increase/Decrease. This was
reverted after discussions.

With this new approach:

- Overflow is detected in both cases (and proved by test).
- Thread safety is guaranteed by atomic Add/Sub operations.
- Unit tests prove all assumptions and thread safety for all cases.

NOTE: The issue is now fully fixed...

Originally, the bug was only about reading counter value in parallel to
atomic operation (multiple LOAD/STORE simultaneously, and some not
protected by atomics).

Previous fix, made by commit 1e54ce292, fixed data race but overflow errors
could be reported more than once. It also had atomic operations for each
given argument (if multiple arguments).

This version addresses all concerns.
```

Heavily influenced by conversations after https://github.com/aquasecurity/tracee/pull/3265 (and offline).

### DIFF
--- a/pkg/counter/counter.go
+++ b/pkg/counter/counter.go
@@ -12,73 +12,86 @@ type Counter struct {
 	value uint64
 }
 
-// NewCounter creates a new counter with the given initial value.
 func NewCounter(initialValue uint64) Counter {
 	return Counter{value: initialValue}
 }
 
-// Increase increases the counter by the given value (thread-safe).
-func (c *Counter) Increase(values ...uint64) error {
-	_, err := c.IncreaseAndRead(values...)
+// Increment
+
+// Increment increments counter by given value (default: 1, thread-safe).
+func (c *Counter) Increment(x ...uint64) error {
+	var err error
+
+	val := uint64(1)
+	if len(x) != 0 {
+		for _, v := range x {
+			val += v
+		}
+		val-- // initial 1
+	}
+
+	// NOTE: checking if val > 0 adds ~200ns/op in benchmarking: not worth if
+	//       not proved that amount of times Increment(0) is called makes it
+	//       worth it.
+	_, err = c.IncrementValueAndRead(val)
+
 	return err
 }
 
-// IncreaseAndRead increases the counter by the given value (thread-safe) and returns the new value.
-func (c *Counter) IncreaseAndRead(values ...uint64) (uint64, error) {
-	if len(values) == 0 {
-		values = append(values, 1)
+// IncrementValueAndRead increments counter by given value and returns the new value (thread-safe).
+func (c *Counter) IncrementValueAndRead(x uint64) (uint64, error) {
+	var err error
+
+	n := atomic.AddUint64(&c.value, x)
+	if n < x {
+		err = errors.New("counter overflow")
 	}
 
-	var n uint64
-
-	for _, value := range values {
-		n = atomic.AddUint64(&c.value, value)
-		if n < value {
-			return n, errors.New("counter overflow")
-		}
-	}
-
-	return n, nil // return last known value (atomically incremented)
+	return n, err
 }
 
-// Decrease decreases the counter by the given value (thread-safe).
-func (c *Counter) Decrease(values ...uint64) error {
-	_, err := c.DecreaseAndRead(values...)
+// Decrement
+
+// Decrement decrements counter by given value (default: 1, thread-safe).
+func (c *Counter) Decrement(x ...uint64) error {
+	val := uint64(1)
+	if len(x) != 0 {
+		for _, v := range x {
+			val += v
+		}
+		val-- // initial 1
+	}
+
+	// NOTE: checking if val > 0 adds ~200ns/op in benchmarking: not worth if
+	//       not proved that amount of times Decrement(0) is called makes it
+	//       worth it.
+	_, err := c.DecrementValueAndRead(val)
+
 	return err
 }
 
-// DecreaseAndRead decreases the counter by the given value (thread-safe) and returns the new value.
-func (c *Counter) DecreaseAndRead(values ...uint64) (uint64, error) {
-	if len(values) == 0 {
-		values = append(values, 1)
+// DecrementValueAndRead decrements counter by given value and returns the new value (thread-safe).
+func (c *Counter) DecrementValueAndRead(x uint64) (uint64, error) {
+	var err error
+
+	n := atomic.AddUint64(&c.value, ^uint64(x-1))
+	if n > math.MaxUint64-x {
+		err = errors.New("counter underflow")
 	}
 
-	var n uint64
-
-	for _, value := range values {
-		if value == 0 {
-			continue
-		}
-		n = atomic.AddUint64(&c.value, ^uint64(value-1))
-		if n == math.MaxUint64 {
-			return n, errors.New("counter underflow")
-		}
-	}
-
-	return n, nil // return last known value (atomically decremented)
+	return n, err
 }
 
-// Read returns the current value of the counter (thread-safe).
-func (c *Counter) Read() uint64 {
-	return atomic.LoadUint64(&c.value)
-}
+// Setters and Getters
 
-// Set sets the counter to the given value (thread-safe).
 func (c *Counter) Set(value uint64) {
 	atomic.StoreUint64(&c.value, value)
 }
 
-// Format implements fmt.Formatter (thread-safe).
+func (c *Counter) Get() uint64 {
+	return atomic.LoadUint64(&c.value)
+}
+
 func (c *Counter) Format(f fmt.State, r rune) {
-	f.Write([]byte(strconv.FormatUint(c.Read(), 10)))
+	f.Write([]byte(strconv.FormatUint(c.Get(), 10)))
 }
--- a/pkg/counter/counter_test.go
+++ b/pkg/counter/counter_test.go
@@ -8,46 +8,96 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-//
-// Increase
-//
+// Increment
 
-// TestZeroedIncrease tests that the counter is not increased when 0 is given.
-func TestZeroedIncrease(t *testing.T) {
-	expected := uint64(1)
-	c := NewCounter(1)
-
-	err := c.Increase(0)
-
-	require.NoError(t, err)
-	require.Equal(t, expected, c.Read())
-}
-
-// TestIncreaseAndRead tests that the counter is increased by 1 and returns the new value.
-func TestIncreaseAndRead(t *testing.T) {
+func TestIncrement(t *testing.T) {
 	expected := uint64(1)
 	c := NewCounter(0)
 
-	n, err := c.IncreaseAndRead()
+	err := c.Increment()
+
+	require.NoError(t, err)
+	require.Equal(t, expected, c.Get())
+}
+
+func TestIncrementWithValue(t *testing.T) {
+	expected := uint64(9)
+	c := NewCounter(0)
+
+	err := c.Increment(9)
+
+	require.NoError(t, err)
+	require.Equal(t, expected, c.Get())
+}
+
+func TestIncrementWithMultipleValue(t *testing.T) {
+	expected := uint64(9)
+	c := NewCounter(0)
+
+	err := c.Increment(3, 3, 3)
+
+	require.NoError(t, err)
+	require.Equal(t, expected, c.Get())
+}
+
+func TestIncrementAndRead(t *testing.T) {
+	expected := uint64(9)
+	c := NewCounter(0)
+
+	n, err := c.IncrementValueAndRead(9)
 
 	require.NoError(t, err)
 	require.Equal(t, expected, n)
-	require.Equal(t, expected, c.Read())
+	require.Equal(t, expected, c.Get())
 }
 
-// TestIncrease tests that the counter is increased by 1.
-func TestIncrease(t *testing.T) {
-	expected := uint64(1)
-	c := NewCounter(0)
+func TestZeroedIncrementWithValue(t *testing.T) {
+	expected := uint64(9)
+	c := NewCounter(9)
 
-	err := c.Increase()
+	err := c.Increment(0)
 
 	require.NoError(t, err)
-	require.Equal(t, expected, c.Read())
+	require.Equal(t, expected, c.Get())
 }
 
-// TestIncrease_MultipleThreads is a stress test to check thread safety for increased.
-func TestIncrease_MultipleThreads(t *testing.T) {
+func TestZeroedIncrementWithMultipleValue(t *testing.T) {
+	expected := uint64(9)
+	c := NewCounter(9)
+
+	err := c.Increment(0, 0, 0)
+
+	require.NoError(t, err)
+	require.Equal(t, expected, c.Get())
+}
+
+func TestIncrementOverflow(t *testing.T) {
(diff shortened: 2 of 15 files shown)


pkg/counter/counter.go
```diff
@@ -12,73 +12,74 @@ type Counter struct {
...
-	if len(values) == 0 {
-		values = append(values, 1)
+// Increment increments the counter by given value (default: 1, thread-safe).
+func (c *Counter) Increment(x ...uint64) error {
+	val := uint64(1)
+	if len(x) != 0 {
+		for _, v := range x {
+			val += v
+		}
+		val--
```
geyslan, line 28: `val++`?

geyslan: Oh, never mind.

pkg/counter/counter.go
```diff
@@ -12,73 +12,74 @@ type Counter struct {
...
-	return err
-}
+// Increment
 
-// IncreaseAndRead increases the counter by the given value (thread-safe) and returns the new value.
-func (c *Counter) IncreaseAndRead(values ...uint64) (uint64, error) {
-	if len(values) == 0 {
-		values = append(values, 1)
+// Increment increments the counter by given value (default: 1, thread-safe).
+func (c *Counter) Increment(x ...uint64) error {
```
geyslan, line 22: Taking care of corner cases like `x = [0]` or `x = [0,0,...,0]` and sum overflow, I would suggest changing to:

```go
func (c *Counter) Increment(x ...uint64) error {
	val := uint64(0)
	if len(x) == 0 {
		val = 1
	} else {
		for _, v := range x {
			// Check for overflow
			if val > math.MaxUint64-v {
				return errors.New("increment values sum overflow")
			}
			val += v
		}
	}

	if val == 0 {
		return nil
	}

	_, err := c.IncrementValueAndRead(val)
	return err
}
```

pkg/counter/counter.go
```diff
@@ -12,73 +12,74 @@ type Counter struct {
...
+		}
+		val--
 	}
+	_, err := c.IncrementValueAndRead(val)
+	return err
+}
 
-	var n uint64
+// IncrementValueAndRead increments the counter by given value and returns the new value (thread-safe).
+func (c *Counter) IncrementValueAndRead(x uint64) (uint64, error) {
```
geyslan, line 35: Here we can simplify to:

```go
func (c *Counter) IncrementValueAndRead(x uint64) (uint64, error) {
	n := atomic.AddUint64(&c.value, x)
	if n < x {
		return n, errors.New("counter overflow")
	}

	return n, nil
}
```

pkg/counter/counter.go
```diff
@@ -12,73 +12,74 @@ type Counter struct {
...
-	return err
-}
+// Decrement
 
-// DecreaseAndRead decreases the counter by the given value (thread-safe) and returns the new value.
-func (c *Counter) DecreaseAndRead(values ...uint64) (uint64, error) {
-	if len(values) == 0 {
-		values = append(values, 1)
+// Decrement decrements the counter by given value (default: 1, thread-safe).
+func (c *Counter) Decrement(x ...uint64) error {
```
geyslan, line 49: The same checks for Decrement:

```go
func (c *Counter) Decrement(x ...uint64) error {
	val := uint64(0)
	if len(x) == 0 {
		val = 1
	} else {
		for _, v := range x {
			// Check for overflow in summation
			if val > math.MaxUint64-v {
				return errors.New("decrement values sum overflow")
			}
			val += v
		}
	}

	if val == 0 {
		return nil
	}

	_, err := c.DecrementValueAndRead(val)
	return err
}
```

pkg/counter/counter_test.go
```diff
@@ -9,45 +9,87 @@ import (
...
+	require.NoError(t, err)
+	require.Equal(t, expected, n)
+	require.Equal(t, expected, c.Get())
+}
+
+func TestZeroedIncrementWithValue(t *testing.T) {
+	expected := uint64(9)
+	c := NewCounter(9)
+
+	err := c.Increment(0)
```
geyslan, line 60:
```suggestion
	err := c.Increment(0, 0, 0)
```

pkg/counter/counter_test.go
```diff
@@ -66,95 +108,99 @@ func TestIncrease_MultipleThreads(t *testing.T) {
...
-// TestDecreaseAndRead tests that the counter is decreased by 1 and returns the new value.
-func TestDecreaseAndRead(t *testing.T) {
-	expected := uint64(0)
-	c := NewCounter(1)
+func TestZeroedDecrementWithValue(t *testing.T) {
+	expected := uint64(9)
+	c := NewCounter(9)
 
-	n, err := c.DecreaseAndRead()
+	err := c.Decrement(0)
```
geyslan, line 163:
```suggestion
	err := c.Decrement(0, 0, 0
```

geyslan (changes requested):
I reviewed and identified some corner cases to be checked. Also suggested a change in the zeroed value tests to contemplate slice of zeroes.

pkg/counter/counter.go
geyslan: Just a correction in my suggestion, the sum overflow check should be: `if val+v < val {`.

pkg/counter/counter.go
rafaeldtinoco: Could you elaborate ?

geyslan: I wrongly thought that was a typo. But noticed that it was removing the initial 1.

pkg/counter/counter.go
rafaeldtinoco: Your suggestion is:

- to short circuit zero values (mm'okay)
- not to continue increasing in case of an overflow (why?) wouldn't it be better to mark how many overflows there was somewhere else instead ? and if we did, i would have to keep track of instanced counters (which is not what we've done so far).

pkg/counter/counter.go
rafaeldtinoco: The only logical difference I see, despite styling, is not checking x > 0 (which is actually a leftover from other attempts of mine). If that is the case, than yep, that can be removed.

pkg/counter/counter.go
rafaeldtinoco: ditto

pkg/counter/counter_test.go
rafaeldtinoco: image

I guess you want to test the "short circuit" for one or more zeroes, either way code works.

pkg/counter/counter.go
rafaeldtinoco: How much faster do you think we will be checking for zeros instead of just doing the ops ?

rafaeldtinoco: My version:

```
goos: linux
goarch: amd64
pkg: github.com/aquasecurity/tracee/pkg/counter
cpu: 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
BenchmarkIncrement
BenchmarkIncrement-6   	 3241423	       357.1 ns/op	       0 B/op	       0 allocs/op
PASS
coverage: 32.1% of statements
```

Your version:

```
goos: linux
goarch: amd64
pkg: github.com/aquasecurity/tracee/pkg/counter
cpu: 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
BenchmarkIncrement
BenchmarkIncrement-6   	 2338204	       506.2 ns/op	       0 B/op	       0 allocs/op
PASS
coverage: 34.4% of statements
ok  	github.com/aquasecurity/tracee/pkg/counter	1.737s
```

pkg/counter/counter_test.go
rafaeldtinoco: Im adding this test AND benchmark tests.

pkg/counter/counter_test.go
rafaeldtinoco: Im adding this test and bencmark tests.

pkg/counter/counter.go
geyslan: >     * not to continue increasing in case of an overflow (why?) wouldn't it be better to mark how many overflows there was somewhere else instead?

If we ignore sum overflow, we masquerade wrong inc/decs on the second stage (Inc/DcrementValueAndRead). Suppose we overflowed (wrapped) the sum val to 1 and ignored the overflow. The current value is 2, adding that 1 from the sum, we got a value not overflowed so not detected as an error. That's why I consider that all stages need to detect possible wrapping and err.

pkg/counter/counter.go
geyslan: Yep, the suggestion was merely figurative as I copied n pasted local changes. Ignore the styling changes.

rafaeldtinoco: Okay then, removing the x > 0 leftover, thx.

pkg/counter/counter.go
rafaeldtinoco: You mean Increment(1, Math.MaxUint64) would have 2 overflows and not only 1 ?

pkg/counter/counter_test.go
geyslan: Yes, I wanted to test the possible sum resulting in zero since the PR is decrementing val by 1 in the case of len(x) != 0, even sum being 0 (underflowing), what would result in a wrong inc/dec.

pkg/counter/counter.go
geyslan: > How much faster do you think we will be checking for zeros instead of just doing the ops ?

Checking the sum result for zero would avoid the atomic Add and its wrapping check.

pkg/counter/counter_test.go
rafaeldtinoco: I see the point, and despite the shortcircuit or not its worth testing that.

pkg/counter/counter.go
rafaeldtinoco: In theory you are right (about checking wraps twice) but I don't think increases will be made by MaxUint64() in any realistic case, so I don't think its worth checking apart from seeing if the *actual* counter was overflowed.

rafaeldtinoco: From 357 ns/op -> 500 ns/op by checking if val is zero.
IF testing only Increment(0) in the benchmark loop we do ~141 ns/op.

I don't think its worth the short circuit unless we know the frequency of zeros we're adding in overall tracee execution (orelse we're ALWAYS adding ~100ns/op to something that might not be that frequent).

rafaeldtinoco: So Id rather have the code I proposed for the reasons specified above.

geyslan: I really don't see current use cases for slice of uints (params). So I would vote to get rid of that API; it's making the counter complex for not real world cases. WDYT?

rafaeldtinoco:
I've included benchmark tests, and removed the extra check for `x > 0` that was a leftover. All the rest I believe we are good.

pkg/counter/counter.go
rafaeldtinoco: > In theory you are right (about checking wraps twice) but I don't think increases will be made by MaxUint64() in any realistic case, so I don't think its worth checking apart from seeing if the _actual_ counter was overflowed (when we consider benchmarking).

Added a comment in the git log saying that isn't supported (and we are not doing for performance reasons).

rafaeldtinoco:
Forgot the decrease(0,0,0) case test. Will add.

pkg/counter/counter.go
rafaeldtinoco: Funny discovery: 

```go
func BenchmarkDecrement(b *testing.B) {
	c := NewCounter(uint64(b.N))

	for i := b.N; i > 0; i-- {
		err := c.Decrement()
		if err != nil {
			b.Fatal(err)
		}
	}
}

func BenchmarkDecrementWithValue(b *testing.B) {
	c := NewCounter(uint64(b.N))

	for i := b.N; i > 0; i-- {
		err := c.Decrement(1)
		if err != nil {
			b.Fatal(err)
		}
	}
}
```

The benchmark shows that the second is almost 2x slower than the first =) (likely because of stack management, argument passing and calling convention).

rafaeldtinoco: Even funier: having more than 1 argument is faster than having a single argument (likely because how golang decides to pass the argument: through registers or stack, because of calling convention as well). Crazy.

rafaeldtinoco:
@geyslan Added tests for (0) and (0,0,0). Added benchmarks for all cases:

- Inc()
- Inc(x)
- Inc(0)
- Inc(0,0,0)

The extra cost for checking if val > 0 (about 200ns) is almost 1/2 of the cost of the function entirely (so unless > 50% of the calls of Increment are Increment(0) it wouldn't be worth adding a short circuit).

Hope that answers your concerns.

pkg/counter/counter.go
rafaeldtinoco: > I really don't see current use cases for slice of uints (params). So I would vote to get rid of that API; it's making the counter complex for not real world cases. WDYT?

I agree (from the beginning), it would make the counter faster actually (not having to deal with a slice each inc/dec).

geyslan: > I agree (from the beginning), it would make the counter faster actually (not having to deal with a slice each inc/dec).

And would have a coherent logic without unlikely corner cases.

pkg/counter/counter.go
NDStrahilevitz: Might be nice to add a comment here, but not critical.

pkg/counter/counter_test.go
```diff
@@ -238,5 +268,127 @@ func TestIncreaseDecrease_MultipleThreads(t *testing.T) {
 
 	wg.Wait()
 
-	require.Equal(t, expected, c.Read())
+	require.Equal(t, expected, c.Get())
+}
+
+//
+// Benchmarks
```
NDStrahilevitz, line 275: nit: maybe separate to `counter_bench_test.go` file?

NDStrahilevitz (approved):
Simple and more importantly thread-safe, LGTM. Added some styling suggestions but nothing critical.

> NOTE: IF we decide to remove the []uint64 as the argument of the API, I can change this PR.

That said, i'm definitely in favor of this.

pkg/counter/counter.go
```diff
@@ -12,73 +12,86 @@ type Counter struct {
...
-	for _, value := range values {
-		n = atomic.AddUint64(&c.value, value)
-		if n < value {
-			return n, errors.New("counter overflow")
+	val := uint64(1)
+	if len(x) != 0 {
+		for _, v := range x {
+			val += v
 		}
+		val--
```
yanivagman, line 30:
```suggestion
        var val uint64
        if len(x) == 0 {
            val = 1
        } else {
		for _, v := range x {
			val += v
		}
```

yanivagman (approved):
LGTM

rafaeldtinoco:
Working on fixing:

```
Stats: {EventCount:{value:1148} EventsFiltered:{value:0} NetCapCount:{value:0} BPFLogsCount:{value:0} ErrorCount:{value:0} LostEvCount:{value:0} LostWrCount:{value:0} LostNtCapCount:{value:0} LostBPFLogsCount:{value:0}}
```

Issue #3279, opened 2023-06-28
derived events not triggering for filtered base events


josedonizetti:
eg: 

```yaml
name: pol1
description: general policy
scope:
  - global
defaultActions: 
  - log
rules:
  - event: cgroup_mkdir
    filter:
      - args.cgroup_path!=/system.slice*
  - event: container_create
```

PR #3280, opened 2023-06-28
fix: derived event not triggering if base filtered

josedonizetti:

Fix https://github.com/aquasecurity/tracee/issues/3279

The bug happens because we pass a different event to [`matchPolicies`](https://github.com/aquasecurity/tracee/blob/main/pkg/ebpf/events_pipeline.go#L522) inside the for, instead of passing the [event that goes down the pipeline](https://github.com/aquasecurity/tracee/blob/main/pkg/ebpf/events_pipeline.go#L531), so the logic of [changing the bitmap](https://github.com/aquasecurity/tracee/blob/main/pkg/ebpf/events_pipeline.go#L366) is lost, because it happened on different events.

### 2. Explain how to test it

```
name: pol1
description: general policy
scope:
  - global
defaultActions: 
  - log
rules:
  - event: cgroup_mkdir
    filters:
      - args.cgroup_path!=/system.slice*
  - event: container_create
```

```
./dist/tracee -p policy.yaml
```

```
docker run ubuntu:latest ls
```

The `container_created` should be triggered but not the `cgroup_mkdir`.

### DIFF
--- a/pkg/ebpf/events_pipeline.go
+++ b/pkg/ebpf/events_pipeline.go
@@ -510,16 +510,16 @@ func (t *Tracee) deriveEvents(ctx context.Context, in <-chan *trace.Event) (
 					t.handleError(err)
 				}
 
-				for i, derivative := range derivatives {
+				for i := range derivatives {
 					// Skip events that dont work with filtering due to missing types being handled.
 					// https://github.com/aquasecurity/tracee/issues/2486
-					switch events.ID(derivative.EventID) {
+					switch events.ID(derivatives[i].EventID) {
 					case events.SymbolsLoaded:
 					case events.SharedObjectLoaded:
 					case events.PrintMemDump:
 					default:
 						// Derived events might need filtering as well
-						if t.matchPolicies(&derivative) == 0 {
+						if t.matchPolicies(&derivatives[i]) == 0 {
 							_ = t.stats.EventsFiltered.Increment()
 							continue
 						}


yanivagman (approved):
nice catch!
Please cherry-pick to v0.16.0 branch as well

pkg/ebpf/events_pipeline.go
```diff
@@ -519,7 +519,7 @@ func (t *Tracee) deriveEvents(ctx context.Context, in <-chan *trace.Event) (
 					case events.PrintMemDump:
 					default:
 						// Derived events might need filtering as well
-						if t.matchPolicies(&derivative) == 0 {
+						if t.matchPolicies(&derivatives[i]) == 0 {
```
geyslan, line 522: Good catch.

What about to remove the copied var usage above (derivative) and make all accesses via index to avoid further errors like this?

```go
				for i := range derivatives {
					// Skip events that dont work with filtering due to missing types being handled.
					// https://github.com/aquasecurity/tracee/issues/2486
switch events.ID(derivatives[i].EventID) {
```

yanivagman: Good idea. This will also avoid the extra copy of creating derivative instance in the loop

geyslan: > Good catch.
> 
> What about to remove the copied var usage above (derivative) and make all accesses via index to avoid further errors like that?
> 
> ```go
> 				for i := range derivatives {
> 					// Skip events that dont work with filtering due to missing types being handled.
> 					// https://github.com/aquasecurity/tracee/issues/2486
> switch events.ID(derivatives[i].EventID) {
> ```

And the code suggestion was just to get us on the same page. Please be my guest on code style or naming, nits, anything 😄.

geyslan (approved):

Issue #3207, opened 2023-06-07
signatures with filtered base events not triggered


josedonizetti:
## Description

```

name: signature_events

description: traces all signature events

defaultAction: log

scope:

  - global

rules:

  - event: ptrace

    filter:

      - userId=1000

  - event: anti_debugging

```

## Output of `tracee version`:

```
(paste your output here)
```

## Output of `uname -a`:

```
(paste your output here)
```

josedonizetti:
@geyslan I'll create a separate issue to address your policy, because the bug for derived events is different than the bug for signatures.

PR #3281, opened 2023-06-28
fix: signature event not triggering if base filtered

josedonizetti:
Fix https://github.com/aquasecurity/tracee/issues/3207

This PR builds a map between a base event and its signatures, allowing us to check after the `decode` stage if the event should be passed down the pipeline because there is a signature depending on it, or in case the userland filter out the event, ignore it and continue. 

### 2. Explain how to test it

```

name: signature_events
description: traces all signature events
defaultActions: 
  - log
scope:
  - global
rules:
  - event: ptrace
    filters:
      - userId=1000
  - event: anti_debugging
```

```
./dist/tracee -p policy.yaml
```

```
sudo strace ls
```

The `anti_debugging` signature should be triggered even though we are filtering the `ptrace` event for the uid 1000.

### DIFF
--- a/pkg/ebpf/tracee.go
+++ b/pkg/ebpf/tracee.go
@@ -77,6 +77,7 @@ type Tracee struct {
 	eventsSorter     *sorting.EventsChronologicalSorter
 	eventProcessor   map[events.ID][]func(evt *trace.Event) error
 	eventDerivations derive.Table
+	eventSignatures  map[events.ID]bool
 	// Artifacts
 	fileHashes     *lru.Cache[string, fileExecInfo]
 	capturedFiles  map[string]int64
@@ -181,6 +182,7 @@ func GetCaptureEventsList(cfg config.Config) map[events.ID]eventConfig {
 
 func (t *Tracee) handleEventsDependencies(eventId events.ID, submitMap uint64) {
 	definition := events.Definitions.Get(eventId)
+
 	eDependencies := definition.Dependencies
 	for _, dependentEvent := range eDependencies.Events {
 		ec, ok := t.events[dependentEvent.EventID]
@@ -190,6 +192,10 @@ func (t *Tracee) handleEventsDependencies(eventId events.ID, submitMap uint64) {
 		}
 		ec.submit |= submitMap
 		t.events[dependentEvent.EventID] = ec
+
+		if events.IsASignatureEvent(eventId) {
+			t.eventSignatures[dependentEvent.EventID] = true
+		}
 	}
 }
 
@@ -205,12 +211,13 @@ func New(cfg config.Config) (*Tracee, error) {
 	// Create Tracee
 
 	t := &Tracee{
-		config:        cfg,
-		done:          make(chan struct{}),
-		writtenFiles:  make(map[string]string),
-		readFiles:     make(map[string]string),
-		capturedFiles: make(map[string]int64),
-		events:        GetEssentialEventsList(),
+		config:          cfg,
+		done:            make(chan struct{}),
+		writtenFiles:    make(map[string]string),
+		readFiles:       make(map[string]string),
+		capturedFiles:   make(map[string]int64),
+		events:          GetEssentialEventsList(),
+		eventSignatures: make(map[events.ID]bool),
 	}
 
 	// Initialize capabilities rings soon
--- a/pkg/filters/args.go
+++ b/pkg/filters/args.go
@@ -106,7 +106,7 @@ func (filter *ArgFilter) Parse(filterName string, operatorAndValues string, even
 	}
 
 	// if the event is a signature event, we allow filtering on dynamic argument
-	if !argFound && !eventDefinition.IsASignatureEvent() {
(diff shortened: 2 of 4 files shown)


geyslan (approved):
LGTM

Tested with:

`sudo ./dist/tracee -p pol1.yaml -p pol2.yaml -p pol3.yaml -o json` and `sudo strace ls`

```json
{"timestamp":1687954737858348592,"threadStartTime":1687954737858256434,"processorId":2,"processId":214948,"cgroupId":15952,"threadId":214948,"parentProcessId":214946,"hostProcessId":214948,"hostThreadId":214948,"hostParentProcessId":214946,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"strace","hostName":"hb","containerId":"","container":{},"kubernetes":{},"eventId":"6018","eventName":"anti_debugging","matchedPolicies":["pol1","pol3"],"argsNum":0,"returnValue":0,"syscall":"","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"args":[],"metadata":{"Version":"1","Description":"A process used anti-debugging techniques to block a debugger. Malware use anti-debugging to stay invisible and inhibit analysis of their behavior.","Tags":null,"Properties":{"Category":"defense-evasion","Kubernetes_Technique":"","Severity":1,"Technique":"Debugger Evasion","external_id":"T1622","id":"attack-pattern--e4dc8c01-417f-458d-9ee0-bb0617c1b391","signatureID":"TRC-102","signatureName":"Anti-Debugging detected"}}}
```

`cat pol*.yaml`

```yaml
name: pol1
description: traces all signature events
defaultActions:
  - log
scope:
  - global
rules:
  - event: ptrace
    filters:
      - userId=1000
  - event: anti_debugging
```

```yaml
name: pol2
description: traces all signature events
defaultActions:
  - log
scope:
  - global
rules:
  - event: ptrace
    filters:
      - userId=1001
```

```yaml
name: pol3
description: traces all signature events
defaultActions:
  - log
scope:
  - global
rules:
  - event: anti_debugging
    filters:
      - userId=0
```

pkg/ebpf/tracee.go
```diff
@@ -190,6 +195,10 @@ func (t *Tracee) handleEventsDependencies(eventId events.ID, submitMap uint64) {
 		}
 		ec.submit |= submitMap
 		t.events[dependentEvent.EventID] = ec
+
+		if isSignatureEvent {
+			t.eventSignatures[dependentEvent.EventID] = true
```
yanivagman, line 200: Can't we just initialize this map above and get rid of the isSignatureEvent variable?

josedonizetti: didn't understand, can you expand?

yanivagman: seems like this was fixed in the recent push

pkg/filters/args.go
```diff
@@ -106,7 +106,7 @@ func (filter *ArgFilter) Parse(filterName string, operatorAndValues string, even
 	}
 
 	// if the event is a signature event, we allow filtering on dynamic argument
-	if !argFound && !eventDefinition.IsASignatureEvent() {
+	if !argFound && !events.IsASignatureEvent(eventDefinition.ID32Bit) {
```
yanivagman, line 109: Are you sure we need to use ID32Bit and not ID? ID32Bit is intended to map 64 bit syscalls to their matching 32 bit compat syscalls

geyslan (approved):
LGTM

pkg/filters/args.go
josedonizetti: @yanivagman maybe, but what I see is -> 
[`Event`](https://github.com/aquasecurity/tracee/blob/main/pkg/events/events.go#L177C1-L177C1) which is what we get when we do an event definition [`Get`](https://github.com/aquasecurity/tracee/blob/main/pkg/events/events.go#L239) or [`GetSafe`](https://github.com/aquasecurity/tracee/blob/main/pkg/events/events.go#L245), and Event only has [`ID32Bit`](https://github.com/aquasecurity/tracee/blob/main/pkg/events/events.go#L178), which has type `ID`, the field ID itself only exists for [`eventDependency`](https://github.com/aquasecurity/tracee/blob/main/pkg/events/events.go#L170), but both have the same type. 

When we define an `Event` for the event definitions table, I only see we setting `ID32Bit` -> https://github.com/aquasecurity/tracee/blob/main/pkg/events/events.go#L509

Isn't it correct than?

yanivagman: The key of the events map in the eventsDefinitions struct is also of type ID.
You can simply use the id local variable that is already defined above in this function

josedonizetti: sure

yanivagman (approved):
LGTM

pkg/filters/args.go
rafaeldtinoco: Related: https://github.com/aquasecurity/tracee/pull/3293

PR #3284, opened 2023-06-28
fix(pkg/counter): finish making counter atomic (#3276)

rafaeldtinoco:
Last changes renamed Increment/Decrement to Increase/Decrease. This was reverted after discussions.

With this new approach:

- Overflow is detected in both cases (and proved by test).
- Thread safety is guaranteed by atomic Add/Sub operations.
- Unit tests prove all assumptions and thread safety for all cases.
- Benchmark tests can be used to test other approaches in the future.
- Overflow of a sum of all given arguments isn't detected (avoid overhead).

NOTE: The issue is now fully fixed...

Originally, the bug was only about reading counter value in parallel to atomic operation (multiple LOAD/STORE simultaneously, and some not protected by atomics).

Previous fix, made by commit 1e54ce292, fixed data race but overflow errors could be reported more than once. It also had atomic operations for each given argument (if multiple arguments).

This version addresses all concerns.

commit: 0b97409c5 (main), cherry-pick

### DIFF
--- a/pkg/counter/counter.go
+++ b/pkg/counter/counter.go
@@ -12,73 +12,86 @@ type Counter struct {
 	value uint64
 }
 
-// NewCounter creates a new counter with the given initial value.
 func NewCounter(initialValue uint64) Counter {
 	return Counter{value: initialValue}
 }
 
-// Increase increases the counter by the given value (thread-safe).
-func (c *Counter) Increase(values ...uint64) error {
-	_, err := c.IncreaseAndRead(values...)
+// Increment
+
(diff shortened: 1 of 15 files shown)


rafaeldtinoco:
@NDStrahilevitz note the last line:

commit: XXXX (where), what

when = main for now
what = backport or cherry-pick

if backport

[backport]
- explanations

NDStrahilevitz (approved):
Format noted, i'll attempt the cherry-pick for #3285 if you don't mind. LGTM.

PR #3298, opened 2023-07-04
events: update syscall_pathname for security_file_open

OriGlassman:

syscall_pathname was fetched only for open(at) syscalls, thus was empty for security_file_open originated from execve(at) syscall

### 2. Explain how to test it

./tracee -f e=security_file_open


NDStrahilevitz:
Test case doesn't include a reproducer for execve and execveat. Would you mind adding such a reproducer (and potentially an instrumentation test?)

rafaeldtinoco (approved):
LGTM

Issue #3303, opened 2023-07-07
Bug: metrics cmdline does not enable metrics


rafaeldtinoco:
## Description

After commit:

commit 0213c0883
Author: Jose Donizetti <jdbjunior@gmail.com>
Date:   Thu May 18 17:45:28 2023

    tracee: add ready callback
    
    The ready callback is executed once tracee is fully initialized,
    and ready to receive events. This feature will be used for the
    `/healthz` endpoint, which will only start to respond "OK" once
    tracee is ready.

The following commandline:

```
$ sudo ./dist/tracee --metrics
```

Stopped registering prometheus stats. That happens because the prometheus metrics are registered at the end of Tracee.New() but it will fail because it depends on `t.config.metricsEnabled` which is enabled by the new introduced callback IF `Server.metricsendpoint` is also enabled (and it is only enabled after `Tracee.New()` is called).

I already have a patch that fixes this, will propose a PR (checking if tracee-rules metrics are being enabled).

## Observations

This also brakes the `make -f builder/Makefile.performance pyroscope-start` dashboard (observed by Geyslan). So I'm checking what needs fixing (will need perf studies for event definitions).

PR #3304, opened 2023-07-07
fix(server): re-enable prometheus counters.

rafaeldtinoco:
Commit: 0213c08 ("tracee: add ready callback") made prometheus stats to stop being registered due to a timing issue. This patch re-enables it by registering prometheus counters in the callback function itself instead of in Tracee.New() (too early).

Fixes: #3303

> Still checking more stuff.


geyslan:
> I'll redo the panels created in [#3297](https://github.com/aquasecurity/tracee/pull/3297#issuecomment-1620795046) on top of this working PR and raise the dashboard related one.

Ready: https://github.com/rafaeldtinoco/tracee/pull/7

rafaeldtinoco:
Im adding an integration test to test if metrics are enabled and this will be ready.

rafaeldtinoco:
So, if the "test-performance" target does not run (it has a skip if the user can't execute anything using "sudo -n" (with no password)) then I'll have to run the performance tests in a jenkins runner or change the runner config a bit. Lets see.

tests/testutils/tracee.go
```diff
@@ -0,0 +1,154 @@
...
+type TraceeStatus int
+
+const (
+	TraceeStarted TraceeStatus = iota
+	TraceeFailed
+	TraceeTimedout
+	TraceeAlreadyRunning
+)
+
+// RunningTracee is a wrapper for a running tracee process as a regula process.
```
geyslan, line 37:
```suggestion
// RunningTracee is a wrapper for a running tracee process as a regular process.
```

geyslan (approved):
LGTM

Issue #3239, opened 2023-06-14
Support dynamic loading of policies


geyslan:
Context: https://github.com/aquasecurity/tracee/pull/3209#issuecomment-1591171759

- [x] #3469
  - [x] `populateBPFMaps()` `events_map` logic
  - [x] `initBPF()` `ProcessTreeFilter` logic
- [x] #3731
- [x] #3805
- [ ] Fixes
  - [ ] Snapshots Pruning and Inner Map creation rollback
    References:
    - https://github.com/aquasecurity/tracee/issues/4019#issuecomment-2118124368
    - #3912
  - [ ] Free GetBTFFDByID() resources - [Reference](https://github.com/aquasecurity/tracee/issues/4019#issuecomment-2118124368)
- [ ] Refactorings (required)
  - [x] https://github.com/aquasecurity/tracee/pull/3906
  - [x] https://github.com/aquasecurity/tracee/pull/3955
  - [x] https://github.com/aquasecurity/tracee/pull/3956
  - [x] https://github.com/aquasecurity/tracee/pull/3963
  - [x] https://github.com/aquasecurity/tracee/pull/3966
  - [x] https://github.com/aquasecurity/tracee/pull/3979
  - [x] https://github.com/aquasecurity/tracee/pull/3997
  - [ ] Matched Rules (instead of Policies - https://github.com/aquasecurity/tracee/pull/4484
  - [ ] PolicyManager API Refactor - https://github.com/aquasecurity/tracee/pull/4484
- [ ] ...

geyslan:
Open this [PR](https://github.com/aquasecurity/tracee/pull/3249) as a PoC for discussion. It raised deadlocks.

geyslan:
#3250 uses snapshots (circular buffer).

yanivagman:
@geyslan I thought about the following API to be added:

AddPolicy(Policy) error
UpdatePolicy(Policy) error
DeletePolicy(Policy) error
ListPolicies()

WDYT?

geyslan:
@yanivagman considering that this API should be to PolicyManager, perhaps:

DeletePolicy(name) error

instead of the policyObject as parameter.

yanivagman:
> @yanivagman considering that this API should be to PolicyManager, perhaps:
> 
> DeletePolicy(name) error
> 
> instead of the policyObject as parameter.

Yes, I agree.
I thought about the following API functions:
[...]

PR #3305, opened 2023-07-08
Introduce Policies versioning (map of maps)

### LINKED ISSUE #3469: Introduce policies versioning (map of maps)

geyslan:
Close #3469

d80c9a4ec **feat(userland): add support for policy versioning**

```
Add policy versioning, which will support runtime policy updates.

This also:

- Adds the singleton Snapshots circular buffer (for policy versioning).
- Introduces the Cloner interface (used for all policy objects).
- Moves the filters' ebpf logic to a new ebpf file (filter types are now
  ebpf-agnostic).
- Splits filters_test.go into multiple files.
- Removes the unused containers filter (it was just a wrapper for
  StringFilter).
- Removes unused RandStringRunes function from sets_test.go.
- Makes other minor refactorings/changes.
```

53dd343ea **feat(ebpf): add support for policy versioning**

```
Make use of map of maps to versionize policies. This is a first step
towards supporting runtime policy updates.
```

### 3. Other comments


go.mod
```diff
@@ -173,3 +172,6 @@ require (
 )
 
 replace github.com/kubernetes/cri-api => k8s.io/cri-api v0.23.5-rc.0
+
+replace github.com/aquasecurity/tracee/types => ./types
+
```
geyslan, line 177: This commit will be dropped when types change gets merged.

go.mod
```diff
@@ -175,3 +175,4 @@ replace github.com/kubernetes/cri-api => k8s.io/cri-api v0.23.5-rc.0
 
 replace github.com/aquasecurity/tracee/types => ./types
 
+replace github.com/aquasecurity/libbpfgo => ../libbpfgo
```
geyslan, line 178: This commit will be dropped when libbpfgo bump (with map of maps) gets merged.

geyslan: This depends on https://github.com/aquasecurity/libbpfgo/pull/356 and https://github.com/aquasecurity/libbpfgo/pull/366.

geyslan: https://github.com/aquasecurity/libbpfgo/pull/356 was merged. It's now depending only on https://github.com/aquasecurity/libbpfgo/pull/366 merge.

geyslan: All merged. Raising a change to get the upstream libbpfgo.

.github/workflows/pr.yaml
```diff
@@ -153,6 +153,10 @@ jobs:
       - verify-analyze-code
     runs-on: ubuntu-latest
     steps:
+      - name: Discover init
+        run: |
+          ps -p 1 -o comm=
+          stat /sbin/init
```
geyslan, line 159: Trying understand what can be causing: https://github.com/aquasecurity/tracee/actions/runs/6074223465/job/16479631240?pr=3305#step:4:411

geyslan:
@yanivagman e2e 1056 failed due to required type changes. github workflow is green.

pkg/policy/snapshots.go
```diff
@@ -0,0 +1,86 @@
...
+
+func (ps *Policies) StoreSnapshot() error {
+	snapshots.murw.Lock()
+	defer snapshots.murw.Unlock()
+
+	snapshots.lastVersion++ // new version
+
+	ps.rwmu.Lock()
+	ps.version = snapshots.lastVersion
+	ps.rwmu.Unlock()
```
yanivagman, line 42: Here you protect version with rwmu but in policies.go you do:
```
func (ps *Policies) Version() uint32 {
	return atomic.LoadUint32(&ps.version)
}
```
I would say you should protect it in one way or the other, and maybe you can use atomic integer instead of these protections

pkg/policy/policies.go
```diff
@@ -213,6 +268,489 @@ func (ps *Policies) FilterableInUserlandMap() map[*Policy]int {
...
+	// populate proc info map with binaries to track
+	for bin := range binEqualities {
+		procs := binsProcs[bin.Path]
+		for _, p := range procs {
+			binBytes := make([]byte, maxBpfBinPathSize)
+			copy(binBytes, bin.Path)
+			binBytesCopy := (*[maxBpfBinPathSize]byte)(binBytes)
+			procInfo := procInfo{
+				newProc:        false,
+				followPolicies: 0,
```
yanivagman, line 682: These assumption of default values are no longer correct if we update the policies dynamically.
Since it was assumed we call this function on init, it was safe to say that this is not a newProc (started after tracee) and that not match occured (thus follow can be set to false for this process).

We can comment it here as a todo and handle later

pkg/policy/policies.go
```diff
@@ -132,11 +149,49 @@ func (ps *Policies) set(id int, p *Policy) error {
...
 
-	ps.Compute()
+	ps.compute()
 
 	return nil
 }
 
+// TODO: runtime API should encapsulate the following calls:
+//       1. policies.DeepCopy() to get a clone before to apply changes
+// 	     2. policies.StoreSnapshot() to get the new version snapshot stored
```
yanivagman, line 159: nit: indentation

pkg/policy/policies.go
```diff
@@ -213,6 +268,489 @@ func (ps *Policies) FilterableInUserlandMap() map[*Policy]int {
...
+		opts := &bpf.BPFMapCreateOpts{
+			BTFFD:                 uint32(btfFD),
+			BTFKeyTypeID:          info.BTFKeyTypeID,
+			BTFValueTypeID:        info.BTFValueTypeID,
+			BTFVmlinuxValueTypeID: info.BTFVmlinuxValueTypeID,
+			MapFlags:              info.MapFlags,
+			MapExtra:              info.MapExtra,
+			MapIfIndex:            info.IfIndex,
+		}
+		newInnerMap, err := bpf.CreateMap(
```
yanivagman, line 487: Do we also delete the map at some point? I don't think I saw it in the code

pkg/policy/policies.go
```diff
@@ -132,11 +149,49 @@ func (ps *Policies) set(id int, p *Policy) error {
 	ps.policiesArray[id] = p
 	ps.filterEnabledPoliciesMap[p] = id
 
-	ps.Compute()
+	ps.compute()
 
 	return nil
 }
 
+// TODO: runtime API should encapsulate the following calls:
```
yanivagman, line 157: This API call will be protected by the policies mutex, right?

pkg/policy/policies.go
```diff
@@ -213,6 +268,489 @@ func (ps *Policies) FilterableInUserlandMap() map[*Policy]int {
...
+			if err := procInfoMap.Update(unsafe.Pointer(&p), unsafe.Pointer(&procInfo)); err != nil {
+				return errfmt.WrapError(err)
+			}
+		}
+	}
+
+	return nil
+}
+
+func (ps *Policies) UpdateBPF(bpfModule *bpf.Module, cts *containers.Containers) error {
```
yanivagman, line 696: Is it the responsibility of the caller to have the policies mutex locked?

yanivagman (changes requested):
Nice work @geyslan !

pkg/bufferdecoder/decoder_test.go
```diff
@@ -13,23 +13,24 @@ import (
...
+		HostPpid:        555,
+		Uid:             9876,
+		MntID:           1357,
+		PidID:           3758,
+		Comm:            [16]byte{1, 3, 5, 3, 1, 5, 56, 6, 7, 32, 2, 4},
+		UtsName:         [16]byte{5, 6, 7, 8, 9, 4, 3, 2},
+		EventID:         0,
+		PoliciesVersion: 1,
+		Retval:          0,
+		StackID:         0,
```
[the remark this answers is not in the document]
geyslan: It's missing indeed. Adding it.

pkg/bufferdecoder/protocol.go
```diff
@@ -48,7 +50,7 @@ type Context struct {
 }
 
 func (Context) GetSizeBytes() int {
-	return 128
+	return 136
```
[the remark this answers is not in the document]
geyslan: I agree, going to reuse the padding space.

pkg/cmd/cobra/cobra.go
```diff
@@ -162,6 +162,9 @@ func GetTraceeRunner(c *cobra.Command, version string) (cmd.Runner, error) {
 	}
 
 	cfg.Policies = policies
+	if err := cfg.Policies.StoreSnapshot(); err != nil {
+		return runner, err
+	}
```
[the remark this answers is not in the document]
geyslan: I put it just after the policies assignment to state that every policies creation/change must be considered as a snapshot. But I gonna think about it.

pkg/ebpf/tracee.go
```diff
@@ -248,12 +250,16 @@ func New(cfg config.Config) (*Tracee, error) {
 
 	// Handle all essential events dependencies
 
+	// TODO: extract this to a function to be called from here and from
+	// policies changes.
 	for id, evt := range t.eventsState {
```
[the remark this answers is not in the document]
geyslan: @yanivagman right above we iterate the policies for events to trace and update their state.

So I ask @rafaeldtinoco if we must take care of it from Policies or from Events?

pkg/filters/string.go
```diff
@@ -220,78 +217,61 @@ func (f *StringFilter) FilterOut() bool {
...
+	maps.Copy(n.notContains, f.notContains)
+	n.enabled = f.enabled
+
+	return n
+}
+
 type BPFStringFilter struct {
 	StringFilter
-	mapName string
 }
```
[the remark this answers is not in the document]
geyslan: > Is there a reason to keep the BPFStringFilter if we remove the mapName from it?

No there's not.

> We can use StringFilter instead, no?

Yeah, I had pondered about this and forgot to put TODO comments, since the BPFFilters won't be for BPF anymore. They will only provide the equality values.

pkg/ebpf/tracee.go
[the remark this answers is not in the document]
yanivagman: Actually you might be right.
Event state is derived from policy state so should be probably moved to the policy manager

pkg/policy/snapshots.go
```diff
@@ -0,0 +1,86 @@
...
+
+func GetLastSnapshot() (*Policies, error) {
+	snapshots.murw.RLock()
+	defer snapshots.murw.RUnlock()
+
+	if snapshots.storedCnt == 0 {
+		return nil, errfmt.Errorf("no snapshots stored")
+	}
+
+	return snapshots.snaps[snapshots.storedCnt-1].policies, nil
```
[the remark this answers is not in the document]
geyslan: Yep. Nice catch. And `GetSnapshot()` requires fixing. I opted to introduce a new holder `lastIdx`, so it would clarify and avoid index recalculation.

pkg/policy/snapshots.go
geyslan: Indeed. I'm going to use atomic for that.

pkg/policy/policies.go
geyslan: No, it's missing here. Fixing.

pkg/policy/policies.go
geyslan: No, currently we lack that. It should be tackled through the version expiration.

pkg/policy/policies.go
geyslan: Yes, it will. @josedonizetti those are the current steps available in this PR for the runtime API.

pkg/ebpf/tracee.go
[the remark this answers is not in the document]
geyslan: @josedonizetti FIY.

pkg/filters/string.go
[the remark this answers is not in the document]
geyslan: `// TODO:` added.

pkg/bufferdecoder/protocol.go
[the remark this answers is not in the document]
geyslan: Reordered and it fit.

pkg/policy/snapshots.go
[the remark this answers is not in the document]
geyslan: Fixed. Introduced `snapshots_test.go`.

pkg/policy/snapshots.go
geyslan: Fixed.

pkg/cmd/cobra/cobra.go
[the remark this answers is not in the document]
geyslan: This will be revisited, since we have plans to move Policies into PolicyManager.

pkg/policy/policies.go
geyslan: As we have discussed, we're going to move all Events and Policies into PolicyManager.

geyslan:
> Nice work @geyslan !

Thanks for the review (first pass). I've already addressed almost everything and postponed the rest for later tasks:

- `// TODO`.
- Didn't touch the proctree logic as we have a great PR in the oven.
- Didn't move policies/eventstate logic to policyManager, waiting ongoing logic merge.

pkg/policy/policies.go
```diff
@@ -224,6 +286,572 @@ func (ps *Policies) FilterableInUserlandMap() map[*Policy]int {
...
+
+		// store pointer to the new inner map version
+		ps.bpfInnerMaps[innerMapName] = newInnerMap
+	}
+
+	return nil
+}
+
+func (ps *Policies) createNewEventsMapVersion(bpfModule *bpf.Module, eventsState map[events.ID]events.EventState) error {
+	// Initialize events parameter types map
```
yanivagman, line 570: Events parameters should be constant and I don't see a reason for it to be changed between policy updates - maybe move to init so we won't repeat this logic on every policy update?

pkg/ebpf/c/common/buffer.h
```diff
@@ -351,6 +351,12 @@ statfunc int events_perf_submit(program_data_t *p, u32 id, long ret)
     u32 size = sizeof(event_context_t) + sizeof(u8) +
                p->event->args_buf.offset; // context + argnum + arg buffer size
 
+    // Check, at the last moment, if the policies version changed during the event processing.
+    // If so, skip the event submission as it is considered stale.
```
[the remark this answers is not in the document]
geyslan: Previous snapshots would only work for buffered events already sent in an end-to-end policy version window. It turns out that as the event is composed of different steps/programs, if the version is different, the rules that correspond to that event may be wrong in sequential steps/programs, so I think that in the final phase if the version is different, it is not safe to use this event. In other words: it was matched/composed in more than one rule windows.

pkg/ebpf/events_pipeline.go
```diff
@@ -576,8 +590,13 @@ func (t *Tracee) sinkEvents(ctx context.Context, in <-chan *trace.Event) <-chan
 				continue
 			}
 
+			policies, err := policy.GetSnapshot(event.PoliciesVersion)
```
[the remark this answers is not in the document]
geyslan: My first attempts were using the policies pointer:

https://github.com/aquasecurity/tracee/pull/3250/files#diff-e74de87a065b53a5a49ea37519b1c19a99f64fa311aef231e728e823a591023cR38

This broke because it has a set of fewer allowed types. To get around this, I replaced unsafe.Pointer with a uint64 value... but this causes panic when the GC releases this pointer. One solution would be to have a dictionary to save this pipeline in use pointer (keeping it hot/referenced) and get it again in the next steps of reading this dictionary. But I assume (not benchmarked) that, in the vast majority of cases, circular buffering will perform better or at worst similar to golang map retrieval (hashing, internal bucket lookup, check return if found, etc). WDYT?

pkg/ebpf/c/common/buffer.h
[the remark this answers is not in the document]
geyslan: Rethinking it, if we keep it hot in a go map, the uit64 field with the right policies version address would be coerced during the whole pipeline instead of re-retrieved.

It would work unlike the event pointer pool which stores unused pointers, this map would store used policy pointers to prevent them from being freed. It might work. LMKWYT.

pkg/ebpf/events_pipeline.go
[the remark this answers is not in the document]
geyslan: Rethinking it, if we keep it hot in a go map, the uit64 field with the right policies version address would be coerced during the whole pipeline instead of re-retrieved.

It would work unlike the event pointer pool which stores unused pointers, this map would store used policy pointers to prevent them from being freed. It might work. LMKWYT.

pkg/policy/policies.go
geyslan: Yep, they're constant indeed. We need here only to tackle eventStates. Going to move eventsParams to initialization.

.github/workflows/pr.yaml
geyslan: This will be handled via https://github.com/aquasecurity/tracee/issues/3440

go.mod
```diff
@@ -5,8 +5,8 @@ go 1.19
 require (
 	github.com/IBM/fluent-forward-go v0.2.1
 	github.com/Masterminds/sprig/v3 v3.2.3
-	github.com/aquasecurity/libbpfgo v0.4.8-libbpf-1.2.0.0.20230509162948-80f41e18e690
-	github.com/aquasecurity/libbpfgo/helpers v0.4.6-0.20230321190037-f591a2c5734f
+	github.com/aquasecurity/libbpfgo v0.5.0-libbpf-1.2
```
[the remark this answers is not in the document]
geyslan: Done https://github.com/aquasecurity/tracee/pull/3482.

pkg/ebpf/c/common/buffer.h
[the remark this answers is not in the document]
geyslan: As we discuss and decide offline, we will always submit the event and leave it to the user to fetch the corresponding policies version object.

pkg/ebpf/c/common/filtering.h
```diff
@@ -189,6 +217,10 @@ statfunc u64 compute_scopes(program_data_t *p)
         return 0;
     }
 
+    //
+    // unversioned filters
```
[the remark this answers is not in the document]
geyslan: We've remembered why bool filters are not versioned: they are not stored in specific maps, they are part of the config.

pkg/ebpf/events_pipeline.go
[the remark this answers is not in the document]
geyslan: We decided to return to this optimisation in the future.

pkg/policy/policy.go
```diff
@@ -1,21 +1,35 @@
...
+	PidNSFilterMap    = "pid_ns_filter"
+	UTSFilterMap      = "uts_ns_filter"
+	CommFilterMap     = "comm_filter"
+	CgroupIdFilterMap = "cgroup_id_filter"
+	BinaryFilterMap   = "binary_filter"
+
+	// TODO: ProcessTreeFilterMap must be versioned
 	ProcessTreeFilterMap = "process_tree_map"
-	CgroupIdFilterMap    = "cgroup_id_filter"
-	ContIdFilter         = "cont_id_filter"
```
geyslan, line 17: `ContIdFilter` was an unused const.

pkg/policy/policies.go
```diff
@@ -1,40 +1,57 @@
...
 	"github.com/aquasecurity/tracee/pkg/events"
 	"github.com/aquasecurity/tracee/pkg/filters"
 	"github.com/aquasecurity/tracee/pkg/utils"
+	"github.com/aquasecurity/tracee/pkg/utils/proc"
 )
 
 const (
-	allPoliciesOn = 0xFFFFFFFFFFFFFFFF
+	MaxPolicies   int    = 64
+	allPoliciesOn uint64 = ^uint64(0)
```
geyslan, line 23: Using `^uint64(0)` to avoid wrong assignment (missing Fs).

rafaeldtinoco:
I want to review this one before its merged.

geyslan:
It is green on E2E 1166. ❤️

geyslan:
Ready for review.

geyslan:
Conflicts solved. Ready for review again.

pkg/events/derive/derive.go
```diff
@@ -42,20 +42,23 @@ func (t Table) Register(deriveFrom, deriveTo events.ID, deriveCondition func() b
...
 	for id, deriveFn := range deriveFns {
 		if deriveFn.Enabled() {
-			derivative, errs := deriveFn.DeriveFunction(event)
+			derivs, errs := deriveFn.DeriveFunction(event)
 			for _, err := range errs {
-				errors = append(errors, deriveError(id, err))
+				deriveErrors = append(deriveErrors, deriveError(id, err))
+			}
+			for _, deriv := range derivs {
+				deriv.PoliciesVersion = event.PoliciesVersion
```
josedonizetti, line 55: I think this should be done at [`buildDerivedEvent`](https://github.com/aquasecurity/tracee/blob/main/pkg/events/derive/derive.go#L130)

pkg/events/derive/derive.go
```diff
@@ -42,20 +42,23 @@ func (t Table) Register(deriveFrom, deriveTo events.ID, deriveCondition func() b
...
 func (t Table) DeriveEvent(event trace.Event) ([]trace.Event, []error) {
-	derivatives := []trace.Event{}
-	errors := []error{}
+	derivedEvents := []trace.Event{}
+	deriveErrors := []error{}
 	deriveFns := t[events.ID(event.EventID)]
 	for id, deriveFn := range deriveFns {
 		if deriveFn.Enabled() {
-			derivative, errs := deriveFn.DeriveFunction(event)
+			derivs, errs := deriveFn.DeriveFunction(event)
```
josedonizetti, line 50: I think `derivatives` is a better name, no?

common/interfaces.go
```diff
@@ -0,0 +1,6 @@
+package common
+
+// Cloneable is an interface for objects that can be cloned
+type Cloneable interface {
```
josedonizetti, line 4: `Cloneable` seems too javaish, more golang style would be `Cloner` no? But I think this interface is too generic but used for a very specific purpose, Filters, I think we shove move it into the filter package `FilterCloner`. It seems out place for all filters to return a `common.Cloneable` since they are all filters. Or maybe we could have the method inside the `Filter` interface, and just return Filter?

pkg/ebpf/c/common/filtering.h
```diff
@@ -189,6 +203,10 @@ statfunc u64 compute_scopes(program_data_t *p)
...
```
josedonizetti, line 207: This is confusing for me. Is unversioned filters that aren't part of the `pols_version`? You did explain to me off line that it isn't the case, so perhaps improve the comments here?

pkg/ebpf/tracee.go
```diff
@@ -1691,6 +1677,7 @@ func (t *Tracee) triggerMemDump(event trace.Event) []error {
 
 	errs := []error{}
 
+	// TODO: consider to iterate over given policies when policies are changed
```
josedonizetti, line 1680: It seems to me you for now didn't change `t.config.Policies` for the initialization, and are planning to do this change later, you only using `policy.Snapshots()` in dynamic places for this pr, which is okay for me, though, I think the methods to `Subscribe`, `Unsubscribe`, etc from the GRPC API aren't in the initialization category, and must be updated https://github.com/aquasecurity/tracee/blob/main/pkg/ebpf/tracee.go#L1857-L1946

pkg/policy/snapshots.go
```diff
@@ -0,0 +1,192 @@
...
+// PruneSnapshotsOlderThan prunes snapshots older than a given duration.
+func (s *snapshots) PruneSnapshotsOlderThan(d time.Duration) []error {
+	s.murw.Lock()
+	defer s.murw.Unlock()
+
+	if s.storedCnt <= 1 {
+		return nil
+	}
+
+	if s.prune == nil {
```
josedonizetti, line 151: This is a bit odd to me, because it seems the prune function is mandatory, though one can create a snapshot without it, it is optional to use the `Set` method to pass it or not, if the func is mandatory it should be required by the constructor, one should not be able to create a snapshot without it. Wdyt?

josedonizetti (commented):
@geyslan  Nice work! I've added a few comments. I'll keep reading it, as I will try to implement a few of the TODOs to help here, so if I think of anything else I'll add other comments, but this looks really cool!

rafaeldtinoco:
Ill review this soon as well!

pkg/events/derive/derive.go
geyslan: Yeah, we can revert those names.

pkg/events/derive/derive.go
geyslan: Ok, as long as it is a function accessible via policyManager.

common/interfaces.go
geyslan: I agree with the nomenclature, it should be `Cloner`.

> But I think this interface is too generic but used for a very specific purpose, Filters, I think we shove move it into the filter package FilterCloner. It seems out place for all filters to return a common.Cloneable since they are all filters. Or maybe we could have the method inside the Filter interface, and just return Filter

It's not just Filters that implement this interface, `Policy` and `Policies` do so:

image

pkg/ebpf/c/common/filtering.h
geyslan: Let's change those comments for something like: `// equality filters (using versioned maps)` and `boolean filters (not using versioned maps)`

pkg/policy/snapshots.go
geyslan: `SetPruneFunc()` is temporary and deserves a comment about it. Good point!

As the [periodic control (go routine with Ticker)](https://github.com/aquasecurity/tracee/pull/3305/files/0cbe2ab53a11b05a9fd1ce0acbcc2e6146df7047#diff-5938b4408c6b7a00649735759bab2c23686f6275834942ee729903bc77dfbe5aR141-R143) to Call `PruneSnapshotsOlderThan()` has not yet been implemented, `SetPruneFunc()` and the checks regarding the lack of prune were inserted for test purpose. So you're right, we should make the pruner func mandatory, setting it on the snapshots constructor.

geyslan:
> @geyslan Nice work! I've added a few comments. I'll keep reading it, as I will try to implement a few of the TODOs to help here, so if I think of anything else I'll add other comments, but this looks really cool!

@josedonizetti Thanks a million for reviewing this. I'll solve the recent conflict to get it ready again for any help that may come. ❤️

pkg/ebpf/c/common/filtering.h
geyslan: @josedonizetti this is solved by: https://github.com/aquasecurity/tracee/pull/3735/commits/ec9c7e5718f46f135c05209341d5be380570be24

geyslan:
It's ready for review again.

pkg/policy/policies.go
```diff
@@ -1,36 +1,55 @@
...
-	uidFilterMin            uint64
-	uidFilterMax            uint64
-	pidFilterMin            uint64
-	pidFilterMax            uint64
-	uidFilterableInUserland bool
-	pidFilterableInUserland bool
-
-	filterableInUserland    uint64 // bitmap of policies that must be filtered in userland
-	containerFiltersEnabled uint64 // bitmap of policies that have at least one container filter type enabled
+	rwmu sync.RWMutex
```
rafaeldtinoco, line 29: Ok, either you do a global lock or a fine grained lock. You want to protect a deep copy of the policies array, which is fine, but you leave other parts unprotected (Count() for example is not thread safe). I would make this entire type thread-safe or thread-unsafe. Currently its half-half.

pkg/cmd/cobra/cobra.go
```diff
@@ -232,6 +232,7 @@ func GetTraceeRunner(c *cobra.Command, version string) (cmd.Runner, error) {
 	}
 
 	cfg.Policies = policies
+	policy.Snapshots().Store(cfg.Policies)
```
rafaeldtinoco, line 235: This is a very good singleton practice/way (instead of saving instanced objects in Tracee singleton). Im just afraid that there should be an effort to make all others like this one somehow. There are many instanciated objects that could use the same approach.

rafaeldtinoco:
Im literally missing only the eBPF filters portion (file) now, will finish tomorrow.

Things I like from the PR so far:

- the singleton initialization (Once)
- multiple tests added to each filter type
- filter eBPF logic split from regular filter logic

Left some questions/comments.

pkg/policy/policies.go
geyslan: Yep. When I put that I was concerned only with the cloning as Policies object is immutable, considering the upcoming versioning, but you pinpointed correctly: we have public methods reading/writing to inner values. Gonna make it full thread-safe.

pkg/cmd/cobra/cobra.go
geyslan: Nice. We can expand this pattern in the future.

pkg/policy/ebpf.go
```diff
@@ -0,0 +1,726 @@
...
+		uidEqualities:      make(map[uint64]equality),
+		pidEqualities:      make(map[uint64]equality),
+		mntNSEqualities:    make(map[uint64]equality),
+		pidNSEqualities:    make(map[uint64]equality),
+		cgroupIdEqualities: make(map[uint64]equality),
+		utsEqualities:      make(map[string]equality),
+		commEqualities:     make(map[string]equality),
+		procTreeEqualities: make(map[uint64]equality),
+		binaryEqualities:   make(map[filters.NSBinary]equality),
+	}
```
rafaeldtinoco, line 673: You could document all this much better. What this function does, when is it called, why. Also, the maps. Index is UID, then PID then MNTNS, etc. You could have a description describing equality (meaning in what policy that is enabled, and in which one is is a "!=" value). Etc. This code is way too obscure for an open source project (not because of you). It is very hard to any 3rd party contributor to understand what goes on in this part.

pkg/policy/ebpf.go
```diff
@@ -0,0 +1,726 @@
...
+		return errfmt.WrapError(err)
+	}
+
+	return nil
+}
+
+func (ps *Policies) createNewFilterMapsVersion(bpfModule *bpf.Module) error {
+	mapsNames := map[string]string{ // inner map name: outer map name
+		UIDFilterMap:         UIDFilterMapVersion,
+		PIDFilterMap:         PIDFilterMapVersion,
```
rafaeldtinoco, line 315: Again, this should be describing the outer/inner relation, why you're creating filter maps and what they mean.

pkg/policy/ebpf.go
```diff
@@ -0,0 +1,726 @@
...
+	pidNSEqualities    map[uint64]equality
+	cgroupIdEqualities map[uint64]equality
+	utsEqualities      map[string]equality
+	commEqualities     map[string]equality
+	procTreeEqualities map[uint64]equality
+	binaryEqualities   map[filters.NSBinary]equality
+}
+
+// TODO: refactor this method deduplicating code
+func (ps *Policies) computeFilterEqualities(fEqs *filtersEqualities, cts *containers.Containers) error {
```
rafaeldtinoco, line 73: Needs documentation. Very obscure for new comers to the project. Consider 2 or 3 new engineers arriving into the project and having to read/maintain this. This should be the mindset when documenting (IMO).

pkg/policy/ebpf.go
rafaeldtinoco: image

image

After sometime not visiting the policies filtering logic, I could only understand/remember taking a look at some old notes of my own. This mean we could do the same, but in a structured way as documentation.

rafaeldtinoco:
@geyslan Im done with my review. I cannot find any flaw (have not tested it, just read and made logic judgement). I believe it touches sensitive filtering logic, so I would make sure to do lots of tests before merging (not only unit testing, but making sure each policy has a specific filter, or set of filters, including == and !=s and making sure at least 5 consecutive version changes are good, things like that).

Either way, it was a very nice work of you, as usual. Congrasts!

rafaeldtinoco (approved):
LGTM after concurrency issues are addressed, documentation is improved and some of my comments are taken into consideration. I would recommend some testing like said before. Other than that, it's a very nice work.

common/interfaces.go
```diff
@@ -0,0 +1,6 @@
+package common
```
NDStrahilevitz, line 1: I think `utils` already serves as this.

pkg/events/derive/derive.go
```diff
@@ -132,6 +132,8 @@ func buildDerivedEvent(baseEvent *trace.Event, skeleton deriveBase, argsValues [
 		return trace.Event{}, unexpectedArgCountError(skeleton.Name, len(skeleton.Params), len(argsValues))
 	}
 	de := *baseEvent
+	// TODO: this logic must be accessible from the policy manager
```
NDStrahilevitz, line 135: Which logic? The deriving from a base? Asking so that the comment's instructions can be more clear.

pkg/ebpf/tracee.go
```diff
@@ -815,7 +832,7 @@ func (t *Tracee) getOptionsConfig() uint32 {
 	return cOptVal
 }
 
-func (t *Tracee) computeConfigValues() []byte {
+func (t *Tracee) computeConfigValues(newPolicies *policy.Policies) []byte {
```
NDStrahilevitz, line 835: I think this function could (and should) be moved to the policy snapshots.

pkg/ebpf/tracee.go
```diff
@@ -1229,6 +1185,37 @@ func (t *Tracee) populateBPFMaps() error {
...
+		updateProcTree,
+	); err != nil {
+		return errfmt.WrapError(err)
+	}
+
+	// Compute config map values based on new policies
+
+	configVal := t.computeConfigValues(newPolicies)
+
+	bpfConfigMap, err := t.bpfModule.GetMap("config_map")
```
NDStrahilevitz, line 1207: and then the logic under this moved to the upper `UpdateBPF` call.

NDStrahilevitz (commented):
Really nice work. I hope what I've written below will be useful.

pkg/ebpf/c/common/filtering.h
```diff
@@ -209,80 +227,102 @@ statfunc u64 compute_scopes(program_data_t *p)
...
     if (p->config->pid_filter_enabled_scopes) {
         u64 filter_out_scopes = p->config->pid_filter_out_scopes;
         u64 mask = ~p->config->pid_filter_enabled_scopes;
         u64 max = p->config->pid_max;
         u64 min = p->config->pid_min;
         // the user might have given us a tid - check for it too
         res &=
-            uint_filter_range_matches(filter_out_scopes, &pid_filter, context->host_pid, max, min) |
-            uint_filter_range_matches(filter_out_scopes, &pid_filter, context->host_tid, max, min) |
+            uint_filter_range_matches(
```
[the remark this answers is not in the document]
geyslan: I agree with you, however I suspect we have plans to bring range filter entirely into bpf, then we could tackle that optimisation.

pkg/policy/snapshots.go
```diff
@@ -0,0 +1,194 @@
...
+	version  uint16
+	policies *Policies
+}
+
+// snapshots is a circular buffer of snapshots.
+type snapshots struct {
+	murw        sync.RWMutex
+	lastVersion uint16
+	snaps       [maxSnapshots]*snapshot
+	nextIdx     int
```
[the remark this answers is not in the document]
geyslan: `lastVersion` size is just aligned to the ebpf part. `lastIdx` by instance also signals no spashots stored `-1`. Moreover snapshots object is a singleton.

common/interfaces.go
geyslan: Yeah, I initially thought as you. But I'm just trying to open a common place to have code that's really common to all. `utils` is kinda "we don't know where to put this" option.

pkg/events/derive/derive.go
geyslan: Gonna improve this.

pkg/ebpf/tracee.go
geyslan: Well, the problem is this would reinject the bpf logic into the Policies/Snapshots types, which is what we're trying to avoid: coupling things together.

geyslan:
While addressing review comments I detected that versioned maps are being generated in duplicity. Gonna check this asap after finish the ongoing changes.

```
❯ sudo bpftool map list | grep _1 
340: hash  name events_map_1  flags 0x0
341: hash  name pid_filter_1  flags 0x0
342: hash  name mnt_ns_filter_1  flags 0x0
343: hash  name uts_ns_filter_1  flags 0x0
344: hash  name comm_filter_1  flags 0x0
347: hash  name binary_filter_1  flags 0x0
348: hash  name uid_filter_1  flags 0x0
349: hash  name pid_ns_filter_1  flags 0x0
350: hash  name events_map_1  flags 0x0
351: hash  name uid_filter_1  flags 0x0
352: hash  name pid_ns_filter_1  flags 0x0
353: hash  name comm_filter_1  flags 0x0
355: hash  name binary_filter_1  flags 0x0
356: hash  name pid_filter_1  flags 0x0
357: hash  name mnt_ns_filter_1  flags 0x0
358: hash  name uts_ns_filter_1  flags 0x0
```

pkg/ebpf/tracee.go
NDStrahilevitz: I think that the existence of `UpdateBPF` in snapshots show that the snapshot object is the coupling point. That's not a bad thing, there must be some point of coupling if we want to make policies meaningful in bpf logic.
Maybe, the `UpdateBPF` function should be taken out of the snapshots object so that the coupling will be out of the object and into some "effect function".

common/interfaces.go
NDStrahilevitz: It doesn't seem like we use this interface for anything other than cloning interface, so it's not common and there's no guarantee it will be. Furthermore, will there be some kind of cloner which returns a different cloner? 
If anything, `Clone` could have been part of the filter interface with the method `Clone() Filter`.

pkg/ebpf/c/common/filtering.h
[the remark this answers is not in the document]
geyslan: Well, this simplifies it based on your idea: https://github.com/aquasecurity/tracee/pull/3305/commits/80bdf06b290b84da91067557ecb824a11d812703

See the chagned file itself for better reading: https://github.com/aquasecurity/tracee/pull/3305/files#diff-c046893996f2e2a6e44a8caac7a14abbf47b47486ee5ceaa9029c34fc5f94a99

common/interfaces.go
geyslan: I'm ok in placing it in `utils`. About being Cloners, we have `Policy` and `Policies` besides `Filter`, take a look:

https://github.com/aquasecurity/tracee/pull/3305#discussion_r1411378820

pkg/ebpf/tracee.go
geyslan: @NDStrahilevitz after analyse this better I agree with you. We really need to split logic and diminish the tracee file. However, the complexity involved is also huge. For example, the `computeConfigValues` takes care of "config_map" which has a lot from policies but it's a tracee generic config place. And splitting or moving its logic can compromise the configVal `config_entry` encode understanding. What I propose now is to update this [EPIC](https://github.com/aquasecurity/tracee/issues/3239) with next steps, so we can tackle then meanwhile.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -981,6 +989,7 @@ int uprobe_lkm_seeker_submitter(struct pt_regs *ctx)
 
     // Uprobes are not triggered by syscalls, so we need to override the false value.
     p.event->context.syscall = NO_SYSCALL;
+    p.event->context.policies_version = p.config->policies_version;
```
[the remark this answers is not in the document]
geyslan: Regarding this, I'll wait @OriGlassman @AlonZivony feedback.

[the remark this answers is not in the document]
geyslan: > Because note that there can be a policy version change in the time between the uprobe trigger and the ebpf program starting.

I had put a check on the last stage of the event submission, checking the event context version against the current one from config_entry; if different discard the event. We had discussions about that and agreed that it would be better to send the event even with wrong version. But we could reintroduce it again to set the version as 0, so userland would know that it's a rogue event.

geyslan:
Moving it to [v0.21.0](https://github.com/aquasecurity/tracee/milestone/38).

pkg/ebpf/c/tracee.bpf.c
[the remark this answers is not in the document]
NDStrahilevitz: > checking the event context version against the current one from config_entry;
But even the event context version can be later than the one that should be injected from the trigger. So that kind of corruption would be missed. This isn't a huge problem for our trigger events right now since they are infrequent, but if we had events which are more frequent, it may be more noticable.
I'd at least document the limitation, it may not be worth fixing (I think eBPF cookies in later kernels could be used to solve this for example).

yanivagman (commented):
This is a very nice work Geyslan!
A few comments ahead and we can merge this soon

pkg/ebpf/c/maps.h
```diff
@@ -78,17 +97,30 @@ enum tail_call_id_e
...
+BPF_HASH_INNER(comm_filter, string_filter_t, eq_t, 256);           // filter events by command name prototype
+BPF_HASH_OUTER(comm_filter_version, comm_filter, 64);              // map of command name filters maps
+BPF_HASH_INNER(cgroup_id_filter, u32, eq_t, 256);                  // filter events by cgroup id prototype
+BPF_HASH_OUTER(cgroup_id_filter_version, cgroup_id_filter, 64);    // map of cgroup id filters maps
+BPF_HASH_INNER(binary_filter, binary_t, eq_t, 256);                // filter events by binary path and mount namespace prototype
+BPF_HASH_OUTER(binary_filter_version, binary_filter, 64);          // map of binary filters maps
+BPF_HASH_INNER(process_tree_map, u32, eq_t, 10240);                // filter events by the ancestry of the traced process
+BPF_HASH_OUTER(process_tree_map_version, process_tree_map, 64);    // map of process tree maps
+BPF_HASH_INNER(events_map, u32, event_config_t, MAX_EVENT_ID);     // map to persist event configuration data
+BPF_HASH_OUTER(events_map_version, events_map, 64);                // map of events maps
```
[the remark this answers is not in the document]
geyslan: It's according the new map creation/insertion. Rafael also noticed this size. It was put as 64 just to align the max number of policies, however, we can reduce it to 4-8. I assume it would be unlikely to have to keep more than that, even if we compute new policies very quickly.

pkg/ebpf/c/common/network.h
```diff
@@ -153,8 +153,10 @@ cgrpctxmap_t cgrpctxmap_eg SEC(".maps");    // saved info SKB caller <=> SKB egr
 typedef struct net_task_context {
     struct task_struct *task;
     task_context_t taskctx;
+    s32 syscall;
```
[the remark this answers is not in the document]
geyslan: IIRC due to respective userland EventContext field:

https://github.com/aquasecurity/tracee/blob/c6d88a038edd2ba0f8da09497d19892c31c26780/pkg/bufferdecoder/protocol.go#L47

pkg/ebpf/tracee.go
geyslan: No, this won't break Subscribe since we're not changing policies in userland yet and the assignment of the Policies is done in:

https://github.com/aquasecurity/tracee/pull/3305/files#diff-93e56b073f4afff2e526135681adfab063a5ef036eb2735daa066b9bb9ec242bR234-R235

pkg/ebpf/c/common/filtering.h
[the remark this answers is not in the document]
geyslan: I already added it as a task in the [EPIC](https://github.com/aquasecurity/tracee/issues/3239). Gonna create the issue and put the comment.

pkg/ebpf/c/maps.h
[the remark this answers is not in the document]
geyslan: Anyway, it would keep up to 64 references to the respective filter map:

```
❯ sudo bpftool map show name events_map_vers
78: hash_of_maps  name events_map_vers  flags 0x0
        key 2B  value 4B  max_entries 64  memlock 9280B
        pids tracee(51042)
```

Changing max_entries to 2 only reduced memlock by ~53%. The usage is minimal.

```
❯ sudo bpftool map show name events_map_vers
188: hash_of_maps  name events_map_vers  flags 0x0
        key 2B  value 4B  max_entries 2  memlock 4320B
        pids tracee(53850)
```

yanivagman (approved):
This is a very nice work Geyslan!
A few comments ahead and we can merge this soon

PR #3306, opened 2023-07-09
docs: Fix incorrect flags and usage documentation for unlinkat

kdrag0n:

unlinkat(2) does *not* accept any flags other than AT_REMOVEDIR. It cannot unlink files without following symlinks. This fixes the documentation accordingly.

### 2. Explain how to test it

N/A

### 3. Other comments

Man page: https://linux.die.net/man/2/unlinkat
Kernel implementation: https://github.com/torvalds/linux/blob/66d8fc0539b0d49941f313c9509a8384e4245ac1/fs/namei.c#L4433

I came across this documentation on Google and noticed that it was incorrectly generated by ChatGPT. It might be worth reconsidering this documentation strategy as potentially misleading content can get indexed.

### DIFF
--- a/docs/docs/events/builtin/syscalls/unlinkat.md
+++ b/docs/docs/events/builtin/syscalls/unlinkat.md
@@ -9,12 +9,10 @@ unlinkat is a system call that deletes a file name, relative to an open director
 
 Unlinkat also allows a user to delete links without traversing whole directory hierarchies, by providing the location of the file (directory) and its name as two separate arguments.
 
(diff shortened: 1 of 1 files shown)


rafaeldtinoco:
Yes, thanks @kdrag0n and you are correct. We have an issue opened with relative high priority about re-generating our docs and reviewing them. Also, the prompts used to generate these first ones wasn't the greated, nor real man pages were used (we used its knowledge base directly). 

I'll try to address this soon!

rafaeldtinoco (approved):
LGTM

Issue #3310, opened 2023-07-12
tracee deadlocking 


josedonizetti:
## Description

Tracee seems to deadlock when deploying to kubernetes. This can be emulated with kind, minikube very consistently. 

## Output of `tracee version`:

```
(paste your output here)
```

## Output of `uname -a`:

```
(paste your output here)
```

PR #3311, opened 2023-07-12
pkg/containers: fix deadlock

josedonizetti:
The write Lock() is not getting released because of the defer, blocking the read RLock(), and the pipeline.

Fix https://github.com/aquasecurity/tracee/issues/3310

We had a deadlock introduced on the last release on the PR https://github.com/aquasecurity/tracee/pull/3285

The write [`Lock()`](https://github.com/aquasecurity/tracee/blob/main/pkg/containers/containers.go#L383) isn't getting release due to the [`defer`](https://github.com/aquasecurity/tracee/blob/main/pkg/containers/containers.go#L384), which blocks the read [`RLock`](https://github.com/aquasecurity/tracee/blob/main/pkg/containers/containers.go#L398) and the pipeline.

To fix it, I remove the defer and did the lock() and unlock() on the critical path, also removed the `return`, we only update the cgroupMap inside the if logic, then move on to get the cgroupInfo and return. There are other ways of fixing it, let me know if you think it should be different.

### 2. Explain how to test it

The simplest to simulate the bug is to run tracee into kind or minukube.

### DIFF
--- a/pkg/containers/containers.go
+++ b/pkg/containers/containers.go
@@ -380,6 +380,8 @@ func (c *Containers) GetCgroupInfo(cgroupId uint64) CgroupInfo {
 		// cgroupInfo in the Containers struct. An empty subPath will make
 		// getCgroupPath() to walk all cgroupfs directories until it finds the
 		// directory of given cgroupId.
+		var cgroupInfo CgroupInfo
+
 		c.cgroupsMutex.Lock()
 		defer c.cgroupsMutex.Unlock()
 
@@ -387,12 +389,14 @@ func (c *Containers) GetCgroupInfo(cgroupId uint64) CgroupInfo {
 		if err == nil {
 			var stat syscall.Stat_t
 			if err = syscall.Stat(path, &stat); err == nil {
-				info, err := c.cgroupUpdate(cgroupId, path, time.Unix(stat.Ctim.Sec, stat.Ctim.Nsec))
-				if err == nil {
-					return info
+				cgroupInfo, err = c.cgroupUpdate(cgroupId, path, time.Unix(stat.Ctim.Sec, stat.Ctim.Nsec))
+				if err != nil {
+					logger.Errorw("cgroupUpdate", "error", err)
 				}
 			}
 		}
+
+		return cgroupInfo
 	}
 
 	c.cgroupsMutex.RLock()


NDStrahilevitz:
IMO this should be backported, thanks for finding this (and sorry for introducing it 😅).

josedonizetti:
@NDStrahilevitz can you review? Also, do you think backport, or can we release 0.16.1?

pkg/containers/containers.go
```diff
@@ -381,18 +381,17 @@ func (c *Containers) GetCgroupInfo(cgroupId uint64) CgroupInfo {
...
-				info, err := c.cgroupUpdate(cgroupId, path, time.Unix(stat.Ctim.Sec, stat.Ctim.Nsec))
-				if err == nil {
-					return info
+				_, err := c.cgroupUpdate(cgroupId, path, time.Unix(stat.Ctim.Sec, stat.Ctim.Nsec))
+				if err != nil {
+					logger.Errorw("cgroupUpdate", "error", err)
 				}
 			}
 		}
+		c.cgroupsMutex.Unlock()
```
NDStrahilevitz, line 394: I think we can return after here by...

pkg/containers/containers.go
```diff
@@ -381,18 +381,17 @@ func (c *Containers) GetCgroupInfo(cgroupId uint64) CgroupInfo {
...
-		defer c.cgroupsMutex.Unlock()
-
 		path, err := cgroup.GetCgroupPath(c.cgroups.GetDefaultCgroup().GetMountPoint(), cgroupId, "")
 		if err == nil {
 			var stat syscall.Stat_t
 			if err = syscall.Stat(path, &stat); err == nil {
-				info, err := c.cgroupUpdate(cgroupId, path, time.Unix(stat.Ctim.Sec, stat.Ctim.Nsec))
-				if err == nil {
-					return info
+				_, err := c.cgroupUpdate(cgroupId, path, time.Unix(stat.Ctim.Sec, stat.Ctim.Nsec))
```
NDStrahilevitz, line 388: ...doing this
```suggestion
				info, err = c.cgroupUpdate(cgroupId, path, time.Unix(stat.Ctim.Sec, stat.Ctim.Nsec))
```

Of course, that would require `var` declaring before.
This would save us from RLocking and RUnlocking after this path.

NDStrahilevitz (approved):
Correct as is, so merge like this if you want, or optimize as I suggested if you think it's ok.

NDStrahilevitz:
About the backporting, are you asking if we should branch out again to a new release branch, or backport to the existing v0.16.0 branch? If so, I believe we agreed on always doing the latter from now on.

josedonizetti:
@NDStrahilevitz changed, can you take another look? pls

yanivagman (approved):
LGTM

rafaeldtinoco:
Cherry-picked at https://github.com/aquasecurity/tracee/pull/3315 to v0.16.0 branch (will be at v0.16.1 release as well).

Issue #3313, opened 2023-07-12
policies: remove defaultActions and actions


josedonizetti:
The current defaultActions and actions in the policies are giving too much flexibility without the need, let's simplify it for now, while the action we support are only to output (file, webhook and fluentd). 

- defaultActions and Actions shouldn't be required
- for now only support either log (backwork compatibility) and print

PR #3314, opened 2023-07-12
feature: remove policy actions

josedonizetti:

Fix https://github.com/aquasecurity/tracee/issues/3313

This PR simplifies policies by removing defaultActions, and actions. The multiple actions feature creates a lot of complexity, which is not required. 

This allows policy creation without the need to specify any action (default or per event). The default action is always output and will depend on the tracee output config to use file, webhook, fluentd, or multiple printers. 

### 2. Explain how to test it

All the examples below should work:

```
name: dig
scope: 
  - binary=/usr/bin/dig
rules:
```

or, for backward compatibility

```
name: dig
defaultActions:
  - log
scope: 
  - binary=/usr/bin/dig
rules:
```


yanivagman (approved):
LGTM

PR #3321, opened 2023-07-13
remove help command, create flags markdown docs

### LINKED ISSUE #3299: Help subcommand doesn't work as expected
## Description

Help subcommand says the follwing:
```
./dist/tracee help
Tracee uses eBPF technology to tap into your system and give you
access to hundreds of events that help you understand how your system behaves.

Usage:
  tracee [flags]
  tracee [command]

Available Commands:
  completion  Generate the autocompletion script for the specified shell
->  help        Help about any command
  ...
```

However, when I want to get help about the output flag, I get the following:
```
./dist/tracee help output
Unknown help topic [output]
```

While I would expect to get help about the output flag.

Adding to the confusion, not only `help` subcommand exists, but also `--help` flag:
```
./dist/tracee --help output
Control how and where output is printed.
...
```

I think we should do the following:
1. Remove support for --help flag
2. Add help about flags in the help subcommand

Alternatively (and easier to implement), we should disable the `help` subcommand and only keep the `--help` flag

```
v0.16.0
```

## Output of `uname -a`:

```
(paste your output here)
```

geyslan:
Close #3299

5bac134c6 **fix(flags)!: del help cmd, add md files for flags** _<sub>(2023/jul/13) Geyslan Gregório \<geyslan@gmail.com\></sub>_

```
This removes the default cobra help command, leaving only the root
--help flag.

The remaining - in code - help flags strings are only used by the old
tracee-ebpf command.

With this, new binary flags help are inaccessible. To circumvent this,
the help flags are now documented in markdown files (not included to
online docs) to be used as source for man pages generation (next step).

BREAKING CHANGE: help command is gone, use --help flag instead.
```

#### Help flag

`sudo ./dist/tracee --help`

```
Tracee uses eBPF technology to tap into your system and give you
access to hundreds of events that help you understand how your system behaves.

Usage:
  tracee [flags]
  tracee [command]

Available Commands:
  analyze     Analyze past events with signature events [Experimental]
  completion  Generate the autocompletion script for the specified shell
  list        List traceable events
  list-caps   List available capabilities
  version     Print the version number of Tracee

Flags:
  -s, --scope stringArray            [uid|comm|container...]            Select workloads to trace by defining filter expressions
  -e, --events stringArray           [name|name.args.pathname...]       Select events to trace and event filters
  -p, --policy stringArray           [file|dir]                         Path to a policy or directory with policies
  -o, --output stringArray           [json|none|webhook...]             Control how and where output is printed (default [table])
  -c, --capture stringArray          [write|exec|network...]            Capture artifacts that were written, executed or found to be suspicious
      --config string                <file>                             Global config file (yaml, json between others - see documentation)
      --containers                                                      Enable container info enrichment to events. This feature is experimental and may cause unexpected behavior in the pipeline
      --crs stringArray              <runtime:socket>                   Define connected container runtimes
      --signatures-dir stringArray   <dir>                              Directories where to search for signatures in CEL (.yaml), OPA (.rego), and Go plugin (.so) formats
      --rego stringArray             [partial-eval|aio]                 Control event rego settings
  -b, --perf-buffer-size int         <size>                             Size, in pages, of the internal perf ring buffer used to submit events from the kernel (default 1024)
      --blob-perf-buffer-size int    <size>                             Size, in pages, of the internal perf ring buffer used to send blobs from the kernel (default 1024)
  -a, --cache stringArray            [type|mem-cache-size]              Control event caching queues (default [none])
      --metrics                                                         Enable metrics endpoint
      --healthz                                                         Enable healthz endpoint
      --pprof                                                           Enable pprof endpoints
      --pyroscope                                                       Enable pyroscope agent
      --listen-addr string           <url:port>                         Listening address of the metrics endpoint server (default ":3366")
  -C, --capabilities stringArray     [bypass|add|drop]                  Define capabilities for tracee to run with
      --install-path string          <dir>                              Path where tracee will install or lookup it's resources (default "/tmp/tracee")
  -l, --log stringArray              [debug|info|warn...]               Logger options (default [info])
  -h, --help                         help for tracee

Additional help topics:
  tracee            

Use "tracee [command] --help" for more information about a command.
```

#### Help command (removed)

`sudo ./dist/tracee help`

```
Error: unknown command "help" for "tracee"
```

### DIFF
--- a/cmd/tracee-ebpf/main.go
+++ b/cmd/tracee-ebpf/main.go
@@ -31,7 +31,7 @@ func main() {
 			// Logger Setup
 			logger.Init(logger.NewDefaultLoggingConfig())
 
-			flags.PrintAndExitIfHelp(c, false)
+			flags.PrintAndExitIfHelp(c)
 
 			if c.Bool("list") {
 				cmd.PrintEventList(false) // list events
--- a/cmd/tracee/cmd/analyze.go
+++ b/cmd/tracee/cmd/analyze.go
@@ -27,25 +27,6 @@ import (
 func init() {
 	rootCmd.AddCommand(analyze)
 
-	hfFallback := rootCmd.HelpFunc()
-	// Override default help function to support help for flags.
-	// Since for commands the usage is tracee help <command>, for flags
-	// the usage is tracee --help <flag>.
-	// e.g. tracee analyze --help rego
-	//      tracee analyze -h rego
-	analyze.SetHelpFunc(func(cmd *cobra.Command, args []string) {
-		if len(args) > 2 && (args[1] == "--help" || args[1] == "-h") {
-			flagHelp := flags.GetHelpString(args[2], true)
-			if flagHelp != "" {
-				fmt.Fprintf(os.Stdout, "%s\n", flagHelp)
-				os.Exit(0)
-			}
-		}
-
-		// If flag help was not found, fallback to default help function
-		hfFallback(cmd, args)
-	})
-
 	// flags
 
 	// events
--- a/cmd/tracee/cmd/list.go
+++ b/cmd/tracee/cmd/list.go
@@ -29,7 +29,6 @@ var listCmd = &cobra.Command{
 	Long:    ``,
 	Run: func(cmd *cobra.Command, args []string) {
 		// Get signatures to update event list
-
 		sigsDir, err := cmd.Flags().GetStringArray("signatures-dir")
 		if err != nil {
 			logger.Fatalw("Failed to get signatures-dir flag", "err", err)
--- a/cmd/tracee/cmd/root.go
+++ b/cmd/tracee/cmd/root.go
@@ -12,7 +12,6 @@ import (
 	"github.com/spf13/viper"
 
 	cmdcobra "github.com/aquasecurity/tracee/pkg/cmd/cobra"
-	"github.com/aquasecurity/tracee/pkg/cmd/flags"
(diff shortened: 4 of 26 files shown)


geyslan:
When this is merged, open an issue to generate man pages for flags using the introduced markdown files.

josedonizetti (commented):
@geyslan Not sure I understand why we added `list-caps`? Also, is the idea to merge into this release, and the man page would only be provided in the upcoming release?

geyslan:
> @geyslan Not sure I understand why we added `list-caps`? Also, is the idea to merge into this release, and the man page would only be provided in the upcoming release?

The main branch help capabilities generates the help with a list of available capabilities, as we're removing them (placing in markdowns), we need to provide similar functionality, so after discussion we've decided to make it a flag for the already existent list command `list --caps`. I'll updated it soon. 👍🏼

geyslan:
> > @geyslan Not sure I understand why we added `list-caps`? Also, is the idea to merge into this release, and the man page would only be provided in the upcoming release?
> 
> The main branch help capabilities generates the help with a list of available capabilities, as we're removing them (placing in markdowns), we need to provide similar functionality, so after discussion we've decided to make it a flag for the already existent list command `list --caps`. I'll updated it soon. 👍🏼

We've discussed this offline and decided to remove the available capabilities output for now while we don't set a system/debug option to centralise system information fetching.

josedonizetti (approved):

PR #3325, opened 2023-07-18
fix (cgroups): already dead edge case

### LINKED ISSUE #3324: Performance issues in short lived cgroup environments
Certain kubernetes version make use of short lived cgroups for various tasks (for example log rotation). These cgroups will generate events and a very quick cgroup_rmdir event.
As such, many related events to the cgroup will attempt to query its directory through the recursive path and not find it.

This is a cause for significant event loss in such environments.

NDStrahilevitz:
```
commit b3d043c558b50526fef1c7e65c8d89db8f984d28
Date:   Tue Jul 18 14:59:12 2023 +0000

    fix (cgroups): already dead edge case
    
    Certain kubernetes version make use of short lived cgroups for various
    tasks (for example log rotation). These cgroups will generate events and
    a very quick cgroup_rmdir event.
    As such, many related events to the cgroup will attempt to query its
    directory through the recursive path and not find it.
    
    Commit adds a "Dead" field to the cgroup info to indicate a cgroup which
    has already been removed. Various logical sections can refer to it if
    its relevant to them, and more importantly, additional queries will not
    be attempted.
    
    Bonus: optimize away additional Stat syscall in containers by returning
    the directory ctime in cgroup.GetCgroupPath.
```

Fix #3324

@rafaeldtinoco I want to backport this to v0.16.0 branch if that's ok with you.

### DIFF
--- a/pkg/ebpf/controlplane/controller.go
+++ b/pkg/ebpf/controlplane/controller.go
@@ -137,7 +137,7 @@ func (p *Controller) processCgroupMkdir(args []trace.Argument) error {
 	if err != nil {
 		return errfmt.WrapError(err)
 	}
-	if info.Container.ContainerId == "" {
+	if info.Container.ContainerId == "" && !info.Dead {
 		// If cgroupId is from a regular cgroup directory, and not the
 		// container base directory (from known runtimes), it should be
 		// removed from the containers bpf map.
--- a/pkg/cgroup/cgroup.go
+++ b/pkg/cgroup/cgroup.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 	"strings"
 	"syscall"
+	"time"
 
 	"github.com/aquasecurity/tracee/pkg/errfmt"
 	"github.com/aquasecurity/tracee/pkg/logger"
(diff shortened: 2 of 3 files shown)


pkg/ebpf/controlplane/controller.go
```diff
@@ -137,7 +137,7 @@ func (p *Controller) processCgroupMkdir(args []trace.Argument) error {
 	if err != nil {
 		return errfmt.WrapError(err)
 	}
-	if info.Container.ContainerId == "" {
+	if info.Container.ContainerId == "" && !info.Dead {
```
yanivagman, line 140: Can this ever happen?
It means that we get mkdir event after rmdir event, doesn't it?

yanivagman (approved):

pkg/ebpf/controlplane/controller.go
NDStrahilevitz: It does, and it was the original indicator for the error (although it was on a tracee v0.11.1 installation).
This can happen because it's a race between userspace and ebpf. The entry can be deleted either here OR in the cgroup_rmdir program.
I guess this would also require a misorder in events received, so this doesn't fully resolve the case, but the result here is only an annoying error log.

Issue #3323, opened 2023-07-17
k8s extend default policy


josedonizetti:
For Kubernetes we are installing a default policy with [only signatures](https://github.com/aquasecurity/tracee/blob/main/deploy/kubernetes/tracee/tracee.yaml#L1-L46), while the default set of events if you are using tracee without policies is much more significant. 

```
./dist/tracee list | grep default 
```

Let's have the same default events for both tracee started without policies, and tracee deployed in kubernetes with a default policy.

Issue #2725, opened 2023-02-15
Initial higher level events


AsafEitani:
This is about adding the basic higher level events - process execution, process termination and file deletion.

yanivagman:
part of #1310

yanivagman:
We will continue tracking this as part of #1310

PR #3328, opened 2023-07-19
feature(k8s): add default events to k8s policies

josedonizetti:

Fix https://github.com/aquasecurity/tracee/issues/3323

Adds a new policy to have all default events collected by default into kubernetes. 

### 2. Explain how to test it

```
# create a k8s cluster
minikube start

# deploy tracee
helm install tracee ./deploy/helm/tracee

# check it is running
kubectl get pods
kubectl logs -f ds/tracee
```

### DIFF
--- a/deploy/helm/tracee/templates/tracee-policies.yaml
+++ b/deploy/helm/tracee/templates/tracee-policies.yaml
@@ -6,6 +6,52 @@ metadata:
   labels:
     {{- include "tracee.labels" . | nindent 4 }}
 data:
+  default_events.yaml: |-
+    name: default_events
+    description: tracee default events
+    scope:
+      - global
+    rules:
+      - event: creat
+      - event: chmod
+      - event: fchmod
+      - event: chown
+      - event: fchown
+      - event: lchown
+      - event: ptrace
+      - event: setuid
+      - event: setgid
+      - event: setpgid
+      - event: setsid
+      - event: setreuid
+      - event: setregid
+      - event: setresuid
+      - event: setresgid
+      - event: setfsuid
+      - event: setfsgid
+      - event: init_module
+      - event: fchownat
+      - event: fchmodat
+      - event: setns
+      - event: process_vm_readv
+      - event: process_vm_writev
+      - event: finit_module
+      - event: memfd_create
+      - event: move_mount
+      - event: sched_process_exec
+      - event: security_inode_unlink
+      - event: security_socket_connect
+      - event: security_socket_accept
+      - event: security_socket_bind
+      - event: security_sb_mount
+      - event: container_create
+      - event: container_remove
+      - event: net_packet_icmp
+      - event: net_packet_icmpv6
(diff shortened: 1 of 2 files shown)


rafaeldtinoco (approved):
Possibly something you agreed with someone else (based on the issue description and an "Okay" emoji from Itay). The change looks technically correct so I'm approving. I just wonder why this change is being made. I thought that the "detections only" policy was quite nice (since users might not be interested in many of these default events). What is the idea behind this change ? (Out of curiosity only, feel free to merge).

josedonizetti:
@rafaeldtinoco Good question. For me, the idea is to be consistent with tracee without policies. Why do we have so many default events if you boot tracee with the binary or docker, but not for k8s? The experiences should be consistent. 
Why k8s is detections only? But for binary/docker we also have some other forensics events?

rafaeldtinoco:
> Why k8s is detections only? But for binary/docker we also have some other forensics events?

Yep, @itaysk and @yanivagman to answer/decide. I'm not sure why we're mixing tracing and detection events in the default experience.

IMHO, "detections only" seems more appropriate to be the default but, TBH, maybe this should require feedback from users and/or something similar (and not something I would be the best person to opinionate).

Like I said, change is "technically correct", as long as people are on the same page, I'm good.

josedonizetti:
@rafaeldtinoco Yeah, also not sure about the mix for the default events, I think our goal is to have higher level as default events along signatures in the future, but we still didn't implement them, so we are using the closest events? 

But I'm moving this to 0.18.0 so we can think a little more about it.

itaysk:
> I'm not sure why we're mixing tracing and detection events

the technical separation between "trace" and "signatures" was unnatural for users. for example "mem_prot_alert" is technically an event but logically a signature, "container created" was technically a signature but logically an event. Users went through lengths in order to get both streams, so instead of trying to justify this separation, we canceled it and made "everything as an event".  Now there is no more "trace" or "signature" it's just events. 

> users might not be interested in many of these default events

I do agree that the selected default events are probably too verbose, I would suggest to add just execution events for now.

josedonizetti:
> I do agree that the selected default events are probably too verbose, I would suggest to add just execution events for now.

Yes, that is why I think moving to 0.18.0 makes sense, because we will be adding `process_executed` and `process_exit`, we could have them + signatures are the defaults.

rafaeldtinoco:
> for example "mem_prot_alert" is technically an event but logically a signature, "container created" was technically a signature but logically an event. 

Indeed, indeed. Never stopped to think about this, super true.

> because we will be adding `process_executed` and `process_exit`, we could have them + signatures are the defaults.

Yep. That would be better IMO.

josedonizetti:
I'm closing this PR while higher level events aren't implemented -> https://github.com/aquasecurity/tracee/issues/2725, but there is a issue tracking it so we can implement it later https://github.com/aquasecurity/tracee/issues/3323

Issue #3322, opened 2023-07-13
k8s compatible configuration


itaysk:
1. Tracee moved from cli-first UX to yaml-first UX https://github.com/aquasecurity/tracee/issues/2900
2. This enabled easier configuration in k8s with ConfigMaps https://github.com/aquasecurity/tracee/issues/3118 https://github.com/aquasecurity/tracee/issues/2050
3. As a future step, we talked about making Tracee Kubernetes operator some day
4. A future operator will depend on CRDs which depends on the configuration structure we define today.

I think we should consider making the basic Tracee configuration entities Kubernetes-compatible, so that when it's time for CRD, there will be no breaking changes. (This issue is NOT to discuss a future Tracee Operator).

What does "kubernetes-compatible" mean? just the following:

```yaml
apiVersion:
kind:
metadata:
  name:
spec:
  <here goes the config file content>
```

For tracee config, there's no issue since it currently contains just config data. If we wanted to make it a CRD we just add the aforementioned wrapping around it.
For policies, there's a little overlap, namely in:
1. version: related: https://github.com/aquasecurity/tracee/issues/3320
2. name

PR #3330, opened 2023-07-20
feature(k8s): policy k8s compatible

josedonizetti:

Fix https://github.com/aquasecurity/tracee/issues/3322

This PR is changing policies to be compatible with Kubernetes. So a user will be able to use the same policy on the filesystem and also as a future Kubernetes CRD. 

Before the policy was:
````
name: dig
scope: 
  - binary=/usr/bin/dig
rules:
````

After this PR it will be:
```
apiVersion: aquasecurity.github.io/v1beta1
kind: TraceePolicy
metadata:
  name: dig
  annotations:
spec:
  scope: 
    - binary=/usr/bin/dig
  rules:
```

This PR also makes policies versioned, so if can be backward compatible if we decide to extend/change it later. 
I also made descriptions non mandatory, as we don't have `description` under metadata, due to a future CRD implementation, and annotations are not required in kubernetes.

### 2. Explain how to test it

Try one of the examples policies.

### 3. Other comments

This PR will break the CNDR e2e tests, so if I'll be creating a PR to fix the tests prior to merging this if approved.

### CODE REVIEW

File: docs/docs/events/overview.md [hunk 1]
```diff
@@ -18,12 +18,17 @@ Tracing `execve` events with [policies]:
...
-name: sample policy
-description: traces execve events
-scope:
-  - global
-rules:
-  - event: execve
+apiVersion: aquasecurity.github.io/v1beta1
+kind: TraceePolicy
+metadata:
+	name: sample policy
```
yanivagman, line 24: Not directly related to this PR but we should consider enforcing policy names without spaces (which may cause parsing issues when listed in the event output), and maybe also limit its size

yanivagman (approved):
LGTM

docs/docs/events/overview.md
josedonizetti: I created an issue to keep track of it https://github.com/aquasecurity/tracee/issues/3333

Issue #3333, opened 2023-07-24
policy: validate names to a max size, and whitespaces


josedonizetti:
context -> https://github.com/aquasecurity/tracee/pull/3330#discussion_r1271861230

PR #3335, opened 2023-07-25
fix: validate policy names are rfc 1123

josedonizetti:

Fix https://github.com/aquasecurity/tracee/issues/3333

Policy names must be k8s compatible, which means defined as [RFC 1123](https://tools.ietf.org/html/rfc1123). 
So the name must:

contain no more than 63 characters
contain only lowercase alphanumeric characters, '-' or '.'
start with an alphanumeric character
end with an alphanumeric character

https://kubernetes.io/docs/concepts/overview/working-with-objects/names/


geyslan:
@josedonizetti is this ready for review?

josedonizetti:
@geyslan yes

geyslan (approved):
LGTM

Issue #3327, opened 2023-07-19
Not equal mntns and pidns are not respected


geyslan:
## Description

Events from mount and pid namespaces are being wrongly sent when defined as "not equal".

When equal is set, and it's different from the origin namespace it is correctly filtered out.

```
sudo ./dist/tracee --scope "mntns!=$(readlink /proc/$$/ns/mnt | awk -F'[][]' '{print $2}')" --scope "pidns!=$(readlink /proc/$$/ns/pid | awk -F'[][]' '{print $2}')" -o table-verbose
```

![image](https://github.com/aquasecurity/tracee/assets/3372117/92e79ede-9b2b-4af6-a557-0c3a7713d7d9)

## Output of `tracee version`:

main branch

`Tracee version: "v0.14.2-141-gd8fa9648e"`

## Output of `uname -a`:

`Linux hb 6.1.38-1-MANJARO #1 SMP PREEMPT_DYNAMIC Wed Jul  5 23:49:30 UTC 2023 x86_64 GNU/Linux`

## Additional details 

I started some initial debugging and noticed that in compute_scopes() this event, when checked against the mnt ns filter (for example), has its bitmap cleared correctly. This implies that elsewhere it is either setting the bit incorrectly or not getting into the compute_scopes() logic.

PR #3337, opened 2023-07-25
fix(ebpf): size of mntns/pidns filters key holders

geyslan:
Close: #3327 

bb8c4c8ac **fix(ebpf): size of mntns/pidns filters key holders** _<sub>(2023/jul/25) Geyslan Gregório \<geyslan@gmail.com\></sub>_

```
Even though the namespace id is a 32-bit integer, the key size for the
mntns/pidns filters has been set to 64-bit. This caused these filters to
not work correctly as their key value was being read beyond limits.

Chore: Add integration tests for mntns/pidns filters.
```

### 2. Explain how to test it

See #3327.


rafaeldtinoco (approved):
LGTM

PR #3340, opened 2023-07-26
refactor(events): new event definitions (mutable vs immutable data)

rafaeldtinoco:
## Observation

1. This is not fully done. I can't make all the refactor at once, I need things merged first to understand future needs for the Extensions approach (and also to give some "briefing moment" in between the big changes, so others can suggest things as they use and also provide feedback).

  - This is true for the EventState struct, which will be fully encapsulated inside the EventGroup, for example (@josedonizetti FYI)

## Changes

401c33693 **fix(events): fix arm64 tailcalls indexes for unsupported syscalls**
1aded80d7 **chore(cmd): improve list of events output using definitions**
ea4247e90 **tests(definition): include GetDefinitions tests**
f89a8c315 **refactor(events): make Events become Definition**
24f5051f8 **refactor(events): Events.GetEventByID returns only the event**
4a69b5300 **refactor(events): encapsulate more event dependencies**
c9ea9821b **refactor(events): start core event group for extensions**
9d2495e47 **refactor(events): rename types to Event and EventGroup**
de4c87ae5 **fix(events): remove leftover from tailcalls initialization**
4208b5ce6 **fix(events): fix syntax and static checks issues**
dbe9eedc2 **chore(revive): no arg limit check**
01608804d **refactor(events): the only mutable type is event definition group**
90bf9f356 **refactor(events): make tailcall dependencies immutable**
9512b8640 **refactor(events): ksymbols and probes are immutable**
564dd53f4 **refactor(events): begin of event definition refactor**
9d3f96e0e **refactor(events): start re-organizing event definitions**
01e4aeb99 **fix(filter): remove unneeded workaround**

commit 401c33693

    fix(events): fix arm64 tailcalls indexes for unsupported syscalls
    
    Bring back the workaround not to index, in tailcall maps, the eBPF
    programs with event (syscall) indexes that are unsupported (other
    architectures should flag unsupported syscalls with IDs >= 10000).

commit 1aded80d7

    chore(cmd): improve list of events output using definitions
    
    - tabled event list output
    - include --wide option for better readability
    - use recent factored event definitions and event groups concept

commit ea4247e90

    tests(definition): include GetDefinitions tests

commit f89a8c315

    refactor(events): make Events become Definition
    
    - differentiate Events and Event Definitions
    - rename Event Group to Definition Group

commit 24f5051f8

    refactor(events): Events.GetEventByID returns only the event
    
    - Creates IsEventDefined() for event ID checking
    - Keeps a single GetEventByID() method
    - Refactor method usage
    - Minor styling changes

commit 4a69b5300

    refactor(events): encapsulate more event dependencies
    
    Encapsulates Event Definition Dependencies:
    - KSymbols
    - Probes
    - Capabilities (and simplify its type)
    
    Also:
    - Unify Events.NewEvent() and NewEventFull()
    - Rename methods as reviews are done
    - Change direct field access for the encapsulation methods

commit c9ea9821b

    refactor(events): start core event group for extensions

commit 9d2495e47

    refactor(events): rename types to Event and EventGroup
    
    - normalize syscall names to be exported (for future extensions)

commit de4c87ae5

    fix(events): remove leftover from tailcalls initialization

commit 4208b5ce6

    fix(events): fix syntax and static checks issues

commit dbe9eedc2

    chore(revive): no arg limit check

commit 01608804d

    refactor(events): the only mutable type is event definition group

commit 90bf9f356

    refactor(events): make tailcall dependencies immutable

commit 9512b8640

    refactor(events): ksymbols and probes are immutable

commit 564dd53f4

    refactor(events): begin of event definition refactor
    
    1. create CoreEventDefinitionGroup (TODO: "core" extension).
    2. EventDefinition and EventDefinitionGroup types.
    3. Make EventDefinition encapsulated.
    4. Keep temporary EvtDef instances (due to Dependencies).

commit 9d3f96e0e

    refactor(events): start re-organizing event definitions
    
    - turn the event dependency ID type just a slice of IDs
    - split code into multiple files (might still change) to organize

commit 01e4aeb99

    fix(filter): remove unneeded workaround
    
    Since 4.19 support was dropped (non CORE) there is no need for this
    workaround to exist anymore.


rafaeldtinoco:
Failing in arm64 because of:

```go
-                       for _, index := range tailCall.GetIndexes() {
-                               if index >= uint32(events.MaxCommonID) {
-                                       logger.Debugw(
-                                               "Removing index from tail call (over max event id)",
-                                               "tail_call_map", tailCall.GetMapName(),
-                                               "index", index,
-                                               "max_event_id", events.MaxCommonID,
-                                               "pkgName", pkgName,
-                                       )
-                                       tailCall.DelIndex(index) // remove undef syscalls (eg. arm64)
-                               }
-                       }
```

So I'll need to do a per arch event definition declaration (since we're now doing immutable declarations). might split the "syscall event group definition" for architecture (common, amd64, arm64 files, for example, for them).

pkg/events/event.go.go
```diff
@@ -0,0 +1,114 @@
+package events
+
+import (
+	"github.com/aquasecurity/tracee/types/trace"
+)
+
+type Event struct {
```
josedonizetti, line 7: This seems a bit confusing to me, because we will have `events.Event` and `trace.Event`, but one one is the event, the other is the definition of the event. The previous name seemed to be more appropriate `EventDefinition`. WDYT?

pkg/events/event.go.go
```diff
@@ -0,0 +1,114 @@
...
+
+	d.Events = append(d.Events, depsID...)
+
+	evt.dependencies = d
+
+	return evt
+}
+
+// NewEvent creates a new Event with all its fields set.
+func NewEventFull(
```
josedonizetti, line 38: maybe we can have only this constructor, named `NewEvent` and the other `NewEvent`which seems to be used to create an event from signature is forced to pass all values?

pkg/events/group.go
```diff
@@ -0,0 +1,200 @@
...
+import (
+	"sync"
+
+	"github.com/aquasecurity/tracee/pkg/errfmt"
+	"github.com/aquasecurity/tracee/pkg/logger"
+)
+
+// TODO: add states to the EventGroup struct (to keep states of events from that group)
+type EventState struct {
+	Submit uint64 // event that should be submitted to userspace (by policies bitmap)
```
josedonizetti, line 12: why make this public with get/sets?

pkg/events/group.go
```diff
@@ -0,0 +1,200 @@
...
+	evt, ok := e.events[givenEvt]
+	if !ok {
+		logger.Debugw("event context: event id not found", "id", givenEvt)
+		return Event{} // TODO: add an return error (or use only GetOk)
+	}
+
+	return evt
+}
+
+// TODO: remove this function ?
```
josedonizetti, line 107: is it possible? if so, i would also prefer.

pkg/events/group.go
```diff
@@ -0,0 +1,200 @@
...
+// Length returns the number of events in the event group.
+func (e *EventGroup) Length() int {
+	e.mutex.RLock()
+	defer e.mutex.RUnlock()
+	return len(e.events)
+}
+
+// GetAllEvents returns a new map of existing event instances (at the time of the call).
+// TODO: iterate internally after refactor is done ?
+func (e *EventGroup) GetAllEvents() map[ID]Event {
```
josedonizetti, line 130: `GetEvents`?

pkg/events/event.go.go
```diff
@@ -0,0 +1,114 @@
...
+
+func (e *Event) GetID() ID {
+	return e.id
+}
+
+func (e *Event) GetID32Bit() ID {
+	return e.id32Bit
+}
+
+func (e *Event) GetName() string {
```
josedonizetti, line 73: For getters of private fields, do you think we always need `GetXXX`, could it be `Name()`? wdyt?

josedonizetti (commented):
@rafaeldtinoco I added a couple of questions, most about naming, nothing huge, let me know what you think

pkg/events/group.go
rafaeldtinoco: This will be in the next parts of the refactor (this PR got too big). You can consider this PR a "1st phase". Its hard to make all changes in a single PR (it would take quite sometime and its better to have the refactor rolling so other PRs can rebase and suggest things/changes. 

This EventState will be a full encapsulated type, etc etc.

pkg/events/event.go.go
rafaeldtinoco: Yes this was in my original PR and it is definitely something we need (I'll make that change). I was actually making that change yesterday before my last push.

pkg/events/event.go.go
rafaeldtinoco: I like GetXXXX because of IDE completion (I want to get something I type Get and it shows me a list of what I can Get). I think its a good practice and I hope we can make it the default if others accept (on other refactors as well).

pkg/events/ids.go
yanivagman, line 1: In a previous PR you took the IDs into their own file (which I thought is nice) and now you merge them back to the event file. Any reason for that?

yanivagman (commented):
I like this change very much
Mostly nit comments ahead

pkg/events/event.go.go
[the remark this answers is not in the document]
rafaeldtinoco: Heheh no!

pkg/cmd/flags/filter.go
```diff
@@ -107,8 +107,8 @@ func prepareEventsToTrace(eventFilter eventFilter, eventsNameToID map[string]eve
 	isExcluded := make(map[events.ID]bool)
 
 	// build a map: k:set, v:eventID
-	for id, event := range events.Definitions.Events() {
-		for _, set := range event.Sets {
+	for id, event := range events.CoreEventDefinitionGroup.Events() {
```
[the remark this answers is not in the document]
rafaeldtinoco: Yep, not done yet, part of future changes (when extensions is started and their event groups are created).

pkg/events/dependencies.go
```diff
@@ -23,7 +24,7 @@ type Probe struct {
 
 type KSymbol struct {
 	Symbol   string
-	Required bool // tracee fails if symbol is not found
```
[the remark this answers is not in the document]
rafaeldtinoco: totally agree (and it was an accident, I should have reformatted it and not removed them).

pkg/ebpf/tracee.go
```diff
@@ -927,7 +927,12 @@ func (t *Tracee) validateKallsymsDependencies() {
 
 	// Cancel events with missing symbols dependencies
 	for eventToCancel, missingDepSyms := range missingSymsPerEvent {
-		logger.Errorw("Event canceled because of missing kernel symbol dependency", "missing symbols", missingDepSyms, "event", events.CoreEventDefinitionGroup.Get(eventToCancel).GetName())
+		can := events.CoreEventDefinitionGroup.Get(eventToCancel)
```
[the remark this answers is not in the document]
rafaeldtinoco: Yep, "cancel" is usually a cancel function from a `ContextWithCancel`, thats why I did not name it. Will change.

rafaeldtinoco:
I marked as solved the changes I already fixed in my local branch. Won't push it now, will do it tomorrow, so I can address all the rest.

pkg/events/dependencies.go
```diff
@@ -0,0 +1,171 @@
...
+type Event struct {
+	ID32Bit      ID
+	Name         string
+	DocPath      string // Relative to the 'doc/events' directory
+	Internal     bool
+	Syscall      bool
+	Dependencies Dependencies
+	Sets         []string
+	Params       []trace.ArgMeta
+}
```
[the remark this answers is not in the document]
rafaeldtinoco: It is now called "Definitions" and placed under definitions.go.
A event group has turned into "definitions group".
This way, we can clearly see what is an "event definition" and what is an "event".
All the places calling the methods had their variables renamed as well (to eventDefinition or equivalent).

pkg/cmd/flags/filter.go
```diff
@@ -104,9 +104,9 @@ func prepareEventsToTrace(eventFilter eventFilter, eventsNameToID map[string]eve
 	isExcluded := make(map[events.ID]bool)
 
 	// build a map: k:set, v:eventID
-	for id, eventDefinition := range events.Core.GetDefinitions() {
-		for _, set := range eventDefinition.GetSets() {
-			setsToEvents[set] = append(setsToEvents[set], id)
+	for _, eventDefinition := range events.Core.GetDefinitions() {
+		for _, sets := range eventDefinition.GetSets() {
```
yanivagman, line 108: we are ranging the sets, so set should be kept singular

pkg/ebpf/tracee.go
```diff
@@ -507,33 +507,50 @@ func (t *Tracee) generateInitValues() (InitValues, error) {
...
 		if events.Core.GetDefinitionByID(events.ID(index)).IsSyscall() {
-			err := t.probes.Attach(probes.SyscallEnter__Internal)
-			if err != nil {
-				return errfmt.WrapError(err)
-			}
-			err = t.probes.Attach(probes.SyscallExit__Internal)
-			if err != nil {
-				return errfmt.WrapError(err)
+			// Optimization: enable enter/exit probes only if at least one syscall is enabled.
+			once.Do(func() {
```
yanivagman, line 537: Why do we need that?
In the future, we may want to use this function dynamically to load new events that have tail call dependencies

yanivagman (commented):
Some commits in this PR undo or add on top of previous commits of this PR so they can probably be squashed.

I also think that this PR (that is already big enough) tries to do things that can be on their own PR like remove network workaround for older kernel (the first commit) or reformatting the list of events to be in a table. Consider taking them out of this PR into their own.

pkg/ebpf/tracee.go
rafaeldtinoco: Agreed. I just added the once.Do() because it makes no sense calling it over and over for the same thing (so there is a refactor need here). Also, the code was already here and I agree to your statement.

yanivagman (approved):
LGTM

PR #3345, opened 2023-07-27
[v0.17.0] fix(flags)!: del help cmd, add md files for flags (#3321)

geyslan:

f68b5c95d **fix(flags)!: del help cmd, add md files for flags (#3321)** _<sub>(2023/jul/27) Geyslan Gregório \<geyslan@gmail.com\></sub>_

```
This removes the default cobra help command, leaving only the root
--help flag.

The remaining - in code - help flags strings are only used by the old
tracee-ebpf command.

With this, new binary flags help are inaccessible. To circumvent this,
the help flags are now documented in markdown files (not included to
online docs) to be used as source for man pages generation (next step).

BREAKING CHANGE: help command is gone, use --help flag instead.

commit: d882cbf (main), cherry-pick
```

### 3. Other comments

### DIFF
--- a/cmd/tracee-ebpf/main.go
+++ b/cmd/tracee-ebpf/main.go
@@ -31,7 +31,7 @@ func main() {
 			// Logger Setup
 			logger.Init(logger.NewDefaultLoggingConfig())
 
-			flags.PrintAndExitIfHelp(c, false)
+			flags.PrintAndExitIfHelp(c)
 
 			if c.Bool("list") {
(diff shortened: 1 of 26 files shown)


rafaeldtinoco:
> BREAKING CHANGE: help command is gone, use --help flag instead.

Breaking changes aren't made for release branches. In this case this was softened due to a v0.17.0 not existing (yet) but we should not have breaking changes in release branches for sure.

PR #3348, opened 2023-07-27
chore(build): add LOGFROM flag to check-pr rule

geyslan:

cf7137649 **chore(build): add LOGFROM flag to check-pr rule**

commit cf7137649

```
With this flag, the git log output is filtered to only include commits
from the specified commit to the current HEAD.

make check-pr                # from main (default)
make check-pr LOGFROM=v0.1.0 # from tag/branch
make check-pr LOGFROM=HEAD~1 # from commit
make check-pr LOGFROM=f3a4b2 # from commit

This also changes check-pr output format and adds information for using
its output in the PULL_REQUEST_TEMPLATE.md file.
```
### 2. Explain how to test it

See commit above.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -15,7 +15,9 @@ PS: DO NOT JUMP THE CHECKLIST. GO BACK AND READ, ALWAYS!
 
 ### 1. Explain what the PR does
 
-<!-- Best advice is to put copy & paste your very well written git logs -->
+<!-- Best advice is to put copy & paste "make check-pr" PR Comment output -->
+
+"Replace me with `make check-pr` output"
 
 ### 2. Explain how to test it
 
--- a/Makefile
+++ b/Makefile
@@ -858,6 +858,8 @@ check-err: \
 # pull request verifier
 #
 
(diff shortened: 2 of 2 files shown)


rafaeldtinoco:
can we change a bit the format from "check-pr"? I'm not using it because I don't like the format. I use it to check everything needed, but I wont copy and paste the results.

2 things that bothers me:

1. Format:

93a5329 chore(build): add LOGFROM flag to check-pr rule (2023/jul/27) Geyslan Gregório <geyslan@gmail.com>

I believe we can make: <hash> description only

2. The "more" or "less" my pager does when `make check-pr` finishes.

Can we make those adjustments in this PR ?

rafaeldtinoco:
image

Feedback from this change:

I think the last list of commits should include the subject as well, to avoid empty stuff when the commit logs have a single subject line with no other lines.

Also, I still have a pager automatically called for the last part (it would be good not to have it so I can copy and paste all the output at once).

rafaeldtinoco:
@geyslan FYI

geyslan:
Ok, raising other PR with that change.

`xclip -selection clipboard` can be used but it would demand to check the tool and possibly override clipboard content. So, related to that, I'll let it as it is.

PR #3351, opened 2023-07-31
chore(build): change check-pr output format

geyslan:

4ef7356d6 **chore(build): change check-pr output format**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -866,7 +866,7 @@ check-pr: \
 	| .check_$(CMD_GIT)
 #
 	@echo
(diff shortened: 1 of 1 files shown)


geyslan:
For a body-less single commit.

image

geyslan:
I'm not liking how its outputting for body-less commits. I'll try other approach soon.

rafaeldtinoco (approved):
LGTM

geyslan:
Now, if the commit doesn't have body it's outputted again in the full list.

I think it is better:

image

PR #3352, opened 2023-07-31
fix(filter): remove unneeded workaround

rafaeldtinoco:
709399bc9 **fix(filter): remove unneeded workaround**

commit 709399bc9

```
Since 4.19 support was dropped (non CORE) there is no need for this
workaround to exist anymore.
```


yanivagman (approved):
LGTM

PR #3355, opened 2023-08-01
feature: validate kernel version

### LINKED ISSUE #3354: Kernel version validation
In order to avoid undefined behavior when running tracee in unsupported kernels, explicit kernel version validation needs to be added.
This is especially needed in applications using tracee as a library, as they do not have the implicit validation done in the tracee image through available BTFs.

NDStrahilevitz:
```
Date:   Tue Aug 1 12:29:06 2023 +0000

    feature: validate kernel version
    
    Tracee so far never explicitly validated kernel version when running.
    It is implicitly validated through available BTFs, and in NON-CORE it
    was validated through a compile time directive.
    When using tracee as a library this is an issue as no validation is done
    and undefined behavior may occur in unsupported kernels.
```

```
Date:   Tue Aug 1 12:59:15 2023 +0000

    chore(go.mod): bump libbpfgo helpers
    
    Include fix to osinfo parsing for centos.
```

Fix #3354

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"fmt"
 	"io"
 
 	"github.com/aquasecurity/libbpfgo/helpers"
@@ -9,6 +10,7 @@ import (
 	"github.com/aquasecurity/tracee/pkg/errfmt"
 	"github.com/aquasecurity/tracee/pkg/events"
 	"github.com/aquasecurity/tracee/pkg/events/queue"
+	"github.com/aquasecurity/tracee/pkg/logger"
 	"github.com/aquasecurity/tracee/pkg/policy"
 	"github.com/aquasecurity/tracee/pkg/signatures/engine"
 	"github.com/aquasecurity/tracee/types/trace"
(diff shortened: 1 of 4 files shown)


pkg/config/config.go
```diff
@@ -37,6 +38,28 @@ type Config struct {
...
+	// Default minimum kernel version
+	minimalKernelVersion := "5.4"
+	if c.OSInfo.GetOSReleaseID() == helpers.CENTOS {
+		// In centos tracee is supported in 4.18 kernel as well
+		minimalKernelVersion = "4.18"
+	}
+
+	res, err := c.OSInfo.CompareOSBaseKernelRelease(minimalKernelVersion)
+	if err != nil {
+		// TODO  what should we do?
```
geyslan, line 53: @rafaeldtinoco some suggestion?

geyslan (approved):
Asked @rafaeldtinoco about what TODO, but LGTM overall.

geyslan:
@NDStrahilevitz in the first integration test case: https://github.com/aquasecurity/tracee/actions/runs/5727088418/job/15518969106?pr=3355#step:4:373

Somehow the Validate failed for github environment.

pkg/config/config.go
```diff
@@ -37,6 +38,28 @@ type Config struct {
 
 // Validate does static validation of the configuration
 func (c Config) Validate() error {
+	// Validate kernel version
+	// For valid versions see https://aquasecurity.github.io/tracee/latest/getting-started/installing/prerequisites/
+
+	// Default minimum kernel version
+	minimalKernelVersion := "5.4"
+	if c.OSInfo.GetOSReleaseID() == helpers.CENTOS {
```
yanivagman, line 46: not only centos but also rhel
Actually, 4.19 is the minimal version now and 5.4 is required for the network features only

pkg/config/config.go
yanivagman: I don't think we should return an error if compare method itself failed.
Better to continue opportunistically in that case

pkg/config/config.go
```diff
@@ -37,6 +38,28 @@ type Config struct {
...
+	res, err := c.OSInfo.CompareOSBaseKernelRelease(minimalKernelVersion)
+	if err != nil {
+		// TODO  what should we do?
+		return fmt.Errorf("couldn't check kernel version compatibility: %v", err)
+	}
+	if res == helpers.KernelVersionNewer {
+		return fmt.Errorf("unsupported kernel version, must be at least %s", minimalKernelVersion)
+	}
+	if res == helpers.KernelVersionInvalid {
+		return fmt.Errorf("invalid kernel version to compare: %s", minimalKernelVersion)
```
yanivagman, line 60: Can this happen?

rafaeldtinoco: Kernels < 5.2 (or even 5.4) that aren't RHEL/CentOS/Alma/Rocky/Oracle would probably be unsupported (unless we start testing and open issues for the load problems they likely have. 

We have to have in mind the following:

- If the kernel isn't supported by a DISTRO or by UPSTREAM (kernel.org) then we can't support.
- If the kernel does not support BTF, it has to be present in BTFHUB for us to support.

We probably work in all kernels >= 5.4 with embedded BTF I believe (despite the distro version).

rafaeldtinoco: Kernels supported by upstream:

image

pkg/config/config.go
NDStrahilevitz: I went by the prerequisite given in the tracee manual. I see that our BTFs don't align with this check. It could be the case that this PR may be too premature, as we need a better image of the validation.

NDStrahilevitz:
Turned out to be unnecessary, closing for now.

PR #3361, opened 2023-08-03
feat: filter file capture by ELF type

### LINKED ISSUE #3359: Capture files ELF filter
Most of the time we are most interested in capturing ELF files when capturing read/write of files.
To reduce the load of captured files, we should had a filter in the eBPF code checking the file header to check for ELF files, and filter by it if the user request so.

AlonZivony:

Add a filter for file IO capture filtering ELF files. ELF files are the most important files to capture in security research, so filtering according to ELF files is very effective to reduce noise.

fix #3359

### DIFF
--- a/pkg/ebpf/c/capture_filtering.h
+++ b/pkg/ebpf/c/capture_filtering.h
@@ -15,11 +15,12 @@
 #define FILTER_NORMAL_FILES (1 << (FILTER_FILE_TYPE_START_BIT + 0))
 #define FILTER_PIPE_FILES   (1 << (FILTER_FILE_TYPE_START_BIT + 1))
 #define FILTER_SOCKET_FILES (1 << (FILTER_FILE_TYPE_START_BIT + 2))
+#define FILTER_ELF_FILES    (1 << (FILTER_FILE_TYPE_START_BIT + 3))
 #define FILTER_STDIN_FILES  (1 << (FILTER_FILE_FD_START_BIT + STDIN))
 #define FILTER_STDOUT_FILES (1 << (FILTER_FILE_FD_START_BIT + STDOUT))
 #define FILTER_STDERR_FILES (1 << (FILTER_FILE_FD_START_BIT + STDERR))
 
-#define FILTER_FILE_TYPE_MASK ((FILTER_SOCKET_FILES << 1) - FILTER_NORMAL_FILES)
+#define FILTER_FILE_TYPE_MASK ((FILTER_ELF_FILES << 1) - FILTER_NORMAL_FILES)
 #define FILTER_FDS_MASK       ((FILTER_STDERR_FILES << 1) - FILTER_STDIN_FILES)
 
 #define CAPTURE_READ_TYPE_FILTER_IDX  0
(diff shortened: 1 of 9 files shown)


pkg/ebpf/c/capture_filtering.h
```diff
@@ -93,15 +99,41 @@ statfunc bool filter_file_type(void *ctx, void *filter_map, size_t map_idx, stru
...
                 type_filter_match = true;
             }
         }
+        if (*type_filter & FILTER_ELF_FILES) {
+            file_id_t file_id = get_file_id(file);
+            file_id.ctime = 0;
+            if (start_pos == 0) {
+                // Check if header is matching ELF header
+                u8 header[FILE_MAGIC_HDR_SIZE] = {};
+                fill_file_header(io_data, header);
```
yanivagman, line 119: nit: usually destination (header) argument comes before source (io_data) argument, so I would switch the arguments order to avoid confusion

pkg/ebpf/c/capture_filtering.h
```diff
@@ -93,15 +99,41 @@ statfunc bool filter_file_type(void *ctx, void *filter_map, size_t map_idx, stru
...
+                if (header[0] == 0x7f && header[1] == 0x45 && header[2] == 0x4c &&
+                    header[3] == 0x46) {
+                    type_filter_match = true;
```
yanivagman, line 122: nit: since you changed the `if...else if... else` pattern to be `if...;if...;` and it seems that once any of the filters matched you set type_filter_match to true (thus an OR filter), you can skip any other checks after a match and go directly to the return code (e.g. using goto)

pkg/ebpf/c/capture_filtering.h
```diff
@@ -93,15 +99,41 @@ statfunc bool filter_file_type(void *ctx, void *filter_map, size_t map_idx, stru
...
             if (imode_mode & S_IFREG) {
                 type_filter_match = true;
             }
         }
+        if (*type_filter & FILTER_ELF_FILES) {
+            file_id_t file_id = get_file_id(file);
+            file_id.ctime = 0;
+            if (start_pos == 0) {
+                // Check if header is matching ELF header
+                u8 header[FILE_MAGIC_HDR_SIZE] = {};
```
yanivagman, line 118: Shouldn't this array be initialized to 0?
This can be important if header size is less than `FILE_MAGIC_HDR_SIZE`

pkg/ebpf/c/maps.h
```diff
@@ -117,6 +117,7 @@ BPF_HASH(logs_count, bpf_log_t, bpf_log_count_t, 4096);            // logs count
 BPF_PERCPU_ARRAY(scratch_map, scratch_t, 1);                       // scratch space to avoid allocating stuff on the stack
 BPF_LRU_HASH(file_modification_map, file_mod_key_t, int, 10240);   // hold file data to decide if should submit file modification event
 BPF_LRU_HASH(io_file_path_cache_map, file_id_t, path_buf_t, 5);    // store cache for IO operations path
+BPF_LRU_HASH(elf_files_map, file_id_t, bool, 16);                  // store cache for file ELF type check
```
yanivagman, line 120: Are 16 entries enough here?

pkg/ebpf/c/capture_filtering.h
AlonZivony: So using the syntax of `x = {}` should initialized the value to zero as far as I know.
I have checked it once with @roikol and it worked.

pkg/ebpf/c/maps.h
AlonZivony: So the cache here should be for a repetitive write/read probe trigger because of a large file write/read.
The goal here is not to cache it for the case that someone tries to bypass our check (he could then write the first bytes only in the end, and bypass us anyways).

yanivagman (approved):
LGTM

PR #3384, opened 2023-08-13
fix: print_mem_dump fails on missing symbol

### LINKED ISSUE #3382: Old kernel symbols dependency semantics broken
Previously, kernel symbols dependency semantics differentiated between three cases:
1. nil dependency - ksymbols irrelevant for the event.
2. empty array event - ksymbols required but none need to be preloaded specifically.
3. non-empty array - ksymbols required and specific symbols need to be preloaded.

Due to breaking this semantic, events such as `print_mem_dump` now fail when loading on their own:
```
tracee -e print_mem_dump.args.symbol_name=compat_filldir64
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x1ae51d1]

goroutine 1 [running]:
github.com/aquasecurity/tracee/pkg/ebpf.(*Tracee).triggerMemDump(_, {0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...})
        /home/ubuntu/tracee/pkg/ebpf/tracee.go:1691 +0xd91
github.com/aquasecurity/tracee/pkg/ebpf.(*Tracee).Run(0xc001952000, {0x315d968, 0xc000716140})
        /home/ubuntu/tracee/pkg/ebpf/tracee.go:1323 +0x165
github.com/aquasecurity/tracee/pkg/cmd.Runner.Run({{0xc000330000, 0xc000338070, 0xc00033b400, 0xc001d053d0, {0x0, 0x0}, 0x400, 0x400, 0x0, {0x0, ...}, ...}, ...}, ...)
        /home/ubuntu/tracee/pkg/cmd/tracee.go:87 +0x385
github.com/aquasecurity/tracee/cmd/tracee/cmd.glob..func4(0xc0007bfde0?, {0xc000040190?, 0x2?, 0x2?})
        /home/ubuntu/tracee/cmd/tracee/cmd/root.go:61 +0x258
github.com/spf13/cobra.(*Command).execute(0x3d39ac0, {0xc000040190, 0x2, 0x2})
        /home/ubuntu/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:944 +0x847
github.com/spf13/cobra.(*Command).ExecuteC(0x3d39ac0)
        /home/ubuntu/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:1068 +0x3bd
github.com/spf13/cobra.(*Command).Execute(...)
        /home/ubuntu/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:992
github.com/aquasecurity/tracee/cmd/tracee/cmd.Execute()
        /home/ubuntu/tracee/cmd/tracee/cmd/root.go:336 +0x34
main.main()
        /home/ubuntu/tracee/cmd/tracee/main.go:11 +0x19
```

### LINKED ISSUE #3383: print_mem_dump fails when one symbol doesn't exist on system
The `print_mem_dump` event can dump the function of many different functions loaded in the kernel, as many as are loaded in its "argument" parameters.
However, its handler code currently hard-fails when any symbol query fails. This means the event will not be generated, despite still being useful for other symbols.
```
tracee -e print_mem_dump.args.symbol_name=compat_filldir64
{"level":"warn","ts":1691929975.2542894,"msg":"print_mem_dump returned an error","error":"ebpf.(*Tracee).triggerMemDump: symbol not found: system___arm64_sys_compat_filldir64"}
^C
End of events stream
Stats: {EventCount:{value:0} EventsFiltered:{value:0} NetCapCount:{value:0} BPFLogsCount:{value:0} ErrorCount:{value:0} LostEvCount:{value:0} LostWrCount:{value:0} LostNtCapCount:{value:0} LostBPFLogsCount:{value:0}}
```

NDStrahilevitz:
Fix #3382
commit cbac4047cf596712a353542a0a9b6586c8cf3844
```
Date:   Sun Aug 13 12:38:37 2023 +0000

    fix: non-nil empty ksymbols dependency
    
    The events group refactor (commit aa3fd66) changed the semantics of
    ksymbols dependency so that kernel symbols would only be loaded if a
    specific list was set.
    This broke several events which require kernel symbols in general, but
    not any specifically (being set through argument "parameter" lists).
    Fix this by bringing back previous semantics.
```

Fix #3383 
commit b57331706b726343af011af9a85c2711dd25c633
```
Date:   Sun Aug 13 12:41:05 2023 +0000

    fix: print_mem_dump fail on missing symbol
    
    print_mem_dump trigger currently fails when failing to query a symbol.
    However print_mem_dump can potentially include many symbols in its
    parameters, and shouldn't fail the whole trigger based on one symbol.
    Skip the error return when a symbol query fails and log a WARN level
    message instead.
```

### DIFF
--- a/pkg/ebpf/tracee.go
+++ b/pkg/ebpf/tracee.go
@@ -1700,7 +1700,8 @@ func (t *Tracee) triggerMemDump(event trace.Event) error {
 						}
 					}
 					if err != nil {
-						return errfmt.WrapError(err)
+						logger.Warnw("print_mem_dump: failed to get symbol info", "symbol", name)
+						continue
 					}
 				}
 				eventHandle := t.triggerContexts.Store(event)
--- a/pkg/ebpf/tracee.go
+++ b/pkg/ebpf/tracee.go
@@ -497,8 +497,8 @@ func (t *Tracee) generateInitValues() (InitValues, error) {
 		if !events.Core.IsDefined(evt) {
 			return initVals, errfmt.Errorf("event %d is undefined", evt)
 		}
-		for range events.Core.GetDefinitionByID(evt).GetDependencies().GetKSymbols() {
-			initVals.Kallsyms = true // only if length > 0
+		if events.Core.GetDefinitionByID(evt).GetDependencies().GetKSymbols() != nil {
+			initVals.Kallsyms = true
 		}
 	}
 
--- a/pkg/events/definition_dependencies.go
+++ b/pkg/events/definition_dependencies.go
@@ -38,10 +38,12 @@ func (d Dependencies) GetIDs() []ID {
 	return d.ids
 }
 
+// GetKSymbols returns the KSymbols dependencies of an event.
+// If nil is returned no ksymbols are needed.
+// If an empty array is returned - symbols are needed, but none should specifically
+// be loaded ahead of time.
+// A non-empty array indicates that some symbols should be queried and loaded ahead of time.
 func (d Dependencies) GetKSymbols() []KSymbol {
-	if d.kSymbols == nil {
-		return []KSymbol{}
-	}
 	return d.kSymbols
 }
 


pkg/ebpf/tracee.go
```diff
@@ -1700,7 +1700,8 @@ func (t *Tracee) triggerMemDump(event trace.Event) error {
 						}
 					}
 					if err != nil {
-						return errfmt.WrapError(err)
+						logger.Warnw("print_mem_dump: failed to get symbol info", "symbol", name)
```
NDStrahilevitz, line 1703: Maybe this should be DEBUG level instead?

yanivagman (approved):
LGTM

PR #3388, opened 2023-08-15
update squidfunk/mkdocs-material.git

sulaiman-coder:


josedonizetti (changes requested):
@sulaiman-coder Thank yo for opening a PR. Though I'm not sure this is something we want to do as we use the [insiders fork](https://squidfunk.github.io/mkdocs-material/insiders/). Is there a reason why you changing it from the insiders' fork?

Issue #3208, opened 2023-06-07
GRPC v1 API 


josedonizetti:
There Tracee API server involves several changes which we are tracking at https://github.com/aquasecurity/tracee/issues/2991 across releases. 

This issue is tracking the work to be done to have a v1 GRPC API for 0.17.0. The v1 implementation should support the endpoints discussed:

```
service TraceeService {
    rpc GetEventDefinition(GetEventDefinitionRequest) returns (GetEventDefinitionResponse);
    rpc StreamEvents(StreamEventsRequest) returns (stream StreamEventsResponse);

    // global actions, enable/disable event
    rpc EnableEvent(EnableEventRequest) returns (EnableEventResponse);
    rpc DisableEvent(DisableEventRequest) returns (DisableEventResponse);

    rpc GetVersion(GetVersionRequest) returns (GetVersionResponse);
}

service diagnostic {
    GetMetrics()
    ChangeLogLevel() 
    DumpStack()
}
```

- [x] GRPC Server
- [x] Stream abstraction to be able to hook new output streams into the pipeline 
- [x] Migrate api to root dir (eg: "./api")
- [x] StreamEvents endpoint
- [x] Add Enable/Disable rule to Tracee
- [x] Add Enable/Disable rule endpoint
- [x] Add GetEventDefinition endponit
- [x] Add Enable/Disable event
- [x]  Add Enable/Disable event endpoint
- [x] Version endpoint
- [x] Add Diagnostic service for ChangeLogLevel
- [x] Add Diagnostic service for DumpStack
- [x] Add Diagnostic service for GetMetrics

PR #3389, opened 2023-08-15
feat: add tracee rpc service

josedonizetti:
Adds tracee rpc service with a first endpoint to return tracee's version

This is the first PR to support a [v1 for GRPC](https://github.com/aquasecurity/tracee/issues/3208). It adds a Tracee rpc service with an endpoint to return its version. The rpc now is declared as `v1beta1` because we are still discussing the endpoints, but once we decide it we will be changing it to be `v1`. 

The PR also adds a Makefile target to generate the code from the proto definition. 

### 2. Explain how to test it

```
# do a change to the proto file
make protoc
```


types/api/v1beta1/tracee.proto
```diff
@@ -0,0 +1,16 @@
...
+message GetVersionRequest {
+}
+
+message GetVersionResponse {
+    string version = 1;
+}
+
+service TraceeService {
+    rpc GetVersion(GetVersionRequest) returns (GetVersionResponse);
+}
```
geyslan, line 16: nit: new line

```suggestion
}

```

Makefile
```diff
@@ -36,6 +36,7 @@ CMD_STATICCHECK ?= staticcheck
 CMD_STRIP ?= llvm-strip
 CMD_TOUCH ?= touch
 CMD_TR ?= tr
+CMD_PROTOC ?= protoc
```
geyslan, line 39: Running `make protoc` on Manjaro:

```
protoc \
        --go_out=. \
        --go_opt=paths=source_relative \
        --go-grpc_out=. \
        --go-grpc_opt=paths=source_relative ./types/api/v1beta1/*.proto
protoc-gen-go: program not found or is not executable
Please specify a program using absolute path or make sure the program is available in your PATH system variable
--go_out: protoc-gen-go: Plugin failed with status code 1.
make: *** [Makefile:919: protoc] Error 1
```

geyslan: The same happened on Ubuntu (vagrant), probably we have to update the Vagrantfile as well.

types/api/v1beta1/tracee.proto
josedonizetti: Is there a default agreed for this we follow?

geyslan: I think we're following the github alert.

image

But indeed not discussed between us.

josedonizetti: 🤔 I don't think it is an alert as something bad hahaha, but I don't mind adding a new line, maybe we can find some protobuf linter that keeps an eye on this.

geyslan (approved):
LGTM

Tested:

image

PR #3394, opened 2023-08-16
Add grpc diagnostic

josedonizetti:

This PR is part of the work to support a [v1 for GRPC](https://github.com/aquasecurity/tracee/issues/3208). This PR implements the diagnostic rpc service. I've tested with `-race` no data races were found here. 

### 2. Explain how to test it

The below clients can be used to test the GRPC API:

```go
import (
        pb "github.com/aquasecurity/tracee/types/api/v1beta1"
)
```

```go
func main() {
        var opts []grpc.DialOption

        if err != nil {
        }

        client := pb.NewDiagnosticServiceClient(conn)

        changelog := &pb.ChangeLogLevelRequest{
                Level: pb.LogLevel_Info,
        }

        _, err = client.ChangeLogLevel(context.Background(), changelog)
        if err != nil {
                log.Fatal(err)
        }

        fmt.Println("Log level changed to DEBUG")
}
```

```go
func main() {
	var opts []grpc.DialOption

	if err != nil {
	}

	client := pb.NewDiagnosticServiceClient(conn)

	stacktrace, err := client.GetStacktrace(context.Background(), &pb.GetStacktraceRequest{})
	if err != nil {
		log.Fatal(err)
	}

	fmt.Println(string(stacktrace.Stacktrace))
}
```

```go
func main() {
	var opts []grpc.DialOption

	if err != nil {
	}

	client := pb.NewDiagnosticServiceClient(conn)

	metrics, err := client.GetMetrics(context.Background(), &pb.GetMetricsRequest{})
	if err != nil {
		log.Fatal(err)
	}

	fmt.Println("EventCount", metrics.EventCount)
	fmt.Println("EventsFiltered", metrics.EventsFiltered)
	fmt.Println("NetCapCount", metrics.NetCapCount)
	fmt.Println("BPFLogsCount", metrics.BPFLogsCount)
	fmt.Println("ErrorCount", metrics.ErrorCount)
	fmt.Println("LostEvCount", metrics.LostEvCount)
	fmt.Println("LostWrCount", metrics.LostWrCount)
	fmt.Println("LostNtCapCount", metrics.LostNtCapCount)
	fmt.Println("LostBPFLogsCount", metrics.LostBPFLogsCount)
}
```

### 3. Other comments

depends on https://github.com/aquasecurity/tracee/pull/3395

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/Masterminds/sprig/v3 v3.2.3
 	github.com/aquasecurity/libbpfgo v0.4.8-libbpf-1.2.0.0.20230509162948-80f41e18e690
 	github.com/aquasecurity/libbpfgo/helpers v0.4.6-0.20230321190037-f591a2c5734f
-	github.com/aquasecurity/tracee/types v0.0.0-20230816112254-7c722620761f
+	github.com/aquasecurity/tracee/types v0.0.0-20230817130859-7d4b8c20396c
 	github.com/containerd/containerd v1.7.0
 	github.com/docker/docker v23.0.5+incompatible
 	github.com/golang/protobuf v1.5.3
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -88,7 +88,7 @@ const (
 // Tracee offers aggregation and filtering support on top of any logger implementation complying to it's interface.
 type LoggingConfig struct {
 	Logger        LoggerInterface
-	loggerConfig  LoggerConfig
+	LoggerConfig  LoggerConfig
 	Filter        LoggerFilter
 	Aggregate     bool
 	FlushInterval time.Duration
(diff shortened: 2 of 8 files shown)


pkg/logger/logger.go
```diff
@@ -88,7 +88,7 @@ const (
 // Tracee offers aggregation and filtering support on top of any logger implementation complying to it's interface.
 type LoggingConfig struct {
 	Logger        LoggerInterface
-	loggerConfig  LoggerConfig
+	LoggerConfig  LoggerConfig
```
josedonizetti, line 91: I don't like this refactoring, but it had to be done because [flags create a LoggingConfig](https://github.com/aquasecurity/tracee/blob/main/pkg/cmd/flags/logger.go#L292), and if this field wasn't exposed it would be nil, generating a panic when you change the log level. But it is weird to allow the user to create a struct but have methods depending on it, we should enforce it by perhaps having these fields private and a constructor function. Nonetheless, I'll create a separate issue to handle this refactor.

go.mod
```diff
@@ -173,3 +173,5 @@ require (
 )
 
 replace github.com/kubernetes/cri-api => k8s.io/cri-api v0.23.5-rc.0
+
+replace github.com/aquasecurity/tracee/types v0.0.0-20230816112254-7c722620761f => ./types
```
josedonizetti, line 177: will be changed once this is merged https://github.com/aquasecurity/tracee/pull/3395

geyslan (approved):
LGTM

Tested the new three GPRC services.

image

image

For debug `ChangeLogLevel` I've put a dummy `logger.Debugw()` and tested changing it to Debug and returning it to Info.

PR #3395, opened 2023-08-16
feat: add diagnostic rpc

josedonizetti:

This PR is part of the work to support a [v1 for GRPC](https://github.com/aquasecurity/tracee/issues/3208). It adds a diagnostic rpc for tracee debugging, this rpc can be used to get a stacktracee live from tracee, its metrics and change log levels. This PR also fixes the version on the tracee.proto, which should have been `v1beta1`


geyslan (approved):
LGTM.

I would consider just to move the [types/api/v1beta1/tracee.proto](https://github.com/aquasecurity/tracee/pull/3395/files#diff-5b3b85e281bd24353a6559822f8d6293c641c633aa7ccfe78f3195d2e0627c29) fix to a proper commit like "fix(protobuf): tracee.proto go_package".

types/api/v1beta1/diagnostic.proto
```diff
@@ -0,0 +1,50 @@
...
+option go_package = "github.co/aquasecurity/tracee/api/v1beta1";
+
+package tracee.v1beta1;
+
+message GetMetricsRequest {
+}
+
+message GetMetricsResponse {
+    uint64 EventCount = 1;
+	uint64 EventsFiltered = 2;
```
yanivagman, line 12: nit: indentation

types/api/v1beta1/diagnostic.proto
```diff
@@ -0,0 +1,50 @@
...
+	uint64 ErrorCount = 5;
+	uint64 LostEvCount = 6; 
+	uint64 LostWrCount = 7; 
+	uint64 LostNtCapCount = 8; 
+	uint64 LostBPFLogsCount = 9;
+}
+
+enum LogLevel {
+	Info = 0;
+	Debug = 1;
```
yanivagman, line 24: Like discussed offline - info and debug should be switched

PR #3408, opened 2023-08-22
[DRAFT] playing around process tree (signal event definition)

rafaeldtinoco:
Based on the reviews at: https://github.com/aquasecurity/tracee/pull/3364

I would like some feedback before moving further.

I have created "signal events", which are regular event definitions but for the control plane. This allows us to initialize the event dependencies just like we do for all other events (removing probe group creation from the control plane).

I have also created an example (unfinished) for the SchedProcessFork to check if duplicating the eBPF programs like this is what was asked.

@yanivagman Please provide feedback if this is good so I can move further (based on your inputs in the last PR).

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -5939,6 +5939,8 @@ CGROUP_SKB_HANDLE_FUNCTION(proto_tcp_http)
     return 1; // NOTE: might block HTTP here if needed (return 0)
 }
 
+// clang-format on
+
 //
 // Control Plane Programs
 //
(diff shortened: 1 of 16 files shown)


pkg/utils/hash.go
```diff
@@ -0,0 +1,77 @@
+package utils
+
+import (
+	"encoding/binary"
+	"unsafe"
+)
+
+// MurMurHash 3 x86 32-bit (https://en.wikipedia.org/wiki/MurmurHash): Small (u32), simple (for C
```
rafaeldtinoco, line 8: I have also created a golang version of the Murmur Hash 32 as requested.

types/trace/trace.go
```diff
@@ -42,10 +42,18 @@ type Event struct {
 	Syscall               string       `json:"syscall"`
 	StackAddresses        []uint64     `json:"stackAddresses"`
 	ContextFlags          ContextFlags `json:"contextFlags"`
-	Args                  []Argument   `json:"args"` // Arguments are ordered according their appearance in the original event
+	EntityID              uint32       `json:"-"`    // task unique identifier (*)
```
rafaeldtinoco, line 45: Lets see if having the current task hash is enough to maintain the tree (instead of having parent starttime and parent hash).

pkg/events/core.go
```diff
@@ -10528,6 +10538,120 @@ var CoreEvents = map[ID]Definition{
...
+			{Type: "int", Name: "parent_tid"},
+			{Type: "int", Name: "parent_ns_tid"},
+			{Type: "int", Name: "parent_pid"},
+			{Type: "int", Name: "parent_ns_pid"},
+			{Type: "unsigned long", Name: "parent_start_time"},
+			{Type: "int", Name: "child_tid"},
+			{Type: "int", Name: "child_ns_tid"},
+			{Type: "int", Name: "child_pid"},
+			{Type: "int", Name: "child_ns_pid"},
+			{Type: "unsigned long", Name: "child_start_time"},
```
rafaeldtinoco, line 10595: Since its a new event, we can control different arguments for the signal ebpf programs (I think this is what was requested in the other review as well).

pkg/events/core.go
```diff
@@ -10528,6 +10538,120 @@ var CoreEvents = map[ID]Definition{
 		},
 	},
 	//
+	// Begin of Signal Events (Control Plane)
+	//
+	SignalCgroupMkdir: {
```
rafaeldtinoco, line 10543: I tested the existing events and they look fine.

pkg/ebpf/probes/probe_group.go
```diff
@@ -206,6 +206,11 @@ func NewDefaultProbeGroup(module *bpf.Module, netEnabled bool) (*ProbeGroup, err
 		ModuleLoad:                 NewTraceProbe(RawTracepoint, "module:module_load", "tracepoint__module__module_load"),
 		ModuleFree:                 NewTraceProbe(RawTracepoint, "module:module_free", "tracepoint__module__module_free"),
 		LayoutAndAllocate:          NewTraceProbe(KretProbe, "layout_and_allocate", "trace_ret_layout_and_allocate"),
+		SignalCgroupMkdir:          NewTraceProbe(RawTracepoint, "cgroup:cgroup_mkdir", "cgroup_mkdir_signal"),
```
rafaeldtinoco, line 209: In the future we can try to split the signal events definitions and signal events probe group from the core (not doing it now to make things easier).

pkg/ebpf/controlplane/controller.go
```diff
@@ -8,78 +8,54 @@ import (
...
 		cgroupManager:  cgroupManager,
 		enrichEnabled:  enrichEnabled,
 	}
+
 	p.signalBuffer, err = bpfModule.InitPerfBuf("signals", p.signalChan, p.lostSignalChan, 1024)
 	if err != nil {
 		return nil, err
 	}
 
-	p.probeGroup = probes.NewProbeGroup(bpfModule, map[probes.Handle]probes.Probe{
```
rafaeldtinoco, line 52: Let tracee to deal with event probes/dependencies initializations.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -6010,9 +6013,74 @@ int cgroup_rmdir_signal(struct bpf_raw_tracepoint_args *ctx)
...
+int sched_process_fork_signal(struct bpf_raw_tracepoint_args *ctx)
+{
+    controlplane_signal_t *signal = init_controlplane_signal();
+    if (unlikely(signal == NULL))
+        return 0;
+
+    long ret = 0;
+    program_data_t p = {};
+    if (!init_program_data(&p, ctx))
+        return 0;
```
rafaeldtinoco, line 6031: not needed (this is just a draft)

AlonZivony (approved):
This review is only on the last commit.
Looks good to me.
You will probably need to pass the task start time of the task in all the signals. Maybe we can create some general values in the future which will be passed for all signals, similar to the normal events.

rafaeldtinoco:
> This review is only on the last commit.
> Looks good to me.
> You will probably need to pass the task start time of the task in all the signals. Maybe we can create some general values in the future which will be passed for all signals, similar to the normal events.

Cool. Thanks for taking a look. I'll close this PR and continue in this path, and force push the original PR once things are done.

PR #3409, opened 2023-08-22
fix: log level should match zap log priority

josedonizetti:

Fixex log level priority to match zap log level priority (debug has the lowest priority) https://pkg.go.dev/go.uber.org/zap@v1.25.0/zapcore#DebugLevel


geyslan (approved):
LGTM

Issue #3396, opened 2023-08-16
verbose log for cgroups_mkdir


josedonizetti:
we are logging an error for `cgroup_mkdir` logs which are not related to a container, this is polluting tracee's output, let's change the log level to `debug`. 

```
{"level":"error","ts":1691287215.0058243,"msg":"error processing control plane signal","error":"controlplane.(*Controller).processCgroupMkdir: failed to get lookup key 824648583784 from map containers_map: no such file or directory"}
```

PR #3410, opened 2023-08-22
fix: ignore error for cgroups that doesn't exist

josedonizetti:

Fix https://github.com/aquasecurity/tracee/issues/3396

When running tracee on gke we see a lot of errors due to deletion of groups that don't exist (probably because they are not related to containers), on this PR we ignore this error, and add a debug logging for whenever it happens.


geyslan (approved):
LGTM

yanivagman (approved):
LGTM

PR #3411, opened 2023-08-23
feat: add streams

josedonizetti:

This PR is part of the work to support a [v1 for GRPC](https://github.com/aquasecurity/tracee/issues/3208). It adds Streams support, streams represent a subscriber of tracee events, the subscriber can specific which policies to receive events from. Streams will be used by grpc, where each grpc client can create multiple streams for tracee to consume events.

### 2. Explain how to test it

The GRPC endpoint will later be implemented like:

```
func (s *TraceeService) StreamEvents(in *pb.StreamEventsRequest, grpcStream pb.TraceeService_StreamEventsServer) error {
	stream, err := s.tracee.Subscribe(in.Policies)
	if err != nil {
		return err
	}
	defer s.tracee.Unsubscribe(stream)

	mask := fmutils.NestedMaskFromPaths(in.GetMask().GetPaths())

	for e := range stream.ReceiveEvents() {
		eventProto := convertTraceeEventToProto(e)
		mask.Filter(eventProto)
		grpcStream.Send(eventProto)
	}

	return nil
}
```

### DIFF
--- a/pkg/ebpf/tracee.go
+++ b/pkg/ebpf/tracee.go
@@ -42,6 +42,7 @@ import (
 	"github.com/aquasecurity/tracee/pkg/pcaps"
 	"github.com/aquasecurity/tracee/pkg/policy"
 	"github.com/aquasecurity/tracee/pkg/signatures/engine"
+	"github.com/aquasecurity/tracee/pkg/streams"
 	"github.com/aquasecurity/tracee/pkg/utils"
 	"github.com/aquasecurity/tracee/pkg/utils/proc"
 	"github.com/aquasecurity/tracee/pkg/utils/sharedobjs"
@@ -116,6 +117,8 @@ type Tracee struct {
 	// Specific Events Needs
 	triggerContexts trigger.Context
 	readyCallback   func(gocontext.Context)
+	// Streams
+	streamsManager *streams.StreamsManager
 }
 
 func (t *Tracee) Stats() *metrics.Stats {
@@ -215,6 +218,7 @@ func New(cfg config.Config) (*Tracee, error) {
 		capturedFiles:   make(map[string]int64),
 		eventsState:     GetEssentialEventsList(),
 		eventSignatures: make(map[events.ID]bool),
+		streamsManager:  streams.NewStreamsManager(),
 	}
 
 	// Initialize capabilities rings soon
@@ -1432,6 +1436,11 @@ func updateCaptureMapFile(fileDir *os.File, filePath string, capturedFiles map[s
 
 // Close cleans up created resources
 func (t *Tracee) Close() {
+	// clean up (unsubscribe) all streams connected if tracee is done
+	if t.streamsManager != nil {
+		t.streamsManager.Close()
+	}
+
 	if t.probes != nil {
 		err := t.probes.DetachAll()
 		if err != nil {
@@ -1735,4 +1744,35 @@ func (t *Tracee) ready(ctx gocontext.Context) {
 //go:noinline
 func (t *Tracee) triggerMemDumpCall(address uint64, length uint64, eventHandle uint64) error {
 	return nil
+}
+
+// SubscribeAll returns a stream subscribed to all policies
+func (t *Tracee) SubscribeAll() *streams.Stream {
+	return t.subscribe(policy.AllPoliciesOn)
+}
+
+// Subscribe returns a stream subscribed to selected policies
+func (t *Tracee) Subscribe(policyNames []string) (*streams.Stream, error) {
+	var policyMask uint64
+
+	for _, policyName := range policyNames {
+		p, err := t.config.Policies.LookupByName(policyName)
+		if err != nil {
+			return nil, err
+		}
+		utils.SetBit(&policyMask, uint(p.ID))
+	}
+
+	return t.subscribe(policyMask), nil
+}
+
+func (t *Tracee) subscribe(policyMask uint64) *streams.Stream {
+	// TODO: the channel size matches the pipeline channel size,
+	// but we should make it configurable in the future.
+	return t.streamsManager.Subscribe(policyMask, 10000)
+}
+
+// Unsubscribe unsubscribes stream
+func (t *Tracee) Unsubscribe(s *streams.Stream) {
+	t.streamsManager.Unsubscribe(s)
 }
--- a/pkg/policy/policies.go
+++ b/pkg/policy/policies.go
@@ -176,16 +176,27 @@ func (ps *Policies) Delete(id int) error {
 	return nil
 }
 
-func (ps *Policies) Lookup(id int) (*Policy, error) {
+// LookupById returns a policy by ID.
+func (ps *Policies) LookupById(id int) (*Policy, error) {
 	if !isIDInRange(id) {
 		return nil, PoliciesOutOfRangeError(id)
 	}
 
 	p := ps.policiesArray[id]
 	if p == nil {
-		return nil, PolicyNotFoundError(id)
+		return nil, PolicyNotFoundByIDError(id)
 	}
 	return p, nil
+}
+
+// LookupByName returns a policy by name.
+func (ps *Policies) LookupByName(name string) (*Policy, error) {
+	for p := range ps.Map() {
+		if p.Name == name {
+			return p, nil
+		}
+	}
+	return nil, PolicyNotFoundByNameError(name)
 }
 
 // MatchedNames returns a list of matched policies names based on
--- a/pkg/streams/streams.go
+++ b/pkg/streams/streams.go
@@ -1,0 +1,114 @@
+package streams
+
+import (
+	"context"
+	"sync"
+
+	"github.com/aquasecurity/tracee/types/trace"
+)
+
+// Stream is a stream of events
+type Stream struct {
+	// policy mask is a bitmap of policies that this stream is interested in
+	policyMask uint64
+	// events is a channel that is used to receive events from the stream
+	events chan trace.Event
+}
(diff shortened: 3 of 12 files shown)


pkg/streams/streams.go
```diff
@@ -0,0 +1,108 @@
...
+	if s.shouldIgnorePolicy(event) {
+		return
+	}
+
+	// currently we are blocking if the channel is full,
+	// though I think we should instead drop events, because if you have
+	// two streams, one should not stop the other from receiving events,
+	// if they are not fast enough. But there also use cases where blocking
+	// might be favor, so th idea is to later make this configurable.
+	// TODO: allow this to be configurable (drop/block) (josedonizetti)
```
geyslan, line 36: awesome drop/block idea.

pkg/streams/streams.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// Publish publishes an event to all streams
+func (sm *StreamsManager) Publish(ctx context.Context, event trace.Event) {
+	sm.mutex.Lock()
+	defer sm.mutex.Unlock()
+
+	for stream := range sm.subscribers {
+		stream.publish(ctx, event)
+	}
```
geyslan, line 97: This loop can block, right? How about each stream having its publish() triggered in its own goroutine?

pkg/streams/streams.go
```diff
@@ -0,0 +1,108 @@
...
+
+// Unsubscribe removes a stream from the manager
+func (sm *StreamsManager) Unsubscribe(stream *Stream) {
+	sm.mutex.Lock()
+	defer sm.mutex.Unlock()
+
+	// check if stream is subscribed
+	if _, ok := sm.subscribers[stream]; ok {
+		delete(sm.subscribers, stream)
+		close(stream.events)
```
geyslan, line 86: Maybe close first and delete later? Wondering if another goroutine could have started using the StreamsManager (somehow) and continue with an already deleted reference even if closed. Just being overly cautious. 🤔

pkg/streams/streams.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+}
+
+// Close closes all streams
+func (sm *StreamsManager) Close() {
+	sm.mutex.Lock()
+	defer sm.mutex.Unlock()
+	for stream := range sm.subscribers {
+		delete(sm.subscribers, stream)
+		close(stream.events)
```
geyslan, line 106: Ditto.

geyslan (approved):
LGTM 🚀

pkg/ebpf/tracee.go
```diff
@@ -42,15 +42,17 @@ import (
...
 	"github.com/aquasecurity/tracee/pkg/utils/sharedobjs"
 	"github.com/aquasecurity/tracee/types/trace"
 )
 
 const (
-	pkgName          = "tracee"
-	maxMemDumpLength = 127
+	pkgName            = "tracee"
+	maxMemDumpLength   = 127
+	allPoliciesEnabled = 0xffffffffffffffff
```
rafaeldtinoco, line 55: I had this in my PR:

image

and then I also replaced:

image

What do you think ?

pkg/streams/streams.go
rafaeldtinoco: Like we've spoken on teams, I believe each stream should have its own limited (size set by user ?) circular buffer and the events should come from this circular buffer. This would allow you to drop published events in the tail of the buffer if they were not consumed fast enough (which can happen depending on the network throughput/bandwidth and how fast clients consume the streams).

pkg/streams/streams.go
rafaeldtinoco: I feel that encapsulation here is half baked (with manager closing the stream channel). I believe Unsubscribe should remove the stream from the manager, but what if others are still using it (I believe Geyslan thought the same). I would encapsulate entirely Stream and the StreamManager, subscbribe and unsubscribe a stream from the manager but without touching the stream. For the stream, if multiple threads are using it, you can even keep a "refcount" and only allow channel closure when refcount is 0, or anything like it (if ever needed).

josedonizetti: I don't think so! When considering channel semantics, we can always read from a closed channel, but we can never write to a closed channel, or the runtime panics. Considering the subscribers map is the state to know if we publish to a channel or not, we first stop publishing (writing), and then we notify readers, there won't be further data on the channel by closing it, the stream interface exposes a read only channel for this reason specially.

pkg/streams/streams.go
josedonizetti: Same answer as above.

rafaeldtinoco (commented):
Overall I like this change a lot. Left some comments on encapsulation and minor stuff.

pkg/policy/policies.go
```diff
@@ -176,18 +176,29 @@ func (ps *Policies) Delete(id int) error {
...
 	p := ps.policiesArray[id]
 	if p == nil {
-		return nil, PolicyNotFoundError(id)
+		return nil, PolicyNotFoundByIDError(id)
 	}
 	return p, nil
 }
 
+// LookupByName returns a policy by name.
+func (ps *Policies) LookupByName(name string) (*Policy, error) {
```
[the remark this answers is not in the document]
josedonizetti: Indeed, a good change for later, I'll create an item about it on the API task so we can improve it.

pkg/streams/streams.go
josedonizetti: Agreed, I'll create an item on the API server task to improve it later.

pkg/streams/streams.go
josedonizetti: Can you explain further? Because it doesn't seem like a benefit for me, blocking will always happen at some point because we are not using the unblocking  semantics to send events. Adding more channels/goroutines will not change it, blocking will still happen, and when all channel blocks this pressure returns to the pipeline. That is why there are comments about dropping events and using a queue per stream.

pkg/ebpf/tracee.go
josedonizetti: Nice! Is this in a commit I can cherry-pick? Or do you mind if I do the change here?

pkg/streams/streams.go
rafaeldtinoco: If you think its good like this feel free to move on, I have no blockers.

josedonizetti: @rafaeldtinoco when I wrote the answer your comment wasn't appearing, let me read it, the first answer was for @geyslan comment about closing before.

pkg/streams/streams.go
geyslan: > That is why there are comments about dropping events and using a queue per stream.

This is the right path already discussed 👍🏼. Ignore my comment above.

pkg/streams/streams.go
josedonizetti: > I feel that encapsulation here is half baked (with manager closing the stream channel). I believe Unsubscribe should remove the stream from the manager, but what if others are still using it (I believe Geyslan thought the same). I would encapsulate entirely Stream and the StreamManager, subscbribe and unsubscribe a stream from the manager but without touching the stream. For the stream, if multiple threads are using it, you can even keep a "refcount" and only allow channel closure when refcount is 0, or anything like it (if ever needed).

@rafaeldtinoco But shouldn't the multiple goroutines (threads) know a channel won't be receiving data anymore? Or else the routine my keep running thinking events will be sent, but the publisher was removed, no data is coming there anymore. So closing works a signal, that once you drain the latest events from this steam either with `e,ok <-` or with `range`, you can stop the work, and close the routine.

geyslan: > I feel that encapsulation here is half baked (with manager closing the stream channel). I believe Unsubscribe should remove the stream from the manager, but what if others are still using it (I believe Geyslan thought the same).

Yeah I thought the same indeed, but as @josedonizetti remembered us, it seems that StreamsManager will be used in a very controlled area (not to be expanded in the codebase), so I agree that deleting it first will automatically remove it for writing which occurs only here:

Screenshot from 2023-08-23 14-47-35

Where we have a mutex control.

And closing it will make the channel only readable. So, yes I also have no blockers. 👍🏼

rafaeldtinoco: Alright. Fair enough then.

pkg/ebpf/tracee.go
rafaeldtinoco: Not at all, put this change in your PR, I'll rebase once yours is merged.

geyslan (approved):

pkg/streams/streams.go
josedonizetti: As we spoke on the call, it will be implemented in a future pr.

pkg/ebpf/tracee.go
```diff
@@ -1736,3 +1745,34 @@ func (t *Tracee) ready(ctx gocontext.Context) {
...
 }
+
+// SubscribeAll returns a stream subscribed to all policies
+func (t *Tracee) SubscribeAll() *streams.Stream {
+	return t.subscribe(policy.AllPoliciesOn)
+}
+
+// Subscribe returns a stream subscribed to selected policies
+func (t *Tracee) Subscribe(policyNames []string) (*streams.Stream, error) {
+	var policyMask uint64
```
yanivagman, line 1756: Not directly related to this PR, but it might be nice if we will define a new type for the policy mask instead of using uint64. This way, it will be easier to modify the number of policies we have, if and when we will decide to do it.

yanivagman (approved):
LGTM

PR #3414, opened 2023-08-24
refactor: getStackAddresses doesn't return an err

josedonizetti:

Minor refactor, `getStackAddresses` doesn't return na error.


geyslan (approved):
LGTM

Issue #3403, opened 2023-08-22
docs: syscalls as Tracee events


AnaisUrlichs:
Just to clarify, I can use syscalls as events in the Tracee Policy?

e.g. https://aquasecurity.github.io/tracee/v0.17/docs/events/builtin/syscalls/open/

```
spec:
    scope:
      - container
    rules:
      - event: open
```

geyslan:
Yep, take a look at [examples/policies/openat_args_pahtname.yaml](https://github.com/aquasecurity/tracee/blob/main/examples/policies/openat_args_pahtname.yaml)

```
❯ cat examples/policies/openat_args_pahtname.yaml
apiVersion: aquasecurity.github.io/v1beta1
kind: TraceePolicy
metadata:
  name: openat-args-pathname
  annotations:
    description: traces openat under /tmp/*
spec:
  scope: 
    - global
  rules:
    - event: openat
      filters:
        - args.pathname=/tmp*
```

AnaisUrlichs:
Cool, thank you @geyslan

AnaisUrlichs:
Will leave this issue open until I documented it

yanivagman:
> Will leave this issue open until I documented it

Which documentation do you want to add other than https://aquasecurity.github.io/tracee/v0.17/docs/events/builtin/syscalls/open/?

AnaisUrlichs:
@yanivagman I was planning on detailing in the Policies section where people can find information on the events they can add to the policy -- I might be learning this right now but a completely new users will not know -- will tag this issue in the PR

Issue #3398, opened 2023-08-18
docs: Why does it say "tracee will start with a sane default"


AnaisUrlichs:
What is sane default? I don't understand https://aquasecurity.github.io/tracee/v0.17/docs/events/overview/

josedonizetti:
Hmm, the idea is that tracee starts with a set of [default events](https://aquasecurity.github.io/tracee/v0.17/docs/events/overview/#default-events) to help the user. But unfortunately the default set of events currently is different if you using the binary/docker or kubernetes, we want to fix it after adding a few higher level events, which will take over as the default events.

AnaisUrlichs:
Thank you for the explanation -- to me the "sane" part in the sentence did not make sense

josedonizetti:
@AnaisUrlichs  gotcha, can you do a PR to fix/improve it? :)

PR #3416, opened 2023-08-25
docs: changes to the policies documentation

AnaisUrlichs:
https://github.com/aquasecurity/tracee/issues/3403

and

https://github.com/aquasecurity/tracee/issues/3398


docs/docs/policies/rules.md
```diff
@@ -7,6 +7,62 @@ Rules determine which events a policy should trace.
 An event can match all occurrences of events for a specific scope, or specific events depending on its filters.
 Events support three types of filters: `context`, `arguments` and `return value`. 
 
+### Type of Events
+
+You can add as events either of the following:
```
josedonizetti, line 12:
```suggestion
You can add events as either of the following:
```
?

docs/docs/policies/rules.md
```diff
@@ -7,6 +7,62 @@ Rules determine which events a policy should trace.
...
+spec:
+	scope:
+	    - global
+	rules:
+	    event: net_packet_ipv4
+```
+
+**[A behavioural Signature](../events/builtin/signatures.md)**
+
+To specified one of the behavioral signatures as event, simply provide the name of the signature in the YAML manifest of the Tracee Policy:
```
josedonizetti, line 42:
```suggestion
To specify one of the behavioral signatures as an event,
```
???

docs/docs/policies/rules.md
```diff
@@ -7,6 +7,62 @@ Rules determine which events a policy should trace.
...
+
+```bash
+spec:
+	scope:
+	    - global
+	rules:
+	    event: TRC-101
```
josedonizetti, line 49:
```suggestion
	    event: TRC-101
```
TRCs are internal ids, we need to use the event name, soon TRCs won't be exposed anymore. 
eg: https://github.com/aquasecurity/tracee/blob/5d12c73031aaf8edc41552fdbbf091a71b6642fa/signatures/golang/proc_mem_access.go#L32

josedonizetti (changes requested):

docs/docs/policies/rules.md
AnaisUrlichs: What do you mean by ` soon TRCs won't be exposed anymore`?

josedonizetti: We will remove the TRCs ids from tracee.

josedonizetti: @AnaisUrlichs This one still needs to be fixed before we can merge this PR

AnaisUrlichs: I changed it

docs/docs/policies/index.md
```diff
@@ -43,6 +43,6 @@ This policy applies to any workload (global) and will log the dropped_executable
 While specifying event filters is optional, policies must have the `name`, `description`, `scope` and `rules` fields.
 
 !!! Note
-    A current limitation is that only one rule can be defined per any event type in a policy
+    Note that only one rule can be defined per any event type in a policy
```
AnaisUrlichs, line 46: I don't see why this is a limitation -- I know users should be aware of it but I would not call it a limitation

AnaisUrlichs:
@yanivagman @itaysk I have added the names for the behavioral signatures -- I think they are all added correctly but for the following, I could not find the right name && the list provided by `tracee list` misses one. While the table in the docs has 31 events  -- the list in the rules section of the CLI only has 30 events 

this one currently is wrong and misses the event name:
Screenshot 2023-09-01 at 12 05 36

Note the name here for the event on the left is just the placeholder

docs/docs/policies/index.md
yanivagman: Ok, but I would add the word "currently" since we want to support more than one rule per event in the future

yanivagman:
> @yanivagman @itaysk I have added the names for the behavioral signatures -- I think they are all added correctly but for the following, I could not find the right name && the list provided by `tracee list` misses one. While the table in the docs has 31 events -- the list in the rules section of the CLI only has 30 events
> 
> this one currently is wrong and misses the event name: Screenshot 2023-09-01 at 12 05 36
> 
> Note the name here for the event on the left is just the placeholder

This is its name: https://github.com/aquasecurity/tracee/blob/main/signatures/golang/kubernetes_certificate_theft_attempt.go#L31

Not sure why it is not printed in the list

docs/docs/policies/index.md
AnaisUrlichs: oki, sounds good, will add it

docs/docs/events/builtin/signatures.md
```diff
@@ -1,36 +1,36 @@
...
-TRC-1031 | System request key configuration modification | An attempt to modify and activate the System Request Key configuration file was detected. The system request key allows immediate input to the kernel through simple key combinations. Adversaries may use this feature to immediately shut down or restart a system. With read access to kernel logs, host related information such as listing tasks and CPU registers may be disclosed and could be used for container escape.
+Name   | ID   |Description                               |Full Description
+-------|------|------------------------------------------|----
+stdio_over_socket | TRC-101 | Process standard input/output over socket detected | A process has its standard input/output redirected to a socket. This behaviour is the base of a  Reverse Shell attack, which is when an interactive shell being invoked from a target machine back to the attacker's machine, giving it interactive control over the target. Adversaries may use a Reverse Shell to retain control over a compromised target while bypassing security measures like network firewalls.
```
josedonizetti, line 5: @AnaisUrlichs since we are touching this part, I don't think we should expose the `TRC-*` ids anymore, since they are internal, and we won't expose it on the new event structure. @yanivagman thougths?

AnaisUrlichs: So you want me to remove the IDs? but it is useful to have them to find the section in the source code

josedonizetti: Yeah, i think removing makes sense, but let's wait yaniv feedback, my understanding is that we will not expose it on the event structure anymore, and our OSS signatures won't have this ID anymore, it will be something internal only, so we shouldn't trust on it to find anything. 

How are you using it? Didn't undestand.

yanivagman: Yes, we can remove the.
The name can be used as an id and also when searching in the sources

josedonizetti (approved):

PR #3423, opened 2023-08-29
Fix closed chan read

geyslan:
Close: https://github.com/aquasecurity/libbpfgo/issues/122

68f094814 **fix(ebpf): close perfbuf instead of just stop**
32a028c8c **fix(ebpf): closed channel read**

68f094814 **fix(ebpf): close perfbuf instead of just stop**

```
This was an old TODO comment removed by https://github.com/aquasecurity/tracee/pull/296/files#diff-82aa2eb77c3ce782b8ebfb70eaba8a1a5a8a2dce9f3349352a3a3c0f909478f1L604
```

32a028c8c **fix(ebpf): closed channel read**

```
This libbpfgo issue https://github.com/aquasecurity/libbpfgo/issues/122
it's actually a Tracee problem. It was caused by read attempts on closed
channels, when libbpgo had already closed them.
```

### 2. Explain how to test it

Run tracee and kill it via `^C`.

One must want to put a print call after the `!ok` check:

```go
func (t *Tracee) processLostEvents() {
	logger.Debugw("Starting processLostEvents goroutine")
	defer logger.Debugw("Stopped processLostEvents goroutine")

	for {
		select {
		case lost, ok := <-t.lostEvChannel:
			if !ok { // channel closed
				return
			}

			// Won't be printed as before.
			fmt.Printf("Lost %d events\n", lost)

			if err := t.stats.LostEvCount.Increment(lost); err != nil {
				logger.Errorw("Incrementing lost event count", "error", err)
			}
			logger.Warnw(fmt.Sprintf("Lost %d events", lost))

		// Since this is an end-state goroutine, it should be terminated only
		// when Tracee done channel is closed.
		case <-t.done:
			return
		}
	}
}
```

### DIFF
--- a/pkg/ebpf/bpf_log.go
+++ b/pkg/ebpf/bpf_log.go
@@ -191,16 +191,15 @@ func (t *Tracee) processBPFLogs(ctx context.Context) {
 				logger.Errorw("Incrementing BPF logs count", "error", err)
 			}
 
-		case lost := <-t.lostBPFLogChannel:
-			// When terminating tracee-ebpf the lost channel receives multiple "0 lost events" events.
-			// This check prevents those 0 lost events messages to be written to stderr until the bug is fixed:
-			// https://github.com/aquasecurity/libbpfgo/issues/122
-			if lost > 0 {
-				if err := t.stats.LostBPFLogsCount.Increment(lost); err != nil {
-					logger.Errorw("Incrementing lost BPF logs count", "error", err)
-				}
-				logger.Warnw(fmt.Sprintf("Lost %d ebpf logs events", lost))
+		case lost, ok := <-t.lostBPFLogChannel:
+			if !ok { // channel closed
+				continue
 			}
+
+			if err := t.stats.LostBPFLogsCount.Increment(lost); err != nil {
+				logger.Errorw("Incrementing lost BPF logs count", "error", err)
+			}
+			logger.Warnw(fmt.Sprintf("Lost %d ebpf logs events", lost))
 
 		case <-ctx.Done():
 			return
--- a/pkg/ebpf/tracee.go
+++ b/pkg/ebpf/tracee.go
@@ -1366,16 +1366,16 @@ func (t *Tracee) Run(ctx gocontext.Context) error {
 	t.ready(ctx)          // executes ready callback, non blocking
 	<-ctx.Done()          // block until ctx is cancelled elsewhere
 
-	// Stop perf buffers
+	// Close perf buffers
 
-	t.eventsPerfMap.Stop()
+	t.eventsPerfMap.Close()
 	if t.config.BlobPerfBufferSize > 0 {
-		t.fileWrPerfMap.Stop()
+		t.fileWrPerfMap.Close()
 	}
 	if pcaps.PcapsEnabled(t.config.Capture.Net) {
-		t.netCapPerfMap.Stop()
+		t.netCapPerfMap.Close()
 	}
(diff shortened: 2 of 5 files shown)


geyslan:
e2e test (1042) is green.

pkg/ebpf/bpf_log.go
```diff
@@ -191,16 +191,15 @@ func (t *Tracee) processBPFLogs(ctx context.Context) {
...
-			// This check prevents those 0 lost events messages to be written to stderr until the bug is fixed:
-			// https://github.com/aquasecurity/libbpfgo/issues/122
-			if lost > 0 {
-				if err := t.stats.LostBPFLogsCount.Increment(lost); err != nil {
-					logger.Errorw("Incrementing lost BPF logs count", "error", err)
-				}
-				logger.Warnw(fmt.Sprintf("Lost %d ebpf logs events", lost))
+		case lost, ok := <-t.lostBPFLogChannel:
+			if !ok { // channel closed
+				continue
```
yanivagman, line 196: Just for my understanding - if we continue here since the channel is closed, will we ever reach the ctx.Done() case below?

pkg/ebpf/tracee.go
```diff
@@ -1368,14 +1368,14 @@ func (t *Tracee) Run(ctx gocontext.Context) error {
 
 	// Stop perf buffers
```
yanivagman, line 1369: nit: Stop -> Close

yanivagman (approved):
LGTM,
If this fixes the mentioned libbpfgo issue, don't forget to close it there as well

josedonizetti:
@geyslan before merging I want to review this one, sorry I dind't have the time yet

pkg/ebpf/bpf_log.go
geyslan: @yanivagman Yes, since `<-ctx.Done()` is another select case clause. But your question made me ponder whether we should continue (putting a warning log) or return; I said this because the channel is controlled by another party and when closed it is the end of its useful life. @josedonizetti when you pick this for review, please take this in consideration.

josedonizetti: @geyslan Although it will get to the ctx done, if the ctx is not done and the channel is closed it will trigger a never blocking loop, because the select will always match on the closed channel, and continue, consuming CPU. When would this chan be closed?  Why do we care about the closing for `lostBPFLogChannel` but not for `bpfLogsChannel`?

geyslan: > Although it will get to the ctx done, if the ctx is not done and the channel is closed it will trigger a never blocking loop, because the select will always match on the closed channel, and continue, consuming CPU.

Indeed.

> When would this chan be closed?

They're finally closed by libbpfgo PerfBuffer.Stop() which is called by PerfBuffer.Close(), which in its turn is called after tracee's `ctx.Done()`:

Screenshot from 2023-09-05 08-06-00

> Why do we care about the closing for lostBPFLogChannel but not for bpfLogsChannel?

lostBPFLogChannel is not buffered.

geyslan: @josedonizetti your review opened my eyes to the correct cause of the "lost 0 events" output. Although the problem is mentioned in different goroutines, the cause is only in `processLostEvents`. Please take a look at this new PR: https://github.com/aquasecurity/tracee/pull/3438

geyslan:
Closed due to https://github.com/aquasecurity/tracee/pull/3438

PR #3430, opened 2023-09-01
Add enable disable policies

josedonizetti:

This PR is part of the work to support a [v1 for GRPC](https://github.com/aquasecurity/tracee/issues/3208).  It adds support to enable/disable policies globally. The GRPC endpoints will be implemented in a following PR. 

This is a temporary solution, the enable/disable policies in the future should be done by `policies.Policies`, and probably load and unload configuration from kernel layer. But to support a feature we are creating some code that handles it by dropping the events, instead of unloading a policy. Also I created a specific structure to handle it, to avoid mutex code scattered around the tracee struct.

### 3. Other comments

depends on https://github.com/aquasecurity/tracee/pull/3411


pkg/ebpf/events_pipeline.go
```diff
@@ -569,6 +569,14 @@ func (t *Tracee) sinkEvents(ctx context.Context, in <-chan *trace.Event) <-chan
 				continue
 			}
 
+			// is the policy globally disabled?
+			event.MatchedPoliciesUser = t.policiesManager.Match(event.MatchedPoliciesUser)
```
geyslan, line 573: A note to my future self @geyslan: this just drops disabled events at the final stage of the pipeline. As discussed offline, it will be replaced by a mechanism that also updates kernel space. Probably the policies versioning.

pkg/ebpf/policies_manager.go
```diff
@@ -0,0 +1,43 @@
...
+
+func newPoliciesManager() *policiesManager {
+	return &policiesManager{
+		mutex:      sync.Mutex{},
+		policyMask: 0,
+	}
+}
+
+// Match returns the matched policies
+func (p *policiesManager) Match(matchedPolicy uint64) uint64 {
```
geyslan, line 23: Perhaps Match(`matchedPolicies` uint64)?

pkg/ebpf/policies_manager_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/aquasecurity/tracee/pkg/policy"
+)
+
+func TestPoliciesManagerEnablePolicy(t *testing.T) {
+	policiesManager := &policiesManager{policyMask: 0}
+
+	for i := 0; i < 64; i++ {
```
geyslan, line 15: Here you can use `policy.MaxPolicies` instead of 64. We have plans to increase it.

pkg/ebpf/policies_manager_test.go
```diff
@@ -0,0 +1,81 @@
...
+		policiesManager.EnablePolicy(i)
+	}
+
+	assert.Equal(t, uint64(policy.AllPoliciesOn), policiesManager.policyMask)
+}
+
+func TestPoliciesManagerDisablePolicy(t *testing.T) {
+	policiesManager := &policiesManager{policyMask: policy.AllPoliciesOn}
+
+	for i := 0; i < 64; i++ {
```
geyslan, line 25: Ditto.

pkg/ebpf/policies_manager.go
```diff
@@ -0,0 +1,43 @@
...
+	"sync"
+
+	"github.com/aquasecurity/tracee/pkg/utils"
+)
+
+// policiesManager is a thread-safe mask of enabled policies
+type policiesManager struct {
+	mutex      sync.Mutex
+	policyMask uint64
+}
```
geyslan, line 13: I like the idea of `policesManager`. We'll be able to move to it all bpf logic that's scattered throughout the codebase.

We have a lot of policy settings being done at different tracee initialization steps that need to be brought into a handler like this. Please later take a look at the `TODO:`s inserted by this PR: https://github.com/aquasecurity/tracee/pull/3305/files

geyslan (approved):
LGTM

pkg/ebpf/policies_manager.go
```diff
@@ -0,0 +1,43 @@
...
+	defer p.mutex.Unlock()
+
+	utils.SetBit(&p.policyMask, uint(id))
+}
+
+// DisablePolicy disables the given policy
+func (p *policiesManager) DisablePolicy(id int) {
+	p.mutex.Lock()
+	defer p.mutex.Unlock()
+	utils.ClearBit(&p.policyMask, uint(id))
```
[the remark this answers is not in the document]
josedonizetti: NL?

yanivagman (commented):
This reminds me of the functionality you implemented in streams (being able to request specific policies) - what is the benefit of it being here as well?

Other than the above concern, I do like the idea of having "policy manager". We can use it to keep state about the currently loaded policies, both scope and rules, which can then be used to do dynamic changes when a policy is modified

josedonizetti:
As we spoke, this won't be needed on the mvp of this API.

PR #3431, opened 2023-09-01
Add enable/disable rule

josedonizetti:

This PR is part of the work to support a [v1 for GRPC](https://github.com/aquasecurity/tracee/issues/3208).  It adds support to enable/disable rules per policy globally. The GRPC endpoints will be implemented in a following PR. 

This is a temporary solution, the enable/disable rules in the future should by extensions code @rafaeldtinoco is currently creating/refactoring, and probably load and unload configuration from kernel layer. But to support a feature now, we are creating some code that handles it by dropping the events, instead of unloading it. Also I created a specific structure to handle it, to avoid mutex code scattered around the tracee struct.

### DIFF
--- a/pkg/ebpf/events_pipeline.go
+++ b/pkg/ebpf/events_pipeline.go
@@ -561,6 +561,13 @@ func (t *Tracee) sinkEvents(ctx context.Context, in <-chan *trace.Event) <-chan 
 				continue // might happen during initialization (ctrl+c seg faults)
 			}
 
+			// Is the rule disabled for the matched policies?
+			if !t.policyManager.IsRuleEnabled(event.MatchedPoliciesUser, events.ID(event.EventID)) {
+				logger.Debugw("event dropped because matched rule disabled", "event", event.EventName)
+				t.eventsPool.Put(event)
+				continue
+			}
+
 			// Only emit events requested by the user and matched by at least one policy.
 			id := events.ID(event.EventID)
 			event.MatchedPoliciesUser &= t.eventsState[id].Emit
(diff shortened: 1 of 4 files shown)


geyslan (approved):
It has the same logic of #3430. So as reviewed there, this LGTM too.

yanivagman:
Didn't we agree on having the API enable/disable a rule and not an event?

josedonizetti:
@yanivagman I did the change to have enable/disable rule per policy, can you review? pls.

pkg/ebpf/events_pipeline.go
```diff
@@ -561,6 +561,13 @@ func (t *Tracee) sinkEvents(ctx context.Context, in <-chan *trace.Event) <-chan
 				continue // might happen during initialization (ctrl+c seg faults)
 			}
 
+			// Is the rule disabled for the matched policies?
+			if !t.policyManager.IsRuleEnabled(event.MatchedPoliciesUser, events.ID(event.EventID)) {
+				logger.Debugw("event dropped because it is disabled", "event", event.EventName)
```
yanivagman, line 566:
```suggestion
				logger.Debugw("event dropped because matched rule is disabled", "event", event.EventName)
```

pkg/ebpf/tracee.go
```diff
@@ -119,6 +119,8 @@ type Tracee struct {
 	readyCallback   func(gocontext.Context)
 	// Streams
 	streamsManager *streams.StreamsManager
+	// policyManager manages rules enabled/disable per policy
```
yanivagman, line 122: In the near future, it will also mange policy state, including state, so maybe we better not restrict the comment to rule enable/disable

yanivagman (approved):
LGTM
See nit comment

PR #3433, opened 2023-09-03
Fix(capture): fix verifier issue with elf capture

AlonZivony:

Force the verifier to use the right registers for read, avoiding verifier errors about reading of possible negative value with bpf_probe_read.


yanivagman (approved):
LGTM

PR #3436, opened 2023-09-05
Revert "fix: print_mem_dump fails on missing symbol (#3384)"

AlonZivony:

This reverts commit 1a47a4e4e3b565d6e06169696c533c63d94c3c1f. This is done because the commit cause the "symbols_loaded" event to fail and also cause many error messages that weren't seen before.


rafaeldtinoco:
What brakes the E2E tests is the following change:

image

But the change Nadav did to the event definition ksymbols dependency isn't good IMO (and can be mitigated by a very simple change).

I'm going to merge this reversion and suggest a PR that fixes: https://github.com/aquasecurity/tracee/issues/3382.

Im going to leave the issue: https://github.com/aquasecurity/tracee/issues/3383 opened to fix it in a further moment (as the behavior does not break anything and it's a "good to have" change).

Issue #2870, opened 2023-03-14
New Event Strucutre


itaysk:
#2355 changed the primary user experience of Tracee to be event oriented (previously events were considered internal and hidden from the user). Therefore:
1. The event schema needs to be formalized and stabilized. Since it's no longer internal.
2. The event structure needs to be generalized. Since events are will now be used for detections, captures and more.

I'm summarizing here a few other discussion and their conclusions. This is how the new event schema will look like:

1. `name` - will become the event identifier, instead of "event id".
2. `metadata`
    1. `description`
    2. `severity`
    4. `misc` - `map[string]interface{}`
 3. `context`
     1. all the current event context attributes
     2. `container` - all container related context
     3. `kubernetes` - all k8s related context
      5. stackTraces - if enabled, will be here
      6. `syscall` - moved here
      7. `flags` - moved here (consider renaming)
 4. args (consider renaming)
     1. every event attribute
      2. returnValue (if relevant will appear here)

We also discussed versioning the event schema, but not including the version with each event, for efficiency.

itaysk:
while this represent the complete event schema, I also think we should try to emit minimal events. For example, I suppose most people won't need the full description with every event, or the full stack trace. We already allow conditionally including some elements (stack trace, syscall, exec-env), and I think we need to expand this control to all event fields. for discussing this I'll create another issue.

NDStrahilevitz:
@itaysk Do you think we should maybe use the `protocol.Event` feature for this (see `types/protocol`)? We could add a `ContentType` header to differentiate events. Or do we want to stick with `trace.Event` with split up context fields?

itaysk:
Actually I'm not very familiar with how protocol is used now. I thought it's supposed to become obsolete in the unified binary approach. Can you elaborate on your proposal or perhaps add an example?

NDStrahilevitz:
The protocol is more a way to differentiate different payloads going into the rules engine.
For example in CNDR it is used to define the ability to receive events sent directly from CNDR into the signature (for example to set data storage in the signature).
The same way the protocol can define a different payload for signature inputs, it can be used to define different event payloads with different shapes (for example, before we changed detections to be `trace.Event` we could have defined it it to be a different payload in the `protocol.Event` instead).
I am not sure if we _want_ to currently support different payloads for events going into the engine (since the engine is now supposed to be internal).

NDStrahilevitz:
May I suggest we move the event definition to a `proto` file while we're at it?
[...]

PR #3437, opened 2023-09-05
feat: add version and description to definition

josedonizetti:

This PR is part of the work to support a [v1 for GRPC](https://github.com/aquasecurity/tracee/issues/3208), and also the https://github.com/aquasecurity/tracee/issues/2870, because the new event structure won't have metadata, and instead static data from the event will come from the definition, so users will be able to query the grpc endpoint to get a definition, but there is two important fields missing from the definition:

- the event description
- the event version, which will be bumped if we change the Params, to inform users that the event schema changed

### DIFF
--- a/pkg/events/definition.go
+++ b/pkg/events/definition.go
@@ -8,6 +8,8 @@ type Definition struct {
 	id           ID // TODO: use id ?
 	id32Bit      ID
 	name         string
+	version      Version
+	description  string
 	docPath      string // Relative to the 'doc/events' directory
 	internal     bool
 	syscall      bool
@@ -20,6 +22,8 @@ func NewDefinition(
 	id ID,
 	id32Bit ID,
 	name string,
+	version Version,
+	description string,
 	docPath string,
 	internal bool,
 	syscall bool,
@@ -31,6 +35,8 @@ func NewDefinition(
 		id:           id,
 		id32Bit:      id32Bit,
 		name:         name,
+		version:      version,
+		description:  description,
 		docPath:      docPath,
 		internal:     internal,
 		syscall:      syscall,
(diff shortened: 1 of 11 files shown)


josedonizetti:
@yanivagman @rafaeldtinoco what do you think? If you agree, I'll add a `v1.0.0` version to every event defined on the core.

pkg/events/definition.go
```diff
@@ -5,9 +5,18 @@ import (
...
+	id32Bit ID
+	name    string
+
+	// 1. Major field is bumped whenever some data the event used to have was changed
+	// e.g. a field was renamed or removed
+	// 2. Minor field is bumped whenever a non breaking change occurs
+	// e.g. a new field was added to the event
+	// 3. Patch field is bumped whenever something is changed in the way the event works internally
+	// e.g. some bug was fixed in the code
+	version      string
```
yanivagman, line 18: Maybe we should set the version to be a struct so we can enforce the use of semantic versioning:
```
type version struct {
    int major
    int minor
    int patch
}
```

WDYT?

josedonizetti: @yanivagman at first I thought a string is simpler, and we could validate at the event definition creation, but it doesn't block someone from creating it without the constructor and have a bad version, so if going with a struct, maybe let's use the one from the semver package direclty? https://github.com/Masterminds/semver/blob/master/version.go#L47

josedonizetti: AH, I remembered why I chose not to validate now, because the signatures are using their own versioning schema https://github.com/aquasecurity/tracee/blob/main/signatures/golang/docker_abuse.go#L27, enforcing semver now will require research to change their signatues, is it something we want to do now?

yanivagman (approved):
LGTM

Issue #3441, opened 2023-09-05
Add argv to security_bprm_check


AlonZivony:
To better use the `security_bprm_check` event, it will be useful to have the `argv` of the execution and maybe even the environment variables.

PR #3442, opened 2023-09-05
feat: add argv to security_bprm_check

AlonZivony:

To allow deeper analysis of the executed program before the execution, a new field for the argv given is added.

fix: #3441

### DIFF
--- a/docs/docs/events/builtin/extra/security_bprm_check.md
+++ b/docs/docs/events/builtin/extra/security_bprm_check.md
@@ -1,0 +1,30 @@
+# security_bprm_check
+
+## Intro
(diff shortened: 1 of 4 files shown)


docs/docs/events/builtin/extra/security_bprm_check.md
```diff
@@ -0,0 +1,30 @@
+# security_bprm_check
```
yanivagman, line 1: Thanks for adding this.
For it to be exposed in the docs, you need to add it to the navigation bar as well.

yanivagman (approved):
LGTM

PR #3443, opened 2023-09-05
fix(definitions): ksymbols dependencies handled wrongly

rafaeldtinoco:
commit fad5fc043 (HEAD -> ksymboldepsregression, rafaeldtinoco/ksymboldepsregression)
Date:   Tue Sep 5 16:33:13 2023

    fix(event): enable ksymbols engine for PrintMemDump
    
    Commit 1a47a4e4e attempted to fix this issue by reverting some changes
    made during the event definitions (and dependencies) refactor. That
    change caused some regressions because of another fix made together,
    but, since it was reverted, this commit fixes the same issue with a much
    simpler approach (at least until the event definitions refactor work is
    finished).
    
    Fixes: #3382

### DIFF
--- a/pkg/events/core.go
+++ b/pkg/events/core.go
@@ -10344,7 +10344,13 @@ var CoreEvents = map[ID]Definition{
 			ids: []ID{
 				DoInitModule,
 			},
-			kSymbols: []KSymbol{},
+			kSymbols: []KSymbol{
+				// Special case for this event: Single symbol, common to all kernel versions. Placed
+				// here so the ksymbols engine is always enabled, during tracee startup. The symbols
+				// are resolved dynamically, during runtime depending on the arguments passed to
+				// the event.
+				{symbol: "_stext", required: true},
+			},
 			capabilities: Capabilities{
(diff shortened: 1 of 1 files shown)


yanivagman:
logic looks similar to #3436 - any other thing done here than just reverting the recently merged commit?

rafaeldtinoco:
> logic looks similar to #3436 - any other thing done here than just reverting the recently merged commit?

Just put this on draft. I'm still testing things.

rafaeldtinoco:
What brakes the E2E tests is the following change:

image

But the change Nadav did to the event definition ksymbols dependency isn't good IMO (and can be mitigated by a very simple change).

I'm going to merge this reversion and suggest a PR that fixes: https://github.com/aquasecurity/tracee/issues/3382.

Im going to leave the issue: https://github.com/aquasecurity/tracee/issues/3383 opened to fix it in a further moment (as the behavior does not break anything and it's a "good to have" change).

yanivagman (approved):
LGTM

PR #3447, opened 2023-09-06
Add definition proto

josedonizetti:
### 1. Explain what the PR does

This PR is part of the work to support a [v1 for GRPC](https://github.com/aquasecurity/tracee/issues/3208), it adds the `EventDefinition` proto, and the `GetEventDefinition` rpc call.

PR #3450, opened 2023-09-06
Add definition grpc endpoint

josedonizetti:

This PR is part of the work to support a [v1 for GRPC](https://github.com/aquasecurity/tracee/issues/3208), it add the implementation to the `GetEventDefinition` rpc call.  

### 2. Explain how to test it

```
func main() {
	var opts []grpc.DialOption

	if err != nil {
	}

	client := pb.NewTraceeServiceClient(conn)

	definitions, err := client.GetEventDefinition(
		context.Background(),
		&pb.GetEventDefinitionRequest{Name: "openat"},
	)
	if err != nil {
		log.Fatal(err)
	}

        for _, definition := range definitions.Definitions {
		v := fmt.Sprintf("%d.%d.%d", definition.Version.Major, definition.Version.Minor, definition.Version.Patch)
		fmt.Println("Id:", definition.Id)
		fmt.Println("Name:", definition.Name)
		fmt.Println("Description:", definition.Description)
		fmt.Println("Version:", v)
		fmt.Printf("Tags: %v\n", definition.Tags)
		fmt.Printf("Params: %v\n", definition.Params)
	}
}
```

### 3. Other comments

depends on https://github.com/aquasecurity/tracee/pull/3447 and https://github.com/aquasecurity/tracee/pull/3437

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/Masterminds/sprig/v3 v3.2.3
 	github.com/aquasecurity/libbpfgo v0.4.8-libbpf-1.2.0.0.20230509162948-80f41e18e690
 	github.com/aquasecurity/libbpfgo/helpers v0.4.6-0.20230321190037-f591a2c5734f
-	github.com/aquasecurity/tracee/types v0.0.0-20230817130859-7d4b8c20396c
+	github.com/aquasecurity/tracee/types v0.0.0-20230912172206-3c8a64ecec2c
 	github.com/containerd/containerd v1.7.0
 	github.com/docker/docker v23.0.5+incompatible
 	github.com/golang/protobuf v1.5.3
--- a/pkg/server/grpc/tracee.go
+++ b/pkg/server/grpc/tracee.go
@@ -2,7 +2,9 @@ package grpc
 
 import (
 	"context"
+	"fmt"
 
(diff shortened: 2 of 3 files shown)


go.mod
```diff
@@ -173,3 +173,5 @@ require (
 )
 
 replace github.com/kubernetes/cri-api => k8s.io/cri-api v0.23.5-rc.0
+
+replace github.com/aquasecurity/tracee/types v0.0.0-20230817130859-7d4b8c20396c => ./types
```
josedonizetti, line 177: TODO: replace after https://github.com/aquasecurity/tracee/pull/3447 was merged

pkg/server/grpc/tracee.go
```diff
@@ -11,6 +13,30 @@ type TraceeService struct {
 	pb.UnimplementedTraceeServiceServer
 }
 
+func (s *TraceeService) GetEventDefinition(ctx context.Context, in *pb.GetEventDefinitionRequest) (*pb.GetEventDefinitionResponse, error) {
+	id, ok := events.Core.GetDefinitionIDByName(in.Name)
+	if !ok {
+		return nil, fmt.Errorf("event %s not found", in.Name)
+	}
```
yanivagman, line 20: We also need to support the case where no name is given, which means we want to get all event definitions

yanivagman (approved):
LGTM

Issue #3454, opened 2023-09-10
Document ARM64 quickstart


yanivagman:
The current documentation mentions a docker image with the latest tag, however, this image is for x86 architecture only. We should document arm64 image usage as well

geyslan:
Additional info:

- https://github.com/aquasecurity/tracee/pull/3464
- https://github.com/aquasecurity/libbpfgo/blob/bd31ff0bdba11491d7d31df7675274e6829b769f/docs/Vagrantfile.md#darwin

PR #3456, opened 2023-09-11
docs: running tracee docker container on amd64 and arm64

AnaisUrlichs:
Regarding the issue detailed in #3454

### DIFF
--- a/docs/getting-started/docker-quickstart.md
+++ b/docs/getting-started/docker-quickstart.md
@@ -1,6 +1,20 @@
 # Docker Quickstart
 
-You can easily start experimenting with Tracee using the Docker image as follows:
+This section details how you can run Tracee through a container image.
+
+## Prerequisites
+
+!!! The container image currently does not support AArch64 architecture. Please ensure that you are on x64 system.
+
+Additionally, please ensure that Docker or another container runtime is working on your machine. 
+
+Note that we will provide support for multiple platforms, such as AArch64, in upcoming releases.
+In the meantime, have a look at [the Kubernetes guide](./kubernetes-quickstart).
+## Run the Tracee container images
+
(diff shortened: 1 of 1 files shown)


docs/getting-started/docker-quickstart.md
```diff
@@ -1,6 +1,20 @@
...
+This section details how you can run Tracee through a container image.
+
+## Prerequisites
+
+!!! The container image currently does not support AArch64 architecture. Please ensure that you are on x64 system.
+
+Additionally, please ensure that Docker or another container runtime is working on your machine. 
+
+Note that we will provide support for multiple platforms, such as AArch64, in upcoming releases.
+In the meantime, have a look at [the Kubernetes guide](./kubernetes-quickstart).
```
yanivagman, line 12: That is not 100% correct.
We actually support aarch64 architecture, and we have the aarch64 tag for that.
If one wants to use it, he can, and should replace latest tag with aarch64 tag.
In the future, we can try to support both architectures in the same docker image using a multi platform image https://docs.docker.com/build/building/multi-platform/

rafaeldtinoco:
@AnaisUrlichs FYI git log messages:

1. keep the "imperative" mood (use fix x,y,z and not fixing x,y,z)
2. use the subject format we have recently adopted if possible

There are tons of git message best practices around there but they are too prolix for me to point you to for simple stuff. So I would recommend your git messages like:

docs(area_being_documented): include tutorial for x
docs(another_area): rewrite paragraph for better understanding
...

If you have any doubts, use "git log" and check the previous commits and they will give you an idea of how to format your git log.

Thank you

Cheers!

geyslan (approved):
LGTM. Gonna squash and reword commits.

AnaisUrlichs:
Made further changes in accordance with the working solution https://github.com/aquasecurity/tracee/pull/3537

PR #3463, opened 2023-09-11
feat: add enable/disable rule rpc

josedonizetti:


yanivagman (approved):
How are we going to support the case policy name is not given (so we should disable the rule in all policies)?

PR #3465, opened 2023-09-12
feat: Add new event structure

josedonizetti:
This PR is the first part of https://github.com/aquasecurity/tracee/issues/2870. It adds the new Event structure without the event data.


josedonizetti:
@yanivagman This is marked as draft but can you do a first review? There are a few questions marked with `TODO` and also an important change on the event data, due to limitations, I endup going with the design made by Daniel when we has investigating this topic -> https://github.com/aquasecurity/tracee/blob/main/pkg/signatures/celsig/wrapper/event.proto#L89

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,162 @@
+syntax = "proto3";
+
+option go_package = "github.co/aquasecurity/tracee/api/v1beta1";
+
+package tracee.v1beta1;
+
+message Event {
+    int64 timestamp = 1;
+    int64 id = 2;
```
yanivagman, line 9: not sure if it has an impact, but 64 bit is probably to much.
Also id should be unsigned

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,162 @@
...
+    string name = 3;
+    // TODO: version? are we going to have version here, or just at the definition?
```
yanivagman, line 11: Version is static information about the event, so at least for now, just in the definition.

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,162 @@
...
+    map<string, string> tags = 4; // TODO:  is this type correct?
```
yanivagman, line 12: should not have values, just an array of strings

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,162 @@
...
+    map<string, string> labels = 5; // TODO:  is this type correct?
```
yanivagman, line 13: The type is correct, but since we don't have labels implemented yet (and not sure when we will) do you think we need to include it?

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,162 @@
...
+    Policies policies = 6;
+    Context context = 7;
+    repeated EventData data = 8;
+    optional Threat threat = 9;
+}
+
+message Policies {
+    repeated string matched = 1;
+    repeated string actions = 2;
```
yanivagman, line 22: since we don't have it implemented yet - do you think we should include it now?

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,162 @@
...
+}
+
+message Context {
+    Process process = 1;
```
yanivagman, line 26: Theoretically we may have events that are not specific to some process (e.g. a threat detected since some processes communicated in some way) so maybe we should make it optional as well?

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,162 @@
...
+    optional Container container = 2;
+    optional Kubernetes kubernetes = 3;
```
yanivagman, line 28: k8s is shorter and fits otel semantic conventions

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,162 @@
...
+}
+
+message Process  {
+    int64 executionTime = 1;
+    Executable executable = 2;
+    int64 pid = 3;
+    int64 namespaced_pid = 4;
```
yanivagman, line 35: int32 should be enough

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,162 @@
...
+    uint32 entity_id = 5;
+    User real_user = 6;
+    User user = 7;
+    Parent parent = 8;
```
yanivagman, line 39: Parent should be of type Process (nested).
By default, we only populate the fields mentioned in the issue, but an event can choose to add more information if required

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,162 @@
...
+    Executable executable = 3;
+    int64 start = 4;
+    optional Parent parent = 5;
+}
+
+message Thread {
+    int64 start = 1;
+    string name = 2;
+    int64 tid = 3;
+    int64 namespaced_tid = 4;
```
yanivagman, line 65: int32 should be enough

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,162 @@
...
+    repeated string capabilities = 5;
+    string syscall = 6;
+    bool compat = 7;
+    repeated uint64 user_stack_strace = 8;
```
yanivagman, line 69: This should probably have its own type.
Now we only have the address but we will want to add the symbol (function) name as well

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,162 @@
...
+}
+
+message Container {
+    string id = 1;
+    string name = 2; 
+    ContainerImage image = 3;
+    bool started = 4;
+    int64 start = 5;
+    int64 pid = 6;
```
yanivagman, line 78: int32 should be enough.
Same qestion about a field which doesn't currently exist (there are some more fields like this, but I see you included them all, so I guess we want them included?)

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,162 @@
...
+
+message KubernetesNamespace {
+    string name = 1;
+}
+
+message Threat {
+    string description = 1;
+    MitreTatic mitre_tactic = 2;
+    MitreTechinique mitre_technique = 3;
+    string severity = 4;
```
yanivagman, line 106: should probably be an enum with the possible values

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,162 @@
...
+  optional uint64 uint64Value = 4;
+  optional int32 int32Value = 5;
+  optional int64 int64Value = 6;
+  optional TriggeringEvent triggering_event = 7;
+}
+
+message TriggeringEvent {
+    int64 id = 1;
+    string name = 2;
+    // data -> TODO: what does this hold?
```
yanivagman, line 161: Should be the same as `repeated EventData data = 8;`

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,162 @@
...
+    string id = 1;
+    string external_id = 2;
+    string name = 3;
+}
+
+// TODO:  what do you think about event data like this? Please read below the reasoning
+// document the Value message.
+message EventData {
+  string name = 1;
+  ValueType valueType = 2;
```
yanivagman, line 123: I don't think these "metadata" fields should be part of what we send.
The schema of a specific event is well defined, and can be retrieved from the event definition.
For the specific use case of "parse arguments" where we change the type, I think we should have a different field in the schema with a different name and type (not implemented today).

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,162 @@
...
+//
+message Value {
+  optional string stringValue = 1;
+  repeated string stringArrayValue = 2;
+  optional uint32 uint32Value = 3;
+  optional uint64 uint64Value = 4;
+  optional int32 int32Value = 5;
+  optional int64 int64Value = 6;
+  optional TriggeringEvent triggering_event = 7;
+}
```
yanivagman, line 156: Questions:
1. What if there is more than one value of a specific type? E.g. two strings. How is that supported?
2. since stringArrayValue is repeated - does it mean it is not optional? in other words, stringArrayValue will always appear, no?
3. There are other types not included here - are you going to add them as well?

yanivagman (changes requested):

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,126 @@
+syntax = "proto3";
+
+option go_package = "github.co/aquasecurity/tracee/api/v1beta1";
+
+package tracee.v1beta1;
+
+message Event {
+    int64 timestamp = 1; // google timestamp
```
yanivagman, line 8: how is it a google timestamp?

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,126 @@
...
+message Context {
+    optional Process process = 1;
+    optional Container container = 2;
+    optional K8S k8s = 3;
+}
+
+message Process  {
+    Executable executable = 1;
+    int64 pid = 2;
+    int32 namespaced_pid = 3;
```
yanivagman, line 30: shouldn't be of the same type as namespaced_pid?

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,126 @@
...
+    optional Process parent = 6;
+    Thread thread = 7;
+}
+
+message Executable {
+    string name = 1;
+}
+
+message User {
+    int64 id = 1;
```
yanivagman, line 42: I think type should be uint32

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,126 @@
...
+}
+
+message Thread {
+    int64 start = 1; // google timestamp
+    string name = 2;
+    int64 tid = 3;
+    int32 namespaced_tid = 4;
```
yanivagman, line 49: ditto

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,126 @@
...
+    repeated string capabilities = 5;
+    string syscall = 6;
+    bool compat = 7;
+    optional uint64 UserStackStracee = 8;
```
yanivagman, line 53:
```suggestion
    optional uint64 UserStackTrace = 8;
```

yanivagman: Also should be of type UserStackStracee and not uint64

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,126 @@
...
+
+message EventData {
+    string name = 1;
+    oneof value {
+        TriggeringEvent triggering_Event = 2;
+    }
+}
+
+message TriggeringEvent {
+    int32 id = 1;
```
yanivagman, line 123: in message Event this is of type uint32

types/api/v1beta1/event.proto
josedonizetti: it is a todo, i'll change it

types/api/v1beta1/event.proto
josedonizetti: both will be uint32

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,126 @@
...
+    optional Process process = 1;
+    optional Container container = 2;
+    optional K8S k8s = 3;
+}
+
+message Process  {
+    Executable executable = 1;
+    int32 pid = 2;
+    uint32 entity_id = 3;
+    int32 namespaced_pid = 4;
```
yanivagman, line 34: nit: better switch pid and entity_id places since pid better be near namespaced_pid

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,126 @@
...
+    User real_user = 5;
+    Thread thread = 6;
+    optional Parent parent = 7;
+}
+
+message Parent  {
+    Executable executable = 1;
+    int32 pid = 2;
+    uint32 entity_id = 3;
+    optional int32 namespaced_pid = 4;
```
yanivagman, line 44: ditto

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,162 @@
...
+}
+
+message Kubernetes {
+    Pod pod = 1;
+    KubernetesNamespace namespace = 2;
+}
+
+message Pod {
+    string name = 1;
+    string uid = 2;
```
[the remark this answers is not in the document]
josedonizetti: pod ids are `0ae5c03d-5fb3-4eb9-9de8-2bd4b51606ba`, I think string is the one that fits best to represent it. WDYT?

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,126 @@
...
+message User {
+    uint32 id = 1;
+}
+
+message Thread {
+    google.protobuf.Timestamp start = 1;
+    string name = 2;
+    int32 tid = 3;
+    int32 namespaced_tid = 4;
+    repeated string capabilities = 5;
```
yanivagman, line 63: We don't have this field today

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,126 @@
...
+message UserStackTrace {
+    repeated uint64 addresses = 1;
+}
+
+message Container {
+    string id = 1;
+    string name = 2; 
+    ContainerImage image = 3;
+    bool started = 4;
+    google.protobuf.Timestamp start = 5;
```
yanivagman, line 78: we don't have this field today

types/api/v1beta1/event.pb.go
```diff
@@ -0,0 +1,1680 @@
...
+	}
+	return ""
+}
+
+type User struct {
+	state         protoimpl.MessageState
+	sizeCache     protoimpl.SizeCache
+	unknownFields protoimpl.UnknownFields
+
+	Id uint32 `protobuf:"varint,1,opt,name=id,proto3" json:"id,omitempty"`
```
yanivagman, line 518: I see that all the fields are generated with json omitempty.
But for some fields we don't want to omit if empty.
For example here - uid 0 is root, and we definitely don't want to drop it.

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,124 @@
...
+    User real_user = 5;
+    Thread thread = 6;
+    optional Parent parent = 7;
+}
+
+message Parent  {
+    Executable executable = 1;
+    int32 pid = 2;
+    optional int32 namespaced_pid = 3;
+    uint32 entity_id = 4;
```
yanivagman, line 44: I know that this is nit of nits :-) but do you mind placing entity_id between executable and pid?
Then all optional fields will be grouped together (if changing this please also do the same for the Process message)

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,124 @@
...
+
+message ContainerImage {
+    string id = 1;
+    repeated string repo_digests = 2;
+    string name = 3;
+}
+
+message K8S {
+    Pod pod = 1;
+    K8sNamespace namespace = 2;
```
yanivagman, line 87: Another nit: the struct is called `K8S` but here you use `K8s` with `s` - don't we want to be consistent about it?

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,124 @@
...
+    string name = 1;
+    string uid = 2;
+    map<string, string> labels = 3;
+}
+
+message K8sNamespace {
+    string name = 1;
+}
+
+// TODO: confirm which values we should have here,
```
yanivagman, line 100: let's not forget this todo

types/api/v1beta1/event.pb.go
josedonizetti: Oh, true.. We can't have primitives for this semantic than, because 0 will be dropped, best to use a wrapper https://protobuf.dev/reference/java/api-docs/com/google/protobuf/UInt32Value, than if not set it is actually nil, instead of 0. WDTY?

types/api/v1beta1/event.proto
josedonizetti: who can confirm which values we want? research?

josedonizetti: I asked Idan

types/api/v1beta1/event.proto
```diff
@@ -0,0 +1,124 @@
...
+// the ones below I got from AVD
+enum Severity {
+    INFO = 0;
+    LOW = 1;
+    MEDIUM = 2;
+    HIGH = 3;
+}
```
roikol, line 107: Also need to add critical (4)

types/api/v1beta1/event.pb.go
```diff
@@ -0,0 +1,1665 @@
...
+	}
+	return ""
+}
+
+type User struct {
+	state         protoimpl.MessageState
+	sizeCache     protoimpl.SizeCache
+	unknownFields protoimpl.UnknownFields
+
+	Id *wrappers.UInt32Value `protobuf:"bytes,1,opt,name=id,proto3" json:"id,omitempty"`
```
yanivagman, line 519: It is still omit empty with the wrapper

types/api/v1beta1/event.proto
josedonizetti: Thanks Roi

types/api/v1beta1/event.pb.go
josedonizetti: Yes, but omit empty for a struct isn't a problem right?

yanivagman (approved):
LGTM

PR #3466, opened 2023-09-12
feat: add event enable/disable

josedonizetti:
The event manager manages the state (disabled/enabled) of events globably on tracee.

This PR is part of the work to support a [v1 for GRPC](https://github.com/aquasecurity/tracee/issues/3208), it adds an enable/disable method to the policyManager, with a state struct to hold the information about the event that is enable/disbale globally as well as per policy, this struct in the future can hold the event emit/submit mask and also a policyMask to enable/disable policies globally.

### DIFF
--- a/pkg/ebpf/events_pipeline.go
+++ b/pkg/ebpf/events_pipeline.go
@@ -570,9 +570,9 @@ func (t *Tracee) sinkEvents(ctx context.Context, in <-chan *trace.Event) <-chan 
 				continue // might happen during initialization (ctrl+c seg faults)
 			}
 
-			// Is the rule disabled for the matched policies?
-			if !t.policyManager.IsRuleEnabled(event.MatchedPoliciesUser, events.ID(event.EventID)) {
-				logger.Debugw("event dropped because matched rule disabled", "event", event.EventName)
+			// Is the event enabled for the policies or globally?
+			if !t.policyManager.IsEnabled(event.MatchedPoliciesUser, events.ID(event.EventID)) {
+				logger.Debugw("event dropped because it is not enabled", "event", event.EventName)
 				t.eventsPool.Put(event)
 				continue
 			}
--- a/pkg/ebpf/policy_manager.go
+++ b/pkg/ebpf/policy_manager.go
@@ -10,14 +10,33 @@ import (
 // policyManager is a thread-safe struct that manages the enabled policies for each rule
 type policyManager struct {
 	mutex sync.Mutex
-	rules map[events.ID]uint64
+	rules map[events.ID]*eventState
+}
+
+// eventState is a struct that holds the state of a given event
+type eventState struct {
+	policyMask uint64
+	enabled    bool
 }
 
 func newPolicyManager() *policyManager {
 	return &policyManager{
 		mutex: sync.Mutex{},
-		rules: make(map[events.ID]uint64),
+		rules: make(map[events.ID]*eventState),
 	}
+}
+
+// IsEnabled tests if a event, or a policy per event is enabled (in the future it will also check if a policy is enabled)
+// TODO: add metrics about an event being enabled/disabled, or a policy being enabled/disabled?
+func (pm *policyManager) IsEnabled(matchedPolicies uint64, ruleId events.ID) bool {
+	pm.mutex.Lock()
+	defer pm.mutex.Unlock()
+
+	if !pm.isEventEnabled(ruleId) {
+		return false
+	}
+
+	return pm.isRuleEnabled(matchedPolicies, ruleId)
 }
 
 // IsRuleEnabled returns true if a given event policy is enabled for a given rule
@@ -25,12 +44,35 @@ func (pm *policyManager) IsRuleEnabled(matchedPolicies uint64, ruleId events.ID)
 	pm.mutex.Lock()
 	defer pm.mutex.Unlock()
 
-	policyMask, ok := pm.rules[ruleId]
+	return pm.isRuleEnabled(matchedPolicies, ruleId)
+}
+
+// not synchronized, use IsRuleEnabled instead
+func (pm *policyManager) isRuleEnabled(matchedPolicies uint64, ruleId events.ID) bool {
+	state, ok := pm.rules[ruleId]
 	if !ok {
 		return false
 	}
 
-	return policyMask&matchedPolicies != 0
+	return state.policyMask&matchedPolicies != 0
+}
+
+// IsEventEnabled returns true if a given event policy is enabled for a given rule
+func (pm *policyManager) IsEventEnabled(evenId events.ID) bool {
+	pm.mutex.Lock()
+	defer pm.mutex.Unlock()
+
+	return pm.isEventEnabled(evenId)
+}
+
+// not synchronized, use IsEventEnabled instead
+func (pm *policyManager) isEventEnabled(evenId events.ID) bool {
+	state, ok := pm.rules[evenId]
+	if !ok {
+		return false
+	}
+
+	return state.enabled
 }
 
 // EnableRule enables a rule for a given event policy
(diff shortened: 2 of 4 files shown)


yanivagman (commented):
Not sure we need an event manager - can you please explain what other responsibilities it will have in the future other than enabling/disabling an event?
If we choose to keep the event manager, I think it should include all the state about the events:
- The events that were selected by the user
- The event configuration (submit/emit)
- The policies that have this event enabled (so this means moving the policy bitmap of an event from policy manager to the event manager).

One thing that I didn't see here is calling the existing API to unload/load a signature event - do you plan to add it as well?

josedonizetti:
@yanivagman I think in the future the event manager should be the one responsible for [eventsState](https://github.com/aquasecurity/tracee/blob/main/pkg/ebpf/tracee.go#L73), which hold the events the user selected and also the the emit, submit mask. 

But I don't think it should have the policyBitmap, because disabling an event globably is different than disabling a rule in a policy, if I have a mask for the `event.SecurityBPF` of `0b0011` on the policyManager which means we have this event as a rule into policy 0 and 1, if I disable the event, I should stop emitting/submitting it, but later when I enable it back the policies that have it should still have the mask of `0b0011`, right? We shouldn't change the policyBitmask. That is why I didn't want to do a enable/disable event into the policymanager. 

Also, not refactoring the `eventsState` now is a way to simplify this change to deliver the API, because eventsState is used everywhere, so encapsulating it first, and refactoring later should be the approach here.

WDTY?

pkg/ebpf/events_pipeline.go
```diff
@@ -561,6 +561,13 @@ func (t *Tracee) sinkEvents(ctx context.Context, in <-chan *trace.Event) <-chan
 				continue // might happen during initialization (ctrl+c seg faults)
 			}
 
+			// Is the event disabled?
+			if !t.eventManager.IsEventEnabled(events.ID(event.EventID)) {
+				logger.Debugw("event dropped because it is disabled", "event", event.EventName)
```
geyslan, line 566: An idea for the future: instead of logging the drop (which is noisy, even in a debugging, in a huge submission of a disabled event), we could count them like we do with lost events.

josedonizetti:
> > But I don't think it should have the policyBitmap, because disabling an event globably is different than disabling a rule in a policy, if I have a mask for the `event.SecurityBPF` of `0b0011` on the policyManager which means we have this event as a rule into policy 0 and 1, if I disable the event, I should stop emitting/submitting it, but later when I enable it back event should still have the mask of the policies `0b0011`, right? We shouldn't change the policyBitmask for an event disable/enable. That is why I didn't want to do a enable/disable event into the policymanager.
> 
> The fact that we can keep the policy bitmap in the event state doesn't say we should modify it if the event was disable/enabled. Per each event, its state can be composed of:
> 
> 1. Which policies selected it - emit, submit, etc
> 2. Is the event enabled/disabled globally

Ah! Ok, so let's merge both policyManager and eventManager? I think I prefer the name policyManager, and it has both states inside a structure like eventState{policy.Bitmap, bool (for enabled/disabled which later becomes emit/subimt}, and we have one check only check if an event is enabled instead of checking first if the event is enabled, and then if the rule is enabled (later if the policy is enabled). WDYT? or you think two maps, one for each is best? 


> > Unloading/loading a signature should be done later, I want to first finish the API to have it ready in the case we want to release an RC, the unloading/loading can't be done without changing any public API.
> 
> I guess you meant CAN be done, right? If so then yes, I agree

Yes, exactly, we do it after having the API working as an internal change.

pkg/ebpf/events_pipeline.go
josedonizetti: Nice! Will add a TODO about it, should be part of the metrics.

geyslan:
> Ah! Ok, so let's merge both policyManager and eventManager? I think I prefer the name policyManager, and it has both states inside a structure like eventState{policy.Bitmap, bool (for enabled/disabled which later becomes emit/subimt}, ...

I also do prefer naming it `policyManager`. 👍🏼

After that type merge I'll start the relocation of the scattered Policies and EventState logic into the `policyManager`, as pointed in this EPIC issue: https://github.com/aquasecurity/tracee/issues/3239

pkg/ebpf/policy_manager.go
```diff
@@ -10,47 +10,129 @@ import (
...
 func (pm *policyManager) IsRuleEnabled(matchedPolicies uint64, ruleId events.ID) bool {
 	pm.mutex.Lock()
 	defer pm.mutex.Unlock()
 
-	policyMask, ok := pm.rules[ruleId]
+	return pm.isRuleEnabled(matchedPolicies, ruleId)
+}
+
+// not synchronized, use IsRuleEnabled instead
+func (pm *policyManager) isRuleEnabled(matchedPolicies uint64, ruleId events.ID) bool {
```
geyslan, line 51: `isRuleEnabled()` is the unique without mutex, do you have plans to use this internally?

geyslan (approved):
LGTM. There's a question ahead.

pkg/ebpf/policy_manager.go
josedonizetti: Actually, there are two without mutex `isRuleEnabled` and `isEventEnabled`, because they are used in two places either `IsRuleEnabled` or `IsEventEnabled` and `IsEnabled`.

pkg/ebpf/policy_manager.go
```diff
@@ -10,47 +10,129 @@ import (
...
+// TODO: add metrics about an event being enabled/disabled, or a policy being enabled/disabled?
+func (pm *policyManager) IsEnabled(matchedPolicies uint64, ruleId events.ID) bool {
+	pm.mutex.Lock()
+	defer pm.mutex.Unlock()
+
+	if !pm.isEventEnabled(ruleId) {
+		return false
 	}
+
+	return pm.isRuleEnabled(matchedPolicies, ruleId)
```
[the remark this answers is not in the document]
josedonizetti: I'm using `ruleId` where in the future it is a rule, although now it is type `events.ID` and I'm using `eventId` where it is an event, rule depends on passing policy information, event doesn't

pkg/ebpf/policy_manager.go
```diff
@@ -10,47 +10,129 @@ import (
...
+func (pm *policyManager) IsEventEnabled(evenId events.ID) bool {
+	pm.mutex.Lock()
+	defer pm.mutex.Unlock()
+
+	return pm.isEventEnabled(evenId)
+}
+
+// not synchronized, use IsEventEnabled instead
+func (pm *policyManager) isEventEnabled(evenId events.ID) bool {
+	state, ok := pm.rules[evenId]
```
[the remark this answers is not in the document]
josedonizetti: Actually, there are two without mutex isRuleEnabled and isEventEnabled, because they are used in two places either IsRuleEnabled or IsEventEnabled and IsEnabled.

pkg/ebpf/policy_manager.go
geyslan: Oh, they are, indeed. Thanks. 👍🏼

pkg/ebpf/policy_manager.go
```diff
@@ -10,47 +10,129 @@ import (
...
+
+	pm.rules[ruleId] = state
+}
+
+// EnableEvent enables a given event
+func (pm *policyManager) EnableEvent(eventId events.ID) {
+	pm.mutex.Lock()
+	defer pm.mutex.Unlock()
+
+	state, ok := pm.rules[eventId]
```
yanivagman, line 117: So how will this code look like in the future when we will have rule ids different than event id?
We will not be able to use pm.rules map anymore, right?

josedonizetti: Probably, but it is an internal structure we should be able to refactor it without affecting anything that is consuming it. Right? It is encapsulated.

yanivagman (approved):
LGTM

PR #3467, opened 2023-09-12
feat: add enable/disable event rpc

josedonizetti:

This PR is part of the work to support a [v1 for GRPC](https://github.com/aquasecurity/tracee/issues/3208), it adds rpc definition of enable/disable event.  The implementation will come in a follow up pr.


yanivagman (approved):
LGTM

PR #3468, opened 2023-09-13
Process tree should have signal event processors for analyze mode.

rafaeldtinoco:
Process tree should have "signal event" processors. They should be:

- called right after the processEvents (processEvent()) pipeline stage
- registered closely to what event processing callbacks function has
- provide an example accessing/updating the process tree (setuid)

Special case:

- deal with the "analyze" case in tracee (try to create the entire process tree from the regular events if analyze mode is on)


yanivagman:
Can you please rebase?

rafaeldtinoco:
Closing this as it is outdated and will pursue other approach.

Issue #3460, opened 2023-09-11
log missing kernel symbol and continue execution


josedonizetti:
Tracee should not abort when a kernel symbol is not found, only log it.

josedonizetti:
@geyslan I belive @yanivagman asked you to implement this one, because it is related to the OSX bug you looking into now. I'll assign it to you, k?

PR #3476, opened 2023-09-18
chore(ebpf): not fail if some ksymbol is not found

geyslan:
Close: #3460 

7b76c016f **chore(ebpf): not fail if some ksymbol is not found**

```
Continue the execution even if some ksymbol is not found or valid.
```

### DIFF
--- a/pkg/ebpf/ksymbols.go
+++ b/pkg/ebpf/ksymbols.go
@@ -37,20 +37,6 @@ func SendKsymbolsToMap(bpfKsymsMap *libbpfgo.BPFMap, ksymbols map[string]*helper
 	return nil
 }
 
-// ValidateKsymbolsTable checks if the addresses in the table are valid by
-// checking a specific symbol address. The reason for the addresses to be
-// invalid is if the capabilities required to read the kallsyms file are not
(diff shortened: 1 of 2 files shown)


yanivagman (approved):
LGTM

josedonizetti:
@geyslan does this cover fully what was the issue https://github.com/aquasecurity/tracee/pull/3476 about? Tracee should have a config for failopen? I'm not sure the config itself is required, though tracee should not fail for any event that can't be loaded in a given kernel, not only ksymbol, but a probe not supporte as well for example?

A quick look and it seems the code doesn't cover it, so it shouldn't close the issue, although you renamed the issue I created, did you create another one for the fail open?

geyslan:
@josedonizetti sorry, I forgot to update you after the offline sync. For now it was only requested to not fail when a ksymbol is not found or is invalid. So I removed the previous `ValidateKsymbolsTable` check, which was just checking for "security_file_open" (and aborting when not found/invalid), and instead let the later called `validateKallsymsDependencies` to log when that happens.

geyslan:
Oh well, news from the front. In our offline sync, it was decided to broad this behaviour to probes etc, leaving the failure reserved only for errors. Opening an issue for this.

Issue #3462, opened 2023-09-11
rename policy apiversion and kind


josedonizetti:
apiVersion: tracee.aquasec.com/v1beta1
kind: Policy
...
```

PR #3484, opened 2023-09-19
fix: rename policy apiVersion and kind

josedonizetti:
Fix https://github.com/aquasecurity/tracee/issues/3462

This PR renames policy apiVersion and kind to be:

```
apiVersion: tracee.aquasec.com/v1beta1
kind: Policy
metadata:
  name: dig
  annotations:
    description: traces dns events from the dig executable
spec:
  scope:
    - executable=/usr/bin/dig
  rules:

```


yanivagman (approved):
LGTM

PR #3485, opened 2023-09-19
fix: remove params from event definition proto

josedonizetti:
Remove `params` from event defintion protobuf.


yanivagman (approved):
LGTM

PR #3488, opened 2023-09-20
Feature: add process tree data source

AlonZivony:

ce696ac95 **chore(proctree): add process tree e2e tests**
7098db0b0 **feat(proctree): create process tree data source**

### DIFF
--- a/pkg/proctree/datasource.go
+++ b/pkg/proctree/datasource.go
@@ -1,0 +1,198 @@
+package proctree
+
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/aquasecurity/tracee/types/datasource"
+	"github.com/aquasecurity/tracee/types/detect"
+)
+
+// DataSource is an implementation to detect.Datasource interface, enveloping the ProcessTree type.
+// It exposes all the relevant information of the tree using the interface methods.
+type DataSource struct {
+	procTree *ProcessTree
+}
+
+func NewDataSource(processTree *ProcessTree) *DataSource {
+	return &DataSource{procTree: processTree}
+}
+
+func (ptds *DataSource) Get(key interface{}) (map[string]interface{}, error) {
+	switch typedKey := key.(type) {
+	case datasource.ProcKey:
+		process, found := ptds.procTree.GetProcessByHash(typedKey.EntityId)
+		if !found {
+			return nil, detect.ErrDataNotFound
+		}
+		return map[string]interface{}{
+			"process_info": ptds.exportProcessInfo(process, typedKey.Time),
+		}, nil
+	case datasource.ThreadKey:
+		thread, found := ptds.procTree.GetThreadByHash(typedKey.EntityId)
+		if !found {
+			return nil, detect.ErrDataNotFound
+		}
+		return map[string]interface{}{
+			"thread_info": ptds.exportThreadInfo(thread, typedKey.Time),
+		}, nil
+	case datasource.LineageKey:
+		process, found := ptds.procTree.GetProcessByHash(typedKey.EntityId)
+		if !found {
+			return nil, detect.ErrDataNotFound
+		}
+		return map[string]interface{}{
+			"process_lineage": ptds.exportProcessLineage(process, typedKey.Time, typedKey.MaxDepth),
+		}, nil
+	default:
+		return nil, detect.ErrKeyNotSupported
+	}
+}
+
+func (ptds *DataSource) Keys() []string {
+	return []string{"datasource.ProcKey", "datasource.ThreadKey", "datasource.LineageKey"}
+}
+
+func (ptds *DataSource) Schema() string {
+	schemaMap := map[string]string{
+		"process_info":    "datasource.ProcessInfo",
+		"thread_info":     "datasource.ThreadInfo",
+		"process_lineage": "datasource.ProcessLineage",
+	}
+	schema, _ := json.Marshal(schemaMap)
+	return string(schema)
+}
+
+func (ptds *DataSource) Version() uint {
+	return 1 // TODO: Change to semantic versioning
+}
+
+func (ptds *DataSource) Namespace() string {
+	return "tracee"
+}
+
+func (ptds *DataSource) ID() string {
+	return "process_tree"
+}
+
+// exportProcessInfo return a representation of the given Process information relevant to given time
+// as the expected datasource process structure.
+func (ptds *DataSource) exportProcessInfo(
+	process *Process,
+	queryTime time.Time,
+) datasource.ProcessInfo {
+	info := process.GetInfo()
+	executable := process.GetExecutable()
+	interpreter := process.GetInterpreter()
+	interp := process.GetInterp()
+
+	existingChildren := make(map[int]uint32)
+	for _, childHash := range process.GetChildren() {
+		child, ok := ptds.procTree.GetProcessByHash(childHash)
+		if !ok {
+			continue
+		}
+		childInfo := child.GetInfo()
+		if childInfo.IsAliveAt(queryTime) {
+			existingChildren[childInfo.GetPid()] = childHash
+		}
+	}
+
+	existingThreads := make(map[int]uint32)
+	for _, threadHash := range process.GetThreads() {
+		thread, ok := ptds.procTree.GetThreadByHash(threadHash)
+		if !ok {
+			continue
+		}
+		threadInfo := thread.GetInfo()
+		if threadInfo.IsAliveAt(queryTime) {
+			existingThreads[threadInfo.GetPid()] = threadHash
+		}
+	}
+
+	infoFeed := info.GetFeedAt(queryTime)
+
+	return datasource.ProcessInfo{
+		EntityId:          process.GetHash(),
+		Pid:               infoFeed.Pid,
+		NsPid:             infoFeed.NsPid,
+		Ppid:              infoFeed.PPid,
+		ContainerId:       "",         // TODO: Add
+		Cmd:               []string{}, // TODO: Add
+		ExecutionBinary:   exportFileInfo(executable, queryTime),
+		Interpreter:       exportFileInfo(interpreter, queryTime),
+		Interp:            exportFileInfo(interp, queryTime),
+		StartTime:         info.GetStartTime(),
+		ExecTime:          time.Unix(0, 0), // TODO: Add
+		ExitTime:          info.GetExitTime(),
+		ParentEntityId:    process.GetParentHash(),
+		ThreadsIds:        existingThreads,
+		ChildProcessesIds: existingChildren,
+		IsAlive:           info.IsAliveAt(queryTime),
+	}
+}
+
+// exportThreadInfo return a representation of the given Thread information relevant to given time
+// as the expected datasource thread structure.
+func (ptds *DataSource) exportThreadInfo(
+	thread *Thread,
+	queryTime time.Time,
+) datasource.ThreadInfo {
+	info := thread.GetInfo()
+	infoFeed := info.GetFeedAt(queryTime)
+	return datasource.ThreadInfo{
+		EntityId:  thread.GetHash(),
+		Tid:       infoFeed.Tid,
+		NsTid:     infoFeed.NsTid,
+		Pid:       infoFeed.Pid,
+		UserId:    infoFeed.Uid,
+		GroupId:   infoFeed.Gid,
+		StartTime: info.GetStartTime(),
+		ExitTime:  info.GetExitTime(),
+		Name:      infoFeed.Name,
+		IsAlive:   info.IsAliveAt(queryTime),
+	}
+}
+
+// exportProcessLineage return a representation of the given Process information and up to a given
+// depth of its ancestors as the expected datasource process lineage structure.
+// The information of each struct is relevant to the fork time of its dependent from it.
+func (ptds *DataSource) exportProcessLineage(
+	process *Process,
+	queryTime time.Time,
+	maxDepth int,
+) datasource.ProcessLineage {
+	lineage := datasource.ProcessLineage{
+		ptds.exportProcessInfo(process, queryTime),
+	}
+	currentProcess := process
+	var found bool
+	var iterationQueryTime time.Time
+	for depth := 0; depth < maxDepth; depth++ {
+		// We don't want to get parents processes which are not part of the container.
+		if currentProcess.GetInfo().GetNsPid() == 1 {
+			break
+		}
+		iterationQueryTime = currentProcess.GetInfo().GetStartTime()
+		currentProcess, found = ptds.procTree.GetProcessByHash(currentProcess.GetParentHash())
+		if !found {
+			break
+		}
+		lineage = append(lineage, ptds.exportProcessInfo(currentProcess, iterationQueryTime))
+	}
+	return lineage
+}
+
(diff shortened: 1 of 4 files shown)


types/datasource/proctree.go
```diff
@@ -0,0 +1,80 @@
...
+
+type NamespacesIds struct {
+	// TODO: Support all namespaces
+	Pid   int
+	Mount int
+}
+
+// ProcessLineage is a representation of a process and its ancestors until the oldest ancestor
+// known in the tree.
+// The lineage is only relevant for the container the process resides in.
```
AlonZivony, line 57: Currently I am not checking for container processes, should add it

pkg/proctree/datasource.go
```diff
@@ -0,0 +1,210 @@
...
+		Inode:  fileInfoFeed.Inode,
+		Device: fileInfoFeed.Dev,
+		Ctime:  time.Unix(0, int64(fileInfoFeed.Ctime)),
+		Mode:   fileInfoFeed.InodeMode,
+	}
+}
+
+// isTaskAlive return whether the task is alive in the given time, for as far as the information
+// can expose this information.
+func isTaskAlive(taskInfo *TaskInfo, queryTime time.Time) bool {
```
rafaeldtinoco, line 198: This should be part of the proctree API. There is a taskInfo method called `HasExited()`. Maybe we should call it `IsTaskAlive()` and create another one `IsTaskAliveAt()`. Then remove this function from here.

pkg/proctree/datasource.go
```diff
@@ -0,0 +1,210 @@
...
+	interp := process.GetInterp()
+
+	existingChildren := make(map[int]uint32)
+	for _, childHash := range process.GetChildren() {
+		child, ok := ptds.procTree.GetProcessByHash(childHash)
+		if !ok {
+			continue
+		}
+		childInfo := child.GetInfo()
+		if isTaskAlive(childInfo, queryTime) {
```
rafaeldtinoco, line 96: https://github.com/aquasecurity/tracee/pull/3488/files#r1332304648

pkg/proctree/datasource.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	existingThreads := make(map[int]uint32)
+	for _, threadHash := range process.GetThreads() {
+		thread, ok := ptds.procTree.GetThreadByHash(threadHash)
+		if !ok {
+			continue
+		}
+		threadInfo := thread.GetInfo()
+		if isTaskAlive(threadInfo, queryTime) {
```
rafaeldtinoco, line 108: ditto

pkg/proctree/datasource.go
```diff
@@ -0,0 +1,210 @@
...
+		ExecutionBinary:   *exportFileInfo(executable, queryTime),
+		Interpreter:       *exportFileInfo(interpreter, queryTime),
+		Interp:            *exportFileInfo(interp, queryTime),
+		StartTime:         info.GetStartTime(),
+		ExecTime:          time.Unix(0, 0), // TODO: Add
+		ExitTime:          info.GetExitTime(),
+		ParentHash:        process.GetParentHash(),
+		ThreadsIds:        existingThreads,
+		ChildProcessesIds: existingChildren,
+		IsAlive:           isTaskAlive(info, queryTime),
```
rafaeldtinoco, line 131: ditto (1)

pkg/proctree/datasource.go
```diff
@@ -0,0 +1,210 @@
...
+		Hash:     thread.GetHash(),
+		Tid:      infoFeed.Tid,
+		NsTid:    infoFeed.NsTid,
+		Pid:      infoFeed.Pid,
+		UserId:   infoFeed.Uid,
+		GroupId:  infoFeed.Gid,
+		ForkTime: info.GetStartTime(),
+		ExitTime: info.GetExitTime(),
+		Name:     "",
+		IsAlive:  isTaskAlive(info, queryTime),
```
rafaeldtinoco, line 153: ditto (1)

pkg/proctree/datasource.go
```diff
@@ -0,0 +1,210 @@
...
+		"process_info":    "datasource.ProcessInfo",
+		"thread_info":     "datasource.ThreadInfo",
+		"process_lineage": "datasource.ProcessLineage",
+	}
+	schema, _ := json.Marshal(schemaMap)
+	return string(schema)
+}
+
+func (ptds *DataSource) Version() uint {
+	return 1
```
rafaeldtinoco, line 67: the datasource, its type and the signature should walk together, right ? we should make a constant and put on top of those files with a good comment explaining any change should bump the version to a specific format.

I usually like always-increasing versions such as "202309202130" (YYYYMMDDhhmm of the last change). But its not mandatory and you guys will be the one bumping this, so feel free.

tests/e2e-inst-signatures/e2e-proctree_data_source.go
```diff
@@ -0,0 +1,304 @@
...
+
+func (sig *e2eProcessTreeDataSource) OnEvent(event protocol.Event) error {
+	eventObj, ok := event.Payload.(trace.Event)
+	if !ok {
+		return fmt.Errorf("failed to cast event's payload")
+	}
+
+	switch eventObj.EventName {
+	case "sched_process_exec":
+		err := sig.checkThread(&eventObj)
```
rafaeldtinoco, line 65: Shouldn't you check if the event is from a process or a thread and then call the appropriate sig.checkXXX for it ? I'm concerned about this: in the checkProcess() you have:

```golang
	threadExist := false
	for tid := range processInfo.ThreadsIds {
		if tid == eventObj.HostThreadID {
			threadExist = true
			break
		}
	}
```

But this will only work if the task is a thread and not a process. What I'm saying is:

A process (or thread group leader) PID=1000 TID=1000 is executed.
- a proctree process is created (there is no thread 1000 on it)

A thread PID=1000 TID=1001 is executed:
- the proctree process entry for 1000 gets a thread 1001
- the proctree thread entry for 1001 is created

tests/e2e-inst-signatures/e2e-proctree_data_source.go
```diff
@@ -0,0 +1,304 @@
...
+	if processInfo.ExecutionBinary.Path != pathname {
+		return fmt.Errorf("process info execution binary in data source (%s) did not match known info from event (%s)",
+			processInfo.ExecutionBinary.Path, pathname)
+	}
+	return nil
+}
+
+// checkLineage check that the lineage from the tree matches the information we know from the
+//
+//	// bash script that ran.
```
rafaeldtinoco, line 203: nit on formatting.

types/datasource/proctree.go
```diff
@@ -0,0 +1,80 @@
...
+	Hash        uint32
+	Tid         int
+	NsTid       int
+	Pid         int
+	UserId      int
+	GroupId     int
+	ForkTime    time.Time
+	ExitTime    time.Time
+	Name        string
+	ProcessHash uint32
```
rafaeldtinoco, line 36: Is this hash from the "thread group leader" ? Maybe we should use the same names both sides (process tree and here).

tests/e2e-inst-signatures/scripts/proctree_data_source.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/bash
```
rafaeldtinoco, line 1: We should use something that can test childs and threads in a proper way. Something like:

```C
include <stdio.h>
#include <stdlib.h>
#include <pthread.h>
#include <unistd.h>
#include <sys/syscall.h>

#define SECS 360000

void* worker_function(void* arg) {

    pid_t pid = getpid();
    pid_t tid = syscall(SYS_gettid);
    pid_t ppid = getppid();

    printf("Thread (ppid=%d, pid=%d, tid=%d) started. Sleeping for %d seconds...\n", ppid, pid, tid, SECS);
    sleep(SECS);
    printf("Thread finished sleeping.\n");

    return NULL;
}

void* special_worker_function(void* arg) {

    pid_t pid = getpid();
    pid_t tid = syscall(SYS_gettid);
    pid_t ppid = getppid();

    printf("Thread (ppid=%d, pid=%d, tid=%d) started. Creating another thread...\n", ppid, pid, tid);

    pthread_t thread;
    int status;

    status = pthread_create(&thread, NULL, worker_function, NULL);
    if (status) {
        fprintf(stderr, "Error creating thread: %d\n", status);
    }

    status = pthread_join(thread, NULL);
    if (status) {
        fprintf(stderr, "Error joining thread: %d\n", status);
    }

    return NULL;
}

int main() {
    pthread_t thread1, thread2;
    int status;

    pid_t pid = getpid();
    pid_t tid = syscall(SYS_gettid);
    pid_t ppid = getppid();

    printf("Process (ppid=%d, pid=%d, tid=%d) started.\n", ppid, pid, tid);

    // Create first thread
    status = pthread_create(&thread1, NULL, worker_function, NULL);
    if (status) {
        fprintf(stderr, "Error creating thread 1: %d\n", status);
        exit(EXIT_FAILURE);
    }

    // Create second thread
    status = pthread_create(&thread2, NULL, special_worker_function, NULL);
    if (status) {
        fprintf(stderr, "Error creating thread 2: %d\n", status);
        exit(EXIT_FAILURE);
    }

    // Wait for first thread to finish
    status = pthread_join(thread1, NULL);
    if (status) {
        fprintf(stderr, "Error joining thread 1: %d\n", status);
        exit(EXIT_FAILURE);
    }

    // Wait for second thread to finish
    status = pthread_join(thread2, NULL);
    if (status) {
        fprintf(stderr, "Error joining thread 2: %d\n", status);
        exit(EXIT_FAILURE);
    }

    printf("All threads have finished.\n");
    return 0;
}
```

It tests:
- main process
- thread group leader
- a regular thread
- a thread that spawns another thread

We can later add a thread that calls exec.

types/datasource/proctree.go
```diff
@@ -0,0 +1,80 @@
...
+	IsAlive     bool
+}
+
+type FileInfo struct {
+	Path   string
+	Hash   string // TODO: should we call it SHA256 or Hash?
```
rafaeldtinoco, line 42: lets call it Hash (and we can decide what algorithm to use, we can even change it later in newer versions without renaming the var).

tests/e2e-inst-signatures/e2e-proctree_data_source.go
AlonZivony: In that case, the process tree is not consistent.
From the user pov, the process thread (tid == pid) is a thread. It should have a thread entry as well as a process entry.
For each TID in the system should be a matching thread in the tree, and for each PID in the system should be a matching process in the tree. This is the expected behavior when exposing a getter function for both.
I see that the problem in the implementation is that we coupled the `TaskInfo` into the `Process` struct, which doesn't make sense from user-mode pov. The process is the shared resources of all its threads, not a task. The current `TaskInfo` could be called also `ThreadInfo`, as the tasks in the kernel are thread info + process info (pointers to shared data like mm, etc.).
Anyways, do you want me to open an issue on it?

types/datasource/proctree.go
AlonZivony: So this connects to my previous comment.
There should be a clear distinction between processes and threads.
I think we have too much of a mish-mash between kernel terms and user terms.
This should be the process hash, because the user is interested in the process information. Not necessarily the thread group leader information (the `TaskInfo` of the group leader is irrelevant for most use cases of the process  tree). 
* I took a closer look on the current design, and I see that some crucial process information is accessible only through the `TaskInfo` struct of the process. We need to discuss the design here to think what is the best API here.

pkg/proctree/datasource.go
AlonZivony: Hmm good question.
Your suggested versioning is much clearer.
The only issue is that the containers datasource is using version "1" right now, so I tried to fit the convention.
If we agree that this is preferable than I will also change the version of the containers datasource.

types/datasource/proctree.go
AlonZivony: Agree, just need to decide on the name

tests/e2e-inst-signatures/e2e-proctree_data_source.go
rafaeldtinoco: https://github.com/aquasecurity/tracee/pull/3494

pkg/proctree/datasource.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+func NewDataSource(processTree *ProcessTree) *DataSource {
+	return &DataSource{procTree: processTree}
+}
+
+func (ptds *DataSource) Get(key interface{}) (map[string]interface{}, error) {
+	switch typedKey := key.(type) {
+	case datasource.ProcKey:
+		process, found := ptds.procTree.GetProcessByHash(typedKey.Hash)
```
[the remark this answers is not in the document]
AlonZivony: Yea, because the PID is insufficient to be a unique identifier.
In the future we can add a method of caching PIDs` hashes, but it is a feature and doesn't require to change current API.

AlonZivony:
Answering https://github.com/aquasecurity/tracee/pull/3488#discussion_r1333274141
I agree, but currently the version is `uint`, so it is not supported to implement it like this.
If you prefer, we can open an issue and do a refactor after this PR changing the datasources version to be a string in the `major.minor.patch` fashion.

AlonZivony:
Has errors with the e2e tests.
Should figure how to solve them.

pkg/proctree/datasource.go
AlonZivony: I close it as the current versioning does not support semantic versioning.
It will require future refactor.

types/datasource/proctree.go
AlonZivony: I believe in the end that having more user-friendly names here is more appropriate.
Tell me what you think.

yanivagman (approved):
LGTM,
needs types PR to be merged first

rafaeldtinoco:
https://github.com/aquasecurity/tracee/pull/3509 merged, feel free to rebase.

AlonZivony:
@yanivagman @rafaeldtinoco needs someone to restart the tests

AlonZivony:
My local e2e test is not working.
I guess at this point that the reason for it is the race condition between the feed of the proctree and the actual events.
Still trying to figure it out.

rafaeldtinoco:
> My local e2e test is not working.
> I guess at this point that the reason for it is the race condition between the feed of the proctree and the actual events.
> Still trying to figure it out.

Like we've just discussed, the only way to get rid of this would be to get "both" sources of events enabled (with https://github.com/aquasecurity/tracee/pull/3498).

Then the proctree entry would always exist since the fork/exec/exit events would come in the same pipeline (at least for this test that doesn't have a filled pipeline).

I'm pretty sure in regular situations the signal events will happen first, but, still ... its an event oriented architecture, can't guarantee a transaction with unordered events without holding the pipeline.

AlonZivony:
I decided to make this PR only on the data source, and putting the e2e tests on another PR.
The e2e tests doesn't fail because of the datasource, and the tests are mergeable after code freeze.
Will push the datasource commit alone soon.

rafaeldtinoco:
Rebased at https://github.com/aquasecurity/tracee/pull/3522 (solving conflicts mostly)

PR #3490, opened 2023-09-20
fix: regenerate definition.pb.go

josedonizetti:
forgot to regenerate definition.pb.go after the last merge removing params


yanivagman (approved):
LGTM

PR #3492, opened 2023-09-20
fix: GetEventDefinition doesn't have params

josedonizetti:
Update types and fix GetEventDefinition endpoint, because EventDefinition doesn't have params anymore.


yanivagman (approved):
LGTM

PR #3493, opened 2023-09-21
Add streams endpoint

josedonizetti:

This PR adds the stream endpoint to the grpc server, which closes the implementation of the v1beta1 of the grpc API. The user needs pass which policies to create a stream, and optionally he can filter with fields to get as a response with the field mask. 

### 2. Explain how to test it

test streams:

```
func main() {
	kk := keepalive.ClientParameters{
		Time:                10 * time.Second, // send pings every 10 seconds if there is no activity
		Timeout:             time.Second,      // wait 1 second for ping ack before considering the connection dead
		PermitWithoutStream: true,             // send pings even without active streams
	}

	var opts []grpc.DialOption
	opts = append(opts, grpc.WithKeepaliveParams(kk))

	if err != nil {
	}
	defer conn.Close()

	client := pb.NewTraceeServiceClient(conn)

	stream, err := client.StreamEvents(context.Background(), &pb.StreamEventsRequest{
		Policies: []string{"signature-events", "test"},
	})
	if err != nil {
		log.Fatal(err)
	}
	for {
		event, err := stream.Recv()
		if err == io.EOF {
			break
		}
		if err != nil {
			log.Fatal(err)
		}
		fmt.Printf("receive event: %v", event)
	}
}
```

test fieldmask:

```
func main() {
	kk := keepalive.ClientParameters{
		Time:                10 * time.Second, // send pings every 10 seconds if there is no activity
		Timeout:             time.Second,      // wait 1 second for ping ack before considering the connection dead
		PermitWithoutStream: true,             // send pings even without active streams
	}

	var opts []grpc.DialOption
	opts = append(opts, grpc.WithKeepaliveParams(kk))

	if err != nil {
	}
	defer conn.Close()

	client := pb.NewTraceeServiceClient(conn)

	stream, err := client.StreamEvents(context.Background(), &pb.StreamEventsRequest{
		Policies: []string{"signature-events", "test"},
		Mask: &fieldmaskpb.FieldMask{
			Paths: []string{
				"name",
			},
		},
	})
	if err != nil {
		log.Fatal(err)
	}
	for {
		event, err := stream.Recv()
		if err == io.EOF {
			break
		}
		if err != nil {
			log.Fatal(err)
		}
		fmt.Println("receive event", event)
	}
}
```

### DIFF
--- a/pkg/server/grpc/tracee.go
+++ b/pkg/server/grpc/tracee.go
@@ -3,16 +3,55 @@ package grpc
 import (
 	"context"
 	"fmt"
+	"time"
+
+	"github.com/golang/protobuf/ptypes/timestamp"
+	"github.com/mennanov/fmutils"
+	"google.golang.org/protobuf/types/known/timestamppb"
+	"google.golang.org/protobuf/types/known/wrapperspb"
 
 	tracee "github.com/aquasecurity/tracee/pkg/ebpf"
 	"github.com/aquasecurity/tracee/pkg/events"
+	"github.com/aquasecurity/tracee/pkg/streams"
 	"github.com/aquasecurity/tracee/pkg/version"
 	pb "github.com/aquasecurity/tracee/types/api/v1beta1"
+	"github.com/aquasecurity/tracee/types/trace"
 )
 
 type TraceeService struct {
 	pb.UnimplementedTraceeServiceServer
 	tracee *tracee.Tracee
+}
+
+func (s *TraceeService) StreamEvents(in *pb.StreamEventsRequest, grpcStream pb.TraceeService_StreamEventsServer) error {
+	var stream *streams.Stream
+	var err error
+
+	if len(in.Policies) == 0 {
+		stream = s.tracee.SubscribeAll()
+	} else {
+		stream, err = s.tracee.Subscribe(in.Policies)
+		if err != nil {
+			return err
+		}
+	}
+	defer s.tracee.Unsubscribe(stream)
+
+	mask := fmutils.NestedMaskFromPaths(in.GetMask().GetPaths())
+
+	for e := range stream.ReceiveEvents() {
+		// TODO: this conversion is temporary, we will use the new event structure
+		// on tracee internals, so the event received by the stream will already be a proto
+		eventProto := convertTraceeEventToProto(e)
+		mask.Filter(eventProto)
+
+		err := grpcStream.Send(&pb.StreamEventsResponse{Event: eventProto})
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
 
 func (s *TraceeService) EnableEvent(ctx context.Context, in *pb.EnableEventRequest) (*pb.EnableEventResponse, error) {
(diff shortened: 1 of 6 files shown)


pkg/server/grpc/tracee.go
```diff
@@ -83,3 +111,91 @@ func convertDefinitionToProto(d events.Definition) *pb.EventDefinition {
...
+		Id:        uint32(e.EventID),
+		Name:      e.EventName,
+		Policies: &pb.Policies{
+			Matched: e.MatchedPolicies,
+		},
+		Context: &pb.Context{
+			Process: &pb.Process{
+				// Executable
+				EntityId:      wrapperspb.UInt32(e.EntityID),
+				Pid:           wrapperspb.UInt32(uint32(e.ProcessID)),
```
yanivagman, line 135: Should be `e.HostProcessID`

pkg/server/grpc/tracee.go
```diff
@@ -83,3 +111,91 @@ func convertDefinitionToProto(d events.Definition) *pb.EventDefinition {
...
+				NamespacedPid: wrapperspb.UInt32(uint32(e.PIDNS)),
```
yanivagman, line 136: Should be `e.ProcessID`

pkg/server/grpc/tracee.go
```diff
@@ -83,3 +111,91 @@ func convertDefinitionToProto(d events.Definition) *pb.EventDefinition {
...
+				RealUser: &pb.User{
+					Id: wrapperspb.UInt32(uint32(e.UserID)),
+				},
+				Thread: &pb.Thread{
+					Start: timestamppb.New(time.Unix(int64(e.ThreadStartTime), 0)),
+					Name:  e.ProcessName,
+					Tid:   wrapperspb.UInt32(uint32(e.ThreadID)),
```
yanivagman, line 143: Should be e.HostThreadID

pkg/server/grpc/tracee.go
```diff
@@ -83,3 +111,91 @@ func convertDefinitionToProto(d events.Definition) *pb.EventDefinition {
...
+					// NamespacedTid
```
yanivagman, line 144: Should be `e.ThreadID`

pkg/server/grpc/tracee.go
```diff
@@ -83,3 +111,91 @@ func convertDefinitionToProto(d events.Definition) *pb.EventDefinition {
...
+					Syscall:        e.Syscall,
+					Compat:         e.ContextFlags.ContainerStarted,
+					UserStackTrace: userStackTrace,
+				},
+				Parent: &pb.Parent{
+					Pid: wrapperspb.UInt32(uint32(e.ParentProcessID)),
```
yanivagman, line 150: Should be `e.HostParentProcessID`

pkg/server/grpc/tracee.go
```diff
@@ -83,3 +111,91 @@ func convertDefinitionToProto(d events.Definition) *pb.EventDefinition {
...
+			MitreTactic: &pb.MitreTatic{
+				Name: "",
+			},
+			MitreTechnique: &pb.MitreTechnique{
+				Id:         e.Metadata.Properties["id"].(string),
+				ExternalId: e.Metadata.Properties["external_id"].(string),
+				Name:       "",
+			},
+			Severity: getSeverity(e),
+		},
```
yanivagman, line 182: This should be null if not a threat detection event

pkg/server/grpc/tracee.go
```diff
@@ -83,3 +118,189 @@ func convertDefinitionToProto(d events.Definition) *pb.EventDefinition {
...
+		Policies: &pb.Policies{
+			Matched: e.MatchedPolicies,
+		},
+		Context: eventContext,
+		Threat:  getThreat(e),
+	}
+
+	if e.Timestamp != 0 {
+		event.Timestamp = timestamppb.New(time.Unix(int64(e.Timestamp), 0))
+	}
```
yanivagman, line 148: does this mean that timestamp field is optional? because it shouldn't be

pkg/server/grpc/tracee.go
```diff
@@ -83,3 +118,189 @@ func convertDefinitionToProto(d events.Definition) *pb.EventDefinition {
...
+func getProcess(e trace.Event) *pb.Process {
+	var userStackTrace *pb.UserStackTrace
+
+	if len(e.StackAddresses) > 0 {
+		userStackTrace = &pb.UserStackTrace{
+			Addresses: e.StackAddresses,
+		}
+	}
+
+	var thredStartTime *timestamp.Timestamp
```
yanivagman, line 162: typo: `threadStartTime`

pkg/server/grpc/tracee.go
```diff
@@ -83,3 +118,189 @@ func convertDefinitionToProto(d events.Definition) *pb.EventDefinition {
...
+
+func getContainer(e trace.Event) *pb.Container {
+	if e.ContainerID == "" && e.Container.Name == "" {
+		return nil
+	}
+
+	container := &pb.Container{
+		Id:   e.Container.ID,
+		Name: e.Container.Name,
+	}
```
yanivagman, line 199: nit: first you use `e.ContainerID` and then `e.Container.ID` - better stick with one of those

pkg/server/grpc/tracee.go
```diff
@@ -3,18 +3,53 @@ package grpc
...
 
 type TraceeService struct {
 	pb.UnimplementedTraceeServiceServer
 	tracee *tracee.Tracee
 }
 
+func (s *TraceeService) StreamEvents(in *pb.StreamEventsRequest, grpcStream pb.TraceeService_StreamEventsServer) error {
+	if len(in.Policies) == 0 {
+		return fmt.Errorf("policies must be provided")
+	}
```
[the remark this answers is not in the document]
josedonizetti: Oh, thought we wanted to always enforce policies, maybe let's do the `*` option like we did for event definitions?

pkg/server/grpc/tracee.go
josedonizetti: it is not optional, though I wasn't sure if the current event structure always have a timestamp, so I tested to check it before conversion.

pkg/server/grpc/tracee.go
```diff
@@ -3,18 +3,50 @@ package grpc
...
+	"github.com/aquasecurity/tracee/types/trace"
 )
 
 type TraceeService struct {
 	pb.UnimplementedTraceeServiceServer
 	tracee *tracee.Tracee
 }
 
+func (s *TraceeService) StreamEvents(in *pb.StreamEventsRequest, grpcStream pb.TraceeService_StreamEventsServer) error {
+	stream, err := s.getStream(in.Policies)
```
yanivagman, line 27: nit: I think it will be better to inline getStream since (1) it is short, and (2) it then makes it clear why we have Unsubscribe below

pkg/server/grpc/tracee.go
```diff
@@ -55,6 +87,18 @@ func (s *TraceeService) GetVersion(ctx context.Context, in *pb.GetVersionRequest
 	return &pb.GetVersionResponse{Version: version.GetVersion()}, nil
 }
 
+func (s *TraceeService) getStream(policies []string) (*streams.Stream, error) {
+	if len(policies) == 0 {
+		return nil, fmt.Errorf("policies must be provided, pass '*' string to subscribe to all policies")
```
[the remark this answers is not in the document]
josedonizetti: so, if empty we subscribe to everything, instead of `*`, will change it

yanivagman (approved):
LGTM

PR #3495, opened 2023-09-21
Log probe attachment error, continue running

geyslan:

0bf2f1799 **chore(ebpf): refactor validateKallsymsDependencies**

```
This refactors validateKallsymsDependencies to be more readable spliting
the logic into a new function: checkUnavailableKSymbols.
```

633bda9fc **feature:(ebpf): don't fail on probe attachment err**

```
On probe attachment error, remove events related to the probe, log the
error and continue execution.
```

### 2. Explain how to test it

Change a probe name to simulate a error on attachment:

```diff
diff --git a/pkg/ebpf/probes/probe_group.go b/pkg/ebpf/probes/probe_group.go
index 5c66c75b6..872ff0159 100644
--- a/pkg/ebpf/probes/probe_group.go
+++ b/pkg/ebpf/probes/probe_group.go
@@ -109,7 +109,7 @@ func NewDefaultProbeGroup(module *bpf.Module, netEnabled bool) (*ProbeGroup, err
                SyscallExit__Internal:      NewTraceProbe(RawTracepoint, "raw_syscalls:sys_exit", "tracepoint__raw_syscalls__sys_exit"),
                SchedProcessFork:           NewTraceProbe(RawTracepoint, "sched:sched_process_fork", "tracepoint__sched__sched_process_fork"),
                SchedProcessExec:           NewTraceProbe(RawTracepoint, "sched:sched_process_exec", "tracepoint__sched__sched_process_exec"),
-               SchedProcessExit:           NewTraceProbe(RawTracepoint, "sched:sched_process_exit", "tracepoint__sched__sched_process_exit"),
+               SchedProcessExit:           NewTraceProbe(RawTracepoint, "sched:sched_process_exit", "WRONGtracepoint__sched__sched_process_exit"),
```

Run tracee and see that even a probe missing it continues working:

```
sudo ./dist/tracee -s comm=uname -e sched_process_exec,sched_process_exit
{"level":"error","ts":1695328115.5947068,"msg":"Event canceled because of missing probe dependency","missing probe":6,"event":"sched_process_exit"}
17:28:43:409881  1000   uname            994453  994453  0                sched_process_exec        cmdpath: /usr/bin/uname, pathname: /usr/bin/uname, dev: 271581188, inode: 5506450, ctime: 1683546656379101996, inode_mode: 33261, interpreter_pathname: <nil>, interpreter_dev: <nil>, interpreter_inode: <nil>, interpreter_ctime: <nil>, argv: [uname], interp: /usr/bin/uname, stdin_type: S_IFCHR, stdin_path: /dev/pts/5, invoked_from_kernel: 0, env: <nil>
^C
```


geyslan:
e2e 1111 is green. Ready for reviews.

pkg/ebpf/tracee.go
```diff
@@ -897,57 +897,54 @@ func (t *Tracee) computeConfigValues() []byte {
...
-				if ok {
-					symEvents = append(symEvents, id)
-				} else {
-					symEvents = []events.ID{id}
-				}
-				symsToDependentEvents[symDep.GetSymbol()] = symEvents
+func (t *Tracee) checkUnavailableKSymbols(eventsState map[events.ID]events.EventState) map[events.ID][]string {
+	reqKSyms := []string{}
+	kSymbolsToEvents := make(map[string][]events.ID)
+	unavailableKSymsToEvents := make(map[events.ID][]string)
```
yanivagman, line 903: nit: it's not accurate to call it "...ToEvents" since the map value is an array of symbols (and not events)

yanivagman (approved):
LGTM with a nit comment

pkg/ebpf/tracee.go
geyslan: Changing it to `unavailableKSymsForEventID`.

yanivagman: SGTM

pkg/ebpf/tracee.go
```diff
@@ -1176,11 +1174,25 @@ func (t *Tracee) attachProbes() error {
...
-				return errfmt.Errorf("failed to attach required probe: %v", err)
+		err = t.probes.Attach(probe.GetHandle(), t.cgroups)
+		if err != nil && probe.IsRequired() {
+			// if a required probe fails to attach, cancel all events that depend on it
+			for _, evtID := range evtsIDs {
+				evtName := events.Core.GetDefinitionByID(evtID).GetName()
+				logger.Errorw(
+					"Event canceled because of missing probe dependency",
+					"missing probe", probe.GetHandle(), "event", evtName,
+				)
```
rafaeldtinoco, line 1193: Sorry for the late comment here @geyslan, but I'm catching up with latest changes before moving on. I believe this change missed "event dependencies", right ? So if a symbol isn't available you are disabling the event directly related to that dependency (all of them) but not the events that depend on that event. 

The direct example, that you're likely solving in the other PR (For the print syscall table), could be:

- HiddenKernelModuleSeeker (depends on symbols "modules", "module_kset", "mod_tree")

So if one of those symbols isn't available, you would disable it directly in the eventsState. But you wouldn't disable the event that depends on it (that could, and probably would) be enabled:

- HiddenKernelModule

Right ? And then there is the fact of the new policy manager (which would also have to have the rules, using those events, disabled from each policy that uses them).

Is that all correct ? And are you addressing all those in the other PR ? Just to accelerate my understanding. Thank you.

geyslan: > So if one of those symbols isn't available, you would disable it directly in the eventsState. But you wouldn't disable the event that depends on it (that could, and probably would) be enabled

After your feedback, I'm making changes to the new PR tackling the case of events dependency.

> And then there is the fact of the new policy manager (which would also have to have the rules, using those events, disabled from each policy that uses them).

Currently, in the event cancellation phase, it would not reach the point of being propagated to BPF filters. So disabling it in policyManager is not necessary now.

In the future, all that logic (eventState/policies) must reside in the policy manager, indeed.

PR #3496, opened 2023-09-22
feat: add event_data to event proto

josedonizetti:

This PR adds the `event_data` structure to the event protobuf.


types/api/v1beta1/event_data.proto
```diff
@@ -0,0 +1,99 @@
...
+    oneof value {
+        google.protobuf.Int32Value int32 = 1; // intT
+        google.protobuf.UInt32Value uint32 = 2; // uintT
+        google.protobuf.Int64Value int64 = 3; // longT
+        google.protobuf.UInt64Value uint64 = 4; // ulongT
+        google.protobuf.UInt64Value off = 5; // offT
+        google.protobuf.UInt32Value mode = 6; // modeT
+        google.protobuf.UInt32Value dev = 7; // devT
+        google.protobuf.UInt64Value size = 8; // sizeT
+        google.protobuf.UInt64Value pointer = 9; // pointerT
```
yanivagman, line 19: do we need to repeat those types?

types/api/v1beta1/event_data.proto
```diff
@@ -0,0 +1,99 @@
+syntax = "proto3";
+
+option go_package = "github.co/aquasecurity/tracee/api/v1beta1";
+
+package tracee.v1beta1;
+
+import "google/protobuf/wrappers.proto";
+
+message EventValue {
+    oneof value {
```
yanivagman, line 10: I'm missing types for triggerEvent and for args (array of ordered system call arguments)

types/api/v1beta1/event_data.proto
```diff
@@ -0,0 +1,99 @@
...
+message IntArrayValue {
+    repeated google.protobuf.Int32Value array = 1;
+}
+	
+message UInt64ArrayValue {
+    repeated google.protobuf.UInt64Value array = 1;   
+}
+	
+// TODO: what do you think we don't have a uint16 type in protobuf,
+// so we have a message to inform the size, so the user converts the internal value
```
yanivagman, line 48: What is "a message to inform the size"?

types/api/v1beta1/event_data.proto
josedonizetti: oh, true, we have the name in the map, so we don't have to repeat the types here.

types/api/v1beta1/event_data.proto
josedonizetti: `message Uint16Value` -> a message specific for the type, though the user will have to convert because we don't have uint16 support in protobuf. Makes sense?

types/api/v1beta1/event_data.proto
```diff
@@ -0,0 +1,145 @@
...
+        SyscallArgsValue syscall_args = 16;
+        TriggerEvent trigger_event = 17;
+    }
+}
+
+message TriggerEvent {
+    uint32 id = 1;
+    string name = 2;
+    // TODO: is this correct?
+    map<string, EventValue> data = 3;
```
yanivagman, line 35: Mostly yes, maybe with the exception that TriggerEvent can't contain another TriggerEvent

types/api/v1beta1/event_data.proto
```diff
@@ -0,0 +1,145 @@
...
+    repeated Capability cap_effective = 13;
+    repeated Capability cap_bounding = 14;
+    repeated Capability cap_ambient = 15;
+}
+	
+message TimespecValue {
+    google.protobuf.DoubleValue value = 1;
+}
+
+// TODO: the syscall args type will have an array of event data?
```
yanivagman, line 84: Yes, but again with the exception it won't have TriggerEvent and SyscallArgsValue as one of its types.

types/api/v1beta1/event_data.proto
```diff
@@ -0,0 +1,145 @@
...
+message SyscallArgsValue {
+    repeated EventValue args = 1;
+}
+
+message SockAddrValue {
```
yanivagman, line 89: Needs to match a syscall sock address

types/api/v1beta1/event_data.proto
```diff
@@ -0,0 +1,145 @@
...
+    google.protobuf.UInt32Value gid = 2;
+    google.protobuf.UInt32Value suid = 3;
+    google.protobuf.UInt32Value sgid = 4;
+    google.protobuf.UInt32Value euid = 5;
+    google.protobuf.UInt32Value egid = 6;
+    google.protobuf.UInt32Value fsuid = 7;
+    google.protobuf.UInt32Value fsgid = 8;
+    google.protobuf.UInt32Value user_namespace = 9;
+    google.protobuf.UInt32Value secure_bits = 10;
+    repeated Capability cap_inheritable = 11;
```
yanivagman, line 73: Note that having capability this way will require some transformation from the bitmap we have today

types/api/v1beta1/event_data.proto
```diff
@@ -0,0 +1,135 @@
...
+        google.protobuf.StringValue string = 5; // strT
+        google.protobuf.BytesValue bytes = 6; // bytesT
+        google.protobuf.BoolValue bool = 7; // boolT
+        StringArrayValue string_array = 10; // strArrT, argsArrT
+        IntArrayValue int_array = 11; // intArr2T
+        UInt64ArrayValue uint64_array = 12; // uint64ArrT
+        SockAddrValue sockaddr = 13; // sockaddrT
+        CredValue cred = 14; // credT
+        TimespecValue timespec = 15; // timespecT
+        OrderedArgsValue ordered_args = 16;
```
yanivagman, line 24: Sorry for nitpicking, maybe we should simply call it args (and ArgsValue)?

yanivagman (approved):
just one naming suggestion to consider, LGTM

PR #3498, opened 2023-09-22
fix(proctree): allow regular events to create proctree nodes

rafaeldtinoco:
commit 0ac586115 (HEAD -> proctreestuff, rafaeldtinoco/proctreestuff)
Date:   Thu Sep 21 23:37:10 2023

    fix(proctree): allow regular events to create proctree nodes
    
    It is not impossible, although it is unlikely, that a fork/exec/exit
    event arrives first in the regular events pipeline. There is also a
    possibility that the signal event is lost.
    
    This change allows tracee to choose which source of events the process
    tree should be fed with. Using the control plane (and the fork, exec and
    exit signal events) is preferrable, but user can also opt to use regular
    sched events OR even both.
    
    There is also another need for this: the analyze mode. If the signatures
    rely in data source entries to work, when running tracee analyze mode
    there will be a need to "process events" from the regular pipeline in a
    way that those data sources exist in memory for the signature processing
    as well (TBD later).
    
    There is also the case where the pipeline is almost empty, and the
    userland (and go runtime) context switching may cause one event
    processor to need proc tree data that hasn't yet been saved. Having
    "both" as the source for proc tree is needed in this case.
    
    This commit also:
    
    - creates a generic way to normalize timing args to relative time (or not)

commit 05bedbc6f
Date:   Tue Sep 19 08:38:11 2023

    chore(events_processor): create a file for processor functions
    
    ... preparing terrain for refactoring and for process tree processors.

### DIFF
--- a/pkg/cmd/flags/proctree.go
+++ b/pkg/cmd/flags/proctree.go
@@ -13,12 +13,16 @@ func procTreeHelp() string {
 	return `Select different options for the process tree.
 
 Example:
-  --proctree enabled            | will enable the process tree with default settings (disabled by default).
-  --proctree process-cache=8192 | will cache up to 8192 processes in the tree (LRU cache).
-  --proctree thread-cache=4096  | will cache up to 4096 threads in the tree (LRU cache).
+  --proctree source=[none|events|signals|both]
+      none         | process tree is disabled (default).
+      events       | process tree is built from events.
+      signals      | process tree is built from signals.
+      both         | process tree is built from both events and signals.
+  --proctree process-cache=8192  | will cache up to 8192 processes in the tree (LRU cache).
+  --proctree thread-cache=4096   | will cache up to 4096 threads in the tree (LRU cache).
 
 Use comma OR use the flag multiple times to choose multiple options:
-  --proctree process-cache=X,thread-cache=Y
+  --proctree source=A,process-cache=B,thread-cache=C
   --proctree process-cache=X --proctree thread-cache=Y
 `
 }
@@ -27,9 +31,12 @@ func PrepareProcTree(cacheSlice []string) (proctree.ProcTreeConfig, error) {
 	var err error
 
 	config := proctree.ProcTreeConfig{
+		Source:           proctree.SourceNone, // disabled by default
 		ProcessCacheSize: proctree.DefaultProcessCacheSize,
 		ThreadCacheSize:  proctree.DefaultThreadCacheSize,
 	}
+
+	cacheSet := false
 
 	for _, slice := range cacheSlice {
 		if strings.HasPrefix(slice, "help") {
@@ -42,9 +49,21 @@ func PrepareProcTree(cacheSlice []string) (proctree.ProcTreeConfig, error) {
 		values := strings.Split(slice, ",")
 
 		for _, value := range values {
-			if strings.HasPrefix(value, "enabled") {
-				logger.Debugw("proctree is enabled")
-				config.Enabled = true
+			if strings.HasPrefix(value, "source=") {
+				option := strings.TrimPrefix(value, "source=")
+				switch option {
+				case "none":
+					config.Source = proctree.SourceNone
+				case "events":
+					config.Source = proctree.SourceEvents
+				case "signals":
+					config.Source = proctree.SourceSignals
+				case "both":
+					config.Source = proctree.SourceBoth
+				default:
+					return config, fmt.Errorf("unrecognized proctree source option: %v", option)
+				}
+				cacheSet = true // at least the default ones
 				continue
 			}
 			if strings.HasPrefix(value, "process-cache=") {
@@ -56,7 +75,7 @@ func PrepareProcTree(cacheSlice []string) (proctree.ProcTreeConfig, error) {
 				if size >= 4096 { // minimum size is 4096 (or the default is used)
 					config.ProcessCacheSize = size
 				}
-				config.Enabled = true
+				cacheSet = true
 				continue
 			}
 			if strings.HasPrefix(value, "thread-cache=") {
@@ -68,14 +87,19 @@ func PrepareProcTree(cacheSlice []string) (proctree.ProcTreeConfig, error) {
 				if size >= 4096 { // minimum size is 4096 (or the default is used)
 					config.ThreadCacheSize = size
 				}
-				config.Enabled = true
+				cacheSet = true
 				continue
 			}
 			err = fmt.Errorf("unrecognized proctree option format: %v", value)
 		}
 	}
 
-	if config.Enabled {
+	if cacheSet && config.Source == proctree.SourceNone {
+		return config, fmt.Errorf("proctree cache was set but no source was given")
+	}
+
+	if config.Source != proctree.SourceNone {
+		logger.Debugw("proctree is enabled and it source is set to", "source", config.Source.String())
 		logger.Debugw("proctree cache size", "process", config.ProcessCacheSize, "thread", config.ThreadCacheSize)
 	}
 
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -499,56 +499,61 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
     if (!init_program_data(&p, ctx))
         return 0;
 
-    // Note: we don't place should_trace() here, so we can keep track of the cgroups in the system
+    // NOTE: proc_info_map updates before should_trace() as the entries are needed in other places.
+
     struct task_struct *parent = (struct task_struct *) ctx->args[0];
     struct task_struct *child = (struct task_struct *) ctx->args[1];
 
-    u64 start_time = get_task_start_time(child);
+    // Information needed before the event:
+    int parent_pid = get_task_host_tgid(parent);
+    u64 child_start_time = get_task_start_time(child);
+    int child_pid = get_task_host_tgid(child);
+    int child_tid = get_task_host_pid(child);
+    int child_ns_pid = get_task_ns_tgid(child);
+    int child_ns_tid = get_task_ns_pid(child);
+
+    // Update the task_info map with the new task's info
 
     task_info_t task = {};
     __builtin_memcpy(&task, p.task_info, sizeof(task_info_t));
     task.recompute_scope = true;
-    task.context.tid = get_task_ns_pid(child);
-    task.context.host_tid = get_task_host_pid(child);
-    task.context.start_time = start_time;
+    task.context.tid = child_ns_tid;
+    task.context.host_tid = child_tid;
+    task.context.start_time = child_start_time;
     ret = bpf_map_update_elem(&task_info_map, &task.context.host_tid, &task, BPF_ANY);
     if (ret < 0)
         tracee_log(ctx, BPF_LOG_LVL_DEBUG, BPF_LOG_ID_MAP_UPDATE_ELEM, ret);
 
-    int parent_pid = get_task_host_pid(parent);
-    int child_pid = get_task_host_pid(child);
+    // Update the proc_info_map with the new process's info (from parent)
 
-    int parent_tgid = get_task_host_tgid(parent);
-    int child_tgid = get_task_host_tgid(child);
-
-    proc_info_t *c_proc_info = bpf_map_lookup_elem(&proc_info_map, &child_tgid);
+    proc_info_t *c_proc_info = bpf_map_lookup_elem(&proc_info_map, &child_pid);
     if (c_proc_info == NULL) {
-        // this is a new process (and not just another thread) - add it to proc_info_map
-
-        proc_info_t *p_proc_info = bpf_map_lookup_elem(&proc_info_map, &parent_tgid);
+        // It is a new process (not another thread): add it to proc_info_map.
+        proc_info_t *p_proc_info = bpf_map_lookup_elem(&proc_info_map, &parent_pid);
         if (unlikely(p_proc_info == NULL)) {
-            // parent proc should exist in proc_map (init_program_data should have set it)
+            // parent should exist in proc_info_map (init_program_data sets it)
             tracee_log(ctx, BPF_LOG_LVL_WARN, BPF_LOG_ID_MAP_LOOKUP_ELEM, 0);
             return 0;
         }
 
-        bpf_map_update_elem(&proc_info_map, &child_tgid, p_proc_info, BPF_NOEXIST);
-        c_proc_info = bpf_map_lookup_elem(&proc_info_map, &child_tgid);
-        // appease the verifier
+        // Copy the parent's proc_info to the child's enty.
+        bpf_map_update_elem(&proc_info_map, &child_pid, p_proc_info, BPF_NOEXIST);
+        c_proc_info = bpf_map_lookup_elem(&proc_info_map, &child_pid);
         if (unlikely(c_proc_info == NULL)) {
             tracee_log(ctx, BPF_LOG_LVL_WARN, BPF_LOG_ID_MAP_LOOKUP_ELEM, 0);
             return 0;
         }
 
-        c_proc_info->follow_in_scopes = 0;
-        c_proc_info->new_proc = true;
+        c_proc_info->follow_in_scopes = 0; // updated later if should_trace() passes (follow filter)
+        c_proc_info->new_proc = true;      // started after tracee (new_pid filter)
     }
 
-    // update process tree map if the parent has an entry
+    // Update the process tree map (filter related) if the parent has an entry.
+
     if (p.config->proc_tree_filter_enabled_scopes) {
-        u32 *tgid_filtered = bpf_map_lookup_elem(&process_tree_map, &parent_tgid);
+        u32 *tgid_filtered = bpf_map_lookup_elem(&process_tree_map, &parent_pid);
         if (tgid_filtered) {
-            ret = bpf_map_update_elem(&process_tree_map, &child_tgid, tgid_filtered, BPF_ANY);
+            ret = bpf_map_update_elem(&process_tree_map, &child_pid, tgid_filtered, BPF_ANY);
             if (ret < 0)
                 tracee_log(ctx, BPF_LOG_LVL_DEBUG, BPF_LOG_ID_MAP_UPDATE_ELEM, ret);
         }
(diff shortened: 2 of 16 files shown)


pkg/ebpf/controlplane/controller.go
```diff
@@ -64,7 +64,7 @@ func (ctrl *Controller) Start() {
 // Run runs the controller.
 func (ctrl *Controller) Run(ctx context.Context) {
 	ctrl.ctx = ctx
-	ctrl.debug(false) // change this to "true" to enable process tree printing periodically.
+	ctrl.debug(true) // change this to "true" to enable process tree printing periodically.
```
AlonZivony, line 67: This shouldn't be set

pkg/proctree/proctree.go
```diff
@@ -39,8 +39,31 @@ const (
...
+		return "events"
+	case SourceBoth:
+		return "signals and events"
+	}
+	return "unknown"
+}
+
 type ProcTreeConfig struct {
-	Enabled          bool
+	Source           SourceType
```
AlonZivony, line 66: I don't agree with putting this config here.
The source of the process tree is irrelevant, as you are not using it anywhere.
In the end, it determines how the object which has reference to the process tree (`Tracee`) is feeding it, which is not touching the process tree logic at all.
You should export it to the general config.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -6203,32 +6247,32 @@ int sched_process_exec_signal(struct bpf_raw_tracepoint_args *ctx)
...
+    
     // The proc_info interpreter field is set by "load_elf_phdrs" kprobe program.
-    save_str_to_buf(&signal->args_buf, &proc_info->interpreter.pathname, 10);                           // interpreter_pathname
-    save_to_submit_buf(&signal->args_buf, &proc_info->interpreter.id.device, sizeof(dev_t), 11);        // interpreter_dev
-    save_to_submit_buf(&signal->args_buf, &proc_info->interpreter.id.inode, sizeof(unsigned long), 12); // interpreter_inode
-    save_to_submit_buf(&signal->args_buf, &proc_info->interpreter.id.ctime, sizeof(u64), 13);           // interpreter_ctime
+    save_str_to_buf(&signal->args_buf, &proc_info->interpreter.pathname, 10);                    // interpreter path
+    save_to_submit_buf(&signal->args_buf, &proc_info->interpreter.id.device, sizeof(dev_t), 11); // interpreter device number
+    save_to_submit_buf(&signal->args_buf, &proc_info->interpreter.id.inode, sizeof(u64), 12);    // interpreter inode number
+    save_to_submit_buf(&signal->args_buf, &proc_info->interpreter.id.ctime, sizeof(u64), 13);    // interpreter creation time
```
yanivagman, line 6275: ctime stands for change time

pkg/ebpf/controlplane/processes.go
```diff
@@ -2,123 +2,82 @@ package controlplane
...
-	if err != nil {
-		return errfmt.Errorf("error parsing child_start_time: %v", err)
+	errs = append(errs, err)
+
+	// Handle errors
+	for _, err := range errs {
+		if err != nil {
+			return err
+		}
 	}
```
yanivagman, line 73: Using this approach we loose the specific error message (e.g. "error parsting child_pid...") which will make debugging harder

pkg/ebpf/events_pipeline.go
```diff
@@ -225,13 +227,17 @@ func (t *Tracee) decodeEvents(outerCtx context.Context, sourceChan chan []byte)
 				}
 			}
 
+			// TODO: o timestamp do evento do fork eh do parent (nao da child)
+
```
yanivagman, line 209: Forgot to remove this?

pkg/ebpf/tracee.go
```diff
@@ -363,10 +358,34 @@ func (t *Tracee) Init(ctx gocontext.Context) error {
...
+					t.eventsState[events.SchedProcessFork] = policy.AlwaysSubmit
+					t.eventsState[events.SchedProcessExec] = policy.AlwaysSubmit
+					t.eventsState[events.SchedProcessExit] = policy.AlwaysSubmit
+				}
+				// Submit signal sched_process_{fork,exec,exit} events so process tree can be built
+				signalEvts := func() {
+					t.eventsState[events.SignalSchedProcessFork] = policy.AlwaysSubmit
+					t.eventsState[events.SignalSchedProcessExec] = policy.AlwaysSubmit
+					t.eventsState[events.SignalSchedProcessExit] = policy.AlwaysSubmit
+				}
```
yanivagman, line 377: What's the reason to have both? Only for analyze mode?
It feels redundant to process the same data twice

pkg/proctree/proctree_output.go
```diff
@@ -80,7 +80,6 @@ func (pt *ProcessTree) String() string {
 	newTable := func() *tablewriter.Table {
 		table := tablewriter.NewWriter(buffer)
 		table.SetHeader([]string{"Ppid", "Tid", "Pid", "Date", "Comm", "Children", "Threads"})
-		// If debug() is enabled:
 		// table.SetHeader([]string{"Ppid", "Tid", "Pid", "StartTime", "Hash", "CMD", "Children", "Threads"})
```
yanivagman, line 83: nit: Maybe add a different comment saying this line is for debug?

yanivagman (dismissed):

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -499,56 +499,76 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
     if (!init_program_data(&p, ctx))
         return 0;
 
-    // Note: we don't place should_trace() here, so we can keep track of the cgroups in the system
-    struct task_struct *parent = (struct task_struct *) ctx->args[0];
+    // no should_trace() here so we can keep track of the cgroups in the system
+
     struct task_struct *child = (struct task_struct *) ctx->args[1];
+    struct task_struct *leader = get_leader_task(child);
+    struct task_struct *parent = get_leader_task(get_parent_task(leader));
```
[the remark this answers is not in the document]
rafaeldtinoco: I know and that was the intent. If the proccess tree is built using the "parent" semantics pointing to a real process, why sched_process_XXX events should be different ? Specially considering we might (will) attach "artifacts" (used files, used sockets, etc) to a "real process" (and not per thread).

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -499,56 +499,76 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
...
+    int parent_tid = get_task_host_pid(parent);
+    int parent_ns_pid = get_task_ns_tgid(parent);
+    int parent_ns_tid = get_task_ns_pid(parent);
+    // If the "real parent" is a thread, we need to get the leader's info
+    u64 leader_start_time = get_task_start_time(leader);
+    int leader_pid = get_task_host_tgid(leader);
+    int leader_tid = get_task_host_pid(leader);
+    int leader_ns_pid = get_task_ns_tgid(leader);
+    int leader_ns_tid = get_task_ns_pid(leader);
+    // If the "real parent" parent is a thread, we need to get the child's info as well
```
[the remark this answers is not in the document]
rafaeldtinoco: Same thing you did in the previous comment: the task (process or thread) that did the fork. 
Parent = the process that generated the thread group which "current" is part of
Leader = the leader of the thread group which "current" is part of (it could even be itself)
Child = current task (if a process, or the initial thread, leader == child, if a thread, leader != child)
This meaning is already base for the process tree and something Alon is currently working with.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -499,56 +499,76 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
...
+            // parent should exist in proc_info_map (init_program_data sets it)
             tracee_log(ctx, BPF_LOG_LVL_WARN, BPF_LOG_ID_MAP_LOOKUP_ELEM, 0);
             return 0;
         }
 
-        bpf_map_update_elem(&proc_info_map, &child_tgid, p_proc_info, BPF_NOEXIST);
-        c_proc_info = bpf_map_lookup_elem(&proc_info_map, &child_tgid);
-        // appease the verifier
-        if (unlikely(c_proc_info == NULL)) {
-            tracee_log(ctx, BPF_LOG_LVL_WARN, BPF_LOG_ID_MAP_LOOKUP_ELEM, 0);
```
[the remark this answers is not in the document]
rafaeldtinoco: It was proforma verification for the verifier only. The map update would only fail if the item already exists, which would still allow the lookup to succeed. Because of the lookup, the LRU map wouldn't evict the key between the two.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -557,25 +577,31 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
...
 
-    // follow every pid that passed the should_trace() checks (used by the follow filter)
+    // Always follow every pid that passed the should_trace() checks (follow filter)
     c_proc_info->follow_in_scopes = p.task_info->matched_scopes;
 
-    if (should_submit(SCHED_PROCESS_FORK, p.event) || p.config->options & OPT_PROCESS_INFO) {
-        int parent_ns_pid = get_task_ns_pid(parent);
-        int parent_ns_tgid = get_task_ns_tgid(parent);
-        int child_ns_pid = get_task_ns_pid(child);
-        int child_ns_tgid = get_task_ns_tgid(child);
```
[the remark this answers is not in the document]
rafaeldtinoco: I can move the following:

```C
        u64 parent_start_time = get_task_start_time(parent);
        int parent_tid = get_task_host_pid(parent);
        int parent_ns_pid = get_task_ns_tgid(parent);
        int parent_ns_tid = get_task_ns_pid(parent);
        // If the "real parent" is a thread, we need to get the leader's info
        u64 leader_start_time = get_task_start_time(leader);
        int leader_pid = get_task_host_tgid(leader);
        int leader_tid = get_task_host_pid(leader);
        int leader_ns_pid = get_task_ns_tgid(leader);
        int leader_ns_tid = get_task_ns_pid(leader);
```

to after the should_submit check. This helps saving a few cycles as well (mostly because of the bpf helpers overhead I believe, cause the instructions per-se would be irrelevant).

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: oh, yep, Ill fix it.

pkg/ebpf/controlplane/processes.go
rafaeldtinoco: Argval tells the name of the argument and its type whenever there is an error.

pkg/ebpf/processor_funcs.go
```diff
@@ -21,112 +18,7 @@ import (
...
-	if t.eventProcessor == nil {
-		t.eventProcessor = make(map[events.ID][]func(evt *trace.Event) error)
-	}
-
-	// no need to error check since we know the error is initialization related
-	t.RegisterEventProcessor(events.VfsWrite, t.processWriteEvent)
-	t.RegisterEventProcessor(events.VfsWritev, t.processWriteEvent)
-	t.RegisterEventProcessor(events.KernelWrite, t.processWriteEvent)
-	t.RegisterEventProcessor(events.SchedProcessExec, t.processSchedProcessExec)
-	t.RegisterEventProcessor(events.SchedProcessFork, t.processSchedProcessFork)
```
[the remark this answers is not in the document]
rafaeldtinoco: No internal rule using "sched_process_fork" event was accessing "start_time" in any means I could find. All of them access pid/tid or are helpers for binary data LRU or the process tree (still being worked).

Check it yourself: 

```
$ for file in $(find . -name *.go -exec grep -il sched_process_for
k {} \;); do grep -Fil time $file; done
```

pkg/ebpf/tracee.go
rafaeldtinoco: Maximize chances of having the proctree entry whenever a pipeline event happens and needs the proctree. 

If, by any chance, the signal event hasn't happened yet, the proctree entry will be created by the fork/exec/exit pipeline event (before the entry is needed by a subsequent event).

pkg/proctree/proctree.go
rafaeldtinoco: Yep but then I would create a single item in the general config, related to the process tree, and apart from an already existing struct configuring the tree. No right or wrong.

AlonZivony: I am for having one config for the tree in the `proctree` package, and one in the general config.
You then only need to translate what you are configured with to the `proctree` config.
This ensures the independence of both, so each only includes what is relevant for itself.

pkg/ebpf/c/tracee.bpf.c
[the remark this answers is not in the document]
AlonZivony: I see the problem here, and I have to agree here with @yanivagman.
Someone using the `sched_process_fork` might be interested in the real parent task, and not necessarily the leader or the process parent.
This is something worth discussing.
Currently, even getting the thread from the process tree is impossible because of lacking of information in the general event (the thread start time is since epoch and not since start time). It will require us to add at least something to query the tree on the thread to the event structure. Maybe we should consider adding the process information too? This might ease this problem.
However, giving the thread, process and parent process hash is a bit extreme maybe (though each hash is only 32bit), and start time will be even more so (especially when the start time of the task since boot is probably irrelevant to the user except for the process tree).
So I am for either giving thread, process and parent process hashes in each event, or giving only the thread and process while adding argument for the parent process to the `sched_process_fork` event.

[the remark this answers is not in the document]
rafaeldtinoco: > Currently, even getting the thread from the process tree is impossible because of lacking of information in the general event (the thread start time is since epoch and not since start time). It will require us to add at least something to query the tree on the thread to the event structure. 

In each event you have "Entity ID (tid) + Starttime (not epoch) Hash". You can find that event's task (process or thread) using the entity ID, no ?

I'll review https://github.com/aquasecurity/tracee/pull/3503 to check what you mean.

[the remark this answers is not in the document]
rafaeldtinoco: > You are right that you get the task entity id. However, it is nit exported using json, so it is not applicable in analyze mode.

OHHHHH true true true, shit.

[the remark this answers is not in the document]
rafaeldtinoco: image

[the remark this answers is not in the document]
yanivagman: Parent entity id is going to be part of the new event structure anyway: https://github.com/aquasecurity/tracee/issues/2870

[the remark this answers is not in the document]
yanivagman: Yet, I think we should keep the sematics of having the real parent (be it a thread or not)

pkg/ebpf/processor.go
```diff
@@ -0,0 +1,178 @@
...
+	t.RegisterEventProcessor(events.SecurityPostReadFile, processKernelReadFile)
+	t.RegisterEventProcessor(events.SchedProcessExec, t.processSchedProcessExec)
+	t.RegisterEventProcessor(events.DoInitModule, t.processDoInitModule)
+	t.RegisterEventProcessor(events.HookedProcFops, t.processHookedProcFops)
+	t.RegisterEventProcessor(events.PrintNetSeqOps, t.processTriggeredEvent)
+	t.RegisterEventProcessor(events.PrintSyscallTable, t.processTriggeredEvent)
+	t.RegisterEventProcessor(events.PrintMemDump, t.processTriggeredEvent)
+	t.RegisterEventProcessor(events.PrintMemDump, t.processPrintMemDump)
+	// Run processors below for all events
+	t.RegisterEventProcessor(events.Undefined, t.normalizeTimings)
```
yanivagman, line 103: To be more clear, maybe instead of Undefined we should call it "all"?

pkg/ebpf/processor_funcs.go
```diff
@@ -308,13 +214,7 @@ func (t *Tracee) processSchedProcessExec(event *trace.Event) error {
 	return nil
 }
 
-func (t *Tracee) processSchedProcessFork(event *trace.Event) error {
-	return t.convertArgMonotonicToEpochTime(event, "start_time")
-}
```
yanivagman, line 313: I want to get the opinion of @roikol  or @AlonZivony before removing that

pkg/ebpf/processor.go
AlonZivony: I agree

pkg/ebpf/processor_funcs.go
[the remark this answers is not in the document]
AlonZivony: The idea here was not doing it for a specific signature.
The idea was that getting time not since epoch is not useful except for low-level code.
The user expect to get all times in unix nano seconds (aka since epoch), as it is with event timestamp and other times.
Actually, I thing we even should create a type for time, which automatically upon parsing from the kernel will be changed to unix time so all times will always match what the user expects.

To conclude, I am againts removing it and for even expanding it to all the time arguments we currently have.

pkg/ebpf/processor_funcs.go
AlonZivony: I will write it also here - I am against removing it and event for adding new time arguments to the functionality.

pkg/ebpf/processor_funcs.go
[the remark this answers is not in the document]
rafaeldtinoco: Sounds good to me (about the time type). About Fork, if all you want is for the "event" to contain the NS since epoch, to normalize with all other times, then it sounds good to me and I can keep it.

pkg/ebpf/processor.go
rafaeldtinoco: SGTM

pkg/ebpf/processor_funcs.go
rafaeldtinoco: https://github.com/aquasecurity/tracee/pull/3498#discussion_r1337884152

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -499,56 +499,76 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
     if (!init_program_data(&p, ctx))
         return 0;
 
-    // Note: we don't place should_trace() here, so we can keep track of the cgroups in the system
-    struct task_struct *parent = (struct task_struct *) ctx->args[0];
+    // no should_trace() here so we can keep track of the cgroups in the system
```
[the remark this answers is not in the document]
rafaeldtinoco: proc_info has to always be populated, here or in the signal event (for whenever this is done).

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -499,56 +499,62 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
...
```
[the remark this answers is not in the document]
rafaeldtinoco: for now cgroup_info_map processing is done here (and not in the signal event). removing comment.

pkg/proctree/proctree.go
rafaeldtinoco: I removed the config variable from inside proctree (so its only dedicated to the singleton Config).

pkg/ebpf/processor_funcs.go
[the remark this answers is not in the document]
rafaeldtinoco: image

I thought that checking if user asked for relative time or not was better. This way all timings are either relative or not:

image

Bad thing is that all events using "timing arguments" should declare themselves and convert their args like the fork event example.

pkg/ebpf/c/tracee.bpf.c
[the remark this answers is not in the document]
rafaeldtinoco: The process tree REALLY needs 2 things that the regular sched_process_fork() does not provide:

- leader info
- parent (a real process) info

Since we don't want to touch the existing fork event (its parent) nor add fields (unless needed by something else) I created a small logic of only submitting "leader" and "up_parent" IF a config flag exists (whether proctree is using the regular events as source). 

image

And then, in userland, after processed by proc tree, I remove these extra arguments (because even if the proc tree is not enabled they would be displayed as "nil", and we dont want them displayed at all):

image

image

This "solves" both problems. And, yes, using regular pipeline will be needed as observed by @AlonZivony in his recent tests. Something to discuss later (when the pipeline is small, specially for tests, there are good chances that the fork signal event would arrive after the "exec" regular event, for example).

pkg/ebpf/tracee.go
rafaeldtinoco: BTW there was a bug here (the sched_process_fork,exec,exit events would not be printed because this was too changing t.eventsState too late). I moved this logic to sooner step and it works alright.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -557,25 +562,70 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
...
+
+        // Child (might be a lwp or a process, sched_process_fork trace is calle by clone() also).
+        save_to_submit_buf(&p.event->args_buf, (void *) &child_tid, sizeof(int), 5);
+        save_to_submit_buf(&p.event->args_buf, (void *) &child_ns_tid, sizeof(int), 6);
+        save_to_submit_buf(&p.event->args_buf, (void *) &child_pid, sizeof(int), 7);
+        save_to_submit_buf(&p.event->args_buf, (void *) &child_ns_pid, sizeof(int), 8);
+        save_to_submit_buf(&p.event->args_buf, (void *) &child_start_time, sizeof(u64), 9);
+
+        // Process tree information (if needed).
+        if (p.config->options & OPT_FORK_PROCTREE) {
```
yanivagman, line 592: 1. Do we need this information now that @AlonZivony  added the start time of all the three in every event?
2. If you think we still need it, I don't think it needs to be gated with OPT_FORK_PROCTREE - this will cause inconsistencies in the event output between when the proc tree is enabled or not, so better to always send it as part of the event output

yanivagman (approved):
Overall LGTM,
please see below questions before merging this

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -557,25 +562,70 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
...
+            // Both, the thread group leader and the "up_parent" (the first process, not lwp, found
+            // as a parent of the child in the hierarchy), are needed by the userland process tree.
+            // The userland process tree default source of events is the signal events, but there is
+            // an option to use regular event for maintaining it as well (and it is needed for some
+            // situatins). These arguments will always be removed by userland event processors.
+            struct task_struct *leader = get_leader_task(child);
+            struct task_struct *up_parent = get_leader_task(get_parent_task(leader));
```
AlonZivony, line 599: Not really important, but maybe `proc_parent`?
`up_parent` is a bit vague, but the comment is very clear so it is not crucial.

pkg/ebpf/controlplane/processes.go
```diff
@@ -1,128 +1,76 @@
...
+	errs = append(errs, err)
+	childStartTime, err := parse.ArgVal[uint64](args, "start_time") // child_start_time
+	errs = append(errs, err)
 
-	if childHash != taskHashVerifier ||
-		parentHash != parentHashVerifier ||
-		leaderHash != leaderHashVerifier {
-		return errfmt.Errorf("eBPF Hash does not match")
+	// Handle errors
+	for _, err := range errs {
```
AlonZivony, line 64: Hmm, is it not preferred to just return the error when you get it?
I don't see any pros to this approach.
If you wanted to feed the tree anyways that it would have made sense...

pkg/ebpf/c/tracee.bpf.c
AlonZivony: We still need it.
My thought here is the following - if we want to have different arguments than the normal event, we should create a new event and submit it with the same first arguments and add on top of it the additional arguments (we do the same with magic_write I think).
We already have the concept of internal events, so I think it would be cleaner.

pkg/ebpf/processor_proctree.go
```diff
@@ -0,0 +1,245 @@
...
+			ChildNsTid:      childNsTid,
+			ChildPid:        childPid,
+			ChildNsPid:      childNsPid,
+			ChildStartTime:  childStartTime,
+		},
+	)
+}
+
+// procTreeForkRemoveArgs removes arguments needed for the process tree only (when source is events).
+func (t *Tracee) procTreeForkRemoveArgs(event *trace.Event) error {
```
AlonZivony, line 100: I don't like the idea here of inconsistent event through the pipeline, even if it only adding arguments.
I feel we might have troubles with it in the future.

AlonZivony (changes requested):
The main issue is the `sched_process_fork` which is not the best approach in my opinion.
Overall looks very good except for it.
Most of my comments are about names and comments.

pkg/cmd/flags/proctree.go
```diff
@@ -42,9 +49,21 @@ func PrepareProcTree(cacheSlice []string) (proctree.ProcTreeConfig, error) {
...
+				case "events":
+					config.Source = proctree.SourceEvents
+				case "signals":
+					config.Source = proctree.SourceSignals
+				case "both":
+					config.Source = proctree.SourceBoth
+				default:
+					return config, fmt.Errorf("unrecognized proctree source option: %v", option)
+				}
+				cacheSet = true // at least the default ones
```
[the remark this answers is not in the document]
AlonZivony: @rafaeldtinoco is it a preparation for making the process tree mandatory?

pkg/ebpf/c/tracee.bpf.c
yanivagman: Why do we still need it? Those are just extra fields - we shouldn't care about sending them since we don't remove any of the existing fields.
In the future, we are going to have process_fork user-friendly event anyway, so having more arguments here shouldn't be a problem

AlonZivony: We need the arguments because I don't pass the start times to the event, only to the decode step.
I don't think that increasing the size of the event drastically for an internal use and exposing all the confusing terms of `up_parent` and the likes is good for us. I believe that if this is only for an internal use, than it should be an internal event.
However, the question here is what do we do with the analyze mode?
The analyze mode will need this arguments, no?

pkg/ebpf/controlplane/processes.go
rafaeldtinoco: readability (1 line in between the calls only)

pkg/cmd/flags/proctree.go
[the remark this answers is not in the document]
rafaeldtinoco: Yes, but Im not sure we will set is to mandatory for this release. Disabled by default.

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: > I don't think that increasing the size of the event drastically for an internal use and exposing all the confusing terms of up_parent and the likes is good for us. I believe that if this is only for an internal use, than it should be an internal event.

That was the moto (for having the args, send the values only when needed, and remove the args before submitting to end user). Yes, we can create another event, but it would be the 3rd sched_process_fork event.

> However, the question here is what do we do with the analyze mode?
> The analyze mode will need this arguments, no?

Analyze mode nowadays don't do any kind of "process event" approach, so it wouldn't be ready even if we had something different. I consider this as "preparing the terrain".

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: proc_parent also means nothing related to what it means, I had a very similar name before :\.. thats why I opt for a good comment.

pkg/ebpf/processor_proctree.go
rafaeldtinoco: I thought it was creative :P. My problem with another event is to have only for SchedProcessFork and not for the others, seems weird as well.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1099,76 +1152,86 @@ int tracepoint__sched__sched_process_exec(struct bpf_raw_tracepoint_args *ctx)
...
+    save_to_submit_buf(&p.event->args_buf, &inode_mode, sizeof(umode_t), 5);     // inode mode
+
+    // NOTES:
+    // - inter is the real interpreter (sh, bash, python, perl, ...)
+    // - interpreter is the binary interpreter (ld.so), also known as the loader
+    // - interpreter might be the same as executable (so there is no interpreter)
+
+    // Check if there is an interpreter and if it is different from the executable:
+
+    bool interInodeExists = proc_info->interpreter.id.inode != 0;
```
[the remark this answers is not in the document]
rafaeldtinoco: I would break lines just because of that, huge name for C. I'll call something else instead.

pkg/ebpf/c/tracee.bpf.c
rafaeldtinoco: It was agreed that the analyze mode will be dealt in v0.20 most likely (and an issue will be opened for that).

rafaeldtinoco: I'm keeping the OPT_FORK_PROCTREE because there is extra processing in getting leader and up_parent information that isn't really needed if process tree is not using events source.

I'm keeping the argument removal not to create a 3rd sched_process_fork event (besides the regular and the signal one). With that, the argument removal logic will be kept so we don't expose a big amount of arguments for SchedProcessFork event to the user (and up_parent wouldn't make much sense in that regard).

With that, I'm either solving, mitigating or workarounding all the concerns brought and yet being functional in time for the release. We can always change things later.

pkg/ebpf/processor_proctree.go
rafaeldtinoco: https://github.com/aquasecurity/tracee/pull/3498#discussion_r1340157044

PR #3499, opened 2023-09-22
fix: stackaddress needs to have symbol name

josedonizetti:
Fix user stack address, the addresses must have the symbol name.


yanivagman (approved):
LGTM

PR #3497, opened 2023-09-22
fix(ebpf): fix symbol name on error

geyslan:
Close: #3383 

### 1. Explain what the PR does

826d30f022b872fb5e02b6be1a2aff3a43fea4c2 **fix(ebpf): fix symbol name on error**

```
When not finding a symbol, the error message was showing the symbol name
prefixed with the last attempted symbol name. Now it shows all attempted
symbol names.

This also fixes a bug where it just returned after the first option
lookup failure, instead of trying all.
```

### 2. Explain how to test it

`sudo ./dist/tracee -e print_mem_dump.args.symbol_name=compat_filldir64,do_sys_open,bleh,proc_sys_open,compat_filldir`

```
TIME             UID    COMM             PID     TID     RET              EVENT                     ARGS
{"level":"warn","ts":1695346496.5952117,"msg":"Memory dump","error":"ebpf.(*Tracee).triggerMemDump: policy 0: invalid symbols provided to print_mem_dump event: {system_,sys_,__x64_sys_,__arm64_sys_}compat_filldir64"}
{"level":"warn","ts":1695346496.5952415,"msg":"Memory dump","error":"ebpf.(*Tracee).triggerMemDump: policy 0: invalid symbols provided to print_mem_dump event: {system_,sys_,__x64_sys_,__arm64_sys_}bleh"}
21:00:00:000000  0                       0       0       0                print_mem_dump            bytes: [102 15 31 0 15 31 68 0 0 72 131 236 32 101 72 139 4 37 40 0 0 0 72 137 68 36 24 137 208 247 194 0 0 32 0 116 58 37 0 0 43 0 49 201 72 137 226 72 137 4 36 72 137 76 36 8 72 199 68 36 16 0 0 0 0 232 106 245 255 255 72 139 84 36 24 101 72 43 20 37 40 0 0 0 117 37 72 131 196 32 195 204 204 204 204 129 225 255 15 0 0 37 195 255 127 0 129 226 64 0 64 0 186 0 0 0 0 72 15 68 202 235 177 232 112 155 140], address: 0xffffffffbc97cdf0, length: 127, caller_context_id: 1, arch: x86_64, symbol_name: do_sys_open, symbol_owner: system
[...]

PR #3501, opened 2023-09-25
Revert "fix(ebpf): fix symbol name on error"

geyslan:
Reverts aquasecurity/tracee#3497

After that, symbols_loader is failing in e2e. There is more info in the tests discussion. @yanivagman FYI.


geyslan:
@josedonizetti just waiting to have green signal from all tests.

josedonizetti (approved):

PR #3503, opened 2023-09-26
feat(event): export thread, process and parent entity id

AlonZivony:

4febd8745 **feat(api): add thread entity id to protobuf**
eb156abd3 **feat(event): export thread, process and parent entity ids**

### 2. Explain how to test it

Manually checked that hashes matches the values from the process tree.


rafaeldtinoco:
Please check https://github.com/aquasecurity/tracee/pull/3498/files, specifically:

image

and

image

I had the "same problem" for the proctree when using regular events as "source". My problem could be "mitigated" doing the time normalization after the proctree is processed.

I believe that, in here, you're saying that, due to the fact that you need, from an already "time normalized" event, pick the correct hash. Correct ?

types/trace/trace.go
```diff
@@ -42,8 +42,9 @@ type Event struct {
 	Syscall               string       `json:"syscall"`
 	StackAddresses        []uint64     `json:"stackAddresses"`
 	ContextFlags          ContextFlags `json:"contextFlags"`
-	EntityID              uint32       `json:"-"`    // task unique identifier (*)
-	Args                  []Argument   `json:"args"` // args are ordered according their appearance in the original event
+	ThreadEntityId        uint32       `json:"threadEntityId"`  // thread task unique identifier (*)
+	ProcessEntityId       uint32       `json:"processEntityId"` // process unique identifier (*)
```
rafaeldtinoco, line 46: I'm +1 to have parentEntityId. I have planned to have the 2 from the beginning (process and parent). Then after our discussions, the thread group leader appeared and now for me the 3 should walk together. I bet there will be a future need to access parent and we won't have parent start_time to calculate the hash.

rafaeldtinoco:
Lets have the 3 IDs available ? This way we can always find ourselves in the proctree, and we can also find parent and leader for artifacts used and anything related.

types/trace/trace.go
```diff
@@ -856,7 +858,9 @@ func jsonConvertToProtoDNSQuestionType(argMap map[string]interface{}) (ProtoDNSQ
 	}, nil
 }
 
-func jsonConvertToProtoDNSResourceRecordType(argMap map[string]interface{}) (ProtoDNSResourceRecord, error) {
+func jsonConvertToProtoDNSResourceRecordType(argMap map[string]interface{}) (
```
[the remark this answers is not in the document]
rafaeldtinoco: There are more at the end of this file.

types/api/v1beta1/event.pb.go
```diff
@@ -573,11 +573,12 @@ type Thread struct {
 
 	Start          *timestamp.Timestamp  `protobuf:"bytes,1,opt,name=start,proto3" json:"start,omitempty"`
 	Name           string                `protobuf:"bytes,2,opt,name=name,proto3" json:"name,omitempty"`
-	Tid            *wrappers.UInt32Value `protobuf:"bytes,3,opt,name=tid,proto3" json:"tid,omitempty"`
-	NamespacedTid  *wrappers.UInt32Value `protobuf:"bytes,4,opt,name=namespaced_tid,json=namespacedTid,proto3" json:"namespaced_tid,omitempty"`
-	Syscall        string                `protobuf:"bytes,5,opt,name=syscall,proto3" json:"syscall,omitempty"`
-	Compat         bool                  `protobuf:"varint,6,opt,name=compat,proto3" json:"compat,omitempty"`
-	UserStackTrace *UserStackTrace       `protobuf:"bytes,7,opt,name=user_stack_trace,json=userStackTrace,proto3,oneof" json:"user_stack_trace,omitempty"`
+	EntityId       *wrappers.UInt32Value `protobuf:"bytes,3,opt,name=entity_id,json=entityId,proto3" json:"entity_id,omitempty"`
```
rafaeldtinoco, line 576: Did you edit this file by hand ? This file is automatically generated from the proto file.

types/api/v1beta1/event.pb.go
```diff
@@ -626,6 +627,13 @@ func (x *Thread) GetName() string {
 	return ""
 }
```
rafaeldtinoco, line 629: ditto

pkg/bufferdecoder/decoder.go
```diff
@@ -68,15 +68,19 @@ func (decoder *EbpfDecoder) DecodeContext(ctx *Context) error {
...
+	// 4 byte padding
 	// task_context end
 
-	ctx.EventID = events.ID(int32(binary.LittleEndian.Uint32(decoder.buffer[offset+96 : offset+100])))
-	ctx.Syscall = int32(binary.LittleEndian.Uint32(decoder.buffer[offset+100 : offset+104]))
-	ctx.MatchedPolicies = binary.LittleEndian.Uint64(decoder.buffer[offset+104 : offset+112])
-	ctx.Retval = int64(binary.LittleEndian.Uint64(decoder.buffer[offset+112 : offset+120]))
-	ctx.StackID = binary.LittleEndian.Uint32(decoder.buffer[offset+120 : offset+124])
-	ctx.ProcessorId = binary.LittleEndian.Uint16(decoder.buffer[offset+124 : offset+126])
+	ctx.EventID = events.ID(int32(binary.LittleEndian.Uint32(decoder.buffer[offset+112 : offset+116])))
```
rafaeldtinoco, line 78: Padding seems correct in pahole.

pkg/ebpf/c/common/context.h
```diff
@@ -32,6 +32,13 @@ init_context(void *ctx, event_context_t *context, struct task_struct *task, u32
     context->task.mnt_id = get_task_mnt_ns_id(task);
     context->task.pid_id = get_task_pid_ns_id(task);
     context->task.uid = bpf_get_current_uid_gid();
+    context->task.task_hash_id = hash_task_id(context->task.host_tid, get_task_start_time(task));
+    struct task_struct *leader = get_leader_task(task);
+    context->task.leader_hash_id =
+        hash_task_id(get_task_host_tgid(leader), get_task_start_time(leader));
+    struct task_struct *parent = get_leader_task(get_parent_task(leader));
```
rafaeldtinoco, line 39: Perfect. SO for this case we are considering the parent a "real process" that originated the "thread group" "current" is currently in. Same thing as the process tree.

pkg/bufferdecoder/protocol.go
```diff
@@ -48,7 +52,7 @@ type Context struct {
 }
 
 func (Context) GetSizeBytes() int {
-	return 128
+	return 144
```
rafaeldtinoco, line 55: Correct.

rafaeldtinoco:
@AlonZivony I believe there are some minor issues/things to be addressed (a padding, a .proto file change + make proto run). Everything else seems ok to me. Feel free to place the type change PR so we can merge that before this one.

rafaeldtinoco (commented):
@AlonZivony I believe there are some minor issues/things to be addressed (a padding, a .proto file change + make proto run). Everything else seems ok to me. Feel free to place the type change PR so we can merge that before this one.

types/api/v1beta1/event.pb.go
AlonZivony: Oh, didn't know this.
I will revert the change

pkg/bufferdecoder/protocol.go
```diff
@@ -37,6 +37,10 @@ type Context struct {
 	PidID           uint32
 	Comm            [16]byte
 	UtsName         [16]byte
+	TaskHash        uint32
+	LeaderHash      uint32
+	ParentHash      uint32
+	_               uint32
```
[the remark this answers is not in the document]
AlonZivony: You are correct.
Because the read is manual it didn't effect the tests though.
I will fix

types/api/v1beta1/event.pb.go
AlonZivony: What file is the source of this?

AlonZivony:
Will push the changes after #3505 will be merged

rafaeldtinoco:
@AlonZivony 

```
go get -u github.com/aquasecurity/tracee/types@9c94c39ad
git commit -a -m 'chore(types): upgrade to latest types including entity ids'
```

as the first commit of this PR, then the rest of the PR will take care of the type changes, likely.

yanivagman (commented):
Isn't it too much to send all the hashes on every event?

yanivagman:
> Isn't it too much to send all the hashes on every event?

I mean, from bpf to userspace.
All of these can be extracted from the process tree without sending them from bpf code

AlonZivony:
Will fix the 

> > Isn't it too much to send all the hashes on every event?
> 
> I mean, from bpf to userspace. All of these can be extracted from the process tree without sending them from bpf code

You are right that you can get them from user-mode, but it is a pretty expensive procedure.
We might want to calculate them only before sending to user space, instead of for each hook triggered, but it will complicate things.
Changing where they are calculated can be done in the future without breaking anything as we don't use the hash in the kernel currently, so I don't think its too much to merge now and maybe change in the future.

pkg/server/grpc/tracee_test.go
```diff
@@ -42,7 +44,9 @@ func Test_convertEventWithProcessContext(t *testing.T) {
 	assert.Equal(t, uint32(6), protoEvent.Context.Process.Parent.Pid.Value)
 	assert.Equal(t, uint32(7), protoEvent.Context.Process.RealUser.Id.Value)
 	assert.Equal(t, uint32(8), protoEvent.Id)
-	assert.Equal(t, uint32(9), protoEvent.Context.Process.EntityId.Value)
+	assert.Equal(t, uint32(9), protoEvent.Context.Process.Thread.EntityId.Value)
+	assert.Equal(t, uint32(10), protoEvent.Context.Process.EntityId.Value)
+	assert.Equal(t, uint32(10), protoEvent.Context.Process.Parent.EntityId.Value)
```
rafaeldtinoco, line 49: 10 should 11 here.

rafaeldtinoco:
>>> Isn't it too much to send all the hashes on every event?
>>
>> I mean, from bpf to userspace. All of these can be extracted from the process tree without sending them from bpf code
> 
> You are right that you can get them from user-mode, but it is a pretty expensive procedure. We might want to calculate them only before sending to user space, instead of for each hook triggered, but it will complicate things. Changing where they are calculated can be done in the future without breaking anything as we don't use the hash in the kernel currently, so I don't think its too much to merge now and maybe change in the future.

Actually, if we missed the SignalSchedProcEvent, for example, then we would have no information about parent, and leader, and we wouldn't be able to find the relation (or to create their nodes). Reading procfs wouldn't be always possible, for mitigating this, due to fast spawning processes.

Imagining the situation we do have the nodes in the process tree (for parent and leader) but not the leader and parent hashes... by the time the event is decoded, we would have to check something like "recent pids" cache (since we don't have leader and parent start time information), so it would involve keeping non unique data (recent PIDs and/or start_time and/or their hashes) for something that is meant to be unique (like the hash IDs, which is unique no matter what, as long as there are no hash collisions).

With that said, considering we're talking about adding a u64 to each event (2 u32 hashes), I think we should stick with the hashes for parent, leader and task and consider this as "unique task identifiers" (considering PIDs and TIDs aren't necessarily unique during tracee lifetime).

rafaeldtinoco (approved):
LGTM. There is a tiny test change need (and the kernel tests run) before this can be merged, IMO.

yanivagman:
> With that said, considering we're talking about adding a u64 to each event (2 u32 hashes), I think we should stick with the hashes for parent, leader and task and consider this as "unique task identifiers" (considering PIDs and TIDs aren't necessarily unique during tracee lifetime). The benefits look way bigger than the disadvantages.

Actually there are now 144 bytes and before there were 128, adding 16 bytes in total. Another point to be made is that now we need to calculate the hash in bpf code for every event, which might degrade performance.

pkg/bufferdecoder/decoder.go
```diff
@@ -68,15 +68,19 @@ func (decoder *EbpfDecoder) DecodeContext(ctx *Context) error {
...
+	ctx.Syscall = int32(binary.LittleEndian.Uint32(decoder.buffer[offset+116 : offset+120]))
+	ctx.MatchedPolicies = binary.LittleEndian.Uint64(decoder.buffer[offset+120 : offset+128])
+	ctx.Retval = int64(binary.LittleEndian.Uint64(decoder.buffer[offset+128 : offset+136]))
+	ctx.StackID = binary.LittleEndian.Uint32(decoder.buffer[offset+136 : offset+140])
+	ctx.ProcessorId = binary.LittleEndian.Uint16(decoder.buffer[offset+140 : offset+144])
 	// 2 byte padding
```
yanivagman, line 84: There is no padding here now

pkg/ebpf/c/common/context.h
```diff
@@ -32,6 +32,13 @@ init_context(void *ctx, event_context_t *context, struct task_struct *task, u32
...
+    context->task.parent_hash_id =
+        hash_task_id(get_task_host_tgid(parent), get_task_start_time(parent));
```
yanivagman, line 41: The hashing happens on init_context, which means not only we do the calculation for every event sent to userspace, but for every event in general, which may have performance implications

yanivagman (commented):
I think we should check the performance implications of this change before merging it:
1. Event header size is now 12.5% bigger
2. For every event (not only those sent to userspace) we calculate 3 hashes in bpf code

This needs to be checked against a high poad of events. For example, using magic_write event and sched_switch

pkg/bufferdecoder/decoder.go
AlonZivony: There is, but we are not reading it

rafaeldtinoco:
> Actually there are now 144 bytes and before there were 128, adding 16 bytes in total. Another point to be made is that now we need to calculate the hash in bpf code for every event, which might degrade performance.

Fair enough. I see Alon changed from hash to start_time for both, leader and parent. That was another idea we discussed recently, LGTM as well.

yanivagman (approved):
LGTM
@rafaeldtinoco I'm leaving the merge to you after your review

PR #3506, opened 2023-09-26
Types change for #3503

rafaeldtinoco:
Types change for https://github.com/aquasecurity/tracee/pull/3503


josedonizetti (approved):
Looks good. After this is merged, we need to update types on main, and the new fields must be reflected on the conversion from `*trace.Event` to `proto.Event` -> https://github.com/aquasecurity/tracee/blob/main/pkg/server/grpc/tracee.go#L126

rafaeldtinoco:
@AlonZivony FTR

PR #3509, opened 2023-09-27
feat(types): add proctree datasource

AlonZivony:
This is the types changes for the PR of the proctree datasource #3054


yanivagman (approved):
LGTM

PR #3510, opened 2023-09-27
fix: convert `[]trace.Argument` into `map<string,EventData>`

josedonizetti:
This PRs converts `[]trace.Argument` into `map<string,EventData>`, this code is temporary because later the event structure will be integrated into tracee, but I've created a test for all types to make sure we have cover all supported arguments.


pkg/server/grpc/tracee.go
```diff
@@ -274,21 +283,16 @@ func getThreat(e trace.Event) *pb.Threat {
 		}
 	}
 
-	if _, ok := e.Metadata.Properties["external_id"]; ok {
-		if val, ok := e.Metadata.Properties["external_id"].(string); ok {
-			mitreTechniqueExternal = val
-		}
-	}
```
yanivagman, line 281: So actually this is the field we should assign to mitreTechniqueId (T1234) and not the "id" property (we will probably want to change that in the future for all signatures to be "id")

pkg/server/grpc/tracee_test.go
```diff
@@ -124,11 +124,11 @@ func Test_convertEventWithThreat(t *testing.T) {
...
+	assert.NoError(t, err)
 
 	assert.Equal(t, "An attempt to abuse the Docker UNIX ..", protoEvent.Threat.Description)
-	assert.Equal(t, "privilege-escalation", protoEvent.Threat.MitreTactic.Name)
-	assert.Equal(t, "Exploitation for Privilege Escalation", protoEvent.Threat.MitreTechnique.Name)
-	assert.Equal(t, "attack-pattern--b21c3b2d", protoEvent.Threat.MitreTechnique.Id)
-	assert.Equal(t, "T1068", protoEvent.Threat.MitreTechnique.ExternalId)
+	assert.Equal(t, "privilege-escalation", protoEvent.Threat.Mitre.Tactic.Name)
+	assert.Equal(t, "Exploitation for Privilege Escalation", protoEvent.Threat.Mitre.Technique.Name)
+	assert.Equal(t, "attack-pattern--b21c3b2d", protoEvent.Threat.Mitre.Technique.Id)
```
yanivagman, line 133: Regarding my previous comment - the Id should be the T1068 and not the attack-...

pkg/server/grpc/tracee.go
```diff
@@ -42,10 +42,14 @@ func (s *TraceeService) StreamEvents(in *pb.StreamEventsRequest, grpcStream pb.T
 	for e := range stream.ReceiveEvents() {
 		// TODO: this conversion is temporary, we will use the new event structure
 		// on tracee internals, so the event received by the stream will already be a proto
-		eventProto := convertTraceeEventToProto(e)
+		eventProto, err := convertTraceeEventToProto(e)
+		if err != nil {
+			return err
+		}
```
yanivagman, line 48: Failing here will break the receieve events loop - shouldn't we just log the error and continue?

pkg/server/grpc/tracee.go
```diff
@@ -42,10 +42,14 @@ func (s *TraceeService) StreamEvents(in *pb.StreamEventsRequest, grpcStream pb.T
...
+
 		mask.Filter(eventProto)
 
-		err := grpcStream.Send(&pb.StreamEventsResponse{Event: eventProto})
+		err = grpcStream.Send(&pb.StreamEventsResponse{Event: eventProto})
 		if err != nil {
 			return err
 		}
```
yanivagman, line 55: ditto

pkg/server/grpc/tracee.go
josedonizetti: Yes, agreed.

pkg/server/grpc/tracee.go
josedonizetti: I don't think so, if we can't send is because the connection on the other side isn't there anymore for a a given reason, so we should stop sending and unsubscribe the client. WDYT?

yanivagman: yes, sounds reasonable

yanivagman (approved):
LGTM

PR #3511, opened 2023-09-27
feat(event): add executable info

AlonZivony:

Enrich the events in the pipeline with the executable info of the current process.

### DIFF
--- a/pkg/ebpf/processor.go
+++ b/pkg/ebpf/processor.go
@@ -103,6 +103,10 @@ func (t *Tracee) registerEventProcessors() {
 	// Convert all time relate args to nanoseconds since epoch (add more events as needed)
 	t.RegisterEventProcessor(events.SchedProcessFork, t.processSchedProcessFork)
 	t.RegisterEventProcessor(events.All, t.normalizeEventCtxTimes)
+	if t.config.ProcTree.Source != proctree.SourceNone {
+		// Can only be enriched if process tree is enabled
+		t.RegisterEventProcessor(events.All, t.addBinaryInformation)
+	}
 }
 
 func initKernelReadFileTypes() {
--- a/pkg/ebpf/processor_funcs.go
+++ b/pkg/ebpf/processor_funcs.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"path/filepath"
 	"strconv"
+	"time"
 
 	"golang.org/x/sys/unix"
 	"kernel.org/pub/linux/libs/security/libcap/cap"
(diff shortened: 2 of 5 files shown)


rafaeldtinoco:
Ok, there was a very very recent discussion about creating a pipeline step that I could use for "feeding the process tree with event data" and I was told not to, to use the processEvents instead.

@yanivagman your input here is needed.

AlonZivony:
> Ok, there was a very very recent discussion about creating a pipeline step that I could use for "feeding the process tree with event data" and I was told not to, to use the processEvents instead.
> 
> @yanivagman your input here is needed, since there isn't much difference in this approach.

It is even preferable.
Will soon push with the change.

pkg/ebpf/events_pipeline.go
```diff
@@ -644,6 +653,55 @@ func (t *Tracee) getStackAddresses(stackID uint32) []uint64 {
...
+							event.HostProcessID,
+							event.HostThreadID,
+						), "error", "not found",
+					)
+				} else {
+					executable := currentProcess.GetExecutable()
+
+					if executable != nil {
+						event.Executable = trace.Executable{
+							ExecutionTime: 0, // TODO: Get exec time
```
yanivagman, line 685: Execution time is a property of the process and not of the executable. Let's omit it for now (anyway it is marked todo here)

yanivagman (changes requested):
Like Rafael said, I don't think we need a dedicated pipeline stage for this enrichment

rafaeldtinoco:
> > Ok, there was a very very recent discussion about creating a pipeline step that I could use for "feeding the process tree with event data" and I was told not to, to use the processEvents instead.
> > @yanivagman your input here is needed, since there isn't much difference in this approach.
> 
> It is even preferable. Will soon push with the change.

Im doing the process events change in my PR. I'll rebase on top of yours (datasource) and this one should be rebased on top of mine. If you are already gone, I can do it.

pkg/ebpf/processor_funcs.go
```diff
@@ -378,3 +379,26 @@ func (t *Tracee) normalizeEventArgTime(event *trace.Event, argName string) error
...
+			"tid", event.HostThreadID,
+			"event name", event.EventName,
+			"timestamp", event.Timestamp,
+			"error", "not found in process tree",
+		)
+		return nil
+	}
+	executable := currentProcess.GetExecutable()
+
+	if executable != nil {
```
rafaeldtinoco, line 400: executable wont ever be nil, its initialized together with the node.

pkg/ebpf/processor.go
```diff
@@ -103,6 +103,10 @@ func (t *Tracee) registerEventProcessors() {
 	// Convert all time relate args to nanoseconds since epoch (add more events as needed)
 	t.RegisterEventProcessor(events.SchedProcessFork, t.processSchedProcessFork)
 	t.RegisterEventProcessor(events.All, t.normalizeEventCtxTimes)
+	if t.config.ProcTree.Source != proctree.SourceNone {
+		// Can only be enriched if process tree is enabled
+		t.RegisterEventProcessor(events.All, t.addBinaryInformation)
+	}
```
rafaeldtinoco, line 109: This should go bellow:

```go
	t.RegisterEventProcessor(events.SchedProcessFork, t.procTreeForkRemoveArgs)
```

Idea is:

// Process events for the proctree before any regular event processing

// Process events from the regular pipeline

// Convert all time related args for the event printers/streams

rafaeldtinoco:
Alright, since we're branching out, I'm fixing this PR in another PR and merging it @AlonZivony. FYIO

pkg/ebpf/processor_funcs.go
AlonZivony: But isn't it a best practice to check anyways?
Its not important, so if you want to fix it for me you can (I won't be available until Sunday)

rafaeldtinoco:
Being merged at: https://github.com/aquasecurity/tracee/pull/3526

PR #3512, opened 2023-09-27
fix: make types uniform with wrappers

josedonizetti:

Changes types to be uniform with protobuf wrappers, the inside type should be named `Value`. Also refactor the array name `IntArrayValue` to have the size, like other types `Int32ArrayValue`.


yanivagman (approved):
LGTM

PR #3524, opened 2023-09-28
feat(types): add executable to event

AlonZivony:

Add to event an executable member, to enable enrichment using process tree.


AlonZivony:
Types change for #3511

PR #3526, opened 2023-09-28
feature(proctree): enrich events with executable from proctree

rafaeldtinoco:
Rebase of https://github.com/aquasecurity/tracee/pull/3511

### DIFF
--- a/pkg/ebpf/processor_proctree.go
+++ b/pkg/ebpf/processor_proctree.go
@@ -2,8 +2,10 @@ package ebpf
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/aquasecurity/tracee/pkg/events/parse"
+	"github.com/aquasecurity/tracee/pkg/logger"
 	"github.com/aquasecurity/tracee/pkg/proctree"
 	"github.com/aquasecurity/tracee/pkg/utils"
 	"github.com/aquasecurity/tracee/types/trace"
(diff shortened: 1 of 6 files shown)


pkg/ebpf/processor_proctree.go
```diff
@@ -243,3 +245,38 @@ func (t *Tracee) procTreeExitProcessor(event *trace.Event) error {
...
+				"event name", event.EventName,
+				"timestamp", event.Timestamp,
+				"error", "not found in process tree",
+			)
+		}
+		return nil // threads don't have executable info in the process tree
+	}
+
+	// Event timestamp is changed to relative (or not) at the end of all processors only.
+	eventTimestamp := time.Unix(0, int64(event.Timestamp)+int64(t.bootTime))
```
yanivagman, line 272: How is this related to this PR?
Also - isn't it already taken care of by normalize time processor?

yanivagman (commented):
After this is merged, there should be a follow up PR to add executable to the new event structure and initialize it @josedonizetti

pkg/ebpf/processor_proctree.go
rafaeldtinoco: Event timestamp is ns since boot and proctree timestamp is ns since epoch. The time normalization (that will convert time to be epoch or not) runs at the end of all processors (so it does not disturb what processors expect).

rafaeldtinoco:
All the errors on tests were related to timeouts (in the very last step of the tests).

rafaeldtinoco:
> After this is merged, there should be a follow up PR to add executable to the new event structure and initialize it @josedonizetti

I think that has been already taken care of:

image

If not we can "fix" it after the branch out (since it is happening today).

yanivagman:
> > After this is merged, there should be a follow up PR to add executable to the new event structure and initialize it @josedonizetti
> 
> I think that has been already taken care of:
> 
> image
> 
> If not we can "fix" it after the branch out (since it is happening today).

Should also have "path" in it, which is missing, and the value needs to be initialized

PR #3536, opened 2023-10-02
fix(decoder): read uint16 from buffer

geyslan:

3d617243a **fix(decoder): read uint16 from buffer**

```
ProcessorId is uint16. This fixes the reading to 2 bytes instead of 4.
```


AlonZivony:
LGTM.
I guess this is not really a bug right now, because the convention to `uint16` should have removed the two extra bytes either way (as it is little endian). But good job identifying it - might have been a problem in the future if another field would have been inserted after it.

AlonZivony (approved):
LGTM.
I guess this is not really a bug right now, because the convention to uint16 should have removed the two extra bytes either way (as it is little endian). But good job identifying it - might have been a problem in the future if another field would have been inserted after it.

geyslan:
> LGTM. I guess this is not really a bug right now, because the convention to uint16 should have removed the two extra bytes either way (as it is little endian). But good job identifying it - might have been a problem in the future if another field would have been inserted after it.

Yep. It's my concern since I calculated wrongly based on the size of the last byte slice.

rafaeldtinoco:
Nice catch @geyslan

PR #3538, opened 2023-10-02
fix(networking): fix net_packet_http errors on empty OK

rafaeldtinoco:
The net_packet_http and net_packet_http_response events are failing for empty replies (such as HTTP/1.x 200 OK replies, for example). Ignore if there is some EOF trying to parse response... it usually means there is a single code response with no body.

### DIFF
--- a/pkg/events/derive/net_packet_http.go
+++ b/pkg/events/derive/net_packet_http.go
@@ -3,6 +3,7 @@ package derive
 import (
 	"bufio"
 	"bytes"
+	"io"
 	"net/http"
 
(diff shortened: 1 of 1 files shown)


rafaeldtinoco:
@geyslan this might be something disturbing the HTTP tests (not sure), but definitely something causing error output where it shouldn't.

rafaeldtinoco:
> LGTM, just a question: `deriveHTTPEvents` will continue returning an error `parsePacketError` in such cases, is it expected?

There are 3 rets from eventToProtoHTTPResponse. As long as "err" == nil and "h" == nil, there won't be an error afaict.

PR #3539, opened 2023-10-02
Fix event data

josedonizetti:

This PR fixes a few bugs on event data. 

1 - if the Value is nil, it returns nil instead of trying to parse
2 - it parse based on the actual type, instead of the metadata
3 - it adds the trigger event information


pkg/ebpf/finding.go
```diff
@@ -107,7 +107,7 @@ func getCType(t interface{}) string {
 	}
 }
 
-func getMetadataFromSignatureMetadata(sigMetadata detect.SignatureMetadata) *trace.Metadata {
+func getMetadataFromSignatureMetadata(sigMetadata detect.SignatureMetadata, triggerEvent trace.Event) *trace.Metadata {
```
yanivagman, line 110: What's the idea of putting this on metadata and not directly in `getArguments`?

pkg/server/grpc/event_data.go
```diff
@@ -84,195 +52,204 @@ func getEventData(e trace.Event) (map[string]*pb.EventValue, error) {
...
-		}
-	default:
-		return nil, errfmt.Errorf("unknown arg type: %s - %v", arg.Name, arg.Type)
+			},
+		}, nil
 	}
 
-	if eventValue == nil {
-		return nil, errfmt.Errorf("can't convert event data: %s - %v - %T", arg.Name, arg.Value, arg.Value)
+	if v, ok := arg.Value.(uintptr); ok {
```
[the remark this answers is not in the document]
josedonizetti: true, I had the switch before because of the meta type, forgot to change here

pkg/ebpf/finding.go
josedonizetti: I was trying to not change the actual arguments today, so I thought it would be best o pass as properties. But i can change to args, what you think?

pkg/server/grpc/event_data.go
```diff
@@ -80,193 +48,220 @@ func getEventData(e trace.Event) (map[string]*pb.EventValue, error) {
 		}
 	}
 
+	triggerEvent, err := getTriggerEvent(e)
```
yanivagman, line 51: As discussed offline, let's take the opportunity to rename to `triggeredBy`

pkg/ebpf/finding.go
yanivagman: I think it is ok to add it to the arguments, since this data is missing today

yanivagman (approved):
LGTM

PR #3542, opened 2023-10-03
fix: add path to executable

josedonizetti:
make executable optional, and add path


types/api/v1beta1/event.proto
```diff
@@ -50,6 +50,7 @@ message Parent  {
 
 message Executable {
     string name = 1;
```
yanivagman, line 52: I wonder if we should make name optional as well since it can be computed from path (basename), wdyt?

yanivagman (approved):
LGTM

geyslan (approved):
LGTM

types/api/v1beta1/event.proto
rafaeldtinoco: Waiting on resolution for this one to merge (or not)

PR #3544, opened 2023-10-03
feat: Change hooked syscall event

OriGlassman:

change hooked_syscall event

### 2. Explain how to test it

tracee -e=hooked_syscall


rafaeldtinoco:
Some preliminary comments (you said this was ready on your side).

I'm getting:

```
panic: interface conversion: interface {} is string, not map[string]interface {}

goroutine 108 [running]:
plugin/unnamed-a7e1df0e51bf802afb26eccc7dd59f329a2028bc.(*SyscallTableHooking).OnEvent(0x7f094def4fa0, {{{{0x224ddb2, 0xe}, {0x222e00d, 0x4}, {0x223178d, 0x6}}, 0x0}, {0x2218940, 0xc004b96000}})
        /home/rafaeldtinoco/work/projects/tracee/signatures/golang/syscall_table_hooking.go:60 +0x447
github.com/aquasecurity/tracee/pkg/signatures/engine.signatureStart({0x7f094dd1c480, 0x7f094def4fa0}, 0x70614d7865646e49?, 0x672e6c61636f6c00?)
        /home/rafaeldtinoco/work/projects/tracee/pkg/signatures/engine/engine.go:80 +0x162
created by github.com/aquasecurity/tracee/pkg/signatures/engine.(*Engine).Start in goroutine 66
        /home/rafaeldtinoco/work/projects/tracee/pkg/signatures/engine/engine.go:118 +0xd8
```

Because of the signature metadata "Data" field wrong type conversion in some case.

rafaeldtinoco:
When mocking the error, you're missing the following syscalls:

```
$ sudo ./dist/tracee --output json --events hooked_syscall | jq '.a
rgs[0].value'
"waitpid"
"oldfstat"
"oldstat"
"stime"
"stty"
"gtty"
"ftime"
"signal"
"oldolduname"
"sigaction"
"sigsuspend"
"sigpending"
"oldlstat"
"readdir"
"vm86old"
"ipc"
"sigreturn"
"sigprocmask"
"bdflush"
"afs_syscall"
"llseek"
"old_select"
"vm86"
"old_getrlimit"
"mmap2"
"truncate64"
"geteuid16"
"getegid16"
"setreuid16"
"setregid16"
"getgroups16"
```
From the string slice in core_amd64.go. I think a map indexed by syscall id would be better there.

rafaeldtinoco:
I have tested hooking a syscall with:

https://github.com/rafaeldtinoco/hijack/blob/main/hijack.c

and, despite being able to:

```
[22372.068909] uname() intercepted!
```

I could not get a detection. Is the command line:

```
sudo ./dist/tracee --output json --events hooked_syscall
```

enough ?

signatures/golang/syscall_table_hooking.go
```diff
@@ -49,23 +48,16 @@ func (sig *SyscallTableHooking) OnEvent(event protocol.Event) error {
...
-			sig.cb(detect.Finding{
-				SigMetadata: metadata,
-				Event:       event,
-				Data:        map[string]interface{}{"Hooked syscalls": hookedSymbolSlice},
-			})
-		}
+		sig.cb(detect.Finding{
+			SigMetadata: metadata,
+			Event:       event,
+			Data:        eventObj.Args[0].Value.(map[string]interface{}),
```
rafaeldtinoco, line 59: This is the cause of panic for me.

pkg/events/derive/hooked_syscall.go
```diff
@@ -3,68 +3,69 @@ package derive
...
-var SyscallsToCheck = make([]string, 0)
-var MaxSupportedSyscallID = events.IoPgetevents // Was the last syscall introduced in the minimum version supported 4.18
+var (
+	reportedHookedSyscalls *lru.Cache[int32, uint64]
+)
+
+// InitHookedSyscall initialize lru
+func InitHookedSyscall() error {
+	var err error
+	reportedHookedSyscalls, err = lru.New[int32, uint64](700)
```
rafaeldtinoco, line 23: this hard coded 700 should be a const at the top of the file and 700 in eBPF code your MAX_SYS_CALL_TABLE_SIZE Is 500. Perhaps they should match.

pkg/events/core_amd64.go
```diff
@@ -902,25 +902,458 @@ const (
...
-		"ptrace",
-		"getdents64",
-		"openat",
-		"bpf",
-		"execveat",
-	}
+const SyscallNotImplemented = "NOT_IMPLEMENTED"
+
+// Order matters
+var SyscallSymbolNames = [...]string{
```
rafaeldtinoco, line 908: This could be a map[ID]string declaration I think. Opened to other ideas, but a slice (and matching the index of this slice visually) would be harder to maintain.

pkg/ebpf/hooked_syscall_table.go
```diff
@@ -0,0 +1,106 @@
...
+	defer logger.Debugw("Stopped hookedSyscallTable goroutine")
+
+	if t.eventsState[events.HookedSyscall].Emit == 0 {
+		return
+	}
+
+	if runtime.GOARCH != "amd64" {
+		logger.Debugw("hookedSyscallTable integrity checker is only supported on amd64")
+		return
+	}
```
rafaeldtinoco, line 35: This is a no go for me. The same mechanism would work in arm64 (The only arch we support) and `core_arm64.go` could be changed in the same or in a similar way.

docs/docs/events/builtin/extra/hooked_syscalls.md
```diff
@@ -1,71 +0,0 @@
...
-#### Purpose
-Detection of syscall hooking.
-
-## Example Use Case
-The `hooked_syscalls` event could be used as part of a broader system integrity monitoring solution. For example, a security engineer could use it to raise alerts or run further investigations if unexpected syscall hooking activities are detected. This could aid in the early detection and mitigation of malware or rootkit infections.
-Example:
-
-```console
-tracee -e hooked_syscalls.args.check_syscalls=<syscall>,<syscall>,...`
-```
```
rafaeldtinoco, line 65: This is good. We have to remove the "args" working as "input mechanism" for detection logics. Thanks for that.

OriGlassman:
> I have tested hooking a syscall with:
> 
> https://github.com/rafaeldtinoco/hijack/blob/main/hijack.c
> 
> and, despite being able to:
> 
> ```
> [22372.068909] uname() intercepted!
> ```
> 
> I could not get a detection. Is the command line:
> 
> ```
> sudo ./dist/tracee --output json --events hooked_syscall
> ```
> 
> enough ?

The event is periodic. If you don't want to wait a few minutes, try running it AFTER the table is hooked, as it runs immediately when tracee starts up.

OriGlassman:
> waitpid

Do we support 32bit? these are only 32 bit system syscalls

rafaeldtinoco:
> > waitpid
> 
> Do we support 32bit? these are only 32 bit system syscalls

Yep, but maybe because of internal system call version routing logic (32-bit userland vs 64-bit userland) the 64-bit symbols need to exist as well :\. I do have them enabled:

```
$ sudo cat /proc/kallsyms | grep -E "sys_waitpid"
ffffffff870a3d80 T __x64_sys_waitpid
ffffffff870a3da0 T __ia32_sys_waitpid

sudo cat /proc/kallsyms | grep -E "_[e|s]text"
ffffffff87000000 T _stext
ffffffff88001f02 T _etext
```

Either way, I get "fake" hook detections for those syscalls. If you decide to consider, blacklist, etc, up to you.

rafaeldtinoco:
Sorry, I missed the timer logic for the detection earlier on. Now:

```console
$ sudo ./dist/tracee --output json --events hooked_syscall
{"timestamp":1697762297084454645,"threadStartTime":1697762293432704767,"processorId":2,"processId":306503,"cgroupId":4846,"threadId":306515,"parentProcessId":306502,"hostProcessId":306503,"hostThreadId":306515,"hostParentProcessId":306502,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"tracee","executable":{"path":""},"hostName":"rugged","containerId":"","container":{},"kubernetes":{},"eventId":"2017","eventName":"hooked_syscall","matchedPolicies":[""],"argsNum":4,"returnValue":0,"syscall":"","stackAddresses":[0],"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":2644688824,"processEntityId":1587116564,"parentEntityId":4262210630,"args":[{"name":"syscall_name","type":"const char*","value":"uname"},{"name":"hooked.address","type":"const char*","value":"ffffffffc1758000"},{"name":"hooked.function_name","type":"const char*","value":"hooked_uname"},{"name":"hooked.owner","type":"const char*","value":"hijack"}]}
```

looks good for amd64 (haven't tested yet arm64, just the tracee execution). 

I need to check why the check-code logic isn't passing (its passing locally so it has probably something to do with the runner picked).

rafaeldtinoco:
Now sure why tests are broken here (checking format), but after a rebase they pass (https://github.com/aquasecurity/tracee/pull/3592). Also, there is an eBPF fix for kernels 6.5+ so this PR would need a rebase either way. I'll close this PR and merge that one if tests pass there (based on an offline msg of you saying you were done with this PR).

rafaeldtinoco:
Being merged at https://github.com/aquasecurity/tracee/pull/3592 because of rebase needs and small test fix.

PR #3550, opened 2023-10-03
fix: add missing types

josedonizetti:
This PR creates the network types for grpc based on https://github.com/aquasecurity/tracee/blob/main/types/trace/network_trace.go


rafaeldtinoco (approved):
LGTM. Events seem like existing ones, the types will have to be translated anyway, so its all good on my side.

PR #3551, opened 2023-10-04
feat(events): create access_remote_vm event

### LINKED ISSUE #3518: Writing to a process memory event
We need an event that is triggered by writing to another memory process, using /proc/<pid>/mem for example.
It should include internal information about the memory.

AlonZivony:

An event for accessing the memory of a process externally (can be the same process) by the mem file of the process in procfs.

789bd6940 **test(events): add e2e test to access_remote_vm**
fd6e01ad3 **feat(events): create access_remote_vm event**
0854a889e **feat(ebpf): support 7 arguments saving for kretprobe**
491ba0491 **feat(events): add probe relevance attribute**

Fix #3518

### DIFF
--- a/docs/docs/events/builtin/extra/access_remote_vm.md
+++ b/docs/docs/events/builtin/extra/access_remote_vm.md
@@ -1,0 +1,34 @@
+# access_remote_vm
+
+## Intro
+access_remote_vm - gain access to the virtual memory of a separate process through the use of the procfs mem file.
+
+## Description
+This event marks access attempt of a process to the virtual memory of another process using the procfs mem file associated with that specific process (/proc/<pid>/mem).
+It is a more elaborated event than the `security_file_open` of the mem file.
+
+## Arguments
+* `remote_pid`: `int`[K] - PID of the process the memory area belongs to.
+* `start_address`: `void *`[K] - Start address of the operation.
+* `gup_flags`: `unsigned int`[K] - Flags for get_user_pages operation.
+* `vm_flags`: `unsigned long`[K] - Virtual memory flags.
+* `mapped.path`: `const char*`[K] - Path of the mapped file, or the name of memory area if no file is mapped.
+* `mapped.device_id`: `dev_t`[K,OPT] - Device ID of the mapped file.
+* `mapped.inode_number`: `unsigned long`[K,OPT] - Inode number of the mapped file.
+* `mapped.ctime`: `unsigned long`[K,OPT] - Creation time of the mapped file.
+
+## Hooks
+### get_user_pages_remote
+#### Type
+kprobe + kretprobe
+#### Purpose
+The main function that implements the access to the virtual memory area of the other process.
+
+### generic_access_phys
+#### Type
+kprobe
+#### Purpose
+A fallback function, implementing the `access` method of the `vma_operations` struct for most of the vmas. It is used to access special memory areas.
+
+## Related Events
+`security_file_open`,`security_mmap_file`,`vfs_write`,`vfs_writev`,`vfs_read`,`vfs_readv`
--- a/pkg/ebpf/c/common/memory.h
+++ b/pkg/ebpf/c/common/memory.h
@@ -4,6 +4,7 @@
 #include <vmlinux.h>
 
 #include <common/common.h>
+#include <common/ksymbols.h>
 
 // PROTOTYPES
 
@@ -12,7 +13,11 @@ statfunc unsigned long get_arg_start_from_mm(struct mm_struct *);
 statfunc unsigned long get_arg_end_from_mm(struct mm_struct *);
 statfunc unsigned long get_env_start_from_mm(struct mm_struct *);
 statfunc unsigned long get_env_end_from_mm(struct mm_struct *);
+statfunc struct task_struct *get_owner_task_from_mm(struct mm_struct *);
+statfunc struct file *get_mapped_file_from_vma(struct vm_area_struct *);
+statfunc struct mount *real_mount(struct vfsmount *);
 statfunc unsigned long get_vma_flags(struct vm_area_struct *);
+statfunc int get_vma_location(struct vm_area_struct *, size_t, char *);
 
 // FUNCTIONS
 
@@ -41,6 +46,16 @@ statfunc unsigned long get_env_end_from_mm(struct mm_struct *mm)
     return BPF_CORE_READ(mm, env_end);
 }
 
+statfunc struct task_struct *get_owner_task_from_mm(struct mm_struct *mm)
+{
+    return BPF_CORE_READ(mm, owner);
+}
+
+statfunc struct file *get_mapped_file_from_vma(struct vm_area_struct *vma)
+{
+    return BPF_CORE_READ(vma, vm_file);
+}
+
 statfunc unsigned long get_vma_flags(struct vm_area_struct *vma)
 {
     return BPF_CORE_READ(vma, vm_flags);
@@ -49,6 +64,66 @@ statfunc unsigned long get_vma_flags(struct vm_area_struct *vma)
 statfunc struct mount *real_mount(struct vfsmount *mnt)
 {
     return container_of(mnt, struct mount, mnt);
+}
+
+// NOTE: The function below use "special_mapping_vmops" and "legacy_special_mapping_vmops"
+// symbols. Ensure these symbols are available when invoking this function.
+
+// Returns true if the specified Virtual Memory Area (VMA) is a special mapping, such as
+// [vdso]. Returns false otherwise.
+statfunc bool is_special_mapping(struct vm_area_struct *vma)
+{
+    char special_mapping_vmops_symbol[22] = "special_mapping_vmops";
+    char legacy_special_mapping_vmops_symbol[29] = "legacy_special_mapping_vmops";
+
+    void *special_mapping_vmops_addr = get_symbol_addr(special_mapping_vmops_symbol);
+    void *legacy_special_mapping_vmops_addr = get_symbol_addr(legacy_special_mapping_vmops_symbol);
+
+    const struct vm_operations_struct *vm_ops = BPF_CORE_READ(vma, vm_ops);
+
+    if (!vm_ops)
+        return false;
+
+    if (vm_ops == special_mapping_vmops_addr || vm_ops == legacy_special_mapping_vmops_addr)
+        return true;
+
+    return false;
+}
+
+// Returns the descriptive name of the specified Virtual Memory Area (VMA). This function
+// identifies special VMAs like [stack], [heap], and [vdso]. It does not resolve VMAs
+// mapped to files; such VMAs require alternative methods for detail retrieval.
+statfunc int get_vma_location(struct vm_area_struct *vma, size_t max_len, char *location)
+{
+    // Avoid buffer overflow and satisfy the verifier.
+    if (max_len < 7) {
+        return 0;
+    }
+
+    struct mm_struct *vm_mm = BPF_CORE_READ(vma, vm_mm);
+
+    if (vm_mm) {
+        u64 start_stack = BPF_CORE_READ(vm_mm, start_stack);
+        u64 start_brk = BPF_CORE_READ(vm_mm, start_brk);
+        u64 brk = BPF_CORE_READ(vm_mm, brk);
+        u64 vm_start = BPF_CORE_READ(vma, vm_start);
+        u64 vm_end = BPF_CORE_READ(vma, vm_end);
+
+        if (vm_start <= start_stack && vm_end >= start_stack) {
+            __builtin_memcpy(location, "[stack]", 7);
+
+        } else if (vm_start <= brk && vm_end >= start_brk) {
+            __builtin_memcpy(location, "[heap]", 6);
+
+        } else if (is_special_mapping(vma)) {
+            struct vm_special_mapping *special_mapping = BPF_CORE_READ(vma, vm_private_data);
+            const char *name = BPF_CORE_READ(special_mapping, name);
+            if (name)
+                bpf_probe_read_str(location, max_len, name);
+        }
+    }
+
+    return 0;
 }
 
 #endif
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -4966,6 +4966,132 @@ int BPF_KPROBE(trace_ret_exec_binprm2)
     return events_perf_submit(&p, PROCESS_EXECUTION_FAILED, ret);
 }
 
+// Submit "access_remote_vm" event for the given vma and info.
+statfunc int submit_access_remote_vm(program_data_t *p,
+                                     struct vm_area_struct *vma,
+                                     void *address,
+                                     unsigned int gup_flags)
+{
+    unsigned long vm_flags = get_vma_flags(vma);
+    struct mm_struct *mm = BPF_CORE_READ(vma, vm_mm);
+    file_info_t file_info = {};
+
+    char location[64] = {0};
+
+    // Get mapped memory information
+    struct file *file = get_mapped_file_from_vma(vma);
+    file_info = get_file_info(file);
+
+    // If there is no file that is mapped to the vma, try to get memory location name instead.
+    if (!file_info.pathname_p)
+        get_vma_location(vma, sizeof(location), location);
+
+    struct task_struct *task = get_owner_task_from_mm(mm);
+    u32 pid = get_task_host_tgid(task);
+
+    save_to_submit_buf(&p->event->args_buf, &pid, sizeof(u32), 0);
+    save_to_submit_buf(&p->event->args_buf, &address, sizeof(void *), 1);
+    save_to_submit_buf(&p->event->args_buf, &gup_flags, sizeof(unsigned int), 2);
+    save_to_submit_buf(&p->event->args_buf, &vm_flags, sizeof(unsigned long), 3);
+    if (file_info.pathname_p != NULL) {
+        save_str_to_buf(&p->event->args_buf, file_info.pathname_p, 4);
+        save_to_submit_buf(&p->event->args_buf, &file_info.id.device, sizeof(dev_t), 5);
+        save_to_submit_buf(&p->event->args_buf, &file_info.id.inode, sizeof(unsigned long), 6);
+        save_to_submit_buf(&p->event->args_buf, &file_info.id.ctime, sizeof(u64), 7);
+    } else {
+        save_str_to_buf(&p->event->args_buf, location, 4);
+    }
+
+    return events_perf_submit(p, ACCESS_REMOTE_VM, 0);
+}
+
+SEC("kprobe/generic_access_phys")
+int BPF_KPROBE(trace_generic_access_phys)
+{
+    program_data_t p = {};
+    if (!init_program_data(&p, ctx))
+        return 0;
+
+    if (!should_trace(&p))
+        return 0;
+
+    struct vm_area_struct *vma = (struct vm_area_struct *) PT_REGS_PARM1(ctx);
+    unsigned long addr = (unsigned long) PT_REGS_PARM2(ctx);
+    int write = (int) PT_REGS_PARM5(ctx);
+
+    return submit_access_remote_vm(&p, vma, (void *) addr, write);
+}
+
+SEC("kprobe/get_user_pages_remote")
+TRACE_ENT_FUNC(get_user_pages_remote, ACCESS_REMOTE_VM);
+
+// The "get_user_pages_remote" function has changed in kernel 6.5.
+// The part of finding the relevant vma was extracted to the new function
+// "get_user_page_vma_remote", leaving the logic of finding the relevant pages in this function. The
+// vma is what interest us in the ACCESS_REMOTE_VM, so this probe should only be used for kernels
+// prior to 6.5.
+SEC("kretprobe/get_user_pages_remote")
+int BPF_KPROBE(trace_ret_get_user_pages_remote)
+{
+    args_t saved_args;
+    if (load_args(&saved_args, ACCESS_REMOTE_VM) != 0) {
+        // missed entry or not traced
+        return 0;
+    }
+    del_args(ACCESS_REMOTE_VM);
+
+    program_data_t p = {};
+    if (!init_program_data(&p, ctx))
+        return 0;
+
+    if (!should_trace(&p))
+        return 0;
+
+    int ret_val = PT_REGS_RC(ctx);
+    // If the page cannot be pinned the return code would be <= 0, this could mean that
+    // the page is a memory mapped page, for example. In this case, the access to that
+    // page has to be done through "vma->vm_ops->access", which we trace for most cases
+    // (all normal cases) using generic_access_phys. Let the flow continue so it would be
+    // traced there.
+    if (ret_val <= 0)
+        return 0;
+
+    unsigned long address;
+    unsigned int gup_flags;
+    struct vm_area_struct **vmas;
+    struct vm_area_struct *vma = NULL;
+    unsigned long vm_flags = 0;
+    file_info_t file_info = {};
+
+    // The first argument of the function was removed in kernel version 5.9.
+    // We need to adjust the arguments retrival to the kernels to avoid misinformation.
+    // In kernel 5.9 the type bpf_iter_seq_array_map_info was added. By checking
+    // its existence, we can determine the current kernel version.
+    if (!bpf_core_type_exists(struct bpf_iter_seq_array_map_info)) { // kernel version < 5.9
+        address = (unsigned long) saved_args.args[2];
+        gup_flags = (unsigned int) saved_args.args[4];
+        vmas = (struct vm_area_struct **) saved_args.args[6];
+    } else { // kernel version >= 5.9
+        address = (unsigned long) saved_args.args[1];
+        gup_flags = (unsigned int) saved_args.args[3];
+        vmas = (struct vm_area_struct **) saved_args.args[5];
+    }
+
+    // Without a vma we can't get enough information for the event. Moreover, this will
+    // filter out the calls for it from the execve flow.
+    if (!vmas)
+        return 0;
+
+    // In the flow of interest, via the access_remote_vm function, the logic iterate over
+    // the pages 1 by 1. This means that 1 and only 1 vma would be resolved here if the
+    // function succeed.
+    bpf_core_read(&vma, sizeof(void *), vmas);
+    if (!vma)
+        return 0;
+
+    return submit_access_remote_vm(&p, vma, (void *) address, gup_flags);
+}
+
 // clang-format off
 
 // Network Packets (works from ~5.2 and beyond)
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -5093,10 +5093,36 @@ int BPF_KPROBE(trace_ret_get_user_pages_remote)
 }
 
 // Callback function for bpf_find_vma
-static long kprobe_vma_callback(struct task_struct *task, struct vm_area_struct *vma, void *callback_ctx) {
+statfunc long kprobe_vma_callback(struct task_struct *task, struct vm_area_struct *vma, void *callback_ctx) {
     struct vm_area_struct **out_vma = (struct vm_area_struct **)callback_ctx;
     if (vma)
         *out_vma = vma;
+    return 0;
+}
+
+typedef struct vma_iteration_ctx {
+    program_data_t *p;
+    struct task_struct *task;
+    u64 addr;
+    u64 len;
+    unsigned int gup_flags;
+} vma_iteration_ctx_t;
+
+statfunc long iterate_submit_access_remote_vm(u32 index, void *ctx) {
+    vma_iteration_ctx_t *iter_ctx = (vma_iteration_ctx_t *)ctx;
+    struct vm_area_struct *vma;
+    int ret = bpf_find_vma(iter_ctx->task, iter_ctx->addr, (void *)kprobe_vma_callback, (void *)&vma, 0);
+    if (ret != 0)
+        return 1;
+
+    submit_access_remote_vm(iter_ctx->p, vma, (void *) iter_ctx->addr, iter_ctx->gup_flags);
+    u64 vm_end = BPF_CORE_READ(vma, vm_end);
+    u64 len_in_vma = vm_end - iter_ctx->addr;
+    len_in_vma += 1; // Range access include last byte - addresses range of 0-4 consist of 5 bytes.
+    if (iter_ctx->len <= len_in_vma)
+        return 1;
+    iter_ctx->addr += len_in_vma;
+    iter_ctx->len -= len_in_vma;
     return 0;
 }
 
@@ -5112,18 +5138,16 @@ int BPF_KPROBE(trace_access_remote_vm)
 	
 	if (!should_submit(ACCESS_REMOTE_VM, p.event))
         return 0;
-	
-	struct vm_area_struct *vma;
+
 	struct mm_struct *mm = (struct mm_struct *)PT_REGS_PARM1(ctx);
-	struct task_struct *task = get_owner_task_from_mm(mm);
-	u64 addr = (u64)PT_REGS_PARM2(ctx);
-	unsigned int gup_flags = (unsigned int)PT_REGS_PARM5(ctx);
+	vma_iteration_ctx_t vma_iteration_ctx = {};
+	vma_iteration_ctx.task = get_owner_task_from_mm(mm);
+	vma_iteration_ctx.addr = (u64)PT_REGS_PARM2(ctx);
+	vma_iteration_ctx.len = (u64)PT_REGS_PARM4(ctx);
+	vma_iteration_ctx.gup_flags = (unsigned int)PT_REGS_PARM5(ctx);
+	vma_iteration_ctx.p = &p;
 
-	int ret = bpf_find_vma(task, addr, (void *)kprobe_vma_callback, (void *)&vma, 0);
-    if (ret != 0)
-        return 0;
-
-	return submit_access_remote_vm(&p, vma, (void *) addr, gup_flags);
+    return bpf_loop(MAX_VMA_ITERATIONS, (void *)iterate_submit_access_remote_vm, &vma_iteration_ctx, 0);
 }
 
 // clang-format off
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -5092,6 +5092,40 @@ int BPF_KPROBE(trace_ret_get_user_pages_remote)
     return submit_access_remote_vm(&p, vma, (void *) address, gup_flags);
 }
 
+// Callback function for bpf_find_vma
+static long kprobe_vma_callback(struct task_struct *task, struct vm_area_struct *vma, void *callback_ctx) {
+    struct vm_area_struct **out_vma = (struct vm_area_struct **)callback_ctx;
+    if (vma)
+        *out_vma = vma;
+    return 0;
+}
+
+SEC("kprobe/get_user_pages_remote")
+int BPF_KPROBE(trace_access_remote_vm)
+{
+	program_data_t p = {};
+    if (!init_program_data(&p, ctx))
+        return 0;
+
+    if (!should_trace(&p))
+        return 0;
+	
+	if (!should_submit(ACCESS_REMOTE_VM, p.event))
+        return 0;
+	
+	struct vm_area_struct *vma;
+	struct mm_struct *mm = (struct mm_struct *)PT_REGS_PARM1(ctx);
+	struct task_struct *task = get_owner_task_from_mm(mm);
+	u64 addr = (u64)PT_REGS_PARM2(ctx);
+	unsigned int gup_flags = (unsigned int)PT_REGS_PARM5(ctx);
+
+	int ret = bpf_find_vma(task, addr, (void *)kprobe_vma_callback, (void *)&vma, 0);
+    if (ret != 0)
+        return 0;
+
+	return submit_access_remote_vm(&p, vma, (void *) addr, gup_flags);
+}
+
 // clang-format off
 
 // Network Packets (works from ~5.2 and beyond)
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -70,6 +70,7 @@ env:
     PROCTREE_DATA_SOURCE
     DNS_DATA_SOURCE
     WRITABLE_DATA_SOURCE
+    ACCESS_REMOTE_VM
 jobs:
   #
   # DOC VERIFICATION
@@ -326,6 +327,7 @@ jobs:
       - generate-matrix
     runs-on: "github-self-hosted_ami-${{ matrix.ami }}_${{ github.event.number }}-${{ github.run_id }}_${{ matrix.sufix }}"
     strategy:
+      fail-fast: false
       matrix:
         include: ${{fromJson(needs.generate-matrix.outputs.matrix01)}}
     env:
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/IBM/fluent-forward-go v0.2.1
 	github.com/Masterminds/sprig/v3 v3.2.3
 	github.com/aquasecurity/libbpfgo v0.6.0-libbpf-1.3
-	github.com/aquasecurity/libbpfgo/helpers v0.4.6-0.20231123142329-37c4b843a539
+	github.com/aquasecurity/libbpfgo/helpers v0.4.6-0.20231127142843-5e798c565c08
 	github.com/aquasecurity/tracee/api v0.0.0-20231222010915-0f73aad36da9
 	github.com/aquasecurity/tracee/types v0.0.0-20231231104405-a33891f29d17
 	github.com/containerd/containerd v1.7.0
--- a/go.sum
+++ b/go.sum
@@ -67,6 +67,8 @@ github.com/aquasecurity/libbpfgo v0.6.0-libbpf-1.3 h1:mhDe1mAZR80LjnsCnteS+R2/Ee
 github.com/aquasecurity/libbpfgo v0.6.0-libbpf-1.3/go.mod h1:0rEApF1YBHGuZ4C8OYI9q5oDBVpgqtRqYATePl9mCDk=
 github.com/aquasecurity/libbpfgo/helpers v0.4.6-0.20231123142329-37c4b843a539 h1:axIHZ3la2/wcqMYO9TUyKO/lMGYizEKyNIodbwQBOkE=
 github.com/aquasecurity/libbpfgo/helpers v0.4.6-0.20231123142329-37c4b843a539/go.mod h1:1fGKke5pgH4xYvZ7HqDbLSi/R5zfRFH2K+c9kLp9L34=
+github.com/aquasecurity/libbpfgo/helpers v0.4.6-0.20231127142843-5e798c565c08 h1:hlnYuC6sWi3yYi9HNsHLemDfbf6F787GCAkP7kwnqBc=
+github.com/aquasecurity/libbpfgo/helpers v0.4.6-0.20231127142843-5e798c565c08/go.mod h1:1fGKke5pgH4xYvZ7HqDbLSi/R5zfRFH2K+c9kLp9L34=
 github.com/aquasecurity/tracee/api v0.0.0-20231222010915-0f73aad36da9 h1:1y+/rhVRNZK8s/ZyqXhtqjVLbKs/F9jPjBUn9jcujQs=
 github.com/aquasecurity/tracee/api v0.0.0-20231222010915-0f73aad36da9/go.mod h1:QJG2PABXucOsFVO85tQsKxV4c1GUhcjww/Kw+Wv7Y/c=
 github.com/aquasecurity/tracee/types v0.0.0-20231231104405-a33891f29d17 h1:tgcl7BuyBLDng+47KKhyOB9gdsq3V9dKoapNyRDchwg=
--- a/pkg/cmd/initialize/callbacks.go
+++ b/pkg/cmd/initialize/callbacks.go
@@ -24,7 +24,7 @@ var (
 	// 3. %s = check_map_func_compatibility (function name)
 	// 4. %x = offset (ignored in this check)
 	// 5. %s = No such file or directory
-	libbpfgoKprobePerfEventRegexp = regexp.MustCompile(`libbpf:.*prog '(?:trace_check_map_func_compatibility|trace_utimes_common)'.*failed to create kprobe.*perf event: No such file or directory`)
+	libbpfgoKprobePerfEventRegexp = regexp.MustCompile(`libbpf:.*prog '(?:trace_check_map_func_compatibility|trace_utimes_common|trace_generic_access_phys)'.*failed to create kprobe.*perf event: No such file or directory`)
 
 	// triggered by: libbpf/src/libbpf.c->bpf_program__attach_fd()
 	// "libbpf: prog '%s': failed to attach to %s: %s\n"
--- a/pkg/ebpf/c/common/arguments.h
+++ b/pkg/ebpf/c/common/arguments.h
@@ -42,6 +42,7 @@ statfunc int load_args(args_t *args, u32 event_id)
     args->args[3] = saved_args->args[3];
     args->args[4] = saved_args->args[4];
     args->args[5] = saved_args->args[5];
+    args->args[6] = saved_args->args[6];
 
     return 0;
 }
--- a/pkg/ebpf/c/common/consts.h
+++ b/pkg/ebpf/c/common/consts.h
@@ -20,6 +20,8 @@
 #define MAX_SYS_CALL_TABLE_SIZE 500
 #define MAX_MEM_DUMP_SIZE   127
 
+#define MAX_VMA_ITERATIONS 20
+
 
 #define MAX_STR_ARR_ELEM      38 // TODO: turn this into global variables set w/ libbpfgo
 #define MAX_ARGS_STR_ARR_ELEM 15
--- a/pkg/ebpf/c/common/probes.h
+++ b/pkg/ebpf/c/common/probes.h
@@ -11,13 +11,13 @@
 #include <common/filtering.h>
 
 #if defined(bpf_target_x86)
-#define PT_REGS_PARM7(ctx) \
-      ({ \
-          unsigned long reg; \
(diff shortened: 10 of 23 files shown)


pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4861,6 +4861,104 @@ int BPF_KPROBE(trace_ret_exec_binprm2)
...
+    if (!should_trace(&p))
+        return 0;
+
+    struct vm_area_struct *vma = (struct vm_area_struct *) PT_REGS_PARM1(ctx);
+    unsigned long addr = (unsigned long) PT_REGS_PARM2(ctx);
+    int write = (int) PT_REGS_PARM5(ctx);
+    return submit_access_remote_vm(&p, vma, (void *) addr, write);
+}
+
+SEC("kprobe/get_user_pages_remote")
```
rafaeldtinoco, line 4919: Okay, I bet this is:

access_remote_vm -> __access_remote_vm -> get_user_page_vma_remote -> get_user_pages_remote.

Call hierarchy shows that even `get_user_pages_remote` is only called when `access_remote_vm` is called and it is almost always only called when `/proc/[pid]/mem` file is read/write (through mem_read and mem_write handlers for the proc_mem_operations). All this to say that I think we will be fine performance wise for this event if it become default in Tracee or CNDR (other execution paths usually come from using PTRACE in debugging step-by-step calls and similar).

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4861,6 +4861,104 @@ int BPF_KPROBE(trace_ret_exec_binprm2)
...
+
+    program_data_t p = {};
+    if (!init_program_data(&p, ctx))
+        return 0;
+
+    if (!should_trace(&p))
+        return 0;
+
+    int ret_val = PT_REGS_RC(ctx);
+    // if failed, then the operation is done using vm_ops->acess instead and will be traced by it
```
rafaeldtinoco, line 4940: Improve this comment saying that "if the page cannot be pinned it could meant that the page is a memory mapped page, for example, and then the access to that page has to be done through vma->vm_ops->access". NOte that there is a mispell in  `->acess` in the current sentence as well.

rafaeldtinoco: You can mention that thats why the other probe also exists for this event.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4861,6 +4861,104 @@ int BPF_KPROBE(trace_ret_exec_binprm2)
...
+    struct vm_area_struct **vmas = (struct vm_area_struct **) saved_args.args[5];
+    struct vm_area_struct *vma = NULL;
+    unsigned long vm_flags = 0;
+    file_info_t file_info = {};
+    // Without a vma we can't get enough information for the event.
+    // Moreover, this will filter out the calls for it from the execve flow.
+    if (!vmas)
+        return 0;
+
+    bpf_core_read(&vma, sizeof(void *), vmas);
```
rafaeldtinoco, line 4955: You can mention that you're reading only the first VMA because `access_remote_vm->get_user_pages_remote` logic walks pages 1 by 1. It looks too detailed but it helps people with less experience in MM & VMA to understand why you're doing what you're doing.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4861,6 +4861,104 @@ int BPF_KPROBE(trace_ret_exec_binprm2)
     return events_perf_submit(&p, PROCESS_EXECUTION_FAILED, ret);
 }
 
+// Submit "access_remtoe_vm" event for the given vma and info.
```
rafaeldtinoco, line 4864: nit: spell on "remote" word

pkg/ebpf/c/common/memory.h
```diff
@@ -41,11 +43,53 @@ statfunc unsigned long get_env_end_from_mm(struct mm_struct *mm)
...
 {
     return BPF_CORE_READ(vma, vm_flags);
 }
 
+// Return the name of the location related to the given vma. For example [stack], [heap] and [vdso].
+// This function doesn't resolve file mapped vmas, so if the vma is a memory mapped file its
+// details should be collected in a different way.
+statfunc int get_vma_location(struct vm_area_struct *vma, size_t max_len, char *location)
+{
+    // Avoid buffer overflow and staisfy the verifier.
```
rafaeldtinoco, line 66: nit: spell on satisfy.

pkg/ebpf/c/common/memory.h
```diff
@@ -41,11 +43,53 @@ statfunc unsigned long get_env_end_from_mm(struct mm_struct *mm)
...
+        if (vm_start <= start_stack && vm_end >= start_stack) {
+            __builtin_memcpy(location, "[stack]", 7);
+        } else if (vm_start <= brk && vm_end >= start_brk) {
+            __builtin_memcpy(location, "[heap]", 6);
+        } else {
+            struct vm_special_mapping *special_mapping = BPF_CORE_READ(vma, vm_private_data);
+            const char *name = BPF_CORE_READ(special_mapping, name);
+            if (name) {
+                bpf_probe_read_str(location, max_len, name);
+            }
```
rafaeldtinoco, line 87: I can see that you're trying to pick VDSO name through the "special mapping" used by `vm_private_data`. But kernel has a function `vma_is_special_mapping()` that tells it if its a special mapping or not before.

Isn't this a bit risky ? You're tracing ANY read/write to the /proc/[pid]/mem (virtually all virtual address space from the process). Imagine if the process has MMIO pages with a device that has `vma->vm_private_data` pointing to something else (that isn't `vm_special_mapping`). You can argue its safe, because its only reads, but you would get trash that could by accident fuzz our internal logic, no ? I can see that the `vm_private_data` has different types for GPUs, SCSI, USB, frame buffer device and the hugetlbfs (probably others).

ALso, I'm not sure its worth checking if `vma->vm_ops` points to special mapping functions in order to do that BPF_CORE_READ. WDYT ?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4861,6 +4861,104 @@ int BPF_KPROBE(trace_ret_exec_binprm2)
...
+        save_to_submit_buf(&p->event->args_buf, &file_info.id.inode, sizeof(unsigned long), 6);
+        save_to_submit_buf(&p->event->args_buf, &file_info.id.ctime, sizeof(u64), 7);
+    } else {
+        save_str_to_buf(&p->event->args_buf, location, 4);
+    }
+
+    return events_perf_submit(p, ACCESS_REMOTE_VM, 0);
+}
+
+SEC("kprobe/generic_access_phys")
```
rafaeldtinoco, line 4903: Okay, this **could** be problematic if we run workloads doing userland MMAP IO to PCI devices, for example (an environment with GPUs through PCI, maybe ? Don't know). I think we should add a note here. If you check, the `vm_operations->access` function pointer can be set to `generic_access_phys` at some points, and, besides the char device driver responsible for /proc/pid/mem, there is also the MMIO code for PCI devices (when mapped in userland to bypass kernel). Again, I don't know if that would be true, but a note here for further awareness would be good.

rafaeldtinoco:
I loved the event! It LGTM, definitely. 

There are just minor spell nits, some comments to be added to make it easier to understand your intents (if my interpretation is correct).

A small "point to clear" would be the VMA naming strategy/logic. I left a question for you there.

Also, If you could add a simple `e2e-inst` trigger for this type of event I would also appreciate (like we already spoke).

Nice work!

pkg/ebpf/c/common/memory.h
AlonZivony: Yea your are right that this is not consistent way.
The `vma->vm_ops` sound pretty good solution to me, and its not hard to check for it.
I will try to dig into it and think if there is another easy way to satisfy this concern.

pkg/ebpf/c/tracee.bpf.c
AlonZivony: I will write it in a comment, but from what I have seen the `access` method isn't called by many flows in the kernel, so I don't think the load will be significant.
For example, in kernel 5.16 it is used only in  `__access_remote_vm` and in `kernfs_vma_access`, which is also an implementation of the `vma_ops->access`.
Notice that even in the declaration of the struct, it is written that this method is only called by `__access_remote_vm`:
```C
	/* called by access_process_vm when get_user_pages() fails, typically
	 * for use by special VMAs. See also generic_access_phys() for a generic
	 * implementation useful for any iomem mapping.
	 */
	int (*[access](https://elixir.bootlin.com/linux/latest/C/ident/access))(struct [vm_area_struct](https://elixir.bootlin.com/linux/latest/C/ident/vm_area_struct) *[vma](https://elixir.bootlin.com/linux/latest/C/ident/vma), unsigned long addr,
		      void *buf, int len, int [write](https://elixir.bootlin.com/linux/latest/C/ident/write));
```
So I don't think we need to be concerned about the performance here more than in `__access_remote_vm`.

docs/docs/events/builtin/extra/access_remote_vm.md
```diff
@@ -0,0 +1,34 @@
+# access_remote_vm
+
+## Intro
+access_remote_vm - gain access to the virtual memory of a separate process through the use of the procfs mem file.
+
+## Description
+This event marks access attempt of a process to the virtual memory of another process using the procfs mem file associated with that specific process (/proc/<pid>/mem).
```
rafaeldtinoco, line 7: There is a signature in the open source called "proc_mem_code_injection" that checks for /proc/pid/mem file openings with WRITE flags (which @itaysk referred to some time ago). 

This signature is way more specific (picking a process doing reads from /proc/<pid>/mem and mapping to its own virtual address space).

While the 2 are different, with this one recognizing parts of the other process memory space that are being mapped, dealing with memory mapped addresses, etc, ... the first one (signature) would be triggered even before this event would be triggered, right ? 

So, in theory, this events are specific to a behavior that could be caught by that signature (probably with WAY less overhead than the signature). Could you confirm the understanding ?

AlonZivony: You are right - this is much easier way to implement these signatures.
In addition, it also allow to catch other flows in the kernel that access a process memory as a 3rd party - for example during execution, and with uprobes.
Above all else, this event give you much detailed information about the access, like the name of the memory mapped area accessed this way or the mmapped file.

rafaeldtinoco: I believe the change should be in a separate PR and, if the other events can use a more performatic hook I'm all in (as checking for all opens is less performatic). Thanks for clarifying!

rafaeldtinoco:
@AlonZivony mind rebasing this one and solving the little conflict it has ?

rafaeldtinoco:
@AlonZivony 

image

rafaeldtinoco (changes requested):
I left you https://github.com/AlonZivony/tracee/pull/3 with 3 commits there. But, after my super nit change, on the first patch, the tests still fails.

I believe the test should use "dd" instead of "tail" to access the stack address from the mem file.

Something like: 
dd if=/proc/pid/mem of=/dev/null bs=4k count=1 skip=offset
to read the first page from the stack address ?

I also believe there is a timing issue - between tracee to start and the test to be executed - that makes it fail in 1/2 of the runs (per my tests it failed in lots of runners). 

image


About timing issue, if you check the proctree example, I had a sleep in the first time the event happens to deal with those kind of things (but the sleep could also be in beginning of the test script, etc, needs checking, Ill leave that to you).

AlonZivony:
> I believe the test should use "dd" instead of "tail" to access the stack address from the mem file.
> 
> Something like: dd if=/proc/pid/mem of=/dev/null bs=4k count=1 skip=offset to read the first page from the stack address ?
 
So this was my first try in doing it, but the problem here is that `dd` needs to work on a round address, else it fails.
But sometimes the stack address I got was not rounded like `dd` needs. So I figured using `tail` with jump of a specific offset will do the trick, even if it gets an error at the end of the read..

About the timing issue, I will try to figure it out.

AlonZivony:
@rafaeldtinoco the tests seems to work now with your changes.
Do you want to rerun the tests to see if it is persistent?
I will push again we we finish giving you also credit for the code.

AlonZivony:
Thanks Rafa for solving the conflict :)

rafaeldtinoco:
I tried to fix the conflict using web interface but it creates a merge and messes up with history and other things. I'll rebase at: https://github.com/aquasecurity/tracee/pull/3696 (and run tests multiple times). Then Ill merge. @AlonZivony ;)

rafaeldtinoco:
image

Tests passed quickly because there were no tests. The workflow had a bad syntax so the tests did not run. Weird.

rafaeldtinoco:
Let me check what is going on in that PR (just a rebase of this one).

rafaeldtinoco:
Alright, so there is a problem here. The event fails in some kernels and it doesn't in some others (https://github.com/aquasecurity/tracee/actions/runs/6895648514/job/18760277375?pr=3696).

I think there are 2 issues:

1. Related to timing (might happen to any kernel)
2. Event does not seem to work in kernels v6.5.

To give a small test (and not to rely in "tail" or "dd" for testing this, specially because procfs mem file is deprecated), I created the following code just to play with this event:

```C
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <fcntl.h>
#include <unistd.h>
#include <sys/uio.h>

// NOTE: compile with "gcc -D_GNU_SOURCE"

#define BUFFER_SIZE 1024

int find_stack_address(pid_t pid, unsigned long *start, unsigned long *end) {
    char maps_path[256];
    FILE *maps;
    char line[BUFFER_SIZE];

    sprintf(maps_path, "/proc/%d/maps", pid);
    maps = fopen(maps_path, "r");
    if (maps == NULL) {
        perror("fopen");
        return -1;
    }

    while (fgets(line, sizeof(line), maps)) {
        if (strstr(line, "[stack]")) {
            sscanf(line, "%lx-%lx", start, end);
            fclose(maps);
            return 0;
        }
    }

    fclose(maps);
    return -1;
}

int read_stack(pid_t pid, unsigned long start, unsigned long end) {
    int output_fd = open("stack.dump", O_WRONLY | O_CREAT | O_TRUNC, 0666);
    if (output_fd < 0) {
        perror("open");
        return -1;
    }

    char buffer[BUFFER_SIZE];
    struct iovec local[1];
    struct iovec remote[1];

    local[0].iov_base = buffer;
    remote[0].iov_base = (void *)start;
    size_t remaining = end - start;

    while (remaining > 0) {
        local[0].iov_len = remote[0].iov_len = (remaining > BUFFER_SIZE) ? BUFFER_SIZE : remaining;
        ssize_t nread = process_vm_readv(pid, local, 1, remote, 1, 0);

        if (nread <= 0) {
            perror("process_vm_readv");
            close(output_fd);
            return -1;
        }

        write(output_fd, buffer, nread);
        remote[0].iov_base += nread;
        remaining -= nread;
    }

    close(output_fd);
    return 0;
}

int main(int argc, char *argv[]) {
    if (argc != 2) {
        fprintf(stderr, "Usage: %s <pid>\n", argv[0]);
        return EXIT_FAILURE;
    }

    pid_t pid = atoi(argv[1]);
    unsigned long stack_start, stack_end;

    if (find_stack_address(pid, &stack_start, &stack_end) != 0) {
        fprintf(stderr, "Failed to find stack address for PID %d\n", pid);
        return EXIT_FAILURE;
    }

    if (read_stack(pid, stack_start, stack_end) != 0) {
        fprintf(stderr, "Failed to read stack for PID %d\n", pid);
        return EXIT_FAILURE;
    }

    printf("Stack dump completed for PID %d\n", pid);
    return EXIT_SUCCESS;
}
```

And saw that it does not use the execution flow you expected/programmed for this event. YOu can check this by executing:

```console
sudo perf ftrace --graph-opts depth=10 -- ./this-program $$ 2>&1 > ./this-program-trace.txt
```

in a kernel v6.5. You will see that the execution path for `process_vm_readv` (and likely for the procfs inode file operations reader handler for /proc/pid/mem file) is different:

```
 3)               |  __x64_sys_process_vm_readv() {
 3)               |    process_vm_rw() {
 3)               |      process_vm_rw_core.constprop.0() {
 3)               |        find_get_task_by_vpid() {
 3)   0.094 us    |          __rcu_read_lock();
 3)   0.283 us    |          __rcu_read_unlock();
 3)   0.761 us    |        }
 3)               |        mm_access() {
 ...
 3)               |        pin_user_pages_remote() {
 3)               |          __gup_longterm_locked() {
 3)               |            __get_user_pages() {
 3)               |              gup_vma_lookup() {
 3)               |                find_vma() {
 3)   0.098 us    |                  __rcu_read_lock();
 3)   0.099 us    |                  __rcu_read_unlock();
 3)   0.493 us    |                }
 3)   0.670 us    |              }
 ...
```

I would say that the logic accessing the other process VMAs is likely `__get_user_pages()` for recent kernels. So, I believe there are a few things you need to clarify:

- When reading from procfs we're likely using __get_user_pages() as well. I can only see `get_user_pages_remote` being used by `execve` code in v6.5, but haven't looked extensively.

- Is there any case (still) that uses `get_user_pages_remote` ? 

And some other doubts:

- Does this event cover aio_read() case when async I/O is used ?
- What about mmap'ed() reads ?
- What about io_uring method (that might use mmap as well) ?

Question: couldn't an attacker read the procfs mem file using mmap() to that file ? The file operations for "read()" and "mmap()" are different (so the execution path would be different).

## SUMMARY

- fix timing issues in all kernels but new ones
- investigate new hooks to be used in v6.5 kernels

rafaeldtinoco (changes requested):
Per https://github.com/aquasecurity/tracee/pull/3551#issuecomment-1815674510

AlonZivony:
> in a kernel v6.5. You will see that the execution path for process_vm_readv (and likely for the procfs inode file operations reader handler for /proc/pid/mem file) is different

I will address this at first.
You are wrong in assuming that the proc mem file is acting the same as `process_vm_readv`. It acts the same way as it was till now.
Moreover, the `process_vm_readv` is not the flow I address with this event.
The event documentation is also specifying that this event is for accessing of the proc mem file. You use the `process_vm_readv` for it, which is not included in the goal of this event. It can be, but its expanding it, not a requirement. I am not sure it will be wise to do it in the same event, as again, this will serve other purpose than monitoring proc mem file usage.

About `mmap`, I don't think its a problem as the proc mem file doesn't implement a `mmap` file operation.
Lastly, the `aio_read` and `io_uring` will need to both pass through the file operation. So I don't have any concern here either.

Tell me if you have further concerns.

AlonZivony:
Now about the failing tests.
I haven't seen a single test failing till now, so I am not sure to what you are referring. I can't really understand the problem without seeing it, so if you can reset the tests for me to see or send me a test suite that failed I will dig deeper.
However, I don't see why should we have a sleep in the test like in the process tree. There is no state here to initialize. If there is  a timing issue here, it is an issue with all e2e tests and should be fixed by the tests environment. However, I can add for now a sleep if you think it will solve it.

rafaeldtinoco:
> You are wrong in assuming that the proc mem file is acting the same as `process_vm_readv`. It acts the same way as it was till now. Moreover, the process_vm_readv is not the flow I address with this event.

> The event documentation is also specifying that this event is for accessing of the proc mem file. 

Sure, sure, my mistake. I thought the execution path was the same for some reason (too many context switches in my head about diff stuff). Ok, so let me test with a simple tail then and lets see.

> I am not sure it will be wise to do it in the same event, as again, this will serve other purpose than monitoring proc mem file usage.

Alright, but be aware that there are many detections from this signature coming from regular processes (not sure your signature will filter for specific stuff, but systemd, dbus-daemon, etc are all being "detected").

```
{"timestamp":1700404629445561300,"threadStartTime":1700403164364553700,"processorId":2,"processId":388,"cgroupId":500,"threadId":388,"parentProcessId":1,"hostProcessId":388,"hostThreadId":388,"hostParentProcessId":1,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"systemd-journal","executable":{"path":""},"hostName":"jammy","containerId":"","container":{},"kubernetes":{},"eventId":"781","eventName":"access_remote_vm","matchedPolicies":[""],"argsNum":5,"returnValue":0,"syscall":"read","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":1669026443,"processEntityId":1669026443,"parentEntityId":1844877023,"args":[{"name":"remote_pid","type":"int","value":12862},{"name":"start_address","type":"void *","value":"0x7ffe885c2a36"},{"name":"gup_flags","type":"unsigned int","value":32768},{"name":"vm_flags","type":"unsigned long","value":1048947},{"name":"mapped.path","type":"const char*","value":"[stack]"},{"name":"mapped.device_id","type":"dev_t","value":null},{"name":"mapped.inode_number","type":"unsigned long","value":null},{"name":"mapped.ctime","type":"unsigned long","value":null}]}
{"timestamp":1700404629445597200,"threadStartTime":1700403164364553700,"processorId":2,"processId":388,"cgroupId":500,"threadId":388,"parentProcessId":1,"hostProcessId":388,"hostThreadId":388,"hostParentProcessId":1,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"systemd-journal","executable":{"path":""},"hostName":"jammy","containerId":"","container":{},"kubernetes":{},"eventId":"781","eventName":"access_remote_vm","matchedPolicies":[""],"argsNum":5,"returnValue":0,"syscall":"read","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":1669026443,"processEntityId":1669026443,"parentEntityId":1844877023,"args":[{"name":"remote_pid","type":"int","value":12862},{"name":"start_address","type":"void *","value":"0x7ffe885c2a2a"},{"name":"gup_flags","type":"unsigned int","value":32768},{"name":"vm_flags","type":"unsigned long","value":1048947},{"name":"mapped.path","type":"const char*","value":"[stack]"},{"name":"mapped.device_id","type":"dev_t","value":null},{"name":"mapped.inode_number","type":"unsigned long","value":null},{"name":"mapped.ctime","type":"unsigned long","value":null}]}
{"timestamp":1700404629446916600,"threadStartTime":1700403164364553700,"processorId":2,"processId":388,"cgroupId":500,"threadId":388,"parentProcessId":1,"hostProcessId":388,"hostThreadId":388,"hostParentProcessId":1,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"systemd-journal","executable":{"path":""},"hostName":"jammy","containerId":"","container":{},"kubernetes":{},"eventId":"781","eventName":"access_remote_vm","matchedPolicies":[""],"argsNum":5,"returnValue":0,"syscall":"read","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":1669026443,"processEntityId":1669026443,"parentEntityId":1844877023,"args":[{"name":"remote_pid","type":"int","value":12862},{"name":"start_address","type":"void *","value":"0x7ffe885c2a36"},{"name":"gup_flags","type":"unsigned int","value":32768},{"name":"vm_flags","type":"unsigned long","value":1048947},{"name":"mapped.path","type":"const char*","value":"[stack]"},{"name":"mapped.device_id","type":"dev_t","value":null},{"name":"mapped.inode_number","type":"unsigned long","value":null},{"name":"mapped.ctime","type":"unsigned long","value":null}]}
{"timestamp":1700404629446944300,"threadStartTime":1700403164364553700,"processorId":2,"processId":388,"cgroupId":500,"threadId":388,"parentProcessId":1,"hostProcessId":388,"hostThreadId":388,"hostParentProcessId":1,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"systemd-journal","executable":{"path":""},"hostName":"jammy","containerId":"","container":{},"kubernetes":{},"eventId":"781","eventName":"access_remote_vm","matchedPolicies":[""],"argsNum":5,"returnValue":0,"syscall":"read","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":1669026443,"processEntityId":1669026443,"parentEntityId":1844877023,"args":[{"name":"remote_pid","type":"int","value":12862},{"name":"start_address","type":"void *","value":"0x7ffe885c2a2a"},{"name":"gup_flags","type":"unsigned int","value":32768},{"name":"vm_flags","type":"unsigned long","value":1048947},{"name":"mapped.path","type":"const char*","value":"[stack]"},{"name":"mapped.device_id","type":"dev_t","value":null},{"name":"mapped.inode_number","type":"unsigned long","value":null},{"name":"mapped.ctime","type":"unsigned long","value":null}]}
{"timestamp":1700404629453558300,"threadStartTime":1700403164364553700,"processorId":2,"processId":388,"cgroupId":500,"threadId":388,"parentProcessId":1,"hostProcessId":388,"hostThreadId":388,"hostParentProcessId":1,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"systemd-journal","executable":{"path":""},"hostName":"jammy","containerId":"","container":{},"kubernetes":{},"eventId":"781","eventName":"access_remote_vm","matchedPolicies":[""],"argsNum":5,"returnValue":0,"syscall":"read","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":1669026443,"processEntityId":1669026443,"parentEntityId":1844877023,"args":[{"name":"remote_pid","type":"int","value":12864},{"name":"start_address","type":"void *","value":"0x7ffd1f5dda85"},{"name":"gup_flags","type":"unsigned int","value":32768},{"name":"vm_flags","type":"unsigned long","value":1048947},{"name":"mapped.path","type":"const char*","value":"[stack]"},{"name":"mapped.device_id","type":"dev_t","value":null},{"name":"mapped.inode_number","type":"unsigned long","value":null},{"name":"mapped.ctime","type":"unsigned long","value":null}]}
{"timestamp":1700404629453606400,"threadStartTime":1700403164364553700,"processorId":2,"processId":388,"cgroupId":500,"threadId":388,"parentProcessId":1,"hostProcessId":388,"hostThreadId":388,"hostParentProcessId":1,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"systemd-journal","executable":{"path":""},"hostName":"jammy","containerId":"","container":{},"kubernetes":{},"eventId":"781","eventName":"access_remote_vm","matchedPolicies":[""],"argsNum":5,"returnValue":0,"syscall":"read","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":1669026443,"processEntityId":1669026443,"parentEntityId":1844877023,"args":[{"name":"remote_pid","type":"int","value":12864},{"name":"start_address","type":"void *","value":"0x7ffd1f5dda81"},{"name":"gup_flags","type":"unsigned int","value":32768},{"name":"vm_flags","type":"unsigned long","value":1048947},{"name":"mapped.path","type":"const char*","value":"[stack]"},{"name":"mapped.device_id","type":"dev_t","value":null},{"name":"mapped.inode_number","type":"unsigned long","value":null},{"name":"mapped.ctime","type":"unsigned long","value":null}]}
{"timestamp":1700404629456401700,"threadStartTime":1700403167598710300,"processorId":1,"processId":647,"cgroupId":2662,"threadId":647,"parentProcessId":1,"hostProcessId":647,"hostThreadId":647,"hostParentProcessId":1,"userId":102,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"dbus-daemon","executable":{"path":""},"hostName":"jammy","containerId":"","container":{},"kubernetes":{},"eventId":"781","eventName":"access_remote_vm","matchedPolicies":[""],"argsNum":5,"returnValue":0,"syscall":"read","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":3695380312,"processEntityId":3695380312,"parentEntityId":1844877023,"args":[{"name":"remote_pid","type":"int","value":12864},{"name":"start_address","type":"void *","value":"0x7ffd1f5dda85"},{"name":"gup_flags","type":"unsigned int","value":32768},{"name":"vm_flags","type":"unsigned long","value":1048947},{"name":"mapped.path","type":"const char*","value":"[stack]"},{"name":"mapped.device_id","type":"dev_t","value":null},{"name":"mapped.inode_number","type":"unsigned long","value":null},{"name":"mapped.ctime","type":"unsigned long","value":null}]}
```

...

rafaeldtinoco:
Another thing, are you sure you want 1 event per VMA ? You may gets lots of events from the same action:

```json
{"timestamp":1700404668293820200,"threadStartTime":1700404668293119700,"processorId":1,"processId":12905,"cgroupId":5040,"threadId":12905,"parentProcessId":12901,"hostProcessId":12905,"hostThreadId":12905,"hostParentProcessId":12901,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"tail","executable":{"path":""},"hostName":"jammy","containerId":"","container":{},"kubernetes":{},"eventId":"781","eventName":"access_remote_vm","matchedPolicies":[""],"argsNum":5,"returnValue":0,"syscall":"read","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":1550028928,"processEntityId":1550028928,"parentEntityId":2360377836,"args":[{"name":"remote_pid","type":"int","value":12901},{"name":"start_address","type":"void *","value":"0x7fff05c58000"},{"name":"gup_flags","type":"unsigned int","value":16},{"name":"vm_flags","type":"unsigned long","value":1048947},{"name":"mapped.path","type":"const char*","value":"[stack]"},{"name":"mapped.device_id","type":"dev_t","value":null},{"name":"mapped.inode_number","type":"unsigned long","value":null},{"name":"mapped.ctime","type":"unsigned long","value":null}]}
{"timestamp":1700404668293824500,"threadStartTime":1700404668293119700,"processorId":1,"processId":12905,"cgroupId":5040,"threadId":12905,"parentProcessId":12901,"hostProcessId":12905,"hostThreadId":12905,"hostParentProcessId":12901,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"tail","executable":{"path":""},"hostName":"jammy","containerId":"","container":{},"kubernetes":{},"eventId":"781","eventName":"access_remote_vm","matchedPolicies":[""],"argsNum":5,"returnValue":0,"syscall":"read","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":1550028928,"processEntityId":1550028928,"parentEntityId":2360377836,"args":[{"name":"remote_pid","type":"int","value":12901},{"name":"start_address","type":"void *","value":"0x7fff05c58fff"},{"name":"gup_flags","type":"unsigned int","value":16},{"name":"vm_flags","type":"unsigned long","value":1048947},{"name":"mapped.path","type":"const char*","value":"[stack]"},{"name":"mapped.device_id","type":"dev_t","value":null},{"name":"mapped.inode_number","type":"unsigned long","value":null},{"name":"mapped.ctime","type":"unsigned long","value":null}]}
{"timestamp":1700404668293828600,"threadStartTime":1700404668293119700,"processorId":1,"processId":12905,"cgroupId":5040,"threadId":12905,"parentProcessId":12901,"hostProcessId":12905,"hostThreadId":12905,"hostParentProcessId":12901,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"tail","executable":{"path":""},"hostName":"jammy","containerId":"","container":{},"kubernetes":{},"eventId":"781","eventName":"access_remote_vm","matchedPolicies":[""],"argsNum":5,"returnValue":0,"syscall":"read","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":1550028928,"processEntityId":1550028928,"parentEntityId":2360377836,"args":[{"name":"remote_pid","type":"int","value":12901},{"name":"start_address","type":"void *","value":"0x7fff05c59000"},{"name":"gup_flags","type":"unsigned int","value":16},{"name":"vm_flags","type":"unsigned long","value":1048947},{"name":"mapped.path","type":"const char*","value":"[stack]"},{"name":"mapped.device_id","type":"dev_t","value":null},{"name":"mapped.inode_number","type":"unsigned long","value":null},{"name":"mapped.ctime","type":"unsigned long","value":null}]}
{"timestamp":1700404668293837800,"threadStartTime":1700404668293119700,"processorId":1,"processId":12905,"cgroupId":5040,"threadId":12905,"parentProcessId":12901,"hostProcessId":12905,"hostThreadId":12905,"hostParentProcessId":12901,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"tail","executable":{"path":""},"hostName":"jammy","containerId":"","container":{},"kubernetes":{},"eventId":"781","eventName":"access_remote_vm","matchedPolicies":[""],"argsNum":5,"returnValue":0,"syscall":"read","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":1550028928,"processEntityId":1550028928,"parentEntityId":2360377836,"args":[{"name":"remote_pid","type":"int","value":12901},{"name":"start_address","type":"void *","value":"0x7fff05c59fff"},{"name":"gup_flags","type":"unsigned int","value":16},{"name":"vm_flags","type":"unsigned long","value":1048947},{"name":"mapped.path","type":"const char*","value":"[stack]"},{"name":"mapped.device_id","type":"dev_t","value":null},{"name":"mapped.inode_number","type":"unsigned long","value":null},{"name":"mapped.ctime","type":"unsigned long","value":null}]}
{"timestamp":1700404668293842200,"threadStartTime":1700404668293119700,"processorId":1,"processId":12905,"cgroupId":5040,"threadId":12905,"parentProcessId":12901,"hostProcessId":12905,"hostThreadId":12905,"hostParentProcessId":12901,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"tail","executable":{"path":""},"hostName":"jammy","containerId":"","container":{},"kubernetes":{},"eventId":"781","eventName":"access_remote_vm","matchedPolicies":[""],"argsNum":5,"returnValue":0,"syscall":"read","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":1550028928,"processEntityId":1550028928,"parentEntityId":2360377836,"args":[{"name":"remote_pid","type":"int","value":12901},{"name":"start_address","type":"void *","value":"0x7fff05c5a000"},{"name":"gup_flags","type":"unsigned int","value":16},{"name":"vm_flags","type":"unsigned long","value":1048947},{"name":"mapped.path","type":"const char*","value":"[stack]"},{"name":"mapped.device_id","type":"dev_t","value":null},{"name":"mapped.inode_number","type":"unsigned long","value":null},{"name":"mapped.ctime","type":"unsigned long","value":null}]}
{"timestamp":1700404668293846500,"threadStartTime":1700404668293119700,"processorId":1,"processId":12905,"cgroupId":5040,"threadId":12905,"parentProcessId":12901,"hostProcessId":12905,"hostThreadId":12905,"hostParentProcessId":12901,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"tail","executable":{"path":""},"hostName":"jammy","containerId":"","container":{},"kubernetes":{},"eventId":"781","eventName":"access_remote_vm","matchedPolicies":[""],"argsNum":5,"returnValue":0,"syscall":"read","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":1550028928,"processEntityId":1550028928,"parentEntityId":2360377836,"args":[{"name":"remote_pid","type":"int","value":12901},{"name":"start_address","type":"void *","value":"0x7fff05c5afff"},{"name":"gup_flags","type":"unsigned int","value":16},{"name":"vm_flags","type":"unsigned long","value":1048947},{"name":"mapped.path","type":"const char*","value":"[stack]"},{"name":"mapped.device_id","type":"dev_t","value":null},{"name":"mapped.inode_number","type":"unsigned long","value":null},{"name":"mapped.ctime","type":"unsigned long","value":null}]}
{"timestamp":1700404668293851000,"threadStartTime":1700404668293119700,"processorId":1,"processId":12905,"cgroupId":5040,"threadId":12905,"parentProcessId":12901,"hostProcessId":12905,"hostThreadId":12905,"hostParentProcessId":12901,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"tail","executable":{"path":""},"hostName":"jammy","containerId":"","container":{},"kubernetes":{},"eventId":"781","eventName":"access_remote_vm","matchedPolicies":[""],"argsNum":5,"returnValue":0,"syscall":"read","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":1550028928,"processEntityId":1550028928,"parentEntityId":2360377836,"args":[{"name":"remote_pid","type":"int","value":12901},{"name":"start_address","type":"void *","value":"0x7fff05c5b000"},{"name":"gup_flags","type":"unsigned int","value":16},{"name":"vm_flags","type":"unsigned long","value":1048947},{"name":"mapped.path","type":"const char*","value":"[stack]"},{"name":"mapped.device_id","type":"dev_t","value":null},{"name":"mapped.inode_number","type":"unsigned long","value":null},{"name":"mapped.ctime","type":"unsigned long","value":null}]}
{"timestamp":1700404668293859000,"threadStartTime":1700404668293119700,"processorId":1,"processId":12905,"cgroupId":5040,"threadId":12905,"parentProcessId":12901,"hostProcessId":12905,"hostThreadId":12905,"hostParentProcessId":12901,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"tail","executable":{"path":""},"hostName":"jammy","containerId":"","container":{},"kubernetes":{},"eventId":"781","eventName":"access_remote_vm","matchedPolicies":[""],"argsNum":5,"returnValue":0,"syscall":"read","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":1550028928,"processEntityId":1550028928,"parentEntityId":2360377836,"args":[{"name":"remote_pid","type":"int","value":12901},{"name":"start_address","type":"void *","value":"0x7fff05c5bfff"},{"name":"gup_flags","type":"unsigned int","value":16},{"name":"vm_flags","type":"unsigned long","value":1048947},{"name":"mapped.path","type":"const char*","value":"[stack]"},{"name":"mapped.device_id","type":"dev_t","value":null},{"name":"mapped.inode_number","type":"unsigned long","value":null},{"name":"mapped.ctime","type":"unsigned long","value":null}]}
{"timestamp":1700404668293863400,"threadStartTime":1700404668293119700,"processorId":1,"processId":12905,"cgroupId":5040,"threadId":12905,"parentProcessId":12901,"hostProcessId":12905,"hostThreadId":12905,"hostParentProcessId":12901,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"tail","executable":{"path":""},"hostName":"jammy","containerId":"","container":{},"kubernetes":{},"eventId":"781","eventName":"access_remote_vm","matchedPolicies":[""],"argsNum":5,"returnValue":0,"syscall":"read","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":1550028928,"processEntityId":1550028928,"parentEntityId":2360377836,"args":[{"name":"remote_pid","type":"int","value":12901},{"name":"start_address","type":"void *","value":"0x7fff05c5c000"},{"name":"gup_flags","type":"unsigned int","value":16},{"name":"vm_flags","type":"unsigned long","value":1048947},{"name":"mapped.path","type":"const char*","value":"[stack]"},{"name":"mapped.device_id","type":"dev_t","value":null},{"name":"mapped.inode_number","type":"unsigned long","value":null},{"name":"mapped.ctime","type":"unsigned long","value":null}]}
{"timestamp":1700404668293867500,"threadStartTime":1700404668293119700,"processorId":1,"processId":12905,"cgroupId":5040,"threadId":12905,"parentProcessId":12901,"hostProcessId":12905,"hostThreadId":12905,"hostParentProcessId":12901,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"tail","executable":{"path":""},"hostName":"jammy","containerId":"","container":{},"kubernetes":{},"eventId":"781","eventName":"access_remote_vm","matchedPolicies":[""],"argsNum":5,"returnValue":0,"syscall":"read","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":1550028928,"processEntityId":1550028928,"parentEntityId":2360377836,"args":[{"name":"remote_pid","type":"int","value":12901},{"name":"start_address","type":"void *","value":"0x7fff05c5cfff"},{"name":"gup_flags","type":"unsigned int","value":16},{"name":"vm_flags","type":"unsigned long","value":1048947},{"name":"mapped.path","type":"const char*","value":"[stack]"},{"name":"mapped.device_id","type":"dev_t","value":null},{"name":"mapped.inode_number","type":"unsigned long","value":null},{"name":"mapped.ctime","type":"unsigned long","value":null}]}
{"timestamp":1700404668293871900,"threadStartTime":1700404668293119700,"processorId":1,"processId":12905,"cgroupId":5040,"threadId":12905,"parentProcessId":12901,"hostProcessId":12905,"hostThreadId":12905,"hostParentProcessId":12901,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"tail","executable":{"path":""},"hostName":"jammy","containerId":"","container":{},"kubernetes":{},"eventId":"781","eventName":"access_remote_vm","matchedPolicies":[""],"argsNum":5,"returnValue":0,"syscall":"read","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":1550028928,"processEntityId":1550028928,"parentEntityId":2360377836,"args":[{"name":"remote_pid","type":"int","value":12901},{"name":"start_address","type":"void *","value":"0x7fff05c5d000"},{"name":"gup_flags","type":"unsigned int","value":16},{"name":"vm_flags","type":"unsigned long","value":1048947},{"name":"mapped.path","type":"const char*","value":"[stack]"},{"name":"mapped.device_id","type":"dev_t","value":null},{"name":"mapped.inode_number","type":"unsigned long","value":null},{"name":"mapped.ctime","type":"unsigned long","value":null}]}
{"timestamp":1700404668293877500,"threadStartTime":1700404668293119700,"processorId":1,"processId":12905,"cgroupId":5040,"threadId":12905,"parentProcessId":12901,"hostProcessId":12905,"hostThreadId":12905,"hostParentProcessId":12901,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"tail","executable":{"path":""},"hostName":"jammy","containerId":"","container":{},"kubernetes":{},"eventId":"781","eventName":"access_remote_vm","matchedPolicies":[""],"argsNum":5,"returnValue":0,"syscall":"read","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":1550028928,"processEntityId":1550028928,"parentEntityId":2360377836,"args":[{"name":"remote_pid","type":"int","value":12901},{"name":"start_address","type":"void *","value":"0x7fff05c5dfff"},{"name":"gup_flags","type":"unsigned int","value":16},{"name":"vm_flags","type":"unsigned long","value":1048947},{"name":"mapped.path","type":"const char*","value":"[stack]"},{"name":"mapped.device_id","type":"dev_t","value":null},{"name":"mapped.inode_number","type":"unsigned long","value":null},{"name":"mapped.ctime","type":"unsigned long","value":null}]}
{"timestamp":1700404668293885000,"threadStartTime":1700404668293119700,"processorId":1,"processId":12905,"cgroupId":5040,"threadId":12905,"parentProcessId":12901,"hostProcessId":12905,"hostThreadId":12905,"hostParentProcessId":12901,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"tail","executable":{"path":""},"hostName":"jammy","containerId":"","container":{},"kubernetes":{},"eventId":"781","eventName":"access_remote_vm","matchedPolicies":[""],"argsNum":5,"returnValue":0,"syscall":"read","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":1550028928,"processEntityId":1550028928,"parentEntityId":2360377836,"args":[{"name":"remote_pid","type":"int","value":12901},{"name":"start_address","type":"void *","value":"0x7fff05c5e000"},{"name":"gup_flags","type":"unsigned int","value":16},{"name":"vm_flags","type":"unsigned long","value":1048947},{"name":"mapped.path","type":"const char*","value":"[stack]"},{"name":"mapped.device_id","type":"dev_t","value":null},{"name":"mapped.inode_number","type":"unsigned long","value":null},{"name":"mapped.ctime","type":"unsigned long","value":null}]}
```

rafaeldtinoco:
> About `mmap`, I don't think its a problem as the proc mem file doesn't implement a `mmap` file operation.

True.

> Lastly, the `aio_read` and `io_uring` will need to both pass through the file operation. So I don't have any concern here either.

Indeed asyncio works good (so io_uring):

```json
{"timestamp":1700405842419459000,"threadStartTime":1700405842419344000,"processorId":1,"processId":13141,"cgroupId":5289,"threadId":13142,"parentProcessId":13129,"hostProcessId":13141,"hostThreadId":13142,"hostParentProcessId":13129,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"asyncaio","executable":{"path":""},"hostName":"jammy","containerId":"","container":{},"kubernetes":{},"eventId":"781","eventName":"access_remote_vm","matchedPolicies":[""],"argsNum":5,"returnValue":0,"syscall":"pread64","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":3626678476,"processEntityId":1852968592,"parentEntityId":2070396318,"args":[{"name":"remote_pid","type":"int","value":13129},{"name":"start_address","type":"void *","value":"0x7ffd739c7000"},{"name":"gup_flags","type":"unsigned int","value":16},{"name":"vm_flags","type":"unsigned long","value":1048947},{"name":"mapped.path","type":"const char*","value":"[stack]"},{"name":"mapped.device_id","type":"dev_t","value":null},{"name":"mapped.inode_number","type":"unsigned long","value":null},{"name":"mapped.ctime","type":"unsigned long","value":null}]}
{"timestamp":1700405842419771400,"threadStartTime":1700405842419344000,"processorId":1,"processId":13141,"cgroupId":5289,"threadId":13142,"parentProcessId":13129,"hostProcessId":13141,"hostThreadId":13142,"hostParentProcessId":13129,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"asyncaio","executable":{"path":""},"hostName":"jammy","containerId":"","container":{},"kubernetes":{},"eventId":"781","eventName":"access_remote_vm","matchedPolicies":[""],"argsNum":5,"returnValue":0,"syscall":"pread64","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":3626678476,"processEntityId":1852968592,"parentEntityId":2070396318,"args":[{"name":"remote_pid","type":"int","value":13129},{"name":"start_address","type":"void *","value":"0x7ffd739c7400"},{"name":"gup_flags","type":"unsigned int","value":16},{"name":"vm_flags","type":"unsigned long","value":1048947},{"name":"mapped.path","type":"const char*","value":"[stack]"},{"name":"mapped.device_id","type":"dev_t","value":null},{"name":"mapped.inode_number","type":"unsigned long","value":null},{"name":"mapped.ctime","type":"unsigned long","value":null}]}
{"timestamp":1700405842420135700,"threadStartTime":1700405842419344000,"processorId":1,"processId":13141,"cgroupId":5289,"threadId":13142,"parentProcessId":13129,"hostProcessId":13141,"hostThreadId":13142,"hostParentProcessId":13129,"userId":0,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"asyncaio","executable":{"path":""},"hostName":"jammy","containerId":"","container":{},"kubernetes":{},"eventId":"781","eventName":"access_remote_vm","matchedPolicies":[""],"argsNum":5,"returnValue":0,"syscall":"pread64","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":3626678476,"processEntityId":1852968592,"parentEntityId":2070396318,"args":[{"name":"remote_pid","type":"int","value":13129},{"name":"start_address","type":"void *","value":"0x7ffd739c7800"},{"name":"gup_flags","type":"unsigned int","value":16},{"name":"vm_flags","type":"unsigned long","value":1048947},{"name":"mapped.path","type":"const char*","value":"[stack]"},{"name":"mapped.device_id","type":"dev_t","value":null},{"name":"mapped.inode_number","type":"unsigned long","value":null},{"name":"mapped.ctime","type":"unsigned long","value":null}]}
```

rafaeldtinoco:
> I haven't seen a single test failing till now, so I am not sure to what you are referring. I can't really understand the problem without seeing it, so if you can reset the tests for me to see or send me a test suite that failed I will dig deeper.

Fix your pr.yaml file and force push. Wait for tests to finish and you will see. I dont need to reset anything on my side. Your pr.yaml is wrong.

rafaeldtinoco:
> However, I don't see why should we have a sleep in the test like in the process tree. There is no state here to initialize. If there is a timing issue here, it is an issue with all e2e tests and should be fixed by the tests environment. However, I can add for now a sleep if you think it will solve it.

Never said it was a definitive answer. My feedback was "it might be related", but it might be an entire different thing. When you fix pr.yaml and tests fail you will be able to see. Let me know if you need a VM with the same images that are failing (but it might be intermittent among all of them, without considering v6.5 which always fails due to other reasoning).

rafaeldtinoco:
For the record:

https://gist.github.com/rafaeldtinoco/bb0bcf795fb2c821136db01417ca669a
https://gist.github.com/rafaeldtinoco/7042acfa78d3c7ca3f31ce8147d54255

Those files show the execution path in a 5.19 (event works) vs a 6.5 (event does not work).

```
rafaeldtinoco@rugged:mine$ gh gist create asyncio.perf.5.19.0-50-generic
- Creating gist asyncio.perf.5.19.0-50-generic
✓ Created secret gist asyncio.perf.5.19.0-50-generic
https://gist.github.com/rafaeldtinoco/bb0bcf795fb2c821136db01417ca669a

rafaeldtinoco@rugged:mine$ gh gist create asyncio.perf.6.5.9-arch2-1
- Creating gist asyncio.perf.6.5.9-arch2-1
✓ Created secret gist asyncio.perf.6.5.9-arch2-1
https://gist.github.com/rafaeldtinoco/7042acfa78d3c7ca3f31ce8147d54255
```

Im using https://gist.github.com/rafaeldtinoco/1e619d9d25a4a934b0ef577304125b87 to test.

AlonZivony:
> Another thing, are you sure you want 1 event per VMA ? You may gets lots of events from the same action:

You are right that it might spam, but it is significant.
Each vma has its own name and own permissions, and you don't see operations on the file done frequently or in massive amount.
It might be a cause of malicious spam, but it is true for almost any event.

rafaeldtinoco:
Looks like the tests failed but when looking at the output the event (at least on 4.18) worked. Needs some fine tuning I believe.

AlonZivony:
You can see that the problem is the version check, because the `remote_pid` value is 0.
I will try to find another type.

geyslan:

yanivagman:
Not going to be implemented

PR #3554, opened 2023-10-05
fix: cap_capable should parse arguments

yanivagman:

When choosing output option parse-arguments, cap_capable event didn't show the capability string.
This PR fixes it.


rafaeldtinoco (approved):
LGTM, but it is weird why it was like that :\

yanivagman:
> LGTM, but it is weird why it was like that :\

Yes, probably some code refactor missed that. It doesn't make sense that it was under the case of sysEnter/sysExit

PR #3557, opened 2023-10-06
fix: add api root dir

josedonizetti:
Move api to rootdir for package stability. This needs to bem merged before I clean up the old code, and replace the imports, or else `go` doesn't have a module to download.


api/v1beta1/event.proto
```diff
@@ -0,0 +1,133 @@
...
+}
+
+message Parent  {
+    optional Executable executable = 1;
+    google.protobuf.UInt32Value entity_id = 2;
+    google.protobuf.UInt32Value pid = 3;
+    optional google.protobuf.UInt32Value namespaced_pid = 4;
+    optional User real_user = 5;
+    optional Thread thread = 6;
+    optional Process parent = 7;
```
yanivagman, line 48: Shouldn't this be of type Parent?
Or actually, why do we even need message Parent - can't we just use Process for Parent as well?

api/v1beta1/event.proto
```diff
@@ -0,0 +1,123 @@
...
+    optional K8s k8s = 3;
+}
+
+message Process  {
+    Executable executable = 1;
+    google.protobuf.UInt32Value entity_id = 2;
+    google.protobuf.UInt32Value pid = 3;
+    google.protobuf.UInt32Value namespaced_pid = 4;
+    User real_user = 5;
+    Thread thread = 6;
```
yanivagman, line 37: Since Process is now also used for Parent, we should add optional for:
Executable
User
Thread

Makefile
```diff
@@ -329,7 +329,7 @@ $(OUTPUT_DIR)/libbpf/libbpf.a: \
 		DESTDIR=$(abspath ./$(OUTPUT_DIR)/libbpf/) \
 		OBJDIR=$(abspath ./$(OUTPUT_DIR)/libbpf/obj) \
 		INCLUDEDIR= LIBDIR= UAPIDIR= prefix= libdir= \
-		install install_uapi_headers
+		install install_ugit _headers
```
geyslan, line 332: unseen typing?

yanivagman (approved):
LGTM

PR #3558, opened 2023-10-06
Use api module

josedonizetti:
This PR finalize the https://github.com/aquasecurity/tracee/issues/3208, it updates the code to use the new API module: https://github.com/aquasecurity/tracee/tree/main/api


yanivagman:
Commit name should be changed (now its parse arguments related)

yanivagman (approved):
LGTM

PR #3559, opened 2023-10-06
proctree + datasource fixes and adjustments

rafaeldtinoco:
commit 559a649f8 (HEAD -> proctree-datasource-fixes, rafaeldtinoco/proctree-datasource-fixes)
Date:   Fri Oct 6 10:31:41 2023

    fix(proctree): fix issues when updating from procfs
    
    - raise channel buffer due to initialization needs (all pids)
    - check if fields from stat and status are valid
    - take care of PIDs 0 and 1 in the right manner
    - only update nodes if they are incomplete in basic fields

commit 72d31d570
Date:   Fri Oct 6 10:28:43 2023

    fix(proctree): remove unused capabilities change
    
    The routing that is in charge of reading procfs is not the same as the
    one being called for initialization. Capabilities were being raised for
    no reason in the initialization.
    
    The loop in charge for reading PIDs from a channel and reading procfs,
    in order to update the process tree, can (or not) have the capabilities
    elevated in a later moment (if/when needed, for example, when
    /proc/pid/exe needs to be read).

commit b407883a4
Date:   Fri Oct 6 10:28:08 2023

    chore(proctree): keep important fields in the table all the time

commit 2e274ee4e
Date:   Fri Oct 6 10:23:48 2023

    fix(proctree): fix proctree feed in many places
    
    - only feed parent if it isn't found or is incomplete
    - only feed leader if it isn't found or is incomplete
    - always create process and thread nodes on exit (due to out-of-order)
    - uid and gid fields should be -1 for no changes

commit 3dc7c9187
Date:   Fri Oct 6 10:22:12 2023

    fix(datasource): thread list is indexed by tid and not pid

commit 628d4269d
Date:   Tue Oct 3 15:23:16 2023

    chore(proctree): documentation and style changes

I'm not adding the test yet because it is not needed for the release. I have to adjust the test in order to add it as a github actions test (it works locally and I used it to provide these fixes, but needs to finish the automation).

### DIFF
--- a/pkg/proctree/datasource.go
+++ b/pkg/proctree/datasource.go
@@ -118,7 +118,7 @@ func (ptds *DataSource) exportProcessInfo(
 		}
 		threadInfo := thread.GetInfo()
 		if threadInfo.IsAliveAt(queryTime) {
-			aliveThreads[threadInfo.GetPid()] = threadHash
+			aliveThreads[threadInfo.GetTid()] = threadHash
 		}
 	}
 
--- a/pkg/proctree/datasource.go
+++ b/pkg/proctree/datasource.go
@@ -9,7 +9,6 @@ import (
 )
 
 // DataSource is an implementation to detect.Datasource interface, enveloping the ProcessTree type.
-// It exposes all the relevant information of the tree using the interface methods.
 type DataSource struct {
 	procTree *ProcessTree
 }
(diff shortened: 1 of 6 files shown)


pkg/proctree/datasource.go
```diff
@@ -118,7 +118,7 @@ func (ptds *DataSource) exportProcessInfo(
 		}
 		threadInfo := thread.GetInfo()
 		if threadInfo.IsAliveAt(queryTime) {
-			aliveThreads[threadInfo.GetPid()] = threadHash
+			aliveThreads[threadInfo.GetTid()] = threadHash
```
yanivagman, line 121: I don't understand. Threads are of a process, and nit of another thread, no? So should we get the threads of some pid?

pkg/proctree/proctree_feed.go
```diff
@@ -46,34 +46,43 @@ func (pt *ProcessTree) FeedFromFork(feed ForkFeed) error {
...
 				Tid:         int(feed.ParentTid),
 				Pid:         int(feed.ParentPid),
 				NsTid:       int(feed.ParentNsTid),
 				NsPid:       int(feed.ParentNsPid),
 				StartTimeNS: feed.ParentStartTime,
-				PPid:        -1, // do not change the parent PID
-				NsPPid:      -1, // do not change the ns parent PID
-				Uid:         -1, // do not change the parent UID
-				Gid:         -1, // do not change the parent GID
+				PPid:        0,  // do not change the parent ppid
```
yanivagman, line 58: 1. There exists pud number 0 (the sched task of the kernel), so why not using -1?
2. Since you have "parent" in the comment, should be parent pid and not parent ppid. Same for nsppid below

pkg/proctree/datasource.go
rafaeldtinoco: This is a temporary map for the process data source query. Each query (for a single process) also provides a map of existing children (processes) and threads (lwps).

The fix is about indexing by thread PID. It was creating a single entry for multiple threads and replacing that entry over and over. That happened because all threads had the same PID but not the same TID.

pkg/proctree/proctree_feed.go
rafaeldtinoco: 1) yes, I believe all the values could be -1.
2) No, its parent ppid in this case. I'm not updating the parent ppid value.

yanivagman (approved):
LGTM

PR #3571, opened 2023-10-13
fix(event): make sure parent points to a real process

rafaeldtinoco:
Both, process tree AND the event context should have "parent" pointing to a real task and not a potential thread. Event context still had it pointing to the real parent and not what is being called "up parent".

### DIFF
--- a/pkg/ebpf/c/common/context.h
+++ b/pkg/ebpf/c/common/context.h
@@ -3,6 +3,7 @@
 
 #include <vmlinux.h>
 
+#include <common/logging.h>
 #include <common/task.h>
 #include <common/cgroups.h>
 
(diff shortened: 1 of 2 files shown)


rafaeldtinoco:
I'll run E2E tests before merging this one, just to make sure.

rafaeldtinoco:
It passed the E2E. I need to release v0.18.1 today so I might merge this and address any concerns in a later patch if needed, hopefully wont be needed.

geyslan (approved):
LGTM

pkg/ebpf/c/common/context.h
```diff
@@ -22,49 +23,61 @@ init_context(void *ctx, event_context_t *context, struct task_struct *task, u32
...
+    // Task Info on Host
     context->task.host_tid = id;
     context->task.host_pid = id >> 32;
-    context->task.host_ppid = get_task_ppid(task);
+    context->task.host_ppid = get_task_pid(up_parent); // always a real process (not a lwp)
+    // Namespaces Info
     context->task.tid = get_task_ns_pid(task);
     context->task.pid = get_task_ns_tgid(task);
-    context->task.ppid = get_task_ns_ppid(task);
+    context->task.ppid = get_task_ns_pid(up_parent); // ns pid for up_parent
```
AlonZivony, line 38: This is not necessarily the true value here.
PID 1 of a PID namespace have no parent in its namespace.
This results that we get the PID of the parent under its own namespace.
Unless we want this behavior, I would have expected to get PPID=0 in this case.
WDYT?

Issue #3570, opened 2023-10-13
Create man option for flags


geyslan:
https://github.com/aquasecurity/tracee/pull/3321 removed help for flags and introduced the respective markdowns (not indexed in the documentation) to be used to create man page files.

- [x] Create `man` option, e.g.: `tracee man scope`.
- [x] Update docs https://aquasecurity.github.io/tracee/v0.18/docs/filters/filtering/

PR #3572, opened 2023-10-13
Introduce cmd man option for flags

geyslan:
Close: #3570

d1603559c **chore(docs): index flags man pages**
d92517eb2 **chore(docs): update flag help example**
e36b15fab **chore(cmd/man): add man cmd and subcmds**

d1603559c **chore(docs): index flags man pages**

```
This makes the same base flag markdown helps used for man page
generation to be available in the online documentation.
```

e36b15fab **chore(cmd/man): add man cmd and subcmds**

```
man cmd open page for a specified flag name (sub command).
```

### 2. Explain how to test it

```
❯ ./dist/tracee man                                
Print man page for a specified flag name

Usage:
  tracee man
  tracee man [command]

Aliases:
  man, m

Available Commands:
  cache        cache flag help
  capabilities capabilities flag help
  capture      capture flag help
  config       config flag help
  crs          containers flag help
  events       events flag help
  log          log flag help
  output       output flag help
  rego         rego flag help
  scope        scope flag help

Flags:
  -h, --help   help for man

Use "tracee man [command] --help" for more information about a command.
```

`./dist/tracee man cache`

```
TRACEE-CACHE(1)                         Tracee Cache Flag Manual                        TRACEE-CACHE(1)

   NAME
       tracee --cache - Select different cache types for the event pipeline queueing

   SYNOPSIS
       tracee --cache [none|cache-type=<type>] [--cache mem-cache-size=<size>]

   DESCRIPTION
       The --cache flag allows you to select different cache types for the event pipeline queueing.

       Possible options for cache-type are:

       • none: No event caching in the pipeline (default) - similar to --cache none.

       • mem: Enables caching events in memory.

       If  cache-type=mem is chosen, you can also set the memory cache size in megabytes (MB) using the
       mem-cache-size option.  This option only works when cache-type=mem.

   EXAMPLES
       • To cache events in memory using the default values, use the following flag:

                --cache cache-type=mem

       • To cache events in memory and set the memory cache size to 1024 MB, use the following flag:

                --cache cache-type=mem --cache mem-cache-size=1024

       • To disable event caching in the pipeline, use the following flag:

                --cache none

                                                2023/10                                 TRACEE-CACHE(1)
```

image

image


docs/docs/filters/filtering.md
```diff
@@ -1,9 +1,9 @@
 # Tracing Event Filtering
 
 ```console
-sudo ./dist/tracee --scope help
-sudo ./dist/tracee --events help
-sudo ./dist/tracee --scope xxx --events xxx
+./dist/tracee man --help
```
rafaeldtinoco, line 4: This does not make any sense to me. I have never seen any userland tool that had "man" together with "--help". Also, having a "man" command without dashes, and a subcommand with dashes is very weird.

Also, man without "--help" has the same output, so why not to just leave "tracee man" here ?

rafaeldtinoco (approved):

rafaeldtinoco:
LGTM.

Before merging:

1. I would address what I pointed as a comment.
2. Im not entirely sure about the path for the man pages directory. (docs/docs/flags/man is weird).

Also, IMO, the whole point of the suggestion of having the documentation done at a single place, and that this documentation would generate the man pages for the CLI, was that we could also have the documentation available (mkdocs).

I would open an issue mentioning that we have to publish this markdown files inside the official documentation (so we can always update a single place). Also, every time one of the "source files" is modified, we should warn that a man page generation is needed (and block PRs).

docs/docs/filters/filtering.md
geyslan: `--help/-h` flags are automatic for any command in cobra. I changed the man command to print the same as the -h flag if it doesn't contain any other arg. So, `tracee man` = `tracee man -h`: https://github.com/aquasecurity/tracee/pull/3572/files#diff-35ddad12850bedf904c52eb37d70d82187698a3531c352b2dcd62e6e43447cfaR37-R41

> Also, having a "man" command without dashes, and a subcommand with dashes is very weird.

It is the main logic of the cobra and GNU flags, if it has dashes it is a flag, not a command/subcommand.

> Also, man without "--help" has the same output, so why not to just leave "tracee man" here ?

I'll change the doc, removing the --help flag.

geyslan:
@rafaeldtinoco, I think I addressed all feedback and got it in the online doc as well.

rafaeldtinoco:
@geyslan feel free to merge! nice changes!

PR #3577, opened 2023-10-16
fix(events): make ns parent pid exist only in same ns

rafaeldtinoco:
Fix for a situation observed by Alon at:

https://github.com/aquasecurity/tracee/pull/3571#discussion_r1361217873


AlonZivony (approved):
LGTM

Issue #3440, opened 2023-09-05
Integration tests timeout


geyslan:
## Description

Depending on the host, integration tests sometimes block on the specific test `Test_EventFilters/pid:_trace_events_from_pid_1`. Despite being occurring during this test case, it looks more like there might be some sync issue or a deadlock scenario.

```
GOOS=linux CC=clang GOARCH=amd64 CGO_CFLAGS="-I/vagrant/dist/libbpf" CGO_LDFLAGS="-lelf  -lz  /vagrant/dist/libbpf/libbpf.a" \
go build -o ./dist/syscaller ./tests/integration/syscaller/cmd
=== RUN   Test_EventFilters
=== RUN   Test_EventFilters/container:_event:_trace_only_events_from_new_containers
    tracee.go:97: started tracee...
    tracee.go:197: stopped tracee
=== RUN   Test_EventFilters/mntns/pidns:_trace_events_only_from_mount/pid_namespace_0
    tracee.go:97: started tracee...
    tracee.go:197: stopped tracee
=== RUN   Test_EventFilters/mntns:_trace_events_from_all_mount_namespaces_but_current
    tracee.go:97: started tracee...
    tracee.go:197: stopped tracee
=== RUN   Test_EventFilters/pidns:_trace_events_from_all_pid_namespaces_but_current
    tracee.go:97: started tracee...
    tracee.go:197: stopped tracee
=== RUN   Test_EventFilters/comm:_mntns:_pidns:_event:_trace_events_set_in_a_single_policy_from_current_pid/mount_namespaces
    tracee.go:97: started tracee...
    tracee.go:197: stopped tracee
=== RUN   Test_EventFilters/comm:_event:_trace_events_set_in_a_single_policy_from_ping_command
    tracee.go:97: started tracee...
    tracee.go:197: stopped tracee
=== RUN   Test_EventFilters/comm:_event:_trace_events_set_in_a_single_policy_from_ping_command#01
    tracee.go:97: started tracee...
    tracee.go:197: stopped tracee
=== RUN   Test_EventFilters/event:_args:_trace_event_set_in_a_specific_policy_with_args_pathname_finishing_with_'ls'
    tracee.go:97: started tracee...
    tracee.go:197: stopped tracee
=== RUN   Test_EventFilters/event:_args:_trace_event_set_in_a_specific_policy_with_args_pathname_starting_with_*_wildcard
    tracee.go:97: started tracee...
    tracee.go:197: stopped tracee
[...]

PR #3579, opened 2023-10-17
chore(tests): re-enable 'trace events from pid 1'

geyslan:
Close: #3440 

538e6be62 **chore(tests): re-enable 'trace events from pid 1'**

```
This re-enables 'pid 1 trace events' since it was unable to be
reproduced. If it happens again, hopefully with the use of a shuffle
seed it can be reproduced.

During analysis, another issue was detected with Test_changeOwnComm. The
raw fork is not safe at all, even with all the precautions taken, which
is why it was removed and the test takes place in the same thread now.
Additional measures were taken to prevent the goroutine from escaping
the OS thread: calling runtime.GOMAXPROCS(1) and changeOwnComm()
from init().

This also removes leftover code (eventOutput, waitForTraceeOutput).
```

### DIFF
--- a/tests/integration/event_filters_test.go
+++ b/tests/integration/event_filters_test.go
@@ -566,39 +566,39 @@ func Test_EventFilters(t *testing.T) {
 			useSyscaller: false,
 			test:         ExpectAnyOfEach,
 		},
-		// {
-		// 	name: "pid: trace events from pid 1",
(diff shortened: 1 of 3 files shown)


geyslan:
I ran it more than 10 times in a row without any failures (local and github workflow).

rafaeldtinoco (approved):
LGTM

PR #3583, opened 2023-10-18
chore(workflow): enable single image for multiple arches

rafaeldtinoco:


.github/workflows/release.yaml
```diff
@@ -103,20 +82,54 @@ jobs:
...
-          docker image tag tracee:latest aquasec/tracee:${ARCH}
           docker image tag tracee:latest aquasec/tracee:${ARCH}-${TAG}
-          docker image push aquasec/tracee:${ARCH}
           docker image push aquasec/tracee:${ARCH}-${TAG}
+        shell: bash
       - name: Sign Docker image
         run: |
           TAG=$(echo ${{ github.event.inputs.ref }} | sed -e "s/v//gI")
           ARCH=$(uname -m)
-          cosign sign -y $(docker inspect --format='{{index .RepoDigests 0}}' aquasec/tracee:${ARCH})
```
yanivagman, line 121: Do we have the signing step for the new image?

rafaeldtinoco: It is not an image, its a manifest. We can sign the manifest as well in a near-future moment, but its 2 different things (the images from the manifest are currently signed).

Issue #3584, opened 2023-10-18
Cloud Foundry container runtime support


hagitaquasec:
## Description

1. deploy tracee in TAS environment and use tracee to detect "container_create" events
2. deploy cloud-foundry application using the "cf" cli. 
3. expected to receive "container_create" event for the new created container. But the event is missing.

## Output of `tracee version`:
```
v0.11.1
```

## Output of `uname -a`:

```
Linux 2c3cb5b4-c159-485f-9b85-138cd8011e42 5.15.0-56-generic #62-Ubuntu SMP Tue Nov 22 19:54:14 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

```

## Additional details 
The reason why the event is missing, is because tracee does not detect the events as container events. The "container_id" field is empty.
"container_id" is empty because the regex of TAS containers has a different format than crio/containerd formats. e.g:
8a37cb6f-385b-4dd3-732c-34d6
afbd0a34-4372-4c89-64ff-08c5

rafaeldtinoco:
Context about supporting this:

https://github.com/aquasecurity/tracee/pull/3586#discussion_r1364804067

Issue #3003, opened 2023-04-18
Unknown container runtime on some kubernetes clusters


NDStrahilevitz:
## Description

1. Create a containerd gke cluster
2. Run tracee with `-f e=cgroup_mkdir,container_create`
3. Observe that cgroup paths are of the form `kubepods/<besteffort|burstable>/podXXXX/<container_id>`
4. Observe that container_create events have the `runtime` argument set to `unknown`

## Output of `tracee -v`:

Tracee version "v0.13.1"

## Output of `uname -a`:

Linux gke-enforcer-overhea-prometheus-node--49423563-kggk 5.15.0-1027-gke #32-Ubuntu SMP Tue Jan 24 11:53:18 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

NDStrahilevitz:
Waiting for the process tree to be added.
Then I can query kubelet from the process tree and resolve the kubernetes default runtime from there.

geyslan:
Well, things seem to have changed in GKE since this issue was opened. Following are analyzes in two environments (cos_containerd and ubuntu_containerd).

### cos_containerd

#### env

`uname -a`

```
Linux geyslan@gke-cluster-1-default-pool-29cfd713-gdzs 5.15.109+ #1 SMP Fri Jun 9 10:57:30 UTC 2023 x86_64 Intel(R) Xeon(R) CPU @ 2.20GHz GenuineIntel GNU/Linux
```

`tr \\0 ' ' < /proc/"$(pgrep kubelet)"/cmdline`

```
[...]

PR #3586, opened 2023-10-18
fix: fail to parse container-id of TAS (cloud-founry) garden applicat…

hagitaquasec:
…ions

The regex check does not match.
It should match the following patterns:

ca94fb1d-8d1e-42f9-529c-a8af
e660266c-de34-45e9-6102-d881

fix: #3584

### DIFF
--- a/pkg/containers/containers.go
+++ b/pkg/containers/containers.go
@@ -241,7 +241,8 @@ func (c *Containers) EnrichCgroupInfo(cgroupId uint64) (cruntime.ContainerMetada
 }
 
 var (
-	containerIdFromCgroupRegex = regexp.MustCompile(`^[A-Fa-f0-9]{64}$`)
+	containerIdFromCgroupRegex       = regexp.MustCompile(`^[A-Fa-f0-9]{64}$`)
+	gardenContainerIdFromCgroupRegex = regexp.MustCompile(`^[0-9a-fA-F]{8}(-[0-9a-fA-F]{4}){4}$`)
 )
 
 // getContainerIdFromCgroup extracts container id and its runtime from path.
@@ -252,8 +253,8 @@ func getContainerIdFromCgroup(cgroupPath string) (string, cruntime.RuntimeId, bo
 	// search from the end to get the most inner container id
 	for i := len(cgroupParts) - 1; i >= 0; i = i - 1 {
 		pc := cgroupParts[i]
-		// container id is at least 64 characters long
-		if len(pc) < 64 {
+		// container id is at least 28 characters long
+		if len(pc) < 28 {
 			continue
 		}
 
@@ -292,6 +293,10 @@ func getContainerIdFromCgroup(cgroupPath string) (string, cruntime.RuntimeId, bo
 			// return first match: closest to root dir path component
 			// (to have container id of the outer container)
 			// container root determined by being matched on the last path part
+			return id, runtime, i == len(cgroupParts)-1
+		}
+		if matched := gardenContainerIdFromCgroupRegex.MatchString(id); matched {
+			runtime = cruntime.Garden
 			return id, runtime, i == len(cgroupParts)-1
 		}
 	}
--- a/pkg/containers/runtime/runtime.go
+++ b/pkg/containers/runtime/runtime.go
@@ -43,6 +43,7 @@ const (
 	Containerd
 	Crio
 	Podman
+	Garden
 )
 
 var runtimeStringMap = map[RuntimeId]string{
@@ -51,6 +52,7 @@ var runtimeStringMap = map[RuntimeId]string{
 	Containerd: "containerd",
 	Crio:       "crio",
 	Podman:     "podman",
+	Garden:     "garden",
 }
 
 func (runtime RuntimeId) String() string {
@@ -69,6 +71,9 @@ func FromString(str string) RuntimeId {
 		return Podman
 	case "containerd":
 		return Containerd
+	case "garden":
+		return Garden
+
 	default:
 		return Unknown
 	}


pkg/containers/containers.go
```diff
@@ -241,7 +241,8 @@ func (c *Containers) EnrichCgroupInfo(cgroupId uint64) (cruntime.ContainerMetada
 }
 
 var (
-	containerIdFromCgroupRegex = regexp.MustCompile(`^[A-Fa-f0-9]{64}$`)
+	containerIdFromCgroupRegex       = regexp.MustCompile(`^[A-Fa-f0-9]{64}$`)
+	gardenContainerIdFromCgroupRegex = regexp.MustCompile(`^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}$`)
```
rafaeldtinoco, line 245: This could be replaced by:

```go
gardenContainerIdFromCgroupRegex = regexp.MustCompile(`^[0-9a-fA-F]{8}(-[0-9a-fA-F]{4}){4}$`)
```

so regex could be less prolix.

pkg/containers/containers.go
```diff
@@ -294,6 +295,10 @@ func getContainerIdFromCgroup(cgroupPath string) (string, cruntime.RuntimeId, bo
 			// container root determined by being matched on the last path part
 			return id, runtime, i == len(cgroupParts)-1
 		}
+		if matched := gardenContainerIdFromCgroupRegex.MatchString(id); matched {
+			runtime = cruntime.Garden
```
rafaeldtinoco, line 299: Yep, but there is a problem here, returning a runtime isn't enough for the entire "runtime" logic for tracee. If you look at `pkg/containers/runtime/sockets.go`:

```go
// check default paths for all supported container runtimes and aggregate them
func Autodiscover(onRegisterFail func(err error, runtime RuntimeId, socket string)) Sockets {
	register := func(sockets *Sockets, runtime RuntimeId, socket string) {
		err := sockets.Register(runtime, socket)
		if err != nil {
			onRegisterFail(err, runtime, socket)
		}
	}
	sockets := Sockets{}
	const (
		defaultContainerd = "/var/run/containerd/containerd.sock"
		defaultDocker     = "/var/run/docker.sock"
		defaultCrio       = "/var/run/crio/crio.sock"
		defaultPodman     = "/var/run/podman/podman.sock"
	)

	register(&sockets, Containerd, defaultContainerd)
	register(&sockets, Docker, defaultDocker)
	register(&sockets, Crio, defaultCrio)
	register(&sockets, Podman, defaultPodman)

	return sockets
}
```

we would have to add a socket for the default socket created by garden. (`/var/vcap/data/garden/run/garden.sock` maybe ?). And then we would register this socket.

2nd step would be to create a `ContainerEnricher`. You will find the following ones:

1. pkg/containers/runtime/containerd.go
2. pkg/containers/runtime/crio.go
3. pkg/containers/runtime/docker.go

They all implement the logic to connect to the runtimes and query information about the running containers. This is so tracee can enrich the events with information about the running containers.

If you pick "docker", for example, at `pkg/containers/runtime/docker.go`, it talks to docker socket and queries information about the container given a container ID:

```go
	metadata := ContainerMetadata{
		ContainerId: containerId,
	}
	resp, err := e.client.ContainerInspect(ctx, containerId)
	if err != nil {
		return metadata, errfmt.WrapError(err)
	}
	container := (*resp.ContainerJSONBase)
	metadata.Name = container.Name

	// Docker prefixes a '/' token to local containers.
	// This can cause some confusion so we remove it if relevant.
	if strings.HasPrefix(metadata.Name, "/") {
		metadata.Name = container.Name[1:]
	}

	// get initial image name from docker's container config
	if resp.Config != nil {
		metadata.Image = resp.Config.Image

		// if in k8s extract pod data from the labels
		if resp.Config.Labels != nil {
			labels := resp.Config.Labels
			metadata.Pod = PodMetadata{
				Name:      labels[PodNameLabel],
				Namespace: labels[PodNamespaceLabel],
				UID:       labels[PodUIDLabel],
				Sandbox:   e.isSandbox(labels),
			}
		}
	}
...
```

rafaeldtinoco (commented):
So, like said in the comment previously, we need not only the parsing logic for existing containers but also the enriching logic from the runtime (garden-runc in this case).

When you run tracee with:

```console
$ sudo ./dist/tracee --log debug --containers --proctree source=both --output option:sort-events --output none --output option:parse-arguments --events uname 2>&1 | grep runtime
{"L":"DEBUG","T":"2023-10-18T23:40:11.603-0300","M":"RuntimeSockets: failed to register default","socket":"crio","error":"runtime.(*Sockets).Register: failed to register runtime socket stat /var/run/crio/crio.sock: no such file or directory","origin":"flags:pkg/cmd/flags/containers.go:47","calls":"PrepareContainers.func1() < Autodiscover.func1() < Autodiscover() < PrepareContainers() < GetTraceeRunner() < glob..func4() < (*Command).execute() < (*Command).ExecuteC() < (*Command).Execute() < Execute() < main()"}
{"L":"DEBUG","T":"2023-10-18T23:40:11.603-0300","M":"RuntimeSockets: failed to register default","socket":"podman","error":"runtime.(*Sockets).Register: failed to register runtime socket stat /var/run/podman/podman.sock: no such file or directory","origin":"flags:pkg/cmd/flags/containers.go:47","calls":"PrepareContainers.func1() < Autodiscover.func1() < Autodiscover() < PrepareContainers() < GetTraceeRunner() < glob..func4() < (*Command).execute() < (*Command).ExecuteC() < (*Command).Execute() < Execute() < main()"}
{"L":"DEBUG","T":"2023-10-18T23:40:11.622-0300","M":"Enricher","error":"containers.(*runtimeInfoService).Register: error registering enricher: unsupported runtime crio","origin":"containers:pkg/containers/containers.go:73","calls":"New() < (*Tracee).Init() < Runner.Run() < glob..func4() < (*Command).execute() < (*Command).ExecuteC() < (*Command).Execute() < Execute() < main()"}
{"L":"DEBUG","T":"2023-10-18T23:40:11.622-0300","M":"Enricher","error":"containers.(*runtimeInfoService).Register: error registering enricher: unsupported runtime podman","origin":"containers:pkg/containers/containers.go:83","calls":"New() < (*Tracee).Init() < Runner.Run() < glob..func4() < (*Command).execute() < (*Command).ExecuteC() < (*Command).Execute() < Execute() < main()"}
```

you will see that tracee tries to register all "known" runtime sockets (the default places where they can be found) during its initialization. For garden, it would also try to find garden socket at the default place in order to communicate to the runtime and enrich the events as they should be enriched.

pkg/containers/containers.go
rafaeldtinoco: We would have to have a runtime implementation to talk to cloud foundry API (there is a golang API so shouldn't be difficult). Then, through this API we would get information about the container.

Then, like you saw above (for docker), we have to check if cloud foundry sets labels that can be read for us to populate the "k8s" enrichment (Name, Namespace, UID, ...).

This is what will enrich tracee events with information from a specific container! I hope that helps, and, we probably need an escalation for this feature to be done if needed by CNDR (unless someone else implements it, then I can review for sure).

rafaeldtinoco: This is what will enrich tracee events with information from a specific container! 

Note that, from my quick study about cloud foundry, looks like there might be many combinations to check/support, right ? `garden-k8s`, `garden-runc` (and probably not the `garden-linux` and `guardian` old container managers).

I hope that helps, and, we probably need an escalation for this feature to be done if needed by CNDR (unless someone else implements it, then I can review for sure). Feel free to ping me!

rafaeldtinoco: @itaysk FYI.

rafaeldtinoco: Spoke to @hagitaquasec offline and we are gonna flag this as a partial change for garden-runc. This regex change will provide the "container_id" event context only. Any enrichment for garden (or cloud-foundry) would have to be done in another PR/change.

geyslan: > I do know this was under Nadav's radar (specially after https://github.com/aquasecurity/tracee/issues/3003), but I wanted to bring attention to it:

> - I believe our engine list might not be up-to-date with existing container engine setups (and we might need to refactor the code).
> - k8s enrichment should be done talking to k8s API server.

@rafaeldtinoco I've closed the #3003 since it was considering only GKE that I tested and didn't detect any issue. But this is the continuation of the analysis in other env/platforms: https://github.com/aquasecurity/tracee/issues/3003#issuecomment-1771567819

rafaeldtinoco: Thanks Geyslan, left you a comment there, thanks for all the effort with that.

rafaeldtinoco:
I'm merging this commit at PR https://github.com/aquasecurity/tracee/pull/3591 with just a few comments added. Thanks for the contribution @hagitaquasec

PR #3594, opened 2023-10-20
chore(events): rename syscall hooking arguments

rafaeldtinoco:
After merging the syscall hooking changes, the E2E signature needs fixing (which Im providing) and then I realized the argument names contained "." (which isn't something used in any other event so far). I changed the names and I'm providing fixes to the e2e tests (with signature and signature test changes as well).


geyslan (approved):
LGTM

Issue #3595, opened 2023-10-20
hooked_syscall event needs more privileges in older kernels


rafaeldtinoco:
## Description

Tracee had recently 2 commits added regarding hooked_syscall (reverted at https://github.com/aquasecurity/tracee/pull/3597):

commit c0f4b9314 (origin/main, origin/HEAD)
Author: Rafael David Tinoco <rafaeldtinoco@gmail.com>
Date:   Fri Oct 20 17:52:08 2023

    chore(events): rename syscall hooking arguments (#3594)

commit 0f49c0a24
Author: Ori Glassman <ori.glassman1@gmail.com>
Date:   Wed Sep 27 05:40:45 2023

    feat(events): refactor hooked_syscall event

But, since the event was only being tested by E2E, this issue in kernel 5.4 wasn't observed:

```
$ sudo ./dist/tracee --output json --events hooked_syscall
{"level":"error","ts":1697839175.6181216,"msg":"Error populating expected syscall table array: failed to update map expected_sys_call_table: operation not permitted"}
```

This particular permission issue can be mitigated by:

```go
	err = capabilities.GetInstance().EBPF(
		func() error {
			return t.populateExpectedSyscallTableArray(expectedSyscallTableMap)
		},
	)
	if err != nil {
		logger.Errorw("Error populating expected syscall table array: " + err.Error())
		return
	}
```

Not sure if the map updates require other capabilities in different kernels, or because older kernels capabilities are different (cap_sys_admin + paranoid value in sysfs).

But then, there is probably something else going on in these older kernels:

```
ubuntu@ip-172-31-16-208:~/tracee-rafael$ sudo ./dist/tracee --output json --events hooked_syscall
{"level":"error","ts":1697840545.6672323,"msg":"Error populating expected syscall table array: symbol not found: system___x64_sys_close_range"}

ubuntu@ip-172-31-16-208:~/tracee-rafael$ sudo ./dist/tracee --output json --events hooked_syscall
{"level":"error","ts":1697840565.2434964,"msg":"Error populating expected syscall table array: symbol not found: system___x64_sys_pidfd_getfd"}

ubuntu@ip-172-31-16-208:~/tracee-rafael$ sudo ./dist/tracee --output json --events hooked_syscall
{"level":"error","ts":1697840571.7882419,"msg":"Error populating expected syscall table array: symbol not found: system___x64_sys_process_madvise"}

ubuntu@ip-172-31-16-208:~/tracee-rafael$ uname -a
[...]

PR #3596, opened 2023-10-20
fix(event): fix hooked_syscall privilege in older kernels

rafaeldtinoco:
Fixes: #3595


rafaeldtinoco:
There is something else to be investigated for 5.4 kernels (not only this).

Issue #3593, opened 2023-10-20
kubernetes enrichment not working on kind/minikube


josedonizetti:
Below is a tabled summary of the results of the analysis from #3003:

rafaeldtinoco:
I'm also getting this at E2E:

```

[2023-10-26T02:45:22.061Z] 2023-10-26 02:45:21,973 [root] INFO Checking test results on json foramt

[2023-10-26T02:45:22.061Z] 2023-10-26 02:45:21,973 [root] ERROR json: tracee has errors:

[2023-10-26T02:45:22.061Z] {"level":"error","ts":1698288232.5669699,"msg":"error triggering container enrich in control plane","error":"containers.(*Containers).EnrichCgroupInfo: container already deleted"}

[2023-10-26T02:45:22.061Z] 

[2023-10-26T02:45:22.061Z] 2023-10-26 02:45:21,973 [root] ERROR json: signatures were not triggered on cv_interactive_container: ["Interactive terminal in container detected"]
```

now that container enrichment was enabled by default at https://github.com/aquasecurity/tracee/pull/3610. I might take a look tomorrow for this specific issue if you need/want.

rafaeldtinoco:
Behavior was introduced at: https://github.com/aquasecurity/tracee/pull/3325

josedonizetti:
@rafaeldtinoco maybe if we don't want to change this logic now, at least for this release we could change the log level to debug, because the log error is appearing a lot now on k8s.

PR #3598, opened 2023-10-22
fix: enrichment for kind and minikube

josedonizetti:
fix https://github.com/aquasecurity/tracee/issues/3593

The PR https://github.com/aquasecurity/tracee/pull/3325 introduced an optimization to avoid querying CRI on short live cgroups, this works well for all bigger gke platforms I tested (aws, azure, gke, digital ocean), but it doesn't work for minikube/kind, because the cgroup path is not accessible to a container https://github.com/aquasecurity/tracee/blob/main/pkg/containers/containers.go#L365-L374

As this is a special case for development tools only, I did a hack to not validate this whenever it is running on minikube or kind, so enrichment works. 

The PR uses the downward API to know the name of the node, if the name of the node is either minikube or kind it ignores the cgroup dead valiation.

### DIFF
--- a/deploy/helm/tracee/templates/daemonset.yaml
+++ b/deploy/helm/tracee/templates/daemonset.yaml
@@ -43,6 +43,10 @@ spec:
           env:
             - name: LIBBPFGO_OSRELEASE_FILE
               value: /etc/os-release-host
+            - name: NODE_NAME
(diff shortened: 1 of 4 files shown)


geyslan (approved):
LGTM

PR #3600, opened 2023-10-22
Bugfix/capture write address error

AlonZivony:

2826f58ef **fix(ebpf): fix prefix check in capture write**

Fix ##3599


AlonZivony:
Has yet to be checked against kernel 6.5

Issue #309, opened 2020-11-15
Add io_uring tracepoints


yanivagman:
io_uring is a relatively new and interesting interface between the user space and the kernel.
It is built around a ring buffer in memory shared between user space and the kernel, that allows the submission of operations (and collecting the results) without the need to call into the kernel in many cases. 

 With io_uring, a program doesn't have to call system calls to performs operations like open, close, read, write and more (for more info: https://lwn.net/Articles/810414/). For this reason, it is also interesting from a security perspective.

We should add support in tracee for the relevant io_uring tracepoints, as defined in:
https://github.com/torvalds/linux/blob/master/include/trace/events/io_uring.h

yanivagman:
@AsafEitani @idanr1986 @roikol @AlonZivony any plans to add these events? I think they are a blind spot we might have in newer kernels

rafaeldtinoco:
I'm afraid that is also true for aio_writes/aio_reads, no ? One could use async I/O instead of regular I/O as a blind spot I believe.

io_submit() -> aio_read/aio_write() -> call_read/write_iter() -> file_operations "write_iter" implementation (all kinds of backing drivers such as block, char, net, vhost, ext4, fat, fuse, nfs, overlays, ...).

itaysk:
io_uring was merged into libuv which means it might get wide adoption soon https://github.com/libuv/libuv/pull/3952

PR #3603, opened 2023-10-23
feat(events): add io_uring visibility

roikol:

```
Date:   Mon Oct 23 16:52:27 2023 +0300

    feat(events): add io_uring visibility

    add events to allow visibility to io_uring operations in the system.
    the added events are:
    io_uring_create - creation of io_uring queue
    io_uring_submit_req - submission of io_uring request
    io_write - operation of write using io_uring
```

### 3. Other comments

close: #309


roikol:
current status:

**io_uring_create event doesn't work on kernels v5.1 - v5.4:**
what should've solved it is was the probe `io_sq_offload_start` which exists in kernels v5.1 - v5.4, but for some reason this probe fails to load. 
we need to find another alternative probe to get the information from.

**the tests don't pass because of the context of the event:**
io_uring mechanism in the kernel changes the context of the task - 
 - if polling mode is used in `io_uring_create` syscall, then eveny io_uring operation is done under the context of a kernel polling thread.
 - if the operation that is being submitted to run in io_uring is blocking, then the kernel assigns a kernel worker thread to do the operation.
in this PR - i attempted to "seat" where these context changes are happening, and override the current context (either kernel polling thread or kernel worker thread) with the original context of the issuer (the userspace process).
currently i only override the `p.event->context`, but i believe that `p.task_info` should be taken care of as well.

**missing probes warnings at tracee startup:**
i've used different probes for different kernel versions to support those events. so some of them won't be available in all the kernel versions.
this is expected, but we may want the silent those warnings.

geyslan:

NDStrahilevitz:
Closing in favor of #5069

PR #3606, opened 2023-10-24
Feature/virtual mem flags parse

AlonZivony:

feat(events): parse access_remote_vm args

Parse vm and gup flags arguments of access_remote_vm.

### DIFF
--- a/docs/docs/events/builtin/extra/access_remote_vm.md
+++ b/docs/docs/events/builtin/extra/access_remote_vm.md
@@ -1,0 +1,34 @@
+# access_remote_vm
+
+## Intro
+access_remote_vm - gain access to the virtual memory of a separate process through the use of the procfs mem file.
+
+## Description
+This event marks access attempt of a process to the virtual memory of another process using the procfs mem file associated with that specific process (/proc/<pid>/mem).
+It is a more elaborated event than the `security_file_open` of the mem file.
+
+## Arguments
+* `remote_pid`: `int`[K] - PID of the process the memory area belongs to.
+* `start_address`: `void *`[K] - Start address of the operation.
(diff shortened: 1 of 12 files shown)


AlonZivony:
@rafaeldtinoco can you tell me your opinion on the last commit?

rafaeldtinoco (changes requested):
The first commit is being already rebased/reviewed in another PR. The last commit is about the flags parsing and I think we should do the version distincting inside the helper package with a generic function that opts for one or another (or something similar).

AlonZivony:
> The first commit is being already rebased/reviewed in another PR. The last commit is about the flags parsing and I think we should do the version distincting inside the helper package with a generic function that opts for one or another (or something similar).

The problem with this approach is that you don't know how to user want to use the API.
If this library will be used on the flag from a different machine than the one it was traced in, then the function will return the wrong flag.
So I think the current API is good, but we can add a function of `ParseGUPFlagsCurrentOS` with the logic I have here.

rafaeldtinoco:
> If this library will be used on the flag from a different machine than the one it was traced in, then the function will return the wrong flag.

You're probably thinking about analyze mode. In that case, we're also missing OSinfo information from the running environment, no ? Thinking about it, like said previously, I believe we should add `metadata` to the output file (that would include information about running environment). And more on that, if that is true than the function from the helper should accept "current running version" argument and decide by itself.

Either way, I think the decision to parse legacy or not should be delegated to the helper itself. `ParseCurrent` or `ParseSpecific(version)` or anything like it, up to you.

geyslan:

yanivagman:
Not going to be implemented

PR #3608, opened 2023-10-25
docs(signatures): add signatures documentation files

rafaeldtinoco:
Still need to integrate signature files in mkdocs and check rendering.


rafaeldtinoco:
I am not yet changing "all terms" from "signature" to "security events". I'll do another PR only doing that (making sure the same term is used in all documentation, but still keep "signature" when it makes sense).

For now, its about the default security events (signatures) documentation.

Issue #3182, opened 2023-05-31
structured config file 


josedonizetti:
Currently, several elements of the configfile are configured like flags with strings, it would be good to have a structure, for example, instead of

```
output:
    - webhook:HTTP://localhost:8080?timeout=5
```

we had

```
output:
    - webhook:
          url: http://localhost:8080
          timeout=5
```

the same for other elements, for webhook we could use it to support alias so the user can use different wehooks on his policies

```
output:
    - webhook:
         name: webhook1
          url: http://localhost:8080
          timeout=5
   - webhook:
         name: webhook2
          url: http://localhost:8081
          timeout=5
```

geyslan:
Those options are bound to viper, so in order to bring the structured options and keep the old way of passing string flags (similar to cli flags), we must rely on type coercion to detect if the `output` options (in this example) are set as legacy or structured.

We have all these that can be structured:

```yaml
cache:           # [type|mem-cache-size]
    - none
proctree:        # [process|thread]
    - none
capabilities: [] # [bypass|add|drop]
crs: []          # <runtime:socket>
log:             # [debug|info|warn...]
    - info
output:          # [json|none|webhook...]
    - json
rego: []         # [partial-eval|aio]
```

For each one, we should do type checks and try to detect if it's legacy (string) or structured.

So, to not touch in so much code right now I ask if the following approach (add option-structured) should be a fit:

```yaml
output:
    - webhook:HTTP://localhost:8080?timeout=5s
output-structured:
    - webhook:
        name: webhook1
        host: http://localhost
        port: 8000
        timeout: 5s
    - webhook:
        name: webhook2
        host: http://localhost
        port: 9000
        timeout: 5s
```

geyslan:
I believe we'll be able to keep both use in the same keys:

![image](https://github.com/aquasecurity/tracee/assets/3372117/72c397a1-8888-4538-9619-951d79dde5e5)

PR #3612, opened 2023-10-25
Handle structured config file

geyslan:
Close: #3182 

d3f9ba9fb **chore(config): add gotemplate and content-type**
86c120975 **chore!: change crs flag/key to cri**
d1532749d **feat(docs): update docs for structured config file**
9092b9941 **feat(cobra): add parser for structured config file**

d3f9ba9fb **chore(config): add gotemplate and content-type**

```
Since #3622, webhooks can be configured with a gotemplate and a
content-type. This commit adds these two fields to the config file.
```

86c120975 **chore!: change crs flag/key to cri**

```
BREAKING CHANGE: change crs flag to cri
```

d1532749d **feat(docs): update docs for structured config file**

```
This also changes the default config file to be a structured YAML file
and adds a new example config file 'global_config_cli.yaml'  that uses
the CLI format.
```

9092b9941 **feat(cobra): add parser for structured config file**

```
This commit adds a parser for a structured config file using viper.

Now, the config file can be written in a more structured way, like:

cache:
    type: mem
    size: 1024

The old way (cli flags) is still supported.

cache:
    - cache-type=mem
    - mem-cache-size=556
```

### 2. Explain how to test it

`sudo ./dist/tracee --config ./examples/config/global_config.yaml`
`sudo ./dist/tracee --config ./examples/config/global_config_cli.yaml`

### DIFF
--- a/examples/config/global_config.yaml
+++ b/examples/config/global_config.yaml
@@ -1,10 +1,37 @@
 blob-perf-buffer-size: 1024
 cache:
-    - none
+    # - none # cache-type={none,mem}
+    - cache-type=none
+    # - mem-cache-size=556
+# cache:
+    # type: mem
+    # size: 256
 proctree:
-    - none
-capabilities: []
-crs: []
+    # - source=events
+    source: events
+    cache:
+        process: 8192
+        thread: 4096
+capabilities:
+    # - bypass=false
+    # - add=cap_kill,cap_syslog
+    bypass: true
+    add:
+        - cap_sys_admin
+        - cap_syslog
+    drop:
+        - cap_chown
+crs:
+    - containerd:/var/run/containerd/containerd.sock
+    - docker:/var/run/docker.sock
+    - runtime:
+        name: containerd
+        socket: /var/run/containerd/containerd.sock
+    - runtime:
+        name: docker
+        socket: /var/run/docker.sock
+
+
 healthz: false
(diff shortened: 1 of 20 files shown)


geyslan:
This doesn't inject the new configuration structures to be used later by flag validations. That would be interesting, but the effort would be much greater. In any case, these structures take care of outputting the flags - `CLIFlagger.flags()` - in the format they need to be parsed (prepared).

examples/config/global_config.yaml
```diff
@@ -1,21 +1,136 @@
...
+    bypass: false
+    # add:
+    #     - cap_sys_admin
+    #     - cap_syslog
+    # drop:
+    #     - cap_chown
+
 containers: false
-crs: []
+crs:
```
yanivagman, line 21: We should rename this to cri, which is much more clear (can be in a different PR)

yanivagman (commented):
Looks good

examples/config/global_config.yaml
geyslan: Done: 33686e61ac786002789b4ae560c8bd685f14e972

rafaeldtinoco (approved):
I have to check Lunar 6.2 test for hooked_syscall, so Ignore that error. Loved the amount of tests, for each specific option. Job well done. LGTM.

PR #3615, opened 2023-10-26
fix(containers): dead containers might have been enriched already

rafaeldtinoco:


rafaeldtinoco:
Converted to DRAFT since it does not solve the E2E issue (I'll have to dig it further).

Issue #3607, opened 2023-10-24
Improve docs structure


itaysk:
every section should have an overview page with talks about the purpose of the section and how to use it via config file.
then there can be other dedicated pages for cli flags, and for k8s experiences.

policies section
move filters page here as the cli experience
if/when we have kubernetes operator is will join here as k8x experience

integrations
container enrichment - we are making it default (separate task). move the setup related content to installation or advanced. then can remove the page.

events
restructures as:
- syscalls events
- network events
- security events
- lsm events
- containers events
- misc

create page for every security event like the rest
delete analyze page or reframe around testing custom signatures

rafaeldtinoco:
1. Container Enrichment being the default now: https://github.com/aquasecurity/tracee/pull/3610 (and cmdline safeguard --no-containers for this release, to be removed in the next). I changed documentation to reflect that.

> Still need "Integrations" session name to be changed I believe.

2. Events restructured:

I'm still going to change https://aquasecurity.github.io/tracee/v0.18/docs/events/overview/ index page. I have mostly added signatures documentation (https://aquasecurity.github.io/tracee/dev/docs/events/builtin/signatures/)

After some thinking, I think the "security events" (Signatures) pages should be the first in the menu (because syscalls list is extensive, if user opens it they will just forget about the rest).

For now I kept:

![image](https://github.com/aquasecurity/tracee/assets/7395852/74da19c3-26f8-43cb-bcef-6a7345290fe3)

But I believe "extra" events should become "misc" ? and maybe add a "lsm events" explaining what they mean (avoiding TOCTOU etc).

itaysk:
> But I believe "extra" events should become "misc" ? and maybe add a "lsm events" explaining what they mean (avoiding TOCTOU etc).

@rafaeldtinoco is this different from what the issue description suggests?

- syscalls events
- network events
- security events
- lsm events
- containers events
- misc

rafaeldtinoco:
Just in order (having security events first). I'll also remove the word "events" at the end of each of them (keeping "Events" -> Security, Network, Misc, LSM, Syscalls, ...)

AnaisUrlichs:
I think the Tracee docs overcomplicates things into separate terminology -- at the moment:

The `Events` Section details how to configure Tracee, including events it should track through the CLI https://aquasecurity.github.io/tracee/dev/docs/events/overview/#read-in-cli 

[...]

PR #3618, opened 2023-10-26
docs: merge tracee policies and filtering

AnaisUrlichs:
Based on the following issue: https://github.com/aquasecurity/tracee/issues/3607
Merging the policies and filtering section of the docs

### DIFF
--- a/docs/docs/events/overview.md
+++ b/docs/docs/events/overview.md
@@ -6,13 +6,15 @@ This section documents all of the different events that Tracee exposes.
 
 Tracee uses eBPF technology to tap into your system and give you access to hundreds of events that help you understand how your system behaves. The events can be specified either through CLI with [filters] or with [policies].
 
-eg:
+### Using the CLI
 
 Tracing `execve` events with [filters]:
 
 ```console
 tracee --events execve
 ```
+
+### Through the Tracee Helm Chart installation
 
 Tracing `execve` events with [policies]:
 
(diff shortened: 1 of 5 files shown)


docs/docs/policies/filtering.md
```diff
@@ -1,4 +1,6 @@
-# Tracing Event Filtering
+# Tracing Events through the Tracee CLI by applying filters
```
itaysk, line 1: "Applying filters with Tracee CLI"?

docs/docs/policies/filtering.md
```diff
@@ -1,4 +1,6 @@
...
+
+This section showcases how to trace events through the Tracee CLI by applying filters.
```
itaysk, line 3: how to filter events through the Tracee CLI?

docs/docs/events/overview.md
itaysk, line 1: I think the entire page needs to be rewritten, let's pair on this

docs/docs/policies/index.md
```diff
@@ -7,7 +9,7 @@ Lastly, policies require a scope. The scope details which resources the policy a
 
 You can load multiple (up to 64) policies into Tracee using the --policy flag providing a path to the policy file.
 
-Following is a sample policy:
+Following is a sample policy for the Tracee Kubernetes usage:
```
itaysk, line 12: no that's actually a valid policy regardless of k8s

docs/docs/policies/index.md
```diff
@@ -47,6 +49,17 @@ While specifying event filters is optional, policies must have the `name`, `desc
 
 More information about defining a scope and the available filters can be found in the next sections.
 
+## Tracee Kubernetes Policy configuration
+
+Through Querying the Tracee Configmap, users can access the default configuration and apply changes:
+```console
+kubectl edit ds/tracee -n tracee-system
```
itaysk, line 56: you talk about querying the ConfigMap but show how to edit a workload. (also s/querying/editing/)

docs/docs/policies/index.md
```diff
@@ -47,6 +49,17 @@ While specifying event filters is optional, policies must have the `name`, `desc
...
+```
+
+## Tracee CLI Policy configuration
+
+Applying Tracee Policies in the CLI is further detailed within the [filtering section.](./filtering)
```
itaysk, line 61: actually, you can use a policy file on the cli as well

itaysk (changes requested):
this one is a bit complicated, I suggest we pair to rewrite it

docs/docs/policies/index.md
AnaisUrlichs: *"Tracee Configmap that is part of the Daemonset"

docs/docs/policies/index.md
AnaisUrlichs: https://aquasecurity.github.io/tracee/dev/docs/config/overview/ should I link here? we should add an example onto this page -- what do you think?

AnaisUrlichs: Also, I have a bit of an issue with the ways Policies can be applied/changed through Tracee -- 

The Policy section details https://aquasecurity.github.io/tracee/dev/docs/policies/ providing an example Policy for Kubernetes
And tracee can also be configured in the CLI through flags and the `--config` command which is detailed here: https://aquasecurity.github.io/tracee/dev/docs/config/overview/
Under Config in the docs -- so I could link to it from the Policies section
Here is my issue, also under `Config` in the docs, is detailed how to Configure Tracee in Kubernetes but that example seems very specific to configuring Tracee, not to applying policies https://aquasecurity.github.io/tracee/dev/docs/config/kubernetes/

So, it would be a bit confusing to link to the `Config` section from the `Policies overview` section -- what do you think of 
Adding to the Policy Overview section an example of creating a config file for the CLI to modify policies and providing it through the `--config` flag and then saying for "further options on using the `--config` flag and configuring Tracee logs" to look at the `Config` section of the docs

docs/docs/policies/index.md
AnaisUrlichs: I believe it will be a separate ConfigMap in the Operator

docs/docs/policies/index.md
itaysk: exactly https://github.com/aquasecurity/tracee/pull/3330

PR #3619, opened 2023-10-26
feature(k8s): add default events to k8s policies

josedonizetti:

Fix https://github.com/aquasecurity/tracee/issues/3323

This PR adds the default events to the kubernetes default policy, so users have the same experience if they are using k8s or the tracee binary.


rafaeldtinoco (approved):
LGTM

PR #3610, opened 2023-10-25
chore(containers): make feature flag a safe guard

rafaeldtinoco:
commit b39d88457 (HEAD -> no-containers-default, rafaeldtinoco/no-containers-default)
Author: Rafael David Tinoco <rafaeldtinoco@gmail.com>
Date:   Wed Oct 25 12:09:09 2023

    fix(definitions): debug shows wrong logs for sig events

commit b067fd385
Author: Rafael David Tinoco <rafaeldtinoco@gmail.com>
Date:   Wed Oct 25 11:51:14 2023

    chore(containers): make feature flag a safe guard
    
    - make default behavior (--containers) to enrich containers
    - keep a safeguard measure (change cmdline flag to --no-containers)

PR #3623, opened 2023-10-26
chore(k8s): remove containers config option

josedonizetti:

This PR removes the `containers` config from the kubernertes files, the flag was removed after the PR https://github.com/aquasecurity/tracee/pull/3610


geyslan (approved):
LGTM

PR #3631, opened 2023-10-29
fix: remove noisy debug log

josedonizetti:
Removing this debug, it is too noisy, really hard to test anything.


rafaeldtinoco:
Yep, I wanted this for long time now. Thanks.

rafaeldtinoco (approved):
LGTM

Issue #3637, opened 2023-10-31
revisit dependency on kconfig probing


itaysk:
when initializing, Tracee probes the running kernel config via either  `/boot/config-$(uname-r)` or `/proc/config.gz`.
On some systems, [as it turns out](https://github.com/orbstack/orbstack/issues/761), it is not an accurate enough. 

After checking with @rafaeldtinoco he indicated that "we're only using kconfig to check for "syscall wrappers" (which are enabled in almost every distro and arch)". 
If that's the case we should check if we can remove the flaky kconfig checking.
If there's another reason to relying on it, we should consider disabling it with a flag (in which case I'd also consider skipping the os-release detection).
either way we should update the documentation appropriately.

PR #3642, opened 2023-10-31
fix(initialize): do not fail kconfig options

geyslan:
Close: #3637 

8a45f7a95590a51f2104ba530e438fa9f5a65ee0 **fix(initialize): do not fail kconfig options**

```
When initializing, we should not fail if the kernel config options
are not found. Let it fail if the option is required further down
the line.
```

### DIFF
--- a/docs/docs/deep-dive/os-info.md
+++ b/docs/docs/deep-dive/os-info.md
@@ -76,13 +76,15 @@ Tracee needs access to kconfig file (/proc/config.gz OR /boot/config-$(uname -r)
 2. Provide kconfig variables to its eBPF counter-part (so eBPF program take decisions)
 
 !!! Warning
-    Tracee **should NOT fail** when it cannot find a kconfig file:
+    Tracee **should NOT fail** when it cannot find a kconfig file or needed options:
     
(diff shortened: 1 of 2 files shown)


itaysk:
nit: this is not a chore since it's a change in behavior. it could be a feature or a bug. I opened the issue as a bug since it contradicted my expectation.

itaysk:
is this behavior consistent with os-release not found/not including a key? should it be?

geyslan:
> is this behavior consistent with os-release not found/not including a key? should it be?

The difference is the logger level (debug for os-release miss):

https://github.com/aquasecurity/tracee/blob/f949b0fff6a72617cbb30c5233ce730d8de65db6/pkg/cmd/cobra/cobra.go#L82-L92

rafaeldtinoco (approved):
LGTM (as a mitigation). I would open a new issue for the kconfig refactoring (removing from libbpfgo, making sure it is still needed in tracee, make it simpler if it is). I know there are signatures that assume things that are only possible to know from kconfig (at least in an easy way), such as size of pages for some architectures, paging level hierarchy size, etc.

@geyslan

PR #3646, opened 2023-10-31
fix: structured k8s config

josedonizetti:
Fix k8s config file. The PR isn't adding all options but the most common ones, if a user needs to optmize all options he can use the option to inject a full config file, I'll create an issue to add the remaining options in the next release.


geyslan:
@josedonizetti a conflict emerged.

PR #3647, opened 2023-11-01
docs: add kubernetes operator

josedonizetti:
Adds docs for kubernetes section under policies. I didn't link it to mkdocs, because I'm not sure where it should go, considering we are changing the structure.

### DIFF
--- a/docs/docs/policies/kubernetes.md
+++ b/docs/docs/policies/kubernetes.md
@@ -1,0 +1,35 @@
+# Kubernetes
+
+## Custom Resource Definition
+
+Tracee policies can be seamlessly integrated into Kubernetes using Custom Resource Definitions (CRDs). When Tracee is installed, the CRD is automatically applied, including a default policy. One can interact with Tracee policies as follows:
+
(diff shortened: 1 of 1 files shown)


docs/docs/policies/kubernetes.md
```diff
@@ -0,0 +1,35 @@
...
+```shell
+kubectl get policies.tracee.aquasec.com
+```
+
+One can manage policies using standard kubectl commands. For example, to create, update, or delete a policy:
+
+Create: Apply a new policy using the kubectl apply command.
+
+```shell
+kubectl apply -f your-policy.yaml
```
rafaeldtinoco, line 18: Should we point to an example of policy here ? Or to our "policy examples" directory ? Of cours one could simply view the `policies.tracee.aquasec.com/default-policy` installed, but maybe pointing to examples could be good (or another already existing page)

josedonizetti: @rafaeldtinoco @itaysk is using the same structure on the k8s guide, so I'll mark this as done.

PR #3649, opened 2023-11-01
fix: k8s static deploy

josedonizetti:
Fix k8s static


rafaeldtinoco (approved):
LGTM. Better this way, thanks for addressing it.

PR #3616, opened 2023-10-26
docs: redo readme and getting started

itaysk:
related to #3605 

1. improved readme/index
2. moved installation under docs

I didn't touch the individual install guides yet, it will be easier to to it one-by-one after this PR is merged

PR #3652, opened 2023-11-01
docs: kubernetes guide

itaysk:
#3616 

I didn't run the commands since I have issues on my env, appreciate if you took a close look at those.


docs/docs/install/kubernetes.md
```diff
@@ -47,17 +49,35 @@ tracee-fcjmp   1/1     Running   0          4m11s
...
+
+When you are ready to apply a policy, it's as easy as `kubectl apply -f your-policy.yaml`. More details [here](../policies/usage/kubernetes.md).
 
-## Exercising a security event
+## Configuring Tracee
+
+In some cases you will need to configure Tracee to your preferences. For example, to change the output event format, or to set a different log level. To learn more about available configuration options please see the [configuration](../install/config/index.md) section.
+
+When you are ready to make a configuration change, it's as easy as `kubectl edit cm -n tracee tracee`. More details [here](../install/config/kubernetes.md).
```
josedonizetti, line 78: As we show how to install with helm, I would prefer said one can customize the config through helm. Something like:

```
You can inject your own config file if installing with helm, with `helm .. --set-file traceeConfig=/path/to/file`, or if you want just to customize one option, our helm supports each config option individually, for example to increase your cache size `helm install .. --set config.cache.size=1024`
```

docs/docs/install/kubernetes.md
```diff
@@ -47,17 +49,35 @@ tracee-fcjmp   1/1     Running   0          4m11s
...
-## Interacting with Tracee
+Once installed, Tracee immediately starts producing events. Since Tracee is deployed as a DaemonSet, a Tracee Pod is running on every node in the cluster. Every Tracee Pod is monitoring the node it is running on.
 
-Once installed, Tracee immediately starts producing system activity events, such as processes and containers activity, network activity, and more. To see the events that Tracee produces, use can use the `kubectl logs` command.
+## Viewing Events
+
+The easiest way to tap into the log stream of all Tracee Pods is with the `kubectl logs` command:
 
 ```console
-kubectl logs -f daemonset/tracee -n tracee-system
```
AnaisUrlichs, line 55: the Helm Chart is usually installed in the tracee-system namespace, why don't we add it here?

docs/docs/install/kubernetes.md
itaysk: I took your suggestion to use helm but didn't explain all the possible options (helm values file, helm value override, kustomize), since I think these more complete information belongs in the `install/config/kubernetes` doc

AnaisUrlichs (approved):

Issue #3650, opened 2023-11-01
revise default policy


itaysk:
currently it's too noisy. probably want to keep: signatures + container events

we can create another example policy with the things we removed which is more verbose

rafaeldtinoco:
I think we should also do for the docker image, right ?

https://github.com/aquasecurity/tracee/pull/3654#issuecomment-1789928851

If that is true (and without touching the "default set" of events).

rafaeldtinoco:
The container image default events was made just like the regular k8s deployment at the PR https://github.com/aquasecurity/tracee/pull/3687.

PR #3654, opened 2023-11-01
Fix k8s default policy

josedonizetti:
Review k8s default policy

Fix https://github.com/aquasecurity/tracee/issues/3650

I understood the review was only for kubernetes, correct?


geyslan (approved):
LGTM

rafaeldtinoco:
So the fix for the container image is at: https://github.com/aquasecurity/tracee/pull/3687

PR #3673, opened 2023-11-05
Refactor/full analyze mode

### LINKED ISSUE #3520: Change analyze mode to use the normal pipeline
The current analyze mode is pretty much a replacement to the previous **tracee-rules** binary.
This is making it quite shallow, as many new features has been added since the union of the binaries, making them inaccessible to the analyze mode.
To support new features like the process tree we should use the normal pipeline of the events, starting from the step after decoding the events. However, we will need to skip things that are not accessible from a remote machine, like file system operations, etc.

AlonZivony:

Fix #3520


AlonZivony:
@rafaeldtinoco @josedonizetti the current analyze mode is working (checked manually) with process tree.
I need you LGTM to do the work left of tests, format and documentation.

AlonZivony:
Optional feature - if stdin of tracee upon startup does not point to /dev/pts (normal stdin path), then run automatically `tracee analyze`.
For example:
`cat tracee_output.json | ./tracee -e ld_preload` will run in analyze mode on the events from `tracee_output.json`.
This will make the behavior similar to `python` for example.

A point of thought:
Do we really want analyze command?
We can just assume by the use of the `--input` flag that we are in analyze mode, avoiding the clumsy command line.

NDStrahilevitz:
@AlonZivony When you go back to working on this, is it reasonable in your opinion to include the DNS Cache?

AlonZivony:
> @AlonZivony When you go back to working on this, is it reasonable in your opinion to include the DNS Cache?

This PR is not giving the analyze mode all the options of the normal mode.
You can see that I didn't support many flags of Tracee.
I am not sure about this DNS cache implementation you want, but I guess it could be possible. But not in this PR (it is big enough as it is)

pkg/events/usermode.go
```diff
@@ -38,12 +39,13 @@ func InitNamespacesEvent() trace.Event {
 	initNamespacesArgs := getInitNamespaceArguments()
 
 	initNamespacesEvent := trace.Event{
```
NDStrahilevitz, line 41: I suggest you turn this to a data source.

pkg/producer/producer.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"fmt"
+	"github.com/aquasecurity/tracee/pkg/config"
+	"github.com/aquasecurity/tracee/pkg/errfmt"
+	"github.com/aquasecurity/tracee/types/trace"
+)
+
+// EventsProducer is a type that is able to generate events
+type EventsProducer interface {
```
NDStrahilevitz, line 11: I think the eBPF event decoder could be considered such an object no? This could be a useful abstraction.

pkg/cmd/flags/input.go
```diff
@@ -0,0 +1,76 @@
+package flags
+
+import (
+	"fmt"
+	"github.com/aquasecurity/tracee/pkg/capabilities"
+	"github.com/aquasecurity/tracee/pkg/config"
+	"github.com/aquasecurity/tracee/pkg/errfmt"
+	cap2 "kernel.org/pub/linux/libs/security/libcap/cap"
```
NDStrahilevitz, line 8: why alias?

pkg/ebpf/tracee.go
```diff
@@ -122,6 +123,8 @@ type Tracee struct {
 	streamsManager *streams.StreamsManager
 	// policyManager manages policy state
 	policyManager *policyManager
+	// producer produce events in analyze mode instead of eBPF programs
+	producer producer.EventsProducer
```
NDStrahilevitz, line 127: Perhaps not relevant for this PR, but I see a future where this is a `one-tracee-many-producers` relation. @josedonizetti WDYT?

pkg/ebpf/processor.go
```diff
@@ -118,6 +120,8 @@ func (t *Tracee) registerEventProcessors() {
 		// Convert all time relate args to nanoseconds since epoch.
 		// NOTE: Make sure to convert time related args (of your event) in here.
 		t.RegisterEventProcessor(events.SchedProcessFork, t.processSchedProcessFork)
+	}
+	if !t.config.Output.ExportAnalyze {
```
NDStrahilevitz, line 124: Why disable? This adds another layer of confusion to normalization. I'd also urge again that normalization is done as the first procession as its easier to reason about for a user writing an extension.

NDStrahilevitz, line 124: Why disable? This adds another layer of confusion to normalization. I'd also urge again that normalization is done as the first procession as its easier to reason about for a user writing an extension.

pkg/ebpf/tracee.go
```diff
@@ -820,6 +834,10 @@ func (t *Tracee) getOptionsConfig() uint32 {
 		cOptVal = cOptVal | optForkProcTree // tell sched_process_fork to be prolix
 	}
 
+	if t.config.Output.ExportAnalyze {
```
NDStrahilevitz, line 837: Sounds like exportanalyze should just enable the process tree, instead of having its own logic.

NDStrahilevitz (commented):
The commits need to be organized. It also seems like this is still a WIP. Are you planning on merging this release or the next one?
Anyway, there should also be a test for the mode before merging.

AlonZivony:
> The commits need to be organized. It also seems like this is still a WIP. Are you planning on merging this release or the next one? Anyway, there should also be a test for the mode before merging.

You are right, this is more of a POC PR.
I didn't want to implement the specifics before the overall idea is approved.
You can think of this as a DR stage, where I want to have approval on all the patterns, abstractions and general solution.
As you noticed, I didn't implement tests, kept many features blocked for this to work in this current stage and didn't document to much (as it is prone to change).
If I get the OK for this, we can discuss how to move on with the full implementation.

pkg/events/usermode.go
AlonZivony: So for now this is used as state initializer.
This info could be also saved to a datasource.

pkg/producer/producer.go
AlonZivony: Yea, this was the idea :)
I tried to fit to the current design

pkg/ebpf/processor.go
AlonZivony: So we need the timestamps to match the kernel times in the analyze mode.
We can do it by either change the timestamps in the back when reading in analyze mode, or when exporting.
For now it was more convenient for this POC to export kernel times.

pkg/ebpf/tracee.go
AlonZivony: You are right :)

pkg/events/usermode.go
NDStrahilevitz: I'm saying we make this a builtin data source, instead of having an event which initializes some shared state (which is what data sources are for). I think this will only be in v0.21 or v0.22, so this can be done as a prep PR.

AlonZivony: But you will still need an event to initialize the datasource in analyze mode...

NDStrahilevitz:
> > The commits need to be organized. It also seems like this is still a WIP. Are you planning on merging this release or the next one? Anyway, there should also be a test for the mode before merging.
> 
> You are right, this is more of a POC PR. I didn't want to implement the specifics before the overall idea is approved. You can think of this as a DR stage, where I want to have approval on all the patterns, abstractions and general solution. As you noticed, I didn't implement tests, kept many features blocked for this to work in this current stage and didn't document to much (as it is prone to change). If I get the OK for this, we can discuss how to move on with the full implementation.

The ability added in this PR is necessary for us to deprecate tracee-ebpf and tracee-rules. So this, or something similar is necessary as well. And it does seem like a useful feature to have anyway.

I think the concept of the event producer is good, and I already have some ideas on where to use it in the future. 

I think it would be even better if we would make the pipeline its own object, which we can build as needed depending on the context (basically making it more declarative). I think it would make implementing this idea much nicer. I'm less sure of how easy it would be as a PR (probably not as it would be a big refactor), but I also fear it is becoming more necessary as we approach extensions/plugins and basically allowed external change of behavior in tracee.

pkg/events/usermode.go
NDStrahilevitz: I had in mind using signals for things like that. In addition to what i've written below, I think --export-analyze should give a separate output of signals, which can be fed back to the control plane running in analyze.

You're right, this needs to be an event or a signal. But, it doesn't need to be an externally visible event. So initialize a data source with this info after the userspace code which would previously emit the event. If running with --export-analyze, dump it as a signal as well.

geyslan:

geyslan:
@AlonZivony will you be able to rebase this?

AlonZivony:
> @AlonZivony will you be able to rebase this?

I have a different branch for the PR, as I made some changes after the reviews.
I will close this PR for now so it won't be confusing.

PR #3679, opened 2023-11-09
Feature: DNS Cache datasource

### LINKED ISSUE #3678: Data source for dns<->ip mapping
It would be useful for networking related signatures to be able to check the relation of ip/dns addresses to the other and vice versa.
Tracee can track these relations through it's DNS events and resolving queries itself where needed, and provide them to signatures through a data source. 

The data source query should give the following (a more accurate schema will be given once the PR is finalized)
[string: ip] -> matched dns records
[string: dns] -> matched A and AAAA records + dns aliases and cname

NDStrahilevitz:

9ebaceef5 **docs: add dns cache**
0b785792e **feature: add a dns:ip cache**
96b581fb1 **fix(dns): wrong dns argument type**
0bfbe8f46 **feature: add generic set data structure**

### 2. Explain how to test it

An instrumentation test and in-code e2e test were added.

### 3. Other comments

The PR currently uses a processor on `net_packet_dns` event to populate the cache. This meant that I had to add an additional `processEvent` call in the `deriveEvents` stage of the pipeline.
A possible improvement in the future is to migrate (or add) a control plane signal for this cache.

The cache currently uses the second node based design proposed.

Resolves #3678


rafaeldtinoco:
Ill review this tomorrow, just wanted a heads up on the feature flag. Alon and I discussed about having a feature flag per "enabled feature that would enrich the proc tree" or "just enriching if that event is being traced".

I have also thought about a multi-var feature-flag like for proctree enriching... Something like --proctree enrich:socket,dns,file,bleh,blah together with all other proctree cmdline options available.

Long story short: I don't think we should have feature flag for this particular dns cache feature. Its either one that would us allow to enable many proctree enrichers (such as this one) OR none needed (tracing the event that causes the enrichment would be enough).

What is your thought ?

NDStrahilevitz:
> Long story short: I don't think we should have feature flag for this particular dns cache feature. Its either one that would us allow to enable many proctree enrichers (such as this one) OR none needed (tracing the event that causes the enrichment would be enough).
> 
> What is your thought ?

I think that since this doesn't enrich or touch the process tree (maybe only currently), it has no reason to be a part of its feature flags. I'm not sure where this would plug into the process tree (at least not directly), as the closest the process tree could get to it would be through open fds maybe, and then if those fds are network sockets you could enrich the relevant dns of the connection.

I also don't like just equating the enable flag to enabling the relevant dns event (and ideally i would have a signal eBPF program here, but I didn't want to delay the PR further), because it obfuscates how the dns cache should be enabled. If a user tries to access the dns cache but it's empty because the `net_packet_dns_response` event wasn't enabled in his policy, that's just kind of annoying and unintuitive. The PR is still missing docs, but if it had docs, that would be the only place where this "enabling" method could be located, it's not self documenting.

rafaeldtinoco:
> I'm not sure where this would plug into the process tree (at least not directly), as the closest the process tree could get to it would be through open fds maybe, and then if those fds are network sockets you could enrich the relevant dns of the connection.

Having a per process list of resolved DNS entries would allow blocking network connections of a process that has solved a blacklisted DNS name (just one example, there are many). Think of the process tree as a place to save all this information (opened files, queries names, IPs communicated with, etc).

rafaeldtinoco:
> If a user tries to access the dns cache but it's empty because the `net_packet_dns_response` event wasn't enabled in his policy, that's just kind of annoying and unintuitive. The PR is still missing docs, but if it had docs, that would be the only place where this "enabling" method could be located, it's not self documenting.

This is the problem we need to solve (for data sources & process tree features). I believe we should solve this now or else it might be too late when we have multiple features done.

rafaeldtinoco:
> (and ideally i would have a signal eBPF program here, but I didn't want to delay the PR further)

We also need to discuss signal events and their future. We need to think what we really want from them. For the process tree, they are helpful for busy pipelines, for example, but couldn't live without the regular events logic in place as well. Either way, when we agreed to have them, we missed the discussion of "up to what point" and "what for".

pkg/dnscache/dnscache.go
```diff
@@ -0,0 +1,204 @@
...
+	"github.com/aquasecurity/tracee/types/trace"
+)
+
+var (
+	ErrDnsCacheIsNotInitialized = errors.New("dns cache is not initialized")
+)
+
+const (
+	dnsCacheSize        = 500
+	dnsCacheTTL         = time.Minute * 5
```
rafaeldtinoco, line 24: For TTL, we have to distinguish the 2 different TTL concepts (at least internally) I believe.

1) The name resolution TTL: used by the resolvers (clients) mostly.
2) A internal cache timeout of a cached entry.

Some people say it should be the same as the resolution TTL.

A problem that I thought (for you to discuss with the sponsors of this need being worked):

When doing round-robin load balancing, for example, some DNS servers might return ALL IPs for a specific A, CNAME, AAAA query. Then the client decides which one to use. Some other DNS servers might return one different IP each time you resolve that same A, CNAME, AAAA entry, for example.

My point is, if this datasource is ever used to distinguish to which host a process talked to, then we will need to correlate the process and a specific entry in this `Set` you created (and not index by name used, as one process might have talked to IP1, the other to IP2, the other to IP3, and all of them talking to the same name, and vice-versa).

pkg/dnscache/dnscache.go
```diff
@@ -0,0 +1,204 @@
...
+
+// Add adds the DNS response into the DNS' LRU cache.
+func (dc *DNSCache) Add(dnsResponse trace.DnsResponseData) error {
+	switch dnsResponse.QueryData.QueryType {
+	case "A":
+	case "AAAA":
+	case "PTR":
+	default:
+		return nil // Do not handle other query types
+	}
```
rafaeldtinoco, line 51: I wont ask you to rewrite using dnsRequest, but why didnt you use it ? I mean, the DnsRequest is the generic event that should always exist (instead of Request/Response events that were added only to keep compatibility with existing signatures back then).

rafaeldtinoco:
Alright, gave a good first look. I believe its solid. I left comments here and there mostly for the `Set` utils package. I believe you could have avoided string parsing for `AAAA` `CNAME` `PTR` and things like that if you had used the common `net_packet_dns` event (instead of the response, which was made for keeping signatures working mostly).

Either way, the code per se is good IMO. There are major discussions about why keep data, what for, etc, but the logic is solid and makes sense.

rafaeldtinoco (commented):
left comments

pkg/ebpf/processor.go
```diff
@@ -97,6 +97,15 @@ func (t *Tracee) registerEventProcessors() {
 	// Processors regitered even if process tree source is disabled.
 	t.RegisterEventProcessor(events.SchedProcessFork, t.procTreeForkRemoveArgs)
 
+	//
+	// DNS Cache Processors
+	//
+
+	if t.config.EnableDNSCache {
+		// TODO(nadav): Migrate to control plane signals?
```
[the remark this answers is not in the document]
NDStrahilevitz: > IMO network events should not be part of the signal plane.
Not all network events, I wanted a signal specifically for this one, so it won't take other event's space in the perf buffer if not traced for emission. This isn't as mission critical as the container signals for example, so it's not a must, which is why I left it as a question.

> Also, why "DNS Cache Processors". Planning others ? Which ones ?
Just sticking to how comments were ordered. I could make it a smaller title.

pkg/utils/set/set.go
```diff
@@ -0,0 +1,76 @@
...
+	s.mutex.RLock()
+	defer s.mutex.RUnlock()
+	_, ok := s.uniqueSet[item]
+	return ok
+}
+
+func (s *Set[T]) Items() []T {
+	s.mutex.RLock()
+	defer s.mutex.RUnlock()
+	return s.items
```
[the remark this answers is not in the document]
NDStrahilevitz: Good idea.

pkg/utils/set/set.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func (s *Set[T]) Has(item T) bool {
+	s.mutex.RLock()
+	defer s.mutex.RUnlock()
+	_, ok := s.uniqueSet[item]
+	return ok
+}
+
+func (s *Set[T]) Items() []T {
```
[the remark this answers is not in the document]
NDStrahilevitz: I guess I can add these : )

pkg/utils/set/set.go
```diff
@@ -0,0 +1,76 @@
...
+	for _, item := range items {
+		set.uniqueSet[item] = struct{}{}
+	}
+	for item := range set.uniqueSet {
+		set.items = append(set.items, item)
+	}
+	return &set
+}
+
+func (s *Set[T]) Empty() bool {
```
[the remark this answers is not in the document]
NDStrahilevitz: I opted to add a Clear method instead to make it distinct from emptying.

pkg/dnscache/dnscache.go
```diff
@@ -0,0 +1,204 @@
...
+
+	return addresses, nil
+}
+
+func (dc *DNSCache) addDnsCname(query string, answer string) error {
+	//
+	// Note: Do not use any locks here.
+	//       It is internal function.
+	//       Only the caller (from exposed function must use locking).
```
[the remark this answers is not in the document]
NDStrahilevitz: Leftover from copying, but I do prefer leaving just to be sure.

pkg/dnscache/dnscache.go
NDStrahilevitz: So about the two TTL types, since this is based on what the sponsors originally developed, I assume they are ok with only using an internal cache TTL. We could, in addition, develop an Expirable generic struct and make the set items of this type, WDYT? 

About the process distinction:
If this is something we want I believe we have two options I can think of:
1. Make the set entry a struct which includes the process which communicated to it (like storing a "foreign key" to the process tree
2. The process tree could hold a dns cache per process (doesn't sound right to me)

pkg/dnscache/dnscache.go
NDStrahilevitz: Because I didn't want to filter (again, as it's already filtered in the derivation) in the function wether it was a request or response.

NDStrahilevitz:
I want to add configuration for the cache LRU expiry time and max size before merging.

pkg/dnscache/dnscache.go
rafaeldtinoco: I believe this conversation thread was elucidated today in our talk. I wanted to make sure all types were covered when I asked this (and I thought that CNAME + A would come together in the response). This might or might not be true. And I think we already covered all the other cases today.

NDStrahilevitz: > (and I thought that CNAME + A would come together in the response)

From what I've gathered, it seems to be true for "A" requests, servers send the CNAME and then the A records corresponding to it, but it probably depends on the server. I think the current implementation should usually cover us, but I may want to further change how DNS records are linked to each other before merging.

rafaeldtinoco: Depends on the query as well. "set q=a" would send a response with all together.

NDStrahilevitz:
@rafaeldtinoco I've added an alternative implementation of the cache backend using a tree with indexed nodes (tell me if there is a better/more correct way to describe it). Please review and contrast with the original implementation, which i've also iterated upon. I believe the tree cache is more correct in how it caches results, but i'm not sure if it will be good enough performance wise on the querying side. Anyway, let's decide and i'll remove the irrelevant commits before merging.

pkg/dnscache/nodednscache.go
```diff
@@ -0,0 +1,132 @@
...
+
+	"github.com/aquasecurity/tracee/types/detect"
+	"github.com/aquasecurity/tracee/types/trace"
+)
+
+type NodeCache struct {
+	queryRoots   *expirable.LRU[string, *dnsQueryNode]
+	queryIndices map[string]*dnsQueryNode
+
+	results *expirable.LRU[string, nodeCacheQuery]
```
NDStrahilevitz, line 18: leftover

pkg/dnscache/nodednscache.go
```diff
@@ -0,0 +1,139 @@
...
+func NewNodeDNSCache(config Config) *NodeCache {
+	ttl := time.Second * time.Duration(config.TTL)
+	nc := &NodeCache{}
+	nc.queryRoots = expirable.NewLRU[string, *dnsQueryNode](config.CacheSize, nc.evictQueryRoot, ttl)
+	nc.queryIndices = make(map[string]*dnsQueryNode)
+	nc.lock = new(sync.RWMutex)
+	return nc
+}
+
+func (nc *NodeCache) Add(dns *trace.ProtoDNS) error {
```
NDStrahilevitz, line 30: @rafaeldtinoco this implementation will use `net_packet_dns` instead of the legacy event like you wanted for the original one.

pkg/cmd/flags/dnscache.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"github.com/aquasecurity/tracee/pkg/dnscache"
+)
+
+func dnsCacheHelp() string {
+	return `Select different options for the DNS cache.
+
+Example:
+  --dnscache enable  | enable with default values (see below). Using other options will override them, but is redundant.
+  --dnscache size=X  | will cache up to X dns and ip cross references in the LRU cache (default: 500).
```
rafaeldtinoco, line 16: 500 is a very small number, no ? A simple webpage load could solve more than 20 or 30 names depending on images coming from multiple sources (including ads etc). Check the example below, 54 and 82 dns request events in 1 second from a simpple browser session:

```console
$ sudo ./dist/tracee --install-path /tmp/tracee --cache cache-type=mem --cache mem-cache-size=512 --output option:sort-events --output json --output option:parse-arguments --events net_packet_dns_request | linecount 
0 lines/sec
0 lines/sec
6 lines/sec
10 lines/sec
0 lines/sec
2 lines/sec
0 lines/sec
6 lines/sec
7 lines/sec
29 lines/sec
54 lines/sec
82 lines/sec
10 lines/sec
16 lines/sec
1 lines/sec
7 lines/sec
14 lines/sec
3 lines/sec
11 lines/sec
0 lines/sec
18 lines/sec
0 lines/sec
13 lines/sec
17 lines/sec
4 lines/sec
51 lines/sec
36 lines/sec
15 lines/sec
23 lines/sec
2 lines/sec
10 lines/sec
16 lines/sec
5 lines/sec
17 lines/sec
0 lines/sec
10 lines/sec
0 lines/sec
0 lines/sec
0 lines/sec
```

pkg/cmd/flags/dnscache.go
```diff
@@ -0,0 +1,75 @@
...
+  --dnscache ttl=Y   | dns and ip entries will expire after Y seconds (default: 300 seconds).
```
rafaeldtinoco, line 17: default TTL is usually 3600 or 86400. Why such a short TTL ? My line of thoughts: if we keep the entry more time, and there are multiple IPs for the same entry, then its good to have all of the associated to the same entry. 300 secs is usually used in very dynamic envs when DNS owners want fast propagation, or during changes so they can make sure all is good before raising it, or related to DNS caching (which is not what we are doing, we are caching "information" about DNS, not DNS itself). No ?

pkg/utils/set/set.go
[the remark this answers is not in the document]
rafaeldtinoco: I dont think we should have unsafe methods. Full encapsulation only. And I think we should enforce that pattern from now on. One of our biggest problems for refactoring is/was lack of encapsulation. IF a caller needs the "internal" data, either register a callback pointer to a function, deal with a channel for readiness completion OR be part of the object. Is that ok for you ?

pkg/dnscache/events
```diff
@@ -0,0 +1,8 @@
...me":3399319006836545726,"processorId":4,"processId":2546510,"cgroupId":145296,"threadId":2546511,"parentProcessId":2546328,"hostProcessId":2546510,"hostThreadId":2546511,"hostParentProcessId":2546328,"userId":1000,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"isc-net-0000","executable":{"path":""},"hostName":"ip-172-31-75-17","containerId":"","container":{},"kubernetes":{},"eventId":"2006","eventName":"net_packet_dns","matchedPolicies":[""],"argsNum":5,"returnValue":0,"syscall":"sendmmsg","stackAddresses":[0],"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":458119841,"processEntityId":1413552547,"parentEntityId":3216664721,"args":[{"name":"src","type":"const char*","value":"127.0.0.1"},{"name":"dst","type":"const char*","value":"127.0.0.53"},{"name":"src_port","type":"u16","value":56900},{"name":"dst_port","type":"u16","value":53},{"name":"proto_dns","type":"trace.ProtoDNS","value":{"ID":20274,"QR":0,"opCode":"query","AA":0,"TC":0,"RD":1,"RA":0,"Z":0,"responseCode":"no error","QDCount":1,"ANCount":0,"NSCount":0,"ARCount":0,"questions":[{"name":"uol.com","type":"CNAME","class":"IN"}],"answers":[],"authorities":[],"additionals":[]}}]}
```
rafaeldtinoco, line 1: I believe this file was an accident.

pkg/dnscache/dnscache.go
```diff
@@ -0,0 +1,248 @@
...
+
+	ipList, ok := dc.dnsToRecords.Get(dns)
+	if !ok {
+		return nil, detect.ErrDataNotFound
+	}
+
+	return ipList.addresses.Items(), nil
+}
+
+// GetIPsForDNS returns IPs by passed DNS from the LRU cache.
```
rafaeldtinoco, line 108: nit: GetDNSRecords

pkg/dnscache/dnscache.go
```diff
@@ -0,0 +1,248 @@
...
+func (dc *DNSCache) GetIPsForDNS(dns string) ([]string, error) {
+	dc.rwLock.RLock()
+	defer dc.rwLock.RUnlock()
+
+	ipList, ok := dc.dnsToRecords.Get(dns)
+	if !ok {
+		return nil, detect.ErrDataNotFound
```
rafaeldtinoco, line 102: I dont think you reuse errors from another package (you're only importing that package because of those AND its weird to use another packages error, IMO).

pkg/dnscache/dnscache.go
```diff
@@ -0,0 +1,248 @@
...
+	//       Only the caller (from exposed function must use locking).
+
+	if records, ok := dc.dnsToRecords.Get(query); !ok {
+		dc.dnsToRecords.Add(query, dnsRecords{
+			addresses: set.New[string](),
+			alias:     answer,
+		})
+	} else {
+		// DNS CNAME should always come first in the set
+		records.alias = answer
```
rafaeldtinoco, line 166: I believe this is ok as CNAMEs need to be unique (can't have one domain name pointing to two CNAMEs). The CNAME change would mean the domain was reconfigured (And its okay for us to overwrite the value).

pkg/dnscache/dnscache.go
```diff
@@ -0,0 +1,248 @@
...
+	//       Only the caller (from exposed function must use locking).
+
+	if records, ok := dc.dnsToRecords.Get(query); !ok {
+		dc.dnsToRecords.Add(query, dnsRecords{
+			addresses: set.New[string](),
+			cname:     answer,
+		})
+	} else {
+		// DNS CNAME should always come first in the set
+		records.cname = answer
```
rafaeldtinoco, line 147: I believe this is ok as CNAMEs need to be unique (can't have one domain name pointing to two CNAMEs). The CNAME change would mean the domain was reconfigured (And its okay for us to overwrite the value).

pkg/cmd/flags/dnscache.go
NDStrahilevitz: These were the default values taken from the original internal implementation, and they went through no tuning, TBH completely guessed. Yes according to this test it should be much more... Perhaps even no capacity and only relying on expiry time? Though i'm worried that would easily cause high memory usage.

pkg/cmd/flags/dnscache.go
NDStrahilevitz: Well, the LRU shares expiry throughout it's entries. So entries can't have a unique TTL (otherwise I could take it from the answer). By this logic we should aim at the ceiling for "overall" TTL (so probably 86400 should be chosen), since if an entry had short TTL, the node would probably have to requery it anyway, thus overriding it in the cache.

pkg/dnscache/events
NDStrahilevitz: It's not, it's to demo the node based dns cache (see the test file's `init` function where it loads events from here).

NDStrahilevitz: Maybe it's better to move this to a string literal in the test file WDYT.

pkg/dnscache/dnscache.go
NDStrahilevitz: Yeah, there's a leaky abstraction here... I just didn't want to create an error value just so I can translate it again in the data source layer, but this isn't good practice either...

pkg/cmd/flags/dnscache.go
rafaeldtinoco: Yep, I like the expiration. I would do this... calculate the size per "average entry" (a mid sized name, with up to 5 entries, etc). What is the acceptable memory footprint per this feature, and how much names can we fit in ? THat will give us a better choice of default values.

pkg/dnscache/events
rafaeldtinoco: To be honest its not a bad idea but I find weird to have files flying in the source tree structure just to input tests, maybe we could name those in a certain pattern _test.json so we can identify them fast ? WDYT ?

pkg/dnscache/nodednscache.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+// addChildNodes builds new DNS nodes from a ProtoDNS object and attaches them to a parent node
+func (nc *NodeCache) addChildNodes(dns *trace.ProtoDNS, parent *dnsQueryNode) {
+	for _, answer := range dns.Answers {
+		node := makeNodeFromAnswer(parent, &answer)
+		if node != nil {
+			parent.next = append(parent.next, node)
+			nc.queryIndices[node.value] = node
+		}
```
rafaeldtinoco, line 115: In here, consider two different DNS queries to the same IP as an answer. Using the IP address as the index (since you're now creating a node from the answer) will cause the second query's answer to overwrite the first in nc.queryIndices, no ?

rafaeldtinoco:
I like the entire code but the `node` and `nodednscache` part. 

I don't like we are trying to optimize using multiple data structures, before even knowing this is needed, and mixing the index from queries and answers (raised a thought in the code). Particularly worried about all possible queries/answers combinations.

> I also think the code needs better documentation (about your intents and example of strings whenever they are being considered, small texts explaining reasonings, etc).

From what I see there are 2 possible initial ways to do this, based on the following **requirements** (which I believe we are on the same page):

1) Given a DNS name, to get the resulting IN A entry (after all aliases are solved).
2) Given an IP, to pick the initial query name that resulted to it.

From the way you **implemented**, I believe you tried to address lots of valid concerns:

1) Being able to construct the full resolution path (through node links)
2) Being memory efficient by having LRU of nodes and time evictions
3) Allow fast queries through keys
4) Memory deduplication through tree structure, etc.

I think that we should have a complete and **simpler approach** before trying all (or any) optimizations. I have created pseudo code mixed with garbage mixed with go at the following gist:

- https://gist.github.com/rafaeldtinoco/0e1a77963c8284c363af08ca2c7b9292

NOTE: Im not suggesting you use my code (specially because it does not work). I wanted to think so I started playing a bit, that's all. It might give you ideas if you accept my feedback. I don't think we should worry so much about access time if the LRU entry is already hashed (and even in recursive walks, like my example, it might fast enough).

So, at the end, Im suggesting that you take an simpler approach until proved something else is needed.

With all that said, I **like very much everything else** to be honest. Its like this part "does not fit yet" the rest, which was very good IMO.

pkg/utils/set/set.go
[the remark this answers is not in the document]
NDStrahilevitz: Not sure why it doesn't reflect here, in the current version `Items` returns a copy and the original was renamed `ItemsUnsafe`. I think that as reviewers we know to be extra careful if we see a code section using the Unsafe methods (which are unused in this code). IMO our issue so far has not been only lack of encapsulation but lack of safe encapsulation and unawareness it was unsafe. The code here is neither of these things, it is correctly encapsulated and unsafety is explicitly mentioned. 
I can remove it anyway, but I think if we use the Set in more critical sections it could be useful for reads. Using a callback instead does not guarantee any read safety.

pkg/cmd/flags/dnscache.go
NDStrahilevitz: I think that if I rework the node implementation (which I will explain why I prefer to do so below), I will rework expiry to be an internal mechanism instead of relying on a global one provided by the LRU. So we can have per item expiry based on actual TTL.

pkg/cmd/flags/dnscache.go
NDStrahilevitz: See above.

pkg/dnscache/events
NDStrahilevitz: Yep, sounds good.

pkg/dnscache/nodednscache.go
NDStrahilevitz: Hmmm, so we can do two options here:
1. make queryIndices a map of value to a list of valid nodes, and keep the graph as a tree
2. give up on a tree structure and have the option for multiple parents (which I think would basically make it just a graph).

I slightly prefer option 2.

NDStrahilevitz:
@rafaeldtinoco I have the intuition that the simpler approach as it is will accidentally break for unexpected queries. 
Also we have in a previous internal conversation assessed that having access to the query history, specifically from an IP up, is actually just as much of a requirement (both for the future dns enrichment in connect events, and for the dnscache data source results).
In addition, the memory efficiency with the current LRU implementation could easily explode (see the amount of queries you generated), especially if we only rely on expiry, so I think that is also a valid concern.

I would like to continue anyway with this approach, improving its documentation as you mentioned.

rafaeldtinoco:
> I would like to continue anyway with this approach, improving its documentation as you mentioned.

OK, I can accept that just because you're trying hard to optimize things, but please expand tests then. 

You can keep tracee opened for some hours and compress the events yaml file using gzip and do a gzip reader for the test. I would like to have many different queries tested. 

I would also go for a "needle in the haystack" test, trying to find a specific complex case query (multiple aliases, multiples IPs and even a reverse lookup for the same IPs) among everything that was cached. 

Also, anytime something is overwritten I would like it to be logged in debug mode.

I would feel more confident with that approach, and possibly others you have in mind.

pkg/dnscache/nodednscache.go
rafaeldtinoco: I believe we should not pick PTR resolutions now. Mainly because a reverse IP could be something totally different than any other query to that name. I think you don't have to be authoritative to what you respond for PTR resolutions.

About having a graph, in order to extract Names or IPs from any data structure we might have, I believe we could have a Bidirectional Graph (it would have duplicates when compared to DAGs BUT it would allow going back and forth from any point).

image

This example is simple and looks like a double linked list, but if we make things more complicated:

image

Also, I would suggest (even require) that you use a good/known/supported package to create such data structures (https://github.com/gonum/gonum/tree/master for example).

pkg/cmd/flags/dnscache.go
NDStrahilevitz: I believe this is addressed in the current implementation.

pkg/dnscache/events
NDStrahilevitz: done.

pkg/dnscache/dnscache.go
NDStrahilevitz: Fixed.

pkg/utils/set/set.go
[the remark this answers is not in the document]
NDStrahilevitz: I've found a usage for the unsafe versions in the internal methods of the dnscache. I can see the criticism of this being a premature optimization, but i do think it's better as is. The name indicates that any code change near it should be done with care.

pkg/utils/set/set.go
```diff
@@ -0,0 +1,118 @@
+package set
+
+import (
+	"fmt"
+)
+
+// Set is a generic thread-safe data structure representing a unique collection of items
```
rafaeldtinoco, line 7: Ok, I see your point here, specially after the creation of threadsafe.go. So, we shouldn't call this Set thread-safe. It is actually a thread unsafe type considering it stores the set in a slice, which is thread-unsafe (considering you have methods iterating the slice, or copying it, that might happen in parallel). Despite the semantics of what we call thread-safety, or any discussion of that regareds, what if we just call this "Set" (saying its not thread safe) and using the threadsafe set you created (and even recommending it is always used). Then I would remove the "ItemsUnsafe" method from this also (everything is technically unsafe). Then, if you want to use this unsafe set and synchronize access to it through an external lock/mutex, up to the user (just like any other thread-unsafe type).

pkg/utils/set/threadsafe.go
```diff
@@ -0,0 +1,88 @@
+package set
+
+import (
+	"sync"
+)
+
+type ThreadSafeSet[T, H comparable] struct {
```
rafaeldtinoco, line 7: Yep, this is good and should the the "default" Set. I would even call it Set and the other something else, but that is not important.

pkg/utils/set/set.go
```diff
@@ -0,0 +1,118 @@
...
+		set.uniqueSet[item] = item
+	}
+	for item := range set.uniqueSet {
+		set.items = append(set.items, item)
+	}
+	return &set
+}
+
+// New builds a new set object
+func NewWithHash[T comparable, HashType comparable](hashFunc func(T) HashType, items ...T) *Set[T, HashType] {
```
rafaeldtinoco, line 33: Looks like too much for this type (as its already a set). But since you are already implementing the ThreadSafeSet type, if we created a HashSet type we would also have to create a ThreadSafeHashSet. IMO it would be better to distinguish both, but I dont think it really matters (since you have a constructor that creates a default basic hash).

pkg/dnscache/dnscache_test.go
```diff
@@ -0,0 +1,106 @@
...
+	assert.ElementsMatch(t, res.DNSResults(), []string{"dailymaverick.co.za"})
+	assert.Equal(t, res.IPResults(), []string{"2606:4700:10::6816:1f"})
+
+	// test reverse ip query
+	res, err = dnsCache.Get("231.92.80.190.in-addr.arpa")
+	require.NoError(t, err)
+	assert.Equal(t, res.DNSResults(), []string{"231-92-pool.dsl.gol.net.gy"})
+	assert.Equal(t, res.IPResults(), []string{"231.92.80.190"})
+
+	// possibly add more tests?
```
rafaeldtinoco, line 105: I would add a very small LRU cache (so evictions happen very frequently) and stress test paralelization. I have doubts if you won't be able to cause races in the TryLock() call I commented about.

rafaeldtinoco (commented):
Leaving comments for us to talk. We're getting there...

pkg/utils/set/set.go
NDStrahilevitz: I'd rather keep the `ItemsUnsafe` method just for the usefulness of a readonly non copying method. Maybe call it `ItemsMutable`?

pkg/dnscache/dnscache.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+}
+
+// evictQueryRoot is the eviction function for the query roots LRU
+func (nc *DNSCache) evictQueryRoot(addr string, node *dnsNode) {
+	// eviction may sometimes occur in the context of an Add transaction.
+	// In these cases locking is unnecessary, since the evict is done in
+	// sequence with adding.
+	locked := nc.lock.TryLock()
+	if locked {
```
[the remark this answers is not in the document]
NDStrahilevitz: So this is unlikely because essentially two cases are possible in the current code (and only one can actually be triggered - the 2nd one):
1. A manual eviction is started, in this case the eviction acquires the lock before any other tx or waits for it. We don't have code manually calling eviction right now.
2. This happens in the context of adding an element to the LRU, this can only happen in the context of the tx mutex already being locked, so the TryLock fails and skips the defer.

pkg/utils/set/threadsafe.go
NDStrahilevitz: In hashicorp's LRU they call the non thread safe version (wrapped by the threadsafe one) `SimpleLRU` so we can do the same semantics `SimpleSet` and `Set`.

pkg/dnscache/dnscache_test.go
NDStrahilevitz: Already explained why I don't think the deadlock is likely (actually the TryLock resolved an initial deadlock bug), but no harm in adding, and if it proves me wrong all the better.

pkg/utils/set/set.go
NDStrahilevitz: The comment itself I just forgot to change.

pkg/dnscache/dnscache_test.go
NDStrahilevitz: Ok i've added such a stress test, it helped finding other incidental bugs, but there's no other deadlock.

pkg/dnscache/nodednscache.go
rafaeldtinoco: BTW, for the record, we gave up on this idea as @NDStrahilevitz saw things would get more complicated and documenting/re-structuring his initial data structures were enough.

PR #3681, opened 2023-11-12
refactor(engine): feed engine with signatures events

### LINKED ISSUE #3680: Signatures feed into the early pipeline stages
Currently, signatures are both events and not events.
They are converted to events, but are only inserted into the "sink" stage of the events pipeline.
This makes little sense, as this disables the option to use signatures in signatures (though they are an event), and even maybe processing them inside Tracee for other uses in the future.
I think the signatures should at least be fed to the input of the signatures engine, instead of the "sink" stage.

AlonZivony:
Feed the signatures engine with its own output instead of feeding it to the output channel.
This allows signatures to use other signatures as their selector.

cfa2e5dae **refactor(engine): feed engine with signatures events**

fix #3680

### DIFF
--- a/pkg/ebpf/signature_engine.go
+++ b/pkg/ebpf/signature_engine.go
@@ -18,8 +18,9 @@ func (t *Tracee) engineEvents(ctx context.Context, in <-chan *trace.Event) (<-ch
 	out := make(chan *trace.Event)
 	errc := make(chan error, 1)
 
-	engineOutput := make(chan detect.Finding, 100)
-	engineInput := make(chan protocol.Event)
+	engineOutput := make(chan detect.Finding, 10000)
+	engineInput := make(chan protocol.Event, 10000)
+	engineOutputEvents := make(chan *trace.Event, 10000)
 	source := engine.EventSources{Tracee: engineInput}
 
 	// Prepare built in data sources
@@ -45,6 +46,35 @@ func (t *Tracee) engineEvents(ctx context.Context, in <-chan *trace.Event) (<-ch
 
 	go t.sigEngine.Start(ctx)
 
+	// Create a function for feeding the engine with an event
+	feedFunc := func(event *trace.Event) {
+		if event == nil {
+			return // might happen during initialization (ctrl+c seg faults)
+		}
+
+		id := events.ID(event.EventID)
+
+		// if the event is marked as submit, we pass it to the engine
+		if t.eventsState[id].Submit > 0 {
+			err := t.parseArguments(event)
+			if err != nil {
+				t.handleError(err)
+				return
+			}
+
+			// Get a copy of our event before sending it down the pipeline.
+			// This is needed because a later modification of the event (in
+			// particular of the matched policies) can affect engine stage.
+			eventCopy := *event
+			// pass the event to the sink stage, if the event is also marked as emit
+			// it will be sent to print by the sink stage
+			out <- event
+
+			// send the event to the rule event
+			engineInput <- eventCopy.ToProtocol()
+		}
+	}
+
 	// TODO: in the upcoming releases, the rule engine should be changed to receive trace.Event,
 	// and return a trace.Event, which should remove the necessity of converting trace.Event to protocol.Event,
 	// and converting detect.Finding into trace.Event
@@ -58,31 +88,9 @@ func (t *Tracee) engineEvents(ctx context.Context, in <-chan *trace.Event) (<-ch
 		for {
 			select {
 			case event := <-in:
-				if event == nil {
-					continue // might happen during initialization (ctrl+c seg faults)
-				}
-
-				id := events.ID(event.EventID)
-
-				// if the event is marked as submit, we pass it to the engine
-				if t.eventsState[id].Submit > 0 {
-					err := t.parseArguments(event)
-					if err != nil {
-						t.handleError(err)
-						continue
-					}
-
-					// Get a copy of our event before sending it down the pipeline.
-					// This is needed because a later modification of the event (in
-					// particular of the matched policies) can affect engine stage.
-					eventCopy := *event
-					// pass the event to the sink stage, if the event is also marked as emit
-					// it will be sent to print by the sink stage
-					out <- event
-
-					// send the event to the rule event
-					engineInput <- eventCopy.ToProtocol()
-				}
+				feedFunc(event)
+			case event := <-engineOutputEvents:
+				feedFunc(event)
 			case <-ctx.Done():
 				return
 			}
@@ -108,7 +116,7 @@ func (t *Tracee) engineEvents(ctx context.Context, in <-chan *trace.Event) (<-ch
 					continue
 				}
 
-				out <- event
+				engineOutputEvents <- event
 			case <-ctx.Done():
 				return
 			}
--- a/tests/e2e-inst-signatures/e2e-signature_derivation.go
+++ b/tests/e2e-inst-signatures/e2e-signature_derivation.go
@@ -1,0 +1,61 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/aquasecurity/tracee/types/detect"
+	"github.com/aquasecurity/tracee/types/protocol"
+	"github.com/aquasecurity/tracee/types/trace"
+)
+
+type e2eSignatureDerivation struct {
+	cb detect.SignatureHandler
+}
+
+func (sig *e2eSignatureDerivation) Init(ctx detect.SignatureContext) error {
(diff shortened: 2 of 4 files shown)


AlonZivony:
I am debating if we should maybe change the `Source` of the selector to be "engine" or something similar.
This will allow the to easily distinguish between signatures and normal events in other signatures, and allow a signature to use all signature events in the selector (like we do in the policy with tags).

rafaeldtinoco (changes requested):
I like the idea of deriving the signatures (I believe the proctree example had a good "usecase" or, at least, an idea that could be used in that regards).

I think the PR needs improvements:

1. You need to better document existing and new logic and functions (please watch out for formatting/syntax, capitalize sentences, keep good line height, etc).

2. I'm afraid of the loop reintroduction. That could lead us to infinite loops if the mask isn't cleared by the signature engine in some occasion. Could you at least document (in a comment ?) what condition makes it safe ? Maybe we should have a counter for the event (can't rely in the event pointer in a LRU map because every time the event comes it has a new address since it was io.copied()).

3. I think we need to fix the "Args" part. When reintroducing the event, the `FindingToEvent->NewEvent->getArguments` logic nests the `Args->TriggeredBy` over and over and over:

```json
{
  "DifferencesInArgs": {
    "Event1": {
      "Args": [
        {
          "Name": "triggeredBy",
          "Value": {
            "args": [
              {
                "Name": "file_path",
                "Type": "const char*",
                "Value": "/home/rafaeldtinoco/work/projects/tracee/file_modification.txt"
              },
              {
                "Name": "dev",
                "Type": "dev_t",
                "Value": 271581188
              },
              {
                "Name": "inode",
                "Type": "unsigned long",
                "Value": 17214308
              },
              {
                "Name": "old_ctime",
                "Type": "unsigned long",
                "Value": 1699921967398957690
              },
              {
                "Name": "new_ctime",
                "Type": "unsigned long",
                "Value": 1699921967402291023
              }
            ],
            "id": 774,
            "name": "file_modification",
            "returnValue": 0
          }
        }
      ]
    },
    "Event2": {
      "Args": [
        {
          "Name": "triggeredBy",
          "Value": {
            "args": [
              {
                "Name": "triggeredBy",
                "Value": {
                  "args": [
                    {
                      "Name": "file_path",
                      "Type": "const char*",
                      "Value": "/home/rafaeldtinoco/work/projects/tracee/file_modification.txt"
                    },
                    {
                      "Name": "dev",
                      "Type": "dev_t",
                      "Value": 271581188
                    },
                    {
                      "Name": "inode",
                      "Type": "unsigned long",
                      "Value": 17214308
                    },
                    {
                      "Name": "old_ctime",
                      "Type": "unsigned long",
                      "Value": 1699921967398957690
                    },
                    {
                      "Name": "new_ctime",
                      "Type": "unsigned long",
                      "Value": 1699921967402291023
                    }
                  ],
                  "id": 774,
                  "name": "file_modification",
                  "returnValue": 0
                }
              }
            ],
            "id": 6001,
            "name": "FILE_MODIFICATION",
            "returnValue": 0
          }
        }
      ]
    }
  }
}
```

AlonZivony:
> 1. You need to better document existing and new logic and functions (please watch out for formatting/syntax, capitalize sentences, keep good line height, etc).

Sure, though I feel it is already well documented in the code. You want in the documentation man pages?

> 2. I'm afraid of the loop reintroduction. That could lead us to infinite loops if the mask isn't cleared by the signature engine in some occasion. Could you at least document (in a comment ?) what condition makes it safe ? Maybe we should have a counter for the event (can't rely in the event pointer in a LRU map because every time the event comes it has a new address since it was io.copied()).

I don't really see a reason to be concerned here.
The only case this might be problematic is if the signatures will create a `Finding` which is of the same event type of the input event, which doesn't makes sense in the current way signatures work.
If a user creates two signatures which are derived from one another, then this is just a very bad logic from his part and its not our responsebillity (like endless recursion is not prevented by the compiler).
However, if you think this is necessary, I will add a warning in the documentation about creating mutual derivation of signatures or create a signature that its `Finding` is of a normal event type.

> 3. I think we need to fix the "Args" part. When reintroducing the event, the `FindingToEvent->NewEvent->getArguments` logic nests the `Args->TriggeredBy` over and over and over. This is actually a good question, @NDStrahilevitz, you asked me about multiple derivations from a single event. Lets assume we will allow that to happen (for both events and signature events), how should "Args" and "triggeredBy" look like in that case ?

I don't see a problem with nested `derivedBy` argument, but I agree this is open to discussion. @NDStrahilevitz tell us what you think

pkg/ebpf/signature_engine.go
```diff
@@ -20,6 +20,7 @@ func (t *Tracee) engineEvents(ctx context.Context, in <-chan *trace.Event) (<-ch
 
 	engineOutput := make(chan detect.Finding, 100)
 	engineInput := make(chan protocol.Event)
+	engineOutputEvents := make(chan *trace.Event, 100)
```
NDStrahilevitz, line 23: Maybe both this and the above should be a bigger channel?

pkg/ebpf/signature_engine.go
```diff
@@ -20,6 +20,7 @@ func (t *Tracee) engineEvents(ctx context.Context, in <-chan *trace.Event) (<-ch
 
 	engineOutput := make(chan detect.Finding, 100)
 	engineInput := make(chan protocol.Event)
```
NDStrahilevitz, line 22: The engine input should for sure, as we usually use 10000 as the channel buffers in the pipeline.

pkg/ebpf/signature_engine.go
```diff
@@ -45,6 +46,35 @@ func (t *Tracee) engineEvents(ctx context.Context, in <-chan *trace.Event) (<-ch
 
 	go t.sigEngine.Start(ctx)
 
+	// Create a function for feeding the engine with an event
+	feedFunc := func(event *trace.Event) {
```
NDStrahilevitz, line 50: Nice refactor. Since this file is already made in particular for the engine related stuff, do you mind making this a top level function in the file? See if it looks better, if it doesn't no need.

NDStrahilevitz (commented):
Not -1ing because @rafaeldtinoco  raised the concerns relevant for it already.

I'll answer to the points:
1. Documentation - 100% required for this kind of change. Inside the code it didn't seem bad at all to me, but I think that the actual docs should be updated on this.
2. I tend to agree with @AlonZivony  on the risk of an infinite loop, but just to be sure we should test this by loading two signatures, one regular one derived, add a print in the engine post selector stage, and trigger the initial signature. Then we can see if an infinite loop is generated anyway.
3. While I think one nested triggeredBy is ok, i'm concerned with any more than that. I'm not sure it's worthwhile for us to limit it though, so it should be left to the user's concern when writing signatures, and documented as a potential "risk" factor of the feature.

pkg/ebpf/signature_engine.go
AlonZivony: Hmm I don't think it will look better as it uses the channels of this stage.
So unless we pass to it all channels each time, or create a function that return this function (which looks the same to me like the current state or even worst) I don't think we can improve the function here.

AlonZivony:
@rafaeldtinoco @NDStrahilevitz
After going over the current documentation, it seems that there is no documentation page that go into the details of creating a signature.
It means that the fact that you can't use signatures in the events selector until now was not documented anywhere.
Unless you want me to write such detailed documentation myself (which either way shouldn't be part of this PR) then I don't think further documentation is required.
Feel free to reply to this if I missed something or you have an idea of a documentation you do want me to write.

rafaeldtinoco:
> @rafaeldtinoco @NDStrahilevitz After going over the current documentation, it seems that there is no documentation page that go into the details of creating a signature. 

We have:

https://aquasecurity.github.io/tracee/dev/docs/events/custom/overview/
https://aquasecurity.github.io/tracee/dev/docs/events/custom/golang/

but they are very small. The fact that golang signatures aren't good end-user experience is definitely what made us not to extend them (as we had the rego x go-cel x golang signatures discussions).

> It means that the fact that you can't use signatures in the events selector until now was not documented anywhere. 

Sure, and this was only "a thing" after the "everything as event" feature was done (very recently) anyway.

> Unless you want me to write such detailed documentation myself (which either way shouldn't be part of this PR) then I don't think further documentation is required.

Nah, its alright from myside.

rafaeldtinoco (approved):
LGTM. Nadav, Alon and I discussed this offline and agreed that the recursive triggeredBy is ok for now, all other items were discussed and already taken care.

PR #3685, opened 2023-11-13
Add Parquet output Issue #3682

### LINKED ISSUE #3682: Add Parquet output
Parquet is a columnar storage file format that is commonly used in the context of big data processing frameworks, such as Apache Spark and Apache Hive. The format is designed to be highly efficient for both storage and processing, especially in scenarios involving large-scale data analytics. Here are some reasons why Parquet is often used:

Columnar Storage: Parquet stores data in a columnar format, which means that values from the same column are stored together. This storage layout is more efficient for analytics queries that often involve reading specific columns rather than entire rows.

Compression: Parquet supports various compression algorithms, enabling efficient use of storage space. It reduces the amount of disk space needed to store large datasets, making it cost-effective.

Predicate Pushdown: Some query engines, like Apache Spark, can take advantage of predicate pushdown with Parquet. This means that certain filter conditions can be pushed down to the storage layer, minimizing the amount of data that needs to be read during query execution.

Schema Evolution: Parquet supports schema evolution, allowing you to evolve your data schema over time without requiring modifications to existing data or affecting backward compatibility.

Compatibility with Big Data Ecosystem: Parquet is widely used in the big data ecosystem, and many big data processing frameworks have built-in support for reading and writing Parquet files. This makes it easier to integrate Parquet with existing data processing workflows.

Performance: Due to its columnar storage and other optimizations, Parquet can offer improved performance for analytics queries, especially when dealing with large datasets.

When working with large-scale data analytics, Parquet can be a suitable choice for storing and processing your data efficiently. It provides benefits in terms of storage space, query performance, and compatibility with popular big data tools and frameworks. However, the choice of file format depends on your specific use case and the tools you are using in your data processing pipeline.

shani-gold:
Adding parquet as a format output
#https://github.com/aquasecurity/tracee/issues/3682

### 2. Explain how to test it
sudo ./dist/tracee --output parquet:test.parquet --events proc,syscalls,sched_switch,sched_process_fork,network_events


rafaeldtinoco:
Why can't we adapt the marshalling on the existing event type (and internal types) ?

```go
type Event struct {
    // Existing fields with added Parquet tags
    Timestamp             int          `json:"timestamp" parquet:"name=timestamp, type=INT64"`
    // ... other int fields with similar tags
    CgroupID              uint         `json:"cgroupId" parquet:"name=cgroupId, type=INT64"`
    // ... other fields
    ProcessName           string       `json:"processName" parquet:"name=processName, type=BYTE_ARRAY, convertedtype=UTF8"`
    // ... other string fields with similar tags
    Container             Container    `json:"container,omitempty" parquet:"name=container"`
    Kubernetes            Kubernetes   `json:"kubernetes,omitempty" parquet:"name=kubernetes"`
    // ... other struct fields with similar tags
    MatchedPoliciesKernel uint64       `json:"-" parquet:"name=matchedPoliciesKernel, type=INT64"`
    MatchedPoliciesUser   uint64       `json:"-" parquet:"name=matchedPoliciesUser, type=INT64"`
    MatchedPolicies       []string     `json:"matchedPolicies,omitempty" parquet:"name=matchedPolicies, type=MAP, convertedtype=LIST, valuetype=BYTE_ARRAY, valueconvertedtype=UTF8"`
    StackAddresses        []uint64     `json:"stackAddresses" parquet:"name=stackAddresses, type=MAP, convertedtype=LIST, valuetype=INT64"`
    // ... other fields
    Metadata              *Metadata    `json:"metadata,omitempty" parquet:"name=metadata"`
}
```
```go
type Container struct {
    ID          string `json:"id,omitempty" parquet:"name=id, type=BYTE_ARRAY, convertedtype=UTF8"`
    // ... other fields with similar tags
}
```
```go
type Kubernetes struct {
    PodName      string `json:"podName,omitempty" parquet:"name=podName, type=BYTE_ARRAY, convertedtype=UTF8"`
    // ... other fields with similar tags
}
```
```go
type Metadata struct {
    Version     string `parquet:"name=version, type=BYTE_ARRAY, convertedtype=UTF8"`
    Description string `parquet:"name=description, type=BYTE_ARRAY, convertedtype=UTF8"`
    Tags        []string `parquet:"name=tags, type=MAP, convertedtype=LIST, valuetype=BYTE_ARRAY, valueconvertedtype=UTF8"`
    Properties  map[string]interface{} // Requires custom serialization, e.g., as a JSON string
}
```

Instead of having new ones ? Which type made you try to duplicate the event type ? Why ? Were you worried about a particular type serialization ? 

Another "problem" is that we're changing the tracee event to a new structure (using protobuf). Check `pkg/server/grpc/tracee.go`->`convertTraceeEventToProto`. Basically you can expect the tracee event format to be what protobuf is defining at: `./api/v1beta1/event.pb`.

I believe that, if we use the same existing types, and rely on custom marshaling/unmarshalling logic only, at least for now, then the eventual migration to the protobuf types will be smoother on our side (and merging this change would be easier).

rafaeldtinoco (commented):
Please sign the contributor agreement when possible.

pkg/cmd/printer/printer_parquet.go
```diff
@@ -0,0 +1,176 @@
...
+	ContextFlags        ParquetContextFlags `parquet:"name=contextFlags"`
+	ThreadEntityId      int32               `parquet:"name=threadEntityId, type=INT32, convertedtype=UINT_32"`
+	ProcessEntityId     int32               `parquet:"name=processEntityId, type=INT32, convertedtype=UINT_32"`
+	ParentEntityId      int32               `parquet:"name=parentEntityId, type=INT32, convertedtype=UINT_32"`
+	Args                string              `parquet:"name=args, type=BYTE_ARRAY, convertedtype=JSON"`
+}
+
+/*
+The following structs are not necessary: ParquetContainer, ParquetKubernetes, ParquetContextFlags, ParquetFile.
+
```
rafaeldtinoco, line 54: Yep, check my comment for the PR. I think we should try to have only the "really needed types", serialize with a custom marshaller whatever is needed and can't be automatically type-converted, and go from there (at least until we have protobufs fully done, having the event structure way more stable than it is now (soon).

shani-gold: I couldn't add the parquet tags to existing types (all the inner structs), since they are on a different package:
github.com/aquasecurity/tracee/types v0.0.0-20231013014739-b32a168ee6a8

Since the Event struct contains types which are not supported by parquet, I can't use it.
For example I can add 
`CgroupID uint json:"cgroupId" parquet:"name=cgroupId, type=INT64, convertedtype=UINT_64"`
But I get `error reflect: call of reflect.Value.Int on uint32 Value`

And I couldn't find a way to implement a custom marshal function.
The marshal function doesn't support it:
https://github.com/xitongsys/parquet-go/blob/8ca067b2bd324788a77bf61d4e1ef9a5f8b4b1d2/marshal/marshal.go#L293C16-L293C16

It calls directly to the ParquetStruct Marshal function

rafaeldtinoco: Ok so here is the thing, I played around a bit with parquet, because it was new to me and I wanted to understand better how it worked. 

It ended up being good and bad. Good that I played and bad related to how this `logical types` of parquet cast unsigned values (and also due to the issue https://github.com/aquasecurity/tracee/issues/3690 just discovered).

I was going to suggest you using json schema to parquet (https://gist.github.com/rafaeldtinoco/fbe58a020fd0d73ea37ce6980520f6ea) so we didnt have to maintain type duplicates. You would complain back to me saying tracee event type -> json -> parquet could add overhead, I would request you to check if that was too bad.

During this process, I discovered https://github.com/aquasecurity/tracee/issues/3690. It basically means that many other fields from the tracee event struct should be unsigned integers (I would say from 4 to 8 fields should have their types changed). Not sure how we missed that, but during the decode phase (from eBPF to our internal types) we converted them to Ints and never checked for overflows.

Good thing that parquet does not support unsigned values (actually it's a terrible thing :P) so I detected the overflow when generating the parquet file, I believe.

Ok, with all that said, and thinking about the near future, we're moving away from these internal types to the new protobuf types you will find in `github.com/aquasec/tracee/api` (specially in `event.pb`.

The protobuf types are declared correctly (uint32s, uints64) but it might impose a very hard time for you IF you rely in a parallel Parquet type to each type automatically generated from the protobuf (and actually I don't think we want parallel types just for marshalling logic).

That explains why I was going after the JSON Schema (to give you an alternative). Why ? 

1. I believe it might be easier for you to convert types from json Integers (than to worry about golang types coming from internal structures) to parquet types.

2. Its likely that tracee will eventually move into a grpc model where the engine will be a grpc server providing streams. Those streams will come serialized with protobuf types but we could also have an already marshalled version. You can see a very simple example at https://github.com/josedonizetti/tracee-grpc-client/blob/main/examples/stream_events/main.go for now.

3. Of course you can work through parquet marshalling and our protobuf types, I just dont want to commit to accept your printer now (when we fix the wrong field types), having duplicate types, and then discover your printer wont work with our near future protobuf work (and either remove your printer or request you to change and be blocked until you do).

Does it make sense for you ?

Suggestion of course of action:

- We fix the event struct types
- you rely in json schema for the printer (similar to my gist)
- you investigate the protobuf types we have for further actions
- by relying in the json schema you will probably remove headaches from protobuf (and we could have a stream of json marshaled types when grpc streams are finished). 

With all that, Im trying to give you a "roadmap" if you stick to tracee as your "engine" (for the project of yours). 

I hope it helps. No hard commitments on our side, but for sure a best effort trying to help.

rafaeldtinoco:
@shani-gold Left you some comments about possible outcomes. Feel free to suggest something else if you have in mind and, please, sign the contributor agreement if you are willing to have this merged. Thank you.

rafaeldtinoco:
@shani-gold ping ?

geyslan:

yanivagman:
Closing due to inactivity

PR #3687, opened 2023-11-13
chore(container): same default events as k8s deployment

rafaeldtinoco:
- all signatures enabled by default
- container creation and removal events only
- json output just like k8s deployment

anything else only if explicitly asked by user.


rafaeldtinoco:
I have deployed: https://github.com/aquasecurity/tracee/actions/runs/6853978375 after this was merged. A test on the dev image generated by that workflow will be good to know if all is good. (Of course already tested locally).

PR #3691, opened 2023-11-15
Feature/proctree query time

### LINKED ISSUE #3700: Process Tree queries are missing query time
The queries of the Process Tree are giving information according to a query time.
This means that the query time is very relevant in the context of the information.
For example, given a specific process information, you might want to query the matching task information (aka thread information). However, to get the info relevant to the given process info, you must query the exact same time. This is very hard to do in a standalone functions without the timestamp integration into the information given. If this info will move between different code components, we can't expect the use to couple them together himself, as it is part of the info.
This problem is even more severe with the lineage query, which gives each ancestor in a different time (according to fork time of its child).
I think that some sort of coupling of the information with the query time is required.

AlonZivony:

7fa990fb7 **feat(proctree): fill timestamp of proctree info objects**

Fix #3700

### 2. Explain how to test it

e2e test

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/aquasecurity/libbpfgo v0.5.0-libbpf-1.2
 	github.com/aquasecurity/libbpfgo/helpers v0.4.6-0.20231123142329-37c4b843a539
 	github.com/aquasecurity/tracee/api v0.0.0-20231013014739-b32a168ee6a8
-	github.com/aquasecurity/tracee/types v0.0.0-20231123143520-9a6b89efc320
+	github.com/aquasecurity/tracee/types v0.0.0-20231128135314-cfe4d6426ccc
 	github.com/containerd/containerd v1.7.0
 	github.com/docker/docker v24.0.7+incompatible
 	github.com/golang/protobuf v1.5.3
--- a/tests/e2e-inst-signatures/e2e-proctree_data_source.go
+++ b/tests/e2e-inst-signatures/e2e-proctree_data_source.go
@@ -23,26 +23,18 @@ const (
 
 type e2eProcessTreeDataSource struct {
 	cb            detect.SignatureHandler
-	processTreeDS detect.DataSource
+	processTreeDS *helpers.ProcessTreeDS
 }
 
 // Init is called once when the signature is loaded.
 func (sig *e2eProcessTreeDataSource) Init(ctx detect.SignatureContext) error {
 	sig.cb = ctx.Callback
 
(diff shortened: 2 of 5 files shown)


AlonZivony:
@rafaeldtinoco WDYT?

go.mod
```diff
@@ -181,4 +181,5 @@ require (
 	kernel.org/pub/linux/libs/security/libcap/psx v1.2.68 // indirect
 )
 
+replace github.com/aquasecurity/tracee/types => ./types
```
rafaeldtinoco, line 184: temporary, needs a PR type change (just marking here not to forget).

tests/e2e-inst-signatures/e2e-proctree_data_source.go
```diff
@@ -161,6 +162,9 @@ func (sig *e2eProcessTreeDataSource) checkThread(eventObj *trace.Event) error {
 	if threadInfo.Pid != eventObj.HostProcessID {
 		return fmt.Errorf(debug("no match for pid"))
 	}
+	if threadInfo.Timestamp != queryTime {
+		return fmt.Errorf(debug("no match for info timestamp"))
+	}
```
rafaeldtinoco, line 167: Code is correct. Logic is correct, etc. But, why ? The issue does not describe the "why" very well.

If you query for a process, or a thread, or a tree, at a specific point in time, you will get the nearest changelog timestamp item. Why there is a need to know the exact point in time ? (Im not saying I'm against). What would be the benefit ?

rafaeldtinoco (changes requested):
LGTM. Needs a better reasoning and a PR changing the types before this can be merged. Im putting a -1 for now until both are satisfied. Request my review in the types change, please. I'll merge and you can rebase this and change types dependency to that commit.

AlonZivony:
@rafaeldtinoco I probably need your opinion on the following question:
Should the query time be part of the info objects, or be part of an enveloping type to the info objects?
In my opinion it is easier to manage as part of the info object, but I am not sure that the timestamp is an info of the process in a way that justify this coupling.
We can create a generic `TimeRelevantInfo` type, that has information and timestamp in it, and return it instead of the process and thread info objects.
WDYT?

rafaeldtinoco:
> We can create a generic `TimeRelevantInfo` type, that has information and timestamp in it, and return it instead of the process and thread info objects.

I think I prefer this idea better Alon, if you also think that.

rafaeldtinoco (changes requested):
We agreed the location of Timestamp will be a struct called TimeRelevantInfo instead of duplicates under other Info structs.

rafaeldtinoco (approved):
LGTM

PR #3692, opened 2023-11-15
feat(signatures): add simple proctree datasource envelope

AlonZivony:

I donate this signatures helper code in the hope that it will help other people to use the process tree datasource.

07371088b **feat(signatures): add simple proctree datasource envelope**

### DIFF
--- a/signatures/helpers/proctree.go
+++ b/signatures/helpers/proctree.go
@@ -1,0 +1,124 @@
+package helpers
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/aquasecurity/tracee/types/datasource"
+	"github.com/aquasecurity/tracee/types/detect"
+	"github.com/aquasecurity/tracee/types/trace"
+)
+
+// ProcessTreeDS is an envelope to the process tree datasource API, to make it intuitive and easy
+// to use.
+type ProcessTreeDS struct {
+	ds detect.DataSource
+}
+
+// InitProcessTreeDS init a datasource envelopment instance with the process tree datasource.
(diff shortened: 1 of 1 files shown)


rafaeldtinoco:
This is actually the reason why I started the discussion about having data sources with generic types or having the package types to include the types of whatever we're saving there.

Like you do in this change, you wrap the type casting/checking, to make it easier to be used. I believe this change, for now, is a good change. Im just afraid of having this code not being used by anything hanging there and not tested.

I believe, besides having a simple unit testing (not sure is needed) but, maybe have a signature using the feature (and helper). Do you think we can open source a simple signature using process tree ? Or include a feature in one of the existing signatures ?

rafaeldtinoco (approved):
LGTM

AlonZivony:
> @AlonZivony
> 
> This is actually the reason why I started the discussion about having data sources with generic types or having the package types to include the types of whatever we're saving there.
> 
> Like you do in this change, you wrap the type casting/checking, to make it easier to be used. I believe this change, for now, is a good change. Im just afraid of having this code not being used by anything hanging there and not tested.
> 
> I believe, besides having a simple unit testing (not sure is needed) but, maybe have a signature using the feature (and helper). Do you think we can open source a simple signature using process tree ? Or include a feature in one of the existing signatures ?

Hmm I think we can add a feature of adding lineage to all signatures upon trigger for example.
Also, we can use this helper type now for the e2e test of the process tree. WDYT?

rafaeldtinoco:
> Also, we can use this helper type now for the e2e test of the process tree. WDYT?

Yep, I would go for this **at least**. Because then if something brakes we will get on the tests right away.

PR #3706, opened 2023-11-22
Packet direction flag

### LINKED ISSUE #3701: Add packet direction flag
In order to implement #3697 on top of TCP packet, and since it's useful information in general, the ingress/egress direction of packets should be added to the `net_packet` family of events.

This can be added either as an argument to all these events, or it should be added to a new `Network` context section.
If it's decided that a section should be added instead, it's likely more "correct" to move the following arguments there as well: src, dst, src_port, src_dst.

NDStrahilevitz:

d3c1b3d83 **chore(tests): test packet direction in net e2e**
13dff15d2 **docs(network): mention new metadata argument**
5345cf963 **feature(network):  add packet metadata argument**
18b4663b7 **refactor(network): parse payload in helper**
d8c8d1b70 **chore(go.mod): bump types**

### 2. Explain how to test it

A test was added in the HTTP e2e test.
In addition, one can see for themselves by running any of the changed net packet events (for example `net_packet_icmp` and looking at the new `metadata` argument.

### 3. Other comments

Resolve #3701


tests/e2e-net-signatures/e2e-http.go
```diff
@@ -78,3 +91,15 @@ func (sig *e2eHTTP) OnSignal(s detect.Signal) error {
 }
 
 func (sig *e2eHTTP) Close() {}
+
+func testHttpDirectionAndPacketDirection(md *trace.PacketMetadata, http *trace.ProtoHTTP) bool {
+	// This test is done in the context of curl, but if we were running in a server
```
NDStrahilevitz, line 96:
```suggestion
	// This test is done in the context of curl, but if the test was running a server
```

rafaeldtinoco (approved):
LGTM (after you rebase on top of the types change).

PR #3712, opened 2023-11-28
feat(types): time relevant info for proctree

AlonZivony:

Add a time relevant info object for proctree queries. This will give the query answer the time context it is relevant for.

This is a type change for #3691


rafaeldtinoco (approved):
LGMT

Issue #3576, opened 2023-10-16
Add file hashes to more events


AlonZivony:
There are some useful events that are related to files, which can benefit from adding file hash to them.
Events I have noticed:
* `shared_object_loaded` and the symbols events that derive from it
* Kernel module events

rafaeldtinoco:
Closed by https://github.com/aquasecurity/tracee/pull/3715 at https://github.com/aquasecurity/tracee/pull/3721

PR #3715, opened 2023-11-29
Feature/hash so

AlonZivony:

9fb37130e **feat(events): add hash to symbols_loaded event**
b1e21418e **feat(events): add hash to shared_object_loaded event**

Related to #3576


rafaeldtinoco:
@AlonZivony LGTM, I like it. YOu need to (re)generate the man pages, like the error is telling you to. After that, since all tests passed I can merge right away.

rafaeldtinoco:
image

You need to update the man pages (make man).

AlonZivony:
So it is weird, because I **have** ran `make man`.
You can see that there are .1 pages in the first commit. If I run it again there is no change.

geyslan:
@AlonZivony it's odd. 🤔

I noticed that other .1 files were changed by a lower version of pandoc but the required one.

geyslan:
> @AlonZivony it's odd. 🤔
> 
> I noticed that other .1 files were changed by a lower version of pandoc but the required one.

Well, this made the required changes: https://github.com/aquasecurity/tracee/pull/3720

The comment in pr.yaml was misleading and the docker image was not using a pinned pandoc version; but now it's back on track.

rafaeldtinoco:
> The comment in pr.yaml was misleading and the docker image was not using a pinned pandoc version; but now it's back on track.

With this said, I believe @AlonZivony should rebase, push and then we can merge after tests succeed.

rafaeldtinoco:
Merged by https://github.com/aquasecurity/tracee/pull/3721

PR #3718, opened 2023-11-29
fix(processor): change event context timestamps normalization order

rafaeldtinoco:
Previous comits have changed order for:

t.RegisterEventProcessor(events.All, t.normalizeEventCtxTimes)

It is important that all the time normalization is registered at the end because all the processors are appended to a slice and executed in order and time normalization should be the last thing done when processing events.

NOTE: All the processors for `events.All` will, no matter what, be executed after the `specific event` processors. The problem is to have another `events.All` processor register after this line, then it would run after the normalization is done.

### DIFF
--- a/pkg/ebpf/processor.go
+++ b/pkg/ebpf/processor.go
@@ -80,15 +80,6 @@ func (t *Tracee) RegisterEventProcessor(id events.ID, proc func(evt *trace.Event
 // registerEventProcessors registers all event processors, each to a specific event id.
 func (t *Tracee) registerEventProcessors() {
 	//
-	// Event Timestamps Normalization Processors
-	//
-
-	// Convert all time relate args to nanoseconds since epoch.
-	// NOTE: Make sure to convert time related args (of your event) in here.
-	t.RegisterEventProcessor(events.SchedProcessFork, t.processSchedProcessFork)
-	t.RegisterEventProcessor(events.All, t.normalizeEventCtxTimes)
-
-	//
 	// Process Tree Processors
 	//
 
@@ -130,6 +121,16 @@ func (t *Tracee) registerEventProcessors() {
 	t.RegisterEventProcessor(events.PrintNetSeqOps, t.processTriggeredEvent)
 	t.RegisterEventProcessor(events.PrintMemDump, t.processTriggeredEvent)
 	t.RegisterEventProcessor(events.PrintMemDump, t.processPrintMemDump)
+
+	//
+	// Event Timestamps Normalization Processors
+	//
+
+	// NOTE: Make sure to convert time related args (of your event) BELOW HERE:
+	t.RegisterEventProcessor(events.SchedProcessFork, t.processSchedProcessFork)
+
+	// Keep this last, orelse it will override event context times before other "All" processors.
+	t.RegisterEventProcessor(events.All, t.normalizeEventCtxTimes)
 }
 
 func initKernelReadFileTypes() {


NDStrahilevitz (changes requested):
@rafaeldtinoco I rely on the normalized time in the dns processing stage. Now, I think that the base event should already be normalized in the derivation stage of the `net_packet_dns`, so it may not break it, but i'd err on the side of caution, so i'm adding a -1 on this specific note, please put the dns processor after the normalization one (so move it to be last).

On a more general note, I have some concerns on this being the last processing done. if we ever allow users to register their own processors then it seems reasonable that authors could rely on a normalized time no?

pkg/ebpf/processor.go
```diff
@@ -130,6 +121,16 @@ func (t *Tracee) registerEventProcessors() {
...
+
+	//
+	// Event Timestamps Normalization Processors
+	//
+
+	// NOTE: Make sure to convert time related args (of your event) BELOW HERE:
+	t.RegisterEventProcessor(events.SchedProcessFork, t.processSchedProcessFork)
+
+	// Keep this last, orelse it will override event context times before other "All" processors.
+	t.RegisterEventProcessor(events.All, t.normalizeEventCtxTimes)
```
NDStrahilevitz, line 133: Please move the DNS Cache processor above this.

rafaeldtinoco:
> On a more general note, I have some concerns on this being the last processing done. if we ever allow users to register their own processors then it seems reasonable that authors could rely on a normalized time no?

There was a discussion when this was done like that regarding we should (or should not) have normalized AND non normalized timestamps in the events, and @AlonZivony and I agreed to make all processors work with the raw timing and normalize timings at the end. 

Maybe we can just add a flag saying if the processor needs normalized events or not ? (instead of relying in a hard coded logic ?)

pkg/ebpf/processor.go
rafaeldtinoco: Check my comment below.

rafaeldtinoco: So, DNS cache IS above it. Assuming you meant "below" instead (because you need the non normalized time in the DNS cache), then I you can use `Tracee.getOrigEvtTimestamp()` and revert the timestamp for your specific needs (some other stages later in the pipeline do that).

rafaeldtinoco:
> so i'm adding a -1 on this specific note, please put the dns processor after the normalization one (so move it to be last).

Ok, so actually there is "a problem" I think. The positional change does not change the result for DNS case I believe. The event timestamp normalization happens after the dns event processor before or after this change (or any specific event processor). No matter if its on top or at the bottom.

The positional change only fixes `events.All` processors (specific ones will run before the All ones).

image

First processEvent() will run all processors for the DNS event, and then it will run all the processors registered for "All" (actually Any) event. Meaning that independent of the order we put there, your timestamp use is already happening BEFORE the timestamp is normalized.

That was done on purpose on my side when I did the code, changing event timestamp should happen only at the end of all processors. Also, there are further pipeline stages that might need the original timestamp and they can call `(t *Tracee) getOrigEvtTimestamp()` to get the original timestamp if needed.

With that said, are you sure you depend on normalized timestamps ? Have you tested the `expiredAfter` logic ?

NDStrahilevitz:
> Ok, so actually there is "a problem" I think. The positional change does not change the result for DNS case I believe. The event timestamp normalization happens after the dns event processor before or after this change (or any specific event processor). No matter if its on top or at the bottom.

@rafaeldtinoco What's probably going on here, is that I get the normalized time from the `net_packet_dns_base` which goes through processing #1. Recall that i've added a 2nd call for processing in `deriveEvents`. The `net_packet_dns` event will be generated with the already normalized timestamp. This is why we're fine. So I'll +1. FWIW I still think it's wrong that event processors need to use unnormalized time, if it's needed to split the field into normalized and unnormalized I think that's ok. Alternatively we can add a wrapper pipeline object that includes the event and necessary context which shouldn't be part of the event itself (I think we've discussed this idea before).

AlonZivony:
> Alternatively we can add a wrapper pipeline object that includes the event and necessary context which shouldn't be part of the event itself (I think we've discussed this idea before).

An interesting idea. I am for it overall.
Anther option might be to expose more complicated types from the event, and exporting for the Timestamp field a `GetEpoch` and `GetMonotonic` methods.

rafaeldtinoco:
> Anther option might be to expose more complicated types from the event, and exporting for the Timestamp field a `GetEpoch` and a `GetMonotonic` methods.

I like this approach! https://github.com/aquasecurity/tracee/issues/3726

rafaeldtinoco:
BTW, recent merge from one of @AlonZivony's PR already had this change so I'm closing this.

PR #3725, opened 2023-12-03
feature: writeable data source

### LINKED ISSUE #3704: Add WriteableDataSource interface
## Summary

Add a new capability to Tracee — the ability for data sources to have an optional "Write" method.

## Background

In Tracee, data sources are currently used as a means to access data external to the signature context. They serve as a kind of bridge or interface to this data. The key aspect here is that these data sources are 'opaque', which means their internal workings or the nature of the external data they access are not transparent or directly manipulated by signatures. Initially, the design of these data sources did not include the functionality to write (or modify) the data they can access; they were only meant for reading it in signatures.
Historical Context: In Tracee, any writing of data was an internal process to the context of the data source. Tracee had certain "state managers" that were modified by events occurring within Tracee. These state managers then provided data to what are known as 'signatures' through the data sources. Thus, any writing in these data sources was internal to Tracee. In simple terms, data sources were used as a means to read the data stored or managed by Tracee internally, but they weren't used to modify data, only to expose it in signatures.

## Why

The proposal indicates that these data sources, while currently limited to reading internal state data, could theoretically be expanded to access and possibly modify data in external systems like databases or web endpoints.
The new feature is intended for users who don't want to or can't set up their own service to write data that is accessible through a Tracee data source. In such cases, Tracee could offer a direct method to transfer or write this data to the external sources. 

## Design

The following interfaces would be added:
1. in detect.go:
```
type WriteableDataSource interface {
    Datasource
    Write(key, value any) error
}

var ErrDataWriteFailed = errors.New("data source failed to unpack or write data")
```

2. The following change will be done in the existing data source interface
```
type Datasource interface {
    ...
    Get(key interface{}) (map[string]interface{}, error) -> Get(key any) (any, error)
    ...
}
```

3. The following gRPC methods and messages will be added to the existing TraceeService:
```
message WriteDataSourceRequest {
    string id;
    string namespace;
   google.anypb key;
   google.anypb value;
}

message WriteDataSourceResponse{}

service TraceeService {
    ...
    rpc WriteDataSource(WriteDataSourceRequest) returns (WriteDataSourceResponse)
    rpc StreamDataSource(stream WriteDataSourceRequest) returns (WriteDataSourceResponse)
}
```

The RPC will return OK for a successful write, UNIMPLEMENTED if the data source isn not writable, INVALID_ARGUMENT if the `ErrDataWriteFailed` is returned from the `datasource.Write` method, and INTERNAL for any other error.

NDStrahilevitz:

dec01f2ad **tests(inst): add writable data source test**
c4a7eed1a **feature(grpc): support data source writing**
90c10dc91 **chore(go.mod): bump types and api**

### 2. Explain how to test it

Added e2e instrumentation test.

### 3. Other comments

Resolve #3704


pkg/server/grpc/datasource.go
```diff
@@ -0,0 +1,82 @@
...
+		}
+		return nil, status.Errorf(codes.Internal, "internal error when writing to data source: %v", err)
+	}
+
+	return &pb.WriteDataSourceResponse{}, nil
+}
+
+func (s *DataSourceService) StreamDataSource(stream pb.DataSourceService_StreamDataSourceServer) error {
+	var writeable detect.WriteableDataSource
+	var keyBatch []interface{}
```
NDStrahilevitz, line 50: This needs to be documented for authors.

api/v1beta1/datasource.proto
```diff
@@ -0,0 +1,20 @@
...
+    string namespace = 2;
+    google.protobuf.Value key = 3;
+    google.protobuf.Value value = 4;
+}
+
+message WriteDataSourceResponse{}
+
+service DataSourceService {
+    rpc WriteDataSource(WriteDataSourceRequest) returns (WriteDataSourceResponse);
+    rpc StreamDataSource(stream WriteDataSourceRequest) returns (WriteDataSourceRequest);
```
josedonizetti, line 19:
```suggestion
    rpc StreamDataSource(stream WriteDataSourceRequest) returns (WriteDataSourceResponse);
```

josedonizetti, line 19: perhaps rename to `StreamWrite`?

api/v1beta1/datasource.proto
```diff
@@ -0,0 +1,20 @@
...
+    string id = 1;
+    string namespace = 2;
+    google.protobuf.Value key = 3;
+    google.protobuf.Value value = 4;
+}
+
+message WriteDataSourceResponse{}
+
+service DataSourceService {
+    rpc WriteDataSource(WriteDataSourceRequest) returns (WriteDataSourceResponse);
```
josedonizetti, line 18: perhaps rename to `Write`?

types/detect/detect.go
```diff
@@ -94,3 +94,18 @@ type DataSource interface {
 
 var ErrDataNotFound = errors.New("requested data was not found")
 var ErrKeyNotSupported = errors.New("queried key is not supported")
+var ErrFailedToUnmarshal = errors.New("given value could not be unmarshaled")
+
+type WriteableDataSource interface {
```
josedonizetti, line 99: perhaps rename to `DataSourceWriter`? I think `Writer` instead of `Writeable` matches golang naming better.

josedonizetti (commented):
@NDStrahilevitz  found one typo, and suggested name suggestions. Is write the only operation required? If I added a DNS that no longer makes sense, what should be the process to clear it from the datasource?

api/v1beta1/datasource.proto
NDStrahilevitz: Thanks for catching this :)

NDStrahilevitz: Should we change to `Write` as well since this is in the context of the `DataSourceService` anyway? What do you think?

api/v1beta1/datasource.proto
NDStrahilevitz: Ok, then as said above. I agree, sounds good.

types/detect/detect.go
NDStrahilevitz: I think if we do that then `DataSource` should be changed to `DataSourceReader` too no? But I don't think so anyway because a `WritableDataSource` necessarily has to implement the full `DataSource` interface, and not just be able to write into one. That's why the name conveys additional behavior and not a functionality (this data source is _also_ writable as opposed to this is the _method_ to write to a particular data source).

NDStrahilevitz:
Types PR in #3759.

rafaeldtinoco:
Please rebase, merged the types PR. Will review this one asap.

tests/e2e-inst-test.sh
```diff
@@ -13,7 +13,7 @@ SCRIPT_TMP_DIR=/tmp
 TRACEE_TMP_DIR=/tmp/tracee
 
 # Default test to run if no other is given
-TESTS=${INSTTESTS:=VFS_WRITE}
+TESTS=${INSTTESTS:=WRITABLE_DATA_SOURCE}
```
NDStrahilevitz, line 16: @rafaeldtinoco I forgot to revert this, if you want to merge as is and can't wait for me to fix, go ahead and edit this out.

NDStrahilevitz:
@denisgersh FYI if you want to review.

NDStrahilevitz:
TODO: add the instrumentation test in the pr.yaml file.

Makefile
```diff
@@ -682,6 +682,8 @@ E2E_INST_SRC := $(shell find $(E2E_INST_DIR) \
 		-type f \
 		-name '*.go' \
 		! -name '*_test.go' \
+		-not -path '$(E2E_INST_DIR)/scripts/*' \
```
rafaeldtinoco, line 685: change previous line to '-not' or use '!' like previous line.

tests/e2e-inst-signatures/scripts/ds_writer/ds_writer.go
```diff
@@ -0,0 +1,32 @@
...
+	)
+	if err != nil {
+		panic(fmt.Errorf("failed to dial tracee grpc server: %v", err))
+	}
+	client := v1beta1.NewDataSourceServiceClient(conn)
+	_, err = client.Write(context.Background(), &v1beta1.WriteDataSourceRequest{
+		Id:        "demo",
+		Namespace: "e2e_inst",
+		Key:       structpb.NewStringValue("bruh"),
+		Value:     structpb.NewStringValue("moment"),
```
rafaeldtinoco, line 27: Can we do just a bit more ? I mean, its a data source, maybe feed it with 100k entries and pick a single one ? Maybe a parallel write ? This is the **only** place we would be testing it in open source.

rafaeldtinoco (commented):
You should add more comments to functions and specific variables in this code.

IMO you should also put the external data source example in the documentation as an example how one could implement it (allowing open source users to benefit from the feature and flagging this in the release).

Left a comment about expanding the grpc client + datasource feed test.

tests/e2e-inst-signatures/scripts/ds_writer/ds_writer.go
NDStrahilevitz: done, tell me what you think

rafaeldtinoco (approved):
LGTM. Feel free to merge whenever wanted (after including the docs you mentioned probably).

rafaeldtinoco:
https://github.com/aquasecurity/tracee/pull/3763 has been merged. I believe you need to update types before merging this one now. All good for me (on latest changes).

PR #3729, opened 2023-12-04
fix: filter dispatching to signatures

### LINKED ISSUE #3656: Signatures receive events from pipeline even when are not used
## Description

It seems that signatures `OnEevent` function is called even when they are not used.
Just noticed when for debugging I have put a print inside the `ld_preload` signature, and it printed even when I ran tracee in the following way:
```bash
sudo ./dist/tracee -e sched_process_exec -o json:./out.json -o option:exec-env
```
I don't see any reason for this to happen, as the unused events eBPF programs are not used when the event is not used.

```
Tracee version: "v0.18.0-rc-70-gcc534be9b"
```

## Output of `uname -a`:

```
Linux Alon-Zivony 6.3.6-microsoft-standard-WSL2 #1 SMP PREEMPT_DYNAMIC Mon Jun 12 11:59:43 IDT 2023 x86_64 x86_64 x86_64 GNU/Linux
```

NDStrahilevitz:

1840ee1d540 **fix: clean up untraced signatures**

### 2. Explain how to test it

See the linked below issue.

### 3. Other comments

Fix #3656

### DIFF
--- a/pkg/signatures/engine/engine.go
+++ b/pkg/signatures/engine/engine.go
@@ -18,8 +18,18 @@ const ALL_EVENT_TYPES = "*"
 
 // Config defines the engine's configurable values
 type Config struct {
-	// Enables the signatures engine to run in the events pipeline
-	Enabled             bool
+	// Engine-in-Pipeline related configuration
+	Enabled          bool             // Enables the signatures engine to run in the events pipeline
+	SigNameToEventID map[string]int32 // Cache of loaded signature event names to event ids, used to filter in dispatching
+
+	// Callback from tracee to determine if event should be dispatched to signature.
+	// This is done as a callback becaues importing the events package breaks compilation for the
+	// tracee-rules binary.
+	// When tracee-rules is removed, and the policy coordinator is implemented (PR #3305)
+	// this solution should be abandoned in favor of using it alongside the engine.
+	ShouldDispatchEvent func(eventIdInt32 int32) bool
+
+	// General engine configuration
 	SignatureBufferSize uint
 	Signatures          []detect.Signature
 	DataSources         []detect.DataSource
@@ -248,7 +258,30 @@ drain:
 }
 
 func (engine *Engine) dispatchEvent(s detect.Signature, event protocol.Event) {
+	if engine.config.Enabled {
+		// Do this test only if engine runs as part of the event pipeline
+		if ok := engine.filterDispatchInPipeline(s, event); !ok {
+			return
+		}
+	}
+
 	engine.signatures[s] <- event
+}
+
+func (engine *Engine) filterDispatchInPipeline(s detect.Signature, event protocol.Event) bool {
+	md, err := s.GetMetadata()
+	if err != nil {
+		logger.Warnw(fmt.Sprintf("event %s not dispatched to signature: no metadata", event.Selector().Name))
+		return false
+	}
+	evtName := md.EventName
+	id, ok := engine.config.SigNameToEventID[evtName] // use specialized cache to avoid definiton lookup
+	if !ok {
+		logger.Warnw(fmt.Sprintf("event %s not dispatched to signature: no eventname declared for siganture %s", event.Selector().Name, md.ID))
+		return false
+	}
+
+	return engine.config.ShouldDispatchEvent(id)
 }
 
 // TODO: This method seems not to be used, let's confirm inside the team and remove it if not needed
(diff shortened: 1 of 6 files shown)


rafaeldtinoco:
@josedonizetti I will let you review this one. I would not merge until E2E tests are cleared though.

rafaeldtinoco:
> Agreed, also want to clear with E2E, but seems like they are currently broken.

That is exactly why I said that. Not because this change.

NDStrahilevitz:
> That is exactly why I said that. Not because this change.

Meant that I want to also specifically pre-run this on the E2E before merging, not just (hopefully) have it green post merge. Anyway we agree :p

rafaeldtinoco:
> > That is exactly why I said that. Not because this change.
> 
> Meant that I want to also specifically pre-run this on the E2E before merging, not just (hopefully) have it green post merge. Anyway we agree :p

I agree to you agreeing with my agreement. Thanks.

rafaeldtinoco:
Related to this change ? https://github.com/aquasecurity/tracee/actions/runs/7090197920/job/19296758500?pr=3729#step:4:2873

NDStrahilevitz:
> Member

Possibly

> Related to this change ? https://github.com/aquasecurity/tracee/actions/runs/7090197920/job/19296758500?pr=3729#step:4:2873

Seems possible since I did a small change with event dependencies, I need to check. Didn't bother checking for now since I was occupied with other things and e2es were still red.

rafaeldtinoco:
> Didn't bother checking for now since I was occupied with other things and e2es were still red.

NO rush on my end! Take your time and thanks.

pkg/signatures/engine/engine.go
```diff
@@ -18,8 +18,14 @@ const ALL_EVENT_TYPES = "*"
 
 // Config defines the engine's configurable values
 type Config struct {
-	// Enables the signatures engine to run in the events pipeline
-	Enabled             bool
+	// Engine-in-Pipeline related configuration
+	Enabled          bool             // Enables the signatures engine to run in the events pipeline
+	SigNameToEventID map[string]int32 // Cache of loaded signature event names to event ids, used to filter in dispatching
+	// Callback from tracee to determine if event should be dispatched to signature.
+	// This is done as a callback becaues importing the events package breaks compilation.
```
rafaeldtinoco, line 25: It brakes compilation because of tracee-rules, right ?

```
# github.com/aquasecurity/tracee/cmd/tracee-rules
/usr/lib/go/pkg/tool/linux_amd64/link: running clang failed: exit status 1
/bin/ld: /tmp/go-link-2441279820/000028.o: in function `_cgo_a979db4ce92c_Cfunc_bpf_btf_get_fd_by_id':
/tmp/go-build/btf.cgo2.c:54:(.text+0x12): undefined reference to `bpf_btf_get_fd_by_id'
/bin/ld: /tmp/go-link-2441279820/000030.o: in function `_cgo_a979db4ce92c_Cfunc_perf_buffer__free':
/tmp/go-build/buf-perf.cgo2.c:49:(.text+0x4): undefined reference to `perf_buffer__free'
/bin/ld: /tmp/go-link-2441279820/000030.o: in function `_cgo_a979db4ce92c_Cfunc_perf_buffer__poll':
/tmp/go-build/buf-perf.cgo2.c:67:(.text+0x26): undefined reference to `perf_buffer__poll'
/bin/ld: /tmp/go-link-2441279820/000031.o: in function `_cgo_a979db4ce92c_Cfunc_ring_buffer__free':
/tmp/go-build/buf-ring.cgo2.c:49:(.text+0x4): undefined reference to `ring_buffer__free'
/bin/ld: /tmp/go-link-2441279820/000031.o: in function `_cgo_a979db4ce92c_Cfunc_ring_buffer__poll':
/tmp/go-build/buf-ring.cgo2.c:67:(.text+0x26): undefined reference to `ring_buffer__poll'
```

Meaning, tracee-rules doesn't link libbpf on its binary. It wouldn't be a problem for single binary, right ? Funny because this change is to fix single-binary.

- Either way I would improve this comment to mention all that (historically wise).

A callback is a creative way of picking eventStates, but when we consider the "eventStates" and the fact that we will have a policy coordinator (which @josedonizetti is working on I believe), then this is temporary. So, together with the comment I would mention that it should be changed by the policy orchestrator (forgot the name now) and maybe link the issue to it.

rafaeldtinoco (commented):
Left single comment about next steps IMO, LGTM as a mitigation for the current problem.

rafaeldtinoco (approved):
LGTM after comment is addressed for next steps.

pkg/signatures/engine/engine.go
NDStrahilevitz: Do you know what the issue id for that is? And i'll expand the comment appropriately and merge.

PR #3735, opened 2023-12-05
Remove BPF map macros

### LINKED ISSUE #3731: Currently, bpf macros (e.g. BPF_HASH, BPF_ARRAY, ...) make it impossible to create maps with different properties and they instantiate maps as anonymous structures (do not create the types necessary for the map of maps approach).

geyslan:
Close #3731

0e1786b0a **chore: remove bpf map macros**

```
Define the bpf maps manually, instead of using the macros.

Context: #3731

It also does some renaming and changes filter comments.
```

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -547,12 +547,12 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
     // Update the process tree map (filter related) if the parent has an entry.
 
     if (p.config->proc_tree_filter_enabled_scopes) {
-        u16 pols_version = p.event->context.policies_version;
+        u16 version = p.event->context.policies_version;
         // Give the compiler a hint about the map type, otherwise libbpf will complain
         // about missing type information. i.e.: "can't determine value size for type".
-        // TODO: remove map macros and use typedef for map structs #3508.
-        void *inner_proc_tree_map = &process_tree_map;
-        inner_proc_tree_map = bpf_map_lookup_elem(&process_tree_map_version, &pols_version);
+        process_tree_map_t *inner_proc_tree_map = &process_tree_map;
+
+        inner_proc_tree_map = bpf_map_lookup_elem(&process_tree_map_version, &version);
         if (inner_proc_tree_map != NULL) {
             eq_t *tgid_filtered = bpf_map_lookup_elem(inner_proc_tree_map, &parent_pid);
             if (tgid_filtered) {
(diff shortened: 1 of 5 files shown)


geyslan:
This has for review only the commit 60847f8743183feae6106393810248ab13c0d37e. The others came from https://github.com/aquasecurity/tracee/issues/3239 (be warned).

pkg/ebpf/c/types.h
```diff
@@ -26,6 +26,8 @@ typedef struct task_context {
     u64 parent_start_time;        // parent process task leader's monotonic start time
 } task_context_t;
 
+typedef u16 policies_version_t;
```
rafaeldtinoco, line 29: This is a no IMO. Why to add complexity (besides a huge var name).

rafaeldtinoco:
@geyslan mind rebasing this one ?

geyslan:
As this is not priority and is on top of https://github.com/aquasecurity/tracee/issues/3239, I moved it to [v0.21.0](https://github.com/aquasecurity/tracee/milestone/38).

pkg/ebpf/c/types.h
NDStrahilevitz: FWIW, I prefer the explicitness, not sure how relevant the type change will actually be (if we will need to change it).

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -664,11 +679,23 @@ BPF_HASH(new_module_map, u64, kernel_new_mod_t, MAX_NUM_MODULES);
...
-             kernel_new_mod_t,
-             50); // Likewise for module insertion
+struct recent_deleted_module_map {
+    __uint(type, BPF_MAP_TYPE_LRU_HASH);
+    __uint(max_entries, 50);
+    __type(key, u64);
+    __type(value, kernel_deleted_mod_t);
+} recent_deleted_module_map SEC(".maps");
+
+typedef struct recent_deleted_module_map recent_deleted_module_map_t;
```
NDStrahilevitz, line 689: Are you typedefing these because it will be versioned and we will need to give compiler hints or for consistency's sake?

pkg/ebpf/c/types.h
```diff
@@ -536,7 +532,12 @@ enum file_modification_op
     FILE_MODIFICATION_DONE,
 };
 
-// Type for values representing file types filters
-typedef u32 file_type_filter_t;
+#define MAX_STACK_ADDRESSES 1024 // max amount of diff stack trace addrs to buffer
```
NDStrahilevitz, line 537: I'd keep these consts near their usage, instead of in a separate file (unless they're widely used).

pkg/ebpf/c/tracee.bpf.c
geyslan: The initial idea emerged from that case (compiler hints) but as the removal of macros was already intended the typedef are to let all type definitions in the same pattern of the types.h ones.

pkg/ebpf/c/types.h
geyslan: Ok, I'll revert this. Checked again and noticed that we only use typedef for structs but https://github.com/aquasecurity/tracee/blob/f3fa64f0d3bd8c8203dd8199c182375f78d295ec/pkg/ebpf/c/types.h#L540.

geyslan:
@josedonizetti @rafaeldtinoco would this break your current efforts?

geyslan:
> @josedonizetti @rafaeldtinoco would this break your current efforts?

Rebased on top of https://github.com/aquasecurity/tracee/pull/3802 and got no conflicts.

Issue #3741, opened 2023-12-06
Add labels map to event definition proto


josedonizetti:
The labels will be used to add extra information, like name of a signature for example.

PR #3742, opened 2023-12-06
feature: Add name and properties to Threat, and add Threat to Event definition.

josedonizetti:
Fix https://github.com/aquasecurity/tracee/issues/3741

Add name and properties to Threat, and add Threat to Event definition.


rafaeldtinoco (approved):
LGTM technically correct, feel free to merge or ask someone else for the +1 on naming and conventions if you think it is needed.

yanivagman (approved):
LGTM

api/v1beta1/definition.proto
```diff
@@ -4,12 +4,15 @@ option go_package = "github.co/aquasecurity/tracee/api/v1beta1";
...
 
+import "api/v1beta1/threat.proto";
+
 message EventDefinition {
 	int32 id = 1;
 	string name = 2;
 	Version version= 3;
 	string description = 4 ;
 	repeated string tags = 5;
+	optional Threat threat = 6;
```
NDStrahilevitz, line 15: nit: I think that this should be called properties to align with the internal event definition concept, or vice versa, your call.

NDStrahilevitz (approved):

api/v1beta1/definition.proto
josedonizetti: hmm, I think Threat is best, because we already have properties on it.

Issue #3745, opened 2023-12-06
exec-hash doesn't set the sha256 arg on the first call


geyslan:
I have detected that on the first event with a given filepath, exec-hash calculates the hash but doesn't set the sha256 arg.

`sudo ./dist/tracee -s comm=uname -e sched_process_exec -o option:exec-hash --capabilities add=cap_sys_ptrace -o json`

#### First call on the host

```
host pid: 3199690, pid: 3199690, mntns: 4026531841
device: 66308, inode: 5506453
sourceFilePath: /proc/1/root/usr/bin/uname
capturedFileID: host:/usr/bin/uname
filePath: /usr/bin/uname, castedSourceFileCtime: 1697222770429185913

[...]

Issue #3734, opened 2023-12-05
Improve exec hash performance


itaysk:
Today we include hash for binaries in some events, most notably the process execution events. A couple of performance improvements have been suggested:
1. more efficient caching: today we do cache the calculated hashes to avoid recalculating subsequent executions. but the cache is scoped to the mount namespace, for example same container executed many times will still recalculate the hash. we can improve it by sharing hash calculations across namespaces.
2. use a more efficient hash function. it appears there's an alternative library that could easily win performance gain.

In the future, we should also look into redesigning the event processing pipeline to be more asynchronous.

rafaeldtinoco:
> In the future, we should also look into redesigning the event processing pipeline to be more asynchronous.

You should open a new issue about this particular part, describing what you really mean here. I thought about this long time ago (specially to cope with event buffer buffering effects, to parallelize events processing, etc). I even thought each extension could have its own set of events, probes, maps and ring/perf buffers (considering there would be some bpf maps/progs that would be common to all). Either way, its a big design decision/change that should be somewhere else IMO.

geyslan:
I'm on this. Detected that the first event with exec-hash calculates the hash but is not updated with it:

`sudo ./dist/tracee -s comm=uname -e sched_process_exec -o option:exec-hash --capabilities add=cap_sys_ptrace -o json`

#### First call on the host

```
host pid: 3199690, pid: 3199690, mntns: 4026531841
device: 66308, inode: 5506453
sourceFilePath: /proc/1/root/usr/bin/uname
capturedFileID: host:/usr/bin/uname
filePath: /usr/bin/uname, castedSourceFileCtime: 1697222770429185913

[...]

PR #3746, opened 2023-12-06
Improve performance of exec-hash

geyslan:
Close: #3734 
Close: #3745

f24813f12 **feat(ebpf): expand exec-hash output option**
43f46089b **feat(ebpf): optimize computeFileHash**
e90ad8576 **fix(ebpf): fix getFileHash()**
7fcc7341b **fix(bucketscache): GetBucket was not thread safe**
4d04a15d4 **chore(bucketscache): add benchmark**

f24813f12 **feat(ebpf): expand exec-hash output option**

```
The user can now choose between the following options:

- none (default)
- pathname
- dev-inode
- digest-inode

'pathname' option recalculates the file hash if the inode's creation
time (ctime) differs, which can occur in different namespaces even for
identical pathnames.

'dev-inode' option generally offers better performance compared to the
pathname option, as it bypasses the need for recalculation by
associating the creation time (ctime) with the device (dev) and inode
pair.

'digest-inode' option is the most efficient, as it keys the hash to a
pair consisting of the container image digest and inode. This approach,
however, necessitates container enrichment.
```

43f46089b **feat(ebpf): optimize computeFileHash**

```
Improve the computeFileHash function, by using minio/sha256-simd.

Benchmarking indicates the current implementation is nearly three times
faster than the old version (21,311 ns/op vs. 61,857 ns/op) while
maintaining similar memory efficiency (33,040 B/op vs. 33,056 B/op) and
the same number of memory allocations per operation (5 allocs/op).

goos: linux
goarch: amd64
pkg: github.com/aquasecurity/tracee/pkg/ebpf
cpu: 12th Gen Intel(R) Core(TM) i7-12700H
BenchmarkComputeFileHashOld-20             19257             61857 ns/op           33056 B/op          5 allocs/op
BenchmarkComputeFileHashCurrent-20         56769             21311 ns/op           33040 B/op          5 allocs/op
```

e90ad8576 **fix(ebpf): fix getFileHash()**

```
First call to getFileHash() for a given filename was returning an
empty string.

Context: #3745
```

7fcc7341b **fix(bucketscache): GetBucket was not thread safe**

```
GetBucket was not thread safe since it was returning a reference to a
bucket instead of its copy.
```

4d04a15d4 **chore(bucketscache): add benchmark**

```
It demonstrate that the current algorithm of addBucketItem with multiple
lock/unlock) is optimal compared to a single lock/unlock.

goos: linux
goarch: amd64
pkg: github.com/aquasecurity/tracee/pkg/bucketscache
cpu: 12th Gen Intel(R) Core(TM) i7-12700H
BenchmarkAddBucketItemCurrent-20                     464           2586038 ns/op           2118 B/op           5 allocs/op
BenchmarkAddBucketItemWithOneLock-20                 122           9461730 ns/op            424 B/op           1 allocs/op
```

### DIFF
--- a/docs/docs/flags/output.1.md
+++ b/docs/docs/flags/output.1.md
@@ -11,7 +11,7 @@ tracee **\-\-output** - Control how and where output is printed
 
 ## SYNOPSIS
 
-tracee **\-\-output** <format[:file,...]\> | gotemplate=template[:file,...] | forward:url | webhook:url | option:{stack-addresses,exec-env,relative-time,exec-hash,parse-arguments,parse-arguments-fds,sort-events} ...
+tracee **\-\-output** <format[:file,...]\> | gotemplate=template[:file,...] | forward:url | webhook:url | option:{stack-addresses,exec-env,relative-time,exec-hash={pathname,dev-inode,digest-inode},parse-arguments,parse-arguments-fds,sort-events} ...
 
 
 ## DESCRIPTION
@@ -48,7 +48,10 @@ Other options:
   - **exec-env**: When tracing execve/execveat, show the environment variables that were used for execution.
   - **relative-time**: Use relative timestamp instead of wall timestamp for events.
   - **exec-hash**: When tracing some file related events, show the file hash (sha256).
-    - Affected events: sched_process_exec, shared_object_loaded
+    - Affected events: *sched_process_exec*, *shared_object_loaded*
+    - **pathname** option recalculates the file hash if the inode's creation time (ctime) differs, which can occur in different namespaces even for identical pathnames.
+    - **dev-inode** option generally offers better performance compared to the **pathname** option, as it bypasses the need for recalculation by associating the creation time (ctime) with the device (dev) and inode pair.
+    - **digest-inode**" option is the most efficient, as it keys the hash to a pair consisting of the container image digest and inode. This approach, however, necessitates container enrichment.
   - **parse-arguments**: Do not show raw machine-readable values for event arguments. Instead, parse them into human-readable strings.
   - **parse-arguments-fds**: Enable parse-arguments and enrich file descriptors (fds) with their file path translation. This can cause pipeline slowdowns.
   - **sort-events**: Enable sorting events before passing them to the output. This may decrease the overall program efficiency.
--- a/pkg/ebpf/processor_funcs.go
+++ b/pkg/ebpf/processor_funcs.go
@@ -2,7 +2,6 @@ package ebpf
 
 import (
 	"bytes"
-	"errors"
 	"fmt"
 	"path/filepath"
 	"strconv"
@@ -11,7 +10,7 @@ import (
 	"golang.org/x/sys/unix"
 
 	"github.com/aquasecurity/tracee/pkg/capabilities"
-	"github.com/aquasecurity/tracee/pkg/containers"
+	"github.com/aquasecurity/tracee/pkg/config"
 	"github.com/aquasecurity/tracee/pkg/errfmt"
 	"github.com/aquasecurity/tracee/pkg/events"
 	"github.com/aquasecurity/tracee/pkg/events/parse"
@@ -121,8 +120,9 @@ func (t *Tracee) processSchedProcessExec(event *trace.Event) error {
 	} else {
 		t.pidsInMntns.AddBucketItem(uint32(event.MountNS), uint32(event.HostProcessID))
 	}
+
 	// capture executed files
-	if t.config.Capture.Exec || t.config.Output.CalcHashes {
+	if t.config.Capture.Exec || t.config.Output.CalcHashes != config.CalcHashesNone {
 		filePath, err := parse.ArgVal[string](event.Args, "pathname")
 		if err != nil {
 			return errfmt.Errorf("error parsing sched_process_exec args: %v", err)
@@ -147,6 +147,7 @@ func (t *Tracee) processSchedProcessExec(event *trace.Event) error {
 			if containerId == "" {
 				containerId = "host"
 			}
+
 			capturedFileID := fmt.Sprintf("%s:%s", containerId, filePath)
 			// capture exec'ed files ?
 			if t.config.Capture.Exec {
@@ -177,8 +178,27 @@ func (t *Tracee) processSchedProcessExec(event *trace.Event) error {
 				}
 			}
 			// check exec'ed hash ?
-			if t.config.Output.CalcHashes {
-				err := t.addHashArg(event, filePath, castedSourceFileCtime)
+			if t.config.Output.CalcHashes != config.CalcHashesNone {
+				dev, err := parse.ArgVal[uint32](event.Args, "dev")
+				if err != nil {
+					return errfmt.Errorf("error parsing sched_process_exec args: %v", err)
+				}
+				ino, err := parse.ArgVal[uint64](event.Args, "inode")
+				if err != nil {
+					return errfmt.Errorf("error parsing sched_process_exec args: %v", err)
+				}
+
+				fid := &fileId{
+					dev:             dev,
+					ino:             ino,
+					ctime:           castedSourceFileCtime,
+					pathname:        filePath,
+					mountNS:         event.MountNS,
+					contId:          containerId,
+					contImageDigest: event.Container.ImageDigest,
+				}
+
+				err = t.addHashArg(event, fid)
 				if err != nil {
 					return err
 				}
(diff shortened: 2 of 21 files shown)


pkg/ebpf/tracee.go
```diff
@@ -1641,11 +1641,12 @@ func computeFileHashAtPath(fileName string) (string, error) {
 }
 
 func computeFileHash(file *os.File) (string, error) {
-	h := sha256.New()
+	h := miniosha.New()
```
NDStrahilevitz, line 1644: Will this work cross platform? That was my initial concern with using this library, that I couldn't assess if it would work everywhere.

pkg/ebpf/processor_funcs.go
```diff
@@ -366,32 +373,29 @@ func (t *Tracee) normalizeEventArgTime(event *trace.Event, argName string) error
...
+	}
+	hash, err := t.getFileHash(fileId, fileName, ctime, event.MountNS, event.ContainerID)
 
-		// Container FS unreachable can happen because of race condition on any system,
-		// so there is no reason to return an error on it
-		if errors.Is(err, containers.ErrContainerFSUnreachable) {
-			logger.Debugw("failed to calculate hash", "error", err, "mount NS", event.MountNS)
-			err = nil
-		}
+	event.Args = append(
```
NDStrahilevitz, line 383: Shouldn't you error check first?

pkg/ebpf/tracee.go
```diff
@@ -59,6 +60,45 @@ type fileExecInfo struct {
 	Hash      string
 }
 
+// fileUniqueId is an unique identifier for a file, based on its device and inode.
+type fileUniqueId struct {
```
NDStrahilevitz, line 64: I guess you confirmed that this keying works across containers? So two busybox containers will share inode+device for its `/bin/busybox` files?

pkg/ebpf/processor_funcs.go
geyslan: It was just a removal of outer check, but will adjust that. Thanks.

pkg/ebpf/tracee.go
NDStrahilevitz: I just checked their source code, they do a hardware check themselves and fallback to the default stdlib hash if needed. So we should be ok on this!

pkg/ebpf/processor_funcs.go
```diff
@@ -137,6 +138,12 @@ func (t *Tracee) processSchedProcessExec(event *trace.Event) error {
 		for _, pid := range pids { // will break on success
 			err = nil
 			sourceFilePath := fmt.Sprintf("/proc/%s/root%s", strconv.Itoa(int(pid)), filePath)
+
+			fileId := &fileUniqueId{}
+			if err := fileId.Stat(sourceFilePath); err != nil {
```
AlonZivony, line 143: Why don't we use the inode and device from the event?

pkg/ebpf/tracee.go
geyslan: Good point. I finished it late at night so I didn't check it until now - I'm still sleepy 😴.

Unfortunately the key dev/inode won't work, since each new container has its own mountns (device).

But the good news is that the inode is the same. So we could key the hash with `origin/inode`, origin being `host` or `ImageDigest`

- `"host,inode"`
- `"alpine@sha256:34871e7290500828b39e22294660bee86d966bc0017544e848dd9a255cdf59e0,inode"`

The unique **minor** problem is that `ImageDigest` is only available on the second event arrived from that container, so the hash shouldn't be computed for that case.

WDYT?

```
# alpine 1

FROM Args: dev: 88, ino: 9580902
ContainerID: fd19818e0560b652e09ab923dc26cc48d91947e1cc6c961bf2939a26033a9a13
ImageDigest: alpine@sha256:34871e7290500828b39e22294660bee86d966bc0017544e848dd9a255cdf59e0
fileId: &{dev:88 ino:9580902}
08:30:54:354089  fd19818e0560      fd19818e0560  4026533824   4026533828   0      uname            15      15      1       0                sched_process_exec        cmdpath: /bin/uname, pathname: /bin/busybox, dev: 88, inode: 9580902, ctime: 1701897542997954943, inode_mode: 33261, interpreter_pathname: <nil>, interpreter_dev: <nil>, interpreter_inode: <nil>, interpreter_ctime: <nil>, argv: [uname], interp: /bin/uname, stdin_type: S_IFCHR, stdin_path: /dev/pts/0, invoked_from_kernel: 0, env: <nil>, sha256: 8c6864b97c2b9ff892c859c59363482cb19a546a33a204b83a24f29048b2118f

#alpine 2

FROM Args: dev: 48, ino: 9580902
ContainerID: 1e804fb5bc6bbe0704149564dfb24b372068b4e96cd2f04de37df0ebd7e2fcbc
ImageDigest: alpine@sha256:34871e7290500828b39e22294660bee86d966bc0017544e848dd9a255cdf59e0
fileId: &{dev:48 ino:9580902}
08:31:07:491494  1e804fb5bc6b      1e804fb5bc6b  4026533219   4026533223   0      uname            14      14      1       0                sched_process_exec        cmdpath: /bin/uname, pathname: /bin/busybox, dev: 48, inode: 9580902, ctime: 1701897542997954943, inode_mode: 33261, interpreter_pathname: <nil>, interpreter_dev: <nil>, interpreter_inode: <nil>, interpreter_ctime: <nil>, argv: [uname], interp: /bin/uname, stdin_type: S_IFCHR, stdin_path: /dev/pts/0, invoked_from_kernel: 0, env: <nil>, sha256: 8c6864b97c2b9ff892c859c59363482cb19a546a33a204b83a24f29048b2118f
```

pkg/ebpf/processor_funcs.go
AlonZivony: You don't really need to check, as either way you want to add  the hash arg.
The only difference is that if it fails, the value will be empty.
Though, it might be good idea to put instead of an empty string of an hash an `nil` value if it fails🤔

pkg/ebpf/processor_funcs.go
```diff
@@ -406,5 +410,14 @@ func (t *Tracee) processSharedObjectLoaded(event *trace.Event) error {
 		return nil
 	}
 
-	return t.addHashArg(event, filePath, int64(fileCtime))
+	if t.config.Output.CalcHashes {
+		fileId := &fileUniqueId{}
+		if err := fileId.Stat(filePath); err != nil {
```
AlonZivony, line 415: Again, you can just use the inode and device from the event instead of probing the file.

pkg/ebpf/tracee.go
geyslan: By the way, I am stating for the device/inode while should be just getting those from `event.Args`. Gonna change that.

AlonZivony, line 64: I have a problem with calling this `FileUniqueId` as this is incorrect.
The hash is also dependent on the `ctime`, so it should be included.
We have the concept of file unique ID in the eBPF code, you can take a look over there - https://github.com/aquasecurity/tracee/blob/main/pkg/ebpf/c/types.h#L216

pkg/ebpf/tracee.go
geyslan: I was doing the same. Neat!

pkg/ebpf/tracee.go
```diff
@@ -59,6 +60,45 @@ type fileExecInfo struct {
...
+}
+
+// Stat attempts to retrieve the device and inode numbers from the given path.
+// This function should be called immediately after obtaining the path.
+// This ensures that even if the path is lost or changes, that file can still
+// be identified and accessed using its device and inode numbers.
+func (f *fileUniqueId) Stat(path string) error {
+	onceHashCapsAdd.Do(
+		func() {
+			err := capabilities.GetInstance().BaseRingAdd(cap.SYS_PTRACE)
```
AlonZivony, line 76: I know that you didn't implement it, but I am just checking if someone knows - don't we also need `CAP_DAC_OVERRIDE` for this?

AlonZivony (changes requested):
Overall looks good

pkg/ebpf/processor_funcs.go
geyslan: Yep, I'll do. https://github.com/aquasecurity/tracee/pull/3746#discussion_r1418842702

pkg/ebpf/tracee.go
geyslan: Good point. I can change it to embed ctime, and instead of dev/inode have a string with the parsed origin as mentioned in https://github.com/aquasecurity/tracee/pull/3746#discussion_r1418838352

pkg/ebpf/processor_funcs.go
geyslan: I agree. `nil` as we have for other args.

pkg/ebpf/tracee.go
AlonZivony: And another question - from what I know device ID is not related to mount namespaces.
From what I know, it just changes the mounting points, but devices are kernel structures.
I am not 100% sure of myself though, I just curios to know from where have you brought this information.

geyslan: @AlonZivony take a look at this new approach which let the choice for the user: f24813f12

pkg/ebpf/tracee.go
```diff
@@ -59,6 +59,44 @@ type fileExecInfo struct {
 	Hash      string
 }
 
+type fileId struct {
```
AlonZivony, line 62: I would have made a `fileId` struct for the first 3, and envelope it with another type.
I see us in the future use the 3 file identifiers for other places.

docs/docs/flags/output.1.md
```diff
@@ -48,7 +48,10 @@ Other options:
   - **exec-env**: When tracing execve/execveat, show the environment variables that were used for execution.
   - **relative-time**: Use relative timestamp instead of wall timestamp for events.
   - **exec-hash**: When tracing some file related events, show the file hash (sha256).
-    - Affected events: sched_process_exec, shared_object_loaded
+    - Affected events: *sched_process_exec*, *shared_object_loaded*
+    - **pathname** option recalculates the file hash if the inode's creation time (ctime) differs, which can occur in different namespaces even for identical pathnames.
```
AlonZivony, line 52: Hmm I don't think this documentation is correct.
This option means that the key for finding the hash is the path of the file alone.
Either way, the ctime is checked for all options and if it doesn't match the hash is recalculated.
To be honest, I am not sure what the `pathname` option is good for. You want it for same container deployment, where the path for the same file is constant?

AlonZivony (commented):
Looks very good.
Needs some clarification about the use of the 3 hash keys options.

NDStrahilevitz:
Closing in favor of #3752 since Geyslan is on vacation.

PR #3759, opened 2023-12-13
Writeable datasource types

### LINKED ISSUE #3757: Cannot reuse the same gRPC unix socket path on restart of Tracee
## Description

Let's say the Tracee being used as a service which in case of restart
it will use exactly the same configuration as on the first start.

In the configuration file specified to use a specific path for gRPC socket (unix socket).

In such "mechanism" Tracee will fail to start again on 2nd+ starts.
The cause of this issue is that Tracee doesn't remove the socket file
in case it is already exist at the specified path in the configuration file.

https://github.com/aquasecurity/tracee/blob/c780a063e81fd007cd83d3302f05e31625b9ce1c/pkg/cmd/flags/grpc.go#L25
https://github.com/aquasecurity/tracee/blob/c780a063e81fd007cd83d3302f05e31625b9ce1c/pkg/server/grpc/server.go#L27

Steps for reproduction:
1. Start Tracee with a configuration file that specifies gRPC socket path to `/tmp/tracee.sock`.
2. Restart the Tracee using the same configuration file.

Expected: Tracee will still start listening on the specified gRPC socket path.
Actual: The `bind: address already in use` error will be caught and Tracee will shutdown.

Of course there is workaround for whoever starts the Tracee process to remove the socket file first.
But, I suggest to do it in Tracee - before starting listening - remove the socket file in case it exists.

Example (check the `else if protocol == "unix" { ... }` that I added):
```go
func New(protocol, listenAddr string) (*Server, error) {
	if protocol == "tcp" {
		listenAddr = ":" + listenAddr
	} else if protocol == "unix" {
		_, err := os.Stat(listenAddr)
		if err == nil {
			// File exists, remove it.
			err = os.Remove(listenAddr)
			if err != nil {
				return nil, err
			}
		} else if !os.IsNotExist(err) {
			// File exists, but cannot be stated.
			return nil, err
		}
	}

	lis, err := net.Listen(protocol, listenAddr)
	if err != nil {
		return nil, err
	}

	return &Server{listener: lis, protocol: protocol, listenAddr: listenAddr}, nil
}
```

```bash
$ ./tracee version
Tracee version: "v0.19.0-1-g47951f66"
```

## Output of `uname -a`:

```bash
$ uname -a
Linux ip-10-0-1-8 6.2.0-1017-aws #17~22.04.1-Ubuntu SMP Fri Nov 17 21:07:13 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```

## Additional details 

N/A

NDStrahilevitz:

Add types for PR #3725
Fix #3757 (opportunistically)

dfa6fe56a **feature(types): add WritableDatasource interface**
ef3bcd602 **feat(ebpf): synchronize ready callback to pipeline**
c831127db **fix(grpc): cleanup listen address**

### 2. Explain how to test it

For the bugfix: run tracee with `--grpc-listen-addr` set to something (tested with both unix and tcp) close tracee and run again. Note that tracee will not fail anymore.

### DIFF
--- a/pkg/cmd/flags/grpc.go
+++ b/pkg/cmd/flags/grpc.go
@@ -1,6 +1,7 @@
 package flags
 
 import (
+	"os"
 	"strings"
 
 	"github.com/aquasecurity/tracee/pkg/errfmt"
@@ -20,6 +21,14 @@ func PrepareGRPCServer(listenAddr string) (*grpc.Server, error) {
 
 	if len(addr[1]) == 0 {
 		return nil, errfmt.Errorf("grpc address cannot be empty")
+	}
+
+	// cleanup listen address if needed, for example if a panic happened
+	if _, err := os.Stat(addr[1]); err == nil {
+		err := os.Remove(addr[1])
+		if err != nil {
+			return nil, errfmt.Errorf("failed to cleanup gRPC listening address (%s): %v", addr[1], err)
+		}
 	}
 
 	return grpc.New(addr[0], addr[1])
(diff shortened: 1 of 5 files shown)


rafaeldtinoco (approved):
LGTM

PR #3760, opened 2023-12-13
Bugfix/parse finding type correctly

### LINKED ISSUE #3758: Finding argument type creating does not match its parsing
## Description

In `finding.go`, in the `getCType` function, we translate the type of the argument value to the matching string of the type.
However, the types there doesn't match the types supported by the `UnmarshalJSON` function:
`uint64` is translated to `unsigned long long`, but the `UnmarshalJSON` is expecting the following:
```Go
case "unsigned long", "u64", "off_t", "size_t":
	tmp, err := strconv.ParseUint(num.String(), 10, 64)
	if err != nil {
		return err
	}
	arg.Value = uint64(tmp)
```
The same is true with `int64`.
This result the following error for example:
"unrecognized argument type unsigned long long of argument Container time"

```
Tracee version: "v0.19.0"
```

AlonZivony:

Fix #3758

### DIFF
--- a/pkg/ebpf/finding.go
+++ b/pkg/ebpf/finding.go
@@ -101,6 +101,7 @@ func getArguments(f detect.Finding, triggerEvent trace.Event) []trace.Argument {
 }
 
 // TODO: we probably should have internal types instead of using kernel, or golang types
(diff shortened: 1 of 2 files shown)


rafaeldtinoco (approved):
LGTM

PR #3763, opened 2023-12-14
feat(types): improve datasource write api

NDStrahilevitz:
for #3725 - improve the write api making batch write implementations more convenient for users.


rafaeldtinoco (approved):
LGTM

Issue #3705, opened 2023-11-22
Add ExportedDataSources to .so plugins


NDStrahilevitz:
Currently our shared object plugins only support exporting signatures. Data Sources should be added as well. This can be done in the same way we use a convention for the `ExportedSignatures` symbol, we can add an `ExportedDataSources` symbol.

yanivagman:
Actually I'm not sure we should invest in the shared object (.so) anymore since it simply doesn't work well (e.g. it is required to compile the .so with the same go version as tracee was compiled with). In addition to that, when tracee is compiled statically it can't load dynamic libraries (and we want to support statically compiled binary so tracee can run without any other dependency).
The only use case I can think of for having .so is to update signatures in runtime, however, this use case can also be achieved by updating the tracee binary itself. In the future, we should consider a different mechanism to update signatures, but I don't think .so will be the best solution.

NDStrahilevitz:
Since extensions, or any other replacement for the .so for that matter, are not on this milestone, and we have a need for externally pluggable data sources due to #3704 in this milestone, I think this is still needed, especially since it's not a very large change (data sources and signatures should be very similar in terms of exporting and registering).
When we replace the .so we can move both of these together.

rafaeldtinoco:
> Since extensions, or any other replacement for the .so for that matter, are not on this milestone, and we have a need for externally pluggable data sources due to #3704 in this milestone, I think this is still needed,

You have a good point.

PR #3765, opened 2023-12-17
feat(signatures): expose signatures helpers as Go module

AlonZivony:

To allow users the option to use the up-to-date helpers for Go signatures, the helpers were exported to be an external Go module.

f9a60e590 **feat(signatures): expose signatures helpers as Go module**

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,8 @@ require (
 	github.com/aquasecurity/libbpfgo v0.6.0-libbpf-1.3.0.20240111220235-90dbffffbdab
 	github.com/aquasecurity/libbpfgo/helpers v0.4.6-0.20240111220235-90dbffffbdab
 	github.com/aquasecurity/tracee/api v0.0.0-20240118133111-07e087b036d4
-	github.com/aquasecurity/tracee/types v0.0.0-20231231104405-a33891f29d17
+	github.com/aquasecurity/tracee/signatures/helpers v0.0.0-00010101000000-000000000000
+	github.com/aquasecurity/tracee/types v0.0.0-20240122122429-7f84f526758d
 	github.com/containerd/containerd v1.7.0
 	github.com/docker/docker v24.0.7+incompatible
 	github.com/golang/protobuf v1.5.3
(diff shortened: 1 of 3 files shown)


AlonZivony:
This will require a follow-up PR to remove the `replace` in the `go.mod` file.

yanivagman:
Please see my comment https://github.com/aquasecurity/tracee/issues/3705#issuecomment-1826081041

AlonZivony:
Using signatures as an .so or built in is not so important to determine what to do here in my opinion.
If we import the Tracee module, we need to have sophisticated make file to run unit tests because of the cgo and all.
It is very inconvenient to work with Tracee as a dependency, and also inaccurate because Tracee use the signatures, and not vice-versa.
So dividing the helpers from Tracee to allow more healthy dependencies is the best approach regardless of the way the signatures will be integrated into the Tracee in the end

NDStrahilevitz (approved):
FWIW, i'm +1 for this change.

go.mod
```diff
@@ -180,3 +181,7 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	kernel.org/pub/linux/libs/security/libcap/psx v1.2.68 // indirect
 )
+
+replace github.com/aquasecurity/tracee/signatures/helpers => ./signatures/helpers
```
yanivagman, line 185: Please replace this after merged

yanivagman (approved):
LGTM

go.mod
```diff
@@ -180,3 +181,7 @@ require (
...
+
+replace github.com/kubernetes/cri-api => k8s.io/cri-api v0.23.5-rc.0
```
yanivagman, line 187: Why is that?

signatures/helpers/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/aquasecurity/tracee/signatures/helpers
+
+go 1.19
```
yanivagman, line 3: Please update to 1.21

yanivagman (changes requested):

yanivagman (approved):
LGTM

Issue #3771, opened 2023-12-18
finding.Data parsing to event data fails if not one of the expected types


josedonizetti:
related: https://github.com/aquasecurity/tracee/issues/3767

we need a dynamic solution for arguments which we don't expect a type

PR #3776, opened 2023-12-20
Fix dynamic data arguments1

josedonizetti:

PR #3777, opened 2023-12-20
Fix dynamic data arguments

josedonizetti:
Fix https://github.com/aquasecurity/tracee/issues/3771

This PR fixes the parsing of types on arguments (signatures) that aren't supported by grpc event data. The idea is that users can create signature and pass whatever information as `finding.Data` to fit their needs, but this information must be converted properly to protobuf so the GRPC api can return it when streaming events.

I've tried a few solutions, but it always endup using reflection if tracee was the one responsible to convert the type, which is slow and cubersome, for example, we are only able to read exported fields, nested need deeper logic for convertion. So instead a better solution is to ask the user which information he wants to send on the grpc by implementing an interface, if the interface is implemented tracee calls it and use that data for the argument (kuddos to @NDStrahilevitz for the idea).

Also, we don't break the logic anymore if we can't parse an argument, we log and send the event with the missing data if needed.

### Depends

https://github.com/aquasecurity/tracee/pull/3776


go.mod
```diff
@@ -180,3 +180,7 @@ require (
 )
 
 replace github.com/kubernetes/cri-api => k8s.io/cri-api v0.23.5-rc.0
+
+replace github.com/aquasecurity/tracee/api v0.0.0-20231213190735-f6f40e03b772 => github.com/josedonizetti/tracee/api v0.0.0-20231220122058-c0a887f467f6
+
+replace github.com/aquasecurity/tracee/types v0.0.0-20231219022131-aa8b62c87118 => github.com/josedonizetti/tracee/types v0.0.0-20231220122058-c0a887f467f6
```
josedonizetti, line 186: replace once merged -> https://github.com/aquasecurity/tracee/pull/3776

yanivagman (approved):
LGTM

PR #3779, opened 2023-12-26
Protected finding data

### LINKED ISSUE #3748: Concurrent map iteration and write in finding
## Description

No reproducer yet.

Concurrent iteration happens at this code section:
image
On the `range`.
Not clear from the logs where the concurrent write is, and there doesn't seem to be a write in this function, nor any other goroutine accessing a finding at the same time.

tracee commit a57c534036b81a65174e6892e417551406a93d8d

## Output of `uname -a`:

```
(paste your output here)
```

## Additional details 

relevant section of the map iteration from the log (no write section was found for some reason):
```
fatal error: concurrent map iteration and map write

goroutine 79 [running]:
github.com/aquasecurity/tracee/pkg/ebpf.getArguments({0xc00397edb0, {{{{...}, {...}, {...}}, 0x0}, {0x2ccafc0, 0xc0038645a0}}, {{0x7f7bb10d4380, 0x7}, {0x7f7bb10d200c, ...}, ...}}, ...)
	/tracee/pkg/ebpf/finding.go:65 +0xc5
github.com/aquasecurity/tracee/pkg/ebpf.newEvent(_, {0xc00397edb0, {{{{...}, {...}, {...}}, 0x0}, {0x2ccafc0, 0xc0038645a0}}, {{0x7f7bb10d4380, 0x7}, ...}}, ...)
	/tracee/pkg/ebpf/finding.go:28 +0x7b
github.com/aquasecurity/tracee/pkg/ebpf.FindingToEvent({0xc00397edb0, {{{{...}, {...}, {...}}, 0x0}, {0x2ccafc0, 0xc0038645a0}}, {{0x7f7bb10d4380, 0x7}, {0x7f7bb10d200c, ...}, ...}})
	/tracee/pkg/ebpf/finding.go:24 +0x13e
github.com/aquasecurity/tracee/pkg/ebpf.(*Tracee).engineEvents.func2()
	/tracee/pkg/ebpf/signature_engine.go:100 +0x17a
created by github.com/aquasecurity/tracee/pkg/ebpf.(*Tracee).engineEvents
	/tracee/pkg/ebpf/signature_engine.go:92 +0x7d8
```

NDStrahilevitz:

f72fd506 **fix(finding): use thread safe finding data access**
ff65fc6 **chore(all): change all detect.Finding to pointers**
156569ea **chore(go.mod): bump types**

Fix #3748

### 2. Explain how to test it

Make sure no regression happens, this change should not fundamentally change current behavior, but rather allow correct signature authoring.

### 3. Design choice

I initially wanted to break the Finding struct by adding the following struct:
```golang
type FindingData struct {
     data map[string]any
      l *sync.RWMutex
}
```

And add the thread-safe methods to that struct. This worked perfectly fine, with the api break being functionally not hard to fix with search and replaces. Only one problem happened, the output templates in tracee-rules were broken. I've made multiple attempts at fixing it, including adding custom json marshalling, but nothing worked. The only thing which worked with no breaking was keeping the existing `Data` field in place and untouched.

This introduced the problem of adding the mutex into the existing struct: if we add the mutex as a pointer then using the thread-safe methods as is would cause a nil pointer reference panic. This means that either all users now need to initialize the mutex themselves, or use a constructor which is not enforced for use (except for the otherwise occuring crash). That meant that the mutex must be a non pointer value.
That lead to the problem of mutex copying. mutex copying is an issue, since the locks are now unshared, which would mean that our problem is left unresolved. the way to use non pointer mutexes is if the object itself (the finding) is always passed around as a reference. Which meant changing the whole codebase to do just that.

Why was I still ok with choosing this solution?
1. Because as explained, I couldn't find any other viable one which didn't break templates
2. Because templates and tracee-rules and all that fun stuff will likely be removed once we finish the "everything-is-an-event" project. Alongside that, the finding struct can also be removed, thereby removing this sad design choice.


go.mod
```diff
@@ -171,12 +171,14 @@ require (
...
+	golang.org/x/net v0.19.0 // indirect
+	golang.org/x/sync v0.5.0 // indirect
 	golang.org/x/text v0.14.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	kernel.org/pub/linux/libs/security/libcap/psx v1.2.68 // indirect
 )
 
 replace github.com/kubernetes/cri-api => k8s.io/cri-api v0.23.5-rc.0
+
+replace github.com/aquasecurity/tracee/types => ./types
```
NDStrahilevitz, line 184: TODO: types PR and bump

AlonZivony (approved):
LGTM

PR #3780, opened 2023-12-26
chore(derive/http): change log level when packets are malformed

NDStrahilevitz:

Deriving a net packet event for HTTP when given a malformed request/response would previously return an error. This error is useless for users as there always exists a chance for packet corruption when passing through the eBPF perf buffer. As such, log in WARN level instead.

### 2. Explain how to test it

Before `net_packet_helpers.go:632` add the following line:
image
Then compile and note the logs working.
`tracee --events net_packet_http_request`
`{"level":"warn","ts":1703593308.6970172,"msg":"attempted to derive net_packet_http_request event from malformed packet, event will be skipped","error":"invalid method \"G\\x00T\""}`


AlonZivony (approved):
LGTM

PR #3784, opened 2023-12-31
fix(events): restore dependency in hooked_syscall

### LINKED ISSUE #3783: Hooked syscalls not triggered by kernel module loading
## Description

1. `tracee --events hooked_syscall`
2. Install diamorphine (https://github.com/m0nad/Diamorphine)
3. note that no `hooked_syscall` event is immediately sent

The cause is that the `do_init_module` dependency was removed in the event definition.

```
Tracee version: v0.19.0-112-g1c5ca9564
```

## Output of `uname -a`:

```
Linux ip-172-31-75-171 6.2.0-1017-aws #17~22.04.1-Ubuntu SMP Fri Nov 17 21:07:13 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```

NDStrahilevitz:

629883533:
```
fix(events): restore dependency in hooked_syscall
    
The event dependency for do_init_module in the hooked_syscall event was
removed in commit 43a3eac. As a result, installing a kernel module did
not immediately trigger a syscall table integrity check.
```

Fix #3783

### 2. Explain how to test it

See issue.

### DIFF
--- a/pkg/events/core.go
+++ b/pkg/events/core.go
@@ -10401,6 +10401,7 @@ var CoreEvents = map[ID]Definition{
 		dependencies: Dependencies{
 			ids: []ID{
 				SyscallTableCheck,
+				DoInitModule,
 			},
 		},
 		sets: []string{},


rafaeldtinoco:
IIRC we discussed this during the review of that code change.

NDStrahilevitz:
> IIRC we discussed this during the review of that code change.

I just checked https://github.com/aquasecurity/tracee/pull/3544 and didn't find such a discussion. Is it possible the discussion was internal?
Anyway, there are still references to that code in the processing of `do_init_module`, and there's no comment explaining the removal in the code. I've also talked with @AlonZivony before opening the PR, he said he saw no issue in bringing it back, though it is @OriGlassman who developed, so his input would be good here.

rafaeldtinoco:
> > IIRC we discussed this during the review of that code change.
> 
> I just checked #3544 and didn't find such a discussion. Is it possible the discussion was internal? Anyway, there are still references to that code in the processing of `do_init_module`, and there's no comment explaining the removal in the code. I've also talked with @AlonZivony before opening the PR, he said he saw no issue in bringing it back, though it is @OriGlassman who developed, so his input would be good here.

I just tried to bring him to the discussion because of a "timing" logic that currently exists and was discussed between him and me (https://github.com/aquasecurity/tracee/pull/3544#issuecomment-1763440603). 

Before, as soon as you loaded the kernel module, if there was a hooked syscall you would get right away (after the module was loaded). Now, there is a time window between checks (for hooked syscalls).

NOTE: I didnt check if its directly related to the lack of the dependency, nor went to the code to check, nor said I agree or disagree to your change (fyio).

NDStrahilevitz:
> Before, as soon as you loaded the kernel module, if there was a hooked syscall you would get right away (after the module was loaded). Now, there is a time window between checks (for hooked syscalls).

Yes, and that is because the dependency was removed. FWIW from this discussion it could mean it was intentional on his part. Regardless, for the product, this doesn't work so well for whatever reason. I don't want to add another configuration flag _just_ for this event (we need event parameters/settings whatever for this) to decide between the two methods, so I propose just having them live side by side for now.
Unrelated but the timing window is random (between 10 to 300 seconds), and I'm not sure that is documented anywhere, and it's not configurable.

> NOTE: I didnt check if its directly related to the lack of the dependency, nor went to the code to check, nor said I agree or disagree to your change (fyio).
No harm done, don't worry.

rafaeldtinoco:
> Unrelated but the timing window is random (between 10 to 300 seconds), and I'm not sure that is documented anywhere, and it's not configurable.

The change was intentional and the range randomness was also intentional. Whatever research and product team agrees seems fine, there is no right or wrong, just useful or not. Its good that you're looking into it.

OriGlassman:
There's support for both periodic triggering and per module load triggering (in do_init_module userspace code:
image
)
So it's indeed a bug. 
do_init_module dependency should be added as it is the focal point for triggering other logic such as mem dump and seq ops.

A note about using do_init_module: The reason we use do_init_module, obviously, is that it's where the init code of the module is ran and we wanna check if the init code did something suspicious. 

Using kprobe on do_init_module to do this, might trigger the logic (such as hooked_syscall event) before the init code of the module actually runs. 
Since time passes between kernelspace until do_init_module userspace code in tracee runs, most of the times the init functions of the module finishes.
In cases that the init code execution time is long, we might run before the "interesting" part of the init code of the module actually runs, so this should probably be a kretprobe.

I'll leave a side note that using kretprobe has a downside unfortunately, since an attacker might hang the do_init_module (by using sleep etc), and the kretprobe will never fire. This can be addressed by checking for hanging do_init_module, or maybe by using some other hook (if exist) other than do_init_module.

OriGlassman (approved):

NDStrahilevitz:
@OriGlassman Thanks for the detailed response!

@rafaeldtinoco FYI looks like this can go ahead.

rafaeldtinoco (approved):
LGTM after explanations.

Issue #3795, opened 2024-01-04
revive 1.3.5 doesn't accept go.mod replace


geyslan:
https://github.com/aquasecurity/tracee/actions/runs/7412617218/job/20169799474#step:3:411

geyslan:
It was solved by revive end: https://github.com/mgechev/revive/commit/e33fb872bbf45934cd6110cc66995df1fdc3d84f

PR #3796, opened 2024-01-04
Pin revive version

geyslan:
Close: #3795

186bdedf4 **chore: bump k8s.io/cri-api**

```
- Removes go.mod replace directive for github.com/kubernetes/cri-api.
- Bump it to v0.29.0.

More info: https://kubernetes.io/blog/2022/12/09/kubernetes-v1-26-release/#cri-v1alpha2-removed
```

c51481be9 **fix: pin revive to e33fb87**

```
revive v1.3.5 wrongly includes a replace directive which breaks the
go install command.

This commit pins revive to e33fb87 that is the fix for this issue.
```


rafaeldtinoco (approved):
LGTM

go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	github.com/google/gopacket v1.1.19
 	github.com/hashicorp/golang-lru v0.5.4
 	github.com/hashicorp/golang-lru/v2 v2.0.2
-	github.com/kubernetes/cri-api v0.27.1
+	github.com/kubernetes/cri-api v0.23.5-rc.0
```
NDStrahilevitz, line 18: I think I used the replace because I couldn't get the imports working. I think we _should_ use the newer versions and remove the replace instead if possible.

geyslan: Don't merge yet. Making changes and will trigger a E2E test. Then squash the commits.

geyslan:
> LGTM

Thanks @rafaeldtinoco. I've made changes addressing @NDStrahilevitz concerns. LMK if it's ok.

geyslan:
Commit 186bdedf40f6e3a1bf03a054facc842dd325e270 has been reported to break some local environments. This is due to the local use of golang version from 1.21.

To avoid this, use the same version (1.20.8) of the github workflow (tests) locally:

https://github.com/aquasecurity/tracee/blob/main/.github/actions/build-dependencies/action.yaml#L18

PR #3797, opened 2024-01-07
fix(ebpf): fix hidden_kernel_module error in some kernels

OriGlassman:
On some kernels, while iterating module_kset, we need to check if we've encountered the first module we iterated on, marking we iterated over all modules instead of continuing iterating indefinitely (max amount of iterations - causing a BPF_LOG_ID_UNSPEC error message)

On some kernels, while iterating module_kset, we need to check if we've encountered the first module we iterated on, marking we iterated over all modules instead of continuing iterating indefinitely.

### 2. Explain how to test it
/tracee -e=hidden_kernel_module

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -749,6 +749,7 @@ statfunc bool is_hidden(u64 mod)
 statfunc bool find_modules_from_module_kset_list(program_data_t *p)
 {
     char module_kset_sym[12] = "module_kset";
+    struct module *first_mod = NULL;
     struct kset *mod_kset = (struct kset *) get_symbol_addr(module_kset_sym);
     struct list_head *head = &(mod_kset->list);
     struct kobject *pos = list_first_entry_ebpf(head, typeof(*pos), entry);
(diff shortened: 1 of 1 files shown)


pkg/ebpf/c/tracee.bpf.c
```diff
@@ -770,6 +772,17 @@ statfunc bool find_modules_from_module_kset_list(program_data_t *p)
         if (mod_kobj) {
             struct module *mod = BPF_CORE_READ(mod_kobj, mod);
             if (mod) {
+                if (first_mod_name == NULL) {
```
AlonZivony, line 775: Can't you just compare the module address?
Why do you need the address of the name?

OriGlassman: Totally right. Was too deep with the kobjects. Thanks!

rafaeldtinoco (approved):
LGTM

rafaeldtinoco:
Is this solving the recently reported issue (from an user) where IR code wasn't unrolling this loop more than 60 times (other than the 600 default) ? @AlonZivony @OriGlassman

OriGlassman:
> Is this solving the recently reported issue (from an user) where IR code wasn't unrolling this loop more than 60 times (other than the 600 default) ?

Not aware of such issue, can you provide a link?

rafaeldtinoco:
> @rafaeldtinoco
> 
> > Is this solving the recently reported issue (from an user) where IR code wasn't unrolling this loop more than 60 times (other than the 600 default) ?
> 
> Not aware of such issue, can you provide a link?

https://aquasecurity.slack.com/archives/C03MEDREVL7/p1703659096759989

PR #3800, opened 2024-01-11
fix(capture): restore absolute time in pcap frames

### LINKED ISSUE #3799: Pcap packets timestamp is relative
Currently when capturing network packets into a `pcap` file a relative time is given:
image

This is done because the network events don't go through the normal pipeline, thus their timestamp isn't modified:
```Go
func (t *Tracee) processNetCaptureEvents(ctx context.Context) {
	logger.Debugw("Starting processNetCaptureEvents goroutine")
	defer logger.Debugw("Stopped processNetCaptureEvents goroutine")

	var errChanList []<-chan error

	// source pipeline stage (re-used from regular pipeline)
	eventsChan, errChan := t.decodeEvents(ctx, t.netCapChannel)
	errChanList = append(errChanList, errChan)

	// process events stage (network capture only)
	errChan = t.processNetCapEvents(ctx, eventsChan)
	errChanList = append(errChanList, errChan)

	// pipeline started, wait for completion.
	if err := t.WaitForPipeline(errChanList...); err != nil {
		logger.Errorw("Pipeline", "error", err)
	}
}
```
The normal timestamp change is now done in the `t.normalizeEventCtxTimes` processor function.
Later in the flow the packet timestamp is initialized using the event timestamp:
```Go
func (p *Pcap) write(event *trace.Event, payload []byte) error {
	info := gopacket.CaptureInfo{
		Timestamp:     time.Unix(0, int64(event.Timestamp)),
		CaptureLength: int(len(payload)),
		Length:        int(len(payload)),
	}
````

This problem was probably introduced when we moved the time change to support the process tree processor functions.
Anyways, we should probably just add a step in the network capture pipeline that calls the `t.normalizeEventCtxTimes` function.

Thanks to @oshaked1 for noticing it.

AlonZivony:

1763cd440 **fix(capture): restore absolute time in pcap frames**

Use the normal process events step in the pipeline for the network capture buffer pipeline as well.
This fixes the issue that the conversion of the monotonic time to absolute time moved from the decode step to the process step, leaving the network capture events timestamp monotonic.

Fix #3799

### DIFF
--- a/pkg/ebpf/capture.go
+++ b/pkg/ebpf/capture.go
@@ -15,9 +15,9 @@ import (
 	"github.com/aquasecurity/tracee/pkg/utils"
 )
 
-func (t *Tracee) processFileCaptures(ctx context.Context) {
-	logger.Debugw("Starting processFileCaptures go routine")
-	defer logger.Debugw("Stopped processFileCaptures go routine")
+func (t *Tracee) handleFileCaptures(ctx context.Context) {
+	logger.Debugw("Starting handleFileCaptures go routine")
+	defer logger.Debugw("Stopped handleFileCaptures go routine")
 
 	const (
 		// stat_S_IFMT uint32 = 0170000 // bit mask for the file type bit field
--- a/pkg/ebpf/net_capture.go
+++ b/pkg/ebpf/net_capture.go
@@ -29,9 +29,9 @@ const (
 	familyIpv6
 )
 
-func (t *Tracee) processNetCaptureEvents(ctx context.Context) {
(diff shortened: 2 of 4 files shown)


AlonZivony:
@rafaeldtinoco I didn't do it in this PR, but I would have preferred to just put the network capture event processor in the normal `processEvent` step.
This will require to create priorities for the registration, so processor functions could be called before or after the timestamp change to absolute time.
Anyways it seems to me like a change that should be in its dedicated PR, so not relevant to here.

yanivagman (approved):
LGTM

yanivagman (approved):
LGTM

Issue #3805, opened 2024-01-15
Config policies field need versioning


geyslan:
It's required to manage policy version for config policies fields.

geyslan:
`p->config` filter fields need to be versioned. Extracting all related filter fields (from `config_entry`) into a separate struct/map of maps might be a solution.

PR #3809, opened 2024-01-17
Make policies config versioned

geyslan:
Close: #3805 

Part of #3239.

172f53367122d979cd2e41334ed69b475bd52c5e **chore: make policies config versioned**

```
This extracts the policy-related fields from config_entry_t into a new,
separate struct (policies_config_t), which continues to be stored in the
config_entry_t struct, as well as in the versioned policies_config_map.

This is required to enable access to the appropriate policies_config_t
based on the policy version.
```

### DIFF
--- a/pkg/ebpf/c/common/filtering.h
+++ b/pkg/ebpf/c/common/filtering.h
@@ -9,11 +9,12 @@
 
 // PROTOTYPES
 
+statfunc policies_config_t *get_policies_config(program_data_t *);
 statfunc void *get_filter_map(void *, u16);
 statfunc u64 uint_filter_range_matches(u64, void *, u64, u64, u64);
 statfunc u64 binary_filter_matches(u64, void *, proc_info_t *);
 statfunc u64 equality_filter_matches(u64, void *, void *);
-statfunc u64 bool_filter_matches(u64, bool val);
+statfunc u64 bool_filter_matches(u64, bool);
 statfunc u64 compute_scopes(program_data_t *);
 statfunc u64 should_trace(program_data_t *);
 statfunc u64 should_submit(u32, event_data_t *);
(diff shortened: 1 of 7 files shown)


geyslan:
Gonna document it on NBD.

pkg/ebpf/c/common/filtering.h
```diff
@@ -190,147 +208,164 @@ statfunc u64 bool_filter_matches(u64 filter_out_scopes, bool val)
...
-        u64 mask = ~p->config->pid_filter_enabled_scopes;
-        u64 max = p->config->pid_max;
-        u64 min = p->config->pid_min;
+    if (policies_cfg->pid_filter_enabled_scopes) {
+        u64 filter_out_scopes = policies_cfg->pid_filter_out_scopes;
+        u64 mask = ~policies_cfg->pid_filter_enabled_scopes;
+        u64 max = policies_cfg->pid_max;
+        u64 min = policies_cfg->pid_min;
 
+        filter_map = get_filter_map(&pid_filter_version, version);
```
NDStrahilevitz, line 279: Shouldn't there be a null check here afterwards for the verifier?

NDStrahilevitz (approved):
LGTM

pkg/policy/ebpf.go
```diff
@@ -685,56 +687,240 @@ func (ps *Policies) UpdateBPF(
...
+
+// createNewPoliciesConfigMap creates a new version of the policies config map
+func (ps *Policies) createNewPoliciesConfigMap(bpfModule *bpf.Module) error {
+	version := ps.Version()
+	newInnerMap, err := createNewInnerMap(bpfModule, PoliciesConfigMap, version)
+	if err != nil {
+		return errfmt.WrapError(err)
+	}
+
+	// policies_config_version  u16, uid_filter
```
yanivagman, line 769: is this correct - uid_filter?

yanivagman (approved):
LGTM!

pkg/ebpf/c/common/filtering.h
geyslan: That checking is already being done in:

- https://github.com/aquasecurity/tracee/blob/17b2d6464d5c4a6c5d43eeb07bac4a28e0b28460/pkg/ebpf/c/common/filtering.h#L68
- https://github.com/aquasecurity/tracee/blob/17b2d6464d5c4a6c5d43eeb07bac4a28e0b28460/pkg/ebpf/c/common/filtering.h#L92
- https://github.com/aquasecurity/tracee/blob/17b2d6464d5c4a6c5d43eeb07bac4a28e0b28460/pkg/ebpf/c/common/filtering.h#L152

pkg/policy/ebpf.go
geyslan: Nope, it's a c&p leftover. Fixing it.

Issue #3811, opened 2024-01-17
fix: event data should use an array instead of map


josedonizetti:
Fix event_data to be an array, instead of a map, for better json readbility.

PR #3812, opened 2024-01-17
Fix event data structure

josedonizetti:
Fix https://github.com/aquasecurity/tracee/issues/3811


josedonizetti:
@yanivagman Let me know what you think.

NDStrahilevitz:
This will break the product, are we synchronised with them on merging this change?

josedonizetti:
@NDStrahilevitz yes, we did!

PR #3815, opened 2024-01-20
valediction: tracee extensions (beginning of)

rafaeldtinoco:
I had a meeting with @geyslan and @josedonizetti about this PR. I would like one of the two to open a new PR with this branch of mine and make changes on top of each. When it is ready it can be merged.

From what I can see, the only missing things are:

1. Fix Integration tests to work with new extensions.

2. Fix DNS_DATA_SOURCE issue (network events seem to be fine, but the DNS_DATA_SOURCE signature can't find the dns entry). It is related to my latest change (creating the netpacket extension).

3. Make sure E2E tests pass.

----

Long story short:

A new extension is a "file.go" put inside `pkg/extensions` directory. It defines its own probes, events and initialization. In the future each extension should be on its own package but because event definitions are declarative it couldn't be done now.

Of course this is the beginning of a longer work but it shall be the base to allow files to be added and removed from the tree and the compilation to include events when they are added.

The eBPF code needs to be split (maps and programs) per extension, then each extension will have its own `Module` (check `extensions/extensions.go`).

I hope you guys have fun with this. Note that it is far from perfect and it was a huge refactoring work, always trying to make sure things did not brake (and not necessarily the best picture/idea I wanted).

Please keep credits on the existing commits.

My very best regards!


rafaeldtinoco:
I still intend to fix any issue this PR might have with github tests and create 1 extension as an example (outside the core extension). Hopefully will be able to address all that tomorrow.

geyslan:
@geyslan Gonna return to this in the near future.

PR #3819, opened 2024-01-22
chore: remove replace of signatures helpers

AlonZivony:

Use signatures helpers module directly instead of replace.


yanivagman (approved):
LGTM

PR #3823, opened 2024-01-24
grpc: fix nil arguments

josedonizetti:
Fixes https://github.com/aquasecurity/tracee/issues/3821

### DIFF
--- a/pkg/server/grpc/event_data.go
+++ b/pkg/server/grpc/event_data.go
@@ -42,6 +42,19 @@ func getEventData(e trace.Event) ([]*pb.EventValue, error) {
 			return nil, err
 		}
 
+		// if convertStruct was not able to convert an argument is because we don't support the conversion
+		if eventValue == nil {
+			logger.Errorw(
+				"Can't convert event argument. Please add it as a GRPC event data type or implement detect.FindingDataStruct interface.",
+				"name",
+				arg.Name,
+				"type",
+				fmt.Sprintf("%T", arg.Value),
+			)
+
+			continue
+		}
+
 		eventValue.Name = arg.ArgMeta.Name
 		data = append(data, eventValue)
 	}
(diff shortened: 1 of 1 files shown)


pkg/server/grpc/event_data.go
```diff
@@ -77,6 +73,10 @@ func getEventValue(arg trace.Argument) (*pb.EventValue, error) {
 // based on the value type
 func parseArgument(arg trace.Argument) (*pb.EventValue, error) {
 	switch v := arg.Value.(type) {
+	case nil:
```
geyslan, line 76: Should we consider a default case as well?

josedonizetti: what is the default case?

josedonizetti: We don't return error if the type is not supported, we try to convert it to a structure, and log a message if we can't. We want to send the event even if an argument isn't supported yet: 

https://github.com/aquasecurity/tracee/blob/main/pkg/server/grpc/event_data.go#L341
https://github.com/aquasecurity/tracee/blob/main/pkg/server/grpc/event_data.go#L731

geyslan: Got it! 👍🏻

pkg/server/grpc/event_data.go
```diff
@@ -42,6 +42,11 @@ func getEventData(e trace.Event) ([]*pb.EventValue, error) {
 			return nil, err
 		}
 
+		// if convertStruct was not able to convert an argument is because we don't support the convertion
+		if eventValue == nil {
+			continue
```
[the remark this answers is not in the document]
josedonizetti: It should never happen to our own types, yeah, but user provided types on signature data, might happen if `convertStruct` can't conver the data: 

https://github.com/aquasecurity/tracee/blob/main/pkg/server/grpc/event_data.go#L740
https://github.com/aquasecurity/tracee/blob/main/pkg/server/grpc/event_data.go#L757

we log the convertion to structure which is our default case https://github.com/aquasecurity/tracee/blob/main/pkg/server/grpc/event_data.go#L732, but I guess a log here also worth.

yanivagman (approved):
LGTM

PR #3825, opened 2024-01-25
Fix existing containers tracking by selecting longest cgroup mount path

### LINKED ISSUE #3824: Incorrect Container Tracking in TAS Cloud-Foundry Environment Due to Cgroup Mount Path Selection
## Description
There is an issue in the container tracking mechanism for existing containers within TAS environments. The current implementation selects the last matching record in `/proc/mounts` for cgroup mounts. However, this approach has encountered challenges, particularly in scenarios where the mount paths are listed in a reverse order compared to our initial assumptions.
As a result, existing containers are not being tracked at all in cloud-foundry environment.

**Observed Behavior:**

In specific setups, the order of mount paths in `/proc/mounts` differs from the usual pattern. The last record found, which our system currently relies on to identify the container's mount path, often does not correctly reflect the actual host mount. This leads to a failure in accurately detecting and tracking existing containers.

**Expected Behavior:**

The container tracking mechanism should reliably identify and track containers regardless of the order of mount paths in `/proc/mounts`. The correct host mount path should be determined in a way that is not dependent on the order of the records.

### Steps to Reproduce:

1. **Deploy an Application**: Start by running any application within a TAS (Tanzu Application Service) environment.

2. **Install Tracee**: Proceed to install Tracee in the same TAS environment.

3. **Observe Application Tracking**: After the installation of Tracee, observe the behavior related to application tracking. You may notice that the existing application, which was running prior to the Tracee installation, is not being tracked as expected.

**Suggested Fix:**

A potential solution is to modify the logic for selecting the mount path by considering the length of the path, rather than its position in the list. By selecting the longest mount path, we can achieve a more accurate representation of the host mounts.

```
v0.19.0
```

## Output of `uname -a`:

```
Linux 5c79c746-5998-4cb1-9a97-7bd3d46d76fe 6.2.0-37-generic #38~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Nov  2 18:01:13 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

```

hagitaquasec:
Change the logic to pick the longest cgroup mount path in /proc/mounts. resolving issues with tracking existing containers in TAS Cloud-Foundry environments where mount paths are in reverse order.
This replaces the previous method of using the last record found, ensuring more accurate host mount detection.

Resolves #3824

Modify the logic for selecting the mount path by considering the maximum length of the path, rather than its position in the list.

### How to test it

1. Deploy an Application: Run any application within a TAS (Tanzu Application Service) environment.
2. Install Tracee: Proceed to install Tracee in the same TAS environment.
3. Observe Application Tracking: After the installation of Tracee, observe the behavior related to existing application tracking. If they are tacked correctly.

Also may verify that existing containers are tracked correctly in other environments, e.g: K8S, OCP, etc.

### DIFF
--- a/pkg/mount/mount.go
+++ b/pkg/mount/mount.go
@@ -202,12 +202,14 @@ func SearchMountpoint(fstype string, search string) (string, error) {
 		}
 	}()
 
+	maxLen := 0
 	scanner := bufio.NewScanner(file)
 	for scanner.Scan() {
 		line := strings.Split(scanner.Text(), " ")
 		mountpoint := line[1]
 		currFstype := line[2]
-		if fstype == currFstype && strings.Contains(mountpoint, search) {
+		if fstype == currFstype && strings.Contains(mountpoint, search) && len(mountpoint) > maxLen {
+			maxLen = len(mountpoint)
 			mp = mountpoint
 		}
 	}


yanivagman:
@hagitaquasec what is the assumption you make when choosing the longest mount path? Why will it be the correct one?

hagitaquasec:
@yanivagman 
In `SearchMountpoint` we are targeting the mountpoint associated with /sys. Specifically It searches for the pattern "/sys/fs/cgroup/cpuset". I assume that the mounted path that we actually need matches /sys/fs/cgroup/cpuset directly or ${prefix}/sys/fs/cgroup/cpuset. Which means that it is always longer then the container path.
As discussed previously with @NDStrahilevitz, a more robust solution would be to use `/proc/1/root/sys/`.. since tracee is running in host pid namespace. But when this change is implemented, it will be applied comprehensively throughout the project.

NDStrahilevitz:
> @yanivagman In `SearchMountpoint` we are targeting the mountpoint associated with /sys. Specifically It searches for the pattern "/sys/fs/cgroup/cpuset". I assume that the mounted path that we actually need matches /sys/fs/cgroup/cpuset directly or ${prefix}/sys/fs/cgroup/cpuset. Which means that it is always longer then the container path. As discussed previously with @NDStrahilevitz, a more robust solution would be to use `/proc/1/root/sys/`.. since tracee is running in host pid namespace. But when this change is implemented, it will be applied comprehensively throughout the project.

Hagit FYI, we may stick with mounting and add the volume. Then we would need this PR.

rafaeldtinoco:
#3815 is touching the cgroups package FYIO.

PR #3832, opened 2024-01-30
feat: allow webhook configuration via helm values

ndegory:

903fb1188 **chore: bump chart version to 0.19.1**
32a892a26 **feat: allow webhook configuration via helm values**

the webhook section of the tracee config map can not be configured anymore since the refactoring in version 0.19.0 of the Helm chart. This PR allows to configure the webhook. Choice has been made to map the webhook keys (instead of just dumping the values as is in the config map), similarly to what already exists for other portions of this configmap's values file content.

### 2. Explain how to test it

```
helm template tracee ./deploy/helm/tracee --show-only templates/tracee-config.yaml
```

output doesn't include the webhook configuration (no reg).

```
cat > ci.yaml <<EOF
config:
  output:
    webhook:
      - name: "named-webhook"
        protocol: http
        host: localhost
        port: 8080
        timeout: 3s
        goTemplate: /tracee/templates/simple.tmpl
        contentType: application/json
      - goTemplate: /tracee/templates/simple.tmpl
EOF
helm template tracee ./deploy/helm/tracee --show-only templates/tracee-config.yaml -f ci.yaml
```

output contains this webhook configuration:
```
    output:
        json:
            files:
                - stdout
        options:
            parse-arguments: true
            stack-addresses: false
            exec-env: false
            relative-time: true
            exec-hash: dev-inode
            sort-events: false
        webhook:
            - named-webhook:
                protocol: http
                host: localhost
                port: 8080
                timeout: 3s
                gotemplate: /tracee/templates/simple.tmpl
                content-type: application/json
            - webhook1:
                protocol: http
                host: localhost
                port: 8080
                timeout: 3s
                gotemplate: /tracee/templates/simple.tmpl
                content-type: application/json
```

### 3. Other comments

webhook content example in config samples: https://github.com/aquasecurity/tracee/blob/main/examples/config/global_config.yaml

### DIFF
--- a/deploy/helm/tracee/templates/tracee-config.yaml
+++ b/deploy/helm/tracee/templates/tracee-config.yaml
@@ -30,8 +30,8 @@ data:
         level: {{ .Values.config.log.level }}
     output:
         {{ .Values.config.output.format }}:
-          files:
-            - stdout
+            files:
+                - stdout
         options:
             parse-arguments: {{ .Values.config.output.options.parseArguments }}
             stack-addresses: {{ .Values.config.output.options.stackAddresses }}
@@ -39,6 +39,16 @@ data:
             relative-time: {{ .Values.config.output.options.relativeTime }}
             exec-hash: {{ .Values.config.output.options.execHash }}
             sort-events: {{ .Values.config.output.options.sortEvents }}
+        {{- with .Values.config.output.webhook }}
+        webhook:
+            - {{ .name | default "webhook1" }}:
+                protocol: {{ .protocol | default "http" }}
+                host: {{ .host | default "localhost" }}
+                port: {{ .port | default "8080" }}
+                timeout: {{ .timeout | default "3s" }}
+                gotemplate: {{ .goTemplate }}
+                content-type: {{ .contentType | default "application/json" }}
+        {{- end }}
     {{- if .Values.config.blobPerfBufferSize }}
     blob-perf-buffer-size: {{ .Values.config.blobPerfBufferSize}}
     {{- end }}
--- a/deploy/helm/tracee/values.yaml
+++ b/deploy/helm/tracee/values.yaml
@@ -82,6 +82,16 @@ config:
       relativeTime: true
       execHash: dev-inode
       sortEvents: false
+    # uncomment config.output.webhook to enable a single webhook
+    # to configure multiple webhooks, use the traceeConfig field
+    # webhook:
+    #   name: "webhook1"
+    #   contentType: "application/json"
+    #   goTemplate: ""
+    #   host: ""
+    #   port: "8080"
+    #   protocol: http
+    #   timeout: 3s
 
 defaultPolicy: true
 


deploy/helm/tracee/Chart.yaml
```diff
@@ -18,7 +18,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: "0.19.0"
+version: "0.19.1"
```
josedonizetti, line 21: let's keep the same version, as we are releasing the chart along with tracee, the 0.20.0 release will be soon, and we can bump it before releasing

josedonizetti (commented):
@ndegory Thanks so much for the pr. Left a comment! Also, would you be able to address another request related to it that a user asked on slack?

When using gotemplate in the config, we need to have the template file in the container. I was thinking about having a specific configmap for it, if the user pass a template we can apply it on the cluster, and then mount it on the daemonset. WDYT?

ndegory:
> When using gotemplate in the config, we need to have the template file in the container. I was thinking about having a specific configmap for it, if the user pass a template we can apply it on the cluster, and then mount it on the daemonset. WDYT?

yes, that would be useful. So far I've been using a post renderer patch to do that, but I'd prefer to have this available in the chart. I'll add it to the PR.

geyslan:
As still in testing, pushed it to next milestone.

josedonizetti:
@ndegory Sorry for the delay, it was holidays in Brazil. Do you have an example of how you would use this option? Because currently we allow two options:

passing a full config file:
```
 helm install tracee aqua/tracee \
        --namespace tracee --create-namespace \
        --set-file traceeConfig=myconfig.yaml
```

or 

passing specific options:
```
helm install tracee aqua/tracee \
        --namespace tracee --create-namespace \
        --set config.blobPerfEventSize=1024
```

And when using single options, I was trying to not support multiple values as it make the helm syntax complex, for example, webhook now could be configured with `--set webhook=FULL_URL`

- https://github.com/aquasecurity/tracee/blob/main/deploy/helm/tracee/values.yaml#L57 
- https://github.com/aquasecurity/tracee/blob/main/deploy/helm/tracee/templates/daemonset.yaml#L38-L40

Though I like your change, as webhook should be under the config options, not outside by itself, maybe I'm wondering if perhaps we should simplify the syntax to allow one webhook to be configured, and if someone wants more options they can use the whole file option, but not sure, WDYT?

nowdegory:
@josedonizetti 

> currently we allow two options

don't you also support passing a values file with `--values`? This is what most people would use when more a few values need to be set.

> perhaps we should simplify the syntax to allow one webhook to be configured, and if someone wants more options they can use the whole file option, but not sure, WDYT?

Works for me, I'll update the PR.

nowdegory:
I cannot update the PR description, the optional webhook configuration in the config section is not an array anymore:
```
config:
  output:
    webhook:
        protocol: http
        host: localhost
        port: 8080
        timeout: 3s
        goTemplate: /tracee/templates/simple.tmpl
        contentType: application/json
 ```
I tested it successfully with a values file, but also with the --set flag:
```
$ helm template tracee ./deploy/helm/tracee --show-only templates/tracee-config.yaml --set config.output.webhook.goTemplate=/tracee/templates/simple.tmpl --set config.output.webhook.host=example.com
```

josedonizetti (approved):

PR #3834, opened 2024-01-31
fix(ebpf): fix hidden_kernel_module not found symbol

OriGlassman:

Change hidden_kernel_module logic so that it's not based on layout_and_allocate symbol which is not always present in all kernels

### 2. Explain how to test it


pkg/ebpf/c/tracee.bpf.c
```diff
@@ -964,8 +955,7 @@ statfunc int check_is_proc_modules_hooked(program_data_t *p)
         }
 
         if (unlikely(mod_base_addr == 0)) { // Module memory was possibly tampered.. submit an error
-            ret = 7;
-            break;
+            return 7;
```
AlonZivony, line 958: ditto - use const

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -979,15 +969,15 @@ statfunc int check_is_proc_modules_hooked(program_data_t *p)
...
+                return 0;
             }
 
             // Module was not seen in proc modules and there was no recent insertion, report.
             lkm_seeker_send_to_userspace(pos, &flags, p);
         }
     }
 
-    return ret;
+    return -5;
```
AlonZivony, line 980: ditto - use const

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1072,19 +1067,26 @@ int uprobe_lkm_seeker(struct pt_regs *ctx)
...
-    if (init_shown_modules() != 0) {
-        tracee_log(ctx, BPF_LOG_LVL_ERROR, BPF_LOG_ID_UNSPEC, 1);
+    int ret = init_shown_modules();
+    if (ret != 0) {
+        tracee_log(ctx, BPF_LOG_LVL_ERROR, BPF_LOG_ID_UNSPEC, ret);
         return 1;
     }
 
+    // On first run, do a scan only relevant for modules that were inserted prior tracee started.
+    if (!hidden_old_mod_scan_done) {
```
AlonZivony, line 1077: I am not sure how, but you probably need to make sure to do an atomic operation here.
I think `__sync_val_compare_and_swap` might work.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1168,7 +1178,7 @@ int lkm_seeker_new_mod_only_tail(struct pt_regs *ctx)
 
     u64 start_scan_time = check_new_mods_only(&p);
     if (start_scan_time == 0) {
-        tracee_log(ctx, BPF_LOG_LVL_ERROR, BPF_LOG_ID_UNSPEC, 6);
+        tracee_log(ctx, BPF_LOG_LVL_ERROR, BPF_LOG_ID_UNSPEC, -7);
```
AlonZivony, line 1181: ditto - const

AlonZivony (changes requested):
All looks great.
Only issue is using consts for all error values.

pkg/ebpf/c/tracee.bpf.c
OriGlassman: Since this is a uprobe triggered based, and is triggered from userspace from a single location, this code can run only once at a time

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -964,8 +959,7 @@ statfunc int check_is_proc_modules_hooked(program_data_t *p)
         }
 
         if (unlikely(mod_base_addr == 0)) { // Module memory was possibly tampered.. submit an error
-            ret = 7;
-            break;
+            return -7;
```
AlonZivony, line 962: Still needs a const here

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -950,8 +946,7 @@ statfunc int check_is_proc_modules_hooked(program_data_t *p)
         pos = n;
         n = list_next_entry_ebpf(n, list);
         if (&pos->list == head) {
-            ret = 0;
-            break;
+            return 0;
```
AlonZivony, line 949: Should be `MOD_NOT_HIDDEN`?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -979,15 +973,15 @@ statfunc int check_is_proc_modules_hooked(program_data_t *p)
             // filled (userspace) - so assume it happened max 2 seconds prior to
             // that.
             if (start_scan_time_init_shown_mods - (2 * 1000000000) < last_module_insert_time) {
-                continue;
+                return 0;
```
AlonZivony, line 976: `MOD_NOT_HIDDEN`

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1095,19 +1101,22 @@ int lkm_seeker_proc_tail(struct pt_regs *ctx)
     if (!init_tailcall_program_data(&p, ctx))
         return -1;
 
-    int ret = check_is_proc_modules_hooked(&p);
-    if (ret != 0) {
+    int ret = find_modules_from_module_kset_list(&p);
+    if (ret == HID_MOD_RACE_CONDITION) {
+        tracee_log(ctx, BPF_LOG_LVL_WARN, BPF_LOG_ID_UNSPEC, ret);
```
AlonZivony, line 1106: Maybe we should add a parser of the values in user-mode to the const name

pkg/ebpf/c/tracee.bpf.c
AlonZivony: This is not a good reason not implementing it thread-safe, as there is a risk here.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1072,19 +1071,26 @@ int uprobe_lkm_seeker(struct pt_regs *ctx)
     }
 
     start_scan_time_init_shown_mods = bpf_ktime_get_ns();
```
AlonZivony, line 1073: This should probably be an atomic write

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -692,14 +692,13 @@ struct recent_deleted_module_map {
...
 
-struct recent_inserted_module_map {
-    __uint(type, BPF_MAP_TYPE_LRU_HASH);
-    __uint(max_entries, 50); // likewise for module insertion
-    __type(key, u64);
-    __type(value, kernel_new_mod_t);
-} recent_inserted_module_map SEC(".maps");
-
-typedef struct recent_inserted_module_map recent_inserted_module_map_t;
+static u64 start_scan_time_init_shown_mods = 0;
```
AlonZivony, line 695: I think we need to consider when using globals in the eBPF code that they may run from multiple cores.
This means that synchronizations problems are a risk.
Even if a program cannot run in parallel, there is the risk of core caches not updating in time.
I think each use of global values should be implemented using atomic operations.

AlonZivony (changes requested):

pkg/ebpf/c/tracee.bpf.c
OriGlassman: No.. not hidden refers to a specific module. This just indicates the search is over

pkg/ebpf/c/tracee.bpf.c
OriGlassman: ditto

pkg/ebpf/c/tracee.bpf.c
OriGlassman: Don't think it's critical. What's more important IMO is tracee's supporting log with a message from eBPF... but that's not for this PR.

geyslan: > What's more important IMO is tracee's supporting log with a message from eBPF...

Ori, `tracee_log()` is not working? Did I get you right?

In fact, `BPF_LOG_ID_UNSPEC` won't help us at all if we use it for everything. When a new log message needs to be used, I suggest implementing it in ebpf and userland:

https://github.com/aquasecurity/tracee/blob/b8f55163665b01654022eaab32858088a28e6096/pkg/ebpf/c/types.h#L393-L407

https://github.com/aquasecurity/tracee/blob/b8f55163665b01654022eaab32858088a28e6096/pkg/ebpf/bpf_log.go#L18-L61

OriGlassman: tracee_log is working. What I meant is adding a string argument to tracee_log that will be passed to userspace as is.
Like a userspace logger... tracee_log("Hello from eBPF") for that matter.
With the current tracee_log, I'm using the return value to distinguish the type of error.

geyslan: > Like a userspace logger... tracee_log("Hello from eBPF") for that matter.

Doing that would increase a lot the logic of it. We've already faced inst limit exceeded before using it as it is (only the id of the log).

For now you can define ids in bpf and in userland letting the latter decode the log correctly, showing an static string (from errorMap). Just use your actual ret values as IDs.

AlonZivony (approved):
LGTM

geyslan:

OriGlassman:
> We have this update #3875
> 
> Please rebase your PR against main to make use of the new workflow setup.

Rebased

yanivagman (approved):
LGTM

PR #3835, opened 2024-01-31
fix: skip timestamp normalizing in derived events

### LINKED ISSUE #3828: Tracee outputs bogus timestamps
## Description

When running tracee with JSON output, some events contain bogus timestamps (very large or even negative numbers). In these cases, both the timestamp and threadStartTime fields contain (slightly different) invalid timestamps.

```
Tracee version: "v0.19.0"
```

## Output of `uname -a`:

```
Linux *********** 5.15.133.1-microsoft-standard-WSL2 #1 SMP Thu Oct 5 21:02:42 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```

## Additional details 

Example of invalid timestamps (rest of event truncated):
```
{'timestamp': 33259358447748, 'threadStartTime': 33259355176567,
{'timestamp': -8207706047079014194, 'threadStartTime': -8207706047079194457,
{'timestamp': 8532537232442238823, 'threadStartTime': 8532537232441643203,
{'timestamp': -1381702870323630858, 'threadStartTime': -1381702870324338633,
{'timestamp': 6826036438253662974, 'threadStartTime': 6826036438252929247,
```

Example of a valid timestamp:
```
{'timestamp': 1706534053545998934, 'threadStartTime': 1706534053543890523,
```

Unique events with invalid timestamps:
```
{'security_file_mprotect', 'security_bprm_check', 'do_mmap', 'shared_object_loaded', 'security_file_open', 'sched_process_exec', 'execve'}
```

NDStrahilevitz:

**a78e7c129e**: fix: skip timestamp normalizing in derived events

### 2. Explain how to test it

1. `tracee --events container_create`
2. `docker run --rm ubuntu`
3. Timestamp should make sense

### 3. Other comments
 
Fix #3828

### DIFF
--- a/pkg/ebpf/processor_funcs.go
+++ b/pkg/ebpf/processor_funcs.go
@@ -331,6 +331,12 @@ func (t *Tracee) processPrintMemDump(event *trace.Event) error {
 // normalizeEventCtxTimes normalizes the event context timings to be relative to tracee start time
 // or current time in nanoseconds.
 func (t *Tracee) normalizeEventCtxTimes(event *trace.Event) error {
+	eventId := events.ID(event.EventID)
+	if eventId > events.MaxCommonID && eventId < events.MaxUserSpace {
+		// derived events are normalized from their base event, skip the processing
+		return nil
+	}
+
 	//
 	// Currently, the timestamp received from the bpf code is of the monotonic clock.
 	//


pkg/ebpf/processor_funcs.go
```diff
@@ -331,6 +331,12 @@ func (t *Tracee) processPrintMemDump(event *trace.Event) error {
 // normalizeEventCtxTimes normalizes the event context timings to be relative to tracee start time
 // or current time in nanoseconds.
 func (t *Tracee) normalizeEventCtxTimes(event *trace.Event) error {
+	eventId := events.ID(event.EventID)
+	if eventId > events.MaxCommonID && eventId < events.MaxUserSpace {
```
josedonizetti, line 335: nitpick: and we extract for this with better naming? `events.IsDerivedEvent()`
wdyt?

josedonizetti (approved):
Looks good, left a nitpick

pkg/ebpf/processor_funcs.go
geyslan: > nitpick: can we extract for this with better naming? `events.IsDerivedEvent()`
> wdyt?

LGTM as well. The method suggested is a good idea.

NDStrahilevitz: I think that when we do extensions, we will need specified "sections" per event type. Then we can easily classify by section location. I assume that that change will break this one, so we don't necessarily need this right now.

PR #3836, opened 2024-01-31
fix: change missing probe log level

josedonizetti:
It should be a warn when an event can't be load becaues of a missing probe.


geyslan (approved):
LGTM

NDStrahilevitz (approved):
Can we add in the log the reason the probe is missing (the error message of the failed attachment)?

yanivagman (approved):
LGTM

josedonizetti:
@NDStrahilevitz error added!

PR #3841, opened 2024-02-01
chore: remove gob printer

josedonizetti:
Remove the gob printer. This is a legacy printer which was used by `tracee-rules` and `tracee-ebpf`, which are no longer supported and will be soon removed from the repo.


josedonizetti:
depends on https://github.com/aquasecurity/tracee/pull/3842

geyslan:
> depends on #3842

It was already merged. Let's get rid of gob (finally 😀)?

geyslan (approved):

PR #3842, opened 2024-02-01
fix: e2e-net-tests should use unified binary

josedonizetti:
Fixing e2e tests to use the `tracee` binary instead of `tracee-rules` and `tracee-ebpf`.

### DIFF
--- a/tests/e2e-kernel-test.sh
+++ b/tests/e2e-kernel-test.sh
@@ -61,7 +61,7 @@ info
 set -e
 make -j$(nproc) all
 set +e
-if [[ ! -x ./dist/tracee-ebpf || ! -x ./dist/tracee-rules ]]; then
+if [[ ! -x ./dist/tracee ]]; then
     error_exit "could not find tracee executables"
 fi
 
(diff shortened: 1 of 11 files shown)


tests/e2e-kernel-test.sh
```diff
@@ -59,9 +59,9 @@ info "= COMPILING TRACEE ============================================"
 info
 # make clean # if you want to be extra cautious
 set -e
-make -j$(nproc) all
+# make -j$(nproc) all
```
yanivagman, line 62: No need to make here?

NDStrahilevitz: ditto

NDStrahilevitz (approved):
LGTM, but please fix the commented area before merging.

tests/e2e-kernel-test.sh
```diff
@@ -61,7 +61,7 @@ info
 set -e
 make -j$(nproc) all
 set +e
-if [[ ! -x ./dist/tracee-ebpf || ! -x ./dist/tracee-rules ]]; then
+if [[ ! -x ./dist/tracee ]]; then
     error_exit "could not find tracee executables"
```
geyslan, line 65:
```suggestion
    error_exit "could not find tracee executable"
```

tests/e2e-kernel-test.sh
```diff
@@ -76,26 +76,16 @@ for TEST in $TESTS; do
...
-        --output option:parse-arguments \
-        --scope container=new \
-        --events $events \
-        2>$SCRIPT_TMP_DIR/ebpf-$$ |
-        ./dist/tracee-rules \
-            --input-tracee=file:stdin \
-            --input-tracee format:gob \
-            --rules $TEST 2>&1 |
+        --output json \
+        --scope container=new 2>&1 |
```
geyslan, line 85:
```suggestion
        --scope container=new 2>&1 \
        | tee "$SCRIPT_TMP_DIR/build-$$" &
```

Added backslash at the end of the line and moved the pipe to the beginning of the next. This makes the intent more clear.

Removed the second `2>&1` after tee since redirecting both stdout and stderr is not necessary; tee already captures both streams.

WDYT?

geyslan (approved):
LGTM, put a request related to missing \ and commented other stuff.

Issue #3817, opened 2024-01-22
No default `exec-hash` option


AlonZivony:
## Description

Recently we have changed the `exec-hash` feature to increase its performance (#3752 ).
To do so, options for the hash caching were introduced.
However, the implementation for choosing the caching key requires the user to specify it by configuration.
This broke the previous API for this feature, requiring users to update their config file or CMD usage.
The use of `-o option:exec-hash` is not supported anymore, and the form of `-o option:exec-hash=<option>` became mandatory.
There is no real reason for breaking the previous API, which could have been preserved by simply providing default algorithm if non was chosen.
Even without the previous API to consider, providing a default option seem wise as it is not trivial to understand which option suits the user more.

NDStrahilevitz:
I see two possible options:

1. Conditional default - if enrichment is enabled (which is the default) use the digest based option, otherwise use dev+inode.
2. always use dev+inode

@yanivagman @AlonZivony  WDYT?

AlonZivony:
If it is easy to implement, I think that the first option is better as the user won't probably know that once he activates the enrichment he should switch manually.

PR #3852, opened 2024-02-08
Set exec-hash default option

geyslan:
Close: #3817 

354c152ff **fix: typo**
e481dbba9 **fix(flags): rollback exechash cli api**

e481dbba9 **fix(flags): rollback exechash cli api**

```
exechash can be set again as 'option:exec-hash' which will set it now as
'dev-inode' as default.
```

### DIFF
--- a/docs/docs/flags/output.1.md
+++ b/docs/docs/flags/output.1.md
@@ -2,7 +2,7 @@
 title: TRACEE-OUTPUT
 section: 1
 header: Tracee Output Flag Manual
-date: 2023/10
+date: 2024/02
 ...
 
 ## NAME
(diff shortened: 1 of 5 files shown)


geyslan:
I didn't tackled this [first option](https://github.com/aquasecurity/tracee/issues/3817#issuecomment-1904433102) since it would demand to propagate information of the flag chosen through the parsing. But can be tackled via other issue.

AlonZivony (approved):
LGTM.
However, this does not address what @NDStrahilevitz offered - when the containers enrichment is active using the `digest-inode` option is preferable.
IMO, this is good enough, as the enrichment is not necessarily stable for all events.
However, this will probably result users never use the digest option, even though it is better overall except if enrichment is not stable.

PR #3853, opened 2024-02-08
chore(k8s): prepare v0.20.0 release

geyslan:


geyslan:
[Failed](https://github.com/aquasecurity/tracee/actions/runs/7832926710/job/21372729301?pr=3853#step:8:227) since the image is still not created.

PR #3865, opened 2024-02-15
chore: add run-tracee.sh script

geyslan:

This script is used to run tracee easily. It is a wrapper around the docker run command.


josedonizetti:
@geyslan this is marked as draft, you still working on it or expecting a review already?

josedonizetti:
@geyslan hahaha, sorry man, but I was expecting you would change it to something else. LOL
Yeah, hiding the docker command like that won't be an improvement, better to be explicit about what we are doing :*

Issue #3867, opened 2024-02-16
[Chore] Update opa version from 0.48.0 to something more recent (eliminate CVEs)


hangrymuppet:
opa 0.48.0 is quite old (over a year) and is triggering a high number of CVEs counts in our security scanners.

Most of them are probably not exploitable because of the way opa tool is used by tracee, but it is creating noise.

josedonizetti:
@hangrymuppet OH! Thanks for catching it, will fix it.

josedonizetti:
@hangrymuppet -> https://github.com/aquasecurity/tracee/pull/3887

PR #3868, opened 2024-02-17
chore: bump opa to 0.61.0

josedonizetti:
Fix https://github.com/aquasecurity/tracee/issues/3867


geyslan:
Perhaps splitting the go.mod bump into its own commit?

geyslan (approved):

PR #3873, opened 2024-02-20
Fix `process_execute_failed` event error

AlonZivony:


tests/e2e-inst-signatures/e2e-process_execute_failed.go
```diff
@@ -0,0 +1,107 @@
...
+
+func (sig *e2eProcessExecuteFailed) OnEvent(event protocol.Event) error {
+	eventObj, ok := event.Payload.(trace.Event)
+	if !ok {
+		return fmt.Errorf("failed to cast event's payload")
+	}
+
+	switch eventObj.EventName {
+	case "init_namespaces":
+		// The event does not support kernel versions 5.7 and older, so pass the test if run in such
```
NDStrahilevitz, line 58: Was this the case for the event before? Or is the only way to fix the event breaking it for kernels older than 5.7? I'd be cautious introducing the fix if so.

AlonZivony: Previous symbols was not  reliable, any kernel could have missed it depending its compilation and configuration.
The new probes are an LSM function and sys_exit tail call, which should be reliable from kernel 5.8 on.
Supporting older kernels is very hard, as there are no real function there to sit on with access to the `bprm` object to extract from the execution information for the event.

NDStrahilevitz: If the previous version doesn't give false positives, then I don't think it's responsible to break the event, and the fixed version should be supplied in addition (maybe we can add a simple kernel version condition to probe dependencies. could be a useful feature in general).
Otherwise, we should deprecate the event in versions which don't have the fix.

AlonZivony: To be honest, the best solution will be to run the current event if possible, as it is much simpler implementation.
In the case the symbol is missing, then we can use the new and more reliable solution.
However, this is not possible in the current architecture and the refactor is much too complicated for this PR as I see it.

AlonZivony: @NDStrahilevitz you can go over the new PR at #3881

AlonZivony:
Moved to #3881

PR #3875, opened 2024-02-20
Use EXECUTION_TYPE label for github self host runner

sharon-amir:
Also use short label name 'graas' instead of 'github-self-hosted'


geyslan (approved):

geyslan:
Thanks @sharon-amir! 👍🏼

Issue #154, opened 2020-07-22
possibly incorrect mode_t size


itaysk:
noticed that we read `mode_t` as `uint32` but it might be `unsigned short`? (architecture dependent?)

https://github.com/aquasecurity/tracee/blob/de741857ace9779a0207a7c346ae4ee70bf206fc/tracee/tracee.go#L904
https://elixir.bootlin.com/linux/latest/source/include/linux/types.h#L18

itaysk:
on x86_64 it's int32 which is ok

PR #3881, opened 2024-02-21
Fix `process_execute_failed` only for supported new kernels

### LINKED ISSUE #3356: Missing symbols on newer kernels (> ~6.3)
## Description

I noticed that my 6.3.6 machine is missing the following symbols:
* `exec_binprm` - part of the `process_execute_failed` event
* `layout_and_allocate` - part of the `module_load` event

```
v0.16.1
```

## Output of `uname -a`:

```
Linux Alon-Zivony 6.3.6-microsoft-standard-WSL2 #1 SMP PREEMPT_DYNAMIC Mon Jun 12 11:59:43 IDT 2023 x86_64 x86_64 x86_64 GNU/Linux
```

AlonZivony:

9fc0ae606 **fix(events): change process_execute_failed probes**
e9094a9c1 **feat(events): add probe relevance attribute**

Fix #3356


pkg/events/derive/process_execute_failed.go
```diff
@@ -0,0 +1,145 @@
...
+
+// execFailedGenerator is the object which implement the ProcessExecuteFailed event derivation
+type execFailedGenerator struct {
+	execEndInfo *lru.Cache[int, execEndInfo]
+	baseEvents  *lru.Cache[int, *trace.Event]
+	deriveBase  deriveBase
+}
+
+// newProcessExecuteFailedGenerator initialize a new generator for the ProcessExecuteFailed event.
+func newProcessExecuteFailedGenerator() (*execFailedGenerator, error) {
```
NDStrahilevitz, line 48: Why make this a singleton if it's not publicly accessible? Just construct the struct once outside the returned function and it will live in that scope.

pkg/events/definition_dependencies.go
```diff
@@ -75,19 +77,63 @@ func (d Dependencies) GetCapabilities() Capabilities {
...
 
+// IsRequired determine if the probe is required for the event to work properly
 func (p Probe) IsRequired() bool {
 	return p.required
 }
 
+// IsRelevant determine if the probe should be attached in current environment
+// If a required probe is not relevant in the environment, then the event won't be cancelled as an
+// attempt to attach it won't be initiated in the first place.
+func (p Probe) IsRelevant(osInfo *helpers.OSInfo) (bool, error) {
```
NDStrahilevitz, line 113: nit: IsKernelCompatible

NDStrahilevitz (commented):
Much better IMO. Thanks for addressing.

NDStrahilevitz:
@josedonizetti or @geyslan I think one of you should review this as well due to the new kernel compatibility check.

pkg/events/core.go
```diff
@@ -12935,7 +12936,17 @@ var CoreEvents = map[ID]Definition{
 		version: NewVersion(1, 0, 0),
 		sets:    []string{"proc"},
 		dependencies: Dependencies{
-			ids: []ID{ExecuteFinished, SecurityBprmCredsForExec},
+			ids: []ID{ExecuteFinished, SecurityBprmCredsForExec}, // For kernel version >= 5.8
```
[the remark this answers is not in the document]
AlonZivony: So I added in the events the kernel compatibility to the probes

pkg/events/derive/process_execute_failed.go
AlonZivony: Hmm its just not how it works right now with other functions.
I had initializations from the calling function with other derived events but they were removed.
Anyways, its more comftable API and I don't see a problem with doing it like this.

pkg/events/core.go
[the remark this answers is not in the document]
AlonZivony: You want to add also kernel dependency to event dependencies?

geyslan:
> @josedonizetti or @geyslan I think one of you should review this as well due to the new kernel compatibility check.

@AlonZivony when you move it for review, ping us. 👍🏼

pkg/events/derive/process_execute_failed.go
NDStrahilevitz: If so, then it would also be redundant in them as well (unless they require public global access, not the case here as the singleton isn't exported or used in another internal file).

AlonZivony:
> > @josedonizetti or @geyslan I think one of you should review this as well due to the new kernel compatibility check.
> 
> @AlonZivony when you move it for review, ping us. 👍🏼

This is ready for review :)

geyslan:
> > > @josedonizetti or @geyslan I think one of you should review this as well due to the new kernel compatibility check.
> > 
> > 
> > @AlonZivony when you move it for review, ping us. 👍🏼
> 
> This is ready for review :)

Is this a fix, a feature or a chore? A bit of all? 😀 

If a fix, please open one issue to link it exposing the reasons. Tks.

I'm reviewing it now. 👍🏼

pkg/events/core.go
```diff
@@ -12878,21 +12880,71 @@ var CoreEvents = map[ID]Definition{
...
+		name:     "execute_finished",
+		version:  NewVersion(1, 0, 0),
+		sets:     []string{"proc"},
+		internal: true,
+		dependencies: Dependencies{
+			probes: []Probe{
+				{handle: probes.SyscallExit__Internal, required: true},
+			},
+			tailCalls: []TailCall{
+				{"sys_exit_init_tail", "syscall__execveat__execute_failed", []uint32{uint32(Execve), uint32(Execveat)}},
```
geyslan, line 12895: Could be a leftover?

Kernels:

- `Linux hb0 6.6.10-1-MANJARO #1 SMP PREEMPT_DYNAMIC Fri Jan  5 17:38:36 UTC 2024 x86_64 GNU/Linux`
- `Linux ip-172-31-5-229 5.4.0-137-generic #154-Ubuntu SMP Thu Jan 5 17:03:22 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux`

`sudo ./dist/tracee -e process_execute_failed`

```json
{"level":"fatal","ts":1708628341.161495,"msg":"Tracee runner failed","error":"cmd.Runner.Run: error initializing Tracee: ebpf.(*Tracee).Init: ebpf.(*Tracee).initBPF: ebpf.(*Tracee).populateBPFMaps: failed to initialize tail call: ebpf.(*Tracee).initTailCall: could not get BPF program syscall__execveat__execute_failed: failed to find BPF program syscall__execveat__execute_failed: no such file or directory"}
```

geyslan (changes requested):
LGTM when addressed a few things.

geyslan:
> LGTM when addressed a few things.

@AlonZivony I have this PR #3848 that touches on some parts that you also touched on (e.g. attaching probe). I think it's worth taking a look just to catch on upcoming changes.

tests/e2e-inst-test.sh
```diff
@@ -129,7 +129,7 @@ for TEST in $TESTS; do
         --output option:parse-arguments \
         --log file:$SCRIPT_TMP_DIR/tracee-log-$$ \
         --signatures-dir "$SIG_DIR" \
-        --scope comm=echo,mv,ls,tracee,proctreetester,ping,ds_writer \
+        --scope comm=echo,mv,ls,tracee,proctreetester,ping,ds_writer,process_execute \
```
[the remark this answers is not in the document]
AlonZivony: So the comm is 16 bytes in the kernel, for a string.
So in practice, the comm received is 15 characters - hence the `process_execute`

tests/e2e-inst-signatures/scripts/process_execute_failed.sh
```diff
@@ -0,0 +1,16 @@
...
+
+exit_err() {
+    echo -n "ERROR: "
+    echo $@
+    exit 1
+}
+
+echo "echo hello > /dev/null" > test.sh
+chmod +x test.sh
+ls . > /dev/null # necessary for the execution path to work
```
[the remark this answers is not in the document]
AlonZivony: Just noticed that I get an error about the path executed if I don't use it.
I am not really sure why, but I just kept it.

pkg/events/core.go
AlonZivony: You are right, I changed the name but didn't update it here.

AlonZivony: It didn't pop up in the tests because I forgot to add the e2e to the `pr.yml` file...

AlonZivony:
@geyslan I will give you a notice when the PR is ready again.
I have some issues with older kernels, I hope that I will soon make it work properly.

AlonZivony:
@geyslan ok now it passes all the tests and ready for review :)

tests/e2e-inst-test.sh
```diff
@@ -129,7 +129,7 @@ for TEST in $TESTS; do
         --output option:parse-arguments \
         --log file:$SCRIPT_TMP_DIR/tracee-log-$$ \
         --signatures-dir "$SIG_DIR" \
-        --scope comm=echo,mv,ls,tracee,proctreetester,ping,ds_writer \
+        --scope comm=echo,mv,ls,tracee,proctreetester,ping,ds_writer,process_execute,tracee-ebpf \
```
geyslan, line 132: Why tracee-ebpf is required here?

geyslan (approved):
LGTM

yanivagman (changes requested):
Relying on kernel version to check if a symbol exists or not is not a good approach.
Commits can be backported (e.g. in RHEL), so we need a more robust way for these kind of checks.

AlonZivony:
I opened #3983 to fix it quickly for now, as it is not using any choosing mechanism of probes.
When the fallback mechanism will be available we will be able to change the event to use it.

PR #3883, opened 2024-02-21
helm: remove old webhook config

josedonizetti:
Remove old way of configuring webhook. One should now use the configfile -> https://github.com/aquasecurity/tracee/pull/3832

### DIFF
--- a/deploy/helm/tracee/templates/daemonset.yaml
+++ b/deploy/helm/tracee/templates/daemonset.yaml
@@ -35,10 +35,6 @@ spec:
           args:
             - --config
             - /tracee/config.yaml
-          {{- if .Values.webhook }}
-            - --output
-            - webhook:{{ .Values.webhook }}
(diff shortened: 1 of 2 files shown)


yanivagman (approved):
LGTM

geyslan:
@rscampos would you mind picking this?

geyslan:
@rscampos I'm pushing it, ok?

yanivagman:
@josedonizetti should we keep this PR open?

josedonizetti:
@yanivagman Looking into it.

josedonizetti:
@yanivagman  PR updated. The idea is still valid, the way used to configure webhook here should be avoid, as we have the config file, which we prefer to be used. Though we have zero docs about tracee in k8s in the repo, not sure if we do want it or not, could be a future PR, if you think it makes sense.

Issue #3863, opened 2024-02-14
grpc: create a ListValue type that has an array of event_value


josedonizetti:
This would allow us to remove specific array types like (unit64_array, str_array) etc.

josedonizetti:
created this issue just to keep track, still not sure about this change as I'm not sure how it will look on the json, due to the recursive event value, I'll run a few test and report back here for us to decide @yanivagman

josedonizetti:
https://github.com/aquasecurity/tracee/pull/3888#issuecomment-1959306625

PR #3888, opened 2024-02-21
grpc: add array proto

josedonizetti:
Fix https://github.com/aquasecurity/tracee/issues/3863

The json for an array argument looks like:

```
{
      "name": "interpreter_inode",
      "u_int64": "39488898"
    },
    {
      "name": "interpreter_ctime",
      "u_int64": "1708552191219117250"
    },
    {
      "name": "argv",
      "array": {
        "value": [
          {
            "str": "ip6tables"
          },
          {
            "str": "-w"
          },
          {
            "str": "5"
          },
          {
            "str": "-W"
          },
          {
            "str": "100000"
          },
          {
            "str": "-S"
          },
          {
            "str": "KUBE-PROXY-CANARY"
          },
          {
            "str": "-t"
          },
          {
            "str": "mangle"
          }
        ]
      }
    },
    {
      "name": "interp",
      "str": "/usr/sbin/ip6tables"
    },
    {
      "name": "stdin_type",
      "str": "S_IFCHR"
    },
```

It simplifies a bit, by removing the duplicated array types per type, a little more complicated to create a type due to the recursion -> https://github.com/aquasecurity/tracee/pull/3889/files#diff-d0e942a6b82d5096d1c7e1cd83a4e6ec88bb15eb3f0e017afc1baa9e159c03f9R148-R162  but not too bad

The only part bugging me a bit, is the idea that for a grpc client, there will be an array of EventValue, which has a filed Name, but it is empty when inside the array. 

@yanivagman WDYT? Makes sense?


josedonizetti:
@yanivagman and I spoke about it, and decide to keep with fixed type arrays.

Issue #3891, opened 2024-02-22
Cannot filter on `sys_enter.args.syscall`


oshaked1:
## Description

Filtering on the syscall argument of sys_enter does not work, neither with a syscall name nor with its number.
e.g. `sys_enter.args.syscall=321` or `sys_enter.args.syscall=bpf` do not work.

## Output of `tracee version`:

```
Tracee version: v0.20.0
```

## Output of `uname -a`:

```
Linux ****** 5.15.133.1-microsoft-standard-WSL2 #1 SMP Thu Oct 5 21:02:42 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```

## Additional details 

The problem appears to be in the following check from `pkt/filters/args.go` line 59:

```go
if argName != "syscall" {
    argVal = fmt.Sprint(argVal)
}
```

Conversion of the argument value to string form doesn't happen specifically for the syscall argument.

geyslan:
@oshaked1 thanks for this.

I have an initial solution to make `sys_enter.args.syscall=bpf` asap. Making both (number/name) to work would demand changes in the filter parser.

oshaked1:
Sounds great, filtering on the number is not a priority for me I just mentioned it to clarify that no method worked.

PR #3893, opened 2024-02-22
fix(filters): handle syscall arg

geyslan:
Close: #3891

8995e8dd3 **fix(filters): handle syscall arg**

```
Handle syscall arg either by its name or its number for the events
SysEnter and SysExit.

E.g.:

  - sys_enter.args.syscall=open
  - sys_enter.args.syscall=2

It uses a handler function to be passed to the filter constructor.

This also adds unit tests for Filter().

```

### 2. Explain how to test it

`sudo ./dist/tracee -e sys_enter.args.syscall=bpf`

Trigger some bpf syscall.


josedonizetti (commented):
As this is a specific case, can we add a test? https://github.com/aquasecurity/tracee/blob/main/pkg/filters/args_test.go

pkg/filters/args.go
```diff
@@ -55,10 +58,21 @@ func (filter *ArgFilter) Filter(eventID events.ID, args []trace.Argument) bool {
 		if !found {
 			return false
 		}
-		// TODO: use type assertion instead of string conversion
-		if argName != "syscall" {
-			argVal = fmt.Sprint(argVal)
+
+		argVal = fmt.Sprint(argVal)
```
NDStrahilevitz, line 62: This will break other filters which aren't string based.

NDStrahilevitz (dismissed):

pkg/filters/args.go
```diff
@@ -55,10 +58,21 @@ func (filter *ArgFilter) Filter(eventID events.ID, args []trace.Argument) bool {
...
+		if argName == "syscall" {
+			// TODO: this only handles args set as syscall name, not syscall id.
+			// To handle syscall id, changes are needed in the parser to store
+			// an unique identifier (or syscall name or the syscall id).
+			syscallID, err := strconv.Atoi(argVal.(string))
```
NDStrahilevitz, line 67: The type conversion is unnecessary as you're already working with a string.

pkg/filters/args.go
geyslan: I think I didn't get this. Previously we always assign the `argVal` interface with a string returned from fmt.Sprint() when the argName isn't "syscall". So I assumed that is safe to assign it always.

pkg/filters/args.go
geyslan: the assign above doesn't change the argVal, so the assertion is required.

pkg/filters/args.go
geyslan: @NDStrahilevitz let me know if it still have to be changed.

pkg/filters/args.go
```diff
@@ -55,10 +58,21 @@ func (filter *ArgFilter) Filter(eventID events.ID, args []trace.Argument) bool {
 		if !found {
 			return false
 		}
-		// TODO: use type assertion instead of string conversion
-		if argName != "syscall" {
-			argVal = fmt.Sprint(argVal)
+
+		argVal = fmt.Sprint(argVal)
+		if argName == "syscall" {
```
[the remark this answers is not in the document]
geyslan: argVal is coming from Filter() args parameter, in other words, from the event.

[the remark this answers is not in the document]
AlonZivony: I am not talking about `argVal`.
I am talking about what we compare it to.
I think we should transform the filtered syscall name to a number, instead of transforming the received syscall number to its name.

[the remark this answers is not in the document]
geyslan: @AlonZivony you've opened my eyes. Thanks. Going to push the suggested solution.

pkg/filters/args.go
NDStrahilevitz: This logic is contingent on the fact that we default to string filters for all arguments. In both cases, it's not entirely "correct", but the defaulting also isn't.

yanivagman (approved):
LGTM

PR #3895, opened 2024-02-22
fix: wrong print_mem_dump errors about args filter

### LINKED ISSUE #3781: Warning for each policy not giving filters for print_mem_dump
## Description

For some reason,
Tracee prints a warning for each policy that doesn't give filter for address or symbol in the `print_mem_dump` event.
It is expected that if a policy chose the event and didn't give a filter, there will be a warning.
But even policies that don't choose the event get a warning if at least one policy chose it.
The errors given:
```
{"level":"warn","ts":1703767495.6483238,"msg":"Memory dump","error":"ebpf.(*Tracee).triggerMemDump: policy 3: no address or symbols were provided to print_mem_dump event. please provide it via -e print_mem_dump.args.address=<hex address>, -e print_mem_dump.args.symbol_name=<owner>:<symbol> or -e print_mem_dump.args.symbol_name=<symbol> if specifying a system owned symbol"}
{"level":"warn","ts":1703767495.6483612,"msg":"Memory dump","error":"ebpf.(*Tracee).triggerMemDump: policy 4: no address or symbols were provided to print_mem_dump event. please provide it via -e print_mem_dump.args.address=<hex address>, -e print_mem_dump.args.symbol_name=<owner>:<symbol> or -e print_mem_dump.args.symbol_name=<symbol> if specifying a system owned symbol"}
{"level":"warn","ts":1703767495.6483715,"msg":"Memory dump","error":"ebpf.(*Tracee).triggerMemDump: policy 5: no address or symbols were provided to print_mem_dump event. please provide it via -e print_mem_dump.args.address=<hex address>, -e print_mem_dump.args.symbol_name=<owner>:<symbol> or -e print_mem_dump.args.symbol_name=<symbol> if specifying a system owned symbol"}
{"level":"warn","ts":1703767495.6483796,"msg":"Memory dump","error":"ebpf.(*Tracee).triggerMemDump: policy 0: no address or symbols were provided to print_mem_dump event. please provide it via -e print_mem_dump.args.address=<hex address>, -e print_mem_dump.args.symbol_name=<owner>:<symbol> or -e print_mem_dump.args.symbol_name=<symbol> if specifying a system owned symbol"}
{"level":"warn","ts":1703767495.6483893,"msg":"Memory dump","error":"ebpf.(*Tracee).triggerMemDump: policy 1: no address or symbols were provided to print_mem_dump event. please provide it via -e print_mem_dump.args.address=<hex address>, -e print_mem_dump.args.symbol_name=<owner>:<symbol> or -e print_mem_dump.args.symbol_name=<symbol> if specifying a system owned symbol"}
```

This is the relevant code that produce these warnings:
```Go
// triggerMemDump is used by a Uprobe to trigger an eBPF program
// that prints the first bytes of requested symbols or addresses
func (t *Tracee) triggerMemDump(event trace.Event) []error {
	if _, ok := t.eventsState[events.PrintMemDump]; !ok {
		return nil
	}

	errs := []error{}

	for p := range t.config.Policies.Map() {
		printMemDumpFilters := p.ArgFilter.GetEventFilters(events.PrintMemDump)
		if len(printMemDumpFilters) == 0 {
			errs = append(errs, errfmt.Errorf("policy %d: no address or symbols were provided to print_mem_dump event. "+
				"please provide it via -e print_mem_dump.args.address=<hex address>"+
				", -e print_mem_dump.args.symbol_name=<owner>:<symbol> or "+
				"-e print_mem_dump.args.symbol_name=<symbol> if specifying a system owned symbol", p.ID))

			continue
		}
```

```
v0.19.0
```

## Output of `uname -a`:

```
(paste your output here)
```

AlonZivony:

Avoid errors about missing print_mem_dump args filters when the policy does not choose the event.
This was caused when one policy chose it, but others didn't.
Fix #3781

### DIFF
--- a/pkg/ebpf/tracee.go
+++ b/pkg/ebpf/tracee.go
@@ -1587,11 +1587,27 @@ func (t *Tracee) triggerMemDump(event trace.Event) []error {
 		return nil
 	}
 
-	errs := []error{}
+	var errs []error
 
-	// TODO: consider to iterate over given policies when policies are changed
-	for it := t.config.Policies.CreateAllIterator(); it.HasNext(); {
+	// We want to use the policies of relevant to the triggering event
+	policies, err := policy.Snapshots().Get(event.PoliciesVersion)
+	if err != nil {
+		logger.Errorw("Error getting policies for print_mem_dump event", "error", err)
+		// For fallback, try to use latest polcies
+		policies, err = policy.Snapshots().GetLast()
+		if err != nil {
+			logger.Errorw("Error getting last snapshots policies for print_mem_dump event", "error", err)
+			// Last resort is to use the policies from the config
+			policies = t.config.Policies
+		}
+	}
+	for it := policies.CreateAllIterator(); it.HasNext(); {
 		p := it.Next()
+		// This might break in the future if PrintMemDump will become a dependency of another event.
+		_, isChosen := p.EventsToTrace[events.PrintMemDump]
+		if !isChosen {
+			continue
+		}
 		printMemDumpFilters := p.ArgFilter.GetEventFilters(events.PrintMemDump)
 		if len(printMemDumpFilters) == 0 {
 			errs = append(errs, errfmt.Errorf("policy %d: no address or symbols were provided to print_mem_dump event. "+


pkg/ebpf/tracee.go
```diff
@@ -1572,6 +1572,11 @@ func (t *Tracee) triggerMemDump(event trace.Event) []error {
 
 	// TODO: consider to iterate over given policies when policies are changed
 	for p := range t.config.Policies.Map() {
+		// This might break in the future if PrintMemDump will become a dependency of another event.
+		_, isChosen := p.EventsToTrace[events.PrintMemDump]
```
yanivagman, line 1576: @geyslan correct me if I'm wrong, but I think this is not the best way to handle this issue.
Event states should be used instead.

geyslan: You're right, Yaniv. I'm attempting to change how EventStates are managed with #3848. For these peculiar events (PrintMemDump, HookedSyscall, ...) I've put comments for further efforts.

https://github.com/aquasecurity/tracee/pull/3848/files#diff-93a0a7535081b0454fec716306a8f60ab9de4de5d300d7799d8792eb0865e87cR218-R233

geyslan: That PR takes care of keeping the states per policies version:

image

In future efforts we'll have a policyManager (system wise = all policies versions in activity) aware of all events states.

The most relevant discussion for now is whether this type of event like PrintMemDump (initial) will only be necessary initially. If so, I would consider detaching their logic from Policies, because they fall into categories of option flags and not events to be managed by policies and so on. WDYT?

AlonZivony: Anyways, this PR is a fix of an issue we now have.
@yanivagman I see no reason not to fix it like this for now, and when we have the policies events states use it instead.

geyslan:
@AlonZivony I remembered already merging this but I was wrong, do you mind rebasing it and ping me again for review? Tks.

geyslan (approved):
LGTM

PR #3896, opened 2024-02-22
fix(events): fix ftrace_hook

OriGlassman:

We need to count how many ftrace based hooks will be placed on each symbol. eventsState may contain duplicate events due to dependencies. To get the real count, we consider the program name and the prob type. Furthermore, added logic that addresses the situation where there may be multiple k[ret]probes from a single probe request (due to multiple symbols at different locations).

### 2. Explain how to test it

./tracee -e=ftrace_hook

### DIFF
--- a/pkg/ebpf/tracee.go
+++ b/pkg/ebpf/tracee.go
@@ -1555,6 +1555,23 @@ func (t *Tracee) getSelfLoadedPrograms(kprobesOnly bool) map[string]int {
 		logger.Debugw("self loaded program", "event", event, "program", program)
 	}
 
+	type probeMapKey struct {
+		programName string
+		probeType   probes.ProbeType
+	}
+	// We need to count how many ftrace based hooks will be placed on each symbol.
+	// eventsState may contain duplicate events due to dependencies.
+	// To get the real count, we consider the program name and the prob type.
+	// For example:
+	// event state may contain the following entries:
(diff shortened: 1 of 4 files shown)


geyslan:
@OriGlassman I have this PR https://github.com/aquasecurity/tracee/pull/3848 related to Events States. I think it's worth taking a look just to catch on upcoming changes.

pkg/ebpf/tracee.go
```diff
@@ -1576,6 +1592,18 @@ func (t *Tracee) getSelfLoadedPrograms(kprobesOnly bool) map[string]int {
...
+				}
+
+				_, found := uniqueHooksMap[key]
+				if found { // Already counted this hook
+					continue
+				}
+
+				uniqueHooksMap[key] = struct{}{}
+
 				log(definition.GetName(), p.GetProgramName())
```
AlonZivony, line 1607: I know this is not part of the current code, but shouldn't we use the `logger.Infow` here?

AlonZivony (approved):
LGTM
Had only one comment but its out of the PR scope

pkg/ebpf/tracee.go
OriGlassman: This is a change done by a previous maintainer so not sure the motivation

yanivagman (approved):
LGTM

PR #3898, opened 2024-02-23
Move `AttachProbes` call out of initBPF

### LINKED ISSUE #3900: Refactor/relocate attach/detach probes to after tracee main loop has already started
This is required to start to have go routines clean state even without policies set.

geyslan:
Close #3900

On top of #3848

a8833424e **chore: refactor equality computation**
64ab539e1 **chore: extract probe attach to after main loop**

a8833424e **chore: refactor equality computation**

```
This moves the equality logic to equality.go, and uses Generics to
deduplicate the code.
```

64ab539e1 **chore: extract probe attach to after main loop**

```
Decouple initialization of Policies (eBPF) and sequential call to
AttachProbes() from main loop.

This also:

- Refactor creation/population and pruning of Policies related bpf maps.
- Enforces the use of inner objects when in a controlled (locked) state.
- Removes Policies.SetVersion().
```

### 3. Other comments

**_Consider for review only the commits above._**

### DIFF
--- a/pkg/cmd/cobra/cobra.go
+++ b/pkg/cmd/cobra/cobra.go
@@ -205,35 +205,6 @@ func GetTraceeRunner(c *cobra.Command, version string) (cmd.Runner, error) {
 		return runner, errors.New("policy and event flags cannot be used together")
 	}
 
-	// Try to get policies from kubernetes CRD, policy files and CLI in that order
-
-	var k8sPolicies []v1beta1.PolicyInterface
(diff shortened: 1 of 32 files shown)


geyslan:
There's so much intertwined work that, for now, I'll be tackling this in https://github.com/aquasecurity/tracee/pull/3912

geyslan:
Closing due to possible changes to the overall architecture.

PR #3899, opened 2024-02-25
fix: improve performance of magic_write event

yanivagman:
Currently, when magic_write event is selected on its own it introduces a very high overhead on the system. This is due to the reason that the logic used for magic_write event is mixed with the logic used for vfs_write(v) events and capture write, thus a lot of extra unnecessary calculations are being performed.

Fix that issue by seperating the logic used for magic_write from any other logic.


NDStrahilevitz (approved):
Logic looks correct overall. Would be good to include the performance improvement and the benchmark method in the commit or PR description before merging.

PR #3901, opened 2024-02-28
ebpf: don't send magic_write with zero bytes

yanivagman:


AlonZivony (approved):
LGTM

Issue #559, opened 2021-02-17
Use bpf_probe_read_user and bpf_probe_read_kernel helpers


yanivagman:
On newer kernels (>=5.5), there are four new bpf helper functions which are safer to use. These are: 
bpf_probe_read_user(),  bpf_probe_read_user_str(), bpf_probe_read_kernel(), bpf_probe_read_kernel_str()
(For more info about these helpers also see https://github.com/iovisor/bcc/blob/master/docs/reference_guide.md#data)

We should start using these functions instead of bpf_probe_read\[\_str\]() functions as libbpf will automatically fall back to bpf_probe_read\[\_str\]() if bpf_probe_read_{kernel, user}\[\_str\]() are not supported on older kernels, so it's always safe to use bpf_probe_read_{kernel, user}\[\_str\]() and not worry about kernel version compatibility

cdelzotti:
Hello ! Is this issue still relevant ? If so, is it possible to work on it ?

yanivagman:
Hello! This was partially implemented as can be seen here:
https://github.com/aquasecurity/tracee/blob/main/pkg/ebpf/c/tracee.bpf.c#L324
However, we still don't make use of xxx_str() functions, and xxx_kernel[_str].

So yes! You can go ahead with this one.
I think the challenges here are to make sure things work as expected with the new helpers, and distinguishing user pointers from kernel pointers.

cdelzotti:
Sorry for the long delay, I got caught in school projects and exams. I see the problem and I understand what to do, but I don't see how to verify what I'm doing. Tests seem to have the same outputs whether I use `kernel_str()` or `user_str()`. How can I be sure do do not make a mistake ?

PR #3906, opened 2024-03-04
Improve bpf policies config access

geyslan:

f1b2d39e **chore: reduce bpf map lookups for policies config**

```
Instead of always retrieving the policies config map version in the
outer and inner maps, create a holder in the event_data and update it
only when config entry version changes.
```

### DIFF
--- a/pkg/ebpf/c/common/context.h
+++ b/pkg/ebpf/c/common/context.h
@@ -113,6 +113,7 @@ statfunc proc_info_t *init_proc_info(u32 pid, u32 scratch_idx)
 statfunc void init_task_info_scratch(u32 tid, scratch_t *scratch)
 {
     scratch->task_info.syscall_traced = false;
+    scratch->task_info.policies_version = 0;
     scratch->task_info.recompute_scope = true;
     scratch->task_info.container_state = CONTAINER_UNKNOWN;
     bpf_map_update_elem(&task_info_map, &tid, &scratch->task_info, BPF_NOEXIST);
@@ -192,7 +193,6 @@ statfunc int init_program_data(program_data_t *p, void *ctx)
     // in some places we don't call should_trace() (e.g. sys_exit) which also initializes
     // matched_policies. Use previously found scopes then to initialize it.
     p->event->context.matched_policies = p->task_info->matched_scopes;
-    p->event->context.policies_version = p->config->policies_version;
 
     // check if we need to recompute scope due to context change
     if (context_changed(&p->task_info->context, &p->event->context.task))
@@ -209,6 +209,20 @@ statfunc int init_program_data(program_data_t *p, void *ctx)
     }
 
 out:
+    if (unlikely(p->event->context.policies_version != p->config->policies_version)) {
+        // copy policies_config to event data
+        long ret = bpf_probe_read_kernel(
+            &p->event->policies_config, sizeof(policies_config_t), &p->config->policies_config);
+        if (unlikely(ret != 0))
+            return 0;
+
+        p->event->context.policies_version = p->config->policies_version;
+    }
+    if (p->task_info->policies_version != p->event->context.policies_version) {
+        p->task_info->policies_version = p->event->context.policies_version;
+        p->task_info->recompute_scope = true;
+    }
+
     if (container_lookup_required) {
         u32 cgroup_id_lsb = p->event->context.task.cgroup_id;
         u8 *state = bpf_map_lookup_elem(&containers_map, &cgroup_id_lsb);
--- a/pkg/ebpf/c/common/filtering.h
+++ b/pkg/ebpf/c/common/filtering.h
@@ -9,7 +9,6 @@
 
 // PROTOTYPES
 
-statfunc policies_config_t *get_policies_config(program_data_t *);
 statfunc void *get_filter_map(void *, u16);
 statfunc u64 uint_filter_range_matches(u64, void *, u64, u64, u64);
 statfunc u64 binary_filter_matches(u64, void *, proc_info_t *);
@@ -51,23 +50,6 @@ statfunc u64 should_submit(u32, event_data_t *);
 #define FILTER_MIN_NOT_SET ULLONG_MAX
 
 // FUNCTIONS
-
-statfunc policies_config_t *get_policies_config(program_data_t *p)
-{
-    u16 version = p->event->context.policies_version;
-
-    if (likely(version == p->config->policies_version)) {
-        return &p->config->policies_config;
-    } else {
-        policies_config_map_t *policies_config_map;
-        policies_config_map = bpf_map_lookup_elem(&policies_config_version, &version);
-        if (unlikely(policies_config_map == NULL))
-            return NULL;
-
-        u32 zero = 0;
-        return bpf_map_lookup_elem(policies_config_map, &zero);
-    }
-}
 
 // get_filter_map returns the filter map for the given version and outer map
 statfunc void *get_filter_map(void *outer_map, u16 version)
@@ -214,13 +196,7 @@ statfunc u64 compute_scopes(program_data_t *p)
         return 0;
 
     proc_info_t *proc_info = p->proc_info;
-
-    policies_config_t *policies_cfg = get_policies_config(p);
-    if (unlikely(policies_cfg == NULL)) {
-        // policies_config should be set by userland
-        return 0;
-    }
-
+    policies_config_t *policies_cfg = &p->event->policies_config;
     u64 res = ~0ULL;
 
     //
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -542,9 +542,7 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
 
     // Update the process tree map (filter related) if the parent has an entry.
 
-    policies_config_t *policies_cfg = get_policies_config(&p);
-    if (unlikely(policies_cfg == NULL))
-        return 0;
+    policies_config_t *policies_cfg = &p.event->policies_config;
 
     if (policies_cfg->proc_tree_filter_enabled_scopes) {
         u16 version = p.event->context.policies_version;
@@ -1025,7 +1023,6 @@ int uprobe_lkm_seeker_submitter(struct pt_regs *ctx)
 
     // Uprobes are not triggered by syscalls, so we need to override the false value.
     p.event->context.syscall = NO_SYSCALL;
-    p.event->context.policies_version = p.config->policies_version;
     p.event->context.matched_policies = ULLONG_MAX;
 
     u32 trigger_pid = bpf_get_current_pid_tgid() >> 32;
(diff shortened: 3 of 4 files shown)


geyslan:
Green in 1402 (E2E).

pkg/ebpf/c/common/context.h
```diff
@@ -215,6 +218,33 @@ statfunc int init_program_data(program_data_t *p, void *ctx)
 }
 // clang-format on
 
+statfunc int setup_policies(program_data_t *p)
+{
+    if (p->event->context.policies_version)
+        return 0; // already set
```
yanivagman, line 224: When will this happen?

pkg/ebpf/c/common/context.h
```diff
@@ -215,6 +218,33 @@ statfunc int init_program_data(program_data_t *p, void *ctx)
...
+
+    u32 zero = 0;
+    policies_config_t *policies_config;
+    policies_config = bpf_map_lookup_elem(policies_config_map, &zero);
+    if (unlikely(policies_config == NULL))
+        return 1;
+
+    // copy policies_config to event
+    if (bpf_probe_read(&p->event->policies_config, sizeof(policies_config_t), policies_config) != 0)
+        return 1;
```
yanivagman, line 240: Why do we need all of this?
Can't we just use p->config->policies_config like we did in get_policies_config()?

geyslan: In `get_policies_config()` we were not caching the policies_config, only short-circuiting to use the one in p->config, but if a different version comes, we were using the pointer of the map directly.

Now, we're caching it in `p->event->policies_config` avoiding lookups.

geyslan: And we can't rely on `p->config->policies_config` since it can change meanwhile (new versions).

geyslan: This was solved in the last pushes. Thanks.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -543,8 +543,6 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
     // Update the process tree map (filter related) if the parent has an entry.
 
     policies_config_t *policies_cfg = get_policies_config(&p);
```
yanivagman, line 545: since `get_policies_config()` function is now one line, I think it will be simpler to just put it inline

pkg/ebpf/c/common/filtering.h
```diff
@@ -214,13 +202,7 @@ statfunc u64 compute_scopes(program_data_t *p)
         return 0;
 
     proc_info_t *proc_info = p->proc_info;
-
     policies_config_t *policies_cfg = get_policies_config(p);
```
yanivagman, line 205: since `get_policies_config()` function is now one line, I think it will be simpler to just put it inline

pkg/ebpf/c/common/context.h
```diff
@@ -129,6 +130,23 @@ statfunc task_info_t *init_task_info(u32 tid, u32 scratch_idx)
     return bpf_map_lookup_elem(&task_info_map, &tid);
 }
 
+statfunc int init_policies_config(program_data_t *p)
+{
+    u16 version = p->config->policies_version;
+
+    // copy policies_config to event data
+    long ret = bpf_probe_read(
```
yanivagman, line 138: We should start using `bpf_probe_read_kernel` and `bpf_probe_read_user` in new code that we add.
In the future we will need to fully tackle #559

pkg/ebpf/c/common/context.h
```diff
@@ -129,6 +130,23 @@ statfunc task_info_t *init_task_info(u32 tid, u32 scratch_idx)
...
+        &p->event->policies_config, sizeof(policies_config_t), &p->config->policies_config);
+    if (unlikely(ret != 0))
+        return 1;
+
+    // finally set the version and set recompute_scope to true
+    p->event->context.policies_version = version;
```
yanivagman, line 144: No need for the version variable since it is only used here
```suggestion
    p->event->context.policies_version = p->config->policies_version;
```

pkg/ebpf/c/common/context.h
```diff
@@ -209,6 +226,11 @@ statfunc int init_program_data(program_data_t *p, void *ctx)
     }
 
 out:
+    if (p->event->context.policies_version == 0) {
```
yanivagman, line 229: Why do we have this condition?
Where did we ever initialized `p->event->context.policies_version` before?

pkg/ebpf/c/common/context.h
geyslan: You're right, it's a bad habit.

pkg/ebpf/c/common/context.h
geyslan: Since there are cases of calling `init_program_data()` (reentrant) and we already have set the policy_version and config via `init_policies_config()`. I've put this print to check such cases:

image

Ran: `sudo ./dist/tracee -e sched_process_exec`

And got this output:

```
...
          docker-616209  [011] ...2. 54180.265695: bpf_trace_printk: 
policies_version: 1
p->event->context.eventid 714
p->event->context.syscall 35

           <...>-616213  [031] ...2. 54180.265705: bpf_trace_printk: 
policies_version: 1
p->event->context.eventid 714
p->event->context.syscall 202

          docker-616215  [030] ...2. 54180.265709: bpf_trace_printk: 
policies_version: 1
p->event->context.eventid 0
p->event->context.syscall 202

           sleep-616207  [000] ...2. 54180.271460: bpf_trace_printk: 
policies_version: 1
p->event->context.eventid 714
p->event->context.syscall 231

     cpuUsage.sh-616201  [008] ...2. 54180.271677: bpf_trace_printk: 
policies_version: 1
p->event->context.eventid 0
p->event->context.syscall 56

     cpuUsage.sh-616216  [000] ...21 54180.271936: bpf_trace_printk: 
policies_version: 1
p->event->context.eventid 714
p->event->context.syscall 59
...
```

So I assume that it's working as a short-circuit.

yanivagman: I think this is happening not due to it being reentrant, but because p->event is received from index 0 of a percpu map, which may have already been initialized from a previous event.

Actually this is even worse - after setting policies_version to 1, we will never be able to enter again the config init, thus we will never get a newer policy version

geyslan: You're right. I had concerns about tail calls calling again init_program_data, but it's not the case since they make use of `init_tailcall_program_data()` instead.

pkg/ebpf/c/common/context.h
```diff
@@ -209,6 +224,9 @@ statfunc int init_program_data(program_data_t *p, void *ctx)
     }
 
 out:
+    if (unlikely(init_policies_config(p) != 0))
```
yanivagman, line 227: Not many lines in this function and it is called once, so maybe we should put it inline to simplify?

pkg/ebpf/c/common/context.h
```diff
@@ -129,6 +130,21 @@ statfunc task_info_t *init_task_info(u32 tid, u32 scratch_idx)
...
+{
+    // copy policies_config to event data
+    long ret = bpf_probe_read_kernel(
+        &p->event->policies_config, sizeof(policies_config_t), &p->config->policies_config);
+    if (unlikely(ret != 0))
+        return 1;
+
+    // finally set the version and set recompute_scope to true
+    p->event->context.policies_version = p->config->policies_version;
+    p->task_info->recompute_scope = true;
```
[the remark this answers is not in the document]
geyslan: In this case, we can assume that's the condition for the entire logic, right? E.g.:

image

Raising the change.

pkg/ebpf/c/common/context.h
```diff
@@ -209,6 +208,18 @@ statfunc int init_program_data(program_data_t *p, void *ctx)
     }
 
 out:
+    if (p->event->context.policies_version != p->config->policies_version) {
```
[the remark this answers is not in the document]
geyslan: Actually the check is against event's policies_version and config's policies_version. `p->config` is always fetched: line [155](https://github.com/aquasecurity/tracee/pull/3906/files/f1b2d39ee0c6393818c70abf96cda272a87f30d6#diff-2d5cf85fbb13b213de3c3fc135cc46d3cc2c30d7436ba65c232991acf47d39afR155).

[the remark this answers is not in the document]
yanivagman: But where did we set p->event.context.policies_version?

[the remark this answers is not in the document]
geyslan: I assume we don't need pre-assign it, since BPF_MAP_TYPE_ARRAY and BPF_MAP_TYPE_PERCPU_ARRAY are pre-allocated and zero-initialised when created.

I've made a small test to make sure:

```diff
 out:
+    if (p->event->context.policies_version == 0 || p->config->policies_version == 0)
+        bpf_printk("\nevent->policies_version is %d\n"
+                     "config->policies_version is %d\n"
+                     "for cpu id %d\n",
+                        p->event->context.policies_version,
+                        p->config->policies_version,
+                        p->event->context.processor_id);
+
     if (p->event->context.policies_version != p->config->policies_version) {
         // copy policies_config to event data
```

```
❯ sudo timeout 30 cat /sys/kernel/debug/tracing/trace_pipe > trace_output; grep -c bpf_trace_printk trace_output    
32
```

```
❯ nproc
32
```

All prints show a percpu event entry with a stale (in this case, zero-initialised) policies_version.

```
❯ head trace_output 
            code-1545838 [018] ...2. 115176.478099: bpf_trace_printk: 
event->policies_version is 0
config->policies_version is 1
for cpu id 18

            code-1566656 [026] ...21 115176.478513: bpf_trace_printk: 
event->policies_version is 0
config->policies_version is 1
for cpu id 26
```

So I believe we can rely on this array bpf map property and only fetch a new policies_config when we have a difference between event->policies_version and config->policies_version. WDYT?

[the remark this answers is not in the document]
yanivagman: Ok, I think I see what you mean. We know that event_data_map will return a zero intialized struct on the first time we fetch it, and we know that policy version is set to 1 at start (right?), then we perform the first update.
The next events that follow are not going to touch these fields, and we can assume they are valid as long as policy version haven't changed. Is that correct?

yanivagman (approved):
LGTM

pkg/ebpf/c/common/context.h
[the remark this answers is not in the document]
yanivagman: Ok, maybe one last concern.
Consider the following scenario: 
1. Current policies version is 2
2. Current task scope was computed with policies version 1, and recompute_scope is false
3. The previous event already updated `p->event->context.policies_version` so it is now set to 2

Now since `p->event->context.policies_version` = `p->config->policies_version` we will never set recompute_scope of the current task to 2. Isn't that right?
If so, I think we should also have policies_version as part of task_info, so we will know to which policies_version the matched_scopes value is relevant, and add the following lines after this block:

```
if (p->event->context.policies_version != p->config->policies_version) {
    // copy policies_config to event data
    long ret = bpf_probe_read_kernel(
        &p->event->policies_config, sizeof(policies_config_t), &p->config->policies_config);
    if (unlikely(ret != 0))
        return 0;

    // finally update the version and set recompute_scope to true
    p->event->context.policies_version = p->config->policies_version;
}
if (p->task_info->policies_version != p->event->context.policies_version)
    p->task_info->recompute_scope = true;
```

WDYT?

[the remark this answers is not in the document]
geyslan: Per event context was dealt but not for task context change. Good catch. Raising an update.

PR #3913, opened 2024-03-10
Change inotify_watch event to security_path_notify

oshaked1:

Changed inotify_watch event to security_path_notify.
inotify_watch used the inotify_find_inode kernel function which is static and cannot be hooked on all kernels.
This replacement event hooks security_path_notify instead (using a kprobe), which is always available.
This function also handles notify requests from dnotify and fanotify, which are used for similar purposes.
The new event also includes the mask and object type fields,
which can be used to determine the exact characteristics of the watch.

### 2. Explain how to test it

Run tracee with `--events security_path_notify`, generate the event by registering a FS watch using any of the `dnotify`, `inotify` or `fanotify` APIs.


NDStrahilevitz:
@yanivagman Are we ok with straight up replacing the event and breaking users? I think we should start adding new events and deprecating older versions if needed instead as a practice.
Alternatively, we could switch the probes in the event to the new one, and not touch the event at all, besides adding the new argument.

oshaked1:
> > @yanivagman Are we ok with straight up replacing the event and breaking users? I think we should start adding new events and deprecating older versions if needed instead as a practice. Alternatively, we could switch the probes in the event to the new one, and not touch the event at all, besides adding the new argument.
> 
> You are probably right. We should keep the old event, and mark him as deprecated somehow. Then we can in a future major release remove at once all deprecated mark events.

I added back the old event. How should it be marked as deprecated?

yanivagman:
Since we're still in v0.x.y and the event is very specific and still new, I'm ok with removing it. In the future we should add the mechanism to mark events as deprecated

AlonZivony (approved):
LGTM
A very high quality code :)

pkg/events/parse_args.go
```diff
@@ -280,6 +280,19 @@ func ParseArgs(event *trace.Event) error {
...
+				fsNotifyMaskArgument := helpers.ParseFsNotifyMask(mask)
+				parseOrEmptyString(maskArg, fsNotifyMaskArgument, nil)
+			}
+		}
+		if objTypeArg := GetArg(event, "obj_type"); objTypeArg != nil {
+			if objType, isUint := objTypeArg.Value.(uint32); isUint {
+				objTypeArgument, err := helpers.ParseFsNotifyObjType(uint64(objType))
+				parseOrEmptyString(objTypeArg, objTypeArgument, err)
+			}
+		}
```
yanivagman, line 295: Please note that if you use a threat detection event which is based on this event you should not rely on the parsed arguments values but on the raw arguments values

tests/e2e-inst-signatures/scripts/fsnotify_tester.c
```diff
@@ -0,0 +1,117 @@
...
+    if (dir_result != 0 && errno != EEXIST) {
+        perror("mkdir");
+        cleanup_and_exit(EXIT_FAILURE);
+    }
+}
+
+static void handle_dnotify_event(int sig, siginfo_t *si, void *ucontext)
+{
+    return;
+}
```
yanivagman, line 38: What's the use of that?

yanivagman (changes requested):

tests/e2e-inst-signatures/scripts/fsnotify_tester.c
oshaked1: Forgot to remove it, fixed.

yanivagman (approved):
LGTM

Issue #3914, opened 2024-03-12
Failed to find a process in proc_info_map


yanivagman:
## Description

In environments with a big amount of CPUs (96 CPUs) we got the following warnings:

"{"level":"warn","ts":1695678822.5690784,"msg":"Failed to find a map element","id":2,"type":"BPF_LOG_ID_MAP_LOOKUP_ELEM","ret":0,"cpu":82,"file":"./pkg/ebpf/c/tracee.bpf.c","line":603,"count":1}

"{"level":"warn","ts":1710099814.7776074,"msg":"Failed to find a map element","id":2,"type":"BPF_LOG_ID_MAP_LOOKUP_ELEM","ret":0,"cpu":70,"file":"./pkg/ebpf/c/tracee.bpf.c","line":695,"count":1}

These warnings then repeat themselves a lot of times.

## Output of `tracee version`:

```
v0.11.1
```

## Output of `uname -a`:

```
(paste your output here)
```

## Additional details 
It seems that in such environments the current map size of 10240 entries is not enough, and gets filled quickly.
A quick mitigation then will be to set a bigger map size.

The root cause of this issue is that we assume that if task_info exists for some task in the task_info_map, then also the proc_info of the process to which the task belongs to also has an entry in the map, but that is not the case.
So the real fix to this issue will be to change all the places that try to get an entry from proc_info_map and make this assumption. Instead, we should reinitialize proc_info_t with the matching pid. One problem to implement the reinitialization is that we already lost some of the information like if it was a "new process", the binary name and for which scopes this process should be followed (used by the follow filter).

PR #3918, opened 2024-03-14
Fix proc info lru

yanivagman:

    fix: always initialize proc_info map
    
    In environments with a high amount of CPUs the current proc_info map
    size of 10240 entries might not be enough, and gets filled quickly.
    In these cases we got an error of missing proc_info entry.
    
    A quick mitigation then will be to set a bigger map size.
    
    The root cause of this issue is that we assume that if task_info exists
    for some task in the task_info_map, then also the proc_info of the
    process to which the task belongs to also has an entry in the map,
    but that is not the case.
    
    To fix this issue, add proc_info lookup and initialization in
    init_program_data() function. In addition, init proc_info of the
    relevant pid in case it was not found in other places.
    
    Fix: #3914

### 2. Explain how to test it

To test, I changed proc_info_map size to 50, then reproduced the issue.
Then see that the issue is fixed with this change

### DIFF
--- a/pkg/ebpf/c/common/context.h
+++ b/pkg/ebpf/c/common/context.h
@@ -10,7 +10,10 @@
 // PROTOTYPES
 
 statfunc int init_context(void *, event_context_t *, struct task_struct *, u32);
-statfunc task_info_t *init_task_info(u32, u32, u32);
+statfunc void init_proc_info_scratch(u32, scratch_t *);
+statfunc proc_info_t *init_proc_info(u32, u32);
+statfunc void init_task_info_scratch(u32, scratch_t *);
+statfunc task_info_t *init_task_info(u32, u32);
 statfunc bool context_changed(task_context_t *, task_context_t *);
 statfunc int init_program_data(program_data_t *, void *);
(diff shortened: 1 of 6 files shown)


geyslan (approved):
Looks great!

pkg/ebpf/c/common/context.h
```diff
@@ -160,16 +170,21 @@ statfunc int init_program_data(program_data_t *p, void *ctx)
 
     bool container_lookup_required = true;
 
-    p->task_info = bpf_map_lookup_elem(&task_info_map, &p->event->context.task.host_tid);
+    u32 host_pid = p->event->context.task.host_pid;
+    p->proc_info = bpf_map_lookup_elem(&proc_info_map, &host_pid);
+    if (unlikely(p->proc_info == NULL)) {
+        p->proc_info = init_proc_info(host_pid, p->scratch_idx);
```
geyslan, line 176: As I didn't detect the initialisation of `p->scratch_idx` can we assume that it is always 0 here, the default value?

geyslan: Yes, we can, since we initialise p as `program_info_t p = {};` in all programs.

PR #3920, opened 2024-03-14
chore: use map instead of stack to store task_info

yanivagman:

Currently we use the stack to initialize task_info in the sched_process_fork program. Since task_info_t has a size of ~200 bytes, we should opt to put it on a map, and not on the stack.


geyslan (approved):
LGTM

Issue #2847, opened 2023-03-12
execve nil argv, envp


OriGlassman:
## Description
execve that fails seems to return with nil argv and envp, even though they aren't really empty.

![image](https://user-images.githubusercontent.com/39296766/224556452-4bc46a2d-504c-4a9f-8b8d-1732e2962577.png)

## Output of `tracee -v`:

```
Tracee version "v0.12.0-16-ge01881a2"
```

## Output of `uname -a`:

```
Linux Ori-Glassman 5.19.0-35-generic #36~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Fri Feb 17 15:17:25 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
```

yanivagman:
This is the expected behavior for failed execve.
We submit execve events on syscall entry since after execve replaces the binary image, pointers become invalid, and this include the argv and envp pointers. Only for failed exec we ALSO submit the event on syscall exit, so it is possible to see the reason for failure through the return code. So if you look one line above the failed exec line in the output, you will be able to see the matching argv and envp pointers.

PR #3922, opened 2024-03-19
fix: show argv on failed execve events

yanivagman:

Failed execve calls never showed argv nor envp.
That is since it was handled by the generic syscall handler, which doesn't handle string array arguments.

Fix by adding a dedicated handler at execve(at) exit.

### 3. Other comments

Fix: #2847


AlonZivony (commented):
The code itself looks great.
I think we should add an e2e test for it to make sure that it works well as intended. WDYT?

yanivagman:
> The code itself looks great. I think we should add an e2e test for it to make sure that it works well as intended. WDYT?

Agree. Let's merge this PR since it is a required fix and I tested it manually.
We are starting an effort to improve functional tests in Tracee for all the events originated from BPF code. We should add this test there as well

AlonZivony (approved):

PR #3923, opened 2024-03-20
Optimize init program

yanivagman:

    perf: use lazy program data initialization
    
    We currently initialize all task context related fields on the beginning
    of our BPF programs. Reading all the relevant fields takes some CPU
    cycles although these fields are not always used. In many cases, for
    example when filters are applied, the event is not sent to userspace and
    is just dropped without any side effects.
    
    Reading only the essential fields and leaving all other fields
    initialization to the submit stage will avoid wasting these cycles, and
    will also allow us to perform some of the event enrichment in userspace
    instead (not part of this PR).
    
    To do this change, we have to remove the matched_policies which were
    cached in task_info since we can't check if "context changed" anymore.
    Considering that the average user will only use one or two scope filters
    (e.g. container id or binary name), computing the scope on every run is
    not a big overhead - running BPF statistics with and without the
    change verified this is indeed the case, and with the change the
    performance is even slightly better.
    
    Although this change does not introduce a visible performance gain for
    most of the events, it is clear for some other specific event that it
    does. For example, the hidden_inode event which attach a program to the
    filldir64 function has a 50% performance gain.

### DIFF
--- a/pkg/ebpf/c/common/buffer.h
+++ b/pkg/ebpf/c/common/buffer.h
@@ -3,6 +3,7 @@
 
 #include <vmlinux.h>
 
+#include <common/context.h>
 #include <common/hash.h>
 #include <common/network.h>
 
@@ -447,6 +448,10 @@ statfunc int events_perf_submit(program_data_t *p, u32 id, long ret)
 {
     p->event->context.eventid = id;
     p->event->context.retval = ret;
+
+    init_task_context(&p->event->context.task, p->event->task, p->config->options);
+    // keep task_info updated
+    bpf_probe_read_kernel(&p->task_info->context, sizeof(task_context_t), &p->event->context.task);
 
     // Get Stack trace
     if (p->config->options & OPT_CAPTURE_STACK_TRACES) {
--- a/pkg/ebpf/c/common/context.h
+++ b/pkg/ebpf/c/common/context.h
@@ -9,7 +9,7 @@
 
 // PROTOTYPES
 
-statfunc int init_context(void *, event_context_t *, struct task_struct *, u32);
+statfunc int init_task_context(task_context_t *, struct task_struct *, u32);
 statfunc void init_proc_info_scratch(u32, scratch_t *);
 statfunc proc_info_t *init_proc_info(u32, u32);
 statfunc void init_task_info_scratch(u32, scratch_t *);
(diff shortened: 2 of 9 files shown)


pkg/ebpf/c/tracee.bpf.c
```diff
@@ -349,7 +359,7 @@ SEC("raw_tracepoint/sys_execve")
 int syscall__execve(void *ctx)
 {
     program_data_t p = {};
-    if (!init_program_data(&p, ctx))
+    if (!init_tailcall_program_data(&p, ctx))
```
geyslan, line 362: Are those changes to tailcall fixes?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3012,16 +3019,17 @@ do_file_io_operation(struct pt_regs *ctx, u32 event_id, u32 tail_call_id, bool i
         // missed entry or not traced
         return 0;
     }
+    del_args(event_id);
```
geyslan, line 3022: Why the arg deletion should occur even if not traceable etc?

geyslan (commented):
That's great!

I've done the first pass and did some questions.

pkg/ebpf/c/common/buffer.h
```diff
@@ -448,6 +449,10 @@ statfunc int events_perf_submit(program_data_t *p, u32 id, long ret)
     p->event->context.eventid = id;
     p->event->context.retval = ret;
 
+    init_task_context(&p->event->context.task, p->event->task, p->config->options);
+    // keep task_info updated
+    bpf_probe_read(&p->task_info->context, sizeof(task_context_t), &p->event->context.task);
```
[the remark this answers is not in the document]
yanivagman: Your'e right, it's a good opportunity to make these changes in the code I touch

pkg/ebpf/c/common/context.h
```diff
@@ -218,27 +166,25 @@ statfunc int init_program_data(program_data_t *p, void *ctx)
...
+            case CONTAINER_STARTED:
+            case CONTAINER_EXISTED:
+                p->event->context.task.flags |= CONTAINER_STARTED_FLAG;
         }
     }
 
-    // update task_info with the new context
-    bpf_probe_read(&p->task_info->context, sizeof(task_context_t), &p->event->context.task);
+    // initialize matched_policies to all policies match
+    p->event->context.matched_policies = ~0ULL;
```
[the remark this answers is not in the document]
NDStrahilevitz: We've had issues before with these events in particular which is why this was placed. So they need to be tested specifically to make sure they didn't break.

pkg/ebpf/c/common/context.h
```diff
@@ -22,74 +22,53 @@ statfunc void reset_event_args(program_data_t *);
...
 
-    context->task.flags = 0;
+    tsk_ctx->flags = 0;
 
     if (is_compat(task))
-        context->task.flags |= IS_COMPAT_FLAG;
+        tsk_ctx->flags |= IS_COMPAT_FLAG;
 
     // Program name
-    __builtin_memset(context->task.comm, 0, sizeof(context->task.comm));
```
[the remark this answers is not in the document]
yanivagman: init_program_data() now does `__builtin_memset(&p->event->context.task, 0, sizeof(p->event->context.task));` so it covers all the fields of the task context. I'll also add this initialization as part of `init_task_info()` and then we should be ok with all the strings initalized here

pkg/ebpf/c/tracee.bpf.c
NDStrahilevitz: I assume because the args now live in saved_args. So no need to keep them around in the "inbetween" map.

pkg/ebpf/c/maps.h
```diff
@@ -49,7 +49,7 @@ typedef struct containers_map containers_map_t;
 // persist args between function entry and return
 struct args_map {
     __uint(type, BPF_MAP_TYPE_HASH);
-    __uint(max_entries, 1024);
+    __uint(max_entries, 10240);
```
NDStrahilevitz, line 52: What's the reason for increasing?

pkg/ebpf/c/common/context.h
```diff
@@ -22,74 +22,53 @@ statfunc void reset_event_args(program_data_t *);
...
-    if (unlikely(ret < 0)) {
-        tracee_log(ctx, BPF_LOG_LVL_ERROR, BPF_LOG_ID_GET_CURRENT_COMM, ret);
-        return -1;
-    }
+    bpf_get_current_comm(&tsk_ctx->comm, sizeof(tsk_ctx->comm));
 
     // UTS Name
     char *uts_name = get_task_uts_name(task);
-    if (uts_name) {
-        __builtin_memset(context->task.uts_name, 0, sizeof(context->task.uts_name));
```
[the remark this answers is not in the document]
yanivagman: ditto

pkg/ebpf/c/common/context.h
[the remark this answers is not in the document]
NDStrahilevitz: +1, I recall this as well.

pkg/ebpf/c/common/context.h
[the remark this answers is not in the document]
geyslan: Ok, but as all of them are calling init_program_data we already reset matched_policies with all bits on.

pkg/ebpf/c/tracee.bpf.c
yanivagman: using the tailcall version of the `init_program_data` keeps all the fields of `p->event` untouched, which also includes the `matched_policies` value that might have been set in the calling program

pkg/ebpf/c/common/buffer.h
```diff
@@ -448,6 +449,10 @@ statfunc int events_perf_submit(program_data_t *p, u32 id, long ret)
     p->event->context.eventid = id;
     p->event->context.retval = ret;
 
+    init_task_context(&p->event->context.task, p->event->task, p->config->options);
```
NDStrahilevitz, line 452: So this has no chance of failure now, which is because some instances of error handling were removed. What's the rationale behind the decision?

pkg/ebpf/c/common/context.h
```diff
@@ -21,75 +21,53 @@ statfunc void reset_event_args(program_data_t *);
...
     // Cgroup ID
     if (options & OPT_CGROUP_V1) {
-        context->task.cgroup_id = get_cgroup_v1_subsys0_id(task);
+        tsk_ctx->cgroup_id = get_cgroup_v1_subsys0_id(task);
     } else {
-        context->task.cgroup_id = bpf_get_current_cgroup_id();
+        tsk_ctx->cgroup_id = bpf_get_current_cgroup_id();
     }
 
-    // Context timestamp
```
NDStrahilevitz, line 84: All the four below will be enriched in userspace? Or is there another reason for removing them?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -77,6 +77,11 @@ int sys_enter_init(struct bpf_raw_tracepoint_args *ctx)
         task_info = init_task_info(tid, 0);
         if (unlikely(task_info == NULL))
             return 0;
+
+        int zero = 0;
+        config_entry_t *config = bpf_map_lookup_elem(&config_map, &zero);
+        if (config != NULL)
```
NDStrahilevitz, line 83: Maybe a `likely` around this conditional?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -210,6 +215,11 @@ int sys_exit_init(struct bpf_raw_tracepoint_args *ctx)
...
```
NDStrahilevitz, line 221: ditto

pkg/ebpf/c/tracee.bpf.c
NDStrahilevitz: Guess they are, as this program is a tailcall from the syscall enter/exit probe.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -5500,25 +5510,15 @@ int BPF_KPROBE(trace_sock_alloc_file)
     if (!is_socket_supported(sock))
         return 0;
 
-    // initialize program data
-
-    program_data_t p = {};
-    if (!init_program_data(&p, ctx))
```
NDStrahilevitz, line 5518: This is fine, because now filtering only will be done at the submit point of the packet right?
But then, we could potentially have much more packets being processed in the cgroup program side. Not sure if this will overall be better for performance, and it should be benchmarked in particular (I will run this branch on my cluster manually as well, but you should confirm with your oneliner too).
This may of course be redundant as we introduce @josedonizetti's refactor, but until then...

NDStrahilevitz (commented):
I will run this full branch on my cluster before giving a +1. Had some questions too.

Have to note on the filldir64 case in particular: I believe the case there isn't this optimization in particular. Note that we have a quick terminating conditional based on a kprobe context argument (the inode_number == 0 check). We could have just as well moved it to the head of the function before the init,trace,submit function trio, and had just the same effect.

pkg/ebpf/c/tracee.bpf.c
geyslan: Or `unlikely` and return if NULL.

pkg/ebpf/c/maps.h
yanivagman: Since I removed the should_trace() check in TRACE_ENT_FUNC macro, we may have to save more arguments between entry and return (theoretically, assuming filtering would have prevent that save) - so I allocated a bigger map size to mitigate this

pkg/ebpf/c/common/buffer.h
yanivagman: We had error handling in 2 places:
1. bpf_get_current_comm fails only when there is no task (https://elixir.bootlin.com/linux/latest/source/kernel/bpf/helpers.c#L261) - this will never happen or we would have failed earlier anyway
2 . reading uts name using bpf_probe_read_kernel_str shouldn't happen as well, and if it does we should continue anyway (no reason to drop the event in that case)

Considering the above + for simplicity + to save some instructions for all programs (since all call this function) - I removed this error handling

pkg/ebpf/c/common/context.h
yanivagman: exactly - they all moved to `init_program_data() ` but stack_id, which is set in the submit stage anyway

pkg/ebpf/c/tracee.bpf.c
yanivagman: This saves the initialization here since in any case we will do the initialization in `trace_ret_sock_alloc_file`.
The next few lines in this function just saves the arguments of the function, which has no effect on performance (tested)

geyslan:
It passed locally, I can't see why it failed when triggered by the action: https://github.com/aquasecurity/tracee/actions/runs/8360692120/job/22887161311.

Anyhoo, I'm gonna replay this test in Github.

geyslan (approved):
Great.

NDStrahilevitz (approved):
Performance test turned out good, let's merge.

PR #3926, opened 2024-03-21
Add security_path_notify test to PR workflow

oshaked1:

Add test for security_path_notify event to PR workflow.


oshaked1:
The security_path_notify function was introduced in kernel version 5.4. Currently there is no method to conditionally load probes based on kernel version, so the test fails on incompatible kernels.

geyslan:
https://github.com/aquasecurity/tracee/actions/runs/8373435136/job/22926492619?pr=3926#step:5:1468

```
INFO: SECURITY_PATH_NOTIFY: FAILED, stderr from tracee:
{"level":"warn","ts":1711017749.6577032,"msg":"Cancelling event and its dependencies because of missing probe","missing probe":103,"event":"security_path_notify","error":"probes.(*TraceProbe).attach: symbol not found: security_path_notify"}
```

@oshaked1 is this PR a demonstration of the lack of probes loading by kernel version? If so, please open an issue referring this, so we can close it for now.

@AlonZivony is this related to your work on dependencies?

AlonZivony:
> @AlonZivony is this related to your work on dependencies?

Yea, it is partly related.
We will need some mechanism to load only needed programs. It will use the dependencies mechanism that I created.
Afterwards we will use the fallback mechanism I want to merge using the dependencies mechanism to load other programs in the case of failure.

yanivagman (approved):
LGTM

Issue #3928, opened 2024-03-25
make check-pr is broken for shells but bash


geyslan:
## Description

The shell syntax might not be compliant with shells but bash.

```
if [[ "$$(git show --no-patch --format=%b $$commit)" ]]; then \
```

## Additional details 

Reported in https://github.com/aquasecurity/tracee/pull/3927

PR #3929, opened 2024-03-25
fix: make check-pr compliant with different shells

geyslan:
Close: #3928

ea5e6f445 **fix: make check-pr compliant with different shells**

```
It also now deals with trailing newlines in the commit bodies, printing
them following the same pattern.
```


yanivagman (approved):
LGTM

PR #3936, opened 2024-03-27
fix(ebpf): hidden_kernel_module fix max iterations

OriGlassman:

A change was depicted on kernel 6.5 where mod_tree contains duplicate modules and thus MAX_NUM_MODULES is not enough to iterate on the tree, which yielded a warning.
The fix is to use bpf_loop helper, which is available on kernels 5.17 and above, and do more iterations.

### 2. Explain how to test it

./dist/tracee -e=hidden_kernel_module

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -656,7 +656,8 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
 }
 
 // number of iterations - value that the verifier was seen to cope with - the higher, the better
-#define MAX_NUM_MODULES 450
+#define MAX_NUM_MODULES       450
+#define MAX_NUM_LOOPS_MODULES 100000
 
 enum
 {
@@ -763,7 +764,6 @@ statfunc int init_shown_modules()
         if (&pos->list == head) {
             return 0;
         }
-
         bpf_map_update_elem(&modules_map, &pos, &ker_mod, BPF_ANY);
     }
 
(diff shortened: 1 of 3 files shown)


pkg/ebpf/c/tracee.bpf.c
```diff
@@ -855,13 +855,69 @@ statfunc struct latch_tree_node *__lt_from_rb(struct rb_node *node, int idx)
...
+        } else {
+            curr = rb_nod.node;
+            struct latch_tree_node *ltn = __lt_from_rb(curr, loop_ctx->idx);
+            struct module *mod = BPF_CORE_READ(container_of(ltn, struct mod_tree_node, node), mod);
+
+            int ret = is_hidden((u64) mod);
+            if (ret == MOD_HIDDEN) {
+                lkm_seeker_send_to_userspace(mod, &loop_ctx->flags, loop_ctx->prog_data);
+            } else if (ret == HID_MOD_RACE_CONDITION) {
+                return 1;
```
AlonZivony, line 887: Maybe create an error for this case?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -855,13 +855,69 @@ statfunc struct latch_tree_node *__lt_from_rb(struct rb_node *node, int idx)
...
+    struct walk_mod_tree_ctx *loop_ctx = (struct walk_mod_tree_ctx *) ctx;
+    struct rb_node *curr = loop_ctx->curr;
+
+    if (curr != NULL) {
+            rb_node_t rb_nod = {.node = curr};
+            bpf_map_push_elem(&walk_mod_tree_queue, &rb_nod, BPF_EXIST);
+
+            loop_ctx->curr = BPF_CORE_READ(curr, rb_left); // Move left
+    } else {
+        rb_node_t rb_nod;
```
AlonZivony, line 875: rb_node?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -657,6 +657,7 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
 
 // number of iterations - value that the verifier was seen to cope with - the higher, the better
 #define MAX_NUM_MODULES 450
+#define MAX_NUM_LOOPS_MODULES 100000
```
AlonZivony, line 660: Its a bit much, no?
Why not 10k or even less?

AlonZivony (commented):
Looks good

pkg/ebpf/c/tracee.bpf.c
OriGlassman: It stops when finishes - so I want a value that if it happens, it'll certainly indicate a problem. 10k is also probably ok but doesn't make much different than 100k.

geyslan:
@OriGlassman is this still in the current milestone?

geyslan:
@OriGlassman I'm converting this to draft as it's been in conflict for a while. Feel free to return it again.

PR #3938, opened 2024-03-27
Policies pointer

geyslan:

62f60e83e **chore: go.mod (types) bump**
e08ca0d87 **perf: pass policies pointer in the event**

e08ca0d87 **perf: pass policies pointer in the event**

```
Instead of retrieving the snapshot of the policies at each stage of the
pipeline, do this only once at the beginning of the pipeline and pass
it embedded in the event.
```


types/trace/trace.go
```diff
@@ -8,47 +8,48 @@ import (
...
+	PIDNS                 int            `json:"pidNamespace"`
+	ProcessName           string         `json:"processName"`
+	Executable            File           `json:"executable"`
+	HostName              string         `json:"hostName"`
+	ContainerID           string         `json:"containerId"`
+	Container             Container      `json:"container,omitempty"`
+	Kubernetes            Kubernetes     `json:"kubernetes,omitempty"`
+	EventID               int            `json:"eventId,string"`
+	EventName             string         `json:"eventName"`
+	Policies              unsafe.Pointer `json:"-"`
```
geyslan, line 39: This commit is gonna be in other PR.

geyslan:
This optimization is still blocked since https://github.com/aquasecurity/tracee/pull/3841 only removed gob from tracee. tracee-rules still uses it.

pkg/ebpf/events_pipeline.go
```diff
@@ -256,6 +255,14 @@ func (t *Tracee) decodeEvents(ctx context.Context, sourceChan chan []byte) (<-ch
 			evt.ProcessEntityId = utils.HashTaskID(eCtx.HostPid, eCtx.LeaderStartTime)
 			evt.ParentEntityId = utils.HashTaskID(eCtx.HostPpid, eCtx.ParentStartTime)
 
+			policies, err := policy.Snapshots().Get(eCtx.PoliciesVersion)
+			if err != nil {
+				t.handleError(err)
+				t.eventsPool.Put(evt)
+				continue
+			}
+			evt.Policies = unsafe.Pointer(policies)
```
AlonZivony, line 264: Why are you using the `unsafe.Pointer` instead of just a pointer to the policies?

AlonZivony (commented):
I don't fully understand the pros of this change.
We are removing the information about the policies version from the event, making it hard for the user to understand it.
From what I do understand, this removes the need to search in the snapshots for the version of the policies each time queried, which is nice in idea because you no longer need to have an access to the policies snapshots to know about the policies relevant for the event. But won't any code that will need to go over the policies have a pointer to the policies to begin with? Can you give an example for where you need this direct access?
I am just a bit afraid of coupling the events with the policies.

pkg/ebpf/events_pipeline.go
geyslan: `types` would require importing the policy package to know about `*Policies`, which isn't the intention.

geyslan:
> But won't any code that will need to go over the policies have a pointer to the policies to begin with?

Only through the Snapshots Get*() since we'll be working based on a specific version.

> Can you give an example for where you need this direct access?

The idea is that if you get a event from the pipeline, anywhere in the codebase, you can get the version information directly from the Policies pointer. I think that keeping the version field available can induce the user to look for the Policies pointer in the Snapshots again.

geyslan:
> From what I do understand, this removes the need to search in the snapshots for the version of the policies each time queried,

Yes. In the future, we will have pruning working in Snapshots. It may happen that in a sequential pipeline stage this version of the policy is already pruned (timed out) and inaccessible via Snapshots, therefore keeping the pointer from the beginning will give us a greater guarantee of output from that event.

pkg/ebpf/events_pipeline.go
AlonZivony: But it is what you sort of do, no? Making the event to depend on the policies implementation.

geyslan:
@AlonZivony https://github.com/aquasecurity/tracee/pull/3940#issuecomment-2025371989

I'll be changing this, changing types won't be necessary anymore.

geyslan:
I was enveloping trace.Event type through the pipeline...

image


but reached a point (sig engine) that would require to change types again... as it's not the ideal currently, I'm postponing it.

PR #3939, opened 2024-03-27
chore!: remove gob support from tracee-rules

geyslan:

ec644981c **chore!: remove gob support from tracee-rules**

```
This finishes the removal of gob support after #3841.
```


yanivagman (approved):
LGTM

PR #3940, opened 2024-03-28
perf(types)!: use policies pointer

geyslan:

28adc668f **perf(types)!: use policies pointer**

```
Instead of holding the version of the policies, hold a pointer to the
policies object.
```


geyslan:
I synced with Yaniv and we decided on another strategy.

Issue #3950, opened 2024-04-03
Wrong fd in security socket connect


yanivagman:
## Description

Running Tracee with the `security_socket_connect` event shows a bad fd.
E.g.:
`12:34:13:735877  0      pool             31995   31999   0                security_socket_connect   sockfd: 1915682817, type: SOCK_STREAM, remote_addr: map[sa_family:AF_UNIX sun_path:/run/dbus/system_bus_socket]`

## Output of `tracee version`:

```
v0.20.0
```

## Output of `uname -a`:

```
(paste your output here)
```

PR #3951, opened 2024-04-03
fix: security_socket_connect wrong fd

yanivagman:

The security_scoket_connect event sent the wrong fd number. This was due to using a wrong way to check if the binary is a 32bit binary.

Fix: #3950

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -2612,9 +2612,8 @@ int BPF_KPROBE(trace_security_socket_connect)
     void *args_buf = &p.event->args_buf;
     void *to = (void *) &sys->args.args[0];
 
-#if defined(bpf_target_x86) // only i386 binaries uses socketcall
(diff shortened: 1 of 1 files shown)


geyslan:
Perhaps there's more places to fix, like `case SYSCALL_SOCKETCALL:`: https://github.com/aquasecurity/tracee/commit/5040b00bdc8b2bddc102a13b35f394cd42acfb50

yanivagman:
> Perhaps there's more places to fix, like `case SYSCALL_SOCKETCALL:`: [5040b00](https://github.com/aquasecurity/tracee/commit/5040b00bdc8b2bddc102a13b35f394cd42acfb50)

In all these other places there's the `case SYSCALL_SOCKETCALL` which should only be called from i386 binaries, so I think it is no problem there (there might be a different problem there though, since I see that it uses args[1] as the file descriptor, while this is actually a void pointer, but this is not related to this fix)

geyslan (approved):
LGTM

PR #3953, opened 2024-04-03
Add `suspicious_syscall_source` event

### LINKED ISSUE #4002: Add `suspicious_syscall_source` event
Add a new event, check_syscall_source, which reports invocations of syscalls from unusual code locations (stack, heap and anonymous VMAs).

oshaked1:

Add a new event, `suspicious_syscall_source`, which reports invocations of syscalls from unusual code locations (stack, heap and anonymous VMAs).
It is implemented using generic syscall kprobes, which trigger the analysis only for syscalls that were selected using event parameters, e.g. `tracee -e suspicious_syscall_source.args.syscall=open,openat`.
Closes #4002 

### 2. Explain how to test it

The easiest way to test the event is using the included tester program, which executes a small shellcode that invokes the `exit` syscall from the stack/heap/anonymous VMA according to the command line argument. Note that it must be compiled with `-z execstack` to be able to run the shellcode from the stack.

Another way of testing the event (for arbitrary syscalls) is by packing a program:

```bash
# compile statically, so libc syscall wrappers are not loaded to separate (legitimate) code areas
gcc -o my_program -static my_program.c

# pack the program
upx -o my_program_upx my_program

# run it
./my_program_upx
```


oshaked1:
The ebpf code for this event contains a call to the `bpf_find_vma` helper if it exists (an alternative logic is implemented for older kernels where the helper doesn't exist).

Because this helper uses a callback function or in verifier terms a "bpf-to-bpf call", this program cannot be combined with tail calls (which it uses) on some older kernels, even though the helper call would not be called on those kernels.

As a result, this event must be split into 2 separate programs that will be loaded conditionally based on kernel version. Currently no such mechanism exists, but is in the works. When such a mechanism will be implemented this event will be split into 2 programs so it can be merged.

geyslan:
@oshaked1 I'm converting this to draft since we're tackling the release and filtering in ready PRs, ok? Please be welcome to set status again the later.

pkg/ebpf/event_parameters.go
```diff
@@ -0,0 +1,97 @@
...
+// For example, an event can use one of its parameters to populate eBPF maps,
+// or perhaps attach eBPF programs according to the parameters.
+func (t *Tracee) handleEventParameters() error {
+	// Iterate through registerd event parameter handlers
+	for eventID, handler := range eventParameterHandlers {
+		// Make sure this event is selected
+		if _, err := t.eventsDependencies.GetEvent(eventID); err != nil {
+			continue
+		}
+		// Get the event parameters for all policies
```
yanivagman, line 30: Let's add a comment about using data filters to get event prameters. We should add a TODO here that event parameters should be extracted from a dedicated parameteres field in the future

pkg/ebpf/event_parameters.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/aquasecurity/tracee/pkg/ebpf/probes"
+	"github.com/aquasecurity/tracee/pkg/events"
+	"github.com/aquasecurity/tracee/pkg/filters"
+	"github.com/aquasecurity/tracee/pkg/logger"
+)
+
+type eventParameterHandler func(t *Tracee, eventParams []map[string]filters.Filter[*filters.StringFilter]) error
+
+var eventParameterHandlers = map[events.ID]eventParameterHandler{
+	events.SuspiciousSyscallSource: attachSuspiciousSyscallSourceProbes,
```
yanivagman, line 16: Since you created event parameters, we can move print_mem_dump in here as well (I'm not sure if there are other events that abuse data filters as parameters)

pkg/ebpf/tracee.go
```diff
@@ -125,6 +125,8 @@ type Tracee struct {
 	// This does not mean they are required for tracee to function.
 	// TODO: remove this in favor of dependency manager nodes
 	requiredKsyms []string
+	// Probes created for suspicious_syscall_source event
+	suspiciousSyscallSourceProbes *probes.ProbeGroup
```
yanivagman, line 129: I don't think it is a good idea to add probe groups into the tracee struct.
It makes the struct coupled to specific events, and we want to decouple things, not the other way around.

yanivagman (changes requested):

pkg/ebpf/c/common/common.h
```diff
@@ -62,7 +62,7 @@ static __inline int has_prefix(char *prefix, char *str, int n)
     }
 
     // prefix is too long
-    return 0;
+    return 1;
```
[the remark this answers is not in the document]
oshaked1: I changed it to match the logic of the macro version of `has_prefix`, which returns 1 if the prefix is longer than n bytes.
If 0 is returned in such case, then comparing a string to a prefix with the same length (essentially using it as `strcmp`) would falsely return 0.

pkg/ebpf/c/common/memory.h
```diff
@@ -51,4 +57,123 @@ statfunc struct mount *real_mount(struct vfsmount *mnt)
...
+        if (!alerted_find_vma_unsupported) {
+            tracee_log(ctx, BPF_LOG_LVL_WARN, BPF_LOG_FIND_VMA_UNSUPPORTED, 0);
+            alerted_find_vma_unsupported = true;
+        }
+        return NULL;
+    }
+
+    // TODO: we don't support NOMMU systems yet (looking up VMAs on them requires walking the VMA
+    // linked list)
+    if (!get_kconfig(MMU)) {
```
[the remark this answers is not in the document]
oshaked1: Ok I will remove it

pkg/ebpf/c/common/common.h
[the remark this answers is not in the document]
yanivagman: Ok, so then we should fix the specific case that string has the same size as prefix.
Yet, we don't want to return true if prefix is longer than original string

pkg/ebpf/tracee.go
oshaked1: Where should it be added? The main probe group is also part of the tracee struct

pkg/ebpf/event_parameters.go
oshaked1: `print_mem_dump` uses event filters to populate the required ksymbols. `handleEventParameters` is called at the end of initialization, because it needs to happen after `initBPF`, which is called after the kernel symbols are initialized.

pkg/ebpf/c/common/common.h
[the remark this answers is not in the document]
oshaked1: I resolved this issue in #4394 - I will rebase onto it and use the new `strncmp` when it's merged

pkg/ebpf/tracee.go
oshaked1: So where should I put it? I also plan to add additional "syscall checker" events that will all utilize the same probe group which will make it less not specific to a single event but for a "class" of events

oshaked1: @yanivagman I added a map to the tracee struct that holds extra probe groups (indexed by a given name) that can be added dynamically.

go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/IBM/fluent-forward-go v0.2.2
 	github.com/Masterminds/sprig/v3 v3.2.3
 	github.com/aquasecurity/libbpfgo v0.7.0-libbpf-1.4.0.20240729111821-61d531acf4ca
+	github.com/aquasecurity/libbpfgo/helpers v0.4.5
```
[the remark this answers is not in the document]
oshaked1: Removed it, it was left after code I deleted

yanivagman (approved):
LGTM

PR #3954, opened 2024-04-03
Add `ArgVal` signature helper

oshaked1:

Add `ArgVal` signature helper to extract a generic type from event arguments.

This function is simply copied from `pkg/events/parse/params.go` so it is now available for signatures.

### 3. Other comments

This PR is needed for https://github.com/aquasecurity/tracee/pull/3953

### DIFF
--- a/signatures/helpers/arguments_helpers.go
+++ b/signatures/helpers/arguments_helpers.go
@@ -7,6 +7,25 @@ import (
 	"github.com/aquasecurity/tracee/types/trace"
 )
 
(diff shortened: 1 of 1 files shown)


geyslan:
What about a pkg to keep the same logic and be imported by both instead of duplicating it? Perhaps [utils](https://github.com/aquasecurity/tracee/tree/main/pkg/utils)?

yanivagman (approved):
LGTM

PR #3960, opened 2024-04-04
fix: parse arguments only if chosen by flag

AlonZivony:

Fix an the issue that cause the parse arguments to always run. It was probably caused to support signatures using parsed arguments. From now on it should be chosen directly in these cases.

### DIFF
--- a/pkg/cmd/cobra/cobra.go
+++ b/pkg/cmd/cobra/cobra.go
@@ -320,9 +320,6 @@ func GetTraceeRunner(c *cobra.Command, version string) (cmd.Runner, error) {
 	runner.Printer = p
 	runner.InstallPath = traceeInstallPath
 
-	// parse arguments must be enabled if the rule engine is part of the pipeline
-	runner.TraceeConfig.Output.ParseArguments = true
-
 	runner.TraceeConfig.EngineConfig = engine.Config{
 		Enabled:          true,
 		SigNameToEventID: sigNameToEventId,


pkg/cmd/cobra/cobra.go
```diff
@@ -320,9 +320,6 @@ func GetTraceeRunner(c *cobra.Command, version string) (cmd.Runner, error) {
 	runner.Printer = p
 	runner.InstallPath = traceeInstallPath
 
-	// parse arguments must be enabled if the rule engine is part of the pipeline
-	runner.TraceeConfig.Output.ParseArguments = true
```
geyslan, line 324: It seems that kernel tests require it on: https://github.com/aquasecurity/tracee/actions/runs/8555281070/job/23442355573?pr=3960#step:7:74

NDStrahilevitz:
Pretty sure this is by design until we convert all our signatures to non parsed args.

NDStrahilevitz:
Actually, even if we convert all our signatures, we would introduce another problem, that the instant some opts to use the flag they would break the (now converted) signatures. So basically it would reintroduce the test failures from the opposite end.
BTW, marking that this resolves the last checkpoint from #2177. What we're missing is a solution for #3008.

AlonZivony:
> Pretty sure this is by design until we convert all our signatures to non parsed args. I should have a branch with a commit doing that lying somewhere on my machine if we opt for that.

The more reasonable option is to require the user to use the flag if he wants to use the signatures.
Currently the flag doesn't do anything as it is overridden when the signatures engine is used, which is always.

The main problem right now with this is the fact that Tracee is made to use the unparsed arguments.
As implementing the full analyze mode requires streaming unparsed events into Tracee we can either not parse them from the start or unparse them in the analyze mode. The later option is the better one probably, especially if we can assuming from the stream whether the arguments are parsed or not. However, the analyze mode introduce too much code as is and is hard to merge. For now it makes more sense that it will only support unparsed events, but it requires that the `parse-arguments` option will enable to opt-out the parsing, which is currently not an option.

In the end, it makes more sense to give the user the option to choose whether he wants parsed or unparsed arguments.

geyslan:
@rscampos I don't know how much you have tackled the analyze mode until now, but this is linked to it.

yanivagman:
Already done in https://github.com/aquasecurity/tracee/pull/4331

PR #3962, opened 2024-04-04
Run x86_64 & aarch64 builds in parallel

hangrymuppet:
I don't think the aarch64 build depends on the x86_64 build after a cursory look. This should reduce build time by over and hour and let me have new builds at 11PM instead of 1AM in my timezone 😁

### DIFF
--- a/.github/workflows/release-snapshot.yaml
+++ b/.github/workflows/release-snapshot.yaml
@@ -57,7 +57,6 @@ jobs:
       #   shell: bash
   release-snapshot-aarch64:
     name: Release Snapshot (aarch64)
-    needs: [ release-snapshot-x86_64 ]
     runs-on: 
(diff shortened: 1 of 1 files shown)


geyslan:
I really don't know why this requirement was firstly inserted. Perhaps due to limitation of parallel processing since the entire procedure takes a lot of resources. @yanivagman any clue?

yanivagman:
> I really don't know why this requirement was firstly inserted. Perhaps due to limitation of parallel processing since the entire procedure takes a lot of resources. @yanivagman any clue?

Not sure why as well. We can probably remove this requirement

geyslan:
> Not sure why as well. We can probably remove this requirement

Ok, I'm going to merge it and see what happens.

geyslan:
@hangrymuppet tks again.

geyslan:
It takes about 45min less. Great.

image

PR #3963, opened 2024-04-05
Add Iterator generic interface, debut it in Policies

geyslan:
5dc0a1bbc **chore: add CreateAllIterator**
688c30403 **chore: add CreateUserlandIterator**
7de062cbd **chore: introduce PoliciesIterator**
faef01c64 **chore: introduce Iterator generic interface**

5dc0a1bbc **chore: add CreateAllIterator**

```
Use CreateAllIterator instead of directly accessing the policies map,
which is a private field.

This also removes a stale Policies validation from the Config struct
method.

Signed-off-by: Geyslan Gregório <geyslan@gmail.com>
```

688c30403 **chore: add CreateUserlandIterator**

```
Use CreateUserlandIterator instead of directly accessing userland
policies slice, which is a private field.

Signed-off-by: Geyslan Gregório <geyslan@gmail.com>
```

faef01c64 **chore: introduce Iterator generic interface**

```
The Iterator pattern can be used to traverse a collection of elements
without exposing its underlying representation.

Signed-off-by: Geyslan Gregório <geyslan@gmail.com>
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -8,7 +8,6 @@ import (
 	"github.com/aquasecurity/tracee/pkg/containers/runtime"
 	"github.com/aquasecurity/tracee/pkg/dnscache"
 	"github.com/aquasecurity/tracee/pkg/errfmt"
-	"github.com/aquasecurity/tracee/pkg/events"
 	"github.com/aquasecurity/tracee/pkg/events/queue"
 	"github.com/aquasecurity/tracee/pkg/policy"
 	"github.com/aquasecurity/tracee/pkg/proctree"
@@ -39,22 +38,6 @@ type Config struct {
 
 // Validate does static validation of the configuration
 func (c Config) Validate() error {
-	// Policies
-	for _, p := range c.Policies.Map() {
-		if p == nil {
-			return errfmt.Errorf("policy is nil")
-		}
-		if p.EventsToTrace == nil {
-			return errfmt.Errorf("policy [%d] has no events to trace", p.ID)
-		}
-
-		for e := range p.EventsToTrace {
-			if !events.Core.IsDefined(e) {
-				return errfmt.Errorf("invalid event [%d] to trace in policy [%d]", e, p.ID)
-			}
-		}
-	}
-
 	// Buffer sizes
 	if (c.PerfBufferSize & (c.PerfBufferSize - 1)) != 0 {
 		return errfmt.Errorf("invalid perf buffer size - must be a power of 2")
--- a/pkg/events/derive/symbols_collision.go
+++ b/pkg/events/derive/symbols_collision.go
@@ -29,8 +29,9 @@ func SymbolsCollision(soLoader sharedobjs.DynamicSymbolsLoader, policies *policy
 	symbolsCollisionFilters := map[string]filters.Filter{}
 
 	// pick white and black lists from the filters (TODO: change this)
-	for _, policies := range policies.Map() {
-		f := policies.ArgFilter.GetEventFilters(events.SymbolsCollision)
+	for it := policies.CreateAllIterator(); it.HasNext(); {
+		p := it.Next()
+		f := p.ArgFilter.GetEventFilters(events.SymbolsCollision)
 		maps.Copy(symbolsCollisionFilters, f)
 	}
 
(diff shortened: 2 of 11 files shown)


geyslan:
@AlonZivony as you suggested I bring in the Iterator pattern.

pkg/config/config.go
```diff
@@ -39,22 +38,6 @@ type Config struct {
 
 // Validate does static validation of the configuration
 func (c Config) Validate() error {
-	// Policies
-	for _, p := range c.Policies.Map() {
```
geyslan, line 43: Stale validation.

pkg/events/derive/symbols_collision.go
```diff
@@ -29,8 +29,10 @@ func SymbolsCollision(soLoader sharedobjs.DynamicSymbolsLoader, policies *policy
 	symbolsCollisionFilters := map[string]filters.Filter{}
 
 	// pick white and black lists from the filters (TODO: change this)
-	for _, policies := range policies.Map() {
-		f := policies.ArgFilter.GetEventFilters(events.SymbolsCollision)
+	it := policies.CreateAllIterator()
+	for it.HasNext() {
```
AlonZivony, line 33: Suggestion for a different and maybe more readable way to write it (it is more C like, so I am not sure if it is really better):
```Go
for it := policies.CreateAllIterator; it.HasNext(); p := it.GetNext() {
	f := p.ArgFilter.GetEventFilters(events.SymbolsCollision)
	maps.Copy(symbolsCollisionFilters, f)
}
```
I think it has the pro of not exposing the `it` variable after the loop, as it is only created for iterating over it.

pkg/policy/policies_iterator.go
```diff
@@ -0,0 +1,54 @@
...
+	index    int
+}
+
+// HasNext returns true if there are more policies to iterate.
+func (i *PoliciesIterator) HasNext() bool {
+	return i.index < len(i.policies)
+}
+
+// GetNext returns the next policy in the iteration.
+func (i *PoliciesIterator) GetNext() *Policy {
```
geyslan, line 19: Returning a pointer is not the ideal result, but keep in mind that Policy will become a read-only interface in the future, so this will be resolved.

pkg/utils/utils.go
```diff
@@ -13,6 +13,15 @@ type Cloner interface {
 	Clone() Cloner
 }
 
+// Iterator is a generic interface for iterators.
+type Iterator[T any] interface {
+	// HasNext returns true if there are more elements to iterate.
+	HasNext() bool
+
+	// GetNext returns the next element in the iteration.
+	GetNext() T
```
NDStrahilevitz, line 22:
```suggestion
	Next() T
```

pkg/policy/policies_iterator.go
```diff
@@ -0,0 +1,54 @@
...
+	if i.HasNext() {
```
NDStrahilevitz, line 20: stylistic nit to revert the conditional and short return nil here.

pkg/policy/policies_iterator.go
```diff
@@ -0,0 +1,54 @@
...
+func (ps *Policies) CreateUserlandIterator() utils.Iterator[*Policy] {
+	return &PoliciesIterator{
+		policies: ps.userlandPolicies,
+		index:    0,
+	}
+}
+
+// CreateAllIterator returns a new iterator for all policies.
+//
+// This iterator should not be used in hot paths since to avoid keeping
```
NDStrahilevitz, line 41: WDYT?
```suggestion
// WARNING: This iterator should not be used in hot paths. In order to avoid keeping
// the list of policies in memory, create the iterator on the fly from its
// respective map instead.
```

NDStrahilevitz: BTW, how bad is the memory footprint that we shouldn't keep the list in memory?

NDStrahilevitz (approved):
LGTM

pkg/policy/policies_iterator.go
geyslan: 256 * policy = up to 16MB... for each policies version in memory.

If better performance is required for that use, in the future we can keep that list static.

geyslan: In hindsight, I think I'll make it static right now.

pkg/policy/policies_iterator.go
geyslan: Good point. Sticking to the happy path whenever possible.

PR #3965, opened 2024-04-08
feat(events): introduce events fallbacks

AlonZivony:

Introduce the concept of events failure because of dependencies.
Add events the ability to declare a fallback dependencies, to be used upon failure.
This allows events to implement not necessarily supported implementation,
for new kernels for example, with other implementation to cover unsupported scenarios.


NDStrahilevitz:
IMO, if we're going by an approach where we can replace an entire set of dependency with another, then instead of having a recursive definition, each event should just have an array of definition so the map becomes `map[events.ID][]EventDefinition`. I think it would be slightly more readable, WDYT?

AlonZivony:
This PR is missing e2e tests.
We can have them by changing the current dependencies e2e tests, adding an event with failed dependencies and good fallback.
I fear I won't be able to do it now :(.

yanivagman:
This PR was replaced by https://github.com/aquasecurity/tracee/pull/4866 and another followup PR that will be created

PR #3966, opened 2024-04-08
chore: make Cloner generic

geyslan:

64ac6e62 **chore: make Cloner generic**

```
This simplifies the Cloner interface usage since it is not necessary to
assert the Clone return to the specific type.

This also includes compile-time checks to ensure that the Clone method
is implemented for all types that should be cloned.

Signed-off-by: Geyslan Gregório <geyslan@gmail.com>
```

### DIFF
--- a/pkg/filters/args.go
+++ b/pkg/filters/args.go
@@ -12,20 +12,23 @@ import (
 )
 
 type ArgFilter struct {
-	filters map[events.ID]map[string]Filter
+	filters map[events.ID]map[string]Filter[*StringFilter]
 	enabled bool
 }
 
+// Compile-time check to ensure that ArgFilter implements the Cloner interface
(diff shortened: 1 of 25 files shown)


pkg/filters/args.go
```diff
@@ -12,20 +12,22 @@ import (
 )
 
 type ArgFilter struct {
-	filters map[events.ID]map[string]Filter
+	filters map[events.ID]map[string]Filter[*StringFilter]
 	enabled bool
 }
 
+var _ utils.Cloner[*ArgFilter] = &ArgFilter{}
```
NDStrahilevitz, line 19: This is for type assertion right? Maybe add a comment so it won't be accidentally removed.

NDStrahilevitz (approved):

PR #3967, opened 2024-04-09
feat(dependencies): further adopt dependencies API to ksymbols and probes

### LINKED ISSUE #3933: Cancelled events probes are not cleaned properly
## Description

If an event fails during attachment stage, it will be cancelled by removing it from the `eventsState` map:
```Go
// cancelEventFromEventState cancels an event and all its dependencies from the eventsState map.
func (t *Tracee) cancelEventFromEventState(evtID events.ID) {
	delete(t.eventsState, evtID)
	evtDef := events.Core.GetDefinitionByID(evtID)
	for _, evtDeps := range evtDef.GetDependencies().GetIDs() {
		t.cancelEventFromEventState(evtDeps)
	}
}
```

However, this is hardly enough.
If the event has other probes that were attached already, there is no effort to detach them (if no other event depends on these probes).
Moreover, the policies maps are not affected from this cancellation, resulting that the `should_trace` function for the cancelled event in the eBPF code will still return true. This means that we will still get the event from the eBPF code. This is a huge waste of resources, and might also result bugs if the event's other probes should clean or do some other logic with the failed probe.

```
(paste your output here)
```

## Output of `uname -a`:

```
(paste your output here)
```

AlonZivony:

This PR expands and uses the dependencies API introduced in #3931 for other events dependencies handling.
It adds the probes dependencies to the dependencies manager, and use it to manage probes cancelling.
This PR also add the dependencies watchers for adding and removing the probes, and use the dependencies to cancel events with missing ksymbols.

Fix #3933

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -6798,3 +6798,50 @@ int sched_process_exit_signal(struct bpf_raw_tracepoint_args *ctx)
 }
 
 // END OF Control Plane Programs
+
+// Tests
+
+SEC("kprobe/empty_kprobe")
+int BPF_KPROBE(empty_kprobe)
+{
+    return 0;
+}
+
+SEC("raw_tracepoint/exec_test")
+int tracepoint__exec_test(struct bpf_raw_tracepoint_args *ctx)
+{
+    // Check if test file was executed
+    struct linux_binprm *bprm = (struct linux_binprm *) ctx->args[2];
+    if (bprm == NULL)
+        return -1;
+    struct file *file = get_file_ptr_from_bprm(bprm);
+    void *file_path = get_path_str(__builtin_preserve_access_index(&file->f_path));
+    if (file_path == NULL || !has_prefix("/tmp/test", file_path, 9))
+        return 0;
+
+    // Submit all test events
+    int ret = 0;
+    program_data_t p = {};
+    if (!init_program_data(&p, ctx, NO_EVENT_SUBMIT))
+        return 0;
+
+    if (!evaluate_scope_filters(&p))
+        return 0;
+
+    if (!reset_event(p.event, EXEC_TEST))
+        return 0;
+    if (evaluate_scope_filters(&p))
+        ret |= events_perf_submit(&p, 0);
+
+    if (!reset_event(p.event, TEST_MISSING_KSYMBOLS))
+        return 0;
+    if (evaluate_scope_filters(&p))
+        ret |= events_perf_submit(&p, 0);
+
+    if (!reset_event(p.event, TEST_FAILED_ATTACH))
+        return 0;
+    if (evaluate_scope_filters(&p))
+        ret |= events_perf_submit(&p, 0);
+
+    return 0;
+}
--- a/pkg/ebpf/tracee.go
+++ b/pkg/ebpf/tracee.go
@@ -1004,50 +1004,83 @@ func (t *Tracee) initKsymTableRequiredSyms() error {
 	return nil
 }
 
-// getUnavKsymsPerEvtID returns event IDs and symbols that are unavailable to them.
-func (t *Tracee) getUnavKsymsPerEvtID() map[events.ID][]string {
-	unavSymsPerEvtID := map[events.ID][]string{}
+// getUnavailbaleKsymbols return all kernel symbols missing from given symbols
+func getUnavailbaleKsymbols(ksymbols []events.KSymbol, kernelSymbols *environment.KernelSymbolTable) []events.KSymbol {
+	var unavailableSymbols []events.KSymbol
 
-	evtDefSymDeps := func(id events.ID) []events.KSymbol {
-		depsNode, _ := t.eventsDependencies.GetEvent(id)
-		deps := depsNode.GetDependencies()
-		return deps.GetKSymbols()
-	}
-
-	for evtID := range t.eventsState {
-		for _, symDep := range evtDefSymDeps(evtID) {
-			sym, err := t.kernelSymbols.GetSymbolByName(symDep.GetSymbolName())
-			symName := symDep.GetSymbolName()
-			if err != nil {
-				// If the symbol is not found, it means it's unavailable.
-				unavSymsPerEvtID[evtID] = append(unavSymsPerEvtID[evtID], symName)
-				continue
-			}
-			for _, s := range sym {
-				if s.Address == 0 {
-					// Same if the symbol is found but its address is 0.
-					unavSymsPerEvtID[evtID] = append(unavSymsPerEvtID[evtID], symName)
-				}
+	for _, ksymbol := range ksymbols {
+		sym, err := kernelSymbols.GetSymbolByName(ksymbol.GetSymbolName())
+		if err != nil {
+			// If the symbol is not found, it means it's unavailable.
+			unavailableSymbols = append(unavailableSymbols, ksymbol)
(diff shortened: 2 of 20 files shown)


AlonZivony:
This PR was extracted from #3965 and will serve it for its logic

AlonZivony:
This however won't solve the issue that we need to update the `events_map` in the eBPF so events removed won't be there anymore. @geyslan is this something you can do? I am a bit lost with how to do it with the new policies versioning mechanism.

geyslan:
> This however won't solve the issue that we need to update the `events_map` in the eBPF so events removed won't be there anymore. @geyslan is this something you can do? I am a bit lost with how to do it with the new policies versioning mechanism.

For sure, could you put TODO comments?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -6762,3 +6762,52 @@ int sched_process_exit_signal(struct bpf_raw_tracepoint_args *ctx)
...
+    if (evaluate_scope_filters(&p))
+        ret |= events_perf_submit(&p, 0);
+
+    if (ret == 0)
+        // This is not a guarantee that the event will be submitted once, but it is good enough for
+        // tests as the purpose is to not create too much of a load.
+        did_submit = true;
+
+    return 0;
+}
```
NDStrahilevitz, line 6813: nit: missing new line

tests/integration/dependencies_test.go
```diff
@@ -0,0 +1,264 @@
...
+			},
+		)
+	}
+	restoreLogger()
+}
+
+// testLogs searches for given logs and test when input channel closes if all
+// logs were received.
+// It also returns a channel with the result of the test - whether all logs were found.
+func testLogs(
```
NDStrahilevitz, line 129: testing logs is a rather flaky kind of test. Are you sure we need it? If so, lets make the logging based on constants, so the test won't break easily if we change the logs for whatever reason.

Also, since we already added the testutils logger, and _this_ is the way to test it, I think the function should be placed alongside. That way if we do want to keep testing logs, we can reuse this technique.

tests/integration/dependencies_test.go
```diff
@@ -0,0 +1,264 @@
...
+		close(outChan)
+	}()
+	return outChan
+}
+
+// testEvents searches for the given events in the events channel, and when the
+// run is completed test whether the given events have been found.
+// It gives the option for searching for event that should and shouldn't be
+// found to enable precise tests.
+func testEvents(
```
NDStrahilevitz, line 178: I believe @geyslan made similar functions in his filter tests. Impossible to reuse them?

AlonZivony: I will look into it

pkg/ebpf/tracee.go
```diff
@@ -862,50 +862,68 @@ func (t *Tracee) newConfig(cfg *policy.PoliciesConfig, version uint16) *Config {
 	}
 }
 
-// getUnavKsymsPerEvtID returns event IDs and symbols that are unavailable to them.
-func (t *Tracee) getUnavKsymsPerEvtID() map[events.ID][]string {
-	unavSymsPerEvtID := map[events.ID][]string{}
+// getUnavailbaleKsymbols return all kernel symbols missing from given symbols
+func getUnavailbaleKsymbols(ksymbols []events.KSymbol, kernelSymbols *helpers.KernelSymbolTable) []events.KSymbol {
```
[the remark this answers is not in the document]
AlonZivony: After some more thoughts about the subject, I believe that this is unnecessary.
To reliably cache I will need to refresh the cache upon refresh to the kernel symbols object, which will require either coupling or larger scale change to allow it without coupling.
We have maybe 5 symbols dependencies in the code today, so I believe there will be no real impact after implementing this so it is probably not worth the effort.

tests/integration/event_filters_test.go
```diff
@@ -2189,14 +2158,44 @@ func ExpectAtLeastOneForEach(t *testing.T, cmdEvents []cmdEvents, actual *eventB
...
-
 				// if we got here, it means we found a match and can stop searching
-				found = true
-				break
+				return true, nil
+			}
+			return false, nil
+		}
+
+		// second stage: validate events
```
AlonZivony, line 2167: I have added support for unexpected events only in this function (`ExpectAtLeastOneForEach`) as I only used this one.
If its seems good to you @NDStrahilevitz I can add it to all other functions as well.
I admit that I think that the current API created by @geyslan is a bit too complicated - it has too many functions so the differences between them is hard to understand. Anyways this PR is not about the testing infrastructure, so I only open it here for discussion.

AlonZivony:
After a conversation with @yanivagman we arrived to the following problems that should be addressed with the watchers mechanism:
1. Add watchers order should be from dependencies to dependants, and removed watchers as well. The dependants cannot work without the dependencies, so there is no logic in having them without their dependencies.
2. All watchers should be called after all the tree management operations are finished.
3. As some watchers might call `RemoveEvent`, we should either call watchers by order of invocation, or cancel irrelevant watchers (for example, cancel add watchers on event that its remove was called already).

yanivagman (commented):
Nice work @AlonZivony 
The automatic handling of event dependencies will make it easier to implement dynamic policy modification in the future. @geyslan FYI

pkg/events/dependencies/manager.go
```diff
@@ -41,6 +44,15 @@ func (m *Manager) GetEvent(id events.ID) (*EventNode, bool) {
 	return node, true
 }
 
+// GetHandle returns the given handle node managed by the Manager
+func (m *Manager) GetHandle(handle probes.Handle) (*HandleNode, bool) {
```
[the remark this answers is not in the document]
AlonZivony: Its telling if the value was found.
It could be removed as we return a pointer, but its matching the `GetEvent` function.

pkg/events/dependencies/manager.go
```diff
@@ -185,3 +211,23 @@ func (m *Manager) removeDependants(node *EventNode) {
...
+	if !ok {
+		handleNode = NewHandleNode(handle, []events.ID{dependant})
+		m.handles[handle] = handleNode
+	} else {
+		handleNode.addDependant(dependant)
+	}
+}
+
+// removeNode removes the node from the tree.
+func (m *Manager) removeHandle(handle *HandleNode) bool {
```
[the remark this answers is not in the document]
AlonZivony: It represents if the remove was successful in the sense of if the handle existed in the manager.
I can replace it with an error, and create an error for "value not found". But as currently the remove cannot  fail, the `bool` value seems cleaner to me.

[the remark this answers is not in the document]
yanivagman: I think using `error` for non existing value is a a common practice

tests/integration/event_filters_test.go
geyslan: > I admit that I think that the current API created by @geyslan is a bit too complicated

I agree with you @AlonZivony. Integration API must be rethought.

pkg/events/dependencies/manager.go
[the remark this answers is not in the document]
geyslan: > I can replace it with an error, and create an error for "value not found".

So the check against it should be `if errors.Is(err, dependencies.ValueNotFoundError) {`. It's a good practice if we'll have more error types, if not... the check against the bool is cleaner.

yanivagman (approved):
Mostly nit comments,
great work!

PR #3971, opened 2024-04-10
Fix Policies Cloning

geyslan:

6185bdcda **chore: improve Clone tests by using google/go-cmp**
227367504 **chore(filters): add ValueHandler function type**
02e4a1238 **chore(policy): add inner methods for iteration**
392535aea **fix(policy): iterate array to preserve order**
239087aca **fix(policy): Policies Clone method**

227367504 **chore(filters): add ValueHandler function type**

```
It is a function type that can be used to handle the value of a filter
when it is parsed.

This also removes a dead code.
```

02e4a1238 **chore(policy): add inner methods for iteration**

```
Depending o the use case, the iteration of the policies can be done
by allFromArray() or allFromMap().
```

392535aea **fix(policy): iterate array to preserve order**

```
To fill userlandPolicies, we need to iterate over the array of policies
to preserve the order, since the map iteration order is not guaranteed.
```

239087aca **fix(policy): Policies Clone method**

```
Use Set instead of Add to preserve the order of the policies.

This also removes the unnecessary call to compute() method as it is
already called by the Set inners.
```

### DIFF
--- a/pkg/policy/policies.go
+++ b/pkg/policy/policies.go
@@ -128,7 +128,11 @@ func (ps *Policies) compute() {
 
 	userlandList := []*Policy{}
 	ps.filterableInUserland = 0
-	for _, p := range ps.policiesMapByID {
+	for _, p := range ps.policiesArray {
+		if p == nil {
+			continue
+		}
+
 		if p.ArgFilter.Enabled() ||
 			p.RetFilter.Enabled() ||
 			p.ContextFilter.Enabled() ||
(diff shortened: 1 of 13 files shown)


pkg/policy/policies.go
```diff
@@ -296,14 +296,12 @@ func (ps *Policies) Clone() utils.Cloner {
 		if p == nil {
 			continue
 		}
-		if err := nPols.Add(p.Clone().(*Policy)); err != nil {
+		if err := nPols.Set(p.Clone().(*Policy)); err != nil {
```
NDStrahilevitz, line 299: With the new generic change the type inference can be dropped right?

geyslan: Yep, must be. I'll rebase it. Tks.

PR #3972, opened 2024-04-11
fix: capture io

yanivagman:

commit c3fc330 accidentally deleted vfs_read/write arguments which are also used by the capture tail call. Let's restore them.

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -3028,19 +3028,17 @@ do_file_io_operation(struct pt_regs *ctx, u32 event_id, u32 tail_call_id, bool i
         // missed entry or not traced
         return 0;
     }
-    del_args(event_id);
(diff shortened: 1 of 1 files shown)


pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3028,17 +3028,20 @@ do_file_io_operation(struct pt_regs *ctx, u32 event_id, u32 tail_call_id, bool i
...
+    // We shouldn't call del_args(event_id) here as the arguments are also used by the tail call
 
     program_data_t p = {};
-    if (!init_program_data(&p, ctx))
+    if (!init_program_data(&p, ctx)) {
+        del_args(event_id);
         return 0;
+    }
 
     if (!should_trace(&p))
```
geyslan, line 3039: Shouldn't we also `del_args()` in `!should_trace()`?

geyslan: What about a goto bail?

geyslan (approved):
LGTM

PR #3975, opened 2024-04-14
chore: remove OPT_PROCESS_INFO

yanivagman:

OPT_PROCESS_INFO is not used anymore and can't be selected by the user. Remove old code that still uses it.


geyslan (approved):
LGTM

PR #3981, opened 2024-04-16
fix(analyze): bind flags with viper

### LINKED ISSUE #3980: Analyze mode flags don't work
## Description

For some reason the code that bind the analyze mode flags to the viper command have been commented out by this commit - 2555190ad647805433a34a9b8a933f8816ba3619.

This is the comment left about this descition:
```Go
	// TODO: decide if we want to bind this flag to viper, since we already have a similar
	// flag in rootCmd, conflicting with each other.
	// The same goes for the other flags (signatures-dir, rego), also in rootCmd.
```
The answer here is yes, because the viper bind is done per command, so we need to bind it to the analyze command that run.
The result is that the flags of the analyze command are not usable.
On the way I propose to add a log flag too.

```
(paste your output here)
```

## Output of `uname -a`:

```
(paste your output here)
```

AlonZivony:

Restore the binding of the analyze mode flags with viper. This should fix the issue that they weren't usable since they were commented out.

Fix #3980

### DIFF
--- a/cmd/tracee/cmd/analyze.go
+++ b/cmd/tracee/cmd/analyze.go
@@ -122,7 +122,11 @@ tracee analyze --events anti_debugging events.json`,
 			logger.Fatalw("No signature event loaded")
 		}
 
-		fmt.Printf("Loading %d signature events\n", len(sigs))
+		logger.Infow(
+			"Signatures loaded",
+			"total", len(sigs),
+			"signatures", getSigsNames(sigs),
+		)
 
 		_ = initialize.CreateEventsFromSignatures(events.StartSignatureID, sigs)
 
@@ -226,4 +230,17 @@ func bindViperFlag(cmd *cobra.Command, flag string) {
 	if err != nil {
 		logger.Fatalw("Error binding viper flag", "flag", flag, "error", err)
 	}
+}
+
+func getSigsNames(sigs []detect.Signature) []string {
+	var sigsNames []string
+	for _, sig := range sigs {
+		sigMeta, err := sig.GetMetadata()
+		if err != nil {
+			logger.Warnw("Failed to get signature metadata", "err", err)
+			continue
+		}
+		sigsNames = append(sigsNames, sigMeta.Name)
+	}
+	return sigsNames
 }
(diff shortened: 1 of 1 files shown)


geyslan:
@AlonZivony https://github.com/aquasecurity/tracee/actions/runs/8704678483/job/23873691072?pr=3981#step:4:1106

cmd/tracee/cmd/analyze.go
```diff
@@ -37,39 +37,26 @@ func init() {
...
-	// if err != nil {
-	// 	fmt.Fprintf(os.Stderr, "Error: %s\n", err)
-	// 	os.Exit(1)
-	// }
+
+	analyze.Flags().StringArrayP(
+		"log",
+		"l",
+		[]string{"info"},
+		"[debug|info|warn...]\t\tLogger options",
```
geyslan, line 58: The other flags must also be escaped with `\t` to align the column.

```
sudo ./dist/tracee analyze -h
Analyze allow you to explore signature events with past events.

...

Flags:
  -e, --events stringArray           Define which signature events to load
  -h, --help                         help for analyze
  -l, --log stringArray              [debug|info|warn...]		Logger options (default [info])
      --rego stringArray             Control event rego settings
      --signatures-dir stringArray   Directory where to search for signatures in OPA (.rego) and Go plugin (.so) formats
```

geyslan (approved):
LGTM, tested the returned flags and they seem to work nicely.

---

This seems not related to your PR but would be nice to have a better output in these cases:

```
sudo ./dist/tracee analyze
Error: accepts 1 arg(s), received 0
```

```
sudo ./dist/tracee analyze -l -e anti_debugging events.json
Error: accepts 1 arg(s), received 2
```

cmd/tracee/cmd/analyze.go
AlonZivony: Nice catch

AlonZivony:
> LGTM, tested the returned flags and they seem to work nicely.
> 
> This seems not related to your PR but would be nice to have a better output in these cases:
> 
> ```
> sudo ./dist/tracee analyze
> Error: accepts 1 arg(s), received 0
> ```
> 
> ```
> sudo ./dist/tracee analyze -l -e anti_debugging events.json
> Error: accepts 1 arg(s), received 2
> ```

The first case is OK in my opinion.
The second is unclear. We can just remove the args constraints, but it should be in another PR so it will have its own commit.

PR #3985, opened 2024-04-17
fix(ebpf): use debug error level instead of error

geyslan:

d621c3644 **fix(ebpf): use log debug level instead of error...**

```
... for internal requirements.

Considering that if GetLast() failed t.config.Policies is nil, this also
removes the use of t.config.Policies. Furthermore, the goal is to
concetrate all Policies access in the PolicyManager.
```


geyslan:
It's green in E2E 1465.

geyslan:
Merging it to unblock E2E.

PR #3989, opened 2024-04-18
feat(proctree): control procfs query by config

### LINKED ISSUE #3988: Process tree procfs scan can't be configured
## Description

We received some complaints about the performance of the process tree feature.
The main performance issue with it is the fact that it scans procfs dirs to recover data that is missing in it.
This is done mainly in the case of lost events, which means that there is a load on the system, so adding IO overhead is making things worst.
We should at least allow the user to choose if he wants this level of information reliability at the cost of the performance.

```
(paste your output here)
```

## Output of `uname -a`:

```
(paste your output here)
```

AlonZivony:

Allow the procfs scanning and querying to be configured by the user. For now Tracee only enable to disable the procfs querying upon lost events. This should help to improve performance on high-load systems.

Fix #3988


yanivagman (commented):
LGTM,
please document `disable-procfs-query` new flag exposed in the CLI

yanivagman (approved):
LGTM

PR #3995, opened 2024-04-19
Rename context filters to scope filters

yanivagman:

This commit updates the terminology within the Tracee codebase, replacing occurrences of "context filters" with "scope filters". By doing so, it enhances the clarity and intuitiveness of the user interface and documentation.

### 3. Other comments

Fix: #3996

### DIFF
--- a/docs/man/events.1
+++ b/docs/man/events.1
@@ -26,8 +26,8 @@ tracee \f[B]--events\f[R] [<event-name1(,[-]event-name2\&...)> |
 <set1(,[-]event-name1,[-]event-name2\&...)> |
 <event1.args.arg-field[=|!=]value> |
 <event1.retval[=|!=|<|>|<=|>=]value> |
-<event1.context.context-field[=|!=|<|>|<=|>=]value> |
-<event.context.container>] \&...
+<event1.scope.scope-field[=|!=|<|>|<=|>=]value> |
+<event.scope.container>] \&...
 .SS DESCRIPTION
 .PP
 The \f[B]--events\f[R] flag allows you to select which events to trace
@@ -49,8 +49,8 @@ Event return value: Filter events based on their return value using
 The event return value expression follows the syntax of a numerical
 expression.
 .IP \[bu] 2
-Event context fields: Filter events based on the non-argument fields
-defined in the trace.Event struct using `event-name.context.field'.
+Event scope fields: Filter events based on the non-argument fields
+defined in the trace.Event struct using `event-name.scope.field'.
 Refer to the json tags in the trace.Event struct located in the
 types/trace package for the correct field names, and the event filtering
 section in the documentation for a full list.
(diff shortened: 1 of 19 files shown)


docs/man/events.1
```diff
@@ -49,8 +49,8 @@ Event return value: Filter events based on their return value using
 The event return value expression follows the syntax of a numerical
 expression.
 .IP \[bu] 2
-Event context fields: Filter events based on the non-argument fields
-defined in the trace.Event struct using `event-name.context.field'.
+Event scope fields: Filter events based on the non-argument fields
+defined in the trace.Event struct using `event-name.scope.field'.
```
geyslan, line 53: I think that this assertion can be a bit misleading, I mean many of the former context fields are part of the trace.Event. WDYT?

pkg/cmd/flags/policy.go
```diff
@@ -81,13 +81,13 @@ func PrepareFilterMapsFromPolicies(policies []k8s.PolicyInterface) (PolicyScopeM
 					continue
 				}
 
-				// at this point we know the filter is an event context filter
-				// context filters are provided without "context." prefix so we need to add it
-				evtContextFlags, err := parseEventFlag(fmt.Sprintf("%s.context.%s", r.Event, f))
+				// at this point we know the filter is an event scope filter
+				// scope filters are provided without "scope." prefix so we need to add it
+				evtScopeFlags, err := parseEventFlag(fmt.Sprintf("%s.scope.%s", r.Event, f))
```
geyslan, line 86: I think we should name it `evtScopeOpts` differentiating it from the `--scope` flags naming.

pkg/ebpf/events_pipeline.go
```diff
@@ -325,8 +325,8 @@ func (t *Tracee) matchPolicies(event *trace.Event) uint64 {
 		// Do the userland filtering
 		//
 
-		// 1. event context filters
-		if !p.ContextFilter.Filter(*event) {
+		// 1. event scope filters
+		if !p.ScopeFilter.Filter(*event) {
```
geyslan, line 329:
```suggestion
		if !p.EventScopeFilter.Filter(*event) {
```
Based on my assumption of differentiating it above. WDYT?

geyslan (approved):
LGTM. Put just some naming suggestions.

docs/man/events.1
yanivagman: We are renaming context filters to scope filters, and I didn't change anything in this comment.
Isn't it still true that scope filters are to filter non-argument fields in the trace.Event? Why is it misleading?

pkg/cmd/flags/policy.go
yanivagman: I think having evt (enent) in the name already differentiates it from the "global" policy scope, does't it?

geyslan: You're right. I concur.

docs/man/events.1
geyslan: Yep, it's not your changes, I just glanced at that sentence and, thinking as a new user, didn't quickly understand the meaning. We can see how to improve it in the future. It's definitely not related to this PR.

PR #3997, opened 2024-04-19
chore: refactor equality computation

geyslan:

405b8bedf **chore: refactor equality computation**

```
This moves the equality logic to equality.go, and uses Generics to
deduplicate the code.

This also changes procTreeEqualities map key to uint32, the right type
for the key.
```


yanivagman (approved):
LGTM

PR #3999, opened 2024-04-23
Bumps

geyslan:

8c62b638a **chore(deps): bump go toolchain to 1.22.3**
2ac1f8565 **chore(builder): bump pandoc to 3.2**
396cbd5ed **chore(builder): pin protoc versions**
435b5932a **chore: use NewClient instead of Dial (deprecated)**
a9aacbfbd **chore: bump go.mod and go.sum**
8d69d7d6e **chore(types): bump go.mod and go.sum**
b99be7a4d **chore(signatures): bump go.mod and go.sum**
94ae61cfa **chore(api): bump go.mod and go.sum**

2ac1f8565 **chore(builder): bump pandoc to 3.2**

```
It also updates all man by generating them with the new pandoc version.
```

396cbd5ed **chore(builder): pin protoc versions**

```
This doesn't upgrade versions to avoid changes in the protoc output
which would require updating the generated code.
```

435b5932a **chore: use NewClient instead of Dial (deprecated)**

```
grpc pkg introduced NewClient in v1.63.0.

SA1019  grpc.Dial is deprecated: use NewClient instead.  Will be supported throughout 1.x.
```

a9aacbfbd **chore: bump go.mod and go.sum**

```
Bump go minimum version to 1.22.0.
```

8d69d7d6e **chore(types): bump go.mod and go.sum**

```
Bump go minimum version to 1.22.0.
```

b99be7a4d **chore(signatures): bump go.mod and go.sum**

```
Bump go minimum version to 1.22.0.
```

94ae61cfa **chore(api): bump go.mod and go.sum**

```
Bump go minimum version to 1.22.0.
```


geyslan:
https://github.com/aquasecurity/tracee/actions/runs/8801749859/job/24155898433?pr=3999#step:7:908

That error is being tackled: https://github.com/kubernetes-sigs/controller-runtime/issues/2788#issuecomment-2072347985

Keeping it as draft until the solution is up.

geyslan:
After this be merged. A sequential PR should be created bumping the main `go.mod`, specifically:

```
github.com/aquasecurity/tracee/api
github.com/aquasecurity/tracee/signatures/helpers
github.com/aquasecurity/tracee/types
```

So, only after that, we should update internal tests.

geyslan:
@josedonizetti you may notice in f2d6129 that I pinned the protoc version to ensure that the generated files are not modified. But I would like to know if there is no synchronization problem with the client side when updating the versions of these files. If there's, what should be the steps to take? Or would you take care of bumping the protoc in a different PR?

api/go.mod
```diff
@@ -1,16 +1,15 @@
 module github.com/aquasecurity/tracee/api
 
-go 1.21
+go 1.22.3
```
yanivagman, line 3: Maybe a silly question, but why do we need to specify the patch version? won't `1.22` be enough?

rscampos (approved):
LGTM

geyslan:
@rscampos Thanks for reviewing it. I'm holding it for a while since it's good to have @NDStrahilevitz and @josedonizetti feedback as well, since it will demand changes to internal tests and beyond.

types/go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/aquasecurity/tracee/types
 
-go 1.21
+go 1.22.0
```
NDStrahilevitz, line 3: Shouldn't we upgrade to go1.22.3 considering the toolchain directive below?

builder/Dockerfile.protoc
```diff
@@ -1,11 +1,14 @@
...
-	unzip protoc-23.4-linux-x86_64.zip -d /usr/local
+RUN wget https://github.com/protocolbuffers/protobuf/releases/download/v${PROTOC_VERSION}/${PROTOC_ZIP} && \
+	unzip ${PROTOC_ZIP} -d /usr/local
 
-RUN go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.28 && \
-	go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.2 && \
-	go install github.com/mitchellh/protoc-gen-go-json@master
+RUN go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.28.1 && \
+	go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.2.0 && \
+	go install github.com/mitchellh/protoc-gen-go-json@49905733154f04e47d685de62c2cc2b72613b69e # master
```
NDStrahilevitz, line 12: nit: no version tag that can be given here?

NDStrahilevitz (approved):

types/go.mod
geyslan: Good question, it was my first change - if go and toolchain are the same, the toolchain will be omitted.

But I'm trying to establish a pattern of updates by saying that we are migrating to 1.22.0 (a newer version of go), while at the same time enforcing the use of a new version in the development environment and CI/CD pipelines without requiring immediately all users or environments upgrade - toolchain configuration will automatically download and use it for building without changing the user's environment.

This allows us to bring in a new toolchain with each patch release and ensures that the codebase remains compatible with older versions during the migration process.

NDStrahilevitz: Ok, sounds good to me.

builder/Dockerfile.protoc
geyslan: It went unnoticed. Solved. Tks.

api/go.mod
geyslan: I decided to use toolchain only in the main go.mod. The others when imported by tracee will have already the toolchain version enforced.

PR #4000, opened 2024-04-23
Wait for apt locks

geyslan:

d4a0dfa0a **fix(tests): address shellcheck SC1091**

```
Silence shellcheck SC1091 in e2e-install-deps.sh.
```

95ea66abd **fix(tests): address shellcheck SC2001**

```
- SC2001: See if you can use ${variable//search/replace} instead.

In this was used % (shortest match - suffix) since it's more appropriate
for the use case.
```

3700b4018 **fix(tests): address shellcheck SC2005**

```
- SC2005: Useless echo? Instead of 'cmd; echo $?' just use 'cmd'.
```

d06ef237f **fix(tests): address shellcheck SC2086**

```
- SC2086: Double quote to prevent globbing and word splitting.
```

a4475b83a **fix(tests): call wait_for_apt_locks before apt-get**

```
Reduce the window of opportunity for apt-get to fail due to locks by
calling wait_for_apt_locks before each apt-get command.
```

### DIFF
--- a/tests/e2e-install-deps.sh
+++ b/tests/e2e-install-deps.sh
@@ -195,7 +195,7 @@ install_clang_from_github() {
         LLVM_URL=$LLVM_URL"clang+llvm-14.0.6-aarch64-linux-gnu.tar.xz"
     fi
 
-    LLVM_FILE=$(echo $(basename $LLVM_URL))
+    LLVM_FILE=$(basename $LLVM_URL)
     LLVM_DIR=$(echo $LLVM_FILE | sed 's:.tar.xz::g')
 
     # Download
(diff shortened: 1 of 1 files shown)


geyslan:
https://github.com/aquasecurity/tracee/commit/a4475b83a5705300712afd416f3eab92506525a8 is a fix attempt for https://github.com/aquasecurity/tracee/pull/3945#issuecomment-2072302792

The other commits are timely.

rscampos:
The shellcheck fixes seem to be OK, I've double checked each one. Will take a look wait_for_apt_locks before apt's.

rscampos (approved):
LGTM

PR #4003, opened 2024-04-24
chore(ebpf): use core type size checks

### LINKED ISSUE #3774: Tracee should always rely on get_type_size() instead of sizeof() for correct CORE type sizes
There are many places where sizeof() is used when it shouldn't. The type being measured by sizeof() has to be either a built-in type (int/long/...) or a local type (fully declared in a header). There are many places where the type being "sized" is a CO-RE (possible CO-RE) type:

EXamples:

```diff
diff --git a/pkg/ebpf/c/common/buffer.h b/pkg/ebpf/c/common/buffer.h
index 49d374272..6d32d421e 100644
--- a/pkg/ebpf/c/common/buffer.h
+++ b/pkg/ebpf/c/common/buffer.h
@@ -29,7 +29,7 @@ statfunc buf_t *get_buf(int idx)
 }
 
 // biggest elem to be saved with 'save_to_submit_buf' should be defined here:
-#define MAX_ELEMENT_SIZE sizeof(struct sockaddr_un)
+#define MAX_ELEMENT_SIZE get_type_size(struct sockaddr_un)
 
 statfunc int save_to_submit_buf(args_buffer_t *buf, void *ptr, u32 size, u8 index)
 {
@@ -319,7 +319,7 @@ statfunc int save_sockaddr_to_buf(args_buffer_t *buf, struct socket *sock, u8 in
         get_network_details_from_sock_v4(sk, &net_details, 0);
         get_local_sockaddr_in_from_network_details(&local, &net_details, family);
 
-        save_to_submit_buf(buf, (void *) &local, sizeof(struct sockaddr_in), index);
+        save_to_submit_buf(buf, (void *) &local, get_type_size(struct sockaddr_in), index);
     } else if (family == AF_INET6) {
         net_conn_v6_t net_details = {};
         struct sockaddr_in6 local;
@@ -327,11 +327,11 @@ statfunc int save_sockaddr_to_buf(args_buffer_t *buf, struct socket *sock, u8 in
         get_network_details_from_sock_v6(sk, &net_details, 0);
         get_local_sockaddr_in6_from_network_details(&local, &net_details, family);
 
-        save_to_submit_buf(buf, (void *) &local, sizeof(struct sockaddr_in6), index);
+        save_to_submit_buf(buf, (void *) &local, get_type_size(struct sockaddr_in6), index);
     } else if (family == AF_UNIX) {
         struct unix_sock *unix_sk = (struct unix_sock *) sk;
         struct sockaddr_un sockaddr = get_unix_sock_addr(unix_sk);
-        save_to_submit_buf(buf, (void *) &sockaddr, sizeof(struct sockaddr_un), index);
+        save_to_submit_buf(buf, (void *) &sockaddr, get_type_size(struct sockaddr_un), index);
     }
     return 0;
 }
@@ -397,13 +397,13 @@ statfunc int save_args_to_submit_buf(event_data_t *event, args_t *args)
                     bpf_probe_read(&family, sizeof(short), (void *) args->args[i]);
                     switch (family) {
                         case AF_UNIX:
-                            size = sizeof(struct sockaddr_un);
+                            size = get_type_size(struct sockaddr_un);
                             break;
                         case AF_INET:
-                            size = sizeof(struct sockaddr_in);
+                            size = get_type_size(struct sockaddr_in);
                             break;
                         case AF_INET6:
-                            size = sizeof(struct sockaddr_in6);
+                            size = get_type_size(struct sockaddr_in6);
                             break;
                         default:
                             size = sizeof(short);
@@ -419,7 +419,7 @@ statfunc int save_args_to_submit_buf(event_data_t *event, args_t *args)
                 rc = save_to_submit_buf(&(event->args_buf), (void *) (args->args[i]), size, index);
                 break;
             case TIMESPEC_T:
-                size = sizeof(struct __kernel_timespec);
+                size = get_type_size(struct __kernel_timespec);
                 rc = save_to_submit_buf(&(event->args_buf), (void *) (args->args[i]), size, index);
                 break;
         }
diff --git a/pkg/ebpf/c/common/filesystem.h b/pkg/ebpf/c/common/filesystem.h
index 6cd6b5b4e..43d1931ef 100644
--- a/pkg/ebpf/c/common/filesystem.h
+++ b/pkg/ebpf/c/common/filesystem.h
@@ -171,14 +171,14 @@ statfunc size_t get_path_str_buf(struct path *path, buf_t *out_buf)
     }
 
     struct path f_path;
-    bpf_probe_read(&f_path, sizeof(struct path), path);
+    bpf_probe_read(&f_path, get_type_size(struct path), path);
     char slash = '/';
     int zero = 0;
     struct dentry *dentry = f_path.dentry;
     struct vfsmount *vfsmnt = f_path.mnt;
     struct mount *mnt_parent_p;
     struct mount *mnt_p = real_mount(vfsmnt);
-    bpf_probe_read(&mnt_parent_p, sizeof(struct mount *), &mnt_p->mnt_parent);
+    bpf_probe_read(&mnt_parent_p, sizeof(void *), &mnt_p->mnt_parent);
     u32 buf_off = (MAX_PERCPU_BUFSIZE >> 1);
     struct dentry *mnt_root;
     struct dentry *d_parent;
```

This can lead to bad logic or verifier issues due to bad sizing of copies.

NDStrahilevitz:

115610859 **chore(ebpf): replace sizeof with bpf type helpers**

```
When using sizeof on kernel structs, false results may happen due to the
sizeof builtin not being designed to take into account btf relocations.

Use the libbpf helpers bpf_core_type_size and bpf_core_field_size where
relevant.
```

cfd849e5c **chore(ebpf): remove get_type_size macro**

```
Use the builtin libbpf core helper instead, which
the macro was an alias of.
```

### 3. Other comments

Resolve #3774

### DIFF
--- a/pkg/ebpf/c/common/filesystem.h
+++ b/pkg/ebpf/c/common/filesystem.h
@@ -171,7 +171,7 @@ statfunc size_t get_path_str_buf(struct path *path, buf_t *out_buf)
     }
 
     struct path f_path;
-    bpf_probe_read_kernel(&f_path, sizeof(struct path), path);
+    bpf_probe_read_kernel(&f_path, bpf_core_type_size(struct path), path);
     char slash = '/';
     int zero = 0;
     struct dentry *dentry = f_path.dentry;
--- a/pkg/ebpf/c/common/filesystem.h
+++ b/pkg/ebpf/c/common/filesystem.h
@@ -170,12 +170,10 @@ statfunc size_t get_path_str_buf(struct path *path, buf_t *out_buf)
         return 0;
     }
 
-    struct path f_path;
-    bpf_probe_read_kernel(&f_path, bpf_core_type_size(struct path), path);
     char slash = '/';
     int zero = 0;
-    struct dentry *dentry = f_path.dentry;
-    struct vfsmount *vfsmnt = f_path.mnt;
+    struct dentry *dentry = BPF_CORE_READ(path, dentry);
+    struct vfsmount *vfsmnt = BPF_CORE_READ(path, mnt);
     struct mount *mnt_parent_p;
     struct mount *mnt_p = real_mount(vfsmnt);
     bpf_core_read(&mnt_parent_p, sizeof(struct mount *), &mnt_p->mnt_parent);
--- a/pkg/ebpf/c/common/network.h
+++ b/pkg/ebpf/c/common/network.h
@@ -360,7 +360,7 @@ statfunc volatile unsigned char get_sock_state(struct sock *sock)
 statfunc struct ipv6_pinfo *get_inet_pinet6(struct inet_sock *inet)
 {
     struct ipv6_pinfo *pinet6_own_impl;
-    bpf_core_read(&pinet6_own_impl, sizeof(pinet6_own_impl), &inet->pinet6);
+    bpf_core_read(&pinet6_own_impl, sizeof(struct ipv6_pinfo *), &inet->pinet6);
     return pinet6_own_impl;
 }
 
@@ -369,6 +369,7 @@ statfunc struct sockaddr_un get_unix_sock_addr(struct unix_sock *sock)
     struct unix_address *addr = BPF_CORE_READ(sock, addr);
     int len = BPF_CORE_READ(addr, len);
     struct sockaddr_un sockaddr = {};
+    // NOTE(nadav.str): stack allocated, so runtime core size check is avoided
     if (len <= sizeof(struct sockaddr_un)) {
         bpf_probe_read(&sockaddr, len, addr->name);
     }
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -5849,11 +5849,11 @@ int BPF_KPROBE(cgroup_bpf_run_filter_skb)
     switch (family) {
         case PF_INET:
             eventctx->retval |= family_ipv4;
-            l3_size = get_type_size(struct iphdr);
+            l3_size = bpf_core_type_size(struct iphdr);
             break;
         case PF_INET6:
             eventctx->retval |= family_ipv6;
-            l3_size = get_type_size(struct ipv6hdr);
+            l3_size = bpf_core_type_size(struct ipv6hdr);
             break;
         default:
             return 1;
@@ -5880,7 +5880,7 @@ int BPF_KPROBE(cgroup_bpf_run_filter_skb)
                 return 1;
 
             if (nethdrs->iphdrs.iphdr.ihl > 5) { // re-read IP header if needed
-                l3_size -= get_type_size(struct iphdr);
+                l3_size -= bpf_core_type_size(struct iphdr);
                 l3_size += nethdrs->iphdrs.iphdr.ihl * 4;
                 bpf_core_read(nethdrs, l3_size, data_ptr);
             }
@@ -5978,11 +5978,11 @@ statfunc u32 cgroup_skb_generic(struct __sk_buff *ctx, void *cgrpctxmap)
     switch (family) {
         case PF_INET:
             dest = &nethdrs->iphdrs.iphdr;
-            size = get_type_size(struct iphdr);
+            size = bpf_core_type_size(struct iphdr);
             break;
         case PF_INET6:
             dest = &nethdrs->iphdrs.ipv6hdr;
-            size = get_type_size(struct ipv6hdr);
+            size = bpf_core_type_size(struct ipv6hdr);
             break;
         default:
             return 1; // verifier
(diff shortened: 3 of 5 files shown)


NDStrahilevitz:
This breaks verifier right now :/

pkg/ebpf/c/common/network.h
```diff
@@ -369,7 +369,7 @@ statfunc struct sockaddr_un get_unix_sock_addr(struct unix_sock *sock)
     struct unix_address *addr = BPF_CORE_READ(sock, addr);
     int len = BPF_CORE_READ(addr, len);
     struct sockaddr_un sockaddr = {};
-    if (len <= sizeof(struct sockaddr_un)) {
+    if (len <= bpf_core_type_size(struct sockaddr_un)) {
```
yanivagman, line 372: ditto

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2641,7 +2644,7 @@ int BPF_KPROBE(trace_security_socket_connect)
         // Workaround for sockaddr_un struct length (issue: #1129).
         struct sockaddr_un sockaddr = {0};
         bpf_probe_read(&sockaddr, (u32) addr_len, (void *) address);
-        stsb(args_buf, (void *) &sockaddr, sizeof(struct sockaddr_un), 2);
+        stsb(args_buf, (void *) &sockaddr, bpf_core_type_size(struct sockaddr_un), 2);
```
yanivagman, line 2647: ditto

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2764,14 +2769,19 @@ int BPF_KPROBE(trace_security_socket_bind)
         }
     } else if (sa_fam == AF_UNIX) {
 #if defined(__TARGET_ARCH_x86) // TODO: this is broken in arm64 (issue: #1129)
+        // NOTE(nadav.str): sizeof used instead of bpf_core_type_size in this case.
+        // bpf_core_type_size is evaluated at runtime. thus the verifier isn't satisfied
+        // with the boundary established below for addr_len. therefore, bpf_probe_read
+        // causes the verifier to fail.
         if (addr_len <= sizeof(struct sockaddr_un)) {
             struct sockaddr_un sockaddr = {};
```
yanivagman, line 2777: ditto

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2871,7 +2881,8 @@ statfunc u32 send_bin_helper(void *ctx, void *prog_array, int tail_call)
             // Handle the rest of write recursively
             bin_args->start_off += bin_args->full_size;
             struct iovec io_vec;
-            bpf_probe_read(&io_vec, sizeof(struct iovec), &bin_args->vec[bin_args->iov_idx]);
+            bpf_probe_read(
```
yanivagman, line 2884: ditto

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2958,7 +2969,8 @@ statfunc u32 send_bin_helper(void *ctx, void *prog_array, int tail_call)
         // Handle the rest of write recursively
         bin_args->start_off += bin_args->full_size;
         struct iovec io_vec;
-        bpf_probe_read(&io_vec, sizeof(struct iovec), &bin_args->vec[bin_args->iov_idx]);
+        bpf_probe_read(
```
yanivagman, line 2972: ditto

yanivagman (commented):
In some places we read a struct into a stack variable which was preallocated by the compiler according to the known size at compile time. I believe this may cause future issues and we should avoid that (so just leave as is for now)

pkg/ebpf/c/common/filesystem.h
```diff
@@ -171,7 +171,7 @@ statfunc size_t get_path_str_buf(struct path *path, buf_t *out_buf)
     }
 
     struct path f_path;
-    bpf_probe_read_kernel(&f_path, sizeof(struct path), path);
+    bpf_probe_read_kernel(&f_path, bpf_core_type_size(struct path), path);
```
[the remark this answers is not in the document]
NDStrahilevitz: Ok, seems reasonable to move back to `sizeof` in order to correspond with the the compiler preallocation.
However, doesn't this imply that the compiler might preallocate the wrong size?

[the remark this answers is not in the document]
NDStrahilevitz: Actually, it seems that in this function we can and should just rely on BPF_CORE_READ, and remove the probe_read on the struct itself, WDYT?

NDStrahilevitz:
> I'm not sure it's a good idea to do that here.
The compiler preallocates space on the stack for f_path using the size of struct path known at compile time.
If the size of struct path is different on the user environment, it can override other other variables on the stack

@yanivagman it seems to me we should rather avoid stack allocations for kernel structs. That is, it seems that if sizeof of the stack allocated variable may turn out wrong across kernels, then the stack allocation size itself might be wrong. Not even sure we can rely on the stack allocation to be the maximum bound (then it wouldn't be such a problem), because we rely on a minimal vmlinux.h file.
Question is, is my assessment correct, and if so, should we make the effort in the scope of this PR?

pkg/ebpf/c/common/filesystem.h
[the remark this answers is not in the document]
NDStrahilevitz: Made the above change.

yanivagman (approved):
LGTM

PR #4004, opened 2024-04-26
fix event definitions api

josedonizetti:
Fix event definitions api

https://github.com/aquasecurity/tracee/issues/4005


geyslan (approved):
LGTM

PR #4010, opened 2024-04-28
fix(analyze): support signatures based on signatures

### LINKED ISSUE #4009: Analyze mode doesn't support sigs-on-sigs
## Description

The current tracee analyze mode is not well maintained or tested, so the sigs-on-sigs option is not yet supported yet.
For compatibility we should support it for now until the full analyze mode will be implemented.

```
 v0.20.0-93-ge8d7a3516
```

AlonZivony:

Support signatures based on signatures in the analyze mode.
This is done in the purpose of making all signatures that are compatible with tracee to be also compatible with the analyze mode.

Fix #4009

### DIFF
--- a/cmd/tracee/cmd/analyze.go
+++ b/cmd/tracee/cmd/analyze.go
@@ -240,7 +240,7 @@ func getSigsNames(sigs []detect.Signature) []string {
 			logger.Warnw("Failed to get signature metadata", "err", err)
 			continue
 		}
-		sigsNames = append(sigsNames, sigMeta.Name)
+		sigsNames = append(sigsNames, sigMeta.ID)
 	}
 	return sigsNames
 }
--- a/cmd/tracee/cmd/analyze.go
+++ b/cmd/tracee/cmd/analyze.go
@@ -1,36 +1,27 @@
 package cmd
 
 import (
-	"bufio"
-	"context"
-	"encoding/json"
-	"fmt"
 	"os"
-	"os/signal"
-	"syscall"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 
+	"github.com/aquasecurity/tracee/pkg/analyze"
 	"github.com/aquasecurity/tracee/pkg/cmd/flags"
 	"github.com/aquasecurity/tracee/pkg/cmd/initialize"
-	tracee "github.com/aquasecurity/tracee/pkg/ebpf"
 	"github.com/aquasecurity/tracee/pkg/events"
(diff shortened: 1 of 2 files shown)


cmd/tracee/cmd/analyze.go
```diff
@@ -240,7 +240,7 @@ func getSigsNames(sigs []detect.Signature) []string {
 			logger.Warnw("Failed to get signature metadata", "err", err)
 			continue
 		}
-		sigsNames = append(sigsNames, sigMeta.Name)
+		sigsNames = append(sigsNames, sigMeta.ID)
```
geyslan, line 310: I think this change would require changing getSigsNames to getSigsIDs, wouldn't?

geyslan (commented):
On a cursory review, it looks ok, even though I haven't tested it yet.

How about moving the pipeline logic into a new file like `analyze_pipeline.go`?

cmd/tracee/cmd/analyze.go
AlonZivony: You are right👌

geyslan:
Hey @AlonZivony I think this is good for merging right? Could you just rebase again?

pkg/analyze/pipeline.go
```diff
@@ -0,0 +1,190 @@
...
+			case event := <-sigsEvents:
+				printInput <- event
+				engineInput <- event
+			case <-ctx.Done():
+				return
+			}
+		}
+	}()
+
+	go sigEngine.Start(ctx)
```
geyslan, line 72: Is missing a wg.Add(1)?

pkg/analyze/pipeline.go
```diff
@@ -0,0 +1,190 @@
...
+// printEvents starts a goroutine that prints the received events if they are chosen.
+// It is similar to the Sink stage in the normal pipeline.
+func printEvents(ctx context.Context, wg *sync.WaitGroup, eventsToPrint []string, eventsInput <-chan protocol.Event) {
+	wg.Add(1)
+
+	eventsToPrintMap := make(map[string]struct{}, len(eventsToPrint))
+	for _, eventName := range eventsToPrint {
+		eventsToPrintMap[eventName] = struct{}{}
+	}
+	go func() {
```
geyslan, line 162: nit: move the (156 line) wg.Add(1) to here.

AlonZivony:
I close this PR as we plan to merge the full analyze mode, which will fix the issue.

PR #4011, opened 2024-04-29
[v0.21.0] fix(events): fix process_execute_failed missing symbol for new kernels

geyslan:

commit 8f9660686dc62717567d68b3985c2cd0b5275979

    fix(events): change process_execute_failed probes
    
    The previous probe was missing from different distros and kernels.
    The new probes are safer, but only exist starting from v5.8 of the kernel.
    
    commit: a496a9e (main), cherry-pick

commit c8cf0412a47c8819fddbf20a279ef07c4991dc8f

    fix(events): create usermode events for submit
    
    Fix an issue that usermode events are only created if they should be
    emitted.
    
    commit: cc1eea7 (main), cherry-pick


rscampos (approved):
LGTM

Issue #3475, opened 2023-09-18
Check type cast results


geyslan:
## Description

It's important to check type cast results to avoid panic.

Revive `unchecked-type-assertion` rule provided this output:

```
make check-lint                            
Linting golang code...
pkg/signatures/benchmark/signature/golang/code_injection.go:94:20: type cast result is unchecked in request.Value.(string) - type assertion will panic if not matched
pkg/signatures/celsig/wrapper/wrapper.go:113:8: type cast result is unchecked in source.Value.(string) - type assertion will panic if not matched
pkg/signatures/celsig/wrapper/wrapper.go:118:8: type cast result is unchecked in source.Value.([]string) - type assertion will panic if not matched
pkg/signatures/celsig/wrapper/wrapper.go:125:9: type cast result is unchecked in source.Value.(uint32) - type assertion will panic if not matched
pkg/signatures/celsig/wrapper/wrapper.go:138:8: type cast result is unchecked in source.Value.(uint64) - type assertion will panic if not matched
pkg/signatures/celsig/wrapper/wrapper.go:143:8: type cast result is unchecked in source.Value.(int32) - type assertion will panic if not matched
pkg/signatures/celsig/wrapper/wrapper.go:148:8: type cast result is unchecked in source.Value.(int64) - type assertion will panic if not matched
pkg/signatures/celsig/wrapper/wrapper.go:153:8: type cast result is unchecked in source.Value.(map[string]string) - type assertion will panic if not matched
pkg/signatures/benchmark/signature/golang/anti_debugging.go:59:19: type cast result is unchecked in request.Value.(string) - type assertion will panic if not matched
pkg/signatures/celsig/signature.go:72:13: type cast result is unchecked in out.Value().(bool) - type assertion will panic if not matched
pkg/signatures/celsig/library.go:67:11: type cast result is unchecked in lhs.Value().(*wrapper.Event) - type assertion will panic if not matched
pkg/signatures/celsig/library.go:68:13: type cast result is unchecked in rhs.Value().(string) - type assertion will panic if not matched
pkg/signatures/celsig/library.go:77:11: type cast result is unchecked in lhs.Value().(*wrapper.Event) - type assertion will panic if not matched
pkg/signatures/celsig/library.go:78:13: type cast result is unchecked in rhs.Value().(string) - type assertion will panic if not matched
[...]

PR #4014, opened 2024-05-02
chore: code refactor to add assertion

rscampos:

fix: #3475 

3645eaa16 **chore: code refactor to add assertion**
e1ec4b2a8 **chore: enable unchecked-type-assertion**

3645eaa16 **chore: code refactor to add assertion**

```
Refactored code to fix the unchecked type cast issue by
incorporating error handling after the type assertion.
```

### DIFF
--- a/pkg/server/grpc/tracee.go
+++ b/pkg/server/grpc/tracee.go
@@ -345,20 +345,23 @@ func getThreat(description string, metadata map[string]interface{}) *pb.Threat {
 }
 
 func getSeverity(metadata map[string]interface{}) pb.Severity {
-	switch metadata["Severity"].(int) {
-	case 0:
-		return pb.Severity_INFO
-	case 1:
-		return pb.Severity_LOW
-	case 2:
-		return pb.Severity_MEDIUM
-	case 3:
-		return pb.Severity_HIGH
-	case 4:
-		return pb.Severity_CRITICAL
+	severityValue, ok := metadata["Severity"].(int)
+	if ok {
+		switch severityValue {
+		case 0:
+			return pb.Severity_INFO
+		case 1:
+			return pb.Severity_LOW
+		case 2:
+			return pb.Severity_MEDIUM
+		case 3:
+			return pb.Severity_HIGH
+		case 4:
+			return pb.Severity_CRITICAL
+		}
 	}
 
-	return -1
+	return pb.Severity_INFO
 }
 
 func getStackAddress(stackAddresses []uint64) []*pb.StackAddress {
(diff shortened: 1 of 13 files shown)


pkg/server/grpc/tracee.go
```diff
@@ -345,17 +345,20 @@ func getThreat(description string, metadata map[string]interface{}) *pb.Threat {
...
-	case 1:
-		return pb.Severity_LOW
-	case 2:
-		return pb.Severity_MEDIUM
-	case 3:
-		return pb.Severity_HIGH
-	case 4:
-		return pb.Severity_CRITICAL
+	severityValue, ok := metadata["Severity"].(int)
+	if ok {
```
yanivagman, line 349: nit:
```suggestion
	if !ok {
	    return -1
	}
```

yanivagman (approved):
One nit comment, otherwise LGTM

pkg/server/grpc/tracee.go
rscampos: Thanks for the review. If we do so we still need do add another return in the final in case the switch is not satisfied (e.g: severityValue is 5).

Another thing, just realized it, but I think this func should follow the same logic as threat.pb.go:
https://github.com/aquasecurity/tracee/blob/cd6759e3c3e4abde2d50cf277fa59606461321b9/api/v1beta1/threat.pb.go#L136-L141

If any error is found, we should return Severity_INFO (0) and not (-1). If so, it should be:
```
func getSeverity(metadata map[string]interface{}) pb.Severity {
	severityValue, ok := metadata["Severity"].(int)
	if ok {
		switch severityValue {
		case 0:
			return pb.Severity_INFO
		case 1:
			return pb.Severity_LOW
		case 2:
			return pb.Severity_MEDIUM
		case 3:
			return pb.Severity_HIGH
		case 4:
			return pb.Severity_CRITICAL
		}
	}

	return pb.Severity_INFO
}
```

yanivagman: Right. The nit here was just about removing the extra indentation of the switch case

tests/integration/capture_test.go
rscampos, line 1: @yanivagman could you take quick glance on this file also?

pkg/server/grpc/tracee.go
```diff
@@ -345,20 +345,23 @@ func getThreat(description string, metadata map[string]interface{}) *pb.Threat {
...
+		case 2:
+			return pb.Severity_MEDIUM
+		case 3:
+			return pb.Severity_HIGH
+		case 4:
+			return pb.Severity_CRITICAL
+		}
+	}
+
+	return pb.Severity_INFO
```
rscampos, line 364: @yanivagman rework on this function and now is return `pb.Severity_INFO` as default

yanivagman: makes sense

PR #4015, opened 2024-05-02
fix(events): hidden_kernel_module - change history scan behaviour

OriGlassman:

The MAX_NUM_MODULES was recently decreased from 600 to 450 and is anyway relatively small. In case there are more modules to iterate on than the number of the iterations done, it will report the rest as false positives. To mitigate that, an LRU holds the suspected events and waits until the scan finishes: if the scan finished successfully, sends the events and otherwise deletes them.

This was introduced in kernels 6.5 where the mod_tree contains duplicate nodes, and 450 iterations may not be enough.

When it's possible to use bpf_loop (currently RHEL verifier fails if the helper exists in the code; specific program loading might solve it), this patch needs to be reverted and the patch that uses bpf_loop needs to be used instead.

### 2. Explain how to test it

### DIFF
--- a/pkg/events/derive/hidden_kernel_module.go
+++ b/pkg/events/derive/hidden_kernel_module.go
@@ -24,6 +24,7 @@ import (
 
 var (
 	foundHiddenKernModsCache *lru.Cache[uint64, struct{}]
+	eventsFromHistoryScan    *lru.Cache[*trace.Event, struct{}]
 	allModsMap               *bpf.BPFMap
 	newModuleOnlyMap         *bpf.BPFMap
 	recentDeletedModulesMap  *bpf.BPFMap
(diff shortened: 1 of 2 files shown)


pkg/events/derive/hidden_kernel_module.go
```diff
@@ -24,19 +24,21 @@ import (
 
 var (
 	foundHiddenKernModsCache *lru.Cache[uint64, struct{}]
+	eventsFromHistoryScan    *lru.Cache[*trace.Event, struct{}]
```
AlonZivony, line 27: I think initializing these values in the function as locals instead of globals will be easier to maintain and test.
As you are returning a function, you can initialize in it whatever values you want.

pkg/events/derive/hidden_kernel_module.go
```diff
@@ -62,65 +64,125 @@ func deriveHiddenKernelModulesArgs() deriveArgsFunction {
...
 
-		return []interface{}{addrHex, name, srcversion}, nil
+	return res, nil // Send all the events
+}
+
+// extractFromEvent extract arguments from the trace.Argument
+func extractFromEvent(args []trace.Argument, address uint64) []interface{} {
+	// Parse module name if possible
+	var name string
+	nameBytes, err := parse.ArgVal[[]byte](args, "name")
```
AlonZivony, line 158: The logic of parsing bytes and turning them to string can be extracted to a helper function.

yanivagman (approved):
LGTM

PR #4016, opened 2024-05-02
capture: fixes and tests

### LINKED ISSUE #4013: Capture IO is received empty
## Description

When using the capture of read or write, the captured files in Tracee's output directory are created but their content is missing.
After some digging into it, it seems to not be a new problem.
The capture have worked well in version 0.19.0.
However, at some point capture files were not written anymore at all, which was fixed in #3972.
This fix had only restored the capture files creation though, without restoring the content capturing.

```
(paste your output here)
```

## Output of `uname -a`:

```
(paste your output here)
```

NDStrahilevitz:

d92f13f43 **tests(integration): add capture tests**
8ad9959a4 **tests(inst): add vfs_writev test**
bb7cd569e **fix: restore io capture output**

bb7cd569e **fix: restore io capture output**

```
Commit 3ddf73d changed various probe_read calls to probe_read_kernel.
However, io capture relies on reading userspace buffers, as such the
pointers being read are userspace pointers. Therefore the relevant calls
in some of these changes, affecting IO capture, were changed to
probe_read_user.
```

### 2. Explain how to test it

1. With the new integration test
OR
1. Run tracee with `tracee -c read=/home/ubuntu/output* -c write=/home/ubuntu/output*`
2. `echo Hello > /home/ubuntu/output.txt`
3. `cat /home/ubuntu/output.txt`
4. Check `/tmp/tracee/out/host` for the relevant files and confirm their content (both should be `Hello`)

### 3. Other comments

Resolve #4013

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -757,7 +757,7 @@ $(OUTPUT_DIR)/syscaller: \
 test-integration: \
 	.checkver_$(CMD_GO) \
 	$(OUTPUT_DIR)/syscaller \
-	tracee-ebpf
+	tracee
 #
 	@$(GO_ENV_EBPF) \
 	$(CMD_GO) test \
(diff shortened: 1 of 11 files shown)


yanivagman (approved):
Good catch, LGTM,
For some reason, the github actions to run tests on a new PR are not working as expected, so we can't see if the new tests work as they should.
Let's wait to see the new tests pass as expected before merging this PR.
We also need to backport it to v0.21.0

Issue #3885, opened 2024-02-21
helm: rename `traceeConfig` to `traceeConfigFile`


josedonizetti:
There has been some confusion around [`traceeConfig`](https://github.com/aquasecurity/tracee/blob/main/deploy/helm/tracee/values.yaml#L59) option on the helm values, the idea behind it is to allow the user to inject a full config file, using `helm --set-file`, when configuring individual values, we can use https://github.com/aquasecurity/tracee/blob/main/deploy/helm/tracee/values.yaml#L61.

So, let's rename it to `traceeConfigFile` and add a comment explaining it on the values.yaml

rscampos:
@josedonizetti what do you think: instead of using `traceeConfigFile` we simply use `configFile`? I thought this way because the `config` (that already exists) don't have `tracee` in the field.

PR #4007, opened 2024-04-26
chore(k8s): prepare v0.21.0 release

geyslan:

PR #4018, opened 2024-05-02
chore(helm): rename helm field config file

rscampos:

fix: #3885 

799c68b3f **docs(helm): rename helm field config file**
e86f9f425 **chore(helm): rename helm field config file**

### DIFF
--- a/deploy/helm/tracee/templates/tracee-config.yaml
+++ b/deploy/helm/tracee/templates/tracee-config.yaml
@@ -6,9 +6,9 @@ metadata:
   labels:
     {{- include "tracee.labels" . | nindent 4 }}
 data:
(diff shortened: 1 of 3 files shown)


rscampos:
Think the test helm broken due that there isn't `tracee:0.21.0`

yanivagman:
> Think the test helm broken due that there isn't `tracee:0.21.0`

Yes, you are right. @geyslan I think we might need to revert https://github.com/aquasecurity/tracee/pull/4007 and merge it just before cutting the release

geyslan:
> Yes, you are right. @geyslan I think we might need to revert #4007 and merge it just before cutting the release

@rscampos please rebase against main to trigger the actions. Tks.

PR #4020, opened 2024-05-05
fix(events): hidden_kernel_module - change history scan behaviour

### LINKED ISSUE #4021: hidden kernel module potential FP
## Description

 The MAX_NUM_MODULES was recently decreased from 600 to 450 and is anyway relatively small.
  In case there are more modules to iterate on than the number of the
  iterations done, it will report the rest as false positives.
    

```
(paste your output here)
```

## Output of `uname -a`:

```
(paste your output here)
```

OriGlassman:

    fix(events): hidden_kernel_module - change history scan behaviour
    
    The MAX_NUM_MODULES was recently decreased from 600 to 450 and is anyway relatively small.
    In case there are more modules to iterate on than the number of the
    iterations done, it will report the rest as false positives.
    To mitigate that, an LRU holds the suspected events and waits until the
    scan finishes: if the scan finished successfully, sends the events and
    otherwise deletes them.
    
    This was introduced in kernels 6.5 where the mod_tree contains 3x nodes
    as before, and 450 iterations are not enough if there are about 450/3
    modules loaded in the system.
    
    When it's possible to use bpf_loop (currently RHEL verifier fails if the
    helper exists in the code; specific program loading might solve it),
    this patch needs to be reverted and the patch that uses bpf_loop needs
    to be used instead.

Closes #4021

### 2. Explain how to test it

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -715,7 +715,9 @@ int tracepoint__sched__sched_process_fork(struct bpf_raw_tracepoint_args *ctx)
 }
 
 // number of iterations - value that the verifier was seen to cope with - the higher, the better
-#define MAX_NUM_MODULES 450
+#define MAX_NUM_MODULES         450
+#define HISTORY_SCAN_FAILURE    0
+#define HISTORY_SCAN_SUCCESSFUL 1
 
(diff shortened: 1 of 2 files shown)


yanivagman (approved):
LGTM

PR #4022, opened 2024-05-05
feat(proctree): control procfs query by config

AlonZivony:

Allow the procfs scanning and querying to be configured by the user.
For now Tracee only enable to disable the procfs querying upon lost events.
This should help to improve performance on high-load systems.


yanivagman (approved):
LGTM

PR #4025, opened 2024-05-06
fix(ebpf): use kprobes for execute_finished

### LINKED ISSUE #4026: `execute_finished` tracepoints are missing when debugfs is not accessible
## Description

When running tracee with the `process_execute_failed` event on a container which is not accessible to debugfs, the following errors are printed:
```
libbpf: failed to determine tracepoint 'syscalls/sys_exit_execveat' perf event ID: No such file or directory
libbpf: prog 'execute_finished': failed to create tracepoint 'syscalls/sys_exit_execveat' perf event: No such file or directory
libbpf: failed to determine tracepoint 'syscalls/sys_exit_execve' perf event ID: No such file or directory
libbpf: prog 'execute_finished': failed to create tracepoint 'syscalls/sys_exit_execve' perf event: No such file or directory
```
Tracepoints seem to depend on debugfs, which is currently not a requirement of Tracee.
This should be fixed until we determine that we accept debugfs as a dependency.

commit f5876313a84346da38dc885947f046b1a82779bf

## Output of `uname -a`:

```
(paste your output here)
```

AlonZivony:

Recently, the process_execute_failed event implementation had been changed to use the new inner execute_finished event. This event has used syscall tracepoints in its implementation. However, tracepoints rely on debugfs, which was not a requirement of tracee until now. To remove this requirement (at least for now), move to use architecture-specific kprobes instead.

Fix #4026

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -5043,17 +5043,17 @@ int BPF_KPROBE(trace_security_bprm_creds_for_exec2)
     return execute_failed_tail2(ctx);
 }
 
-SEC("tracepoint/execute_finished")
-int execute_finished(struct sys_exit_tracepoint_args *args)
+SEC("kretprobe/execute_finished")
+int BPF_KPROBE(execute_finished)
 {
     program_data_t p = {};
-    if (!init_program_data(&p, args, EXECUTE_FINISHED))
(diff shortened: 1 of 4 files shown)


OriGlassman (approved):

yanivagman (approved):
LGTM
Please backport to v0.21.0 branch as well

PR #4027, opened 2024-05-07
fix: network event context

### LINKED ISSUE #3949: Regression in process info
## Description

A recent optimization effort (https://github.com/aquasecurity/tracee/commit/f806cb4b71b8f4082e182adf6c201c1e81f05753) moved some parts of the program data initialization to the event submission stage. This results in the information that is used to construct pcap file names to be missing (the process name portion of the pcap file is empty).

I also noticed that (somewhat expectedly), using process info inside a bpf program results in incorrect data. Some bpf programs may rely on this data which will result in unexpected behavior.

There may be more side effects for this change, anywhere that process info is used and isn't in a regular event submitted to user space.

Commit f806cb4b71b8f4082e182adf6c201c1e81f05753

NDStrahilevitz:

714777811 **fix(ebpf): initialize network event task context**

```
Commit f806cb4b changed context initialization to occur in the submit
stage. However, network events do not have access to that context during
their submit stage (since they happen in a cgroup_skb program).

Therefore, add a call to init_task_context in the pre cgroup bpf program
(cgroup_bpf_run_filter_skb).
```

37716feb2 **tests(network): add task context validation**

```
Since network events have a different eBPF program architecture, they
can't initialize the task context the standard way.
Therefore, add a validation in the e2e tests so that task context
initialization doesn't break.
```

### 2. Explain how to test it

Added validation in existing e2e network tests.

Manually test by running `tracee --events net_packet_http` and then `curl yourfavoritesite.com`

### 3. Other comments

Fix #3949

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -5920,6 +5920,10 @@ int BPF_KPROBE(cgroup_bpf_run_filter_skb)
 
     neteventctx.bytes = 0; // event arg size: no payload by default (changed inside skb prog)
 
+    // initialize task context before submit since it will not be available when
+    // submitting the network event.
+    init_task_context(&eventctx->task, p.event->task, p.config->options);
+
     // TODO: log collisions
     bpf_map_update_elem(cgrpctxmap, &indexer, &neteventctx, BPF_NOEXIST);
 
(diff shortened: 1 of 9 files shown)


yanivagman (approved):
LGTM
Needs backport to v0.21.0 branch

oshaked1 (approved):
LGTM, fixed the issue for me

PR #4030, opened 2024-05-08
fix(ebpf): use kprobes for execute_finished

AlonZivony:

Recently, the process_execute_failed event implementation had been changed to use the new inner execute_finished event.
This event has used syscall tracepoints in its implementation.
However, tracepoints rely on debugfs, which was not a requirement of tracee until now.
To remove this requirement (at least for now), move to use architecture-specific kprobes instead.

This is a cherry pick of #4025


yanivagman (approved):
LGTM

PR #4032, opened 2024-05-08
feat(events): use same tailcalls for process_execute_failed

AlonZivony:

Change process_execute_failed programs to use the same tailcall programs for both exec_binprm and security_bprm_cred_for_exec. This reduces code duplication and risk of errors


yanivagman (approved):
LGTM

PR #4034, opened 2024-05-09
fix: avoid logging warnings for non-ELF so loading

### LINKED ISSUE #4033: `symbols_loaded` log warning on parsing non-ELF files
## Description

We received the following WARN log:
```
symbols_loaded        {"object loaded": {"Id":{"Inode":19457,"Device":187,"Ctime":1714802203398053391},"Path":"/usr/lib/jvm/java-11-openjdk-amd64/lib/server/classes.jsa","MountNS":4026532917}, "error": "sharedobjs.(*HostSymbolsLoader).GetExportedSymbols: sharedobjs.(*HostSymbolsLoader).loadSOSymbols: sharedobjs.loadSharedObjectDynamicSymbols: bad magic number '[162 171 11 240]' in record at byte 0x0"}
```
It makes sense, as the loaded file is not ELF. Only ELF files export symbols for this event.
However, the more logical thing to do is avoid any warning and avoid parsing non-ELF files as the event is not meant for them. The fact is that it seems some high-level languages like Java can load non-ELF files. We might want to have some event on such cases though.

```
(paste your output here)
```

## Output of `uname -a`:

```
(paste your output here)
```

AlonZivony:

The symbols_loaded event currently log all errors received in its logic.
However, there is a case that an object is loaded to a process which is not ELF (for example, jsa files of Java).
As the event is not meant to parse such files, avoid logging errors in parsing them.

Fix #4033

### DIFF
--- a/pkg/events/derive/symbols_collision_test.go
+++ b/pkg/events/derive/symbols_collision_test.go
@@ -351,7 +351,7 @@ func TestSymbolsCollisionArgsGenerator_FindSOCollision(t *testing.T) {
 
 			for _, testCase := range testCases {
 				t.Run(testCase.name, func(t *testing.T) {
-					mockLoader := initLoaderMock(false)
+					mockLoader := initLoaderMock(nil)
 					gen := initSOCollisionsEventGenerator(
 						mockLoader,
 						filterTestCase.blackList,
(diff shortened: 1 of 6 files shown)


yanivagman (approved):
LGTM

PR #4040, opened 2024-05-10
chore(events): decrease SO loader error log level to debug

AlonZivony:

Errors from the SO loader are pretty frequent because of race conditions. As the user has nothing to do with it, the log level should be lowered from WARN.


yanivagman (approved):
LGTM

yanivagman (changes requested):
@AlonZivony tests are failing with these changes

yanivagman (approved):

PR #4053, opened 2024-05-16
feat(events): create global event id enum

### LINKED ISSUE #1098: Add a global event id enum
Today we expose event id as part of the new event structure (https://github.com/aquasecurity/tracee/blob/main/api/v1beta1/event.proto#L14)
This id, however, is not useful to the user that can't map it to a specific event.

Moreover, the existing event ids are architecture specific, meaning that our internal event ids are different for different architectures (more specifically, the event ids for syscalls are different).

To solve both of these problems, we need to define a global event id enum. For most of tracee's events, it will be just a simple mapping. However, tracee also support dynamic events, such as rego threat detection events. Such dynamic events can be assigned with an ID that is bigger than 10000 (meaning that all IDs below 10,000 are reserved to tracee built-in events).

rscampos:

fix #1098 

53d93eabb **fix(tests): event ID translation**
4806e8aa1 **feat(events): create global event id enum**

4806e8aa1 **feat(events): create global event id enum**

```
- create an array for translation internal ID to external ID;
- External syscall event starts at 1;
- External common event starts at 1000;
- External originated from user-space event start at 2000.
```


pkg/server/grpc/tracee.go
```diff
@@ -19,6 +19,456 @@ import (
 	"github.com/aquasecurity/tracee/types/trace"
 )
 
+// Syscall table map - internal syscall ID and external ID
+var syscallTranslationMap = map[events.ID]pb.Id{
```
rscampos, line 23: @geyslan and I were discuss about using a array instead of mapping here, to be more performatic... and for each arch have one array, WDYT @yanivagman

geyslan: Maybe one array would fit both archs range. The array index being the `syscall number` and its value being our `pb.Id`.

In this scenario, the array would only be filled at initialization.

api/v1beta1/event.proto
```diff
@@ -9,9 +9,597 @@ import "google/protobuf/wrappers.proto";
...
+    capture_net_packet = 4005;
+    capture_bpf = 4006;
+    capture_file_read = 4007;
+
+    // Signal meta-events
+    signal_cgroup_mkdir = 5000;
+    signal_cgroup_rmdir = 5001;
+    signal_sched_process_fork = 5002;
+    signal_sched_process_exec = 5003;
+    signal_sched_process_exit = 5004;
```
yanivagman, line 597: I think none of these should be exposed to the user since they are all internal to tracee

api/v1beta1/event.proto
```diff
@@ -9,9 +9,597 @@ import "google/protobuf/wrappers.proto";
...
+    io_pgetevents_time32 = 440;
+    recvmmsg_time32 = 441;
+    mq_timedsend_time32 = 442;
+    mq_timedreceive_time32 = 443;
+    rt_sigtimedwait_time32 = 444;
+    futex_time32 = 445;
+    sched_rr_get_interval_time32 = 446;
+
+    // Common events (used by all architectures).
+    net_packet_base = 700;
```
yanivagman, line 465: We previously tried to make event ids "packed" so 700 was chosen as the start of common events.
Since these values now become part of the API, I think it makes more sense to start the count from 1000.

api/v1beta1/event.proto
```diff
@@ -9,9 +9,597 @@ import "google/protobuf/wrappers.proto";
 import "api/v1beta1/event_data.proto";
 import "api/v1beta1/threat.proto";
 
+enum Id {
```
yanivagman, line 12: I think `Id` on its own may be confusing, maybe `EventId`?

pkg/server/grpc/tracee.go
yanivagman: good idea

pkg/server/grpc/tracee.go
```diff
@@ -138,10 +588,26 @@ func convertDefinitionToProto(d events.Definition) *pb.EventDefinition {
...
+		if idExternal, ok := syscallTranslationMap[events.ID(e.EventID)]; ok {
+			return pb.Id(idExternal)
+		}
+	} else { // If not syscall, access enum directly using event ID
+		// Return external ID just if found
+		if _, ok := pb.Id_name[int32(e.EventID)]; ok {
+			return pb.Id(e.EventID)
+		}
+	}
+	return pb.Id_unspecified
```
yanivagman, line 603: Initializing an array at start, as suggested in a previous comment, will make this translation a simple access to the array at EventID index, which is better considering the fact this will be called for every event

api/v1beta1/event.proto
rscampos: make sense, since `trace.Event` also uses `EventId`, tks

pkg/server/grpc/tracee.go
```diff
@@ -138,10 +706,19 @@ func convertDefinitionToProto(d events.Definition) *pb.EventDefinition {
 	}
 }
 
+func getGlobalID(e trace.Event) pb.EventId {
+	idExternal := EventTranslationMap[events.ID(e.EventID)]
+	if idExternal != 0 {
+		return pb.EventId(idExternal)
```
yanivagman, line 712: Do we need the casting? won't `return idExternal` be enough?

pkg/server/grpc/tracee.go
```diff
@@ -19,6 +19,574 @@ import (
 	"github.com/aquasecurity/tracee/types/trace"
 )
 
+// Event translation maps - internal to external Event ID
+var EventTranslationMap = [10200]pb.EventId{
```
yanivagman, line 23: Shouldn't the size of this array be 10,000? Any event id higher than 10000 should be considered non built-in event

pkg/server/grpc/tracee.go
```diff
@@ -138,10 +706,19 @@ func convertDefinitionToProto(d events.Definition) *pb.EventDefinition {
 	}
 }
 
+func getGlobalID(e trace.Event) pb.EventId {
+	idExternal := EventTranslationMap[events.ID(e.EventID)]
+	if idExternal != 0 {
+		return pb.EventId(idExternal)
+	}
+	return pb.EventId_unspecified
```
yanivagman, line 714: In case the mapping doesn't exist the event might be a non "built-in" event of trace (higher than 10,000), so I think we should return the event id as is (enums in proto 3 are considered "open" enums, so it should be possible to specify id's out of the enum range)

pkg/server/grpc/tracee.go
NDStrahilevitz: > Maybe one array would fit both archs range. The array index being the `syscall number` and its value being our `pb.Id`.
> 
> In this scenario, the array would only be filled at initialization - based on the arch.

We previously used build tags for this. So we can use the already existing code for that in `pkg/events/core/core_<arch>.go`, and refactor it into the array.

pkg/server/grpc/tracee.go
rscampos: You are right, think the size of this array should be 10,000, but today when we defined the following line, the value of `events.Stat` will depend on each arch we are compiling it.
```
events.Stat: pb.EventId_stat,
```
if amd64, `events.Stat` will be 4
if arm64, `events.Stat` will be 10,001 (because is [unsupported](https://github.com/aquasecurity/tracee/blob/baaf585ae06e9d19028739fc90f00e9f7859501e/pkg/events/core_arm64.go#L354-L357))

So I think the unsupported range would be below 10,000, e.g: 9,000. In this way will be perfect possible to define array size of 10,000.

pkg/server/grpc/tracee.go
rscampos: Agreed, so I think here we have 2 possible cases:
1. if below 10,000, check the translation table in order to get the external ID. If no external ID is found, we should return `pb.EventId_unspecified`.
2. If is higher than 10,000, we don't need to check the translation table and simple return id as is.

Have I doubt for 1: if event ID is not defined in the translation table, we always return `pb.EventId_unspecified` instead of id as id, right?

go.mod
```diff
@@ -45,6 +45,8 @@ require (
 	sigs.k8s.io/controller-runtime v0.16.3
 )
 
+replace github.com/aquasecurity/tracee/api => github.com/rscampos/tracee/api v0.0.0-20240520202449-d199291ce929
```
geyslan, line 48: don't forget to split this pr into 2, go.mod enters api first, main go.mod follows it.

pkg/server/grpc/tracee.go
```diff
@@ -19,6 +19,574 @@ import (
 	"github.com/aquasecurity/tracee/types/trace"
 )
 
+// EventTranslationTable - translate internal to external Event ID
```
geyslan, line 22: nit:

```suggestion
// EventTranslationTable translates internal to external protobuf Event Id
```

geyslan (approved):
LGTM

go.mod
rscampos: @geyslan splited the PR into 2, if you don't mind to do the last review, tks again :)

PR #4058, opened 2024-05-19
fix(proctree): limit tree memory usage and reduce it

AlonZivony:

This PR add limitation on the changelog objects sizes, and reduces drastically the memory usage of the tree by not saving the interpreter and interp information in the tree.
This should reduce the chance for memory leakage.
By rough calculation, before the PR the process tree would take around 150MB of memory on average.
After this change, it should be reduced by at least 60MB on average.

### DIFF
--- a/pkg/changelog/changelog.go
+++ b/pkg/changelog/changelog.go
@@ -1,6 +1,7 @@
 package changelog
 
 import (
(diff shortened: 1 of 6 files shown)


yanivagman (approved):
LGTM

PR #4059, opened 2024-05-21
fix(build): static build might require zstd lib

geyslan:
Close: https://github.com/aquasecurity/tracee/issues/4060

ac9bdb162 **chore: temp fix for install libzstd on AMIs**
656df9ac8 **chore: install required bpftool from btfhub**
e40109f69 **fix(build): static build might require zstd lib**

ac9bdb162 **chore: temp fix for install libzstd on AMIs**

```
AMIs have to be updated and installations like this commented out.
```

e40109f69 **fix(build): static build might require zstd lib**

```
libelf might be built with zstd support in some environments, so for
static builds (mainly) it is necessary to ensure that the zstd library
is available.

https://sourceware.org/git/?p=elfutils.git;a=commit;h=ed688a59b4d4f5ccf6ef15244e5a9139f71769a3
```

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -95,7 +95,7 @@ Vagrant.configure("2") do |config|
       ln -s "$path" "${path%-*}"
     done
 
-    apt-get install --yes zlib1g-dev libelf-dev
+    apt-get install --yes zlib1g-dev libelf-dev libzstd-dev
     apt-get install --yes protobuf-compiler
     apt-get install --yes linux-tools-"$(uname -r)" ||
       apt-get install --yes linux-tools-generic
--- a/tests/e2e-install-deps.sh
+++ b/tests/e2e-install-deps.sh
@@ -264,6 +264,22 @@ remove_golang_os_packages() {
     apt-get --purge autoremove -y
 }
 
+install_libzstd_os_packages() {
+    case $ID in
+    "ubuntu")
+        wait_for_apt_locks
+        apt-get install -y libzstd-dev
+    ;;
+    "almalinux")
+        yum install -y libzstd-devel
+    ;;
+    *)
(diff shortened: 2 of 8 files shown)


geyslan:
This would make room for https://github.com/aquasecurity/btfhub/pull/117.

geyslan:
https://github.com/aquasecurity/tracee/actions/runs/9166787615/job/25202867174?pr=4059

`e2e-install-deps.sh` should have (for now) the installation of zstd.

tests/e2e-install-deps.sh
```diff
@@ -264,6 +264,22 @@ remove_golang_os_packages() {
...
+    "almalinux")
+        yum install -y libzstd-devel
+    ;;
+    *)
+        echo "Unsupported OS: $ID"
+        exit 1
+    ;;
+    esac
+}
+
```
geyslan, line 282: All this would be unnecessary if we opt for an easier way of updating the AMIs in a bunch. Anyway, after its future upgrade, those steps should be commented/removed.

Vagrantfile
```diff
@@ -95,7 +95,7 @@ Vagrant.configure("2") do |config|
       ln -s "$path" "${path%-*}"
     done
 
-    apt-get install --yes zlib1g-dev libelf-dev
+    apt-get install --yes zlib1g-dev libelf-dev libzstd-dev
```
geyslan, line 98: As we discussed @rscampos, the setup of the bpftool in Vagrantfile must be done in a further PR (since it requires a system upgrade). Thanks for noticing that. 👍🏼

rscampos: I'll upgrade Vagrantfile in order use bpftool :)

rscampos (approved):
LGTM

PR #4069, opened 2024-05-21
moving printers to the new event structure

josedonizetti:
starting to work on  https://github.com/aquasecurity/tracee/issues/2870


josedonizetti:
@yanivagman  guessing this doesn't make sense anymore right? It was converting the old trace type into the new one for printers, so what people would see on tracee output would be the new format.

yanivagman:
> @yanivagman guessing this doesn't make sense anymore right? It was converting the old trace type into the new one for printers, so what people would see on tracee output would be the new format.

We are going to remove tracee-ebpf and tracee-rules soon. Then it will make total sense

PR #4076, opened 2024-05-26
fix: select cgroup mountpoint with the smallest inode number

yanivagman:

When running Tracee from within a container, multiple mount points of the cgroup filesystem may exist. In such cases, we need to ensure that we select the mountpoint from the host cgroup namespace. By checking for an inode equal to 1, we can identify the cgroupfs mountpoint belonging to the host cgroup namespace. This ensures that Tracee can discover preexisting containers using the host cgroup filesystem.

Additionally, a warning log is emitted if the cgroup mountpoint does not have an inode of 1, indicating that it might not be part of the host cgroup namespace.


denisgersh (approved):

denisgersh (approved):
LGTM.

PR #4078, opened 2024-05-26
fix(changelog): size enforcement corrupted cache

AlonZivony:

Fix bug with changelog max size enforcement that caused the remove of wrong entries from the timestamps cache, corrupting it.


yanivagman (approved):
LGTM

PR #4080, opened 2024-05-27
fix(proctree): fix clock type differences

### LINKED ISSUE #4074: Proctree hash calculation for procfs entries is wrong
## Description

We noticed that we get the following debug message when using the process tree:
```
{"L":"DEBUG","T":"2024-05-22T21:43:46.634+0300","M":"error enriching event executable info","pid":126,"tid":2248847,"event name":"setresuid","timestamp":36462467638300,"error":"not found in process tree","origin":"ebpf:pkg/ebpf/processor_proctree.go:260","calls":"(*Tracee).procTreeAddBinInfo() < (*Tracee).processEvent()"}
```
After some research, we found out that the hash received from the `ClockTicksToNsSinceBootTime` function in the process tree procfs operations return a wrong start time of the process/thread.
This results a wrong hashing of the `EntitiyId`.
The difference is in the format of the times received.
The kernel times we use are `MONOTONIC`, while the time received in the procfs is `BOOTTIME`. They defer from each other by the fact that monotonic does not include the sleep times. Hence, botttime is the more accurate format.

### Suggested Solution
We should probably use only boottime if it is possible, as it is more accurate.
The only thing preventing us from doing so is the fact that the events timestamps use the `bpf_ktime_get_ns` helper which return monotonic time.
A new helper, `bpf_ktime_get_boot_ns`, was introduced in kernel 5.8 that use the boottime.
So, we should transform all times used in tracee from monotonic to boottime for kernels 5.8 and above.
To do so, we need to use the `boottime_start` field of the `task_struct` for all tasks start times we use.
We will also need to check that the epoch time transform to the time will work with the boottime (or does it need some modification).
We will need to cancel the procfs features of the proctree for kernels older than 5.8 (can be done using the kallsyms).

## Credits
The issue was found and the solutions were discussed with @OriGlassman and @yanivagman.

AlonZivony:

Tracee had an issue that all the times from the kernel were monotonic. However, many user-mode times like times in procfs are of boottime. As boottime is more accurate, change to use boottime if possible (mainly newer kernels).

Fix #4074

### DIFF
--- a/pkg/ebpf/c/common/common.h
+++ b/pkg/ebpf/c/common/common.h
@@ -1,15 +1,17 @@
+#pragma once
+
 #ifndef __COMMON_COMMON_H__
-#define __COMMON_COMMON_H__
+    #define __COMMON_COMMON_H__
 
-#include <vmlinux.h>
+    #include <vmlinux.h>
 
-#include <bpf/bpf_core_read.h>
+    #include <bpf/bpf_core_read.h>
 
-#include <maps.h>
+    #include <maps.h>
 
 // PROTOTYPES
 
-#define statfunc static __always_inline
+    #define statfunc static __always_inline
 
 // FUNCTIONS & MACROS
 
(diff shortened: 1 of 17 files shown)


yanivagman:
Closing in favor of #4117

PR #4086, opened 2024-05-29
feat: remove default usage of parse-arguments

### LINKED ISSUE #2177: Remove need for parse-arguments

NDStrahilevitz:

The PR removes the default usage of parse-arguments. Data parsing will now be the default behavior only in the context of rego signatures. Go signatures have all been refactored to use raw data arguments. This should increase userspace pipeline performance in most use cases, and as such reduce the risk margin of event loss.

9c4cd8d5a **feat(rego): parse event arguments in sig**
bfd96e01b **feat(cmd): disable arg parsing by default**
a007d2d2c **feat(sigs): refactor to use nonparsed arguments**
6f0c1c9b0 **chore(go.mod): update sig helpers**
9c4cd8d5a **feat(rego): parse event arguments in sig**
```
Removing default argument parsing will break rego signatures, whose
implementation depends on parsed arguments. In order to keep the option
of readability in REGO signatures, events will be parsed by default in
the context of evaluating these kind of signatures.
This will also ensure that they will not be broken depending on the
selection of parse-arguments.
```
9d45a6522 **feat(helpers): robust int and uint arg helpers**
```
Add a new uint argument helpers, which extracts a data argument from an
event by name and returns it if it was one of the uint types.
Additionally, add further checks for int64 and int types in the already
existing int argument helper.
```

e14baa074 **feat(helpers): unparsed flag helpers**

```
```

### 3. Other comments
Resolve #2177

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -180,3 +180,5 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	kernel.org/pub/linux/libs/security/libcap/psx v1.2.68 // indirect
 )
+
+replace github.com/aquasecurity/tracee/signatures/helpers => ./signatures/helpers
--- a/go.sum
+++ b/go.sum
@@ -27,8 +27,6 @@ github.com/aquasecurity/libbpfgo v0.7.0-libbpf-1.4 h1:rQ94U12Xlz2tncE8Rxnw3vpp/9
 github.com/aquasecurity/libbpfgo v0.7.0-libbpf-1.4/go.mod h1:iI7QCIZ3kXG0MR+FHsDZck6cYs1y1HyZP3sMObBg0sk=
 github.com/aquasecurity/libbpfgo/helpers v0.4.6-0.20240313150344-0080df4914d9 h1:BMnpLW1ouA1xjEIQUVx6g/orqgVkTixdfRIJk+vYlFE=
 github.com/aquasecurity/libbpfgo/helpers v0.4.6-0.20240313150344-0080df4914d9/go.mod h1:JIFvPdLaEVJ6cS2lt52lcyUia9LkpxtTVijOaYesi5Q=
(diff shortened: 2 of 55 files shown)


NDStrahilevitz:
So this breaks compilation of tracee-rule. Maybe we can merge this after we remove those binaries.

NDStrahilevitz:
So this breaks compilation of tracee-rule. Maybe we can merge this after we remove those binaries.

go.mod
```diff
@@ -180,3 +180,5 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	kernel.org/pub/linux/libs/security/libcap/psx v1.2.68 // indirect
 )
+
+replace github.com/aquasecurity/tracee/signatures/helpers => ./signatures/helpers
```
NDStrahilevitz, line 184: TODO: split PR and remove this

geyslan:
Closing in favour of https://github.com/aquasecurity/tracee/pull/4086.

yanivagman:
> Closing in favour of #4086.

This is #4086 - was it closed mistakenly?

geyslan:
> > Closing in favour of #4086.
> 
> This is #4086 - was it closed mistakenly?

That comment was wrong actually. Fixed it.

PR #4092, opened 2024-05-30
Fix ksymbols mem consumption

yanivagman:

    ksymbols: reintroduce lazy symbol queries
    
    Add new required symbol and addresses lists to the KSymbolTable helper.
    When either list is given, the data structure will only save required
    symbols during scanning. This should reduce memory usage for users
    who can mostly guarentee that symbols are known ahead of time.
    
    Dynamic querying is still allowed in this mode, though it will incur
    a penalty as rescanning is required.

------------------------------------------------------------

    feat(ksymbols): restore lazy ksyms implementtion
    
    ahead of runtime.


geyslan:
Thank Yaniv. The last one is https://github.com/aquasecurity/tracee/pull/4093. I'm closing this.

PR #4093, opened 2024-05-31
Fix ksymbols mem consumption

geyslan:

### DIFF
--- a/pkg/ebpf/probes/probe_group.go
+++ b/pkg/ebpf/probes/probe_group.go
@@ -7,14 +7,11 @@ import (
 
 	"github.com/aquasecurity/tracee/pkg/errfmt"
 	"github.com/aquasecurity/tracee/pkg/logger"
-	"github.com/aquasecurity/tracee/pkg/utils/environment"
 )
 
 //
 // ProbeGroup
 //
-
-var kernelSymbolTable *environment.KernelSymbolTable
 
 // ProbeGroup is a collection of probes.
 type ProbeGroup struct {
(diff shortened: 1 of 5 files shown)


geyslan:
@NDStrahilevitz @yanivagman I fixed a null deref (unordered use of the `t.kernelSymbols`) but it continued failing on the HOOKED_SYSCALL test.

The approach of `onlyRequired` has problems with events which have non deterministic syms requirements - hooked_syscall in this case. I've just commented `WithRequiredSymbols` out, in the initialization, to make the tests pass.

NDStrahilevitz:
> The approach of `onlyRequired` has problems with events which have non deterministic syms requirements - hooked_syscall in this case. I've just commented `WithRequiredSymbols` out, in the initialization, to make the tests pass.
> 

That sounds like a bug, when I coded the new approach I wanted it to also support dynamic(non-deterministic) queries during runtime. The idea was that when a symbol that wasn't previously required was queried, the structure would call refresh and append the newly requested symbol/address to the required list.
> --- EDIT
> 
> A possible solution is to check if an event like the aforementioned was chosen and disable the WithRequiredSymbols option. But be aware that soon we'll have runtime changes so the ksyms should have an API for easy options set (even after initialized).

So then we would go back to square 1 and have no memory reduction (CNDR uses hooked_symbols so especially there). Let's try and fix the bug beforehand.

yanivagman:
The problem with hooked_syscall event was related to missing dependency (CAP_SYSLOG).
Fixed here: https://github.com/aquasecurity/tracee/pull/4095/commits/54cc21a36f9db306ef8dafa939d9d5f882c61ed9

yanivagman:
Closing in favor of #4095

PR #4095, opened 2024-06-02
Fix ksymbols mem consumption

yanivagman:

```
fix: add missing dependency for hooked_syscalls event
The hooked_syscalls event requires CAP_SYSLOG in order to
refresh its symbol table when reading /proc/kallsyms.
Add this missing dependency.
```

```
feat(ksymbols): restore lazy ksyms implementation

ahead of runtime.
```

### DIFF
--- a/pkg/ebpf/probes/trace.go
+++ b/pkg/ebpf/probes/trace.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/aquasecurity/tracee/pkg/errfmt"
 	"github.com/aquasecurity/tracee/pkg/logger"
+	"github.com/aquasecurity/tracee/pkg/utils/environment"
 )
 
 // NOTE: thread-safety guaranteed by the ProbeGroup big lock.
@@ -22,7 +23,23 @@ const (
 	KretProbe            // github.com/iovisor/bcc/blob/master/docs/reference_guide.md#1-kp
 	Tracepoint           // github.com/iovisor/bcc/blob/master/docs/reference_guide.md#3-tracep
 	RawTracepoint        // github.com/iovisor/bcc/blob/master/docs/reference_guide.md#7-raw-tracep
+	InvalidProbeType
 )
+
+func (t ProbeType) String() string {
+	switch t {
+	case KProbe:
+		return "kprobe"
+	case KretProbe:
+		return "kretprobe"
+	case Tracepoint:
+		return "tracepoint"
+	case RawTracepoint:
+		return "raw_tracepoint"
+	}
+
+	return "invalid"
+}
 
 // When attaching a traceProbe, by handle, to its eBPF program:
 //
@@ -94,7 +111,20 @@ func (p *TraceProbe) attach(module *bpf.Module, args ...interface{}) error {
 		// kernels it won't fail when trying to attach to a symbol that has
 		// multiple addresses.
 		//
-		syms, err := kernelSymbolTable.GetSymbolByName(p.eventName)
+
+		var ksyms *environment.KernelSymbolTable
+
+		for _, arg := range args {
+			switch a := arg.(type) {
+			case *environment.KernelSymbolTable:
+				ksyms = a
+			}
+		}
+		if ksyms == nil {
+			return errfmt.Errorf("trace probes needs kernel symbols table argument")
+		}
+
+		syms, err := ksyms.GetSymbolByName(p.eventName)
 		if err != nil {
 			goto rollback
 		}
--- a/pkg/ebpf/tracee.go
+++ b/pkg/ebpf/tracee.go
@@ -121,6 +121,11 @@ type Tracee struct {
 	policyManager *policyManager
 	// The dependencies of events used by Tracee
 	eventsDependencies *dependencies.Manager
+
+	// Ksymbols needed to be kept alive in table.
+	// This does not mean they are required for tracee to function.
+	// TODO: remove this in favor of dependency manager nodes
+	requiredKsyms []string
 }
 
 func (t *Tracee) Stats() *metrics.Stats {
@@ -238,6 +243,7 @@ func New(cfg config.Config) (*Tracee, error) {
 			func(id events.ID) events.Dependencies {
 				return events.Core.GetDefinitionByID(id).GetDependencies()
 			}),
+		requiredKsyms: []string{},
 	}
 
 	t.eventsDependencies.SubscribeAdd(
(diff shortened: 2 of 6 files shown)


pkg/ebpf/probes/probes.go
```diff
@@ -148,4 +148,5 @@ const (
 	ExecuteAtFinishedARM
 	ExecuteFinishedCompatARM
 	ExecuteAtFinishedCompatARM
+	NoProbe = 10000
```
NDStrahilevitz, line 151: This is a leftover, we can remove.

pkg/ebpf/probes/trace.go
```diff
@@ -22,8 +23,26 @@ const (
...
+	switch t {
+	case KProbe:
+		return "kprobe"
+	case KretProbe:
+		return "kretprobe"
+	case Tracepoint:
+		return "tracepoint"
+	case RawTracepoint:
+		return "raw_tracepoint"
+	case InvalidProbeType:
```
NDStrahilevitz, line 39: nit: remove and add as a return outside the switch

pkg/ebpf/probes/trace.go
```diff
@@ -77,6 +96,19 @@ func (p *TraceProbe) attach(module *bpf.Module, args ...interface{}) error {
 		return errfmt.WrapError(err)
 	}
 
+	var ksyms *environment.KernelSymbolTable
```
NDStrahilevitz, line 99: I think we may want to move this code section inside the switch case in the kprobe case. Otherwise, other traceprobes which don't need this will require it for attachment.

pkg/ebpf/tracee.go
```diff
@@ -862,6 +890,104 @@ func (t *Tracee) newConfig(cfg *policy.PoliciesConfig, version uint16) *Config {
 	}
 }
 
+func (t *Tracee) initKsymTableRequiredSyms() error {
```
NDStrahilevitz, line 893: We need to make sure we always initialize the text segment in relevant events. Perhaps we may want it as a symbol always present. WDYT?

pkg/utils/environment/kernel_symbols.go
```diff
@@ -22,110 +22,150 @@ type KernelSymbol struct {
...
 	owner string
 }
-
 type addrAndOwner struct {
 	addr  uint64
 	owner string
 }
 
+// KernelSymbolTable manages kernel symbols with multiple maps for fast lookup.
 type KernelSymbolTable struct {
```
NDStrahilevitz, line 35: I think we may want to add the option to batch require and query symbols. This would be useful in cases like `hooked_symbols` where we may receive multiple new symbols to query at once. The current API would have to make much more rescans than is necessary, since we can batch request all the new symbols at once, and only then trigger the scan.

NDStrahilevitz (dismissed):
Note the last comment in particular, that is an oversight I made in the original implementation.

geyslan:
Tested and the args from `hooked_syscall` are being populated correctly.

Related to the memory footprint this clearly brings a smaller initial mem consumption (RSS): 249496 (PR) < 456036 (main)

### This PR

#### Before the hook

```
ps -p `pgrep tracee` -o pid,vsz,rss,cmd
    PID    VSZ   RSS CMD
 412217 4201868 249496 ./dist/tracee -e hooked_syscall
```

#### After the hook

```
[hb0 tracee]# ps -p `pgrep tracee` -o pid,vsz,rss,cmd
    PID    VSZ   RSS CMD
 413535 3768752 166612 ./dist/tracee -e openat
```

It freed mem.


### main branch

#### Before the hook

```
[hb0 tracee]# ps -p `pgrep tracee` -o pid,vsz,rss,cmd
    PID    VSZ   RSS CMD
 423066 4918396 456036 ./dist/tracee -e hooked_syscall
```

#### After the hook

the same

geyslan:
@NDStrahilevitz I addressed some of your comments. The remainder is left for further consideration.

NDStrahilevitz:
> @NDStrahilevitz I addressed some of your comments. The remainder is left for further consideration.

My last comment seems rather critical to me, @yanivagman WDYT? I fear that if we don't address it we will introduce a performance regression while a `hooked_symbol` event is triggered.

pkg/ebpf/tracee.go
yanivagman: Shouldn't this be handled by event dependencies?

pkg/utils/environment/kernel_symbols.go
yanivagman: This can be a future optimization in a different PR.
For now, it is critical to resolve the memory footprint issue

pkg/ebpf/tracee.go
NDStrahilevitz: It should, that's why the first part was about double-checking that.

pkg/utils/environment/kernel_symbols.go
NDStrahilevitz: I was likely mistaken about having a relevant case for this right now in the codebase. Though it could be a possibility in the future, so it would still be good to add this API (in another PR), so that we don't accidentally introduce this regression in the future.

Anyway, resolving this one.

NDStrahilevitz (approved):
LGTM once e2e fixes are resolved.

geyslan:
@yanivagman @NDStrahilevitz ignore the wip (3cacd038cf03ea9d18c0ea4a6a6319d4f0795d29).

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -4300,12 +4300,12 @@ int tracepoint__module__module_load(struct bpf_raw_tracepoint_args *ctx)
     struct module *mod = (struct module *) ctx->args[0];
 
     if (event_is_selected(HIDDEN_KERNEL_MODULE_SEEKER, p.event->context.policies_version)) {
-        u64 insert_time = bpf_ktime_get_ns();
+        u64 insert_time = p.event->context.ts;
         kernel_new_mod_t new_mod = {.insert_time = insert_time};
         u64 mod_addr = (u64) mod;
         // new_module_map - must be after the module is added to modules list,
         // otherwise there's a risk for race condition
-        bpf_map_update_elem(&new_module_map, &mod_addr, &new_mod, BPF_ANY);
```
NDStrahilevitz, line 4308: @OriGlassman Proposed this internally afaik, should we keep this?

PR #4097, opened 2024-06-02
Fix ksymbols mem consumption 0.21

### LINKED ISSUE #4098: Increased memory usage compared to v0.19.0
## Description

After merging Pull Request #3802, we have observed a significant increase in the memory usage of Tracee. The memory consumption has risen by more than 100MB compared to the previous state. 

```
v0.20.0
```

## Output of `uname -a`:

```
(paste your output here)
```

yanivagman:

    ahead of runtime.

   Also fix a missing dependency (CAP_SYSLOG) of hooked_syscall event

### 3. Other comments

Fix #4098


yanivagman:
Wrong branch

PR #4105, opened 2024-06-04
fix(changelog): fix bugs with indexes

### LINKED ISSUE #4103: Proctree return wrong thread name for sched_process_exec
## Description

I noticed in the e2e tests that if I check the task name it fails.
It seems to return the name before the execution, even though all other information matches the after execution.

AlonZivony:

Fix multiple bugs in the pursuit to fix the issue that thread name received from proctree was incorrect.

Fix #4103

### 2. Explain how to test it

Updated all tests to the new use-cases


yanivagman (approved):
LGTM

Issue #4111, opened 2024-06-10
Remove dependencies of `pkg/utils/environment`


geyslan:
https://github.com/aquasecurity/tracee/pull/4096#discussion_r1623634047

Remove internal dependencies of `pkg/utils/environment` getting rid of possible cyclic imports.

_Originally posted by @yanivagman in https://github.com/aquasecurity/tracee/pull/4096#discussion_r1632352099_

PR #4113, opened 2024-06-10
New modules `logger`, `errfmt` and `environment`

geyslan:
Part of #4111 solution.

1122db1f1 **chore: use environment module**
6ee5a54cf **chore: mv pkg/utils/environment into environment mod**
4e7c7deaf **chore: use errfmt module**
efae8e68e **chore: move pkg/errfmt into errfmt module**
0e22c48dd **chore: use logger module**
c21142205 **chore: move pkg/logger into logger module**

### 3. Other comments

Well, there's no straight way of doing this relocation. So, after this merge will have at least 2 more PRs to point to the right modules hashes.


rscampos (approved):
LGTM

geyslan:
@AlonZivony do you think this would help with https://github.com/aquasecurity/tracee/issues/4111 matter?

yanivagman:
I'm just thinking about a couple of things:
1. Should we have so many modules? Maybe we can converge some?
2. Should they be under the root of the tree? Can't we put them under pkg?
3. Do we really want errfmt as a module (it doesn't have much logic in it).
 
Maybe we should remove the error handling from within the environment package and let the caller handle any returned errors?

geyslan:
> A few thoughts:
> 
>     1. Should we have so many modules? Maybe we can converge some?

We should not have. I agree that converging them is the best approach to us. But we have to have in mind that splitting should avoid cyclic import issues.

> 
>     2. Should they be under the root of the tree? Can't we put them under pkg?

If modules, bring them up to avoid cyclic import issues.

> 
>     3. Do we really want errfmt as a module (it doesn't have much logic in it).

No, but it can be part of a converged module, like `utils` - I know, I know, utils is the worse naming option, but since we would converge different pkgs into it, they could be imported like `utils/errfmt`, `utils/environment`...

> 
> 
> Maybe we should remove the error handling from within the environment package and let the caller handle any returned errors?

Well, this is an option too.

geyslan:
We are choosing another approach. https://github.com/aquasecurity/tracee/pull/4125

PR #4117, opened 2024-06-12
fix(proctree): fix clock type differences

rscampos:

Tracee had an issue that all the times from the kernel were monotonic. However, many user-mode times like times in procfs are of boottime. As boottime is more accurate, change to use boottime if possible (mainly newer kernels).

If boottime is not available (old kernels), it's still possible to use the proctree option, but the proctree won't be built from procfs. Instead, it will be built from some events.

Fix https://github.com/aquasecurity/tracee/issues/4074

6be76aee5 **Enable time normalization for proctree**
9142b16ec **Enable time normalization for control plane**
0d4ea1022 **Check eBPF helper func symbol & Time normalization init**
aa68a61f5 **Time package**
51d562b5d **eBPF data plane - use boottime whenever possible**

6be76aee5 **Enable time normalization for proctree**

```
```

9142b16ec **Enable time normalization for control plane**

```
```

0d4ea1022 **Check eBPF helper func symbol & Time normalization init**

```
- Check if the eBPF helper function for getting boottime is available;
otherwise, use monotonic time. If boottime is available and proctree
is enabled: initialize proctree using procfs and querying.

- Initialize time normalization based on the option chosen by the user
(relative or absolute time).

```

aa68a61f5 **Time package**

```
- Time normalization interface can be implemented by relative or
  absolute
time;
- Create package time under pkg/time/;
- Update utils function for time.

```

51d562b5d **eBPF data plane - use boottime whenever possible**

```
- Replaced every instance of getting the current timestamp with the
function get_current_time_in_ns(). Internally, bpf_ktime_get_boot_ns()
(boottime) is used for newer kernels, while bpf_ktime_get_ns (monotonic)
is used for older kernels. Boottime is used whenever possible;

- Changed MAX_NUM_MODULES from 450 to 440.

```

### 2. Explain how to test it

Need to enable proctree:
`sudo ./dist/tracee --events kill --proctree source=both`

### 3. Other comments

Credits to @AlonZivony, who did most of the work.

### DIFF
--- a/pkg/ebpf/processor_funcs.go
+++ b/pkg/ebpf/processor_funcs.go
@@ -353,37 +353,10 @@ func (t *Tracee) normalizeEventCtxTimes(event *trace.Event) error {
 		// derived events are normalized from their base event, skip the processing
 		return nil
 	}
-
-	//
-	// Currently, the timestamp received from the bpf code is of the monotonic clock.
-	//
-	// TODO: The monotonic clock doesn't take into account system sleep time.
-	// Starting from kernel 5.7, we can get the timestamp relative to the system boot time
-	// instead which is preferable.
-
-	if t.config.Output.RelativeTime {
-		// monotonic time since tracee started: timestamp - tracee starttime
-		event.Timestamp = event.Timestamp - int(t.startTime)
-		event.ThreadStartTime = event.ThreadStartTime - int(t.startTime)
-	} else {
-		// current ("wall") time: add boot time to timestamp
-		event.Timestamp = event.Timestamp + int(t.bootTime)
-		event.ThreadStartTime = event.ThreadStartTime + int(t.bootTime)
-	}
+	event.Timestamp = t.timeNormalizer.NormalizeTime(event.Timestamp)
+	event.ThreadStartTime = t.timeNormalizer.NormalizeTime(event.ThreadStartTime)
 
 	return nil
-}
-
-// getOrigEvtTimestamp returns the original timestamp of the event.
-// To be used only when the event timestamp was normalized via normalizeEventCtxTimes.
-func (t *Tracee) getOrigEvtTimestamp(event *trace.Event) int {
-	if t.config.Output.RelativeTime {
-		// if the time was normalized relative to tracee start time, add the start time back
-		return event.Timestamp + int(t.startTime)
-	}
-
-	// if the time was normalized to "wall" time, subtract the boot time
-	return event.Timestamp - int(t.bootTime)
 }
 
 // processSchedProcessFork processes a sched_process_fork event by normalizing the start time.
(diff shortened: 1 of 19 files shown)


pkg/utils/time.go
```diff
@@ -76,13 +76,19 @@ func getBootTimeInJiffies() int64 {
 // Boot time functions
 //
 
+// Common clock IDs
+const (
+	CLOCK_MONOTONIC = unix.CLOCK_MONOTONIC // Time since a boot (not including time spent in suspend)
+	CLOCK_BOOTTIME  = unix.CLOCK_BOOTTIME  // Time since a boot (including time spent in suspend)
+)
+
 // GetStartTimeNS returns the system start time in nanoseconds (using CLOCK_MONOTONIC).
```
yanivagman, line 85: There are many comments in this file that still refer to CLOCK_MONOTONIC, like this one

pkg/ebpf/processor_funcs.go
```diff
@@ -353,37 +353,16 @@ func (t *Tracee) normalizeEventCtxTimes(event *trace.Event) error {
...
 func (t *Tracee) getOrigEvtTimestamp(event *trace.Event) int {
-	if t.config.Output.RelativeTime {
-		// if the time was normalized relative to tracee start time, add the start time back
-		return event.Timestamp + int(t.startTime)
-	}
-
-	// if the time was normalized to "wall" time, subtract the boot time
-	return event.Timestamp - int(t.bootTime)
+	return t.timeNormalizer.GetOriginalTime(event.Timestamp)
 }
```
yanivagman, line 366: We can remove this wrapper if it only returns a different function

pkg/utils/time.go
rscampos: good catch, missed this detail... now can be both, depending on kernel version

pkg/ebpf/normalizetime/time.go
```diff
@@ -0,0 +1,57 @@
...
+	GetOriginalTime(timeNs int) int
+}
+
+// CreateTimeNormalizerByConfig create a TimeNormalizer according to given configuration using
+// runtime functions.
+func CreateTimeNormalizerByConfig(relative bool, clockId int) TimeNormalizer {
+	if relative {
+		return NewRelativeTimeNormalizer(int(utils.GetStartTimeNS(int32(clockId))))
+	}
+	return NewAbsoluteTimeNormalizer(int(utils.GetBootTimeNS(int32(clockId))))
```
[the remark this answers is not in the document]
rscampos: Good catch, make more sense to use the same start timestamp to create the relative normalizer, to be consistent.

For making this change, its necessary to move the start timestamp measure up (today the measure is in the after the execution of Tracee Init). I don't think this is problem, right? (measuring the time before ending the Init)

pkg/ebpf/tracee.go
```diff
@@ -530,6 +524,52 @@ func (t *Tracee) Init(ctx gocontext.Context) error {
...
+					err = nil
+				}
+			}
+			return err
+		},
+		cap.SYSLOG,
+	)
+	if err != nil {
+		return errfmt.WrapError(err)
+	}
```
yanivagman, line 548: As discussed offline, this should be done using libbpf probing function

yanivagman (approved):
Overall LGTM,
Added a previously discussed comment that requires pushing the libbpfgo probing function first. We can do it in a later PR, or if you prefer, before merging this PR.

rscampos:
Thank you for the review @yanivagman , will merge and as soon as I finished the libbpfgo probe func, I'll do the refactor.

PR #4120, opened 2024-06-13
Refactor/analyze mode no proctree

### LINKED ISSUE #3520: Change analyze mode to use the normal pipeline

AlonZivony:

Fix #3520


pkg/events/usermode.go
```diff
@@ -49,6 +49,27 @@ func InitNamespacesEvent() trace.Event {
...
+	def := Core.GetDefinitionByID(InitTraceeData)
+	params := def.GetParams()
+	args := []trace.Argument{
+		{ArgMeta: params[0], Value: bootTime},
+		{ArgMeta: params[1], Value: startTime},
+	}
+
+	initTraceeDataEvent := trace.Event{
+		Timestamp:   int(time.Now().UnixNano()),
+		ProcessName: "tracee-ebpf",
```
yanivagman, line 63: let's rename to tracee, since tracee-ebpf cmd will be removed soon

pkg/ebpf/processor_proctree.go
```diff
@@ -98,39 +98,6 @@ func (t *Tracee) procTreeForkProcessor(event *trace.Event) error {
...
 }
 
-// procTreeForkRemoveArgs removes arguments needed for the process tree only (when source is events).
-func (t *Tracee) procTreeForkRemoveArgs(event *trace.Event) error {
-	argsToRemove := []string{
-		"up_parent_tid",
-		"up_parent_ns_tid",
-		"up_parent_pid",
-		"up_parent_ns_pid",
-		"up_parent_start_time",
```
yanivagman, line 108: If we make these fields permanent, we should rename them to something more clear.
I suggest calling it `parent_process_xxx` instead of `up_parent_xxx`

yanivagman (changes requested):
I only reviewed the first two commits since I think they can be extracted into their own PRs.
Let's merge them first and then continue the review with a rebased version

pkg/ebpf/processor_proctree.go
rscampos: Pls ref to https://github.com/aquasecurity/tracee/pull/4167#issuecomment-2204087350

yanivagman:
Closing this PR as some of the commits here were already merged while others will be done differently.

PR #4122, opened 2024-06-13
Added event containing full payload for all packets

### LINKED ISSUE #4121: Add event containing full packet payloads
Some use cases require access to the full payload of network packets in the event stream, instead of needing to analyze capture files written separately.

oshaked1:

Add event containing the full payload of all network packets.

Closes #4121 

### 2. Explain how to test it

Run with the `net_packet_raw` event enabled.


yanivagman (approved):
LGTM

Issue #3987, opened 2024-04-18
Remove policy limitations


yanivagman:
Today Tracee policies have two limitations that we need to remove:
1. Only up to 64 policies are supported
2. Only 1 rule with the same event can be specified in a policy

Removing the second limitation is probably more important since it will allow a user to write rules that specify different sets of filters to the same event, thus supporting logical OR operation between the filter sets.

PR #4125, opened 2024-06-14
wip ...

geyslan:
Close: #3987


geyslan:
This is the package with the fewest changes and it seems impractical to continue. Therefore, the necessary small steps will be addressed first, even if intermediate and removed in the future. One of them is the pseudo-encapsulation of the eventsStates into Policies.

PR #4126, opened 2024-06-16
perf: Optmize event parameter decoding in pipeline

yanivagman:

Refactor the event decoding stage in the Tracee pipeline to fetch event parameters and related data (e.g., event name) from the event definition only once. Previously, the `GetParams` function was called multiple times within `DecodeArguments`, leading to unnecessary overhead.

This optimization improves the efficiency of event processing, particularly for events with a large number of parameters or frequent occurrences.

### DIFF
--- a/pkg/ebpf/controlplane/signal.go
+++ b/pkg/ebpf/controlplane/signal.go
@@ -30,8 +30,10 @@ func (sig *signal) Unmarshal(buffer []byte) error {
 		return errfmt.Errorf("failed to get event %d configuration", sig.id)
 	}
 	eventDefinition := events.Core.GetDefinitionByID(sig.id)
-	sig.args = make([]trace.Argument, len(eventDefinition.GetParams()))
-	err = ebpfDecoder.DecodeArguments(sig.args, int(argnum), eventDefinition, sig.id)
+	evtParams := eventDefinition.GetParams()
+	evtName := eventDefinition.GetName()
+	sig.args = make([]trace.Argument, len(evtParams))
+	err = ebpfDecoder.DecodeArguments(sig.args, int(argnum), evtParams, evtName, sig.id)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/ebpf/controlplane/signal.go
```diff
@@ -30,8 +30,10 @@ func (sig *signal) Unmarshal(buffer []byte) error {
 		return errfmt.Errorf("failed to get event %d configuration", sig.id)
 	}
 	eventDefinition := events.Core.GetDefinitionByID(sig.id)
-	sig.args = make([]trace.Argument, len(eventDefinition.GetParams()))
-	err = ebpfDecoder.DecodeArguments(sig.args, int(argnum), eventDefinition, sig.id)
+	evtParams := eventDefinition.GetParams()
+	evtName := eventDefinition.GetName()
+	sig.args = make([]trace.Argument, len(evtParams))
```
geyslan, line 35: Would we reduce initialization overhead setting len to 0?

```suggestion
	sig.args = make([]trace.Argument, 0, len(evtParams))
```

pkg/ebpf/events_pipeline.go
```diff
@@ -180,8 +180,10 @@ func (t *Tracee) decodeEvents(ctx context.Context, sourceChan chan []byte) (<-ch
 				continue
 			}
 			eventDefinition := events.Core.GetDefinitionByID(eventId)
-			args := make([]trace.Argument, len(eventDefinition.GetParams()))
-			err := ebpfMsgDecoder.DecodeArguments(args, int(argnum), eventDefinition, eventId)
+			evtParams := eventDefinition.GetParams()
+			evtName := eventDefinition.GetName()
+			args := make([]trace.Argument, len(evtParams))
```
geyslan, line 185: ditto?

geyslan (approved):
LGTM

PR #4129, opened 2024-06-17
chore: rem logger and errfmt as deps from env pkg

geyslan:
Close: #4111

1d8278c95 **chore: rem logger and errfmt as dependencies**

```
Remove logger and errfmt as dependencies from the environment package.
```

### 3. Other comments


yanivagman (approved):
LGTM

PR #4130, opened 2024-06-17
chore: possible fix for size/RSS issue

rscampos:

a1ad19fb3 **chore: Possible fix for size/RSS issue**

```
Using scheme package increases the binary size because it includes
additional code and dependencies related to Kubernetes. It also
increases the RSS memory.

Because we don't need the entire scheme package, only
apimachinery/runtime package was used to create a new scheme.
```


yanivagman (approved):
LGTM

PR #4132, opened 2024-06-19
[v0.21.0] chore: possible fix for size/RSS issue

rscampos:

c943766f8 **chore: Possible fix for size/RSS issue**

```
Using scheme package increases the binary size because it includes
additional code and dependencies related to Kubernetes. It also
increases the RSS memory.

Because we don't need the entire scheme package, only
apimachinery/runtime package was used to create a new scheme.

commit: 3b5175d (main), cherry-pick
```


geyslan (approved):
LGTM

PR #4137, opened 2024-06-20
fix(ci): dev tag is the latest snapshot

geyslan:


yanivagman:
@geyslan is that change correct?
Isn't dev tag out of main and not the latest release branch?

geyslan:
Continuation: https://github.com/aquasecurity/tracee/pull/4137

PR #4141, opened 2024-06-21
fix(ci): make release rule to have prerequisites

geyslan:

82336fef5 **fix(ci): make release rule to have prerequisites**

```
These changes ensure that the 'release' rule will require the other
rules to be executed (splitting the commands into different rules). This
way, if any some of the rules fail, the build will be aborted right away.
```


rscampos (approved):
LGTM

Issue #4156, opened 2024-06-26
Integration Tests failing


geyslan:
## Description

It was detected problems with `Test_EventsDependencies` and `Test_TraceeCapture` so far.

`sudo make test-integration`

### Manjaro

```
=== RUN   Test_TraceeCapture
=== RUN   Test_TraceeCapture/capture_write/read
    capture_test.go:85:   --- started tracee ---
    event_filters_test.go:1869: Cooling down for 5s
=== RUN   Test_TraceeCapture/capture_write/readv
    event_filters_test.go:1869: Cooling down for 2s
    capture_test.go:85:   --- started tracee ---
    event_filters_test.go:1869: Cooling down for 5s
=== RUN   Test_TraceeCapture/capture_pipe_write/read
    event_filters_test.go:1869: Cooling down for 2s
    capture_test.go:85:   --- started tracee ---
    event_filters_test.go:1869: Cooling down for 5s
    capture_test.go:329: assertion failed: 
        --- input
        +++ →
        @@ -1,2 +1,8 @@
        +Hello World!
        +Hello World!
        +Hello World!
        +Hello World!
        +Hello World!
        +Hello World!
         Hello World!
...

=== NAME  Test_TraceeCapture
    capture_test.go:106: found unexpected goroutines:
        [Goroutine 1278 in state syscall, with syscall.Syscall6 on top of the stack:
        syscall.Syscall6(0xf7, 0x1, 0x4f98e, 0xc001046d90, 0x1000004, 0x0, 0x0)
                /root/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.4.linux-amd64/src/syscall/syscall_linux.go:91 +0x45
        os.(*Process).blockUntilWaitable(0xc001cc24e0)
                /root/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.4.linux-amd64/src/os/wait_waitid.go:32 +0x9f
        os.(*Process).wait(0xc001cc24e0)
                /root/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.4.linux-amd64/src/os/exec_unix.go:22 +0x4b
        os.(*Process).Wait(...)
                /root/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.4.linux-amd64/src/os/exec.go:134
        os/exec.(*Cmd).Wait(0xc001cd26e0)
[...]

PR #4157, opened 2024-06-26
Fix Integration Tests

geyslan:
Close: #4156 

74bfd7d41 **fix(tests): refactor Test_TraceeCapture**
4e1f89918 **fix(tests): Test_EventsDependencies**
f493e2890 **chore: turn logger thread-safe**

74bfd7d41 **fix(tests): refactor Test_TraceeCapture**

```
This test was failing when other tests were run before it and failed,
tracee running was not stopped and some garbage was left in the system.

Two major changes were made:

1. Do not fail the test before stopping tracee.
2. Always remove capture files after the test (do cleanup).
```

4e1f89918 **fix(tests): Test_EventsDependencies**

```
Test_EventsDependencies overrides the default logger to capture the logs
and check them as part of the test.

Before it was synced to pass in the github actions, what wasn't
happening locally or in a different environment.

Since the logger was not thread safe, it was causing the test to fail
(data race). That forced the previous commit which turned Logger into a
thread safe logger.

This fixes the problem refactoring it to override and restore the logger
in the correct steps.
```

### DIFF
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"go.uber.org/zap"
@@ -157,6 +158,9 @@ func aggregateLog(l *Logger, lvl Level, msg string, ci *callerInfo) bool {
 // It does NOT override those options.
 // For the case where innerAggregation is set to true, it will only aggregate if pkg logger's aggregation config is also set to true.
 func Log(lvl Level, innerAggregation bool, msg string, keysAndValues ...interface{}) {
+	pkgMutex.RLock()
+	defer pkgMutex.RUnlock()
+
 	if innerAggregation {
 		switch lvl {
 		case DebugLevel:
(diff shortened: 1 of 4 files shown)


pkg/logger/logger.go
```diff
@@ -341,18 +366,27 @@ func SetLogger(l LoggerInterface) {
 
 // GetLogger gets the package-level base logger
 func GetLogger() LoggerInterface {
+	pkgMutex.RLock()
+	defer pkgMutex.RUnlock()
```
yanivagman, line 370: Ditto

pkg/logger/logger.go
```diff
@@ -322,16 +341,22 @@ func (l *Logger) Sync() error {
 var (
 	// Package-level Logger
 	pkgLogger = &Logger{}
+	pkgMutex  = &sync.RWMutex{}
 )
 
 // Current returns the package-level base logger
 func Current() *Logger {
+	pkgMutex.RLock()
+	defer pkgMutex.RUnlock()
```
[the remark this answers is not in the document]
geyslan: We need to, since any chunk of memory accessed without synchronicity is subject to data race. I had the same doubt, because in this case it is just a pointer copy, but it is not guaranteed to be an atomic read.

pkg/logger/logger.go
geyslan: ditto.

yanivagman (approved):
LGTM

PR #4160, opened 2024-06-28
chore(logger): safe guard before locking

geyslan:

baa97d3fa **chore(logger): safe guard before locking**


rscampos (approved):
LGTM

PR #4161, opened 2024-06-28
chore: make dependencies manager a singleton

geyslan:

83c39cf83 **chore: mv dependencies tests to same pkg**
5b6c9fce4 **chore: make dependencies manager a singleton (temp)**

5b6c9fce4 **chore: make dependencies manager a singleton (temp)**

```
This tidies up the dependencies manager by making it a singleton,
temporarily. It's a necessary step to decouple the manager from the
Tracee object without touching much code, allowing it to be accessed
from other parts of the codebase and be detached from other logic.

It introduces an also temporary ResetManagerFromTests() to reset the
manager state between tests.

When elements are decoupled and redesign, the manager will not be a
singleton anymore, but it will be accessed via dependency injection.

NOTE: It continues being not thread-safe. This will be addressed in a
future PR.
```

### DIFF
--- a/pkg/ebpf/ksymbols.go
+++ b/pkg/ebpf/ksymbols.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/aquasecurity/tracee/pkg/errfmt"
 	"github.com/aquasecurity/tracee/pkg/events"
+	"github.com/aquasecurity/tracee/pkg/events/dependencies"
 	"github.com/aquasecurity/tracee/pkg/logger"
 )
 
@@ -26,7 +27,7 @@ func (t *Tracee) UpdateKallsyms() error {
 
 	// Wrap long method names.
 	evtDefSymDeps := func(id events.ID) []events.KSymbol {
-		depsNode, _ := t.eventsDependencies.GetEvent(id)
+		depsNode, _ := dependencies.GetManagerInstance().GetEvent(id)
 		deps := depsNode.GetDependencies()
 		return deps.GetKSymbols()
 	}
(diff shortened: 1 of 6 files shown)


yanivagman:
Why do we want it to be a singleton?
In general, I'm against using singletons in tracee and prefer to define clear interfaces for each component and use dependency injection to provide references to the necessary components to each other. This maintains loose coupling and makes it easier to test and modify components in isolation.

geyslan:
DI is good as well but currently to decouple modules from Tracee we should inject those modules (changing function/method signatures) of all related to it. This demands a clear vision of the relation of the elements.

And it also only works if the injected is thread safe what dependencies manager isn't yet.

About the coupling of the singleton in tests... We can use methods to control its visibility only for the test build tag. That's a change that I was about to do to this PR, so the Reset function wouldn't be available to production, only in tests.

Anyhoo, if you do prefer I can step back, make dependencies manager thread-safe first and in futures PRs:

- initialize it and inject it to the policy manager creation, removing it from the Tracee object.
- decouple event states from Tracee making it private to the policy manager.

WDYT?

Last note: the main focus on making the transitional objects into singletons is to make the code logic easy in the tidying process, without touching much code. When the objects with relations are moved and tidied, the singleton pattern should be removed leaving the injection to do the job.

geyslan:
Correction:

- initialize it and inject it to the policy manager creation, ~removing it from~ leaving it to the Tracee object as well.

As it would be thread safe, all that demands access to it should get a copy from Tracee's.

yanivagman:
> Last note: the main focus on making the transitional objects into singletons is to make the code logic easy in the tidying process, without touching much code. When the objects with relations are moved and tidied, the singleton pattern should be removed leaving the injection to do the job.

If this step is only transitional and the goal is to make the logic easy before removing the singleton, I'm ok with it.
I'll complete the review with that comment in mind.

yanivagman (approved):
LGTM, considering https://github.com/aquasecurity/tracee/pull/4161#issuecomment-2199762849

PR #4163, opened 2024-06-28
move PolicyManager to policy pkg

geyslan:

209a44f9e **chore: move policy manager into policy pkg**

```
Place PolicyManager in its proper home, the policy package.
```


geyslan:
One more tidying.

geyslan:
https://github.com/aquasecurity/tracee/pull/4165 covers this.

PR #4165, opened 2024-07-01
Tidying Policy Manager

geyslan:

93aaaf98a **chore: tidy up policy manager and related code**
857d8ed16 **chore(policy): change manager mutex type to rw**
209a44f9e **chore: move policy manager into policy pkg**

9341d3510 **chore: tidy up policy manager and related code**

```
- Remove Snapshot() use. In the future it could be managed by the policy
  manager itself.
- After PolicyManager initialization (first stages), Policies are not
  accessed directly anymore, but through the PolicyManager.
- t.config.Policies is now only transient and used for PolicyManager
  initialization.
- Policy version is deprecated and to be removed soon.
```

209a44f9e **chore: move policy manager into policy pkg**

```
Place PolicyManager in its proper home, the policy package.
```

### DIFF
--- a/pkg/cmd/cobra/cobra.go
+++ b/pkg/cmd/cobra/cobra.go
@@ -231,7 +231,6 @@ func GetTraceeRunner(c *cobra.Command, version string) (cmd.Runner, error) {
 	}
 
 	cfg.Policies = policies
-	policy.Snapshots().Store(cfg.Policies)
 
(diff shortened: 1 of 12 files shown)


geyslan:
This encompasses https://github.com/aquasecurity/tracee/pull/4163.

yanivagman (approved):
LGTM

PR #4166, opened 2024-07-02
feat(events): create tracee_info event

rscampos:
Part of #3520

5c02ebf4e **feat(events): create tracee_info event**

```
Create an event that export Tracee's data upon startup.

```

### 3. Other comments

Credits to @AlonZivony, who did the work.

### DIFF
--- a/docs/docs/events/builtin/extra/tracee_info.md
+++ b/docs/docs/events/builtin/extra/tracee_info.md
@@ -1,0 +1,27 @@
+# tracee_info
+
+## Intro
+
+tracee_info - An event that exports some relevant data of Tracee upon startup.
+
+## Description
+
+This event, created in user-mode during Tracee's initialization, is typically the first event emitted. It provides valuable metadata about Tracee's configuration and runtime environment, which can be helpful for event processing and troubleshooting.
+
+The event was created also with Tracee's File Source in mind, to provide information about how Tracee ran during the original capture.
+
+## Arguments
+
+* `boot_time`:`u64`[U] - the boot time of the system that Tracee is running on, relative to the Unix epoch.
+* `start_time`:`u64`[U] - the time the Tracee process started relative to system boot time.
+* `version`:`const char*`[U] - Tracee version.
+
+## Hooks
+
+## Example Use Case
+
+The event could be used to calculate the relative time of events since Tracee's start.
+
+## Related Events
+
+`init_namespaces`
--- a/pkg/ebpf/tracee.go
+++ b/pkg/ebpf/tracee.go
@@ -1731,6 +1731,14 @@ func (t *Tracee) invokeInitEvents(out chan *trace.Event) {
 
 	// Initial namespace events
 
+	matchedPolicies = policiesMatch(t.eventsState[events.TraceeInfo])
+	if matchedPolicies > 0 {
+		traceeDataEvent := events.TraceeInfoEvent(t.bootTime, t.startTime)
+		setMatchedPolicies(&traceeDataEvent, matchedPolicies, t.policyManager)
+		out <- &traceeDataEvent
+		_ = t.stats.EventCount.Increment()
+	}
+
 	matchedPolicies = policiesMatch(t.eventsState[events.InitNamespaces])
 	if matchedPolicies > 0 {
 		systemInfoEvent := events.InitNamespacesEvent()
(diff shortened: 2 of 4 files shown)


yanivagman (approved):
LGTM

docs/docs/events/builtin/extra/tracee_info.md
```diff
@@ -0,0 +1,27 @@
...
+## Description
+
+This event, created in user-mode during Tracee's initialization, is typically the first event emitted. It provides valuable metadata about Tracee's configuration and runtime environment, which can be helpful for event processing and troubleshooting.
+
+The event was created also with Tracee's File Source in mind, to provide information about how Tracee ran during the original capture.
+
+## Arguments
+
+* `boot_time`:`u64`[U] - the boot time of the system Tracee run in since epoch.
+* `start_time`:`u64`[U] - the Tracee time elapse since system boot. This indicates when the Tracee process started relative to the system boot time.
```
yanivagman, line 16:
```suggestion
* `start_time`:`u64`[U] - the time elapsed since Tracee started, relative to the system boot time. This indicates when the Tracee process started relative to the system boot time.
```

docs/docs/events/builtin/extra/tracee_info.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Description
+
+This event, created in user-mode during Tracee's initialization, is typically the first event emitted. It provides valuable metadata about Tracee's configuration and runtime environment, which can be helpful for event processing and troubleshooting.
+
+The event was created also with Tracee's File Source in mind, to provide information about how Tracee ran during the original capture.
+
+## Arguments
+
+* `boot_time`:`u64`[U] - the boot time of the system Tracee run in since epoch.
```
yanivagman, line 15:
```suggestion
* `boot_time`:`u64`[U] - the boot time of the system that Tracee is running on, relative to the Unix epoch.
```

yanivagman:
> @yanivagman @rscampos Doesn't this fit the datastore framework much more? Static data which should be accessible over time shouldn't be exposed as a transient event. As this data fits under the `traceectl status` future command, shouldn't this be statically held and exposed under a gRPC method? And a signature could access it through the datastore API.

The first reason for creating this event is to be used by the file source (aka "analyze mode"). Since we only expose events in the output file, I don't think data store would fit for this purpose

NDStrahilevitz:
I think we should invest in a more robust way to convert datastores in general into sources for the analyze mode in this case, rather than adding directly as a pipeline event (although it could be a solution that datastores emit events).

An example for such a design could be that each datastore would implement a "commitlog" interface. So changes can only be done through commits, which the datastore needs to keep track of (if its designed to be used in analyze mode later - maybe some datastore authors in plugins will not have this intention). Then, we could set a flag that dumps the commit-log after tracee's run. Maybe we would have `CommitDataStore` which inherits `DataStore`, and only `CommitDataStore`s can work in analyze mode.

pkg/ebpf/tracee.go
```diff
@@ -1731,6 +1731,14 @@ func (t *Tracee) invokeInitEvents(out chan *trace.Event) {
 
 	// Initial namespace events
 
+	matchedPolicies = policiesMatch(t.eventsState[events.TraceeInfo])
+	if matchedPolicies > 0 {
```
geyslan, line 1735: I'm getting rid of this `eventState` direct access soon. JFIY. Lookups related to applied events will be done via t.policyManager.

yanivagman:
> I think we should invest in a more robust way to convert datastores in general into sources for the analyze mode in this case, rather than adding directly as a pipeline event (although it could be a solution that datastores emit events).
> 
> An example for such a design could be that each datastore would implement a "commitlog" interface. So changes can only be done through commits, which the datastore needs to keep track of (if its designed to be used in analyze mode later - maybe some datastore authors in plugins will not have this intention). Then, we could set a flag that dumps the commit-log after tracee's run. Maybe we would have `CommitDataStore` which inherits `DataStore`, and only `CommitDataStore`s can work in analyze mode.

I agree that we should invest in data stores and to keep its state for later analysis. This will require a careful design and implementation though. So for now, I think we can continue with the approach of having the "tracee_info" event, which can also signal the start of the event stream (thus it can be seen as an event on its own)

PR #4167, opened 2024-07-02
Fix(events): don't remove fork excess args

rscampos:
Part of https://github.com/aquasecurity/tracee/issues/3520

05ffffb98 **Fix(events): don't remove fork excess args**

```
Stop removing arguments from the `sched_process_fork` event.
This is removing an edge case making it easier to use the event.
It also makes it easier to reproducing the process tree.

The real parent process variable name 'up_parent_xxx' has been
renamed to 'parent_process_xxx'.

```

### 3. Other comments

Credits to @AlonZivony, who did the work.


yanivagman:
> @yanivagman , if I understood correctly from this comment ([09f4685#r1660126189](https://github.com/aquasecurity/tracee/commit/09f4685cdd3fddcbca1b6f1c6a460889342b85c3#r1660126189)), the ideia is to replace the following to `parent_process_xxx` instead of `up_parent_xxx` in the core.go?
> 
> https://github.com/aquasecurity/tracee/blob/3cab2050cae54f64760fb9dfc583da3f1f9ecabb/pkg/events/core.go#L11175-L11179
> 
> But this way won't be confused with:
> 
> https://github.com/aquasecurity/tracee/blob/3cab2050cae54f64760fb9dfc583da3f1f9ecabb/pkg/events/core.go#L11162-L11166
> 
> If it's the case, should we try something different like `ancestor_xxx`?

I agree that it is confusing.
I think if we will have parent_xxx and ancestor_xxx it will also be confusing.
It would be probably best to rename both so we will have parent_process_xxx and parent_task_xxx (which can also be a thread), but that may break existing signatures. For now, I think we can go with parent_process_xxx and keep parent_xxx as is, in the future we may break the other fields by introducing a new version of the event. WDYT?

yanivagman (approved):
LGTM

PR #4169, opened 2024-07-03
fix(pipeline): add ebpf caps in stack addres query

### LINKED ISSUE #3964: Missing capability for `option:stack-addresses`
## Description

Using the stack addresses option (`-o option:stack-addresses`) doesn't work (addresses array always comes out empty). Running with `--log debug` results in multiple of the following error:

```
{"L":"DEBUG","T":"2024-04-07T07:22:15.869Z","M":"failed to get StackAddress","error":"failed to lookup value 0xc001604070 in map stack_addresses: operation not permitted","origin":"ebpf:pkg/ebpf/events_pipeline.go:652","calls":"(*Tracee).getStackAddresses()"}
```

Running with `--capabilities add=cap_bpf` fixes the issue. `cap_bpf` needs to be added automatically before querying the stack addresses map.

```
Tracee version: v0.20.0
```

## Output of `uname -a`:

```
Linux *********** 5.15.146.1-microsoft-standard-WSL2 #1 SMP Sun Mar 10 18:17:47 IST 2024 x86_64 x86_64 x86_64 GNU/Linux
```

NDStrahilevitz:

36f4eb123 **fix(pipeline): add ebpf caps in stack addres query**

### 2. Explain how to test it

1. `tracee -o option:stack-addresses -o json`
2. Confirm stack addresses field is filled with a non-empty array

### 3. Other comments

Fix #3964

### DIFF
--- a/pkg/ebpf/events_pipeline.go
+++ b/pkg/ebpf/events_pipeline.go
@@ -9,6 +9,7 @@ import (
 	"unsafe"
 
 	"github.com/aquasecurity/tracee/pkg/bufferdecoder"
+	"github.com/aquasecurity/tracee/pkg/capabilities"
 	"github.com/aquasecurity/tracee/pkg/errfmt"
 	"github.com/aquasecurity/tracee/pkg/events"
 	"github.com/aquasecurity/tracee/pkg/logger"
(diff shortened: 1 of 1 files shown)


yanivagman:
This change may have a serious impact on performance if stack-addresses is selected.
Now on every event it will be required to change the capabilities of the tracee process, meaning performing a syscall and context switch to kernel.
I think a better approach is to set the required capbility for the entire run in case stack-addresses was selected.

NDStrahilevitz:
> This change may have a serious impact on performance if stack-addresses is selected. Now on every event it will be required to change the capabilities of the tracee process, meaning performing a syscall and context switch to kernel. I think a better approach is to set the required capbility for the entire run in case stack-addresses was selected.

You're right, this wasn't taken into consideration. This is a hotpath so we should likely revert it until we solve it in a better way (probably the one you've suggested).

PR #4171, opened 2024-07-03
fix(tests): remove named pipe if it exists

geyslan:

0d4e979bc **fix(tests): remove named pipe if it exists**

### 3. Other comments

This guarantees that the named pipe creation won't fail when `readWritePipe()` is called.


rscampos (approved):
LGTM

Issue #4172, opened 2024-07-03
dependencies manager is not thread-safe


geyslan:
## Description

Dependencies manager is not thread-safe, so trying to turn it so, using a mutex, causes it to dead lock.

PR #4173, opened 2024-07-03
Fix deps deadlock

geyslan:
Close: #4172 

2f8a61a0f **chore: run dependencies tests in parallel**
3616c30e9 **fix: dependencies manager deadlock**

2f8a61a0f **chore: run dependencies tests in parallel**

```
It also reformats some tests blocks.
```

3616c30e9 **fix: dependencies manager deadlock**

```
Since this turns the manager thread-safe, it takes the opportunity to
remove the temporary singleton pattern, reintroducing the manager as a
Tracee's member.
```

### DIFF
--- a/pkg/events/dependencies/manager.go
+++ b/pkg/events/dependencies/manager.go
@@ -19,19 +19,12 @@ const (
 	IllegalNodeType NodeType = "illegal"
 )
 
-var (
-	// Temporary singleton as a transition for decoupling. Manager will be a dependency injected.
-	managerInstance *Manager
-	once            sync.Once
-	mu              sync.Mutex // workaround mutex to protect the singleton instance in the tests
-)
-
 // Manager is a management tree for the current dependencies of events.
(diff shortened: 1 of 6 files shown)


pkg/events/dependencies/manager.go
```diff
@@ -136,13 +121,22 @@ func (m *Manager) UnselectEvent(id events.ID) bool {
 // no longer valid).
 // It returns if managed to remove the event, as it might not be present in the tree.
 func (m *Manager) RemoveEvent(id events.ID) error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	return m.removeEvent(id)
+}
+
+func (m *Manager) removeEvent(id events.ID) error {
```
geyslan, line 130: This is the dead lock fix. `RemoveEvent()`, a public method, was being called again after a lock by `removeEventDependents()`.

yanivagman (approved):
LGTM

PR #4178, opened 2024-07-09
feat(caps): base ebpf capabilities

NDStrahilevitz:

6bd0b3956 **feat(caps): base ebpf capabilities**

```
Add the ability to add the ebpf capabilities to the base ring, instead
of allocating a specific ring. When the feature is enabled, requests for
the eBPF ring will seamlessly avoid a ring switch, and the callback will
be called as is.
```

### 2. Explain how to test it

Stack addresses should remain working.

### DIFF
--- a/pkg/capabilities/capabilities.go
+++ b/pkg/capabilities/capabilities.go
@@ -27,14 +27,20 @@ const (
 )
 
 type Capabilities struct {
-	have   *cap.Set
-	all    map[cap.Value]map[RingType]bool
-	bypass bool
-	lock   *sync.Mutex // big lock to guarantee all threads are on the same ring
+	have     *cap.Set
+	all      map[cap.Value]map[RingType]bool
+	bypass   bool
+	baseEbpf bool
+	lock     *sync.Mutex // big lock to guarantee all threads are on the same ring
+}
+
+type Config struct {
+	Bypass   bool
+	BaseEbpf bool
 }
 
 // Initialize initializes the "caps" instance (singleton).
-func Initialize(bypass bool) error {
+func Initialize(cfg Config) error {
 	var err error
 
 	once.Do(func() {
@@ -44,20 +50,23 @@ func Initialize(bypass bool) error {
 		caps.lock.Lock()
 		defer caps.lock.Unlock()
 
-		err = caps.initialize(bypass)
+		err = caps.initialize(cfg)
 	})
 
 	return errfmt.WrapError(err)
 }
 
 // GetInstance returns current "caps" instance. It initializes capabilities if
-// needed, bypassing the privilege dropping by default.
+// needed, bypassing the privilege dropping by default, and not adding eBPF to the base.
 func GetInstance() *Capabilities {
 	capsMutex.Lock()
 	defer capsMutex.Unlock()
 
 	if caps == nil {
-		err := Initialize(true)
+		err := Initialize(Config{
+			Bypass:   true,
+			BaseEbpf: false,
+		})
 		if err != nil {
 			return nil
 		}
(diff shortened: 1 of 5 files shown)


pkg/capabilities/capabilities.go
```diff
@@ -187,7 +224,7 @@ func (c *Capabilities) EBPF(cb func() error) error {
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
-	if !c.bypass {
+	if !c.bypass || !c.baseEbpf {
```
yanivagman, line 227: Shouldn't this be `&&`?
We want to enter here only if both bypass and base ebpf is false, no?

pkg/capabilities/capabilities.go
```diff
@@ -196,7 +233,7 @@ func (c *Capabilities) EBPF(cb func() error) error {
 
 	errCb := cb() // callback
 
-	if !c.bypass {
+	if !c.bypass || !c.baseEbpf {
```
yanivagman, line 236: ditto

pkg/ebpf/tracee.go
```diff
@@ -275,7 +275,16 @@ func New(cfg config.Config) (*Tracee, error) {
 
 	// Initialize capabilities rings soon
 
-	err = capabilities.Initialize(t.config.Capabilities.BypassCaps)
+	useBaseEbpf := func(cfg config.Config) bool {
+		return cfg.Output.StackAddresses
+	}
```
yanivagman, line 280: Can't we simply add the required bpf capability to the base ring if StackAddresses is selected?

yanivagman (commented):
I don't see we removed the ebpf ring request from the hot path in the pipeline - shouldn't we do that?

NDStrahilevitz:
> I don't see we removed the ebpf ring request from the hot path in the pipeline - shouldn't we do that?

Now there's no need, because the EBPF callback automatically skips the capability switch if we include EBPF caps in the base. So its essentially like a mini-bypass.

pkg/capabilities/capabilities.go
NDStrahilevitz: Yep you're right.

pkg/ebpf/tracee.go
NDStrahilevitz: So if in the future we add other features which require a similar approach, we can add the logic here instead of directly coupling to stackaddresses.

yanivagman: Why coupling?
I mean, to expose some method to add more capabilities to the base ring, and then we can remove all the other changes in this PR, and just add the capability using this method

NDStrahilevitz: There already is such a method, but the eBPF caps specifically aren't constant across systems. So I thought initializing a specific bypass for eBPF in the code location where we do the calculation of what are these eBPF caps, would be easier.
About coupling, I meant that this is slightly less coupled, or at least that it makes the intention that this condition coud change in the future more clear, then doing `ebpfBase: cfg.Output.StackAddresses`.

yanivagman:
> > I don't see we removed the ebpf ring request from the hot path in the pipeline - shouldn't we do that?
> 
> Now there's no need, because the EBPF callback automatically skips the capability switch if we include EBPF caps in the base. So its essentially like a mini-bypass.

But this method still tries to acquire the capabilities lock in the hot path

NDStrahilevitz:
> > > I don't see we removed the ebpf ring request from the hot path in the pipeline - shouldn't we do that?
> > 
> > 
> > Now there's no need, because the EBPF callback automatically skips the capability switch if we include EBPF caps in the base. So its essentially like a mini-bypass.
> 
> But this method still tries to acquire the capabilities lock in the hot path

It wouldn't be the first lock acquisition we have inside the pipeline. The cgroups tracking for example does that. In any case, the performance overhead from capability switching was resolved. If there is lock contention because of this, we can reconsider, I'm not sure there will be (seeing as in similar situations so far we haven't).

yanivagman (approved):
Ok, we can continue with this change for now.
In the future, we can rethink how capabilities are handled.

PR #4179, opened 2024-07-09
chore: introduce eventFlags to policy manager

geyslan:

19320445b **chore: introduce eventFlags to policy manager**

```
It replaces the inner eventState type being a holder for the flags of an
event with a proper API.

In next efforts, it will replace the event state map embedded in Tracee.
```


yanivagman (approved):
LGTM

PR #4182, opened 2024-07-10
feat(ebpf): configurable pipeline channel size

NDStrahilevitz:

236d33ef0 **feat(ebpf): configurable pipeline channel size**

236d33ef0 **feat(ebpf): configurable pipeline channel size**

```
Allow configuring the event channels in pipeline stages using the
`pipeline-channel-size` flag and config entry. Rigorous testing of the
effect of playing with this value has yet to be done, so the default
value is currently 10000 as hardcoded before.

Tentatitevely, reducing the size of this configuration value may help
reduce the mean memory usage, perhaps reduce channel contention, but may
be at the cost of a greater risk for event loss.
```

### 2. Explain how to test it

1. `tracee --pipeline-channel-size X`
2. Configure in a config file with `pipeline-channel-size: X` at the root level

### DIFF
--- a/cmd/tracee/cmd/root.go
+++ b/cmd/tracee/cmd/root.go
@@ -212,6 +212,16 @@ func initCmd() error {
 		return errfmt.WrapError(err)
 	}
 
+	rootCmd.Flags().Int(
+		"pipeline-channel-size",
+		10000,
+		"<size>\t\t\t\tSize, in event objects, of each pipeline stage's output channel",
+	)
+	err = viper.BindPFlag("pipeline-channel-size", rootCmd.Flags().Lookup("pipeline-channel-size"))
+	if err != nil {
+		return errfmt.WrapError(err)
+	}
+
 	rootCmd.Flags().StringArrayP(
 		"cache",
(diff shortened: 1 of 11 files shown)


geyslan (approved):
LGTM

The failing integration seems related to capture.

cmd/tracee/cmd/root.go
```diff
@@ -212,6 +212,16 @@ func initCmd() error {
 		return errfmt.WrapError(err)
 	}
 
+	rootCmd.Flags().Int(
+		"pipeline-channel-size",
+		10000,
+		"<size>\t\t\t\tSize, in event objects, of each pipeline stage's output channel",
+	)
+	err = viper.BindPFlag("blob-perf-buffer-size", rootCmd.Flags().Lookup("blob-perf-buffer-size"))
```
yanivagman, line 220: `blob-perf-buffer-size`? shouldn't this be `pipeline-channel-size`?

NDStrahilevitz: It should be.

PR #4183, opened 2024-07-10
chore(ebpf): optimize filldir64 program

NDStrahilevitz:

0aa855cbf **chore(ebpf): optimize filldir64 program**

```
The filldir64 program runs very frequently, and as such incurs great
overhead. However, its usecase only requires the submission of events
from non 0 process inodes. Since this filter is faster than scope
evaluation, move this condition check before program data initialization
and scope evaluation.

Optimization reduces the average runtime to ~18% of the previous
overhead. Note, that even for more run amounts, the overall runtime is
now 27% from the previous version.

Before:
PROGRAM: filldir64 (type: kprobe, runtime: 61047110 ns, amount: 53135 times, average: 1148 ns)
After:
PROGRAM: filldir64 (type: kprobe, runtime: 16507056 ns, amount: 80350 times, average: 205 ns)
```

### 2. Explain how to test it

`hidden_inodes` event should work the same.

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -1557,6 +1557,11 @@ int tracepoint__sched__sched_switch(struct bpf_raw_tracepoint_args *ctx)
 SEC("kprobe/filldir64")
 int BPF_KPROBE(trace_filldir64)
 {
+    // only inode=0 is relevant, simple filter prior to program run
+    unsigned long process_inode_number = (unsigned long) PT_REGS_PARM5(ctx);
+    if (process_inode_number != 0)
+        return 0;
+
     program_data_t p = {};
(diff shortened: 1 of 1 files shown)


OriGlassman (approved):

geyslan:
Awesome. ❤️

I have only one question: as `evaluate_scope_filters()` updates `p`, it could somehow jeopardise some expected behaviour?

Doubt:

Commit says:

`However, its usecase only requires the submission of events
from non 0 process inodes.`

Wouldn't it be the opposite? `... only ... from 0 process...`?

yanivagman (approved):
LGTM

NDStrahilevitz:
> I have only one question: as `evaluate_scope_filters()` updates `p`, it could somehow jeopardise some expected behaviour?

I don't think so, since this isn't a program_data struct which is shared later on.

PR #4184, opened 2024-07-11
chore: use libbpfgo to check bpf helper func

rscampos:

ee975c0eb **chore: use libbpfgo to check bpf helper func**

```
Instead of checking kernel symbols, determine if the helper is available
based on the BPF program type. To achieve this, use the
BPFHelperIsSupported() function from libbpfgo.
```

### DIFF
--- a/pkg/ebpf/tracee.go
+++ b/pkg/ebpf/tracee.go
@@ -3,12 +3,14 @@ package ebpf
 import (
 	gocontext "context"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"os"
 	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
+	"syscall"
 	"unsafe"
 
 	"kernel.org/pub/linux/libs/security/libcap/cap"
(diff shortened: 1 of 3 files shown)


pkg/ebpf/tracee.go
```diff
@@ -541,20 +543,21 @@ func (t *Tracee) Init(ctx gocontext.Context) error {
...
 	// otherwise, we get a warning.
 	usedClockID := traceetime.CLOCK_BOOTTIME
-	err = capabilities.GetInstance().Specific(
+	err = capabilities.GetInstance().EBPF(
 		func() error {
+			supported, err := bpf.BPFHelperIsSupported(bpf.BPFProgTypeKprobe, bpf.BPFFuncKtimeGetBootNs)
+
+			// only report if operation not permitted
+			if errors.Is(err, syscall.EPERM) {
+				return err
```
geyslan, line 552: To avoid shadowing and confusion. `e` is just a suggestion, check if it also doesn't shadow.

```suggestion
			supported, e := bpf.BPFHelperIsSupported(bpf.BPFProgTypeKprobe, bpf.BPFFuncKtimeGetBootNs)

			// only report if operation not permitted
			if errors.Is(e, syscall.EPERM) {
				return err
```

pkg/ebpf/tracee.go
```diff
@@ -541,20 +543,21 @@ func (t *Tracee) Init(ctx gocontext.Context) error {
...
+			}
+
 			// If bpf_ktime_get_boot_ns is not available, eBPF will generate events based on monotonic time.
```
geyslan, line 555:
```suggestion
			// If BPFFuncKtimeGetBootNs is not available, eBPF will generate events based on monotonic time.
```

geyslan (approved):
LGTM

pkg/ebpf/tracee.go
rscampos: good catch, thank you for the suggestion but think `innerErr` would be better, WDYT?

PR #4185, opened 2024-07-12
chore(cap): check if cap is supported before set/unset

rscampos:

93b17685d **chore(cap): check if cap is supported before set/unset**

```
set or unset a capability without proper checking can result in a panic.
```

### 2. Explain how to test it

Test the following code in an older kernel version <5.8.
```
err = capabilities.GetInstance().Specific(
	func() error {
		return nil
	},
	cap.BPF,
	cap.PERFMON,
)
if err != nil {
	return errfmt.WrapError(err)
}
```

### 3. Other comments

If `CAP_BPF`+`CAP_PERFMON` is used on kernel version <5.8, we get a `panic: assignment to entry in nil map` when set/unset the map that was initialized. It happens because both cap are supported only on kernel version >=5.8.

### DIFF
--- a/pkg/capabilities/capabilities.go
+++ b/pkg/capabilities/capabilities.go
@@ -409,17 +409,23 @@ func (c *Capabilities) setProc() error {
 
 func (c *Capabilities) set(t RingType, values ...cap.Value) error {
 	for _, v := range values {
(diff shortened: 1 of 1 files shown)


yanivagman (approved):
LGTM

PR #4186, opened 2024-07-12
fix(build): add the include to 3rdparty libbpf during libbpfgo compilation

rscampos:

b1e17b7a6 **fix(build): include libbpf during libbpfgo compilation**

```
An issue appears when we try to compile libbpfgo on an older kernel that
doesn't have the XDP_FLAGS_REPLACE defined in
/usr/include/linux/if_link.h. To overcome this, it is necessary to
include the directory of libbpf under 3rdparty during the compilation,
since install_uapi_headers doesn't install if_link.h.
```

### 3. Other comments

Thank you @geyslan for the guidance through this issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -237,7 +237,6 @@ env:
 	@echo "CGO_EXT_LDFLAGS_RULES    $(CGO_EXT_LDFLAGS_RULES)"
 	@echo ---------------------------------------
 	@echo "GO_ENV_EBPF              $(GO_ENV_EBPF)"
-	@echo "GO_ENV_RULES             $(GO_ENV_RULES)"
 	@echo ---------------------------------------
 	@echo "TRACEE_SRC               $(TRACEE_SRC)"
 	@echo "TRACEE_SRC_DIRS          $(TRACEE_SRC_DIRS)"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -349,6 +348,7 @@ $(OUTPUT_DIR)/btfhub:
 LIBBPF_CFLAGS = "-fPIC"
 LIBBPF_LDFLAGS =
 LIBBPF_SRC = ./3rdparty/libbpf/src
+LIBBPF_INCLUDE = $(abspath ./3rdparty/libbpf/include/uapi)
```
geyslan, line 351: At line 401 we have `-I ./3rdparty/include \` for `$(OUTPUT_DIR)/tracee.bpf.o` target. I ask you, how about to also change it to use the variable?

A doubt: the full uapi path should be used? wouldn't only `-I ./3rdparty/include` work?

rscampos: We need to include full `./3rdparty/libbpf/include/uapi` in order to work. So I'll rename to `LIBBPF_INCLUDE_UAPI` as discussed offline, tks for the tip.

geyslan (approved):
LGTM

PR #4189, opened 2024-07-15
fix: inner error inside check for ebpf func

rscampos:

0ac18e3e4 **fix: inner error inside check for ebpf func**


geyslan (approved):
LGTM

PR #4191, opened 2024-07-16
Events flags embedding

geyslan:

4dfd7da92 **chore: inject deps manager into policy manager**
9aff9d2bb **chore(policy): introduce ManagerConfig**
443869ea7 **fix: validateKallsymsDependencies() nil deref**

4dfd7da92 **chore: inject deps manager into policy manager**

```
It makes the access to event flags be done through the policy manager,
which is responsible for setting up the dependencies manager.

This also:

- Turn the event flags access thread-safe.
- Renames manager and config types.
- Fix t.eventSignatures[eventId] logic, now accessible via policy
  manager IsRequiredBySignature(), since the former was only flagging
  the event as signature, but not as required by one.
```

### 3. Other comments

This is a enabler for https://github.com/aquasecurity/tracee/issues/3987.

### DIFF
--- a/pkg/ebpf/events_pipeline.go
+++ b/pkg/ebpf/events_pipeline.go
@@ -268,9 +268,9 @@ func (t *Tracee) decodeEvents(ctx context.Context, sourceChan chan []byte) (<-ch
 			// thus the need to continue with those within the pipeline.
 			if t.matchPolicies(evt) == 0 {
 				_, hasDerivation := t.eventDerivations[eventId]
-				_, hasSignature := t.eventSignatures[eventId]
+				reqBySig := t.policyManager.IsRequiredBySignature(eventId)
 
-				if !hasDerivation && !hasSignature {
+				if !hasDerivation && !reqBySig {
 					_ = t.stats.EventsFiltered.Increment()
 					t.eventsPool.Put(evt)
 					continue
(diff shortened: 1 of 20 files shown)


geyslan:
It's passing all the tests. Gonna polish it and leave it for review asap.

pkg/ebpf/events_pipeline.go
```diff
@@ -268,9 +268,9 @@ func (t *Tracee) decodeEvents(ctx context.Context, sourceChan chan []byte) (<-ch
 			// thus the need to continue with those within the pipeline.
 			if t.matchPolicies(evt) == 0 {
 				_, hasDerivation := t.eventDerivations[eventId]
-				_, hasSignature := t.eventSignatures[eventId]
+				isSignature := t.policyManager.IsSignature(eventId)
```
yanivagman, line 271: I don't think this change is correct.
We want to check if the event have signatures that depend on it (`hasSignature`) and not if the event itself is a signature (`isSignature`)

yanivagman (commented):
Looks like a very good refactor required for future changes!

pkg/ebpf/events_pipeline.go
geyslan: I think I've changed the nomenclature but need to double check the logic. Soon I'll be returning to this. Tks.

geyslan: Well, you are right. But I detected the main issue. As I only moved logic into the new place, I noticed that it didn't save the events in the map as required by sigs, it just set if it's a signature. So, at that stage it didn't work as a shortcut.

My last change includes:

```
  Fix t.eventSignatures[eventId] logic, now accessible via policy
  manager IsRequiredBySignature(), since the former was only flagging
  the event as signature, but not as required by one.
```

geyslan: E2E 1783: ✅

rscampos (approved):
LGTM

pkg/ebpf/events_pipeline.go
rscampos: @geyslan, I tested it and compared it with the main branch (which, as you mentioned, has a bug).

Issue #3870, opened 2024-02-19
enrichment error message


josedonizetti:
I keep seem messages like when running on minikube. Not sure if they are a  bug because the cgroup that they are trying to enrich isn't from a container. Thought it gives the wrong message to a user, it seems like something is not working. Does it make sense to keep logging them as errors? Maybe change the severity? Anything else?

```
{
    "level":"error",
    "ts":1708089649.258455,
    "msg":"error enriching container in control plane",
    "error":"containers.(*Containers).EnrichCgroupInfo: cgroup 74730: no containerId (path /user.slice/user-1000.slice/user@1000.service/app.slice/snap.kubectl.kubectl-186be11b-11d3-462e-821c-83ce6d742804.scope)",
    "cgroup_id":74730
}
```

josedonizetti:
@NDStrahilevitz  WDYT?

NDStrahilevitz:
Seems to me there's a bug here, as this pattern should not match to a container but to a pod.
So the log did its work in identifying the bug. I'll add this to the milestone and likely self assign later.

josedonizetti:
@NDStrahilevitz AH, nice, so it is indeed a bug! Thankyou

NDStrahilevitz:
@josedonizetti Can you try reproducing this? I got an error with the enrichment that the container doesn't exist (because I didn't manually setup the minikube dokcer socket), which means I did get the container id, unlike your case. Maybe different minikube versions?

PR #4193, opened 2024-07-17
fix(controlplane): filter unnecessary enriches

NDStrahilevitz:

613986ac5 **fix(controlplane): filter unnecessary enriches**

```
The condition for enrichment in the control plane was faulty. It relied
on the previous check for container relevant cgroups + not dead dirs.
This, however, is obviously not enough. Therefore the condition for
enrichment trigger now includes checking if the cgroup is a container's
root directory.
```

### 2. Explain how to test it

`tracee -o json`
`docker run --rm -d ubuntu`
Enrichment should work as before, with no new errors.

### DIFF
--- a/pkg/ebpf/controlplane/cgroups.go
+++ b/pkg/ebpf/controlplane/cgroups.go
@@ -47,8 +47,9 @@ func (ctrl *Controller) processCgroupMkdir(args []trace.Argument) error {
 		return errfmt.WrapError(err)
 	}
 
-	if !ctrl.enrichDisabled {
-		// If cgroupId belongs to a container, enrich now (in a goroutine)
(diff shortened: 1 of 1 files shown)


geyslan (approved):
LGTM

geyslan:
It arose in upcoming changes: https://github.com/aquasecurity/tracee/actions/runs/9973741398/job/27564629591?pr=4191#step:5:1455

```json
{"level":"error","ts":1721217866.2903428,"msg":"error enriching container in control plane","error":"containers.(*Containers).EnrichCgroupInfo: cgroup 4633: no containerId (path /system.slice/motd-news.service)","cgroup_id":4633}
```

NDStrahilevitz:
Possibly relevant #3870

PR #4197, opened 2024-07-22
improve flag parsing performance

geyslan:
### 1. Explain what the PR does

8d86c0c5d **chore(events): optimize ParseArgs for uintptr conv**
d4144ae50 **chore(events): improve flag parsing performance**

8d86c0c5d **chore(events): optimize ParseArgs for uintptr conv**

```
- Improved execution time by 20.78% and reduced byte operations by 55%.
- Changed loop to iterate by index for direct modification of
  argument values instead of iterating again via SetArgValue.
- Replaced string concatenation with strconv.AppendUint for efficient
  string construction.

Running tool: /home/gg/.goenv/versions/1.22.4/bin/go test -benchmem
-run=^$ -tags ebpf
-bench ^(BenchmarkParseArgs_Uintptr|BenchmarkParseArgs_Old)$
github.com/aquasecurity/tracee/pkg/events -benchtime=10000000x -race

goos: linux
goarch: amd64
pkg: github.com/aquasecurity/tracee/pkg/events
cpu: AMD Ryzen 9 7950X 16-Core Processor
=== RUN   BenchmarkParseArgs_Uintptr
BenchmarkParseArgs_Uintptr
BenchmarkParseArgs_Uintptr-32     10000000               609.9 ns/op           144 B/op          7 allocs/op
=== RUN   BenchmarkParseArgs_Old
BenchmarkParseArgs_Old
BenchmarkParseArgs_Old-32         10000000               769.7 ns/op           320 B/op          8 allocs/op
PASS
ok      github.com/aquasecurity/tracee/pkg/events       14.829s
```

d4144ae50 **chore(events): improve flag parsing performance**

```
Move helpers functions out of the ParseArgs method which reduces the
overhead of defining closures on each call to ParseArgs.

It reduces ~13,3% of processing time in 10mi iterations, as benchmarked
below:

Running tool: /home/gg/.goenv/versions/1.22.4/bin/go test -benchmem
-run=^$ -tags ebpf -bench ^(BenchmarkParseArgs|BenchmarkParseArgsOld)$
github.com/aquasecurity/tracee/pkg/events -benchtime=10000000x -race

goos: linux
goarch: amd64
pkg: github.com/aquasecurity/tracee/pkg/events
cpu: AMD Ryzen 9 7950X 16-Core Processor
=== RUN   BenchmarkParseArgs
BenchmarkParseArgs
BenchmarkParseArgs-32           10000000 2190 ns/op  0 B/op  0 allocs/op
[...]

PR #4198, opened 2024-07-23
improve flag parsing performance continuation

geyslan:

3adc191b0 **chore(events): parseSyscall with no lock**

```
Bypass the lock contention accessing the read-only map directly.
```

115a85c86 **chore(events): reduce ParseArgs complexity**

```
This reduces the complexity of the ParseArgs function by extracting
different parts of it into separate helpers. This makes the code
easier to read and understand and also makes it smaller.

Before:
1a080c0      14863 T github.com/aquasecurity/tracee/pkg/events.ParseArgs

After:
1a0cee0      11632 T github.com/aquasecurity/tracee/pkg/events.ParseArgs

It's a reduction of 21.73% in the size of the function what should also
improve cache locality and performance.

The benchmark tests also show a significant improvement in performance,
with a reduction of 6.4% in the time it takes to parse the arguments.

-run=^$ -tags ebpf -bench ^(BenchmarkParseArgsPrev|BenchmarkParseArgs)$
github.com/aquasecurity/tracee/pkg/events -benchtime=20000000x -race

goos: linux
goarch: amd64
pkg: github.com/aquasecurity/tracee/pkg/events
=== RUN   BenchmarkParseArgsPrev
BenchmarkParseArgsPrev
BenchmarkParseArgsPrev-32               20000000              2002 ns/op               0 B/op          0 allocs/op
=== RUN   BenchmarkParseArgs
BenchmarkParseArgs
BenchmarkParseArgs-32                   20000000              1873 ns/op               0 B/op          0 allocs/op
```

### 3. Other comments

Note that the benchmark measurements are non-deterministic.

### DIFF
--- a/pkg/events/parse_args_helpers.go
+++ b/pkg/events/parse_args_helpers.go
@@ -1,0 +1,238 @@
+package events
+
+import (
+	"github.com/aquasecurity/tracee/pkg/events/parsers"
+	"github.com/aquasecurity/tracee/pkg/logger"
+	"github.com/aquasecurity/tracee/types/trace"
+)
+
+func parseMMapProt(arg *trace.Argument, prot uint64) {
+	mmapProtArgument := parsers.ParseMmapProt(prot)
+	arg.Type = "string"
+	arg.Value = mmapProtArgument.String()
+}
+
+func parseSocketDomainArgument(arg *trace.Argument, domain uint64) {
+	arg.Type = "string"
+	socketDomainArgument, err := parsers.ParseSocketDomainArgument(domain)
+	if err != nil {
+		arg.Value = ""
+		return
+	}
+	arg.Value = socketDomainArgument.String()
+}
+
+func parseSocketType(arg *trace.Argument, typ uint64) {
+	arg.Type = "string"
+	socketTypeArgument, err := parsers.ParseSocketType(typ)
+	if err != nil {
+		arg.Value = ""
+		return
+	}
+	arg.Value = socketTypeArgument.String()
+}
+
+func parseInodeMode(arg *trace.Argument, mode uint64) {
+	arg.Type = "string"
+	inodeModeArgument, err := parsers.ParseInodeMode(mode)
+	if err != nil {
+		arg.Value = ""
+		return
+	}
+	arg.Value = inodeModeArgument.String()
+}
+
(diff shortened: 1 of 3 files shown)


geyslan:
@NDStrahilevitz when reviewing this, consider only the two last commits since it's based on top of #4197.

pkg/events/parse_args_helpers.go
```diff
@@ -68,13 +68,14 @@ func parseMemProtAlert(arg *trace.Argument, alert uint32) {
...
 
 func parseSyscall(arg *trace.Argument, id int32) {
-	if Core.IsDefined(ID(id)) {
-		eventDefinition := Core.GetDefinitionByID(ID(id))
-		if eventDefinition.IsSyscall() {
-			arg.Value = eventDefinition.GetName()
-			arg.Type = "string"
-		}
+	// bypass the lock contention accessing the read-only map directly
+	def, ok := CoreEvents[ID(id)]
```
rscampos, line 72: Is there any issue of doing that, accessing directly and no using through the Core API?

rscampos (approved):
LGTM

geyslan:
> Regarding the size of the function: reduction of 3455 bytes (22,33%) in the new version but the size of the Tracee binary increase by 736 bytes (probably because you create another file with the helpers function).

Exactly. Tks for bringing that data.

pkg/events/parse_args_helpers.go
geyslan: Let me double check this.

geyslan: Well, good catch @rscampos. Currently this change isn't a problem, but in the future it will cause data races due to the upcoming runtime changes to `CoreEvents`. That made me wonder why keep the non-changing definitions (syscalls) under lock contention? Would it be feasible to segregate them from the sigs? @yanivagman @NDStrahilevitz WDYT?

geyslan: Well, this brings a significant improvement in CPU time, although it will cause some concurrency issues in the future. I'm in favor of merging it (giving more information in the comment) to be a starting point for discussion about internal event segregation. If that's not feasible, we can easily fall back to the lock-contention way of retrieving the definition.

PR #4200, opened 2024-07-24
chore(parsers): optimize ParseMmapProt

geyslan:

ae857dafa **chore(parsers): optimize ParseMmapProt**

```
Refactor ParseMmapProt to use mmapProtArgs slice.

- Introduced mmapProtArgs slice for cleaner and more maintainable code.
- Replaced multiple if statements with a loop and string builder.
- Improved performance: reduced allocs/op from 17 to 15, and B/op from
  392 to 352.
- Execution time slightly improved from 1603 ns/op to 1536 ns/op
  (benchmark file used included).
- Reduced function size from 1534 to 677 bytes (55.8% less), what
  should improve locality and reduce cache misses.
```

### DIFF
--- a/pkg/events/parsers/data_parsers.go
+++ b/pkg/events/parsers/data_parsers.go
@@ -1443,6 +1443,16 @@ var (
 
 // revive:enable
 
+var mmapProtArgs = []MmapProtArgument{
+	PROT_READ,
+	PROT_WRITE,
+	PROT_EXEC,
+	PROT_SEM,
+	// PROT_NONE, // special case checked before the loop in ParseMmapProt
+	PROT_GROWSDOWN,
+	PROT_GROWSUP,
+}
+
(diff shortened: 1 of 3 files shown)


rscampos:
I successfully reproduced the benchmark:
```
goos: linux
goarch: arm64
pkg: github.com/aquasecurity/tracee/pkg/events/parsers
=== RUN   BenchmarkParseMmapProtOld
BenchmarkParseMmapProtOld
BenchmarkParseMmapProtOld-4             10000000               423.9 ns/op           392 B/op         17 allocs/op
=== RUN   BenchmarkParseMmapProtNew
BenchmarkParseMmapProtNew
BenchmarkParseMmapProtNew-4             10000000               395.1 ns/op           352 B/op         15 allocs/op
PASS
ok      github.com/aquasecurity/tracee/pkg/events/parsers       12.142s
```

PR #4202, opened 2024-07-25
fix: Prevent loading syscall-specific BPF programs for non-syscall events

yanivagman:

Previously, tracee incorrectly assumed all tailcall indexes were syscall IDs, leading to unnecessary loading of syscall-specific BPF programs. This caused a resource leak due to unused tracepoint attachments. The issue is now fixed by only removing the code which attaches these programs, as we explicitly define it in the event dependencies.

### 2. Explain how to test it

Run tracee with
`sudo ./dist/tracee -e sched_process_exec`
Before this fix, sys_enter and sys_exit events were attached, after this fix they are not anymore.
Use the following command to verify:
`sudo bpftool link`


OriGlassman (approved):
LGTM

PR #4203, opened 2024-07-25
fix: Ensure correct event dependency for process_execute_failed

yanivagman:

The process_execute_failed event was not functioning correctly due to a missing dependency on the sys_enter probe.
This commit attaches the required probe, allowing the event to trigger as expected.


OriGlassman (approved):
LGTM

PR #4204, opened 2024-07-25
Reduce save_args_to_submit_buf

geyslan:

### DIFF
--- a/pkg/ebpf/c/common/buffer.h
+++ b/pkg/ebpf/c/common/buffer.h
@@ -339,6 +339,33 @@ statfunc int save_sockaddr_to_buf(args_buffer_t *buf, struct socket *sock, u8 in
 
 #define DEC_ARG(n, enc_arg) ((enc_arg >> (8 * n)) & 0xFF)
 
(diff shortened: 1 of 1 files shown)


geyslan:
The idea of caching the sizes (hardcoded) in the stack to be retrieved by the type (index) only works in kernels after 5.10. Below that we face the wrath of the verifier: "variable stack access var_off". So this is being closed in favour of #4205.

In the future, when older kernels are no longer supported, perhaps this can be applied and tested for performance improvements.

geyslan:
The idea of caching the sizes (hardcoded) in the stack to be retrieved by the type (index) only works in kernels after 5.10. Below that we face the wrath of the verifier: "variable stack access var_off". So this is being closed in favour of https://github.com/aquasecurity/tracee/pull/4205.

In the future, when older kernels are no longer supported, perhaps this can be applied and tested for performance improvements.

geyslan:
This approach makes `save_args_to_submit_buf` 6.74ns faster (ns/per event).

PR #4209, opened 2024-07-29
chore: release bpf object memory

rscampos:

12c4a512a **chore: release bpf object memory**

```
These are the two references to the BPF object read from the file. Both
need to be set to nil for the Go garbage collector to reclaim the
memory.
```

### 2. Explain how to test it

To observe a real change in RSS, it's necessary to run Tracee with `GOGC=5` to trigger the GC cycle more frequently. This approach allows for a clearer view of the improvement in RSS size.

`sudo env GOGC=5 ./dist/tracee --events kill`

RSS size before this commit:
```
ps -p `pgrep tracee` -o pid,vsz,rss,cmd
    PID    VSZ   RSS CMD
  29655 2269948 101512 ./dist/tracee --events kill
```

RSS size in this commit:
```
ps -p `pgrep tracee` -o pid,vsz,rss,cmd
    PID    VSZ   RSS CMD
  26268 2265800 87796 ./dist/tracee --events kill
```

With this chore, we achieve around a 14MB reduction in RSS, which corresponds to the size of each BPF object. This reduction was possible after a small fix in this libbpfgo [PR#451](https://github.com/aquasecurity/libbpfgo/pull/451)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ toolchain go1.22.4
 require (
 	github.com/IBM/fluent-forward-go v0.2.2
 	github.com/Masterminds/sprig/v3 v3.2.3
-	github.com/aquasecurity/libbpfgo v0.7.0-libbpf-1.4.0.20240715144951-c8f76d124d7d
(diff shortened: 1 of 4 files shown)


geyslan (approved):
LGTM

PR #4212, opened 2024-07-30
`set_fs_pwd`: don't rely on sys_enter hook

oshaked1:

Get the argument for chdir from the task's registers instead of the arguments saved by sys_enter.

### 2. Explain how to test it

The test for this event was updated to also check compat binaries.

### 3. Other comments

Changed the behavior of the `get_task_pt_regs` function, see relevant commit.


yanivagman:
Merged #4228 which seems to be a dup of this one.
I would consider keeping the tests from this PR though

PR #4222, opened 2024-08-01
fix: necessary to Init engine before Start

rscampos:

814993d43 **fix: necessary to Init engine before Start**


geyslan (approved):
LGTM

PR #2085, opened 2022-08-21
events: add raw_ptrace

NDStrahilevitz:
## Initial Checklist

- [ ] There is an issue describing the need for this PR.
- [x] I have discussed this change with another maintainer.
- [x] Git log contains summary of the change.
- [x] Git log contains motivation and context of the change.
- [ ] If part of an EPIC, PR git log contains EPIC number.
- [ ] If part of an EPIC, PR was added to EPIC description.

## Description (git log)

```
Author: Nadav Strahilevitz <nadav.strahilevitz@aquasec.com>
Date:   Sun Aug 21 12:17:43 2022 +0000

    events: add raw_ptrace
    
    raw_ptrace has the same structure as the ptrace events but does not rely
    on the syscall sys_enter and sys_exit probes in order to submit.
    This will allow tracing ptrace events while using the lite enter and
    exit probes later on.

```

Fixes: #issue_number

## Type of change

- [ ] Bug fix (non-breaking change fixing an issue, preferable).
- [ ] Quick fix (minor non-breaking change requiring no issue, use with care)
- [ ] Code refactor (code improvement and/or code removal)
- [x] New feature (non-breaking change adding functionality).
- [ ] Breaking change (cause existing functionality not to work as expected).

## How Has This Been Tested?

Tests being included in this PR:

- [ ] Test File A
- [ ] Test File B

Reproduce the test by running:

- command 01
- command 02

## Final Checklist:

Pick "Bug Fix" or "Feature", delete the other and mark appropriate checks.

- [ ] I have made corresponding changes to the documentation.
- [ ] My code follows the style guidelines (C and Go) of this project.
- [ ] I have performed a self-review of my own code.
- [ ] I have commented all functions/methods created explaining what they do.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] My changes generate no new warnings.
- [ ] I have added tests that prove my fix, or feature, is effective.
- [ ] New and existing unit tests pass locally with my changes.
- [ ] Any dependent changes have been merged and published before.

## Git Log Checklist:

My commits logs have:

- [ ] Subject starts with "subsystem|file: description".
- [ ] Do not end the subject line with a period.
- [ ] Limit the subject line to 50 characters.
- [ ] Separate subject from body with a blank line.
- [ ] Use the imperative mood in the subject line.
- [ ] Wrap the body at 72 characters.
- [ ] Use the body to explain what and why instead of how.

PR #4231, opened 2024-08-05
feat(ebpf): make ptrace no rely on sys_enter/exit

OriGlassman:
make ptrace no rely on sys_enter/exit

### 2. Explain how to test it
./tracee -e=ptrace


pkg/ebpf/c/tracee.bpf.c
```diff
@@ -5133,6 +5133,31 @@ int BPF_KPROBE(trace_security_task_setrlimit)
     return events_perf_submit(&p, 0);
 }
 
+SEC("kprobe/ptrace")
```
NDStrahilevitz, line 5136: Not sure if there is a raw tracepoint for this, but i'm pretty sure that even a regular tracepoint has better performance than a kprobe.

NDStrahilevitz: According to @yanivagman in #2085, where I suggested something similar with a regular tracepoint, there should be an option for a raw tracepoint.

OriGlassman: If you succeed at doing a raw_tracepoint on a syscall that's indeed preferred. Let me know if that's the case.
tracepoint requires debugfs which is not currently a requirement of tracee.

yanivagman:
Replaced by #4256

PR #4232, opened 2024-08-05
make arch_prctl not rely on sys_enter/exit

OriGlassman:
make arch_prctl not rely on sys_enter/exit

### 2. Explain how to test it
./tracee -e=arch_prctl


yanivagman:
Replaced by #4256

PR #4234, opened 2024-08-06
fix: set engine to nil - sig benchmark

rscampos:

0d193c0bc **fix: set engine to nil - sig benchmark**

```
In order to avoid engine leak, set engine to nil and force GC.
```

### 3. Other comments

Thanks to @geyslan for spotting the Engine leak in every benchmark iteration. When a large benchtime value is used, the tasks get stuck.


geyslan (approved):
LGTM

PR #4236, opened 2024-08-07
fix: Remove unnecessary check for syscall wrapper in sys_enter tracepoint

yanivagman:

The sys_enter raw tracepoint always provides direct access to pt_regs,
regardless of the architecture's use of a syscall wrapper. This PR removes
the redundant conditional check and simplifies the logic for extracting system
call arguments.


geyslan (approved):
LGTM

PR #4238, opened 2024-08-08
feat(ebpf): use bpf_task_pt_regs when available

### LINKED ISSUE #4239: use bpf_task_pt_regs when available

OriGlassman:
Thanks @oshaked1 for writing this commit!

Tracee's `get_task_pt_regs` function mimics the logic of this helper, but relies on assuming some values defined in the kernel. This commit changes this function to use the helper if it is available. This helper must receive a task_struct with BTF info obtained from `bpf_get_current_task_btf`. From my experimentation, the verifier cannot determine whether a task_struct contains BTF info if it was stored outside of the stack. This means that this function cannot acutally retrieve the registers of any task except for the current one. The function name was changed to reflect this requirement, and it no longer receives a task_struct as a parameter.

### 2. Explain how to test it
./tracee

### 3. Other comments

Closes #4239


yanivagman (approved):
LGTM

PR #4245, opened 2024-08-12
fix(ebpf): use correct syscall id for compat

OriGlassman:
 use correct syscall id for compat

### 2. Explain how to test it
./tracee

### DIFF
--- a/pkg/ebpf/c/common/task.h
+++ b/pkg/ebpf/c/common/task.h
@@ -42,11 +42,22 @@ statfunc int get_task_flags(struct task_struct *task)
 statfunc int get_current_task_syscall_id(void)
 {
     // There is no originated syscall in kernel thread context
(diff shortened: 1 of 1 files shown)


NDStrahilevitz:
Is this intended to resolve the wrong `0xFFFFFFFF` syscalls on exit?

OriGlassman:
> Is this intended to resolve the wrong `0xFFFFFFFF` syscalls on exit?

For Compat syscalls, there was no translation to the non compat id. This meant that a check like
`
if (sys_id == SYSCALL_SOCKETCALL)
    ....
`
was never fulfilled (since sys_id was 102, but SYSCALL_SOCKETCALL is 476. After translation, sys_id is 476 as expected).

yanivagman (approved):
LGTM

PR #4246, opened 2024-08-12
chore: drop kernel support

geyslan:
Close: #4244 

### 1. Explain what the PR does

74768d6a1 **chore: drop kernel support**

PR #4247, opened 2024-08-12
remove e2e tests for kernels 5.4 and 4.18 on ARM

OriGlassman:
eBPF doesn't not support reading userspace memory in those kernels and below, leading to missing data. Thus, Tracee should not support those kernels.


geyslan:
dup of #4246

geyslan (approved):
LGTM

Issue #4248, opened 2024-08-12
PR action failing due to docker pull limitation


geyslan:
## Description

https://github.com/aquasecurity/tracee/actions/runs/10352531493/job/28653358900#step:5:1520

```
Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
Unable to find image 'ubuntu:latest' locally
docker: Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit.
See 'docker run --help'.
```

PR #4249, opened 2024-08-12
fix(ci): authenticate docker to bypass limitation

geyslan:
Close: #4248 

d8f972bce **fix(ci): authenticate docker to bypass limitation**


geyslan:
Other approach is in progress.

PR #4252, opened 2024-08-14
feat(time)!: epoch timestamps as standard

### LINKED ISSUE #3820: Processor functions order lead to problems with timestamps
## Description

Recently we changed the pipeline to have the `processEvent` stage, in which processor functions are called for the specific event.
We added support there for processor functions which are applied to all events, and used the `normalizeEventCtxTimes` this way.
However, this introduced some difficulties.
The processor functions for all events are called only after all the specific events processor functions are called.
This means that no processor function can use the timestamps as they are still monotonic (instead of absolute since epoch).
An example for a place struggling with it is the processor for `sched_process_exec` which creates the capture exec files. It uses the timestamp for the files captured, so since the change the times are not useful anymore.
This is not the only place which struggles with this.
We should introduce a mechanism for assigning processor functions both before after the time fix, according to their need.
I guess that in the future we can expect more problems like this caused by ordering.

```
(paste your output here)
```

## Output of `uname -a`:

```
(paste your output here)
```

### LINKED ISSUE #4287: feature: epoch timestamps as a standard
## Feature Proposal

Feature will do the following:
1. Simplify normalization logic - only monotonic/boot <-> epoch, remove timeNormalizer object
2. Remove `relative-time` flag - tracee only outputs epoch timestamps
3. Normalization done asap - ideally at point of event decoding, arguments processed first - new code can and should rely on the assumption of normalized timestamps

### Discussed in https://github.com/aquasecurity/tracee/discussions/4273

<div type='discussions-op-text'>

<sup>Originally posted by **NDStrahilevitz** September  3, 2024</sup>
Hey all.

As I'm working on code related to the event timestamps right now, I was struck by the option that no one is really using the relative time feature. The feature is currently enabled by appending the `-o option:relative-time` flag to tracee, or setting it in the config file. Once enabled, timestamp values in the event context and arguments are printed relative to the point in time in which tracee's process begun (so instead of ts: epoch_time_at_record -> ts: time_from_boot_at_record - tracee_recorded_starttime_from_boot).

I believe we have the following reasons to remove:
1. No intrinsic advantage over epoch time: relative time does not allow a comparison or timeline assessment of event occurence, especially given a crash/restart in a tracee daemonset. The only intrinsic advantage is a more or less precise duration from tracee's start, but even that can be approximated through the first event or other means on the host.
2. The plurality of timestamp outputs complicates the usage and implementation of ad-hoc analysis in the planned analyze mode (https://github.com/aquasecurity/tracee/issues/3520). The two solutions proposed for these are either "marking" the timestamp, such that any operation done on it can rely on the mark + awareness of ad-hoc analysis (which couples it to the notion of time), and/or relying on the `tracee_info` event which solely exists currently to inform of the state set for tracee in some pipeline work.
3. In contrast, it seems to me that by removing the relative time feature we could simply rely on the any event source converting its timestamps to epoch as standard. There will be no different outputs for events, anyone who wants to convert tracee event times to some "relative from" timestamp can do it on their own, per their requirements. Any addition of custom events will require, by standard, to set epoch timestamps on all relevant fields. This turns the issue of timestamps to an issue of standard compliance, and a compliance which is certainly possible.

Any thoughts? Any who are directly in favor or against?</div>

NDStrahilevitz:
14861156a **feat(time): normalize all time to epoch**
fbbe9f9c2 **feat(output)!: remove relative-time**
1a55bdb7f **chore(pipeline): streamline time args normalization**
18d3abdae **chore: refactor time pkg**

14861156a **feat(time): normalize all time to epoch**

```
Remove all timeNormalizer object logic. Instead use simple normalization
functions and normalize all timestamps in events asap:
1. Context timestamp normalization moved to decode stage
2. Relevant timestamp arguments normalized in processing stage,
   registered to run first.
```

1a55bdb7f **chore(pipeline): streamline time args normalization**

```
Instead of using a helper, and then using it in a specialized processor
function - use function currying to enable creating generic time argument
processors.
```
### 2. Explain how to test it

1. running with `-o option:relative-time` should fail
2. event context timestamps should be epoch
3. sched_process_fork timestamps in args should be epoch (relevant args are `parent_start_time`, `start_time`, `leader_start_time`, `parent_process_start_time`)

### 3. Other comments
Resolve #4287 
Resolve #3820


pkg/ebpf/tracee.go
```diff
@@ -1034,6 +1038,9 @@ func (t *Tracee) initKsymTableRequiredSyms() error {
 			}
 		}
 	}
+
+	// bpf boot time helper
+	t.requiredKsyms = append(t.requiredKsyms, "bpf_ktime_get_boot_ns")
```
geyslan, line 1043: I got you point, but it turns out a duplication as we already check it here: https://github.com/aquasecurity/tracee/blob/4b048001b82d4a0c8ac68943d077c95492852b6a/pkg/ebpf/tracee.go#L548

NDStrahilevitz: I recall there was check doing this with the ksymbol table before, though maybe i'm mistaken. If this is the only check this change is redundant as you say.

NDStrahilevitz:
@geyslan Is the PR ok besides the change you've commented on? I will drop that commit since it's irrelevant.

geyslan (approved):
LGTM

geyslan:
When rebasing, please put ! in the [breaking changes commits](https://www.conventionalcommits.org/en/v1.0.0/), like I did in the PR header. Tks.

NDStrahilevitz:
@rscampos Could you review this?

pkg/time/time.go
```diff
@@ -77,42 +76,77 @@ func getBootTimeInJiffies() int64 {
...
+// Run this function ASAP. Not running this function first will cause wrong behaviour
+// in other functions of the package.
+//
+// Reference points can be set from two differents clocks: CLOCK_MONOTONIC or CLOCK_BOOTTIME.
+// Tracee bpf code tries to use boottime clock if available, otherwise uses monotonic clock.
+// ClockGettime get time elapsed since start (boot) so tracee can calculate event timestamps.
+func Init(clockID int32) error {
+	var err error
+	initTimeOnce.Do(func() {
+		startTimeMonotonic, errIn := getClockTimeNS(clockID)
```
rscampos, line 98: Nit: The variable `startTimeMonotonic` could represent either the result from a Monotonic or Boottime clock, depending on the value of `clockID`. Therefore, it may be a good idea to consider renaming the variable (also in the rest of the function) to better reflect this.

pkg/ebpf/controlplane/processes.go
```diff
@@ -208,7 +214,7 @@ func (ctrl *Controller) procTreeExitProcessor(args []trace.Argument) error {
 
 	return ctrl.processTree.FeedFromExit(
 		proctree.ExitFeed{
-			TimeStamp:  uint64(ctrl.timeNormalizer.NormalizeTime(int(timestamp))), // time of exit is already a timestamp
+			TimeStamp:  time.BootToEpochNS(timestamp), // time of exit is already a times)p
```
rscampos, line 217: comment typo: `timestamp`

rscampos (approved):
LGTM

PR #4259, opened 2024-08-21
process_execute_failed: don't rely on sys_enter

oshaked1:

Use task registers to extract `argv` and `envp` instead of getting them from sys_enter.


yanivagman (approved):
LGTM

PR #4263, opened 2024-08-27
[v0.22.0] fix(build): cyclic dependency in makefile

geyslan:

02adacf5c **fix(build): cyclic dependency in makefile**

```
$(GOENV_MK) is a body level include, so using it as a prerequisite and
as a target dependency caused a cyclic dependency.

Use .eval_goenv rule (and placeholder file) to evaluate variables to be
stored in goenv.mk and use the placeholder as a prerequisite for the
targets that need the goenv.mk content.

commit: dd763f4 (main), cherry-pick
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,11 @@
 # dependencies, like source files, changed)
 .check*
 
+# .eval_* files are placeholders as .check* above, but focused on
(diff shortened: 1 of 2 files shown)


rscampos (approved):
LGTM

PR #4274, opened 2024-09-03
fix: release snapshot target arch

rscampos:

0e019b598 **fix: release snapshot target arch**

### DIFF
--- a/builder/Dockerfile.alpine-tracee-container
+++ b/builder/Dockerfile.alpine-tracee-container
@@ -2,7 +2,6 @@
 # Creates the official tracee containers.
 #
 
-ARG TARGETARCH
 ARG BTFHUB=0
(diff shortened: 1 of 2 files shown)


builder/Dockerfile.alpine-tracee-container
```diff
@@ -2,7 +2,6 @@
 # Creates the official tracee containers.
 #
 
-ARG TARGETARCH
```
geyslan, line 5: This should not be global, that's why it's not being accessible inside the stages. See https://docs.docker.com/build/building/variables.

geyslan (approved):
Despite it's working on my machine 🤣, we need the fix asap. Tks @rscampos. In the future we can try to use the TARGETARCH and BUILDPLATFORM variables (for instance) in different environments.

LGTM.

PR #4277, opened 2024-09-04
fix: create a Finding copy with Data copied too

geyslan:

499ae1ce2 **fix: create a Finding copy with Data copied too**

```
This avoids template internal (json parsing) datarace when iterating the
map and attempting to modify it.

Co-authored-by: Nadav Strahilevitz <nadav.strahilevitz@aquasec.com>
```


NDStrahilevitz (approved):
LGTM

geyslan:
Closing, as this is no longer needed. The data race was caused elsewhere.

It is worth noting that, since `tracee-rules` uses template pkg and it tries to parse the `Finding` object, the latter should always have read-only fields (be aware of slices and maps - reference types).

PR #4279, opened 2024-09-05
argument parsers improvements

geyslan:
Close: #4476 
Close: #4477 
Close: #4478 

This is a sequence of the effort already started by https://github.com/aquasecurity/tracee/pull/4200. As it was getting huge, the continuation will come after.

---

Improvements include reducing the size of the parser logic by using slices instead of maps whenever possible. This allows for direct access to values via index, and when direct access isn't feasible, iterating through slices is generally more efficient than fetching values from a map.

Return string only instead of the Argument type since it's the only value used.

This PR reduces the size of the final binary by ~56KB.

---

cb55f8f81 **fix(parsers): fix openFlagsValues O_LARGEFILE**
96763694d **feat: parse dirfd for special case AT_FDCWD**
af50e8b95 **perf: chore: add flag to ParseSocketDomainArgument**
4fc5f7cff **perf: chore: improve ParseSocketcallCall**
35613a5d2 **perf: chore: add flags ParsePtraceRequestArgument**
083cfd4ed **perf: chore: add missing flags to ParseBPFCmd**
6c2ca9886 **perf: chore: add missing flags to ParsePrctlOption**
88f4f92eb **perf: chore: add missing flags to ParseCapability**
9b8fe5834 **feat: perf: fix: AT flags parsing**
c99ebed90 **perf: fix: improve ParseAccessMode**
86b68aa53 **perf: fix: improve ParseOpenFlagArgument**
9a328a9a2 **perf: improve ParseCloneFlags**
2baff0489 **chore: add comment about raw values**
2d5e43aef **chore: deprecate internal interface**

cb55f8f81 **fix(parsers): fix openFlagsValues O_LARGEFILE**

```
O_LARGEFILE is different on arm64 and amd64, 0400000 and 00100000
respectively.
```

96763694d **feat: parse dirfd for special case AT_FDCWD**

```
syscalls with dirfd arg now parse for special case AT_FDCWD when
ParseArgumentsFDs is true.
```

af50e8b95 **perf: chore: add flag to ParseSocketDomainArgument**

```
Add missing flag to ParseSocketDomainArgument:
    AF_MCTP

Use slice instead of maps. This allows for direct access to values via
index.

Return string only instead of the Argument type since it's the only
value used.
```

4fc5f7cff **perf: chore: improve ParseSocketcallCall**

```
Use slice instead of maps. This allows for direct access to values via
index.

Return string only instead of the Argument type since it's the only
value used.
```

35613a5d2 **perf: chore: add flags ParsePtraceRequestArgument**

```
Add missing flags to ParsePtraceRequestArgument:
    PTRACE_GET_THREAD_AREA, PTRACE_SET_THREAD_AREA, PTRACE_ARCH_PRCTL,
    PTRACE_SYSEMU, PTRACE_SYSEMU_SINGLESTEP, PTRACE_SINGLEBLOCK,
    PTRACE_GET_RSEQ_CONFIGURATION,
    PTRACE_SET_SYSCALL_USER_DISPATCH_CONFIG and
    PTRACE_GET_SYSCALL_USER_DISPATCH_CONFIG.

Iterate slice instead of fetching a map.

Return string only instead of the Argument type since it's the only
value used.
```

083cfd4ed **perf: chore: add missing flags to ParseBPFCmd**

```
Add missing flags to ParseBPFCmd:
    BPF_PROG_BIND_MAP, BPF_TOKEN_CREATE

Use slice instead of maps. This allows for direct access to values via
index.

Return string only instead of the Argument type since it's the only
value used.
```

6c2ca9886 **perf: chore: add missing flags to ParsePrctlOption**

```
Add missing flags to ParsePrctlOption:
    PR_SET_IO_FLUSHER, PR_GET_IO_FLUSHER, PR_SET_SYSCALL_USER_DISPATCH,
    PR_PAC_SET_ENABLED_KEYS, PR_PAC_GET_ENABLED_KEYS, PR_SCHED_CORE,
    PR_SME_SET_VL, PR_SME_GET_VL, PR_SET_MDWE, PR_GET_MDWE,
    PR_SET_MEMORY_MERGE and PR_GET_MEMORY_MERGE.

Iterate slice instead of fetching a map.

Return string only instead of the Argument type since it's the only
value used.
```

88f4f92eb **perf: chore: add missing flags to ParseCapability**

```
Add missing flags to ParseCapability:
CAP_PERFMON, CAP_BPF and CAP_CHECKPOINT_RESTORE

Use slice instead of maps. This allows for direct access to values via
index.

Return string only instead of the Argument type since it's the only
value used.
```

9b8fe5834 **feat: perf: fix: AT flags parsing**

```
Introduce ParseFaccessatFlag to parse faccessat flags.
Introduce ParseFchmodatFlag to parse fchmodat flags.

Iterate slice instead of conditional branch.

Return string only instead of the Argument type since it's the only
value used.

Fix ParseExecFlag (now ParseExecveatFlag) to check only related flags.
```

c99ebed90 **perf: fix: improve ParseAccessMode**

```
Iterate slice instead of conditional branch.

Return string only instead of the Argument type since it's the only
value used.

Fix logic, since it wasn't printing F_OK flag alone.
```

86b68aa53 **perf: fix: improve ParseOpenFlagArgument**

```
When possible iterates slice instead of conditional branch.
Return string only instead of the Argument type since it's the only
value used.

Fix logic, since it wasn't printing O_RDONLY flag alone.
```

9a328a9a2 **perf: improve ParseCloneFlags**

```
Iterate slice instead of conditional branch.

Return string only instead of the Argument type since it's the only
value used.

| BenchmarkParseCloneFlags-32 | Before   | After  |
|-----------------------------|----------|--------|
| Execution Time (ns/op)      | 1117     | 696.7  |
| Memory Allocated (B/op)     | 1192     | 736    |
| Allocations per Operation   | 27       | 25     |
```

2d5e43aef **chore: deprecate internal interface**

```
Dealing with a concrete type instead of an interface is more efficient.

systemFunctionArgument interface will be replaced by
SystemFunctionArgument struct in future efforts.
```


geyslan:
This is only the beginning. So far this was done:

- performance improvements
- fix ParseOpenFlagArgument that wasn't printing O_RDONLY flag alone.
- fix ParseAccessMode that wasn't printing F_OK flag alone.
- fix ParseExecFlag (now ParseExecveatFlag) to check only related flags.
- fix Faccessat to parse flags instead of mode (wrong arg).
- feat: *at syscalls with dirfd arg now parses for special case AT_FDCWD.

geyslan:
This is a sequence of the effort already started by https://github.com/aquasecurity/tracee/pull/4200. As this was getting huge, the continuation will come after.

geyslan:
After all refactor done, I pretend to split it in different files. I also have an idea to try to reduce ParseArgs() logic, but gonna play with it only after finishing the whole data parsers.

pkg/events/parsers/data_parsers.go
```diff
@@ -13,11 +13,27 @@ import (
 	"github.com/aquasecurity/tracee/pkg/utils/environment"
 )
 
-type SystemFunctionArgument interface {
+
+type systemFunctionArgument interface {
```
yanivagman, line 39: Why making the interface private?
Isn't the idea behind interfaces to be exposed between different packages?

pkg/events/parsers/data_parsers.go
```diff
@@ -13,6 +13,28 @@ import (
...
+#include <linux/ptrace.h>
+#include <sys/socket.h>
+
+// defined in fcntl.h but it conflicts with linux/fcntl.h
+#define F_OK 0
+#define X_OK 1
+#define W_OK 2
+#define R_OK 4
+*/
+import "C"
```
yanivagman, line 37: I think we should avoid using cgo not in libbfgo so we can switch to pure go library in the future if will be required

yanivagman (commented):
I reviewed the first 2 PRs and added some doubts.
I assume the comments there will affect the rest of the code so I will continue the review after clarifying these points

pkg/events/parsers/data_parsers.go
geyslan: It is not used (imported) so I just made it private via https://github.com/aquasecurity/tracee/commit/c07114173ce275f33a73e3bdf468d59118303e2e by creating a public type with same name - since the type or its pointer is cheaper than dealing with an interface.

pkg/events/parsers/data_parsers.go
geyslan: Ok, we can avoid that, but all values should be inserted again as raw.

pkg/events/parsers/data_parsers.go
geyslan: As we discussed I'm going to remove the private interface (enforcer). 👍🏼

pkg/events/parsers/data_parsers.go
geyslan: Done.

pkg/events/parsers/data_parsers.go
geyslan: Well, in the future we'll remove that interface, other still not refactored parsers consumes it.

pkg/events/parse_args.go
```diff
@@ -23,7 +23,8 @@ func ParseArgs(event *trace.Event) error {
 		}
 	}
 
-	switch ID(event.EventID) {
+	evtID := ID(event.EventID)
```
geyslan, line 26: After all refactoring (optimisation) is complete, I plan to make ParseArgs() smaller (cache wise), fetching parsers from a slice or map based on its eventId. It will certainly help to deduplicate code and expand the parsing to other events.

pkg/events/parse_args.go
```diff
@@ -133,12 +145,20 @@ func ParseArgs(event *trace.Event) error {
...
 				parseInodeMode(modeArg, uint64(mode))
 			}
 		}
+		if evtID == Fchmodat {
+			if flagsArg := GetArg(event, "flags"); flagsArg != nil {
+				if flags, isInt32 := flagsArg.Value.(int32); isInt32 {
+					parseFchmodatFlag(flagsArg, uint64(flags))
+				}
+			}
+		}
```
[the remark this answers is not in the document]
geyslan: While changing this package (some PRs before and this) I noticed that ParseArgs won't escalate well, it gets bigger and bigger at every new supported function and/or args etc. So I'm not focused on it now.

After optimising the inner parsers, the main idea is to return to ParseArgs. To reduce it, identifying common helper parses for every system func, map them and call it accordingly. See: https://github.com/aquasecurity/tracee/pull/4279#discussion_r1888924182

pkg/events/parsers/data_parsers.go
```diff
@@ -40,1239 +56,889 @@ func optionIsContainedInArgument(rawArgument uint64, option uint64) bool {
...
+}
 
-func ParseExecFlag(rawValue uint64) (ExecFlagArgument, error) {
-	if rawValue == 0 {
-		return ExecFlagArgument{}, nil
+// ParseFaccessatFlag parses the `flags` bitmask argument of the `faccessat` syscall.
+func ParseFaccessatFlag(flags uint64) (string, error) {
+	if flags == 0 {
+		return "", nil
 	}
```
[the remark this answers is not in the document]
geyslan: ditto.

pkg/events/parsers/data_parsers.go
```diff
@@ -40,1239 +56,889 @@ func optionIsContainedInArgument(rawArgument uint64, option uint64) bool {
...
-	CAP_MAC_ADMIN
-	CAP_SYSLOG
-	CAP_WAKE_ALARM
-	CAP_BLOCK_SUSPEND
-	CAP_AUDIT_READ
-)
+func ParseFchmodatFlag(flags uint64) (string, error) {
+	if flags == 0 {
+		return "", nil
+	}
```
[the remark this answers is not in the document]
geyslan: ditto

pkg/events/parsers/data_parsers.go
```diff
@@ -40,1239 +56,889 @@ func optionIsContainedInArgument(rawArgument uint64, option uint64) bool {
...
-func ParseCapability(rawValue uint64) (CapabilityFlagArgument, error) {
-	v, ok := capabilitiesMap[rawValue]
-	if !ok {
-		return 0, fmt.Errorf("not a valid capability value: %d", rawValue)
+// ParseExecveatFlag parses the `flags` bitmask argument of the `execveat` syscall.
+// http://man7.org/linux/man-pages/man2/execveat.2.html
+func ParseExecveatFlag(flags uint64) (string, error) {
+	if flags == 0 {
+		return "", nil
 	}
```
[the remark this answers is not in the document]
geyslan: ParseCapability has a different logic, see: https://github.com/aquasecurity/tracee/pull/4279/files/06328f83a7a91d6b44e28d14bf1a86aa707e327e#diff-b7f34045358e980276266706956ef8a73b7a42c2df811875ccb375a863bc911dR445

pkg/events/parsers/data_parsers.go
```diff
@@ -63,6 +63,11 @@ type CloneFlagArgument struct {
 
 // revive:disable
 
+// Use always raw values for the constants, since unix/syscall constants are not
+// always set to the same values.
+// For example, `O_LARGEFILE` is defined as 0x8000 (00100000) in C include,
+// but as 0x0 in unix package.
+
```
yanivagman: This comment is not so clear to me. Use raw values where exactly?

pkg/events/parsers/data_parsers.go
```diff
@@ -760,156 +760,71 @@ func ParsePtraceRequestArgument(request uint64) (string, error) {
...
+	if call > SYS_NET_LAST {
+		return "", fmt.Errorf("not a valid socketcall call value: %d", call)
 	}
 
-	return 0, fmt.Errorf("not a valid socketcall call value: %d", rawValue)
+	idx := int(call - SYS_NET_FIRST)
+	return sysNetValues[idx].String(), nil
 }
 
-type SocketDomainArgument uint64
```
yanivagman, line 1108: What about socket domain type? No need for this?

pkg/events/parsers/data_parsers.go
geyslan: For all `rawValue`'s set in every XXXFlagArgument. It's just a hint to not trust in every const under Go libraries:

https://cs.opensource.google/go/x/sys/+/refs/tags/v0.29.0:unix/zerrors_linux_amd64.go;l=182
https://cs.opensource.google/go/x/sys/+/refs/tags/v0.29.0:unix/zerrors_linux_arm64.go;l=183

Under linux code:

<details>

image

</details>

The main issue is that those consts are fetched from the uapi, but it can be different based on the `#define`s set like `_LARGEFILE64_SOURCE` and `_FILE_OFFSET_BITS`, in this context. If used, those will set `O_LARGEFILE` as `0` if the binary is 64 bits. But if the binary is 32 bits, those macros will set `O_LARGEFILE` as `00100000` (x86_64) or `00400000` (arm64), it makes difference only for the glibc that wrappers calls like `fcntl64`.

```c
#define _LARGEFILE64_SOURCE
#define _FILE_OFFSET_BITS 64

#include <stdio.h>
#include <fcntl.h>

int main(void)
{
    printf("%o\n", O_LARGEFILE);
    fcntl(0, F_SETFD, O_LARGEFILE);
    return 0;
}
```

When built as 64 bits it calls:

`fcntl(0, F_SETFD, 0)` - O_LARGEFILE = 0


When buit as 32 bits it calls:

`fcntl64(0, F_SETFD, 0x8000 /* FD_??? */)` - O_LARGEFILE = 0x8000

pkg/events/parsers/data_parsers.go
geyslan: I'm abolishing the different types since all of them have the same behaviour of String and Value. I'm focusing on declaring flags (based on origin - include files) and declaring slices like `socketDomainValues = []SystemFunctionArgument{` which hold a instance copy of the related flags already made public. For the particular case of `SocketDomain`, `socketDomainValues` store the values and they are parsed by `ParseSocketDomainArgument()`.

You can better understand the idea looking the `AT_*` declarations and the "`faccessatFlagsValues`, `fchmodatFlagsValues` and `execveAtFlagsValues`" holders which make use of only a fragment of those flags based on the event to be parsed. Before we were iterating the whole bunch for events that don't use those flags.

geyslan:
@yanivagman rebased with a new fix.

yanivagman (approved):
LGTM

PR #4286, opened 2024-09-10
update syscall table: lookup_dcookie is removed

OriGlassman:
Kernels 6.7 and above does not have lookup_dcookie syscall anymore

update syscall table: lookup_dcookie is removed


geyslan:
@rscampos v0.22.1

PR #4294, opened 2024-09-10
[v0.22.1] container enrichment fixes and improvements

rscampos:

3f810f7c7 **feat(enrich): improve containerd image info enrich**

```
Make the image info query in containerd enrichment more robust.
Procedure now begins by first querying the containerd image service,
and only then using the cri directly as a fallback.

Additionally, fix a typo in the CRI query which appended the image name
as its digest, even when found.

commit: 126133a (main), cherry-pick
```

8b74d3959 **fix(enrich): silence noncontainer cgroup errors**

```
In case enrichment is requested on a non container cgroup, return the
metadata struct with an empty container id instead. User is responsible
for handling this "not-found" case by themselves.
Apply this in locations where enrichment is called (pipeline and control
plane).

commit: e4215bd (main), cherry-pick
```

### DIFF
--- a/pkg/containers/containers.go
+++ b/pkg/containers/containers.go
@@ -194,8 +194,11 @@ func (c *Containers) cgroupUpdate(
 
 // EnrichCgroupInfo checks for a given cgroupId if it is relevant to some running
 // container. It then calls the runtime info service to gather additional data from the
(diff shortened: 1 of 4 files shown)


geyslan (approved):
LGTM

PR #4297, opened 2024-09-11
Update Readme.md

LJTian:
Fix docs address


geyslan:
@LJTian thank you for this contribution. @itaysk do you think we should move that link from home to docs overview?

LJTian:
I have some opinions:
1. The context information in Readme.md points to the document and should not jump to the homepage.
2. However, there is no detailed introduction in the document, so I think it should be supplemented.

ShohamBit:
i think it is batter that it points to the hime and nit directly to the docs section, it's just make more sense to me

itaysk:
I don't have a strong opinion, but it does intentionally point to the home and not docs. I don't think this change is needed

geyslan:
Closing as not needed. @LJTian Nevertheless, please be our guest to contribute. If you have any questions, just ask us.

Issue #4255, opened 2024-08-18
Possible bug in Test_EventFilters when a case has many `cmdEvent` and each` cmdEvent` has different number of `expectEvent`. 


canoriz:
## Description

This is not an existing bug but a possible bug. I am developing a feature (a new kind of event), when adding a test for my feature, I encountered a bug in integration-tests. My test case looks like this, after running `ping`, tracee should emit `events.SchedProcessExec` and `events.MyDevelopingEvent`.
```go
		{
			name: "xxx",
			policyFiles: []testutils.PolicyFileWithID{
				{
					Id: 1,
					PolicyFile: v1beta1.PolicyFile{
						Metadata: v1beta1.Metadata{
							Name: "xxx",
						},
						Spec: k8s.PolicySpec{
							Scope:          []string{"global"},
							DefaultActions: []string{"log"},
							Rules: []k8s.Rule{
								{
									Event:   "sched_process_exec",
									Filters: []string{"comm=ping"},
									...
								},
								{
									Event:   "sched_process_exit",
									Filters: []string{"comm=who"},
								},
							},
						},
					},
				},
			},
			cmdEvents: []cmdEvents{
				newCmdEvents(
					"who",
					0,
					1*time.Second,
					[]trace.Event{
						expectEvent(anyHost, "who", testutils.CPUForTests, anyPID, 0, events.SchedProcessExit, orPolNames("xxx"), orPolIDs(1)),
					},
					[]string{},
					false,
				),
				newCmdEvents(
					"ping -c1 0.0.0.0",
					0,
					1*time.Second,
					[]trace.Event{
						expectEvent(anyHost, "ping", testutils.CPUForTests, anyPID, 0, events.SchedProcessExec, orPolNames("xxx"), orPolIDs(1)),
						expectEvent(anyHost, "ping", testutils.CPUForTests, anyPID, 0, events.MyDevelopingEvent, orPolNames("xxx"), orPolIDs(1)),
					},
					[]string{},
					true,
				),
			},
			useSyscaller: false,
			coolDown:     0,
			test:         ExpectAllInOrderSequentially,
		},

```
[...]

PR #4305, opened 2024-09-17
fix(tests): possible out of range in integration

geyslan:
Close: #4255

5cf2d9d1b **fix(tests): use ExpectAtLeastOneForEach**

```
Two test cases were using ExpectAllInOrderSequentially helper function
and passing by luck, since they emit more events than expected only
beyond the expected events boundary.

For that cases we should use ExpectAtLeastOneForEach helper function
instead.
```

34be7e038 **fix(tests): possible out of range in integration**

```
ExpectAllInOrderSequentially might try to access an index out of range
depending on the number of events that are being checked.

More about in issue #4255.
``` in issue #4255.
```

### DIFF
--- a/tests/integration/event_filters_test.go
+++ b/tests/integration/event_filters_test.go
@@ -771,7 +771,7 @@ func Test_EventFilters(t *testing.T) {
 			},
 			useSyscaller: false,
 			coolDown:     0,
-			test:         ExpectAllInOrderSequentially,
+			test:         ExpectAtLeastOneForEach,
 		},
 		{
 			name: "pid: trace new (should be empty)",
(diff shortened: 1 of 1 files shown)


tests/integration/event_filters_test.go
```diff
@@ -2469,17 +2469,19 @@ func ExpectAllInOrderSequentially(t *testing.T, cmdEvents []cmdEvents, actual *e
...
 		if checkSets {
 			syscallsInSets = getAllSyscallsInSets(cmd.sets)
 		}
 
 		// compare the expected events with the actual events in the same order
-		for evtIdx, expEvt := range cmd.expectedEvents {
-			actEvt := actEvtsCopy[cmdIdx*len(cmd.expectedEvents)+evtIdx]
+		for _, expEvt := range cmd.expectedEvents {
+			actEvt := actEvtsCopy[actEvtIdx]
+			actEvtIdx++
```
NDStrahilevitz, line 2484: So this will go from 0 to some n, and then from n to n+m, and so on... Is this the intended behavior?

geyslan: Yep, as events must be found in order. The index before `cmdIdx*len(cmd.expectedEvents)+evtIdx` could cause an out of bounds access for special events like the created by https://github.com/aquasecurity/tracee/issues/4255#issue-2471922475.

But after your question I realised that it still can cause it. Gonna double check that.

geyslan: Done.

Issue #3872, opened 2024-02-20
Data race in ProcessTree


subkin13:
Could be a write and read to pt.procfsOnce.
Solution: move ```if pt.procfsOnce == nil { ``` before ```if pt.procfsChan == nil { ```

```go
func (pt *ProcessTree) feedFromProcFSLoop() {
	go func() {
		for {
			select {
			case <-time.After(15 * time.Second):
				pt.procfsOnce = new(sync.Once) // reset the once every 15 seconds
			case <-pt.ctx.Done():
				return
			case givenPid := <-pt.procfsChan:
				err := pt.FeedFromProcFS(givenPid)
				if err != nil && debugMsgs {
					logger.Debugw("proctree from procfs (loop)", "err", err)
				}
			}
		}
	}()
}

// FeedFromProcFSAsync feeds the process tree with data from procfs asynchronously.
func (pt *ProcessTree) FeedFromProcFSAsync(givenPid int) {
	if pt.procfsChan == nil {
		logger.Debugw("starting procfs proctree loop") // will tell if called more than once
		pt.procfsChan = make(chan int, 1000)
		pt.feedFromProcFSLoop()
	}
	if pt.procfsOnce == nil {
		pt.procfsOnce = new(sync.Once)
	}
```

geyslan:
> Could be a write and read to pt.procfsOnce.

Thanks @subkin13. It really seems to be the case.

Perhaps a mutex control only for the `procfsOnce`? @NDStrahilevitz WDYT?

NDStrahilevitz:
@geyslan Not so familiar with this code i'm afraid. @AlonZivony Would be more relevant here.
Anyway, from a glance at this, it seems like this procfsOnce should be initialized on construction, then no nil check is required. If there is an additional condition to trigger procfsOnce, it should be explicit and perhaps concurrency controlled if needed.

AlonZivony:
@geyslan can you elaborate on the issue here?
The problem is read and write in the same time to the `procfsOnce ` variable?

AlonZivony:
@geyslan can you elaborate on the issue here?

geyslan:
> @geyslan can you elaborate on the issue here?

The goroutine triggered by `feedFromProcFSLoop()` might timeout, right? If so it could be writing to `pt.procfsOnce` while it's read in?

[...]

PR #4307, opened 2024-09-17
fix(proctree): possible sync.Once data race

geyslan:
Close: #3872 

97dc77af6 **fix(proctree): possible sync.Once data race**


NDStrahilevitz (approved):
Makes much more sense

PR #4308, opened 2024-09-17
chore: remove deprecated debug-shell

geyslan:

b6e07d3e0 **chore: remove deprecated debug-shell**


geyslan:
We would need runners with public IPs. As far as I know, it's not possible.

NDStrahilevitz (approved):

PR #4310, opened 2024-09-18
sched_process_exec: don't drop event in capture exec

OriGlassman:
Currently, as part of capturing executed files, tracee tries to copy the file. In case the copy fails, tracee drops the event. This patch changes the behavior to log a debug message and continue the normal flow.

sched_process_exec: don't drop event in capture exec

### 2. Explain how to test it
./tracee -e=sched_process_exec -c exec


geyslan (approved):
LGTM.

I believe it's the simplest workaround for now.

Tks, @OriGlassman.

PR #4312, opened 2024-09-18
chore: make analyze mode work with sigs of sigs

geyslan:
`analyze` mode wasn't aware of signature's selected events working as signatures. Now it also loads all of the selected.

ef6441f57 **chore: make analyze mode work with sigs of sigs**

### DIFF
--- a/pkg/signatures/signature/signature.go
+++ b/pkg/signatures/signature/signature.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	embedded "github.com/aquasecurity/tracee"
+	"github.com/aquasecurity/tracee/pkg/events"
 	"github.com/aquasecurity/tracee/pkg/logger"
 	"github.com/aquasecurity/tracee/pkg/signatures/regosig"
 	"github.com/aquasecurity/tracee/types/detect"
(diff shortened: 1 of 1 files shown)


geyslan:
Tested with current signatures (which depend only on base events) and still works.

Be aware that in case selected are sigs, the input must include events from all base events that are selected by the selected signatures themselves.

NDStrahilevitz:
@geyslan From my end, I feel like I'm missing context on how this solved the issue, and what didn't work in the first place. Because on its face i'd assume that:
1. signatures are converted to event definitions
2. a sig-on-sig selector filters for these event names created in step 1
3. therefore signatures selecting for signatures should work

So i'm not sure what in these steps was faulty and how this code fixes it.

geyslan:
@NDStrahilevitz sorry, this should be closed already, it was only raised to help the team on debugging. There's another PR that continued this. https://github.com/aquasecurity/tracee/pull/4314

geyslan:
Anyway, this piece only loaded the all other required sigs of a sig that declares them as Selected. Without it tracee warned about the missing sigs.

PR #4314, opened 2024-09-19
[DRAFT] Sig on sigs

AsafEitani:

### DIFF
--- a/cmd/tracee-rules/main.go
+++ b/cmd/tracee-rules/main.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/aquasecurity/tracee/pkg/cmd/initialize/initialize_sigs"
+	"github.com/aquasecurity/tracee/pkg/events"
 	"io"
 	"os"
 	"os/signal"
@@ -26,6 +28,8 @@ import (
 const (
 	signatureBufferFlag = "sig-buffer"
 )
+
+var inputs engine.EventSources
 
 func main() {
 	app := &cli.App{
@@ -116,8 +120,6 @@ func main() {
 				return nil
 			}
 
-			var inputs engine.EventSources
-
 			opts, err := parseTraceeInputOptions(c.StringSlice("input-tracee"))
 			if err == errHelp {
 				printHelp()
@@ -142,7 +144,7 @@ func main() {
 			if err != nil {
 				return err
 			}
-
+			_ = initialize_sigs.CreateEventsFromSignatures(events.StartSignatureID, sigs)
 			config := engine.Config{
 				SignatureBufferSize: c.Uint(signatureBufferFlag),
 				Signatures:          sigs,
@@ -295,13 +297,13 @@ func listEvents(w io.Writer, sigs []detect.Signature) {
 		}
 	}
 
-	var events []string
+	var e []string
 	for k := range m {
-		events = append(events, k)
+		e = append(e, k)
 	}
 
-	sort.Slice(events, func(i, j int) bool { return events[i] < events[j] })
-	fmt.Fprintln(w, strings.Join(events, ","))
+	sort.Slice(e, func(i, j int) bool { return e[i] < e[j] })
+	fmt.Fprintln(w, strings.Join(e, ","))
 }
 
 func isRoot() bool {
--- a/cmd/tracee-rules/main.go
+++ b/cmd/tracee-rules/main.go
@@ -4,8 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/aquasecurity/tracee/pkg/cmd/initialize/initialize_sigs"
-	"github.com/aquasecurity/tracee/pkg/events"
 	"io"
 	"os"
 	"os/signal"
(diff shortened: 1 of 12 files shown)


geyslan (changes requested):
For a simple anti_debugging signature, tracee-ebpf and analyze continues to work, but I can't say for the specific signature of signatures... I don't have a proper input file.

pkg/signatures/signature/signature.go
```diff
@@ -50,16 +51,88 @@ func Find(target string, partialEval bool, signaturesDir []string, signatures []
...
+		processedSigs[sigMeta.EventName] = struct{}{}
+
+		// add the current signature to the result list
+		sigsOfsigs = append(sigsOfsigs, sig)
+
+		// get selected signatures from this sig and process them
+		selected := getSelectedSigsFromSig(loadedSigMap, sig)
+		for sel := range selected {
+			if selSig, ok := loadedSigMap[sel]; ok {
+				processSigsOfSig(selSig) // recursively process the found signature
```
geyslan, line 97: If !ok, emit warning of a missing signature.

cmd/tracee-rules/main.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/aquasecurity/tracee/pkg/cmd/initialize"
```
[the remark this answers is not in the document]
AsafEitani: resolved by decoupling the sigs from the `initialize` package

cmd/tracee-rules/main.go
```diff
@@ -142,7 +144,7 @@ func main() {
 			if err != nil {
 				return err
 			}
-
+			_ = initialize.CreateEventsFromSignatures(events.StartSignatureID, sigs)
```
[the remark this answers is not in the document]
AsafEitani: resolved by decoupling the sigs from the `initialize` package

AsafEitani:
@geyslan Hi, need your assistance here.
1. Lint in failing for code I didn't change. What do you want me to rename it to?
2. How do we add filters for the signatures that are sent back to the input channel? Currently all findings are sent to all the signatures as events. We should filter that so only the selected events are sent.

cmd/tracee-rules/main.go
```diff
@@ -27,6 +29,8 @@ const (
 	signatureBufferFlag = "sig-buffer"
 )
 
+var inputs engine.EventSources
```
NDStrahilevitz, line 32: What's the purpose of moving this?

cmd/tracee-rules/main.go
```diff
@@ -142,7 +144,7 @@ func main() {
 			if err != nil {
 				return err
 			}
-
+			_ = initialize_sigs.CreateEventsFromSignatures(events.StartSignatureID, sigs)
```
NDStrahilevitz, line 147: Why is there no error check?

cmd/tracee-rules/output.go
```diff
@@ -57,8 +57,17 @@ func setupOutput(w io.Writer, webhook string, webhookTemplate string, contentTyp
...
+			switch e := res.Event.Payload.(type) {
 			case trace.Event:
+				select {
+				case _, ok := <-inputs.Tracee:
+					if !ok {
+						logger.Debugw("Tracee input channel closed")
+						return
+					}
+				default:
+					inputs.Tracee <- e.ToProtocol()
```
NDStrahilevitz, line 69: So the `inputs` from the previous file is now a global variable for this? What's the reasoning?

NDStrahilevitz (changes requested):
Ok, if I understand it right from the code, the issue was that just as events have event dependencies which we need to ensure are loaded even if the users does not select them, so goes the case now for signatures.

Now, since I know the internal usecase for this, If I got the code right then I'm not sure that it is actually the desired behavior. The desired usecase was for a signature aggregating data on the behavior of all other signatures with certain characteristics. But this does not imply that we would want to load each such signature. Rather, it implies that it would aggregate this info based on the selected signatures, which sounds about right.

Now, I might be wrong about what the code does. Whether this is the case or not, it lacks documentation on:
1. What is the resolved gap in the code (the code is meant to handle signatures as dependent on others, it didn't work because of X)
2. How the gap was resolved (Signatures which are dependencies are now selected for emission - previously they were only loaded)
3. More concrete documentation in the code explaining the logic, the word "processing" for example was used as a generic descriptor for a concrete action - finding dependent signatures which are available in the plugin but not selected for emitting.

Finally, please fix the style so that the tests pass.

AsafEitani:
@NDStrahilevitz So the purpose of the PR is to allow signatures to process other signatures.
The original issue was that the sigs on sigs works on the one binary, but not in `analyze` mode or `tracee-rules`, this PR fixes that.
This includes the event selection process as well as feeding the signatures as events back to the input channel, mainly the latter.
Most of the code here was either written by @geyslan or copied from other places in Tracee.

cmd/tracee-rules/main.go
AsafEitani: feeding the signatures as events back to the input channel.

cmd/tracee-rules/main.go
AsafEitani: copied from the one binary mode.

cmd/tracee-rules/output.go
AsafEitani: feeding back the signatures.

geyslan:
I'm debugging internal integration (bump) and faced this commit, it seems related to this current effort: https://github.com/aquasecurity/tracee/pull/3681/commits

NDStrahilevitz:
Turned to draft in favor of #4327

NDStrahilevitz:
Closing in favor of #4327.

Issue #4313, opened 2024-09-19
fatal error: concurrent map iteration and map write


JoeyRedfield:
## Description
I execute the command `sudo docker run --name tracee -it --rm   --pid=host --cgroupns=host --privileged   -v /etc/os-release:/etc/os-release-host:ro   -v /var/run:/var/run:ro   -v /boot:/boot:ro   -v /lib/modules:/lib/modules:ro   aquasec/tracee:latest   -o json`, an error occurred after running for a period of time: 

```bash
fatal error: concurrent map iteration and map write

goroutine 37 [running]:
reflect.mapiternext(0xdc1669?)
	/tracee/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.4.linux-amd64/src/runtime/map.go:1392 +0x13
reflect.(*MapIter).Next(0xc0032cd328?)
	/tracee/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.4.linux-amd64/src/reflect/value.go:2005 +0x74
encoding/json.mapEncoder.encode({0x4?}, 0xc00317e200, {0x2222fa0?, 0xc0006d5b78?, 0x2155ee0?}, {0xd?, 0x0?})
	/tracee/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.4.linux-amd64/src/encoding/json/encode.go:745 +0x334
encoding/json.structEncoder.encode({{{0xc000174fc8, 0x4, 0x4}, 0xc000f128d0, 0xc000f12900}}, 0xc00317e200, {0x23a3dc0?, 0xc0006d5b40?, 0xc0022fe948?}, {0x0, ...})
	/tracee/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.4.linux-amd64/src/encoding/json/encode.go:704 +0x21e
encoding/json.ptrEncoder.encode({0xc0022fe948?}, 0xc00317e200, {0x2102ca0?, 0xc0022fe948?, 0xc0022fe948?}, {0xb?, 0x0?})
	/tracee/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.4.linux-amd64/src/encoding/json/encode.go:876 +0x23c
encoding/json.structEncoder.encode({{{0xc000620008, 0x20, 0x23}, 0xc000f12990, 0xc000f129c0}}, 0xc00317e200, {0x24be340?, 0xc0022fe780?, 0xc0032cd7d0?}, {0x0, ...})
[...]

Issue #4298, opened 2024-09-11
Signatures escape same Metadata.Properties map field to heap


geyslan:
## Description

This can cause a data race as detected in: https://github.com/aquasecurity/tracee/actions/runs/10814517981/job/30001186085?pr=4252#step:6:592

```
INFO: 
INFO: = TEST: TRC-1010 =================================================
INFO: 
INFO: 
INFO: UP AND RUNNING
INFO:

...

goroutine 20 [running]:
reflect.mapiternext(0x2104360?)
	/tmp/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.4.linux-arm64/src/runtime/map.go:1392 +0x1c
reflect.(*MapIter).Next(0x4000057300?)
	/tmp/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.4.linux-arm64/src/reflect/value.go:2005 +0x78
encoding/json.mapEncoder.encode({0x4000057468?}, 0x4000d1fd80, {0x2145200?, 0x4000d1fd78?, 0x20782c0?}, {0xd?, 0x0?})
	/tmp/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.4.linux-arm64/src/encoding/json/encode.go:745 +0x28c
encoding/json.structEncoder.encode({{{0x40004e3d48, 0x4, 0x4}, 0x4002fcad80, 0x4002fcadb0}}, 0x4000d1fd80, {0x22c5d00?, 0x4000d1fd40?, 0x96?}, {0xa8?, ...})
	/tmp/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.4.linux-arm64/src/encoding/json/encode.go:704 +0x190
encoding/json.ptrEncoder.encode({0x4000057688?}, 0x4000d1fd80, {0x2025240?, 0x4000c803a8?, 0x4000d1fd80?}, {0xb?, 0x0?})
	/tmp/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.4.linux-arm64/src/encoding/json/encode.go:876 +0x1ec
encoding/json.structEncoder.encode({{{0x4000392c08, 0x20, 0x23}, 0x4002fcae40, 0x4002fcae70}}, 0x4000d1fd80, {0x23e0000?, 0x4000c801e0?, 0x4000057818?}, {0xfc?, ...})
[...]

PR #4316, opened 2024-09-20
fix: clone sig metadata properties

rscampos:
fix: #4298
fix: #4313 

0b164b1a1 **fix: clone sig metadata properties**

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/Masterminds/sprig/v3 v3.2.3
 	github.com/aquasecurity/libbpfgo v0.7.0-libbpf-1.4.0.20240729111821-61d531acf4ca
 	github.com/aquasecurity/tracee/api v0.0.0-20240905132323-d1eaeef6a19f
-	github.com/aquasecurity/tracee/signatures/helpers v0.0.0-20240607205742-90c301111aee
+	github.com/aquasecurity/tracee/signatures/helpers v0.0.0-20240920144223-9d62cbdd8935
 	github.com/aquasecurity/tracee/types v0.0.0-20240607205742-90c301111aee
 	github.com/containerd/containerd v1.7.21
 	github.com/docker/docker v26.1.5+incompatible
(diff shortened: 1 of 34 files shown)


go.mod
```diff
@@ -43,6 +43,8 @@ require (
 	sigs.k8s.io/controller-runtime v0.18.2
 )
 
+replace github.com/aquasecurity/tracee/signatures/helpers => ./signatures/helpers
+
```
geyslan, line 47: Should be removed right?

signatures/helpers/helpers.go
```diff
@@ -435,3 +437,9 @@ func GetProtoHTTPByName(
 
 	return trace.ProtoHTTP{}, fmt.Errorf("protocol HTTP: type error (should be trace.ProtoHTTP, is %T)", arg.Value)
 }
+
+func CloneProperties(m *detect.SignatureMetadata) {
+	// do a shallow clone of Properties map getting a new reference
+	// avoiding leaking the original pointer
```
geyslan, line 443:
```suggestion
// CloneMetadataProperties makes a shallow clone of the Metadata
// Properties map, ensuring that there is always a new reference to it.
//
// TODO: since this helper is a workaround to avoid data races,
// perhaps a better solution would be to convert Properties into
// a concrete structure.
func CloneMetadataProperties(m *detect.SignatureMetadata) {
```

WDYT of an outer comment and rename the helper?

geyslan (approved):
LGTM. Put some wondering.

geyslan:
This issue's impacting latest docker images. So we should release a new minor after merging and testing.

go.mod
rscampos: Yeah, will remove it.

signatures/helpers/helpers.go
rscampos: Its better! tks!

geyslan (approved):
LGTM

NDStrahilevitz:
Forgive me for asking, but are we any better off now than the original version where the function just returned the metadata as a struct? So maybe we have now (at best) (mem footprint)=original-(no copy)+(clone the properties). 
But we lost ease of authoring and readability which are in the context of signatures more important IMO.

rscampos:
I think you’re right, @NDStrahilevitz. Initially, we saw a performance improvement, but then an issue arose. After discussing it with @geyslan, we’re considering a rollback. Thanks!

geyslan:
We're already reverted related changes.

PR #4324, opened 2024-09-24
Data filter in kernel

rscampos:

03b6b4dd5 **docs(filters): add restrictions when kernel data filter is used**
97dd25f1f **test(filters): kernel data filter**
40305940e **feat(filters): equalities for kernel data filter**
0d9825254 **feat(filters): eBPF management for kernel data filter**
0c31dbe07 **feat(ebpf): enable kernel data filter in eBPF program**
86ac860da **feat(ebpf): kernel data filter logic**
a499bc1e3 **feat(ebpf): Enable BPF_F_NO_PREALLOC for LPM TRIE**

03b6b4dd5 **docs(filters): add restrictions when kernel data filter is used**

```
- Add the restrictions applicable when the kernel-space data filter is
available for an event field.
```

97dd25f1f **test(filters): kernel data filter**

```
- Add MatchTypes{} and KernelDataFilter{} in cmp.AllowUnexported;
- Kernel data filters restrict pathnames to 255 characters and
disallow 'contains' filters; unit tests have been added to validate
these restrictions;
- Integration tests for specific events added, covering three filter
types with "equal" and "not equal" conditions.
```

40305940e **feat(filters): equalities for kernel data filter**

```
- method equalities created for data filter;
- method computeDataFilterEqualities created for kernel data filter;
- handle corner case when one policy uses a substring (path) of another
  policy;
- disable data filter (only pathname) for selected events;
- Kernel data filters restrict pathnames to 255 characters and disallow
'contains' filters - added functions to enforce that.
```

0d9825254 **feat(filters): eBPF management for kernel data filter**

```
- eBPF map definition for exact, prefix, suffix match;
- create updateDataFilterLPMBPF and updateDataFilterBPF to populate eBPF
  maps;
- config map fields for exact, prefix and suffix;
- Create the function createNewDataFilterMapsVersion in order to create
the inner maps based on version and event id.
```

0c31dbe07 **feat(ebpf): enable kernel data filter in eBPF program**

```
- how to enable data filter in the eBPF program using the function
evaluate_data_filters.
```

86ac860da **feat(ebpf): kernel data filter logic**

```
- function load_str_from_buf created to retrieve str value based on index;
- function reverse_string created to revert an string in order to enable suffix;
- function evaluate_data_filters/match_data_filters created to apply: exact, prefix and suffix match;
- eBPF maps for exact, prefix and suffix. eBPF map for hold temporary LPM TRI key;
- extend event_config to have data filter config per event: used for exact, prefix and suffix match;
- save offset at the specified index in the function save_str_to_buf.
```

### 2. Explain how to test it

The method for defining data filters in Tracee remains the same. However, for the `security_file_open` and `magic_write` events, if the pathname is used as a filter, the event is now filtered at the eBPF data plane, preventing it from being sent to user space for filtering.

**Notes for the reviewer:** The following sections contain commands I used to test with policies. The results for each test group are also included. Both the policies and results are located in the zip file provided in each section. The results of some tests may vary depending on the Linux version and libraries, especially when the "not equal" operator is used.

- [Only exact match](#only-exact-match)
- [Only prefix match](#only-prefix-match)
- [Only suffix match](#only-suffix-match)
- [Mixed (exact/prefix/suffix) match](#mixed-match)
- [Mixed (exact/prefix/suffix) match (same policy)](#mixed-match-same-policy)
- [Ensuring Multiple Policy Matches when LPM Trie is used - Prefix](#corner-case-prefix-match)
- [Ensuring Multiple Policy Matches when LPM Trie is used - Suffix](#corner-case-suffix-match)
- [Performance results](#performance)

<h2 id="only-exact-match">Only exact match</h2>

### Maps
```
% sudo bpftool map dump name config_map
...
"enabled_policies": 7,
"data_filter_prefix_enabled": 0,
"data_filter_suffix_enabled": 0,
"data_filter_exact_enabled": 7,
"data_filter_prefix_match_if_key_missing": 0,
"data_filter_suffix_match_if_key_missing": 0,
"data_filter_exact_match_if_key_missing": 4,
...
```
```
## dump data_filter_exact
% sudo bpftool map dump id 24016
[{
        "key": {
            "path": "/etc/networks"
        },
        "value": {
            "equal_in_scopes": 2,
            "equality_set_in_scopes": 6
        }
    },{
        "key": {
            "path": "/etc/netconfig"
        },
        "value": {
            "equal_in_scopes": 1,
            "equality_set_in_scopes": 1
        }
    }
]
```
### Cmds
The results of each of the following lines are in the JSON file (results_exact.json):
> % more /etc/netconfig  # json line 1 (sfo-exact-match-1)
> % more /etc/networks  # json line 2 (sfo-exact-match-2)
> % cat /etc/networks     # json line 3,4,5 (sfo-exact-match-3)

[exactly_policies_results.zip](https://github.com/user-attachments/files/17319737/exactly_policies_results.zip)

<h2 id="only-prefix-match">Only prefix match</h2>

### Maps
```
% sudo bpftool map dump name config_map
...
"enabled_policies": 7,
"data_filter_prefix_enabled": 7,
"data_filter_suffix_enabled": 0,
"data_filter_exact_enabled": 0,
"data_filter_prefix_match_if_key_missing": 4,
"data_filter_suffix_match_if_key_missing": 0,
"data_filter_exact_match_if_key_missing": 0,
...
```
```
## dump data_filter_prefix
% sudo bpftool map dump id 24583
[{
        "key": {
            "prefix_len": 96,
            "path": "/etc/network"
        },
        "value": {
            "equal_in_scopes": 1,
            "equality_set_in_scopes": 5
        }
    },{
        "key": {
            "prefix_len": 72,
            "path": "/etc/pass"
        },
        "value": {
            "equal_in_scopes": 2,
            "equality_set_in_scopes": 2
        }
    }
]
```
### Cmds
The results of each of the following lines are in the JSON file (results_prefix.json):
> % more /etc/networks # json line 1 (sfo-prefix-match-1)
> % sudo cp /etc/networks /etc/networks.bkp; more /etc/networks.bkp # json line 2 (sfo-prefix-match-1)
> % more /etc/passwd # json line 3 (sfo-prefix-match-2)
> % cat /etc/networks # json line 4,5,6 (sfo-prefix-match-3)

[prefix_policies_results.zip](https://github.com/user-attachments/files/17319873/prefix_policies_results.zip)

<h2 id="only-suffix-match">Only suffix match</h2>

### Maps
```
% sudo bpftool map dump name config_map
...
"enabled_policies": 7,
"data_filter_prefix_enabled": 0,
"data_filter_suffix_enabled": 7,
"data_filter_exact_enabled": 0,
"data_filter_prefix_match_if_key_missing": 0,
"data_filter_suffix_match_if_key_missing": 4,
"data_filter_exact_match_if_key_missing": 0,
...
```
```
## dump data_filter_suffix
% sudo bpftool map dump id 24583
[{
        "key": {
            "prefix_len": 48,
            "path": "dwssap"
        },
        "value": {
            "equal_in_scopes": 2,
            "equality_set_in_scopes": 2
        }
    },{
        "key": {
            "prefix_len": 72,
            "path": "gifnocten"
        },
        "value": {
            "equal_in_scopes": 1,
            "equality_set_in_scopes": 5
        }
    }
]
```
### Cmds
The results of each of the following lines are in the JSON file (results_suffix.json):

> % more /etc/netconfig # json line 1 (sfo-suffix-match-1)
> % cp /etc/netconfig /tmp/netconfig; more /tmp/netconfig # json line 2 (sfo-suffix-match-1)
> % more /etc/passwd # json line 3 (sfo-suffix-match-2)
> % cat /etc/netconfig # json line 4,5,6 (sfo-suffix-match-3)

[suffix_policies_results.zip](https://github.com/user-attachments/files/17319846/suffix_policies_results.zip)

<h2 id="mixed-match">Mixed (exact/prefix/suffix) match</h2>

In this section, you can see all string matches working together. The command `cat /etc/netconfig` triggers three policies simultaneously, while the command `cat /etc/host.conf` triggers two policies.

### Maps
```
% sudo bpftool map dump name config_map
...
"enabled_policies": 31,
"data_filter_prefix_enabled": 18,
"data_filter_suffix_enabled": 12,
"data_filter_exact_enabled": 1,
"data_filter_prefix_match_if_key_missing": 0,
"data_filter_suffix_match_if_key_missing": 0,
"data_filter_exact_match_if_key_missing": 0,
...
```
```
## dump exact
[{
        "key": {
            "path": "/etc/netconfig"
        },
        "value": {
            "equal_in_scopes": 1,
            "equality_set_in_scopes": 1
        }
    }
]

## dump prefix
[{
        "key": {
            "prefix_len": 72,
            "path": "/etc/host"
        },
        "value": {
            "equal_in_scopes": 16,
            "equality_set_in_scopes": 16
        }
    },{
        "key": {
            "prefix_len": 64,
            "path": "/etc/net"
        },
        "value": {
            "equal_in_scopes": 2,
            "equality_set_in_scopes": 2
        }
    }
]

## dump suffix
[{
        "key": {
            "prefix_len": 40,
            "path": "fnoc."
        },
        "value": {
            "equal_in_scopes": 8,
            "equality_set_in_scopes": 8
        }
    },{
        "key": {
            "prefix_len": 72,
            "path": "gifnocten"
        },
        "value": {
            "equal_in_scopes": 4,
            "equality_set_in_scopes": 4
        }
    }
]
```
### Cmds
The results of each of the following lines are in the JSON file (results_mixed.json):

> % cat /etc/network/fan # json line 1 (sfo-prefix-match-4)
> % cp /etc/netconfig /tmp/netconfig; cat /tmp/netconfig # json line 2 (sfo-suffix-match-4)
> % cat /etc/netconfig # json line 3 (sfo-exact-match-5,sfo-prefix-match-4,sfo-suffix-match-4)
> % cat /etc/host.conf # json line 4 (sfo-suffix-match-5,sfo-prefix-match-5)
> % cat /etc/sysctl.conf # json line 5 (sfo-suffix-match-5)

[mixed_policies_results.zip](https://github.com/user-attachments/files/17328950/mixed_policies_results.zip)

<h2 id="mixed-match-same-policy">Mixed (exact/prefix/suffix) match (same policy)</h2>

In this section, you can see all string matches working together in the same policy. The command `more /etc/netconfig` triggers three policies simultaneously.

### Maps
```
% sudo bpftool map dump name config_map
...
"enabled_policies": 7,
"data_filter_prefix_enabled": 3,
"data_filter_suffix_enabled": 6,
"data_filter_exact_enabled": 7,
"data_filter_prefix_match_if_key_missing": 3,
"data_filter_suffix_match_if_key_missing": 4,
"data_filter_exact_match_if_key_missing": 6,
...
```
```
## dump exact
[{
        "key": {
            "str": "/etc/ld.so.cache"
        },
        "value": {
            "equals_in_policies": 1,
            "key_used_in_policies": 5
        }
    },{
        "key": {
            "str": "/etc/netconfig"
        },
        "value": {
            "equals_in_policies": 0,
            "key_used_in_policies": 2
        }
    }
]

## dump prefix
[{
        "key": {
            "prefix_len": 64,
            "str": "/etc/net"
        },
        "value": {
            "equals_in_policies": 1,
            "key_used_in_policies": 1
        }
    },{
        "key": {
            "prefix_len": 72,
            "str": "/usr/lib/"
        },
        "value": {
            "equals_in_policies": 0,
            "key_used_in_policies": 3
        }
    }
]

## dump suffix
[{
        "key": {
            "prefix_len": 120,
            "str": "3.6.os.ofnitbil"
        },
        "value": {
            "equals_in_policies": 0,
            "key_used_in_policies": 4
        }
    },{
        "key": {
            "prefix_len": 72,
            "str": "6.os.cbil"
        },
        "value": {
            "equals_in_policies": 0,
            "key_used_in_policies": 4
        }
    },{
        "key": {
            "prefix_len": 88,
            "str": "ehcac.os.dl"
        },
        "value": {
            "equals_in_policies": 2,
            "key_used_in_policies": 2
        }
    }
]
```
### Cmds
The results of each of the following lines are in the JSON file (results_mixed_same_policy.json):

> % more /etc/netconfig

Results: results_mixed_same_policy.json
> json line 1 (sfo-mixed-pol-1, sfo-mixed-pol-2)
> json line 2 (sfo-mixed-pol-3)
> json line 3 (sfo-mixed-pol-3)
> json line 4 (sfo-mixed-pol-1, sfo-mixed-pol-3)
> json line 5 (sfo-mixed-pol-3)

[results_mixed_same_policy_results.zip](https://github.com/user-attachments/files/18009193/results_mixed_same_policy_results.zip)

<h2 id="corner-case-prefix-match">Ensuring Multiple Policy Matches when LPM Trie is used - Prefix</h2>

**Corner case description:** When using the LPM Trie, it always returns the longest matching string. A corner case arises when one policy (e.g., **policy1**) uses a substring of another policy (e.g., **policy2**). For instance, if **policy1** covers `/etc/net*` and **policy2** covers `/etc/netconf*`, a lookup for `/etc/netconfig` currently only returns **policy2** because it is the longest match. However, it should return both **policy1** and **policy2**.

**A potential solution (implemented):** If one suffix or prefix overlaps with another, we can simply combine their bitmaps in user space. No additional logic is required in kernel space to handle this corner case.

### Maps
```
% sudo bpftool map dump name config_map
...
"enabled_policies": 15,
"data_filter_prefix_enabled": 15,
"data_filter_suffix_enabled": 0,
"data_filter_exact_enabled": 0,
"data_filter_prefix_match_if_key_missing": 8,
"data_filter_suffix_match_if_key_missing": 0,
"data_filter_exact_match_if_key_missing": 0,
...
```
```
## dump prefix
[{
        "key": {
            "prefix_len": 96,
            "path": "/etc/netconf"
        },
        "value": {
            "equal_in_scopes": 3,
            "equality_set_in_scopes": 11
        }
    },{
        "key": {
            "prefix_len": 96,
            "path": "/etc/network"
        },
        "value": {
            "equal_in_scopes": 7,
            "equality_set_in_scopes": 7
        }
    },{
        "key": {
            "prefix_len": 64,
            "path": "/etc/net"
        },
        "value": {
            "equal_in_scopes": 3,
            "equality_set_in_scopes": 3
        }
    },{
        "key": {
            "prefix_len": 48,
            "path": "/etc/n"
        },
        "value": {
            "equal_in_scopes": 1,
            "equality_set_in_scopes": 1
        }
    },{
        "key": {
            "prefix_len": 72,
            "path": "/usr/lib/"
        },
        "value": {
            "equal_in_scopes": 0,
            "equality_set_in_scopes": 8
        }
    }
]
```
_Note: Policy 4 includes lines to exclude library entries from the output. These lines are solely for cleaning up the output to simplify the testing in this section._

**Policy 2** (prefix _/etc/net_) overlaps with **Policy 1** (prefix _/etc/n_), as _/etc/n_ is a substring of _/etc/net_. This is why the key with the path **"/etc/net"** has **equality_set_in_scopes=3**, indicating that both **Policy 1** and **Policy 2** are part of the same equality set. Additionally, **equal_in_scopes=3** shows that **Policy 1** and **Policy 2** are considered equal in their scopes.

**Policy 3** (prefix _/etc/network_) encompasses both **Policy 1** (prefix _/etc/n_) and **Policy 2** (prefix _/etc/net_). Consequently, the key with the path "/etc/network" has **equality_set_in_scopes=7**, which signifies that all three policies are present within the same scope. Similarly, **equal_in_scopes=7** indicates that **Policy 1**, **Policy** 2, and **Policy 3** are equal in scopes.

**Policy 4** (prefix _/etc/netconf_) also includes both **Policy 1** (prefix _/etc/n_) and **Policy 2** (prefix _/etc/net_). Therefore, the key with the path **"/etc/netconf"** has **equality_set_in_scopes=11**, which means that **Policy 1**, **Policy 2**, and **Policy 4** are all part of the same scope. However, because **Policy 4** was defined with the condition **data.pathname!=/etc/netconf**, **equal_in_scopes=3**, meaning that only **Policy 1** and **Policy 2** are considered equal in scopes, while **Policy 4** is excluded from that equality.

In summary, **Policy 2**, **Policy 3**, and **Policy 4** derive bits from other policies, reflecting their interdependencies and overlaps in scope.

### Cmds
The results of each of the following lines are in the JSON file (results_corner_case_prefix.json):

> % more /etc/netconfig; json line 1 (cc-sfo-prefix-match-1; cc-sfo-prefix-match-2)
> % more /etc/networks; json line 2 (cc-sfo-prefix-match-2; cc-sfo-prefix-match-3; cc-sfo-prefix-match-4; cc-sfo-prefix-match-1)
> % sudo cp /etc/netconfig /etc/na; more /etc/na; json line 3 (cc-sfo-prefix-match-4; cc-sfo-prefix-match-1)

[cc_prefix_policies_results.zip](https://github.com/user-attachments/files/17424048/cc_prefix_policies_results.zip)

<h2 id="corner-case-suffix-match">Ensuring Multiple Policy Matches when LPM Trie is used - Suffix</h2>

### Maps
```
% sudo bpftool map dump name config_map
...
"enabled_policies": 7,
"data_filter_prefix_enabled": 4,
"data_filter_suffix_enabled": 7,
"data_filter_exact_enabled": 4,
"data_filter_prefix_match_if_key_missing": 4,
"data_filter_suffix_match_if_key_missing": 4,
"data_filter_exact_match_if_key_missing": 4,
...
```
```
## dump suffix
[{
        "key": {
            "prefix_len": 104,
            "path": "gifnocten/cte"
        },
        "value": {
            "equal_in_scopes": 3,
            "equality_set_in_scopes": 7
        }
    },{
        "key": {
            "prefix_len": 72,
            "path": "gifnocten"
        },
        "value": {
            "equal_in_scopes": 3,
            "equality_set_in_scopes": 3
        }
    },{
        "key": {
            "prefix_len": 48,
            "path": "gifnoc"
        },
        "value": {
            "equal_in_scopes": 2,
            "equality_set_in_scopes": 2
        }
    }
]
```
_Note: Policy 3 includes lines to exclude library entries from the output. These lines are solely for cleaning up the output to simplify the testing in this section._

**Policy 1** (suffix _netconfig_) overlaps with **Policy 2** (suffix _config_), as _config_ is a substring of _netconfig_. This is why the key with the path **"gifnocten"** (which is a reversed representation of **netconfig**) has **equality_set_in_scopes=3** and this indicates that both **Policy 1** and **Policy 2** are contained within the same scope. However, **equal_in_scopes=3** shows that **Policy 1** and **Policy 2** are equal in scopes.

**Policy 2** (suffix _config_) only contains the bitmap of **Policy 2** (**equality_set_in_scopes=2** and **equal_in_scopes=2**).

**Policy 3** (suffix _etc/netconfig_) contains both **Policy 1** (suffix _netconfig_) and **Policy 2** (suffix _config_). Therefore, the key with the path **"gifnocten/cte"** (representing the reverse of **etc/netconfig**) has **equality_set_in_scopes=7**. This value indicates that all three policies are present within the scope. However, **equal_in_scopes=3** shows that only **Policy 1** and **Policy 2** are equal in scopes, whereas **Policy 3** is disabled in this scope because it was defined using **data.pathname!=etc/netconfig**.

### Cmds
The results of each of the following lines are in the JSON file (results_corner_case_suffix.json):

> % more /etc/netconfig; json line 1 (cc-sfo-suffix-match-1; cc-sfo-suffix-match-2)
> % more /etc/ssh/ssh_config; json line 2 (cc-sfo-suffix-match-2; cc-sfo-suffix-match-3)
> % cp /etc/netconfig /tmp/netaconfig; more /tmp/netaconfig; json line 3 (cc-sfo-suffix-match-2; cc-sfo-suffix-match-3)

[cc_suffix_policies_results.zip](https://github.com/user-attachments/files/17423150/cc_suffix_policies_results.zip)

- [Performance results](#performance)

<h2 id="performance">Performance results</h2>

To measure performance, the `bpftool` tool was used to collect and export `bpf_program_runtime_ns` and `bpf_program_amount` metrics to Grafana for analyzing eBPF program latency. Exporting this data to Grafana allows for measuring latency within a specific time window using the following logic:
> rate(bpf_program_runtime_ns[1m]) / rate(bpf_program_amount[1m])

**Improvements in eBPF Latency for Some Events (Compared to the Old Version of Tracee):**
**security_file_open**: ~80% improvement (both with and without stress).
**security_mmap_file**: 74% improvement without stress and 48% with stress.
**magic_write**: 54% improvement without stress and 0% with stress.

**Improvements in Lost Events:**
**security_file_open**: ~97.65% improvement.
**security_mmap_file**: ~95.04% improvement.
**magic_write**: ~99.40% improvement.

By applying filters in the eBPF plane, fewer events are placed in the perf buffer, which significantly reduces eBPF latency. The primary benefit of performing filtering in the kernel is the reduction in the number of events added to the perf buffer, leading to improved overall performance.

#### TODO
First Phase:
- [x]  Evaluate and integrate all three types of string-based filters simultaneously. Currently, they operate independently and need to be combined.
- [x] Complete the implementation and testing of filters for exact matches, prefix, and suffix. Including both equal and not equal for each of one these three filters.
- [x] In the function `save_str_to_buf()`, add the argument offset based on its index to facilitate direct access for the `load_str_from_buf()` function.
- [x] Document the testing steps for exact match, prefix, and suffix filters. Include policies and expected results for the review process.
- [x] When using the LPM Trie, it always returns the longest matching string. A corner case arises when one policy (e.g., **policy1**) uses a substring of another policy (e.g., **policy2**). For instance, if **policy1** covers `/etc/net*` and **policy2** covers `/etc/netconf*`, a lookup for `/etc/netconfig` currently only returns **policy2** because it is the longest match. However, it should return both **policy1** and **policy2**. A potential solution (work in progress) is to combine equality when such a corner case is detected.
- [x] Document the testing steps for corner case (prefix and suffix) filters. Include policies and expected results for the review process.
- [x] The filters for exact match, prefix, and suffix work with a total of 128 bytes (including the NULL terminator). Therefore, a maximum of 127 characters can be defined in the policy or provided via the CLI.
- [x] Improve the method for defining (in user-space) which events should have an in-kernel filter enabled. The current logic was added as a proof of concept and requires rework.
- [x] Measure performance between filter in user-space (older version) and filter in kernel-space (new version).
- [x] Create inner maps based on version and event id - in this way event id will be removed from the key.
- [x] Extend the filters to use 255 characters for exact, prefix and suffix. This is possible after removing event id from key, freeing 4 bytes.
- [x] Combine the bitmaps (using OR operation) when multiples filters are used in the same policy (match_data_filters function);
- [x] Need to add some validations/warning for event ids that has data kernel filter (e.g.: disable filter contains or 255 characters filter limit);
- [x] Need to add unit tests and integration tests;
- [x] Update docs.

Second Phase:

- The data kernel filter is currently enabled for `security_file_open`, `magic_write`, and `security_mmap_file`. Extend support to other events.
- Currently the index for retrieving the pathname in evaluate_data_filters is explicitly defined. While this works, it would be better to dynamically retrieve the index based on the event ID for greater flexibility.
- Currently only string filter is allowed. Design and implement a generic approach for other type of filters.


pkg/policy/ebpf.go
```diff
@@ -706,6 +829,7 @@ func (ps *policies) computePoliciesConfig() *PoliciesConfig {
 		}
 
 		cfg.EnabledScopes |= 1 << offset
+		cfg.EnabledDataFilters |= 1 << offset
```
geyslan, line 832: Shouldn't this be?

```suggestion
		cfg.EnabledDataFilters |= (ExactlyEnabledDataFilters | PrefixEnabledDataFilters | cfg.PostfixEnabledDataFilters) & (1 << offset)
```

pkg/filters/data.go
```diff
@@ -37,6 +116,17 @@ func (af *DataFilter) Filter(eventID events.ID, data []trace.Argument) bool {
 		return true
 	}
 
+	// don't Filter this special case
```
geyslan, line 119: What do you think of checking a `kernelFilter` flag instead (it could be set on `Parse`)?

pkg/filters/data.go
```diff
@@ -23,6 +30,78 @@ func NewDataFilter() *DataFilter {
...
+	// selected events
+	eventIDs := []events.ID{events.SecurityFileOpen, events.MagicWrite, events.SecurityMmapFile}
+
+	// selected data name
+	dataField := "pathname"
+
+	for _, eventID := range eventIDs {
+		if filterMap, ok := af.filters[eventID]; ok {
+			if pathName, ok := filterMap[dataField]; ok {
+				if filter, ok := pathName.(*StringFilter); ok {
```
geyslan, line 77: We can easily use the happy path here.

pkg/ebpf/c/types.h
```diff
@@ -284,6 +284,17 @@ typedef struct path_filter {
     char path[MAX_PATH_PREF_SIZE];
 } path_filter_t;
 
+typedef struct data_filter_key {
+    u32 eventid;
```
geyslan, line 288:
```suggestion
    u32 event_id;
```
nit: I know that there's still some `eventid`, but it would be nice to standardise them as well.

pkg/ebpf/c/types.h
```diff
@@ -284,6 +284,17 @@ typedef struct path_filter {
...
+    char path[MAX_PATH_PREF_SIZE];
+} data_filter_key_t;
+
+typedef struct data_filter_lpm_key {
+    u32 prefixlen;
```
geyslan, line 293:
```suggestion
    u32 prefix_len;
```

geyslan (commented):
This is just a first pass (skimming) review.

Tested and working with:

sudo ./dist/tracee -e security_file_open.data.pathname=/etc/passwd
sudo ./dist/tracee -e security_file_open.data.pathname='/etc/passwd*'
sudo ./dist/tracee -e security_file_open.data.pathname='*passwd-'

Amazing, @rscampos! 👏🏼

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -2209,6 +2209,9 @@ int BPF_KPROBE(trace_security_file_open)
     save_to_submit_buf(&p.event->args_buf, &ctime, sizeof(u64), 4);
     save_str_to_buf(&p.event->args_buf, syscall_pathname, 5);
 
+    if (!evaluate_data_filters(&p, 0))
```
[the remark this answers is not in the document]
rscampos: Good point. @yanivagman also make a suggestion and I've implemented that, seems to be simpler and don't need to change the `save_*to*_buf` API.

We just need to save the offset based on the index in the new `args_buffer_t` field called `args_offset`:
https://github.com/aquasecurity/tracee/blob/575dc3ea4945eaf0893251e5577e466fa880932c/pkg/ebpf/c/common/buffer.h#L160-L161

After that we can directly retrieve it without the for-loop:
https://github.com/aquasecurity/tracee/blob/575dc3ea4945eaf0893251e5577e466fa880932c/pkg/ebpf/c/common/buffer.h#L116

WDYT?

rscampos:
@geyslan I've pushed some changes to how we retrieve the string from args in args_buffer_t. To make it work, I added a field to args_buffer_t and modified the save_str_to_buf function.

itaysk:
> When using the LPM Trie, it always returns the longest matching string. A corner case arises when one policy (e.g., policy1) uses a substring of another policy (e.g., policy2). For instance, if policy1 covers /etc/net* and policy2 covers /etc/netconf*, a lookup for /etc/netconfig currently only returns policy2 because it is the longest match. However, it should return both policy1 and policy2. A potential solution (work in progress) is to combine equality when such a corner case is detected.

This is important, it could have been a security vulnerability. 
If you can't filter it correctly in kernel, perhaps an acceptable solution is to filter it again in userspace. We would would still gain the performance improvement from reducing events submission attemps in the bpf program, and add a second pass in userspace to ensure correctness.

rscampos:
> This is important, it could have been a security vulnerability.
If you can't filter it correctly in kernel, perhaps an acceptable solution is to filter it again in userspace. We would would still gain the performance improvement from reducing events submission attemps in the bpf program, and add a second pass in userspace to ensure correctness.

Thank you for commenting on this, @itaysk. If filtering in the kernel isn't possible, I'll definitely try this solution.

yanivagman:
> > This is important, it could have been a security vulnerability.
> > If you can't filter it correctly in kernel, perhaps an acceptable solution is to filter it again in userspace. We would would still gain the performance improvement from reducing events submission attemps in the bpf program, and add a second pass in userspace to ensure correctness.
> 
> Thank you for commenting on this, @itaysk. If filtering in the kernel isn't possible, I'll definitely try this solution.

There is no need to filter in userspace for such a corner case. It is possible to define the result of the longest prefix match to contain the results of the policies with the sub strings as well, since the return value of the map is matched policies (or rules in the future)

yanivagman (changes requested):
Looks good Raphael!
I did a first review on this draft, looking forward to see it merged

pkg/ebpf/c/common/buffer.h
```diff
@@ -98,6 +104,48 @@ statfunc int save_bytes_to_buf(args_buffer_t *buf, void *ptr, u32 size, u8 index
     return 0;
 }
 
+statfunc int load_str_from_buf(args_buffer_t *buf, char *str, u8 index)
```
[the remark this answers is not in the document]
rscampos: good point... I've followed the pattern from other functions like:
> save_str_to_buf(args_buffer_t *buf, void *ptr, u8 index)

if we change the parameters from `load_str_from_buf` maybe would be good to also change the order of the others... WYDT?

pkg/ebpf/c/common/buffer.h
```diff
@@ -98,6 +104,48 @@ statfunc int save_bytes_to_buf(args_buffer_t *buf, void *ptr, u32 size, u8 index
...
+        // Skip size
+        offset += sizeof(u32);
+
+        if (offset + sizeof(u32) > ARGS_BUF_SIZE) {
+            return 0;
+        }
+
+        if (size > MAX_PATH_PREF_SIZE) {
+            return 0;
+        }
```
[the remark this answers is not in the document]
rscampos: Upon reviewing the logic, it appears that the condition should account for the NULL character as follows:

```c
        if (size > MAX_PATH_PREF_SIZE - 1) {
            return 0;
        }
```
Considering your point: If we only copy part of the string, this approach would work for prefix matching but could result in incorrect matches for exact and suffix matches. Here's an example:

Assume: `MAX_PATH_PREF_SIZE=12`:

Using all three filter types:
**exact match:** `/etc/passwd`
**prefix match:** `/etc/pass*`
**suffix match:** `*passwd`

An access path like `/etc/passwd123` would truncate to `/etc/passwd` (size 11), which would:
- Falsely match an exact match filter for `/etc/passwd`.
- Incorrectly align with a suffix match filter for `*passwd`.

Does it make sense?

I think we have two possible options here:
1. Only allow paths with a maximum length of `MAX_PATH_PREF_SIZE - 1`.
2. Allow partial strings only for prefix matching, while requiring the entire string for exact and suffix matching.

Option 1 is complete, but Option 2 is also possible and requires some minor adjustments to the function and its caller.

pkg/ebpf/c/types.h
```diff
@@ -285,6 +285,17 @@ typedef struct path_filter {
...
+typedef struct data_filter_key {
+    u32 event_id;
+    char path[MAX_PATH_PREF_SIZE];
+} data_filter_key_t;
+
+typedef struct data_filter_lpm_key {
+    u32 prefix_len;
+    u32 event_id;
+    char path[MAX_PATH_PREF_SIZE];
+} data_filter_lpm_key_t;
```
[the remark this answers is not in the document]
rscampos: In fact, to make this more generic, we need to consider a better approach. I'll write down these considerations and come back later. Thanks.

pkg/policy/ebpf.go
```diff
@@ -358,6 +373,9 @@ func (ps *policies) updateProcTreeFilterBPF(procTreeEqualities map[uint32]equali
 const (
 	maxBpfBinPathSize = 256 // maximum binary path size supported by BPF (MAX_BIN_PATH_SIZE)
 	bpfBinFilterSize  = 264 // the key size of the BPF binary filter map entry
+
+	maxBpfPathnameSize = 128 // Maximum path size supported by BPF (MAX_PATH_PREF_SIZE)
```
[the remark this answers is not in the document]
rscampos: Yeah, make sense. I'll change the other references as well, tks.

pkg/policy/ebpf.go
rscampos: `cfg.EnabledDataFilters` should hold the bitmap of the policies used instead of holding the bitmap of the enabled data filters.

Now this is replaced by `cfg.EnabledPolicies`

pkg/ebpf/c/common/buffer.h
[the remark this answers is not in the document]
yanivagman: Yep, agree. Anyway, it doesn't have to be done in this PR of course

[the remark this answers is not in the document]
rscampos: Sure, I'll handle this refactor after this PR... but for me make sense as well.

pkg/ebpf/c/common/filtering.h
```diff
@@ -168,6 +170,25 @@ statfunc u64 bool_filter_matches(u64 match_bitmap, bool bool_value)
...
+    uint i;
+
+    // don't count null-termination since we will force it at the end
+    len = (len - 1) & MAX_DATA_FILTER_STR_SIZE_MASK;
+
+    // Copy with safe bounds checking
+    for (i = 0; i < len; i++) {
+        if (i >= MAX_DATA_FILTER_STR_SIZE) {
+            break;
+        }
```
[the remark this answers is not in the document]
rscampos: I just tried it `for (i = 0; i < len && i < MAX_DATA_FILTER_STR_SIZE; i++) {` but for some reason the eBPF Verify complain about it. For some reason the if condition that I've added was necessary just for one of the kernels I've tested - x86 Ubuntu 22.04.4 LTS (kernel 5.15.0-124-generic).

pkg/ebpf/c/common/buffer.h
```diff
@@ -98,6 +104,44 @@ statfunc int save_bytes_to_buf(args_buffer_t *buf, void *ptr, u32 size, u8 index
...
+    __builtin_memcpy(&size, &(buf->args[offset]), sizeof(u32));
+
+    // Skip size
+    offset += sizeof(u32);
+
+    if (offset + sizeof(u32) > ARGS_BUF_SIZE)
+        return 0;
+
+    if (size > MAX_DATA_FILTER_STR_SIZE - 1)
+        return 0;
```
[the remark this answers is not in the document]
rscampos: valuable nit for sure, make total sense.. tks again!

pkg/ebpf/c/common/buffer.h
```diff
@@ -10,6 +10,7 @@
 // PROTOTYPES
 
 statfunc buf_t *get_buf(int);
+statfunc data_filter_lpm_key_t *get_data_filter_buf(int);
```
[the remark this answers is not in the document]
rscampos: `get_data_filter_buf` will return a `data_filter_lpm_key_t` struct, which internally holds a `str` field (renamed from `pathname`). We plan to add other types of fields inside `data_filter_lpm_key_t` soon, so I think it's better to use a more generic name like `get_data_filter_buf`. Today make more sense to call `get_string_data_filter_buf` but I think soon will be better `get_data_filter_buf`. WDYT?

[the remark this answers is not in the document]
rscampos: Yeah... today, since we only have string filters, the data filter is very tightly coupled with the string filter, so `get_string_data_filter_buf ` make sense... but on other hand, we soon will add other field type filters, so maybe we should let it generic as `get_data_filter_buf`? WDYT?

pkg/filters/data.go
rscampos: Good ideia. I've called `af.EnableKernelFilter(id, dataName)` from `Parse`. The `EnableKernelFilter` should enable the filter in kernel for some events. After that we just need to check on `Filter` is the flag for a specific `eventID` was set.

pkg/ebpf/c/common/buffer.h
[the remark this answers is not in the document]
yanivagman: But prefix filter will only be relevant to string filters, won't it?
What other field types may fit LPM trie?

pkg/ebpf/c/common/filtering.h
```diff
@@ -323,13 +344,99 @@ statfunc u64 match_scope_filters(program_data_t *p)
...
+    // exact and prefix matching
+    data_filter_lpm_key_t *key_prefix = get_data_filter_buf(DATA_FILTER_BUF1_IDX);
+    if (key_prefix == NULL)
+        return 0;
+
+    // get event ID
+    key_prefix->event_id = p->event->context.eventid;
+
+    // get string based on index
+    __builtin_memset(key_prefix->str, 0, sizeof(key_prefix->str));
```
yanivagman, line 376: Do we really need this step if a step after we copy into `key_prefix->str`?
Since we are giving the string length to the lpm trie lookup - I don't think it is actually required to zero the unused bytes

pkg/ebpf/c/common/consts.h
```diff
@@ -24,8 +24,12 @@
 #define MAX_STR_ARR_ELEM      38 // TODO: turn this into global variables set w/ libbpfgo
 #define MAX_ARGS_STR_ARR_ELEM 15
 #define MAX_PATH_PREF_SIZE    64
+#define MAX_DATA_FILTER_STR_SIZE 128
```
yanivagman, line 27: Can you please remind me why this can't be 256 bytes long?

pkg/ebpf/c/common/buffer.h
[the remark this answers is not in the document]
geyslan: It would be worth having a `// TODO:` comment.

pkg/ebpf/c/common/buffer.h
```diff
@@ -98,6 +104,45 @@ statfunc int save_bytes_to_buf(args_buffer_t *buf, void *ptr, u32 size, u8 index
...
+    if (index > buf->argnum)
+        return 0;
+
+    offset = buf->args_offset[index];
+
+    if (offset == INVALID_ARG_OFFSET)
+        return 0;
+
+    // Ensure there is enough space for read index (u8)
+    if (offset + sizeof(u8) > ARGS_BUF_SIZE)
```
geyslan, line 122: Even though we're acquainted to operator precedence, I would enclose it to always communicate what we want explicitly. But I don't know if the current linter settings agree with me. Would you mind checking it?

```suggestion
    if ((offset + sizeof(u8)) > ARGS_BUF_SIZE)
```

pkg/ebpf/c/common/buffer.h
```diff
@@ -98,6 +104,45 @@ statfunc int save_bytes_to_buf(args_buffer_t *buf, void *ptr, u32 size, u8 index
...
+        return 0;
+
+    // Skip index
+    offset += sizeof(u8);
+
+    // Copy the size
+    __builtin_memcpy(&size, &(buf->args[offset]), sizeof(u32));
+
+    // Check size against max length, leaving space for null terminator
+    if (size > MAX_DATA_FILTER_STR_SIZE - 1)
```
geyslan, line 132: If greater should we abort filtering or consider the string filter up to the max?

If aborting is the answer, could we avoid this check by enforcing the set of the filter in userspace? If larger than max, I suppose it shouldn't be set at first.

pkg/ebpf/c/common/buffer.h
```diff
@@ -98,6 +104,45 @@ statfunc int save_bytes_to_buf(args_buffer_t *buf, void *ptr, u32 size, u8 index
...
+        return 0;
+
+    // Skip size
+    offset += sizeof(u32);
+
+    if (offset + sizeof(u32) > ARGS_BUF_SIZE)
```
geyslan, line 138: nit: as above.

pkg/ebpf/c/common/consts.h
geyslan: IIRC perf tests were done and it was decided to use 128.

pkg/ebpf/c/common/filtering.h
geyslan, line 184: We should break or return? Currently this writes nul char to dst even if len is 0.

nit: only use curly braces in blocks with more than one line or if the else block is not onelined.

```suggestion
        if (i >= MAX_DATA_FILTER_STR_SIZE)
            break;
```

pkg/ebpf/c/common/filtering.h
```diff
@@ -168,6 +170,25 @@ statfunc u64 bool_filter_matches(u64 match_bitmap, bool bool_value)
     return match_bitmap ^ (bool_value ? ~0ULL : 0);
 }
 
+statfunc void reverse_string(char *dst, char *src, int len)
+{
+    uint i;
+
+    // don't count null-termination since we will force it at the end
+    len = (len - 1) & MAX_DATA_FILTER_STR_SIZE_MASK;
```
geyslan, line 178: Is safe to consider that `len` is always passed as greater than 0?

pkg/ebpf/c/common/filtering.h
```diff
@@ -323,13 +344,99 @@ statfunc u64 match_scope_filters(program_data_t *p)
...
+// - u8 index: Index of the string data to be used as filter.
+statfunc u64 match_data_filters(program_data_t *p, u8 index)
+{
+    policies_config_t *policies_cfg = &p->event->policies_config;
+    u64 res = ~0ULL;
+
+    u16 version = p->event->context.policies_version;
+    void *filter_map = NULL;
+
+    if (!(policies_cfg->data_filter_exact_enabled || policies_cfg->data_filter_prefix_enabled ||
```
geyslan, line 361: nit: as above.

pkg/ebpf/c/common/filtering.h
```diff
@@ -323,13 +344,99 @@ statfunc u64 match_scope_filters(program_data_t *p)
...
+          policies_cfg->data_filter_suffix_enabled)) {
+        return policies_cfg->enabled_policies;
+    }
```
geyslan, line 364: nit: as above.

pkg/ebpf/c/common/filtering.h
```diff
@@ -323,13 +344,99 @@ statfunc u64 match_scope_filters(program_data_t *p)
...
+    key_prefix->event_id = p->event->context.eventid;
+
+    // get string based on index
+    __builtin_memset(key_prefix->str, 0, sizeof(key_prefix->str));
+    u32 len = load_str_from_buf(&p->event->args_buf, key_prefix->str, index);
+    if (!len)
+        return 0;
+
+    // len + size of event id
+    // prefixlen need to be multipled by 8
```
geyslan, line 382: Could you insert a reference or explain the multiplying requirement?

pkg/ebpf/c/vmlinux.h
```diff
@@ -788,6 +788,11 @@ struct iovec {
     __kernel_size_t iov_len;
 };
 
+enum
+{
+    BPF_F_NO_PREALLOC = 1,
```
geyslan, line 793:
```suggestion
    BPF_F_NO_PREALLOC = (1U << 0),
```

As in the kernel code, making it crystal as a bitmask.

pkg/filters/data.go
```diff
@@ -37,6 +183,12 @@ func (af *DataFilter) Filter(eventID events.ID, data []trace.Argument) bool {
 		return true
 	}
 
+	// No need to filter the following event IDs as they have already
+	// been filtered in the kernel
+	if af.kernelDataFilter.IsKernelFilterEnabled(eventID, "pathname") {
```
geyslan, line 188: As this is called by the pipeline (hotpath), we should think in a way of check it against integers rather strings.

geyslan (commented):
@rscampos I did another pass, good work! Let's keep the pace.

pkg/ebpf/c/common/filtering.h
rscampos: Good point, just added a ref :)

pkg/ebpf/c/common/buffer.h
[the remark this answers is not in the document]
rscampos: Updated load_str_from_buf to calculate the size and offset dynamically based on the filter type. For suffix filters, the offset is adjusted to load/reverse the last bytes of the string. This allows us to match prefixes or suffixes of pathnames that exceed 255 bytes.

pkg/ebpf/c/common/filtering.h
```diff
@@ -323,13 +344,99 @@ statfunc u64 match_scope_filters(program_data_t *p)
...
+        data_filter_lpm_key_t *key_suffix = get_data_filter_buf(DATA_FILTER_BUF2_IDX);
+
+        if (key_suffix == NULL)
+            return 0;
+
+        key_suffix->prefix_len = key_prefix->prefix_len;
+        key_suffix->event_id = key_prefix->event_id;
+
+        // reverse string for suffix match
+        __builtin_memset(key_suffix->str, 0, sizeof(key_suffix->str));
```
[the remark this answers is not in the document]
rscampos: You're right... in this case, I took this additional step unnecessarily. Good catch!

pkg/ebpf/c/common/consts.h
rscampos: We initially used `event id`, and the next available mask was 128 (for reverse string). However, now we're using 256 bytes since event id has been removed from the key of the inner map.

pkg/ebpf/c/common/buffer.h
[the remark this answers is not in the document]
rscampos: After some refactoring, I think it makes sense to have separate functions. I changed them to `get_string_data_filter_buf` and `get_string_data_filter_lpm_buf`. I'm also considering renaming some variables and maps, such as `data_filter_lpm_key_t` (to `str_data_filter_lpm_key_t`). WDYT?

rscampos:
@yanivagman @geyslan Thank you for all the reviews. I've just pushed some updates based on the discussions in this PR:

- Create `createNewDataFilterMapsVersion` in order to create the inner maps based on version and event id. This use the new approach as discussed with @yanivagman ;
	- `createNewDataFilterMapsVersion` is only used for data filters - in future it should have a common way to create any `version` + `version id` (used by match rules);
- Because event id was removed from key struct, now its possible to use a 255 characters for exact, prefix and suffix;
- Combine the bitmaps (using OR operation) when multiples filters are used in the same policy (`match_data_filters` function);
- Some fields from some `event id` are filtered in kernel. For those scenarios, the definition of such filter in CLI/policy acts with two restrictions: 1) max 255 for pathname and 2) don't allow the option contains (e.g: security_file_open.data.pathname=**\*net**\*);
- Add unit tests and integration tests;

pkg/ebpf/c/common/filtering.h
```diff
@@ -11,12 +11,15 @@
 // PROTOTYPES
 
 statfunc void *get_filter_map(void *, u16);
+statfunc void *get_filter_map_eventid(void *, u16, u32);
```
yanivagman, line 14: nit: consider renaming to `get_event_filter_map`

pkg/ebpf/c/common/buffer.h
```diff
@@ -98,6 +131,75 @@ statfunc int save_bytes_to_buf(args_buffer_t *buf, void *ptr, u32 size, u8 index
...
+    // Skip index
+    offset += sizeof(u8);
+
+    // Ensure there is enough space for read size (u32)
+    if (offset + sizeof(u32) > ARGS_BUF_SIZE)
+        return 0;
+
+    // Copy the size
+    __builtin_memcpy(&size, &(buf->args[offset]), sizeof(u32));
+    decoded_size = size;
```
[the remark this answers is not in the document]
rscampos: Tks... I missed this

pkg/ebpf/c/common/context.h
```diff
@@ -128,6 +128,8 @@ statfunc int init_program_data(program_data_t *p, void *ctx, u32 event_id)
 
     p->event->args_buf.offset = 0;
     p->event->args_buf.argnum = 0;
+    __builtin_memset(p->event->args_buf.args_offset, 0xFF, sizeof(p->event->args_buf.args_offset));
```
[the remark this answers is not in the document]
rscampos: We use `0xFF` as the initial value for the `args_offset` buffer to mark all index values as invalid. Using a different value (e.g., 0x00 or 0x80) could lead to confusion, as it might be interpreted as a valid value.

This behavior is verified in the following logic, where INVALID_ARG_OFFSET is defined as `0xFFFF`:

```c
    if (offset == INVALID_ARG_OFFSET)
        return 0;
```

pkg/utils/utils.go
```diff
@@ -73,3 +73,12 @@ func Max(x, y uint64) uint64 {
 func GenerateRandomDuration(min, max int) time.Duration {
 	return time.Duration(rand.Intn(max-min+1)+min) * time.Second
 }
+
+func ReverseString(s string) string {
+	bytes := []byte(s)
+
+	for i, j := 0, len(bytes)-1; i < j; i, j = i+1, j-1 {
+		bytes[i], bytes[j] = bytes[j], bytes[i]
```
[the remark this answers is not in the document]
rscampos: good point... reworked on that in a simple way... tks!

pkg/ebpf/c/common/buffer.h
rscampos: @geyslan the `load_str_from_buf` was complete reworded ... and its aligned with your doubt... if you don't mind, pls take a look on the function again

yanivagman (approved):
LGTM

pkg/ebpf/c/common/buffer.h
rscampos: valid point! checked and the current linter settings accepted both. tks

pkg/filters/data.go
rscampos: Good point... I'll add a comment to revisit it in the future.

pkg/filters/data.go
```diff
@@ -11,18 +11,164 @@ import (
...
+		}
+
+		fieldName, ok := filterMap[dataField]
+		if !ok {
+			continue
+		}
+
+		filter, ok := fieldName.(*StringFilter)
+		if !ok {
+			continue
```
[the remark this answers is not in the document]
rscampos: Done! tks!

pkg/ebpf/c/common/buffer.h
```diff
@@ -98,6 +144,73 @@ statfunc int save_bytes_to_buf(args_buffer_t *buf, void *ptr, u32 size, u8 index
...
+                return 0;
+            break;
+
+        case FILTER_TYPE_PREFIX:
+            size = size > MAX_DATA_FILTER_STR_SIZE ? MAX_DATA_FILTER_STR_SIZE : size;
+            break;
+
+        case FILTER_TYPE_SUFFIX:
+            if (size > MAX_DATA_FILTER_STR_SIZE) {
+                offset += size - (MAX_DATA_FILTER_STR_SIZE);
```
geyslan, line 191: nit:

```suggestion
                offset += size - MAX_DATA_FILTER_STR_SIZE;
```

pkg/filters/data.go
```diff
@@ -7,33 +7,187 @@ import (
...
+			notExactMatch:  false,
+			prefixMatch:    false,
+			notPrefixMatch: false,
+			suffixMatch:    false,
+			notSuffixMatch: false,
+		},
+	}
+}
+
+// EnableKernelFilter sets the kernel filter flag for an event and field
```
geyslan, line 49:
```suggestion
// enableKernelFilter sets the kernel filter flag for an event and field
```

pkg/filters/data.go
```diff
@@ -157,74 +324,162 @@ func (af *DataFilter) Parse(filterName string, operatorAndValues string, eventsN
...
+	}
+
+	// Enable the kernel filter if restrictions are satisfied
+	df.enableKernelFilterArg(id, dataName)
+	return val, nil
+}
+
+// CheckKernelFilterRestrictions enforces restrictions for kernel-filtered
+// fields: 1) values cannot use "contains" (e.g., start and end with "*");
+// and 2) Maximum length for the value is 255 characters.
```
geyslan, line 380:
```suggestion
// checkKernelFilterRestrictions enforces restrictions for kernel-filtered fields:
// 1) Values cannot use "contains" (e.g., start and end with "*").
// 2) Maximum length for the value is 255 characters.
```

pkg/filters/data.go
```diff
@@ -157,74 +324,162 @@ func (af *DataFilter) Parse(filterName string, operatorAndValues string, eventsN
...
+
+	// Disallow "*" and "**" as invalid values
+	if val == "*" || val == "**" {
+		return InvalidValue(val)
+	}
+
+	// Check for "contains" type filtering
+	if len(val) > 1 && val[0] == '*' && val[len(val)-1] == '*' {
+		return InvalidFilterType()
+	}
```
geyslan, line 394: Using `strings.Count()` will consider cases as `"*some*hing"` invalid as well.

Would this suggestion work as expected?

```suggestion
	// Disallow a single "*" as invalid value
	if val == "*" {
		return InvalidValue(val)
	}

	// Check for "contains" type filtering
	if len(val) > 1 && strings.Count(val, "*") > 1 {
		return InvalidFilterType()
	}
```

tests/integration/event_filters_test.go
```diff
@@ -1700,6 +1700,336 @@ func Test_EventFilters(t *testing.T) {
...
+						"02200000000230000000024000000002500056789; more /tmp/00000100000000200000000030000000004000000000500000000060000000007"+
+						"0000000008000000000900000000010000000001100000000120000000013000000001400000000150000000016000000001700000000180000000"+
+						"01900000000200000000021000000002200000000230000000024000000002500056789'",
+					0,
+					1*time.Second,
+					[]trace.Event{
+						expectEvent(anyHost, "more", testutils.CPUForTests, anyPID, 0, events.SecurityFileOpen, orPolNames("sfo-pol-1", "sfo-pol-2"), orPolIDs(1, 2),
+							expectArg("pathname", "/tmp/00000100000000200000000030000000004000000000500000000060000000007000000000800000000090000000001"+
+								"0000000001100000000120000000013000000001400000000150000000016000000001700000000180000000019000000002000000000210000000"+
+								"02200000000230000000024000000002500056789")),
```
geyslan, line 2024: These file names will be hard to handle in future changes. Is it possible to use alpha chars instead?

geyslan (approved):
LGTM. I also put some suggestions. Great work, @rscampos ❤️

pkg/filters/data.go
rscampos: That makes sense, thank you for highlighting this. Today if we define a filter like this`/tmp/file\*`, this will be added as prefix (maybe the correct should be exact). In the end both ways will match a file access since `/tmp/file*` will be matched by the prefix `/tmp/file`.

pkg/policy/rules.go
```diff
@@ -0,0 +1,158 @@
...
+
+import (
+	"github.com/aquasecurity/tracee/pkg/events"
+	"github.com/aquasecurity/tracee/pkg/filters"
+	"github.com/aquasecurity/tracee/pkg/utils"
+)
+
+type RulesData struct {
+	Data                        map[events.ID]RuleData
+	kernelDataFilterMatchStates KernelDataFilterMatchStates
```
yanivagman, line 11: I don't think we should add this extra Data hierarchy level.
This is related to the fact that you added `kernelDataFilterMatchStates` in here, but I think this is not the correct abstraction level to add it into.
The current policy struct represents the way a policy in Tracee looks like:

```
type Policy struct {
        **// Policy metadata:**
	ID                int
	Name              string
        **// All scope filters are below:**
	UIDFilter         *filters.UIntFilter[uint32]
	PIDFilter         *filters.UIntFilter[uint32]
	NewPidFilter      *filters.BoolFilter
	MntNSFilter       *filters.UIntFilter[uint64]
	PidNSFilter       *filters.UIntFilter[uint64]
	UTSFilter         *filters.StringFilter
	CommFilter        *filters.StringFilter
	ContFilter        *filters.BoolFilter
	NewContFilter     *filters.BoolFilter
	ContIDFilter      *filters.StringFilter
	ProcessTreeFilter *filters.ProcessTreeFilter
	BinaryFilter      *filters.BinaryFilter
	Follow            bool
        **// All rules are grouped, just like in a policy of Tracee**
	Rules             map[events.ID]RuleData
}

type RuleData struct {
	EventID     events.ID
        **// All the per-rule filters are here:**
	ScopeFilter *filters.ScopeFilter
	DataFilter  *filters.DataFilter
	RetFilter   *filters.IntFilter[int64]
}
```

So, I think the kernel data filter logic should be pushed to a lower level object, probably during kernel filter map calculation in ebpf.go

yanivagman (changes requested):

pkg/policy/rules.go
rscampos: Good ideia. I did another version where I called a new function`computeDataFilterEqualities` inside the function `computeFilterEqualities` to compute kernel data filter. The rest of the logic remains the same.

https://github.com/aquasecurity/tracee/blob/f8b46e634dade61fdac21d0d65d4c580016f5099/pkg/policy/equality.go#L227

After 1) computing kernel data filter equalities, we proceed to 2) generate the config_map. To do this, we need to determine if at least one of the following types was set: exact, not exact, prefix, not prefix, suffix, or not suffix. Since we already iterate over all policies in step 1, I added a field to the policies struct to store this computed data. WDYT?

yanivagman (approved):
LGTM
Great work @rscampos !

rscampos:
Folks @yanivagman @geyslan,

Thank you for all the feedbacks! Learned a lot of good things during this work!

geyslan:
Congrats for this amazing new feature! 🚀🥳

PR #4327, opened 2024-09-25
analyze: enable sigs consuming sigs

### LINKED ISSUE #4326: Support Ad-Hoc analysis of signatures which consume other signatures
We have previously introduced the feature of outputting signatures based on inputs from previous signature findings in the tracee one-binary version (#3681). This is exactly why this feature does not currently work in ad-hoc analysis modes such as analyze and rules. Since we plan on deprecating rules, we should at least support this in analyze mode.

NDStrahilevitz:
df567ebc7 **fix(analyze): enable sigs consuming sigs**
ad56d06f8 **fix(findings): clone properties before rewriting**
84ceba74a **chore: move finding event conversion to a package**
3b3299707 **chore: decouple sig init from init**

df567ebc7 **fix(analyze): enable sigs consuming sigs**

```
Signatures consuming other signatures relied on the single binary
reprocessing finding events into the pipeline. This did not occur in
analyze mode. Introduce that mechanism so that it now works.

Co-authored-by: Asaf Eitani <asaf.eitani@aquasec.com>
```

84ceba74a **chore: move finding event conversion to a package**

```
Opportunistic refactor. Logic does not relate to eBPF and does relate to
event data. Also allows importing this logic without importing eBPF
related code.
```

3b3299707 **chore: decouple sig init from init**

```
This allows initializing signature to event data without importing eBPF
initialization logic, which require specific compilation tags.

Co-authored-by: Asaf Eitani <asaf.eitani@aquasec.com>
```

### 2. Explain how to test it

Currently we have internal tests for it. Confirmed it was working using those.

### 3. Other comments

Resolve #4326

### DIFF
--- a/cmd/tracee/cmd/analyze.go
+++ b/cmd/tracee/cmd/analyze.go
@@ -13,7 +13,7 @@ import (
 	"github.com/spf13/viper"
 
 	"github.com/aquasecurity/tracee/pkg/cmd/flags"
-	"github.com/aquasecurity/tracee/pkg/cmd/initialize"
+	"github.com/aquasecurity/tracee/pkg/cmd/initialize/sigs"
 	tracee "github.com/aquasecurity/tracee/pkg/ebpf"
 	"github.com/aquasecurity/tracee/pkg/events"
 	"github.com/aquasecurity/tracee/pkg/logger"
@@ -106,7 +106,7 @@ tracee analyze --events anti_debugging events.json`,
 			signatureEvents = nil
 		}
 
-		sigs, _, err := signature.Find(
+		signatures, _, err := signature.Find(
 			rego.RuntimeTarget,
 			rego.PartialEval,
 			viper.GetStringSlice("signatures-dir"),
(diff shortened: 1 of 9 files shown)


cmd/tracee/cmd/analyze.go
```diff
@@ -211,18 +212,21 @@ func produce(ctx context.Context, inputFile *os.File, engineInput chan protocol.
 	}
 }
 
-func process(finding *detect.Finding) {
-	event, err := tracee.FindingToEvent(finding)
-	if err != nil {
-		logger.Fatalw("Failed to convert finding to event", "err", err)
-	}
+func findingProcessor(engineInput chan<- protocol.Event) func(finding *detect.Finding) {
```
geyslan, line 215: Do you plan reusing this helper?

geyslan (approved):
LGTM

geyslan:
I tested it again since I noticed you removed the recursive loading what made the initial behaviour stands. For each selected event that is a signature, analyze emits an error:

```
{"level":"error","ts":1727381548.0349305,"msg":"Failed to load event dependency","event":"internal_sig"}
...
```

NDStrahilevitz:
> I tested it again since I noticed you removed the recursive loading what made the initial behaviour stands. For each selected event that is a signature, analyze emits an error:
> 
> ```
> {"level":"error","ts":1727381548.0349305,"msg":"Failed to load event dependency","event":"internal_sig"}
> ...
> ```

Weird, I haven't seen it in my test. Maybe because I checked with grep, let me see.

NDStrahilevitz:
Manually tested on both tracee-rules and analyze.
@geyslan please review at your pace.

geyslan (approved):
LGTM

NDStrahilevitz:
Waiting for a final review by @geyslan before merging. 
I've made changes such that the feedback concern is handled entirely in the engine context and not in analyze. Later on, the feedback should be included into derived events and fed into the processing stage.
Further, i've resolved some crash issues that were introduced in initial versions of the PR, a concurrency safety issue in finding->event conversion, and made sure that analyze mode remains as deterministic as it was before (it is msotly deterministic in output size but not in order, which may affect output size given a signature dependent on order).

geyslan:
SGTM.

PR #4328, opened 2024-09-25
fix(mount): reintroduce root path requirement

### LINKED ISSUE #4311: Cgroupfs mounting doesn't work on EKS
### Case
AWS seems to setup pods such that their cgroup filesystem are already mounted on their own cgroup root (see the reproduction steps below for an example). This means that when tracee searches for an already managed cgroupfs it finds that one which obscures the host cgroup fs. In addition, it prevents mounting the actual cgroup filesystem.

PR https://github.com/aquasecurity/tracee/pull/4076 broke the previous fix for EKS mounting. We previously confirmed through `/proc/self/mountinfo` that the root of the filepath mount was `/`, which means that it must be the host. This was changed in favor of merely detecting the lowest inode and considering that the host cgroup, which is apparently false.

### Possible Solutions
1. Detect the specific case of a mounted cgroupfs on top of a pod through the mounted pattern. Then handle it in particular by unmounting and remounting.
2. Make it so finding *no* root filesystem mount triggers the unmount and remounting

### Reproduction Steps
Deploy tracee on EKS
exec into the pod
`cat /proc/self/mountinfo`
confirm that the mount looks something like:
`1575 1572 0:28 /kubepods.slice/kubepods-besteffort.slice/kubepods-besteffort-pod6795064b_0010_4189_a6fd_fcb7a661cc2c.slice/cri-containerd-0c5972fc0d0b7d4c0889ea8010098f07b294449bd948384d2222085be107d8a1.scope /sys/fs/cgroup/cpuset rw,nosuid,nodev,noexec,relatime master:11 - cgroup cgroup rw,cpuset`

HEAD

## Output of `uname -a`:

```
Linux tracee-7fgc7 5.10.224-212.876.amzn2.x86_64 #1 SMP Thu Aug 22 16:55:24 UTC 2024 x86_64 GNU/Linux
```

NDStrahilevitz:

### 2. Explain how to test it

1. Deploy tracee to an EKS env
2. sh into it with `kubectl exec -it <pod_name> -- /bin/sh`
3. `cat /proc/self/mountinfo`
4. Confirm that an appropriate cgroupfs mount exists for cpuset or cgroup2
5. Importantly, if it is known that the deployed tracee instance has cgroupfs mounted via k8s or docker, there should be no additional mount for the relevant cgroup version (cgroup v2 may still be mounted by tracee in cgroupv1 envs)
6. `ls` that cgroupfs folder and confirm that `kubepods.slice` exists and other pods are visible
7. Optionally: check the debug logs at tracee's startup for info on the mount object and confirm it aligns with expected result per the environment.

### 3. Other comments

Resolve #4311

### DIFF
--- a/pkg/mount/mount.go
+++ b/pkg/mount/mount.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 	"syscall"
 
@@ -217,11 +218,11 @@ func IsFileSystemSupported(fsType string) (bool, error) {
 // - int: The inode number of the matching mountpoint.
 // - error: Any error encountered while reading the /proc/mounts file.
 func SearchMountpointFromHost(fstype string, search string) (string, int, error) {
+	const mountRootIndex = 3
 	const mountpointIndex = 4
-	const fsTypeIndex = 8
 
-	mp := ""
-	inode := 0
+	mp := ""   // matched mountpoint search var
+	inode := 0 // matched mountpoint's inode
 
 	file, err := os.Open(procMounts)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


yanivagman:
@NDStrahilevitz can you please confirm this fix was tested both on EKS and TAS environments?

NDStrahilevitz:
Just confirmed it works in TAS as well as EKS. GTG @yanivagman

yanivagman (approved):
LGTM

PR #4329, opened 2024-09-25
fix(ebpf): set pipeline chan size from config

geyslan:

5ee8cafcf **fix(ebpf): set pipeline chan size from config**

```
Make all pipeline channels have the size set from the config option.
```


NDStrahilevitz (changes requested):

NDStrahilevitz:
Also i'd argue this is a fix kind commit and not a chore as it is a bug that these sizes were not set by the config.

NDStrahilevitz (approved):
LGTM

Issue #4330, opened 2024-09-26
`sched_process_fork` emits errors instead of events


oshaked1:
## Description

The `sched_process_fork` event is broken, instead of events the following errors come out:

```
{"level":"error","ts":1727330855.408098,"msg":"Error processing event","event":"sched_process_fork","error":"ebpf.(*Tracee).registerEventProcessors.(*Tracee).normalizeTimeArg.func1: argument parent_process_start_time of event sched_process_fork is not of type uint64"}
```

To see this simply run

```
sudo dist/tracee -e sched_process_fork
```

## Output of `tracee version`:

```
Tracee version: v0.22.0-36-g795a95b1b
```

PR #4332, opened 2024-09-26
register normalizeTimeArg processor only when proctree is on

geyslan:
Close: #4330

0b35baa4d **fix(ebpf): register processor conditionally**

```
Register SchedProcessFork processors, including normalizeTimeArg,
only if proctree is enabled.
```

206e16072 **fix(ebpf): try to assert only when it is not nil**

```
The normalizeTimeArg function was not checking if the value was nil
before asserting it as an uint64.

This also adds the type to the error message.
```

### 2. Explain how to test it

`sudo ./dist/tracee -e sched_process_fork`
`sudo ./dist/tracee --proctree source=both -e sched_process_fork` to have related args filled.

### DIFF
--- a/pkg/ebpf/processor.go
+++ b/pkg/ebpf/processor.go
@@ -79,26 +79,24 @@ func (t *Tracee) RegisterEventProcessor(id events.ID, proc func(evt *trace.Event
 // registerEventProcessors registers all event processors, each to a specific event id.
 func (t *Tracee) registerEventProcessors() {
 	//
-	// Event Timestamps Normalization
-	//
-
-	// Convert all time relate args to nanoseconds since epoch.
-	// NOTE: Make sure to convert time related args (of your event) in here, so that
-	// any later code has access to normalized time arguments.
-	t.RegisterEventProcessor(events.SchedProcessFork, t.normalizeTimeArg(
-		"start_time",
-		"parent_start_time",
-		"parent_process_start_time",
-		"leader_start_time",
-	))
-
-	//
 	// Process Tree Processors
 	//
 
 	// Processors registered when proctree source "events" is enabled.
 	switch t.config.ProcTree.Source {
 	case proctree.SourceEvents, proctree.SourceBoth:
+		// Event Timestamps Normalization
+		//
+		// Convert all time relate args to nanoseconds since epoch.
+		// NOTE: Make sure to convert time related args (of your event) in here, so that
+		// any later code has access to normalized time arguments.
+		t.RegisterEventProcessor(events.SchedProcessFork, t.normalizeTimeArg(
+			"start_time",
+			"parent_start_time",
+			"parent_process_start_time",
+			"leader_start_time",
+		))
+
 		t.RegisterEventProcessor(events.SchedProcessFork, t.procTreeForkProcessor)
 		t.RegisterEventProcessor(events.SchedProcessExec, t.procTreeExecProcessor)
 		t.RegisterEventProcessor(events.SchedProcessExit, t.procTreeExitProcessor)
--- a/pkg/ebpf/processor_funcs.go
+++ b/pkg/ebpf/processor_funcs.go
@@ -361,9 +361,13 @@ func (t *Tracee) normalizeTimeArg(argNames ...string) func(event *trace.Event) e
 			if arg == nil {
 				return errfmt.Errorf("couldn't find argument %s of event %s", argName, event.EventName)
 			}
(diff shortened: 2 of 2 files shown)


NDStrahilevitz (commented):
In what case will a time arg be nil?

geyslan:
> In what case will a time arg be nil?

Currently, when not using proctree, see:

`sudo ./dist/tracee -e sched_process_fork`

```
17:35:12:005276  1000   IPDL Background  16017   16052   0                sched_process_fork        parent_tid: 16052, ..., **parent_process_start_time: <nil>**, leader_tid: <nil>, leader_ns_tid: <nil>, leader_pid: <nil>, leader_ns_pid: <nil>, leader_start_time: <nil>
```

`sudo ./dist/tracee --proctree source=both -e sched_process_fork`

```
17:37:29:009715  1000   firefox          16017   16017   0                sched_process_fork        parent_tid: 16017, ..., **parent_process_start_time: 1727389186434828788**, leader_tid: 16017, leader_ns_tid: 16017, leader_pid: 16017, leader_ns_pid: 16017, leader_start_time: 1727389260327760332
```

NDStrahilevitz:
> > In what case will a time arg be nil?
> 
> Currently, when not using proctree, see:
> 
> `sudo ./dist/tracee -e sched_process_fork`
> 
> ```
> 17:35:12:005276  1000   IPDL Background  16017   16052   0                sched_process_fork        parent_tid: 16052, ..., **parent_process_start_time: <nil>**, leader_tid: <nil>, leader_ns_tid: <nil>, leader_pid: <nil>, leader_ns_pid: <nil>, leader_start_time: <nil>
> ```
> 
> `sudo ./dist/tracee --proctree source=both -e sched_process_fork`
> 
> ```
> 17:37:29:009715  1000   firefox          16017   16017   0                sched_process_fork        parent_tid: 16017, ..., **parent_process_start_time: 1727389186434828788**, leader_tid: 16017, leader_ns_tid: 16017, leader_pid: 16017, leader_ns_pid: 16017, leader_start_time: 1727389260327760332
> ```

That's odd. Sounds to me like that is the bug.

geyslan:
If you're not using proctree, it should have been fed by the control plane signal, right?

NDStrahilevitz:
Ok so these arguments are sort of "metadata" arguments which are submitted on a need basis for the process tree. I then question what merit they have as arguments. Arguments or as want to call it "event data" should reflect user informative data on the event. If this is not such data, that is, it is some sort of "metadata" argument, then that reflects an issue in the event format.
Further, it implies that the processing should be applied conditionally. WDYT?
@yanivagman @itaysk I think we need to take this usecase into account in https://github.com/aquasecurity/tracee/issues/2870 (event specific "data" which is relevant for internal logic but not for the user)

geyslan:
> Further, it implies that the processing should be applied conditionally. WDYT?

It is:

https://github.com/aquasecurity/tracee/blob/1a0065b370320fd55eae7eb952719f0e785ec2a0/pkg/ebpf/processor.go#L100-L102

NDStrahilevitz:
@geyslan 

The relevant processor is here https://github.com/aquasecurity/tracee/blob/1a0065b370320fd55eae7eb952719f0e785ec2a0/pkg/ebpf/processor.go#L88

geyslan:
Nice. This makes sense to me now. So yeah, I believe it should be in the condition below.

pkg/ebpf/processor.go
```diff
@@ -78,27 +78,25 @@ func (t *Tracee) RegisterEventProcessor(id events.ID, proc func(evt *trace.Event
...
-	))
-
 	//
 	// Process Tree Processors
 	//
 
 	// Processors registered when proctree source "events" is enabled.
 	switch t.config.ProcTree.Source {
 	case proctree.SourceEvents, proctree.SourceBoth:
+		// Event Timestamps Normalization
```
NDStrahilevitz, line 88: Logically this works. But I would rather have it in a separate section to convey the intention of all timestamp arguments being registered and processed first. I do think that an effort in another PR of yours may help us move this logic from the processing alltogether, I'll elaborate there.

NDStrahilevitz (changes requested):

NDStrahilevitz (approved):
I want to resolve this properly with the changes to argument typing (and thus move this logic to decoding timestamp args).
So for now, lets merge this and more importantly cherry-pick to the v0.22.3 release so that users don't receive the repeated errors.

PR #4334, opened 2024-09-27
Perf event writes metric

geyslan:

bae153278 **chore(performance): update tracee dashboard**
5a5c2e7ce **chore: add perfbuf metric per event (METRICS=1)**
5fe36d177 **chore(go.mod): bump api to latest 1f796cb**

5a5c2e7ce **chore: add perfbuf metric per event (METRICS=1)**

```
Enabled only when built with METRICS=1.

BPFPerfEventSubmitAttemptsCount and BPFPerfEventSubmitFailuresCount
count the number of events processed by the eBPF programs and written to
or attempted to be written to the perf buffer.

It is incremented right after the attempt of writing the event to the
perf buffer, making it possible to measure if the that event was
successfully written to the perf buffer or not.

This metric can be used to monitor the performance of individual eBPF
events and to detect potential bottlenecks.
```

image

### 2. Explain how to test it

`METRICS=1 make tracee`

`sudo ./dist/tracee -e socket_dup,execve --metrics --pyroscope --pprof`

`make -f builder/Makefile.performance dashboard-start`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -167,6 +167,10 @@ else
 	GO_DEBUG_FLAG = -w
 endif
 
+ifeq ($(METRICS),1)
+	BPF_DEBUG_FLAG += -DMETRICS
+endif
+
 ifeq ($(UNAME_M),x86_64)
 	ARCH = x86_64
 	LINUX_ARCH = x86
@@ -423,6 +427,7 @@ $(OUTPUT_DIR)/tracee.bpf.o: \
 	$(TRACEE_EBPF_OBJ_HEADERS)
 #
 	$(CMD_CLANG) \
+		$(BPF_DEBUG_FLAG) \
 		-D__TARGET_ARCH_$(LINUX_ARCH) \
 		-D__BPF_TRACING__ \
 		-DCORE \
@@ -501,6 +506,7 @@ $(OUTPUT_DIR)/tracee: \
 		-ldflags="$(GO_DEBUG_FLAG) \
 			-extldflags \"$(CGO_EXT_LDFLAGS_EBPF)\" \
 			-X github.com/aquasecurity/tracee/pkg/version.version=$(VERSION) \
+			-X github.com/aquasecurity/tracee/pkg/version.metrics=$(METRICS) \
 			" \
 		-v -o $@ \
 		./cmd/tracee
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/IBM/fluent-forward-go v0.2.2
 	github.com/Masterminds/sprig/v3 v3.2.3
 	github.com/aquasecurity/libbpfgo v0.7.0-libbpf-1.4.0.20240729111821-61d531acf4ca
-	github.com/aquasecurity/tracee/api v0.0.0-20241202151435-715b6290fb6a
+	github.com/aquasecurity/tracee/api v0.0.0-20241203172838-1f796cb64289
 	github.com/aquasecurity/tracee/signatures/helpers v0.0.0-20241009193135-0b23713fa9f9
 	github.com/aquasecurity/tracee/types v0.0.0-20241008181102-d40bc1f81863
 	github.com/containerd/containerd v1.7.21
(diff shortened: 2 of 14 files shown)


go.mod
```diff
@@ -179,3 +179,5 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	kernel.org/pub/linux/libs/security/libcap/psx v1.2.70 // indirect
 )
+
+replace github.com/aquasecurity/tracee/api => ./api
```
geyslan, line 183: This commit must be changed to reflect the new api.

api/v1beta1/diagnostic.proto
```diff
@@ -12,11 +12,12 @@ message GetMetricsResponse {
 	uint64 EventsFiltered = 2;
 	uint64 NetCapCount = 3;
 	uint64 BPFLogsCount = 4;
-	uint64 ErrorCount = 5;
-	uint64 LostEvCount = 6; 
-	uint64 LostWrCount = 7; 
-	uint64 LostNtCapCount = 8; 
-	uint64 LostBPFLogsCount = 9;
+	uint64 BPFPerfEventWrites = 5;
```
geyslan, line 15: This commit must be merged first in a separate PR.

Makefile
```diff
@@ -501,6 +504,7 @@ $(OUTPUT_DIR)/tracee: \
 		-ldflags="$(GO_DEBUG_FLAG) \
 			-extldflags \"$(CGO_EXT_LDFLAGS_EBPF)\" \
 			-X github.com/aquasecurity/tracee/pkg/version.version=$(VERSION) \
+			-X github.com/aquasecurity/tracee/pkg/version.debug=$(DEBUG) \
```
geyslan, line 507: Perhaps to change that flag to enable the metric with standard build. I mean, making possible to get that metric but with stripped binaries.

geyslan:
After offline discussion, registering another way of approaching this: by enabling the metric with a flag instead of requiring a rebuild.

It's feasible considering an increment of program sizes, an additional branch check (metric enabled) and a dangling map.

geyslan:
I'm working on a stress tool on top of this branch. Gonna let you know soon.

Makefile
geyslan: Update grafana dashboard.

pkg/metrics/collector.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/aquasecurity/tracee/pkg/counter"
+	"github.com/aquasecurity/tracee/pkg/logger"
+)
+
+type Collector struct {
+	m           sync.RWMutex
+	description string
+	values      map[uint64]uint64
```
geyslan, line 16: @NDStrahilevitz I have fixed this key/value, but perhaps they can be generic. WDYT?

geyslan: Already made it generic.

performance/dashboard/docker-compose.yml
```diff
@@ -11,7 +9,7 @@ services:
     volumes:
       - /proc:/host/proc:ro
       - /sys:/host/sys:ro
-      - /:/rootfs:ro
+      - /:/rootfs:ro,rslave
```
geyslan, line 12: Opportunistic change to accommodate new version:

`Error response from daemon: invalid mount config: must use either propagation mode "rslave" or "rshared" when mount source is within the daemon root, daemon root: "/var/lib/docker", bind mount source: "/", propagation: "rprivate"`

geyslan:
@NDStrahilevitz @yanivagman I would appreciate your +1.

yanivagman (approved):
LGTM, see one nit comment

pkg/ebpf/tracee.go
```diff
@@ -63,7 +64,9 @@ type Tracee struct {
 	running   atomic.Bool
 	done      chan struct{} // signal to safely stop end-stage processing
 	OutDir    *os.File      // use utils.XXX functions to create or write to this file
-	stats     metrics.Stats
+	// Metrics
+	stats *metrics.Stats
+	// Engine
```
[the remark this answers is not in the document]
geyslan: You're right. It's a leftover.

Issue #4335, opened 2024-10-04
Event sched_process_exec is sending ArgMeta with nil value in gRPC


trvll:
## Description

The event `sched_process_exec` is received on gRPC client with ArgMeta value nil which causes the parser to fail when trying to convert protobuf to tracee.Event on client side.

Tested on v0.22.1 at Linux ce08d0625488 6.2.0-1018-aws #18~22.04.1-Ubuntu SMP Wed Jan 10 22:54:16 UTC 2024 x86_64 Linux

![image](https://github.com/user-attachments/assets/b040421e-19c1-482f-9d8f-4090e8260298)

![image](https://github.com/user-attachments/assets/6559243b-6579-4bc0-ad1f-d040a94d282e)

PR #4336, opened 2024-10-04
Handle zero-value types for unavailable fields - ArgMeta

geyslan:
Close: #4335

2f31471ef **fix(events): setuid16 arg name typo**
473762535 **fix: handle zero-value types for unavailable fields**
41eeb29f2 **chore(go.mod): bump types to latest version**

2f31471ef **fix(events): setuid16 arg name typo**

```
'git log --grep old_old_uid_t' in Linux kernel code didn't return any
results.
```

473762535 **fix: handle zero-value types for unavailable fields**

```
Some event arguments, such as "interpreter_*" from sched_process_exec,
were not being populated because the kernel might not always provide
them. This was causing errors during the gRPC proto conversion, as
the values couldn't be asserted - they were nil.

With this change, ArgMeta now holds the Zero value of the type, ensuring
that assignment is always possible without requiring additional parsing
in the pipeline.

This also might help parsing since the Zero field will always have
value, making it easier to assert its type instead of checking for
Name field (string).
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/aquasecurity/libbpfgo v0.7.0-libbpf-1.4.0.20240729111821-61d531acf4ca
 	github.com/aquasecurity/tracee/api v0.0.0-20240905132323-d1eaeef6a19f
 	github.com/aquasecurity/tracee/signatures/helpers v0.0.0-20240607205742-90c301111aee
-	github.com/aquasecurity/tracee/types v0.0.0-20240607205742-90c301111aee
+	github.com/aquasecurity/tracee/types v0.0.0-20241008181102-d40bc1f81863
 	github.com/containerd/containerd v1.7.21
 	github.com/docker/docker v26.1.5+incompatible
 	github.com/golang/protobuf v1.5.4
--- a/pkg/events/definition_group.go
+++ b/pkg/events/definition_group.go
@@ -5,6 +5,7 @@ import (
 	"sync"
 
 	"github.com/aquasecurity/tracee/pkg/errfmt"
+	"github.com/aquasecurity/tracee/pkg/events/parse"
 	"github.com/aquasecurity/tracee/pkg/logger"
 )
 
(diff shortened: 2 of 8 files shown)


geyslan:
@trvll do you mind testing?

pkg/events/definition_group.go
```diff
@@ -123,6 +124,9 @@ func (d *DefinitionGroup) AddBatch(givenDefs map[ID]Definition) error {
 	defer d.mutex.Unlock()
 
 	for id, def := range givenDefs {
+		for i := range def.params {
+			def.params[i].Zero = parse.ArgZeroValueFromType(def.params[i].Type)
```
geyslan, line 128: Filling the respective zero value in the initialisation saves us of parsing further to discover the type of the nil arg.

go.mod
```diff
@@ -179,3 +179,5 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	kernel.org/pub/linux/libs/security/libcap/psx v1.2.70 // indirect
 )
+
+replace github.com/aquasecurity/tracee/types => ./types
```
geyslan, line 183: To be dropped when types changes be merged at first.

NDStrahilevitz:
I'll review this thoroughly tomorrow.
From a quick glance I have these to say:
1. We should use the effort of type conversion here and already move to hardcoded `ArgType` enums. These should be used instead of the strings. (We can then add a timestamp argument type and have it autonormalized in the decode stage)
2. I believe we're handling this issue or a similar one through the argument helper here: https://github.com/aquasecurity/tracee/blob/main/signatures/helpers/arguments_helpers.go#L35. What is the difference in issue? I think further elaboration in the resolved issue would have helped clear the difference.

types/trace/trace.go
```diff
@@ -136,8 +136,9 @@ type Argument struct {
 
 // ArgMeta describes an argument
 type ArgMeta struct {
-	Name string `json:"name"`
-	Type string `json:"type"`
+	Name string      `json:"name"`
+	Type string      `json:"type"`
+	Zero interface{} `json:"-"` // Zero value of the argument
```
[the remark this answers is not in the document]
geyslan: It won't, it's just used internally. As the `Argument.Value` will always be set (zero or not), the `ArgMeta.Zero` is not necessary beyond initial stages of pipeline.

pkg/events/definition_group.go
```diff
@@ -123,6 +124,10 @@ func (d *DefinitionGroup) AddBatch(givenDefs map[ID]Definition) error {
 	defer d.mutex.Unlock()
 
 	for id, def := range givenDefs {
+		for i := range def.params {
+			// set zero value for each parameter based on its type name
```
NDStrahilevitz, line 128:
```suggestion
			// set zero value in the argument definition once so it can be 
			// reused without recalculation later.
```

NDStrahilevitz: I think it conveys the intention of the code better.

PR #4340, opened 2024-10-08
chore(types): add Zero field to ArgMeta

geyslan:
Requirement for https://github.com/aquasecurity/tracee/pull/4336

d40bc1f81 **chore(types): add Zero field to ArgMeta**

```
This is field for internal use which holds the zero value
of ArgMeta.Type.
```


rscampos (approved):
LGTM

PR #4331, opened 2024-09-26
feat: remove default usage of parse-arguments

geyslan:
Close: https://github.com/aquasecurity/tracee/issues/2177

### 1. Explain what the PR does

47451f562 **feat(cmd): disable arg parsing by default**
cb48e1fca **fix: possible overwrite of Args on Derive stage**
3cfe5c5a4 **chore(ebpf): copy event before parsing it**
0fce1839b **feat(rego): parse event arguments in sig**
ec478cc6a **feat(sigs): refactor to use nonparsed arguments**
ddc219a3a **chore(go.mod): bump sig helpers to 0b23713f**

3cfe5c5a4 **chore(ebpf): copy event before parsing it**

```
Refactor 'event feeding to engine' logic to copy the event before
parsing it, cloning its arguments.
```

0fce1839b **feat(rego): parse event arguments in sig**

```
Removing default argument parsing will break rego signatures, whose
implementation depends on parsed arguments. In order to keep the option
of readability in REGO signatures, events will be parsed by default in
the context of evaluating these kind of signatures.
This will also ensure that they will not be broken depending on the
selection of parse-arguments.

Co-authored-by: Geyslan Gregório <geyslan@gmail.com>
```

ec478cc6a **feat(sigs): refactor to use nonparsed arguments**

```
Co-authored-by: Geyslan Gregório <geyslan@gmail.com>
```

### 2. Explain how to test it

`-e ptrace -o json`

```json
{"timestamp":1727372470993574813,"threadStartTime":1727372470988592715,"processorId":30,"processId":250996,"cgroupId":15795,"threadId":250996,"parentProcessId":250785,"hostProcessId":250996,"hostThreadId":250996,"hostParentProcessId":250785,"userId":1000,"mountNamespace":4026531841,"pidNamespace":4026531836,"processName":"strace","executable":{"path":""},"hostName":"hb0","containerId":"","container":{},"kubernetes":{},"eventId":"101","eventName":"ptrace","matchedPolicies":[""],"argsNum":4,"returnValue":0,"syscall":"ptrace","stackAddresses":null,"contextFlags":{"containerStarted":false,"isCompat":false},"threadEntityId":1253568537,"processEntityId":1253568537,"parentEntityId":1331977861,"args":[{"name":"request","type":"long","value":24},{"name":"pid","type":"pid_t","value":250999},{"name":"addr","type":"void*","value":0},{"name":"data","type":"void*","value":0}]}
```

`-e ptrace -o json -o option:parse-arguments`

```json
[...]

PR #4345, opened 2024-10-09
Signatures helpers improvement

geyslan:
Required by: #4331

0b23713fa **feat(helpers): robust int and uint arg helpers**

```
Add a new uint argument helpers, which extracts a data argument from an
event by name and returns it if it was one of the uint types.
Additionally, add further checks for int64 and int types in the already
existing int argument helper.
```

4d5e47de0 **feat(helpers): unparsed flag helpers**

```

```


rscampos (approved):
LGTM

PR #4353, opened 2024-10-16
feat!: separate data field decoding types

NDStrahilevitz:

b5489a559 **feat!(events): hardtype and streamline event decoding**
e0cb894bc **go.mod: temp replace**
9f7f0fa2b **chore(trace): align with normalized data types**
00c808fa1 **feat(trace): add pointer type alias**

b5489a559 **feat!(events): hardtype and streamline event decoding**

```
1. Reduce the type variance in data fields via hardcoding possible types
   decodable from the kernel.
2. (Almost) all event data type field names (post-decode) should now
   conform to go type names. This will ensure a more consistent and
   simpler format for event data.
3. Add data field decode types to the events package
4. Make corresponding changes in eBPF and test code
5. Move some functions around, rename and improve documentation in
   bufferdecoder.
6. Implement decode type to data field type conversion logic in the
   eventsreader package. Each decode type is mapped to a default data
   field type.
6. Replace all event pointer types with `trace.Pointer`
7. Replace all places where the code depends on uintptr for presenting
   pointer data with the new alias type. This means that pointers are
   invariantly represented by a 64bit type. This change makes no effect
   on lower level code which actually deals with pointers.
```

00c808fa1 **feat(trace): add pointer type alias**

```
Decoding and encoding pointer data fields has so far relied on the
uintptr go type. But this type is architecture dependent, and tracee
should be able to parse 64-bit pointers in any case (tracee with ebpf
capabilities is only supported on 64-bit machines, but tracee analyze
should potentially be able to differ. In any case we shouldn't hang
user presentations on architecture speicifc features).
```

### 2. Explain how to test it

1. Nothing should functionally change
2. Events will be emitted with new and simplified field types (BREAKING CHANGE)
3. Timestamps will now be normalized entirely in the decode stage


pkg/bufferdecoder/eventsreader.go
```diff
@@ -66,67 +34,66 @@ func readArgFromBuff(id events.ID, ebpfMsgDecoder *EbpfDecoder, params []trace.A
 		return 0, arg, errfmt.Errorf("invalid arg index %d", argIdx)
 	}
 	arg.ArgMeta = params[argIdx]
-	argType := GetParamType(arg.Type)
+	argType := GetDecodeType(arg.Type)
 
 	switch argType {
-	case u8T:
+	case trace.U8_T:
```
geyslan, line 40: How about reducing the size of `readArgFromBuff()` by replacing the switch/if branches with an already initialized array containing pointer functions for each type (ebpfMsgDecoder.Decode*)?

NDStrahilevitz: This is basically the planned next step. I plan to move the decoding mapping to a hash map similar to the processor and derive functions. In this way we, and plugin authors, can easily introduce new decoding strategies.

NDStrahilevitz: @geyslan I've opted not to make this step in this PR but rather in a followup so that we merge smaller chunks at a time as this is a large scale change.

pkg/ebpf/c/common/buffer.h
```diff
@@ -339,39 +352,35 @@ statfunc int save_sockaddr_to_buf(args_buffer_t *buf, struct socket *sock, u8 in
...
-
-#define ARG_TYPE_MAX_ARRAY (u8) TIMESPEC_T // last element defined in argument_type_e
+     (u64) 1 << U16_T | (u64) 1 << U8_T | (u64) 1 << UINT64_ARR_T | (u64) 1 << POINTER_T |         \
+     (u64) 1 << BYTES_T | (u64) 1 << STR_ARR_T | (u64) 1 << U8_T)
 
-// Ensure that only values that can be held by an u8 are assigned to sizes.
+// Ensure that only values that can be held by an u32 are assigned to sizes.
 // If the size is greater than 255, assign 0 (making it evident) and handle it as a special case.
-static u8 type_size_table[ARG_TYPE_MAX_ARRAY + 1] = {
+static u32 type_size_table[ARG_TYPE_MAX_ARRAY + 1] = {
```
yanivagman, line 368: Why do we need to change to u32?

pkg/ebpf/c/common/buffer.h
```diff
@@ -30,35 +31,46 @@ statfunc buf_t *get_buf(int idx)
...
 
-    // Read into buffer
-    if (bpf_probe_read(&(buf->args[buf->offset + 1]), size, ptr) == 0) {
-        // We update offset only if all writes were successful
-        buf->offset += size + 1;
-        buf->argnum++;
-        return 1;
-    }
+    // Read into buffer after the argument index
+    if (bpf_probe_read_kernel(&(buf->args[buffer_index_offset]), size, ptr) < 0)
```
yanivagman, line 65: How do you know that we are reading from a kernel pointer here?

pkg/ebpf/tracee.go
```diff
@@ -66,11 +66,11 @@ type Tracee struct {
...
 	sigEngine *engine.Engine
 	// Events
-	eventsSorter     *sorting.EventsChronologicalSorter
-	eventsPool       *sync.Pool
-	eventsParamTypes map[events.ID][]bufferdecoder.ArgType
-	eventProcessor   map[events.ID][]func(evt *trace.Event) error
-	eventDerivations derive.Table
+	eventsSorter       *sorting.EventsChronologicalSorter
+	eventsPool         *sync.Pool
+	eventArgumentTypes map[events.ID][]trace.DecodeAs
```
yanivagman, line 71: Arguments/Parameters should be replaced by data fields, which is the terminology used by the new event structure

yanivagman (changes requested):
Not sure what we are trying to achieve in this PR

pkg/ebpf/c/common/buffer.h
NDStrahilevitz: leaving at u8 introduced BTF errors. Moving to u32 moved the error to the verifier which I could eventually sort out. In addition I believe @geyslan suggested a move to u32 for some reason which I can't recall...

pkg/ebpf/c/common/buffer.h
NDStrahilevitz: This is used only in syscall argument extraction which I think is copied from the bpf context first. Anyway we can't be certain and there would be no harm done moving back to the generic probe_read.

pkg/ebpf/tracee.go
NDStrahilevitz: Renaming to `eventDataDecodeTypes` is ok for you?

NDStrahilevitz:
> Not sure what we are trying to achieve in this PR

The PR has two goals:
1. Reduce the type variance in data field declaration. We use concrete types instead, this is already achieved in the context of the below "decode types"
2. Decouple the single data type into a 1. decode strategy type (encoding the protocol between kernel and decoder, this should be a registerable method in the future) 2. presentation data type which encodes what kind of data the user will receive when reading the value (a float, an int, a string, some array, a timestamp, so on). 

Currently only the first goal is achieved.

pkg/ebpf/c/common/buffer.h
geyslan: When I implemented such an array, I was lucky enough to pass through the verifier with flying colours. But when you changed its elements, the verifer surely started complaining about misaligned accesses or even reverted to its familiar state of "not knowing how to count". Changing the size of the array elements to the word size would be a fix, as you did.

pkg/ebpf/c/common/buffer.h
NDStrahilevitz: This actually caused issues in certain syscalls so I reverted.

NDStrahilevitz:
@geyslan @rscampos As we keep introducing new event PRs I will have to rebase this PR over and over. Could we please close this PR before merging new events? I also don't want it to go stale and be lost over time.

geyslan:
@NDStrahilevitz rest assured. It will be the first handled as soon as we leave the code freeze.

geyslan:
@NDStrahilevitz ping me when you sync it main, then I'll start on it.

pkg/ebpf/c/common/buffer.h
```diff
@@ -376,8 +388,9 @@ statfunc int save_str_arr_to_buf(args_buffer_t *buf, const char __user *const __
 statfunc int save_args_str_arr_to_buf(
     args_buffer_t *buf, const char *start, const char *end, int elem_num, u8 index)
 {
-    // Data saved to submit buf: [index][len][arg_len][arg #][null delimited string array]
-    // Note: This helper saves null (0x00) delimited string array into buf
+    // Data saved to submit buf: [index][len][arg_len][arg #][array of null delimited string]
+    // Note: This helper saves null (0x00) delimited string array into buf in the format:
+    // [[str1][\n][str2][\n][...][strn][\n])]
```
geyslan, line 393: This comment is not so clear to me.

First doubt: Is the delimiter of this array elements a single NUL char or a NULL pointer?
Second doubt: Is the final delimiter of the array the same size of the elements delimiter?

pkg/ebpf/c/common/buffer.h
```diff
@@ -43,7 +44,7 @@ statfunc data_filter_lpm_key_t *get_string_data_filter_lpm_buf(int idx)
 }
 
 // biggest elem to be saved with 'save_to_submit_buf' should be defined here:
-#define MAX_ELEMENT_SIZE bpf_core_type_size(struct sockaddr_un)
+#define MAX_ELEMENT_SIZE sizeof(struct sockaddr_un)
```
geyslan, line 47: Why this change?

pkg/ebpf/c/common/buffer.h
```diff
@@ -82,29 +83,40 @@ statfunc int save_to_submit_buf(args_buffer_t *buf, void *ptr, u32 size, u8 inde
...
         return 0;
 
-    barrier();
-    if (buf->offset > ARGS_BUF_SIZE - 1)
+    u32 buffer_index_offset = buf->offset + 1; // buffer offset after writing the index
+    if (buffer_index_offset > ARGS_BUF_SIZE)
         return 0;
 
-    // Save argument index
+    // Save argument index (add 1 to index later)
```
geyslan, line 95: Where this increment is done?

geyslan (commented):
I did a first pass. It looks neat so far. I've put some questions.

tests/integration/capture_test.go
```diff
@@ -69,7 +69,7 @@ func Test_TraceeCapture(t *testing.T) {
 		},
 		{
 			name:           "capture packet context",
-			coolDown:       0 * time.Second,
+			coolDown:       2 * time.Second,
```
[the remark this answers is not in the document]
NDStrahilevitz: I don't know, maybe? Anyway it's likely best to keep with my change for good measures since we start full fledged tracee instances in these tests.

types/trace/trace.go
```diff
@@ -134,11 +134,105 @@ type Argument struct {
 	Value interface{} `json:"value"`
 }
 
+// DecodeAs is an enum that encodes the argument types that an
```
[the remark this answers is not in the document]
NDStrahilevitz: Legacy of the binary division. I think we could unify the packages once we rm -rf tracee-rules.

types/trace/trace.go
```diff
@@ -134,11 +134,105 @@ type Argument struct {
...
+const (
+	ARGS_ARR_T DecodeAs = iota + 128
+	BOOL_T
+	FLOAT_T
+	FLOAT64_T
+	MAX_TRACEE_DECODES
+)
+const (
+	USER_DEFINED_DEOCDE_BEGIN DecodeAs = iota + 256
+	USER_DEFINED_DEOCDE_END   DecodeAs = iota + 256
```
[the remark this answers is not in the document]
NDStrahilevitz: Nice catch

types/trace/trace.go
```diff
@@ -134,11 +134,105 @@ type Argument struct {
...
+)
+
+func (d DecodeAs) String() string {
+	// in the future register these along deocde strategies
+	switch d {
```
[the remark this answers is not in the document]
NDStrahilevitz: I think it would be best to make this change code-wide then no? It would also simplify all ours plans for plugin registration.

pkg/events/parse/params.go
```diff
@@ -28,74 +26,41 @@ func ArgVal[T any](args []trace.Argument, argName string) (T, error) {
...
-	case "s16",
-		"short":
+	case "int16":
 		return int16(0)
-	case "u16",
-		"unsigned short",
-		"old_gid_t",
-		"old_uid_t",
-		"umode_t":
+	case "uint16":
```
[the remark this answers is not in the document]
NDStrahilevitz: IIRC the problem with umode_t is that it has kernel dependent type aliases no? In any case I don't think we should shift our userspace events along with the internal kernel types - the whole point of this PR is decoupling the two. The author of the event should accommodate in his typing the edge case, most likely fit the largest possible size and submit the argument appropriately.

pkg/ebpf/c/common/buffer.h
NDStrahilevitz: Each string is size variable. The indication for the end of a string is just a null terminator char. You can know which \n is the final one using the arg_len provided beforehands.

pkg/ebpf/c/common/buffer.h
NDStrahilevitz: Probably a verifier issue at some point. I will check while rebasing if I can revert this.

pkg/ebpf/c/common/buffer.h
NDStrahilevitz: I think I meant `argnum` here as it should, under ideal conditions, correspond to the index.

pkg/ebpf/c/common/buffer.h
geyslan: So would it be like this?

`NUL = '\0'`

`[ char_arr1[n], NUL, char_arr2[n], NUL, ... char_arrn[n], NUL ]`

If so, consider it as a suggestion.

Another way, even more expressive might be:

```
    [ char_arr1[0]  char_arr1[1]  ...  char_arr1[n-1]  '\0',
      char_arr2[0]  char_arr2[1]  ...  char_arr2[n-1]  '\0',
      ...
      char_arrn[0]  char_arrn[1]  ...  char_arrn[n-1]  '\0' ]
```

NDStrahilevitz:
@geyslan Just rebased and adjusted the newly added events since I've originally pushed this PR.

geyslan (approved):
LGTM, put some sparse thoughts, be my guest to ignore them if not aligned with the effort.

pkg/bufferdecoder/eventsreader.go
```diff
@@ -196,53 +164,45 @@ func readArgFromBuff(id events.ID, ebpfMsgDecoder *EbpfDecoder, fields []trace.A
...
+	case "char*":
+		return trace.STR_T
 	case "const char*const*": // used by execve(at) argv and env
-		return strArrT
-	case "const char**": // used by sched_process_exec argv and envp
-		return argsArrT
-	case "const struct sockaddr*", "struct sockaddr*":
-		return sockAddrT
+		return trace.STR_ARR_T
+	case "const char**": // used by sched_process_exec argv and env
```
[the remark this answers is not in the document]
NDStrahilevitz: In the next PR im renaming these to go types. Presentation types are userfacing, they should be unrelated to C types. DecodeAs represents a strategy for transferring buffer data, so it should in some sense correlate to C. I will bring over the next PR into this one to make it a cohesive whole.

pkg/ebpf/c/common/buffer.h
```diff
@@ -455,39 +468,35 @@ statfunc int save_sockaddr_to_buf(args_buffer_t *buf, struct socket *sock, u8 in
 
 #define DEC_ARG(n, enc_arg) ((enc_arg >> (8 * n)) & 0xFF)
 
+// types whose arguments needs to be directly in type_size_table (arg = (void *) args->args[i])
 #define BITMASK_INDIRECT_VALUE_TYPES                                                               \
-    ((u64) 1 << STR_T | (u64) 1 << SOCKADDR_T | (u64) 1 << INT_ARR_2_T | (u64) 1 << TIMESPEC_T)
+    ((u64) 1 << INT_ARR_2_T | (u64) 1 << STR_T | (u64) 1 << SOCKADDR_T | (u64) 1 << TIMESPEC_T)
```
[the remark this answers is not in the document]
NDStrahilevitz: Ok, I can do this rename.

pkg/events/core.go
```diff
@@ -313,8 +313,8 @@ var CoreEvents = map[ID]Definition{
 		syscall: true,
 		sets:    []string{"syscalls", "fs", "fs_file_attr"},
 		fields: []trace.ArgMeta{
-			{Type: "const char*", Name: "pathname"},
-			{Type: "struct stat*", Name: "statbuf"},
+			{DecodeAs: trace.STR_T, Type: "string", Name: "pathname"},
+			{DecodeAs: trace.POINTER_T, Type: "void*", Name: "statbuf"},
```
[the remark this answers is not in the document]
NDStrahilevitz: I'm not sure about this. My purpose was to remove user facing c-types. In the end we are only presenting here a pointer, some memory address. The only useful reason I could see to bring it back is if for some reason the user wants to create a matching pointer in some other program.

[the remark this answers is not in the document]
geyslan: What I suggested it's just a comment with the correspondent type for debugging and stuff - a plus for the eyes. I agree with using the DecodeAs as you changed: a pointer.

pkg/bufferdecoder/eventsreader.go
```diff
@@ -13,40 +13,6 @@ import (
...
-	strArrT
-	sockAddrT
-	bytesT
-	u16T
-	credT
-	intArr2T
-	uint64ArrT
-	u8T
-	timespecT
-)
```
yanivagman, line 41: As mentioned in a previous commit comment, I think this enum (for that matter, DecodeAs enum is the same as this one) should be kept here since it is the business of the decoder, and not of any other part in the code, and more specifically' shuoldn't be part of the types packages

pkg/dnscache/dns_events_test.gz
yanivagman, line 1: how is this file removal related?

pkg/bufferdecoder/decoder.go
```diff
@@ -29,11 +31,61 @@ var ErrBufferTooShort = errors.New("can't read context from buffer: buffer too s
...
+	// add an identity function per decode type for its string presentation and an undefined type
+	for decode := range present {
+		present[decode][decode.String()] = identity
+		// note(nadav.str): this allows defining data fields with a type - should we allow it?
+		present[decode][""] = identity
+	}
+
+	return present
+}
+
```
yanivagman, line 88: I'm afraid performing this map lookup for every data field of every event may have performance implications. Did you check for that?

yanivagman (changes requested):

pkg/bufferdecoder/decoder.go
geyslan: I believe it can be converted to array, right @NDStrahilevitz?

NDStrahilevitz: At the second level I think we would inevitably have a map, but the first layer could be converted to a slice.

pkg/dnscache/dns_events_test.gz
NDStrahilevitz: Breaking changes to decode and encode. This file included json events for testing purposes, so they had to be regenerated.

types/trace/trace.go
```diff
@@ -135,11 +135,106 @@ type Argument struct {
 	Value interface{} `json:"value"`
 }
 
+// DecodeAs is an enum that encodes the argument types that an
+// eBPF program may write to the shared buffer. In practice they designate
+// either/or a type which is specifically encoded as an argument type in eBPF
+// or a strategy of passing an argument on the submission buffer via some
+// function.
+type DecodeAs uint16
```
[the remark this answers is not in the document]
NDStrahilevitz: Because I need this knowledge in the data field metadata. The only way I can think of bypassing this is switching to using a wrapper inside the event definitions which will be like `DataFieldWrapper{trace.ArgMeta, DecodeAs}`

pkg/bufferdecoder/decoder.go
```diff
@@ -13,12 +13,14 @@ import (
...
+	"github.com/aquasecurity/tracee/pkg/time"
 	"github.com/aquasecurity/tracee/types/trace"
 )
 
 type EbpfDecoder struct {
-	buffer []byte
-	cursor int
+	buffer  []byte
+	cursor  int
+	present map[trace.DecodeAs]map[string]func(any) (any, error)
```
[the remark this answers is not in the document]
NDStrahilevitz: It means here how data will be presented the moment it leaves decoding. The point of this PR is splitting the knowledge of how the data looks internally as decoded from whatever source, and how tracee and users consuming tracee will consume that data. So when decoding ends, it neatly packages it to something "presentable", notably it will provide timestamps as already normalized.

pkg/bufferdecoder/eventsreader.go
NDStrahilevitz: See above.

pkg/bufferdecoder/decoder.go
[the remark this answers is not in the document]
yanivagman: > It means here how data will be presented the moment it leaves decoding

So maybe simply call it "decode"?

types/trace/trace.go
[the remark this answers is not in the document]
yanivagman: So let's use a wrapper then. We should keep decode an internal business of Tracee and not expose it to the user

pkg/bufferdecoder/decoder.go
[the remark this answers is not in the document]
NDStrahilevitz: Sounds right actually. Maybe `decodeIntoType` is a bit more explicit to distinguish the byte->type decoding, since here its type->type.

NDStrahilevitz:
@yanivagman waiting for your review.
@geyslan or @rscampos feel free to give your take as well.

pkg/policy/ebpf.go
```diff
@@ -318,6 +318,15 @@ func (ps *policies) createNewEventsMapVersion(
...
 		fields := eventsFields[id]
+
+		/*
+			each event may have at most 8 argument data types stored.
+			in the map, the arguments are stored in a 64bit sized bitmap where each 8 bits
+			represent some argument type.
+			For example consider an event with two int arguments (argType = 1) and one string
+			(argType = 10):
+			0 0 0 0 0 a 1 1 - would be the encoded argument types bitmap
+		*/
```
geyslan, line 329: Just a rewrite suggestion, be my guest to ignore it.

```suggestion
		/*
			Each event can have up to 8 argument data types stored.
			These argument types are encoded in a 64-bit bitmap where:
			- Each byte (8 bits) represents a single argument type.
			- Therefore, a maximum of 8 argument types can be represented.

			For example, consider an event with:
			- Two integer arguments (argType = 1)
			- One string argument (argType = 10)

			The bitmap would be encoded as (in hex representation):
			00 00 00 00 00 0A 01 01 (0x00000000000A0101)

			Breaking down the bitmap from right to left (least significant byte first):

			| Byte Position | Argument Type          |
			|---------------|------------------------|
			| 0             | 0x01 (int argument)    |
			| 1             | 0x01 (int argument)    |
			| 2             | 0x0A (string argument) |
			| 3 - 7         | 0x00 (padding, unused) |

		*/
```

pkg/events/parse/params.go
```diff
@@ -43,74 +43,44 @@ func ArgVal[T any](args []trace.Argument, argName string) (T, error) {
...
 		return []string{}
 	case "bool":
 		return false
 	case "float":
 		return float32(0)
 	case "float64":
 		return float64(0)
-	case "slim_cred_t":
+	case "time.Time":
+		// TODO: is this the right choice? Maybe abuse the any and return int(0)?
```
geyslan, line 81: What's the caveat?

pkg/ebpf/processor_proctree.go
```diff
@@ -44,11 +45,13 @@ func (t *Tracee) procTreeForkProcessor(event *trace.Event) error {
 	if err != nil {
 		return err
 	}
-	forkFeed.ParentStartTime, err = parse.ArgVal[uint64](event.Args, "parent_process_start_time")
+	parentStartTime, err := parse.ArgVal[time.Time](event.Args, "parent_process_start_time")
```
geyslan, line 48: Maybe I overlooked this change in the rest of the PR, anyway, why the arg in this stage already has a time.Time instead of uint64?

pkg/bufferdecoder/eventsreader.go
```diff
@@ -367,9 +351,10 @@ func readStringFromBuff(ebpfMsgDecoder *EbpfDecoder) (string, error) {
 	return string(res), nil
 }
 
-// readStringVarFromBuff reads a null-terminated string from `buff`
-// max length can be passed as `max` to optimize memory allocation, otherwise pass 0
-func readStringVarFromBuff(decoder *EbpfDecoder, max int) (string, error) {
+// readVarStringFromBuffer reads a null-terminated string from the ebpf buffer where the size is not
+// known. The helper will read from the buffer char-by-char until it hits the null terminator
+// or the given max length. The cursor will then skip to the point in the buffer after the max length.
+func readVarStringFromBuffer(decoder *EbpfDecoder, max int) (string, error) {
```
geyslan, line 357: You already know that this is being touched in https://github.com/aquasecurity/tracee/pull/4634/files#diff-d9cde7aaf9961101ae53f694c6932d5943eceb87ed114291d51e88ec2d535969R376. I'm just referencing it for both of us (more for myself) to remember later.

geyslan (approved):
LGTM

Put some suggestions and doubts.

types/trace/trace.go
```diff
@@ -229,7 +235,7 @@ func (arg *Argument) UnmarshalJSON(b []byte) error {
 	var err error
 
 	switch arg.Type {
-	case "const char*const*", "const char**":
+	case "[]string":
```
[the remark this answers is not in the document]
NDStrahilevitz: 1. Didn't catch that error message, i'll realign
2. I don't think that would be relevant because we are unmarshalling a json text here. This arg.Type is just a type hint for a generic.

pkg/bufferdecoder/decoder.go
```diff
@@ -27,20 +30,73 @@ var ErrBufferTooShort = errors.New("can't read context from buffer: buffer too s
...
-// ReadAmountBytes returns the total amount of bytes that decoder has read from its buffer up until now.
-func (decoder *EbpfDecoder) ReadAmountBytes() int {
+// BytesRead returns the total amount of bytes that decoder has read from its buffer up until now.
+func (decoder *EbpfDecoder) BytesRead() int {
+	return decoder.cursor
+}
+
+// MoveCursor moves the buffer cursor over n bytes. Returns the new cursor position.
+func (decoder *EbpfDecoder) MoveCursor(n int) int {
+	decoder.cursor += n
```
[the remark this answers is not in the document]
NDStrahilevitz: Not an integer overflow, but there should probably be a bound check here.

pkg/events/parse/params.go
NDStrahilevitz: That the 0 time is the epoch beginning so it can seem slightly off when rendered.

pkg/ebpf/processor_proctree.go
NDStrahilevitz: Because a motivation of the PR was enabling "decode time processing". So arguments representing some timestamps can be outputted directly as timestamps, and more importantly, normalized. This should help avoid duplicate normalization in analyze mode.
So you can see i've changed the event definition respectively.

types/trace/trace.go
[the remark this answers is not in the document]
NDStrahilevitz: Done in the latest change.

pkg/bufferdecoder/decoder.go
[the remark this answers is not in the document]
NDStrahilevitz: Done in the latest change.

types/trace/trace.go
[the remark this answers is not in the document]
geyslan: > 2. I don't think that would be relevant because we are unmarshalling a json text here. This arg.Type is just a type hint for a generic.

I meant, could we change both ends to communicate `[]string` instead of `[]interface{}` (performance wise)?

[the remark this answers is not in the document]
NDStrahilevitz: Oh, I don't think thats an option. (go default) Json unmarshaling just works that way because we've defined argument values to be generic.

pkg/ebpf/processor.go
```diff
@@ -90,12 +90,12 @@ func (t *Tracee) registerEventProcessors() {
...
-			"parent_start_time",
-			"parent_process_start_time",
-			"leader_start_time",
-		))
+		// t.RegisterEventProcessor(events.SchedProcessFork, t.normalizeTimeArg(
+		// 	"start_time",
+		// 	"parent_start_time",
+		// 	"parent_process_start_time",
+		// 	"leader_start_time",
+		// ))
```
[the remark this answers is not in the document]
NDStrahilevitz: yep

geyslan:
@NDStrahilevitz are you merging it later this week?

pkg/events/data/decode.go
```diff
@@ -0,0 +1,78 @@
...
+	ARGS_ARR_T DecodeAs = iota + 128
+	BOOL_T
+	FLOAT_T
+	FLOAT64_T
+	MAX_TRACEE_DECODES
+)
+const (
+	USER_DEFINED_DECODE_BEGIN DecodeAs = iota + 256
+	USER_DEFINED_DECODE_END   DecodeAs = iota + 256
+)
```
yanivagman, line 46: DecodeAs is passed to bpf code encoded as u8, capable of encoding up to 256 types.
If we overflow the u8 being encoded, we may misinterpret this data in bpf code, won't we?

pkg/ebpf/processor_proctree.go
```diff
@@ -66,10 +69,11 @@ func (t *Tracee) procTreeForkProcessor(event *trace.Event) error {
 	if err != nil {
 		return err
 	}
-	forkFeed.LeaderStartTime, err = parse.ArgVal[uint64](event.Args, "leader_start_time")
+	leaderStartTime, err := parse.ArgVal[time.Time](event.Args, "leader_start_time")
```
yanivagman, line 72: @geyslan we need to make sure proctree functionality isn't affected with these changes (I remember we had issues with process id hashes not matching after making such changes previosuly). Do we have tests for that?

pkg/bufferdecoder/eventsreader.go
```diff
@@ -160,89 +133,97 @@ func readArgFromBuff(id events.ID, ebpfMsgDecoder *EbpfDecoder, fields []trace.A
...
-		return uint(argIdx), arg, errfmt.Errorf("error unknown arg type %v", argType)
+		return uint(argIdx), arg, errfmt.Errorf("error unknown arg type %v", decodeType)
 	}
 	if err != nil {
 		return uint(argIdx), arg, errfmt.WrapError(err)
 	}
-	arg.Value = res
+
+	// note(nadav.str): this allows defining data fields with a type - should we allow it?
+	if decodeType.String() == arg.Type || arg.Type == "" {
```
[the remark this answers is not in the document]
NDStrahilevitz: Yep. I'll rewrite the comment to fix the meaning.

pkg/bufferdecoder/decoder.go
```diff
@@ -27,20 +30,80 @@ var ErrBufferTooShort = errors.New("can't read context from buffer: buffer too s
...
 
+type presentorFunc func(any) (any, error)
+type TypeDecoder []map[string]presentorFunc
+
+func NewDataPresentor() TypeDecoder {
+	// type default presentor - handled as a special case in decoding
+	// if no specific presentor is specified
+	identity := func(a any) (any, error) {
+		return a, nil
+	}
```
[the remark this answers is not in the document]
NDStrahilevitz: leftover from when I autoassigned this. So yes, this can be inlined.

pkg/bufferdecoder/eventsreader.go
```diff
@@ -160,89 +133,97 @@ func readArgFromBuff(id events.ID, ebpfMsgDecoder *EbpfDecoder, fields []trace.A
...
+		if err != nil {
+			return uint(argIdx), arg, errfmt.WrapError(err)
+		}
+	}
+
 	return uint(argIdx), arg, nil
 }
 
-func GetFieldType(fieldType string) ArgType {
+func GetDecodeType(fieldType string) data.DecodeAs {
```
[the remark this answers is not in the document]
NDStrahilevitz: leftover from the rewrite. it can be removed.

pkg/ebpf/c/types.h
```diff
@@ -159,32 +159,31 @@ typedef struct args {
 } args_t;
 
 // NOTE: If any fields are added to argument_type_e, the array type_size_table
-// (and related defines) must be updated accordingly.
+// (and related defines) must be updated accordingly. Corresponds to the ArgType enum in
+// pkg/bufferdecoder/eventsreader.go.
```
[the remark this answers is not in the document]
NDStrahilevitz: leftover, will fix.

pkg/events/data/decode.go
NDStrahilevitz: That's a blunder. There are no userdefined decode types yet, but it should be aligned. I'll resection and document.

yanivagman (approved):
LGTM,
Please rebase on types PR before merging this one

Issue #4346, opened 2024-10-10
GetProtoHTTPByName for net_packet_http_request


HenrikWittemeier:
## Description
Hi,
im trying to write a go signature in which i need the HTTPRequest Header Data of the Event net_packet_http_request. 
I found the function GetProtoHTTPByName(eventObj, "http_proto") that works the event net_packet_http, but i found no way to do similar with the net_packet_http_request event.

Thanks for your help!

###Steps to reproduce

1. Write a Go Signature that uses the event net_packet_http_request and try to get the http_request field of the event.
2. Compile and run it 
3. See the error  `protocol HTTP: type error (should be trace.ProtoHTTP, is trace.ProtoHTTPRequest)"}"}]`

## Output of `tracee version`:

```
Tracee Version v0.22.0
```

## Output of `uname -a`:

```
Linux tracee-hqmbk 6.8.0-45-generic #45-Ubuntu SMP PREEMPT_DYNAMIC Fri Aug 30 12:02:04 UTC 2024 x86_64 GNU/Linux
```

rscampos:
Hello @HenrikWittemeier,

The `GetProtoHTTPByName` helper currently only works for the `net_packet_http event`, not for the `net_packet_http_request` event. While it is possible to retrieve the request header from `net_packet_http_request` using `GetTraceeArgumentByName`, we have decided to create a new helper, `GetProtoHTTPRequestByName`, to streamline this process.

The `GetProtoHTTPRequestByName` helper will function similarly to `GetProtoHTTPByName`, making it easier to handle HTTP requests for the `net_packet_http_request` event. Note that is necessary to use `http_request` instead of `http_proto`.
```
httpRequest, err := helpers.GetProtoHTTPRequestByName(eventObj, "http_request")
if err != nil {
	return err
}
```

PR #4354, opened 2024-10-16
feat(test): e2e integration test for new helpers

rscampos:
fix: #4346

f457c2bf6 **feat(test): e2e integration test for new helpers**

### 2. Explain how to test it

1. Duplicate the http.go signature file.
2. Update the event to `net_packet_http_request` or `net_packet_http_response`.
3. Use the appropriate helper: `GetProtoHTTPRequestByName` or `GetProtoHTTPResponseByName`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/Masterminds/sprig/v3 v3.2.3
 	github.com/aquasecurity/libbpfgo v0.7.0-libbpf-1.4.0.20240729111821-61d531acf4ca
 	github.com/aquasecurity/tracee/api v0.0.0-20241203172838-1f796cb64289
-	github.com/aquasecurity/tracee/signatures/helpers v0.0.0-20241009193135-0b23713fa9f9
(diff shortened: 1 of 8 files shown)


NDStrahilevitz:
@rscampos Maybe use these helpers in the existing signatures if possible? For example add themto the instrumentation tests. Then we can merge.

go.mod
```diff
@@ -43,6 +43,8 @@ require (
 	sigs.k8s.io/controller-runtime v0.18.2
 )
 
+replace github.com/aquasecurity/tracee/signatures/helpers => ./signatures/helpers/
```
NDStrahilevitz, line 46: Remove and update through types PR

NDStrahilevitz (approved):
LGTM, although we should probably use something similar to the flow events in DNS instead of these events (which are meant for legacy compatibility iirc).

NDStrahilevitz:
@rscampos Merge at will once you do the types PR (which you can merge without +1 as far as i'm concerned)

PR #4355, opened 2024-10-17
fix: print err when parseArgument() fails

geyslan:

dad7dae43 **fix: print err when parseArgument() fails**

```
When parseArgument() fails, error message should be printed to help
debugging.
```


geyslan:
We were having an issue with Lineage that was hard to catch without that error message.

NDStrahilevitz (approved):

PR #4357, opened 2024-10-18
Fix arg zero parse types and core typo

geyslan:

7eafda0b2 **fix(core): ctx_id* types of io_* events typo**

```
io_destroy, io_getevents, io_submit and io_cancel were using an
inexistent type for ctx_id.

The same for io_setup's ctx_idp parameter.
```

2a3f0dd4c **fix(parse): correct types sizes**

```
This considers the size of the types based on vmlinux.h.
```


NDStrahilevitz (approved):
LGTM. This will require some fixes in my other PR I guess...

PR #4366, opened 2024-10-28
feat(events): change log level in hooked_syscall

OriGlassman:
When unable to locate a syscall symbol, instead of printing an error and terminate the hook checker goroutine, be more graceful: print a warning and skip hook check only for the specific syscall

When unable to locate a syscall symbol, instead of printing an error and terminate the hook checker goroutine, be more graceful: print a warning and skip hook check only for the specific syscall

### 2. Explain how to test it
./tracee -e=hooked_syscall


yanivagman (approved):
LGTM

PR #4367, opened 2024-10-28
fix(events): check if init finished in hidden kernel module

OriGlassman:
On startup, there could be a case where a kernel module is being loaded before the hidden kernel module initialization function is called and finished.

On startup, there could be a case where a kernel module is being loaded before the hidden kernel module initialization function is called and finished.

### 2. Explain how to test it


yanivagman (approved):
LGTM

PR #4376, opened 2024-11-06
Refactor filter matching

yanivagman:

### DIFF
--- a/pkg/ebpf/c/types.h
+++ b/pkg/ebpf/c/types.h
@@ -294,43 +294,42 @@ typedef struct ksym_name {
 } ksym_name_t;
 
 typedef struct equality {
-    // bitmask with scopes on which a equal '=' filter is set
-    // its bit value will depend on the filter's equality precedence order
-    u64 equal_in_scopes;
-    // bitmask with scopes on which a filter equality is set
-    u64 equality_set_in_scopes;
+    // bitmap indicating which policies have a filter that uses the '=' operator (0 means '!=')
+    u64 equals_in_policies;
+    // bitmap indicating which policies have a filter that utilize the provided key
+    u64 key_used_in_policies;
 } eq_t;
 
 typedef struct policies_config {
-    // enabled scopes bitmask per filter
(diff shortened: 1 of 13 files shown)


geyslan (approved):
LGTM

pkg/ebpf/c/types.h
```diff
@@ -294,43 +294,43 @@ typedef struct ksym_name {
...
+    u64 uts_ns_filter_enabled;
+    u64 comm_filter_enabled;
+    u64 cgroup_id_filter_enabled;
+    u64 cont_filter_enabled;
+    u64 new_cont_filter_enabled;
+    u64 new_pid_filter_enabled;
+    u64 proc_tree_filter_enabled;
+    u64 bin_path_filter_enabled;
+    u64 follow_filter_enabled;
+    // bitmap indicating the default match behavior for policies
```
geyslan, line 319: What about explaining more `*_filter_default_match`? The enabled above is crystal, but default_match not so much.

pkg/ebpf/c/types.h
```diff
@@ -294,43 +294,43 @@ typedef struct ksym_name {
...
 typedef struct equality {
-    // bitmask with scopes on which a equal '=' filter is set
-    // its bit value will depend on the filter's equality precedence order
-    u64 equal_in_scopes;
-    // bitmask with scopes on which a filter equality is set
-    u64 equality_set_in_scopes;
+    // bitmap indicating which policies have a scope filter that uses the '=' operator (0 means
+    // '!=')
+    u64 equals_in_policies;
+    // bitmap indicating which policies have a scope filter that utilize the provided key
```
rscampos, line 300: Same here with word "scope".

pkg/ebpf/c/types.h
rscampos: good point @geyslan , maybe something like this if I understood correctly: `default behavior to match certain events even if no explicit key match occurs`

rscampos (approved):
LGTM

PR #4377, opened 2024-11-08
fix(engine): restrict finding feedback

NDStrahilevitz:
1. Findings were previously wrongly fedback to the engine even after an error in converting to events. This issue was resolved.
2. Finding feedbacks in general are now restricted by the engine config being set to single-binary mode (which now logically includes analyze mode).

### 2. Explain how to test it

tracee-rules shouldn't crash or emit errors when using a signature plugin that relies on signature feedback. The signature will not work.
In analyze the signature will work just as before.


NDStrahilevitz:
@yanivagman Confirmed this resolves our e2e issues.

yanivagman (approved):
LGTM

PR #4378, opened 2024-11-10
 fix hidden_kernel_module history scan for kernels >6.2

OriGlassman:

fix(events): fix hidden_kernel_module history scan for kernels >6.2
    
    - also address potential slice out of bounds for name argument in
      hidden_kernel_module.go

### 2. Explain how to test it


yanivagman:
@OriGlassman  is this WIP or something to review?
If only a test - please mark as draft or close this PR

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -747,7 +747,8 @@ enum
 
 struct modules_map {
     __uint(type, BPF_MAP_TYPE_HASH);
-    __uint(max_entries, MAX_NUM_MODULES);
+    __uint(max_entries, 2 * MAX_NUM_MODULES); // Do not decrease as it won't be big enough to hold
```
yanivagman, line 750: consider using another constant for this to be more clear, e.g MAX_MODULES_MAP_ENTRIES

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -970,12 +983,152 @@ statfunc int walk_mod_tree(program_data_t *p, struct rb_node *root, int idx)
...
+        lkm_seeker_send_to_userspace((struct module *) HISTORY_SCAN_SUCCESSFUL, &flags, &p);
+    } else {
+        tracee_log(ctx, BPF_LOG_LVL_WARN, BPF_LOG_ID_HID_KER_MOD, ret);
+        lkm_seeker_send_to_userspace((struct module *) HISTORY_SCAN_FAILURE, &flags, &p);
+    }
+
+    bpf_tail_call(ctx, &prog_array, TAIL_HIDDEN_KERNEL_MODULE_PROC);
+
+    return -1;
+}
```
yanivagman, line 1131: Why do we need 4 different programs that do the same thing?
Can't you use the context map (that you added) and use one program that tail calls itself recursively?

OriGlassman: Great point.. Changed that.

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -970,12 +1033,29 @@ statfunc int walk_mod_tree(program_data_t *p, struct rb_node *root, int idx)
...
-};
+    if (loop_result == HID_MOD_COMPLETED_ITERATIONS) {
+        flags = HISTORY_SCAN_FINISHED;
+        lkm_seeker_send_to_userspace((struct module *) HISTORY_SCAN_SUCCESSFUL, &flags, &p);
+        bpf_tail_call(ctx, &prog_array, TAIL_HIDDEN_KERNEL_MODULE_PROC);
+    } else if (loop_result == HID_MOD_RACE_CONDITION || iteration_num == MOD_TREE_LOOP_DEPTH) {
+        flags = HISTORY_SCAN_FINISHED;
+        tracee_log(ctx, BPF_LOG_LVL_WARN, BPF_LOG_ID_HID_KER_MOD, loop_result ^ iteration_num);
+        lkm_seeker_send_to_userspace((struct module *) HISTORY_SCAN_FAILURE, &flags, &p);
+        bpf_tail_call(ctx, &prog_array, TAIL_HIDDEN_KERNEL_MODULE_PROC);
```
yanivagman, line 1046: This line can be extracted out of these if-else statements

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -786,6 +797,7 @@ u64 last_module_insert_time = 0;
 bool hidden_old_mod_scan_done = false;
 static const int HID_MOD_RACE_CONDITION = -1;
 static const int HID_MOD_UNCOMPLETED_ITERATIONS = -2;
+static const int HID_MOD_COMPLETED_ITERATIONS = 0;
```
yanivagman, line 800: I don't see this variable is being assigned any value different than 0, why not make it a constant then?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -763,6 +765,15 @@ struct new_module_map {
 
 typedef struct new_module_map new_module_map_t;
 
+struct module_context_map {
+    __uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+    __uint(max_entries, 3);
+    __type(key, u32);
+    __type(value, u64);
+} module_context_map SEC(".maps");
```
yanivagman, line 773: It will be more efficient to define the map value to be a struct that contains node addr, idx, and iteration num.
By doing that, only one map lookup will be required to access all three, and less code will be required both for getting and setting these values

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -970,12 +1033,29 @@ statfunc int walk_mod_tree(program_data_t *p, struct rb_node *root, int idx)
...
-    struct latch_tree_root root;
-};
+    if (loop_result == HID_MOD_COMPLETED_ITERATIONS) {
+        flags = HISTORY_SCAN_FINISHED;
+        lkm_seeker_send_to_userspace((struct module *) HISTORY_SCAN_SUCCESSFUL, &flags, &p);
+        bpf_tail_call(ctx, &prog_array, TAIL_HIDDEN_KERNEL_MODULE_PROC);
+    } else if (loop_result == HID_MOD_RACE_CONDITION || iteration_num == MOD_TREE_LOOP_DEPTH) {
+        flags = HISTORY_SCAN_FINISHED;
+        tracee_log(ctx, BPF_LOG_LVL_WARN, BPF_LOG_ID_HID_KER_MOD, loop_result ^ iteration_num);
+        lkm_seeker_send_to_userspace((struct module *) HISTORY_SCAN_FAILURE, &flags, &p);
```
yanivagman, line 1045: Do we really need both a warning and sending the event to userspace?

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1221,19 +1314,18 @@ int lkm_seeker_mod_tree_tail(struct pt_regs *ctx)
...
-    // Report to userspace that the history scan finished successfully
-    lkm_seeker_send_to_userspace((struct module *) HISTORY_SCAN_SUCCESSFUL, &flags, &p);
+    // if (ret == 0) {
+    //     lkm_seeker_send_to_userspace((struct module *) HISTORY_SCAN_SUCCESSFUL, &flags, &p);
+    // } else if (ret == HID_MOD_UNCOMPLETED_ITERATIONS) { // Continue doing more iterations
+    //     bpf_tail_call(ctx, &prog_array, TAIL_HIDDEN_KERNEL_MODULE_MODTREE_1);
+    // } else {
+    // tracee_log(ctx, BPF_LOG_LVL_WARN, BPF_LOG_ID_HID_KER_MOD, ret);
+    // lkm_seeker_send_to_userspace((struct module *) HISTORY_SCAN_FAILURE, &flags, &p);
+    // }
```
yanivagman, line 1326: If this code is not needed anymore - let's remove it

yanivagman (changes requested):

pkg/ebpf/c/tracee.bpf.c
OriGlassman: In case the if else statements are not fulfilled, it updates the map and does another recursive tail call.
If I extract this tail call (Line 1046) out of the if-else statement, there won't be the map update and further recursive call).

pkg/ebpf/c/tracee.bpf.c
OriGlassman: It is a constant, isn't it?

pkg/ebpf/c/tracee.bpf.c
OriGlassman: The event must be submitted to userspace to mark that the scan finished.
I could remove the tracee_log and log it in userspace - but it will lose the context of where in the code it happened etc. Passing this information to userspace sounds like an hassle to me where there's the tracee_log API

pkg/ebpf/c/tracee.bpf.c
yanivagman: What do you mean by "not fulfilled"?
I was refering to the fact that both the if block and else block contains this same tail call at the end of the block, so I don't see a reason why not to take it out:
```
   if (loop_result == HID_MOD_COMPLETED_ITERATIONS) {
        flags = HISTORY_SCAN_FINISHED;
        lkm_seeker_send_to_userspace((struct module *) HISTORY_SCAN_SUCCESSFUL, &flags, &p);
    } else if (loop_result == HID_MOD_RACE_CONDITION || iteration_num == MOD_TREE_LOOP_DEPTH) {
        flags = HISTORY_SCAN_FINISHED;
        tracee_log(ctx, BPF_LOG_LVL_WARN, BPF_LOG_ID_HID_KER_MOD, loop_result ^ iteration_num);
        lkm_seeker_send_to_userspace((struct module *) HISTORY_SCAN_FAILURE, &flags, &p);
    }
bpf_tail_call(ctx, &prog_array, TAIL_HIDDEN_KERNEL_MODULE_PROC);
```

OriGlassman: As said, in that case the code highlighted below won't happen (note that it's not if-else.. it's if and else if):
image

yanivagman (approved):
LGTM

PR #4381, opened 2024-11-10
feat(events): add chmod_common event

OriGlassman:


yanivagman:
We don't backport new features into v0.22 branch, only bug fixes

PR #4384, opened 2024-11-11
fix: optimize proctree memory consumption

geyslan:
Close: #4409

f8963e007 **chore(proctree): lower process default cache size**
4a5bb5d0f **chore(proctree): del Process interpreter FileInfo**
d02ce615c **chore(changelog): reduce mem footprint**
9823a73b9 **chore(changelog): benchmark changelog.Set()**
6b805bbc1 **chore(utils): add PrintStructSizes() helper**

f8963e007 **chore(proctree): lower process default cache size**

```
Adjust the default the process cache size to 16,384, setting a 2:1 ratio
between thread and process caches.

With GOGC=5 and both caches stressed, was observed an average heap usage
of 77MB.
```

4a5bb5d0f **chore(proctree): del Process interpreter FileInfo**

```
Remove the interpreter FileInfo fields from the Process struct.
This reduces the memory footprint of the Proctree by 8MB at average.
```

d02ce615c **chore(changelog): reduce mem footprint**

```
The current Changelog structure consumes a significant amount of memory
due to the allocation of metadata for each field/instance. As the number
of fields increases, the memory usage grows linearly. Approximately 240
bytes per field were observed just for metadata, excluding the actual
data and pointers for each field.

To reduce memory consumption, the new changelog.Changelog[T] type was
created.

Changelog[T] implementation uses a single slice to store changes for
a single field of type T. This approach is more memory-efficient than
the previous implementation, since the map is no longer used. Its main
advantage is achievable by setting T as a complete structure type, which
gets rid of the need for metadata of each field.

Avoid distributing Changelog[t] across multiple fields, as this would
negate the purpose of reducing memory usage by increasing the metadata
associated with each Changelog. However, one might consider using it as
structure fields in scenarios where a mutable field (with many entries
and high frequency of change) has a larger size than the sum of the
Changelog metadata overhead and the size of the outer data structure
itself.

FileInfo and TaskInfo were modified to use the new Changelog[T] being
T the respective feed structures.

---

| Caches | GOGC | Branch | *Heap Use | *Heap   | Diff of | Proctree |
|        |      |        | (Avg)     | Growth  | main    |          |
|--------|------|--------|--------- -|------- -|---------|----------|
| -      | 5    | main   | 18        | -       | -       | off      |
| 32768  | 5    | main   | 209       | 191     | -       | on       |
|---------------------------------- --------- ----------------------|
| -      | 5    | new    | 18        | -       | -       | off      |
| 32768  | 5    | new    | 96        | 78      | -59.16% | on       |

* in MB

With GOGC set to 5, the new implementation reduces average heap usage by
approximately 59% when using cache sizes of 32,768.

The "Heap Use" and "Heap Growth" columns serve as a good indicator of
memory consumption and can assist in determining optimal cache sizes.

---

The Set method was hugely improved, reducing the number of allocations,
memory usage, and execution time. The benchmark results are as follows:

-run=^$ -tags ebpf -bench ^Benchmark_Set$
github.com/aquasecurity/tracee/pkg/changelog -benchtime=10000000x

goos: linux
goarch: amd64
pkg: github.com/aquasecurity/tracee/pkg/changelog

Benchmark_Set/All_Scenarios-32 10000000  322.8 ns/op  112 B/op  3 allocs/op
Benchmark_Set/Within_Limit-32  10000000  506.8 ns/op  496 B/op  5 allocs/op
PASS
ok      github.com/aquasecurity/tracee/pkg/changelog    367.236s

| Scenario      | ns/op. (%)   | B/op Reduc. (%)   | Alloc. Reduc. (%) |
|:--------------|:-------------|:------------------|:------------------|
| All Scenarios | -75.9%       | -88.7%            | -92.9%            |
| Within Limit  | -76.2%       | -85.5%            | -91.7%            |
```

9823a73b9 **chore(changelog): benchmark changelog.Set()**

```
-run=^$ -tags ebpf -bench ^Benchmark_Set$
github.com/aquasecurity/tracee/pkg/changelog -benchtime=10000000x

goos: linux
goarch: amd64
pkg: github.com/aquasecurity/tracee/pkg/changelog
Benchmark_Set/All_Scenarios-32 10000000  1337 ns/op   992 B/op  42 allocs/op
Benchmark_Set/Within_Limit-32  10000000  2130 ns/op  3424 B/op  60 allocs/op
PASS
ok      github.com/aquasecurity/tracee/pkg/changelog    350.043s
```

### DIFF
--- a/docs/docs/advanced/data-sources/builtin/process-tree.md
+++ b/docs/docs/advanced/data-sources/builtin/process-tree.md
@@ -14,7 +14,7 @@ The underlying structure is populated using the core `sched_process_fork`, `sche
 
 > Introducing this secondary event source is strategic: it reduces interference with actively traced events, leading to more accurate and granular updates in the process tree.
 
-The number of processes retained in the tree hinges on cache size. We have two separate caches at play: one for processes and another for threads. Both default to a size of 32K, supporting tracking for up to 32,768 processes and the same number of threads. It's worth noting that these are LRU caches: once full, they'll evict the least recently accessed entries to accommodate fresh ones.
+The number of processes retained in the tree hinges on cache size. We have two separate caches at play: one for processes and another for threads. The default cache size for processes is 16K, supporting tracking for up to 16,384 processes, while the thread cache is 32K, supporting tracking for up to 32,768 threads. On average, a configuration ratio of 2:1 (thread:cache) is defined, as one thread is created for every process. It's worth noting that these are LRU caches: once full, they'll evict the least recently accessed entries to accommodate fresh ones.
 
 The process tree query the procfs upon initialization and during runtime to fill missing data:
 * During initialization, it runs over all procfs to fill all existing processes and threads
@@ -34,7 +34,7 @@ Example:
       signals      | process tree is built from signals.
       both         | process tree is built from both events and signals.
   --proctree process-cache=8192   | will cache up to 8192 processes in the tree (LRU cache).
-  --proctree thread-cache=4096    | will cache up to 4096 threads in the tree (LRU cache).
+  --proctree thread-cache=16384   | will cache up to 16384 threads in the tree (LRU cache).
   --proctree process-cache-ttl=60 | will set the process cache element TTL to 60 seconds.
   --proctree thread-cache-ttl=60  | will set the thread cache element TTL to 60 seconds.
   --proctree disable-procfs-query | Will disable procfs quering during runtime
--- a/pkg/proctree/proctree.go
+++ b/pkg/proctree/proctree.go
@@ -35,7 +35,7 @@ import (
 //
 
 const (
-	DefaultProcessCacheSize = 32768
+	DefaultProcessCacheSize = 16384
 	DefaultThreadCacheSize  = 32768
 	DefaultProcessCacheTTL  = time.Second * 120
 	DefaultThreadCacheTTL   = time.Second * 120


pkg/proctree/process.go
```diff
@@ -22,22 +22,15 @@ type Process struct {
...
 	return &Process{
 		processHash: hash,
 		parentHash:  0,
 		info:        NewTaskInfo(),
-		executable:  NewFileInfo(executableChangelogSize),
-		interpreter: NewFileInfo(interpreterChangelogSize),
-		interp:      NewFileInfo(interpChangelogSize),
+		executable:  NewFileInfo(),
+		interpreter: NewFileInfo(),
+		interp:      NewFileInfo(),
```
geyslan, line 33: This also turns interpreters fields on, previously disabled by: https://github.com/aquasecurity/tracee/pull/4058

So, the heap reduction already considers the interpreter fields enabled again.

geyslan:
@trvll FYI.

NDStrahilevitz:
I will give this a full review tomorrow. I would like to ask beforehand why we gave up on generic comparable types entirely? Was there a significant memory footprint deriving from that usage?

geyslan:
> I would like to ask beforehand why we gave up on generic comparable types entirely?

Currently I can't see any specific use case for generic implementation.

> Was there a significant memory footprint deriving from that usage?

Before tackling this PR I've tested replacing all generic changelog fields with primitive ones (no-changelog branch). You can notice a huge decrement of heap in use after a stress that don't force changelog store more than one change (in other words, that stores an unique changelog value per field).

image

geyslan:
@NDStrahilevitz 

> Currently I can't see any specific use case for generic implementation.

I've got a considerable reduction by using generics, so hold on. Soon it will be pushed.

pkg/changelog/changelog.go
```diff
@@ -1,224 +1,204 @@
...
-// We need to find the last change that occurred before or exactly at the targetTime. The binary
-// search loop finds the position where a new entry with the targetTime timestamp would be inserted
-// to maintain chronological order:
-//
-// This position is stored in "left".
-//
-// So, to get the last entry that occurred before the targetTime, we need to access the previous
-// position, which is left-1.
+// Entries is the main structure that manages a list of changes (entries).
+// It keeps track of the configured flags for each kind of entry and the list of recorded entries.
```
NDStrahilevitz, line 31:
```suggestion
// It keeps track of specifically configured members indicated by MemberKind identifiers.
// When instantiating an Entries struct, one must supply a relevant mapping between the desired unique members
// and the maximum amount of changes that member can track.
```

pkg/changelog/changelog.go
```diff
@@ -1,224 +1,204 @@
...
-		return returnZero[T]()
-	}
-
-	return clv.changes[len(clv.changes)-1].value
+// ATTENTION: You should use Entries within a struct and provide methods to access it,
+// coordinating access through your struct mutexes. DO NOT EXPOSE the Entries object directly to
+// the outside world as it is not thread-safe.
+type Entries[T comparable] struct {
+	entryFlagsMap EntryFlagsMap // Configuration map defining flags for each member kind.
+	entries       []entry[T]    // List of recorded entries.
```
NDStrahilevitz, line 38: I am not sure at what length of entries this would be less optimal than a map. I guess you have made a comparison between this and (intuitively) a map from MemberKind to a time ordered slice? 

Alternatively, instead of a map, since we know at instantiation time the exact count of MemberKind, I think we can create a slice with known access indices per MemberKind (and I think these can just be the enum id). Then each index points to the entry array. This would also mean that we can remove the MemberKind metadata from the `entry[T]` object itself.

pkg/changelog/changelog.go
```diff
@@ -1,224 +1,204 @@
...
+			indexes = append(indexes, idx)
 		}
-		if !clv.changes[index].timestamp.Equal(targetTime) { // sanity check only (time exists already)
-			logger.Debugw("changelog internal error: timestamp mismatch")
+	}
+
+	// if there are entries for kind check if the last entry has the same value
+	if len(indexes) > 0 {
+		lastIdx := indexes[len(indexes)-1]
+		if e.entries[lastIdx].value == value && t.After(e.entries[lastIdx].t) {
```
NDStrahilevitz, line 79: This is basically a case of keeping the timestamp warm. Is this a documented need?

pkg/changelog/changelog.go
```diff
@@ -1,224 +1,204 @@
...
-	// Mark the timestamp as set.
-	clv.timestamps[targetTime] = struct{}{}
-
-	clv.enforceSizeBoundary()
+
+	//
+	// as there is no space, replace an entry
+	//
+
+	replaceIdx := indexes[len(indexes)-1] // index to replace
```
NDStrahilevitz, line 111:
```suggestion
	replaceIdx := indexes[len(indexes)-1] // default index to replace
```

NDStrahilevitz (commented):
I think I have a very access possible optimization to add, and some general questions and comment suggestions.

Overall while I don't like removing the explicit member references, this will be much more memory efficient and at least we kept type safety. Well done.

pkg/proctree/taskinfo.go
```diff
@@ -27,38 +27,53 @@ type TaskInfoFeed struct {
...
-	uid         *ch.Changelog[int]    // variable (process uid can be changed)
-	gid         *ch.Changelog[int]    // variable (process gid can be changed)
-	startTimeNS uint64                // this is a duration, in ns, since boot (immutable)
-	exitTimeNS  uint64                // this is a duration, in ns, since boot (immutable)
-	mutex       *sync.RWMutex
+	tid            int                        // immutable
+	pid            int                        // immutable
+	nsTid          int                        // immutable
+	nsPid          int                        // immutable
+	startTimeNS    uint64                     // this is a duration, in ns, since boot (immutable)
```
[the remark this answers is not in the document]
geyslan: Processes and Threads are created by hash, so even if the pid/tid is equal, the timestamp won't be. 

https://github.com/aquasecurity/tracee/blob/aa615f307aaba096b1a66dd7669c525b2260f1d3/pkg/proctree/proctree.go#L208

https://github.com/aquasecurity/tracee/blob/aa615f307aaba096b1a66dd7669c525b2260f1d3/pkg/proctree/proctree_procfs.go#L210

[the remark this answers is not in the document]
NDStrahilevitz: Right, it's a composite key after all. My bad :)

pkg/changelog/changelog.go
geyslan: > I am not sure at what length of entries this would be less optimal than a map. I guess you have made a comparison between this and (intuitively) a map from MemberKind to a time ordered slice?

As this effort focused on reducing the memory footprint, there's a chance that after some size the slice becomes less optimized than a map. But I believe it should be a larger slice, since the hash computation of the map will be more expensive than an iteration which don't trigger a cache miss. The benchmark that has a test case with 15 entries is verifiable anyway, so if we need in the future to use Entries to hold so many changes, we should start by creating incremental size test cases to detect the threshold.

> Alternatively, instead of a map, since we know at instantiation time the exact count of MemberKind, I think we can create a slice with known access indices per MemberKind (and I think these can just be the enum id). Then each index points to the entry array. This would also mean that we can remove the MemberKind metadata from the entry[T] object itself as its now defined through the access index.

This sounds promising, give me some time to check it. 👍🏼

pkg/changelog/changelog.go
geyslan: Actually this is a shortcircuit to avoid the previous coalesce logic. Instead of always inserting the value, even if it's equal to later enforce the boundaries and remove the old ones, the shortcircuit always checks if it's an update or an insertion always keeping the slice tidy.

pkg/changelog/changelog.go
geyslan: We can convert it to a slice, but MemberKind metadata is required to set and get the entry of specific kind. Recall that the `entries` is flat for all member kinds.

NDStrahilevitz: I'm suggesting a slice of slices. so the subslice is already accessed under the assumption of some MemberKind which is guaranteed by the access index.

geyslan: It works, but doing that we will increase again the metadata in one pointer (8 bytes) + slice head (24 bytes) per kind and certainly we'll have more cache misses.

geyslan: We continued this discussion offline and as @NDStrahilevitz suggested I improved it to use slices instead of maps in the Entries creation what improved the cpu time. That's really good. Tks for the tip, Nadav.

I've benchmarked as well using slice of slices (inner as a member kind one) instead of a single flat. It indeed improved cpu time but did increase the memory footprint as well. Said that, for our specific use case, this current implementation serve us well enough for now.

NDStrahilevitz (approved):
LGTM. Adding from our internal discussions that we can consider the slice-of-slices optimization in the future, in case we want to reduce CPU time here (at the cost of ~12mb from your tests).

geyslan:
> LGTM. Adding from our internal discussions that we can consider the slice-of-slices optimization in the future, in case we want to reduce CPU time here (at the cost of ~12mb from your tests). Nice work!

For future reference, this is the slice of slice draft:

https://github.com/geyslan/tracee/commit/85ab679145f536646048d77a7b486de9d39ae804

pkg/changelog/changelog.go
geyslan: https://github.com/aquasecurity/tracee/pull/4384#issuecomment-2515377613

geyslan:
We decided to create inner helpers to Task and File objects to wrap the entries use in more one level - it will make it more readable. Changes coming soon.

pkg/changelog/changelog.go
```diff
@@ -6,219 +6,202 @@ import (
...
+// It keeps track of specifically configured members indicated by MemberKind identifiers.
+// When instantiating an Entries struct, one must supply a relevant mapping between the desired
+// unique members and the maximum amount of changes that member can track.
+//
+// ATTENTION: You should use Entries within a struct and provide methods to access it,
+// coordinating access through your struct mutexes. DO NOT EXPOSE the Entries object directly to
+// the outside world as it is not thread-safe.
+type Entries[T comparable] struct {
+	entryFlags []MaxEntries // Configuration slice defining flags for each member kind.
+	entries    []entry[T]   // List of recorded entries.
```
yanivagman, line 38: entries is tightly coupled to entryFlags. This can be simplified to:
```
type Entries[T comparable] struct {
    entries []struct {
        maxEntries MaxEntries
        entries    []entry[T]
    }
}
```

Also consider using slice of slices and remove `MemberKind` from `entry`

yanivagman (changes requested):

pkg/changelog/changelog.go
```diff
@@ -6,219 +6,202 @@ import (
...
+// Entries is the main structure that manages a list of changes (entries).
+// It keeps track of specifically configured members indicated by MemberKind identifiers.
+// When instantiating an Entries struct, one must supply a relevant mapping between the desired
+// unique members and the maximum amount of changes that member can track.
+//
+// ATTENTION: You should use Entries within a struct and provide methods to access it,
+// coordinating access through your struct mutexes. DO NOT EXPOSE the Entries object directly to
+// the outside world as it is not thread-safe.
+type Entries[T comparable] struct {
+	entryFlags []MaxEntries // Configuration slice defining flags for each member kind.
```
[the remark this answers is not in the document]
geyslan: For providing information about the field itself: https://github.com/aquasecurity/tracee/pull/4384/commits/1d22a3c1702f1d7dbe2d90df0cc514aea375b96a#diff-46cf73873bd7fa0d747d9bc7ea904019c47f4ebeaacd3a01382cd8623b479e2eR13-R15

I don't see in this specific case another reason.

pkg/changelog/changelog.go
geyslan: Ok, I'm right now implementing the changes above in a linked list. Soon gonna get results after stress.

```go
type entry[T comparable] struct {
	time  time.Time // Timestamp of the change.
	value T         // Value of the change.
	prev  *entry[T] // Pointer to the previous entry in the linked list.
}

type entries[T comparable] struct {
	maxEntries MaxEntries // Maximum number of entries for this kind.
	size       uint8      // Current number of entries for this kind.
	tail       *entry[T]  // Tail pointer for the linked list of this kind.
}
```

pkg/changelog/changelog.go
```diff
@@ -6,219 +6,229 @@ import (
...
+	timestamp time.Time // Timestamp of the change.
+	value     T         // Value of the change.
+	prev      *entry[T] // Pointer to the previous entry in the linked list.
+}
+
+// entryList is an internal structure that manages a list of changes (entries) for a specific kind of entry.
+type entryList[T comparable] struct {
+	maxEntries MaxEntries // Maximum number of entries.
+	size       uint8      // Current number of entries.
+	tail       *entry[T]  // Tail pointer for the linked list of entries.
```
yanivagman, line 202: Why don't we simply use a slice of entry[T] instead?

NDStrahilevitz: That was my suggestion to Geyslan to potentially overcome the RSS cost. By using linked lists instead of arrays we maintain the same struct layout as before (switching the memberkind for the pointer), but get rid of the slice header. considering the minimal sizes of theharrays we have anyway, linked lists are likely just as good, since we did linear search in the arrays anyway (binary probably would be a very minor imporvement).
In particular the linked lists may be better in terms of memory because we have so little entries per list.

geyslan: That's right. Using node pointers we get rid of 16 extra bytes of slice metadata (len and cap). That was a great idea.

yanivagman: I don't think I understand why linked lists will be better with respect to memory.
Using linked list we will be able to remove the prev pointer inside the entry, so an entry will look like this:
```
type entry[T comparable] struct {
	timestamp time.Time // Timestamp of the change.
	value     T         // Value of the change.
}
```
This is 8 bytes per each entry. Since entry is the most inner struct and you can have more than 1 entry in the list, it can easily become 16 or even 24 extra bytes if we consider 3 entries in the list.
In addition to that, we will also be able to remove this extra "size" variable, which is part of the slice metadata. This code will be more elegant, and I think that in most cases won't be less good with respect to memory. I suggest the following struct to be used:
```
type Entries[T comparable] struct {
    entries []struct {
        maxEntries MaxEntries
        entries    []entry[T]
    }
}

type entry[T comparable] struct {
	timestamp time.Time // Timestamp of the change.
	value     T         // Value of the change.
}
```

geyslan: Ok, handling this today. Soon I'll be bringing more data.

geyslan: Ok, after analysis, we opted for slice of slices. Changes pushed again.

pkg/changelog/changelog.go
```diff
@@ -6,219 +6,221 @@ import (
...
 
-// The changelog package provides a changelog data structure. It is a list of changes, each with a
-// timestamp. The changelog can be queried for the value at a given time.
+// Public
 
-// ATTENTION: You should use Changelog within a struct and provide methods to access it,
-// coordinating access through your struct mutexes. DO NOT EXPOSE the changelog object directly to
-// the outside world as it is not thread-safe.
+// NewEntries initializes a new `Entries` structure using the provided flags.
+func NewEntries[T comparable](flags []MaxEntries) *Entries[T] {
```
yanivagman, line 60: flags is confusing IMO, consider renaming to `maxEntries`

pkg/proctree/process.go
yanivagman: I think we can remove these two fields for now - this should save a few extra bytes.
We can add them in the future when required.

pkg/proctree/fileinfo.go
```diff
@@ -21,243 +22,379 @@ type FileInfoFeed struct {
...
-	path      *ch.Changelog[string] // file path
-	dev       *ch.Changelog[int]    // device number of the file
-	ctime     *ch.Changelog[int]    // creation time of the file
-	inode     *ch.Changelog[int]    // inode number of the file
-	inodeMode *ch.Changelog[int]    // inode mode of the file
-	mutex     *sync.RWMutex
+	mutableStrings *changelog.Entries[string]
+	mutableInts    *changelog.Entries[int]
+	mutex          *sync.RWMutex
 }
```
yanivagman, line 58: The whole fields of FileInfo are tightly coupled - they all change together (unlike taskInfo) - so what do you think about putting the whole FileInfo struct under one changelog.Entries[FileInfo]? This may reduce memory a bit more plus keeping the original field names of FileInfo for direct access (so we can remove specific setters and getters for each one of them)

pkg/proctree/taskinfo.go
```diff
@@ -27,343 +27,418 @@ type TaskInfoFeed struct {
...
-		mutex:  &sync.RWMutex{},
+		mutableStrings: changelog.NewEntries[string](taskInfoMutableStringsFlags),
+		mutableInts:    changelog.NewEntries[int](taskInfoMutableIntsFlags),
+		mutex:          &sync.RWMutex{},
 	}
 }
 
 // NewTaskInfoFromFeed creates a new task with values from the given feed.
-func NewTaskInfoFromFeed(feed TaskInfoFeed) *TaskInfo {
+func NewTaskInfoNewFromFeed(feed TaskInfoFeed) *TaskInfo {
```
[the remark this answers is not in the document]
geyslan: Oh, thanks, that's definitely a typo.

pkg/proctree/fileinfo.go
NDStrahilevitz: Just mentioning that in this case flat slice/linked list will almost certainly be better since a slice of slices will be a redundancy (there is only one memberkind). Maybe there should a be a version of entries for a single memberkind usecase.

geyslan: I implemented it in #4427.

pkg/proctree/process.go
geyslan: Done, we got extra 8MB with 32k cache size (both): https://github.com/aquasecurity/tracee/pull/4384/commits/8f80d65132b0fba11c60dbcac5685efbcc047e4b

pkg/proctree/fileinfo.go
geyslan: Got it back here.

pkg/changelog/entry.go
```diff
@@ -0,0 +1,146 @@
+package changelog
+
+import "time"
+
+// entry is an internal structure representing a single change in the entryList.
+// It includes the timestamp and the value of the change.
+type entry[T comparable] struct {
+	tsUnixNano int64 // timestamp of the change (nanoseconds since epoch)
```
geyslan, line 8: Replaced time.Time with int64 (UnixNano). I believe that's a safe change since it's a internal field for checking just time passing diffs. It got us ~4MB extra bytes (proctree with 32k elements in both caches).

yanivagman (approved):
LGTM, good work @geyslan 
Let's just remove changelog kind if it is not currently used

docs/docs/advanced/data-sources/builtin/process-tree.md
```diff
@@ -34,7 +34,7 @@ Example:
       signals      | process tree is built from signals.
       both         | process tree is built from both events and signals.
   --proctree process-cache=8192   | will cache up to 8192 processes in the tree (LRU cache).
-  --proctree thread-cache=4096    | will cache up to 4096 threads in the tree (LRU cache).
+  --proctree thread-cache=16384    | will cache up to 16384 threads in the tree (LRU cache).
```
yanivagman, line 37: This doesn't match the documented values of 16384 processes and 32K threads

yanivagman (approved):
LGTM

PR #4389, opened 2024-11-20
perf: remove sys_enter/exit dependency from default event set

yanivagman:

This PR removes the dependency of the default event set on sys_enter/sys_exit probes to improve performance when no specific event is selected.

To do that, update default event set and signatures that were depend on events that used these probes.

### DIFF
--- a/signatures/golang/illegitimate_shell.go
+++ b/signatures/golang/illegitimate_shell.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"fmt"
-	"strings"
 
 	"github.com/aquasecurity/tracee/signatures/helpers"
 	"github.com/aquasecurity/tracee/types/detect"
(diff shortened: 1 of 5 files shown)


signatures/golang/illegitimate_shell.go
```diff
@@ -54,16 +53,15 @@ func (sig *IllegitimateShell) OnEvent(event protocol.Event) error {
...
 		for _, webServersProcessName := range sig.webServersProcessNames {
-			if webServersProcessName == eventObj.ProcessName {
-				pathname, err := helpers.GetTraceeStringArgumentByName(eventObj, "pathname")
-				if err != nil {
-					return err
-				}
-
+			if webServersProcessName == prevComm {
 				for _, shellName := range sig.shellNames {
-					if strings.HasSuffix(pathname, shellName) {
```
itaysk, line 66: I'm not sure but can't processName be controlled by the user, whereas pathName can't?

yanivagman: pathname can be controlled by the user as well - altering the binary name, for example, will change the path suffix

PR #4390, opened 2024-11-21
Refactor: Restructure event and rename context

yanivagman:

This commit introduces two key changes to the Tracee event structure:

- Renames the `context` field to `workload` to provide a more specific and clear description of the execution environment.
- Moves the `triggeredBy` field from the `data` section to the root of the event structure. This ensures the `data` section is strictly defined by the event-specific schema, improving consistency and clarity.

These changes constitute a breaking change to the API. However, as `traceectl` is not yet merged, it is acceptable to introduce this breaking change now to avoid future complications.


geyslan (approved):
LGTM. @ShohamBit would you mind testing traceectl with this change?

PR #4394, opened 2024-11-24
fix(epbf): fix behavior of `has_prefix() ` and add `strncmp()`

### LINKED ISSUE #4395: `has_prefix()` behaves incorrectly
There are 2 implementations for `has_prefix()`, one as a macro and one as a function (depending on the compiler version).
They both behave differently when no difference was found between the prefix and the string in `n` iterations.

The macro returns 1 (true), which is an issue if the prefix is longer than `n`, while the function returns 0 (false) which is incorrect if the prefix and the string are equal.

oshaked1:
Closes #4395 

There are 2 implementations for `has_prefix()`, one as a macro and one as a function (depending on the compiler version). They both behave differently in cases where no difference was found in `n` iterations - the macro returns 1 (true), which is an issue if the prefix is longer than `n`, while the function returns 0 (false) which is wrong if the prefix and string are identical. An additional check was added to `has_prefix()` to account for prefixes longer than `n`, while still returning true when the strings are equal.

Additionally, an `strncmp()` macro and function were added, because most uses of `has_prefix()` become clearer when using `strncmp()`.

Only a single usage of `has_prefix()` (in `filter_file_path()`) cannot use `strncmp()` because the prefix length can only be determined at runtime which makes usage of an unrolled loop impossible. Instead, it must use `has_prefix()` which accounts for prefixes shorter than the string, unlike `strncmp()`.

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -1925,7 +1925,7 @@ send_bpf_perf_attach(program_data_t *p, struct file *bpf_prog_file, struct file 
     bpf_probe_read_kernel_str(
         &class_system, REQUIRED_SYSTEM_LENGTH, BPF_CORE_READ(tp_class, system));
     class_system[REQUIRED_SYSTEM_LENGTH - 1] = '\0';
-    if (has_prefix("syscalls", class_system, REQUIRED_SYSTEM_LENGTH)) {
+    if (strncmp("syscalls", class_system, REQUIRED_SYSTEM_LENGTH - 1) == 0) {
         is_syscall_tracepoint = true;
     }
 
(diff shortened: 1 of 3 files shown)


pkg/ebpf/c/tracee.bpf.c
```diff
@@ -3140,7 +3140,7 @@ statfunc int capture_file_write(struct pt_regs *ctx, u32 event_id, bool is_buf)
     // otherwise the capture will overwrite itself.
     int pid = 0;
     void *path_buf = get_path_str_cached(file);
-    if (path_buf != NULL && has_prefix("/dev/null", (char *) path_buf, 10)) {
+    if (path_buf != NULL && strncmp("/dev/null", (char *) path_buf, 9) == 0) {
```
yanivagman, line 3143: In this case, don't we want to compare 10 bytes to take into account null byte as well?

oshaked1: You're right I'll change it

yanivagman (approved):
LGTM

PR #4398, opened 2024-11-25
refactor: Rename event parameters to fields

yanivagman:

This PR renames the `parameters` field in event definitions to `fields`. This change reflects the evolving nature of Tracee, which now supports various event types beyond just syscalls.

The term "fields" more accurately describes the structured data within events, aligning with the concept of event schemas and the `data` section in the new event structure.

This renaming also avoids future confusion with the planned introduction of "event parameters," which will be configurable settings for modifying event behavior.

### DIFF
--- a/pkg/bufferdecoder/decoder.go
+++ b/pkg/bufferdecoder/decoder.go
@@ -89,13 +89,13 @@ func (decoder *EbpfDecoder) DecodeContext(eCtx *EventContext) error {
 // DecodeArguments decodes the remaining buffer's argument values, according to the given event definition.
 // It should be called last, and after decoding the argnum with DecodeUint8.
 //
-// Argument array passed should be initialized with the size of len(evtParams).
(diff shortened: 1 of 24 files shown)


NDStrahilevitz:
What's the logical difference between the `fields` section and the `data` section? `fields` represents the metadata for the concrete `data` found in an event? If so, why not just align the names?

yanivagman:
> What's the logical difference between the `fields` section and the `data` section? `fields` represents the metadata for the concrete `data` found in an event? If so, why not just align the names?

I agree that clear terminology is crucial. Here's why I chose 'fields' for the event definition and 'data' for the output:

`data` in Output: This represents the actual event payload – the concrete values captured at runtime. It's the raw information associated with the event, like the arguments passed to a syscall or the contents of a network packet.

`fields` in Definition: This acts as the metadata or schema for the event. It defines the structure, types, and possible values within the data section.

While the complete term might be "data fields," using "data" for the output and "fields" for the definition provides contextual clarity and avoids redundancy.

geyslan (approved):
LGTM

PR #4400, opened 2024-11-28
Refactor policy

yanivagman:

commit 9e6ddcd658dc282f041046292eb2801381742457
Author: Yaniv Agman <yanivagman@gmail.com>
Date:   Thu Nov 28 17:32:23 2024 +0200

    test(pkg/cmd/flags): add unit tests for policy parsing functions
    
    Add unit tests for recently added policy parsing functions:
    - parseScopeFilters
    - createSinglePolicy
    - parseEventFilters

commit ec0ec43512941b492c81d0d2bb84c075fddbebe2
Author: Yaniv Agman <yanivagman@gmail.com>
Date:   Thu Nov 28 14:16:57 2024 +0200

    refactor: simplify CreatePolicies function structure
    
    Split the large CreatePolicies function into smaller, focused functions
    for better readability and maintainability. The main changes:
    
    - Split policy creation logic into separate functions
    - Simplified function signatures and parameter passing
    - Made variable naming more consistent
    - Improved error handling organization
    
    No functional changes, only code reorganization.

commit d575f3d90cf544c5558cedf315a6522b85c870e5
Author: Yaniv Agman <yanivagman@gmail.com>
Date:   Mon Nov 25 17:54:48 2024 +0200

    refactor: align filters with policy rule structure
    
    Restructure filters to match tracee's policy structure by moving event-specific
    filters into policy.Rules. Main changes:
    
    - Move DataFilter, ScopeFilter and RetFilter from top-level maps into
      policy.Rules[eventID]
    - Remove filter.go and its flag as event filtering is now handled through
      the policy structure
    - Adjust all filter operations to work with the new Rules-based structure
    
    This change better reflects tracee's policy design where each event has
    its own set of filters under a rule, preparing the stage for moving from
    matched_policies to matched_rules resolution.

commit cc8947dd6013b038366b6dd808883e1e44fe3eb3
Author: Yaniv Agman <yanivagman@gmail.com>
Date:   Thu Nov 28 11:48:17 2024 +0200

    docs: split filter help into scope and events help
    
    Split the filter help message into separate scope and events help messages
    to match the CLI structure. Moved documentation from the deprecated filter
    flag into their respective --scope and --events flag help messages.


pkg/filters/data.go
```diff
@@ -12,7 +11,7 @@ import (
 )
 
 type DataFilter struct {
-	filters map[events.ID]map[string]Filter[*StringFilter]
+	filters map[string]Filter[*StringFilter]
```
geyslan, line 14: @rscampos this conflicts with your effort.

pkg/filters/data.go
```diff
@@ -21,39 +20,31 @@ var _ utils.Cloner[*DataFilter] = &DataFilter{}
...
+		filters: map[string]Filter[*StringFilter]{},
 		enabled: false,
 	}
 }
 
 // GetEventFilters returns the data filters map for a specific event
 // writing to the map may have unintentional consequences, avoid doing so
-func (af *DataFilter) GetEventFilters(eventID events.ID) map[string]Filter[*StringFilter] {
-	return af.filters[eventID]
+// TODO: encapsulate by replacing this function with "GetFieldFilter(fieldName string) StringFilter"
```
geyslan, line 30: If possible, `GetFieldFilter(field int)` should receive an int.

pkg/policy/policies_compute.go
```diff
@@ -126,9 +126,20 @@ func (ps *policies) updateUserlandPolicies() {
...
+
+		// Check filters under Rules
+		for _, rule := range p.Rules {
+			if rule.DataFilter.Enabled() ||
+				rule.RetFilter.Enabled() ||
+				rule.ScopeFilter.Enabled() {
+				hasUserlandFilters = true
+				break
+			}
+		}
```
geyslan, line 139: Could this logic set an internal flag to be checked via a Rules public method as a shortcut in the pipeline?

pkg/policy/policy.go
```diff
@@ -22,12 +19,17 @@ type Policy struct {
...
-	DataFilter        *filters.DataFilter
-	ScopeFilter       *filters.ScopeFilter
 	ProcessTreeFilter *filters.ProcessTreeFilter
 	BinaryFilter      *filters.BinaryFilter
 	Follow            bool
+	Rules             map[events.ID]RuleData
+}
+
+type RuleData struct {
+	EventName   string
```
geyslan, line 29: An int rather?

pkg/policy/policy.go
```diff
@@ -22,12 +19,17 @@ type Policy struct {
 	ContFilter        *filters.BoolFilter
 	NewContFilter     *filters.BoolFilter
 	ContIDFilter      *filters.StringFilter
-	RetFilter         *filters.RetFilter
-	DataFilter        *filters.DataFilter
-	ScopeFilter       *filters.ScopeFilter
 	ProcessTreeFilter *filters.ProcessTreeFilter
 	BinaryFilter      *filters.BinaryFilter
 	Follow            bool
+	Rules             map[events.ID]RuleData
```
geyslan, line 25: I'm aware that you're just following the current structure definition but, just as a reminder, we'll better encapsulate it converting all public objects to private and make them accessible only by specific methods as we already started in policy.Manager.

geyslan:
I've skimmed it and put some thoughts. When the conflicts are resolved, gonna check it again.

pkg/filters/data.go
yanivagman: I think using a string here will be much more clear...
Anyway, I don't see this part as a bottlenceck, if this is your concern. Thinking about the logic - we still do a map lookup using a string of the data field

pkg/policy/policies_compute.go
yanivagman: We already set `ps.filterableInUserland` a few lines below and use it in the pipeline by `FilterableInUserland()`, what else do you suggest?

pkg/policy/policy.go
yanivagman: Yes, I agree we should eventually encapsulate these stuff and use a clear interface instead of accessing public fields. This is definitely something we will need to do

pkg/policy/policies_compute.go
geyslan: We already have a short circuit:

https://github.com/aquasecurity/tracee/blob/8d22a0791f98d4a1df4e30c95bdb9d9eb3c6801f/pkg/ebpf/events_pipeline.go#L307-L311

But I think it would be faster if we only check a boolean (the `hasUserlandFilters` promoted as a Policies field). So instead of acquiring mutex and doing the bitwise, we would only be required to acquire and return true or false.

https://github.com/aquasecurity/tracee/blob/8d22a0791f98d4a1df4e30c95bdb9d9eb3c6801f/pkg/policy/policy_manager.go#L565-L570

Also, as `PolicyManager` has its own mutex, these atomic aren't necessary anymore (reminiscences of another era):

https://github.com/aquasecurity/tracee/blob/8d22a0791f98d4a1df4e30c95bdb9d9eb3c6801f/pkg/policy/policies.go#L88-L90

yanivagman: But this `hasUserlandFilters` variable doesn't takes the UID and PID filters into account as done below, so what do you suggest exactly?
Anyway, I'm not saying we shouldn't do such a shortcut to avoid mutexes, just that I'm not sure it is related to this PR that handles policy structure change

geyslan: We already discussed this offline. 👍🏼

geyslan (approved):
LGTM

pkg/cmd/flags/scope.go
```diff
@@ -6,6 +6,60 @@ import (
...
+
+Special Values:
+- 'new': For container and pid filters to match newly created ones
+- 'host:' prefix: For executable paths in host namespace
+- namespace ID prefix: For executable paths in specific mount namespace
+
+Examples:
+  Process Filters:
+    --scope pid=new                                   | trace new processes
+    --scope pid=509,1709                              | trace PIDs 510 or 1709
```
rscampos, line 39: Is it a typo?
Shouldn't be: `trace PIDs 509 or 1709`?

pkg/cmd/flags/scope.go
```diff
@@ -6,6 +6,60 @@ import (
...
+
+  Namespace Filters:
+    --scope mntns=4026531839                          | trace specific mount namespace
+    --scope pidns!=4026531835                         | exclude specific PID namespace
+    --scope executable=host:/usr/bin/ls               | trace ls in host mount namespace
+    --scope executable=4026532447:/usr/bin/ls         | trace ls in specific mount namespace
+
+  Process Tree:
+    --scope tree=476164                               | trace process and its descendants
+    --scope tree=3212,5200 --scope tree!=3215         | trace descendants of 3213 or 5200, except 3215's
```
rscampos, line 59: Is it a typo?
Shouldn't be: `trace descendants of 3212 or 5200`?

pkg/filters/data.go
```diff
@@ -21,39 +20,31 @@ var _ utils.Cloner[*DataFilter] = &DataFilter{}
...
 func NewDataFilter() *DataFilter {
 	return &DataFilter{
-		filters: map[events.ID]map[string]Filter[*StringFilter]{},
+		filters: map[string]Filter[*StringFilter]{},
 		enabled: false,
 	}
 }
 
-// GetEventFilters returns the data filters map for a specific event
+// GetFieldFilters returns the data filters map for a specific event
```
rscampos, line 28: since GetFieldFilters doesn't accept events.ID anymore, should we remove the comment part "for a specific event"?

pkg/policy/policy.go
```diff
@@ -22,12 +19,17 @@ type Policy struct {
...
+}
+
+type RuleData struct {
+	EventID     events.ID
```
rscampos, line 29: Is this `EventID` necessary since we already have `events.ID` in Policy.Rules? I tried searching the code but couldn't find any place where it's needed - or maybe we'll need it in the future.

rscampos (approved):
LGTM

pkg/policy/policy.go
yanivagman: My thought here was that a rule is composed from an event and its filters (soon action will be added as well), so I placed it here as well. This might be relevant in the future if we will want to pass a pointer to the Rule without giving a reference to the policy itself

PR #4402, opened 2024-12-02
fix(epbf): fix incorrect parsed syscall name

### LINKED ISSUE #4401: Incorrect syscall name for compat (32bit) processes
## Description

The `syscall` context field contains an incorrect syscall name if the process is compat.

```
main
```

oshaked1:
Commit b21174d8ef0065626e02d716e782d7f3e65a3519 introduced syscall ID translation for compat processes in eBPF, which makes the translation in userspace redundant. The redundant translation caused an incorrect syscall name to be displayed.

Closes #4401


yanivagman:
> Commit [b21174d](https://github.com/aquasecurity/tracee/commit/b21174d8ef0065626e02d716e782d7f3e65a3519) introduced syscall ID translation for compat processes in eBPF, which makes the translation in userspace redundant. The redundant translation caused an incorrect syscall name to be displayed.
> 
> Closes #4401

It looks like this commit just translates from 32bit to 64bit and not the other way around. I don't see anything redundant with this change since we still need to convert back to 32bit syscall id in userspace

oshaked1:
We don't need to convert back to 32-bit, the syscall name is derived from the 64-bit syscall ID by looking up the tracee event with that ID, which should be the syscall event. If the syscall ID that comes from eBPF is already 64-bit, the translation returns an unrelated ID.

yanivagman:
> We don't need to convert back to 32-bit, the syscall name is derived from the 64-bit syscall ID by looking up the tracee event with that ID, which should be the syscall event. If the syscall ID that comes from eBPF is already 64-bit, the translation returns an unrelated ID. The translation is not from 64-bit back to 32-bit, it is the opposite.

But then you return the 64bit syscall name, and not the 32bit syscall name.
The real problem here that needs to be fixed is tha `compatTranslationMap` given as an argument to `parseSyscallID` is a mapping from 32bit to 64bit:
```
sysCompatTranslation := events.Core.IDs32ToIDs()
parseSyscallID(int(eCtx.Syscall), flags.IsCompat, sysCompatTranslation)
```

So the solution should be to create a reverse map from 64bit to 32bit and give it as an argument to this function

oshaked1:
But 64 to 32 bit mapping is useless, there is nothing we can do with the 32-bit ID as the syscall name is retrieved from the 64-bit counterpart.

We could build a mapping of all 32-bit syscall names and that way the correct syscall can be displayed in case translation to 64-bit ID is not possible (which could happen for certain syscalls that only exist on 32-bit).

But for now, my point is that currently an incorrect name is displayed, and my change fixes it.

oshaked1:
> The correct way to go is to build such a mapping of 32bit syscall ids to their respective names, similar to what we have here for 64bit syscalls: https://github.com/aquasecurity/tracee/blob/main/pkg/events/core_amd64.go#L917

I agree that it would be better, but for now we should fix the current issue. My fix does not get rid of any preexisting functionality.

yanivagman (approved):
LGTM

geyslan:
@oshaked1 we're getting this error randomly:

https://github.com/aquasecurity/tracee/actions/runs/12677319392/job/35332416412?pr=4414#step:4:1248

```
INFO: WRITABLE_DATA_SOURCE: FAILED, stderr from tracee:
{"level":"error","ts":1736364222.3955824,"msg":"Ignored key without a value.","ignored":-237}
{"level":"error","ts":1736364222.3954601,"msg":"No syscall event with id %d"}
```

PR #4403, opened 2024-12-02
feat(events): add stack_pivot event

### LINKED ISSUE #4404: Add stack_pivot event
Add stack_pivot event, which detects usage of the stack pivoting technique as part of a ROP exploit, by monitoring selected syscalls and verifying the location of the stack pointer.

oshaked1:
Closes #4404

Add `stack_pivot` event, which detects usage of the stack pivoting technique as part of a ROP exploit, by monitoring selected syscalls and verifying the location of the stack pointer.

### 2. Explain how to test it

Run tracee as follows:

```bash
sudo dist/tracee -e stack_pivot.args.syscall=exit_group
```

Compile and run the tester program:

```bash
gcc -o stack_pivot tests/e2e-inst-signatures/scripts/stack_pivot.c
./stack_pivot
```

After 15 seconds a stack pivot event should be triggered.

### 3. Other comments

Like `suspicious_syscall_source`, this event makes use of event parameters to determine which syscalls should be monitored. The new probe group created for `suspicious_syscall_source` was made generic for "syscall checker" events, and each time a probe is triggered only the "checkers" which selected that syscall are run.


pkg/events/core.go
```diff
@@ -13065,7 +13065,14 @@ var CoreEvents = map[ID]Definition{
 		id:      SuspiciousSyscallSource,
 		id32Bit: Sys32Undefined,
 		name:    "suspicious_syscall_source",
-		sets:    []string{},
+		dependencies: Dependencies{
+			probes: []Probe{
+				{handle: probes.WakeUpNewTask, required: false},    // for thread stack tracking
+				{handle: probes.SchedProcessExec, required: false}, // for thread stack tracking
+				{handle: probes.SchedProcessExit, required: false}, // for thread stack tracking
```
yanivagman, line 13073: These can be removed if you will use task_info_map instead of self maintaing another task map

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -5281,40 +5285,66 @@ statfunc void check_suspicious_syscall_source(void *ctx, struct pt_regs *regs, u
...
-        default:
-            vma_type_str = "unknown";
-            break;
-    }
+    save_to_submit_buf(&p.event->args_buf, &syscall, sizeof(syscall), 0);
+    save_to_submit_buf(&p.event->args_buf, &ip, sizeof(ip), 1);
+    save_str_to_buf(&p.event->args_buf, (void *) vma_type_str, 2);
+    save_to_submit_buf(&p.event->args_buf, &vma_start, sizeof(vma_start), 3);
+    save_to_submit_buf(&p.event->args_buf, &vma_size, sizeof(vma_size), 4);
+    save_to_submit_buf(&p.event->args_buf, &vma_flags, sizeof(vma_flags), 5);
```
yanivagman, line 5298: Your commits are mixing stack_pivot and suspicious_syscalls events.
This commit states it adds the stack pivot event, however, it also modifies logic of the suspicious syscall event.

pkg/events/core.go
oshaked1: I will still need `sched_process_exec` because I need to reset the thread stack info whenever an exec occurs, but `sched_process_exit` can be removed (and `wake_up_new_task` will be replaced with `sched_process_fork_signal`)

pkg/ebpf/c/tracee.bpf.c
oshaked1: The only thing I changed is moving the logic to get `vma_type_str` into a separate function (because it's used by `check_stack_pivot` as well).

yanivagman (approved):
LGTM

PR #4414, opened 2024-12-05
Introduce evt trigger

geyslan:
Close: #4412

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 .PHONY: all | env
-all: tracee-ebpf tracee-rules signatures tracee
+all: tracee-ebpf tracee-rules signatures tracee evt
 
 #
 # make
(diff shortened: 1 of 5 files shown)


NDStrahilevitz:
Just to note that at first I thought the PR refers to the uprobe triggering mechanism, we may want to rethink one of the names...
Anyway, promising tool.

geyslan:
The current idea is to have this new binary `evt` with two main commands `trigger` and `stress`. The former would only run an underlying script related to a single Tracee event chosen. The former would accept multiple events or policies to run Tracee first and in the sequence the `evt trigger` itself for each event chosen.

How do you see the name conflicting with uname? What do you suggest then? I'm all ears.

NDStrahilevitz:
> Member

It's just that there is this package:
https://github.com/aquasecurity/tracee/tree/main/pkg/events/trigger
Which conflicts in name. So we need to rename on of them imo since we have the concept of "triggered events (through uprobes)" as a, rather underdeveloped, internal mechanism of tracee.

geyslan:
> > Member
> 
> It's just that there is this package: https://github.com/aquasecurity/tracee/tree/main/pkg/events/trigger Which conflicts in name. So we need to rename on of them imo since we have the concept of "triggered events (through uprobes)" as a, rather underdeveloped, internal mechanism of tracee.

I don't think they conflict with each other, perhaps in resemblance only, see:

pkg/events/trigger - pkg
cmd/evt/cmd/trigger - cmd

They in different hierarchies. Anyway, we might change one or both in the future if it starts to confuse us. 👍🏼

geyslan:
I'm going to provide documentation after reviews.

geyslan:
@NDStrahilevitz @rscampos I believe this is ready to merge if you guys don't have anything against it.

NDStrahilevitz (approved):
LGTM. 
This is arch specific though I guess? Different arches (probably even different dists), could result in different event sets being outputted. This is particularly relevant for the `common` set which you reuse across multiple events. Any plans or notes you have on that?

geyslan:
> This is arch specific though I guess? ... Any plans or notes you have on that?

Yes, it is. We have plans to make triggers run in containers, as we will likely abandon Tracee's `tree` scope and `follow`. Then, pinning the container image versions we'll certainly have the same set of events.

PR #4426, opened 2024-12-12
Refactor: Remove Rego signature support

yanivagman:

This PR removes support for Rego-based signatures in Tracee.

Rego signatures, while easier to write than compiled Go signatures, introduced performance overhead. This removal simplifies the Tracee codebase and prepares for the future introduction of Wasm-based signatures, which will offer both improved performance and extended capabilities.

In the meantime, users can leverage the following:

- **Go-based signatures:** For optimal performance and custom event handling.
- **Tracee policies:** For basic event filtering needs, providing a convenient way to define filters directly within the policy.

Tracee policies currently address the most common use case for signatures. Future development will focus on expanding Tracee's capabilities to accommodate more advanced scenarios.


geyslan (approved):
LGTM. There's only a unit test err, perhaps related to formatting.

Issue #4391, opened 2024-11-22
struct inode ctime fields changed again with Kernel >= 6.11


agalauner-r7:
## Description

Running tracee on a kernel with version >= 6.11 fails with the below error.

Reason for this is that the `ctime` fields in the `inode` struct changed again: https://github.com/torvalds/linux/commit/3aa63a569c64e708df547a8913c84e64a06e7853

```
[...]
; if (bpf_core_field_exists(inode->__i_ctime)) { // Version >= 6.6 @ filesystem.h:61
3941: (15) if r1 == 0x0 goto pc+2     ; R1_w=0
3944: <invalid CO-RE relocation>
failed to resolve CO-RE relocation <byte_off> [620] struct inode___older_v66.i_ctime (0:0 @ offset 0)
processed 3273 insns (limit 1000000) max_states_per_insn 0 total_states 226 peak_states 226 mark_read 173
-- END PROG LOAD LOG --
{"level":"warn","ts":1732264697.5150177,"msg":"libbpf: prog 'tracepoint__sched__sched_process_exec': failed to load: -22"}
{"level":"warn","ts":1732264697.518446,"msg":"libbpf: failed to load object ''"}
{"level":"fatal","ts":1732264697.5203826,"msg":"Tracee runner failed","error":"cmd.Runner.Run: error initializing Tracee: ebpf.(*Tracee).Init: ebpf.(*Tracee).initBPF: failed to load BPF object: invalid argument"}
```

## Output of `tracee version`:

```
Tracee version: v0.22.4
```

## Output of `uname -a`:

```
Linux 500576e7ccce 6.11.6-arch1-1 #1 SMP PREEMPT_DYNAMIC Fri, 01 Nov 2024 03:30:41 +0000 x86_64 GNU/Linux
```

## Additional details 

I would fix it myself and submit a PR, but there are multiple ways to do it and I am not sure what's the preferred way.

Right now you check using the CO-RE framework if a new field for kernels between 6.6 and 6.10 is present. If so, read that, otherwise cast the struct into an older mock version and use CO-RE to read the old field.

This doesn't work if we have three different versions now.

There is a way to check for the current linux kernel version by declaring an external variable:

```
extern int LINUX_KERNEL_VERSION __kconfig;

if (LINUX_KERNEL_VERSION >= KERNEL_VERSION(6, 11, 0)) {
    /* we are on v6.11+ */
}
```

So, naively, I ended up with code like this:

```c
[...]

Issue #3385, opened 2023-08-13
vma_area_struct first member is not address since 6.4


AlonZivony:
## Description

In our code of `security_file_mprotect` we assume that the first argument is address:
```c
if (should_submit_mprotect) {
        void *addr = (void *) sys->args.args[0];
```
However, this is not true.
The first value is `struct vma_area_struct*`, which has the following structure:
```c
struct [vm_area_struct](https://elixir.bootlin.com/linux/v6.0.19/C/ident/vm_area_struct) {
	/* The first cache line has the info for VMA tree walking. */

	unsigned long [vm_start](https://elixir.bootlin.com/linux/v6.0.19/C/ident/vm_start);		/* Our start address within vm_mm. */
	unsigned long [vm_end](https://elixir.bootlin.com/linux/v6.0.19/C/ident/vm_end);		/* The first byte after our end address
					   within vm_mm. */
``` 
However, this structure has changed in kernel v6.4:
```c
struct [vm_area_struct](https://elixir.bootlin.com/linux/latest/C/ident/vm_area_struct) {
	/* The first cache line has the info for VMA tree walking. */

	union {
		struct {
			/* VMA covers [vm_start; vm_end) addresses within mm */
			unsigned long [vm_start](https://elixir.bootlin.com/linux/latest/C/ident/vm_start);
			unsigned long [vm_end](https://elixir.bootlin.com/linux/latest/C/ident/vm_end);
		};
#ifdef [CONFIG_PER_VMA_LOCK](https://elixir.bootlin.com/linux/latest/K/ident/CONFIG_PER_VMA_LOCK)
		struct [rcu_head](https://elixir.bootlin.com/linux/latest/C/ident/rcu_head) [vm_rcu](https://elixir.bootlin.com/linux/latest/C/ident/vm_rcu);	/* Used for deferred freeing. */
#endif
	};
```

This means that starting from kernel v6.4, this operation might result false address in the event.
I didn't check about other places we use this struct but it might be relevant for other places as well.
**This is not urgent issue**, but it will prove to be problematic as time pass.

AlonZivony:
@rafaeldtinoco @yanivagman for your attention

yanivagman:
The current code is already correct https://github.com/aquasecurity/tracee/blob/main/pkg/ebpf/c/tracee.bpf.c#L3699:
- Gets VMA struct pointer from PT_REGS_PARM1(ctx)
- Extracts actual address from syscall args using get_syscall_arg1()
[...]

Issue #2904, opened 2023-03-22
Event "magic_write" does not work correctly in ARM64 kernels < 5.4


rafaeldtinoco:
## Description

When checking why signature **TRC-1022** wasn't working in ARM64 kernels below 5.4, I've seen that the "bytes" argument coming from that event, in that architecture, seems to be always empty:

```
{"name":"bytes","type":"bytes","value":"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA="},{"name":"dev","type":"dev_t","value":54},{"name":"inode","type":"unsigned long","value":1314215}]}
```

Making the signature 1022 to ignore the file as it does not think the file is an ELF file.

## Output of `tracee -v`:

```
Tracee version "v0.12.0-74-g3e62e7a8"
```

## Output of `uname -a`:

```
Linux ip-172-31-4-24 5.4.0-122-generic #138-Ubuntu SMP Wed Jun 22 15:05:39 UTC 2022 aarch64 aarch64 aarch64 GNU/Linux
```

## Additional details 

Run tracee and filter for "magic_write" event in an ARM64 Focal image (can be the same AMI that is used in tracee kernel tests).

**NOTE**: TRC-1022 should be re-enabled in ARM64 tests whenever this issue is fixed.

Context: https://github.com/aquasecurity/tracee/issues/2821

yanivagman:
We no longer support kernels < 5.4

PR #4428, opened 2024-12-13
chore(GH): pin ubuntu (22.04) version for gh runners

rscampos:

8c14f78d4 **chore(GH): pin ubuntu (22.04) version for gh runners**

```
Currently, ubuntu:latest uses Ubuntu version 24.04, which includes
kernel 6.8. Since Tracee needs to be tested with kernel 6.8, it is
better to pin the version of the Ubuntu runners.

After testing Tracee with kernel 6.8 and confirming that everything
works correctly, this commit can be reverted.
```

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -84,7 +84,7 @@ jobs:
   #
   verify-docs:
     name: Verify Documentation
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - name: Checkout code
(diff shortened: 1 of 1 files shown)


geyslan (approved):
LGTM

geyslan:
Linking for context:

- https://github.com/aquasecurity/tracee/issues/2904
- https://github.com/aquasecurity/tracee/issues/3385
- https://github.com/aquasecurity/tracee/issues/4391

PR #4441, opened 2024-12-18
chore: add kernel 6.11 and 6.12 in matrix images

rscampos:

d0cd0a473 **chore: add kernel 6.11 and 6.12 in matrix images**

```
- Add kernel 6.11 and 6.12, both for x86_64 and aarch64;
- Both kernels are based on Ubuntu 24.04 LTS (codename Noble).
```


yanivagman (approved):
LGTM

PR #4442, opened 2024-12-18
fix empty arguments resolution

### LINKED ISSUE #3892: Failure in argument name resolution results in empty value
## Description

When tracee tries to resolve a numeric argument to a string (e.g. cmd value of bpf syscall), if the resolution fails, the event field will contain an empty string.

For example, running the following command, which uses a new eBPF feature not supported on my kernel:

```bash
sudo bpftool gen skeleton -L hello.bpf.o > hello.skel.h
```

Results in the following tracee event:

```
11:37:28:562345  0      bpftool          211739  211739  -22              bpf                       cmd: , attr: 0x7ffc011cdec0, size: 8
```

The strace output shows that this command is not supported, which explains why the resolution fails:

```
bpf(0x24 /* BPF_??? */, 0x7ffe6338f210, 8) = -1 EINVAL (Invalid argument)
```

This is only one example of incorrect handling of failed name resolutions, another example I found is ptrace commands, and there are possibly many others.

An example of a correctly handled name resolution is the syscall name of sys_enter, where an unknown syscall will result in the syscall number as the value.

```
Tracee version: v0.20.0
```

## Output of `uname -a`:

```
Linux ********* 5.15.133.1-microsoft-standard-WSL2 #1 SMP Thu Oct 5 21:02:42 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```

ShohamBit:
- fix issue #3892 
- fix the empty result of failed parse helper argument
- add test to parse_args_helper.go

`make check-pr`:

1cbdc3d1dc **tests: add argument parsers tests**
69d55cbe77 **fix: empty arguments resolution**

69d55cbe77 **fix: empty arguments resolution**

```
When tracee tries to resolve a numeric argument to a string (e.g. cmd value of
bpf syscall), if the resolution fails, the event field will contain an empty
string.

Return the raw value as a string in case of a failed resolution.
```
### 2. Explain how to test it

In order to test you can run the test in the file with this command 
``` bash
GOOS=linux CC=clang GOARCH=amd64 CGO_CFLAGS="-I/tracee/dist/libbpf -I/tracee/3rdparty/libbpf/include/uapi" CGO_LDFLAGS="-L/tracee/dist/libbpf/obj -lbpf" \
go test \
	-tags ebpf \
        -v \
        -short \
	-race \
	-coverprofile=coverage.txt \
	./pkg/events -run TestParseArgsHelpers
```

- this will run the test functions for parse_args_helpers.go file
-  you can  change `TestParseArgsHelpers` to a function name like: `TestParseBPFCmd` to test a more specific 
### 3. Other comments

Resolves #3892

### DIFF
--- a/pkg/events/parse_args_helpers.go
+++ b/pkg/events/parse_args_helpers.go
@@ -1,6 +1,8 @@
 package events
 
 import (
+	"strconv"
+
 	"github.com/aquasecurity/tracee/pkg/events/parsers"
 	"github.com/aquasecurity/tracee/pkg/logger"
 	"github.com/aquasecurity/tracee/types/trace"
@@ -16,7 +18,7 @@ func parseSocketDomainArgument(arg *trace.Argument, domain uint64) {
 	arg.Type = "string"
 	socketDomainArgument, err := parsers.ParseSocketDomainArgument(domain)
 	if err != nil {
-		arg.Value = ""
+		arg.Value = strconv.FormatUint(domain, 10)
 		return
 	}
 	arg.Value = socketDomainArgument.String()
@@ -26,7 +28,7 @@ func parseSocketType(arg *trace.Argument, typ uint64) {
 	arg.Type = "string"
 	socketTypeArgument, err := parsers.ParseSocketType(typ)
 	if err != nil {
-		arg.Value = ""
+		arg.Value = strconv.FormatUint(typ, 10)
 		return
 	}
 	arg.Value = socketTypeArgument.String()
@@ -36,7 +38,7 @@ func parseInodeMode(arg *trace.Argument, mode uint64) {
 	arg.Type = "string"
 	inodeModeArgument, err := parsers.ParseInodeMode(mode)
 	if err != nil {
-		arg.Value = ""
+		arg.Value = strconv.FormatUint(mode, 10)
 		return
 	}
 	arg.Value = inodeModeArgument.String()
@@ -46,7 +48,7 @@ func parseBPFProgType(arg *trace.Argument, progType uint64) {
 	arg.Type = "string"
 	bpfProgTypeArgument, err := parsers.ParseBPFProgType(progType)
 	if err != nil {
-		arg.Value = ""
+		arg.Value = strconv.FormatUint(progType, 10)
 		return
 	}
 	arg.Value = bpfProgTypeArgument.String()
@@ -56,7 +58,7 @@ func parseCapability(arg *trace.Argument, capability uint64) {
 	arg.Type = "string"
 	capabilityFlagArgument, err := parsers.ParseCapability(capability)
 	if err != nil {
-		arg.Value = ""
+		arg.Value = strconv.FormatUint(capability, 10)
 		return
 	}
 	arg.Value = capabilityFlagArgument.String()
@@ -74,12 +76,13 @@ func parseSyscall(arg *trace.Argument, id int32) {
 	// NOTE: This might cause data races in the future if the map is modified.
 	// One solution to keep better CPU time is to segregate the map into two maps:
 	// one for proper core (read-only) events and another for the dynamic events.
+	arg.Type = "string"
 	def, ok := CoreEvents[ID(id)]
 	if !ok || !def.IsSyscall() {
+		arg.Value = strconv.FormatInt(int64(id), 10)
 		return
 	}
 
-	arg.Type = "string"
 	arg.Value = def.GetName()
 }
 
@@ -87,7 +90,7 @@ func parsePtraceRequestArgument(arg *trace.Argument, req uint64) {
 	arg.Type = "string"
 	ptraceRequestArgument, err := parsers.ParsePtraceRequestArgument(req)
 	if err != nil {
-		arg.Value = ""
+		arg.Value = strconv.FormatUint(req, 10)
 		return
 	}
 	arg.Value = ptraceRequestArgument.String()
@@ -97,7 +100,7 @@ func parsePrctlOption(arg *trace.Argument, opt uint64) {
 	arg.Type = "string"
 	prctlOptionArgument, err := parsers.ParsePrctlOption(opt)
 	if err != nil {
-		arg.Value = ""
+		arg.Value = strconv.FormatUint(opt, 10)
 		return
 	}
 	arg.Value = prctlOptionArgument.String()
@@ -105,19 +108,19 @@ func parsePrctlOption(arg *trace.Argument, opt uint64) {
 
 func parseSocketcallCall(arg *trace.Argument, call uint64) {
 	arg.Type = "string"
-	socketcallArgument, err := parsers.ParseSocketcallCall(call)
+	socketCallArgument, err := parsers.ParseSocketcallCall(call)
 	if err != nil {
-		arg.Value = ""
+		arg.Value = strconv.FormatUint(call, 10)
 		return
 	}
-	arg.Value = socketcallArgument.String()
+	arg.Value = socketCallArgument.String()
 }
 
 func parseAccessMode(arg *trace.Argument, mode uint64) {
 	arg.Type = "string"
 	accessModeArgument, err := parsers.ParseAccessMode(mode)
 	if err != nil {
-		arg.Value = ""
+		arg.Value = strconv.FormatUint(mode, 10)
 		return
 	}
 	arg.Value = accessModeArgument.String()
@@ -127,7 +130,7 @@ func parseExecFlag(arg *trace.Argument, flags uint64) {
 	arg.Type = "string"
 	execFlagArgument, err := parsers.ParseExecFlag(flags)
 	if err != nil {
-		arg.Value = ""
+		arg.Value = strconv.FormatUint(flags, 10)
 		return
 	}
 	arg.Value = execFlagArgument.String()
@@ -137,7 +140,7 @@ func parseOpenFlagArgument(arg *trace.Argument, flags uint64) {
 	arg.Type = "string"
 	openFlagArgument, err := parsers.ParseOpenFlagArgument(flags)
 	if err != nil {
-		arg.Value = ""
+		arg.Value = strconv.FormatUint(flags, 10)
 		return
 	}
 	arg.Value = openFlagArgument.String()
@@ -147,7 +150,7 @@ func parseCloneFlags(arg *trace.Argument, flags uint64) {
 	arg.Type = "string"
 	cloneFlagArgument, err := parsers.ParseCloneFlags(flags)
 	if err != nil {
-		arg.Value = ""
+		arg.Value = strconv.FormatUint(flags, 10)
 		return
 	}
 	arg.Value = cloneFlagArgument.String()
@@ -157,7 +160,7 @@ func parseBPFCmd(arg *trace.Argument, cmd uint64) {
 	arg.Type = "string"
 	bpfCommandArgument, err := parsers.ParseBPFCmd(cmd)
 	if err != nil {
-		arg.Value = ""
+		arg.Value = strconv.FormatUint(cmd, 10)
 		return
 	}
 	arg.Value = bpfCommandArgument.String()
@@ -167,7 +170,7 @@ func parseSocketLevel(arg *trace.Argument, level uint64) {
 	arg.Type = "string"
 	socketLevelArgument, err := parsers.ParseSocketLevel(level)
 	if err != nil {
-		arg.Value = ""
+		arg.Value = strconv.FormatUint(level, 10)
 		return
 	}
 	arg.Value = socketLevelArgument.String()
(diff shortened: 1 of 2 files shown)


yanivagman:
Let's change all parsers inside this file that currently return an empty value on an error to return the raw value instead

ShohamBit:
> Let's change all parsers inside this file that currently return an empty value on an error to return the raw value instead.

Okay, but I think the code is still going to look weird because if we just print the information without any need for processing it behind the scenes, why would we write the function like this:

``` go
func funcArg(arg *trace.Argument, val uint64) {
    arg.Type = "string"
    argument, err := parsers.funcArg(mode) 
    if err != nil {
        arg.Value = "" 
        return 
    }
    arg.Value = argument.String() 
}
```

instead of like this:

``` go 
func funcArg(arg *tracee.Argument, val uint64) {
    arg.Type = "string"
    arg.Value = val 
}
```

or just remove it entirely?

It looks pretty weird that we check for an error and don't even use it.

Note: Not every function behaves like this, but most of them do.

oshaked1:
> It looks pretty weird that we check for an error and don't even use it.

We do use the error, that way we know to use the raw value.

pkg/events/parse_args_helpers.go
```diff
@@ -46,7 +46,7 @@ func parseBPFProgType(arg *trace.Argument, progType uint64) {
 	arg.Type = "string"
 	bpfProgTypeArgument, err := parsers.ParseBPFProgType(progType)
 	if err != nil {
-		arg.Value = ""
+		arg.Value = progType
```
oshaked1, line 49: If the raw value is used, the value type is inconsistent. We should either change the type accordingly or (more preferrable in my opinion) format the raw value as a string.

pkg/events/parse_args_helpers.go
```diff
@@ -157,7 +157,7 @@ func parseBPFCmd(arg *trace.Argument, cmd uint64) {
 	arg.Type = "string"
 	bpfCommandArgument, err := parsers.ParseBPFCmd(cmd)
 	if err != nil {
-		arg.Value = ""
+		arg.Value = cmd
```
oshaked1, line 160: Same as above

pkg/events/parse_args_helpers.go
yanivagman: I agree - better to format the raw argument as a string

yanivagman:
> > Let's change all parsers inside this file that currently return an empty value on an error to return the raw value instead.
> 
> Okay, but I think the code is still going to look weird because if we just print the information without any need for processing it behind the scenes, why would we write the function like this:
> 
> ```go
> func funcArg(arg *trace.Argument, val uint64) {
>     arg.Type = "string"
>     argument, err := parsers.funcArg(mode) 
>     if err != nil {
>         arg.Value = "" 
>         return 
>     }
>     arg.Value = argument.String() 
> }
> ```
> 
> instead of like this:
> 
> ```go
> func funcArg(arg *tracee.Argument, val uint64) {
>     arg.Type = "string"
>     arg.Value = val 
> }
> ```
> 
> or just remove it entirely?
> 
> It looks pretty weird that we check for an error and don't even use it.
> 
> Note: Not every function behaves like this, but most of them do.

Because the parsers return the parsed value of the argument. For example, if some (fake) argument has a raw value 0x4 which means F_WRITE, the output string will be "F_WRITE". However, if it has a value of 0x20, but the parsers don't know this value, an error will be returned. In this case, we can simply return the string value of the raw argument, which is "0x20"

pkg/events/parse_args_helpers.go
ShohamBit: Yes, that sounds good. However, as discussed earlier, raw values are used when Tracee can't identify the data.  In this code, forming the raw value isn't the best practice because that happens in the pkg/parsers module.

yanivagman: This should work: `fmt.Sprint(rawVal)`

pkg/events/parse_args_helpers.go
```diff
@@ -213,7 +215,7 @@ func parseBpfHelpersUsage(arg *trace.Argument, helpersList []uint64) {
 	}
 
 	arg.Type = "const char**"
-	arg.Value = usedHelpers
+	arg.Value = fmt.Sprint(usedHelpers)
```
oshaked1, line 218: The `usedHelpers` slice is already the finalized value, it shouldn't be formatted as a string. Instead, we should handle parsing errors inside the loop by adding the raw helper number to the slice (formatted as a string)

pkg/events/parse_args_helpers.go
```diff
@@ -235,10 +237,10 @@ func parseBpfAttachType(arg *trace.Argument, attachType int32) {
...
 		attTypeName = "uretprobe"
 	default:
-		arg.Value = ""
+		arg.Value = fmt.Sprint(attachType)
 		logger.Errorw("Unknown attach_type got from bpf_attach event")
 		return
 	}
 
-	arg.Value = attTypeName
+	arg.Value = fmt.Sprint(attTypeName)
```
oshaked1, line 245: `attTypeName` is already a string, no need to format it.

pkg/events/parse_args_helpers.go
```diff
@@ -235,10 +237,10 @@ func parseBpfAttachType(arg *trace.Argument, attachType int32) {
 	case 5:
 		attTypeName = "uretprobe"
 	default:
-		arg.Value = ""
+		arg.Value = fmt.Sprint(attachType)
```
oshaked1, line 240: It would look nicer if we assigned to `attTypeName` instead of directly to `arg.Value` and remove the return, for consistency with rest of the cases.

pkg/events/parsers/data_parsers.go
```diff
@@ -1094,10 +1094,10 @@ var socketcallCallMap = map[uint64]SocketcallCallArgument{
 	SYS_SENDMMSG.Value():    SYS_SENDMMSG,
 }
 
-// ParseSocketcallCall parses the `call` argument of the `socketcall` syscall
+// ParseSocketCall parses the `call` argument of the `socketcall` syscall
 // http://man7.org/linux/man-pages/man2/socketcall.2.html
 // https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/net.h
-func ParseSocketcallCall(rawValue uint64) (SocketcallCallArgument, error) {
+func ParseSocketCall(rawValue uint64) (SocketcallCallArgument, error) {
```
oshaked1, line 1100: The previous name makes more sense because the function is parsing the `call` argument of the `socketcall` syscall.

pkg/events/parse_args_helpers.go
oshaked1: @ShohamBit The issue isn't solved for this function yet because helpers with parsing errors will not be included in the list. We should add the raw value for helpers that couldn't be parsed.

pkg/events/parse_args_helpers.go
```diff
@@ -208,7 +208,7 @@ func parseBpfHelpersUsage(arg *trace.Argument, helpersList []uint64) {
 			// helper number <i> is used. get its name from libbpfgo
 			bpfHelper, err := parsers.ParseBPFFunc(uint64(i))
 			if err != nil {
-				continue
+				usedHelpers = append(usedHelpers, fmt.Sprint(bpfHelper))
```
oshaked1, line 211: Still need the continue after this line, otherwise `.String()` will be called on a uint64.

ShohamBit: ok will fix

oshaked1: Just noticed that you are `bpfHelpers` as the raw value, it should be `i`.

ShohamBit: done

pkg/events/parse_args_helpers.go
ShohamBit: @oshaked1 is it solved now?

pkg/events/parse_args_helpers.go
```diff
@@ -185,15 +187,15 @@ func parseGetSocketOption(arg *trace.Argument, opt uint64, evtID ID) {
 	if err == nil {
 		arg.Value = optionNameArgument.String()
 	} else {
-		arg.Value = ""
+		arg.Value = fmt.Sprint(optionNameArgument)
```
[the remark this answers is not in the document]
ShohamBit: fixed

oshaked1:
LGTM

oshaked1 (approved):

ShohamBit:
@yanivagman I added tests for parse_args_helpers.go

pkg/events/parse_args_helpers.go
```diff
@@ -16,7 +18,7 @@ func parseSocketDomainArgument(arg *trace.Argument, domain uint64) {
 	arg.Type = "string"
 	socketDomainArgument, err := parsers.ParseSocketDomainArgument(domain)
 	if err != nil {
-		arg.Value = ""
+		arg.Value = fmt.Sprint(domain)
```
geyslan, line 21: It seems that all types are previously known, right? In this case, even being a corner case, I would recommend using `strconv.Format{Uint,Int}()` since it's a straightforward conversion which doesn't use the expensive reflection.

Take a look:

```
BenchmarkAll/Sprint_1-32         	10000000	        47.03 ns/op	      16 B/op	       2 allocs/op
BenchmarkAll/Strconv_1-32        	10000000	        16.00 ns/op	       5 B/op	       1 allocs/op
BenchmarkAll/Sprint_2-32         	10000000	        49.21 ns/op	      16 B/op	       2 allocs/op
BenchmarkAll/Strconv_2-32        	10000000	        15.59 ns/op	       5 B/op	       1 allocs/op
```

pkg/events/parsers/data_parsers.go
```diff
@@ -206,9 +206,9 @@ func (o OpenFlagArgument) String() string { return o.stringValue }
 // http://man7.org/linux/man-pages/man2/open.2.html
 // https://elixir.bootlin.com/linux/v5.5.3/source/include/uapi/asm-generic/fcntl.h
 func ParseOpenFlagArgument(rawValue uint64) (OpenFlagArgument, error) {
-	if rawValue == 0 {
-		return OpenFlagArgument{}, nil
-	}
+	// if rawValue == 0 {
+	// 	return OpenFlagArgument{}, nil
+	// }
```
geyslan, line 211: I have an open PR awaiting approval that fix these cases: https://github.com/aquasecurity/tracee/pull/4279/commits/c892271001fde142a3daac0ffb4ffeb2b8a92446#diff-b7f34045358e980276266706956ef8a73b7a42c2df811875ccb375a863bc911dR206-R208

It touches the tests as well. If you have a window to review it, please assign yourself as a reviewer, I'll be glad. After we can sync both PRs.

geyslan (commented):
Put some thoughts.

docs/man/scope.1
```diff
@@ -57,7 +57,7 @@ if starting with `*'.
 .PP
 NOTE: Expressions containing `*' token must be escaped!
 .PP
-you can learn more about the wildcard in the \f[CR]event\f[R] section
+you can learn more about the wildcard in the \f[CR]event\f[R] section.
```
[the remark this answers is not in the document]
ShohamBit: I saw that a GitHub test was falling because of this and just ran: `make -f ./builder/Makefile.man`

pkg/events/parsers/data_parsers.go
ShohamBit: I reviewed the relevant parts of this PR and PR #4279. Since our PRs are not synchronized, there might be some conflicts. I'll wait for PR #4279 to be merged and then rebase mine.

pkg/events/parse_args_helpers.go
```diff
@@ -157,7 +160,7 @@ func parseBPFCmd(arg *trace.Argument, cmd uint64) {
 	arg.Type = "string"
 	bpfCommandArgument, err := parsers.ParseBPFCmd(cmd)
 	if err != nil {
-		arg.Value = ""
+		arg.Value = strconv.FormatUint(cmd, 10)
```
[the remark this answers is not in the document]
ShohamBit: well as you can see the cmd arg is unit64 and not int64, do you suggest we change it to int64?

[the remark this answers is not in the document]
rscampos: I was discussing it with @geyslan and decided to open this issue https://github.com/aquasecurity/tracee/issues/4504. Realized that there are some casting between int and uint that should be avoid (at least from my understanding).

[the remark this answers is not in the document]
ShohamBit: I reviewed this issue and think it is a great addition, also removes unnecessary casting that will also resolute in better performance

yanivagman (approved):
LGTM

Issue #4443, opened 2024-12-18
static release artefacts are being clobbered


geyslan:
## Description

https://github.com/aquasecurity/tracee/blob/b56ee730ca389400fe0b84d3a8d63fb145745614/builder/Makefile.release#L220-L223

Those rule commands are not .ONESHELL, so running in different shells (order).

Issue #4154, opened 2024-06-25
Change version format (snapshots and main branch)


geyslan:
## Description

For snapshot releases and main branch builds version should be
 
`main-d83c07b5b`
 
instead of
 
`v0.21.0-rc-103-gd83c07b5b`

The latter doesn't communicate precisely, since even after the official release, it still uses the release candidate tag as its ancestor (due to the release tag pointing to the release branch, not to main).

geyslan:
This might be solved automatically by changing the release approach (still in analysis). Let's see. @rscampos @itaysk

PR #4444, opened 2024-12-18
fix release tarball static binaries & make daily building faster

geyslan:
Close: #4154
Close: #4443

1517687da **chore(ci): use new make snapshot rule**
7e3eab7bb **chore(builder): version format & quicker snapshot**
feca8a5da **fix(builder): avoid static binary clobbering**

1517687da **chore(ci): use new make snapshot rule**

```
This also removes INSTANCE_TYPE from the snapshot workflow, as it is
expected to be not so resource intensive now.
```

7e3eab7bb **chore(builder): version format & quicker snapshot**

```
This changes the version format (details on #4154).

Also, this makes the snapshot (dev daily) building quicker by not
embedding the BTF files in the tracee binaries.

As daily images are not meant for production, this should not be a
problem, also considering that the majority of the contributors are
running latest kernels.
```

### 2. Explain how to test it

`make -f builder/Makefile.release snapshot` to quickly get a `tracee:dev` image (without BTF embedding).
`make -f builder/Makefile.release release RELEASE_GITHUB=0` to get a `tracee:latest` image.

After check versions:

```
docker run --name tracee -it --rm --pid=host --cgroupns=host --privileged -v /etc/os-release:/etc/os-release-host:ro -v /var/run:/var/run:ro tracee:dev version
Tracee version: fix-release-building-b56ee730c
```

```
docker run --name tracee -it --rm --pid=host --cgroupns=host --privileged -v /etc/os-release:/etc/os-release-host:ro -v /var/run:/var/run:ro tracee:latest version
Tracee version: v0.22.0
```

Now, when working on your branches, you'll get a crystal binary version.

```
make tracee
./dist/tracee version
Tracee version: 4154-4443-release-building-1517687da
```

### DIFF
--- a/.github/workflows/release-snapshot.yaml
+++ b/.github/workflows/release-snapshot.yaml
@@ -17,7 +17,6 @@ jobs:
     runs-on: 
       - graas_ami-0cdf7ad6d9627da45_${{ github.event.number }}${{ github.run_attempt }}-${{ github.run_id }}
       - EXECUTION_TYPE=LONG
-      - INSTANCE_TYPE=2XLARGE
     permissions:
       contents: read
       packages: write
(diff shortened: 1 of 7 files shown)


yanivagman (approved):
LGTM

PR #4454, opened 2024-12-18
chore(test): use cat cmd to trigger magic_write event

rscampos:

ed57c0016 **chore(test): use cat cmd to trigger magic_write event**

```
- On newer kernels, the cp command no longer uses vfs_write. As
a result, using cp to trigger magic_write will not work.
```


geyslan (approved):
LGTM

PR #4455, opened 2024-12-18
chore: pin ubuntu:latest for gh runners / trigger magic_write event

rscampos:

9e6e4d4fd **chore(test): use more cmd to trigger magic_write event**
e442ebd73 **Revert "chore(GH): pin ubuntu (22.04) version for gh runners"**

e442ebd73 **Revert "chore(GH): pin ubuntu (22.04) version for gh runners"**

```
This reverts commit ad74ef5a637b5c6fd1830b12f58a9cb897643762.
```

### 3. Other comments

The GitHub runner currently uses ubuntu:latest (Ubuntu 24.04 LTS), which includes kernel 6.8. Initially, the version of the Ubuntu runner was pinned because Tracee needed to be tested with kernel 6.8.

After conducting tests on kernel 6.8 and confirming that everything worked as expected, the pinning was reverted.

The reason for pinning and later reverting the version was due to two failing integration tests. These tests relied on the command `cp /etc/netconfig /tmp/netconfig` to generate a `magic_write` event. However, starting from kernel 6.8, this command no longer triggers a `magic_write` event. Instead, the tests were updated to use the `more` command, which works as expected with kernel 6.8.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -84,7 +84,7 @@ jobs:
   #
   verify-docs:
     name: Verify Documentation
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout code
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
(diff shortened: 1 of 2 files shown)


geyslan:
--- FAIL: Test_EventFilters/comm:_event:_data:_trace_event_magic_write_set_in_multiple_policies_using_multiple_filter_types (13.13s)
    --- PASS: Test_EventFilters/comm:_event:_data:_trace_event_security_mmap_file_using_multiple_filter_types (9.11s)
    --- FAIL: Test_EventFilters/comm:_event:_data:_trace_event_security_file_open_and_magic_write_using_multiple_filter_types (14.13s)

geyslan (approved):
LGTM

Issue #4325, opened 2024-09-24
Kernel symbols are updated without required capabilities


oshaked1:
## Description

The kernel symbol table utility relies on `CAP_SYSLOG` in order to be able to read from `/proc/kallsyms` and get actual addresses (otherwise addresses are zeroed-out).

When `NewKernelSymbolTable` is called when initializing tracee, the correct capability is held. But later calls to helpers like `GetSymbolByName` do not acquire `CAP_SYSLOG`, even though the helper may end up reading `/proc/kallsyms` again.

This causes lookups for symbols that were not initially added to `t.requiredKsyms` to produce unusable results (and no error is generated). An example for such a lookup occurs when fetching the address for attaching a kprobe belonging to an event that is needed as a dependency (and not selected directly).

The simplest way to reproduce this error is by deleting `dist/signatures/builtin.so` (because some signatures select events that list `CAP_SYSLOG` as a dependency), and running tracee without any arguments:

```
$ sudo dist/tracee > /dev/null
{"level":"warn","ts":1727174815.8476572,"msg":"libbpf: prog 'trace_load_elf_phdrs': failed to create kprobe '+0x0' perf event: Invalid argument"}
```

As can be seen, this program which is an (unrequired) dependency of `sched_process_exec`, is being loaded to address 0 and fails.

geyslan:
I was able to reproduce it:

1. `rm dist/signatures/builtin.so`
2. `sudo dist/tracee -o none`

```
{"level":"warn","ts":1727195441.2103937,"msg":"libbpf: prog 'trace_load_elf_phdrs': failed to create kprobe '+0x0' perf event: Invalid argument"}
```

@oshaked1 are you working on solving this? LMK.

oshaked1:
I could solve it by acquiring the capability in the `refresh` method instead of relying on the callers to acquire it. Do you think that would be a good solution?

geyslan:
> I could solve it by acquiring the capability in the `refresh` method instead of relying on the callers to acquire it. Do you think that would be a good solution?

I think we opted to not acquire caps in inner functions since they can be called for other outers acquires, is that correct @NDStrahilevitz?

NDStrahilevitz:
> > I could solve it by acquiring the capability in the `refresh` method instead of relying on the callers to acquire it. Do you think that would be a good solution?
> 
> I think we opted to not acquire caps in inner functions since they can be called for other outers acquires, is that correct @NDStrahilevitz?

[...]

PR #4464, opened 2024-12-25
feat(ksymbols): reimplement ksymbols

### LINKED ISSUE #4463: Significant initialization slowdown from kallsyms parsing
The current kallsyms implementation uses a lazy approach where only symbols marked as required ahead of time are stored. When looking up a symbol that wasn't marked as required fails, the symbol is added ad-hoc to the required symbols and the symbol table is refreshed (the entire `/proc/kallsyms` is read and parsed).

For the most part, symbols that will be looked up are added to the required symbols ahead of time. But there are some cases involving dependencies and kprobe attachments where there are lookup failures which result in a re-read of `/proc/kallsyms`. From a check I performed, a standard execution of tracee with no arguments resulted in `/proc/kallsyms` being parsed 11 times, accounting for more than 10% of the initialization time (around 0.7 seconds from a total of 5.8 seconds). This slowdown becomes much worse with more kprobe attachments. For example,. when supplying a long syscall list (~70 syscalls) for the `susipcious_syscall_source` event, the initialization takes 22 seconds, 15 of which are just from updating the symbol table.

oshaked1:

The previous ksymbols implementation used a lazy lookup method, where only symbols marked as required ahead of time were stored. Trying to lookup a symbol that was not stored resulted in `/proc/kallsyms` being read and parsed in its entirety.
While most symbols being looked up were registered as required ahead of time, some weren't (in particular symbols needed for kprobe attachment) which incurred significant overhead when tracee is being initialized.

This new implementation stores all symbols, or if a `requiredDataSymbolsOnly` flag is used when creating the symbol table (used by default), only non-data symbols are stored (and required data symbols must be registered before updating). Some additional memory usage optimizations are included, for example encoding symbol owners as an index into a list of owner names, and also lazy symbol name lookups where the map of symbol name to symbol is populated only for symbols that were looked up once.

From measurements I performed, the extra memory consumption is around 21MB (from ~159MB to ~180MB when running tracee with no arguments on my machine).

Under the hood, this ksymbols implementation uses a generic symbol table implementation that can be used by future code for managing executable file symbols.

A significant advantage gained by storing all non-data symbols is the ability to lookup a function symbol that contains a given code address, a feature that I plan to use in the future.

This PR closes #4463 and renders #4325 irrelevant (because `/proc/kallsyms` reads no-longer happen "spontaneously").

### DIFF
--- a/pkg/ebpf/tracee.go
+++ b/pkg/ebpf/tracee.go
@@ -78,9 +78,8 @@ type Tracee struct {
 	writtenFiles   map[string]string
 	netCapturePcap *pcaps.Pcaps
 	// Internal Data
-	readFiles     map[string]string
-	pidsInMntns   bucketscache.BucketsCache // first n PIDs in each mountns
-	kernelSymbols *environment.KernelSymbolTable
+	readFiles   map[string]string
+	pidsInMntns bucketscache.BucketsCache // first n PIDs in each mountns
 	// eBPF
 	bpfModule     *bpf.Module
 	defaultProbes *probes.ProbeGroup
@@ -123,6 +122,9 @@ type Tracee struct {
 	policyManager *policy.Manager
 	// The dependencies of events used by Tracee
 	eventsDependencies *dependencies.Manager
+	// A reference to a environment.KernelSymbolTable that might change at runtime.
+	// This should only be accessed using t.getKernelSymbols() and t.setKernelSymbols()
+	kernelSymbols atomic.Pointer[environment.KernelSymbolTable]
 	// Ksymbols needed to be kept alive in table.
 	// This does not mean they are required for tracee to function.
 	// TODO: remove this in favor of dependency manager nodes
@@ -135,6 +137,14 @@ func (t *Tracee) Stats() *metrics.Stats {
 
 func (t *Tracee) Engine() *engine.Engine {
 	return t.sigEngine
+}
+
+func (t *Tracee) getKernelSymbols() *environment.KernelSymbolTable {
+	return t.kernelSymbols.Load()
+}
+
+func (t *Tracee) setKernelSymbols(kernelSymbols *environment.KernelSymbolTable) {
+	t.kernelSymbols.Store(kernelSymbols)
 }
 
 // New creates a new Tracee instance based on a given valid Config. It is expected that it won't
@@ -362,12 +372,13 @@ func (t *Tracee) Init(ctx gocontext.Context) error {
 
 	err = capabilities.GetInstance().Specific(
 		func() error {
-			t.kernelSymbols, err = environment.NewKernelSymbolTable(
-				environment.WithRequiredSymbols(t.requiredKsyms),
-			)
-			// Cleanup memory in list
-			t.requiredKsyms = []string{}
-			return err
+			// t.requiredKsyms may contain non-data symbols, but it doesn't affect the validity of this call
+			kernelSymbols, err := environment.NewKernelSymbolTable(true, true, t.requiredKsyms...)
+			if err != nil {
+				return err
+			}
+			t.setKernelSymbols(kernelSymbols)
+			return nil
 		},
 		cap.SYSLOG,
 	)
@@ -604,13 +615,13 @@ func (t *Tracee) initDerivationTable() error {
 		events.SyscallTableCheck: {
 			events.HookedSyscall: {
 				Enabled:        shouldSubmit(events.SyscallTableCheck),
-				DeriveFunction: derive.DetectHookedSyscall(t.kernelSymbols),
+				DeriveFunction: derive.DetectHookedSyscall(t.getKernelSymbols()),
 			},
 		},
 		events.PrintNetSeqOps: {
 			events.HookedSeqOps: {
 				Enabled:        shouldSubmit(events.HookedSeqOps),
-				DeriveFunction: derive.HookedSeqOps(t.kernelSymbols),
+				DeriveFunction: derive.HookedSeqOps(t.getKernelSymbols()),
 			},
 		},
 		events.HiddenKernelModuleSeeker: {
@@ -913,17 +924,11 @@ func getUnavailbaleKsymbols(ksymbols []events.KSymbol, kernelSymbols *environmen
 	var unavailableSymbols []events.KSymbol
 
 	for _, ksymbol := range ksymbols {
-		sym, err := kernelSymbols.GetSymbolByName(ksymbol.GetSymbolName())
+		_, err := kernelSymbols.GetSymbolByName(ksymbol.GetSymbolName())
 		if err != nil {
 			// If the symbol is not found, it means it's unavailable.
 			unavailableSymbols = append(unavailableSymbols, ksymbol)
 			continue
-		}
-		for _, s := range sym {
-			if s.Address == 0 {
-				// Same if the symbol is found but its address is 0.
-				unavailableSymbols = append(unavailableSymbols, ksymbol)
-			}
 		}
 	}
 	return unavailableSymbols
@@ -944,7 +949,7 @@ func (t *Tracee) validateKallsymsDependencies() {
 	}
 
 	validateEvent := func(eventId events.ID) bool {
-		missingDepSyms := getUnavailbaleKsymbols(evtDefSymDeps(eventId), t.kernelSymbols)
+		missingDepSyms := getUnavailbaleKsymbols(evtDefSymDeps(eventId), t.getKernelSymbols())
 		shouldFailEvent := false
 		for _, symDep := range missingDepSyms {
 			if symDep.IsRequired() {
@@ -1159,7 +1164,7 @@ func (t *Tracee) attachEvent(id events.ID) error {
 		return err
 	}
 	for _, probe := range depsNode.GetDependencies().GetProbes() {
-		err := t.defaultProbes.Attach(probe.GetHandle(), t.cgroups, t.kernelSymbols)
+		err := t.defaultProbes.Attach(probe.GetHandle(), t.cgroups, t.getKernelSymbols())
 		if err == nil {
 			continue
 		}
@@ -1192,7 +1197,7 @@ func (t *Tracee) attachProbes() error {
 				logger.Errorw("Got node from type not requested")
 				return nil
 			}
-			err := t.defaultProbes.Attach(probeNode.GetHandle(), t.cgroups, t.kernelSymbols)
+			err := t.defaultProbes.Attach(probeNode.GetHandle(), t.cgroups, t.getKernelSymbols())
 			if err != nil {
 				return []dependencies.Action{dependencies.NewCancelNodeAddAction(err)}
 			}
@@ -1722,7 +1727,7 @@ func (t *Tracee) triggerSeqOpsIntegrityCheck(event trace.Event) {
 	}
 	var seqOpsPointers [len(derive.NetSeqOps)]uint64
 	for i, seqName := range derive.NetSeqOps {
-		seqOpsStruct, err := t.kernelSymbols.GetSymbolByOwnerAndName("system", seqName)
+		seqOpsStruct, err := t.getKernelSymbols().GetSymbolByOwnerAndName("system", seqName)
 		if err != nil {
 			continue
 		}
@@ -1816,7 +1821,7 @@ func (t *Tracee) triggerMemDump(event trace.Event) []error {
 
 					continue
 				}
-				symbol, err := t.kernelSymbols.GetSymbolByOwnerAndName(owner, name)
+				symbol, err := t.getKernelSymbols().GetSymbolByOwnerAndName(owner, name)
 				if err != nil {
 					if owner != "system" {
 						errs = append(errs, errfmt.Errorf("policy %d: invalid symbols provided to print_mem_dump event: %s - %v", p.ID, field, err))
@@ -1828,7 +1833,7 @@ func (t *Tracee) triggerMemDump(event trace.Event) []error {
 					prefixes := []string{"sys_", "__x64_sys_", "__arm64_sys_"}
 					var errSyscall error
 					for _, prefix := range prefixes {
-						symbol, errSyscall = t.kernelSymbols.GetSymbolByOwnerAndName(owner, prefix+name)
+						symbol, errSyscall = t.getKernelSymbols().GetSymbolByOwnerAndName(owner, prefix+name)
 						if errSyscall == nil {
 							err = nil
 							break
--- a/pkg/utils/environment/kernel_symbols.go
+++ b/pkg/utils/environment/kernel_symbols.go
@@ -2,319 +2,283 @@ package environment
 
 import (
 	"bufio"
-	"fmt"
+	"io"
 	"os"
 	"strconv"
 	"strings"
-	"sync"
+
+	"github.com/aquasecurity/tracee/pkg/errfmt"
+	"github.com/aquasecurity/tracee/pkg/utils"
 )
 
 const (
-	kallsymsPath = "/proc/kallsyms"
-	chanBuffer   = 112800 // TODO: check if we really need this buffer size
+	// Kernel symbols do not have an associated size, so we define a sensible size
+	// limit to prevent unrelated symbols from being returned for an address lookup
+	maxSymbolSize = 0x100000
+
+	ownerShift          = 48                           // Number of bits to shift the owner into the upper 16 bits
+	addressMask         = (1 << ownerShift) - 1        // Mask to extract the address from the addressAndOwner field
+	kernelAddressPrefix = uint64(0xffff) << ownerShift // Precomputed prefix for kernel addresses
 )
 
+// KernelSymbol is a friendly representation of a kernel symbol.
 type KernelSymbol struct {
 	Name    string
-	Type    string
 	Address uint64
 	Owner   string
 }
-type nameAndOwner struct {
-	name  string
-	owner string
-}
-type addrAndOwner struct {
-	addr  uint64
-	owner string
+
+// kernelSymbolInternal is a memory efficient representation of
+// a kernel symbol, used internally for storing all symbols.
+type kernelSymbolInternal struct {
+	name string
+	// We save only the low 48 bits of the address, as all (non-percpu) symbols are at 0xffffXXXXXXXXXXXX
+	// Owner is a 16-bit index into a slice of seen owners for the symbol table this symbol belongs to.
+	// It can only be translated to the owner name if we have the symbol table.
+	// To conserve memory, we encode both of them as a single 64-bit integer where the lower 48-bits
+	// are the address and the hight 16-bits are the owner index.
+	addressAndOwner uint64
 }
 
-// KernelSymbolTable manages kernel symbols with multiple maps for fast lookup.
+func newKernelSymbolInternal(name string, address uint64, owner uint16) *kernelSymbolInternal {
+	return &kernelSymbolInternal{
+		name:            name,
+		addressAndOwner: (uint64(owner) << ownerShift) | (address & addressMask),
+	}
+}
+
+func (ks kernelSymbolInternal) Name() string {
+	return ks.name
+}
+
+func (ks kernelSymbolInternal) Address() uint64 {
+	// Convert truncated address to the real kernel address
+	return kernelAddressPrefix | (ks.addressAndOwner & addressMask)
+}
+
+func (ks kernelSymbolInternal) owner() uint16 {
+	return uint16(ks.addressAndOwner >> ownerShift)
+}
+
+func (ks kernelSymbolInternal) Contains(address uint64) bool {
+	symbolAddr := ks.Address()
+	return symbolAddr <= address && symbolAddr+maxSymbolSize > address
+}
+
+func (ks kernelSymbolInternal) Clone() kernelSymbolInternal {
+	return kernelSymbolInternal{
+		name:            ks.name,
+		addressAndOwner: ks.addressAndOwner,
+	}
+}
+
 type KernelSymbolTable struct {
-	symbols       map[string][]*KernelSymbol
-	addrs         map[uint64][]*KernelSymbol
-	symByName     map[nameAndOwner][]*KernelSymbol
-	symByAddr     map[addrAndOwner][]*KernelSymbol
-	requiredSyms  map[string]struct{}
-	requiredAddrs map[uint64]struct{}
-	onlyRequired  bool
-	updateLock    sync.Mutex
-	updateWg      sync.WaitGroup
+	symbols *utils.SymbolTable[kernelSymbolInternal]
+
+	// Used for memory efficient representation of symbol owners
+	idxToSymbolOwner []string
+	symbolOwnerToIdx map[string]uint16
 }
 
-func symNotFoundErr(v interface{}) error {
-	return fmt.Errorf("symbol not found: %v", v)
+// Creates a new KernelSymbolTable that will be populated from a reader.
+// If lazyNameLookup is true, the mapping from name to symbol will be populated
+// only when a failed lookup occurs. This reduces memory footprint at the cost
+// of the time it takes to lookup a symbol name for the first time.
+// If requiredDataSymbolsOnly is true, only the data symbols passed in the
+// optional requiredDataSymbols argument will be added.
+func NewKernelSymbolTableFromReader(reader io.Reader, lazyNameLookup bool, requiredDataSymbolsOnly bool, requiredDataSymbols ...string) (*KernelSymbolTable, error) {
+	kst := &KernelSymbolTable{
+		symbols:          utils.NewSymbolTable[kernelSymbolInternal](lazyNameLookup),
+		idxToSymbolOwner: []string{"system"},
+		symbolOwnerToIdx: map[string]uint16{"system": 0},
+	}
+
+	if err := kst.update(reader, requiredDataSymbolsOnly, requiredDataSymbols); err != nil {
+		return nil, err
+	}
+
+	return kst, nil
 }
 
-// NewKernelSymbolTable initializes a KernelSymbolTable with optional configuration functions.
-func NewKernelSymbolTable(opts ...KSymbTableOption) (*KernelSymbolTable, error) {
-	k := &KernelSymbolTable{}
-	for _, opt := range opts {
-		if err := opt(k); err != nil {
-			return nil, err
-		}
-	}
-
-	// Set onlyRequired to true if there are required symbols or addresses
-	k.onlyRequired = k.requiredAddrs != nil || k.requiredSyms != nil
-
-	// Initialize maps if they are nil
-	if k.requiredSyms == nil {
-		k.requiredSyms = make(map[string]struct{})
-	}
-	if k.requiredAddrs == nil {
-		k.requiredAddrs = make(map[uint64]struct{})
-	}
-
-	return k, k.Refresh()
-}
-
-// KSymbTableOption defines a function signature for configuration options.
-type KSymbTableOption func(k *KernelSymbolTable) error
-
-// WithRequiredSymbols sets the required symbols for the KernelSymbolTable.
-func WithRequiredSymbols(reqSyms []string) KSymbTableOption {
-	return func(k *KernelSymbolTable) error {
-		k.requiredSyms = sliceToValidationMap(reqSyms)
-		return nil
-	}
-}
-
-// WithRequiredAddresses sets the required addresses for the KernelSymbolTable.
-func WithRequiredAddresses(reqAddrs []uint64) KSymbTableOption {
-	return func(k *KernelSymbolTable) error {
-		k.requiredAddrs = sliceToValidationMap(reqAddrs)
-		return nil
-	}
-}
-
-// TextSegmentContains returns true if the given address is in the kernel text segment.
-func (k *KernelSymbolTable) TextSegmentContains(addr uint64) (bool, error) {
-	k.updateLock.Lock()
-	defer k.updateLock.Unlock()
-
-	segStart, segEnd, err := k.getTextSegmentAddresses()
+// Creates a new KernelSymbolTable that will be populated from /proc/kallsyms.
+// If lazyNameLookup is true, the mapping from name to symbol will be populated
+// only when a failed lookup occurs. This reduces memory footprint at the cost
+// of the time it takes to lookup a symbol name for the first time.
+// If requiredDataSymbolsOnly is true, only the data symbols passed in the
+// optional requiredDataSymbols argument will be added.
+func NewKernelSymbolTable(lazyNameLookup bool, requiredDataSymbolsOnly bool, requiredDataSymbols ...string) (*KernelSymbolTable, error) {
+	file, err := os.Open("/proc/kallsyms")
 	if err != nil {
-		return false, err
-	}
-
-	return addr >= segStart && addr < segEnd, nil
-}
-
-// GetSymbolByName returns all the symbols with the given name.
-func (k *KernelSymbolTable) GetSymbolByName(name string) ([]KernelSymbol, error) {
-	k.updateLock.Lock()
-	defer k.updateLock.Unlock()
-
-	if err := k.validateOrAddRequiredSym(name); err != nil {
-		return nil, err
-	}
-
-	symbols, exist := k.symbols[name]
-	if !exist {
-		return nil, symNotFoundErr(name)
-	}
-
-	return copySliceOfPointersToSliceOfStructs(symbols), nil
-}
-
-// GetSymbolByAddr returns all the symbols with the given address.
-func (k *KernelSymbolTable) GetSymbolByAddr(addr uint64) ([]KernelSymbol, error) {
-	k.updateLock.Lock()
-	defer k.updateLock.Unlock()
-
-	if err := k.validateOrAddRequiredAddr(addr); err != nil {
-		return nil, err
-	}
-
-	symbols, exist := k.addrs[addr]
-	if !exist {
-		return nil, symNotFoundErr(addr)
-	}
-
-	return copySliceOfPointersToSliceOfStructs(symbols), nil
-}
-
-// GetSymbolByOwnerAndName returns all the symbols with the given owner and name.
-func (k *KernelSymbolTable) GetSymbolByOwnerAndName(owner, name string) ([]KernelSymbol, error) {
-	k.updateLock.Lock()
-	defer k.updateLock.Unlock()
-
-	if err := k.validateOrAddRequiredSym(name); err != nil {
-		return nil, err
-	}
-
-	symbols, exist := k.symByName[nameAndOwner{name, owner}]
-	if !exist {
-		return nil, symNotFoundErr(nameAndOwner{name, owner})
-	}
-
-	return copySliceOfPointersToSliceOfStructs(symbols), nil
-}
-
-// GetSymbolByOwnerAndAddr returns all the symbols with the given owner and address.
-func (k *KernelSymbolTable) GetSymbolByOwnerAndAddr(owner string, addr uint64) ([]KernelSymbol, error) {
-	k.updateLock.Lock()
-	defer k.updateLock.Unlock()
-
-	if err := k.validateOrAddRequiredAddr(addr); err != nil {
-		return nil, err
-	}
-
-	symbols, exist := k.symByAddr[addrAndOwner{addr, owner}]
-	if !exist {
-		return nil, symNotFoundErr(addrAndOwner{addr, owner})
-	}
-
-	return copySliceOfPointersToSliceOfStructs(symbols), nil
-}
-
-// getTextSegmentAddresses gets the start and end addresses of the kernel text segment.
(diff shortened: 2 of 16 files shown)


pkg/utils/symbol_table.go
```diff
@@ -0,0 +1,212 @@
...
+	"sync"
+)
+
+// The Symbol interface defines what is needed from a symbol implementation in
+// order to facilitate the lookup functionalities provided by SymbolTable.
+// Implementations of Symbol can hold various types of information relevant to
+// the type of symbol they represent.
+type Symbol interface {
+	Name() string
+	Address() uint64
```
NDStrahilevitz, line 15: Considering the test implementation this seems to be more along the lines of a base address right? As `Contains` should return whether some address is inside an address range defined by the symbol's implemenetation.

Could you maybe add some fast one-line docs to each interface method?

pkg/utils/symbol_table.go
```diff
@@ -0,0 +1,212 @@
...
+// SymbolTable is used to hold information about symbols (mapping from symbolic
+// names used in code to their address) in a certain executable.
+// It can be used to hold symbols from an ELF binary, or symbols of the entire
+// kernel and its modules.
+// It provides functions to lookup symbols by address and name.
+type SymbolTable[T Symbol] struct {
+	mu sync.RWMutex
+	// All symbols sorted by their address in descending order,
+	// for quick binary searches by address.
+	sortedSymbols []*T
```
NDStrahilevitz, line 28: Why `*T`? `Symbol` is already a pointer by definition.

pkg/utils/symbol_table.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// Adds a slice of symbols to the symbol table.
+func (st *SymbolTable[T]) AddSymbols(symbols []*T) {
+	st.mu.Lock()
+	defer st.mu.Unlock()
+
+	// Add the new symbols to the sorted slice (which now becomes unsorted).
+	// Allocate the slice with the needed capacity to avoid overallocation.
+	oldSymbols := st.sortedSymbols
```
NDStrahilevitz, line 56: Pretty sure this "copy" is unsafe as its just a pointer reassignment. 
I think you should rather allocate the new slice, use on copy the old and parameter slice into the new allocation, then assign that one to the object field (while the prior slice will be given to GC).

pkg/utils/symbol_table.go
```diff
@@ -0,0 +1,212 @@
...
+	newLen := len(oldSymbols) + len(symbols)
+	st.sortedSymbols = make([]*T, 0, newLen)
+	st.sortedSymbols = append(st.sortedSymbols, oldSymbols...)
+
+	// If lazyNameLookup is false, we update the name to symbol mapping for
+	// each new symbol
+	if !st.lazyNameLookup {
+		for _, symbol := range symbols {
+			st.sortedSymbols = append(st.sortedSymbols, symbol)
```
NDStrahilevitz, line 65: This is redundant, you could've just as well copied here the same way as the `else` section, which would be more time efficient. 
Note how you're not relying on `st.sortedSymbols` at all in the rest of the block. So you can actually skip the `if..else` copy first like in the `else`, then only do this loop in a single `if`.

pkg/utils/symbol_table.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	// Lazy name lookup is enabled, perform a linear search to find the requested name
+	symbols := []*T{}
+	for _, symbol := range st.sortedSymbols {
+		if (*symbol).Name() == name {
+			symbols = append(symbols, symbol)
+		}
+	}
+	st.mu.RUnlock()
```
NDStrahilevitz, line 110: You should for safety's sake push one of these into the start of the following `if` block, and then duplicate this call before the last `return`.

pkg/utils/environment/kernel_symbols.go
```diff
@@ -2,303 +2,223 @@ package environment
...
-		Name:    symbolName,
-		Type:    symbolType,
-		Address: symbolAddr,
-		Owner:   symbolOwner,
-	}
+func (kst *KernelSymbolTable) ForEachSymbol(callback func(*KernelSymbol)) {
+	kst.symbols.ForEachSymbol(callback)
 }
 
 // copySliceOfPointersToSliceOfStructs converts a slice of pointers to a slice of structs.
```
NDStrahilevitz, line 224: As this is left unused inside the package, and I see you've copied it elsewhere, this can be removed, no reason to leave dangling code.

pkg/utils/environment/kernel_symbols.go
```diff
@@ -2,303 +2,223 @@ package environment
...
-	Owner   string
-}
-type nameAndOwner struct {
-	name  string
-	owner string
+// Kernel symbols do not have an associated size, so we define a sensible size
+// limit to prevent unrelated symbols from being returned for an address lookup
+const maxSymbolSize = 0x100000
+
+var ownersMu sync.RWMutex
```
NDStrahilevitz, line 19: The following owner tracking data structures should be localized to the struct I think. Since this package doesn't expose a singleton (which it certainly could given that a kernel symbol table is a single system file) I don't think we should mix global and struct variables.
That is unless you want to make this a singleton.

pkg/ebpf/tracee.go
```diff
@@ -362,12 +362,16 @@ func (t *Tracee) Init(ctx gocontext.Context) error {
 
 	err = capabilities.GetInstance().Specific(
 		func() error {
-			t.kernelSymbols, err = environment.NewKernelSymbolTable(
-				environment.WithRequiredSymbols(t.requiredKsyms),
-			)
+			t.kernelSymbols = environment.NewKernelSymbolTable(true, true)
+			// t.requiredKsyms may contain non-data symbols, but it doesn't affect the validity of this call
```
NDStrahilevitz, line 366: Since plenty of the symbols i've attempted to include here belong to the `T` symbols, I think trimming that logic will save both RSS, loading time, and code complexity.
If you do it, please do so in a separate commit.

NDStrahilevitz (commented):
Nice work overall, though I do have some comments in mind.

pkg/utils/symbol_table.go
oshaked1: Are you sure it is? From my understanding if I use `T` then `sortedSymbols` contains all of the symbols inline, which I tried but it made it harder to manage symbol references in the name to symbol map.

pkg/utils/symbol_table.go
oshaked1: It's not a copy, I'm just saving a reference to the old slice in a temporary variable before assigning an empty slice to the struct variable.

pkg/utils/symbol_table.go
oshaked1: Not sure I understand what you mean, why is that safer?

pkg/utils/environment/kernel_symbols.go
oshaked1: It's used in the test, should I remove it from the package and call the inner version in the test?

pkg/ebpf/tracee.go
oshaked1: I could trim the symbols included for attachment purposes, but anything included as a ksymbol event dependency needs to stay becuase we don't know if it's a data symbol or not. I think the effect on RSS and loading time would be negligible, but maybe it's worth it for the sake of code complexity.

pkg/utils/symbol_table.go
NDStrahilevitz: I assume you're using a mutex because different goroutines can call this data structure's methods from multiple locations. By unlocking here you're leaving a small gap where another goroutine can "sneak in", do changes and then unlock you back inside the `if` post changes. 
That's why the gap between locks and unlocks should be as small as possible - so this unlock should be done near the inner `Lock` of the `if` or right before the return.

pkg/utils/environment/kernel_symbols.go
NDStrahilevitz: I think you've copied it into the test file no? Then you're using that copy anyway and this is dangling.

pkg/utils/environment/kernel_symbols.go
NDStrahilevitz: What purpose does the index:owner mapping serve? I see that when you parse the file you do string->int and when you call `Owner()` you do int->string (via array), but I don't see where the int plays a role in itself. Why not just use strings overall?
Further, if you expect that an array of owner string will be rather small (up to 5 lets say), then the map duplication is redundant (consider that in event args we prefer arrays to a map), and you could also drop the mutex.

pkg/ebpf/tracee.go
NDStrahilevitz: I think that would still leave us with plenty of symbols down. Aren't the mem_dump symbols also mostly `T` symbols?

pkg/utils/symbol_table.go
NDStrahilevitz: The overall operation you're doing is copying the old into a new extended. I was pinning to the particular line here as its:
1. The first of the operation
2. Seemed problematic to me

pkg/utils/symbol_table.go
NDStrahilevitz: IIRC say you have a struct `S` which implements some interface `I`. Then you can pass `s S` as `I` only with `&s`.

NDStrahilevitz: So what you're doing is requiring concretely a `**KernelSymbol` instead of a `*KernelSymbol` as `Symbol`.

pkg/utils/symbol_table.go
oshaked1: Understood :thumbsup:

pkg/utils/environment/kernel_symbols.go
oshaked1: No, `getTheOnlySymbol()` in the test file uses `KernelSymbolTable.ForEachSymbol()`

pkg/utils/symbol_table.go
oshaked1: > IIRC say you have a struct `S` which implements some interface `I`. Then you can pass `s S` as `I` only with `&s`.

If I understand correctly, this applies only if the struct methods are implemented using a pointer receiver, notice how I implemented them without it and given a non-pointer type when creating the underlying symbol table.

pkg/utils/symbol_table.go
oshaked1: I still don't understand why it's unsafe

pkg/utils/environment/kernel_symbols.go
oshaked1: The purpose is to reduce memory usage, storing a uint16 takes less memory than an owner string. On my machine there are 103 possible owners so I'm not sure it would be smart to drop the map.

oshaked1: Maybe it wouldn't be that bad storing the name, on my machine 30622/230472 symbols are owned by modules, and the system symbols could have a memory efficient representation (like a nil)

pkg/ebpf/tracee.go
oshaked1: > I think that would still leave us with plenty of symbols down. Aren't the mem_dump symbols also mostly `T` symbols?

We have no way to distinguish between T and D symbols requested by `print_mem_dump` (until they are found in the kallsyms file).

pkg/utils/symbol_table.go
NDStrahilevitz: Pointer receivers tend to be more efficient but that's besides the point. Yes you're correct but then why add the indirection at this point? Just stick to the (hopefully) stack allocated objects. Now you're forcing heap escape. If we're going through pointer eventually might as well just author the struct as a pointer receiver and make the code simpler.

pkg/utils/symbol_table.go
NDStrahilevitz: You assign the _pointer_ of `st.sortedSymbols` to `oldSymbols`. This isn't a copy.
Then to the pointer of `st.sortedSymbols` you assign the `make`. This may be reassigning a pointer which is safe or assigning to the content of the pointer. It's probably the former, but if the latter that is unsafe as you are essentially overriding the content of `oldSymbols` as well.
In any case since there is no sorted insertion here, i'm not sure why you do the whole reallocation, just append the items, simpler and quicker, and sometimes you will avoid reallocation if the slice has preallocated space.

pkg/utils/environment/kernel_symbols.go
NDStrahilevitz: I'm referring to `copySliceOfPointersToSliceOfStructs `, I see now it might not be clear from the snippet.

pkg/utils/environment/kernel_symbols.go
NDStrahilevitz: > The purpose is to reduce memory usage, storing a uint16 takes less memory than an owner string. On my machine there are 103 possible owners so I'm not sure it would be smart to drop the map.

That's why I said it depends on the expected count.

> Maybe it wouldn't be that bad storing the name, on my machine 30622/230472 symbols are owned by modules, and the system symbols could have a memory efficient representation (like a nil)

That does sound like a lot and storing the string directly would be costly... So going with the int was a good call.

I still don't like the mixed approach but maybe that means the kernelsymbol table should be a singleton. Tentative on this for now.

pkg/utils/symbol_table.go
oshaked1: > Then to the pointer of `st.sortedSymbols` you assign the `make`. This may be reassigning a pointer which is safe or assigning to the content of the pointer.

It's definitely the former, I'm replacing the pointer stored in `st.sortedSymbols` with a new one, this doesn't affect the slice that is now in `oldSymbols`.

> In any case since there is no sorted insertion here, i'm not sure why you do the whole reallocation, just append the items, simpler and quicker, and sometimes you will avoid reallocation if the slice has preallocated space.

I'm reallocating to avoid an overallocation of the slice, which will never be released. Also regular insertion is likely to result in at least 1 reallocation anyway.

pkg/utils/environment/kernel_symbols.go
oshaked1: Oh you're right, I missed that.

pkg/utils/symbol_table.go
oshaked1: Not sure I understand what you're referring to with stack allocated objects. The reason I chose to use `*T` for `sortedSymbols` and `symbolsByName` is to make sure we're dealing with pointers, otherwise a future user of `SymbolTable` could use a non-pointer type and all of the symbols in `symbolsByName` will be duplicated instead of using their pointers. I guess this also means that if I use `*T` then a future user could use a pointer type which will result in double indirection.

If you think using `T` and passing `*KernelSymbol` as the symbol type is preferrable I'll change it, let me know.

pkg/utils/environment/kernel_symbols.go
oshaked1: @NDStrahilevitz I solved this by moving the owner mapping into the `KernelSymbolTable` struct and splitting the `KernelSymbol` struct into two: `kernelSymbolInternal` stores the owner as an index and it cannot be translated without context of the symbol table, and `KernelSymbol` stores the translated owner and is only created when returning symbols to the user.

oshaked1:
I added an additional memory optimization - kernel symbols now only store the lower 32 bits of the address with the assumption that all addresses begin with `0xffffffff`. We ignore any symbols whose address doesn't start with `0xffffffff`, which is only percpu symbols. This eliminates 8 bytes per symbol (because before there was inefficient padding after the owner), for a total memory saving of around 3-4MB.

pkg/utils/environment/kernel_symbols.go
NDStrahilevitz: Nice, that is a clever optimization :)

pkg/ebpf/tracee.go
NDStrahilevitz: > We have no way to distinguish between T and D symbols requested by `print_mem_dump` (until they are found in the kallsyms file).

Right, I was talking about it more heuristically on the benefit for most cases removing the required request. I guess we can leave that as is for now.

pkg/utils/symbol_table.go
NDStrahilevitz: Aaaah I see, you want to share guarantee that we share the memory between the slice and the map. gotcha. 
Since you have made the table generic, you could still leave it up the implementor to decide if he wants internally pointers or not, via the symbol implementation. This is an internal use package so not that critical but a point to consider if we ever release it independently.

NDStrahilevitz (changes requested):
I have one critical request to make: please avoid the mixed mutex transactions in the kernel symbols table. From experience this tends to cause transaction mixes where one write operation will happen in the middle of a mixed operation for example. Imagine the following methods m1 and m2 where m1 is w and m2 has rw. The following could occur:
1. m2 - r
2. m1 - wait for w
3. m2 - release r
4. m1 - back for w
5. m1 - release
6. m2 - back to w, wiht r assumptions changed due to m1.

Please either pick for each method that it is R or W and make the lock last for the whole operation. You could even opt for a regular mutex instead of a RWMutex, I don't think we have very frequent reads or writes to this struct anyway.

oshaked1:
Symbols lookups could be very frequent in the future (stack trace processing). Using a write lock for the entire duration of `KernelSymbolTable.UpdateFromReader` will prevent symbol lookups for a significant duration. The same applies to `SymbolTable.LookupByName`.

In both functions, the write operations only add new data, they don't change or remove existing data. In the case of `SymbolTable.LookupByName`, the worst case scenario for an outdated assumption means we add the same name to symbol mapping twice (the added data will always be the same). For `KernelSymbolTable.UpdateFromReader`, the worst case scenario is the same owner gets added twice to `idxToSymbolOwner`, but all data remains valid.

I could solve the issue with `UpdateFromReader` by adding a third lock that makes sure only a single update operation can happen at a time (which prevents 2 concurrent goroutines from wanting to add a new symbol owner).

NDStrahilevitz:
> I could also change the API of the kernel symbol table so that reading `/proc/kallsyms` happens once when creating the symbol table, and if the user wants to update it, he must create a new one. This prevents the need for locks at all.
> 
> It also solves a race condition where if a lookup happens between `kst.symbols.Clear()` and `kst.symbols.AddSymbols(symbols)` the lookup will fail.

Wouldn't this be an issue for your stack trace feature if you miss a symbol in the initial request?

> In both functions, the write operations only add new data, they don't change or remove existing data. In the case of `SymbolTable.LookupByName`, the worst case scenario for an outdated assumption means we add the same name to symbol mapping twice (the added data will always be the same). For `KernelSymbolTable.UpdateFromReader`, the worst case scenario is the same owner gets added twice to `idxToSymbolOwner`, but all data remains valid.
> I could solve the issue with `UpdateFromReader` by adding a third lock that makes sure only a single update operation can happen at a time (which prevents 2 concurrent goroutines from wanting to add a new symbol owner).

So this is a valid concern... I can't think of a better solution, and it sounds reasonable that these internal structures should have separate access control. So i'm ok with that. Make the change and i'll +1.
@yanivagman @geyslan tagging you in case you want to drop a review as well.

oshaked1:
> Wouldn't this be an issue for your stack trace feature if you miss a symbol in the initial request?

Do you mean with the current implementation? If so then yes, I only noticed it now.

> So this is a valid concern... I can't think of a better solution, and it sounds reasonable that these internal structures should have separate access control. So i'm ok with that. Make the change and i'll +1.

Which solution do prefer? IMO the second one (RO symbol table, create a new one instead of updating) is favorable.

Additionally, I could change `SymbolTable.symbolsByName` to an LRU which would solve the concurrency issue. It would also allow us to control the memory usage of name mappings.

NDStrahilevitz:
> Additionally, I could change `SymbolTable.symbolsByName` to an LRU which would solve the concurrency issue. It would also allow us to control the memory usage of name mappings.

That is just equivalent to giving it a separate mutex, and I don't see why we would want to limit the owner number. It might be slightly more readable. BTW, you may want to use the `set.Set` data structure for this instead, as the owner symbols just are a set, and I believe i've included a mutex with it.

> > Wouldn't this be an issue for your stack trace feature if you miss a symbol in the initial request?
> 
> Do you mean with the current implementation? If so then yes, I only noticed it now.

I mean if you make it RO and require full regeneration with a new symbol you may easily encounter multiple regenerations per stack trace extraction.
 
> > So this is a valid concern... I can't think of a better solution, and it sounds reasonable that these internal structures should have separate access control. So i'm ok with that. Make the change and i'll +1.
> 
> Which solution do prefer? IMO the second one (RO symbol table, create a new one instead of updating) is favorable.
> 

Which is why the RO symbol table seems like a bad choice for your future needs, unless I am missing something in what you've suggested. Therefore I suggest you use a set, LRU, mutex on the owners, such that the overall concurrency mutex is limited to its relevant data.

oshaked1:
> > Additionally, I could change `SymbolTable.symbolsByName` to an LRU which would solve the concurrency issue. It would also allow us to control the memory usage of name mappings.
> 
> That is just equivalent to giving it a separate mutex, and I don't see why we would want to limit the owner number. It might be slightly more readable. BTW, you may want to use the `set.Set` data structure for this instead, as the owner symbols just are a set, and I believe i've included a mutex with it. Though it may not fit with the int<->string translation scheme you've made.

`symbolsByName` is not for the owners, it's for the cached symbol name to symbol structure mapping (which probably should be size limited). The owners cannot be a set for the reason you mentioned, there is no way to index into it.

> I mean if you make it RO and require full regeneration with a new symbol you, I think you may easily encounter multiple regenerations per stack trace extraction.

There is no regeneration on failed request, only manually (used in `processDoInitModule`). Creating a new kernel symbol table instead of updating it will ensure that for the users, the update is atomic (a new structure gets assigned to `t.kernelSymbols`).

pkg/utils/environment/kernel_symbols.go
```diff
@@ -2,319 +2,284 @@ package environment
...
-const (
-	kallsymsPath = "/proc/kallsyms"
-	chanBuffer   = 112800 // TODO: check if we really need this buffer size
+	"github.com/aquasecurity/tracee/pkg/errfmt"
+	"github.com/aquasecurity/tracee/pkg/utils"
 )
 
+// Kernel symbols do not have an associated size, so we define a sensible size
+// limit to prevent unrelated symbols from being returned for an address lookup
+const maxSymbolSize = 0x100000
```
geyslan, line 17: Consider using more consts like:

```go
const (
	ownerShift          = 48                           // Number of bits to shift the owner into the upper 16 bits
	addressMask         = (1 << ownerShift) - 1        // Mask to extract the address from the addressAndOwner field
	kernelAddressPrefix = uint64(0xffff) << ownerShift // Precomputed prefix for kernel addresses
)

```

pkg/utils/environment/kernel_symbols.go
```diff
@@ -2,319 +2,284 @@ package environment
...
-	// Initialize maps if they are nil
-	if k.requiredSyms == nil {
-		k.requiredSyms = make(map[string]struct{})
-	}
-	if k.requiredAddrs == nil {
-		k.requiredAddrs = make(map[uint64]struct{})
+func newKernelSymbolInternal(name string, address uint64, owner uint16) *kernelSymbolInternal {
+	return &kernelSymbolInternal{
+		name:            name,
+		addressAndOwner: (uint64(owner) << 48) | (address & ((1 << 48) - 1)),
```
geyslan, line 41: Based on the consideration above, this could be:

```suggestion
		addressAndOwner: (uint64(owner) << ownerShift) | (address & addressMask),
```

pkg/utils/environment/kernel_symbols.go
```diff
@@ -2,319 +2,284 @@ package environment
...
-	return k, k.Refresh()
+func (ks kernelSymbolInternal) Name() string {
+	return ks.name
 }
 
-// KSymbTableOption defines a function signature for configuration options.
-type KSymbTableOption func(k *KernelSymbolTable) error
+func (ks kernelSymbolInternal) Address() uint64 {
+	// Convert truncated address to the real kernel address
+	return (0xffff << 48) | (ks.addressAndOwner & ((1 << 48) - 1))
```
geyslan, line 51: As the prefix (all 1) overwrites owner bits I believe this can be simplified to:

```suggestion
	return kernelAddressPrefix | ks.addressAndOwner
```

pkg/utils/environment/kernel_symbols.go
```diff
@@ -2,319 +2,284 @@ package environment
...
+}
 
-// WithRequiredSymbols sets the required symbols for the KernelSymbolTable.
-func WithRequiredSymbols(reqSyms []string) KSymbTableOption {
-	return func(k *KernelSymbolTable) error {
-		k.requiredSyms = sliceToValidationMap(reqSyms)
-		return nil
-	}
+func (ks kernelSymbolInternal) owner() uint16 {
+	return uint16(ks.addressAndOwner >> 48)
```
geyslan, line 55: ditto:

```suggestion
	return uint16(ks.addressAndOwner >> ownerShift)
```

pkg/utils/environment/kernel_symbols.go
```diff
@@ -2,319 +2,284 @@ package environment
...
 }
 
-// WithRequiredAddresses sets the required addresses for the KernelSymbolTable.
-func WithRequiredAddresses(reqAddrs []uint64) KSymbTableOption {
-	return func(k *KernelSymbolTable) error {
-		k.requiredAddrs = sliceToValidationMap(reqAddrs)
-		return nil
-	}
+func (ks kernelSymbolInternal) Contains(address uint64) bool {
+	return ks.Address() <= address && ks.Address()+maxSymbolSize > address
```
geyslan, line 59: consider calling it once:

```suggestion
	addr := ks.Address()
	return addr <= address && addr+maxSymbolSize > address
```

geyslan (commented):
First pass so far. I've only reviewed `kernelSymbolInternal` with some suggestions that could make the code easier to read and change - later I'm going to reach the rest.

pkg/utils/symbol_table.go
```diff
@@ -0,0 +1,215 @@
...
+			name := (*symbol).Name()
+			if symbols, found := st.symbolsByName[name]; found {
+				st.symbolsByName[name] = append(symbols, symbol)
+			} else {
+				st.symbolsByName[name] = []*T{symbol}
+			}
+		}
+	}
+
+	// Sort the symbols slice by address in descending order
```
geyslan, line 79: Perhaps sorting it in ascending order would bring a small performance and API features (as boolean found or the ordered index where a value should be inserted when not found) by using `slices.BinarySearch()` or `slices.BinarySearchFunc()`. WDYT?

pkg/utils/symbol_table.go
```diff
@@ -0,0 +1,215 @@
...
+		return nil, ErrSymbolNotFound
+	}
+
+	// Lazy name lookup is enabled, perform a linear search to find the requested name
+	symbols := []*T{}
+	for _, symbol := range st.sortedSymbols {
+		if (*symbol).Name() == name {
+			symbols = append(symbols, symbol)
+		}
+	}
```
geyslan, line 111: WDYT about moving this reading logic into a private method, so all read mutex control would be managed by it... i.e.: `symbols := st.lookupByName()`

Then we should write lock only if symbols len is > 0.

If you agree, provide comments about the locking in the inner call to warn about unexpected deadlocks if it's used out of context.

pkg/utils/symbol_table.go
```diff
@@ -0,0 +1,215 @@
...
+		return symbols, nil
+	}
+
+	// Lazy name lookup is disabled, the lookup failed
+	if !st.lazyNameLookup {
+		st.mu.RUnlock()
+		return nil, ErrSymbolNotFound
+	}
+
+	// Lazy name lookup is enabled, perform a linear search to find the requested name
```
geyslan, line 105: Depending on the size of sortedSymbols this won't be optimal, but I believe that to keep metadata keyed by name would be worse (memory wise), is that right?

pkg/utils/symbol_table.go
```diff
@@ -0,0 +1,215 @@
...
+	// Find the first symbol at an address smaller than or equal to the requested address
+	idx := sort.Search(len(st.sortedSymbols),
+		func(i int) bool {
+			return address >= (*st.sortedSymbols[i]).Address()
+		})
+
+	// Not found or not exact match
+	if idx == len(st.sortedSymbols) || (*st.sortedSymbols[idx]).Address() != address {
+		return nil, ErrSymbolNotFound
+	}
```
geyslan, line 142: As commented above, I suppose that by using BinarySearch this check might not be necessary.

pkg/utils/symbol_table.go
```diff
@@ -0,0 +1,215 @@
...
+
+	// Not found or the symbol doesn't contain this address
+	if idx == len(st.sortedSymbols) || !(*st.sortedSymbols[idx]).Contains(address) {
+		return nil, ErrSymbolNotFound
+	}
+
+	return st.sortedSymbols[idx], nil
+}
+
+func (st *SymbolTable[T]) ForEachSymbol(callback func(symbol *T)) {
```
geyslan, line 200: This method name makes me to believe that it will do something for each symbol (write to it), but we're just locking for read. Is that right? Couldn't the caller to pass a callback changing the symbol inadvertently?

pkg/utils/symbol_table_test.go
```diff
@@ -0,0 +1,336 @@
...
+			if !reflect.DeepEqual(*(st.symbolsByName[tc.symbols[tc.expectedMappings[i]].name][0]), *tc.symbols[tc.expectedMappings[i]]) {
+				t.Errorf("st.symbolsByName[\"%s\"] = %v, expected %v", tc.symbols[tc.expectedMappings[i]].name, *(st.symbolsByName[tc.symbols[tc.expectedMappings[i]].name][0]), *tc.symbols[tc.expectedMappings[i]])
+				continue
+			}
+		}
+	}
+}
+
+// TestLookupByAddressExact tests the LookupByAddressExact function
+func TestLookupByAddressExaxt(t *testing.T) {
```
geyslan, line 204:
```suggestion
func TestLookupByAddressExact(t *testing.T) {
```

geyslan (commented):
I did one pass more.

Besides my thoughts put, I would recommend you bring also a test file for concurrency like this: https://github.com/aquasecurity/tracee/blob/main/pkg/capabilities/capabilities_test.go

I didn't reviewed `KernelSymbolTable` yet. It will wait for the next pass. 👍🏼

geyslan (commented):
Final pass, I believe. Anything, just ping me.

pkg/utils/environment/kernel_symbols.go
oshaked1: I did `return kernelAddressPrefix | (ks.addressAndOwner & addressMask)` even though as you said the prefix is all 1's, because IMO its more readable.

pkg/utils/symbol_table.go
oshaked1: I'm not sure I understand what's the benefit of sorting in ascending order. I already perform binary searches (using `sort.Search`).

pkg/utils/symbol_table.go
oshaked1: This has the opposite effect of what @NDStrahilevitz commented regarding race conditions with mixed mutex transactions (assumption from read lock is outdated by the time we acquire the write lock) because it places the read unlock and the write lock further apart.

As I commented above, the worst case scenario for such a race condition is duplicate work with no negative effect on the data (the same name is looked up linearly more than once). WDYT? Is such a race condition even worth consideration?

pkg/utils/symbol_table.go
geyslan: slices.BinarySearchFunc() returns the index and found boolean. So you'll be able to shortcut the later checking in LookupByAddressExact(), by instance.

pkg/utils/symbol_table.go
oshaked1: As I said above, I do use binary search. A failed search will return `idx == len(st.sortedSymbols`. The search condition is that the address of the current symbol is smaller or equal to the address being looked up, so we have to verify that it's equal.

pkg/utils/symbol_table.go
oshaked1: You're right, do you think we should return a copy of the symbol to the user?

pkg/utils/symbol_table.go
geyslan: Consider that you're only checking for the existence of symbols with the name passed. I believe that the worst scenario is getting RLock avoiding a write to it (to happen) that would be the finding to update the symbolsByName. So it will be a mismatch even the symbol write locked and waiting for the reading release. What are the odds?

If it is susceptible to happen very often, I would use a write lock for the entire scope then.

oshaked1:
@geyslan some of your remaining comments are regarding lock behavior in `kernel_symbols.go`. As a I proposed in a previous comment, I could change the API such that `KernelSymbolTable` is RO and when it needs to be updated (currently only happens when a kernel module is loaded) a new one is created instead. This would completely remove the need for locks in this file.

WDYT? @NDStrahilevitz @yanivagman it would be great if you could weigh in as well

NDStrahilevitz:
@oshaked1 I worry that jt would expose an easy attack against tracee, simply load and unload a module many times. of course this is already suspicious behavior, and not the only "smoke screen" tactic possible, but adding another one isn't great. That said considering it is not the only such tactic maybe it shouldn't be a blocker for the option.

a way to circumvent this altogether would be if you could cache the delta per module load. if we know for sure that the module in a load loop is the same one, there's no need to calculate the differences each time. although this might be too memory heavy.

Overall no strong opinion on going that way, going RO has many obvious benefits, but the exploit surface slightly worries me.

NDStrahilevitz:
> This attack is also a problem with the current implementation, because we clear the underlying symbol table and add the symbols from `/proc/kallsyms` again. The RO implementation is actually more resistant to this attack becuase lookups can still happen normally while a new kernel symbol table is being created, and only when it's ready `t.kernelSymbols` will be replaced atomically.

True, you're right. So, on my end, I see no issue with moving to an RO implementation.

oshaked1:
@NDStrahilevitz @geyslan I changed `KernelSymbolTable` to RO and guarded `t.kernelSymbols` behind an atomic getter and setter.

pkg/utils/symbol_table.go
oshaked1: I'm not sure how this function behaves when there are multiple symbols with the same requested address, I need to find all of them. `sort.Search` guarantees to find me the first match and then I iterate forward, finding the rest. I could also iterate backwards, but I don't think this would end up being simpler than the current method.

pkg/utils/symbol_table.go
oshaked1: I'm not sure I understood what you mean. The odds of 2 lookups for the same name at very similar times that result in a duplicate linear search is very low (there aren't many name lookups). IMO it's not worth accounting for this.

Anyway regarding your proposal, we would need to return from the internal `st.lookupByName` an indication on whether `st.symbolsByName` needs to be updated, I'm not sure there is much point in separating the two.

pkg/utils/symbol_table.go
yanivagman, line 1: Shouldn't we place this file under environment? I don't think this logic will have any use out of ksymbols context, isn't that so?

pkg/ebpf/tracee.go
```diff
@@ -137,6 +139,14 @@ func (t *Tracee) Engine() *engine.Engine {
...
 }
 
+func (t *Tracee) getKernelSymbols() *environment.KernelSymbolTable {
+	return (*environment.KernelSymbolTable)(atomic.LoadPointer(&t.kernelSymbols))
+}
+
+func (t *Tracee) setKernelSymbols(kernelSymbols *environment.KernelSymbolTable) {
+	atomic.StorePointer(&t.kernelSymbols, unsafe.Pointer(kernelSymbols))
+}
+
```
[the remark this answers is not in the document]
oshaked1: @yanivagman The atomic operations protect from simultaneous reads of `t.kernelSymbols` and an update that replaces it. While there is probably no actual risk, go defines such simultaneous access as a data race so it's just to be safe.

pkg/ebpf/tracee.go
```diff
@@ -123,6 +122,9 @@ type Tracee struct {
 	policyManager *policy.Manager
 	// The dependencies of events used by Tracee
 	eventsDependencies *dependencies.Manager
+	// A reference to a environment.KernelSymbolTable that might change at runtime.
+	// This should only be accessed using t.getKernelSymbols() and t.setKernelSymbols()
+	kernelSymbols unsafe.Pointer
```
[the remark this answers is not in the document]
oshaked1: It's just to ensure that it's only accessed using the getter and setter.

pkg/utils/symbol_table.go
oshaked1: I plan to use it in the future for ELF symbols

NDStrahilevitz (approved):
LGTM after all the changes, nice work.

geyslan:
@yanivagman I still need to check RSS of this.

geyslan:
I'm merging this since we've already discussed regarding the limit of RSS increase, and as explained above it's below.

Issue #4433, opened 2024-12-13
Extend In-Kernel Data Filtering for other types of filtering


rscampos:
The first phase of the in-kernel data filter focused on implementing string-based filters (**exact**, **prefix**, **suffix**). The goal of this task, which comprises the third phase, is to extend the filter mechanism to support types beyond string-based filters.

PR #4468, opened 2025-01-06
fix(ebpf): kernel data filter combine bitmap

rscampos:

3e7df505a **test(filters): kernel data filter**
d66134be7 **fix(ebpf): kernel data filter combine bitmap**

3e7df505a **test(filters): kernel data filter**

```
Add two different test cases:
- test1: policy1 without in-kernel filter and policy2 with in-kernel
  filter;
- test2: policy1 with two filters (one in-kernel filter and other non
  in-kernel filter) and policy2 with in-kernel filter;
```

d66134be7 **fix(ebpf): kernel data filter combine bitmap**

```
- Issue: The final bitmap (res) only contains the enabled bit for
the policy that was matched during the kernel data filter logic.
- Solution: After completing the filter logic, combine the bitmap
for policies without filters to ensure they are correctly enabled
in the final result.
```

### 2. Explain how to test it

Start Tracee with the following two policies:
```
sudo ./dist/tracee -p sfo-01.yaml -p sfo-02.yaml
12:18:40:780375  1000   more             442654  442654  0                security_file_open        pathname: /sys/devices/virtual/dmi/id/bios_date, flags: O_RDONLY|O_NOFOLLOW, dev: 25, inode: 1995, ctime: 1736186855386691414, syscall_pathname: /sys/class/dmi/id/bios_dat
```

Access the following file:
```
more /sys/class/dmi/id/bios_date
Mon, 21 Oct 2024 04:03:58
```
[policies.zip](https://github.com/user-attachments/files/18336864/policies.zip)

### 3. Other comments

A bug was identified and reported by @OriGlassman in the new kernel data filter.

1. **Scenario:** There are two policies with `security_file_open`:
    - **Policy 1:** Uses `security_file_open` without a filter.
    - **Policy 2:** Uses `security_file_open` with pathname filter (which uses in-kernel filter).
2. **Problem:** The final bitmap (`res`) only contains the enabled bit for the policy that was matched during the kernel data filter logic.
    - Example: The resulting bitmap is `10`, meaning only Policy 2 is enabled.
    - However, the correct result should be `11`, where Policy 1 is also enabled because it does not use a filter.
3. **Root Cause:**
    - In Tracee, the scope logic (`match_scope_filters`) works because it initializes all bits to `1` and disables them as needed throughout the code.
    - However, when dealing with the in-kernel filter, we need to use an OR operation between different types of string filters. This requires starting the bitmap with `0` and enabling bits during the filter logic.
    - Unfortunately, this approach only enables policies with in-kernel filters, omitting policies without filters (e.g., Policy 1).
4. **Solution:**  
    After completing the filter logic, combine the bitmap for policies without filters to ensure they are correctly enabled in the final result.

### DIFF
--- a/tests/integration/event_filters_test.go
+++ b/tests/integration/event_filters_test.go
@@ -1951,6 +1951,142 @@ func Test_EventFilters(t *testing.T) {
 			test:         ExpectAtLeastOneForEach,
 		},
 		{
+			name: "comm: event: data: trace event security_file_open set in multiple policies (with and without in-kernel filter)",
+			policyFiles: []testutils.PolicyFileWithID{
+				{
+					Id: 1,
+					PolicyFile: v1beta1.PolicyFile{
+						Metadata: v1beta1.Metadata{
+							Name: "sfo-pol-1",
+						},
+						Spec: k8s.PolicySpec{
+							Scope: []string{
+								"comm=more",
+							},
+							DefaultActions: []string{"log"},
+							Rules: []k8s.Rule{
+								{
+									Event: "security_file_open",
+									Filters: []string{
+										"data.syscall_pathname=/sys/class/dmi/id*",
+									},
+								},
+							},
+						},
+					},
+				},
+				{
+					Id: 2,
+					PolicyFile: v1beta1.PolicyFile{
+						Metadata: v1beta1.Metadata{
+							Name: "sfo-pol-2",
+						},
+						Spec: k8s.PolicySpec{
+							Scope: []string{
+								"comm=more",
+							},
+							DefaultActions: []string{"log"},
+							Rules: []k8s.Rule{
+								{
+									Event: "security_file_open",
+									Filters: []string{
+										"data.pathname=/etc/pam.d/*",
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			cmdEvents: []cmdEvents{
+				newCmdEvents(
+					"more /sys/class/dmi/id/bios_date",
+					0,
+					1*time.Second,
+					[]trace.Event{
+						expectEvent(anyHost, "more", testutils.CPUForTests, anyPID, 0, events.SecurityFileOpen, orPolNames("sfo-pol-1"), orPolIDs(1), expectArg("syscall_pathname", "/sys/class/dmi/id/bios_date")),
+					},
+					[]string{},
+				),
+				newCmdEvents(
+					"more /etc/pam.d/other",
+					0,
+					1*time.Second,
+					[]trace.Event{
+						expectEvent(anyHost, "more", testutils.CPUForTests, anyPID, 0, events.SecurityFileOpen, orPolNames("sfo-pol-2"), orPolIDs(2), expectArg("pathname", "/etc/pam.d/other")),
+					},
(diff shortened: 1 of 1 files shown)


pkg/ebpf/c/common/filtering.h
```diff
@@ -357,6 +357,9 @@ statfunc u64 match_data_filters(program_data_t *p, u8 index)
     u64 explicit_disable_policies = 0;
     u64 explicit_enable_policies = 0;
     u64 default_enable_policies = 0;
+    // Determine policies that do not use any type of string filter (exact, prefix, suffix)
+    u64 mask_no_str_filter_policies =
+        ~str_filter->exact_enabled & ~str_filter->prefix_enabled & ~str_filter->suffix_enabled;
```
geyslan, line 362: In the future, this could potentially be precomputed in userspace. Does the string_filter_config_t struct have any unused padding that could accommodate it? If so, precomputing this could also optimize the check at line 353. That flag could represent either enabled or disabled states, effectively encapsulating both scenarios. This would streamline the logic by providing a single source of truth for the checks.

pkg/ebpf/c/common/filtering.h
```diff
@@ -434,8 +437,11 @@ statfunc u64 match_data_filters(program_data_t *p, u8 index)
     // 2. Default Enabled Policies: Policies that are enabled by default (default_enable_policies)
     // remain enabled only if they are not explicitly disabled (explicit_disable_policies).
     res = explicit_enable_policies | (default_enable_policies & ~explicit_disable_policies);
+    // Combine policies that use string filters with those that do not
```
geyslan, line 440: NOTE: Not directly related to this PR, but I noticed that the logic on lines 384-386, 407-409, and 428-430 could be refactored into a helper function for better reusability and clarity.

In fact, it might be worth considering whether the [entire type match logic](https://github.com/aquasecurity/tracee/pull/4468/files#diff-c046893996f2e2a6e44a8caac7a14abbf47b47486ee5ceaa9029c34fc5f94a99R389-R410) (prefix, suffix) could be consolidated into a single helper function to streamline the code further.

yanivagman (approved):
LGTM

pkg/ebpf/c/common/filtering.h
rscampos: Good point, @geyslan. I think this optimization is worth it. Precomputing it in user space is a good idea. @yanivagman also suggested that maybe we should precompute the bitmaps in user-space (we need to do some PoCs to see if its possible). I think we can make an effort in the near future to enumerate and handle those places. WDYT?

pkg/ebpf/c/common/filtering.h
rscampos: thank you to point that... I think its possible to do some refactor for clarity... specially for each filter ending logic where there are some duplicate logic. The next phase (https://github.com/aquasecurity/tracee/issues/4433), we'll expend to other type of filters, so I'll review the string filter logic... WDYT?

geyslan: Awesome, ping me then so we can discuss further.

PR #4475, opened 2025-01-09
chore(ci): enable dependabot version updates

geyslan:

e5c4df45a **chore(ci): enable dependabot version updates**


rscampos (approved):
LGTM

I've double checked the gomod's and the places where Dockerfile is used and seems to be a good configuration.

Issue #4481, opened 2025-01-10
CodeQL "Incorrect conversion between integer types" warnings


geyslan:
## Description

https://github.com/aquasecurity/tracee/security/code-scanning

yanivagman:
@geyslan is this still an issue?

geyslan:
It's not. Closing it.

PR #4482, opened 2025-01-10
fix(filters): int conversion without check

geyslan:
Takes care of some of #4481. 

4bef591ce **fix(filters): int conversion without check**

```
This silences some CodeQL "Incorrect conversion between integer types"
warnings.
```

### DIFF
--- a/pkg/filters/binary.go
+++ b/pkg/filters/binary.go
@@ -1,6 +1,7 @@
 package filters
 
 import (
+	"math"
 	"strconv"
 	"strings"
 
@@ -32,6 +33,9 @@ func getHostMntNS() (uint32, error) {
 	ns, err = proc.GetProcNS(1, "mnt")
 	if err != nil {
 		return 0, errfmt.WrapError(err)
+	}
+	if ns < 0 || ns > math.MaxUint32 {
+		return 0, errfmt.Errorf("invalid mnt namespace %d", ns)
 	}
 
 	return uint32(ns), nil
@@ -82,6 +86,9 @@ func (f *BinaryFilter) Parse(operatorAndValues string) error {
 			} else {
 				mntNS, err := strconv.Atoi(mntAndPath[0])
 				if err != nil {
+					return InvalidValue(val)
+				}
+				if mntNS < 0 || mntNS > math.MaxUint32 {
 					return InvalidValue(val)
 				}
 				bin.MntNS = uint32(mntNS)
--- a/pkg/filters/data.go
+++ b/pkg/filters/data.go
@@ -2,6 +2,7 @@ package filters
 
 import (
 	"fmt"
+	"math"
 	"strconv"
 	"strings"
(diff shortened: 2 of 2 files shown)


pkg/ebpf/event_parameters.go
```diff
@@ -70,11 +71,16 @@ func attachSuspiciousSyscallSourceProbes(t *Tracee, eventParams []map[string]fil
 			if err != nil {
 				return err
 			}
-			if !events.Core.IsDefined(events.ID(syscallID)) {
+			if syscallID < 0 || syscallID > math.MaxInt32 {
```
NDStrahilevitz, line 74: Shouldn't this kind of check be resolved inside the IsDefined method? Is this the particular code which fixes the CodeQL issue?

geyslan: > Shouldn't this kind of check be resolved inside the IsDefined method?

IsDefined only checks for the existence of a events.ID value, it's agnostic to the origin of that value, so I believe it's sane to check it right after the Atoi conversion, since we're truncating the output (int - 64) to events.ID (int32).

> Is this the particular code which fixes the CodeQL issue?

Yep. It's an attempt. Let's see if the issue is closed automagically after this change.

geyslan: BTW, I would like to ask you if these conversions https://github.com/aquasecurity/tracee/security/code-scanning/2 are ok. I mean, if they are the full size of those values, why the API providing them uses int (or other larger types) instead?

rscampos:
@geyslan LGTM

Realized that there are other 2 places also with the same issue (https://github.com/aquasecurity/tracee/security/code-scanning):
https://github.com/aquasecurity/tracee/blob/ca5ebd4210339301d816824625ff229d224f356b/pkg/ebpf/tracee.go#L301-L303

Following the same approach:
```
    for mountNS, pid := range mntNSProcs {
        if mountNS < 0 || mountNS > math.MaxUint32 {
            return errfmt.Errorf("invalid mnt namespace %d", mountNS)
        }
        if pid < 0 || pid > math.MaxUint32 {
            return errfmt.Errorf("invalid pid %d", pid)
        }

        t.pidsInMntns.AddBucketItem(uint32(mountNS), uint32(pid))
    }
```

https://github.com/aquasecurity/tracee/blob/409b3ccf605665918f386a6265b1da87251252a9/pkg/utils/proc/ns.go#L66-L68
Following the same approach:
```
for mountNS, p := range mountNSTimeMap {
	if p.pid > math.MaxInt32 {
		return nil, errfmt.Errorf("invalid pid %d", p.pid)
	}
	mountNSToFirstProcess[mountNS] = int(p.pid)
}
```

geyslan:
I didn't touch the others since they deserve a better analysis (API wise).

rscampos (approved):
LGTM

PR #4540, opened 2025-01-21
Procfs improvements

geyslan:
Ran these in `main` and `procfs-work` branches:

`dist/tracee --metrics --pyroscope --pprof -s tree=2609608 -e sched_process_exec,sched_process_fork,sched_process_exit --proctree source=both  --proctree disable-procfs -o none`
`dist/tracee --metrics --pyroscope --pprof -s tree=2241901 -e sched_process_exec,sched_process_fork,sched_process_exit --proctree source=both  -o none`

Stressor details:

Details
`8 threads with 1_000_000 ops each` running on:

MemTotal: 64923992 kB (64GB)

```
| Metric        | NO procfs (%) | WITH procfs  (%) |
|---------------|--------------:|-----------------:|
| CPU avg       | -2.65%        | -28.46%          |
| Malloc avg    | -1.56%        | -22.60%          |
| Heap avg      | -1.94%        | +5.65%           |
| Heap obj avg  | -0.52%        | +1.82%           |
```

**NOTE:** You may notice that the `WITH procfs` Heap values ​​have increased - it is a side effect of the CPU time improvement in the proc package. Since the process termination window has been reduced (from feed request to proc file reading), more child processes have been added to the Process still living in LRU.

Close: #4547 

c5be62d6e **perf(proctree): remove stat call**
d474eeb0d **chore: introduce builders with specific fields**
c7178632a **perf(proc): improve ns**
8fc7eb389 **perf(proc): improve status file parsing**
7471ae27b **perf(proctree/proc): align fields to real size**
764ba4abd **perf(proc): improve stat file parsing**
2afe59496 **perf(proc): introduce ReadFile for /proc**
d6818c109 **perf(proc): use formatters for procfs file paths**

c5be62d6e **perf(proctree): remove stat call**

```
Calling stat on /proc/<pid> would only increase the window for
process termination between the stat call and the read of the file.

This also replaces fmt.Sprintf with string concatenation and
strconv.FormatInt for better performance.
```

d474eeb0d **chore: introduce builders with specific fields**

```
- NewProcStatFields()
- NewThreadStatFields()
- NewProcStatusFields()
- NewThreadStatusFields()
```

c7178632a **perf(proc): improve ns**

```
Reduce ProcNS memory footprint by using the right member type sizes -
namespace id is an uint32, since it is the inode number in
struct ns_common.

This change also improves the performance of GetAllProcNS(), GetProcNS()
and GetMountNSFirstProcesses().
```

8fc7eb389 **perf(proc): improve status file parsing**

```
Remove the use of library functions to parse the status file and instead
parse it manually (on the fly) to reduce the number of allocations and
improve performance.
```

7471ae27b **perf(proctree/proc): align fields to real size**

```
Propagate values based on its real size which in most cases is smaller
than int (64-bit). This change reduces the memory footprint or at least
the stress on the stack/heap.
```

764ba4abd **perf(proc): improve stat file parsing**

```
Remove the use of library functions to parse the stat file and instead
parse it manually (on the fly) to reduce the number of allocations and
improve performance.

chore(proc): align parsing of stat field with the formats size

This also align parsing sizes with the formats to avoid wrong parsing
of the stat file. The internal fields are represented aligned with the
actual kernel fields to avoid any confusion (signed/unsigned).
```

2afe59496 **perf(proc): introduce ReadFile for /proc**

```
`os.ReadFile` is not efficient for reading files in `/proc` because it
attempts to determine the file size before reading. This step is
unnecessary for `/proc` files, as they are virtual files with sizes
that are often reported as unknown or `0`.

`proc.ReadFile` is a new function designed specifically for reading
files in `/proc`. It reads directly into a buffer and is more efficient
than `os.ReadFile` because it allows tuning the initial buffer size to
better suit the characteristics of `/proc` files.

-run=^$ -tags ebpf -bench ^BenchmarkReadFile$
github.com/aquasecurity/tracee/pkg/utils/proc -benchtime=10000000x

goos: linux
goarch: amd64
pkg: github.com/aquasecurity/tracee/pkg/utils/proc
BenchmarkReadFile/ProcFSReadFile/Empty_File-32        10000000  3525 ns/op  408 B/op  4 allocs/op
BenchmarkReadFile/OsReadFile/Empty_File-32            10000000  4070 ns/op  872 B/op  5 allocs/op
BenchmarkReadFile/ProcFSReadFile/Small_File-32        10000000  3961 ns/op  408 B/op  4 allocs/op
BenchmarkReadFile/OsReadFile/Small_File-32            10000000  4538 ns/op  872 B/op  5 allocs/op
BenchmarkReadFile/ProcFSReadFile/Exact_Buffer_Size-32 10000000  4229 ns/op  920 B/op  5 allocs/op
BenchmarkReadFile/OsReadFile/Exact_Buffer_Size-32     10000000  4523 ns/op  872 B/op  5 allocs/op
BenchmarkReadFile/ProcFSReadFile_/proc/self/stat-32   10000000  4043 ns/op  408 B/op  4 allocs/op
BenchmarkReadFile/OsReadFile_/proc/self/stat-32       10000000  4585 ns/op  872 B/op  5 allocs/op
PASS
ok  	github.com/aquasecurity/tracee/pkg/utils/proc	334.751s
```

d6818c109 **perf(proc): use formatters for procfs file paths**

```
Since the type of the converted primitive is already known, formatter
helpers should be used to construct procfs file paths instead of relying
on `fmt.Sprintf`. Using `fmt.Sprintf` is relatively costly due to its
internal formatting logic, which is unnecessary for simple path
construction.
```

### DIFF
--- a/pkg/proctree/proctree_procfs.go
+++ b/pkg/proctree/proctree_procfs.go
@@ -2,8 +2,8 @@ package proctree
 
 import (
 	"fmt"
+	"math"
 	"os"
-	"strconv"
 	"sync"
 	"time"
 
@@ -40,10 +40,10 @@ func (pt *ProcessTree) feedFromProcFSLoop() {
 }
 
 // FeedFromProcFSAsync feeds the process tree with data from procfs asynchronously.
-func (pt *ProcessTree) FeedFromProcFSAsync(givenPid int) {
+func (pt *ProcessTree) FeedFromProcFSAsync(givenPid int32) {
 	if pt.procfsChan == nil {
 		logger.Debugw("starting procfs proctree loop") // will tell if called more than once
-		pt.procfsChan = make(chan int, 1000)
+		pt.procfsChan = make(chan int32, 1000)
 		pt.feedFromProcFSLoop()
 	}
 
@@ -59,7 +59,7 @@ func (pt *ProcessTree) FeedFromProcFSAsync(givenPid int) {
 }
 
 // FeedFromProcFS feeds the process tree with data from procfs.
-func (pt *ProcessTree) FeedFromProcFS(givenPid int) error {
+func (pt *ProcessTree) FeedFromProcFS(givenPid int32) error {
 	procDir := "/proc"
 
 	// If a PID is given, only deal with that process
@@ -77,7 +77,7 @@ func (pt *ProcessTree) FeedFromProcFS(givenPid int) error {
 		if !dir.IsDir() {
 			continue
 		}
-		pid, err := strconv.Atoi(dir.Name())
+		pid, err := proc.ParseInt32(dir.Name())
 		if err != nil {
 			continue
 		}
@@ -93,7 +93,7 @@ func (pt *ProcessTree) FeedFromProcFS(givenPid int) error {
 //
 
 // getProcessByPID returns a process by its PID.
-func getProcessByPID(pt *ProcessTree, givenPid int) (*Process, error) {
+func getProcessByPID(pt *ProcessTree, givenPid int32) (*Process, error) {
 	if givenPid <= 0 {
 		return nil, errfmt.Errorf("invalid PID")
 	}
@@ -120,7 +120,7 @@ func getProcessByPID(pt *ProcessTree, givenPid int) (*Process, error) {
 }
 
 // dealWithProc deals with a process from procfs.
-func dealWithProc(pt *ProcessTree, givenPid int) error {
+func dealWithProc(pt *ProcessTree, givenPid int32) error {
 	if givenPid <= 0 {
 		return errfmt.Errorf("invalid PID")
 	}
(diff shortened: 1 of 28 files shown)


rscampos:
Good ideia about replacing ReadFile :)
The benchmark results for ReadFile on /proc are close to yours. The `ns/op` improvements `16,01%`, `14,79%`,`12,76%` and `14,85%`. Tested on Arm64.

```
BenchmarkReadFile/ProcFSReadFile/Empty_File-4         	  288828	      4053 ns/op	     408 B/op	       4 allocs/op
BenchmarkReadFile/OsReadFile/Empty_File-4             	  246860	      4826 ns/op	     856 B/op	       5 allocs/op
BenchmarkReadFile/ProcFSReadFile/Small_File-4         	  244072	      5012 ns/op	     408 B/op	       4 allocs/op
BenchmarkReadFile/OsReadFile/Small_File-4             	  212805	      5882 ns/op	     856 B/op	       5 allocs/op
BenchmarkReadFile/ProcFSReadFile/Exact_Buffer_Size-4  	  236595	      5045 ns/op	     920 B/op	       5 allocs/op
BenchmarkReadFile/OsReadFile/Exact_Buffer_Size-4      	  207237	      5783 ns/op	     856 B/op	       5 allocs/op
BenchmarkReadFile/ProcFSReadFile_/proc/self/stat-4    	  242958	      4895 ns/op	     408 B/op	       4 allocs/op
BenchmarkReadFile/OsReadFile_/proc/self/stat-4        	  207411	      5749 ns/op	     856 B/op	       5 allocs/op
```

pkg/utils/proc/stat_test.go
```diff
@@ -45,14 +52,14 @@ func TestProcStatSize(t *testing.T) {
...
+var statContent = "3529367 (Isolated (((Web))) Co) S 3437358 3433422 3433422 0 -1 4194560 " +
 	"50679 0 0 0 566 643 0 0 20 0 29 0 46236871 2609160192 33222 " +
 	"18446744073709551615 94165013317536 94165014109840 140730010890672 " +
 	"0 0 0 0 16846850 1082134264 0 0 0 17 29 0 0 0 0 0 94165014122560 " +
 	"94165014122664 94165887094784 140730010895394 140730010895699 " +
-	"140730010895699 140730010898399 0\n"
+	"140730010895699 140730010898399 -1\n"
 
-func TestProcStatParsing(t *testing.T) {
+func Test_newProcStat(t *testing.T) {
```
rscampos, line 62: Wouldn't be good to add some more test cases?
With different `statContent` and maybe with different `statDefaultFields`?

pkg/utils/proc/status_test.go
```diff
@@ -72,12 +131,12 @@ func TestProcStatusParsing(t *testing.T) {
 			name: "Correct parsing of mock status file",
```
rscampos, line 131: I think maybe we should add some more tests here (if its the case).

pkg/proctree/proctree_procfs.go
```diff
@@ -308,38 +308,43 @@ func dealWithThread(pt *ProcessTree, givenPid, givenTid int32) error {
...
+		}
 		if debugMsgs {
 			logger.Debugw("dealWithProc", "pid", givenPid, "err", err)
 		}
+
 		return err
 	}
 
-	taskPath := fmt.Sprintf("/proc/%v/task", givenPid)
+	taskPath := "/proc/" + strconv.FormatInt(int64(givenPid), 10) + "/task"
```
rscampos, line 323: Shouldn't we add `getTaskPath` in formatters?

rscampos (approved):
good work!! @geyslan, LGTM

pkg/proctree/proctree_procfs.go
geyslan: I believe I didn't created it since it's the unique use. But good idea, let me bring it.

pkg/utils/proc/stat_test.go
geyslan: It would be nice having it for sure. Currently the stat (and status) is well coupled with its inner parts, so to provide different defaultFields and structs in a test, I believe that we would need to make use of 2 interfaces. I'm for maturing the idea, and if new fields are required we definitively should refactor it.

pkg/utils/proc/status_test.go
geyslan: ditto.

PR #4544, opened 2025-01-22
feat: reapply: remove default usage of parse-arguments

geyslan:
Close: #4541

This reverts #4331 back.

### 1. Explain what the PR does

c922aaa60 **Reapply "feat(sigs): refactor to use nonparsed arguments"**
f680998bc **Revert "Add workaround: Revert to using raw ... argument values in engine stage".**
1f00c689d **chore(deps): bump signatures/helpers to f57a1ebab9**

c922aaa60 **Reapply "feat(sigs): refactor to use nonparsed arguments"**

```
This reverts commit b46a2b6bde588825877ea45ec690a7dc871fe73c.
```

f680998bc **Revert "Add workaround: Revert to using raw ... argument values in engine stage".**

```
This reverts commit a481d11d2ad3eda1ea9c5a187a8574e0c46e9e33.
```

Issue #4541, opened 2025-01-22
Revert back to use raw arguments for threat detection events


yanivagman:
We should revert back the revert in PR:
https://github.com/aquasecurity/tracee/pull/4461

Before doing that, we should first update the helpers package similar to what was done here:
https://github.com/aquasecurity/tracee/pull/4345/commits/4d5e47de02ff2e1214db1a167e38ba50dd7aa98c
However, it is better not to use the parsers package this time, since it adds dependency of the helpers package on tracee itself, which we should avoid.

PR #4545, opened 2025-01-22
feat(helpers): reapply: unparsed flag helpers

geyslan:
This is part of #4541 to make possible to merge #4544.

ed62bef6a **feat(helpers): unparsed flag helpers**

```

It makes use of syscall package to get constants for the flags instead
of Tracee's parsers package.

```


geyslan:
I've cancelled the tests since the change is not reflect on main go.mod yet.

yanivagman (approved):
LGTM

PR #4548, opened 2025-01-23
fix(events): fix incorrect handling of event parameters

oshaked1:
If an event that uses parameters is enabled but not present in all policies, a nil access occurs and tracee panics. This fix makes sure the event is present in each policy before accessing its parameters.


geyslan:
@yanivagman would this conflict with yours?

yanivagman (approved):

PR #4550, opened 2025-01-23
fix(ebpf): fix insufficiently accurate detection of golang heaps

oshaked1:
The previous check resulted in false positives for regular mmap allocations (without explicit address hints), resulting in random test failures for `suspicious_syscall_source` (VMA type was detected as golang heap instead of anonymous).

Continuation of https://github.com/aquasecurity/tracee/pull/4403.


geyslan (approved):
LGTM

PR #4557, opened 2025-01-27
fix(ebpf): treat sched_process_exit corner case

geyslan:
Close: #4558

cccaf7f83 **fix(ebpf): treat sched_process_exit corner cases**

```
The sched_process_exit event may be triggered by a standard exit, such
as a syscall, or by alternative kernel paths, making it unsafe to assume
that it is always associated with a syscall exit.

do_exit and do_exit_group, while typically invoked by the exit and
exit_group syscalls, can also be reached through internal kernel
mechanisms such as signal handling. A concrete example of this occurs
when a syscall returns, enters signal handling, and subsequently calls
do_exit after get_signal. Both get_signal and do_exit involve
tracepoints.

A real execution flow illustrating this scenario in the kernel is as
follows:

entry_SYSCALL_64
  ├── do_syscall_64
  ├── syscall_exit_to_user_mode
  ├── __syscall_exit_to_user_mode_work
  ├── exit_to_user_mode_prepare
  ├── exit_to_user_mode_loop
  ├── arch_do_signal_or_restart
  ├── get_signal  (has signal_deliver tracepoint)
  ├── do_group_exit
  └── do_exit  (has sched_process_exit tracepoint)
```

### 2. Explain how to test it

Run this before on main to get sporadic errors as negative syscall numbers (triggered by signals):

`INSTTESTS="WRITABLE_DATA_SOURCE" ./tests/e2e-inst-test.sh`

After that, test this PR by running the same command above and make sure that there's no error since `sched_process_exit` will just submit `NO_SYSCALL` in such cases.

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -1521,6 +1521,24 @@ int tracepoint__sched__sched_process_exit(struct bpf_raw_tracepoint_args *ctx)
     if (!init_program_data(&p, ctx, SCHED_PROCESS_EXIT))
         return 0;
 
+    // The syscall number cannot be trusted in the following cases:
+    //
+    // 1. If the task was terminated due to a signal (PF_SIGNALED is set), the syscall
+    //    context may be inconsistent.
+    //
+    // 2. If the task was not signaled:
+    //    - A kernel thread (PF_KTHREAD is set) is not expected to have a valid syscall context, so
+    //      the function init_program_data has already set its syscall number as NO_SYSCALL (-1).
+    //    - If PF_KTHREAD is not set but the syscall value is negative, it may be due to
+    //      an invalid or clobbered context.
+    //
+    // In any of these cases, we explicitly mark the syscall number as NO_SYSCALL (-1) to avoid
+    // misinterpretation.
+    int task_flags = get_task_flags(p.event->task);
+    if ((task_flags & PF_SIGNALED) ||
+        (!(task_flags & PF_KTHREAD) && (p.event->context.syscall < 0)))
+        p.event->context.syscall = NO_SYSCALL;
+
     // evaluate matched_policies before removing this pid from the maps
     evaluate_scope_filters(&p);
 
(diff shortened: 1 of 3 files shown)


pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1521,6 +1521,11 @@ int tracepoint__sched__sched_process_exit(struct bpf_raw_tracepoint_args *ctx)
     if (!init_program_data(&p, ctx, SCHED_PROCESS_EXIT))
         return 0;
 
+    // the caller might not be an exit like syscall, but a signal handler
+    if ((p.event->context.syscall < 0) &&
```
geyslan, line 1525: Gonna see tomorrow if checking against expected syscalls is a better solution, since I suspect that we can have syscall wrongly populated as positive as well.

oshaked1:
@geyslan did you understand why -237 was showing up as the syscall number? From what I understand it should always be -1 when not in syscall context

geyslan:
> @geyslan did you understand why -237 was showing up as the syscall number? From what I understand it should always be -1 when not in syscall context

I believe this is due to kernel signal handling. These syscalls are triggered by the Go runtime, and on their return, the process group has already received a signal.

Relevant references:

[arch/x86/kernel/signal.c#L333-L341](https://github.com/torvalds/linux/blob/e2ee2e9b159094527ae7ad78058b1316f62fc5b7/arch/x86/kernel/signal.c#L333-L341)
[kernel/signal.c#L3036-L3037](https://github.com/torvalds/linux/blob/e2ee2e9b159094527ae7ad78058b1316f62fc5b7/kernel/signal.c#L3036-L3037)

image

Since `do_exit()` sets the tracepoint in the [middle](https://github.com/torvalds/linux/blob/e2ee2e9b159094527ae7ad78058b1316f62fc5b7/kernel/exit.c#L929) of its execution, the value in the register we fetch for the syscall number may not be reliable. Depending on the origin of the call, this register could contain a non-related value, potentially clobbered by the interrupted syscall or the signal handler itself.

oshaked1 (approved):
LGTM

yanivagman (approved):
LGTM

PR #4562, opened 2025-01-28
fix(epbf): fix handling of compat tasks in syscall checkers

oshaked1:
Convert to 64bit syscall ID before checking if the syscall is selected.
Additionally, remove false positive of VDSO triggered syscalls being detected as suspicious.

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -5320,6 +5320,13 @@ statfunc void check_suspicious_syscall_source(void *ctx, struct pt_regs *regs, u
     if (vma_is_file_backed(vma))
         return;
 
+    // In 32-bit compat mode, syscalls may be invoked by calling into a VDSO provided
+    // syscall handler (the vsyscall mechanism).
+    // Even the 64-bit VDSO may invoke syscalls, as a fallback mechanism.
+    // In such cases, we don't know where was the code that called into the VDSO.
+    if (vma_is_vdso(vma))
+        return;
+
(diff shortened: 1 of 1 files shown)


geyslan (approved):
LGTM

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -5320,6 +5320,12 @@ statfunc void check_suspicious_syscall_source(void *ctx, struct pt_regs *regs, u
     if (vma_is_file_backed(vma))
         return;
 
+    // In 32-bit compat mode, syscalls may be invoked by calling into a VDSO provided
+    // syscall handler (the vsyscall mechanism).
+    // In such cases, we don't know where was the code that called into the VDSO.
+    if (is_compat(task) && vma_is_vdso(vma))
```
geyslan, line 5326: Just a doubt: wouldn't full 64-bit also make use of VDSO?

oshaked1: Calling arbitrary syscalls using the VDSO is something that only exists for 32-bit x86 applications (see https://elixir.bootlin.com/linux/v6.13/source/arch/x86/entry/vdso/vdso32/system_call.S#L15). This is meant to speed up the old syscall method of `int 0x80` by potentially using the (backwards incompatible) `sysenter` instruction.

oshaked1: Turns out that also the 64-bit VDSO may invoke syscalls as a fallback mechanism, so I removed the `is_compat()` check.

PR #4563, opened 2025-01-29
feat(events): convert syscall arg to name at processing stage

oshaked1:
Since signatures now receive unparsed event arguments, conversion of syscall IDs to names became an issue since the signature has no way to know whether the event was generated on an x86_64 or ARM64 system. To solve this, we convert the syscall ID argument to its name at the event processing stage, so it happens regardless of argument parsing. This is applied to the `suspicious_syscall_source` and `stack_pivot` events.

### DIFF
--- a/pkg/ebpf/processor.go
+++ b/pkg/ebpf/processor.go
@@ -131,6 +131,8 @@ func (t *Tracee) registerEventProcessors() {
 	t.RegisterEventProcessor(events.PrintMemDump, t.processTriggeredEvent)
 	t.RegisterEventProcessor(events.PrintMemDump, t.processPrintMemDump)
 	t.RegisterEventProcessor(events.SharedObjectLoaded, t.processSharedObjectLoaded)
+	t.RegisterEventProcessor(events.SuspiciousSyscallSource, t.convertSyscallIDToName)
+	t.RegisterEventProcessor(events.StackPivot, t.convertSyscallIDToName)
 
 	//
 	// Uprobe based events processors
--- a/pkg/ebpf/processor_funcs.go
+++ b/pkg/ebpf/processor_funcs.go
@@ -487,3 +487,25 @@ func (t *Tracee) removeIrrelevantContext(event *trace.Event) error {
 
 	return nil
 }
+
+func (t *Tracee) convertSyscallIDToName(event *trace.Event) error {
+	syscallArg := events.GetArg(event, "syscall")
+	if syscallArg == nil {
+		return errfmt.Errorf("cannot find syscall argument")
+	}
+
+	syscallID, ok := syscallArg.Value.(int32)
+	if !ok {
+		return errfmt.Errorf("cannot convert syscall arg to ID")
+	}
+	syscallDef := events.Core.GetDefinitionByID(events.ID(syscallID))
+	// no need to check for NotValid() since it is syscall only if it's a valid event
+	if !syscallDef.IsSyscall() {
+		return errfmt.Errorf("invalid syscall ID %d", syscallID)
+	}
+
+	syscallArg.Type = "string"
+	syscallArg.Value = syscallDef.GetName()
+
+	return nil
+}
--- a/pkg/events/core.go
+++ b/pkg/events/core.go
@@ -13075,7 +13075,7 @@ var CoreEvents = map[ID]Definition{
 		},
 		sets: []string{},
 		fields: []trace.ArgMeta{
-			{Type: "int", Name: "syscall"},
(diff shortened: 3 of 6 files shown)


geyslan:
As discussed with @oshaked1 offline, this is required to consume tracee output through analyze, in a context that analyze (signatures) doesn't know the architecture from which the events were originated. I just wonder if providing that info in the event wouldn't be better - avoiding translations (int -> string -> int). RFC @yanivagman @NDStrahilevitz @rscampos.

NDStrahilevitz:
> As discussed with @oshaked1 offline, this is required to consume tracee output through analyze, in a context that analyze (signatures) doesn't know the architecture from which the events were originated. I just wonder if providing that info in the event wouldn't be better - avoiding translations (int -> string -> int). RFC @yanivagman @NDStrahilevitz @rscampos.

I think providing the host info, architecture in particular, is the solution and what immediately came to mind when I read the issue now.

yanivagman:
We actually have a solution exactly to that problem in the grpc server.
We assign a unique event ID to tracee events, so no matter what the CPU architecture is, the event id is the same:
https://github.com/aquasecurity/tracee/blob/main/pkg/server/grpc/tracee.go#L24

oshaked1:
> We actually have a solution exactly to that problem in the grpc server. We assign a unique event ID to tracee events, so no matter what the CPU architecture is, the event id is the same: https://github.com/aquasecurity/tracee/blob/main/pkg/server/grpc/tracee.go#L24
> 
> Once we will change the pipeline structure to use the protobuf struct, we can assign this event id in the decode stage

@yanivagman how would that help with parsing syscall IDs?

NDStrahilevitz:
> > We actually have a solution exactly to that problem in the grpc server. We assign a unique event ID to tracee events, so no matter what the CPU architecture is, the event id is the same: https://github.com/aquasecurity/tracee/blob/main/pkg/server/grpc/tracee.go#L24
> > 
> > Once we will change the pipeline structure to use the protobuf struct, we can assign this event id in the decode stage
> 
> @yanivagman how would that help with parsing syscall IDs?

It would make tracee ids the syscall standard. so you consult the tracee event ids rather than the architecture syscall id standard.

I don't think this has to be the solution though short term. there are other reasons to include this kind of host info in the event, considering that tracee usually runs as a daemonset.

oshaked1:
> > > We actually have a solution exactly to that problem in the grpc server. We assign a unique event ID to tracee events, so no matter what the CPU architecture is, the event id is the same: https://github.com/aquasecurity/tracee/blob/main/pkg/server/grpc/tracee.go#L24
> > > Once we will change the pipeline structure to use the protobuf struct, we can assign this event id in the decode stage
> > 
> > 
> > @yanivagman how would that help with parsing syscall IDs?
> 
> It would make tracee ids the syscall standard. so you consult the tracee event ids rather than the (unknown) architecture syscall id standard.
> 
> I don't think this has to be the solution though short term. there are other reasons to include this kind of host info in the event, considering that tracee usually runs as a daemonset.

I'm not sure I understand... Given a syscall ID in an event field, how do I find the corresponding syscall event if it has some sort of unique ID that is unrelated to the syscall ID? Or was the intention to add the unique event ID as an event field (or replace the syscall ID with the corresponding event unique ID) in the decode stage?

yanivagman:
> > > > We actually have a solution exactly to that problem in the grpc server. We assign a unique event ID to tracee events, so no matter what the CPU architecture is, the event id is the same: https://github.com/aquasecurity/tracee/blob/main/pkg/server/grpc/tracee.go#L24
> > > > Once we will change the pipeline structure to use the protobuf struct, we can assign this event id in the decode stage
> > > 
> > > 
> > > @yanivagman how would that help with parsing syscall IDs?
> > 
> > It would make tracee ids the syscall standard. so you consult the tracee event ids rather than the (unknown) architecture syscall id standard.
> > 
> > I don't think this has to be the solution though short term. there are other reasons to include this kind of host info in the event, considering that tracee usually runs as a daemonset.
> 
> I'm not sure I understand... Given a syscall ID in an event field, how do I find the corresponding syscall event if it has some sort of unique ID that is unrelated to the syscall ID? Or was the intention to add the unique event ID as an event field (or replace the syscall ID with the corresponding event unique ID) in the decode stage?

You can compare the unique event id to the enum given in tracee API (protobuf).
Anyway, this solution I mentioned is for event id, and in the case discussed here we discuss the syscall field. So it can be given as an inspiration for how to avoid the conversion from int to string and back to int. I'm also ok with always converting the syscall field to a string

yanivagman (approved):
LGTM

Issue #4580, opened 2025-02-05
Time normalisation causing proctree hashes mismatches


geyslan:
## Description

This was detected during the last proctree optimisation effort (checking proctree entries).

Entries never populated due to hash mismatching:
 
![Image](https://github.com/user-attachments/assets/8fa91920-7a42-4c53-ae58-a4fb6e46eb9a)

Duplicated entries (pid, tid, ppid) but with different hashes:

![Image](https://github.com/user-attachments/assets/a29ff376-efa3-4580-af95-f4094f329c65)

Issue #3878, opened 2024-02-20
Error handling event by signature (PROCTREE_DATA_SOURCE)


geyslan:
## Description

https://github.com/aquasecurity/tracee/actions/runs/7980171604/job/21789214615?pr=3877#step:5:1575

`./tests/e2e-inst-test.sh` failed with 

```json
{"level":"error","ts":1708465206.4616668,"msg":"Handling event by signature Process Tree Data Source Test: process (tid: 12665 pid: 12665, ppid: 12657, time: 1708465191356487053, hash: 3194477383) no match for ppid"}
```

```
INFO: = TEST: PROCTREE_DATA_SOURCE ==============================================
INFO: 
INFO: 
INFO: UP AND RUNNING
INFO: 
INFO: 
INFO: PROCTREE_DATA_SOURCE: FAILED, stderr from tracee:
{"level":"error","ts":1708465206.4616668,"msg":"Handling event by signature Process Tree Data Source Test: process (tid: 12665 pid: 12665, ppid: 12657, time: 1708465191356487053, hash: 3194477383) no match for ppid"}
INFO: PROCTREE_DATA_SOURCE: FAILED, events from tracee:
[...]

PR #4582, opened 2025-02-08
Fix ProcessTree rogue entries

geyslan:
Close: #4580
Close: #3878

f2e413059 **chore(proctree): mov hash computation to userspace**
5d943de68 **chore(proctree): add debug comment**
71bb922af **fix(proctree): exec event already brings data**
0ed6a1a3a **fix(proctree): treat kthread as special case**
fa4bc3283 **fix(proctree): time normalization momentum**
0e9065f8f **fix(tests): kill only tracee (-x exact match)**
ab4e18441 **chore(proctree): add exittime to debug**

71bb922af **fix(proctree): exec event already brings data**

```
Populate Process TaskInfo with data from sched_process_exec event.
Without it, the ProcessTree when set as source=events will contain
entries with no relevant data.
```

0ed6a1a3a **fix(proctree): treat kthread as special case**

```
kthread was being added only as parent, so its taskinfo was not being
updated.
```

fa4bc3283 **fix(proctree): time normalization momentum**

```
ProcessTree was populated with rogue entries because time normalization
was not performed at the correct moment, leading to a mismatch between
hashes from signals and those from events.
```

### 2. Explain how to test it

Uncomment https://github.com/aquasecurity/tracee/pull/4582/files#diff-72195925b110a2e9579f4bd9491d18052a9fb1875d2b3757987cfdfc288a1cffR183 and run tracee with `--proctree source=both`, `--proctree source=events`, `--proctree source=signals` for checking the current state of the ProcessTree. One might also wanna run `INSTTESTS="PROCTREE_DATA_SOURCE" ./tests/e2e-inst-test.sh` changing the source type inside that script before.

One won't see anymore:

- Rogue entries (hashes without pid, tid or ppid)
- Different hashes for the same pid, tid and ppid
- pid 2 as a rogue entry

### DIFF
--- a/pkg/events/parse_args.go
+++ b/pkg/events/parse_args.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/aquasecurity/tracee/pkg/errfmt"
 	"github.com/aquasecurity/tracee/pkg/events/parsers"
+	traceetime "github.com/aquasecurity/tracee/pkg/time"
 	"github.com/aquasecurity/tracee/types/trace"
 )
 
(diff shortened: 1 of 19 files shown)


NDStrahilevitz:
@geyslan updating here our offline discussion:

Instead of changing the normalization ordering, we will align the timestamp arguments  between the signal and event. This will:
1. Give userspace the option to work on the same kind of timestamp, that is, userspace will have access to the kernel determined timestamp from the buffer
2. Wether the hash will be calculated from the normalized or kernel original timestamp is irrelevant to the current PR, the point is consistency
3. By submitting the timestamp from the kernel we can reduce the instruction size of the signal ebpf program since we will not hash in kernel time

Please note if i've missed or misrepresented anything by mistake

geyslan:
> By submitting the timestamp from the kernel we can reduce the instruction size of the signal ebpf program since we will not hash in kernel time

It's a trade-off... we reduce computation time but enlarge the signal buffer with more args. Anyway, I believe the latter is better.

NDStrahilevitz:
> > By submitting the timestamp from the kernel we can reduce the instruction size of the signal ebpf program since we will not hash in kernel time
> 
> It's a trade-off... we reduce computation time but enlarge the signal buffer with more args. Anyway, I believe the latter is better.

Hmmm, I was under the assumption we were submitting the pids and tids. Then, this could actually increase instruction size (-3hash,+6submits).

geyslan:
> Hmmm, I was under the assumption we were submitting the pids and tids. Then, this could actually increase instruction size (calculation is: -3hash,+3 new submits, +0(?) switching out hash submit to timestamp submit).

For each process task level relation we have to submit a pair of id and timestart instead of their hash only.

pkg/events/parse_args.go
```diff
@@ -274,24 +275,54 @@ func ParseArgsFDs(event *trace.Event, origTimestamp uint64, fdArgPathMap *bpf.BP
...
 	if arg == nil {
 		return fmt.Errorf("event %s has no argument named %s", event.EventName, argName)
 	}
 	arg.Value = value
 	return nil
 }
 
+// NormalizeTimeArgs normalizes the time arguments of an event, converting them to
+// nanoseconds since the epoch.
+func NormalizeTimeArgs(eventID ID, args []trace.Argument, timeArgNames []string) error {
```
NDStrahilevitz, line 299: nit: I don't think we should pass the eventID just for the error message in the callee, the caller should handle that.

NDStrahilevitz (commented):
Is the [this commit](https://github.com/aquasecurity/tracee/pull/4582/commits/eff68f42f64f00a7f72fa9c4fa8f8d80e77caef1) relevant given the later commit [chore(proctree): mov hash computation to userspace](https://github.com/aquasecurity/tracee/pull/4582/commits/2aabbd473148c781abb9ec4039bc18b289b0de4c)?

geyslan:
> Is the [this commit](https://github.com/aquasecurity/tracee/pull/4582/commits/eff68f42f64f00a7f72fa9c4fa8f8d80e77caef1) relevant given the later commit [chore(proctree): mov hash computation to userspace](https://github.com/aquasecurity/tracee/pull/4582/commits/2aabbd473148c781abb9ec4039bc18b289b0de4c)?

I didn't squash them because the previous fixes are atomic. This last one being atomic as well makes it easier to revert the hash computation relocation if decided for.

rscampos (approved):
LGTM

PR #4585, opened 2025-02-12
Fix clang format

geyslan:

12a7e77bb **chore(ebpf): format apply based on the latest fix**
d3fc3e14e **fix: remove duplicate entry in .clang-format**

d3fc3e14e **fix: remove duplicate entry in .clang-format**

```
PenaltyReturnTypeOnItsOwnLine was duplicated in the .clang-format file.
It was crashing the clangd server.

Since the value of PenaltyReturnTypeOnItsOwnLine was being wrongly
interpreted, this opportunistically changes the other Penalty* values.
```

### DIFF
--- a/pkg/ebpf/c/tracee.bpf.c
+++ b/pkg/ebpf/c/tracee.bpf.c
@@ -937,11 +937,10 @@ statfunc int find_modules_from_module_kset_list(program_data_t *p)
 
 #pragma unroll
     for (int i = 0; i < MAX_NUM_MODULES; i++) {
-        if (BPF_CORE_READ(n, name) ==
-            NULL) { // Without this the list seems infinite. Also, using pos
-                    // here seems incorrect as it starts from a weird member
+        // Without this the list seems infinite.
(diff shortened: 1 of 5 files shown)


geyslan:
In future updates, these could be relevant:

- https://github.com/git/git/blob/master/.clang-format
- https://github.com/torvalds/linux/blob/master/.clang-format
- https://github.com/freebsd/freebsd-src/blob/main/.clang-format

pkg/ebpf/c/tracee.bpf.c
```diff
@@ -1034,7 +1034,7 @@ int lkm_seeker_modtree_loop(struct pt_regs *ctx)
                 }
 
                 /* We have visited the node and its left subtree.
-                Now, it's right subtree's turn */
+Now, it's right subtree's turn */
```
rscampos, line 1037: Its missing 4 tabs here

rscampos (approved):
LGTM

PR #4588, opened 2025-02-13
fix: filldir64 event

rscampos:

5d6c82279 **chore: update go mod/sum**
5baa18d6c **fix: filldir64 event**

5baa18d6c **fix: filldir64 event**

```
The second parameter (`name`) of filldir64 is currently being
interpreted as the `process name`, whereas it should be
interpreted as the directory entry name (`dirent`).
```

### 2. Explain how to test it

In order to generate the event `hidden_inodes` its necessary to trigger `filldir64`. `filldir64` is triggered by `getdents64` (library function readdir). Flow: readdir -> getdents64 -> [filldir64](https://elixir.bootlin.com/linux/v6.13/source/fs/readdir.c#L345)

```
% gcc -o readdir readdir.c
% ./readdir /
```

**readdir source code:**

<details>
<summary>Source code</summary>

```c
#include <stdio.h>
#include <stdlib.h>
#include <dirent.h>

int main(int argc, char *argv[]) {
    // Check if the directory path is provided
    if (argc != 2) {
        fprintf(stderr, "Usage: %s <directory_path>\n", argv[0]);
        return EXIT_FAILURE;
    }
    // Open the directory
    DIR *dir = opendir(argv[1]);
    if (dir == NULL) {
        perror("opendir");
        return EXIT_FAILURE;
    }
    printf("Contents of directory %s:\n", argv[1]);
    struct dirent *entry;
    while ((entry = readdir(dir)) != NULL) {
        printf("Name: %s", entry->d_name);
        // Additional information
        switch (entry->d_type) {
            case DT_REG:
                printf(" (Regular file)\n");
                break;
            case DT_DIR:
                printf(" (Directory)\n");
                break;
            case DT_LNK:
                printf(" (Symbolic link)\n");
                break;
            default:
                printf(" (Other)\n");
                break;
        }
    }
    // Close the directory
    closedir(dir);
    return EXIT_SUCCESS;
}

```

</details>

```
sudo ./dist/tracee -s comm=readdir -e hidden_inodes.data.hidden_dirent=tmp
08:12:55:823956  1000   readdir          2470122 2470122 0                hidden_inodes             hidden_dirent: tmp
```

### 3. Other comments

Notes: this C example is a partial PoC because don't trigger the event unless we comment out the lines:
```c
    // only inode=0 is relevant, simple filter prior to program run
    unsigned long dirent_inode_number = (unsigned long) PT_REGS_PARM5(ctx);
    if (dirent_inode_number != 0)
        return 0;
```
For a fully PoC would be necessary to trigger the event and some how set the `inode=0` before calling `filldir64`. Maybe using a LKM to hook `filldir64` and set to 0 before passing to the original function.


rscampos:
@OriGlassman FYI

pkg/events/core.go
```diff
@@ -11900,7 +11900,7 @@ var CoreEvents = map[ID]Definition{
 		},
 		sets: []string{},
 		fields: []trace.ArgMeta{
-			{Type: "char*", Name: "hidden_process"},
+			{Type: "char*", Name: "hidden_dirent"},
```
geyslan, line 11903:
```suggestion
			{Type: "char*", Name: "dirent_name"},
```

Although the event is used to catch hidden processes, the arg field is the name of the dirent with zeroed inode.

The event, in my opinion, should be called `ZeroedInodes`, since whoever indirectly calls filldir64 may pass a zeroed inode. @NDStrahilevitz WDYT?

NDStrahilevitz: I'm not sure if research uses this event "raw" or passes it to a signature for further inspection. So their input is the relevant one. @OriGlassman @oshaked1

geyslan: It's used as data filter.

rscampos: As pointed out by @geyslan in this [reference](https://www.kernel.org/doc/html/latest/filesystems/ext4/directory.html), always assuming that inode=0 is explicitly used by an attacker to hide an inode may lead to a false positive... if unused dir uses `inode=0` as described:
```
The end of the entire directory is of course signified by reaching the end of the file. Unused directory entries are signified by inode = 0.
```

Maybe one possibility is changing the name as Geyslan suggested.

There is one helper from research which is using the field interpreting as `hidden process` whereas should be `hidden dirent` or `unused dirent` or something similar.

geyslan:
Postponing this, since we need to release v0.23.0 asap. @rscampos

rscampos:
Hey @geyslan, updated the event name as you suggested... tks
updated the API https://github.com/aquasecurity/tracee/pull/4636. After merging the API I'll able to update this PR.

geyslan (approved):
LGTM

PR #4589, opened 2025-02-13
fix(build): libbpf output

geyslan:

0172da02c **fix(build): libbpf output**

```
It also removes the 3rdparty/include/stdarg.h file, which is not needed.
```


Makefile
```diff
@@ -450,7 +467,7 @@ TRACEE_SRC_DIRS = ./cmd/ ./pkg/ ./signatures/
 TRACEE_SRC = $(shell find $(TRACEE_SRC_DIRS) -type f -name '*.go' ! -name '*_test.go')
 GO_TAGS_EBPF = core,ebpf
 CGO_EXT_LDFLAGS_EBPF =
-CUSTOM_CGO_CFLAGS = "-I$(abspath $(OUTPUT_DIR)/libbpf) -I$(LIBBPF_INCLUDE_UAPI)"
```
geyslan, line 453: We could use /usr/include here, but as our goal is to not depend on environment, let's keep using uapi provided by libbpf.

rscampos (approved):
LGTM

